PR #121, opened 2019-05-07
Problem: OpenAPI schema for Uploads is wrong

dkliban:
Solution: add multiple serializers for Uploads

This patch introduces a ViewSet that uses SwaggerAutoSchema to describe the REST API
provided by drf_chunked_uploads. This also required adding some new serializers. Since
the post and put methods provided by drf_chunked_uploads are used to perform multiple
actions, a custom Router was needed to route all the calls to the correct methods.

fixes: #4760
https://pulp.plan.io/issues/4760


pulpcore/app/serializers/content.py
```diff
@@ -218,16 +217,46 @@ class Meta:
...
     file = serializers.FileField(
         write_only=True,
+        required=True
     )
 
-    class Meta(ChunkedUploadSerializer.Meta):
+
+class UploadPOSTSerializer(base.ModelSerializer):
+    """Serializer for creating chunked uploads from entire file."""
+    file = serializers.FileField(
```
asmacdo: please add help_text

pulpcore/app/serializers/content.py
```diff
@@ -218,16 +217,46 @@ class Meta:
 
 class UploadSerializer(base.ModelSerializer):
     """Serializer for chunked uploads."""
-    viewname = 'uploads:upload-detail'
-
     _href = base.IdentityField(
         view_name='upload-detail',
     )
+    file = serializers.FileField(
+        write_only=True,
```
asmacdo: please add help text

pulpcore/app/serializers/content.py
dkliban: added.

pulpcore/app/serializers/content.py
dkliban: added

dkliban:
I've added all the helptext and the link to the REST docs.

PR #124, opened 2019-05-09
Split Distribution into two objects

### LINKED ISSUE #4785: Post-publish hook

We are part of the packages.microsoft.com team and want a more optimal way to clear certain files from our Content Delivery Network when we publish new when publishing new content

We would like a script to clear out the cache in our CDN anytime new content is published so we are able to resolve those cache issues. We would want a post-publish hook to run a script and this is most likely where the place should be in the code https://github.com/pulp/pulpcore/blob/00c885db9d866274562fd2621a23381422602f62/pulpcore/app/models/publication.py#L204

The current solution we are working on is having to monitor the publish tasks and wait for one to complete before clearing files from the CDN cache service.

bmbouter:
The Distribution contained options which should not be mixed.
Specifically the repository and repository_version options go together
and the publication goes by itself.

This PR splits that object into two new objects
RepositoryVersionDistribution and PublicationDistribution. The two
models are not detail objects, and require the plugin writer to declare
a Distribution detail object. It was already this way before this PR.

Required PR: https://github.com/pulp/pulpcore-plugin/pull/97
Required PR: https://github.com/pulp/pulp_file/pull/219

https://pulp.plan.io/issues/4785
closes #4785


pulpcore/app/viewsets/publication.py
```diff
@@ -55,52 +57,55 @@ class ContentGuardViewSet(NamedModelViewSet,
...
-            return True
-        else:
-            return super().is_master_viewset()
+
+class PublicationDistributionViewSet(BaseDistributionViewSet):
+    """
+    ViewSet for PublicationDistribution objects.
+    """
+
+    endpoint_name = 'publication'
```
asmacdo: can this be `publications` instead?

bmbouter: yes the plural is more consistent with the other endpoints. I'm also going to switch the one below as well.

pulpcore/app/viewsets/publication.py
```diff
@@ -55,52 +57,55 @@ class ContentGuardViewSet(NamedModelViewSet,
...
+                              mixins.ListModelMixin,
+                              AsyncCreateMixin,
+                              AsyncRemoveMixin,
+                              AsyncUpdateMixin):
     """
     Provides read and list methods and also provides asynchronous CUD methods to dispatch tasks
     with reservation that lock all Distributions preventing race conditions during base_path
     checking.
     """
-    endpoint_name = 'distributions'
```
dkliban: you need to bring this line back.

bmbouter: the issue is that it will be overridden by the subclasses further down. Now that we're switching the models to be abstract let's re-look at it.

pulpcore/app/serializers/publication.py
```diff
@@ -170,52 +208,18 @@ class DistributionSerializer(BaseDistributionSerializer):
...
 
-        mutex_keys = ['publication', 'repository', 'repository_version']
-        in_use_keys = []
-        for mkey in mutex_keys:
-            if mkey in data:
-                in_use_keys.append(mkey)
-
-        if len(in_use_keys) > 1:
-            msg = _("The attributes {keys} must be used exclusively.".format(keys=in_use_keys))
+        if 'repository' in data and 'repository_version' in data:
```
gmbnomis: Do you know what happens if those parameters are set to `null`? I don't know if they disappear or result in `None`. If the latter, it is better to use `data.get('repository') and data.get('repository_version')` here

pulpcore/content/handler.py
```diff
@@ -94,16 +94,16 @@ def _match_distribution(path):
...
-            Distribution: The matched distribution.
+            BaseDistribution: The matched distribution.
 
         Raises:
             PathNotResolved: when not matched.
         """
         base_paths = Handler._base_paths(path)
         try:
-            return Distribution.objects.get(base_path__in=base_paths)
+            return BaseDistribution.objects.get(base_path__in=base_paths)
```
bmbouter: This seems problematic to me. This would include custom distributions that are also subclasses of `BaseDistribution`. Isn't that an issue?

pulpcore/app/viewsets/publication.py
```diff
@@ -95,12 +99,12 @@ def async_reserved_resources(self, instance):
     @classmethod
     def is_master_viewset(cls):
```
gmbnomis: I think this method is not needed anymore. This Viewset now has an associated Model that actually is a master model.

pulpcore/content/handler.py
dkliban: I think BaseDistribution should provide a property called 'served_by_pulpcore' and default it to True.

pulpcore/app/models/publication.py
```diff
@@ -232,23 +232,35 @@ class BaseDistribution(MasterModel):
...
 
     Relations:
         publication (models.ForeignKey): Publication to be served.
+    """
+
+    publication = models.ForeignKey(Publication, null=True, on_delete=models.SET_NULL)
+
+    class Meta:
+        abstract = True
+        default_related_name = '_distributions'
```
bmbouter: I don't think L#247 and L#255 can be the same? It's working for me, but that's because there is exactly 1 concrete class definition with that default_related_name. I think additional classes with `default_related_name = '_distributions'` would cause an ambiguation error. Note that it needs to match the `_distributions` of `PublicationSerializer` [here](https://github.com/pulp/pulpcore/blob/d75e94a5e4e0a6d7ed495571f6ea00c25fe2a8bd/pulpcore/app/serializers/publication.py#L24-L29).

pulpcore/content/handler.py
bmbouter: ok so we know we don't love this, but I think I'm going to file it as a follow story since that's an additive change and bugfix. Sound ok?

pulpcore/app/models/publication.py
gmbnomis: Yes, I wondered how this works as well. Your assumption why this works makes sense. Do we need the default_related_name in RepositoryVersionDistribution at all? And if so, do we need to set it to `_distributions` as well?

pulpcore/content/handler.py
bmbouter: I filed the issue here, and I also commented on why maybe it won't happen?  https://pulp.plan.io/issues/4809#note-1

pulpcore/app/models/publication.py
bmbouter: I think we do need it because we need to represent the backref in other objects like the PublicationSerializer. I don't think these backrefs will make themselves and if we don't specify them they'll just end up with stranger names.

Since we want them all to have unique names we need to remove these and have the actual detail methods name them. If we allow these to be inherited we'll end up with duplicates for sure. I'm removing these and adding it to `FileDistribution` now.

bmbouter: They'll also have to make their serializer, e.g. `FilePublicationSerializer` include the backref name to match `default_related_name` but that's just the way it is.

dkliban: That sounds right to me.

gmbnomis: Hmm, I wonder whether the root cause is that these models are abstract. This worked with Distribution/FileDistribution/CookbookDistribution before, but Distribution was not abstract, i.e. there was an actual DB table with an id to relate to.

bmbouter: Ironically I made them abstract because when they weren't django wouldn't start complaining about the ambiguity of `_distributions` between these two models specifically. That's when I realized (a) they should be abstract and (b) these names must be unique for the backref to not raise this exception.

bmbouter: I'm about to push a version that also updates the handler some. Let's see what you think of it.

pulpcore/app/viewsets/publication.py
bmbouter: I agree let me remove it now.

pulpcore/content/handler.py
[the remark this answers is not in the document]
bmbouter: I pushed some refinements to the handler code (needed updates) but I also filed this for us to maybe do better in the future in this area:  https://pulp.plan.io/issues/4809

pulpcore/app/viewsets/publication.py
bmbouter: yup it works the same without this.

pulpcore/app/serializers/publication.py
bmbouter: They disappear when I was testing it I was testing with one or the other. I'm also going to write some test issues which will add a functional test for this aspect to give us confidence over time.

pulpcore/app/models/publication.py
gmbnomis: > django wouldn't start complaining about the ambiguity of `_distributions` between these two models specifically.

Why do you expect Django to complain in this case? My understanding is that Django 'injects' a kind of a pseudo field into the model/serializer of the model the foreign key is from. Thus, Publication, Repository, and RepositoryVersion each have a '_distributions' field that one can use in the respective serializer.

gmbnomis: I just cobbled together something: pulp_cookbook has two distributor types (one for 'cookbook', one for 'cookbook2', RepositorySerializer has a _distributions field showing related RepositoryVersionDistribution models (obviously, this does not make any sense). And:

```
$ http :24817$PUBLICATION_HREF
{
    "_created": "2019-05-09T22:30:06.686178Z",
    "_distributions": [
        "/pulp/api/v3/distributions/cookbook/cookbook/95f06efb-5331-441d-bd08-459bc806ff30/"
    ],
    "_href": "/pulp/api/v3/publications/cookbook/cookbook/e003db72-78b9-4ce1-8fb2-066d92608170/",
    "_type": "cookbook.cookbook",
    "publisher": "/pulp/api/v3/publishers/cookbook/cookbook/e0ceb4a7-2dba-4267-8454-44b7dee8f073/",
    "repository": "http://localhost:24817/pulp/api/v3/repositories/%3CRepository:%20foo%3E/",
    "repository_version": "/pulp/api/v3/repositories/8f5b97e4-5485-4b4c-9ae4-f91a14c1a0a2/versions/1/"
}

$ http :24817$REPO_HREF
{
    "_created": "2019-05-09T22:27:28.355982Z",
    "_distributions": [
        "/pulp/api/v3/distributions/cookbook/cookbook2/f38d83a7-282d-457f-ab77-9f1c79d2df54/"
    ],
    "_href": "/pulp/api/v3/repositories/8f5b97e4-5485-4b4c-9ae4-f91a14c1a0a2/",
    "_latest_version_href": "/pulp/api/v3/repositories/8f5b97e4-5485-4b4c-9ae4-f91a14c1a0a2/versions/1/",
    "_versions_href": "/pulp/api/v3/repositories/8f5b97e4-5485-4b4c-9ae4-f91a14c1a0a2/versions/",
    "description": "",
    "name": "foo"
}
```

bmbouter: I'm going to wait to merge this until I can do some testing also tomorrow morning.

@gmbnomis is the idea just to bring back `PublicationSerializer._distributions` to replace `FilePublicationSerializer._file_distributions` ?

I don't think we want to make `models.PublicationDistribution` and `RepositoryVersionDistribution` models non-abstract because then the inheritance tree becomes 3 large and with concrete tables that is too much for the ORM I think.

gmbnomis: Yes, that is is the idea. You are right, we should avoid deep inheritance levels if possible. I think we should keep it as you implemented it and make the REST API nicer by a small adaptation on the plugin side: https://github.com/pulp/pulp_file/pull/219/files#r282762667

gmbnomis:
LGTM. This improves the plugin API and REST API for Distribution a lot! And having the plugin writer implement a distributions field in the publication serializer is a minor nuisance compared to excessive joining in the content app's hot path.

pulpcore/tests/functional/api/using_plugin/test_crd_publications.py
```diff
@@ -80,7 +80,7 @@ def test_02_read_publication_with_specific_fields(self):
 
         Permutate field list to ensure different combinations on result.
         """
-        fields = ('_href', '_created', '_distributions', 'publisher')
+        fields = ('_href', '_created', '_file_distributions', 'publisher')
```
kersommoura: Every plugin will have now a different field for this? like...`_rpm_distributions` and so on.

bmbouter: I'm going to switch this to `distributions` see here for more commentary:  https://github.com/pulp/pulp_file/pull/219#discussion_r282948011

PR #127, opened 2019-05-09
containers: Update to support Pulp 3.0 rc2

mikedep333:
Update to support installation from git nightly
Also some changes necessary for pulp-operator to work
Also some badly needed fixes that were always needed

Other/included fixes/improvements:

> 
> README.md: Update container list
> List all of those that currently can be installed at container build
> time.
> 
> Update for new Pulp 3 default ports
> 
> Update commands to start services; including them matching the systemd
> unit more closely
> 
> Set DJANGO_SETTINGS_MODULE in all container images
> 
> Fix possibly applicable security vulns in images by running `yum update`
> 
> Install nightly version of Pulp (currently limited to always doing so)
> 
> run migrations
> 
> Set pulp admin password if defined
> 
> Run django-admin collectstatic at runtime (so that /var/lib/pulp
> can be a persistent volume. i.e., so that stopping a container
> does not delete all your data.)
> 

These have been successfully used with the entire operator:
https://github.com/mikedep333/pulp-operator/tree/summit-demo
And the patched redis only from here:
https://github.com/mikedep333/carafe/tree/summit-demo

Also, to clarify, these are not supposed to be in a perfect state. Just a significantly better state. Much more work remains to be done in the operator, and a moderate amount remains here. Largely because the containers are dependent on each other, they need to be used with an operator. Theoretically you could use another solution as well.

### DIFF
--- a/containers/README.md
+++ b/containers/README.md
@@ -1,6 +1,6 @@
 # Pulp 3 Containers
 
-This directory contains assets and tooling for building Pulp 3 container images. The current image is an all-in-one image with different runtime scripts to assume each of the roles: pulp-core, pulp-worker and pulp-resource-manager.
+This directory contains assets and tooling for building Pulp 3 container images. The current image is an all-in-one image with different runtime scripts to assume each of the roles: pulp-api, pulp-content, pulp-worker and pulp-resource-manager.
 
 ## Build
 
@@ -10,7 +10,9 @@ The base image can be built with the help of an Ansible script. To build the bas
 
 The image can be customized to include any number of plugins as a build argument:
 
-    ansible-playbook build.yaml -e '{"plugins": ["pulp_file", "pulp_ansible"]}'
+    ansible-playbook build.yaml -e '{"plugins": ["pulp_file", "pulp_ansible", "pulp_cookbook", "pulp_docker", "pulp_maven", "pulp_python"]}'
+
+    ansible-playbook build.yaml -e '{"plugins": ["git+https://github.com/pulp/pulp_file.git", "git+https://github.com/pulp/pulp_ansible.git", "git+https://github.com/gmbnomis/pulp_cookbook.git", "git+https://github.com/pulp/pulp_docker.git", "git+https://github.com/pulp/pulp_maven.git", "git+https://github.com/pulp/pulp_python.git"]}'
 
 ## Push Image to Registry
 
--- a/containers/images/pulp-api/Dockerfile
+++ b/containers/images/pulp-api/Dockerfile
@@ -2,9 +2,15 @@ FROM centos:7
 
 ARG PLUGINS=""
 
+# mariadb_devel needed to avoid error installing pulpcore[mysql] (post-rc1)
+# otherwise, `mysql_config` is not found
+# It also needs gcc at least, so we install the dev tools package group.
+# And it needs Python.h
 RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
+		yum -y update && \
 		yum -y install epel-release centos-release-scl && \
 		yum -y install wget git rh-python36-python-pip && \
+		yum -y install @development mariadb-devel rh-python36-python-devel && \
 		yum clean all
 
 ENV LANG=en_US.UTF-8
@@ -16,15 +22,25 @@ ENV DJANGO_SETTINGS_MODULE=pulpcore.app.settings
 RUN mkdir -p /etc/pulp
 
 RUN scl enable rh-python36 'pip install gunicorn'
-RUN scl enable rh-python36 'pip install pulpcore'
-RUN scl enable rh-python36 'pip install pulpcore-plugin pulpcore[postgres] pulpcore[mysql]'
+RUN scl enable rh-python36 'pip install git+https://github.com/pulp/pulpcore.git'
+RUN scl enable rh-python36 'pip install git+https://github.com/pulp/pulpcore-plugin.git'
+RUN scl enable rh-python36 'pip install git+https://github.com/pulp/pulpcore.git#egg=pulpcore[postgres]'
+RUN scl enable rh-python36 'pip install git+https://github.com/pulp/pulpcore.git#egg=pulpcore[mysql]'
 RUN scl enable rh-python36 'pip install $PLUGINS'
 
-RUN scl enable rh-python36 "django-admin collectstatic --noinput"
+RUN mkdir -p /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/pulpcore/app/migrations
+RUN mkdir -p /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/pulpcore/app/migrations
+RUN mkdir -p /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/pulp_file/app/migrations
+RUN mkdir -p /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/pulp_ansible/app/migrations
+RUN mkdir -p /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/pulp_cookbook/app/migrations
+RUN mkdir -p /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/pulp_docker/app/migrations
+RUN mkdir -p /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/pulp_maven/app/migrations
+RUN mkdir -p /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/pulp_python/app/migrations
 
 COPY container-assets/wait_on_postgres.py /usr/bin/wait_on_postgres.py
 COPY container-assets/wait_on_database_migrations.sh /usr/bin/wait_on_database_migrations.sh
+COPY container-assets/pulp-common-entrypoint.sh /pulp-common-entrypoint.sh
 COPY container-assets/pulp-api /usr/bin/pulp-api
 
-
+ENTRYPOINT ["/pulp-common-entrypoint.sh"]
 CMD ["/usr/bin/pulp-api"]
--- a/containers/images/pulp-api/container-assets/pulp-api
+++ b/containers/images/pulp-api/container-assets/pulp-api
@@ -2,6 +2,17 @@
 
 /usr/bin/wait_on_postgres.py
 
-scl enable rh-python36 'django-admin migrate --noinput'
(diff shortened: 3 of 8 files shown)


containers/images/pulp-worker/container-assets/pulp-worker
```diff
@@ -3,4 +3,6 @@
 /usr/bin/wait_on_postgres.py
 /usr/bin/wait_on_database_migrations.sh
 
-exec scl enable rh-python36 "rq worker --url 'redis://${REDIS_SERVICE_HOST}:${REDIS_SERVICE_PORT}' -n reserved_resource_worker_${PULP_WORKER_NUMBER}@${HOSTNAME} -w 'pulpcore.tasking.worker.PulpWorker'"
+# TODO: Set ${PULP_WORKER_NUMBER} to the Pod Number
+# In the meantime, the hostname provides uniqueness.
+exec scl enable rh-python36 "rq worker --url 'redis://${REDIS_SERVICE_HOST}:${REDIS_SERVICE_PORT}' -n reserved-resource-worker-${PULP_WORKER_NUMBER}@${HOSTNAME} -w 'pulpcore.tasking.worker.PulpWorker' -c 'pulpcore.rqconfig'"
```
bmbouter: I guess this is working, but I'm not sure if you want to mix the commandline approach w/ the `-c 'pulpcore.rqconfig` approach. Maybe one or the other? For k8s maybe commandline is better? The -c option would be set from an environment variable as the other option.

containers/images/pulp-api/Dockerfile
```diff
@@ -16,15 +22,25 @@ ENV DJANGO_SETTINGS_MODULE=pulpcore.app.settings
 RUN mkdir -p /etc/pulp
 
 RUN scl enable rh-python36 'pip install gunicorn'
-RUN scl enable rh-python36 'pip install pulpcore'
-RUN scl enable rh-python36 'pip install pulpcore-plugin pulpcore[postgres] pulpcore[mysql]'
+RUN scl enable rh-python36 'pip install git+https://github.com/pulp/pulpcore.git'
+RUN scl enable rh-python36 'pip install git+https://github.com/pulp/pulpcore-plugin.git'
+RUN scl enable rh-python36 'pip install git+https://github.com/pulp/pulpcore.git#egg=pulpcore[postgres]'
```
bmbouter: I guess we want both to be baked in is that the idea?

containers/images/pulp-api/container-assets/pulp-common-entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
+
+exec "$@"
```
bmbouter: I guess this is useful?

containers/images/pulp-api/Dockerfile
mikedep333: That's what we did previously I think when we installed from pip. This is similar to the discussion over how to handle pulp's content plugins with containers.

containers/images/pulp-api/container-assets/pulp-common-entrypoint.sh
mikedep333: The last line is necessary for an entrypoint.

The entrypoint is how you are supposed to set env vars for example. I am going to look into just doing the scl sourcing (a little bit more complex) in the entrypoint as well later on.

containers/images/pulp-worker/container-assets/pulp-worker
mikedep333: I am not sure. I am under the impression that multiple settings get set via pulpcore.rqconfig, but the commandline is overriding a small number of them. (Reading more about redis is something on my to-do list.)

containers/images/pulp-api/container-assets/pulp-common-entrypoint.sh
mikedep333: Also, in previous versions of this branch (before I squashed the commits), I did even more in the entrypoint.

bmbouter: That is fine we can improve it over time. Thanks for the info.

containers/images/pulp-worker/container-assets/pulp-worker
bmbouter: There are six and they get set here:  https://github.com/pulp/pulpcore/blob/master/pulpcore/rqconfig.py#L4

containers/README.md
```diff
@@ -10,7 +10,9 @@ The base image can be built with the help of an Ansible script. To build the bas
 
 The image can be customized to include any number of plugins as a build argument:
 
-    ansible-playbook build.yaml -e '{"plugins": ["pulp_file", "pulp_ansible"]}'
+    ansible-playbook build.yaml -e '{"plugins": ["pulp_file", "pulp_ansible", "pulp_cookbook", "pulp_docker", "pulp_maven", "pulp_python"]}'
+
+    ansible-playbook build.yaml -e '{"plugins": ["git+https://github.com/pulp/pulp_file.git", "git+https://github.com/pulp/pulp_ansible.git", "git+https://github.com/gmbnomis/pulp_cookbook.git", "git+https://github.com/pulp/pulp_docker.git", "git+https://github.com/pulp/pulp_maven.git", "git+https://github.com/pulp/pulp_python.git"]}'
```
ehelms: Might be cleaner to encourage users to create a vars file and use it.

containers/images/pulp-api/Dockerfile
```diff
@@ -2,9 +2,15 @@ FROM centos:7
...
 
+# mariadb_devel needed to avoid error installing pulpcore[mysql] (post-rc1)
+# otherwise, `mysql_config` is not found
+# It also needs gcc at least, so we install the dev tools package group.
+# And it needs Python.h
 RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
+		yum -y update && \
 		yum -y install epel-release centos-release-scl && \
 		yum -y install wget git rh-python36-python-pip && \
+		yum -y install @development mariadb-devel rh-python36-python-devel && \
```
ehelms: Given this, we should consider using a multi-stage build container to avoid the development bits being needed in the production container and minimizing the size of the resulting image. Where are you tracking issues for container related deployments? I can file things there (as I would also ike to file an issue to consider switching to Universal Base Image (UBI) 8.)

containers/images/pulp-api/container-assets/pulp-api
```diff
@@ -2,6 +2,17 @@
 
 /usr/bin/wait_on_postgres.py
 
-scl enable rh-python36 'django-admin migrate --noinput'
+# Generating /var/lib/pulp/static at runtime rather than at container build time
+# facilitates all of /var/lib/pulp being a separate volume.
+scl enable rh-python36 "django-admin collectstatic --noinput"
```
ehelms: I read the comment but I am still struggling with why generate static files on container start/restart. These files seem like they should be... well static for an installation.

containers/images/pulp-api/container-assets/pulp-api
```diff
@@ -2,6 +2,17 @@
...
 
-exec scl enable rh-python36 "gunicorn -b 0.0.0.0:8000 pulpcore.app.wsgi:application"
+#TODO: Determine list of installed plugins by inspecting image contents
+scl enable rh-python36 "django-admin makemigrations file ansible cookbook docker maven python"
```
ehelms: Are plugins not yet shipping their migrations? Even so, this feels like a build time effort not a runtime effort.

containers/images/pulp-api/container-assets/pulp-common-entrypoint.sh
ehelms: I tend not to see the value in this entrypoint over the basic entrypoint.

containers/images/pulp-api/container-assets/pulp-common-entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
```
ehelms: Does this value ever get set to anything different? If not, can it be set on the images during buildtime?

containers/images/pulp-content/container-assets/pulp-content
```diff
@@ -3,4 +3,8 @@
 /usr/bin/wait_on_postgres.py
 /usr/bin/wait_on_database_migrations.sh
 
-exec scl enable rh-python36 "gunicorn -b 0.0.0.0:8000 pulpcore.content:server"
+exec scl enable rh-python36 "gunicorn pulpcore.content:server \
+--bind 0.0.0.0:24816 \
+--worker-class 'aiohttp.GunicornWebWorker' \
+-w 2 \
```
ehelms: This could be `$THREADS` or something similar to allow customization at runtime depending on a users setup.

containers/images/pulp-api/container-assets/pulp-api
mikedep333: Not yet. The bigger problem is that makemigrations requires access to the db. So it has to be runtime, not build time.

containers/images/pulp-api/container-assets/pulp-api
mikedep333: I agree, but I cannot think of a better way of handling it that isn't a lot of work up front / maintenance. Multiple subdirs under /var/lib/pulp/ **need** to be persistent storage. This one **should** be static and at build time.

containers/images/pulp-api/Dockerfile
mikedep333: I think these should be installed only when we install from any git/nightly source. That is an improvement I'd like to make later on; auto-detection for whether these packages are needed.

containers/README.md
mikedep333: I agree, I'll do that in future PRs.

containers/images/pulp-api/container-assets/pulp-common-entrypoint.sh
mikedep333: Once I move/adapt the scl command into it, I think it will pay off for reducing duplicate code.

containers/images/pulp-api/container-assets/pulp-common-entrypoint.sh
mikedep333: My bad. I didn't know about the ENV syntax. I'm adding this to a list of future improvements.

PR #135, opened 2019-05-16
Problem: REST API does not accept null for string values

dkliban:
Solution: update serializers to accept null

re: #4676
https://pulp.plan.io/issues/4676

### DIFF
--- a/docs/release-notes/pulpcore/3.0.x.rst
+++ b/docs/release-notes/pulpcore/3.0.x.rst
@@ -1,6 +1,18 @@
 Pulp 3.0 Release Notes
 ======================
 
+3.0.0rc3
+========
+
+`Comprehensive list of changes and bugfixes for rc 3 <https://github.com/pulp/pulpcore/compare/3.0.0rc2...3.0.0rc3>`_.
+
+Breaking Changes
+----------------
+
+All the string fields in the REST API no longer accept an empty string as a value. These fields
+now accept null instead.
+
+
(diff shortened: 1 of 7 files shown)


bmbouter:
Are there any alternatives to this implementation? I'm concerned from the [DRF docs](https://www.django-rest-framework.org/api-guide/fields/#charfield) that say "doing so means that there will be two differing types of empty value permissible for string representations, which can lead to data inconsistencies and subtle application bugs."

dkliban:
The alternative is to not accept empty strings and only accept null.

bmbouter:
Adding WIP since we have a meeting to discuss tomorrow w/ @jlsherrill

dkliban:
@bmbouter This is  ready for re-review.

bmbouter:
@dkliban actually it needs a breaking change 3.0 release note...

docs/release-notes/pulpcore/3.0.x.rst
```diff
@@ -1,6 +1,18 @@
...
 
+3.0.0rc3
+========
+
+`Comprehensive list of changes and bugfixes for rc 3 <https://github.com/pulp/pulpcore/compare/3.0.0rc2...3.0.0rc3>`_.
+
+Breaking Changes
+----------------
+
+All the string fields in the REST API no longer accept an empty string as a value. These fields
```
bmbouter: this looks great, thanks. +1 to merge

PR #138, opened 2019-05-20
Add flake8-docstrings to test_requirements

mdellweg:
When running flake8 linting without this package, flake reports less
errors than the automatic ci pipeline. Putting i here will ensure, that
is is installed on all pulplift development boxes.

[noissue]

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -3,6 +3,7 @@ asynctest
 codecov
 coverage
 flake8
+flake8-docstrings
 mock
 pytest
 git+https://github.com/PulpQE/pulp-smash.git#egg=pulp-smash


mdellweg:
Hmm, seems, like i was wrong. But it was definitely true for deb plugin, where the docstrings are linted.

bmbouter:
@mdellweg so what should we do with this PR?

mdellweg:
Good question. For starters, i think it is odd, that the enforcing of some linting rules is solely dependent on the existence of a package in the system. But that is the part we cannot control.

On the other side, I think it would be beneficial to follow the same linting policy throughout the whole project. Being it with or without docstring linting should be up to discussion.

The third thing the original problem shows, is that we should install dev&test requirements of plugins in the source boxes of pulplift. I can have a look into that.

kersommoura:
For pulp-smash we used to have a dev section on setup.py with all linters, and a Makefile to call those linters.

https://github.com/PulpQE/pulp-smash/blob/f14a3d7bd391761c5f22bd908555b22734c6e261/setup.py#L51

A time ago we change to pre-commit with  pre-commit hooks - pre-commit will install the packages for the required hooks in a separate env and run them. No need to add the package as dependency.

See: https://github.com/PulpQE/pulp-smash/commit/34ac7e24c1da6cc0146734132e786fec3eca6b0a

bmbouter:
@dkliban this thread has some of the linting ideas that we may want to incorporate into the managed CI tooling.

bmbouter:
@mdellweg Travis fails with ~800 linting errors with this type of linting enabled. If we are to merge this we'll have to have all that done. What do you want to do? I'm ok w/ whatever you want (or don't want) to do.

mdellweg:
@bmbouter That depends on whether we want to lint those docstrings or not. But we can close the ticket here an take the discussion to the mailinglist if you prefer.

bmbouter:
@mdellweg I'm +1 on us linting those docstrings too. Those become part of our docs and if we can increase consistency that would be a good thing. Some mailing list discussion would be good since AIUI this change would show up with new Travis errors in any CI pipeline that uses pulpcore (which is all of them). Is that right?

mdellweg:
I am not sure. In pulp_deb and pulp_ansible at least they are already active in the travis pipeline (that was my original problem to understand why they did not pop up locally).
But with the new pipelines you mentioned i have even less of an idea.

bmbouter:
I see. Well then it's probably safe to add here if someone is willing to make all the changes. The big win I hope for is for us to add this type of linting to the managed CI pipeline @dkliban is starting.

mdellweg:
@bmbouter Moved the discussion to the mailinglist. Shall we close this one?

bmbouter:
+1 to closing this for now, and we can bring it back w/ a plan to resolve the epic linting issues it discovers.

PR #146, opened 2019-05-21
Add basic CRUD tests to distributions

kersommoura:
Add basic CRUD tests to distributions.

ref:#4862

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/test_distributions.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_distributions.py
@@ -1,0 +1,88 @@
+# coding=utf-8
+"""Tests that perform actions over distributions."""
+import unittest
+
+from pulp_smash import api, config, utils
+from pulp_smash.pulp3.utils import gen_distribution
+
+from pulpcore.tests.functional.api.using_plugin.constants import (
+    FILE_DISTRIBUTION_PATH,
+)
+from pulpcore.tests.functional.api.using_plugin.utils import skip_if
+from pulpcore.tests.functional.api.using_plugin.utils import set_up_module as setUpModule  # noqa
+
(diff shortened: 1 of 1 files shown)


kersommoura:
More tests will be added around this endpoint. https://docs.pulpproject.org/en/3.0/nightly/restapi.html#tag/distributions

pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -0,0 +1,88 @@
...
+from pulp_smash.pulp3.utils import gen_distribution
+
+from pulpcore.tests.functional.api.using_plugin.constants import (
+    FILE_DISTRIBUTION_PATH,
+)
+from pulpcore.tests.functional.api.using_plugin.utils import skip_if
+from pulpcore.tests.functional.api.using_plugin.utils import set_up_module as setUpModule  # noqa
+
+
+class CRUDFileDistributionTestCase(unittest.TestCase):
```
daviddavis: s/FileDistribution/PublicationDistribution/. Instead of saying that you are testing FileDistributions, I would say that you are testing PublicationDistributions. The end goal is not to test pulp_file's distributions but rather pulpcore's PublicationDistribution. It would also be nice to test RepositoryVersionDistribution too but not sure how you would do that.

kersommoura: @daviddavis thanks for your review. We will add test for https://pulp.plan.io/issues/4839, https://pulp.plan.io/issues/4846 what it seems to be covering what you suggested.  Idea was to add basic test cover the distributions endpoint. Since we did not have the basic CRUD tests for this one. I will update per your comments.

PR #150, opened 2019-05-22
Adding checksum serializer for content

daviddavis:
refs #4827
https://pulp.plan.io/issues/4827

See https://github.com/pulp/pulp_file/pull/233 for usage.

### DIFF
--- a/pulpcore/app/serializers/content.py
+++ b/pulpcore/app/serializers/content.py
@@ -91,6 +91,51 @@ class MultipleArtifactContentSerializer(BaseContentSerializer):
         fields = BaseContentSerializer.Meta.fields + ('_artifacts',)
 
 
+class ContentChecksumSerializer(serializers.Serializer):
+    """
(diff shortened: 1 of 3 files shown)


pulpcore/app/serializers/content.py
```diff
@@ -91,6 +91,51 @@ class Meta:
...
+        checksum='sha384',
+    )
+
+    sha512 = fields.ContentArtifactChecksumField(
+        help_text=_("The SHA-512 checksum if available."),
+        checksum='sha512',
+    )
+
+    class Meta:
+        model = models.Artifact
```
gmbnomis: Sorry for being too late on this, but the model is not Artifact here (`ContentArtifactChecksumField` expects Content). It works anyway for the intended use (as subclasses will overwrite model), but I think it is misleading.

daviddavis: Opened https://github.com/pulp/pulpcore/pull/151

PR #153, opened 2019-05-24
Add test to create publication using repo version

kersommoura:
Add test to create publication only using repository version.
Add test to create publication using repository that does not have any
version.

closes:#4874


kersommoura:
@rochacbruno, any comment?

PR #154, opened 2019-05-24
Expand create publication distribution test case

kersommoura:
Expand create publication distribution test case. Use a publication
field when creating a distribution. Update assertions about just created
distribution.

closes:#4839


kersommoura:
@rochacbruno, any comment?

PR #155, opened 2019-05-24
Update response_handler used by create_file_publication

kersommoura:
Update response_handler used by create_file_publication. Then the
parsing of returned tasks will be done by the response_handler, not the
function itself.

'[noissue]'


kersommoura:
@rochacbruno, any comment?

PR #157, opened 2019-05-25
Add serializer information to status api response

mdellweg:
[noissue]


mdellweg:
Reminder to self: Refresh more often.

PR #158, opened 2019-05-28
Add back test for overlapping `base_path`

kersommoura:
Test overlapping of base_path when creating distributions.

Assuming that a distributions base_path is foo/bar. Assert that
this base_path or part of it can not be used as a new base_path for
a different distribution. No overlapping.

closes:#4882

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/test_distributions.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_distributions.py
@@ -168,3 +168,92 @@ class CRUDPublicationDistributionTestCase(unittest.TestCase):
         # Verify the update
         distribution = self.client.get(self.distribution['_href'])
         self.assertEqual(string, distribution[attr], self.distribution)
+
+
+class DistributionBasePathTestCase(unittest.TestCase):
+    """Test possible values for ``base_path`` on a distribution.
+
+    This test targets the following issues:
+
+    * `Pulp #2987 <https://pulp.plan.io/issues/2987>`_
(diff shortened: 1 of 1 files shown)


pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -86,3 +88,92 @@ def do_partially_update_attr(self, attr):
...
+        body['base_path'] = body['base_path'].replace('-', '/')
+        distribution = cls.client.post(FILE_DISTRIBUTION_PATH, body)
+        cls.distribution = cls.client.get(distribution['_href'])
+
+    @classmethod
+    def tearDownClass(cls):
+        """Clean up resources."""
+        cls.client.delete(cls.distribution['_href'])
+
+    def test_spaces(self):
```
rochacbruno: By the name of the test case `test_spaces` I can't tell if this is a **positive** or a **negative** test, what do you think about naming functions more explicitly like: 

- test_cannot_use_spaces
- test_negative_create_with_spaces (I usually follow this pattern)
- test_spaces_are_not_allowed

It make the function name longer but it is more easy to understand the scenario when analyzing the results on Jenkins.

bherrin3: I would prefer adding the extra string into the name only when it is a failure case (assume all other tests are positive cases). 

I prefer ``test_negative_foo``, if asked.

kersommoura:
Update per comments.

kersommoura:
Closing in favor of https://github.com/pulp/pulpcore/pull/159

PR #166, opened 2019-06-06
Verify that content is served from publication distribution

kersommoura:
Assert that published metadata and content is served from a publication
distribution.

closes:#4847

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/test_distributions.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_distributions.py
@@ -1,21 +1,29 @@
 # coding=utf-8
 """Tests that perform actions over distributions."""
+import csv
+import hashlib
 import unittest
+from functools import reduce
+from urllib.parse import urljoin
+
+from requests.exceptions import HTTPError
 
 from pulp_smash import api, config, utils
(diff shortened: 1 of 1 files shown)


pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -257,3 +265,83 @@ def try_update_distribution(self, **kwargs):
...
+        """Verify that content is served over publication distribution."""
+        repo = self.client.post(REPO_PATH, gen_repo())
+        self.addCleanup(self.client.delete, repo['_href'])
+
+        remote = self.client.post(FILE_REMOTE_PATH, gen_file_remote())
+        self.addCleanup(self.client.delete, remote['_href'])
+
+        sync(self.cfg, remote, repo)
+        repo = self.client.get(repo['_href'])
+        last_version_href = get_versions(repo)[-1]['_href']
```
rochacbruno: Question: is this `get_versions` function needed? repo does not have a `_latest_version_href` field?

kersommoura: I will check this one.

PR #172, opened 2019-06-11
Allow users to remove all content units from repo versions

daviddavis:
fixes #4901
https://pulp.plan.io/issues/4901

### DIFF
--- a/pulpcore/app/serializers/repository.py
+++ b/pulpcore/app/serializers/repository.py
@@ -246,12 +246,17 @@ class RepositoryVersionSerializer(ModelSerializer, NestedHyperlinkedModelSeriali
 
 class RepositoryVersionCreateSerializer(ModelSerializer, NestedHyperlinkedModelSerializer):
     add_content_units = serializers.ListField(
-        help_text=_('A list of content units to add to a new repository version'),
-        write_only=True
+        help_text=_('A list of content units to add to a new repository version. This content is '
+                    'added after remove_content_units are removed.'),
+        write_only=True,
+        required=False
     )
     remove_content_units = serializers.ListField(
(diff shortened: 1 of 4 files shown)


pulpcore/app/serializers/repository.py
```diff
@@ -246,11 +246,14 @@ class Meta:
...
 class RepositoryVersionCreateSerializer(ModelSerializer, NestedHyperlinkedModelSerializer):
     add_content_units = serializers.ListField(
-        help_text=_('A list of content units to add to a new repository version'),
+        help_text=_('A list of content units to add to a new repository version. This content is '
+                    'added after remove_content_units are removed.'),
         write_only=True
     )
     remove_content_units = serializers.ListField(
-        help_text=_('A list of content units to remove from the latest repository version'),
+        help_text=_("A list of content units to remove from the latest repository version. "
```
bmbouter: I imagined that `['*']` would be valid but `['someurlhere/UUID', '*']` to not be. It's not a big deal but since all other entries are ignored perhaps this should fail hard? what do you think?

pulpcore/app/tasks/repository.py
```diff
@@ -82,12 +82,16 @@ def add_and_remove(repository_pk, add_content_units, remove_content_units, base_
             as the initial set of content for our new RepositoryVersion
     """
     repository = models.Repository.objects.get(pk=repository_pk)
+    latest = models.RepositoryVersion.latest(repository)
```
bmbouter: Maybe move this down to L#93 in the if? (not required)

pulpcore/app/serializers/repository.py
daviddavis: Done. Can you check out the change?

PR #176, opened 2019-06-17
Problem: travis config can't be updated using travis config generator

dkliban:
Solution: use travis config generator to configure travis for pulpcore

[noissue]


.travis/script.sh
```diff
@@ -26,31 +29,40 @@ wait_for_pulp() {
...
 if [ "$TEST" = 'docs' ]; then
-  django-admin runserver 24817 >> ~/django_runserver.log 2>&1 &
   sleep 5
   cd docs
   make html
+  cd ..
+
+    if [ -x $POST_DOCS_TEST ]; then
+        $POST_DOCS_TEST
+    fi
```
daviddavis: Why is this code block indented?

dkliban: The template was indented like that. I'll fix the template and this will get updated with the next push.

PR #178, opened 2019-06-19
Problem: Pulp's OpenAPI schema has verbose Operation Ids

dkliban:
Solution: add ability to request shorter operation ids in the schema
    
This patch adds ability to specify 'bindings' query parameter when requesting the OpenAPI
schema. When specified, a short OperationId is used for each operation.
    
This patch also changes how the OpenAPI schema operations are grouped together. Each resource
now has its own tag associated with it. This results in better REST API docs and better
experience for users of bindings.
    
The tag for a resource is determined from it's URL. The plugin name is omitted from the tag. For
example:
    
URL: /pulp/api/v3/repositories/
Tag: Repositories
    
URL: /pulp/api/v3/repositories/uuid-of-some-sort/versions/
Tag: Repositories: Versions
    
URL: /pulp/api/v3/content/file/files/
Tag: Content: Files
    
URL: /pulp/api/v3/remotes/file/file/
Tag: Remotes: File
    
Required PR: https://github.com/pulp/pulp-openapi-generator/pull/18
    
re: #4989
https://pulp.plan.io/issues/4989

### DIFF
--- a/.travis/publish_client_pypi.sh
+++ b/.travis/publish_client_pypi.sh
@@ -12,7 +12,7 @@ if [[ $DESCRIPTION == 'tags/'$REPORTED_VERSION ]]; then
   export VERSION=${REPORTED_VERSION}
 else
   export EPOCH="$(date +%s)"
-  export VERSION=${REPORTED_VERSION}.dev.${EPOCH}
+  export VERSION=${REPORTED_VERSION}.${EPOCH}
 fi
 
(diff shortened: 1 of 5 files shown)


mdellweg:
I like the resulting bindings.
And if two plugins provide a content type packages, it is still namespaced by the python client module, i guess ?!?

dkliban:
> I like the resulting bindings.
> And if two plugins provide a content type packages, it is still namespaced by the python client module, i guess ?!?

Yes, even if it is called the same thing, the object is provided by a different package.

PR #182, opened 2019-06-21
Add support for parallel chunks and sha256

daviddavis:
Also removed drf-chunked-upload.

fixes #4488,#4486


docs/workflows/upload-publish.rst
```diff
@@ -5,20 +5,20 @@ Chunked Uploads
 ---------------
 
 For large file uploads, Pulp provides an `Uploads API <../../restapi.html#tag/uploads>`_. To begin
-uploading a file in chunks, an initial PUT request must be sent to the ``/pulp/api/v3/uploads``
+uploading a file in chunks, an initial POST request must be sent to the ``/pulp/api/v3/uploads``
 endpoint::
 
-    http --form PUT :24817/pulp/api/v3/uploads/ file@./chunkaa 'Content-Range:bytes 0-6291455/32095676'
+    http POST :24817/pulp/api/v3/uploads/ size=8675309
```
goosemania: How about making the total size=10485759? So the later calls are valid and not exceeding the file size.

pulpcore/app/models/upload.py
```diff
@@ -0,0 +1,68 @@
...
+    Fields:
+
+        file (models.FileField): The stored file.
+        size (models.BigIntegerField): The size of the file in bytes.
+        completed_at (models.DateTimeField): Time when the upload is committed
+    """
+
+    file = models.FileField(null=False, max_length=255)
+    size = models.BigIntegerField()
+    completed_at = models.DateTimeField(null=True)
```
goosemania: Maybe it's out of the scope of this PR, but we seem to be inconsistent with naming of datetime fields.
Some of them with `_at` and some of them without.
The only ones in pulpcore which use `_at` are `started_at` and `finished_at` on the Task model.
I don't have strong preference, and maybe `_at` is better, since it removes any ambiguity around the field being boolean or datetime. Any thoughts?

pulpcore/app/models/upload.py
```diff
@@ -0,0 +1,68 @@
...
+        Append a chunk to an upload.
+
+        Args:
+            chunk (File): Binary file to append to the upload file.
+            offset (int): First byte position to write chunk to.
+        """
+        if not self.file:
+            self.file.save(os.path.join('upload', str(self.pk)), ContentFile(''))
+
+        self.file.open(mode='r+b')
```
goosemania: It seems like file can be open many times and I saw `close` only when user commits the upload (in `sha256` property).
Is it intentional, do I miss anything?

pulpcore/app/viewsets/upload.py
```diff
@@ -1,84 +1,95 @@
...
+from pulpcore.app.serializers import UploadChunkSerializer, UploadCommitSerializer, UploadSerializer
+from pulpcore.app.viewsets.base import NamedModelViewSet
 
 
-class UploadViewSet(GenericViewSet, ChunkedUploadView, CreateModelMixin, DestroyModelMixin):
+class UploadViewSet(NamedModelViewSet,
+                    mixins.CreateModelMixin,
+                    mixins.RetrieveModelMixin,
+                    mixins.UpdateModelMixin,
+                    mixins.ListModelMixin):
```
goosemania: Is DestroyModelMixin left on purpose? For the other story that will deal with removal of incomplete chunked uploads?

goosemania:
@daviddavis , I forgot to mention CHANGES. I think it makes sense to add this issue there for the release notes. Thank you!

docs/workflows/upload-publish.rst
```diff
@@ -5,20 +5,20 @@ Chunked Uploads
...
 
 This returns an upload href (e.g. ``/pulp/api/v3/uploads/a8b5a7f7-2f22-460d-ab20-d5616cb71cdd/``) that can
 be used for subsequent chunks::
 
-    http --form PUT :24817/pulp/api/v3/uploads/a8b5a7f7-2f22-460d-ab20-d5616cb71cdd/ file@./chunkbb 'Content-Range:bytes 6291456-10485759/32095676'
+    http --form PUT :24817/pulp/api/v3/uploads/a8b5a7f7-2f22-460d-ab20-d5616cb71cdd/ file@./chunk1 'Content-Range:bytes 6291456-10485759/*'
```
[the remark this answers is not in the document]
daviddavis: Ok, I do have a two chunk example at the end of the docs.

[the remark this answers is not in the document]
daviddavis: I'll add two calls to this section though.

pulpcore/app/viewsets/upload.py
daviddavis: Yes, exactly.

pulpcore/app/models/upload.py
daviddavis: No, good catch. Will fix it.

pulpcore/app/models/upload.py
```diff
@@ -0,0 +1,69 @@
...
+            chunk (File): Binary file to append to the upload file.
+            offset (int): First byte position to write chunk to.
+        """
+        if not self.file:
+            self.file.save(os.path.join('upload', str(self.pk)), ContentFile(''))
+
+        with self.file.open(mode='r+b') as file:
+            file.seek(offset)
+            file.write(chunk.read())
+            file.close()
```
goosemania: I think this is not needed, `with` takes care of that.

pulpcore/app/models/upload.py
```diff
@@ -0,0 +1,69 @@
...
+
+        self.chunks.create(offset=offset, size=len(chunk))
+
+    @property
+    def sha256(self, rehash=False):
```
goosemania: Where is `rehash` used?

daviddavis: It's not. I left it in case it was ever needed.

daviddavis:
@dkliban I removed patch. The file parameter is showing up fine for me in the schema for the put call. Not sure why it's not showing up for you?

pulpcore/app/viewsets/upload.py
daviddavis: As an FYI to anyone else reading this PR. The delete functionality has some open/unresolved discussions. See https://pulp.plan.io/issues/4988 and https://pulp.plan.io/issues/4981.

daviddavis:
So this change is blocked by https://github.com/axnsan12/drf-yasg/issues/386#issuecomment-505143807. I see two possible workarounds until the problem is fixed. Both are pretty awful:

1. We can set the parser_classes for the Upload viewset and force all requests to multipart. This would be a simple one-line change but it would prevent users from creating uploads, etc using json which would be really strange.
2. We can break out the `update` endpoint into its own view. This is more work but we could explicitly set the media type for just the update action. This is probably the better fix but it's quite a bit more work.

@goosemania @dkliban thoughts?

goosemania:
@daviddavis , can it be done at the mixins level or parsers can only be changed for the whole viewset?
E.g. would it be possible to define a custom mixins.UpdateModelMixin with a specific parser set, and include this mixin in the viewset instead of the standard one?

daviddavis:
I think only the whole viewset. I tried creating an UploadUpdate mixin that defined the update action and  `parser_classes` but it didn't work.

dkliban:
I was able to get the schema generating correctly by adding the following hack to the generator:

```
[dkliban@localhost pulpcore]$ git diff
diff --git a/pulpcore/app/openapigenerator.py b/pulpcore/app/openapigenerator.py
index 878fde0fe..02c298eef 100644
--- a/pulpcore/app/openapigenerator.py
+++ b/pulpcore/app/openapigenerator.py
@@ -253,7 +253,8 @@ class PulpAutoSchema(SwaggerAutoSchema):
     def get_operation(self, operation_keys):
         consumes = self.get_consumes()
         produces = self.get_produces()
-
+        if operation_keys[0] == 'uploads' and operation_keys[1] == 'update':
+            consumes = ['multipart/form-data', 'application/x-www-form-urlencoded']
         body = self.get_request_body_parameters(consumes)
         query = self.get_query_parameters()
         parameters = body + query

```

daviddavis:
@dkliban I added your patch but tweaked it a bit. Please review the change.

@goosemania and @dkliban travis is green and I think this PR should be good to go. Let me know if I missed anything.

pulpcore/app/models/upload.py
```diff
@@ -0,0 +1,66 @@
...
+        if getattr(self, '_sha256', None) is None or rehash is True:
+            sha256 = hashlib.sha256()
+            with self.file.open(mode='rb') as file:
+                for chunk in file.chunks():
+                    sha256.update(chunk)
+                    self._sha256 = sha256.hexdigest()
```
goosemania: Do you need to update `self._sha256` for every chunk? Does it make sense to do it once outside of the for loop?

PR #183, opened 2019-06-22
Adds content-app to status API

### LINKED ISSUE #4881: tests accessing the content host access the api host instead
As seen here:

https://github.com/pulp/pulpcore/blob/d8239beacde38fd73fa9a1d335d14d07db8a3e3c/pulpcore/tests/functional/__init__.py#L1084C33-L1084C33

When functional tests like pulp-rpm's test_download_content.py go to access the content host via download_content_unit(), they are actually accessing the API hostname instead. Since bindings_cfg.host is always the API hostname.

bmbouter:
A new model is introduced called `ContentAppStatus` is introduced which
will record status check-in records from the Content App.

The Content App will check in periodically checkin and write to the
`ContentAppStatus` object.

To tell the ContentAppStatus records apart for multiple Content Apps
each Content App now requires a name. The name is auto-selected as
'{pid}@{hostname}'.

The Status API viewset now shows a 'content_apps' section which lists
the ContentAppStatus records, showing their names and last heartbeat
time.

Adds a doc and feature changelog entries.

https://pulp.plan.io/issues/4881
closes #4881


CHANGES/4881.feature
```diff
@@ -0,0 +1,4 @@
+The Content App now heartbeat periodically and are shown in the Status API ``/pulp/api/v3/status/``
```
daviddavis: Not sure I understand this sentence. Did you mean "heartbeats" here?

pulpcore/app/models/status.py
```diff
@@ -0,0 +1,88 @@
...
+from django.utils import timezone
+
+from pulpcore.app.models import Model
+
+
+class ContentAppStatusManager(models.Manager):
+
+    def online(self):
+        """
+        Returns a queryset of ``ContentAppStatus`` objects that have checked in recently.
```
daviddavis: Minor gripe but I would probably s/have checked in recently/are online/ to prevent mixed terminology.

pulpcore/content/__init__.py
```diff
@@ -1,19 +1,46 @@
...
+
 app = web.Application()
 
 CONTENT_MODULE_NAME = 'content'
 
 
+async def _heartbeat():
+    name = '{pid}@{hostname}'.format(pid=os.getpid(), hostname=socket.gethostname())
+    heartbeat_interarrival = settings.CONTENT_APP_TTL // 4
+    i8ln_msg = _("Content App '{name}'heartbeat written, sleeping for '{interarrival}' seconds")
```
daviddavis: missing space after '{name}'

pulpcore/content/__init__.py
```diff
@@ -1,19 +1,46 @@
...
+log = logging.getLogger(__name__)
+
 app = web.Application()
 
 CONTENT_MODULE_NAME = 'content'
 
 
+async def _heartbeat():
+    name = '{pid}@{hostname}'.format(pid=os.getpid(), hostname=socket.gethostname())
+    heartbeat_interarrival = settings.CONTENT_APP_TTL // 4
```
daviddavis: Interested to know why you chose the word interarrival here vs say interval?

pulpcore/app/models/status.py
```diff
@@ -0,0 +1,88 @@
...
+        To be considered 'missing', a Content App must have a timestamp older than
+        ``SETTINGS.CONTENT_APP_TTL``.
+
+        Returns:
+            bool: True if the content app is considered missing, otherwise False
+        """
+        now = timezone.now()
+        age_threshold = now - timedelta(seconds=settings.CONTENT_APP_TTL)
+
+        return self.last_heartbeat < age_threshold
```
daviddavis: What about returning `not online()` to de-dupe logic?

CHANGES/4881.feature
bmbouter: yeah this doesn't make sense. Fixed in next push.

pulpcore/app/models/status.py
bmbouter: yup fixing now.

pulpcore/content/__init__.py
bmbouter: I'm going to switch to interval because I think that's a more commonly used term. I chose interarrival because in queueing theory to describe a traffic pattern (like heartbeats) you talk about arrival times, and the inter-arrivals between arrival times. It's a holdover from my research life I think. Regardless interval is I think a more widely understood term in this context.

pulpcore/content/__init__.py
bmbouter: Thank you. fixed in next push.

PR #184, opened 2019-06-25
Problem: client fails to publish to PyPI

dkliban:
Solution: remove the extra '.' from the version string

[noissue]


dkliban:
I don't think this is needed anymore

PR #187, opened 2019-06-27
Remove unused field Remote.validate

asmacdo:
This field was intended for plugin use, but so far plugins validate
unconditionally. For now, we remove the field from core but plugins can
add it on their own if they have a use for it.

https://pulp.plan.io/issues/4714
fixes #4714


asmacdo:
Closing to reopen as https://github.com/pulp/pulpcore/pull/192 because travis didnt pick up the commit message.

PR #190, opened 2019-06-27
Core's Remote only validates 'immediate'

bmbouter:
The core's default mode policy='immediate' is the only one that is
guaranteed by all plugins to implement. Users of plugins that haven't
enabled additional modes should receive an error when then create a
Remote.

Now users using policy != 'immediate' will receive an error.

https://pulp.plan.io/issues/4990
closes #4990

### DIFF
--- a/pulpcore/app/serializers/repository.py
+++ b/pulpcore/app/serializers/repository.py
@@ -111,9 +111,10 @@ class RemoteSerializer(MasterModelSerializer):
         min_value=1
     )
     policy = serializers.ChoiceField(
-        help_text="The policy to use when downloading content. The possible values include: "
-                  "'immediate', 'on_demand', and 'cache_only'. 'immediate' is the default.",
-        choices=models.Remote.POLICY_CHOICES,
+        help_text="The policy to use when downloading content.",
+        choices=(
+            (models.Remote.IMMEDIATE, 'When syncing, download all metadata and content now.')
+        ),
         default=models.Remote.IMMEDIATE
     )
 
--- a/CHANGES/4990.bugfix
+++ b/CHANGES/4990.bugfix
@@ -1,0 +1,1 @@
+Core's serializer should only validate when policy='immediate' (the default).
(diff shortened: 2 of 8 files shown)


pulpcore/app/serializers/repository.py
```diff
@@ -117,6 +117,14 @@ class RemoteSerializer(MasterModelSerializer):
         default=models.Remote.IMMEDIATE
     )
 
+    def validate_policy(self, value):
+        """
+        Check that the policy option is 'immediate'.
+        """
+        if value != 'immediate':
+            raise serializers.ValidationError("This plugin only supports policy='immediate'.")
```
daviddavis: i18n

pulpcore/app/serializers/repository.py
```diff
@@ -117,6 +117,14 @@ class RemoteSerializer(MasterModelSerializer):
         default=models.Remote.IMMEDIATE
     )
 
+    def validate_policy(self, value):
+        """
+        Check that the policy option is 'immediate'.
+        """
+        if value != 'immediate':
```
daviddavis: Why not use the IMMEDIATE constant here?

daviddavis:
I think you need to update this as well:

https://github.com/pulp/pulpcore/blob/ae6be0d89c5df26ac1cfdc876f3b131aa6e9bcf8/pulpcore/app/serializers/repository.py#L114-L117

And also, I think if you change the CHOICES, you can remove the validate method.

pulpcore/app/serializers/repository.py
bmbouter: fixed in next version. ty.

pulpcore/app/serializers/repository.py
bmbouter: I should use the constant here. fixed in next version.

bmbouter:
I did it that way first, but then I switched it to this for 2 reasons. I like that the error message indicates the plugin writer hasn't added support versus a validation error which says that option isn't validating. Also as this inherits to plugin Remote serializers plugin writers won't have to override this field's options too so it's simpler for plugin writers.

PR #201, opened 2019-07-05
Increasing artifact size field to hold more than 2GB

### LINKED ISSUE #4998: [RFE?] Keep worker process running in case of db connection error
From time to time my Postgres instance restarts (not related to Pulp) which, in turn, causes a database connection error and pulpcore worker process exits (and pod restarts).

Is there a way to handle these connection drops on pulpcore worker side? For example, the main process from pulpcore-api pods is a gunicorn app that forks children processes and whenever the connection to the database is lost, the api pods will throw exceptions but the main gunicorn process will keep running (only the forks will die), so this causes no pod restart. Is it possible to provide a similar behavior in pulpcore-worker pods so that the main process does not exit in case of database failure?  

**Additional context**
* We understand that the main problem is on the database side and the idea of this issue is just a "workaround".
* If we manage to find a way to keep the process running (and not restart the pod) will it handle/ensure the reconnection to the database?

daviddavis:
fixes #4998


CHANGES/4998.bug
```diff
@@ -0,0 +1 @@
+Increased artifact size field to prevent 500 errors for artifacts > 2GB in size.
```
bmbouter: I think this needs to be 4998.bugfix  https://github.com/hawkowl/towncrier#news-fragments

daviddavis:
@bmbouter fixed the changelog entry :+1:

PR #207, opened 2019-07-12
Add test to create artifact using chunk upload

kersommoura:
Add test to create artifact using chunk upload.

https://pulp.plan.io/issues/4197
ref: #4197

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/test_upload.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_upload.py
@@ -1,0 +1,84 @@
+# coding=utf-8
+"""Tests related to content upload."""
+import hashlib
+import unittest
+from random import shuffle
+from urllib.parse import urljoin
+
+from pulp_smash import api, config
+from pulp_smash.utils import http_get
+from pulp_smash.pulp3.constants import ARTIFACTS_PATH, UPLOAD_PATH
+
+from pulpcore.tests.functional.api.using_plugin.constants import (
+    FILE_CHUNKED_PART_1_URL,
+    FILE_CHUNKED_PART_2_URL,
+    FILE_TO_BE_CHUNKED_URL,
+)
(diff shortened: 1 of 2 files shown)


pulpcore/tests/functional/api/using_plugin/test_upload.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        """Create class-wide variables."""
+        cls.cfg = config.get_config()
+        cls.client = api.Client(cls.cfg)
+
+    def test_create_artifact(self):
+        """Test creation of artifact using upload of files in chunks."""
+        file = http_get(FILE_TO_BE_CHUNKED_URL)
```
daviddavis: I would move this to `setUpClass` so if a second test gets created it doesn't also try to download the file again.

pulpcore/tests/functional/api/using_plugin/test_upload.py
```diff
@@ -0,0 +1,83 @@
...
+        size_file = len(file)
+
+        first_chunk = http_get(FILE_CHUNKED_PART_1_URL)
```
daviddavis: Splitting up file in the code would be more efficient. Downloading the file and its chunks downloads twice as much data 20MB vs 10MB.

kersommoura: That was my first approach, but the machinery to do this was getting a bit complicated. What about we move forward with this approach, and if it is indeed slow we update?

daviddavis: That's fine. At least move this to setUpClass though so we're not downloading the chunks for each upload test

kersommoura: That one is already done.

kersommoura:
@bherrin3, I will merge this one. If you have any comment, I will send a follow up PR.

PR #212, opened 2019-07-16
Add test for testing the delete method of uploads

lubosmj:
closes #5145
https://pulp.plan.io/issues/5145

### DIFF
--- a/pulpcore/tests/functional/api/test_upload.py
+++ b/pulpcore/tests/functional/api/test_upload.py
@@ -1,13 +1,15 @@
 # coding=utf-8
 """Tests related to content upload."""
+import os
 import hashlib
 import unittest
 from random import shuffle
 from urllib.parse import urljoin
 
-from pulp_smash import api, config
+from pulp_smash import api, cli, config
 from pulp_smash.utils import http_get
-from pulp_smash.pulp3.constants import ARTIFACTS_PATH, UPLOAD_PATH
+from pulp_smash.exceptions import CalledProcessError
+from pulp_smash.pulp3.constants import ARTIFACTS_PATH, UPLOAD_PATH, MEDIA_PATH
 
 from pulpcore.tests.functional.api.using_plugin.constants import (
     FILE_CHUNKED_PART_1_URL,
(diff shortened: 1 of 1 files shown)


pulpcore/tests/functional/api/test_crd_uploads.py
```diff
@@ -0,0 +1,90 @@
...
+    FILE_TO_BE_CHUNKED_URL,
+)
+
+
+class UploadsDeleteFileSystemTestCase(unittest.TestCase):
+    """Test a deletion of an uploaded content
+
+    This test targets the following issue:
+
+    * `Pulp #4186 <https://pulp.plan.io/issues/5092>`_
```
dkliban: link and issue are not the same

kersommoura: Good job. 

This test could be moved to the  `pulpcore/tests/functional/api/using_plugin/test_upload.py`, and reuse the code already there.

lubosmj: I think that the test is not related to the plugin. Is it still necessary to move it to a different destination?

pulpcore/tests/functional/api/test_upload.py
```diff
@@ -82,3 +84,74 @@ def test_create_artifact(self):
...
+class UploadsDeleteFileSystemTestCase(unittest.TestCase):
+    """Test a deletion of an uploaded content
+
+    This test targets the following issue:
+
+    * `Pulp #5092 <https://pulp.plan.io/issues/5092>`_
+    """
+
+    @classmethod
+    def setUpClass(cls):
```
dkliban: This setUpClass method looks teh same as that of the ChunkedUploadTestCase class. You should just add the test_delete_upload() method to that class.

pulpcore/tests/functional/api/test_upload.py
```diff
@@ -22,22 +24,56 @@
...
+        # fetch a name of the upload from the corresponding _href
+        upload_name = upload_request['_href'].replace('/pulp/api/v3/uploads/', '')[:-1]
+
+        self.addCleanup(self.client.delete, upload_request['_href'])
+        cmd = ('ls', os.path.join(MEDIA_PATH, 'upload', upload_name))
+        self.cli_client.run(cmd, sudo=True)
+
+        # delete
+        self.doCleanups()
+        with self.assertRaises(CalledProcessError):
```
dkliban: What do you think about adding an assertion that you can perform an 'ls' before you do the delete?

dkliban: I now see that you already do this on line 69.

PR #213, opened 2019-07-16
Fix breaking changes from DRF 3.10

CodeHeeler:
Django restframework 3.10.0 has introduced breaking changes
Unpinning DRF now that a drf-yasg blocker has been fixed

fixes #5125
https://pulp.plan.io/issues/5125

### DIFF
--- a/pulpcore/app/openapigenerator.py
+++ b/pulpcore/app/openapigenerator.py
@@ -255,10 +255,11 @@ class PulpAutoSchema(SwaggerAutoSchema):
         produces = self.get_produces()
 
         multipart = ['multipart/form-data', 'application/x-www-form-urlencoded']
-        if 'file' in [param['type'] for param in self.get_request_body_parameters(multipart)]:
-            # automatically set the media type to form data if there's a file
-            # needed due to https://github.com/axnsan12/drf-yasg/issues/386
-            consumes = multipart
+        if self.method != 'GET':
+            if 'file' in [param['type'] for param in self.get_request_body_parameters(multipart)]:
+                # automatically set the media type to form data if there's a file
+                # needed due to https://github.com/axnsan12/drf-yasg/issues/386
+                consumes = multipart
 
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -7,12 +7,12 @@
...
     'Django~=2.2',  # LTS version, switch only if we have a compelling reason to
     'django-filter',
-    'djangorestframework<3.10',
+    'djangorestframework',
     'djangorestframework-queryfields',
     'drf-nested-routers',
     'drf-yasg',
     'gunicorn',
-    'PyYAML',
+    'PyYAML>=5.1',
```
dkliban: why does it need to be at least 5.1?

CodeHeeler: DRF now pins there because of a human readability feature in OpenAPI schema generation:
https://github.com/encode/django-rest-framework/pull/6680

I'm not sure if this is strictly necessary, what are your thoughts?

pulpcore/app/openapigenerator.py
```diff
@@ -255,10 +255,13 @@ def get_operation(self, operation_keys):
         produces = self.get_produces()
 
         multipart = ['multipart/form-data', 'application/x-www-form-urlencoded']
-        if 'file' in [param['type'] for param in self.get_request_body_parameters(multipart)]:
-            # automatically set the media type to form data if there's a file
-            # needed due to https://github.com/axnsan12/drf-yasg/issues/386
-            consumes = multipart
+        import pydevd
+        pydevd.settrace('localhost', port=3013, stdoutToServer=True, stderrToServer=True)
```
dkliban: This breakpoint is probably causing most of the test failures on travis.

setup.py
dkliban: I don't think it's necessary to put this minimum requirement here. Let's remove it.

CodeHeeler: Fine by me. Removed.

PR #216, opened 2019-07-17
4982 chunk checksums

fabricio-aguiar:
Enable sending checksum for chunk upload.

https://pulp.plan.io/issues/4982
closes #4982

### DIFF
--- a/docs/workflows/upload-publish.rst
+++ b/docs/workflows/upload-publish.rst
@@ -16,6 +16,10 @@ be used for chunks. Chunks can be uploaded in any order or in parallel::
     http --form PUT :24817/pulp/api/v3/uploads/a8b5a7f7-2f22-460d-ab20-d5616cb71cdd/ file@./chunk2 'Content-Range:bytes 6291456-10485759/*'
     http --form PUT :24817/pulp/api/v3/uploads/a8b5a7f7-2f22-460d-ab20-d5616cb71cdd/ file@./chunk1 'Content-Range:bytes 0-6291455'
 
+Note: You can send an optional sha256 argument::
+
+    http --form PUT :24817/pulp/api/v3/uploads/a8b5a7f7-2f22-460d-ab20-d5616cb71cdd/ file@./chunk1 'Content-Range:bytes 0-6291455' sha256=7ffc86295de63e96006ce5ab379050628aa5d51f816267946c71906594e13870
+
 Once all chunks have been uploaded, a final POST request with the file md5 can be sent to complete the
 upload::
 
--- a/pulpcore/app/models/upload.py
+++ b/pulpcore/app/models/upload.py
@@ -3,6 +3,7 @@ import os
 
 from django.core.files.base import ContentFile
 from django.db import models
+from rest_framework import serializers
 
 from pulpcore.app.models import Model
 
@@ -22,7 +23,7 @@ class Upload(Model):
     size = models.BigIntegerField()
     completed = models.DateTimeField(null=True)
 
-    def append(self, chunk, offset):
+    def append(self, chunk, offset, sha256=None):
         """
         Append a chunk to an upload.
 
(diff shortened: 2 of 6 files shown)


dkliban:
@fabricio-aguiar You need to amend the commit message[0]. Docs for this can be found here[1].


[0] https://travis-ci.com/pulp/pulpcore/jobs/217055495#L355
[1] https://docs.pulpproject.org/en/3.0/nightly/contributing/git.html#commit-message

pulpcore/app/models/upload.py
```diff
@@ -22,7 +23,7 @@ class Upload(Model):
     size = models.BigIntegerField()
     completed = models.DateTimeField(null=True)
 
-    def append(self, chunk, offset):
+    def append(self, chunk, offset, sha256):
```
daviddavis: This `sha256` param should probably be optional (ie `sha256=None`).

pulpcore/tests/functional/api/test_upload.py
```diff
@@ -56,6 +78,40 @@ def test_create_artifact(self):
...
+        response = self.client.post(
+            ARTIFACTS_PATH, {'upload': upload_request['_href']}
+        )
+
+        artifact = self.client.get(response['_href'])
+        self.addCleanup(self.client.delete, artifact['_href'])
+
+        self.assertEqual(artifact['sha256'], self.file_sha256, artifact)
+
+    def test_create_artifact_wrong_checksum(self):
```
daviddavis: Since you can create an artifact with a wrong checksum, we should probably name this test `test_upload_chunk_wrong_checksum` to be more specific.

pulpcore/tests/functional/api/test_upload.py
```diff
@@ -56,6 +78,40 @@ def test_create_artifact(self):
...
+        """Test creation of artifact using upload of files in chunks passing wrong checksum."""
+        data = {'sha256': "WRONG CHECKSUM"}
+        with self.assertRaises(HTTPError):
+            self.upload_chunks(data)
```
daviddavis: Can you check the status code too? I think it should be 400.

pulpcore/tests/functional/api/test_upload.py
```diff
@@ -42,7 +44,27 @@ def setUpClass(cls):
         cls.file_sha256 = hashlib.sha256(cls.file).hexdigest()
         cls.size_file = len(cls.file)
 
-    def test_create_artifact(self):
+        first_chunk = http_get(FILE_CHUNKED_PART_1_URL)
```
daviddavis: FYI, once I merge https://github.com/pulp/pulpcore/pull/215, this will be `self.first_chunk`.

daviddavis:
I don't see the sha256 param in the api docs when I go to http://localhost:24817/pulp/api/v3/docs/api#operation/uploads_update. I think it needs to be added to the serializer.

Also, can you update docs/workflows/upload-publish.rst and add a note that you can specify an optional sha256 checksum when uploading a chunk?

pulpcore/tests/functional/api/test_upload.py
kersommoura: In general we rely on the response handler to verify status code. 
https://github.com/PulpQE/pulp-smash/blob/e53e5a6a909a3a108a6a70e7f0bbfaefec7bb7b2/pulp_smash/api.py#L167

docs/workflows/upload-publish.rst
```diff
@@ -16,6 +16,9 @@ be used for chunks. Chunks can be uploaded in any order or in parallel::
     http --form PUT :24817/pulp/api/v3/uploads/a8b5a7f7-2f22-460d-ab20-d5616cb71cdd/ file@./chunk2 'Content-Range:bytes 6291456-10485759/*'
     http --form PUT :24817/pulp/api/v3/uploads/a8b5a7f7-2f22-460d-ab20-d5616cb71cdd/ file@./chunk1 'Content-Range:bytes 0-6291455'
 
+Note: You can send an optional sha256 argument:
+    http --form PUT :24817/pulp/api/v3/uploads/a8b5a7f7-2f22-460d-ab20-d5616cb71cdd/ file@./chunk1 'Content-Range:bytes 0-6291455' sha256=7ffc86295de63e96006ce5ab379050628aa5d51f816267946c71906594e13870
```
daviddavis: You need an empty line before this one otherwise it won't render properly. See: https://github.com/pulp/pulpcore/blob/81af19d3f37774210c786a7048f02f07e32937ca/docs/workflows/upload-publish.rst

PR #220, opened 2019-07-19
Add irc-build-notifications to travis

ggainey:


daviddavis:
These settings are going to get wiped out the next time we sync with the plugin_template config unless you also add them to template_config.yml.

ggainey:
I somehow missed that pulpcore is also under the control of pulp_template - dropping this PR, will submit a new one with the change in the right place!

PR #222, opened 2019-07-19
display chunks on upload payload

fabricio-aguiar:
closes #5150

### DIFF
--- a/pulpcore/app/serializers/upload.py
+++ b/pulpcore/app/serializers/upload.py
@@ -5,6 +5,24 @@ from pulpcore.app import models
 from pulpcore.app.serializers import base
 
 
+class UploadChunkSerializer(serializers.Serializer):
+    file = serializers.FileField(
+        help_text=_("A chunk of the uploaded file."),
+    )
+
+    sha256 = serializers.CharField(
+        help_text=_("The SHA-256 checksum of the chunk if available."),
+        required=False,
+        allow_null=True,
+    )
+
+
+class UploadChunkDetailSerializer(base.ModelSerializer):
+    class Meta:
+        model = models.UploadChunk
+        fields = ('offset', 'size')
+
+
 class UploadSerializer(base.ModelSerializer):
     """Serializer for chunked uploads."""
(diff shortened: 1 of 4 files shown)


pulpcore/app/serializers/upload.py
```diff
@@ -37,6 +37,23 @@ class UploadChunkSerializer(serializers.Serializer):
     )
 
 
+class UploadChunkDetailSerializer(base.ModelSerializer):
+    class Meta:
+        model = models.UploadChunk
+        fields = ('offset', 'size')
+
+
+class UploadDetailSerializer(UploadSerializer):
```
daviddavis: nitpick - I would probably move this under UploadSerializer.

pulpcore/tests/functional/api/test_upload.py
```diff
@@ -123,6 +124,40 @@ def test_upload_chunk_wrong_checksum(self):
 
             assert response.status_code == 400
 
+    def test_upload_response(self):
+        """Test response when creating artifact using upload of files in chunks."""
```
daviddavis: This should probably be something like "Test upload responses when creating an upload and uploading chunks" since you aren't creating an artifact in this test.

pulpcore/tests/functional/api/test_upload.py
```diff
@@ -123,6 +124,41 @@ def test_upload_chunk_wrong_checksum(self):
 
             assert response.status_code == 400
 
+    def test_upload_response(self):
+        """Test upload responses when creating an upload and uploading chunks."""
+        self.client.response_handler = api.echo_handler
```
kersommoura: We can create a client, and just modified the handler when necessary the method [using_handler](https://pulp-smash.readthedocs.io/en/latest/api/pulp_smash.api.html?highlight=using_handler#pulp_smash.api.Client.using_handler). Then we do not need to mutate the client. Besides that, the safe_handler, at least verify the code status. A bit more robust. Suggest to use echo_handler just when necessary.

PR #224, opened 2019-07-22
using correct handler

fabricio-aguiar:
ref #5150

### DIFF
--- a/pulpcore/tests/functional/api/test_upload.py
+++ b/pulpcore/tests/functional/api/test_upload.py
@@ -108,15 +108,13 @@ class ChunkedUploadTestCase(unittest.TestCase):
 
     def test_upload_chunk_wrong_checksum(self):
         """Test creation of artifact using upload of files in chunks passing wrong checksum."""
-        self.client.response_handler = api.echo_handler
-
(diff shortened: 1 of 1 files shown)


pulpcore/tests/functional/api/test_upload.py
```diff
@@ -126,37 +124,35 @@ def test_upload_chunk_wrong_checksum(self):
...
-        self.client.response_handler = api.echo_handler
-
         upload_request = self.client.post(
             UPLOAD_PATH, {'size': self.size_file}
         )
 
         expected_keys = ['_href', '_created', 'size', 'completed']
 
-        self.assertEquals([*upload_request.json()], expected_keys)
+        self.assertEquals([*upload_request], expected_keys)
```
kersommoura:
```suggestion
        self.assertEqual([*upload_request], expected_keys, upload_request)
```


`assertEquals` is deprecated. See: https://docs.python.org/2/library/unittest.html#deprecated-aliases

Adding a message to the assertion in case of failure can be helpful in case of failure. 

See: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertEqual

PR #227, opened 2019-07-22
Upload creating artifact asynchronously 

fabricio-aguiar:
https://pulp.plan.io/issues/5087

### DIFF
--- a/.travis/test_bindings.py
+++ b/.travis/test_bindings.py
@@ -49,7 +49,7 @@ def upload_file_in_chunks(file_path):
         file_path (str): path to the file being uploaded to Pulp
 
     Returns:
-        upload object
+        artifact object
     """
     size = os.stat(file_path).st_size
     chunk_size = 200000
@@ -74,8 +74,12 @@ def upload_file_in_chunks(file_path):
                                         content_range=content_range)
             offset += chunk_size
             sha256hasher.update(chunk)
-        uploads.commit(upload.href, UploadCommit(sha256=sha256hasher.hexdigest()))
-    return upload
+
+        commit_response = uploads.commit(upload.href, UploadCommit(sha256=sha256hasher.hexdigest()))
+        created_resources = monitor_task(commit_response.task)
+        artifact = artifacts.read(created_resources[0])
+
+    return artifact
 
 
 # Configure HTTP basic authorization: basic
@@ -106,9 +110,7 @@ with NamedTemporaryFile() as downloaded_file:
         'pulp-large_1mb_test-packageA-0.1.1-1.fc14.noarch.rpm')
     response.raise_for_status()
     downloaded_file.write(response.content)
-    upload = upload_file_in_chunks(downloaded_file.name)
-    pprint(upload)
-    artifact = artifacts.create(upload=upload.href)
+    artifact = upload_file_in_chunks(downloaded_file.name)
     pprint(artifact)
 
 # Create a File Remote
--- a/pulpcore/app/tasks/upload.py
+++ b/pulpcore/app/tasks/upload.py
@@ -1,0 +1,46 @@
+from datetime import datetime
+from gettext import gettext as _
+from logging import getLogger
+
+from rest_framework import serializers
+
+from pulpcore.app import models
+from pulpcore.app.models import CreatedResource
+from pulpcore.app.serializers import ArtifactUploadSerializer
+from pulpcore.tasking.util import get_url
+
+log = getLogger(__name__)
+
+
+def commit(upload_id, sha256):
+    """
+    Commit the upload and mark it as completed.
+
+    Commit a :class:`~pulpcore.app.models.Upload`
+
+    Args:
+        upload_id (int): The upload primary key
+        sha256 (str): The checksum for the uploaded file
+
+    """
+    try:
+        upload = models.Upload.objects.get(pk=upload_id)
+    except models.Upload.DoesNotExist:
+        log.info(_('The upload was not found. Nothing to do.'))
+        return
+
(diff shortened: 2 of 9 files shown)


.travis/test_bindings.py
```diff
@@ -74,8 +74,12 @@ def upload_file_in_chunks(file_path):
                                         content_range=content_range)
             offset += chunk_size
             sha256hasher.update(chunk)
-        uploads.commit(upload.href, UploadCommit(sha256=sha256hasher.hexdigest()))
-    return upload
+
+        sync_response = uploads.commit(upload.href, UploadCommit(sha256=sha256hasher.hexdigest()))
```
daviddavis: `sync_response` seems odd. Maybe `commit_response`?

.travis/test_bindings.py
```diff
@@ -74,8 +74,12 @@ def upload_file_in_chunks(file_path):
...
+        created_resources = monitor_task(sync_response.task)
+        artifact = artifacts.read(created_resources[0])
+
+    return upload, artifact
```
daviddavis: I would just return artifact

.travis/test_bindings.py
```diff
@@ -106,9 +110,8 @@ def upload_file_in_chunks(file_path):
         'pulp-large_1mb_test-packageA-0.1.1-1.fc14.noarch.rpm')
     response.raise_for_status()
     downloaded_file.write(response.content)
-    upload = upload_file_in_chunks(downloaded_file.name)
+    upload, artifact = upload_file_in_chunks(downloaded_file.name)
     pprint(upload)
```
daviddavis: And remove this line.

pulpcore/app/tasks/upload.py
```diff
@@ -0,0 +1,45 @@
...
+
+from pulpcore.app import models
+from pulpcore.app.models import CreatedResource
+from pulpcore.app.serializers import ArtifactSerializer
+from pulpcore.tasking.util import get_url
+
+log = getLogger(__name__)
+
+
+def commit(upload_id, data):
```
daviddavis: Recommend that you accept `sha256` instead of `data`.

pulpcore/app/tasks/upload.py
```diff
@@ -0,0 +1,45 @@
...
+    """Commit the upload and mark it as completed."""
```
daviddavis: Can you document the arguments? See here as an example:

https://github.com/pulp/pulpcore/blob/master/pulpcore/app/tasks/repository.py#L11-L17

pulpcore/app/tasks/upload.py
```diff
@@ -0,0 +1,45 @@
...
+
+    try:
+        sha256 = data['sha256']
+    except KeyError:
+        raise serializers.ValidationError(_("Checksum not supplied."))
+
+    if sha256 != upload.sha256:
+        raise serializers.ValidationError(_("Checksum does not match upload."))
+
+    if upload.completed is not None:
```
daviddavis: We should drop this field. It's no longer necessary since we delete the upload once the artifact has been created.

daviddavis:
Oh, can you make sure `upload` is no longer a parameter for the artifact create endpoint:

https://docs.pulpproject.org/en/3.0/nightly/restapi.html#operation/artifacts_create

fabricio-aguiar:
thanks @dkliban and @daviddavis

.travis/test_bindings.py
```diff
@@ -74,8 +74,12 @@ def upload_file_in_chunks(file_path):
...
             offset += chunk_size
             sha256hasher.update(chunk)
-        uploads.commit(upload.href, UploadCommit(sha256=sha256hasher.hexdigest()))
-    return upload
+
+        commit_response = uploads.commit(upload.href, UploadCommit(sha256=sha256hasher.hexdigest()))
+        created_resources = monitor_task(commit_response.task)
+        artifact = artifacts.read(created_resources[0])
+
+    return artifact
```
dkliban: Please update the doc string with the new return type.

pulpcore/app/tasks/upload.py
```diff
@@ -0,0 +1,49 @@
...
+    except models.Upload.DoesNotExist:
+        log.info(_('The upload was not found. Nothing to do.'))
+        return
+
+    log.info(_('Commiting the upload %(i)d and mark it as completed'), {'i': upload_id})
+
+    if not sha256:
+        raise serializers.ValidationError(_("Checksum not supplied."))
+
+    if sha256 != upload.sha256:
```
dkliban: The file is read at this point and it's sha256 checksum is calculated here.

pulpcore/app/tasks/upload.py
```diff
@@ -0,0 +1,49 @@
...
+        raise serializers.ValidationError(_("Checksum does not match upload."))
+
+    upload.completed = datetime.now()
+    upload.save()
+
+    data = {'upload': get_url(upload)}
+    serializer = ArtifactUploadSerializer(data=data)
```
dkliban: The file is read again and all of it's checksums needed to store an Artifact are calculated here. 

We should avoid reading the file twice. I think it will be better to just use the ArtifactSerializer to check the checksum. You can add the 'sha256' field to the data on line 43 and teh ArtifactSerialize will assert that that value is equal to the real one. 

Then you can also remove lines 37 and 38.

daviddavis: Also, we should probably remove the sha256 method from Upload since it's no longer needed. That would prevent us from needlessly calculating upload sha256 somewhere in the code.

daviddavis:
Thanks @fabricio-aguiar!

PR #230, opened 2019-07-24
Problem: ruby client is not tested

dkliban:
Solution: run the .travis/test_bindings.rb with each PR

re: #4896
https://pulp.plan.io/issues/4896

### DIFF
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -65,6 +65,29 @@ if [ "$TEST" = 'bindings' ]; then
   ./generate.sh pulp_file python
   pip install ./pulp_file-client
   python $TRAVIS_BUILD_DIR/.travis/test_bindings.py
+
+  if [ ! -f $TRAVIS_BUILD_DIR/.travis/test_bindings.rb ]
+  then
+    exit
+  fi
+
+  rm -rf ./pulpcore-client
+
+  ./generate.sh pulpcore ruby
+  cd pulpcore-client
+  gem build pulpcore_client
+  gem install --both ./pulpcore_client-0.gem
+  cd ..
+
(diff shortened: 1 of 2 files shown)


.travis/script.sh
```diff
@@ -65,6 +65,23 @@ if [ "$TEST" = 'bindings' ]; then
   ./generate.sh pulp_file python
   pip install ./pulp_file-client
   python $TRAVIS_BUILD_DIR/.travis/test_bindings.py
+  if [ ! -f $TRAVIS_BUILD_DIR/.travis/test_bindings.rb ]
```
daviddavis: What's the purpose of this check?

.travis/script.sh
```diff
@@ -65,6 +65,23 @@ if [ "$TEST" = 'bindings' ]; then
...
+  gem build pulpcore_client
+  gem install --both ./pulpcore_client-0.gem
+  cd ..
+  rm -rf ./pulp_file-client
+  ./generate.sh pulp_file ruby
+  cd pulp_file-client
+  gem build pulp_file_client
+  gem install --both ./pulp_file_client-0.gem
+  cd ..
+  ruby $TRAVIS_BUILD_DIR/.travis/test_bindings.rb
```
daviddavis: This is kind of hard to read. Can you organize this by adding in some line breaks?

.travis/script.sh
dkliban: It only tests ruby bindings if the file exists. I will open a PR to plugin_template with this change.

.travis/script.sh
dkliban: Sure. I'll do it.

PR #238, opened 2019-07-30
test for 5167

fabricio-aguiar:
ref #5167

related to https://github.com/pulp/pulp_ansible/pull/146

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/constants.py
+++ b/pulpcore/tests/functional/api/using_plugin/constants.py
@@ -105,3 +105,17 @@ standard repositories, i.e. :data:`RPM_SIGNED_FIXTURE_URL` and
 :data:`RPM_UNSIGNED_FIXTURE_URL`.  This matches the format output by the
 "content_summary" field on "../repositories/../versions/../".
 """
+
+ANSIBLE_GALAXY_URL = "https://galaxy.ansible.com/api/v1/roles/"
+
+NAMESPACE_ANSIBLE = "?namespace__name=ansible"
+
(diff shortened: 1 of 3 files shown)


bmbouter:
@fabricio-aguiar It's so great that you are writing smash tests along with your work. The testing strategy has changed some over time and I think the idea is that all tests should go with the plugin they test. Then we'll run some set of those tests on pulpcore to ensure we aren't breaking those plugins in some way. How about moving this test to be [here](https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/tests/functional/api/collection/test_sync.py) instead and the CI of pulp_ansible will start to run it?

@kersommoura is what I'm saying here right?

kersommoura:
@bmbouter, yeah. Idea is to keep in pulpcore, just the features related to pulpcore itself. Some features of pulpcore requires a plugin in order to be exercised, and in this case we are adding tests in the `using_plugin` dir. Most of those tests that require an extra plugin we have been using `file_plugin`. 
@fabricio-aguiar, this is a question that we have been asking for a while. If this test is exercising a pulp-ansible feature, my suggestion is to move this test over the other repo.

kersommoura:
@fabricio-aguiar +1 for your approach of writing tests together with feature implementation. Excellent initiative.

fabricio-aguiar:
thanks @bmbouter and @kersommoura 
I am closing this PR to move it to https://github.com/pulp/pulp_ansible/pull/146

bmbouter:
@fabricio-aguiar thank you!

PR #250, opened 2019-08-02
Use local fixtures instead of the ones from fedorapeople.org

daviddavis:
[noissue]

### DIFF
--- a/.travis/create_fixtures.sh
+++ b/.travis/create_fixtures.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+git clone https://github.com/PulpQE/pulp-fixtures.git
+cd pulp-fixtures
+make all
+cp -R fixtures /var/www/html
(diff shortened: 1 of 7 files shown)


.travis/create_fixtures.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+git clone https://github.com/PulpQE/pulp-fixtures.git
+cd pulp-fixtures
+make all
```
kersommoura: Perhaps we could just generate the fixtures used by pulpcore.

All the fixtures requires a few [dependencies]( https://github.com/pulp/pulp-ci/blob/8ea394b588d0d3e1ee19aaac7fa11d177e3b8fb4/ci/jjb/jobs/pulp-fixtures-publisher.yaml#L22).

daviddavis:
This approach is too complex IMO.

PR #254, opened 2019-08-06
Allow artfacts to be pre-stored in the Pulp 3 storage

goosemania:
re #3810
https://pulp.plan.io/issues/3810

### DIFF
--- a/pulpcore/app/models/fields.py
+++ b/pulpcore/app/models/fields.py
@@ -43,15 +43,20 @@ class ArtifactFileField(FileField):
         file = model_instance.file
         artifact_storage_path = self.upload_to(model_instance, '')
 
-        if file.name != artifact_storage_path and file.name.startswith(
-                os.path.join(settings.MEDIA_ROOT, 'artifact')):
+        already_in_place = (
+            file.name in [artifact_storage_path,
+                          os.path.join(settings.MEDIA_ROOT, artifact_storage_path)])
+        is_in_artifact_storage = file.name.startswith(os.path.join(settings.MEDIA_ROOT, 'artifact'))
+
+        if not already_in_place and is_in_artifact_storage:
             raise ValueError(_('The file referenced by the Artifact is already present in '
                                'Artifact storage. Files must be stored outside this location '
                                'prior to Artifact creation.'))
 
         move = (file._committed and file.name != artifact_storage_path)
         if move:
-            file._file = TemporaryDownloadedFile(open(file.name, 'rb'))
+            if not already_in_place:
+                file._file = TemporaryDownloadedFile(open(file.name, 'rb'))
             file._committed = False
 
         return super().pre_save(model_instance, add)


pulpcore/app/models/fields.py
```diff
@@ -43,15 +43,20 @@ def pre_save(self, model_instance, add):
...
+
+        if not has_the_existing_path and is_in_artifact_storage:
             raise ValueError(_('The file referenced by the Artifact is already present in '
                                'Artifact storage. Files must be stored outside this location '
                                'prior to Artifact creation.'))
 
         move = (file._committed and file.name != artifact_storage_path)
         if move:
-            file._file = TemporaryDownloadedFile(open(file.name, 'rb'))
+            if not has_the_existing_path:
```
goosemania: If the file is in /var/lib/pulp/artifact, don't move it, just commit. What do you think?

daviddavis: Are we allowing plugin writers to write artifacts to other locations in `/var/lib/pulp/artifact` that don't conform to [our current logic in get_artifact_path](https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/storage.py#L116-L127)? If so, then how do we prevent collisions?

pulpcore/app/models/fields.py
```diff
@@ -43,15 +43,20 @@ def pre_save(self, model_instance, add):
         file = model_instance.file
         artifact_storage_path = self.upload_to(model_instance, '')
 
-        if file.name != artifact_storage_path and file.name.startswith(
-                os.path.join(settings.MEDIA_ROOT, 'artifact')):
+        has_the_existing_path = (
+            file.name in [artifact_storage_path,
+                          os.path.join(settings.MEDIA_ROOT, artifact_storage_path)])
```
daviddavis: Not sure I understand this change. Why are we checking `os.path.join(settings.MEDIA_ROOT, artifact_storage_path)` in addition to `artifact_storage_path`?

pulpcore/app/models/fields.py
daviddavis: If we are assuming that everything `/var/lib/pulp/artifact` must conform to `get_artifact_path` then I guess my concern would be if we don't move the artifact then it's location actually does conform to the logic in `get_artifact_path`

pulpcore/app/models/fields.py
goosemania: `file.name` can be absolute and can be relative. We used to check only relative, and now I'm checking both.
I think it depends if you provided a file object or a string as a `file` for Artifact object instantiation.

pulpcore/app/models/fields.py
goosemania: Maybe my comment was not precise enough. I'll try again:
If the file is in `MEDIA_ROOT` + `path which would be produce by get_artifact_path`, don't move it, just commit.

> Are we allowing plugin writers to write artifacts to other locations in /var/lib/pulp/artifact that don't conform to our current logic in get_artifact_path?

No, we are not allowing other locations than  `MEDIA_ROOT` + `get_artifact_path()`.

> I could see plugin writers accidentally putting files into the wrong location in /var/lib/pulp/artifact/.

Do you mean the case when the mistake is in sha256 digest which plugin writers used?  I think it's checked in the `has_the_existing_path`.

I should probably rename it to `has_the_same_path`. Or any other name suggestions are welcome.

daviddavis: I guess we don't want to move it in the case where we have a hard link? That makes sense to me. What about calling the variable `already_in_place` or something?

Also, this might be a great time to use unit tests to make sure we're hitting/supporting all these possible cases. This code is complex and thus fragile.

goosemania: Name is updated

PR #235, opened 2019-07-25
Implement filter for duplicated reserved_resources

lubosmj:
closes #5120
https://pulp.plan.io/issues/5120

PR #257, opened 2019-08-06
Add test for filtering created_resources fields

lubosmj:
closes #5180
https://pulp.plan.io/issues/5180

### DIFF
--- a/pulpcore/tests/functional/api/test_tasks.py
+++ b/pulpcore/tests/functional/api/test_tasks.py
@@ -3,8 +3,13 @@
 import unittest
 
 from pulp_smash import api, config, utils
-from pulp_smash.pulp3.constants import P3_TASK_END_STATES, REPO_PATH, TASKS_PATH
-from pulp_smash.pulp3.utils import gen_repo
+from pulp_smash.pulp3.constants import (
+    BASE_DISTRIBUTION_PATH,
+    P3_TASK_END_STATES,
+    REPO_PATH,
+    TASKS_PATH
+)
+from pulp_smash.pulp3.utils import gen_repo, gen_distribution
 from requests import HTTPError
 
 from pulpcore.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
@@ -145,3 +150,49 @@ class TasksTestCase(unittest.TestCase):
     def filter_tasks(self, criteria):
         """Filter tasks based on the provided criteria."""
         return self.client.get(TASKS_PATH, params=criteria)
+
+
+class FilterTaskCreatedResourcesTestCase(unittest.TestCase):
+    """Perform filtering over task resources.
+
+    This test targets the following issue:
+
+    * `Pulp #5180 <https://pulp.plan.io/issues/5180>`_
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """Create class-wide variables."""
+        cls.client = api.Client(config.get_config(), api.json_handler)
+        cls.task = {}
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the created task and related resources."""
+        cls.client.delete(cls.task['_href'])
+
+        created_distribution = cls.task['created_resources'][0]
+        response = cls.client.delete(created_distribution)
+
+        cls.client.delete(response['task'])
+
+    def test_01_create_task_with_reserved_resources(self):
+        """Create a task that will create a new distribution."""
+        distribution_path = '{}file/file/'.format(BASE_DISTRIBUTION_PATH)
+        response = self.client.post(distribution_path, gen_distribution())
+
+        task = self.client.get(response['task'])
+        self.task.update(task)
+
+    @skip_if(bool, 'task', False)
+    def test_02_read_fields_created_resources_only(self):
+        """Read created resources from the requested fields."""
+        task = self.client.get(
+            self.task['_href'],
+            params={'fields': 'created_resources'}
+        )
+        self.assertEqual(
+            self.task['created_resources'],
+            task['created_resources'],
+            task
+        )


kersommoura:
@lubosmj, thank you for writing test this.

I inspected this code locally, and I would like to suggest those new tests to be added into a new class.

We will avoid repetition of test names, and the use of decorators per required variable will be easier to follow.

```python
    @skip_if(bool, 'created_resources_task', False)
```

New class...
```python
TaskCreatedResourcesTestCase(unittet.TestCase):

```

pulpcore/tests/functional/api/test_tasks.py
```diff
@@ -3,7 +3,12 @@
 import unittest
 
 from pulp_smash import api, config, utils
-from pulp_smash.pulp3.constants import P3_TASK_END_STATES, REPO_PATH, TASKS_PATH
+from pulp_smash.pulp3.constants import (
+    P3_TASK_END_STATES,
+    REPO_PATH,
+    TASKS_PATH,
+    BASE_DISTRIBUTION_PATH
+)
```
kersommoura: niptick.  We sort this alphabetically.

pulpcore/tests/functional/api/test_tasks.py
```diff
@@ -132,15 +148,37 @@ def test_02_search_by_invalid_name(self):
...
         self.assertEqual(len(search_results['results']), 0)
 
+    @skip_if(bool, 'created_resources_task', False)
+    def test_02_read_fields_created_resources_only(self):
+        """Read created resources from the requested fields."""
+        task = self.client.get(
+            self.created_resources_task['_href'],
+            params={'fields': 'created_resources'}
+        )
+        self.assertEqual(
```
kersommoura: Adding a third parameter as msg to the assertions can be helpful in case of failures.

See: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual

pulpcore/tests/functional/api/test_tasks.py
```diff
@@ -132,15 +148,37 @@ def test_02_search_by_invalid_name(self):
...
+            self.created_resources_task['created_resources'],
+            task['created_resources']
```
kersommoura:
```suggestion
            task['created_resources'],
            task
```

lubosmj:
Thank you for the review. I really appreciate that.

I have one question. The decorator skip_if() has to be used within the new class FilterTaskCreatedResourcesTestCase, or not?

kersommoura:
``@skif_if`` decorator is used if we need to assure that certain variable was created previously. It is a way to assure that the test will not fail in case a certain variable to exist and does not. If we move those tests to a new class, and we need to assure a certain variable in order to run a certain test, it is a good use case for this decorator.

pulpcore/tests/functional/api/test_tasks.py
```diff
@@ -38,6 +45,15 @@ def test_01_create_task(self):
         response = self.client.patch(repo['_href'], attrs)
         self.task.update(self.client.get(response['task']))
 
+    def test_01_create_task_with_reserved_resources(self):
+        """Create a task that will create a new distribution."""
+        attrs = {'name': 'name', 'base_path': 'base_path'}
```
kersommoura: There is a utils function `gen_distribution` perhaps we could just call this one to create this test data.

pulpcore/tests/functional/api/test_tasks.py
```diff
@@ -134,14 +140,71 @@ def test_02_search_by_invalid_name(self):
...
+        response = self.client.post(distribution_path, gen_distribution())
+
+        task = self.client.get(response['task'])
+        self.task.update(task)
+
+    @skip_if(bool, 'task', False)
+    def test_02_read_fields_created_resources_only(self):
+        """Read created resources from the requested fields."""
+        task = self.client.get(
+            self.task['_href'],
```
kersommoura: Is it the goal to filter in the general task path or a specific task?
```suggestion
     TASKS_PATH,
```

pulpcore/tests/functional/api/test_tasks.py
```diff
@@ -134,14 +140,71 @@ def test_02_search_by_invalid_name(self):
...
+
+    @skip_if(bool, 'task', False)
+    def test_03_delete_distribution(self):
+        """Delete the created distribution."""
+        created_distribution = self.task['created_resources'][0]
+        self.client.delete(created_distribution)
+        with self.assertRaises(HTTPError):
+            self.client.get(created_distribution)
+
+    def test_04_delete_tasks(self):
```
kersommoura: Since the goal of this test to verify the filter of created resources. This test can be removed. Deletion of tasks is already covered by other tests.

lubosmj: Isn't this method something like tearDown()? At fist, we have allocated resources (created a new task) and after assertions, we have released the resources. Why is the deletion of this task covered in a separate tests?

bmbouter:
Is this test for the code in https://github.com/pulp/pulpcore/pull/235/files ? If so I think it would be cool if these PRs were combined so the tests is merged along w/ the code it tests. If they are unrelated then two PRs makes sense.

lubosmj:
No, this test code is not related to the aforementioned PR, however, it is still related to the filtering. This test targets https://github.com/pulp/pulpcore/pull/228.

The main difference between these PRs is the way of filtering. In this test, we are filtering tasks by fields (?fields=created_resources). For https://github.com/pulp/pulpcore/pull/228 we would be required to filter tasks by reserved_resources (?reserved_resources=/path/to/reserved/resource).

bmbouter:
@lubosmj cool that makes sense. Thank you.

kersommoura:
@bmbouter, we talked about those tests and they have different goals.

pulpcore/tests/functional/api/test_tasks.py
kersommoura: The previous test has as a goal perform actions - CRUD over the task system. That is why there is the delete test, I will follow up with others about clean up of indirect created resources. Thanks.

pulpcore/tests/functional/api/test_tasks.py
lubosmj: In order to trigger the method which was previously modified (https://github.com/pulp/pulpcore/pull/228), it is sufficient enough to create an HTTP GET request that targets a single task.

lubosmj: Because of that, I do not need to delete all tasks in a database in the method setUpClass(). I do can just query the right one.

PR #259, opened 2019-08-09
Avoid collisions between plugin models

### LINKED ISSUE #4681: file:// sync deletes files from directory
**Version**
Pulpcore 3.39

**Describe the bug**
When syncing file:// repositories, files are disappearing after the sync.

**To Reproduce**
1) Copy these two repositories to the FS:
  - https://github.com/Katello/katello/tree/master/test/fixtures/test_repos/file1
  - https://github.com/Katello/katello/tree/master/test/fixtures/test_repos/file2
2) Sync one, then the other
3) See that some files disappeared.
  - In my case, file2 lost every file except PULP_MANIFEST

**Expected behavior**
No files disappear.

**Additional context**
This also occurred with RPM content type files.

fabricio-aguiar:
Making mandatory declaring default_related_name
closes #4681

https://pulp.plan.io/issues/4681

# NOTE: DO NOT MERGE UNTIL AUGUST 19, 2019

### DIFF
--- a/pulpcore/app/models/base.py
+++ b/pulpcore/app/models/base.py
@@ -1,7 +1,9 @@
+from gettext import gettext as _
 import uuid
 
 from django.db import models
 from django.db.models import options
+from django.db.models.base import ModelBase
 
 
 class Model(models.Model):
@@ -35,7 +37,24 @@ class Model(models.Model):
         return str(self)
 
 
-class MasterModel(Model):
+class MasterModelMeta(ModelBase):
+    def __new__(cls, name, bases, attrs, **kwargs):
+        """Override __new__ to set the default_related_name."""
+        if Model not in bases and MasterModel not in bases:  # Only affects "Detail" models.
+            meta = attrs.get("Meta")
+            default_related_name = getattr(
+                meta, "default_related_name", None)
+            abstract = getattr(meta, "abstract", None)
+
+            if not default_related_name and not abstract:
+                raise Exception(_("The 'default_related_name' option has not been set for "
+                                  "{class_name}").format(class_name=name))
+
+        new_class = super().__new__(cls, name, bases, attrs, **kwargs)
+        return new_class
+
+
+class MasterModel(Model, metaclass=MasterModelMeta):
     """Base model for the "Master" model in a "Master-Detail" relationship.
 
     Provides methods for casting down to detail types, back up to the master type,
--- a/CHANGES/4681.misc
+++ b/CHANGES/4681.misc
@@ -1,0 +1,1 @@
+Avoid collisions between plugin models.


CHANGES/4681.bugfix
```diff
@@ -0,0 +1 @@
+Avoid collisions between plugin models.
```
asmacdo: This is good for user-facing changelog IMO, but we need something for plugin writers that indicates that this is a breaking change. One solution would be adding a ` .feature` CHANGE pointing to this issue in https://github.com/pulp/pulpcore-plugin/pull/119

pulpcore/app/models/base.py
```diff
@@ -35,7 +37,24 @@ def __repr__(self):
         return str(self)
 
 
-class MasterModel(Model):
+class MasterModelMeta(ModelBase):
+    def __new__(cls, name, bases, attrs, **kwargs):
+        """Override __new__ to set the default_related_name."""
+        if Model not in bases and MasterModel not in bases:
```
asmacdo: Could you add a comment explicitly stating that this only affects "detail" models?

fabricio-aguiar: Sure!

CHANGES/4681.bugfix
daviddavis: Adding a `4681.feature` to pulpcore-plugin makes sense to me. I think this would be a great way to document changes to the plugin API. 

cc @bmbouter

bmbouter: @daviddavis I agree, even though the changes occur here, we should represent this in the pulpcore-plugin release notes instead. I think of the changelog for pulpcore as the changelog for end users, and the changelog for pulpcore-plugin as the changelog for plugin writers.

@asmacdo, I don't think end-users will understand `4681.feature` for the reasons above. For that reason removing `4681.feature` and keeping `4681.bugfix` make the most sense to me.

pulpcore/app/models/base.py
```diff
@@ -35,7 +37,24 @@ def __repr__(self):
...
+    def __new__(cls, name, bases, attrs, **kwargs):
+        """Override __new__ to set the default_related_name."""
+        if Model not in bases and MasterModel not in bases:  # Only affects "Detail" models.
+            meta = attrs.get("Meta")
+            default_related_name = getattr(
+                meta, "default_related_name", None)
+            abstract = getattr(meta, "abstract", None)
+
+            if not default_related_name and not abstract:
+                raise Exception(_("The default_related_name option has not been set for "
```
bmbouter: Maybe put single quotes around `default_related_name` so it's clearer for the user?

CHANGES/4681.bugfix
daviddavis: Should we rename this file to 4681.misc then?

CHANGES/4681.feature
```diff
@@ -0,0 +1,2 @@
+Breaking change on how to subclass Master/Detail models in plugins.
+Please refer to https://github.com/pulp/pulpcore-plugin/pull/119
```
daviddavis: I think you can drop this file based on the above conversation.

PR #272, opened 2019-08-22
Add load of plugin default settings.

rochacbruno:
- Includes default settings from `{plugin_name}/app/settings.py`
- This settings overrides the data existing in `/etc/pulp/settings.py`
- This settings are overridden by env vars prefixed with `PULP_`
- So a good practice for plugin writers is to prefix the default configs with plugin name such as `RPM`, `FILE`, `ANSIBLE` etc...
 
example:

 `ANSIBLE_SOMEKEY = 'value'` on  `{plugin_name}/app/settings.py` can be overriden via 
`export PULP_ANSIBLE_SOMEKEY='othervalue'`

closes #5290
https://pulp.plan.io/issues/5290

### DIFF
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -223,7 +223,10 @@ settings = dynaconf.DjangoDynaconf(
     GLOBAL_ENV_FOR_DYNACONF='PULP',
     ENV_SWITCHER_FOR_DYNACONF='PULP_ENV',
     SETTINGS_MODULE_FOR_DYNACONF='/etc/pulp/settings.py',
-    INCLUDES_FOR_DYNACONF=['/etc/pulp/plugins/*'],
+    INCLUDES_FOR_DYNACONF=[
+        '{}.app.settings'.format(plugin_name)
+        for plugin_name in INSTALLED_PULP_PLUGINS
+    ],
     ENVVAR_FOR_DYNACONF='PULP_SETTINGS',
 )
 # HERE ENDS DYNACONF EXTENSION LOAD (No more code below this line)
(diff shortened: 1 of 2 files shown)


pulpcore/app/settings.py
```diff
@@ -223,7 +223,10 @@
     GLOBAL_ENV_FOR_DYNACONF='PULP',
     ENV_SWITCHER_FOR_DYNACONF='PULP_ENV',
     SETTINGS_MODULE_FOR_DYNACONF='/etc/pulp/settings.py',
-    INCLUDES_FOR_DYNACONF=['/etc/pulp/plugins/*'],
+    INCLUDES_FOR_DYNACONF=[
+        '{}.app.settings'.format(plugin_name)
```
rochacbruno: Maybe it needs to be
```suggestion
        'pulp_{}.app.settings'.format(plugin_name) 
```

rochacbruno: And that may cause problems with plugin naming, like `pulp-certguard` uses `-` while `pulp_rpm` uses `_` I am really not sure..

Another solution is to try to extract the proper path from `entry_point` object https://github.com/pulp/pulpcore/blob/master/pulpcore/app/settings.py#L63

bmbouter: I think what you have here is right. Based on my testing with pulp_file and pulp_ansible installed and this PR checked out, this is being set to `['pulp_file.app.settings', 'pulp_ansible.app.settings']`. Based on what you are saying is correct.

bmbouter:
This one is tricky with changelogs because even though the change is here, the changelog entry needs to be added to here instead with another PR:  https://github.com/pulp/pulpcore-plugin/tree/master/CHANGES  I can make this changelog entry with my docs PR.

Travis will still require a CHANGES entry for this PR though, so I recommend using `5290.misc` for this one.

PR #274, opened 2019-08-23
Update constants related to docker

kersommoura:
Update constants related to docker plugin.

'[noissue]'


kersommoura:
Tested against Pulp installation from source Aug/23.

PR #288, opened 2019-09-09
Removing code from task errors.

fabricio-aguiar:
closes #5282
https://pulp.plan.io/issues/5282

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 build
 dist
 *.egg-info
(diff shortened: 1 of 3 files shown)


CHANGES/5282.misc
```diff
@@ -0,0 +1 @@
+Removing code from task errors.
```
daviddavis: I think this should probably be a `.removal` file since we are removing a field from the API.

PR #290, opened 2019-09-10
Update the version number of dynaconf

lubosmj:
A feature for loading python modules was added in the version
2.0.4. The PR https://github.com/pulp/pulpcore/pull/272 utilizing
this feature was already merged without updating the dependencies.

Before this commit, settings defined in different plugins were not
loaded at all. A new version of dynaconf also adds a spupport for
environment variables for nested data.

[noissue]


setup.py
```diff
@@ -17,7 +17,7 @@
     'rq~=1.1.0',
     'redis~=3.1.0',
     'setuptools>=41.0.1,<41.3.0',
-    'dynaconf>=2.0.3,<2.2.0',
+    'dynaconf>=2.1.0,<2.2.0',
```
daviddavis: Would recommend "~=2.1.0". I don't know why dependabot keeps switching off `~=`.

lubosmj: You mean to use only "~=2.1.0" instead of ">=2.1.0,<2.2.0"?

PR #293, opened 2019-09-10
Replace ProgressBar with ProgressReport

fabricio-aguiar:
closes #5444
Required PR: https://github.com/pulp/pulpcore-plugin/pull/127
https://pulp.plan.io/issues/5444


pulpcore/app/migrations/0001_initial.py
```diff
@@ -386,17 +386,6 @@ class Migration(migrations.Migration):
             },
             bases=('core.progressreport',),
         ),
-        migrations.CreateModel(
```
bmbouter: Can we apply this as a new migration instead? That will allow us to keep pre-production upgrades simpler.

fabricio-aguiar: yes, I will change it

fabricio-aguiar: @bmbouter apparently django doesn't make migration to remove proxy model:
```
(pulp) [vagrant@pulp3-source-fedora30 ~]$ django-admin makemigrations
No changes detected
```

bmbouter:
This PR started as a bugfix, but then it transitioned more into doing the work of [5444](https://pulp.plan.io/issues/5444) which I believe is right because we want to get the big parts right and then handle bugs on them. Can this PR become dedicated to 5444 and not include the bugfix? My concern is that we would be spreading out the 5444 (possible breaking change) across more PRs and time instead of all at once. What do you think?

CHANGES/5444.misc
```diff
@@ -0,0 +1 @@
+Using `ProgressReport` for known and unknown items count.
```
bmbouter: This would be good as a `.feature` so it can be shown (because of the `misc` template not being shown).

pulpcore/app/models/progress.py
```diff
@@ -116,121 +115,6 @@ def __exit__(self, type, value, traceback):
...
-    multiple instances of a specific ProgressBar will diverge as they are written to from separate
-    threads.
-    """
-
-    class Meta:
-        proxy = True
-
     def increment(self):
         """
         Increment done count and save the progress bar.
```
bmbouter: Can `progress bar` be updated please.

pulpcore/app/models/progress.py
```diff
@@ -31,8 +31,7 @@ class ProgressReport(Model):
         code (models.CharField): identifies the type of progress report
```
bmbouter: Can L#25 be updated since it's no longer a base model, but to be used directly.

pulpcore/app/models/progress.py
```diff
@@ -106,8 +105,8 @@ def __exit__(self, type, value, traceback):
         See the context manager documentation for more info on __exit__ parameters
         """
         self._using_context_manager = False
-        if self.total is None and self.done != 0:
-            self.total = self.done
+        if self.total is None:
```
bmbouter: I'm not sure why we have this conditional case. Can we do away with this part?

dralley:
Since we need to redo migrations for this change anyways, I decided to make another change that would also require redoing the migrations, and we can merge them around the same time to avoid doing it more than once.

https://github.com/pulp/pulpcore/pull/301

pulpcore/app/models/progress.py
```diff
@@ -20,8 +20,6 @@
 
 class ProgressReport(Model):
     """
-    A base model for all progress reporting.
```
mdellweg: I would keep that line and just remove the 'base'.

pulpcore/app/models/progress.py
```diff
@@ -116,139 +111,24 @@ def __exit__(self, type, value, traceback):
...
-        >>> spinner.save() # spinner is saved as 'waiting'
-        >>> with spinner:
-        >>>     # spinner is saved as 'running'
-        >>>     publish_metadata()
-        >>> # spinner is saved as 'completed' if no exception is raised or 'failed' otherwise
-
-    You can also use this short form which handles all necessary save() calls:
-
-        >>> with ProgressSpinner(message='Publishing Metadata'):
-        >>>     publish_metadata()
```
mdellweg: It's a pity, that we loose all those good code examples.

bmbouter: I agree and didn't realize that. @fabricio-aguiar can we move these back to the docstrings we are keeping?

pulpcore/app/models/progress.py
```diff
@@ -116,139 +111,24 @@ def __exit__(self, type, value, traceback):
...
-        >>>         handle(file)
-
-    ProgressBar objects are associated with a Task and auto-discover and populate the task id when
-    saved.
-
-    When using threads to update a ProgressBar in parallel, it is recommended that all threads
-    share the same in-memory instance. Django does not synchronize in-memory model instances, so
-    multiple instances of a specific ProgressBar will diverge as they are written to from separate
-    threads.
-    """
```
mdellweg: ... I think the documentation should transmogrified into the foremer base model.

PR #294, opened 2019-09-11
Rename ContentSerializer fields

mdellweg:
fixes #5428
https://pulp.plan.io/issues/5428


CHANGES/5428.misc
```diff
@@ -0,0 +1 @@
+Rename the fields on the ContentSerializers to not start with underscore.
```
daviddavis: This needs to be `5428.removal` since this is a breaking change.

pulpcore/app/serializers/content.py
```diff
@@ -27,16 +27,26 @@ class Meta:
...
         write_only=True,
     )
 
+    def __init__(self, *args, **kwargs):
+        """
+        Initializer for SingleArtifactContentSerializer
+        """
+        super().__init__(*args, **kwargs)
+
+        # If the content model wants to safe the relative_path, we mark the field as readable
```
daviddavis: I'm not clear on this comment (to safe doesn't make sense). Are you saying something like `If the content model has a saved relative_path, we mark the field as readable`?

mdellweg: Sth like `If the model has it's own relativ_path field (in the database), we mark the field as not write_only.`

PR #300, opened 2019-09-13
filter plugin managed repo

fabricio-aguiar:
ref #5421
https://pulp.plan.io/issues/5421


bmbouter:
Can we brainstorm on some alternative names instead of `is_subrepo`? My concern is that it's not self-descriptive enough to make sense for other plugins. I'm thinking of a situation like docker where the repo is created by the plugin at docker push perhaps, but it isn't a sub-repo par-se. Maybe a name like `plugin_managed` or `user_created`? What do you think?

bmbouter:
@fabricio-aguiar, @dralley, @daviddavis I want to help plan this feature a little more so I'll recap some of what I saw on IRC for clarity here.

Is the idea that Repository.parent_repo would be an optional foreign key with the reverse set named `child_repos`? Both of these would be exposed in the `RepositorySerializer`?

Also what about the `on_delete` behavior?

CHANGES/5421.feature
```diff
@@ -0,0 +1 @@
+Filter sub repositories.
```
bmbouter: Can the changelog be updated to indicate plugin managed instead?

pulpcore/app/serializers/repository.py
```diff
@@ -30,6 +30,10 @@ class RepositorySerializer(ModelSerializer):
         help_text=_('A unique name for this repository.'),
         validators=[UniqueValidator(queryset=models.Repository.objects.all())]
     )
+    plugin_managed = serializers.BooleanField(
+        help_text=_('True if plugin managed to create a sub-repository.'),
```
bmbouter: Can this sub-repository be rewritten . I think it's more like `True if the plugin manages this repository and users typically do not interact with it`.  Or maybe `True if managed by a plugin` although that doesn't tell you much more than the label.

pulpcore/app/viewsets/repository.py
```diff
@@ -41,10 +41,14 @@
...
 class RepositoryFilter(BaseFilterSet):
     name = filters.CharFilter()
+    plugin_managed = filters.BooleanFilter()
 
     class Meta:
         model = Repository
-        fields = {'name': NAME_FILTER_OPTIONS}
+        fields = {
+            'name': NAME_FILTER_OPTIONS,
+            'plugin_managed': ['exact'],
```
bmbouter: Can this be set to False somehow? I think `plugin_managed=True` should be disincluded from RepositoryList by default.

daviddavis:
I have some concerns about modeling the parent-child relationship in core at this point. Namely, we don't know what other plugins needs will look like. For one, they might want to have a many-to-many relationship instead of one-to-many. Also, they may want to store additional information in this relationship. I think having a boolean for now is a safer approach as it would not preclude a parent-child repo relationship later.

CHANGES/5421.feature
daviddavis: Also, we should probably create a PR against pulpcore-plugin with `5421.feature` that says something like, "Added plugin_managed field on repository that plugins can set for repositories that should be managed by plugins and not by users".

daviddavis:
I see that we're just now excluding all plugin_managed repos from being returned in the API. Was that intentional? I thought we were going to have a plugin_managed filter field and defaulting it to `False`?

fabricio-aguiar:
> I see that we're just now excluding all plugin_managed repos from being returned in the API. Was that intentional? I thought we were going to have a plugin_managed filter field and defaulting it to `False`?
> 
> Edit: I am fine with not returning plugin_managed repos in the API but if we do that, we probably shouldn't expose the field as it will always be False.

I probably misunderstood then, I will work on this

pulpcore/app/models/repository.py
```diff
@@ -23,6 +23,8 @@ class Repository(Model):
 
         name (models.CharField): The repository name.
         description (models.TextField): An optional description.
+        plugin_managed (models.BooleanField): True if the plugin manages this repository and users
+                                              typically do not interact with it.
```
bmbouter: The usual style is to line wrap this 4 chars. Can this be made like that for consistency (see example below)?

pulpcore/app/viewsets/repository.py
```diff
@@ -53,7 +63,7 @@ class RepositoryViewSet(NamedModelViewSet,
                         mixins.RetrieveModelMixin,
                         mixins.ListModelMixin,
                         mixins.DestroyModelMixin):
-    queryset = Repository.objects.all()
+    queryset = Repository.objects.exclude(plugin_managed=True)
```
bmbouter: :+1:

pulpcore/app/viewsets/repository.py
bmbouter: This was resolved by the queryset L#56. Thanks.

pulpcore/app/viewsets/repository.py
daviddavis: Doesn't this mean that plugin_managed repos are never returned? I tested by querying `/pulp/api/v3/repositories/?plugin_managed=False` and it seems to confirm that.

fabricio-aguiar: plugin_managed will only be returned if you do:
`/pulp/api/v3/repositories/?plugin_managed=True`
otherwise, it won't be returned

daviddavis: Right, nevermind me. :+1:

daviddavis: Ah, I see how this works. The filter overrides the queryset. But now it looks like other filters don't work:

```
$ http ":24817/pulp/api/v3/repositories/?plugin_managed=True&name=nil"

{
    "count": 1,
    "next": null,
    "previous": null,
    "results": [
        {
            "_created": "2019-09-17T17:57:01.986549Z",
            "_href": "/pulp/api/v3/repositories/885643cc-49ec-402f-840e-948bd341798c/",
            "_latest_version_href": null,
            "_versions_href": "/pulp/api/v3/repositories/885643cc-49ec-402f-840e-948bd341798c/versions/",
            "description": null,
            "name": "something",
            "plugin_managed": true
        }
    ]
}
```

CHANGES/5421.feature
fabricio-aguiar: https://github.com/pulp/pulpcore-plugin/pull/131

PR #301, opened 2019-09-17
Remove custom JSONField implementation

dralley:
This was quick to do and since it requires a redo of all the migrations I figured it would be good timing to do it at the same time as Fabricio's PR that will also need to redo the migrations


daviddavis:
This change LGTM. Just need to fix Travis.

PR #303, opened 2019-09-18
Store PublishedMetadata files in artifact storage

dkliban:
This patch changes PublishedMetadata into Content. This requires migrating the database.
The migration that comes with this change will only work on an empty database. It will
fail to migrate any existing PublishedMetadata. This is a backwards incompatible change.

The PublishedArtifact now has a classmethod called 'create_from_file'. This method creates
PublishedMetadata along with an Artifact, ContentArtifact, and PublishedArtifact.

Required PR: https://github.com/pulp/pulp_file/pull/278/

closes: #5304
https://pulp.plan.io/issues/5304


pulpcore/app/models/publication.py
```diff
@@ -168,23 +157,58 @@ class Meta:
...
 
-    file = models.FileField(upload_to=_storage_path, max_length=255)
+        Returns:
+            PublishedMetadata: a saved instance of PublishedMetadata
+        """
+
+        with transaction.atomic():
+            artifact = Artifact.init_and_validate(file=PulpTemporaryUploadedFile.from_file(file))
+            try:
+                with transaction.atomic():
```
daviddavis: What's this needed for? This code is already in a transaction (line 190).

dkliban: This is needed so the outer transaction doesn't roll back. You can read more about it here https://docs.djangoproject.com/en/2.2/topics/db/transactions/#controlling-transactions-explicitly

pulpcore/content/handler.py
```diff
@@ -213,14 +213,6 @@ def _permit(request, distribution):
                 else:
                     return await self._stream_content_artifact(request, StreamResponse(), ca)
 
-            # published metadata
-            try:
-                pm = publication.published_metadata.get(relative_path=rel_path)
-            except ObjectDoesNotExist:
-                pass
-            else:
-                return self._handle_file_response(pm.file)
```
daviddavis: :+1:

bmbouter:
This code all looks good to me. We do need changes in the pulpcore-plugin with it including its docs area.

pulpcore/app/models/publication.py
```diff
@@ -168,23 +157,58 @@ class Meta:
         )
 
 
-class PublishedMetadata(PublishedFile):
+class PublishedMetadata(Content):
```
bmbouter: If the default constructor is not usable, can we override it and have it raise an Exception? Wdyt?

dkliban: The default constructor is usable for creating the PublishedMetadata content unit. It is used by Django internals. However, plugin writers are discouraged from using it because it requires them to create Artifact, ContentArtifact, and PublishedArtifact all on their own.

bmbouter: Thank you for the info.

dralley:
@dkliban could you rebase and rename the migration since there is a new migration #7 now?

pulpcore/app/models/publication.py
```diff
@@ -168,23 +156,58 @@ class Meta:
...
```
goosemania: @dkliban , what do you think about defining a `TYPE` class attribute?  Every other `Content` has it and there is an assumption that every Content will define it. It's used for generating a [ _type](https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/base.py#L106) which is user-facing and stored in the DB.
It can potentially help with the logic for an orhan cleanup.

goosemania:
@dkliban , a part of the redmine ticket is a change in the orhpan cleanup logic. Will it be handled in this PR or a separate one? If separate, maybe it's worth filing a separate redmine task. What do you think?

dkliban:
@goosemania I don't think a change in orphan cleanup is actually needed. The PublishedMetadata is going to be removed when a Publication is deleted[0].

[0] https://github.com/pulp/pulpcore/pull/303/files#diff-4cf3b583f3d4f0b078706dca1308e77aR172

dkliban:
@goosemania You are right. We need to make sure orphan cleanup doesn't consider PublishedMetadata orphaned. I'll update my PR.

pulpcore/app/tasks/orphan.py
```diff
@@ -11,6 +11,7 @@ def orphan_cleanup():
     # Content cleanup
     content = Content.objects.exclude(pk__in=RepositoryContent.objects.values_list('content_id',
                                                                                    flat=True))
+    content = content.exclude(_type='core.{}'.format(PublishedMetadata.TYPE))
```
goosemania: @dkliban , just a suggestion. Maybe, just `_type=PublishedMetadata._type`?, so we are not reconstructing the _type in multiple places?

PR #305, opened 2019-09-19
Add context to the general_create task

mdellweg:
re #5403
https://pulp.plan.io/issues/5403

### DIFF
--- a/pulpcore/app/tasks/base.py
+++ b/pulpcore/app/tasks/base.py
@@ -11,8 +11,9 @@ def general_create(app_label, serializer_name, *args, **kwargs):
 
     """
     data = kwargs.pop('data', None)
+    context = kwargs.pop('context', {})
     serializer_class = get_plugin_config(app_label).named_serializers[serializer_name]
-    serializer = serializer_class(data=data)
+    serializer = serializer_class(data=data, context=context)
     serializer.is_valid(raise_exception=True)
     serializer.save()
     resource = CreatedResource(content_object=serializer.instance)
--- a/CHANGES/5403.feature
+++ b/CHANGES/5403.feature
@@ -1,0 +1,1 @@
+Add the possibility to pass context to the general_create task.


pulpcore/app/tasks/base.py
```diff
@@ -11,8 +11,9 @@ def general_create(app_label, serializer_name, *args, **kwargs):
 
     """
     data = kwargs.pop('data', None)
+    context = kwargs.pop('context', {})
     serializer_class = get_plugin_config(app_label).named_serializers[serializer_name]
-    serializer = serializer_class(data=data)
+    serializer = serializer_class(data=data, context=context)
```
bmbouter: Will this cause an issue when `context` is passed as a kwarg to other Serializers besides `SingleArtifactContentUploadSerializer` do you think?

mdellweg: They should know how to handle / ignore context.

See:
https://www.django-rest-framework.org/api-guide/serializers/#including-extra-context

bmbouter:
Also what about docs for this feature? I think some small updates to this section specifically would be good: https://docs.pulpproject.org/en/3.0/nightly/workflows/upload-publish.html

What do you think?

bmbouter:
I couldn't push to your branch. Maybe I'm doing this wrong so I'll post the commands here:

```
[bmbouter@localhost pulpcore]$ git push mdellweg HEAD:upload_create
ERROR: Permission to mdellweg/pulpcore.git denied to bmbouter.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

[bmbouter@localhost pulpcore]$ git remote -vv
mdellweg	git@github.com:mdellweg/pulpcore.git (fetch)
mdellweg	git@github.com:mdellweg/pulpcore.git (push)
<snip>
```

I opened another PR here:  https://github.com/pulp/pulpcore/pull/309

mdellweg:
@bmbouter 
Thanks for adding the Docs.

PR #308, opened 2019-09-23
Remove NamePagination class

mdellweg:
Sorting of the results can be accomplished on the queryset.

fixes #5489
https://pulp.plan.io/issues/5489


daviddavis:
LGTM. Mind sending out a '[BREAKING]' announcement to pulp-dev?

mdellweg:
@daviddavis So this is going to be a *thing*?
I'll write something. Is there an estimated time of merge?

daviddavis:
Yes, but only until 3.0 GA probably as we shouldn't have breaking changes after 3.0 GA (due to semver). 

Let's merge tomorrow I think. I don't think there are plugins using this.

PR #310, opened 2019-09-27
Set Content-Type and Content-Encoding headers in content-app responses

dkliban:
closes: #5507
https://pulp.plan.io/issues/5507

### DIFF
--- a/CHANGES/5507.bugfix
+++ b/CHANGES/5507.bugfix
@@ -1,0 +1,1 @@
+content-app sets Content-Type and Content-Encoding headers for all responses.
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -1,4 +1,5 @@
 import logging
+import mimetypes
 import os
 from gettext import gettext as _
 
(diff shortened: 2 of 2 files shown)


CHANGES/5507.bugfix
```diff
@@ -0,0 +1,2 @@
+content-app now sets Content-Type and Content-Encoding headers for all responses.
+
```
bmbouter: nitpick: you probably don't need the extra line, just a newline after your first line.

pulpcore/content/handler.py
```diff
@@ -393,10 +395,17 @@ def _handle_file_response(self, file):
         Returns:
             The :class:`aiohttp.web.FileResponse` for the file.
         """
+        file = content_artifact.artifact.file
+        content_type, encoding = mimetypes.guess_type(content_artifact.relative_path)
```
bmbouter: :+1:

bmbouter:
I reviewed this once more and it still looks good to me.

PR #315, opened 2019-09-30
fixing repository filter

fabricio-aguiar:
closes #5516
https://pulp.plan.io/issues/5516

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -304,8 +304,9 @@ class RepositoryVersion(Model):
                 # now add any content that's in the base_version but not in version
                 version.add_content(base_version.content.exclude(pk__in=version.content))
 
-            resource = CreatedResource(content_object=version)
-            resource.save()
+            if not repository.plugin_managed:
+                resource = CreatedResource(content_object=version)
+                resource.save()
             return version
 
     @staticmethod
--- a/pulpcore/app/viewsets/repository.py
+++ b/pulpcore/app/viewsets/repository.py
@@ -40,20 +40,10 @@ from pulpcore.tasking.tasks import enqueue_with_reservation
 
 class RepositoryFilter(BaseFilterSet):
     name = filters.CharFilter()
-    plugin_managed = filters.BooleanFilter(method='filter_plugin_managed')
-
-    def filter_plugin_managed(self, queryset, name, value):
-        if value:
-            return Repository.objects.filter(**self.data.dict())
-
-        return queryset
 
     class Meta:
         model = Repository
-        fields = {
-            'name': NAME_FILTER_OPTIONS,
-            'plugin_managed': ['exact'],
-        }
+        fields = {'name': NAME_FILTER_OPTIONS}
 
 
 class RepositoryViewSet(NamedModelViewSet,
--- a/CHANGES/5516.removal
+++ b/CHANGES/5516.removal
@@ -1,0 +1,1 @@
+Removing filter for `plugin_managed` repositories.


pulpcore/app/viewsets/repository.py
```diff
@@ -40,13 +40,30 @@
 
 class RepositoryFilter(BaseFilterSet):
     name = filters.CharFilter()
-    plugin_managed = filters.BooleanFilter(method='filter_plugin_managed')
+    plugin_managed = filters.BooleanFilter(method='filter_plugin_managed', initial=False)
```
bmbouter: We we ok with the downsides stated here?

https://django-filter.readthedocs.io/en/master/guide/tips.html#using-initial-values-as-defaults

fabricio-aguiar: the other way that I tried didn't work, then @daviddavis suggested this approach

daviddavis: Those are arguments against using a default value for a filter. The other option is to not have a default value and show all repos including plugin managed ones or--and I am actually thinking this might be the best way to go--not show plugin managed repos to the user. If plugins want to expose their plugin managed repos, we should let them do it.

bmbouter: Having `plugin_managed=True` not able to be visible/modified/deleted/managed by users sounds very straightforward. With this change though a user couldn't see the rpms listed themselves. Would that be a problem?

@jlsherrill this change would cause kitckstart repos to be forever hidden, but usable by virtue of the single DistributionTree content unit. How does this sound?

daviddavis: Not forever hidden necessarily--just hidden until/if pulp_rpm decides to expose it. I think of this change as ceding control of plugin managed repos to plugins.

bmbouter: That makes sense to me.

bmbouter: I'm +1 on on that if @jlsherrill acks it

bmbouter: Would this change also be removing the ability to list the repository via core URL as a detail view? If we loose that, then the pulp_rpm DistributionTree serializer couldn't show the repo because it wouldn't have a URL to link to. I guess we would have to add those views to pulp_rpm, but in practice that doesn't seem as good.

So is having the list view only always disinclude `plugin_managed=True` but having the detail work an option?

jlsherrill: I think this is fine from a katello perspective.  I can't foresee any issues with it at this time.

daviddavis: @bmbouter I would probably defer to allowing plugins to define the detail view. They may not want to show these repos at all to the user.

fabricio-aguiar: @bmbouter I've found this bug when I was writing performance tests for rpm, with this solution I was able to run these tests: https://github.com/pulp/pulp_rpm/pull/1459

bmbouter: @daviddavis I can see what you're saying. Let's remove them altogether as you suggest.

What should we do for pulp_rpm then? I believe the DistributionTree serializer links to the repo, so if we do nothing those links would be 404's. Maybe we modify the serializer to not include that data in the serializer? Or?

fabricio-aguiar: I was getting 404 before this PR, with this PR I got all green on https://github.com/pulp/pulp_rpm/pull/1459

daviddavis: @bmbouter what about an endpoint like `/pulp/api/v3/content/rpm/distribution_trees/<uuid>/repositories/` ?

fabricio-aguiar: we already have `repositories` under addons/variants
```
GET /pulp/api/v3/distribution_trees/e4fc57f4-b066-4f0c-86cb-255170d1a747/
HTTP 200 OK
Allow: GET, DELETE, HEAD, OPTIONS
Vary: Accept
```

```json
{
    "_href": "/pulp/api/v3/distribution_trees/e4fc57f4-b066-4f0c-86cb-255170d1a747/",
    "header_version": "1.2",
    "release_name": "CentOS",
    "release_short": "CentOS",
    "release_version": "7",
    "release_is_layered": false,
    "base_product_name": null,
    "base_product_short": null,
    "base_product_version": null,
    "arch": "x86_64",
    "build_timestamp": 1567770561.0,
    "instimage": null,
    "mainimage": "LiveOS/squashfs.img",
    "discnum": null,
    "totaldiscs": null,
    "addons": [],
    "checksums": [],
    "images": [
        {
            "name": "kernel",
            "path": "images/pxeboot/vmlinuz",
            "platforms": "x86_64, xen",
            "artifact": null
        },
        {
            "name": "initrd",
            "path": "images/pxeboot/initrd.img",
            "platforms": "x86_64, xen",
            "artifact": null
        },
        {
            "name": "boot.iso",
            "path": "images/boot.iso",
            "platforms": "x86_64",
            "artifact": null
        }
    ],
    "variants": [
        {
            "variant_id": "CentOS",
            "uid": "CentOS",
            "name": "CentOS",
            "type": "variant",
            "packages": ".",
            "repository": "http://pulp3-source-fedora30.localhost.example.com/pulp/api/v3/repositories/755ab7e2-5e5f-4129-8208-88b7b479f122/",
            "source_packages": null,
            "source_repository": null,
            "debug_packages": null,
            "debug_repository": null,
            "identity": null
        }
    ]
}
```

bmbouter: I think the idea is that the list view for `/pulp/api/v3/repositories/` would not show `plugin_managed=True` repositories and the user couldn't filter them to be shown. Also they wouldn't be viewable at the detail view `/pulp/api/v3/repositories/<UUID>/`. @daviddavis is this right?

daviddavis: @bmbouter correct. I would remove the repository field for variants (for now).

daviddavis:
So talking with @jlsherrill and @goosemania on IRC, I think we can just have core hide plugin_managed repos.

So next steps:
1. Remove the plugin_managed filter
2. Set `queryset = Repository.objects.filter(plugin_managed=False)` in `RepositoryViewSet`
3. Remove the repository field on variant in pulp_rpm

I would also remove the parent code for now. We can add it later when needed.

fabricio-aguiar:
> So talking with @jlsherrill and @goosemania on IRC, I think we can just have core hide plugin_managed repos.
> 
> So next steps:
> 
> 1. Remove the plugin_managed filter
> 2. Set `queryset = Repository.objects.filter(plugin_managed=False)` in `RepositoryViewSet`
> 3. Remove the repository field on variant in pulp_rpm
> 
> I would also remove the parent code for now. We can add it later when needed.

it is needed due to kickstart, a `RepositoryVersion` is created for each plugin managed repo

pulpcore/app/models/repository.py
```diff
@@ -304,8 +304,9 @@ def create(cls, repository, base_version=None):
                 # now add any content that's in the base_version but not in version
                 version.add_content(base_version.content.exclude(pk__in=version.content))
 
-            resource = CreatedResource(content_object=version)
-            resource.save()
+            if not repository.plugin_managed:
+                resource = CreatedResource(content_object=version)
```
fabricio-aguiar: due to https://github.com/PulpQE/pulp-smash/blob/master/pulp_smash/api.py#L263-L264

CHANGES/5516.bugfix
```diff
@@ -0,0 +1 @@
+Fixing filter for `plugin_managed` repositories.
```
daviddavis: This should probably be `5516.removal` with something like `Hiding plugin_managed repos from API`.

bmbouter:
Would this still show in the list/view, or is that coming as another PR?

fabricio-aguiar:
> Would this still show in the list/view, or is that coming as another PR?

it doesn't shoe on list/view,
from what I understood, it would be done on the plugins side, is that correct @daviddavis ?

daviddavis:
@fabricio-aguiar correct

bmbouter:
@fabricio-aguiar @daviddavis Thank you, this sounds great.

PR #324, opened 2019-10-04
Fix schemas

fabricio-aguiar:
closes #5311
https://pulp.plan.io/issues/5311

### DIFF
--- a/pulpcore/app/viewsets/base.py
+++ b/pulpcore/app/viewsets/base.py
@@ -8,7 +8,7 @@ from django_filters.rest_framework import filterset
 from drf_yasg.utils import swagger_auto_schema
 from rest_framework import viewsets
 from rest_framework.generics import get_object_or_404
-from rest_framework.schemas import AutoSchema
+from rest_framework.schemas.openapi import AutoSchema
 from rest_framework.serializers import ValidationError as DRFValidationError
 
 from pulpcore.app import tasks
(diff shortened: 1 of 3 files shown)


pulpcore/app/viewsets/upload.py
```diff
@@ -60,7 +60,7 @@ def get_serializer_class(self):
 
     def get_serializer_context(self):
         context = super().get_serializer_context()
-        if self.action == 'update':
+        if self.action == 'update' and self.kwargs.get("pk"):
```
bmbouter: Is this related to the 500? I'm just curious.

pulpcore/app/viewsets/base.py
```diff
@@ -8,7 +8,7 @@
 from drf_yasg.utils import swagger_auto_schema
 from rest_framework import viewsets
 from rest_framework.generics import get_object_or_404
-from rest_framework.schemas import AutoSchema
+from rest_framework.schemas.openapi import AutoSchema
```
bmbouter: I can see how this fixes the error.

pulpcore/app/viewsets/upload.py
fabricio-aguiar: No, it is for fixing this:
```
AssertionError: Expected view UploadViewSet to be called with a URL keyword argument named "pk". Fix your URL conf, or set the `.lookup_field` attribute on the view correctly.
```

bmbouter: Makes sense. Thanks!

PR #330, opened 2019-10-07
Load /etc/pulp/settings.py last when loading settings

dkliban:
fixes: #5425
https://pulp.plan.io/issues/5425

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,4 +103,5 @@ notifications:
     - 'Change view : %{compare_url}'
     - 'Build details : %{build_url}'
     - '%{message}'
+
 ...
--- a/.travis/before_install.sh
+++ b/.travis/before_install.sh
@@ -19,6 +19,11 @@ if [ -f $PRE_BEFORE_INSTALL ]; then
     $PRE_BEFORE_INSTALL
 fi
 
+if [[ -n $(echo -e $COMMIT_MSG | grep -P "Required PR:.*" | grep -v "https") ]]; then
+  echo "Invalid Required PR link detected in commit message. Please use the full https url."
+  exit 1
+fi
+
 export PULP_PR_NUMBER=$(echo $COMMIT_MSG | grep -oP 'Required\ PR:\ https\:\/\/github\.com\/pulp\/pulpcore\/pull\/(\d+)' | awk -F'/' '{print $7}')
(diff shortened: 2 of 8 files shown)


dkliban:
In it's current state, this patch makes it impossible to set PULP_SETTINGS environment variable. This settings is supposed to specify where the settings file is located. However, this patch hard codes it to /etc/pulp/settings.py

bmbouter:
I see what you're saying. I think we should merge this for now, and file a separate issue dedicated to this. We'll need input from @rochacbruno I think on how to have dynaconf allow this. Wdyt about this path?

rochacbruno:
@bmbouter @dkliban I am doing some experiments, let me see if I can get in to a solution that fits both requirements (PULP_SETTINGS envvar + correct order of loading)

dkliban:
I opened an issue to track the problem with PULP_SETTINGS not working.

rochacbruno:
@dkliban @bmbouter 

https://github.com/rochacbruno/dynaconf/releases/tag/2.2.0

Released with:

```py
from dynaconf import LazySettings

settings = LazySettings(
  PRELOAD_FOR_DYNACONF=["/path/*", "other/settings.toml"],                # <-- Loaded first
  SETTINGS_FILE_FOR_DYNACONF="/etc/foo/settings.py",                      # <-- Loaded second (the main file)
  INCLUDES_FOR_DYNACONF=["other.module.settings", "other/settings.yaml"]  # <-- Loaded at the end
)

```

https://dynaconf.readthedocs.io/en/latest/guides/advanced_usage.html#preloading-files

Now you can do:

```py
# HERE STARTS DYNACONF EXTENSION LOAD (Keep at the very bottom of settings.py)
# Read more at https://dynaconf.readthedocs.io/en/latest/guides/django.html
import dynaconf  # noqa
settings = dynaconf.DjangoDynaconf(
    __name__,
    GLOBAL_ENV_FOR_DYNACONF='PULP',
    ENV_SWITCHER_FOR_DYNACONF='PULP_ENV',
    PRELOAD_FOR_DYNACONF=[
        '{}.app.settings'.format(plugin_name)
        for plugin_name in INSTALLED_PULP_PLUGINS
    ],
    SETTINGS_FILE_FOR_DYNACONF= "/etc/pulp/settings.py",
    ENVVAR_FOR_DYNACONF='PULP_SETTINGS',
)
``` 

Must load plugin files first and keep PULP_SETTINGS envvar working.

PR #331, opened 2019-10-07
Add repo_key support for `add_content`

bmbouter:
The `Content` model now supports a `repo_key` attribute which defaults
to (). Plugin writers can specify this, and the tuple field names will
automatically replace another unit of the same type with these field
names.

Required PR: https://github.com/pulp/pulp_file/pull/290

https://pulp.plan.io/issues/5008
closes #5008

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -1,10 +1,14 @@
 """
 Repository related Django models.
 """
+from collections import defaultdict
 from contextlib import suppress
+from gettext import gettext as _
+import logging
 
 import django
 from django.db import models, transaction
+from django.db.models import Q
 from django.urls import reverse
 
 from pulpcore.app.util import get_view_name_for_model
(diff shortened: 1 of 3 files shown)


pulpcore/app/models/repository.py
```diff
@@ -351,9 +358,23 @@ def add_content(self, content):
...
+            if item.repo_key == ():
+                continue
+            unit_q_dict = {
+                field: getattr(item, field) for field in item.repo_key
+            }
+            query_for_repo_duplicates_by_type[item._meta.model] |= Q(**unit_q_dict)
+
+        for model in query_for_repo_duplicates_by_type:
+            qs = model.objects.filter(query_for_repo_duplicates_by_type[model])
+            self.remove_content(qs)
```
daviddavis: Maybe add a logger debug message here that dupe content is being removed? Your call.

PR #336, opened 2019-10-11
Change relative_path from CharField to TextField

fabricio-aguiar:
closes #4544
https://pulp.plan.io/issues/4544


daviddavis:
Did you check other repos? I know for example pulp_file has relative_path:

https://github.com/pulp/pulp_file/blob/master/pulp_file/app/models.py#L26

fabricio-aguiar:
> Did you check other repos? I know for example pulp_file has relative_path:
> 
> https://github.com/pulp/pulp_file/blob/master/pulp_file/app/models.py#L26

No, I haven't, I thought the issue was specific for pulpcore

daviddavis:
I think it would be good to at least file issues for other plugins. I suspect though that pulp_file is the only plugin that has a relative_path.

fabricio-aguiar:
> I think it would be good to at least file issues for other plugins. I suspect though that pulp_file is the only plugin that has a relative_path.

I will check it out and open PR for all plugins that have it

PR #342, opened 2019-10-16
Remove hard coded `/etc/pulp/settings.py` file

bmbouter:
This changes Pulp to not assume any settings file is needed or looked
for at all. It encourages the user to set one via the `PULP_SETTINGS`
Environment Variable. The settings documentation was updated to reflect
this.

The Installer will continue working as is. The PR below causes the
installer to set `PULP_SETTINGS` correctly allowing settings to work the
same way in those environments.

Required PR: https://github.com/pulp/ansible-pulp/pull/182

https://pulp.plan.io/issues/5560
closes #5560

### DIFF
--- a/docs/installation/configuration.rst
+++ b/docs/installation/configuration.rst
@@ -10,10 +10,20 @@ to configure Pulp in a few ways:
 By Configuration File
 ---------------------
 
-Non-default settings can be specified in the ``/etc/pulp/settings.py``. The presence of this file is
-optional. The expected location and format can be changed by specifying the ``PULP_SETTINGS``
-environment variable. Dynaconf supports `settings in multiple file formats <https://dynaconf.
-readthedocs.io/en/latest/guides/examples.html>`_
+To configure Pulp by settings file, you must set the format and location of the config file by
+specifying the ``PULP_SETTINGS`` environment variable. For example, if you wanted to use Python to
+specify your configuration, and provide it at ``/etc/pulp/settings.py`` you could::
+
+    export PULP_SETTINGS=/etc/pulp/settings.py
+
+
+Or in a systemd file you could::
+
+    Environment="PULP_SETTINGS=/etc/pulp/settings.py" as the Ansible Installer does.
+
+
+Dynaconf supports `settings in multiple file formats <https://dynaconf.readthedocs.io/en/latest/
+guides/examples.html>`_
 
 This file should have permissions of:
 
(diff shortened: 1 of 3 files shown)


docs/installation/configuration.rst
```diff
@@ -10,10 +10,20 @@ to configure Pulp in a few ways:
...
+Users can specify the location and format of the config file by specifying the ``PULP_SETTINGS``
+environment variable. For example, if you wanted to use Python to specify your configuration, and
+provide it at ``/etc/pulpsettings.py`` you could::
+
+    export PULP_SETTINGS=/etc/pulp/settings.py
+
+
+Or in a systemd file you could::
+
+    Environment="PULP_SETTINGS={{ pulp_config_dir }}/settings.py" as the Ansible Installer does.
```
mikedep333: pulp_config_dir is a variable provided via ansible, so users would not know it. Replace "{{ pulp_config_dir }}" with "/etc/pulp"

docs/installation/configuration.rst
```diff
@@ -10,10 +10,20 @@ to configure Pulp in a few ways:
 By Configuration File
 ---------------------
 
-Non-default settings can be specified in the ``/etc/pulp/settings.py``. The presence of this file is
-optional. The expected location and format can be changed by specifying the ``PULP_SETTINGS``
-environment variable. Dynaconf supports `settings in multiple file formats <https://dynaconf.
-readthedocs.io/en/latest/guides/examples.html>`_
+Users can specify the location and format of the config file by specifying the ``PULP_SETTINGS``
```
mikedep333: I would make it clear that if you are providing the config via config file, the PULP_SETTINGS **must** be defined.

docs/installation/configuration.rst
bmbouter: Great point. Thanks you. Fixed in next push.

docs/installation/configuration.rst
bmbouter: I agree. I rewrote this and pushed again.

docs/installation/configuration.rst
```diff
@@ -10,10 +10,20 @@ to configure Pulp in a few ways:
...
 
-Non-default settings can be specified in the ``/etc/pulp/settings.py``. The presence of this file is
-optional. The expected location and format can be changed by specifying the ``PULP_SETTINGS``
-environment variable. Dynaconf supports `settings in multiple file formats <https://dynaconf.
-readthedocs.io/en/latest/guides/examples.html>`_
+To configure Pulp by settings file, you must set the format and location of the config file by
+specifying the ``PULP_SETTINGS`` environment variable. For example, if you wanted to use Python to
+specify your configuration, and provide it at ``/etc/pulp/settings.py`` you could::
+
+    export PULP_SETTINGS=/etc/pulp/settings.py
```
mikedep333: I just realized that in [docs/installation/instructions.rst](https://github.com/pulp/pulpcore/blob/master/docs/installation/instructions.rst) , they'll have to set this as well. That's where they run django-admin commands. However, DJANGO_SETTINGS_MODULE is not listed there either. There are some other files under "docs" with "django-admin" as well, but it may not be necessary to update them.

bmbouter: Yeah we'll have to improve those docs some. With the CI finished I'll merge now, and we can make more improvements later.

PR #345, opened 2019-10-21
Set default `ref_name` attribute for a ModelSerializers's `Meta` class.

### LINKED ISSUE #5574: Fix aiofiles UB in 3.22 to <23.3.0 instead of <=
aiofiles==23.3.0 was never released, and our new strict-UB-check is failing trying to install it.

gmbnomis:
Use the `__init_subclass__` method added in Python 3.6 to customize
subclasses of pulpcore's ModelSerializer class.

If it is not yet defined, set the `Meta.ref_name` attribute according to
the best practice established within Pulp (`'<app label>_<model name>'`).

closes #5574
https://pulp.plan.io/issues/5574

### DIFF
--- a/.travis/test_bindings.py
+++ b/.travis/test_bindings.py
@@ -2,9 +2,9 @@ from pulpcore.client.pulpcore import (ApiClient as CoreApiClient, ArtifactsApi, 
                                       Repository, RepositoriesApi, RepositoriesVersionsApi,
                                       TasksApi, Upload, UploadCommit, UploadsApi)
 from pulpcore.client.pulp_file import (ApiClient as FileApiClient, ContentFilesApi,
-                                       DistributionsFileApi, FileDistribution,
-                                       PublicationsFileApi, RemotesFileApi, FileRemote,
-                                       RepositorySyncURL, FilePublication)
+                                       DistributionsFileApi, FileFileDistribution,
+                                       PublicationsFileApi, RemotesFileApi, FileFileRemote,
+                                       RepositorySyncURL, FileFilePublication)
 from pprint import pprint
 from time import sleep
 import hashlib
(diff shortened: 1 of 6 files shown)


daviddavis:
This seems similar to [the `default_related_name` change we made not too long ago](https://github.com/pulp/pulpcore/blob/a9bf2e54dec7f9a576e39fea191d586eb84769f3/pulpcore/app/models/base.py#L49-L51). However we use raise an exception instead of automatically setting it for the plugin writer. Any reason to follow a different pattern for `ref_name`? Should we switch `default_related_name` code to automatically be set as well?

gmbnomis:
> This seems similar to [the `default_related_name` change we made not too long ago](https://github.com/pulp/pulpcore/blob/a9bf2e54dec7f9a576e39fea191d586eb84769f3/pulpcore/app/models/base.py#L49-L51). However we use raise an exception instead of automatically setting it for the plugin writer. Any reason to follow a different pattern for `ref_name`? Should we switch `default_related_name` code to automatically be set as well?

@daviddavis I was not aware of this discussion. As an alternative, we could provide a class decorator for `Meta` (as originally proposed in issue 5574) and use `__init_subclass__` to raise an error if a Serializer derived from Pulp's ModelSerializer in a plugin does not define `ref_name`. That would be more explicit. (OTOH the current automatic solution just implements a default value that is easy to override by defining a `ref_name` in the Meta class. Effectively, we switch from the default magic done by drf yasg to magic done by pulpcore. we could probably achieve the same for `default_related_name`)

gmbnomis:
> @gmbnomis This change makes sense to me. The OpenAPI schema change that is produced by it could be improved by removing the name spacing from objects provided by pulpcore. Would that be possible?

Yes that is possible. Actually, this PR is outdated a.t.m.; I would like to get a consensus on the way forward before l update the implementation and adapt the rest.

My current solution proposal is at https://pulp.plan.io/issues/5574#note-6.

PR #347, opened 2019-10-22
Move pulpcore-plugin into pulpcore as a subpackage

dralley:
re: #5580
https://pulp.plan.io/issues/5580


CHANGES/5580.misc
```diff
@@ -0,0 +1 @@
+Depend on pulpcore, directly, instead of pulpcore-plugin.
```
bmbouter: I usually put a newline at the end of these snippets so they concatenate together for sure.

pulpcore/stages/test_artifactdownloader.py
```diff
@@ -0,0 +1,364 @@
+import asyncio
```
bmbouter: I think these `test`_` files are in the code base itself, but I expected them in `pulpcore/tests` https://github.com/pulp/pulpcore/tree/master/pulpcore/tests  There should likely be both unit and functional tests. Also I checked Travis and they are not running there. The Travis tests need to pass with these the pulpcore-plugin tests also.

pulpcore/plugin/__init__.py
```diff
@@ -0,0 +1,7 @@
+__version__ = '0.1.0rc8.dev'
```
bmbouter: I was thinking we will not have another version number for this package. We had talked about doing that, but since we couldn't depend on it with setuptools, we can just use `pulpcore`. For example with the changelog, we'll add custom towncrier extensions which will build two "sections" of the changelog, one for users, one for plugin writers. Both would use the version of pulpcore itself though (not this one). @daviddavis @dralley @dkliban does this sound right?

bmbouter: This is such a big PR though I'm in favor of merging as-is and fixing in a separate PR.

daviddavis: That sounds right to be based on my memory of our discussion.

dkliban: This is exactly as I remember it.

dralley: Line deleted

pulpcore/stages/test_artifactdownloader.py
dralley: Fixed

PR #352, opened 2019-10-28
Provide storage information in the status API

ekohl:
Currently Katello has the ability to query the Pulp storage usage. This is implemented via the smart_proxy_pulp project. This requires the Smart Proxy to be colocated with Pulp. The colocation also makes deployment in Kubernetes harder. It won't work with S3 unless a lot of duplication with Pulp itself is done. By providing it in the Pulp API, this can be
avoided.

Currently it's only implemented for local filesystem storage.

This is currently more of a draft to see what the ideas about implementing this are. If this is not desired at all or if this should be a plugin, I'd rather not waste my time writing a fully working implementation. If there is, I'll look into writing the correct tests and making sure this works properly.

https://pulp.plan.io/issues/5631

### DIFF
--- a/pulpcore/app/views/status.py
+++ b/pulpcore/app/views/status.py
@@ -1,6 +1,9 @@
 import logging
+import shutil
 from gettext import gettext as _
 
+from django.conf import settings
+from django.core.files.storage import default_storage
 from drf_yasg.utils import swagger_auto_schema
 from pkg_resources import get_distribution
 from rest_framework.response import Response
@@ -13,6 +16,16 @@ from pulpcore.app.settings import INSTALLED_PULP_PLUGINS
 from pulpcore.tasking.connection import get_redis_connection
 
 _logger = logging.getLogger(__name__)
+
+
+def _disk_usage():
+    if settings.DEFAULT_FILE_STORAGE == 'pulpcore.app.models.storage.FileSystem':
+        try:
+            return shutil.disk_usage(default_storage.location)
+        except Exception:
+            _logger.exception(_('Failed to determine disk usage'))
+
+    return None
 
 
 class StatusView(APIView):
@@ -61,7 +74,8 @@ class StatusView(APIView):
             'missing_workers': missing_workers,
             'online_content_apps': online_content_apps,
             'database_connection': db_status,
-            'redis_connection': redis_status
+            'redis_connection': redis_status,
+            'storage': _disk_usage(),
         }
 
         context = {'request': request}
--- a/pulpcore/tests/functional/api/test_status.py
+++ b/pulpcore/tests/functional/api/test_status.py
@@ -43,6 +43,14 @@ STATUS = {
                 }
             },
         },
+        'storage': {
+            'type': 'object',
+            'properties': {
+                'total': {'type': 'integer'},
+                'used': {'type': 'integer'},
+                'free': {'type': 'integer'},
+            },
+        },
     }
 }
 
@@ -96,3 +104,4 @@ class StatusTestCase(unittest.TestCase):
         self.assertEqual(status['missing_workers'], [])
         self.assertNotEqual(status['online_workers'], [])
         self.assertNotEqual(status['versions'], [])
+        self.assertIsNotNone(status['storage'])
--- a/CHANGES/5631.feature
+++ b/CHANGES/5631.feature
@@ -1,0 +1,1 @@
+Add storage information to the status API. Currently limited to disk space information.
--- a/pulpcore/app/serializers/status.py
+++ b/pulpcore/app/serializers/status.py
@@ -39,6 +39,27 @@ class RedisConnectionSerializer(serializers.Serializer):
     )
 
 
+class StorageSerializer(serializers.Serializer):
+    """
+    Serializer for information about the storage system
+    """
+
+    total = serializers.IntegerField(
+        min_value=0,
+        help_text=_('Total number of bytes')
+    )
+
+    used = serializers.IntegerField(
+        min_value=0,
+        help_text=_('Number of bytes in use')
+    )
+
+    free = serializers.IntegerField(
+        min_value=0,
+        help_text=_('Number of free bytes')
+    )
+
+
 class StatusSerializer(serializers.Serializer):
     """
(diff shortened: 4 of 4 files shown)


pulpcore/tests/functional/api/test_status.py
```diff
@@ -67,15 +77,17 @@ def test_get_authenticated(self):
 
         Verify the response with :meth:`verify_get_response`.
         """
-        self.verify_get_response(self.client.get(STATUS_PATH))
+        with mock.patch('pulpcore.app.views.status.disk_usage', return_value=None):
```
ekohl: This will need some more thought to properly isolate it depending on config.

pulpcore/app/views/status.py
```diff
@@ -15,6 +18,16 @@
 _logger = logging.getLogger(__name__)
 
 
+def disk_usage():
```
bmbouter: Can this either become a private method with the leading `_` or have a docstring added?

pulpcore/tests/functional/api/test_status.py
bmbouter: I think pulp-smash has facilities to run remote commands on the machine to get the number of bytes. Maybe @kersommoura can provide some details.

I would also be ok with just asserting that section is not None since 100% of our tests currently use that storage backend. It could be nearly impossible the grab the exact number of bytes at all times. Also checking within 5% or something would also be fine.

I think going with the "lighter check" option that doesn't verify the number itself, but it's existence is what I recommend. What do you think?

I don't think we can use mock in the functional tests though because running on remote systems won't work.

bmbouter:
@ekohl I'm really excited about this feature. I see it's still in Draft, but I went ahead and left some comments. Feel free to ping me again for a re-review. Thanks!

bmbouter:
@ekohl also lmk if you have any questions or need any help.

pulpcore/tests/functional/api/test_status.py
ekohl: I guess we can verify it has all the keys with some value. Not sure if the schema also verifies that already.

pulpcore/app/views/status.py
ekohl: Good point.

ekohl:
It is green, but I do see a warning:
```
  /home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/jsonschema/validators.py:928: DeprecationWarning: The metaschema specified by $schema was not found. Using the latest draft to validate, but this will raise an error in the future.

    cls = validator_for(schema)
```

Was this already present before my change?

Either way, I think this is ready for review now.

pulpcore/app/views/status.py
```diff
@@ -15,6 +18,16 @@
 _logger = logging.getLogger(__name__)
 
 
+def _disk_usage():
+    if settings.DEFAULT_FILE_STORAGE == 'pulpcore.app.models.storage.FileSystem':
```
ekohl: I've copied this from other places, but it looks the `Storage` class doesn't have a location property. Would this be preferred or a `hasattr(default_storage, 'location')`?

daviddavis: I think checking `settings.DEFAULT_FILE_STORAGE` is better IMO.

daviddavis:
@ekohl this looks great. Would it be possible to also expose `settings.DEFAULT_FILE_STORAGE` in the status api? That way we can show something for non-filesystem installs. I will understand though if this is beyond on the scope of the amount of work you're willing to do.

ekohl:
My idea was that for any supported backend (my understanding is that it's currently local storage or AWS) the storage status is implemented. I have my doubts about sharing internal data. Should storage be transparent to the outside? Since it's easier to add than to remove, I opted not to do this now.

daviddavis:
You're correct that currently we only support local storage or S3. I can understand your hesitation to expose the type of storage. I think that's fine to exclude it.

However, thinking through this a bit more, I'm wondering how we can support other storage types? Our api schema doesn't support polymorphism so we can't show a different set of fields for each storage type.

ekohl:
Correct. So what's the common set? I thought that we will always have a bytes used. We may not have an upper limit in some horizontal scaling solution in which case bytes free is also not available. Currently the schema may not reflect that though.

daviddavis:
I think the easiest solution would be to have a superset of fields that we always display (regardless of what `settings.DEFAULT_FILE_STORAGE` is) and just return None or blank for a field if it's not applicable. I will look to see if we can gather total/used/free from S3 though.

ekohl:
> **Edit:** I don't see a way to get total/free with S3 and from what I know about S3, I am not sure they make sense (ie disk space is pay per use).

Right, that's what I meant with `total` should be `None` which implies `free` is also `None`. However, can we get how much is used? Is it expensive?

Do we want the whole "value" to be `None` (current implementation) or return a dict with empty values?

daviddavis:
> Right, that's what I meant with total should be None which implies free is also None. However, can we get how much is used? Is it expensive?

I think we should just file a follow up issue to do this. I can do it.

> Do we want the whole "value" to be None (current implementation) or return a dict with empty values?

If by empty dict you mean `{"total": None, "free": None, "used": None}`, then I think that's the best option. I don't think the current implementation will work with the bindings unless you set `storage` in `StorageSerializer` to `required=False`. For whatever reason the bindings validate responses so if `storage` is returned as `None`, the bindings will raise an exception.

daviddavis:
Actually, the more I think about it, I think setting `required=False` on storage in StorageSerializer is the better option. Let's go with that and I will file a follow up issue to support S3 storage info in the status api.

daviddavis:
I created https://pulp.plan.io/issues/5651. If you can set `storage` to `required=False`, I good to merge this.

ekohl:
> If you can set `storage` to `required=False`, I good to merge this.

Done. Please do note I'm 100% relying on CI and reviewers because I haven't actually created a dev instance for this.

daviddavis:
I tested it out and it worked well.

bmbouter:
I'm +1 to merge this once green. Thanks @ekohl !

PR #354, opened 2019-10-28
Typed Repositories

dralley:


pulpcore/app/serializers/publication.py
```diff
@@ -26,12 +26,11 @@ class PublicationSerializer(ModelSerializer):
         queryset=models.RepositoryVersion.objects.all(),
         required=False,
     )
-    repository = RelatedField(
+    repository = DetailRelatedField(
```
ipanova: @dralley  this line needs similar change https://github.com/pulp/pulpcore/blob/master/pulpcore/app/serializers/publication.py#L195

pulpcore/app/models/repository.py
```diff
@@ -52,6 +54,55 @@ class Repository(Model):
     class Meta:
         verbose_name_plural = 'repositories'
 
+    def new_version(self, base_version=None):
```
gmbnomis: I like that you moved this method to here. (RepositoryVersion should not modify Repository.)

pulpcore/app/serializers/publication.py
dralley: Good catch -- it looks like we're missing test coverage of this in pulpcore.  I assume the docker plugin hits this?  Should we then have a test for this in pulpcore using the docker plugin?

pulpcore/app/models/repository.py
```diff
@@ -52,6 +54,55 @@ class Repository(Model):
...
+        with transaction.atomic():
+            version = RepositoryVersion(
+                repository=self,
+                number=int(self.last_version) + 1,
+                base_version=base_version)
+            self.last_version = version.number
+            self.save()
+            version.save()
+
+            if base_version:
```
gmbnomis: But I don't like this part. Basically, the `RepositoryVersion(...)` constructor above may create an invalid `RepositoryVersion` instance (i.e. it has the wrong content) which is "salvaged" here by adjusting the content of the version from the outside.

Should we split up the creation of RepositoryVersion into two parts? RepositoryVersion is able to create a RepositoryVersion  with the correct content. And Repository calls the creation method with the correct parameters?

pulpcore/app/util.py
```diff
@@ -55,6 +56,8 @@ def get_view_name_for_model(model_obj, view_action):
     # Import this here to prevent out-of-order plugin discovery
     from pulpcore.app.urls import all_routers
 
+    if isinstance(model_obj, models.MasterModel):
```
gmbnomis: Could you explain why this is needed? The call site you introduced seems to already do a cast. Is there another place in the code that needs this cast?

.travis/test_bindings.py
```diff
@@ -139,7 +139,7 @@ def upload_file_in_chunks(file_path):
 # Monitor the sync task
 created_resources = monitor_task(sync_response.task)
 
-repository_version_1 = repoversions.read(created_resources[0])
+repository_version_1 = filerepoversions.read(created_resources[0])
```
dralley:
```
The task was successful.
  File "../../pulpcore/.travis/test_bindings.py", line 142, in <module>
    repository_version_1 = filerepoversions.read(created_resources[0])
  File "/home/vagrant/devel/pulp-openapi-generator/pulpcore-client/pulpcore/client/pulp_file/api/repositories_file_versions_api.py", line 352, in read
    return self.read_with_http_info(file_repository_version_href, **kwargs)  # noqa: E501
  File "/home/vagrant/devel/pulp-openapi-generator/pulpcore-client/pulpcore/client/pulp_file/api/repositories_file_versions_api.py", line 441, in read_with_http_info
    collection_formats=collection_formats)
  File "/home/vagrant/devel/pulp-openapi-generator/pulpcore-client/pulpcore/client/pulpcore/api_client.py", line 343, in call_api
    _preload_content, _request_timeout, _host)
  File "/home/vagrant/devel/pulp-openapi-generator/pulpcore-client/pulpcore/client/pulpcore/api_client.py", line 182, in __call_api
    return_data = self.deserialize(response_data, response_type)
  File "/home/vagrant/devel/pulp-openapi-generator/pulpcore-client/pulpcore/client/pulpcore/api_client.py", line 254, in deserialize
    return self.__deserialize(data, response_type)
  File "/home/vagrant/devel/pulp-openapi-generator/pulpcore-client/pulpcore/client/pulpcore/api_client.py", line 282, in __deserialize
    klass = getattr(pulpcore.client.pulpcore.models, klass)
AttributeError: module 'pulpcore.client.pulpcore.models' has no attribute 'RepositoryVersion'
```

gmbnomis: @dralley I had a look into this. Further binding fixes to make the PR green are here https://github.com/pulp/pulpcore/pull/373/commits/3e60881c403f7a79e8621360334ef4026cd1b57e

PR #358, opened 2019-10-29
Introduce CONTENT_ORIGIN setting

bmbouter:
This also adds a check that the user specified CONTENT_ORIGIN at
startup. Pulp refuses to start if it is not set with an error message
for the user.

Required PR: https://github.com/PulpQE/pulp-smash/pull/1227
Required PR: https://github.com/pulp/ansible-pulp/pull/185
Required PR: https://github.com/pulp/pulp_file/pull/303

https://pulp.plan.io/issues/5629
re #5629

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -92,7 +92,8 @@ spec:
     password: pulp
     admin_password: pulp
   pulp_settings:
-     content_host: $(hostname):24816
+     content_origin: http://$(hostname):24816
+    
 CRYAML
 
 # Install k3s, lightweight Kubernetes
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -20,6 +20,10 @@ export FUNC_TEST_SCRIPT=$TRAVIS_BUILD_DIR/.travis/func_test_script.sh
 export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
 
 if [ "$TEST" = 'docs' ]; then
+  
+    export PULP_CONTENT_ORIGIN=http://$(hostname):24816
+  
+
   cd docs
   make html
   cd ..
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -10,8 +10,12 @@ https://docs.djangoproject.com/en/1.11/ref/settings/
 
 import os
 from contextlib import suppress
+from gettext import gettext as _
 from importlib import import_module
 from pkg_resources import iter_entry_points
+
+from django.core.exceptions import ImproperlyConfigured
+
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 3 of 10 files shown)


pulpcore/app/wsgi.py
```diff
@@ -7,11 +7,19 @@
...
+from gettext import gettext as _
 import os
 
+from django.conf import settings
 from django.core.wsgi import get_wsgi_application  # noqa
+from django.core.exceptions import ImproperlyConfigured
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "pulpcore.app.settings")
 
+try:
```
bmbouter: I put it here so that only pulp-api would fail, but I realize now that maybe all processes should be guaranteed to have it and stop the user if they don't. @daviddavis wdyt?

daviddavis: +1 to having all services fail

bmbouter: I will move this there.

pulpcore/app/settings.py
```diff
@@ -234,3 +238,9 @@
     ENVVAR_FOR_DYNACONF='PULP_SETTINGS',
 )
 # HERE ENDS DYNACONF EXTENSION LOAD (No more code below this line)
+
+
+try:
```
bmbouter: @rochacbruno here is where I put the check. I tested it and it works, but is this how you would have done it?

bmbouter:
I tried to use [dynaconf validators](https://dynaconf.readthedocs.io/en/latest/guides/validation.html) with the diff below. I'm going submit the code without dynaconf validators for 2 reasons:

1) The error message I think will be confusing to users because most users use Pulp without much awareness they are also using dynaconf.  The error message is: `dynaconf.validator.ValidationError: CONTENT_ORIGIN is required in env DEVELOPMENT`

2) I couldn't get it to work. Maybe @rochacbruno can tell me why and we can adopt them later. I would like to use them.

```
from dynaconf import settings, Validator

settings.validators.register(
    # Ensure some password cannot exist
    Validator('CONTENT_ORIGIN', must_exist=True),
)

settings.validators.validate()
```

.travis/install.sh
```diff
@@ -92,7 +92,8 @@ spec:
     password: pulp
     admin_password: pulp
   pulp_settings:
-     content_host: $(hostname):24816
+     content_origin: http://$(hostname):24816
```
ipanova: if pulp-operator auto sets this, can we remove the setting from travis installation scripts?

template_config.yml
```diff
@@ -21,7 +21,7 @@ plugin_dash_short: pulpcore
 plugin_name: pulpcore
 plugin_snake: pulpcore
 pulp_settings:
-  content_host: $(hostname):24816
+  content_origin: http://$(hostname):24816
```
ipanova: this line can also be removed

.travis/script.sh
```diff
@@ -20,6 +20,10 @@ export FUNC_TEST_SCRIPT=$TRAVIS_BUILD_DIR/.travis/func_test_script.sh
 export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
 
 if [ "$TEST" = 'docs' ]; then
+  
+    export PULP_CONTENT_ORIGIN=http://$(hostname):24816
```
ipanova: same here

daviddavis:
Closing in favor of https://github.com/pulp/pulpcore/pull/370

PR #360, opened 2019-10-30
Make file required on ArtifactSerializer

mdellweg:
When refactoring the upload mechanics away from the artifacts endpoint, this
has been missed.

[noissue]

### DIFF
--- a/pulpcore/app/serializers/content.py
+++ b/pulpcore/app/serializers/content.py
@@ -157,12 +157,11 @@ class ArtifactSerializer(base.ModelSerializer):
     file = serializers.FileField(
         help_text=_("The stored file."),
         allow_empty_file=True,
-        required=False
     )
 
     size = serializers.IntegerField(
         help_text=_("The size of the file in bytes."),
-        required=False
+        required=False,
     )
 
     md5 = serializers.CharField(
--- a/pulpcore/tests/unit/serializers/test_content.py
+++ b/pulpcore/tests/unit/serializers/test_content.py
@@ -1,0 +1,31 @@
+from unittest import TestCase
+
+import mock
+from pulpcore.app.serializers import ArtifactSerializer
+
+
+class TestArtifactSerializer(TestCase):
+
+    def test_validate_file_checksum(self):
+        mock_file = mock.MagicMock(size=42)
+        mock_file.hashers.__getitem__.return_value.hexdigest.return_value = 'asdf'
+
+        data = {'file': mock_file}
+        serializer = ArtifactSerializer(data=data)
+        self.assertTrue(serializer.is_valid())
+        new_data = serializer.validated_data
+        self.assertEqual(new_data, {
+            'file': mock_file,
+            'size': 42,
+            'md5': 'asdf',
+            'sha1': 'asdf',
+            'sha224': 'asdf',
+            'sha256': 'asdf',
+            'sha384': 'asdf',
+            'sha512': 'asdf',
+        })
+
+    def test_emtpy_data(self):
+        data = {}
+        serializer = ArtifactSerializer(data=data)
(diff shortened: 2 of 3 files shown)


pulpcore/tests/unit/serializers/test_content.py
```diff
@@ -0,0 +1,31 @@
+from unittest import TestCase
+
+import mock
+from pulpcore.app.serializers import ArtifactSerializer
+
+
+class TestArtifactSerializer(TestCase):
```
daviddavis: :+1:

pulpcore/app/serializers/content.py
```diff
@@ -157,12 +157,12 @@ class ArtifactSerializer(base.ModelSerializer):
     file = serializers.FileField(
         help_text=_("The stored file."),
         allow_empty_file=True,
-        required=False
+        required=True,
```
daviddavis: I wonder if it's better to not set `required` for consistency with how we do it on other fields. On the other hand, I suppose this is more explicit?

mdellweg: It was set to `False` because there was an alternative source (`upload`) added, together with a validation, that either one was specified. This test has been removed when the upload mechanism went elsewhere, but the required was not reenacted. And i believe it must be `True`, to ensure that the `validate` method can access that field.

mdellweg: Sorry, complete misunderstood you. You meant leave its default value untouched?
Whatever you like more...

mdellweg:
@daviddavis Sorry, i don't see the slightest reason, why this change should break the bindings.
If you want me to fix this, im gonna need help.

daviddavis:
@mdellweg I ran the bindings tests locally and they seemed to pass. I tried to rerun Travis but now it seems that Travis is having degraded service. I can look into it more tomorrow when Travis is hopefully fixed.

daviddavis:
@mdellweg I think this PR needs the latest Travis changes. Can you rebase?

daviddavis:
I debugged the bindings issue. It looks like the signature on the artitfact create method changes with this PR:

**Before**

```ruby
    # Create an artifact
    # @param [Hash] opts the optional parameters
    # @option opts [File] :file The stored file.
    def create(opts = {})
      data, _status_code, _headers = create_with_http_info(opts)
      data
    end
```

**After**

```ruby
    # Create an artifact
    # @param file [File] The stored file.
    # @return [Artifact]
    def create(file, opts = {})
      data, _status_code, _headers = create_with_http_info(file, opts)
      data
    end
```

However, I'm not sure exactly why this happens since I don't see other required parameters as explicit arguments in the ruby bindings?

dkliban:
This makes sense. The openapi-generator-cli makes required parameters into positional arguments. The test bindings scripts need to be updated.

daviddavis:
@dkliban thanks.

@mdellweg this should fix the bindings test:

```diff
diff --git a/.travis/test_bindings.rb b/.travis/test_bindings.rb
index 390388c56..c4a37abd0 100644
--- a/.travis/test_bindings.rb
+++ b/.travis/test_bindings.rb
@@ -122,7 +122,7 @@ repository_version_1 = @repoversions_api.read(created_resources[0])
 
 # Create an artifact from a local file
 file_path = File.join(ENV['TRAVIS_BUILD_DIR'], '.travis/test_bindings.rb')
-artifact = @artifacts_api.create({file: File.new(file_path)})
+artifact = @artifacts_api.create(File.new(file_path))
 
 # Create a FileContent from the artifact
 filecontent_response = @filecontent_api.create('foo.tar.gz', {artifact: artifact.pulp_href})
```

mdellweg:
Let's see.

mdellweg:
Thank you!
@daviddavis @dkliban

PR #364, opened 2019-11-05
Return HTML listing of directories served by content app

### LINKED ISSUE #5378: Add a scheduled cleanup task to keep the tasks table reasonable
We have seen substancial performance degredation on tasks tables that are several million rows long. While this is probably not uncommon in a usual Pulp installation it is a tedious task for the admin to call purge tasks regularly.
We should just schedule it. To give admins the control they are used to, this should be guarded by a setting (maybe down to different tasks states) for how long tasks are supposed to be finished before vacuumed.

Question: Are we confident to add a reasonable time as default, or does this feature need to be off by default?

dkliban:
This patch enables users to browse each distribution. However, the /pulp/content/ endpoint
continues to return a 404.

When a user requests the 'base_url' of a Distribution or any path that ends in '/' inside the
Distribution's 'base_url' path, the pulpcore-content app looks for either a PublishedArtifact
or ContentArtifact with a relative path equal to <requested path>'index.html'. If no such
relative path is found for the RepositoryVersion or Publication, an HTML page is dynamically
generated and returned.

closes #5378
https://pulp.plan.io/issues/5378

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -1,13 +1,14 @@
 import logging
 import mimetypes
 import os
+import re
 from gettext import gettext as _
 
 import django  # noqa otherwise E402: module level not at top of file
 django.setup()  # noqa otherwise E402: module level not at top of file
 
 from aiohttp.client_exceptions import ClientResponseError
-from aiohttp.web import FileResponse, StreamResponse
+from aiohttp.web import FileResponse, StreamResponse, HTTPOk
 from aiohttp.web_exceptions import HTTPForbidden, HTTPFound, HTTPNotFound
 from django.conf import settings
 from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
@@ -21,6 +22,7 @@ from pulpcore.app.models import (
     RepositoryVersion,
 )
 
+from jinja2 import Template
 
 log = logging.getLogger(__name__)
 
@@ -111,7 +113,7 @@ class Handler:
         """
         tree = []
         while True:
-            base = os.path.split(path.strip('/'))[0]
+            base = os.path.split(path)[0]
             if not base:
                 break
             tree.append(base)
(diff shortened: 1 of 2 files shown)


pulpcore/content/handler.py
```diff
@@ -195,6 +197,65 @@ def response_headers(path):
...
+        """
+        Generate HTML with directory listing of the path.
+
+        This method expects either a repo_version or a publication in addition to a path. This
+        method generates HTML directory list of a path inside the repository version or
+        publication.
+
+        Args:
+            repo_version (:class:`~pulpcore.app.models.RepositoryVersion`): The repository version
+            publication (:class:`~pulpcore.app.models.Publication`): Publication
```
daviddavis: You're missing `path` parameter.

pulpcore/content/handler.py
```diff
@@ -195,6 +197,65 @@ def response_headers(path):
...
+        template = Template("""
+        <html>
+        <body>
+        <ul>
+        {% for name in dir_list %}
+        <li><a href="{{ name }}">{{ name }}</a></li>
+        {% endfor %}
+        </ul>
+        </body>
+        </html>
```
daviddavis: Maybe indent this for readability?

pulpcore/content/handler.py
```diff
@@ -195,6 +197,65 @@ def response_headers(path):
...
+                    result = re.match(r'({})([^\/]*)(\/*)'.format(path), ca.relative_path)
+                    directory_list.add('{}{}'.format(result.groups()[1], result.groups()[2]))
+
+        if repo_version:
+            cas = ContentArtifact.objects.filter(
+                content__in=repo_version.content,
+                relative_path__startswith=path)
+            for ca in cas:
+                result = re.match(r'({})([^\/]*)(\/*)'.format(path), ca.relative_path)
+                directory_list.add('{}{}'.format(result.groups()[1], result.groups()[2]))
```
daviddavis: What about DRYing up this code into a nested function?

pulpcore/content/handler.py
dkliban: Will do.

pulpcore/content/handler.py
dkliban: I will

pulpcore/content/handler.py
dkliban: thanks. i'll add it.

PR #365, opened 2019-11-06
Thaw queues on awaiting stuck futures

mdellweg:
fixes #5668

### DIFF
--- a/pulpcore/plugin/stages/api.py
+++ b/pulpcore/plugin/stages/api.py
@@ -109,22 +109,36 @@ class Stage:
         batch = []
         shutdown = False
         no_block = False
+        thaw_queue_event = asyncio.Event()
 
         def add_to_batch(content):
             nonlocal batch
             nonlocal shutdown
             nonlocal no_block
+            nonlocal thaw_queue_event
+
             if content is None:
                 shutdown = True
                 log.debug(_('%(name)s - shutdown.'), {'name': self})
             else:
                 if not content.does_batch:
                     no_block = True
+                content._thaw_queue_event = thaw_queue_event
                 batch.append(content)
 
+        get_listener = asyncio.ensure_future(self._in_q.get())
+        thaw_event_listener = asyncio.ensure_future(thaw_queue_event.wait())
         while not shutdown:
-            content = await self._in_q.get()
-            add_to_batch(content)
+            done, pending = await asyncio.wait(
+                [thaw_event_listener, get_listener], return_when=asyncio.FIRST_COMPLETED
+            )
+            if thaw_event_listener in done:
+                thaw_event_listener = asyncio.ensure_future(thaw_queue_event.wait())
+                no_block = True
+            if get_listener in done:
+                content = await get_listener
+                add_to_batch(content)
+                get_listener = asyncio.ensure_future(self._in_q.get())
             while not shutdown:
                 try:
                     content = self._in_q.get_nowait()
@@ -140,9 +154,14 @@ class Stage:
                         'name': self,
                         'length': len(batch),
                     })
+                for content in batch:
+                    content._thaw_queue_event = None
+                thaw_queue_event.clear()
                 yield batch
                 batch = []
                 no_block = False
+        thaw_event_listener.cancel()
+        get_listener.cancel()
 
     async def put(self, item):
         """
--- a/pulpcore/plugin/stages/content_stages.py
+++ b/pulpcore/plugin/stages/content_stages.py
@@ -160,15 +160,20 @@ class ResolveContentFutures(Stage):
     different content type `Bar`. Consider this code in FirstStage::
 
         # Create d_content and d_artifact for a `foo_a`
-        foo_a = DeclarativeContent(..., does_batch=False)
-        foo_a_future = foo_a.get_or_create_future()  # This is awaitable
+        foo_a = DeclarativeContent(...)
+        # Send it in the pipeline
+        await self.put(foo_a)
 
         ...
 
-        foo_a_content = await foo_a_future  # awaits until the foo_a reaches this stage
+        foo_a_content = await foo_a.resolution()  # awaits until the foo_a reaches this stage
 
     This creates a "looping" pattern, of sorts, where downloaded content at the end of the pipeline
     can introduce new additional to-be-downloaded content at the beginning of the pipeline.
+    On the other hand, it can impose a substantial performance decrement of batching content in the
+    earlier stages.
+    As a rule of thumb, sending more items into the pipeline first and awaiting their resolution
+    later is better.
     """
 
     async def run(self):
@@ -179,6 +184,5 @@ class ResolveContentFutures(Stage):
             The coroutine for this stage.
         """
         async for d_content in self.items():
-            if d_content.future is not None:
-                d_content.future.set_result(d_content.content)
+            d_content.resolve()
             await self.put(d_content)
--- a/pulpcore/plugin/stages/models.py
+++ b/pulpcore/plugin/stages/models.py
@@ -102,43 +102,58 @@ class DeclarativeContent:
         d_artifacts (list): A list of zero or more
             :class:`~pulpcore.plugin.stages.DeclarativeArtifact` objects associated with `content`.
         extra_data (dict): A dictionary available for additional data to be stored in.
-        does_batch (bool): If `False`, prevent batching mechanism to block this item.
-            Defaults to `True`.
-        future (:class:`~asyncio.Future`): A future that gets resolved to the
-            :class:`~pulpcore.plugin.models.Content` in the
-            :class:`~pulpcore.plugin.stages.ResolveContentFutures` stage. See the
-            :class:`~pulpcore.plugin.stages.ResolveContentFutures` stage for example usage.
 
     Raises:
         ValueError: If `content` is not specified.
     """
 
-    __slots__ = ('content', 'd_artifacts', 'extra_data', 'does_batch', 'future')
+    __slots__ = (
+        'content',
+        'd_artifacts',
+        'extra_data',
+        '_future',
+        '_thaw_queue_event',
+        '_resolved',
+    )
 
-    def __init__(self, content=None, d_artifacts=None, extra_data=None, does_batch=True):
+    def __init__(self, content=None, d_artifacts=None, extra_data=None):
         if not content:
             raise ValueError(_("DeclarativeContent must have a 'content'"))
         self.content = content
         self.d_artifacts = d_artifacts or []
         self.extra_data = extra_data or {}
-        self.does_batch = does_batch
-        self.future = None
+        self._future = None
+        self._thaw_queue_event = None
+        self._resolved = False
 
-    def get_or_create_future(self):
+    @property
+    def does_batch(self):
+        """Whether this content is being awaited on and must therefore not wait forever in batches.
+        When overwritten in subclasses, a `True` value must never be turned into `False`.
         """
-        Return the existing or a new future.
+        return not self._resolved and self._future is None
 
-        If you rely on this future in a the course of the pipeline, consider clearing the
-        `does_batch` attribute to prevent deadlocks.
-        See the :class:`~pulpcore.plugin.stages.ResolveContentFutures` stage for example usage.
+    async def resolution(self):
+        """Coroutine that waits for the content to be saved to database.
+        Returns the content unit."""
+        if self._resolved:
+            # Already resolved ~> shortcut
+            return self.content
+        if self._future is None:
+            # We do not yet have a future
+            self._future = asyncio.get_event_loop().create_future()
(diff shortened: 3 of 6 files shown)


bmbouter:
@mdellweg can you give me more insight into this change? It's a plugin writer facing feature so we also need a pulp.plan.io entry for it.

mdellweg:
@gmbnomis care to take a look?

CHANGES/5668.feature
```diff
@@ -0,0 +1 @@
+Allow get_or_create_future without deactivating does_batch on DeclarativeContent from the beginning.
```
bmbouter: This release note is for the plugin API, can it be moved to the plugin API folder in CHANGES?

pulpcore/plugin/stages/models.py
```diff
@@ -86,6 +86,25 @@ def __init__(self, artifact=None, url=None, relative_path=None, remote=None, ext
         return download_result
 
 
+class DCFuture(asyncio.Future):
```
bmbouter: This will need an autodoc entry so it shows in the docs also somewhere on this page https://github.com/pulp/pulpcore/blob/master/docs/plugins/api-reference/stages.rst

pulpcore/plugin/stages/models.py
```diff
@@ -113,7 +132,7 @@ class DeclarativeContent:
         ValueError: If `content` is not specified.
     """
 
-    __slots__ = ('content', 'd_artifacts', 'extra_data', 'does_batch', 'future')
+    __slots__ = ('content', 'd_artifacts', 'extra_data', 'does_batch', 'future', 'thaw_queue_event')
```
bmbouter: The `future` and the `thaw_queue_event` seem very similar. I'd like to think more about how we can provide this value to plugin writers without adding more facilities they need to use. Part of the concern is that if the plugin writer needs to take extra care to use this facility, how is that different than the plugin writer taking extra care to clear the `does_batch` attribute like we document [here](https://github.com/pulp/pulpcore/blob/62e6a3b280c1240b0e495bed2cd460a8bb4b5ab4/pulpcore/plugin/stages/models.py#L132).

bmbouter:
@mdellweg how would you feel in us including this in 3.1? I see the benefits so that is clear, but I'm concerned about how we're offering them. I want to work to find a way to bring this benefit to plugin writers without adding the the options/facilities they need to think about.

pulpcore/plugin/stages/models.py
gmbnomis: This class looks like a mixture between a proxy to a future and a subclassed  future. But I don't get why. Wouldn't the following work?

```python
class DCFuture(asyncio.Future):
    """A special future, that can thaw the queue it is stuck in."""
    __slots__ = ["d_content"]

    def __init__(self, d_content):
        self.d_content = d_content

    def __await__(self):
        self.d_content.does_batch = False
        if self.d_content.thaw_queue_event:
            self.d_content.thaw_queue_event.set()
        return super().__await__()
```

pulpcore/plugin/stages/models.py
gmbnomis: I wonder whether `thaw_queue_event` needs to be stored in content at all. If I understand your PR correctly, we need `thaw_queue_event` only if there is a future. As we use our own future type here anyway, couldn't we store it in the future?

bmbouter:
I'm branching 3.0 now. If we are able to resolve this before the actual GA I'd be OK cherry picking it since it's a) additive only and b) only affects plugin writers. Let's continue the collaboration!

pulpcore/plugin/stages/models.py
mdellweg: That is actually something we need to discuss, because as i see it, this is inner plumbing. The user does not even need to know about that class.

pulpcore/plugin/stages/models.py
mdellweg: @bmbouter The way this change works, the user isn't supposed to do _more_ than before. The only thing that is new, is that he can omit to set `does_batch=False`. So we can even safely remove that warning, that using futures can jam the queue.
@gmbnomis That is a matter of perspective, but this version allows to create the future, even after the d_content has been injected into the queue. Also just setting it might not even be slower than looking if there is a future and then conditionally set it.

pulpcore/plugin/stages/models.py
mdellweg: Right. I wanted to make it a subclass (so that is _is_ a future). But then i didn't see a way to inject it into:
`self.future = asyncio.get_event_loop().create_future()`
So it became a proxy, and i forgot about the inheritance. Thinking about this now, we can probably make the DC the proxy itself.

mdellweg: > Thinking about this now, we can probably make the DC the proxy itself.

This would turn the DeclarativeContent into an awaitable future-like object itself. Though i like the concept, it would be backwards incompatible, and rather involved. I am not sure it's worth persuing it today.

gmbnomis: I even like the proper proxy you changed it into now better. The async trigger mechanism is an optional feature of DC, but not its main feature.

pulpcore/plugin/stages/models.py
gmbnomis: @mdellweg I thought about this difference, but injecting the `d_content` into the queue and adding the future later is a bug IMO. The queue operation should be async and we don't know when it comes back. The DC may have passed the entire pipeline before one can actually attach a future to it.

My argument was not about speed (which is probably in the same ball park), but about isolation. If the implementation of the "thaw queue" feature can be self-contained in DCFuture, I think we should do that.

pulpcore/plugin/stages/models.py
mdellweg: I've always seen the DC as some kind of promised future, so i thinks it feels kind of natural to await them. As a result of a discussion with @bmbouter i prepared another (breaking) solution, where the DC's can be awaited directly.
https://github.com/mdellweg/pulpcore/tree/awaitable_dc

gmbnomis: Oh, we are in "we can have a nicer (but breaking) API" land? 

It's great that your change gets rid of the exposed Future, since Futures are considered low-level objects that should not be exposed in user level APIs (https://docs.python.org/3/library/asyncio-future.html#future-object).

But, as said, let's please not await DC directly:

1. It is a pretty uncommon thing to do. The basic building block of an async API seems to be the coroutine, not an awaitable object. See for example the high-level APIs offered by asyncio (https://docs.python.org/3/library/asyncio.html). None of them seem to offer a directly awaitable object.
2. It is more explicit, we could e.g. have `async def wait_for_resolution(self)` which is clearer IMHO than just awaiting the DC (and can have a nice docstring)
3. It allows us to enhance the API later, we could e.g. add a `async def wait_for_download(self)` to await an earlier event. Or we can add optional parameters if the need arises.

wdyt?

mdellweg: > Oh, we are in "we can have a nicer (but breaking) API" land?

Since it was rescheduled to 3.1 that is at least an option. And i am throwing ideas here (which is why i did not just replace this branch).
If what you are saying is you liked to use it like
```python
await self.put(d_content)
content = await d_content.resolution()
```
instead of 
```python
await self.put(d_content)
content = await d_content
```
or the current version
```python
fut = d_content.get_or_create_future()
await self.put(d_content)
content = await fut
```
i think, i can have a working example in no time.

mdellweg:
@gmbnomis is this what you expected to see?
Sure documentation still needs to be adjusted...

pulpcore/plugin/stages/models.py
```diff
@@ -113,32 +113,36 @@ class DeclarativeContent:
         ValueError: If `content` is not specified.
     """
 
-    __slots__ = ('content', 'd_artifacts', 'extra_data', 'does_batch', 'future')
+    __slots__ = ('content', 'd_artifacts', 'extra_data', '_future', '_thaw_queue_event', '_resolved')
 
-    def __init__(self, content=None, d_artifacts=None, extra_data=None, does_batch=True):
```
bmbouter: @ipanova does docker use does_batch?

pulpcore/plugin/stages/api.py
```diff
@@ -109,22 +109,37 @@ class MyStage(Stage):
...
                 batch.append(content)
 
         while not shutdown:
-            content = await self._in_q.get()
-            add_to_batch(content)
+            done, pending = await asyncio.wait(
+                [thaw_event_listener, get_listener], return_when=asyncio.FIRST_COMPLETED
+            )
+            if thaw_event_listener in done:
+                await thaw_event_listener
```
bmbouter: Like we mentioned on the call, I don't see how awaiting here productive since `thaw_event_listener` is already in `done`.

pulpcore/plugin/stages/api.py
```diff
@@ -109,22 +109,37 @@ class MyStage(Stage):
...
             nonlocal shutdown
             nonlocal no_block
+            nonlocal thaw_queue_event
+
             if content is None:
                 shutdown = True
                 log.debug(_('%(name)s - shutdown.'), {'name': self})
             else:
-                if not content.does_batch:
+                if not content._resolved and content._future is not None:
```
simon-baatz: I think the stage should not access `_resolved` and `_future` directly (this is different from `_thaw_queue_event` which iserves as the communication means between Stage and DC).

A `does_batch` property in DC would be a nice way to keep this in DC I guess. A plugin writer could even derive from DC and use a different mechanism to decide what to batch.

gmbnomis:
> @gmbnomis is this what you expected to see?

Yes, I like the interface much more than the old one.

pulpcore/plugin/stages/api.py
mdellweg: Ah, i see now, where this is going. But i do not want to leave it as a variable on the DC, because we cannot just clear it after resolving without surprise. So i'm going to make it a property. A derived DC can then alter the logic to it's need (for example wait only til artifacts are ready).

mdellweg:
@gmbnomis, @bmbouter ^

pulpcore/plugin/stages/models.py
```diff
@@ -102,43 +102,54 @@ class DeclarativeContent:
...
-            # If on 3.7, we could preferrably use get_running_loop()
-            self.future = asyncio.get_event_loop().create_future()
-        return self.future
+        self._future = None
+        self._thaw_queue_event = None
+        self._resolved = False
+
+    @property
+    def does_batch(self):
+        return not self._resolved and self._future is None
```
gmbnomis: Should we add a note here that an overridden method must never 'overrule' the original implementation (i.e. never turn a "does not batch" into an "does batch")

pulpcore/plugin/stages/content_stages.py
```diff
@@ -160,12 +160,13 @@ class ResolveContentFutures(Stage):
...
-        foo_a = DeclarativeContent(..., does_batch=False)
-        foo_a_future = foo_a.get_or_create_future()  # This is awaitable
+        foo_a = DeclarativeContent(...)
+        # Send it in the pipeline
+        await self.put(foo_a)
 
         ...
 
-        foo_a_content = await foo_a_future  # awaits until the foo_a reaches this stage
+        foo_a_content = await foo_a.resolution()  # awaits until the foo_a reaches this stage
```
gmbnomis: Should we mention somewhere that, as a general rule, the pipeline will work more efficient if waiting for resolution is done as late as possible? (that's a terrible wording, but hopefully you understand what I am trying to say).

gmbnomis:
> @gmbnomis, @bmbouter ^

As said, I like the new interface much better than the old one. It's @bmbouter's call whether this backwards incompatible change is ok at this point.

Regarding the "thaw queue" feature, I am still struggling with the added complexity (given that the benefit will only be seen in very specific situations and if the plugin writer knows what she/he is doing). OTOH, this may be the price we have to pay to ease the use of pipeline feedback loops without risking hangs all the time.

pulpcore/plugin/stages/content_stages.py
mdellweg: I know exactly what you mean, and i will try to write something.

bmbouter:
@mdellweg Travis is failing, maybe try rebasing?

I'm ok adding it at this point in time because we are still going to have breaking changes in the plugin API for a few Y releases. When merged this would just go into 3.1 and not backported to 3.0.

This overall seems safer so I'm inclined to accept it. Last call for blocking concerns.

mdellweg:
@bmbouter rebasing helped.

bmbouter:
I want to do a final review on this soon. Sorry for the delay

bmbouter:
@mdellweg to complete this work please send an email to pulp-dev identifying this as a breaking change with the subject starting with `[breaking]`. All plugin API breaking changes need to follow that policy once merged.

PR #366, opened 2019-11-06
Documented a gotcha with the default ACL used by django-storages

joeydumont:
[noissue]

Documented the fact that when the `BlockPublicAcls` option is set either at the bucket or account level, the default value for `AWS_DEFAULT_ACL` of `public-read` will result in AccessDenied errors. To fix it, simply set this setting to None.


joeydumont:
[noissue]

PR #367, opened 2019-11-06
pulpcore-content return  HTML list of avaiable distributions

dkliban:
This patch enables the user to browse :24816/pulp/content/ to see the list of available
distributions.

closes: #5397
https://pulp.plan.io/issues/5397

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -85,6 +85,23 @@ class Handler:
 
     distribution_model = None
 
+    async def list_distributions(self, request):
+        """
+        The handler for an HTML listing all distributions
+
+        Args:
+            request (:class:`aiohttp.web.request`): The request from the client.
+
+        Returns:
+            :class:`aiohttp.web.HTTPOk`: The response back to the client.
+        """
+        if self.distribution_model is None:
+            distributions = BaseDistribution.objects.only("base_path").all()
+        else:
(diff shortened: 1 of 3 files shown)


pulpcore/content/handler.py
```diff
@@ -197,6 +214,31 @@ def response_headers(path):
...
+        Returns:
+            String representing HTML of the directory listing.
+        """
+        template = Template("""
+        <!DOCTYPE html>
+        <html>
+            <body>
+                <ul>
+                {% for name in dir_list %}
+                    <li><a href="{{ name }}">{{ name }}</a></li>
```
gmbnomis: I don't know jinja2 in detail. Does this need to be HTML escaped?

dkliban: I don't think so, but what is your concern?

gmbnomis: file names containing characters like `"`, `>`, `&` which have special meaning in HTML and would not display correctly/render invalid HTML.

daviddavis: I think `{{ name|e }}` should probably work.

https://jinja.palletsprojects.com/en/2.10.x/templates/#html-escaping

dkliban: thanks @gmbnomis and @daviddavis ... i'll update the PR.

PR #369, opened 2019-11-07
Add Repository.finalize_new_version

bmbouter:
This PR adds a central place for plugin writers to put validation and 
content modification code.

* `RepositoryVersion.__exit__` calls `finalize_new_version`
* Moves the `/modify/` endpoint to `pulpcore.plugin.actions` as a mixin
  named `ModifyRepositoryActionMixin`.

* Renames `Repository.repo_key` to `Repository.repo_key_fields`.
* Remove the `RemoveDuplicates` stage for plugin writers to instead use 
  `Repository.finalize_new_version`
* Remove the implicit usage of `RemoveDuplicates` from
 `RepositoryVerison.add_content` and `RepositoryVersion.remove_content`.
* Make the `RemoveDuplicates` available as
  `pulpcore.plugin.repo_version_utils.remove_duplicates`.

As an unrelated change, it also replaces the the `>>>` python
codeblocks in various docblocks with python codeblock using `::` and 
indention.

Required PR: https://github.com/pulp/pulp_file/pull/307

https://pulp.plan.io/issues/3541
re #3541

### DIFF
--- a/CHANGES/plugin_api/3541.removal
+++ b/CHANGES/plugin_api/3541.removal
@@ -1,0 +1,4 @@
+Renamed the Content.repo_key to be Content.repo_unit_key. Also the calling of `RemoveDuplicates`
+no longer happens in `RepositoryVersion.add_content` and instead is intended for plugins to call
+from `Repository.finalize_new_version(new_version)`. Also the `pulpcore.plugin.RemoveDuplicates`
+Stage was removed.
--- a/docs/plugins/plugin-writer/concepts/sync_pipeline/sync_pipeline.rst
+++ b/docs/plugins/plugin-writer/concepts/sync_pipeline/sync_pipeline.rst
@@ -36,8 +36,12 @@ following order:
    3. :class:`pulpcore.plugin.stages.ArtifactDownloader`
    4. :class:`pulpcore.plugin.stages.ArtifactSaver`
    5. :class:`pulpcore.plugin.stages.ContentSaver`
-   6. :class:`pulpcore.plugin.stages.RemoveDuplicates`
+   6. :class:`pulpcore.plugin.stages.RemoteArtifactSaver`
    7. :class:`pulpcore.plugin.stages.ResolveContentFutures`
+   8. :class:`pulpcore.plugin.stages.ContentAssociation`
+
+If the `mirror=True` optional parameter is passed to `DeclarativeVersion` the pipeline also runs
+:class:`pulpcore.plugin.stages.ContentUnassociation` at the end.
 
 On-demand synchronizing
 -----------------------
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -244,7 +244,7 @@ class Content(MasterModel, QueryMixin):
         _artifacts (models.ManyToManyField): Artifacts related to Content through ContentArtifact
     """
     TYPE = 'content'
-    repo_key = ()
+    repo_key_fields = ()  # Used by pulpcore.plugin.repo_version_utils.remove_duplicates
 
     _artifacts = models.ManyToManyField(Artifact, through='ContentArtifact')
 
@@ -287,10 +287,11 @@ class Content(MasterModel, QueryMixin):
         Plugin writers are expected to override this method with an implementation for a specific
         content type.
 
-        For example:
-            >>> if path.isabs(relative_path):
-            >>>     raise ValueError(_("Relative path can't start with '/'."))
-            >>> return FileContent(relative_path=relative_path, digest=artifact.sha256)
+        For example::
+
+            if path.isabs(relative_path):
+                raise ValueError(_("Relative path can't start with '/'."))
+            return FileContent(relative_path=relative_path, digest=artifact.sha256)
 
         Args:
             artifact (:class:`~pulpcore.plugin.models.Artifact`): An instance of an Artifact
--- a/pulpcore/plugin/actions.py
+++ b/pulpcore/plugin/actions.py
@@ -1,0 +1,59 @@
+from drf_yasg.utils import swagger_auto_schema
+from rest_framework.decorators import action
+
+from pulpcore.app import tasks
+from pulpcore.app.models import Content, Repository, RepositoryVersion
+from pulpcore.app.response import OperationPostponedResponse
+from pulpcore.app.serializers import (
+    AsyncOperationResponseSerializer,
+    RepositoryAddRemoveContentSerializer
+)
+from pulpcore.tasking.tasks import enqueue_with_reservation
+
+
+class ModifyRepositoryActionMixin:
+
+    @swagger_auto_schema(
+        operation_description="Trigger an asynchronous task to create a new repository version.",
+        responses={202: AsyncOperationResponseSerializer}
+    )
+    @action(detail=True, methods=["post"], serializer_class=RepositoryAddRemoveContentSerializer)
+    def modify(self, request, pk):
+        """
+        Queues a task that creates a new RepositoryVersion by adding and removing content units
+        """
+        add_content_units = []
+        remove_content_units = []
+        repository = Repository.objects.get(pk=pk)
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        if 'base_version' in request.data:
+            base_version_pk = self.get_resource(request.data['base_version'], RepositoryVersion).pk
+        else:
+            base_version_pk = None
+
+        if 'add_content_units' in request.data:
+            for url in request.data['add_content_units']:
+                content = self.get_resource(url, Content)
+                add_content_units.append(content.pk)
+
+        if 'remove_content_units' in request.data:
+            for url in request.data['remove_content_units']:
+                if url == '*':
+                    remove_content_units = [url]
+                    break
+                else:
+                    content = self.get_resource(url, Content)
+                    remove_content_units.append(content.pk)
+
+        result = enqueue_with_reservation(
+            tasks.repository.add_and_remove, [repository],
+            kwargs={
+                'repository_pk': pk,
+                'base_version_pk': base_version_pk,
+                'add_content_units': add_content_units,
+                'remove_content_units': remove_content_units
+            }
+        )
+        return OperationPostponedResponse(result, request)
--- a/pulpcore/plugin/repo_version_utils.py
+++ b/pulpcore/plugin/repo_version_utils.py
@@ -1,0 +1,47 @@
+from collections import defaultdict
+from gettext import gettext as _
+import logging
+
+from django.db.models import Q
+
+
+_logger = logging.getLogger(__name__)
+
+
+def remove_duplicates(repository_version):
+    """
+    Inspect content additions in the `RepositoryVersion` and replace repository duplicates.
+
+    Some content can have two instances A and B which are unique, but cannot both exist together in
+    one repository. For example, pulp_file's content has `relative_path` for that file within the
+    repository.
+
+    Any content newly added to the :class:`~pulpcore.plugin.models.RepositoryVersion` is checked
+    against existing content in the :class:`~pulpcore.plugin.models.RepositoryVersion` with newer
+    "repository duplicates" replace existing "repository duplicates". Each Content model can define
+    a `repo_key_fields` attribute with the field names to be compared. If all `repo_key_fields`
+    contain the same value for two content units, they are considered "repository duplicates".
+
+    After instantiating `RemoveDuplicates` call it with the `run()` method and pass in the
+    :class:`~pulpcore.plugin.models.RepositoryVersion` to be checked and possibly modified as a
+    parameter to `run()`.
+
+    Args:
+        repository_version: The :class:`~pulpcore.plugin.models.RepositoryVersion` to be checked
+            and possibly modified.
+    """
+    query_for_repo_duplicates_by_type = defaultdict(lambda: Q())
+    for item in repository_version.added():
+        detail_item = item.cast()
+        if detail_item.repo_key_fields == ():
+            continue
+        unit_q_dict = {
+            field: getattr(detail_item, field) for field in detail_item.repo_key_fields
+        }
+        item_query = Q(**unit_q_dict) & ~Q(pk=detail_item.pk)
+        query_for_repo_duplicates_by_type[detail_item._meta.model] |= item_query
+
+    for model in query_for_repo_duplicates_by_type:
+        _logger.debug(_("Removing duplicates for type: {}".format(model)))
+        qs = model.objects.filter(query_for_repo_duplicates_by_type[model])
+        repository_version.remove_content(qs)
--- a/pulpcore/plugin/stages/__init__.py
+++ b/pulpcore/plugin/stages/__init__.py
@@ -8,7 +8,6 @@ from .artifact_stages import (  # noqa
 from .association_stages import (  # noqa
     ContentAssociation,
     ContentUnassociation,
-    RemoveDuplicates
 )
 from .content_stages import ContentSaver, QueryExistingContents, ResolveContentFutures  # noqa
 from .declarative_version import DeclarativeVersion  # noqa
--- a/pulpcore/plugin/stages/declarative_version.py
+++ b/pulpcore/plugin/stages/declarative_version.py
@@ -9,20 +9,20 @@ from .artifact_stages import (
     QueryExistingArtifacts,
     RemoteArtifactSaver,
 )
-from .association_stages import ContentAssociation, ContentUnassociation, RemoveDuplicates
+from .association_stages import ContentAssociation, ContentUnassociation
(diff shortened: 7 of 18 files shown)


pulpcore/plugin/stages/declarative_version.py
```diff
@@ -85,17 +88,16 @@ def __init__(self, first_stage, repository, mirror=False, remove_duplicates=None
         3. Call the :meth:`~pulpcore.plugin.stages.DeclarativeVersion.create` method on your
            :class:`~pulpcore.plugin.stages.DeclarativeVersion` instance
 
-        Here is an example:
+        Here is an example::
 
-        >>> first_stage = MyFirstStage(remote)
-        >>> DeclarativeVersion(first_stage, repository).create()
+            first_stage = MyFirstStage(remote)
+            DeclarativeVersion(first_stage, repository).create()
```
dralley: Should this be ```DeclarativeVersion(first_stage, repository_version...```?

pulpcore/plugin/stages/declarative_version.py
```diff
@@ -85,17 +88,16 @@ def __init__(self, first_stage, repository, mirror=False, remove_duplicates=None
...
-        Example using remove_duplicates:
+        Example using remove_duplicates::
 
-        # This will enforce that within a repository version, `FileContent.relative_path` is
-        # unique.
-        >>> remove_dupes = [{'model': FileContent, 'field_names': ['relative_path']}]
-        >>> DeclarativeVersion(first_stage, repository, remove_duplicates=remove_dupes).create()
+            # This enforces that within a repository version, `FileContent.relative_path` is unique.
+            remove_dupes = [{'model': FileContent, 'field_names': ['relative_path']}]
+            DeclarativeVersion(first_stage, repository, remove_duplicates=remove_dupes).create()
```
dralley: Ditto

pulpcore/plugin/stages/declarative_version.py
bmbouter: yes it should be. Fixed in next push. ty.

pulpcore/plugin/stages/declarative_version.py
bmbouter: fixed in next push, ty.

pulpcore/plugin/stages/declarative_version.py
```diff
@@ -15,65 +15,69 @@
...
            :class:`~pulpcore.plugin.stages.DeclarativeContent` with
            :class:`~pulpcore.plugin.stages.ResolveContentFutures`
-        10. Remove duplicate content in the repository version if `remove_duplicates` is given by
-            :class:`~pulpcore.plugin.stages.RemoveDuplicates`
-        11. Associate all content units with the new
-            :class:`~pulpcore.plugin.models.RepositoryVersion` with
-            :class:`~pulpcore.plugin.stages.ContentAssociation`
-        12. Unassociate any content units not declared in the stream (only when mirror=True)
+        9. Remove duplicate content in the repository version if `remove_duplicates` is given by
+           :class:`~pulpcore.plugin.stages.RemoveDuplicates`
```
goosemania: I think it's no longer a Stage.
Btw, should it be mentioned in the changelog that it's no longer a stage?

bmbouter: Thank you, I realize now that besides this documentation line I need to remove the stage itself!

bmbouter: pushed

pulpcore/plugin/actions.py
```diff
@@ -0,0 +1,76 @@
...
+
+        base_version_pk = self._get_base_version_pk(request)
+        add_content_units = self._handle_add_content_units(request)
+        remove_content_units = self._handle_remove_content_units(request)
+        result = self._dispatch_task(repository, base_version_pk, add_content_units,
+                                     remove_content_units)
+
+        return OperationPostponedResponse(result, request)
+
+    def _get_base_version_pk(self, request):
```
bmbouter: I'll make these public methods with docstrings if we want to keep this implementation before merging.

pulpcore/plugin/actions.py
```diff
@@ -0,0 +1,76 @@
...
+from pulpcore.app.models import Content, Repository, RepositoryVersion
+from pulpcore.app.response import OperationPostponedResponse
+from pulpcore.app.serializers import (
+    AsyncOperationResponseSerializer,
+    RepositoryAddRemoveContentSerializer
+)
+from pulpcore.tasking.tasks import enqueue_with_reservation
+
+
+class ModifyActionMixin:
```
dralley: Do you think we should name this ```ModifyRepositoryActionMixin``` just to add clarity that it works (only) with repository viewsets?

dralley:
I'm not sure I understand the name "vadifier", but otherwise, I really like this change.

docs/plugins/plugin-writer/concepts/sync_pipeline/sync_pipeline.rst
```diff
@@ -36,8 +36,12 @@ following order:
    3. :class:`pulpcore.plugin.stages.ArtifactDownloader`
    4. :class:`pulpcore.plugin.stages.ArtifactSaver`
    5. :class:`pulpcore.plugin.stages.ContentSaver`
-   6. :class:`pulpcore.plugin.stages.RemoveDuplicates`
+   6. :class:`pulpcore.plugin.stages.RemoteArtifactSaver`
```
mdellweg: :+1:

pulpcore/plugin/repo_version_utils.py
```diff
@@ -0,0 +1,52 @@
...
+from gettext import gettext as _
+import logging
+
+from django.db.models import Q
+
+
+_logger = logging.getLogger(__name__)
+
+
+class RemoveDuplicates:
```
mdellweg: What is the reason, this is wrapped as a task? I do not see any state that is saved.

pulpcore/plugin/stages/__init__.py
```diff
@@ -7,8 +7,7 @@
 )
 from .association_stages import (  # noqa
     ContentAssociation,
-    ContentUnassociation,
-    RemoveDuplicates
+    ContentUnassociation
```
mdellweg: black would like to keep the comma.

mdellweg:
For calling the so called vadifier, it looks to me that a vadify method on the repository class should be enough to keep the latter clean. At least if we do not need the ability to pass different validation rules based on some parameters to the sync call, but allowing that calls for trouble anyway imho.

pulpcore/plugin/repo_version_utils.py
bmbouter: Do you mean s/task/class/ ? The thinking is that a plugin writer may want to subclass and customize it. Should I make it a function instead and then a subclasser could call it just the same? Or should I make it a static method? What do you think?

bmbouter:
@mdellweg yes @gmbnomis also had a similar observation/suggestion regarding putting the vadifier as a repository class attribute so the subclasser can define it in one place. That makes sense. I believe we do have use cases where user provided params will need to configure these, but we don't have any "right now" so I'm a little unsure. Here's what I was thinking at least:

The one I keep thinking of is: "As a user uploading, I can retain the previous N versions of an rpm for downgrade purposes, where N is supplied at upload time". This is a bit contrived because if someone did want a feature like that they probably want it all the time and not specified call by call. That's the idea though.

One option is to ignore that use case for now until we actually have one. If we do that, I agree it would be best to have a class attribute the subclasser could define to set the vadifier.

gmbnomis:
> @mdellweg yes @gmbnomis also had a similar observation/suggestion regarding putting the vadifier as a repository class attribute so the subclasser can define it in one place. That makes sense. I believe we do have use cases where user provided params will need to configure these, but we don't have any "right now" so I'm a little unsure. Here's what I was thinking at least:
> 
> The one I keep thinking of is: "As a user uploading, I can retain the previous N versions of an rpm for downgrade purposes, where N is supplied at upload time". This is a bit contrived because if someone did want a feature like that they probably want it all the time and not specified call by call. That's the idea though.

What comes to mind:
- Add "recursive" flag for the modify operation (which will recursively add dependencies), but not for sync. 
- As seen with the recursive flag, vadification may be different in different operations. We had discussions to automatically fixup repo versions (add or remove content based on other content). This might not be necessary for sync (we have full upstream metadata).

Because vadifiers also include modification, I think they don't belong to the repo version. They belong to the respective operation creating a repo version.  The vadifiers are there to allow to reuse common parts (which may be 100% in many cases).

bmbouter:
@gmbnomis and @mdellweg I talked w/ @dralley, @daviddavis, @dkliban, and @goosemania; here is what we came up with:

* We don't have a use case where the user wants to provide configuration of the vadifier currently. Even if we end up with one the plugin writer could still work around it because they are the owner of the viewset now.
* Therefore moving the vadifier to a central place that `SingleContentArtifactUploadSerializer.create`, `DeclarativeVersion.run`, and `modify` can call is good to avoid passing it through the tasking system and into the core machinery with optional params.
* We observe that it would be useful for the `vodifier` to have access to the Repository instance. For example a "retain N" feature for pulp_rpm. The user won't want to specify that call-by-call, but instead set it on a Repository itself as, e.g. `retain=2`, which now they can do with typed repos. Then having the vadifier have access to that Repository instance would be good. Thus making it an instance method is useful because it has access to self.
* I got feedback that the vadifier name was confusing. We're planning to name this `def new_version_finalize(self, new_version)`.
* This would be useful for various other things also so it creates more value for the plugin writer, e.g. notifying a third party service upon finalization maybe?

I'm going to add this in as an additional commit to demonstrate the idea with the hopes of merging it. What do you think of this? Will this work your your needs?

daviddavis:
Does `finalize_new_version` make sense over `new_version_finalize`?

bmbouter:
Yes I'll use `finalize_new_version` instead.

mdellweg:
If i read correctly, the suggestion now is to make the validitator an instance method of repository, and any possible parameter to it a field on said repository. This is exactly, what i thought it to be, as i see this method to put some common (for all operations) constraints on the resulting repository versions.
And for the example of recursively adding dependencies, i would go as far as stating, that if one wants that (repo cosure) on a specific repository, than that validification should also be used on sync.
We would even need to think about the possibility, that changing such a flag on the repository might create a new repoversion to satisfy it.

pulpcore/plugin/actions.py
bmbouter: yes, I'm switching it.

pulpcore/plugin/repo_version_utils.py
mdellweg: yes, sorry, i meant class. And instead of subclassing it and replacing its only method you could just as well provide a different function.
It might still be interesting to provide a remove_duplicates function by core, that can be (re-)used in the Repository.finalize_new_version method.

mdellweg:
> Because vadifiers also include modification, I think they don't belong to repo. They belong to the respective operation creating a repo version. The vadifiers are there to allow to reuse common parts (which may be 100% in many cases).

And i would argue here, that they belong to the repository to ensure the same constraints for all (future) versions independently of the operation performed.

pulpcore/plugin/repo_version_utils.py
bmbouter: I switched this to a functional named `remove_duplicates` in my next push.

pulpcore/plugin/repo_version_utils.py
```diff
@@ -0,0 +1,47 @@
...
+    :class:`~pulpcore.plugin.models.RepositoryVersion` to be checked and possibly modified as a
+    parameter to `run()`.
+
+    Args:
+        repository_version: The :class:`~pulpcore.plugin.models.RepositoryVersion` to be checked
+            and possibly modified.
+    """
+    query_for_repo_duplicates_by_type = defaultdict(lambda: Q())
+    for item in repository_version.added():
+        detail_item = item.cast()
```
dralley: Future performance improvement

* Sort these into groups based on their ```pulp_type``` which is present on the master Content model.
* Look up the detail content models that represent the ```pulp_type``` strings
* Query the detail content models directly, in bulk, provided a list of PKs
* Then within each type group check for duplicates

pulpcore/plugin/actions.py
```diff
@@ -0,0 +1,59 @@
...
+
+        if 'add_content_units' in request.data:
+            for url in request.data['add_content_units']:
+                content = self.get_resource(url, Content)
+                add_content_units.append(content.pk)
+
+        if 'remove_content_units' in request.data:
+            for url in request.data['remove_content_units']:
+                if url == '*':
+                    remove_content_units.append(url)
```
mdellweg: Does `*` mean remove everything?
Should we do
```python
if url == "*":
    remove_content_units=["*"]
    break
```

pulpcore/app/models/content.py
```diff
@@ -244,7 +244,7 @@ class Content(MasterModel, QueryMixin):
         _artifacts (models.ManyToManyField): Artifacts related to Content through ContentArtifact
     """
     TYPE = 'content'
-    repo_key = ()
+    repo_key_fields = ()
```
mdellweg: Maybe add a comment, that this is used with `remove_duplicates`.

pulpcore/plugin/actions.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+    @action(detail=True, methods=["post"], serializer_class=RepositoryAddRemoveContentSerializer)
+    def modify(self, request, pk):
+        """
+        Queues a task that creates a new RepositoryVersion by adding and removing content units
+        """
+        add_content_units = []
+        remove_content_units = []
+        repository = Repository.objects.get(pk=pk)
+        repository.cast()
```
mdellweg: Is this `detail_repository = repository.cast()` ?

pulpcore/plugin/repo_version_utils.py
bmbouter: Thanks for filing @dralley https://pulp.plan.io/issues/5701

pulpcore/plugin/actions.py
bmbouter: these methods got removed.

pulpcore/app/models/content.py
bmbouter: Added in next push.

pulpcore/plugin/actions.py
bmbouter: Yes this line is unproductive since the result is never saved. It's also not needed since we only use the `pk` and that's the same for both master and detail. Also I verified that when the line is removed a repository PATCH and a later 'modify' call both use the same reserved_resource so they will be serialized correctly. I'm pushing the change that removes this line now.

pulpcore/plugin/actions.py
bmbouter: Yes I agree with the `break` change because when at least one `remove_content_units` specifies to remove all, additional `remove_content_units` params specified are unproductive.

I am going to leave `remove_content_units=[url]` because it DRYs the definition of `*`.

CHANGES/plugin_api/3541.removal
```diff
@@ -0,0 +1,4 @@
+Renamed the Content.repo_key to be Content.repo_unit_key. Also the calling of `RemoveDuplicates`
```
goosemania: Did you mean Content.repo_key_fields?

goosemania: And also RemoveDuplicates is now `remove_duplicates`

bmbouter: yes but I want to fix in separate PR because I need to unblock the other Prs with the skiptest in this one. Is that ok?

pulpcore/plugin/repo_version_utils.py
```diff
@@ -0,0 +1,47 @@
...
+    one repository. For example, pulp_file's content has `relative_path` for that file within the
+    repository.
+
+    Any content newly added to the :class:`~pulpcore.plugin.models.RepositoryVersion` is checked
+    against existing content in the :class:`~pulpcore.plugin.models.RepositoryVersion` with newer
+    "repository duplicates" replace existing "repository duplicates". Each Content model can define
+    a `repo_key_fields` attribute with the field names to be compared. If all `repo_key_fields`
+    contain the same value for two content units, they are considered "repository duplicates".
+
+    After instantiating `RemoveDuplicates` call it with the `run()` method and pass in the
```
goosemania: remove_duplicates

bmbouter:
Thank you! I will make the followup PR now and ensure it's merged pre-branch and pre-RC8.

CHANGES/plugin_api/3541.removal
bmbouter: followup PR here: https://github.com/pulp/pulpcore/pull/382

pulpcore/plugin/repo_version_utils.py
bmbouter: Followup PR here: https://github.com/pulp/pulpcore/pull/382

PR #374, opened 2019-11-09
Set CONTENT_ORIGIN in Travis when publishing docs

### LINKED ISSUE #5686: pulp_file fails to sync file repos literally containing URL encoding strings like %3a in a file name
### Version
Tested using `python3.11-pulp-file-1.15.1` on a Katello instance

### Describe the bug
If a file in a file repo literally contains for example `%3a` in the filename, then this repo cannot be synced to pulp_file, because the download URL unparses the `%3a` to `:` and runs into a 404, e.g:
```
404, message='Not Found', url=URL('http://my-file-server/pub/file_repo/stupid-file-name-:-blub')
```

### To Reproduce
I manually create a local file repo as follows:
```
mkdir file_repo
touch file_repo/stupid-file-name-%3a-blub
echo "blub" > file_repo/stupid-file-name-%3a-blub
bash make_pulp_manifest.sh
```
Here `make_pulp_manifest.sh` is:
```
TARGET_PATH=file_repo

for TARGET_FILE in "${TARGET_PATH}/"*
do
  FILE="$(basename "${TARGET_FILE}")"
  SHA256="$(sha256sum "${TARGET_FILE}" | cut -d ' ' -f 1)"
  SIZE="$(stat --printf="%s" "${TARGET_FILE}")"
  echo "${FILE},${SHA256},${SIZE}" >> ${TARGET_PATH}/PULP_MANIFEST
done
```
Feel free to create the PULP_MANIFEST file any other way!

Now serve this file repo somewhere and sync it to pulp_file and observe the sync error.
```
404, message='Not Found', url=URL('http://my-file-server/pub/file_repo/stupid-file-name-:-blub')
```

### Expected behavior

Sync succeeds. Not sure how difficult it is to differentiate between a URL that contains URL encodings that should be converted and one like this case where the URL encoding string is literally part of the filename/path.

### Additional context

The issue was originally discovered via the real world use case of trying to synchronize a mounted Ubuntu 22.04 iso into pulp_file. Example "problem files" from the mounted Ubuntu iso:
```
# ls -l /var/www/html/pub/ubuntu22/22.04-x86_64/pool/main/o/openssh/
total 468
-r--r--r--. 1 root root 435336 Jan  3  2024 openssh-server_1%3a8.9p1-3ubuntu0.6_amd64.deb
-r--r--r--. 1 root root  38716 Jan  3  2024 openssh-sftp-server_1%3a8.9p1-3ubuntu0.6_amd6
```

daviddavis:
fixes #5686

### DIFF
--- a/.travis/publish_docs.sh
+++ b/.travis/publish_docs.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+export PULP_CONTENT_ORIGIN=http://$(hostname):24816
 django-admin runserver 24817 >> ~/django_runserver.log 2>&1 &
 sleep 5
 
--- a/CHANGES/5686.misc
+++ b/CHANGES/5686.misc
@@ -1,0 +1,1 @@
+Set CONTENT_ORIGIN when publishing docs.


daviddavis:
Looks like bindings publish job doesn't require CONTENT_ORIGIN to be set manually. Adding WIP until I figure out why the docs publish job needs it but bindings don't.

bmbouter:
Ignore this if this is remedial, but I want to share some info. What I learned was that the pulp-operator bakes the settings inside the containers itself, but the docs build runs outside that and they import the settings [here](https://github.com/pulp/pulpcore/blob/master/docs/conf.py#L31-L32).

I'm not sure why the docs need settings.py set. Does anyone know?

Perhaps it's because the docs have this line and the bindings don't is my theory on why the docs publish job needs it set but the bindings don't. Wdyt about this?

daviddavis:
@bmbouter that makes sense to me. I've updated this PR.

PR #377, opened 2019-11-11
Convert remaining CharField to TextField

dralley:
closes: #5609
https://pulp.plan.io/issues/5609

Exception, the artifact hashes have a preset length.  While we don't gain (or lose) anything from using CharField, it might provide a little tiny amount of extra validation, so I left it as-is.


dev_requirements.txt
```diff
@@ -5,3 +5,4 @@ ipython
 codecov
 coverage
 flake8
+pytest
```
CodeHeeler: Did you mean to make this change as part of this same PR?

pulpcore/app/models/progress.py
```diff
@@ -106,7 +106,7 @@ class ProgressReport(Model):
             it will be set to the current task_id.
     """
     message = models.TextField()
-    code = models.CharField(max_length=36)
+    code = models.TextField()
```
daviddavis: Why remove the max_length here? I thought code was supposed to be a short name for the progress report.

dralley: Sure, but how do we decide how short is too short? I feel like we should just make sure plugin writers know they need to keep it short by convention

PR #378, opened 2019-11-12
Update requirements for dependencies

daviddavis:
Also bump gunicorn version requirement.

[noissue]


evgeni:
:+1: this should fix *my* issues on EL7.

daviddavis:
@mikedep333 will this cause any problems with installing Pulp? This change will compile psycopg2 and it requires some dependencies like psql-devel I believe.

evgeni:
`postgresql-devel` on CentOS7.

mikedep333:
@daviddavis @evgeni [It requires](http://initd.org/psycopg/docs/install.html#build-prerequisites) (on CentOS7, and similar Fedora names):
- postgresql-devel
- postgresql (clients only. for /usr/bin/pg_config)
- python3-devel (I'm fairly certain this is needed for venvs)

I think these are  reasonable RPM requirements to maintain a clean venv without system-wide packages. This change should be made to https://github.com/pulp/pulp_rpm_prerequisites

The alternative would be to use the system packages like python3-psycopg2, but then system-wide packages would always be necessary within the venv.

evgeni:
The linked repo says it's only for pulp-rpm, not for pulpcore?

mikedep333:
@evgeni Correct. Because pulpcore does not need any new RPMs currently.

mikedep333:
@evgeni My bad. Fixing my comment now.

PR #380, opened 2019-11-12
Don't create a new repository version if no content was added or removed.

dkliban:
closes: #3308
https://pulp.plan.io/issues/3308

### DIFF
--- a/CHANGES/3308.feature
+++ b/CHANGES/3308.feature
@@ -1,0 +1,1 @@
+New repository version is not created if no content was added or removed.
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -629,9 +629,13 @@ class RepositoryVersion(Model):
             self.delete()
         else:
             self.repository.finalize_new_version(self)
-            self.complete = True
-            self.save()
-            self._compute_counts()
+            no_change = not self.added() and not self.removed()
+            if no_change:
+                self.delete()
+            else:
+                self.complete = True
+                self.save()
+                self._compute_counts()
 
     def __str__(self):
         return "<Repository: {}; Version: {}>".format(self.repository.name, self.number)
--- a/pulpcore/tests/functional/api/using_plugin/test_crd_publications.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_crd_publications.py
@@ -4,10 +4,11 @@ import unittest
 from itertools import permutations
 
 from pulp_smash import api, config
-from pulp_smash.pulp3.utils import gen_distribution, gen_repo, sync
+from pulp_smash.pulp3.utils import gen_distribution, gen_repo, get_content, modify_repo, sync
 from requests.exceptions import HTTPError
 
 from pulpcore.tests.functional.api.using_plugin.constants import (
+    FILE_CONTENT_NAME,
     FILE_DISTRIBUTION_PATH,
     FILE_PUBLICATION_PATH,
     FILE_REMOTE_PATH,
@@ -181,7 +182,9 @@ class PublicationRepositoryParametersTestCase(unittest.TestCase):
 
     def test_create_only_using_repoversion(self):
         """Create a publication only using repository version."""
-        repo = self.create_sync_repo(3)
+        repo = self.create_sync_repo()
+        for file_content in get_content(repo)[FILE_CONTENT_NAME]:
+            modify_repo(self.cfg, repo, remove_units=[file_content])
         version_href = self.client.get(repo['versions_href'])[1]['pulp_href']
         publication = create_file_publication(self.cfg, repo, version_href)
         self.addCleanup(self.client.delete, publication['pulp_href'])
@@ -213,7 +216,7 @@ class PublicationRepositoryParametersTestCase(unittest.TestCase):
                 ctx.exception.response
             )
 
-    def create_sync_repo(self, number_syncs=1):
+    def create_sync_repo(self):
         """Create and sync a repository.
 
         Given the number of times to be synced.
@@ -224,6 +227,5 @@ class PublicationRepositoryParametersTestCase(unittest.TestCase):
         remote = self.client.post(FILE_REMOTE_PATH, gen_file_remote())
         self.addCleanup(self.client.delete, remote['pulp_href'])
 
-        for _ in range(number_syncs):
-            sync(self.cfg, remote, repo)
+        sync(self.cfg, remote, repo)
         return self.client.get(repo['pulp_href'])
--- a/pulpcore/tests/functional/api/using_plugin/test_distributions.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_distributions.py
@@ -7,7 +7,15 @@ from functools import reduce
 from urllib.parse import urljoin
 
 from pulp_smash import api, config, utils
-from pulp_smash.pulp3.utils import gen_distribution, gen_repo, get_added_content, get_versions, sync
+from pulp_smash.pulp3.utils import (
+    gen_distribution,
+    gen_repo,
+    get_added_content,
+    get_content,
+    get_versions,
+    modify_repo,
+    sync
+)
 from requests.exceptions import HTTPError
 
 from pulpcore.tests.functional.api.using_plugin.constants import (
@@ -75,8 +83,11 @@ class CRUDPublicationDistributionTestCase(unittest.TestCase):
             self.client.post(FILE_REMOTE_PATH, gen_file_remote())
         )
         # create 3 repository versions
-        for _ in range(3):
-            sync(self.cfg, self.remote, self.repo)
+        sync(self.cfg, self.remote, self.repo)
+        self.repo = self.client.get(self.repo['pulp_href'])
+        for file_content in get_content(self.repo)[FILE_CONTENT_NAME]:
+            modify_repo(self.cfg, self.repo, remove_units=[file_content])
+
         self.repo = self.client.get(self.repo['pulp_href'])
 
         versions = get_versions(self.repo)
--- a/pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
@@ -256,9 +256,9 @@ class SyncChangeRepoVersionTestCase(unittest.TestCase):
         Do the following:
 
         1. Create a repository, and a remote.
-        2. Sync the repository an arbitrary number of times.
-        3. Verify that the repository version is equal to the previous number
-           of syncs.
+        2. Sync the repository.
+        3. Remove all content - one by one.
+        3. Verify that the repository version is equal to the number of operations.
         """
         cfg = config.get_config()
         client = api.Client(cfg, api.json_handler)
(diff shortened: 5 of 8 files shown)


CHANGES/3308.feature
```diff
@@ -0,0 +1,2 @@
+New repository version is not created if no content was added or removed.
+
```
bmbouter: nitpick: this extra line is probably not necessary.

pulpcore/app/models/repository.py
```diff
@@ -631,9 +631,12 @@ def __exit__(self, exc_type, exc_value, traceback):
         if exc_value:
             self.delete()
         else:
-            self.complete = True
-            self.save()
-            self._compute_counts()
+            if not self.added() and not self.removed():
+                self.delete()
```
goosemania: It's a matter of preference, just sharing an idea.

It's good that the `if` for errors is separate from the `if` for changes in a repo version because semantically they are very different. At the same time, the chain of nested if/else is not always very readable.

An alternative would be something like this:
```
no_changes = not self.added() and not self.removed()
if exc_value or no_changes:
    self.delete()
￼else:
￼    self.complete = True
￼    self.save()
￼    self._compute_counts()
```

pulpcore/tests/functional/api/using_plugin/test_crd_publications.py
```diff
@@ -181,7 +181,9 @@ def setUpClass(cls):
 
     def test_create_only_using_repoversion(self):
         """Create a publication only using repository version."""
-        repo = self.create_sync_repo(3)
+        repo = self.create_sync_repo()
+        for file_content in get_content(repo)["file.file"]:
```
dralley: There's a constant for this "file.file" string

pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
```diff
@@ -270,14 +270,14 @@ def test_all(self):
         remote = client.post(FILE_REMOTE_PATH, body)
         self.addCleanup(client.delete, remote['pulp_href'])
 
-        number_of_syncs = randint(1, 10)
-        for _ in range(number_of_syncs):
-            sync(cfg, remote, repo)
-
+        sync(cfg, remote, repo)
+        repo = client.get(repo['pulp_href'])
+        for file_content in get_content(repo)["file.file"]:
```
dralley: ditto on the constant

pulpcore/tests/functional/api/using_plugin/test_tasks.py
```diff
@@ -180,37 +181,48 @@ def test_read_fields_created_resources_only(self):
...
-        cls.client = api.Client(config.get_config(), api.page_handler)
-
+        cls.cfg = config.get_config()
+        cls.client = api.Client(cls.cfg, api.page_handler)
+        cls.remote = cls.client.post(FILE_REMOTE_PATH, gen_file_remote())
         cls.repository = cls.client.post(FILE_REPO_PATH, gen_repo())
+        response = sync(cls.cfg, cls.remote, cls.repository)
+        cls.created_repo_version = response['pulp_href']
+        cls.repository = cls.client.get(cls.repository['pulp_href'])
+        for file_content in get_content(cls.repository)["file.file"]:
```
dralley: Ditto on the constant

Also you will want to make sure you modify the ```tearDownClass``` to destroy the resources that you are creating in ```setUpClass```

pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -7,7 +7,8 @@
 from urllib.parse import urljoin
 
 from pulp_smash import api, config, utils
-from pulp_smash.pulp3.utils import gen_distribution, gen_repo, get_added_content, get_versions, sync
+from pulp_smash.pulp3.utils import gen_distribution, gen_repo, get_added_content, get_content,\
```
dralley: This should probably be broken out with ```( )```

pulpcore/tests/functional/api/using_plugin/test_crd_publications.py
dkliban: Thank you! i've added use of it everywhere.

pulpcore/tests/functional/api/using_plugin/test_tasks.py
dkliban: I updated the teardownclass method.

pulpcore/tests/functional/api/using_plugin/test_tasks.py
```diff
@@ -180,37 +185,48 @@ def test_read_fields_created_resources_only(self):
...
     """
 
     @classmethod
     def setUpClass(cls):
         """Create class-wide variables."""
-        cls.client = api.Client(config.get_config(), api.page_handler)
-
+        cls.cfg = config.get_config()
+        cls.client = api.Client(cls.cfg, api.page_handler)
+        cls.remote = cls.client.post(FILE_REMOTE_PATH, gen_file_remote())
```
dralley: This remote needs to be deleted after the test finishes.

dkliban: Thank you! I updated the tearDownClass method.

dkliban: done!

pulpcore/app/models/repository.py
```diff
@@ -628,7 +628,8 @@ def __exit__(self, exc_type, exc_value, traceback):
         """
         Save the RepositoryVersion if no errors are raised, delete it if not
         """
-        if exc_value:
+        no_change = not self.added() and not self.removed()
+        if exc_value or no_change:
             self.delete()
```
gmbnomis: What happens to the task's created_resource entry if there is no change? It is set when creating the new version, but I don't see it being unset anywhere?

pulpcore/app/models/repository.py
```diff
@@ -628,7 +628,8 @@ def __exit__(self, exc_type, exc_value, traceback):
         """
         Save the RepositoryVersion if no errors are raised, delete it if not
         """
-        if exc_value:
+        no_change = not self.added() and not self.removed()
```
gmbnomis: The way this is implemented right now applies to all finalizations of a repo version. I thought when wanted to implement it for sync only?

I think there is a use case for creating repo versions with empty changes for /modify: If you want to set an old repo version to be the latest, just do a modify giving the old repo_version as the base version and no artifacts to add/remove

dkliban: If this old version is different from the latest version, then self.added() and self.removed() will contain values that reflect those differences. If you are trying to create the exact same version as the latest, then no new version will be created. What do you think?

pulpcore/app/models/repository.py
dkliban: I believe the delete cascades to the CreatedResource which inherits from GenericRelationModel. 

https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/task.py#L411
https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/generic.py#L22

gmbnomis: Ahh, thank you. Relational databases are magic.

pulpcore/app/models/repository.py
gmbnomis: I just realized that my mental model of added/removed was wrong when discussing validation/modification with @bmbouter. Hooks for add_content/remove_content actually are different from inspecting added/removed content during finalization.

But you are right of course, .added/.removed is the difference to the previous version.

gmbnomis: This convo made me thinking. How does the user of the API finds out to which repo version the "no change" result refers to?

For example, suppose I want to sync and publish

Prior to this change: After the sync, I can use the created repository version resource I got back in the task result for publication. As a user, I am sure that I publish the result of the sync.

With this change: I can do the same if a get a created resource back. But when I don't, this may happen in different scenarios:
a) the latest version is the same as the published one and there is no change -> no need to publish
b) the latest version is not the same as the published one and there is no change between the latest version and the previous one -> publication required

To find out whether I need to publish, I could try to query the repository and get the latest version. But I can't be sure that this is the latest version the sync has seen (another task may have sneaked in). Thus, I may publish the result of a different operation.

To rule out the "sneak in" case, I might be able to infer from the timestamps (latest version vs. task finish) that this has happened.

I discussed this with @dkliban and we came out with two options:
1. In the no change case, put the latest version used for comparison into the created resource field. Although the version already existed and wasn't created)
2.  Accepting that there is no repo version information is better than providing information in the wrong field

This choice isn't nice (to put it mildly). I would love to hear about other solution proposals.

bmbouter: I'd opt for option (2) because having the data be incorrect is a liability because the user can't distinguish between the one that just ran versus the one that ran two days ago.

It will be a bit difficult for integrators to receive a CreatedResource 80% of the time when calling sync(), and then one day they don't get one. We should document this up-front in the docs (maybe a story could be filed to do so and I could fix soon). It's a case integrators need to think about and be aware of and think about how they want to handle. I think that would be better than us handling it in this way for them, although it will take more effort from them.

Also in terms of a human user, I am hoping for a log-streaming addition to the task API in the future which would allow the sync code to log data directly to the human user. `pulp_ansible` has a plugin-based version of this. If that feature is available in the future I think human users would know easily.

pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -75,8 +83,11 @@ def test_01_create(self):
             self.client.post(FILE_REMOTE_PATH, gen_file_remote())
         )
         # create 3 repository versions
-        for _ in range(3):
```
kersommoura: Could we still keep a test to sync a few times using the same remote, and assuring that no repository version was created?

pulpcore/app/models/repository.py
daviddavis: I'd like to propose option 3, which we have discussed quite a bit. We'd eventually provide an option to sync that would allow users to always create a repo version even if nothing changed. I'd suggest we NOT pursue this until users request it or end up in the situation that @gmbnomis describes but it does solve @gmbnomis's problem should we ever encounter it.

mdellweg: I agree, that reporting something as created that isn't is wrong. (So this will be option 2)
As a long term solution, the tasks could be much more verbose (For failed tasks we already put the whole stacktrace there.). The CreatedResource model could turn into an AffectedResource that has some additional attributes.
But for the short term integrators, the proper solution imo is to query for the last repoversion, trigger the sync/modification (probably even based on that one) and in case none got created, use the result of before. Then there is no option for any other action to sneak in.

pulpcore/tests/functional/api/using_plugin/test_distributions.py
dkliban: We have tests like that in pulp-file. https://github.com/pulp/pulp_file/pull/311/files

daviddavis:
The code in this PR looks good to me. Can you update the docs though to state that new repo versions aren't created if content doesn't change?

pulpcore/app/models/repository.py
```diff
@@ -625,7 +625,8 @@ def __exit__(self, exc_type, exc_value, traceback):
         """
         Finalize and save the RepositoryVersion if no errors are raised, delete it if not
         """
-        if exc_value:
+        no_change = not self.added() and not self.removed()
+        if exc_value or no_change:
             self.delete()
         else:
             self.repository.finalize_new_version(self)
```
gmbnomis: That's at the wrong place now. a finalizer may for example render a changed version into one without a change.

pulpcore/app/models/repository.py
gmbnomis: @daviddavis and @mdellweg yes, both proposals would solve the problem.

@mdellweg If you can't set the base_version (which is the case for sync), querying the latest repo version first does not help. Between the query and the sync, another task may sneak in. For example, another sync. Then the second sync would most probably produce no change and, consequently, one would not publish the repo version.

pulpcore/app/models/repository.py
dralley: Note that because of how GFKs work the cascade doesn't actually happen in the database, it's enforced by Django.  So it might be worthwhile to double check all of this.

pulpcore/app/models/repository.py
mdellweg: @gmbnomis So what you are saying is "If you have two users, one syncs the reopsitory daily at 8 and the other one at 9, the second one will never realize, that he needs to republish." It does not even help to look at that in a more Einsteinian way than Newtonian.
And we surely do not want to resync the same thing from another base_version.
In the end, it again sounds like `sync` should be 'find or create a repository version that equals upstream'. Alternatively, the sync task must be chained with the publish task that runs even if the new version has not been created by that sync task.

PR #385, opened 2019-11-14
Delete repository version if exception is raised during finalize_new_…

dkliban:
…version().

closes: #5712
https://pulp.plan.io/issues/5712

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -628,14 +628,18 @@ class RepositoryVersion(Model):
         if exc_value:
             self.delete()
         else:
-            self.repository.finalize_new_version(self)
-            no_change = not self.added() and not self.removed()
-            if no_change:
+            try:
+                self.repository.finalize_new_version(self)
+                no_change = not self.added() and not self.removed()
+                if no_change:
+                    self.delete()
+                else:
+                    self.complete = True
+                    self.save()
+                    self._compute_counts()
+            except Exception:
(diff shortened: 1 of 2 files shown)


pulpcore/app/models/repository.py
```diff
@@ -628,14 +628,18 @@ def __exit__(self, exc_type, exc_value, traceback):
...
-            if no_change:
+            try:
+                self.repository.finalize_new_version(self)
+                no_change = not self.added() and not self.removed()
+                if no_change:
+                    self.delete()
+                else:
+                    self.complete = True
+                    self.save()
+                    self._compute_counts()
```
goosemania: Do we care only about catching  exceptions for finalize or for everything else as well? I thought that for finalize only but I can be mistaken.

dkliban: I wasn't sure, so I decided to include everything else also. @bmbouter what do you tihnk?

bmbouter: We do need to catch and exception if emitted from `finalize_new_version` (I think).

daviddavis: Perhaps catching exceptions for finalize only is better until we have reason to catch exceptions for other things?

bmbouter: Since we can't provide exception handling here, any exception emitted is a fatal exception. When a fatal exception occurs I think we have to delete the RepositoryVerison.

PR #386, opened 2019-11-14
Delete repository version if exception is raised during finalize_new_…

dkliban:
…version().

closes: #5712
https://pulp.plan.io/issues/5712


bmbouter:
This needs to go against master and I'll cherry pick it back to 3.0

PR #393, opened 2019-11-18
Assuring pk is present when adding repository content

fabricio-aguiar:
https://pulp.plan.io/issues/5746
closes #5746

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -475,13 +475,14 @@ class RepositoryVersion(Model):
 
         repo_content = []
         for content_pk in content.exclude(pk__in=self.content).values_list('pk', flat=True):
-            repo_content.append(
-                RepositoryContent(
-                    repository=self.repository,
-                    content_id=content_pk,
-                    version_added=self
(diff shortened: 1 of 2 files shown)


pulpcore/app/models/repository.py
```diff
@@ -475,13 +475,14 @@ def add_content(self, content):
 
         repo_content = []
         for content_pk in content.exclude(pk__in=self.content).values_list('pk', flat=True):
-            repo_content.append(
-                RepositoryContent(
-                    repository=self.repository,
-                    content_id=content_pk,
-                    version_added=self
+            if content_pk:
```
gmbnomis: I am wondering if this fixes a symptom and not the root cause: `content` is supposed to be a Queryset. How can it happen that a content unit in the DB has no primary key?

fabricio-aguiar: you are right, I was about to close this PR, as I saw it breaking in another case

fabricio-aguiar:
Addressing issue here:
https://github.com/pulp/pulp_rpm/pull/1519

PR #395, opened 2019-11-19
Fix repository parameter name in OpenAPI schema

dkliban:
The prefix for the *_repository_version_href was getting applied to *_repository_href slug
names. This patch limits it to only *_repository_version_href.

re: #5760
https://pulp.plan.io/issues/5760

### DIFF
--- a/pulpcore/app/openapigenerator.py
+++ b/pulpcore/app/openapigenerator.py
@@ -9,7 +9,7 @@ from drf_yasg.inspectors import SwaggerAutoSchema
 from drf_yasg.openapi import Parameter
 from drf_yasg.utils import filter_none, force_real_str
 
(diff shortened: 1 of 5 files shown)


pulpcore/app/openapigenerator.py
```diff
@@ -106,7 +106,7 @@ def get_paths(self, endpoints, components, request, public):
                             resource_model = view.queryset.model
                         resource_name = self.get_parameter_name(resource_model)
                         prefix_ = None
-                        if issubclass(view_cls, RepositoryVersionViewSet):
+                        if issubclass(resource_model, RepositoryVersion):
```
fabricio-aguiar: it would take me a lot of time to discover it, thanks @dkliban !

PR #405, opened 2019-11-21
Move Rest API to its own section.

bmbouter:
We originally wanted the left-side navigation link to directly link to
the ../restapi.html page, but sphinx doesn't allow for that. The
left-navigation comes from the toctree, which requires each entry to be
a page.

https://pulp.plan.io/issues/5722
closes #5722

### DIFF
--- a/CHANGES/5722.doc
+++ b/CHANGES/5722.doc
@@ -1,0 +1,1 @@
+Adding dedicated `Rest API` left-navigation section.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -98,8 +98,9 @@ Table of Contents
    from-pulp-2
    components
    installation/index
-   plugins/index
    workflows/index
+   plugins/index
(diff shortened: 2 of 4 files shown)


bmbouter:
The new page looks like this:

image

ipanova:
@bmbouter we keep forgetting to regenerate the docs.

bmbouter:
> @bmbouter we keep forgetting to regenerate the docs.

I don't understand what you mean. I regenerated the docs locally, and travis will generate them again once merged.

ipanova:
> > @bmbouter we keep forgetting to regenerate the docs.
> 
> I don't understand what you mean. I regenerated the docs locally, and travis will generate them again once merged.

disregard, i guess i got confused by this PR https://github.com/pulp/pulp_rpm/pull/1501
if travis generates the docs then we should not take care of the in our PRs

PR #407, opened 2019-11-22
Release resources when worker is cleaned up

bmbouter:
https://pulp.plan.io/issues/5673
closes #5673

### DIFF
--- a/pulpcore/tasking/util.py
+++ b/pulpcore/tasking/util.py
@@ -54,6 +54,7 @@ def cancel(task_id):
         task_status.state = TASK_STATES.CANCELED
         task_status.save()
         _delete_incomplete_resources(task_status)
+        task_status.release_resources()
 
     _logger.info(_('Task canceled: {id}.').format(id=task_id))
     return task_status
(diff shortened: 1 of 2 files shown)


pulpcore/tasking/util.py
```diff
@@ -54,6 +54,7 @@ def cancel(task_id):
         task_status.state = TASK_STATES.CANCELED
         task_status.save()
         _delete_incomplete_resources(task_status)
+        task_status.release_resources()
```
dralley: What's the distinction between ```_delete_incomplete_resources()``` and ```task_statis.release_resources()```?  Are they similar enough to be merged into one?

bmbouter: They serve different purposes, but the names aren't very good. This new line releases "ReservedResource" model instances; these are strings. `_delete_incompleteresources` is about deleting CreatedResource objects, e.g. `RepositoryVersion` objects.

daviddavis: Specifically, CreatedResources that have `complete` and have it set to False.

PR #409, opened 2019-11-22
cherry-pick: Problem: Travis will eventually check out the wrong plugin branches for pulpcore 3.0.

mikedep333:
Solution:
1. Regenerate with latest plugin-template,
2. "repurpose" its var pulpcore_branch as the branch for
pulp_file: master/master & 3.0/0.1
3. Specify the current only branch for pulp-certguard: master
4. Add plugin-template's standard bash settings for robustness
and verbosity to pre_before_install.sh

Also cleanup the effectively empty .travis/post_script.sh

[noissue]

re: #5782
plugin-template needs to support both the pulpcore 3.0 and master branch for plugins
https://pulp.plan.io/issues/5782


daviddavis:
Would it be possible instead of repurposing `pulpcore_branch` to use `additional_plugins` instead?

mikedep333:
@daviddavis Yes, but it would require noticeably more refactoring work across the pulpcore-added https://github.com/pulp/pulpcore/blob/master/.travis/pre_before_install.sh & plugin-template.

@dkliban and I thought this would be a good compromise for unbreaking CI quickly for pulp_container, without the quality of the plugin-template being deteriorated at all. The workaround only exists in pulpcore's added files.

daviddavis:
What sort of refactoring? I was just imaging you set `additional_plugins` and do something like: 

```
yq ".additional_plugins | .[] | select(.name == \"pulp_file\") | .branch" template_config.yml
```

Also, I am not sure the change to pulp_certguard in additional_plugins does anything [since we're checking it out in pre_before_install.sh](https://github.com/pulp/pulpcore/blob/master/.travis/pre_before_install.sh#L13).

mikedep333:
@daviddavis That's a very good idea, but it requires more refactoring in plugin-template. Since plugin-template currently treats pulp_file as the primary $PLUGIN, and would try to install pulp_file twice.

Therefore, I want to do it in a later PR.

And the entire line (without the literal-double-quotes-in-branch-name-error that I spent the last hour fixing) would be:
```
PULP_FILE_BRANCH=$(yq -r ".additional_plugins | .[] | select(.name == \"pulp_file\") | .branch" template_config.yml)
```

daviddavis:
Is installing it twice a problem though? Anyway, handling it in a follow up PR is fine with me. My main objection is that it's nonobvious to someone reading template_config.yml. If it's a temporary measure though, then that's fine.

PR #420, opened 2019-11-26
Add validation to check for file path overlaps

daviddavis:
fixes #5559
https://pulp.plan.io/issues/5559

pulp_file PR: https://github.com/pulp/pulp_file/pull/325

### DIFF
--- a/pulpcore/app/files.py
+++ b/pulpcore/app/files.py
@@ -1,8 +1,10 @@
 import hashlib
 import os
+from gettext import gettext as _
 
 from django.core.files.uploadedfile import TemporaryUploadedFile
 from django.core.files.uploadhandler import TemporaryFileUploadHandler
+from pygtrie import StringTrie
 
 
 class PulpTemporaryUploadedFile(TemporaryUploadedFile):
@@ -90,3 +92,41 @@ class TemporaryDownloadedFile(TemporaryUploadedFile):
         if name is None:
             name = getattr(file, 'name', None)
         self.name = name
+
+
+def validate_file_paths(paths):
+    """
+    Check for valid POSIX paths (ie ones that aren't duplicated and don't overlap).
+
+    Overlapping paths are where one path terminates inside another (e.g. a/b and a/b/c).
+
+    This function will raise an exception at the first dupe or overlap it detects. We use a trie (or
+    prefix tree) to keep track of which paths we've already seen.
+
+    Args:
+        paths (iterable of str): An iterable of strings each representing a relative path
+
+    Raises:
+        ValueError: If any path overlaps another
+    """
+    overlap_error = _("The path for file '{path}' overlaps: {conflicts}")
+
+    path_trie = StringTrie(separator="/")
+    for path in paths:
+        if path in path_trie:
+            # path duplicates a path already in the trie
+            raise ValueError(_("Path is duplicated: {path}").format(path=path))
+
+        if path_trie.has_subtrie(path):
+            # overlap where path is 'a/b' and trie has 'a/b/c'
+            conflicts = [item[0] for item in path_trie.items(prefix=path)]
+            raise ValueError(overlap_error.format(path=path, conflicts=(", ").join(conflicts)))
+
+        prefixes = list(path_trie.prefixes(path))
+        if prefixes:
+            # overlap where path is 'a/b/c' and trie has 'a/b'
+            conflicts = [prefix.key for prefix in prefixes]
+            raise ValueError(overlap_error.format(path=path, conflicts=(", ").join(conflicts)))
+
+        # if there are no overlaps, add it to our trie and continue
+        path_trie[path] = True
--- a/pulpcore/tests/unit/test_files.py
+++ b/pulpcore/tests/unit/test_files.py
@@ -1,0 +1,36 @@
+from unittest import TestCase
+
+from pulpcore.app.files import validate_file_paths
+
+
+class TestValidateFilePaths(TestCase):
+    def test_valid_paths(self):
+        """
+        Test for valid paths.
+        """
+        paths = ['a/b', 'a/c/b', 'PULP_MANIFEST', 'b']
+        validate_file_paths(paths)
+
+    def test_dupes(self):
+        """
+        Test for two duplicate paths.
+        """
+        paths = ['a/b', 'PULP_MANIFEST', 'PULP_MANIFEST']
+        with self.assertRaisesRegex(ValueError, "Path is duplicated: PULP_MANIFEST"):
+            validate_file_paths(paths)
+
+    def test_overlaps(self):
+        """
+        Test for overlapping paths.
+        """
+        paths = ['a/b', 'a/b/c']
+        with self.assertRaises(ValueError):
+            validate_file_paths(paths)
+
+        paths = ['a/b/c', 'a/b']
+        with self.assertRaises(ValueError):
+            validate_file_paths(paths)
+
+        paths = ['b/c', 'a/b', 'b']
+        with self.assertRaises(ValueError):
+            validate_file_paths(paths)
--- a/CHANGES/plugin_api/5559.feature
+++ b/CHANGES/plugin_api/5559.feature
@@ -1,0 +1,3 @@
+Added artifact path overlap checks for repo versions and publications. Plugin writers should call
+``validate_version_paths()`` or ``validate_publication_paths()`` during the finalize step when
+creating RepositoryVersions or Publications (respectively).
--- a/doc_requirements.txt
+++ b/doc_requirements.txt
@@ -11,6 +11,7 @@ drf-yasg
 dynaconf
 plantuml
 pyyaml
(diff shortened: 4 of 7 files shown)


pulpcore/app/models/publication.py
```diff
@@ -107,6 +110,46 @@ def delete(self, **kwargs):
             CreatedResource.objects.filter(object_id=self.pk).delete()
             super().delete(**kwargs)
 
+    def validate_path_overlap(self):
```
bmbouter: What calls this?

daviddavis: It's an open question. Right now I put it in the plugin (see https://github.com/pulp/pulp_file/pull/325) but I'm probably going to move it to `Publication.__exit__`.

bmbouter: Oh I didn't see the other PR. Cool. If we can have it called automatically that would be great. +1 to moving it to `Publication.__exit__`.

pulpcore/app/models/publication.py
```diff
@@ -125,6 +155,12 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             exc_val (Exception): (optional) Instance of exception raised.
             exc_tb (types.TracebackType): (optional) stack trace.
         """
+        try:
+            self.validate_path_overlap()
```
gmbnomis: Is `validate_path_overlap()` really generic enough so that every plugin wants to have it? 

- In pulp_cookbook this check is completely superfluous, as the path is derived from the name and version of the cookbook (which is the repo version key, i.e. is guaranteed to be unique anyway) 
- In a plugin that wants to provide a "static" read-only rest interface (and does not care about exporting to a POSIX file system), I could imagine paths like "endpoint", "endpoint/a", "endpoint/b" to be useful.

Additionally, plugin writers might want to provide additional validations in their derived Publications and it feels odd forcing them to override `validate_path_overlap()` to accomplish this.

I think we should follow the pattern of `Repository.finalize_new_version()` here, i.e. define a (set of) common helper functions and a `Publication.finalize_new_publication()` hook that can be used to define the validations in the plugin.

pulpcore/app/models/repository.py
```diff
@@ -128,7 +129,7 @@ def finalize_new_version(self, new_version):
         Returns:
 
         """
-        pass
+        new_version.validate_path_overlap()
```
gmbnomis: As said above, I think this check is not generic enough. Moreover, if a plugin uses `PublishedArtifacts`, overlapping content paths in a repo version may be perfectly fine.

pulpcore/app/models/repository.py
```diff
@@ -654,6 +655,24 @@ def _compute_counts(self):
                     counts_list.append(count_obj)
             RepositoryVersionContentDetails.objects.bulk_create(counts_list)
 
+    def validate_path_overlap(self):
```
gmbnomis: Shouldn't this live in `repo_version_utils.py`?

pulpcore/app/models/publication.py
```diff
@@ -107,6 +110,33 @@ def delete(self, **kwargs):
...
+        paths += self.published_artifact.values_list("relative_path", flat=True)
+        paths += self.published_metadata.values_list("relative_path", flat=True)
+
+        if self.pass_through:
+            paths += ContentArtifact.objects. \
+                filter(content__pk__in=self.repository_version.content). \
+                values_list("relative_path", flat=True)
+
+        for path in paths:
+            matches = [p for p in paths if re.match(f"^{path}/.*", p)]
```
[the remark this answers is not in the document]
daviddavis: Yes, this was a first pass, knowing that I would have to optimize it later. I'll get a benchmark and look at optimizing it.

pulpcore/app/models/repository.py
daviddavis: I did this to be consistent with publication. I guess I should define `publication_utils.py`?

pulpcore/app/models/publication.py
daviddavis: See https://github.com/pulp/pulpcore/pull/420#discussion_r350971599. I didn't have the validate functionality automatically called but @bmbouter suggested that it should be. I'm fine with your proposal though. @bmbouter thoughts?

pulpcore/app/models/repository.py
gmbnomis: Yes, I would prefer that to make it consistent with the validation mechanism we introduced for repo versions.

pulpcore/app/files.py
```diff
@@ -90,3 +92,35 @@ def __init__(self, file, name=None):
...
+    Args:
+        paths (list): A list of strings each representing a relative path
+
+    Raises:
+        ValueError: If any path overlaps another
+    """
+    error = _("The path for file '{path}' overlaps: {conflicts}")
+
+    path_trie = StringTrie(separator="/")
+    for path in set(paths):
```
gmbnomis: Hmm, the `set` makes me wonder whether we need to look at duplicate paths during validation. For example, in pulp file there could be a "PULP_MANIFEST" content unit and a "PULP_MANIFEST" meta data unit. Or is there another mechanism to prevent that?

pulpcore/app/files.py
```diff
@@ -90,3 +92,35 @@ def __init__(self, file, name=None):
...
+            raise ValueError(error.format(path=path, conflicts=(", ").join(conflicts)))
+
+        prefixes = list(path_trie.prefixes(path))
+        if prefixes:
+            # overlap where path is 'a/b/c' and trie has 'a/b'
+            conflicts = [prefix.key for prefix in prefixes]
+            raise ValueError(error.format(path=path, conflicts=(", ").join(conflicts)))
+
+        # if there are no overlaps, add it to our trie and continue
+        path_trie[path] = True
```
gmbnomis: Nice!

pulpcore/app/files.py
daviddavis: There is no mechanism to prevent that currently. I debated about whether to handle that here or file a separate issue to create a new mechanism to do so. The former would be more performant (only having to traverse the paths once) while the latter would be more flexible (in case plugin writers want one validation but not the other). I can't think of such a case though where a plugin writer might want to prevent dupe paths but not path overlaps?

gmbnomis: Yes, "dupe only" looks like a very special case.

If I understand your code correctly, it looks like it is actually more work to not check for duplicates. (If we iterate over `paths` directly (instead of `set(path)`), dupes should be caught by the check on prefixes, because a path is its own prefix).

If that is indeed the case, we can save time and, especially, memory when including the check for duplicates. And, if it turns out that someone really needs a "dupe only" check, we could add a separate check later. wdyt?

pulpcore/app/models/publication.py
bmbouter: Sorry to just be responding now, but I'm convinced we should introduce `Publication.finalize_new_publication()` as you suggest. That would make this type of checking an opt-in thing. Am I understanding this correctly?

daviddavis: Yes, that's my understanding. Also, the repo version check would be opt-in too.

daviddavis: I broke off this hook into a separate PR since it seemed like it should be a separate commit/changelog/etc 

https://github.com/pulp/pulpcore/pull/426

daviddavis:
@bmbouter @gmbnomis this should be ready for re-review.

pulpcore/app/files.py
```diff
@@ -90,3 +92,41 @@ def __init__(self, file, name=None):
...
+    """
+    Check for valid POSIX paths (ie ones that aren't duplicated and don't overlap).
+
+    Overlapping paths are where one path terminates inside another (e.g. a/b and a/b/c).
+
+    This function will raise an exception at the first dupe or overlap it detects. We use a trie (or
+    prefix tree) to keep track of which paths we've already seen.
+
+    Args:
+        paths (list): A list of strings each representing a relative path
```
gmbnomis: nitpick: this actually needs only an iterable of strings

pulpcore/tests/unit/test_files.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def test_overlaps(self):
+        """
+        Test for overlapping paths.
+        """
+        paths = ['a/b', 'a/b/c']
+        with self.assertRaises(ValueError):
+            validate_file_paths(paths)
+
+        paths = ['a/b/c', 'a/b']
```
gmbnomis: Suggestion: Also include test cases in which an overlap adds multiple levels, e.g. `['a/b/c/d', 'a/b']`  and `['a/b', 'a/b/c/d']`. 

I'm pretty certain that your implementation does not have problems with it, but it is a conceptually different case that needs to be caught.

bmbouter:
I'm going to merge now to include in the next (and hopefully final) RC. @daviddavis are you able to send any final test changes in another PR?

PR #428, opened 2019-12-02
Have BaseModel save() call full_clean()

daviddavis:
fixes #5828
https://pulp.plan.io/issues/5828


bmbouter:
I wrote some about this on the list: I think we should close for now. https://www.redhat.com/archives/pulp-dev/2019-December/msg00018.html  Please reopen if you think we should.

PR #432, opened 2019-12-03
WIP - Improve memory performance on syncing

fabricio-aguiar:
https://pulp.plan.io/issues/5688
ref #5688

### DIFF
--- a/CHANGES/5688.bugfix
+++ b/CHANGES/5688.bugfix
@@ -1,0 +1,1 @@
+Improve speed and memory performance.
--- a/pulpcore/app/models/progress.py
+++ b/pulpcore/app/models/progress.py
@@ -135,9 +135,11 @@ class ProgressReport(BaseModel):
         kwargs (dict): keyword arguments to be passed on to the real save
         """
         now = timezone.now()
+        batch_interval = kwargs.pop("batch_interval", BATCH_INTERVAL)
+        saving_periodically = (batch_interval != BATCH_INTERVAL) or self._using_context_manager
 
-        if self._using_context_manager and self._last_save_time:
-            if now - self._last_save_time >= datetime.timedelta(milliseconds=BATCH_INTERVAL):
+        if saving_periodically and self._last_save_time:
+            if now - self._last_save_time >= datetime.timedelta(milliseconds=batch_interval):
                 super().save(*args, **kwargs)
                 self._last_save_time = now
         else:
@@ -177,7 +179,7 @@ class ProgressReport(BaseModel):
             self.state = TASK_STATES.FAILED
         self.save()
 
-    def increment(self):
+    def increment(self, *args, **kwargs):
         """
         Increment done count and save the progress report.
 
@@ -188,7 +190,7 @@ class ProgressReport(BaseModel):
         if self.total:
             if self.done > self.total:
                 _logger.warning(_('Too many items processed for ProgressReport %s') % self.message)
-        self.save()
+        self.save(*args, **kwargs)
 
     def iter(self, iter):
         """
--- a/pulpcore/app/util.py
+++ b/pulpcore/app/util.py
@@ -71,3 +71,20 @@ def get_view_name_for_model(model_obj, view_action):
             if registered_viewset is viewset:
                 return '-'.join((base_name, view_action))
     raise LookupError('view not found')
+
+
+def batch_qs(qs, batch_size=1000):
+    """
+    Returns a queryset batch in the given queryset.
+
+    Usage:
+        # Make sure to order your querset
+        article_qs = Article.objects.order_by('id')
+        for qs in batch_qs(article_qs):
+            for article in qs:
+                print article.body
+    """
+    total = qs.count()
+    for start in range(0, total, batch_size):
+        end = min(start + batch_size, total)
+        yield qs[start:end]
--- a/pulpcore/plugin/repo_version_utils.py
+++ b/pulpcore/plugin/repo_version_utils.py
@@ -1,10 +1,10 @@
-from collections import defaultdict
 from gettext import gettext as _
 import logging
 
 from django.db.models import Q
 
-from pulpcore.app.models import ContentArtifact
+from pulpcore.app.models import Content, ContentArtifact
+from pulpcore.app.util import batch_qs
 from pulpcore.app.files import validate_file_paths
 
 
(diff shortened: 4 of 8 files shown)


pulpcore/app/models/progress.py
```diff
@@ -135,9 +135,11 @@ def save(self, *args, **kwargs):
         kwargs (dict): keyword arguments to be passed on to the real save
         """
         now = timezone.now()
+        batch_interval = kwargs.pop("batch_interval", BATCH_INTERVAL)
+        saving_periodically = (batch_interval != BATCH_INTERVAL) or self._using_context_manager
 
-        if self._using_context_manager and self._last_save_time:
-            if now - self._last_save_time >= datetime.timedelta(milliseconds=BATCH_INTERVAL):
+        if saving_periodically and self._last_save_time:
+            if now - self._last_save_time >= datetime.timedelta(milliseconds=batch_interval):
```
fabricio-aguiar: In some loopings, it was creating a lot of queries for updating the `done` count.
slight improvement in memory consumption

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -41,15 +42,33 @@ class QueryExistingArtifacts(Stage):
...
+                    update_kwargs(d_artifact.artifact, all_kwargs)
+
+            for k, v in all_kwargs.items():
+                key = k
+                value = v[0]
 
-            for artifact in Artifact.objects.filter(all_artifacts_q):
+                if len(v) > 1:
+                    key = f"{k}__in"
+                    value = v
```
fabricio-aguiar: As @dralley suggested on the email thread, sometimes **OR** is not efficient for queries, so I gathered the data to make an **IN** clause
slight improvement in memory consumption

pulpcore/plugin/stages/content_stages.py
```diff
@@ -36,16 +36,46 @@ class QueryExistingContents(Stage):
...
+            for k, v in all_kwargs.items():
+                for key, value in v.items():
+                    _key = key
+                    _value = value[0]
+
+                    if len(value) > 1:
+                        _key = f"{key}__in"
+                        _value = value
+
+                    content_q_by_type[k] |= Q(**{_key: _value})
```
fabricio-aguiar: As @dralley suggested on the email thread, sometimes **OR** is not efficient for queries, so I gathered the data to make an **IN** clause
slight improvement in memory consumption

pulpcore/plugin/stages/content_stages.py
```diff
@@ -36,16 +36,46 @@ class QueryExistingContents(Stage):
...
 
             for model_type in content_q_by_type.keys():
-                for result in model_type.objects.filter(content_q_by_type[model_type]):
+                for result in model_type.objects.filter(content_q_by_type[model_type]).iterator():
```
fabricio-aguiar: avoiding to cache queries

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -41,15 +42,33 @@ class QueryExistingArtifacts(Stage):
...
+
+                all_artifacts_q |= Q(**{key: value})
+
+            for artifact in Artifact.objects.filter(all_artifacts_q).iterator():
```
fabricio-aguiar: avoiding to cache queries

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -130,7 +148,7 @@ def _add_to_pending(coro):
                                 content_get_task = None
                         else:
                             pb.done += task.result()  # download_count
-                            pb.save()
+                            pb.save(batch_interval=2000)
```
fabricio-aguiar: it was saving around 50 times in a row

CHANGES/5688.bugfix
```diff
@@ -0,0 +1 @@
+Improve memory performance.
```
bmbouter: Can it also call out the speed improvements.

pulpcore/plugin/stages/artifact_stages.py
bmbouter: I'm concerned about these changes in this module and the code complexity. This code is already pretty complicated, and having it be less readable will be more difficult in the future. Do we know much this is improving the memory/runtime?

pulpcore/app/models/progress.py
```diff
@@ -177,7 +179,7 @@ def __exit__(self, type, value, traceback):
             self.state = TASK_STATES.FAILED
         self.save()
 
-    def increment(self):
+    def increment(self, *args, **kwargs):
```
bmbouter: I didn't expect we want to pass through `*args` and `**kwargs`.

pulpcore/app/models/progress.py
```diff
@@ -188,7 +190,7 @@ def increment(self):
         if self.total:
             if self.done > self.total:
                 _logger.warning(_('Too many items processed for ProgressReport %s') % self.message)
-        self.save()
+        self.save(*args, **kwargs)
```
bmbouter: I didn't expect these arg/kwargs.

pulpcore/app/models/progress.py
bmbouter: Is the issue that the default BATCH_INTERVAL is too slow? If so can we increase that because that is the situation the throttling of progress reports was for. Maybe set BATCH_INTERVAL to 2000?

pulpcore/plugin/repo_version_utils.py
```diff
@@ -13,6 +12,23 @@
 __all__ = ["remove_duplicates"]
 
 
+def batch_qs(qs, batch_size=1000):
```
bmbouter: Can we keep this internal by renaming `batch_qs` -> `_batch_qs`? Once plugins need it we can add it to the plugin API making it public.

pulpcore/plugin/stages/api.py
```diff
@@ -81,7 +81,7 @@ class MyStage(Stage):
             log.debug(_('%(name)s - next: %(content)s.'), {'name': self, 'content': content})
             yield content
 
-    async def batches(self, minsize=50):
+    async def batches(self, minsize=500):
```
bmbouter: :+1:

pulpcore/plugin/stages/api.py
```diff
@@ -163,7 +163,7 @@ def __str__(self):
         return '[{id}] {name}'.format(id=id(self), name=self.__class__.__name__)
 
 
-async def create_pipeline(stages, maxsize=100):
+async def create_pipeline(stages, maxsize=1000):
```
bmbouter: :+1:

pulpcore/plugin/repo_version_utils.py
```diff
@@ -31,21 +47,45 @@ def remove_duplicates(repository_version):
         repository_version: The :class:`~pulpcore.plugin.models.RepositoryVersion` to be checked
             and possibly modified.
     """
-    query_for_repo_duplicates_by_type = defaultdict(lambda: Q())
-    for item in repository_version.added():
-        detail_item = item.cast()
-        if detail_item.repo_key_fields == ():
+    repository_type = repository_version.repository.pulp_type.split(".")[-1]
```
dralley: I was actually working on this in parallel since there was an issue filed for it.  Here's my implementation https://github.com/pulp/pulpcore/compare/master...dralley:performance?diff=split&expand=1#diff-fb9784727f42efcd0b0bcae38b9daa30R34-R59

pulpcore/app/models/progress.py
fabricio-aguiar: we can increase it, but the problem is that currently, the BATCH_INTERVAL is only taken account when ProgressReport is a context manager, and for RPM almost all ProgressReport are not context manager

pulpcore/plugin/repo_version_utils.py
fabricio-aguiar: @dralley  your solution is much more elegant, I think, from my solution we can take the part when I test if queryset is not empty before calling remove_content, and maybe take the batch queryset also.
Currently remove_content does not test if queryset is empty, and due that it consumes a lot of time.

pulpcore/app/models/progress.py
bmbouter: The context manager is what I recommend when using progress reports. Can the usage of Progress Reports adopt them?

pulpcore/plugin/repo_version_utils.py
bmbouter: Can we bring the implementation from @dralley into this PR, or remove this PR's changes and let the other PR go forward in parallel?

fabricio-aguiar: I will remove this change form this PR, this PR is totally WIP, I tested some tweaks if they reduce memory consumption, I push. When I am able to sync everything without comment any part of the code, I will start to adjust the code. Probably many things will be removed from this PR.

fabricio-aguiar:
Closing in favor of https://github.com/pulp/pulpcore/pull/440

pulpcore/app/util.py
```diff
@@ -71,3 +71,20 @@ def get_view_name_for_model(model_obj, view_action):
             if registered_viewset is viewset:
                 return '-'.join((base_name, view_action))
     raise LookupError('view not found')
+
+
+def batch_qs(qs, batch_size=1000):
```
fabricio-aguiar: from here:
https://medium.com/@chakrabortypritish/django-optimization-how-to-avoid-memory-mishaps-1e55b2aa947c

PR #437, opened 2019-12-03
containers: Use Fedora 30's test Errata for libcomps

mikedep333:
until they release the Errata.

So that python3-libcomps can have .egg-info metadata available
for pip to see that libcomps is already satisfied for pulp_rpm.

They already released the corresponding Errata for Fedora 31,
so this should be very safe.

Also, cleanup currently unneeded -devel packages.

[noissue]

re: #5840
as a user of pulp_rpm_prerequisites I have libcomps installed from pypi
https://pulp.plan.io/issues/5840

### DIFF
--- a/containers/images/pulp/Dockerfile.j2
+++ b/containers/images/pulp/Dockerfile.j2
@@ -31,11 +31,13 @@ RUN		dnf -y update && \
 		dnf -y install python3-psycopg2 && \
 		dnf -y install glibc-langpack-en && \
 		dnf -y install python3-createrepo_c && \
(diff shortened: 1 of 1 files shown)


mikedep333:
Status: The [F30](https://bodhi.fedoraproject.org/updates/FEDORA-2019-0d122cc67a) & F31 errata were resubmitted with a fix (actually a newer upstream version of libcomps: 0.1.14 instead of 0.1.12)

I'll re-run the pulp_rpm tests (tomorrow) once it is in the testing repo.

PR #440, opened 2019-12-05
Improve speed and memory performance of syncing

fabricio-aguiar:
https://pulp.plan.io/issues/5688
ref #5688

### DIFF
--- a/pulpcore/app/models/progress.py
+++ b/pulpcore/app/models/progress.py
@@ -15,7 +15,7 @@ from pulpcore.constants import TASK_CHOICES, TASK_STATES
 _logger = logging.getLogger(__name__)
 
 # number of ms between save() calls when _using_context_manager is set
-BATCH_INTERVAL = 500
+BATCH_INTERVAL = 2000
 
 
 class ProgressReport(BaseModel):
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -10,7 +10,7 @@ import django
 from django.db import models, transaction
 from django.urls import reverse
 
-from pulpcore.app.util import get_view_name_for_model
+from pulpcore.app.util import batch_qs, get_view_name_for_model
 from pulpcore.download.factory import DownloaderFactory
 from pulpcore.exceptions import ResourceImmutableError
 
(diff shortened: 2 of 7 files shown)


pulpcore/app/util.py
```diff
@@ -71,3 +71,20 @@ def get_view_name_for_model(model_obj, view_action):
             if registered_viewset is viewset:
                 return '-'.join((base_name, view_action))
     raise LookupError('view not found')
+
+
+def batch_qs(qs, batch_size=1000):
```
fabricio-aguiar: https://medium.com/@chakrabortypritish/django-optimization-how-to-avoid-memory-mishaps-1e55b2aa947c

pulpcore/app/models/repository.py
```diff
@@ -496,6 +500,9 @@ def remove_content(self, content):
         if self.complete:
             raise ResourceImmutableError(self)
 
+        if not content or not content.count():
```
bmbouter: What motivated adding this?

fabricio-aguiar: SQLAlchemy enforces to not pass an empty list due performance problems: https://developpaper.com/query-empty-list-problem-analysis-of-sql-alchemy-in/
So I did it to see how it would affect performance on django orm, and I notice a 300 seconds difference from my previous test without this.

bmbouter: Yup that makes sense. Thank you.

pulpcore/app/models/progress.py
```diff
@@ -15,7 +15,7 @@
 _logger = logging.getLogger(__name__)
 
 # number of ms between save() calls when _using_context_manager is set
-BATCH_INTERVAL = 500
+BATCH_INTERVAL = 2000
```
fao89, line 18: @gerrod3 @bmbouter @daviddavis  maybe we need to change these numbers for pulp_ansible

pulpcore/plugin/stages/api.py
```diff
@@ -81,7 +81,7 @@ async def run(self):
             log.debug(_('%(name)s - next: %(content)s.'), {'name': self, 'content': content})
             yield content
 
-    async def batches(self, minsize=50):
+    async def batches(self, minsize=500):
```
fao89, line 84: @gerrod3 @bmbouter @daviddavis  maybe we need to change these numbers for pulp_ansible

pulpcore/plugin/stages/api.py
```diff
@@ -163,7 +163,7 @@ def __str__(self):
         return '[{id}] {name}'.format(id=id(self), name=self.__class__.__name__)
 
 
-async def create_pipeline(stages, maxsize=100):
+async def create_pipeline(stages, maxsize=1000):
```
fao89, line 166: @gerrod3 @bmbouter @daviddavis  maybe we need to change these numbers for pulp_ansible

PR #441, opened 2019-12-05
Improve performance of duplicate removal, and prevent content from contaminating typed repos

dralley:


pulpcore/plugin/repo_version_utils.py
```diff
@@ -31,21 +31,32 @@ def remove_duplicates(repository_version):
...
-        repository_version.remove_content(qs)
+
+        if new_content_qs.count():
+            find_dup_qs = Q()
+            for content_unit in new_content_qs:
+                unit_q_dict = {
+                    field: getattr(content_unit, field) for field in type_obj.repo_key_fields
+                }
+                item_query = Q(**unit_q_dict) & ~Q(pk=content_unit.pk)
+                find_dup_qs |= item_query
```
fabricio-aguiar: I would do a batch queryset here, RHEL 7 has about 27000 packages, so this query would result in 27000 objects + 27000 unit_q_dicts

pulpcore/plugin/repo_version_utils.py
```diff
@@ -31,21 +30,32 @@ def remove_duplicates(repository_version):
...
-        query_for_repo_duplicates_by_type[detail_item._meta.model] |= item_query
-
-    for model in query_for_repo_duplicates_by_type:
-        _logger.debug(_("Removing duplicates for type: {}".format(model)))
-        qs = model.objects.filter(query_for_repo_duplicates_by_type[model])
-        repository_version.remove_content(qs)
+
+        if new_content_qs.count():
+            find_dup_qs = Q()
+            for content_unit in new_content_qs:
```
fabricio-aguiar: if you do:
```suggestion
            repo_key_fields_with_pk = type_obj.repo_key_fields + ('pk',)
            for content_unit in new_content_qs.values(*repo_key_fields_with_pk):
```
it would already bring the result in dicts, so you won't need to do a dict comprehension for each item
```python
pk = str(content_unit.pop("pk"))
item_query = Q(**content_unit) & ~Q(pk=pk)
```

pulpcore/tests/unit/models/test_repo_version.py
```diff
@@ -1,47 +0,0 @@
-from unittest.mock import patch
```
dralley: All this is moved to ```pulp_file```.  It has to be, because "core.content" is no longer a valid type of content to be in a repository, and you're not supposed to be creating a generic "Repository" anymore either.

CHANGES/5701.feature
```diff
@@ -0,0 +1 @@
+Pulp will do validation that a new repository version contains only content which is supported by the Repository type. Using the same a-priori knowledge of content types, increase performance of duplicate removal.
```
bmbouter: Can this be line wrapped to 100 chars?

bmbouter: Also can a newline character be added to the end of each CHANGES file.

pulpcore/plugin/repo_version_utils.py
```diff
@@ -63,3 +77,19 @@ def validate_version_paths(version):
         validate_file_paths(paths)
     except ValueError as e:
         raise ValueError(_("Cannot create repository version. {err}.").format(err=e))
+
+
+def batch_qs(qs, batch_size=1000):
```
bmbouter: Can this be de-duped to use the one from @fabricio-aguiar 's PR?

pulpcore/plugin/repo_version_utils.py
```diff
@@ -31,21 +31,36 @@ def remove_duplicates(repository_version):
...
+
+        if new_content_qs.count() and existing_content.count():
+            _logger.debug(_("Removing duplicates for type: {}".format(type_obj.get_pulp_type())))
+
+            for batch in batch_qs(new_content_qs):
+                find_dup_qs = Q()
+
+                for content_dict in batch:
+                    content_pk = content_dict.pop('pk')
+                    item_query = Q(**content_dict) & ~Q(pk=content_pk)
```
gmbnomis: Do we need `& ~Q(pk=content_pk)` here? It look like these pk are filtered out by `.filter(pk__in=existing_content)` below anyway?

pulpcore/plugin/repo_version_utils.py
```diff
@@ -31,21 +31,36 @@ def remove_duplicates(repository_version):
...
-    query_for_repo_duplicates_by_type = defaultdict(lambda: Q())
-    for item in repository_version.added():
-        detail_item = item.cast()
-        if detail_item.repo_key_fields == ():
+    added_content = repository_version.added().only('pk')
+    existing_content = repository_version.content.exclude(pk__in=added_content)
+    repository = repository_version.repository.cast()
+    content_types = {ctype.get_pulp_type(): ctype for ctype in repository.CONTENT_TYPES}
+
+    for pulp_type in content_types.keys():
```
gmbnomis:
```suggestion
    for pulp_type, type_obj in content_types.items():
```

and remove the next line.

pulpcore/plugin/repo_version_utils.py
dralley: Fair point, probably not.

dralley:
Comments addressed

pulpcore/plugin/repo_version_utils.py
```diff
@@ -31,21 +31,34 @@ def remove_duplicates(repository_version):
...
+            for batch in batch_qs(new_content_qs):
+                find_dup_qs = Q()
+
+                for content_dict in batch:
+                    item_query = Q(**content_dict)
+                    find_dup_qs |= item_query
+
+                duplicates_qs = type_obj.objects.filter(pk__in=existing_content)\
+                                                .filter(find_dup_qs)\
+                                                .only('pk')
```
gmbnomis: This is really neat. I must remember to put it my toolbox.

gmbnomis:
> Comments addressed

@dralley this looks really great!

pulpcore/app/models/repository.py
```diff
@@ -648,11 +666,25 @@ def __exit__(self, exc_type, exc_value, traceback):
...
             try:
-                self.repository.finalize_new_version(self)
+                repository = self.repository.cast()
+                repository.finalize_new_version(self)
                 no_change = not self.added() and not self.removed()
                 if no_change:
                     self.delete()
                 else:
+                    content_types_seen = set(
+                        self.content.values_list('pulp_type', flat=True).distinct()
```
dralley: We might be able to get away with just checking the content used in the added content units.  But I think I'll let that wait until I can start looking at the query generation.  We'll definitely be doing more optimization work post-GA

pulpcore/plugin/stages/association_stages.py
```diff
@@ -79,7 +78,6 @@ def __init__(self, new_version, *args, **kwargs):
         with ProgressReport(message='Un-Associating Content', code='unassociating.content') as pb:
             async for queryset_to_unassociate in self.items():
                 self.new_version.remove_content(queryset_to_unassociate)
-                pb.done = pb.done + queryset_to_unassociate.count()
-                pb.save()
+                pb.increase_by(queryset_to_unassociate.count())
```
bmbouter: Should we remove the auto-throttling from pulpcore's ProgressReport context manager? I'm concerned that we already have a mechanism (auto-throttle saving) that handles this, but we're side stepping that both here and in pulp_rpm.

dralley: We're not actually side stepping that here, the mechanism is inside of ```.save()```, and ```.increase_by()``` calls ```.save()```, so it's functionally equivalent.

In ```pulp_rpm```, yes, I agree.

bmbouter: That's true. Still I'm getting the feeling that auto-throttling is an anti-feature. It's here to help plugin writers to not think about these things, but in practice plugin writers probably need to think about their data processing in batches. The throttling is here to serve plugin writers, but with plugins avoiding it I don't see who its serving. I'm +1 to remove it either in this PR or as a separate issue. It can be in 3.1 also it doesn't matter to me when. It's important to me that we don't have core offering a feature in its plugin API that isn't a good fit for plugin writers.

So can we remove this? and if so shall I file a separate story to?

dralley: Well, ```pulp_file``` was using it, ```pulp_cookbook``` uses it, ```pulp_container``` uses it in one place (but not another), and ```pulp_rpm``` is using it *now*.  I'm not sure the statement that plugins are avoiding it is justified.

With that said, yeah, it does seem a bit hacky.  I could go either way with removing it.

bmbouter: > Well, `pulp_file` was using it, `pulp_cookbook` uses it, `pulp_container` uses it in one place (but not another), and `pulp_rpm` is using it _now_. I'm not sure the statement that plugins are avoiding it is justified.

It's accurate to say that plugins managing their own updates are not relying on pulpcore to provide that feature and avoiding its use.

> 
> With that said, yeah, it does seem a bit hacky. I could go either way with removing it.

I filed https://pulp.plan.io/issues/5855 and started a discussion on pulp-dev here for feedback: https://www.redhat.com/archives/pulp-dev/2019-December/msg00032.html

bmbouter:
I read through again and it looks good to me. Thank you @dralley and @gmbnomis

PR #452, opened 2019-12-11
Add a "not equal" lookup for filters

daviddavis:
fixes #5868
https://pulp.plan.io/issues/5868

### DIFF
--- a/pulpcore/app/models/fields.py
+++ b/pulpcore/app/models/fields.py
@@ -2,7 +2,8 @@ import os
 from gettext import gettext as _
 
 from django.conf import settings
-from django.db.models import FileField
+from django.db.models import FileField, Lookup
+from django.db.models.fields import Field
 
 from pulpcore.app.files import TemporaryDownloadedFile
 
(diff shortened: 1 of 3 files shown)


pulpcore/app/models/fields.py
```diff
@@ -60,3 +61,17 @@ def pre_save(self, model_instance, add):
             file._committed = False
 
         return super().pre_save(model_instance, add)
+
+
+@Field.register_lookup
+class NotEqualLookup(Lookup):
+    # this is copied from https://docs.djangoproject.com/en/3.0/howto/custom-lookups/
+    lookup_name = 'ne'
```
daviddavis: Feedback welcome here on what we want to name this. The django example used `ne`.

daviddavis:
I'm looking at a package called django-rest-framework-filters since it has support for negating filters:

https://github.com/philipn/django-rest-framework-filters#automatic-filter-negationexclusion

daviddavis:
I had a chance to look into django-rest-framework-filters and it seems to work pretty well (at least for searching on NOT conditions). Moreover, you can combine it with other lookups (e.g. `GET /pulp/api/v3/repositories/file/file/name__contains!=RHEL`). 

The only downside is we have to use a prerelease version from September 2018 (1.0.0.dev0) and it looks like there is [no timeline for a 1.0.0 release](https://github.com/philipn/django-rest-framework-filters/issues/287). 

I'm torn and don't have a strong opinion on whether to just go with this PR or use drf-filters.

bmbouter:
I also don't have a strong opinion. If it's working and that project is already dedicated to that type of feature set I lean towards using `django-rest-framework-filters`. @dralley do you see a downside to using `django-rest-framework-filters`?

dralley:
I agree that we should use the 3rd party package with one caveat:

We should make sure it's compatible with the range filters we currently support on some fields

https://github.com/philipn/django-rest-framework-filters#multiwidget-is-incompatible
https://github.com/pulp/pulpcore/blob/master/pulpcore/app/viewsets/base.py#L26

daviddavis:
I plan to merge this on Monday February 10th. django-rest-framework-filters uses `!` for NE fields which is a nonstarter for us since field names with `!` won't work with the bindings.

PR #457, opened 2019-12-17
Generating fixture internally

fabricio-aguiar:
https://pulp.plan.io/issues/5873
closes #5873

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ addons:
     packages:
       - httpie
       - jq
+      - nginx
   # postgres versions provided by el7 RHSCL (lowest supportable version)
   postgresql: '9.6'
 before_install: .travis/before_install.sh
--- a/.travis/nginx.conf
+++ b/.travis/nginx.conf
@@ -1,0 +1,49 @@
+# TODO: Support IPv6.
+# TODO: Configure SSL certificates.
+# TODO: Maybe serve multiple `location`s, not just one.
+
+# Gunicorn docs suggest this value.
+worker_processes 1;
+pid $TRAVIS_BUILD_DIR/tmp/nginx.pid;
+
+error_log $TRAVIS_BUILD_DIR/tmp/error.log;
+
+events {
+    worker_connections 1024;  # increase if you have lots of clients
+    accept_mutex off;  # set to 'on' if nginx worker_processes > 1
+}
+
+http {
+    # If left at the default of 1024, nginx emits a warning about being unable
+    # to build optimal hash types.
+    types_hash_max_size 4096;
+
+    client_body_temp_path $TRAVIS_BUILD_DIR/tmp/nginx_client_body;
+    fastcgi_temp_path     $TRAVIS_BUILD_DIR/tmp/nginx_fastcgi_temp;
+    proxy_temp_path       $TRAVIS_BUILD_DIR/tmp/nginx_proxy_temp;
+    scgi_temp_path        $TRAVIS_BUILD_DIR/tmp/nginx_scgi_temp;
+    uwsgi_temp_path       $TRAVIS_BUILD_DIR/tmp/nginx_uwsgi_temp;
+
+    server {
+        # Gunicorn docs suggest the use of the "deferred" directive on Linux.
+        listen 8000 default_server deferred;
+        server_name localhost;
+
+        # The default client_max_body_size is 1m. Clients uploading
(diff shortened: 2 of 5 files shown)


.travis/before_script.sh
```diff
@@ -40,6 +49,11 @@ if [[ "$TEST" == 'pulp' || "$TEST" == 'performance' ]]; then
     $CMD_PREFIX dnf install -yq lsof which dnf-plugins-core
 fi
 
+sudo kubectl delete pod $PULP_API_POD
+cd $TRAVIS_BUILD_DIR/../pulp-operator
+.travis/pulp-operator-check-and-wait.sh
+cd $TRAVIS_BUILD_DIR/../pulpcore
+
```
mikedep333: Are these changes specific to pulpcore?

If so, they need to be in PRE_BEFORE_SCRIPT or POST_BEFORE_SCRIPT.

If they need to be in pulpcore, they need to be in the plugin-template. There would be a corresponding PR in plugin-template reviewed at the same time as pulpcore's PR.

.travis/pre_before_script.sh
```diff
@@ -0,0 +1,74 @@
...
+        location /fixtures/ {
+	        alias $TRAVIS_BUILD_DIR/tmp/var/www/html/;
+            autoindex on;
+            autoindex_exact_size off;
+	    }
+        error_log $TRAVIS_BUILD_DIR/tmp/error.log;
+		access_log $TRAVIS_BUILD_DIR/tmp/access.log;
+    }
+}
+NGINXCONF
```
daviddavis: Any chance we could put this into a separate file inside the `.travis` directory?

.travis/pre_before_script.sh
```diff
@@ -0,0 +1,74 @@
...
+for i in $(cat Makefile | grep "^fixtures\/file" | sed "s/://g"); do
+    make "$i"
+done
+mv ./fixtures/* ../pulpcore/tmp/var/www/html/
+cd ../pulpcore
+
+nginx -c `pwd`/tmp/etc/nginx.conf
+
+cat .travis/pulp-smash-config.json | \
+    jq 'setpath(["custom","fixtures_origin"]; "http://localhost:8000/fixtures/")' > temp.json
```
daviddavis: Why are you setting fixtures_origin here instead of just setting it in pulp-smash-config.json? Is this to avoid conflicts with plugins?

daviddavis:
This is looking great. Very excited to have this. :+1:

.travis/pre_before_script.sh
fabricio-aguiar: Exactly, if I declare it in pulp-smash-config.json, it would force all plugins to have local fixtures

.travis/pre_before_script.sh
fabricio-aguiar: done!

.travis/nginx.conf
```diff
@@ -0,0 +1,49 @@
...
+        # Path to Pulp's static files.
+        root $TRAVIS_BUILD_DIR/tmp/www;
+
+        location /fixtures/ {
+	        alias $TRAVIS_BUILD_DIR/tmp/var/www/html/;
+            autoindex on;
+            autoindex_exact_size off;
+	    }
+        error_log $TRAVIS_BUILD_DIR/tmp/error.log;
+		access_log $TRAVIS_BUILD_DIR/tmp/access.log;
```
daviddavis: nitpick -- the indentation is inconsistent in this file.

.travis.yml
```diff
@@ -35,6 +35,7 @@ addons:
     packages:
       - httpie
       - jq
+      - nginx
```
mikedep333: Where is the corresponding change to plugin-template?

Without it, it would be blown away by it being run.

fabricio-aguiar:
> It looks like we should refactor this into an option for plugin-template. Only a few variables would be involved: A boolean generate local fixtures (which affects nginx in .travis.yml) in template_config.yml, blocks of code based in before_script.sh if it is true, and we'd change the pulp-smash-config.json based on the plugin name.
> 
> Note that you seemed to indicate that pulp-smash-config.json is part of plugin-template, but I do not see it there.

pulp-smash-config.json from pulpcore is used in all plugins:
https://github.com/pulp/plugin_template/blob/master/templates/travis/.travis/before_script.sh.j2#L23-L27

PR #460, opened 2019-12-18
List added/removed content according to a given RepositoryVersion

fabricio-aguiar:
https://pulp.plan.io/issues/5706
closes #5706

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -105,7 +105,7 @@ class Repository(MasterModel):
                 # now add any content that's in the base_version but not in version
                 version.add_content(base_version.content.exclude(pk__in=version.content))
 
-            if Task.current:
+            if Task.current():
                 resource = CreatedResource(content_object=version)
                 resource.save()
             return version
@@ -396,6 +396,17 @@ class RepositoryVersion(BaseModel):
         get_latest_by = 'number'
         ordering = ('number',)
 
+    def _relationships(self):
+        """
+        Returns a set of repository_content for a repository version
+
+        Returns:
+            django.db.models.QuerySet: The repository_content that is contained within this version.
+        """
+        return RepositoryContent.objects.filter(
+            repository=self.repository, version_added__number__lte=self.number
+        ).exclude(version_removed__number__lte=self.number)
+
     @property
     def content(self):
         """
@@ -415,12 +426,7 @@ class RepositoryVersion(BaseModel):
             >>>     ...
             >>>
         """
-        relationships = RepositoryContent.objects.filter(
-            repository=self.repository, version_added__number__lte=self.number
-        ).exclude(
-            version_removed__number__lte=self.number
-        )
-        return Content.objects.filter(version_memberships__in=relationships)
+        return Content.objects.filter(version_memberships__in=self._relationships())
 
     def _content_old(self):
         """
@@ -449,19 +455,39 @@ class RepositoryVersion(BaseModel):
         """
         Artifact.objects.filter(content__pk__in=self.content)
 
-    def added(self):
+    def added(self, base_version=None):
         """
+        Args:
+            base_version (pulpcore.app.models.RepositoryVersion): an optional repository version
+
         Returns:
             QuerySet: The Content objects that were added by this version.
         """
-        return Content.objects.filter(version_memberships__version_added=self)
+        if not base_version:
+            return Content.objects.filter(version_memberships__version_added=self)
 
-    def removed(self):
+        return Content.objects.filter(
+            version_memberships__in=self._relationships()
+        ).exclude(
+            version_memberships__in=base_version._relationships()
+        )
+
+    def removed(self, base_version=None):
         """
+        Args:
+            base_version (pulpcore.app.models.RepositoryVersion): an optional repository version
+
         Returns:
             QuerySet: The Content objects that were removed by this version.
         """
-        return Content.objects.filter(version_memberships__version_removed=self)
+        if not base_version:
+            return Content.objects.filter(version_memberships__version_removed=self)
+
+        return Content.objects.filter(
+            version_memberships__in=base_version._relationships()
+        ).exclude(
+            version_memberships__in=self._relationships()
+        )
 
     def contains(self, content):
         """
--- a/pulpcore/tests/unit/models/test_repository.py
+++ b/pulpcore/tests/unit/models/test_repository.py
@@ -1,0 +1,65 @@
+from itertools import compress
+
+from django.test import TestCase
+from pulpcore.plugin.models import Content, Repository, RepositoryVersion
+
+
+class RepositoryVersionTestCase(TestCase):
+
+    def setUp(self):
+        self.repository = Repository.objects.create()
+        self.repository.CONTENT_TYPES = [Content]
+        self.repository.save()
+
+        contents = []
+        for _ in range(0, 4):
+            contents.append(Content(pulp_type="core.content"))
+
+        Content.objects.bulk_create(contents)
+        self.pks = [c.pk for c in contents]
+
+    def test_add_and_remove_content(self):
+        contents = Content.objects.filter(pk__in=self.pks)
+        with self.repository.new_version() as version1:
+            version1.add_content(contents)  # v1 == all contents
+
+        to_remove = contents[0:2]
+        with self.repository.new_version() as version2:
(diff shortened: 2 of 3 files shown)


CHANGES/5706.feature
```diff
@@ -0,0 +1 @@
+List added/removed content according to a given RepositoryVersion
```
daviddavis: This is quite ambiguous. What about, "Added an optional parameter base_version to RepositoryVersion add() and removed() methods."

Also, this should probably be in `CHANGE/plugin_api/`.

daviddavis:
For extra bonus points, what about adding some unit tests?

pulpcore/app/models/repository.py
```diff
@@ -449,19 +449,33 @@ def artifacts(self):
...
+            base_version (pulpcore.app.models.RepositoryVersion): an optional repository version
+
         Returns:
             QuerySet: The Content objects that were added by this version.
         """
-        return Content.objects.filter(version_memberships__version_added=self)
+        if not base_version:
+            return Content.objects.filter(version_memberships__version_added=self)
+        versions = RepositoryVersion.objects.filter(
+            number__in=range(base_version.number + 1, self.number + 1), repository=self.repository)
```
daviddavis: I thought through this logic a bit and I think there is a problem. Consider the case where you have versions 1, 2, 3, 4 and you call added(1) on version 4. If there's a unit that was added in 2 and removed in 3, it'll be returned which is not correct. 

The only way I can think of off the top of my head is to get the content for version 4 and version 1 and subtract.

bmbouter:
@gmbnomis what do you think about this?

pulpcore/app/models/repository.py
```diff
@@ -449,19 +449,59 @@ def artifacts(self):
         """
         Artifact.objects.filter(content__pk__in=self.content)
 
-    def added(self):
+    def added(self, base_version=None):
```
gmbnomis: This looks complicated and may take a long time to run if there are many version between the `base_version` and the current version. I was thinking about:

```

    def _relationships(self):
        return RepositoryContent.objects.filter(
            repository=self.repository, version_added__number__lte=self.number
        ).exclude(version_removed__number__lte=self.number)

    @property
    def content(self):
        return Content.objects.filter(version_memberships__in=self._relationships())


    def added(self, base_version=None):
        if not base_version:
            return Content.objects.filter(version_memberships__version_added=self)

        return Content.objects.filter(
            version_memberships__in=self._relationships()
        ).exclude(
            version_memberships__in=base_version._relationships()
        )

    def removed(self, base_version=None):
        if not base_version:
            return Content.objects.filter(version_memberships__version_removed=self)

        return Content.objects.filter(
            version_memberships__in=base_version._relationships()
        ).exclude(
            version_memberships__in=self._relationships()
        )
```
I hadn't the time to look into the cost of the queries in your implementation and in the one from above.

The queries above may be costly (we need a join with the repository version table to get the version numbers to do the range query on version numbers). However, I don't know how this relates to your queries and we do .content queries all the time anyway...

pulpcore/tests/unit/models/test_repository.py
```diff
@@ -0,0 +1,58 @@
...
+
+        self.assertTrue(self.pks[0] not in removed_pks)
+        self.assertTrue(self.pks[1] in removed_pks)
+        self.assertTrue(self.pks[2] not in removed_pks)
+        self.assertTrue(self.pks[3] not in removed_pks)
+
+        self.assertTrue(self.pks[0] in added_pks)
+        self.assertTrue(self.pks[1] not in added_pks)
+        self.assertTrue(self.pks[2] in added_pks)
+        self.assertTrue(self.pks[3] in added_pks)
```
gmbnomis: I would prefer to have additional comparisons in the test e.g. version 3 against all other versions.

pulpcore/tests/unit/models/test_repository.py
```diff
@@ -0,0 +1,58 @@
...
+
+        added_pks = version3.added(version0).values_list('pk', flat=True)
+        removed_pks = version3.removed(version0).values_list('pk', flat=True)
+
+        self.assertTrue(self.pks[0] not in removed_pks)
+        self.assertTrue(self.pks[1] in removed_pks)
+        self.assertTrue(self.pks[2] not in removed_pks)
+        self.assertTrue(self.pks[3] not in removed_pks)
+
+        self.assertTrue(self.pks[0] in added_pks)
```
gmbnomis: Proposal: It's less explicit, but `self.assertCountEqual(added_pks, compress(self.pks, [1, 0, 1, 1]))` is more concise (using compress from itertools).

pulpcore/app/models/repository.py
```diff
@@ -105,7 +105,7 @@ def new_version(self, base_version=None):
                 # now add any content that's in the base_version but not in version
                 version.add_content(base_version.content.exclude(pk__in=version.content))
 
-            if Task.current:
+            if Task.current():
```
gmbnomis: Oh, now I know why I had to work around this in my unit tests. Thank you!

pulpcore/tests/unit/models/test_repository.py
```diff
@@ -0,0 +1,58 @@
...
+        self.assertEqual(version2.added().count(), 0)
+        self.assertEqual(version3.added().count(), 1)
+
+        self.assertEqual(version0.removed().count(), 0)
+        self.assertEqual(version1.removed().count(), 0)
+        self.assertEqual(version2.removed().count(), 2)
+        self.assertEqual(version3.removed().count(), 0)
+
+        self.assertEqual(version3.added(version0).count(), 3)
+        self.assertEqual(version3.removed(version0).count(), 1)
```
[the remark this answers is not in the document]
fabricio-aguiar: I believe I misunderstood the problem, I specifically designed this solution to show that one element was removed during all the versions.

pulpcore/app/models/repository.py
fabricio-aguiar: definitely your queries are better, thank you!

pulpcore/app/models/repository.py
```diff
@@ -449,19 +449,48 @@ def artifacts(self):
         """
         Artifact.objects.filter(content__pk__in=self.content)
 
-    def added(self):
+    def _relationships(self):
```
gmbnomis: I think we should use `_relationships()` in the `content` property as well.

pulpcore/app/models/repository.py
```diff
@@ -449,19 +449,48 @@ def artifacts(self):
...
         """
+        Returns:
+            QuerySet: The Content objects that were added until this version.
```
gmbnomis: Docstring needs to be updated

gmbnomis:
@fabricio-aguiar @daviddavis @bmbouter I have written two different fixes for https://pulp.plan.io/issues/5707 based on this fix for 5706. I would love to get feedback on your view on the pros and cons of these implementations, see https://pulp.plan.io/issues/5707#note-8.

@fabricio-aguiar Thank you for the test rig added by this PR!

pulpcore/app/models/repository.py
```diff
@@ -396,6 +396,17 @@ class Meta:
         get_latest_by = 'number'
         ordering = ('number',)
 
+    def _relationships(self):
```
daviddavis: This is the only hangup I have with this PR. `_relationships` sounds generic. Would there be a better name?

gmbnomis:
Should this be cherry picked? While it isn't a bugfix, it will be necessary to fix `remove_duplicates()`. https://github.com/pulp/pulpcore/blob/7c1aee68391d19befed1ad33c30d94857b7d7eab/pulpcore/plugin/repo_version_utils.py#L34

pulpcore/app/models/repository.py
bmbouter: What about `_find_content`? @gmbnomis and @daviddavis what do you think?

daviddavis: Hmm, it's not really Content though. It's RepositoryContent. What about `_repository_content`?

bmbouter: > Hmm, it's not really Content though. It's RepositoryContent. What about `_repository_content`?

This sounds good to me.

bmbouter:
I think we do want to cherry pick it because we're planning to fix bugs on top of it in 3.0.z like https://pulp.plan.io/issues/5707 ? @daviddavis wdyt?

PR #467, opened 2019-12-27
Use libeatmydata to speed up travis

fabricio-aguiar:
**Does not seem to be effective**
https://pulp.plan.io/issues/5871
closes #5871

### DIFF
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -19,10 +19,18 @@ export FUNC_TEST_SCRIPT=$TRAVIS_BUILD_DIR/.travis/func_test_script.sh
 # this script.
 export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
 
+if [ -r /usr/lib/libeatmydata/libeatmydata.so ]; then
+  # much faster package installation
+  export LD_PRELOAD='/usr/lib/libeatmydata/libeatmydata.so'
+elif [ -r /usr/lib/*/libeatmydata.so ]; then
+  # much faster package installation
+  export LD_PRELOAD='/usr/$LIB/libeatmydata.so'
+fi
+
 if [ "$TEST" = 'docs' ]; then
-  
+
     export PULP_CONTENT_ORIGIN=http://$(hostname):24816
-  
+
 
   cd docs
   make html
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,7 @@ addons:
       - httpie
       - jq
       - nginx
+      - eatmydata
   # postgres versions provided by el7 RHSCL (lowest supportable version)
   postgresql: '9.6'
 before_install: .travis/before_install.sh
--- a/CHANGES/5871.misc
+++ b/CHANGES/5871.misc
@@ -1,0 +1,1 @@
+Use libeatmydata to speed up travis


fabricio-aguiar:
**Does not seem to be effective**

bmbouter:
@fabricio-aguiar am I reading correctly that this did not reduce the Travis runtime?

fabricio-aguiar:
> @fabricio-aguiar am I reading correctly that this did not reduce the Travis runtime?

yes, I just opened this PR to be a record that it did not improve travis performance

bmbouter:
@fabricio-aguiar thank you. Want to close this PR and also the associated Redmine issue?

fabricio-aguiar:
Does not seem to be effective

.travis/script.sh
```diff
@@ -19,10 +19,18 @@ export FUNC_TEST_SCRIPT=$TRAVIS_BUILD_DIR/.travis/func_test_script.sh
 # this script.
 export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
 
+if [ -r /usr/lib/libeatmydata/libeatmydata.so ]; then
+  # much faster package installation
+  export LD_PRELOAD='/usr/lib/libeatmydata/libeatmydata.so'
+elif [ -r /usr/lib/*/libeatmydata.so ]; then
+  # much faster package installation
+  export LD_PRELOAD='/usr/$LIB/libeatmydata.so'
```
mikedep333: Sorry for not reviewing this earlier.

Is $LIB actually set? I just ran docker containers for ubuntu:xenial & ubuntu:bionic, and it was not.

We can also just hardcode the Ubuntu/Debian x86_64 path for now:
export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libeatmydata.so

mikedep333:
Also, I think this small commit would only speed up travis a little bit, so be careful when analyzing results. We'd have to put that code block in more scripts that have disk I/O. And any code running in containers would need implementation as well.

It might be best to do initial performance testing by doing a travis debug in your own fork.

(Of course, this may all be for nothing if travis is already doing the equivalent of libeatmydata at the hypervisor level.)

dralley:
Is Fedora 32 going to break this trick?  

```
When we build the Python interpreter with the -fno-semantic-interposition compiler/linker flag, we can achieve a performance gain of 5% to 27% depending on the workload. Link time optimizations and profile guided optimizations also have a greater impact when python3 is built this way.

As a negative side effect, it disables the LD_PRELOAD feature: it's no longer possible to override symbols in libpython with LD_PRELOAD. 
```

https://fedoraproject.org/wiki/Changes/PythonNoSemanticInterpositionSpeedup

.travis/script.sh
fabricio-aguiar: it was not set, I made some changes, but performance still pretty similar:
https://travis-ci.com/fabricio-aguiar/pulpcore/builds/146811533
code:
https://github.com/pulp/pulpcore/compare/master...fabricio-aguiar:5871
got the USE_CHROOT from here:
https://github.com/sunny11mind/NumPy/blob/b48dc92402b1e928b6dc26214ca6ae6d445d50dd/.travis.yml#L16-L17

PR #471, opened 2020-01-02
Adds a note on where django-admin is found.

bmbouter:
The note identifies django-admin's location for virtualenv, system-wide,
and Katello environments.

https://pulp.plan.io/issues/5859
closes #5859

### DIFF
--- a/docs/installation/instructions.rst
+++ b/docs/installation/instructions.rst
@@ -81,8 +81,18 @@ PyPI Installation
 
 9. Run Django Migrations::
 
-   $ django-admin migrate --noinput
-   $ django-admin reset-admin-password --password admin
+    $ django-admin migrate --noinput
+    $ django-admin reset-admin-password --password admin
+
+.. note::
+
+    ``django-admin`` is provided by Django and is available wherever Django is installed. For
+    virtualenv based installs, that is from the activated virtualenv. For system-wide Django
+    installations, ``django-admin`` should be on your system path.
+
+    Katello users use an alternate binary, ``python3-django-admin`` which is available on the system
+    path. In Katello environments, the ``django-admin`` command is for the Django environment of
+    Pulp 2.
 
 10. Collect Static Media for live docs and browsable API::
 
--- a/CHANGES/5859.doc
+++ b/CHANGES/5859.doc
@@ -1,0 +1,1 @@
+Clarifies where ``django-admin`` is found. Also notes the alternate name Katello environments use.


bmbouter:
Thanks @daviddavis !

@evgeni can you let me know what you think also? I'm not 100% sure on the binary name for example.

docs/installation/instructions.rst
```diff
@@ -81,8 +81,18 @@ PyPI Installation
...
+    $ django-admin migrate --noinput
+    $ django-admin reset-admin-password --password admin
+
+.. note::
+
+    ``django-admin`` is provided by Django and is available wherever Django is installed. For
+    virtualenv based installs that is ``/path/to/your/pulpvenv/bin/django-admin``. For system-wide
+    Django installations, ``django-admin`` should be on your system path.
+
+    Katello users use an alternate binary, ``pulp3-django-admin`` which is available on the system
```
evgeni:
```suggestion
    Katello users use an alternate binary, ``python3-django-admin`` which is available on the system
```

docs/installation/instructions.rst
```diff
@@ -81,8 +81,18 @@ PyPI Installation
...
 
-   $ django-admin migrate --noinput
-   $ django-admin reset-admin-password --password admin
+    $ django-admin migrate --noinput
+    $ django-admin reset-admin-password --password admin
+
+.. note::
+
+    ``django-admin`` is provided by Django and is available wherever Django is installed. For
+    virtualenv based installs that is ``/path/to/your/pulpvenv/bin/django-admin``. For system-wide
```
evgeni: When you `activate` a virtualenv, its `bin` folder gets added to `PATH`, so you don't *need* to specify the full path (even tho you can, and that will execute stuff *inside* the venv even if you didn't `activate`). Not sure we should be in the business to teach users how to use virtual envs, but thought I'd still raise it here.

ekohl:
>     2\. We don't mention that you need to set `DJANGO_SETTINGS_MODULE=pulpcore.app.settings` for `django-admin` to actually work (well, we mention it in `docs/contributing/architecture/app-layout.rst`, but that's not in the default "how to install path" of the user).

In addition to that, there's also `PULP_SETTINGS`. In the installer we set this to `/etc/pulp/settings.py` and for every django-admin invocation.

docs/installation/instructions.rst
bmbouter: Thank you! Change made in next push.

docs/installation/instructions.rst
bmbouter: I agree, I made a change. Let me know what you think.

bmbouter:
> Corrected the binary name and added a comment inline.
> 
> Two things I'd like to raise more globally:
> 
>     1. `django-admin` is also mentioned in the following files:
>        
>        * docs/components.rst
>        * docs/contributing/tests.rst
>        * docs/installation/authentication.rst
>        * docs/installation/instructions.rst
>        * docs/plugins/api-reference/profiling.rst
>        * docs/plugins/plugin-writer/concepts/index.rst
>        * docs/plugins/plugin-writer/concepts/subclassing/models.rst
>        * docs/plugins/reference/how-plugins-work.rst
>          What should happen with those?
I also was worried about this. I made some changes and then backed them out. I decided one place is right for now because a) Half of these are not user facing, and b) all users will have gone through the instructions first at least to apply migrations.
> 
>     2. We don't mention that you need to set `DJANGO_SETTINGS_MODULE=pulpcore.app.settings` for `django-admin` to actually work (well, we mention it in `docs/contributing/architecture/app-layout.rst`, but that's not in the default "how to install path" of the user).

This is a compelling point. I was thinking that setting these things for services is ok because the installer takes of systemd for example, but this means that all users currently need to apply `DJANGO_SETTINGS_MODULE` to run migrations for example. Let's get some other input from others also.

bmbouter:
I rewrote the issue https://pulp.plan.io/issues/5859 once again, to use the wrapper. With the DJANGO_SETTINGS_MODULE that seems like something we should provide.

Here is a question about what it should be named:  https://pulp.plan.io/issues/5859#note-15

bmbouter:
I'm closing this issue so that we can bring forward 5859 as a new feature. We're trying to have it provide autocompletion also which I'm not sure how to do at the moment. Suggestions are welcome.

PR #474, opened 2020-01-03
Allow storages.backends.azure_storage.AzureStorage as value DEFAULT_F…

makesse:
In order to use Azure Blob storage as location for the artifacts stored by Pulp 'storages.backends.azure_storage.AzureStorage' needs to be added as a valid value for this setting. I will add the required instructions on how to configure Azure storage next to the already available S3 documentation


makesse:
Well, that became a commit hell quickly...

PR #479, opened 2020-01-07
Remove redis-py dep, update setuptools dep

dralley:
[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,15 +14,14 @@ requirements = [
     'djangorestframework-queryfields~=1.0.0',
     'drf-nested-routers~=0.91.0',
     'drf-yasg~=1.17.0',
+    'dynaconf>=2.2,<2.3',
     'gunicorn>=19.9,<20.1',
+    'pygtrie~=2.3.2',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -18,8 +18,7 @@
     'psycopg2>=2.7,<2.9',
     'PyYAML>=5.1.1,<5.4.0',
     'rq>=1.1,<1.3',
-    'redis~=3.1.0',
-    'setuptools>=39.2.0,<44.1.0',
+    'setuptools>=39.2.0',
```
dralley: Relying on an upwardly-bounded version of setuptools seems like a bad idea since I think it gets updated along with pip.

setup.py
```diff
@@ -14,15 +14,14 @@
     'djangorestframework-queryfields~=1.0.0',
     'drf-nested-routers~=0.91.0',
     'drf-yasg~=1.17.0',
+    'dynaconf>=2.2,<2.3',
```
dralley: alphabetize

PR #482, opened 2020-01-07
Avoid superfluous RepositoryContent records

gmbnomis:
When adding and removing the same content in a repository version, don't
update the RepositoryContent record (resulting in an superfluous
RepositoryContent for which version_added == version_removed == current
version).  Instead, just delete the existing record.  This fixes two
problems:

1. Such content is shown as added and removed.
2. Re-adding the content again in the same version leads to a
   IntegrityError because the uniqueness constraint is violated.

Handle the "remove then re-add" case analogously.

https://pulp.plan.io/issues/5707
closes #5707

### DIFF
--- a/pulpcore/tests/unit/models/test_repository.py
+++ b/pulpcore/tests/unit/models/test_repository.py
@@ -12,16 +12,16 @@ class RepositoryVersionTestCase(TestCase):
         self.repository.save()
 
         contents = []
-        for _ in range(0, 4):
+        for _ in range(0, 5):
             contents.append(Content(pulp_type="core.content"))
 
         Content.objects.bulk_create(contents)
         self.pks = [c.pk for c in contents]
 
     def test_add_and_remove_content(self):
-        contents = Content.objects.filter(pk__in=self.pks)
+        contents = Content.objects.filter(pk__in=self.pks[:4])
         with self.repository.new_version() as version1:
-            version1.add_content(contents)  # v1 == all contents
+            version1.add_content(contents)  # v1 == four content units 0-4
 
         to_remove = contents[0:2]
         with self.repository.new_version() as version2:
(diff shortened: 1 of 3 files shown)


pulpcore/tests/unit/models/test_repository.py
```diff
@@ -63,3 +63,94 @@ def test_add_and_remove_content(self):
...
+        self.assertCountEqual(added_pks, compress(self.pks, added), added_pks)
+        self.assertCountEqual(removed_pks, compress(self.pks, removed), removed_pks)
+
+    def test_normalize_repository_content(self):
+        """Verify that add_content/delete_content normalize RepositoryContent records properly."""
+        with self.repository.new_version() as version1:
+            version1.add_content(self.content_qs(self.pks[:5]))  # v1 == content 0-4
+
+        # v2 content:
+        # 0 leave as is
```
daviddavis: These numbers confused me until I realized that they're indexes. Mind updating this to `# pks[0] ...`?

daviddavis: Actually sorry, I see that these are the indexes for each of the assertions below. This test is long and rather hard to follow. Any chance you could break it up into smaller tests?

daviddavis:
This looks great. Thanks for the PR. Should we close out https://github.com/pulp/pulpcore/pull/482 ?

pulpcore/tests/unit/models/test_repository.py
gmbnomis: Yes, you have a point there. This test was for solution 2/2a originally and I wanted to ensure that all add/remove "histories" are handled correctly by normalization. This is not necessary in here, since there is no need for normalization when saving the version.

I think I will rework this into:

1. Simple test: add, then remove
2. Simple test: remove, then add
3. More complex test with content that is not touched at all, content that is just added or removed and content that is added/removed repeatedly. (to verify that there are no interaction effects and to verify that adding/removing multiple content units with differing "normalization needs" at once works correctly.)

3 will probably be still a little complex, but I will add more documentation.

wdyt?

daviddavis: That sounds great. :+1: from me.

PR #486, opened 2020-01-08
Copy files when they originate out of MEDIA_ROOT

### LINKED ISSUE #5941: Task metrics reporting an incorrect number of unblocked tasks
**Describe the bug**
If a number of unblocked tasks are in the "complete state", like `canceled` or `failed`, somehow they will be showed in the task metrics, leading to some confusing conclusions.

**To Reproduce**
Create some tasks with different exclusive resources, like different `domain`. Make them fail, or cancel them.
It's gonna show a number of unblocked tasks waiting in the queue, which is not true.

**Expected behavior**
It's  expected to only show unblocked tasks that are waiting in the queue to be executed.

bmbouter:
Django would move files in cases where the source was a file:///
repository.

https://pulp.plan.io/issues/5941
closes #5941

https://pulp.plan.io/issues/5870
closes #5870

### DIFF
--- a/pulpcore/app/models/storage.py
+++ b/pulpcore/app/models/storage.py
@@ -1,74 +1,26 @@
-import errno
 import os
 from uuid import uuid4
 
 from django.conf import settings
-from django.core.files import File
+from django.core.files import locks
+from django.core.files.move import file_move_safe
 from django.core.files.storage import FileSystemStorage
 
 
 class FileSystem(FileSystemStorage):
     """
-    Django's FileSystemStorage with modified save() behavior
+    Django's FileSystemStorage with modified _save() and get_available_name behaviors
 
-    The FileSystemStorage backend provided by Django is designed to work with File,
-    TemporaryUploadedFile, and InMemoryUploadedFile objects. The latter two inherit from the first.
-    The type of File determines how the storage backend places it in its final destination. The
-    Artifact upload API always uses a TemporaryUploadedFile, which writes the file to a temporary
-    location on disk during the upload. Here is how FileSystemStorage saves a file to its
-    destination for the two File types we are interested in:
-
-    TemporaryUploadedFile
-    ------------------------------
-    1) is name available?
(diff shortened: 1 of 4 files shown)


pulpcore/app/models/storage.py
```diff
@@ -1,93 +1,38 @@
...
-
-    The non-atomic nature of 3b makes it possible for corrupted Artifacts to be saved if
-    /var/lib/pulp/tmp and /var/lib/pulp/artifact are on separate filesystems. An interrupted save()
-    operation will result in a partial file being placed into /var/lib/pulp/artifact. In such
-    situations, Pulp will not be able to replace the file with the correct version in the
-    future and the user will have to manually remove the partial file.
+    Django's FileSystemStorage with modified _save() and get_available_name behaviors
+
+    The _save() will try to reflink the file first, and then fall back to copy similar to what
+    `cp --reflink=auto` does.
```
daviddavis: This needs to be updated.

bmbouter: good call. fixed in next push.

pulpcore/app/models/storage.py
```diff
@@ -97,20 +43,56 @@ def save(self, name, content, max_length=None):
...
+                            mode = 'wb' if isinstance(chunk, bytes) else 'wt'
+                            _file = os.fdopen(fd, mode)
+                        _file.write(chunk)
+                finally:
+                    locks.unlock(fd)
+                    if _file is not None:
+                        _file.close()
+                    else:
+                        os.close(fd)
+        except FileExistsError:
```
dkliban: Can you refactor this except so that it is more closely located to the code that emits the exception?

bmbouter: yes! I pushed that just now.

bmbouter: That fails on the open due to the 0o666 open mode, so the exception handler being down here is needed.

PR #489, opened 2020-01-13
Adding django-rest-framework-filters for NOT filters

daviddavis:
https://github.com/philipn/django-rest-framework-filters

fixes #5868
https://pulp.plan.io/issues/5868

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ requirements = [
     'backoff',
     'coreapi~=2.3.3',
     'Django~=2.2.3',  # LTS version, switch only if we have a compelling reason to
-    'django-filter~=2.2.0',
+    'djangorestframework-filters==1.0.0.dev0',
     'djangorestframework~=3.10.2',
     'djangorestframework-queryfields~=1.0.0',
     'drf-nested-routers~=0.91.0',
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -9,7 +9,7 @@
     'backoff',
     'coreapi~=2.3.3',
     'Django~=2.2.3',  # LTS version, switch only if we have a compelling reason to
-    'django-filter~=2.2.0',
+    'djangorestframework-filters==1.0.0.dev0',
```
daviddavis: I tried using `>=` here but it didn't work (it installed 0.11.1). 

It looks [doubtful about whether we'll see a 1.0 release](https://github.com/philipn/django-rest-framework-filters/issues/287) which makes me nervous.

daviddavis:
Waiting on a response to https://github.com/philipn/django-rest-framework-filters/issues/287#issuecomment-574312703

dralley:
@daviddavis How easily could the not-equal filter be vendored?

daviddavis:
@dralley I don't think it would be too hard although I haven't vendored anything in Python before. I could look into it.

My concern though is having to maintain the django-rest-framework-filters code. Like right now, [it looks like it doesn't support django 3](https://github.com/philipn/django-rest-framework-filters/issues/326).

daviddavis:
@dralley brought up a good point. We can't name fields with `!` since these field names are used in the bindings and `!` is a special character in many languages (including Python).

PR #492, opened 2020-01-14
REMOTE_USER auth shouldn't trigger csrf rejects

bmbouter:
Move all authentication fully to DRF. We were incorrectly configuring
webserver auth support in django and not DRF. This ports the setting
`REMOTE_USER_ENVIRON_NAME` to work with DRF instead of Django.

It adjusts the settings.py so there are removal claims even though this
will likely go into a z-release. It's an important FYI for the user.

https://pulp.plan.io/issues/5955
closes #5955

### DIFF
--- a/pulpcore/app/middleware.py
+++ b/pulpcore/app/middleware.py
@@ -1,7 +1,0 @@
-from django.contrib.auth.middleware import RemoteUserMiddleware
-
-from pulpcore.app import settings
-
-
-class PulpRemoteUserMiddleware(RemoteUserMiddleware):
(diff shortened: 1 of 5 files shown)


pulpcore/app/middleware.py
```diff
@@ -1,7 +0,0 @@
-from django.contrib.auth.middleware import RemoteUserMiddleware
-
```
bmbouter: Maybe we should keep this file even though we don't use it a user may want to?

bmbouter: Actually let's not. Pulp uses DRF and we should focus on the mechanisms it provides.

PR #498, opened 2020-01-15
Update travis to fix Required PR failure for closed PRs

daviddavis:
Required PR: https://github.com/pulp/pulp-smash/pull/1228

This PR requires a pulp-smash PR ^^^ as a proof of concept. If the PR passes, then we should be good.

[noissue]


daviddavis:
Example of the failure (before applying this fix): https://travis-ci.com/pulp/pulpcore/jobs/275851298

PR #512, opened 2020-01-21
Add an interface for a signing service

lubosmj:
closes #5946
https://pulp.plan.io/issues/5946


docs/plugins/reference/metadata-signing.rst
```diff
@@ -0,0 +1,41 @@
+.. _metadata-signing:
+
+Metadata Signing
+================
+
```
dkliban: Plugin writers wishing to enable users to sign metadata need to add a 'metadata_signing_service' field to their implementation of a Repository. This field should be exposed to REST API users so they can specify which signing service should be used to sign the metadata when creating a publication. That foreign key can be used to retrieve the Signing Service during the publish task.

docs/plugins/reference/metadata-signing.rst
```diff
@@ -0,0 +1,47 @@
...
+``AsciiArmoredDetachedSigningService``. Learn more about the creation of the signing script in the section
+:ref:`administration`.
+
+The following procedure may be taken into account for the plugin writers:
+
+    1. Retrieve a desired signing script from a database:
+
+       .. code-block:: python
+
+           from pulpcore.app.models.content import AsciiArmoredDetachedSigningService
```
bmbouter: Can the example show a user's Repository, e.g. `FileRepository` that has a `ForeignKey` to this model. Then the database access example is taking the Repository instance for the operation and accessing it's `metadata_signing` attribute (the FK) instead of querying with `objects.get()`.

pulpcore/app/models/content.py
```diff
@@ -370,12 +372,55 @@ class Meta:
 
 class SigningService(BaseModel):
```
bmbouter: The SigningService model should not be instantiated directly, so I think we should make it an abstract model.

pulpcore/app/models/content.py
```diff
@@ -370,12 +372,55 @@ class Meta:
...
     """
-    A model used for producing signed Artifacts.
+    A model used for producing signed artifacts.
```
bmbouter: s/signed artifacts/signatures/ The term Artifacts means a formal thing so I think it's clearer to state it produces signatures.

pulpcore/app/models/content.py
```diff
@@ -370,12 +372,55 @@ class Meta:
...
 
     Fields:
-        name (models.TextField): A unique name describing a script used for the signing.
+        name (models.TextField): A unique name describing a script used for a signing.
```
bmbouter: s/a signing/signing/

pulpcore/app/models/content.py
```diff
@@ -370,12 +372,55 @@ class Meta:
...
+    def sign(self, filename):
+        """
+        Sign a passed file.
+
+        Subclasses are required to implement this method. The method shall invoke an external
+        signing script that creates a signature for the passed filename.
+
+        Args:
+            filename (str): A relative path to a file which is intended to be signed.
+
```
bmbouter: Can a `Returns:` section be added for this docstring. Plugin writers will use that to know what to expect.

bmbouter: Oh actually I see it defined on the subclass. Ignore this comment.

pulpcore/app/models/content.py
```diff
@@ -370,12 +372,55 @@ class Meta:
...
+        Raises:
+            RuntimeError: If a return code of the script is not equal to 0, indicating a failure.
+
+        Returns:
+            A dictionary in the following format::
+
+                {"file": "signed_file.xml", "signature": "signed_file.asc", "key": "public.key"}
+
+        """
+        completed_process = subprocess.run(
```
bmbouter: Since `script` is defined in the `SigningService` class, we should move the calling of that script in `sign()` also to that class. This would replace the `NotImplentedError` which I think is ok. The docstring Return can move there too. Then this class can just be:

```
class AsciiArmoredDetachedSigningService(SigningService):
    """
    A model used for creating detached ascii armored signatures.
    """

    def save():
        """put validation stuff here"""
```

lubosmj: I believe, in the future, the output format of signing scripts will be different for different content types. Now, we can sign only metadata and we do expect the output of the script to be written to stdout. Is not this thing going to be changed by any means? If so, I can proceed with proposed changes. I think that every signing service should handle the signing process by itself.

bmbouter: After thinking more about this I agree with you. Let's leave the script calling in `AsciiArmoredDetachedSigningService` like you already have it. Thanks!

PR #513, opened 2020-01-21
Re-pin redis version to fix freezing issue

dralley:
[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,7 @@ requirements = [
     'pygtrie~=2.3.2',
     'psycopg2>=2.7,<2.9',
     'PyYAML>=5.1.1,<5.4.0',
+    'redis~=3.1.0',  # pinned due to https://github.com/andymccurdy/redis-py/issues/1138
     'rq>=1.1,<1.3',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -19,6 +19,7 @@
     'pygtrie~=2.3.2',
     'psycopg2>=2.7,<2.9',
     'PyYAML>=5.1.1,<5.4.0',
+    'redis~=3.1.0',
```
bmbouter: Can we have an issue tracking the defect (even a 1-liner to have the issue number), and have a comment here linking to it?

mdellweg:
@gmbnomis may have found a solution: https://github.com/andymccurdy/redis-py/issues/1138#issuecomment-576924979

setup.py
mdellweg: https://pulp.plan.io/issues/6038

bmbouter: @dralley can you add the comment here also? I'm concerned that without it someone could upgrade and subtly re-introduce the issue. Wdyt?

mdellweg:
Thank you!

PR #514, opened 2020-01-22
WIP - Validate Content data in the Stages API

fabricio-aguiar:
https://pulp.plan.io/issues/5927
closes #5927

### DIFF
--- a/pulpcore/plugin/stages/content_stages.py
+++ b/pulpcore/plugin/stages/content_stages.py
@@ -1,11 +1,52 @@
 from collections import defaultdict
+from itertools import chain
 
+from django.apps import apps
 from django.db import IntegrityError, transaction
 from django.db.models import Q
 
 from pulpcore.plugin.models import ContentArtifact
+from pulpcore.plugin.serializers import SingleArtifactContentSerializer
+from pulpcore.app import files
 
 from .api import Stage
+
+
+def validate_content(content):
+    """Validate Content."""
+    data = {}
+    opts = content._meta
+    fields = chain(
+        opts.concrete_fields, opts.private_fields, opts.many_to_many, opts.related_objects
+    )
+    label = content._meta.app_label
+    model = content._meta.model
+    serializer = apps.get_app_config(label).model_serializers[model]
+    serializer_fields = serializer.Meta.fields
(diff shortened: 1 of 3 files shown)


daviddavis:
Your code looks good. My concern here though is that this code is magical in that Content is automatically matched to a serializer without any input from the plugin writer. It's possible that plugin may have more than one serializer for a Content model depending on different contexts (like in pulp_rpm we have two for Package).

fabricio-aguiar:
> Your code looks good. My concern here though is that this code is magical in that Content is automatically matched to a serializer without any input from the plugin writer. It's possible that plugin may have more than one serializer for a Content model depending on different contexts (like in pulp_rpm we have two for Package).

I agree, do you think the solution would be: require plugin writer to specify the serializer on DeclarativeContent?

pulpcore/plugin/stages/content_stages.py
```diff
@@ -1,13 +1,54 @@
...
+                data[f.name] = list(f.value_from_object(content).values_list('pk', flat=True))
+        else:
+            data[f.name] = f.value_from_object(content)
+
+    if issubclass(serializer, SingleArtifactContentSerializer):
+        data["relative_path"] = data.get("relative_path", ".")
+        data["file"] = data.get("file", files.PulpTemporaryUploadedFile("name", "", 3, ""))
+
+    # avoiding deferred_validate on SingleArtifactContentUploadSerializer
+    serializer = serializer(data=data, context={"request": {}})
```
fabricio-aguiar: I'm not happy with these tricks to validating SingleArtifactContentSerializer and SingleArtifactContentUploadSerializer

bmbouter:
I'd like to help with this when I get back from PTO. I think this PR is too fragile to merge.

pulpcore/plugin/stages/content_stages.py
fabricio-aguiar: Alternative:
https://github.com/pulp/pulpcore/compare/master...fabricio-aguiar:5927.2

fabricio-aguiar:
Closing in favor of: https://github.com/pulp/pulpcore/pull/532

PR #516, opened 2020-01-25
Move Pull Request Template to .github directory

dralley:
[noissue]


bmbouter:
@dralley is this where it's supposed to go?

dralley:
@bmbouter It's allowed to go in either place (they're both documented as locations where it will be found) but it's a better location since any future integrations we might add can go in there also.  Also it just makes sense to have it hidden.

PR #518, opened 2020-01-27
[WIP/experiment] Add prometheus metrics support

alikins:
**Note:** This is mostly just an example of how prometheus metrics support could be added to pulpcore (and presumably, all the pulp plugins).

I'm looking for a more general way to set up this configuration, if that is possible.
For example, I'm not sure this is something that could be done with dynaconf 
(it seems like dynaconf's django support doesn't allow changing INSTALLED_APPS or MIDDLEWARE.)

So looking for suggestions on the best way to approach this.

> This enables pulpcore to export pulp metrics to
> be scraped by prometheus.
> 
> It uses the django module 'django_prometheus'
> django app/middleware to do this.
> (https://github.com/korfuri/django-prometheus)
> The result is a set of metrics exposed at /metrics

### DIFF
--- a/pulpcore/app/urls.py
+++ b/pulpcore/app/urls.py
@@ -120,7 +120,7 @@ urlpatterns = [
 ]
 
 with suppress(ImportError):
-    urlpatterns.append(url('', include('django_prometheus.urls')))
+    urlpatterns.append(url("", include("django_prometheus.urls")))
 
 api_info = openapi.Info(
     title="Pulp 3 API",
--- a/pulpcore/app/urls.py
+++ b/pulpcore/app/urls.py
@@ -1,4 +1,5 @@
 """pulp URL Configuration"""
+from contextlib import suppress
 import logging
 
 from django.conf.urls import include, url
@@ -117,6 +118,9 @@ urlpatterns = [
     url(r"^{api_root}orphans/".format(api_root=API_ROOT), OrphansView.as_view()),
     url(r"^auth/", include("rest_framework.urls")),
 ]
+
+with suppress(ImportError):
+    urlpatterns.append(url('', include('django_prometheus.urls')))
 
 api_info = openapi.Info(
     title="Pulp 3 API",
--- a/CHANGES/6462.feature
+++ b/CHANGES/6462.feature
@@ -1,0 +1,9 @@
+Add optional prometheus metrics support
+
+This enables pulpcore to export pulp metrics to
+be scraped by prometheus if the django module 'django_prometheus'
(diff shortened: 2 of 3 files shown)


alikins:
Example content of /metrics https://gist.github.com/alikins/39435297bad1637c4d3bc2edcb96f6b9

Grafana screenshot showing some basic metrics: https://imgur.com/a/xbn7uOK

bmbouter:
dynaconf will let you override the installed apps like any attribute in the settings module. Still though, I prefer the way this PR is positioning it as an optional dependency that "just works" if its installed.

pulpcore/app/urls.py
```diff
@@ -115,6 +115,7 @@ def __repr__(self):
     url(r'^{api_root}status/'.format(api_root=API_ROOT), StatusView.as_view()),
     url(r'^{api_root}orphans/'.format(api_root=API_ROOT), OrphansView.as_view()),
     url(r'^auth/', include('rest_framework.urls')),
+    url('', include('django_prometheus.urls')),
```
cutwater, line 118: Because `django_prometheus` is an optional dependency, urls must be configured conditionally too.

alikins: good catch

alikins:
**Note:** https://github.com/pulp/pulpcore/pull/518/commits/a1ac736b5bcba429236c3a9d588db2eb3001b783 (making django_prometheus imports optional in urls.py) is untested at the moment

bmbouter:
I filed the installer change here: https://pulp.plan.io/issues/6808

RCMariko:
And is there an issue already written up for this support? Good to see this moving forward.

alikins:
@bmbouter I can up the examples from https://github.com/pulp/pulpcore/pull/518#issuecomment-578821518 and add some docs into docs rst somewhere.

alikins:
@bmbouter One thing I was wondering about is if it would be useful to add a settings.py variable to explicitly disable prometheus even if the required django-prometheus modules are importable.

alikins:
@bmbouter 

> dynaconf will let you override the installed apps like any attribute in the settings module. Still though, I prefer the way this PR is positioning it as an optional dependency that "just works" if its installed.

That was my first approach, but didn't have much luck getting that to work. Could be revisited.

daviddavis:
We're closing this until the requested changes can be made. Please reopen.

ralphbean:
Chiming in here just to say that a user may want to see prometheus metrics around tasks - for example, their duration, total count, and error counts - in addition to the httpd request metrics already implemented here.

PR #520, opened 2020-01-27
Adds ALLOWED_IMPORT_PATHS for file:// urls

bmbouter:
Remote.url now validate to only allow file:// paths that are a subpath
of a configured ALLOWED_IMPORT_PATHS setting. This setting comes with
docs. URLs starting with other protocol handlers, e.g. http:// or
https:// are unaffected.

https://pulp.plan.io/issues/5974
closes #5974


bmbouter:
@gmbnomis What do you think of this PR?

Note also I wanted to have the validation occur at sync time, but what I realized is that sync code is fully in the control of the plugin so every plugin would need to add it. Is preventing any file:// from saving enough? What do you recommend?

docs/settings.rst
```diff
@@ -199,6 +199,19 @@ REMOTE_USER_ENVIRON_NAME
    Defaults to ``'REMOTE_USER'``.
 
 
+.. _allowed-import-paths:
+
+ALLOWED_IMPORT_PATHS
+^^^^^^^^^^^^^^^^^^^^
+
+   One or more real filesystem paths that Remote's with filesystem paths can import from. For
```
daviddavis: nitpick -- s/Remote's/Remotes/

daviddavis:
I feel like we should probably call out in the docs that ALLOWED_IMPORT_PATH is only checked at Remote creation/update and not before sync. Users might mistakenly think they can edit it on the fly.

docs/settings.rst
bmbouter: fixed in next push, ty.

bmbouter:
I added a note to the docs about it only be checked at creation or update time.

bmbouter:
Actually I realized we could put an additional check in the FileDownloader here and then I could remove this note and make it better. Wdyt?

pulpcore/download/file.py
```diff
@@ -27,7 +27,13 @@ def __init__(self, url, **kwargs):
             url (str): The url to the file. This is expected to begin with `file://`
             kwargs (dict): This accepts the parameters of
                 :class:`~pulpcore.plugin.download.BaseDownloader`.
+
+        Raises:
+            ValidationError: When the url starts with `file://`, but is not a subfolder of a path in
+                the ALLOWED_IMPORT_PATH setting.
         """
+        from pulpcore.app.serializers import RemoteSerializer
```
bmbouter: This is not great, but I tried to work around the circuilar import several ways and it was still there.

daviddavis: maybe leave a comment?

bmbouter: Anytime I see these I think circular import. I will if you want tho just lmk.

PR #532, opened 2020-02-11
Validate Content data in the Stages API

fabricio-aguiar:
https://pulp.plan.io/issues/5927
closes #5927

### DIFF
--- a/CHANGES/plugin_api/5927.feature
+++ b/CHANGES/plugin_api/5927.feature
@@ -1,0 +1,1 @@
+Validate Content data in the Stages API
--- a/pulpcore/app/apps.py
+++ b/pulpcore/app/apps.py
@@ -81,6 +81,7 @@ class PulpPluginAppConfig(apps.AppConfig):
         from pulpcore.app.serializers import ModelSerializer
 
         self.named_serializers = {}
+        self.model_serializers = {}
         if module_has_submodule(self.module, SERIALIZERS_MODULE_NAME):
             # import the serializers module and track any discovered serializers
             serializers_module_name = '{name}.{module}'.format(
@@ -93,6 +94,9 @@ class PulpPluginAppConfig(apps.AppConfig):
                     # gets registered in the named_serializers registry.
                     if obj is not ModelSerializer and issubclass(obj, ModelSerializer):
                         self.named_serializers[objname] = obj
+                        model = getattr(obj.Meta, "model", None)
+                        if model:
+                            self.model_serializers[obj.Meta.model] = obj
                 except TypeError:
                     # obj isn't a class, issubclass exploded but obj can be safely filtered out
                     continue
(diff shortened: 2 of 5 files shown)


pulpcore/plugin/stages/models.py
```diff
@@ -109,17 +111,19 @@ class DeclarativeContent:
 
     __slots__ = (
         'content',
+        'serializer',
```
bmbouter: Conceptually, each content type pairs with one serializer, so from an object oriented perspective what about defining this relationship on the model? I need to look into the loader code to see what is there already also.

daviddavis: I was thinking we could have situations where we have multiple serializers for a Content model. To give an example, for rpm packages that are uploaded we don't validate name on upload (it's fetched from the artifact) but on sync, we might want to validate it exists. That's what led me to believe we should tell the stages api which serializer to use.

bmbouter: I can see the thinking. I imagined the uploading of an artifact that isn't actually a content unit yet. For example in pulp_ansible the tarball is just a tarball until it's imported so the Collection model has one way to represent an already saved one. Uploads wouldn't use that one.

Part of my motivation to change this is that using this code means setting it item-by-item which when dealing with mixed types in sync code is more complicated. Having it declared on the object is easier which I think is helpful to the integrator.

daviddavis: > I can see the thinking. I imagined the uploading of an artifact that isn't actually a content unit yet. For example in pulp_ansible the tarball is just a tarball until it's imported so the Collection model has one way to represent an already saved one. Uploads wouldn't use that one.

This kind of makes sense to me in the one-shot upload case. However, you can also directly create Content in the API. This seems like it should use a serializer tied to the Content model. As a plugin writer, I think I'd be confused about which serializer I'm supposed to set on the model if I need to have different ones for uploading/directly creating content in the API vs sync.

> Part of my motivation to change this is that using this code means setting it item-by-item which when dealing with mixed types in sync code is more complicated. Having it declared on the object is easier which I think is helpful to the integrator.

I can see this motivation and agree we should try to avoid having to set serializers on an item by item basis. Perhaps for now we can go with defining `serializer` on models but in the future  allow plugin writers to optionally set a mapping with contexts:

```python
class FileContent(Content):
    serializer = FileContentSerializer

class TestContent(Content):
    serializer = {'upload': TestContentUploadSerializer, 'sync': 'TestContentSerializer'}
```

bmbouter: Defining the serializer on the model for now makes sense and adding more later I think does too. Especially since we are favoring DRF validation with serializers over Django's own model validation, I think we need plugin writers to express how something should validate the model.

fabricio-aguiar:
Closing as the story will be rewritten: https://www.redhat.com/archives/pulp-dev/2020-April/msg00075.html

PR #536, opened 2020-02-11
RelatedField return empty strings instead of None

fabricio-aguiar:
### Reference: 
https://pulp.plan.io/issues/6069#note-7

### DIFF
--- a/pulpcore/app/serializers/base.py
+++ b/pulpcore/app/serializers/base.py
@@ -4,6 +4,7 @@ from urllib.parse import urljoin
 from django.core.validators import URLValidator
 from drf_queryfields.mixins import QueryFieldsMixin
 from rest_framework import serializers
+from rest_framework.fields import get_attribute
 from rest_framework_nested.relations import (
     NestedHyperlinkedIdentityField,
     NestedHyperlinkedRelatedField,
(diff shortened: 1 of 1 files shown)


pulpcore/app/serializers/base.py
```diff
@@ -156,9 +157,23 @@ def _view_name(self, obj):
...
+        if isinstance(obj, str):
+            return obj
         view_name = self._view_name(obj)
         return super().get_url(obj, view_name, request, *args, **kwargs)
 
+    def get_attribute(self, instance):
+        # Return empty string instead of None
+        try:
+            attr = get_attribute(instance, self.source_attrs)
+        except Exception:
```
dkliban: What kind of exception is possible here?

fabricio-aguiar: BuiltinSignatureError, KeyError, AttributeError:
https://github.com/encode/django-rest-framework/blob/master/rest_framework/fields.py#L449-L488

fabricio-aguiar:
This PR is a PoC, I don't know if the best approach is changing `_DetailFieldMixin`, maybe a better option would be creating a specific field for it like `StringDetailRelatedField`. I just wanted to see how bindings would behave if instead of getting None it received an empty string.

PR #543, opened 2020-02-17
Considering base version when removing duplicates

fabricio-aguiar:
https://pulp.plan.io/issues/5964
closes #5964

### DIFF
--- a/pulpcore/plugin/repo_version_utils.py
+++ b/pulpcore/plugin/repo_version_utils.py
@@ -31,8 +31,8 @@ def remove_duplicates(repository_version):
         repository_version: The :class:`~pulpcore.plugin.models.RepositoryVersion` to be checked
             and possibly modified.
     """
-    added_content = repository_version.added()
-    existing_content = repository_version._content_old()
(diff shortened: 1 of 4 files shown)


pulpcore/plugin/repo_version_utils.py
```diff
@@ -31,7 +31,7 @@ def remove_duplicates(repository_version):
         repository_version: The :class:`~pulpcore.plugin.models.RepositoryVersion` to be checked
             and possibly modified.
     """
-    added_content = repository_version.added()
+    added_content = repository_version.added(base_version=repository_version.base_version)
     existing_content = repository_version._content_old()
```
gmbnomis: I think `existing_content` has to be adapted as well. IMHO, we should get rid of `_content_old()` altogether now that we have `previous()`.

And, in light of these subtle problems, I suggest to add a test (this could be similar to the reproducer in the issue).

fabricio-aguiar: Agreed, I will start to work on tests for it

PR #555, opened 2020-02-19
Repository.latest_version() considering deleted versions

fabricio-aguiar:
https://pulp.plan.io/issues/6147
closes #6147


pulpcore/app/models/repository.py
```diff
@@ -740,6 +740,9 @@ def delete(self, **kwargs):
                 # and delete the version
                 repo_relations.filter(version_added=self).delete()
                 repo_relations.filter(version_removed=self).update(version_removed=None)
+                if self.number == self.repository.last_version:
+                    self.repository.last_version = self.number - 1
+                    self.repository.save()
```
dralley: I haven't tested this out, but could you check this scenario?  What happens if you create 3 new repository versions, so you have:

0, 1, 2, 3
repository.last_version = 3

If you delete 3 and then 2 you would probably get the correct result:

0, 1,
repository.last_version = 1

But if you delete 2 and then 3, then I think with this algorithm you would get:

0, 1
repository.last_version = 2

I could be wrong, it's just an edge case we should check :)

fabricio-aguiar: I will write a test for it

dralley: Probably just 2 versions is enough to test this, version 1 isn't really doing much as-is.

dralley: I guess this also brings a question to mind -- do we want repository version numbers to be re-usable?  Because if we set ```number``` as ```last_version + 1``` and we reset ```last_version``` appropriately, then that means that it will be possible to overwrite the value of ```${REPO_HREF}/versions/2/``` with a new version 2 that contains different content.  It's not monotonically increasing.

dralley: Solution from IRC:

* Rename ```Repository.last_version``` to ```Repository.next_version``` and do a data migration to add +1 to its value
* When creating a new repository version, use the value of ```Repository.next_version``` and if it is successfully created increment this value.
* To determine the latest repository version, query the repository versions with a descending sort on the version ```number```, and pick the ```.first()``` one.

dkliban: This sounds right to me. Does this  change the REST API ?

daviddavis: @dkliban it shouldn't. We don't expose that field.

dralley: It shouldn't, no.  ```last_version``` isn't serialized and ```next_version``` doesn't need to be either.

dkliban: Great! thanks for confirming that.

pulpcore/app/migrations/0022_rename_last_version.py
```diff
@@ -0,0 +1,29 @@
...
+from django.db import migrations, transaction
+
+
+def plus_one(apps, schema_editor):
+    # re: https://pulp.plan.io/issues/6147
+    with transaction.atomic():
+        Repository = apps.get_model('core', 'Repository')
+        if 'last_version' in dir(Repository):
+            for repository in Repository.objects.all().only('last_version'):
+                repository.last_version = int(repository.last_version) + 1
```
dralley: I think you could try something like this, it might be more efficient: https://stackoverflow.com/questions/12661253/how-to-bulk-update-with-django

The main benefit though is that I think ```.save()``` would change the ```last_updated``` timestamp and while I don't think that's necessarily a problem we should avoid it if we can help it.

pulpcore/app/models/repository.py
```diff
@@ -93,10 +92,8 @@ def new_version(self, base_version=None):
         with transaction.atomic():
             version = RepositoryVersion(
                 repository=self,
-                number=int(self.last_version) + 1,
```
dralley: I love that this logic was made simpler!

pulpcore/app/models/repository.py
```diff
@@ -748,7 +760,7 @@ def delete(self, **kwargs):
                 RepositoryContent.objects.filter(version_removed=self) \
                     .update(version_removed=None)
                 CreatedResource.objects.filter(object_id=self.pk).delete()
-                self.repository.last_version = self.number - 1
+                self.repository.next_version = self.number - 1
```
dralley: I think this line should be deleted entirely.  We should only increment ```repository.next_version``` once the repository version has been successfully created and finalized, so we shouldn't ever need to decrement it.  If the repository version wasn't completed then it shouldn't have been incremented so we shouldn't need to decrement it here.

pulpcore/app/models/repository.py
```diff
@@ -719,6 +716,21 @@ def _squash(self, repo_relations, next_version):
...
+
+        Args:
+            args (list): list of positional arguments for Model.save()
+            kwargs (dict): dictionary of keyword arguments to pass to Model.save()
+        """
+        with transaction.atomic():
+            adding = self._state.adding
+            super().save(*args, **kwargs)
+            if adding:
+                self.repository.next_version = self.number + 1
```
dralley: I think this actually needs to happen here: https://github.com/pulp/pulpcore/pull/555/files#diff-ae59196c41f139ec2b7e3ae3aa581a45R832

pulpcore/app/migrations/0022_rename_last_version.py
```diff
@@ -0,0 +1,28 @@
...
+
+from django.db import migrations, transaction
+from django.db.models import F
+
+
+def plus_one(apps, schema_editor):
+    # re: https://pulp.plan.io/issues/6147
+    with transaction.atomic():
+        Repository = apps.get_model('core', 'Repository')
+        if 'last_version' in dir(Repository):
```
dralley: I'm not totally sure I understand what the conditional does, do some Repository objects not have this field?

fabricio-aguiar: it is for not breaking if for any reason someone re-run the migration, because after migration `last_version` will no longer exist

dralley: I don't think it's possible to re-run a migration unless you go out of your way to do so, the most recent migration is stored somewhere and it picks up from there afterwards

bmbouter: It's not possible to rerun a migration because we don't provide downgrade migrations. Is this part of the logic still needed?

fabricio-aguiar: I just added it in case someone would rerun, since it is not possible to rerun, it is not needed

CHANGES/6147.removal
```diff
@@ -0,0 +1 @@
+Replacing Repository.last_version with Repository.next_version
```
bmbouter: Can we rephrase this some? Can it be `Renaming Repository.last_version to Repository.next_version`

PR #558, opened 2020-02-21
Allow to filter SigningService by name

mdellweg:


pulpcore/app/viewsets/content.py
```diff
@@ -138,3 +149,4 @@ class SigningServiceViewSet(NamedModelViewSet,
     endpoint_name = 'signing-services'
     queryset = SigningService.objects.all()
     serializer_class = SigningServiceSerializer
+    filterset_class = SigningServiceFilter
```
lubosmj: How about using `filterset_fields` (https://django-filter.readthedocs.io/en/master/guide/rest_framework.html#using-the-filterset-fields-shortcut)? Then you do not need to create a separate class.
```suggestion
    filterset_fields = ['name']
```

pulpcore/app/models/content.py
```diff
@@ -459,7 +459,10 @@ def sign(self, filename):
 
         """
         completed_process = subprocess.run(
-            [self.script, filename], stdout=subprocess.PIPE, stderr=subprocess.PIPE
+            [self.script, filename],
+            env={},
```
mdellweg: I added this, because the script inherited the complete environment of Django.

bmbouter:
@mdellweg Can we add a changelog entry for this? Since it's a new feature I try to have an entry for any non-trivial change. To do that file a story (even a short one is fine) and then use that number for your commit and CHANGES entry please.

pulpcore/plugin/models/__init__.py
```diff
@@ -24,5 +24,6 @@
     RepositoryContent,
     RepositoryVersion,
     RepositoryVersionDistribution,
+    SigningService,
```
bmbouter: This probably needs it's own changelog entry also please.

mdellweg:
I will add tickets and changelogs next week, when I'm back in the office.
If it turns out to be urgent, please take any action needed.

bmbouter:
@mdellweg next week sounds fine. My interest was more to make it clear to anyone coming across the PR that it's status is reviewed and waiting for changelog changes. Thank you!

mdellweg:
ping @bmbouter

PR #560, opened 2020-02-25
Stop sending basic auth credentials to redirect location

daviddavis:
Pulp is getting redirected and sending the Remote's credentials to the
redirected location. Looks like if we set the credentials on the request
instead of the session, aiohttp will only send the credentials to
redirected locations when the domains match which is the behavior we
want.

### DIFF
--- a/CHANGES/6227.bugfix
+++ b/CHANGES/6227.bugfix
@@ -1,0 +1,1 @@
+Stopped HttpDownloader sending basic auth credentials to redirect location if domains don't match.
--- a/pulpcore/download/factory.py
+++ b/pulpcore/download/factory.py
@@ -106,15 +106,8 @@ class DownloaderFactory:
 
         conn = aiohttp.TCPConnector(**tcp_conn_opts)
 
-        auth_options = {}
-        if self._remote.username and self._remote.password:
-            auth_options['auth'] = aiohttp.BasicAuth(
-                login=self._remote.username,
-                password=self._remote.password
-            )
-
         timeout = aiohttp.ClientTimeout(total=None, sock_connect=600, sock_read=600)
-        return aiohttp.ClientSession(connector=conn, timeout=timeout, **auth_options)
+        return aiohttp.ClientSession(connector=conn, timeout=timeout)
 
     def build(self, url, **kwargs):
         """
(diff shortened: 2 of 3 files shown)


alikins:
Looks okay to me at first glance. 

Is basic-auth the only auth mechanism supported when syncing with automation-hub?

bmbouter:
> Looks okay to me at first glance.
> 
> Is basic-auth the only auth mechanism supported when syncing with automation-hub?

Basic and certificate configuration via `client_cert` and `client_key` https://pulp-ansible.readthedocs.io/en/latest/restapi.html#operation/remotes_ansible_collection_create

The server side also supports both types as well.

It depends what happens with @cutwater and the adjustment of how this content is served to users.

bmbouter:
@cutwater having pulp_ansible perform validation directly is important because non automation hub users should still be able to have their CLI's auth to pulp_ansible also.

alikins:
@bmbouter I had forgotten that c.r.c still supports client cert auth as well.

Mostly curious if pulp_ansible supports (or if it makes sense to support) the keycloak based "long term" auth tokens that ansible-galaxy uses ('auth_url' and 'token' described at https://docs.ansible.com/ansible/devel/galaxy/user_guide.html#configuring-the-ansible-galaxy-client)

The current CollectionRemote doesn't seem to have fields for it (https://pulp-ansible.readthedocs.io/en/latest/restapi.html#operation/remotes_ansible_ansible_create).
Probably not a requirement at the moment, but may become one in the future.

Either way, regardless of the original auth mechanism, when download request redirect to s3, the auth headers need to be dropped when making the requests to s3.
And this PR seems to be doing that correctly :+1:

bmbouter:
@alikins Pulp content app checks user authorization to that content using a pluggable mechanism we call a certguard, not a great name because it suggests it's only for certificates. It's [here](https://pulp-certguard.readthedocs.io/en/latest/) We have an X.509 certguard for example Webservers also can provide auth, but doing it in Pulp allows us to extend the RBAC authorization in a way the webserver wouldn't be as easy to do.

I would like to see pulp_ansible be able to use a certguard to authorize the client's request. We would want to reuse that machinery in the galaxy API also to authorize them in that service also. Certguard can have it's checks called from there as well.

Would you be willing to file a story linking to the docs about how that auth works on the client and server?

PR #562, opened 2020-02-26
Update pip inside of the container

dralley:
A new version of pip is needed to use Manylinux2014 pre-built wheels.
Pre-built wheels allows us to avoid installing some of the RPMs.

[noissue]

### DIFF
--- a/containers/images/pulp/Dockerfile.j2
+++ b/containers/images/pulp/Dockerfile.j2
@@ -32,8 +32,6 @@ RUN		dnf -y update && \
 		dnf -y install glibc-langpack-en && \
 		dnf -y install python3-createrepo_c && \
 		dnf -y install python3-libmodulemd && \
-		dnf -y install python3-libcomps && \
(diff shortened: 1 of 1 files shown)


containers/images/pulp/Dockerfile.j2
```diff
@@ -43,6 +41,10 @@ RUN ln -s /usr/bin/python3 /usr/bin/python
 # /usr/local/bin will be before it in the PATH .
 RUN ln -s /usr/bin/pip3 /usr/local/bin/pip
 
+# Need to update pip inside of the container so that manylinux2014
+# packages are supported
+RUN pip install --user --upgrade pip
```
mikedep333: We install everything system-wide in the container, so drop "--user". (These are single-process containers.)

And wouldn't it be better to just upgrade from Fedora 30 to Fedora 31 instead? We would go from pip 19.03 to 19.1.1 that way.

PR #571, opened 2020-03-05
Added MANIFEST.in part to web server snippets docs

### LINKED ISSUE #6249: Report correlation ID in Django API responses
As a user who wants is troubleshooting a failure, at times on some installations, it would be useful to have the correlation ID used for a given Django API call. Returning this data as a response header makes sense to me.

bmbouter:
The MANIFEST.in part of the webserver was missing from the snippets
guide for plugin writers.

https://pulp.plan.io/issues/6249
closes #6249


bmbouter:
@gmbnomis and @daviddavis FYI I tried to improve these docs with this rewrite.

PR #573, opened 2020-03-08
Handle no-previous-version case in remove_duplicates util

goosemania:
[noissue]


goosemania:
@daviddavis , should I mark it to cherry-pick? I think it's a bugfix.

daviddavis:
:+1:

PR #584, opened 2020-03-16
Refactor the existing SigningService classes

quba42:
fixes #6342
https://pulp.plan.io/issues/6342

This is based on my experience of implementing the `AptReleaseSigingService` class here: https://github.com/pulp/pulp_deb/pull/160

I found that I replicated some of the exact same functions as for the `AsciiArmoredDetachedSigningService` since the only thing that is really different for different signing services is the validation. I therefore moved the reusable functions into the parent class, and split out a dedicated `validate` function from the existing `save` function. The Idea is that this is the one and only function that needs to be provided by each subclass of `SigningService`.

### DIFF
--- a/docs/plugins/reference/metadata-signing.rst
+++ b/docs/plugins/reference/metadata-signing.rst
@@ -9,14 +9,43 @@ Metadata Signing
     releases.
 
 Plugin writers wishing to enable users to sign metadata need to add a new field ``metadata_signing_service``
-to their implementation of a repository. This field should be exposed to users who consume content
-via REST API. The users may afterwards specify which kind of a signing service will be used to sign the
+to their implementation of a repository and/or publication. This field should be exposed to users who consume
+content via REST API. The users may afterwards specify which signing service will be used to sign the
 metadata when creating a publication.
 
-In order to sign metadata, plugin writers are required to call the method ``sign()`` of the subclasses
-inheriting from the model ``SigningService``. This method invokes a signing script which is provided by
-an administrator and creates a detached ascii-armored signature if called via the class
-``AsciiArmoredDetachedSigningService``.
+Every signing service will always be an instance of a subclass of the ``SigningService`` model. Plugin
+writers may either use the existing ``AsciiArmoredDetachedSigningService``, or use that as a reference for
+writing their own signing service model.
+
+The ``SigningService`` base class already provides the fully implemented ``sign()`` method, the signature of
+the ``validate()`` method (which must be implemented by each subclass), and the ``save()`` method (which
+calls the ``validate()`` method, but is otherwise fully implemented).
+
+In order to sign metadata, plugin writers are required to call the ``sign()`` method of the signing service
+being used. This method invokes the signing script (or other executable) which is provided by the
+administrator who instantiates a concrete signing service. Instantiating/creating a concrete signing service
+will ultimately call the ``save()`` method, which will in turn call ``validate()``. As a result, it is up to
+the ``validate()`` method to ensure the signing service script provided by the administrator actually provides
+any signatures, signature files, and return values, as required by the individual ``SigningService`` subclass.
+
+This is why implementing a signing service model other than ``AsciiArmoredDetachedSigningService`` simply
+requires inheriting from ``SiginingService`` and then implementing ``validate()``.
+
+.. note::
+    The existing ``AsciiArmoredDetachedSigningService`` requires a signing script that creates a detached
+    ascii-armored signature file, and prints valid JSON in the following format to stdout:
+
+        {"file": "filename", "signature": "filename.asc", "key": "public.key"}
+
+    Here "filename" is a path to the original file that was signed (passed to the signing script by the
+    ``sign()`` method), "filename.asc" is a path to the signature file created by the script, and "public.key"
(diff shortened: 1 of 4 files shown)


quba42:
I was just hoping to get some feedback for now, to see if people find this desirable.

pulpcore/app/models/content.py
```diff
@@ -379,51 +379,98 @@ class SigningService(BaseModel):
...
+            [self.script, filename],
+            env={},
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+
+        if completed_process.returncode != 0:
+            raise RuntimeError(str(completed_process.stderr))
+
+        return json.loads(completed_process.stdout)
```
mdellweg: This could throw `JSONDecodeError`. Should we catch that (and translate it into a PulpError)?

quba42: This would have the advantage of taking the onus of meaningfully reporting that error away from the validate function in the child classes, so I am for it.

quba42: For now, I have moved the exception handling from the verify method of the child class to the sign method in the parent class.
This could be changed to some dedicated error type at a later time as a separate issue.

pulpcore/app/models/content.py
```diff
@@ -379,54 +379,99 @@ class SigningService(BaseModel):
...
         gpg = gnupg.GPG()
         with tempfile.TemporaryDirectory() as temp_directory_name:
             with tempfile.NamedTemporaryFile(dir=temp_directory_name) as temp_file:
-                temp_file.write(b"random-data")
-                try:
-                    data = self.sign(temp_file.name)
-                except RuntimeError:
-                    raise
+                temp_file.write(b"arbitrary data")
+                temp_file.flush()
```
mdellweg: :+1:

dkliban:
@quba42 Is this ready for a review?

quba42:
Yes.

pulpcore/app/models/content.py
```diff
@@ -379,54 +379,99 @@ class SigningService(BaseModel):
...
+            An absolute path to the external signing script (or executable).
 
     """
     name = models.TextField(db_index=True, unique=True)
     script = models.TextField()
 
     def sign(self, filename):
         """
-        Sign a passed file.
+        Pass the filename parameter to an external script (or executable).
```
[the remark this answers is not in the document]
quba42: I have amended the first line of the docstring to try and accommodate the feedback, thanks!

quba42:
> These changes look good to me. However, I believe these changes should be also properly documented (https://docs.pulpproject.org/plugins/reference/metadata-signing.html, see the note at the bottom). Currently, we say that only the subclass `AsciiArmoredDetachedSigningService` returns a dictionary object. After merging this pull request, it will be the class `SigningService` by default.

My initial quick read of the docs suggested It does not need much change. However having looked at it again, I have tried to amend the docs in a way that is hopefully helpful.

docs/plugins/reference/metadata-signing.rst
```diff
@@ -9,14 +9,41 @@ Metadata Signing
...
+The ``SigningService`` base class already provides the fully implemented ``sign()`` method, the signature of
+the ``validate()`` method (which must be implemented by each subclass), and the ``save()`` method (which
+calls the ``validate()`` method, but is otherwise fully implemented).
+
+In order to sign metadata, plugin writers are required to call the ``sign()`` method of the signing service
+being used. This method invokes the signing script (or other executable) which is provided by the
+administrator who instantiates a concrete signing service. Instantiating/creating a concrete signing service
+will ultimately call the ``save()`` method, which will in turn call ``validate()``. As a result, it is up to
+the ``validate()`` method to ensure the signing service script provided by the administrator actually provides
+any signatures, signature files, and return values, as required by the individual signing service.
```
quba42: This should end with `by the individual "SigningService" subclass.`. Richt now it sounds like "as required by the individual instantiated signing service object.

docs/plugins/reference/metadata-signing.rst
```diff
@@ -9,14 +9,41 @@ Metadata Signing
...
+
+        {"file": "filename", "signature": "filename.asc", "key": "public.key"}
+
+    Here "filename" is a path to the original file that was signed (passed to the signing script by the
+    ``sign()`` method), "filename.asc" is a path to the signature file created by the script, and "public.key"
+    is a path to the signature file containing the public key used by the script.
+
+    This json is converted to a python dict and returned by the ``sign()`` method. If an error occurs, a
+    runtime error is raised instead. All of this is enforced by the ``validate()`` method at the time of
+    instantiation.
```
quba42: This note box could also include a cross-reference to the existing "Metadata Signing" workflow documentation (https://docs.pulpproject.org/en/master/nightly/workflows/signed-metadata.html?highlight=signing%20service)

docs/plugins/reference/metadata-signing.rst
quba42: Done.

docs/plugins/reference/metadata-signing.rst
quba42: Done.

PR #585, opened 2020-03-16
Adding PulpExporter model and endpoints

daviddavis:
Also adding a new Export model and refactoring existing Exporters.

fixes #6328
https://pulp.plan.io/issues/6328


pulpcore/app/openapigenerator.py
```diff
@@ -118,7 +118,11 @@ def get_paths(self, endpoints, components, request, public):
                             resources[resource_path] = param_name
                             resource_example[resource_path] = self.get_example_uri(path)
                             path = path.replace(resource_path, '{%s}' % resources[resource_path])
-                        example = resource_example[resource_other_path]
+                        try:
+                            example = resource_example[resource_other_path]
+                        except KeyError:
+                            # also try resouce_path
+                            example = resource_example[resource_path]
```
daviddavis: @dkliban I'm getting `KeyError: '/pulp/api/v3/exporters/file/filesystem/{pulp_id}/'` here without my change. It looks like resource_path is `/pulp/api/v3/exporters/file/filesystem/{exporter_pk}/` which is in resource_example but resource_other_path (`/pulp/api/v3/exporters/file/filesystem/{pulp_id}/`) is not. 

This seems to be due to the fact that I have a nested master-detail viewset at `/pulp/api/v3/exporters/file/filesystem/{exporter_pk}/exports/`. Not sure if this is the right fix. Any thoughts?

CHANGES/6328.feature
```diff
@@ -0,0 +1 @@
+Added a new APIs for PulpExorpters and Exports.
```
bmbouter: spelling PulpExorpters

bmbouter: it would also be cool if the API url was written here so urls could see it easily.

pulpcore/app/models/exporter.py
```diff
@@ -2,23 +2,99 @@
...
     """
-    A base class that provides logic to export a set of content. This set of content can be a
-    publication, repo version, etc.
+    A class that represents an Export.
+
+    Fields:
+
+        params (models.JSONField): A set of parameters used to create the export
+
+    Relations:
```
bmbouter: two `Relations` sections?

pulpcore/app/models/exporter.py
```diff
@@ -2,23 +2,99 @@
...
+        exporter (models.ForeignKey): The Exporter that created this Export.
+    """
+    params = JSONField(null=True)
+    task = models.ForeignKey("Task", on_delete=models.PROTECT)
+    exporter = models.ForeignKey("Exporter", on_delete=models.CASCADE)
+
+
+class ExportedResource(GenericRelationModel):
+    """
+    A class to represent anything that was exported in an Export. This includes repo versions,
```
bmbouter: Can the first line be a single line only for pep 257 compliance?

pulpcore/app/serializers/fields.py
```diff
@@ -278,6 +278,48 @@ def to_representation(self, value):
             ))
 
 
+class ExportsIdentityFromExporterField(DetailIdentityField):
+    view_name = 'exporters-detail'
+
+    def __init__(self, view_name=None, **kwargs):
+        assert view_name is None, 'The `view_name` must not be set.'
```
bmbouter: I learned recently that assert probably shouldn't be relied on for production code because many python interpreters used in production are compiled w/ flags that specifically ignore `assert`. I used to do this pattern a lot, but now I don't. wdyt about this?

pulpcore/app/serializers/fields.py
```diff
@@ -278,6 +278,48 @@ def to_representation(self, value):
...
+        super().__init__(view_name=self.view_name, **kwargs)
+
+    def get_url(self, obj, view_name, request, *args, **kwargs):
+        return super().get_url(obj, self.view_name, request, *args, **kwargs) + "exports/"
+
+
+class ExportFieldGetURLMixin:
+    view_name = 'exports-detail'
+
+    def __init__(self, view_name=None, **kwargs):
```
bmbouter: Because of the diamond inheritance problem I believe mixins are discouraged from providing methods are that are guaranteed to already exist on the objects they are mixing into. In other words I don't think a mixin should provide `__init__`. wdyt?

pulpcore/app/tasks/export.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def fs_publication_export(exporter_pk, publication_pk):
+    """
+    Export a publication to the file system.
+
+    Args:
+        exporter_pk (str): FileSystemExporter pk
+        publication_pk (str): Publication pk
+        publication_pk (str): RepositoryVersion pk
```
bmbouter: duplicate line here

pulpcore/app/serializers/fields.py
daviddavis: That's also my understanding and so I think this assertion makes sense since this assertion is geared towards developers/plugin writers and not users.

pulpcore/app/serializers/fields.py
daviddavis: I copied this from [RepositoryVersionFieldGetURLMixin](https://github.com/pulp/pulpcore/blob/master/pulpcore/app/serializers/fields.py#L190-L192). I don't have a strong opinion though.

daviddavis: Hey so it seems what when I try to remove this method I get an error from DRF:

```
  File "/home/vagrant/devel/pulpcore/pulpcore/app/serializers/__init__.py", line 36, in <module>
    from .exporter import (  # noqa
  File "/home/vagrant/devel/pulpcore/pulpcore/app/serializers/exporter.py", line 44, in <module>
    class ExportSerializer(ModelSerializer):
  File "/home/vagrant/devel/pulpcore/pulpcore/app/serializers/exporter.py", line 48, in ExportSerializer
    pulp_href = ExportIdentityField()
  File "/usr/local/lib/pulp/lib64/python3.7/site-packages/rest_framework/relations.py", line 425, in __init__
    assert view_name is not None, 'The `view_name` argument is required.'
AssertionError: The `view_name` argument is required.
```

pulpcore/app/models/exporter.py
daviddavis: Fixed.

pulpcore/app/serializers/fields.py
daviddavis: Filed an issue to refactor and fix this pattern: https://pulp.plan.io/issues/6389. Feel free to add onto it.

PR #597, opened 2020-03-19
Dramatically increase re-sync performance

dralley:
Increase the performance of the QueryExisting* stages. pulp_created=None
was originally picked because it wouldn't match with anything, and an empty
Q() object matches with everything. But it's a lookup on an unindexed field
that requires looking at an entirely different table, so it ends up
being incredibly inefficient. It turns out that there is a way to
accomplish the same thing that gets compiled out of the query entirely.

[noissue]

### DIFF
--- a/pulpcore/plugin/stages/api.py
+++ b/pulpcore/plugin/stages/api.py
@@ -78,7 +78,7 @@ class Stage:
             content = await self._in_q.get()
             if content is None:
                 break
-            log.debug(_('%(name)s - next: %(content)s.'), {'name': self, 'content': content})
+            log.debug('%(name)s - next: %(content)s.', {'name': self, 'content': content})
             yield content
 
     async def batches(self, minsize=500):
--- a/pulpcore/plugin/stages/content_stages.py
+++ b/pulpcore/plugin/stages/content_stages.py
@@ -37,7 +37,7 @@ class QueryExistingContents(Stage):
             The coroutine for this stage.
         """
         async for batch in self.batches():
-            content_q_by_type = defaultdict(lambda: Q(pulp_created=None))
+            content_q_by_type = defaultdict(lambda: Q(pk__in=[]))
             for d_content in batch:
                 if d_content.content._state.adding:
                     model_type = type(d_content.content)
--- a/CHANGES/6373.bugfix
+++ b/CHANGES/6373.bugfix
@@ -1,0 +1,1 @@
+Improved resync performance by up to 2x with a change to the content stages.
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -42,7 +42,7 @@ class QueryExistingArtifacts(Stage):
             The coroutine for this stage.
         """
         async for batch in self.batches():
-            all_artifacts_q = Q(pulp_created=None)
+            all_artifacts_q = Q(pk__in=[])
             for d_content in batch:
                 for d_artifact in d_content.d_artifacts:
                     one_artifact_q = d_artifact.artifact.q()


dralley:
## Performance tests

TL;DR improves the speed of file plugin re-sync nearly double, RPM plugin re-sync and re-migration by around 33%

### File Plugin, 30k units, on_demand

Before:

- Initial sync: 72 seconds
- Re-sync: 63 seconds

After:

- Initial sync: 72 seconds
- Re-sync: 33 seconds

### Rpm Plugin, ~10k units (centos), on_demand

Before:

- Initial sync: 61 seconds
- Re-sync: 48 seconds

After:

- Initial sync: 53 seconds
- Re-sync: 33 seconds

### Migration plugin, migration of 1 repository, ~10k units (centos) from Pulp 2

Before:

- Initial migration: 88 seconds
- Re-migration: 77 seconds

After:

- Initial migration: 88 seconds
- Re-migration: 54 seconds

dralley:
This probably deserves a changelog entry but there's no issue filed.  Should I file an issue? Is there a way to add an entry without doing so?

dralley:
Even with this improvement, the QueryExistingContents() stage is really expensive.  It's 50% (*exactly* 50%) of the time spent in a resync (of the file plugin), and 33% of the time used in a re-migration (of an RPM repository). @bmbouter @fabricio-aguiar do you have any ideas on how to speed it up further?

Resync:
file_resync

goosemania:
@dralley, thanks for all the investigation. Even though it's a problem, It's great to see that we'll definitely save a lot by tracking remotes and not going through all the content and thus not using QueryExistingContent Stage that much during re-migration.

pulpcore/plugin/stages/content_stages.py
```diff
@@ -37,7 +37,7 @@ class QueryExistingContents(Stage):
             The coroutine for this stage.
         """
         async for batch in self.batches():
-            content_q_by_type = defaultdict(lambda: Q(pulp_created=None))
+            content_q_by_type = defaultdict(lambda: Q(pk__in=[]))
```
goosemania: sweet :+1:

pulpcore/plugin/stages/api.py
```diff
@@ -78,7 +78,7 @@ class MyStage(Stage):
             content = await self._in_q.get()
             if content is None:
                 break
-            log.debug(_('%(name)s - next: %(content)s.'), {'name': self, 'content': content})
```
dralley: This translation provided almost no value but was responsible for 6% of the runtime of a file-re-sync.

bmbouter:
I think performance fixes would benefit from a changelog entry. We use `[noissue]` less and less these days. The code changes look great tho!

bmbouter:
Is it correct that this whole workload runs in a single worker? If so, in terms of speeding this up significantly, my recommendation would be to explore running multiple pipelines in parallel across several workers each handling disjoint parts of the pulp2 migration.

dralley:
@bmbouter That's certainly something to explore but it would be great to exhaust all the opportunities present here first, since every improvement made here will also speed up re-sync performance by a huge margin.

bmbouter:
Improving this will help all workflows I agree. I observe though that with each change it's improving by only a few percent. What that tells me is that we'll have to start to do draconian things to our code, e.g. postgres stored procedures, etc, etc to get improvements beyond a few percent. If you want to continue down that road that's ok w/ me. The usual techniques to use there are a mixture of cprofile for improve the python code's runtime, and then SQL EXPLAIN ANALYZE on the queries executed in the stages taking the most time.

My take on the workload is that it is heavily db constrained, and yet we have lots of indexes already and I don't see one missing. Adding indexes is the most common way I know how to make a query "go faster", but I don't see that opportunity since we already have indexes on all the tables involved: https://gist.github.com/bmbouter/20e7743efb0906f25cbd4c45860cfd82

So how to you solve a large problem faster? Divide it into sub-problems! Also from a human effort and benefit to our users perspective a parallel worker approach would give a much higher return I think.

bmbouter:
@dalley I would lgtm this if it could have a changelog entry. How do you feel about adding that?

fabricio-aguiar:
> Even with this improvement, the QueryExistingContents() stage is really expensive. It's 50% (_exactly_ 50%) of the time spent in a resync (of the file plugin), and 33% of the time used in a re-migration (of an RPM repository). @bmbouter @fabricio-aguiar do you have any ideas on how to speed it up further?
> 
> Resync:
> file_resync

The only thing I can think is: replace `OR` with `IN` on:
```python
content_q_by_type[model_type] = content_q_by_type[model_type] | unit_q
```

PR #599, opened 2020-03-19
Checking for dupe content in finalize

daviddavis:
fixes #6362
https://pulp.plan.io/issues/6362

Plugin example: https://github.com/pulp/pulp_file/pull/368

I decided not to update remove_duplicates to handle checking for duplicate content because (a) it would violate the single responsibility principle as remove_duplicates removes existing content from the repo version if it collides with content being added. 

Also (b) [remove_duplicates is called at the beginning of finalize before processing happens](https://github.com/pulp/pulp_rpm/blob/417541e71655f10045f5e6f5f8cb156dff7972ff/pulp_rpm/app/models/repository.py#L127) while validation ought to happen at the end.

### DIFF
--- a/CHANGES/plugin_api/6362.feature
+++ b/CHANGES/plugin_api/6362.feature
@@ -1,0 +1,1 @@
+Added two new repo validation methods (validate_repo_version and validate_duplicate_content).
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -553,33 +553,6 @@ class RepositoryVersion(BaseModel):
         if self.complete:
             raise ResourceImmutableError(self)
 
-        error_messages = []
-        for type_obj in self.repository.CONTENT_TYPES:
-            if type_obj.repo_key_fields == ():
-                continue
-
-            pulp_type = type_obj.get_pulp_type()
-            repo_key_fields = type_obj.repo_key_fields
-            new_content_total = type_obj.objects.filter(
-                pk__in=content.filter(pulp_type=pulp_type)
-            ).count()
-            unique_new_content_total = type_obj.objects.filter(
-                pk__in=content.filter(pulp_type=pulp_type)
-            ).distinct(*repo_key_fields).count()
-
-            if unique_new_content_total < new_content_total:
-                error_messages.append(_(
-                    "More than one {pulp_type} content with the duplicate values for {fields}."
-                    ).format(
-                        pulp_type=pulp_type,
-                        fields=", ".join(repo_key_fields),
-                    )
-                )
-        if error_messages:
-            raise ValueError(
-                _("Cannot create repository version. {msg}").format(msg=", ".join(error_messages))
-            )
-
         repo_content = []
         to_add = set(content.exclude(pk__in=self.content).values_list('pk', flat=True))
 
--- a/pulpcore/plugin/repo_version_utils.py
+++ b/pulpcore/plugin/repo_version_utils.py
@@ -15,7 +15,11 @@ __all__ = ["remove_duplicates"]
 
 def remove_duplicates(repository_version):
     """
-    Inspect content additions in the `RepositoryVersion` and replace repository duplicates.
+    Inspect content additions in the `RepositoryVersion` and remove existing repository duplicates.
+
+    This function will inspect the content being added to a repo version and remove any existing
+    content which would collide with the content being added to the repository version. It does not
+    inspect the content being added for duplicates.
 
     Some content can have two instances A and B which are unique, but cannot both exist together in
     one repository. For example, pulp_file's content has `relative_path` for that file within the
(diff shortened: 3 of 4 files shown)


pulpcore/plugin/repo_version_utils.py
```diff
@@ -82,3 +120,16 @@ def validate_version_paths(version):
         validate_file_paths(paths)
     except ValueError as e:
         raise ValueError(_("Cannot create repository version. {err}.").format(err=e))
+
+
+def validate_repo_version(version):
```
bmbouter: is the idea this would be configured to be called by plugins? or is this called for all plugins today?

daviddavis: Not sure I completely understand your question. This is just a helper method plugins can call during finalize if they want to. Here's an example from pulp_file: https://github.com/pulp/pulp_file/pull/368.

daviddavis: It's sort of like a wrapper for any other validate_* methods in repo_version_utils.

bmbouter: Yup that answers it. You're saying plugin opt into this and this is a helpful wrapper. Sounds good to me.

pulpcore/app/models/repository.py
```diff
@@ -553,33 +553,6 @@ def add_content(self, content):
         if self.complete:
             raise ResourceImmutableError(self)
 
-        error_messages = []
```
bmbouter: To confirm my understanding of this code being removed. It wouldn't automatically remove the duplicate, it would just fail. Is that right?

daviddavis: Exactly. This is the crux of the problem. When duplicate content gets added to a repo version, it fails before giving the plugin writer the chance to deal with the problem during finalize.

pulpcore/plugin/repo_version_utils.py
```diff
@@ -67,6 +67,44 @@ def remove_duplicates(repository_version):
...
+    """
+    error_messages = []
+
+    for type_obj in version.repository.CONTENT_TYPES:
+        if type_obj.repo_key_fields == ():
+            continue
+
+        pulp_type = type_obj.get_pulp_type()
+        repo_key_fields = type_obj.repo_key_fields
+        new_content_total = type_obj.objects.filter(
```
bmbouter: I'm not 100% sure, but I think this would be more efficient if we used queryset caching that django provides. Instead of forming the same queryset twice, maybe form one and then distinguish them with the `.count()` versus `.distinct.....`. I think we'll avoid making the same query twice. I don't think it's a huge deal but I do recommend it. Wdyt?

bmbouter:
Does this need any docs changes? I suspect there needs to be something.

CHANGES/plugin_api/6362.feature
```diff
@@ -0,0 +1 @@
+Added two new repo validation methods (validate_repo_version and validate_duplicate_content).
```
fabricio-aguiar: Great idea!

pulpcore/plugin/repo_version_utils.py
```diff
@@ -67,6 +67,44 @@ def remove_duplicates(repository_version):
                 repository_version.remove_content(duplicates_qs)
 
 
+def validate_duplicate_content(version):
```
fabricio-aguiar: much better! it makes easy for plugin writers to customize

goosemania:
@daviddavis , thanks! Looks good to me.

daviddavis:
@gmbnomis My comment about `remove_duplicates` only removing existing content is about the existing nature of the function and not what it *ought* to be doing based on its name. I agree that it's not clear and could be confusing to plugin writers. I went ahead and updated the docstring for `remove_duplicates`. I fully agree with that move.

As to renaming `remove_duplicates`, I'm a bit torn because it would be a lot of work but I do agree that its name is not entirely accurate. I could go either way. Any thoughts @goosemania @bmbouter @fabricio-aguiar ?

daviddavis:
@gmbnomis I talked on IRC with @bmbouter and @fabricio-aguiar. We'd like to merge this and open a separate issue to consider renaming remove_duplicates since we agreed that its name is not accurate. Wdyt?

gmbnomis:
> @gmbnomis I talked on IRC with @bmbouter and @fabricio-aguiar. We'd like to merge this and open a separate issue to consider renaming remove_duplicates since we agreed that its name is not accurate. Wdyt?

@daviddavis yes, please go ahead. You have addressed the most important bit (updating the documentation). Of course, the name is still misleading, but if you are hesitant to change it because it creates issues with plugins/work for plugins, we can keep it.

daviddavis:
@gmbnomis cool, sounds good. I opened https://pulp.plan.io/issues/6416.

PR #600, opened 2020-03-20
Enabled write_only fields on bindings

fao89:
https://pulp.plan.io/issues/6346
closes #6346

### DIFF
--- a/CHANGES/6346.bugfix
+++ b/CHANGES/6346.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug that prevented write_only fields from being present in the API docs and bindings
--- a/pulpcore/app/openapigenerator.py
+++ b/pulpcore/app/openapigenerator.py
@@ -7,7 +7,7 @@ from drf_yasg import openapi
 from drf_yasg.generators import OpenAPISchemaGenerator
 from drf_yasg.inspectors import SwaggerAutoSchema
 from drf_yasg.openapi import Parameter
-from drf_yasg.utils import filter_none, force_real_str
+from drf_yasg.utils import filter_none, force_real_str, get_serializer_ref_name, no_body
 
 from pulpcore.app.models import RepositoryVersion
 
@@ -439,18 +439,96 @@ class PulpAutoSchema(SwaggerAutoSchema):
 
         return tags
 
-    def serializer_to_schema(self, serializer):
+    def get_default_response_serializer(self):
         """
-        Convert a serializer to an OpenAPI Schema.
-        Patch: https://github.com/axnsan12/drf-yasg/issues/70#issuecomment-485050813
+        Return the default response serializer for this endpoint.
+
+        Reference: https://github.com/axnsan12/drf-yasg/issues/70#issuecomment-485050813
+        """
+        body_override = self._get_request_body_override()
+        if body_override and body_override is not no_body:
+            return body_override
+
+        serializer = super().get_view_serializer()
+
+        if not serializer:
+            return
+
+        read_fields = OrderedDict()
+        for field_name, field in serializer.fields.items():
+            if not field.write_only:  # Removing write_only fields
+                read_fields[field_name] = field
+
+        serializer.fields = read_fields
+        return serializer
+
+    def get_view_serializer(self):
+        """
+        Return the appropriate serializer.
+
+        Reference: https://github.com/axnsan12/drf-yasg/issues/70#issuecomment-485050813
         """
 
-        if self.method.lower() == "get":
-            new_fields = OrderedDict()
-            for field_name, field in serializer.fields.items():
-                if not field.write_only:  # Removing write_only fields
-                    new_fields[field_name] = field
+        serializer = super().get_view_serializer()
 
-            serializer.fields = new_fields
+        if not serializer:
+            return
 
-        return super().serializer_to_schema(serializer)
+        read_fields = OrderedDict()
+        write_fields = OrderedDict()
+        for field_name, field in serializer.fields.items():
+            if not field.write_only:  # Removing write_only fields
+                read_fields[field_name] = field
+            if not field.read_only:  # Removing read_only fields
+                write_fields[field_name] = field
+
+        if len(read_fields) == len(write_fields):
+            return serializer
+
+        contain_write_fields = len(serializer.fields) > len(read_fields)
+        if self.method.lower() == "get" or not contain_write_fields:
+            serializer.fields = read_fields
+            return serializer
+
+        return self._convert_serializer(serializer, WriteOnly)
+
+    def _convert_serializer(self, serializer, new_class):
+        """
+        Convert a serializer into Read/Write only serializer.
+
+        Reference: https://github.com/axnsan12/drf-yasg/issues/70#issuecomment-485050813
+        """
+        serializer_meta = getattr(serializer, "Meta", BlankMeta)
+        name = f"Pulp{get_serializer_ref_name(serializer)}"
+
+        class CustomSerializer(new_class, serializer.__class__):
+            class Meta(serializer_meta):
+                ref_name = name
+
+        new_serializer = CustomSerializer(data=serializer.data)
+        return new_serializer
+
+
+class WriteOnly:
+    """
+    Mixin for removing read_only fields from serializers.
+
+    Reference: https://github.com/axnsan12/drf-yasg/issues/70#issuecomment-485050813
+    """
+
+    def get_fields(self):
+        new_fields = OrderedDict()
+        for file_name, field in super().get_fields().items():
+            if not field.read_only:
+                new_fields[file_name] = field
+        return new_fields
+
+
+class BlankMeta:
+    """
+    Creates a blank Meta for serializers
+
+    Reference: https://github.com/axnsan12/drf-yasg/issues/70#issuecomment-485050813
+    """
+
+    pass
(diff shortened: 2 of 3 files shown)


CHANGES/6346.bugfix
```diff
@@ -0,0 +1 @@
+Enabled `write_only` fields on bindings
```
daviddavis: Any chance you make this more descriptive and include having fixed a bug that prevented write_only fields from being present in the api docs and bindings?

daviddavis:
I tested this and it worked great :+1: :tada:

fabricio-aguiar:
Closing in favor of https://github.com/pulp/pulpcore/pull/611
please read the notes on: https://pulp.plan.io/issues/6346

pulpcore/app/openapigenerator.py
```diff
@@ -383,18 +383,95 @@ def get_tags(self, operation_keys):
...
+
+    def _convert_serializer(self, serializer, new_class):
+        """
+        Convert a serializer into Read/Write only serializer.
 
-        return super().serializer_to_schema(serializer)
+        Reference: https://github.com/axnsan12/drf-yasg/issues/70#issuecomment-485050813
+        """
+        serializer_meta = getattr(serializer, "Meta", BlankMeta)
+        name = f"Pulp{get_serializer_ref_name(serializer)}"
```
fabricio-aguiar: `PulpfileFileRemote`

pulpcore/app/openapigenerator.py
```diff
@@ -383,18 +383,95 @@ def get_tags(self, operation_keys):
...
+
+        if len(read_fields) == len(write_fields):
+            return serializer
+
+        contain_write_fields = len(serializer.fields) > len(read_fields)
+        if self.method.lower() == "get" or not contain_write_fields:
+            serializer.fields = read_fields
+            return serializer
+
+        return self._convert_serializer(serializer, WriteOnly)
```
fabricio-aguiar: "$ref": "#/definitions/Pulpfile.FileRemote"
```python
pulp_file.PulpfileFileRemote(
    name=None,
    url=None,
    ca_cert=None,
    client_cert=None,
    client_key=None,
    tls_validation=None,
    proxy_url=None,
    username=None,
    password=None,
    download_concurrency=None,
    policy='immediate',
    local_vars_configuration=None,
)
```

pulpcore/app/openapigenerator.py
```diff
@@ -383,18 +383,95 @@ def get_tags(self, operation_keys):
...
+        if not serializer:
+            return
+
+        read_fields = OrderedDict()
+        for field_name, field in serializer.fields.items():
+            if not field.write_only:  # Removing write_only fields
+                read_fields[field_name] = field
+
+        serializer.fields = read_fields
+        return serializer
```
fabricio-aguiar: "$ref": "#/definitions/file.FileRemote"
```python
pulp_file.FileFileRemote(
    pulp_href=None,
    pulp_created=None,
    name=None,
    url=None,
    ca_cert=None,
    client_cert=None,
    client_key=None,
    tls_validation=None,
    proxy_url=None,
    pulp_last_updated=None,
    download_concurrency=None,
    policy='immediate',
    local_vars_configuration=None,
)
```

fabricio-aguiar:
Reopening it, as an option for #613

pulpcore/app/openapigenerator.py
dkliban: Which model will be called PulpfileFileRemote? Is it the one used for Creating or for Reading?

fabricio-aguiar: "Pulp" + ref_name will be used for write_only models

bmbouter: I think having Read prepended to the name for the generated Read serializer would be the best. I conclude this from two reasons:

* the in-code definition includes both read and write fields, so the one that gets the custom name is the one not stored in-code, so that would be the Write serializer
* having 'Write' prepended is the most self-descriptive name I could think of

Does this make sense? What do you think?

dkliban: Are you suggesting a name such as WriteFileFileRemote? and are you also suggesting ReadFileFileRemote? or just the former? I was thinking that we could append Write instead, e.g. FileFileRemoteWrite

bmbouter: After reading your note I think `FileFileRemoteWrite` is the best. It would keep them grouped more similarly and makes more sense to me than what I was thinking before.

bmbouter: For the read side should it be `FileFileRemote` or `FileFileRemoteRead` ?

dkliban: It would be nice if we could do FileFileRemoteRead, but I am not sure how easy that would be to implement. What do you think @fabricio-aguiar ?

fabricio-aguiar: it is easier to make the change, but my concern is:
what people have been using so far is: `FileFileRemote`, if I change to `FileFileRemoteRead` it could be a breaking change for katello if they are using some of these models. 
My intention with this PR was:
 - Hey user, you've been using, `FileFileRemote`, here's a new model (FileFileRemoteWrite) for you in case you need some write fields

fabricio-aguiar: Here a practical case:
https://github.com/pulp/pulpcore/blob/master/.travis/test_bindings.py#L134
people generally use `RepositorySyncURL` which would be `RepositorySyncURLRead`

pulpcore/app/serializers/repository.py
```diff
@@ -83,15 +83,15 @@ class RemoteSerializer(ModelSerializer):
...
+                  'checksum of the username on GET.',
         required=False,
         allow_null=True,
     )
-    password = serializers.CharField(
-        help_text='The password to be used for authentication when syncing.',
-        write_only=True,
+    password = SecretCharField(
+        help_text='The password to be used for authentication when syncing. Returns SHA256 '
+                  'checksum of the password on GET.',
```
mdellweg, line 94: Is that a given for `SecretCharField`? I once learned that hashing for cryptographic reasons should always be salted to be safe from rainbow-table attacks.
I know this is a little bit off topic.

daviddavis: I think this is a good point and one I raised when we were discussing converting write_only fields to SecretCharFields. The latter is less secure especially for things like passwords.

dkliban:
The openapi generator should only modify serializers that have write_only fields. The modification should produce two serializers: OriginalNameRead and OriginalNameWrite. 

Any serializer that is only used for 'write only' operations should not need to have fields designated as write_only.

fabricio-aguiar:
> Any serializer that is only used for 'write only' operations should not need to have fields designated as write_only.

how can I determine if a serializer is only used for 'write only'?

pulpcore/app/openapigenerator.py
dkliban: > > Any serializer that is only used for 'write only' operations should not need to have fields designated as write_only.
> 
> how can I determine if a serializer is only used for 'write only'?

Good question. We have some serializers that we only use for processing POST requests. e.g.: RepositorySyncURLSerializer or RecursiveManageSerializer in pulp_container. We should remove write_only=True from those serializers because we know that they are only used for POST requests.

daviddavis: This is why I suggested we address https://pulp.plan.io/issues/6421 before continuing on this PR. I suspect there are a few instances of write_only that probably shouldn't be write_only.

pulpcore/app/serializers/repository.py
bmbouter: If we go forward with SecretCharField we should use a salt, but first I'm going to share a perspective about why we shouldn't use RBAC instead of it.

### Use RBAC instead

The SecretCharField is definitely needed if we continue to allow all users to GET data for all Remotes (in this example). We are about to being a push for RBAC though, and what if we resolve the concern of one user reading the data from another with that mechanism instead? I think that would be better because we would avoid:

* The GET-then-PATCH issue where the user accidentally overwrites the data of a SecretCharField with its sha result would be totally avoided
* It would be more usable because the user wouldn't have to compute the hash(orig_password) or hash(salt + orig_password)
* It would be secure

Optional idea: we could have fine-grained permissions that allow you to read all fields on a Remote or just the "non-sensitive" fields perhaps...


### Use a salt if that's what we are going to use SecretCharField

If we do keep `SecretCharField` I'm convinced a salt is a good idea. Based on my reading the hash rate of a GPU or FPGA hardware-dedicated hashers along with their pricepoints "short" passwords are definitly at risk. What is too short has too many factors to name here, but I'd say 8-16 chars per passwords are likely crackable without a crazy amount of time or money. Thus using a long salt would push the keyspace out enough that the attach would become totally infeasable, say a 64 char salt.

mdellweg:
> The openapi generator should only modify serializers that have write_only fields. The modification should produce two serializers: OriginalNameRead and OriginalNameWrite.

I think, i do not agree, because this would render the addition of the first write_only field to an existing serializer a breaking change in the face of the consumer of the bindings.

dkliban:
> > The openapi generator should only modify serializers that have write_only fields. The modification should produce two serializers: OriginalNameRead and OriginalNameWrite.
> 
> I think, i do not agree, because this would render the addition of the first write_only field to an existing serializer a breaking change in the face of the consumer of the bindings.

That is a good point. I was only thinking about already existing serializers when I sent a note to pulp-list[0]. However, I did not think aobut future changes that might add write_only fields.

I propose that we stop using write_only fields. That's the only way to stop this from happening in the future. 

[0] https://www.redhat.com/archives/pulp-list/2020-April/msg00031.html

fao89:
Closing in favor of https://github.com/pulp/pulpcore/pull/695

fao89:
New PR: https://github.com/pulp/pulpcore/pull/707

PR #601, opened 2020-03-21
Normalize PK path parameter names used in openapi schema generation

dkliban:
This ensures that when the endpoints are sorted, the parent endpoint appears before the
endpoints nested under it.

fixes: #6372
https://pulp.plan.io/issues/6372


pulpcore/app/openapigenerator.py
```diff
@@ -1,3 +1,4 @@
+
```
daviddavis: nitpick - unneeded.

PR #607, opened 2020-03-24
Added ALLOWED_EXPORT_PATHS setting

daviddavis:
fixes #6335


pulpcore/app/models/exporter.py
```diff
@@ -100,7 +100,13 @@ def _export_to_file_system(self, content_artifacts):
 
         Args:
             content_artifacts (django.db.models.QuerySet): Set of ContentArtifacts to export
+
+        Raises:
+            ValidationError: When path is not in the ALLOWED_EXPORT_PATHS setting
         """
+        from pulpcore.app.serializers import ExportSerializer
```
bmbouter: Circular import avoided?

bmbouter: I want to confirm, but we have this issue in several places so we shouldn't solve with this piece of work here.

daviddavis: Haha yes. I remember [the discussion we had on the import PR where I suggested we should leave a comment](https://github.com/pulp/pulpcore/pull/520#discussion_r372606433). I can do that if it's not clear.

pulpcore/app/serializers/exporter.py
```diff
@@ -24,6 +25,25 @@ class ExporterSerializer(ModelSerializer):
...
+            value: The user-provided value path to be validated.
+
+        Raises:
+            ValidationError: When path is not in the ALLOWED_EXPORT_PATHS setting.
+
+        Returns:
+            The validated value.
+        """
+        for allowed_path in settings.ALLOWED_EXPORT_PATHS:
+            user_provided_realpath = os.path.realpath(value)
```
bmbouter: :+1:

pulpcore/app/models/exporter.py
bmbouter: nope, this is fine.

PR #610, opened 2020-03-25
Port pulpcore-content to use Application Runners

bmbouter:
Application Runners in aiohttp Server are the recommended way to
configure additional aspects of aiohttp such configuring the header
length. Plugins like `pulp-certguard` require this additional
configuration.

This PR ports the `pulp-content` app to use Application Runners. Also
port the `pulpcore.content.app` entry point that gunicorn uses to it
also.

https://pulp.plan.io/issues/6397
closes #6397

### DIFF
--- a/bin/pulp-content
+++ b/bin/pulp-content
@@ -1,8 +1,24 @@
 #!/usr/bin/env python
 
+import asyncio
+
 import aiohttp
+
+import django  # noqa otherwise E402: module level not at top of file
+django.setup()  # noqa otherwise E402: module level not at top of file
+
+from django.conf import settings
 
 from pulpcore.content import server
 
 
-aiohttp.web.run_app(server(), port=24816)
+async def work():
+    pulp_content_app = await server()
+    runner = aiohttp.web.AppRunner(pulp_content_app, max_line_size=16380, max_field_size=16380)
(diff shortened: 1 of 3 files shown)


bin/pulp-content
```diff
@@ -1,8 +1,24 @@
 #!/usr/bin/env python
 
+import asyncio
+
 import aiohttp
 
+import django  # noqa otherwise E402: module level not at top of file
+django.setup()  # noqa otherwise E402: module level not at top of file
+
+from django.conf import settings
```
mdellweg, line 10: I do not see those settings to be used anywhere (yet?).

pulpcore/content/worker.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class PulpGunicornWebWorker(GunicornWebWorker):
+
+    def _get_application_runner_instance(self, app, **kwargs):
+        return super()._get_application_runner_instance(
+            app,
+            max_line_size=16380,
+            max_field_size=16380
+        )
```
mdellweg, line 11: Is this `PulpGunicornWebWorker` used anywhere?

daviddavis:
Closing due to inactivity.

PR #616, opened 2020-03-27
Upload chunks to a local file system

lubosmj:
As of now, every single chunk is going to be "uploaded" to a local file system instead of the defined default file storage.

In addition to that, the test test_delete_upload was refactored because it did not properly verify whether the committed upload was removed from the file system or not. It rather tested if the upload was deleted after sending an HTTP DELETE request.

closes #6253


CHANGES/6253.feature
```diff
@@ -0,0 +1 @@
+Made chunked uploads to be stored in a local file system instead of a default file storage
```
daviddavis: I was thinking this is probably a bugfix or removal instead of a feature?

daviddavis: Maybe the feature aspect of this change is the new setting `CHUNKED_UPLOAD_DIR`.

pulpcore/app/settings.py
```diff
@@ -40,6 +40,7 @@
 
 FILE_UPLOAD_TEMP_DIR = os.path.join(MEDIA_ROOT, 'tmp/')
 WORKING_DIRECTORY = os.path.join(MEDIA_ROOT, 'tmp/')
+CHUNKED_UPLOAD_DIR = os.path.join(MEDIA_ROOT, 'tmp/')
```
daviddavis: I'd probably separate out CHUNKED_UPLOAD_DIR into a new directory called 'upload/'.

pulpcore/app/models/upload.py
```diff
@@ -1,24 +1,28 @@
...
 
+from django.conf import settings
 from django.core.files.base import ContentFile
+from django.core.files.storage import FileSystemStorage
 from django.db import models
 from rest_framework import serializers
 
 from pulpcore.app.models import BaseModel
 
+tmp_fs = FileSystemStorage(location=settings.CHUNKED_UPLOAD_DIR)
```
daviddavis: What's the reasoning behind defining this variable?

daviddavis:
Overall, this looks great. A couple small comments.

pulpcore/app/settings.py
lubosmj: Uploads are already stored to the directory "upload" https://github.com/pulp/pulpcore/blob/f9731dfa128d9b690192899700b1c0a20d62897b/pulpcore/app/models/upload.py#L33

pulpcore/app/models/upload.py
lubosmj: I wanted to clearly state that this FileSystemStorage is temporary. A may do the following in the model as well:
```suggestion
    file = models.FileField(null=False, max_length=255, storage=tmp_fs = FileSystemStorage(location=settings.CHUNKED_UPLOAD_DIR))
```

PR #617, opened 2020-03-30
Allow tasks to spawn other tasks

dralley:
And record their relationships in the database.

closes: #6282
https://pulp.plan.io/issues/6282


dralley:
Still missing:

* migration (vagrant box still loading)
* proper tests

CHANGES/6282.feature
```diff
@@ -0,0 +1 @@
+Tasks can now spawn other tasks, and the parent/child relationships are recorded in the database.
```
bmbouter: I think two release notes would be more helpful for our two readers personas. The field additions to the task serializer make sense for users, but a `plugin_api/6282.feature` for plugin writer identifying that they can now dispatch tasks from tasks would make sense for them.

pulpcore/app/viewsets/task.py
```diff
@@ -32,18 +32,21 @@ class TaskFilter(BaseFilterSet):
...
         model = Task
         fields = {
+            'name': ['contains'],
             'state': ['exact', 'in'],
             'worker': ['exact', 'in'],
-            'name': ['contains'],
             'started_at': DATETIME_FILTER_OPTIONS,
             'finished_at': DATETIME_FILTER_OPTIONS,
+            'parent_task': ['exact'],
+            'child_tasks': ['exact'],
```
dkliban: Maybe 'exact' and 'in' is appropriate for child_tasks?

dralley: ```in``` causes problems.  I can't debug it without having some appropriate data first -- still trying to get everything manually tested.

dralley:
Screenshot from 2020-03-30 17-10-34

pulpcore/app/viewsets/task.py
dralley: Just having ```in``` in the list of acceptable formats causes this:

```
Mar 30 21:14:47 pulp2-nightly-pulp3-source-centos7.localhost.example.com gunicorn[14656]: File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django_filters/widgets.py", line 201, in value_from_datadict
Mar 30 21:14:47 pulp2-nightly-pulp3-source-centos7.localhost.example.com gunicorn[14656]: return value.split(',')
Mar 30 21:14:47 pulp2-nightly-pulp3-source-centos7.localhost.example.com gunicorn[14656]: AttributeError: 'list' object has no attribute 'split'
```

I'm not inclined to debug this right now unless it is absolutely necessary.  In this circumstance ```exact``` actually behaves more like you would expect ```contains``` to behave.  You can only provide one value at a a time, but it will match any task that contains it regardless of how many others it contains.

Screenshot from 2020-03-30 17-13-33

PR #618, opened 2020-03-30
Added Pulp import support

daviddavis:
fixes #6329


pulpcore/app/models/importer.py
```diff
@@ -0,0 +1,80 @@
...
+    A class that maps upstream repo names to downstream repositories.
+
+    Fields:
+
+        upstream_repo (models.TextField): The name of the upstream repo
+
+    Relations:
+
+        pulp_importer (models.ForeignKey): The associated pulp importer
+        repository (models.ForeignKey): The downstream repository
```
ggainey: We prob want to add some commentary about our use of upstream/downstream and export/import

pulpcore/app/serializers/fields.py
```diff
@@ -320,6 +320,48 @@ def get_object(self, view_name, view_args, view_kwargs):
         return self.get_queryset().get(**lookup_kwargs)
 
 
+class ImportsIdentityFromImporterField(DetailIdentityField):
```
daviddavis, line 323: This code was copied over from repo version and export. I have a task to refactor all of these fields: https://pulp.plan.io/issues/6389

pulpcore/app/models/importer.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Fields:
+        name (models.TextField): The importer unique name.
+    """
+    name = models.TextField(db_index=True, unique=True)
+
+
+class PulpImporter(Importer):
+    """
+    A class that can be used to import exports from other Pulp instances.
```
bmbouter, line 39: nitpick: can all models s/class/model/.

pulpcore/app/models/importer.py
```diff
@@ -0,0 +1,82 @@
...
+    Relations:
+        task (models.ForeignKey): The Task that ran the import
+        importer (models.ForeignKey): The Importer that imported the export
+    """
+    params = JSONField(null=True)
+    task = models.ForeignKey("Task", on_delete=models.PROTECT)
+    importer = models.ForeignKey("Importer", on_delete=models.CASCADE)
+
+
+class Importer(MasterModel):
```
bmbouter, line 27: Does this have to be subclassed or should it receive a concrete table in the DB? If it does have to be subclassed I think its metaclass get `abstract = True`. Also if it has to be subclassed, the docstring could call that out also.

pulpcore/app/models/repository.py
```diff
@@ -612,6 +612,20 @@ def remove_content(self, content):
             version_removed=None)
         q_set.update(version_removed=self)
 
+    def set_content(self, content):
+        """
+        Use add_content and remove_content to update repo version content.
```
bmbouter, line 617: I don't understand how this descrpition relates to `set_content`'s functionlity itself.

bmbouter: After reading the code I do though. Maybe just calling out that it sets the content using add_content and remove_content...

bmbouter: If this is clear enough for you all you can ignore this request.

pulpcore/app/models/importer.py
daviddavis: I thought the fact it says it's a "base class" in the docstring calls that out. But I can be more clear.

As for `abstract = True`. I checked other MasterModels and don't see that set?

pulpcore/app/models/repository.py
daviddavis: Not sure I fully understand your suggestion here. If I update this to `Sets the repo version content using add_content and remove_content` would that work?

bmbouter: I think `Sets the repo version content by calling remove_content() then add_content()` would have caused me to understand it.

pulpcore/app/models/importer.py
bmbouter: I think we are supposed to use it in this case since our intent is to share the field across other models but not to make this itself a model https://docs.djangoproject.com/en/3.0/topics/db/models/#abstract-base-classes

The reason I think it's not used in `MasterModel` itself is because since it's a multi-table inheritance design MasterModel actually is a concrete table.

bmbouter: My suggestion does not make sense. This is the "Base Model" and `PulpImporter` is the "Master Model" so both need to be concrete.

PR #619, opened 2020-03-31
Problem: Pulp containers do not allow commands like "/bin/bash"

mikedep333:
to be run, only "bash"

Solution: Only apply the /bin/ logic for the 4 pulp-* commands.

fixes: #6420
Pulp containers do not allow commands like "/bin/bash" to be run, only "bash"
https://pulp.plan.io/issues/6420


mikedep333:
FYI, I tested running several `docker run` commands as well. To make sure that the quotes around "$@" do not interfere.

PR #621, opened 2020-03-31
As a user, I have Task Groups

dralley:
closes: #6414
https://pulp.plan.io/issues/6414


dralley:
TODO:

* TaskGroup serializes counts of tasks in each state
* Tests

dralley:
Automated tests for this are just as difficult as for task parent/children, so @bmbouter @dkliban If you want to just merge this with some manual testing, I'll file an issue to eventually write automated tests for both.

pulpcore/app/serializers/task.py
```diff
@@ -117,6 +123,27 @@ class Meta:
                                                 'worker')
 
 
+class TaskGroupSerializer(ModelSerializer):
+    description = serializers.CharField(
+        help_text=_("A description of the task group.")
+    )
+
+    waiting = TaskGroupStatusCountField(state='waiting')
```
dkliban, line 131: What about some help text for these fields?

bmbouter:
@dralley I'm +1 to merging this with your hand testing done and filing auotmated tests for later.

One idea I had for the automated tests is that we should have some way for them to load additional viewsets and tasks in the RQ registry if there is a PULP_TEST env var set or something. Something that keeps them unloaded on production systems and the tests automatically skip if they aren't loaded. Just an idea I had.

PR #622, opened 2020-03-31
Rename `pulp-content` script to `pulpcore-content`

bmbouter:
The `bin/pulp-content` script was never renamed along with the other
systemd service names. This has been renamed for consistency.

Note: this is not related to the container name `pulp-content` which is
still the correct container name.

https://pulp.plan.io/issues/6403
closes #6403


mikedep333:
I thought I submitted an issue a few months ago about renaming it, but for other reasons.

Let me look for it.

mikedep333:
Yeah, this is the issue:
https://pulp.plan.io/issues/6130

PR #628, opened 2020-04-03
Repair repository version

mdellweg:
fixes #5613

### DIFF
--- a/CHANGES/5613.feature
+++ b/CHANGES/5613.feature
@@ -1,0 +1,4 @@
+Added support for repairing a RepositoryVersion by redownloading corrupted artifact files.
+Sending a POST request to
+``/pulp/api/v3/repositories/<plugin>/<type>/<repository-uuid>/versions/<version-number>/repair/``
+will trigger a task that scans all associated artfacts and attempts to fetch missing or corrupted ones again.
--- a/docs/workflows/repairing-repository-versions.rst
+++ b/docs/workflows/repairing-repository-versions.rst
@@ -1,0 +1,67 @@
+.. _repository-version-repairing:
+
+Repository Version Repairing
+----------------------------
+
+.. note::
+
+    Repairing a Repository Version is in tech-preview and may change in backwards incompatible
+    ways in future releases.
+
+
+Pulp allows users to repair repository versions where artifacts in the storage went missing or
+got corrupted in some way (i.e. bit rot).
+
+.. warning::
+
+    This repair feature looks for missing or corrupted files that are supposed to be in the
+    internal storage. It attempts a redownload of these files from known upstream urls.
+    In case upstream stopped serving the corresponding files Pulp is unable to remedy the
+    situation.
+
+To start a repair task on a specific repository version, send a POST request to its `repair`
+endpoint:
+
+.. code-block:: bash
+
+    $ REPAIR_TASK=$(http POST :24817${REPOSITORY_VERSION}repair/ | jq -r '.task')
+    $ http --body :24817$REPAIR_TASK
+
+    {
+        "child_tasks": [],
+        "created_resources": [],
+        "error": null,
+        "finished_at": "2020-04-07T08:36:52.373633Z",
+        "name": "pulpcore.app.tasks.repository.repair_version",
+        "parent_task": null,
+        "progress_reports": [
+            {
+                "code": "repair.repaired",
+                "done": 2,
+                "message": "Repair corrupted units",
+                "state": "completed",
+                "suffix": null,
+                "total": null
+            },
+            {
+                "code": "repair.corrupted",
+                "done": 2,
+                "message": "Identify corrupted units",
+                "state": "completed",
+                "suffix": null,
+                "total": null
+            }
+        ],
+        "pulp_created": "2020-04-07T08:36:52.274985Z",
+        "pulp_href": "/pulp/api/v3/tasks/530302b4-8674-4db3-8a13-99febef80830/",
+        "reserved_resources_record": [
+            "/pulp/api/v3/repositories/file/file/47a3f651-aaa6-4026-b649-130c45ab38ea/"
+        ],
+        "started_at": "2020-04-07T08:36:52.348381Z",
+        "state": "completed",
+        "task_group": null,
+        "worker": "/pulp/api/v3/workers/f2fe2811-74a1-463f-93d2-53c7b302115c/"
+    }
+
+The result of this task can be read in the `progress_report` section.
+If the number of `done` differs between the reports, pulp was unable to repair all artifacts.
--- a/pulpcore/app/tasks/repository.py
+++ b/pulpcore/app/tasks/repository.py
@@ -1,5 +1,8 @@
 from gettext import gettext as _
 from logging import getLogger
+from functools import partial
+import asyncio
+import hashlib
 
 from django.db import transaction
 
@@ -47,7 +50,7 @@ def delete_version(pk):
     guarantee.
 
     Args:
-        pk (int): the primary key for a RepositoryVersion to delete
+        pk (uuid): the primary key for a RepositoryVersion to delete
 
     Raises:
         models.RepositoryVersion.DoesNotExist: if there is not a newer version to squash into.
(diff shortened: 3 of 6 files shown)


pulpcore/app/tasks/repository.py
```diff
@@ -66,6 +68,58 @@ def delete_version(pk):
         version.delete()
 
 
+def repair_version(pk):
+    """
+    Repair the artifacts associated with this repository version that have suffered from bit rod.
```
dkliban, line 73: s/rod/rot/

pulpcore/app/tasks/repository.py
```diff
@@ -66,6 +68,58 @@ def delete_version(pk):
...
+
+    Args:
+        pk (int): the primary key for a RepositoryVersion to delete
```
dkliban, line 76: s/delete/repair/

pulpcore/tests/functional/api/using_plugin/test_repair.py
```diff
@@ -0,0 +1,105 @@
...
+        # STEP 2
+        content = choice(get_content(repo)[FILE_CONTENT_NAME])
+        if "FileSystem" in settings.DEFAULT_FILE_STORAGE:
+            # Verify that the artifact is present on disk.
+            artifact_path = os.path.join(MEDIA_PATH,
+                                         self.api_client.get(content['artifact'])['file'])
+            cmd = ('sed', '-i', '-e', r'$a bit rot', artifact_path)
+            self.cli_client.run(cmd, sudo=True)
+        else:
+            self.fail("Corrupting files on this storage platform is not supported.")
```
dkliban, line 82: s/Corrupting/Repairing/

pulpcore/app/tasks/repository.py
mdellweg:
```suggestion
    Repair the artifacts associated with this repository version that have suffered from bit rot.
```

pulpcore/app/tasks/repository.py
mdellweg:
```suggestion
        pk (int): the primary key for a RepositoryVersion to repair
```

pulpcore/tests/functional/api/using_plugin/test_repair.py
mdellweg: No, actually this is about having direct access to the file from the testing code to create a situation in which repairing is desired.

pulpcore/app/tasks/repository.py
```diff
@@ -66,6 +68,58 @@ def delete_version(pk):
...
+    """
+
+    def _repair_ca(content_artifact):
```
bmbouter, line 79: I think we should make this either a utility function outside of this function, or transform it into a class with two methods and make `_repair_ca` static.

mdellweg: At some point at least i want to turn it into an async callback, but that might just not affect your suggestion.

mdellweg: Actually, we can move this method to the ContentArtifact model if you like.

pulpcore/app/tasks/repository.py
```diff
@@ -66,6 +68,57 @@ def delete_version(pk):
         version.delete()
 
 
+def repair_version(pk):
+    """
+    Repair the artifacts associated with this repository version that have suffered from bit rod.
```
bmbouter, line 73: s/rod/rot/

bmbouter: Also maybe just referring to them as corrupted or missing is better because it captures the likely more common case that it's been a user modification that went wrong.  Fitting that into 1 sentence for PEP 257 tho may be challenging.

pulpcore/app/tasks/repository.py
```diff
@@ -66,6 +68,57 @@ def delete_version(pk):
         version.delete()
 
 
+def repair_version(pk):
```
bmbouter, line 71: s/pk/repository_version_pk/ would be more descriptive (so I think it's better).

pulpcore/app/tasks/repository.py
```diff
@@ -66,6 +68,57 @@ def delete_version(pk):
...
+    """
+    Repair the artifacts associated with this repository version that have suffered from bit rod.
+
+    Args:
+        pk (int): the primary key for a RepositoryVersion to delete
```
bmbouter, line 76: The pk's are uuids.

pulpcore/app/tasks/repository.py
```diff
@@ -66,6 +68,57 @@ def delete_version(pk):
...
+        for remote_artifact in content_artifact.remoteartifact_set.all():
+            downloader = remote_artifact.remote.get_downloader(remote_artifact)
+            dl_result = loop.run_until_complete(downloader.run())
+            if dl_result.artifact_attributes["sha256"] == content_artifact.artifact.sha256:
+                with open(dl_result.path, "rb") as src:
+                    filename = content_artifact.artifact.file.name
+                    content_artifact.artifact.file.delete(save=False)
+                    content_artifact.artifact.file.save(filename, src, save=False)
+                return True
+            log.warn("Redownload failed from {}.".format(remote_artifact.url))
```
bmbouter, line 90: Can we i8ln these strings plz?

pulpcore/app/tasks/repository.py
bmbouter: Moving this to `ContentArtifact` with an async internal method and a more descriptive name is probably what I recommend rather than a stray utility function that takes a reference to the ContentArtifact. What about a name like `_check_and_redownload_if_corrupted` or some descriptive name like that?

bmbouter: Although .... maybe making it an async function that handles a single content unit will not correctly parallelize this because when `run_until_complete` instructs one of the ContentArtifact async co-routines to run it won't yield back control to the reactor. I need to think some about it.

bmbouter: Ok I thought about it some, and I think as it's written now you will have the issue described, and the trick to avoid is that you wouldn't run coroutines inside of a coroutine with the run_until_complete() instead you'd call it with the `await` keyword which does yield control back to the reactor. So making it a coroutine internal method on ContentArtifact will work fine, you'll just have to change its implementation to use `await` instead of `run_until_complete`.

Feel free to setup a 1-hour time for you and I to meet on Monday to talk it over if you want; my calendar is up to date.

Overall great work so far!

mdellweg: Oh, yes. That is roughly what i meant with "turn into async".

pulpcore/app/viewsets/repository.py
```diff
@@ -214,6 +215,26 @@ def destroy(self, request, repository_pk, number):
         )
         return OperationPostponedResponse(async_result, request)
 
+    @swagger_auto_schema(
+        operation_description="Trigger an asynchronous task to delete "
+                              "a repositroy version.",
+        responses={202: AsyncOperationResponseSerializer}
+    )
+    @action(detail=True, methods=['post'])
```
mdellweg, line 223: Is this the right method? I think it should _not_ be `GET`, but it could be any P action like `PATCH`, `POST` or `PUT`.

CHANGES/5613.feature
```diff
@@ -0,0 +1 @@
+Added a repair endpoint and task to RepositoryVersion to redownload corrupted artifact files.
```
bmbouter, line 1: Can we add the url endpoint written out in this release note please? It uses sphinx notation so that would need to be surrounded by double backticks. I usually proof my note on https://livesphinx.herokuapp.com/

CHANGES/5613.feature
```diff
@@ -0,0 +1,4 @@
+Added support for repairing a RepositoryVersion by redownloading corrupted artifact files.
+Sending a POST request to
+``/pulp/api/v3/repositories/<plugin>/<type>/<repository-uuid>/versions/<version-number>/repair/``
+will trigger a task that scans all associated artfacts and attempts do fetch missing or corrupted ones again.
```
bmbouter, line 4: s/do/to/

docs/workflows/repairing-repository-versions.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Pulp allows users to repair repository versions where artifacts in the storage went missing or
+got corrupted in some way (i.e. bit rot).
+
+.. warning::
+
+    This repair feature looks for missing or corrupted files that are supposed to be in the
+    internal storage. It attempts a redownload of these files from known upstream urls.
+    In case upstream stopped serving the corresponding files Pulp is unable to remedy the
+    situation.
```
bmbouter, line 20: Consider adding to the end of this sentence:   "... and the task will fail".

docs/workflows/repairing-repository-versions.rst
```diff
@@ -0,0 +1,67 @@
...
+        "reserved_resources_record": [
+            "/pulp/api/v3/repositories/file/file/47a3f651-aaa6-4026-b649-130c45ab38ea/"
+        ],
+        "started_at": "2020-04-07T08:36:52.348381Z",
+        "state": "completed",
+        "task_group": null,
+        "worker": "/pulp/api/v3/workers/f2fe2811-74a1-463f-93d2-53c7b302115c/"
+    }
+
+Plugin writers are then able to sign selected content by the provided script. To learn more
```
bmbouter, line 66: Is this paragraph supposed to be here?

docs/workflows/repairing-repository-versions.rst
mdellweg: It will not fail. It will report a different number of repaired than corrupted entries.

But i can add this...

docs/workflows/repairing-repository-versions.rst
mdellweg: nooooooo

pulpcore/app/tasks/repository.py
```diff
@@ -66,6 +69,81 @@ def delete_version(pk):
...
+    Artifact files that have suffered from bit rot, were altered or have gone missing will be
+    attempted to be refetched fron an associated upstream.
+
+    Args:
+        repository_version_pk (uuid): the primary key for a RepositoryVersion to delete
+    """
+
+    try:
+        version = models.RepositoryVersion.objects.get(pk=repository_version_pk)
+    except models.RepositoryVersion.DoesNotExist:
```
bmbouter, line 136: I expected this exception to raise actually.

pulpcore/app/viewsets/repository.py
bmbouter: There is a bit of grey area here, but I think POST is right actually. PATCH isn't right because the stateful resource doesn't yet exist. Also PUT isn't right because the client isn't determining the URL portion of the resource to be created. The resource created is a task and the server side selects the UUID. So I think what you have here is right.

pulpcore/tests/functional/api/using_plugin/test_repair.py
```diff
@@ -0,0 +1,120 @@
...
+    set_up_module as setUpModule
+)
+
+
+class RepairRepositoryVersionTestCase(unittest.TestCase):
+    """Test whether corrupted files can be redownloaded.
+
+    This test targets the following issues:
+
+    * `Pulp #5613 <https://pulp.plan.io/issues/5613>`_
```
bmbouter, line 34: I try not to tie these to issues since a feature could be spread across many issues over the lifetime of the codebase. Instead I recommend describing the feature it's testing in its own terms here. This is a style preference, so feel free to ignore.

pulpcore/app/tasks/repository.py
mdellweg: sure.

docs/workflows/repairing-repository-versions.rst
bmbouter: It's fine as is, that is a reasonable outcome.

bmbouter: Actually more than reasonable, it needs to stay as is otherwise we won't be able to complete the task as the user or katello expects.

PR #629, opened 2020-04-03
PulpExporter backend, first cut

ggainey:
Wires up PulpExporter to PulpExport to producing an export.
- supports pulpcore entities only
- exports to filesystem-tree
- full-exports only

fixes #6135


CHANGES/6135.feature
```diff
@@ -0,0 +1,3 @@
+Filled in wiring for PulpExporter/PulpExport/django-import-export code. Invoking the ```/pulp/api/v3/exporters/core/pulp/``` API will create a PulpExporter. POSTing to that exporter via ```/pulp/api/v3/exporters/core/pulp/<exporter-uuid>/exports/``` will generate an export-tar.gz file at ```<exporter.path>/export-<export-UUID>-YYYYMMDD_hhMM.tar.gz```
+
+
```
daviddavis, line 3: A few things:

* You have some extra line breaks here
* I'd recommend wrapping lines at 100 chars
* These files are rST so you use double backticks to escape text

pulpcore/app/importexport.py
```diff
@@ -0,0 +1,115 @@
...
+    resource = ContentArtifactResource(repository_version)
+    _write_export(export.tarfile, resource, dest_dir)
+
+    resource = RepositoryVersionResource(repository_version)
+    _write_export(export.tarfile, resource, dest_dir)
+
+    # find and export any ModelResource found in pulp_<repo-type>.app.modelresource
+    import importlib
+    import inspect
+    from pulpcore.app.apps import get_plugin_config
```
daviddavis, line 102: Can any of these imports be moved to the top of the file? I suspect importlib and inspect can?

pulpcore/app/models/exporter.py
```diff
@@ -164,3 +147,43 @@ def export_repository_version(self, repository_version):
...
         abstract = True
+
+
+class PulpExport(Export):
+    """
+    A class that provides export-files that can be imported into other Pulp instances.
+
+    Fields:
+
+        tarfile : a full path for this particular export-instance's export-tar-gz-file
```
daviddavis, line 158: It's an actual tar object and not a path, no?

pulpcore/app/tasks/export.py
```diff
@@ -44,3 +58,77 @@ def fs_repo_version_export(exporter_pk, repo_version_pk):
         ).format(exporter=exporter.name, repo_version=repo_version.pk, path=exporter.path)
     )
     exporter.export_repository_version(repo_version)
+
+
+def export_tarfile(pulp_export):
```
daviddavis, line 63: I'd probably move this to be a method or property on PulpExport.

daviddavis: Or rather, I'd have the method/property return the full path to the tarfile.

daviddavis:
A couple things:

* A lot of the docstring still refers to exporting to the filesystem. Probably needs to be updated.
* This should work even if artifacts are stored in S3. I suspect it doesn't but I'm cool with filing a follow up issue to address this.

ggainey:
> * A lot of the docstring still refers to exporting to the filesystem. Probably needs to be updated.

I think I caught most of these now

> * This should work even if artifacts are stored in S3. I suspect it doesn't but I'm cool with filing a follow up issue to address this.

Def will need its own issue

CHANGES/6135.feature
```diff
@@ -0,0 +1,4 @@
+Filled in wiring for PulpExporter/PulpExport/django-import-export code. Invoking the
```
bmbouter, line 1: What about removing this first sentence? As a user it would confuse me. The rest of the changelog is really nice.

pulpcore/app/importexport.py
```diff
@@ -0,0 +1,113 @@
...
+    RepositoryVersionResource,
+    RepositoryResource,
+)
+
+from pulpcore.app.settings import MEDIA_ROOT
+
+
+def _write_export(the_tarfile, resource, dest_dir=None):
+    """
+    Write the JSON export for the specified resource to the specified tarfile. The resulting file
```
bmbouter, line 26: We don't have a linter checking it (sorry), but for PEP 257 compliance can a single sentence be added above this summarizing the docstring. Similar for all docstrings in this PR.

pulpcore/app/importexport.py
```diff
@@ -0,0 +1,113 @@
...
+    ContentResource,
+    ContentArtifactResource,
+    RepositoryVersionResource,
+    RepositoryResource,
+)
+
+from pulpcore.app.settings import MEDIA_ROOT
+
+
+def _write_export(the_tarfile, resource, dest_dir=None):
```
bmbouter, line 24: Maybe this should mentoin it's using UTF8 as the encoded format that it produces? I'm not sure, just a thought.

pulpcore/app/importexport.py
```diff
@@ -0,0 +1,113 @@
...
+
+    Args:
+        export (django.db.models.PulpExport): export instance that's doing the export
+        artifacts (django.db.models.Artifacts): list of artifacts in all repos being exported
+        last_export (django.db.models.PulpExport): previous export of owning Exporter
+
+    Raises:
+        ValidationError: When path is not in the ALLOWED_EXPORT_PATHS setting
+    """
+    # TODO: notice the existence of a last_export and not-full and do an incremental export
```
bmbouter, line 61: Generally for Pulp3 in pulpcore we've either filed TODOs as followup work, implemented them pre-merge, or chosen to not track them. I'm ok w/ any of those outcomes for the TODOs, but I do think we want to not merge this with any TODOs in the code.

pulpcore/app/importexport.py
```diff
@@ -0,0 +1,113 @@
...
+    _write_export(export.tarfile, resource, dest_dir)
+
+    resource = RepositoryVersionResource(repository_version)
+    _write_export(export.tarfile, resource, dest_dir)
+
+    # find and export any ModelResource found in pulp_<repo-type>.app.modelresource
+    plugin_name = repository_version.repository.pulp_type.split('.')[0]
+    cfg = get_plugin_config(plugin_name)
+
+    plugin_resource_modulename = cfg.name + '.modelresource'
```
bmbouter, line 105: We have this "plugin and discovery" part of the code that runs and analyzes the plugins at startup time [here](https://github.com/pulp/pulpcore/blob/master/pulpcore/app/apps.py#L79-L128) and then at runtime when that code is looked up, we access the data and references already created in memory. It would be great if we can continue that pattern here. To do that, I think move the discovery and indexing of these objects to the apps.py and access the info here instead.

pulpcore/app/modelresource.py
```diff
@@ -0,0 +1,76 @@
...
+    A ModelResource that has-a repository-version that can be used to limit its export, and a
+    queryset that is derived from that repository-version.
+
+    A plugin-writer will subclass the ModelResources for their Models from QueryModelResource,
+    and use it to define the limiting query (
+
+    Attributes:
+
+        repo_version (models.RepositoryVersion): The RepositoryVersion whose content we would like
+        to export
```
bmbouter, line 24: The second line I believe is usually indented 4 spaces which when it's rendered in Sphinx with autodoc I think matters.

CHANGES/6135.feature
ggainey: Changed the sentence to "Added support for exporting pulp-repo-versions.", since that's really what 6135 is all about. Sound good?

pulpcore/app/models/exporter.py
```diff
@@ -164,3 +148,52 @@ def export_repository_version(self, repository_version):
 
     class Meta:
         abstract = True
+
+
+class PulpExport(Export):
+    """
+    A class that provides export-files that can be imported into other Pulp instances.
```
bmbouter, line 155: s/class/model/

pulpcore/app/models/exporter.py
```diff
@@ -164,3 +148,52 @@ def export_repository_version(self, repository_version):
...
+
+        tarfile : a tarfile.Tarfile instance for this export to write into
+        sha256 (models.CharField): The SHA-256 checksum of the tarfile after export completes
+    """
+    tarfile = None
+    sha256 = models.CharField(max_length=64, null=True)
+
+    def export_tarfile_path(self):
+        """
+        Return the full tarfile name where the specified PulpExport should store its export
```
bmbouter, line 167: Can the normal Napolean style docstring format be used here to structure the  data. This could be interpreted by autodoc which expects that to render the things in a more structured way.

pulpcore/app/models/exporter.py
```diff
@@ -164,3 +148,52 @@ def export_repository_version(self, repository_version):
...
+                                               str(self.pulp_id),
+                                               datetime.utcnow().strftime("%Y%m%d_%H%M"))
+
+    class Meta:
+        default_related_name = '%(app_label)s_pulp_export'
+
+
+class PulpExporter(Exporter):
+    """
+    A class that controls creating exports that can be imported into other Pulp instances.
```
bmbouter, line 180: Similar here, I think calling these out as models is more useful than classes.

pulpcore/app/serializers/exporter.py
```diff
@@ -90,23 +97,33 @@ class Meta:
...
-    Serializer for pulp exports.
+    Serializer for pulp exporters.
     """
     path = serializers.CharField(
         help_text=_("File system location for the pulp export.")
     )
 
+    repositories = serializers.PrimaryKeyRelatedField(queryset=models.Repository.objects.all(),
+                                                      many=True)
+    # TODO: this needs to be an href
```
bmbouter, line 108: Does this need to change? I'm not sure. Wdyt?

pulpcore/app/tasks/export.py
```diff
@@ -44,3 +57,65 @@ def fs_repo_version_export(exporter_pk, repo_version_pk):
...
+
+    from pulpcore.app.serializers.exporter import ExporterSerializer
+    ExporterSerializer.validate_path(pulp_exporter.path, check_is_dir=True)
+
+    repositories = pulp_exporter.repositories.all()
+    export = PulpExport.objects.create(exporter=pulp_exporter, task=Task.current(), params=None)
+    tarfile_fp = export.export_tarfile_path()
+    os.makedirs(pulp_exporter.path, exist_ok=True)
+
+    # TODO: catch filesystem errors here
```
bmbouter, line 83: This one could be removed, a later a bugfiled from a user who expected more.

pulpcore/app/tasks/export.py
```diff
@@ -44,3 +57,65 @@ def fs_repo_version_export(exporter_pk, repo_version_pk):
...
+    with tarfile.open(tarfile_fp, 'w:gz') as tar:
+        export.tarfile = tar
+        CreatedResource.objects.create(content_object=export)
+
+        # 1) Spit out all Artifacts, ArtifactResource.json, and RepositoryResource.json
```
bmbouter, line 88: These could move to the docstring for accessability in autodoc also.

pulpcore/app/tasks/export.py
```diff
@@ -44,3 +57,65 @@ def fs_repo_version_export(exporter_pk, repo_version_pk):
...
+        # 3) Spit out all *resource JSONs per-repo-version
+
+        artifacts = []
+        repo_versions = []
+        # Gather up the versions and artifacts
+        for repo in repositories:
+            version = repo.latest_version()
+            # Check version-content to make sure we're not being asked to export an on_demand repo
+            content_artifacts = ContentArtifact.objects.filter(content__in=version.content)
+            if content_artifacts.filter(artifact=None).exists():
```
bmbouter, line 99: :+1:

pulpcore/app/viewsets/exporter.py
```diff
@@ -49,6 +60,12 @@ class ExporterViewSet(NamedModelViewSet,
     filterset_class = ExporterFilter
 
 
+class PulpExporterViewSet(ExporterViewSet):
```
bmbouter, line 63: Needs a docstring.

pulpcore/app/viewsets/exporter.py
```diff
@@ -61,13 +78,36 @@ class ExportViewSet(NamedModelViewSet,
...
+class PulpExportViewSet(ExportViewSet):
+    """
+    ViewSet for viewing exports from a PulpExporter.
+    """
+    parent_viewset = PulpExporterViewSet
+    serializer_class = PulpExportSerializer
+    queryset = PulpExport.objects.all()
+
+    @swagger_auto_schema(
+        # TODO: Why does this work for openapi and PulpExportSerializer fails with a 500?!?
```
bmbouter, line 96: This todo seems like it needs resolution before merging.

pulpcore/app/viewsets/exporter.py
```diff
@@ -61,13 +78,36 @@ class ExportViewSet(NamedModelViewSet,
...
+        request_body=PublicationExportSerializer,
+        operation_description="Trigger an asynchronous task to export a set of repositories",
+        responses={202: AsyncOperationResponseSerializer}
+    )
+    def create(self, request, exporter_pk):
+        """
+        Export a set of repository-versions.
```
bmbouter, line 103: This needs a full docstring I think.

pulpcore/app/importexport.py
ggainey: With this exception, the rest of the docstrings are single sentences - am I missing something?

bmbouter: You're not I put it there as a blanket statement. I read the PRs top to bottom

pulpcore/app/serializers/exporter.py
ggainey: When you GET an Exporter, current code shows its related Exports as an list-of-Export-UUIDs. This isn't very RESTful - you have to know the API for to ask for a specific Export to look at one in particular. If this were an array-of-hrefs, you could go directly to the export.

bmbouter: Oh I see. Thank you for explaining that. In that case then I think we want to make this change pre-merge. I don't know how hard that is to do, but that is what I recommend.

pulpcore/app/viewsets/exporter.py
ggainey: Then I am going to need help - there is no obvious error logged anywhere when this is 'PulpExportSerializer', the attempt to get the API simply fails with a 500. @daviddavis and I spent A LOT of time trying to understand what is happening there.

Given that import/export is proposed as a tech-preview for 3.3, I would be willing to go out with it this way.

pulpcore/app/models/exporter.py
ggainey: Concur, assumed the initial comment was %s/class/model/

pulpcore/app/viewsets/exporter.py
daviddavis: When I made [these changes](https://gist.github.com/daviddavis/40e1ffd1261b21ec398bfe633531d3a4) I did not see a 500 error. Are you still seeing one?

pulpcore/app/importexport.py
ggainey: We're aiming for tech-preview of this feature for 3.3. In that context, these are reminders not so much for 'what' has to happen before GA, so much as 'where'. As an example - we have a story for incremental-update, #6136. The TODOs that refer to incremental-update aren't "we need to implement incremental-update', they are '*when* you are implementing incremental update, these are places you need to think about'. If this were GA, I think it would def be worth a pass to clean up all TODOs via one of the above - but it doesn't feel as useful for tech-preview code.

Thoughts?

pulpcore/app/importexport.py
ggainey: Good thought, refactored

pulpcore/app/serializers/exporter.py
ggainey: This is another one I'm going to need a pointer or three on - I'm sure there is a Simple Piece of Django Magic that will make this happen, but I've spent *way* too much time looking for it....

pulpcore/app/viewsets/exporter.py
ggainey: @daviddavis yes. That was part of Friday's (and Sunday's) fun, alas.

daviddavis: Looks like the problem is that ExportedResources should be readonly and the yasg doesn't know how to generate docs for this field. This change worked for me:

https://gist.github.com/daviddavis/c60c8aa9e9203a96deb725d4f3ac819b

ggainey: And *that* did the trick! Thanks David, I would never have figured it out

pulpcore/app/serializers/exporter.py
ggainey: ExportRelatedField was The Magic - fixed, will be in next push

pulpcore/app/importexport.py
ggainey: The remaining TODOs were either 'implement incremental-export here' or 'possible performance improvement here?'. We have a story for incrementals, and performance-improvements should be made after measurements show us where the actual problems are! Removed TODOs.

pulpcore/app/models/exporter.py
```diff
@@ -164,3 +148,52 @@ def export_repository_version(self, repository_version):
...
+    sha256 = models.CharField(max_length=64, null=True)
+
+    def export_tarfile_path(self):
+        """
+        Return the full tarfile name where the specified PulpExport should store its export
+        """
+        # EXPORTER-PATH/export-EXPORTID-YYYYMMDD_HHMM.tar.gz
+        return "{}/export-{}-{}.tar.gz".format(self.exporter.path,
+                                               str(self.pulp_id),
+                                               datetime.utcnow().strftime("%Y%m%d_%H%M"))
```
daviddavis, line 172: I thought we agreed to remove the timestamp since it's not necessary.

CHANGES/6135.feature
```diff
@@ -0,0 +1,4 @@
+Added support for exporting pulp-repo-versions. Invoking the
+``/pulp/api/v3/exporters/core/pulp/`` API will create a PulpExporter. POSTing to that
+exporter via ``/pulp/api/v3/exporters/core/pulp/<exporter-uuid>/exports/`` will generate
+an export-tar.gz file at ``<exporter.path>/export-<export-UUID>-YYYYMMDD_hhMM.tar.gz``
```
daviddavis, line 4: Probably don't need the line about creating PulpExporters since that's called out in https://github.com/pulp/pulpcore/blob/master/CHANGES/6328.feature

pulpcore/app/apps.py
```diff
@@ -127,6 +135,22 @@ def import_urls(self):
...
+        """
+        if module_has_submodule(self.module, MODELRESOURCE_MODULE_NAME) and \
+                self.name != "pulpcore.app":
+            modelrsrc_module_name = '{name}.{module}'.format(
+                name=self.name, module=MODELRESOURCE_MODULE_NAME)
+            self.modelresource_module = import_module(modelrsrc_module_name)
+            self.exportable_classes = []
+            for (classname, cls) in inspect.getmembers(self.modelresource_module, inspect.isclass):
+                if cls.__module__ == self.modelresource_module.__name__:
+                    self.exportable_classes.append(cls)
```
daviddavis, line 152: This is great.

pulpcore/app/tasks/export.py
```diff
@@ -44,3 +57,65 @@ def fs_repo_version_export(exporter_pk, repo_version_pk):
...
+    2) Spit out one repositoryresource JSON w/all repositories
+    3) Spit out all *resource JSONs in per-repo-version directories
+    4) Compute and store the sha256 of the resulting tar.gz
+
+    Args:
+        pulp_exporter (models.PulpExporter): PulpExporter instance
+
+    Raises:
+        ValidationError: When path is not in the ALLOWED_EXPORT_PATHS setting,
+        OR path exists and is not a directory
```
daviddavis, line 77: nitpick - this should be indented.

daviddavis:
A couple small comments. Other than that, LGTM.

pulpcore/app/models/exporter.py
ggainey: Uh, I don't remember that. As an admin who's dealt with files like this, having the filename tell me when it was generated is/can be very useful when someone says "find me the export from last Thursday". I'd like to leave it as is.

CHANGES/6135.feature
ggainey: kk, will change.

pulpcore/app/tasks/export.py
ggainey: Ah, understood - also realized I misinterpreted a similar comment from bmbouter, fixing both.

pulpcore/app/models/exporter.py
daviddavis: The only thing I am thinking is that we're not tracking the file path + name. If we remove timestamp then `export_tarfile_path` always returns the file path of the export. Maybe one solution if we want to keep timestamp in the file name would be to store the file path in the database? Or maybe it's not worth keeping a history of the export file path + name?

pulpcore/app/apps.py
ggainey: Cool!

pulpcore/app/models/exporter.py
ggainey: Ooooh - oh, not tracking the actual filename is def a problem. Use case: user defines exporter, exports, updates exporter to change path, exports. Ugh.

OK - I think I can add 'export filename' to PulpExport today. It needs to be there, alongside the sha256.

ggainey: filename added to PulpExport. db-migration recreated. Good catch!

pulpcore/app/apps.py
```diff
@@ -127,6 +135,22 @@ def import_urls(self):
                 name=self.name, module=URLS_MODULE_NAME)
             self.urls_module = import_module(urls_module_name)
 
+    def import_modelresources(self):
```
bmbouter, line 138: :+1: thank you for moving this here

pulpcore/app/importexport.py
```diff
@@ -0,0 +1,102 @@
...
+
+    data = dataset.json.encode('utf8')
+    info = tarfile.TarInfo(name=dest_filename)
+    info.size = len(data)
+    the_tarfile.addfile(info, io.BytesIO(data))
+
+
+def export_artifacts(export, artifacts, last_export=None):
+    """
+    Export a set of Artifacts, their matching db-content, and the db-content of the owning
```
bmbouter, line 51: Can the first line not line-wrap to comply with PEP 257 please? We should really add a lint checker for PEP257.

pulpcore/app/modelresource.py
```diff
@@ -0,0 +1,76 @@
...
+)
+from pulpcore.app.models.repository import (
+    Repository,
+    RepositoryVersion,
+)
+
+
+class QueryModelResource(resources.ModelResource):
+    """
+    A ModelResource that has-a repository-version that can be used to limit its export, and a
```
bmbouter, line 15: Similar docstring single-top line request here.

pulpcore/app/tasks/export.py
```diff
@@ -44,3 +57,66 @@ def fs_repo_version_export(exporter_pk, repo_version_pk):
         ).format(exporter=exporter.name, repo_version=repo_version.pk, path=exporter.path)
     )
     exporter.export_repository_version(repo_version)
+
+
+def pulp_export(pulp_exporter):
+    """
+    Create a PulpExport instance and invoke it to export the repositories specified by the
```
bmbouter, line 64: single line docstring request here too.

pulpcore/app/importexport.py
ggainey: done

pulpcore/app/modelresource.py
ggainey: done

pulpcore/app/tasks/export.py
ggainey: done - noted and corrected some errors in the doc in the process, thanks for making me look again :)

pulpcore/app/apps.py
ggainey: looked like the right home for it!

PR #635, opened 2020-04-08
Teach pulp_export() to cast repositories to their Detail for plugin use

ggainey:

### DIFF
--- a/CHANGES/6466.bugfix
+++ b/CHANGES/6466.bugfix
@@ -1,0 +1,1 @@
+Taught pulp_export() to cast() repositories so plugins start with the right Detail type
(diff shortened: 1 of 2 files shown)


CHANGES/6466.bugfix
```diff
@@ -0,0 +1 @@
+Taught pulp_export() to cast() repositories so plugins start with the right Detail type
```
daviddavis, line 1: I think we should probably just drop this changelog since 6466 hasn't been released yet anyway.

daviddavis: Nevermind. I see that 6466 is a new issue.

ggainey: 6466 isn't 6135 - this is a new issue. And pulpbot complains at me if "fixes #1234" exists but there is not CHANGES for 1234

ggainey: overlapping comments ftw!

ggainey:
Closing this PR - this turns out to be Not The Problem

PR #636, opened 2020-04-08
Fix storage testing

mdellweg:

### DIFF
--- a/CHANGES/6450.bugfix
+++ b/CHANGES/6450.bugfix
@@ -1,0 +1,1 @@
+Changed storage checking in tests to use the correct value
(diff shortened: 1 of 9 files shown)


CHANGES/6450.bugfix
```diff
@@ -0,0 +1 @@
+Changed storage checking in tests to use the correct value
```
bmbouter, line 1: @daviddavis and @mdellweg I was thinking that since user's don't see test issues as bugs we should only use .misc entries for these types of changes next time. Having these that are already merged I think is ok since it's not a big deal. I wanted to comment to share my opinion for next time and to see what you all think.

daviddavis: That makes sense to me. Although I think we can still change this to a misc file. @mdellweg can you open a PR to do so?

daviddavis: Sorry, I should have caught that in my review.

mdellweg: Sorry i just thought ISSUE->bugfix.
https://github.com/pulp/pulpcore/pull/640

PR #639, opened 2020-04-09
Add index for pulp_type

dralley:
[noissue]


dralley:
After some minor benchmarking, it doesn't really seem like I was getting the benefits I was hoping for.  Maybe I'm testing the wrong things -- orphan cleanup and duplicate removal should see some benefit but I don't really have time to design a benchmark for those things.

The container plugin might benefit as well, but ditto there.

I'm not sure it's worth advertising without hard numbers.

bmbouter:
If the benefits aren't compelling I think we should do nothing. If you do think we should merge this I believe it needs a changelog entry and it needs to land in a Y-release because it ships an additional migration.

dralley:
Closing, will reconsider later on

PR #645, opened 2020-04-13
StableOrderingFilter only for NamedModelViewSet

fabricio-aguiar:
https://pulp.plan.io/issues/6347
fixes #6347

Problem:
builds failing for pulpcore, pulp_ansible, pulp_file

Observations:
- Noticed some SQL errors on logs;
- Commented `StableOrderingFilter` on settings, all tests passed;
- Focused om the following test:
```python 
pytest pulp_ansible/tests/functional/api/collection/v2/test_sync.py::SyncCollectionsFromPulpServerTestCase::test_sync_collections_from_pulp
```
- Noticed test was breaking on [GalaxyCollectionDetailView](https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/galaxy/views.py#L100), it was not entering the `get` method.
- As this view is `generics.RetrieveAPIView` , I can see some django magic is breaking due to `StableOrderingFilter` 
Still not figured out why, but when I added it just on `NamedModelViewSet` it fixed the tests

from the docs I infer that filter backends only works for `GenericAPIView` and its subclasses:
https://www.django-rest-framework.org/api-guide/filtering/#setting-filter-backends


bmbouter:
@fabricio-aguiar what does this need to pass in Travis?

fabricio-aguiar:
> @fabricio-aguiar what does this need to pass in Travis?

skip the intermittent test

daviddavis:
This code LGTM. One thing: we should probably need to update the changelog message for #6347 to say that it's only for NamedViewSets.

pulpcore/app/viewsets/base.py
```diff
@@ -56,6 +57,33 @@ def _allows_filters(self, path, method):
...
+    Reference: https://github.com/encode/django-rest-framework/issues/6886#issuecomment-547120480
+    """
+
+    def get_ordering(self, request, queryset, view):
+        """
+        Ordering is set by a comma delimited ?ordering=... query parameter.
+
+        The `ordering` query parameter can be overridden by setting
+        the `ordering_param` value on the OrderingFilter or by
+        specifying an `ORDERING_PARAM` value in the API settings.
```
daviddavis, line 73: nitpick - can you wrap this at 100 chars?

bmbouter:
I was thinking the same thing; LGTM, but the changelog needs to be updated.

CHANGES/6347.feature
```diff
@@ -1,2 +1,2 @@
-Indroduced `ordering` keyword, which orders the results by specified field.
+Indroduced `ordering` keyword on `NamedViewSets`, which orders the results by specified field.
```
bmbouter, line 1: Actually I think this won't be understandable by users.

pulpcore/app/viewsets/base.py
fabricio-aguiar: I don't know how to summarize it, suggestion?

daviddavis: This text is fine. It's just not wrapped at 100 chars.

fabricio-aguiar: Oh! I thought wrap == summarize, now I got it

PR #653, opened 2020-04-16
Added docs for the export-side of import/export

ggainey:
ref #6364

### DIFF
--- a/docs/plugins/plugin-writer/concepts/subclassing/import-export.rst
+++ b/docs/plugins/plugin-writer/concepts/subclassing/import-export.rst
@@ -1,0 +1,61 @@
+.. _subclassing_import-export:
+
+Pulp Import/Export
+==================
+
+The Pulp Import/Export process is based around the `Django Import/Export library <https://django-import-export.readthedocs.io/en/latest/>`_ .
+To be 'exportable/importable', your plugin must define a ``modelresource`` module at
+``<plugin>/app/modelresource.py``. The module must contain a ModelResource subclasses
+for each Model you want to expose, and it must define an ``IMPORT_ORDER`` ordered list
+for all such ModelResources.
+
+QueryModelResource
+~~~~~~~~~~~~~~~~~~
+
+If you don't need to do anything "special" to export your Model you can subclass
+``QueryModelResource``. This only requires you to provide the ``Meta.model`` class for the
+Model being export/imported, and to override the ``set_up_queryset(self)`` method to
+define a limiting filter based on the self.repo_version provided by ``QueryModelResource``.
+
+An example ``QueryModelResource`` subclasses, for import/exporting the ``Bar`` Model
+from ``pulp_foo``, would look like this::
+
+    class BarResource(QueryModelResource):
+        """
+        Resource for import/export of foo_bar entities
+        """
+
+        def set_up_queryset(self):
+            """
+            :return: Bars specific to a specified repo-version.
+            """
+            return Bar.objects.filter(pk__in=self.repo_version.content)
+
+        class Meta:
+            model = Bar
+
+
+modelresource.py
+~~~~~~~~~~~~~~~~
+
+A simple ``modelresource.py`` module might look like this::
+
+    from pulpcore.app.modelresource import QueryModelResource
+    from pulp_file.app.models import FileContent
+
+    class FileContentResource(QueryModelResource):
+        """
+        Resource for import/export of file_filecontent entities
+        """
+
+        def set_up_queryset(self):
+            """
+            :return: FileContents specific to a specified repo-version.
+            """
+            return FileContent.objects.filter(pk__in=self.repo_version.content)
+
+        class Meta:
+            model = FileContent
+
+
+    IMPORT_ORDER = [FileContentResource]
--- a/docs/workflows/import-export.rst
+++ b/docs/workflows/import-export.rst
@@ -1,11 +1,168 @@
-Import and Export
-=================
+Pulp Import and Export
+======================
 
 .. warning::
     Support for import and export is provided as a tech preview in Pulp 3. Functionality may not
     work or may be incomplete. Also, backwards compatibility when upgrading is not guaranteed.
 
+Overview
+^^^^^^^^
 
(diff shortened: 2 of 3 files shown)


docs/plugins/plugin-writer/concepts/subclassing/import-export.rst
```diff
@@ -0,0 +1,61 @@
+.. _subclassing_import-export:
+
+Import/Export
+=============
+
+The Pulp Import/Export process is based around the `Django Import/Export library <https://django-import-export.readthedocs.io/en/latest/>`_ .
+To be 'export/importable', your plugin must define a ``modelresource`` module at
```
daviddavis, line 7: maybe exportable to match importable.

docs/plugins/plugin-writer/concepts/subclassing/import-export.rst
```diff
@@ -0,0 +1,61 @@
...
+``<plugin>/app/modelresource.py`` The module must contain a `ModelResource` subclasses
```
daviddavis, line 8: Missing period.

docs/plugins/plugin-writer/concepts/subclassing/import-export.rst
```diff
@@ -0,0 +1,61 @@
+.. _subclassing_import-export:
+
+Import/Export
```
daviddavis, line 3: Probably needs to be "Pulp Import/Export" since there are other types of import/export stuff the plugin needs to worry about.

docs/workflows/import-export.rst
```diff
@@ -5,7 +5,166 @@ Import and Export
...
+2. That Exporter is requested to produce and execute an Export for the current
+:term:`repository versions<RepositoryVersion>` of the specified
+:term:`repositories<Repository>`
+
+3. The resulting ``.tar.gz`` Export is transferred to the appropriate Downstream.
+
+4. On the Downstream Pulp instance, an Importer is defined, that maps the incoming
+Upstream Repositories to matching Downstream Repositories
+
+5. That Importer is requested to produce an execute an Import, pointing to the provided
```
daviddavis, line 30: produce and execute?

docs/workflows/import-export.rst
```diff
@@ -5,7 +5,166 @@ Import and Export
...
+export file from the Upstream.
+
+In order to minimize space utilization, import/export operates on sets of
+:term:`repositories<Repository>`. This allows the Export operation to export shared
```
daviddavis, line 34: probably keep Repositories capitalized to be consistent with Repository.

docs/workflows/import-export.rst
```diff
@@ -5,7 +5,166 @@ Import and Export
...
+:term:`artifacts<Artifact>` only once per-export, rather than once for each
```
daviddavis, line 35: artifact can be capitalized too I think.

docs/workflows/import-export.rst
daviddavis: There are several instances of `repositories` btw.

docs/workflows/import-export.rst
```diff
@@ -5,7 +5,166 @@ Import and Export
...
+:term:`Repository` being exported.
+
+.. note::
+
+    Export will not operate on :term:`repository versions<RepositoryVersion>` that have
```
daviddavis, line 40: Repository Versions. There are probably other instances that should be calitalized.

docs/workflows/import-export.rst
```diff
@@ -5,7 +5,166 @@ Import and Export
...
+    the settings options ``ALLOWED_IMPORT_PATHS`` and ``ALLOWED_EXPORT_PATHS``.
+    These default to empty, if not explicitly set attempts to import or export will fail
+    with a validation error like
+
+        ``"Path '/tmp/exports/' is not an allowed export path"``
+
+Definitions
+^^^^^^^^^^^
+Upstream
+    Pulp instance whose repository-version(s) we want to export
```
daviddavis, line 57: Repository Version(s)

docs/workflows/import-export.rst
```diff
@@ -5,7 +5,166 @@ Import and Export
...
+Downstream
+    Pulp instance that will be importing those repository-version(s)
```
daviddavis, line 59: Repository Version(s)

docs/workflows/import-export.rst
```diff
@@ -5,7 +5,166 @@ Import and Export
...
+``ZOO_UUID`` and ``ISOFILE_UUID``
+
+Set up 'zoo' repository"::
+
+    # Create the repository
+    export ZOO_HREF=$(http POST http://localhost:24817/pulp/api/v3/repositories/rpm/rpm/ name=zoo | jq -r '.pulp_href')
+    #
+    # Find and extract its UUID
+    ZOO_UUID=${ZOO_HREF##/pulp/api/v3/repositories/rpm/rpm/}
+    ZOO_UUID=${ZOO_UUID%%/}
```
daviddavis, line 109: With https://github.com/pulp/pulpcore/pull/651, it's not longer necessary to extract the UUIDs.

docs/workflows/import-export.rst
```diff
@@ -5,7 +5,166 @@ Import and Export
...
+This export file can now be transferred to a Downstream Pulp instance, and imported.
+
+Updating an Exporter
+--------------------
+
+You can update an Exporter to modify a subset of its fields::
 
+    http PATCH http://localhost:24817${EXPORTER_HREF} path=/tmp/newpath
+    ls /tmp/newpath/
+    export-f00523e1-451d-4795-af67-d694a80b9eaa-20200416_2018.tar.gz
```
daviddavis, line 167: This makes it seem like a previous export file will be moved if you update the exporter. I would probably just remove this line and the one prior.

docs/workflows/import-export.rst
ggainey: Creating an exporter requires the UUIDs of the repositories, not the hrefs. Ultimately, we probably want it to require repository-names instead - but right now, you need the UUIDs in isolation to pass to the next step.

ggainey: Old code is old - retested with #651 and you are correct.

PR #655, opened 2020-04-17
Add NoArtifactContentUploadSerializer

lubosmj:
closes #6281

### DIFF
--- a/pulpcore/app/files.py
+++ b/pulpcore/app/files.py
@@ -33,6 +33,12 @@ class PulpTemporaryUploadedFile(TemporaryUploadedFile):
         instance = cls(name, '', file.size, '', '')
         instance.file = file
         data = file.read()
+
+        # calling the method read() moves the file's pointer to the end of the file object,
+        # thus, it is necessary to reset the file's pointer position back to 0 in case of
+        # calling the method read() again from another place
+        file.seek(0)
+
         for hasher in hashlib.algorithms_guaranteed:
             instance.hashers[hasher].update(data)
         return instance
--- a/pulpcore/app/tasks/base.py
+++ b/pulpcore/app/tasks/base.py
@@ -1,5 +1,28 @@
 from pulpcore.app.apps import get_plugin_config
-from pulpcore.app.models import CreatedResource
+from pulpcore.app.models import Artifact, CreatedResource
+from pulpcore.app.files import PulpTemporaryUploadedFile
+
+
+def general_create_from_temp_file(app_label, serializer_name, *args, **kwargs):
+    """
+    Create a model instance from contents stored in a temporary Artifact.
+
+    A caller should always pass the dictionary "data", as a keyword argument, containing the
+    href to the temporary Artifact. Otherwise, the function does nothing.
+
+    This function calls the function general_create() to create a model instance.
+    Data passed to that function already contains a serialized artifact converted
+    to PulpTemporaryUploadFile that will be deleted afterwards.
+    """
+    data = kwargs.pop("data", None)
+    if data and "artifact" in data:
+        named_model_view_set = get_plugin_config(app_label).viewsets_module.NamedModelViewSet
+        artifact = named_model_view_set.get_resource(data.pop("artifact"), Artifact)
+
+        data["file"] = PulpTemporaryUploadedFile.from_file(artifact.file)
+
+        general_create(app_label, serializer_name, data=data, *args, **kwargs)
+        artifact.delete()
 
 
 def general_create(app_label, serializer_name, *args, **kwargs):
--- a/pulpcore/plugin/viewsets/content.py
+++ b/pulpcore/plugin/viewsets/content.py
@@ -1,3 +1,5 @@
+from collections import namedtuple
+
 from drf_yasg.utils import swagger_auto_schema
 
 from django.db.utils import IntegrityError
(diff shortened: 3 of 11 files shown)


pulpcore/app/tasks/base.py
```diff
@@ -1,5 +1,30 @@
 from pulpcore.app.apps import get_plugin_config
-from pulpcore.app.models import CreatedResource
+from pulpcore.app.models import Artifact, CreatedResource
+from pulpcore.app.files import PulpTemporaryUploadedFile
+
+
+def general_create_from_temp_file(app_label, serializer_name, *args, **kwargs):
+    """
+    Create a model instance out of contents stored in a temporary Artifact.
```
mdellweg, line 8:
```suggestion
    Create a model instance from contents stored in a temporary Artifact.
```

pulpcore/app/tasks/base.py
```diff
@@ -1,5 +1,30 @@
...
+    """
+    data = kwargs.pop("data", None)
+    if data and "artifact" in data:
+        named_model_view_set = get_plugin_config(app_label).viewsets_module.NamedModelViewSet
+        artifact = named_model_view_set.get_resource(data.pop("artifact"), Artifact)
+
+        data["file"] = PulpTemporaryUploadedFile.from_file(artifact.file)
+        # PulpTemporaryUploadedFile calls file.read() when instantiating, thus, it is necessary
+        # to reset the file's object position back to 0
+        data["file"].seek(0)
```
mdellweg, line 22: Should the seek be moved there then?

pulpcore/app/tasks/base.py
```diff
@@ -1,5 +1,30 @@
...
+
+        general_create(app_label, serializer_name, data=data, *args, **kwargs)
+        artifact.delete()
+    else:
```
mdellweg, line 26: Is falling through here a supported usecase or should it be an error?

pulpcore/plugin/viewsets/content.py
```diff
@@ -28,17 +33,86 @@ def get_deferred_context(self, request):
...
+        """
+        Initialize content data which will be passed to a created task.
+
+        Args:
+            serializer: A serializer retrieved from request data containing already validated data
+            request (rest_framework.request.Request): the current HTTP request being handled
+        """
+        raise NotImplementedError("Subclasses must implement this method.")
+
+    def init_tasks_function_type(self):
```
mdellweg, line 73: Instead of calling a function here, you could use class constants here.
And i would guess defaulting to `general_create` sounds reasonable.

pulpcore/app/tasks/base.py
lubosmj: Where it should be moved? I do not understand.

pulpcore/app/tasks/base.py
lubosmj: Can you elaborate more on your question?

pulpcore/app/tasks/base.py
mdellweg: I was thinking about your comment saying basically it needs to be done before `file.read()`.
So if this call to `file.read()` is under our control and it requires the file pointer to be at position 0 the seek should happen immediately before that read.

pulpcore/app/tasks/base.py
mdellweg: Is there a case, where you would call this specialized function just to hand everything over to `general_create` instead of calling `general_create` directly?

pulpcore/app/tasks/base.py
lubosmj: Call to file.read() is under "our" control in PulpTemporaryUploadedFile:
https://github.com/pulp/pulpcore/blob/7f29f46e10c94b1fa0f9d7a8873f872db8fc9b7c/pulpcore/app/files.py#L35

After reading data, the file's object pointer remains at the end of the file (as usually in Python). To keep the default behavior of having the pointer set to 0 when the instance of an uploaded file is normally created, it is required to call the method seek() here or in PulpTemporaryUploadedFile itself. I decided to do it here instead of modifying the existing method from_file().

For instance, the following line would read an empty string instead of the uploaded content without calling seek():
https://github.com/pulp/pulp_rpm/blob/12b5738ecdcc90a6b0d3c902848c6443cd96c1eb/pulp_rpm/app/serializers.py#L552.

pulpcore/app/tasks/base.py
lubosmj: No, I think I should then remove the if-else statement. Thank you for pointing this out.

pulpcore/app/tasks/base.py
mdellweg: So your code here is repairing the assumptions for another piece of code that was violated by yet another.
Maybe calling `seek(0)` immediately before every call to read is the sane thing to do?

lubosmj: Hmm. I have got your point. But I think it would be better to just put the call seek(0) in PulpTemporaryUploadedFile.from_file() after read() to reset the pointer or leave it as it is now. I reckon it is not very plugin-writers-friendly to call seek(0) before every call of read(). And I do not know whether plugin writers already deal with this thing or not. What if they want to read the content byte by byte? I cannot even override the method read() due to that. I just wanted to not break the existing code.

mdellweg: I agree that resetting the file after read is the polite way to do it.
And it should i general be a rather inexpensive call.

lubosmj: I moved the call to PulpTemporaryUploadedFile.

pulpcore/app/files.py
```diff
@@ -33,6 +33,12 @@ def from_file(cls, file):
         instance = cls(name, '', file.size, '', '')
         instance.file = file
         data = file.read()
+
+        # calling the method read() moves the file's pointer to the end of the file object,
+        # thus, it is necessary to reset the file's pointer position back to 0 in case of
+        # calling the method read() again from another place
+        file.seek(0)
```
mdellweg, line 40: :1:

pulpcore/plugin/viewsets/content.py
lubosmj: I used a class variable to store function types instead of calling this function.

pulpcore/app/files.py
lubosmj: @daviddavis, what do you thing about this?

daviddavis: Looks good to me. :+1:

PR #659, opened 2020-04-21
Add warning for SigningServirce if signing script changes

Manisha15:
SigningService issues a warning if the signing script has changed on disk

fixes #6291
https://pulp.plan.io/issues/6291


pulpcore/app/migrations/0025_auto_20200421_1002.py
```diff
@@ -0,0 +1,18 @@
...
+        ('core', '0024_signingservice_sha_256'),
+    ]
+
+    operations = [
+        migrations.RenameField(
+            model_name='signingservice',
+            old_name='sha_256',
+            new_name='sha256',
+        ),
+    ]
```
mdellweg, line 18: Can you redo the migration as one?

pulpcore/app/models/content.py
```diff
@@ -412,6 +414,9 @@ def sign(self, filename):
             stderr=subprocess.PIPE,
         )
 
+        if self.sha256 != self.hash_value(self.script):
+            warnings.warn('Provided signing script does not match original signing script', Warning)
```
mdellweg, line 418: What is the purpose of the `Warning` argument here?

Manisha15: `Warning` is a preexisting class, if no argument is used then it is by default UserWarning. For eg,
```
>>> warnings.warn('Provided signing script does not match original signing script.')
<string>:1: UserWarning: Provided signing script does not match original signing script.

>>>warnings.warn('Provided signing script does not match original signing script.', Warning)
<string>:1: Warning: Provided signing script does not match original signing script.

File: pulpcore/app/migrations/0025_auto_20200421_1002.py (hunk 1 above)
Comment (Manisha15) [reply pulpcore/app/migrations/0025_auto_20200421_1002.py]: Done.

File: pulpcore/app/models/content.py (hunk 2 above)
Comment (quba42) [reply pulpcore/app/models/content.py:418]: Perhaps we should use `_logger.warning` here? (As is already the case in other places in pulpcore.)

File: pulpcore/app/models/content.py [hunk 3]
```diff
@@ -441,8 +446,19 @@ def save(self, *args, **kwargs):
         Save a signing service to the database (unless it fails to validate).
         """
         self.validate()
+        self.sha256 = self.hash_value(self.script)
```
quba42, line 449: I think this should be set before the call to `self.validate()`?
Won't validate call sign which will check the hash and be confused if it has not been set yet?

pulpcore/app/models/content.py
```diff
@@ -424,7 +429,7 @@ def sign(self, filename):
 
     def validate(self):
         """
-        Ensure that the external signing script produces the desired beahviour.
+        Ensure that the external signing script produces the desired behaviour.
```
quba42, line 432: oops! my bad.

pulpcore/app/models/content.py
Manisha15: Yes you're Right. Fixed!

quba42:
Since the tests have now turned green for the first time, could we move this towards final review? @mdellweg @dkliban 

As far as I can tell this is now the "minimal consensus" version of this change.
It adds a non intrusive warning for something that _could_ be an operational error.

There have been voices (@mibanescu) who would prefer a hard error in combination with some kind of update mechanism (?) for existing singing services. (I am personally sympathetic towards a hard error. :wink:) However, anything further could still be done in a followup task. Opinions?

mdellweg:
This is a very nice piece of work.
My biggest concern here is that the warning going into the system log is of very little value. That is why i am, too, leaning towards a hard error.
Can someone elaborate a little more on the need to change that script?
Is changing it not a good reason to revalidate it still works?

quba42:
Maybe the new behaviour could be: If there is a new script (changed hash), the validate function is called, and if that fails then there is a hard error. If it changed and validates then only a warning?

A hard error for a script that fails to validate is surely uncontroversial.

mdellweg:
If we verify the script only immediately before using it, we can just as well let it fail _while_ we use it. I don't see the benefit there.

bmbouter:
I'm late to review this, I'm sorry about that. I think there are possible legitimate use cases where the SigningService script could have during normal operation. Here are two examples:

* Say the hostname/ip of the signing server or the gpg key ring path legitimately changes?

* Also what about cases where the signing service script is on several servers in a Pulp cluster and they don't have exactly the same contents. For example perhaps the script contains a key itself to sign it's request to the signing server with and this is per-server? It's not something I would prefer but it's a possibility.

I'm also not sure this is that helpful in increasing the security of Pulp. If an attacker has already compromised the SigningServiceScript itself you probably also have the ability to read the credentials to connect to the database.

daviddavis:
Adding WIP label pending pulp-dev discussion.

bmbouter:
A while ago we had a meeting about changes needed for the signing service. It was kind of prompted by this PR because we knew we needed to unblock it. The recording is here: https://youtu.be/uecwUFJTWno Also there was some email discussion about next steps here: https://www.redhat.com/archives/pulp-dev/2020-June/msg00062.html and here: https://www.redhat.com/archives/pulp-dev/2020-June/msg00083.html

I think we need to do those before we can move forward with this PR. This PR has been open for so long I think we need to close it, but my hope is that we can work towards the next steps discussed on the call. One that is done I think we can resolve this original goal of this PR pretty easily as built-in validation is part of the plan. I'd like to help provide review and a guidance earlier and better than I did with this PR. My apologies for that.

bmbouter:
I opened [this questions ticket](https://github.com/tfranzel/drf-spectacular/issues/152) against `drf-spectacular` asking about the state of their use of semver or if we should continue version pinning.

PR #660, opened 2020-04-21
Single container

mdellweg:


mdellweg:
This is a draft, because it should demonstrate the changes in https://github.com/pulp/plugin_template/pull/210

PR #661, opened 2020-04-21
DRAFT exporter testcases

ggainey:
Required some tweaks to model/serializer code to handle deletes and binding-calls.

[noissue]


pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
```diff
@@ -0,0 +1,280 @@
...
+            repos.append(a_repo)
+            remotes.append(remote)
+        return repos, remotes
+
+    @classmethod
+    def setUpClass(cls):
+        """Create class-wide variables."""
+        import pydevd_pycharm
+        pydevd_pycharm.settrace('192.168.1.109', port=3014, stdoutToServer=True,
+                                stderrToServer=True)
```
daviddavis, line 79: :smiley_cat:

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+        Utility routine to delete an exporter.
+
+        Sets last_exporter to null to make it possible. Also removes the export-directory
+        and all its contents.
+        """
+        if path.exists(exporter.path):
+            shutil.rmtree(exporter.path)
+        # NOTE: you have to manually undo 'last-export' if you really really REALLY want to
+        #  delete an Exporter. This is...probably correct?
```
daviddavis, line 111: Hmm. I think deleting the exporter should just cascade delete the exports?

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+class PulpExportTestCase(BaseExporterCase):
+    """Test PulpExport CRDL methods (Update is not allowed)."""
+
+    def _gen_export(self, exporter):
+        """Create and read back an export for the specified PulpExporter."""
+        # TODO: at this point we can't create an export unless we do string-surgery on the
+        #  exporter-href because there's no way to get just-the-id
+        export_response = self.exports_api.create(exporter.pulp_href.split("/")[-2], {})
```
daviddavis, line 187: @dkliban, @ggainey and I talked about this and it seems unexpected that you have to pass in a uuid and not an href here. Any thoughts?

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
ggainey: argh

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
```diff
@@ -0,0 +1,280 @@
...
+        task = self.client.get(export_response.task)
+        resources = task["created_resources"]
+        self.assertEqual(1, len(resources))
+        export_href = resources[0]
+        export = self.exports_api.read(export_href)
+        self.assertIsNotNone(export)
+        return export
+
+    def test_export(self):
+        """Issue and evaluate a PulpExport (this is both C and R)."""
```
daviddavis, line 198: C and R?

daviddavis: Oh, create and read?

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+    Base functionality for Exporter and Export test classes
+
+    The export process isn't possible without repositories having been sync'd - arranging for
+    that to happen once per-class (instead of once-per-test) is the primary purpose of this parent
+    class.
+    """
+    @classmethod
+    def _setup_repositories(cls):
+        """Cretae and sync a number of repositories to be exported."""
```
daviddavis, line 57: Cretae

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
ggainey: Create and Read - but it def needs to be clearer

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
ggainey: We protect from deleting the last_export of a PulpExporter (see https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/exporter.py#L198) - iirc that kept PulpExporter from being deleted if it had ever been exported, unless/until you explicitly set last_export to null first. I will experiment some more tho, a lot happened between when I wrote that comment and now.

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
```diff
@@ -0,0 +1,280 @@
...
+
+            # make sure the exporter knows it's gone
+            exporter = self.exporter_api.read(exporter.pulp_href)
+            exports = self.exports_api.list(exporter.pulp_href.split("/")[-2]).results
+            self.assertEqual(2, len(exports))
+
+            # Now try to delete the last_export export and fail
+            try:
+                self._delete_export(last_export)
+            except Exception:
```
daviddavis, line 271: Is it possible to use `assertRaises` here? Something like:

```python
with self.assertRaises(ApiException) as ae:
    self._delete_export(last_export)
self.assertEqual(ae.exception.status_code, 400)
```

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
daviddavis: Ok. I do see value in preventing a user from accidentally deleting an exporter that has exported but unsetting a field to delete an object seems weird.

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
ggainey: that works. returned-error-code is actually a 500, because the failure is a django.db.models.deletion.ProtectedError : ("Cannot delete some instances of model 'PulpExport' because they are referenced through a protected foreign key: 'PulpExporter.last_export'"

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
ggainey: You are not wrong. There is probably a better way to do this, will leave this comment open and think about it overnight.

daviddavis: Ok sounds good. I'm fine with leaving it this way as long as it's clear to the user that the last_export has to be unset in order to delete.

daviddavis:
@ggainey you should be able to set ALLOWED_PATHS in the template config. See pulp_ansible [here](https://github.com/pulp/pulp_ansible/blob/master/template_config.yml#L29) as an example.

ggainey:
> @ggainey you should be able to set ALLOWED_PATHS in the template config. See pulp_ansible [here](https://github.com/pulp/pulp_ansible/blob/master/template_config.yml#L29) as an example.

Awesome - changes pushed, let's see what happens!

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
ggainey: Opened https://pulp.plan.io/issues/6555 to track flattening this - closing the conversation here.

pulpcore/app/serializers/exporter.py
```diff
@@ -90,6 +90,7 @@ class ExportSerializer(ModelSerializer):
 
     params = serializers.JSONField(
         help_text=_('Any additional parameters that were used to create the export.'),
+        required=False,
```
daviddavis, line 93: This probably should be `readonly=True` instead of `required=False`

pulpcore/app/serializers/exporter.py
```diff
@@ -133,6 +134,7 @@ class PulpExporterSerializer(ExporterSerializer):
         queryset=models.PulpExport.objects.all(),
         many=False,
         required=False,
+        allow_null=True,
```
daviddavis, line 137: This too.

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
ggainey: Opened https://pulp.plan.io/issues/6556 to track flattening this - please move conversation there.

pulpcore/app/serializers/exporter.py
ggainey: Moving the discussion to https://pulp.plan.io/issues/6544

PR #663, opened 2020-04-22
Activate Black formatting

mdellweg:
[noissue]


daviddavis:
This is great. I'm glad someone finally got around to this. A couple things. There is [an issue](https://pulp.plan.io/issues/5020) for this work. Also, this is a huge change and I think we should probably send out a notice to pulp-dev with a merge date (April 30?) to give developers a chance to merge any open PRs, etc.

mdellweg:
I figured right after a Release might be the right time. I'll take that issue then.

daviddavis:
Talked to @bmbouter and we decided to squash and merge. Thank you @mdellweg!

PR #664, opened 2020-04-22
Add child tasks for processing repo version imports

daviddavis:
fixes #6484

### DIFF
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -1,5 +1,6 @@
 import json
 import os
+import re
 import tempfile
 import tarfile
 from gettext import gettext as _
@@ -18,12 +19,14 @@ from pulpcore.app.models import (
     PulpImporter,
     Repository,
     Task,
+    TaskGroup,
 )
 from pulpcore.app.modelresource import (
     ArtifactResource,
     ContentArtifactResource,
     ContentResource,
 )
+from pulpcore.tasking.tasks import enqueue_with_reservation
 
 log = getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


ggainey:
Grabbed the PR and ran it through my test-harness. In general, everything is working and the child tasks are reporting as I'd expect, and the code looks reasonable to me.

The one suggestion I was going to make is "it would be nice if the parent task had a summary of the children so I don't have to go look at them individually" - which, as it turns out, is exactly what [6559](https://pulp.plan.io/issues/6559) is opened to address! So I'm good with this change.

pulpcore/app/tasks/importer.py
```diff
@@ -33,6 +35,68 @@
...
+def import_repository_version(destination_repo_pk, source_repo_pk, tar_path):
+    """
+    Import a repository version from a Pulp export.
+
+    Args:
+        destination_repo_pk (str): Primary key of Repository to import into.
+        source_repo_pk (str): Primary key of the Repository in the export.
+        tar_path (str): A path to export tar.
+    """
+    def repo_version_path(src_repo):
```
bmbouter, line 55: Can this become a utility function maybe `_repo_version_path` or have this object become a Class with two static methods on it?

pulpcore/app/tasks/importer.py
```diff
@@ -96,26 +149,9 @@ def repo_version_path(temp_dir, src_repo):
...
-                # Create the repo version
-                with dest_repo.new_version() as new_version:
-                    new_version.set_content(content)
-
-    return importer
+                enqueue_with_reservation(
+                    import_repository_version,
+                    [dest_repo],
+                    args=[dest_repo.pk, src_repo['pulp_id'], path],
+                    task_group=None,
```
bmbouter, line 156: I think setting a task_group is preferred for the use of this pattern.

daviddavis: Yea, I left this blank because I wasn't sure how to set this up and so I filed https://pulp.plan.io/issues/6559. Are there any docs or examples that I can read up on? I didn't see any.

dkliban: There are no docs. But the idea is that you create TaskGroup object, save it, and pass it in to all the tasks.

PR #667, opened 2020-04-24
Add view_name_pattern to DetailRelatedField

### LINKED ISSUE #6521: Replace f-strings with .format() for proper gettext support
**Version**
main branch

**Describe the bug**
Some translatable strings are currently written using f-strings inside `_()` calls. This is problematic because `gettext` might not parse f-strings correctly.

**Expected behavior**
Translatable strings with `gettext` should use `.format()`.

mdellweg:
fixes #6521
https://pulp.plan.io/issues/6521

### DIFF
--- a/CHANGES/6521.bugfix
+++ b/CHANGES/6521.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug where the wrong kind of error was being raised for href parameters of mismatched types.
--- a/CHANGES/plugin_api/6521.feature
+++ b/CHANGES/plugin_api/6521.feature
@@ -1,0 +1,1 @@
+Added view_name_pattern to DetailRelatedField and DetailIdentityField to properly identify wrong resource types.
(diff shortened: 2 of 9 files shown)


dkliban:
@mdellweg I think the CHANGELOG entry should go under the CHANGES/plugin_api directory since this is a plugin API change.

daviddavis:
It might even make sense to have two entries: one in CHANGES for the bugfix and one in CHANGES/plugin_api.

Other than that, this LGTM.

CHANGES/6521.bugfix
```diff
@@ -0,0 +1 @@
+Added view_name_pattern to serialzer fields to properly identify wrong resource types and prevent server errors.
```
daviddavis, line 1: nitpick but I think these changelog messages should be written for an end user and I'm not sure an end user will understand this.

PR #670, opened 2020-04-24
Rehome QueryModelResource to pulpcore.plugin.importexport

ggainey:
fixes #6514

### DIFF
--- a/docs/plugins/plugin-writer/concepts/subclassing/import-export.rst
+++ b/docs/plugins/plugin-writer/concepts/subclassing/import-export.rst
@@ -13,9 +13,10 @@ QueryModelResource
 ~~~~~~~~~~~~~~~~~~
 
 If you don't need to do anything "special" to export your Model you can subclass
-``QueryModelResource``. This only requires you to provide the ``Meta.model`` class for the
-Model being export/imported, and to override the ``set_up_queryset(self)`` method to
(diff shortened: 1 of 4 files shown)


CHANGES/6514.removal
```diff
@@ -0,0 +1 @@
+Rehomed QueryModelResource to pulpcore.plugin.importexport.
```
daviddavis, line 1: I think this should go in the plugin api changelog (`CHANGES/plugin_api`).

docs/plugins/plugin-writer/concepts/subclassing/import-export.rst
```diff
@@ -40,7 +40,7 @@ modelresource.py
 
 A simple ``modelresource.py`` module might look like this::
 
-    from pulpcore.app.modelresource import QueryModelResource
+    from pulpcore.plugin.importexport import QueryModelResource
```
daviddavis, line 43: Good catch :+1:

CHANGES/6514.removal
ggainey: concur, moved

docs/plugins/plugin-writer/concepts/subclassing/import-export.rst
ggainey: :)

PR #671, opened 2020-04-24
Add version.json to top level of export.tar.gz

ggainey:
closes #6572

### DIFF
--- a/pulpcore/app/importexport.py
+++ b/pulpcore/app/importexport.py
@@ -1,5 +1,6 @@
 import os
 import io
+import json
 import tarfile
 import tempfile
 
@@ -40,6 +41,25 @@ def _write_export(the_tarfile, resource, dest_dir=None):
     info = tarfile.TarInfo(name=dest_filename)
     info.size = len(data)
     the_tarfile.addfile(info, io.BytesIO(data))
+
+
+def export_versions(export, version_info):
+    """
+    Write a JSON list of plugins and their versions as 'versions.json' to export.tarfile
+
+    Output format is [{"component": "<pluginname>", "version": "<pluginversion>"},...]
+
+    Args:
+        export (django.db.models.PulpExport): export instance that's doing the export
+        version_info (set): set of (distribution-label,version) tuples for repos in this export
+    """
+    # build the version-list from the distributions for each component
+    versions = [{"component": label, "version": version} for (label, version) in version_info]
+
+    version_json = json.dumps(versions).encode("utf8")
+    info = tarfile.TarInfo(name="versions.json")
+    info.size = len(version_json)
+    export.tarfile.addfile(info, io.BytesIO(version_json))
 
 
 def export_artifacts(export, artifacts, last_export=None):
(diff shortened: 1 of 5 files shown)


pulpcore/app/importexport.py
```diff
@@ -44,6 +47,26 @@ def _write_export(the_tarfile, resource, dest_dir=None):
...
+    Output format is [{"component": "<pluginname>", "version": "<pluginversion>"},...]
+
+    Args:
+        export (django.db.models.PulpExport): export instance that's doing the export
+    """
+    components = ['pulpcore'] + INSTALLED_PULP_PLUGINS
+    versions = [{
+        'component': component,
+        'version': get_distribution(component).version
+    } for component in components]
```
daviddavis, line 63: This captures all plugins and not just the ones used to generate the export?

ggainey: Oooohhh - good point, and worth pursuing. OK, will mull over the right way to make that happen.

ggainey: New version does this - but the code required, to get from core_repository.pulp_type to distribution.version, is...ugly. Will re-evaluate this w/ @bmbouter Monday, when brains are less fried.

bmbouter: In looking at this I've learned the splitting code is not going to work in cases when the python package name and the python distribution name are different. I've written this up as a separate issue [here](https://pulp.plan.io/issues/6671).

So until pulpcore can improve the state for plugin writers here's an alternative that uses only 1 split to get the package name and rely on the assumption the distribution is the same (for now).

```
from django.apps import apps
from pkg_resources import get_distribution
app_label = ContentType.objects.get_for_model(FileRepository).app_label
app_config_module = apps.get_app_config(app_label).name
maybe_the_distribution_name = app_config_module.split('.')[0]
version = get_distribution(maybe_the_distribution_name).version   ## hope for the best
```

bmbouter: When [feature 6671](https://pulp.plan.io/issues/6671) is implemented, I think this would become:

```
from django.apps import apps
from pkg_resources import get_distribution
app_label = ContentType.objects.get_for_model(FileRepository).app_label
version = apps.get_app_config(app_label).version
```

ggainey: Looks much better. I implemented def get_version_from_model(in_model) in pulpcore.app.util, so we can perhaps take advantage of this in other places.

pulpcore/app/tasks/export.py
```diff
@@ -88,9 +89,21 @@ def pulp_export(pulp_exporter):
 
         artifacts = []
         repo_versions = []
+        vers_info = set()
+        vers_info.add(("pulpcore", get_distribution("pulpcore").version))
+
+        from pulpcore.app.util import get_version_from_model
```
bmbouter, line 95: Are these able to be moved to the imports section?

ggainey: Back in The Before Times, no. However, we've encapsulated/changed code enough that the "Applications not yet loaded" exception no longer happens - moved.

PR #677, opened 2020-04-28
Add task groups to the Plugin API

dralley:
Somehow this never got added to the plugin API even though I remember making the change, it must have gotten accidentally stashed or something.  There's already a changelog entry... we can send this out in a Y release.

re: #6414
https://pulp.plan.io/issues/6414

### DIFF
--- a/CHANGES/plugin_api/6603.feature
+++ b/CHANGES/plugin_api/6603.feature
@@ -1,0 +1,1 @@
+Add TaskGroup to the plugin API.
--- a/pulpcore/plugin/models/__init__.py
+++ b/pulpcore/plugin/models/__init__.py
@@ -29,4 +29,5 @@ from pulpcore.app.models import (  # noqa
     RepositoryVersionDistribution,
     SigningService,
     Task,
+    TaskGroup,
 )


daviddavis:
Actually, I wonder if we should file an issue + add a changelog entry for this.

daviddavis:
I thought this over and I think we need a new issue that we can add to a 3.3.1 milestone plus changelog entry that will show up in the 3.3.1 changelog.

dralley:
The changelog entry that already went out in 3.3.0 already mentions "plugin writers", so it arguably doesn't need a changelog entry. wdyt?

daviddavis:
If #6414 were unreleased (and at MODIFIED) then I think it would be ok. My understanding of our release process though is that we can't/shouldn't attach changes to released issues.

bmbouter:
> If #6414 were unreleased (and at MODIFIED) then I think it would be ok. My understanding of our release process though is that we can't/shouldn't attach changes to released issues.

This is also my understanding. An issue can't be fixed in two releases. Additionally it creates issues for any downstream during their cherry picking process.

PR #678, opened 2020-04-28
Add content handler to distribution

pieterlexis:
See https://pulp.plan.io/issues/6570

This implements a `content_handler` function in `BaseDistribution` to allow plugin writers to serve non-Artifact content from their plugin. It also adds a `content_handler_directory_listing` function in `BaseDistribution` that allows plugin writers to add these 'files' to the directory listings.

On how this can be used, see https://github.com/pulp/pulp_rpm/pull/1687

Please comment on the code (and where perhaps an `await` or `async` could be added). Once the code is "good enough", I'll add tests.

### DIFF
--- a/CHANGES/plugin_api/6570.feature
+++ b/CHANGES/plugin_api/6570.feature
@@ -1,0 +1,1 @@
+Added support for Distributions to provide non-Artifact content via a content_handler.
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -254,11 +254,11 @@ class Handler:
 
     async def list_directory(self, repo_version, publication, path):
         """
-        Generate HTML with directory listing of the path.
+        Generate a set with directory listing of the path.
 
         This method expects either a repo_version or a publication in addition to a path. This
-        method generates HTML directory list of a path inside the repository version or
-        publication.
+        method generates a set of strings representing the list of a path inside the repository
+        version or publication.
 
         Args:
             repo_version (:class:`~pulpcore.app.models.RepositoryVersion`): The repository version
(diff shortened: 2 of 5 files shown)


dkliban:
@pieterlexis This looks good to me. I don't think you need to add any async/awaits for these method definitions/calls. 

Update the commit message with `closes #6570` and add a CHANGES/plugin_api/ file.

https://docs.pulpproject.org/contributing/git.html#commit-message
https://docs.pulpproject.org/contributing/index.html#workflow

pulpcore/content/handler.py
```diff
@@ -335,7 +339,9 @@ async def _match_and_stream(self, path, request):
                     rel_path = index_path
                 except ObjectDoesNotExist:
                     dir_list = await self.list_directory(None, publication, rel_path)
-                    return HTTPOk(headers={"Content-Type": "text/html"}, body=dir_list)
+                    dir_list.update(distro.content_handler_list_directory(rel_path))
```
bmbouter, line 342: Would you mind updating the docstring to include a paragraph calling out this method's use of calling of `list_directory` and `content_handler_list_directory`? Having knowledge that `content_handler_list_directory` could overwrite content from `list_directory` is specifically the part that I'm interested in having documented (even though its a private method).

If this change is too vague, and we can handle that separately since the docstring is already silent on the listing of directories at all.

CHANGES/plugin_api/6570.feature
```diff
@@ -0,0 +1 @@
+Added support for Distributions to provide non-Artifact content via a content_handler.
```
bmbouter, line 1: This changelog is great. In addition to it, can you add a `CHANGES/plugin_api/6570.removal` to call out the backwards incompatible change in the `list_directory` to use set() and not HTML as the return type? The backwards incompatible change is fine, but I would like it to be more clear for any plugin writers who already implemented `list_directory` in their subclasses and may need to know when upgrading to 3.4.0+.

bmbouter:
@pieterlexis thank you for this contribution! :+1: :medal_sports:

PR #682, opened 2020-05-04
Add 'all_tasks_dispatched' to TaskGroup

dralley:
closes: #6591
https://pulp.plan.io/issues/6591


pulpcore/app/migrations/0030_taskgroup_all_tasks_dispatched.py
```diff
@@ -0,0 +1,24 @@
...
+
+    dependencies = [
+        ('core', '0029_export_delete'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='taskgroup',
+            name='all_tasks_dispatched',
+            field=models.BooleanField(default=False),
```
dralley, line 21: Since the default is "false" we need to go back through and change all of the existing ones to True.

daviddavis:
The code looks great. I think since we expose this field to end users, we should probably have a changelog entry for end users too.

daviddavis:
LGTM. Thanks @dralley! @bmbouter or @dkliban do you want to give this a quick review before we merge?

CHANGES/6591.feature
```diff
@@ -0,0 +1 @@
+Task groups now have an 'all_tasks_dispatched' field which denotes that no more tasks will spawn as part of this group.
```
bmbouter, line 1: nitpick: can this have a newline char at the end? I think it will make the changelog look inconsistent otherwise.

CHANGES/plugin_api/6591.feature
```diff
@@ -0,0 +1 @@
+TaskGroups now have an 'all_tasks_dispatched' field that can be used to notify systems that no further tasks will be dispatched for a TaskGroup. Plugin writers should call ".finish()" on all TaskGroups created once they are done using them to set this field.
```
bmbouter, line 1: nitpick: can this have a newline char at the end I think it will make the changleog look inconsistent otherwise.

bmbouter: Can this be wrapped to 100 chars please?

docs/plugins/plugin-writer/concepts/subclassing/viewsets.rst
```diff
@@ -92,7 +94,11 @@ this is accomplished.
             # Since tasks are asynchronous, we return a 202
             return OperationPostponedResponse(result, request)
 
-See :class:`~pulpcore.plugin.tasking.enqueue_with_reservation` for more details.
+If a "task_group" is provided as an optional keyword argument, then the deployed task will be
```
bmbouter, line 97: Can we make this section linkable as `task_groups` and link directly to it?

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -69,6 +69,14 @@ Tasks are deployed from Views or Viewsets, please see :ref:`kick-off-tasks`.
    tasks/export
 
 
+**Task Groups**
```
bmbouter, line 72: Also can a new issue be filed so this can have a .doc changelog entry? We track docs changes for plugin writers so I think we need to do that.

docs/plugins/plugin-writer/concepts/subclassing/viewsets.rst
dralley: I'm not sure what you mean exactly

bmbouter: I mean can above this paragraph something like this be added? I'm not 100% if RST allows it is why I'm a bit unsure.

```
.. _dispatching-task-groups:
```

or maybe


```
.. _task-groups:
```

dralley: well, I formatted it this way because that is the name of the kwarg for ```enqueue_with_reservation```.  I could put the link somewhere near here though I guess?

bmbouter: yeah something linkable would be good. This is an optional thing, the main thing is the changelog for me.

CHANGES/plugin_api/6591.feature
bmbouter, line 1: Can this be line wrapped like all pulp docs are (to 100 chars).

pulpcore/app/models/task.py
```diff
@@ -422,6 +422,15 @@ def release_resources(self):
 
 class TaskGroup(BaseModel):
     description = models.TextField()
+    all_tasks_dispatched = models.BooleanField(default=False)
+
+    def finish(self):
+        """
+        Set 'all_tasks_dispatched' to True so that API users can know that there are no
```
bmbouter, line 429: Can this docstring get a single line summary as its top line to be PEP 257 compliant

PR #683, opened 2020-05-04
Remove coreapi dependency

mdellweg:
fixes #6583
https://pulp.plan.io/issues/6583

### DIFF
--- a/CHANGES/6583.misc
+++ b/CHANGES/6583.misc
@@ -1,0 +1,1 @@
+Removed dependency on coreapi.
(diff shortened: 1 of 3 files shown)


evgeni:
Well, that was easy 😅

cc @ekohl

mdellweg:
It was worth a try to just remove the dependency.

ekohl:
Nice! What was it providing?

mdellweg:
I think, it was the predecessor of openapi. And it wasn't even used anymore in pulp.

PR #684, opened 2020-05-04
Fix response when trying to cancel a completed task

mdellweg:
fixes #6465
https://pulp.plan.io/issues/6465

### DIFF
--- a/pulpcore/app/viewsets/task.py
+++ b/pulpcore/app/viewsets/task.py
@@ -23,7 +23,7 @@ from pulpcore.app.viewsets.custom_filters import (
     ReservedResourcesFilter,
     CreatedResourcesFilter,
 )
-from pulpcore.constants import TASK_INCOMPLETE_STATES
+from pulpcore.constants import TASK_INCOMPLETE_STATES, TASK_STATES
 from pulpcore.tasking.util import cancel as cancel_task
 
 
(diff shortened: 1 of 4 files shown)


pulpcore/app/viewsets/task.py
```diff
@@ -79,8 +79,10 @@ def partial_update(self, request, pk=None, partial=True):
         if request.data["state"] != "canceled":
             raise ValidationError(_("The only acceptable value for 'state' is 'canceled'."))
         task = cancel_task(task.pk)
+        # Check whether task is actually canceled
+        http_status = None if task.state == TASK_STATES.CANCELED else status.HTTP_409_CONFLICT
         serializer = self.serializer_class(task, context={"request": request})
-        return Response(serializer.data)
+        return Response(serializer.data, status=http_status)
```
dralley, line 85: So if ```http_status=None```, it defaults to 200?

dralley: @bmbouter What should we do if the user tries to cancel a task that is already in canceled state?

bmbouter: I expect pulp not to error. The user can specify cancel to the same task over and over without error.

dralley: Cool, then this is the correct behavior.

PR #687, opened 2020-05-06
Validate and serialize parameters used to invoke Export

ggainey:
This only implements the infrastructure to validate incoming parameters
and store them in the 'params' field, it does not implement the
functions (full/incremental, dry-run, export-by-version) being requested.

Also cleaned up some black-related format issues.

closes #6544


pulpcore/app/viewsets/exporter.py
```diff
@@ -116,8 +117,13 @@ def create(self, request, exporter_pk):
         Generates a Task to export the set of repositories assigned to a specific PulpExporter.
         """
         exporter = PulpExporter.objects.get(pk=exporter_pk).cast()
+        serializer = PulpExportSerializer(data=request.data)
+        serializer.exporter = exporter
```
ggainey, line 121: At serialize-validation-time I need the PulpExporter that is going to own the PulpExport, in order to get its repositories so I can validate explicit-versions if they are passed in. I'm not sure this is the 'right' way to accomplish this. The problem is, there is no PulpExport (which 'knows' it's Exporter) until after we have already passed validation. Thoughts?

pulpcore/app/serializers/exporter.py
```diff
@@ -93,6 +97,8 @@ class ExportSerializer(ModelSerializer):
     params = serializers.JSONField(
         help_text=_("Any additional parameters that were used to create the export."),
         required=False,
+        read_only=True,
+        allow_null=True,
```
daviddavis, line 101: If this field is read_only, do allow_null and required need to be set?

pulpcore/app/serializers/exporter.py
```diff
@@ -105,21 +111,76 @@ class PulpExportSerializer(ExportSerializer):
...
+        num_repos = self.exporter.repositories.count()
+        if num_repos != len(versions):
+            raise serializers.ValidationError(
+                _(
+                    "Number of versions does not match the number of Repositories for the owning "
+                    + "Exporter!"
+                ).format(num_repos, len(versions))
+            )
+
+        # make sure the specified versions 'belong to' the exporter.reos
```
daviddavis, line 163: Her name is `reos` and she dances on the sand :woman_dancing:

pulpcore/app/tasks/export.py
```diff
@@ -64,7 +63,7 @@ def fs_repo_version_export(exporter_pk, repo_version_pk):
     exporter.export_repository_version(repo_version)
 
 
-def pulp_export(pulp_exporter):
+def pulp_export(pulp_exporter, the_export):
```
daviddavis, line 66: Is `pulp_exporter` needed anymore since you can fetch it from `the_export`?

pulpcore/app/serializers/exporter.py
ggainey: As is Traditional Among Our People :) - will fix

pulpcore/app/tasks/export.py
ggainey: Ahhh, good point! concur

pulpcore/app/serializers/exporter.py
ggainey: Mmmm, yeah good point - there are a few places I need to clean this up

pulpcore/app/viewsets/exporter.py
ggainey: See http://www.tomchristie.com/rest-framework-2-docs/api-guide/serializers#including-extra-context for the right way to add extra info for a serializer to use. Code updated.

pulpcore/app/serializers/exporter.py
daviddavis, line 101: You marked this as resolved?

ggainey: Ugh, I did and fixed those - I may have rebased after, let me recheck

ggainey: Bah, no not a rebase-issue, a 'checked all the *other* places and failed to fix the one you actually noticed' issue. Ay de me. Fixing.

PR #695, opened 2020-05-08
Return username and password when requesting Remote

fao89:
https://pulp.plan.io/issues/6346
closes #6346

### DIFF
--- a/CHANGES/6346.bugfix
+++ b/CHANGES/6346.bugfix
@@ -1,0 +1,2 @@
+Remote fields `username` and `password` show up in:
+REST docs, API responses, and are available in the bindings.
--- a/CHANGES/plugin_api/6346.removal
+++ b/CHANGES/plugin_api/6346.removal
@@ -1,0 +1,1 @@
+Fields: `username` and `password` will be returned to the rest API user requesting a `Remote`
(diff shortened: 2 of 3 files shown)


CHANGES/6346.bugfix
```diff
@@ -0,0 +1 @@
+Fixed: `Remote` fields `username` and `password` not showing up in REST docs
```
bmbouter, line 1: I believe they were also not showing up in the openAPI schema definition

bmbouter: I'm trying to clarify it's not only affecting docs but also bindings. Also for changelogs I think we want to describe the solution to the problem not the problem itself. Put together, here's what that looks like to me:

Remote fields `username` and `password` show up in REST docs, API responses, and are available in the bindings.

How does this sound?

fao89: sounds good

bmbouter:
With two acks and some IRC feedback I'm going to merge this. Thanks @fao89 !

snecklifter:
@bmbouter @fao89 thanks for this, when can we expect to see this in a stable release?

bmbouter:
It will release with pulpcore 3.4.0. Yesterday on IRC I proposed tentative dates of May 19th or 26th. I'll bring up the 3.4.0 release timeline today at open floor today.

PR #696, opened 2020-05-11
Teach PulpExport to export specific repo-versions when invoked with versions:= .

ggainey:
closes #6566

### DIFF
--- a/pulpcore/app/tasks/export.py
+++ b/pulpcore/app/tasks/export.py
@@ -63,9 +63,44 @@ def fs_repo_version_export(exporter_pk, repo_version_pk):
     exporter.export_repository_version(repo_version)
 
 
+def _get_versions_to_export(the_exporter, the_export):
+    """
+    Return repo-versions to be exported.
+
+    versions is based on exporter-repositories and how the export-cmd was
+    invoked.
+    """
+    repositories = the_exporter.repositories.all()
+    # Figure out which RepositoryVersions we're going to be exporting
+    # Use repo.latest unless versions= was specified
+    if the_export.validated_versions is not None:
+        versions = the_export.validated_versions
+    else:
+        versions = [r.latest_version() for r in repositories]
+    return versions
+
+
+def _get_versions_info(the_exporter):
+    """
+    Return plugin-version-info based on plugins are responsible for exporter-repositories.
+    """
+    repositories = the_exporter.repositories.all()
+
+    # extract plugin-version-info based on the repositories we're exporting from
+    vers_info = set()
+    # We always need to know what version of pulpcore was in place
+    vers_info.add(("pulpcore", get_distribution("pulpcore").version))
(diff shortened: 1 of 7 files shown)


pulpcore/tasking/tasks.py
```diff
@@ -185,7 +185,8 @@ def enqueue_with_reservation(
         args (tuple): The positional arguments to pass on to the task.
         kwargs (dict): The keyword arguments to pass on to the task.
         options (dict): The options to be passed on to the task.
-        task_group (models.TaskGroup): A TaskGroup to add the created Task to.
+        task_group (pulpcore.app.models.TaskGroupmodels.TaskGroup): A TaskGroup to add the created
```
daviddavis, line 188: I think the module path is messed up.

pulpcore/app/viewsets/exporter.py
```diff
@@ -116,12 +116,21 @@ def create(self, request, exporter_pk):
...
         serializer = PulpExportSerializer(data=request.data, context={"exporter": exporter})
         serializer.is_valid(raise_exception=True)
 
+        # Invoke the export
         export = PulpExport.objects.create(exporter=exporter, params=request.data)
-
+        export.validated_versions = (
+            serializer.validated_data["versions"]
+            if "versions" in serializer.validated_data
+            else None
```
daviddavis, line 132: this might be a personal preference but multi-line ternaries are always hard for me to decipher.

pulpcore/app/tasks/export.py
```diff
@@ -63,9 +63,27 @@ def fs_repo_version_export(exporter_pk, repo_version_pk):
...
+    if the_export.validated_versions is not None:
+        versions = the_export.validated_versions
+    else:
+        versions = [r.latest_version() for r in repositories]
+
+    # extract plugin-version-info based on the repositories we're exporting from
+    vers_info = set()
+    vers_info.add(("pulpcore", get_distribution("pulpcore").version))
+    for r in repositories:
+        vers_info.add(get_version_from_model(r.cast()))
```
daviddavis, line 79: I would probably split this into a separate function. It took me a minute or two of head scratching to figure out what was going on.

daviddavis:
Looks good overall. A couple suggestions and my less than favorable opinion of ternaries in Python.

pulpcore/app/tasks/export.py
ggainey: This loop is the whole point of _get_versions_info - making it into yet another function doesn't feel like it's helping anything. I will add commentary around the loop explaining, tho

ggainey: I take it back, I can see how this would be useful - changing

pulpcore/app/viewsets/exporter.py
ggainey: Should be using .get(key, default) in this context anyway - fixing

PR #697, opened 2020-05-12
Make ssl certs and keys accessible through RESTAPI

mdellweg:
fixes #6691
https://pulp.plan.io/issues/6691

### DIFF
--- a/CHANGES/6691.removal
+++ b/CHANGES/6691.removal
@@ -1,0 +1,1 @@
+Content of ssl certificates and keys changed to be return their full value instead of sha256 through REST API.
--- a/pulpcore/app/serializers/repository.py
+++ b/pulpcore/app/serializers/repository.py
@@ -11,7 +11,6 @@ from pulpcore.app.serializers import (
     DetailRelatedField,
     LatestVersionField,
     ModelSerializer,
-    SecretCharField,
     RepositoryVersionIdentityField,
     RepositoryVersionRelatedField,
     RepositoryVersionsIdentityFromRepositoryField,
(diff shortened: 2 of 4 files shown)


CHANGES/6691.removal
```diff
@@ -0,0 +1 @@
+Removed SecretCharField from the codebase.
```
daviddavis, line 1: This probably needs to go in CHANGES/plugin_api.

daviddavis: Oh, we never exposed it via the plugin api. Ignore :smiley_cat:

mdellweg: I did not see it exposed in the plugin api.

daviddavis: I think it might be better to say here that fields that returned sha256 (ca_cert, client_cert, client_key) now return the full value.

pulpcore/app/serializers/repository.py
```diff
@@ -52,26 +51,23 @@ class RemoteSerializer(ModelSerializer):
         validators=[UniqueValidator(queryset=models.Remote.objects.all())],
     )
     url = serializers.CharField(help_text="The URL of an external content source.",)
-    ca_cert = SecretCharField(
+    ca_cert = serializers.CharField(
         help_text="A string containing the PEM encoded CA certificate used to validate the server "
         "certificate presented by the remote server. All new line characters must be "
         "escaped. Returns SHA256 checksum of the certificate file on GET.",
```
daviddavis, line 57: Also this help text needs to be updated.

CHANGES/6691.bugfix
```diff
@@ -0,0 +1 @@
+Content of ssl certificates and keys changed to be accessible through REST API.
```
mdellweg, line 1:
```suggestion
Content of ssl certificates and keys changed to be return their full value instead of sha256 through REST API.
```
@daviddavis like this?

pulpcore/app/serializers/repository.py
mdellweg:
```suggestion
        "escaped.",
```

pulpcore/app/serializers/repository.py
```diff
@@ -52,26 +51,23 @@ class RemoteSerializer(ModelSerializer):
...
-        write_only=False,
         required=False,
         allow_null=True,
     )
-    client_cert = SecretCharField(
+    client_cert = serializers.CharField(
         help_text="A string containing the PEM encoded client certificate used for authentication. "
         "All new line characters must be escaped. Returns SHA256 checksum of the "
         "certificate file on GET.",
```
mdellweg, line 64:
```suggestion
        "All new line characters must be escaped.",
```

pulpcore/app/serializers/repository.py
```diff
@@ -52,26 +51,23 @@ class RemoteSerializer(ModelSerializer):
...
-        write_only=False,
         required=False,
         allow_null=True,
     )
-    client_key = SecretCharField(
+    client_key = serializers.CharField(
         help_text="A PEM encoded private key used for authentication. Returns SHA256 checksum of "
         "the certificate file on GET.",
```
mdellweg, line 70:
```suggestion
        help_text="A PEM encoded private key used for authentication.",
```

bmbouter:
@mdellweg The code all looks good, my only suggestions are around the changelogs.

1. Remove the CHANGES/6691.removal file
2. Rename CHANGES/6691.bugfix -> CHANGES/6691.removal so it shows up in the backwards compatible area

PR #707, opened 2020-05-19
Splitting seralizers into read and write ones

fao89:
https://pulp.plan.io/issues/6775
closes #6775

### DIFF
--- a/pulpcore/app/openapigenerator.py
+++ b/pulpcore/app/openapigenerator.py
@@ -7,7 +7,7 @@ from drf_yasg import openapi
 from drf_yasg.generators import OpenAPISchemaGenerator
 from drf_yasg.inspectors import SwaggerAutoSchema
 from drf_yasg.openapi import Parameter
-from drf_yasg.utils import filter_none, force_real_str
+from drf_yasg.utils import filter_none, force_real_str, get_serializer_ref_name, no_body
 from rest_framework import serializers
 
 from pulpcore.app.models import Export, Import, RepositoryVersion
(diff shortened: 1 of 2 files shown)


pulpcore/app/openapigenerator.py
```diff
@@ -459,18 +459,96 @@ def get_tags(self, operation_keys):
...
+
+        read_fields = OrderedDict()
+        write_fields = OrderedDict()
+        for field_name, field in serializer.fields.items():
+            if not field.write_only:  # Removing write_only fields
+                read_fields[field_name] = field
+            if not field.read_only:  # Removing read_only fields
+                write_fields[field_name] = field
+
+        if len(read_fields) == len(write_fields):
```
mdellweg, line 505: I think that comparison is too risky. Can you compare the `keys()` for equality?

pulpcore/app/openapigenerator.py
```diff
@@ -459,18 +459,96 @@ def get_tags(self, operation_keys):
...
+            return serializer
+
+        contain_write_fields = len(serializer.fields) > len(read_fields)
+        if self.method.lower() == "get" or not contain_write_fields:
+            serializer.fields = read_fields
+            return serializer
```
mdellweg, line 511:
```suggestion
        contain_write_only_fields = len(serializer.fields) > len(read_fields)
        if self.method.lower() == "get" or not contain_write_only_fields:
            serializer.fields = read_fields
            return serializer
```
Also: are `serializer.fields` and `read_fields` not exactly the same in this case?

pulpcore/app/openapigenerator.py
```diff
@@ -459,18 +459,96 @@ def get_tags(self, operation_keys):
...
+
+        return self._convert_serializer(serializer, WriteOnly)
```
mdellweg, line 513: I belive we agreed to perform this transformation on the read-serializer and not the write-serializer.

dkliban: That's right ... we want the write serializer to keep the original name and the read serializer to change to OriginalNameRead. 

The motivation for this outlined here: https://www.redhat.com/archives/pulp-dev/2020-April/msg00101.html

pulpcore/app/serializers/content.py
```diff
@@ -33,7 +33,6 @@ class SingleArtifactContentSerializer(BaseContentSerializer):
     relative_path = serializers.CharField(
         help_text=_("Path where the artifact is located relative to distributions base_path"),
         validators=[fields.relative_path_validator],
-        write_only=True,
```
mdellweg, line 36: That use of `write_only ` should be preserved. As the database for content does not have that field.

fao89: Many plugins are counting with a response with `relative_path`, keeping it as `write_only` would break many plugins :
https://travis-ci.com/github/pulp/pulpcore/jobs/338483661#L1866
https://github.com/pulp/pulp_deb/blob/7c90e1dea8583c9bb838ce77014243adaacceee8/pulp_deb/tests/functional/api/test_crud_content_unit.py#L77

mdellweg: That's interesting, because in those cases, the `write_only`ness should be deactivated dynamically in the `__init__` of the serializer. Are you assigning the model attribute on the ReadOnly serializer you are creating on the fly?

fao89: I'll put a `rpdb` to see why it is ignoring the `__init__`

fao89: it didn't stop at my `rpdb`, so drf_yasg has other methods for delivering the responses, I'll investigate more

fao89: with `ReadOnly` mixin I was getting the fields from the original serializer class, so the `__init__` was not triggered. So instead of creating a new serializer class from the mixin, I'm doing it by modifying the serializer object.

pulpcore/app/openapigenerator.py
```diff
@@ -459,18 +459,48 @@ def get_tags(self, operation_keys):
...
+        serializer_meta = getattr(original_serializer, "Meta", BlankMeta)
+        read_ref_name = f"{get_serializer_ref_name(original_serializer)}Read"
+        read_fields = OrderedDict()
+        for field_name, field in original_serializer.fields.items():
+            if not field.write_only:
+                read_fields[field_name] = field
+
+        class ReadOnlySerializer(original_serializer.__class__):
+            class Meta(serializer_meta):
+                ref_name = read_ref_name
```
mdellweg, line 492: So this new class is there to change the ref_name. Anyway we can get away without crating that transient class by just setting that `ref_name` on the instance `original_serializer` (like we do with the fields two lines down)?

fao89: I tried, but it changes the original serialize, and I ended up with SerializerNameReadReadRead

mdellweg: Then just ignore this one.

dkliban:
I tested this by creating a PythonPythonPackageContent. Worked great!

PR #712, opened 2020-05-20
Fix issues where tasks records remain in the 'waiting' state forever.

dralley:


dralley:
Still just a partial fix.

bmbouter:
I'm wondering if the fix would involve some sort of change in [this area](https://github.com/pulp/pulpcore/blob/master/pulpcore/tasking/worker.py#L63) where the queue is created.

dralley:
I think it might be a little racy.  If I step through with a debugger, it works...

bmbouter:
What about reproducing it like this?
1. Add a long sleep statement to a file sync, like 300 seconds.
2. Create N+1 repositories to sync into where N is the number of pulp workers. So 2 workers by default, create 3 repos.
3. Initiate a sync of the N+1 repos. I expect N to start and 1 to be 'waiting'
4. `sudo pkill -9 -f redis; sudo systemctl start redis`
5. I think the 1 'waiting' task will be stuck.

dralley:
I might know what it is.  The task is spawned onto a queue with the name of the worker they are assigned to.  This worker is determined via ```_acquire_worker``` with a database query.  When redis goes down, all the workers get killed off, but their records remain in the database for 30 seconds until they get cleaned up in the background.

If it happens quickly enough, it could try to assign the task to a dead worker, and dump the task into a queue for the dead worker.  That should be easy enough to verify...

dralley:
Confirmed that ^^ is the problem.  I halted execution just after the worker was assigned and waited about 15 seconds, and refreshed the record.  It went from ```online=True``` && ```missing=False``` to the reverse.

So we have a window of some not-insignificant timeframe where tasks are lost by being assigned to dead workers -- and their reserved resources block subsequent tasks from progressing as well.

bmbouter:
Pulp's worker design currently is supposed to handle this. It's ok for the resource manager to route work to dead workers. It's not ideal but that's a separate issue. So to me the problem is that cleanup mechanism isn't working.

That code lives here: https://github.com/pulp/pulpcore/blob/master/pulpcore/tasking/services/worker_watcher.py#L53-L170

dralley:
@bmbouter A little more detail than I mentioned the other day:

W/ the current state of my branch I'm only able to reproduce this about 20% of the time instead of 80% of the time like it was before, which is good I suppose but also quite frustrating to actually debug.

* 20% assigned to dead worker, waiting forever
* 20% cancelled (what is expected to happen)
* 60% waits a few moments on the resource manager queue presumably and then completes on a live worker

dralley:
Seems like at the point in time where the worker is being marked offline, the task hasn't been associated with the worker yet, so this query comes up empty.

https://github.com/pulp/pulpcore/blob/master/pulpcore/tasking/services/worker_watcher.py#L163

It does end up associated shortly thereafter.  This is the race that causes them not to be canceled.

PR #719, opened 2020-05-22
Teach exporter to understand, validate, and respect chunk_size= parameter

ggainey:
Since we can now have multiple-output-files, replaced filename/sha256
columns with output_file_info, a JSONField which is a dictionary of
filename: hash pairs.
closes #6736


pulpcore/app/tasks/export.py
```diff
@@ -141,62 +144,99 @@ def pulp_export(the_export):
...
-        export_versions(the_export, plugin_version_info)
-        # Export the top-level entities (artifacts and repositories)
-        # Note: we've already handled "what about incrementals" when building the 'artifacts' list
-        export_artifacts(the_export, artifacts)
-        # Export the repository-version data, per-version
-        for version in versions_to_export:
-            export_content(the_export, version)
-            ExportedResource.objects.create(export=the_export, content_object=version)
 
+def compute_hash(filename):
```
daviddavis, line 196: I'm wondering if this should be `_compute_hash(filename)`?

pulpcore/app/tasks/export.py
```diff
@@ -141,62 +144,99 @@ def pulp_export(the_export):
...
             sha256_hash.update(byte_block)
-        the_export.sha256 = sha256_hash.hexdigest()
-    the_export.filename = tarfile_fp
-    the_export.save()
-    pulp_exporter.last_export = the_export
-    pulp_exporter.save()
+        return sha256_hash.hexdigest()
+
+
+def do_export(pulp_exporter, tar, the_export):
```
daviddavis, line 205: And `_do_export(...)`?

daviddavis:
Looks good. I think this probably requires a `.removal` file in CHANGES since we are removing/moving fields.

ggainey:
> Looks good. I think this probably requires a `.removal` file in CHANGES since we are removing/moving fields.

Good catch, added!

PR #725, opened 2020-05-26
Removes Content-Encoding header from pulpcore-content responses.

dkliban:
This header is not needed because the server is not actually encoding the response.

fixes: #6831
https://pulp.plan.io/issues/6831


dralley:
Needs a changelog entry

PR #726, opened 2020-05-26
Views can specify the tag name

fao89:
https://pulp.plan.io/issues/6832
closes #6832

### DIFF
--- a/CHANGES/plugin_api/6832.feature
+++ b/CHANGES/plugin_api/6832.feature
@@ -1,0 +1,1 @@
+Views can specify the tag name with `pulp_tag_name`
--- a/docs/plugins/reference/how-to-doc-api.rst
+++ b/docs/plugins/reference/how-to-doc-api.rst
@@ -69,3 +69,18 @@ as a GET parameter. For example for pulp_rpm only endpoints use a query like thi
 .. code-block:: bash
 
     curl -o api.json http://localhost:24817/pulp/api/v3/docs/api.json?plugin=pulp_rpm
+
+OpenAPI Tags
+------------
+
+Tags are used to group operations logically into categories, such as RepositoriesRpm,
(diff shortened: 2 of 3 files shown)


CHANGES/6832.feature
```diff
@@ -0,0 +1 @@
+Views can specify the tag name with `pulp_tag_name`
```
daviddavis, line 1: Should this go into the plugin_api folder?

daviddavis:
Can you add some docs for the plugin writer too?

CHANGES/6832.feature
fao89: not sure

fao89:
> Can you add some docs for the plugin writer too?

yes, I didn't add before because I thought it would go as a hidden feature on 3.4.0

CHANGES/6832.feature
daviddavis: I think the main difference is that changes in `CHANGES/plugin_api` are for plugin writers while `CHANGES` are for endusers. If I understand this change correctly, it seems more geared toward plugin writers?

daviddavis:
> yes, I didn't add before because I thought it would go as a hidden feature on 3.4.0

I think this'll probably land in 3.5.0.

CHANGES/6832.feature
fao89: yep, it is for plugin writers, I'll move to `CHANGES/plugin_api` then

fao89:
> Can you add some docs for the plugin writer too?

Is it a good place to add the docs?
https://docs.pulpproject.org/plugins/reference/how-to-doc-api.html

daviddavis:
I'm not sure. This might also be a good place:

https://docs.pulpproject.org/plugins/plugin-writer/concepts/subclassing/viewsets.html

@bmbouter any thoughts?

PR #745, opened 2020-06-06
Remove import/export's dependency on pulp_ids.

### LINKED ISSUE #6807: [PULP-659] Remove legacy django storages
Django legacy storage was depreated and marked for removal in 3.85.

References:
- **Deprecation issue:** https://github.com/pulp/pulpcore/issues/5404
- **Discourse post:** https://discourse.pulpproject.org/t/action-required-upgrade-your-storage-settings-before-pulpcore-3-85/2072/2
- **Jira**: https://issues.redhat.com/browse/PULP-659

ggainey:
Dealing this required slimming down the fields we export/import,
so #6515 got addressed along the way as well.

closes #6807
closes #6515


pulpcore/app/tasks/importer.py
```diff
@@ -40,17 +39,22 @@
...
-        resource = resource_class()
-        return resource.import_data(data, raise_errors=True)
+    try:
+        log.info(_("Importing file {}.").format(fpath))
+        with open(fpath, "r") as json_file:
+            data = Dataset().load(json_file.read(), format="json")
+            resource = resource_class()
+            log.info(_("...Importing resource {}.").format(resource.__class__.__name__))
+            return resource.import_data(data, raise_errors=True)
+    except AttributeError:
```
daviddavis, line 49: When does AttributeError get raised?

ggainey: > When does AttributeError get raised?

When the export-file isn't 'importable' for some reason. Without this, it can be hard to figure out exactly which import-class a failure-stack-trace is from, with log-entries being interleaved due to import being multithreaded now.

daviddavis:
This change looks good to me. So export_id on the Content in the downstream Pulp stores the pulp_id of the upstream Pulp server? Does it handle the case where the pulp_id in the upstream Pulp changes? I'm imagining this could happen if a Content gets removed/orphan cleaned up, and then created again later.

ggainey:
> This change looks good to me. So export_id on the Content in the downstream Pulp stores the pulp_id of the upstream Pulp server? Does it handle the case where the pulp_id in the upstream Pulp changes? I'm imagining this could happen if a Content gets removed/orphan cleaned up, and then created again later.

For Content, export_id is Just Another Field, that will be updated when the matching natural-key-content gets imported/updated. The flow is a little complicated, but should Just Work.

Consider the following scenario:

1. In upstream, package foo-1.2.3-4 has pulp_id 100, relative_path '/foo.rpm', and is exported and then imported.
2. There now exists in downstream a core_content entry, pulp_id 2000, export_id 100, and a matching rpm_rpmpackage entry. core_contentartifact has content_id 2000 and relative_path '/foo.rpm'.
3. Upstream deletes and recreates foo-1.2.3-4, so that it now has export_id 157.
4. On import, we end up updating core_content 2000 (since natural-key works), export_id 157.
5. Core_contentartifact import happens. We attempt to create a core_contentartifact by looking for export_id=157. We find pulp_id 2000, find the existing core_contentartifact record, and the update happens correctly.

I **think** we're good in this case.

daviddavis:
Code LGTM. I am going to test this out. @bmbouter or @dkliban can we get a second review?

CHANGES/6515.bugfix
```diff
@@ -0,0 +1 @@
+Removed unnecessary fields from the import/export transfer.
```
bmbouter, line 1: I think this should be `.removal`. Even though it's tech preview the changelog should show removals that require db migrations.

pulpcore/app/models/content.py
```diff
@@ -257,6 +257,8 @@ class Content(MasterModel, QueryMixin):
 
     objects = BulkCreateManager()
 
+    export_id = models.UUIDField(null=True)  # Used by PulpImport/Export processing
```
bmbouter, line 260: This is a significant change because all plugins will have to make migrations. The docstring above needs it to be explained. What does this do and why here?

daviddavis: So it's not as significant as it looks. It's on the base content table and doesn't need to be populated. It's essentially data that's stored during an import to link up Content to ContentArtifact since the Content table doesn't have a natural key.

ggainey: None of the plugins need migrations, it's happening in core_content only - pulp_file and pulp_rpm already work with it, only pulpcore needed a migration.

See https://github.com/pulp/pulpcore/pull/745#issuecomment-640777700 for an explanation. The net is, core_contentartifact has no 'natural key' other than pulp_id, and doesn't know what it's 'pointing to' at this part of the process. export_id gives us a way to teach ContentArtifactResource a way to find the right ContentResource to point itself to when being imported.

CHANGES/6515.bugfix
ggainey: Sure, can do.

ggainey: Is having a removal going to want pulpcore's semantic versioning to want to go to 4.0? Because that would be inappropriate - it's why we released this as 'tech preview', we knew we were likely to make changes.

pulpcore/app/modelresource.py
```diff
@@ -7,36 +11,66 @@
...
+
+
 #
 # Artifact and Repository are different from other import-export entities, in that they are not
 # repo-version-specific.
 #
 class ArtifactResource(QueryModelResource):
     class Meta:
         model = Artifact
+        exclude = (
```
bmbouter, line 24: I'm a bit out of the loop on this sorry. Why are these being excluded even if they aren't being used as the primary identifier anymore?

ggainey: Excluded fields 'belong to' the downstream - if they're included, the import-process will use them. That's inappropriate for created/updated dates, and **disastrous** for pulp_id.

bmbouter: Oh I think I see what you're saying. So these are the fields that will be contained in the export itself?

pulpcore/app/models/content.py
bmbouter: I see what you all are saying about the detail models not needing a migration. Is the data only stored while an import is running and then when the import completes it's no longer stored? I'm trying to determine if this is a field to be stored in the long-term or not.

CHANGES/6515.bugfix
daviddavis: No there are two cases where having a removal (or backwards incompatible change) are permitted where we can keep pulpcore at 3.x: if there's a bug to fix or if the code is at tech preview. I think this meets both of these given that you're fixing a bug and import/export is tech preview.

ggainey: OK cool, just didn't want it to automagically increment during the release process :)

pulpcore/app/models/content.py
ggainey: In the database, a given exported Detail ends up in two tables - core_content, and the detail-specific table, like say file_filecontent. This field exists *only* in core_content. It is filled when/if a given detail is imported - which action results in both the core_content and the <detail>content table-entries being created and filled in, because that's how django-import-export works. It then exists in the table (and the model, but nobody cares about its existence other than import/export). Import of ContentArtifact relies on export_id existing, to have any hope of recreating the core/artifact mapping being imported.

pulpcore/app/modelresource.py
ggainey: Yes exactly - this is about controlling what ends up in the export-file.

pulpcore/app/models/content.py
```diff
@@ -257,6 +257,8 @@ class Content(MasterModel, QueryMixin):
 
     objects = BulkCreateManager()
 
+    upstream_id = models.UUIDField(null=True)  # Used by PulpImport/Export processing
```
bmbouter, line 260: Can this be grouped with the fields above (the `upstream_id`)?

bmbouter: Also this needs a docstring update above please.

pulpcore/app/modelresource.py
```diff
@@ -7,36 +11,66 @@
...
+class ContentArtifactResource(QueryModelResource):
 
-    class Meta:
-        model = Content
-        fields = ("pulp_id", "pulp_type")
+    artifact = fields.Field(
+        column_name="artifact", attribute="artifact", widget=ForeignKeyWidget(Artifact, "sha256")
+    )
 
+    def before_import_row(self, row, **kwargs):
```
bmbouter, line 55: This probably should have a docstring.

pulpcore/app/modelresource.py
```diff
@@ -7,36 +11,66 @@
...
+            "pulp_created",
+            "pulp_last_updated",
+            "content",
+        )
 
 
 #
-# Content, and ContentArtifact are per-repo-version import/exports, and can
-# follow the same pattern as a plugin writer would follow
+# ContentArtifact is different from other import-export entities because it has no 'natural key'
```
bmbouter, line 45: I think this should be moved to the class docstring.

ggainey: Done!

pulpcore/app/modelresource.py
ggainey: Done!

pulpcore/app/models/content.py
ggainey: Both done!

CHANGES/6807.bugfix
```diff
@@ -0,0 +1 @@
+Cured import/export's addiction to pulp_ids.
```
goosemania, line 1: Is it clear for a user who would read a changelog what's fixed?
maybe it can be .misc, so implementation details are not exposed to a user.

ggainey: This entire PR is nothing but low-level implementation change. From the user's POV, there isn't anything to see here - import/export works as before. From a plugin-author's POV, that is a meaningful changelog (or at least meaningful enough to go look at the attached issue). I honestly don't see how moving this to 'misc' changes anything?

goosemania: Moving to .misc makes the changelog not published/visible.
I'm not sure if it's a good proposal or not. We have a dedicated plugin-api changelog, though this change is not plugin api per se.

We can leave it as is, it's just quite confusing for users, imo, and the changelog is for them.

ggainey: As a side comment - ".misc makes the changelog not published/visible" was news to me. Our docs say nothing about it, and I see that there is one line in the towncrier doc that could be taken as implying that, but it never occurred to me that it means "not emitted".

ggainey: I'll rename the file and repush.

bmbouter: I was going to leave a similar comment. I also thought a user reading this would not understand it.

pulpcore/app/modelresource.py
```diff
@@ -7,36 +11,77 @@
...
+
+    def before_import_row(self, row, **kwargs):
+        """
+        Fixes the content-ptr of an incoming content-artifact row at import time.
+
+        Finds the 'original uuid' of the Content for this row, looks it up as the
+        'upstream_id' of imported Content, and then replaces the Content-pk with its
+        (new) uuid.
+
+        :param row: incoming import-row representing a single ContentArtifact
```
bmbouter, line 65: We use a different style guide so that all autodoc can work as expected https://docs.pulpproject.org/contributing/style-guide.html?highlight=napoleon#example-docstring

ggainey: And done

PR #748, opened 2020-06-09
RemoteArtifact size field type

pavelpicka:
Change RemoteArtifacts' size from Integer to BigInteger to reflect same field in Artifact.

closes: #6717
https://pulp.plan.io/issues/6717


CHANGES/6717.misc
```diff
@@ -0,0 +1 @@
+RemoteArtifact size field changed to BigIntegerField to reflect Artifact class.
```
daviddavis, line 1: This probably is a bugfix as misc entry texts don't appear on the changelog and we want this to be visible to users.

Also, I would also probably mention that this fixes `integer out of range` errors when syncing. Maybe something like: "Fixed ``integer out of range`` errors during sync by changing RemoteArtifact size field to BigIntegerField".

pulpcore/app/models/content.py
```diff
@@ -360,7 +360,7 @@ class RemoteArtifact(BaseModel, QueryMixin):
     """
 
     url = models.TextField(validators=[validators.URLValidator])
-    size = models.IntegerField(null=True)
+    size = models.BigIntegerField(null=True)
```
goosemania, line 363: Please fix the doc string here and on the Artifact model

CHANGES/6717.misc
pavelpicka: changed to bugfix

pulpcore/app/models/content.py
pavelpicka: fixed docsctring for both models

goosemania:
@pavelpicka , I noticed one more thing. Could you rename the migration file to have a meaningful name. It's considered a good practice not to have auto_*. It's easier to look through the list of migrations to find the one you need.

The majority of migrations are named, we have some auto_* ones but I believe it's just an oversight.

PR #749, opened 2020-06-13
Rewrite QueryExistingContent stage for performance

dralley:
Making the following assumptions:

* Even though total uniqueness is only guaranteed by the full natural key, most content types
  have a single attribute which gets you 99.9% of the way there, e.g. "digest" for file content and
  "pkgId" for RPM content and "filename" for Python content.  It is far, far cheaper to query by a
  single attribute than to query for sets of many attributes, both in terms of query construction
  (ORM has a high overhead here) and the ability of the database to optimize it.
* If we were to query by a single attribute to find existing content, we may get back "false positives"
  which have a duplicate value of this attribute but with a different overall natural key. If discriminating
  and ignoring these is sufficiently cheap and if there are sufficiently few of them, then the overall
  performance can be increased.

The idea is that instead of sending PostgreSQL queries that look like this:

```
SELECT ... FROM rpm_package WHERE
   (name='tmux' AND epoch='0' AND version='2' AND release='1' AND arch='x86_64' AND ...) OR
   (name='ssh' AND  epoch='0' AND version='3' AND release='2' AND arch='x86_64' AND ...) OR
   ...
   x100
```

We send it vastly more efficient (to execute, and to construct) queries that look like this:

```
SELECT ... FROM rpm_package WHERE name IN {'tmux', 'ssh', ..., x100}
```

And, by making it sufficiently inexpensive to compare natural keys between the
units from the database query (filled with potential false positives) and the
incoming content (the batch), we can gain a large overall net win in terms of
performance.

[noissue]

### DIFF
--- a/pulpcore/plugin/stages/content_stages.py
+++ b/pulpcore/plugin/stages/content_stages.py
@@ -1,7 +1,7 @@
 from collections import defaultdict
+from itertools import chain
 
 from django.db import IntegrityError, transaction
-from django.db.models import Q
 
 from pulpcore.plugin.models import ContentArtifact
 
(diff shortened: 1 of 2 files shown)


dralley:
# File plugin, on-demand sync, 30k files

## Initial results

Initial sync: 51 seconds
Re-sync: 23 seconds

## Rewritten content stage

Initial sync: 40 seconds
Re-sync: 10 seconds

CProfile of resync perfrormance before and after - QueryExistingContent drops from 50% of runtime to 5% of runtime

original_content_resync

rewritten_content_resync

dralley:
# Rpm plugin, on-demand sync, ~11,500 packages

Times are averages of 3 runs.  I used CentOS 7 online repo.

## Initial results

Initial sync: 46 seconds
Re-sync: 25 seconds

## Rewritten content stage

Initial sync: 39 seconds
Re-sync: 14 seconds

dralley:
For what it's worth, I haven't and can't easily test performance where there actually are a lot of false positives, like if you have 100 different versions of an RPM or something and only one of them is the right one, but I'm not concerned for two reasons:

* The extra work being done here is still pretty cheap in a relative sense, I don't think it would scale too poorly even with a bunch of false positives
* There's obvious workarounds:
    * many content types like RPM have a built-in checksum that could be placed at the front of the list in place of name
    * we could also do more than one round of filtering, and it would still be faster when structured this way rather than a list of natural keys OR'd together


there are a bunch of ways to address that.  RPMs have a checksum which we could place at the front of the list instead of name, which would pretty much resolve the issue straight away.

dralley:
@ipanova I fixed the issue so that now all DeclarativeContent that are passed in should be passed back out, but it is possible that the order of the DeclarativeContent within the batch itself might change.  Is that a problem?

ipanova:
@dralley if the order changes only *within* the batch then it should not be an issue, important not to mix up batches because in container plugin and migration plugin we rely on the order the content types we send into the pipeline

daviddavis:
@gmbnomis any chance you can help us review this change?

pulpcore/plugin/stages/content_stages.py
```diff
@@ -37,29 +37,75 @@ async def run(self):
...
+                        key_attribute_by_type[model_type] = key_attribute
+
+                    key_attribute_values_by_type[model_type].add(
+                        getattr(d_content.content, key_attribute)
+                    )
+
+                else:
+                    # Content that is already saved is already unique / deduplicated, no further
+                    # processing needed.
+                    deduplicated_content.append(d_content)
```
mdellweg, line 68: Instead of collecting them we could just hand them to the next stage right away.

dralley: Probably, yeah

pulpcore/plugin/stages/content_stages.py
```diff
@@ -37,29 +37,75 @@ async def run(self):
...
-            for d_content in batch:
+
+                    # This could produce false positive duplicates because it matches only on a
+                    # single attribute of the content -- but we handle this when processing the
+                    # results of this query.
+                    try:
+                        key_attribute = key_attribute_by_type[model_type]
+                    except KeyError:
+                        key_attribute = model_type.natural_key_fields()[0]
+                        key_attribute_by_type[model_type] = key_attribute
```
dralley, line 59: As I mentioned, the first attribute in the natural key is usually one of the better ones to use, but not necessarily the best for all plugins.  We could either tell plugin writers to put their most unique field first, or we could add a parameter to the ```class Meta``` to define it explicitly, which I think should be optional because requiring it would be a breaking change.

mdellweg:
> @dralley if the order changes only _within_ the batch then it should not be an issue, important not to mix up DC from different batches because in container plugin and migration plugin we rely on the order the content types we send into the pipeline

That is a bad idea, and quite honestly, i cannot believe it. At the very least the DownloadArtifact stage spits out the content_units once they are finished downloading. There is no such guarantee for the order. Also since each stage wraps up their own batches, there is no guarantee, that the same content units will make up the same batch in different stages.

dralley:
@mdellweg Just for clarity, you're referring to the second part, where some plugins depend on content from batches not being intermixed?

mdellweg:
Yes. That is my concern. Because there is no such guarantee, that the batches propagate through the stages.
The "Bad idea" is to depend on the order of declarative content to traverse the pipeline.

dralley:
My only concern is whether this assumption holds for all plugin types and whether the consequences of it not holding would be worse performance for those plugin types.

> * Even though total uniqueness is only guaranteed by the full natural key, **most content types
    have a single attribute which gets you 99.9% of the way there**, e.g. "digest" for file content and
    "pkgId" for RPM content and "filename" for Python content. It is far, far cheaper to query by a
    single attribute than to query for sets of many attributes, both in terms of query construction
    (ORM has a high overhead here) and the ability of the database to optimize it.
> * If we were to query by a single attribute to find existing content, we may get back "false positives"
which have a duplicate value of this attribute but with a different overall natural key. If discriminating
and ignoring these is sufficiently cheap and **if there are sufficiently few of them**, then the overall
performance can be increased.

The container plugin and ansible plugin are what I'm not sure about, primarily. @ipanova @fao89 @daviddavis could you weigh in here?

ipanova:
> > @dralley if the order changes only _within_ the batch then it should not be an issue, important not to mix up DC from different batches because in container plugin and migration plugin we rely on the order the content types we send into the pipeline
> 
> That is a bad idea, and quite honestly, i cannot believe it. At the very least the DownloadArtifact stage spits out the content_units once they are finished downloading. There is no such guarantee for the order. Also since each stage wraps up their own batches, there is no guarantee, that the same content units will make up the same batch in different stages.

That's true, that's why futures are used for this purpose. This way you can control what content type is being processed and in which order.

ipanova:
> My only concern is whether this assumption holds for all plugin types and whether the consequences of it not holding would be worse performance for those plugin types.
> 
> > * Even though total uniqueness is only guaranteed by the full natural key, **most content types
> >   have a single attribute which gets you 99.9% of the way there**, e.g. "digest" for file content and
> >   "pkgId" for RPM content and "filename" for Python content. It is far, far cheaper to query by a
> >   single attribute than to query for sets of many attributes, both in terms of query construction
> >   (ORM has a high overhead here) and the ability of the database to optimize it.
> > * If we were to query by a single attribute to find existing content, we may get back "false positives"
> >   which have a duplicate value of this attribute but with a different overall natural key. If discriminating
> >   and ignoring these is sufficiently cheap and **if there are sufficiently few of them**, then the overall
> >   performance can be increased.
> 
> The container plugin and ansible plugin are what I'm not sure about, primarily. @ipanova @fao89 @daviddavis could you weigh in here?

I tested some in pulp_container and have not seen any major issues, i mean everything works as expected - content is correctly synced and relations are created. I have tested though 2 small/medium repos

mdellweg:
> That's true, that's why futures are used for this purpose. This way you can control what content type is being processed and in which order.

I agree. Futures are a primitive to schedule work just about when other work is finished. So they are a synchronization tool (as mutexes and conditions are).

dralley:
The regression in "subsequent sync" performance concerns me a little bit because it is the most common case.   It is good to know that it is minor at worst, but I think it still warrants some further investigation on my part to see if there is anything that can be done.

PR #751, opened 2020-06-15
Rewrite QueryExistingArtifacts stage for performance

dralley:
Use "IN" instead of "OR" which is easier for PostgreSQL to optimize.

[noissue]

### DIFF
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -1,8 +1,9 @@
 import asyncio
+from collections import defaultdict
 from gettext import gettext as _
 import logging
 
-from django.db.models import Q, Prefetch, prefetch_related_objects
+from django.db.models import Prefetch, prefetch_related_objects
 
 from pulpcore.plugin.models import Artifact, ContentArtifact, ProgressReport, RemoteArtifact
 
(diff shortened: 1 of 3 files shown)


dralley:
Measured on top of my previous PR:

File Plugin:

initial sync: 37 seconds (vs 40 before)
re-sync: 7 seconds (vs 10 before)

RPM Plugin:

initial sync: 36 seconds (vs 39 before)
resync: 9 seconds  (vs 14 before)

dralley:
Here's some Cprofile graphs from a branch that includes both these changes and the changes made in https://github.com/pulp/pulpcore/pull/749

Initial sync

file_rewritten_artifact_initial

Re-sync

file_rewritten_artifact_resync

daviddavis:
@gmbnomis any chance you can help us review this change?

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -42,21 +43,39 @@ async def run(self):
...
             for d_content in batch:
                 for d_artifact in d_content.d_artifacts:
-                    one_artifact_q = d_artifact.artifact.q()
-                    if one_artifact_q:
-                        all_artifacts_q |= one_artifact_q
+                    if d_artifact.artifact._state.adding:
+                        for digest_type in Artifact.COMMON_DIGEST_FIELDS:
+                            digest_value = getattr(d_artifact.artifact, digest_type)
+                            if digest_value:
+                                artifact_digests_by_type[digest_type].append(digest_value)
```
mdellweg, line 58: If we collected the d_artifacts by diegest in dictionary here, we could make a lookup for all `.keys()` and we didn't need the triple of nested loops further down to match them up again.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -42,21 +43,39 @@ async def run(self):
...
         """
         async for batch in self.batches():
-            all_artifacts_q = Q(pk__in=[])
+            artifact_digests_by_type = defaultdict(list)
+
+            # For each unsaved artifact, check its digests in the order of COMMON_DIGEST_FIELDS
+            # and the first digest which is found is added to the list of digests of that type.
+            # We assume that in general only one digest is provided and that it will be
+            # sufficient to identify the Artifact.
             for d_content in batch:
```
mdellweg, line 52: Can we assume here, that already saved content does not need its artifacts to be reevaluated?
This could speed up resyncs dramatically.

dralley: Hmmm, not sure.  It's an interesting idea.  Probably the most useful thing to focus on now is the ```RemoteArtifactSaver``` stage though since it's now 30% of the runtime.

pulpcore/plugin/stages/artifact_stages.py
dralley: I thought about that, I think it's probably viable but I was trying to be (probably unnecessarily) cautious about not messing up any of the relationships.  It should be fine as long as the objects act like references and not copies (which I think is the case).

mdellweg: Yes, exactly, you pass the d_artifacts around as references and change their contained reference to the artifacts you found. I am sure this will work.
I just realized however, that you might have the same artifact digest in d_artifacts coming from different content_units, so you probably need to make it a dictionary of digest -> list of d_artifacts.

dralley:
The combination of both of these PRs improves migration performance a little bit.

Before:

Intial migration: 890 seconds

Rewritten stages:

Intial migration: 784 seconds

PR #752, opened 2020-06-17
Add a Pulp user-agent string to all aiohttp requests

dralley:


pulpcore/download/factory.py
```diff
@@ -15,6 +19,17 @@
...
 
+def user_agent():
+    """
+    Produce a User-Agent string to identify Pulp and relevant system info.
+    """
+    pulp_version = get_distribution('pulpcore').version
+    python = "{} {}.{}.{}-{}{}".format(sys.implementation.name, *sys.version_info)
+    uname = platform.uname()
+    system = f"{uname.system} {uname.machine})"
+    return f"pulp/{pulp_version} ({python}, {system}) (aiohttp {aiohttp_version})"
```
dralley, line 30: ```pulp/3.5.0.dev0 (cpython 3.8.3-final0, Linux x86_64)) (aiohttp 3.6.2)```

dralley:
I still need to verify that it actually sends the header

pulpcore/download/__init__.py
```diff
@@ -1,4 +1,4 @@
 from .base import BaseDownloader, DownloadResult  # noqa
-from .factory import DownloaderFactory  # noqa
+from .factory import DownloaderFactory, user_agent  # noqa
```
dralley, line 2: There's prooobably no need to export this?

bmbouter: What do you mean export?

pulpcore/download/factory.py
bmbouter: I think this should be s/pulp/pulpcore/

pulpcore/download/__init__.py
dralley: Make it available from the plugin API

bmbouter: Let's not for now. Generally things are added to the plugin api at the time they are used.

pulpcore/download/factory.py
dralley: I'm ambivalent on that one because plugins can use these downloaders as well, do they not?  Traffic could either be driven by pulpcore (artifact download) or by the plugin itself, so I wouldn't want to be specific in an inaccurate way in the user agent.

bmbouter: This is inaccurate in all cases. `pulp` refers to nothing actually.

dralley: How about ```pulpproject```?

dralley:
@bmbouter Fixed

pulpcore/download/factory.py
bmbouter: I was thinking the same thing. Thank you!

CHANGES/6954.feature
```diff
@@ -0,0 +1 @@
+Provide a Pulp user agent string with all aiohttp requests by default.
```
bmbouter, line 1: Since it's a Pulp changelog I was thinking `/Pulp //`

PR #753, opened 2020-06-17
Describe permit in ContentGuard

mdellweg:


docs/plugins/reference/content-protection.rst
```diff
@@ -17,7 +17,10 @@ The ``ContentGuard`` is a Master/Detail object provided at
 github.com/pulp/pulpcore/blob/master/pulpcore/app/models/publication.py#L192-L202>`_.
 
 In your plugin code, subclass ``ContentGuard`` and optionally add additional fields as necessary to
-perform the authentication and authorization. As with all Master/Detail objects a ``TYPE`` class
+perform the authentication and authorization.
+Then overwrite the ``permit`` method so that it returns ``None`` if access is granted and throws a
+``PermissionError`` on denial.
+As with all Master/Detail objects a ``TYPE`` class
```
bmbouter, line 23:
```suggestion
perform the authentication and authorization. Then overwrite the ``permit`` method so that it returns ``None`` if access
is granted and throws a ``PermissionError`` on denial. As with all Master/Detail objects a ``TYPE`` class
```

docs/plugins/reference/content-protection.rst
```diff
@@ -17,9 +17,11 @@ The ``ContentGuard`` is a Master/Detail object provided at
 github.com/pulp/pulpcore/blob/master/pulpcore/app/models/publication.py#L192-L202>`_.
 
 In your plugin code, subclass ``ContentGuard`` and optionally add additional fields as necessary to
-perform the authentication and authorization. As with all Master/Detail objects a ``TYPE`` class
-attribute is needed which is then used in the URL. For ``ContentGuard`` detail objects the URL
-structure is::
+perform the authentication and authorization.
+Then overwrite the ``permit`` method so that it returns ``None`` if access is granted and throws a
```
bmbouter, line 21: This still seems to have the random newlines in it.

mdellweg: I think, that is now what you meant.
I still think, from a `git diff` perspective it is more productive to wrap lines at whole phrases.
It would only put the one sentence you changed in the commit, and not the rest of its paragraph.

PR #755, opened 2020-06-18
Add GroupPgressReport model and serializer.

ipanova:
[noissue]


pulpcore/app/serializers/progress.py
```diff
@@ -26,15 +26,31 @@ class ProgressReportSerializer(ModelSerializer):
...
+    message = serializers.CharField(
+        help_text=_("The message shown to the user for the group progress report."), read_only=True
+    )
+    code = serializers.CharField(
+        help_text=_("Identifies the type of group progress report'."), read_only=True
+    )
+    total = serializers.IntegerField(help_text=_("The total count of items."), read_only=True)
+    done = serializers.IntegerField(
+        help_text=_("The count of items already processed. Defaults to 0."), read_only=True
+    )
```
goosemania, line 49: Does it make sense to use PositiveIntegerField? I don't think there is a use case for negative numbers here. If it's the same as progress report has, then maybe it's worth keeping it the same for consistency.

ipanova: it does make sense to use a PositiveIntField, i don't think we want to have a negative result.
ProgressReport has same IntField as GroupReport so for consistency this update should happen in both places.

CHANGES/7011.removal
```diff
@@ -0,0 +1 @@
+Removed suffix field from ProgressBar.
```
bmbouter, line 1: To give some background on where it came from, the model layer was designed to align with the [progress](https://pypi.org/project/progress/) library in the CLI. The idea is that the CLI would report task progress as it happens using that library, and plguin writers could influence the "suffix" that is used there.

This isn't used today so I'm ok w/ removing it, but I wanted to share the thinking about why it exists in the first place.

dralley:
Looks good, but see my comment on the other PR about possibly adding some helpers (if you think it would improve the code)

CHANGES/6858.feature
```diff
@@ -0,0 +1 @@
+Added GroupProgressReport to track progress in a TaskGroup.
```
bmbouter, line 1: Is it possible to put an example of the URL here for users to see.

pulpcore/app/models/progress.py
```diff
@@ -215,3 +215,70 @@ def iter(self, iter):
         for x in iter:
             yield x
             self.increment()
+
+
+class GroupProgressReport(BaseModel):
```
bmbouter, line 220: I haven't built the docs but do the existing autodoc statements bring this into the documentation for plugin writers? If not can an autodoc statement be added?

bmbouter: Also should a paragraph or few lines of documentation be added here in addition? https://docs.pulpproject.org/plugins/plugin-writer/concepts/index.html#tasks There is a `Task Groups` section at the bottom there.

CHANGES/6858.feature
ipanova: it does not have a separate endpot
```
$ http GET $BASE_ADDR/pulp/api/v3/task-groups/ca57fbd1-d970-4b09-8258-81753a0dda76/
HTTP/1.1 200 OK
Allow: GET, HEAD, OPTIONS
Connection: close
Content-Length: 411
Date: Tue, 23 Jun 2020 11:54:10 GMT
Server: gunicorn/20.0.4
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "all_tasks_dispatched": true,
    "canceled": 0,
    "completed": 3,
    "description": "Migration Sub-tasks",
    "failed": 0,
    "group_progress_reports": [
        {
            "code": "create.repo_version",
            "done": 2,
            "message": "Repo version creation",
            "total": 2
        },
        {
            "code": "create.distribution",
            "done": 4,
            "message": "Distribution creation",
            "total": 4
        }
    ],
    "pulp_href": "/pulp/api/v3/task-groups/ca57fbd1-d970-4b09-8258-81753a0dda76/",
    "running": 0,
    "skipped": 0,
    "waiting": 0
}

```

bmbouter: Oh right! I forgot this only adds progress reporting not the entire task group interface. Let me look back at the comments to see if they make sense.

ipanova:
@bmbouter ready for re-review

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -73,8 +73,41 @@ Tasks are deployed from Views or Viewsets, please see :ref:`kick-off-tasks`.
 
 Sometimes you may want to create many tasks to perform different parts of one larger piece of work,
 but you need a simple means to track the progress of these many tasks. Task Groups serve this purpose
-by providing details on the number of associated tasks in each possible state. For more details,
-please see :ref:`kick-off-tasks`.
+by providing details on the number of associated tasks in each possible state.
+For more details, please see :ref:`kick-off-tasks`.
```
bmbouter, line 77: This is a nit, but these lines probably did not need to change.

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -73,8 +73,41 @@ Tasks are deployed from Views or Viewsets, please see :ref:`kick-off-tasks`.
...
+
+**GroupProgressReports**
+
+GroupProgressReport can track progress of each task in that group. GroupProgressReport needs to be
```
bmbouter, line 81: This part all looks good. Thanks for adding the docs.

pulpcore/app/models/progress.py
bmbouter: What I was trying to say with the autodoc is that we have this "Plugin API Reference" section [here](https://docs.pulpproject.org/plugins/api-reference/index.html) which is driven by [this portion of the docs](https://github.com/pulp/pulpcore/blob/master/docs/plugins/api-reference/). The idea is that our plugin API uses `.. automodule::` statements to bring in the object documentation as a full reference in this section.

In looking through that section though, it's a mess and out of date enough to the point that I'm not going to ask you to even put this in there because it needs its own revision. I filed this instead for us to fix later https://pulp.plan.io/issues/7045

ipanova:
github had some issues which led to the state when changes has been merged but PR has not changed the state. closing out this PR.

PR #763, opened 2020-06-27
don't go into an infinite loop when `CONTENT_PATH_PREFIX` is followed by `/`

SimonPe:
this would cause `path` to start with a `/`
and because
```
>>> os.path.split('/path')
('/', 'path')
>>> os.path.split('/')
('/', '')
```
`base` would never end up as `None` but stay as `/` resulting in an infinite loop.

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -151,7 +151,7 @@ class Handler:
         tree = []
         while True:
             base = os.path.split(path)[0]
-            if not base:
+            if not base.lstrip("/"):
                 break
             tree.append(base)
             path = base
--- a/CHANGES/7066.bugfix
+++ b/CHANGES/7066.bugfix
@@ -1,0 +1,1 @@
+Fixed denial of service caused by extra slashes in content urls.


daviddavis:
Thank you for the contribution! In order to move this forward, we'd need to file an issue. Would you mind doing so: https://pulp.plan.io/projects/pulp/issues/new ? Also, I'm not sure if I fully understand the reproducer so if you could include some steps to reproduce this, that'd be great.

SimonPe:
@daviddavis https://pulp.plan.io/issues/7066

pulpcore/content/handler.py
```diff
@@ -151,7 +151,7 @@ def _base_paths(path):
         tree = []
         while True:
             base = os.path.split(path)[0]
-            if not base:
+            if not base or base == '/':
```
bmbouter, line 154: @daviddavis in the example given this would match in both cases, so I was thinking this would break in too many cases. I didn't run it though, maybe in running it it makes sense?

bmbouter: On chat @daviddavis explained that the example isn't indicative of the problem like `/pulp/content//` is. With an example like `/pulp/content//` I can see how this is solving the problem.

daviddavis:
This works for me. Originally, the content app hangs and now it returns 404 when I hit `/pulp/content//`.

However, the problem is not fully fixed. `/pulp/content///` still hangs:

```
>>> os.path.split('//')
('//', '')
```

I think we need to check for 1 to N number of slashes.

SimonPe:
@daviddavis thanks for catching that, and thanks to that catch I realised we can just do `.lstrip('/')`

commit amended to do so.

daviddavis:
Awesome, thanks. This looks good to me.

In order to get the CI passing, you have [to add "fixes #7066" to your commit message](https://docs.pulpproject.org/contributing/git.html#commit-message) and [a changelog entry](https://docs.pulpproject.org/contributing/git.html#changelog-update) (`CHANGES/7066.bugfix`).

bmbouter:
This revision looks good to me.

daviddavis:
Hey @SimonPe. We'd love to get this merged. Do you think you could address the changelog and commit message changes?

SimonPe:
sorry, was a bit busy with other priorities the last couple of weeks, should be fine now.

PR #765, opened 2020-06-30
Make the permitted changelog extensions more prominent

quba42:
[noissue]

Just a suggestion.
I find myself repeatedly scanning the page for these and not finding them :wink:


quba42:
Should I reformat the paragraph such that lines are the same length again? (Would make the diff a bit unreadable...)

daviddavis:
Yes, can you wrap the texts and we'll merge this. Thank you. :+1:

PR #767, opened 2020-07-01
Fixed a number of typos in import/export docs.

ggainey:
Also made examples consistent.

### DIFF
--- a/CHANGES/6919.doc
+++ b/CHANGES/6919.doc
@@ -1,0 +1,1 @@
+Fixed a number of typos in the import/export workflow docs.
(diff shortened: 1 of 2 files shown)


CHANGES/6919.bugfix
```diff
@@ -0,0 +1 @@
+Fixed a number of typos in the import/export workflow docs.
```
daviddavis, line 1: I think this should probably be a `.doc` entry. What do you think?

ggainey: Well, mostly I think I'm an idiot. Fixing, one sec.

PR #770, opened 2020-07-02
Update password variable term

melcorr:
fixes #7065

Make passwords consistent across Ansible installer and Pulp core docs.

### DIFF
--- a/CHANGES/7065.doc
+++ b/CHANGES/7065.doc
@@ -1,0 +1,1 @@
+Made password variable consistent with Ansible installer example playbook
--- a/docs/installation/instructions.rst
+++ b/docs/installation/instructions.rst
@@ -40,9 +40,9 @@ Create your pulp_install.yml playbook to use with the installer (Uncomment the d
    - hosts: all
      vars:
        pulp_settings:
-         secret_key: secret
+         secret_key: << YOUR SECRET HERE >>
          content_origin: "http://{{ ansible_fqdn }}"
-       pulp_default_admin_password: password
+       pulp_default_admin_password: << YOUR SECRET HERE >>
        pulp_install_plugins:
          # galaxy-ng: {}
          # pulp-ansible: {}
@@ -127,7 +127,7 @@ PyPI Installation
 9. Run Django Migrations::
 
    $ pulpcore-manager migrate --noinput
-   $ pulpcore-manager reset-admin-password --password admin
+   $ pulpcore-manager reset-admin-password --password << YOUR SECRET HERE >>
 
 
 .. note::
@@ -236,4 +236,3 @@ These services can then be started by running::
     sudo systemctl start pulpcore-api
     sudo systemctl start pulpcore-worker@1
     sudo systemctl start pulpcore-worker@2
-


mdellweg:
I like the idea. However, i see another one inline 130
`   $ pulpcore-manager reset-admin-password --password admin`
where, for inconsistencies sake, we set the admin-password to `admin`.

melcorr:
@fao89 @mikedep333 hey guys, @mdellweg makes an excellent point. With regards to the Ansible installer docs here and in the Ansible installer guide, would you be comfortable with us using the same <<>> around all variable values. Consistency is key.

docs/installation/instructions.rst
```diff
@@ -42,7 +42,7 @@ Create your pulp_install.yml playbook to use with the installer (Uncomment the d
        pulp_settings:
          secret_key: secret
```
fao89, line 43: we need the same here

melcorr:
I am happy to implement feedback from @fao89 - but is there any chance that I could convince you guys of the original premise of the issue I opened: 

"Replace the Ansible installer section in the Pulpcore docs with just a description and redirect to the Pulp Ansible Installer docs"

Unless we are stricter with the docs or figure out a way to have a single source across different doc sites, there will be a lot of inconsistencies. If we link to the primary and most-likely-to-be-up-to-date docs, I think that we might avoid future confusion.

WDYT?

mdellweg:
Absolutely, if we have documentation for the ansible installer somewhere, we should not even try to keep a (usually outdated) copy of it on the main website. I think, this is the purpose, hyperlinks were invented for.

fao89:
> I am happy to implement feedback from @fao89 - but is there any chance that I could convince you guys of the original premise of the issue I opened:
> 
> "Replace the Ansible installer section in the Pulpcore docs with just a description and redirect to the Pulp Ansible Installer docs"
> 
> Unless we are stricter with the docs or figure out a way to have a single source across different doc sites, there will be a lot of inconsistencies. If we link to the primary and most-likely-to-be-up-to-date docs, I think that we might avoid future confusion.
> 
> WDYT?

I agree with you, just for sharing the context, this brief explanation came from here:
https://pulp.plan.io/issues/6658
By that time we didn't have https://pulp-installer.readthedocs.io/en/latest/

mikedep333:
> I am happy to implement feedback from @fao89 - but is there any chance that I could convince you guys of the original premise of the issue I opened:
> 
> "Replace the Ansible installer section in the Pulpcore docs with just a description and redirect to the Pulp Ansible Installer docs"
> 
> Unless we are stricter with the docs or figure out a way to have a single source across different doc sites, there will be a lot of inconsistencies. If we link to the primary and most-likely-to-be-up-to-date docs, I think that we might avoid future confusion.
> 
> WDYT?

I can go either way. I think a version of the Installer quickstart in our main docs (like we do now) has benefit, but it also increases our maintenance, and makes people more likely to overlook our Installer docs.

CHANGES/7065.doc
```diff
@@ -0,0 +1 @@
+Make password variable consistent with Ansible installer example playbook
```
lubosmj, line 1:
```suggestion
Made password variable consistent with Ansible installer example playbook
```

We decided to use the past simple tense for changelog messages. Refer to https://www.mail-archive.com/pulp-dev@redhat.com/msg03803.html to learn more.

melcorr:
I am so sorry that it took so long for me to get back to this. 
It kept getting pushed down my list. 
As we discussed in the previous sprint planning, I am OK with leaving this as-is for now and if we get several complaints about inconsistencies etc, we can then maybe revisit having only one source of truth.

PR #773, opened 2020-07-07
Remove non-model-fields from Meta.fields of filters

ggainey:
Changes include:
  * Meta.fields list may only reference fields that exist in Meta.model
  * 'exact' is redundant in Meta.fields qualifiers (per Mongo docs)
  * Upload.completed was removed in cf5bad, remove referencing filter

closes #6915

### DIFF
--- a/CHANGES/6915.bugfix
+++ b/CHANGES/6915.bugfix
@@ -1,0 +1,1 @@
+Corrected a number of filters to be django-filter-2.3.0-compliant.
--- a/pulpcore/app/viewsets/task.py
+++ b/pulpcore/app/viewsets/task.py
@@ -122,8 +122,6 @@ class WorkerFilter(BaseFilterSet):
         fields = {
             "name": NAME_FILTER_OPTIONS,
             "last_heartbeat": DATETIME_FILTER_OPTIONS,
-            "online": ["exact"],
-            "missing": ["exact"],
         }
 
     def filter_online(self, queryset, name, value):
--- a/pulpcore/app/viewsets/content.py
+++ b/pulpcore/app/viewsets/content.py
@@ -35,13 +35,12 @@ class ArtifactFilter(BaseFilterSet):
     class Meta:
         model = Artifact
         fields = {
-            "repository_version": ["exact"],
-            "md5": ["exact"],
-            "sha1": ["exact"],
-            "sha224": ["exact"],
-            "sha256": ["exact"],
-            "sha384": ["exact"],
-            "sha512": ["exact"],
+            "md5",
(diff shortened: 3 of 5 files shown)


daviddavis:
> 'exact' is redundant in Meta.fields qualifiers (per Mongo docs)

Mongo docs?

ggainey:
> > 'exact' is redundant in Meta.fields qualifiers (per Mongo docs)
> 
> Mongo docs?

Oh geez. Lemme fix and repush, thanks for the catch.

CHANGES/6915.bugfix
```diff
@@ -0,0 +1 @@
+Corrected a number of filters to be django-filter-2.3.0-compliant.
```
daviddavis, line 1: I'm conflicted about whether this should be a misc or a bugfix. You think bugfix?

daviddavis:
Tested this out and it seems the filters still work so :+1:

CHANGES/6915.bugfix
ggainey: It's a bugfix - it was harmless, but Worng, and in the next release of django-filters it will stop working at all. We should never have been doing this.

pulpcore/app/viewsets/task.py
```diff
@@ -122,8 +122,6 @@ class Meta:
         fields = {
             "name": NAME_FILTER_OPTIONS,
             "last_heartbeat": DATETIME_FILTER_OPTIONS,
-            "online": ["exact"],
-            "missing": ["exact"],
```
dralley, line 126: I don't think this is semantically equivalent.  Does this not tell django-filters that you cannot filter on the "online" or "missing" fields? 

I'm also entirely confused why this is necessary.  This form is still present in the ```django-filter``` [docs](https://django-filter.readthedocs.io/en/stable/ref/filterset.html) and I can't imagine they would break this in a minor release anyways (we're on 2.2 currently).  Why change it?

daviddavis: It's redundant since the online and missing filters are defined above. You can still filter on online and missing. Before django-filter 2.3, it just ignored these lines but with django-filter 2.3 it now complains that these fields aren't actually fields.

dralley: You're right, I guess I'm just perplexed that they went from allowing it without a warning to making it a hard error in a point release.

daviddavis: Yea, me too. I did some digging and it looks like this is the issue: https://github.com/carltongibson/django-filter/issues/1013

PR #775, opened 2020-07-07
Revert "Splitting seralizers into read and write ones"

fao89:
Reverts pulp/pulpcore#707

### DIFF
--- a/CHANGES/7087.removal
+++ b/CHANGES/7087.removal
@@ -1,0 +1,2 @@
+Removes the Write models from the OpenAPI schema.
+Brings back the models that were accidentally removed from the OpenAPI schema in 3.4.0 release.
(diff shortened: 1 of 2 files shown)


fao89:
Should we file an issue and add a changelog?
@bmbouter @daviddavis @dkliban

mdellweg:
Is reverting not introducing the original problem?

dkliban:
> Is reverting not introducing the original problem?

Yes, we decided to abandon this approach to solving the problem. Instead we are going to switch to using OpenAPI v3 for pulpcore 3.6. We can achieve this by switching to drf-spectacular. drf-spectacular provides a guide for migrating from drf-yasg. 

https://drf-spectacular.readthedocs.io/en/latest/
https://drf-spectacular.readthedocs.io/en/latest/drf_yasg.html

CHANGES/7087.removal
```diff
@@ -0,0 +1 @@
+Revert: Fixed a bug that prevented write_only fields from being present in the API docs and bindings
```
dkliban, line 1: Removes the Write models from the OpenAPI schema. This also brings back the models that were accidentally removed from the OpenAPI schema in 3.4.0 release.

PR #780, opened 2020-07-09
Fixed WorkerDirectory.delete() causing infinite recursion when permission errors occur

gerrod3:
WorkerDirectory.delete() now only tries one additional time to delete its directory tree after encountering a PermissionError

fixes #6504
https://pulp.plan.io/issues/6504

### DIFF
--- a/pulpcore/tasking/services/storage.py
+++ b/pulpcore/tasking/services/storage.py
@@ -73,16 +73,23 @@ class WorkerDirectory:
         """
         Delete the directory.
 
-        On permission denied - an attempt is made to recursively fix the
+        On permission denied - an attempt is made to fix the
         permissions on the tree and the delete is retried.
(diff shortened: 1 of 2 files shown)


pulpcore/tasking/services/storage.py
```diff
@@ -69,20 +69,29 @@ def create():
...
 
+    def _delete(self):
+        """
+        Helper method for delete
+        """
+        try:
+            shutil.rmtree(self.path)
+        except FileNotFoundError:
+            pass
+        except PermissionError:
```
dralley, line 80: This isn't strictly necessary since it's just being raised afterwards -- it doesn't hurt anything though.

If you do decide to remove it, could you also move this method below the ```delete()``` method? Both of these are no big deal though.

PR #788, opened 2020-07-14
Allow Remotes to be associated to Repositories

daviddavis:
fixes #7015
https://pulp.plan.io/issues/7015


daviddavis:
Example:

```
http POST $BASE_ADDR/pulp/api/v3/remotes/file/file/ \
    name='bar' \
    url='https://fixtures.pulpproject.org/file/PULP_MANIFEST'
export REMOTE_HREF=$(http $BASE_ADDR/pulp/api/v3/remotes/file/file/ | jq -r '.results[] | select(.name == "bar") | .pulp_href')

http POST $BASE_ADDR/pulp/api/v3/repositories/file/file/ name=foo remotes:="[\"$REMOTE_HREF\"]" \
export REPO_HREF=$(http $BASE_ADDR/pulp/api/v3/repositories/file/file/ | jq -r '.results[] | select(.name == "foo") | .pulp_href')

http POST $BASE_ADDR$REPO_HREF'sync/'
```

pulpcore/app/serializers/repository.py
```diff
@@ -164,6 +183,21 @@ class RepositorySyncURLSerializer(serializers.Serializer):
         ),
     )
 
+    def validate(self, data):
+        try:
+            repo = models.Repository.objects.get(pk=self.context["repository_pk"])
+        except KeyError:
+            remote = None
+        else:
+            remote = repo.remotes.first()
```
dralley, line 192: I'm not really sure I feel about this - picking the first remote out of many.  It feels arbitrary even though it is probably the best decision if we do truly want to associate multiple remotes.  I guess I just don't really understand the use case that well.

@bmbouter , question here: https://pulp.plan.io/issues/7015#note-23

dralley: I don't want to block merging or anything since you guys do understand the use cases, so don't take this as a -1

daviddavis: The idea was to only support a single remote on repository but have a many-to-many join from repos to remotes in case some day we wanted to support associating multiple remotes with repos (and not break semver).

That said, I don't think this is happening in the near term (ie in pulp 3) and it's confusing for users so I updated the PR to only have a remote field on Repository.

dralley: So the idea is we're only externally supporting a single remote, but using a ManyToMany field anyway to possibly make future migration easier?

dralley: Idk, it's not difficult at all to move from a foreign key to a many-to-many.  If Pulp 3 is unlikely to ever use a many-to-many relationship then I think we should avoid premature complexity in that respect.

daviddavis: Yea, the idea was to have a `remotes` list field in the API but validate that the user is only passing in a single remote even though the field was a list. That way if we wanted to support multiple remotes later, we wouldn't have to change the API and break semantic versioning.

This was my original idea but I had second thoughts and thus changed this PR to only have a single `remote` field on repository in the API because having a `remotes` field but only allowing a single remote was awkward and confusing. Also, I don't think we're going to add support for syncing from multiple remotes anytime soon.

Anyway, interested if you or @bmbouter have thoughts on which approach you think is better.

daviddavis: > Idk, it's not difficult at all to move from a foreign key to a many-to-many. If Pulp 3 is unlikely to ever use a many-to-many relationship then I think we should avoid premature complexity in that respect.

Agreed.

CHANGES/7015.feature
```diff
@@ -0,0 +1,2 @@
+Added ability to associate Remotes to Repositories so users no longer have to specify Remote when
```
bmbouter, line 1:
```suggestion
Added ability to associate a Remote with a Repository so users no longer have to specify Remote when
```

How about this?

pulpcore/app/serializers/repository.py
```diff
@@ -147,12 +162,10 @@ class Meta:
...
 class RepositorySyncURLSerializer(serializers.Serializer):
     remote = DetailRelatedField(
-        required=True,
+        required=False,
         view_name_pattern=r"remotes(-.*/.*)-detail",
         queryset=models.Remote.objects.all(),
-        help_text=_("A URI of the repository to be synchronized."),
-        label=_("Remote"),
-        error_messages={"required": _("The remote URI must be specified.")},
+        help_text=_("A remote to sync from. Will override the remote set on repository."),
```
bmbouter, line 168:
```suggestion
        help_text=_("A remote to sync from. This will override a remote set on repository."),
```

PR #790, opened 2020-07-17
Remote certificate doc strings

pavelpicka:
PULP doesn't advice to escape newlines for ca_cert, client_cert and client_key fields.

closes: #6735
https://pulp.plan.io/issues/6735

### DIFF
--- a/CHANGES/plugin_api/6735.removal
+++ b/CHANGES/plugin_api/6735.removal
@@ -1,0 +1,1 @@
+Newlines in certificate string (ca_cert, client_cert, client_key) on Remotes are not required to be escaped.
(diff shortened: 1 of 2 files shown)


CHANGES/6735.removal
```diff
@@ -0,0 +1 @@
+Newlines in certificate string (ca_cert, client_cert, client_key) on Remote repositories are not required to escape.
```
bmbouter, line 1:
```suggestion
Newlines in certificate string (ca_cert, client_cert, client_key) on Remote repositories are not required to be
escaped.

```

CHANGES/6735.removal
```diff
@@ -0,0 +1 @@
+Newlines in certificate string (ca_cert, client_cert, client_key) on Remote repositories are not required to be escaped.
```
bmbouter, line 1: Sorry to ask for another comment on the same line. I'm thinking more about this, and really this is a plugin writer change because pulpcore itself doesn't have a remote. So can you do two things?

1) Can you rename this file to `CHANGES/plugin_api/6735.removal`. This will cause it to show up in the plugin writer changelog.

2) Also another slight revision:  `Newlines in certificate string (ca_cert, client_cert, client_key) on Remotes are not required to be escaped.`

PR #792, opened 2020-07-17
Streamed responses are properly returned to handler to be logged

### LINKED ISSUE #7002: Ruby bindings send PRNs as hrefs to Pulp API
**Version**
```
      "versions": {
        "deb": "3.7.0",
        "rpm": "3.32.0",
        "core": "3.85.1",
        "file": "3.85.1",
        "ostree": "2.5.0",
        "python": "3.19.0",
        "ansible": "0.28.0",
        "certguard": "3.85.1",
        "container": "2.26.1"
      }
```

**Describe the bug**
I noticed a couple API endpoints that reportedly should accept both hrefs and PRNs 404 when a PRN is sent:

1)

```
> <UploadResponse> update(content_range, upload_href, file, opts)

Upload a file chunk

Upload a chunk for an upload.
```

2) 
```
> <TaskResponse> read(task_href, opts)

Inspect a task

Error I captured:
 777f8c1a | ActionController::RoutingError (No route matches [GET] "/prn:core.task:01997816-58f6-71e2-9bd9-e49554e2404a"):
```

**To Reproduce**
Use the Ruby bindings to read a Task is the most straight-forward one.

Note, there may be other endpoints too. I've only identified 2. Perhaps the PRN/href mixin isn't working specifically with the Ruby bindings?

**Expected behavior**
PRN should be used where an HREF is expected.

**Additional context**
Katello is removing all use of HREFs to prepare for Pulp 4.

gerrod3:
When a request for content that was on_demand occured for the first time, no logs where present due to the response not properly being return to the content handler, despite the response actually occuring.

fixes #7002
https://pulp.plan.io/issues/7002

### DIFF
--- a/CHANGES/7002.bugfix
+++ b/CHANGES/7002.bugfix
@@ -1,0 +1,1 @@
+First time on demand content requests appear in the access log.
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -490,6 +490,7 @@ class Handler:
         for remote_artifact in content_artifact.remoteartifact_set.all():
             try:
                 response = await self._stream_remote_artifact(request, response, remote_artifact)
+                return response
 
             except ClientResponseError:
                 continue
@@ -643,4 +644,7 @@ class Handler:
         if remote.policy != Remote.STREAMED:
             self._save_artifact(download_result, remote_artifact)
         await response.write_eof()
+
+        if response.status == 404:
+            raise HTTPNotFound()
         return response


pulpcore/content/handler.py
```diff
@@ -494,6 +494,9 @@ async def _stream_content_artifact(self, request, response, content_artifact):
             except ClientResponseError:
                 continue
 
+        if response.status == 200:
```
dkliban, line 497: You should simply return the response here. The status check should happen in _stream_remote_artifact(). If the status is 404, you need to raise the HTTPNotFound in _stream_remote_artifact().

pulpcore/content/handler.py
```diff
@@ -643,4 +644,7 @@ async def finalize():
         if remote.policy != Remote.STREAMED:
             self._save_artifact(download_result, remote_artifact)
         await response.write_eof()
+
+        if response.status == 404:
+            return HTTPNotFound
```
dkliban, line 649: raise HTTPNotFound()

CHANGES/7002.bugfix
```diff
@@ -0,0 +1 @@
+A StreamResponse is now returned in _stream_content_artifact() if the response is 200 OK allowing it to be logged.
```
dkliban, line 1: This should be a user facing message. Something along the line of  "First time on demand content requests appear in the access log."

PR #793, opened 2020-07-17
Introducing PulpTemporaryFile

fao89:
https://pulp.plan.io/issues/6749
closes #6749


pulpcore/app/models/content.py
```diff
@@ -241,6 +245,122 @@ def init_and_validate(file, expected_digests=None, expected_size=None):
...
+
+        Args:
+            args (list): list of positional arguments for Model.delete()
+            kwargs (dict): dictionary of keyword arguments to pass to Model.delete()
+        """
+        super().delete(*args, **kwargs)
+        self.file.delete(save=False)
+
+    @staticmethod
+    def init_and_validate(file, expected_digests=None, expected_size=None):
```
daviddavis, line 297: Is this the same as the one in Artifact? If so, what about making it into a mixin?

daviddavis: Actually, I see that it'll differ.

pulpcore/app/models/content.py
```diff
@@ -241,6 +245,122 @@ def init_and_validate(file, expected_digests=None, expected_size=None):
...
+            for algorithm, expected_digest in expected_digests.items():
+                if expected_digest != hashers[algorithm].hexdigest():
+                    raise DigestValidationError()
+        else:
+            expected_digests = {"sha256": hashers["sha256"].hexdigest()}
+
+        if Artifact.objects.filter(**expected_digests).count():
+            raise ValidationError(_("Artifact already exists."))
+
+        return PulpTemporaryFile(file=file)
```
fao89, line 350: @daviddavis this part is different than the `Artifact` one

pulpcore/app/models/content.py
```diff
@@ -241,6 +245,122 @@ def init_and_validate(file, expected_digests=None, expected_size=None):
...
+
+        if expected_digests:
+            for algorithm, expected_digest in expected_digests.items():
+                if expected_digest != hashers[algorithm].hexdigest():
+                    raise DigestValidationError()
+        else:
+            expected_digests = {"sha256": hashers["sha256"].hexdigest()}
+
+        if Artifact.objects.filter(**expected_digests).count():
+            raise ValidationError(_("Artifact already exists."))
```
daviddavis, line 348: I think that temporary files can be created even if there are Artifacts that already exist.

fao89: I did this because of this test:
https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/tests/functional/api/collection/v2/test_upload.py#L51-L52
I can update the test

pulpcore/app/models/storage.py
```diff
@@ -110,6 +110,21 @@ def get_artifact_path(sha256digest):
...
+
+    Args:
+        pulp_id (uuid): pulp_id the file for the PulpTemporaryFile
+
+    Returns:
+        A string representing the absolute path where a file backing the PulpTemporaryFile should be
+        stored
+    """
+    pulp_id = str(pulp_id)
+    return os.path.join("files", pulp_id[0:2], pulp_id[2:])
```
daviddavis, line 125: I think this should be `tmp/files/...`

pulpcore/app/models/content.py
daviddavis: No, I think the test makes sense but that we should probably leave it up to plugins to check whether an Artifact already exists before creating a PulpTemporaryFile. The thought being that some plugins might want to still create a PulpTemporaryFile in some cases even if the Artifact exists. But maybe this is not a valid use case.

cc @bmbouter

pulpcore/app/models/content.py
```diff
@@ -241,6 +245,122 @@ def init_and_validate(file, expected_digests=None, expected_size=None):
...
+
+        return PulpTemporaryFile(file=file)
+
+    def to_artifact(self):
+        """Creates an Artifact from PulpTemporaryFile."""
```
daviddavis, line 353: This docstring is missing a `Returns` section.

CHANGES/6749.feature
```diff
@@ -0,0 +1 @@
+`PulpTemporaryFile` for handling temporary files between the viewset and triggered tasks
```
daviddavis, line 1: This needs to be a plugin changelog entry.

daviddavis:
Looks good. Don't forget to add some plugin writer docs.

CHANGES/6749.feature
bmbouter: Agreed, and indicate it was added

pulpcore/app/models/content.py
```diff
@@ -241,6 +245,122 @@ def init_and_validate(file, expected_digests=None, expected_size=None):
         return Artifact(**attributes)
 
 
+class PulpTemporaryFile(BaseModel):
+    """
+    A temporary file.
```
bmbouter, line 250: Can this be more descriptive but still one like, e.g. `A temporary filed saved to the storage backend. Commonly used to pass data to one or more tasks.`

pulpcore/app/models/content.py
bmbouter, line 297: I didn't imagine we would have this method at all. To me, plugin writers will save a file, use the file, delete the file. If a user case comes up later we could add something then. @fao89 and @daviddavis what do you both think?

pulpcore/app/models/content.py
```diff
@@ -241,6 +245,122 @@ def init_and_validate(file, expected_digests=None, expected_size=None):
...
+
+        Args:
+            name (str): Original name of uploaded file. It is ignored by this method because the
+                pulp_id is used to determine a file path instead.
+        """
+        return storage.get_temp_file_path(self.pulp_id)
+
+    file = fields.ArtifactFileField(null=False, upload_to=storage_path, max_length=255)
+
+    def save(self, *args, **kwargs):
```
bmbouter, line 270: Are the `save` and `delete` both the same as Artifact? If so maybe we can make these two methods an ancestor class to both?

pulpcore/app/models/content.py
daviddavis: That's fine but I think the import code is using this to check the checksum digest. We'll have to move that code into pulp_ansible.

pulpcore/app/models/content.py
fao89: they are the same

pulpcore/app/models/content.py
fao89: I kept it because it validates the checksum, and I like to raise failures the earliest possible

pulpcore/app/models/content.py
```diff
@@ -241,6 +245,122 @@ def init_and_validate(file, expected_digests=None, expected_size=None):
...
+                    raise DigestValidationError()
+        else:
+            expected_digests = {"sha256": hashers["sha256"].hexdigest()}
+
+        if Artifact.objects.filter(**expected_digests).count():
+            raise ValidationError(_("Artifact already exists."))
+
+        return PulpTemporaryFile(file=file)
+
+    def to_artifact(self):
```
bmbouter, line 352: Do we need this method? Maybe we do I'm not sure. I was thinking users would pass their file handle directly in as `PulpTemporaryFile(file=my_file).save()`. What would happen if we deleted this, and `init_and_validate` and then called it with `PulpTemporaryFile(file=my_file).save()`?

daviddavis: It seems pretty handy to have a method that converts a PulpTemporaryFile into an Artifact.

fao89: it would fail, as we cannot save the same `Artifact` twice, it would happen with or without this method. At some point, you want to turn the `PulpTemporaryFile` into an `Artifact`, and if it already exists, we will get an `IntegrityError`.

pulpcore/app/models/content.py
bmbouter: +1 to moving them to an ancestor class that is shared by both then. Is that possible?

pulpcore/app/models/content.py
bmbouter: OK that's understandable. The internals need to change then though because it shouldn't be using `Artifact` in any places.

bmbouter:
Can a new docs section be added indicating to plugin writers that this is available for passing file data to one or more tasks with some examples? I think it would go in a new subsection [here](https://docs.pulpproject.org/plugins/plugin-writer/concepts/index.html#tasks). I think that would be named `Making Temporary Files Available to Tasks`.

pulpcore/app/models/content.py
fao89: `init_and_validate` is validating `checkums` and `size`, so it would fail at some point when turning `PulpTemporaryFile` into `Artifact`, I'm trying to raise the failures earlier.
`to_artifact` is just a convenient method for not forgetting to call `delete` when you move `PulpTemporaryFile` to `Artifact`.

pulpcore/app/models/storage.py
bmbouter: agreed.

pulpcore/app/models/content.py
bmbouter: Yup I didn't understand it; I thought it was part of the internals. I see how it's used in the pulp_ansible PR now, and in seeing the usage I agree.

pulpcore/app/models/content.py
daviddavis: Are these methods needed for PulpTemporaryFile? I'm guessing they were needed for some special cases in Artifact.

pulpcore/app/models/storage.py
```diff
@@ -110,6 +110,21 @@ def get_artifact_path(sha256digest):
...
+    Determine the absolute path where a file backing the PulpTemporaryFile should be stored.
+
+    Args:
+        pulp_id (uuid): pulp_id the file for the PulpTemporaryFile
+
+    Returns:
+        A string representing the absolute path where a file backing the PulpTemporaryFile should be
+        stored
+    """
+    pulp_id = str(pulp_id)
```
bmbouter, line 124: nitpick: Generally I try to not modify a variable and reassign it. So maybe call this var `pulp_id_str = `.

fao89:
> Can a new docs section be added indicating to plugin writers that this is available for passing file data to one or more tasks with some examples? I think it would go in a new subsection [here](https://docs.pulpproject.org/plugins/plugin-writer/concepts/index.html#tasks). I think that would be named `Making Temporary Files Available to Tasks`.

I have no idea how to explain the need for this new model

mdellweg:
Is this temporary file using the django storage framework (and i think it should, because there are usecases without any other shared storage.)?

pulpcore/app/models/content.py
```diff
@@ -159,47 +195,102 @@ def is_equal(self, other):
...
+        attributes = {"size": size, "file": file}
+        for algorithm in Artifact.DIGEST_FIELDS:
+            attributes[algorithm] = hashers[algorithm].hexdigest()
+
+        return Artifact(**attributes)
+
+
+class PulpTemporaryFile(HandleTempFilesMixin, BaseModel):
+    """
+    A temporary filed saved to the storage backend.
```
alikins, line 256:
```suggestion
    A temporary file saved to the storage backend.
```

pulpcore/app/models/content.py
alikins: Suppose it could be a @classmethod on Artifact, a `from_pulp_temporary_file(temp_file)`. Though the code is about equal parts Artifact and PulpTemporaryFile, so doesn't really matter.

pulpcore/app/models/content.py
fao89: thank you!

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -48,6 +48,16 @@ Any action that can run for a long time should be an asynchronous task. Plugin w
 to understand the internals of the pulpcore tasking system, workers automatically execute tasks from
 RQ, including tasks deployed by plugins.
 
+
+**Making Temporary Files Available to Tasks**
+
+Sometimes, files must be brought forward from the viewset to the executing task, that are not meant
```
daviddavis, line 54:
```suggestion
Sometimes, files must be brought forward from the viewset to the executing task, that may or may not
```

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -48,6 +48,16 @@ Any action that can run for a long time should be an asynchronous task. Plugin w
...
+to end up being artifacts in the end. On the other hand, different pulp services are not guaranteed
```
daviddavis, line 55:
```suggestion
end up being artifacts in the end. On the other hand, different pulp services are not guaranteed
```

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -48,6 +48,16 @@ Any action that can run for a long time should be an asynchronous task. Plugin w
...
+to share a common filesystem (like /usr/share/pulp).
+:term:`PulpTemporaryFile<PulpTemporaryFile>` is the alternative for creating  files with the same
+storage technology that the artifacts use.
```
daviddavis, line 58: This is great. Can you add a small code example of instantiating a PulpTemporaryFile, saving it, and then turning it into an Artifact?

pulpcore/app/models/content.py
daviddavis: This still needs to be addressed. @bmbouter any thoughts?

bmbouter: Yes I expect a new `PulpTemporaryFile` would still be created without error even there is a corresponding Artifact that already exists also.

pulpcore/app/models/content.py
bmbouter: @alikins I was thinking the same thing.

@fao89 am I reading correctly that the `Artifact.init_and_validate` now accepts a `PulpTemporaryFile` also? Should we just use that and remove this one? What do you think should be done?

fao89: `Artifact` is not accepting `PulpTemporaryFile` yet, this is alikins suggestion - I can implement it

pulpcore/app/models/content.py
fao89: so what should I do with this test?
https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/tests/functional/api/collection/v2/test_upload.py#L51-L52

pulpcore/app/models/content.py
bmbouter: Oh I thought it did from reading this docstring https://github.com/pulp/pulpcore/pull/793/files but I see now in the code it doesn't. +1 to having it take a `PulpTemporaryFile` also just to make the existing interface even more usable.

pulpcore/app/models/content.py
bmbouter: I think that test should receive an error on the task later when it realizes the Artifact sha256 already exists. So it would monitor the task and see "oh the task failed w/ a new error message". I think it will fail with a duplicate key on the sha256 when the Artifact.save() is called.

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -48,6 +48,29 @@ Any action that can run for a long time should be an asynchronous task. Plugin w
...
+
+Sometimes, files must be brought forward from the viewset to the executing task, that may or may
+not end up being artifacts in the end. On the other hand, different pulp services are not
+guaranteed to share a common filesystem (like /usr/share/pulp).
+:term:`PulpTemporaryFile<PulpTemporaryFile>` is the alternative for creating  files with the same
+storage technology that the artifacts use.
+
+.. code-block:: python
+
+    # Example 1 - Saving a temporary file:
```
mdellweg, line 62: should we suggest that this is usually in the viewset?

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -48,6 +48,29 @@ Any action that can run for a long time should be an asynchronous task. Plugin w
...
+    temp_file = PulpTemporaryFile(file=my_file).save()
+
+    # Example 2 - Validating the digest and saving:
+    temp_file = PulpTemporaryFile.init_and_validate(
+        my_file, expected_digests={'md5': '912ec803b2ce49e4a541068d495ab570'}
+    )
```
mdellweg, line 68: Does this still need `.save()`?

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -48,6 +48,29 @@ Any action that can run for a long time should be an asynchronous task. Plugin w
...
+
+    # Example 3 - From PulpTemporaryFile to Artifact:
+    artifact = Artifact.from_pulp_temporary_file(temp_file)
```
mdellweg, line 71: This will typically happen in the task context, right?
Can we add an else branch to show that, alternatively, the task must clean up the temp_file?
Or should it even be try-except?

pulpcore/app/models/content.py
```diff
@@ -75,7 +77,39 @@ def q(self):
...
+        Args:
+            args (list): list of positional arguments for Model.save()
+            kwargs (dict): dictionary of keyword arguments to pass to Model.save()
+        """
+        try:
+            super().save(*args, **kwargs)
+            self.file.close()
+        except Exception:
+            self.file.close()
+            raise
```
mdellweg, line 98:
```suggestion
        try:
            super().save(*args, **kwargs)
        finally:
            self.file.close()
```

pulpcore/app/models/content.py
```diff
@@ -240,6 +248,100 @@ def init_and_validate(file, expected_digests=None, expected_size=None):
...
+                values don't match the digest of the data
+            :class:`~pulpcore.exceptions.SizeValidationError`: When the ``expected_size`` value
+                doesn't match the size of the data
+
+        Returns:
+            An in-memory, unsaved :class:`~pulpcore.plugin.models.PulpTemporaryFile`
+        """
+        if isinstance(file, str):
+            with open(file, "rb") as f:
+                hashers = {n: hashlib.new(n) for n in Artifact.DIGEST_FIELDS}
```
mdellweg, line 321:
```suggestion
                hashers = {n: hashlib.new(n) for n in expected_digests.keys()}
```

pulpcore/app/models/content.py
```diff
@@ -240,6 +248,100 @@ def init_and_validate(file, expected_digests=None, expected_size=None):
...
+                size = 0
+                while True:
+                    chunk = f.read(1048576)  # 1 megabyte
+                    if not chunk:
+                        break
```
mdellweg, line 326:
```suggestion
                for chunk in iter(lambda: f.read(1048576), b""):
```

pulpcore/app/models/content.py
```diff
@@ -240,6 +248,100 @@ def init_and_validate(file, expected_digests=None, expected_size=None):
...
+        Args:
+            name (str): Original name of uploaded file. It is ignored by this method because the
+                pulp_id is used to determine a file path instead.
+        """
+        return storage.get_temp_file_path(self.pulp_id)
+
+    file = fields.ArtifactFileField(null=False, upload_to=storage_path, max_length=255)
+
+    @staticmethod
+    def init_and_validate(file, expected_digests=None, expected_size=None, validate_artifact=False):
```
mdellweg, line 294: I am wondering whether this need to be it's own function, or if we can fold these optional validations in the `__init__` method.

docs/glossary.rst
```diff
@@ -7,6 +7,10 @@ Glossary
         A file. They usually belong to a :term:`content unit<Content>` but may be used
         elsewhere (e.g. for PublishedArtifacts).
 
+    :class:`~pulpcore.app.models.PulpTemporaryFile`
+        A temporary filed saved to the storage backend. Commonly used to pass data to one or more
```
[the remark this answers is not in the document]
fao89: thank you!

docs/plugins/plugin-writer/concepts/index.rst
fao89: I don't know how can I express it, any suggestion?

docs/plugins/plugin-writer/concepts/index.rst
fao89: yep, I forgot it, thanks!

pulpcore/app/models/content.py
fao89: my idea was to mimic `Artifact` method

docs/plugins/plugin-writer/concepts/index.rst
fao89: great suggestion! I put a try-except

pulpcore/app/models/content.py
fao89: I think the lambda is less readable

docs/plugins/plugin-writer/concepts/index.rst
mdellweg: Maybe like
```
# Example:
# (in the view)
# variant 1 - Saving a temporary file:
...
# variant 2 - --"-- with additional validation:
...
# (in the task)
# Using the temporary file to create an artifact:
...
```

docs/glossary.rst
```diff
@@ -7,6 +7,10 @@ Glossary
         A file. They usually belong to a :term:`content unit<Content>` but may be used
         elsewhere (e.g. for PublishedArtifacts).
 
+    :class:`~pulpcore.app.models.PulpTemporaryFile`
```
bmbouter, line 10: I actually think this glossary is for end-user facing terms and I don't believe they would know about this. So my suggestion is to remove this from the glossary.

pulpcore/app/models/content.py
mdellweg: What i see is that if both of the `expected_*` parameters are omitted from the call this function is really the same as just instanciating a `PulpTemporaryFile` the old fashioned way.
Also if there is no expected digests, we are consuming reading the whole file for no additional benefit. Can you add a `if expected_digests:` there?

pulpcore/app/models/content.py
```diff
@@ -240,6 +246,102 @@ def init_and_validate(file, expected_digests=None, expected_size=None):
...
+        """
+        Callable used by FileField to determine where the uploaded file should be stored.
+
+        Args:
+            name (str): Original name of uploaded file. It is ignored by this method because the
+                pulp_id is used to determine a file path instead.
+        """
+        return storage.get_temp_file_path(self.pulp_id)
+
+    file = fields.ArtifactFileField(null=False, upload_to=storage_path, max_length=255)
```
bmbouter, line 289: This being `ArtifactFileField` instead of `FileField` will cause [this extra functionality](https://github.com/pulp/pulpcore/blob/b94abd64d76ea4554e6750ff38ce458eaa888cc8/pulpcore/app/models/fields.py#L44) to run. Is that important or is `FileField` right?

fao89: From the docstring I guess we don't need `ArtifactFileField`

PR #794, opened 2020-07-19
Taught export to produce, and import to understand, a table-of-conten…

ggainey:
…ts (toc) file.

Emitted 'next to' the export file or files, named <export-file>-toc.json.

Consists of keys "meta" and "files". "files" is a dictionary of export-file/checksums.
"meta" contains the "file", "chunk_size", and "global_hash" of the export.

Added toc= to import. Import will find and validate the checksums of any chunk_files,
reassemble them into a single .tar.gz, and pass that along to the rest of the
import process. Deletes chunks as it goes, to minimize disk-space.

Updated import-export docs to describe TOC file and its use.

closes #6737

### DIFF
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -1,6 +1,8 @@
+import hashlib
 import json
 import os
 import re
+import subprocess
 import tempfile
 import tarfile
 from gettext import gettext as _
@@ -148,7 +150,7 @@ def import_repository_version(destination_repo_pk, source_repo_name, tar_path):
             new_version.set_content(content)
 
 
-def pulp_import(importer_pk, path):
+def pulp_import(importer_pk, path, toc):
     """
     Import a Pulp export into Pulp.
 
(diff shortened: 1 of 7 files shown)


pulpcore/app/tasks/importer.py
```diff
@@ -165,6 +175,124 @@ def destination_repo(source_repo_name):
...
         return Repository.objects.get(name=dest_repo_name)
 
+    def validate_toc(toc_filename):
+        """
+        Check validity of table-of-contents file.
+
+        table-of-contents must:
+          * exist
+          * be valid JSON
+          * point to chunked-export-files that exist 'next to' the 'toc' file
```
daviddavis, line 185: Doesn't this also check the checksums of the file?

ggainey: It does indeed, thanks

ggainey: Updated docstrings and repushed!

PR #795, opened 2020-07-20
 Enable plugin writers to inherit from ImmutableRepositoryViewSet

lubosmj:
Required by https://github.com/pulp/pulp_container/pull/127.

closes #7191

### DIFF
--- a/CHANGES/plugin_api/7191.feature
+++ b/CHANGES/plugin_api/7191.feature
@@ -1,0 +1,1 @@
+Enabled plugin writers to create immutable repository ViewSets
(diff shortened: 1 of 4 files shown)


lubosmj:
Since there is a requirement to access a few additional classes within a plugin, I opened this pull request. We may discuss the way how we allow users to access these specific classes; for example, whether we want to have `pulpcore.plugin.viewsets.AsyncRemoveMixin` or `pulpcore.plugin.viewsets.mixin_ops.AsyncRemoveMixin`.

ipanova:
@lubosmj technically this is a plugin api change. can you open an issue + add a changelog entry please?

PR #797, opened 2020-07-21
Set a default DJANGO_SETTINGS_MODULE env var in content app

ekohl:
This aligns the content service to the WSGI service (API) which already sets a default env var for DJANGO_SETTINGS_MODULE. By providing a default, the systemd service doesn't have to duplicate this and makes configuration easier.

fixes #7179

### DIFF
--- a/CHANGES/7179.bugfix
+++ b/CHANGES/7179.bugfix
@@ -1,0 +1,1 @@
+Set a default DJANGO_SETTINGS_MODULE env var in content app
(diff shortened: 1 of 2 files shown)


ekohl:
Note this mirrors wsgi.py here:
https://github.com/pulp/pulpcore/blob/09d05da6f38c74e3574d3d256f890b23d76cb3d0/pulpcore/app/wsgi.py#L14

ekohl:
Updated from misc to bugfix.

PR #798, opened 2020-07-21
Updates template config's branch for pulp_file

dkliban:
[noissue]

### DIFF
--- a/.travis/Containerfile.j2
+++ b/.travis/Containerfile.j2
@@ -1,4 +1,4 @@
-FROM pulp/pulp-ci:latest
+FROM {{ ci_base | default("pulp/pulp-ci:latest") }}
 
 {% if s3_test | default(false) %}
 # Hacking botocore (https://github.com/boto/botocore/pull/1990)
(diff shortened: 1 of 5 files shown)


daviddavis:
Here is the error:

```
File "/usr/local/lib/python3.7/site-packages/pulp_file/app/modelresource.py", line 1, in <module>
    from pulpcore.plugin.importexport import BaseContentResource
ImportError: cannot import name 'BaseContentResource' from 'pulpcore.plugin.importexport' (/usr/local/lib/python3.7/site-packages/pulpcore/plugin/importexport.py)
```

Here is the line that it's referring to in master:

https://github.com/pulp/pulp_file/blob/master/pulp_file/app/modelresource.py#L1

And 1.1: 

https://github.com/pulp/pulp_file/blob/1.1/pulp_file/app/modelresource.py#L1

So the CI is still installing master somehow.

mdellweg:
Right, i think, you need to reapply the plugin template to really perform the change.

PR #799, opened 2020-07-21
Change the default deployment layout

ekohl:
This changes the default deployment layout. The main change is that MEDIA_ROOT gets its own directory. This allows limiting the file permissions in a shared Pulp 2 + Pulp 3 deployment and the SELinux file contexts. Another benefit is compatibility with django_extensions' unreferenced_files command which lists all files in MEDIA_ROOT that are not in the database.

Other paths are kept on the same absolute paths, but the diff looks bigger because they used derive from MEDIA_ROOT.

The documentation is updated to show the latest best practices.

fixes #7178

### DIFF
--- a/CHANGES/7178.feature
+++ b/CHANGES/7178.feature
@@ -1,0 +1,9 @@
+Change the default deployment layout
+
+This changes the default deployment layout. The main change is that MEDIA_ROOT gets its own
+directory. This allows limiting the file permissions in a shared Pulp 2 + Pulp 3 deployment and the
+SELinux file contexts. Another benefit is compatibility with django_extensions' unreferenced_files
+command which lists all files in MEDIA_ROOT that are not in the database.
+
+Other paths are kept on the same absolute paths. The documentation is updated to show the latest
+best practices.
--- a/pulpcore/app/checks.py
+++ b/pulpcore/app/checks.py
@@ -1,0 +1,43 @@
+from pathlib import Path
+
+from django.conf import settings
+from django.core.checks import Warning as CheckWarning, register
+
+
+@register(deploy=True)
+def storage_paths(app_configs, **kwargs):
+    warnings = []
+
+    if settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem":
+        try:
+            media_root_dev = Path(settings.MEDIA_ROOT).stat().st_dev
+        except OSError:
+            media_root_dev = None
+            warnings.append(
+                CheckWarning(
+                    "Your MEDIA_ROOT setting points to a path that does not exist.",
+                    id="pulpcore.W001",
+                )
+            )
+
+        try:
+            upload_temp_dir_dev = Path(settings.FILE_UPLOAD_TEMP_DIR).stat().st_dev
+        except OSError:
+            upload_temp_dir_dev = None
+            warnings.append(
+                CheckWarning(
+                    "Your FILE_UPLOAD_TEMP_DIR setting points to a path that does not exist.",
+                    id="pulpcore.W002",
+                )
+            )
+
+        if media_root_dev and media_root_dev != upload_temp_dir_dev:
+            warnings.append(
+                CheckWarning(
+                    "MEDIA_ROOT and FILE_UPLOAD_TEMP_DIR are on different filesystems. "
+                    "It is highly recommended that these live on the same filesystem",
+                    id="pulpcore.W003",
+                )
+            )
+
+    return warnings
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -8,12 +8,12 @@ For the full list of settings and their values, see
 https://docs.djangoproject.com/en/1.11/ref/settings/
 """
 
-import os
 import sys
 
 from contextlib import suppress
 from gettext import gettext as _
 from importlib import import_module
+from pathlib import Path
 from pkg_resources import iter_entry_points
 
 from django.core.exceptions import ImproperlyConfigured
@@ -21,8 +21,8 @@ from django.db import connection
 
 from pulpcore import constants
 
-# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
-BASE_DIR = os.path.dirname(os.path.abspath(__file__))
+# Build paths inside the project like this: BASE_DIR / ...
+BASE_DIR = Path(__file__).absolute().parent
 
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/1.11/howto/deployment/checklist/
@@ -32,7 +32,8 @@ DEBUG = False
 
 ALLOWED_HOSTS = ["*"]
 
-MEDIA_ROOT = "/var/lib/pulp/"
+DEPLOY_ROOT = Path("/var/lib/pulp")
+MEDIA_ROOT = str(DEPLOY_ROOT / "media")  # Django 3.1 adds support for pathlib.Path
 
 ADMIN_SITE_URL = "admin/"
 
@@ -40,12 +41,12 @@ ADMIN_SITE_URL = "admin/"
 # https://docs.djangoproject.com/en/1.11/howto/static-files/
 
 STATIC_URL = "/assets/"
-STATIC_ROOT = os.path.join(MEDIA_ROOT, STATIC_URL.lstrip("/"))
+STATIC_ROOT = DEPLOY_ROOT / STATIC_URL.strip("/")
 
 DEFAULT_FILE_STORAGE = "pulpcore.app.models.storage.FileSystem"
 
-FILE_UPLOAD_TEMP_DIR = os.path.join(MEDIA_ROOT, "tmp/")
-WORKING_DIRECTORY = os.path.join(MEDIA_ROOT, "tmp/")
+FILE_UPLOAD_TEMP_DIR = DEPLOY_ROOT / "tmp"
+WORKING_DIRECTORY = FILE_UPLOAD_TEMP_DIR
 CHUNKED_UPLOAD_DIR = "upload"
 
 # List of upload handler classes to be applied in order.
@@ -118,7 +119,7 @@ ROOT_URLCONF = "pulpcore.app.urls"
 TEMPLATES = [
     {
         "BACKEND": "django.template.backends.django.DjangoTemplates",
-        "DIRS": [os.path.join(BASE_DIR, "templates")],
+        "DIRS": [BASE_DIR / "templates"],
         "APP_DIRS": True,
         "OPTIONS": {
             "context_processors": [
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -55,7 +55,7 @@ DEFAULT_FILE_STORAGE
 MEDIA_ROOT
 ^^^^^^^^^^
 
-   The location where Pulp will store files. By default this is `/var/lib/pulp/`.
+   The location where Pulp will store files. By default this is `/var/lib/pulp/media`.
 
    If you're using S3, point this to the path in your bucket you want to save files. See the
    :ref:`storage documentation <storage>` for more info.
@@ -65,7 +65,7 @@ MEDIA_ROOT
    * mode: 750
    * owner: pulp (the account that pulp runs under)
    * group: pulp (the group of the account that pulp runs under)
-   * SELinux context: system_u:object_r:var_lib_t:s0
+   * SELinux context: system_u:object_r:pulpcore_var_lib_t:s0
 
 LOGGING
 ^^^^^^^
@@ -138,10 +138,10 @@ WORKING_DIRECTORY
 
    It should have permissions of:
 
-   * mode: 755
+   * mode: 750
    * owner: pulp (the account that pulp runs under)
    * group: pulp (the group of the account that pulp runs under)
-   * SELinux context: unconfined_u:object_r:var_lib_t:s0
+   * SELinux context: system_u:object_r:pulpcore_var_lib_t:s0
 
 .. note::
 
--- a/pulpcore/app/apps.py
+++ b/pulpcore/app/apps.py
@@ -185,6 +185,10 @@ class PulpAppConfig(PulpPluginAppConfig):
     # The version of this app
     version = "3.10.0.dev"
 
+    def ready(self):
+        super().ready()
+        from . import checks  # noqa
+
 
 def _populate_access_policies(sender, **kwargs):
     print(f"Initialize missing access policies for {sender.label}.")
--- a/pulpcore/app/models/storage.py
+++ b/pulpcore/app/models/storage.py
@@ -98,7 +98,7 @@ class FileSystem(FileSystemStorage):
 
 def get_artifact_path(sha256digest):
     """
-    Determine the absolute path where a file backing the Artifact should be stored.
+    Determine the relative path where a file backing the Artifact should be stored.
 
     Args:
         sha256digest (str): sha256 digest of the file for the Artifact
@@ -112,14 +112,14 @@ def get_artifact_path(sha256digest):
 
 def get_temp_file_path(pulp_id):
     """
-    Determine the absolute path where a file backing the PulpTemporaryFile should be stored.
+    Determine the relative path where a file backing the PulpTemporaryFile should be stored.
 
     Args:
         pulp_id (uuid): An identifier identifying the file for the PulpTemporaryFile
 
     Returns:
-        A string representing the absolute path where a file backing the PulpTemporaryFile should be
-        stored
+        A string representing the path where a file backing the PulpTemporaryFile should be
+        stored. This path is relative to the storage.
     """
     pulp_id_str = str(pulp_id)
     return os.path.join("tmp/files", pulp_id_str[:2], pulp_id_str[2:])
--- a/pulpcore/tests/functional/api/test_crd_artifacts.py
+++ b/pulpcore/tests/functional/api/test_crd_artifacts.py
@@ -7,7 +7,7 @@ import unittest
 
 from pulp_smash import api, cli, config, utils
 from pulp_smash.exceptions import CalledProcessError
-from pulp_smash.pulp3.constants import ARTIFACTS_PATH, MEDIA_PATH
+from pulp_smash.pulp3.constants import ARTIFACTS_PATH
 from pulp_smash.pulp3.utils import delete_orphans
 from requests.exceptions import HTTPError
 
@@ -166,6 +166,7 @@ class ArtifactsDeleteFileSystemTestCase(unittest.TestCase):
         storage = utils.get_pulp_setting(cli_client, "DEFAULT_FILE_STORAGE")
         if storage != "pulpcore.app.models.storage.FileSystem":
             self.skipTest("this test only works for filesystem storage")
+        media_root = utils.get_pulp_setting(cli_client, "MEDIA_ROOT")
 
         api_client = api.Client(cfg, api.json_handler)
 
@@ -173,7 +174,7 @@ class ArtifactsDeleteFileSystemTestCase(unittest.TestCase):
         files = {"file": utils.http_get(FILE_URL)}
         artifact = api_client.post(ARTIFACTS_PATH, files=files)
         self.addCleanup(api_client.delete, artifact["pulp_href"])
-        cmd = ("ls", os.path.join(MEDIA_PATH, artifact["file"]))
+        cmd = ("ls", os.path.join(media_root, artifact["file"]))
         cli_client.run(cmd, sudo=True)
 
         # delete
--- a/pulpcore/tests/functional/api/using_plugin/test_orphans.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_orphans.py
@@ -7,7 +7,6 @@ from random import choice
 from pulp_smash import cli, config, utils
 from pulp_smash.exceptions import CalledProcessError
 from pulp_smash.pulp3.bindings import monitor_task
-from pulp_smash.pulp3.constants import MEDIA_PATH
 from pulp_smash.pulp3.utils import (
     delete_orphans,
     delete_version,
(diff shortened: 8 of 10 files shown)


ekohl:
This is the result of playing around with it and previous experience with another Django project.

@bmbouter @mikedep333 please have a look

@jlsherrill @ehelms @wbclark you may also find this interesting.

If this is agreed on, it'll also need to be reflected in pulpcore-selinux. I don't mind making that patch, but first there should be agreement.

mikedep333:
@ekohl I'm probably not understanding this correctly, but I disagree with part of your design. It seems to contradict previous designs, including security (group access is good. World access is insecure.)

https://pulp.plan.io/projects/pulp/activity?from=2020-07-21

pulpcore/app/settings.py
```diff
@@ -28,19 +27,20 @@
 
 ALLOWED_HOSTS = ["*"]
 
-MEDIA_ROOT = "/var/lib/pulp/"
+DEPLOY_ROOT = Path("/var/lib/pulp")
+MEDIA_ROOT = DEPLOY_ROOT / "media"
```
ekohl, line 31: Offline discussion was to change this to `artifact` (which already exists) and then add a migration that updates the database.
```suggestion
MEDIA_ROOT = DEPLOY_ROOT / "artifact"
```

That means the following code needs updates:
https://github.com/pulp/pulpcore/blob/09d05da6f38c74e3574d3d256f890b23d76cb3d0/pulpcore/app/models/fields.py#L47-L66

Before proceeding with this, I'd like to get some more feedback if this is really desired because it looks like the entire `ArtifactFileField.pre_save` becomes redundant then.

Another approach would be to use a migration that looks at all models with a `FileField`. Then if it doesn't exist, look a directory higher and move it into `MEDIA_ROOT` https://github.com/django-extensions/django-extensions/blob/d747e09885e90d4c5f702f0bb4f0ea42e63f4eb0/django_extensions/management/commands/unreferenced_files.py#L27-L43 can be the basis of that.

ekohl:
Updated to at least use `pathlib` instead of `path.py` correctly so the tests show if I broke something.

daviddavis:
Moving to WIP due to inactivity.

ekohl:
For what it's worth: I was waiting on a reply from maintainers on how to proceed @ https://github.com/pulp/pulpcore/pull/799#discussion_r458870543

ekohl:
Summarizing a discussion with @mikedep333 and @dkliban. Keep me honest if I missed anything

The easiest migration plan is to keep the `artifact` directory inside `MEDIA_ROOT`. That means no application code change is needed and no DB migration is needed. Both are a good thing. Exact paths inside `MEDIA_ROOT` are not important to the admin, but a quick migration with minimal to no downtime is.

To support this, the following steps should be taken:

* Upgrade pulpcore-selinux to have the correct file contexts for the new locations
* Prepare layout (see script below)
* Update settings
* Restart all services
* Remove symlink

The migration script is as follows:

```python
#!/usr/bin/env python3
from pathlib import Path

DEPLOY_ROOT = Path('/var/lib/pulp')
LEGACY_DIR = DEPLOY_ROOT / 'artifact'
NEW_MEDIA_ROOT = DEPLOY_ROOT / 'media'
DESTINATION = NEW_MEDIA_ROOT / LEGACY_DIR.name

if LEGACY_DIR.is_dir() and not LEGACY_DIR.is_symlink():
    if DESTINATION.is_dir():
        raise SystemExit(f'Target {DESTINATION} already exists')

    NEW_MEDIA_ROOT.mkdir(exist_ok=True)
    LEGACY_DIR.rename(DESTINATION)
    LEGACY_DIR.symlink_to(DESTINATION)
```

This means that the files are available at both locations at the same time. There is a tiny window between the `rename` and `symlink_to` calls where content can't be found, but on a modern storage system that should be negligible.

After all services use the new settings, nothing should be using the symlink anymore and can be cleaned up.

mikedep333:
The installer team reviewed this, and we are good with the design.

ekohl:
Excellent, I'll try to find time to get Travis green on this PR.

CHANGES/7178.feature
```diff
@@ -0,0 +1,5 @@
+Change the default deployment layout
+
+This changes the default deployment layout. The main change is that MEDIA_ROOT gets its own directory. This allows limiting the file permissions in a shared Pulp 2 + Pulp 3 deployment and the SELinux file contexts. Another benefit is compatibility with django_extensions' unreferenced_files command which lists all files in MEDIA_ROOT that are not in the database.
```
bmbouter, line 3: Can this be line-wrapped so each line is < 100 chars. (pulp style guide requirement)

mikedep333:
To merge this, we'll also need a corresponding PR for pulp_installer that does the symlink.

ekohl:
Ah, it turns out Django's storage system only gained support for `pathlib.Path` in Django 3.1. That's disappointing. Casting that to a `str`.

ekohl:
Ok, the current failures are in pulp_smash because it's still looking in the old locations.

ekohl:
Rebased to resolve a conflict. Haven't looked into pulp_smash yet.

At this point I've written the Foreman installer side ([config](https://github.com/theforeman/puppet-pulpcore/pull/115) and [migration](https://github.com/theforeman/foreman-installer/pull/583) as a Ruby port of https://github.com/pulp/pulpcore/pull/799#issuecomment-684957879).

bmbouter:
@ekohl we have not been able to make the pulp-smash changes a priority. My apologies for that. To get this going we'll need to ask a bit more from you. Can you collab with @dkliban so he can guide you to open a PR against pulp-smash?

ekohl:
@bmbouter did you talk to Marek? He reminded me that this was still open yesterday ... I've also been unable to prioritize this but I'd be happy to get to the point where I can cross it off as completed. I'll reach out.

bmbouter:
@ekohl actually I didn't talk to Marek. This came up in our weekly PR review at the pulpcore meeting.

bmbouter:
@ekohl thank you!

ekohl:
I at least rebased it and resolved a merge conflict. Will see what CI reports about current failures.

ekohl:
@dkliban from what I can see, I need to modify this:
https://github.com/pulp/pulp-smash/blob/45aa7cd9fb7562b304b9af26968b03f22708ecdc/pulp_smash/pulp3/constants.py#L33

Now I can easily create a PR to change it, but can you help me to understand how to deal with versioning? Do you need to deal with testing existing versions (such as 3.7)?

dkliban:
@ekohl That is definitely the right place to change the value. However, this will break CI for older branches of pulpcore that also use the master branch of pulp-smash for testing. What if we added a new constant?

ekohl:
For context, this is currently how it's used:
```console
$ rg MEDIA_ pulpcore/tests/
pulpcore/tests/functional/api/test_crd_artifacts.py
10:from pulp_smash.pulp3.constants import ARTIFACTS_PATH, MEDIA_PATH
178:        cmd = ("ls", os.path.join(MEDIA_PATH, artifact["file"]))

pulpcore/tests/functional/api/using_plugin/test_orphans.py
10:from pulp_smash.pulp3.constants import MEDIA_PATH
95:            artifact_path = os.path.join(MEDIA_PATH, artifacts_api.read(content["artifact"]).file)
127:            cmd = ("ls", os.path.join(MEDIA_PATH, artifact.file))

pulpcore/tests/functional/api/using_plugin/test_repair.py
7:from pulp_smash.pulp3.constants import MEDIA_PATH, BASE_PATH
72:        artifact1_path = os.path.join(MEDIA_PATH, self.api_client.get(content1["artifact"])["file"])
76:        artifact2_path = os.path.join(MEDIA_PATH, self.api_client.get(content2["artifact"])["file"])
```

One solution is to import the media path from Pulpcore's settings directly. The other is introducing something like `MODERN_MEDIA_PATH` or call it `MEDIA_ROOT`. They all have their downsides.

Could you help me understand why this needs to be defined in pulp_smash? Is it it ensure the path doesn't magically change? That can help steer the right implementation.

dkliban:
It does not need to be defined in pulp_smash. We recently added a method for getting pulp's settings. Here is an example of how it is used[0]. We should definitely just get the value using this method. 

[0] https://github.com/pulp/pulpcore/blob/master/pulpcore/tests/functional/api/test_status.py#L60

ekohl:
Ok, let's see what CI thinks about this. I included #1047 because I did some cleanups in the same test file there.

ekohl:
Uploading now gives a Bad Request. I suspect it has to do with a rebase which now includes 1b6c7360c30cc7d9e2d3b9fc5062ac1f7d69c2de.

I could use some guidance on how to get the exact reason why it is a bad request.

dkliban:
@ekohl When looking at the raw logs for the CI[0]

```
2020-12-03T18:48:08.9934604Z pulp: django.security.SuspiciousFileOperation:ERROR: The joined path (/var/lib/pulp/upload/2ea2ec37-fb6b-4500-8a12-2727ebdac46f) is located outside of the base path component (/var/lib/pulp/media)
2020-12-03T18:48:08.9937042Z pulp: django.request:WARNING: Bad Request: /pulp/api/v3/uploads/3bb6790f-ca5d-474d-ba3a-07f36423e40a/
2020-12-03T18:48:08.9939563Z pulp: django.security.SuspiciousFileOperation:ERROR: The joined path (/var/lib/pulp/upload/0f2ce388-cf14-4c83-812e-b780527fba3d) is located outside of the base path component (/var/lib/pulp/media)
2020-12-03T18:48:08.9941985Z pulp: django.request:WARNING: Bad Request: /pulp/api/v3/uploads/827bc302-6861-4105-b20f-063f5046a384/
2020-12-03T18:48:08.9944392Z pulp: django.security.SuspiciousFileOperation:ERROR: The joined path (/var/lib/pulp/upload/65a290de-d336-4e7d-b8e8-24feeac0fd43) is located outside of the base path component (/var/lib/pulp/media)
2020-12-03T18:48:08.9946842Z pulp: django.request:WARNING: Bad Request: /pulp/api/v3/uploads/b3d2f0b8-89cb-49e9-89ef-2ddb66c606ca/
2020-12-03T18:48:08.9948528Z pulp: django.request:WARNING: Bad Request: /pulp/api/v3/uploads/d87467ef-8307-4c17-bdc9-0d13aefe1b9d/
2020-12-03T18:48:08.9950167Z pulp: django.request:WARNING: Bad Request: /pulp/api/v3/uploads/d87467ef-8307-4c17-bdc9-0d13aefe1b9d/
2020-12-03T18:48:08.9952637Z pulp: django.security.SuspiciousFileOperation:ERROR: The joined path (/var/lib/pulp/upload/cf597e3b-7ded-48e1-aa1a-f58e51afd86a) is located outside of the base path component (/var/lib/pulp/media)
2020-12-03T18:48:08.9955180Z pulp: django.request:WARNING: Bad Request: /pulp/api/v3/uploads/14a37658-85c9-435d-af96-e4b7f8be61ec/
2020-12-03T18:48:08.9956903Z pulp: django.request:WARNING: Method Not Allowed: /pulp/api/v3/workers/0c8be5ac-5006-4f0f-9291-05356c4132ec/
```

ekohl:
@dkliban any idea why its not running CI anymore?

dkliban:
@ekohl  The CI is not running cause you need to rebase again. There is currently a conflict.

ekohl:
Odd, when I wrote that it didn't say so. Later when I refreshed, it did.

ekohl:
I think this is now ready for review.

pulpcore/app/checks.py
```diff
@@ -0,0 +1,43 @@
...
+                CheckWarning(
+                    "Your FILE_UPLOAD_TEMP_DIR setting points to a path that does not exist.",
+                    id="pulpcore.W002",
+                )
+            )
+
+        if media_root_dev and media_root_dev != upload_temp_dir_dev:
+            warnings.append(
+                CheckWarning(
+                    "MEDIA_ROOT and FILE_UPLOAD_TEMP_DIR are on different filesystems. "
```
ekohl, line 37: I'm not 100% sure if there are file moves between these two. That was the reason that they should be on the same FS, right?

dkliban (commented):
I filed an installer story to go along with this change. https://pulp.plan.io/issues/8011. @ekohl Can you please take a look at it and provide feedback?

@ekohl It doesn't look like we need to make any changes to the pulpcore-selinux repository, is my understanding correct?

ekohl:
> I filed an installer story to go along with this change. https://pulp.plan.io/issues/8011. @ekohl Can you please take a look at it and provide feedback?

I [summarized the steps](https://github.com/pulp/pulpcore/pull/799#issuecomment-684957879), including a minimal migration script. I'd recommend to at least link that in the issue, but possibly copy over some info.

> @ekohl It doesn't look like we need to make any changes to the pulpcore-selinux repository, is my understanding correct?

https://github.com/pulp/pulpcore-selinux/commit/d7305fbbd01fdb0e7f9bbd6c412fd7f389495fd8#diff-c0c5c879ce2e9f03d675c8a4efe177ed5f2ee6a8ff41e4a15bc7500d16ae32ef did that already.

dkliban:
> > I filed an installer story to go along with this change. https://pulp.plan.io/issues/8011. @ekohl Can you please take a look at it and provide feedback?
> 
> I [summarized the steps](https://github.com/pulp/pulpcore/pull/799#issuecomment-684957879), including a minimal migration script. I'd recommend to at least link that in the issue, but possibly copy over some info.
> 

Thank you! I thought it would be easier to not do a move but simply use a hardlink. However, I have now learned that you can't hardlink a directory. So we will have the installer perform the operations in the script you provided. 

> > @ekohl It doesn't look like we need to make any changes to the pulpcore-selinux repository, is my understanding correct?
> 
> [pulp/pulpcore-selinux@d7305fb#diff-c0c5c879ce2e9f03d675c8a4efe177ed5f2ee6a8ff41e4a15bc7500d16ae32ef](https://github.com/pulp/pulpcore-selinux/commit/d7305fbbd01fdb0e7f9bbd6c412fd7f389495fd8#diff-c0c5c879ce2e9f03d675c8a4efe177ed5f2ee6a8ff41e4a15bc7500d16ae32ef) did that already.

pulpcore/app/migrations/0024_use_local_storage_for_uploads.py
```diff
@@ -14,6 +14,6 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='upload',
             name='file',
-            field=models.FileField(max_length=255, storage=django.core.files.storage.FileSystemStorage(location='/var/lib/pulp/upload/'), upload_to=''),
+            field=models.FileField(max_length=255, storage=django.core.files.storage.FileSystemStorage(location='/var/lib/pulp/upload'), upload_to=''),
```
dkliban, line 17: is this change intentional? It's in a migration that has already been released

ekohl: Yes. The default for it has changed and if I didn't modify this, Django insisted on a new migration. Since the default is at runtime only (default is not part of the database) I thought it was safe to do so. Given the user may even change the setting, it probably can't even live in the database.

dkliban (approved):
Thank you for this change! Once we merge this we should also merge the installer PR.

https://github.com/pulp/pulp_installer/pull/510/

pulpcore/app/settings.py
```diff
@@ -32,21 +32,22 @@
 
 ALLOWED_HOSTS = ["*"]
 
-MEDIA_ROOT = "/var/lib/pulp/"
+DEPLOY_ROOT = Path("/var/lib/pulp")
+MEDIA_ROOT = str(DEPLOY_ROOT / "media")  # Django 3.1 adds support for pathlib.Path
```
bmbouter, line 36: Will this work on Django 2.2 though? We currently run on 2.2. https://github.com/pulp/pulpcore/blob/master/requirements.txt#L5

dkliban: Yes, I believe the comment is about not being able to use pathlib.Path and using a string instead.

ekohl: Indeed. With Django 3.1+ it can be
```suggestion
MEDIA_ROOT = DEPLOY_ROOT / "media"
```

pulpcore/app/settings.py
```diff
@@ -32,21 +32,22 @@
 
 ALLOWED_HOSTS = ["*"]
 
-MEDIA_ROOT = "/var/lib/pulp/"
+DEPLOY_ROOT = Path("/var/lib/pulp")
```
bmbouter, line 35: Is this a net-new setting for Pulp? Or is setting and existing Django setting? We need docs added for it here: https://docs.pulpproject.org/pulpcore/settings.html

pulpcore/app/settings.py
```diff
@@ -32,21 +32,22 @@
...
 STATIC_URL = "/assets/"
-STATIC_ROOT = os.path.join(MEDIA_ROOT, STATIC_URL.lstrip("/"))
+STATIC_ROOT = DEPLOY_ROOT / STATIC_URL.strip("/")
 
 DEFAULT_FILE_STORAGE = "pulpcore.app.models.storage.FileSystem"
 
-FILE_UPLOAD_TEMP_DIR = os.path.join(MEDIA_ROOT, "tmp/")
-WORKING_DIRECTORY = os.path.join(MEDIA_ROOT, "tmp/")
-CHUNKED_UPLOAD_DIR = os.path.join(MEDIA_ROOT, "upload/")
+FILE_UPLOAD_TEMP_DIR = DEPLOY_ROOT / "tmp"
```
bmbouter, line 48: Also the docs https://docs.pulpproject.org/pulpcore/settings.html probably need updates to identify their defaults being tied up with `DEPLOY_ROOT`.

pulpcore/app/settings.py
ekohl: It is new.

pulpcore/app/settings.py
dkliban: https://github.com/pulp/pulpcore/pull/799#issuecomment-686571440

pulpcore/app/settings.py
dkliban: It's a Pulp only setting that's being added with this PR. +1 on adding it to the docs. I've added a reference to it in the installer docs[0].

[0] https://github.com/pulp/pulp_installer/pull/510/files#diff-c26c81d8d68fbf1c0dfa6018bb2b2256d54bae4729158279ab601b6f34247fffR58

bmbouter (changes requested):
Requesting some docs updates please.

bmbouter:
Also does this need any changes to the pulpcore-selinux fc files? https://github.com/pulp/pulpcore-selinux/blob/master/pulpcore.fc

dkliban:
> Also does this need any changes to the pulpcore-selinux fc files? https://github.com/pulp/pulpcore-selinux/blob/master/pulpcore.fc

No, see https://github.com/pulp/pulpcore/pull/799#issuecomment-748158729

pulpcore/app/settings.py
ekohl: @bmbouter before I start documenting, could you weigh in if this should actually do something? Documenting a setting that doesn't change anything feels useless.

bmbouter: Thank you for the questions (and this effort in general).

There's no need to document settings that can't be used. Let me split my answer in two parts:
1) I had imagined `DEPLOY_ROOT` could be set by the user via dynaconf and if so, it would cause all of the variables that use it to also change. If that's not possible and users would just need to set N variables to effectively change `/var/lib/pulp/` then I don't think there is anything to document.

2) This does change the default value of at least `MEDIA_ROOT` thought right? That default should be updated in the docs if so.

Let me know if there is more I should respond with. I'm excited about this change, ty!

ekohl: >     1. I had imagined `DEPLOY_ROOT` could be set by the user via dynaconf and if so, it would cause all of the variables that use it to also change. If that's not possible and users would just need to set N variables to effectively change `/var/lib/pulp/` then I don't think there is anything to document.

I don't think this is true. Though I didn't test it, I can't see how it would work.
```python
from pathlib import Path
DEPLOY_ROOT = Path('/var/lib/pulp')
MEDIA_PATH = DEPLOY_ROOT / 'media'
assert id(DEPLOY_ROOT) != id(MEDIA_ROOT)
assert repr(MEDIA_PATH) == "PosixPath('/var/lib/pulp/media')"
```
Here `MEDIA_PATH` is a new instance of `Path`. If you set a new value for `DEPLOY_ROOT` via dynaconf, it won't modify anything that copied its value.

Perhaps dynaconf can do something with lazy settings (https://www.dynaconf.com/envvars/#lazy-values) suggests it can do something like that), I have no idea how that would work.

That made me think that perhaps it should be `_DEPLOY_ROOT` but if you can use lazy values, perhaps it becomes useful again. I'm looking for advice here since I don't know dynaconf at all.

>     2\. This does change the default value of at least `MEDIA_ROOT` thought right? That default should be updated in the docs if so.

It does indeed.

bmbouter: That's ok, I don't think we need "one var to reroot many other vars". This PR didn't have that goal, and I won't want to introduce additional goals. What you've already got looks great.

I think just documenting the change for the defualt of `MEDIA_ROOT` is all we need.

bmbouter:
@ekohl @dkliban what handles moving the files on existing installations to make them consistent with the new `MEDIA_ROOT` location?

ekohl:
https://github.com/pulp/pulpcore/pull/799#issuecomment-684957879 summarizes the migration. https://github.com/theforeman/foreman-installer/blob/39003aaed97cb51285761463bdb84cf8bf0cdbca/hooks/pre/34-pulpcore_directory_layout.rb is what foreman-installer uses. https://github.com/pulp/pulp_installer/pull/510 is still open for Pulp itself.

pulpcore/app/settings.py
ekohl: > I think just documenting the change for the defualt of `MEDIA_ROOT` is all we need.

FYI: that's already done here:

https://github.com/pulp/pulpcore/pull/799/files#diff-1d78ba60e1efa4b7b08982e1bbaeceffefea079d402b7d58e57c0cfc7e6c78d8R58

I also took the opportunity to fix the SELinux context type.

ekohl:
FYI: I opened https://github.com/pulp/pulpcore/pull/1074 which extracted a small part of this PR to cherry pick as a regression fix in 3.9. For this reason I this PR on it.

ekohl:
Rebased now that a805312435188aaeeb82b38700598af1c1d001aa has been merged. I've dropped the database migration changes because of it.

bmbouter (approved):
Thank you for everything you've done for this! I really appreciate both the great contribution and patience to get it right.

ekohl:
I'm glad we got there.

PR #802, opened 2020-07-24
Fix the response's structure returned by OperationPostponedResponse

lubosmj:
closes #7196

### DIFF
--- a/pulpcore/app/response.py
+++ b/pulpcore/app/response.py
@@ -11,8 +11,7 @@ class OperationPostponedResponse(Response):
     like the following::
 
         {
-            "pulp_href": "https://example.com/pulp/api/v3/tasks/adlfk-bala-23k5l7-lslser",
-            "task_id": "adlfk-bala-23k5l7-lslser"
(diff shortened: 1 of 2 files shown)


pulpcore/app/response.py
```diff
@@ -12,7 +12,6 @@ class OperationPostponedResponse(Response):
 
         {
             "pulp_href": "https://example.com/pulp/api/v3/tasks/adlfk-bala-23k5l7-lslser",
-            "task_id": "adlfk-bala-23k5l7-lslser"
```
mdellweg, line 15: This is not the whole story...
The response looks like `{"task": "/pulp/api/v3/tasks/<uuid>"}`.

CHANGES/7196.doc
```diff
@@ -0,0 +1 @@
+Removed the redundant field task_id in the docstring for OperationPostponedResponse
```
daviddavis, line 1: Are these docs visible to the end user? If not, I'd probably recommend a `.misc` changelog entry.

PR #806, opened 2020-07-28
Namespace href placeholder in schema generation

mdellweg:
fixes #7209
https://pulp.plan.io/issues/7209

### DIFF
--- a/pulpcore/openapi/__init__.py
+++ b/pulpcore/openapi/__init__.py
@@ -17,8 +17,6 @@ from rest_framework import serializers
 from rest_framework.generics import GenericAPIView
 from rest_framework.views import APIView
 
-from pulpcore.app.models import RepositoryVersion
-
 
 class PulpAutoSchema(AutoSchema):
     """Pulp Auto Schema."""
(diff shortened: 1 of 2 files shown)


mdellweg:
@dkliban There was an additional problem with the content view versions href, where we ended up having
`"{python_repository_version_href}versions/"` which should have been `"{python_repository_href}versions/"`
and now is `"{python_python_repository_href}version/"`

pulpcore/openapi/__init__.py
```diff
@@ -185,10 +188,18 @@ def convert_endpoint_path_params(self, path, view):
         else:
             resource_model = view.queryset.model
         if resource_model:
-            prefix_ = None
+            prefix = None
             if issubclass(resource_model, RepositoryVersion):
-                prefix_ = view.parent_viewset.endpoint_name
-            param_name = self.get_parameter_slug_from_model(resource_model, prefix_)
+                if path.endswith("/versions/"):
```
mdellweg, line 193: maybe `if view._is_list_view():` suites better here.
Combined with `if getattr(view, "parent_viewset", None):` above it might even be generic. I will have a try.

pulpcore/openapi/__init__.py
```diff
@@ -220,10 +232,10 @@ def parse(self, request, public):
                 continue
 
             # Converting path params
```
mdellweg, line 234:
```suggestion
```

mdellweg: This information seems pretty redundant given the name of the called method.

mdellweg:
> it looks good, but I believe this change may break some tests

How about this, it looks like it didn't. :+)
It had flake issues however.

fao89:
> > it looks good, but I believe this change may break some tests
> 
> How about this, it looks like it didn't. :+)
> It had flake issues however.

I forgot hrefs become positional args

PR #808, opened 2020-07-28
Omit a view/viewset from the OpenAPI schema

fao89:
https://pulp.plan.io/issues/7133
closes #7133


mdellweg:
Sorry, i wanted to comment this on the ticket earlier.
But a quote from the DRF-Docs:
>  You may disable schema generation for a view by setting schema to None:
```
class CustomView(APIView):
    ...
    schema = None  # Will not appear in schema
```

As i understand, spectacular extends the DRF-Schemas, so this might already work for us.

mdellweg:
I tried this now, and it turns out that this mechanism already works without this change. It would be nice to document it however.
how about here: https://github.com/pulp/pulpcore/blob/master/docs/plugins/plugin-writer/concepts/index.rst#custom-api-url-routes

mdellweg:
Yes this looks good! I'd still love to see a cross reference on in the "custom api routes" section, since i consider that the main usecase for this feature. Something like: `Consider <a href=...>omitting</a> those views from the api docs.`

PR #810, opened 2020-07-28
`HttpDownloader` control on response codes 400+

fao89:
https://pulp.plan.io/issues/7117
closes #7117

### DIFF
--- a/CHANGES/plugin_api/7117.feature
+++ b/CHANGES/plugin_api/7117.feature
@@ -1,0 +1,2 @@
+Enalbing plugin writers to have more control on `HttpDownloader` response codes 400+
+by subclassing `HttpDownloader` and overwriting `raise_for_status` method
--- a/pulpcore/download/http.py
+++ b/pulpcore/download/http.py
@@ -119,6 +119,7 @@ class HttpDownloader(BaseDownloader):
         proxy=None,
         proxy_auth=None,
         headers_ready_callback=None,
+        silence_errors_for_response_status_codes=set(),
         **kwargs,
     ):
         """
@@ -134,6 +135,8 @@ class HttpDownloader(BaseDownloader):
                 as its argument. The callback will be called when the response headers are
                 available. The dictionary passed has the header names as the keys and header values
                 as its values. e.g. `{'Transfer-Encoding': 'chunked'}`
+            silence_errors_for_response_status_codes (iterable): An iterable of response exception
+                codes to be ignored when raising exception. e.g. `{404}`
             kwargs (dict): This accepts the parameters of
                 :class:`~pulpcore.plugin.download.BaseDownloader`.
         """
@@ -149,7 +152,21 @@ class HttpDownloader(BaseDownloader):
         self.proxy = proxy
         self.proxy_auth = proxy_auth
         self.headers_ready_callback = headers_ready_callback
+        self.silence_errors_for_response_status_codes = silence_errors_for_response_status_codes
         super().__init__(url, **kwargs)
+
+    def raise_for_status(self, response):
+        """
+        Raise error if aiohttp response status is >= 400 and not silenced.
+
+        Args:
+            response (aiohttp.ClientResponse): The response to handle.
+
+        Raises:
+               aiohttp.ClientResponseError: When the response status is >= 400.
+        """
+        if response.status not in self.silence_errors_for_response_status_codes:
+            response.raise_for_status()
 
     async def _handle_response(self, response):
         """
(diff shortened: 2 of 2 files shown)


pulpcore/download/http.py
```diff
@@ -119,6 +119,7 @@ def __init__(
         proxy=None,
         proxy_auth=None,
         headers_ready_callback=None,
+        skip_status=[],
```
bmbouter, line 122: This name could be clearer to me. Maybe `silence_errors_for_response_status_codes`. It's long but it's descriptive, while this could mean that it skips downloading something somehow.

bmbouter: I think this would be better as `set()`.

pulpcore/download/http.py
```diff
@@ -134,6 +135,7 @@ def __init__(
                 as its argument. The callback will be called when the response headers are
                 available. The dictionary passed has the header names as the keys and header values
                 as its values. e.g. `{'Transfer-Encoding': 'chunked'}`
+            skip_status (list): A list of status to be ignored when raising exception. e.g. `[404]`
```
bmbouter, line 138: If it was used as a `set()` then this would probably become s/list/iterable/

pulpcore/download/http.py
fao89: yep, this name is not good, I put as placeholder, thank you for the suggestion!

CHANGES/plugin_api/7117.feature
```diff
@@ -0,0 +1 @@
+`HttpDownloader` control on response codes 400+
```
bmbouter, line 1: This could use a longer description that the plugin writer can have control through the subclass defining the method and separately that they can specify a list of codes that would be silenced.

pulpcore/download/http.py
```diff
@@ -134,6 +135,8 @@ def __init__(
                 as its argument. The callback will be called when the response headers are
                 available. The dictionary passed has the header names as the keys and header values
                 as its values. e.g. `{'Transfer-Encoding': 'chunked'}`
+            silence_errors_for_response_status_codes (set): A list of status to be ignored when
+            raising exception. e.g. `{404}`
```
bmbouter, line 139: This wrapped line needs an indent (see the form above).

CHANGES/plugin_api/7117.feature
```diff
@@ -0,0 +1,2 @@
+Enalbing plugin writers to have more control on `HttpDownloader` response codes 400+
+By subclassing `HttpDownloader` and overwriting `raise_for_status` method
```
bmbouter, line 2:
```suggestion
by subclassing `HttpDownloader` and overwriting `raise_for_status` method
```

pulpcore/download/http.py
```diff
@@ -134,6 +135,8 @@ def __init__(
                 as its argument. The callback will be called when the response headers are
                 available. The dictionary passed has the header names as the keys and header values
                 as its values. e.g. `{'Transfer-Encoding': 'chunked'}`
+            silence_errors_for_response_status_codes (set): A list of status to be ignored when
```
bmbouter, line 138: Technically this can be s/set/iterable/

bmbouter: Also when describing it an iterable is more general.

bmbouter:
```suggestion
            silence_errors_for_response_status_codes (iterable): An iterable of statuses to be ignored when
```

pulpcore/download/http.py
```diff
@@ -149,8 +152,22 @@ def __init__(
         self.proxy = proxy
         self.proxy_auth = proxy_auth
         self.headers_ready_callback = headers_ready_callback
+        self.silence_errors_for_response_status_codes = silence_errors_for_response_status_codes
         super().__init__(url, **kwargs)
 
+    def raise_for_status(self, response):
+        """
+        Raise error if aiohttp response status is >= 400
```
bmbouter, line 160:
```suggestion
        Raise error if aiohttp response status is >= 400 and not silenced.
```

pulpcore/download/http.py
```diff
@@ -134,6 +135,8 @@ def __init__(
                 as its argument. The callback will be called when the response headers are
                 available. The dictionary passed has the header names as the keys and header values
                 as its values. e.g. `{'Transfer-Encoding': 'chunked'}`
+            silence_errors_for_response_status_codes (iterable): An iterable of statuses to be
```
bmbouter, line 138:
```suggestion
            silence_errors_for_response_status_codes (iterable): An iterable of response exception codes to be
```

PR #812, opened 2020-07-29
Stop preventing to render the status endpoint

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/openapi/__init__.py
+++ b/pulpcore/openapi/__init__.py
@@ -14,8 +14,6 @@ from rest_framework.schemas.utils import get_pk_description
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiParameter
 from rest_framework import serializers
-from rest_framework.generics import GenericAPIView
-from rest_framework.views import APIView
 
 
 class PulpAutoSchema(AutoSchema):
@@ -281,12 +279,6 @@ class PulpSchemaGenerator(SchemaGenerator):
             if not self.has_view_permissions(path, method, view):
                 continue
 
-            if "docs" in path or "status" in path:
-                continue
-
-            if not self.contains_serializer(view):
-                continue
-
             schema = view.schema
 
             path = self.convert_endpoint_path_params(path, view, schema)
@@ -337,19 +329,6 @@ class PulpSchemaGenerator(SchemaGenerator):
             result[path][method.lower()] = operation
 
         return result
-
-    def contains_serializer(self, view):
-        """Check if view contains serializer."""
-        if isinstance(view, GenericAPIView):
-            return True
-        if isinstance(view, APIView):
-            if callable(getattr(view, "get_serializer", None)):
-                return True
-            if callable(getattr(view, "get_serializer_class", None)):
-                return True
-            if hasattr(view, "serializer_class"):
-                return True
-        return False
 
     def get_schema(self, request=None, public=False):
         """ Generate a OpenAPI schema. """


mdellweg:
@fao89 You may know better, why you excluded those views from the schema generation...
At least status we do want to have. It's rendered like this now:

``` json
{
  "get": {
    "operationId": "status_read",
    "description": "Returns app information including the version of pulpcore and loaded pulp plugins,\nknown workers, database connection status, and messaging connection status",
    "summary": "Inspect status of Pulp",
    "tags": [
      "Status"
    ],
    "responses": {
      "200": {
        "content": {
          "application/json": {
            "schema": {
              "$ref": "#/components/schemas/StatusResponse"
            }
          }
        },
        "description": ""
      }
    }
  }
}
```

fao89:
> @fao89 You may know better, why you excluded those views from the schema generation...
> At least status we do want to have. It's rendered like this now:
> 
> ```json
> {
>   "get": {
>     "operationId": "status_read",
>     "description": "Returns app information including the version of pulpcore and loaded pulp plugins,\nknown workers, database connection status, and messaging connection status",
>     "summary": "Inspect status of Pulp",
>     "tags": [
>       "Status"
>     ],
>     "responses": {
>       "200": {
>         "content": {
>           "application/json": {
>             "schema": {
>               "$ref": "#/components/schemas/StatusResponse"
>             }
>           }
>         },
>         "description": ""
>       }
>     }
>   }
> }
> ```

I removed because I thought we did not want it on the bindings

mdellweg:
Well, then i say: We want it. ;)

pulpcore/openapi/__init__.py
```diff
@@ -213,10 +213,7 @@ def parse(self, request, public):
             if not self.has_view_permissions(path, method, view):
                 continue
 
-            if "docs" in path or "status" in path:
-                continue
-
-            if not self.contains_serializer(view):
```
fao89, line 219: this is for another thing, drf-spectacular logs error when the view does not have serializer:
https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/openapi.py#L761-L792

fao89: and this method is only used here, so if we decide to remove this call, we need to remove the method

mdellweg: with seven different plugins installed, i do not see a single one trigger that code path. Nor did i see the drf-spectacular error logging you mentioned. Also i would assume that a "proper pulp view(set)" should provide serializers. And all the others should use the `schema=None` method of your other PR #808 . Therefore i am going to remove this method also.
Thanks for pointing out.

mdellweg: I deleted the test for `"docs" in path` also, since it didn't trigger either. Seems like spectacular knew that documenting the documentation doesn't add any value.
And if we ever have a pulp_docs content type, we should not hide it. ;)

fao89: `pulp_ansible` was one of them? I remember `pulp_ansible` required lots of changes in `pulpcore`

mdellweg: No, it wasn't. I can look into that next week.

mdellweg: I do not see an error when using this with pulp_ansible.
However pulp_ansible might make use of the `schema=None` feature to prevent rendering galaxy endpoints here.

fao89: I agree, `schema=None` is a better solution!

mdellweg: https://github.com/pulp/pulp_ansible/pull/338

PR #814, opened 2020-07-30
Revert "`HttpDownloader` control on response codes 400+"

fao89:
Reverts pulp/pulpcore#810


fao89:
> We also need to revert the release note. I recommend using `git revert` to create the commit.

the subclass thing remains, so the changelog still making sense, I just fixed a typo

PR #824, opened 2020-08-04
Enable the swaggerui view supported by spectactular

alikins:
Enable the swaggerui view supported by spectactular

Based on example at:
https://drf-spectacular.readthedocs.io/en/latest/readme.html#take-it-for-a-spin

Fixes #7271
https://pulp.plan.io/issues/7271

### DIFF
--- a/CHANGES/7271.feature
+++ b/CHANGES/7271.feature
@@ -1,0 +1,1 @@
+Enable swagger-ui drf-spectactular view
--- a/pulpcore/app/urls.py
+++ b/pulpcore/app/urls.py
@@ -7,6 +7,7 @@ from drf_spectacular.views import (
     SpectacularJSONAPIView,
     SpectacularYAMLAPIView,
     SpectacularRedocView,
+    SpectacularSwaggerView,
 )
 from rest_framework import permissions
 from rest_framework.schemas import get_schema_view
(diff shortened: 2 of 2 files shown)


fao89 (approved):

bmbouter:
@dkliban can you look at this?

daviddavis:
This should fix the test failure by making the check case insensitive: https://github.com/pulp/pulpcore/pull/897

daviddavis:
@alikins can you rebase and this should pass.

mdellweg:
I am a little bit concerned that this adds a feature we already have from drf. And i would not want to support two things providing the exact same thing.
So is this significantly better than what DRF provides? Should we stop using the other one when switching on this?
Maybe I'm missing something here.

fao89:
> I am a little bit concerned that this adds a feature we already have from drf. And i would not want to support two things providing the exact same thing.
> So is this significantly better than what DRF provides? Should we stop using the other one when switching on this?
> Maybe I'm missing something here.

drf-spectacular extends DRF (I did not test it, but maybe DRF would break, since drf-spectacular adds resources in some classes)

fao89 (commented):
When trying the URL, it generates this curl request:
```
curl -X GET "http://6ddadf73fb8c.ngrok.io/pulp/api/v3/ansible/collections/" -H  "accept: application/json" -H  "X-CSRFToken: fyT7miy1KGZMjBDwb0AQ3k9s6A5R6iXzVyJ1fZcYI9EbGgOFZXijbccKEy6UnESR"
```
which fails to authenticate:
image

fao89:
Actually, the swagger view works, I got 403 because I was logged in locally and used swagger from ngrok
image

daviddavis:
@alikins can you add a note to the changelog that this is a tech preview and rebase?

daviddavis:
I'm closing due to inactivity. If we can get this PR passing and add some docs, then would be ideal. Please reopen if you have time.

PR #835, opened 2020-08-10
Use PulpTemporaryFile instead of Artifact when passing temporary data

### LINKED ISSUE #7270: Return useful information to user from base task-actions
With https://github.com/pulp/pulpcore/pull/7233 we stopped returning potentially-sensitive info in failed task error-messages. We, alas, also *lost* information that the creators of those tasks need, in order to be able to figure out what happened.

As an example - general_create() validates the incoming serializer here https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/base.py#L41 . That used to tell the user *exactly* what the problem was.  Now, all you get back is the generic "something went wrong" error.

We need to get the validation-error-msgs back to the user - things like "I expected JSON and didn't get it", "the URL you're syncing from isn't working", "I couldn't find the file you said to use", etc, are things the API-user should be able to see, not just the person who has access to the logs.

Whether we should be validating serializers in the tasks or in the viewset is a question for some other, later, day...

lubosmj:
closes #7270

### DIFF
--- a/pulpcore/app/tasks/base.py
+++ b/pulpcore/app/tasks/base.py
@@ -1,28 +1,22 @@
 from pulpcore.app.apps import get_plugin_config
-from pulpcore.app.models import Artifact, CreatedResource
+from pulpcore.app.models import CreatedResource, PulpTemporaryFile
 from pulpcore.app.files import PulpTemporaryUploadedFile
 
 
-def general_create_from_temp_file(app_label, serializer_name, *args, **kwargs):
+def general_create_from_temp_file(app_label, serializer_name, temp_file_pk, *args, **kwargs):
     """
-    Create a model instance from contents stored in a temporary Artifact.
+    Create a model instance from contents stored in a temporary file.
 
-    A caller should always pass the dictionary "data", as a keyword argument, containing the
-    href to the temporary Artifact. Otherwise, the function does nothing.
-
-    This function calls the function general_create() to create a model instance.
-    Data passed to that function already contains a serialized artifact converted
-    to PulpTemporaryUploadFile that will be deleted afterwards.
+    A task which executes this function takes the ownership of a temporary file and deletes it
+    afterwards. This function calls the function general_create() to create a model instance.
     """
-    data = kwargs.pop("data", None)
-    if data and "artifact" in data:
-        named_model_view_set = get_plugin_config(app_label).viewsets_module.NamedModelViewSet
-        artifact = named_model_view_set.get_resource(data.pop("artifact"), Artifact)
+    temp_file = PulpTemporaryFile.objects.get(pk=temp_file_pk)
 
-        data["file"] = PulpTemporaryUploadedFile.from_file(artifact.file)
+    data = kwargs.pop("data", {})
+    data["file"] = PulpTemporaryUploadedFile.from_file(temp_file.file)
 
-        general_create(app_label, serializer_name, data=data, *args, **kwargs)
-        artifact.delete()
+    general_create(app_label, serializer_name, data=data, *args, **kwargs)
+    temp_file.delete()
 
 
 def general_create(app_label, serializer_name, *args, **kwargs):
--- a/pulpcore/plugin/viewsets/content.py
+++ b/pulpcore/plugin/viewsets/content.py
@@ -9,7 +9,7 @@ from pulpcore.plugin.serializers import (
     ArtifactSerializer,
     AsyncOperationResponseSerializer,
 )
-from pulpcore.plugin.models import Artifact
+from pulpcore.plugin.models import Artifact, PulpTemporaryFile
 from pulpcore.plugin.tasking import enqueue_with_reservation
 from pulpcore.plugin.viewsets import (
     ContentViewSet,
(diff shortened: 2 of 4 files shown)


lubosmj:
This change was tested on the pulp_rpm plugin because UpdateRecordSerializer benefits from NoArtifactContentUploadSerializer (https://github.com/pulp/pulp_rpm/blob/bbf70f539d9524d53b777ebdd63d4e445e6120f3/pulp_rpm/app/serializers/advisory.py#L203).

pulpcore/app/tasks/base.py
```diff
@@ -1,28 +1,27 @@
...
-    Create a model instance from contents stored in a temporary Artifact.
-
-    A caller should always pass the dictionary "data", as a keyword argument, containing the
-    href to the temporary Artifact. Otherwise, the function does nothing.
+    Create a model instance from contents stored in a temporary file.
 
-    This function calls the function general_create() to create a model instance.
-    Data passed to that function already contains a serialized artifact converted
-    to PulpTemporaryUploadFile that will be deleted afterwards.
+    A caller should always pass the dictionary "data", as a keyword argument, containing
```
dralley, line 12: If it should always be present, I kinda think it should be an explicit keyword argument, although I'm interested in what others think and if there's any good reasons not to do that.

dralley: Otherwise, everything looks good

CHANGES/7270.misc
```diff
@@ -0,0 +1,2 @@
+Replaced an Artifact object, which was used for storing temporary data in
```
bmbouter, line 1: Can this become CHANGES/plugin_api/7270.misc

pulpcore/app/tasks/base.py
bmbouter: If it's always required then I think it should be an arg even instead of a keyword arg since it can't have a reasonable default.

lubosmj: There is such a statement because I wanted to explicitly note that this parameter is mandatory. I could write a similar docstring like we did for the function `general_create()`, but I decided to do write it more precisely.

Also, in the function `general_create()` we have the field `artifact` that is stored within `data` too. I wonder why this should not be done in exactly the same way.
https://github.com/pulp/pulpcore/blob/d78416238190111ff2cc5f53e8542d949b7aff0d/pulpcore/plugin/viewsets/content.py#L56-L65
Do you still want to make that change?

dralley: I think I lean towards "yes" but I'm not familiar with this part of the codebase.  Curious if there's any reasons why not to.

daviddavis:
@lubosmj what's the status of this work?

pulpcore/app/tasks/base.py
```diff
@@ -1,28 +1,22 @@
...
-    Create a model instance from contents stored in a temporary Artifact.
+    Create a model instance from contents stored in a temporary file.
 
-    A caller should always pass the dictionary "data", as a keyword argument, containing the
-    href to the temporary Artifact. Otherwise, the function does nothing.
-
-    This function calls the function general_create() to create a model instance.
-    Data passed to that function already contains a serialized artifact converted
-    to PulpTemporaryUploadFile that will be deleted afterwards.
+    A caller should always pass a reference to the temporary file. This function
```
mdellweg, line 10: `temp_file_pk` is a mandatory arg, so it should be a "must".
More important, we should mention that the task takes ownership of that temporary file and deletes it afterwards.

pulpcore/plugin/viewsets/content.py
```diff
@@ -22,11 +22,6 @@
 class ContentUploadViewSet(ContentViewSet):
     """A base ContentViewSet with added upload functionality."""
 
-    def __init__(self, *args, **kwargs):
-        """Set a default task's type that creates a new content from an uploaded Artifact."""
-        super().__init__(*args, **kwargs)
-        self._task_function_type = tasks.base.general_create
```
mdellweg, line 28: This makes the base class an abstract base class, right?

pulpcore/plugin/viewsets/content.py
```diff
@@ -64,53 +97,6 @@ def create(self, request):
...
-
-class NoArtifactContentUploadViewSet(ContentUploadViewSet):
-    """A ViewSet for uploads that do not require to store an uploaded content as an Artifact."""
-
-    def __init__(self, *args, **kwargs):
-        """Set a task's function type that creates a new content using a temporary Artifact."""
-        super().__init__(*args, **kwargs)
-        self._task_function_type = tasks.base.general_create_from_temp_file
-
-    def init_content_data(self, serializer, request):
```
mdellweg, line 86: Was it impossible to prepare the tempfile in this polymorphic manner?

mdellweg (changes requested):

pulpcore/plugin/viewsets/content.py
lubosmj: Kind of. I am going to update the docstring. The classes which inherit from this ViewSet have in common one thing, that is the method `get_deffered_context()`.

pulpcore/plugin/viewsets/content.py
lubosmj: I was requested to change the task's function parameters. Therefore, if I would stick to the current pattern, I would be required to add an additional method that would deal with the function parameters.

I am not sure how polymorphism could serve better in this case. Because the ViewSets do not have many things in common at the moment. I opted to move backwards and just added a new class that works with temporary files.

mdellweg: Thanks for the explanation.

bmbouter (approved):
This all looks ok to me. Thanks @lubosmj !

bmbouter:
I'm LGTM on this, @mdellweg are you?

mdellweg:
I think, every comment has been addressed. :+1:

PR #836, opened 2020-08-10
Extra require: postgres, s3, azure, prometheus and test

fao89:
https://pulp.plan.io/issues/6844
closes #6844

Ref: https://github.com/pypa/pip/issues/3147#issuecomment-144656466

### DIFF
--- a/CHANGES/6844.feature
+++ b/CHANGES/6844.feature
@@ -1,0 +1,1 @@
+Extra require: s3, azure, prometheus and test
--- a/setup.py
+++ b/setup.py
@@ -6,6 +6,9 @@ with open("README.md") as f:
 with open("requirements.txt") as requirements:
     requirements = requirements.readlines()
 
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -19,6 +22,13 @@
     url="https://pulpproject.org",
     python_requires=">=3.6",
     install_requires=requirements,
+    extras_require={
+        "postgres": ["psycopg2-binary"],
```
daviddavis, line 26: I think the set of users that would be interested in using `psycopg2-binary` is very small. I would either drop this or give it a very specific name like `psycopg2-binary`.

CHANGES/6844.feature
```diff
@@ -0,0 +1 @@
+Extra require: postgres, s3, azure, prometheus and test
```
daviddavis, line 1: s/postgres//

PR #840, opened 2020-08-11
Decouple plugin filter apidoc from "bindings"

mdellweg:
fixes #7306
https://pulp.plan.io/issues/7306


daviddavis:
Tested and this worked for me. Merging.

PR #844, opened 2020-08-12
Delete associated files from the storage

lubosmj:
closes #7316

### DIFF
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -95,17 +95,6 @@ class HandleTempFilesMixin:
         finally:
             self.file.close()
 
-    def delete(self, *args, **kwargs):
-        """
-        Deletes Model and the file associated with the Model
-
-        Args:
-            args (list): list of positional arguments for Model.delete()
-            kwargs (dict): dictionary of keyword arguments to pass to Model.delete()
-        """
-        super().delete(*args, **kwargs)
-        self.file.delete(save=False)
-
 
 class Artifact(HandleTempFilesMixin, BaseModel):
     """
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,7 @@ aiofiles
 backoff
 Django~=2.2.14  # LTS version, switch only if we have a compelling reason to
 django-currentuser~=0.5.1
+django-cleanup~=5.0.0
 django-filter~=2.3.0
 django-guardian~=2.3.0
 django-import-export~=2.3.0
--- a/CHANGES/7316.feature
+++ b/CHANGES/7316.feature
@@ -1,0 +1,1 @@
+Added the django-cleanup handlers for removing files stored within FileField
--- a/CHANGES/plugin_api/7316.feature
+++ b/CHANGES/plugin_api/7316.feature
@@ -1,0 +1,2 @@
+Enabled the automatic removal of files, which are stored in FileField, when a corresponding
+model's delete() method is invoked
--- a/pulpcore/app/models/upload.py
+++ b/pulpcore/app/models/upload.py
@@ -46,17 +46,6 @@ class Upload(BaseModel):
 
         self.chunks.create(offset=offset, size=len(chunk))
 
-    def delete(self, *args, **kwargs):
-        """
-        Deletes Upload model and the file associated with the model
-
-        Args:
-            args (list): list of positional arguments for Model.delete()
-            kwargs (dict): dictionary of keyword arguments to pass to Model.delete()
-        """
-        super().delete(*args, **kwargs)
-        self.file.delete(save=False)
-
 
 class UploadChunk(BaseModel):
     """
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -65,6 +65,8 @@ INSTALLED_APPS = [
     "rest_framework",
     # pulp core app
     "pulpcore.app",
+    # the cleanup config has to be placed last
+    "django_cleanup.apps.CleanupConfig",
 ]
 
 # Enumerate the installed Pulp plugins during the loading process for use in the status API


pulpcore/app/models/content.py
```diff
@@ -299,6 +301,14 @@ def storage_path(self, name):
 
     file = models.FileField(null=False, upload_to=storage_path, max_length=255)
 
+    @hook(BEFORE_DELETE)
```
mdellweg, line 304: It needs to be after delete. I'm sorry. You will end in an inconsistent state in case of a rolled_back transaction.

what about:
```
def delete(self):
    with transaction.atomic() as ta:
        ta.on_commit(lambda: delete_from_storage(self.file.name))
        super().delete()
```

mdellweg: I'm wondering django didn't come up with this.

daviddavis: This is [what django-cleanup does](https://github.com/un1t/django-cleanup/blob/master/django_cleanup/handlers.py#L104).

lubosmj: The problem is that when the model is being removed because of ondelete.CASCADE, then the method model.delete() is not invoked: https://docs.djangoproject.com/en/3.0/ref/models/fields/#django.db.models.CASCADE.

lubosmj: Seems like django-cleanup resolves all the mentioned issues.

pulpcore/app/models/content.py
```diff
@@ -283,8 +283,9 @@ class PulpTemporaryFile(HandleTempFilesMixin, BaseModel):
 
     Fields:
 
-        file (models.FileField): The stored file. This field should be set using an absolute path to
-            a temporary file. It also accepts `class:django.core.files.File`.
+        file (models.FileField): The stored file. This field accepts `class:django.core.files.File`.
+            The field can be set using an absolute path to a temporary file. Note that the content
+            of the temporary file will not be automatically copied to the storage backend.
```
mdellweg, line 288: How is it transferred then?
Maybe:
Note that the content of the temporary file will only copied to the storage backend once save() is called. It will be immutable afterwards.

lubosmj: It is not copied. It is not transferred at all.

```
In [1]: import tempfile                                                                                               

In [2]: with tempfile.NamedTemporaryFile("ab") as tf: 
   ...:     temp_file = PulpTemporaryFile(file=tf.name) 
   ...:     temp_file.save() 
   ...:                                                                                                               

In [3]: PulpTemporaryFile.objects.all()                                                                               
Out[3]: <QuerySet [<PulpTemporaryFile: pk=db5b57ba-1f65-4f0c-85b5-6d01c84b7d77>]>

In [4]: PulpTemporaryFile.objects.all()[0].file                                                                       
Out[4]: <FieldFile: /tmp/tmp7ttftw60>

In [5]: PulpTemporaryFile.objects.all()[0].file.read()
OSError: File does not exist: tmp/tmp7ttftw60

```

requirements.txt
```diff
@@ -4,6 +4,7 @@ aiofiles
 backoff
 Django~=2.2.14  # LTS version, switch only if we have a compelling reason to
 django-currentuser~=0.5.1
+django-cleanup~=5.0.0
```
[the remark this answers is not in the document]
lubosmj: Yes, it work in both cases.

pulpcore/app/models/content.py
mdellweg: I'm confused. Isn't this how it's supposed to work in the first place? Also should the file not be in the django-storage location after saving?

mdellweg: Also i get:
`SuspiciousFileOperation: The joined path (/tmp/tmpxko75ex9) is located outside of the base path component (/var/lib/pulp)`

daviddavis: Oh wow, so [this code](https://github.com/pulp/pulpcore/blob/0f30cfefdfedc53cf79dc6ce9d12ff9e33351e96/pulpcore/app/models/content.py#L290-L298) does nothing?

daviddavis: If we use [ArtifactFileField](https://github.com/pulp/pulpcore/blob/b94abd64d76ea4554e6750ff38ce458eaa888cc8/pulpcore/app/models/fields.py#L11) then it should move it into place. I'll open an issue for this.

lubosmj: We decided to have FileField instead: https://github.com/pulp/pulpcore/pull/793#discussion_r459023880

mdellweg: So the example we provide does not even work?
How is this PulpTemporaryFile supposed to be used?

requirements.txt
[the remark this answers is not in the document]
daviddavis: A drawback (or perhaps benefit) of django-cleanup is devs/plugin writers must specify a decorator if they do not want their FileField files deleted. I'm conflicted about this approach because while I don't like the implicit nature of it, I do find it surprising that django doesn't delete files when models get deleted.

CC @pulp/core for more visibility.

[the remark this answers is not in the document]
daviddavis: If we do use `django-cleanup` we can remove these:

https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/upload.py#L49-L58
https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/content.py#L98-L107

pulpcore/app/models/content.py
daviddavis: I believe this works if you pass in a File object (or rather a `PulpTemporaryUploadedFile`). What doesn't work is if you pass in a file path to a file on disk. We need to fix the latter.

requirements.txt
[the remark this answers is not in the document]
mdellweg: And given @lubosmj 's insight it will also work properly for cascaded deletes.

CHANGES/plugin_api/7316.bugfix
```diff
@@ -0,0 +1 @@
+Added the django-cleanup handlers for removing temporary files from the storage
```
daviddavis, line 1: I would probably make this a `.removal` or `.feature`. Or better yet, also add a `.feature` saying that files for FileFields will automatically be cleaned up using django-cleanup.

daviddavis:
Also, I think we can remove these methods with this change:

https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/upload.py#L49-L58

https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/content.py#L98-L107

CHANGES/plugin_api/7316.bugfix
bmbouter: I agree

PR #860, opened 2020-08-21
OpenAPI: do not discard components without properties

fao89:
https://pulp.plan.io/issues/7347
closes #7347


pulpcore/openapi/__init__.py
```diff
@@ -182,6 +183,25 @@ def _map_serializer_field(self, field, direction):
             del mapped["additionalProperties"]
         return mapped
 
+    def resolve_serializer(self, serializer, direction):
+        """Serializer to component."""
+        component = super().resolve_serializer(serializer, direction)
```
mdellweg, line 188: The result of this call is only used, if the condition below is not met. Should we move it to "else"?

fao89: nice catch!

ipanova (approved):

PR #862, opened 2020-08-25
Store toc_info as a separate attribute on PulpExport.

ggainey:
Updated/corrected Field comments on PulpExport. Added a line to test
for toc_info to tests.


CHANGES/7221.feature
```diff
@@ -0,0 +1 @@
+Added a toc_info attribute to PulpExport to make accessing the export-TOC easier.
```
daviddavis, line 1: I would maybe include in here what this field is (ie it contains full-path filename and SHA256-checksum).

ggainey: Good call, updated text

PR #864, opened 2020-08-25
Fixing groups API validation

fao89:
https://pulp.plan.io/issues/7329
closes #7329

### DIFF
--- a/pulpcore/app/viewsets/user.py
+++ b/pulpcore/app/viewsets/user.py
@@ -1,7 +1,9 @@
 from gettext import gettext as _
+import uuid
 
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group, Permission
+from django.core.exceptions import FieldError
 from django.shortcuts import get_object_or_404
 from drf_spectacular.utils import extend_schema
 from guardian.models.models import GroupObjectPermission
@@ -18,6 +20,8 @@ from pulpcore.app.serializers.user import (
     UserSerializer,
 )
 
+User = get_user_model()
+
 
 class UserViewSet(
     NamedModelViewSet, mixins.RetrieveModelMixin, mixins.ListModelMixin,
@@ -31,7 +35,7 @@ class UserViewSet(
 
     endpoint_name = "users"
     serializer_class = UserSerializer
-    queryset = get_user_model().objects.all()
+    queryset = User.objects.all()
 
     @extend_schema(description="List user permissions.",)
     @action(detail=True, methods=["get"], serializer_class=PermissionSerializer)
@@ -98,7 +102,12 @@ class GroupModelPermissionViewSet(NamedModelViewSet):
 
             data[key] = value
 
-        return get_object_or_404(Permission, **data)
+        try:
+            permission = Permission.objects.get(**data)
+        except (Permission.MultipleObjectsReturned, Permission.DoesNotExist, FieldError) as exc:
+            raise ValidationError(str(exc))
+
+        return permission
 
     @extend_schema(description="Retrieve a model permission from a group.")
     def retrieve(self, request, group_pk, pk):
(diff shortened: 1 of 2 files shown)


pulpcore/app/viewsets/user.py
```diff
@@ -98,7 +100,12 @@ def get_model_permission(self, request):
 
             data[key] = value
 
-        return get_object_or_404(Permission, **data)
+        try:
+            permission = Permission.objects.get(**data)
+        except (Permission.MultipleObjectsReturned, Permission.DoesNotExist, FieldError) as exc:
+            raise ValidationError(_(str(exc)))
```
mdellweg, line 106: I do not believe, gettext works in this situation. How is it supposed to extract the translatable string here?

pulpcore/app/viewsets/user.py
```diff
@@ -263,7 +293,11 @@ def create(self, request, group_pk):
             raise ValidationError(
                 _("Please provide one of the following values for User: 'pk', 'id', 'username'")
             )
-        user = get_object_or_404(get_user_model(), **request.data)
+        User = get_user_model()
```
mdellweg, line 296: Would it be OK do make this call at the top level? Or is django insufficiently initialized at that point?

pulpcore/app/viewsets/user.py
```diff
@@ -263,7 +293,11 @@ def create(self, request, group_pk):
...
+        try:
+            user = User.objects.get(**request.data)
+        except (User.MultipleObjectsReturned, User.DoesNotExist, FieldError) as exc:
```
mdellweg, line 299: How can `MultipleObjectsReturned` happen, if we require to provide `pk`, `id` or `username`?
Is this overly cautious?

fao89: it was a bad copy/paste

pulpcore/app/viewsets/user.py
fao89: I do not know how gettext works, `str(exc)` returns a string I imagined it would work

mdellweg: Well, I thought it must be able to extract the strings without calling the code. So it probably looks for calls to ’_("...")’.

dkliban: The exception text should already be translated if a translation is available. Please remove the _() around str(exc).

pulpcore/app/viewsets/user.py
```diff
@@ -164,10 +173,33 @@ class GroupObjectPermissionViewSet(NamedModelViewSet):
...
+    def get_object_pk(self, request):
+        """Get object pk."""
+
+        if "obj" not in request.data:
+            raise ValidationError(_("Please provide 'obj' value"))
+        try:
+            obj_pk = request.data["obj"].strip("/").split("/")[-1]
+            uuid.UUID(obj_pk)
+        except (AttributeError, ValueError):
+            raise ValidationError(_(f"Invalid value for 'obj': {request.data['obj']}"))
```
mdellweg, line 185: Stackoverflow says nope:
https://stackoverflow.com/questions/49797658/how-to-use-gettext-with-python-3-6-f-strings

```suggestion
            raise ValidationError(_("Invalid value for 'obj': " + str(request.data['obj'])))
```
Or the good old "format" of course (Usually this helps people translating if they see the whole string template).
```suggestion
            raise ValidationError(_("Invalid value for 'obj': {obj}").format(request.data['obj']))
```

pulpcore/app/viewsets/user.py
```diff
@@ -164,10 +173,33 @@ class GroupObjectPermissionViewSet(NamedModelViewSet):
...
+        if "permission" not in request.data:
+            raise ValidationError(_("Please provide 'permission' value"))
+
         codename = request.data["permission"].split(".")[-1]
-        return get_object_or_404(Permission, codename=codename)
+
+        try:
+            permission = Permission.objects.get(codename=codename)
+        except (Permission.MultipleObjectsReturned, Permission.DoesNotExist, FieldError) as exc:
+            raise ValidationError(_(str(exc)))
```
mdellweg, line 200:
```suggestion
            raise ValidationError(str(exc))
```
Won't work at all, but the exception should only use translatable strings anyway.

pulpcore/app/viewsets/user.py
```diff
@@ -263,7 +295,10 @@ def create(self, request, group_pk):
             raise ValidationError(
                 _("Please provide one of the following values for User: 'pk', 'id', 'username'")
             )
-        user = get_object_or_404(get_user_model(), **request.data)
+        try:
+            user = User.objects.get(**request.data)
+        except (User.DoesNotExist, FieldError) as exc:
+            raise ValidationError(_(str(exc)))
```
mdellweg, line 301: same here
```suggestion
            raise ValidationError(str(exc))
```

mdellweg (changes requested):
Not sure if i'm the right one to review.
But at least with the gettext stuff i am pretty confident now.

mdellweg (approved):

pulpcore/app/viewsets/user.py
```diff
@@ -274,7 +309,7 @@ def destroy(self, request, group_pk, pk):
         Remove a user from a group.
         """
         group = Group.objects.get(pk=group_pk)
-        user = get_object_or_404(get_user_model(), pk=pk)
+        user = get_object_or_404(User, pk=pk)
```
dkliban, line 312: Is this going to be a problem for galaxy_ng?

fao89: nope, at the beginning of this file (line 22) I declared:
```python 
User = get_user_model()
```

PR #867, opened 2020-08-26
Fix canceled tasks being processed by worker

dralley:
When a "scheduling" job runs on the resource manager, first check
whether its task has been canceled prior to queuing the new job.

The job that runs on the resource manager has a different RQ job ID than
the one which ultimately does the work (which uses the task ID as a job
ID) therefore if util.cancel(task_id) is called on a task that only
exists on the resource manager queue, canceling the RQ job won't do
anything because the job does not yet exist.

Also remove a workaround which is no longer needed.

closes: #7389
https://pulp.plan.io/issues/7389


dralley:
It seems like it's OK to have two jobs with the same ID, and it works in my testing, but I asked for feedback on whether this is a bad idea: https://github.com/rq/rq/issues/1208#issuecomment-681074871

dralley:
So it seems like commit 1 might be problematic?  A couple of tests failed, not sure why exactly... Commit 3 seems to work well though.

dralley:
I dropped the 2nd commit reverting the first to make it easier to review, but we'd use only one of the two strategies

pulpcore/tasking/tasks.py
```diff
@@ -76,6 +74,10 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
     redis_conn = connection.get_redis_connection()
     task_status = Task.objects.get(pk=inner_task_id)
     task_name = func.__module__ + "." + func.__name__
+
+    if task_status.state == "canceled":
```
bmbouter, line 78: I can see how this is helpful, but we already have a mechanism to handle this [here](https://github.com/pulp/pulpcore/blob/master/pulpcore/tasking/worker.py#L104-L109). Since we have code that handles this I don't think we should add more.

pulpcore/tasking/tasks.py
```diff
@@ -15,9 +15,7 @@
 _logger = logging.getLogger(__name__)
 
 
-# workaround for long-running tasks https://github.com/rq/rq/issues/872
-# Pulp tasks should never run more than one Julian year
-TASK_TIMEOUT = 31557600
+TASK_TIMEOUT = -1  # -1 for infinite timeout
```
bmbouter, line 18: :+1:

pulpcore/tasking/tasks.py
dralley: We don't, that's actually the problem.

Our "task" is 2 separate RQ jobs - one for the resource manager which then spawns a second one on a worker.

Only Job#2 shares the same ID as the task PK.  If job#1 is on the resource manager queue when the cancel request goes out, we're going to ask to cancel a job with an ID that does not yet exist.  So what happens is, we cancel the task, it goes to canceled state, it doesn't cancel the RQ job on the resource manager queue because that one has a different ID, it does cancel the other job... that doesn't exist yet, and then the resource manager job is processed and spawns the 2nd job.

dralley: Commit #2 was a strategy of "well, if we make job#1 have the same Job ID, then it won't matter where the task is because it'll get cancelled regardless.  But, it seems like that's a no-no or at least has some side effects that I'm not sure what is the cause of them.

bmbouter: I see thanks for pointing this out. I expect RQ to handle checking if a task is canceled or not and what we're doing is bringing that feature into our application code. The reason I think it's not comes from the `_queue_reserved_task` not having the task_id be the task ID of final task executing (like we had with celery).

The issue with having the `_queue_reserved_task` not use the same task ID is that we can never cancel the orphan_cleanup task while it is running. That's a problem since all tasks are expected to cancel immediately.

You mentioned there were problems with job#1 and job#2 having the same task ID, what are those?

bmbouter: Maybe our data model needs to include in it an unserialized record of the reserved_resource job ID (aka job#1) and when we cancel we cancel both...

dralley: @bmbouter See the test failures

bmbouter: In looking at the cancel implementation it deletes the job itself, so part of the issue here I think is that:

1. The `_queue_reserved_resource` job ID is not being cancelled and yet the actual tasks job Id (job#2) has not yet been created in redis yet...
2. The user calls cancel which deletes the job ID that hasn't been created
3. The `_queue_resserved_resource` job runs
4. The "canceled task" doesn't know it's canceled...

dralley: Yup, all of that sounds correct.

I'm not sure I understand why this is true, though.

> The issue with having the _queue_reserved_task not use the same task ID is that we can never cancel the orphan_cleanup task while it is running. That's a problem since all tasks are expected to cancel immediately.

I don't think we necessarily have any problems with canceling running tasks, and I don't think orphan cleanup is any different apart from the fact that they run on the resource manager itself.

bmbouter: Have we tested the orphan cleanup cancellation? The orphan cleanup is significantly different w.r.t cancellation because of where it runs.

With the new information that the `_queue_reserved_task` does not use the same task ID I don't see how it could work. Confirming that I think is important to guide the right solution in this PR also. What do you think?

dralley: It runs on the resource manager but it still has 2 stages just like everything else.  It just gets re-queued on the resource manager basically.  I haven't tested it but that's what the code does.

dralley: @bmbouter What do you think is our best path forwards?

bmbouter: Interesting...  I see that it does queue the 2nd task, but it actually runs it synchronously. [`is_async=False`](https://github.com/pulp/pulpcore/blob/master/pulpcore/tasking/tasks.py#L90). So orphan cleanup probably does work, but it's strange at best.

My take is that we should integrate cancellation to explicitly track-and-cancel the `_queue_reserved_task` also. To do this I think:

1. Add a new attribute to `Task` named `resource_manager_job_id` and have it never be shown to the user (unserialized).
2. Have the `enqueue_with_reservation` populate that field.
3. Have cancellation cancel both [here](https://github.com/pulp/pulpcore/blob/f06db1eed0dc6e328b8f2e57c44665e10bcaa083/pulpcore/tasking/util.py#L48)
4. Additionally, we probably should issue the KILLKEY for both tasks regardless of state [here](https://github.com/pulp/pulpcore/blob/f06db1eed0dc6e328b8f2e57c44665e10bcaa083/pulpcore/tasking/util.py#L47)

Also as an aside, maybe we can move [this time-delay hack](https://github.com/pulp/pulpcore/blob/f06db1eed0dc6e328b8f2e57c44665e10bcaa083/pulpcore/tasking/util.py#L53-L57) to the kill-thread itself. We shouldn't have the viewset handling those things I think given that the task knows it needs to die.

dralley: @bmbouter Will "lockless Pulp" resdesign this area once again?  If so, do you object to just merging the temporary solution already implemented in commit#1 until we re-evaluate the larger design here?

bmbouter: I perceive lockless Pulp as being much further than 2 months (although I am very in favor of it).

Given that I would like us to not merge this. My reasoning is that I want to avoid picking up task flow-control in our task code when RQ has features to do that for us. Consider a situation when a user performs a mass-cancel. With this fix, the resource-manager will still process each task only to no-op on the first line. This would be both surprising to users and inefficient. Our design uses two RQ tasks** for one Pulp task, so I think our cancel logic should match that. Would you be ok with that other fix type that I'm hoping for?

**: well really it's three because there is the `_release_resources` task also, but that is never canceled.

dralley: I'm perfectly OK with the design, I'd just rather avoid migrations and temporary architectural changes.  If it's going to be 6+ months before we change anything in this area then I don't mind proceeding with the suggested path though.

bmbouter: @dralley thank you for your willingness to go this way for the fix. Are you planning to proceed with this work? Should we continue to use this PR or a new one?

dralley: Updated.

It's weird though, ```job.cancel()``` doesn't work but ```job.delete()``` does.  Not sure why.  In any case, this patch seems to work.

pulpcore/tasking/util.py
```diff
@@ -41,10 +41,14 @@ def cancel(task_id):
         return task_status
 
     redis_conn = connection.get_redis_connection()
+
     job = Job(id=str(task_status.pk), connection=redis_conn)
+    resource_job = Job(id=str(task_status._resource_job_id), connection=redis_conn)
 
     if job.is_started:
         redis_conn.sadd(TASKING_CONSTANTS.KILL_KEY, job.get_id())
```
bmbouter, line 49: Does it make sense (here or in another PR) to just issue the kill key regardless?

I'm concerned about the race condition where either the resource_job or the actual job had not yet started but then starts and the deletion doesn't stop the worker from already receiving the job.

Note your PR isn't introducing this problem, but I think it's a general weakness in our design.

pulpcore/tasking/tasks.py
```diff
@@ -212,16 +211,18 @@ def as_url(r):
...
     if current_job:
-        current_task = Task.objects.get(pk=current_job.id)
-        parent_kwarg["parent_task"] = current_task
+        # set the parent task of the spawned task to the current task ID (same as rq Job ID)
+        parent_kwarg["parent_task"] = Task.objects.get(pk=current_job.id)
+    task_args = (func, inner_task_id, list(resources), args, kwargs, options)
+
+    q = Queue("resource-manager", connection=redis_conn)
+    resource_job = q.enqueue(_queue_reserved_task, args=task_args, job_timeout=TASK_TIMEOUT)
     Task.objects.create(
```
bmbouter, line 220: Regarding the order of queue dispatch relative to Task creation at transactional commit, I remembered an article from I read a long time ago that suggests we do this slightly differently. Check out the `transaction.on_commit` suggestion here? https://testdriven.io/blog/celery-database-transactions/

This is not exactly related to your PR, but in our overall interest to get these race conditions worked out.

pulpcore/app/models/task.py
```diff
@@ -352,6 +352,11 @@ class Task(BaseModel, AutoDeleteObjPermsMixin, AutoAddObjPermsMixin):
         "TaskGroup", null=True, related_name="tasks", on_delete=models.SET_NULL
     )
 
+    # TODO: find a solution that makes this unnecessary
+    # The purpose of this is to enable cancelling the job scheduled on the resource manager
+    # as it has a separate job ID that is not the task ID.
+    _resource_job_id = models.UUIDField(null=True)
```
bmbouter, line 358: I expect this will always be set, so `null=False`? That's also the default I think so this cold be `models.UUIDField()`.

CHANGES/7389.bugfix
```diff
@@ -0,0 +1 @@
+Fixed an edge case where canceled tasks might be processed if Pulp is in a bad state.
```
bmbouter, line 1: Maybe remove the `if Pulp is in a bad state`? I think even under normal operation this design issue would have been experienced.

pulpcore/tasking/util.py
dralley: All we can do without a task blacklist/denylist is to reduce the duration of that window.  Even if we issue the kill key, the killing happens in a background thread that only polls once a second or so.  The original check would have actually handled that situation because when the worker does receive the job, it checked that the task wasn't already canceled first.

pulpcore/tasking/tasks.py
dralley: That's an interesting idea, I'll play around with that a little bit.

dralley: @bmbouter The struggle is that, well, we've just added the resource job ID to the task model, so if we commit the task first then we need to do another update afterwards.  Also in this order if RQ fails to enqueue the task, it won't leave a task record lying around in waiting.

bmbouter: Yes but conversely if this happens to reach the worker prior to the transactional commit reaching postgresql the task will fail here https://github.com/pulp/pulpcore/blob/master/pulpcore/tasking/tasks.py#L77

Also consider that in some situations if rq is run with `SimpleWorker` this enqueue actually becomes a synchronous call. This change means we could never use `SimpleWorker`. Granted we don't do that today, but if we're picking an order I think the other way was better.

pulpcore/tasking/util.py
bmbouter: I don't want to make this PR too complicated, it accomplishes what we came in here to do.

The issue I'm observing (for another PR another day) is that it's not about canceling the task "quickly" it's more about cancelling it at all. With the PR merged we could still:

1. Have job.delete() called without the kill key issued
2. Just after, the worker started processing the job
3. The task runs for a long time without the cancel having notified the worker.

The "one-time check" at the top of any job I don't think solves this fully either because the write-read race condition could still occur in an order where at the time of checking it was not cancelled, the job starts, and just after the cancel transaction is committed. This is the benefit I see in having the kill key involved is that while it doesn't cancel the job in under 1 second it's guaranteed to always cancel a job (which could be running for minutes).

pulpcore/app/models/task.py
dralley: Going forwards, but not for existing Task records.

bmbouter: yeah that's true ... we would need to do that with a data migration. That annoyance aside, I think it's worth it to get the data model accurate.

pulpcore/tasking/tasks.py
dralley: @bmbouter So I made the migration change and then tried to change this order as well, and ran into the problem that now we can't save the Task with a null ```_resource_job_id```.

Filling it in with a fake UUID here temporarily is a step too far for me.  Go back to nullable?

bmbouter: I agree, I believe we want to manufacture the real UUID and specify it as `job_id` kwarg to the _queue_reserved_task itself. This is also how we do it [here also](https://github.com/pulp/pulpcore/blob/master/pulpcore/tasking/tasks.py#L95).

pulpcore/app/models/task.py
bmbouter: Shouldn't the `null=True` be removed?

pulpcore/tasking/tasks.py
```diff
@@ -208,20 +207,31 @@ def as_url(r):
...
+    with transaction.atomic():
+        Task.objects.create(
+            pk=inner_task_id,
+            _resource_job_id=resource_task_id,
+            state=TASK_STATES.WAITING,
+            task_group=task_group,
+            name=f"{func.__module__}.{func.__name__}",
+            **parent_kwarg,
+        )
+        transaction.on_commit(callback)
```
bmbouter, line 235: Oh shiz you did this, hell yeah!

bmbouter (approved):
Thank you! I really appreciate you making it and making it this way.

pulpcore/tasking/util.py
```diff
@@ -41,10 +42,17 @@ def cancel(task_id):
...
+
+    if resource_job.is_started:
```
bmbouter, line 52: Thank you for this also.

bmbouter:
I'm ok w/ this migration rename (for clarity) and I'll try to give my migrations a clearer name from the start next time. +1 to merging this if it passes.

PR #868, opened 2020-08-26
Change docs publish location

dkliban:


template_config.yml
```diff
@@ -26,13 +26,15 @@ plugin_dash: pulpcore
...
 plugin_snake: pulpcore
+publish_docs_to_pulpprojectdotorg: true
 pulp_settings:
   allowed_export_paths:
   - /tmp
   allowed_import_paths:
   - /tmp
 pulpcore_branch: null
 pulpcore_pip_version_specifier: null
+pulpprojectdotorg_key_id: d1a778bda808
```
fao89, line 37: where the key come from? Should it be private?

dkliban: This is a value that is generated by Travis CLI when you encrypt a file[0]. It is then used to decrypt the key file[1].

[0] https://github.com/pulp/pulpcore/blob/master/.travis/pulp-infra.enc
[1] https://github.com/pulp/pulpcore/blob/master/.travis/publish_docs.sh#L5

PR #879, opened 2020-09-01
Re-orders migrations 43 and 44

dkliban:
This will allow cherry-picking migration 43 (old 44) into 3.6.1

[noissue]

### DIFF
--- a/pulpcore/app/migrations/0043_temp_file_artifact_field.py
+++ b/pulpcore/app/migrations/0043_temp_file_artifact_field.py
@@ -8,7 +8,7 @@ import pulpcore.app.models.fields
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('core', '0043_toc_attribute'),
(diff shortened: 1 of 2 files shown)


mdellweg:
Somehow, I am not convinced, this is a good idea. Are we breaking upgrading to the next major release if we generate a new migration with the cherry pick?

dralley:
Not unless we've already released one of those migrations, which I don't *think* we have.

However, I'm not really thrilled about this either.

dkliban:
We have not released these migrations. They are only present on the master branch. I am re-ordering them on the master branch and then plan to cherry-pick this commit to the 3.6 branch.

bmbouter:
I am not in favor of reordering master for a z-release. I recommend not shipping migrations in z-stream and closing this pr. I put this as a discussion point for Sept 4th open floor also.

dkliban:
It is not critical to ship this bug fix with 3.6.1. Closing this PR.

PR #885, opened 2020-09-02
Building changelog for 3.6.2

dkliban:
[noissue]

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,24 @@ Changelog
 
 .. towncrier release notes start
 
+3.6.2 (2020-09-02)
+==================
+REST API
+--------
+
+No significant changes.
+
+
+Plugin API
+----------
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -17,6 +17,24 @@ Changelog
...
+No significant changes.
+
+
+Plugin API
+----------
+
+Bugfixes
+~~~~~~~~
+
+- Remove customized operation_id from OrphansView
```
bmbouter, line 34: Is this not an API fix?

dkliban: It was a fix of the OpenAPI schema. I guess it should be part of the REST API section, but it was already released this way. So we would want to make a separate commit to fix this changelog.

bmbouter (approved):
Thank you for this and release 3.6.3

PR #886, opened 2020-09-03
Allow The ``AccessPolicy.permissions_assignment`` to be null.

bmbouter:
The ``AccessPolicy.permissions_assignment`` can now be null, which some
viewset endpoints may require.

closes #7448


CHANGES/7448.bugfix
```diff
@@ -0,0 +1,2 @@
+The ``AccessPolicy.permissions_assignment`` can now be null, which some viewset endpoints may
```
dkliban, line 1: did you mean to have both a user facing changelog and plugin api changelog entyr?

bmbouter: I did because plugin writers can meaningfully use it as null, but users can also modify the policy too so they may want the same.

docs/plugins/plugin-writer/concepts/rbac/adding_automatic_permissions.rst
```diff
@@ -22,6 +22,10 @@ to be run when new objects are created. These do not run automatically; your mod
 ``pulpcore.plugin.models.AutoAddObjPermsMixin`` on the model as described in the
 :ref:`enabling_new_object_permission_creation` section.
 
+The ``AccessPolicy.permissions_assignment`` attribute is optional because not all AccessPolicy
+objects create objects. If no objects are created by an endpoint, there does not need to be a
+``permissions_assignemnt`` attribute.
```
dkliban, line 27: s/permissions_assignemnt/permissions_assignment/

dkliban (approved):
One small typo in the docs that should be fixed before merging.

bmbouter:
Will do, ty!

PR #891, opened 2020-09-08
Fix tasks stuck in waiting state on redis connection failure

dralley:
closes: #6449
https://pulp.plan.io/issues/6449

### DIFF
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -3,8 +3,9 @@ import time
 import uuid
 from gettext import gettext as _
 
-from django.db import IntegrityError, transaction
+from django.db import IntegrityError
 from django.db.models import Model
+from redis.exceptions import ConnectionError as RedisConnectionError
 from rq import Queue
 from rq.job import Job, get_current_job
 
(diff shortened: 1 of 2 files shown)


pulpcore/tasking/tasks.py
```diff
@@ -215,22 +224,22 @@ def as_url(r):
...
-            _resource_job_id=resource_task_id,
-            state=TASK_STATES.WAITING,
-            task_group=task_group,
-            name=f"{func.__module__}.{func.__name__}",
-            **parent_kwarg,
-        )
-        transaction.on_commit(callback)
+    except RedisConnectionError as e:
+        task.set_failed(e, None)
+        raise
```
dralley, line 243: I figure the most consistent thing to do here is to make sure the task record is created, but make sure it's marked failed if the queue message couldn't be sent.  Alternatively, we could make sure that the failure to send the queue message fails a transaction, thus not creating a task record at all.  In the latter case, the ```transaction.on_commit(callback)``` idea would make sense, but not so much if we always want the Task record.

dralley: Also we don't have to raise the exception.  Raising it returns a 500 server error.  Not raising it would provide the task record but should always be failed if when checked.

dralley: Also should we be releasing the resources directly from ```set_failed``` and ```set_completed```?  Since there's no redis connection clearly the release resources task can't be sent.

bmbouter: I don't think we need to do anything additional to release resources at this point in the call flow. The reserved resources are created in the `_queue_reserved_task` job. Even in that job, if the Redis connection fails it occurs prior to the lock acquisition, so I see no issue in that call flow either.

bmbouter: Can we keep the `transaction.on_commit` callback though? I think having the task dispatch post-commit is still desirable due to race-condition concerns.

dralley: Django uses [autocommit](https://docs.djangoproject.com/en/3.1/topics/db/transactions/#django-s-default-transaction-behavior), which means that every individual modification to the database runs in a transaction (unless you wrap it in a transaction context manager, in which case they run in the same transaction).  So the create() call here has an internal transaction, so there's no point wrapping a single DB op in a transaction.

Also since the I/O is synchronous, there's no race condition here.  create() blocks until the change is committed.

bmbouter (approved):
Let's merge this. It would have been slightly nicer to use a contextmanager to de-duplicate the try/except...set_failed, but if that's done it can be done later. Thank you for fixing this!

pulpcore/tasking/tasks.py
bmbouter: I agree, thank you

PR #894, opened 2020-09-08
Added support for specifying/limiting content-checksums used by Pulp.

### LINKED ISSUE #5216: Allow anonymous browsing of container distributions
It's possible to browse, for example, Pulp file repositories without authenticating by going to <pulp_url>/pulp/content/<repo>, but when we try to do the same for container distributions (i.e., <pulp_url>/pulp/content/<distribution>) we get 403s (since we're unauthenticated in the browser). I assume this is somehow related to pulp_container content guards/access policies but I'm not sure how I can allow unauthenticated browsing of available container images. Any thoughts on how I can enable that workflow?

ggainey:
settings.ALLOWED_CONTENT_CHECKSUMS now drives the other checksum-related
fields of Artifact (DIGEST_FIELDS, COMMON_DIGEST_FIELDS, RELIABLE_DIGEST_FIELDS)

closes #5216


pulpcore/app/settings.py
```diff
@@ -228,6 +228,38 @@
...
+# adding here won't make a new type-of checksum available
+ALL_KNOWN_CONTENT_CHECKSUMS = ("md5", "sha1", "sha224", "sha256", "sha384", "sha512")
+
+# What kinds of checksums is this pulp-instance _allowed to use_ ?
+# NOTE : "sha256"" IS REQUIRED - Pulp will fail to start if it is not found in this set
+# NOTE: specifying checksums that are not listed under ALL_KNOWN_CONTENT_CHECKSUMS will fail
+#       at startup
+# NOTE: If running in a FIPS environment, change this line to the following:
+# ALLOWED_CONTENT_CHECKSUMS = set(("sha1", "sha224", "sha256", "sha384", "sha512"))
+ALLOWED_CONTENT_CHECKSUMS = set(("md5", "sha1", "sha224", "sha256", "sha384", "sha512"))
```
mdellweg, line 242: I thought this kind of information should be in the "other" `settings.py` finally located at `/etc/pulp/settings.py`.

pulpcore/app/settings.py
```diff
@@ -228,6 +228,38 @@
...
+
+if "sha256" not in ALLOWED_CONTENT_CHECKSUMS:
```
mdellweg, line 244: And this test is probably slated for after including dynaconf.

pulpcore/tests/unit/serializers/test_content.py
```diff
@@ -13,19 +14,10 @@ def test_validate_file_checksum(self):
...
-                "sha224": "asdf",
-                "sha256": "asdf",
-                "sha384": "asdf",
-                "sha512": "asdf",
-            },
-        )
+        self.assertEqual(new_data["file"], mock_file)
+        self.assertEqual(new_data["size"], 42)
+        for csum in models.Artifact.DIGEST_FIELDS:
+            self.assertEqual(new_data[csum], "asdf")
```
mdellweg, line 20: Can you add a second loop to test that all excluded algorithms saved "None"?

pulpcore/app/models/content.py
```diff
@@ -132,43 +217,29 @@ def storage_path(self, name):
...
 
     # Reliable digest fields ordered by algorithm strength.
-    RELIABLE_DIGEST_FIELDS = DIGEST_FIELDS[:-3]
+    RELIABLE_DIGEST_FIELDS = reliable_checksums()
 
     def q(self):
         if not self._state.adding:
             return models.Q(pk=self.pk)
-        for digest_name in self.DIGEST_FIELDS:
+        for digest_name in Artifact.DIGEST_FIELDS:
```
mdellweg, line 242: You should not need this change. Class specifiers are avaliable in the instances unless overwritten.

pulpcore/app/models/content.py
```diff
@@ -11,13 +11,98 @@
...
+
+    reliables = (
+        "sha512",
+        "sha384",
+        "sha256",
+    )
+    checksum_cache["reliable_checksums"] = [
+        h for h in reliables if h in settings.ALLOWED_CONTENT_CHECKSUMS
+    ]
+    return checksum_cache["reliable_checksums"]
```
mdellweg, line 105: You asked for comments about the memoization.
Well, first of all i see you calling those functions exactly once at importing the module for the first time. So, what is the memoization for?
Having realized that, it should be sufficient to write a corresponding list comprehension at definition of those constants.
See suggestion there.

pulpcore/app/models/content.py
```diff
@@ -132,43 +217,29 @@ def storage_path(self, name):
...
     # (Better average-case performance in some algorithms with fallback)
-    COMMON_DIGEST_FIELDS = (
-        "sha256",
-        "sha512",
-        "sha384",
-        "sha224",
-        "sha1",
-        "md5",
-    )
+    COMMON_DIGEST_FIELDS = checksums_by_frequency()
```
mdellweg, line 234:
```suggestion
    COMMON_DIGEST_FIELDS = [ algo for algo in ("sha256", "sha512", "sha384", "sha224", "sha1", "md5") if algo in settings.ALLOWED_CONTENT_CHECKSUMS ]
```

mdellweg (changes requested):
Thank you for letting me get a very early look on this **brill** PR.

Running it in the FIPS environment exposed at least one more location where a not allowed hash algorithm is used: pulpcore/app/files.py

docs/settings.rst
```diff
@@ -232,3 +233,26 @@ PROFILE_STAGES_API
...
+
+.. warning::
+    Specifying checksums that are not listed under ``ALL_KNOWN_CONTENT_CHECKSUMS`` will cause
+    Pulp to fail to start.
+
+.. note::
+    If running in a FIPS environment, before starting Pulp change the value to the allowed
+    algorithm-list, something like the following:
+
+    ``ALLOWED_CONTENT_CHECKSUMS = set(("sha1", "sha224", "sha256", "sha384", "sha512"))``
```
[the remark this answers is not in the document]
ggainey: Indeed - I was following the existing code, but I agree we can simplify.

pulpcore/app/models/content.py
ggainey: Heh, fair - personally, I prefer the explicit namespace.

pulpcore/app/models/content.py
ggainey: Ah great - that crystallizes the murky "there is something not right about this" I was feeling last night, thanks.

pulpcore/app/settings.py
mdellweg: ... as dynaconf is loading the on-site-config from whatever specific location (commonly /etc/pulp/settings.py)

pulpcore/app/models/content.py
mdellweg: I know, it's academic, but you could subclass the Artifact and define a different value there. Then only self does
*the-right-thing (tm)*.

ggainey: Fair enough - plus, 'consistency' is more important than 'personal quirk' anyway :)

ggainey:
> Thank you for letting me get a very early look on this **brill** PR.
> 
> Running it in the FIPS environment exposed at least one more location where a not allowed hash algorithm is used: pulpcore/app/files.py

Concur - see changes in files.py with latest repush

docs/settings.rst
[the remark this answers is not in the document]
ggainey: Updated to use the simpler/more-direct syntax

pulpcore/app/settings.py
ggainey: Moved to post-dynaconf to allow site-settings to override

pulpcore/tests/unit/serializers/test_content.py
ggainey: Concur, moved

pulpcore/app/models/content.py
ggainey: Removed the overly-complicated mess to follow suggestion below

pulpcore/app/models/content.py
ggainey: Concur, applied to each of these attributes

docs/settings.rst
```diff
@@ -232,3 +233,26 @@ PROFILE_STAGES_API
...
       Profiling stages is provided as a tech preview in Pulp 3.0. Functionality may not fully work
       and backwards compatibility when upgrading to future Pulp releases is not guaranteed.
+
+.. _allowed-content-checksums:
+
+ALLOWED_CONTENT_CHECKSUMS
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    The set of content-checksums this pulp-instance **allowed to use**. This list is a
+    proper subset of the checksums defined by the Artifact model.
```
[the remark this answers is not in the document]
ggainey: Concur, clarified thie whole section (I think)

pulpcore/tests/unit/serializers/test_content.py
ggainey: Done

pulpcore/app/settings.py
ggainey: Concur, moved post-dynaconf

docs/settings.rst
```diff
@@ -232,3 +233,31 @@ PROFILE_STAGES_API
...
+    .. warning::
+      Specifying checksums that are not listed under ``ALL_KNOWN_CONTENT_CHECKSUMS`` will cause
+      Pulp to fail to start.
+
+.. note::
+
+   If running in a security-conscious environment, before starting Pulp change the value to the
+   allowed algorithm-list, something like the following:
+
+   ``ALLOWED_CONTENT_CHECKSUMS = {"sha1", "sha224", "sha256", "sha384", "sha512"}``
```
bmbouter, line 260: I think we should disinclude sha1 also given that is has been shown to be shattered. https://shattered.io/

daviddavis:
This code looks great to me. Impressive turnaround time on this.

Question: would it be possible to test this out by defining `ALLOWED_CONTENT_CHECKSUMS` in https://github.com/pulp/pulpcore/blob/master/.travis/settings.py.j2?

pulpcore/app/settings.py
```diff
@@ -228,6 +228,20 @@
     },
 }
 
+# What content-identifying checksums algorithms does pulp _know about_?
+# NOTE that models.Artifact must include a field for each possible checksum; simply
+# adding here won't make a new type-of checksum available
+ALL_KNOWN_CONTENT_CHECKSUMS = {"md5", "sha1", "sha224", "sha256", "sha384", "sha512"}
```
bmbouter, line 234: I think we need to move this out of settings.py because it's not safe for users to override this. Instead can this become a constant somewhere, maybe on the Artifact object itself, or perhaps https://github.com/pulp/pulpcore/blob/master/pulpcore/constants.py

docs/settings.rst
```diff
@@ -232,3 +233,31 @@ PROFILE_STAGES_API
...
+    proper subset of the checksums defined by the Artifact model. You may safely list fewer
+    algorithms than the Artifact model supports (although see the warning below regarding ``sha256``),
+    but adding unknown algorithms will cause unexpected behavior.
+
+    .. warning::
+      Due to its use as a primary content-identifier, "sha256"" **IS REQUIRED**. Pulp will
+      fail to start if it is not found in this set.
+
+    .. warning::
+      Specifying checksums that are not listed under ``ALL_KNOWN_CONTENT_CHECKSUMS`` will cause
```
bmbouter, line 252: Can we list the explicit entire set here instead of referring to `ALL_KNOWN_CONTENT_CHECKSUMS`. If we move `ALL_KNOWN_CONTENT_CHECKSUMS` to a constant (I think we need to) user's won't know the values it provides.

pulpcore/app/models/content.py
```diff
@@ -132,38 +138,36 @@ def storage_path(self, name):
...
+    sha224 = models.CharField(max_length=56, null=True, unique=False, db_index=True)
     sha256 = models.CharField(max_length=64, null=False, unique=True, db_index=True)
-    sha384 = models.CharField(max_length=96, null=False, unique=True, db_index=True)
-    sha512 = models.CharField(max_length=128, null=False, unique=True, db_index=True)
+    sha384 = models.CharField(max_length=96, null=True, unique=True, db_index=True)
+    sha512 = models.CharField(max_length=128, null=True, unique=True, db_index=True)
 
     objects = BulkCreateManager()
 
     # All digest fields ordered by algorithm strength.
```
bmbouter, line 150:
```suggestion
    # All available digest fields ordered by algorithm strength.
```

pulpcore/app/models/content.py
```diff
@@ -132,38 +138,36 @@ def storage_path(self, name):
...
-        "sha1",
-        "md5",
-    )
+    DIGEST_FIELDS = [
+        algo
+        for algo in ("sha512", "sha384", "sha256", "sha224", "sha1", "md5")
+        if algo in settings.ALLOWED_CONTENT_CHECKSUMS
+    ]
 
     # All digest fields ordered by relative frequency
```
bmbouter, line 157:
```suggestion
    # All available digest fields ordered by relative frequency
```

pulpcore/app/settings.py
```diff
@@ -228,6 +228,20 @@
...
+
+# What kinds of checksums is this pulp-instance _allowed to use_ ?
+# NOTE : "sha256"" IS REQUIRED - Pulp will fail to start if it is not found in this set
+# NOTE: specifying checksums that are not listed under ALL_KNOWN_CONTENT_CHECKSUMS will fail
+#       at startup
+# NOTE: If running in a FIPS environment, change this line to the following:
+#           ALLOWED_CONTENT_CHECKSUMS = {"sha1", "sha224", "sha256", "sha384", "sha512"}
+#       in your on-site configuration (commonly /etc/pulp/settings.py)
```
bmbouter, line 242: There are various ways to set this, I recommend not naming just one and instead removing the "how to set" recommendation.

docs/settings.rst
```diff
@@ -232,3 +233,31 @@ PROFILE_STAGES_API
...
+      Pulp to fail to start.
+
+.. note::
+
+   If running in a security-conscious environment, before starting Pulp change the value to the
+   allowed algorithm-list, something like the following:
+
+   ``ALLOWED_CONTENT_CHECKSUMS = {"sha1", "sha224", "sha256", "sha384", "sha512"}``
+
+   in your on-site configuration (commonly /etc/pulp/settings.py)
```
bmbouter, line 262: There are various ways to set this, I recommend not naming just one and instead removing the "how to set" recommendation.

You could refer users to the docs on it tho https://docs.pulpproject.org/pulpcore/installation/configuration.html

pulpcore/app/settings.py
```diff
@@ -228,6 +228,20 @@
...
+# NOTE that models.Artifact must include a field for each possible checksum; simply
+# adding here won't make a new type-of checksum available
+ALL_KNOWN_CONTENT_CHECKSUMS = {"md5", "sha1", "sha224", "sha256", "sha384", "sha512"}
+
+# What kinds of checksums is this pulp-instance _allowed to use_ ?
+# NOTE : "sha256"" IS REQUIRED - Pulp will fail to start if it is not found in this set
+# NOTE: specifying checksums that are not listed under ALL_KNOWN_CONTENT_CHECKSUMS will fail
+#       at startup
+# NOTE: If running in a FIPS environment, change this line to the following:
+#           ALLOWED_CONTENT_CHECKSUMS = {"sha1", "sha224", "sha256", "sha384", "sha512"}
```
bmbouter, line 241: Can we remove the FIPS related recommendation L#240 and 241. The available "FIPS" algorithms are different depending on if a user is government or not. I think the safe thing is for us to remain silent on this.

pulpcore/app/settings.py
```diff
@@ -254,3 +267,23 @@
...
+
+unknown_algs = ALLOWED_CONTENT_CHECKSUMS.difference(ALL_KNOWN_CONTENT_CHECKSUMS)
+if unknown_algs:
+    raise ImproperlyConfigured(
+        _(
+            "ALLOWED_CONTENT_CHECKSUMS may only contain algorithms known to pulp - see "
+            "ALL_KNOWN_CONTENT_CHECKSUMS for the allowed list. Unknown algorithms "
+            "provided: {}".format(unknown_algs)
+        )
+    )
```
bmbouter, line 289: Ok this comment is an idea I'm just now having, so it's not in the original story. Let me know if this is too much to request. What if an additional check was made that would:

1) Pick a random Artifact from the db
2) Check that all the checksums present in `ALLOWED_CONTENT_CHECKSUMS` are set and only those are set. If not, the user is changing this variable post-content-being-saved and should be instructed via `ImproperlyConfigured` that they should set it to: {name the set of set attributes} on the tested Artifact.

pulpcore/app/models/content.py
```diff
@@ -132,38 +138,36 @@ def storage_path(self, name):
...
     # (Better average-case performance in some algorithms with fallback)
-    COMMON_DIGEST_FIELDS = (
-        "sha256",
-        "sha512",
-        "sha384",
-        "sha224",
-        "sha1",
-        "md5",
-    )
+    COMMON_DIGEST_FIELDS = [
```
bmbouter, line 159: I think these would be nicer with a property and moved to the module level. I'm read python every day and I'm even a bit unsure of what this is providing. This could become:

It could roughly be:

```
_COMMON_DIGEST_FIELDS = []
for algo in ("sha256", "sha512", "sha384", "sha224", "sha1", "md5")
    if algo in settings.ALLOWED_CONTENT_CHECKSUMS:
        _COMMON_DIGEST_FIELDS.append(algo)

class Artifact(HandleTempFilesMixin, BaseModel):
    <fields>

    @property
    def COMMON_DIGEST_FIELDS(self):
        return _COMMON_DIGEST_FIELDS    
```

Similarly for the other ones.

docs/settings.rst
ggainey: Per following comments and some pondering, I'm just removing my 'helpful' suggestions - security-conscious users know what they need, they don't it from us/here!

pulpcore/app/settings.py
ggainey: Concur, moving to constants.py

pulpcore/app/models/content.py
```diff
@@ -238,6 +243,8 @@ def init_and_validate(file, expected_digests=None, expected_size=None):
 
         if expected_digests:
             for algorithm, expected_digest in expected_digests.items():
+                if algorithm not in hashers:
+                    raise UnsupportedDigestValidationError()
```
bmbouter, line 247: We can continue to do this here, but I think we need this same `raise UnsupportedDigestValidationError` in a new `Artifact.__init__(*args, **kwargs)` call. The reason is that plugin writers are not required to use `init_and_validate` and while they should they don't have to. Also by not having it on the Artifact() instantiation itself when `DeclarativeArtifact` creates in-memory Artifacts in FirstStage of a Stages pipeline the error won't be raised until much later in the pipeline.

If we did it in `Artifact.__init__(...)` itself, we probably wouldn't need it in either of these places then because underneath they both call `Artifact.__init__(...) at some point.

pulpcore/app/models/content.py
ggainey: Oh, good suggestion - changing all three instances

pulpcore/app/models/content.py
ggainey: concur

pulpcore/app/settings.py
ggainey: See prev "removing helpful suggestions"

docs/settings.rst
ggainey: Good suggestion, adding pointer and removing suggestions

pulpcore/app/settings.py
ggainey: See prev RE removing :)

pulpcore/app/settings.py
ggainey: I think this is a reasonable idea - but I think I'd like to see it as a refinement. Can we get an issue opened for it, and if we can fit it into the current crunch it'll be in the next release, and if not, we'll have FIPS-compliance available and won't forget to add that when we can.

pulpcore/app/models/content.py
ggainey: Lots of ways to do this. Part of the issue here (imnsho) is that Black's decision about how list-comprehensions 'should' look is...kind of painful.

In any event - perfectly happy to change for readability.

pulpcore/app/settings.py
bmbouter: That works for me, I added it here: https://pulp.plan.io/issues/7487

Since we can't change settings during functional tests run I don't believe we can reasonably test for https://pulp.plan.io/issues/7487

mdellweg: That test would give a false sense of security. Consider pulp being started with [a,b,c] and an artifact A1 was created. Then a turned bad and it is reconfigured [b,c]. A second artifact A2 is created. Now someone configures [a,b,c] again (which is wrong). This test would only report the error in about half the cases (and you can tweak this ration in every direction).

pulpcore/app/models/content.py
mdellweg: doesn't need to be a property here. You can just assign the value.
Read-onlyness of constants is more of a convention in python anyway.

ggainey: So, @property defines instance-level not class-level - and this change as-is results in a lot of "TypeError: argument of type 'property' is not iterable" errors in the several/many places that take advantage of these entities.

Note that none of these are new - where they exist in Artifact is where they've been for...quite some time, and there is code that relies on them. I can make creating the list be more explicit, but I think further change is more invasive than I'm confortable with, and doesn't actually gain us much. What do you think?

docs/settings.rst
ggainey: Concur

pulpcore/app/settings.py
daviddavis: Yup, we could perhaps do a select like `SELECT * FROM artifacts WHERE a IS NULL or b IS NULL ...`.

ggainey: @mdellweg ah, good point. What we're really going for here is, "once you're up and running for a while, deciding to suddenly limit the available checksums is....Fraught With Peril." This feels like something that we def need to think on more - let's discuss under 7487, if we come up with an answer great, if not oh well - but either way, it's a 'future enhancement' :)

pulpcore/app/models/content.py
bmbouter: If it's causing problems it's not worth it. But can we try what @mdellweg is suggesting as the only other thing we'd try to gain the readability? Basically have it:

```
_COMMON_DIGEST_FIELDS = []
for algo in ("sha256", "sha512", "sha384", "sha224", "sha1", "md5")
    if algo in settings.ALLOWED_CONTENT_CHECKSUMS:
        _COMMON_DIGEST_FIELDS.append(algo)

class Artifact(HandleTempFilesMixin, BaseModel):
    <fields>

    COMMON_DIGEST_FIELDS = _COMMON_DIGEST_FIELDS  
```

ggainey: Was in the process of making this exact change :)  Looks to work, we'll go w/ @mdellweg 's suggestion here

pulpcore/app/settings.py
bmbouter: Excellent points. I revised https://pulp.plan.io/issues/7487 as such w/ attribution

mdellweg:
I can confirm, this peachy PR is passing the `pulp_file` test suite (unit & functional) on a the centos8-fips box.

pulpcore/app/models/content.py
ggainey: Added __init__ to Artifact - left this code in place as well

pulpcore/app/models/content.py
```diff
@@ -132,38 +170,36 @@ def storage_path(self, name):
...
-    RELIABLE_DIGEST_FIELDS = DIGEST_FIELDS[:-3]
+    COMMON_DIGEST_FIELDS = _COMMON_DIGEST_FIELDS
+
+    # Available, reliable digest fields ordered by algorithm strength.
+    RELIABLE_DIGEST_FIELDS = _RELIABLE_DIGEST_FIELDS
+
+    # Digest-fields that are NOT ALLOWED
+    FORBIDDEN_DIGESTS = _FORBIDDEN_DIGESTS
+
+    def __init__(self, *args, **kwargs):
```
ggainey, line 195: @bmbouter @mdellweg More eyes please - is this what we want for Artifact.__init__?

pulpcore/tests/unit/models/test_content.py
```diff
@@ -64,3 +70,13 @@ def test_read_temp_file(self):
             temp_file.save()
 
         assert b"temp file test" in temp_file.file.read()
+
+
+class ArtifactAlgorithmTestCase(TestCase):
+    def test_set_forbidden(self):
```
ggainey, line 76: This test will fail, on a box that has denied md5, only if the remaining code in tyhis PR is broken. How useful is this? (I mean, it helped me find/fix problems in the code, but those fixes are all in place now...)

pulpcore/tests/unit/serializers/test_content.py
```diff
@@ -13,19 +15,21 @@ def test_validate_file_checksum(self):
...
+
+        all_csums = {"md5", "sha1", "sha224", "sha256", "sha384", "sha512"}
+        for csum in all_csums.difference(models.Artifact.DIGEST_FIELDS):
+            self.assertFalse(csum in new_data, f"Found forbidden checksum {csum}")
+
+        # This part of the test will only fire if the system-under-test has forbidden
```
ggainey, line 27: SImilar comment to models/test_content

ggainey:
> This code looks great to me. Impressive turnaround time on this.
> 
> Question: would it be possible to test this out by defining `ALLOWED_CONTENT_CHECKSUMS` in https://github.com/pulp/pulpcore/blob/master/.travis/settings.py.j2?

@daviddavis , I have...no insight here. I've done a lot of local testing by setting in /etc/pulp/settings.py.

mdellweg:
> > This code looks great to me. Impressive turnaround time on this.
> > Question: would it be possible to test this out by defining `ALLOWED_CONTENT_CHECKSUMS` in https://github.com/pulp/pulpcore/blob/master/.travis/settings.py.j2?
> 
> @daviddavis , I have...no insight here. I've done a lot of local testing by setting in /etc/pulp/settings.py.

You would add that to the plugins template_config and regenerate the travis directory.

pulpcore/app/models/content.py
```diff
@@ -213,7 +249,8 @@ def init_and_validate(file, expected_digests=None, expected_size=None):
                 values don't match the digest of the data
             :class:`~pulpcore.exceptions.SizeValidationError`: When the ``expected_size`` value
                 doesn't match the size of the data
-
+            :class:`~pulpcore.exceptions.UnexpectedDigestValidationError`: When the
```
mdellweg, line 252:
```suggestion
            :class:`~pulpcore.exceptions.UnsupportedDigestValidationError`: When the
```

mdellweg: also "...when any of the expected_digests..."

ggainey: Nice catch, updated both init_and_validate methods

pulpcore/app/models/content.py
```diff
@@ -11,13 +13,50 @@
...
+# Available, reliable digest fields ordered by algorithm strength.
+_RELIABLE_DIGEST_FIELDS = []
+for alg in ("sha512", "sha384", "sha256"):
+    if alg in settings.ALLOWED_CONTENT_CHECKSUMS:
+        _RELIABLE_DIGEST_FIELDS.append(alg)
+
+# Digest-fields that are NOT ALLOWED
+_FORBIDDEN_DIGESTS = set(ALL_KNOWN_CONTENT_CHECKSUMS).difference(
+    set(settings.ALLOWED_CONTENT_CHECKSUMS)
+)
```
bmbouter, line 51: ty for all this.

pulpcore/constants.py
```diff
@@ -43,3 +43,9 @@
 
 
 API_ROOT = "pulp/api/v3/"
+
+# What content-identifying checksums algorithms does pulp _know about_?
+# NOTE that models.Artifact must include a field for each possible checksum; simply
+# adding here won't make a new type-of checksum available. sha256 MUST be here,
+# as Pulp relies on it to identify entities.
+ALL_KNOWN_CONTENT_CHECKSUMS = {"md5", "sha1", "sha224", "sha256", "sha384", "sha512"}
```
bmbouter, line 51: :+1:

pulpcore/tests/unit/serializers/test_content.py
```diff
@@ -2,6 +2,8 @@
 
 import mock
 from pulpcore.app.serializers import ArtifactSerializer
+from pulpcore.app import models
```
bmbouter, line 5: The test suite is expected to be run from a third party box and may not have pulpcore packages installed on it, or if they are, they may not be configured the same way. In those environments this import will fail or return incorrect data.

To resolve this, I think do two things:
1) Have the test suite expect that `ALLOWED_CONTENT_CHECKSUMS = {"sha1", "sha224", "sha256", "sha384", "sha512"}`. This is the same approach [used by the import/export tests](https://github.com/pulp/pulpcore/blob/40f7cc3bb28830d9944a9908f971ea7002702b16/pulpcore/tests/functional/api/using_plugin/test_pulpimport.py#L4) which require the user to set settings to specific values for the test suite to pass.
2) Have travis fulfill that requirement by adding in the var here: https://github.com/pulp/pulpcore/blob/ab3e4ebbf4fafa93576dc03b8940752ecc56ca1e/.travis/install.sh#L78

mdellweg: This last thing would need to go into the plugins template_config.yaml to survive regenerating the ci files.

mdellweg: Also i just realize that this is a unit test. And i fail to see why it should be ok to import `ArtifactSerializer` but not `Artifact`.
Is there any context missing?

ggainey: Updated template_config.yml and regen'd. Turns out you cannot use sets and dynaconf together; tweaked ALLOWED_CONTENT_CHECKSUMS to be a list, explicitly used a set() around it wherever I needed to use difference/intersection.

ggainey: Also, yes, as noted by @mdellweg , these are unit-tests not functional-tests, so it's ok to rely on being able to import directly (as they all do)

pulpcore/tests/unit/models/test_content.py
ggainey: Addressed by setting ALLOWED_ template_config.yml to not-include md5 by default.

ggainey:
> > > This code looks great to me. Impressive turnaround time on this.
> > > Question: would it be possible to test this out by defining `ALLOWED_CONTENT_CHECKSUMS` in https://github.com/pulp/pulpcore/blob/master/.travis/settings.py.j2?
> > 
> > 
> > @daviddavis , I have...no insight here. I've done a lot of local testing by setting in /etc/pulp/settings.py.
> 
> You would add that to the plugins template_config and regenerate the travis directory.
> 
> Like here: https://github.com/pulp/pulpcore/blob/master/template_config.yml#L30

And done!

mdellweg:
I was finally (sort of) able to pass the functional testsuite of pulpcore (pagination fails due to remaining repos from earlier tests of the same run) in fips environment.

daviddavis (approved):

pulpcore/app/models/content.py
bmbouter: This could use a docstring that especially calls out the `:raises` parts. Also maybe the main docstring should get an update too?

bmbouter:
@ggainey is there a functional test that can assert that a sync'd content unit does not have 'md5' set?

ggainey:
> @ggainey is there a functional test that can assert that a sync'd content unit does not have 'md5' set?

I only have unit-tests in this commit. Hm.

bmbouter:
> > @ggainey is there a functional test that can assert that a sync'd content unit does not have 'md5' set?
> 
> I only have unit-tests in this commit. Hm.

I think you could likely extend this test to check on the md5 aspects of things pretty easily.  https://github.com/pulp/pulpcore/blob/master/pulpcore/tests/functional/api/test_crd_artifacts.py#L52-L57  wdyt?

ggainey:
> > > @ggainey is there a functional test that can assert that a sync'd content unit does not have 'md5' set?
> > 
> > 
> > I only have unit-tests in this commit. Hm.
> 
> I think you could likely extend this test to check on the md5 aspects of things pretty easily. https://github.com/pulp/pulpcore/blob/master/pulpcore/tests/functional/api/test_crd_artifacts.py#L52-L57 wdyt?

Fine idea - added testing for "did you get md5 back from a valid upload?" and a new test for "trying to upload while specifying md5 as the checksum should fail" (both assuming md5 is not present in ALLOWED_CONTENT_CHECKSUMS"

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -79,8 +79,12 @@ def test_upload_empty_file(self):
     def _do_upload_valid_attrs(self, data, files):
         """Upload a file with the given attributes."""
         artifact = self.client.post(ARTIFACTS_PATH, data=data, files=files)
+        # assumes ALLOWED_CONTENT_CHECKSUMS does NOT contain "md5"
+        self.assertTrue(artifact["md5"] is None, "MD5 {}".format(artifact["md5"]))
         self.addCleanup(self.client.delete, artifact["pulp_href"])
         read_artifact = self.client.get(artifact["pulp_href"])
+        # assumes ALLOWED_CONTENT_CHECKSUMS does NOT contain "md5"
+        self.assertTrue(read_artifact["md5"] is None)
```
bmbouter, line 87: All these tests look greattttt! :+1:

bmbouter:
I'm LGTM once the `Artifact.__init__` docstring is added. Thank you @ggainey !

pulpcore/app/models/content.py
bmbouter: Thank you!

ggainey: Pushed!

bmbouter (approved):
Thank you for this very excellent piece of work. I really appreciate how you brought this together so quickly. :+1: :1st_place_medal: :rocket:

pulpcore/tests/functional/api/test_crd_artifacts.py
ggainey: awesome!

PR #898, opened 2020-09-10
Upgrade dynaconf

fao89:
Fix installer CI issue.

[dynaconf 2.22](https://github.com/rochacbruno/dynaconf/blob/2.2.2/setup.py#L50) is conflicting with [black](https://github.com/psf/black/blob/master/setup.py#L73)

Ref: https://pulp.plan.io/issues/7479


bmbouter (approved):
I'm good with this.

dkliban (approved):

PR #902, opened 2020-09-11
Fix duplicate key error after incomplete sync task

dralley:
closes: #6463
https://pulp.plan.io/issues/6463

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -94,6 +94,10 @@ class Repository(MasterModel):
             pulpcore.app.models.RepositoryVersion: The Created RepositoryVersion
         """
         with transaction.atomic():
+            latest_version = self.versions.latest()
+            if not latest_version.complete:
+                latest_version.delete()
+
             version = RepositoryVersion(
                 repository=self, number=int(self.next_version), base_version=base_version
             )
@@ -108,6 +112,7 @@ class Repository(MasterModel):
             if Task.current():
                 resource = CreatedResource(content_object=version)
                 resource.save()
+
             return version
 
     def finalize_new_version(self, new_version):
--- a/CHANGES/6463.bugfix
+++ b/CHANGES/6463.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug where creating an incomplete repository version (via canceled or failed task) could cause future operations to fail.


pulpcore/app/models/repository.py
```diff
@@ -105,9 +109,6 @@ def new_version(self, base_version=None):
                 # now add any content that's in the base_version but not in version
                 version.add_content(base_version.content.exclude(pk__in=version.content))
 
-            if Task.current():
-                resource = CreatedResource(content_object=version)
-                resource.save()
```
dralley, line 110: This doesn't really belong here because the version hasn't been "created" until it has been completed.  It makes more sense in ```__exit__```  IMO.

pulpcore/app/models/repository.py
```diff
@@ -810,6 +811,10 @@ def __exit__(self, exc_type, exc_value, traceback):
                     self.repository.save()
                     self.save()
                     self._compute_counts()
+
+                    if Task.current() and not hasattr(self.repository, '_user_hidden'):
```
dralley, line 815: Yes, this is hacky : /  But so, imo, is having the entire method be duplicated.

pulpcore/app/models/repository.py
```diff
@@ -810,6 +811,10 @@ def __exit__(self, exc_type, exc_value, traceback):
                     self.repository.save()
                     self.save()
                     self._compute_counts()
+
+                    if Task.current() and not getattr(self.repository, "_user_hidden", False):
```
dkliban, line 815: Could we add some plugin writer docs about this property?

dralley: I'm not sure to what degree we actually want to advertise this, I kept it underscore-prefixed under the assumption that it would be semi-private.  If you think it should be documented though, I can do that.

dkliban: i don't feel strongly about it. I dont' even know of the best place to document it.

dralley: @bmbouter, what do you think?

pulpcore/app/models/repository.py
ipanova: yeah I agree, the question is - what is a better choice hacky code or duplicated code. Theoretically this does not really belong to pulpcore since sub_repos are only pure pulp_rpm concept ( for now?)

dralley: I think it basically comes down to whether we want to have to remember to copy every change to this function to the RPM plugin, every time, at risk of bad things happening if we don't.

ipanova: agreed.

dkliban (approved):

ipanova (approved):

pulpcore/app/models/repository.py
bmbouter: I think we need to either make this official or not use it. We need to continue to keep the plugin API official, documented, and versioned. To introduce a callback like that we would need to:

* rename it form _user_hidden to user_hidden, to make it an official hook with a docstring
* Create a base implementation and allow plugins to customize something that is already there
* Not have the default carried here
* Add it into the plugin writer docs somewhere. It can be short, I don't want to ask for so many docs given we have gaps already in so many areas. I want to avoid creating willful new gaps though.

My apologies for the late-in-the-game review.

bmbouter (changes requested):
left requested changes in comment

pulpcore/app/models/repository.py
dralley: I think my hesitation is that we don't have any other plugin that would forseeably use this feature, and if one were to appear, we might want to design better implementation than this.  Honestly that applies to the entire concept of subrepos in general and was part of the reason we didn't put the concept into core way back when.

Would you consider it sufficient to write a detailed-ish description of what and why and commenting the code inline where applicable, or linking to the issue where it is described, along with a disclaimer that "if you want to do something like this, stop and tell us, because we should come up with something better"?

bmbouter: I hear you on the hesitation to add an official call interface when only one plugin is using it. I share that concern. My conclusion is different though. Generally the plan for adding call interfaces to `pulpcore.plugin` is that it's to be used by two or more plugins. For example, pulp_container carries a lot of code, but it's been withheld from the plugin API until it is. By that measure, until this interface is used by two plugins we would not be adding it here. With RPM being the only user can we do either one or two things:

1) keep all the customization in RPM for now until it's used by two or more plugins
2) decompose this call into several supporting functions that would allow a plugin to not carry so much duplicated code.

I'm uncomfortable with soft or unofficial call interfaces because we've worked so hard to get out of that situation with Pulp2. So I'll counter-propose with these alternative options instead. What do you think about these options?

dralley: @bmbouter What I could do is retain the duplicated code, and just add the change in both places.

That does however mean that creation of the ```created_resource``` would need to stay in ```.new_version()```, because we have no ability to override ```__exit__``` from the plugin side.  That wouldn't be the worst thing (it is how things are currently after all) although it doesn't make quite as much sense as having it in ```__exit__``` personally.  But I'm fine with it if we need to do so to avoid changing pulpcore.

bmbouter: +1 to keeping the duplicated code since it's just in one plugin. I'm not understanding what prevents us from moving the `CreatedResource` to `__exit__` here too. I'm confused.

dralley: ```__exit__``` is on ```RepositoryVersion``` which plugins don't override.  The RPM plugin needs to disable creating ```CreatedResource``` when they make an new repository version for a subrepo, but the only two ways to do that are:

1. Have a marker on the Repository which the code in ```RepositoryVersion``` can look for (the approach currently in this PR) to disable saving the ```CreatedResource```
2.  Have the ```CreatedResource``` created by a piece of code that the plugin has the ability to override (what we were doing before)

bmbouter: oic. Thank you for explaining this to me. Yeah I recommend either make the configuration point official so sublasses can control it (and maybe name it `create_created_resource_record` or some self-descriptive name instead of _user_hidden) or leave the changes for this fix in the plugin for now.

bmbouter (approved):
Thanks!

PR #904, opened 2020-09-14
Pin drf-spectacular

fao89:
[noissue]

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ djangorestframework~=3.10.3
 djangorestframework-queryfields~=1.0.0
 drf-access-policy~=0.6.2
 drf-nested-routers~=0.91.0
-drf-spectacular>=0.9.12
+drf-spectacular==0.9.13
 dynaconf>=2.2,<4.0
 gunicorn>=19.9,<20.1
 jinja2
--- a/pulpcore/openapi/__init__.py
+++ b/pulpcore/openapi/__init__.py
@@ -9,13 +9,17 @@ from drf_spectacular.plumbing import (
     ResolvedComponent,
     build_basic_type,
     build_parameter_type,
+    build_root_object,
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -12,7 +12,7 @@ djangorestframework~=3.10.3
 djangorestframework-queryfields~=1.0.0
 drf-access-policy~=0.6.2
 drf-nested-routers~=0.91.0
-drf-spectacular>=0.9.12
+drf-spectacular==0.9.13
```
mdellweg, line 15: #906 contains all but this line. But given the way we now very specifically tweak the spectacular pluming, it might be a good idea to pin this to a single version.

bmbouter: Does drf-spectafular use semver? If they do we should be safe in using `~=`. I don't understand the specific of this tweak tho.

daviddavis: Their [0.9.13 release](https://github.com/tfranzel/drf-spectacular/blob/master/CHANGELOG.rst#0913-2020-09-13) included breaking changes so I think they don't use semver or maybe it's because they haven't reached 1.0 yet.

bmbouter: Yeah I looked and they don't declare anything w.r.t semver. We should ask them at some point ... but later. +1 to the version pin, it's the safe thing to do.

daviddavis: I wonder if we should also make this change against master too. If we don't, then 3.7 will go out with `drf-spectacular>=0.9.12` which could be problematic if a user has 0.9.12.

mdellweg: The change that is breaking us here is following the specifications more strictly. So they might have declared it a bugfix even with semver. It is the special way in which we don't follow the specs to get a nicer bindings experience. At some point we should try to fix this operationId thing in the bindings templates.

PR #906, opened 2020-09-14
Addressing drf-spectacular 0.9.13 breaking changes

fao89:
[noissue]


fao89:
@pulp/core without this fix, we don't have CI: https://travis-ci.com/github/pulp

dkliban (approved):

daviddavis:
Will users potentially hit this problem? Like if a user installs 3.6 today, I imagine they will get drf-spectacular 0.9.13 which will cause them to hit a problem. If this is the case, we should add a changelog so users can see we have fixed the problem in the next release of pulpcore.

dkliban:
Yes, users will hit this problem if they install today.

PR #914, opened 2020-09-16
Upload chunks as separate files

lubosmj:
closes #4498


daviddavis:
I think the main problem with this solution is that it's tied to S3 and we support a variety of other storage backends besides S3 (e.g. Azure). I think the option of where to store upload chunks should be to either use the default storage backend or to override that and use the filesystem.

pulpcore/app/models/upload.py
```diff
@@ -32,32 +29,65 @@ def append(self, chunk, offset, sha256=None):
...
+    is represented in the migration in this way:
 
-        self.chunks.create(offset=offset, size=len(chunk))
+        field=models.FileField(
+            storage=pulpcore.app.models.storage.FileSystem('/var/lib/pulp/')
+        )
 
+    Because of that, we are not able to dynamically change the storage used for chunked uploads
+    without applying additional migrations after performing relevant changes to the settings file.
+    Support for a callable storage is new in Django 3.1.
```
daviddavis, line 57: With this limitation in django 2, it's creating a rather complex solution to have a separate storage setting. I'm inclined to ditch the setting altogether and just store chunks in the default storage.

ipanova:
> I think the main problem with this solution is that it's tied to S3 and we support a variety of other storage backends besides S3 (e.g. Azure). I think the option of where to store upload chunks should be to either use the default storage backend or to override that and use the filesystem.
I agree that we should not be tied to S3 only but use the default storage backend. I am not so sure about idea of using the local fs  because this will be an issue for, for example when having 100 docker puh operations simultaneously whereas layers are quire big and it might happen that local fs won't be enough

daviddavis:
> I agree that we should not be tied to S3 only but use the default storage backend. I am not so sure about idea of using the local fs because this will be an issue for, for example when having 100 docker puh operations simultaneously whereas layers are quire big and it might happen that local fs won't be enough

I agree but if we give users the choice of where to store their upload chunks, I don't think we should prevent them from using a filesystem.

mdellweg:
I thought it would just use the temporary file storage.

daviddavis:
The problem with just using temporary file storage is that it's horribly inefficient in some cases. It requires that Pulp downloads the chunks, assemble the files, and then re-uploads the chunks back into S3. This is after having already having uploaded the chunks into S3 in the first place. And it also presents the same problem @ipanova outlined in that users could potentially run out of filesystem space.

daviddavis:
I agree though that using temporary storage only for now makes sense (or at least until we're on django 3.1 and can offer some configuration options). I would document the issues I describe above though so users are aware.

lubosmj:
@daviddavis, speaking of the option for temporary storage (i.e. FileSystemStorage), users would have to deal with migrations again, as I mentioned in the documentation.

The only thing I can actually do, is to note that `In Pulp, we do support only default_storage and FileSystemStorage for storing chunked uploads; other options are not guaranteed to work properly`. Yet, I am not sure if this is you what you meant by your last comment. And of course, I will remove the wrapper around the S3 storage.

mdellweg:
Sorry to say it like that, but having the user create site-local migrations is not an option at all.
We struggle a lot not to generate or cherry-pick migrations on release branches, already.

daviddavis:
@lubosmj my last comment is that I think we should ditch the storage setting altogether for now and just use default_storage for chunk uploads.

@mdellweg I'm not sure what you mean by users creating "site-local migrations"

mdellweg:
@daviddavis referring to this: https://github.com/pulp/pulpcore/pull/914/files#diff-f47f8ca652dac550c8bc9d449ca0d253R165

daviddavis:
What happens to users that have upload chunks in their database and then they upgrade?

docs/workflows/upload-publish.rst
```diff
@@ -49,7 +49,7 @@ This queues a task that creates an artifact, and the upload gets deleted and can
 
 Putting this altogether, here is an example that uploads a 1.iso file in two chunks::
 
-   curl -O https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/file-large/1.iso
+   curl -O https://fixtures.pulpproject.org/file-large/1.iso
```
daviddavis, line 52: :rocket:

lubosmj:
> What happens to users that have upload chunks in their database and then they upgrade?

They could probably lose track of uploads since the file field was deleted by the migration. Will users perform chunked uploading during the upgrade by any chance? I thought that once an upload is finished, all users create an artifact from it; then, there should not be any uploads left behind.

I may modify the migration to move all uploads to the directory used for chunked uploads. Then, uploaded files will be referenced by UploadChunks models instead of Upload models.

daviddavis:
@lubosmj I brought this topic up at the pulpcore team meeting. The consensus was to add a release note saying that users should delete all uploads before upgrading. So I think we want a XXXX.removal saying that upload chunks are being moved from the filesystem to the default storage the user has configured and users need to delete any chunked uploads before upgrading.

mdellweg:
Honestly i understood that we expect the administrator to clean the diskspace after the migration.
But at least we need a note that pending uploads will not survive the upgrade.

pulpcore/app/models/upload.py
```diff
@@ -32,32 +27,49 @@ def append(self, chunk, offset, sha256=None):
...
+class UploadChunk(HandleTempFilesMixin, BaseModel):
     """
     A chunk for an uploaded file.
 
+    At the moment, it is not possible to dynamically change the storage used for uploaded chunks
+    without creating additional site-local migrations. Due to this, each uploaded chunk is stored
+    in the default file storage.
+
+    Support for callable storage was added in Django 3.1 and Pulp uses the Django 2.2 LTS version.
+    Refer to https://docs.djangoproject.com/en/3.1/topics/files/#using-a-callable.
```
daviddavis, line 50: This text is great but I would probably just put it in the issue and not have it live in our code.

pulpcore/app/models/storage.py
```diff
@@ -96,6 +96,20 @@ def _save(self, name, content, max_length=None):
         return str(name).replace("\\", "/")
 
 
+def get_file_path(directory, identifier):
```
daviddavis, line 99: The function name and docstring seem a bit two generic. What about `get_sha256_path` or something?

daviddavis:
Overall, looks good to me. :+1: 

@mdellweg or @ipanova could you also take a look?

pulpcore/app/models/storage.py
daviddavis: Also, I lean towards maybe not even having a function for such a small bit of logic.

ipanova: i agree. i would not extra this into a separate function

ipanova (approved):
looks good to me, minor comments that should not be considered as a blocker.

pulpcore/app/models/upload.py
```diff
@@ -32,32 +27,49 @@ def append(self, chunk, offset, sha256=None):
...
 
-        with self.file.open(mode="r+b") as file:
-            file.seek(offset)
-            file.write(chunk_read)
-
-        self.chunks.create(offset=offset, size=len(chunk))
+        upload_chunk = UploadChunk(
+            upload=self, offset=offset, size=len(chunk), sha256=current_sha256
+        )
+        upload_chunk.file.save("", ContentFile(chunk_read))
```
ipanova, line 38: why do you have empty string here?

ipanova: i guess we do notreally care what the name of the file for the uploaded chunk

pulpcore/app/models/storage.py
mdellweg: What about `get_hashed_file_path`?

pulpcore/app/models/upload.py
```diff
@@ -32,32 +27,49 @@ def append(self, chunk, offset, sha256=None):
...
 
+    def storage_path(self, name):
+        """
+        Callable used by FileField to determine where the uploaded file should be stored.
+
+        Args:
+            name (str): Original name of uploaded file. It is ignored by this method because the
+                pulp_id is used to determine a file path instead.
+        """
+        return storage.get_upload_chunk_file_path(self.sha256)
```
mdellweg, line 69: Should this really use digests? What if two users upload the same chunk? Will they be deduplicaated or will one of them see a "415 cannot upload that chunk"?

pulpcore/app/models/upload.py
lubosmj: I used the same technique like we use in PulpTemporaryFile. A chunk's filename does not depend on the provided filename, but rather on pulp_id which is known when saving the chunk.

CHANGES/4498.removal
```diff
@@ -0,0 +1,2 @@
+The local file system directory used for uploaded chunks was changed and moved to the default
+storage instead. Users are encouraged to remove all uploaded files before applying this change.
```
daviddavis, line 2: small change: 

s/remove all uploaded/remove all uncommitted uploaded/

daviddavis:
I thought we talked about using pulp_id or uuid to store the chunks in storage and not sha256?

daviddavis:
Oh, I see now you are using `pulp_id_str` to put it in a nested dir. I think this is probably overkill. I don't think that there will be more than a 100 chunks at any one time.

pulpcore/app/models/upload.py
```diff
@@ -32,32 +27,39 @@ def append(self, chunk, offset, sha256=None):
...
 
-class UploadChunk(BaseModel):
+class UploadChunk(HandleTempFilesMixin, BaseModel):
     """
     A chunk for an uploaded file.
 
     Fields:
 
+        file (fields.ArtifactFileField): A file where the uploaded chunk is stored.
         upload (models.ForeignKey): Upload this chunk belongs to.
```
ipanova, line 46: please add `sha256`

pulpcore/app/models/upload.py
```diff
@@ -32,32 +27,39 @@ def append(self, chunk, offset, sha256=None):
...
+        Callable used by FileField to determine where the uploaded file should be stored.
+
+        Args:
+            name (str): Original name of uploaded file. It is ignored by this method because the
+                pulp_id is used to determine a file path instead.
+        """
+        return storage.get_upload_chunk_file_path(self.pulp_id)
+
+    file = fields.FileField(null=False, upload_to=storage_path, max_length=255)
+    sha256 = models.CharField(max_length=64, null=True)
```
ipanova, line 62: what is the purpose of this field if it is not being used when UploadChunk is created? please whether remove it completely or use it during the creation of upload chunk.

lubosmj:
@daviddavis , after the latest changes, nested directories are not created. However, chunks file names are still equal to pulp_id.

daviddavis:
Cool, I would probably get rid of the `get_upload_chunk_file_path` function too but if you think it's useful to have, then that's fine.

daviddavis (approved):
Added one optional comment. Looks good enough to merge to me. 

Thanks @lubosmj!

lubosmj:
Yes, I think that it is useful to have that function!

ipanova (approved):

PR #917, opened 2020-09-18
Fixed failing filesystem export due to undefined validate_path method

daviddavis:
Just removing the call since we're already validating the path in the serializer. Also, adding some tests so we don't break this functionality again later.

Required PR: https://github.com/pulp/pulp_file/pull/429

fixes #7521


goosemania (approved):
Thanks, and nice tests, easy to read!

PR #918, opened 2020-09-18
Added check for ALLOWED_CONTENT_CHECKSUMS that Artifacts are missing

daviddavis:
fixes #7487

### DIFF
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -15,6 +15,7 @@ from importlib import import_module
 from pkg_resources import iter_entry_points
 
 from django.core.exceptions import ImproperlyConfigured
+from django.db import connection
 
 from pulpcore import constants
 
@@ -281,3 +282,21 @@ if unknown_algs:
             "provided: {}".format(unknown_algs)
         )
     )
+for checksum in ALLOWED_CONTENT_CHECKSUMS:
+    # can't import Artifact here so use a direct db connection
+    try:
+        with connection.cursor() as cursor:
+            cursor.execute(f"SELECT count(pulp_id) FROM core_artifact WHERE {checksum} IS NULL")
+            row = cursor.fetchone()
+    except Exception:
+        # our check could fail if the table hasn't been created yet or we can't get a db connection
+        pass
+    else:
+        if row[0] > 0:
+            raise ImproperlyConfigured(
+                _(
+                    "ALLOWED_CONTENT_CHECKSUMS cannot contain checksum '{}' since checksums can "
+                    "only be removed and not added to ALLOWED_CONTENT_CHECKSUMS in existing "
+                    "deployments."
+                ).format(checksum)
+            )
--- a/CHANGES/7487.feature
+++ b/CHANGES/7487.feature
@@ -1,0 +1,2 @@
+Added check to prevent users from adding checksums to ``ALLOWED_CONTENT_CHECKSUMS`` if there are
+Artifacts without those checksums.


pulpcore/app/settings.py
```diff
@@ -281,3 +282,15 @@
...
         )
     )
+for checksum in ALLOWED_CONTENT_CHECKSUMS:
+    # can't import Artifact here so use a direct db connection
+    with connection.cursor() as cursor:
+        cursor.execute(f"SELECT count(pulp_id) FROM core_artifact WHERE {checksum} IS NULL")
+        row = cursor.fetchone()
+        if row[0] > 0:
+            raise ImproperlyConfigured(
+                _(
```
bmbouter, line 292: I think the i8ln wants to have the replacement occur on the returned value from `_()`. This may work when strings are collected perhaps tho?

bmbouter (approved):
This is great thanks!

bmbouter (approved):
This looks good to me. Thanks!

mdellweg (approved):

pulpcore/app/settings.py
```diff
@@ -281,3 +282,20 @@
...
+            cursor.execute(f"SELECT count(pulp_id) FROM core_artifact WHERE {checksum} IS NULL")
+            row = cursor.fetchone()
+    except Exception:
+        # our check could fail if the table hasn't been created yet or we can't get a db connection
+        pass
+    else:
+        if row[0] > 0:
+            raise ImproperlyConfigured(
+                _(
+                    "ALLOWED_CONTENT_CHECKSUMS cannot contain checksum '{}' since there are "
```
ipanova, line 298: The way it is phrased right now, it might give an impression that Artifacts have some issue because they miss information about X checksum. Since this will be a user facing error, how about we re-phrase and give  a higher level explanation:
"ALLOWED_CONTENT_CHECKSUMS cannot contain checksum '{}' since there is already content in the DB and ALLOWED_CONTENT_CHECKSUMS cannot be altered" ( or something along the lines)

ipanova (approved):

pulpcore/app/settings.py
mdellweg: I think, what it tried to capture was that it is actually ok to reduce the set of allowed checksums, but never ok to add some back.
And given the fact that tomorrow one of those might be considered to not be collision free anymore, I think this is an important feature.

daviddavis: Agreed. I've updated the message.

pulpcore/app/settings.py
```diff
@@ -281,3 +282,21 @@
...
+    except Exception:
+        # our check could fail if the table hasn't been created yet or we can't get a db connection
+        pass
+    else:
+        if row[0] > 0:
+            raise ImproperlyConfigured(
+                _(
+                    "ALLOWED_CONTENT_CHECKSUMS cannot contain checksum '{}' since there is "
+                    "already content in the database and checksums cannot be added to "
+                    "ALLOWED_CONTENT_CHECKSUMS (only removed)."
```
mdellweg, line 300:
```suggestion
                    "ALLOWED_CONTENT_CHECKSUMS cannot contain checksum '{}' since checksums can only be removed and not added to "
                    "ALLOWED_CONTENT_CHECKSUMS in existing deployments."
```

My take: From a user perspective the _why_ should not matter too much. It is simply not allowed.
That is what I would say. Maybe it can be articulated better.

PR #922, opened 2020-09-21
Remove orphan content in a loop to clean up subrepos' content as well.

goosemania:
closes #7460
https://pulp.plan.io/issues/7460

### DIFF
--- a/pulpcore/app/tasks/orphan.py
+++ b/pulpcore/app/tasks/orphan.py
@@ -32,26 +32,33 @@ def queryset_iterator(qs, batchsize=2000, gc_collect=True):
 def orphan_cleanup():
     """
     Delete all orphan Content and Artifact records.
+    Go through orphan Content multiple times to remove content from subrepos.
     This task removes Artifact files from the filesystem as well.
-    """
-    content = Content.objects.filter(version_memberships__isnull=True).exclude(
-        pulp_type=PublishedMetadata.get_pulp_type()
-    )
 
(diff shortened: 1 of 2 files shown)


pulpcore/app/tasks/orphan.py
```diff
@@ -32,26 +32,33 @@ def queryset_iterator(qs, batchsize=2000, gc_collect=True):
...
-        c.delete()
+    while True:
+        content = Content.objects.filter(version_memberships__isnull=True).exclude(
+            pulp_type=PublishedMetadata.get_pulp_type()
+        )
+        content_count = content.count()
+        if not content_count:
+            break
+
+        progress_bar.total += content_count
```
goosemania, line 55: I could do:
```
if not content:
    break

progress_bar.total += content.count()
```

I guess because of all the queryset issues I saw in the past, I'm hesitant to use any non-queryset specific operations on querysets because sometimes they behave differently from what one would expect :) The bool() should be fine though :)

daviddavis: I prefer `not content.exists()` over `not content`. But I think what you have here is fine since it uses only one query to check for content and count it.

daviddavis:
This LGTM. @bmbouter or @dralley maybe you could also review?

dralley (approved):

bmbouter (approved):
This looks fine to me. Thank you!

PR #924, opened 2020-09-22
Fix typos

melcorr:
[noissue]

I shared docs with wibbit
& saw some typos
I tried to also add some consistency
to the TaskGroup section

### DIFF
--- a/docs/plugins/plugin-writer/concepts/index.rst
+++ b/docs/plugins/plugin-writer/concepts/index.rst
@@ -5,7 +5,7 @@ Plugin Concepts
 
 Like the Pulp Core itself, all Pulp Plugins are Django Applications, and could be created like any
 other Django app with ``pulpcore-manager startapp <your_plugin>``. However, instead of writing all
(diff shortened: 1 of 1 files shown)


docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -106,7 +106,7 @@ not end up being artifacts in the end. To tackle this, one should use ``PulpTemp
         temp_file.delete()
 
 When dealing with a clustered deployment, different pulp services are not guaranteed to share a
-common filesystem (like /usr/share/pulp). ``PulpTemporaryFile`` is the alternative for creating
+common file system (like /usr/share/pulp). ``PulpTemporaryFile`` is the alternative for creating
```
ipanova, line 109: I think we mostly write this as one word. Apparently both versions are correct https://en.wikipedia.org/wiki/File_system just weird to see it written separated

melcorr: Haha. And for me it is weird to see it together. I will revert. Thanks for the explanation :)

ipanova (approved):

PR #927, opened 2020-09-22
Provide a way to get PK from resource URI without evaluating query

dralley:
re: #7483
https://pulp.plan.io/issues/7483


daviddavis:
The `extract_pk` method looks good to me. Is it possible to reuse it in `get_resource`?

dralley:
@daviddavis I was thinking about that earlier and decided that it couldn't be, the purpose is slightly orthogonal.  The is code should only ever return a PK or a failure, the other code should fall back to doing a multi-attribute search.  Re-using it would mean re-doing part of the work.

daviddavis (approved):

PR #932, opened 2020-09-24
Enable on demand sync for RepositoryVersionDistribution

fao89:
https://pulp.plan.io/issues/7568
closes #7568

### DIFF
--- a/CHANGES/7568.bugfix
+++ b/CHANGES/7568.bugfix
@@ -1,0 +1,1 @@
+Enable content streaming for RepositoryVersionDistribution
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -442,7 +442,12 @@ class Handler:
             except ObjectDoesNotExist:
                 pass
             else:
-                return self._serve_content_artifact(ca, headers)
+                if ca.artifact:
+                    return self._serve_content_artifact(ca, headers)
+                else:
+                    return await self._stream_content_artifact(
+                        request, StreamResponse(headers=headers), ca
+                    )
 
         if distro.remote:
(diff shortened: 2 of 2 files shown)


pulpcore/content/handler.py
```diff
@@ -442,7 +442,12 @@ async def _match_and_stream(self, path, request):
             except ObjectDoesNotExist:
                 pass
             else:
-                return self._serve_content_artifact(ca, headers)
+                if ca.artifact:
+                    return self._serve_content_artifact(ca, headers)
+                else:
+                    return await self._stream_content_artifact(
```
ipanova, line 448: :+1:

mdellweg:
Nice.
Do we have tests in that area? I doubt that it can easily be tested with pulp_file.

CHANGES/7568.bugfix
```diff
@@ -0,0 +1 @@
+Enable on demand sync for RepositoryVersionDistribution
```
ipanova, line 1: how about "enable content streaming.."

ipanova:
> Nice.
> Do we have tests in that area? I doubt that it can easily be tested with pulp_file.

I don't think this can be tested with the pulp_file since it has a publication, it should be a plugin that has distribution only, like pulp_container just not subclassed handler. I think pulp_ansible uses directly pulpcore's content handler

fao89:
> Nice.
> Do we have tests in that area? I doubt that it can easily be tested with pulp_file.

it is being tested here: https://github.com/pulp/pulp_ansible/pull/373

ipanova (approved):

bmbouter (approved):
This looks good to me. Thanks @fao89 !

PR #933, opened 2020-09-25
Set ProgressReport states to canceled when task is canceled

dralley:
closes: #4921
https://pulp.plan.io/issues/4921


dralley:
Should we add a migration to update existing progress reports?  I don't think it's necessary but it's something that we could do.

dkliban:
@dralley Thanks for the fix! I don't think it's worth the trouble to write a migration.

ipanova (approved):

PR #934, opened 2020-09-25
Get rid of failure message on task cancellation

dralley:
_release_resources() task checks whether task status is still set to
"running". In the case of cancelation, that is still true if the
transaction hasn't been committed yet.

closes: #4559
https://pulp.plan.io/issues/4559

### DIFF
--- a/pulpcore/tasking/util.py
+++ b/pulpcore/tasking/util.py
@@ -37,14 +37,18 @@ def cancel(task_id):
 
     if task_status.state in TASK_FINAL_STATES:
         # If the task is already done, just stop
-        msg = _("Task [{task_id}] already in a completed state: {state}")
-        _logger.info(msg.format(task_id=task_id, state=task_status.state))
+        msg = _("Task [{task_id}] already in a final state: {state}")
+        _logger.debug(msg.format(task_id=task_id, state=task_status.state))
         return task_status
 
-    redis_conn = connection.get_redis_connection()
+    _logger.info(_("Canceling task: {id}").format(id=task_id))
 
+    redis_conn = connection.get_redis_connection()
     job = Job(id=str(task_status.pk), connection=redis_conn)
     resource_job = Job(id=str(task_status._resource_job_id), connection=redis_conn)
+
+    task_status.state = TASK_STATES.CANCELED
+    task_status.save()
 
     if job.is_started:
         redis_conn.sadd(TASKING_CONSTANTS.KILL_KEY, job.get_id())
@@ -59,16 +63,13 @@ def cancel(task_id):
     time.sleep(1.5)
 
     with transaction.atomic():
-        task_status.state = TASK_STATES.CANCELED
         for report in task_status.progress_reports.all():
             if report.state not in TASK_FINAL_STATES:
                 report.state = TASK_STATES.CANCELED
                 report.save()
-        task_status.save()
         _delete_incomplete_resources(task_status)
         task_status.release_resources()
 
-    _logger.info(_("Task canceled: {id}.").format(id=task_id))
     return task_status
 
 
--- a/pulpcore/tasking/worker.py
+++ b/pulpcore/tasking/worker.py
@@ -6,6 +6,8 @@ import sys
 import threading
 import time
 
+from gettext import gettext as _
+
 from rq import Queue
 from rq.worker import Worker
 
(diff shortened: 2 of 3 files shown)


pulpcore/tasking/util.py
```diff
@@ -41,11 +41,15 @@ def cancel(task_id):
...
 
-    redis_conn = connection.get_redis_connection()
+    _logger.info(_("Canceling task: {id}").format(id=task_id))
 
+    redis_conn = connection.get_redis_connection()
     job = Job(id=str(task_status.pk), connection=redis_conn)
     resource_job = Job(id=str(task_status._resource_job_id), connection=redis_conn)
 
+    task_status.state = TASK_STATES.CANCELED
+    task_status.save()
```
dralley, line 51: Commit the change of task status before the running task is canceled, therefore, before ```_release_resources()``` task will run.

pulpcore/tasking/util.py
```diff
@@ -59,16 +63,14 @@ def cancel(task_id):
...
     with transaction.atomic():
-        task_status.state = TASK_STATES.CANCELED
         for report in task_status.progress_reports.all():
             if report.state not in TASK_FINAL_STATES:
                 report.state = TASK_STATES.CANCELED
                 report.save()
-        task_status.save()
+
         _delete_incomplete_resources(task_status)
         task_status.release_resources()
```
dralley, line 72: The fact that tasks start running prior to completion of ```cancel()``` means that this is probably a little bit racy.  That is true both before and after this change.  Basically incomplete resources could hang around while new tasks start running.

pulpcore/tasking/worker.py
```diff
@@ -101,10 +104,23 @@ def perform_job(self, job, queue):
...
-        def check_kill(conn, id, interval=1):
+        def check_kill(conn, job_id, interval=1):
             while True:
-                res = conn.srem(TASKING_CONSTANTS.KILL_KEY, id)
+                res = conn.srem(TASKING_CONSTANTS.KILL_KEY, job_id)
                 if res > 0:
+                    try:
+                        task = Task.objects.get(pk=job_id)
+                        task.state = TASK_STATES.CANCELED
+                        task.save()
```
bmbouter, line 114: Is this codepath here just for extra insurance in case the workhorse is killed before the `cancel()` does it?

dralley: Basically.

* cancel() sets the task to canceled, but it's in a transaction, so the change isn't committed until cancel() finishes doing its work including cleaning up incomplete resources and stuff
* The workhorse gets killed
* The next task (release_resources) immediately starts running
* If cancel(), which is running in parallel, still hasn't comitted its transaction, so release_resources() thinks that the task died out of the blue because the task state is still "running", and complains about it.

I'm a little surprised that all the RQ stuff starts happening faster than cancel() is able to complete, but that's what I'm seeing.

bmbouter: Yup yup. This is an ok thing to have, but I'm wondering if we can find a way to not have these state changes in so many places. 

I commented above on maybe moving the transaction above the issuing of the kill key. I *think* that would keep it in one place. wdyt?

pulpcore/tasking/util.py
```diff
@@ -60,15 +61,16 @@ def cancel(task_id):
...
     with transaction.atomic():
         task_status.state = TASK_STATES.CANCELED
+        task_status.save()
+
         for report in task_status.progress_reports.all():
             if report.state not in TASK_FINAL_STATES:
                 report.state = TASK_STATES.CANCELED
                 report.save()
-        task_status.save()
+
```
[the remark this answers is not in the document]
dralley: I didn't want to do that because there would be a higher likelihood that the progress reports would not reflect reality, which is somewhat unavoidable but it was marginally convincing.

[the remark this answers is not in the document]
bmbouter: That's true but I still think we should consider it. One of the concerns with two places setting it is:

1. It's a race with two writes going to the same object in the db
2. The workhorse sets the task state first
3. The cancel() transaction fails
4. The other parts of the transaction that fail are never recovered from.

I would rather have progress reports our of date for a moment than this problem and doing the same thing in multiple places.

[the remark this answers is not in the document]
bmbouter: From a high level, having two sections contain the same code that perform a write-write race on the same data is telling me something is not right with our design. I think removing the part from check_kill would resolve this.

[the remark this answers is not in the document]
dralley: @bmbouter How does this look

pulpcore/tasking/util.py
bmbouter: This is great!

bmbouter (approved):
This is looking exactly right (to me). Thanks @dralley !

daviddavis (approved):

PR #937, opened 2020-09-27
Handles duplicate content in the stages API.

dkliban:
If the very first Content in a batch has already been prefetched in a previous call to prefetch_related_objects,
Django does not set the attribute specified in 'to_attr' when prefetching the rest of the items in the batch. As
a result, the '_remote_artifact_saver_cas' attribute is not set on the Content instances and AttributeError is
raised when '_remote_artifact_saver_cas' is accessed by the RemoteArtifactSaverStage.

This patch removes any existing '_remote_artifact_saver_cas' attributes on Content before calling into Django's
prefetch_related_objects().

fixes: #7147
https://pulp.plan.io/issues/7147

### DIFF
--- a/CHANGES/7147.bugfix
+++ b/CHANGES/7147.bugfix
@@ -1,0 +1,1 @@
+Properly handle duplicate content during synchronization and migration from Pulp 2 to 3.
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -258,6 +258,11 @@ class RemoteArtifactSaver(Stage):
         """
         remotes_present = set()
         for d_content in batch:
+            # If the attribute is set in a previous batch on the very first item in this batch, the
+            # rest of the items in this batch will not get the attribute set during prefetch.
+            # https://code.djangoproject.com/ticket/32089
+            if hasattr(d_content.content, "_remote_artifact_saver_cas"):
+                delattr(d_content.content, "_remote_artifact_saver_cas")
             for d_artifact in d_content.d_artifacts:
                 if d_artifact.remote:
                     remotes_present.add(d_artifact.remote)


bmbouter:
Is there some sort of test that can test this? There may not be, but I was wondering.

dkliban:
@bmbouter I am unable to reproduce this bug outside of the migration plugin where 2 repositories with the same content are being migrated.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -278,7 +278,10 @@ def _needed_remote_artifacts(self, batch):
         )
         needed_ras = []
         for d_content in batch:
-            for content_artifact in d_content.content._remote_artifact_saver_cas:
+            content_artifacts = getattr(d_content.content, "_remote_artifact_saver_cas", None)
+            if not content_artifacts:
+                content_artifacts = d_content.content.contentartifact_set.all()
```
ipanova, line 283: :+1:

CHANGES/7147.bugfix
```diff
@@ -0,0 +1 @@
+Fixed a bug that prevented syncing and migrating of duplicate content in a single task.
```
ipanova, line 1: s/syncing/synchronization/
s/migrating/migration/

ipanova (approved):

CHANGES/7147.bugfix
dkliban: thanks! i'll update it

goosemania (dismissed):

pulpcore/plugin/stages/artifact_stages.py
bmbouter, line 283: Disclaimer: I could be way wrong about this. I'm very unsure.

Is it possible the right thing to do when `d_content.content._remote_artifact_saver_cas == None` is to not save any instead of saving all? I'm asking because the queryset restrictions on the prefetching I think are for more than performance, they are identifying which of the objects need to be saved at all. If the result is None, shouldn't we save None?

goosemania: @bmbouter , I think that what you are saying is that we need to differentiate the `_remote_artifact_saver_cas == None` and the absence of `_remote_artifact_saver_cas` and handle them differently, because currently they are handled in the same way. Do I understand you correctly?
I agree that for the case when `_remote_artifact_saver_cas` is explicitly `None`, we probably should not save anything. However when this attribute is absent it means that there was a duplicated content (but with different remote) and for this we likely need to create a remote artifact.  Likely - because we definitely need to do it on the first run but we don;t have a mechanism to identify if remote artifacts are needed for duplicated content on the subsequent reruns.

bmbouter: Yeah `None` is not the same as having the attribute not set at all, so thanks for separating those two things.

I think I'm still unsure of why the attribute is not being set. How do we know when this attribute is absent it means there was duplicated content (but with a different remote)?

A related question: are we sure that creating all of these object in the situation where the attribute is not set at all is more correct than not creating them? I keep wondering if the absense of the attribute means there is nothing for our code to do.

I'm still very lost on all this, so apologies for having to get more explanation on a PR with two acks already.

goosemania: @bmbouter, that's a good discussion, let's answer all questions. I removed my ack, so you feel better ;)

I base my understanding on @dkliban's commit message `The prefetch_related_objects method provided by Django fails to set the 'to_attr' for duplicate model instances that have already been prefetched as part of the call into prefetch_related_objects.` , so I'll let @dkliban answer your questions :)

dkliban:
prefetch_related_objects ends up deduplicating the list of instances when that list is converted to a queryset. Here is how this transformation occurs:

*  List of model instances passed in to `prefetch_related_objects()` is stored as `obj_list`[0]
*  The value of `obj_list` is replaced by the return value of `prefetch_one_level()`[1]
   * `prefetch_one_level()` passes the list of instances RelatedManager's `prefetcher.get_prefetch_queryset()`[2] 
   * `get_prefetch_queryset()` transforms the list of model instances into a queryset[3] and returns it as `rel_qs` inside `prefetch_one_level()`[1]
   * `rel_qs` queryset is then converted back to a list called `all_related_objects`[4]
   * `all_related_objects` is returned and it's value it replaces the value of `obj_list` inside `prefetch_related_objects()`[1]
* `obj_list` is then replaced by a new list that is built using cached models referenced by `obj_list`

[0] https://github.com/django/django/blob/stable/2.2.x/django/db/models/query.py#L1569
[1] https://github.com/django/django/blob/stable/2.2.x/django/db/models/query.py#L1625
[2] https://github.com/django/django/blob/stable/2.2.x/django/db/models/query.py#L1737-L1738
[3] https://github.com/django/django/blob/stable/2.2.x/django/db/models/fields/related_descriptors.py#L622-L623
[4] https://github.com/django/django/blob/stable/2.2.x/django/db/models/query.py#L1755
[5] https://github.com/django/django/blob/stable/2.2.x/django/db/models/query.py#L1641-L1645

dkliban:
I finally figured out what is actually happening in Django's `prefetch_related_objects()`. I've updated the PR description as well as the commit message to reflect these findings. Here are some more details.

When Django determines that the very first instance in the list `is_fetched`[0], it does not call into the the `prefetch_one_level()`[1]. The attributed specified in the `to_attr` parameter is only set in the `prefetch_one_level()` method[2].

This patch removes the `_remote_artifact_saver_cas` attribute from Content before calling into the `prefetch_related_objects()` method. As a result, Django treats every instance as if it has never been prefetched before. 

[0] https://github.com/django/django/blob/stable/2.2.x/django/db/models/query.py#L1605-L1609
[1] https://github.com/django/django/blob/stable/2.2.x/django/db/models/query.py#L1624
[2] https://github.com/django/django/blob/stable/2.2.x/django/db/models/query.py#L1799

dralley:
I applied this patch on a setup that was experiencing the error and can confirm that it does seem to fix it.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -258,6 +258,10 @@ def _needed_remote_artifacts(self, batch):
         """
         remotes_present = set()
         for d_content in batch:
+            # If the attribute is set in a previous batch on the very first item in this batch, the
+            # rest of the items in this batch will not get the attribute set during prefetch.
+            if hasattr(d_content.content, '_remote_artifact_saver_cas'):
+                delattr(d_content.content, '_remote_artifact_saver_cas')
```
goosemania, line 264: Black complains that you use single quotes ;)

goosemania:
Great job! Thanks for thorough investigation , @dkliban ! Thank for prompting it, @bmbouter :)

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -258,6 +258,10 @@ def _needed_remote_artifacts(self, batch):
         """
         remotes_present = set()
         for d_content in batch:
+            # If the attribute is set in a previous batch on the very first item in this batch, the
```
bmbouter, line 261: This is very cool. Since this is working around a bug in Django, for posterity, can the Django bug be opened and this comment also link to it? I'd lgtm that.

goosemania (approved):

pulpcore/plugin/stages/artifact_stages.py
dkliban: Yes, I plan to open a bug later today.

dkliban: https://code.djangoproject.com/ticket/32089

bmbouter (approved):
I think this is a great investigation and a good workaround. Thank you!

pulpcore/plugin/stages/artifact_stages.py
dralley: Is it a bug, or just a "quirk"?

PR #944, opened 2020-09-29
Expose pulpcore_version to plugin writers

dralley:
Give plugin writers a way to access the pulpcore version.
This is possiblyuseful in case they want to support more than one
version of the plugin API at the same time.

[noissue]


daviddavis:
Any particular reason not to add a plugin api changelog? I would expect any change to the plugin api to have one.

dralley:
No, I can add one

daviddavis (approved):
LGTM. Thanks!

dkliban (approved):

PR #948, opened 2020-10-01
Configurable URL for Django admin site

cutwater:
closes: #7637
https://pulp.plan.io/issues/7637


ipanova (approved):

bmbouter (changes requested):
Can you add a setting to the settings page. Each setting has an entry in the docs here https://github.com/pulp/pulpcore/blob/master/docs/settings.rst#pulp-settings

bmbouter (approved):
Thanks!

PR #949, opened 2020-10-02
Building changelog for 3.7.1

dkliban:
[noissue]

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,24 @@ Changelog
 
 .. towncrier release notes start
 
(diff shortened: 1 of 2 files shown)


goosemania:
@dkliban , is it a cherry-pick? If yes, I suggest to be explicit  (`git cherry-pick -x`) about it next time, so it's clear that it's a cherry-pick and from which commit.

goosemania (approved):

dkliban:
It is a cherry-pick. I'll use -x.

PR #952, opened 2020-10-06
Document that we do not support backporting migrations

daviddavis:
fixes #7657

### DIFF
--- a/docs/bugs-features.rst
+++ b/docs/bugs-features.rst
@@ -89,6 +89,9 @@ For **Backport Requests**
 Only bug fixes from newer versions can be requested to be backported. No features can be backported
 to older versions.
 
+Django migrations are order dependent so we cannot backport any change that has database migrations.
+Thus z-stream releases will not add new migrations.
+
 Please include:
 
 #. A redmine issue number or URL for the bug fix to backport. If possible, relate those two issues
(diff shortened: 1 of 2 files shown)


ipanova (approved):

docs/bugs-features.rst
```diff
@@ -89,6 +89,9 @@ For **Backport Requests**
 Only bug fixes from newer versions can be requested to be backported. No features can be backported
 to older versions.
 
+Django migrations are order dependent so we cannot backport any change that has database migrations.
+And thus we also do not ship new migrations in z-stream releases.
```
bmbouter, line 93: nitpick, wdyt?

```suggestion
Thus we do not ship migrations in z-stream releases.
```

daviddavis: Your suggestion looks good to me except for dropping the word 'new' here. We do ship migrations in z-stream releases but just not new ones. So for users, if they upgrade from say Pulp 3.9 to 3.10.2, they'll still get the migrations from 3.10.0.

daviddavis: Perhaps this should be, "Thus z-stream releases will not add new migrations."

bmbouter: +1 to your wording. Agreed.

bmbouter (approved):
Thank you!

PR #953, opened 2020-10-06
Added check for ALLOWED_CONTENT_CHECKSUMS that Artifacts are missing

daviddavis:
fixes #7487

### DIFF
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -15,6 +15,7 @@ from importlib import import_module
 from pkg_resources import iter_entry_points
 
 from django.core.exceptions import ImproperlyConfigured
+from django.db import connection
 
 from pulpcore import constants
 
@@ -283,3 +284,24 @@ if unknown_algs:
             "provided: {}".format(unknown_algs)
         )
     )
+try:
+    for checksum in ALLOWED_CONTENT_CHECKSUMS:
+        # can't import Artifact here so use a direct db connection
+        with connection.cursor() as cursor:
+            cursor.execute(f"SELECT count(pulp_id) FROM core_artifact WHERE {checksum} IS NULL")
+            row = cursor.fetchone()
+            if row[0] > 0:
+                raise ImproperlyConfigured(
+                    _(
+                        "ALLOWED_CONTENT_CHECKSUMS cannot contain checksum '{}' since checksums "
+                        "can only be removed and not added to ALLOWED_CONTENT_CHECKSUMS in "
+                        "existing deployments."
+                    ).format(checksum)
+                )
+except ImproperlyConfigured as e:
+    raise e
+except Exception:
+    # our check could fail if the table hasn't been created yet or we can't get a db connection
+    pass
+finally:
+    connection.close()
--- a/CHANGES/7487.feature
+++ b/CHANGES/7487.feature
@@ -1,0 +1,2 @@
+Added check to prevent users from adding checksums to ``ALLOWED_CONTENT_CHECKSUMS`` if there are
+Artifacts without those checksums.


pulpcore/app/settings.py
```diff
@@ -283,3 +284,24 @@
...
+                        "deployments."
+                    ).format(checksum)
+                )
+except ImproperlyConfigured as e:
+    raise e
+except Exception:
+    # our check could fail if the table hasn't been created yet or we can't get a db connection
+    pass
+finally:
+    connection.close()
```
ipanova, line 307: :zap:

bmbouter (approved):
Looks good, thank you!

ipanova:
the change looks good to me however i don't understand why travis fails.

ipanova (approved):

pulpcore/app/settings.py
mdellweg: I'd prefer if we could be more specific about the exception we want to `pass`. We would not need to reraise `ImproperlyConfigured` in that case.
Also is it possible, that `connection.close()` raises an error?

daviddavis: > I'd prefer if we could be more specific about the exception we want to pass. We would not need to reraise ImproperlyConfigured in that case.

I assume you're referring to line 303? There could be a wide variety of exceptions that get raised here depending on the state of the database. I don't think we want to raise ImproperlyConfigured if, for example, the database hasn't been initialized yet.

My thought about catching a generic `Exception` is that there's a wide variety of different exceptions that could be raised. In most cases, I think we just want to forgo checking ALLOWED_CONTENT_CHECKSUMS. 

> Also is it possible, that connection.close() raises an error?

This is possible. I will add a try/except.

bmbouter: Moving the connection.close() into the existing try/except I think would make sense.

daviddavis: I can do that. I didn't put the `connection.close()` in the try block because I assumed that it might not be reached if an exception gets thrown and thus the connection might not get closed properly. But I guess at that point, if an exception gets thrown, the process is mostly likely going to exit anyway?

mdellweg: I was concerned about the case when you throw a random exception inside the try block. Let's say it was an exception due to `connection.*` not working properly. Now you ignore that exception with `pass`, but `connection.close()` is called regardlessly (that's what `finally` is for). So my question is: How likely will `connection.close()` throw an exception if `connection.cursor()` threw one already?

daviddavis: It seems likely actually. That would leave me to believe that we should move `connection.close()` to the try block so we only execute it if `connection.cursor()` executed without throwing an exception. Is that your thinking as well?

mdellweg: i would still attempt to close it in finally to be sure and just ignore any exception there.

daviddavis: So this change?

```suggestion
    try:
        connection.close()
     except Exception:
         pass
```

mdellweg: Jeps!

bmbouter: So will this work out so well for us? We had to add this explicit close because not having it was secretly leaving a connection open. If we silence an error here and continue we're back in that situation. If there is a problem to the extent that the `close()` call fails, I think it would be better for users to know that and have that be a fatal exception for Pulp.

daviddavis: It depends. I think the problem with not catching `close()` exceptions is that we might be halting the code in cases where we actually want the code to proceed. I did some more testing though and it looks my previous statement about `close()` was incorrect. `close()` looks like it doesn't tend to raise exceptions in cases where `connection.cursor()` does (eg there's no db initialized, bad db password, etc). So I think not rescuing here is probably the correct choice.

bmbouter: I think the code as-is is the strongest, but if @mdellweg and @daviddavis, you'd like to do it differently feel free to adjust. I'm leaving the lgtm I put on it with support for either approach.

mdellweg: @daviddavis thanks that is a good analysis and probably the right answer to my question.
@bmbouter I'm retiring my concerns.

PR #957, opened 2020-10-09
Using check-manifest to update manifest

daviddavis:
refs #7656

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,3 +67,23 @@ exclude = '''
   | docs
 )/
 '''
+
+[tool.check-manifest]
+ignore = [
+    ".bumpversion.cfg",
+    "CHANGES/**",
+    "CONTRIBUTING.rst",
+    "HISTORY.rst",
+    "dev_requirements.txt",
+    "doc_requirements.txt",
+    "docs/**",
+    "flake8.cfg",
+    "template_config.yml",
+    ".travis/**",
+    ".travis.yml",
+    ".coveragerc",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -67,3 +67,23 @@ exclude = '''
...
+
+[tool.check-manifest]
+ignore = [
+    ".bumpversion.cfg",
+    "CHANGES/**",
+    "CONTRIBUTING.rst",
+    "HISTORY.rst",
+    "dev_requirements.txt",
+    "doc_requirements.txt",
+    "docs/**",
```
bmbouter, line 79: Does this mean we're not including the docs in our shipped python wheels? Or are they still included in the sdists? Do we publish sdists or wheels only?

daviddavis: I'm hoping someone with more knowledge of our packaging weighs in here but I believe we aren't shipping docs with our wheels. Not sure about your other questions.

goosemania (approved):

pyproject.toml
goosemania: FWIW, [MANIFEST.in affects sdists only](https://packaging.python.org/guides/distributing-packages-using-setuptools/#manifest-in), not bdists like wheels. So I believe docs are not included in the sdists.
Not sure what is included in to bdist by default.

bmbouter: Great, thank you.

bmbouter (approved):
Thanks!

PR #968, opened 2020-10-16
Pass an adjusted request object to the serializer

lubosmj:
closes #7718

### DIFF
--- a/pulpcore/app/serializers/task.py
+++ b/pulpcore/app/serializers/task.py
@@ -25,12 +25,23 @@ class CreatedResourceSerializer(RelatedField):
                 return None
         except AttributeError:
             pass
-        viewset = get_viewset_for_model(data.content_object)
 
-        # serializer contains all serialized fields because we are passing
-        # 'None' to the request's context
-        serializer = viewset.serializer_class(data.content_object, context={"request": None})
+        # query parameters can be ignored because we are looking just for 'pulp_href'; still,
+        # we need to use the request object due to contextual references required by some
+        # serializers
+        request = self._get_request_without_query_params()
+
+        viewset = get_viewset_for_model(data.content_object)
+        serializer = viewset.serializer_class(data.content_object, context={"request": request})
         return serializer.data.get("pulp_href")
+
+    def _get_request_without_query_params(self):
+        """Remove all query parameters from the request object."""
+        request = self.context["request"]
+        request.query_params._mutable = True
+        request.query_params.clear()
+        request.query_params._mutable = False
+        return request
 
     class Meta:
         model = models.CreatedResource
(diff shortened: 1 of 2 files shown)


lubosmj:
This pull request is required by https://github.com/pulp/pulp_container/pull/169.

pulpcore/app/serializers/task.py
```diff
@@ -25,13 +25,24 @@ def to_representation(self, data):
...
             pass
-        viewset = get_viewset_for_model(data.content_object)
 
-        # serializer contains all serialized fields because we are passing
-        # 'None' to the request's context
-        serializer = viewset.serializer_class(data.content_object, context={"request": None})
+        # query parameters can be ignored because we are looking just for 'pulp_href'; still,
+        # we need to use the request object due to contextual references required by some
+        # serializers
+        request = self.get_request_without_query_params()
```
mdellweg, line 32: Apart from the minimalistic approach, what is the reason to remove the query parameters?

ipanova: i believe because of this issue https://github.com/pulp/pulpcore/commit/ac75012ba21e593bd3f017e81f3f56dc57a0de83

ipanova (approved):

pulpcore/app/serializers/task.py
```diff
@@ -25,13 +25,24 @@ def to_representation(self, data):
...
+
+        viewset = get_viewset_for_model(data.content_object)
+        serializer = viewset.serializer_class(data.content_object, context={"request": request})
         return serializer.data.get("pulp_href")
 
+    def get_request_without_query_params(self):
```
bmbouter, line 38: Maybe make this an internal method? `_get_request_without_query_params`.

pulpcore/app/serializers/task.py
lubosmj: @mdellweg, this change was made due to this: https://github.com/wimglenn/djangorestframework-queryfields/blob/b0aed3eb51625b20d187aa2ca1ec5a2982b979e2/drf_queryfields/mixins.py#L42. If we do not add `pulp_href` to request's query params, it is simply excluded and the serializer has got no access to `pulp_href` in our case.

The issue @ipanova is referring to, had allowed us to return from QueryFieldsMixin before excluding fields: https://github.com/wimglenn/djangorestframework-queryfields/blob/b0aed3eb51625b20d187aa2ca1ec5a2982b979e2/drf_queryfields/mixins.py#L19. Thus, `pulp_href` was available to the serializer all the time.

In this PR, I discarded query parameters for the sake of the `pulp_href` presence.

dkliban (approved):

PR #970, opened 2020-10-18
Always set a default for DJANGO_SETTINGS_MODULE

ekohl:
6a2e7f304c31adcbf12d964753f639c5e3ea4cce already set this in one place for the handler, but this does it everywhere before calling django.setup. This means the services files don't need to.

### DIFF
--- a/CHANGES/7720.feature
+++ b/CHANGES/7720.feature
@@ -1,0 +1,1 @@
+Always set a default for DJANGO_SETTINGS_MODULE. This means the services files don't need to.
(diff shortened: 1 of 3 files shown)


mdellweg (approved):
:+1: for consistency!

ekohl:
Btw, I did think about some small wrapper file that does `django.setup()` after having the default `DJANGO_SETTINGS_MODULE` but not sure where you would place that.

dkliban (approved):
I am approving this change. @bmbouter, what do you think about this change?

bmbouter (approved):
This is great.

PR #980, opened 2020-10-22
Fix file descriptor leak in repo version repair

dralley:
closes: #7735
https://pulp.plan.io/issues/7735

### DIFF
--- a/CHANGES/7735.bugfix
+++ b/CHANGES/7735.bugfix
@@ -1,0 +1,1 @@
+Fixed a file descriptor leak during repository version repair operations.
--- a/pulpcore/app/tasks/repository.py
+++ b/pulpcore/app/tasks/repository.py
@@ -58,12 +58,16 @@ async def _repair_ca(content_artifact, repaired=None):
     return False
 
 
+CHUNK_SIZE = 1024 * 1024  # 1 Mb
+
+
 def _verify_ca(content_artifact):
     try:
         # verify files digest
         hasher = hashlib.sha256()
-        for chunk in iter(lambda: content_artifact.artifact.file.read(1024 * 1024), b""):
+        for chunk in content_artifact.artifact.file.chunks(CHUNK_SIZE):
             hasher.update(chunk)
+        content_artifact.artifact.file.close()
         return hasher.hexdigest() == content_artifact.artifact.sha256
     except FileNotFoundError:
         return False
--- a/pulpcore/app/tasks/repository.py
+++ b/pulpcore/app/tasks/repository.py
@@ -35,8 +35,9 @@ def delete_version(pk):
             return
 
         log.info(
-            _("Deleting and squashing version %(v)d of repository %(r)s"),
-            {"v": version.number, "r": version.repository.name},
+            _("Deleting and squashing version {num} of repository '{repo}'").format(
+                num=version.number, repo=version.repository.name
+            )
         )
 
         version.delete()
@@ -84,8 +85,8 @@ async def _repair_repository_version(version):
         ) as repaired:
             query_set = models.ContentArtifact.objects.filter(
                 content__in=version.content
-            ).prefetch_related("artifact")
-            for content_artifact in query_set:
+            ).select_related("artifact")
+            for content_artifact in query_set.iterator():
                 if not content_artifact.artifact:
                     continue
                 if not await loop.run_in_executor(None, partial(_verify_ca, content_artifact)):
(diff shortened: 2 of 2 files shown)


CHANGES/7735.bugfix
```diff
@@ -0,0 +1 @@
+Fix a file descriptor leak during repository version repair operations.
```
ipanova, line 1: s/Fix/Fixed

pulpcore/app/tasks/repository.py
```diff
@@ -84,8 +84,10 @@ async def _repair_repository_version(version):
         ) as repaired:
             query_set = models.ContentArtifact.objects.filter(
                 content__in=version.content
-            ).prefetch_related("artifact")
-            for content_artifact in query_set:
+            ).select_related(
+                "artifact"
+            )
+            for content_artifact in query_set.iterator():
```
ipanova, line 90: :+1

pulpcore/app/tasks/repository.py
```diff
@@ -58,12 +58,16 @@ async def _repair_ca(content_artifact, repaired=None):
...
+
+
 def _verify_ca(content_artifact):
     try:
         # verify files digest
         hasher = hashlib.sha256()
-        for chunk in iter(lambda: content_artifact.artifact.file.read(1024 * 1024), b""):
+        for chunk in content_artifact.artifact.file.chunks(CHUNK_SIZE):
             hasher.update(chunk)
+        content_artifact.artifact.file.close()
```
dralley, line 70: The "actual fix"

ipanova (approved):

dkliban (approved):

pulpcore/app/tasks/repository.py
```diff
@@ -80,8 +85,8 @@ async def _repair_repository_version(version):
         ) as repaired:
             query_set = models.ContentArtifact.objects.filter(
                 content__in=version.content
-            ).prefetch_related("artifact")
-            for content_artifact in query_set:
+            ).select_related("artifact")
+            for content_artifact in query_set.iterator():
```
goosemania, line 89: This should improve memory usage, right?
I suggest to make such improvements visible through the changelog.
Not insisting to do it for this one. Just keep in mind that calling out any perf/memory improvements in our blog or release notes is a good advertisement of the fact that we care about the scalability and quality of the project.

dralley: Probably, but not by a huge amount.  ContentArtifacts are pretty small and we're dealing with, at most, however many of them are present in the RepositoryVersion.

My general guideline is that I don't want to advertise "probablies" that weren't measured, and I don't really think it's worth measuring, and the quality will speak for itself long-term.

pulpcore/app/tasks/repository.py
```diff
@@ -58,12 +59,16 @@ async def _repair_ca(content_artifact, repaired=None):
...
 
+CHUNK_SIZE = 1024 * 1024  # 1 Mb
+
+
 def _verify_ca(content_artifact):
     try:
         # verify files digest
         hasher = hashlib.sha256()
-        for chunk in iter(lambda: content_artifact.artifact.file.read(1024 * 1024), b""):
+        for chunk in content_artifact.artifact.file.chunks(CHUNK_SIZE):
```
mdellweg, line 69: :1st_place_medal:

pulpcore/app/tasks/repository.py
goosemania: I agree with the time-will-tell approach.
I just think that these are different target audiences we are talking about :) , people who are users already and will be happy with time that pulp is getting better and notice performance improvements, and people who look at projects and evaluate which one to choose. The latter ones potentially skim through the changelog or blog posts.

ekohl:
I just ran into this, but isn't there a `with` block equivalent? if there's an exception the file still isn't closed. While a bit of an edge case, it's still good to cover IMHO.

dralley:
@ekohl Well, I'm not sure if it could possibly throw an exception other than `FileNotFound`, but at least in that case there would be no open file.

ekohl:
Never assume it won't fail? I think this actually works and is safer:
```python
with content_artifact.artifact.file as fp:
    for chunk in fp.chunks(CHUNK_SIZE):
        hasher.update(chunk)
```

dralley:
I don't believe these objects work in that manner

ekohl:
`django.core.files.File` does implement the protocol:
https://github.com/django/django/blob/225d96533a8e05debd402a2bfe566487cc27d95f/django/core/files/base.py#L102-L106
(I linked latest but in 2.2 it's the same)

PR #981, opened 2020-10-23
Upload chunks as separate files

lubosmj:
closes #4498


lubosmj:
This pull request is the same as https://github.com/pulp/pulpcore/pull/914. The only difference here is that the settings file is not changed and therefore no compatibility issues should persist (https://github.com/pulp/pulpcore/pull/973).

daviddavis:
So this would store chunks in /var/lib/pulp/upload/ by default?

lubosmj:
@daviddavis, yes, to the place where uploads were stored before. Do you suggest me to create a new setting called CHUNKED_UPLOAD_DIR_NEW which will point to /var/lib/pulp/chunks/?

daviddavis:
No I think /var/lib/pulp/upload is good.

daviddavis (approved):

CHANGES/4498.feature
```diff
@@ -0,0 +1 @@
+Made uploaded chunks to be stored as separated files in the default storage.
```
mdellweg, line 1: @daviddavis  Should we mention, that this finally lifts the requirement that the api-nodes have a shared storage?

CHANGES/4498.feature
```diff
@@ -0,0 +1,3 @@
+Made uploaded chunks to be stored as separate files in the default storage. This feature allows
+Pulp nodes to have a shared storage when dealing with clustered deployment. Each node can now
+access all uploaded chunks as they were uploaded directly to that node.
```
mdellweg, line 3: I think it is: This feature removes the need for a share storage of pulp api nodes, as the chunks are now stored individually in the shared storage and are therefore accessible by all nodes.

lubosmj: Ah, okay, I will reword it. Thank you!

CHANGES/4498.feature
```diff
@@ -0,0 +1,3 @@
+Made uploaded chunks to be stored as separate files in the default storage. This feature removes
+the need for a share storage of pulp api nodes, as the chunks are now stored individually in the
+shared storage and are therefore accessible by all nodes.
```
mdellweg, line 3: :+1:

mdellweg (approved):

pulpcore/app/tasks/upload.py
```diff
@@ -22,7 +25,14 @@ def commit(upload_id, sha256):
         log.info(_("The upload was not found. Nothing to do."))
         return
 
-    file = files.PulpTemporaryUploadedFile.from_file(upload.file.file)
+    chunks = models.UploadChunk.objects.filter(upload=upload).order_by("offset")
```
mdellweg, line 28: Probably not in this pr, but can we add a check, that the chunks uploaded neither overlap nor fail to cover the whole file?
The overlap check can be added to where we save the upload chunk, and then a "sum of all chunk sizes == expected file size" should be sufficient.

pulpcore/app/models/upload.py
```diff
@@ -32,43 +27,49 @@ def append(self, chunk, offset, sha256=None):
...
+        return storage.get_upload_chunk_file_path(self.pulp_id)
+
+    file = fields.FileField(null=False, upload_to=storage_path, max_length=255)
     upload = models.ForeignKey(Upload, on_delete=models.CASCADE, related_name="chunks")
     offset = models.BigIntegerField()
     size = models.BigIntegerField()
+
+    def delete(self, *args, **kwargs):
+        """
+        Delete UploadChunk model and the file associated with the model
```
daviddavis, line 68: :+1:

PR #983, opened 2020-10-23
Raise an error when a user tries to update a signing service

lubosmj:
closes #7701

### DIFF
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -18,6 +18,7 @@ from django.core import validators
 from django.core.files.storage import default_storage
 from django.db import IntegrityError, models, transaction
 from django.forms.models import model_to_dict
+from django_lifecycle import BEFORE_UPDATE, hook
 
 from pulpcore.constants import ALL_KNOWN_CONTENT_CHECKSUMS
 from pulpcore.app.models import MasterModel, BaseModel, fields, storage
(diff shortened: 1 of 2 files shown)


lubosmj:
When a user performs a single update, the error is correctly raised. However, this approach has a limitation when executing bulk operations (such as SigningService.objects.filter(name="s").update(name="b")):

> Be aware that the update() method is converted directly to an SQL statement. It is a bulk operation for direct updates. It doesn’t run any save() methods on your models, or emit the pre_save or post_save signals (which are a consequence of calling save()), or honor the auto_now field option. If you want to save every item in a QuerySet and make sure that the save() method is called on each instance [https://docs.djangoproject.com/en/2.2/topics/db/queries/#updating-multiple-objects-at-once].

pulpcore/app/models/content.py
```diff
@@ -609,6 +610,13 @@ def save(self, *args, **kwargs):
         self.validate()
         super().save(*args, **kwargs)
 
+    @hook(BEFORE_UPDATE)
+    def on_update(self):
+        raise RuntimeError(
+            "The signing service is immutable. It is advised to create a new signing service "
```
bmbouter, line 616: Can this be i8ln with gettext please?

bmbouter (changes requested):
This looks good once the i8ln is fixed. Thank you!

mdellweg (approved):
The requested change has been made.

bmbouter (approved):
Thank you!

PR #987, opened 2020-10-27
Adding missing pytest dependency

daviddavis:
fixes #7753

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ cover
 *,cover
 tags
 .pytest_cache
+.venv
 
 # KDevelop
 .kdev4/
--- a/CHANGES/7753.misc
+++ b/CHANGES/7753.misc
@@ -1,0 +1,1 @@
+Added missing pytest dependency.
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,40 @@
...
+	# TODO: pbindings won't work here. need to recreate this functionality
+	pbindings pulpcore python
+
+test-unit: setup-virtualenv
+	$(VENV)/bin/django-admin test --noinput
+
+test-functional: setup-virtualenv bindings
+	$(VENV)/bin/pytest -v -r sx --color=yes --pyargs pulpcore.tests.functional
+
+test: test-unit test-functional
```
mdellweg, line 40: This is not, what the commit message promises.

mdellweg (commented):
I think the .venv parts are unrelated, too.

daviddavis:
> I think the .venv parts are unrelated, too.

How about a separate commit then?

mdellweg:
> > I think the .venv parts are unrelated, too.
> 
> How about a separate commit then?

I thought the .venv things were related to that makefile you were drafting.

daviddavis:
No, not necessarily. I am using .venv to store my dependencies when I work on pulpcore outside of our vagrant environment. It's a common convention to ignore `.venv`. See [black's config](https://github.com/psf/black/blob/989ea69bd1d0a02aa6c5794d941c122f7b4958b4/pyproject.toml#L19) as an example.

daviddavis:
Here's another example: https://github.com/github/gitignore/blob/master/Python.gitignore#L109

mdellweg (approved):

goosemania (approved):

PR #991, opened 2020-10-28
Add a univeral Pulp repair endpoint

dralley:
Create a new endpoint /pulp/api/v3/repair/ which can attempt to fix corruption
across all of Pulp.

closes: #7755
https://pulp.plan.io/issues/7755


dralley:
TODO:

* Smash test
* Expose the ```check_hashes``` option to the user?
    * Allowing the user to disable checksum verification and only check for *missing* files would be much faster (orders of magnitude), and if that user already has a fancy corruption-detecting filesystem such as ZFS or Btrfs, might be a better value proposition the majority of the time.

dralley:
@dkliban @mdellweg Before I start on the smash tests, could you please sanity check the code and give feedback on whether we should expose ```check_hashes```?

pulpcore/app/tasks/repository.py
```diff
@@ -59,39 +63,44 @@ async def _repair_ca(content_artifact, repaired=None):
...
+            query_set = models.ContentArtifact.objects.exclude(artifact__isnull=True)
+
+            if subset:
+                query_set = query_set.filter(content__in=subset)
+
+            for content_artifact in query_set.select_related("artifact").iterator():
+                artifact = content_artifact.artifact
+
+                if check_hashes:
+                    valid = _verify_artifact(artifact)
```
mdellweg, line 96: This command is calculating a checksum from a file on disk, which takes some time. At the same time the event loop might be downloading content (async code) that is blocked by this. To avoid this, `_verify_artifact` should be offloaded to another thread by `loop.run_in_executor` (until the django storage framework allows `chunks` to be used as an async iterator).

pulpcore/app/views/repair.py
```diff
@@ -0,0 +1,44 @@
...
+
+        async_result = enqueue_with_reservation(repair_all_artifacts, [], args=[verify_checksums])
+
+        return OperationPostponedResponse(async_result, request)
+
+
+class RepairSerializer(serializers.Serializer):
+    verify_checksums = fields.BooleanField(
+        required=False,
+        default=True,
```
dralley, line 39: I think there's actually a solid argument to make this default to ```False```, since the gulf in performance is so large, and the use case for detecting bitrot more limited.

pulpcore/app/views/repair.py
```diff
@@ -0,0 +1,44 @@
...
+
+        async_result = enqueue_with_reservation(repair_all_artifacts, [], args=[verify_checksums])
+
+        return OperationPostponedResponse(async_result, request)
+
+
+class RepairSerializer(serializers.Serializer):
+    verify_checksums = fields.BooleanField(
+        required=False,
+        default=False,
```
dralley, line 39: I think ```False``` is a better default than ```True``` for a couple of reasons.

* Reading every byte of every artifact out from disk and then computing the hashes will likely degrade the performance of the entire system since we're talking about potentially hundreds of gigabytes (or more) of data.
* Depending on the underlying storage technology (RAID, filesystem checksumming), error detection and correction could possibly be handled without our intervention.
* From recent experience, it seems like "whoops, files are gone" incidents could be more likely than random corruptions.

pulpcore/app/tasks/repository.py
```diff
@@ -44,7 +50,15 @@ def delete_version(pk):
 
 
 async def _repair_ca(content_artifact, repaired=None):
-    for remote_artifact in content_artifact.remoteartifact_set.all():
+    remotes = content_artifact.remoteartifact_set.all()
+
+    if not remotes:
+        log.warn(
+            _("Artifact {} is unrepairable - no remote source".format(content_artifact.artifact))
```
dralley, line 57: i.e. ```PublishedMetadata```, which we store via the ```Artifact``` machinery as well [0].  Should we print a more meaningful message here or is this ok?

[0] https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/publication.py#L185

pulpcore/app/tasks/repository.py
```diff
@@ -56,49 +70,67 @@ async def _repair_ca(content_artifact, repaired=None):
...
-                if not await loop.run_in_executor(None, partial(_verify_ca, content_artifact)):
-                    corrupted.increment()
-                    log.warn(_("Digest mismatch for {}").format(content_artifact))
+    ) as corrupted, ProgressReport(
+        message="Repair corrupted units", code="repair.repaired"
+    ) as repaired:
+
+        with ThreadPoolExecutor(max_workers=2) as checksum_executor:
+            for content_artifact in query_set.select_related("artifact").iterator():
+                artifact = content_artifact.artifact
```
dralley, line 108: In the case where ```verify_checksums=True```, we're now checking for the existence of each file first, separately.  This is a little slower, but probably insignificant, and it gives us better reporting.  This way missing files aren't reported as digest mismatches.

pulpcore/app/tasks/repository.py
```diff
@@ -44,7 +50,15 @@ def delete_version(pk):
 
 
 async def _repair_ca(content_artifact, repaired=None):
-    for remote_artifact in content_artifact.remoteartifact_set.all():
+    remotes = content_artifact.remoteartifact_set.all()
```
mdellweg, line 53: sounds like this variable should be named `remote_artifacts`.
Especially when i read it i thought "Remotes".

pulpcore/app/tasks/repository.py
mdellweg: I guess it's ok.
We can easily grep for `unrepairable` if a user ever hit's this.

pulpcore/app/tasks/repository.py
```diff
@@ -56,49 +70,67 @@ async def _repair_ca(content_artifact, repaired=None):
...
         log.warn(_("Redownload failed from {}.").format(remote_artifact.url))
-    return False
 
-
-CHUNK_SIZE = 1024 * 1024  # 1 Mb
+    return False
 
 
-def _verify_ca(content_artifact):
+def _verify_artifact(artifact):
```
mdellweg, line 77: Maybe add a comment why this should run in the executor.
(I know, i should have done that in the first place.)

pulpcore/app/tasks/repository.py
```diff
@@ -56,49 +70,67 @@ async def _repair_ca(content_artifact, repaired=None):
...
-            ).select_related("artifact")
-            for content_artifact in query_set.iterator():
-                if not content_artifact.artifact:
-                    continue
-                if not await loop.run_in_executor(None, partial(_verify_ca, content_artifact)):
-                    corrupted.increment()
-                    log.warn(_("Digest mismatch for {}").format(content_artifact))
+    ) as corrupted, ProgressReport(
+        message="Repair corrupted units", code="repair.repaired"
+    ) as repaired:
```
mdellweg, line 104: Yay for less indentation!

pulpcore/app/tasks/repository.py
```diff
@@ -121,7 +153,20 @@ def repair_version(repository_version_pk):
     )
 
     loop = asyncio.get_event_loop()
-    loop.run_until_complete(_repair_repository_version(version))
+    loop.run_until_complete(_repair_artifacts_for_content(subset=version.content))
```
mdellweg, line 156: shouldn't this allow the `verify_checksums` too?
for compatibility reasons probably with True as default.

pulpcore/app/views/repair.py
```diff
@@ -0,0 +1,44 @@
...
+    def post(self, request):
+        """
+        Repair artifacts.
+        """
+        serializer = RepairSerializer(data=request.data)
+        serializer.is_valid()
+
+        verify_checksums = serializer.validated_data["verify_checksums"]
+
+        async_result = enqueue_with_reservation(repair_all_artifacts, [], args=[verify_checksums])
```
mdellweg, line 31: Have you thought about locking?
I think, the only dangerous operation is the redownloading of the artifact. Does that interfer with any other workflows?

Should we limit the number of concurrent repair actions to 1?

I'm fine, if "No action required" is the answer.

pulpcore/tests/functional/api/using_plugin/test_repair.py
```diff
@@ -25,17 +26,21 @@
...
+    "pulpcore.app.models.storage.FileSystem",
+]
+
 
-    SUPPORTED_STORAGE_FRAMEWORKS = [
-        "django.core.files.storage.FileSystemStorage",
-        "pulpcore.app.models.storage.FileSystem",
-    ]
+@unittest.skipIf(
+    settings.DEFAULT_FILE_STORAGE not in SUPPORTED_STORAGE_FRAMEWORKS,
```
mdellweg, line 39: @dkliban didn't you change the way how settings can be accessed recently?

pulpcore/app/views/repair.py
dralley: I don't think we can realistically lock on every single artifact, especially since we won't know which ones it's operating on without duplicating some of the logic outside of the function.  I'm not sure how dangerous it really is considering that any affected artifacts are already broken.

I don't *think* it should interfere with anything else.

mdellweg: I agree. Thanks for sharing a thought about it. I had the same feeling and now it's much more comfortable.

pulpcore/app/tasks/repository.py
dralley: Can I add it to the comment in the location where it's called?

pulpcore/app/tasks/repository.py
dralley: Added

pulpcore/tests/functional/api/using_plugin/test_repair.py
dralley: We must have missed a spot, but I updated it to use "the new way"

pulpcore/app/tasks/repository.py
mdellweg: Sure

mdellweg (approved):

pulpcore/app/serializers/repository.py
```diff
@@ -264,6 +264,17 @@ class Meta:
         )
 
 
+class RepositoryVersionRepairSerializer(serializers.Serializer):
```
ipanova, line 267: can we use ValidateFieldsMixin so we check on unknown fields? same for the RepairViewset.

pulpcore/app/serializers/repository.py
```diff
@@ -264,6 +264,17 @@ class Meta:
...
+    verify_checksums = fields.BooleanField(
+        required=False,
+        default=True,
+        help_text=_(
+            "If enabled, will verify that the checksum of all stored files matches what saved in "
```
ipanova, line 272: the help_text is not accurate. the checksum verification is enabled by default.

pulpcore/app/tasks/repository.py
```diff
@@ -56,52 +70,72 @@ async def _repair_ca(content_artifact, repaired=None):
...
-        content_artifact.artifact.file.close()
-        return hasher.hexdigest() == content_artifact.artifact.sha256
+        artifact.file.close()
+        return hasher.hexdigest() == artifact.sha256
     except FileNotFoundError:
         return False
 
 
-async def _repair_repository_version(version):
+async def _repair_artifacts_for_content(subset=None, verify_checksums=True):
```
ipanova, line 89: i believe verify_checksums does not need to be set to True by default, it is always set through the serializer

pulpcore/app/views/repair.py
```diff
@@ -0,0 +1,44 @@
...
+
+        return OperationPostponedResponse(async_result, request)
+
+
+class RepairSerializer(serializers.Serializer):
```
ipanova, line 36: can we move this to /serializers/repair.py

CHANGES/7755.feature
```diff
@@ -0,0 +1 @@
+Added support for repairing Pulp by detecting and redownloading missing or corrupted artifact files. Sending a POST request to ``/pulp/api/v3/repair/`` will trigger a task that scans all artifacts for missing files in Pulp storage, and will attempt to redownload them from the original remote. Specifying ``verify_checksums=True`` when POSTing to the same endpoint will additionally verify that the checksums of each file in Pulp storage match the ones saved in the database, and will attempt to redownload corrupted files.
```
ipanova, line 1: @dralley I suggest adding docs about this feature

pulpcore/app/tasks/repository.py
dralley: That's true but I wanted it to be a keyword argument, and keyword arguments need a default.  It doesn't actually hurt anything.

ipanova (approved):

ipanova:
@dralley maybe also fix in the commit the typo s/univeral/universal/ :laughing:

pulpcore/app/views/repair.py
```diff
@@ -0,0 +1,44 @@
...
+    verify_checksums = fields.BooleanField(
+        required=False,
```
ipanova, line 38: @dralley let's make this consistent repo version repair endpoint and have this enabled by default

dralley:
@ipanova See documentation and feel free to merge if it looks good since I'll be out tomorrow.

After some discussion with Brian we went the "other way" and decided to default to including checksum verification.

docs/workflows/repairing-pulp.rst
```diff
@@ -0,0 +1,117 @@
...
+                "message": "Identify corrupted units",
+                "state": "completed",
+                "suffix": null,
+                "total": null
+            }
+        ],
+        "pulp_created": "2020-04-07T08:36:52.274985Z",
+        "pulp_href": "/pulp/api/v3/tasks/530302b4-8674-4db3-8a13-99febef80830/",
+        "reserved_resources_record": [
+            "/pulp/api/v3/repositories/file/file/47a3f651-aaa6-4026-b649-130c45ab38ea/"
```
ipanova, line 59: i assume it was copy/paste, because we are not locking on the repos when repairing all artifacts.

pulpcore/app/serializers/repair.py
```diff
@@ -0,0 +1,16 @@
...
+
+from rest_framework import fields, serializers
+
+from pulpcore.app.serializers import ValidateFieldsMixin  # noqa
+
+
+class RepairSerializer(serializers.Serializer, ValidateFieldsMixin):
+    verify_checksums = fields.BooleanField(
+        required=False,
+        default=False,
```
ipanova, line 11: default=True

pulpcore/app/serializers/repair.py
```diff
@@ -0,0 +1,16 @@
...
+        help_text=_(
+            "If enabled, will verify that the checksum of all stored files matches what saved in "
```
ipanova, line 13: "Will verify that the checksum of all stored files matches what saved in the database. Otherwise only the existence of the files will be checked. Enabled by default"

pulpcore/app/viewsets/repository.py
```diff
@@ -197,20 +198,24 @@ def destroy(self, request, repository_pk, number):
...
         responses={202: AsyncOperationResponseSerializer},
     )
     @action(detail=True, methods=["post"])
     def repair(self, request, repository_pk, number):
         """
-        Queues a task to repair currupted artifacts corresponding to a RepositoryVersion
+        Queues a task to repair corrupted artifacts corresponding to a RepositoryVersion
         """
         version = self.get_object()
+        serializer = RepairSerializer(data=request.data)
```
ipanova, line 210: :+1:

docs/workflows/repairing-pulp.rst
```diff
@@ -0,0 +1,117 @@
...
+    ways in future releases.
+
+
+Pulp provides some features for self-repair in cases where artifacts in the storage went missing or
+got corrupted in some way (i.e. bit rot).
+
+.. warning::
+
+    This repair feature looks for missing or corrupted files that are supposed to be in the
+    internal storage. It attempts a redownload of these files from known upstream urls.
```
ipanova, line 18: s/the internal storage/ the storage/

docs/workflows/repairing-pulp.rst
```diff
@@ -0,0 +1,117 @@
...
+    In case upstream stopped serving the corresponding files, or these files were uploaded
+    directly into Pulp or were created by Pulp itself (i.e. generated metadata), Pulp is
```
ipanova, line 20: s/Pulp/ Pulp repair feature/

ipanova (changes requested):
small things but needs to be fixed.

dralley:
Updated

ipanova (approved):
:100:

PR #997, opened 2020-10-28
Remove SELinux from installer note

melcorr:
[noissue]

Discussed this with @mikedep333 last week. 

Let me know if any changes are needed.

### DIFF
--- a/CHANGES/7793.doc
+++ b/CHANGES/7793.doc
@@ -1,0 +1,1 @@
+Removed outdated reference stating Pulp did not have an SELinux policy.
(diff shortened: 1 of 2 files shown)


mikedep333 (approved):

bmbouter (changes requested):
@melcorr I'm thinking this is significantly enough for it to show up in the changelog. If you agree, add a file to the CHANGES/xxxx.doc entry with a changelog. The XXXX would be a yet-to-be-filed issue in Redmine , which also should have the 'documentation' tag. The Redmine issue can be sparse, with maybe just a title that says "Remove references claiming Pulp does not have an SELinux policy."

Would you be willing to jump through these extra hoops for the changelog entry?

bmbouter (approved):
Thanks!

daviddavis (approved):

PR #1008, opened 2020-11-05
As a user, I have artifacts compliant with the allowed checksums.

ipanova:
Added command to populate missing checksums.

closes #7561
https://pulp.plan.io/issues/7561

### DIFF
--- a/CHANGES/7561.feature
+++ b/CHANGES/7561.feature
@@ -1,0 +1,2 @@
+Enabled users to add checksums to ALLOWED_CONTENT_CHECKSUMS by allowing them to populate checksums
+with handle-artifact-checksums command.
--- a/CHANGES/7914.feature
+++ b/CHANGES/7914.feature
@@ -1,0 +1,1 @@
+Added check to prevent Pulp to start if there are Artifacts with forbidden checksums.
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -262,6 +262,11 @@ ALLOWED_CONTENT_CHECKSUMS
 
       ``{"md5", "sha1", "sha224", "sha256", "sha384", "sha512"}``
 
+    .. warning::
+      If Pulp fails to start because forbidden checkums have been identified or required ones are
+      missing, run ``pulpcore-manager handle-artifact-checksums`` command. This command is in
+      tech-preview and may change in backwards incompatible ways in future releases.
+
 ADMIN_SITE_URL
 ^^^^^^^^^^^^^^
 
--- a/pulpcore/app/management/commands/handle-artifact-checksums.py
+++ b/pulpcore/app/management/commands/handle-artifact-checksums.py
@@ -1,0 +1,88 @@
+import hashlib
+import os
+
+from gettext import gettext as _
+
+from django.conf import settings
+from django.core.management import BaseCommand, CommandError
+from pulpcore import constants
+from pulpcore.plugin.models import Artifact
+
+import logging
+
+log = logging.getLogger("")
+
+CHUNK_SIZE = 1024 * 1024  # 1 Mb
+
+
+class Command(BaseCommand):
+    """
+    Django management command for populating or removing checksums on artifacts.
+    """
+
+    help = _("Handle missing and forbidden checksums on the artifacts")
+
+    def _download_artifact(self, artifact, checksum, file_path):
+        restored = False
+        for ca in artifact.content_memberships.all():
+            if not restored:
+                for ra in ca.remoteartifact_set.all():
+                    remote = ra.remote.cast()
+                    if remote.policy == "immediate":
+                        self.stdout.write(_("Restoring missing file {}").format(file_path))
+                        downloader = remote.get_downloader(ra)
+                        dl_result = downloader.fetch()
+                        # FIXME in case url is not available anymore this will break
+                        if dl_result.artifact_attributes["sha256"] == artifact.sha256:
+                            with open(dl_result.path, "rb") as src:
+                                filename = artifact.file.name
+                                artifact.file.save(filename, src, save=False)
+                            setattr(artifact, checksum, dl_result.artifact_attributes[checksum])
+                            restored = True
+                            break
+            else:
+                break
+        return restored
+
+    def handle(self, *args, **options):
+
+        log.setLevel(logging.ERROR)
+        hrefs = set()
+        for checksum in settings.ALLOWED_CONTENT_CHECKSUMS:
+            params = {f"{checksum}__isnull": True}
+            artifacts_qs = Artifact.objects.filter(**params)
+            for a in artifacts_qs:
+                hasher = hashlib.new(checksum)
+                try:
+                    for chunk in a.file.chunks(CHUNK_SIZE):
+                        hasher.update(chunk)
+                    a.file.close()
+                    setattr(a, checksum, hasher.hexdigest())
+                except FileNotFoundError:
+                    file_path = os.path.join(settings.MEDIA_ROOT, a.file.name)
+                    restored = self._download_artifact(a, checksum, file_path)
+                    if not restored:
+                        hrefs.add(file_path)
+
+            if artifacts_qs:
+                self.stdout.write(_("Updating artifacts with missing checksum {}").format(checksum))
+                Artifact.objects.bulk_update(objs=artifacts_qs, fields=[checksum], batch_size=1000)
+        if hrefs:
+            raise CommandError(
+                _("Some files that were missing could not be restored: {}").format(hrefs)
+            )
+
+        forbidden_checksums = set(constants.ALL_KNOWN_CONTENT_CHECKSUMS).difference(
+            settings.ALLOWED_CONTENT_CHECKSUMS
+        )
+        for checksum in forbidden_checksums:
+            search_params = {f"{checksum}__isnull": False}
+            update_params = {f"{checksum}": None}
+            artifacts_qs = Artifact.objects.filter(**search_params)
+            if artifacts_qs:
+                self.stdout.write(
+                    _("Removing forbidden checksum {} from database").format(checksum)
+                )
+                artifacts_qs.update(**update_params)
+
+        self.stdout.write(_("Finished aligning checksums with settings.ALLOWED_CONTENT_CHECKSUMS"))
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -9,6 +9,8 @@ https://docs.djangoproject.com/en/1.11/ref/settings/
 """
 
 import os
+import sys
+
 from contextlib import suppress
 from gettext import gettext as _
 from importlib import import_module
@@ -281,20 +283,41 @@ if unknown_algs:
             "provided: {}".format(unknown_algs)
         )
     )
+
+FORBIDDEN_CHECKSUMS = set(constants.ALL_KNOWN_CONTENT_CHECKSUMS).difference(
+    ALLOWED_CONTENT_CHECKSUMS
+)
 try:
-    for checksum in ALLOWED_CONTENT_CHECKSUMS:
-        # can't import Artifact here so use a direct db connection
-        with connection.cursor() as cursor:
+    with connection.cursor() as cursor:
+        for checksum in ALLOWED_CONTENT_CHECKSUMS:
+            # can't import Artifact here so use a direct db connection
             cursor.execute(f"SELECT count(pulp_id) FROM core_artifact WHERE {checksum} IS NULL")
             row = cursor.fetchone()
             if row[0] > 0:
+                if len(sys.argv) >= 2 and sys.argv[1] == "handle-artifact-checksums":
+                    break
                 raise ImproperlyConfigured(
                     _(
-                        "ALLOWED_CONTENT_CHECKSUMS cannot contain checksum '{}' since checksums "
-                        "can only be removed and not added to ALLOWED_CONTENT_CHECKSUMS in "
-                        "existing deployments."
+                        "There have been identified artifacts missing checksum '{}'. "
+                        "Run 'pulpcore-manager handle-artifact-checksums' first to populate "
+                        "missing artifact checksums."
                     ).format(checksum)
                 )
+        for checksum in FORBIDDEN_CHECKSUMS:
+            # can't import Artifact here so use a direct db connection
+            cursor.execute(f"SELECT count(pulp_id) FROM core_artifact WHERE {checksum} IS NOT NULL")
+            row = cursor.fetchone()
+            if row[0] > 0:
+                if len(sys.argv) >= 2 and sys.argv[1] == "handle-artifact-checksums":
+                    break
+                raise ImproperlyConfigured(
+                    _(
+                        "There have been identified artifacts with forbidden checksum '{}'. "
+                        "Run 'pulpcore-manager handle-artifact-checksums' first to unset "
+                        "forbidden checksums."
+                    ).format(checksum)
+                )
+
 except ImproperlyConfigured as e:
     raise e
 except Exception:
--- a/pulpcore/download/factory.py
+++ b/pulpcore/download/factory.py
@@ -80,7 +80,10 @@ class DownloaderFactory:
         }
         self._session = self._make_aiohttp_session_from_remote()
         self._semaphore = asyncio.Semaphore(value=remote.download_concurrency)
-        atexit.register(self._session.close)
+        atexit.register(self._session_cleanup)
+
+    def _session_cleanup(self):
+        asyncio.get_event_loop().run_until_complete(self._session.close())
 
     def _make_aiohttp_session_from_remote(self):
         """


pulpcore/app/management/commands/populate-missing-checksums.py
```diff
@@ -0,0 +1,57 @@
...
+            for a in artifacts_qs:
+                hasher = hashlib.new(checksum)
+                try:
+                    for chunk in a.file.chunks(CHUNK_SIZE):
+                        hasher.update(chunk)
+                    a.file.close()
+                    setattr(a, checksum, hasher.hexdigest())
+                except FileNotFoundError:
+                    if options["skip-unrecoverable-artifacts"]:
+                        setattr(a, checksum, "")
```
ipanova, line 45: this is the best i managed to come up with.

pulpcore/app/settings.py
```diff
@@ -289,11 +291,13 @@
             cursor.execute(f"SELECT count(pulp_id) FROM core_artifact WHERE {checksum} IS NULL")
             row = cursor.fetchone()
             if row[0] > 0:
+                if len(sys.argv) >= 2 and sys.argv[1] == "populate-missing-checksums":
```
mdellweg, line 294: :grinning: :+1:

pulpcore/app/management/commands/populate-missing-checksums.py
```diff
@@ -0,0 +1,57 @@
...
+            help="Will not raise error on artifacts that miss the file and are unrecoverable",
+        )
+
+    def handle(self, *args, **options):
+
+        for checksum in settings.ALLOWED_CONTENT_CHECKSUMS:
+            params = {f"{checksum}__isnull": True}
+            artifacts_qs = Artifact.objects.filter(**params)
+            # FIXME what if one artifact has multipiple checkums that are missing
+            # improve to not read file multiple times?
```
daviddavis, line 35: What about this?

```python
q = Q()
for checksum in settings.ALLOWED_CONTENT_CHECKSUMS:
    q |= Q(**{f"{checksum}__isnull": True}) 
    q |= Q(**{f"{checksum}":""})

for artifact in Artifact.objects.filter(q):
    hashers = []
    for checksum in settings.ALLOWED_CONTENT_CHECKSUMS:
        if not getattr(atrifact.checksum):
            hashers = hashlib.new(cheksum)
# ...
```

CHANGES/7561.feature
```diff
@@ -0,0 +1,3 @@
+Enabled users to add checksums to ALLOWED_CONTENT_CHECKSUMS.
+    
+Added command to populate missing checksums.
```
daviddavis, line 3: Enabled users to add checksums to ALLOWED_CONTENT_CHECKSUMS by allowing them to populate checksums with `populate-missing-checksums` command.

pulpcore/app/management/commands/populate-missing-checksums.py
daviddavis: So this will unset the checksum if the file does not exist and the user uses `--skip-unrecoverable-artifacts`?

pulpcore/app/management/commands/populate-missing-checksums.py
daviddavis: Actually I see that you wouldn't be able to do a bulk create later on though. I guess it's a trade off between fewer queries vs opening a file fewer times. I think probably using a bulk create is better since users will likely be only fixing/adding one checksum at a time.

pulpcore/app/management/commands/populate-missing-checksums.py
ipanova: this will not unset the checksum but set it to empty string.
That is to cover case when you have sha224 null but you miss the file, you will have instead sha224='' this will at least unblock you. i am hoping that once we fix this one https://pulp.plan.io/issues/7791 on the re-upload of the file we should fix checksums as well.

daviddavis:
One other thought: maybe update the error message in settings.py to let users know they can fix their system by runing this new command.

ipanova:
> One other thought: maybe update the error message in settings.py to let users know they can fix their system by runing this new command.

I have updated it but i have not given much thought on the wording
```
                raise ImproperlyConfigured(
                    _(

                        "There have been identified artifacts with missing checksum '{}'. "
                        "Run pulpcore-manager populate-missing-checksums first to populate "
                        "artifacts with checksum."
                    ).format(checksum)

```

pulpcore/app/management/commands/populate-missing-checksums.py
ipanova: yeah i agree that the chance of the artifact having more than one missing checksum is really small.

pulpcore/app/settings.py
mdellweg: You know what? You can use that same check to allow migrations.
Or maybe just only deny services (pulp-api, pulp-content, ...) to run.

daviddavis:
Sounds good. Just a couple tweaks:

```
                raise ImproperlyConfigured(
                    _(

                        "There have been identified artifacts with missing checksum '{}'. "
                        "Run 'pulpcore-manager populate-missing-checksums' first to populate "
                        "missing artifact checksums."
                    ).format(checksum)
```

pulpcore/app/settings.py
ipanova: that's good to know, but in any case this command is needed, so users can fix missing checksums as many times as needed whereas the migration will be ran just once

mdellweg: I'm just saying, it would be nice if you could also run `migrate`, `shell`, `shell_plus` and maybe `dbshell` if that tests fails, because that could be your tool to fix such issues.

daviddavis: So maybe like a denylist instead of an allowlist?

ipanova: generally we dis-advise users to directly manipulate with the db because they can bring it into such state we won't be able to fix it, or for example each user might get the db in the different state and it would be harder for us to fix and take into account so many options. I am hesitant to add those commands to the skip list because of the mentioned concerns, it gives more leeway meaning more options how to shoot yourself in the foot

pulpcore/app/management/commands/populate-missing-checksums.py
```diff
@@ -0,0 +1,55 @@
...
+                except FileNotFoundError:
+                    if options["skip-unrecoverable-artifacts"]:
+                        setattr(a, checksum, "")
+                    else:
+                        raise CommandError(
+                            _(
+                                "Some artifacts miss the file, run repair feature first. "
+                                "To skip srtifacts that can not be repaired with the repair "
+                                "feature use option --skip-unrecoverable-artifacts next time "
+                                "you run this command."
```
mdellweg, line 50: I don't think we can get away with this, as you won't be able to trigger the repair command until the digest check succeeds.

ipanova: Couple of folks met today to discuss this pr and here's some thoughts:
1) incorporate  the repair endpoint functionality in the populate-checksums command. For the artifacts that miss files and do not have remote artifact just set checksum to ''. As next step, this needs to addressed so we in general can re-upload missing/corrupted files https://pulp.plan.io/issues/7791
2) For the artifacts that miss files and do not have remote artifact  set checksum to '' and issue a message that would ask user to run global repair because some artifacts miss files( Nuance - repair feature will not recalculate the checksums and everything that was set to '' will stay so. Although next run of populate-checksums command could fix that.) As next step, this needs to addressed so we in general can re-upload missing/corrupted files https://pulp.plan.io/issues/7791

Thigs to consider:
1)Probability that a user will have the situation when artifact will miss checksum and file is quite low - how much resources we want to invest to cover that case?
2) populate-checksums command might need to be backported to 3.8

pulpcore/download/http.py
```diff
@@ -210,6 +210,6 @@ async def _run(self, extra_data=None):
             self.raise_for_status(response)
             to_return = await self._handle_response(response)
             await response.release()
-        if self._close_session_on_finalize:
+        #if self._close_session_on_finalize:
```
ipanova, line 213: @bmbouter i got this error when running the command
`pulp: asyncio:ERROR: Unclosed client session` because this flag `_close_session_on_finalize` is set to False. Do you have more background in this area?
After i have commented out this line, session does get closed but then it says it has never been awaited
`sys:1: RuntimeWarning: coroutine 'ClientSession.close' was never awaited`

bmbouter: Yes each aiohttp request needs a session. When making many requests aiohttp wants us to reuse the session for a variety of reasons (see upstream docs), so the DownloaderFactory in many cases generates one. The downloaders have a behavior  in them that will auto-create and then auto-close a session if one is not created outside of the downloader. It's in that case that `_close_session_on_finalize = True` to provide the auto-close functionality, but if one is provided outside of the downloaders it should not close it with `_close_session_on_finalize = False`.

Python watches every created co-routine to ensure it's awaited as a programmer correctness helpfulness feature. If the session is made by the downloader it needs to be closed by being awaited and Python is saying that it isn't being closed in that case.

pulpcore/app/management/commands/populate-missing-checksums.py
```diff
@@ -0,0 +1,75 @@
...
+    help = _("Populates missing checksums on the artifacts")
+
+
+    def _download_artifact(self, artifact, checksum):
+        restored = False
+        for ca in artifact.content_memberships.all():
+            if not restored:
+                for ra in ca.remoteartifact_set.all():
+                    downloader = ra.remote.get_downloader(ra)
+                    dl_result = downloader.fetch()
```
ipanova, line 30: @bmbouter i am going to leave here the steps just in case:
1.sync a repo
2. set checksum to none and remo file frol fs
2In [1]: ra=RemoteArtifact.objects.first()

In [2]: ra.content_artifact.artifact.sha1=None

In [3]: ra.content_artifact.artifact.save()

In [4]: ra.content_artifact.artifact.file
Out[4]: <FieldFile: artifact/42/983ae40ef0e9aef0c1f95aa4cd7adcc7820b5da1ccf091670eb6d77efddb23>

In [5]:                                                                                                                                                       
Do you really want to exit ([y]/n)? y
(pulp) [vagrant@pulp3-source-fedora32 _scripts]$ rm /var/lib/pulp/artifact/42/983ae40ef0e9aef0c1f95aa4cd7adcc7820b5da1ccf091670eb6d77efddb23
3. run the command

ipanova:
this PR is blocked until we figure out why this line causes the session not be closed https://github.com/pulp/pulpcore/blob/master/pulpcore/download/factory.py#L84

pulpcore/download/factory.py
```diff
@@ -81,7 +81,10 @@ def __init__(self, remote, downloader_overrides=None):
         }
         self._session = self._make_aiohttp_session_from_remote()
         self._semaphore = asyncio.Semaphore(value=remote.download_concurrency)
-        atexit.register(self._session.close)
+        atexit.register(self._session_cleanup)
```
ipanova, line 84: @bmbouter problem was in atexit not handling coroutines

pulpcore/app/management/commands/populate-missing-checksums.py
```diff
@@ -0,0 +1,82 @@
...
+
+    help = _("Populates missing checksums on the artifacts")
+
+    def _download_artifact(self, artifact, checksum):
+        restored = False
+        for ca in artifact.content_memberships.all():
+            if not restored:
+                for ra in ca.remoteartifact_set.all():
+                    remote = ra.remote.cast()
+                    if remote.policy == "immediate":
```
ipanova, line 31: added check for download policy, so we do not download on_demand content

pulpcore/download/factory.py
ipanova: this codepath probably has never worked. during sync we instantiate the session with the _close_session_on_finalize set to True so session was properly closed in _run() and not with atexit

pulpcore/app/settings.py
```diff
@@ -289,9 +289,9 @@
     ALLOWED_CONTENT_CHECKSUMS
 )
 try:
-    for checksum in ALLOWED_CONTENT_CHECKSUMS:
-        # can't import Artifact here so use a direct db connection
-        with connection.cursor() as cursor:
+    with connection.cursor() as cursor:
```
ipanova, line 292: @daviddavis is there a reason why were creating a new cursor each time instead of re-using same one?

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -0,0 +1,82 @@
...
+                    a.file.close()
+                    setattr(a, checksum, hasher.hexdigest())
+                except FileNotFoundError:
+                    restored = self._download_artifact(a, checksum)
+                    if not restored:
+                        hrefs.add(os.path.join(settings.MEDIA_ROOT, a.file.name))
+
+            if artifacts_qs:
+                self.stdout.write(_("Updating artifacts with missing checksum {}").format(checksum))
+                Artifact.objects.bulk_update(objs=artifacts_qs, fields=[checksum], batch_size=1000)
```
ipanova, line 68: I just realized, that once this command will be executed, Artifacts and RemoteArtifacts will differ. Is it ok if we will address this in the following work related to handling of RAs?

bmbouter: I think that's ok. They differ already because RemoteArtifacts store only the checksums the original metadata defined while Artifacts store all checksums available to Pulp.

CHANGES/7914.feature
```diff
@@ -0,0 +1 @@
+Added check to prevent Pulp to start if there are Artifacts with forbidden checksums.
```
daviddavis, line 1: s/to start/from starting/

daviddavis:
```suggestion
Added check to prevent Pulp from starting if there are Artifacts with forbidden checksums.
```

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -0,0 +1,82 @@
...
+import logging
+
+log = logging.getLogger("")
+
+CHUNK_SIZE = 1024 * 1024  # 1 Mb
+
+
+class Command(BaseCommand):
+    """
+    Django management command for populating missing checksums on the artifacts
```
daviddavis, line 20:
```suggestion
    Django management command for populating or removing checksums on the artifacts
```

bmbouter:
@ipanova I think if you rebase the bindings test will pass? I saw a PR fixing bindings get merged very recently IIRC.

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -0,0 +1,82 @@
...
+            )
+        self.stdout.write(_("Successfully re-populated all missing checksums"))
+
+        FORBIDDEN_CHECKSUMS = set(constants.ALL_KNOWN_CONTENT_CHECKSUMS).difference(
+            settings.ALLOWED_CONTENT_CHECKSUMS
+        )
+        for checksum in FORBIDDEN_CHECKSUMS:
+            search_params = {f"{checksum}__isnull": False}
+            update_params = {f"{checksum}": None}
+            self.stdout.write(_("Setting fobidden checksums {} to None").format(checksum))
```
daviddavis, line 81:
```suggestion
            self.stdout.write(_("Setting forbidden checksum {} to None").format(checksum))
```

pulpcore/app/settings.py
```diff
@@ -282,20 +284,41 @@
             "provided: {}".format(unknown_algs)
         )
     )
+
+FORBIDDEN_CHECKSUMS = set(constants.ALL_KNOWN_CONTENT_CHECKSUMS).difference(
```
daviddavis, line 288: I would try to avoid using a constant in this file lest it be confused for a setting.

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -0,0 +1,82 @@
...
+            artifacts_qs = Artifact.objects.filter(**search_params).update(**update_params)
```
daviddavis, line 82: Should this OR together the search_params?

pulpcore/app/settings.py
ipanova: If i make it a setting that means a user can configure it, i don't think we want the user to configure anything except the ALLOWED_CONTENT_CHECKSUMS.

daviddavis: Yes, exactly. My concern is that someone looking at this file might think it's a setting since it's a constant. Does it need to be a constant?

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -0,0 +1,82 @@
...
+    help = _("Handle missing and forbidden checksums on the artifacts")
+
+    def _download_artifact(self, artifact, checksum):
+        restored = False
+        for ca in artifact.content_memberships.all():
+            if not restored:
+                for ra in ca.remoteartifact_set.all():
+                    remote = ra.remote.cast()
+                    if remote.policy == "immediate":
+                        self.stdout.write(_("Restoring missing file."))
```
daviddavis, line 32: It might be worthwhile to output maybe the path or href of `artifact` here.

daviddavis (approved):
I tested this out and it worked great. :+1: 

I noticed one thing during test that I left a comment for.

bmbouter:
I've been testing this and it's been working well so far. Here's what I did:

1. Sync an Artifact into Pulp
2. Verify I can enter `shell_plus` (this goes through the startup check). Once in there, verify all checksums are computed on the Artifact.
3. Remove sha1 from `ALLOWED_CONTENT_CHECKSUMS`
4. Attempt to start Pulp, it refuses to start with the error telling me to use `pulpcore-manager handle-artifact-checksums`
5. I run `pulpcore-manager handle-artifact-checksums`
6. I can then use `shell_plus` and verify it removed the sha1 checksum
7. Then I add back sha1 to `ALLOWED_CONTENT_CHECKSUMS`.
8. I attempt to start Pulp, it refuses to start with the error telling me to use `pulpcore-manager handle-artifact-checksums`
9. I run `pulpcore-manager handle-artifact-checksums`
10. I can then use `shell_plus` and verify it has sha1 again.

docs/settings.rst
```diff
@@ -262,6 +262,11 @@ ALLOWED_CONTENT_CHECKSUMS
 
       ``{"md5", "sha1", "sha224", "sha256", "sha384", "sha512"}``
 
+    .. warning::
+      If Pulp fails to start because missing or forbidden checkums have been identified, run
```
bmbouter, line 266: Maybe also say "or required ones are missing" ?

bmbouter:
I was trying to test the re-download workflow, but I couldn't get it to work. I tried:

1. Sync one Artifact into Pulp
2. Delete it from the filesystem
3. Run `pulpcore-manager handle-artifact-checksums` and I expected it to be restored, but it wasn't
4. I then adjusted the checksums in `ALLOWED_CONTENT_CHECKSUMS` a bit and reran `pulpcore-manager handle-artifact-checksums` a few times but it still didn't restore.

What am I doing wrong?

ipanova:
> I was trying to test the re-download workflow, but I couldn't get it to work. I tried:
> 
>     1. Sync one Artifact into Pulp
> 
>     2. Delete it from the filesystem
> 
>     3. Run `pulpcore-manager handle-artifact-checksums` and I expected it to be restored, but it wasn't
> 
>     4. I then adjusted the checksums in `ALLOWED_CONTENT_CHECKSUMS` a bit and reran `pulpcore-manager handle-artifact-checksums` a few times but it still didn't restore.
> 
> 
> What am I doing wrong?


   

1. Sync one Artifact into Pulp
2. Delete it from the filesystem
3. Run pulpcore-manager handle-artifact-checksums and I expected it to be restored, but it wasn't (that's expected)
4. I then adjusted the checksums in ALLOWED_CONTENT_CHECKSUMS a bit ( removed sha1) and reran pulpcore-manager handle-artifact-checksums a few times but it still didn't restore.( that's expected)
5. I then adjusted the checksums in ALLOWED_CONTENT_CHECKSUMS a bit ( added back sha1) and reran pulpcore-manager handle-artifact-checksums a few times but it still didn't restore.( that's NOT expected). file should be restored, sha1 populated.

daviddavis:
This worked for me:

1. Removed a checksum from ALLOWED_CONTENT_CHECKSUMS
2. Ran the handle artifact checksums command
3. Confirmed the checksum was gone from the data model
4. Remove the file
5. Added back the checksum to ALLOWED_CONTENT_CHECKSUMS
6. Ran the handle artifact checksums command
7. Confirmed the file was on the filesystem 

The command only downloads the file if it's needed to recompute the checksum.

bmbouter:
I was testing against pulp_ansible and I think it isn't creating the records in the db correctly, but that is an issue with its code not this PR. When deleting content, and causing the command to recalculate the checksum now I get `CommandError: Some files that were missing could not be restored: {'/var/lib/pulp/artifact/36/0548ba80e3dce478b7915ca89a5613dc80c650a55b96f5491012a8297e12ac'}`

pulpcore/app/management/commands/handle-artifact-checksums.py
bmbouter: +1

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -0,0 +1,82 @@
...
+                        hrefs.add(os.path.join(settings.MEDIA_ROOT, a.file.name))
+
+            if artifacts_qs:
+                self.stdout.write(_("Updating artifacts with missing checksum {}").format(checksum))
+                Artifact.objects.bulk_update(objs=artifacts_qs, fields=[checksum], batch_size=1000)
+        if hrefs:
+            raise CommandError(
+                _("Some files that were missing could not be restored: {}").format(hrefs)
+            )
+        self.stdout.write(_("Successfully re-populated all missing checksums"))
```
bmbouter, line 73: This is printed in cases where it's not actually changing things like a no-op. So maybe something like:
`Finished aligning checksums with settings.ALLOWED_CONTENT_CHECKSUMS`.

Or perhaps print nothing?

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -0,0 +1,82 @@
...
+
+        forbidden_checksums = set(constants.ALL_KNOWN_CONTENT_CHECKSUMS).difference(
+            settings.ALLOWED_CONTENT_CHECKSUMS
+        )
+        for checksum in forbidden_checksums:
+            search_params = {f"{checksum}__isnull": False}
+            update_params = {f"{checksum}": None}
+            self.stdout.write(_("Setting forbidden checksum {} to None").format(checksum))
```
bmbouter, line 81: This may be a bit confusing to a user? Perhaps `Removing forbidden checksum {} from database`.

pulpcore/app/settings.py
```diff
@@ -281,20 +283,41 @@
...
+        for checksum in ALLOWED_CONTENT_CHECKSUMS:
+            # can't import Artifact here so use a direct db connection
             cursor.execute(f"SELECT count(pulp_id) FROM core_artifact WHERE {checksum} IS NULL")
             row = cursor.fetchone()
             if row[0] > 0:
+                if len(sys.argv) >= 2 and sys.argv[1] == "handle-artifact-checksums":
+                    break
+                raise ImproperlyConfigured(
+                    _(
+                        "There have been identified artifacts with missing checksum '{}'. "
```
bmbouter, line 301:
```suggestion
                        "There have been identified artifacts missing required checksum '{}'. "
```

docs/settings.rst
```diff
@@ -262,6 +262,11 @@ ALLOWED_CONTENT_CHECKSUMS
 
       ``{"md5", "sha1", "sha224", "sha256", "sha384", "sha512"}``
 
+    .. warning::
+      If Pulp fails to start because forbidden checkums have been identified or requires ones are
```
bmbouter, line 266:
```suggestion
      If Pulp fails to start because forbidden checkums have been identified or required ones are
```

bmbouter (approved):
Thank you so much for making this! I think our users will really appreciate it. :tada: :100:

bmbouter:
I two acks ... I'm merging.

PR #1012, opened 2020-11-06
Add a new field that holds the value of the public key to SigningService

lubosmj:
closes #7700


lubosmj:
In order to utilize this change in pulp_rpm, the following has to be done:

```diff
diff --git a/pulp_rpm/app/tasks/publishing.py b/pulp_rpm/app/tasks/publishing.py
index 86cc91e4..061febc1 100644
--- a/pulp_rpm/app/tasks/publishing.py
+++ b/pulp_rpm/app/tasks/publishing.py
@@ -5,9 +5,11 @@ import shutil
 from tempfile import NamedTemporaryFile
 
 import createrepo_c as cr
+import tempfile
 import libcomps
 
 from django.core.files import File
+from django.core.files.base import ContentFile
 from django.core.files.storage import default_storage as storage
 from django.db.models import Q
 
@@ -527,12 +529,16 @@ def create_repomd_xml(content, publication, checksum_types, extra_repomdrecords,
             file=File(open(sign_results['signature'], 'rb'))
         )
 
-        # publish a public key required for further verification
-        PublishedMetadata.create_from_file(
-            relative_path=os.path.join(repodata_path, os.path.basename(sign_results['key'])),
-            publication=publication,
-            file=File(open(sign_results['key'], 'rb'))
-        )
+        with tempfile.NamedTemporaryFile() as tf:
+            tf.write(signing_service.public_key.encode("utf-8"))
+            tf.flush()
+
+            # publish a public key required for further verification
+            PublishedMetadata.create_from_file(
+                relative_path=os.path.join(repodata_path, os.path.basename("public.key")),
+                publication=publication,
+                file=File(open(tf.name, "rb")),
+            )
     else:
         PublishedMetadata.create_from_file(
             relative_path=os.path.join(repodata_path, os.path.basename(repomd_path)),

```

pulpcore/app/serializers/content.py
```diff
@@ -260,10 +260,13 @@ class SigningServiceSerializer(base.ModelSerializer):
 
     pulp_href = base.IdentityField(view_name="signing-services-detail")
     name = serializers.CharField(help_text=_("A unique name used to recognize a script."))
+    public_key = serializers.CharField(
```
mdellweg, line 263: would it be possible, that we save the fingerprint of that key to the database too, and report that in the serializer?

pulpcore/app/models/content.py
```diff
@@ -598,6 +602,9 @@ def sign(self, filename):
         except json.JSONDecodeError:
             raise RuntimeError("The signing service script did not return valid JSON!")
 
+        if self.public_key:
+            return_value["key"] = self.public_key
```
mdellweg, line 606: WRT the deprecation policy, i believe we need to be a bit more careful here.
The old interface to the plugin returned a filename in `return_value["key"]`, so we should do the same by writing the value of this db field to such a file. I have no idea, how we can warn about the deprecation when accessing that field other than subclassing the `return_value` dict however. In that case we can make the whole "write to file" process lazy.

The "new" interface could be `return_value["public_key"]`.

mdellweg (changes requested):

CHANGES/7700.feature
```diff
@@ -0,0 +1 @@
+Added a new field used for storing the value of the public key in SigningService.
```
bmbouter, line 1: Can the field name be added here please

pulpcore/app/models/content.py
```diff
@@ -560,21 +561,24 @@ class SigningService(BaseModel):
...
+        public_key (models.TextField):
+            An absolute path to a public key.
         script (models.TextField):
-            An absolute path to the external signing script (or executable).
+            An absolute path to an external signing script (or executable).
 
     """
 
     name = models.TextField(db_index=True, unique=True)
+    public_key = models.TextField(default="")
```
bmbouter, line 573: I imagined this would be a required field. All SigningServices use PKI which use public keys is what I was thinking.

pulpcore/app/models/content.py
```diff
@@ -560,21 +561,24 @@ class SigningService(BaseModel):
 
     Fields:
         name (models.TextField):
-            A unique name describing the script (or executable) used for signing.
+            A unique name describing a script (or executable) used for signing.
+        public_key (models.TextField):
+            An absolute path to a public key.
         script (models.TextField):
-            An absolute path to the external signing script (or executable).
+            An absolute path to an external signing script (or executable).
```
bmbouter, line 568: This should be updated to store the key itself, not a path.

pulpcore/app/models/content.py
```diff
@@ -560,21 +561,24 @@ class SigningService(BaseModel):
 
     Fields:
         name (models.TextField):
-            A unique name describing the script (or executable) used for signing.
+            A unique name describing a script (or executable) used for signing.
+        public_key (models.TextField):
+            An absolute path to a public key.
```
bmbouter, line 566: This should be updated to store the key itself, not a path.

pulpcore/app/models/content.py
```diff
@@ -619,6 +626,15 @@ def save(self, *args, **kwargs):
         Save a signing service to the database (unless it fails to validate).
         """
         self.validate()
+
+        if not self.public_key:
```
bmbouter, line 630: I imagine all new SigningServices created will require users to put in a public key with this change. I don't think this needs to go through the deprecation cycle since plugin writers had a plugin key returned to them before and they still will. The deprecation policy I didn't imagine applied to the "scripts", users upgrading will need to take action at that time to recreate their SigningServices. This is my take at least.

pulpcore/app/serializers/content.py
bmbouter: This would be good, but also we could add this later?

pulpcore/app/models/content.py
mdellweg: I think this is a migration thing. What about existing signing_services?
IIRC django won't even let you add a required field without providing a default.

bmbouter: With 3.9 signing servies are immutable so I think we've already gotten to the point where users will need to make them new.

pulpcore/app/models/content.py
mdellweg: This is about saving new SigningServices? Yes, they should be up to date. And we should not allow to create deprecated stuff in the database.
I would expect to see this warning whenever existing preexisting ones were used.

pulpcore/app/models/content.py
mdellweg: So yes to required. Can we also not default it to anything and provide a more verbose default only for the migration?

bmbouter: We could provide a migration which sets the value at migration time. It could sidestep the hook by calling https://rsinger86.github.io/django-lifecycle/advanced/#suppressing-hooked-methods in the data migration itself. What do you think about that idea?

mdellweg: I thought django would ask you for a migration default if you don't have it on the model.

bmbouter: I was imaging a migraiton that has 3 parts:

1. Add the field but make it optional
2. Perform the data migration populating the field
3. SQL update to make that field required (now that all existing data also has it). This last part would "match" the model code marking the field as required.

lubosmj: I do not quite understand what are you trying to achieve with those 3 parts, @bmbouter. How do you want to populate the data from an old SigningService object if there was not any `public_key` present back then? It will contain the same data like before. Do you want to extract the value of the public key by executing a signing script and manually reading the value of `return_value["key"]` within the migration?

bmbouter: > I do not quite understand what are you trying to achieve with those 3 parts, @bmbouter. How do you want to populate the data from an old SigningService object if there was not any `public_key` present back then? It will contain the same data like before. Do you want to extract the value of the public key by executing a signing script and manually reading the value of `return_value["key"]` within the migration?

That's what I was thinking yes. If the migration fails, I think it should probably put in dummy data and log to the user an error and print the error to stderr as well so it's shown in the foreground too. We can't have the migraitons "fail to apply"

@mdellweg what do you think about ^?

mdellweg: Agreed, we must make the migration apply cleanly at any cost.
Trying to extract the public key is a convenience we offer the user, but it could fail for various reasons (revoked/expired keys is probably just one of them).
So i think, we should make the failing units unusable somehow to force the user to reinstanciate them. Having dummy data in the key field breaks the promise we give our plugins IMHO.

bmbouter: @mdellweg agreed. In terms of making the failing units unusable, I was thinking it would play out like:

1. The migration logs them and puts an empty string in the public_key field.
2. Then the SQL alter to make public_key required can be applied without error and the transaction can finish.
3. The code expects the public_key data to be accurate and so in cases where it's just `''` the code will receive a RuntimeError. Or perhaps it could even check if its '' and gracefully refuse. Either way is ok with me.

bmbouter, line 573: I don't expect this to have a default. It's a required field so it can't have one.

pulpcore/app/models/content.py
bmbouter: With [this comment](https://github.com/pulp/pulpcore/pull/1012/files#r523015114) I was thinking we would not be ever getting the public_key from the script anymore.

pulpcore/app/models/content.py
```diff
@@ -658,19 +674,21 @@ def validate(self):
...
-                            raise RuntimeError(
-                                "A signature could not be verified or a trust level is too low. "
-                                "The signing script may generate invalid signatures."
-                            )
-                        elif verified.pubkey_fingerprint != import_result.fingerprints[0]:
-                            raise RuntimeError(
-                                "Fingerprints of a provided public key and a verified public key "
-                                "are not equal. The signing script is probably not valid."
-                            )
+                    if self.public_key:
```
bmbouter, line 677: Similarly here, I was thinking [with this comment](https://github.com/pulp/pulpcore/pull/1012/files#r523015114) we are always getting the key from the database and not from the signing script ever.

pulpcore/app/models/content.py
```diff
@@ -658,19 +674,21 @@ def validate(self):
...
+                    import_result = gpg.import_keys(key_value)
+                    verified = gpg.verify_file(fp, temp_file.name)
+                    if verified.trust_level is None or verified.trust_level < verified.TRUST_FULLY:
+                        raise RuntimeError(
+                            "The signature could not be verified or the trust level is too "
+                            "low. The signing script may generate invalid signatures."
+                        )
+                    elif verified.pubkey_fingerprint != import_result.fingerprints[0]:
+                        raise RuntimeError(
+                            "Fingerprints of the provided public key and the verified public "
```
bmbouter, line 692: This should be i8ln with `gettext`.

lubosmj:
I addressed the requested changes. Please take a look at the PR again, @mdellweg and @bmbouter. I tested the PR on pulp_rpm (with the changes from the comment https://github.com/pulp/pulpcore/pull/1012#issuecomment-723228250 and without them) and everything seems to be working.

CHANGES/plugin_api/7700.deprecation
```diff
@@ -0,0 +1,2 @@
+Access to the path of the public key of a signing service was deprecated. The value of the public
+key is now directly accessible via the model instance as SigningService.public_key.
```
bmbouter, line 2:
```suggestion
key is now expected to be saved in the model instance as `SigningService.public_key`.
```

docs/workflows/signed-metadata.rst
```diff
@@ -87,9 +96,10 @@ The example below demonstrates how a signing service can be created using ``gpg`
...
+                   "pulp_created": "2020-11-06T15:42:20.645197Z",
```
mdellweg, line 100: Should we shorten this in the docs with `-----BEGIN PGP PUBLIC KEY BLOCK-----\n\n [...] \n-----END PGP PUBLIC KEY BLOCK-----\n`?
I don't see much value for the reader to see the raw key material.

pulpcore/app/migrations/0053_add_public_key.py
```diff
@@ -0,0 +1,68 @@
...
+            temp_file.write(b'arbitrary data')
+            temp_file.flush()
+
+            for service in SigningService.objects.all():
+                try:
+                    sign_results = sign(service, temp_file.name)
+                except RuntimeError:
+                    warnings.warn(
+                        "The public key migration for '{}' failed. Consider creating a new signing "
+                        "service instead.".format(service.name),
```
mdellweg, line 41: If this migration failed, will the resource be non functional? We should add this to the warning.

pulpcore/app/migrations/0053_add_public_key.py
```diff
@@ -0,0 +1,68 @@
...
+                        RuntimeWarning,
+                    )
+                else:
+                    with open(sign_results['key'], 'rb') as key:
+                        service.public_key = key.read()
```
mdellweg, line 46: Should we add another try-warn safety layer here, too?

pulpcore/app/models/content.py
```diff
@@ -602,11 +606,19 @@ def sign(self, filename):
...
             raise RuntimeError("The signing service script did not return valid JSON!")
 
+        return_value["key"] = self.public_key
+
+        msg = _(
+            "The public key will no longer be accessible via 'return_value'. This is way is "
+            "deprecated as of the next release (3.10.0). Use 'signing_service.public_key' instead."
+        )
+        warnings.warn(msg, DeprecationWarning)
+
```
mdellweg, line 616: Ideally this warning will only be triggered when someone accesses this key in the dict.
Is it worth subclassing dict for this?

pulpcore/app/serializers/content.py
mdellweg: It may be beneficial to do it now, to not have to "calculate" it in an upcoming migration.

mdellweg (commented):
I don't have any major concerns with the current state. Just some ideas.
Also the deprecation window has probably moved a version.

lubosmj:
@mdellweg, fingerprints are now calculated during the migration. Have a look.

pulpcore/app/migrations/0053_add_public_key.py
```diff
@@ -0,0 +1,100 @@
...
+        "be valid.".format(service.name),
+        RuntimeWarning,
+    )
+
+
+def sign_temp_file(service, temp_file):
+    try:
+        return sign(service, temp_file.name)
+    except RuntimeError:
+        print_warning(service)
```
mdellweg, line 43: If you do not reraise this exception, the function will return `None` leading to `update_model_fields` to fail with a `KeyError`. I suggest to reraise the exception and catch-ignore it in the loop in line 86, so so subsequent services are still attempted to be migrated, but the whole migration doesn't throw any thing.

pulpcore/app/models/content.py
```diff
@@ -564,21 +565,42 @@ class SigningService(BaseModel):
...
+                "deprecated as of the next release (3.10.0). Use 'signing_service.public_key' "
+                "instead."
+            )
+
+        def __getitem__(self, item):
+            if item == "key":
+                # raise the deprecation warning when a plugin writer tries to access the public key
+                warnings.warn(self.msg, DeprecationWarning)
+
+            return super().__getitem__(item)
```
mdellweg, line 596: This is exactly what i meant.
I feel like i created a monster. ;)

pulpcore/app/migrations/0053_add_public_key.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def update_model_fields(sign_results, service, gpg):
+    try:
+        with open(sign_results['key'], 'rb') as key:
+            public_key = key.read()
+            service.public_key = public_key
+
+            import_result = gpg.import_keys(public_key)
+            service.pubkey_fingerprint = import_result.fingerprints[0]
```
mdellweg, line 53: :+1:

mdellweg (changes requested):

pulpcore/app/migrations/0053_add_public_key.py
```diff
@@ -0,0 +1,100 @@
...
+            model_name='signingservice',
+            name='public_key',
+            field=models.TextField(default=''),
+        ),
+        migrations.AddField(
+            model_name='signingservice',
+            name='pubkey_fingerprint',
+            field=models.TextField(default=''),
+        ),
+        migrations.RunPython(migrate_public_key_values),
```
mdellweg, line 89: Can you please provide a down migration here?

https://docs.djangoproject.com/en/3.1/ref/migration-operations/#runpython

lubosmj: I do not quite understand what you want me to do. What is the down migration?

lubosmj: Were you referring to the `reverse_code` parameter?

mdellweg: Yes, exactly. And it is probably an empty lambda expression, but it allows to rollback migrations.

lubosmj: I added `migrations.RunPython.noop`.

pulpcore/app/migrations/0053_add_public_key.py
```diff
@@ -0,0 +1,105 @@
...
+        raise RuntimeError()
+
+    return return_value
+
+
+def print_warning(service):
+    warnings.warn(
+        "The public key migration for '{}' failed. Consider creating a new signing "
+        "service instead and try to run the migration again. The resource will not "
+        "be valid.".format(service.name),
```
mdellweg, line 34: You cannot "run the migration again".
How about:
```suggestion
        "The public key migration for '{}' failed.  The resource will not "
        "be valid. Consider creating a new signing service instead."
        "".format(service.name),
```

pulpcore/app/migrations/0053_add_public_key.py
```diff
@@ -0,0 +1,105 @@
...
+            temp_file.flush()
+
+            for service in SigningService.objects.all():
+                try:
+                    sign_results = sign_temp_file(service, temp_file)
+                except RuntimeError:
+                    pass
+                else:
+                    update_model_fields(sign_results, service, gpg)
+
```
mdellweg, line 76: There's a newline missing.

pulpcore/app/migrations/0053_add_public_key.py
```diff
@@ -0,0 +1,105 @@
...
+        migrations.RunPython(migrate_public_key_values, reverse_code=migrations.RunPython.noop),
+        migrations.AlterField(
+            model_name='signingservice',
+            name='public_key',
+            field=models.TextField(),
+        ),
+        migrations.AlterField(
+            model_name='signingservice',
+            name='pubkey_fingerprint',
+            field=models.TextField(),
```
mdellweg, line 103: What happens if two signing services failed to migrate?
Is it correct that this is not a problem (?) because:
* the fields are not unique
* the database allows the fields to stay empty (but the services proved to be unusable anyway)
* removing the default makes the field required for new services

Remember: As long as the database is in a "supported state" (including misconfigured singing services) the migration must pass without exposing an uncaught exception. And i heard, we do not allow migrations to be touched later.

mdellweg (approved):
I believe, this now catches all the exceptions we could encounter.

For the sake of simplicity i'd rather have a single try-ecxept-warn-pass around the best-effort migration of every single singing service. But i'm not going to block this PR there.

Still a second pair of eyes on the db-migration would be good. @bmbouter

pulpcore/app/models/content.py
```diff
@@ -622,6 +645,15 @@ def save(self, *args, **kwargs):
         """
         Save a signing service to the database (unless it fails to validate).
         """
+        if not self.public_key:
+            raise RuntimeError(
+                _(
+                    "The public key needs to be specified during the new instance creation. Other "
+                    "ways of providing the public key are considered deprecated as of the release "
+                    "(3.9.0)."
```
bmbouter, line 653: This needs to show 3.10.0 I think right?

pulpcore/app/models/content.py
```diff
@@ -564,21 +565,42 @@ class SigningService(BaseModel):
...
+            An absolute path to an external signing script (or executable).
 
     """
 
     name = models.TextField(db_index=True, unique=True)
+    public_key = models.TextField()
+    pubkey_fingerprint = models.TextField()
     script = models.TextField()
 
+    class _WarningsDict(dict):
```
bmbouter, line 581: Please move this class definition to the module level. It's ok to still call it `_WarningsDict` though.

pulpcore/app/models/content.py
bmbouter: Also all the `3.10.0` strings in this PR can just say `3.10` I think.

bmbouter (changes requested):
@lubosmj this is excellent. I'm requesting a few small changes and then I'll lgtm.

Please fix the version deprecation strings to be 3.10 and move the _WarningsDict to the module level and we'll be done!

bmbouter (approved):
Thank you @lubosmj ! This is excellent! I really appreciate all the effort you've put into this.

PR #1022, opened 2020-11-12
Minor resync perf improvement

dralley:
10% of resync time (file plugin) was spent inside these tight loops that we can replace with cheaper dictionary lookups.  No queries changed.

Time spent inside Python code of the QueryExistingContent stage (excluding function calls) drops from 11.4% to 0.4%

[noissue]

### DIFF
--- a/pulpcore/plugin/stages/content_stages.py
+++ b/pulpcore/plugin/stages/content_stages.py
@@ -38,27 +38,19 @@ class QueryExistingContents(Stage):
         """
         async for batch in self.batches():
             content_q_by_type = defaultdict(lambda: Q(pk__in=[]))
+            d_content_by_nat_key = defaultdict(list)
             for d_content in batch:
                 if d_content.content._state.adding:
                     model_type = type(d_content.content)
                     unit_q = d_content.content.q()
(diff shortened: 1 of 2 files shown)


dralley:
~10-15% improvement to resync times, ~3-4% improvement to initial sync times

pulpcore/plugin/stages/content_stages.py
```diff
@@ -38,27 +38,19 @@ async def run(self):
...
-                            continue
-                        not_same_unit = False
-                        for field in type_natural_key_fields:
-                            in_memory_digest_value = getattr(d_content.content, field)
-                            if in_memory_digest_value != getattr(result, field):
-                                not_same_unit = True
-                                break
-                        if not_same_unit:
-                            continue
+                    for d_content in d_content_by_nat_key[result.natural_key()]:
```
mdellweg, line 51: Yay for using the information we already had!
This should especially scale very well with larger batch sizes.

mdellweg (approved):

dralley:
@daviddavis or @bmbouter could one of you review also?

bmbouter (approved):
This makes so much more sense. lgtm, great! :cake:

PR #1023, opened 2020-11-12
Fix test that causes /tmp to be deleted

daviddavis:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/test_filesystemexport.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_filesystemexport.py
@@ -157,8 +157,8 @@ class FilesystemExporterTestCase(BaseExporterCase):
 
     def test_delete(self):
         """Delete a pulpExporter."""
-        exporter = self.exporter_api.create({"name": "test", "path": "/tmp"})
-        self._delete_exporter(exporter)
+        exporter = self.exporter_api.create({"name": "test", "path": "/tmp/abc"})
+        self.exporter_api.delete(exporter.pulp_href)
         with self.assertRaises(ApiException) as ae:
             self.exporter_api.read(exporter.pulp_href)
         self.assertEqual(404, ae.exception.status)


ggainey (approved):
Looks like the right thing to do for this test, thanks!

mdellweg:
I'm more concerned that pulp seems to be able to delete tmp at all.
Should we maybe enforce that an export directory must have at least one path component added to what is in `ALLOWED_EXPORT_PATHS`?

daviddavis:
You say pulp but it's actually [the test](https://github.com/pulp/pulpcore/blob/11d65657667f77a29aa835e83bfa052d2a7f444d/pulpcore/tests/functional/api/using_plugin/test_filesystemexport.py#L106) that's removing the directory. Not sure if you understood that.

ggainey:
Yeah, there's not much we can do about a test that issues "sudo rm -rf /tmp"

mdellweg:
You are referring to here?
https://github.com/pulp/pulpcore/pull/1023/files#diff-fad902aff06ba0d737e00cfff5643e7273887b43d2330b0f298a616f689d71feL106
Yep, then i was wrong thinking pulp deleted the directory with the exporter.

pulpcore/tests/functional/api/using_plugin/test_filesystemexport.py
```diff
@@ -157,8 +157,8 @@ def test_list(self):
 
     def test_delete(self):
         """Delete a pulpExporter."""
-        exporter = self.exporter_api.create({"name": "test", "path": "/tmp"})
-        self._delete_exporter(exporter)
+        exporter = self.exporter_api.create({"name": "test", "path": "/tmp/abc"})
+        self.exporter_api.delete(exporter.pulp_href)
```
mdellweg, line 161: Now it looks like changing either of those two lines would solve the problem.

ggainey: Yes - "belt and suspenders", make the path be a directory inside-tmp, and don't use the function-with-side-effects since this particular teest doesn't ever create an export (and hence doesn't need to clean it up).

mdellweg (approved):

ipanova (approved):

PR #1024, opened 2020-11-12
Adding check that exporter path exists

### LINKED ISSUE #7829: Add domain support to TaskSchedule
`TaskSchedule` has no domain field - `dispatch_scheduled_tasks()` always dispatches tasks in the default domain context. Add `pulp_domain` to `TaskSchedule` so scheduled tasks run in the correct domain.

daviddavis:
Adding check that exporter path exists before trying to create and modify it.

fixes #7829

### DIFF
--- a/pulpcore/app/tasks/export.py
+++ b/pulpcore/app/tasks/export.py
@@ -168,8 +168,9 @@ def pulp_export(the_export):
 
         tarfile_fp = the_export.export_tarfile_path()
 
-        os.makedirs(pulp_exporter.path, exist_ok=True)
-        os.chmod(pulp_exporter.path, 0o775)  # let owner and group read and write the directory
+        path = Path(pulp_exporter.path)
+        if not path.is_dir():
+            path.mkdir(mode=0o775, parents=True)
 
         rslts = {}
         if the_export.validated_chunk_size:
--- a/CHANGES/7829.bugfix
+++ b/CHANGES/7829.bugfix
@@ -1,0 +1,2 @@
+Fixed bug where exporter directory existed and was writable but not owned by worker process and thus
+not chmod-able.


pulpcore/app/tasks/export.py
```diff
@@ -168,8 +168,9 @@ def pulp_export(the_export):
 
         tarfile_fp = the_export.export_tarfile_path()
 
-        os.makedirs(pulp_exporter.path, exist_ok=True)
-        os.chmod(pulp_exporter.path, 0o775)  # let owner and group read and write the directory
+        path = Path(pulp_exporter.path)
+        if not path.is_dir():
+            path.mkdir(mode=0o775, parents=True)
```
mdellweg, line 173: The only difference i see in this change is that the chmod part is only applied, when the directory does not already exist.
(And maybe closing a race condition window by setting the mode while creating).
That is not exactly matching how i read the changelog.
Also, how about first making sure the directories exist, and then executing the chmod only if we don't have the necessary permissions?
Paraphrasing:
```
        path = Path(pulp_exporter.path)
        if not path.is_dir():
            path.mkdir(mode=0o775, parents=True)
        elif cannot_write:
            os.chmod(path, 0o775)
```
If the chmod fails, it shall not be.

daviddavis: The problem I am solving is that the directory exists and is writable but the worker doesn’t own it and thus can’t chmod it. Agreed that the change log and commit message need to be updated to convey that. 

If the directory exists though and is not writable, I think it should just fail and not try to chmod it. But I can be convinced otherwise. The initial problem this chmod tried to solve was that pulp was creating directories that weren’t writable.

mdellweg: Yes, you are probably right that we should not touch permissions on directories already existing.

mdellweg (approved):

ipanova (approved):

PR #1026, opened 2020-11-13
Added fields controlling connection-timeouts to Remote.

### LINKED ISSUE #7201: Support timeout customization for synchronous File upload API
I am frustrated when attempting to use the synchronous file upload endpoint (`post/pulp/api/v3/content/file/files/upload/`) to fetch large files from external URLs. Currently, the API does not allow users to specify custom timeout parameters.

When pulling large datasets—for example, the OpenStreetMap planet file at `https://planet.openstreetmap.org/pbf/planet-latest.osm.pbf`—the connection exceeds the default `aiohttp` limits, causing the upload to fail without any way for the user to extend the grace period for high-latency or high-volume transfers.

I would like the `post/pulp/api/v3/content/file/files/upload/` endpoint to accept the following optional fields in its request body to align with other Pulp download behaviors:

* **`total_timeout`**: (number or null) Customizes `aiohttp.ClientTimeout.total`.
* **`connect_timeout`**: (number or null) Customizes `aiohttp.ClientTimeout.connect`.
* **`sock_connect_timeout`**: (number or null) Customizes `aiohttp.ClientTimeout.sock_connect`.
* **`sock_read_timeout`**: (number or null) Customizes `aiohttp.ClientTimeout.sock_read`.

These should default to `null`, allowing the system to fall back to standard `aiohttp` defaults unless explicitly overridden by the user.

I considered creating a **Remote** to handle the download, but this is not possible in this use case. A File Remote requires a PULP_MANIFEST file to be present at the source to index the content. The source in question (e.g., OpenStreetMap) provides direct links to large binary blobs without a Pulp-compatible manifest, making the synchronous file_url upload the only viable method.

**Additional context**
This feature would bring parity between the synchronous File upload API and other download-related tasks in Pulp that already utilize `aiohttp` timeout configurations.

Example payload structure requested:

```json
{
  "file_url": "https://planet.openstreetmap.org/pbf/planet-latest.osm.pbf",
  "relative_path": "planet-latest.osm.pbf",
  "total_timeout": 3600,
  "sock_read_timeout": 600
}

```

ggainey:
closes #7201


pulpcore/download/factory.py
```diff
@@ -118,8 +120,17 @@ def _make_aiohttp_session_from_remote(self):
         headers = {"User-Agent": user_agent()}
 
         conn = aiohttp.TCPConnector(**tcp_conn_opts)
+        total = self._remote.total_timeout
+        if self._remote.sock_connect_timeout:
+            sock_connect = self._remote.sock_connect_timeout
+        else:
+            sock_connect = settings.DEFAULT_CONNECTION_TIMEOUT_SECONDS
```
mdellweg, line 127:
```suggestion
        sock_connect = self._remote.sock_connect_timeout or settings.DEFAULT_CONNECTION_TIMEOUT_SECONDS
```
Lazy evaluation ;)

ggainey: Elegance is a good thing :)

pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
```diff
@@ -192,3 +197,92 @@ def test_negative_create_repo_with_invalid_parameter(self):
...
+        # Compare initial-attrs vs remote created in setUp
+        self._compare_results(self.remote_attrs, self.remote)
+
+    def test_update(self):
+        data = {"download_concurrency": 66, "policy": "immediate"}
+        self.client.patch(self.remote["pulp_href"], data)
+        new_remote = self.client.get(self.remote["pulp_href"])
+        self._compare_results(data, new_remote)
+
+    def test_timeout_attributes(self):
```
bmbouter, line 246: Can this test be broken up into different indivudal tests?

pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
```diff
@@ -192,3 +197,92 @@ def test_negative_create_repo_with_invalid_parameter(self):
         response = api.Client(self.cfg, api.echo_handler).post(FILE_REPO_PATH, gen_repo(foo="bar"))
         assert response.status_code == 400
         assert response.json()["foo"] == ["Unexpected field"]
+
+
+class CRUDRemoteTestCase(unittest.TestCase):
```
bmbouter, line 202: It would be really great if these were bindings tests instead of `requests` API calls. We get more value out of testing the bindings because it ensures the api schema producing those bindings is also correct. Most of our existing tests are `requests` based, but net-new tests I hope can all be bindings based. We've been porting pulp_ansible to bindings tests a lot recently, here's an example: https://github.com/pulp/pulp_ansible/pull/423/files#diff-d8871ad5c606deb92237ce777a88530a3522c2ae112096a0e105fb42f8337cf7R71

docs/settings.rst
```diff
@@ -266,3 +266,12 @@ ADMIN_SITE_URL
 ^^^^^^^^^^^^^^
 
     The Django admin site URL. Defaults to ``admin/``.
+
+DEFAULT_CONNECTION_TIMEOUT_SECONDS
```
bmbouter, line 270: I didn't expect a system-wide default setting. I imagined we would defer to the defaults in aiohttp in all cases.

CHANGES/7201.feature
```diff
@@ -0,0 +1 @@
+Exposed aiohttp.ClientTimeout fields in Remote.
```
bmbouter, line 1: It would be cool if this called out the fields by name each wrapped in doublebackticks (sphinx escaping format).

bmbouter (changes requested):
Despite my variety of comments, this PR is bigtime going in the right direction.

docs/settings.rst
ggainey: Current behavior is a hardcoded sock_connect=600. aiohttp default is 300. Letting the admin override the hardcoded default is better than what we have now.

Also, being on the wrong side of a bad upstream connection means *all* your remotes are going to have problems - it would be nice to let the admin set a global value, rather than having to set every single remote individually. Thinking about it, I'd suggest letting the admin set *all* the ClientTimeout values here, with either empty or 0 meaning "take the aiohttp defaults".  Thoughts?

pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
ggainey: Makes sense, can do

pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
ggainey: Sure, makes sense

docs/settings.rst
ggainey: Just realized current-behavior also hardcodes sock_read=600 as well.

CHANGES/7201.feature
ggainey: good idea

docs/settings.rst
ggainey: I added DEFAULT_SOCKET_READ_TIMEOUT_SECONDS so that we out of the box recreate the current situation (sock_read and connect default to 600s). Still waiting on "should we allow instance-wide defaults for all four ClientTimeout settings"

pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
ggainey: and done!

bmbouter: thank you!

docs/settings.rst
bmbouter: I want to make a case, shocking as it may be
That we defer default values to aiohttp
Previous pulp versions I know
Set them to 600 though,
I do not think it should be

Verse aside, I want us to seriously consider changing these. Those values would be 300 seconds, not 600. Here's the reasoning:

* We can change code-defaults without breaking semver which applies strictly to integration points with Pulp like the API
* I don't see a reason why these values are better than the aiohttp defaults. They do more downloading than we do.
* Pulp should introduce features as requested, global options like these weren't yet requested
* There is precedent for this, look at how this was changed with the download concurrency. If there is a desire for users to "get back their heavior" they can use a mass-set update like `pulpcore-manager dbshell -c SQL COMMANDS HERE`.

ggainey: I don't have a huge issue with not-doing this, honestly - as long as what we do instead is "not hardcode 'our' defaults", and just let aiohttp be controlling unless the remote is explicitly changed". I have an allergy to setting things like this in code.

So to be clear - the proposal is "remove DEFAULT_SOCKET_READ_TIMEOUT_SECONDS and DEFAULT_CONNECTION_TIMEOUT_SECONDS , **and** the 600-second-hard-coded-override, and let aiohttp win", yes?

bmbouter: yes, that's what I'm proposing. Maybe save a copy of your existing commit in case other reviewers feel differently. But yes, I would lgtm that.

bmbouter: Also maybe a .deprecated release note calling out the 600 -> 300 timeout changes.

ggainey: Sounds like a fine plan :)

CHANGES/7201.feature
```diff
@@ -0,0 +1,5 @@
+Exposed ``aiohttp.ClientTimeout`` fields in ``Remote`` as ``connect_timeout``,
+``sock_connect_timeout``, ``sock_read_timeout``, and ``total_timeout``.
+
+This replaces the previous hard-coded 600 second timeout for sock_connect and sock_read,
```
bmbouter, line 4: :+1:

pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
```diff
@@ -192,3 +203,98 @@ def test_negative_create_repo_with_invalid_parameter(self):
...
+            "sock_read_timeout": None,
+        }
+        self.remotes_api.partial_update(self.remote.pulp_href, data)
+        time.sleep(1)
+        new_remote = self.remotes_api.read(self.remote.pulp_href)
+        self._compare_results(data, new_remote)
+
+    def test_delete(self):
+        self.remotes_api.delete(self.remote.pulp_href)
+        time.sleep(1)
```
bmbouter, line 297: Can we switch these sleeps to `monitor_task`? You can see an example of it here https://github.com/pulp/pulpcore/blob/98c591ae46a6350e4c96e325f3e1b5e3362c11bd/pulpcore/tests/functional/utils.py#L62

ggainey: > Can we switch these sleeps to `monitor_task`? You can see an example of it here

Ahh perfect - much better, will do!

ggainey: Done!

bmbouter (approved):
This looks great to me. Thank you so much @ggainey !

goosemania (approved):
Thanks!

PR #1027, opened 2020-11-14
Added pre_save hook to Artifact to enforce checksum rules.

ggainey:
Closes #7696

### DIFF
--- a/CHANGES/plugin_api/7696.feature
+++ b/CHANGES/plugin_api/7696.feature
@@ -1,0 +1,1 @@
+Added pre_save hook to Artifact to enforce checksum rules implied by ALLOWED_CONTENT_CHECKSUMS.
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -18,13 +18,14 @@ from django.core import validators
 from django.core.files.storage import default_storage
 from django.db import IntegrityError, models, transaction
 from django.forms.models import model_to_dict
-from django_lifecycle import BEFORE_UPDATE, hook
+from django_lifecycle import BEFORE_UPDATE, BEFORE_SAVE, hook
 
 from pulpcore.constants import ALL_KNOWN_CONTENT_CHECKSUMS
 from pulpcore.app.models import MasterModel, BaseModel, fields, storage
 from pulpcore.exceptions import (
     DigestValidationError,
     SizeValidationError,
+    MissingDigestValidationError,
 )
 
 # All available digest fields ordered by algorithm strength.
@@ -202,24 +203,34 @@ class Artifact(HandleTempFilesMixin, BaseModel):
     # Digest-fields that are NOT ALLOWED
     FORBIDDEN_DIGESTS = _FORBIDDEN_DIGESTS
 
-    def __init__(self, *args, **kwargs):
+    @hook(BEFORE_SAVE)
+    def before_save(self):
         """
-        Initialization override to reject use of disallowed checksum algorithms.
+        Pre-save hook that validates checksum rules prior to allowing an Artifact to be saved.
 
-        Args:
-            args: positional arguments
-            kwargs: keyword arguments
+        An Artifact with any checksums from the FORBIDDEN set will fail to save while raising
+        an UnsupportedDigestValidationError exception.
+
+        Similarly, any checksums in DIGEST_FIELDS that is NOT set will raise a
+        MissingDigestValidationError exception.
 
         Raises:
             :class: `~pulpcore.exceptions.UnsupportedDigestValidationError`: When any of the
-                keys in kwargs are found on the FORBIDDEN_DIGESTS list
+                keys on FORBIDDEN_DIGESTS are set for the Artifact
+            :class: `~pulpcore.exceptions.MissingDigestValidationError`: When any of the
+                keys on DIGEST_FIELDS are found to be missing from the Artifact
         """
-        bad_keys = self.FORBIDDEN_DIGESTS.intersection(kwargs.keys())
+        bad_keys = [k for k in self.FORBIDDEN_DIGESTS if getattr(self, k)]
         if bad_keys:
             raise UnsupportedDigestValidationError(
                 _(f"Checksum algorithms {bad_keys} are forbidden for this Pulp instance.")
             )
-        super().__init__(*args, **kwargs)
+
+        missing_keys = [k for k in self.DIGEST_FIELDS if not getattr(self, k)]
+        if missing_keys:
+            raise MissingDigestValidationError(
+                _(f"Missing required checksum algorithms {missing_keys}.")
+            )
 
     def q(self):
         if not self._state.adding:
--- a/pulpcore/tests/unit/models/test_content.py
+++ b/pulpcore/tests/unit/models/test_content.py
@@ -1,9 +1,13 @@
 import os
 import tempfile
+from unittest import mock
 
 from django.core.files.storage import default_storage as storage
+from django.core.files.uploadedfile import SimpleUploadedFile
+
 from django.conf import settings
-from django.test import TestCase
+from django.test import SimpleTestCase, TestCase
+from pulpcore.plugin.exceptions import MissingDigestValidationError
 from pulpcore.plugin.models import (
     Artifact,
     Content,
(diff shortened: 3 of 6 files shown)


pulpcore/app/models/content.py
```diff
@@ -206,6 +214,9 @@ def __init__(self, *args, **kwargs):
         """
         Initialization override to reject use of disallowed checksum algorithms.
 
+        Note that this will also be saved in the pre_save hook below. It's generally a good thing
+        to catch errors as close to where they're made as possible, so we look here as well.
+
```
mdellweg, line 219: I'm not entirely sure, this is worth the computing power, as it is still possible to add forbidden fields between initialization and saving of the artifact.
Given Artifacts are probably the most central entity in Pulp, this could trade a measurable performance hit for not being safe on the other side.
The before_safe hook is really our parachute here.

pulpcore/tests/unit/models/test_content.py
```diff
@@ -80,3 +83,35 @@ def test_set_forbidden(self):
                 a = Artifact(md5="asdf")  # noqa
         else:
             pass
+
+    def test_direct_set_forbidden(self):
+        # This will only fire on a Pulp instance that has forbidden md5 in settings.py
```
mdellweg, line 88: This is a unittest. Do you think you can mock that value there?

pulpcore/app/models/content.py
ggainey: The cost we're talking about here, is doing a set-intersection on 2 sets of <10 elements, and a single 'if' - microseconds, literally, at worst. If your repo has a million Artifacts, you might add ones of milliseconds to a sync.

There are definitely cases we have to catch at save-time, but the majority-use is "provide the checksums at init". Having this check here, means catching those immediately, rather than going through "manipulate the Artifact and whatever points to it and then fail at save-time".

I'm willing to be overridden here - but I think this is a useful check with minimal performance cost.

pulpcore/tests/unit/models/test_content.py
ggainey: Fine idea, repush inbound!

pulpcore/tests/unit/models/test_content.py
```diff
@@ -72,11 +76,60 @@ def test_read_temp_file(self):
...
+            return_value=set(["sha224"]),
+        ):
+            with self.assertRaises(UnsupportedDigestValidationError) as udv:  # noqa
+                a = Artifact(sha224="asdf")  # noqa
+
+    @mock.patch(
+        "pulpcore.app.models.Artifact.FORBIDDEN_DIGESTS",
+        new_callable=mock.PropertyMock,
+        return_value=set(["md5"]),
+    )
```
mdellweg, line 93: :1st_place_medal: 
I just don't understand, why you are mixing the two technologies here.
I'd make them both decorators.

ggainey: Eh, reasonable point. Was trying to keep the "what we're testing" as close to the test as possible, but I can see it being confusing. Lemme fix.

mdellweg (approved):
Given that you will change the mocking in the tests to a unified appearance, i am going to leave an ACK on this.

CHANGES/7696.feature
```diff
@@ -0,0 +1 @@
+Added pre_save hook to Artifact to enforce checksum rules implied by ALLOWED_CONTENT_CHECKSUMS.
```
bmbouter, line 1: I think this changelog should be for plugin writers since users never call or have access to the save calls directly. So `/CHANGES/plugin_api/7696.feature`.

pulpcore/app/models/content.py
```diff
@@ -58,6 +58,14 @@ class UnsupportedDigestValidationError(Exception):
     pass
 
 
+class MissingDigestValidationError(Exception):
```
bmbouter, line 61: It's not a big thing, but I'm a little uneasy with these custom exceptions declared here and there. Instead can we use Python exceptions and emit them in the right places along with `Raises` docstrings identifying the scenarios they are called?

pulpcore/tests/unit/models/test_content.py
```diff
@@ -72,11 +76,60 @@ def test_read_temp_file(self):
...
 
-class ArtifactAlgorithmTestCase(TestCase):
-    def test_set_forbidden(self):
-        # This will only fire on a Pulp instance that has forbidden md5 in settings.py
-        if "md5" not in Artifact.DIGEST_FIELDS:
-            with self.assertRaises(UnsupportedDigestValidationError) as udv:  # noqa
-                a = Artifact(md5="asdf")  # noqa
-        else:
-            pass
+class ArtifactAlgorithmTestCase(SimpleTestCase):
```
bmbouter, line 79: :+1:

pulpcore/tests/unit/models/test_content.py
```diff
@@ -72,11 +76,60 @@ def test_read_temp_file(self):
...
+    @mock.patch(
+            "pulpcore.app.models.Artifact.FORBIDDEN_DIGESTS",
+            new_callable=mock.PropertyMock,
+            return_value=set(["sha224"]),
+        )
+    def test_set_forbidden(self, mock_DIGEST_FIELDS):  # noqa
```
bmbouter, line 85: Why are there so many `# noqa` ?

pulpcore/app/models/content.py
bmbouter: I don't see the change in this PR, but if the question: "should we also be checking in `Artifact.__init__()`, I think we should not do both. Checking it in two places with each runtime on each system, is not worth it to benefit a programmer at development who will still experience the error just at save time instead. My opinion at least.

pulpcore/app/models/content.py
ggainey: I prefer specific exceptions because it makes code that looks for exceptions a lot cleaner/easier to understand. If you use generic exceptions, code that wants to respond to specific problems (especially code higher up in a call-chain than the immediate caller) has to do string-matches on exception-messages. If we ever end up localizing messages, it makes such code even more difficult.

However, if The Pulp Way is to use generic python exceptions, I'm ok to adhere to that.

pulpcore/app/models/content.py
daviddavis: I also agree. I am hoping that no user ever hits these errors--only developers/plugin writers while they are developing.

pulpcore/tests/unit/models/test_content.py
ggainey: PEP8 doesn't like caps-in-argument-names, and pycharm complains about it. I 'assumed' flake and/or black would as well - investigation shows they don't care, so I can just tell pycharm to not worry about it and remove the #noqas :)

pulpcore/app/models/content.py
daviddavis: I also like having specific exceptions. Would a central place to define exceptions solve your concern @bmbouter?

pulpcore/app/models/content.py
ggainey: okeydoke, will do

CHANGES/7696.feature
ggainey: good catch

pulpcore/app/models/content.py
bmbouter: Yes. And if the others can be moved there to that would be extra great.

pulpcore/plugin/models/__init__.py
```diff
@@ -36,6 +36,7 @@
     Task,
     TaskGroup,
     UnsupportedDigestValidationError,
+    MissingDigestValidationError,
```
bmbouter, line 39: We probably want to move this then to `pulpcore.plugin.exceptions` then.

pulpcore/plugin/models/__init__.py
```diff
@@ -36,6 +36,7 @@
     Task,
     TaskGroup,
     UnsupportedDigestValidationError,
```
bmbouter, line 38: Moving this one will reqiure the 1-cycle deprecation ...

bmbouter (changes requested):
Requested change on where custom exceptions are defined. It's looking great otherwise.

pulpcore/app/models/content.py
ggainey: @bmbouter I'd like to make the consolidation effort its own Thing - see https://pulp.plan.io/issues/7908. That will let this PR/issue remain a .feature, and keep 7908 as strictly a .deprecation. (Obviously, the exception(s) added with this PR I'll move to the right place as part of this PR, repushing shortly)

CHANGES/plugin_api/7696.feature
```diff
@@ -0,0 +1 @@
+Added pre_save hook to Artifact to enforce checksum rules implied by ALLOWED_CONTENT_CHECKSUMS.
```
bmbouter, line 1: Either this PR or the next one I think needs to add a `CHANGES/plugin_api/XXXX.feature` that announces the exceptions available at `pulpcore.plugin.exceptions`. @ggainey is that this PR ok the next one? I'm ok with either.

ggainey: I'm thinking it would make the most sense as part of https://pulp.plan.io/issues/7908 - will add a note to that effect to that issue.

bmbouter (approved):
Thank you! This is an important safety feature.

bmbouter:
I see two approvals!

PR #1029, opened 2020-11-17
Fix example for AccessPolicyFromDB and namespace access policies

mdellweg:
[noissue]


pulpcore/app/migrations/0042_rbac_for_tasks.py
```diff
@@ -46,12 +46,16 @@ def populate_access_policy(apps, schema_editor):
...
+
 class Migration(migrations.Migration):
 
     dependencies = [
         ('core', '0041_accesspolicy'),
     ]
 
     operations = [
-        migrations.RunPython(populate_access_policy),
+        migrations.RunPython(populate_access_policy, reverse_code=remove_access_policy),
```
bmbouter, line 60: Generally we never modify a migration once it's shipped.

pulpcore/app/migrations/0050_namespace_access_policies.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0049_add_file_field_to_uploadchunk'),
+    ]
+
+    operations = [
+        migrations.RunPython(namespace_access_policies_up, reverse_code=namespace_access_policies_down),
```
bmbouter, line 27: fancy :+1:

CHANGES/7845.bugfix
```diff
@@ -0,0 +1 @@
+Properly namespaced the `viewset_name` in `AccessPolicy` to avoid naming conflicts in plugins.
```
bmbouter, line 1: I think this should be only for plugin writers. Users never modify/populate this data since this field is part of the contract between plugin writers and pulpcore.plugin. It's tech preview and generally unused in plugin code so I don't think it needs to go through the deprecation cycle, but I do think it should be labeled in the plugin writer changelog as a .removal.

pulpcore/app/access_policy.py
```diff
@@ -29,5 +29,10 @@ def get_policy_statements(self, request, view):
         Returns:
             The access policy statements in drf-access-policy policy structure.
         """
-        access_policy_obj = AccessPolicyModel.objects.get(viewset_name=view.__class__.__name__)
+        try:
+            access_policy_obj = AccessPolicyModel.objects.get(
+                viewset_name=view.__class__.view_name()
+            )
+        except AccessPolicyModel.NotFound:
```
bmbouter, line 36: How long will we carry this fallback style behavior? This seems like it could go through the deprecation cycle with a fallback like this... If so it needs warnings that it'll be removed in 3.10 and a deprecation changelog entry.

pulpcore/app/access_policy.py
```diff
@@ -29,5 +29,10 @@ def get_policy_statements(self, request, view):
         Returns:
             The access policy statements in drf-access-policy policy structure.
         """
-        access_policy_obj = AccessPolicyModel.objects.get(viewset_name=view.__class__.__name__)
+        try:
+            access_policy_obj = AccessPolicyModel.objects.get(
+                viewset_name=view.__class__.view_name()
```
bmbouter, line 34: oh yeahhhh :cake:

bmbouter (changes requested):
This is looking really good. I requested some changes, but all mostly small things I think.

CHANGES/7845.bugfix
mdellweg: But users see it and need that name to look the policy up they want to modify.

bmbouter: Oh that's true. Actually that got me thinking ... from a user usability perspective `remotes-file/file` is probably just as mysterious a string as `FileRemoteViewSet`. Random idea: what if instead of `viewset_name` it was `viewset_url` and it actually contained a url? Maybe that won't work well at all, just a thought.

pulpcore/app/migrations/0042_rbac_for_tasks.py
mdellweg: :disappointed: I'll remove it.

CHANGES/7845.bugfix
mdellweg: Would that be unique? I thought you can hook up the viewset with multiple urls. (And if it's possible, someone will do it...)

mdellweg: I only see a `urlpattern()` method leading to
```
In [17]: FileRepositoryViewSet.urlpattern()
Out[17]: 'repositories/file/file'

In [18]: TaskViewSet.urlpattern()
Out[18]: 'tasks'
```
That might be ok. Going to check, where it is defined, and if it's guaranteed to exist.

mdellweg: Turns out, both live on the `NamedModelViewSet`. With the `view_name` being a tiny little bit more collision free. But things go up in flames anyway if the `urlpattern` isn't unique. So they should be equaly fine with the url to be more user friendly.

pulpcore/app/access_policy.py
bmbouter: This looks good.

pulpcore/app/models/task.py
```diff
@@ -352,7 +352,7 @@ class Task(BaseModel, AutoDeleteObjPermsMixin, AutoAddObjPermsMixin):
     # as it has a separate job ID that is not the task ID.
     _resource_job_id = models.UUIDField()
 
-    ACCESS_POLICY_VIEWSET_NAME = "TaskViewSet"
+    ACCESS_POLICY_VIEWSET_NAME = "tasks"
```
bmbouter, line 355: So this is just the url representing `/pulp/api/v3/tasks/` ?

mdellweg: Yes, it is.

bmbouter (approved):
This all looks good to me. Thank you!

ipanova (approved):

PR #1031, opened 2020-11-20
Add progress reporting for imports

daviddavis:
fixes #6559

### DIFF
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -10,6 +10,7 @@ from logging import getLogger
 
 from django.conf import settings
 from django.core.files.storage import default_storage
+from django.db.models import F
 
 from pkg_resources import DistributionNotFound, get_distribution
 from rest_framework.serializers import ValidationError
@@ -20,6 +21,8 @@ from pulpcore.app.models import (
     Artifact,
     Content,
     CreatedResource,
+    GroupProgressReport,
+    ProgressReport,
     PulpImport,
     PulpImporter,
     Repository,
@@ -30,6 +33,7 @@ from pulpcore.app.modelresource import (
     ArtifactResource,
     ContentArtifactResource,
 )
+from pulpcore.constants import TASK_STATES
 from pulpcore.tasking.tasks import enqueue_with_reservation
 
 log = getLogger(__name__)
@@ -110,6 +114,13 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
     dest_repo = Repository.objects.get(pk=destination_repo_pk)
     importer = PulpImporter.objects.get(pk=importer_pk)
 
+    pb = ProgressReport(
+        message=f"Importing content for {dest_repo.name}",
+        code="import.repo.version.content",
+        state=TASK_STATES.RUNNING,
+    )
+    pb.save()
+
     with tempfile.TemporaryDirectory() as temp_dir:
         # Extract the repo file for the repo info
         with tarfile.open(tar_path, "r:gz") as tar:
@@ -178,6 +189,15 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
             with dest_repo.new_version() as new_version:
                 new_version.set_content(content)
 
+        content_count = content.count()
+        pb.total = content_count
+        pb.done = content_count
+        pb.state = TASK_STATES.COMPLETED
+        pb.save()
+
+    gpr = TaskGroup.current().group_progress_reports.filter(code="import.repo.versions")
+    gpr.update(done=F("done") + 1)
+
 
 def pulp_import(importer_pk, path, toc):
     """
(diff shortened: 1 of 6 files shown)


pulpcore/app/tasks/importer.py
```diff
@@ -178,6 +182,20 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
             with dest_repo.new_version() as new_version:
                 new_version.set_content(content)
 
+        content_count = content.count()
+        data = dict(
+            message=f"Importing content for {dest_repo.name}",
+            code="import-repo-version-content",
```
goosemania, line 188: I think we have an unspoken and unwritten convention :) to use dots in the `code` as a word separator. I don't have any preference, just for the sake of consistency.

pulpcore/app/tasks/importer.py
```diff
@@ -337,18 +358,27 @@ def validate_and_assemble(toc_filename):
...
-            src = os.path.join(temp_dir, base_path)
-            dest = os.path.join(settings.MEDIA_ROOT, base_path)
-
-            if not default_storage.exists(dest):
-                with open(src, "rb") as f:
-                    default_storage.save(dest, f)
+        data = dict(message="Importing Artifacts", code="import-artifacts",
+                    total=len(ar_result.rows))
+        with ProgressReport(**data) as pb:
+            for row in pb.iter(ar_result.rows):
```
goosemania, line 364: It's user friendly to show progress incrementing progress report counters for each artifact however it might affect performance I think. 
The whole operation is not fast anyway , so maybe it's not a concern

pulpcore/app/tasks/importer.py
```diff
@@ -337,18 +358,27 @@ def validate_and_assemble(toc_filename):
...
+                if not default_storage.exists(dest):
+                    with open(src, "rb") as f:
+                        default_storage.save(dest, f)
 
         with open(os.path.join(temp_dir, REPO_FILE), "r") as repo_data_file:
             data = json.load(repo_data_file)
+            gpr = GroupProgressReport(
+                message="Importing repository versions",
+                code="import-repo-versions",
+                total=len(data),
```
goosemania, line 379: Maybe we should add `done` here as well. What do you think?
The default is 0 https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/progress.py#L276

pulpcore/app/tasks/importer.py
daviddavis: I thought that's what this code is doing?

goosemania: It is. Just shared that performance might be a potential concern here.

ggainey: At this step we're physically moving Artifact binaries to their final resting-places on the filesystem; I expect the I/O time will completely swamp the time to update the progress report. SSDs are fast, but it's a LOT of bytes :)

pulpcore/app/tasks/importer.py
ggainey: Well, darn, I did not realize this - pulpcore/app/importexport.py uses dashes as well :(

ggainey:
After running CentOS8-base through the export/import dance, I have a couple of observations from the end-user's POV. 

With a large export-file, it can take an appreciable time (ones of minutes) before we get any progress reporting in the parent task or task-group. Similarly, once we dispatch the child-tasks-per-repo, there's a lot of startup going on before we have anything in the task-progress-reports. I'm not sure that this kind of reporting fits into the progress-report framework. As it stands, to really see what's going on, the user wants to be watching the systemctl logs in addition to the progress-reports.

Maybe we just need to document this expectation somewhere?

pulpcore/app/tasks/importer.py
dralley: We really need to look into async file I/O

dralley: I guess technically threads are fine for this purpose as well.  If it becomes a thing that people complain about being slow anyways.

pulpcore/app/tasks/importer.py
daviddavis: yea, that's where I got it from.

daviddavis:
I think I addressed all the feedback. 

I also added some additional progress reporting and updated some of the existing progress reporting. Namely, I added progress reports to validating and recombining the chunks so progress reporting starts instantly when the task starts. Also, I create the import version progress report at the beginning of the child task code instead of at the end so there is a progress report immediately available. 

It's not perfect but I think that should solve the gaps between when the tasks start and when progress reporting starts.

goosemania:
Binding tests fail because fedorapeople was unavailable and they reference old fixture repos:
```
$ git grep repos.fedora
.ci/assets/bindings/test_bindings.py:        "https://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/test_bandwidth_repo/"
.ci/assets/bindings/test_bindings.py:remote_url = "https://repos.fedorapeople.org/pulp/pulp/demo_repos/test_file_repo/PULP_MANIFEST"
.ci/assets/bindings/test_bindings.rb:remote_url = 'https://repos.fedorapeople.org/pulp/pulp/demo_repos/test_file_repo/PULP_MANIFEST'
```
Do we need move those to the fixtures.pulpproject.org or they didn't migrate for a reason?

goosemania (approved):
Thank you, @daviddavis !

daviddavis:
fixtures.pulpproject.org is a deployment of the pulp-fixtures repo (formerly at https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/) and the demo repos were created outside of pulp-fixtures. We could move them into pulp-fixtures or onto the pulpproject.org server but I haven't done so because of how infrequently we use them in Pulp 3.

ggainey:
> It's not perfect but I think that should solve the gaps between when the tasks start and when progress reporting starts.

Doing some testing right now, and def concur - you've answered my user-experience issues completely.

ggainey (approved):

dkliban (approved):

PR #1032, opened 2020-11-20
Add Github Actions CI

dralley:
[noissue]


dralley:
That's kinda weird, why did this one trigger immediately when the others only trigger after being merged?

daviddavis (approved):

mdellweg (approved):
Should we merge this one and fix the docs workflow later?

functest_requirements.txt
```diff
@@ -1,6 +1,7 @@
 django
 dynaconf
 pulp-smash>=1!0.12.0
+pulp_file  # import/export functional tests import from the pulp_file functional tests
```
fao89, line 4: I believe this is bringing the outdated version of drf-spectacular

fao89: it installs pulp-file 1.3.0 which requires pulpcore<3.9,>=3.7, which uninstalls source pulpcore and gets 3.9 from PyPI

mdellweg: Just found the line
https://github.com/pulp/pulpcore/pull/1032/checks?check_run_id=1442157564#step:7:263

mdellweg: Do we need functest_requirements in the docs workflow?

.github/workflows/scripts/before_install.sh
```diff
@@ -0,0 +1,110 @@
...
+
+
+git clone --depth=1 https://github.com/pulp/pulp_file.git --branch master
+if [ -n "$PULP_FILE_PR_NUMBER" ]; then
+  cd pulp_file
+  git fetch --depth=1 origin pull/$PULP_FILE_PR_NUMBER/head:$PULP_FILE_PR_NUMBER
+  git checkout $PULP_FILE_PR_NUMBER
+  cd ..
+fi
+
```
fao89, line 92: maybe we need to add here:
```python
pip install ./pulp_file
```

functest_requirements.txt
mdellweg: Also we should try to not install pulpcore and pulp_file even for the functional tests. They should really be independent (I'm just not sure how far we are with that effort.).

functest_requirements.txt
```diff
@@ -1,6 +1,7 @@
 django
 dynaconf
 pulp-smash>=1!0.12.0
+git+https://github.com/pulp/pulp_file.git#egg=pulp-file  # import/export functional tests import from the pulp_file functional tests
```
mdellweg, line 4: That file is used as test_requirements in setup.py
And this format seems to break stuff.

pulpcore/tests/functional/api/using_plugin/utils.py
```diff
@@ -81,3 +89,23 @@ def create_file_publication(cfg, repo, version_href=None):
     else:
         body = {"repository": repo["pulp_href"]}
     return api.Client(cfg).post(FILE_PUBLICATION_PATH, body)
+
+
+def monitor_task(task_href):
```
mdellweg, line 94: monitor_task should be taken from pulp_smash.

fao89: @dralley monitor_task from pulp_smash requires rewriting some tests, example: https://github.com/pulp/pulp_rpm/pull/1801/files#r528923043

daviddavis (approved):

pulpcore/tests/functional/api/using_plugin/utils.py
dralley: I don't think they do in this limited case, as none of these tasks are expected to ever fail.

PR #1033, opened 2020-11-23
Updated pulplift references in the docs

mdellweg:
fixes #7878
https://pulp.plan.io/issues/7878

This depends on https://github.com/pulp/pulp_installer/pull/498

### DIFF
--- a/docs/contributing/dev-setup.rst
+++ b/docs/contributing/dev-setup.rst
@@ -3,8 +3,9 @@
 Developer Setup
 ===============
 
-To ease developer setup, we have `Pulplift <https://github.com/pulp/pulplift>`_ which is based on
-the `Forklift <https://github.com/theforeman/forklift>`_ project and utilizes
+To ease developer setup, we have `Pulplift
+<https://pulp-installer.readthedocs.io/en/latest/pulplift/>`_ as part of the installer repository,
+which is based on the `Forklift <https://github.com/theforeman/forklift>`_ project and utilizes
 `Ansible <https://docs.ansible.com/ansible/index.html>`_ roles and playbooks to provide supported
 `Vagrant <https://docs.vagrantup.com/>`_ boxes that are more consistent with the user experience.
 
@@ -19,9 +20,6 @@ has read and write permissions, it doesn't matter where your **development direc
 You will need ``pulp/pulpcore`` at a minimum::
 
     $ git clone https://github.com/pulp/pulpcore.git
-
-This repository is for Pulp 3 only, development is done agains the master branch of
-each.
 
 Additionally, you will need at least one plugin.::
 
@@ -39,7 +37,7 @@ Installation
 ------------
 
 We recommend using ``pulplift`` for developer installations. Follow the instructions in the
-`README.md <https://github.com/pulp/pulplift/#pulplift>`_.
+`installer documentation <https://pulp-installer.readthedocs.io/en/latest/pulplift/>`_.
 
 It is also possible to use the `Ansible roles
 <https://github.com/pulp/pulp_installer#pulp-3-ansible-installer>`_ directly, if you prefer not to
--- a/CHANGES/7878.doc
+++ b/CHANGES/7878.doc
@@ -1,0 +1,1 @@
+Updated the reference to the new location of pulplift at the installer repository in the development section.


docs/contributing/dev-setup.rst
```diff
@@ -20,8 +21,7 @@ You will need ``pulp/pulpcore`` at a minimum::
 
     $ git clone https://github.com/pulp/pulpcore.git
 
-This repository is for Pulp 3 only, development is done agains the master branch of
-each.
+This repository is for Pulp 3 only, development is done against the master branch of each.
```
dralley, line 24: "development is done against the master branch of each." is really confusing in this context as it's not clear what "each" is referring to.  It would be better if we either mentioned this after we mention the plugins, or if we just leave it out.

docs/contributing/dev-setup.rst
```diff
@@ -3,7 +3,8 @@
 Developer Setup
 ===============
 
-To ease developer setup, we have `Pulplift <https://github.com/pulp/pulplift>`_ which is based on
+To ease developer setup, we have `Pulplift <https://pulp-installer.readthedocs.io/en/latest/pulplift/>`_
+as part of the installer repository, which is based on
```
bmbouter, line 7: nitpick: Can this paragraph be wrapped more. I know it makes the blame more complicated, but this way is less readable in unrendered.

docs/contributing/dev-setup.rst
bmbouter: +1 to remove this sentence altogether.

docs/contributing/dev-setup.rst
```diff
@@ -38,8 +38,8 @@ The current base branch on this repository is the master branch.
 Installation
 ------------
 
-We recommend using ``pulplift`` for developer installations. Follow the instructions in the
-`README.md <https://github.com/pulp/pulplift/#pulplift>`_.
+We recommend using ``pulplift`` for developer installations.
```
bmbouter, line 41: Nitpick here also to not wrap at the end of the sentence. Dense paragraphs are more readable I think.

bmbouter (changes requested):
requested some nitpick changes. Thanks in advance for considering them.

docs/contributing/dev-setup.rst
mdellweg: Not sure how to properly wrap those.

But it's not only about blame. It's every single future review on these lines that touches more lines than needed. And those unwieldy hyper references prove to me that this is actually source code to be compiled into nice looking html, pdf or whatever.

docs/contributing/dev-setup.rst
```diff
@@ -39,8 +41,8 @@ Installation
...
 
 We recommend using ``pulplift`` for developer installations. Follow the instructions in the
-`README.md <https://github.com/pulp/pulplift/#pulplift>`_.
+`installer documentation <https://pulp-installer.readthedocs.io/en/latest/pulplift/>`_.
 
-It is also possible to use the `Ansible roles
-<https://github.com/pulp/pulp_installer#pulp-3-ansible-installer>`_ directly, if you prefer not to
-use Vagrant.
+It is also possible to use the
+`Ansible roles <https://github.com/pulp/pulp_installer#pulp-3-ansible-installer>`_ directly, if you
```
bmbouter, line 47: Sphinx renders links correctly if they are split with a newline.

docs/contributing/dev-setup.rst
```diff
@@ -3,10 +3,15 @@
 Developer Setup
 ===============
 
-To ease developer setup, we have `Pulplift <https://github.com/pulp/pulplift>`_ which is based on
-the `Forklift <https://github.com/theforeman/forklift>`_ project and utilizes
-`Ansible <https://docs.ansible.com/ansible/index.html>`_ roles and playbooks to provide supported
-`Vagrant <https://docs.vagrantup.com/>`_ boxes that are more consistent with the user experience.
+To ease developer setup, we have
```
bmbouter, line 6: This paragraphs is much less readable.

bmbouter (approved):
This is looking good to me.

dkliban (approved):

bmbouter:
I see two lgtm's, I'm merging.

PR #1035, opened 2020-11-23
Add asciinema terminal params

melcorr:
* Update record-a-demo file

[noissue]

We discussed this in a team meeting on Nov 16th and agreed it would be good to restrict the terminal size so we don't have embedded videos of various shapes and sizes on the demo page of the website. 

Should I also add a line to tag me, or send the video to me, so I can add it to the site?


mdellweg (approved):
:+1:

daviddavis (approved):

daviddavis:
> Should I also add a line to tag me, or send the video to me, so I can add it to the site?

Actually yea, that would be good.

melcorr:
Done, thanks!

PR #1036, opened 2020-11-24
Add access policies with post_migrate signal

mdellweg:
fixes #7710
https://pulp.plan.io/issues/7710

@bmbouter Time for a first look at my design of this.

### DIFF
--- a/docs/plugins/plugin-writer/concepts/rbac/access_policy.rst
+++ b/docs/plugins/plugin-writer/concepts/rbac/access_policy.rst
@@ -130,63 +130,64 @@ AccessPolicy instance to exist.
 Shipping a Default Access Policy
 --------------------------------
 
-To ship an access policy, write a data migration that creates an ``AccessPolicy`` instance. Here's
-an example of code to create an instance, which would be contained in a data migration.
+To ship a default access policy, define a dictionary named ``DEFAULT_ACCESS_POLICY`` as a class
+attribute on a subclass of ``NamedModelViewSet`` containing both ``statements`` and
+``permissions_assignment``. The ``AccessPolicy`` instance will be then be created in the
+``pulp_migrate`` signal handler.
+
+Here's an example of code to define a default policy:
 
 .. code-block:: python
 
-    from pulpcore.plugin.models import AccessPolicy
+    class FileRemoteViewSet(RemoteViewSet):
 
-    FILE_REMOTE_STATEMENTS = [
-        {
-            "action": ["list"],
-            "principal": "authenticated",
-            "effect": "allow",
-        },
-        {
-            "action": ["create"],
-            "principal": "authenticated",
-            "effect": "allow",
-            "condition": "has_model_perms:file.add_fileremote",
-        },
-        {
-            "action": ["retrieve"],
-            "principal": "authenticated",
-            "effect": "allow",
-            "condition": "has_model_or_obj_perms:file.view_fileremote",
-        },
-        {
-            "action": ["update", "partial_update"],
-            "principal": "authenticated",
-            "effect": "allow",
-            "condition": "has_model_or_obj_perms:file.change_fileremote",
-        },
-        {
-            "action": ["destroy"],
-            "principal": "authenticated",
-            "effect": "allow",
-            "condition": "has_model_or_obj_perms:file.delete_fileremote",
-        },
-    ]
+    <...>
+        DEFAULT_ACCESS_POLICY = {
+            "statements": [
+                {
+                    "action": ["list"],
+                    "principal": "authenticated",
+                    "effect": "allow",
+                },
+                {
+                    "action": ["create"],
+                    "principal": "authenticated",
+                    "effect": "allow",
+                    "condition": "has_model_perms:file.add_fileremote",
+                },
+                {
+                    "action": ["retrieve"],
+                    "principal": "authenticated",
+                    "effect": "allow",
+                    "condition": "has_model_or_obj_perms:file.view_fileremote",
+                },
+                {
+                    "action": ["update", "partial_update"],
+                    "principal": "authenticated",
+                    "effect": "allow",
+                    "condition": "has_model_or_obj_perms:file.change_fileremote",
+                },
+                {
+                    "action": ["destroy"],
+                    "principal": "authenticated",
+                    "effect": "allow",
+                    "condition": "has_model_or_obj_perms:file.delete_fileremote",
+                },
+            ],
 
-    FILE_REMOTE_PERMISSIONS_ASSIGNMENT = [
-        {
-            "function": "add_for_object_creator",
-            "parameters": None,
-            "permissions": [
-                "file.view_fileremote", "file.change_fileremote", "file.delete_fileremote"
-            ]
+            "permissions_assignment": [
+                {
+                    "function": "add_for_object_creator",
+                    "parameters": None,
+                    "permissions": [
+                        "file.view_fileremote", "file.change_fileremote", "file.delete_fileremote"
+                    ]
+                },
+            ],
         }
-    ]
+        <...>
 
-    AccessPolicy.objects.create(
-        viewset_name="remotes/file/file",
-        statements=FILE_REMOTE_STATEMENTS,
-        permissions_assignment=FILE_REMOTE_PERMISSIONS_ASSIGNMENT
-    )
-
-The actual ``AccessPolicy`` statement is created at the end. The other data structures store the
-default policy. For en explanation of the ``permissions_assignment`` see the
+For an explanation of the ``permissions_assignment`` see the
 :ref:`shipping_a_default_new_object_policy` documentation.
 
 
--- a/pulpcore/app/apps.py
+++ b/pulpcore/app/apps.py
@@ -2,6 +2,7 @@ from collections import defaultdict
 from importlib import import_module
 
 from django import apps
+from django.db.models.signals import post_migrate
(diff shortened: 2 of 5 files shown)


bmbouter:
Overall I think this is strongly headed in the right direction. While running the code I ran into this problem with this PR and a freshly migrated to latest (e.g. `pclean`) system. Then I run

```
(pulp) [vagrant@pulp3-source-fedora32 pulp_ansible]$ pulpcore-manager flush
You have requested a flush of the database.
This will IRREVERSIBLY DESTROY all data currently in the 'pulp' database,
and return each table to an empty state.
Are you sure you want to do this?

    Type 'yes' to continue, or 'no' to cancel: yes
  File "/usr/local/lib/pulp/lib64/python3.8/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/pulp/lib64/python3.8/site-packages/django/core/management/__init__.py", line 375, in execute
  File "/usr/local/lib/pulp/lib64/python3.8/site-packages/django/core/management/base.py", line 323, in run_from_argv
  File "/usr/local/lib/pulp/lib64/python3.8/site-packages/django/core/management/base.py", line 364, in execute
  File "/usr/local/lib/pulp/lib64/python3.8/site-packages/django/core/management/commands/flush.py", line 80, in handle
    emit_post_migrate_signal(verbosity, interactive, database)
  File "/usr/local/lib/pulp/lib64/python3.8/site-packages/django/core/management/sql.py", line 45, in emit_post_migrate_signal
    models.signals.post_migrate.send(
  File "/usr/local/lib/pulp/lib64/python3.8/site-packages/django/dispatch/dispatcher.py", line 173, in send
    return [
  File "/usr/local/lib/pulp/lib64/python3.8/site-packages/django/dispatch/dispatcher.py", line 174, in <listcomp>
    (receiver, receiver(signal=self, sender=sender, **named))
TypeError: _populate_access_policies() missing 1 required positional argument: 'apps'
```

mdellweg:
That strikes me as a rather unfortunate turn of events. The `apps` arg is documented to be passed unconditionally. ;(
https://docs.djangoproject.com/en/2.2/ref/signals/

mdellweg:
This will stay a draft until #1029 landed.

mdellweg:
Also i need to adjust the docs after #1029 merge.

bmbouter:
I've confirmed running the `pulpcore-manager flush` causes the AccessPolicy to still be present.

```
(pulp) [vagrant@pulp3-source-fedora32 ~]$ pulpcore-manager shell -c 'from pulpcore.app.models.access_policy import AccessPolicy;print(AccessPolicy.objects.all().count())'
1
(pulp) [vagrant@pulp3-source-fedora32 ~]$ pulpcore-manager flush
You have requested a flush of the database.
This will IRREVERSIBLY DESTROY all data currently in the 'pulp' database,
and return each table to an empty state.
Are you sure you want to do this?

    Type 'yes' to continue, or 'no' to cancel: yes
Initialize missing access policies for core.
Initialize missing access policies for ansible.
Initialize missing access policies for file.
(pulp) [vagrant@pulp3-source-fedora32 ~]$ pulpcore-manager shell -c 'from pulpcore.app.models.access_policy import AccessPolicy;print(AccessPolicy.objects.all().count())'
1
```

bmbouter:
I've confirmed this correctly does not overwrite a custom modified policy which is important too.

```
(pulp) [vagrant@pulp3-source-fedora32 ~]$ pulpcore-manager shell -c 'from pulpcore.app.models.access_policy import AccessPolicy;print(AccessPolicy.objects.first().statements)'
[{'action': ['list'], 'effect': 'allow', 'principal': 'authenticated'}, {'action': ['retrieve'], 'effect': 'allow', 'condition': 'has_model_or_obj_perms:core.view_task', 'principal': 'authenticated'}, {'action': ['destroy'], 'effect': 'allow', 'condition': 'has_model_or_obj_perms:core.delete_task', 'principal': 'authenticated'}, {'action': ['update', 'partial_update'], 'effect': 'allow', 'condition': 'has_model_or_obj_perms:core.change_task', 'principal': 'authenticated'}]
(pulp) [vagrant@pulp3-source-fedora32 ~]$ pulpcore-manager shell -c "from pulpcore.app.models.access_policy import AccessPolicy;foo = AccessPolicy.objects.first();foo.statements = {};foo.save()"
(pulp) [vagrant@pulp3-source-fedora32 ~]$ pulpcore-manager shell -c 'from pulpcore.app.models.access_policy import AccessPolicy;print(AccessPolicy.objects.first().statements)'
{}
(pulp) [vagrant@pulp3-source-fedora32 ~]$ pulpcore-manager migrate core
Operations to perform:
  Apply all migrations: core
Running migrations:
  No migrations to apply.
Initialize missing access policies for core.
Initialize missing access policies for ansible.
Initialize missing access policies for file.
(pulp) [vagrant@pulp3-source-fedora32 ~]$ pulpcore-manager shell -c 'from pulpcore.app.models.access_policy import AccessPolicy;print(AccessPolicy.objects.first().statements)'
{}
```

bmbouter (approved):
This looks great to me, and it tested well too. :+1: :tada:

docs/plugins/plugin-writer/concepts/rbac/access_policy.rst
```diff
@@ -130,63 +130,64 @@ AccessPolicy instance to exist.
 Shipping a Default Access Policy
 --------------------------------
 
-To ship an access policy, write a data migration that creates an ``AccessPolicy`` instance. Here's
-an example of code to create an instance, which would be contained in a data migration.
+To ship a default access policy, define a dictionary named ``DEFAULT_ACCESS_POLICY`` as a class
```
goosemania, line 133: What happens if I changed my mind and I want new defaults?
If the changes are additive, is it fine just to change it?
Maybe it's worth documenting the implication of changing the default policies and how other policies rely on their existence and so on. What do you think?

pulpcore/app/apps.py
```diff
@@ -82,6 +83,9 @@ def ready(self):
         self.import_serializers()
         self.import_urls()
         self.import_modelresources()
+        post_migrate.connect(
+            _populate_access_policies, sender=self, dispatch_uid="my_unique_identifier"
```
goosemania, line 87: Where is the `dispatch_uid` used? Would we benefit from having more meaningful id  or not really?

pulpcore/app/apps.py
```diff
@@ -164,3 +169,18 @@ class PulpAppConfig(PulpPluginAppConfig):
     # with manage.py, etc. This cannot contain a dot and must not conflict with the name of a
     # package containing a Django app.
     label = "core"
+
+
+def _populate_access_policies(sender, **kwargs):
+    print(f"Initialize missing access policies for {sender.label}.")
+    apps = kwargs.get("apps")
+    if apps is None:
+        from django.apps import apps
```
goosemania, line 178: What is the case when the `apps` is `None`?
[Docs](https://docs.djangoproject.com/en/2.2/ref/signals/#post-migrate) say that we should not use global `apps` registry if I read it correctly.

docs/plugins/plugin-writer/concepts/rbac/access_policy.rst
bmbouter: I agree users need to understand that they can change behaviors, and also when/why/how etc. For those user facing items, we don't have docs yet, all we have are these plugin writer docs. We should document that a default is only applied once, for example what if a plugin writer needs to adjust the default policy... These issues were already here though so I think they can be handled separately. If one of us should push a commit to his PR I'm ok with that also.

pulpcore/app/apps.py
bmbouter: I don't think it's used since in this case the receiver of the signal isn't interested in where the it was dispatched from.

bmbouter: I didn't expect it to be a required kwarg tho....

pulpcore/app/apps.py
bmbouter: In the early-on testing of this PR the docs say apps is always set but in practice it was not. Maybe that's a Django bug perhaps?

We do something similar here https://github.com/pulp/pulpcore/blob/master/pulpcore/app/apps.py#L15-L24 which is how Pulp handles "app discovery" and inspection at startup time.

This brings up another question: should we restrict to apps that are of type `PulpPluginAppConfig`. If so maybe we just switch to using the yielded items from `pulp_plugin_configs` https://github.com/pulp/pulpcore/blob/master/pulpcore/app/apps.py#L24 or maybe the viewsets loaded in pulpcore.apps ?

goosemania: Thanks for the background.
We can be more precise that we look only at pulp apps, however in this case I guess it's quite safe to load the global one, because we are accessing one specific model and not doing any discovery or iteration.

docs/plugins/plugin-writer/concepts/rbac/access_policy.rst
goosemania: Yeah, I meant to make it clear from a plugin writer point of view that changing defaults might not be an desirable thing.
A follow-up PR with any clarification sounds good. I didn't mean to block this PR with my comment.

pulpcore/app/apps.py
goosemania: This can also go in a follow-up PR if we want to merge this PR as is.

goosemania (approved):
Thanks!

PR #1040, opened 2020-11-30
Replace custom task cancellation with RQ 1.7 functionality

dralley:
closes: #7902
https://pulp.plan.io/issues/7902

### DIFF
--- a/pulpcore/tasking/util.py
+++ b/pulpcore/tasking/util.py
@@ -4,6 +4,8 @@ from gettext import gettext as _
 
 from django.db import transaction
 from django.urls import reverse
+from rq.command import send_stop_job_command
+from rq.exceptions import InvalidJobOperation, NoSuchJobError
 from rq.job import Job, get_current_job
 from rq.worker import Worker
 
@@ -12,7 +14,6 @@ from pulpcore.app.util import get_view_name_for_model
 from pulpcore.constants import TASK_FINAL_STATES, TASK_STATES
 from pulpcore.exceptions import MissingResource
 from pulpcore.tasking import connection
-from pulpcore.tasking.constants import TASKING_CONSTANTS
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


dralley:
RQ feature: https://github.com/rq/rq/pull/1376

daviddavis (approved):
This is awesome. Good find @dralley.

pulpcore/tasking/util.py
```diff
@@ -50,18 +50,16 @@ def cancel(task_id):
...
+        send_stop_job_command(redis_conn, resource_job.get_id())
+    except (InvalidJobOperation, NoSuchJobError):
+        # We don't care if the job isn't currently running when we try to cancel
+        pass
 
     resource_job.delete()
     job.delete()
 
-    # A hack to ensure that we aren't deleting resources still being used by the workhorse
-    time.sleep(1.5)
```
dralley, line 63: Maybe we should keep this but reduce it to ~0.1 seconds?

bmbouter: Yeah I'm not entirely sure what to do. I'm trying to think through it; so the resources delete while the task is in use, and it receives a database write error in the task at some point?

bmbouter: Maybe that's ok?

dralley: There's an inherent delay in between when the cancellation is triggered and when it actually happens.  Before that delay was due to polling, now it's message passing so it should be faster but technically still present.

dralley: The main thing is that we should avoid cancellations being marked as failures.

dralley: @bmbouter I'm thinking I should just err towards caution and add it back for now.

pulpcore/tasking/util.py
```diff
@@ -50,18 +51,19 @@ def cancel(task_id):
...
+        send_stop_job_command(redis_conn, job.get_id())
+        send_stop_job_command(redis_conn, resource_job.get_id())
+    except (InvalidJobOperation, NoSuchJobError):
+        # We don't care if the job isn't currently running when we try to cancel
+        pass
 
-    if resource_job.is_started:
-        redis_conn.sadd(TASKING_CONSTANTS.KILL_KEY, resource_job.get_id())
+    # A hack to ensure that we aren't deleting resources still being used by the workhorse
+    time.sleep(0.5)
```
bmbouter, line 62: :+1:

bmbouter (approved):
Thank you for switching us back to the upstream feature.

PR #1041, opened 2020-11-30
Add headers option is aiohttp client session

ThikaXer:
This PR is needed to add support for synchronising oracle repos using ULN (PR #1896 in pulp_rpm)

https://github.com/pulp/pulp_rpm/pull/1896


dralley:
@ThikaXer Could you rebase the PR?  Looks like there is a conflicting file.

bmbouter:
@ThikaXer Sorry for the long delay in responding to this. Thank you for opening this PR. I've realized that in addition to the needs of oracle linux sync, there is another use case that motivates having users be able to configure headers for downloaders. For that reason I think we should add them to the base `Remote` object so every `Remote` subclass (including RPM's) will have them.

I filed this story to captures that: https://pulp.plan.io/issues/8083

Your PR mostly does it, except can you add the new field and then also have the DownloaderFactory set it from the Remote object instaed of a kwarg?

This is my suggestion on how to move this forward. Any questions or concerns are welcome. I think our goals are aligned, but let me know if I'm misreading your goal.

Thank you!

bmbouter (changes requested):
Requesting addition attribute on `Remote` object.

ThikaXer:
> @ThikaXer Sorry for the long delay in responding to this. Thank you for opening this PR. I've realized that in addition to the needs of oracle linux sync, there is another use case that motivates having users be able to configure headers for downloaders. For that reason I think we should add them to the base `Remote` object so every `Remote` subclass (including RPM's) will have them.
> 
> I filed this story to captures that: https://pulp.plan.io/issues/8083
> 
> Your PR mostly does it, except can you add the new field and then also have the DownloaderFactory set it from the Remote object instaed of a kwarg?
> 
> This is my suggestion on how to move this forward. Any questions or concerns are welcome. I think our goals are aligned, but let me know if I'm misreading your goal.
> 
> Thank you!

@bmbouter Thank you for your review. I'm now trying to integrate your suggestions. However, I'm having difficulties:
In the requirements of pulpcore, it says `Django~=2.2.17  # LTS version, switch only if we have a compelling reason to`, but the Django JSONField is only available from version 3.1, see https://docs.djangoproject.com/en/3.1/ref/models/fields/#field-types .

Would it be an alternative to set the headers field eg as a text field? This would loose the dict structure, though. Or would this be a compelling reason to switch the django version? Would that influence the rest of the code?
Thank you for your help!

daviddavis:
JSONField was introduced in django 3.1 but prior to 3.1, it is available as a postgresql only datatype from django.contrib. This should work:

```python
from django.contrib.postgres.fields import JSONField
```

ThikaXer:
Thank you for your help. Is this implementation what you meant? Thanks!

quba42:
@ThikaXer Since @bmbouter opened an issue (https://pulp.plan.io/issues/8083), it is probably worth adjusting your commit message to replace `[noissue]` with:
```
fixes #8083
https://pulp.plan.io/issues/8083
```

pulpcore/download/http.py
```diff
@@ -134,6 +135,7 @@ def __init__(
                 as its argument. The callback will be called when the response headers are
                 available. The dictionary passed has the header names as the keys and header values
                 as its values. e.g. `{'Transfer-Encoding': 'chunked'}`
+            headers (dict): Headers for a websitecall
```
bmbouter, line 138:
```suggestion
            headers (dict): Headers to be submitted with the request.
```

pulpcore/download/http.py
```diff
@@ -143,7 +145,7 @@ def __init__(
         else:
             timeout = aiohttp.ClientTimeout(total=None, sock_connect=600, sock_read=600)
             conn = aiohttp.TCPConnector({"force_close": True})
-            self.session = aiohttp.ClientSession(connector=conn, timeout=timeout)
+            self.session = aiohttp.ClientSession(connector=conn, timeout=timeout, headers=headers)
```
bmbouter, line 148: :+1:

pulpcore/download/factory.py
```diff
@@ -15,7 +15,15 @@
...
 
-PROTOCOL_MAP = {"http": HttpDownloader, "https": HttpDownloader, "file": FileDownloader}
+import json
+
+
+PROTOCOL_MAP = {
+    "http": HttpDownloader,
+    "https": HttpDownloader,
+    "file": FileDownloader,
+    "uln": HttpDownloader,
```
bmbouter, line 25: Is `uln` an official protocol handler? As in do folks commonly write urls with `uln://` at the front? If it's not official we can move it to the DownloadFactory subclass in pulp_rpm. Let me know if you have any questions on how to do that, or if it's something you believe should be kept here.

pulpcore/app/serializers/repository.py
```diff
@@ -137,6 +139,9 @@ class RemoteSerializer(ModelSerializer):
         help_text="aiohttp.ClientTimeout.sock_read (q.v.) for download-connections.",
         min_value=0.0,
     )
+    headers = JSONField(
+        help_text="Headers for aiohttp.Clientsession",
```
bmbouter, line 143: I think I've led you slightly astray here when I recommended the JSONField for the serializer. What would be better is to have this be a ListField with children as dictionaries. The problem if we have a generic JSONField is we can't validate data in a way that guarantee's aiohttp can accept it. For example they could create {'asdf':{'anotherdict': {'yetanother': False}}} which wouldn't work when aiohttp goes to run. Sorry for this confusion I introduced.

Technically the [ClientSession docs]() state the `headers` field is "May be either iterable of key-value pairs". So to make that I recommend having your serializer field be similar to [this one in pulp_ansible](https://github.com/pulp/pulp_ansible/blob/c2e353c746c2b478486acb78c852e48d2b2a82f5/pulp_ansible/app/serializers.py#L299-L301).

The model field you've made can be left as-is.

pulpcore/app/models/repository.py
```diff
@@ -266,6 +268,7 @@ class Remote(MasterModel):
     sock_read_timeout = models.FloatField(
         null=True, validators=[MinValueValidator(0.0, "Timeout must be >= 0")]
     )
+    headers = JSONField(blank=True, null=True)
```
bmbouter, line 271: Do we really need `blank=True`? I am thinking we maybe don't because there are so many existing fields that use `null=True` without `blank=True`. We use DRF for validation so I think that is why.

bmbouter (changes requested):
Thank you for submitting this! Post any questions or comments. I'm not sure if it's possible, but if we can merge this by end of the day on Jan 27th that would be great to get it included in Pulpcore 3.10 which is releasing on the 28th.

pulpcore/app/serializers/repository.py
mdellweg: Isn't the header parameter a multidict even, because you can have the same key more then once?
e.g.
```
Link: '<SOMEURL>; rel="next"'
Link: '<OTHERURL>; rel="prev"'
```

pulpcore/download/http.py
ThikaXer: Done :)

pulpcore/app/models/repository.py
ThikaXer: Yes, I think we need it. I got an error when first using only `null=True` and found this discussion 
https://stackoverflow.com/questions/36209336/in-django-1-9-whats-the-convention-for-using-jsonfield-native-postgres-jsonb
explaining that one needs `null=True, blank=True` to make a JSON field optional in django.

pulpcore/app/serializers/repository.py
bmbouter: @mdellweg yes it is that's a good point; I agree. I believe we'll be facilitating that as something like: `[{'Link': '<SOMEURL>; rel="next"'}, {'Link: '<OTHERURL>; rel="prev"'}]`. Does this sound right to you?

pulpcore/download/factory.py
bmbouter: @ThikaXer what about this question?

ThikaXer: I don't think `uln` is an official protocol handler, at least it is not listed in https://en.wikipedia.org/wiki/List_of_URI_schemes
Still, Oracle uses `uln://repository` as a short hand notation for their repositories. At the moment, the API Endpoint to ULN seems broken, so I can't test the ULN repository sync :( . I guess, we could omit this line. Is
https://github.com/pulp/pulp_rpm/blob/f85c5eb0a04d5f7baf7bb336bb7635511dd723ff/pulp_rpm/app/models/repository.py#L82
enough to substitute this? Thanks for your help!

pulpcore/app/serializers/repository.py
mdellweg: Either that, or just a list of pairs (=list of two).

pulpcore/download/factory.py
bmbouter: Yes moving it to the pulp_rpm subclassed `DownloaderFactory` is the right way to handle this. Thank you for moving it there.

bmbouter (approved):
Thank you for this contribution!

bmbouter (changes requested):
Oh actually I need to ask for two small bookeeping things.

1) Can you mark yourself as the assignee on the redmine issue: https://pulp.plan.io/issues/8083
2) Also can you edit your commit message to have:
* a shorter first line (it's also out of date with uln removed)
* A description
* Remove the brackets from around `fixes...` because the automation won't recognize it with the brackets. The brackets are only used for `[noissue]`.

bmbouter:
@ThikaXer if you don't have permissions to assign yourself to the issue please as in #pulp-dev on IRC to have your redmine account ID added to the 'contributors' group.

dkliban (approved):

dkliban:
I added @ThikaXer to the contributor group on pulp.plan.io.

bmbouter (approved):
Thank you!

bmbouter:
Issue:  https://pulp.plan.io/issues/8083

PR #1045, opened 2020-12-02
Moved UnsupportedDigestValidationError to its new home in pulpcore.pl…

ggainey:
…ugin.exceptions.

closes #7908

### DIFF
--- a/CHANGES/plugin_api/7908.deprecation
+++ b/CHANGES/plugin_api/7908.deprecation
@@ -1,0 +1,5 @@
+The ``pulpcore.plugin.models.UnsupportedDigestValidationError`` is being deprecated and
+will be removed in 3.10.
+
+It can now be found at ``pulpcore.plugin.exceptions.UnsupportedDigestValidationError``
+instead; please change any code that imports it to access it from its new location.
--- a/pulpcore/plugin/models/__init__.py
+++ b/pulpcore/plugin/models/__init__.py
@@ -1,6 +1,9 @@
 # Models are exposed selectively in the versioned plugin API.
 # Any models defined in the pulpcore.plugin namespace should probably be proxy models.
 
+# THIS WILL BE DEPRECATED In 3.10 - LOOK TO .plugins.exceptions INSTEAD
+from pulpcore.exceptions.validation import UnsupportedDigestValidationError  # noqa
+
 from pulpcore.app.models import (  # noqa
     AccessPolicy,
     AutoAddObjPermsMixin,
@@ -35,7 +38,6 @@ from pulpcore.app.models import (  # noqa
     SigningService,
     Task,
     TaskGroup,
-    UnsupportedDigestValidationError,
     Upload,
     UploadChunk,
 )
--- a/pulpcore/tests/unit/models/test_content.py
+++ b/pulpcore/tests/unit/models/test_content.py
@@ -7,13 +7,16 @@ from django.core.files.uploadedfile import SimpleUploadedFile
 
 from django.conf import settings
 from django.test import SimpleTestCase, TestCase
-from pulpcore.plugin.exceptions import MissingDigestValidationError
+from pulpcore.plugin.exceptions import (
+    UnsupportedDigestValidationError,
+    MissingDigestValidationError,
+)
+
 from pulpcore.plugin.models import (
     Artifact,
     Content,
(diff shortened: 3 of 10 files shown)


CHANGES/plugin_api/7908.deprecation
```diff
@@ -0,0 +1,4 @@
+The pulpcore.plugin.models.UnsupportedDigestValidationError is being deprecated and will be removed in 3.10.
```
daviddavis, line 1: nitpick - should probably escape these class names this with double backticks.

daviddavis (approved):
One small comment. Other than that, LGTM.

pulpcore/plugin/models/__init__.py
```diff
@@ -1,6 +1,9 @@
 # Models are exposed selectively in the versioned plugin API.
 # Any models defined in the pulpcore.plugin namespace should probably be proxy models.
 
+# THIS WILL BE DEPRECATED SOON - LOOK TO .plugins.exceptions INSTEAD
```
bmbouter, line 4: I think calling out the version of Pulp this will be removed would be good. Even if we forget to specifically in that version, we'll easily know it's safe anytime at that version or later.

pulpcore/tests/unit/models/test_content.py
```diff
@@ -7,13 +7,16 @@
...
+    MissingDigestValidationError,
+)
+
 from pulpcore.plugin.models import (
     Artifact,
     Content,
     ContentArtifact,
     PulpTemporaryFile,
-    UnsupportedDigestValidationError,
+    UnsupportedDigestValidationError,  # will fail when UDVE is fully deprecated/removed from here
```
bmbouter, line 19: Can't we switch this to the new defined location? Since it's internal code we don't have to continue importing through the old location.

CHANGES/plugin_api/7908.deprecation
```diff
@@ -0,0 +1,3 @@
+The ``pulpcore.plugin.models.UnsupportedDigestValidationError`` is being deprecated and will be removed in 3.10.
+
+It can now be found at ``pulpcore.plugin.exceptions.UnsupportedDigestValidationError`` instead; please change any code that imports it to access it from its new location.
```
bmbouter, line 3: Can this be line wrapped to 70 chars.

bmbouter:
One small nitpick: can the commit be reworked to fit to have the commit title be shorted to fit within the char limit for git commit titles? https://docs.pulpproject.org/pulpcore/contributing/git.html#commit-message

pulpcore/tests/unit/models/test_content.py
ggainey: I left it in deliberately in a "this will fail if we do the removal and have (somehow) forgotten that it's a deprecation that needs to be specifically announced"

CHANGES/plugin_api/7908.deprecation
ggainey: I wish our local tests would test all the rules.

ggainey:
> One small nitpick: can the commit be reworked to fit to have the commit title be shorted to fit within the char limit for git commit titles? https://docs.pulpproject.org/pulpcore/contributing/git.html#commit-message

See prev comment. Will do.

CHANGES/plugin_api/7908.deprecation
bmbouter: me too

pulpcore/plugin/models/__init__.py
ggainey: +1

CHANGES/plugin_api/7908.deprecation
ggainey: but done :)

pulpcore/tests/unit/models/test_content.py
bmbouter: I don't understand. Keeping this here won't help us remember to remove the deprecated one later, but it will keep the deprecated one in use which we could switch off of now.

ggainey: Okeydoke - switched!

bmbouter (approved):
Thank you @ggainey ! :1st_place_medal: :tada:

PR #1046, opened 2020-12-02
Adds additional lock cleanup to worker cleanup

bmbouter:
As another layer of security to guard against lock cleanup not occuring
due to Redis not delivering the _release_resource task, ensure all locks
are also cleaned up even for tasks that are in their final states.

closes #7907


dralley (approved):

daviddavis (approved):
LGTM

PR #1054, opened 2020-12-10
Fix import/export bug when sha384 or sha512 are null

daviddavis:
fixes #7836

### DIFF
--- a/pulpcore/app/modelresource.py
+++ b/pulpcore/app/modelresource.py
@@ -8,6 +8,7 @@ from pulpcore.app.models.content import (
     ContentArtifact,
 )
 from pulpcore.app.models.repository import Repository
+from pulpcore.constants import ALL_KNOWN_CONTENT_CHECKSUMS
 from pulpcore.plugin.importexport import QueryModelResource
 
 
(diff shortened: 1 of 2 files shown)


ipanova (approved):

pulpcore/app/modelresource.py
```diff
@@ -19,6 +19,24 @@
...
+            row (tablib.Dataset row): incoming import-row representing a single Variant.
+            kwargs: args passed along from the import() call.
+
+        """
+        # the export converts None to blank strings but sha384 and sha512 have unique constraints
+        # that get triggered if they are blank. convert them back into None if they are blank.
+        if row["sha384"] == "":
+            row["sha384"] = None
+        if row["sha512"] == "":
+            row["sha512"] = None
```
mdellweg, line 38: Should we not do it for all digests for consistency?

daviddavis: It's not necessary since the other checksums aren't unique. I don't think it would hurt though.

pulpcore/app/modelresource.py
```diff
@@ -19,6 +20,23 @@
 # repo-version-specific.
 #
 class ArtifactResource(QueryModelResource):
+    """Resource for import/export of artifacts."""
+
+    def before_import_row(self, row, **kwargs):
+        """
+        Sets sha384 and sha512 to None if they are blank strings.
```
mdellweg, line 27:
```suggestion
        Sets digests to None if they are blank strings.
```

mdellweg (approved):

pulpcore/app/modelresource.py
mdellweg: It feels weird to do something with only a subset of those otherwise very symmetric digests.

PR #1056, opened 2020-12-10
Add filter to users and groups endpoints

mdellweg:
fixes #7975
https://pulp.plan.io/issues/7975

### DIFF
--- a/pulpcore/app/viewsets/user.py
+++ b/pulpcore/app/viewsets/user.py
@@ -5,14 +5,16 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group, Permission
 from django.core.exceptions import FieldError
 from django.shortcuts import get_object_or_404
+from django_filters.rest_framework import DjangoFilterBackend
 from drf_spectacular.utils import extend_schema
 from guardian.models.models import GroupObjectPermission
 from rest_framework import mixins, status
 from rest_framework.decorators import action
+from rest_framework.filters import OrderingFilter
 from rest_framework.response import Response
 from rest_framework.serializers import ValidationError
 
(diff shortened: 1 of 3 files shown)


daviddavis (approved):
LGTM. A test would be nice but I leave that at your discretion.

mdellweg:
Writing a test for the user filters seems to be hard, as long as you cannot manipulate them though the rest API.

pulpcore/app/viewsets/user.py
```diff
@@ -23,6 +25,23 @@
...
+class UserFilter(BaseFilterSet):
+    """
+    FilterSet for User.
+    """
+
+    class Meta:
+        model = User
+        fields = {
+            "username": ["exact", "in", "contains"],
+            "first_name": ["exact", "in", "contains"],
```
ipanova, line 37: i suggest adding 'icontains' for case insensitive match on the fileds where it makes sense

ipanova (approved):
a suggestion but lgtm in an ycase https://github.com/pulp/pulpcore/pull/1056/files#r542696899

pulpcore/app/viewsets/user.py
```diff
@@ -32,6 +51,8 @@ class UserViewSet(NamedModelViewSet, mixins.RetrieveModelMixin, mixins.ListModel
     """
 
     endpoint_name = "users"
+    filterset_class = UserFilter
+    filter_backends = (OrderingFilter, DjangoFilterBackend)
```
ipanova, line 55: will you also specify ordering?

mdellweg: I added some tests too.

ipanova (approved):

PR #1058, opened 2020-12-14
Prevent tasking system deadlocks when Redis is restarted/data lost

dralley:


pulpcore/tasking/worker_watcher.py
```diff
@@ -148,11 +160,10 @@ def mark_worker_offline(worker_name, normal_shutdown=False):
...
     if not normal_shutdown:
         msg = _("The worker named %(name)s is missing. Canceling the tasks in its queue.")
-        msg = msg % {"name": worker_name}
-        _logger.error(msg)
+        _logger.error(msg % {"name": worker_name})
     else:
-        msg = _("Cleaning up shutdown worker '%s'.") % worker_name
-        _logger.info(msg)
+        _logger.info(_("Worker '%s' shutdown") % worker_name)
+        _logger.info(_("Cleaning up shutdown worker '%s'.") % worker_name)
```
mdellweg, line 166: Nitpick: I have seen us using `.format(...)` in the other places.

mdellweg (commented):
Is there a 1 to 1 relationship of workers and queues? In that case it should be possible merge the cleanup into one routine.

As you do not enque the resource job anymore, you should have a look at this line too:
https://github.com/pulp/pulpcore/blob/25e92d0d65bcac08994a420b8c45a2094233fe32/pulpcore/tasking/util.py#L49

dralley:
This resource job is actually the "resource manager job" whereas the one I got rid of was the "release resources job".  So unfortunately we still need it since they do different things.

pulpcore/tasking/worker_watcher.py
```diff
@@ -115,19 +120,33 @@ def check_worker_processes():
     _logger.debug(msg)
 
 
-def handle_worker_offline(worker_name):
-    """
-    This is a generic function for handling workers going offline.
+def check_dropped_queues():
```
bmbouter, line 123: This name could be clearer I think. Somehow working in the idea that it also cancels would do it for me.

bmbouter: Maybe `check_and_cancel_missing_tasks` ?

pulpcore/tasking/worker_watcher.py
```diff
@@ -115,19 +120,33 @@ def check_worker_processes():
...
+    """Cancel any unexecuted Tasks which are no longer in the RQ registry.
 
-    _delete_worker() task is called to handle any work cleanup associated with a worker going
-    offline. Logging at the info level is also done.
+    In some situations such as a restart of Redis, Jobs can be dropped from the Redis
+    queues and "forgotten". Therefore the Task will never be marked completed in Pulp.
```
bmbouter, line 127: Another sentence here would be good I think. It could add that it's problematic because tasks that never complete never release their resources and that causes the tasking system to deadlock.

pulpcore/tasking/worker_watcher.py
```diff
@@ -115,19 +120,33 @@ def check_worker_processes():
...
+        state__in=TASK_INCOMPLETE_STATES, worker__in=Worker.objects.online_workers()
+    )
+
+    for task in assigned_and_unfinished_tasks:
+        try:
+            Job.fetch(str(task.pk), connection=redis_conn)
+        except NoSuchJobError:
+            cancel(task.pk)
+
+    # Also go through all of the tasks that were still queued up on the resource manager
```
bmbouter, line 144: Niiiiice :+1:

CHANGES/7912.bugfix
```diff
@@ -0,0 +1 @@
+Provide a mechanism to automatically resolve issues and prevent deadlocks when Redis experiences data loss (such as a restart).
```
bmbouter, line 1: nitpick: add a newline at the end?

bmbouter:
I ran some hand tests where I simulated OOM kill against a long running task and it was marked as failed nicely.

bmbouter (changes requested):
I think this PR is very excellent. I am requesting very small docstring, function name, and a newline change.

Thank you for making this! It's much simpler and better! :1st_place_medal:

dkliban (approved):
This is a great improvement!

pulpcore/tasking/worker_watcher.py
```diff
@@ -115,19 +120,35 @@ def check_worker_processes():
...
+    We go through all of the tasks which are in an incomplete state and check that RQ
+    still has a record of the job. If not, we cancel it.
     """
-    msg = _("Worker '%s' shutdown") % worker_name
-    _logger.info(msg)
-    mark_worker_offline(worker_name, normal_shutdown=True)
+    redis_conn = connection.get_redis_connection()
+
+    assigned_and_unfinished_tasks = Task.objects.filter(
+        state__in=TASK_INCOMPLETE_STATES, worker__in=Worker.objects.online_workers()
```
dralley, line 137: @bmbouter Should this be ```worker__isnull=False```?  In theory I guess you could unplug a machine and after a while the worker is no longer "online", but it's task would still be assigned.

dralley: Although I suppose the worker watcher should handle that

bmbouter: Yeah I tested this actually I pkilled all workers including parents and their workhourse children. They remained in the running state until the worker watcher timeout occurred and they were transitioned to cancelled by the resource manager running its heartbeat check.

bmbouter (approved):
Thank you!

PR #1059, opened 2020-12-16
Handle request=None in openapi schema generation

alikins:
No-Issue

### DIFF
--- a/pulpcore/openapi/__init__.py
+++ b/pulpcore/openapi/__init__.py
@@ -389,7 +389,7 @@ class PulpSchemaGenerator(SchemaGenerator):
                 continue
 
             # Removes html tags from OpenAPI schema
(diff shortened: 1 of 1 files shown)


pulpcore/openapi/__init__.py
```diff
@@ -389,7 +389,7 @@ def parse(self, request, public):
                 continue
 
             # Removes html tags from OpenAPI schema
-            if "include_html" not in request.query_params:
+            if request and "include_html" not in request.query_params:
```
mdellweg, line 392: As "include_html" needs to be added, i'd like this to default to strip the tags:
```suggestion
            if not (request or "include_html" in request.query_params):
```

mdellweg:
This has been solved in
https://github.com/pulp/pulpcore/pull/1065

PR #1061, opened 2020-12-21
User get_user_model in access_policy

mdellweg:
In order to support customized user models, the User model should be
determined with the get_user_model method.

[noissue]


mdellweg:
I am not even sure, that this fixes a bug, but django docs ...

https://docs.djangoproject.com/en/2.1/topics/auth/customizing/#referencing-the-user-model

fao89 (approved):
Nice catch!

daviddavis (approved):

ipanova (approved):

PR #1062, opened 2020-12-22
Have a more descriptive artifact error message

### LINKED ISSUE #3387: File System Exporter should have an option to generate listing files
When Katello uses the File System exporter it does not generate the listing files in CDN directory structure. Katello does not directly have access to the pulp's content directories. Due to this it is not able to easily generate `Listing` files. It relies right now on the hammer export command being run from the same machine as pulp.

Pulp exporter should ideally have an option or a flag to generate `listing` files when fs exporter is run.

Couple of things to consider here 
- This is only relevant to YUM/KS repos
- This probably only useful for certain Yum repos. But a flag in the exporter would take care of that.

daviddavis:
fixes #3387

### DIFF
--- a/pulpcore/app/serializers/content.py
+++ b/pulpcore/app/serializers/content.py
@@ -1,4 +1,3 @@
-import hashlib
 from gettext import gettext as _
 
 from django.db import transaction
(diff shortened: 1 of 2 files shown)


daviddavis:
It appears that the ordering of checksums here is not deterministic somehow. Sometimes I get `Artifact with sha256 checksum of '...' already exists.` while other times I get `Artifact with sha384 checksum of '...' already exists.`

daviddavis:
The problem is that `hashlib.algorithms_guaranteed` returns a set which isn't ordered (and so therefore the error message one gets is nondeterministic). I solved it by iterating through `Artifact.DIGEST_FIELDS`.

ipanova:
@daviddavis we have just recently discussed this topic a bit specifically how to improve the upload experience. I am hoping that we can instead handle dup artifact during the upload and not  fail with any of the errors but just return href to the existing artifact. Here is a story with more info https://pulp.plan.io/issues/7791and https://pulp.plan.io/issues/7114. What do you think?

daviddavis:
@ipanova thanks. I knew about 7114 but hadn't seen the discussion on 7791. I left some comments.

ipanova:
> @ipanova thanks. I knew about 7114 but hadn't seen the discussion on 7791. I left some comments.

ty, your observation about backwards incompatible change of return code is valid, i have updated the issue accordingly.

mdellweg (approved):

ipanova (approved):

CHANGES/3387.feature
```diff
@@ -0,0 +1 @@
+Updated duplicate artifact error message to be more descriptive.
```
ipanova, line 1: I would probably keep just one entry log.

pulpcore/app/serializers/content.py
```diff
@@ -219,24 +218,24 @@ def validate(self, data):
...
-                    data[algorithm] = digest
-                if algorithm in models.Artifact.RELIABLE_DIGEST_FIELDS:
-                    validator = UniqueValidator(
-                        models.Artifact.objects.all(),
-                        message=_("{0} checksum must be unique.").format(algorithm),
-                    )
-                    validator.instance = None
-
-                    validator(digest, self.fields[algorithm])
+        for algorithm in reversed(models.Artifact.DIGEST_FIELDS):
```
ipanova, line 221: out of curiosity question,  any reason to use reversed() here? digest_fields is a list so  message should not change as it was changing with algorithms_guaranteed which was a set?

ipanova: probably because of perf reasons since we started with the weakest alg

daviddavis: Yea, so like right now you can get one of a variety of different error messages depending on the order of checksum types. With this change, you will always get the same error message with the same checksum type. Without this reverse, it starts at the highest so the error message would be with sha512. However, with the reversed, you get the lowest unique checksum sha256. sha256 seems like the more useful checksum to display to the user since it's how we identify artifacts on the filesystem, file content, etc.

PR #1064, opened 2020-12-23
Restrict default permissions to admin users

mdellweg:
fixes #8018
https://pulp.plan.io/issues/8018

### DIFF
--- a/CHANGES/8018.bugfix
+++ b/CHANGES/8018.bugfix
@@ -1,0 +1,1 @@
+Changed the default permission class to ``IsAdminUser`` to protect endpoints not yet guarded by an access policy from users without permission.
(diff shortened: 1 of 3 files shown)


mdellweg:
@jlsherrill We want to confirm with katello if you are always using the admin user.
This should not impact you then.

jlsherrill:
That's correct!

ipanova (approved):

ipanova:
@bmbouter PTAL

bmbouter:
@newswangerd FYI on this, does galaxy_ng use any non-admin users that you don't define Permission classes explicitly for?

bmbouter (changes requested):
I think we need a plugin writer `.removal` note calling out this breaking change. For example if a plugin is shipping a viewset that they need non-admin users to have access to they would want to see this in the plugin writer changelog. What do you all think about this?

newswangerd:
> @newswangerd FYI on this, does galaxy_ng use any non-admin users that you don't define Permission classes explicitly for?

Thanks for the heads up @bmbouter. I don't think this should affect galaxy_ng too much.

bmbouter (approved):
Thanks!

PR #1065, opened 2020-12-23
Validate OpenAPI schema

fao89:
[noissue]

### DIFF
--- a/.ci/scripts/schema.py
+++ b/.ci/scripts/schema.py
@@ -1,0 +1,23 @@
+"""
+Customizing OpenAPI validation.
+
+OpenAPI requires paths to start with slashes:
+https://spec.openapis.org/oas/v3.0.3#patterned-fields
+
+But some pulp paths start with curly brackets e.g. {artifact_href}
+This script modifies drf-spectacular schema validation to accept slashes and curly brackets.
+"""
+import json
+from drf_spectacular.validation import JSON_SCHEMA_SPEC_PATH
+
+with open(JSON_SCHEMA_SPEC_PATH) as fh:
+    openapi3_schema_spec = json.load(fh)
+
+properties = openapi3_schema_spec["definitions"]["Paths"]["patternProperties"]
+# Making OpenAPI validation to accept paths starting with / and {
+if "^\\/|{" not in properties:
+    properties["^\\/|{"] = properties["^\\/"]
+    del properties["^\\/"]
+
+with open(JSON_SCHEMA_SPEC_PATH, "w") as fh:
+    json.dump(openapi3_schema_spec, fh)
--- a/.github/workflows/scripts/script.sh
+++ b/.github/workflows/scripts/script.sh
@@ -31,6 +31,11 @@ if [[ "$TEST" = "docs" || "$TEST" = "publish" ]]; then
   make PULP_URL="http://pulp" html
   cd ..
 
+  echo "Validating OpenAPI schema..."
+  cat $PWD/.ci/scripts/schema.py | cmd_stdin_prefix bash -c "cat > /tmp/schema.py"
+  cmd_prefix bash -c "python /tmp/schema.py"
+  cmd_prefix bash -c "pulpcore-manager spectacular --file pulp_schema.yml --validate"
+
   if [ -f $POST_DOCS_TEST ]; then
     source $POST_DOCS_TEST
   fi
--- a/pulpcore/app/serializers/user.py
+++ b/pulpcore/app/serializers/user.py
@@ -37,11 +37,11 @@ class PermissionSerializer(serializers.Serializer):
     permission = PermissionField(source="*", read_only=True)
     obj = ContentObjectField(help_text=_("Content object."), source="*", read_only=True)
 
-    def get_id(self, obj):
+    def get_id(self, obj) -> int:
         """Get model/object permission id."""
         return obj.id
 
-    def get_pulp_href(self, obj):
+    def get_pulp_href(self, obj) -> str:
         """Get model/object permission pulp_href."""
         group_pk = self.context.get("group_pk")
 
--- a/pulpcore/openapi/__init__.py
+++ b/pulpcore/openapi/__init__.py
@@ -389,7 +389,7 @@ class PulpSchemaGenerator(SchemaGenerator):
                 continue
 
             # Removes html tags from OpenAPI schema
-            if "include_html" not in request.query_params:
+            if request is None or "include_html" not in request.query_params:
                 operation["description"] = strip_tags(operation["description"])
 
             # operationId as actions [list, read, sync, modify, create, delete, ...]


pulpcore/app/serializers/user.py
```diff
@@ -37,10 +39,12 @@ class PermissionSerializer(serializers.Serializer):
     permission = PermissionField(source="*", read_only=True)
     obj = ContentObjectField(help_text=_("Content object."), source="*", read_only=True)
 
+    @extend_schema_field(OpenApiTypes.INT)
     def get_id(self, obj):
```
mdellweg, line 43: I figured, that it's also possible to use `def get_id(self, obj) -> int:`, because spectacular is looking for type hints.

pulpcore/openapi/__init__.py
```diff
@@ -423,8 +424,9 @@ def parse(self, request, public):
             if path.startswith("/"):
                 path = path[1:]
 
-            if not path.startswith("{"):
-                path = urljoin(self.url or "/", path)
+            path = urljoin(self.url or "/", path)
```
mdellweg, line 427: So this will lead to paths like `"/{task_href}"` right? And after proper substitution it will be `//pulp/api/v3/tasks/5/` with the double slashes we do not want. I urge you to at least test this with the current implementation of pulp-cli and squeezer.
I'd rather make an exception to the linter here if possible.

pulpcore/openapi/__init__.py
```diff
@@ -389,7 +390,7 @@ def parse(self, request, public):
                 continue
 
             # Removes html tags from OpenAPI schema
-            if "include_html" not in request.query_params:
+            if request and "include_html" not in request.query_params:
```
mdellweg, line 393:
```suggestion
            if request is None or "include_html" not in request.query_params:
```
As `include_html` is to be requested specifically, we should default to `strip_tags`.

pulpcore/openapi/__init__.py
```diff
@@ -454,4 +456,11 @@ def get_schema(self, request=None, public=False):
         server_url = "http://localhost:24817" if not request else request.build_absolute_uri("/")
         result["servers"] = [{"url": server_url}]
 
-        return normalize_result_object(result)
+        schema = normalize_result_object(result)
+
+        if request and "bindings" not in request.query_params:
+            # Validation fails when path does not start with /
+            # But bindings require some paths without /
+            validate_schema(schema)
```
mdellweg, line 464: Is this function dependant on the `--validate` flag?

fao89: No, it isn't.
`--validate` calls this, but I put it here just to fail early

pulpcore/openapi/__init__.py
fao89: it is a temporary thing, I did it just to please the CI, but OpenAPI requires it to start with slash: https://spec.openapis.org/oas/v3.0.3#patterned-fields
And I think we need to discuss a better way to handle this, I was thinking about introducing a new query param, maybe `?validate=true`. An exception to the linter would be good, but so far I couldn't figure out how to do it.

pulpcore/app/serializers/user.py
mdellweg: Not saying that it's the preferred solution...

pulpcore/openapi/__init__.py
mdellweg: Hmm, this sounds a bit like you are going to volkswagen it (make it comply selectively in a test environment).

pulpcore/openapi/__init__.py
mdellweg: so it's for debugging only?

fao89: not exactly debugging (but yes, I'm testing what is the best way to be sure the schema is valid)
`--validate` will fail only if you do: `pulpcore-manager spectacular --validate`
`validate_schema` will give a `500` when you hit `/pulp/api/v3/docs/api.json`

pulpcore/openapi/__init__.py
fao89: I think I found the place we need to change for solving the slashes thing:
https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/validation/openapi3_schema.json#L60
https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/validation/__init__.py#L6

pulpcore/openapi/__init__.py
mdellweg: I think there's a good reason for the explicit `--validate` flag on the commandline tool.

daviddavis:
I am not 100% clear on what .ci/scripts/schema.py is doing. It looks like it's fetching, munging, and saving the schema (but I am not sure why)? Could you add some comments to the file to explain its purpose?

daviddavis (approved):
One small suggestion. Other than that, looks good to me.

pulpcore/openapi/__init__.py
fao89: > I am not 100% clear on what .ci/scripts/schema.py is doing. It looks like it's fetching, munging, and saving the schema (but I am not sure why)? Could you add some comments to the file to explain its purpose?

I've added a docstring

fao89:
> I am not 100% clear on what .ci/scripts/schema.py is doing. It looks like it's fetching, munging, and saving the schema (but I am not sure why)? Could you add some comments to the file to explain its purpose?

I've added a docstring

.ci/scripts/schema.py
```diff
@@ -0,0 +1,23 @@
+"""
+Customizing OpenAPI validation.
+
+OpenAPI requires paths to start with slashes:
+https://spec.openapis.org/oas/v3.0.3#patterned-fields
+
+But some pulp paths starts with curly brackets e.g. {artifact_href}
```
mdellweg, line 7:
```suggestion
But some pulp paths start with curly brackets e.g. "{artifact_href}"
```

.github/workflows/scripts/script.sh
```diff
@@ -31,6 +31,11 @@ if [[ "$TEST" = "docs" || "$TEST" = "publish" ]]; then
   make PULP_URL="http://pulp" html
   cd ..
 
+  echo "Validating OpenAPI schema..."
+  cat $PWD/.ci/scripts/schema.py | cmd_stdin_prefix bash -c "cat > /tmp/schema.py"
+  cmd_prefix bash -c "python /tmp/schema.py"
```
mdellweg, line 36: Just for my clarification: So this is not changing the pulp-api-schema, but the openapi-schema-spec-schema so that it validates out api.

mdellweg (approved):
Thank you! All comments have been addressed, IMHO.

.github/workflows/scripts/script.sh
fao89: exactly!

PR #1066, opened 2021-01-04
Fix openapi doc paths for nested endpoints

mdellweg:
This disambiguates the group permission paths in the api docs and uses
the proper parent viweset href for nested "list" and "create" views.

### DIFF
--- a/pulpcore/app/viewsets/user.py
+++ b/pulpcore/app/viewsets/user.py
@@ -121,6 +121,7 @@ class GroupModelPermissionViewSet(NamedModelViewSet):
     parent_lookup_kwargs = {"group_pk": "group__pk"}
     serializer_class = PermissionSerializer
     queryset = Permission.objects.all()
+    pulp_model_alias = "ModelPermission"
 
     def get_model_permission(self, request):
         """Get model permission"""
@@ -211,6 +212,7 @@ class GroupObjectPermissionViewSet(NamedModelViewSet):
     parent_lookup_kwargs = {"group_pk": "group__pk"}
     serializer_class = PermissionSerializer
     queryset = Permission.objects.all()
+    pulp_model_alias = "ObjectPermission"
 
     def get_object_pk(self, request):
         """Get object pk."""
--- a/pulpcore/openapi/__init__.py
+++ b/pulpcore/openapi/__init__.py
@@ -284,19 +284,21 @@ class PulpSchemaGenerator(SchemaGenerator):
     """Pulp Schema Generator."""
 
     @staticmethod
-    def get_parameter_slug_from_model(model, prefix):
+    def get_parameter_slug_from_model(model, prefix, pulp_model_alias=None):
         """Returns a path parameter name for the resource associated with the model.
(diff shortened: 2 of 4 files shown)


pulpcore/openapi/__init__.py
```diff
@@ -336,23 +338,20 @@ def convert_endpoint_path_params(self, path, view, schema):
             resource_model = view.queryset.model
         if resource_model:
             prefix = None
+            model_alias = getattr(view, "model_alias", None)
             parent_viewset = getattr(view, "parent_viewset", None)
             if parent_viewset:
-                if schema._is_list_view():
+                if view.action in ["list", "create"]:
```
mdellweg, line 344: @ggainey This (i think) is where the export api is weird, as this substitution was not performed on "create" before.

Context: https://github.com/pulp/pulp-cli/blob/develop/pulpcore/cli/core/export.py#L97

fao89: could you explain why only `list` and `create` needs this?

mdellweg: Sure: `list` and `create` are the methods that do not reference an existing nested resource, so they will be like 
`GET {repository_href}/versions/` and `POST {repository_href}/versions/` with `repository_href="/pulp/api/v3/repositories/1234"`. (The reason we didn't see this as a bug with repo_versions is that they don't even have a create call.)

All the other methods like `show`, `update`, `partial_update` and `delete` need a nested resource to work on, in which case their path will look like `{repository_version_href}`.

fao89: gotcha! Thank you!

fao89 (approved):
Thank you!

pulpcore/app/viewsets/user.py
```diff
@@ -121,6 +121,7 @@ class GroupModelPermissionViewSet(NamedModelViewSet):
     parent_lookup_kwargs = {"group_pk": "group__pk"}
     serializer_class = PermissionSerializer
     queryset = Permission.objects.all()
+    model_alias = "ModelPermission"
```
fao89, line 124: I'm wondering if it would be better to make it `pulp_model_alias`, to be sure this name won't conflict with any DRF thing

daviddavis (approved):

pulpcore/app/viewsets/user.py
mdellweg: Will change that.

pulpcore/openapi/__init__.py
ggainey: > @ggainey This (i think) is where the export api is weird, as this substitution was not performed on "create" before.
> 
> Context: https://github.com/pulp/pulp-cli/blob/develop/pulpcore/cli/core/export.py#L97

Just getting to looking at this and at pulp_cli #52. It took some re-reading on my part before I understood what was going on/what you're doing here - but the lightbulb finally lit.  Thanks for the catch, and the explanation!

dkliban (approved):

PR #1070, opened 2021-01-13
Add object labels

daviddavis:
fixes #8065


pulpcore/app/models/base.py
```diff
@@ -1,12 +1,28 @@
...
 from django.db.models import options
 from django.db.models.base import ModelBase
 from django_lifecycle import LifecycleModel
 
 
+class Label(LifecycleModel):
+    """Class for handling resource labels."""
+
+    pulp_id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
```
mdellweg, line 16: When is this cascade triggered? When a ContentType is deleted/uninstalled?

pulpcore/app/serializers/base.py
```diff
@@ -122,6 +122,29 @@ def __init_subclass__(cls, **kwargs):
...
+                instance.pulp_labels.create(key=key, value=value)
+
+    def create(self, validated_data):
+        labels = validated_data.pop("pulp_labels", {})
+        instance = super().create(validated_data)
+        self._update_labels(instance, labels)
+        return instance
+
+    def update(self, instance, validated_data):
+        labels = validated_data.pop("pulp_labels", {})
```
mdellweg, line 143: This will delete all labels on update if none are provided, right?
Also labels can only be updated as the whole bunch if i read that correctly.

mdellweg: What happens if you use "PATCH" instead of "PUT"?

pulpcore/app/models/base.py
daviddavis: That is correct.

pulpcore/app/serializers/base.py
daviddavis: Ah yea, good catch. Pushing fix and a test now.

pulpcore/tests/unit/models/test_repository.py
```diff
@@ -289,3 +290,14 @@ def test_next_version_with_multiple_versions(self):
...
+    def test_label_cascade(self):
+        """Check that a repo's labels are deleted when it's deleted."""
+        repo = Repository.objects.create(name=uuid4())
+        repo.CONTENT_TYPES = [Content]
+        repo.save()
+
+        label = repo.pulp_labels.create(key="brown", value="radagast")
+        self.assertTrue(Label.objects.filter(pk=label.pk).exists())
+        repo.delete()
+        self.assertFalse(Label.objects.filter(pk=label.pk).exists())
```
mdellweg, line 303: :medal_sports:

pulpcore/app/serializers/base.py
```diff
@@ -122,6 +122,55 @@ def __init_subclass__(cls, **kwargs):
...
+        Update the labels for a Model instance.
+
+        Args:
+            instance (pulpcore.app.models.BaseModel): instance with labels to update
+            labels (list): labels to set for the instance
+        """
+        instance.pulp_labels.exclude(key__in=labels.keys()).delete()
+
+        for key, value in labels.items():
+            label = instance.pulp_labels.filter(key=key).first()
```
mdellweg, line 136: Is this the same as
```suggestion
            label = instance.pulp_labels.get(key=key)
```
because there can never be more than one?

pulpcore/app/serializers/base.py
```diff
@@ -122,6 +122,55 @@ def __init_subclass__(cls, **kwargs):
...
+
+        Args:
+            validated_data (dict): Validated data to create instance
+
+        Returns:
+            instance: The created of resource
+        """
+        labels = validated_data.pop("pulp_labels", {})
+        instance = super().create(validated_data)
+        self._update_labels(instance, labels)
```
mdellweg, line 155: should there be transactions? Or does drf handle it for us?

mdellweg (approved):
Just two questions. Nothing to make me stop this.

pulpcore/app/serializers/base.py
daviddavis: I will dig into the drf code and find out.

pulpcore/app/serializers/base.py
daviddavis: Yea, there should only be one. Although `get()` raises an exception if there is no match where as `first()` returns None. I could rewrite this code with `get()` though:

```python
for key, value in labels.items():
    try:
        label = instance.pulp_labels.get(key=key)
        if label.value != value:
            instance.pulp_labels.filter(key=key).update(value=value)
    except Label.DoesNotExist:
        instance.pulp_labels.create(key=key, value=value)
```

pulpcore/app/migrations/0054_label.py
```diff
@@ -0,0 +1,31 @@
...
+import django.db.models.deletion
+import django_lifecycle.mixins
+import uuid
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('contenttypes', '0002_remove_content_type_name'),
+        ('core', '0053_remote_headers'),
```
bmbouter, line 13: Sorry needs one more rebase I think `0054_add_public_key.py`

pulpcore/app/serializers/fields.py
```diff
@@ -374,3 +375,23 @@ def __init__(self, state, *args, **kwargs):
...
         return instance.tasks.filter(state=self.state).count()
+
+
+class LabelsField(serializers.DictField):
+    def to_representation(self, labels):
+        return {label.key: label.value for label in labels.all()}
+
+    def to_internal_value(self, data):
+        if type(data) != dict:
+            raise serializers.ValidationError(_("data must be supplied as a key/value hash"))
```
bmbouter, line 386:
```suggestion
            raise serializers.ValidationError(_("Data must be supplied as a key/value hash."))
```

bmbouter: Similar for validation error strings below too.

bmbouter: I interpret these as sentences but if you don't mean it that way leave them.

pulpcore/app/migrations/0054_label.py
daviddavis: Ha yea, I just reordered this a couple hours ago.

pulpcore/app/models/base.py
```diff
@@ -27,6 +60,7 @@ class BaseModel(LifecycleModel):
     pulp_id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     pulp_created = models.DateTimeField(auto_now_add=True)
     pulp_last_updated = models.DateTimeField(auto_now=True, null=True)
+    pulp_labels = GenericRelation(Label)
```
bmbouter, line 63: I (maybe naively) expected this would generate a migration. But it seems to be only in the serializers? I can see how the `LabelsField` handles receiving and returning data. I think i'm forgetting how GenericRelation works.

CHANGES/8065.feature
```diff
@@ -0,0 +1 @@
+Added pulp_labels to allow users to add key/value data to objects.
```
bmbouter, line 1: Can you add plugin_api release notes on the new objects added to the plugin API please?

pulpcore/app/serializers/fields.py
```diff
@@ -374,3 +375,23 @@ def __init__(self, state, *args, **kwargs):
 
     def get_attribute(self, instance):
         return instance.tasks.filter(state=self.state).count()
+
+
+class LabelsField(serializers.DictField):
```
bmbouter, line 380: Can this have a docstring?

bmbouter: And maybe the methods below here too?

pulpcore/app/models/base.py
daviddavis: Yea, this is the reverse of the `GenericForeignKey`.

pulpcore/app/serializers/base.py
mdellweg: Nah, i think it's ok. Thanks for the clarification.

pulpcore/app/serializers/base.py
```diff
@@ -122,6 +123,58 @@ def __init_subclass__(cls, **kwargs):
...
+        Created the resource from validated_data.
+
+        Args:
+            validated_data (dict): Validated data to create instance
+
+        Returns:
+            instance: The created of resource
+        """
+        labels = validated_data.pop("pulp_labels", {})
+        with transaction.atomic():
```
mdellweg, line 156: :+1:

daviddavis:
Ok I think I addressed everything. Will merge later today if there are no objections.

bmbouter (approved):
Thank you @daviddavis ! :+1: :fireworks:

bmbouter:
I see two lgtm's ... merging!

PR #1071, opened 2021-01-14
Allow looking up names using icontains, contains and startswith.

newswangerd:
fixes #8094

### DIFF
--- a/CHANGES/8094.feature
+++ b/CHANGES/8094.feature
@@ -1,0 +1,1 @@
+Allow querying names on the api using name__icontains, name__contains and name__startswith query parameters.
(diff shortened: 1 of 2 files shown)


bmbouter (approved):
This looks good to me. A test would be nice, but it's mostly a feature we're configuring not creating so I'm ok w/ this.

bmbouter:
@fao89 what do you think?

mdellweg (commented):
I like it.
Do we use this constant in all relevant places consistently?
`TaskFilter`, `UserFilter` and `GroupFilter` look like candidates. But the latter have also `"iexact"`.

Should we export it to the plugin api?

fao89:
> @fao89 what do you think?

Looks good to me

fao89 (approved):

bmbouter:
@dralley @daviddavis what do you think?

dralley (approved):

daviddavis (approved):

PR #1072, opened 2021-01-14
Add readonly repository viewset to the plugin API

goosemania:


mdellweg:
Once you have the commit machinery pass the linting stage and all the CI green, i'd be happy to approve.

pulpcore/app/viewsets/repository.py
```diff
@@ -43,15 +43,13 @@ class Meta:
         fields = {"name": NAME_FILTER_OPTIONS}
 
 
-class ImmutableRepositoryViewSet(
+class ReadOnlyRepositoryViewSet(
     NamedModelViewSet,
     mixins.CreateModelMixin,
```
ipanova, line 48: I bet this should not be here ;)

pulpcore/app/viewsets/repository.py
```diff
@@ -43,15 +43,13 @@ class Meta:
...
     mixins.RetrieveModelMixin,
-    mixins.ListModelMixin,
```
ipanova, line 50: We should probably leave the 'list'

pulpcore/app/viewsets/repository.py
goosemania: It was the typo, I fixed it. maybe at the same time you were reviewing.
As a reminder, it's a draft PR, I suggest not to spend time on it until it moves to a normal one.

daviddavis (approved):

ipanova (approved):

PR #1073, opened 2021-01-15
correct rq requirement

evgeni:
commas are important ;)

[noissue]


evgeni:
also should be backported to 3.9 where it's also wrong :)

dralley (approved):

mdellweg (approved):

PR #1078, opened 2021-01-20
Fix the default value of CHUNKED_UPLOAD_DIR

dkliban:
This patch changes the default value of CHUNKED_UPLOAD_DIR to `upload`.
This value is a relative path inside the MEDIA_ROOT. Doc strings have
also been updated to match the actual behavior of functions used to
form the path used by the UploadChunk's file field. The documentation
was also updated to make it more clear that CHUNKED_UPLOAD_DIR is
supposed to be a relative path inside MEDIA_ROOT.

fixes: #8099
https://pulp.plan.io/issues/8099

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -152,9 +152,11 @@ WORKING_DIRECTORY
 CHUNKED_UPLOAD_DIR
 ^^^^^^^^^^^^^^^^^^
 
-   A directory used exclusively for uploaded chunks. The uploaded chunks are stored in the default
-   storage specified by ``DEFAULT_FILE_STORAGE``. This option allows users to customize the actual
-   place where chunked uploads should be stored within the declared storage.
+   A relative path inside the MEDIA_ROOT directory used exclusively for uploaded chunks. The
+   uploaded chunks are stored in the default storage specified by ``DEFAULT_FILE_STORAGE``. This
+   option allows users to customize the actual place where chunked uploads should be stored within
+   the declared storage. The default, ``upload``, is sufficient for most use cases. A change to
+   this setting only applies to uploads created after the change.
 
 CONTENT_ORIGIN
 ^^^^^^^^^^^^^^
--- a/CHANGES/8099.removal
+++ b/CHANGES/8099.removal
@@ -1,0 +1,1 @@
+CHUNKED_UPLOAD_DIR was converted to a relative path inside MEDIA_ROOT.
--- a/pulpcore/app/models/storage.py
+++ b/pulpcore/app/models/storage.py
@@ -127,12 +127,12 @@ def get_temp_file_path(pulp_id):
 
 def get_upload_chunk_file_path(pulp_id):
     """
-    Determine the absolute path where a file backing an uploaded chunk should be stored.
+    Determine the relative path where a file backing an uploaded chunk should be stored.
 
     Args:
         pulp_id (uuid): An identifier identifying the file for UploadChunk
     Returns:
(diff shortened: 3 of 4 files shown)


daviddavis (approved):
LGTM. Needs a changelog entry.

docs/settings.rst
```diff
@@ -152,9 +152,10 @@ WORKING_DIRECTORY
 CHUNKED_UPLOAD_DIR
 ^^^^^^^^^^^^^^^^^^
 
-   A directory used exclusively for uploaded chunks. The uploaded chunks are stored in the default
-   storage specified by ``DEFAULT_FILE_STORAGE``. This option allows users to customize the actual
-   place where chunked uploads should be stored within the declared storage.
+   A relative path inside the MEDIA_ROOT directory used exclusively for uploaded chunks. The
+   uploaded chunks are stored in the default storage specified by ``DEFAULT_FILE_STORAGE``. This
+   option allows users to customize the actual place where chunked uploads should be stored within
+   the declared storage.
```
ekohl, line 158: Perhaps it should mention that this is only for advanced users. Also, if you change it, there is no code to move files and database references. It is only used for new files.

ekohl (commented):
This is an alternative to https://github.com/pulp/pulpcore/pull/1074.

ekohl (approved):
:+1: for the actual code change. I do think the docs could steer users away from changing it.

ekohl (approved):
If I was my project I'd suggest the imperative mood for the commit message (see https://chris.beams.io/posts/git-commit/#imperative for the reasoning), but I'll leave that up to the actual maintainers. Resulting change looks good to me.

mdellweg (approved):
I don't exactly see, how this variant survives without crying for a new migration, but as the tests are green i suppose it is correct.
I second @ekohl on the commit message style.

daviddavis:
> I don't exactly see, how this variant survives without crying for a new migration, but as the tests are green i suppose it is correct.

It's because [the field uses the `upload_to` callback rather than referring to the setting directly](https://github.com/pulp/pulpcore/blob/6e8f2c9377095fb01a069858daa37c5217fbbdf8/pulpcore/app/models/upload.py#L61).

ekohl:
What I would write as the commit message:

Make CHUNKED_UPLOAD_DIR a relative path

In 1b6c7360c30cc7d9e2d3b9fc5062ac1f7d69c2de uploads were changed to use the default storage (uses settings.MEDIA_ROOT). Anything that's written outside of storage location raises a SuspiciousOperation. That already made the implicit requirement that CHUNKED_UPLOAD_DIR was relative.

Users could hit this if they modified MEDIA_ROOT in their settings but kept CHUNKED_UPLOAD_DIR default.

If a relative path is used, Django prepends the location and it is guaranteed to be a safe location. This changes the default value to be relative and updates the documentation to reflect this.

daviddavis:
Actually now that I think of it: I wonder if we ought to have a `.removal` to say that `CHUNKED_UPLOAD_DIR` should be a relative path now instead of an absolute path?

dkliban:
> Actually now that I think of it: I wonder if we ought to have a `.removal` to say that `CHUNKED_UPLOAD_DIR` should be a relative path now instead of an absolute path?

We could. We already had a .removal related to this setting for the 3.9.0 release.

https://docs.pulpproject.org/pulpcore/changes.html#removals

mdellweg:
Two (green) approvals; all tests passed; no migration needed; last minute comments addressed
I am going to play merge bot now.

PR #1083, opened 2021-01-21
Fix another scenario where canceled tasks could be marked failed

dralley:
closes: #8143
https://pulp.plan.io/issues/8143

### DIFF
--- a/CHANGES/7980.bugfix
+++ b/CHANGES/7980.bugfix
@@ -1,0 +1,1 @@
+Fixed a scenario where canceled tasks could be marked failed.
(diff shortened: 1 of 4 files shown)


dralley:
Can't be merged until https://github.com/rq/rq/pull/1394 is

daviddavis (approved):

dralley:
@daviddavis Yes, it's the same issue as the one you filed.  I've marked mine as duplicate and changed all the references to your older one since it's more specific. 

@bmbouter @daviddavis I need both of you to re-review.  Unfortunately during the RQ PR review process some of the details of the implementation changed which broke this patch, so I had to re-write it.  It should work now however.

bmbouter (approved):
I can see why you added these changes. Thanks @dralley !

daviddavis (approved):

PR #1084, opened 2021-01-22
Auto publish and distribute

dralley:
This is like 10% finished, I just want to discuss ideas

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -12,6 +12,8 @@ from django.core.validators import MinValueValidator
 from django.db import models, transaction
 from django.urls import reverse
 
+from django.contrib.postgres.fields import JSONField
+
 from pulpcore.app.util import batch_qs, get_view_name_for_model
 from pulpcore.constants import ALL_KNOWN_CONTENT_CHECKSUMS
 from pulpcore.download.factory import DownloaderFactory
@@ -20,8 +22,6 @@ from pulpcore.exceptions import ResourceImmutableError
 from .base import MasterModel, BaseModel
 from .content import Artifact, Content
 from .task import CreatedResource, Task
-
-from django.contrib.postgres.fields import JSONField
 
 
 _logger = logging.getLogger(__name__)
@@ -58,6 +58,16 @@ class Repository(MasterModel):
 
     class Meta:
         verbose_name_plural = "repositories"
+
+    def on_new_version(self, version):
+        """Called when new repository versions are created.
+
+        Subclasses are expected to override this to do useful things.
+
+        Args:
+            version: The new repository version.
+        """
+        pass
 
     def save(self, *args, **kwargs):
         """
@@ -913,6 +923,7 @@ class RepositoryVersion(BaseModel):
                     self.repository.save()
                     self.save()
                     self._compute_counts()
+                    repository.on_new_version(self)
             except Exception:
                 self.delete()
                 raise
--- a/pulpcore/plugin/stages/declarative_version.py
+++ b/pulpcore/plugin/stages/declarative_version.py
@@ -134,6 +134,8 @@ class DeclarativeVersion:
     def create(self):
         """
         Perform the work. This is the long-blocking call where all syncing occurs.
+
+        Returns: The created RepositoryVersion or None if it represents no change from the latest.
         """
         with tempfile.TemporaryDirectory(dir="."):
             with self.repository.new_version() as new_version:
@@ -145,3 +147,5 @@ class DeclarativeVersion:
                 stages.append(EndStage())
                 pipeline = create_pipeline(stages)
                 loop.run_until_complete(pipeline)
+
+        return new_version if new_version.complete else None
--- a/CHANGES/7626.feature
+++ b/CHANGES/7626.feature
@@ -1,0 +1,1 @@
+Add support for automatic publishing and distributing.
--- a/functest_requirements.txt
+++ b/functest_requirements.txt
@@ -3,4 +3,5 @@ dynaconf
 pulp-smash @ git+https://github.com/pulp/pulp-smash.git
 pulpcore-client
 pulp-file-client
+pulp-certguard
 pytest
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -156,7 +156,9 @@ class ArtifactDownloader(Stage):
         #    Set to None if stage is shutdown.
         content_get_task = _add_to_pending(content_iterator.__anext__())
 
-        with ProgressReport(message="Downloading Artifacts", code="downloading.artifacts") as pb:
+        with ProgressReport(
+            message="Downloading Artifacts", code="sync.downloading.artifacts"
+        ) as pb:
             try:
                 while pending:
                     done, pending = await asyncio.wait(pending, return_when=asyncio.FIRST_COMPLETED)
--- a/pulpcore/tests/functional/api/using_plugin/test_distributions.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_distributions.py
@@ -104,7 +104,7 @@ class CRUDPublicationDistributionTestCase(unittest.TestCase):
 
         self.publication = self.client.get(self.publication["pulp_href"])
 
-        # content_guard is the only parameter unset.
+        # content_guard and repository parameters unset.
         for key, val in self.distribution.items():
             if key in ["content_guard", "repository"]:
                 self.assertIsNone(val, self.distribution)


pulpcore/app/models/repository.py
```diff
@@ -51,10 +53,68 @@ class Repository(MasterModel):
         "Content", through="RepositoryContent", related_name="repositories"
     )
     remote = models.ForeignKey("Remote", null=True, on_delete=models.SET_NULL)
+    publish_settings = JSONField(null=True)
 
     class Meta:
         verbose_name_plural = "repositories"
 
+    def mutate(self, closure):
```
dralley, line 61: @bmbouter @daviddavis @dkliban 

So here's the problem.  Repo modification lives in core, everything else is in the plugin.  We need a way for modifications to the repo (no matter where they come from) are able to run code present exclusively in the plugin.

There's two ways to do that.

1) move every kind of task that creates new repository versions into the plugin, and copy-and-paste the implementation between all of them.  Lots of duplicate and ad-hoc code and the implementations can start diverging if we don't pay attention.

2) Find a way to bridge the gap by registering the functions necessary for publishing and distributing somewhere that core can use them, and make it convenient(ish) to combine the triggers w/ the automatic side-effects.

I have no idea which way is better - this is a half-assed version of approach #2 just to demonstrate the idea.

What do you guys think?  Would #1 be better or can you think of a better way?

pulpcore/app/models/repository.py
```diff
@@ -51,10 +53,68 @@ class Repository(MasterModel):
...
+            # need to register publish_func somehow so that it can be used from modify code
+            publish_closure = partial(publish_func, *publish_args)
+            ras.chain(publish_closure)
+
+            if self._should_auto_distribute():
+                # need to register distribute_func somehow so that it can be used from modify code
+                distribute_closure = partial(distribute_func, *distribute_args)
+                ras.chain(distribute_closure)
+
+        return ras.build()
```
dralley, line 110: This returns a closure function that encapsulates all of the steps, and all of the arguments they need to execute.

So in the place where we would currently have this code:

```python
        result = enqueue_with_reservation(
            tasks.publish,
            [repository_version.repository],
            kwargs={"repository_version_pk": str(repository_version.pk), "manifest": str(manifest)},
        )
```

You would instead put this:

```python
        result = enqueue_with_reservation(
            repository.modify(partial(tasks.publish, repository_version.repository, repository_version_pk=str(repository_version.pk), manifest=str(manifest))),
            [],
            kwargs={},
        )
```

dralley: Totally open to suggestions on better ways

pulpcore/app/models/repository.py
dkliban: Seems like number 2 would be more DRY.

dralley: I agree, with the extra bonus that we should only need a couple of lines of changes per plugin to support it.

bmbouter: I agree #2 is likely our best path. If I'm understanding this right, we already do something like that now with the [`finalize_new_version()`](https://github.com/pulp/pulpcore/blob/05faf9f04c43e20a7d00841d1925c2ef5e0363dc/pulpcore/app/models/repository.py#L119-L138). The nice thing about that pattern is a) a base implementation can be provided or not, b) the plugin writer can customize it fully.

@dralley is ^ helpful towards moving forward or maybe I misunderstood?

dralley: ```finalize_new_version()``` is intended to be used in a different way, I would be uncomfortable using it this way.  Actually, I don't think it works regardless, because it's called prior to the "completion" of the repository version.

https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/repository.py#L813

Something *like* that could work though.

bmbouter: Yes I mean add a new one. I was giving `finalize_repo_version()` as an example of precedent for such a design.

Plugin writers technically can create a new RepositoryVersion outside of a task, but they are violating the tasking system's locking requirements when they do so. Ultimately that call comes back to a Repository.method, e.g. `Repository.finalize_repo_version()` so I think it's ok. I design-wise think of the pattern as a controller pattern.

pulpcore/app/models/repository.py
bmbouter: I like this pattern. We should think some about how we want to support chain-ability and what other build-in tools could be used instead of `mutate`. Or maybe we need to diy it, I'm not sure.

dralley: I guess there's "Option 3", make an explicit controller class with methods for sync, modify, publish, and distribute, that takes a repository as an argument, and that object can be used to do all actions that spawn tasks.

Then plugin writers would create a really thin subclass with constants that hold the functions used for sync and publish.

pulpcore/app/models/repository.py
dralley: @bmbouter I've been writing an implementation along the lines of what you described, but what you said about the "controller" pattern gave me the idea to just make it explicit. What if we just had a subclassable "RepoController" that would handle spawning all tasks.

It would probably be a cleaner separation of concerns, and it would basically remove the "tasking system" from the Plugin API, and a lot of code that we currently copy and paste.

Something roughly like this (which I haven't tried yet)

```python
class RepoController:  # defined in core
    SYNC_TASK = None
    PUBLISH_TASK = None

    def __init__(self, repository):
        self.repository = repository

    def sync(self, remote=None, *args, **kwargs):
        remote = remote or self.repository.remote
        result = enqueue_with_reservation(
            self._wrap(self.SYNC_TASK),
            [self.repository, remote],
            args=args,
            kwargs=kwargs,
        )
        return result

    def publish(self, repository_version, *args, **kwargs):
        repository_version = repository_version or self.repository.latest_version()
        result = enqueue_with_reservation(
            self.PUBLISH_TASK,
            [self.repository],
            args=args,
            kwargs=kwargs,
        )

    def modify(self, *args, **kwargs):
        result = enqueue_with_reservation(
            self._wrap(tasks.repository.add_and_remove),
            [repository],
            args=args,
            kwargs=kwargs,
        )

    def _wrap(self, *args, **kwargs):
        def inner(func):

            if self.publish_settings and self.PUBLISH_FUNC:
                self.PUBLISH_FUNC(repository_version_pk=rv.pk, **self.publish_settings)

                if self.distribution:
                    self.distribution.update(repository_version=rv)
        return inner


# defined in plugin
class FileRepoController(RepoController):
    SYNC_TASK = tasks.sync
    PUBLISH_TASK = tasks.publish


class FilePublicationViewSet(PublicationViewSet):
    endpoint_name = "file"
    queryset = FilePublication.objects.exclude(complete=False)
    serializer_class = FilePublicationSerializer

    @extend_schema(
        description="Trigger an asynchronous task to publish file content.",
    )
    def create(self, request):
        """
        Publishes a repository.

        Either the ``repository`` or the ``repository_version`` fields can
        be provided but not both at the same time.
        """
        serializer = self.get_serializer(data=request.data)
        repository_version = serializer.validated_data.get("repository_version")
        manifest = serializer.validated_data.get("manifest")

        result = FileRepoController(repository_version.repository).publish(
            repository_version=repository_version,
            manifest=str(manifest)
        )

        return OperationPostponedResponse(result, request)

```

What do you think? Good idea? Too much?

dralley:
@bmbouter @daviddavis All 3 PRs are ready for some review.  Two things:

* Since plugins need to do some work to make the feature work for them, I'm not sure where to put the docs?  It works a little differently depending on whether the plugin is publication based or not.
* The issue discusses being able to create the publish settings and distributions directly from repository creation endpoint by having nested fields, but I didn't do that. A couple of reasons:
    * Since the link between Distributions and Repositories is on the abstract model / detail model instead of on BaseDistribution, it becomes practically impossible to follow the link backwards, so you need knowledge of the distributor type, which core doesn't currently have.
    * DRF makes it really easy to de-serialize objects in a nested way, but nested creation of is not very easy, unfortunately.  It's supposedly possible but a major pain, and even moreso because of the way we do master-detail.
    * It'd be inconsistent because we're not doing that for Remote.

That's a little sad because it means that you need to make 3 or 4 requests (repository, remote, publish_settings, distribution) instead of just 1 or 2 (repository + publish + distribution, remote).  But I don't know that we can realistically do better than that right now.

dralley:
Not sure what to do about the requirement that causes the test failures.

pulpcore/app/models/publication.py
```diff
@@ -7,6 +7,10 @@
 from pulpcore.app.files import PulpTemporaryUploadedFile
 
 
+class PublishSettings(MasterModel):
+    TYPE = "remote"
```
mdellweg, line 11: This looks wrong to me.
Also this PublishSetting reminds me of the Publisher we once ditched.

dralley: Yes, and yes.  I actually made this same observation https://pulp.plan.io/issues/7469#note-12

dralley: And ```Remote``` would be more properly called ```SyncSettings``` since it contains not just a location but a set of instructions on how to sync.  But there's no changing that now.

pulpcore/app/models/publication.py
```diff
@@ -45,9 +49,11 @@ class Publication(MasterModel):
     pass_through = models.BooleanField(default=False)
 
     repository_version = models.ForeignKey("RepositoryVersion", on_delete=models.CASCADE)
+    publish_settings = models.ForeignKey("PublishSettings", null=True, on_delete=models.PROTECT)
 
+    # TODO(3.13) - make "publish_settings" mandatory
```
daviddavis, line 54: Why 3.13? With a one release deprecation policy, I would have expected 3.12.

pulpcore/plugin/stages/declarative_version.py
```diff
@@ -135,6 +135,8 @@ def pipeline_stages(self, new_version):
     def create(self):
         """
         Perform the work. This is the long-blocking call where all syncing occurs.
+
+        Returns: (bool) Whether a new repository version was created.
```
daviddavis, line 139: Looks like this method returns new_version or None.

pulpcore/app/serializers/publication.py
```diff
@@ -165,27 +184,34 @@ class PublicationDistributionSerializer(BaseDistributionSerializer):
...
         allow_null=True,
     )
-    repository_version = RepositoryVersionRelatedField(
-        required=False, help_text=_("RepositoryVersion to be served"), allow_null=True
-    )
 
     class Meta:
         abstract = True
-        fields = BaseDistributionSerializer.Meta.fields + ("repository", "repository_version")
+        fields = BaseDistributionSerializer.Meta.fields + ("repository_version", "repository")
```
daviddavis, line 214: I am not sure I understand the changes here to `RepositoryVersionDistributionSerializer`. Was there some issue with the ordering?

pulpcore/app/serializers/publication.py
```diff
@@ -165,27 +184,34 @@ class PublicationDistributionSerializer(BaseDistributionSerializer):
         queryset=models.Publication.objects.exclude(complete=False),
         allow_null=True,
     )
+    repository = DetailRelatedField(
+        required=False,
+        help_text=_("The latest RepositoryVersion for this Repository will be served."),
+        view_name_pattern=r"repositories(-.*/.*)?-detail",
+        queryset=models.Repository.objects.all(),
+        allow_null=True,
+    )
```
daviddavis, line 193: I'm not sure I understand how this change relates to the feature?

pulpcore/app/models/repository.py
```diff
@@ -54,10 +54,21 @@ class Repository(MasterModel):
...
 
+    def on_new_version(self, version):
+        """Called when new repository versions are created.
+
+        Subclasses are expected to override this to do useful things.
+
+        Args:
+            version: The new repository version.
+        """
+        pass
```
daviddavis, line 70: :+1:

pulpcore/app/models/publication.py
bmbouter: For these types of things can a task be filed and put onto the correct milestone in redmine?

daviddavis: Yea, I was going to suggest that too. I created the 3.13 milestone already.

pulpcore/app/serializers/publication.py
dralley: No issue, I don't know why I changed it.  Maybe it just got left when I moved things around and then moved them back.

pulpcore/app/serializers/publication.py
dralley: @daviddavis The way you set up auto-distribute is by associating the Repository with the Distribution.  Every distribution that points to the repository gets updated when a new repository version is created in that repository.

pulpcore/plugin/stages/declarative_version.py
dralley: I'll fix that

pulpcore/app/serializers/publication.py
daviddavis: Ah ok. The Github diff seemed to indicate that this was on ContentGuardSerializer.

pulpcore/app/serializers/publication.py
```diff
@@ -165,10 +184,17 @@ class PublicationDistributionSerializer(BaseDistributionSerializer):
...
```
dralley, line 193: @bmbouter @daviddavis  I think I was wrong when I said we could make this backwards compatible.  I don't think we can, even with "terrible hacks".  I spent a while discussing it in #django.

Django's runtime ORM stuff doesn't know about the actual database schema, it only knows about the inheritance hierarchy.  If you have new pulpcore and old plugin, and you add ```repository``` to the ```PublicationDistribution``` abstract model, then the field is now in the Django inheritance machinery but not in the database.  

So if you have new pulpcore installed, and older plugins, this happens...

```
FileDistribution.objects.create(name="foo", base_path="bar/baz")
...
...
ProgrammingError: column "repository_id" of relation "file_filedistribution" does not exist
LINE 1: ...ile_filedistribution" ("basedistribution_ptr_id", "repositor...
```

I really have no idea what to do about this.  Like, the only place to go from here would be introspecting the live schema and trying to construct a raw sql query.  Or completely redesigning the feature to put the ```BaseDistribution``` foreign key on ```Repository``` instead.

People have requested this feature from Django but it's not implemented.  https://code.djangoproject.com/ticket/21454

dralley: We added ```repository``` to ```RepositoryVersionDistribution``` before we ever froze the models so we never ran into this problem.

dralley: I have an idea.

* We push pulpcore 3.11 without these changes.
* We put a pulpcore compatibility cap on all the new plugins releases, 1 or 2 releases from now
* We make the switch then
* In the meantime, we just give the UI team custom packages or point them at the PR branches

I think that would still be obeying the policy, right?

ipanova: what about rhui folks? i had an impression they were waiting on these changes as well with a certain deadline. can we check in with them on this topic?

dralley: @ipanova You're right about that.  I'll check with them on the deadlines.

I notice though that the file, ansible and container plugins already have their versions capped <3.12, which would make it a much easier place to sync everything up.

pulpcore/app/migrations/0061_rename_new_distribution_fields.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 2.2.18 on 2021-02-19 16:02
+
+from django.db import migrations, models
+
+class Migration(migrations.Migration):
```
dralley, line 5: The plugin migrations all run before this one does.

pulpcore/app/serializers/publication.py
mdellweg: > Django's runtime ORM stuff doesn't know about the actual database schema, it only knows about the inheritance hierarchy. If you have new pulpcore and old plugin, the new version of the PublicationDistribution abstract model has an extra field repository, then the field is now present in the Django ORM machinery but not in the database.

Maybe you are kind of giving away the solution here. As you say, ORM only "sees the inheritance hierarchy", and the "new version of [the] abstract model" must have the field. What if we introduce a new intermediary base class called `PublicationDistribution2` which initially inherits `PublicationDistribution` and adds the `repository` field? Plugins must then start to inherit from there (while providing the corresponding migrations). We deprecate the `PublicationDistribution` and on deletion, we move all its members over to the `PublicationDistribution2`. And if we are really keen to do, we can take another 2 pulpcore versions to rename it back (but why?).

pulpcore/app/serializers/publication.py
```diff
@@ -16,6 +16,27 @@
...
+        """Remove the UniqueTogetherValidator so that we can do a get_or_create()."""
+        for validator in self.validators.copy():
+            if isinstance(validator, validators.UniqueTogetherValidator):
+                self.validators.remove(validator)
+        super().run_validators(value)
+
+    def create(self, validated_data):
+        """Override the default create to use get_or_create behavior."""
+        instance, _ = self.Meta.model.objects.get_or_create(**validated_data)
+        return instance
```
dralley, line 37: I'm open to disagreement on the way this works, maybe it's a step too far.

The idea here is that we now effectively have all the publish settings duplicated between *Publication and *PublishSettings objects, which is unfortunate.  One way we could unify everything is by moving the data into PublishSettings exclusively, and attaching the PublishSettings to the Publication.

But that would also lead to a lot of duplicate PublishSettings objects unless we implement get_or_create() semantics, which is why I did this.

But if we just ignore the duplication of fields, it does avoid all of these changes, at the expense of the data model being a little less consistent.

CHANGES/plugin_api/7626.feature
```diff
@@ -0,0 +1 @@
+Added a new model `PublishSettings` which plugin writers are expected to subclass. When dispatching a task to create a `Publication`, use ".get_or_create()" to create a `PublishSettings` object and refer your publish tasks to use it. When creating a `Publication` inside your publish task with `Publication.create()`, you should pass the in the `PublishSettings` object so that it may be associated with the `Publication`. A new method `on_new_version()` should be added to repository models to define behavior that takes place when new repository versions are created. For an example of these changes, please see: https://github.com/pulp/pulp_file/pull/467
```
bmbouter, line 1: Can this line-wrap at 100 chars please?

bmbouter: Also sphinx needs double backticks instead of single for rst formatting

bmbouter: This content is great. We also need this to go into a new section of docs in the plugin writer guide I think. Please add that too.

pulpcore/app/serializers/publication.py
bmbouter: Can you clarify for me: with the code I see in this PR this does have everything live on `PublishSettings` exclusively yes?

dralley: Every plugin went and added fields to their own Publication models like this:  https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/models/repository.py#L310-L325

This is how the user can know what settings were used in the creation of their publication.

Now that we have a PublishSettings model, all of those fields are duplicated in the plugins.

The question is, do we leave it like that, or do we try to move towards just creating a PublishSettings any time you publish (with get_or_create() semantics to avoid spamming them) and including that on the publication.

dralley: Maybe I misread your question.  *Most* of the code related to this lives here, but under this framework one thing that would change is that the publish viewset would always get_or_create() a publish settings object and then pass it into the publish task instead of individual settings.

PR #1086, opened 2021-01-25
Checksum type publication check

pavelpicka:
User receive error when try to publish content with disallowed checksum type.

closes: #7988
https://pulp.plan.io/issues/7988

### DIFF
--- a/CHANGES/7988.feature
+++ b/CHANGES/7988.feature
@@ -1,0 +1,1 @@
+Publication creation will fail when there is content which doesn't contain at least one artifact with allowed checksum.
--- a/pulpcore/app/models/publication.py
+++ b/pulpcore/app/models/publication.py
@@ -1,10 +1,14 @@
 from django.db import IntegrityError, models, transaction
 
+from gettext import gettext as _
+
 from .base import MasterModel, BaseModel
 from .content import Artifact, Content, ContentArtifact
 from .repository import Remote, Repository, RepositoryVersion
 from .task import CreatedResource
+from pulpcore.constants import ALLOWED_CHECKSUM_ERROR_MSG
 from pulpcore.app.files import PulpTemporaryUploadedFile
+from pulpcore.plugin.exceptions import UnsupportedDigestValidationError
 
 
 class Publication(MasterModel):
@@ -69,6 +73,25 @@ class Publication(MasterModel):
             Adds a Task.created_resource for the publication.
         """
         with transaction.atomic():
+            # Check if at least some remote artifacts have allowed checksum
+            for ca in ContentArtifact.objects.filter(
+                content__in=repository_version.content, artifact=None
+            ):
+                allowed_checksum_content = False
+                remote_artifacts = ca.remoteartifact_set.all()
+                for checksum in Artifact.DIGEST_FIELDS:
+                    if any([ra for ra in remote_artifacts if getattr(ra, checksum, None)]):
+                        allowed_checksum_content = True
+                        break
+                if not allowed_checksum_content:
+                    raise UnsupportedDigestValidationError(
+                        _(
+                            'Cannot publish repository version "{}" because any remote artifact '
+                            "of {} doesn't contain allowed checksum type. {}".format(
+                                repository_version.pk, ca.content, ALLOWED_CHECKSUM_ERROR_MSG
+                            )
+                        )
+                    )
             publication = cls(pass_through=pass_through, repository_version=repository_version)
             publication.save()
             resource = CreatedResource(content_object=publication)
--- a/pulpcore/constants.py
+++ b/pulpcore/constants.py
@@ -1,3 +1,4 @@
+from gettext import gettext as _
 from types import SimpleNamespace
 
 #: All valid task states.
@@ -49,3 +50,8 @@ API_ROOT = "pulp/api/v3/"
 # adding here won't make a new type-of checksum available. sha256 MUST be here,
 # as Pulp relies on it to identify entities.
 ALL_KNOWN_CONTENT_CHECKSUMS = {"md5", "sha1", "sha224", "sha256", "sha384", "sha512"}
+
+# Shared part of error message for user which setting to use
+ALLOWED_CHECKSUM_ERROR_MSG = _(
+    "You can add allowed checksums with extending the 'ALLOWED_CONTENT_CHECKSUMS' setting."
+)
--- a/pulpcore/tests/unit/models/test_publication.py
+++ b/pulpcore/tests/unit/models/test_publication.py
@@ -1,0 +1,108 @@
+from django.test import TestCase
+from pulpcore.plugin.exceptions import UnsupportedDigestValidationError
+from pulpcore.plugin.models import (
+    Content,
+    ContentArtifact,
+    CreatedResource,
+    Publication,
+    Remote,
+    RemoteArtifact,
+    Repository,
+)
+from mock import patch
+
+
+class PublicationTestCase(TestCase):
+    def setUp(self):
+        self.repository = Repository.objects.create()
+        self.repository.CONTENT_TYPES = [Content]
+        self.repository.save()
+        self.remote = Remote.objects.create()
+        self.remote.url = "www.pulpproject.org"
+        self.remote.name = "testremote"
+        self.remote.save()
+        self.secondremote = Remote.objects.create()
+        self.secondremote.url = "example.pulpproject.org"
+        self.secondremote.name = "exampleremote"
+        self.secondremote.save()
+
+    @patch.object(CreatedResource, "save")
+    def test_publication_create_all_remoteartifact_supported_checksum_type(self, save):
+        save.return_value = True
+        contents = []
+        cntartifact = []
+        for _ in range(2):
+            cnt = Content.objects.create(pulp_type="core.content")
+            ca = ContentArtifact.objects.create(content=cnt)
+            contents.append(cnt)
+            cntartifact.append(ca)
+
+        for ca in cntartifact:
+            RemoteArtifact.objects.create(
+                url="http://example.pulpproject.org/ra",
+                remote=self.remote,
+                content_artifact=ca,
+                sha256="3dfe13d3b8264db6bc66391af110b9fe932e0b9d170d160fbd81b9a5cdfd3040",
+            )
+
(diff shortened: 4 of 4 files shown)


pulpcore/app/models/publication.py
```diff
@@ -69,6 +69,14 @@ def create(cls, repository_version, pass_through=False):
             Adds a Task.created_resource for the publication.
         """
         with transaction.atomic():
+            for artifact in repository_version.artifacts:
+                for checksum in Artifact.DIGEST_FIELDS:
+                    checksum_type = getattr(artifact, checksum)
+            if not checksum_type:
```
ipanova, line 75: should not this be indented so it's inside of the for loop?

ipanova: how is it possible to have artifacts with missing checksums if pulp would not start in the first place? am i missing something?

daviddavis: I only think we need a check for remote artifacts since [we still allow pulp to still start](https://pulp.plan.io/issues/7985).

ipanova: agreed on that

pavelpicka: @ipanova @daviddavis logic fixed and only remote artifacts are check before publish

pavelpicka:
Introduced part of error message we can re-use to point user where to update setting about checksum types.

pulpcore/app/models/publication.py
```diff
@@ -69,6 +70,26 @@ def create(cls, repository_version, pass_through=False):
...
+                    if checksum_type:
+                        break
+                if not checksum_type:
+                    raise ValueError(
+                        "Repository version contains remote artifact "
+                        "({}) with unsupported checksum type, "
+                        "thus can't be pulbished. {}".format(
+                            artifact.url, ALLOWED_CHECKSUM_ERROR_MSG
+                        )
+                    )
```
dralley, line 92: @pavelpicka Shouldn't the criteria be that at least one ```RemoteArtifact``` for an artifact needs to have an allowed checksum rather than that all ```RemoteArtifacts``` do?

pavelpicka: @dralley We need all content have allowed checksum to avoid situations when we won't be able to work with forbidden checksums algorithm/libs (e.g. forbidden by os)

mdellweg: Will all checksums from all RemoteArtifacts be verified once the Artifact is downloaded? In that case, "at least one" sounds suficient.

daviddavis: I think the code here is generally correct although it doesn't match the error message.

The one thing it doesn't take into account is that there could be multiple RemoteArtifacts for a ContentArtifact and one of them could have an allowed checksum that's acceptable. 

I think we need something like this:

```python
for content in repository_version.content:
    for ca in content.contentartifact_set.all():
        if ca.artifact:
            continue  # not on demand
        else:
            remote_artifacts = ca.remoteartifact_set.all()
            for checksum in Artifact.DIGEST_FIELDS:
                if not any(ra for ra in remote_artifacts if getattr(ra, checksum)):
                    raise ValueError("...")
```

daviddavis:
I'd highly encourage some testing on this PR since the logic is complicated. You'll probably have to write a unit test.

pulpcore/app/models/publication.py
ipanova: @pavelpicka we had a short discussion yesterday at the pulpcore team meeting regarding this PR and came to a conclusion that it would make sense to deny a publication only in case when not a single remote artifact from the content artifact would have allowed checksum.
Will you be able to update your PR given this input?

pavelpicka: @ipanova @daviddavis Now the logic check if at least one remote artifact has allowed checksum.

CHANGES/7988.feature
```diff
@@ -0,0 +1 @@
+User will get error when try publish repository version with content not allowed in 'ALLOWED_CONTENT_CHECKSUMS' setting.
```
ipanova, line 1: Can you update this entry?
"Deny publication when...'

pulpcore/app/models/publication.py
```diff
@@ -69,6 +70,29 @@ def create(cls, repository_version, pass_through=False):
...
+            for ca in ContentArtifact.objects.filter(
+                content__in=repository_version.content, artifact=None
+            ):
+                allowed_checksum_content = False
+                remote_artifacts = ca.remoteartifact_set.all()
+                for checksum in Artifact.DIGEST_FIELDS:
+                    if any([ra for ra in remote_artifacts if getattr(ra, checksum, None)]):
+                        allowed_checksum_content = True
+                        break
+                # We don't need to continue, allowed checksum type found
```
ipanova, line 83: we do need to continue here because we need to perform this check for every CA within the repo_version if you are going to break on line 85 you will end up checking only first CA that the repo version contains

ipanova: we should instead check the value of `allowed_checksum_content` when we're out of the forloop, false value will point that none of the remote artifacts from the ca contains allowed checksum and we should raise an error.

pulpcore/app/models/publication.py
```diff
@@ -69,6 +70,29 @@ def create(cls, repository_version, pass_through=False):
...
+                if allowed_checksum_content:
+                    break
+            else:
+                # In case of no remote artifact to check, ensure exception won't be raised
```
daviddavis, line 87: Given that we are looping through content artifacts with no artifacts, this would mean that the content artifact has no artifact or remote artifacts. This case should probably raise an exception. I'd probably just raise the ValueError you have below by removing these lines.

pulpcore/app/models/publication.py
```diff
@@ -69,6 +70,29 @@ def create(cls, repository_version, pass_through=False):
...
+                allowed_checksum_content = True
+            if not allowed_checksum_content:
+                raise ValueError(
+                    'Cannot publish repository version "{}" '
+                    "because none of the content has allowed checksum. {}".format(
```
daviddavis, line 92:
```suggestion
                    "because some of the content cannot be published due to missing available checksums. {}".format(
```

daviddavis: Also, this needs to use gettext.

pulpcore/constants.py
```diff
@@ -49,3 +49,7 @@
 # adding here won't make a new type-of checksum available. sha256 MUST be here,
 # as Pulp relies on it to identify entities.
 ALL_KNOWN_CONTENT_CHECKSUMS = {"md5", "sha1", "sha224", "sha256", "sha384", "sha512"}
+
+# Shared part of error message for user which setting to use
+ALLOWED_CHECKSUM_ERROR_MSG = """You can add allowed checksums with extending
+ the 'ALLOWED_CONTENT_CHECKSUMS' setting."""
```
daviddavis, line 55: This needs i18n/gettext too.

pavelpicka:
@bmbouter, @daviddavis, @ipanova PR back to ready for re-review

pulpcore/app/models/publication.py
```diff
@@ -69,6 +72,25 @@ def create(cls, repository_version, pass_through=False):
...
+                content__in=repository_version.content, artifact=None
+            ):
+                allowed_checksum_content = False
+                remote_artifacts = ca.remoteartifact_set.all()
+                for checksum in Artifact.DIGEST_FIELDS:
+                    if any([ra for ra in remote_artifacts if getattr(ra, checksum, None)]):
+                        allowed_checksum_content = True
+                        break
+                if not allowed_checksum_content:
+                    raise ValueError(
```
daviddavis, line 86: I found out that there's a specific exception for unrecognized / forbidden checksums:
```suggestion
                    raise UnsupportedDigestValidationError(
```

pulpcore/tests/unit/models/test_publish.py
```diff
@@ -0,0 +1,107 @@
+from django.test import TestCase
```
daviddavis, line 1: I would probably call this test `test_publication`. Publish seems too specific and technically we don't publish (we create publications).

pulpcore/constants.py
```diff
@@ -49,3 +50,9 @@
 # adding here won't make a new type-of checksum available. sha256 MUST be here,
 # as Pulp relies on it to identify entities.
 ALL_KNOWN_CONTENT_CHECKSUMS = {"md5", "sha1", "sha224", "sha256", "sha384", "sha512"}
+
+# Shared part of error message for user which setting to use
+ALLOWED_CHECKSUM_ERROR_MSG = _(
+    """You can add allowed checksums with extending
+ the 'ALLOWED_CONTENT_CHECKSUMS' setting."""
```
daviddavis, line 57: Does this need to be split on two lines?

daviddavis:
Three small comments. Other than that, I think this is ready to merge.

CHANGES/7988.feature
```diff
@@ -0,0 +1 @@
+Publication creation will fail when there is content which doesn't contain at least one artifact with allowed checksum.
```
ipanova, line 1: @daviddavis when combining with this PR https://github.com/pulp/pulpcore/pull/1156 do i read correctly that publish will succeed with 1 forbidden( for example md5)  and  1allowed checksum ( sha256) but when we get to streaming the artifact it will fail in any case because that remote artifact has forbidden checksum?

daviddavis: Yes, exactly. The only thing I would mention is that it should continue to the next remote artifact and try to stream it if the first one fails.

ipanova (approved):

daviddavis:
There is one edge case I thought of. Do plugins ever create RemoteArtifacts with no checksums? Looks like it's possible and I believe that certain plugins would have to if they do not have checksums in the metadata. If this is the case, then this code wouldn't work properly as you could never create a publication for these RemoteArtifacts.

If I am correct here, then I think we should probably just reject any publication that would have a RemoteArtifact with a forbidden checksum.

mdellweg:
Rubygems do not have checksums. (Last time i checked.)

daviddavis:
In our FIPS meeting today, we decided to postpone this work indefinitely (until a user requests it or we deem it necessary) and instead just rely on rejecting on-demand content during sync and in the content app before saving it.

PR #1088, opened 2021-01-26
Add repo attribute check to global access conditions

goosemania:
Plugins can leverage it at least in policies for repository version viewsets.

closes #8161
https://pulp.plan.io/issues/8161


bmbouter (approved):
This looks good to me. There are so many docstrings and such little code. I think that is a great thing actually.

Thank you for contributing this!

ipanova (approved):

bmbouter:
I see two lgtm's, merging!

PR #1089, opened 2021-01-26
Added clarifications around backport requests.

goosemania:
[noissue]

### DIFF
--- a/docs/bugs-features.rst
+++ b/docs/bugs-features.rst
@@ -86,11 +86,13 @@ from the filer, so please :ref:`reach out<community>` with a link to your issue.
 For **Backport Requests**
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Only bug fixes from newer versions can be requested to be backported. No features can be backported
-to older versions.
+Only bug fixes from newer released or unreleased versions can be requested to be backported.
+No features can be backported to older versions.
 
 Django migrations are order dependent so we cannot backport any change that has database migrations.
-Thus z-stream releases will not add new migrations.
+Thus Z-stream releases will not add new migrations.
+
+No Z releases are planned by default. Backport requests trigger Z release planning.
 
 Please include:
 


docs/bugs-features.rst
```diff
@@ -86,11 +86,13 @@ from the filer, so please :ref:`reach out<community>` with a link to your issue.
...
-Only bug fixes from newer versions can be requested to be backported. No features can be backported
-to older versions.
+Only bug fixes from newer released or unreleased versions can be requested to be backported.
+No features can be backported to older versions.
 
 Django migrations are order dependent so we cannot backport any change that has database migrations.
-Thus z-stream releases will not add new migrations.
+Thus Z-stream releases will not add new migrations.
+
+No Z releases are planned by default. Backport requests trigger Z release planning.
```
ipanova, line 95: It is worth re-phrasing and mention that bugfixes also trigger Z releases?
I am concerned it's not read as only backports trigger Z releases.

goosemania: That's exactly my take away from the open floor discussion and which was a bit surprising to me.
There are no planned  Z releases and any Z release works with backport requests, since we cherry-pick any fix from a master branch. Folks said that any cherry-pick action requires a backport request.

mdellweg: Still a question might be if every backport leads to one Z release. I think we plan to deliver them in batches if possible. (That is multiple backports in on Z release)

goosemania: IMO, "Backport requests trigger Z release planning", implies that it's not one release per request. `Requests` are in plural and they trigger `planning`. I'm open to phrasing suggestions, if it's not clear that it is not one release per request.

ipanova: @goosemania thanks for sharing the background and clarification.
I still think we would need to have a Z release no matter what in case, for example, a regression has been found. It will just be one more step, most likely for us, to copy the original reported issue and open a backport. I don't foresee the user opening both issue+backport, they already struggle with finding how and where to open pulp bugs.

ipanova (approved):

docs/bugs-features.rst
daviddavis: I hear the concern about not automatically backporting bugfixes to the latest z-stream without a backport release. I'm hoping though that once we have automated our release process, we can consider changing our release processes.

daviddavis (approved):

bmbouter (approved):
This looks good to me.

PR #1090, opened 2021-01-26
Added import-check command, tests, and doc.

ggainey:
closes #7549

### DIFF
--- a/docs/workflows/import-export.rst
+++ b/docs/workflows/import-export.rst
@@ -364,3 +364,30 @@ The command to create an import will return a task that can be used to monitor t
 also see a history of past imports::
 
     http :/pulp/api/v3/importers/core/pulp/f8acba87-0250-4640-b56b-c92597d344b7/imports/
+
+Pre-validating import parameters
+--------------------------------
+
+There are a number of things that can keep an import from being successful, ranging from a specified
+export-file not being available to bad JSON specified for ``repo_mapping``. You can pre-validate your
+proposed import using the ``import-check`` command::
+
+    http POST :/pulp/api/v3/importers/core/pulp/import-check/ \
+      path=/tmp/export-file-path toc=/tmp/export-toc-path repo_mapping:="{\"source\": \"dest\"}"
+
+``import-check`` will validate that:
+
+    * paths are in ``ALLOWED_IMPORT_PATHS``
+    * containing directory exists
+    * containing directory is readable
+    * path/toc file(s) exist and are readable
+    * for TOC, containing directory is writeable
+    * repo_mapping is valid JSON
+
+``import-check`` is a low-overhead synchronous call. It does not attempt to do validations that
+require database access or long-running tasks such as verifying checksums. All parameters are optional. 
+
+.. note::
+
+    For ``path`` and ``toc``, if the ALLOWED_IMPORT_PATHS check fails, no further information will be given.
+
--- a/pulpcore/app/serializers/__init__.py
+++ b/pulpcore/app/serializers/__init__.py
@@ -41,16 +41,19 @@ from .content import (  # noqa
     SingleArtifactContentSerializer,
 )
 from .exporter import (  # noqa
-    ExportSerializer,
     ExporterSerializer,
+    ExportSerializer,
     FilesystemExporterSerializer,
     PublicationExportSerializer,
     PulpExporterSerializer,
     PulpExportSerializer,
 )
 from .importer import (  # noqa
-    ImportSerializer,
+    EvaluationSerializer,
     ImporterSerializer,
+    ImportSerializer,
+    PulpImportCheckResponseSerializer,
+    PulpImportCheckSerializer,
     PulpImporterSerializer,
     PulpImportSerializer,
 )
@@ -83,5 +86,4 @@ from .upload import (  # noqa
     UploadSerializer,
     UploadDetailSerializer,
 )
-
 from .user import GroupSerializer, UserSerializer  # noqa
--- a/pulpcore/app/serializers/importer.py
+++ b/pulpcore/app/serializers/importer.py
@@ -10,6 +10,7 @@ from pulpcore.app.serializers import (
     ImportIdentityField,
     ModelSerializer,
     RelatedField,
+    ValidateFieldsMixin,
 )
 
 
@@ -131,21 +132,16 @@ class PulpImportSerializer(ModelSerializer):
         """
         Check validity of provided 'toc' parameter.
 
-        'toc' must:
-          * be within ALLOWED_IMPORT_PATHS.
-          * be valid JSON
-          * point to chunked-export-files that exist 'next to' the 'toc' file
+        'toc' must be within ALLOWED_IMPORT_PATHS.
 
-        NOTE: this method does NOT validate checksums of the chunked-export-files. That
+        NOTE: this method does NOT validate existence/sanity of export-files. That
         happens asynchronously, due to time/responsiveness constraints.
 
         Args:
             value (str): The user-provided toc-file-path to be validated.
 
         Raises:
-            ValidationError: When toc is not in the ALLOWED_IMPORT_PATHS setting,
-            toc is not a valid JSON table-of-contents file, or when toc points to
-            chunked-export-files that can't be found in the same directory as the toc-file.
+            ValidationError: When toc is not in the ALLOWED_IMPORT_PATHS setting
 
         Returns:
             The validated value.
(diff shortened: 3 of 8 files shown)


pulpcore/tests/functional/api/using_plugin/test_pulpimport.py
```diff
@@ -236,3 +298,118 @@ def test_chunked_import(self):
...
+                for s in result.toc.messages
+            )
+        )
+        # FAILS IN CI, passes locally
+        # self.assertTrue(
+        #     any(
+        #         "directory /tmp/importcheck/nowritedir must allow pulp write-access" in s
+        #         for s in result.toc.messages
+        #     )
+        # )
```
gerrod3, line 412: You know why it doesn't work in the CI? I also got the test to pass locally.

pulpcore/tests/functional/api/using_plugin/test_pulpimport.py
```diff
@@ -236,3 +298,118 @@ def test_chunked_import(self):
...
+        result = self.import_check_api.pulp_import_check_post(body)
+        self.assertEqual(result.path.context, "/tmp/idonotexist")
+        self.assertFalse(result.path.is_valid)
+        self.assertTrue(
+            any("file /tmp/idonotexist does not exist" in s for s in result.path.messages)
+        )
+
+        body = {"toc": "/tmp/idonotexist"}
+        result = self.import_check_api.pulp_import_check_post(body)
+        self.assertEquals(result.toc.context, "/tmp/idonotexist")
```
gerrod3, line 361: Pytest says assertEquals is deprecated.
```suggestion
        self.assertEqual(result.toc.context, "/tmp/idonotexist")
```

pulpcore/tests/functional/api/using_plugin/test_pulpimport.py
```diff
@@ -236,3 +298,118 @@ def test_chunked_import(self):
...
+        body = {"repo_mapping": '{"foo": "bar"'}
+        result = self.import_check_api.pulp_import_check_post(body)
+        self.assertEqual(result.repo_mapping.context, '{"foo": "bar"')
+        self.assertFalse(result.repo_mapping.is_valid)
+        self.assertEqual(result.repo_mapping.messages[0], "invalid JSON")
+
+    def test_import_check_not_allowed(self):
+        body = {"path": "/notinallowedimports"}
+        result = self.import_check_api.pulp_import_check_post(body)
+        self.assertEquals(result.path.context, "/notinallowedimports")
```
gerrod3, line 338: Pytest says assertEquals is deprecated.
```suggestion
        self.assertEqual(result.path.context, "/notinallowedimports")
```

gerrod3 (commented):
Looks good. BTW is the path argument always suppose to point to a tarball when specified? I pointed it to the toc file and the command returned with valid. Should there be a check that the file types for toc and path are json and tar.gz?

pulpcore/tests/functional/api/using_plugin/test_pulpimport.py
```diff
@@ -236,3 +298,118 @@ def test_chunked_import(self):
...
+        }
+        result = self.import_check_api.pulp_import_check_post(body)
+
+        self.assertFalse(result.path.is_valid)
+        self.assertEqual(len(result.path.messages), 1, "Only not-allowed should be returned")
+        self.assertEqual(result.path.messages[0], "/ is not an allowed import path")
+
+        self.assertFalse(result.toc.is_valid)
+        print(result.toc.messages)
+        self.assertTrue(
```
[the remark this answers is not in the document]
ggainey: Indeed! will rmv

pulpcore/app/viewsets/importer.py
```diff
@@ -1,5 +1,6 @@
 from django.http import Http404
 from django_filters.rest_framework import filters
+
 from drf_spectacular.utils import extend_schema
 from rest_framework import mixins
```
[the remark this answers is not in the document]
ggainey: That...is a good catch. No, it's a leftover when this was going to be a ViewSet and there was more code here. Good, that means this file doesn't need to be touched at all.

pulpcore/tests/functional/api/using_plugin/test_pulpimport.py
ggainey: Yeah, looks like I missed a couple of those - been using assertEquals() for too long.

pulpcore/tests/functional/api/using_plugin/test_pulpimport.py
ggainey: I haven't determined why the ci-task thinks that it has write-access to a directory with 755 perms. I only commented this out after four different attempts, decided not to block on it.

ggainey:
> Looks good. BTW is the path argument always suppose to point to a tarball when specified? I pointed it to the toc file and the command returned with valid. Should there be a check that the file types for toc and path are json and tar.gz?

I can't rely on the filename, because users can call the files whatever they want. And I don't want to add heuristics to open the files to look inside, this is supposed to be a top-level sanity-check - are the files in a location we're allowed to read from? do they exist? can I read them? are the directory-permissions correct (where possible)? Anything more than that, will be found when one actually does the import (and there are A LOT more checks there)

pulpcore/tests/functional/api/using_plugin/test_pulpimport.py
daviddavis: You could use tmate to debug: https://github.com/marketplace/actions/debugging-with-tmate

Just add this to the end of [the ci workflow](https://github.com/pulp/pulpcore/blob/master/.github/workflows/ci.yml) and uncomment this test:

```yaml
    - name: Setup tmate session
      uses: mxschmitt/action-tmate@v3
      if: failure()
```

daviddavis (approved):
LGTM.

docs/workflows/import-export.rst
```diff
@@ -364,3 +364,30 @@ The command to create an import will return a task that can be used to monitor t
...
+
+    * paths are in ``ALLOWED_IMPORT_PATHS``
+    * containing directory exists
+    * containing directory is readable
+    * path/toc file(s) exist and are readable
+    * for TOC, containing directory is writeable
+    * repo_mapping is valid JSON
+
+``import-check`` is a low-overhead synchronous call. It does make any attempt to do validations that
+require database access or long-running tasks such as verifying checksums. All parameters are optional.
```
goosemania, line 388: I'm not sure I fully follow the sentence. Did you mean "it does make no attempt", or "it does not make any attempt" or something like that?

pulpcore/app/serializers/__init__.py
```diff
@@ -53,6 +53,9 @@
     ImporterSerializer,
     PulpImporterSerializer,
     PulpImportSerializer,
+    PulpImportCheckSerializer,
+    PulpImportCheckResponseSerializer,
+    EvaluationSerializer,
```
goosemania, line 58: Alphabetical order, please

pulpcore/app/serializers/importer.py
```diff
@@ -170,3 +166,76 @@ class Meta:
...
+    create an importer.
+    """
+
+    path = serializers.CharField(
+        help_text=_("Path to export-tar-gz that will be imported."), required=False
+    )
+    toc = serializers.CharField(
+        help_text=_(
+            "Path to a table-of-contents file describing chunks to be validated, "
+            + "reassembled, and imported."
```
goosemania, line 221: Not important but I don't think you need `+` here

pulpcore/app/views/importer.py
```diff
@@ -0,0 +1,124 @@
+from gettext import gettext as _
+import json
+import os
+from drf_spectacular.utils import extend_schema
+from rest_framework.views import APIView
+from rest_framework.response import Response
+
+from pulpcore.app import settings
+from pulpcore.app.serializers import PulpImportCheckSerializer, PulpImportCheckResponseSerializer
```
goosemania, line 9: mega nitpick, in theory those imports should be in different order ;)

pulpcore/tests/functional/api/using_plugin/test_pulpimport.py
```diff
@@ -25,6 +26,7 @@
     ExportersCoreExportsApi,
     ImportersPulpApi,
     ImportersCoreImportsApi,
+    ImportersCoreImportCheckApi,
```
goosemania, line 29: I don't know why I notice those order things, don't ask me :)
One is not from your PR, to be fair

goosemania (approved):
Thanks for this feature!
I added just some minor comments.  It's up to you, whether to address them or not.

docs/workflows/import-export.rst
ggainey: great catch, fixed

pulpcore/app/serializers/importer.py
ggainey: Embarassing, my Java history shining thru...

pulpcore/app/serializers/__init__.py
ggainey: reordered a number of imports, just for you :)

PR #1093, opened 2021-01-28
Allow anonymous access to the api-doc endpoints

mdellweg:


daviddavis:
Is this broken in =< 3.9? If so, it might be worthwhile to have a changelog entry.

mdellweg:
> Is this broken in <= 3.9? If so, it might be worthwhile to have a changelog entry.

It works on my current sandbox machine.
```json
        {
            "component": "pulpcore",
            "version": "3.9.0"
        }
```

daviddavis (approved):

dkliban (approved):

PR #1094, opened 2021-01-28
Add RBAC to the group endpoints

mdellweg:


mdellweg:
As the group model is not defined in pulpcore, i'm unsure if the auto assignment of permissions works there.
The next question is, whether we want auto permissions on those.

This pr is draft, because i want to add the group users rbac to it.

dkliban (approved):
I tested this using pulp_container functional tests that use these APIs to list groups and assign users to groups. This works great! 

When we add the APIs for managing users, we should add some functional tests to pulpcore that exercise these access policies.

CHANGES/8159.bugfix
```diff
@@ -0,0 +1 @@
+Added RBAC to the endpoint for managing groups.
```
bmbouter, line 1: I think this should be a feature instead of a bugfix.

CHANGES/8160.bugfix
```diff
@@ -0,0 +1 @@
+Added RBAC to the endpoint for managing group users.
```
bmbouter, line 1: I think this should be a feature instead of a bugfix.

CHANGES/8159.bugfix
mdellweg: The ticket is a "issue", but we can change that, right?

bmbouter: Yes just switch to story and maybe retitle it to `As a user, ....`

pulpcore/app/access_policy.py
```diff
@@ -44,3 +45,15 @@ def get_policy_statements(self, request, view):
...
             )
         return access_policy_obj.statements
+
+
+class GroupRelationAccessPolicy(AccessPolicyFromDB):
+    """
+    An AccessPolicy for relations on groups that can refer to the groups permissions.
+    """
+
+    def has_group_perms(self, request, view, action, permission):
```
bmbouter, line 55: I think we should consider splitting this into two access checks and also follow the pattern using in the global access checks, for example the [recent additions there](https://github.com/pulp/pulpcore/blob/master/pulpcore/app/global_access_conditions.py#L195-L280).

This handles condition check as-is handles both model-level and user-level permissions. That's similar to a `has_repo_attr_model_or_obj_perms` so this could probably be called `has_group_attr_model_or_obj_perms`. What we aren't providing is a `has_group_attr_obj_perms`. While not used directly in default policies we should provide both I think because users may want to reconfigure a system with a different policy that disallows user-level permission maybe.

Also maybe this should just live in the global access checks? I'm not totally sure because I can't think of it being reused par-se, but it's an idea.

@mdellweg and @dkliban what do you think about this approach?

Also whatever it becomes or wherever it lives, it needs a docstring please.

bmbouter (changes requested):
This PR is legitttttt. I am asking about where to put that condition check and possibly breaking it into two.

Either way it needs a docstring.

pulpcore/app/access_policy.py
dkliban: +1 to splitting it into two checks

pulpcore/app/global_access_conditions.py
```diff
@@ -278,3 +280,67 @@ def has_repo_attr_model_or_obj_perms(request, view, action, permission):
     if has_repo_attr_obj_perms(request, view, action, permission):
         return True
     return False
+
+
+# `Group` permission checks
+
+
+def has_group_obj_perms(request, view, action, permission):
```
bmbouter, line 288: This all looks great. My only question is: do we want to move these back to the GroupAccessPolicy? If they can only be used there I don't think they should be in the global area.

Apologies if I introduced this extra work.

mdellweg: The same holds true for the ones for repos above. And having them here means there is no special GroupAccessPolicy.

bmbouter: Does it? I imagine there are multiple viewsets that will take a remote as a parameter, or a repo version as a parameter so that makes putting there here valuable to multiple viewsets. Will (or could there be) multiple viewsets that pass group_pk as an argument? If there aren't then it doesn't make sense here. If there are then it does. I don't know if there are or not. Wdyt?

mdellweg: GroupUsers, GroupObjectPermissions, GroupModelPermissions come to my mind.

bmbouter (approved):
Thank you!

PR #1098, opened 2021-01-29
Add filtering to access policies

mdellweg:
Also make viewset_name readonly.

### DIFF
--- a/CHANGES/8185.bugfix
+++ b/CHANGES/8185.bugfix
@@ -1,0 +1,1 @@
+Made the viewset_name property of access policies read only.
--- a/pulpcore/app/serializers/access_policy.py
+++ b/pulpcore/app/serializers/access_policy.py
@@ -27,6 +27,7 @@ class AccessPolicySerializer(ModelSerializer):
     viewset_name = serializers.CharField(
         help_text=_("The name of ViewSet this AccessPolicy authorizes."),
         validators=[UniqueValidator(queryset=models.AccessPolicy.objects.all())],
+        read_only=True,
     )
 
     class Meta:
--- a/CHANGES/8189.feature
+++ b/CHANGES/8189.feature
@@ -1,0 +1,1 @@
+Added filtering for access policies.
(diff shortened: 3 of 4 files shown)


CHANGES/8185.bugfix
```diff
@@ -0,0 +1 @@
+Make the viewset_name property of access policies read only and add filtering for it.
```
bmbouter, line 1: This is new functionality, so I think it's `.feature`.

pulpcore/app/serializers/access_policy.py
```diff
@@ -27,6 +27,7 @@ class AccessPolicySerializer(ModelSerializer):
     viewset_name = serializers.CharField(
         help_text=_("The name of ViewSet this AccessPolicy authorizes."),
         validators=[UniqueValidator(queryset=models.AccessPolicy.objects.all())],
+        read_only=True,
```
bmbouter, line 30: Is this a bugfix?

bmbouter: If so maybe break out into its own issue + PR?

CHANGES/8185.bugfix
mdellweg: I consider the lack of filterability a bug. But i broke it into two issues.

CHANGES/8189.bugfix
```diff
@@ -0,0 +1 @@
+Added filtering for access policies.
```
ipanova, line 1: I would lean towards having this as .feature as well.

ipanova (approved):

CHANGES/8189.bugfix
mdellweg: It's not a bug, it's a feature now. :grinning:

CHANGES/8185.bugfix
```diff
@@ -0,0 +1 @@
+Made the viewset_name property of access policies read only.
```
goosemania, line 1: :+1:

goosemania (approved):
Thanks! More things are now in order.

bmbouter (approved):
Thank you!

PR #1100, opened 2021-01-31
Fix repo attribute check to work with the detail model

goosemania:
re #8161
https://pulp.plan.io/issues/8161


mdellweg (approved):
Looks reasonable to me.

ipanova (approved):

PR #1102, opened 2021-02-01
Switches plugin reporting to use Django name

bmbouter:
The Status API and the OpenAPI schema API now both use the Django name.
This is a backwards incompatible change, but necessary to fix the bug
which was using Python package names incorrectly instead.

This also reapplies the plugin template changes required from the PR
below:

https://github.com/pulp/plugin_template/pull/341/files

closes #8198

### DIFF
--- a/.github/workflows/scripts/before_install.sh
+++ b/.github/workflows/scripts/before_install.sh
@@ -34,7 +34,8 @@ else
 fi
 mkdir .ci/ansible/vars || true
 echo "---" > .ci/ansible/vars/main.yaml
-echo "component_name: pulpcore" >> .ci/ansible/vars/main.yaml
+echo "legacy_component_name: pulpcore" >> .ci/ansible/vars/main.yaml
+echo "component_name: core" >> .ci/ansible/vars/main.yaml
 echo "component_version: '${COMPONENT_VERSION}'" >> .ci/ansible/vars/main.yaml
 
 export PRE_BEFORE_INSTALL=$PWD/.github/workflows/scripts/pre_before_install.sh
--- a/pulpcore/app/views/status.py
+++ b/pulpcore/app/views/status.py
@@ -5,14 +5,13 @@ from gettext import gettext as _
 from django.conf import settings
 from django.core.files.storage import default_storage
 from drf_spectacular.utils import extend_schema
-from pkg_resources import get_distribution
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
+from pulpcore.app.apps import pulp_plugin_configs
 from pulpcore.app.models.status import ContentAppStatus
 from pulpcore.app.models.task import Worker
 from pulpcore.app.serializers.status import StatusSerializer
-from pulpcore.app.settings import INSTALLED_PULP_PLUGINS
 from pulpcore.tasking.connection import get_redis_connection
 
 _logger = logging.getLogger(__name__)
@@ -42,16 +41,22 @@ class StatusView(APIView):
         operation_id="status_read",
         responses={200: StatusSerializer},
     )
-    def get(self, request, format=None):
+    def get(self, request):
         """
-        Returns app information including the version of pulpcore and loaded pulp plugins,
-        known workers, database connection status, and messaging connection status
+        Returns status and app information about Pulp.
+
+        Information includes:
+         * version of pulpcore and loaded pulp plugins
+         * known workers
+         * known content apps
+         * database connection status
+         * redis connection status
+         * disk usage information
         """
-        components = ["pulpcore"] + INSTALLED_PULP_PLUGINS
-        versions = [
-            {"component": component, "version": get_distribution(component).version}
-            for component in components
-        ]
+        versions = []
+        for app in pulp_plugin_configs():
+            versions.append({"component": app.label, "version": app.version})
+
         redis_status = {"connected": self._get_redis_conn_status()}
         db_status = {"connected": self._get_db_conn_status()}
 
--- a/pulpcore/openapi/__init__.py
+++ b/pulpcore/openapi/__init__.py
@@ -18,11 +18,10 @@ from drf_spectacular.plumbing import (
 from drf_spectacular.settings import spectacular_settings
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiParameter
-from pkg_resources import get_distribution
 from rest_framework import mixins, serializers
 from rest_framework.schemas.utils import get_pk_description
 
-from pulpcore.app.settings import INSTALLED_PULP_PLUGINS
+from pulpcore.app.apps import pulp_plugin_configs
 
 
 class PulpAutoSchema(AutoSchema):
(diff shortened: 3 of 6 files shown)


bmbouter:
This changes from 

```
    "versions": [
        {
            "component": "pulpcore",
            "version": "3.10.0.dev0"
        },
        {
            "component": "pulp_ansible",
            "version": "0.7.0.dev0"
        },
        {
            "component": "pulp_file",
            "version": "1.6.0.dev0"
        }
    ]
```

To

```
    "versions": [
        {
            "component": "core",
            "version": "3.10.0.dev"
        },
        {
            "component": "ansible",
            "version": "0.7.0.dev"
        },
        {
            "component": "file",
            "version": "1.6.0.dev"
        }
    ],
```

pulpcore/app/apps.py
```diff
@@ -65,11 +65,11 @@ def __init__(self, app_name, app_module):
             self.version
         except AttributeError:
             msg = _(
-                f"The plugin `{self.label}` is missing a version attribute. Starting with "
+                "The plugin `{}` is missing a version declaration. Starting with "
                 "pulpcore==3.10, plugins are required to define their version on the "
                 "PulpPluginAppConfig subclass."
             )
-            warnings.warn(msg, FutureWarning)
+            raise ImproperlyConfigured(msg.format(self.label))
```
mdellweg, line 72: :+1:

pulpcore/openapi/__init__.py
```diff
@@ -446,11 +445,12 @@ def get_schema(self, request=None, public=False):
...
         }
+
         # Adding plugin version config
-        components = ["pulpcore"] + INSTALLED_PULP_PLUGINS
-        result["info"]["x-pulp-app-versions"] = {
-            component: get_distribution(component).version for component in components
-        }
+        result["info"]["x-pulp-app-versions"] = {}
+        for app in pulp_plugin_configs():
+            result["info"]["x-pulp-app-versions"][app.label] = app.version
```
mdellweg, line 452:
```suggestion
        result["info"]["x-pulp-app-versions"] = {
            app.label: app.version
            for app in pulp_plugin_configs()
        }
```

mdellweg (commented):
Sounds good to me.
Just a coding style suggestion.

goosemania (approved):
Thanks! Looks good to me.

goosemania:
> This changes from
> 
> ```
>     "versions": [
>         {
>             "component": "pulpcore",
>             "version": "3.10.0.dev0"
>         },
>         {
>             "component": "pulp_ansible",
>             "version": "0.7.0.dev0"
>         },
>         {
>             "component": "pulp_file",
>             "version": "1.6.0.dev0"
>         }
>     ]
> ```
> 
> To
> 
> ```
>     "versions": [
>         {
>             "component": "core",
>             "version": "3.10.0.dev"
>         },
>         {
>             "component": "ansible",
>             "version": "0.7.0.dev"
>         },
>         {
>             "component": "file",
>             "version": "1.6.0.dev"
>         }
>     ],
> ```

@jlsherrill , FYI, this change is coming in 3.10. I think Katello uses the status endpoint to check pulp services state but I wonder if you check the installed plugins as well.

pulpcore/app/views/status.py
```diff
@@ -42,16 +41,22 @@ class StatusView(APIView):
...
+         * disk usage information
         """
-        components = ["pulpcore"] + INSTALLED_PULP_PLUGINS
-        versions = [
-            {"component": component, "version": get_distribution(component).version}
-            for component in components
-        ]
+        versions = []
+        for app in pulp_plugin_configs():
+            versions.append({"component": app.label, "version": app.version})
```
mdellweg, line 58: I just realized, this changes the keys and not only the trailing "0" in the versions.
This will break the version check logic in the cli.
Can we get back to the old names`

mdellweg (changes requested):

pulpcore/app/views/status.py
bmbouter: I also thought about this, but I don't see an easy way to. :/ What this change does is hand control over to the plugin for the name it provides. So it's up to the plugins really. Let's talk about it at the pulpcore meeting, I put an agenda item on there.

mdellweg: And BTW. This is what breaks the CI here too.

bmbouter:
@mdellweg any link on where I should look to fix the CI? Also is this something I'll need to also fix in the plugin template?

mdellweg:
> @mdellweg any link on where I should look to fix the CI? Also is this something I'll need to also fix in the plugin template?

https://github.com/pulp/pulpcore/blob/master/.ci/ansible/start_container.yaml#L87 is the check that fails.
https://github.com/pulp/pulpcore/blob/master/.github/workflows/scripts/before_install.sh#L37 is where that variable is set.

mdellweg:
Thinking more about this: The check is really valuable, as it catches oops-we-updated-the-thing-to-test-from-pypi-because-of-dependencies mistakes. But it relies on the assumption that there is only one plugin in the repository.

bmbouter:
@mdellweg I agree it's valuable but we no longer have info required to perform it (right?). All I can think to do is to disable the check. What are our other options? Can you (or someone on the CI team) give me some advice? Thanks!

mdellweg:
> @mdellweg I agree it's valuable but we no longer have info required to perform it (right?). All I can think to do is to disable the check. What are our other options? Can you (or someone on the CI team) give me some advice? Thanks!

For _our_ well curated "single_plugins_in_one_repo" we should still able to find the information what version it is supposed to be, right? I really want to avoid testing a released version instead of failing for dependency issues.

bmbouter:
@mdellweg my apologies. I'm haven't looked into that CI check yet and I'm not up to speed on what it does. Can you elaborate a bit more? Thanks!

mdellweg:
Sure, it makes sure, that the code that is installed inside the container is the one we are supposed to test.
We had previous incidents, where a plugin by not being compatible with the provided version of pulpcore downgraded pulpcore to a compatible one from pypi, and if it weren't for mismatching functional tests, this would have gone unnoticed.

.github/workflows/scripts/before_install.sh
```diff
@@ -34,7 +34,7 @@ else
 fi
 mkdir .ci/ansible/vars || true
 echo "---" > .ci/ansible/vars/main.yaml
-echo "component_name: pulpcore" >> .ci/ansible/vars/main.yaml
+echo "component_name: pulp" >> .ci/ansible/vars/main.yaml
```
mdellweg, line 37:
```suggestion
echo "component_name: core" >> .ci/ansible/vars/main.yaml
```

template_config.yml
```diff
@@ -19,7 +19,7 @@ deploy_daily_client_to_rubygems: true
 deploy_to_pypi: true
 docker_fixtures: true
 docs_test: true
-plugin_app_label: pulpcore
+plugin_app_label: pulp
```
mdellweg, line 22:
```suggestion
plugin_app_label: core
```

pulpcore/openapi/__init__.py
bmbouter: I'd like to decline this change if that's alright. The list comprehension with the iterator at the end is less readable for me. What I wrote is a bit slower though.

mdellweg:
I'm about to say ACK here, but we need to merge 
https://github.com/pulp/plugin_template/pull/341
first and roll it out to all plugins (and their still tested stable branches) first.

bmbouter:
@mdellweg I reapplied the plugin template from https://github.com/pulp/plugin_template/pull/341/files and pushed it here to see if this change will pass with that plugin template change.

mdellweg (approved):

PR #1103, opened 2021-02-01
Enable nesting WorkingDirectory() calls

dralley:
closes: #7815
https://pulp.plan.io/issues/7815

### DIFF
--- a/pulpcore/tasking/storage.py
+++ b/pulpcore/tasking/storage.py
@@ -10,6 +10,10 @@ from rq.job import get_current_job
 class _WorkingDir:
     """
     A base class for temporary working directories.
+
+    TODO: This is very similar to functionality already in the stdlib, why keep our own?
+
+    https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory
     """
 
     # Directory permissions.
@@ -35,8 +39,6 @@ class _WorkingDir:
     def create(self):
         """
         Create the directory.
-
-        The directory is deleted and recreated when already exists.
         """
         os.makedirs(self.path, mode=self.MODE)
 
@@ -71,6 +73,27 @@ class _WorkingDir:
 
     def __str__(self):
         return self.path
+
+    def __enter__(self):
+        """
+        Create the directory and set the CWD to the path.
+
+        Returns: self
+
+        Raises:
+            OSError: On failure.
+        """
+        self.create()
+        self._prev_dir = os.getcwd()
+        os.chdir(self._path)
+        return self
+
+    def __exit__(self, *unused):
+        """
+        Delete the directory (tree) and restore the original CWD.
+        """
+        os.chdir(self._prev_dir)
+        self.delete()
 
 
 def get_worker_path(hostname):
@@ -116,9 +139,32 @@ class WorkerDirectory(_WorkingDir):
             super().create()
 
 
-class WorkingDirectory(_WorkingDir):
+class TaskWorkingDirectory(_WorkingDir):
     """
     RQ Job working directory.
+
+    Path format: <worker-dir>/<task-id>/
+    """
+
+    def __init__(self, job):
+        """
+        Create a WorkingDirectory.
+
+        Args:
+            job (rq.Job): The RQ job to create a working directory for
+
+        Raises:
+            RuntimeError: When used outside of an RQ task.
+        """
+        self.hostname = job.origin
+        self.task_id = job.id
+        self.task_path = os.path.join(get_worker_path(self.hostname), self.task_id)
+        super().__init__(self.task_path)
+
+
+class WorkingDirectory(_WorkingDir):
+    """
+    Provide clean working directory for plugin writers on demand.
 
     Path format: <worker-dir>/<task-id>/<random>/
 
@@ -147,7 +193,8 @@ class WorkingDirectory(_WorkingDir):
         except AttributeError:
             raise RuntimeError(_("May only be used within a Task."))
 
-        self.worker_path = os.path.join(get_worker_path(self.hostname), self.task_id)
+        self.task_path = os.path.join(get_worker_path(self.hostname), self.task_id)
+        super().__init__(self.task_path)
 
     def create(self):
         """
@@ -158,32 +205,10 @@ class WorkingDirectory(_WorkingDir):
         NUM_ATTEMPTS = 10
 
         for try_num in range(NUM_ATTEMPTS):
-            random_segment = "".join(rng.choices(characters, k=8))
-            self._path = os.path.join(self.worker_path, random_segment)
+            random_suffix = "".join(rng.choices(characters, k=8))
+            self._path = os.path.join(self.task_path, "workdir_" + random_suffix)
             try:
                 super().create()
                 break
             except FileExistsError:
                 pass
-
-    def __enter__(self):
-        """
-        Create the directory and set the CWD to the path.
-
-        Returns:
-            pulpcore.tasking.service.WorkingDirectory: self
-
-        Raises:
-            OSError: On failure.
-        """
-        self.create()
-        self._prev_dir = os.getcwd()
-        os.chdir(self._path)
-        return self
-
-    def __exit__(self, *unused):
-        """
-        Delete the directory (tree) and restore the original CWD.
-        """
-        os.chdir(self._prev_dir)
-        self.delete()
--- a/pulpcore/tasking/storage.py
+++ b/pulpcore/tasking/storage.py
@@ -1,4 +1,5 @@
 import os
+import random
 import shutil
 from gettext import gettext as _
 
@@ -6,42 +7,20 @@ from django.conf import settings
 from rq.job import get_current_job
 
 
-class WorkerDirectory:
+class _WorkingDir:
     """
-    The directory associated with a RQ worker.
-
-    Path format: <root>/<worker-hostname>
-
-    Attributes:
-        _path (str): The absolute path.
+    A base class for temporary working directories.
     """
 
     # Directory permissions.
     MODE = 0o700
 
-    @staticmethod
-    def _worker_path(hostname):
-        """
-        Get the root directory path for a worker by hostname.
-
-        Format: <root>/<worker-hostname>
-
-        Args:
-            hostname (str): The worker hostname.
-
-        Returns:
-            str: The absolute path to a worker's root directory.
-        """
-        root = settings.WORKING_DIRECTORY
-        path = os.path.join(root, hostname)
-        return path
-
-    def __init__(self, hostname):
+    def __init__(self, path):
         """
         Args:
-            hostname (str): The worker hostname.
+            path (str): The absolute directory path to use.
         """
-        self._path = self._worker_path(hostname)
+        self._path = path
 
     @property
     def path(self):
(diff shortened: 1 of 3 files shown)


dralley:
@goosemania This is not exactly what was described in the issue, but it's safer because each WorkingDirectory is kept separate rather than having one global working directory for the task.  wdyt?

pulpcore/tasking/storage.py
```diff
@@ -119,41 +133,38 @@ class WorkingDirectory(WorkerDirectory):
...
-        super().__init__(self._hostname())
-        self._path = os.path.join(self._path, self._task_id())
+        for try_num in range(NUM_ATTEMPTS):
+            random_segment = "".join(rng.choices(characters, k=8))
+            self._path = os.path.join(self.worker_path, random_segment)
+            try:
+                super().create()
+                break
+            except FileExistsError:
+                pass
```
dralley, line 167: Alternatively, we could try without using a random suffix first, and only if it collides would we try a random subfolder.

dralley:
Test failures unrelated

mdellweg:
Is this a reimplemetation of https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory ?

dralley:
Yes, basically, I noticed the same thing.  I don't really want to swap the entire implementation right but l now but maybe we could in the future.

pulpcore/tasking/storage.py
```diff
@@ -59,15 +38,7 @@ def create(self):
...
-
-        def create():
-            os.makedirs(self.path, mode=self.MODE)
-
-        try:
-            create()
-        except FileExistsError:
-            self.delete()
-            create()
+        os.makedirs(self.path, mode=self.MODE)
```
goosemania, line 41: Please update the docstring to reflect the changes

goosemania (commented):
In this PR, we recreate worker directory but do not recreate the sub directories in it. Correct? And for every context manager usage plugin will get new sub directory, is my understanding correct? If so and if we choose to do it, we would need to go through the deprecation path I think. It's a change in the current behaviour and it will likely break pulp_rpm which uses this context manager N times throughout sync or publish.
 
I believe one of the problems which was intended to be solved is to always have a directory for a task.
Maybe plugin needs multiple sub dirs, it's a different dimension and you seemed to solve that here, however it would be nice not to let plugin write into other directory by mistake, just because they forgot to use this context manager. And have always a directory set for a task. What do you think?

dralley:
> In this PR, we recreate worker directory but do not recreate the sub directories in it. Correct? 

Correct

>And for every context manager usage plugin will get new sub directory, is my understanding correct?

Correct

>If so and if we choose to do it, we would need to go through the deprecation path I think. It's a change in the current behaviour and it will likely break pulp_rpm which uses this context manager N times throughout sync or publish.

I don't understand why?  Does it assume that different ```WorkingDirectory()``` context manager usages are nonetheless always still using the same working directory?

> I believe one of the problems which was intended to be solved is to always have a directory for a task.

That's still true with this change.  It's just that instead of the ```WorkingDirectory``` returning this task directory, it returns a subdirectory of it.  So instead of using

```/var/lib/pulp/tmp/31326@pulp2-nightly-pulp3-source-centos7.localhost.example.com/325b21f9-f07d-4927-8137-6d592a811281/```

it uses

```/var/lib/pulp/tmp/31326@pulp2-nightly-pulp3-source-centos7.localhost.example.com/325b21f9-f07d-4927-8137-6d592a811281/ub91tswg/```

which is one level deeper.

> however it would be nice not to let plugin write into other directory by mistake just because they forgot to use this context manager

That is true.

**Comment by dralley** [comment]
Just to mention it - the problem with nesting was the "delete and recreate" behavior, which makes sense for the ```WorkerDirectory``` but not so much for the ```WorkingDirectory```.  The second time you enter it would try to use the same directory, notice it already exists, and delete it.  That's not what we want.  We could just have nested calls just re-use the same directory.  But that assumes that the task isn't doing anything internally that could cause path conflicts, which feels like a bad assumption.

**Comment by goosemania** [comment]
> > If so and if we choose to do it, we would need to go through the deprecation path I think. It's a change in the current behaviour and it will likely break pulp_rpm which uses this context manager N times throughout sync or publish.
> 
> I don't understand why? Does it assume that separate, subsequent `WorkingDirectory()` context manager usages are nonetheless always still using the same "dirty" working directory?

Actually, I think I'm wrong. The working directory was recreated every time, so plugin can't rely on any data being there. Disregard this concern then, I agree that no deprecation is needed then.

> 
> > I believe one of the problems which was intended to be solved is to always have a directory for a task.
> 
> That's still true with this change. It's just that instead of the `WorkingDirectory` returning this task directory, it returns a subdirectory of it. So instead of using
> 
> `/var/lib/pulp/tmp/31326@pulp2-nightly-pulp3-source-centos7.localhost.example.com/325b21f9-f07d-4927-8137-6d592a811281/`
> 
> it uses
> 
> `/var/lib/pulp/tmp/31326@pulp2-nightly-pulp3-source-centos7.localhost.example.com/325b21f9-f07d-4927-8137-6d592a811281/<8 random letters>/`
> 
> which is one level deeper.

I understand this part. The story states that pulpcore ensures that there is always a working dir for a task. And it seems to me that with this PR it's still a plugin responsibility. The allowed nesting makes it easier and more usable but it doesn't eliminate the concern that plugins must remember to use WorkingDirectory context manager.

I'm +1 to the approach with subdir , I just think that additionally we should add the working dir creation by pulpcore for every task and plugin can decide whether to have many subdirs or not but at least the task will always have one. What do you think?

File: pulpcore/tasking/storage.py [hunk 3]
```diff
@@ -1,47 +1,30 @@
...
     """
-    The directory associated with a RQ worker.
+    A base class for temporary working directories.
 
-    Path format: <root>/<worker-hostname>
+    TODO: This is very similar to functionality already in the stdlib, why keep our own?
 
-    Attributes:
-        _path (str): The absolute path.
+    https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory
```
bmbouter, line 16: I think we should switch to this and stop carrying so much code that stdlib already provides. Also we don't need to have the WorkerDirectory stuff even reusable, I think we should remove this whole module and move the remaining small pieces around `tempfile.TemporaryDirectory` to `pulpcore.tasking.worker`.

pulpcore/tasking/storage.py
bmbouter: If we switch to the tempfile based implementation doesn't it guarantee it's free of race conditions and it can do the name picking?

pulpcore/tasking/worker.py
```diff
@@ -104,7 +104,8 @@ def perform_job(self, job, queue):
             _set_current_user(user)
             GuidMiddleware.set_guid(task.logging_cid)
 
-        return super().perform_job(job, queue)
+        with TaskWorkingDirectory(job):
```
bmbouter, line 107: Yeah I think this area of the PR with just the right configuration could remove a huge amount of code.

bmbouter:
I also think that since the working directories will be managed every time by pulpcore, we don't even need to carry reusable facilities for working directories. If that's true would could deprecate the ones we have in the plugin API with 3.11 and remove them from 3.12 entirely. What do you think about this also?

pulpcore/tasking/storage.py
dralley: I partially copied this implementation from ```tempfile```.  It'd be free of race conditions regardless due to being namespaced by worker and task.

pulpcore/tasking/worker.py
dralley: The reason I don't want to mess with it right now is because there are *some* differences, namely that I don't think stdlib ```TemporaryDirectory``` will create intermediate directories that don't already exist, and I'm not certain the permissions stuff is the same.

I'm not against a refactor issue being filed though.

pulpcore/tasking/storage.py
dralley: The reason I don't want to mess with it right now is because there are some differences, namely that I don't think stdlib TemporaryDirectory will create intermediate directories that don't already exist, and I'm not certain the permissions stuff is the same. I don't want to scope of my current work to explode too much.

I'm not against a refactor issue being filed though.

dralley:
>I also think that since the working directories will be managed every time by pulpcore, we don't even need to carry reusable facilities for working directories. If that's true would could deprecate the ones we have in the plugin API with 3.11 and remove them from 3.12 entirely. What do you think about this also?

I don't know about this.  The more we enable doing more and different kinds of work within one task, the more we risk the possibility of file collisions in the working directory.  We already hit that once with the migration plugin.  https://pulp.plan.io/issues/7842

pulpcore/tasking/storage.py
bmbouter: Yeah but this piece of work is already a huge refactor. If just unblocking the other issue is the goal (and I can get behind that) then I think just rewriting [this part](https://github.com/pulp/pulpcore/blob/master/pulpcore/tasking/storage.py#L63-L70) to the code below would do:

```
        try:
            os.makedirs(self.path, mode=self.MODE)
        except FileExistsError:
            pass
```

dralley: That would have been enough to make it re-entrant but it wouldn't have solved the collision problem I mentioned earlier.  If a plugin writer asks for a new WorkingDirectory they shold get a clean one.

bmbouter:
I see what you're saying in terms of a) this resolving the problem that blocks you and b) this not changing the behavior plugin writers experience already. I appreciate where you're coming from though. I want to get your work unblocked.

I've not lgtm'd this because I'm stuck on accepting that we're replacing a bunch of code we shouldn't be carrying with an even larger diff of code we shouldn't be carrying. Also code that's duplicated from the stdlib. I don't think this PR necessarily makes our situation worse though, so maybe I should just accept the PR as is.

One thing I could do that would be more helpful is to write the future design on an issue instead of trying to pitch it here.

bmbouter:
> > I also think that since the working directories will be managed every time by pulpcore, we don't even need to carry reusable facilities for working directories. If that's true would could deprecate the ones we have in the plugin API with 3.11 and remove them from 3.12 entirely. What do you think about this also?
> 
> I don't know about this. The more we enable doing more and different kinds of work within one task, the more we risk the possibility of file collisions in the working directory. We already hit that once with the migration plugin. https://pulp.plan.io/issues/7842

I agree when a new tempdir is needed it should be clean. I think the thing Pulp could do better though is have plugin writers use stdlib tools to get those clean directories instead of interfaces we provide. I'll put this onto the issue I write up though, and it can be separate from this PR.

dralley:
+1 to filing a refactor issue, I'll definitely take a look over it.  I just can't get hung up here because even once auto-publish is done, we've got a ton of RPM and migration work to get done by March that I need to help with.

bmbouter (changes requested):
I think this also provides a feature and I want to make sure the release note calls that out. Specifically, it now creates a working directory automatically for each task and changes into it. If that can be clarified some I would lgtm this.

bmbouter:
FYI I filed the refactor issue here: https://pulp.plan.io/issues/8231 at the pulpcore meeting we are planning to do for 3.11, but we should still merge this for now since it only deprecates in 3.11, the code will still be needed.

dralley:
@bmbouter I added to the changelog

bmbouter (approved):
Thanks!

mdellweg (approved):

goosemania (approved):

PR #1108, opened 2021-02-03
Ensure that a task is created in pulp not earlier than the job in redis

goosemania:
Otherwise it can get cancelled when we check for missing tasks/jobs.

re #7912
https://pulp.plan.io/issues/7912

### DIFF
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -214,13 +214,16 @@ def enqueue_with_reservation(
             reservation_record = ReservedResourceRecord.objects.get_or_create(resource=resource)[0]
             TaskReservedResourceRecord.objects.create(resource=reservation_record, task=task)
 
-    task_args = (func, inner_task_id, list(resources), args, kwargs, options)
-    try:
-        q = Queue("resource-manager", connection=redis_conn)
-        q.enqueue(
-            _queue_reserved_task, job_id=resource_task_id, args=task_args, job_timeout=TASK_TIMEOUT
-        )
-    except RedisConnectionError as e:
-        task.set_failed(e, None)
+        task_args = (func, inner_task_id, list(resources), args, kwargs, options)
+        try:
+            q = Queue("resource-manager", connection=redis_conn)
+            q.enqueue(
+                _queue_reserved_task,
+                job_id=resource_task_id,
+                args=task_args,
+                job_timeout=TASK_TIMEOUT,
+            )
+        except RedisConnectionError as e:
+            task.set_failed(e, None)
 
     return Job(id=inner_task_id, connection=redis_conn)


goosemania:
Here is what I saw while debugging:
 - task cancellation always happens for the task in resource manager queue
 - it looks to me like a race between task registered in redis and this regular check for missing tasks
 - I put a breakpoint right before cancellation, and when I tried to check there if a job exists in redis, it does! I believe it's because it was added to redis few milliseconds later than the check happened.

Here is some timing for the job which would have been cancelled if not my breakpoint:
```
In[14]: j = Job.fetch("40b8ce8b-fd8a-4d78-b3ef-9c3aa0788e8e", connection=redis_conn)
In[15]: j
Out[15]: Job('40b8ce8b-fd8a-4d78-b3ef-9c3aa0788e8e', enqueued_at=datetime.datetime(2021, 2, 3, 10, 54, 17, 804527))
In[19]: task.pulp_created
Out[19]: datetime.datetime(2021, 2, 3, 10, 54, 17, 690756, tzinfo=<UTC>)
In[22]: j.created_at
Out[22]: datetime.datetime(2021, 2, 3, 10, 54, 17, 804386)
```
It was created in redis 1+ millisecond later than in pulp.
I logged the time of the check/exception, it was veery close to the job creation time in redis (8.043 vs 8.051 milliseconds) and it takes some time to log it, so my guess is that the check happened earlier than the job was in redis.
`Feb 03 10:54:17 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[30870]: pulp [None]: pulpcore.tasking.worker_watcher:WARNING: 1612349657.8051136`

This PR just puts the part which enqueues job in Redis into the same transaction as the Task creation in Pulp, so the transaction is not committed before the job is enqueued. Alternatively, we can just swap job creation in Redis and Task creation in Pulp. Any preference?

@dralley , @bmbouter , do you agree with my conclusions? do I miss anything?
FWIW, migration tests successfully pass with this PR, even in the torture-tasking-system-by-whipping-the-db-for-every-test-without-shutting-down-the-services mode.

pulpcore/tasking/tasks.py
```diff
@@ -214,13 +214,16 @@ def as_url(r):
...
+        try:
+            q = Queue("resource-manager", connection=redis_conn)
+            q.enqueue(
+                _queue_reserved_task,
+                job_id=resource_task_id,
+                args=task_args,
+                job_timeout=TASK_TIMEOUT,
+            )
+        except RedisConnectionError as e:
+            task.set_failed(e, None)
```
dralley, line 227: Where is the actual change?  I'm just seeing some reformatting

goosemania: The change is in indentation ;) it makes it a part of the transaction

bmbouter: I do think this will significantly improve the situation. Thank you for submitting this. :)

On a somewhat related note, the real issue here is architectural. We created these problems by splitting crucial records between postgresql and redis and redis doesn't support transactions. This is another example of how Pulp would do better to eliminate the use of redis altogether.

bmbouter (approved):
Thanks! :+1:

pulpcore/tasking/tasks.py
goosemania: Agreed on the architectural issue.

dralley (approved):

PR #1110, opened 2021-02-03
PoC: Write only remote password

daviddavis:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
@@ -1,10 +1,11 @@
 """Tests that CRUD repositories."""
+import re
 import time
 import unittest
 from itertools import permutations
 from urllib.parse import urljoin
 
-from pulp_smash import api, config, utils
+from pulp_smash import api, cli, config, utils
 from pulp_smash.pulp3.bindings import monitor_task
 from pulp_smash.pulp3.utils import gen_repo
 
@@ -22,6 +23,7 @@ from pulpcore.tests.functional.utils import skip_if
 from pulpcore.client.pulp_file.exceptions import ApiException
 from pulpcore.client.pulp_file import (
     ApiClient as FileApiClient,
+    FileFileRemote,
     RemotesFileApi,
 )
 
@@ -237,6 +239,9 @@ class CRUDRemoteTestCase(unittest.TestCase):
         self.remote = self.remotes_api.create(self.remote_attrs)
 
     def _compare_results(self, data, received):
+        self.assertFalse(hasattr(received, "password"))
+        data.pop("password", None)
+
         for k in data:
             self.assertEqual(getattr(received, k), data[k])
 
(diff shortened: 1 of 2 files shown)


pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
```diff
@@ -251,6 +256,31 @@ def test_update(self):
...
+        monitor_task(response.task)
+        exc = cli_client.run(["pulpcore-manager", "shell", "-c", shell_cmd])
+        self.assertEqual(exc.stdout.rstrip("\n"), "new")
+
+        # test a PUT request with a new password
+        remote_update = FileFileRemote(name="test_pass", url="http://", password="changed")
+        response = self.remotes_api.update(href, remote_update)
+        monitor_task(response.task)
+        exc = cli_client.run(["pulpcore-manager", "shell", "-c", shell_cmd])
+        self.assertEqual(exc.stdout.rstrip("\n"), "changed")
```
fao89, line 282: sweet!

pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
```diff
@@ -251,6 +256,31 @@ def test_update(self):
...
+        remote = self.remotes_api.create({"name": "test_pass", "url": "http://", "password": "new"})
+        href = remote.pulp_href
+        uuid = re.search(r"/pulp/api/v3/remotes/file/file/([\w-]+)/", href).group(1)
+        shell_cmd = (
+            f"import pulpcore; print(pulpcore.app.models.Remote.objects.get(pk='{uuid}').password)"
+        )
+
+        self.addCleanup(self.remotes_api.delete, href)
+
+        # test a PUT request without a password
```
bmbouter, line 270: Can each of these groupings become their own tests? Does that make sense to do?

bmbouter:
I believe all the possibly sensitive info is [on the Remote MasterModel](https://github.com/pulp/pulpcore/blob/a70ab191d0e431469c04591b6dfaf177a7e6001f/pulpcore/app/models/repository.py#L183-L220), is that right?

### Potentially needing to be write_only
* username
* password
* proxy_url
* proxy_username - not yet merged, but soon https://pulp.plan.io/issues/8167
* proxy_password - not yet merged, but soon https://pulp.plan.io/issues/8167
* client_cert
* client_key

### bmbouter's recommendations
* username - yes because this can be significant
* password - yes, definitely
* proxy_url - no, but it's debatable. Users could set username/password in this. Perhaps we could have the validation disable users from doing that. That would be good. Being able to read back the proxy_url is important for users I think.
* proxy_username - yes because this can be significant
* proxy_password - yes, definitly
* client_cert - no, certs aren't useful without the key
* client_key - yes, definitly

What do you think?

What do you think?

daviddavis:
Makes sense. I filed an issue and added it to the pulpcore agenda topic.

https://pulp.plan.io/issues/8202

PR #1111, opened 2021-02-03
Fixes migration imports to use apps.get_model

bmbouter:
In various places in the migrations Models are imported directly and then
used. This is a hazard for future changes because when the model
definition changes this code can break.

This PR switches data migrations to use apps.get_model(...) intead. It
also fixes a syntax migration I noticed on the downmigrate in migration
50.

closes #8203

### DIFF
--- a/CHANGES/8203.misc
+++ b/CHANGES/8203.misc
@@ -1,0 +1,1 @@
+Switches migrations to import models using `apps.get_model(...)`.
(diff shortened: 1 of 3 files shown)


goosemania (approved):
Just needs a changelog entry

daviddavis (approved):

PR #1113, opened 2021-02-03
Adds the `AccessPolicy.customized` field.

bmbouter:
The `AccessPolicy.customized` fields is a read-only field that defaults
to False. If True, it indicates that the user has modified the policy.

This will be useful for:
* Users to see how a system's RBAC has been modified
* Plugin writers who want to ship changes or bugfixes to default policies
  they previously shipped and don't want to overwrite a user's custom
  policy.

closes #8182


pulpcore/app/viewsets/access_policy.py
```diff
@@ -16,6 +16,7 @@ class Meta:
         model = AccessPolicy
         fields = {
             "viewset_name": NAME_FILTER_OPTIONS,
+            "customized": NAME_FILTER_OPTIONS,
```
mdellweg, line 19: This doesn't really apply here.

pulpcore/app/viewsets/access_policy.py
```diff
@@ -34,3 +35,14 @@ class AccessPolicyViewSet(
...
+
+    def update(self, request, pk, **kwargs):
+        """
+        Update AccessPolicy.
+
+        This ensures that the customized boolean will be set to True anytime the user modifies it.
+        """
+        response = super().update(request, pk, **kwargs)
+        if response.status_code == 200:
+            AccessPolicy.objects.filter(pk=pk).update(customized=True)
```
mdellweg, line 47: I think, this leaves space for a race condition.
Could you just add `"updated": True` in the validation step of the serializer?

pulpcore/tests/functional/api/test_access_policy.py
```diff
@@ -0,0 +1,99 @@
...
+        self.assertEqual(groups_response.count, 1)
+
+        groups_users_response = self.access_policy_api.list(viewset_name="groups/users")
+        self.assertEqual(groups_users_response.count, 1)
+
+        tasks_response = self.access_policy_api.list(viewset_name="tasks")
+        self.assertEqual(tasks_response.count, 1)
+
+    def test_user_can_modify_statements(self):
+        """Test that the user can modify the AccessPolicy.statements"""
```
ipanova, line 42: Can we re-phrase this to 'Test that AccessPolicy.statements can be modified'?
If we want to test that a user can do so, then we should create a user and assign proper perms to him, otherwise all this tests are performed under admin user

bmbouter: Yes let me update the docstring now. Yes this only adds basic serializer tests. Right now there is no multi-user testing of rbac policies at all, so that's bigger than I can fix here. Also these endpoints themselves don't have an AccessPolicy so it's not even multi-user capable atm.

pulpcore/app/viewsets/access_policy.py
bmbouter: Oh yeah I wasn't sure what to write here. What do I write here?

ipanova: that a filter for name which usually a charfield and not boolean
NAME_FILTER_OPTIONS = ["exact", "in", "icontains", "contains", "startswith"] these would not be applicable for boolean

pulpcore/app/viewsets/access_policy.py
bmbouter: I don't understand what is being suggested. I could set `customized=True` in a `AccessPolicySerializer.validate()`, but is that really much different? Also what is the race condition here? Putting it here was inspired by [this code](https://github.com/pulp/pulpcore/blob/master/pulpcore/app/viewsets/upload.py#L59-L80). I want to put it in the best place so let me know where that is.

pulpcore/app/viewsets/access_policy.py
bmbouter: Agreed. Please give a suggestion. I am not very familiar with how these filters are specified.

ipanova: https://django-filter.readthedocs.io/en/stable/ref/filters.html#booleanfilter here is an example how to define a boolean filter.
Just add this field to the filter. `customized = BooleanFilter()...`

daviddavis: `"customized": ["exact"]`

ipanova: the `exact` would work only with `customized=True,true,False,false` but not with `customized=1 or 0`

pulpcore/app/viewsets/access_policy.py
mdellweg: The idea is to get the `customized = True` between the initialization of the serializer an the updating of the AccessPolicy. So setting it customized will be in the very same database call.

In this version, you update the AccessPolicy (in the database) twice, which is not needed, and leaves a window open where the policy is already changed, but customized is still False. Due to circumstances, this window can be open for an arbitrary amount of time.

pulpcore/app/viewsets/access_policy.py
ipanova: i am taking my words back, it does work :)

pulpcore/app/viewsets/access_policy.py
mdellweg: To make this more clear: What i mean is to not touch this file at all, but add a `validate` method in https://github.com/pulp/pulpcore/blob/master/pulpcore/app/serializers/access_policy.py#L32 like
```python
def validate(self, data):
    data = super().validate(data)
    data["customized"] = True
    return data
```

Example to draw from: https://github.com/pulp/pulpcore/blob/master/pulpcore/app/serializers/repository.py#L223

pulpcore/app/viewsets/access_policy.py
bmbouter: I tried `"customized": filters.BooleanFilter(),` and it fails to start with `TypeError: 'BooleanFilter' object is not iterable`.

So for now I put in `"customized": ["exact"]` which does start.

bmbouter: Oh and ty both for helping me with this.

ipanova:
```
class AccessPolicyFilter(BaseFilterSet):
    """
    FilterSet for AccessPolicy.
    """
    customized = filters.BooleanFilter()

    class Meta:
        model = AccessPolicy
        fields = {
            "viewset_name": NAME_FILTER_OPTIONS,
            }
```

This would be the proper way if using BooleanFilter.

pulpcore/app/viewsets/access_policy.py
bmbouter: This is very helpful and I agree it's better. I switched to this implementation, but I am still having an issue. When the bindings call:

```
        response = self.access_policy_api.partial_update(
            tasks_href, {"customized": not task_access_policy.customized}
        )
```

It still sets `customized=True` even though the client side data isn't specifying anything is actually changing.

bmbouter: It does solve one problem though that my response data was out of date so this is definitely better.

bmbouter: This actually worked perfectly thank you for pointing me in the right direction.

```
    def validate(self, data):
        """"
        Validate the AccessPolicy.

        This ensures that the customized boolean will be set to True anytime the user modifies it.
        """
        data = super().validate(data)
        if data:
            data["customized"] = True
        return data
```

pulpcore/app/viewsets/access_policy.py
bmbouter: This is great I'm switching it to use the `customized = filters.BooleanFilter()` as an attribute. Thanks for showing this to me.

bmbouter:
@ipanova @mdellweg @daviddavis I believe this fixes the tests and I think addresses all concerns. Please let me know what you think. Thanks!

daviddavis (approved):

mdellweg (approved):

PR #1116, opened 2021-02-05
Pass blocking IO in content handler to a separate executor

dralley:
The standard async executor is single-threaded (which is perfectly fine for async code)
but blocking IO shreds its throughput. Let's create a separate ThreadPool executor for 
blocking IO so that our async code is actually async.

[noissue]

### DIFF
--- a/pulpcore/content/__init__.py
+++ b/pulpcore/content/__init__.py
@@ -18,7 +18,7 @@ from django.conf import settings  # noqa: E402: module level not at top of file
 from pulpcore.app.apps import pulp_plugin_configs  # noqa: E402: module level not at top of file
 from pulpcore.app.models import ContentAppStatus  # noqa: E402: module level not at top of file
 
-from .handler import Handler  # noqa: E402: module level not at top of file
+from .handler import Handler, loop  # noqa: E402: module level not at top of file
 
 
 log = logging.getLogger(__name__)
@@ -34,10 +34,18 @@ async def _heartbeat():
     i8ln_msg = _("Content App '{name}' heartbeat written, sleeping for '{interarrival}' seconds")
     msg = i8ln_msg.format(name=name, interarrival=heartbeat_interval)
 
+    def get_status_blocking():
+        return ContentAppStatus.objects.get_or_create(name=name)
+
     while True:
-        content_app_status, created = ContentAppStatus.objects.get_or_create(name=name)
-        if not created:
+        content_app_status, created = await loop.run_in_executor(None, get_status_blocking)
+
+        def save_heartbeat_blocking():
             content_app_status.save_heartbeat()
+
+        if not created:
+            await loop.run_in_executor(None, save_heartbeat_blocking)
+
         log.debug(msg)
         await asyncio.sleep(heartbeat_interval)
 
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -1,7 +1,9 @@
+import asyncio
 import logging
 import mimetypes
 import os
 import re
+from concurrent.futures import ThreadPoolExecutor
 from gettext import gettext as _
 
 from aiohttp.client_exceptions import ClientResponseError
@@ -36,6 +38,14 @@ from pulpcore.exceptions import UnsupportedDigestValidationError  # noqa: E402
 from jinja2 import Template  # noqa: E402: module level not at top of file
 
 log = logging.getLogger(__name__)
+
+
+loop = asyncio.get_event_loop()
+# Django ORM is blocking, as are most of our file operations. This means that the
+# standard single-threaded async executor cannot help us. We need to create a separate,
+# thread-based executor to pass our heavy blocking IO work to.
+io_pool_exc = ThreadPoolExecutor(max_workers=2)
+loop.set_default_executor(io_pool_exc)
 
 
 class PathNotResolved(HTTPNotFound):
@@ -116,11 +126,17 @@ class Handler:
         """
         self._reset_db_connection()
 
-        if self.distribution_model is None:
-            base_paths = list(Distribution.objects.values_list("base_path", flat=True))
-        else:
-            base_paths = list(self.distribution_model.objects.values_list("base_path", flat=True))
-        directory_list = ["{}/".format(base_path) for base_path in base_paths]
+        def get_base_paths_blocking():
+            if self.distribution_model is None:
+                base_paths = list(Distribution.objects.values_list("base_path", flat=True))
+            else:
+                base_paths = list(
+                    self.distribution_model.objects.values_list("base_path", flat=True)
+                )
+            return base_paths
+
+        base_paths = await loop.run_in_executor(None, get_base_paths_blocking)
+        directory_list = ["{}/".format(path) for path in base_paths]
         return HTTPOk(headers={"Content-Type": "text/html"}, body=self.render_html(directory_list))
 
     async def stream_content(self, request):
@@ -177,7 +193,13 @@ class Handler:
         base_paths = cls._base_paths(path)
         if cls.distribution_model is None:
             try:
-                return Distribution.objects.get(base_path__in=base_paths).cast()
+                return (
+                    Distribution.objects.select_related(
+                        "repository", "repository_version", "publication", "remote"
+                    )
+                    .get(base_path__in=base_paths)
+                    .cast()
+                )
             except ObjectDoesNotExist:
                 log.debug(
                     _("Distribution not matched for {path} using: {base_paths}").format(
@@ -186,7 +208,9 @@ class Handler:
                 )
         else:
             try:
-                return cls.distribution_model.objects.get(base_path__in=base_paths)
+                return cls.distribution_model.objects.select_related(
+                    "repository", "repository_version", "publication", "remote"
+                ).get(base_path__in=base_paths)
             except ObjectDoesNotExist:
                 log.debug(
                     _("Distribution not matched for {path} using: {base_paths}").format(
@@ -282,41 +306,45 @@ class Handler:
         Returns:
             Set of strings representing the files and directories in the directory listing.
         """
-        if not publication and not repo_version:
-            raise Exception("Either a repo_version or publication is required.")
-        if publication and repo_version:
-            raise Exception("Either a repo_version or publication can be specified.")
 
         def file_or_directory_name(directory_path, relative_path):
             result = re.match(r"({})([^\/]*)(\/*)".format(directory_path), relative_path)
             return "{}{}".format(result.groups()[1], result.groups()[2])
 
-        directory_list = set()
+        def list_directory_blocking():
+            if not publication and not repo_version:
+                raise Exception("Either a repo_version or publication is required.")
+            if publication and repo_version:
+                raise Exception("Either a repo_version or publication can be specified.")
 
-        if publication:
-            pas = publication.published_artifact.filter(relative_path__startswith=path)
-            for pa in pas:
-                directory_list.add(file_or_directory_name(path, pa.relative_path))
+            directory_list = set()
 
-            if publication.pass_through:
+            if publication:
+                pas = publication.published_artifact.filter(relative_path__startswith=path)
+                for pa in pas:
+                    directory_list.add(file_or_directory_name(path, pa.relative_path))
+
+                if publication.pass_through:
+                    cas = ContentArtifact.objects.filter(
+                        content__in=publication.repository_version.content,
+                        relative_path__startswith=path,
+                    )
+                    for ca in cas:
+                        directory_list.add(file_or_directory_name(path, ca.relative_path))
+
+            if repo_version:
                 cas = ContentArtifact.objects.filter(
-                    content__in=publication.repository_version.content,
-                    relative_path__startswith=path,
+                    content__in=repo_version.content, relative_path__startswith=path
                 )
                 for ca in cas:
                     directory_list.add(file_or_directory_name(path, ca.relative_path))
 
-        if repo_version:
-            cas = ContentArtifact.objects.filter(
-                content__in=repo_version.content, relative_path__startswith=path
-            )
-            for ca in cas:
-                directory_list.add(file_or_directory_name(path, ca.relative_path))
+            if directory_list:
+                return directory_list
+            else:
+                raise PathNotResolved(path)
 
-        if directory_list:
-            return directory_list
-        else:
-            raise PathNotResolved(path)
+        return await loop.run_in_executor(None, list_directory_blocking)
 
     async def _match_and_stream(self, path, request):
         """
@@ -351,8 +379,16 @@ class Handler:
             :class:`aiohttp.web.StreamResponse` or :class:`aiohttp.web.FileResponse`: The response
                 streamed back to the client.
         """
-        distro = self._match_distribution(path)
-        self._permit(request, distro)
+
+        def match_distribution_blocking():
+            return self._match_distribution(path)
+
+        distro = await loop.run_in_executor(None, match_distribution_blocking)
+
+        def check_permit_blocking():
+            self._permit(request, distro)
+
+        await loop.run_in_executor(None, check_permit_blocking)
 
         rel_path = path.lstrip("/")
         rel_path = rel_path[len(distro.base_path) :]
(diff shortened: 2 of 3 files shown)


dralley:
Consider this my "day of learning" project :)

We have... a lot... of blocking IO in our supposedly-async content app.

bmbouter:
Thank you for exploring this. Maybe we want to do something like this, but we need to think a lot about it. Is this how we should be increasing throughput?

First let me check my understanding: Am I right in understanding that this doesn't decrease the latency per request, but it increases the number of clients that can be served concurrently (aka increasing client throughput)?

If that's the case how does this compare against:
* Having gunicorn manage the threadpool with more threads per worker: https://docs.gunicorn.org/en/stable/settings.html#threads
* Increase increasing the workers per gunicorn process?
* Increasing the number of gunicorn processes?  <--- idea: if most of our time is just spent waiting in theory we can do this to increase throughput without actually seeing an increase in cpu/ram resources.

dralley:
@bmbouter It improves both the throughput and latency by roughly 2x.  I did notice though that there was no significant improvement beyond one or two threads in the threadpool on my system.  The main benefit is simply not blocking the entire loop waiting on database IO.

I don't have any numbers on how it compares against more processes.  I think more workers is something we might want regardless.  And Adam Winberg did see good results by increasing the count.

Re: gunicorn threads, though - those don't apply to the aiottp worker type we are using, see the note at the bottom.  Having our own threadpool for blocking IO serves the same purpose as having more worker threads in a standard blocking IO http server.

I'm in general agreement that we need to play with the parameters to get the biggest impact, but this PR also has separate value in that right now we have all this async code which is not actually async, which basically violates all the assumptions async code and servers and whatnot are supposed to make.  And Django 3+ has the ability to detect this and complains very very loudly about it (refuses to work unless you set a certain ENV variable).  I haven't tested but this will likely ease the pain of migrating to newer Django when we do so.

mdellweg:
I can only second what you say about breaking the assumptions of async code. With async, the only allowed way to wait for anything is by handing the control back to the main loop (aka calling anything `await <...>`).

bmbouter:
It does offer improvements, but it also offers a huge increase in complexity. We need Pup's code to be very simple. People already find the content app hard to work on.

I want to state an idea to see what you all think: the real measure of our code is how efficient it is is terms of hardware to producing a user experience.

I think that's the key concern actually given that more gunicorn processes can be started and gunicorn workers per process. This is the whole horizontal versus vertical scaling debate. Unless we're increasing the hardware efficiency all we're doing is choosing to vertically scale instead of horizontally scale, and in doing so taking on a significant code complexity.

Maybe scaling vertically is the best thing to do, but maybe it's not. We don't really know until we evaluate a variety of solutions. I'd like us to look at the problem more holistically before we continue down this road.

dralley:
>Unless we're increasing the hardware efficiency all we're doing is choosing to vertically scale instead of horizontally scale, and in doing so taking on a significant code complexity.

This is... the entire purpose of async?  It increases hardware efficiency by eliminating wasted memory needed to scale out to more threads / processes, and eliminating overhead incurred by context switches, at the expense of more complex code. 

>It does offer improvements, but it also offers a huge increase in complexity. We need Pup's code to be very simple. People already find the content app hard to work on.

If this is our goal, then we should just dump "async" content app, and use traditional blocking webserver, and run async code from core within a local scope rather than a global async event loop.  The code will be much simpler.  What we're doing now is the worst of both worlds, the added complexity of using async, with none of the benefits of async, and less horizontal scaling tunables than threaded web server models.

dralley:
Regardless, we have larger issues - just serving the 150-ish requests per second used 60% CPU on all cores.  We can't solve that with more workers, we'd have to reduce the amount of work we're doing.  Whether that means optimization or caching or both.

dralley:
I just did some tests with more workers (4).

pulpcore master handled about 130-135 requests per second with a latency of 0.8 seconds, which is definitely better than it was previously.  However it maxed all 4 CPUs available in the VM.

This PR handled 160-180 with a latency of around 0.6 seconds, also maxing all 4 CPUs available.  So we are hitting some hard resource limits pretty quickly.

dralley:
removed some of my comments -- don't want to drag the discussion away from this particular PR

mdellweg:
I am with dalley here. The whole functionality of the async technology and the aiohttp framework is based on the assumption that no coroutine is wasting time. Blocking calls will introduce lag in places that we can not even anticipate.
And offloading tasks into thread executors is the official way to make originally blocking calls well behaved. Clearly the ultimate goal must be to switch to both async file access as well as an async postgres adapter (and i am yet unsure if this will introduce more or less complexity than the thread executors in the end).

dralley:
@daviddavis @mdellweg This PR does actually make Django 3.2 happy in terms of not complaining about using the ORM in an async context (at least for the content app, the sync pipeline still throws errors everywhere).

bmbouter:
We will wait for 3.13 to release to merge this.

pulpcore/content/handler.py
```diff
@@ -612,13 +711,23 @@ def _serve_content_artifact(self, content_artifact, headers):
...
+                artifact_name = artifact_file.name
+                return (artifact_file, artifact_name)
+
+            (artifact_file, artifact_name) = await loop.run_in_executor(io_pool_exc, blocking_io)
             content_disposition = f"attachment;filename={content_artifact.relative_path}"
             parameters = {"ResponseContentDisposition": content_disposition}
-            url = artifact_file.storage.url(artifact_file.name, parameters=parameters)
+            url = artifact_file.storage.url(artifact_name, parameters=parameters)
             raise HTTPFound(url)
         elif settings.DEFAULT_FILE_STORAGE == "storages.backends.azure_storage.AzureStorage":
```
gerrod3, line 732: Missing an io executor here

pulpcore/content/handler.py
```diff
@@ -364,27 +392,47 @@ async def _match_and_stream(self, path, request):
...
-                    )
-                    publication = publications.latest("repository_version", "-pulp_created")
-                except ObjectDoesNotExist:
-                    pass
+        def blocking_io():
+            # secretly ORM queries
+            publication = getattr(distro, "publication", None)
+            repo_version = getattr(distro, "repository_version", None)
+            repository = getattr(distro, "repository", None)
+            return (publication, repo_version, repository)
```
gerrod3, line 400: We should probably add a `select_related` to the `_match_distribution` call so that we don't do an additional 3 database queries.

gerrod3 (commented):
Another spot you need to add the executor is here https://github.com/pulp/pulpcore/pull/1116/files#diff-453fdf3a6e81f92ff5391cf65f9c6154852133a656d82082021e3beaa887776fR389-R391. This is a hook for plugin writers to define extra responses and can have more blocking IO.

pulpcore/content/handler.py
dralley: I agree, I wanted to do that originally but at the time we were kneecapped by the `BaseDistribution` / `Distribution` split.

pulpcore/content/handler.py
```diff
@@ -38,6 +40,13 @@
 log = logging.getLogger(__name__)
 
 
+loop = asyncio.get_event_loop()
+# Django ORM is blocking, as are most of our file operations. This means that the
+# standard single-threaded async executor cannot help us. We need to create a separate,
+# thread-based executor to pass our heavy blocking IO work to.
+io_pool_exc = ThreadPoolExecutor(max_workers=2)
```
dralley, line 47: I did experiment with the numbers a bit and wasn't seeing much difference past 2 threads, although I wasn't especially thorough.  Once I get back to content app perf work, we can tune this better.

pulpcore/content/handler.py
dralley: Done ^

pulpcore/content/__init__.py
```diff
@@ -34,10 +34,18 @@ async def _heartbeat():
     i8ln_msg = _("Content App '{name}' heartbeat written, sleeping for '{interarrival}' seconds")
     msg = i8ln_msg.format(name=name, interarrival=heartbeat_interval)
 
+    def get_status_blocking():
+        return ContentAppStatus.objects.get_or_create(name=name)
+
     while True:
-        content_app_status, created = ContentAppStatus.objects.get_or_create(name=name)
-        if not created:
+        content_app_status, created = await loop.run_in_executor(io_pool_exc, get_status_blocking)
```
mdellweg, line 41: Would it help readability to use a lambda here?

mdellweg (commented):
I would use `set_default_executor` and then `run_in_executor(None, ...)` to not have to carry that variable around.
Also i'd prefer to use the args argument of `run_in_executor` and `functool.partial` over defining local functions in all places.

https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.set_default_executor

pulpcore/content/handler.py
```diff
@@ -116,11 +125,18 @@ async def list_distributions(self, request):
...
 
-        if self.distribution_model is None:
-            base_paths = list(Distribution.objects.values_list("base_path", flat=True))
-        else:
-            base_paths = list(self.distribution_model.objects.values_list("base_path", flat=True))
-        directory_list = ["{}/".format(base_path) for base_path in base_paths]
+        def get_base_paths_blocking():
+            if self.distribution_model is None:
+                base_paths = list(Distribution.objects.values_list("base_path", flat=True))
+                base_paths.extend(list(Distribution.objects.values_list("base_path", flat=True)))
```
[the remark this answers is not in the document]
gerrod3: I think this is a fragment of the old `BaseDistribution` model, probably from a /s/BaseDistribution/Distribution/g

pulpcore/content/handler.py
```diff
@@ -186,7 +208,9 @@ def _match_distribution(cls, path):
                 )
         else:
             try:
-                return cls.distribution_model.objects.get(base_path__in=base_paths)
+                return cls.distribution_model.objects.select_related(
+                    "repository", "repository_version", "publication"
+                ).get(base_path__in=base_paths)
```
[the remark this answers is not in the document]
gerrod3: Yes, I suggested to use this since we access these values later and they are secretly queries that would needed to be wrapped into a thread executor.

pulpcore/content/handler.py
```diff
@@ -282,41 +306,45 @@ async def list_directory(self, repo_version, publication, path):
...
+            if directory_list:
+                return directory_list
+            else:
+                raise PathNotResolved(path)
 
-        if directory_list:
-            return directory_list
-        else:
-            raise PathNotResolved(path)
+        return await loop.run_in_executor(io_pool_exc, list_directory_blocking)
```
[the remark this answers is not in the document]
dralley: Django 3 has one, fwiw.  https://docs.djangoproject.com/en/3.1/topics/async/#asgiref.sync.sync_to_async

pulpcore/content/handler.py
```diff
@@ -364,27 +400,47 @@ async def _match_and_stream(self, path, request):
...
         if publication:
             if rel_path == "" or rel_path[-1] == "/":
                 try:
                     index_path = "{}index.html".format(rel_path)
-                    publication.published_artifact.get(relative_path=index_path)
+
+                    def get_published_artifact_blocking():
+                        publication.published_artifact.get(relative_path=index_path)
+
+                    await loop.run_in_executor(io_pool_exc, get_published_artifact_blocking)
```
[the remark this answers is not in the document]
dralley: Perhaps, but considering that we can't use that approach in the large majority of cases, I have a mild preference for sticking to one consistent way of doing things.

pulpcore/content/handler.py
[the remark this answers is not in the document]
bmbouter: That sounds good. Does it work with WSGI though? At some point we should consider switching the content app to use ASGI, but we still use WSGI today.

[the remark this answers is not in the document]
dralley: I'm pretty sure it does, but we can find that out later, when we adopt Django 3.

pulpcore/content/handler.py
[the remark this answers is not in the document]
dralley: You're right, I just fixed it.

pulpcore/content/__init__.py
dralley: I don't think so, personally.  In my mind it makes sense to explicitly label the functions as blocking io, it feels more self-documenting.

pulpcore/content/handler.py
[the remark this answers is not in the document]
dralley: I'd rather not roll our own, is what I'm saying. And I think this is fine for now, personally.

pulpcore/content/handler.py
```diff
@@ -351,8 +379,16 @@ async def _match_and_stream(self, path, request):
             :class:`aiohttp.web.StreamResponse` or :class:`aiohttp.web.FileResponse`: The response
                 streamed back to the client.
         """
-        distro = self._match_distribution(path)
-        self._permit(request, distro)
+
+        def match_distribution_blocking():
+            return self._match_distribution(path)
+
+        distro = await loop.run_in_executor(io_pool_exc, match_distribution_blocking)
```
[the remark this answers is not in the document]
dralley: My personal preference is to stick to one consistent way of doing the wrapping, that is explicit about why they are being wrapped (`_blocking` suffix).

Also, I kinda think that making the calls stick out strongly from the rest of the code is a feature, not a bug, considering our long-term goal is to eliminate as many of them as possible.  This is a critical hot loop so we should deliberately try to visibly mark the slow parts as being slow, whether by comment or convention.

dralley:
@mdellweg I think we don't want to change the default executor, the whole point here is that we want to separate executors where the one used for blocking IO does *not* block the other executor.  Or perhaps I misunderstand asyncio?  It's complex as hell so that is quite possible.

mdellweg (approved):
I think, this is ready to go...

The one suggestion i want to reiterate is to use set_default_executor to not carry the `io_pool_exc` variable around.

Django's ORM _is_ thread safe, right?

dralley:
As far as I can tell, yes, it's thread-safe by virtue of creating a new thread-local connection object for every new thread, rather than re-using the connection object.

pulpcore/content/handler.py
gerrod3: Now that these are being brought in through select related, those this need to be in a wrapping block call?

pulpcore/content/handler.py
```diff
@@ -364,27 +399,46 @@ async def _match_and_stream(self, path, request):
...
+
+        (publication, repo_version, repository) = await loop.run_in_executor(
+            io_pool_exc, blocking_io
+        )
+
+        def get_latest_publication_or_version_blocking():
+            nonlocal repo_version
+            nonlocal publication
+            # secretly an ORM query
+            if repository:
```
gerrod3, line 417: Same thing here, this if statement should not be blocking (obviously the code inside is), but having one less blocking call will help with performance.

pulpcore/content/handler.py
```diff
@@ -424,19 +494,33 @@ async def _match_and_stream(self, path, request):
...
+            if repository or repo_version:
+                if repository:
+
+                    def get_latest_version_blocking():
+                        return distro.repository.latest_version()
+
+                    repo_version = await loop.run_in_executor(
+                        io_pool_exc, get_latest_version_blocking
+                    )
+
```
gerrod3, line 512: This is a regression (probably from rebasing). The check for `repository` has already happened and `repo_version`  has already been set.

pulpcore/content/handler.py
```diff
@@ -424,19 +494,33 @@ async def _match_and_stream(self, path, request):
...
                 try:
                     index_path = "{}index.html".format(rel_path)
-                    ContentArtifact.objects.get(
-                        content__in=repo_version.content, relative_path=index_path
-                    )
+
+                    def get_contentartifact_blocking():
+                        ContentArtifact.objects.get(
+                            content__in=repo_version.content, relative_path=index_path
+                        )
```
gerrod3, line 520: Should have a `select_related("artifact")` here for fewer database queries.

pulpcore/content/handler.py
```diff
@@ -459,23 +548,34 @@ async def _match_and_stream(self, path, request):
...
             else:
                 if ca.artifact:
-                    return self._serve_content_artifact(ca, headers)
+                    return await self._serve_content_artifact(ca, headers)
                 else:
                     return await self._stream_content_artifact(
                         request, StreamResponse(headers=headers), ca
                     )
 
         if distro.remote:
```
gerrod3, line 557: I think this a query, not sure how many plugin's use this field.

pulpcore/content/handler.py
```diff
@@ -611,18 +716,23 @@ def _serve_content_artifact(self, content_artifact, headers):
         Returns:
             The :class:`aiohttp.web.FileResponse` for the file.
         """
+
+        def blocking_io():
+            artifact_file = content_artifact.artifact.file
+            artifact_name = artifact_file.name
+            return (artifact_file, artifact_name)
+
+        (artifact_file, artifact_name) = await loop.run_in_executor(io_pool_exc, blocking_io)
```
gerrod3, line 725: If all instances of `content_artifact` have `select_related("artifact")` before this function then this blocking wrapper could be removed.

bmbouter:
This PR looks good to me, but I'd like to give one final look once @gerrod3 and @dralley say it's final. Thanks everyone for this great work!

pulpcore/content/handler.py
dralley: It makes no difference here, because the value is never actually used.  We only do the query to know whether or not it exists (and then do something else in the exception case)

pulpcore/content/handler.py
dralley: I don't *think* so?  But it won't hurt.

pulpcore/content/handler.py
dralley: Same as above, I'll leave a comment to re-evaluate with Django 3.2, which will let us know for sure.

pulpcore/content/handler.py
dralley: Fixed with `select_related()`

pulpcore/content/handler.py
gerrod3: I see, would a `filter().exists()` be better than using a `try/except` block?

gerrod3 (approved):
Looks good to me, just one question about using an exist query.

pulpcore/content/handler.py
dralley: Potentially, yeah

dralley:
@bmbouter This is ready for review

bmbouter (approved):
Thank you so much for putting this together!

PR #1119, opened 2021-02-09
Move CONTENT_ORIGIN check before loading plugin settings

daviddavis:
fixes #8235

### DIFF
--- a/CHANGES/8235.bugfix
+++ b/CHANGES/8235.bugfix
@@ -1,0 +1,1 @@
+Fixed ``AttributeError`` when ``CONTENT_ORIGIN`` is not set.
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -256,6 +256,18 @@ SPECTACULAR_SETTINGS = {
 #       at startup
 ALLOWED_CONTENT_CHECKSUMS = ["md5", "sha1", "sha224", "sha256", "sha384", "sha512"]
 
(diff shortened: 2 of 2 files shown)


bmbouter (approved):
Excellent this is looking good.

daviddavis:
This doesn't work. We have a single call to dynaconf to load the settings (pulpcore and plugins). We'd need to either:

1. Separate these out and have dynaconf load pulpcore, then check `CONTENT_ORIGIN`, then load plugin settings. Not sure if this is possible.
2. Patch plugins (see https://github.com/pulp/pulp_ansible/pull/519)

bmbouter:
Either option seems ok, but I'm also unclear on how to do option (1).

Plugins could be more defensive about this. If they are requiring a required setting and want to more gracefully handle it they could carry a similar exception. We could even move ours into the plugin API to deduplicate things. This is used in some plugins, but I don't think many.

I'm not that familiar with it, but perhaps django checks could be used. Everytime I go to use those though they never provide enough enforcement for the job I'm looking at.

daviddavis:
> I'm not that familiar with it, but perhaps django checks could be used. Everytime I go to use those though they never provide enough enforcement for the job I'm looking at.

This is actually what prompted this bug. I was trying to convert this check into a django check but django checks get run after the settings get loaded so it's too late to check for CONTENT_ORIGIN. There might be a way to add a plugin api function that can call the check though.

PR #1120, opened 2021-02-10
Adds a wrapper around hashlib.new('md5')

bmbouter:
Introducing the new, somewhat questionable
`pulpcore.app.pulp_insecure_md5_hashlib` which is a wrapper around the
real hashlib and specified usedforsecurity=False for the md5 hasher.
All other hashers are instantiated unmodified.

Pulp by default disallows the use of md5 and sha1, so this will only be
used if users went out of their way to also allow md5 to be used. Their
intent should be upheld even in restricted environments. sha1 is usually
not prohibited in restricted environments, but probably should be. Since
it usually is not we only specify usedforsecurity=False for `md5`.

closes #7984


CHANGES/7984.misc
```diff
@@ -0,0 +1,6 @@
+Consolidate all calls to ```hashlib.new` to use the new, somewhat questionable
+``pulpcore.app.pulp_insecure_md5_hashlib``. This adds the ``usedforsecurity=False`` kwarg when
+instantiating the ``md5`` hasher. All other hashers pass through umodified.
```
daviddavis, line 3:
```suggestion
instantiating the ``md5`` hasher. All other hashers pass through unmodified.
```

CHANGES/7984.misc
```diff
@@ -0,0 +1,6 @@
...
+
+``md5`` is disallowed in Pulp by default, so this allows ``md5`` if a user went out of their way to
+express their intent for ``md5`` to be used, it should work even in a restricted environment.
```
daviddavis, line 6: I believe md5 is allowed by default: https://github.com/pulp/pulpcore/blob/master/pulpcore/app/settings.py#L257

pulpcore/app/pulp_insecure_md5_hashlib.py
```diff
@@ -0,0 +1,27 @@
...
+        **kwargs: kwargs to be passed along to the real `hashlib.new()`
+
+    Returns:
+        An instantiated hasher.
+    """
+    if name == 'md5':
+            kwargs['usedforsecurity'] = True
+            return the_real_hashlib.new(*args, **kwargs)
+    else:
+        return the_real_hashlib.new(*args, **kwargs)
```
daviddavis, line 27: :100:

pulpcore/app/pulp_insecure_md5_hashlib.py
```diff
@@ -0,0 +1,27 @@
...
+    Args:
+        name:
+        *args: args to be passed along to the real `hashlib.new()`.
+        **kwargs: kwargs to be passed along to the real `hashlib.new()`
+
+    Returns:
+        An instantiated hasher.
+    """
+    if name == 'md5':
+            kwargs['usedforsecurity'] = True
```
daviddavis, line 24:
```suggestion
            kwargs['usedforsecurity'] = False
```

bmbouter: lol, yup

CHANGES/7984.misc
bmbouter: Man then this really does not feel right.

pulpcore/app/pulp_insecure_md5_hashlib.py
```diff
@@ -0,0 +1,27 @@
...
+        name: The name of the hasher to be instantiated.
+        *args: args to be passed along to the real `hashlib.new()`.
+        **kwargs: kwargs to be passed along to the real `hashlib.new()`
+
+    Returns:
+        An instantiated hasher.
+    """
+    if name == 'md5':
+            kwargs['usedforsecurity'] = False
+            return the_real_hashlib.new(*args, **kwargs)
```
daviddavis, line 25: I think you'll need to add this check too:

```python
if 'usedforsecurity' in inspect.getargspec(hashlib.new)[0]:
```

CHANGES/7984.misc
daviddavis: Agreed.

daviddavis: We could remove it from the defaults and have the installer run our `handle-content-checksums` task when upgrading to 3.11.

pulpcore/app/pulp_insecure_md5_hashlib.py
bmbouter: Yeah so I looked at this, but on a centos8 fips enabled box when I `dnf install python` and run this it gives me `FullArgSpec(args=['name', 'data'], varargs=None, varkw='kwargs', defaults=(b'',), kwonlyargs=[], kwonlydefaults=None, annotations={})`. So first I wasn't sure this was even working.

I think we do need it though so let me add it back in. I thought `hashlib.new() took **kwargs, but in testing it just now I see it does not.

bmbouter: Ohhhhh I'm inspecting new() but I need to inspect hashlib.md5()!

bmbouter: Hmm on my centos8 fips bux I get:

```
[vagrant@centos8-fips ~]$ python3
Python 3.6.8 (default, Apr 16 2020, 01:36:27) 
[GCC 8.3.1 20191121 (Red Hat 8.3.1-5)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import inspect
>>> import hashlib
>>> inspect.getfullargspec(hashlib.md5)
  File "/usr/lib64/python3.6/inspect.py", line 1126, in getfullargspec
    sigcls=Signature)
  File "/usr/lib64/python3.6/inspect.py", line 2273, in _signature_from_callable
    skip_bound_arg=skip_bound_arg)
  File "/usr/lib64/python3.6/inspect.py", line 2097, in _signature_from_builtin
    raise ValueError("no signature found for builtin {!r}".format(func))
ValueError: no signature found for builtin <built-in function openssl_md5>

The above exception was the direct cause of the following exception:

  File "/usr/lib64/python3.6/inspect.py", line 1132, in getfullargspec
    raise TypeError('unsupported callable') from ex
TypeError: unsupported callable
```

daviddavis: Hmm, it seems to work on Fedora 33 with Python 3.9.1:

```
>>> inspect.getfullargspec(hashlib.md5)
FullArgSpec(args=['string'], varargs=None, varkw=None, defaults=(b'',), kwonlyargs=['usedforsecurity'], kwonlydefaults={'usedforsecurity': True}, annotations={})
```

bmbouter: I agree

daviddavis: Looks like this is a limitation against some builds of CPython. From https://docs.python.org/3.7/library/inspect.html#inspect.signature:

> Note: Some callables may not be introspectable in certain implementations of Python. For example, in CPython, some built-in functions defined in C provide no metadata about their arguments.

bmbouter: I think on this FIPS box at least I don't think it can use inspect to get the info it needs.

CHANGES/7984.misc
bmbouter: I think this is what we should do. Are we going to remove just md5 or sha1 and md5 then?

daviddavis: I don't have a huge preference. sha1 is still commonly used in RPM I believe so it might be a hassle to remove it.

daviddavis: A hassle for our users that is.

bmbouter: We should remove both md5 and sha1 from the default list starting with 3.11. sha1 is widely believed to be insecure even though it has not been formally broken. Microsoft, Mozilla, and Google all stopped accepting it in 2017.

pulpcore/app/pulp_insecure_md5_hashlib.py
bmbouter: Agreed. On that system it raised a `TypeError` and also if the caller in kwargs puts a `usedforsecurity` and we provide it again it is also a type error which will fall back to the value the user provided. So I think this is actually a pretty good situation.

daviddavis:
I think `pulp_insecure_md5_hashlib` needs to be added to the plugin api.

dralley (approved):

bmbouter:
So in thinking more about this, we should not have Pulp as an application override the `usedforsecurity` flag in restricted environments. I think it's an unhealthy practice. So instead I propose I:

* keep this PR to consolidate the usage of hasher generation to one place
* rename the `pulpcore.app.pulp_insecure_md5_hashlib` to be `pulp_hashlib` and similarly in the plugin API
* have the `pulpcore.app.pulp_hashlib.new` raise an exception if a hasher is requested that is not allowed
* make a patch available (but don't merge it upstream) that can force `usedforsecurity=False` when md5 is allowed and requested from `pulpcore.app.pulp_hashlib.new` for those codebases and users that want to do so

daviddavis:
This was similar to my proposal of having a specific setting that would enable md5 in FIPS environments but shipping the code separately sounds just as well. How would we ship this patch and what level of support would we offer for it (e.g. would we test it in a FIPS environment)?

bmbouter:
Closing in favor of this PR: https://github.com/pulp/pulpcore/pull/1127/files

PR #1121, opened 2021-02-11
Refactor settings checks using django.core.checks

daviddavis:
fixes #8077

### DIFF
--- a/pulpcore/app/checks.py
+++ b/pulpcore/app/checks.py
@@ -1,7 +1,13 @@
+import sys
+from gettext import gettext as _
 from pathlib import Path
 
 from django.conf import settings
-from django.core.checks import Warning as CheckWarning, register
+from django.core.checks import Error, Warning as CheckWarning, register
+from django.db import connection
+
+from pulpcore import constants
+from pulpcore.app.models import Artifact
 
 
 @register(deploy=True)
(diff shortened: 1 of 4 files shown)


pulpcore/app/checks.py
```diff
@@ -41,3 +45,90 @@ def storage_paths(app_configs, **kwargs):
...
+                id="pulpcore.E003",
+            )
+        )
+
+    for checksum in settings.ALLOWED_CONTENT_CHECKSUMS:
+        if checksum in unknown_checksums:
+            # the E003 check will handle this problem. skip to prevent FieldError.
+            continue
+
+        if Artifact.objects.filter(**{checksum: None}).exists():
```
fao89, line 109: from the logs, it seems it is executed before the migrations, so it breaks because `core_artifact` is not on the database yet

mdellweg: I read, that this is taken care of in line 120.

pulpcore/app/settings.py
```diff
@@ -270,76 +262,3 @@
     ENVVAR_FOR_DYNACONF="PULP_SETTINGS",
 )
 # HERE ENDS DYNACONF EXTENSION LOAD (No more code below this line)
```
mdellweg, line 264: Ha! Now it makes sence!

mdellweg (approved):
Not an expert on django checks, but this looks reasonable.

bmbouter:
When I run `pstop` then `pclean` I get this error with this PR:

```
(pulp) [vagrant@pulp3-source-fedora33 docs]$ pclean
pstasystemctl stop pulpcore-content pulpcore-worker@1 pulpcore-worker@2 pulpcore-resource-manager pulpcore-api
rt
pulp [None]: root:INFO: Executing... "DROP DATABASE "pulp";"
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/core/management/__init__.py", line 375, in execute
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/core/management/base.py", line 323, in run_from_argv
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/core/management/base.py", line 364, in execute
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django_extensions/management/utils.py", line 62, in inner
    ret = func(self, *args, **kwargs)
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django_extensions/management/commands/reset_db.py", line 175, in handle
    cursor.execute(drop_query)
psycopg2.errors.ObjectInUse: database "pulp" is being accessed by other users
DETAIL:  There is 1 other session using the database.
```

bmbouter:
I can't remember what our motivation was to move these to be django.checks. I know that is probably a better mechanism, but also we've had problems using them for our assertions. Whenever I've tried to make a django.check I couldn't be sure it was running at startup for example.

daviddavis:
> When I run pstop then pclean I get this error with this PR:

Ah, I tried to use the `Artifact` model instead of db cursor. Looks like I'l have to switch back to the cursor.

> I can't remember what our motivation was to move these to be django.checks. I know that is probably a better mechanism, but also we've had problems using them for our assertions. Whenever I've tried to make a django.check I couldn't be sure it was running at startup for example.

It's definitely a better mechanism. It's more in line with what django recommends and it keeps business logic out of the settings file where it doesn't belong. I'd be interested in knowing more about the problems you had. For one thing, the storage checks currently in the file are deployment checks so they don't run automatically.

bmbouter:
> It's definitely a better mechanism. It's more in line with what django recommends and it keeps business logic out of the settings file where it doesn't belong.

Agreed on all that; I just couldn't get it working when I did it (see below).

> I'd be interested in knowing more about the problems you had. For one thing, the storage checks currently in the file are deployment checks so they don't run automatically.

It's been a few weeks, but iirc, even when I had a check that should fail in place, when I started each component, the processes still started without failure. I'll test this when you push yours next and if it indeed fails, then I think we're good to go.

PR #1122, opened 2021-02-11
Make the generated openapi spec more concrete and compliant to standard

alikins:
ie, instead of {ansible_repository_href}, the
api urls are of form /pulp/api/v3/repositories/ansible/ansible/

Also makes the generated spec compliant to the openapi standard,
including not allowing endpoints to start with '{' and requiring
endpoints to start with '/'

### DIFF
--- a/pulpcore/openapi/__init__.py
+++ b/pulpcore/openapi/__init__.py
@@ -380,8 +380,6 @@ class PulpSchemaGenerator(SchemaGenerator):
 
             schema = view.schema
 
-            path = self.convert_endpoint_path_params(path, view, schema)
-
             # beware that every access to schema yields a fresh object (descriptor pattern)
             operation = schema.get_operation(path, path_regex, method, self.registry)
 


daviddavis:
I realize that our API is not compatible with the openapi spec but this change would fundamentally alter our API by instead having endpoints/bindings accept ids instead of hrefs (the former of which we're not exposing in responses). It might be better to raise this as a discussion on our mailing lists about the motivations for this change.

bmbouter:
Are we sure it's not compatible with the V3 standard? This is how the google APIs work for example, and there was an openapi schema issue they opened (I don't have the link handy) to have this be supported. I remember believing it wasn't compatible with V2, but it would be with V3.

alikins:
Another option would be to extract the endpoint/url/path generating bits from PulpSchemaGenerator.parse() to a method, and let galaxy_ng override it.  Or maybe it an option or config setting etc.

PR #1125, opened 2021-02-12
Adds additional d_artifact.remote check

bmbouter:
Not all d_artifact objects have remotes, e.g. the migration plugin does
not. Other places in the code account for this, but there is one area
that does not. This adds an additional check to one codepath that checks
for equality of `d_artifact.remote.pk` only if `d_artifact.remote`
exists.

closes #7876

### DIFF
--- a/CHANGES/7876.bugfix
+++ b/CHANGES/7876.bugfix
@@ -1,0 +1,2 @@
+Fixed a bug experienced by the migration plugin where all content objects are assumed to have a
+remote associated with them.
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -293,11 +293,11 @@ class RemoteArtifactSaver(Stage):
                     raise ValueError(
                         msg.format(rp=content_artifact.relative_path, c=d_content.content)
                     )
-                for remote_artifact in content_artifact._remote_artifact_saver_ras:
-                    if remote_artifact.remote_id == d_artifact.remote.pk:
-                        break
-                else:
-                    if d_artifact.remote:
+                if d_artifact.remote:
+                    for remote_artifact in content_artifact._remote_artifact_saver_ras:
+                        if remote_artifact.remote_id == d_artifact.remote.pk:
+                            break
+                    else:
                         remote_artifact = self._create_remote_artifact(d_artifact, content_artifact)
                         needed_ras.append(remote_artifact)
         return needed_ras


pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -294,8 +294,9 @@ def _needed_remote_artifacts(self, batch):
                         msg.format(rp=content_artifact.relative_path, c=d_content.content)
                     )
                 for remote_artifact in content_artifact._remote_artifact_saver_ras:
-                    if remote_artifact.remote_id == d_artifact.remote.pk:
-                        break
+                    if d_artifact.remote:
+                        if remote_artifact.remote_id == d_artifact.remote.pk:
+                            break
```
dralley, line 299: Should we lift the conditional to the outside of the loop since both branches have this same conditional?

CHANGES/7876.bugfix
```diff
@@ -0,0 +1,2 @@
+Prevents a traceback experienced by the migration plugin where all content objects are assumed to
```
ipanova, line 1: can it be re-phrased to the past tense?

bmbouter: Yes. I keep getting this wrong and using imperative (which is what commit messages should be). Whoops!

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -293,11 +293,11 @@ def _needed_remote_artifacts(self, batch):
                     raise ValueError(
                         msg.format(rp=content_artifact.relative_path, c=d_content.content)
                     )
-                for remote_artifact in content_artifact._remote_artifact_saver_ras:
-                    if remote_artifact.remote_id == d_artifact.remote.pk:
-                        break
-                else:
-                    if d_artifact.remote:
+                if d_artifact.remote:
```
bmbouter, line 296: What does seem right to me is that the `d_artifact` object is actually being defined implicitly in the loop on L#267. Is that what you all expected also?

bmbouter: I think it's missing a `for d_artifact in d_content.d_artifacts:`

I believe this section of code is pretty messed up because of this.

mdellweg: It's in line 288, no?

bmbouter: You're right! I see now. Thank you.

Maybe separate from this PR I'll refactor this to be clearer and with equivalent performance. But for now I think this PR is ready for review.

mdellweg (approved):
Ping me for reviewing the refactor here. Can't wait to see it. ;)

ipanova (approved):

PR #1127, opened 2021-02-15
Adds a wrapper around hashlib.new()

bmbouter:
Introducing the new, `pulpcore.app.pulp_hashlib` which provides the
`new` function, a wrapper around the real `hashlib.new()`. This checks
the `ALLOWED_CONTENT_CHECKSUMS` setting and raises an Exception if a
hasher not declared as allowed is defined.

closes #7984


dralley (approved):

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -445,3 +445,13 @@ galaxy_ui.yml>`_ in the installers ``pulp_devel`` role.
 For help contributing or changing a plugin-specific installation, please reach out to the installer
 maintainers either through the developer mailing list (``pulp-dev@redhat.com``) or on Freenode in
 the developer channel, ``#pulp-dev``.
+
+.. _checksum-use-in-plugins:
+
+Checksum Use In Plugins
+-----------------------
+
+The ``ALLOWED_CONTENT_CHECKSUMS`` setting provides the list of allowed checksums for that Pulp
```
ipanova, line 454: s/that/the

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -445,3 +445,13 @@ galaxy_ui.yml>`_ in the installers ``pulp_devel`` role.
...
+installation. Plugin writers are expected to restrict their use of checksums to this list. The
+plugin API provides the ``pulpcore.plugin.pulp_hashlib`` module which provides the ``new`` function.
+This is a wrapper around ``hashlib.new`` and should be used instead of it.
```
ipanova, line 457: can you expand on why plugin writers should use the wrapper? `... it ensures only
allowed hashers listed in ``ALLOWED_CONTENT_CHECKSUMS`` can be instantiated`

pulpcore/app/pulp_hashlib.py
```diff
@@ -0,0 +1,36 @@
+"""A wrapper around `hashlib` providing only hahsers named in settings.ALLOWED_CONTENT_CHECKSUMS"""
```
ipanova, line 1: /s/hahsers/hashers

pulpcore/plugin/pulp_hashlib.py
```diff
@@ -0,0 +1,3 @@
+"""A wrapper around `hashlib`, ensuring only allowed hashers are used"""
```
ipanova, line 1: s/allowed hashers/ allowed hashers specified in the ALLOWED_CONTENT_CHECKSUMS setting

ipanova:
@bmbouter  can this file be removed? I believe it is not related to this PR `docs/static/spinny_disk_hard_drive_benchmark.png `

daviddavis:
The code here looks good to me. Is there a patch to review and test with this?

docs/plugins/plugin-writer/concepts/index.rst
bmbouter: I think `that` instead of `the` is the correct partitive case.

docs/plugins/plugin-writer/concepts/index.rst
bmbouter: Yes let me add some more here.

daviddavis: Also, it came up before and maybe it's worth documenting here: does `ALLOWED_CONTENT_CHECKSUMS` apply to things besides content? For example, the rpm plugin checks the checksums of repo metadata files before deciding whether to open and parse them during syncs:

https://github.com/pulp/pulp_rpm/blob/bc5b2da9ef7a5e67e1458468b07c68187e2855ef/pulp_rpm/app/tasks/synchronizing.py#L172

bmbouter: fixed in next push

pulpcore/app/pulp_hashlib.py
bmbouter: ty!

pulpcore/plugin/pulp_hashlib.py
bmbouter: I used the docstring from the other one because this needs to be a single line and it got too long. I think it brings in the elements needed though.

docs/plugins/plugin-writer/concepts/index.rst
bmbouter: @daviddavis let me add that, but I want to ask to get it right. Is the gist that the `ALLOWED_CONTENT_CHECKSUMS` not only restricts the hashers to be used to compute checksums, but also restricts the checksums plugin code should pass through it, e.g. checksums in publications?

bmbouter: Actually let me push a change and tell me what you all think.

docs/plugins/plugin-writer/concepts/index.rst
ipanova: i see now, i needed to read the sentence couple of times

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -445,3 +445,26 @@ galaxy_ui.yml>`_ in the installers ``pulp_devel`` role.
...
+
+The ``ALLOWED_CONTENT_CHECKSUMS`` setting provides the list of allowed checksums a Pulp installation
+is allowed to handle. This includes two types of "checksum handling":
+
+1. Generating checksums. Only hashers in the ``ALLOWED_CONTENT_CHECKSUMS`` list should be used for
+   checksum generation.
+2. Passing through checksum data to clients. Pulp installations should not deliver checksum data to
+   clients that are not in the ``ALLOWED_CONTENT_CHECKSUMS`` list. For example, the RPM plugin
+   publications contain checksums that Pulp does not generate, and it should restrict the checksum
+   data used in those publications to the set of allowed hashers in ``ALLOWED_CONTENT_CHECKSUMS``.
```
ipanova, line 462: :+1:

ipanova (approved):
ty for addressing the comments :octopus:

bmbouter:
Here's the patch I propose https://gist.github.com/bmbouter/31c45dac7de68eeccc35a9f9564c0f28

I put it into a gist for now because the rpms who would receive it don't have the code to patch onto so we'll have to add to the rpms at upgrade time.

bmbouter:
Thank you all for the reviews!

PR #1129, opened 2021-02-16
Add warning if on-demand content does not have allowed checksum

daviddavis:
fixes #7985

### DIFF
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -13,6 +13,7 @@ import sys
 from contextlib import suppress
 from gettext import gettext as _
 from importlib import import_module
+from logging import getLogger
 from pathlib import Path
 from pkg_resources import iter_entry_points
 
@@ -271,6 +272,8 @@ settings = dynaconf.DjangoDynaconf(
 )
 # HERE ENDS DYNACONF EXTENSION LOAD (No more code below this line)
 
+_logger = getLogger(__name__)
+
 try:
     CONTENT_ORIGIN
 except NameError:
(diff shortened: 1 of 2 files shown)


pulpcore/app/settings.py
```diff
@@ -305,41 +308,48 @@
...
-                )
-
-except ImproperlyConfigured as e:
-    raise e
-except Exception:
-    # our check could fail if the table hasn't been created yet or we can't get a db connection
-    pass
-finally:
-    connection.close()
+                _logger.warn(_("Warning: detected remote content without allowed checksums."))
```
daviddavis, line 347: This is just a placeholder until https://pulp.plan.io/issues/7986 is fixed and we can tell users how to proceed.

ipanova (approved):

pulpcore/app/settings.py
```diff
@@ -305,41 +308,48 @@
...
-                        "Run 'pulpcore-manager handle-artifact-checksums' first to populate "
-                        "missing artifact checksums."
-                    ).format(checksum)
+
+if not (len(sys.argv) >= 2 and sys.argv[1] == "handle-artifact-checksums"):
+    try:
+        with connection.cursor() as cursor:
+            for checksum in ALLOWED_CONTENT_CHECKSUMS:
+                # can't import Artifact here so use a direct db connection
+                cursor.execute(f"SELECT count(pulp_id) FROM core_artifact WHERE {checksum} IS NULL")
```
bmbouter, line 317: One idea I was wondering about (not to redo this PR but in general) is: what if we used the get_model stuff from django for things like this (next time). My understanding is that reconstructs the object on-demand by inspecting the schema the db reports so perhaps it would give us a totally usable model without needing an import. Just an idea for later to share. I don't know if it will work.

bmbouter (approved):
This looks good to me. Thanks @daviddavis !

pulpcore/app/settings.py
daviddavis: The main reason we do this is that otherwise we get these connection errors where db connections are being kept alive somehow. I'll give it a try though.

bmbouter: Cool if you want. I'm curious. It may not help at all tho with that problem.

daviddavis: It raises an exception ("django.core.exceptions.AppRegistryNotReady: Models aren't loaded yet."). We could probably move these checks though to the app config ready hook. 

I think we should investigate more and I've added a comment to the refactoring story: https://pulp.plan.io/issues/8265#note-3

bmbouter: merge on!

PR #1128, opened 2021-02-16
Enable CLI test

daviddavis:
[noissue]

Please be sure you have read our documentation on creating PRs:
https://docs.pulpproject.org/contributing/pull-request-walkthrough.html

PR #1130, opened 2021-02-17
Setting for customizing OpenAPI to use or not pulp_href

fao89:
[noissue]

### DIFF
--- a/pulpcore/openapi/__init__.py
+++ b/pulpcore/openapi/__init__.py
@@ -379,8 +379,8 @@ class PulpSchemaGenerator(SchemaGenerator):
                 continue
 
             schema = view.schema
-
-            path = self.convert_endpoint_path_params(path, view, schema)
+            if spectacular_settings.user_settings.ENABLE_PULP_HREF:
+                path = self.convert_endpoint_path_params(path, view, schema)
 
             # beware that every access to schema yields a fresh object (descriptor pattern)
             operation = schema.get_operation(path, path_regex, method, self.registry)
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -234,6 +234,7 @@ SPECTACULAR_SETTINGS = {
     "SERVE_URLCONF": ROOT_URLCONF,
     "DEFAULT_GENERATOR_CLASS": "pulpcore.openapi.PulpSchemaGenerator",
     "DEFAULT_SCHEMA_CLASS": "pulpcore.openapi.PulpAutoSchema",
+    "ENABLE_PULP_HREF": True,
     "COMPONENT_SPLIT_REQUEST": True,
     "COMPONENT_NO_READ_ONLY_REQUIRED": True,
     "TITLE": "Pulp 3 API",


pulpcore/openapi/__init__.py
```diff
@@ -380,7 +380,8 @@ def parse(self, request, public):
 
             schema = view.schema
 
-            path = self.convert_endpoint_path_params(path, view, schema)
+            if request and "bindings" in request.query_params:
+                path = self.convert_endpoint_path_params(path, view, schema)
```
mdellweg, line 384: I'm pretty sure, this will break both the CLI and squeezer.

daviddavis:
I enabled pthe CLI test against pulpcore](https://github.com/pulp/pulpcore/pull/1128) so if you rebase, GHA should run them and tell you if your changes will work with the CLI.

bmbouter:
I expect the openAPI schema to work the same on all installations. If not calling `path = self.convert_endpoint_path_params(path, view, schema)` produces a schema that works well in all places, e.g. squeezer, bindings, docs, for all plugins etc, then why do we even do it in the first place?

My understanding is that this 'alternate' schema doesn't work everywhere, and if it did we would have used it initially.

mdellweg:
I agree. And having an option that is basically a kill switch for all the tooling we provide sounds like a bad idea.
If both options work, then we may provide a different endpoint with the `conformant` api doc.

Maybe the reason for the choice of this schema was rooted in incapabilities of openapi2.
But i think this is about the fact, that we return items by their href, and therefore that href should be a substitution parameter in the api path. This is our way of dealing with polymorphy, and again maybe openapiv3 can do better.

fao89:
it was just an experiment, 
it indeed breaks CLI and squeezer, and I tested what would be required to make them work with the new schema, and it is a lot.
Exposing an OpenAPI schema that is not compliant with OpenAPI standards is weird, so I think `/pulp/api/v3/docs/api.json` should yield a valid OpenAPI, the bindings currently use a param for modifying it `/pulp/api/v3/docs/api.json?bindings`.
I tested using `/pulp/api/v3/docs/api.json?bindings` on pulp-cli and it also requires lots of changes, because it modifies the `operation_id` e.g. (`distributions_ansible_ansible_list``to `list`).
I thought about introducing a new query param: `/pulp/api/v3/docs/api.json?pulp_href` and I was planning to test it.

> If both options work, then we may provide a different endpoint with the `conformant` api doc.
I was planning to try it 

> Maybe the reason for the choice of this schema was rooted in incapabilities of openapi2.
Yes, it was an effort not to break the bindings when switching from drf-yasg to drf-spectacular, which I realize was a mistake, switching libs was the right moment for a breaking change.

In summary, we have two things that we are calling by the same name, OpenAPI schema and Pulp schema(which we use for our tooling), and I think we need to differentiate them. Not sure exactly what would be the best way to do this, maybe the options are creating a new endpoint (not sure how to manipulate drf-spectacular for that), or introduce new query params (maybe both combined).

PR #1131, opened 2021-02-17
Upgrading drf-spectacular

fao89:
[noissue]

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,7 +14,7 @@ djangorestframework~=3.12.2
 djangorestframework-queryfields~=1.0.0
 drf-access-policy~=0.8.6
 drf-nested-routers==0.92.1
-drf-spectacular==0.11.0
+drf-spectacular==0.13.2
 dynaconf~=3.1.2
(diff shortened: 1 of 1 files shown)


mdellweg (approved):

daviddavis (dismissed):

requirements.txt
```diff
@@ -14,7 +14,7 @@ djangorestframework~=3.12.2
 djangorestframework-queryfields~=1.0.0
 drf-access-policy~=0.8.6
 drf-nested-routers==0.92.1
-drf-spectacular==0.11.0
+drf-spectacular~=0.13
```
daviddavis, line 17: Actually I think we need to pin to a specific z release to ensure we don't get breaking changes.

In the meantime, I am going to investigate why dependabot isn't bumping our dependencies.

mdellweg: Right. Especially, since we are monkey patching spectacular.
:100: @daviddavis

mdellweg (approved):

daviddavis (approved):

PR #1132, opened 2021-02-17
Enable filters on bindings read requests

fao89:
https://pulp.plan.io/issues/8209
closes #8209

### DIFF
--- a/pulpcore/openapi/__init__.py
+++ b/pulpcore/openapi/__init__.py
@@ -4,6 +4,7 @@ from urllib.parse import urljoin
 from django.core.exceptions import FieldDoesNotExist
 from django.utils.html import strip_tags
 from drf_spectacular.drainage import reset_generator_stats
+from drf_spectacular.extensions import OpenApiFilterExtension
 from drf_spectacular.generators import SchemaGenerator
 from drf_spectacular.openapi import AutoSchema
 from drf_spectacular.plumbing import (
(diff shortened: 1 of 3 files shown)


pulpcore/openapi/__init__.py
```diff
@@ -279,6 +280,23 @@ def _get_response_bodies(self):
 
         return response
 
+    def _get_filter_parameters(self):
```
mdellweg, line 283: I believe, only filtering on the list actions is the desired behavior. On the detail actions you only have single item anyway, why filter it?

mdellweg: Then again, the api seems to allow it, so maybe we should document it accordingly.

fao89: we have a use case for it, but I don't remember where, @ipanova could you please refresh my memory?

mdellweg: And yes, the api seems to allow, which deserves to documented properly...

ipanova: this is the particular case we need to enable
`http GET  :24817/pulp/api/v3/content/container/tags/?repository_version=/pulp/api/v3/repositories/container/container-
push/e4bc3b34-b9b0-45b8-a197-51283a3e4412/versions/2/`

If you look at the content filter you will see that it is possible to filter by repo_version https://github.com/pulp/pulpcore/blob/c600ff1b63ea45e6020d980736f2f749b76e8be8/pulpcore/app/viewsets/content.py#L93

mdellweg: And it is possible to filter (the list of) tags by repository versions. All is fine there.
I'm still surprised, that the detailed GET for a single object can be filtered at all. But as i said, i tried and it is indeed possible. So it may as well be documented.
The result of the call you shared is either that single tag, or nothing, right? So the question you are really after is: Is this tag part of this repo_version. And yes, maybe this is the easiest way to get an answer from the pulp api.

ipanova:
```
The result of the call you shared is either that single tag, or nothing, right? So the question you are really after is: Is this tag part of this repo_version. And yes, maybe this is the easiest way to get an answer from the pulp api.
```
yes.

daviddavis:
Closing per our team meeting.

PR #1149, opened 2021-02-22
Gracefully handle unrepresentable objects

mdellweg:
In the object permission view objects are supposed to be represented by
their pulp uri. This is obviously impossible for objects, that do not
have one. This will change the behaviour from failing badly to returning
a string that indicates the problem.

fixes #8048
https://pulp.plan.io/issues/8048

### DIFF
--- a/pulpcore/app/serializers/user.py
+++ b/pulpcore/app/serializers/user.py
@@ -24,9 +24,12 @@ class ContentObjectField(serializers.CharField):
     def to_representation(self, obj):
         content_object = getattr(obj, "content_object", None)
         if content_object:
-            viewset = get_viewset_for_model(obj.content_object)
-            serializer = viewset.serializer_class(obj.content_object, context={"request": None})
-            return serializer.data.get("pulp_href")
+            try:
+                viewset = get_viewset_for_model(obj.content_object)
+                serializer = viewset.serializer_class(obj.content_object, context={"request": None})
+                return serializer.data.get("pulp_href")
+            except LookupError:
+                return "<unaddressable resource>"
 
 
 class PermissionSerializer(serializers.Serializer):
--- a/CHANGES/8048.bugfix
+++ b/CHANGES/8048.bugfix
@@ -1,0 +1,1 @@
+Added a fallback text in the permission view for resources that cannot be represented by a pulp uri.


mdellweg:
This is not solving the problem that we do not allow handling resources that live outside of the pulp url schema. But i believe, it solves the original problem as described in the ticket.

cc @bmbouter

ipanova:
@bmbouter any thoughts on the approach in this PR?

bmbouter:
My preference is that we should not merge this for 3.11, but if you still want to go forward with it and get two lgtm's I'm ok with that. It does solve the immediate problem as described on the ticket. In terms of timeline though, we have time to think about this more; I checked in with galaxy_ng and they indicated we could not deliver this in 3.11 https://pulp.plan.io/issues/8048#note-16 and that would be fine.

My concern is usability. With this change we're making it ok for the Pulp API to refer to objects that aren't representable. From a usability perspective, I'm sure users would prefer a link to this error message. In the context of the ticker for example, the user is listing object permissions, so with this we'll say to the user here's permission XYZ for object We have no idea. So rather than making that OK, I think it would be better to keep it as a hard error, and work to make those unrepresentable objects, link-representable.

I don't want to be dogmatic about my preference so if others feel strongly that we should be doing this I want to hear about that.

mdellweg:
To me the problem is a little bit more subtle.
As long as all your plugins follow he pulp infrastructure (with a NamedViewset for every model), you are fine.
But for objects that don't do this, we have (to my knowledge) no way to associate a well defined URL to them.
OTOH without this bandaid we are not only breaking the access to those non conforming objects, but we break the whole list-my-permissions endpoint, including pulp-like objects.

pulpcore/app/serializers/user.py
```diff
@@ -24,9 +24,12 @@ class ContentObjectField(serializers.CharField):
...
         content_object = getattr(obj, "content_object", None)
         if content_object:
-            viewset = get_viewset_for_model(obj.content_object)
-            serializer = viewset.serializer_class(obj.content_object, context={"request": None})
-            return serializer.data.get("pulp_href")
+            try:
+                viewset = get_viewset_for_model(obj.content_object)
+                serializer = viewset.serializer_class(obj.content_object, context={"request": None})
+                return serializer.data.get("pulp_href")
+            except Exception:
```
ipanova, line 31: The problem with this approach i have is that we are catching all the exceptions and this way we might just miss other issues.
Like for example this one https://pulp.plan.io/issues/8396 after I have applied this patch I no longer get 500 and i see `obj: "<unaddressable resource>"`

ipanova: can we catch explicitly `LookupError`?

ipanova:
@bmbouter I share your view on this, however to reach the middle ground and enable a path forward I believe we should go ahead and merge this. This fallback will enable GET on object_permisssions instead of completely blocking this call and getting a 500.
I am all for properly fixing it, but given the workload when will we get to that? My bet not earlier than 3.13. We can backport this into 3.11.z for the sake of galaxy needs when we have ready a full solution.
```
$ http :24817/pulp/api/v3/groups/6/object_permissions/
HTTP/1.1 200 OK
Allow: GET, POST, HEAD, OPTIONS
Connection: close
Content-Length: 515
Correlation-ID: e19a5414018442fab14c6dd8be24f7b2
Date: Fri, 12 Mar 2021 13:38:09 GMT
Server: gunicorn/20.0.4
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "count": 3,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 48,
            "obj": "<unaddressable resource>",
            "permission": "container.view_containerdistribution",
            "pulp_href": "/pulp/api/v3/groups/6/object_permissions/48/"
        },
        {
            "id": 49,
            "obj": "<unaddressable resource>",
            "permission": "container.pull_containerdistribution",
            "pulp_href": "/pulp/api/v3/groups/6/object_permissions/49/"
        },
        {
            "id": 50,
            "obj": "/pulp/api/v3/pulp_container/namespaces/c12a4293-e408-4d90-bb41-fcd453a078c1/",
            "permission": "container.view_containerpushrepository",
            "pulp_href": "/pulp/api/v3/groups/6/object_permissions/50/"
        }
    ]
}
```

ipanova (approved):

ipanova:
folks, should this close this PR?

bmbouter:
> folks, should this close this PR?

That would be my preference because while it allows the endpoint to work without the user knowing what object the permission is for it's still very unusable.

daviddavis:
Decided to close this today. Sounds like galaxy will use NamedModelViewSets which will fix this problem.

PR #1150, opened 2021-02-22
[wip] Add get_content method to the repository version model.

newswangerd:
This would allow for plugin writers to query a specific content type in a repository and get back the correct django Model for that content type.

So instead of writing a queryset like this
```
container_models.Manifest.objects.filter(pk__in=repo_version.content)
```

I can now write one like this which avoids the inefficient `pk__in` filter.
```
repo_version.get_content(content_qs = Manifest.objects)
```


pulpcore/app/models/repository.py
```diff
@@ -479,11 +479,42 @@ def _content_relationships(self):
             repository=self.repository, version_added__number__lte=self.number
         ).exclude(version_removed__number__lte=self.number)
 
+    def get_content(self, content_qs=None):
```
dralley, line 482: Let's name this ```from_content_qs``` instead, and get rid of the optional parameter.  Down below, the ```.content``` property function can just call

```self.from_content_qs(Content.objects)``` instead.

**Comment by bmbouter** [comment]
Thanks for submitting this. Here are the things I think need to happen to get this in.

1. Make a Story https://pulp.plan.io/issues/new titled 'As a plugin writer, ....' since this is a feature for plugin writers
2. Link your commit to it with `closes #XXXX`
3. Add a changelog for it in CHANGE/plugin_api/XXXX.feature   since it's a new method `RepositoryVersions.get_content()`.

File: CHANGES/plugin_api/8375.feature [hunk 2]
```diff
@@ -0,0 +1,4 @@
+Add a ``get_content`` method to ``pulpcore.app.models.Repoitory`` that accepts a
```
bmbouter, line 1:
```suggestion
Add a ``get_content`` method to ``pulpcore.plugin.models.RepositoryVersion`` that accepts a
```

dralley (approved):

dralley:
@newswangerd squash the commits once approved so the CI can pass

bmbouter (approved):
thank you!

PR #1151, opened 2021-02-23
Turn WORKER_TTL into a setting

mdellweg:
The heartbeat rate will scale proportional to that timeout.

fixes #8291

### DIFF
--- a/CHANGES/8291.feature
+++ b/CHANGES/8291.feature
@@ -1,0 +1,1 @@
+Added the ``WORKER_TTL`` setting, that specifies the interval a worker is considered missing after its last heartbeat.
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -189,6 +189,16 @@ CONTENT_APP_TTL
    Defaults to ``30`` seconds.
 
 
+.. _worker-ttl:
+
+WORKER_TTL
+^^^^^^^^^^
+
+   The number of seconds before a worker should be considered lost.
+
+   Defaults to ``30`` seconds.
+
+
 .. _remote-user-environ-name:
 
 REMOTE_USER_ENVIRON_NAME
--- a/pulpcore/tasking/constants.py
+++ b/pulpcore/tasking/constants.py
@@ -3,8 +3,6 @@ from types import SimpleNamespace
 TASKING_CONSTANTS = SimpleNamespace(
     # The name of resource manager entries in the workers table
     RESOURCE_MANAGER_WORKER_NAME="resource-manager",
-    # The amount of time (in seconds) after which a worker process is considered missing.
-    WORKER_TTL=30,
     # The amount of time (in seconds) between checks
     JOB_MONITORING_INTERVAL=5,
 )
--- a/pulpcore/app/models/task.py
+++ b/pulpcore/app/models/task.py
@@ -11,6 +11,7 @@ from django.db import models, transaction
 from django.utils import timezone
 from rq.job import get_current_job
 
+from pulpcore.app.settings import WORKER_TTL
 from pulpcore.app.models import (
     AutoAddObjPermsMixin,
     AutoDeleteObjPermsMixin,
@@ -150,7 +151,7 @@ class WorkerManager(models.Manager):
                 are considered by Pulp to be 'online'.
         """
         now = timezone.now()
-        age_threshold = now - timedelta(seconds=TASKING_CONSTANTS.WORKER_TTL)
+        age_threshold = now - timedelta(seconds=WORKER_TTL)
 
         return self.filter(last_heartbeat__gte=age_threshold, gracefully_stopped=False)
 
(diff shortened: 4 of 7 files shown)


docs/settings.rst
```diff
@@ -189,6 +189,16 @@ CONTENT_APP_TTL
    Defaults to ``30`` seconds.
 
 
+.. _worker-ttl:
+
+WORKER_TTL
+^^^^^^^^^^^^^^^
```
bmbouter, line 195:
```suggestion
WORKER_TTL
^^^^^^^^^^
```

CHANGES/8291.feature
```diff
@@ -0,0 +1 @@
+Added the ``WORKER_TTL`` setting, that specifies the interval a worker is considered missing after its last heartbeat.
```
bmbouter, line 1: This is well said. :+1:

pulpcore/tasking/constants.py
```diff
@@ -1,10 +1,12 @@
 from types import SimpleNamespace
 
+from pulpcore.app.settings import WORKER_TTL
+
 TASKING_CONSTANTS = SimpleNamespace(
     # The name of resource manager entries in the workers table
     RESOURCE_MANAGER_WORKER_NAME="resource-manager",
     # The amount of time (in seconds) after which a worker process is considered missing.
-    WORKER_TTL=30,
+    WORKER_TTL=WORKER_TTL,
```
bmbouter, line 9: Can we refactor the code to use the setting directly imported from settings instead of importing it here, setting it here, and then having other things import it from here.

pulpcore/tasking/constants.py
```diff
@@ -1,10 +1,12 @@
...
     # The amount of time (in seconds) between checks
-    JOB_MONITORING_INTERVAL=5,
+    JOB_MONITORING_INTERVAL=int(WORKER_TTL / 6),
```
bmbouter, line 11: So I tested to see if this is indeed spacing out the heartbeats and it did not. I think this is the parents RQ process checking in on its child PID the workhorse which is different. I think this should stay at 5.

We do need to do something though because here's what I learned. With a `WORKER_TTL=30` I get one heartbeat every 15 seconds:

```
Feb 23 17:48:40 pulp3-source-fedora33.localhost.example.com rq[31657]: pulp [None]: root:ERROR: HEARTBEAT
Feb 23 17:48:56 pulp3-source-fedora33.localhost.example.com rq[31657]: pulp [None]: root:ERROR: HEARTBEAT
Feb 23 17:49:11 pulp3-source-fedora33.localhost.example.com rq[31657]: pulp [None]: root:ERROR: HEARTBEAT
```

With a `WORKER_TTL=60` I get only one heartbeat every 45 seconds.

```
Feb 23 17:45:49 pulp3-source-fedora33.localhost.example.com rq[31434]: pulp [None]: root:ERROR: HEARTBEAT
Feb 23 17:46:35 pulp3-source-fedora33.localhost.example.com rq[31434]: pulp [None]: root:ERROR: HEARTBEAT
Feb 23 17:47:20 pulp3-source-fedora33.localhost.example.com rq[31434]: pulp [None]: root:ERROR: HEARTBEAT
```

With `WORKER_TTL=120` I get one heartbeat every 105 seconds

```
Feb 23 17:50:41 pulp3-source-fedora33.localhost.example.com rq[31813]: pulp [None]: root:ERROR: HEARTBEAT
Feb 23 17:52:26 pulp3-source-fedora33.localhost.example.com rq[31813]: pulp [None]: root:ERROR: HEARTBEAT
Feb 23 17:54:11 pulp3-source-fedora33.localhost.example.com rq[31813]: pulp [None]: root:ERROR: HEARTBEAT
```

So we're not actually getting more heartbeats, just fewer of them which means a slow db could equally not deliver the one it needs to. It's just being checked less frequently really which I think explains why experimentally changing this on the migration box did work, but not for the reason we think it did.

I tested with this diff:

```
diff --git a/pulpcore/tasking/worker.py b/pulpcore/tasking/worker.py
index d6c9403cd..ab36dad28 100644
--- a/pulpcore/tasking/worker.py
+++ b/pulpcore/tasking/worker.py
@@ -176,6 +176,7 @@ class PulpWorker(Worker):
             args (tuple): unused positional arguments
             kwargs (dict): unused keyword arguments
         """
+        logging.error('HEARTBEAT')
         handle_worker_heartbeat(self.name)
         check_worker_processes()
         if self.is_resource_manager:
```

bmbouter (changes requested):
So along with this change we need to find out how to have RQ space out its heartbeats.

pulpcore/tasking/constants.py
mdellweg: Well, it's only sending the heartbeat 15 seconds before it is meant to timeout. The code is there:

https://github.com/rq/rq/pull/945/files#r185680378

bmbouter:
I looked into the `unlogged` option to fix this and it didn't work out because the FK in postgresql would require us to mark many tables as `unlogged`. So I think we're back to this as our best option.

I filed this: https://github.com/rq/rq/issues/1426

bmbouter:
I tested this PR against RQ and I think it works https://github.com/rq/rq/pull/1427

bmbouter (approved):
Given that the "spacing out" will be handled in RQ itself and our code can't control that anyway, I think this is LGTM now. :+1:

dralley (approved):

bmbouter:
I see two lgtm's, merging!

PR #1152, opened 2021-02-24
Released import/export from tech-preview state.

ggainey:

### DIFF
--- a/CHANGES/8146.doc
+++ b/CHANGES/8146.doc
@@ -1,0 +1,3 @@
+Removed 'tech preview' label from the import/export documentation.
+
+PulpImport/PulpExport is now a fully-supported part of pulp3.
--- a/docs/workflows/import-export.rst
+++ b/docs/workflows/import-export.rst
@@ -1,10 +1,6 @@
 Pulp Import and Export
 ======================
 
-.. warning::
-    Support for import and export is provided as a tech preview in Pulp 3. Functionality may not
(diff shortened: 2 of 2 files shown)


mdellweg (approved):
This needs the fixes stanza

daviddavis (approved):

ipanova:
moved to draft, we might want to keep it in tech-preview for 3.11

CHANGES/8146.feature
```diff
@@ -0,0 +1,3 @@
+Removed 'tech preview' label from the import/export documentation.
+
+PulpImport/PulpExport is now a fully-supported part of pulp3.
```
daviddavis, line 3: This should be a doc file.

ggainey:
Can we mark this as Draft? There may be an API change needed based on katello's feedback

ipanova:
@ggainey any updates here?

ggainey:
> @ggainey any updates here?

We're on-hold pending what happens in ansible-land RE https://pulp.plan.io/issues/8204 - if we need to do this work, it *might* affect the API. Also, while we've addressed all the API-affecting requests from katello, will request a final "ack" from them before moving fwd on this PR/issue.

bmbouter:
I'd like us to get more data on usage of this feature before removing the tech preview. Also I think we should close these PRs that have been open for so long without a clear plan on when the tech preview labels will be removed.

PR #1153, opened 2021-02-24
Released handle-artifact-checksum from tech-preview.

ggainey:
closes #7928
[nocoverage]

### DIFF
--- a/CHANGES/7928.doc
+++ b/CHANGES/7928.doc
@@ -1,0 +1,3 @@
+Removed 'tech preview' label from ``handle-artifact-checksums`` command.
+
+``handle-artifact-checksums`` is now a fully-supported part of Pulp3.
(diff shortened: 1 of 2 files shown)


CHANGES/7928.feature
```diff
@@ -0,0 +1,3 @@
+Removed 'tech preview' label from ``handle-artifact-checksums`` command.
+
+``handle-artifact-checksums`` is now a fully-supported part of Pulp3.
```
daviddavis, line 3: I would have probably gone for a docs changelog since this is not really a new feature. But feature works for me too.

daviddavis (dismissed):

mdellweg (approved):

CHANGES/7928.feature
ggainey: In my head, moving out of tech-preview is more than just changing documentation - it's announcing the feature 'for reals'.

ipanova: i am also leaning towards using .doc rather than .feature. Feature is there and we have already used once .feature entry when introducing it even in the tech preview state.

mdellweg (approved):

goosemania (approved):

ipanova (approved):

PR #1156, opened 2021-02-26
Have the content app handle UnsupportedDigestValidationError 

daviddavis:
In https://pulp.plan.io/issues/8435 we added a check to downloader to
prevent the download of files without allowed checksums. This code 
updates the downloader to handle the exception that gets raised.

fixes #7989

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -30,6 +30,7 @@ from pulpcore.app.models import (  # noqa: E402: module level not at top of file
     Remote,
     RemoteArtifact,
 )
+from pulpcore.exceptions import UnsupportedDigestValidationError  # noqa: E402
 
 from jinja2 import Template  # noqa: E402: module level not at top of file
 
@@ -501,7 +502,12 @@ class Handler:
                 response = await self._stream_remote_artifact(request, response, remote_artifact)
                 return response
 
-            except ClientResponseError:
+            except (ClientResponseError, UnsupportedDigestValidationError) as e:
+                log.warn(
+                    _("Could not download remote artifact at '{}': {}").format(
+                        remote_artifact.url, str(e)
+                    )
+                )
                 continue
 
         raise HTTPNotFound()
--- a/CHANGES/7989.feature
+++ b/CHANGES/7989.feature
@@ -1,0 +1,1 @@
+Added support in content app for properly handling unknown or forbidden digest errors.


pulpcore/app/models/content.py
```diff
@@ -555,6 +555,20 @@ class RemoteArtifact(BaseModel, QueryMixin):
 
     objects = BulkCreateManager()
 
+    def validate_checksums(self):
```
mdellweg, line 558: Would it make sense to turn this into a pre-save-hook?

mdellweg: Oh, i see now, it's checked when using, not saving the RA.

pulpcore/tests/unit/models/test_content.py
```diff
@@ -128,3 +128,29 @@ def test_forgot_something(self, mock_FORBIDDEN_DIGESTS, mock_DIGEST_FIELDS):
...
+        new_callable=mock.PropertyMock,
+        return_value=set(["md5", "sha1"]),
+    )
+    def test_validate_forbidden_checksums(self, mock_FORBIDDEN_DIGESTS):
+        remote_artifact = RemoteArtifact(md5="5eb63bbbe01eeed093cb22bb8f5acdc3")
+        with self.assertRaises(UnsupportedDigestValidationError):
+            remote_artifact.validate_checksums()
+        remote_artifact.sha256 = "b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9"
+        with self.assertRaises(UnsupportedDigestValidationError):
+            remote_artifact.validate_checksums()
```
mdellweg, line 156: :100: for having unittests. They are so much faster.

mdellweg (approved):
Looks good to me. Just thinking whether we should fail on creating RA with bad checksums?

pulpcore/app/models/repository.py
```diff
@@ -343,6 +343,7 @@ def get_downloader(self, remote_artifact=None, url=None, **kwargs):
             raise ValueError(_("get_downloader() requires either 'remote_artifact' and 'url'."))
         if remote_artifact:
             url = remote_artifact.url
+            remote_artifact.validate_checksums()
```
ipanova, line 346: tiny suggestion to rename this to something more concrete `validate_forbidden_checksums`?

pulpcore/app/models/content.py
daviddavis: We could do a pre-save in addition to this change. There is the possibility where a user downloads a bunch of RAs and changes their allowed checksums so we need to validate RAs before serving them.

pulpcore/content/handler.py
```diff
@@ -501,7 +502,12 @@ async def _stream_content_artifact(self, request, response, content_artifact):
                 response = await self._stream_remote_artifact(request, response, remote_artifact)
                 return response
 
-            except ClientResponseError:
+            except (ClientResponseError, UnsupportedDigestValidationError) as e:
```
mdellweg, line 505: Do we want to handle those exceptions with the same message?
It seems to me that a `ClientResponseError` could be an intermittent problem, while having the wrong digests is more like a permanent one.

mdellweg (approved):
One small suggestion, but still ok this way.

pulpcore/content/handler.py
daviddavis: I am not sure I fully understand. Are you referring to the warning below?

mdellweg: Yes, the warning.
It might be helpful to the admin to realize it's a checksum related problem his users hit repeatedly.
I now see: `str(e)` disregard my comment.

ipanova (approved):

PR #1159, opened 2021-03-02
Add check to downloader to prevent use of forbidden checksums

daviddavis:
This is strictly a plugin api feature at the moment since the pulpcore
code always filters by DIGEST_FIELDS on Artifact before sending
digests to the downloader and DIGEST_FIELDS takes
ALLOWED_CONTENT_CHECKSUMS into account.

fixes #7854

### DIFF
--- a/pulpcore/download/base.py
+++ b/pulpcore/download/base.py
@@ -1,12 +1,17 @@
 import asyncio
 from collections import namedtuple
+from gettext import gettext as _
 import logging
 import os
 import tempfile
 
 from pulpcore.app import pulp_hashlib
 from pulpcore.app.models import Artifact
-from pulpcore.exceptions import DigestValidationError, SizeValidationError
+from pulpcore.exceptions import (
+    DigestValidationError,
+    SizeValidationError,
+    UnsupportedDigestValidationError,
+)
 
 
 log = logging.getLogger(__name__)
@@ -96,6 +101,17 @@ class BaseDownloader:
             self.semaphore = asyncio.Semaphore()  # This will always be acquired
         self._digests = {n: pulp_hashlib.new(n) for n in Artifact.DIGEST_FIELDS}
         self._size = 0
+
+        if expected_digests:
+            bad_digests = [
+                key for key in self.expected_digests.keys() if key in Artifact.FORBIDDEN_DIGESTS
+            ]
+            if bad_digests:
+                raise UnsupportedDigestValidationError(
+                    _("Cannot download artifact at '{}' due to bad digests: {}").format(
+                        self.url, (", ").join(bad_digests)
+                    )
+                )
 
     def _ensure_writer_has_open_file(self):
         """
--- a/CHANGES/plugin_api/7853.feature
+++ b/CHANGES/plugin_api/7853.feature
@@ -1,0 +1,1 @@
+Added a digest check to downloaders to prevent use of forbidden checksums.


pulpcore/download/base.py
```diff
@@ -97,6 +102,17 @@ def __init__(
         self._digests = {n: pulp_hashlib.new(n) for n in Artifact.DIGEST_FIELDS}
         self._size = 0
 
+        if expected_digests:
+            bad_digests = [
+                key for key in self.expected_digests.keys() if key in Artifact.FORBIDDEN_DIGESTS
+            ]
+            if bad_digests:
+                raise UnsupportedDigestValidationError(
+                    _("Cannot download artifact at '{}' due to bad digests: {}").format(
```
bmbouter, line 111:
```suggestion
                    _("Cannot download Artifact at '{}' due to bad digests: {}").format(
```

maybe this tho ^? I'm not sure.

bmbouter:
I think this kind of gets us back into the convo from this morning maybe? I think with the default sha224+ setting of 3.11, if someone specifies sha1 and sha224 together in `expected_digests` I think that's ok. I suspect that is maybe what we didn't agree on so far though. Wdyt?

daviddavis:
So we would just ignore the sha1? That seems like a silent error and insecure to me personally.

bmbouter:
We can make it explicit, but we need to do this consistently in various places. Today we already silently ignore the `sha1` so we just need to think about where else we would need to do it differently in addition to this.

Also with this change in approach, now plugin writers have an additional responsibility of filtering out checksums in their plugin code instead of passing into pulpcore and having pulpcore figure out if a situation is unsafe or not.

I can get on board with a plan like this, but how will we organize ^ changes?

daviddavis:
> We can make it explicit, but we need to do this consistently in various places. Today we already silently ignore the sha1 so we just need to think about where else we would need to do it differently in addition to this.

I'm not aware of where this is happening. Can you point me to it?

> Also with this change in approach, now plugin writers have an additional responsibility of filtering out checksums in their plugin code instead of passing into pulpcore and having pulpcore figure out if a situation is unsafe or not.

I didn't think this was a change in approach (at least based on my reading of the tickets in https://pulp.plan.io/issues/7960). I think it's good though that plugins get errors if they attempt to use forbidden checksums. And I only see pulp_rpm and pulp_ansible use the downloader directly and the former is not setting expected checksums while the latter is using sha256 which should always be supported.

bmbouter:
What I was thinking of was actually here where it sorts them by strength and only checks the strongest available, silently ignoring the others: https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/content.py#L311-L312  Then again, now the values of that field actually changed, so I don't think it would pass silently anymore and also there is validation code earlier up.

I thought the downloaders also did this, but they do not https://github.com/pulp/pulpcore/blob/master/pulpcore/download/base.py#L193-L196

bmbouter:
> I didn't think this was a change in approach (at least based on my reading of the tickets in https://pulp.plan.io/issues/7960). I think it's good though that plugins get errors if they attempt to use forbidden checksums. And I only see pulp_rpm and pulp_ansible use the downloader directly and the former is not setting expected checksums while the latter is using sha256 which should always be supported.

Maybe my understand is just out of date or plain wrong due to being pulled in too many directions at once.

pulpcore/download/base.py
bmbouter: @daviddavis wdyt about this ^? My motivation is to have the user understand that it's not bad like invalid, but bad as in blocklisted and how they can change it.

daviddavis:
Closing in favor of https://pulp.plan.io/issues/8435

PR #1160, opened 2021-03-02
Raise error for on-demand content with forbidden checksums

daviddavis:
fixes #7985

### DIFF
--- a/CHANGES/7985.feature
+++ b/CHANGES/7985.feature
@@ -1,1 +1,1 @@
-Added a warning message if there is on-demand content with no allowed checksums.
+Added an error if there is on-demand content with forbidden checksum(s).
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -340,11 +340,17 @@ if not (len(sys.argv) >= 2 and sys.argv[1] == "handle-artifact-checksums"):
                         ).format(checksum)
                     )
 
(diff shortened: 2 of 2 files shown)


daviddavis:
@pulp/core I talked with @bmbouter and we decided to stop Pulp from starting if there are remote artifacts with forbidden checksums.

As a result, I believe we can close out https://github.com/pulp/pulpcore/pull/1086 and https://github.com/pulp/pulpcore/pull/1156 as WONTFIX.

goosemania:
@daviddavis , it looks like a painful user experience to me, could we add docs or more verbose error message about the steps to solve the problem when they see the error? Or are the docs planned to be a part of a different issue?

daviddavis:
@goosemania I'm also very hesitant about this change so feedback/suggestions are welcome. 

Either we have to error when we encounter forbidden on-demand content checksums at start time or at runtime and I am not sure which is less painful for users..?

I think the hard part will be fixing on-demand content. I don't know that Pulp provides good tooling to make this easy. I've filed an issue though that I think may help: https://pulp.plan.io/issues/8324.

We do have a [reporting tool](https://pulp.plan.io/issues/7986) planned for 3.11 which should show users which repo versions are affected. I've added a note to update this error message.

I agree that more docs would be helpful too. So I've filed: https://pulp.plan.io/issues/8325.

PR #1163, opened 2021-03-03
Add proxy username and password to remote

mdellweg:
fixes #8167
https://pulp.plan.io/issues/8167


mdellweg:
Whichever will be merged last, needs to adjust here:

https://github.com/pulp/pulpcore/pull/1165

pulpcore/app/migrations/0060_data_migration_proxy_creds.py
```diff
@@ -0,0 +1,47 @@
+# Generated by Django 2.2.18 on 2021-03-03 10:59
+
+from urllib.parse import urlparse, urlunparse
+
+from django.db import migrations
+
+
+def move_proxy_creds_up(apps, schema_editor):
```
daviddavis, line 8: :rocket:

daviddavis:
The code looks good to me. It would be nice to have some tests (at least for the validation).

CHANGES/8167.feature
```diff
@@ -0,0 +1,2 @@
+The fields ``proxy_username`` and ``proxy_password`` have been added to remotes.
+Credentials can no longer be specified as part of the ``proxy_url``.
```
bmbouter, line 2: Can this release note identify that a migration moves the credentials from `proxy_url` also?

pulpcore/app/serializers/repository.py
```diff
@@ -178,6 +188,51 @@ def validate_url(self, value):
...
+        Check, that proxy credentials are only provided completely and if a proxy is configured.
+        """
+        data = super().validate(data)
+
+        if "proxy_url" in data:
+            proxy_url = data["proxy_url"]
+        elif self.partial:
+            proxy_url = self.instance.proxy_url
+        else:
+            proxy_url = None
```
bmbouter, line 210: This `else` clauses could be avoided if the top of the function defined:

```
proxy_url = None
proxy_username = None
proxy_password = None
```

This is not a required change, just a suggestion to lower the cyclomatic complexity.

pulpcore/app/serializers/repository.py
```diff
@@ -178,6 +188,51 @@ def validate_url(self, value):
                 return value
         raise serializers.ValidationError(_("url '{}' is not an allowed import path").format(value))
 
+    def validate_proxy_url(self, value):
```
bmbouter, line 191: Can some simple CRUD tests be added for these validation cases please?

bmbouter (changes requested):
I left some requested changes. Mainly it's small things, and then a few tests please. Thanks for opening this!

mdellweg:
Needs to be adjusted now given:

https://github.com/pulp/pulpcore/pull/1165

mdellweg:
also fixes
https://pulp.plan.io/issues/8330

pulpcore/app/serializers/repository.py
```diff
@@ -178,6 +189,39 @@ def validate_url(self, value):
...
+        proxy_password = data.get("proxy_password", proxy_password)
+
+        if (proxy_username or proxy_password) and not proxy_url:
+            raise serializers.ValidationError(
+                _("proxy credentials cannot be specified without a proxy")
+            )
+
+        if bool(proxy_username) is not bool(proxy_password):
+            raise serializers.ValidationError(
+                _("proxy username and password can only be specified together")
```
bmbouter, line 220: Can this declare the fields with their exact names?

pulpcore/app/serializers/repository.py
```diff
@@ -178,6 +189,39 @@ def validate_url(self, value):
...
+        proxy_url = self.instance.proxy_url if self.partial else None
+        proxy_url = data.get("proxy_url", proxy_url)
+        proxy_username = self.instance.proxy_username if self.partial else None
+        proxy_username = data.get("proxy_username", proxy_username)
+        proxy_password = self.instance.proxy_password if self.partial else None
+        proxy_password = data.get("proxy_password", proxy_password)
+
+        if (proxy_username or proxy_password) and not proxy_url:
+            raise serializers.ValidationError(
+                _("proxy credentials cannot be specified without a proxy")
```
bmbouter, line 215: Can this declare the exact names also?

pulpcore/app/migrations/0060_data_migration_proxy_creds.py
```diff
@@ -0,0 +1,47 @@
...
+            try:
+                parsed_url = urlparse(remote.proxy_url)
+                url = parsed_url.netloc
+                netloc = remote.proxy_username + ":" + remote.proxy_password + "@" + url
+                remote.proxy_username = None
+                remote.proxy_password = None
+                remote.proxy_url = urlunparse(parsed_url._replace(netloc=netloc))
+                remote.save()
+            except Exception:
+                pass
```
rochacbruno, line 36: Python 3,4+ added `suppress` to avoid the use of `pass` in broader exceptions like this. https://docs.python.org/3.6/library/contextlib.html#contextlib.suppress

```suggestion
            with contextlib.suppress(Exception)
                parsed_url = urlparse(remote.proxy_url)
                url = parsed_url.netloc
                netloc = remote.proxy_username + ":" + remote.proxy_password + "@" + url
                remote.proxy_username = None
                remote.proxy_password = None
                remote.proxy_url = urlunparse(parsed_url._replace(netloc=netloc))
                remote.save()
```

pulpcore/app/migrations/0060_data_migration_proxy_creds.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def move_proxy_creds_up(apps, schema_editor):
+    Remote = apps.get_model('core', 'Remote')
+    for remote in Remote.objects.all():
+        if remote.proxy_url and "@" in remote.proxy_url:
+            try:
+                parsed_url = urlparse(remote.proxy_url)
+                netloc = parsed_url.netloc
+                auth, url = netloc.rsplit("@", maxsplit=1)
```
rochacbruno, line 15: on urllib3 (included as dependency of requests) there is a `parse_url` which already splits auth items we used that on galaxy here: https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/common/proxy_url.py#L42

pulpcore/app/serializers/repository.py
```diff
@@ -91,7 +91,17 @@ class RemoteSerializer(ModelSerializer):
         help_text="If True, TLS peer validation must be performed.", required=False
     )
     proxy_url = serializers.CharField(
```
rochacbruno, line 93: Any reason to not use a `URLField` ?

pulpcore/app/serializers/repository.py
```diff
@@ -91,7 +91,17 @@ class RemoteSerializer(ModelSerializer):
...
+        help_text="The proxy URL. Format: scheme://host:port",
+        required=False,
+        allow_null=True,
+    )
+    proxy_username = serializers.CharField(
+        help_text="The username to authenticte to the proxy.",
+        required=False,
+        allow_null=True,
+    )
+    proxy_password = serializers.CharField(
```
rochacbruno, line 103: adding `style={'input_type': 'password'}` makes the generated API UI form to replace chars with `***` on password.

rochacbruno (approved):
LGTM

I left just a couple of suggestions

pulpcore/app/migrations/0060_data_migration_proxy_creds.py
mdellweg: I was about to say, that i tested this migration up and down and i didn't want to touch it anymore if not needed. But i'll take a look. :+1:

pulpcore/app/migrations/0060_data_migration_proxy_creds.py
mdellweg: Interesting!

pulpcore/app/migrations/0060_data_migration_proxy_creds.py
mdellweg: I do not see, that urllib3 is a dependency of pulpcore yet, so i'm going to stick with the standard library here.

pulpcore/app/serializers/repository.py
daviddavis: We've had problems in the past with `URLField`. Not sure if it's still an issue.

https://github.com/pulp/pulpcore/blob/0db2dcbb043cf6b00f93f0b661a25b1ba90fcd97/docs/plugins/reference/how-plugins-work.rst#serializer-and-openapi-schema

mdellweg:
See also: https://github.com/pulp/pulp-cli/pull/169

daviddavis (approved):

bmbouter (approved):
This looks great, thank you!

pulpcore/app/serializers/repository.py
```diff
@@ -92,10 +92,23 @@ class RemoteSerializer(ModelSerializer):
...
-        help_text="The proxy URL. Format: scheme://user:password@host:port",
+        help_text="The proxy URL. Format: scheme://host:port",
         required=False,
         allow_null=True,
     )
+    proxy_username = serializers.CharField(
+        help_text="The username to authenticte to the proxy.",
+        required=False,
+        allow_null=True,
+        write_only=True,
```
rochacbruno, line 103: I think proxy_username and proxy_url was not meant to be `write_only`. Those 2 fields can return to the get request and only `proxy_password` are write only. @mdellweg

mdellweg: `proxy_url` is readable. and with `proxy_username` i followed the example of `username`.

I think @daviddavis started there.

bmbouter: This came from the original story as planned here: https://pulp.plan.io/issues/8202

It's debateable as to whether username and proxy_username are really sensitive, but as it stands now (for better or worse) they are considered sensitive.

@rochacbruno how does this affect your use of pulpcore? Can you share some insight into this debate either way?

rochacbruno: @bmbouter for us it is ok, we can overwrite that property, I am also not sure if we consider usernames sensitive, I personally think that in terms of UX user will want to know which username is in use, what do you think @newswangerd ?

bmbouter: If it's not causing a problem then I'm inclined to leave it redacted because since it's a subjective thing and security related I'd like to default the most secure option. In that case I think, if its in doubt, consider it sensitive. If it causes a practical problem though, let us know for sure.

newswangerd: @rochacbruno for our serializers I think we should leave usernames exposed. Users will want to know which account they've configured for authentication if they ever have to update a token or need to check permissions.

PR #1166, opened 2021-03-03
Disable Dynaconf loading of .env files

rochacbruno:
In production we don't need django to load .env files.

[noissue]


ekohl:
I can only comment for the Foreman/Katello use case, but there I think this only leads to a more predictable setup. We don't use `.env` and it can break setups if users happen to have a `.env` file present without them knowing.

dralley (approved):

mdellweg (approved):
LGTM

goosemania:
@dralley , @mdellweg, please do not approve/merge PRs without a redmine issue if it's not a release PR or a tiny CI change. I believe this issue has been found by a customer, it's good to share it with other users, there are also other reasons to file an issue, like release/backports tracking.

PR #1170, opened 2021-03-04
Revert "Add warning if on-demand content does not have allowed checksum"

daviddavis:
Revert the RemoteArtifact warning for 3.11 as users don't have a way to remedy the problem and also it doesn't handle plugins who can have RemoteArtifact without checksums.


ipanova (approved):

bmbouter (approved):
Thank you!

PR #1173, opened 2021-03-05
Artifact checksums report

pavelpicka:
`pulpcore-manager handle-artifact-checksums --report` reports local and remote artifacts with forbidden checksum type and approx size of it.

re #7986
https://pulp.plan.io/issues/7986

### DIFF
--- a/pulpcore/app/management/commands/handle-artifact-checksums.py
+++ b/pulpcore/app/management/commands/handle-artifact-checksums.py
@@ -4,9 +4,16 @@ from gettext import gettext as _
 
 from django.conf import settings
 from django.core.management import BaseCommand, CommandError
+from django.db.models import Q, Sum
 from pulpcore import constants
 from pulpcore.app import pulp_hashlib
-from pulpcore.plugin.models import Artifact
+from pulpcore.plugin.models import (
+    Artifact,
+    Content,
+    ContentArtifact,
+    RemoteArtifact,
+    RepositoryVersion,
+)
 
 import logging
 
@@ -21,6 +28,96 @@ class Command(BaseCommand):
     """
 
     help = _("Handle missing and forbidden checksums on the artifacts")
+
+    def add_arguments(self, parser):
+        parser.add_argument("--report", action="store_true")
+        parser.add_argument("--checksums", help=_("Comma separated list of checksums to evaluate"))
+
+    def _print_out_repository_version_hrefs(self, repo_versions):
+        """
+        Print out repository version href from a query of repo versions.
+
+        Args:
+            repo_versions (django.db.models.QuerySet): repository versions query
+
+        Returns None
+        """
+        for repo_version in repo_versions.iterator():
+            self.stdout.write(
+                _(
+                    "/repositories/{plugin}/{type}/{pk}/versions/{number}/".format(
+                        plugin=repo_version.repository.pulp_type.split(".")[0],
+                        type=repo_version.repository.pulp_type.split(".")[1],
+                        pk=str(repo_version.repository.pk),
+                        number=repo_version.number,
+                    )
+                )
+            )
+
+    def _show_on_demand_content(self, checksums):
+        query = Q(pk__in=[])
+        for checksum in checksums:
+            query |= Q(**{f"{checksum}__isnull": False})
+
+        remote_artifacts = RemoteArtifact.objects.filter(query).filter(
+            content_artifact__artifact__isnull=True
+        )
+        ras_size = remote_artifacts.aggregate(Sum("size"))["size__sum"]
+
+        content_artifacts = ContentArtifact.objects.filter(remoteartifact__pk__in=remote_artifacts)
+        content = Content.objects.filter(contentartifact__pk__in=content_artifacts)
+        repo_versions = RepositoryVersion.versions_containing_content(content).select_related(
+            "repository"
+        )
+
+        self.stdout.write(
+            _("Found {} on-demand content units with forbidden checksums.").format(content.count())
+        )
+        if content.count():
+            self.stdout.write(
+                _("There is approx {:.2f}Mb of content to be downloaded.").format(
+                    ras_size / (1024 ** 2)
+                )
+            )
+
+        if repo_versions.exists():
+            self.stdout.write(_("\nAffected repository versions with remote content:"))
+            self._print_out_repository_version_hrefs(repo_versions)
+
+    def _show_immediate_content(self, forbidden_checksums):
+        allowed_checksums = set(
+            constants.ALL_KNOWN_CONTENT_CHECKSUMS.symmetric_difference(forbidden_checksums)
+        )
+        query_forbidden = Q()
+        query_required = Q()
+        for checksum in forbidden_checksums:
+            query_forbidden |= Q(**{f"{checksum}__isnull": False})
+
+        for allowed_checksum in allowed_checksums:
+            query_required |= Q(**{f"{allowed_checksum}__isnull": True})
+
+        artifacts = Artifact.objects.filter(query_forbidden | query_required)
+        content_artifacts = ContentArtifact.objects.filter(artifact__in=artifacts)
+        content = Content.objects.filter(contentartifact__pk__in=content_artifacts)
+        repo_versions = RepositoryVersion.versions_containing_content(content).select_related(
+            "repository"
+        )
+
+        self.stdout.write(
+            _("Found {} downloaded content units with forbidden or missing checksums.").format(
+                content.count()
+            )
+        )
+        if content.count() > 0:
+            self.stdout.write(
+                _("There is approx. {:.2f}Mb content data to be re-hashed.").format(
+                    artifacts.aggregate(Sum("size"))["size__sum"] / (1024 ** 2)
+                )
+            )
+
+        if repo_versions.exists():
+            self.stdout.write(_("\nAffected repository versions with present content:"))
+            self._print_out_repository_version_hrefs(repo_versions)
 
     def _download_artifact(self, artifact, checksum, file_path):
         restored = False
@@ -44,7 +141,36 @@ class Command(BaseCommand):
                 break
         return restored
 
+    def _report(self, allowed_checksums):
+        if allowed_checksums:
+            allowed_checksums = allowed_checksums.split(",")
+            if "sha256" not in allowed_checksums:
+                raise CommandError(_("Checksums must contain sha256"))
+        else:
+            allowed_checksums = settings.ALLOWED_CONTENT_CHECKSUMS
+
+        forbidden_checksums = set(constants.ALL_KNOWN_CONTENT_CHECKSUMS).difference(
+            allowed_checksums
+        )
+
+        self.stderr.write(
+            _(
+                "Warning: the handle-artifact-checksums report is in "
+                "tech preview and may change in the future."
+            )
+        )
+        self._show_on_demand_content(forbidden_checksums)
+        self._show_immediate_content(forbidden_checksums)
+
     def handle(self, *args, **options):
+
+        if options["report"]:
+            self._report(options["checksums"])
+            return
+
+        if options["checksums"] and not options["report"]:
+            self.stdout.write(_("Checksums cannot be supplied without --report argument"))
+            exit(1)
 
         log.setLevel(logging.ERROR)
         hrefs = set()
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -468,6 +468,32 @@ class RepositoryVersion(BaseModel):
         get_latest_by = "number"
         ordering = ("number",)
 
+    @classmethod
+    def versions_containing_content(cls, content):
+        """
+        Returns a query of repository versions containing the provided content units.
+
+        Args:
+            content (django.db.models.QuerySet): query of content
+
+        Returns:
+            django.db.models.QuerySet: Repository versions which contains content.
+        """
+        query = models.Q(pk__in=[])
+        repo_content = RepositoryContent.objects.filter(content__pk__in=content)
+
+        for rc in repo_content.iterator():
+            filter = models.Q(
+                repository__pk=rc.repository.pk,
+                number__gte=rc.version_added.number,
+            )
+            if rc.version_removed:
+                filter &= models.Q(number__lt=rc.version_removed.number)
+
+            query |= filter
+
+        return cls.objects.filter(query)
+
     def _content_relationships(self):
         """
         Returns a set of repository_content for a repository version
--- a/CHANGES/7986.feature
+++ b/CHANGES/7986.feature
@@ -1,0 +1,1 @@
+User can evaluate how many content units are affected with checksum type change with 'pulpcore-manager handle-artifact-checksums --report'.
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -271,6 +271,17 @@ ALLOWED_CONTENT_CHECKSUMS
     ``sha384``, and ``sha512``. After modifying this setting, you likely will need to run
     ``pulpcore-manager handle-artifact-checksums`` or Pulp will refuse to start.
 
+    You can run ``pulpcore-manager handle-artifact-checksums --report`` to find out
+    how many content units are affected with actual ``ALLOWED_CONTENT_CHECKSUMS`` settings.
+    Also, you can run ``pulpcore-manager handle-artifact-checksums --report`` with ``--checksums``
+    argument specifying comma separated list of checksums you want to use to perform this check
+    before changing the setting.
+
+    .. warning::
+      ``--report`` and ``--checksums`` arguments are tech-preview and may change in backwards
+      incompatible ways in future releases.
+
+
     .. warning::
       Due to its use as a primary content-identifier, "sha256"" **IS REQUIRED**. Pulp will
       fail to start if it is not found in this set.


pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -15,13 +16,77 @@
...
+    ras_size = sum([ra.size for ra in remote_artifacts if ra.size])
+
+    content = (
+        ContentArtifact.objects.filter(remoteartifact__pk__in=remote_artifacts)
+        .filter(artifact=None)
+        .values_list("content", flat=True)
+    )
+    repo_versions = RepositoryVersion.has_content(content)
+
+    print(_("Found {} on-demand content units with forbidden checksums.").format(content.count()))
```
daviddavis, line 34: Need to a replace of `print` with `self.stdout.write`.

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -15,13 +16,77 @@
 CHUNK_SIZE = 1024 * 1024  # 1 Mb
 
 
+def _show_on_demand_content(checksums):
+    query = Q(pk__in=[])
+    for checksum in checksums:
+        query |= Q(**{f"{checksum}__isnull": False})
+
+    remote_artifacts = RemoteArtifact.objects.filter(query)
+    ras_size = sum([ra.size for ra in remote_artifacts if ra.size])
```
daviddavis, line 25: This size is not going to be accurate since it's not filtering by artifact=None.

What you could do is change the previous line to:

```
RemoteArtifact.objects.filter(query).filter(content_artifact__artifact__isnull=True)
```

You would drop line 29 then.

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -44,8 +109,29 @@ def _download_artifact(self, artifact, checksum, file_path):
...
+        if checksums:
+            checksums = checksums.split(",")
+            if "sha256" not in checksums:
+                raise CommandError(_("Checksums must contain sha256"))
+        else:
+            checksums = settings.ALLOWED_CONTENT_CHECKSUMS
+
+        forbidden_checksums = set(constants.ALL_KNOWN_CONTENT_CHECKSUMS).difference(checksums)
+
+        _show_on_demand_content(forbidden_checksums)
```
daviddavis, line 122: I'd probably stick in a tech preview warning for now especially if this goes out in 3.11:

```suggestion
        self.stderr.write(_("Warning: the handle-artifact-checksums report is in tech preview and may change in the future."))
        _show_on_demand_content(forbidden_checksums)
```

daviddavis:
Looking good. Will test it out later today.

pulpcore/app/management/commands/handle-artifact-checksums.py
pavelpicka: changed and moved those two functions under the Command class to make use of 'self' sense.

pulpcore/app/management/commands/handle-artifact-checksums.py
pavelpicka: I agree, added.

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -44,8 +117,35 @@ def _download_artifact(self, artifact, checksum, file_path):
...
+
+        forbidden_checksums = set(constants.ALL_KNOWN_CONTENT_CHECKSUMS).difference(checksums)
+
+        self._show_on_demand_content(forbidden_checksums)
+        self.stderr.write(
+            _(
+                "Warning: the handle-artifact-checksums report is in "
+                "tech preview and may change in the future."
+            )
+        )
```
daviddavis, line 136: This needs to go before `_show_on_demand_content`

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -22,6 +23,78 @@ class Command(BaseCommand):
...
+        query_forbidden = Q()
+        query_required = Q()
+        for checksum in checksums:
+            query_forbidden |= Q(**{f"{checksum}__isnull": False})
+
+        for allowed_checksum in allowed_checksums:
+            query_required |= Q(**{f"{allowed_checksum}__isnull": True})
+
+        artifacts = Artifact.objects.filter(query_forbidden & query_required)
```
daviddavis, line 72: I tested this out and I didn't get any results when I had artifacts with forbidden checksums. I think you want to query for artifacts that have forbidden checksums OR missing checksums.

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -22,6 +23,78 @@ class Command(BaseCommand):
...
+        content = ContentArtifact.objects.filter(artifact__in=artifacts).values_list(
+            "content", flat=True
+        )
+
+        repo_versions = RepositoryVersion.has_content(content)
+
+        self.stdout.write(
+            _("Found {} already downloaded content units with forbidden checksums.").format(
```
daviddavis, line 80:
```suggestion
            _("Found {} already downloaded content units with forbidden or missing checksums.").format(
```

daviddavis:
It looks like this is missing the `has_content()` change from the PoC:

```
AttributeError: type object 'RepositoryVersion' has no attribute 'has_content'
```

Bonus points if you write some unit tests for this method.

daviddavis:
Overall, this looks great to me. :+1:  We may have to tweak the criteria around reporting on-demand content but I think with the tech preview label, we can merge this once the issues I pointed out are fixed.

@bmbouter @ipanova maybe you too can review as well?

ipanova:
> Overall, this looks great to me. +1 We may have to tweak the criteria around reporting on-demand content but I think with the tech preview label, we can merge this once the issues I pointed out are fixed.
> 
> @bmbouter @ipanova maybe you too can review as well?

yes, i can get to this on monday.

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -44,8 +117,35 @@ def _download_artifact(self, artifact, checksum, file_path):
...
+            if "sha256" not in checksums:
+                raise CommandError(_("Checksums must contain sha256"))
+        else:
+            checksums = settings.ALLOWED_CONTENT_CHECKSUMS
+
+        forbidden_checksums = set(constants.ALL_KNOWN_CONTENT_CHECKSUMS).difference(checksums)
+
+        self.stderr.write(
+            _(
+                "Warning: the handle-artifact-checksums report is in "
```
ipanova, line 132: if we claim this command is in tech-preview then we should revert this PR https://github.com/pulp/pulpcore/pull/1153

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -22,6 +23,78 @@ class Command(BaseCommand):
 
     help = _("Handle missing and forbidden checksums on the artifacts")
 
+    def add_arguments(self, parser):
+        parser.add_argument("--report", action="store_true")
```
ipanova, line 27: can we add docs about this new functionality of the `handle-artifacts-checksums` command?

pulpcore/app/management/commands/handle-artifact-checksums.py
pavelpicka: I believe only '--report' part should be still in tech-preview. Actual command works ok and will stay, we may change the report part in the future.

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -22,6 +23,78 @@ class Command(BaseCommand):
...
+        if content.count():
+            self.stdout.write(
+                _("There is approx {:.2f}Mb of content to be downloaded.").format(ras_size / 1024)
+            )
+
+        if repo_versions.exists():
+            self.stdout.write(_("\nAffected repository versions with remote content:"))
+            for repo_version in repo_versions:
+                self.stdout.write(
+                    _("{} (version {})").format(repo_version.repository.name, repo_version.number)
```
ipanova, line 57: won't it be more useful to directly provide the href for the affected repo_version?

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -44,8 +117,35 @@ def _download_artifact(self, artifact, checksum, file_path):
...
 
+    def _report(self, checksums):
+        if checksums:
+            checksums = checksums.split(",")
+            if "sha256" not in checksums:
+                raise CommandError(_("Checksums must contain sha256"))
+        else:
+            checksums = settings.ALLOWED_CONTENT_CHECKSUMS
+
+        forbidden_checksums = set(constants.ALL_KNOWN_CONTENT_CHECKSUMS).difference(checksums)
```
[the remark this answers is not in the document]
ipanova: actually, disregard this comment - have not noticed that you are using the `checksums` variable here

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -22,6 +23,78 @@ class Command(BaseCommand):
...
+        repo_versions = RepositoryVersion.has_content(content)
+
+        self.stdout.write(
+            _("Found {} downloaded content units with forbidden or missing checksums.").format(
+                content.count()
+            )
+        )
+        if content.count() > 0:
+            self.stdout.write(
+                _("There is approx. {:.2f}Mb content data to be re-hash.").format(
```
ipanova, line 86: s/re-hash/re-hashed maybe?

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -22,6 +23,78 @@ class Command(BaseCommand):
...
+
+    def _show_on_demand_content(self, checksums):
+        query = Q(pk__in=[])
+        for checksum in checksums:
+            query |= Q(**{f"{checksum}__isnull": False})
+
+        remote_artifacts = RemoteArtifact.objects.filter(query).filter(
+            content_artifact__artifact__isnull=True
+        )
+        ras_size = sum([ra.size for ra in remote_artifacts if ra.size])
```
ipanova, line 38: i wonder if instead of iterating like this we should not user remote_artifacts
.values_list('size', flat=True)

ipanova:
I did some testing and it seems to report correct thing.

pulpcore/app/management/commands/handle-artifact-checksums.py
dralley: ```remote_artifacts.aggregate(Sum('size'))```

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -22,6 +23,78 @@ class Command(BaseCommand):
...
+
+        content = ContentArtifact.objects.filter(
+            remoteartifact__pk__in=remote_artifacts
+        ).values_list("content", flat=True)
+        repo_versions = RepositoryVersion.has_content(content)
```
dralley, line 43: Does this not take a queryset?

pulpcore/app/models/repository.py
```diff
@@ -468,6 +468,23 @@ class Meta:
         get_latest_by = "number"
         ordering = ("number",)
 
+    @classmethod
+    def has_content(cls, content):
```
dralley, line 472: This definitely needs documentation, it's not immediately clear what is going on here.

pulpcore/app/management/commands/handle-artifact-checksums.py
pavelpicka: good idea, but I think getting pulp_type of each repository to return right href (e.g. repositories/`type/type`/uuid) can slow the process

pulpcore/app/management/commands/handle-artifact-checksums.py
pavelpicka: Now the function `has_content()` works with queryset which should help efficiency.

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -22,6 +23,74 @@ class Command(BaseCommand):
...
+        allowed_checksums = set(
+            constants.ALL_KNOWN_CONTENT_CHECKSUMS.symmetric_difference(checksums)
+        )
+        query_forbidden = Q()
+        query_required = Q()
+        for checksum in checksums:
+            query_forbidden |= Q(**{f"{checksum}__isnull": False})
+
+        for allowed_checksum in allowed_checksums:
+            query_required |= Q(**{f"{allowed_checksum}__isnull": True})
```
dralley, line 68: While the normal advice would be to use less queries, in this case, I think it would actually be more efficient to do one query per checksum type, and add the PK's to a set.

Databases are bad with dealing with a lot of OR operations, and I'm not sure if it's able to use the index effectively, as opposed to dealing with one checksum type at a time and being able to do a very fast scan down a single index.

I used that technique [here](https://github.com/pulp/pulpcore/pull/751/) on the artifact stage and it sped up the entire QueryExistingArtifacts stage by more than 30%.

But, this isn't something a user is going to run often, maybe only once or twice, so I don't know that we actually need to worry about it?  I think I would be fine with leaving it as-is, @ipanova what do you think?

Just wanted to share the general idea in case it's useful knowledge in the future.

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -22,6 +23,74 @@ class Command(BaseCommand):
...
+            )
+
+        if repo_versions.exists():
+            self.stdout.write(_("\nAffected repository versions with remote content:"))
+            for repo_version in repo_versions:
+                self.stdout.write(
+                    _("{} (version {})").format(repo_version.repository.name, repo_version.number)
+                )
+
+    def _show_immediate_content(self, checksums):
```
dralley, line 58: Here and in the other method above, ```checksums``` is the set of forbidden checksums, but below on the ``_report()``` method ```checksums``` is the set of allowed checksums.  We should be clearer with the naming, let's use ```forbidden_checksums``` and ```allowed_checksums``` consistently.

dralley: It's fine if we use just ```checksums``` where we interact with the user, I only mean inside our own code.

pulpcore/app/models/repository.py
```diff
@@ -468,6 +468,34 @@ class Meta:
         get_latest_by = "number"
         ordering = ("number",)
 
+    @classmethod
+    def has_content(cls, ca_query):
```
dralley, line 472: Passing ContentArtifact querysets isn't something we do anywhere else in the API that I know of, I would prefer that we just use a Content queryset like most other places.  Line 484 can just be moved outside of the method.

I'm not sure about the name, it sounds (to me) too much like a method you would call to see if a single ```RepositoryVersion``` contains/has some content.  It might be better if we named it something like ```RepositoryVersion.versions_containing_content()```.

pulpcore/app/models/repository.py
```diff
@@ -468,6 +468,34 @@ class Meta:
...
+            ca_query (django.db.models.QuerySet): query of content_artifacts
+
+        Returns:
+            django.db.models.QuerySet: Repository versions which contains content.
+        """
+        query = models.Q(pk__in=[])
+        content = Content.objects.filter(contentartifact__pk__in=ca_query)
+        repo_content = RepositoryContent.objects.filter(content__pk__in=content)
+
+        for rc in repo_content:
```
dralley, line 487: ```.iterator()```

pulpcore/app/management/commands/handle-artifact-checksums.py
ipanova: I agree that it will be slower, but this is already a report for 'what-if' scenario that can take some time because of all the calculations. The question is - what is the purpose of this information? Just to display to the user the amount affected repo versions? Then we can just leave what it is in there now or even do  `len(repo_versions)`. If our goal is to display this information in a usable format so the user would do something about it then we should display the `href` right away so he can directly access  it instead of figuring out what is the href for repo 'foo' and version X

pulpcore/app/management/commands/handle-artifact-checksums.py
ipanova: @dralley your observations are very correct. Because this probably would be ran once I think we can leave this as is.

pulpcore/app/management/commands/handle-artifact-checksums.py
ipanova: +1 this improves the code readability

pulpcore/app/models/repository.py
ipanova: `versions_contain_content`?

pulpcore/app/management/commands/handle-artifact-checksums.py
dralley: Maybe ```repo_versions = RepositoryVersion.has_content(content).select_related('repository')``` would make it cheap enough?  We could copy and paste the href generation code to make sure that we're doing it in the most efficient way, without extra queries.  It's not that much logic, something like 

```python
"/repositories/{plugin}/{type}/{pk}/{number}/".format(
    plugin=rv.repository.get_pulp_type().split('.')[0],
    type=rv.repository.get_pulp_type().split('.')[1],
    pk=str(rv.repository.pk),
    number=rv.number,
)

We need a ```.iterator()``` here also.

File: pulpcore/app/management/commands/handle-artifact-checksums.py [hunk 19]
```diff
@@ -22,6 +29,93 @@ class Command(BaseCommand):
...
+                    artifacts.aggregate(Sum("size"))["size__sum"] / 1024
+                )
+            )
+
+        if repo_versions.exists():
+            self.stdout.write(_("\nAffected repository versions with present content:"))
+            for repo_version in repo_versions.iterator():
+                self.stdout.write(
+                    _(
+                        "/repositories/{plugin}/{type}/{pk}/{number}/".format(
```
dralley, line 110: Put this in a helper method so we don't copy it twice.

pulpcore/app/models/repository.py
```diff
@@ -468,6 +468,33 @@ class Meta:
         get_latest_by = "number"
         ordering = ("number",)
 
+    @classmethod
+    def versions_contain_content(cls, content):
+        """
+        Returns a query of repository versions,
+        which contains content units derived from content_artifacts.
```
dralley, line 475:
```
   def versions_containing_content(cls, content):
        """
        Returns a query of repository versions containing the provided content units.
        """
```

pavelpicka: I still leave 'args and returns' parts of doc string in place.

pulpcore/app/models/repository.py
```diff
@@ -468,6 +468,32 @@ class Meta:
         get_latest_by = "number"
         ordering = ("number",)
 
+    @classmethod
+    def versions_contain_content(cls, content):
```
dralley, line 472: I think this needs to be ```versions_containing_content```

pavelpicka: substituted

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -22,6 +29,94 @@ class Command(BaseCommand):
...
+    def _print_out_repository_version_hrefs(self, repo_versions):
+        """
+        Print out repository version href from a query of repo versions.
+
+        Args:
+            repo_versions (django.db.models.QuerySet): repository versions query
+
+        Returns None
+        """
+        for repo_version in repo_versions.iterator():
```
ipanova, line 45: this produces wrong url. you need to call cast() on the repo and also add the `versions` part to the url

```
Affected repository versions with present content:
/repositories/core/repository/58c1c1a0-4b6a-4140-a959-4bbbe81245aa/1/
```

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -22,6 +29,94 @@ class Command(BaseCommand):
...
+            "repository"
+        )
+
+        self.stdout.write(
+            _("Found {} on-demand content units with forbidden checksums.").format(content.count())
+        )
+        if content.count():
+            self.stdout.write(
+                _("There is approx {:.2f}Mb of content to be downloaded.").format(ras_size / 1024)
+            )
```
ipanova, line 79: correct my math here but we are storing the size in bytes here :)

pulpcore/app/management/commands/handle-artifact-checksums.py
dralley: My bad, I think the correct thing to use would have been ```repository.pulp_type``` rather than ```repository.get_pulp_type()```.

I *think* that should make casting unnecessary

pulpcore/app/management/commands/handle-artifact-checksums.py
pavelpicka: you are right, we store bytes not kb, fixed my math

pulpcore/app/management/commands/handle-artifact-checksums.py
ipanova: @dralley i can confirm repository.pulp_type works.

pavelpicka: Used ``.pulp_type`` instead casting.

ipanova (approved):
Thanks!

dralley (approved):

PR #1178, opened 2021-03-11
Make it easier to change our APIs in the future without breaking compat

dralley:
[noissue]

### DIFF
--- a/pulpcore/app/models/publication.py
+++ b/pulpcore/app/models/publication.py
@@ -47,7 +47,7 @@ class Publication(MasterModel):
     repository_version = models.ForeignKey("RepositoryVersion", on_delete=models.CASCADE)
 
     @classmethod
-    def create(cls, repository_version, pass_through=False):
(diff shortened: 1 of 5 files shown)


mdellweg:
I am not sure, this is a good idea. This way, when a plugin "uses" a new parameter, it would silently be ignored by pulpcore. So whatever the plugin expects that parameter do accomplish will be missed.
I would expect, that new parameters (along with a new feature) will be added with a default value, so plugins do not need to know about it initially. But as they start using it, they'd need to depend on a pulpcore version having it anyway.
Am i missing something here?

dralley:
This makes it possible for an older pulpcore version to remain compatible with a newer plugin, as long as the plugin writer is careful about their intent (passing args / kwargs into a method that function that can't do anything with them).

mdellweg:
I am having a hard time picturing a plugin to adopt a feature that might or might not be there in pulpcore.
Maybe i'm missing a catchy example.

PR #1183, opened 2021-03-12
Make 3.7 branch to use GHA

goosemania:
[noissue]


.github/workflows/scripts/script.sh
```diff
@@ -4,35 +4,49 @@
...
-if [ "$TEST" = "docs" ]; then
+if [[ "$TEST" = "docs" || "$TEST" = "publish" ]]; then
   cd docs
   make PULP_URL="http://pulp" html
   cd ..
 
+  echo "Validating OpenAPI schema..."
+  cat $PWD/.ci/scripts/schema.py | cmd_stdin_prefix bash -c "cat > /tmp/schema.py"
+  cmd_prefix bash -c "python3 /tmp/schema.py"
+  # cmd_prefix bash -c "pulpcore-manager spectacular --file pulp_schema.yml --validate"
```
fao89, line 37: I can see you got the right version

fao89 (approved):
the 2 things I remembered could break the CI were the OpenAPI test and the app label change, both were right here, not sure if have other things that may break a release

goosemania:
@fao89 , it could also be the docs script publishing in the wrong place.
But it seems that this one should work. It worked for the container plugin. Codepaths are a bit different for pulpcore but I think it should be fine.

fao89:
https://github.com/pulp/pulpcore/blob/master/functest_requirements.txt#L6

goosemania:
> https://github.com/pulp/pulpcore/blob/master/functest_requirements.txt#L6

test_bindings.* were missing as well. We'll see what now :)

mdellweg (approved):

PR #1194, opened 2021-03-19
Fixed azure artifacts path download.

ipanova:
closes #8427
https://pulp.plan.io/issues/8427

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -594,14 +594,15 @@ class Handler:
         if settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem":
             filename = content_artifact.artifact.file.name
             return FileResponse(os.path.join(settings.MEDIA_ROOT, filename), headers=headers)
-        elif (
-            settings.DEFAULT_FILE_STORAGE == "storages.backends.s3boto3.S3Boto3Storage"
-            or settings.DEFAULT_FILE_STORAGE == "storages.backends.azure_storage.AzureStorage"
-        ):
+        elif settings.DEFAULT_FILE_STORAGE == "storages.backends.s3boto3.S3Boto3Storage":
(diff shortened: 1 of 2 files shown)


pulpcore/content/handler.py
```diff
@@ -594,15 +594,16 @@ def _serve_content_artifact(self, content_artifact, headers):
...
-            settings.DEFAULT_FILE_STORAGE == "storages.backends.s3boto3.S3Boto3Storage"
-            or settings.DEFAULT_FILE_STORAGE == "storages.backends.azure_storage.AzureStorage"
-        ):
+        elif settings.DEFAULT_FILE_STORAGE == "storages.backends.s3boto3.S3Boto3Storage":
             artifact_file = content_artifact.artifact.file
             content_disposition = f"attachment;filename={content_artifact.relative_path}"
             parameters = {"ResponseContentDisposition": content_disposition}
             url = artifact_file.storage.url(artifact_file.name, parameters=parameters)
             raise HTTPFound(url)
+        elif settings.DEFAULT_FILE_STORAGE == "storages.backends.azure_storage.AzureStorage":
```
ipanova, line 603: @daviddavis should not we also set content-disposition header for Azure artifacts? 
https://docs.microsoft.com/en-us/rest/api/storageservices/set-blob-properties#request-headers-all-blob-types

daviddavis: I would think so but I am not familiar with Azure and I haven't tested with it at all.

ipanova: I agree, I am not familiar with Azure either. I wanted to finalize this community contribution. Can we merge this as-is and add/fix content-disposition support later on whenever we receive a bug?Thoughts?

daviddavis (approved):

goosemania (approved):
Thanks!

PR #1198, opened 2021-03-19
Adds new `Distribution` MasterModel

bmbouter:
Added the following new objects related to a new ``Distribution``
MasterModel:

* ``pulpcore.plugin.models.Distribution`` - A new MasterModel
  ``Distribution`` which replaces the
  ``pulpcore.plugin.models.BaseDistribution``. This now contains the
  ``repository``, ``repository_version``, and ``publication`` fields on
  the MasterModel instead of on the detail models as was done with
  ``pulpcore.plugin.models.BaseDistribution``.

* ``pulpcore.plugin.serializer.DistributionSerializer`` - A serializer
  plugin writers should use with the new
  ``pulpcore.plugin.models.Distribution``.

* ``pulpcore.plugin.viewset.DistributionViewSet`` - The viewset that
  replaces the deprecated
  ``pulpcore.plugin.viewset.BaseDistributionViewSet``.

* ``pulpcore.plugin.viewset.NewDistributionFilter`` - The filter that
  pairs with the ``Distribution`` model.

closes #8384


pulpcore/app/serializers/publication.py
```diff
@@ -129,32 +159,106 @@ class Meta:
...
+        elif repository_in_data and repository_version_in_instance:
+            error = True
+        elif repository_in_instance and repository_version_in_data:
+            error = True
+        else:
+            error = False
+
+        if error:
+            msg = _(
+                "The attributes 'repository' and 'repository_version' must be used exclusively."
```
dralley, line 257: This doesn't necessarily need to be true.  We could have the same semantics as publication-based distributions, where one particular {publication | repository_version} is being pointed to, but the existence of "repository" or not tells it whether or not to auto-distribute.

I'm not opposed to this but we should have a good justification for the different semantics.

pulpcore/app/serializers/publication.py
```diff
@@ -75,7 +75,37 @@ class Meta:
...
+        # look for any base paths nested in path
+        search = path.split("/")[0]
+        q = Q(base_path=search)
+        for subdir in path.split("/")[1:]:
+            search = "/".join((search, subdir))
+            q |= Q(base_path=search)
+
+        # look for any base paths that nest path
+        q |= Q(base_path__startswith="{}/".format(path))
+        qs = models.BaseDistribution.objects.filter(q)
```
mdellweg, line 91: I am missing `models.Distribution` here for symmetry.
Maybe we could make it `self` and `the_other` where `the_other` is specified in the `Meta` class. But it may be prevented by circular dependencies.

bmbouter:
I added the overlapping cross-model base path support. I'm hand-testing it because pulpcore's CI doesn't have two plugins with heterogeneous Distribution types. This code isn't going to be around long enough to justify adding two plugins for dedicated tests. So what I did was:

1) Install pulp_file with this PR checked out:  
2) Made a pulp_file distribution with `pulp file distribution create --name mydistro --base-path foo`
3) Created an ansible repo (needed to create an ansible distribution). `pulp ansible repository create --name ansiblerepo`
4. Created a distribution and correctly received the overlapping error cross-model.

```
pulp ansible distribution create --name asdf --base-path foo/bar --repository ansiblerepo
Error: {"base_path":["Overlaps with existing distribution 'mydistro'"]}
```

bmbouter:
I added the overlapping cross-model name uniqueness support. I'm hand-testing it because pulpcore's CI doesn't have two plugins with heterogeneous Distribution types. This code isn't going to be around long enough to justify adding two plugins for dedicated tests. So what I did was:

1) Install pulp_file with this PR checked out:  https://github.com/pulp/pulp_file/pull/481
2) Made a pulp_file distribution with `pulp file distribution create --name mydistro --base-path foo`
3) Created an ansible repo (needed to create an ansible distribution). `pulp ansible repository create --name ansiblerepo`
4. Created a distribution and correctly received the overlapping error cross-model.

```
pulp ansible distribution create --name mydistro --base-path another --repository ansiblerepo
Error: {"name":["This field must be unique."]}
```

pulpcore/app/models/publication.py
```diff
@@ -311,6 +311,71 @@ def content_handler_list_directory(self, rel_path):
         return set()
 
 
+class Distribution(MasterModel):
+    """
+    A Distribution defines how a publication is distributed by the Content App.
```
dralley, line 316: Or repository version

pulpcore/app/models/publication.py
```diff
@@ -311,6 +311,71 @@ def content_handler_list_directory(self, rel_path):
...
+
+    The ``name`` must be unique.
+
+    The ``base_path`` must have no overlapping components. So if a Distribution with ``base_path``
+    of ``a/path/foo`` existed, you could not make a second Distribution with a ``base_path`` of
+    ``a/path`` or ``a`` because both are subpaths of ``a/path/foo``.
+
+    Subclasses are expected to use either the ``publication`` or ``repository_version`` field, but
+    not both. The content app that serves content is not prepared to serve content both ways at the
+    same time.
```
dralley, line 328: I wish we could unify this either by attaching publication to repository version, or using publication.pass_through for "RepositoryVersionDistributions"

pulpcore/app/serializers/publication.py
```diff
@@ -75,7 +75,37 @@ class Meta:
         fields = ModelSerializer.Meta.fields + ("name", "description")
 
 
-class BaseDistributionSerializer(ModelSerializer):
+class BasePathOverlapMixin:
```
dralley, line 78: Could we add a  TODO: remove in version X  comment?

mdellweg: And file a removal ticket for the corresponding milestone.

pulpcore/app/serializers/publication.py
```diff
@@ -127,34 +239,56 @@ class Meta:
...
+        required=False,
+        help_text=_("Publication to be served"),
+        view_name_pattern=r"publications(-.*/.*)?-detail",
+        queryset=models.Publication.objects.exclude(complete=False),
+        allow_null=True,
+    )
+
+
+class DistributionRepositoryVersionSerializerMixin:
+    repository = DetailRelatedField(
```
bmbouter, line 282: I think `repository` needs to either be copied to `DistributionPublicationSerializerMixin` or it needs to become its own mixin. @daviddavis @dralley @mdellweg wdyt?

pulpcore/app/serializers/publication.py
bmbouter: This was resolved in a later commit so I'm marking as resolved. Please re-comment if its not the case.

pulpcore/app/models/publication.py
bmbouter: fixed in next push, ty.

pulpcore/app/serializers/publication.py
bmbouter: To address this: I put the details on the removal ticket here: https://pulp.plan.io/issues/8386

pulpcore/app/serializers/publication.py
dralley: Why should it be copied if it's now present and usable by both?  I would think it would be part of the standard distribution serializer.

daviddavis: Won't the help text be slightly different for publication distributions vs repo version distributions?

dralley: I don't think it necessarily has to be.  Something like ```The latest version of this repository will be distributed automatically``` is still true even if there's a publication step that happens in the middle.

bmbouter: I'm moving `repository` to `DistributionSerializer` itself. Also I have to remove these Mixins because the tests point out that we can't create a distribution that has a publication due to the error: `{"publication":["Invalid hyperlink - Incorrect URL match."]}`.

To work around that I tried setting:
* `view_name_pattern=r"publication(-.*/.*)?-detail",`
* `view_name_pattern=r"publications(-.*/.*)?-detail",`
* `view_name_pattern="publication-detail",`

None of ^ worked.

dralley: What is the consequence of removing the mixins, unnecessary fields?

bmbouter: Plugins will have to copy the block now shown in the docstring of `DistributionSerializer`. We can't provide a de-duplicated copy.

pulpcore/app/serializers/publication.py
```diff
@@ -115,8 +147,96 @@ class BaseDistributionSerializer(ModelSerializer):
...
+
+    pulp_href = DetailIdentityField(view_name_pattern=r"distributions(-.*/.*)-detail")
+    pulp_labels = LabelsField(required=False)
+
+    base_path = serializers.CharField(
+        help_text=_(
+            'The base (relative) path component of the published url. Avoid paths that \
+                    overlap with other distribution base paths (e.g. "foo" and "foo/bar")'
+        ),
         validators=[UniqueValidator(queryset=models.BaseDistribution.objects.all())],
```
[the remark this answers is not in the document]
daviddavis: That makes sense to me. I think we can just remove it.

daviddavis (approved):
Looks good to me.

pulpcore/app/serializers/publication.py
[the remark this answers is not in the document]
bmbouter: I removed it and pushed.

pulpcore/app/models/publication.py
bmbouter: It would be nice to unify, but some plugins don't need a publication at all so I'm not sure creating an object would make sense to them. I see where you're coming from though. Either way maybe we can still unify in the future somehow.

pulpcore/app/serializers/publication.py
bmbouter: Yeah this was a copy/paste from what we had. The thing really is that we can't have a distribution point to both a `repository` and a `publication` at the same time. Let me try to work on this validation to make that true instead and tell me what you think.

pulpcore/app/serializers/publication.py
```diff
@@ -129,32 +164,115 @@ class Meta:
...
+            "content_guard",
+            "pulp_labels",
+            "name",
+            "repository",
+        )
+
+    def validate(self, data):
+        super().validate(data)
+
+        publication_in_data = "publication" in data
```
bmbouter, line 255: @dralley check this portion out, it's revised.

pulpcore/app/serializers/publication.py
```diff
@@ -129,32 +164,115 @@ class Meta:
...
+        elif publication_in_data and repository_version_in_instance:
+            error = True
+        elif publication_in_instance and repository_version_in_data:
+            error = True
+        else:
+            error = False
+
+        if error:
+            msg = _(
+                "The attributes 'publication' and 'repository_version' must be used exclusively."
```
dralley, line 271: This is very slightly unclear I think.  Maybe say something like 

> Only one of the attributes 'publication' and 'repository_version' may be used at a time.

pulpcore/app/serializers/publication.py
```diff
@@ -116,9 +148,23 @@ class BaseDistributionSerializer(ModelSerializer):
...
+            UniqueValidator(queryset=models.Distribution.objects.all()),
+        ],
     )
 
+    def __init__(self, *args, **kwargs):
+        """ Initialize a BaseDistributionSerializer and emit DeprecationWarnings"""
+        warnings.warn(
+            _(
+                "BaseDistributionSerializer is deprecated and could be removed as early as "
+                "pulpcore==3.13; use pulpcore.plugin.serializers.DistributionSerializer instead."
```
dralley, line 162: I would almost suggest putting a link to the Pulp File PR in here since it is a fairly "involved" migration process.

dralley (approved):
See suggestion, otherwise LGTM

pulpcore/app/serializers/publication.py
bmbouter: Actually one of the other ones calls out the migration specifically. I'm going to merge for now if that's alright.

PR #1200, opened 2021-03-22
Wrap Exporter update and destroy calls in tasks

mdellweg:
Exporter is a resource that needs to be reserved to change it. Running
an export task, while the exporter is being changed can lead to problems
otherwise.

fixes #7438

### DIFF
--- a/CHANGES/7438.bugfix
+++ b/CHANGES/7438.bugfix
@@ -1,0 +1,1 @@
+Added asynchronous tasking to the Update and Delete endpoints of PulpExporter to provide proper locking on resources.
--- a/CHANGES/7438.removal
+++ b/CHANGES/7438.removal
@@ -1,0 +1,1 @@
+The Update and Delete endpoints of Exporters changed to now return 202 with tasks.
--- a/pulpcore/app/viewsets/exporter.py
+++ b/pulpcore/app/viewsets/exporter.py
@@ -21,6 +21,8 @@ from pulpcore.app.serializers import (
 from pulpcore.app.tasks.export import pulp_export
 
 from pulpcore.app.viewsets import (
+    AsyncRemoveMixin,
+    AsyncUpdateMixin,
     BaseFilterSet,
     NamedModelViewSet,
 )
@@ -51,10 +53,10 @@ class ExporterFilter(BaseFilterSet):
 class ExporterViewSet(
     NamedModelViewSet,
     mixins.CreateModelMixin,
-    mixins.UpdateModelMixin,
+    AsyncUpdateMixin,
     mixins.RetrieveModelMixin,
     mixins.ListModelMixin,
-    mixins.DestroyModelMixin,
+    AsyncRemoveMixin,
 ):
     """
     ViewSet for viewing exporters.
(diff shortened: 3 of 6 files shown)


daviddavis:
This code looks good to me. I believe @ggainey was going to double check the code to confirm we needed this.

daviddavis (approved):

mdellweg:
Actually, we have a real problem here. The change changes the pulp_file api schema in a way to make the pulp_file_client package incompatible.

CHANGES/7438.bugfix
```diff
@@ -0,0 +1 @@
+Added asynchronous tasking to the Update and Delete endpoints of PulpExporter to provide proper locking on resources.
```
ipanova, line 1: i think this maybe should be a .feature. This is also a change in behavior

ggainey:
> This code looks good to me. I believe @ggainey was going to double check the code to confirm we needed this.

I've been discussing with jsherrill about katello's use. We can either do it this way and let resource-manager dispatch when it can, *or* we can continue to have this be a synchornous call that checks for "is this resource reserved right now" and return an error immediately if so.  

This PR does it the 'standard' way - but, as Matthias points out, it's a significant API change. This particular change is the main reason imp/exp is still in tech-preview, so making breaking changes is still allowed - but that doesn't make them less painful.

CHANGES/7438.bugfix
ggainey: It's not a new feature even if it is an API change. If this is a feature, then any API change in a tech-preview function would be a feature - and I don't think we've done that in the past?

ggainey:
> Actually, we have a real problem here. The change changes the pulp_file api schema in a way to make the pulp_file_client package incompatible.

Yes, that is an impact of changing the API. See prev comment RE "we're still in tech-preview and this is why".

mdellweg:
> > Actually, we have a real problem here. The change changes the pulp_file api schema in a way to make the pulp_file_client package incompatible.
> 
> Yes, that is an impact of changing the API. See prev comment RE "we're still in tech-preview and this is why".

Let me rephrase: This is changing the pulp_file api without changing pulp_file's version, but just changing pulpcore's version. This means we cannot serve a pulp_file_client package that will work for the lifetime of the corresponding pulp_file version.
All in all, this is a very subtle breakage of the deprecation policy.
What we could do, was
1. Limit the change to the pulp_exporter, which is inconsistent.
2. Do the change for the filesystem_exporter one release ahead (inconsistent for a limited amount of time) and then let pulpcore follow a release later.
3. Do nothing

bmbouter:
+1 to option (2) because our feature set should be consistent.

mdellweg:
> +1 to option (2) because our feature set should be consistent.

https://github.com/pulp/pulp_file/pull/483

daviddavis (approved):
I might add a CHANGES/7438.removal but other than that, looks good to me.

CHANGES/7438.bugfix
goosemania: I'd agree that just a change of API because we decided so, would be a feature.
I can convince myself that it's a bugfix since the ticket describes the problem with the calls being synchronous.

goosemania (approved):

PR #1201, opened 2021-03-22
Add warning if there are remote artifacts with no allowed checksums

daviddavis:
fixes #7985

### DIFF
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -13,6 +13,7 @@ import sys
 from contextlib import suppress
 from gettext import gettext as _
 from importlib import import_module
+from logging import getLogger
 from pathlib import Path
 from pkg_resources import iter_entry_points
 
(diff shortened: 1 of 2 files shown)


pulpcore/app/settings.py
```diff
@@ -341,6 +344,24 @@
                         ).format(checksum)
                     )
 
+            # warn if there are remote artifacts with checksums but no allowed checksums
+            cond = " AND ".join([f"{c} IS NULL" for c in constants.ALL_KNOWN_CONTENT_CHECKSUMS])
+            no_checksum_query = f"SELECT pulp_id FROM core_remoteartifact WHERE {cond}"
+            cond = " AND ".join([f"{c} IS NULL" for c in ALLOWED_CONTENT_CHECKSUMS])
```
ipanova, line 350: does it make sense to make this query smaller and query only those remote artifacts which have no checksums from `ALLOWED_CONTENT_CHECKSUMS` list or am i missing some corner case?

daviddavis: This is what the main part of the query does. However, we have to also exclude remote artifacts at all which is what the subquery is doing.

ipanova: `RemoteArtifacts without any checksums which is perfectly valid and not prohibited.` yeah this is the part  i forgot about, thanks for reminding me

ipanova (approved):

bmbouter (approved):
This looks good.

PR #1204, opened 2021-03-22
Limit how many repo versions Pulp retains

daviddavis:
fixes #8368


CHANGES/8368.feature
```diff
@@ -0,0 +1,2 @@
+Added ability for users to limit how many repo versions Pulp retains by setting
+``version_retention`` on repository.
```
mdellweg, line 2: This field is supposed to hold the number of retained versions, right?
To me "retained_versions" would be more descriptive.

pulpcore/app/models/repository.py
```diff
@@ -179,6 +181,18 @@ def artifacts_for_version(version):
...
+        """Cleanup old repository versions based on version_retention."""
+        if self.version_retention:
+            for version in self.versions.order_by("-number")[self.version_retention :]:
+                _logger.info(
+                    _("Deleting repository version {} due to version retention limit.").format(
+                        version
+                    )
+                )
+                version.delete()
+
```
mdellweg, line 195: I imagined to see this code in the `__exit__` of the version creation context manager.
Does the version creation trigger a save on the repository?

daviddavis: I originally had it there but yes, you're correct: `__exit__` triggers [a save on the repository](https://github.com/pulp/pulpcore/blob/a95cabeef8945bd683e1352ad5d75d57a653105d/pulpcore/app/models/repository.py#L912).

I just saw @bmbouter's comment on https://pulp.plan.io/issues/8368 where he suggests we only trigger version cleanup on `__exit__` and not when updating `retained_versions`. I can make that change if no one objects.

ipanova: we talked some of this during pulpcore meeting and people ended up having different opinions

mdellweg: My take, was that setting the limit will also clean up. And even wondered if we needed to trigger the version deletion in a second place, but then i saw your SAVE_HOOK and thought "oh yeah". I just wondered if it also served the new version creation case, which it does, because we increase the next_version counter there.

CHANGES/8368.feature
dralley: I agree, here's an example of what the RPM plugin does for package versions https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/models/repository.py#L150

pulpcore/app/models/repository.py
dralley: My personal preference is `__exit__`, but I have no strong opinions.

pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
```diff
@@ -1059,3 +1059,60 @@ def test_add_content_with_base_version(self):
         repo = self.client.get(repo["pulp_href"])
 
         self.assertEqual(get_content(repo)[FILE_CONTENT_NAME][0], content2)
+
+
+class RepoVersionRetentionTestCase(unittest.TestCase):
```
dralley, line 1064: For new tests, it would be better to use the bindings, if possible... however, this is already written, so no objection

bmbouter: I agree, we want all new tests to be bindings based.

pulpcore/app/models/repository.py
bmbouter: I have a use case and a "best practices" argument in favor of keeping is so that setting the limit also performs cleanup.

### Use case

Users may not be able to trigger a sync or modify operation again. Either because they don't want to, because they don't have connectivity, the remote host is offline temporarily or permanently, etc. If we don't delete the repository versions at value-set-time those users could never practically use the feature.

### Best Practices Argument

We should keep features separate whenever we can. Software is complicated, and when features start to subtly depend on each other, things get more complicated for everyone. In this case we'd be making the version retention feature depend on the sync/modify feature set, which seems unnecessary and concerning.

pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
daviddavis: Yea, I tried using bindings initially but this file nor the code I reused had bindings. But generally, I agree as well.

pulpcore/app/models/repository.py
ipanova: My worry with this approach is that it leaves no space for the users for a mistake. 
If the user won't have any option to subsequently trigger sync/modify of the repo he could just during repo creation set this retention policy.
We already have a similar feature in rpm plugin related to the content retention policy. Whatever we decide i would prefer both features follow same approach. Meaning that if content retention policy is being set on the repo it should then be triggered right away. I understand that this is purely rpm plugin feature and plugin writers have the freedom to define whatever behavior they want, however it would be great it we stick, where it's possible and makes sense to similar behavior in order to decrease  discrepancies.

ipanova:
@daviddavis since repo version 0 is special, we should always keep it and maybe exclude from the version count to keep, meaning if the user has set retain_versions=1 we will keep latest and 0. thoughts?

pulpcore/app/models/repository.py
```diff
@@ -179,6 +181,18 @@ def artifacts_for_version(version):
         """
         return Artifact.objects.filter(content__pk__in=version.content)
 
+    @hook(AFTER_SAVE)
+    def cleanup_old_versions(self):
+        """Cleanup old repository versions based on version_retention."""
+        if self.version_retention:
+            for version in self.versions.order_by("-number")[self.version_retention :]:
```
ipanova, line 188: self.versions.order_by("-number")[1:self.versions.count()-1]

ipanova: negative indexing seems like is not supported

daviddavis:
@ipanova currently I am waiting on the outcome of https://pulp.plan.io/issues/8454.

ipanova:
> @ipanova currently I am waiting on the outcome of https://pulp.plan.io/issues/8454.

oh ok, i missed this one

mdellweg:
Code looks good.
What i am pondering here is how does it handle versions that are published / attached to a distribution?
Is it playing nice with auto_publish when set to one?

daviddavis:
I'll add a test to cover the case of auto-publish/auto-distribute with `retained_versions`.

mdellweg (approved):

pulpcore/app/serializers/repository.py
```diff
@@ -31,6 +31,15 @@ class RepositorySerializer(ModelSerializer):
     description = serializers.CharField(
         help_text=_("An optional description."), required=False, allow_null=True
     )
+    retained_versions = serializers.IntegerField(
+        help_text=_(
+            "Retain X versions of the repository. Default is null which retains all versions."
+            " This is provided as a tech preview in Pulp 3 and may change in the future."
```
ipanova, line 37: can you add this feature in the docs that it is in 'tech preview'?
Not sure which place would be the best https://docs.pulpproject.org/pulpcore/concepts.html or https://docs.pulpproject.org/pulpcore/plugins/plugin-writer/concepts/tasks/add-remove.html?highlight=repository%20version#repository-versions?

ipanova: another idea would be a create a new workflow for repo versions where we could how to trigger it and when it gets triggered, specifically to clarify that versions are getting cleaned up even on the repo update call.

pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
```diff
@@ -1071,3 +1083,86 @@ def test_add_content_with_base_version(self):
...
+
+        # update retained_versions to 2
+        result = self.repo_api.partial_update(self.repo.pulp_href, {"retained_versions": 2})
+        monitor_task(result.task)
+
+        versions = self.version_api.list(file_file_repository_href=self.repo.pulp_href).results
+        self.assertEqual(len(versions), 2)
+        self.assertEqual(self.repo.latest_version_href.split("/")[-2], "3")
+
+    def test_autodistribute(self):
```
ipanova, line 1156: :+1:

daviddavis:
I decided to go with some workflow docs, which means I need to update the CLI. Setting this back to draft while I do that.

ipanova (approved):

PR #1205, opened 2021-03-23
Allow plugins to unset attribute in NamedModelViewSet

awcrosby:
Allow plugins to set `queryset_filtering_required_permission` attribute to
`None` and have it behave the same way as if the attribute were not present.

fixes #8438
https://pulp.plan.io/issues/8438

### DIFF
--- a/CHANGES/plugin_api/8438.bugfix
+++ b/CHANGES/plugin_api/8438.bugfix
@@ -1,0 +1,1 @@
+Allow plugins to unset the ``queryset_filtering_required_permission`` attribute in ``NamedModelViewSet``.
--- a/pulpcore/app/viewsets/base.py
+++ b/pulpcore/app/viewsets/base.py
@@ -326,12 +326,10 @@ class NamedModelViewSet(viewsets.GenericViewSet):
                 filters[lookup] = self.kwargs[key]
             qs = qs.filter(**filters)
 
-        try:
-            permission_name = self.queryset_filtering_required_permission
-        except AttributeError:
-            pass
-        else:
+        permission_name = getattr(self, "queryset_filtering_required_permission", None)
+        if permission_name:
             qs = get_objects_for_user(self.request.user, permission_name, klass=qs)
+
         return qs
 
     @classmethod


pulpcore/app/viewsets/base.py
```diff
@@ -331,7 +331,8 @@ def get_queryset(self):
         except AttributeError:
             pass
         else:
-            qs = get_objects_for_user(self.request.user, permission_name, klass=qs)
+            if permission_name:
+                qs = get_objects_for_user(self.request.user, permission_name, klass=qs)
```
mdellweg, line 335: Can you replace the try - except - else structure here (L330) with
`getattr(self, "queryset_filtering_required_permission", None)`
That way, we can absorb the `AttributeError` in the `if permission_name` check.

awcrosby: Thanks - updated.

ipanova (approved):

mdellweg (approved):

CHANGES/8438.bugfix
```diff
@@ -0,0 +1 @@
+Allow plugins to unset attribute in NamedModelViewSet
```
mdellweg, line 1: Can this be moved to `CHANGES/plugin` ?

mdellweg (changes requested):

CHANGES/8438.bugfix
mdellweg: Sorry, didn't see this earlier.

awcrosby: No problem. I assume you meant ` CHANGES/plugin_api` as I didn't see ` CHANGES/plugin` listed in the docs, but if not just let me know.

CHANGES/plugin_api/8438.bugfix
```diff
@@ -0,0 +1 @@
+Allow plugins to unset attribute in NamedModelViewSet
```
mdellweg, line 1:
```suggestion
Allow plugins to unset the ``queryset_filtering_required_permission`` attribute in ``NamedModelViewSet``.
```

mdellweg (approved):
Just one last suggestion to the changelog before merging.

PR #1206, opened 2021-03-23
Fixed import deadlock by force-locking Tables Of Interest.

ggainey:
fixes #7904
[nocoverage]

### DIFF
--- a/CHANGES/7904.bugfix
+++ b/CHANGES/7904.bugfix
@@ -1,0 +1,1 @@
+Taught ``PulpImport`` to avoid deadlock by correctly locking tables being imported.
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -10,6 +10,10 @@ from logging import getLogger
 
 from django.conf import settings
 from django.core.files.storage import default_storage
+from django.db import (
+    connection,
+    transaction,
+)
 from django.db.models import F
 
 from pkg_resources import DistributionNotFound, get_distribution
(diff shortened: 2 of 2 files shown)


bmbouter:
Which tables does this lock? Is it the Content table?

ggainey:
> Which tables does this lock? Is it the Content table?

The (newly-pushed) version locks each table it's about to import data into. The previous version suffered from a fatal cut-and-paste error...

While the current state of the PR fixes the deadlock, I'm not happy with it. Let's mark this as "draft", maybe with thought and discussion we can come up with a more elegant/performant solution.

ipanova:
@ggainey  this is marked for 3.12 will this still be the case if you move it to draft state?

ggainey:
> @ggainey this is marked for 3.12 will this still be the case if you move it to draft state?

I don't like this approach, but it may be our best bet. I will have time to do some different experiments in the next couple days. How about I say, "no later than Weds 31-MAR we'll call it for 3.12, or Not"?

bmbouter:
Where in the code are we creating transactions that these inserts compete with?

ggainey:
> Where in the code are we creating transactions that these inserts compete with?

This code competes with itself. Repositories are imported in parallel, import_repository_version() gets enqueued for each repo in the import-file. See https://github.com/pulp/pulpcore/blob/02e157f2e3edf21118be861aafa95530c1f004ac/pulpcore/app/tasks/importer.py#L419

daviddavis (approved):
I can't think of a better solution so LGTM.

ggainey:
Closing - this is not the right way to approach this problem.

PR #1207, opened 2021-03-24
Remove checksum filtering in Remote.get_downloader and add checksum t…

ipanova:
…ype enforcement to downloaders itself

closes #8435

### DIFF
--- a/CHANGES/plugin_api/8435.feature
+++ b/CHANGES/plugin_api/8435.feature
@@ -1,0 +1,1 @@
+Added checksum type enforcement to ``pulpcore.plugin.download.BaseDownloader``.
--- a/CHANGES/plugin_api/8435.removal
+++ b/CHANGES/plugin_api/8435.removal
@@ -1,0 +1,1 @@
+Removed checksum type filtering from ``pulpcore.plugin.models.Remote.get_downloader`` and ``pulpcore.plugin.stages.DeclarativeArtifact.download``.
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -13,6 +13,7 @@ from django.db import models, transaction
 from django.urls import reverse
 
 from pulpcore.app.util import batch_qs, get_view_name_for_model
+from pulpcore.constants import ALL_KNOWN_CONTENT_CHECKSUMS
 from pulpcore.download.factory import DownloaderFactory
 from pulpcore.exceptions import ResourceImmutableError
 
@@ -349,7 +350,7 @@ class Remote(MasterModel):
         if remote_artifact:
             url = remote_artifact.url
             expected_digests = {}
-            for digest_name in Artifact.DIGEST_FIELDS:
+            for digest_name in ALL_KNOWN_CONTENT_CHECKSUMS:
                 digest_value = getattr(remote_artifact, digest_name)
                 if digest_value:
                     expected_digests[digest_name] = digest_value
--- a/pulpcore/download/base.py
+++ b/pulpcore/download/base.py
@@ -1,3 +1,5 @@
+from gettext import gettext as _
+
 import asyncio
 from collections import namedtuple
 import logging
@@ -6,7 +8,11 @@ import tempfile
 
 from pulpcore.app import pulp_hashlib
 from pulpcore.app.models import Artifact
-from pulpcore.exceptions import DigestValidationError, SizeValidationError
+from pulpcore.exceptions import (
+    DigestValidationError,
+    SizeValidationError,
+    UnsupportedDigestValidationError,
+)
 
 
 log = logging.getLogger(__name__)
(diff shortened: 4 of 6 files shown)


CHANGES/8435.misc
```diff
@@ -0,0 +1 @@
+Remove checksum filtering in Remote.get_downloader and add checksum type enforcement to downloaders itself
```
bmbouter, line 1: Can the downloader enforcement become its own CHANGES/plugin_api/8435.feature entry? I think plugin writers would see that behavioral change as a feature they need the FYI on.

pulpcore/app/models/repository.py
```diff
@@ -349,7 +350,7 @@ def get_downloader(self, remote_artifact=None, url=None, **kwargs):
         if remote_artifact:
             url = remote_artifact.url
             expected_digests = {}
-            for digest_name in Artifact.DIGEST_FIELDS:
+            for digest_name in ALL_KNOWN_CONTENT_CHECKSUMS:
```
bmbouter, line 353: :+1:

pulpcore/download/base.py
```diff
@@ -96,6 +102,14 @@ def __init__(
             self.semaphore = asyncio.Semaphore()  # This will always be acquired
         self._digests = {n: pulp_hashlib.new(n) for n in Artifact.DIGEST_FIELDS}
         self._size = 0
+        if self.expected_digests:
```
ipanova, line 105: I see a potential problem with this approach. get_downloader is called in the command `handle-artifact-checksums`
This command is getting called in a the migration 0061. How much should we worry about getting user stuck in the middle of the migration run?:)

should we maybe add a flag to the command which will enable on demand the download of bits? otherwise we would nit try to fix missing files by default when running the command
https://github.com/pulp/pulpcore/blob/master/pulpcore/app/management/commands/handle-artifact-checksums.py#L130

bmbouter: Can you describe the problematic series of events a bit more please?

daviddavis: Trying to think through how this problem would be encountered. It would mean a user is upgrading/running the migration, and somehow has an artifact missing a checksum (meaning that they've added a checksum to their ALLOWED_CONTENT_CHECKSUMS before running the migration)?

ipanova: yes, he would need to have a checksum missing from the list in the ALLOWED_CONTENT_CHECKSUMS and in addition he would need to have missing the file.
I agree that that this situation is unlikely, thought it would still be good to put it on the table for discussion

pulpcore/download/base.py
```diff
@@ -96,6 +102,14 @@ def __init__(
...
+            if not set(self.expected_digests).intersection(set(Artifact.DIGEST_FIELDS)):
+                raise UnsupportedDigestValidationError(
+                    _(
+                        "Content does not contain at least one trusted hasher which is specified in"
+                        "'ALLOWED_CONTENT_CHECKSUMS' setting."
```
daviddavis, line 110: I might add url to this error message. It could help the user to debug should they ever hit this message.

pulpcore/tests/unit/download/test_download_base.py
```diff
@@ -0,0 +1,60 @@
+from unittest import mock
+
+from django.test import TestCase
+
+from pulpcore.download import BaseDownloader
+from pulpcore.plugin.exceptions import UnsupportedDigestValidationError
+
+
+class BaseDownloaderInitTestCase(TestCase):
```
daviddavis, line 9: :+1:

daviddavis (approved):

CHANGES/plugin_api/8435.feature
```diff
@@ -0,0 +1 @@
+Removed checksum filtering in Remote.get_downloader. Added checksum type enforcement to the downloader itself.
```
bmbouter, line 1: apologies @ipanova for the re-request on it. Can this be:

#  CHANGES/plugin_api/8435.feature
Added checksum type enforcement to the downloader itself.

#  CHANGES/plugin_api/8435.removal
Removed checksum type filtering from ``pulpcore.plugin.models.Remote.get_downloader`` and ``pulpcore.plugin.models.Artifact.download``.

CHANGES/plugin_api/8435.removal
```diff
@@ -0,0 +1 @@
+Removed checksum type filtering from `pulpcore.plugin.models.Remote.get_downloader` and `pulpcore.plugin.stages.DeclarativeArtifact.download`.
```
bmbouter, line 1:
```suggestion
Removed checksum type filtering from ``pulpcore.plugin.models.Remote.get_downloader`` and ``pulpcore.plugin.stages.DeclarativeArtifact.download``.
```

bmbouter: The rst syntax needs double backticks for sphinx quoting. It's such a small change though I'm lgtm-ing now.

CHANGES/plugin_api/8435.feature
```diff
@@ -0,0 +1 @@
+Added checksum type enforcement to the downloader itself.
```
bmbouter, line 1:
```suggestion
Added checksum type enforcement to ``pulpcore.plugin.download.BaseDownloader``.
```

bmbouter (approved):
Thanks @ipanova !

bmbouter:
Thanks @ipanova !

PR #1209, opened 2021-03-24
RemoteArtifacts checksum type check

pavelpicka:
RemoteArtifacts are checked for forbidden checksums.

closes: #8423
https://pulp.plan.io/issues/8423

### DIFF
--- a/CHANGES/8423.feature
+++ b/CHANGES/8423.feature
@@ -1,0 +1,1 @@
+Added sync check that raises error when only forbidden checksums are found for on-demand content.
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -558,6 +558,31 @@ class RemoteArtifact(BaseModel, QueryMixin):
 
     objects = BulkCreateManager()
 
+    def validate_checksums(self):
+        """Validate if RemoteArtifact has allowed checksum or no checksum at all."""
+        if not any(
+            [
+                checksum_type
+                for checksum_type in ALL_KNOWN_CONTENT_CHECKSUMS
+                if getattr(self, checksum_type, False)
+            ]
+        ):
+            return
+        if not any(
+            [
+                checksum_type
+                for checksum_type in Artifact.DIGEST_FIELDS
+                if getattr(self, checksum_type, False)
+            ]
+        ):
+            raise UnsupportedDigestValidationError(
+                _(
+                    "On-demand content located at the url {} contains forbidden checksum type,"
+                    "thus cannot be synced."
+                    "You can allow checksum type with 'ALLOWED_CONTENT_CHECKSUMS' setting."
+                ).format(self.url)
+            )
+
     class Meta:
         unique_together = ("content_artifact", "remote")
 
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -323,7 +323,7 @@ class RemoteArtifactSaver(Stage):
 
     @staticmethod
     def _create_remote_artifact(d_artifact, content_artifact):
-        return RemoteArtifact(
+        ra = RemoteArtifact(
             url=d_artifact.url,
             size=d_artifact.artifact.size,
             md5=d_artifact.artifact.md5,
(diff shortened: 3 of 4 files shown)


daviddavis:
The logic here is not quite correct. We want to allow a RemoteArtifact if it (1) has no checksums or (2) has an allowed checksum. So it can have forbidden checksum(s) if also has an allowed checksum. Something like this:

```python
def validate_checksums(self):
    """Check that a RemoteArtifact either doesn't have checksums or has no allowed checksums.""""
    if len([c for c in ALL_KNOWN_CHECKSUMS if get_attr(self, c, False)]) == 0:
        # doesn't have any checksums
        return
    if len([c for c in Artifact.DIGEST_FIELDS if get_attr(self, c, False)]) == 0:
        # doesn't have any allowed checksums
        raise UnsupportedDigestValidationError("...")
```

CHANGES/plugin_api/8423.feature
```diff
@@ -0,0 +1 @@
+RemoteArtifacts are checked for not allowed checksum types and Pulp raises an exception if any found.
```
daviddavis, line 1: I think this is more of a user-facing change than a change that impacts plugin writers.

daviddavis:
Tests look good. Can you add a test that checks a RemoteArtifact with an allowed AND a forbidden checksum? It should not raise an exception.

pulpcore/app/models/content.py
```diff
@@ -558,6 +558,28 @@ class RemoteArtifact(BaseModel, QueryMixin):
...
+            return
+        if not any(
+            [
+                checksum_type
+                for checksum_type in Artifact.DIGEST_FIELDS
+                if getattr(self, checksum_type, False)
+            ]
+        ):
+            raise UnsupportedDigestValidationError(
+                "RemoteArtifact contains forbidden checksum type, thus cannot be synced."
```
daviddavis, line 579: I don't know if users know what RemoteArtifacts are. I would maybe say "On-demand content".

daviddavis: I would also include `self.url` in the message to help users track down the problem.

pavelpicka: Also changed in CHANGES message

daviddavis (approved):

pulpcore/app/models/content.py
```diff
@@ -558,6 +558,30 @@ class RemoteArtifact(BaseModel, QueryMixin):
...
+        if not any(
+            [
+                checksum_type
+                for checksum_type in Artifact.DIGEST_FIELDS
+                if getattr(self, checksum_type, False)
+            ]
+        ):
+            raise UnsupportedDigestValidationError(
+                _(
+                    "On-demand content {} contains forbidden checksum type, thus cannot be synced."
```
ipanova, line 580: if i may suggest : On-demand content located at the url {}...

pulpcore/app/models/content.py
```diff
@@ -558,6 +558,30 @@ class RemoteArtifact(BaseModel, QueryMixin):
 
     objects = BulkCreateManager()
 
+    def validate_checksums(self):
+        """Validate if RemoteArtifact has allowed checksum or no checksum at all."""
+        if not any(
+            [
+                checksum_type
+                for checksum_type in ALL_KNOWN_CONTENT_CHECKSUMS
+                if getattr(self, checksum_type, False)
```
ipanova, line 567: in theory the AttributeError should never be raised

CHANGES/8423.feature
```diff
@@ -0,0 +1 @@
+On-demand content are checked for not allowed checksum types and Pulp raises an exception if any found.
```
ipanova, line 1: s/are checked/is checked/

ipanova: On-demand content is checked for forbidden checksum types and Pulp raises an exception when those are exclusively found.

daviddavis: Both work. I might even use something simpler: "Added sync check that raises error when only forbidden checksums are found for on-demand content".

pulpcore/app/models/content.py
pavelpicka: in theory yes, but I'd rather stay explicit

ipanova (approved):

PR #1210, opened 2021-03-25
Use pulp-smash from source

daviddavis:
fixes #8450

### DIFF
--- a/functest_requirements.txt
+++ b/functest_requirements.txt
@@ -1,6 +1,6 @@
 django
 dynaconf
-pulp-smash  # >=1!0.12.0 - broken: https://github.com/dependabot/dependabot-core/issues/3165
+pulp-smash @ git+https://github.com/pulp/pulp-smash.git
 pulpcore-client
 pulp-file-client
 pytest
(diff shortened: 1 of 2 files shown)


functest_requirements.txt
```diff
@@ -1,6 +1,7 @@
 django
 dynaconf
 pulp-smash  # >=1!0.12.0 - broken: https://github.com/dependabot/dependabot-core/issues/3165
+git+git://github.com/pulp/pulp-smash@master#egg=pulp-smash
```
mdellweg, line 4: I think, it should replace that line.

mdellweg (approved):

bmbouter:
With this do we want to make a CI adjustment which I believe carries code to do the same thing? I'm not sure where that code lives tho.

bmbouter:
Also what about other plugins? The dev environment on the installer installing other plugins may bring in the pypi release of pulp_smash, and that would cause pip to believe this dep is satisfied causing the pulpcore tests to not be runnable?

bmbouter:
Random idea: what if pulp_smash was no longer a released software, and everyone used git checkouts...  That way it would be the same everywhere.

daviddavis:
I emailed pulp-dev asking for objections/concerns about removing pulp-smash from PyPI.

dralley (approved):

PR #1212, opened 2021-03-26
Don't check RemoteArtifact by artifact check

pavelpicka:
Don't check RemoteArtifact by artifact check in stages.

closes: #8445
https://pulp.plan.io/issues/8445

### DIFF
--- a/CHANGES/8445.bugfix
+++ b/CHANGES/8445.bugfix
@@ -1,0 +1,1 @@
+Fix artifact checksum check to not check on-demand content.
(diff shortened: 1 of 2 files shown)


pavelpicka:
I'd like to see @bmbouter review on this one.

bmbouter (approved):
I hand tested this against the reproducer without this patch and it reproduced the problem. I tested it with this patch and it resolved the problem.

Additionally, in reading the code I believe it's what should be done.

Thanks @pavelpicka ! :+1:

daviddavis (approved):

PR #1219, opened 2021-03-30
Fixed artifact_stages edge case for multi-artifact/multi-remote batches.

ggainey:
Only encountered while 2to3-migrating specific content in specific orders.

fixes #8377
[nocoverage]

### DIFF
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -303,8 +303,11 @@ class RemoteArtifactSaver(Stage):
         )
         needed_ras = []
         for d_content in batch:
-            for content_artifact in d_content.content._remote_artifact_saver_cas:
-                for d_artifact in d_content.d_artifacts:
+            for d_artifact in d_content.d_artifacts:
+                if not d_artifact.remote:
+                    continue
+
+                for content_artifact in d_content.content._remote_artifact_saver_cas:
                     if d_artifact.relative_path == content_artifact.relative_path:
                         break
                 else:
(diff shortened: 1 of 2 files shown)


ggainey:
@dralley @goosemania this fix deserves an explanation :) 

The prefetch-work is a great optimization, but it results in `_remote_artifact_saver_cas` containing any remote needed for anything in the batch, for all `d_content.content.artifacts` in the batch.

The existing code essentially said, "for each entity in the batch, for every entry in `_remote_artifact_saver_cas`, find the matching `d_content.d_artifacts` and complain if you can't find one".

The new code switches the nesting to "for each entity in the batch, for each entry in `d_content.d_artifacts`, **IF** the d_artifact has a remote, find a matching `_remote_artifact_saver_cas` . Complain if there isn't one"

This definitely fixes the reported problem. I *believe* it does so sanely. But this code made my brain melt, so I def need more eyes please.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -303,22 +303,25 @@ def _needed_remote_artifacts(self, batch):
         )
         needed_ras = []
         for d_content in batch:
-            for content_artifact in d_content.content._remote_artifact_saver_cas:
-                for d_artifact in d_content.d_artifacts:
+            for d_artifact in d_content.d_artifacts:
+                if not d_artifact.remote:
+                    continue
```
dralley, line 308: This seems reasonable to me.  It's even a bit of an optimization :)

ggainey: I hadn't even thought of that, but you have a point!

dralley (approved):

PR #1222, opened 2021-03-31
Adds basic PoC of new worker type

bmbouter:
This PR is very unpolished, but here is the state of things:

Done to a working point:
* Fallback support allowing this code to work with RQ style workers
* Introduction of the ` USE_NEW_WORKER_TYPE` setting
* The `puplcore-worker` entrypoint
* The heartbeating of workers into the status API the same as before
* Finding the next safe task
* Correctly using advisory locks to synchronize exclusive access to
  resources, and tested locks are released upon worker exit.
* transitioning of task states from running, completed, and failed
* calling into the task code itself

Not done:
* The distributed worker cleanup. Right now the workers do not watch
  each other.
* worker forking
* general docs
* changelogs
* setting the Task.parent_task when a task is dispatched from within a
  task.

The "recording of task reserved resources" to be shown to users in
TaskSerializer is a huge mess. This PR basically avoids that entirely by
introducing `Task.new_reserved_resources_record`. The hope is that once
this worker style is the only type Pulp uses we can delete the 4-ish
models providing the "ReservedResource..." models.

However! The TaskSerializer currently is *not* integrated with
`Task.new_reserved_resources_record`.

Also I used the not-recommended `hashtext` postgresql function to
transform the string to 64-bit data for the advisory locks. It works
well, but it's not safe in the long-term because it's an internal to
postgresql. It was fast and easy for this PoC though. The blake2
implementation I have on the ticket wasn't working so I moved on for
now.


pulpcore/tasking/tasks.py
```diff
@@ -227,3 +267,49 @@ def as_url(r):
...
+        resources = list(_validate_and_get_resources_as_set(resources))
+        with transaction.atomic():
+            task = Task.objects.create(
+                state=TASK_STATES.WAITING,
+                logging_cid=(GuidMiddleware.get_guid() or ""),
+                task_group=task_group,
+                name=f"{func.__module__}.{func.__name__}",
+                args=args_as_json,
+                kwargs=kwargs_as_json,
+                parent_task=None,  # TODO implement me!
```
bmbouter, line 307: This needs resolution.

mdellweg:
This is superseeded by #1261

PR #1223, opened 2021-04-01
Allow plugin writers to hide repositories

daviddavis:
fixes #8487


bmbouter (approved):
Niiiice!

CHANGES/plugin_api/8487.feature
```diff
@@ -0,0 +1 @@
+Added ``user_hidden`` field to ``Repository`` to hide repositories from users.
```
ipanova, line 1: Looking at our code history we removed `plugin_managed` repos because with typed repos we decided to let it on the plugins. I understand that we are bringing this back because of the generic repo list endpoint but it still seems to me unclear - the field is available in core yet all the business is still done in plugins , thereby I don't see this PR as enabling plugin writers to hide repos but more like enable generic repo list endpoint to work properly.
I also would suggest to add this check during version creation https://github.com/pulp/pulpcore/commit/2161bf71ddce76394933fe0d52b09194a2153501#diff-29bc9ed3cae158e3a508ec8cfa3c61538c4ddf9c88d77f033b61005d0af9019cL86 so we at least get rid of this duplicated code in rpm plugin https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/models/repository.py#L251

daviddavis:
Need to update this to handle filtering repos for the new global repo versions endpoint.

CHANGES/plugin_api/8487.feature
daviddavis: > Looking at our code history we removed plugin_managed repos because with typed repos we decided to let it on the plugins. I understand that we are bringing this back because of the generic repo list endpoint but it still seems to me unclear - the field is available in core yet all the business is still done in plugins , thereby I don't see this PR as enabling plugin writers to hide repos but more like enable generic repo list endpoint to work properly.

That's correct. It provides a standard field plugin writers can use and a way to hide repos from the generic repo endpoint.

> I also would suggest to add this check during version creation 2161bf7#diff-29bc9ed3cae158e3a508ec8cfa3c61538c4ddf9c88d77f033b61005d0af9019cL86 so we at least get rid of this duplicated code in rpm plugin https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/models/repository.py#L251

:+1:

pulpcore/app/viewsets/repository.py
```diff
@@ -242,5 +242,5 @@ class ListRepositoryVersionViewSet(
 ):
     endpoint_name = "repository_versions"
     serializer_class = RepositoryVersionSerializer
-    queryset = RepositoryVersion.objects.all()
+    queryset = RepositoryVersion.objects.exclude(repository__user_hidden=True)
```
ipanova, line 245: :+1:

pulpcore/app/models/repository.py
```diff
@@ -58,6 +58,7 @@ class Repository(MasterModel):
         "Content", through="RepositoryContent", related_name="repositories"
     )
     remote = models.ForeignKey("Remote", null=True, on_delete=models.SET_NULL)
+    user_hidden = models.BooleanField(default=False)
```
ipanova, line 61: can you add the field description, also for `retained_versions` please

ipanova (approved):

PR #1225, opened 2021-04-01
Fixed bug where deprecation warnings weren't being logged

daviddavis:
fixes #8499

### DIFF
--- a/docs/plugins/plugin-writer/concepts/index.rst
+++ b/docs/plugins/plugin-writer/concepts/index.rst
@@ -400,7 +400,8 @@ In 3.8 the following changes happen:
 2. The existing method signature ``def foo(a, b)`` is left in-tact.
 3. The ``foo`` method would have the a Python ``DeprecationWarning`` added to it such as::
 
-    warnings.warn("foo() is deprecated and will be removed in pulpcore==3.9; use the_new_foo().", DeprecationWarning)
+    from pulpcore.app.logging import deprecation_logger
+    deprecation_logger.warn("foo() is deprecated and will be removed in pulpcore==3.9; use the_new_foo().")
 
 4. A ``CHANGES/plugin_api/XXXX.deprecation`` changelog entry is created explaining how to port
    plugin code onto the new call interface.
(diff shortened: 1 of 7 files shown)


docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -421,6 +422,17 @@ This also applies to models importable from ``pulpcore.plugin.models``. For exam
 that is being renamed or removed would follow a similar deprecation process described above to allow
 plugin code one release cycle to update their code compatibility.
 
+Logging of deprecation warnings can be disabled by raising the log level for the
```
bmbouter, line 425: I think the first sentence in L#417 should be removed since we don't use those anymore. Maybe replaced with some text indicating that showing warnings to users and developers by default is be design.

pulpcore/deprecation.py
```diff
@@ -0,0 +1,7 @@
+import logging
```
bmbouter, line 1: I like that this keeps the logging adjustment simple as `pulpore.deprecation`, but I wonder if maybe this should be `pulpcore.logging.deprecation`. I actually imagine we'll ship a tasking logger in the future that logs to a task. We could move it later, wdyt?

daviddavis: Are you referring to the module or the logger name? Having a pulpcore.logging module makes sense. Having logging though in the logger name seems redundant. I am pushing another version of this code to address the former. Let me know what you think.

docs/plugins/plugin-writer/concepts/index.rst
daviddavis: Done.

pulpcore/app/logging.py
```diff
@@ -0,0 +1,6 @@
+import logging
+
+
+def deprecation_warning(msg):
+    logger = logging.getLogger("pulpcore.deprecation")
```
bmbouter, line 5: niiiiiice

bmbouter (approved):
Thank you!

daviddavis:
@bmbouter made a slight tweak to have pulpcore.app.logging define a `deprecation_logger` instead of having a function. I think this is more flexible/powerful. Let me know what you think.

pulpcore/tasking/storage.py
```diff
@@ -187,12 +188,11 @@ def __init__(self):
         Raises:
             RuntimeError: When used outside of an RQ task.
         """
-        warnings.warn(
+        deprecation_logger.warn(
             _(
                 "WorkingDirectory is deprecated and will be removed in pulpcore==3.12; "
```
dralley, line 193: Should we fix this to say 3.13?

dralley (approved):

pulpcore/tasking/storage.py
bmbouter: I can look at fixing this in my PR.

PR #1226, opened 2021-04-01
Adds new `dispatch` call and deprecate the old one

bmbouter:
This adds a new `pulpcore.plugin.tasking.dispatch` interface which will
replace the `pulpcore.plugin.tasking.enqueue_with_reservation`
interface. This also deprecates the
`pulpcore.plugin.tasking.enqueue_with_reservation` and causes it to
emit warnings if used.

Additionally the `pulpcore.plugin.viewsets.OperationPostponedResponse`
has been ported to support both the `dispatch` and
`enqueue_with_reservation` interfaces.

closes #8496

### DIFF
--- a/CHANGES/plugin_api/8496.feature
+++ b/CHANGES/plugin_api/8496.feature
@@ -1,0 +1,7 @@
+Adds the ``pulpcore.plugin.tasking.dispatch`` interface which replcaes the
+``pulpcore.plugin.tasking.enqueue_with_reservation`` interface. It is the same except:
+* It returns a ``pulpcore.plugin.models.Task`` instead of an RQ object
+* It does not support the ``options`` keyword argument
+
+Additionally the ``pulpcore.plugin.viewsets.OperationPostponedResponse`` was updated to support both
+the ``dispatch`` and ``enqueue_with_reservation`` interfaces.
--- a/CHANGES/plugin_api/8496.feature
+++ b/CHANGES/plugin_api/8496.feature
@@ -1,4 +1,4 @@
-Adds the ``pulpcore.plugin.tasking.dispatch`` interface which replcaes the
+Adds the ``pulpcore.plugin.tasking.dispatch`` interface which replaces the
 ``pulpcore.plugin.tasking.enqueue_with_reservation`` interface. It is the same except:
 * It returns a ``pulpcore.plugin.models.Task`` instead of an RQ object
 * It does not support the ``options`` keyword argument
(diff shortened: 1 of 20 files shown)


pulpcore/tasking/tasks.py
```diff
@@ -176,6 +177,21 @@ def enqueue_with_reservation(
     Raises:
         ValueError: When `resources` is an unsupported type.
     """
+    logging.warning(
```
bmbouter, line 180: This needs to be refactored to use the facilities here: https://github.com/pulp/pulpcore/pull/1225

dralley (approved):

pulpcore/app/response.py
```diff
@@ -15,11 +15,17 @@ class OperationPostponedResponse(Response):
...
-            result (rq.job.Job): A :class:`rq.job.Job` object used to generate the response.
+            task (pulpcore.plugin.models.Task or rq.job.Job): A
+                :class:`~pulpcore.plugin.models.Task` or :class:`rq.job.Job` object used to generate
+                the response.
             request (rest_framework.request.Request): Request used to generate the pulp_href urls
         """
-        resp = {"task": reverse("tasks-detail", args=[result.id], request=None)}
+        try:
+            pk = task.id
+        except AttributeError:
```
dralley, line 28: What's going on here?

dralley: Ah, I see.  Just put an inline comment about which maps to which (RQ vs Task)

bmbouter: The new `dispatch` interface returns a `Task`, but the old `enqueue_with_reservation` returns a `RQ.job`. Both return types end up getting passed into `OperationPostponedResponse` (see L#21 docstring). If `id` is present it's the RQ.job style, otherwise it's the `Task. Maybe I should put a comment on this in the code?

dralley (approved):

pulpcore/app/response.py
bmbouter: I added the comment.

CHANGES/plugin_api/8496.feature
```diff
@@ -0,0 +1,7 @@
+Adds the ``pulpcore.plugin.tasking.dispatch`` interface which replcaes the
```
mdellweg, line 1:
```suggestion
Adds the ``pulpcore.plugin.tasking.dispatch`` interface which replaces the
```

mdellweg (approved):
One typo left.

PR #1227, opened 2021-04-01
Pass the request's context to the serializers

lubosmj:
In pulp_container, we are relying on the request's context when retrieving the host from a URL in ContainerDistributionSerializer. Since object-level permissions are tied to specific objects, the linked error was triggered only during the serialization of ContainerDistribution objects. Similarly, when a user tried to list permissions which contained task permissions, the missing context caused CreatedResourceSerializer to raise an internal error.

As of this commit, all RBAC-related serializers can benefit from the initial request's context.

closes #8396

### DIFF
--- a/pulpcore/app/util.py
+++ b/pulpcore/app/util.py
@@ -127,3 +127,35 @@ def get_view_urlpattern(view):
     if hasattr(view, "parent_viewset") and view.parent_viewset:
         return os.path.join(view.parent_viewset.urlpattern(), view.urlpattern())
     return view.urlpattern()
+
+
+def get_request_without_query_params(context):
+    """
+    Remove query parameters from a request object.
+
+    Removing query parameters should not influence the features provided by the library
+    'djangorestframework-queryfields'. But, once a user selects which fields should be sent in the
+    API response, any other fields are going to be filtered out even though they are still required
+    for serialization purposes in the future.
+
+    Some serializers need to be aware of the context that triggered the operation. For example, when
+    a Task is serialized, created resources need to be serialized as well. Missing context lead to
+    invalid serialization if the created resources do not have knowledge about the initial request.
+
+    This method modifies the request object in-place, meaning that all other objects that reference
+    this object may be affected.
+
+    Args:
+        context (dict): A context containing the request object.
+
+    Returns:
+        rest_framework.request.Request: a request object without query parameters
+    """
+    request = context.get("request")
+
+    if request is not None:
+        request.query_params._mutable = True
+        request.query_params.clear()
+        request.query_params._mutable = False
+
+    return request
--- a/CHANGES/8396.bugfix
+++ b/CHANGES/8396.bugfix
@@ -1,0 +1,1 @@
+Fixed an issue that caused the request's context to be ignored in the serializers.
--- a/pulpcore/app/serializers/task.py
+++ b/pulpcore/app/serializers/task.py
@@ -12,7 +12,7 @@ from pulpcore.app.serializers import (
     TaskGroupStatusCountField,
 )
 from pulpcore.constants import TASK_STATES
-from pulpcore.app.util import get_viewset_for_model
+from pulpcore.app.util import get_viewset_for_model, get_request_without_query_params
 
 
 class CreatedResourceSerializer(RelatedField):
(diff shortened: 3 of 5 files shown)


pulpcore/app/util.py
```diff
@@ -127,3 +127,10 @@ def get_view_urlpattern(view):
     if hasattr(view, "parent_viewset") and view.parent_viewset:
         return os.path.join(view.parent_viewset.urlpattern(), view.urlpattern())
     return view.urlpattern()
+
+
+def remove_query_params(request):
```
ipanova, line 132: s/remove_query_params/remove_request_query_params/

ipanova:
Looks good to me. @daviddavis  @ggainey  can any of you look into the exporter area and confirm it does not need passed request's context? I see that it is passed to the importer though.

```$ git grep -i "export.*context={"
pulpcore/app/serializers/exporter.py:                exported_resource.content_object, context={"request": None}
pulpcore/app/viewsets/exporter.py:        serializer = PulpExportSerializer(data=request.data, context={"exporter": exporter})
[ipanova@fluffy pulpcore]$ git grep -i "import.*context={"
pulpcore/app/views/importer.py:            crs = PulpImportCheckResponseSerializer(data, context={"request": request})
pulpcore/app/viewsets/importer.py:        serializer = PulpImportSerializer(data=request.data, context={"request": request})



```

ggainey:
We use request.data in a couple of places, but I don't find us to be dependent on request.query_params or context:request - in fact, we deliberately set context:request=None. I think we're good here.

lubosmj:
Since it is not very clear which Serializers/ViewSets really do require the request's context, I updated the method `remove_query_params()` to validate the passed context object before removing the request's query parameters (before this change, I was expecting the request object to be always provided for `CreatedResourceSerializer` and `PermissionField`).

ipanova (approved):

pulpcore/app/util.py
```diff
@@ -127,3 +127,18 @@ def get_view_urlpattern(view):
...
         return os.path.join(view.parent_viewset.urlpattern(), view.urlpattern())
     return view.urlpattern()
+
+
+def get_request_without_query_params(context):
+    """Return the request object without query parameters."""
+    try:
+        request = context["request"]
+    except KeyError:
+        request = None
```
mdellweg, line 137:
```suggestion
    request = context.get("request")
```

pulpcore/app/util.py
```diff
@@ -127,3 +127,18 @@ def get_view_urlpattern(view):
...
+
+    if request is not None:
+        request.query_params._mutable = True
+        request.query_params.clear()
```
mdellweg, line 141: I am worried, that we are modifying the actual request in place. (I know, that's what the code did in the old place too.)
Would it be possible to return a modified copy?
Also what is the reason to wipe the query params anyway?

lubosmj: I may try to copy a `request` object but I do not see any benefits of doing so...

The reason to wipe the query parameters is right here: https://github.com/pulp/pulpcore/pull/968#discussion_r510971808.

mdellweg: As i said, we are not only passing a modified version of the request to the calling function, but we are modifying the only existing copy of the request in the whole stack.
I also do not know if this results in an actual bug (today or in the future), but i think it's a serious side effect. And there's practically no way of knowing what else can be effected by this.

lubosmj: Copying a whole `request` object will lead to a more complex code. The object cannot be simply copied because there are fields, such as `_io.TextIOWrapper`, which cannot be serialized.

The only solution that came to my mind was to iterate over all fields which can be serialized and then to copy them, one by one. I am converting this PR to a draft for now.

mdellweg: Then we maybe should not do it in this PR.

Can you add a comment to recapitulate why we are doing it this way?
And maybe the solution is not to try solving it here, but at `pulp_href` that should be able to cope with any valid request.

mdellweg: To be specific: The only change i request in this PR is to add a comment explaining, that we cannot easily create a copy of the request without that parameters, but we need to remove them to ensure we can always see the pulp_href.

lubosmj:
@mdellweg, I added a few words here: https://github.com/pulp/pulpcore/pull/1227/files#diff-7c39d38f1aefe19ba0c60a9506034c8258a013de8b65c529c7d89c581053f05dR134-R152. Do you think I should rather move it into the code and not exhibit the reasoning to a user through the method's documentation?

mdellweg (approved):

PR #1228, opened 2021-04-02
Fix race condition in handling of reserved resources

dralley:
closes: #8352
https://pulp.plan.io/issues/8352

### DIFF
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -4,7 +4,7 @@ import time
 import uuid
 from gettext import gettext as _
 
-from django.db import IntegrityError, transaction
+from django.db import IntegrityError, transaction, models
 from django.db.models import Model
 from django_guid.middleware import GuidMiddleware
 from redis.exceptions import ConnectionError as RedisConnectionError
@@ -16,11 +16,12 @@ from pulpcore.app.models import (
     ReservedResource,
     ReservedResourceRecord,
     Task,
+    TaskReservedResource,
     TaskReservedResourceRecord,
     Worker,
 )
 from pulpcore.constants import TASK_STATES
-from pulpcore.tasking import connection, util
+from pulpcore.tasking import connection, constants, util
 
 _logger = logging.getLogger(__name__)
 
@@ -42,9 +43,9 @@ def _acquire_worker(resources):
     Raises:
         Worker.DoesNotExist: If no worker is found
     """
-    # Find a worker who already has this reservation, it is safe to send this work to them
+    # Find a worker who already has one of the reservations, it is safe to send this work to them
     try:
-        worker = Worker.objects.with_reservations(resources)
+        worker = Worker.objects.filter(reservations__resource__in=resources).distinct().get()
     except Worker.MultipleObjectsReturned:
         raise Worker.DoesNotExist
     except Worker.DoesNotExist:
@@ -52,8 +53,17 @@ def _acquire_worker(resources):
     else:
         return worker
 
-    # Otherwise, return any available worker
-    return Worker.objects.get_unreserved_worker()
+    # Otherwise, return any available worker at random
+    workers_qs = Worker.objects.online_workers().exclude(
+        name=constants.TASKING_CONSTANTS.RESOURCE_MANAGER_WORKER_NAME
+    )
+    workers_qs_with_counts = workers_qs.annotate(models.Count("reservations"))
+    try:
+        # A randomly-selected Worker instance that has zero ReservedResource entries associated with it.
+        return workers_qs_with_counts.filter(reservations__count=0).order_by("?")[0]
+    except IndexError:
+        # If all Workers have at least one ReservedResource entry.
+        raise Worker.model.DoesNotExist()
 
 
 def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwargs, options):
@@ -120,7 +130,16 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
             continue
 
         try:
-            worker.lock_resources(task_status, resources)
+            # Attempt to lock all resources by their urls. Must be atomic to prevent deadlocks.
+            with transaction.atomic():
+                for resource in resources:
+                    if worker.reservations.filter(resource=resource).exists():
+                        reservation = worker.reservations.get(resource=resource)
+                    else:
+                        reservation = ReservedResource.objects.create(
+                            worker=worker, resource=resource
+                        )
+                    TaskReservedResource.objects.create(resource=reservation, task=task_status)
         except IntegrityError:
             # we have a worker but we can't create the reservations so wait
             time.sleep(0.25)
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -31,8 +31,11 @@ TASK_TIMEOUT = -1  # -1 for infinite timeout
 
 def _acquire_worker(resources):
     """
-    Attempts to acquire a worker for a set of resource urls. If no worker has any of those resources
-    reserved, then the first available worker is returned
+    Attempts to acquire a worker for a set of resource urls. If no worker has any of those
+    resources reserved, then the first available worker is returned.
+
+    Must be done in a transaction, and locks the worker in question until the end of the
+    transaction.
 
     Arguments:
         resources (list): a list of resource urls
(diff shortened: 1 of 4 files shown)


dralley:
@bmbouter This is only half of it, I think for the resource manager bits we may want to get on a call to discuss.  I made an attempt but the complexity of the code makes it quite difficult to put the ```select_for_update``` where it needs to go.

You can go ahead and review this portion though.

daviddavis:
The code looks good to me. I am going to try to test out/debug the code some tomorrow with some sleep statements so I can understand it a bit better.

pulpcore/tasking/tasks.py
```diff
@@ -110,19 +121,29 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
                     task_status.set_failed(e, None)
                 return
 
-        try:
-            worker = _acquire_worker(resources)
-        except Worker.DoesNotExist:
-            # no worker is ready so we need to wait
-            time.sleep(0.25)
-            continue
+        with transaction.atomic():
```
dralley, line 124: I'm about 70% certain we also need a ```select_for_update()``` here... however, apparently it doesn't get along with the ```distinct()``` on line 46.  Any ideas?  Or is it not necessary?

dralley: The "simple" option might be to just ignore locking during `acquire_worker` and just drop a `worker = Worker.objects.select_for_update().get(pk=worker.pk)` on line 127.  I don't like the idea of querying twice but it would work.  And it might be "good enough" if we want a fast patch with minimal risk.

Would still like to do better though.

fao89:
@cutwater

pulpcore/tasking/tasks.py
dralley: I implemented ^

cutwater: @dralley `select_for_update` is 100% needed here to prevent related tables (resource reservation) from modification if another transaction holds a lock. An alternative to double query can be rewriting the query on line 48 to eliminate DISTINCT. However as a workaround and immediate solution it is acceptable in my opinion.

cutwater (changes requested):

pulpcore/tasking/tasks.py
```diff
@@ -112,19 +121,30 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
...
+            try:
+                worker = _acquire_worker(resources)
+
+                worker = Worker.objects.select_for_update().get(pk=worker.pk)
```
daviddavis, line 128: I wonder if this is still susceptible to a race condition? 

1. Worker is returned from `_acquire_worker()`
2. Worker is cleaned up
3. Worker is fetched here with `get()` and assigned reservations

cutwater: Good catch. Yes, it will. Here worker status is not checked.

daviddavis: Alternatively, I believe this should work:

```
Worker.objects.select_for_update.filter(pk__in=Worker.objects.filter(reservations__resource__in=resources))
```

mdellweg: Apart from this question, the rest looks good to me.

dralley: @daviddavis Is this meant to go on line 132 or on line 48?  It has different semantics from line 48.

daviddavis: I think this could potentially work for line 48:

```python
return Worker.objects.select_for_update().filter(pk__in=Worker.objects.filter(reservations__resource__in=resources)).get()
```

cutwater (approved):
LGTM

daviddavis (approved):

mdellweg (approved):

dralley:
@daviddavis This change appears as though it might be causing deadlocks?  If I run the tests locally it hangs on the very first one.

dralley:
@cutwater @mdellweg Please re-review the last patch

pulpcore/tasking/tasks.py
```diff
@@ -55,11 +62,15 @@ def _acquire_worker(resources):
...
+    workers_qs_no_res = (
+        workers_qs.annotate(models.Count("reservations"))
+        .filter(reservations__count=0)
+        .order_by("?")
+    )
     try:
         # A randomly-selected Worker instance that has zero ReservedResource entries
         # associated with it.
-        return workers_qs_with_counts.filter(reservations__count=0).order_by("?")[0]
+        return Worker.objects.select_for_update().filter(pk__in=workers_qs_no_res).first()
```
dralley, line 73: I guess the selected worker isn't necessarily random anymore.

daviddavis: Let's fix that. What if we did this:

```python
workers_qs_no_res = (
        workers_qs.annotate(models.Count("reservations"))
        .filter(reservations__count=0)
        .order_by("?")[0:1]
    )
```

daviddavis: Actually I think this is better:

```suggestion
        return Worker.objects.select_for_update().filter(pk__in=workers_qs_no_res[0:1]).first()
```

mdellweg (approved):

pulpcore/tasking/tasks.py
```diff
@@ -42,18 +46,37 @@ def _acquire_worker(resources):
...
+    try:
+        # A randomly-selected Worker instance that has zero ReservedResource entries
+        # associated with it.
+        return (
+            Worker.objects.select_for_update()
+            .filter(pk__in=workers_qs_no_res)
+            .order_by("?")
+            .first()
+        )
+    except IndexError:
```
cutwater, line 77: If query fetches no rows, `.first()` will not throw `IndexError`, but return `None`

cutwater (changes requested):

pulpcore/tasking/tasks.py
daviddavis: There is no `limit()` method for queryset. You have to use `[0:1]`.

dralley: Updated

cutwater: @daviddavis `.get()` implies `LIMIT 1`, why `[0:1]` needed here?

mdellweg: my experience says, that `get` raises the MultipleReturned exception.

cutwater: Why do you want to get an exception here instead of simply checking `first()` return value?

cutwater: Also `.get()` doesn't raise `IndexError`

pulpcore/tasking/tasks.py
```diff
@@ -42,18 +46,37 @@ def _acquire_worker(resources):
...
+        # associated with it.
+        return (
+            Worker.objects.select_for_update()
+            .filter(pk__in=workers_qs_no_res)
+            .order_by("?")[0:1]
+            .get()
+        )
+    except IndexError:
+        # If all Workers have at least one ReservedResource entry.
+        raise Worker.model.DoesNotExist()
```
cutwater:
```suggestion
    # A randomly-selected Worker instance that has zero ReservedResource entries
    # associated with it.
    return (
        Worker.objects.select_for_update()
        .filter(pk__in=workers_qs_no_res)
        .order_by("?")[0:1]
        .get()
    )
```

cutwater (changes requested):

pulpcore/tasking/tasks.py
cutwater: @mdellweg yes, my bad. `.get()` raises MultipleReturned indeed.

dralley: Early morning drowsiness :)  That's probably the simplest solution, I can do that.

pulpcore/tasking/tasks.py
cutwater: While this approach is offered by official Django docs, it looks odd.

dralley: @cutwater I went with your recommendation to just check the value of what is returned by `first()`

pulpcore/tasking/tasks.py
```diff
@@ -42,18 +46,35 @@ def _acquire_worker(resources):
...
+    )
+    workers_without_res = workers_qs.annotate(models.Count("reservations")).filter(
+        reservations__count=0
+    )
+    # A randomly-selected Worker instance that has zero ReservedResource entries.
+    worker = (
+        Worker.objects.select_for_update().filter(pk__in=workers_without_res).order_by("?").first()
+    )
+
+    if not worker:
```
cutwater, line 73: Minor comment: I'd rather use an explicit check here
```if worker is not None```

File: pulpcore/tasking/tasks.py [hunk 7]
```diff
@@ -42,18 +46,35 @@ def _acquire_worker(resources):
...
+        # If all Workers have at least one ReservedResource entry.
+        raise Worker.model.DoesNotExist()
```
daviddavis, line 75: Should this be `Worker.DoesNotExist()`?

cutwater (changes requested):

pulpcore/tasking/tasks.py
dralley: I think there's an alias - this is copied over so if it didn't work it would have broken before now.  But I'll change it anyway

pulpcore/tasking/tasks.py
dralley: Updated to `if worker is None`

pulpcore/tasking/tasks.py
```diff
@@ -113,20 +134,26 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
...
+                # Attempt to lock all resources by their urls. Must be atomic to prevent deadlocks.
+                for resource in resources:
+                    if worker.reservations.filter(resource=resource).exists():
+                        reservation = worker.reservations.get(resource=resource)
+                    else:
+                        reservation = ReservedResource.objects.create(
+                            worker=worker, resource=resource
+                        )
+                    TaskReservedResource.objects.create(resource=reservation, task=task_status)
+        except (Worker.DoesNotExist, IntegrityError):
```
cutwater, line 150: Having `break` statement at the end of the loop looks odd.

IMO this is more readable:

```

except (Worker.DoesNotExist, IntegrityError):
    # if worker is ready, or we have a worker but we can't create the reservations, wait
    time.sleep(0.25)
else:
    break
```

cutwater (changes requested):

pulpcore/tasking/tasks.py
```diff
@@ -113,17 +134,23 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
...
+            # if worker is ready, or we have a worker but we can't create the reservations, wait
             time.sleep(0.25)
             continue
```
cutwater, line 153: I meant that `continue` is not needed here anymore.

cutwater (approved):
LGTM

cutwater:
@dralley Thank you for addressing my comments.

daviddavis (approved):

dralley:
CI failure looks unrelated

dralley:
@cutwater Merged.  Do you need a new build with this patch or do you plan to upgrade to 3.12?

PR #1229, opened 2021-04-02
Add support for automatic publishing

dralley:
closes: #7626
https://pulp.plan.io/issues/7626

### DIFF
--- a/CHANGES/plugin_api/7626.feature
+++ b/CHANGES/plugin_api/7626.feature
@@ -1,0 +1,1 @@
+Added a new callback method to ``Repository`` named ``on_new_version()``, which runs when a new repository version has been created. This can be used for e.g. automatically publishing or distributing a new repository version after it has been created.
--- a/pulpcore/app/migrations/0062_add_new_distribution_mastermodel.py
+++ b/pulpcore/app/migrations/0062_add_new_distribution_mastermodel.py
@@ -25,7 +25,7 @@ class Migration(migrations.Migration):
                 ('content_guard', models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, to='core.ContentGuard')),
                 ('publication', models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, to='core.Publication')),
                 ('remote', models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, to='core.Remote')),
-                ('repository', models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, to='core.Repository')),
+                ('repository', models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, to='core.Repository', related_name='distributions')),
                 ('repository_version', models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, to='core.RepositoryVersion')),
             ],
             options={
--- a/CHANGES/7626.feature
+++ b/CHANGES/7626.feature
@@ -1,0 +1,1 @@
+Add support for automatic publishing and distributing.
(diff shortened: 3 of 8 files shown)


dralley:
@bmbouter @daviddavis

You'll notice that this implementation is a lot different (simpler) than it was before, and that I ditched PublishSettings objects entirely.

There are many reasons, but basically as I thought about it more and more, I talked myself out of the whole idea.  The reasons in favor kept getting smaller and smaller, and the reasons against got bigger and bigger.  The simplest argument is:

* Most plugins either
    * don't even do publications at all, and therefore have no need to care about publish settings
    * only have one or two settings, which might even be the default 99.9% of the time, like "manifest"

So adding this whole new thing that isn't useful for the majority of plugins is just silly and overcomplicated.  Sure, with plugins like RPM and Debian it means that now the repository has to deal with those settings, and will get kind of cluttered.  But it's "fine".  The cure was way worse than the disease.  

There's other reasons too which I can explain on a call or something.

pulpcore/app/migrations/0062_add_new_distribution_mastermodel.py
```diff
@@ -25,7 +25,7 @@ class Migration(migrations.Migration):
                 ('content_guard', models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, to='core.ContentGuard')),
                 ('publication', models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, to='core.Publication')),
                 ('remote', models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, to='core.Remote')),
-                ('repository', models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, to='core.Repository')),
+                ('repository', models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, to='core.Repository', related_name='distributions')),
```
mdellweg, line 28: Not entirely sure, if this should be a new migration.

ipanova: i believe this is an automatic one produced with 'makemigrations'

mdellweg: It is modifying a migration that is on master, but not yet released. I'm not _saying_ we should not do this, however, just asking.

ipanova: hmm i see what you're saying - i think in the past practice we'd just have a migration on top instead of modifying an unreleased one.

dralley: I can do that if there's a strong desire to.  I just thought that since this work and Brian's is so closely linked, there's not a great need to create an entire new migration just for this one-line change that I probably should have asked for on his PR initially.

So, not a strong opinion, just convenience.

CHANGES/plugin_api/7626.feature
```diff
@@ -0,0 +1 @@
+Added a new callback method to `Repository` named `on_new_version()`, which runs when a new repository version has been created. This can be used for e.g. automatically publishing or distributing a new repository version after it has been created.
```
dralley, line 1: @daviddavis How is this

pulpcore/app/migrations/0062_add_new_distribution_mastermodel.py
mdellweg: Not a strong opinion here.

CHANGES/plugin_api/7626.feature
daviddavis:
```suggestion
Added a new callback method to ``Repository`` named ``on_new_version()``, which runs when a new repository version has been created. This can be used for e.g. automatically publishing or distributing a new repository version after it has been created.
```

daviddavis (approved):

mdellweg (approved):

PR #1239, opened 2021-04-08
Disable django guardian's anonymous user.

ipanova:
closes #8524

### DIFF
--- a/pulpcore/app/apps.py
+++ b/pulpcore/app/apps.py
@@ -3,6 +3,8 @@ from gettext import gettext as _
 from importlib import import_module
 
 from django import apps
+from django.conf import settings
+from django.contrib.auth import get_user_model
 from django.core.exceptions import ImproperlyConfigured
 from django.db.models.signals import post_migrate
 from django.utils.module_loading import module_has_submodule
@@ -189,6 +191,8 @@ class PulpAppConfig(PulpPluginAppConfig):
         super().ready()
         from . import checks  # noqa
 
+        post_migrate.connect(_delete_anon_user, sender=self, dispatch_uid="delete_anon_identifier")
+
 
 def _populate_access_policies(sender, **kwargs):
     from pulpcore.app.util import get_view_urlpattern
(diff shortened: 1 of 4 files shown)


pulpcore/app/migrations/0063_delete_anon_user.py
```diff
@@ -0,0 +1,32 @@
...
+
+user_model = get_user_model()
+
+
+def delete_anon_user(apps, schema_editor):
+    try:
+        anon_user = user_model.objects.get(username='AnonymousUser')
+        anon_user.delete()
+    except user_model.DoesNotExist:
+        pass
```
goosemania, line 15: Why is delete needed if create does get_or_create? For some permissions cleanup?

pulpcore/app/settings.py
```diff
@@ -115,6 +115,10 @@
     "guardian.backends.ObjectPermissionBackend",
 ]
 
+# Disable django guardian anonymous user
+# https://django-guardian.readthedocs.io/en/stable/configuration.html#anonymous-user-name
+ANONYMOUS_USER_NAME = None
```
goosemania, line 120: If it's a big concern and we should prevent guardian anon user to be created, I think having it in the config set to None might not be safe enough. Or we need to validate that it s not set to anything else. What do you think?

CHANGES/8524.bugfix
```diff
@@ -0,0 +1 @@
+Disable django guardian's anonymous user.
```
goosemania, line 1: I think this can go to .misc, or it requires some user friendly explanation of a situation which is being fixed.

pulpcore/app/settings.py
mdellweg: I think, if a user overwrites it deliberately, then we should not block it.

goosemania: The issue description states that this fix is a security concern.
I'm lost then, why it is important to fix it and who is impacted. My understanding is that any non Django native Anonymous can't be handled properly because of DRF policies.
I can see that it may result in rejecting access for an external Anonymous user, like guardian one, but I do not understand the security concern, I think.

mdellweg: You are right. This anonymous user does not play nice with drf-access-policy. And there is nothing we can do about that.
Do you think, every option enrolled in `pulpcore/app/settings.py` is meant to be potentially overwritten by users?
What if we set it after the Dynaconf import?

pulpcore/app/apps.py
```diff
@@ -205,3 +209,14 @@ def _populate_access_policies(sender, **kwargs):
                 AccessPolicy.objects.get_or_create(
                     viewset_name=get_view_urlpattern(viewset), defaults=access_policy
                 )
+
+def _delete_anon_user(sender, **kwargs):
+    if not settings.ANONYMOUS_USER_NAME:
```
mdellweg, line 214:
```suggestion
    if settings.ANONYMOUS_USER_NAME is None:
```

mdellweg (approved):
Looks good to me.
Just the question whether we should enforce this setting is open.

goosemania (approved):

PR #1245, opened 2021-04-09
Content contained in RepoVer and Publication

pavelpicka:
Views to show repository_versions or publications which contain specific content.

closes: #4832
https://pulp.plan.io/issues/4832

### DIFF
--- a/pulpcore/app/models/publication.py
+++ b/pulpcore/app/models/publication.py
@@ -10,6 +10,30 @@ from pulpcore.app.files import PulpTemporaryUploadedFile
 from pulpcore.app.loggers import deprecation_logger
 
 
+class PublicationQuerySet(models.QuerySet):
+    """A queryset that provides publication filtering methods."""
+
+    def with_content(self, content):
+        """
+        Filters publictions that contain the provided content units.
+
+        Args:
+            content (django.db.models.QuerySet): query of content
+
+        Returns:
+            django.db.models.QuerySet: Publications that contain content.
+        """
+        pub_artifact_q = Publication.objects.filter(
+            published_artifact__content_artifact__content__pk__in=content
+        )
+        pass_thru_q = Publication.objects.filter(
+            pass_through=True,
+            repository_version__pk__in=RepositoryVersion.objects.with_content(content),
+        )
+
+        return pub_artifact_q | pass_thru_q
+
+
 class Publication(MasterModel):
     """
     A publication contains metadata and artifacts associated with content
@@ -43,6 +67,8 @@ class Publication(MasterModel):
     """
 
     TYPE = "publication"
+
+    objects = PublicationQuerySet.as_manager()
 
     complete = models.BooleanField(db_index=True, default=False)
     pass_through = models.BooleanField(default=False)
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -19,6 +19,7 @@ from pulpcore.app.util import batch_qs, get_view_name_for_model
 from pulpcore.constants import ALL_KNOWN_CONTENT_CHECKSUMS
 from pulpcore.download.factory import DownloaderFactory
 from pulpcore.exceptions import ResourceImmutableError
+from pulpcore.app.loggers import deprecation_logger
 
 from .base import MasterModel, BaseModel
 from .content import Artifact, Content
@@ -465,6 +466,31 @@ class RepositoryVersionQuerySet(models.QuerySet):
     def complete(self):
         return self.exclude(complete=False)
 
+    def with_content(self, content):
+        """
+        Filters repository versions that contain the provided content units.
+
+        Args:
+            content (django.db.models.QuerySet): query of content
+
+        Returns:
+            django.db.models.QuerySet: Repository versions which contains content.
+        """
+        query = models.Q(pk__in=[])
+        repo_content = RepositoryContent.objects.filter(content__pk__in=content)
+
+        for rc in repo_content.iterator():
+            filter = models.Q(
+                repository__pk=rc.repository.pk,
+                number__gte=rc.version_added.number,
+            )
+            if rc.version_removed:
+                filter &= models.Q(number__lt=rc.version_removed.number)
+
+            query |= filter
+
+        return self.filter(query)
+
 
 class RepositoryVersion(BaseModel):
     """
@@ -494,6 +520,8 @@ class RepositoryVersion(BaseModel):
         repository (models.ForeignKey): The associated repository.
     """
 
+    objects = RepositoryVersionQuerySet.as_manager()
+
     repository = models.ForeignKey(Repository, on_delete=models.CASCADE)
     number = models.PositiveIntegerField(db_index=True)
     complete = models.BooleanField(db_index=True, default=False)
@@ -518,6 +546,10 @@ class RepositoryVersion(BaseModel):
         Returns:
             django.db.models.QuerySet: Repository versions which contains content.
         """
+        deprecation_logger(
+            "This method is deprecated and will be removed in version 3.14. "
+            "Use RepositoryVersion.objects.with_content() instead."
+        )
         query = models.Q(pk__in=[])
         repo_content = RepositoryContent.objects.filter(content__pk__in=content)
 
--- a/pulpcore/app/urls.py
+++ b/pulpcore/app/urls.py
@@ -16,6 +16,7 @@ from rest_framework_nested import routers
 
 from pulpcore.app.apps import pulp_plugin_configs
 from pulpcore.app.views import OrphansView, PulpImporterImportCheckView, RepairView, StatusView
+from pulpcore.app.viewsets import ListRepositoryVersionViewSet
 from pulpcore.constants import API_ROOT
 from pulpcore.openapi import PulpSchemaGenerator
 
@@ -123,6 +124,10 @@ urlpatterns = [
     url(r"^{api_root}repair/".format(api_root=API_ROOT), RepairView.as_view()),
     url(r"^{api_root}status/".format(api_root=API_ROOT), StatusView.as_view()),
     url(r"^{api_root}orphans/".format(api_root=API_ROOT), OrphansView.as_view()),
+    url(
+        r"^{api_root}repository_versions/".format(api_root=API_ROOT),
+        ListRepositoryVersionViewSet.as_view({"get": "list"}),
+    ),
     url(
         r"^{api_root}importers/core/pulp/import-check/".format(api_root=API_ROOT),
         PulpImporterImportCheckView.as_view(),
--- a/pulpcore/app/viewsets/publication.py
+++ b/pulpcore/app/viewsets/publication.py
@@ -1,10 +1,17 @@
 from gettext import gettext as _
 
+from django_filters import Filter
 from django_filters.rest_framework import DjangoFilterBackend, filters
-from rest_framework import mixins
+from rest_framework import mixins, serializers
 from rest_framework.filters import OrderingFilter
 
-from pulpcore.app.models import BaseDistribution, ContentGuard, Distribution, Publication
+from pulpcore.app.models import (
+    BaseDistribution,
+    ContentGuard,
+    Distribution,
+    Publication,
+    Content,
+)
 from pulpcore.app.serializers import (
     BaseDistributionSerializer,
     ContentGuardSerializer,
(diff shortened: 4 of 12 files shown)


pulpcore/app/viewsets/repository.py
```diff
@@ -278,3 +238,18 @@ class RemoteViewSet(
     serializer_class = RemoteSerializer
     queryset = Remote.objects.all()
     filterset_class = RemoteFilter
+
+
+class ListRepositoryVersionViewSet(
+    GenericViewSet,
```
daviddavis, line 244: @mdellweg @dralley we ran into a problem when attempting to use `NamedModelViewSet` (like the other generic endpoints do) since RepositoryVersions aren't typed. [This code in util.py](https://github.com/pulp/pulpcore/blob/0f802b8953aa2d408b5907e79a755420da83173f/pulpcore/app/util.py#L34) expects a model to map to one single NamedModelViewSet. This works for Repository and other typed models since each model class maps to one viewset (eg a FileRepository maps to a FileRepositoryViewSet). However, adding another RepositoryVersion NamedModelViewSet causes RepositoryVersion to map to two viewsets.

Do you have any thoughts or suggestions?

pulpcore/app/viewsets/publication.py
```diff
@@ -39,6 +68,18 @@ class Meta:
         }
 
 
+class ListPublicationViewSet(GenericViewSet, mixins.ListModelMixin):
```
daviddavis, line 71: This can actually extend `NamedModelViewSet` like the other generic endpoints.

pulpcore/app/urls.py
```diff
@@ -123,6 +124,14 @@ def __repr__(self):
     url(r"^{api_root}repair/".format(api_root=API_ROOT), RepairView.as_view()),
     url(r"^{api_root}status/".format(api_root=API_ROOT), StatusView.as_view()),
     url(r"^{api_root}orphans/".format(api_root=API_ROOT), OrphansView.as_view()),
+    url(
+        r"^{api_root}publications_content/".format(api_root=API_ROOT),
```
daviddavis, line 128: Why `publications_content` instead of `publications`?

pulpcore/app/models/repository.py
```diff
@@ -504,6 +532,9 @@ def versions_containing_content(cls, content):
         Returns:
             django.db.models.QuerySet: Repository versions which contains content.
         """
+        warnings.warn(
```
daviddavis, line 535: There's now a deprecation logger. See https://github.com/pulp/pulpcore/pull/1225.

daviddavis:
Some tests would be lovely :smiley_cat:

pulpcore/app/urls.py
pavelpicka: there was conflict with NamedModelViewSets. This gets removed by addressing comment above.

pulpcore/app/models/repository.py
```diff
@@ -504,6 +532,9 @@ def versions_containing_content(cls, content):
         Returns:
             django.db.models.QuerySet: Repository versions which contains content.
         """
+        deprecation_logger(
+            "This method is deprecated. Use RepositoryVersion.objects.with_content() instead."
```
mdellweg, line 536: We usually state at least the version when we deprecate it, so we can easily see when it's save to remove.

pulpcore/app/viewsets/repository.py
mdellweg: And the problem being, that it evades the automatic viewset discovery?

daviddavis: Yea, I'm also concerned that there might be other potential problems now or in the future I guess.

mdellweg: But there is a `FileRepositoryVersionViewSet` already. So we _do_ have multiple viewsets for repository_versions.
There must be something i do not understand.

pavelpicka: Those are mapped to repository, based on type of plugin (this case `File..`) `.../repository/**file**/file/UUID/version/[0-9]`

daviddavis: I made a little PoC to demonstrate the problem. Check out this branch: https://github.com/daviddavis/pulpcore/tree/repo-versions-endpoint

Then run this:

```
In [1]: from pulpcore.app.util import get_viewset_for_model

In [2]: get_viewset_for_model(FileRepository())
Out[2]: pulp_file.app.viewsets.FileRepositoryViewSet

In [3]: get_viewset_for_model(RepositoryVersion())
LookupError: Could not determine ViewSet base name for model <class 'pulpcore.app.models.repository.RepositoryVersion'>
```

mdellweg: And now i see, it is only that you added the second viewset in the _same_ app, what's causing the lookup in utils to fail.
So `get_viewset_for_model(RepositoryVersion())` on my unpatched system returns `pulpcore.app.viewsets.repository.RepositoryVersionViewSet` because pulpcore just happens to be the first app in `pulp_plugin_configs()`. If it were looking at pulp_file, it would see `pulpcore.app.models.repository.RepositoryVersion: [pulp_file.app.viewsets.FileRepositoryVersionViewSet,               pulpcore.app.viewsets.repository.RepositoryVersionViewSet],` and fail as well.

daviddavis: Yes, what's also strange is that `get_viewset_for_model(Repository())` works even though core has four matching viewsets (ImmutableRepositoryViewSet, ListRepositoryViewSet, ReadOnlyRepositoryViewSet, RepositoryViewSet). I believe it uses plugins' named_viewsets:

```
In [11]: from pulpcore.app.apps import pulp_plugin_configs

In [12]: pulp_file_config = [c for c in pulp_plugin_configs() if c.label == "file"][0]

In [13]: pulp_file_config.named_viewsets[Repository]
Out[13]: [pulpcore.app.viewsets.repository.RepositoryViewSet]
```

daviddavis: Yea, I just uninstalled all plugins and ran:

```
In [1]: from pulpcore.app.util import get_viewset_for_model

In [2]: get_viewset_for_model(Repository())

LookupError: Could not determine ViewSet base name for model <class 'pulpcore.app.models.repository.Repository'>
```

I filed a bug here: https://pulp.plan.io/issues/8537

daviddavis: Can you add a comment here saying that we had to use `GenericViewSet` as `NamedModelViewSet` caused `get_viewset_for_model()` to match multiple viewsets?

daviddavis (approved):
One suggestion for a comment about `ListRepositoryVersionViewSet`. Other then that, LGTM.

pavelpicka:
@daviddavis comment about viewset added.

mdellweg (approved):

pulpcore/app/models/publication.py
```diff
@@ -10,6 +10,29 @@
...
+        Args:
+            content (django.db.models.QuerySet): query of content
+
+        Returns:
+            django.db.models.QuerySet: Repository versions which contains content.
+        """
+        query = models.Q(pk__in=[])
+        repo_versions = RepositoryVersion.objects.with_content(content)
+        for rv in repo_versions.iterator():
+            publication_query = Publication.objects.filter(repository_version=rv)
```
dralley, line 29: This is not strictly correct.  I believe the Debian plugin can filter content at Publication time.  @mdellweg is that correct?

I think the best way to implement this may actually be to go through artifacts rather than content.  Publications have a direct link to Artifacts, and we can get a list of Artifacts fairly easily from a list of content.

dralley: That would be two separate methods FYI.

mdellweg: That is correct. "normal" publications for example ignore the cached metadata content units.

Can you delete the repository version a publication is made from?
If not, the publication would still kind of transitively contain everything in the repo-version.

daviddavis: PublishedArtifact [has a link to Content through ContentArtifact](https://github.com/pulp/pulpcore/blob/87f6ee9fa4f7afb8887f71816ca97f1333b6d27f/pulpcore/app/models/publication.py#L162). I believe we could use that? We'd still need to handle pass_through publications which can contain content for which they have no published artifacts.

I believe that publications get deleted when repo versions do if I am reading [this](https://github.com/pulp/pulpcore/blob/87f6ee9fa4f7afb8887f71816ca97f1333b6d27f/pulpcore/app/models/publication.py#L50) correctly.

dralley: Where are pass through publications used?

daviddavis: I know at least [pulp_file](https://github.com/pulp/pulp_file/blob/0e56b0f4c6dca9161c1216c6f16efa1b9d11fc4b/pulp_file/app/tasks/publishing.py#L40) and it looks like [pulp_deb](https://github.com/pulp/pulp_deb/blob/bbfa359ddc654d5dee6696319cdae06c37beee93/pulp_deb/app/tasks/publishing.py#L61) too. pulp_ansible probably will too if/when we switch to publications.

I wrote up a few lines to cover both cases (published artifacts and pass through publications):

```python
pub_artifact_q = Publication.objects.filter(published_artifact__content_artifact__content__pk__in=content)
pass_thru_q = Publication.objects.filter(pass_through=True, repository_version__pk__in=RepositoryVersion.objects.with_content(content))

return pub_artifact_q | pass_thru_q
```

pulpcore/app/models/repository.py
```diff
@@ -504,6 +532,10 @@ def versions_containing_content(cls, content):
         Returns:
             django.db.models.QuerySet: Repository versions which contains content.
         """
+        deprecation_logger(
```
ipanova, line 535: this needs a changelog entry

pulpcore/app/models/publication.py
daviddavis: @dalley @mdellweg does this make sense ^?

pulpcore/app/models/repository.py
pavelpicka: Changelog entry extended by method goes deprecated message.

daviddavis: The changes listed in `CHANGES` are meant for end users. We need a changelog entry for plugin writers (in `CHANGES/plugin_api`). Also, it should use the `.deprecation` extension (`CHANGES/plugin_api/4832.deprecation`)

daviddavis:
I just noticed that the tests don't use bindings. It's recommended new tests should use the bindings.

pulpcore/app/models/repository.py
pavelpicka: changelog divided to .feature and .deprecation

pavelpicka:
tests are using bindings now

daviddavis (approved):
LGTM. Thank you for converting the tests.

@mdellweg @ipanova want to re-review?

pulpcore/app/viewsets/repository.py
```diff
@@ -278,3 +238,20 @@ class RemoteViewSet(
...
+):
+    endpoint_name = "repository_versions"
+    serializer_class = RepositoryVersionSerializer
+    queryset = RepositoryVersion.objects.all()
+    filterset_class = RepositoryVersionFilter
+
+    @classmethod
+    def is_master_viewset(cls):
+        """Do not hide from the routers."""
+        return False
```
mdellweg, line 257: Given the inheritance here, this function serves no purpose anymore.

mdellweg (commented):
There is one superfluous method now. But my approval stands.

pulpcore/app/models/publication.py
```diff
@@ -10,6 +10,30 @@
...
+
+    def with_content(self, content):
+        """
+        Filters publictions that contain the provided content units.
+
+        Args:
+            content (django.db.models.QuerySet): query of content
+
+        Returns:
+            django.db.models.QuerySet: Repository versions which contains content.
```
ipanova, line 24: s/Repository versions which container/Publications that contain

pulpcore/app/viewsets/repository.py
```diff
@@ -132,46 +131,7 @@ def filter(self, qs, value):
...
-        )
-
-        # Get the ranges between paired version_added and version_removed to get all
-        # the versions the content is present in.
-        versions = [list(range(added, removed)) for (added, removed) in version_tuples]
-        # Flatten the list of lists
-        versions = list(itertools.chain.from_iterable(versions))
-
-        return qs.filter(number__in=versions)
+        return qs.with_content([content.pk])
```
ipanova, line 134: the doc block for `RepositoryVersionContentFilter` is outdated

ipanova (approved):
I did some basic testing it seems to do the job :)

PR #1246, opened 2021-04-09
Update CI files

daviddavis:
[noissue]


goosemania (approved):

mdellweg:
I think, when i prepared a release for such an old branch, we decided to use an older version of the plugin_template, because the change in the status api was breaking one or two things.
OTOH I cannot see anything too bad. Did you handpick some changes here?

daviddavis:
> I think, when i prepared a release for such an old branch, we decided to use an older version of the plugin_template, because the change in the status api was breaking one or two things.

This makes sense. I forgot about the changes to the status api.

> OTOH I cannot see anything too bad. Did you handpick some changes here?

The only change I made by hand was commenting out this line again:

https://github.com/pulp/pulpcore/blob/7b33cbdd5d6cf3e888607f5f82e71d48b4fc847f/.github/workflows/scripts/script.sh#L37

It looks like though on 3.7 that [the plugin_app_label was set to pulpcore instead of core](https://git.io/JOeQJ) which is perhaps why the status api changes aren't reflected here.

Though the CI jobs are failing so I think it makes sense not to update the CI files?

PR #1248, opened 2021-04-09
CP for 3.7.5

goosemania:

### DIFF
--- a/.github/workflows/scripts/before_install.sh
+++ b/.github/workflows/scripts/before_install.sh
@@ -96,7 +96,7 @@ if [ -n "$PULP_FILE_PR_NUMBER" ]; then
   cd ..
 fi
 
-git clone --depth=1 https://github.com/pulp/pulp-certguard.git --branch master
+git clone --depth=1 https://github.com/pulp/pulp-certguard.git --branch 1.1
 if [ -n "$PULP_CERTGUARD_PR_NUMBER" ]; then
   cd pulp-certguard
   git fetch --depth=1 origin pull/$PULP_CERTGUARD_PR_NUMBER/head:$PULP_CERTGUARD_PR_NUMBER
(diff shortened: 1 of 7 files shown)


daviddavis:
It looks like this needs to be updated: https://github.com/pulp/pulpcore/blob/7b33cbdd5d6cf3e888607f5f82e71d48b4fc847f/template_config.yml#L7

dralley:
Given that we know there is a hole in the previous QueryExistingContent artifact that was fixed by https://github.com/pulp/pulpcore/commit/50c6705941c5f249fdacda6a52f0bd0c0efa453c, should we backport it?

goosemania:
> It looks like this needs to be updated:
> 
> https://github.com/pulp/pulpcore/blob/7b33cbdd5d6cf3e888607f5f82e71d48b4fc847f/template_config.yml#L7

yup, on it

goosemania:
> Given that we know there is a hole in the previous QueryExistingContent artifact that was fixed by [50c6705](https://github.com/pulp/pulpcore/commit/50c6705941c5f249fdacda6a52f0bd0c0efa453c), should we backport it?

Is it possible to run into the problem with the migration case? Or is it only for sync? 
If you think it is worth backporting, please file a backport and an explanation of the situation in which the bug can be encountered.

dralley:
Whether it gets hit during migrations probably depends on how well Pulp 2 does deduplication.

> As Pulp started looking through the stream of packages to pick out which ones it already knows about and replace them with the existing ones, so that it knows not to save them later, both packages ended up in the same "batch". When we hit the first package, we correctly noticed it was a duplicate, but then we skipped all the rest of the packages, as an optimization, because we didn't expect it to be present twice.  So it *doesn't* get replaced, and so later on we try to save it, and then it blows up.

https://github.com/pulp/pulpcore/blob/3.7/pulpcore/plugin/stages/content_stages.py#L58

goosemania:
@dalley and I agreed not to include content stages commit, because katello seems to start moving to 3.11 in 3.18, see https://pulp.plan.io/issues/8534 .

dralley (approved):

PR #1259, opened 2021-04-13
Add sequential steps for storage docs

melcorr:
If it is OK with everyone, I would like to change the style of these docs to a set of sequential steps with the use of imperative voice. 
I would also like to be more specific. In these Storage docs, there is a mention of "Pulp settings". Is this draft, I've used settings.py although I know that's more than likely wrong. Can we be specific here about the exact file? 

fixes #8477

part of general installation work

### DIFF
--- a/docs/installation/storage.rst
+++ b/docs/installation/storage.rst
@@ -10,6 +10,9 @@ Storage
   to use another storage backend such as Amazon Simple Storage Service (S3), you'll need to
   configure Pulp.
 
+  You can also configure Pulp to use Amazon S3 and Azure storage using the Pulp installer. For more information
+  see the `Pulp installer documentation <https://pulp-installer.readthedocs.io/en/latest/quickstart/#storage>`_
+
 Amazon S3
 ^^^^^^^^^
 
@@ -24,22 +27,20 @@ To complete these steps, consult the official Amazon S3 documentation.
 3. In AWS Identity and Access Management (IAM), create a user that Pulp can use to access your S3 bucket.
 4. Save the access key id and secret access key.
 
-Configuring Pulp
-----------------
+Configuring Pulp to use Amazon S3
+---------------------------------
 
-  To have Pulp use S3, you'll need to install the optional django-storages and boto3 Python packages in the pulp
-  virtual environment::
+To have Pulp use S3, complete the following steps:
+
+1. Install the optional django-storages and boto3 Python packages in the pulp virtual environment::
 
       pip install django-storages[boto3]
 
-  Next you'll need to set ``DEFAULT_FILE_STORAGE`` to ``storages.backends.s3boto3.S3Boto3Storage``
-  in your Pulp settings. At a minimum, you'll also need to set ``AWS_ACCESS_KEY_ID``,
-  ``AWS_SECRET_ACCESS_KEY``, and ``AWS_STORAGE_BUCKET_NAME``. For more S3 configuration options, see
-  the `django-storages documents <https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html>`_.
+2. Depending on which method you use to install or configure Pulp, you must set ``DEFAULT_FILE_STORAGE`` to ``storages.backends.s3boto3.S3Boto3Storage`` in Pulp Settings. For example, if you use the `Pulp installer <https://pulp-installer.readthedocs.io/en/latest/quickstart/>`_, the ``default_file_storage`` is part of the ``pulp_settings`` Ansible variables you can define in your Ansible playbook.
 
-  You will also want to set the ``MEDIA_ROOT`` configuration option. This will be the path in your
-  bucket that Pulp will use. An empty string is acceptable as well if you want Pulp to create its
-  folders in the top level of the bucket.
+3. In that same way, add your Amazon S3 configuration settings to ``AWS_ACCESS_KEY_ID``, ``AWS_SECRET_ACCESS_KEY``, and ``AWS_STORAGE_BUCKET_NAME``. For more S3 configuration options, see the `django-storages documents <https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html>`_.
+
+4. Set the ``MEDIA_ROOT`` configuration option. This will be the path in your bucket that Pulp will use. If you want Pulp to create its folders in the top level of the bucket, an empty string is acceptable.
 
   Here is an example configuration that will use a bucket called ``pulp3`` that is hosted in
   region ``eu-central-1``::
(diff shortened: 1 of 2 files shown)


melcorr:
> Warning: Issue [#8477](https://pulp.plan.io/issues/8477) is not at NEW/ASSIGNED/POST.

Is it bad that I'm using an umbrella task for tangential updates to docs?

mdellweg:
> > Warning: Issue [#8477](https://pulp.plan.io/issues/8477) is not at NEW/ASSIGNED/POST.
> 
> Is it bad that I'm using an umbrella task for tangential updates to docs?

This sounds, like you should use an epic task with subtasks.

melcorr:
> > > Warning: Issue [#8477](https://pulp.plan.io/issues/8477) is not at NEW/ASSIGNED/POST.
> > 
> > 
> > Is it bad that I'm using an umbrella task for tangential updates to docs?
> 
> This sounds, like you should use an epic task with subtasks.

For me this would mean spending equal amounts of time creating subtasks as actually implementing the work, but I can I suppose if I have to !

docs/installation/storage.rst
```diff
@@ -24,22 +27,22 @@ To complete these steps, consult the official Amazon S3 documentation.
...
-  To have Pulp use S3, you'll need to install the optional django-storages and boto3 Python packages in the pulp
-  virtual environment::
+1. Install the optional django-storages and boto3 Python packages in the pulp virtual environment::
 
       pip install django-storages[boto3]
 
-  Next you'll need to set ``DEFAULT_FILE_STORAGE`` to ``storages.backends.s3boto3.S3Boto3Storage``
-  in your Pulp settings. At a minimum, you'll also need to set ``AWS_ACCESS_KEY_ID``,
-  ``AWS_SECRET_ACCESS_KEY``, and ``AWS_STORAGE_BUCKET_NAME``. For more S3 configuration options, see
+2. In your Pulp `settings.py` file, set ``DEFAULT_FILE_STORAGE`` to ``storages.backends.s3boto3.S3Boto3Storage``.
```
mdellweg, line 39: Hmm, usually this file lives in `/etc/pulp/settings.py`, with the container we inject it via a volume and in the end, we are using dynaconf, which to my knowledge brings a lot of options to provide settings.
So referring to "Pulp settings" is probably the most generic way. (Yes this begs for a section called "1001 ways to configure pulp.")

melcorr: Are you saying that this is a generic statement that covers every possible installation type rather than specific instructions? 
Could we even have a generic section at the beginning of the Storage configuration docs where we give a number of options, or one option of where they'd set these values? A recommended option per deployment type? I dunno, something other than a generic statement and hope they figure it out. 
How does a user know where to put these Pulp settings? Or does the average Pulp user just *know*? The Ansible installer docs have some way of being more specific.

melcorr: @mdellweg  I have updated this, calling out an example, but ultimately reverting to the generic way, while trying to alert the reader to the fact that it is generic. 

I also created an issue to correspond. 

Let me know your thoughts.

mdellweg: I think, maybe we could use an even more generic section how to configure settings for pulp.
I do not think that all users _just know_.

melcorr: I would be very interested in helping out with a Settings overview section! 
Thank you for teaching me all these things!

docs/installation/storage.rst
```diff
@@ -73,27 +74,30 @@ Azure Blob storage
...
+1. Install the optional django-storages[azure] package in the pulp virtual environment::
 
       pip install django-storages[azure]
 
-  Next you'll need to set the ``DEFAULT_FILE_STORAGE`` to
-  ``storages.backends.azure_storage.AzureStorage`` in your pulp settings, and configure following
-  parameters in your pulp settings file (for a comprehensive overview of all possible options for
-  the Azure Blob storage backend see the `django-storages[azure] documents
-  <https://django-storages.readthedocs.io/en/latest/backends/azure.html>_`)::
+2. In the set the `settings.py` file, set ``DEFAULT_FILE_STORAGE`` to ``storages.backends.azure_storage.AzureStorage``.
```
ipanova, line 92: s/In the set the/In the `settings.py` file/

melcorr: This is a good catch Ina because I need to rewrite this as per line 39. Thank you.

ipanova (approved):
I like sequential steps and imperative voice - brings much more clarity then we had before.

docs/installation/storage.rst
```diff
@@ -24,22 +27,20 @@ To complete these steps, consult the official Amazon S3 documentation.
...
 
       pip install django-storages[boto3]
 
-  Next you'll need to set ``DEFAULT_FILE_STORAGE`` to ``storages.backends.s3boto3.S3Boto3Storage``
-  in your Pulp settings. At a minimum, you'll also need to set ``AWS_ACCESS_KEY_ID``,
-  ``AWS_SECRET_ACCESS_KEY``, and ``AWS_STORAGE_BUCKET_NAME``. For more S3 configuration options, see
-  the `django-storages documents <https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html>`_.
+2. Depending on which method you use to install or configure Pulp, you must set ``DEFAULT_FILE_STORAGE`` to ``storages.backends.s3boto3.S3Boto3Storage`` in Pulp Settings. For example, if you use the `Pulp installer <https://pulp-installer.readthedocs.io/en/latest/quickstart/>`_, the ``default_file_storage`` is part of the ``pulp_settings`` Ansible variables you can define in your Ansible playbook.
+
+3. In that same file, add your Amazon S3 configuration settings to ``AWS_ACCESS_KEY_ID``, ``AWS_SECRET_ACCESS_KEY``, and ``AWS_STORAGE_BUCKET_NAME``. For more S3 configuration options, see the `django-storages documents <https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html>`_.
```
mdellweg, line 41:
```suggestion
3. In the same way, add your Amazon S3 configuration settings to ``AWS_ACCESS_KEY_ID``, ``AWS_SECRET_ACCESS_KEY``, and ``AWS_STORAGE_BUCKET_NAME``. For more S3 configuration options, see the `django-storages documents <https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html>`_.
```
I'm referring to the very general "Pulp Settings" here.

docs/installation/storage.rst
```diff
@@ -73,27 +74,30 @@ Azure Blob storage
...
 
       pip install django-storages[azure]
 
-  Next you'll need to set the ``DEFAULT_FILE_STORAGE`` to
-  ``storages.backends.azure_storage.AzureStorage`` in your pulp settings, and configure following
-  parameters in your pulp settings file (for a comprehensive overview of all possible options for
-  the Azure Blob storage backend see the `django-storages[azure] documents
-  <https://django-storages.readthedocs.io/en/latest/backends/azure.html>_`)::
+2. Depending on which method you use to install or configure Pulp, you must set ``DEFAULT_FILE_STORAGE`` to ``storages.backends.azure_storage.AzureStorage`` in Pulp Settings. For example, if you use the `Pulp installer <https://pulp-installer.readthedocs.io/en/latest/quickstart/>`_, the ``default_file_storage`` is part of the ``pulp_settings`` Ansible variables you can define in your Ansible playbook.
+3. In the same file, configure the following parameters::
```
mdellweg, line 93: Same suggestion here.

mdellweg (approved):
Thank you!
Consider my last comments as pure suggestions. They may be more pedantic than helpful.

docs/installation/storage.rst
melcorr: Thanks, applied!

PR #1261, opened 2021-04-16
Tasking system redesign

mdellweg:
This is supposed to be the successor of:
https://github.com/pulp/pulpcore/pull/1222


pulpcore/tasking/tasks.py
```diff
@@ -230,6 +216,32 @@ def enqueue_with_reservation(
...
+        deprecation_logger.warn(
+            _(
+                "The argument {obj} is of type {type}, which is not JSON serializable. The use "
+                "of non JSON serializable objects for `args` and `kwargs` to tasks is "
+                "deprecated as of pulpcore==3.12. See the traceback below for more info."
+            ).format(obj=obj, type=type(obj))
+        )
+        if isinstance(obj, uuid.UUID):
+            return str(obj)
+        return json.JSONEncoder.default(self, obj)
```
mdellweg, line 242:
```suggestion
    def default(self, obj):
        if isinstance(obj, uuid.UUID):
            deprecation_logger.warn(
                _(
                    "The argument {obj} is of type {type}, which is not JSON serializable. The use "
                    "of non JSON serializable objects for `args` and `kwargs` to tasks is "
                    "deprecated as of pulpcore==3.12. See the traceback below for more info."
                ).format(obj=obj, type=type(obj))
            )
            return str(obj)
        return json.JSONEncoder.default(self, obj)
```

Also we could just not warn about it and keep the conversion of UUIDs as a convenience to our plugin writers.

pulpcore/tasking/worker.py
```diff
@@ -201,3 +216,107 @@ def handle_warm_shutdown_request(self, *args, **kwargs):
...
+
+
+def worker():
+    """A Pulp worker."""
+    NewPulpWorker().run_forever()
+
+
+def _hash_to_u64(value):
+    _digest = blake2s(value.encode(), digest_size=8).digest()
+    return int.from_bytes(_digest, byteorder='big', signed=True)
```
mdellweg, line 228: @bmbouter The signed here is probably what stopped your implementation from working.

pulpcore/tasking/worker.py
```diff
@@ -201,3 +216,107 @@ def handle_warm_shutdown_request(self, *args, **kwargs):
...
+        for required_resource in task.new_reserved_resources_record:
+            if required_resource in taken_resources:
+                return False
+        return True
+
+    def lock_task_resources(self, task):
+        with connection.cursor() as cursor:
+            required_resources = list(task.new_reserved_resources_record)
+            # TODO Is it necessary / useful to lock on the task too?
+            required_resources.append(util.get_url(task))
```
mdellweg, line 260: I thought it might be a good idea to lock on the task too, because we are about to change its state.

pulpcore/tasking/worker.py
```diff
@@ -201,3 +216,107 @@ def handle_warm_shutdown_request(self, *args, **kwargs):
...
+            # Sort locks to prevent dead locks
+            required_locks = sorted((_hash_to_u64(resource) for resource in required_resources))
```
mdellweg, line 262: Actually, since we are using `pg_try_advisory_lock`, maybe there cannot be deadlocks anyway.

pulpcore/tasking/worker.py
```diff
@@ -201,3 +216,107 @@ def handle_warm_shutdown_request(self, *args, **kwargs):
...
+                            self.perform_task(task)
+                        else:
+                            for required_resources in task.new_reserved_resources_record:
+                                taken_resources.add(required_resources)
+                            continue
+                time.sleep(0.2)
+        finally:
+            self._running = False
+
+    def perform_task(self, task):
```
mdellweg, line 293: Should we wrap this function into a Thread, or even subprocess to ensure memory cleanup and postgres session cleanup?

pulpcore/tasking/tasks.py
```diff
@@ -311,7 +318,24 @@ def dispatch(func, resources, args=None, kwargs=None, task_group=None):
...
+        resources = list(_validate_and_get_resources_as_set(resources))
+        with transaction.atomic():
+            task = Task.objects.create(
+                state=TASK_STATES.WAITING,
+                logging_cid=(GuidMiddleware.get_guid() or ""),
+                task_group=task_group,
+                name=f"{func.__module__}.{func.__name__}",
+                args=args_as_json,
+                kwargs=kwargs_as_json,
+                parent_task=None,  # TODO implement me!
```
mdellweg, line 333: How is a parent task defined?
Can a parent task be considered finished before all child tasks are?

pulpcore/tasking/worker.py
```diff
@@ -201,3 +214,144 @@ def handle_warm_shutdown_request(self, *args, **kwargs):
...
+        asyncio.set_event_loop(asyncio.new_event_loop())
+        while self._running:
+            taken_resources = set()
+            # When batching this query, be sure to use "pulp_created" as a cursor
+            pending_tasks = Task.objects.filter(
+                state__in=[TASK_STATES.RUNNING, TASK_STATES.WAITING]
+            ).order_by("pulp_created")
+            for task in pending_tasks:
+                try:
+                    with AdvisoryLock(_task_lock(task)):
```
mdellweg, line 298: Would be nice if python had something like a ternary context manager, or an if with a finally.

pulpcore/app/models/task.py
```diff
@@ -257,7 +258,12 @@ class Task(BaseModel, AutoDeleteObjPermsMixin, AutoAddObjPermsMixin):
         logging_cid (models.CharField): The logging CID associated with the task
         started_at (models.DateTimeField): The time the task started executing
         finished_at (models.DateTimeField): The time the task finished executing
-        error (pulpcore.app.fields.JSONField): Fatal errors generated by the task
+        error (django.contrib.postgres.fields.JSONField): Fatal errors generated by the task
+        args (django.contrib.postgres.fields.JSONField): The JSON serialized arguments for the task
+        kwargs (django.contrib.postgres.fields.JSONField): The JSON serialized keyword arguments for
+            the task
+        required_resources (django.contrib.postgres.fields.JSONField): The reserved
+            resources required for the task.
```
dralley, line 266: How is the function (or path of the function) stored?

pulpcore/tasking/tasks.py
```diff
@@ -311,7 +322,31 @@ def dispatch(func, resources, args=None, kwargs=None, task_group=None):
...
+    if settings.USE_NEW_WORKER_TYPE:
+        args_as_json = json.dumps(args, cls=UUIDEncoder)
+        kwargs_as_json = json.dumps(kwargs, cls=UUIDEncoder)
+        resources = list(_validate_and_get_resources_as_set(resources))
+        with transaction.atomic():
+            task = Task.objects.create(
+                state=TASK_STATES.WAITING,
+                logging_cid=(GuidMiddleware.get_guid() or ""),
+                task_group=task_group,
+                name=f"{func.__module__}.{func.__name__}",
```
dralley, line 334: Ah, stored here?

mdellweg:
There are still 2 tests failing. I think it is in the way, task_groups are handled. I could use some help.

pulpcore/tasking/worker.py
mdellweg: After good discussions, we came to the conclusion, that:
1. It is absolutely necessary to lock on the task.
2. We don't even need to lock on the resources individually.
3. The task locks provide a nice way to identify abandoned tasks, that need cleanup.

pulpcore/app/models/task.py
```diff
@@ -281,11 +291,12 @@ class Task(BaseModel, AutoDeleteObjPermsMixin, AutoAddObjPermsMixin):
     task_group = models.ForeignKey(
         "TaskGroup", null=True, related_name="tasks", on_delete=models.SET_NULL
     )
+    required_resources = JSONField(null=True)
```
mdellweg, line 294: Should we make this an array field?
https://docs.djangoproject.com/en/3.1/ref/contrib/postgres/fields/#django.contrib.postgres.fields.ArrayField

mdellweg:
Task canceling is not yet implemented. But there only seems to be one blinker test in the the cli part that is (sometimes) able to find that inability.

pulpcore/tasking/worker.py
dralley: That's actually a good question.  My initial response is that it needs to be a subprocess, because things can go Badly Wrong (segfault, OOM killed, unexpected errors, memory leaks) and because we need a way to hard-kill it.

But I believe we have systemd auto-restart service turned on (or could do so, if we don't already), so if something goes Badly Wrong we could *possibly* just let the whole worker die and be respawned by systemd.  That might have further implications we would need to think through though.  We'd be churning the "online workers".

pulpcore/tasking/tasks.py
dralley: parent task = if a task was spawned inside another task, that task is a child, and the task that spawned it is the parent

yes, parent can be marked finished before children.

CHANGES/plugin_api/8501.removal
```diff
@@ -0,0 +1 @@
+The usage of non-JSON serializable types of ``args`` and ``kwargs`` to tasks is no longer supported.
```
dralley, line 1: I would mention that `uuid.UUID` objects get automatically cast to strings.

bmbouter: Oh yeah I think we wanted to remove that if we could, but I'm out of touch with what version that would be done. I couldn't do it when I made the original PoC at that point yet.

mdellweg: The question is, whether we explicitly allow the automatic conversion of UUID because it makes the life of plugin writers easier, and we can provide the conversion without risk.

docs/settings.rst
```diff
@@ -304,12 +308,29 @@ ALLOWED_CONTENT_CHECKSUMS
       If Pulp fails to start because forbidden checkums have been identified or required ones are
       missing, run ``pulpcore-manager handle-artifact-checksums`` command.
 
+
+.. _use-non-rq-pulp-workers:
+
+USE_NEW_WORKER_TYPE
+^^^^^^^^^^^^^^^^^^^
+
+    Pulp has a new experimental worker type which is startable with the ``pulpcore-worker`` command. If
```
ipanova, line 317: Pulp has a new experimental worker type, which if set to True, starts workers with the ``pulpcore-worker`` command. Defaults to ``False``.

mdellweg: That is a good observation. What you describe would be ideal: The user changes the option and restarts pulp, and the startup scripts do the-right-thing (tm).
The reality however today, is that the duty of starting the right worker type is upon the user / the installer / the container setup / the <fill in your deployment mechanism>.
It is even worse today, because the "wrong type of worker" will fail to start up and the systemd files will respawn them infinitely.
I can start to work on the installer side right away, but the problem is that we can never assume, that we can cover _all_ deployments here.
I can however try to merge the old and the new type worker into a single entrypoint.

BTW The single container is already doing the-right-thing.

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -0,0 +1,212 @@
...
+                    (
+                        item.channel == "pulp_worker_wakeup"
+                        for item in connection.connection.notifies
+                    )
+                ):
+                    connection.connection.notifies.clear()
+                    break
+        self.cursor.execute("UNLISTEN pulp_worker_wakeup")
+
+    def superwise_task(self, task):
```
ipanova, line 121: unless the task is really meant to be super wise i think it should be `supervise` :D

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -0,0 +1,212 @@
...
+    """A Pulp worker."""
+    if not settings.USE_NEW_WORKER_TYPE:
+        raise NotImplementedError("This worker is not supposed to run with old style tasks.")
+    NewPulpWorker().run_forever()
+
+
+class NewPulpWorker:
+    def __init__(self):
+        self.name = f"{os.getpid()}@{socket.getfqdn()}"
+        self.worker, created = Worker.objects.get_or_create(name=self.name)
```
ipanova, line 49: Worker.objects.get_or_create(name=self.name)[0], i don't see `created` to be used anywhere

docs/settings.rst
bmbouter: That would be ideal, but we would have to have everyone switch to the new version since that is the long-lived one. That may be doable? Is this what you were thinking @mdellweg?

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -0,0 +1,212 @@
...
+                        self.superwise_task(task)
+                    finally:
+                        self.cursor.execute(f"SELECT pg_advisory_unlock{suffix}(1234)")
+                self.sleep()
+
+
+def _perform_task(task_pk):
+    """This must be called as a subprocess, while the parent holds the advisory lock."""
+    # All processes need to create their own postgres connection
+    connection.connection = None
```
ipanova, line 173: with all these connections opened should not we also close them?

docs/settings.rst
mdellweg: Well, we'd had a single entrypoint in pulpcore setup.py that loads either worker type. And we 'd need the deployments to switch to that entrypoint. Then switching is just a matter of the configuration.
Yes, i think this is doable.

pulpcore/tasking/pulpcore_worker.py
mdellweg: Closing them is the real problem. If you close the connection in the subprocess, the main process will complain that someone closed its connection (and the locks would be gone).
Forgetting about the connection in the subprocess seems to work. And they won't pile up, as it is always the same connection that is going to be forgotton in the subprocess.

docs/settings.rst
mdellweg: I'm not quite sure what to do with the resource-manager however.
If we just close the process, it will keep respawning.

mdellweg: https://github.com/pulp/pulpcore/pull/1261/commits/c33214abf808cca7a0643a6c9a515fdb6572a018 is what i came up with so far. There are probably a bunch of rq config things more to consider.

pulpcore/tasking/pulpcore_worker.py
mdellweg: Maybe we should just call `self.beat()` here, anyway.

pulpcore/tasking/worker_wrapper.py
```diff
@@ -0,0 +1,37 @@
...
+
+        NewPulpWorker().run_forever()
+    else:
+        _logger.info("Starting queued type worker")
+        from rq import Connection
+        from pulpcore.tasking.worker import PulpWorker
+
+        with Connection():
+            worker = PulpWorker("default")
+            worker.work()
```
mdellweg, line 37: @dralley @bmbouter Maybe one of you knows, which set of configuration options we need to digests here.
Probably at least the ones from `pulpcore.rqconfig`?

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -0,0 +1,205 @@
...
+        return handle_worker_heartbeat(self.name)
+
+    def notify_workers(self):
+        self.cursor.execute("NOTIFY pulp_worker_wakeup")
+
+    def cancel_abandoned_task(self, task):
+        _logger.info(f"Canceling Task {task.pk}")
+        rows = Task.objects.filter(pk=task.pk, state=TASK_STATES.RUNNING).update(
+            state=TASK_STATES.CANCELED
+        )
```
dralley, line 56: Maybe we should do this in multiple steps, with locking?  Or is this covered by one of the `advisory_lock`s already.

```python
with transaction.atomic():
    tasks = Task.objects.filter(pk=task.pk, state=TASK_STATES.RUNNING)
    tasks.update(
        state=TASK_STATES.CANCELED
    )
    _delete_incomplete_resources(task)
    # .....
```

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -0,0 +1,205 @@
...
+        self.cursor = connection.cursor()
+        self.worker = self.beat()
+
+    def beat(self):
+        return handle_worker_heartbeat(self.name)
+
+    def notify_workers(self):
+        self.cursor.execute("NOTIFY pulp_worker_wakeup")
+
+    def cancel_abandoned_task(self, task):
```
dralley, line 52: Let's leave some comments for any methods that have an depend on external locks

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def iter_tasks(self):
+        """Iterate over ready tasks and yield each task while holding the lock."""
+
+        while True:
+            taken_resources = set()
+            # When batching this query, be sure to use "pulp_created" as a cursor
+            for task in Task.objects.filter(
+                state__in=[TASK_STATES.RUNNING, TASK_STATES.WAITING]
+            ).order_by("pulp_created"):
```
dralley, line 72: Do we have an index on the date?  Otherwise this may be quite slow.

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -0,0 +1,205 @@
...
+                with suppress(AdvisoryLockError), task:
+                    # Check if someone else changed the task before we got the lock
+                    task.refresh_from_db()
+                    if task.state == TASK_STATES.RUNNING:
+                        # Running task without a lock is invalid
+                        self.cancel_abandoned_task(task)
+                        # Continue looking for the next task
+                        # without considering this tasks resources
+                        continue
```
dralley, line 81: Could you explain the semantics above?  How do we know that there is no lock?  Do we really need the refresh, or can we avoid that race entirely with locking?

pulpcore/tasking/pulpcore_worker.py
mdellweg: I'll see how it works out. I believe i factored it into a function, because the level of indentation went crazy.
Too many `try: with:`...
+1 to stating the locking explicitly.

pulpcore/tasking/pulpcore_worker.py
mdellweg: If you are referring to the refresh i am thinking, yes we need it. Another process could have done the cleanup and released the lock already before we got it. And at that point, we have outdated data.

The logic here is: We have the lock for that task, and it is in "running" state, but we are not running it, so the process that was supposed to run that task must have died (and thereby released the lock prematurely).

docs/settings.rst
mdellweg: We have the new worker entrypoint in master now.

pulpcore/tasking/worker_wrapper.py
mdellweg: That code is merged with another PR now.

pulpcore/tasking/pulpcore_worker.py
mdellweg: On a second thought, i like this function as a way to structure the higher level concept of looping over tasks from handling a single task.

pulpcore/app/migrations/0064_add_new_style_task_columns.py
```diff
@@ -0,0 +1,54 @@
+# Generated by Django 2.2.20 on 2021-04-27 07:51
+
+import django.contrib.postgres.fields
+import django.contrib.postgres.fields.jsonb
+from django.db import migrations, models
+
+
+def copy_required_resources(apps, schema_editor):
+    Task = apps.get_model('core', 'Task')
+    for task in Task.objects.all():
```
dralley, line 10: `.iterator()`

pulpcore/app/migrations/0064_add_new_style_task_columns.py
```diff
@@ -0,0 +1,54 @@
...
+
+    dependencies = [
+        ('core', '0063_repository_retained_versions'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='task',
+            name='args',
+            field=django.contrib.postgres.fields.jsonb.JSONField(null=True),
```
dralley, line 29: I presume we decided against using more specialized data types like `ArrayField` and `HStoreField`?

bmbouter: I prefer JSON for two reasons:
1) json is always save, e.g. never pickle
2) it encourages plugin writers to pass simple data types instead of potentially reusing an object it fetched at dispatch time.

dralley: I'm not sure I understand either of these, could you elaborate a bit?

pulpcore/app/serializers/task.py
```diff
@@ -107,6 +107,10 @@ class TaskSerializer(ModelSerializer):
         view_name="None",  # This is a polymorphic field. The serializer does not need a view name.
     )
     reserved_resources_record = ReservedResourcesSerializer(many=True, read_only=True)
+    required_resources = serializers.ListField(
```
bmbouter, line 110: Maybe this is right, but this part is unexpected to me. When I view task output now I get:

```
      "/pulp/api/v3/repositories/file/file/06f4a9bd-94c7-4a9d-8e5f-f82945a48dfa/",
      "/pulp/api/v3/remotes/file/file/3d033c12-c6fa-4e84-a501-db2cdb96d42c/"
    ],
    "required_resources": [
      "/pulp/api/v3/repositories/file/file/06f4a9bd-94c7-4a9d-8e5f-f82945a48dfa/",
      "/pulp/api/v3/remotes/file/file/3d033c12-c6fa-4e84-a501-db2cdb96d42c/"
    ]
```

Why introduce a new field to store the same info as `reserved_resources_record` ?

pulpcore/app/tasks/importer.py
```diff
@@ -374,8 +374,8 @@ def validate_and_assemble(toc_filename):
     CreatedResource.objects.create(content_object=the_import)
 
     task_group = TaskGroup.objects.create(description=f"Import of {path}")
-    current_task.task_group = task_group
-    current_task.save()
+    Task.objects.filter(pk=current_task.pk).update(task_group=task_group)
+    current_task.refresh_from_db()
```
bmbouter, line 378: Why do we need to call `refresh_from_db()` pattern when we call `task.update(...)` ?

pulpcore/tasking/entrypoint.py
```diff
@@ -22,17 +29,30 @@ def worker(resource_manager, pid):
...
-    sys.argv = args
-    main()
+    if settings.USE_NEW_WORKER_TYPE:
+        if resource_manager:
+            _logger.warn(
+                "Attempting to start a resource-manager with the distributed tasking system"
+            )
+            select.select([], [], [])
+        _logger.info("Starting distributed type worker")
+        from pulpcore.tasking.pulpcore_worker import NewPulpWorker
```
bmbouter, line 39: Is moving this to the imports at the top an option? I suspect it's not but wanted to ask.

pulpcore/app/migrations/0064_add_new_style_task_columns.py
bmbouter: Sure, in Celery and Pulp2 we had an issue where the tasking system would allow complex types as arguments. The concern was that pickled objects could allow for arbitrary code execution. The solution was to switch to using JSON serializers.

The second one is a similar problem, only instead of a security concern it's a stale-data concern where plugin writers could pass an instance of an object they got from the db instead of the pk for the task to then look it up. With JSON they can only pass python primatives.

I'm not sure what typing is allowed for storage in `ArrayField` and `HStoreField` but I know JSON is safe and only allows Python primitives.

pulpcore/app/serializers/task.py
dralley: I agree that we can probably avoid this.

Delete `ReservedResourcesSerializer`, replace it with a method that either gets the arraylist or joins the reserved resources depending on which worker mode Pulp is in.  And then use either a `JsonField` or a `ListField` to display the `List[str]`

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def beat(self):
+        if self.worker.last_heartbeat < timezone.now() - timedelta(seconds=self.heartbeat_period):
+            self.worker = handle_worker_heartbeat(self.name)
+
+    def notify_workers(self):
+        self.cursor.execute("NOTIFY pulp_worker_wakeup")
+
+    def cancel_abandoned_task(self, task):
+        """Cancel and clean up an abandoned task.
```
bmbouter, line 56: Can you clarify for me once more what an abandoned task is?

pulpcore/app/migrations/0064_add_new_style_task_columns.py
dralley: >Sure, in Celery and Pulp2 we had an issue where the tasking system would allow complex types as arguments. The concern was that pickled objects could allow for arbitrary code execution. The solution was to switch to using JSON serializers.

I think for Postgresql though, that isn't an issue.  To the best of my knowledge Django ArrayField and HStoreField don't do any pickling, at least, I couldn't find any in their implementations.

I think the answer is "Python primitives + any datatype that has a SQL type"

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -0,0 +1,212 @@
...
+    try:
+        _logger.info("Starting task {}".format(task.pk))
+
+        # Execute task
+        module_name, function_name = task.name.rsplit(".", 1)
+        module = importlib.import_module(module_name)
+        func = getattr(module, function_name)
+        args = json.loads(task.args) or ()
+        kwargs = json.loads(task.kwargs) or {}
+        with TemporaryDirectory(dir="."):
```
bmbouter, line 198: I verified this was `/var/lib/pulp/tmp/104579@pulp3-source-fedora34.localhost.example.com`. Nice.

pulpcore/app/migrations/0064_add_new_style_task_columns.py
bmbouter: The concern with pickling was when data being passing as args/kwargs is untrusted.

Cases like `ArrayField[Artifact]` are what I want us to prevent, that's basically the second concern. Plugin writers shouldn't pass around instances they got from the db because things change in the db, e.g. imagine passing an Artifact and then orphan cleanup happens.

pulpcore/tasking/util.py
```diff
@@ -44,6 +48,22 @@ def cancel(task_id):
...
+        rows = 0
+        with suppress(AdvisoryLockError), task:
+            if task.state == TASK_STATES.WAITING:
+                rows = Task.objects.filter(pk=task.pk, state=TASK_STATES.WAITING).update(
+                    state=TASK_STATES.CANCELED
+                )
+        # TODO There is the possibility that this task is locked, but not assigned to a worker.
+        if rows != 1:
+            with db_connection.cursor() as cursor:
+                cursor.execute("SELECT pg_notify('pulp_worker_cancel', %s)", (str(task.pk),))
```
bmbouter, line 62: To confirm my understanding, this wakes up all workers right?

bmbouter (changes requested):
Overall this is looking great. I asked a few informational questions. The one thing I think we need to change is the removal of the new serializer field.

Also I tested this locally and it worked great! It's really, really nice!

pulpcore/app/migrations/0064_add_new_style_task_columns.py
dralley: @bmbouter It doesn't pass instances though, it passes `artifact[pk]`, and instantiates it on the other end.  As far as I can tell it is the same thing as passing the PK doing it manually.

bmbouter: What is better about doing it with those other more complex types? Will it work with multi-table inheritance? The benefits of using JSON are simplicity (anyone who knows Python can understand it), and explicitness re-lookup up the data explicitly makes the tasking code very clear.

dralley: HStoreField is string-to-string only, no complex data types.

https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/fields/#django.contrib.postgres.fields.HStoreField

ArrayField only takes subclasses of `Field`, so `CharField`, `IntField`, etc.  So, not arbitrary models, unless you custom-define one.

pulpcore/app/models/task.py
```diff
@@ -346,11 +383,13 @@ def set_failed(self, exc, tb):
             exc (Exception): The exception raised by the task.
             tb (traceback): Traceback instance for the current exception.
         """
-        self.state = TASK_STATES.FAILED
-        self.finished_at = timezone.now()
         tb_str = "".join(traceback.format_tb(tb))
-        self.error = exception_to_dict(exc, tb_str)
-        self.save()
+        Task.objects.filter(pk=self.pk).update(
```
dralley, line 387: Do we not want to apply the same check from [here](https://github.com/pulp/pulpcore/pull/1261/files#diff-b6921076eabe69f8f887286bbac04f5abb8c91b5e2e747b99ea932c5abf17135R367) to avoid setting an already-completed task as failed?

pulpcore/app/tasks/importer.py
dralley: I think it's just being defensive about `current_task` having the correct state.

However it doesn't look like `current_task` is used again, so maybe that isn't necessary.

pulpcore/tasking/pulpcore_worker.py
dralley: Presumably a task that says it is still running, but actually is not.

I'm not sure if this is actually specific to "abandoned" tasks though, it seems like it would generically cancel any task.  In which case maybe it should be named as such?  (we do need to avoid name collisions with older functions however).

dralley:
I was able to get a fairly repeatable hang with this script. The second task gets stuck in "waiting" and remains that way after 5 minutes.  I'm not 100% sure but I don't *think* it blocks other tasks from completing, so that's nice.

```bash
export BASE_ADDR=http://localhost:24817


http POST $BASE_ADDR/pulp/api/v3/remotes/file/file/ name=foo$I url=bar
export REMOTE_HREF=$(http $BASE_ADDR/pulp/api/v3/remotes/file/file/ | jq -r '.results[0] | .pulp_href')

sudo systemctl stop pulpcore-worker@1 pulpcore-worker@2

http PATCH :$REMOTE_HREF name=test$RANDOM
sudo systemctl restart postgresql

sudo systemctl start pulpcore-worker@1 pulpcore-worker@2

# this task will be stuck forever
http PATCH :$REMOTE_HREF name=test$RANDOM
```

docs/components.rst
```diff
@@ -82,7 +82,38 @@ Resource Manager
...
+
+.. note::
+
+   This distributed resource-manager free tasking system is still in tech-preview.
+
+The major differences to the prior tasking system is, that tasks are not routed through a
+``resource-manager``, and not queued into ``rq`` queues. So all necessary information about tasks
+is stored in Pulp's Postgres database as a single source of truth. This version of the tasking
+system consists of a single ``pulpcore-worker`` component consequently, and can be scaled by
+increasing the number of worker processes. Each worker can handle on task at a time, and idle
```
dralley, line 102: typo: s/on/one

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -0,0 +1,212 @@
...
+                    task.refresh_from_db()
+                    if task.state == TASK_STATES.RUNNING:
+                        # A running task without a lock must be abandoned
+                        if self.cancel_abandoned_task(task):
+                            # Continue looking for the next task
+                            # without considering this tasks resources
+                            # as we just released them
+                            continue
+                    if task.state == TASK_STATES.WAITING and not any(
+                        resource in taken_resources for resource in task.required_resources or []
```
dralley, line 94: I would split the `task.required_resources or []` onto a separate line, it feels like a little bit too much going on

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -0,0 +1,212 @@
...
+    def iter_tasks(self):
+        """Iterate over ready tasks and yield each task while holding the lock."""
+
+        while True:
+            taken_resources = set()
+            # When batching this query, be sure to use "pulp_created" as a cursor
+            for task in Task.objects.filter(
+                state__in=[TASK_STATES.RUNNING, TASK_STATES.WAITING]
+            ).order_by("pulp_created"):
+                with suppress(AdvisoryLockError), task:
```
dralley, line 83: I'm not sure I like how implicit these codepaths are.

`task` is used as a context manager and potentially throws an `AdvisoryLockError`, which we deliberately suppress.  If it does throw an `AdvisoryLockError`, what happens?

* I ***think*** it acts like a `try: ...... except: pass``` and just moves on to the next iteration of the loop, but that took a fair amount of time and reading of the `contextlib` documentation.

pulpcore/tasking/util.py
```diff
@@ -44,6 +48,22 @@ def cancel(task_id):
...
+            time.sleep(0.5)
```
dralley, line 63: I'm not sure if we still need the `time.sleep(0.5)` or not, that was an RQ hack.

Also does the TODO above need to be addressed still?

pulpcore/tasking/util.py
```diff
@@ -123,3 +143,32 @@ def get_current_worker():
...
+def _hash_to_u64(value):
+    _digest = blake2s(value.encode(), digest_size=8).digest()
+    return int.from_bytes(_digest, byteorder="big", signed=True)
+
+
+def model_lock(obj):
+    return _hash_to_u64(str(obj.pk))
+
+
+class AdvisoryLock:
```
dralley, line 157: It looks like `Task` just copied and pasted this implementation, maybe we should subclass instead

dralley:
@mdellweg Excellent work!  I'm overall very impressed.  I left a few comments, most are minor, and one relatively small bug that I was able to find.

bmbouter:
This error was reported on IRC by wibbit. He says he see's it in the logs several times:

```
Internal Server Error: /pulp/api/v3/tasks/c142ca54-6490-4a07-85d5-a90f96ee1c49/
  File "/opt/bats/lib/python3.8/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/opt/bats/lib/python3.8/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/opt/bats/lib/python3.8/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/bats/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/opt/bats/lib/python3.8/site-packages/rest_framework/viewsets.py", line 125, in view
    return self.dispatch(request, *args, **kwargs)
  File "/opt/bats/lib/python3.8/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/opt/bats/lib/python3.8/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/opt/bats/lib/python3.8/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/opt/bats/lib/python3.8/site-packages/rest_framework/views.py", line 497, in dispatch
    self.initial(request, *args, **kwargs)
  File "/opt/bats/lib/python3.8/site-packages/pulpcore/app/viewsets/base.py", line 301, in initial
    super().initial(request, *args, **kwargs)
  File "/opt/bats/lib/python3.8/site-packages/rest_framework/views.py", line 415, in initial
    self.check_permissions(request)
  File "/opt/bats/lib/python3.8/site-packages/rest_framework/views.py", line 332, in check_permissions
    if not permission.has_permission(request, self):
  File "/opt/bats/lib/python3.8/site-packages/rest_access_policy/access_policy.py", line 26, in has_permission
    return self._evaluate_statements(statements, request, view, action)
  File "/opt/bats/lib/python3.8/site-packages/rest_access_policy/access_policy.py", line 62, in _evaluate_statements
    matched = self._get_statements_matching_context_conditions(
  File "/opt/bats/lib/python3.8/site-packages/rest_access_policy/access_policy.py", line 181, in _get_statements_matching_context_conditions
    passed = bool(boolExpr.parseString(condition)[0])
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1937, in parseString
    loc, tokens = self._parse(instring, 0)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1677, in _parseNoCache
    loc, tokens = self.parseImpl(instring, preloc, doActions)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 4445, in parseImpl
    return self.expr._parse(instring, loc, doActions, callPreParse=False)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1677, in _parseNoCache
    loc, tokens = self.parseImpl(instring, preloc, doActions)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 4445, in parseImpl
    return self.expr._parse(instring, loc, doActions, callPreParse=False)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1677, in _parseNoCache
    loc, tokens = self.parseImpl(instring, preloc, doActions)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 4237, in parseImpl
    ret = e._parse(instring, loc, doActions)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1677, in _parseNoCache
    loc, tokens = self.parseImpl(instring, preloc, doActions)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 4445, in parseImpl
    return self.expr._parse(instring, loc, doActions, callPreParse=False)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1677, in _parseNoCache
    loc, tokens = self.parseImpl(instring, preloc, doActions)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 4237, in parseImpl
    ret = e._parse(instring, loc, doActions)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1677, in _parseNoCache
    loc, tokens = self.parseImpl(instring, preloc, doActions)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 4445, in parseImpl
    return self.expr._parse(instring, loc, doActions, callPreParse=False)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1677, in _parseNoCache
    loc, tokens = self.parseImpl(instring, preloc, doActions)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 4237, in parseImpl
    ret = e._parse(instring, loc, doActions)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1710, in _parseNoCache
    tokens = fn(instring, tokensStart, retTokens)
  File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1316, in wrapper
    ret = func(*args[limit[0]:])
TypeError: __init__() missing 1 required positional argument: 't'
```

mdellweg:
> This error was reported on IRC by wibbit. He says he see's it in the log below several times.

Looking at the traceback, i believe the real error might be hidden inside the `raise_uncaught_exception`, while the trace we see is then in turn produced by a failure to parse an access policy. We may need to fix the latter to see the former.

pulpcore/tasking/util.py
mdellweg: This is not waking up any workers, but notifying all workers, that are currently listening to the "pulp_worker_cancel" channel. All workers whose task.pk does not match will ignore the message.
Workers is sleep state (without a task) will not listen here.

pulpcore/tasking/entrypoint.py
mdellweg: I put it here to match with the rather voluminous imports of "all the rq libraries".
As this is really a diverting wrapper to call into other code depending on a condition, i thought it was wise to import this other code late/lazy.

pulpcore/app/migrations/0064_add_new_style_task_columns.py
mdellweg: Do you mean this?
```suggestion
    for task in Task.objects.iterator():
```

pulpcore/app/serializers/task.py
mdellweg: The old name did not quite match the purpose for the new use. And yes, i didn't want to change the old code too much.
At this point, i would be confident to completely remove the `ReservedResourceRecord` model in favour of the new HSTORE field.
Is the complaint that the data is shown twice in the task output? Should we stick to the old name only? I guess, out side the internal tasking code it is still a "record of used resources".

pulpcore/app/tasks/importer.py
mdellweg: Maybe i should have used `save` with `update_fields`.
I only learned about it later. But it is crucial to not save all of the task, because it's state is managed elsewhere.
Thinking about it, i remember that the original problem was that the call to save its state as final removed the task_group again. So this change was in line to be consistent about only changing relevant rows.

pulpcore/tasking/pulpcore_worker.py
mdellweg: An abandoned task is is `state=running`, but no worker is holding it's lock.
So the line of reasoning here is:
I am a worker. While looping tasks, i got the lock for this one. It claims to be running, which cannot be true (another worker would hold the lock). Therefore i am on the hook to cancel it.

mdellweg:
> **edit:** I might have accidentally been using the old tasking system? `pulpcore-worker` was printing RQ logs despite my `settings.py` being correct. Not sure if I'm doing something improperly.

@dralley Are you sure, your systemd scripts use the new entry points?

pulpcore/app/migrations/0064_add_new_style_task_columns.py
dralley: Yes

pulpcore/tasking/util.py
mdellweg: Actually, this class isn't used anywhere. I just didn't want it to go to waste, when i folded it into the task.
I'll remove it.

pulpcore/tasking/util.py
mdellweg: The time sleep is there to give the notified worker some time to cancel the task.
The TODO means, there is a slight chance, that a cancel request might not cancel the task because a worker is evaluating this task for pickup atm. And i do not how likely that is to happen, nor how to solve it without checking the cancelled state later. We could state that cancel (in the rest api) is fire and forget, and the user needs to check whether it succeeded.
Another solution might involve a new task state `TO_BE_CANCELLED` that can be set from `RUNNING` or `WAITING` without having the lock.

pulpcore/tasking/pulpcore_worker.py
mdellweg: That is exactly it. I was looking for a combined version of `if` and `with` but no luck.
The exception ensures, that the block inside will only run _with_ the lock in place. But we don't have rescue code. And i think suppress is the pythonic way for this case.
I will add a comment.

pulpcore/app/models/task.py
mdellweg: I'll add it. However, only the worker that is working on that task will ever want to set it to failed.

pulpcore/tasking/pulpcore_worker.py
dralley: An inline comment would be good here as well

dralley: Does this mean that a worker can only clean up its own tasks?

dralley:
>@dralley Are you sure, your systemd scripts use the new entry points?

I think the problem is that the documentation doesn't make it clear that the installer needs to be re-run / the service files edited.  It just says

> To switch to using this worker model, one needs to set ``USE_NEW_WORKER_STYLE=True`` in pulp
settings, and start the worker processes via ``pulpcore-worker``.

> ...

> If ``True``, the ``pulpcore-worker`` command will start workers of the new type.  If ``False``
    it will chainload into the traditional ``rq`` based system.  Also the ``pulpcore-api``
    processes will dispatch their tasks accordingly.  When changing this value, all ``pulpcore``
    (at least api and worker) processes must be restarted.

pulpcore/tasking/util.py
bmbouter: I think it's ok if it's fire-and-foreget if there is guaranteed eventual consistency. So my question is: design wise, how can we ensure 100% of cancel requests eventually cancel? Is that in place with this design or is there more/alternate work needed?

bmbouter:
[idea] What do you think about making a "performance" test that fires say 500 tasks and ensures they all run to completion without error? I suggest this because it seems like a) it would be easy and b) it would be helpful in sanity checking the correctness.

I say performance because I think we do not want this to run as part of the normal "functional" tests which run with every PR.

@mdellweg and @dralley wdyt?

pulpcore/tasking/util.py
bmbouter: Great, ty

pulpcore/app/tasks/importer.py
bmbouter: Personally I think using `update_fields` would be much better. This pattern does concern me slightly. Not a whole lot, but some.

pulpcore/app/serializers/task.py
bmbouter: I think we should stick to the old name, but have it use the new implementation. I don't see a motivation to change this part of the API, even with an additive change. The `reserved_resources_record` is still pretty easy.

dralley:
I agree, although I would actually try to have more of a "tasking torture test suite" than just a set of performance tests.  We could of course do both.

pulpcore/tasking/util.py
mdellweg: No it isnt. ATM it tries to cancel the task right away (holding its lock). If impossible, it sends a signal on the cancel channel. But that will only be heard if some worker is listening for it, or put another way, handling that task.
But the lock could have been held by a worker just to realize the task is not ready yet. And if you try to cancel in that exact timeframe, nothing will happen.
I need to brainstorm a little more what options we do have.

bmbouter:
> I agree, although I would actually try to have more of a "tasking torture test suite" than just a set of performance tests. We could of course do both.

+1 to this. this is roughly what I was trying to express also.

dralley:
BTW my idea of what this would be would include things like "restart postgresql mid-task" as the script I posted earlier does.

wibbit:
> This error was reported on IRC by wibbit. He says he see's it in the log below several times.
> 
> The only pulp code here is `File "/opt/bats/lib/python3.8/site-packages/pulpcore/app/viewsets/base.py", line 301, in initial` and I that likely is in the WSGI layer, but the ISE gives a pulp task endpoint, so that's confusing. Also wibbit reported that for this task when he later requested it, it gave him a 404...
> 
> Perhaps it's failing in the task itself, but very early??
> 
> ```
> Internal Server Error: /pulp/api/v3/tasks/c142ca54-6490-4a07-85d5-a90f96ee1c49/
> Traceback (most recent call last):
>   File "/opt/bats/lib/python3.8/site-packages/django/core/handlers/exception.py", line 34, in inner
>     response = get_response(request)
>   File "/opt/bats/lib/python3.8/site-packages/django/core/handlers/base.py", line 115, in _get_response
>     response = self.process_exception_by_middleware(e, request)
>   File "/opt/bats/lib/python3.8/site-packages/django/core/handlers/base.py", line 113, in _get_response
>     response = wrapped_callback(request, *callback_args, **callback_kwargs)
>   File "/opt/bats/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
>     return view_func(*args, **kwargs)
>   File "/opt/bats/lib/python3.8/site-packages/rest_framework/viewsets.py", line 125, in view
>     return self.dispatch(request, *args, **kwargs)
>   File "/opt/bats/lib/python3.8/site-packages/rest_framework/views.py", line 509, in dispatch
>     response = self.handle_exception(exc)
>   File "/opt/bats/lib/python3.8/site-packages/rest_framework/views.py", line 469, in handle_exception
>     self.raise_uncaught_exception(exc)
>   File "/opt/bats/lib/python3.8/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
>     raise exc
>   File "/opt/bats/lib/python3.8/site-packages/rest_framework/views.py", line 497, in dispatch
>     self.initial(request, *args, **kwargs)
>   File "/opt/bats/lib/python3.8/site-packages/pulpcore/app/viewsets/base.py", line 301, in initial
>     super().initial(request, *args, **kwargs)
>   File "/opt/bats/lib/python3.8/site-packages/rest_framework/views.py", line 415, in initial
>     self.check_permissions(request)
>   File "/opt/bats/lib/python3.8/site-packages/rest_framework/views.py", line 332, in check_permissions
>     if not permission.has_permission(request, self):
>   File "/opt/bats/lib/python3.8/site-packages/rest_access_policy/access_policy.py", line 26, in has_permission
>     return self._evaluate_statements(statements, request, view, action)
>   File "/opt/bats/lib/python3.8/site-packages/rest_access_policy/access_policy.py", line 62, in _evaluate_statements
>     matched = self._get_statements_matching_context_conditions(
>   File "/opt/bats/lib/python3.8/site-packages/rest_access_policy/access_policy.py", line 181, in _get_statements_matching_context_conditions
>     passed = bool(boolExpr.parseString(condition)[0])
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1937, in parseString
>     loc, tokens = self._parse(instring, 0)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1677, in _parseNoCache
>     loc, tokens = self.parseImpl(instring, preloc, doActions)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 4445, in parseImpl
>     return self.expr._parse(instring, loc, doActions, callPreParse=False)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1677, in _parseNoCache
>     loc, tokens = self.parseImpl(instring, preloc, doActions)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 4445, in parseImpl
>     return self.expr._parse(instring, loc, doActions, callPreParse=False)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1677, in _parseNoCache
>     loc, tokens = self.parseImpl(instring, preloc, doActions)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 4237, in parseImpl
>     ret = e._parse(instring, loc, doActions)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1677, in _parseNoCache
>     loc, tokens = self.parseImpl(instring, preloc, doActions)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 4445, in parseImpl
>     return self.expr._parse(instring, loc, doActions, callPreParse=False)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1677, in _parseNoCache
>     loc, tokens = self.parseImpl(instring, preloc, doActions)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 4237, in parseImpl
>     ret = e._parse(instring, loc, doActions)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1677, in _parseNoCache
>     loc, tokens = self.parseImpl(instring, preloc, doActions)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 4445, in parseImpl
>     return self.expr._parse(instring, loc, doActions, callPreParse=False)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1677, in _parseNoCache
>     loc, tokens = self.parseImpl(instring, preloc, doActions)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 4237, in parseImpl
>     ret = e._parse(instring, loc, doActions)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1710, in _parseNoCache
>     tokens = fn(instring, tokensStart, retTokens)
>   File "/opt/bats/lib/python3.8/site-packages/pyparsing.py", line 1316, in wrapper
>     ret = func(*args[limit[0]:])
> TypeError: __init__() missing 1 required positional argument: 't'
> ```

I think this can be traced back to my trying the use of Threads for the gunicorn process as opposed to Processes.

I can 100% cause the error when I have threads configured, however, when threads are not configured, and I use processes, (3 * 15 threads v's 45) I am not able to replicate the issue.

So I think this at least can be ignored (unless of course you expect to be able to use threads :)

pulpcore/tasking/pulpcore_worker.py
mdellweg: Yes and no. When a worker is shut down, it closes its postgres connection (or eventually it times out). So the advisory lock will vanish. That is what an abandoned task is all about.
The next worker looking for tasks will see this task, get it's lock (only now we can consider it to own that task), and clean it up. That way, we can ensure that only one worker will try to clean up that task.
If that worker is killed while cleaning up, the next worker coming at this task will continue.

mdellweg: Having said that, the next worker will see the task being cancelled in the current implementation, so it won't be cleaned up.
Maybe we should first mark the task as cancelling and only when nicely cleaned up set it to cancelled.
That mechanism may solve the other problem identified in the TODO of the cancel call too.

mdellweg:
> > @dralley Are you sure, your systemd scripts use the new entry points?
> 
> I think the problem is that the documentation doesn't make it clear that the installer needs to be re-run / the service files edited. It just says
> 
> > To switch to using this worker model, one needs to set `USE_NEW_WORKER_STYLE=True` in pulp
> > settings, and start the worker processes via `pulpcore-worker`.
> 
> > ...
> 
> > If `True`, the `pulpcore-worker` command will start workers of the new type.  If `False`
> > it will chainload into the traditional `rq` based system.  Also the `pulpcore-api`
> > processes will dispatch their tasks accordingly.  When changing this value, all `pulpcore`
> > (at least api and worker) processes must be restarted.
> 
> So I didn't touch my services apart from changing settings.py and restarting them

I'm not sure this is an issue here. If we expect the change to come in the next version of pulpcore, we can expect the installer to be rerun. (actually that change will already be rolled out with the current version of pulpcore, and it's installer). On the other hand, for manual installations, it is covered by the docs already. What we do not have at all is instructions to update a manual installation. The best hint there is the entry in the changelog.

mdellweg:
To sum up the current state as i see it, we have two discussion points left:
1. Cancelling tasks is (at least theoretically) slightly probablistic.
2. We don not have good tests to torture the (new or old) tasking system.

While i'm happy to sort out the former in this PR (with the help of some good discussion), i'd love to state that the latter is probably out of scope for this PR, because we might want its not limited to this new feature.

bmbouter:
> To sum up the current state as i see it, we have two discussion points left:
> 
>     1. Cancelling tasks is (at least theoretically) slightly probablistic.
> 
>     2. We don not have good tests to torture the (new or old) tasking system.
> 
> 
> While i'm happy to sort out the former in this PR (with the help of some good discussion), i'd love to state that the latter is probably out of scope for this PR, because we might want its not limited to this new feature.
> WDYT: @dralley @bmbouter

I agree with this approach. Have this PR handle cancellation correctly (let us know what we can do to help that along), and have another PR add a test that tortures the tasking system while restarting workers.

dralley:
>What we do not have at all is instructions to update a manual installation. The best hint there is the entry in the changelog.

I think we need this, because the default is to use the old tasking system.  Users need to know that if they want to upgrade and then try the new tasking system briefly, it is not as simple as just doing what is described in the docs.  They need to re-run the installer any time they switch back and forth.

mdellweg:
> > What we do not have at all is instructions to update a manual installation. The best hint there is the entry in the changelog.
> 
> I think we need this, because the default is to use the old tasking system. Users need to know that if they want to upgrade and then try the new tasking system briefly, it is not as simple as just doing what is described in the docs. They need to re-run the installer any time they switch back and forth.

They don't need to. Starting from 3.13 all workers should be started with the new entrypoint. That's the very reason why we introduced it. It can merely be seen as a legacy convenience, that calling rq still works.

dralley:
Ok, I believe I understand now.  The problem is just that my development environment in particular was out of date and needed to be recreated.

dralley:
I agree that the test suite is out of scope for this PR

pulpcore/app/serializers/task.py
```diff
@@ -106,7 +97,10 @@ class TaskSerializer(ModelSerializer):
         read_only=True,
         view_name="None",  # This is a polymorphic field. The serializer does not need a view name.
     )
-    reserved_resources_record = ReservedResourcesSerializer(many=True, read_only=True)
+    reserved_resources_record = serializers.ListField(
```
dralley, line 100: `ListField` should have a `child` defined ideally https://www.django-rest-framework.org/api-guide/fields/#listfield

mdellweg: I added it. And then realized, that we use this serializer read only. But it's better this way.

mdellweg:
> I have a naive thought about cancellation. Perhaps this is maybe already how it works, but here's the idea. What if:
> 
>     1. Setting the task state to cancelled occurred atomically in the webserver's cancel view
> 
>     2. The worker was notified to wakeup and if it's processing that task ID just stop doing so

That's almost what i was thinking. However i thought, the api-worker can set the task to cancelling regardless of any locks (but only from waiting or running), and then send the notify in case any worker is performing that task. And the next worker to get the lock for the task will first attempt to clean up the unfinished created resources, and on success set the state to the final cancelled.
When a worker is trying to start working on the task, it will be unable to set it from cancelling to running, and otoh, if it will transition it from waiting to running, it will already listen for the cancel channel. And if a worker is killed mid-cleanup, the next one will finish the process.

dralley:
Given wibbit's seal of approval after testing at scale, the number of the issues with the old system, and the recommendations we're already willing to provide to Katello, perhaps we should consider providing the new tasking system as the default with the ability to easily fall back to the old system?

dralley (approved):

bmbouter (approved):
I took another look through this, and I think it's good to go. @mdellweg please squash the commits before merging.

mdellweg:
@bmbouter The two commits i intend to quash are labeled with "Adds new worker type" and "Introduce new task state canceling".
And if you insist, i could also squash "Delete unused code `mark_worker_online`" and "Activate new worker style in subset of tests", but i want to keep "Add orphan cleanup stop gap in pulpcore-worker" separate, as i think it can eventually be reverted cleanly.

PR #1263, opened 2021-04-19
Fix pulp export creation task dispatching

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/app/tasks/export.py
+++ b/pulpcore/app/tasks/export.py
@@ -16,10 +16,13 @@ from pulpcore.app.models import (
     ExportedResource,
     Exporter,
     Publication,
+    PulpExport,
+    PulpExporter,
     RepositoryVersion,
     Task,
 )
 from pulpcore.app.models.content import ContentArtifact
+from pulpcore.app.serializers import PulpExportSerializer
 from pulpcore.app.util import get_version_from_model
 from pulpcore.app.importexport import (
     export_versions,
@@ -147,7 +150,7 @@ def _incremental_requested(the_export):
     return (starting_versions_provided or last_exists) and not full
 
 
-def pulp_export(the_export):
+def pulp_export(exporter_pk, params):
     """
     Create a PulpExport to export pulp_exporter.repositories.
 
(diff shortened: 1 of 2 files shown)


pulpcore/app/tasks/export.py
```diff
@@ -156,14 +159,23 @@ def pulp_export(the_export):
     3) Compute and store the sha256 and filename of the resulting tar.gz/chunks
 
     Args:
-        the_export (models.PulpExport): PulpExport instance
+        exporter_pk (str): PulpExporter
+        params (dict): request data
+        validated_data (dict): serializer output
```
ggainey, line 164: I don't see validated_data in the args list

pulpcore/app/tasks/export.py
```diff
@@ -156,14 +159,23 @@ def pulp_export(the_export):
...
 
     Raises:
         ValidationError: When path is not in the ALLOWED_EXPORT_PATHS setting,
             OR path exists and is not a directory
     """
+    pulp_exporter = PulpExporter.objects.get(pk=exporter_pk)
+    serializer = PulpExportSerializer(data=params, context={"exporter": pulp_exporter})
+    serializer.is_valid(raise_exception=True)
```
ggainey, line 172: With this change, we validate the data twice - once at the viewset, so we can return errors quickly, and then again in the task, so we can use the resulting validated data. I don't think this is a huge issue - but it does double the db-accesses for this path. Not sure there's anything to be done about it, given the constraints that drive this change though.

ggainey (changes requested):
A minor nit and an observation, but generally looks good to me! Thanks for doing the work!

pulpcore/app/tasks/export.py
mdellweg: right (it was there for a brief moment...)

pulpcore/app/tasks/export.py
mdellweg: Yes. But my view is, that both validations can be separated by a significant amount of time. And the circumstances might even make once valid data invalid at the point the task is ready to start. So we really should validate it there.

ggainey: That's a good observation, concur.

ggainey (approved):

ipanova (approved):

daviddavis (approved):

PR #1269, opened 2021-04-21
Update signing docs

mdellweg:
Please look at the two commits separately.

In the second one i added a service command to produce a AsciiArmoredSigningService, but i don't know it it is ready for consumption, as it is not yet able to create e.g. a pulp_deb provided one.

### DIFF
--- a/CHANGES/8609.doc
+++ b/CHANGES/8609.doc
@@ -1,0 +1,1 @@
+Updated signing service workflow. Removed old deprecation warning.
--- a/docs/workflows/signed-metadata.rst
+++ b/docs/workflows/signed-metadata.rst
@@ -12,15 +12,25 @@ Administrators can add signing services to Pulp using the command line tools. Us
 may then associate the signing services with repositories that support content signing.
 The example below demonstrates how a signing service can be created using ``gpg``:
 
-1. Create a signing script that accepts a file name as the only argument. The script
+1. Make sure the service user ``pulp`` has access to ``gpg`` and that the keypair is
+   installed in its keyrings. The private key might alternatively be provided by a
+   hardware cryptographic device.
+
+2. Create a signing script that accepts a file name as the only argument. The script
    needs to generate an ascii-armored detached GPG signature for that file. The script
    should then print out a JSON structure with the following format. All the file names
    are relative paths inside the current working directory::
 
-       {"file": "filename", "signature": "filename.asc", "key": "public.key"}
+       {"file": "filename", "signature": "filename.asc"}
 
-   The filename must remain the same for the detached signature, as shown. Below is an
-   example of a signing script:
+   The filename must remain the same for the detached signature, as shown.
+
+   .. note::
+
+      Plugins may provide other signing service classes that may need their JSON output to
+      contain different information.
+
+   Below is an example of a signing script:
 
    .. code-block:: bash
 
@@ -29,13 +39,8 @@ The example below demonstrates how a signing service can be created using ``gpg`
        FILE_PATH=$1
        SIGNATURE_PATH="$1.asc"
 
-       PUBLIC_KEY_PATH="$(cd "$(dirname $1)" && pwd)/public.key"
-
        ADMIN_ID="658285BA1A648083"
        PASSWORD="password"
-
-       # Export a public key
-       gpg --armor --export admin@example.com > $PUBLIC_KEY_PATH
 
        # Create a detached signature
        gpg --quiet --batch --pinentry-mode loopback --yes --passphrase \
@@ -45,24 +50,17 @@ The example below demonstrates how a signing service can be created using ``gpg`
        # Check the exit status
        STATUS=$?
        if [ $STATUS -eq 0 ]; then
-          echo {\"file\": \"$FILE_PATH\", \"signature\": \"$SIGNATURE_PATH\", \
-              \"key\": \"$PUBLIC_KEY_PATH\"}
+          echo {\"file\": \"$FILE_PATH\", \"signature\": \"$SIGNATURE_PATH\"}
        else
           exit $STATUS
        fi
-
-   .. deprecated:: 3.10
-
-        The path to the public key is no longer required to be present in the output of a signing
-        script. Pass the value of the public key during the object creation process instead. The
-        step number 2 below depicts the recommended way of creating a new signing service.
 
    .. note::
 
        Make sure the script contains a proper shebang and Pulp has got valid permissions
        to execute it.
 
-2. Create a signing service consisting of an absolute path to the script and a meaningful
+3. Create a signing service consisting of an absolute path to the script and a meaningful
    name describing the script's purpose. It is possible to insert the signing service in
    to a database by using the ``pulpcore-manager shell_plus`` interactive Python shell. Here is an
    example showing how to create one instance pointing to a script:
@@ -86,7 +84,7 @@ The example below demonstrates how a signing service can be created using ``gpg`
        runs additional checks in order to prevent saving invalid scripts to the database.
        This feature enables administrators to validate their signing scripts in advance.
 
-3. Retrieve and check the saved signing service via REST API::
+4. Retrieve and check the saved signing service via REST API::
 
        $ http :24817/pulp/api/v3/signing-services/
 
--- a/docs/workflows/signed-metadata.rst
+++ b/docs/workflows/signed-metadata.rst
@@ -84,6 +84,11 @@ The example below demonstrates how a signing service can be created using ``gpg`
        runs additional checks in order to prevent saving invalid scripts to the database.
        This feature enables administrators to validate their signing scripts in advance.
 
+   .. note::
+
+      You can use `pulpcore-manager add-signing-service` to add a ``SigningService``.
+      This command is however still in tech-preview.
+
 4. Retrieve and check the saved signing service via REST API::
 
        $ http :24817/pulp/api/v3/signing-services/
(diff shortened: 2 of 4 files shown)


pulpcore/app/management/commands/add-signing-service.py
```diff
@@ -0,0 +1,52 @@
...
+from django.core.management import BaseCommand, CommandError
+
+from pulpcore.app.models import AsciiArmoredDetachedSigningService
+
+
+class Command(BaseCommand):
+    """
+    Django management command for adding a signing service.
+
+    TODO extend this command to create any type of signing service.
```
mdellweg, line 15: How are we supposed to specify them?

docs/workflows/signed-metadata.rst
```diff
@@ -12,15 +12,25 @@ Administrators can add signing services to Pulp using the command line tools. Us
 may then associate the signing services with repositories that support content signing.
 The example below demonstrates how a signing service can be created using ``gpg``:
 
+0. Make sure, the service user ``pulp`` has access to ``gpg`` and that the keypair is
```
ipanova, line 15: if it is not a big of a trouble, can we start numbering of the steps/instructions with `1`?

docs/workflows/signed-metadata.rst
```diff
@@ -84,6 +84,10 @@ The example below demonstrates how a signing service can be created using ``gpg`
        runs additional checks in order to prevent saving invalid scripts to the database.
        This feature enables administrators to validate their signing scripts in advance.
 
+   .. note::
```
ipanova, line 87: is it worth mentioning that this command is in tech-preview since it's a new one?

docs/workflows/signed-metadata.rst
mdellweg: Sure. I am thinking about turning this into a precondition.

docs/workflows/signed-metadata.rst
mdellweg: Good call.

CHANGES/8609.doc
```diff
@@ -0,0 +1 @@
+Updated signing service workflow. Removed old deprecation warning.
```
ggainey, line 1: Need to call out the new add-signing-service in towncrier. That kind of mixes a feature into this .doc - do we switch to .feature? I'm not sure what the right thing to do is here. At a minimum, need to call out the new function.

docs/workflows/signed-metadata.rst
```diff
@@ -12,15 +12,25 @@ Administrators can add signing services to Pulp using the command line tools. Us
 may then associate the signing services with repositories that support content signing.
 The example below demonstrates how a signing service can be created using ``gpg``:
 
-1. Create a signing script that accepts a file name as the only argument. The script
+1. Make sure, the service user ``pulp`` has access to ``gpg`` and that the keypair is
```
ggainey, line 15: Lose the comma.

docs/workflows/signed-metadata.rst
```diff
@@ -12,15 +12,25 @@ Administrators can add signing services to Pulp using the command line tools. Us
...
+   installed in its keyrings. The private key might alternatively be provided by a
+   hardware cryptografic device.
```
ggainey, line 17: cryptographic

ggainey (changes requested):
A few minor nits. The only thing I really need to see, is calling out add-signing-service in towncrier. Looks good otherwise!

docs/workflows/signed-metadata.rst
mdellweg: :man_facepalming:

ggainey (approved):
shippit :)

CHANGES/8609.doc
bmbouter: This seems right to me. If there is a doc part to the work, it gets `.doc` in addition to the other `.feature` part.

bmbouter (approved):
This looks right to me. Thanks!

pulpcore/app/management/commands/add-signing-service.py
```diff
@@ -0,0 +1,64 @@
...
+        key_id = options["key-id"]
+        app_label, service_class = options["class"].split(":")
+
+        SigningService = apps.get_model(app_label, service_class)
+        gpg = gnupg.GPG()
+        key_list = gpg.list_keys(keys=[key_id])
+        if not len(key_list) == 1:
+            raise CommandError(_("There are {} keys matching the key id.").format(len(key_list)))
+        fingerprint = key_list[0]["fingerprint"]
+        print(fingerprint)
```
ipanova, line 52: did you mean to print this? if yes maybe make the message a bit more user friendly?

pulpcore/app/management/commands/add-signing-service.py
```diff
@@ -0,0 +1,64 @@
...
+        parser.add_argument(
+            "name",
+            help=_("Name, the signing_service should get in the database."),
+        )
+        parser.add_argument(
+            "script",
+            help=_("Shell script where the signing service is located."),
+        )
+        parser.add_argument(
+            "key-id",
```
ipanova, line 30: so I always get conflicted with dash or underscores, I see in other management command uderscore used instead. Should we switch to it as well, for consistency? https://github.com/pulp/pulpcore/blob/master/pulpcore/app/management/commands/stage-profile-summary.py#L17

pulpcore/app/management/commands/add-signing-service.py
```diff
@@ -0,0 +1,64 @@
...
+        key_list = gpg.list_keys(keys=[key_id])
+        if not len(key_list) == 1:
+            raise CommandError(_("There are {} keys matching the key id.").format(len(key_list)))
+        fingerprint = key_list[0]["fingerprint"]
+        print(fingerprint)
+        public_key = gpg.export_keys(key_id)
+
+        script_path = Path(script).resolve(strict=True)
+
+        SigningService.objects.create(
```
ipanova, line 57: can use here get_or_create? If this command will be run twice in a row, create() will fail.

pulpcore/app/management/commands/add-signing-service.py
```diff
@@ -0,0 +1,64 @@
...
+        SigningService = apps.get_model(app_label, service_class)
+        gpg = gnupg.GPG()
+        key_list = gpg.list_keys(keys=[key_id])
+        if not len(key_list) == 1:
+            raise CommandError(_("There are {} keys matching the key id.").format(len(key_list)))
+        fingerprint = key_list[0]["fingerprint"]
+        print(fingerprint)
+        public_key = gpg.export_keys(key_id)
+
+        script_path = Path(script).resolve(strict=True)
```
ipanova, line 55: how about catching FileNotFound error?

pulpcore/app/management/commands/add-signing-service.py
```diff
@@ -0,0 +1,64 @@
...
+            default="core:AsciiArmoredDetachedSigningService",
+            required=False,
+            help=_("Signing service class prefixed by the app label separated by a colon."),
+        )
+
+    def handle(self, *args, **options):
+        name = options["name"]
+        script = options["script"]
+        key_id = options["key-id"]
+        app_label, service_class = options["class"].split(":")
```
ipanova, line 44: maybe too defensive, but maybe worth adding validation here, that signign service was provided in the expected format separated by `:`, also note that this  option is not required

pulpcore/app/management/commands/add-signing-service.py
```diff
@@ -0,0 +1,64 @@
...
+
+        SigningService = apps.get_model(app_label, service_class)
```
ipanova, line 46: i would catch a lookup error here :)

pulpcore/app/management/commands/add-signing-service.py
bmbouter: +1 for underscores. In looking at `pulpcore-manager` output underscores seem most common. In another PR maybe we should switch the stage-profile-summary which is still tech preview also.

pulpcore/app/management/commands/add-signing-service.py
mdellweg: Only for debugging purposes.

pulpcore/app/management/commands/add-signing-service.py
mdellweg: You know what? I'll scrap the id part altogether.

pulpcore/app/management/commands/add-signing-service.py
mdellweg: I'm conflicted here, because signing services have unique names. And if the service does exist, it is immutable, so with a get_or_create i will not necessarily get what i ask for.
However, i can improve the error reporting here.

pulpcore/app/management/commands/add-signing-service.py
mdellweg: It has a default value, so it will always be present.
I'll add some validation to tell the user what went wrong.

ipanova: you're right, i have not noticed the `default`

pulpcore/app/management/commands/add-signing-service.py
mdellweg: It is the key you reference, and it can be the fingerprint, last 8 or 16 digits of it, or part of the user id on the key (if unambiguous).

ipanova (approved):
Thank you for addressing all the comments!

PR #1273, opened 2021-04-24
Merge ContentUnassociation stage into the ContentAssociation stage

dralley:
ContentUnassociation doesn't really need to be a separate stage, and
making it one breaks the expectations of the "pipeline" by changing the
semantics of what goes into and out of the queues.

[noissue]


pulpcore/plugin/stages/association_stages.py
```diff
@@ -1,83 +0,0 @@
...
-                        to_delete.remove(d_content.content.pk)
-                    except KeyError:
-                        to_add.add(d_content.content.pk)
-
-                if to_add:
-                    self.new_version.add_content(Content.objects.filter(pk__in=to_add))
-                    pb.increase_by(len(to_add))
-
-            if to_delete:
-                await self.put(Content.objects.filter(pk__in=to_delete))
```
dralley, line 50: We break the expectations of the pipeline here by passing querysets instead of DeclarativeContent, and the semantics are different.  Now they're content to delete rather than things to include.

pulpcore/plugin/stages/association_stages.py
```diff
@@ -1,83 +0,0 @@
...
-        """
-        The coroutine for this stage.
-
-        Returns:
-            The coroutine for this stage.
-        """
-        with ProgressReport(message="Un-Associating Content", code="unassociating.content") as pb:
-            async for queryset_to_unassociate in self.items():
-                self.new_version.remove_content(queryset_to_unassociate)
-                pb.increase_by(queryset_to_unassociate.count())
```
dralley, line 81: And why?  It's not worth it for these two lines of code.

pulpcore/plugin/stages/content_stages.py
```diff
@@ -187,3 +187,55 @@ async def run(self):
...
+                if to_add:
+                    self.new_version.add_content(Content.objects.filter(pk__in=to_add))
+                    pb.increase_by(len(to_add))
+
+            if self.allow_delete and to_delete:
+                with ProgressReport(
+                    message="Un-Associating Content", code="unassociating.content"
+                ) as pb:
+                    self.new_version.remove_content(Content.objects.filter(pk__in=to_delete))
+                    pb.increase_by(len(to_delete))
```
dralley, line 241: Now we just do everything at once.

pulpcore/plugin/stages/declarative_version.py
```diff
@@ -141,9 +146,6 @@ def create(self):
             with self.repository.new_version() as new_version:
                 loop = asyncio.get_event_loop()
                 stages = self.pipeline_stages(new_version)
-                stages.append(ContentAssociation(new_version))
-                if self.mirror:
-                    stages.append(ContentUnassociation(new_version))
```
dralley, line 146: There's no reason we can't just pass a param into the stage rather than toggling an entire stage based on the param.

pulpcore/plugin/stages/declarative_version.py
```diff
@@ -128,6 +132,7 @@ def pipeline_stages(self, new_version):
             ContentSaver(),
             RemoteArtifactSaver(),
             ResolveContentFutures(),
+            ContentAssociation(new_version, self.mirror),
```
dralley, line 135: FWIW, potentially we need to continue appending this for the sake of the plugins that define their own pipeline.

mdellweg: We can deprecate it, while we serve the new stage with a new name.

pulpcore/plugin/stages/content_stages.py
```diff
@@ -187,3 +187,55 @@ async def run(self):
...
+        new_version (:class:`~pulpcore.plugin.models.RepositoryVersion`): The repo version this
+            stage associates content with.
+        args: unused positional arguments passed along to :class:`~pulpcore.plugin.stages.Stage`.
+        kwargs: unused keyword arguments passed along to :class:`~pulpcore.plugin.stages.Stage`.
+    """
+
+    def __init__(self, new_version, allow_delete, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.new_version = new_version
+        self.allow_delete = allow_delete
```
mdellweg, line 213: What about keeping the name `mirror` here for consistency?

pulpcore/plugin/stages/content_stages.py
```diff
@@ -187,3 +187,55 @@ async def run(self):
...
+                        to_add.add(d_content.content.pk)
+
+                if to_add:
+                    self.new_version.add_content(Content.objects.filter(pk__in=to_add))
+                    pb.increase_by(len(to_add))
+
+            if self.allow_delete and to_delete:
+                with ProgressReport(
+                    message="Un-Associating Content", code="unassociating.content"
+                ) as pb:
```
mdellweg, line 239: I think, i would expect the progress report to exist if in mirror mode even, if there is nothing to delete.

mdellweg (commented):
I really like the improvement.
However, i think we should introduce the combined stage with a new name and keep the old ones around unmodified (well, with a deprecation warning) for a couple of releases.

pulpcore/plugin/stages/declarative_version.py
dralley: It looks like, due to being outside the pipeline before, none of the plugins actually use it directly.

So I just need to move it back to where it as and we'll be fine to move forwards.

mdellweg: ... And all plugins will automatically start using it.
:+1:

pulpcore/plugin/stages/content_stages.py
dralley: This is consistent with what it was doing before - the ContentUnassociation stage never ran at all unless mirror mode was active.

pulpcore/plugin/stages/declarative_version.py
mdellweg: Would it be wise to also incorporate the EndStage then?

dralley: In the current state, absolutely.  We're not even passing anything out the end of this stage, and there's no way to modify the pipeline after it.

pulpcore/plugin/stages/content_stages.py
mdellweg: What i mean would be something like:
```suggestion
            if self.allow_delete:
                with ProgressReport(
                    message="Un-Associating Content", code="unassociating.content"
                ) as pb:
                    if to_delete:
```

pulpcore/plugin/stages/declarative_version.py
dralley: The migration plugin does use it : /.  It's kind of a special case though.

dralley: So we can get rid of it from the normal pipeline but we can't delete it

pulpcore/plugin/stages/content_stages.py
dralley: I see what you mean.  Done.

docs/plugins/api-reference/stages.rst
```diff
@@ -67,13 +66,4 @@ Content Related Stages
 
 .. autoclass:: pulpcore.plugin.stages.ResolveContentFutures
 
-
-.. _content-association-stages:
-
-Content Association and Unassociation Stages
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
 .. autoclass:: pulpcore.plugin.stages.ContentAssociation
```
mdellweg, line 69: Do we write it out in the docs, that this stage may not be used explicitly?

dralley: I don't believe we do.

pulpcore/plugin/stages/declarative_version.py
dralley: We can't do that either FYI.  Or maybe we can but "just" deleting it doesn't work and I don't feel like debugging why if we have to keep the EndStage around anyways.

dralley:
@goosemania Added an issue.

mdellweg (approved):

PR #1274, opened 2021-04-25
Fix artifact_stages to not uniquify based on an empty checksum-type.

ggainey:
fixes #8625
[nocoverage]

### DIFF
--- a/CHANGES/8625.bugfix
+++ b/CHANGES/8625.bugfix
@@ -1,0 +1,1 @@
+Fixed on-demand sync/migration of repositories that don't have sha256 checksums.
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -329,7 +329,8 @@ class RemoteArtifactSaver(Stage):
                         break
                 else:
                     remote_artifact = self._create_remote_artifact(d_artifact, content_artifact)
-                    needed_ras[remote_artifact.sha256] = remote_artifact
+                    key = f"{str(content_artifact.pk)}-{str(d_artifact.remote.pk)}"
+                    needed_ras[key] = remote_artifact
 
         return list(needed_ras.values())
 


pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -329,7 +329,13 @@ def _needed_remote_artifacts(self, batch):
                         break
                 else:
                     remote_artifact = self._create_remote_artifact(d_artifact, content_artifact)
-                    needed_ras[remote_artifact.sha256] = remote_artifact
+                    # we want to uniquely-identify the RA - use the first csum it has defined
+                    csums = [
+                        getattr(remote_artifact, a_sum)
+                        for a_sum in ("md5", "sha1", "sha224", "sha256", "sha384", "sha512")
```
dralley, line 335: Should we be using the list of allowed checksums?

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -329,7 +329,13 @@ def _needed_remote_artifacts(self, batch):
...
+                        if getattr(remote_artifact, a_sum, None)
+                    ]
+                    needed_ras[csums[0]] = remote_artifact
```
ggainey, line 338: Another option here is to just use remote_artifact.url . Thoughts on that as a uniqueness constraint?

pulpcore/plugin/stages/artifact_stages.py
ggainey: The remote might not have a supported-csum, especially in the migration case.

ggainey: Although, I should prob create the list from the full-list instead of recreating it here. Onnit.

pulpcore/plugin/stages/artifact_stages.py
goosemania: I think we are trying to avoid RA duplicates in one batch, correct? (see https://pulp.plan.io/issues/8489)

One option is to remove this deduplication code in Python and let the DB do it by enabling ignore_conflicts for bulk_create.
If it's not a good idea, then uniqueness constraint for RA is content_artifact+remote, should we just use their pks as a key?

What do you all think? @ggainey , @dralley

mdellweg: What happens, if you have the same file (matching checksum) in multiple locations (several urls)?
Do we want to reduce it to a unique one?

goosemania: I think that's ok and with alternate content source it will likely be the case. It's fine to have different urls/RAs to choose from for the same content, imo.

mdellweg: That would mean unique by url would be a good idea, right? As long as there are no contradicting checksums...

goosemania: It's likely fine but since the remote data is not in our or user's control, I wouldn't rely on it or try to act based on it (contradicting checksums or not). I'd just use it as is, that's why I suggested to use the fields from uniqueness constraint as a key.

ggainey: > One option is to remove this deduplication code in Python and let the DB do it by enabling ignore_conflicts for bulk_create.

bulk_create(ingore=True) has caused us (or at least me) SO MUCH pain in the last few months, I don't think it's a good idea.

> If it's not a good idea, then uniqueness constraint for RA is content_artifact+remote, should we just use their pks as a key?

contentartifact.pk+remote.pk is a good idea, I like it! Will give that a spin.

dralley: `ignore_conflicts` *should* be ok as long as we're not re-using the objects, but I can't blame you for being wary.

dralley: > contentartifact.pk+remote.pk is a good idea, I like it! Will give that a spin.

FWIW in my proof of concept I basically did this.  I used a "RemoteSource" uniqueness constraint of (remote, artifact).  So +1 to the general idea.

ggainey: > `ignore_conflicts` _should_ be ok as long as we're not re-using the objects, but I can't blame you for being wary.

My current success-rate for "using ignore=True is safe" is 0 for 2...

ggainey:
Tested against pulp_container CI with https://github.com/pulp/pulp_container/pull/300 - doesn't look like we broke anything, whew.

ggainey:
@goosemania @dralley re-review please when you have a moment.

dralley (approved):
This looks straightforwards and reasonable, thanks @ggainey

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -329,7 +329,8 @@ def _needed_remote_artifacts(self, batch):
                         break
                 else:
                     remote_artifact = self._create_remote_artifact(d_artifact, content_artifact)
-                    needed_ras[remote_artifact.sha256] = remote_artifact
+                    key = str(content_artifact.pk) + "-" + str(d_artifact.remote.pk)
```
goosemania, line 332: FWIW, f-strings can be used here nicely as well. Up to you.
`key = f'{content_artifact.pk}-{d_artifact.remote.pk}'`

CHANGES/8625.bugfix
```diff
@@ -0,0 +1 @@
+Taught artifact_stages to uniquely-identify content using a non-empty checksum-type.
```
goosemania, line 1: This needs an update, since the approach changed. Maybe something more user oriented will be more suitable here? E.g. fixed sync/migration from old repos which use other checksums than sha256?

pulpcore/plugin/stages/artifact_stages.py
ggainey: definitely, more-clear is always better

CHANGES/8625.bugfix
ggainey: Ah yes, good idea will do.

pulpcore/plugin/stages/artifact_stages.py
ggainey: done

goosemania (approved):

PR #1278, opened 2021-04-27
Taught PulpImporter to retry (once) on _import_file() failure.

ggainey:
There's a race condition in django-import-export's get_or_init_instance()
that is exercised by importing repo-versions concurrently. We attempt
an import and check for errors, retrying ONCE if encountered. On a
second error, fail the attempt.

The test added for pulp_rpm #7904 cover this case.

fixes #8633
ref #7904
[nocoverage]

### DIFF
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -55,14 +55,14 @@ def _destination_repo(importer, source_repo_name):
     return Repository.objects.get(name=dest_repo_name)
 
 
-def _import_file(fpath, resource_class):
+def _import_file(fpath, resource_class, do_raise=True):
     try:
         log.info(_("Importing file {}.").format(fpath))
         with open(fpath, "r") as json_file:
             data = Dataset().load(json_file.read(), format="json")
             resource = resource_class()
             log.info(_("...Importing resource {}.").format(resource.__class__.__name__))
-            return resource.import_data(data, raise_errors=True)
+            return resource.import_data(data, raise_errors=do_raise)
     except AttributeError:
         log.error(_("FAILURE importing file {}!").format(fpath))
         raise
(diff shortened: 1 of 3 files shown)


pulpcore/app/tasks/importer.py
```diff
@@ -158,7 +158,31 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
...
+            while True:
+                if a_result.has_errors():
+                    log.info(
+                        _("...{} import-errors encountered importing {} from {}, retrying").format(
+                            a_result.totals["error"], filename, rv_name
+                        )
+                    )
+                    # Second try will either succeed, or raise an error
+                    try:
+                        a_result = _import_file(os.path.join(rv_path, filename), res_class)
```
mdellweg, line 177: This will either succeed (as in have no errors) or raise an Exception, right?
In that case, there is no use for the `while True:` construct.

Instead, i would put the first invocation in a `try` construct, catch the specific constraints exception and do one retry in the `except` branch.

ggainey: I left the while-true construct in place because one thought I had was "is retrying exactly once the right thing to do?" I haven't convinced myself that there might be scenarios where multiple-retries are the Right Answer - but neither have I convinced myself one is The Way, either. Any thoughts?

RE retry-inside-except - mmm, yeah could do that. I try to avoid code that expects exceptions, inside an except: block, just because of Bad Experiences with nested exceptions in the past. May well be cleaner/more understandable here to do so tho, especially if "try just once" is the way we go.

ipanova: I would also get rid of while loop in case we end up re-trying once.

ggainey: And done - @ipanova @mdellweg re-review please?

ipanova (approved):

PR #1281, opened 2021-04-27
Fix reservation cleanup race condition

daviddavis:
fixes #8603

### DIFF
--- a/CHANGES/8603.bugfix
+++ b/CHANGES/8603.bugfix
@@ -1,0 +1,1 @@
+Handled a tasking race condition where cleaning up resource reservations sometimes raised an IntegrityError.
--- a/pulpcore/app/models/task.py
+++ b/pulpcore/app/models/task.py
@@ -8,6 +8,7 @@ from gettext import gettext as _
 
 from django.contrib.postgres.fields import JSONField
 from django.db import models
+from django.db.utils import IntegrityError
 from django.utils import timezone
 from rq.job import get_current_job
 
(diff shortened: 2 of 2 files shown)


CHANGES/8603.bugfix
```diff
@@ -0,0 +1 @@
+Fixed race condition when deleting records from ``core_reservedresource`` table.
```
dralley, line 1: Technically we're not fixing the race condition, we're just making sure it doesn't cause a failure.  The race is still there.

I think this is a good hotfix but I wonder if we should try to fix the actual race somehow, by locking.

daviddavis: Ok, I think I've fixed the race condition. It uses a single query to delete the reserved resources that are associated with the task but not any other task.

I'll look at adding a unit test but let me know what you think.

daviddavis: I tried the single query but it looks like calling `delete()` on the queryset actually issues a select and then deletes the individual rows. So that doesn't work.

I can look into locking but I am not sure what to lock on. The issue is that new TaskReservedResource records are being added to the db so I don't think locking on the ReservedResource will help..?

dralley (approved):
Approved, but we should maybe show this to @bmbouter when he gets back next week in case he has any ideas

mdellweg (approved):
I believe, this is safe, as the worst thing that can happen is that we do not delete a reservation.
It will be `get_or_create`ed the next time it will be used anyway.

PR #1283, opened 2021-04-27
Use https container

fao89:
[noissue]

### DIFF
--- a/.github/workflows/scripts/before_install.sh
+++ b/.github/workflows/scripts/before_install.sh
@@ -108,7 +108,7 @@ fi
 
 cd pulp-cli
 pip install -e .
-pulp config create --base-url http://pulp --location tests/settings.toml --no-verify-ssl
+pulp config create --base-url https://pulp --location tests/settings.toml 
 mkdir ~/.config/pulp
 cp tests/settings.toml ~/.config/pulp/settings.toml
 cd ..
--- a/.github/workflows/scripts/install.sh
+++ b/.github/workflows/scripts/install.sh
@@ -62,6 +62,8 @@ VARSYAML
 
 cat >> vars/main.yaml << VARSYAML
 pulp_settings: {"allowed_export_paths": ["/tmp"], "allowed_import_paths": ["/tmp"]}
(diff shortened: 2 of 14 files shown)


.github/workflows/scripts/before_install.sh
```diff
@@ -85,6 +85,7 @@ fi
 cd pulp-openapi-generator
 sed -i -e 's/localhost:24817/pulp/g' generate.sh
 sed -i -e 's/:24817/pulp/g' generate.sh
+sed -i -e 's/http:/https:/g' generate.sh
```
mdellweg, line 88: Hang on! Can we get that hostname as a parameter for `generate.sh`? This is starting to get weird.

.github/workflows/scripts/install.sh
```diff
@@ -91,6 +91,13 @@ fi
 ansible-playbook build_container.yaml
 ansible-playbook start_container.yaml
 
+# Hack: adding pulp CA to certifi.where()
+sudo docker cp pulp:/etc/pulp/certs/ca.crt /usr/local/share/ca-certificates/pulp_ca.crt
+CERT=$(python -c 'import certifi; print(certifi.where())')
```
fao89, line 96: This is Mozilla CA bundle, and I'm appending pulp CA to it

.github/workflows/scripts/script.sh
```diff
@@ -79,6 +81,7 @@ if [[ "$TEST" = 'bindings' || "$TEST" = 'publish' ]]; then
   gem build pulp_file_client
   gem install --both ./pulp_file_client-0.gem
   cd ..
+  export SSL_CERT_FILE=/usr/local/share/ca-certificates/pulp_ca.crt
```
fao89, line 84: but Ruby doesn't use `certifi` lib, so I had to use an OpenSSL env var to point to the CA

fao89:
Blocked by https://github.com/pulp/pulp-oci-images/pull/73 or https://github.com/pulp/pulp-oci-images/pull/84

PR #1284, opened 2021-04-28
3.11 cherrypicks

daviddavis:


goosemania:
It seems like we need to pin pulp_file here https://github.com/pulp/pulpcore/blob/3.11/template_config.yml#L5
And I think maybe the pip version needs to be specified for a release, let me find an example.

goosemania:
Not sure if pip_version_specifier is still used but it used to be used for release pipeline.
https://github.com/pulp/pulp-2to3-migration/blob/0.5/template_config.yml#L7

daviddavis:
@mdellweg I'm pulling in pulp_file 1.6 which is the version that supports pulpcore 3.11 but it looks like it includes the change that makes [the filesystem exporter api async](https://github.com/pulp/pulp_file/commit/6d4fad45ff97763b4a3ef625d0555f21299309e6). I believe I also need to pull in the test fixes in test_filesystemexport.py from 2264c7eca36d24db1c865ef6e52b4e7dfac085cd. Does that sound correct to you?

mdellweg:
> @mdellweg I'm pulling in pulp_file 1.6 which is the version that supports pulpcore 3.11 but it looks like it includes the change that makes [the filesystem exporter api async](https://github.com/pulp/pulp_file/commit/6d4fad45ff97763b4a3ef625d0555f21299309e6). I believe I also need to pull in the test fixes in test_filesystemexport.py from [2264c7e](https://github.com/pulp/pulpcore/commit/2264c7eca36d24db1c865ef6e52b4e7dfac085cd). Does that sound correct to you?

That sounds reasonable. Also those tests really seem to belong in the pulp_file repository.

goosemania (approved):

PR #1285, opened 2021-04-29
3.7.6 CP

goosemania:
backports #8625
https://pulp.plan.io/issues/8625

fixes #8651
https://pulp.plan.io/issues/8651

(cherry picked from commit eda2c890214a76e6f6ffa18cc939d04273b1fa13)

### DIFF
--- a/CHANGES/8651.bugfix
+++ b/CHANGES/8651.bugfix
@@ -1,0 +1,1 @@
+Backported a fix for on-demand sync/migration of repositories that don't have sha256 checksums.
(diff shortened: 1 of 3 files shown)


mdellweg:
I see two options to get the linting complaints out of the way here:
1. Pin black to an older version
2. Just run (the new version of) black against the release branch
Can't say, which one I prefer.

CHANGES/8651.bugfix
```diff
@@ -0,0 +1 @@
+Backported a fix for on-demand sync/migration of repositories that don't have sha256 checksums.
```
ipanova, line 1: suggestion, feel free to ignore -we usually link to the original issue that is being backported

ipanova:
> I see two options to get the linting complaints out of the way here:
> 
>     1. Pin black to an older version
> 
>     2. Just run (the new version of) black against the release branch
>        Can't say, which one I prefer.

I'd go with 2

ipanova (approved):

PR #1294, opened 2021-04-29
Add path to content-disposition header

dannysauer:
I'm using this basic thing to provide the repo path in query strings, making it easier to determine from CloudFront logs what repo and file were requested by a user without having to gather the Pulp logs together.  Before I create the issue, I thought I'd toss the PR out and see if there's any interest in such a thing.  Or if there's anything significant that should be changed to make this more palatable. If it's of interest, I'll open a proper ticket and add a changelog entry.  Otherwise I'll just keep maintaining a local patch.

* Add path to artifact to content-disposition header using a local x-pulp-artifact-path extension attribute
* Add config values S3_REPO_MANGLE_REGEX and S3_REPO_MANGLE_TO to configure values in artifact path which should be replaced (and what they're replaced with), because different URL libraries escape and unescape chars, breaking signatures.

PR #1295, opened 2021-04-29
Stabilize URLs before generating redirects

dannysauer:
The aio URL object mangles query strings when they're passed in. Specifically, it unescapes http entities which are escaped but which don't need escaped.  This is normally ok, but in the case of signed CloudFront URLs, changing anything about the URL invalidates the signature, including the querystring.  By pre-creating a yarl.URL object before raising the redirect exception, the parsing doesn't happen in aiohttp.  Passing in the "encoding=True" stops the reparsing when the object is created.

Fixes #8670
Relates to #1294, and boto/botocore#2363

### DIFF
--- a/CHANGES/8670.bugfix
+++ b/CHANGES/8670.bugfix
@@ -1,0 +1,1 @@
+Altered redirect URL escaping, preventing invalidation of signed URLs for artifacts using cloud storage.
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,3 +27,4 @@ redis>=3.4.0
 rq~=1.8.1
 setuptools>=39.2.0
 whitenoise>=5.0.0,<5.3.0
+yarl>1.0.0
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -9,6 +9,7 @@ from gettext import gettext as _
 from aiohttp.client_exceptions import ClientResponseError
 from aiohttp.web import FileResponse, StreamResponse, HTTPOk
 from aiohttp.web_exceptions import HTTPForbidden, HTTPFound, HTTPNotFound
+from yarl import URL
 
 import django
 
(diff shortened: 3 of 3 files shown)


dannysauer:
I'll do the squash and force-push after confirming that there aren't more linter changes. :)

dannysauer:
Do I need to add yarl to the setup.py, since it's not directly included from Pulp anywhere else?  It's an implicit dependency already pulled in via aiohttp.  I don't think any specific version would be required, since URL has been in Yarl since The Beginning.

CHANGES/8670.bugfix
```diff
@@ -0,0 +1 @@
+Pre-create Yarl.URL object before raising redirect when S3 or Azure storage backends are used, preventing html entinty re-encoding or decoding with signed URLs
```
mdellweg, line 1: For changelogs, we try to describe the problem, that was solved (as it would be interesting to the end user) in past tense. Ideally an admin would be able to recognize his failure in the bugfix section.
The how is more likely to be in the commit message (do not change that one, it's perfect).
So probably something like "Added URL querystring normalization for redirected signed cloud storage URLs to prevent invalidating the signature."

dkliban:
'

> @dkliban I was told the other way around for click yesterday. And i opt for explicit.

@dannysauer Let's be explicit and add it as a requirement. All the code changes look good to me.

dkliban:
@dannysauer How are these changes coming along?

dannysauer:
> @dannysauer How are these changes coming along?

Forgotten about. 😂 Thanks for the reminder.

dannysauer:
Finally got around to rebasing, integrating the changelog suggestion, and adding `yarl` to the requirements.txt.

requirements.txt
```diff
@@ -27,3 +27,4 @@ redis>=3.4.0
 rq~=1.8.1
 setuptools>=39.2.0
 whitenoise>=5.0.0,<5.3.0
+yarl>1.0.0
```
dralley, line 30: Use `yarl~=1.0.0` so that we don't pick up any new major versions.

dralley (changes requested):

dralley (approved):
I changed my mind, we probably don't need to be too selective about the Yarl version since as you mentioned `URL()` is fairly stable.

dannysauer:
> I changed my mind, we probably don't need to be too selective about the Yarl version since as you mentioned `URL()` is fairly stable.

I went back-and-forth on that, and figured it made more sense to just call out the minimum stable version. I'm [using 1.6.3](https://github.com/Kong/docker-pulp/blob/011ee381713ce44476eec8f5e24bc644d62c4d52/pulp-core/requirements.txt#L92) in our Docker container, for what that's worth.

dkliban (approved):
The upgrade tests failed because the test was bad. We've fixed the test since the CI ran on this PR. I am going to merge this change. Thank you for the contribution!

PR #1298, opened 2021-05-04
Fix small mistake

dralley:
[noissue]


mdellweg (approved):
That change is obviously correct.
No test did catch this, because no later stage is really interested in those content units. It only created unawaited tasks.

daviddavis (approved):

PR #1301, opened 2021-05-04
Encrypt Remote fields in the database

daviddavis:
fixes #8192


pulpcore/app/migrations/0064_encrypt_remote_fields.py
```diff
@@ -0,0 +1,89 @@
...
+            field=pulpcore.app.models.fields.EncryptedTextField(null=True),
+        ),
+        migrations.AlterField(
+            model_name='remote',
+            name='username',
+            field=pulpcore.app.models.fields.EncryptedTextField(null=True),
+        ),
+        migrations.RunPython(
+            code=encrypt_remote_fields,
+            reverse_code=unencrypt_remote_fields,
```
mdellweg, line 87: :heart:

pulpcore/app/migrations/0064_encrypt_remote_fields.py
```diff
@@ -0,0 +1,86 @@
...
+
+            if len(records) == 0:
+                break
+
+            for record in records:
+                update = {field: record[i] for i, field in enumerate(fields) if record[i]}
+                if not update:
+                    continue
+
+                Remote.objects.filter(pk=record[0]).update(**update)
```
mdellweg, line 31: I encountered that
`record.save(update_fields=[...])` is doing a similar thing.

daviddavis: Ah, did not know about `update_fields`. I wrote it this way to avoid having to do a select query to retrieve a `remote` object.

pulpcore/app/models/fields.py
```diff
@@ -68,6 +76,39 @@ def pre_save(self, model_instance, add):
...
+    @cached_property
+    def _fernet(self):
+        hkdf = HKDF(
+            algorithm=hashes.SHA256(),
+            length=32,
+            salt=b"pulp-hkdf-salt",
+            info=b"pulp-encrypted-field",
+            backend=default_backend(),
+        )
+        key = base64.urlsafe_b64encode(hkdf.derive(force_bytes(settings.SECRET_KEY)))
```
daviddavis, line 91: I am using `SECRET_KEY` here but the issue suggests creating a new key/setting. Any thoughts?

pulpcore/app/migrations/0064_encrypt_remote_fields.py
mdellweg: They should be äquivalent anyway.

pulpcore/app/models/fields.py
mdellweg: I don't know what is "right" here. But i think the django secret-key is used for cookies and xsrf-tokens.
Maybe that implies it might have a shorter lifetime until rekeying than stuff we want to save to the database.

pulpcore/tests/unit/models/test_remote.py
```diff
@@ -0,0 +1,31 @@
...
+
+class RepositoryTestCase(TestCase):
+    def setUp(self):
+        self.remote = None
+
+    def tearDown(self):
+        if self.remote:
+            self.remote.delete()
+
+    @override_settings(SECRET_KEY="test")
```
mdellweg, line 18: :ok_hand:

matburt:
👍 lgtm from the perspective of https://github.com/ansible/awx

daviddavis:
> +1 lgtm from the perspective of https://github.com/ansible/awx

Hey, thanks for the review. We're making one final tweak to write the key to a secure location on the filesystem (instead of using SECRET_KEY). I can ping you again for a final review when that part is done.

Thanks again for taking a look.

dralley:
As per the pulp-meeting discussion, we should avoid merging any PRs containing migrations before Friday to enable the possibility for 3.14.z-stream migrations.  And we'll decide then how to proceed.

pulpcore/app/migrations/0069_encrypt_remote_fields.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+
+        with connection.cursor() as cursor:
+            cursor.execute(query)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("core", "0068_add_timestamp_of_interest"),
```
fao89, line 62:
```suggestion
        ("core", "0070_rename_retained_versions"),
```

fao89 (commented):
s/0069_encrypt_remote_fields/0071_encrypt_remote_fields

docs/configuration/settings.rst
```diff
@@ -36,6 +36,22 @@ SECRET_KEY
...
+
+  The key is automatically generated by default with pulp_installer and single container image. To
+  generate your own key though, you can use Python::
+
+    pip install cryptography
+    python3 -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())" > /etc/pulp/certs/database_fields.symmetric.key
+
+  Or you can use openssl::
+
+    openssl rand -base64 32 > /etc/pulp/certs/database_fields.symmetric.key
```
mdellweg, line 53: Now the python method looks overly complicated for 256 coin flips.
Would it help the admin if we say "...a symmetric fernet key of 32 random bytes that Pulp..."?

pulpcore/app/models/fields.py
```diff
@@ -68,6 +79,33 @@ def pre_save(self, model_instance, add):
         return super().pre_save(model_instance, add)
 
 
+class EncryptedTextField(TextField):
+    """A field mixin that provides functionality to encrypt data using settings.SECRET_KEY."""
+
+    @cached_property
```
mdellweg, line 85: :1st_place_medal:

pulpcore/app/models/fields.py
```diff
@@ -68,6 +79,33 @@ def pre_save(self, model_instance, add):
...
+    def _fernet(self):
+        _logger.debug(f"Loading encryption key from {settings.DB_ENCRYPTION_KEY}")
+        with open(settings.DB_ENCRYPTION_KEY, "rb") as key_file:
+            hkdf = HKDF(
+                algorithm=hashes.SHA256(),
+                length=32,
+                salt=b"pulp-hkdf-salt",
```
mdellweg, line 92: What is the purpose of this salt value?
If this HKDF is doing some password to key transformation, i do not see the cryptographical benefit of doing so. We already start with 32 bytes of _good_ random data.

pulpcore/app/settings.py
```diff
@@ -289,6 +292,8 @@
 )
 # HERE ENDS DYNACONF EXTENSION LOAD (No more code below this line)
 
+_SKIPPED_COMMANDS_FOR_SETTING_CHECKS = ["handle-artifact-checksums", "migrate", "collectstatic"]
```
mdellweg, line 295: With the migration you introduced, i think the migrate command really needs the encryption key to be present.

docs/configuration/settings.rst
daviddavis: My concern is that some methods of generating the fernet key might not have enough entropy to be secure but I'm fine with having just leaving it up to the user.

pulpcore/app/models/fields.py
daviddavis: I also don't see a benefit but the code requires it: `TypeError: __init__() missing 1 required positional argument: 'salt'`.

It does look like [the cryptography code](https://github.com/pyca/cryptography/blob/f0b2269f7221f06aacf883b58577ffc9627e7522/src/cryptography/hazmat/primitives/kdf/hkdf.py#L39-L40) will use a string of `b"\x00"` but I'd still have to specify `salt=None` and I figure it's better to explicitly set the salt instead of letting the cryptography code implicitly set it.

daviddavis: Actually I think I can eliminate the HKDF and just feed the key straight to `Fernet()`. Will look into that.

pulpcore/app/settings.py
daviddavis: I removed migrate from the skipped commands but now I have a problem: the single container fails to build. I imagine this is because [it creates the key](https://github.com/pulp/pulp-oci-images/blob/98fb9ee3b91d137945a16dfd657eca51fd7de428/assets/pulpcore-api.prep#L9-L15) after migrating. Any thoughts on how to fix this?

daviddavis: Maybe I need to create the key in [the postgres prep file](https://github.com/pulp/pulp-oci-images/blob/d7413e6f0f1f56a6b61288f86939f7261d63da54/assets/000-postgres.prep)?

pulpcore/app/settings.py
```diff
@@ -304,6 +309,17 @@
         )
     )
 
+if not (len(sys.argv) >= 2 and sys.argv[1] in ["sphinx-build", "collectstatic"]):
+    try:
+        with open(DB_ENCRYPTION_KEY, "rb") as key_file:
+            Fernet(key_file.read())
```
mdellweg, line 315: :+1:
I assume this will raise if we don't have enough bits in that file.

mdellweg (approved):
You need to have the key in the right place (single container, installer, operator) obviously, and you need to rerun black.
This looks good to me.

pulpcore/app/migrations/0072_encrypt_remote_fields.py
```diff
@@ -0,0 +1,95 @@
...
+                break
+
+            for record in records:
+                update = {
+                    field: record[i] for i, field in enumerate(fields, 1) if record[i] is not None
+                }
+                if not update:
+                    continue
+
+                Remote.objects.filter(pk=record[0]).update(**update)
```
goosemania, line 35: Just for my understanding...
The encryption happens when you update the fields because you change the type of the fields to EncryptedTextField, correct?
And you read the data for remotes using SQL, so the EncryptedTextField didn't try to unencrypt those plaintext values, correct?

pulpcore/app/models/fields.py
```diff
@@ -68,6 +75,25 @@ def pre_save(self, model_instance, add):
         return super().pre_save(model_instance, add)
 
 
+class EncryptedTextField(TextField):
+    """A field mixin that provides functionality to encrypt data using settings.SECRET_KEY."""
```
goosemania, line 79: Did you mean `settings.DB_ENCRYPTION_KEY`?

pulpcore/app/models/fields.py
```diff
@@ -68,6 +75,25 @@ def pre_save(self, model_instance, add):
         return super().pre_save(model_instance, add)
 
 
+class EncryptedTextField(TextField):
```
goosemania, line 78: I came across the implementation of such a field  https://github.com/orcasgit/django-fernet-fields/blob/master/fernet_fields/fields.py#L22.
It looks old and unmaintained. But what I noticed, they check that this field is not used in indices or uniqueness constraints. We do not use either of those in remotes but since it's a non-remote specific type of field and potentially can be used for other sensitive data, does it make sense to check that it's not used for indexing or uniqueness?
(It can be a separate PR if you agree in general with the concern.)

pulpcore/app/migrations/0072_encrypt_remote_fields.py
mdellweg: That is exactly my understanding.

daviddavis: That's correct.

pulpcore/app/models/fields.py
daviddavis: I'm a little unclear about what would happen if you used an EncryptedTextField as a index or as part of unique constraint but I agree it probably won't work. Will update.

pulpcore/app/settings.py
daviddavis: Yes, exactly:

```
>>> Fernet(b"vFLzZN4ZA/cBrzqXqf2Bs3suzYijkDmP")
  File "/home/daviddavis/.local/lib/python3.9/site-packages/cryptography/fernet.py", line 34, in __init__
    raise ValueError(
ValueError: Fernet key must be 32 url-safe base64-encoded bytes.
```

daviddavis:
I think I addressed everything.

goosemania (approved):
Thank you!

daviddavis:
Going to wait on https://github.com/pulp/pulp_installer/pull/722 before I merge this.

mdellweg (approved):

PR #1303, opened 2021-05-04
Remove BaseDistribution from plugin api

daviddavis:
fixes #8386


pulpcore/content/handler.py
```diff
@@ -177,23 +175,21 @@ def _match_distribution(cls, path):
...
+                )
         else:
             try:
-                model_class = cls.distribution_model
                 return cls.distribution_model.objects.get(base_path__in=base_paths)
             except ObjectDoesNotExist:
                 log.debug(
-                    _("{model_name} not matched for {path} using: {base_paths}").format(
-                        model_name=model_class.__name__, path=path, base_paths=base_paths
+                    _("Distributoin not matched for {path} using: {base_paths}").format(
```
dralley, line 191: typo

pulpcore/app/models/publication.py
```diff
@@ -290,38 +287,9 @@ class BaseDistribution(MasterModel):
...
 
     def __init__(self, *args, **kwargs):
-        """Initialize a BaseDistribution and emit deprecation warnings"""
-        deprecation_logger.warn(
-            _(
-                "BaseDistribution is deprecated and could be removed as early as pulpcore==3.13; "
-                "use pulpcore.plugin.models.Distribution instead."
-            )
+        raise Exception(
+            "Using BaseDistribution is no longer supported. Please use Distribution instead"
```
mdellweg, line 291: Does makemigration instantiate such models? Can you verify that it does not blow up here?
Is "NotImplementedError" a better matching exception?

ipanova: please use `pulpcore.plugin.models.Distribution` instead

daviddavis: > Does makemigration instantiate such models? Can you verify that it does not blow up here?

Good question. I tested and it seems to work.

> Is "NotImplementedError" a better matching exception?

I considered it but reading the description for NotImplementedError and it sounds like it's maybe not a good fit?

https://docs.python.org/3/library/exceptions.html#NotImplementedError

dralley (approved):

mdellweg (approved):
LGTM

PR #1306, opened 2021-05-05
Adds compatability with Django==2.2.21

bmbouter:
Switches the ArtifactFileField handling to use
`django.db.models.fields.Field` and specifies the `name` which causes
Django to no longer try to generate one from the filename.

closes #8691

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ aiodns~=2.0.0
 aiofiles==0.6.0
 backoff~=1.10.0
 click~=7.1.2
-Django==2.2.20  # LTS version, switch only if we have a compelling reason to
+Django~=2.2.23  # LTS version, switch only if we have a compelling reason to
 django-currentuser~=0.5.3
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -3,7 +3,7 @@ aiohttp~=3.7.4
 aiodns~=2.0.0
 aiofiles==0.6.0
 backoff~=1.10.0
-Django==2.2.20  # LTS version, switch only if we have a compelling reason to
+Django @ git+https://github.com/bmbouter/django.git@2.2.21-patched
```
fao89, line 6: can we try patch `2.2.22`?

bmbouter: :+1:  I'm now testing against the pending 2.2.23 release.

daviddavis (approved):

dkliban (approved):

fao89:
FYI @newswangerd

PR #1308, opened 2021-05-06
Skip allowed content checks on collectstatic

ekohl:
Since 9a9a06f1e37f5367e36a24b9aa3aca6ca78dab53 DB access is always needed. In later commits this was extended with additional checks.

However, collectstatic should be a fully offline operation that doesn't access the network. As such, content checking is not needed.

Doing this allows the Foreman/Katello installer to run it more freely.  The alternative is to add additional ordering. However, I'd argue that this is better and also slightly faster.

Fixes #8711

### DIFF
--- a/CHANGES/8711.bugfix
+++ b/CHANGES/8711.bugfix
@@ -1,0 +1,1 @@
+Skip allowed content checks on collectstatic
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -318,7 +318,9 @@ FORBIDDEN_CHECKSUMS = set(constants.ALL_KNOWN_CONTENT_CHECKSUMS).difference(
     ALLOWED_CONTENT_CHECKSUMS
 )
 
-if not (len(sys.argv) >= 2 and sys.argv[1] in ["handle-artifact-checksums", "migrate"]):
+_SKIPPED_COMMANDS_FOR_CONTENT_CHECKS = ["handle-artifact-checksums", "migrate", "collectstatic"]
+
+if not (len(sys.argv) >= 2 and sys.argv[1] in _SKIPPED_COMMANDS_FOR_CONTENT_CHECKS):
     try:
         with connection.cursor() as cursor:
             for checksum in ALLOWED_CONTENT_CHECKSUMS:


daviddavis:
Can you add a changelog entry (`CHANGES/8711.bugfix`)?

ekohl:
Argh, I should give it the correct filename. `git mv CHANGES/8177.bugfix CHANGES/8711.bugfix` should help.

daviddavis (approved):

mdellweg (approved):

bmbouter (approved):
This is a good idea. Thanks!

ekohl:
While CI is running, let me already ask about cherry picking this back to a stable release. Is this something that could be picked to 3.11?

daviddavis:
Hmm that's a good question. It does sort of straddle the line between bug fix and enhancement. I have no objections to backporting this. Thoughts @mdellweg @bmbouter ?

ekohl:
You can argue that it's a regression fix. Prior to 9a9a06f1e37f5367e36a24b9aa3aca6ca78dab53 there was no need for the database during `collectstatic`. After it was. Now it's not particularly critical since it gracefully fails if there is no DB.

The way we hit this was that on an upgrade `collectstatic` ran before `migrate`. I was told there was a migration we needed to run to avoid hitting the check, but we couldn't reach that. Now we can order `collectstatic` after `migrate` but I'm always in favor of handling the variations.

If there is an order requirement, that should be clearly documented.

bmbouter:
> Hmm that's a good question. It does sort of straddle the line between bug fix and enhancement. I have no objections to backporting this. Thoughts @mdellweg @bmbouter ?

I'm ok with a backport. My thinking is that our startup checks were never intended to prevent `collectstatic` which is not actually starting Pulp, so we introduced a bug.

daviddavis:
@ekohl can you file a backport request at https://pulp.plan.io/projects/pulp/issues/new ? Be sure to include the issue number (8711) and the release (3.11).

ekohl:
I opened https://pulp.plan.io/issues/8712. Please verify it's correct. Thanks for your consideration!

PR #1309, opened 2021-05-06
Move directory listing template external [WIP]

dannysauer:
It'd be nice to be able to make the content server's download page a little prettier without having to custom-build pulp.  In order to do that, this PR moves the download page to an external Django template, and adds support to serve static file content from the regular django static location (as opposed to adding the content as a release).  If a conflict occurs between existing content and the static path, the actual pulp content distribution wins out.  So this is backwards compatible for anyone who is using /static in a current publication.

### DIFF
--- a/pulpcore/app/static/directory_listing.css
+++ b/pulpcore/app/static/directory_listing.css
@@ -1,0 +1,3 @@
+ul {
+  list-style-type: square;
+}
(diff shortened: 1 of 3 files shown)


dkliban:
@dannysauer  I am closing this for now. Feel free to re-open when you are ready to work on it.

PR #1313, opened 2021-05-07
Change migration to use django's get_model

daviddavis:
fixes #8656


pulpcore/app/migrations/0061_call_handle_artifact_checksums_command.py
```diff
@@ -1,11 +1,53 @@
...
-    call_command('handle-artifact-checksums')
+    Artifact = apps.get_model("core", "Artifact")
+    paths = []
+
+    # set allowed checksums
+    for checksum in settings.ALLOWED_CONTENT_CHECKSUMS:
+        params = {f"{checksum}__isnull": True}
+        artifacts_qs = Artifact.objects.filter(**params)
+
+        for a in artifacts_qs:
```
dralley, line 23: `.iterator()`

bmbouter:
I ran two tests on this.

## Upgrade and expected to remove sha1 and md5

1. Install a fresh pulpcore==3.10.0 (the one without this migration at all) and pulp_file==1.6.1 and set `ALLOWED_CONTENT_CHECKSUMS = ['md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512']`
2. Sync a file repo
3. verify in the db that all checksums are calculated and present
4. Stop pulp and modify the setting to `ALLOWED_CONTENT_CHECKSUMS = ['sha224', 'sha256', 'sha384', 'sha512']
4. upgrade to this PR and latest pulp_file which includes running this migration
5. verify in the db that sha1 and md5 have been removed and pulp starts without complaint

## Upgrade and expected to add sha1 and md5

1. Install a fresh pulpcore==3.10.0 (the one without this migration at all) and pulp_file==1.6.1 and set `ALLOWED_CONTENT_CHECKSUMS = ['sha224', 'sha256', 'sha384', 'sha512']`
2. Sync a file repo
3. verify in the db that all checksums are calculated and present except for sha1 and md5
4. Stop pulp and modify the setting to `ALLOWED_CONTENT_CHECKSUMS = ['md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512']
4. upgrade to this PR and latest pulp_file which includes running this migration
5. verify in the db that sha1 and md5 have been calculated and are present and pulp starts without complaint

bmbouter (approved):
I also read through the code. This all look good to me. Thank you so much @daviddavis !

pulpcore/app/migrations/0061_call_handle_artifact_checksums_command.py
```diff
@@ -1,11 +1,53 @@
...
+                setattr(a, checksum, hasher.hexdigest())
+            except FileNotFoundError:
+                file_path = os.path.join(settings.MEDIA_ROOT, a.file.name)
+                paths.append(file_path)
+
+        if artifacts_qs:
+            Artifact.objects.bulk_update(objs=artifacts_qs, fields=[checksum], batch_size=1000)
+        if paths:
+            raise Exception(
+                _("Missing files needed to update artifact checksums: {}").format(paths)
```
ipanova, line 38: what is the suggested course of action for the user who has missing files and migration failing in the middle of the upgrade? can we not raise the error but just log that there are missing files? user can run `handle-artifact-checksums` post-upgrade which will download missing files.

mdellweg: Good observation! :+1:

daviddavis: I think we need to allow the migration to proceed and not error. 

Consider the case where I am adding new fields to Artifact for [the orphan cleanup work](https://pulp.plan.io/issues/7659). A user upgrades, runs these migrations, migration throws an error. They have no remedy since they can't run `handle-artifact-checksums` at this point (will complain about the missing field in the table).

So I think the migration should continue and not error. And Pulp will refuse to start if the artifact is missing checksums (or has prohibited checksums).

daviddavis: Or should I use `logger.warn()`?

daviddavis: Going with `logger.warn()`.

pulpcore/app/migrations/0061_call_handle_artifact_checksums_command.py
```diff
@@ -1,11 +1,57 @@
...
+                a.file.close()
+                setattr(a, checksum, hasher.hexdigest())
+            except FileNotFoundError:
+                file_path = os.path.join(settings.MEDIA_ROOT, a.file.name)
+                paths.append(file_path)
+
+        if artifacts_qs:
+            Artifact.objects.bulk_update(objs=artifacts_qs, fields=[checksum], batch_size=1000)
+        if paths:
+            _logger.warn(
```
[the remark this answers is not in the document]
daviddavis: My concern is that sets use a large amount of memory. I think having duplicate warnings is probably acceptable trade off.

pulpcore/app/migrations/0061_call_handle_artifact_checksums_command.py
```diff
@@ -1,11 +1,57 @@
...
-    call_command('handle-artifact-checksums')
+    Artifact = apps.get_model("core", "Artifact")
+    paths = []
+
+    # set allowed checksums
+    for checksum in settings.ALLOWED_CONTENT_CHECKSUMS:
+        params = {f"{checksum}__isnull": True}
+        artifacts_qs = Artifact.objects.filter(**params)
+
+        for a in artifacts_qs:
```
dralley, line 26: @daviddavis If we're concerned about memory usage, we should probably use `.iterator()` with internal batching of the saves, so that we can clear them out of memory. 

```python
      artifacts = []
      for a in artifacts_qs.iterator():
          # do stuff
          artifacts.append(artifact)
          if len(artifacts) >= 1000:
              Artifact.objects.bulk_update(objs=artifacts, fields=[checksum])
              artifacts.clear()
              
```

pulpcore/app/migrations/0061_call_handle_artifact_checksums_command.py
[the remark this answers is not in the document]
daviddavis: I updated the error message to specify which checksum it's missing artifact files for so the error message should no longer be duplicated.

mdellweg (approved):
I see all comments addressed, and even the management command was improved with the iterator and batching.

PR #1314, opened 2021-05-10
Use a common entrypoint for workers

mdellweg:
With this change, workers should be started with `pulpcore-worker` and
the resource-manager with `pulpcore-worker --resource-manager`
accordingly.

fixes #8721
https://pulp.plan.io/issues/8721

### DIFF
--- a/pulpcore/tasking/worker.py
+++ b/pulpcore/tasking/worker.py
@@ -61,16 +61,16 @@ class PulpWorker(Worker):
 
     def __init__(self, queues, **kwargs):
 
-        if kwargs["name"]:
+        if kwargs.get("name"):
             kwargs["name"] = kwargs["name"].replace("%h", socket.getfqdn())
         else:
             kwargs["name"] = "{pid}@{hostname}".format(pid=os.getpid(), hostname=socket.getfqdn())
 
         if kwargs["name"] == TASKING_CONSTANTS.RESOURCE_MANAGER_WORKER_NAME:
-            queues = [Queue("resource-manager", connection=kwargs["connection"])]
+            queues = [Queue("resource-manager", connection=kwargs.get("connection"))]
             self.is_resource_manager = True
         else:
-            queues = [Queue(kwargs["name"], connection=kwargs["connection"])]
+            queues = [Queue(kwargs["name"], connection=kwargs.get("connection"))]
             self.is_resource_manager = False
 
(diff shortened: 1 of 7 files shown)


pulpcore/tasking/worker.py
```diff
@@ -61,16 +61,16 @@ class PulpWorker(Worker):
 
     def __init__(self, queues, **kwargs):
 
-        if kwargs["name"]:
+        if kwargs.get("name"):
```
ipanova, line 64: :+1:

ipanova:
was this supposed to be aligned with 3.13?

mdellweg:
> was this supposed to be aligned with 3.13?

That is the goal, to make the transition to any new tasking system in the future much more smoothly. :fox_face:

mdellweg:
CI failure seems to be due to a new version of sphinx.
I will look into this on another PR.

ipanova:
> > was this supposed to be aligned with 3.13?
> 
> That is the goal, to make the transition to any new tasking system in the future much more smoothly. fox_face

It was a gentle nudge to update redmine :cake:  :smile_cat:

ipanova (approved):

mdellweg:
See also https://github.com/pulp/pulp_installer/pull/616

requirements.txt
```diff
@@ -3,6 +3,7 @@ aiohttp~=3.7.4
 aiodns~=2.0.0
 aiofiles==0.6.0
 backoff~=1.10.0
+click~=7.1.2
```
bmbouter, line 6: I'm not sure what's best but perhaps we should leave this undeclared for now? What you've done is traditionally the best practice I believe. The concern is what happens if another RQ were to upgrade to 8.0 in an RQ z-stream...

bmbouter:
Do we want the CI to switch to the new-style worker commands? I'm not immediately clear on how to do that... What do you recommend?

requirements.txt
mdellweg: Shouldn't be a problem. RQ is pinned to a specific version.

mdellweg:
The CI integration of this should really be handled by an intelligent startup script along the lines of https://github.com/pulp/pulp-oci-images/commit/e96c8497f1aeb71af62eaf4d16e9ddff09e02445
And for now, the workers keep being able to be started via rq unmodified.

bmbouter (approved):
This looks great to me!

PR #1320, opened 2021-05-11
Revert "Removed ``pulpcore.plugin.tasking.enqueue_with_reservation``."

dralley:
Reverts pulp/pulpcore#1318


bmbouter (dismissed):
On the call @ggainey and @dralley identified that reverting this single PR would allow them to release pulp_rpm 3.11 which is needed by a lot of folks. It's not the preferred outcome, but the motivations make sense and this will be reintroduced post-release of pulp_rpm.

PR #1323, opened 2021-05-13
Fixes ability to specify custom headers on a Remote

dkliban:
The Remotes APIs accepts a list of headers to be sent with each request during
sync. However, the headers were not being properly passed to aiohttp. This patch
fixes how the headers are passed to aiohttp.

fixes: #8689
https://pulp.plan.io/issues/8689

### DIFF
--- a/pulpcore/download/factory.py
+++ b/pulpcore/download/factory.py
@@ -3,6 +3,7 @@ import asyncio
 import atexit
 import copy
 from gettext import gettext as _
+from multidict import MultiDict
 import platform
 from pkg_resources import get_distribution
 import ssl
@@ -14,8 +15,6 @@ import aiohttp
 
 from .http import HttpDownloader
 from .file import FileDownloader
-
-import json
 
 
 PROTOCOL_MAP = {
@@ -124,9 +123,13 @@ class DownloaderFactory:
         if sslcontext:
             tcp_conn_opts["ssl_context"] = sslcontext
 
-        headers = {"User-Agent": user_agent()}
+        headers = MultiDict({"User-Agent": user_agent()})
         if self._remote.headers is not None:
-            headers.update(json.loads(self._remote.headers))
+            for header_dict in self._remote.headers:
+                user_agent_header = header_dict.pop("User-Agent", None)
+                if user_agent_header:
+                    headers["User-Agent"] = f"{headers['User-Agent']}, {user_agent_header}"
+                headers.extend(header_dict)
 
         conn = aiohttp.TCPConnector(**tcp_conn_opts)
         total = self._remote.total_timeout
--- a/pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
@@ -133,6 +133,7 @@ class AddRemoveContentTestCase(unittest.TestCase):
         * The ``content_removed_summary`` attribute is correct.
         """
         body = gen_file_remote()
+        body.update({"headers": [{"Connection": "keep-alive"}]})
         self.remote.update(self.client.post(FILE_REMOTE_PATH, body))
         sync(self.cfg, self.remote, self.repo)
         repo = self.client.get(self.repo["pulp_href"])
--- a/CHANGES/8689.bugfix
+++ b/CHANGES/8689.bugfix
@@ -1,0 +1,1 @@
+Fixed ability to specify custom headers on a Remote.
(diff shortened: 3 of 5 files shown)


ipanova (approved):

pulpcore/app/serializers/repository.py
```diff
@@ -168,8 +168,7 @@ class RemoteSerializer(ModelSerializer):
         help_text="aiohttp.ClientTimeout.sock_read (q.v.) for download-connections.",
         min_value=0.0,
     )
-    headers = serializers.ListField(
-        child=serializers.DictField(),
```
bmbouter, line 172: This line was to prevent things like `[1, 4, "somestring"] which won't be valid headers. Having a test asserts those are invalid would be pretty great also. @dkliban wdyt?

pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
```diff
@@ -133,6 +133,7 @@ def test_02_sync_content(self):
         * The ``content_removed_summary`` attribute is correct.
         """
         body = gen_file_remote()
+        body.update({"headers": {"Connection": "keep-alive"}})
```
bmbouter, line 136: Yeah I think this needs to be a list of dicts.

pulpcore/app/serializers/repository.py
dkliban: @bmbouter I can add a test for that.

bmbouter: Oh yeah this happened on IRC, but we recalled this comment in discussion yesterday https://github.com/pulp/pulpcore/pull/1041#discussion_r563525571

pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
```diff
@@ -133,6 +133,7 @@ def test_02_sync_content(self):
         * The ``content_removed_summary`` attribute is correct.
         """
         body = gen_file_remote()
+        body.update({"headers": [{"Connection": "keep-alive"}]})
```
bmbouter, line 136: I would lgtm this as-is, great!

How do you feel about adding a test that asserts Pulp rejecting a header specified like `{"headers": {"Connection": "keep-alive"}}` and `{"headers": [1, 2, 3]}` ?

dkliban: @bmbouter I added a test for this.

pulpcore/download/factory.py
```diff
@@ -124,9 +123,10 @@ def _make_aiohttp_session_from_remote(self):
         if sslcontext:
             tcp_conn_opts["ssl_context"] = sslcontext
 
-        headers = {"User-Agent": user_agent()}
+        headers = MultiDict({"User-Agent": user_agent()})
         if self._remote.headers is not None:
-            headers.update(json.loads(self._remote.headers))
+            for header_dict in self._remote.headers:
+                headers.extend(header_dict)
```
daviddavis, line 129: Do you suppose that users may want to be able to set/override the 'User-Agent' header? If so, then maybe this should be:

```python
headers = MultiDict()
if self._remote.headers is not None:
    for header_dict in self._remote.headers:
        headers.extend(header_dict)
if 'User-Agent' not in headers:
    headers['User-Agent'] = user_agent()
```

dkliban: @daviddavis I specifically did not want users to override it, but perhaps that's not right.

daviddavis: This code allows the User-Agent to be set twice. Is that the right behavior?

dkliban: If the user adds a user-agent string, it will be appended to the user agent string that pulp sets. That's my understanding of how the values for headers with the same name are treated.

daviddavis: Are you sure? I had assumed that it would create multiple headers.

dkliban: The last paragraph here makes me think so: https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

daviddavis: I was able to test it out with aiohttp and it looks it sends multiple User-Agent headers without concatenating them. Here's my code:

https://gist.github.com/daviddavis/34e1b4be12987b05cd0d4990989d2159

daviddavis: Fwiw I noticed that aiohttp sets a default User-Agent which can be overridden by specifying it.

ipanova: `user-agent` came fom this PR  https://github.com/pulp/pulpcore/commit/82136ae1588afffeee844787610c9f4b07fe7f44 and I am not entirely sure we want the user to change the user-agent

ipanova: i checked some of their tests and it seems like they merge headers differently than specified in the rfc https://github.com/pulp/pulpcore/commit/82136ae1588afffeee844787610c9f4b07fe7f44

daviddavis (approved):

ipanova (approved):
Thanks for adding those tests!

PR #1328, opened 2021-05-14
Content app auto-distributes latest publication

gerrod3:
@bmbouter Here is the problem(and assumption) I ran into https://pulp.plan.io/issues/8760#note-2

fixes: #8760
https://pulp.plan.io/issues/8760

### DIFF
--- a/CHANGES/8760.feature
+++ b/CHANGES/8760.feature
@@ -1,0 +1,1 @@
+Content app auto-distributes latest publication if distribution's ``repository`` field is set
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -28,6 +28,7 @@ from pulpcore.app.models import (  # noqa: E402: module level not at top of file
     BaseDistribution,
     ContentArtifact,
     Distribution,
+    Publication,
     Remote,
     RemoteArtifact,
 )
@@ -368,6 +369,20 @@ class Handler:
         headers = self.response_headers(rel_path)
 
         publication = getattr(distro, "publication", None)
+        repository = getattr(distro, "repository", None)
+        repo_version = getattr(distro, "repository_version", None)
+        if repository:
+            repo_version = repository.latest_version()
+            # Search for publication serving the closest latest version
+            if not publication:
+                try:
+                    versions = repository.versions.all()
+                    publications = Publication.objects.filter(
+                        repository_version__in=versions, complete=True
+                    )
+                    publication = publications.latest("repository_version", "-pulp_created")
+                except ObjectDoesNotExist:
+                    pass
 
         if publication:
             if rel_path == "" or rel_path[-1] == "/":
@@ -419,13 +434,7 @@ class Handler:
                             request, StreamResponse(headers=headers), ca
                         )
 
-        repo_version = getattr(distro, "repository_version", None)
-        repository = getattr(distro, "repository", None)
-
-        if repository or repo_version:
-            if repository:
-                repo_version = distro.repository.latest_version()
-
+        if repo_version:
             if rel_path == "" or rel_path[-1] == "/":
                 try:
                     index_path = "{}index.html".format(rel_path)
--- a/pulpcore/tests/functional/api/using_plugin/test_distributions.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_distributions.py
@@ -17,6 +17,12 @@ from pulp_smash.pulp3.utils import (
 )
 from requests.exceptions import HTTPError
 
+from pulpcore.client.pulp_file import (
+    DistributionsFileApi,
+    PublicationsFileApi,
+    RemotesFileApi,
+    RepositoriesFileApi,
+)
 from pulpcore.tests.functional.api.using_plugin.constants import (
     FILE_CONTENT_NAME,
     FILE_DISTRIBUTION_PATH,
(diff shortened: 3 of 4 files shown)


pulpcore/content/handler.py
```diff
@@ -368,6 +368,15 @@ async def _match_and_stream(self, path, request):
         headers = self.response_headers(rel_path)
 
         publication = getattr(distro, "publication", None)
+        repository = getattr(distro, "repository", None)
+        repo_version = getattr(distro, "repository_version", None)
+        if repository:
+            repo_version = distro.repository.latest_version()
+            # Search for publication serving latest_version
+            try:
+                publication = repo_version.publication_set.earliest("pulp_created")
```
mdellweg, line 377: Can the distribution have both a repository and a publication set?

pulpcore/content/handler.py
```diff
@@ -368,6 +368,15 @@ async def _match_and_stream(self, path, request):
         headers = self.response_headers(rel_path)
 
         publication = getattr(distro, "publication", None)
+        repository = getattr(distro, "repository", None)
+        repo_version = getattr(distro, "repository_version", None)
+        if repository:
+            repo_version = distro.repository.latest_version()
```
mdellweg, line 374:
```suggestion
            repo_version = repository.latest_version()
```

pulpcore/content/handler.py
mdellweg: I'm not sure i understand the logic here.
What if we have a repository that needs publications, and a new version has just been created, but the autopublishing is still in progress. Then the content app will attempt to handle this repository version as one without publications, right?

gerrod3: Yes, you would end up getting a 404 if you tried accessing a file that was only created through a publication. If you were to access a piece of content it would probably work. This means for a time you could access content inside of a repository that would not have been exposed due to publication settings.

mdellweg: Should we use the latest version that has a publication instead in those cases?

gerrod3: I think so.

pulpcore/content/handler.py
```diff
@@ -368,6 +369,18 @@ async def _match_and_stream(self, path, request):
...
+        repository = getattr(distro, "repository", None)
+        repo_version = getattr(distro, "repository_version", None)
+        if repository:
+            repo_version = repository.latest_version()
+            # Search for publication serving the closest latest version
+            if not publication:
+                try:
+                    versions = repository.versions.all()
+                    publications = Publication.objects.filter(repository_version__in=versions)
+                    publication = publications.latest("repository_version", "-pulp_created")
```
dralley, line 381: Should we add an index for this?

mdellweg: Example of how to do it:
https://github.com/pulp/pulpcore/pull/1261/files#diff-b6921076eabe69f8f887286bbac04f5abb8c91b5e2e747b99ea932c5abf17135R408-R410

pulpcore/content/handler.py
bmbouter: I agree

pulpcore/content/handler.py
bmbouter: +1 to this

CHANGES/8760.feature
```diff
@@ -0,0 +1 @@
+Content app auto-distributes latest publication if distribution's `repository` field is set
```
bmbouter, line 1:
```suggestion
Content app auto-distributes latest publication if distribution's ``repository`` field is set
```

This goes into the Sphinx site which uses RST and that needs double backticks for literals escaping.

pulpcore/content/handler.py
```diff
@@ -419,13 +432,7 @@ async def _match_and_stream(self, path, request):
...
                         )
 
-        repo_version = getattr(distro, "repository_version", None)
-        repository = getattr(distro, "repository", None)
-
-        if repository or repo_version:
-            if repository:
-                repo_version = distro.repository.latest_version()
-
+        if repo_version:
```
bmbouter, line 435: :+1:

pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -295,6 +296,48 @@ def test_content_served(self):
         )
         self.addCleanup(self.client.delete, distribution["pulp_href"])
 
+        self.assert_content_served(repo, distribution)
+
+    def test_autopublish_content_served(self):
+        """Verify autopublished content is served over publication distributions."""
+        repo_body = {"autopublish": True, "manifest": "PULP_MANIFEST"}
+        repo = self.client.post(FILE_REPO_PATH, gen_repo(**repo_body))
```
bmbouter, line 304: Maybe not worth the rewrite here because of the need to merge this soon, but overall our plan is to have all new tests be bindings based tests and eventually remove all use of `self.client`.

bmbouter:
The code and tests here look good. I'll wait until the index is added and the tiny changelog fix.

bmbouter:
I also did some testing, which due to the tests wasn't really needed, but it did test well when I ran this code with pulp_file FWIW.

pulpcore/content/handler.py
dralley: I've backed off on this, we should investigate indexes but we should probably try to measure the impact rather than adding them speculatively.

gerrod3: Added task to keep track of this: https://pulp.plan.io/issues/8777

dralley (approved):

pulpcore/content/handler.py
daviddavis: I think we want to exclude publications where `complete=False` as well here.

bmbouter: Ah great point!

bmbouter (approved):
This looks good to me, but I want to also see the final from @daviddavis if that's ok.

daviddavis (approved):

PR #1335, opened 2021-05-17
Bring back some of the Distribution validation around "repository"

dralley:
Bring back some of the Distribution validation around "repository"

re: #8761
https://pulp.plan.io/issues/8761

### DIFF
--- a/pulpcore/app/serializers/publication.py
+++ b/pulpcore/app/serializers/publication.py
@@ -263,26 +263,39 @@ class DistributionSerializer(ModelSerializer, BasePathOverlapMixin):
     def validate(self, data):
         super().validate(data)
 
-        publication_in_data = "publication" in data
-        repository_version_in_data = "repository_version" in data
-        publication_in_instance = self.instance.publication if self.instance else None
-        repository_version_in_instance = self.instance.repository_version if self.instance else None
+        repository_provided = data.get("repository", None) or (
+            self.instance and self.instance.repository
+        )
+        repository_version_provided = data.get("repository_version", None) or (
+            self.instance and self.instance.repository_version
+        )
+        publication_provided = data.get("publication", None) or (
+            self.instance and self.instance.publication
+        )
 
-        if publication_in_data and repository_version_in_data:
-            error = True
-        elif publication_in_data and repository_version_in_instance:
-            error = True
-        elif publication_in_instance and repository_version_in_data:
-            error = True
-        else:
-            error = False
-
-        if error:
-            msg = _(
-                "Only one of the attributes 'publication' and 'repository_version' may be used "
-                "simultaneously."
+        if publication_provided and repository_version_provided:
+            raise serializers.ValidationError(
+                _(
+                    "Only one of the attributes 'publication' and 'repository_version' "
+                    "may be used simultaneously."
+                )
             )
-            raise serializers.ValidationError(msg)
+        elif repository_provided and repository_version_provided:
+            raise serializers.ValidationError(
+                _(
+                    "Only one of the attributes 'repository' and 'repository_version' "
+                    "may be used simultaneously."
+                )
+            )
+        # TODO: https://pulp.plan.io/issues/8762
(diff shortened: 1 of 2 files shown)


pulpcore/app/serializers/publication.py
```diff
@@ -263,26 +263,37 @@ class Meta:
...
-
-        if publication_in_data and repository_version_in_data:
-            error = True
-        elif publication_in_data and repository_version_in_instance:
-            error = True
-        elif publication_in_instance and repository_version_in_data:
-            error = True
-        else:
-            error = False
+        repository_provided = "repository" in data or (self.instance and self.instance.repository)
```
dralley, line 266: Should this be `data.get("repository", None) or ...`?  

The current logic will fail to validate even if the user is specifying `publication=null`

bmbouter: My personal opinion is the user shouldn't be specifying the key at all regardless of its value.

dralley: I'm thinking about the case when they want to change it from one type of distribution to another, they are forced to do two steps (unset, and then set) rather than unsetting one and setting the other in the same call.

mdellweg: Are you referring to the extra update calls needed here https://github.com/pulp/pulp-cli/blob/develop/pulpcore/cli/ansible/distribution.py#L134 ?
I agree, that just providing the new state in the update call should be enough.

goosemania:
Please file an issue for this change.

pulpcore/app/serializers/publication.py
dralley: Yes, that looks like a workaround for this (the same issue existed before).

bmbouter: I see. Without thinking about how to implement this, I believe users should use one or the other but not never both. Also it *would* be valuable to allow a user to "switch" in one call, but by doing so they will need to manually unset the other if using a PATCH call. Does this sound right @dralley ?

dralley: @bmbouter I think trying to restrict what parameters are *used* in the call (even if they're explicitly set null) feels kind of un-RESTful, and we should probably just let users say `publication=null repository=null repository_version=......` and have it work.  But of course if they try to actively set more than one, it should fail.

```diff
diff --git a/pulpcore/app/serializers/publication.py b/pulpcore/app/serializers/publication.py
index 1eba1df05..d4259ce00 100644
--- a/pulpcore/app/serializers/publication.py
+++ b/pulpcore/app/serializers/publication.py
@@ -263,11 +263,13 @@ class DistributionSerializer(ModelSerializer, BasePathOverlapMixin):
     def validate(self, data):
         super().validate(data)
 
-        repository_provided = "repository" in data or (self.instance and self.instance.repository)
-        repository_version_provided = "repository_version" in data or (
+        repository_provided = data.get("repository", None) or (
+            self.instance and self.instance.repository
+        )
+        repository_version_provided = data.get("repository_version", None) or (
             self.instance and self.instance.repository_version
         )
-        publication_provided = "publication" in data or (
+        publication_provided = data.get("publication", None) or (
             self.instance and self.instance.publication
         )
 ```

bmbouter: As long as there is no possibility that two could be set then +1. @dralley is that what you are thinking?

dralley: Yes, that's what I'd like to do

bmbouter:
@dralley this PR looks great. I'm about to LGTM it, but I wanted to ask, is there any need for us to write a migration along with this? We could merge this and someone else could do it, but I wanted to see if it was even necessary. Wdyt?

dralley:
I don't think so, not for this.  When we add the database constraints, yes.

bmbouter (approved):
Thanks @dralley !

dkliban (approved):

PR #1336, opened 2021-05-18
Removes click as a direct dependency

bmbouter:
`pulpcore` does not use `click` as a direct dependency, so we should no
longer declare it.

closes #8772

### DIFF
--- a/CHANGES/8772.misc
+++ b/CHANGES/8772.misc
@@ -1,0 +1,1 @@
+Removed click as a direct dependency since ``pulpcore`` does not directly use it.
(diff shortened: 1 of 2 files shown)


dralley (changes requested):
@bmbouter what about https://github.com/pulp/pulpcore/blob/6af4885832b2fb473d8e455a2a3602a8e63765b7/pulpcore/tasking/entrypoint.py ?

bmbouter:
Oh you right, we can't do this at all. I was thinking this all came in 3.14, but no here it is! I'm closing this and reopeneing https://github.com/pulp/pulpcore/pull/1321 instead. Thanks @dralley !

PR #1337, opened 2021-05-18
Revert "[noissue]: Update click requirement from ~=7.1.2 to ~=8.0.0"

bmbouter:
This reverts commit 935fac880e5f0b70e8cccc75fa22b614da96ef47.

Additionally since pulpcore is compatible with both 7.y and 8.y versions
of click we can just ensure we don't break when click 9.0 comes out.

[noissue]

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ aiohttp~=3.7.4
 aiodns~=3.0.0
 aiofiles==0.7.0
 backoff~=1.10.0
-click~=8.0.0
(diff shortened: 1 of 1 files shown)


fao89 (approved):

dralley (approved):

dkliban (approved):

mdellweg (approved):

goosemania:
There are 3 acks, I'm merging in order for a plugin CI to pass.

PR #1349, opened 2021-05-19
Fixes autodoc build errors

bmbouter:
Autodoc was failing to find several python modules, but these failures
didn't cause formal docs build errors. In cases where code was moved,
this PR fixes the paths to resolve the error. In cases where code was
removed, it removes those autodoc entries.

closes #8784

### DIFF
--- a/CHANGES/8784.doc
+++ b/CHANGES/8784.doc
@@ -1,0 +1,1 @@
+Fixed docs build errors reported by autodoc.
--- a/docs/contributing/platform-api/app/auth.rst
+++ b/docs/contributing/platform-api/app/auth.rst
@@ -1,4 +1,4 @@
 pulp.app.auth
-=================
+=============
 
-.. automodule:: pulpcore.app.auth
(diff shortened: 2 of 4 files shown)


daviddavis (approved):
LGTM. Is there a way to check for these errors in our CI?

mdellweg (approved):

goosemania (approved):

bmbouter:
> LGTM. Is there a way to check for these errors in our CI?

I looking into this for a few minutes, but I'm not sure. We already have [Sphinx enabled to treat warnings as errors](https://github.com/pulp/pulpcore/blob/master/docs/Makefile#L5). I tried making that [`-Wn` to turn on nitpick mode](https://www.sphinx-doc.org/en/master/man/sphinx-build.html) which turned up other issues but did not identify the failed. Maybe those failures are preventing us from seeing the autodoc failures since it stops on the first one?

To address that, I filed this.  https://pulp.plan.io/issues/8790

PR #1350, opened 2021-05-20
Add troubleshooting section to documentation

mdellweg:
Currently this section contains tips to recover a jammed tasking system.

fixes #8774

### DIFF
--- a/docs/components.rst
+++ b/docs/components.rst
@@ -85,6 +85,7 @@ Resource Manager
    the same repo should not run in parallel. Generally tasks are serialized at the "repo" level, so
    if you start *N* workers you can process *N* repo sync/modify/publish operations concurrently.
 
+In case your tasking system get's jammed, there is a guide to help :ref:debugging_tasks.
 
 Static Content
 --------------
--- a/docs/workflows/troubleshooting.rst
+++ b/docs/workflows/troubleshooting.rst
@@ -1,0 +1,113 @@
+Troubleshooting
+===============
+
+.. _debugging_tasks:
+
+Debugging Tasks
+---------------
+
+In case your system gets stuck in the processing of pulp tasks, you might want to debug the tasking system.
+
+Please always consider that your system might be in the process of dealing with long running tasks, and other tasks are rightfully waiting on their completion.
+
+Query tasks with the CLI
+------------------------
+
+How many tasks are waiting?
+
+.. code-block:: bash
+
+    pulp task list --state=waiting | jq 'length'
+
+Is anybody running?
+
+.. code-block:: bash
+
+    pulp task list --state=running | jq 'length'
+
+How many have failed?
+
+.. code-block:: bash
+
+    pulp task list --state=failed | jq 'length'
+
+Retrieve the HREF's of running tasks:
+
+.. code-block:: bash
+
+    pulp task list --state=running | jq 'map({.name, .pulp_href})'
+    # Save the HREF of the 3rd (counting starts at zero)
+    TASK_HREF=$(pulp task list --state=running | jq -r 'map(.pulp_href)[3]')
+
+Show the state of a particular task:
+
+.. code-block:: bash
+
+    pulp task show --href "$TASK_HREF"
+
+Cancel a running task:
+
+.. code-block:: bash
+
+    # warning canceling tasks may break higher level workflows
+    pulp task cancel --href "$TASK_HREF"
+
+Deeper Inspection
+-----------------
+
+This requires root access on the pulp server.
+
+Gather status information about the underlying ``rq`` service:
+
+.. code-block:: bash
+
+    rq info
+
+A sample output can look like this:
+
+.. code-block:: bash
+
+    resource-manager | 0
+    84459@pulp3.example.com | 0
+    72284@pulp3.example.com | 0
+    84635@pulp3.example.com | 0
+    7 queues, 0 jobs total
+
+    resource-manager (None None): ?
+    84635@pulp3.example.com (None None): ?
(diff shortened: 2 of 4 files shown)


docs/workflows/troubleshooting.rst
```diff
@@ -0,0 +1,111 @@
+Troubleshooting
+===============
+
+Debugging Tasks
+---------------
+
+In case, your system gets stuck in the processing of pulp tasks, you might want to debug the tasking system.
```
ggainey, line 7: No comma after case

docs/workflows/troubleshooting.rst
```diff
@@ -0,0 +1,111 @@
...
+
+.. code-block:: bash
+
+    pulp task show --href "$TASK_HREF"
+
+Cancel a running task:
+
+.. code-block:: bash
+
+    # possibly dangerous action
```
ggainey, line 50: Is 'cancel' dangerous? If so - let's add some examples of why this might be suboptimal?

docs/workflows/troubleshooting.rst
```diff
@@ -0,0 +1,111 @@
...
+    84635@pulp3.example.com (None None): ?
+    2 workers, 7 queues
+
+    Updated: 2020-03-17 14:08:41.961447
+
+There should at least be one queue for every worker.
+The numbers behind the queues show the count of queued rq jobs.
+If jobs pile up on the resource-manager queue, it's a sign that something might have got stuck.
+In case you cannot resolve the issue, be sure to include the dump of this command with the issue description.
+
```
ggainey, line 84: :+1:

docs/workflows/troubleshooting.rst
```diff
@@ -0,0 +1,111 @@
...
+Find and Remove Stuck ``ReservedResource``
+------------------------------------------
+
+This requires root access on the pulp server.
+
+In case, the automatically scheduled resource cleanup job of a task was not properly executed, some resources can be stuck in a locked state.
```
ggainey, line 90: No comma post-case.

docs/workflows/troubleshooting.rst
```diff
@@ -0,0 +1,111 @@
...
+They need to be removed by hand.
+
+The following operations are meant to be executed inside Pulp's Django shell `pulpcore-manager shell`.
+Be careful, with running those commands, as they are executed without any protection in the python context of the Pulp application.
```
ggainey, line 94: And lose the comma after careful.

docs/workflows/troubleshooting.rst
mdellweg: Well it's probably the least dangerous of the dangerous commands.
And it's danger is, that someone wanted this task to be performed, and yet we hinder it to complete.
It is the only not just investigating command here. How would you put it?

ggainey (approved):
This all looks good to me!

docs/workflows/troubleshooting.rst
```diff
@@ -0,0 +1,111 @@
...
+This requires root access on the pulp server.
+
+In case the automatically scheduled resource cleanup job of a task was not properly executed, some resources can be stuck in a locked state.
+They need to be removed by hand.
+
+The following operations are meant to be executed inside Pulp's Django shell `pulpcore-manager shell`.
+Be careful with running those commands, as they are executed without any protection in the python context of the Pulp application
+If you choose to go down that road, you are on your own.
+
+.. code-block:: bash
```
bmbouter, line 97: Is it python instead?

bmbouter: Maybe bash is fine, it is run in `pulpcore-manager shell`

bmbouter (approved):
This all looks good to me. Thanks @mdellweg !

docs/workflows/troubleshooting.rst
ggainey: > Is it python instead?

Oh, good question! It's run in pulpcore-manager, the code is python. Changing this would maybe make the output a little prettier? Worth a test I think.

mdellweg: Hmm, i thought i get a compilation error with it. But it turns out, that i get the exact same error every second time i run `make html`. This is so weird. It's not even the same error every time...

docs/workflows/troubleshooting.rst
```diff
@@ -0,0 +1,111 @@
...
+===============
+
+Debugging Tasks
+---------------
+
+In case your system gets stuck in the processing of pulp tasks, you might want to debug the tasking system.
+
+Please always consider that your system might be in the process of dealing with long running tasks, and other tasks are rightfully waiting on their completion.
+
+Query tasks through the REST API
```
ipanova, line 11: query tasks with the CLI?

docs/workflows/troubleshooting.rst
```diff
@@ -0,0 +1,111 @@
+Troubleshooting
+===============
+
+Debugging Tasks
```
ipanova, line 4: does it make sense to add a link to this section? https://github.com/pulp/pulpcore/blob/master/docs/components.rst#L61

docs/components.rst
```diff
@@ -85,6 +85,7 @@ Resource Manager
    the same repo should not run in parallel. Generally tasks are serialized at the "repo" level, so
    if you start *N* workers you can process *N* repo sync/modify/publish operations concurrently.
 
+In case your tasking system get's jammed, there is a guide that to help :ref:debugging_tasks.
```
ipanova, line 88: :+1:

ipanova (approved):

PR #1353, opened 2021-05-24
Fixed retained version bug where content wasn't updated

daviddavis:
Fixed a bug in the retained versions code where content wasn't being
properly moved to newer repo versions when old versions were cleaned up.

fixes #8793

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -11,7 +11,7 @@ from asyncio_throttle import Throttler
 from django.core.validators import MinValueValidator
 from django.db import models, transaction
 from django.urls import reverse
-from django_lifecycle import AFTER_SAVE, hook
+from django_lifecycle import AFTER_UPDATE, hook
 
 from django.contrib.postgres.fields import JSONField
 
@@ -193,7 +193,7 @@ class Repository(MasterModel):
         """
         return Artifact.objects.filter(content__pk__in=version.content)
 
-    @hook(AFTER_SAVE)
+    @hook(AFTER_UPDATE, when="retained_versions", has_changed=True)
     def cleanup_old_versions(self):
         """Cleanup old repository versions based on retained_versions."""
         if self.retained_versions:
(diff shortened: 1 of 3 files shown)


pulpcore/app/models/repository.py
```diff
@@ -193,7 +193,7 @@ def artifacts_for_version(version):
         """
         return Artifact.objects.filter(content__pk__in=version.content)
 
-    @hook(AFTER_SAVE)
+    @hook(AFTER_UPDATE, when="retained_versions", has_changed=True)
```
mdellweg, line 196: Nice!

mdellweg (approved):

pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
```diff
@@ -1140,7 +1140,13 @@ def test_retained_versions(self):
 
         versions = self.version_api.list(file_file_repository_href=self.repo.pulp_href).results
         self.assertEqual(len(versions), 1)
-        self.assertEqual(self.repo.latest_version_href.split("/")[-2], "3")
+
+        latest_version = self.version_api.read(
+            file_file_repository_version_href=self.repo.latest_version_href
+        )
+        self.assertEqual(latest_version.number, 3)
+        self.assertEqual(latest_version.content_summary.present["file.file"]["count"], 3)
```
ipanova, line 1148: :+1:

ipanova (approved):

pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
```diff
@@ -1155,7 +1161,15 @@ def test_retained_versions_on_update(self):
...
+
+        # TODO: the count here should be 3 https://pulp.plan.io/issues/8798
+        self.assertEqual(latest_version.content_summary.added["file.file"]["count"], 1)
```
mdellweg, line 1172: I thought 1 _was_ correct here, because we retain 2 versions.
In the test setup we create a bunch of versions, each adding one content unit on top of the last one. And if we retain 2 versions, than the second to last version should add and contain N-1 items, while the last should contain N and add 1.
And that's what i am seeing here.

daviddavis: Yup, the previous version should have an added count of 2. Good catch.

PR #1354, opened 2021-05-24
Release 3.11.2

bmbouter:


dralley (approved):

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.11.2.dev
+current_version = 3.12.0.dev
```
daviddavis, line 2: This should be 3.11.3.dev

bmbouter: Agreed. I'm going to fix this in the release script and then have it used here.

mdellweg (dismissed):

.bumpversion.cfg
bmbouter: Actually I was doing it wrong I pushed again only this time is was created using `patch` instead of `minor`.

daviddavis (approved):

mdellweg (approved):

PR #1355, opened 2021-05-24
Moves Auth to its own section in the docs

bmbouter:
This commit reorganizes the basic auth docs and moves them to their own
section.

closes #8800


docs/authentication/basic.rst
```diff
@@ -0,0 +1,34 @@
...
+user submitted header against an internal database of users. If the username and password match, the
+request is considered authenticated as that username.
+
+Below is an example of a Basic Authentication header::
+
+    Authorization: Basic YWRtaW46cGFzc3dvcmQ=
+
+You can set this header on an `httpie <https://httpie.org/>`_ command as follows::
+
+    http :80/pulp/api/v3/tasks/ Authorization:"Basic YWRtaW46cGFzc3dvcmQ="
```
mdellweg, line 16: This should be `http --auth admin:password ...`

mdellweg (approved):
I left a comment, but maybe you want to make this commit "move only" and improve the docs in another commit.

docs/authentication/basic.rst
ipanova: both ways work, you whether provide in plain text with the `--auth` or specify the header with base64 encoded creds
```
$ http :24817/pulp/api/v3/tasks/ Authorization:"Basic aW5hOmdvb2RwYXNzd29yZA==" --ignore-netrc
HTTP/1.1 403 Forbidden
Allow: GET, HEAD, OPTIONS
Connection: close
Content-Length: 39
Date: Tue, 25 May 2021 11:05:13 GMT
Server: gunicorn/20.0.4
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "detail": "Invalid username/password."
``

File: docs/authentication/index.rst [hunk 2]
```diff
@@ -0,0 +1,44 @@
...
+By default, Pulp supports Basic and Session authentication. The Basic Authentication checks the
+username and password against the internal users database.
+
+.. note::
+    This authentication is only for the REST API. Client's fetching binary data have their identity
+    verified and authorization checked using a :term:`ContentGuard`.
+
+.. warning::
+    Until Role-Based Access Control is added to Pulp, REST API is not safe for multi-user use.
+    Sensitive credentials can be read by any user, e.g. ``Remote.password``, ``Remote.client_key``.
```
ipanova, line 23: I don't think this the case anymore, these fields are write_only.

docs/authentication/basic.rst
mdellweg: I know, it works, i just don't see it good practice in user facing docs.
We also do not specify, that the crazy string should be b64 of "username:password".

docs/authentication/index.rst
bmbouter: Great point. I'll fix this.

bmbouter: I'm also removing from 2 additional places I found.

docs/authentication/basic.rst
bmbouter: I also added some clarification about this in my latest revision.

docs/authentication/basic.rst
```diff
@@ -0,0 +1,38 @@
...
+
+Below is an example of a Basic Authentication header for a username ``admin`` and password
+``password``.::
+
+    Authorization: Basic YWRtaW46cGFzc3dvcmQ=
+
+You can set this header on an `httpie <https://httpie.org/>`_ command using the ``--auth`` option as
+follows::
+
+     http --auth admin:password ...
```
mdellweg, line 20: :+1:

bmbouter:
@ipanova what do you think about these changes?

ipanova (approved):
Looks good, thank you!

PR #1356, opened 2021-05-25
Use Redis to cache responses from content app

gerrod3:
fixes: #8805
https://pulp.plan.io/issues/8805

- [x] Clean up/ Re-do the Cache API
- [x] Finish every TODO commented
- [x] Add tests
- [x] Add changelog
- [x] Add docs explaining new cache
- [ ] (Optional) Add decorator for Django views

Still a heavily WIP PR and lots of work to be done, but it does work and pretty nicely too!  I didn't put much thought into the cache api design(except for the fact that I wanted it to be a decorator added to any handler/view function) so it is a bit of a mess and needs refactoring.  ~The one big problem I ran into was that the content app should use async methods when talking to redis, but all of the other Django code requires non-async methods to talk to redis.  This made designing a shared base class non-trivial.~ Another problem is that aiohttp requests/responses are different from Django requests/responses so an all inclusive cache api would be difficult(but not impossible) to create and would probably need to be in separate objects.  

~*Edit* Since this will go onto pulpcore 3.14 which is upgrading to Django 3.2 which has `asgiref` as a dependency, I have solved the async/sync problem using the `async_to_sync` functions.~

### DIFF
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -189,6 +189,33 @@ CONTENT_APP_TTL
    Defaults to ``30`` seconds.
 
 
+.. _pulp-cache:
+
+CACHE_ENABLED
+^^^^^^^^^^^^^^^^^^
+
+   .. note:: This feature is provided as a tech-preview
+
+   Store cached responses from the content app into Redis. This setting improves the performance
+   of the content app under heavy load for similar requests. Defaults to ``True``.
+
+   .. note::
+     The entire response is not stored in the cache. Only the location of the file needed to
+     recreate the response is stored. This reduces database queries and allows for many
+     responses to be stored inside the cache.
+
+CACHE_SETTINGS
+^^^^^^^^^^^^^
+
+   Dictionary with tunable settings for the cache:
+
+   * ``EXPIRES_TTL`` - Number of seconds entries should stay in the cache before expiring.
+
+   .. note::
+     Set to ``None`` to have entries not expire.
+     Content app responses are always invalidated when the backing distribution is updated.
+
+
 .. _worker-ttl:
 
 WORKER_TTL
--- a/pulpcore/app/models/publication.py
+++ b/pulpcore/app/models/publication.py
@@ -1,10 +1,13 @@
 from django.db import IntegrityError, models, transaction
+from django_lifecycle import hook, AFTER_UPDATE, BEFORE_DELETE
 
 from .base import MasterModel, BaseModel
 from .content import Artifact, Content, ContentArtifact
 from .repository import Remote, Repository, RepositoryVersion
 from .task import CreatedResource
 from pulpcore.app.files import PulpTemporaryUploadedFile
+from pulpcore.cache import Cache
+from dynaconf import settings
 
 
 class PublicationQuerySet(models.QuerySet):
@@ -122,6 +125,19 @@ class Publication(MasterModel):
             Deletes the Task.created_resource when complete is False.
         """
         with transaction.atomic():
+            if settings.CACHE_ENABLED:
+                # Find any publications being served directly
+                base_paths = self.distribution_set.values_list("base_path", flat=True)
+                # Find any publications being served indirectly by auto-distribute feature
+                versions = self.repository.versions.all()
+                pubs = Publication.objects.filter(repository_version__in=versions, complete=True)
+                publication = pubs.latest("repository_version", "-pulp_created")
+                if self.pk == publication.pk:
+                    base_paths |= self.repository.distributions.values_list("base_path", flat=True)
+                # Invalidate cache for all distributions serving this publication
+                if base_paths:
+                    Cache().delete(base_key=base_paths)
+
             CreatedResource.objects.filter(object_id=self.pk).delete()
             super().delete(**kwargs)
 
@@ -282,6 +298,13 @@ class ContentGuard(MasterModel):
         """
         raise NotImplementedError()
 
+    @hook(BEFORE_DELETE)
+    def invalidate_cache(self):
+        if settings.CACHE_ENABLED:
+            base_paths = self.distribution_set.values_list("base_path", flat=True)
+            if base_paths:
+                Cache().delete(base_key=base_paths)
+
 
 class BaseDistribution(MasterModel):
     """
@@ -383,3 +406,22 @@ class Distribution(MasterModel):
             Set of strings for the extra entries in rel_path
         """
         return set()
+
+    @hook(BEFORE_DELETE)
+    @hook(
+        AFTER_UPDATE,
+        when_any=[
+            "base_path",
+            "content_guard",
+            "publication",
+            "remote",
+            "repository",
+            "repository_version",
+        ],
+        has_changed=True,
+    )
+    def invalidate_cache(self):
+        """Invalidates the cache if enabled."""
+        if settings.CACHE_ENABLED:
+            Cache().delete(base_key=self.base_path)
+            # Can also preload cache here possibly
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -8,10 +8,11 @@ import logging
 
 import django
 from asyncio_throttle import Throttler
+from dynaconf import settings
 from django.core.validators import MinValueValidator
 from django.db import models, transaction
 from django.urls import reverse
-from django_lifecycle import AFTER_UPDATE, hook
+from django_lifecycle import AFTER_UPDATE, BEFORE_DELETE, hook
 
 from django.contrib.postgres.fields import JSONField
 
@@ -19,6 +20,8 @@ from pulpcore.app.util import batch_qs, get_view_name_for_model
 from pulpcore.constants import ALL_KNOWN_CONTENT_CHECKSUMS
 from pulpcore.download.factory import DownloaderFactory
 from pulpcore.exceptions import ResourceImmutableError
+
+from pulpcore.cache import Cache
 
 from .base import MasterModel, BaseModel
 from .content import Artifact, Content
@@ -133,6 +136,8 @@ class Repository(MasterModel):
                 resource = CreatedResource(content_object=version)
                 resource.save()
 
+            self.invalidate_cache()
+
             return version
 
     def finalize_new_version(self, new_version):
@@ -206,6 +211,17 @@ class Repository(MasterModel):
                     )
                 )
                 version.delete()
+
+    @hook(BEFORE_DELETE)
+    def invalidate_cache(self):
+        """Invalidates the cache if repository is present."""
+        if settings.CACHE_ENABLED:
+            distributions = self.distributions.all()
+            if distributions.exists():
+                base_paths = distributions.values_list("base_path", flat=True)
+                if base_paths:
+                    Cache().delete(base_key=base_paths)
+                # Could do preloading here for immediate artifacts with artifacts_for_version
 
 
 class Remote(MasterModel):
@@ -427,6 +443,14 @@ class Remote(MasterModel):
             Class: The Class of the content type that should be available at the relative path.
         """
         raise NotImplementedError()
+
+    @hook(BEFORE_DELETE)
+    def invalidate_cache(self):
+        """Invalidates the cache if remote is present."""
+        if settings.CACHE_ENABLED:
+            base_paths = self.distribution_set.values_list("base_path", flat=True)
+            if base_paths:
+                Cache().delete(base_key=base_paths)
 
     class Meta:
         default_related_name = "remotes"
@@ -865,6 +889,11 @@ class RepositoryVersion(BaseModel):
         Deletion of a complete RepositoryVersion should be done in a RQ Job.
         """
         if self.complete:
+            if settings.CACHE_ENABLED:
+                base_paths = self.distribution_set.values_list("base_path", flat=True)
+                if base_paths:
+                    Cache().delete(base_key=base_paths)
+
             repo_relations = RepositoryContent.objects.filter(repository=self.repository)
             try:
                 next_version = self.next()
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -239,6 +239,12 @@ USE_NEW_WORKER_TYPE = True
 
 PROFILE_STAGES_API = False
 
+# https://docs.pulpproject.org/pulpcore/configuration/settings.html#pulp-cache
+CACHE_ENABLED = True
+CACHE_SETTINGS = {
+    "EXPIRES_TTL": 86400,
+}
+
 SPECTACULAR_SETTINGS = {
     "SERVE_URLCONF": ROOT_URLCONF,
     "DEFAULT_GENERATOR_CLASS": "pulpcore.openapi.PulpSchemaGenerator",
--- a/pulpcore/cache/cache.py
+++ b/pulpcore/cache/cache.py
@@ -1,0 +1,234 @@
+import enum
+import json
+
+from aiohttp.web import FileResponse, Response, HTTPSuccessful, Request
+from aiohttp.web_exceptions import HTTPFound
+
+from pulpcore.app.settings import settings
+from pulpcore.tasking.connection import get_redis_connection, get_async_redis_connection
+
+DEFAULT_EXPIRES_TTL = settings.CACHE_SETTINGS["EXPIRES_TTL"]
+
+
+class CacheKeys(enum.Enum):
+    """Available keys to construct the index key for cache entry."""
+
+    path = "path"
+    host = "host"
+    method = "method"
+
+
+class Cache:
+    """Base class for Pulp's cache"""
+
+    default_base_key = "PULP_CACHE"
+    default_expires_ttl = DEFAULT_EXPIRES_TTL
+
+    def __init__(self):
+        """Creates synchronous cache instance"""
+        self.redis = get_redis_connection()
+
+    def get(self, key, base_key=None):
+        """Gets cached entry of key"""
+        base_key = base_key or self.default_base_key
+        if key is None:
+            return self.redis.hgetall(base_key)
+        return self.redis.hget(base_key, key)
+
+    def set(self, key, value, expires=None, base_key=None):
+        """Sets the cached entry at key"""
+        base_key = base_key or self.default_base_key
+        ret = self.redis.hset(base_key, key, value)
+        if expires:
+            self.redis.expire(base_key, expires)
+        return ret
+
+    def exists(self, key=None, base_key=None):
+        """Checks if cached entries exist"""
+        base_key = base_key or self.default_base_key
+        if key:
+            return self.redis.hexists(base_key, key)
+        else:
+            if isinstance(base_key, str):
+                base_key = [base_key]
+            return self.redis.exists(*base_key)
+
+    def delete(self, key=None, base_key=None):
+        """
+        Deletes the cached entry at base_key: key
+
+        If only base_key is supplied then delete all entries under that base_key
+        key can be a list to delete multiple entries under a base_key
+        base_key can be a list to delete multiple sets of entries
+        key and base_key should not both be lists
+        """
+        base_key = base_key or self.default_base_key
+        if key:
+            return self.redis.hdel(base_key, key)
+        if isinstance(base_key, str):
+            base_key = [base_key]
+        return self.redis.delete(*base_key)
+
+
+class AsyncCache:
+    """Base class for asynchronous Pulp Cache"""
+
+    default_base_key = "PULP_CACHE"
+    default_expires_ttl = DEFAULT_EXPIRES_TTL
+
+    def __init__(self):
+        """Creates asynchronous cache instance"""
+        self.redis = get_async_redis_connection()
+
+    async def get(self, key, base_key=None):
+        """Gets cached entry of key"""
+        base_key = base_key or self.default_base_key
+        if key is None:
+            return await self.redis.hgetall(base_key)
+        return await self.redis.hget(base_key, key)
+
+    async def set(self, key, value, expires=None, base_key=None):
+        """Sets the cached entry at key"""
+        base_key = base_key or self.default_base_key
+        ret = await self.redis.hset(base_key, key, value)
+        if expires:
+            await self.redis.expire(base_key, expires)
+        return ret
+
+    async def exists(self, key=None, base_key=None):
+        """Checks if cached entries exist"""
+        base_key = base_key or self.default_base_key
+        if key:
+            return await self.redis.hexists(base_key, key)
+        else:
+            if isinstance(base_key, str):
+                base_key = [base_key]
+            return await self.redis.exists(*base_key)
+
+    async def delete(self, key=None, base_key=None):
+        """
+        Deletes the cached entry at base_key: key
+
+        If only base_key is supplied then delete all entries under that base_key
+        key can be a list to delete multiple entries under a base_key
+        base_key can be a list to delete multiple sets of entries
+        key and base_key should not both be lists
+        """
+        base_key = base_key or self.default_base_key
+        if key:
+            return await self.redis.hdel(base_key, key)
+        if isinstance(base_key, str):
+            base_key = [base_key]
+        return await self.redis.delete(*base_key)
+
+
+class ContentCache(AsyncCache):
+    """Cache object meant to be used for the content app"""
+
+    RESPONSE_TYPES = {
+        "FileResponse": FileResponse,
+        "Response": Response,
+        "Redirect": HTTPFound,
+    }
+
+    def __init__(self, base_key=None, expires_ttl=None, keys=None, auth=None):
+        """
+        Initiates a cache instance to be used for dealing with an aiohttp server
+
+        Args:
+            base_key: a string to group entries under, defaults to DEFAULT_BASE_KEY,
+                      can be a callable taking the request and cache instance as arguments
+            expires_ttl: length in seconds entries should live in the cache, EXPIRES_TTL is default
+            keys: a list of CacheKeys to use for key creation upon entry placement, path is default
+            auth: a callable to check authorization of the request; takes the request, cache
+                  instance, and base_key as arguments.
+        """
+        super().__init__()
+        self.default_base_key = base_key or self.default_base_key
+        self.keys = keys or (CacheKeys.path,)
+        self.default_expires_ttl = expires_ttl or self.default_expires_ttl
+        self.auth = auth
+
+    def __call__(self, func):
+        """Decorator magic call to make the handler cached"""
+        if not settings.CACHE_ENABLED:
+            return func
+
+        async def cached_function(*args):
+            request = self.get_request_from_args(args)
+            bk = self.default_base_key
+            if callable(self.default_base_key):
+                bk = await self.default_base_key(request, self)
+            if self.auth:
+                await self.auth(request, self, bk)
+            key = self.make_key(request)
+            # Check cache
+            response = await self.make_response(key, bk)
+            if response is None:
+                # Cache miss, create new entry
+                response = await self.make_entry(key, bk, func, args, self.default_expires_ttl)
+            return response
+
+        return cached_function
+
+    def get_request_from_args(self, args):
+        """Finds the request object from list of args"""
+        for arg in args:
+            if isinstance(arg, Request):
+                return arg
+
+    async def make_response(self, key, base_key):
+        """Tries to find the cached entry and turn it into a proper response"""
+        entry = await self.get(key, base_key)
+        if not entry:
+            return None
+        entry = json.loads(entry)
+        response_type = entry.pop("type", None)
+        if not response_type or response_type not in self.RESPONSE_TYPES:
+            # Bad entry, delete from cache
+            self.delete(key, base_key)
+            return None
+        response = self.RESPONSE_TYPES[response_type](**entry)
+        response.headers.update({"X-PULP-CACHE": "HIT"})
+        return response
+
+    async def make_entry(self, key, base_key, handler, args, expires=86400):
+        """Gets the response for the request and try to turn it into a cacheable entry"""
+        try:
+            response = await handler(*args)
+        except (HTTPSuccessful, HTTPFound) as e:
+            response = e
+
+        entry = {"headers": dict(response.headers), "status": response.status}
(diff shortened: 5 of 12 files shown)


pulpcore/app/models/repository.py
```diff
@@ -70,7 +71,15 @@ def on_new_version(self, version):
         Args:
             version: The new repository version.
         """
-        pass
+        # Not sure If i should put this here or in new_version
+        # Invalidate Caches, could also do pre-loading here as-well
+        caching = True  # TODO change to setting's value
+        distributions = self.distributions.all()
```
dralley, line 77: This is a way to disable caching, correct?

We definitely need one, as a workaround for customers in case there's a critical bug in the caching logic, and also as an aide to debugging and testing.

gerrod3: Yes it'll be a setting's value.  I think there might be a couple cache-specific settings values that will be needed for user-tuning but this will definitely be one.

dralley:
@gerrod3 I did some performance testing and the results were pretty good!  It does take a while to warm up, during which time the results aren't quite as good, but once it does:

* It hovers between 2 and 2.5x as many requests handled per second - usually around 500.
* 95th percentile latency of around 110ms, median 20ms
* Improved CPU usage (50-60% on all cores, rather than 85+%)

It's actually beating Nginx serving static files!  Granted, it's the default Nginx configuration, and Nginx is only using 15% of the CPU, but that is still quite impressive.

Thank you so much for this work!  I'm very excited to see where we end up with this work.

pulpcore/app/models/publication.py
```diff
@@ -352,13 +370,13 @@ class Distribution(MasterModel):
     name = models.TextField(db_index=True, unique=True)
     base_path = models.TextField(unique=True)
 
-    content_guard = models.ForeignKey(ContentGuard, null=True, on_delete=models.SET_NULL)
-    publication = models.ForeignKey(Publication, null=True, on_delete=models.SET_NULL)
-    remote = models.ForeignKey(Remote, null=True, on_delete=models.SET_NULL)
+    content_guard = models.ForeignKey(ContentGuard, null=True, on_delete=set_null)
```
dralley, line 373: This feels like leaning really hard on an implementation detail.  @bmbouter Do you think signals ([pre_delete](https://docs.djangoproject.com/en/3.2/ref/signals/#pre-delete) specifically) would work better here?

Possibly on line 126 up above, as well.

gerrod3: Yeah I'm going to need to change this to use signals, because changing the `on_delete` requires a migration.

docs/configuration/settings.rst
```diff
@@ -189,6 +189,35 @@ CONTENT_APP_TTL
...
+
+   .. note:: This feature is provided as a tech-preview
+
+   Store cached responses from the content app into Redis. This setting improves the performance
+   of the content app under heavy load for similar requests.
+
+   .. note::
+     The entire response is not stored in the cache. Only the location of the file needed to
+     recreate the response is stored. This reduces database queries and allows for many
+     responses to be stored inside the cache.
```
dralley, line 205: Not sure if this is worth mentioning but the file itself *is* cached, by the operating system.  It'll use any free RAM you have for caching files that are being frequently used and purge them as needed.  So we don't (and shouldn't) ever attempt to do that ourselves.

docs/configuration/settings.rst
```diff
@@ -189,6 +189,35 @@ CONTENT_APP_TTL
...
+
+PULP_SETTINGS
```
dralley, line 207: Should this be `CACHE_SETTINGS`?

pulpcore/cache/cache.py
```diff
@@ -0,0 +1,214 @@
...
+
+    path = "path"
+    host = "host"
+    method = "method"
+
+
+class Cache:
+    """Base class for Pulp's cache"""
+
+    def __init__(self, mode="sync", base_key=None, expires_ttl=None, keys=None, auth=None):
```
dralley, line 25: Let's have a docstring with descriptions of what these kwargs do.

pulpcore/cache/cache.py
```diff
@@ -0,0 +1,214 @@
...
+        self.keys = keys or (CacheKeys.path,)
+        self.default_base_key = base_key or DEFAULT_BASE_KEY
+        self.default_expires_ttl = expires_ttl or DEFAULT_EXPIRES_TTL
+        self.auth = auth
+        # Maybe make the config an argument too?
+        functions = ["get", "set", "exists", "delete", "clear"]
+        pre = "_" if mode != "async" else "_a_"
```
dralley, line 32: What's going on here?  Is it a compatibility hack?

pulpcore/app/models/publication.py
bmbouter: I'm not concerned about it being a migration. on_delete I think is probably a more direct way of handling this (to me).

gerrod3: Ok after testing out the hooks, I can either add a `BeforeDelete` hook to every `ForeignKey` model on the distribution to invalidate the cache or keep the new `on_delete` method and add the required migration. The migration would make this feature not backportable, but the feature does require a new dependency and I'm not sure if that is backportable either.

docs/configuration/settings.rst
gerrod3: Yes!

pulpcore/app/models/publication.py
dralley: @bmbouter my concern is that this is not a documented approach. People do it anyway (apparently, at least there are stack overflow posts about it) but it's not really something the developers want you to be doing.

pulpcore/cache/cache.py
gerrod3: Yeah a compatibility hack. The aioredis library I use has the exact same API as the redis library except everything is async. Before I was using the asgiref `async_to_sync` decorators to make the cache api synchronous when used as a context manager, but I had to create a new connection every time you made a redis call since the conversion caused the aioredis library to disconnect each time since the sync code runs the loop to completion.

Should I separate out the async and sync methods into two separate cache classes?

pulpcore/app/models/publication.py
bmbouter: I thought we were talking about this: https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.SET_NULL

dralley: @bmbouter But this isn't `django.db.models.SET_NULL` it's a custom wrapper around it.

The Django docs say "The possible values for on_delete are found in django.db.models" so this isn't something they expect people to be extending and customizing.

pulpcore/cache/cache.py
dralley: >Should I separate out the async and sync methods into two separate cache classes?

Possibly.  Do we need the synchronous version or can we stick to just aioredis?

gerrod3: We need the synchronous version so that we can invalidate the cache from django. I think it's easier that way then having plugin writers wrap the async methods into some synchronous context themselves.

dralley: Two classes seems a little better to me, yes

pulpcore/app/models/publication.py
bmbouter: I haven't looked closely at this apologies. I was recommending we use the official `django.db.models.SET_NULL`. If we can't then we should use another official thing. Note that `django-lifecycle` allows us to use that also as an alternative to django signals...

pulpcore/app/settings.py
```diff
@@ -239,6 +239,13 @@
 
 PROFILE_STAGES_API = False
 
+# https://docs.pulpproject.org/pulpcore/configuration/settings.html#pulp-cache
+PULP_CACHE_ENABLED = True
```
dralley, line 243: I'm not entirely sure how I feel about enabling it by default right away, I flip back and forth on it.  But the argument against is that we should give it a few weeks to "settle", and we have active communication with several users who are motivated to enable it and let us know immediately if there are issues.  Then we could enable it by default in 3.15 in a couple of weeks.

bmbouter: FWIW I'm in favor of enabling by default. If there is a problem I want to hear about it as early as possible, and I'm concerned without it being default users will not actually be testing which will be the same problem over again, only with 3.15. Also I think the tests are our proof of correctness which give me the confidence in it being a responsible choice.

gerrod3: I agree. Is there a way I can still have the tests for the feature run with the default setting being false?

pulpcore/app/models/publication.py
```diff
@@ -122,6 +125,16 @@ def delete(self, **kwargs):
             Deletes the Task.created_resource when complete is False.
         """
         with transaction.atomic():
+            if settings.PULP_CACHE_ENABLED:
```
dralley, line 128: I can mostly work out what's going on here, but would you mind commenting it to make it more clear?

pulpcore/app/models/publication.py
```diff
@@ -282,6 +295,13 @@ def permit(self, request):
         """
         raise NotImplementedError()
 
+    @hook(BEFORE_DELETE)
+    def invalidate_cache(self):
+        if settings.PULP_CACHE_ENABLED:
```
dralley, line 300: Same.  I will say that one thing which confuses me is using a list as the value of `base_key`.  Maybe `base_key` has a misleading name, but using a list as a "key" feels wrong, especially since the order of the list is not guaranteed.

pulpcore/app/models/publication.py
```diff
@@ -383,3 +403,22 @@ def content_handler_list_directory(self, rel_path):
             Set of strings for the extra entries in rel_path
         """
         return set()
+
+    @hook(BEFORE_DELETE)
+    @hook(
+        AFTER_UPDATE,
```
dralley, line 409: What are the semantics of these hooks and how do they differ from django signals?

pulpcore/app/models/repository.py
```diff
@@ -865,6 +889,11 @@ def delete(self, **kwargs):
         Deletion of a complete RepositoryVersion should be done in a RQ Job.
         """
         if self.complete:
+            if settings.PULP_CACHE_ENABLED:
+                base_paths = self.distribution_set.values_list("base_path", flat=True)
+                if base_paths:
+                    Cache().delete(base_key=base_paths)
```
dralley, line 895: Would it be possible to just call `self.invalidate_cache()`?  Does the hook decorator get in the way of that?

pulpcore/cache/cache.py
```diff
@@ -0,0 +1,233 @@
...
+        base_key = base_key or self.default_base_key
+        if key:
+            return await self.redis.hdel(base_key, key)
+        if isinstance(base_key, str):
+            base_key = [base_key]
+        return await self.redis.delete(*base_key)
+
+    async def clear(self):
+        """Clears the cache"""
+        return await self.redis.flushdb()  # Maybe this is dangerous?
```
dralley, line 121: Extremely dangerous if the user is on the old tasking system.

pulpcore/cache/cache.py
```diff
@@ -0,0 +1,233 @@
...
+        base_key = base_key or self.default_base_key
+        if key:
+            return self.redis.hdel(base_key, key)
+        if isinstance(base_key, str):
+            base_key = [base_key]
+        return self.redis.delete(*base_key)
+
+    def clear(self):
+        """Clears the cache"""
+        return self.redis.flushdb()  # Maybe this is dangerous?
```
dralley, line 70: Extremely dangerous if the user is on the old tasking system.

pulpcore/tasking/connection.py
```diff
@@ -1,12 +1,21 @@
...
 from rq.cli.helpers import get_redis_from_config
 
 from pulpcore.app.settings import settings
 
 _conn = None
+_a_conn = None
 
 
-def get_redis_connection():
+def get_redis_connection(mode="sync"):
```
dralley, line 10: This worries me a little.  I'd prefer to keep the async code separate e.g. `get_aioredis_connection` or `get_async_redis_connection`

pulpcore/app/models/repository.py
gerrod3: I've defined a `self.invalidate_cache()` for Repository and Remote models, but not for RepositoryVersion. The logic for Remotes and RepositoryVersions are the same, but slightly different for Repositories (mainly just the related_name for distributions)

pulpcore/cache/cache.py
dralley: If it's used only for unit tests it should be fine (and it looks like it is)

pulpcore/app/models/publication.py
gerrod3: In redis you can pass multiple values to a delete command. I agree `base_key` could be something different, but not sure what it can be named. You can also pass in a list for `key` as well and have multiple keys deleted under one `base_key`.

pulpcore/cache/cache.py
gerrod3: Yeah, I just have it as part of the api for now, but it's not called anywhere. Maybe once we drop old tasking system this call can be wrapped into a rest API call or used when restarting pulp itself.

pulpcore/app/models/publication.py
gerrod3: Functionally they are almost the exact same as django signals. These hooks work by decorating the `save` and `delete` handlers for lifecycle models, which all of our models are. The one difference I can see is that queryset deletes won't trigger the lifecycle hooks since the models `delete` isn't called, but signals do send a pre-post delete signal for queryset deletes.

pulpcore/cache/cache.py
```diff
@@ -0,0 +1,233 @@
...
+            base_key: a string to group entries under, defaults to DEFAULT_BASE_KEY,
+                      can be a callable taking the request and cache instance as arguments
+            expires_ttl: length in seconds entries should live in the cache, EXPIRES_TTL is default
+            keys: a list of CacheKeys to use for key creation upon entry placement, path is default
+            auth: a callable to check authorization of the request; takes the request, cache
+                  instance, and base_key as arguments.
+        """
+        super().__init__()
+        self.default_base_key = base_key or self.default_base_key
+        self.keys = keys or (CacheKeys.path,)
```
dralley, line 147: Nitpick, `tuple(Cachekeys.path)` might be clearer here.  Speaking for myself, my brain skips over the comma quite often for one-element tuples.

gerrod3: Pretty sure it would need to be `tuple([CacheKeys.path])` or else you would get a not iterable error. When I see parenthesis surrounding a variable I always think tuple.

pulpcore/app/models/repository.py
```diff
@@ -133,6 +136,8 @@ def new_version(self, base_version=None):
                 resource = CreatedResource(content_object=version)
                 resource.save()
 
+            self.invalidate_cache()
```
dralley, line 139: I presume we're invalidating the entire cache every time - that's probably a safe thing to do to start with, but in the future we probably want to be a little more selective.

docs/configuration/settings.rst
```diff
@@ -189,6 +189,35 @@ CONTENT_APP_TTL
...
+
+   Dictionary with two tunable settings for the cache:
+
+   * ``EXPIRES_TTL`` - Number of seconds entries should stay in the cache before expiring.
+
+   .. note::
+     Set to ``None`` to have entries not expire.
+     Content app responses are always invalidated when the backing distribution is updated.
+
+   * ``DEFAULT_BASE_KEY`` - The default base_key to group entries inside the Redis cache.
```
dralley, line 218: I'm not sure a user would understand what this means, it's a little too technical.

pulpcore/content/handler.py
```diff
@@ -141,6 +142,51 @@ def get_base_paths_blocking():
...
+        Args:
+            request (:class:`aiohttp.web.request`): The request from the client.
+            cached (:class:`CacheAiohttp`): The Pulp cache
+
+        Returns:
+            str: The base-path associated with this request
+        """
+        path = request.match_info["path"]
+        base_paths = cls._base_paths(path)
+        m_base_paths = [bp for i, bp in enumerate(base_paths) for _ in range(i + 1)]
```
dralley, line 159: I'm not sure I understand what is going on here. This grows exponentially with the number of base paths and has a bunch of duplicates. If it's meant to be this way it definitely needs some more comments for explanation.

```python
In [2]: base_paths = ["foo/bar", "foo/baz", "baz/cat", "dog/elephant"]                              

In [3]: [bp for i, bp in enumerate(base_paths) for _ in range(i + 1)]                               
Out[3]: 
['foo/bar',
 'foo/baz',
 'foo/baz',
 'baz/cat',
 'baz/cat',
 'baz/cat',
 'dog/elephant',
 'dog/elephant',
 'dog/elephant',
 'dog/elephant']
```

gerrod3: Yes it is a bit obtrusive. It uses the behavior of `_base_paths()` and [redis `exist()`](https://redis.io/commands/exists) to cleverly (at least I think so) find the base_path in one redis call. 

Say the request url is `foo/bar/snu/hello.txt` and the base_path is `foo/bar/snu` then `_base_paths()` will return `['foo/bar/snu', 'foo/bar', 'foo']`.  The redis `exists()` command can check multiple keys at once and returns the total count of all keys present, with duplicate keys being counted multiple times. Therefore by duplicating the keys by their index position I can figure out which is the real base_path from the list with one exists call since we have special code for base_paths to not overlap.

docs/configuration/settings.rst
gerrod3: Yeah I was also thinking the same. Should I just move the `EXPIRES_TTL` out to its own setting since it would be the only other setting in `PULP_CACHE_SETTINGS`? I can't think of other tunable settings that users might want to change.

requirements.txt
```diff
@@ -2,6 +2,7 @@ asyncio-throttle~=1.0
 aiohttp~=3.7.4
 aiodns~=3.0.0
 aiofiles==0.7.0
+aioredis==2.0.0a1
```
evgeni, line 5: pining a first alpha release of a complete rewrite of a library sounds dangerous to me.

dralley: One of the things we were considering is just leaving the feature off by default until the stable release (of aioredis) comes out.  Supposedly they wanted to release 2.0.0 stable this month but they will miss that deadline it seems.

The most important thing is that we want to make it easy to enable this feature later if we find out it's necessary to support the largest installations.

gerrod3: True, but is it more dangerous then using a library or version that is no longer supported? The feature can be turned off through the settings and maybe this is a good reason to have it off by default.

docs/configuration/settings.rst
ehelms: This reads like a namespace, that users would possibly want to tune if they had multiple Pulp instances hitting the same Redis? In that case, is it the 'Default' or just the 'Base Key' ?  The Default wording tends to imply I should not be changing this as its the default and I should instead configure some other setting to change the cache key namespace.

requirements.txt
evgeni: Wouldn't for that `aioredis~=2.0.0a1` or `aioredis>=2.0.0a1` also work? My main concern with the current pin is: you can't install the final 2.0.0 if this pin ships with pulpcore 3.14.0, unless it gets unpinned in a later release.

But now that you said that, I wonder (as we just had that discussed with @ehelms @ekohl and @wbclark): should Katello 4.1 have this feature enabled by default (I read you: no), or enableable at all (I read you: maybe?)

docs/configuration/settings.rst
gerrod3: Yeah, it's sort of a namespace, but only for plugin writers who want to use the cache without creating their own base_key. For caching the responses from the content-app the base_key is determined by the base_path of the distribution, so each distribution has its own "namespace" that can easily be cleared without affecting other entries. This setting is misleading in that sense and not very useful. Proper namespaces could be a useful setting, but that's not how the caching feature is set up currently.

requirements.txt
ehelms: I was implementing this for Katello as present but disabled by default with a PR I have upcoming.

docs/configuration/settings.rst
ehelms: Thanks for sorting that. To recap my understanding, users should not change this but plugin writers can modify it in their code.

ehelms: In this state, is this setting required to be set in `settings.py` for this to function or if omitted will the feature safely fall back to a default?

requirements.txt
evgeni: Given it's an optional dependency, the installer will have to handle the installation of the library

dralley: @evgeni We can certainly upgrade the pin in a later z-release, no problems there.  And both of those suggestions are fine, too.

re: Katello default or not, it's hard to make the risk/reward judgement when we still don't have any hard data from the performance team about the current non-cached content app (that's not their fault, they have a lot on their plate AFAICT).  I think the safest choice is "no" unless they come back to us and say we really need it, at which point it's trivial to enable.

But we do have several users upstream whose installations are large enough that they do really want this and will be enabling it quickly.  So we'll definitely get some testing in that way whether it's the default or not.

ekohl:
There is caching built into Django itself. Is this custom implementation here because Django's built in version is not async?

dralley:
@ekohl That's not the main reason.  We use aiohttp rather than Django for the content app because the task of streaming on_demand files while saving them to disk + Pulp's recordkeeping in the background is too low level for Django.  

But that's another valid reason.  It's a fundamentally async kind of task, and Django 2.2's async support is very poor.

docs/configuration/settings.rst
gerrod3: Now that you mention it, there is no backup currently if this setting was deleted, which would be bad. I think I'm going to remove the setting and have the default be 'PULP_CACHE' . If plugin writers want to put their entries under a different base_key then they will have to specify it themselves.

docs/configuration/settings.rst
```diff
@@ -189,6 +189,35 @@ CONTENT_APP_TTL
    Defaults to ``30`` seconds.
 
 
+.. _pulp-cache:
+
+PULP_CACHE_ENABLED
```
ekohl, line 194: Why is this prefixed with `PULP_`? Won't that mean Dynaconf will make it the env var `PULP_PULP_CACHE_ENABLED`?

docs/configuration/settings.rst
goosemania: +1 to remove the setting, it's confusing for a user I think.

pulpcore/app/settings.py
goosemania: I'm fine with enabling this one by default. This is so easy and harmless to switch on and off (just do not flush the whole redis :)).
IMO, new tasking system as a default was a much riskier decision.

pulpcore/cache/cache.py
goosemania: Could we remove it until we have any use for it? I thought that's a general approach we usually take.
If it's needed for tests only at the moment, let's implement this part in tests.

My concern that it's very dangerous until there is something else in redis apart from caching. And if we a not using a dedicated logical database within the redis, I do not think we want to perform flush for the whole thing.

pulpcore/content/handler.py
```diff
@@ -238,9 +284,10 @@ def _permit(request, distribution):
         """
         guard = distribution.content_guard
         if not guard:
-            return
+            return False
         try:
             guard.cast().permit(request)
+            return True
```
goosemania, line 290: I'd vote for excluding return from try/except clause for conventional reasons of including only code that is expected to produce the exception.

goosemania (commented):
I tested this PR with the old tasking system (to be sure that nothing is accidentally removed in redis), monitored db requests/checked redis keys/values. Went through various scenarios when cache needs to be invalidated. Works well. Thanks! LGTM from me. 
Let me know if you need a formal ack, I'd leave it for the rest who spent more time working with you on this.

docs/configuration/settings.rst
gerrod3: Changed it to just `CACHE_ENABLED`

pulpcore/cache/cache.py
gerrod3: Ok I'll remove it, it's not hard to add back if we need it later.

pulpcore/app/models/publication.py
gerrod3: I added documentation to the `delete` call on the cache api to note the different behavior when passing in lists.

docs/configuration/settings.rst
```diff
@@ -189,6 +189,33 @@ CONTENT_APP_TTL
...
+
+   .. note::
+     The entire response is not stored in the cache. Only the location of the file needed to
+     recreate the response is stored. This reduces database queries and allows for many
+     responses to be stored inside the cache.
+
+CACHE_SETTINGS
+^^^^^^^^^^^^^
+
+   Dictionary with two tunable settings for the cache:
```
bmbouter, line 210: Does this still have two settings? I only see one. It would be good I think to state the default also.

docs/configuration/settings.rst
```diff
@@ -189,6 +189,33 @@ CONTENT_APP_TTL
...
 
+.. _pulp-cache:
+
+CACHE_ENABLED
+^^^^^^^^^^^^^^^^^^
+
+   .. note:: This feature is provided as a tech-preview
+
+   Store cached responses from the content app into Redis. This setting improves the performance
+   of the content app under heavy load for similar requests.
```
bmbouter, line 200: It would be good to state the default I think. Perhaps by example:

`CACHE_ENABLED=True`

pulpcore/cache/cache.py
```diff
@@ -0,0 +1,234 @@
+import enum
+import json
+
+from aiohttp.web import FileResponse, Response, HTTPSuccessful, Request
+from aiohttp.web_exceptions import HTTPFound
+
+from pulpcore.app.settings import settings
+from pulpcore.tasking.connection import get_redis_connection, get_async_redis_connection
+
+DEFAULT_EXPIRES_TTL = settings.CACHE_SETTINGS.get("EXPIRES_TTL", None)
```
bmbouter, line 10: This is a required setting right, and it has a sane default. Given that, I think not being defensive would be better actually. Say someone accidentally set the settings `CACHE_SETTINGS = {"ESPIRES_TTL": None}`.

pulpcore/content/handler.py
bmbouter: I'm not asking for a change here, but I want to share a perspective on the value of readability. Comments describing what code does are tricky in the long term because, often implementations change but comments don't. Readability is one of Pythons core language goals because code is read many more times than it's written. These goals taken together (to me) suggest that one should write code that is extremely easy to read. In this case, that would be avoiding the double list comprehension in favor of a nested loop with absurdly long (and clear) variable names. If it were written that way, it likely wouldn't need a comment at all.

@ehelms I still remember when you taught me this years ago.

bmbouter:
I don't want to ask for the change in this PR due to timeline constraints, but one thing we should think about in the future is what happens when Pulp is configured to cache, but it's not available (netsplit, redis died, whatever). I believe right now it would error, and in some codepaths maybe it should, e.g. cache invalidation perhaps?

Overall though, to me it would be nice if the content app could fall back to it's original non-caching implementation.

bmbouter:
@gerrod3 I left a few small comments about settings reading and the docs, but code-wise this is looking great!

pulpcore/content/handler.py
gerrod3: I agree with your readability comment, but since the whole point of the cache is that it's suppose to be quick I think an exception can be made here since list comprehensions are faster than basic for loops in python. If there is a way to create the same list in a more readable fashion that's just as fast as list comprehension then I'm all for changing it and removing the comments.

dralley: The margins between a for loop and a list comprehension aren't enough IMO to sacrifice readability.  We're IO bound by a huge margin, the cost of network hops is going to swamp the loop.

bmbouter: @dralley I 100% agree with this. Also FWIW interpreters like pypy recompile hotloops like doubly nested loops using JIT compilation at runtime anyway.

dralley (approved):

ehelms:
On Wed, Jun 30, 2021, 2:58 PM bmbouter ***@***.***> wrote:

> I don't want to ask for the change in this PR due to timeline constraints,
> but one thing we should think about in the future is what happens when Pulp
> is configured to cache, but it's not available (netsplit, redis died,
> whatever). I believe right now it would error, and in some codepaths maybe
> it should, e.g. cache invalidation perhaps?
>
> Overall though, to me it would be nice if the content app could fall back
> to it's original non-caching implementation.
>
This one concerns me if not already true. I would expect if the cache is
unavailable to fall back to standard content serving functionality with
some error raised by the app (e.g. in logs or by status API) that a
monitoring system could consume. Give the user a degraded experience but
not affect availability.

dralley:
@ehelms Completely agree - it should be easy to verify / fix though.  If it doesn't fall back correctly, that's a bug for 3.14.1

pulpcore/cache/cache.py
bmbouter: What I mean is `DEFAULT_EXPIRES_TTL = settings.CACHE_SETTINGS["EXPIRES_TTL"]`

That way if a user has misconfigured it, it will raise a KeyError.

bmbouter:
@ehelms I also agree, but with the merge by end of business today, I think we would have to fix in 3.14.1. I've asked on matrix for a bug to be filed and linked here so we're sure.

dralley:
I've filed these two issues to address in 3.14.1 (or at least an early z-release)

* https://pulp.plan.io/issues/8996
* https://pulp.plan.io/issues/8997

bmbouter (approved):
This looks good to me. The changes I requested and I've seen requested have been made. There are two followup lingering items, but we have a merge deadline of today so I'm going to merge and we'll z-stream the other two items.

Thank you @gerrod3, @dralley and everyone who helped with this significant piece of work!

ehelms:
We need some help with the finer points of the configuration, if someone could look at:

 * https://github.com/theforeman/puppet-pulpcore/pull/204#discussion_r663133677
 * https://github.com/theforeman/puppet-pulpcore/pull/204#discussion_r663131039

PR #1302, opened 2021-05-04
Un-remove ContentUnassociation stage, deprecate it

dralley:

PR #1363, opened 2021-05-26
Merge ContentUnassociation stage into the ContentAssociation stage

dralley:
ContentUnassociation doesn't really need to be a separate stage, and
making it one breaks the expectations of the "pipeline" by changing the
semantics of what goes into and out of the queues.

closes: #8827
https://pulp.plan.io/issues/8827


mdellweg (changes requested):
I really like this change.

However, i think we cannot simply remove the `ContentUnassociation` stage and change the signature of the `ContentAssociation` stage.

To adhere to the deprecation policy, i suggest to:
* Deprecate the `ContentUnassociation` stage
* Make the (new) `mirror` flag in the `ContentAssociation` stage default to `False`.

Sadly we would need to keep the bad interface of passing on the queryset to the deprecated `ContentUnassociation` stage for a release longer.

Maybe I can be convinced, that these two stages de facto never were part of the plugin api. Is there a way, we can move them out of the pulpcore/plugin directory in that case?

dralley:
> To adhere to the deprecation policy, i suggest to:
> * Deprecate the ContentUnassociation stage

@mdellweg We did do that :) 

https://pulp.plan.io/issues/8635
https://github.com/pulp/pulpcore/pull/1302

> Maybe I can be convinced, that these two stages de facto never were part of the plugin api. Is there a way, we can move them out of the pulpcore/plugin directory in that case?

FWIW, this is essentially true also, so even this is more of an "abundance of caution".  No plugin uses these directly and it's very difficult to actually do so, the standard API that plugins that create their own pipelines use can't do it because they're injected at the very end of the pipeline by pulpcore.

mdellweg (approved):
I was convinced. ;)

What do you still think about not having them part of the pulpcore.plugin interface?
Or maybe guard by adding "_" in the name?

bmbouter:
I know in practice plugins don't use this, but to be totally technical, they could.

My opinion is that all the parts of the stages API should be available in `pulpcore.plugin`.

dralley:
They could but they would need to patch / override `DeclarativeVersion` itself to do so.

mdellweg:
> I know in practice plugins don't use this, but to be totally technical, they could.
> 
> My opinion is that all the parts of the stages API should be available in `pulpcore.plugin`.

That means (de jure ; not de facto) we can never change the interface of anything in the stages api (even if never used directly by a plugin) in a backwards incompatible way without properly deprecating for a release.

bmbouter:
I'm being pedantic here because I don't know anyone who does this, but plugin writers are allowed to construct their own pipelines without `DeclarativeVersion`.

dralley:
@bmbouter Yes but that mechanism doesn't allow you to influence this part of the pipeline

https://github.com/pulp/pulpcore/blob/master/pulpcore/plugin/stages/declarative_version.py#L143-L147

dralley:
The migration plugin does do that, but they don't use the `Content*Association` stages at all.  They just create the content.

bmbouter (approved):
Thanks @dralley !

PR #1367, opened 2021-05-26
Add timestamp_of_interest to Content and Artifact

daviddavis:
fixes #8823


dralley:
Would it not be possible to re-use `updated_date`?  IIRC we currently never touch it for content or artifacts.

daviddavis:
@dralley that's a good question. I was going to say that updating `pulp_last_updated` would give users the false impression that content and artifacts are being updated every time we use them (e.g. to create a new repo version) but I checked the code and it doesn't look like we expose `pulp_last_updated` to users. @bmbouter @ipanova I believe this was your design. thoughts?

ipanova:
> @dralley that's a good question. I was going to say that updating `pulp_last_updated` would give users the false impression that content and artifacts are being updated every time we use them (e.g. to create a new repo version) but I checked the code and it doesn't look like we expose `pulp_last_updated` to users. @bmbouter @ipanova I believe this was your design. thoughts?

Yeah, we did not want to use last_updated to not give an impression that content/artifact is mutable.

pulpcore/app/models/content.py
```diff
@@ -495,6 +500,11 @@ def init_from_artifact_and_relative_path(artifact, relative_path):
         """
         raise NotImplementedError()
 
+    @hook(AFTER_CREATE)
+    def update_artifact_timestamps(self):
+        """Update timestamp_of_interest on artifacts after content is created."""
+        self._artifacts.update(timestamp_of_interest=now())
```
mdellweg, line 506: https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.DateField.auto_now

You should not need to set "now()", just call save (maybe with `update_fields`) on the object.

daviddavis: Calling `update()` without any fields doesn't update timestamp_of_interest and there is no `update_fields` option for `update`. I could iterate through each object and call `save()` with `update_fields` but that seems less efficient than using `update()` which issues a single query.

daviddavis: I do see some lines below where I am explicitly setting `timestamp_of_interest` before calling `save()`. I will update those lines.

bmbouter:
> @dralley that's a good question. I was going to say that updating `pulp_last_updated` would give users the false impression that content and artifacts are being updated every time we use them (e.g. to create a new repo version) but I checked the code and it doesn't look like we expose `pulp_last_updated` to users. @bmbouter @ipanova I believe this was your design. thoughts?

Yes this is what I was thinking also. FWIW, we should over time likely move `pulp_last_updated` further down the inheritance tree to remove it from some models altogether. Especially the immutable ones For the purposes of this goal though, I'm +1 to using the dedicated field.

dralley:
Re-using the field is not mutually exclusive with presenting a clear picture to the user.  We're currently hiding this field from the serializer - we could in theory just provide a new field under a new name on the serializer, and point it to the `pulp_last_updated` field.

Or we could fix our inheritance so that we're not inheriting fields we don't actually want to use.

bmbouter:
@dralley that makes sense, but for me the issue is clarity in naming for plugin writers. The data didn't actually change so using `pulp_last_updated` feel like the wrong way for plugin writers to update that. That's just my perspective on the naming.

dralley:
It's not clear to me that plugin writers would actually need to know about this though.  

>Added a ``timestamp_of_interest`` field to Content and Artifacts that plugin writers should update
whenever they touch Content or Artifacts. For example, this includes places where Content is
uploaded or added to Repository Versions. This will prevent recently used Content and Artifacts from
being cleaned up when orphan cleanup becomes a non-blocking task in pulpcore 3.15.

When Content and Artifacts are created, no intervention should be needed, because `auto_now_add` should handle it (or the base save handlers).

And RepositoryVersions are generally (always? I'm not personally aware of other cases) created using some form of wrapping API like DeclarativeVersion.

daviddavis:
I think we also have to handle the case where content or artifact already exists. We should update the timestamp but these places don't exist necessarily in pulpcore. Here's [an example from pulp_rpm](https://github.com/pulp/pulp_rpm/blob/f0ffeb61deef3124af08dd45c6f4c348ce2dc6ad/pulp_rpm/app/serializers/package.py#L242-L261).

In the case where repository versions are being created, we need to update the timestamp long before the repo version is created so that the content/artifacts we've fetched from the database don't get cleaned up while the task is running. pulp_container has [repo add and remove endpoints](https://github.com/pulp/pulp_container/blob/862e0e665ee89e3c32c3069f072cdf7223f3c54e/pulp_container/app/viewsets.py#L585-L643) that will need to do this.

dralley:
So, `timestamp_of_interest` is not just a way of determining whether something should be cleaned up, but is also meant to be an ad-hoc locking mechanism?

mdellweg:
Isn't select_for_update only valid in a transaction, or on the same database connection?
The additional complication here, is the a user can upload a content_unit that already existed. In that case the content must be updated to prevent its deletion, before the user can add the content to a repoversion only seconds later.

ipanova:
@dralley in addition to what @mdellweg mentioned, I still think we cannot get away without the `timestamp` in the sync pipeline.

ipanova:
Should not we update the timestamp in this case? https://github.com/pulp/pulpcore/blob/master/pulpcore/content/handler.py#L571

I looked at other tasks, and for example in the import task, I can't say for sure whether there is a window between the imported artfacts and imported repo-versions, which will lead to artifact removal? can someone look and confirm?

pulpcore/app/models/content.py
```diff
@@ -495,6 +500,11 @@ def init_from_artifact_and_relative_path(artifact, relative_path):
         """
         raise NotImplementedError()
 
+    @hook(AFTER_CREATE)
```
ipanova, line 503: why there is the need to update timestamp on artifacts while they are related already to the just created content, which basically won't let orphan task delete those because of the membership?

daviddavis:
I updated content handler to bump the timestamp for the Artifact. I had to also handle the case where the artifact got cleaned up before we could bump the timestamp. Check it out.

For import, django-import-export updates existing Artifacts/Content which bumps the timestamp_of_interest so I don't think there'll be a window for an Artifact/Content to be removed between when it's added/updated and then added to a repo version:

```
In [3]: Artifact.objects.all().values_list("pk", "timestamp_of_interest")
Out[3]: <QuerySet [(UUID('797d7094-d813-41fd-b30b-a6ec0682563f'), datetime.datetime(2021, 6, 15, 18, 42, 54, 678176, tzinfo=<UTC>)), (UUID('ceea5124-80bc-499c-a08b-242bbd12e769'), datetime.d
atetime(2021, 6, 15, 18, 42, 54, 684240, tzinfo=<UTC>)), (UUID('005c87cb-52c6-4bcc-a337-b502827a8093'), datetime.datetime(2021, 6, 15, 18, 42, 54, 690131, tzinfo=<UTC>))]>

In [4]: Content.objects.all().values_list("pk", "timestamp_of_interest")
Out[4]: <QuerySet [(UUID('eb75ca68-5e60-4cfa-906a-9991f2052e57'), datetime.datetime(2021, 6, 15, 18, 42, 55, 526349, tzinfo=<UTC>)), (UUID('dd4072fa-7444-4115-ba28-664c60564ad1'), datetime.d
atetime(2021, 6, 15, 18, 42, 55, 534239, tzinfo=<UTC>)), (UUID('f7de6d03-12ad-4cc6-9efc-a5bba63e9215'), datetime.datetime(2021, 6, 15, 18, 42, 55, 541277, tzinfo=<UTC>))]>

# RUN IMPORT HERE

In [5]: Artifact.objects.all().values_list("pk", "timestamp_of_interest")
Out[5]: <QuerySet [(UUID('797d7094-d813-41fd-b30b-a6ec0682563f'), datetime.datetime(2021, 6, 15, 18, 53, 9, 83206, tzinfo=<UTC>)), (UUID('ceea5124-80bc-499c-a08b-242bbd12e769'), datetime.datetime(2021, 6, 15, 18, 53, 9, 89411, tzinfo=<UTC>)), (UUID('005c87cb-52c6-4bcc-a337-b502827a8093'), datetime.datetime(2021, 6, 15, 18, 53, 9, 95032, tzinfo=<UTC>))]>

In [6]: Content.objects.all().values_list("pk", "timestamp_of_interest")
Out[6]: <QuerySet [(UUID('eb75ca68-5e60-4cfa-906a-9991f2052e57'), datetime.datetime(2021, 6, 15, 18, 53, 9, 954064, tzinfo=<UTC>)), (UUID('dd4072fa-7444-4115-ba28-664c60564ad1'), datetime.datetime(2021, 6, 15, 18, 53, 9, 960436, tzinfo=<UTC>)), (UUID('f7de6d03-12ad-4cc6-9efc-a5bba63e9215'), datetime.datetime(2021, 6, 15, 18, 53, 9, 966206, tzinfo=<UTC>))]>
```

pulpcore/plugin/actions.py
```diff
@@ -50,6 +51,9 @@ def modify(self, request, pk):
                     content = self.get_resource(url, Content)
                     remove_content_units.append(content.pk)
 
+        # set timestamp_of_interest
+        Content.objects.filter(pk__in=add_content_units).update(timestamp_of_interest=now())
```
mdellweg, line 55: I think, we need to "touch" that timestamp individually for all content_units, because if the content is deleted between line 41 and here, the `filter` will not complain.

Another solution would be to wrap this whole block into a shared advisory lock that orphan cleanup needs to hold exclusively between identifying and deleting an orphan.

pulpcore/content/handler.py
```diff
@@ -648,7 +649,15 @@ def _save_artifact(self, download_result, remote_artifact):
...
                     artifact.save()
             except IntegrityError:
-                artifact = Artifact.objects.get(artifact.q())
+                try:
+                    artifact = Artifact.objects.get(artifact.q())
+                    artifact.save(update_fields=["timestamp_of_interest"])
+                except (Artifact.DoesNotExist, DatabaseError):
+                    # it's possible that orphan cleanup deleted the artifact
+                    # so fall back to creating a new artifact again
+                    artifact = Artifact(
```
ipanova, line 658: :+1:

pulpcore/plugin/viewsets/content.py
```diff
@@ -110,6 +110,7 @@ def init_content_data(self, serializer, request):
             except IntegrityError:
                 # if artifact already exists, let's use it
                 artifact = Artifact.objects.get(sha256=artifact.sha256)
+                artifact.save(update_fields=["timestamp_of_interest"])
```
ipanova, line 113: @daviddavis  i think you should take same approach as in content handler where the artifact might have been removed between lines 112 and 113, in that case use artifact from line 102

pulpcore/content/handler.py
ipanova: you should save it though

ipanova (approved):

pulpcore/content/handler.py
```diff
@@ -648,7 +649,16 @@ def _save_artifact(self, download_result, remote_artifact):
                 with transaction.atomic():
                     artifact.save()
             except IntegrityError:
-                artifact = Artifact.objects.get(artifact.q())
+                try:
+                    artifact = Artifact.objects.get(artifact.q())
+                    artifact.save(update_fields=["timestamp_of_interest"])
```
mdellweg, line 654: Would it be useful for readability to define a `touch` method for this call?

mdellweg (approved):

pulpcore/content/handler.py
daviddavis: Sure, should this method just call ‘save(update_fields=["timestamp_of_interest"])’?

mdellweg: That is my thought. And plugin writers would call it like `artifact.touch()`; same for content.

PR #1382, opened 2021-06-01
Consolidates existing auth into top-level section

bmbouter:
This moves all of the authentication into one place and organizes it
into and Overview, Basic, Webserver, and Other sections.

It also deletes the 'disabling auth' docs. Upon testing Django refuses
to start if you follow those docs.

closes #8801


dralley (approved):

goosemania (approved):
Thanks!

PR #1385, opened 2021-06-01
Caching OpenAPI schema

fao89:
https://pulp.plan.io/issues/8830
closes #8830

### DIFF
--- a/CHANGES/8830.feature
+++ b/CHANGES/8830.feature
@@ -1,0 +1,1 @@
+Cache OpenAPI schema
--- a/pulpcore/openapi/__init__.py
+++ b/pulpcore/openapi/__init__.py
@@ -1,7 +1,10 @@
+import json
 import os
 import re
 from urllib.parse import urljoin
+from pathlib import Path
 
+from django.conf import settings
 from django.core.exceptions import FieldDoesNotExist
 from django.utils.html import strip_tags
 from drf_spectacular.drainage import reset_generator_stats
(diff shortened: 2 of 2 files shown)


mdellweg (commented):
:+1:
Loving it.
There is one thing to consider though: If you are in a dev environment, and you are changing some api, this would not be reflected by this caching. And you'd need to delete that file by hand over and over.
I see a few solutions to this:
* Add a setting to not cache it in dev environments.
* Detect if we are in a volatile environment and disable caching accordingly
  * is there a `dev` in any plugin version?
  * is any plugin installed with `pip -e`
* Remove that file with `pstart` and `prestart`

daviddavis:
> Add a setting to not cache it in dev environments.

We don't use Django's [DEBUG setting](https://docs.djangoproject.com/en/3.2/ref/settings/#debug) but maybe we should?

bmbouter:
At the pulpcore meeting we discussed wanting to investigate the memory leak as an alternative fix to this approach.

mdellweg:
I have another approach for this: We could add a post installation step that generates the file and puts it on disk.
Whenever the file is there, we serve it, if not, we do the generate on the fly.

bmbouter:
I think we should close this for a few reasons. 1) I don't think we don't really want caching. It can lead to problems I believe. 2) If we did want caching we wouldn't add the feature to Pulp; I think we'd add it to the dependency providing this endpoint to start with. 3) The folks affected by the memory leak already put in their workaround.

fao89:
we had 2 views for the OpenAPI and the one that most concerned me got removed at https://github.com/pulp/pulpcore/pull/1593
So I agree we can close this PR

PR #1388, opened 2021-06-02
Add method field to FilesystemExporter

daviddavis:
fixes #8695


pulpcore/app/models/exporter.py
```diff
@@ -82,10 +84,14 @@ def _export_to_file_system(self, content_artifacts):
             content_artifacts (django.db.models.QuerySet): Set of ContentArtifacts to export
 
         Raises:
-            ValidationError: When path is not in the ALLOWED_EXPORT_PATHS setting
+            RuntimeError: If there are remote artifacts or when trying to link non-fs files
         """
         if content_artifacts.filter(artifact=None).exists():
-            RuntimeError(_("Remote artifacts cannot be exported."))
+            raise RuntimeError(_("Remote artifacts cannot be exported."))
```
daviddavis, line 90: :grimacing:

daviddavis:
I filed an issue here as this change causes the field to be added on the detail table: https://pulp.plan.io/issues/8860

CHANGES/8695.feature
```diff
@@ -0,0 +1,2 @@
+Added ``method`` field to file system exporters to customize how they export files. Users can now
+export files by writing them to the filesyste, using hardlinks, or using symlinks.
```
bmbouter, line 2:
```suggestion
export files by writing them to the filesystem using hardlinks, or using symlinks.
```

I'm not 100% sure on the comma removal I'm suggesting.

pulpcore/app/models/exporter.py
```diff
@@ -82,10 +86,14 @@ def _export_to_file_system(self, content_artifacts):
             content_artifacts (django.db.models.QuerySet): Set of ContentArtifacts to export
 
         Raises:
-            ValidationError: When path is not in the ALLOWED_EXPORT_PATHS setting
+            RuntimeError: If there are remote artifacts or when trying to link non-fs files
```
bmbouter, line 89:
```suggestion
            RuntimeError: If Artifacts are not downloaded or when trying to link non-fs files
```

This would be clearer to me to avoid confusion with `RemoteArtifacts`. This suggestion is totally optional.

pulpcore/app/models/exporter.py
```diff
@@ -82,10 +86,14 @@ def _export_to_file_system(self, content_artifacts):
...
         """
         if content_artifacts.filter(artifact=None).exists():
-            RuntimeError(_("Remote artifacts cannot be exported."))
+            raise RuntimeError(_("Remote artifacts cannot be exported."))
```
bmbouter, line 92: Maybe similar clarity touch up here about the Artifact not yet being downloaded.

bmbouter (approved):
I left a few comments for string fixes, but code-wise this looks great.

pulpcore/app/models/exporter.py
daviddavis: How about `Undownloaded artifacts cannot be exported.`?

daviddavis: Or better yet: `Cannot export artifacts that haven't been downloaded.`

pulpcore/app/tasks/export.py
```diff
@@ -59,12 +67,17 @@ def _export_to_file_system(path, content_artifacts):
         except FileExistsError:
```
mdellweg, line 67: This would be easier if we just set `exist_ok=True` (Introduced in Python 3.2).

https://docs.python.org/3/library/os.html#os.makedirs

pulpcore/app/tasks/export.py
```diff
@@ -59,12 +67,17 @@ def _export_to_file_system(path, content_artifacts):
...
             pass
 
-        if settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem":
+        if method == FS_EXPORT_METHODS.SYMLINK:
```
mdellweg, line 70: These methods will probably only work with "pulpcore.app.models.storage.FileSystem", right?

pulpcore/app/tasks/export.py
```diff
@@ -59,12 +67,17 @@ def _export_to_file_system(path, content_artifacts):
...
+            src = os.path.join(settings.MEDIA_ROOT, artifact.file.name)
+            os.symlink(src, dest)
+        elif method == FS_EXPORT_METHODS.HARDLINK:
             src = os.path.join(settings.MEDIA_ROOT, artifact.file.name)
             os.link(src, dest)
-        else:
+        elif method == FS_EXPORT_METHODS.WRITE:
             with open(dest, "wb") as f:
                 f.write(artifact.file.read())
```
mdellweg, line 78: This is reading the complete artifact into memory at once. Can we batch it?

mdellweg:
> I filed an issue here as this change causes the field to be added on the detail table: https://pulp.plan.io/issues/8860

Is this still true? Or was this fixed by refactoring lately?

pulpcore/app/tasks/export.py
daviddavis: Yes, there is a check on lines 55-59

daviddavis:
> Is this still true? Or was this fixed by refactoring lately?

Yea, not still true. It was the same problem as the distribution problem we had and the refactoring fixed it.

pulpcore/app/tasks/export.py
mdellweg: I see it now. ;)

mdellweg (approved):

pulpcore/app/tasks/export.py
daviddavis: Ok check my code. I used `:=` because I assume this change won't be backported.

pulpcore/app/tasks/export.py
```diff
@@ -46,25 +47,36 @@ def _export_to_file_system(path, content_artifacts):
...
+            os.symlink(src, dest)
+        elif method == FS_EXPORT_METHODS.HARDLINK:
             src = os.path.join(settings.MEDIA_ROOT, artifact.file.name)
             os.link(src, dest)
-        else:
+        elif method == FS_EXPORT_METHODS.WRITE:
             with open(dest, "wb") as f:
-                f.write(artifact.file.read())
+                while chunk := f.read(1000 * 1000):
+                    f.write(chunk)
```
mdellweg, line 76: I was expecting something like:
https://github.com/pulp/pulpcore/blob/master/pulpcore/app/tasks/repository.py#L90

Maybe
```python
with open(dest, "wb") as f, artifact.file as af:
    for chunk in af.chunks(CHUNK_SIZE):
        f.write(chunk)
```

Not sure if it actually _is_ better.

daviddavis: Wow, this code I wrote is wrong. I must have been on autopilot on Friday.

daviddavis: I fixed this. Want to review one last time before I merge?

mdellweg (approved):

PR #1392, opened 2021-06-03
Refactor FilesystemExporter

daviddavis:
fixes #8860

TODO
- [x] Update tests
- [x] Add validation to prevent both repo version and publication params being supplied
- [x] Save FileExport records when they are exported


dralley:
As per the pulp-meeting discussion, we should avoid merging any PRs containing migrations before Friday to enable the possibility for 3.14.z-stream migrations.  And we'll decide then how to proceed.

pulpcore/app/tasks/export.py
```diff
@@ -33,6 +36,37 @@
...
+        path (str): A path to export the ContentArtifacts to
+        content_artifacts (django.db.models.QuerySet): Set of ContentArtifacts to export
+
+    Raises:
+        ValidationError: When path is not in the ALLOWED_EXPORT_PATHS setting
+    """
+    if content_artifacts.filter(artifact=None).exists():
+        RuntimeError(_("Remote artifacts cannot be exported."))
+
+    for ca in content_artifacts:
```
dralley, line 53: Two things

* This can be a big list, use `.iterator()`
* use `.select_related("artifact")` to avoid the N+1 problem

pulpcore/app/viewsets/exporter.py
```diff
@@ -134,3 +150,49 @@ def create(self, request, exporter_pk):
...
+        serializer = FilesystemExportSerializer(data=request.data, context={"exporter": exporter})
+        serializer.is_valid(raise_exception=True)
+
+        if request.data.get("publication"):
+            publication = self.get_resource(request.data["publication"], Publication)
+
+            task = dispatch(
+                fs_publication_export,
+                [exporter],
+                kwargs={"exporter_pk": str(exporter.pk), "publication_pk": publication.pk},
```
dralley, line 187: `exporter.pk` is wrapped in `str()` but `publication.pk` is not - `str()` isn't strictly needed anymore so we can remove it.  Or use it in both places, whatever you prefer.

pulpcore/app/viewsets/exporter.py
```diff
@@ -134,3 +150,49 @@ def create(self, request, exporter_pk):
...
+        if request.data.get("publication"):
+            publication = self.get_resource(request.data["publication"], Publication)
+
+            task = dispatch(
+                fs_publication_export,
+                [exporter],
+                kwargs={"exporter_pk": exporter.pk, "publication_pk": publication.pk},
+            )
+        else:
+            repo_version = self.get_resource(request.data["repository_version"], RepositoryVersion)
```
mdellweg, line 190: This would allow exporting of repository_versions of any repository type, right?
Don't we have a flag to tell us whether a repo type needs publications? Maybe it's not yet merged...

mdellweg (approved):
Looks good to me.
Is the removing of ``FilesystemExporterSerializer`` and ``PublicationExportSerializer`` something that needs a deprecation round?

pulpcore/app/models/exporter.py
```diff
@@ -62,6 +56,14 @@ class Exporter(MasterModel):
     name = models.TextField(db_index=True, unique=True)
 
 
+class FilesystemExport(Export):
+    """
+    A model that represents an export to the filesystem.
```
dralley, line 61: Apologies if this is already written up somewhere.  What is the use case for tracking / modelling exports?

Because Pulp has no control over what happens with the files on the filesystem post-export, keeping records could actually be extremely misleading.  The very next step in the user pipeline could be running createrepo_c in-place to make a small modification, or moving the directory somewhere else, or deleting it.  Thereafter, the export records seem like they would be - in the best case scenario - not very useful, and in the worst case scenario actively misleading.

Is it required or is there a benefit gained from doing so?

daviddavis: It's meant to keep a historical record of an export. I agree though that it's misleading and perhaps there is a better term for it (`ExportHistory` or something?). This PR doesn't add this concept to Pulp so I think filing an issue would make sense if you feel strongly that we should change it.

dralley: I think I will.  My mental model is that if a user wants their data out of Pulp, it's likely a one-time event, and not something that they'd keep coming back for over and over.

Beyond the use case already mentioned, I would expect that users may want to keep exporting different versions of the same repository to the same path.  And in that case all the records would be incorrect as well.

pulpcore/app/viewsets/exporter.py
daviddavis: > This would allow exporting of repository_versions of any repository type, right?

Correct, I filed an issue here for this: https://pulp.plan.io/issues/9142.

> Don't we have a flag to tell us whether a repo type needs publications? Maybe it's not yet merged...

I think you're referring to [this PR](https://github.com/pulp/pulpcore/pull/1495). The field is on the Distribution though so I'm not totally sure how using a repo version I can determine if the plugin uses publications? If you have a proposal, feel free to put it on 9142.

pulpcore/app/models/exporter.py
dralley: I think the original idea would have been that we should put something in `created_resources`.  I'm not sure if it's worth it though.

daviddavis:
> Is the removing of FilesystemExporterSerializer and PublicationExportSerializer something that needs a deprecation round?

That's a good question. My thought is that (a) there are no plugins using these (pulp_file was but it was removed) and (b) they were in tech preview. We could do it though--it'd be like a day of work. We'd have to rename FilesystemExporterSerializer and bring back PublicationExportSerializer and add deprecation warnings. Then in 3.16, we could remove the models and rename NewFilesystemExporterSerializer back to FilesystemExporterSerializer. wdyt?

dralley:
My personal opinion is that we probably don't need to do so, given the tech-preview status and the aforementioned limited usage.  If we are releasing this in 3.15 though, I really don't want to introduce more API surface that we may then want to think about removing (as per previous comment).

pulpcore/app/models/exporter.py
ggainey: So a **very** common use-case for some of our pulp2 users, is to export Pulp distributions to "somewhere else" on the file system to be served directly via apache/nginx/whatever-webserver; also, the RsyncExporter was used to do the same thing, to a far-away system Tracking the exports lets them know when it last was done, and what happened. Like any workflow, that's useful info (like, say, when an admin is on PTO (or has left the company)). That's one main reason these records exist, and why it takes deliberate action on the admin's part to delete them. 

(The other is, the only way to do 'incremental' updates is to know what you did when, to where. Again, a really common use-case is "export Pulp's content to be served from some Not-Pulp that is Elsewhere", as part of a regular admin workflow.)

dralley: That makes sense, but in this case it cannot actually track when the rsync last happened because this is only the filesystem export component of that - and there is no mechanism for triggering arbitrary external actions on completion.

Plus, it feels in some sense like duplicating the records the tasking system already keeps.

ggainey: Rsync would be its own kind-of Thing, we don't have it yet. But even when we do - you'll have an export, that says "we rsync'd these repos to that location on this datetime". It's not about 'arbitrary' external commands, the Rsync Exporter knows that it lives to do rsync.  But this is FileSystemExporter - which knows it's exporting a distribution to a new location on disk.

RE duplicating task-info - possibly, but I propose that "task" is a much more temporal/temporary kind-of Thing than what we're talking about here.

ggainey: I guess that, while I'm not viscerally tied to this approach, it is the way it is because of experience with actual users running actual workflows.

daviddavis: Any chance we can merge this PR as-is and follow up this discussion on an issue? This work is in tech preview. And it's blocking other work I have (https://pulp.plan.io/issues/8695) and a user is waiting on it (https://github.com/pulp/community/discussions/51).

dralley (approved):

PR #1394, opened 2021-06-04
Adds `/pulp/api/v3/orphans/cleanup/`

bmbouter:
This adds a new endpoint `/pulp/api/v3/orphans/cleanup/`. When called
with `POST` and no parameters it is equivalent to calling `DELETE
/pulp/api/v3/orphans/`. Additionally the optional parameter
`content_hrefs` can be specified and must contain a list of content
hrefs. When `content_hrefs` is specified, only those content units will
be considered to be removed by orphan cleanup.

closes #8658

### DIFF
--- a/CHANGES/8658.feature
+++ b/CHANGES/8658.feature
@@ -1,0 +1,4 @@
+Added new endpoint ``/pulp/api/v3/orphans/cleanup/``. When called with ``POST`` and no parameters
+it is equivalent to calling ``DELETE /pulp/api/v3/orphans/``. Additionally the optional parameter
+``content_hrefs`` can be specified and must contain a list of content hrefs. When ``content_hrefs``
+is specified, only those content units will be considered to be removed by orphan cleanup.
--- a/pulpcore/app/serializers/orphans.py
+++ b/pulpcore/app/serializers/orphans.py
@@ -1,0 +1,34 @@
+from gettext import gettext as _
+
+from rest_framework import fields, serializers
+
+from pulpcore.app.models import Content
+from pulpcore.app.serializers import ValidateFieldsMixin
+
+
+class OrphansCleanupSerializer(serializers.Serializer, ValidateFieldsMixin):
+
+    content_hrefs = fields.ListField(
+        required=False,
+        help_text=_("Will delete specified content and associated Artifacts if they are orphans."),
+    )
+
+    def validate_content_hrefs(self, value):
+        """
+        Check that the content_hrefs is not an empty list and contains all valid hrefs.
+        Args:
+            value (list): The list supplied by the user
+        Returns:
+            The list of pks (not hrefs) after validation
+        Raises:
+            ValidationError: If the list is empty or contains invalid hrefs.
+        """
+        if len(value) == 0:
+            raise serializers.ValidationError("Must not be [].")
+        from pulpcore.app.viewsets import NamedModelViewSet
+
+        pks_to_return = []
+        for href in value:
+            pks_to_return.append(NamedModelViewSet.get_resource(href, Content).pk)
+
+        return pks_to_return
--- a/pulpcore/app/tasks/orphan.py
+++ b/pulpcore/app/tasks/orphan.py
@@ -29,11 +29,14 @@ def queryset_iterator(qs, batchsize=2000, gc_collect=True):
             gc.collect()
 
 
-def orphan_cleanup():
+def orphan_cleanup(content_pks=None):
     """
     Delete all orphan Content and Artifact records.
     Go through orphan Content multiple times to remove content from subrepos.
     This task removes Artifact files from the filesystem as well.
+
+    Kwargs:
+        content_pks (list): A list of content pks. If specified, only remove these orphans.
 
     """
     progress_bar = ProgressReport(
@@ -48,6 +51,9 @@ def orphan_cleanup():
         content = Content.objects.filter(version_memberships__isnull=True).exclude(
             pulp_type=PublishedMetadata.get_pulp_type()
         )
+        if content_pks:
+            content = content.filter(pk__in=content_pks)
+
         content_count = content.count()
         if not content_count:
             break
--- a/pulpcore/tests/functional/api/test_correlation_id.py
+++ b/pulpcore/tests/functional/api/test_correlation_id.py
@@ -4,7 +4,7 @@ import uuid
 from pulp_smash import api, cli, config
 from pulp_smash.pulp3.bindings import monitor_task
 
-from pulpcore.client.pulpcore import ApiClient, OrphansApi, TasksApi
+from pulpcore.client.pulpcore import ApiClient, OrphansCleanupApi, TasksApi
 
 
 class CorrelationIdTestCase(unittest.TestCase):
@@ -23,12 +23,12 @@ class CorrelationIdTestCase(unittest.TestCase):
         )
         cls.cli_client = cli.Client(cls.cfg)
 
-        cls.orphan_api = OrphansApi(cls.client)
+        cls.orphan_cleanup_api = OrphansCleanupApi(cls.client)
         cls.task_api = TasksApi(cls.client)
 
     def test_correlation_id(self):
         """Test that a correlation can be passed as a header and logged."""
-        response, status, headers = self.orphan_api.delete_with_http_info()
+        response, status, headers = self.orphan_cleanup_api.cleanup_with_http_info({})
         monitor_task(response.task)
         task = self.task_api.read(response.task)
         self.assertEqual(headers["Correlation-ID"], self.cid)
--- a/pulpcore/tests/functional/api/test_crd_artifacts.py
+++ b/pulpcore/tests/functional/api/test_crd_artifacts.py
@@ -6,8 +6,8 @@ import unittest
 
 from pulp_smash import api, cli, config, utils
 from pulp_smash.exceptions import CalledProcessError
+from pulp_smash.pulp3.bindings import delete_orphans
 from pulp_smash.pulp3.constants import ARTIFACTS_PATH
-from pulp_smash.pulp3.utils import delete_orphans
 from requests.exceptions import HTTPError
 
 # This import is an exception, we use a file url but we are not actually using
@@ -29,7 +29,7 @@ class ArtifactTestCase(unittest.TestCase):
     def setUpClass(cls):
         """Delete orphans and create class-wide variables."""
         cfg = config.get_config()
-        delete_orphans(cfg)
+        delete_orphans()
         cls.client = api.Client(cfg, api.json_handler)
         cls.file = {"file": utils.http_get(FILE_URL)}
         cls.file_sha256 = hashlib.sha256(cls.file["file"]).hexdigest()
--- a/.github/workflows/scripts/script.sh
+++ b/.github/workflows/scripts/script.sh
@@ -111,7 +111,6 @@ export PYTHONPATH=$REPO_ROOT${PYTHONPATH:+:${PYTHONPATH}}
 
 
 if [[ "$TEST" == "upgrade" ]]; then
-  git checkout ci_upgrade_test -- pulpcore/tests/
 
   # Handle app label change:
   sed -i "/require_pulp_plugins(/d" pulpcore/tests/functional/utils.py
@@ -160,6 +159,7 @@ if [[ "$TEST" == "upgrade" ]]; then
   cd $REPO_ROOT
 
   # Running post upgrade tests
+  git checkout ci_upgrade_test -- pulpcore/tests/
   pytest -v -r sx --color=yes --pyargs -capture=no pulpcore.tests.upgrade.post
   exit
 fi
--- a/pulpcore/app/serializers/__init__.py
+++ b/pulpcore/app/serializers/__init__.py
@@ -57,6 +57,7 @@ from .importer import (  # noqa
     PulpImporterSerializer,
     PulpImportSerializer,
 )
+from .orphans import OrphansCleanupSerializer  # noqa
 from .progress import GroupProgressReportSerializer, ProgressReportSerializer  # noqa
 from .publication import (  # noqa
     ContentGuardSerializer,
--- a/pulpcore/app/urls.py
+++ b/pulpcore/app/urls.py
@@ -16,7 +16,7 @@ from rest_framework_nested import routers
 
 from pulpcore.app.apps import pulp_plugin_configs
 from pulpcore.app.views import OrphansView, PulpImporterImportCheckView, RepairView, StatusView
-from pulpcore.app.viewsets import ListRepositoryVersionViewSet
+from pulpcore.app.viewsets import ListRepositoryVersionViewSet, OrphansCleanupViewset
 from pulpcore.constants import API_ROOT
 from pulpcore.openapi import PulpSchemaGenerator
 
@@ -123,6 +123,10 @@ root_router = routers.DefaultRouter()
 urlpatterns = [
     url(r"^{api_root}repair/".format(api_root=API_ROOT), RepairView.as_view()),
     url(r"^{api_root}status/".format(api_root=API_ROOT), StatusView.as_view()),
+    url(
+        r"^{api_root}orphans/cleanup/".format(api_root=API_ROOT),
+        OrphansCleanupViewset.as_view({"post": "cleanup"}),
+    ),
     url(r"^{api_root}orphans/".format(api_root=API_ROOT), OrphansView.as_view()),
     url(
         r"^{api_root}repository_versions/".format(api_root=API_ROOT),
--- a/pulpcore/app/viewsets/__init__.py
+++ b/pulpcore/app/viewsets/__init__.py
@@ -35,6 +35,7 @@ from .importer import (  # noqa
     PulpImportViewSet,
     PulpImporterViewSet,
 )
+from .orphans import OrphansCleanupViewset  # noqa
 from .publication import (  # noqa
     ContentGuardFilter,
     ContentGuardViewSet,
--- a/pulpcore/app/viewsets/orphans.py
+++ b/pulpcore/app/viewsets/orphans.py
@@ -1,0 +1,28 @@
+from drf_spectacular.utils import extend_schema
+from rest_framework.viewsets import ViewSet
+
+from pulpcore.app.response import OperationPostponedResponse
+from pulpcore.app.serializers import AsyncOperationResponseSerializer, OrphansCleanupSerializer
+from pulpcore.app.tasks import orphan_cleanup
+from pulpcore.tasking.tasks import dispatch
+
+
+class OrphansCleanupViewset(ViewSet):
+    serializer_class = OrphansCleanupSerializer
+
+    @extend_schema(
+        description="Trigger an asynchronous orphan cleanup operation.",
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    def cleanup(self, request):
+        """
+        Triggers an asynchronous orphan cleanup operation.
+        """
+        serializer = OrphansCleanupSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        content_pks = serializer.validated_data.get("content_hrefs", None)
+
+        task = dispatch(orphan_cleanup, [], kwargs={"content_pks": content_pks})
+
+        return OperationPostponedResponse(task, request)
--- a/pulpcore/tests/functional/api/using_plugin/test_content_delivery.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_content_delivery.py
@@ -5,9 +5,9 @@ from random import choice
 from urllib.parse import urljoin
 
 from pulp_smash import api, config, utils
+from pulp_smash.pulp3.bindings import delete_orphans
 from pulp_smash.pulp3.constants import ON_DEMAND_DOWNLOAD_POLICIES
 from pulp_smash.pulp3.utils import (
-    delete_orphans,
     download_content_unit,
     gen_distribution,
     gen_repo,
@@ -53,7 +53,7 @@ class ContentDeliveryTestCase(unittest.TestCase):
         remote is recreated and another sync is triggered.
         """
         cfg = config.get_config()
-        delete_orphans(cfg)
+        delete_orphans()
         client = api.Client(cfg, api.page_handler)
 
         repo = client.post(FILE_REPO_PATH, gen_repo())
--- a/pulpcore/tests/functional/api/using_plugin/test_content_path.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_content_path.py
@@ -2,7 +2,8 @@
 import unittest
 
 from pulp_smash import api, config
-from pulp_smash.pulp3.utils import delete_orphans, gen_remote, gen_repo, sync
+from pulp_smash.pulp3.bindings import delete_orphans
+from pulp_smash.pulp3.utils import gen_remote, gen_repo, sync
 
 from pulpcore.tests.functional.api.using_plugin.constants import (
     FILE_FIXTURE_MANIFEST_URL,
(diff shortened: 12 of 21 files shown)


pulpcore/app/serializers/orphans.py
```diff
@@ -0,0 +1,12 @@
+from gettext import gettext as _
+
+from rest_framework import fields, serializers
+
+from pulpcore.app.serializers import ValidateFieldsMixin  # noqa
```
mdellweg, line 5:
```suggestion
from pulpcore.app.serializers import ValidateFieldsMixin
```
I don't see, why anyone should complain.

pulpcore/app/tasks/orphan.py
```diff
@@ -44,10 +48,21 @@ def orphan_cleanup():
         state="running",
     )
 
+    content_pks = []
+    if content_hrefs:
+        from pulpcore.app.viewsets import NamedModelViewSet
```
mdellweg, line 53: Any reason to not import this at the top?

pulpcore/app/views/orphans.py
```diff
@@ -16,8 +16,28 @@ class OrphansView(APIView):
...
         return OperationPostponedResponse(task, request)
+
+    @extend_schema(
+        description="Trigger an asynchronous task that deletes specified"
+        "orphaned content and their associated artifacts.",
+        summary="Delete orphans",
+        request=OrphansSerializer,
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    def post(self, request, format=None):
```
mdellweg, line 32: Can this functionality be added to the Delete command on the endpoint?
We are still (only) deleting stuff here (well, we create a task, but that's beyond the point).
A quick research has not shown any evidence, that a DELETE request cannot carry a body.

pulpcore/app/tasks/orphan.py
bmbouter: I tried it at the top, but there was a circular import which I couldn't easily resolve without a pretty large refactor. :(

pulpcore/app/views/orphans.py
bmbouter: This is what I also wanted to do and I tried this originally. The issue is that with DELETE the `request=OrphansSerializer` never shows up in the API schema and because of that the bindings can't be correct. Suggestions are welcome, since I also prefer this to be DELETE only and have `content_hrefs` be an optional parameter.

pulpcore/app/views/orphans.py
```diff
@@ -16,8 +16,28 @@ class OrphansView(APIView):
...
+        """
+        Cleans up specified orphaned Content and their associated Artifact in the system.
+        """
+        serializer = OrphansSerializer(data=request.data)
+        serializer.is_valid()
+
+        content_hrefs = serializer.validated_data["content_hrefs"]
```
ipanova, line 39:
```
$ http  POST :24817/pulp/api/v3/orphans/
HTTP/1.1 500 Internal Server Error
Connection: close
Content-Length: 27
Content-Type: text/html
Correlation-ID: 499b1d2ec16a42c0b6cd1af1c4b1d479
Date: Mon, 07 Jun 2021 15:47:29 GMT
Server: gunicorn
Vary: Cookie
X-Frame-Options: SAMEORIGIN

<h1>Server Error (500)</h1>


(pulp) [vagrant@pulp3-source-fedora32 _scripts]$ pk
-bash: pk: command not found
(pulp) [vagrant@pulp3-source-fedora32 _scripts]$ pjournal 
-- Logs begin at Wed 2021-03-17 21:38:37 UTC. --
Jun 07 15:47:29 pulp3-source-fedora32.lan gunicorn[17247]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/rest_framework/views.py", line 469, in handle_exception
Jun 07 15:47:29 pulp3-source-fedora32.lan gunicorn[17247]:     self.raise_uncaught_exception(exc)
Jun 07 15:47:29 pulp3-source-fedora32.lan gunicorn[17247]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
Jun 07 15:47:29 pulp3-source-fedora32.lan gunicorn[17247]:     raise exc
Jun 07 15:47:29 pulp3-source-fedora32.lan gunicorn[17247]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/rest_framework/views.py", line 506, in dispatch
Jun 07 15:47:29 pulp3-source-fedora32.lan gunicorn[17247]:     response = handler(request, *args, **kwargs)
Jun 07 15:47:29 pulp3-source-fedora32.lan gunicorn[17247]:   File "/home/vagrant/devel/pulpcore/pulpcore/app/views/orphans.py", line 39, in post
Jun 07 15:47:29 pulp3-source-fedora32.lan gunicorn[17247]:     content_hrefs = serializer.validated_data["content_hrefs"]
Jun 07 15:47:29 pulp3-source-fedora32.lan gunicorn[17247]: KeyError: 'content_hrefs'
Jun 07 15:47:29 pulp3-source-fedora32.lan gunicorn[17247]: pulp [499b1d2ec16a42c0b6cd1af1c4b1d479]: 127.0.0.1 - admin [07/Jun/2021:15:47:29 +0000] "POST /pulp/api/v3/orphans/ HTTP/1.1" 500 27 "-" "HTTPie/2.4.0"
```

You need to raise the expection when checking validity of the serializer.

pulpcore/app/tasks/orphan.py
```diff
@@ -44,10 +48,21 @@ def orphan_cleanup():
...
+
+        for href in content_hrefs:
+            content_unit = NamedModelViewSet.get_resource(href, Content)
+            content_pks.append(content_unit.pk)
+
     while True:
         content = Content.objects.filter(version_memberships__isnull=True).exclude(
             pulp_type=PublishedMetadata.get_pulp_type()
         )
+        if content_pks:
```
ipanova, line 63: this call will remove all the orphans  `http  POST :24817/pulp/api/v3/orphans/ content_hrefs:=[]`

pulpcore/app/views/orphans.py
ipanova: can we deprecate DELETE and move to POST only where:

`http  POST :24817/pulp/api/v3/orphans/` will remove all orphans
and `http  POST :24817/pulp/api/v3/orphans/ content_hrefs:=['href1, href2']` just the specified ones.

pulpcore/app/tasks/orphan.py
```diff
@@ -44,10 +48,21 @@ def orphan_cleanup():
         state="running",
     )
 
+    content_pks = []
+    if content_hrefs:
+        from pulpcore.app.viewsets import NamedModelViewSet
+
+        for href in content_hrefs:
+            content_unit = NamedModelViewSet.get_resource(href, Content)
```
ipanova, line 56: Whenever I provide an invalid content_href , task gets stuck in `running` and i see this traceback in the logs
```
Jun 07 16:21:25 pulp3-source-fedora32.lan gunicorn[21488]: pulp [551a12a9dd0b459b8ab0026d2e22630e]: 127.0.0.1 - admin [07/Jun/2021:16:21:25 +0000] "GET /pulp/api/v3/tasks/761dfa7a-647a-468a-af4f-dffb44b6c4bc/ HTTP/1.1" 200 509 "-" "HTTPi>
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]: rest_framework.exceptions.ValidationError: [ErrorDetail(string='URI /pulp/api/v3/content/container/tags/ce3740a0-e2c3-44e5-9cf9-97b1a8200816/ not found for content.', code='invalid')]
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     raise DRFValidationError(
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/home/vagrant/devel/pulpcore/pulpcore/app/viewsets/base.py", line 180, in get_resource
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     content_unit = NamedModelViewSet.get_resource(href, Content)
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/home/vagrant/devel/pulpcore/pulpcore/app/tasks/orphan.py", line 56, in orphan_cleanup
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     result = self.func(*self.args, **self.kwargs)
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/rq/job.py", line 732, in _execute
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     self._result = self._execute()
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/rq/job.py", line 709, in perform
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     job.perform()
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/rq/queue.py", line 370, in run_job
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     job = self.run_job(job)
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/rq/queue.py", line 484, in enqueue_job
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     job = self.enqueue_job(job, at_front=at_front)
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/rq/queue.py", line 366, in enqueue_call
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     return self.enqueue_call(
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/rq/queue.py", line 419, in enqueue
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     q.enqueue(
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/home/vagrant/devel/pulpcore/pulpcore/tasking/tasks.py", line 123, in _queue_reserved_task
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     result = self.func(*self.args, **self.kwargs)
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/rq/job.py", line 732, in _execute
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     self._result = self._execute()
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/rq/job.py", line 709, in perform
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     rv = job.perform()
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/rq/worker.py", line 1013, in perform_job
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]: Traceback (most recent call last):
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]: During handling of the above exception, another exception occurred:
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]: pulpcore.app.models.content.Content.DoesNotExist: Content matching query does not exist.
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     raise self.model.DoesNotExist(
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/django/db/models/query.py", line 406, in get
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     return getattr(self.get_queryset(), name)(*args, **kwargs)
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/django/db/models/manager.py", line 82, in manager_method
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:     return model.objects.get(**kwargs)
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]:   File "/home/vagrant/devel/pulpcore/pulpcore/app/viewsets/base.py", line 172, in get_resource
Jun 07 16:21:13 pulp3-source-fedora32.lan rq[21796]: pulp [058f943ad13d488ab1e6f2b259ea0adc]: rq.worker:ERROR: Traceback (most recent call last):
```

pulpcore/app/views/orphans.py
bmbouter: I'm ok with that, but I think we'll need broader agreement to do so. I put this [as an agenda item for tomorrow's pulpcore meeting](https://hackmd.io/XfI9DxymSAu6UJnblh7prQ) to reach a conclusion either way.

pulpcore/app/views/orphans.py
bmbouter: Doesn't `is_valid()` raise that exception? Anyway I'll look into it, ty.

pulpcore/app/tasks/orphan.py
bmbouter: I'll add a serializer validation to prevent this and also add a test for it.

pulpcore/app/tasks/orphan.py
bmbouter: Getting stuck in running is unexpected. I'll see if I can reproduce that because that would be very significant.

I based my implementation off of the `add_content_units` list passed to a `modify` endpoint.](https://docs.pulpproject.org/pulpcore/restapi.html#operation/repositories_file_file_modify) endpoint which also doesn't validate in the serializer. @ipanova do you think I should be validating all of these in the serializer?

pulpcore/app/views/orphans.py
bmbouter: Oh nope, I have to add `raise_exception=True`.  https://www.django-rest-framework.org/api-guide/serializers/#raising-an-exception-on-invalid-data

bmbouter: I tried to write a test for this but the bindings won't let me post without the required param being provided. So I'm adding a test that ensures it stays as required and that the bindings raise and Exception if it's omitted.

Also for the server side, it does return this now:

pulp) [vagrant@pulp3-source-fedora34 pulp_ansible]$ http POST http://localhost/pulp/api/v3/orphans/
HTTP/1.1 400 Bad Request
Allow: POST, DELETE, OPTIONS
Connection: keep-alive
Content-Length: 45
Correlation-ID: f648134759cd445ea7dfff30005b8ca6
Date: Mon, 07 Jun 2021 20:09:06 GMT
Server: nginx/1.20.0
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "content_hrefs": [
        "This field is required."
    ]
}

pulpcore/app/views/orphans.py
daviddavis: We can deprecate it but according to semver, we can't remove it until pulpcore 4.

pulpcore/app/tasks/orphan.py
bmbouter: I went ahead and added the validation in the serializer along with a test for it.

ipanova: I think the data should be validated in the serializer. did you manage to reproduce stuck 'running' task?

ipanova: actually, how about we provide a list of pks already to the orphans_cleanup, like here https://github.com/pulp/pulp_container/blob/master/pulp_container/app/viewsets.py#L609

mdellweg: I believe, that the task needs to be able to cope with missing content units.
You may be able to schedule two cleanup calls with overlapping content one after another (Or in the future even in parallel).

bmbouter: @ipanova Passing along pk's would be more efficient since validation only happens once, I'll switch to that. +1
@mdellweg the tasking code uses a chaining filter with these pk's so if they aren't present they'll just be skipped over.

CHANGES/8658.feature
```diff
@@ -0,0 +1,3 @@
+Adds POST method to ``/pulp/api/v3/orphans/`` endpoint which accepts a list of content hrefs which
```
goosemania, line 1: Can we consider having a verb here to delete orphans.
A POST to the `/pulp/api/v3/orphans/` makes me, as user, think that I'm creating orphans and not removing them from Pulp completely.
One of those to consider:
POST /pulp/api/v3/orphans/delete/
POST /pulp/api/v3/orphans/remove/
POST  /pulp/api/v3/orphans/cleanup/

Outside of scope of this PR, it would be nice to have GET available for the `/pulp/api/v3/orphans/` endpoint which will show the list of orphans.

pulpcore/app/views/orphans.py
bmbouter: Summarizing discussion from the pulpcore meeting here. The next step is for us to reach a resolution via this PR (we couldn't decide on the call). Here are the two prevailing points that stood out to me:

1. We're going to have more parameters, e.g. orphan cleanup time, so having `POST` also handle the base-case of deletion of everything makes sense.
2. If you're calling POST calling it on a verb like `delete` which would be `POST /pulp/api/v3/orphans/delete/` would be more clear than `POST /pulp/api/v3/orphans/`.
3. DELETE on a list endpoint is awkward which also further supports points (1) and (2)

I'll save a copy of this PR as-is locally, but I'm going to:
1. deprecate `DELETE /pulp/api/v3/orphans/`
2. Make `POST /pulp/api/v3/orphans/delete/` delete all orphans.
3. Make `POST /pulp/api/v3/orphans/delete/ content_hrefs=[...]` delete just the specified orphans

Please post more perspectives and feedback on this PR.

daviddavis: That sounds good to me.

ipanova:
My olny suggestion will be to use /cleanup/ and not /delete/. I think we
should stick to "orphan cleanup" since this is the terminology we use in
docs.

bmbouter:
I had also heard the `cleanup` instead of `delete` on the call. I'll use that for my changes today and we can see what folks think.

bmbouter:
I started into this, but then I realized that we should have this occur as a separate piece of work so this change can be identified more clearly on its own. I filed this https://pulp.plan.io/issues/8876 and I'm going to add to this PR as a separate commit.

bmbouter:
I've switched the implementation to use pks, and I'm getting the tests passing, but I'd like to merge this and then I'll implement the fix for https://pulp.plan.io/issues/8876 in a separate PR just after this. It'll just be easier and cleaner to do these unrelated things separately.

pulpcore/app/views/orphans.py
mdellweg: Will the operation_id in the api schema stay the same?

bmbouter: I'll have to look, but can we address in the next PR? I'm going to wait to start on that change until after this one is merged.

bmbouter:
After some matrix discussion with @x9c4 he pointed out that it would be better if this PR used `/cleanup/` right from the start and then the next PR will deprecate the other endpoint. I'm going to do that. @ipanova maybe hold off on review once more. Thanks @x9c4 and @ipanova !

ipanova:
> After some matrix discussion with @x9c4 he pointed out that it would be better if this PR used `/cleanup/` right from the start and then the next PR will deprecate the other endpoint. I'm going to do that. @ipanova maybe hold off on review once more. Thanks @x9c4 and @ipanova !

I agree, I was about to suggest the same. ty @bmbouter

ipanova:
@bmbouter  lmk when this is ready for re-review.

bmbouter:
After about 3 afternoons of work, I've literally git reset back to this PR. It's pretty frustrating. If we change the existing view code to a viewset for orphan cleanup it breaks the bindings in subtle backwards incompatible ways which luckily the tests point out. I feel like I've wasted a lot of time on all this.

bmbouter:
So right now we have this (pic below). After working on this a while I've learned about two downsides of introducing a `/pulp/api/v3/orphans/clean_up/` endpoint:

* You get two sets of bindings, with totally separate objects. Yes one of them is deprecated, but I don't think having multiple bindings is a good integrator experience.
* There is no way I know of to port this code into one viewset, without introducing backwards incompatible bindings changes. Someone who knows DRF better than I do may be able to, but after so much time spent on it I've concluded I'm not that person.

Maybe I should try keeping the same existing view and reusing it in the urls.py as `.../clean_up/` but overall this is a feeling like a big mess.

image

bmbouter:
Actually I can just make a new view, or a new viewset. The downside is that the new thing and the old thing will be totally different bindings objects. If that's what folks think is best I'll do that. @ipanova @mdellweg please give input.

mdellweg:
> Actually I can just make a new view, or a new viewset. The downside is that the new thing and the old thing will be totally different bindings objects. If that's what folks think is best I'll do that. @ipanova @mdellweg please give input.

Not sure i fully understand. But making a new & deprecating the old without further touching sounds good to me.
As i understand, you would route (urls.py) r'^/pulp/api/v3/orphans/clean_up/' to the new view(set), and then a fallback for r'^/pulp/api/v3/orphan/' to the traditional (maybe deprecated) one.
The one thing we need to make sure is that they get different `operation_id`s in the bindings. The new may default to `orphan_clean_up_create` as it is a POST call.

bmbouter:
Thanks for the reply. I want to clarify, not only will it get a new operation_id, but it will be an entirely new object. When I prototyped this before (as one viewset included in two places) I had a `OrphansApi` which has `OrphansApi.delete` and then `OrphansCleanUpApi` which has `OrphansCleanUpApi.clean_up`. Is this ok with everyone?

mdellweg:
It would be OK for me.

bmbouter:
@ipanova @mdellweg I'm happy to say this is *finally* ready for review.

mdellweg (approved):
LGTM

ipanova (approved):

pulpcore/tests/functional/api/test_correlation_id.py
```diff
@@ -23,12 +23,12 @@ def setUpClass(cls):
...
         cls.cli_client = cli.Client(cls.cfg)
 
-        cls.orphan_api = OrphansApi(cls.client)
+        cls.orphan_cleanup_api = OrphansCleanupApi(cls.client)
         cls.task_api = TasksApi(cls.client)
 
     def test_correlation_id(self):
         """Test that a correlation can be passed as a header and logged."""
-        response, status, headers = self.orphan_api.delete_with_http_info()
+        response, status, headers = self.orphan_cleanup_api.cleanup_with_http_info({})
```
mdellweg, line 31: We don't use `status` and `headers` so we should make this
```suggestion
        response = self.orphan_cleanup_api.cleanup({})
```

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -7,13 +7,13 @@
...
 from pulp_smash.exceptions import CalledProcessError
 from pulp_smash.pulp3.constants import ARTIFACTS_PATH
-from pulp_smash.pulp3.utils import delete_orphans
 from requests.exceptions import HTTPError
 
 # This import is an exception, we use a file url but we are not actually using
 # any plugin
 from pulpcore.tests.functional.api.using_plugin.constants import FILE_URL
 from pulpcore.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
+from pulpcore.tests.functional.utils import delete_orphans
```
mdellweg, line 16: There is already a bindings version of this in pulp-smash

https://github.com/pulp/pulp-smash/blob/master/pulp_smash/pulp3/bindings.py#L98

bmbouter: OK I'll switch this back to using that, but I'll need to have its implementation carry a fallback to the old-style because this new cleanup API is only available in 3.14+.

bmbouter:
I'll merge this whenever it passes: https://github.com/pulp/pulp-smash/pull/1257

PR #1397, opened 2021-06-07
Remove FileFilesystemExporter tests

daviddavis:
FileFilesystemExporter is being removed in #8861. In order to get the PR
passing, we have to remove the tests from core since pulp_file's CI runs
pulpcore tests

[noissue]


daviddavis:
I plan to add these back but against the new Exporter in core in https://github.com/pulp/pulpcore/pull/1392.

dralley (approved):

mdellweg (approved):
I'm acking it here.
Do you think the added back tests will look much different? Would it be an option to just skip them in the meantime?

daviddavis:
I think the tests will be very similar. I'll change it to skip the tests.

daviddavis:
It doesn't look like skipping the tests is an option. It still get these sorts of errors:

```
ImportError: cannot import name 'ExportersFilesystemApi' from 'pulpcore.client.pulp_file' 
```

If I try to update the import and remove `ExportersFilesystemApi`, then I will get flake8 errors in the code about `ExportersFilesystemApi` being an undefined name. And so on...

PR #1401, opened 2021-06-09
Extend retry logic to cover more types of errors

dralley:
Extend retry logic to cover all HTTP 5xx errors as well as some types of connection errors, and make the number of retries configurable by the user and plugin writer.

closes: #8881
https://pulp.plan.io/issues/8881
closes: #8867
https://pulp.plan.io/issues/8867

### DIFF
--- a/CHANGES/8867.bugfix
+++ b/CHANGES/8867.bugfix
@@ -1,0 +1,1 @@
+Retry downloads on ``ClientConnectorSSLError``, which appears to be spuriously returned by some CDNs.
--- a/CHANGES/8881.feature
+++ b/CHANGES/8881.feature
@@ -1,0 +1,1 @@
+Downloads from remote sources will now be retried on more kinds of errors, such as HTTP 500 or socket errors.
--- a/CHANGES/plugin_api/8881.deprecation
+++ b/CHANGES/plugin_api/8881.deprecation
@@ -1,0 +1,1 @@
+Plugin writers who create custom downloaders by subclassing ``HttpDownloader`` no longer need to wrap the ``_run()`` method with a ``backoff`` decorator. Consequntly the ``http_giveup`` handler the sake of the ``backoff`` decorator is no longer needed and has been deprecated. It is likely be removed in pulpcore 3.15.
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -257,6 +257,7 @@ class Remote(MasterModel):
     STREAMED = "streamed"
 
     DEFAULT_DOWNLOAD_CONCURRENCY = 10
+    DEFAULT_MAX_RETRIES = 3
 
     POLICY_CHOICES = (
         (IMMEDIATE, "When syncing, download all metadata and content now."),
@@ -293,6 +294,7 @@ class Remote(MasterModel):
     download_concurrency = models.PositiveIntegerField(
         null=True, validators=[MinValueValidator(1, "Download concurrency must be at least 1")]
     )
+    max_retries = models.PositiveIntegerField(null=True)
     policy = models.TextField(choices=POLICY_CHOICES, default=IMMEDIATE)
 
     total_timeout = models.FloatField(
--- a/pulpcore/app/serializers/repository.py
+++ b/pulpcore/app/serializers/repository.py
@@ -144,6 +144,14 @@ class RemoteSerializer(ModelSerializer):
         required=False,
         min_value=1,
     )
+    max_retries = serializers.IntegerField(
+        help_text=(
+            "Maximum number of retry attempts after a download failure. If not set then the "
+            "default value (3) will be used."
+        ),
+        required=False,
+        allow_null=True,
+    )
     policy = serializers.ChoiceField(
         help_text="The policy to use when downloading content.",
         choices=((models.Remote.IMMEDIATE, "When syncing, download all metadata and content now.")),
@@ -263,6 +271,7 @@ class RemoteSerializer(ModelSerializer):
             "pulp_labels",
             "pulp_last_updated",
             "download_concurrency",
+            "max_retries",
             "policy",
             "total_timeout",
             "connect_timeout",
--- a/pulpcore/download/http.py
+++ b/pulpcore/download/http.py
@@ -13,23 +13,45 @@ logging.getLogger("backoff").addHandler(logging.StreamHandler())
 
 
 def http_giveup(exc):
+    """Deprecated."""
+    from pulpcore.app.loggers import deprecation_logger
+
+    deprecation_logger.warning(
+        "http_giveup() is no longer necessary and has been deprecated. "
+        "It will be removed in pulpcore 3.15."
+    )
+    return http_giveup_handler(exc)
+
+
+def http_giveup_handler(exc):
     """
     Inspect a raised exception and determine if we should give up.
 
-    Do not give up when the status code is one of the following:
+    Do not give up when the error is one of the following:
 
-        429 - Too Many Requests
-        502 - Bad Gateway
-        503 - Service Unavailable
-        504 - Gateway Timeout
+        HTTP 429 - Too Many Requests
+        HTTP 5xx - Server errors
+        Socket timeout
+        TCP disconnect
+        Client SSL Error
+
+    Based on the AWS and Google Cloud guidelines:
+        https://docs.aws.amazon.com/general/latest/gr/api-retries.html
+        https://cloud.google.com/storage/docs/retry-strategy
 
     Args:
-        exc (aiohttp.ClientResponseException): The exception to inspect
+        exc (Exception): The exception to inspect
 
     Returns:
         True if the download should give up, False otherwise
     """
-    return exc.code not in [429, 502, 503, 504]
+    if isinstance(exc, aiohttp.ClientResponseError):
+        server_error = 500 <= exc.code < 600
+        too_many_requests = exc.code == 429
+        return not server_error and not too_many_requests
+
+    # any other type of error (pre-filtered by the backoff decorator) shouldn't be fatal
+    return False
 
 
 class HttpDownloader(BaseDownloader):
@@ -121,6 +143,7 @@ class HttpDownloader(BaseDownloader):
         headers_ready_callback=None,
         headers=None,
         throttler=None,
+        max_retries=None,
         **kwargs,
     ):
         """
@@ -138,6 +161,7 @@ class HttpDownloader(BaseDownloader):
                 as its values. e.g. `{'Transfer-Encoding': 'chunked'}`
             headers (dict): Headers to be submitted with the request.
             throttler (asyncio_throttle.Throttler): Throttler for asyncio.
+            max_retries (int): The maximum number of times to retry a download upon failure.
             kwargs (dict): This accepts the parameters of
                 :class:`~pulpcore.plugin.download.BaseDownloader`.
         """
@@ -154,6 +178,7 @@ class HttpDownloader(BaseDownloader):
         self.proxy_auth = proxy_auth
         self.headers_ready_callback = headers_ready_callback
         self.download_throttler = throttler
+        self.max_retries = max_retries
         super().__init__(url, **kwargs)
 
     def raise_for_status(self, response):
@@ -194,16 +219,51 @@ class HttpDownloader(BaseDownloader):
             headers=response.headers,
         )
 
-    @backoff.on_exception(
-        backoff.expo, aiohttp.ClientResponseError, max_tries=10, giveup=http_giveup
-    )
+    async def run(self, extra_data=None):
+        """
+        Run the downloader with concurrency restriction and retry logic.
+
+        This method acquires `self.semaphore` before calling the actual download implementation
+        contained in `_run()`. This ensures that the semaphore stays acquired even as the `backoff`
+        wrapper around `_run()`, handles backoff-and-retry logic.
+
+        Args:
+            extra_data (dict): Extra data passed to the downloader.
+
+        Returns:
+            :class:`~pulpcore.plugin.download.DownloadResult` from `_run()`.
+
+        """
+        retryable_errors = (
+            aiohttp.ClientConnectorSSLError,
+            aiohttp.ClientConnectorError,
+            aiohttp.ClientOSError,
+            aiohttp.ClientPayloadError,
+            aiohttp.ClientResponseError,
+            aiohttp.ServerDisconnectedError,
+            TimeoutError,
+        )
+
+        async with self.semaphore:
+
+            @backoff.on_exception(
+                backoff.expo,
+                retryable_errors,
+                max_tries=self.max_retries + 1,
+                giveup=http_giveup_handler,
+            )
+            async def download_wrapper():
+                return await self._run(extra_data=extra_data)
+
+            return await download_wrapper()
+
     async def _run(self, extra_data=None):
         """
         Download, validate, and compute digests on the `url`. This is a coroutine.
 
-        This method is decorated with a backoff-and-retry behavior to retry some errors.
-        It retries with exponential backoff 10 times before allowing a final exception to
-        be raised.
+        This method is externally wrapped with backoff-and-retry behavior for some errors.
+        It retries with exponential backoff some number of times before allowing a final
+        exception to be raised.
 
         This method provides the same return object type and documented in
         :meth:`~pulpcore.plugin.download.BaseDownloader._run`.
--- a/CHANGES/plugin_api/8881.feature
+++ b/CHANGES/plugin_api/8881.feature
@@ -1,0 +1,1 @@
+Added a field ``DEFAULT_MAX_RETRIES`` to the ``Remote`` base class - plugin writers can override the default number of retries attempted when file downloads failed for each type of remote. The default value is 3.
--- a/pulpcore/app/migrations/0066_download_concurrency_and_retry_changes.py
+++ b/pulpcore/app/migrations/0066_download_concurrency_and_retry_changes.py
@@ -16,4 +16,9 @@ class Migration(migrations.Migration):
             name='download_concurrency',
             field=models.PositiveIntegerField(null=True, validators=[django.core.validators.MinValueValidator(1, 'Download concurrency must be at least 1')]),
         ),
+        migrations.AddField(
+            model_name='remote',
+            name='max_retries',
+            field=models.PositiveIntegerField(null=True),
+        ),
     ]
--- a/pulpcore/download/factory.py
+++ b/pulpcore/download/factory.py
@@ -160,6 +160,12 @@ class DownloaderFactory:
             is configured with the remote settings.
         """
         kwargs["semaphore"] = self._semaphore
+        kwargs["max_retries"] = (
+            kwargs.get("max_retries")
+            or self._remote.max_retries
+            or self._remote.DEFAULT_MAX_RETRIES
+        )
+
         scheme = urlparse(url).scheme.lower()
         try:
             builder = self._handler_map[scheme]


pulpcore/app/models/repository.py
```diff
@@ -286,7 +288,7 @@ class Remote(MasterModel):
     proxy_username = models.TextField(null=True)
     proxy_password = models.TextField(null=True)
 
-    download_concurrency = models.PositiveIntegerField(default=10)
```
dralley, line 289: TL;DR - we represent "default value" with `0`, save that in the db instead of permanently, and then we can override what that default value is with code changes on a per-remote basis.

pulpcore/app/models/repository.py
```diff
@@ -254,6 +254,8 @@ class Remote(MasterModel):
     ON_DEMAND = "on_demand"
     STREAMED = "streamed"
 
+    DEFAULT_DOWNLOAD_CONCURRENCY = 6
```
dralley, line 257: We can leave this at 10 and then override in the RPM plugin.

pulpcore/app/models/repository.py
bmbouter: Can we make this Null being the default instead please? Since the DB and serializers can support null that would make the most sense to me.

dralley: Yes

pulpcore/app/models/repository.py
```diff
@@ -253,6 +253,8 @@ class Remote(MasterModel):
     ON_DEMAND = "on_demand"
     STREAMED = "streamed"
 
+    DEFAULT_DOWNLOAD_CONCURRENCY = 8
```
dralley, line 256: The global default is up for debate, but for the RPM plugin we will probably reduce this number to e.g. 3-6

pulpcore/app/serializers/repository.py
```diff
@@ -136,7 +136,13 @@ class RemoteSerializer(ModelSerializer):
...
     )
     download_concurrency = serializers.IntegerField(
-        help_text="Total number of simultaneous connections.", required=False, min_value=1
+        help_text=(
+            "Total number of simultaneous connections. If not provided or set to zero (0), "
+            "then a reasonable default value will be used."
+        ),
+        required=False,
+        min_value=1,
+        max_value=25,
```
dralley, line 145: Max value can be changed or dropped

pulpcore/download/http.py
```diff
@@ -195,14 +204,17 @@ async def _handle_response(self, response):
...
+        max_tries=3,
+        giveup=http_giveup,
     )
     async def _run(self, extra_data=None):
         """
         Download, validate, and compute digests on the `url`. This is a coroutine.
 
         This method is decorated with a backoff-and-retry behavior to retry HTTP 429 and
-        some 5XX errors. It retries with exponential backoff 10 times before allowing
+        some 5XX errors. It retries with exponential backoff 3 times before allowing
```
dralley, line 217: https://pulp.plan.io/issues/8881#note-5

dralley:
@jlsherrill ^

pulpcore/download/http.py
```diff
@@ -195,14 +204,21 @@ async def _handle_response(self, response):
         )
 
     @backoff.on_exception(
-        backoff.expo, aiohttp.ClientResponseError, max_tries=10, giveup=http_giveup
+        backoff.expo,
+        (
+            aiohttp.ClientResponseError,
+            aiohttp.ServerDisconnectedError,
+            aiohttp.ClientConnectorSSLError,
```
dralley, line 211: Missing anything?  https://docs.aiohttp.org/en/stable/client_reference.html#client-exceptions

pulpcore/download/http.py
```diff
@@ -195,15 +204,22 @@ async def _handle_response(self, response):
...
+        ),
+        max_tries=3,
```
dralley, line 213: There was some discussion about configurable retry count, but this seems difficult to do since it's statically configured.

Maybe instead of wrapping `_run()` with the backoff decorator, we could have `_run()` configure a closure, wrap it with the backoff decorator (using an external value for the retry count to make it configurable), and then execute that?

pulpcore/app/models/repository.py
```diff
@@ -285,7 +287,7 @@ class Remote(MasterModel):
     proxy_username = models.TextField(null=True)
     proxy_password = models.TextField(null=True)
 
-    download_concurrency = models.PositiveIntegerField(default=10)
+    download_concurrency = models.PositiveIntegerField(null=True, validators=[MinValueValidator(1)])
```
mdellweg, line 290: Is Null the default automatically? Does the validator accept Null?

pulpcore/app/serializers/repository.py
```diff
@@ -136,7 +136,13 @@ class RemoteSerializer(ModelSerializer):
         help_text="Timestamp of the most recent update of the remote.", read_only=True
     )
     download_concurrency = serializers.IntegerField(
-        help_text="Total number of simultaneous connections.", required=False, min_value=1
+        help_text=(
+            "Total number of simultaneous connections. If not provided or set to zero (0), "
+            "then a reasonable default value will be used."
+        ),
```
mdellweg, line 142: Can you delete the value by providing `null` with a PATCH call?
Also i think the helptext is outdated.

pulpcore/app/serializers/repository.py
```diff
@@ -136,7 +136,13 @@ class RemoteSerializer(ModelSerializer):
         help_text="Timestamp of the most recent update of the remote.", read_only=True
     )
     download_concurrency = serializers.IntegerField(
-        help_text="Total number of simultaneous connections.", required=False, min_value=1
+        help_text=(
+            "Total number of simultaneous connections. If not provided or set to zero (0), "
+            "then a reasonable default value will be used."
```
ipanova, line 141: what is reasonable default?

pulpcore/app/serializers/repository.py
ipanova: i'd probably vote for not having the upper limit

pulpcore/download/http.py
ipanova: what is the benefit of having this configurable? if the server has intermittent issues they should be solved within the time of 3 re-tries, if the server did not recover within that time it probably does not make sense to keep hitting on his doors.

pulpcore/app/models/repository.py
ipanova: can we add some docs with the guidance on the numbers?

pulpcore/app/serializers/repository.py
dralley: Currently, 8

ipanova: yeah I know :D , I meant to say that the way it is written is very vague  and 'reasonable' is very subjective. Can we re-phrase it to `"Total number of simultaneous connections. If not provided or set to zero (0), then the default value will be used.`

pulpcore/app/serializers/repository.py
bmbouter: I agree

pulpcore/download/http.py
bmbouter: I think different users have different needs. For example a user on a low reliability satellite link (like those gaming machines in southeast asia that use pulp), 3 may just not be enough. Yet other users who have high-reliability and syncing from reliable servers may want to disable retries. If we're going to introduce retries I think we have to make them configurable.

@dralley I was thinking the same thing. Then the retry count would be passed in as a kwarg to the `HttpDownloader.__init__` to configure the closure with.

CHANGES/8881.feature
```diff
@@ -0,0 +1 @@
+Retry logic has been added to file downloads for a wider variety of connection and server errors.
```
bmbouter, line 1: Can this call out the cases more clearly. As users read the changelog I think they could benefit from that.

CHANGES/8867.bugfix
```diff
@@ -0,0 +1 @@
+Retry downloads on ClientConnectorSSLError, which appears to be spuriously returned by some CDNs.
```
bmbouter, line 1:
```suggestion
Retry downloads on ``ClientConnectorSSLError``, which appears to be spuriously returned by some CDNs.
```

CHANGES/8878.feature
```diff
@@ -0,0 +1 @@
+The default download concurrency of 10 was found to be too high for many client types, it has been reduced to 8. Additionally, where before 'download_concurrency' would be set to a default value upon creation, it will now be set NULL (but a default value will still be used).
```
bmbouter, line 1: Can this be line wrapped.

Also s/'download_concurrency'/``download_concurrency``/

CHANGES/plugin_api/8878.feature
```diff
@@ -0,0 +1 @@
+Added a field DEFAULT_DOWNLOAD_CONCURRENCY to the Remote base class - plugin writers can override the number of concurrent downloads for each type of remote. The default value is 8.
```
bmbouter, line 1: Can this be line wrapped also.

```suggestion
Added a field ``DEFAULT_DOWNLOAD_CONCURRENCY`` to the Remote base class - plugin writers can override the number of concurrent downloads for each type of remote. The default value is 8.
```

dralley:
Blocked by https://pulp.plan.io/issues/8899

pulpcore/download/http.py
```diff
@@ -194,16 +206,46 @@ async def _handle_response(self, response):
             headers=response.headers,
         )
 
-    @backoff.on_exception(
-        backoff.expo, aiohttp.ClientResponseError, max_tries=10, giveup=http_giveup
-    )
+    async def run(self, extra_data=None, max_retries=None):
```
dralley, line 209: By putting this in ``run()`` we not only solve the problem but also avoid having plugin writers need to apply the decorator to their custom downloaders.

bmbouter: Why accept `max_retries` two ways?

dralley: We don't have to, but it seemed useful to be able to override whatever the normal value would be for one particular call.

pulpcore/download/http.py
bmbouter: These look good to me. I think more specific error cases will present themselves if necessary.

pulpcore/download/http.py
bmbouter: If it's alright I'd prefer to keep it simpler and just have one way to do it.

pulpcore/download/http.py
```diff
@@ -194,16 +208,47 @@ async def _handle_response(self, response):
...
+        """
+        retryable_errors = (
+            aiohttp.ClientResponseError,
+            aiohttp.ServerDisconnectedError,
+            aiohttp.ClientConnectorSSLError,
+        )
+
+        async with self.semaphore:
+
+            @backoff.on_exception(
```
bmbouter, line 234: We need to think about this some. Heretofore we've recommended plugin writers wrap _run() with the backoff decorator as they override `_run`. For instance here (already released) https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/downloaders.py#L75

I have to think through some options here...

bmbouter: One option (maybe?) is that we could have the existing `http_giveup` log a deprecation warning to be removed in 3.15, and always return True. Then this code could correctly provide the backoff using a newly named giveup style function, e.g. `giveup_on_certain_response_codes`. I think this would work.

dralley: It depends on what behavior it exhibits.  If the worst thing it does is retry a few extra times without actually breaking compatibility then it would be fine, IMO.

bmbouter: I agree from an outcomes perspective that is fine. I think my main concern is that if we're moving the responsibility to the `run()` method we need some controlled way to force plugin writers to stop applying the backoff to their `_run` also. So incorporating your idea here also I'm thinking we leave `http_giveup` alone, but mark as depreacted to be removed in 3.15. Then create `giveup_on_certain_response_codes` and do *not* add it to the plugin api. wdyt?

pulpcore/download/http.py
```diff
@@ -16,20 +16,31 @@ def http_giveup(exc):
...
 
     Args:
-        exc (aiohttp.ClientResponseException): The exception to inspect
+        exc (Exception): The exception to inspect
 
     Returns:
         True if the download should give up, False otherwise
     """
-    return exc.code not in [429, 502, 503, 504]
+    if isinstance(exc, aiohttp.errors.ClientResponseError):
```
bmbouter, line 37: I think this also needs to type check if its either `aiohttp.ServerDisconnectedError` or `aiohttp.ClientConnectorSSLError` and return True for those also.

bmbouter:
You mentioned something wasn't working with this PR, what was it?

dralley:
@bmbouter Instead of returning a `DownloadResult` it's returning `None`

pulpcore/download/http.py
dralley: No, it's correct but kind of confusing :)

This is the `giveup` method, which inspects the error types that are supposed to be retried, and can tell `backoff` that "no, actually you shouldn't retry this one".  And that's what `True` means here.  It's an override for sub-categories of the "retriable" error types.

We want `aiohttp.ClientConnectorSSLError` (etc.) to be retried so therefore we don't want to add any conditions for it in the `giveup` method - it should return false.

pulpcore/download/http.py
dralley: That's fine, but this is a minor enough change that I think I would rather wait a few cycles to remove it, rather than ditching it after just one.

The discussion we had today has me thinking about how should should try to balance backwards compatibility and proper refactoring.  Minor issues like this (again assuming that the worst behavior is extra retries, which is yet to be proven) don't quite reach that threshold.

Maybe we need a policy about only allowing backwards incompatible changes every 2 or 3 releases.  So if 3.14 marks something deprecated, it wouldn't be removed until 3.16 or 3.17, instead of 3.15.  And that would give plugins space to declare compatibility with more than one pulpcore release at a time.

pulpcore/download/http.py
```diff
@@ -194,16 +208,47 @@ async def _handle_response(self, response):
...
+        async with self.semaphore:
+
+            @backoff.on_exception(
+                backoff.expo,
+                retryable_errors,
+                max_tries=self.max_retries + 1,
+                giveup=http_giveup,
+            )
+            async def wrapper():
+                return await self._run(extra_data=extra_data)
```
dralley, line 241: @bmbouter This is ready for final review. My mistake was really stupid, I just forgot the return statement here, that's why it returned None.

dralley:
It works!

```
Jun 17 01:32:02 pulp3-source-fedora33.localhost.example.com pulpcore-worker[66623]: pulp [0b20ff57b9ae4ad4b93b71e2765e5f75]: backoff:INFO: Backing off wrapper(...) for 0.2s (aiohttp.client_exceptions.ClientOSError: [Errno 1] [SSL: SSLV3_ALERT_BAD_RECORD_MAC] sslv3 alert bad record mac (_ssl.c:2633))
Jun 17 01:34:18 pulp3-source-fedora33.localhost.example.com pulpcore-worker[66623]: pulp [0b20ff57b9ae4ad4b93b71e2765e5f75]: backoff:INFO: Backing off wrapper(...) for 0.4s (aiohttp.client_exceptions.ClientOSError: [Errno 1] [SSL: SSLV3_ALERT_BAD_RECORD_MAC] sslv3 alert bad record mac (_ssl.c:2633))
Jun 17 01:35:49 pulp3-source-fedora33.localhost.example.com pulpcore-worker[66623]: pulp [0b20ff57b9ae4ad4b93b71e2765e5f75]: backoff:INFO: Backing off wrapper(...) for 0.2s (aiohttp.client_exceptions.ClientPayloadError: Response payload is not completed)

```

@jlsherrill

pulpcore/download/http.py
```diff
@@ -194,16 +208,51 @@ async def _handle_response(self, response):
...
+
+        """
+        retryable_errors = (
+            aiohttp.ClientConnectorSSLError,
+            aiohttp.ClientConnectorError,
+            aiohttp.ClientOSError,
+            aiohttp.ClientPayloadError,
+            aiohttp.ClientResponseError,
+            aiohttp.ServerDisconnectedError,
+            aiohttp.TimeoutError,
```
dralley, line 233: I handled everything listed in the issue Justin filed - several of those (which were previously unhandled) I did manage to see in my testing.

pulpcore/download/http.py
dralley: @bmbouter It does seem to work fine even if wrapped a second time, the inner one fires first so the behavior is identical to if there was no decorator in the plugin.

pulpcore/app/models/repository.py
```diff
@@ -293,6 +294,7 @@ class Remote(MasterModel):
     download_concurrency = models.PositiveIntegerField(
         null=True, validators=[MinValueValidator(1, "Download concurrency must be at least 1")]
     )
+    max_retries = models.PositiveIntegerField(null=True)
```
ipanova, line 297: I am wondering whether having just `max_tries` would be more explicit rather than introducing `max_retries`?
We could have just `max_tries` which would specify the maximum number of calls to make before giving up and would have minumum value of 1 and it also would match with the `max_tries` from the backoff decorator

bmbouter: `max_tries` feels awkward to me to think about since it mixes the original request and subsequent requests. When I look at all the tools that motivated us to mimic, e.g. curl, wget, etc, they all `retry` as the way the feature is positioned to the user.

pulpcore/download/http.py
bmbouter: It will work fine, but this PR is formally moving something plugin writers used to do to pulpcore so I think we need actual change. Without it it won't be correct. For example, pulp_ansible still will use a retry of 10, but the user says 3 and actually gets 13? So my suggestion is:  deprecate `http_giveup`, introduce a new one instead and disinclude the new one from the plugin API.

CHANGES/8881.removal
```diff
@@ -0,0 +1 @@
+The ``http_giveup`` handler provided for subclasses of ``HttpDownloader`` is no longer needed and has been deprecated. Plugin writers no longer need to wrap the ``_run()`` method of custom downloaders with a ``backoff`` decorator.
```
bmbouter, line 1: This is a plugin change so it would need to go into the plugin API changelog (not the user) changelog. Also use the `deprecation` type https://github.com/pulp/pulpcore/blob/master/pyproject.toml#L45 instead of `.removal`.

pulpcore/download/http.py
```diff
@@ -13,23 +13,45 @@
 
 
 def http_giveup(exc):
+    """Deprecated."""
+    from pulpcore.app.loggers import deprecation_logger
+
+    deprecation_logger.warning(
+        "http_giveup() is no longer necessary and has been deprecated. "
+        "It will be removed in pulpcore 3.15."
```
bmbouter, line 21: Can you file an issue in redine and tag onto 3.15 for us to remove this please (if you haven't already).

pulpcore/download/http.py
```diff
@@ -13,23 +13,45 @@
...
+    )
+    http_giveup_handler(exc)
+
+
+def http_giveup_handler(exc):
```
bmbouter, line 26: :+1:

bmbouter: Doesn't the giveup handler get called for any type of retry with backoff? I may not have the correct understanding, but you indicate which types should even be considered for retry, and if they are of that type and within the retry count, and giveup handler is specified, the giveup handler is called.

I figure it's got to work that way because how could backoff know to call the giveup handler for `ClientConnectorError` but not say `ServerDisconnectedError`?

dralley:
@bmbouter https://pulp.plan.io/issues/8913

pulpcore/download/http.py
```diff
@@ -13,23 +13,45 @@
...
 
 def http_giveup(exc):
+    """Deprecated."""
+    from pulpcore.app.loggers import deprecation_logger
+
+    deprecation_logger.warning(
+        "http_giveup() is no longer necessary and has been deprecated. "
+        "It will be removed in pulpcore 3.15."
+    )
+    http_giveup_handler(exc)
```
bmbouter, line 23: Shouldn't this be?

```suggestion
    return http_giveup_handler(exc)
```

pulpcore/download/http.py
bmbouter: I see. So what you're saying makes sense, when giveup returns True it gives up, and false all other cases is that right?

pulpcore/download/http.py
dralley: Yes.  That's what I get for learning Rust, apparently.

pulpcore/app/settings.py
```diff
@@ -214,6 +214,11 @@
             "level": "INFO",
             "filters": ["correlation_id"],
         },
+        "backoff": {
+            "handlers": ["console"],
+            "level": "INFO",
+            "propagate": False,
+        },
```
bmbouter, line 221: Can you describe what this change brings?

dralley: I think nothing, the default for ``backoff`` is INFO anyway.  I can remove it, I was playing around with the logging levels

bmbouter (approved):
@dralley thank you so much for this work and for working through the zillion changes I asked for.

CHANGES/plugin_api/8881.deprecation
```diff
@@ -0,0 +1 @@
+Plugin writers who create custom downloaders by subclassing ``HttpDownloader`` no longer need to wrap the ``_run()`` method with a ``backoff`` decorator. Consequntly the ``http_giveup`` handler the sake of the ``backoff`` decorator is no longer needed and has been deprecated. It may be removed in pulpcore 3.15.
```
ipanova, line 1: s/It may be removed/It will be removed.
We just recently went though the model verbs at English lessons :hear_no_evil:

ipanova (approved):

pulpcore/app/serializers/repository.py
```diff
@@ -144,6 +144,11 @@ class RemoteSerializer(ModelSerializer):
         required=False,
         min_value=1,
     )
+    max_retries = serializers.IntegerField(
+        help_text="Maximum number of retry attempts after a download failure.",
```
ipanova, line 148: can you add that the default is 3

CHANGES/plugin_api/8881.deprecation
dralley: I only say "may" because I wasn't sure if we should keep it a little longer

ipanova (approved):
thank you! :medal_sports:

PR #1403, opened 2021-06-10
Fixes the post-upgrade test

dkliban:
The test was failing intermitently because the modify repo task was not being monitored
until completion. In cases where it did not finish quickly enough, the rest of the test
would fail.

[noissue]


bmbouter (approved):
Thanks!

dralley (approved):

PR #1406, opened 2021-06-14
Revert tasking system pr

bmbouter:
This PR being reverted broke a plugin API compatibility point. Plugins were running code like this: `Task.objects.filter(reserved_resources_record__resource='pulp_2to3_migration')` and that can't be run currently against master.


daviddavis (approved):

dralley (approved):

bmbouter:
@mdellweg FYI we need a version of these reverted commits that preserves-yet-deprecates the two models removed from the tasking system PR. I think it should still introduce the new list-based resources on the Task model directly, and have the new tasking system use that, but the old models need to still be there...

PR #1409, opened 2021-06-15
Tasking system redesign

mdellweg:


daviddavis:
FYI, I merged a pr yesterday that introduced a 0064 migration but I believe we are thinking of reverting that pr.

pulpcore/app/models/task.py
```diff
@@ -3,25 +3,28 @@
...
     AutoAddObjPermsMixin,
     AutoDeleteObjPermsMixin,
     BaseModel,
     GenericRelationModel,
 )
 from pulpcore.constants import TASK_CHOICES, TASK_FINAL_STATES, TASK_STATES
-from pulpcore.exceptions import exception_to_dict
+from pulpcore.exceptions import AdvisoryLockError, exception_to_dict
 from pulpcore.tasking.constants import TASKING_CONSTANTS
+from pulpcore.app.logging import deprecation_logger
```
goosemania, line 27: it was renamed, should be `pupcore.app.loggers`

bmbouter:
> FYI, I merged a pr yesterday that introduced a 0064 migration but I believe we are thinking of reverting that pr.

What's to motivation to revert your PR?

dralley:
@bmbouter Depending on whether it would make life easier for wibbit as he tries to migrate back onto mainline Pulp.  IIRC his production system is migrated so swapping the order of migrations is going to create a giant mess for him.

But that might be the case now anyway if we're modifying migration 0064.

mdellweg:
> @bmbouter Depending on whether it would make life easier for wibbit as he tries to migrate back onto mainline Pulp. IIRC his production system is migrated so swapping the order of migrations is going to create a giant mess for him.
> 
> But that might be the case now anyway if we're modifying migration 0064.

I am proposing here a solution without modifying the migration at all.

bmbouter:
> @bmbouter Depending on whether it would make life easier for wibbit as he tries to migrate back onto mainline Pulp. IIRC his production system is migrated so swapping the order of migrations is going to create a giant mess for him.
> 
> But that might be the case now anyway if we're modifying migration 0064.

I don't think we can take on the concern of ensuring master never has reverted migrations. Master is unreleased software, and someone who uses I think has to accept the risk. Let's not revert merged work that has nothing to do with the reverting we had to do here.

pulpcore/app/models/task.py
```diff
@@ -247,6 +208,24 @@ def save_heartbeat(self):
...
+def _hash_to_u64(value):
+    _digest = blake2s(value.encode(), digest_size=8).digest()
+    return int.from_bytes(_digest, byteorder="big", signed=True)
+
+
+class TaskManager(models.Manager):
+    def filter(self, *args, **kwargs):
+        value = kwargs.pop("reserved_resources_record__resource", None)
+        if value is not None:
+            deprecation_logger.warn(
```
bmbouter, line 220:
```suggestion
            deprecation_logger.warning(
```

nitpick: because `warn` is deprecated in Python.

dralley:
@bmbouter I understand and agree that in general we can't make that promise, but the amount of effort in this particular case is small, because there have been no migrations added for the past month, so this is the one and only conflict that would cause problems.

I had suggested yesterday to avoid any migration merges until the tasking system PR was re-merged for this reason : /

bmbouter (approved):
Thanks @mdellweg. I didn't think that just a filter(...) addition would be so simple, but I see now that it is. Great!

daviddavis (approved):

dralley (approved):
@mdellweg Neat trick! I didn't know you could do that.

mdellweg:
> @mdellweg Neat trick! I didn't know you could do that.

It's django's answer to git branches i think.

PR #1410, opened 2021-06-15
Improve handling of download concurrency

dralley:
Don't save download concurrency defaults in the database. Make it easier
for individual plugins to override.

closes: #8878
https://pulp.plan.io/issues/8878


pulpcore/app/models/repository.py
```diff
@@ -288,7 +290,7 @@ class Remote(MasterModel):
     proxy_username = models.TextField(null=True)
     proxy_password = models.TextField(null=True)
 
-    download_concurrency = models.PositiveIntegerField(default=10)
+    download_concurrency = models.PositiveIntegerField(null=True, validators=[MinValueValidator(1)])
```
dralley, line 293: @bmbouter You suggested to use `null` for unspecified but unfortunately I don't think DRF lets you make IntegerFields nullable?  Maybe it's a bug?  In any case, as per @mdellweg 's question on the other PR, it is impossible to PATCH this back to a `null` value once it has been set.

Do you think we should investigate that further or would you be OK with a default value of 0?

dralley: @daviddavis @bmbouter @mdellweg https://pulp.plan.io/issues/8899

dkliban (approved):

pulpcore/app/models/repository.py
```diff
@@ -256,6 +256,8 @@ class Remote(MasterModel):
     ON_DEMAND = "on_demand"
     STREAMED = "streamed"
 
+    DEFAULT_DOWNLOAD_CONCURRENCY = 8
```
bmbouter, line 259: I have mixed feelings about this. While it will help some, it will slow down many others. Personally I'd like more discussion before we change this.

dralley: I wouldn't object to backing it out.  The important thing is that plugin writers can decide based on the typical use cases of their plugin.

bmbouter: If that's ok I would like to. I suspect we should make a change though, I would just like to roundup all the known info on what is working and what isn't regarding this default.

bmbouter (approved):
Thanks!

PR #1411, opened 2021-06-15
Remove misleading  badges

fkuep:
#362 and https://github.com/pulp/community/discussions/26
Remove badges, since they don't reflect branches/tags .


fkuep:
It's better to have none , than misleading ones.
This is a minimal next better situation.
If somebody cares to improve the badges so they reflect respective tags and versions, 
that can be re-added again later.

daviddavis:
Could you add `[noissue]` to the commit?

PR #1413, opened 2021-06-15
[noissue]  github issue was pulp/pulpproject.org#362

fkuep:
remove misleading badges.
Can be re-instated if somebody likes to make them reflect tags/branches.
See also: https://github.com/pulp/community/discussions/26

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,3 @@
-![Pulp CI](https://github.com/pulp/pulpcore/workflows/Pulp%20CI/badge.svg)
-[![PyPI](https://img.shields.io/pypi/pyversions/pulpcore.svg)](https://pypi.python.org/pypi/pulpcore)
-
 
 **Acquire, Organize, and Distribute Software**
 


fkuep:
@daviddavis 
Should this work or is: 
``` text
re pulp/pulpproject.org#362
```
a better commit message ?

melcorr:
The commit message has to be like this @fkuep :

```
github issue was pulp/pulpproject.org#362

[noissue]
```

If you do a soft reset of the commit and then change the commit message as above, and `git push origin -f master` you don't need to raise another PR for this. 

I'm sorry, it's a bit painful first time. 
Really appreciate the PR!

daviddavis:
This commit message is fine. There’s some test that seems to be failing but I believe it’s unrelated to this PR.

daviddavis (approved):

fkuep:
> some test that seems to be failing
I saw that test ~"about versions ?" failing yesterday.
That might even have been the reason, I initially saw the badges in failed state.

bmbouter (approved):
Looks good to me. Thanks!

PR #1415, opened 2021-06-16
Update default access policies if unmodified

mdellweg:
fixes #8883

### DIFF
--- a/pulpcore/app/apps.py
+++ b/pulpcore/app/apps.py
@@ -197,7 +197,6 @@ class PulpAppConfig(PulpPluginAppConfig):
 def _populate_access_policies(sender, **kwargs):
     from pulpcore.app.util import get_view_urlpattern
 
-    print(f"Initialize missing access policies for {sender.label}.")
     apps = kwargs.get("apps")
     if apps is None:
         from django.apps import apps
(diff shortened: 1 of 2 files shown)


mdellweg:
I've been asking myself if we need to delete too, if the default policy vanished.

pulpcore/app/apps.py
```diff
@@ -197,7 +197,7 @@ def ready(self):
 def _populate_access_policies(sender, **kwargs):
     from pulpcore.app.util import get_view_urlpattern
 
-    print(f"Initialize missing access policies for {sender.label}.")
+    print(f"Adjust default access policies for {sender.label}.")
```
ipanova, line 200: `Update uncustomized default access policies`, does this sounds better by chance?

ipanova:
> I've been asking myself if we need to delete too, if the default policy vanished.

This is a valid question per se, however I don't think we plan to remove policy completely if introduced once. We can (1) cross that bridge when we come to it or (2) for the sake of completeness we can delete those for which `access_policy is None`

mdellweg:
> > I've been asking myself if we need to delete too, if the default policy vanished.
> 
> This is a valid question per se, however I don't think we plan to remove policy completely if introduced once. We can (1) cross that bridge when we come to it or (2) for the sake of completeness we can delete those for which `access_policy is None`

FWIW: When we removed a default policy, we would also change the viewset to not look for any anymore. So the one in the DB would really only be a Karteileiche (nominal member).

pulpcore/app/apps.py
mdellweg: Not sure if it's easier to understand. Someone else please decide.

ipanova (approved):
I tested this Pr alongside with my chages from rbac policy update in pulp-container and it works well.

pulpcore/app/apps.py
bmbouter: Here's my yet-another-opinion. Let's not log this at all and instead log when a change is actually applied further down.

bmbouter (approved):
I hand tested this and it worked well for me. What I did was:

1. starting with 3.13 and pup_file 1.7 on a fresh install.
2. specified customized=False, on the AccessPolicy using `shell_plus` to simulate a user edited change
3. Checked out this PR and pulp_file master
4. modified the 'tasks' policy in the code prior to applying migrations, sumulating the pulpcore shipping a change to the tasks access policy
5. applied migrations and step through the code to confirm that the 'tasks' change was not applied due to `db_accesS_policy.customized=True`.

@mdellweg if you rebase you'll get a change that was helpful for the upgrade tests. They still fail occassionally and we don't yet know why. You can also merge over that failure I think too, or rebase + rerun until it passes.

Also feel free to do whatever you want with the logging. I left a comment with an idea, but I'm ok with anything as long as the bugfix gets merged tomorrow.

PR #1416, opened 2021-06-16
Close a race condition with ReservedResource

mdellweg:
During a small window between checking it's existance and looking up the
ReservedResource, this object may be deleted in another thread.
This condition will lead to task remaining in waiting state forever.

fixes #8708
https://pulp.plan.io/issues/8708

### DIFF
--- a/CHANGES/8708.bugfix
+++ b/CHANGES/8708.bugfix
@@ -1,0 +1,3 @@
+Add an update row lock on in task dispatching for ``ReservedResource`` to prevent a race where an
+object was deleted that was supposed to be reused. This prevents a condition where tasks ended up in
+waiting state forever.
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -139,7 +139,7 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
 
                 # Attempt to lock all resources by their urls. Must be atomic to prevent deadlocks.
                 for resource in resources:
-                    if worker.reservations.filter(resource=resource).exists():
+                    if worker.reservations.select_for_update().filter(resource=resource).exists():
                         reservation = worker.reservations.get(resource=resource)
                     else:
                         reservation = ReservedResource.objects.create(


CHANGES/8708.bugfix
```diff
@@ -0,0 +1 @@
+Add an update row lock on in task dispatching for ReservedResource to prevent a race where an object was deleted that was supposed to be reused. This prevents a condition where tasks ended up in waiting state forever.
```
bmbouter, line 1: nitpick: can this be wrapped to 100 lines. Also can we use double backticks to escape code literals.

bmbouter:
Without this PR you can reproduce the issue 100% if you add a sleep with the following diff:

```
diff --git a/pulpcore/tasking/tasks.py b/pulpcore/tasking/tasks.py
index ae8a10263..52aa03513 100644
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -140,6 +140,7 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
                 # Attempt to lock all resources by their urls. Must be atomic to prevent deadlocks.
                 for resource in resources:
                     if worker.reservations.filter(resource=resource).exists():
+                        time.sleep(1)
                         reservation = worker.reservations.get(resource=resource)
                     else:
                         reservation = ReservedResource.objects.create(
```

And then run `http POST http://localhost:24817/pulp/api/v3/herminig/tasking_benchmark/ truncate_tasks=true count=5 resources_N=2 resources_K=1 sleep_secs=1` using [`pulp_herminig`](https://github.com/mdellweg/pulp_herminig).

bmbouter:
Then if you use this PR with the `sleep` still in place it reproduces 0%.  Great fix!

bmbouter (approved):
Thanks! This is great.

pulpcore/tasking/tasks.py
```diff
@@ -139,7 +139,7 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
 
                 # Attempt to lock all resources by their urls. Must be atomic to prevent deadlocks.
                 for resource in resources:
-                    if worker.reservations.filter(resource=resource).exists():
+                    if worker.reservations.select_for_update().filter(resource=resource).exists():
```
ggainey, line 142: Great catch, both on spotting the window and finding a succinct way to close it!

ggainey (approved):

dralley (approved):

PR #1418, opened 2021-06-16
Fix test that fails intermittently

daviddavis:
Sometimes the assertion returns 2 which causes the test to fail. There's
probably a test that's not properly cleaning up its repo versions.

[noissue]

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
@@ -1211,8 +1211,7 @@ class ContentInRepositoryVersionViewTestCase(unittest.TestCase):
 
         self.assertEqual(ctx.exception.status, 400)
 
-        # No repository version exists.
-        self.assertEqual(self.repo_ver_api.list().count, 0)
+        initial_rv_count = self.repo_ver_api.list(limit=1).count
 
         repo = self.repo_api.create(gen_repo())
         self.addCleanup(self.repo_api.delete, repo.pulp_href)
@@ -1248,4 +1247,4 @@ class ContentInRepositoryVersionViewTestCase(unittest.TestCase):
         # Test if repositories version with content matches.
         self.assertEqual(rv_search[0]["pulp_href"], repo.latest_version_href)
         # Test total number of repository version. Two for each repository.
-        self.assertEqual(rv_total, 4)
+        self.assertEqual(rv_total - initial_rv_count, 4)


dralley:
@daviddavis The problem is probably that `addCleanup` just doesn't wait for the task to finish.

pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
```diff
@@ -1211,8 +1211,7 @@ def test_all(self):
 
         self.assertEqual(ctx.exception.status, 400)
 
-        # No repository version exists.
-        self.assertEqual(self.repo_ver_api.list().count, 0)
+        initial_rv_count = self.repo_ver_api.list().count
```
mdellweg, line 1214:
```suggestion
        initial_rv_count = self.repo_ver_api.list(limit=0).count
```
Not sure if it really works, but if we only care about the total number there is no need to ask for any detail.

mdellweg (approved):
Anyway, i like this "difference based" version better.

ipanova:
> @daviddavis The problem is probably that `addCleanup` just doesn't wait for the task to finish.

this should fix it https://github.com/pulp/pulp-smash/blob/master/pulp_smash/pulp3/bindings.py#L17

daviddavis:
The problem is finding the test(s) that are leaving repository versions in the database.

ipanova:
> The problem is finding the test(s) that are leaving repository versions in the database.

I understand, my comment was relevant only to the cleanup part - I was mostly suggesting to switch to PulpTestCase which is actually waiting on tasks to finished in the cleanup

daviddavis:
I updated the PR and I think it's ready to merge as-is. I'm also happy to close this out if someone feels like looking into the test(s) that aren't cleaning up repo versions properly.

mdellweg:
I think we should do this anyway, as when running tests locally, i usually don't start with an empty db.

pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
```diff
@@ -1211,8 +1211,7 @@ def test_all(self):
 
         self.assertEqual(ctx.exception.status, 400)
 
-        # No repository version exists.
-        self.assertEqual(self.repo_ver_api.list().count, 0)
+        initial_rv_count = self.repo_ver_api.list(limit=0).count
```
dralley, line 1214: Wouldn't `limit=0` mess with the count?

mdellweg: I don't think so. You will always get the total count and limit only affects the pagination.

mdellweg: Actually upon further investigation, limit=0 will switch off pagination. So the best we can do here is `limit=1`.
But i can confirm, that the count will still be the total count.

daviddavis: Ok, I'll update to `limit=1`.

dralley (approved):

PR #1422, opened 2021-06-17
Close a race condition with ReservedResource

mdellweg:
During a small window between checking it's existance and looking up the
ReservedResource, this object may be deleted in another thread.
This condition will lead to task remaining in waiting state forever.

backports #8708
https://pulp.plan.io/issues/8708

fixes #8908

(cherry picked from commit 18978d85d510d50b10990f509fae5a6a07d30f67)

### DIFF
--- a/.github/workflows/scripts/before_install.sh
+++ b/.github/workflows/scripts/before_install.sh
@@ -100,7 +100,7 @@ cd ..
   pip install --upgrade --force-reinstall ./pulp-smash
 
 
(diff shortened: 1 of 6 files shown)


dralley (approved):

bmbouter:
At the katello integration, we discussed delivering this as an rpm-only patch and not doing a full upstream release for this. https://pulp.plan.io/issues/8908#note-5 I'm going to close for now so we can pursue that approach.

PR #1423, opened 2021-06-17
Add logging_cid filter to task list view

mdellweg:
fixes #8891
https://pulp.plan.io/issues/8891

This should help to find "all the tasks that were triggered by xxx frontend api call".

### DIFF
--- a/pulpcore/app/viewsets/task.py
+++ b/pulpcore/app/viewsets/task.py
@@ -32,6 +32,7 @@ class TaskFilter(BaseFilterSet):
     state = filters.ChoiceFilter(choices=TASK_CHOICES)
     worker = HyperlinkRelatedFilter()
     name = filters.CharFilter()
+    logging_cid = filters.CharFilter()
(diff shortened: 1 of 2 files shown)


pulpcore/app/viewsets/task.py
```diff
@@ -43,9 +44,10 @@ class TaskFilter(BaseFilterSet):
     class Meta:
         model = Task
         fields = {
-            "name": ["contains"],
             "state": ["exact", "in"],
             "worker": ["exact", "in"],
+            "name": ["contains"],
+            "logging_cid": ["contains"],
```
daviddavis, line 50: I wonder if it makes sense to have `exact` and `contains`?

mdellweg: Yeah, why not...
```suggestion
            "logging_cid": ["exact", "contains"],
```

daviddavis (approved):

ipanova (approved):

PR #1424, opened 2021-06-17
Make sure we close file descriptor if error is thrown

dralley:
[noissue]

### DIFF
--- a/pulpcore/app/management/commands/handle-artifact-checksums.py
+++ b/pulpcore/app/management/commands/handle-artifact-checksums.py
@@ -177,9 +177,9 @@ class Command(BaseCommand):
             for a in artifacts_qs.iterator():
                 hasher = pulp_hashlib.new(checksum)
                 try:
-                    for chunk in a.file.chunks(CHUNK_SIZE):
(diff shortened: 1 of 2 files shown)


dralley:
@ekohl

ekohl (approved):
Probably less important, but here is another example:
https://github.com/pulp/pulpcore/blob/912b022ddec8e088ad5fc9a0d2523c6933f22f88/pulpcore/app/management/commands/handle-artifact-checksums.py#L180-L182

ipanova (approved):

mdellweg (commented):
I would approve, but wasn't the "too many open files" a bug reported somewhere?
And somehow, i feel this deserves a changelog.

ekohl:
This is a follow up to 047f974e92f706c0d96baad732cc632b209d3013. I noticed it a long time after it was fixed because it showed in with customer testing so I left https://github.com/pulp/pulpcore/pull/980#issuecomment-861270329. That fixed the happy path (everything works) but this is a further security because we should never assume code doesn't fail.

mdellweg (approved):

PR #1425, opened 2021-06-17
Fix bug around reserved resource cleanup

daviddavis:
fixes #8637


pulpcore/app/models/task.py
```diff
@@ -382,11 +383,8 @@ def release_resources(self):
         for reservation in self.reserved_resources.all():
             TaskReservedResource.objects.filter(task=self.pk).delete()
             if not reservation.tasks.exists():
-                try:
+                with suppress(IntegrityError):
                     reservation.delete()
-                except IntegrityError:
-                    # other tasks have added reservations for this resource
-                    pass
```
daviddavis, line 389: This except block got introduced in https://pulp.plan.io/issues/8603. My theory about why an `IntegrityError` was getting raised was that this sequence was happening:

1. `reservation.delete()` gets called
2. django checks if there are any reservations (for the cascade delete). there are none.
3. another task then reserves the resource
4. django calls delete in the database which causes fk violation and raises the `IntegrityError`.

pulpcore/app/models/task.py
```diff
@@ -382,11 +383,8 @@ def release_resources(self):
         for reservation in self.reserved_resources.all():
             TaskReservedResource.objects.filter(task=self.pk).delete()
             if not reservation.tasks.exists():
-                try:
+                with suppress(IntegrityError):
```
daviddavis, line 386: PROTECT raises a `ProtectedError` which is a subclass of `IntegrityError`.

pulpcore/app/models/task.py
daviddavis: So I think it's wise to be safe and continue to check for `IntegrityError` even though we could potentially just check for `ProtectedError`.

dralley (approved):

pulpcore/app/models/task.py
```diff
@@ -382,11 +383,8 @@ def release_resources(self):
         for reservation in self.reserved_resources.all():
             TaskReservedResource.objects.filter(task=self.pk).delete()
```
mdellweg, line 384: Why do we call this inside the loop?
We could use something like
`list(TaskReservedResource.objects.filter(task=id).values_list("resource_id"))`
to know what needs cleanup afterwards.

pulpcore/app/models/task.py
```diff
@@ -382,11 +383,8 @@ def release_resources(self):
         for reservation in self.reserved_resources.all():
```
mdellweg, line 383: Can we change the name of this variable to `reserved_resource`? I think, it's confusing.

mdellweg (approved):
I think, this is ready as stated in the ticket.
But i still left some suggestions.

pulpcore/app/models/task.py
daviddavis: What's the benefit of not calling it inside the loop? We have to delete each ReservedResource individually since if one is protected, it will cause the entire delete to fail.

mdellweg: We need the loop, yes. But this command will produce an empty query in the second iteration.

daviddavis: I see what you mean. I filed an issue to fix this: https://pulp.plan.io/issues/8943

bmbouter:
This looks good to me also! I see two lgtm's I'm merging.

PR #1427, opened 2021-06-21
Fall back to on_demand if artifact download fails

dralley:
closes: #8873
https://pulp.plan.io/issues/8873

This needs some discussion.  See the issue: Pulp 2 handles content download failures (referring to non-retriable ones like 403 forbidden, 404 not found) by skipping over downloading the content.  Pulp 3 does not do so.

Should we, or should we close this issue?  And if we should, then should we have strict and non-strict mode?

I agree with Justin here but I'm pretty sure we've seen repositories with missing packages before.
```

<jsherrill> gotcha
<jsherrill> yeah it seems rare that a paricular package you'd get a 403, but the rest of the repo is fine
```

### DIFF
--- a/CHANGES/8873.feature
+++ b/CHANGES/8873.feature
@@ -1,0 +1,1 @@
+Files which are unable to be downloaded will fall back to "on_demand" and log a warning rather than causing a sync to fail.
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -196,10 +196,19 @@ class ArtifactDownloader(Stage):
             and not d_artifact.deferred_download
             and not d_artifact.artifact.file
         ]
+        downloaded_count = 0
         if downloaders_for_content:
-            await asyncio.gather(*downloaders_for_content)
+            results = await asyncio.gather(*downloaders_for_content, return_exceptions=True)
+            for d_artifact, result in zip(d_content.d_artifacts, results):
+                if isinstance(result, Exception):
+                    log.warn(
+                        "Download of {} failed due to error: {}.".format(d_artifact.url, result)
+                    )
+                    d_artifact.deferred_download = True
+                else:
+                    downloaded_count += 1
         await self.put(d_content)
-        return len(downloaders_for_content)
+        return downloaded_count
 
 
 class ArtifactSaver(Stage):


bmbouter:
We do need to solve this (counterproposal below) but I don't think this fallback is the right approach. The decision to keep going is a subjective one, and given that the user indicated they want all artifacts to be downloaded I think we need to adhere to what they tell us. Another practical reason to be strict on this has to do with import/export, which can only work on repos where all Artifacts are already downloaded, so if a user is syncing for the purposes of import export, this type of fallback would cause the workflow to go past where the root cause has occurred (at sync time) only for the issue to present itself later (at export time).

We also have this issue here too https://pulp.plan.io/issues/5286

I looked in the tracker for where I wrote this before, but I'll summarize here since I couldn't find it. I propose we have a remote option on the master Remote called `continue_on_http_response_codes` or some similar name. This would accept a list of integers, e.g. `[404]`, or `[403, 404]` and would fall back in http responses for those cases. This way we aren't enabling desired functionality for some set of users at the expense of another.

dralley:
I don't know about requiring a specific list of HTTP reponse codes to bypass, but a concept of `continue_on_error` sounds exactly like the strict / non-strict modes I suggested

bmbouter:
> I don't know about requiring a specific list of HTTP reponse codes to bypass, but a concept of `continue_on_error` sounds exactly like the strict / non-strict modes I suggested

Actually that sounds better. Users likely don't care why they can't have content, just that they should continue anyway. Let's do that.

dralley:
Question: `continue_on_error` seems like something that should be a sync-time option rather than a remote option.  It's more verb-y than even `max_retries` or `download_concurrency`.

But there's not really a good way to get the value from the viewset to the internals of the pipeline

mdellweg:
> Question: `continue_on_error` seems like something that should be a sync-time option rather than a remote option. It's more verb-y than even `max_retries` or `download_concurrency`.
> 
> But there's not really a good way to get the value from the viewset to the internals of the pipeline. Is that something we should try, or should we just add it to the remote?

I can see it as a parameter on the remote. It makes the remote "failsafe". And you can still have a second remote that fails regularly with the same upstream URL. In the end, the Remote is kind of a "sync parameter collection".

bmbouter:
I also think on the remote. My hope is that in the long-term all remote options can be generically overridden as sync options.

bmbouter:
To move this forward, I think we need this brought back to an issue and this PR changed to either draft or closed. @dralley what do you think about that?

dralley:
Sounds fine.  I'll move to draft for now.

ipanova:
> Question: `continue_on_error` seems like something that should be a sync-time option rather than a remote option. It's more verb-y than even `max_retries` or `download_concurrency`.
> 
> But there's not really a good way to get the value from the viewset to the internals of the pipeline. Is that something we should try, or should we just add it to the remote?

+1 to add this to the remote

dralley:
There's also no way to access the remote from `DeclarativeVersion` :(  I think we should put this one on ice until we have time to do a refactor of `DeclarativeVersion`, something we will eventually want to do anyway to better support metadata mirroring.

PR #1428, opened 2021-06-22
Fix bad default value

dralley:
The default value isn't used by normal code due to downloader factory,
but RPM plugin unit tests construct the downloader manually and expect it to not be null.

[noissue]


bmbouter (approved):
I think this makes sense.

daviddavis (approved):

PR #1432, opened 2021-06-22
Separate Architecture from Deployment docs

melcorr:
Move hw & fs info into installation section 
As a user, I expect to find hw requirements before I go to install something. 
To me, it is not architecture.
I cannot tell the difference between deployment & installation in this context either, so removed "deployment". 

fixes #8947

### DIFF
--- a/docs/installation/hardware-requirements.rst
+++ b/docs/installation/hardware-requirements.rst
@@ -1,0 +1,99 @@
+
+Hardware requirements
+=====================
+
+Pulp de-duplicates content, and makes as efficient use of storage space as possible. Even if you
+configure Pulp not to store content, it will still require some local storage.
+
+.. note::
+
+   This section is updated based on your feedback. Feel free to share what your experience is
+   https://pulpproject.org/help/
+
+.. note::
+
+   These are empirical guidelines to give an idea how to estimate what you need. It hugely
+   depends on the scale of the setup (how much content you need, how many repositories you plan
+   to have), frequency (how often you run various tasks) and the workflows (which tasks you
+   perform, which plugin you use) of each specific user.
+
+CPU
+***
+
+CPU count is recommended to be equal to the number of pulp workers. It allows to perform N
+repository operations concurrently. E.g. 2 CPUs, one can sync 2 repositories concurrently.
+
+RAM
+***
+
+Out of all operations the highest memory consumption task is likely synchronization of a remote
+repository. Publication can also be memory consuming, however it depends on the plugin.
+
+For each worker, the suggestion is to plan on 1GB to 3GB. E.g. 4 workers would need 4GB to 12 GB
+For the database, 1GB is likely enough.
+
+The range for the workers is quite wide because it depends on the plugin. E.g. for RPM plugin, a
+setup with 2 workers will require around 8GB to be able to sync large repositories. 4GB is
+likely not enough for some repositories, especially if 2 workers both run sync tasks in parallel.
+
+Disk
+****
+
+For disk size, it depends on how one is using Pulp and which storage is used.
+
+
+Empirical estimation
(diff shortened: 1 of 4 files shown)


docs/installation/hardware-requirements.rst
```diff
@@ -0,0 +1,103 @@
...
+   This section is updated based on your feedback. Feel free to share what your experience is https://pulpproject.org/help/
+
+.. note::
+
+   These are empirical guidelines to give an idea how to estimate what you need. It hugely
+   depends on the scale of the setup (how much content you need, how many repositories you plan
+   to have), frequency (how often you run various tasks) and the workflows (which tasks you
+   perform, which plugin you use) of each specific user.
+
+Pulp behaviour
```
mdellweg, line 16: Not sure if that is well suited here...

melcorr: I find it so ill-suited in all locations. 
I was copying and pasting it into different places! 
Any ideas? Should I just delete? I find it so strange and saying something but also nothing.

docs/installation/hardware-requirements.rst
```diff
@@ -0,0 +1,103 @@
...
+ * Any additional content, one plans to upload to or import into Pulp, needs to be counted as well.
+
+ * DB size needs to be taken into account as well.
+
+E.g. For syncing remote repositories with ``on_demand`` policy and using local storage, one
+would need 50GB + 30% of size of all the repository content + the DB.
+
+.. _filesystem-layout:
+
+Filesystem Layout
```
mdellweg, line 79: Also not sure if this belongs here. Maybe with a comment like "To better estimate diskspace, this table explains where stuff goes."

melcorr: I agree. This section says nothing about what it is useful for. I will add that line. Thanks!

docs/installation/hardware-requirements.rst
mdellweg: I agree.

* Pulp de-duplicates content.
That's a general feature. It's important, because you don't waste too much diskspace.
* There are different policies for downloading content. It is possible not to store any content
   at all.
That is also a feature, but maybe download policies are already described better elsewhere. Maybe that comment serves as a note in the diskspace section.
* If plugin needs to generate metadata for a repository, it will be in the artifact storage,
   even if the download policy is configured not to save any content.
Really don't know what to do with this one. "Even if you don't intend to save anything, you need a (working) storage. Maybe the same type of note again.
* Pulp verifies downloaded artifact checksums locally and artifacts are downloaded/verified in
   parallel, so some local storage is needed, even if the download policy is configured not to save
   any content and an external storage, like S3, is used.
And again: You opt for saving elsewhere? You still need _some_ amount of local storage... again such a note, maybe with less detail of why? (you cannot do anything about it anyway.)

docs/installation/hardware-requirements.rst
```diff
@@ -0,0 +1,99 @@
+
+Hardware requirements
+=====================
+
+Pulp de-duplicates content, and makes as efficient use of storage space as possible. Even if you
```
melcorr, line 5: @mdellweg this is my attempt at summarizing the "Pulp behaviour" section. 
I've removed that as following your comments, it didn't make sense to keep it.

docs/installation/hardware-requirements.rst
```diff
@@ -0,0 +1,99 @@
...
+configure Pulp not to store content, you still require some local storage.
```
mdellweg, line 6:
```suggestion
configure Pulp not to store content, it will still require some local storage.
```

mdellweg (approved):

bmbouter (approved):
This looks good to me. Thank you @melcorr !

PR #1433, opened 2021-06-22
Add subheadings and simplify conceptual intro

melcorr:
Just trying to reduce jargon and make this easier to read. 
It definitely needs a David Davis job done on it. Low priority though. 

fixes #8949

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -1,46 +1,70 @@
 Concepts and Terminology
 ========================
 
-This introduction presents a high level overview of Pulp terminology and concepts. It is designed
-to be understandable to anyone who is familiar with software management even without prior
-knowledge of Pulp. This document favors clarity and accuracy over ease of reading.
+This introduction is designed for anyone who is familiar with software management even without prior
+knowledge of Pulp.
 
-From a user’s perspective Pulp is a tool to manage their content. In this context, “Pulp” refers to
-“pulpcore and one or more plugins”. Because of its dependent relationship with plugins, pulpcore
-can be described as a framework for plugin development.
+From a user’s perspective, Pulp is a tool to manage content. In this context, _Pulp_ refers to
+pulpcore and one or more content plugins.
 
-:term:`pulpcore` is a generalized backend with a REST API and a plugin API. Users will need at
-least one :term:`plugin` to manage :term:`Content`.  Each :term:`type` of content unit (like rpm or
-deb) is defined by a plugin.  Files that belong to a content unit are called
-:term:`Artifacts<Artifact>`. Each content unit can have 0 or many artifacts and artifacts can be
+What is pulpcore?
+-----------------
+
+Throughout Pulp documentation, when you see references to _pulpcore_, this term refers to the main
+python program that provides a platform to which you add content plugins for the types of content
+that you want to manage. In a very general sense, Pulpcore refers to Pulp and its functionality
+without any plugins. Pulpcore provides a REST API and a Plugin API.
+
+Content Management with plugins
+-------------------------------
+
+To manage content, you need at least one content plugin. Each type of content unit, for example RPM
+or Debian, is defined by a plugin. For example, if you want to manage RPM content in Pulp, you must
+install the RPM plugin. Files that belong to a content unit are called
+:term:`Artifacts<Artifact>`. Each content unit can have zero or many artifacts. Artifacts can be
 shared by multiple content units.
 
 .. image:: ./_diagrams/concept-content.png
     :align: center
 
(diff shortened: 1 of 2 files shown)


daviddavis:
This change looks great. Thank you. :+1: 

Can you wrap the lines at 100 characters?

melcorr:
> Can you wrap the lines at 100 characters?

Sure, will do first thing tomorrow. Thank you!

melcorr:
I hope that did it... @daviddavis - I set my editor to wrap at 100 chars but who knows...

daviddavis:
I think it wrapped the lines longer than 100 chars but did not combine lines shorter than 100 chars.

melcorr:
> I think it wrapped the lines longer than 100 chars but did not combine lines shorter than 100 chars.

lol! OK !

daviddavis:
What editor are you using? If you're using vim, you can highlight text and use `gq`.

melcorr:
> What editor are you using? If you're using vim, you can highlight text and use `gq`.

I write a lot so use Atom. 
There are different rules for Foreman docs, so need to mess around :)

melcorr:
> > What editor are you using? If you're using vim, you can highlight text and use `gq`.
> 
> I write a lot so use Atom.
> There are different rules for Foreman docs, so need to mess around :)

What's the logic behind the wrapping at 100 chars btw? 
Is this not just another obstacle to put in the way of a contributor? ;-)

daviddavis:
> What's the logic behind the wrapping at 100 chars btw?
> Is this not just another obstacle to put in the way of a contributor? ;-)

I'm curious about what you would suggest in terms of formatting?

I think the reasoning behind wrapping at 100 is that markup languages like markdown and rst are meant to be read both rendered and unrendered. Like you can look at any of these in github, view diffs, etc and easily read it without having it rendered. And 100 chars fits on pretty much any screen without having to scroll over. 

Personally I would prefer 80 chars since it's pretty standard and I often use split panes in vim with about 160 chars being the max that fits on my laptop screen.

daviddavis:
By the way, vim is great for writing:

https://www.naperwrimo.org/wiki/index.php?title=Vim_for_Writers
https://nickjanetakis.com/blog/vim-is-saving-me-hours-of-work-when-writing-books-and-courses
https://hackernoon.com/vim-for-writers-ee15d2a8f512

daviddavis (approved):

melcorr:
Just wanted some context. 
I've no wisdom to add. TBH I'm quite surprised I haven't encountered this before. 
The only requirement when moving all of the Red Hat Satellite docs out into the Foreman community was from ATIX. They requested that the docs have one sentence per line. 
So I'll jump between the two settings, or use VIM for Pulp docs. 

I've never heard of VIM for nanowrimo ;) 
Thanks for the links. I'll swot up!

daviddavis:
> one sentence per line.

giphy

mdellweg:
The one sentence per line rule makes reading diffs easier.
If you change a single word, and it re-wraps the whole paragraph it may easily take an additional minute to figure that the diff actually reflects a rather trivial change.
In the end mark down is source code (for rendered texts) and imo should be treated that way.
And there is still the option to start a new sentence on a new line _and_ wrap at 100 chars.

daviddavis:
That's true: having a linebreak after each sentence makes diffs smaller when text is changed. To me though, as someone who mostly edits and reads these docs unrendered, it detracts from the readability and I think being able to edit and read the docs unrendered is one of the main reasons we use rst and markdown over other markup languages like html. But this is just my biased opinion. Feel free to overrule me.

docs/concepts.rst
```diff
@@ -1,46 +1,70 @@
...
-From a user’s perspective Pulp is a tool to manage their content. In this context, “Pulp” refers to
-“pulpcore and one or more plugins”. Because of its dependent relationship with plugins, pulpcore
-can be described as a framework for plugin development.
-
-:term:`pulpcore` is a generalized backend with a REST API and a plugin API. Users will need at
-least one :term:`plugin` to manage :term:`Content`.  Each :term:`type` of content unit (like rpm or
-deb) is defined by a plugin.  Files that belong to a content unit are called
-:term:`Artifacts<Artifact>`. Each content unit can have 0 or many artifacts and artifacts can be
+This introduction is designed for anyone who is familiar with software management even without prior
+ knowledge of Pulp.
```
bmbouter, line 5: tiny nitpick
```suggestion
knowledge of Pulp.
```

melcorr: Thakns for your review @bmbouter !

bmbouter (approved):
Thanks!

PR #1434, opened 2021-06-22
Add a remove-plugin command

### LINKED ISSUE #1945: As a user, I can remove a Pulp plugin
Author: @jlsherrill (jsherril@redhat.com)

Redmine Issue: 7822, https://pulp.plan.io/issues/7822

---

### Motivation
There are cases when a user might want to remove a plugin:
 - plugin is no longer in use but it's data in the db is quite large (pulp-2to3-migration case)
 - incompatible plugins

### Deliverables

 - all plugin related data should be removed from the DB, its schema as well
 - artifacts can be left for orphan cleanup, don't need to be removed immediately
 - it should be possible to install this plugin back again, make sure django migrations are applied in this case accordingly

### Original request from Katello

> After a pulp2 to 3 migration, the left over migration data is quite large and needs to be cleaned up.  there needs to be a command or api to clean it up.
> 
> In addition we need a way/instructions to remove the plugin altogether and delete its added schema entirely

goosemania:
closes #1945
Overall approach and related notes are in this doc https://hackmd.io/cKZ-MwkdSUiFKJo0W5zQ_A.

### DIFF
--- a/docs/workflows/plugin-removal.rst
+++ b/docs/workflows/plugin-removal.rst
@@ -1,0 +1,38 @@
+.. _plugin-removal:
+
+Plugin removal
+--------------
+
+.. note::
+
+    Plugin removal feature is in tech-preview and may change in backwards incompatible
+    ways in future releases.
+
+
+In case when one does not need a certain Pulp plugin anymore or there is a plugin which is no
+longer supported and having it installed may prevent from upgrading to the latest releases for
+pulpcore and other Pulp plugins.
+
+Pulp provides an ability to remove its plugins. It might be needed for the following reasons:
+  * a plugin is no longer needed
+  * a plugin is no longer supported (it can block further pulpcore upgrades if maintainers no
+    longer update the plugin to be compatible with the latest pulpcore)
+
+Plugins can be removed one at a time using the `pulpcore-manager` command `remove-plugin`. In this
+example the File plugin is removed:
+
+
+.. code-block::
+
+    $ pulpcore-manager remove-plugin file
+
+
+As a result, all the data related to the plugin will be removed from the Pulp database.
+It is possible to install back the removed plugin if desired and if it's compatible with the
+pulpcore version being used.
+
+
+.. note::
+
+    After `remove-plugin` command has succeeded, the plugin needs to be uninstalled manually.
+    Steps to uninstall depend on how it was originally installed.
--- a/pulpcore/app/apps.py
+++ b/pulpcore/app/apps.py
@@ -100,7 +100,7 @@ class PulpPluginAppConfig(apps.AppConfig):
         post_migrate.connect(
             _populate_access_policies,
             sender=self,
-            dispatch_uid="polulate_access_policies_identifier",
+            dispatch_uid="populate_access_policies_identifier",
         )
         post_migrate.connect(_populate_roles, sender=self, dispatch_uid="populate_roles_identifier")
 
--- a/pulpcore/app/management/commands/remove-plugin.py
+++ b/pulpcore/app/management/commands/remove-plugin.py
@@ -1,0 +1,201 @@
+import time
+
+from gettext import gettext as _
+
+from django.apps import apps
+from django.db import connection, IntegrityError
+from django.db.migrations.exceptions import IrreversibleError
+from django.db.models.signals import post_migrate
+from django.conf import settings
+from django.contrib.contenttypes.models import ContentType
+from django.core.management import BaseCommand, call_command, CommandError
+
+from pulpcore.app.apps import pulp_plugin_configs
+from pulpcore.app.models import AccessPolicy, ContentAppStatus, Worker
+from pulpcore.app.models.role import Role
+from pulpcore.app.util import get_view_urlpattern
+
+DROP_PLUGIN_TABLES_QUERY = """
+DO $$
+  BEGIN
+    EXECUTE format('DROP TABLE %s',
+                    (SELECT STRING_AGG(table_name, ', ')
+                       FROM information_schema.tables
+                         WHERE table_schema = 'public' AND table_name like '{app_label}_%'
+                    )
+                  );
+  END
+$$;
+"""  # noqa
+
+
+class Command(BaseCommand):
+    """
+    Django management command for removing a plugin.
+
+    This command is in tech-preview.
+    """
+
+    help = _(
+        "[tech-preview] Disable a Pulp plugin and remove all the relevant data from the database. "
+        "Destructive!"
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "plugin_name",
+            help=_("Name of a plugin to remove. E.g. file, container, rpm, pulp_2to3_migration."),
+        )
+
+    def _check_pulp_services(self):
+        """
+        Check if any pulp services are running and error out if they are.
+        """
+        is_pulp_running = True
+        waiting_time = max(settings.CONTENT_APP_TTL, settings.WORKER_TTL)
+        check_started = time.time()
+        self.stdout.write(
+            _("Checking if Pulp services are running, it can take up to {}s...").format(
+                waiting_time
+            )
+        )
+        while is_pulp_running and (time.time() - check_started) < waiting_time:
+            is_pulp_running = (
+                ContentAppStatus.objects.online().exists()
+                or Worker.objects.online_workers().exists()
+            )
+            time.sleep(2)
+
+        if is_pulp_running:
+            raise CommandError(
+                _(
+                    "The command can't be used when Pulp services are running. "
+                    "Please stop the services: pulpcore-api, pulpcore-content and all "
+                    "pulpcore-worker@*."
+                )
+            )
+
+    def _remove_indirect_plugin_data(self, app_label):
+        """
+        Remove plugin data not accessible via plugin models.
+
+        Specifically,
+            - remove django content type by app_label (also auth permissions are removed by cascade)
+            - remove default access policies related to the plugin with provided app_label
+            - remove locked roles related to the plugin, do not touch the user defined ones.
+        """
+        ContentType.objects.filter(app_label=app_label).delete()
+        app_config = apps.get_app_config(app_label)
+        viewset_names = []
+        role_names = []
+        for viewset_batch in app_config.named_viewsets.values():
+            for viewset in viewset_batch:
+                viewset_names.append(get_view_urlpattern(viewset))
+                role_names.extend(getattr(viewset, "LOCKED_ROLES", {}).keys())
+
+        AccessPolicy.objects.filter(viewset_name__in=viewset_names, customized=False).delete()
+        Role.objects.filter(name__in=role_names, locked=True).delete()
+
+    def _remove_plugin_data(self, app_label):
+        """
+        Remove all plugin data.
+
+        Removal happens via ORM to be sure that all relations are cleaned properly as well,
+        e.g. Master-Detail, FKs to various content plugins in pulp-2to3-migration.
+
+        In some cases, the order in which models are removed matters, e.g. FK is a part of
+        uniqueness constraint. Try to remove such problematic models later.
+        """
+
+        models_to_delete = set(apps.all_models[app_label].values())
+        prev_model_count = len(models_to_delete) + 1
+        while models_to_delete and len(models_to_delete) < prev_model_count:
+            # while there is something to delete and something is being deleted on each iteration
+            removed_models = set()
+            for model in models_to_delete:
+                self.stdout.write(_("Removing model: {}").format(model))
+                try:
+                    model.objects.filter().delete()
+                except IntegrityError:
+                    continue
+                else:
+                    removed_models.add(model)
+
+            prev_model_count = len(models_to_delete)
+            models_to_delete = models_to_delete - removed_models
+
+        if models_to_delete:
+            # Never-happen case
+            raise CommandError(
+                _(
+                    "Data for the following models can't be removed: {}. Please contact plugin "
+                    "maintainers."
+                ).format(list(models_to_delete))
+            )
+
+        self._remove_indirect_plugin_data(app_label)
+
+    def _drop_plugin_tables(self, app_label):
+        """
+        Drop plugin table with raw SQL.
+        """
+        with connection.cursor() as cursor:
+            cursor.execute(DROP_PLUGIN_TABLES_QUERY.format(app_label=app_label))
+
+    def _unapply_migrations(self, app_label):
+        """
+        Unapply migrations so the plugin can be installed/run django migrations again if needed.
+
+        Make sure no post migration signals are connected/run (it's enough to disable only
+        `populate_access_policy` and `populate_roles` for the requested plugin, so after
+        migration is run, policies are not repopulated but there is no need for any of
+        post_migrate operations to happen.)
+
+        Then, try to unmigrate the clean way, and if it fails, fake it until you make it.
+        A potential reason for the failure can be that some migrations are irreversible.
+        """
+        for app_config in pulp_plugin_configs():
+            post_migrate.disconnect(
+                sender=app_config, dispatch_uid="populate_access_policies_identifier"
+            )
+            post_migrate.disconnect(sender=app_config, dispatch_uid="populate_roles_identifier")
+            if app_config.label == "core":
+                post_migrate.disconnect(sender=app_config, dispatch_uid="delete_anon_identifier")
+
+        try:
+            call_command("migrate", app_label=app_label, migration_name="zero")
+        except (IrreversibleError, Exception):
+            # a plugin has irreversible migrations or some other problem, drop the tables and fake
+            # that migrations are unapplied.
+            self._drop_plugin_tables(app_label)
+            call_command("migrate", app_label=app_label, migration_name="zero", fake=True)
+
+    def handle(self, *args, **options):
+        plugin_name = options["plugin_name"]
+        if plugin_name == "core":
+            raise CommandError(_("Please specify a plugin name, core can't be removed."))
+
+        available_plugins = {app.label for app in pulp_plugin_configs()} - {"core"}
+        if plugin_name not in available_plugins:
+            raise CommandError(
+                _(
+                    "Plugin name is incorrectly specified or plugin is not installed. Please "
+                    "specify one of the following plugin names: {}."
+                ).format(list(available_plugins))
+            )
+
+        self._check_pulp_services()
+
+        self.stdout.write(_("Cleaning up the database for {} plugin...").format(plugin_name))
+        self._remove_plugin_data(app_label=plugin_name)
+
+        self.stdout.write(_("Unapplying {} plugin migrations...").format(plugin_name))
+        self._unapply_migrations(app_label=plugin_name)
+
+        self.stdout.write(
+            _(
+                "Successfully removed the {} plugin data. It is ready to be uninstalled. "
+                "NOTE: Please do uninstall, otherwise `pulp status` might not show you the correct "
+                "list of plugins available."
+            ).format(plugin_name)
+        )
--- a/pulpcore/tests/functional/api/using_plugin/test_remove_plugin.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_remove_plugin.py
@@ -1,0 +1,34 @@
+import pytest
+
+from pulp_smash.pulp3.utils import gen_repo
+
+
+@pytest.mark.nightly
+def test_remove_plugin(
+    cli_client,
+    delete_orphans_pre,
+    file_fixture_gen_file_repo,
+    file_repo_api_client,
+    start_and_check_services,
+    stop_and_check_services,
+):
+    repo_name = "repo for plugin removal test"
+    file_repo_pre_removal = file_repo_api_client.create(gen_repo(name=repo_name))
+
+    assert stop_and_check_services() is True
+
+    res = cli_client.run(["pulpcore-manager", "remove-plugin", "file"])
+    assert "Successfully removed" in res.stdout
+
+    # Without uninstalling the package just run migrations again to mimic the reinstallation
+    # of a plugin at least from pulp's perspective
+    res = cli_client.run(["pulpcore-manager", "migrate", "file"])
+    assert res.stdout.endswith("updated.\n") is True
+
+    assert start_and_check_services() is True
+
+    # create a repo with the same name as before the removal
+    file_repo_post_reinstall = file_fixture_gen_file_repo(name=repo_name)
+
+    assert file_repo_pre_removal.name == file_repo_post_reinstall.name
+    assert file_repo_pre_removal.pulp_href != file_repo_post_reinstall.pulp_href
--- a/CHANGES/1945.feature
+++ b/CHANGES/1945.feature
@@ -1,0 +1,1 @@
+Added a pulpcore-manager command `remove-plugin` to remove a Pulp plugin.
--- a/docs/workflows/index.rst
+++ b/docs/workflows/index.rst
@@ -22,4 +22,5 @@ assumes that the reader is familiar with the fundamentals discussed in the :doc:
    reclaim-disk-space
    import-export
    labels
+   plugin-removal
    troubleshooting
--- a/pulpcore/tests/conftest_pulp_file.py
+++ b/pulpcore/tests/conftest_pulp_file.py
@@ -123,3 +123,13 @@ def file_fixture_gen_remote_client_cert_req(
         return gen_object_with_cleanup(file_remote_api_client, kwargs)
 
     yield _file_fixture_gen_remote_client_cert_req
+
+
+@pytest.fixture
+def file_fixture_gen_file_repo(file_repo_api_client, gen_object_with_cleanup):
+    """A factory to generate a File Repository with auto-deletion after the test run."""
+
+    def _file_fixture_gen_file_repo(**kwargs):
+        return gen_object_with_cleanup(file_repo_api_client, kwargs)
+
+    yield _file_fixture_gen_file_repo


pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def _remove_indirect_plugin_data(self, app_label):
+        """
+        It feels like we are reinventing a wheel, tbh, I can't believe it can be that hard to
+        remove an app from Django.
+        TODO:
+          - remove from auth_permission by contettype_id (cascade delete
+          will hopefully take care of the assigned perms, needs testing)
+          - remove from django_content_type by app_label
+          - all default access policies are in core_accesspolicy, can be identified by view_name?
```
goosemania, line 41: Any thought/ideas are welcome

pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -0,0 +1,150 @@
...
+
+        models_to_delete = set(apps.all_models[app_label].values())
+        prev_model_count = len(models_to_delete) + 1
+        while models_to_delete and len(models_to_delete) < prev_model_count:
+            # while there is something to delete and something is being deleted on each iteration
+            removed_models = set()
+            for model in models_to_delete:
+                self.stdout.write(_("Removing model: {}").format(model))
+                try:
+                    model.objects.filter().delete()
```
goosemania, line 64: This causes an unexpected problem. Any thoughts are welcome.
Setup: 
 - have pulp_rpm and pulp_file
 - remove pulp_rpm
 - removal of pulp_file fails due to relations(?) to the pulp_rpm models (I suspect some django smartness combined with master-detail models)

Here how the failure looks:
```$ pulpcore-manager remove-plugin file
pulp [None]: pulp_2to3_migration.app.plugin:INFO: Plugin pulp_deb is not installed in pulp3 therefore it will not be migrated from pulp2
Checking if Pulp services are running, it can take up to 30s...
Cleaning up the database for file plugin...
Removing model: <class 'pulp_file.app.models.FileContent'>
Removing model: <class 'pulp_file.app.models.FileRemote'>
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UndefinedTable: relation "rpm_rpmrepository" does not exist
LINE 1: ...heck", "rpm_rpmrepository"."sqlite_metadata" FROM "rpm_rpmre...
                                                             ^


The above exception was the direct cause of the following exception:

  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/management/__init__.py", line 375, in execute
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/management/base.py", line 323, in run_from_argv
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/management/base.py", line 364, in execute
  File "/home/vagrant/devel/pulpcore/pulpcore/app/management/commands/remove-plugin.py", line 133, in handle
    self._remove_plugin_data(app_label=plugin_name)
  File "/home/vagrant/devel/pulpcore/pulpcore/app/management/commands/remove-plugin.py", line 51, in _remove_plugin_data
    model.objects.filter().delete()
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/db/models/query.py", line 710, in delete
    collector.collect(del_query)
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/db/models/deletion.py", line 223, in collect
    elif sub_objs:
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/db/models/query.py", line 278, in __bool__
    self._fetch_all()
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/db/models/query.py", line 1242, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/db/models/query.py", line 55, in __iter__
    results = compiler.execute_sql(chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size)
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/db/models/sql/compiler.py", line 1142, in execute_sql
    cursor.execute(sql, params)
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/db/backends/utils.py", line 67, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/db/backends/utils.py", line 76, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/db/utils.py", line 89, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
django.db.utils.ProgrammingError: relation "rpm_rpmrepository" does not exist
LINE 1: ...heck", "rpm_rpmrepository"."sqlite_metadata" FROM "rpm_rpmre...

```

The same happens if you try to remove any FileRemote instance when rpm plugin is removed/uninstalled.
If you look at FileRemote instance, it has some managers/relations/attributes available which has rpm in it.
```
In [4]: dir(my_file_remote_instance)

Out[4]:
[...
 'remote_ptr',
 'remote_ptr_id',
 'remoteartifact_set',
 'repository_set',
 'rpm_rpmremote',
 'rpm_rpmrepository',
 'rpm_ulnremote',
...]
```
This happened when I tried to remove one plugin after another without uninstalling python package. Need to test if it's still teh case with python package installed.

dralley: Does it still happen post-restart?

mdellweg: I think you either need to pip uninstall the plugin or find a way to stop pulpcore from autoloading it after it was "removed".

goosemania: I'm not sure it is safe to pip uninstall, what if it was an rpm install?

mdellweg: right then it should be a rpm uninstall accordingly.
What i meant is, that you'd need to get the source files out of the system, or a way for pulp to not consider loading them.

goosemania:
@bmbouter , @mdellweg , @dralley , could you provide feedback for the approach in general?
I'm waiting on it to proceed with implementing and polishing the details. 
It would be helpful to get feedback today or early next week, since next week is the last one before my pto and it would be good to wrap this work up to the extent possible.
It's not critical though but if you have time, I'd appreciate it.

pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -0,0 +1,150 @@
...
+        Removal happens via ORM to be sure that all relations are cleaned properly as well,
+        e.g. Master-Detail, FKs to various content plugins in pulp-2to3-migration.
+
+        In some cases, the order in which models are removed matters, e.g. FK is a part of
+        uniqueness constraint. Try to remove such problematic models later.
+        """
+
+        models_to_delete = set(apps.all_models[app_label].values())
+        prev_model_count = len(models_to_delete) + 1
+        while models_to_delete and len(models_to_delete) < prev_model_count:
```
mdellweg, line 58: The greedy algorithm ;)

mdellweg:
Yes, i can see this approach going forward.

pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -0,0 +1,150 @@
...
+            # unapplied
+
+            # TODO: maybe catch any error, because if unapplying of non-atomic migrations fails for
+            #  some reason, migrations will be in a bad state, can't apply back but failing should
+            #  help. E.g. pulp_file 0009 migration can be half unapplied if pulp-2to3-migration
+            #  is installed and references pulp_file objects.
+
+            # TODO: drop plugin tables with raw SQL
+
+            call_command("migrate", app_label=app_label, migration_name="zero", fake=True)
```
bmbouter, line 104: nice

pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def handle(self, *args, **options):
+        plugin_name = options["plugin_name"]
+        available_plugins = {app.label for app in pulp_plugin_configs()} - {"core"}
```
bmbouter, line 108: This could also raise a `CommandError` above here if `"core"`.

pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -0,0 +1,150 @@
...
+
+        is_pulp_running = True
+        waiting_time = max(settings.CONTENT_APP_TTL, settings.WORKER_TTL)
+        check_started = time.time()
+        self.stdout.write(
+            _("Checking if Pulp services are running, it can take up to {}s...").format(
+                waiting_time
+            )
+        )
+        while is_pulp_running and time.time() - check_started < waiting_time:
```
bmbouter, line 125: nice

pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -0,0 +1,150 @@
...
+                ContentAppStatus.objects.online().exists()
+                or Worker.objects.online_workers().exists()
+            )
+            time.sleep(2)
+
+        if is_pulp_running:
+            raise CommandError(
+                _(
+                    "The command can't be used when Pulp services are running. "
+                    "Please stop the services."
```
bmbouter, line 136: Maybe identifying the services by name would be helpful here? Just a thought.

pulpcore/app/management/commands/remove-plugin.py
bmbouter: I think it can search through all the viewsets similar to how this code does it: https://github.com/pulp/pulpcore/blob/master/pulpcore/app/apps.py#L204-L218 Roughly the idea is to find the `viewset_name` and then call delete on the `AccessPolicy`. Let me know if I can give more info here.

pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -0,0 +1,150 @@
...
+        parser.add_argument(
+            "plugin_name",
+            help=_("Name of a plugin to remove. E.g. file, container, rpm, pulp_2to3_migration."),
+        )
+
+    def _remove_indirect_plugin_data(self, app_label):
+        """
+        It feels like we are reinventing a wheel, tbh, I can't believe it can be that hard to
+        remove an app from Django.
+        TODO:
```
bmbouter, line 37: Is the plan to resolve all of these before merging?

bmbouter:
Overall this seems like the most reasonable/feasible approach. Thank you for putting this together!

pulpcore/app/management/commands/remove-plugin.py
goosemania: Definitely

pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -0,0 +1,150 @@
...
+        if plugin_name not in available_plugins:
+            raise CommandError(
+                _(
+                    "Plugin name is incorrectly specified or plugin is not installed. Please "
+                    "specify one of the following plugin names: {}."
+                ).format(list(available_plugins))
+            )
+
+        is_pulp_running = True
+        waiting_time = max(settings.CONTENT_APP_TTL, settings.WORKER_TTL)
```
goosemania, line 118: A note for myself. Settings are optional(?).

```
$ pulpcore-manager remove-plugin rpm
pulp: pulp_2to3_migration.app.plugin:INFO: Plugin pulp_deb is not installed in pulp3 therefore it will not be migrated from pulp2
System check identified some issues:

WARNINGS:
?: (guardian.W001) Guardian authentication backend is not hooked. You can add this in settings as eg: `AUTHENTICATION_BACKENDS = ('django.contrib.auth.backends.ModelBackend', 'guardian.backends.ObjectPermissionBackend')`.
  File "/bin/pulpcore-manager", line 11, in <module>
    load_entry_point('pulpcore==3.7.6', 'console_scripts', 'pulpcore-manager')()
  File "/usr/lib/python3.6/site-packages/pulpcore/app/manage.py", line 11, in manage
  File "/usr/lib/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/lib/python3.6/site-packages/django/core/management/__init__.py", line 375, in execute
  File "/usr/lib/python3.6/site-packages/django/core/management/base.py", line 323, in run_from_argv
  File "/usr/lib/python3.6/site-packages/django/core/management/base.py", line 364, in execute
  File "/usr/lib/python3.6/site-packages/pulpcore/app/management/commands/remove-plugin.py", line 118, in handle
    waiting_time = max(settings.CONTENT_APP_TTL, settings.WORKER_TTL)
  File "/usr/lib/python3.6/site-packages/dynaconf/base.py", line 164, in __getattr__
    value = getattr(self._wrapped, name)
AttributeError: 'Settings' object has no attribute 'WORKER_TTL'
```

bmbouter: They got introduced later, in newer versions they'll always be there https://github.com/pulp/pulpcore/blob/master/pulpcore/app/settings.py#L230 (maybe this was already obvious).

goosemania:
In my test, it took ~55 sec to remove RPM plugin with the following amount of migrated data to Pulp3:
> db.units_erratum.count()
4680
> db.units_rpm.count()
13608
> db.repos.count()
1209

On the same system, removing pulp-2to3-migration took ~35 sec.

This doesn't look like a large system, I'd say it's a small one.
What is the acceptable downtime for removing a plugin?

I doubt people will remove rpm plugin, but still, it feels like the system can be down for minutes.
Any thoughts? @bmbouter @dralley  @mdellweg 
Should we require services to be down or can we do some pre-work to require shutting them down later or for a short period of time?
E.g. disable plugin somehow and restart the services, then can safely remove in the background.

pulpcore/app/management/commands/remove-plugin.py
goosemania: I can't do any uninstallation because I do not know how it was installed and I might not have the right perms. User needs to uninstall the plugin themselves.
I seem not to be able to reproduce this problem anymore, it might be that now when all the proper cleanup is happening, it's no longer a problem. I'll keep an eye on it though.

pulpcore/app/management/commands/remove-plugin.py
goosemania: I'm going to list all services,  even if not all of them are running, would that work for you? not sure what is the best way to check if pulpcore-api is running without requiring bindings to be installed (to make an api request and see if it works) or without using `ps`.

pulpcore/app/apps.py
```diff
@@ -98,7 +98,7 @@ def ready(self):
         self.import_urls()
         self.import_modelresources()
         post_migrate.connect(
-            _populate_access_policies, sender=self, dispatch_uid="my_unique_identifier"
+            _populate_access_policies, sender=self, dispatch_uid="populate_access_policy"
```
goosemania, line 101: I thought I'd give this identifier a more meaningful name since I use it. Let me know if there was a purpose to keep it vague.

pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -0,0 +1,192 @@
...
+        A potential reason for the failure can be that some migrations are irreversible.
+        """
+        for app_config in pulp_plugin_configs():
+            post_migrate.disconnect(sender=app_config, dispatch_uid="populate_access_policy")
+            if app_config.label == "core":
+                post_migrate.disconnect(sender=app_config, dispatch_uid="delete_anon_identifier")
+
+        try:
+            call_command("migrate", app_label=app_label, migration_name="zero")
+        except (IrreversibleError, Exception):
```
goosemania, line 158: The intention here is to show that it was mostly for irreversible errors but actually any exception needs to be handled so user do not end up with a non-atomic migration half-applied/unaplied and have no easy way out.
There is no need to have 2 exceptions listed here from the programming point of view. Let me know if it's less confusing to keep a general `Exception` only.

goosemania:
This awaits a test to be written, other than that it is ready for review.

pulpcore/app/apps.py
mdellweg: No. It was copied from the django example, i think.

pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -141,13 +141,17 @@ def _unapply_migrations(self, app_label):
...
+        are not repopulated but there is no need for any of post_migrate operations to happen.)
+
         Then, try to unmigrate the clean way, and if it fails, fake it until you make it.
         A potential reason for the failure can be that some migrations are irreversible.
         """
-        app_config = apps.get_app_config(app_label)
-        post_migrate.disconnect(sender=app_config, dispatch_uid="my_unique_identifier")
-        post_migrate.disconnect(dispatch_uid="delete_anon_identifier")
+        for app_config in pulp_plugin_configs():
+            post_migrate.disconnect(sender=app_config, dispatch_uid="populate_access_policy")
```
mdellweg, line 152: Do you think instead of disconnecting we could try to make the signal handler look for some indicator (existing table, ..., ???) and actually remove the stale access_policies from db?

goosemania: That's a good idea if we have a reliable way to identify those

mdellweg: I tried to find an easy way to see it the applied migration was the last one for the plugin. No luck yet.

goosemania:
Awaits the Roles PR in order to address roles removal.
LOCKED_ROLES related to a plugin should be removed, the rest should not be touched.
Documentation should be clear that custom roles will lose any permissions which belong to the removed plugin. In the extreme case it is possible that the custom role will remain but have no permissions. It's a responsibility of a user to deal with that.

goosemania:
> Awaits the Roles PR in order to address roles removal. LOCKED_ROLES related to a plugin should be removed, the rest should not be touched. Documentation should be clear that custom roles will lose any permissions which belong to the removed plugin. In the extreme case it is possible that the custom role will remain but have no permissions. It's a responsibility of a user to deal with that.

Awaits roles usage in container plugin to be able to test against.

pulpcore/tests/functional/management/using_plugin/test_remove_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+    # Importing here, otherwise it is recognised as a separate test to run
+    from pulpcore.tests.functional.api.using_plugin.test_sync import test_http_sync_no_ssl
+
+    test_http_sync_no_ssl(
+        delete_orphans_pre,
+        file_fixture_gen_remote,
+        file_repo,
+        file_repo_api_client,
+        content_file_api_client,
+    )
```
goosemania, line 48: I have this weird plugin removal test where I'm removing a plugin, then adding it back, and then I'd like to run a test to check that the plugin is functional.
I would like to reuse an existing test and so far I do not see a good way to do it.
If I use fixtures and I pass them as parameters to my test, the records in the db are created before I removed the plugin (aka cleaned the db), so by the time I'd like to call the existing test, my fixtures data is outdated, it would be good to force it to be recreated again. Is there a way? The smallest scope is a test itself, and I need to recreate it within a test :/

As a hack, I was thinking to split the test into multiple ones and enforce the order with [a pytest hook](https://docs.pytest.org/en/6.2.x/_modules/_pytest/hookspec.html#pytest_collection_modifyitems) but it's so hacky that I have not even tried to do that.

This plugin removal test needs to go into every plugin, so I thought to add it to the plugin template, leaving it for a plugin writer to decide which test(s) they want to run after the plugin is installed back. So it would be good to have this PR/test as a showcase for other plugins.

Any thoughts or ideas? @bmbouter @mdellweg

pulpcore/tests/functional/management/using_plugin/test_remove_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+    svc_mgr.stop(pulp_services)
+    # Wait for services to stop
+    sleep(2)
+
+    res = cli.run(["pulpcore-manager", "remove-plugin", "file"])
+    assert "Successfully removed" in res.stdout
+
+    # Without uninstalling the package just run migrations again to mimic the reinstallation
+    # of a plugin at least from pulp's perspective
+    # call_command("migrate", "file")
```
goosemania, line 30: leftovers, needs to be removed

pulpcore/tests/functional/management/using_plugin/test_remove_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+    res = cli.run(["pulpcore-manager", "migrate", "file"])
+    assert res.stdout.endswith("updated.\n") is True
+
+    svc_mgr.start(pulp_services)
+    # Wait for services to start
+    sleep(2)
+
+    # Run any tests here to make sure that everything works as expected.
+    # Importing here, otherwise it is recognised as a separate test to run
+    from pulpcore.tests.functional.api.using_plugin.test_sync import test_http_sync_no_ssl
```
bmbouter, line 40: I think if you `from foo import x as y` and it didn't start with `test_` I believe it would not be collected.

pulpcore/tests/functional/management/using_plugin/test_remove_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # Without uninstalling the package just run migrations again to mimic the reinstallation
+    # of a plugin at least from pulp's perspective
+    # call_command("migrate", "file")
+    res = cli.run(["pulpcore-manager", "migrate", "file"])
+    assert res.stdout.endswith("updated.\n") is True
+
+    svc_mgr.start(pulp_services)
+    # Wait for services to start
+    sleep(2)
```
bmbouter, line 36: Do we have a utility to monitor via the status API when pulp is functional? That would be ideal. If not maybe we make one? It would be a nice pytest fixture for pulp-smash. Not required in this PR tho.

pulpcore/tests/functional/management/using_plugin/test_remove_plugin.py
bmbouter: I'm a little unclear on if the code here is working and you are looking for a better way or if this is not working at all. Also when you say "the records in the db are created before...." what records are these?

pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -0,0 +1,201 @@
...
+class Command(BaseCommand):
+    """
+    Django management command for removing a plugin.
+
+    This command is in tech-preview.
+    """
+
+    help = _(
+        "Disable a Pulp plugin and remove all the relevant data from the database. "
+        "Destructive! [tech-preview]"
```
dralley, line 41: Suggestion: put `[tech-preview]` first

pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -0,0 +1,201 @@
...
+        """
+        is_pulp_running = True
+        waiting_time = max(settings.CONTENT_APP_TTL, settings.WORKER_TTL)
+        check_started = time.time()
+        self.stdout.write(
+            _("Checking if Pulp services are running, it can take up to {}s...").format(
+                waiting_time
+            )
+        )
+        while is_pulp_running and time.time() - check_started < waiting_time:
```
dralley, line 62: Can we clarify the order of operations with `()`?

pulpcore/tests/functional/management/using_plugin/test_remove_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+    file_fixture_gen_remote,
+    file_repo,
+    file_repo_api_client,
+    content_file_api_client,
+):
+    cfg = smash_cfg.get_config()
+    cli = smash_cli.Client(cfg)
+    pulp_host = cfg.hosts[0]
+    svc_mgr = smash_cli.ServiceManager(cfg, pulp_host)
+    pulp_services = ("pulpcore-content", "pulpcore-api", "pulpcore-worker@1", "pulpcore-worker@2")
```
dralley, line 20: This feels like something we ought to have a utility for enumerating

pulpcore/tests/functional/management/using_plugin/test_remove_plugin.py
dralley: +1, although the bindings might make it easy enough to get by without?  I haven't seen an example of it, our tests don't use it https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/api/test_status.py

pulpcore/tests/functional/management/using_plugin/test_remove_plugin.py
goosemania: > "the records in the db are created before...."

Fixture data is generated at the beginning of the test `test_remove_plugin`, e.g. a file repo with uuid XXXX is created in the db. By the time I'm trying to execute the test `test_http_sync_no_ssl`, I've already removed and installed back the plugin, so there are no db records for that plugin. At the same time fixture data is not regenerated, it is cached and outdated, so `test_http_sync_no_ssl` fails because it can't find a file repo which the fixture provides.

goosemania: @bmbouter , fyi, @mdellweg  and I had [a discussion on matrix](https://matrix.to/#/!aVApiNMtnstWbwDcVU:matrix.org/$9bKEdSbYbIBSvZYcWin5dY5CyAp6Wvi3-vVL4JVIA4M?via=libera.chat&via=matrix.org&via=x9c4.spdns.org) that the test looks like it's following a pattern close to the upgrade one, and we might need to have something similar for this test.

pulpcore/tests/functional/management/using_plugin/test_remove_plugin.py
goosemania: could you elaborate?

dralley: I mean we have potentially many pulpcore-workers and we want to start and stop all of them, this makes the assumption that the installation only has two.  If we can't use a wildcard to start / stop all worker services, we should probably enumerate them somehow.

I don't consider this a blocking issue though.

pulpcore/tests/functional/api/using_plugin/test_remove_plugin.py
```diff
@@ -0,0 +1,37 @@
+# import pytest
+
+from pulp_smash import cli as smash_cli, config as smash_cfg
+from pulp_smash.pulp3.utils import gen_repo
+
+cfg = smash_cfg.get_config()
+cli = smash_cli.Client(cfg)
```
mdellweg, line 7: We have session fixtures for these now. ;)

https://github.com/pulp/pulp-smash/blob/main/pulp_smash/pulp3/pytest_plugin/__init__.py#L246

mdellweg (approved):

docs/workflows/plugin-removal.rst
```diff
@@ -0,0 +1,38 @@
...
+    Plugin removal feature is in tech-preview and may change in backwards incompatible
+    ways in future releases.
+
+
+In case when one does not need a certain Pulp plugin anymore or there is a plugin which is no
+longer supported and having it installed may prevent from upgrading to the latest releases for
+pulpcore and other Pulp plugins.
+
+Pulp provides an ability to remove its plugins. It might be needed for the following reasons:
+  * a plugin is no longer need
```
dralley, line 17: s/need/needed

pulpcore/app/apps.py
```diff
@@ -100,7 +100,7 @@ def ready(self):
         post_migrate.connect(
             _populate_access_policies,
             sender=self,
-            dispatch_uid="polulate_access_policies_identifier",
```
dralley, line 103: Is this a separate issue that needs to be backported?  Or would we want to backport this whole feature.

pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -0,0 +1,201 @@
...
+        ContentType.objects.filter(app_label=app_label).delete()
+        app_config = apps.get_app_config(app_label)
+        viewset_names = []
+        role_names = []
+        for viewset_batch in app_config.named_viewsets.values():
+            for viewset in viewset_batch:
+                viewset_names.append(get_view_urlpattern(viewset))
+                role_names.extend(getattr(viewset, "LOCKED_ROLES", {}).keys())
+
+        AccessPolicy.objects.filter(viewset_name__in=viewset_names, customized=False).delete()
```
dralley, line 96: I'm not familiar with AccessPolicy, what happens with `customized=True`?

dralley (changes requested):

pulpcore/app/management/commands/remove-plugin.py
goosemania: `customized=True` is set for access policies created by users, we are not touching them, it's their business to deal with them.

pulpcore/app/apps.py
mdellweg: It was an identifier with and without the typo. ;)

dralley: +1

docs/workflows/plugin-removal.rst
goosemania: pushing

dralley (approved):

pulpcore/tests/functional/api/using_plugin/test_remove_plugin.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+
+from pulp_smash.pulp3.utils import gen_repo
+
+
+@pytest.mark.nightly
+def test_remove_plugin(
```
bmbouter, line 7: I'm really happy with how this turned out. Thanks!

PR #1436, opened 2021-06-23
Set USE_NEW_WORKER_TYPE=True by default

mdellweg:
fixes #8948
https://pulp.plan.io/issues/8948


CHANGES/8948.feature
```diff
@@ -0,0 +1,6 @@
+Activate the new task worker type by default.
+
+.. warning::
+
+   If you intend to stick with the old tasking system, you should configure the
+   ``USE_NEW_WORKER_TYPE`` setting to false before upgrading.
```
mdellweg, line 6: Didn't really know where to put this.
The changelog is part of the release notes was my thought.

docs/configuration/settings.rst
```diff
@@ -314,8 +314,6 @@ ALLOWED_CONTENT_CHECKSUMS
 USE_NEW_WORKER_TYPE
 ^^^^^^^^^^^^^^^^^^^
 
-    .. note:: This new worker-type is still in tech-preview.
```
mdellweg, line 317: IMHO we cannot make this switch _and_ say it's tech preview.

bmbouter (approved):
This looks great!

mdellweg:
I need to adjust/remove the duplicated ci jobs here.

pulpcore/app/settings.py
```diff
@@ -235,7 +235,7 @@
 
 ALLOWED_EXPORT_PATHS = []
 
-USE_NEW_WORKER_TYPE = False
+USE_NEW_WORKER_TYPE = True
```
dralley, line 238: Should should rename this setting to `USE_LEGACY_WORKER_TYPE`?

mdellweg: This name is used in several other repositories already. That's why i never renamed it.

docs/configuration/settings.rst
bmbouter: +1 to removing as tech-preview.

I think adding a  paragraph about switching from the old to the new and from the new to the old would be good. Users need guidance on the details about if they need to drain their tasks or not. Perhaps a `pulp-cli` command to show the number of tasks that need to be cancelled or maybe a `pulpcore-manager shell -c ...` command would help users know if switching is safe.

dralley (approved):

docs/configuration/settings.rst
mdellweg: cli won't work, because you first shut down the api-workers.

dralley (approved):

bmbouter:
The new changes look good. @mdellweg please merge whenever the tests pass. Thank you!

CHANGES/8948.feature
ipanova: maybe add a note in the settings.rst that the default is set to true

dralley: +1 to that suggestion

mdellweg:
I do not understand the upgrade test failing here.

bmbouter:
From what I've seen and heard the upgrade test has intermittent failures unrelated to any piece of work. Given that, I think this should merge without that passing.

PR #1438, opened 2021-06-24
Fix OpenAPI schema tag generation

dkliban:
The tag generation code was deleting the second element from the end instead of deleting the
second element in the list of operation keys. This worked for operations that had 3 elements
in the operations keys list, however, anything longer produced wrong operation tags.

fixes: #8868
https://pulp.plan.io/issues/8868

### DIFF
--- a/CHANGES/8868.bugfix
+++ b/CHANGES/8868.bugfix
@@ -1,0 +1,9 @@
+Fixed OpenAPI schema tag generation for resources that are nested more than 2 levels.
+
+This change is most evident in client libraries generated from the OpenAPI schema.
+
+Prior to this change, the API client for a resource located at
+`/api/v3/pulp/exporters/core/pulp/<uuid>/exports/` was named `ExportersCoreExportsApi`.
+
+After this change, the API client for a resource located at
+`/api/v3/pulp/exporters/core/pulp/<uuid>/exports/` is named `ExportersPulpExportsApi`.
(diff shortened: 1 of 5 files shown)


mdellweg:
Can you add some examples of correctly and incorrectly translated operation tags?
Does this affect the operation_ids?

dkliban:
The examples of changes to the bindings are available in the test changes.

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
```diff
@@ -23,7 +23,7 @@
 from pulpcore.client.pulpcore import (
     ApiClient as CoreApiClient,
     ExportersPulpApi,
-    ExportersCoreExportsApi,
+    ExportersPulpExportsApi,
```
daviddavis, line 26: So this swaps `Core` for `Pulp`? What happens if say we have two exporters of the same type but in different plugins:

```
/exporters/file/whatever/
/exporters/rpm/whatever/
```

dkliban: Since we generate bindings for each plugin separate they would be:

`pulpcore.client.pulp_file.ExportersWhateverApi`
`pulpcore.client.pulp_rpm.ExportersWhateverApi`

daviddavis (approved):

CHANGES/8868.bugfix
```diff
@@ -0,0 +1 @@
+Fixed OpenAPI schema tag generation.
```
daviddavis, line 1: Some more detail or perhaps an example here would be good.

daviddavis: Also, this change perhaps needs a `removal` entry in addition or in lieu of this change since we have a breaking change.

dkliban: Yep ... and I'll also add a note for plugin authors.

mdellweg (approved):

PR #1439, opened 2021-06-24
Deprecated `DELETE /pulp/api/v3/orphans/`

bmbouter:
When calling this endpoint now, a deprecation warning is logged
recommending users instead use the `POST /pulp/api/v3/orphans/cleanup/`
call.

closes #8876

### DIFF
--- a/CHANGES/8876.deprecation
+++ b/CHANGES/8876.deprecation
@@ -1,0 +1,2 @@
+Deprecated the ``DELETE /pulp/api/v3/orphans/`` call. Instead use the
+``POST /pulp/api/v3/orphans/cleanup/`` call.
--- a/pulpcore/app/views/orphans.py
+++ b/pulpcore/app/views/orphans.py
@@ -1,6 +1,7 @@
 from drf_spectacular.utils import extend_schema
 from rest_framework.views import APIView
 
+from pulpcore.app.loggers import deprecation_logger
 from pulpcore.app.response import OperationPostponedResponse
 from pulpcore.app.serializers import AsyncOperationResponseSerializer
 from pulpcore.app.tasks import orphan_cleanup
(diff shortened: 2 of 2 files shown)


bmbouter:
This shows up in the docs as 
image

Additionally, when called, the tests log `pulpcore.deprecation:WARNING: The `DELETE /pulp/api/v3/orphans/` call is deprecated. Use`POST /pulp/api/v3/orphans/cleanup/` instead.`.

bmbouter:
@ipanova @mdellweg This is the one just for deprecation. Also review this please.

dralley (approved):

ipanova (approved):

ipanova:
@bmbouter  can you update smash to use the new endpoint? https://github.com/pulp/pulp-smash/search?q=delete_orphans

ipanova:
also i found this testcase that probably would be good to update https://github.com/pulp/pulpcore/blob/master/pulpcore/tests/functional/api/test_correlation_id.py#L31

mdellweg (approved):

bmbouter:
> @bmbouter can you update smash to use the new endpoint? https://github.com/pulp/pulp-smash/search?q=delete_orphans

Actually my goal is to have us stop using pulp smash. Using pulp smash prevents us from using the bindings directly, which is my goal. I'd rather update Pulp to never call orphan cleanup with pulp smash. I can do that tho.

bmbouter:
> also i found this testcase that probably would be good to update https://github.com/pulp/pulpcore/blob/master/pulpcore/tests/functional/api/test_correlation_id.py#L31

I can fix this.

bmbouter:
Actually I can't fix this until the other PR merges because the new endpoint isn't available in this PR. I can fix it in the other one though. I want to merge this second so I'm going to just wait to merge for now.

PR #1442, opened 2021-06-28
Make new tasking system cd into working dir

bmbouter:
Without this, the new working dir is created, but it isn't changed into
so plugin code writes temp files into the wrong area.

[noissue]

### DIFF
--- a/pulpcore/tasking/pulpcore_worker.py
+++ b/pulpcore/tasking/pulpcore_worker.py
@@ -142,31 +142,32 @@ class NewPulpWorker:
         self.cursor.execute("LISTEN pulp_worker_cancel")
         task.worker = self.worker
         task.save(update_fields=["worker"])
-        task_process = Process(target=_perform_task, args=(task.pk,))
-        task_process.start()
-        while True:
-            r, w, x = select.select(
-                [connection.connection, task_process.sentinel], [], [], self.heartbeat_period
-            )
-            self.beat()
-            if r:
-                connection.connection.poll()
-                if any(
-                    (
-                        item.channel == "pulp_worker_cancel" and item.payload == str(task.pk)
-                        for item in connection.connection.notifies
-                    )
-                ):
-                    connection.connection.notifies.clear()
-                    _logger.info(f"Received signal to cancel current task {task.pk}.")
-                    task_process.terminate()
-                    break
-                if not task_process.is_alive():
-                    break
-        task_process.join()
-        if task.reserved_resources_record:
-            self.notify_workers()
-        self.cursor.execute("UNLISTEN pulp_worker_cancel")
+        with TemporaryDirectory(dir=".") as task_working_dir_rel_path:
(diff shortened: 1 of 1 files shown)


pulpcore/tasking/pulpcore_worker.py
```diff
@@ -206,7 +206,8 @@ def _perform_task(task_pk):
         func = getattr(module, function_name)
         args = json.loads(task.args) or ()
         kwargs = json.loads(task.kwargs) or {}
-        with TemporaryDirectory(dir="."):
+        with TemporaryDirectory(dir=".") as task_working_dir_rel_path:
+            os.chdir(task_working_dir_rel_path)
```
mdellweg, line 210: Doesn't this prevent deletion of the Directory later?
I started looking into giving the tempdir as a parameter to the process creation.

mdellweg: It does not seem possible. but we could still create the tempdir in `supervise_task` and add the dirname as an argument to `_perform_task`. First thing the new process does is cd into it.

mdellweg (approved):
I tried to outline a solution, where the supervising process cleans up the temp dir.
But i am not going to block this one.

pulpcore/tasking/pulpcore_worker.py
gerrod3, line 210: Is the worker it's own process each time a task is run? Would we need to cd back to the original directory so that the next task the worker does is not starting from some non-existent directory?

mdellweg: Yes, each task gets its own process. (To avoid memleaks for example.)
My fear is that deleting a dir is impossible as long as a process is using it. Also I'd prefer to have the cleanup in the outer process, because we may kill this one without warning.

dralley (approved):

bmbouter:
I tested cancellation also. If I make a task run for a long time and then cancel it I see that it cancels in the task record, the subprocess is killed, and there is no working directory leftover!

bmbouter:
@mdellweg I'm merging this so we can get things ready for 3.14. If there is any way this can be improved please let me know or open a PR. Either way is fine with me. Thanks everyone!

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -142,31 +142,32 @@ def supervise_task(self, task):
...
+                        connection.connection.notifies.clear()
+                        _logger.info(f"Received signal to cancel current task {task.pk}.")
+                        task_process.terminate()
+                        break
+                    if not task_process.is_alive():
+                        break
+            task_process.join()
+            if task.reserved_resources_record:
+                self.notify_workers()
+            self.cursor.execute("UNLISTEN pulp_worker_cancel")
```
mdellweg, line 170: We could unindent these 3 lines after `task_process.join()`, but it does not make a whole lot of difference.

PR #1445, opened 2021-06-28
Add environment variables for preliminary Django 3.0+ compat

dralley:
closes: #8975
https://pulp.plan.io/issues/8975


pulpcore/content/__init__.py
```diff
@@ -11,6 +11,10 @@
 import django
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "pulpcore.app.settings")
+# this is terrible, but we do try to manually ensure async safety, and it's the best we
+# can do until django supports async ORM natively.
+os.environ.setdefault("DJANGO_ALLOW_ASYNC_UNSAFE", "true")
```
bmbouter, line 16:
```suggestion
# Until Django supports async ORM natively this is the best we can do given these parts of Pulp run in coroutines.
# We try to ensure it is safe by never passing ORM data between co-routines
os.environ.setdefault("DJANGO_ALLOW_ASYNC_UNSAFE", "true")
```

bmbouter: Personally I don't think using the only option we have is "terrible"

dralley: On one hand yes, on the other hand, they explicitly warn against enabling it in production because of the dragons it could unleash, so...

bmbouter: Is it really terrible though? Django can't know if this is safe or not.

bmbouter: Maybe we could use a statement that doesn't include a subjective judgement at all?

bmbouter (approved):
Thanks @dralley

dkliban (approved):

PR #1446, opened 2021-06-29
Add authentication to content app

gerrod3:
fixes: #8951
https://pulp.plan.io/issues/8951

### DIFF
--- a/pulpcore/content/authentication.py
+++ b/pulpcore/content/authentication.py
@@ -1,0 +1,57 @@
+import asyncio
+import gettext
+import logging
+
+from aiohttp.web import middleware
+from django.http.request import HttpRequest
+from rest_framework.views import APIView
+from rest_framework.exceptions import APIException
+
+log = logging.getLogger(__name__)
+loop = asyncio.get_event_loop()
+_ = gettext.gettext
+
+
+@middleware
+async def authenticate(request, handler):
+    """Authenticates the request to the content app using the DRF authentication classes"""
+    django_request = convert_request(request)
+    fake_view = APIView()
+
+    def _authenticate_blocking():
+        drf_request = fake_view.initialize_request(django_request)
+        try:
+            fake_view.perform_authentication(drf_request)
+        except APIException as e:
+            log.warning(_('"{} {}" "{}": {}').format(request.method, request.path, request.host, e))
+
+        return drf_request
+
+    auth_request = await loop.run_in_executor(None, _authenticate_blocking)
+    request["user"] = auth_request.user
+    request["auth"] = auth_request.auth
+    request["drf_request"] = auth_request
+
+    return await handler(request)
+
+
+def convert_request(request):
+    """
+    Converts an aiohttp Request to a Django HttpRequest
+
+    This does not convert the async body to a sync body for POST requests
+    """
+    djr = HttpRequest()
+    djr.method = request.method
+    upper_keys = {k.upper() for k in request.headers.keys()}
+    # These two headers are specially set by Django without the HTTP prefix
+    h = {"CONTENT_LENGTH", "CONTENT_TYPE"}
+    djr.META = {f"HTTP_{k}" if k not in h else k: request.headers[k] for k in upper_keys}
+    djr.COOKIES = request.cookies
+    djr.path = request.path
+    djr.path_info = request.match_info.get("path", "")
+    djr.encoding = request.charset
+    djr.GET = request.query
+
+    djr._get_scheme = lambda: request.scheme
+    return djr
--- a/CHANGES/8951.feature
+++ b/CHANGES/8951.feature
@@ -1,0 +1,1 @@
+Added authentication to the content app.
--- a/pulpcore/content/__init__.py
+++ b/pulpcore/content/__init__.py
@@ -23,11 +23,12 @@ from pulpcore.app.apps import pulp_plugin_configs  # noqa: E402: module level no
 from pulpcore.app.models import ContentAppStatus  # noqa: E402: module level not at top of file
 
 from .handler import Handler, loop  # noqa: E402: module level not at top of file
+from .authentication import authenticate  # noqa: E402: module level not at top of file
 
 
 log = logging.getLogger(__name__)
 
-app = web.Application()
+app = web.Application(middlewares=[authenticate])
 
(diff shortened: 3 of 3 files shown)


pulpcore/content/authentication.py
```diff
@@ -0,0 +1,56 @@
...
+
+log = logging.getLogger(__name__)
+loop = asyncio.get_event_loop()
+
+
+@middleware
+async def authenticate(request, handler):
+    """Authenticates the request to the content app using the DRF authentication classes"""
+    django_request = convert_request(request)
+    fake_view = APIView()
```
mdellweg, line 17: seen that before XD

pulpcore/content/authentication.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def _authenticate_blocking():
+        drf_request = fake_view.initialize_request(django_request)
+        try:
+            fake_view.perform_authentication(drf_request)
+        except APIException as e:
+            # maybe change to User-Agent?
+            log.warning(f'"{request.method} {request.path}" "{request.host}": {e}')
```
mdellweg, line 25: I think we should use gettext here.

pulpcore/content/authentication.py
```diff
@@ -0,0 +1,56 @@
...
+
+        return drf_request
+
+    auth_request = await loop.run_in_executor(None, _authenticate_blocking)
+    request["user"] = auth_request.user
+    request["auth"] = auth_request.auth
+    request["drf_request"] = auth_request
```
mdellweg, line 32: Is this needed? Or are you anticipating it's use somehow?

mdellweg (commented):
One nitpic about i18n.
And for my understanding: Whatever authentication methods Django is configured to will work here? Will it propagate the headers from DRF's AuthenticationFailed Exception to the response?

pulpcore/content/authentication.py
gerrod3: This is used by the RBAC content guard.

gerrod3:
> And for my understanding: Whatever authentication methods Django is configured to will work here? Will it propagate the headers from DRF's AuthenticationFailed Exception to the response?

In theory yes, any authentication methods Django is configured with should work here. As for the second part, no. If the authentication fails then an error will be raised and I catch that error and only log its occurrence. With this current implementation if no authentication is provided with the request then no error is raised, but either way DRF defaults the auth user to AnonymousUser when there is no authentication or authentication fails.

mdellweg:
> > And for my understanding: Whatever authentication methods Django is configured to will work here? Will it propagate the headers from DRF's AuthenticationFailed Exception to the response?
> 
> In theory yes, any authentication methods Django is configured with should work here. As for the second part, no. If the authentication fails then an error will be raised and I catch that error and only log its occurrence. With this current implementation if no authentication is provided with the request then no error is raised, but either way DRF defaults the auth user to AnonymousUser when there is no authentication or authentication fails.

I am thinking especially about the "Www-Authenticate" header, that may allow some challenge-response authentication, or a redirect to a token server. Maybe we don't need it today...

https://www.django-rest-framework.org/api-guide/authentication/#unauthorized-and-forbidden-responses

dkliban (approved):

pulpcore/content/authentication.py
```diff
@@ -0,0 +1,57 @@
...
+    """Authenticates the request to the content app using the DRF authentication classes"""
+    django_request = convert_request(request)
+    fake_view = APIView()
+
+    def _authenticate_blocking():
+        drf_request = fake_view.initialize_request(django_request)
+        try:
+            fake_view.perform_authentication(drf_request)
+        except APIException as e:
+            log.warning(_('"{} {}" "{}": {}').format(request.method, request.path, request.host, e))
```
mdellweg, line 26: There isn't much left to translate... :facepalm:

pulpcore/content/authentication.py
```diff
@@ -0,0 +1,57 @@
...
+async def authenticate(request, handler):
+    """Authenticates the request to the content app using the DRF authentication classes"""
+    django_request = convert_request(request)
+    fake_view = APIView()
+
+    def _authenticate_blocking():
+        drf_request = fake_view.initialize_request(django_request)
+        try:
+            fake_view.perform_authentication(drf_request)
+        except APIException as e:
```
mdellweg, line 25: What happens, if i provide a user with a wrong password here?
I think, it will raise a `AuthenticationFailed` exception, and we should return 401 or 403.
Also on any exception, will the `user` and `auth` fields be usable in line 31f?

mdellweg (commented):
I think, we need to think about a failed authorization once more.

pulpcore/content/authentication.py
gerrod3: Uhhh, not sure what was there to translate in the first place. The message is of the form `"GET /pulp/content/foo/simple/" "127.0.0.1": Invalid username/password`. The error message is the only thing that could be translated. What is the correct way to make sure the message is translate?

pulpcore/content/authentication.py
gerrod3: 1. It'll raise an `AuthenticationFailed` exception with the message `Invalid username/password`, the expected error code is 401.
2. Yes, `user` will be set to an instance of `AnonymousUser` and `auth` will be set to `None`.

Maybe I should add a setting for strict authentication fails. Brian wanted the auth to still work even if no auth credentials are supplied so that the content app would continue to work like it currently does. If auth fails and you try to get content from a RBAC guard you will get a 403.

pulpcore/content/authentication.py
mdellweg: I was referring to the fact, that `'"{} {}" "{}": {}'` will look rather dull in a translation app. But it is probably the right thing to do anyway.

pulpcore/content/authentication.py
mdellweg: To continue without credentials, is absolutely ok. What I am uneasy about is continuing with invalid credentials.
But i'm not going to block it here.

I thought drf will handle the no-creds-provided case without an exception. That is how i read: https://www.django-rest-framework.org/api-guide/authentication/#how-authentication-is-determined

mdellweg (approved):

PR #1447, opened 2021-06-29
Shields coroutine that is saving Artifact and Content

dkliban:
After the client receives the last chunk of the response, it closes the connection
with the content app. This causes all scheduled coroutines to be cancelled. If the
the thread that is performing the Artifact/Content save has not finished running,
the Content is never saved.

This patch ensures that the Artifact and the Content are fully saved to the
database after being streamed to the user.

fixes: #8980
https://pulp.plan.io/issues/8980

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -784,7 +784,7 @@ class Handler:
             def save_artifact_blocking():
                 self._save_artifact(download_result, remote_artifact)
 
-            await loop.run_in_executor(None, save_artifact_blocking)
+            await asyncio.shield(loop.run_in_executor(None, save_artifact_blocking))
         await response.write_eof()
 
         if response.status == 404:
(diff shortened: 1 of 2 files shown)


bmbouter:
I'd like us to explore a fix more than this workaround.

pulpcore/content/handler.py
```diff
@@ -780,11 +780,8 @@ async def finalize():
         download_result = await downloader.run()
 
         if remote.policy != Remote.STREAMED:
+            self._save_artifact(download_result, remote_artifact)
```
dralley, line 783: Leave a TODO comment about the blocking call so that we don't forget about it

bmbouter (approved):
Thank you so much for re-evaluating this fix and producing this!

dralley (approved):

PR #1448, opened 2021-06-29
Improve the help text for the remote timeout options

quba42:
[noissue]

Adds an additional hint on how to look for the default values and by
extension what one might plausibly change the value to.

This is kind of implied already by the `(q.v.)` abbreviation, but I think there might be the odd user who is helped by it, and I don't think slightly longer help strings in the API docs ever really hurt. Any thoughts?

If there is agreement that this is an improvement I can create an issue to go with it, if desired.


dralley (approved):

bmbouter (approved):
Thanks!

PR #1452, opened 2021-07-01
Reclaim disk space story.

ipanova:
closes #8459

### DIFF
--- a/CHANGES/8459.feature
+++ b/CHANGES/8459.feature
@@ -1,0 +1,1 @@
+Added feature to reclaim disk space for a list of repositories.
--- a/docs/workflows/reclaim-disk-space.rst
+++ b/docs/workflows/reclaim-disk-space.rst
@@ -1,0 +1,78 @@
+.. _reclaim-disk-space:
+
+Reclaim disk space for a list of repositories
+---------------------------------------------
+
+.. note::
+
+    Pulp's reclaim disk space feature is in tech-preview and may change in backwards incompatible
+    ways in future releases.
+
+
+Pulp provides the ability to reclaim disk space for:
+
+  * content that one no longer needs to serve but would like to keep in the repo for historical
+    reasons.
+  * repos that were synced with the on_demand download policy and one would want to clear out
+    downloaded files for those repos. Usually, a repository that was synced with the on_demand
+    download policy will store artifacts locally after they have been requested by the client, but
+    there really isn't a way to have pulp delete the locally stored files and free disk space if
+    these packages are unlikely to be used again.
+
+To start a reclaim task send a POST request to `/pulp/api/v3/repositories/reclaim_space/`.
+
+.. code-block::
+
+    $ RECLAIM_TASK=$(http POST :24817/pulp/api/v3/repositories/reclaim_space/ repo_hrefs:=[\"/pulp/api/v3/repositories/rpm/rpm/b3a6674d-181c-4e72-9412-7cbc747480ad/\"] | qq -r '.task')
+    $ http --body :24817$RECLAIM_TASK
+
+    {
+        "child_tasks": [],
+        "created_resources": [],
+        "error": null,
+        "finished_at": "2021-07-16T15:36:20.650573Z",
+        "logging_cid": "50d1721d205c40f69defb773e32a98ff",
+        "name": "pulpcore.app.tasks.reclaim_space.reclaim_space",
+        "parent_task": null,
+        "progress_reports": [
+           {
+                "code": "reclaim-space.artifact",
+                "done": 35,
+                "message": "Reclaim disk space",
+                "state": "completed",
+                "suffix": null,
+                "total": 35
+            }
+        ],
+        "pulp_created": "2021-07-16T15:36:20.306845Z",
+        "pulp_href": "/pulp/api/v3/tasks/20ee50bd-9392-4ebf-8f1d-d2f15474ebd6/",
+        "reserved_resources_record": [
+            "/pulp/api/v3/repositories/rpm/rpm/b3a6674d-181c-4e72-9412-7cbc747480ad/"
+        ],
+        "started_at": "2021-07-16T15:36:20.370245Z",
+        "state": "completed",
+        "task_group": null,
+        "worker": "/pulp/api/v3/workers/ccc132c4-0445-4f55-b370-32c3662dce3c/"
+     }
+
+As a result of this request, disk space will be freed-up for artifacts that are exclusive to the
+list of provided repos. The content of the repository versions will not change and no repository
+versions will be created or removed.
+
+An optional ``repo_versions_keeplist`` parameter can be specified, that will contain list of repo
+version hrefs which will be excluded from the artifact removal.
+
+By default, the task will not remove artifacts from content that was uploaded directly into Pulp but
+only content that was synced from a remote source.
+
+.. note::
+
+   The task will clean up artifacts regardless of the download policy. The content app will be able
+   to stream artifact if it is locally available, otherwise it will attempt to redownload it from
+   the known upstream urls. In case upstream stopped serving the corresponding file, Pulp won't be
+   able to download and serve it.
+
+.. warning::
+
+    For uploaded content that, as a rule, does not have a remote source url, a ``force`` flag must
+    be specified. This, however, will lead to a non-consumable content afterwards. 
--- a/pulpcore/app/serializers/reclaim.py
+++ b/pulpcore/app/serializers/reclaim.py
@@ -1,0 +1,48 @@
+from gettext import gettext as _
+
+from rest_framework import fields, serializers
+
+from pulpcore.app.models import Repository
+
+from pulpcore.app.serializers import RepositoryVersionRelatedField, ValidateFieldsMixin
+
+
+class ReclaimSpaceSerializer(serializers.Serializer, ValidateFieldsMixin):
+    """
+    Serializer for reclaim disk space operation.
+    """
+
+    repo_hrefs = fields.ListField(
+        required=True,
+        help_text=_("Will reclaim space for the specified list of repos."),
+    )
+    repo_versions_keeplist = RepositoryVersionRelatedField(
+        help_text=_("Will exclude repo versions from space reclaim."),
+        many=True,
+        required=False,
+    )
+    force = fields.BooleanField(
+        required=False,
+        default=False,
+        help_text=_("Will remove Artifacts for uploaded content. Disabled by default"),
+    )
+
+    def validate_repo_hrefs(self, value):
+        """
+        Check that the repo_hrefs is not an empty list and contains all valid hrefs.
+        Args:
+            value (list): The list supplied by the user
+        Returns:
+            The list of pks (not hrefs) after validation
+        Raises:
+            ValidationError: If the list is empty or contains invalid hrefs.
+        """
+        if len(value) == 0:
+            raise serializers.ValidationError("Must not be [].")
+        from pulpcore.app.viewsets import NamedModelViewSet
+
+        hrefs_to_return = []
+        for href in value:
+            hrefs_to_return.append(NamedModelViewSet.get_resource(href, Repository))
+
+        return hrefs_to_return
--- a/pulpcore/app/tasks/reclaim_space.py
+++ b/pulpcore/app/tasks/reclaim_space.py
@@ -1,0 +1,83 @@
+from pulpcore.app.models import (
+    Artifact,
+    Content,
+    ContentArtifact,
+    ProgressReport,
+    PublishedMetadata,
+    Repository,
+    RepositoryVersion,
+)
+
+
+def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
+    """
+    This task frees-up disk space by removing Artifact files from the filesystem for Content
+    exclusive to the list of provided repos.
+
+    Note: content marked as `proctected` will be excluded from the reclaim disk space.
+
+    Kwargs:
+        repo_pks (list): A list of repo pks the disk reclaim space is performed on.
+        keeplist_rv_pks (list): A list of repo version pks that will be excluded from the reclaim
+        disk space.
+        force (bool): If True, uploaded content will be taken into account.
+
+    """
+    rest_of_repos = Repository.objects.exclude(pk__in=repo_pks)
+    c_keep_qs = Content.objects.filter(repositories__in=rest_of_repos)
+    c_reclaim_qs = Content.objects.filter(repositories__in=repo_pks)
+    c_reclaim_qs = c_reclaim_qs.exclude(
+        pk__in=c_keep_qs, pulp_type=PublishedMetadata.get_pulp_type()
+    )
+
+    if keeplist_rv_pks:
+        rv_qs = RepositoryVersion.objects.filter(pk__in=keeplist_rv_pks)
+        rv_content = Content.objects.none()
+        for rv in rv_qs.iterator():
+            rv_content |= rv.content
+        c_reclaim_qs = c_reclaim_qs.exclude(pk__in=rv_content)
+
+    content_distinct = c_reclaim_qs.distinct("pulp_type")
+    unprotected = []
+    for content in content_distinct:
+        if not content.cast().PROTECTED_FROM_RECLAIM:
+            unprotected.append(content.pulp_type)
+
+    ca_qs = ContentArtifact.objects.filter(
+        content__in=c_reclaim_qs.values("pk"), artifact__isnull=False
+    )
+    if not force:
+        ca_qs = ca_qs.filter(remoteartifact__isnull=False)
+    artifact_pks = set()
+    ca_to_update = []
+    for ca in ca_qs.iterator():
+        if ca.content.pulp_type in unprotected:
+            artifact_pks.add(ca.artifact.pk)
+            ca.artifact = None
+            ca_to_update.append(ca)
+
+    ContentArtifact.objects.bulk_update(objs=ca_to_update, fields=["artifact"], batch_size=1000)
+    artifacts_to_delete = Artifact.objects.filter(pk__in=artifact_pks)
+    progress_bar = ProgressReport(
+        message="Reclaim disk space",
+        total=artifacts_to_delete.count(),
+        code="reclaim-space.artifact",
+        done=0,
+        state="running",
+    )
+    progress_bar.save()
+
+    counter = 0
+    interval = 100
+    for artifact in artifacts_to_delete.iterator():
+        # we need to manually call delete() because it cleans up the file on the filesystem
+        artifact.delete()
+        progress_bar.done += 1
+        counter += 1
+
+        if counter >= interval:
+            progress_bar.save()
+            counter = 0
+
+    progress_bar.state = "completed"
+    progress_bar.save()
--- a/pulpcore/app/viewsets/reclaim.py
+++ b/pulpcore/app/viewsets/reclaim.py
@@ -1,0 +1,49 @@
+from drf_spectacular.utils import extend_schema
+from rest_framework.viewsets import ViewSet
+
+from pulpcore.app.response import OperationPostponedResponse
+from pulpcore.app.serializers import AsyncOperationResponseSerializer, ReclaimSpaceSerializer
+from pulpcore.app.tasks import reclaim_space
+from pulpcore.tasking.tasks import dispatch
+
+
+class ReclaimSpaceViewSet(ViewSet):
+    """
+    Viewset for reclaim disk space endpoint.
+    """
+
+    serializer_class = ReclaimSpaceSerializer
+
+    @extend_schema(
+        description="Trigger an asynchronous space reclaim operation.",
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    def reclaim(self, request):
+        """
+        Triggers an asynchronous space reclaim operation.
+        """
+        serializer = ReclaimSpaceSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        repos = serializer.validated_data.get("repo_hrefs", [])
+        keeplist = serializer.validated_data.get("repo_versions_keeplist", [])
+        reclaim_repo_pks = []
+        keeplist_rv_pks = []
+        for repo in repos:
+            reclaim_repo_pks.append(repo.pk)
+        for rv in keeplist:
+            repos.append(rv.repository)
+            keeplist_rv_pks.append(rv.pk)
+        force = serializer.validated_data.get("force", False)
+
+        task = dispatch(
+            reclaim_space,
+            repos,
+            kwargs={
+                "repo_pks": reclaim_repo_pks,
+                "keeplist_rv_pks": keeplist_rv_pks,
+                "force": force,
+            },
+        )
+
+        return OperationPostponedResponse(task, request)
--- a/CHANGES/plugin_api/8459.feature
+++ b/CHANGES/plugin_api/8459.feature
@@ -1,0 +1,2 @@
+Content model has a new boolean class constant ``PROTECTED_FROM_RECLAIM`` for plugins to enable the
+reclaim disk space feature provided by core.
--- a/docs/workflows/index.rst
+++ b/docs/workflows/index.rst
@@ -18,6 +18,7 @@ assumes that the reader is familiar with the fundamentals discussed in the :doc:
    exposing-content
    signed-metadata
    repairing-pulp
+   reclaim-disk-space
    import-export
    labels
    troubleshooting
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -439,6 +439,8 @@ class Content(MasterModel, QueryMixin):
         _artifacts (models.ManyToManyField): Artifacts related to Content through ContentArtifact
     """
 
+    PROTECTED_FROM_RECLAIM = True
+
     TYPE = "content"
     repo_key_fields = ()  # Used by pulpcore.plugin.repo_version_utils.remove_duplicates
     upstream_id = models.UUIDField(null=True)  # Used by PulpImport/Export processing
--- a/pulpcore/app/serializers/__init__.py
+++ b/pulpcore/app/serializers/__init__.py
@@ -73,6 +73,7 @@ from .repository import (  # noqa
     RepositoryVersionSerializer,
 )
 from .repair import RepairSerializer  # noqa
+from .reclaim import ReclaimSpaceSerializer  # noqa
 from .task import (  # noqa
     MinimalTaskSerializer,
     TaskCancelSerializer,
--- a/pulpcore/app/tasks/__init__.py
+++ b/pulpcore/app/tasks/__init__.py
@@ -6,4 +6,6 @@ from .importer import pulp_import  # noqa
 
 from .orphan import orphan_cleanup  # noqa
 
+from .reclaim_space import reclaim_space  # noqa
+
 from .repository import repair_all_artifacts  # noqa
--- a/pulpcore/app/urls.py
+++ b/pulpcore/app/urls.py
@@ -16,7 +16,11 @@ from rest_framework_nested import routers
 
 from pulpcore.app.apps import pulp_plugin_configs
 from pulpcore.app.views import OrphansView, PulpImporterImportCheckView, RepairView, StatusView
-from pulpcore.app.viewsets import ListRepositoryVersionViewSet, OrphansCleanupViewset
+from pulpcore.app.viewsets import (
+    ListRepositoryVersionViewSet,
+    OrphansCleanupViewset,
+    ReclaimSpaceViewSet,
+)
 from pulpcore.constants import API_ROOT
 from pulpcore.openapi import PulpSchemaGenerator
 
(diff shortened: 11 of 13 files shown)


pulpcore/app/viewsets/reclaim.py
```diff
@@ -0,0 +1,45 @@
...
+        for repo in repos:
+            reclaim_repo_pks.append(repo.pk)
+        for rv in keeplist:
+            repos.append(rv.repository)
+            keeplist_rv_pks.append(rv.pk)
+        force = serializer.validated_data.get("force", False)
+
+        task = dispatch(
+            reclaim_space,
+            repos,
```
ipanova, line 37: initial story was about reclaiming disk space only on for a set of repos. Is it worth adding reclaim disk space for *all* the repos? In that case this task should probably be a blocking one.

mdellweg: I'd say, we don't want any globally locking tasks.

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -0,0 +1,59 @@
...
+
+    ca_qs = ContentArtifact.objects.filter(content__in=c_reclaim_qs.values("pk"))
+    if not force:
+        ca_qs = ca_qs.filter(remoteartifact__isnull=False)
+    for ca in ca_qs:
+        ca.artifact = None
+    ContentArtifact.objects.bulk_update(objs=ca_qs, fields=["artifact"], batch_size=1000)
+    # respect timestamp when it will be added
+    artifacts_to_delete = Artifact.objects.filter(content_memberships__isnull=True).exclude(
+        pk__in=artifacts_before
```
ipanova, line 35: this was an attempt for optimization in case there would be tons of orphaned artifacts not really related to this operation they would be removed together with this call and it would take more time

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,67 @@
...
+Reclaim disk space for a list of repositories
+---------------------------------------------
+
+.. note::
+
+    Pulp reclaim disk space feature are in tech-preview and may change in backwards incompatible
+    ways in future releases.
+
+
+Pulp provides some feature for reclaiming disk space for:
```
dralley, line 12: "some" feels like a typo, maybe rwrite as "Pulp provides the ability to reclaim disk space for"?

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,67 @@
...
+
+  * content that one no longer needs to serve but wants to keep it in the repo for historical
```
dralley, line 14: Same with "wants"

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,67 @@
...
+    purpose.
+  * repos that have on_demand download policy and one would want to clear out downloaded files for
+    those repos. Usually, a repository with on_demand download policy will store artifacts locally
+    after they have been requested, but there really isn't a way to have pulp delete the locally
+    stored files and free disk space if these packages are unlikely to be used again.
```
dralley, line 19: It would be a neat idea (for the future) to let a user configure this to be done automatically for old repository versions after a new one is created.

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,67 @@
...
+        "started_at": "2021-07-16T15:36:20.370245Z",
+        "state": "completed",
+        "task_group": null,
+        "worker": "/pulp/api/v3/workers/ccc132c4-0445-4f55-b370-32c3662dce3c/"
+     }
+
+As a result of this request disk space will be freed-up for:
+
+  * artifacts only exclusive to that list of provided repos. No new versions will be created or
+    removed because content set will not change.
```
dralley, line 60: Rewriting this slightly:

> Artifacts that are exclusive to the list of provided repos. The content of the repository versions will not change and no repository versions will be created or removed.

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,67 @@
...
+  * willingly, a ``epo_versions_keeplist`` option can be specified that will contain list of
```
dralley, line 61: Typo, "epo_versions_keeplist"

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,67 @@
...
+    ``repo_version_hrefs`` which will be excluded from Artifact removal.
```
dralley, line 62: "repo_version_hrefs" doesn't look like it's a parameter and it's not a resource type, maybe it should be "repo version ``href``s?

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,67 @@
...
+  * by default, the task will remove only those Artifacts that have a corresponding RemoteArtifact.
+  * for uploaded content that does not have RemoteArtifacts, a force flag should be specified.
+    This, however, will lead to a non-cosumable content afterwards.
```
dralley, line 65: Typo "cosumable"

I'm conflicted about this because `RemoteArtifact` is a very low-level detail and not something that we should expect users to know or care about.  But the risk of corruption makes it important, which feels like a major red flag.

Is a `force` flag a hard requirement?  Do we expect users to have a critical need to be able to delete content for which there is no longer any remote?

If it's not, I would be more comfortable leaving it out.  Or considering implementing something like [method 1d](https://hackmd.io/_3gsUVdyQwy50Nc5pMXN-g) to prevent the entire issue.

pulpcore/app/serializers/reclaim.py
```diff
@@ -0,0 +1,45 @@
...
+
+    repo_hrefs = fields.ListField(
+        required=True,
+        help_text=_("Will reclaim space for the specified list of repos."),
+    )
+    repo_versions_keeplist = RepositoryVersionRelatedField(
+        help_text=_("Will exclude repo versions from space reclaim."),
+        many=True,
+        required=False,
+    )
```
dralley, line 20: A neat idea (for the future - not now) would be to let a user configure this to be done automatically for old repository versions after a new one is created.

pulpcore/app/serializers/reclaim.py
```diff
@@ -0,0 +1,45 @@
...
+        Raises:
+            ValidationError: If the list is empty or contains invalid hrefs.
+        """
+        if len(value) == 0:
+            raise serializers.ValidationError("Must not be [].")
+        from pulpcore.app.viewsets import NamedModelViewSet
+
+        hrefs_to_return = []
+        for href in value:
+            hrefs_to_return.append(NamedModelViewSet.get_resource(href, Repository))
```
dralley, line 43: If this is called with hundreds or thousands of repo hrefs it might be a problem https://pulp.plan.io/issues/7483

But it's less likely to be a big issue then it was with content, where often tens of thousands were being passed at a time.

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -0,0 +1,59 @@
...
+    ContentArtifact,
+    ProgressReport,
+    Repository,
+    RepositoryVersion,
+)
+
+
+def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
+    artifacts_before = list(
+        Artifact.objects.filter(content_memberships__isnull=True).values_list("pk", flat=True)
```
dralley, line 13: If this is only being used once (on line 35), I would recommend not casting it to a list, and not using `values_list`.  Django can just fold it into the query on line 35 and avoid two separate queries where it has to move the results over to Python and then back.

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -0,0 +1,59 @@
...
+        rv_qs = RepositoryVersion.objects.filter(pk__in=keeplist_rv_pks)
+        rv_content = set()
+        for rv in rv_qs.iterator():
+            rv_content.update(rv.content)
+        c_reclaim_qs = set(c_reclaim_qs).difference(rv_content)
+
+    ca_qs = ContentArtifact.objects.filter(content__in=c_reclaim_qs.values("pk"))
+    if not force:
+        ca_qs = ca_qs.filter(remoteartifact__isnull=False)
+    for ca in ca_qs:
```
dralley, line 30: `.iterator()`

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -0,0 +1,59 @@
...
+    c_keep_qs = Content.objects.filter(repositories__in=rest_of_repos)
+    c_reclaim_qs = Content.objects.filter(repositories__in=repo_pks)
+    c_reclaim_qs = c_reclaim_qs.difference(c_keep_qs)
+
+    if keeplist_rv_pks:
+        rv_qs = RepositoryVersion.objects.filter(pk__in=keeplist_rv_pks)
+        rv_content = set()
+        for rv in rv_qs.iterator():
+            rv_content.update(rv.content)
+        c_reclaim_qs = set(c_reclaim_qs).difference(rv_content)
```
dralley, line 25: Can this be done without casting to a set?

pulpcore/app/viewsets/reclaim.py
```diff
@@ -0,0 +1,45 @@
...
+        repos = serializer.validated_data.get("repo_hrefs", [])
+        keeplist = serializer.validated_data.get("repo_versions_keeplist", [])
+        reclaim_repo_pks = []
+        keeplist_rv_pks = []
+        for repo in repos:
+            reclaim_repo_pks.append(repo.pk)
+        for rv in keeplist:
+            repos.append(rv.repository)
+            keeplist_rv_pks.append(rv.pk)
+        force = serializer.validated_data.get("force", False)
```
dralley, line 33: I don't recall if there was an actual reason why we do this, but the RPM Copy Serializer does all of the business logic validation (checking that the hrefs are valid and exist) in the viewset (here) rather than the serializer.

This isn't a suggestion at all, I'm just pointing it out.

docs/workflows/reclaim-disk-space.rst
ipanova: automation hub needs a way to reclaim disk space for content that has been uploaded https://hackmd.io/e78_-T2rQRWvXaGjHT6Kyw#First-workflow-container-where-versions-are-kept-for-history

pulpcore/app/tasks/reclaim_space.py
ipanova: i actually need to evaluate the query here and register the results because if i don't do it then on line 35 when it will be evaluated it will have different results since between lines 12 and 35 i'm actually creating more orphan artifacts.

docs/workflows/reclaim-disk-space.rst
dralley: Then we really need to make sure that it fails as gracefully as possible if purged content is attempted to be downloaded.

pulpcore/app/tasks/reclaim_space.py
dralley: Could the order of lines 32 and 35 be swapped?

CHANGES/8459.feature
```diff
@@ -0,0 +1 @@
+As a user I want to reclaim disk space for a list of repositories.
```
daviddavis, line 1:
```suggestion
Added feature to reclaim disk space for a list of repositories.
```

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,67 @@
+.. _reclaim-disk-space:
+
+Reclaim disk space for a list of repositories
+---------------------------------------------
+
+.. note::
+
+    Pulp reclaim disk space feature are in tech-preview and may change in backwards incompatible
```
daviddavis, line 8:
```suggestion
    Pulp's reclaim disk space feature is in tech-preview and may change in backwards incompatible
```

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,67 @@
...
+    ways in future releases.
+
+
+Pulp provides the ability to reclaim disk space for:
+
+  * content that one no longer needs to serve but would like to keep it in the repo for historical
+    purpose.
```
daviddavis, line 15:
```suggestion
  * content that one no longer needs to serve but would like to keep in the repo for historical reasons
```

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,67 @@
...
+        "reserved_resources_record": [
+            "/pulp/api/v3/repositories/rpm/rpm/b3a6674d-181c-4e72-9412-7cbc747480ad/"
+        ],
+        "started_at": "2021-07-16T15:36:20.370245Z",
+        "state": "completed",
+        "task_group": null,
+        "worker": "/pulp/api/v3/workers/ccc132c4-0445-4f55-b370-32c3662dce3c/"
+     }
+
+As a result of this request disk space will be freed-up for:
```
daviddavis, line 57:
```suggestion
As a result of this request, disk space will be freed-up for:
```

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,67 @@
...
+
+  * artifacts that are exclusive to the list of provided repos. The content of the repository
+    versions will not change and no repository versions will be created or removed.
+  * willingly, a ``repo_versions_keeplist`` option can be specified that will contain list of repo
```
daviddavis, line 61:
```suggestion
  * an optional ``repo_versions_keeplist`` option can be specified that will contain list of repo
```

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,67 @@
...
+    version hrefs which will be excluded from Artifact removal.
+  * by default, the task will remove only those Artifacts that have a corresponding RemoteArtifact.
+  * for uploaded content that does not have RemoteArtifacts, a force flag should be specified.
+    This, however, will lead to a non-consumable content afterwards.
+    
+  The task will trim Artifacts regardless of the download policy. Content app would be able to stream artifact if it is locally available, otherwise with the help of RemoteArtifact (if available) it would download artifact from the remote source.
```
daviddavis, line 67: Does trim here mean delete?

```suggestion
The task will clean up Artifacts regardless of the download policy. The content app will be able to
stream artifact if it is locally available, otherwise using the RemoteArtifact (if available), it
will download artifact from the remote source.
```

pulpcore/app/viewsets/reclaim.py
daviddavis: Yea, I think probably the RPM code ought to be updated to perform its validation in a serializer.

pulpcore/app/tasks/reclaim_space.py
daviddavis: It seems to me that this could be problematic. Won't this clean up orphan artifacts that got created after `artifacts_before` got set?

ipanova: yes it will :/  is your suggestion to directly collect the artifact pks as a loop throug the ca_qs?

docs/workflows/reclaim-disk-space.rst
ipanova: that's a good point, i will need to look into this and confirm

docs/workflows/reclaim-disk-space.rst
ipanova: yes

docs/workflows/reclaim-disk-space.rst
dralley: https://pulp.plan.io/issues/8305 is probably a good place to write the results

pulpcore/app/tasks/reclaim_space.py
daviddavis: That's my thought. Any downsides to that?

ipanova: i guess none, just did not cross my mind earlier since this idea with optimization of orphaned artifacts just gave me tunnel vision :D

pulpcore/app/tasks/reclaim_space.py
ipanova: i have removed this completely and replaced with https://github.com/pulp/pulpcore/pull/1452#discussion_r673113048

docs/workflows/reclaim-disk-space.rst
ipanova: I have not manually tested rpm plugin yet,  but skimmed through pulpcore code - plugins that use directly pulp's content app handler, should be able to gracefully handle this. Uploaded content will have artifact set to none as well it won't have any remoteartifacts so a 404 will just be raised https://github.com/pulp/pulpcore/blob/master/pulpcore/content/handler.py#L681.
Since pulp-container has a subclassed version of handler, i needed to modify couple of lines which led to this output:
```
$ podman pull localhost:24817/test/repo --tls-verify=false
Trying to pull localhost:24817/test/repo:latest...
Error: Error initializing source docker://localhost:24817/test/repo:latest: Error reading manifest latest in localhost:24817/test/repo: StatusCode: 404, 404: Not Found
(pulp) [vagrant@pulp3-source-fedora34 ~]$ 
```

tldr, i think we're fine just need to audit plugins that subclass the Handler.

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,69 @@
...
+
+As a result of this request, disk space will be freed-up for:
+
+  * artifacts that are exclusive to the list of provided repos. The content of the repository
+    versions will not change and no repository versions will be created or removed.
+  * an optional ``repo_versions_keeplist`` parameter can be specified that will contain list of
+    repo version hrefs which will be excluded from Artifact removal.
+  * by default, the task will remove only those Artifacts that have a corresponding RemoteArtifact.
+  * for uploaded content that does not have RemoteArtifacts, a ``force`` flag should be specified.
+    This, however, will lead to a non-consumable content afterwards.
```
dralley, line 65: I still feel like we shouldn't need to explain `RemoteArtifacts` to the user -- and probably the `--force` option should have some kind of high-visibility note or warning box to call out the potential for things to go wrong.  We can make it available but the user should understand that it's something they really really shouldn't do unless they know what they are doing.

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,69 @@
...
+        "worker": "/pulp/api/v3/workers/ccc132c4-0445-4f55-b370-32c3662dce3c/"
+     }
+
+As a result of this request, disk space will be freed-up for:
+
+  * artifacts that are exclusive to the list of provided repos. The content of the repository
+    versions will not change and no repository versions will be created or removed.
+  * an optional ``repo_versions_keeplist`` parameter can be specified that will contain list of
+    repo version hrefs which will be excluded from Artifact removal.
+  * by default, the task will remove only those Artifacts that have a corresponding RemoteArtifact.
```
dralley, line 63: I think the two bullet points about the `--force` option should be in a separate paragraph from the rest (or maybe in a "note" box) for visibility.  Alternatively we should use a big "warning" box such as the one on this page: https://docs.pulpproject.org/pulpcore/workflows/repairing-pulp.html

We can make it available but it should be clear that the user should never use this option without understanding the full implications.  I still think that needing to explain low-level details such as `RemoteArtifact` is a red flag and a usability problem.

We should also call out that if the content cannot be guaranteed to be / stay available at the remote source, either due to changes at the remote source or changes to the remote configuration.

dralley (changes requested):

docs/workflows/reclaim-disk-space.rst
ipanova: @dralley i tried to re-word couple of things is this is more user friendly and we don't talk about low-level stuff such ass 'remote artifact'. Let me know if that looks better to you and if i need to make any further adjustments.

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,74 @@
...
+    into Pulp.
+
+.. warning::
+
+    For uploaded content that, as a rule, does not have a remote source url, a ``force`` flag must
+    be specified. This, however, will lead to a non-consumable content afterwards.
+    
+The task will clean up artifacts regardless of the download policy. The content app will be able
+to stream artifact if it is locally available, otherwise it will attempt to redownload it from the
+known upstream urls. In case upstream stopped serving the corresponding file , Pulp won't be able to
```
dralley, line 73: Extra space before the comma

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,76 @@
...
+  * an optional ``repo_versions_keeplist`` parameter can be specified that will contain list of
+    repo version hrefs which will be excluded from the artifact removal.
+  * by default, the task will remove only those artifacts from the content that were synchronized
+    into Pulp.
+
+.. note::
+
+   The task will clean up artifacts regardless of the download policy. The content app will be able
+   to stream artifact if it is locally available, otherwise it will attempt to redownload it from
+   the known upstream urls. In case upstream stopped serving the corresponding file , Pulp won't be
```
dralley, line 70: Extra space before the comma

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,76 @@
...
+     }
+
+As a result of this request, disk space will be freed-up for:
+
+  * artifacts that are exclusive to the list of provided repos. The content of the repository
+    versions will not change and no repository versions will be created or removed.
+  * an optional ``repo_versions_keeplist`` parameter can be specified that will contain list of
+    repo version hrefs which will be excluded from the artifact removal.
+  * by default, the task will remove only those artifacts from the content that were synchronized
+    into Pulp.
```
dralley, line 64: The meaning of this isn't totally clear, do you mean content that is already part of a repository version?

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,76 @@
...
+
+   The task will clean up artifacts regardless of the download policy. The content app will be able
+   to stream artifact if it is locally available, otherwise it will attempt to redownload it from
+   the known upstream urls. In case upstream stopped serving the corresponding file, Pulp won't be
+   able to download and serve it.
+
+.. warning::
+
+    For uploaded content that, as a rule, does not have a remote source url, a ``force`` flag must
+    be specified. This, however, will lead to a non-consumable content afterwards.
```
dralley, line 76: @ipanova Do you think it would be reasonable to prevent content with no Artifact and no RemoteArtifact from being added to new repository versions via one of the validation hooks?

docs/workflows/reclaim-disk-space.rst
ipanova: i mean content that was synchronized in pulp. there are just 2 ways how to get content into pulp - you whether upload it or synchronize it.

dralley: I think it would make more sense to reverse it, i.e. "by default, the task will *not* remove artifacts from content that were uploaded directly into Pulp"

docs/workflows/reclaim-disk-space.rst
ipanova: don't we fall into the same category of the issue we have with the removal of remote, remoteartifacts are being removed as well? in case it was an on_demand content the ca will not have either artifact nor remoteartifact?

ipanova: what i am trying to say is whatether we'd decide to do with such content that does not have RA and artifact it should be done in a separate PR after figuring out what to do with https://hackmd.io/_3gsUVdyQwy50Nc5pMXN-g
but yeah, you've opened the pandora box here by asking this question :D

dralley: @ipanova I think it can be left for #8305, as long as we actually address #8305 soon :)

ipanova: #soon

ipanova: i have added a comment to that issue with your proposal of the validation hooks so it can be discussed whenever relevant.

docs/workflows/reclaim-disk-space.rst
dralley: Would you mind filing an issue to do that?

docs/workflows/reclaim-disk-space.rst
dralley: Thanks!

docs/workflows/reclaim-disk-space.rst
ipanova: sure, if this is more readable i'll update.

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -0,0 +1,57 @@
...
+    ca_qs = ContentArtifact.objects.filter(content__in=c_reclaim_qs.values("pk"))
+    if not force:
+        ca_qs = ca_qs.filter(remoteartifact__isnull=False)
+    artifact_pks = set()
+    for ca in ca_qs:
+        artifact_pks.add(ca.artifact.pk)
+        ca.artifact = None
+
+    ContentArtifact.objects.bulk_update(objs=ca_qs, fields=["artifact"], batch_size=1000)
+    # respect timestamp when it will be added
```
daviddavis, line 33: Is this is a reference to `timestamp_of_interest`? I probably wouldn't respect that timestamp here. For one thing, it only applies to orphan cleanup and I wouldn't consider this to be part of orphan cleanup.

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -0,0 +1,57 @@
...
+        code="reclaim-space.artifact",
+        done=0,
+        state="running",
+    )
+    progress_bar.save()
+
+    counter = 0
+    interval = 100
+    for artifact in artifacts_to_delete.iterator():
+        # we need to manually call delete() because it cleans up the file on the filesystem
```
daviddavis, line 47: :+1:

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -27,11 +27,14 @@ def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
...
         ca_qs = ca_qs.filter(remoteartifact__isnull=False)
     artifact_pks = set()
-    for ca in ca_qs:
-        artifact_pks.add(ca.artifact.pk)
-        ca.artifact = None
-
-    ContentArtifact.objects.bulk_update(objs=ca_qs, fields=["artifact"], batch_size=1000)
+    ca_to_update = []
+    for ca in ca_qs.iterator():
+        if not ca.content.cast().protected_from_reclaim:
```
dralley, line 32: : /

Is there a chance we could do what we do in the `QueryExistingContent` stage?  It could be simpler since we would be dealing with `pk` instead of `natural_key`

mdellweg:
I believe the `PROTECT_FROM_RECLAIM` shoudhave a note in `CHANGES/plugin_api`.

ipanova:
@daviddavis @dralley can I get a final review/approval? I have submitted PRs to the plugins which depend on this change

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -0,0 +1,70 @@
...
+    Content,
+    ContentArtifact,
+    ProgressReport,
+    PublishedMetadata,
+    Repository,
+    RepositoryVersion,
+)
+
+
+def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
```
daviddavis, line 12: Some docstring here might be helpful so one doesn't have to go look at the serializer to understand the args.

daviddavis (approved):

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,76 @@
...
+
+    Pulp's reclaim disk space feature is in tech-preview and may change in backwards incompatible
+    ways in future releases.
+
+
+Pulp provides the ability to reclaim disk space for:
+
+  * content that one no longer needs to serve but would like to keep in the repo for historical
+    reasons.
+  * repos that have on_demand download policy and one would want to clear out downloaded files for
```
dralley, line 16: s/have/were synced with

docs/workflows/reclaim-disk-space.rst
```diff
@@ -0,0 +1,76 @@
...
+            "/pulp/api/v3/repositories/rpm/rpm/b3a6674d-181c-4e72-9412-7cbc747480ad/"
+        ],
+        "started_at": "2021-07-16T15:36:20.370245Z",
+        "state": "completed",
+        "task_group": null,
+        "worker": "/pulp/api/v3/workers/ccc132c4-0445-4f55-b370-32c3662dce3c/"
+     }
+
+As a result of this request, disk space will be freed-up for:
+
```
dralley, line 58: Maybe we should get rid of the list and switch it to paragraph form?  Points two and three don't match with the caption above.

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -0,0 +1,83 @@
...
+    ContentArtifact.objects.bulk_update(objs=ca_to_update, fields=["artifact"], batch_size=1000)
+    artifacts_to_delete = Artifact.objects.filter(pk__in=artifact_pks)
+    progress_bar = ProgressReport(
+        message="Reclaim disk space",
+        total=artifacts_to_delete.count(),
+        code="reclaim-space.artifact",
+        done=0,
+        state="running",
+    )
+    progress_bar.save()
```
dralley, line 68: Why not use as a context manager?

dralley (approved):

PR #1456, opened 2021-07-01
Fix changelog for 3.14.0

dkliban:
This change also adds a script that will allow the new changelog to be published during a re-run of 3.14.0 release workflow.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,11 @@ on:
       release:
         description: "Release tag (e.g. 3.2.1)"
         required: true
+      before_script:
+        description: |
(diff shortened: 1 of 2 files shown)


mdellweg:
I'm unsure how this script is going to survive the course of time. Wouldn't it be needed on all release branches?

dkliban:
> I'm unsure how this script is going to survive the course of time. Wouldn't it be needed on all release branches?

This script is going to be removed from this branch after we re-publish the docs. I filed an issue to improve this: https://github.com/pulp/plugin_template/issues/433 

I am going to work on this issue now instead of merging this script and then reverting it.

mdellweg (approved):
Approving for not being a permanent solution.

PR #1463, opened 2021-07-05
Fix syncs of repos from file:// urls

dralley:
closes: #9003
https://pulp.plan.io/issues/9003

### DIFF
--- a/CHANGES/9003.bugfix
+++ b/CHANGES/9003.bugfix
@@ -1,0 +1,1 @@
+Fixed a regression preventing syncs from file:// urls.
--- a/pulpcore/download/base.py
+++ b/pulpcore/download/base.py
@@ -76,6 +76,8 @@ class BaseDownloader:
         expected_digests=None,
         expected_size=None,
         semaphore=None,
+        *args,
+        **kwargs,
     ):
         """
         Create a BaseDownloader object. This is expected to be called by all subclasses.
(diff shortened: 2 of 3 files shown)


dralley:
See: https://pulp.plan.io/issues/9003#note-5

bmbouter:
If you can look into writing a test for this that would be really great.

ianballou:
I patched this into my Katello with pulpcore 3.14 and I'm seeing this error in my tests:
```
Katello::Errors::Pulp3Error: [Errno 2] No such file or directory: '/var/lib/pulp/sync_imports/test_repos/zoo/repodata/repomd.xml.asc'
```

There is a `repomd.xml` but no `repomd.xml.asc` on my filesystem.

dralley:
Thanks @ianballou.  That would be an RPM plugin bug - we expect that these files might not always exist but we check for them because there's no manifest that enumerates them otherwise.  The case where the files fail with 404 is handled but not FileNotFound.

https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/tasks/synchronizing.py#L639-L647

That's an easy fix and I can address it in a few minutes.

ianballou:
@dralley I made an issue for the bug I hit above: https://pulp.plan.io/issues/9021

daviddavis (approved):

bmbouter (approved):
We gotta figure the testing out, but I'm hearing we gotta merge this asap. The fix part looks good so lgtm.

PR #1465, opened 2021-07-06
Fix disk usage during the synchronization

lubosmj:

### DIFF
--- a/pulpcore/plugin/stages/api.py
+++ b/pulpcore/plugin/stages/api.py
@@ -81,7 +81,7 @@ class Stage:
             log.debug("%(name)s - next: %(content)s.", {"name": self, "content": content})
             yield content
 
-    async def batches(self, minsize=500):
+    async def batches(self, minsize=250):
         """
         Asynchronous iterator yielding batches of :class:`DeclarativeContent` from `self._in_q`.
 
--- a/CHANGES/8295.bugfix
+++ b/CHANGES/8295.bugfix
@@ -1,0 +1,1 @@
+Improved disk usage during the synchronization.
--- a/CHANGES/plugin_api/8295.feature
+++ b/CHANGES/plugin_api/8295.feature
@@ -1,0 +1,2 @@
+Enabled plugin writers to specify whether source files used for creating artifacts will be
+removed while running the stage ``ArtifactSaver`` in the sync pipeline.
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -1,4 +1,6 @@
 import asyncio
+import os
+
 from collections import defaultdict
 from gettext import gettext as _
 import logging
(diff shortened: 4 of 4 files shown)


pulpcore/plugin/stages/api.py
```diff
@@ -81,7 +81,7 @@ async def run(self):
             log.debug("%(name)s - next: %(content)s.", {"name": self, "content": content})
             yield content
 
-    async def batches(self, minsize=500):
+    async def batches(self, minsize=250):
```
lubosmj, line 84: This value needs to be reevaluated. Having the argument `minsize` set to 500 by default leads to higher disk usage during the synchronization because the pipeline is somehow "blocked" until the `batches` buffer is filled up. For example, the stage `ArtifactSaver` will be run only after downloading 2GB of data (depends on a remote repository).

Other values probably need to be reevaluated as well. See the @fao89's comments: https://github.com/pulp/pulpcore/pull/440/files.

lubosmj:
Seems like there is a design flaw in the synchronization process because it is not possible to determine which downloaded files need to be deleted and when these files need to be deleted. Usually, the procedure is as follows:
1. Check for existing artifacts.
2. Download missing artifacts as raw files (`ArtifactDownloader`).
3. Create artifacts from the raw files (`ArtifactSaver`).
4. Associate all artifacts.
5. ...

The stages `ArtifactDownloader` and `ArtifactSaver` do not share any information and we lose the knowledge about downloaded files right after entering the fourth stage. Thus, we have no idea what files to safely remove. I think we should either merge the stages `ArtifactDownloader` and `ArtifactSaver` or create some triggers which wrap temporarily downloaded files. This will eventually allow us to delete the files when they are not needed anymore. Both cases seem to be only workarounds.

bmbouter:
The saving of an artifact can occur outside of the stages API, e.g. in FirstStage for example. Fixing it here doesn't seem to address the whole issue. I believe the issue is in our FileField and StorageBackend code which is the underlying code moving files into place that this code calls into at `save()` time.

I'm confused because this is the same feedback I gave via Matrix on Friday.

lubosmj:
@bmbouter, I think using `FileField` itself does not resolve the issue. This would mean that `FileField` will need to wrapped by a wrapper that will delete a temporary file after reading it and storing it in the storage. In any case, there **must** be somewhere a temporary file () which has to be removed, eventually.

lubosmj:
The root cause of the issue basically lies here: https://github.com/pulp/pulpcore/blob/354383883032277e7a1f7dc7ddf2dc0a5bc40fad/pulpcore/download/base.py#L122.

lubosmj:
I am converting this PR to a draft since the referenced issue and the issue https://pulp.plan.io/issues/9000 might be related. One common change could resolve both of them.

lubosmj:
The root cause of the problem was hidden in our storage provider. I am closing this PR in favour of https://github.com/pulp/pulpcore/pull/1498. The issue https://pulp.plan.io/issues/9000 will be addressed in a separate PR.

PR #1467, opened 2021-07-07
Validate objects in GroupObjectPermissionViewSet

lubosmj:
closes #8500

### DIFF
--- a/pulpcore/app/viewsets/user.py
+++ b/pulpcore/app/viewsets/user.py
@@ -1,9 +1,9 @@
 from gettext import gettext as _
-import uuid
 
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group, Permission
 from django.core.exceptions import FieldError
+from django.db import IntegrityError
 from django.shortcuts import get_object_or_404
 from django_filters.rest_framework import DjangoFilterBackend
 from drf_spectacular.utils import extend_schema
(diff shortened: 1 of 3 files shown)


pulpcore/app/viewsets/user.py
```diff
@@ -269,13 +270,15 @@ def get_object_pk(self, request):
...
-            uuid.UUID(obj_pk)
-        except (AttributeError, ValueError):
-            raise ValidationError(_("Invalid value for 'obj': {obj}").format(request.data["obj"]))
+            match = resolve(obj_url)
+            obj = NamedModelViewSet.get_resource(obj_url, match.func.cls.queryset.model)
+        except (Resolver404, ValidationError):
+            raise ValidationError(_("Invalid value for 'obj': {}.").format(obj_url))
 
-        return obj_pk
+        return obj.pk
```
dralley, line 281: Can this be replaced with https://github.com/pulp/pulpcore/blob/master/pulpcore/app/viewsets/base.py#L190-L221 or is the use case a little different?

lubosmj: I am leaving this untouched because then I would need to check for a valid pk in the method `create`. I find this way of parsing the pk more robust since we can determine invalid pks straight away.

dralley:
I can't speak to the permissions part (no experience with RBAC yet) but the rest looks great, good job @lubosmj  :+1:

mdellweg (approved):
LGTM

dkliban (approved):

PR #1473, opened 2021-07-08
Adds Range header support to content app

bmbouter:
This unskips the `Range` header and adds support for it to the content
app.

closes #8865

### DIFF
--- a/CHANGES/8865.bugfix
+++ b/CHANGES/8865.bugfix
@@ -1,0 +1,3 @@
+Fixed bug where content app would not respond to ``Range`` HTTP Header in requests when
+``remote.policy`` was either ``on_demand`` or ``streamed``. For example this request is used by
+Anaconda clients.
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -665,7 +665,7 @@ class Handler:
                 return response
 
             except (ClientResponseError, UnsupportedDigestValidationError) as e:
-                log.warn(
+                log.warning(
                     _("Could not download remote artifact at '{}': {}").format(
                         remote_artifact.url, str(e)
                     )
@@ -809,15 +809,50 @@ class Handler:
 
         remote = await loop.run_in_executor(None, cast_remote_blocking)
 
-        async def handle_headers(headers):
+        range_start, range_stop = request.http_range.start, request.http_range.stop
+        if range_start or range_stop:
+            response.set_status(206)
+
+        async def handle_response_headers(headers):
             for name, value in headers.items():
-                if name.lower() in self.hop_by_hop_headers:
+                lower_name = name.lower()
+                if lower_name in self.hop_by_hop_headers:
                     continue
+
+                if response.status == 206 and lower_name == "content-length":
+                    range_bytes = int(value)
+                    start = 0 if range_start is None else range_start
+                    stop = range_bytes if range_stop is None else range_stop
+
+                    range_bytes = range_bytes - range_start
+                    range_bytes = range_bytes - (int(value) - stop)
+                    response.headers[name] = str(range_bytes)
+
+                    response.headers["Content-Range"] = "bytes {0}-{1}/{2}".format(
+                        start, stop - start + 1, int(value)
+                    )
+                    continue
+
                 response.headers[name] = value
             await response.prepare(request)
 
+        data_size_handled = 0
+
         async def handle_data(data):
-            await response.write(data)
+            nonlocal data_size_handled
+            if range_start or range_stop:
+                start_byte_pos = 0
+                end_byte_pos = len(data)
+                if range_start:
+                    start_byte_pos = max(0, range_start - data_size_handled)
+                if range_stop:
+                    end_byte_pos = min(len(data), range_stop - data_size_handled)
+
+                data_for_client = data[start_byte_pos:end_byte_pos]
+                await response.write(data_for_client)
+                data_size_handled = data_size_handled + len(data)
+            else:
+                await response.write(data)
             if remote.policy != Remote.STREAMED:
                 await original_handle_data(data)
 
@@ -826,7 +861,7 @@ class Handler:
                 await original_finalize()
 
         downloader = remote.get_downloader(
-            remote_artifact=remote_artifact, headers_ready_callback=handle_headers
+            remote_artifact=remote_artifact, headers_ready_callback=handle_response_headers
         )
         original_handle_data = downloader.handle_data
         downloader.handle_data = handle_data
--- a/pulpcore/tests/functional/api/using_plugin/test_distributions.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_distributions.py
@@ -4,10 +4,11 @@ import hashlib
 import unittest
 from urllib.parse import urljoin
 
-from pulp_smash import api, config, utils
+from pulp_smash import api, cli, config, utils
 from pulp_smash.pulp3.bindings import delete_orphans
 from pulp_smash.pulp3.utils import (
     download_content_unit,
+    download_content_unit_return_requests_response,
     gen_distribution,
     gen_repo,
     get_content,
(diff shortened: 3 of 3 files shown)


pulpcore/content/handler.py
```diff
@@ -436,6 +436,8 @@ async def _match_and_stream(self, path, request):
             :class:`aiohttp.web.StreamResponse` or :class:`aiohttp.web.FileResponse`: The response
                 streamed back to the client.
         """
+        # import pydevd_pycharm
+        # pydevd_pycharm.settrace('localhost', port=29437, stdoutToServer=True, stderrToServer=True)
```
dralley, line 440: ^

pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -389,9 +388,11 @@ def do_test_content_served(self):
         self.assertEqual(len(pulp_manifest), FILE_FIXTURE_COUNT, pulp_manifest)
 
     def do_range_request_download_test(self):
+        # import pydevd_pycharm
+        # pydevd_pycharm.settrace('localhost', port=29437, stdoutToServer=True, stderrToServer=True)
         file_path = "1.iso"
 
-        headers = {"Range": "bytes=0-9"}  # first 10 bytes
+        headers = {"Range": "bytes=2-11"}  # first 10 bytes
```
dralley, line 395: Fix the comment, and it would be good to do a second request with a different range to make sure the potential caching issues @gerrod3 mentioned aren't there.  Likewise a range request after a non-range-request for the same file.

pulpcore/content/handler.py
bmbouter: Yeah this is not ready for review yet. I want to see what the CI says. I'll convert to draft.

pulpcore/tests/functional/api/using_plugin/test_distributions.py
bmbouter: Good ideal. I'm adding a bunch of tests actually.

pulpcore/content/handler.py
```diff
@@ -809,15 +809,46 @@ def cast_remote_blocking():
 
         remote = await loop.run_in_executor(None, cast_remote_blocking)
 
-        async def handle_headers(headers):
+        rng = request.http_range
```
dralley, line 812: What I would maybe do instead is `range_start, range_stop = request.http_range.start, request.http_range.stop`, because `rng` mentally maps to "random number generator" for me

pulpcore/content/handler.py
```diff
@@ -809,15 +809,46 @@ def cast_remote_blocking():
...
+                    continue
+
+                if response.status == 206 and lower_name == "content-length":
+                    range_bytes = int(value)
+                    start = 0 if rng.start is None else rng.start
+                    stop = range_bytes if rng.stop is None else rng.stop
+
+                    range_bytes = range_bytes - rng.start
+                    range_bytes = range_bytes - (int(value) - stop)
+                    response.headers[name] = str(range_bytes)
```
dralley, line 829: I'm not sure I understand this part.  Why are we rewriting the content-length?

pulpcore/content/handler.py
gerrod3: I agree with associating `rng` with random number generator.

CHANGES/8865.bugfix
```diff
@@ -0,0 +1,2 @@
+Fixed bug where content app would not respond to `Range` HTTP Header in requests, e.g. from Anaconda
+clients.
```
gerrod3, line 2: Is it worth mentioning that this only occurred when content wasn't present in Pulp yet, i.e. `remote.policy != Remote.IMMEDIATE`?

gerrod3 (commented):
I think a test(s) for ranges with files that would be chunked is needed to verify this fix works correctly (see comments below). Use this fixture https://fixtures.pulpproject.org/file-chunked/. Also an additional test for streamed content should be added.

CHANGES/8865.bugfix
dralley: I think so, personally

pulpcore/content/handler.py
bmbouter: I'll change this now. ty

pulpcore/content/handler.py
bmbouter: If we pass the `Content-Length` header back unmodified then even though say the range is asking for 10 bytes of 1024, and we're only returning 10 bytes, the `Content-Length` would show 1024 incorrectly. Also we do this here because some server's don't return the `Content-Length` at which point we don't know enough to set it.

pulpcore/content/handler.py
```diff
@@ -809,15 +809,46 @@ def cast_remote_blocking():
...
+                end_byte_pos = len(data)
+                if rng.start:
+                    start_byte_pos = max(0, rng.start - downloader._size)
+                if rng.stop:
+                    end_byte_pos = min(len(data), rng.stop - downloader._size)
+
+                data_for_client = data[start_byte_pos:end_byte_pos]
+                await response.write(data_for_client)
+            else:
+                await response.write(data)
```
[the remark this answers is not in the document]
bmbouter: I'll add a test using the fixture to cause chunking. I think it does because when I developed the patch I modified chunk size [here](https://github.com/pulp/pulpcore/blob/master/pulpcore/download/http.py#L210) to be 3 bytes for files of size 1024. Adding a test from the start would have been better than that tho.

pulpcore/content/handler.py
```diff
@@ -809,15 +809,46 @@ def cast_remote_blocking():
...
             if remote.policy != Remote.STREAMED:
                 await original_handle_data(data)
```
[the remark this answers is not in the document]
bmbouter: Now this I think would not work. I'll add a streamed test and also use the local var instead of `._size`. ty!

CHANGES/8865.bugfix
bmbouter: I will update this also. Ty!

pulpcore/content/handler.py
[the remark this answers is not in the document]
bmbouter: I added a `test_content_served_streamed_with_range_request_valid_middle_range` and it seemed to work as-is. Hmmmm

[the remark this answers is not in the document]
bmbouter: It's because it's not using chunking! I'm switching all of these tests to use the chunking fixture...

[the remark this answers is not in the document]
bmbouter: Yes this new test is failing. I'm fixing it.

pulpcore/content/handler.py
gerrod3: Not sure if this math for these values are right. Testing with httpie, `http HEAD https://fixtures.pulpproject.org/file/1.iso Range:bytes=40-50` returns `content-length: 11 content-range: bytes 40-50/1024`. Is either the range start or stop auto decremented by `aiohttp`?

daviddavis (approved):

pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -329,24 +330,113 @@ def test_content_served_immediate(self):
...
+        cli_client = cli.Client(cfg)
+        storage = utils.get_pulp_setting(cli_client, "DEFAULT_FILE_STORAGE")
+        if storage != "pulpcore.app.models.storage.FileSystem":
+            self.skipTest("The S3 test API project doesn't handle invalid Range values correctly")
+        self.setup_download_test(
+            "immediate", url="https://fixtures.pulpproject.org/file-chunked/PULP_MANIFEST"
+        )
 
-    def setup_download_test(self, policy):
+        self.assertRaises(
```
dralley, line 397: We should ensure that HTTP 416 is being raised, which is the correct status code.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/416

pulpcore/content/handler.py
gerrod3: Looking at the tests constants below, they seem to be correct after manually checking them against httpie and the fixtures, so I think this math must be correct if the tests pass even if it's not intuitive from the code.

pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -329,24 +330,113 @@ def test_content_served_immediate(self):
...
+        self.setup_download_test("streamed")
+        self.do_test_content_served()
 
-    def test_content_served_immediate_with_range_request(self):
+    def test_content_served_immediate_with_range_request_inside_one_chunk(self):
         """Assert that downloaded content can be properly downloaded with range requests."""
-        self.setup_download_test("immediate")
-        self.do_range_request_download_test()
+        self.setup_download_test(
+            "immediate", url="https://fixtures.pulpproject.org/file-chunked/PULP_MANIFEST"
```
dralley, line 341: Add this URL to constants.py and import it to avoid duplicates.

pulpcore/content/handler.py
bmbouter: Yeah it's a little confusing but aiohttp does modify it by one I believe because start is inclusive and end is exclusive. https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.BaseRequest.http_range

Take this test for example, it [asserts the content-length is 1200401](https://github.com/pulp/pulpcore/pull/1473/files#diff-c2f1bc66265b714bc14821751e842441c1d63c1b629c807144bf1fdaf162ad1aR353) and when I run `http HEAD https://fixtures.pulpproject.org/file-chunked/1.iso Range:bytes=1048176-2248576` I get

```
HTTP/1.1 206 Partial Content
cache-control: private
content-length: 1200401
content-range: bytes 1048176-2248576/10485760
content-type: application/octet-stream
date: Fri, 09 Jul 2021 20:35:02 GMT
etag: "60d0dd9f-a00000"
last-modified: Mon, 21 Jun 2021 18:42:39 GMT
server: nginx/1.14.1
set-cookie: dd6ec0a439e94ec1afa8091bd8cb5f4f=84e1263ecb996dab7f6f4ae499835116; path=/; HttpOnly; Secure; SameSite=None
```

I really really like the thoroughness and attention to detail you're showing here!

dralley (changes requested):

pulpcore/content/handler.py
dralley: @gerrod3 Since the ranges start from 0, index 10 is the 11th byte, and likewise in this instance.  So I think the math is correct.

See the example given in the middle of page 13 https://datatracker.ietf.org/doc/html/rfc7233#section-4.2

dralley (approved):

PR #1478, opened 2021-07-10
Make auto-distribute feature more intuitive

dralley:
closes: #9039
https://pulp.plan.io/issues/9039

### DIFF
--- a/pulpcore/app/models/publication.py
+++ b/pulpcore/app/models/publication.py
@@ -125,13 +125,14 @@ class Publication(MasterModel):
             Deletes the Task.created_resource when complete is False.
         """
         with transaction.atomic():
+            # invalidate cache
             if settings.CACHE_ENABLED:
                 # Find any publications being served directly
                 base_paths = self.distribution_set.values_list("base_path", flat=True)
                 # Find any publications being served indirectly by auto-distribute feature
                 versions = self.repository.versions.all()
                 pubs = Publication.objects.filter(repository_version__in=versions, complete=True)
-                publication = pubs.latest("repository_version", "-pulp_created")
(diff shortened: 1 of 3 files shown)


pulpcore/app/models/publication.py
```diff
@@ -175,13 +175,22 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
-                raise
+            with transaction.atomic():
+                try:
+                    self.finalize_new_publication()
+                    self.complete = True
+                    self.save()
+                except Exception:
+                    self.delete()
+                    raise
+                else:
```
dralley, line 186: @gerrod3 Since the relationship is indirect, I didn't see an obvious way to accomplish this with the hooks.  Are you aware of one?

gerrod3 (approved):

daviddavis (approved):

pulpcore/app/models/publication.py
```diff
@@ -183,6 +184,14 @@ def __exit__(self, exc_type, exc_val, exc_tb):
                 self.delete()
                 raise
 
+            # invalidate cache
+            if settings.CACHE_ENABLED:
+                base_paths = Distribution.objects.filter(
+                    repository=self.repository_version.repository
+                ).values_list("base_path", flat=True)
+                if base_paths:
+                    Cache().delete(base_key=base_paths)
```
dkliban, line 193: It seems like this query won't work. i expected to see something like base_key__in=base_paths

dralley: It's not a query it's a method call on a non-queryset object.

That tripped me up, too, when doing the initial reviews.

dkliban (approved):

PR #1480, opened 2021-07-12
Validate absolute pathnames in remotes' URLs

lubosmj:
Before this change, it was not possible to determine why did the synchronization fail when a user provided a seemingly valid URL. This commit also adds more relevant information to the error message.

Having set `ALLOWED_EXPORT_PATHS` to `["/tmp", "/home/vagrant/test"]`, the following error messages are shown:

```
$ pulp file remote create --name test --url file://error/vagrant/test/centos-7/PULP_MANIFEST
Error: {"url":["The path 'error/vagrant/test/centos-7/PULP_MANIFEST' needs to be an absolute pathname."]}

$ pulp file remote create --name test --url file:///error/vagrant/test/centos-7/PULP_MANIFEST
Error: {"url":["The path '/error/vagrant/test/centos-7/PULP_MANIFEST' is not in allowed import paths"]}
```

closes #9080

### DIFF
--- a/pulpcore/app/serializers/repository.py
+++ b/pulpcore/app/serializers/repository.py
@@ -205,14 +205,14 @@ class RemoteSerializer(ModelSerializer):
         required=False,
     )
 
-    def validate_url(self, value):
+    def validate_url(self, url):
         """
         Check if the 'url' is a ``file://`` path, and if so, ensure it's an ALLOWED_IMPORT_PATH.
 
         The ALLOWED_IMPORT_PATH is specified as a Pulp setting.
 
         Args:
-            value: The user-provided value for 'url' to be validated.
+            url: The user-provided value for 'url' to be validated.
 
         Raises:
             ValidationError: When the url starts with `file://`, but is not a subfolder of a path in
@@ -221,16 +221,24 @@ class RemoteSerializer(ModelSerializer):
         Returns:
             The validated value.
         """
-        if not value.lower().startswith("file://"):
-            return value
+        if not url.lower().startswith("file://"):
+            return url
 
-        user_path = value[7:]
+        user_path = url[7:]
+        if not os.path.isabs(user_path):
+            raise serializers.ValidationError(
+                _("The path '{}' needs to be an absolute pathname.").format(user_path)
+            )
+
+        user_provided_realpath = os.path.realpath(user_path)
 
         for allowed_path in settings.ALLOWED_IMPORT_PATHS:
-            user_provided_realpath = os.path.realpath(user_path)
             if user_provided_realpath.startswith(allowed_path):
-                return value
-        raise serializers.ValidationError(_("url '{}' is not an allowed import path").format(value))
+                return url
+
+        raise serializers.ValidationError(
+            _("The path '{}' does not start with any of the allowed import paths").format(user_path)
+        )
 
     def validate_proxy_url(self, value):
         """
--- a/pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
@@ -396,3 +396,50 @@ class CreatePulpLabelsRemoteTestCase(unittest.TestCase):
         remote = self.api_client.post(FILE_REMOTE_PATH, data=remote_attrs)
         self.addCleanup(self.remotes_api.delete, remote["pulp_href"])
         self.assertEqual(remote["pulp_labels"], self.pulp_labels)
+
+
+class RemoteFileURLsValidationTestCase(unittest.TestCase):
+    """A test case that verifies the validation of remotes' URLs."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Initialize class-wide variables"""
+        cls.cfg = config.get_config()
+
+        cls.api_client = api.Client(cls.cfg, api.json_handler)
+        cls.file_client = FileApiClient(cls.cfg.get_bindings_config())
+        cls.remotes_api = RemotesFileApi(cls.file_client)
+
+    def test_invalid_absolute_pathname(self):
+        """Test the validation of an invalid absolute pathname."""
+        remote_attrs = {
+            "name": utils.uuid4(),
+            "url": "file://error/path/name",
+        }
+        self.raise_for_invalid_request(remote_attrs)
+
+    def test_invalid_import_path(self):
+        """Test the validation of an invalid import pathname."""
+        remote_attrs = {
+            "name": utils.uuid4(),
+            "url": "file:///error/path/name",
+        }
+        self.raise_for_invalid_request(remote_attrs)
+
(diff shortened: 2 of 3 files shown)


pulpcore/app/serializers/repository.py
```diff
@@ -221,16 +221,23 @@ def validate_url(self, value):
...
+        if not url.lower().startswith("file://"):
+            return url
+
+        user_path = url[7:]
+        user_provided_realpath = os.path.realpath(user_path)
+
+        if any(
+            user_provided_realpath.startswith(allowed_path)
+            for allowed_path in settings.ALLOWED_IMPORT_PATHS
+        ):
```
dralley, line 233: To be honest the `for` loop felt a little cleaner.

lubosmj: You can even read it as a valid sentence: "If any user-provided real path starts with (an) allowed path for allowed path in Allowed Import Paths, return URL"! :+1:

dralley: No strong feelings, it's just awkward because of the indentation

dralley:
```$ pulp file remote create --name $REMOTE_NAME --url file://home/vagrant/test/centos-7/PULP_MANIFEST
OK

$ pulp file repository sync --name $REPO_NAME --remote $REMOTE_NAME
Started background task /pulp/api/v3/tasks/b4748fc7-cec9-425d-b39e-46fcb9097153/
Error: Task /pulp/api/v3/tasks/b4748fc7-cec9-425d-b39e-46fcb9097153/ failed: '[ErrorDetail(string="The url 'file://home/vagrant/test/centos-7/PULP_MANIFEST' is not in an allowed import path. The path is interpreted as '/var/lib/pulp/tmp/73438@pulp3-source-fedora33.localhost.example.com/tmp2ft_wsar/home/vagrant/test/centos-7/PULP_MANIFEST'", code='invalid')]'
```

So I understand why the first remote failed to be created (because `file://error/vagrant/test/centos-7/PULP_MANIFEST` doesn't match any of the allowed paths) but I don't understand why the second one was allowed in the first place.  

If I understand correctly, the problem is that `file://` is part of the URL, so the actual path is interpreted as `home/vagrant/...` instead of `/home/vagrant/...`.  That feels like something that we could be catching up front?

dralley:
If Katello uses this functionality at all I think we should create an issue and backport it.  Even though it's not a real bugfix, it's an improvement to support-ability, which is worthwhile.

dralley:
> I think maybe clearest way we could improve the error message would be to strip the file:// prefix and present the user with the raw path that was extracted from their URL.

Discussion on Matrix:  We should probably automatically reject all relative paths with a specific error message, and we can provide the extracted path to make it clear to the user.

The matching should have rejected `home/vagrant/test/...` since it is not the same as `/home/vagrant/test/...` -- hopefully the rejection of relative paths should handle this case automatically.

pulpcore/app/serializers/repository.py
```diff
@@ -221,16 +221,24 @@ def validate_url(self, value):
...
-            return value
+        if not url.lower().startswith("file://"):
+            return url
 
-        user_path = value[7:]
+        user_path = url[7:]
+        if not os.path.isabs(user_path):
+            raise serializers.ValidationError(
+                _("The path '{}' needs to be an absolute pathname.").format(user_path)
+            )
```
dralley, line 231: :+1:

pulpcore/app/serializers/repository.py
```diff
@@ -221,16 +221,24 @@ def validate_url(self, value):
...
 
         for allowed_path in settings.ALLOWED_IMPORT_PATHS:
-            user_provided_realpath = os.path.realpath(user_path)
             if user_provided_realpath.startswith(allowed_path):
-                return value
-        raise serializers.ValidationError(_("url '{}' is not an allowed import path").format(value))
+                return url
+
+        raise serializers.ValidationError(
+            _("The path '{}' is not in allowed import paths").format(user_path)
```
dralley, line 240: What do you think about "The path '{}` is not a subdirectory of one of the allowed import paths"

pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
```diff
@@ -396,3 +396,50 @@ def test_create_remote_using_form(self):
...
+        cls.file_client = FileApiClient(cls.cfg.get_bindings_config())
+        cls.remotes_api = RemotesFileApi(cls.file_client)
+
+    def test_invalid_absolute_pathname(self):
+        """Test the validation of an invalid absolute pathname."""
+        remote_attrs = {
+            "name": utils.uuid4(),
+            "url": "file://error/path/name",
+        }
+        self.raise_for_invalid_request(remote_attrs)
```
dralley, line 419: :+1:

pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
```diff
@@ -396,3 +396,50 @@ def test_create_remote_using_form(self):
...
+            self.addCleanup(self.remotes_api.delete, remote.pulp_href)
+
+        self.assertEqual(ae.exception.status, 400)
+
+    def test_valid_import_path(self):
+        """Test the creation of a remote after passing a valid URL."""
+        remote_attrs = {
+            "name": utils.uuid4(),
+            "url": "file:///tmp/good",
+        }
```
dralley, line 442: Perfect, thanks!

dralley (approved):

dralley:
@lubosmj Please squash the functional tests into the same commit, it makes cherry-picking easier.

pulpcore/app/serializers/repository.py
lubosmj: Actually, it is not a subdirectory. It is a full path to the resource.

dralley: The check is only that the path `startswith` one of the allowed paths.  For example if `/tmp` is allowed then any subdirectory of `/tmp` can be used.

dralley: e.g. https://github.com/pulp/pulp_rpm/pull/2034/files#diff-5a20483ca2f367cb948f26046e71ca30e3aac3e3622a02e4d10b0471966af4fdR150

lubosmj: How can `/tmp/test/centos-7/PULP_MANIFEST` be a subdirectory of `/tmp/`?

lubosmj:
> @lubosmj Please squash the functional tests into the same commit, it makes cherry-picking easier.

I squashed the commits, thanks for pointing it out.

pulpcore/app/serializers/repository.py
dralley: Fair point that an error message the describes "..../PULP_MANIFEST" as a subdirectory would be a little strange.  The updated wording is probably OK.

daviddavis (approved):

PR #1481, opened 2021-07-12
Reduce the cache entry expiration TTL

dralley:
If Redis goes down then its persisted data can get out of sync with
Pulp, because cache invalidation events cannot occur. We should mitigate
the impact by letting cache entries expire after a shorter period of
time.

closes: #8996
https://pulp.plan.io/issues/8996

### DIFF
--- a/CHANGES/8996.misc
+++ b/CHANGES/8996.misc
@@ -1,0 +1,1 @@
+Reduced the default expiration TTL of entries in the Pulp content app cache.
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -242,7 +242,7 @@ PROFILE_STAGES_API = False
 # https://docs.pulpproject.org/pulpcore/configuration/settings.html#pulp-cache
 CACHE_ENABLED = True
 CACHE_SETTINGS = {
-    "EXPIRES_TTL": 86400,
+    "EXPIRES_TTL": 600,  # 10 minutes
 }
 
 SPECTACULAR_SETTINGS = {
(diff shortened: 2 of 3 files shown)


pulpcore/app/settings.py
```diff
@@ -242,7 +242,7 @@
 # https://docs.pulpproject.org/pulpcore/configuration/settings.html#pulp-cache
 CACHE_ENABLED = True
 CACHE_SETTINGS = {
-    "EXPIRES_TTL": 86400,
```
dralley, line 245: 24 hours.

pulpcore/app/settings.py
```diff
@@ -242,7 +242,7 @@
...
+    "EXPIRES_TTL": 480,  # 8 minutes
```
dralley, line 245: According to @jlsherrill , most large users would have their clusters updating over a fairly short timeframe, so an expiration time measured in minutes is reasonable and would still result in high utilization.

We want to reduce the impact of any significant cache-invalidation events.  The ideal amount of time is up for debate though, and perhaps we should consider adding a little bit of randomness to amortize the cost of cache expiration events.

gerrod3 (approved):

ipanova (approved):
i'd probably lead towards 10 mins

daviddavis (approved):

CHANGES/8996.misc
```diff
@@ -0,0 +1 @@
+Reduced the default expiration TTL of entries in the Pulp content app cache.
```
daviddavis, line 1: I feel like this is maybe worth exposing to users.

dralley: We never advertised the default to begin with so I don't really think it's necessary.  However I've just added it to the configuration docs.

PR #1483, opened 2021-07-12
Use proxy auth from the Remote config.

dkliban:
fixes: #9024
https://pulp.plan.io/issues/9024

### DIFF
--- a/CHANGES/9024.bugfix
+++ b/CHANGES/9024.bugfix
@@ -1,0 +1,1 @@
+Use proxy auth from Remote config to download content from a remote repository.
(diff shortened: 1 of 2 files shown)


dralley (approved):

ipanova (approved):

daviddavis:
I imagine we probably want this in 3.14.2?

rochacbruno:
@daviddavis @dkliban @fao89 do we need this change on https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/downloaders.py#L134 ?

fao89:
> @daviddavis @dkliban @fao89 do we need this change on https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/downloaders.py#L134 ?

I believe so, nice catch!

PR #1484, opened 2021-07-13
Conditionally include Redis connection information

ekohl:
Redis is only used when cache is enabled or RQ tasking is used. In other cases, the connection is not needed. We've changed the Foreman installer to [only install Redis if needed](https://github.com/theforeman/puppet-pulpcore/commit/7a5543b1a47de8d76a1cad1fb76e12db7bbf8e45), just to stop wasting resources. However, this leads to warnings.

This PR changes the status view to only include Redis if it's used.

Right now there are no tests since I first want to discuss if this is even the right direction. That's also why this PR is a draft.

https://pulp.plan.io/issues/9070

### DIFF
--- a/CHANGES/9070.bugfix
+++ b/CHANGES/9070.bugfix
@@ -1,0 +1,2 @@
+Set Redis connection information in status to null unless it's used. Redis is
+needed for RQ tasking or content caching.
--- a/pulpcore/app/serializers/status.py
+++ b/pulpcore/app/serializers/status.py
@@ -74,6 +74,9 @@ class StatusSerializer(serializers.Serializer):
         help_text=_("Database connection information")
     )
 
-    redis_connection = RedisConnectionSerializer(help_text=_("Redis connection information"))
+    redis_connection = RedisConnectionSerializer(
+        required=False,
+        help_text=_("Redis connection information"),
+    )
 
     storage = StorageSerializer(required=False, help_text=_("Storage information"))
--- a/pulpcore/app/views/status.py
+++ b/pulpcore/app/views/status.py
@@ -57,7 +57,11 @@ class StatusView(APIView):
         for app in pulp_plugin_configs():
             versions.append({"component": app.label, "version": app.version})
 
-        redis_status = {"connected": self._get_redis_conn_status()}
+        if settings.CACHE_ENABLED or not settings.USE_NEW_WORKER_TYPE:
+            redis_status = {"connected": self._get_redis_conn_status()}
+        else:
+            redis_status = None
+
         db_status = {"connected": self._get_db_conn_status()}
 
         try:
--- a/pulpcore/tests/functional/api/test_status.py
+++ b/pulpcore/tests/functional/api/test_status.py
@@ -1,6 +1,7 @@
 """Test the status page."""
 import unittest
 
+from django.test import override_settings
 from jsonschema import validate
 from pulp_smash import api, cli, config, utils
 from pulp_smash.pulp3.constants import STATUS_PATH
@@ -91,6 +92,7 @@ class StatusTestCase(unittest.TestCase):
         """
         validate(status, self.status_response)
         self.assertTrue(status["database_connection"]["connected"])
+        self.assertIsNotNone(status["redis_connection"])
         self.assertTrue(status["redis_connection"]["connected"])
         self.assertNotEqual(status["online_workers"], [])
         self.assertNotEqual(status["versions"], [])
@@ -98,3 +100,12 @@ class StatusTestCase(unittest.TestCase):
             self.assertIsNotNone(status["storage"])
         else:
             self.assertIsNone(status["storage"])
+
+    @override_settings(CACHE_ENABLED=False, USE_NEW_WORKER_TYPE=True)
+    def verify_get_response_without_redis(self, status):
+        """Verify the response to an HTTP GET call when Redis is not used.
+
+        Verify that redis_connection is null
+        """
+        validate(status, self.status_response)
+        self.assertIsNone(status["redis_connection"])


ekohl:
@dralley I saw you're assigned to https://pulp.plan.io/issues/8997 and I think this can be considered a blocker to that effort. We're now running into this. Please have a look.

pulpcore/app/views/status.py
```diff
@@ -75,10 +74,12 @@ def get(self, request):
             "online_workers": online_workers,
             "online_content_apps": online_content_apps,
             "database_connection": db_status,
-            "redis_connection": redis_status,
             "storage": _disk_usage(),
         }
 
+        if settings.CACHE_ENABLED or settings.USE_NEW_WORKER_TYPE:
```
dralley, line 80: `or not`

dralley:
A few concerns:

* The status page will now change depending on what their settings declare, which is a little strange.  Most What if a downstream user is parsing the status - it's not clear what data they should be expecting back.  I think most status pages would simply list all components, optional or not.
* I thought (re: ehelms email) that Foreman / Katello already did require Redis running for other reasons (Dynflow / Sidekiq)?  Is that not the case?
* re: https://github.com/theforeman/puppet-pulpcore/commit/7a5543b1a47de8d76a1cad1fb76e12db7bbf8e45, if the goal is to only to avoid unnecessary resource usage, wouldn't the best solution be to install Redis but leave it disabled, and make it socket-activated, so that it only comes online if needed?  Otherwise it seems more difficult  than necessary to enable/disable caching.

ekohl:
> The status page will now change depending on what their settings declare, which is a little strange. What if a downstream user is parsing the status - it's not clear what data they should be expecting back. I think most status pages would simply list all components, optional or not.

This is indeed something I also thought about. However, I think the output format should clarify that it's optional then. Right now the output is (simplified):
```json
{
  "redis_connection": {
    "connected": false
  }
}
```

So what should it then become if it's optional?
```json
{
  "redis_connection": false
}
```

Or:
```json
{
  "redis_connection": {
    "connected": false,
    "used": false
  }
}
```

>     * I thought (re: ehelms email) that Foreman / Katello already did require Redis for other reasons (Dynflow / Sidekiq)?  Is that not the case?

Foreman does indeed have Redis (but we support an external Redis as well) but on content proxies there is no Foreman so no Redis. That means we'll reduce resources if we don't install it. Fewer services is generally also appreciated by security folks.

>     * if the goal is to only to avoid unnecessary resource usage, wouldn't the best solution be to install Redis but leave it disabled, and make it socket-activated, so that it only comes online if needed?  Otherwise it seems more difficult  than necessary to enable/disable caching, because the components aren't there.

Our install can handle conditional inclusion, but doesn't handle clean up. I like to socket activation suggestion but I'm not sure if Redis supports that today. I'm looking into unix sockets anyway (https://github.com/theforeman/puppet-pulpcore/pull/220) but it appears that now results in internal server errors. I'll need to dive into that.

daviddavis:
I agree that we want to avoid having option keys in our response. 

I have no objection to the two proposals you mention (just setting connection to false or adding a `used` key). Another option is to return `null` or `{}` for `redis_connection` if Pulp is not using redis.

dralley:
> Another option is to return null or {} for redis_connection if Pulp is not using redis. This would be consistent with the storage key which returns "storage": null when Pulp is not using a filesystem.

I'm in favor of this approach

ekohl:
I like `null`: it's very straight forward and feels natural to me. I've updated the PR to implement this and added a test to verify it. I'm relying on CI to run tests but I think this is ready.

pulpcore/app/serializers/status.py
```diff
@@ -74,6 +74,9 @@ class StatusSerializer(serializers.Serializer):
         help_text=_("Database connection information")
     )
 
-    redis_connection = RedisConnectionSerializer(help_text=_("Redis connection information"))
+    redis_connection = RedisConnectionSerializer(
+        required=False,
```
daviddavis, line 78: I'm assuming you can't return null unless required=False?

CHANGES/9070.bugfix
```diff
@@ -0,0 +1,2 @@
+Remove Redis connection information from status unless it's used, either by RQ
```
daviddavis, line 1: I guess this is perhaps still valid? Maybe better to say "Return null for redis connection key..."

pulpcore/app/serializers/status.py
dralley: I thought `required` was only used for creating resources.  It usually allows the non-presence of the entire key rather than affecting `null`ability

CHANGES/9070.bugfix
ekohl: I debated it but this morning I couldn't think of a better text. Technically it sort of still removes it but I agree explicitly saying that it's null is clearer.

pulpcore/app/serializers/status.py
ekohl: I must admit I didn't even check it and assumed it. Note that storage also already is optional.

ekohl: For completeness, from a current Pulp 3.14 without this patch. When I go to `/pulp/api/v3` using a browser (text/html):

```yaml
    StatusResponse:
      type: object
      description: Serializer for the status information of the app
      properties:
        versions:
          type: array
          items:
            $ref: '#/components/schemas/VersionResponse'
          description: Version information of Pulp components
        online_workers:
          type: array
          items:
            $ref: '#/components/schemas/WorkerResponse'
          description: List of online workers known to the application. An online
            worker is actively heartbeating and can respond to new work
        online_content_apps:
          type: array
          items:
            $ref: '#/components/schemas/ContentAppStatusResponse'
          description: List of online content apps known to the application. An online
            content app is actively heartbeating and can serve data to clients
        database_connection:
          allOf:
          - $ref: '#/components/schemas/DatabaseConnectionResponse'
          description: Database connection information
        redis_connection:
          allOf:
          - $ref: '#/components/schemas/RedisConnectionResponse'
          description: Redis connection information
        storage:
          allOf:
          - $ref: '#/components/schemas/StorageResponse'
          description: Storage information
      required:
      - database_connection
      - online_content_apps
      - online_workers
      - redis_connection
      - versions
```

Note `storage` is not in `required`. While it may describe inputs, it is the closest thing there is to also describe the output. That's why I think it still makes sense.

CHANGES/9070.bugfix
ekohl: Updated. Let me know what you think of it now.

daviddavis (approved):

pulpcore/app/serializers/status.py
dralley: Sounds fine to me

pulpcore/tests/functional/api/test_status.py
```diff
@@ -91,10 +92,20 @@ def verify_get_response(self, status):
...
+        self.assertIsNotNone(status["redis_connection"])
         self.assertTrue(status["redis_connection"]["connected"])
         self.assertNotEqual(status["online_workers"], [])
         self.assertNotEqual(status["versions"], [])
         if self.storage == "pulpcore.app.models.storage.FileSystem":
             self.assertIsNotNone(status["storage"])
         else:
             self.assertIsNone(status["storage"])
+
+    @override_settings(CACHE_ENABLED=False, USE_NEW_WORKER_TYPE=True)
```
dralley, line 104: I had no idea this was a thing, neat!

dralley (approved):

pulpcore/app/serializers/status.py
mdellweg: The thing to check, IMHO is whether the bindings can parse the result without error.

ekohl:
I should probably create an issue for this, but before I do: this changes the API. Does that disqualify it for a cherry pick to 3.14?

The reasoning is that with that we can avoid ever installing Redis in some scenarios.

daviddavis:
This is a good question and I'm a bit torn. I'll add it to Tuesday's pulpcore meeting agenda for us to discuss and report back.

dralley:
I'm also slightly conflicted, it's probably not something we would normally do - but on the other hand, it's not a particularly meaningful change and something Katello would probably want to add to their backport list.

But a lot of our users *are* using the Katello RPMs, most of them even, so we don't necessarily want a divergence between the Katello 3.14 and the upstream 3.14 either.

ekohl:
Thanks! I completely understand being torn because I'd be the same. Perhaps a good question is what @mdellweg said: will the bindings change? If it needs updated bindings, I can see the barrier being much higher. Let me know if I need to do anything.

PR #1490, opened 2021-07-13
Fix upgrade test

fao89:
https://pulp.plan.io/issues/9010
[noissue]

### DIFF
--- a/pulpcore/tests/upgrade/pre/test_publish.py
+++ b/pulpcore/tests/upgrade/pre/test_publish.py
@@ -9,7 +9,6 @@ from pulp_smash.pulp3.utils import (
     get_content,
     gen_distribution,
     get_versions,
-    modify_repo,
 )
 
 from pulpcore.tests.functional.api.using_plugin.constants import FILE_CONTENT_NAME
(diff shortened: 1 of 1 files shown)


pulpcore/tests/upgrade/pre/test_publish.py
```diff
@@ -78,10 +78,15 @@ def test_all(self):
...
+        repo_content = get_content(self.repo.to_dict())[FILE_CONTENT_NAME][:-1]
+        for file_content in repo_content:
+            repository_modify_data = RepositoryAddRemoveContent(
+                remove_content_units=[file_content["pulp_href"]]
+            )
+            modify_response = self.repo_api.modify(self.repo.pulp_href, repository_modify_data)
+            monitor_task(modify_response.task)
         version_hrefs = tuple(ver["pulp_href"] for ver in get_versions(self.repo.to_dict()))
-        non_latest = choice(version_hrefs[:-1])
+        non_latest = choice(version_hrefs[1:-1])
```
dralley, line 89: @fao89 So the fix here is: create multiple repository versions by repeatedly removing content (instead of one version by removing content only once) so that we have a "non-latest" version to pick, correct?

Is this particular change necessary?  I would hope the order the versions are returned in will stay consistent, so we only need to exclude one version off the end (or front) of the list.

fao89: we were deleting all the content, and we use choice, sometimes it picked a version without the content, yielding an empty manifest.
So I changed to not delete all content (ensuring we have at least one) and dropping version 0

dralley: Oh, version 0, that makes perfect sense.

dralley (approved):

dkliban (approved):

PR #1491, opened 2021-07-13
Fix on_demand downloads from file:// urls

gerrod3:
fixes: #9078

### DIFF
--- a/pulpcore/download/base.py
+++ b/pulpcore/download/base.py
@@ -76,6 +76,7 @@ class BaseDownloader:
         expected_digests=None,
         expected_size=None,
         semaphore=None,
+        headers_ready_callback=None,
         *args,
         **kwargs,
     ):
@@ -92,12 +93,17 @@ class BaseDownloader:
             expected_size (int): The number of bytes the download is expected to have.
             semaphore (asyncio.Semaphore): A semaphore the downloader must acquire before running.
                 Useful for limiting the number of outstanding downloaders in various ways.
+            headers_ready_callback (callable): An optional callback that accepts a single dictionary
+                as its argument. The callback will be called when the response headers are
+                available. The dictionary passed has the header names as the keys and header values
+                as its values. e.g. `{'Transfer-Encoding': 'chunked'}`. This can also be None.
         """
         self.url = url
         self._writer = custom_file_object
         self.path = None
         self.expected_digests = expected_digests
         self.expected_size = expected_size
+        self.headers_ready_callback = headers_ready_callback
         if semaphore:
             self.semaphore = semaphore
         else:
@@ -247,12 +253,16 @@ class BaseDownloader:
         Run the downloader.
 
         This is a coroutine that asyncio can schedule to complete downloading. Subclasses are
-        required to implement this method and do two things:
+        required to implement this method and do three things:
 
-        1. Pass all downloaded data to
+        1. Call :meth:`~pulpcore.plugin.download.BaseDownloader.headers_ready_callback` before any
+            data is delivered to :meth:`~pulpcore.plugin.download.BaseDownloader.handle_data` if
+            the method is defined.
+
+        2. Pass all downloaded data to
            :meth:`~pulpcore.plugin.download.BaseDownloader.handle_data`.
 
-        2. Call :meth:`~pulpcore.plugin.download.BaseDownloader.finalize` after all data has
+        3. Call :meth:`~pulpcore.plugin.download.BaseDownloader.finalize` after all data has
            been delivered to :meth:`~pulpcore.plugin.download.BaseDownloader.handle_data`.
 
         It is also expected that the subclass implementation return a
--- a/pulpcore/download/file.py
+++ b/pulpcore/download/file.py
@@ -2,7 +2,7 @@ import os
 
 from urllib.parse import urlparse
 
-import aiofiles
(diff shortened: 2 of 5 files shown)


pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -326,6 +329,20 @@ def test_content_served_on_demand(self):
...
         self.do_test_content_served()
 
+    def test_content_served_on_demand_filesystem(self):
+        """Assert on_demand content from filesystem can be downloaded."""
+        with TemporaryDirectory(dir="/tmp") as td:
+            # Create a filesystem remote
+            for file in ["1.iso", "2.iso", "3.iso", "PULP_MANIFEST"]:
+                path = os.path.join(td, file)
+                with open(path, "wb") as fd:
+                    fd.write(utils.http_get(urljoin(FILE_FIXTURE_URL, file)))
```
gerrod3, line 339: @dkliban I need a way to write these temporary files into the Pulp container for the CI. Any ideas?

dralley: @gerrod3 https://github.com/pulp/pulp_file/pull/528/files#diff-bca5f0b916aca91596114e579ea90e23aec2287619d80a5dbfbbca99697cd303R35-R52

dralley: Although since this would be the third use of it, maybe it needs to be a helper utility in pulp-smash

`def wget_download_fixture(url, path)`

dralley:
For discussion: https://pulp.plan.io/issues/9078#note-6

pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -326,6 +329,21 @@ def test_content_served_on_demand(self):
         self.setup_download_test("on_demand")
         self.do_test_content_served()
 
+    def test_content_served_on_demand_filesystem(self):
+        """Assert on_demand content from filesystem can be downloaded."""
+        if not shutil.which("wget"):
+            unittest.skip("Cannot run file:// tests without wget available.")
+        p_cli = cli.Client(self.cfg)
+        with TemporaryDirectory(dir="/tmp") as td:
```
dralley, line 337: Can we extract this out into a pulp-smash helper utility?

pulpcore/download/file.py
```diff
@@ -2,7 +2,7 @@
 
 from urllib.parse import urlparse
 
-import aiofiles
+import aiofiles, aiofiles.os
```
bmbouter, line 5: This should be left as-is since the dot-access is already available on `aiofiles` as is.

pulpcore/download/base.py
```diff
@@ -76,6 +76,7 @@ def __init__(
         expected_digests=None,
         expected_size=None,
         semaphore=None,
+        headers_ready_callback=None,
```
bmbouter, line 79: We want to fix this differently because we don't want to bring the notion of headers into the the base downloader. Instead we need to probably adjust how the content app handles the starting of the streaming response.

pulpcore/download/file.py
dralley: I don't think that is (necessarily) the case since the `os` module is not part of `__all__`

https://github.com/Tinche/aiofiles/blob/master/src/aiofiles/__init__.py

pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -326,6 +329,21 @@ def test_content_served_on_demand(self):
         self.setup_download_test("on_demand")
         self.do_test_content_served()
 
+    def test_content_served_on_demand_filesystem(self):
+        """Assert on_demand content from filesystem can be downloaded."""
```
dralley, line 333: Please do try to make a generic helper in `pulp_smash` if possible since we have a few different places that are using this approach now.

If it's not possible due to the specifics of this particular usage then that's OK.

pulpcore/download/file.py
bmbouter: If they aren't declaring it in their `_all_` then maybe we shouldn't use it? It's mute anyway though we need to do this PR differently.

dralley: I don't think that's the case, they're just taking a respectful approach to namespaces.  Every function in `aiofiles` maps to a non-async function of the same name so putting everything in `__all__` would cause issues if anyone was silly enough to `import *`

pulpcore/tests/functional/api/using_plugin/test_distributions.py
bmbouter, line 337: Yeah we need the test to assume the system under test is not localhost so we can't use this approach. We need to place the files on the remote system. There is a similar approach here https://github.com/pulp/pulp_file/pull/528

bmbouter (changes requested):
We need to find a way fix this without introducing the notion of an http-only concept to base.

Overall the issue is that the content app is expecting an http downloader here https://github.com/pulp/pulpcore/blob/master/pulpcore/content/handler.py#L869-L871

gerrod3:
Closing since this issue is probably pretty rare to run into and this isn't the best solution to fix it. See https://pulp.plan.io/issues/9078?pn=1#note-9.

PR #1494, opened 2021-07-15
Fixed non-gracefully shutting down workers

mdellweg:
When hitting ctrl-c twice, workers should also kill their supervised
tasks. And shut them down.

backports #8986
https://pulp.plan.io/issues/8986

fixes #9086

(cherry picked from commit 80d7d6d8e24c26d6eb0ff13ce29ed8c764cf2341)

### DIFF
--- a/CHANGES/9086.bugfix
+++ b/CHANGES/9086.bugfix
@@ -1,0 +1,2 @@
+Fixed signal handling to properly kill a task when double ctrl-c is used to shut down a worker fast.
+(backported from #8986)
--- a/pulpcore/tasking/pulpcore_worker.py
+++ b/pulpcore/tasking/pulpcore_worker.py
@@ -234,6 +234,12 @@ class NewPulpWorker:
 
 
 def child_signal_handler(sig, frame):
+    # Reset signal handlers to default
+    # If you kill the process a second time it's not graceful anymore.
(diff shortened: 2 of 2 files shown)


mdellweg:
@daviddavis i believe, this failure is triggered by the update of the ci container to python 3.8.

Do we have a plan to handle z-streams without that support?

daviddavis:
> Do we have a plan to handle z-streams without that support?

My understanding is that our policy is to update release branches with the latest changes from the plugin_template which should fix this.

mdellweg:
> > Do we have a plan to handle z-streams without that support?
> 
> My understanding is that our policy now is to update release branches with the latest changes from the plugin_template which should fix this.

Wouldn't that also bump the python version requirement on a release branch? I don't think, we want that, do we?

daviddavis:
> Wouldn't that also bump the python version requirement on a release branch? I don't think, we want that, do we?

It will bump the CI files but it shouldn't bump the Python version requirement. See my PR here:

https://github.com/pulp/pulpcore/pull/1497

The one thing is we're going to be testing release branches against Python 3.8 now. I could look at producing a Python 3.6 base image though for our release branches if we want to ensure that they work against Python 3.6.

dralley (approved):

PR #1495, opened 2021-07-15
Prevent access to unpublish content from content app

gerrod3:
fixes: #8870
https://pulp.plan.io/issues/8870

### DIFF
--- a/pulpcore/app/models/publication.py
+++ b/pulpcore/app/models/publication.py
@@ -391,6 +391,9 @@ class Distribution(MasterModel):
         repository_version (models.ForeignKey): RepositoryVersion to be served.
     """
 
+    # If distribution serves publications, set by subclasses for proper handling in content app
+    SERVE_FROM_PUBLICATION = False
+
     name = models.TextField(db_index=True, unique=True)
     base_path = models.TextField(unique=True)
 
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -557,7 +557,7 @@ class Handler:
                             request, StreamResponse(headers=headers), ca
                         )
 
-        if repo_version:
+        if repo_version and not publication and not distro.SERVE_FROM_PUBLICATION:
             if rel_path == "" or rel_path[-1] == "/":
                 index_path = "{}index.html".format(rel_path)
 
--- a/pulpcore/tests/functional/api/using_plugin/test_distributions.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_distributions.py
@@ -300,26 +300,12 @@ class ContentServePublicationDistributionTestCase(unittest.TestCase):
 
     def test_nonpublished_content_not_served(self):
         """Verify content that hasn't been published is not served."""
-        # Create a repository
-        self.repo = self.repo_api.create(gen_repo())
-        self.addCleanup(self.repo_api.delete, self.repo.pulp_href)
-
-        # Create a remote
-        self.remote = self.remote_api.create(gen_file_remote())
-        self.addCleanup(self.remote_api.delete, self.remote.pulp_href)
-
-        # Sync the repository.
-        repository_sync_data = RepositorySyncURL(remote=self.remote.pulp_href)
-        sync_response = self.repo_api.sync(self.repo.pulp_href, repository_sync_data)
-        monitor_task(sync_response.task)
-
-        # Create a distribution.
-        response = self.distributions_api.create(
-            {"name": "foo", "base_path": "bar/foo", "repository": self.repo.pulp_href}
-        )
-        distribution_href = monitor_task(response.task).created_resources[0]
-        self.distribution = self.distributions_api.read(distribution_href)
-        self.addCleanup(self.distributions_api.delete, self.distribution.pulp_href)
+        self.setup_download_test("immediate", publish=False)
+        files = ["", "1.iso", "2.iso", "3.iso"]
+        for file in files:
+            with self.assertRaises(HTTPError, msg=f"{file}") as cm:
+                download_content_unit(self.cfg, self.distribution.to_dict(), file)
+            self.assertEqual(cm.exception.response.status_code, 404, f"{file}")
 
     def test_content_served_on_demand(self):
         """Assert that on_demand content can be properly downloaded."""
@@ -407,7 +393,7 @@ class ContentServePublicationDistributionTestCase(unittest.TestCase):
             )
         self.assertEqual(cm.exception.response.status_code, 416)
 
-    def setup_download_test(self, policy, url=None):
+    def setup_download_test(self, policy, url=None, publish=True):
         # Create a repository
         self.repo = self.repo_api.create(gen_repo())
         self.addCleanup(self.repo_api.delete, self.repo.pulp_href)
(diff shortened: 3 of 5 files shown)


gerrod3:
Example PR using this change: https://github.com/pulp/pulp_file/pull/535. I have a feeling the test I wrote for this won't pass since it will not be using the pulp_file PR, but I think pulp_file CI runs the pulpcore tests too so it should pass there since I set it to require this PR.

pulpcore/app/models/publication.py
```diff
@@ -381,6 +381,12 @@ class Distribution(MasterModel):
         repository_version (models.ForeignKey): RepositoryVersion to be served.
     """
 
+    # PLUGIN TYPE CONSTANTS
+    PUBLICATION = "publication"
+    REPOSITORY_VERSION = "repository_version"
+    # Plugin type meant to be set by subclasses for proper handling in content app
+    plugin_type = None
```
dralley, line 388: Maybe we should call it "distribution type" instead of "plugin type"?  It's not strictly accurate because a plugin could contain both types of distributions even though that is not a common thing to need.  "distribution type" feels more specific.

But I guess it depends on how frequently we use that terminology already.  I'm definitely guilty of using that shorthand myself, and I don't really see it in the documentation, but if we do use it already in a few places (that I haven't found yet) it might not be worth it.

@daviddavis @ipanova What do you think?

Either way I would recommend doing the constants slightly differently.

```python
from types import simplenamespace

Types = SimpleNamespace(
    PUBLICATION="publication",
    REPOSITORY_VERSION="repository_version"
)

class RpmDistribution(Distribution):
    distribution_type = Distribution.Types.PUBLICATION
```

pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -300,26 +300,13 @@ def setUp(self):
...
-        monitor_task(sync_response.task)
-
-        # Create a distribution.
-        response = self.distributions_api.create(
-            {"name": "foo", "base_path": "bar/foo", "repository": self.repo.pulp_href}
-        )
-        distribution_href = monitor_task(response.task).created_resources[0]
-        self.distribution = self.distributions_api.read(distribution_href)
-        self.addCleanup(self.distributions_api.delete, self.distribution.pulp_href)
+        # This test was doing nothing before
```
dralley, line 303: Lol, you're right.

pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -407,7 +394,7 @@ def test_content_served_immediate_with_range_request_start_value_larger_than_con
             )
         self.assertEqual(cm.exception.response.status_code, 416)
 
-    def setup_download_test(self, policy, url=None):
+    def setup_download_test(self, policy, url=None, serve="publication"):
```
dralley, line 397: Probably just make this a boolean?

pulpcore/content/handler.py
```diff
@@ -468,7 +468,7 @@ def get_latest_publication_or_version_blocking():
                 nonlocal publication
 
                 # Search for publication serving the closest latest version
-                if not publication:
+                if not publication and distro.plugin_type != distro.REPOSITORY_VERSION:
```
dralley, line 471: Consider flipping this to

```python
if distro.plugin_type == distro.Types.PUBLICATION and not publication:
```

pulpcore/content/handler.py
```diff
@@ -557,7 +557,7 @@ def get_contentartifact_blocking():
                             request, StreamResponse(headers=headers), ca
                         )
 
-        if repo_version:
+        if repo_version and distro.plugin_type != distro.PUBLICATION:
```
dralley, line 560: Same here.

pulpcore/app/models/publication.py
ipanova: how about `needs_publication` or `publication_based` boolean flag?this way we'd avoid introducing more constants

```
class RpmDistribution(Distribution): 
    publication_based = True
```

dralley: That does seem sufficient

pulpcore/tests/functional/api/using_plugin/test_distributions.py
gerrod3: Ok, can you check the CI and help me figure out why the linter is failing?

dralley: That's... weird.  Maybe some kind of weird control character has snuck into the file somehow?

mdellweg:
Can the typo in the commit message be addressed too?
There is "unpublish" where it should read "unpublished".

you know: `git commit --amend`

pulpcore/app/models/publication.py
```diff
@@ -381,6 +381,9 @@ class Distribution(MasterModel):
         repository_version (models.ForeignKey): RepositoryVersion to be served.
     """
 
+    # If distribution serves publications, set by subclasses for proper handling in content app
+    publication_based = False
```
dralley, line 385: We will want to update the plugin template with this change, and probably the plugin writer guide too.

dralley (approved):

pulpcore/app/models/publication.py
gerrod3: I'm thinking of renaming this because of the behavior expected by pulp_python. pulp_python allows for content to be served from a publication or live from a repository by using the fact that you can currently access content if you know where it is stored in the content app. I'm thinking maybe `protect_unpublished` or `only_serve_published`.

dralley: Between those two, the latter makes more sense.  But let's lock down the name soon because the RPM plugin will release tomorrow and the file plugin will need a new release

ipanova: with having little context of the python plugin needs these names do not make much sense to me.

gerrod3: Basically pulp_python uses this bug of being able to access unpublished content to allow users to serve content whether they have created a publication or not. So in a sense pulp_python fits under this term `publication_based`, but doesn't want to use this fix since it'll break its current features. This fix is suppose to stop access to unpublished content when using the auto-distribute feature, so renaming the variable will better suite its purpose.

ipanova: I have mixed feelings on this approach, On one hand we claim that this is a bug and we're trying to fix this but still want to enable certain plugin/s taking advantage of this gap. How plugins that have distribution only should read/understand this? `protect_unpublished` or `only_serve_published`.
How about `serve_from_publication`?
re: python plugin. I am confused. Python plugin has publication, but it seems to be a nominal one because content can be consumed regardless of it's presence? What is the purpose of having it then?

ipanova: before approving this PR, I'd like to agree in the variable naming :+1:

ipanova:
still waiting on automation hub to provide feedback before we take off the tech-preview.

pulpcore/app/models/publication.py
ipanova: as per discussion with @mdellweg we should probably have it as constant in caps

mdellweg: I want to add, that this is supposed to be a Class constant. So we should make it all caps (however we name it).

dralley: +1 to @ipanova's suggestion of `serve_from_publication`

And +1 to class constant.

>re: python plugin. I am confused. Python plugin has publication, but it seems to be a nominal one because content can be consumed regardless of it's presence? What is the purpose of having it then?

IIRC it's because the metadata is sufficiently cheap to create on-the-fly due to the way it's accessed, correct me if I'm wrong @gerrod3.

mdellweg:
```suggestion
    PUBLICATION_BASED = False
```
```suggestion
    SERVE_FROM_PUBLICATION = False
```

In the Exporter refactoring PR we came across the question, how we could derive this value from a repository version to determine whether it's OK to export a repo_version directly or you'd need to go through a publication.
Would it make sense to move this variable to the repository?

mdellweg (approved):

ipanova:
I tested following these steps and it worked well for me https://pulp.plan.io/issues/8870#note-5

ipanova (approved):

pulpcore/app/models/publication.py
ipanova: @mdellweg it should be possible, I think `repo.SERVED_FROM_PUBLICATION` would do it.

pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -300,26 +300,12 @@ def setUp(self):
...
-        sync_response = self.repo_api.sync(self.repo.pulp_href, repository_sync_data)
-        monitor_task(sync_response.task)
-
-        # Create a distribution.
-        response = self.distributions_api.create(
-            {"name": "foo", "base_path": "bar/foo", "repository": self.repo.pulp_href}
-        )
-        distribution_href = monitor_task(response.task).created_resources[0]
-        self.distribution = self.distributions_api.read(distribution_href)
-        self.addCleanup(self.distributions_api.delete, self.distribution.pulp_href)
```
dralley, line 322: Whoops, I think this regression was my fault

dralley (approved):

PR #1437, opened 2021-06-23
Alternate Content Source Models

pavelpicka:
basic models, viewsets and some logic for Alternate Content Sources

closes: #8606
closes: #8607
https://pulp.plan.io/issues/8606
https://pulp.plan.io/issues/8607

PR #1496, opened 2021-07-16
Acs artifact stage

pavelpicka:
New optional stage for ACS. 

Be aware that this PR is on top commit used in https://github.com/pulp/pulpcore/pull/1437. Before merge please rebase and drop commit with ACS models.

### DIFF
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -4,10 +4,16 @@ from gettext import gettext as _
 import logging
 
 from asgiref.sync import sync_to_async
-from django.db.models import Prefetch, prefetch_related_objects
+from django.db.models import Prefetch, prefetch_related_objects, Q
 
 from pulpcore.plugin.exceptions import UnsupportedDigestValidationError
-from pulpcore.plugin.models import Artifact, ContentArtifact, ProgressReport, RemoteArtifact
+from pulpcore.plugin.models import (
+    AlternateContentSource,
+    Artifact,
+    ContentArtifact,
+    ProgressReport,
+    RemoteArtifact,
+)
 from pulpcore.plugin.sync import sync_to_async_iterable
 
 from .api import Stage
@@ -398,3 +404,60 @@ class RemoteArtifactSaver(Stage):
         )
         ra.validate_checksums()
         return ra
+
+
+class ACSArtifactHandler(Stage):
+    """
+    API stage to download :class:`~pulpcore.plugin.models.Artifact` files from Alternate
+    Content Source if available.
+    """
+
+    async def run(self):
+        async for batch in self.batches():
+            acs_exists = sync_to_async(AlternateContentSource.objects.exists)()
+            if acs_exists:
+                # Gather batch d_artifact checksums
+                batch_checksums = defaultdict(list)
+                for d_content in batch:
+                    for d_artifact in d_content.d_artifacts:
+                        for cks_type in d_artifact.artifact.COMMON_DIGEST_FIELDS:
+                            if getattr(d_artifact.artifact, cks_type):
+                                batch_checksums[cks_type].append(
+                                    getattr(d_artifact.artifact, cks_type)
+                                )
+
+                batch_query = Q()
+                for checksum_type in batch_checksums.keys():
+                    batch_query.add(
+                        Q(**{f"{checksum_type}__in": batch_checksums[checksum_type]}), Q.OR
+                    )
+
+                existing_ras = await sync_to_async(list)(
+                    (
+                        RemoteArtifact.objects.acs()
+                        .filter(batch_query)
+                        .only("url", "remote")
+                        .select_related("remote")
+                    )
+                )
+                existing_ras_dict = dict()
+                for ra in existing_ras:
+                    for c_type in Artifact.COMMON_DIGEST_FIELDS:
+                        checksum = await sync_to_async(getattr)(ra, c_type)
+                        if checksum:
+                            existing_ras_dict[checksum] = {
+                                "remote": ra.remote,
+                                "url": ra.url,
+                            }
+
+                for d_content in batch:
+                    for d_artifact in d_content.d_artifacts:
+                        for checksum_type in Artifact.COMMON_DIGEST_FIELDS:
+                            if getattr(d_artifact.artifact, checksum_type):
+                                checksum = getattr(d_artifact.artifact, checksum_type)
+                                if checksum in existing_ras_dict:
+                                    d_artifact.url = existing_ras_dict[checksum]["url"]
+                                    d_artifact.remote = existing_ras_dict[checksum]["remote"]
+
+            for d_content in batch:
+                await self.put(d_content)
--- a/pulpcore/plugin/stages/declarative_version.py
+++ b/pulpcore/plugin/stages/declarative_version.py
@@ -3,6 +3,7 @@ import tempfile
 
 from .api import create_pipeline, EndStage
 from .artifact_stages import (
+    ACSArtifactHandler,
     ArtifactDownloader,
     ArtifactSaver,
     QueryExistingArtifacts,
@@ -17,7 +18,7 @@ from .content_stages import (
 
 
 class DeclarativeVersion:
-    def __init__(self, first_stage, repository, mirror=False):
+    def __init__(self, first_stage, repository, mirror=False, acs=False):
         """
         A pipeline that creates a new :class:`~pulpcore.plugin.models.RepositoryVersion` from a
         stream of :class:`~pulpcore.plugin.stages.DeclarativeContent` objects.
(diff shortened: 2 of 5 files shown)


pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -398,3 +406,55 @@ def _create_remote_artifact(d_artifact, content_artifact):
...
+    A Stages API stage to download :class:`~pulpcore.plugin.models.Artifact` files from Alternative
+    Content Source if available.
+    """
+
+    async def run(self):
+        async for batch in self.batches():
+            if AlternateContentSource.objects.exists():
+                # get first checksum type from d_artifacts to use for search ACS
+                # or fallback for sha256
+                checksum_type = "sha256"
```
daviddavis, line 422: This assumes that all RemoteArtifacts in a batch will have the same checksum type?

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -398,3 +406,55 @@ def _create_remote_artifact(d_artifact, content_artifact):
...
+                    getattr(d_artifact.artifact, checksum_type)
+                    for d_content in batch
+                    for d_artifact in d_content.d_artifacts
+                ]
+                acs_repos = Repository.objects.filter(
+                    alternatecontentsourcepath__in=AlternateContentSourcePath.objects.all()
+                )
+                existing_ras = RemoteArtifact.objects.filter(
+                    content_artifact__content__repositories__in=acs_repos,
+                    **{checksum_type + "__in": batch_remote_artifacts_checksums},
```
daviddavis, line 446:
```suggestion
                    **{f"{checksum_type}__in": batch_remote_artifacts_checksums},
```

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -398,3 +406,55 @@ def _create_remote_artifact(d_artifact, content_artifact):
...
+                ).select_related("remote")
```
daviddavis, line 447: I think you could probably simplify this using [the `acs()` queryset method](https://github.com/pulp/pulpcore/pull/1605/files#diff-81f6a78175bb93934b6beff952646d3ca1ef3731f1ff14492d4ec77bfc3fdf82R574-R576):

```suggestion
                existing_ras = RemoteArtifact.objects.acs().filter(
                    **{f"{checksum_type}__in": batch_remote_artifacts_checksums}
                )
```

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -398,3 +406,55 @@ def _create_remote_artifact(d_artifact, content_artifact):
...
+
+                for d_content in batch:
+                    for d_artifact in d_content.d_artifacts:
+                        if existing_ras.filter(
+                            **{checksum_type: getattr(d_artifact.artifact, checksum_type)}
+                        ).exists():
+                            alternate_ra = existing_ras.filter(
+                                **{checksum_type: getattr(d_artifact.artifact, checksum_type)}
+                            ).first()
```
daviddavis, line 456: These queries are going to execute for each `d_artifact`. I think you could eliminate them by storing a dictionary of remote artifacts and looking them up by checksum.

pulpcore/plugin/stages/artifact_stages.py
pavelpicka: It does.

bmbouter: I only spent a few minutes just now looking at this, but I'm not sure this is a valid assumption.

pavelpicka: Updated to take existing checksum from each d_artifact and look for ACS by it.

pulpcore/plugin/stages/artifact_stages.py
pavelpicka: to take URL it works well, but still we need to 'get' remote (not only the remote_id).

daviddavis: Right, I was imagining that the dict would relate checksum type + checksum digest to remote.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -398,3 +404,56 @@ def _create_remote_artifact(d_artifact, content_artifact):
...
+        async for batch in self.batches():
+            if AlternateContentSource.objects.exists():
+                # Gather batch d_artifact checksums
+                batch_checksums = {}
+                for d_content in batch:
+                    for d_artifact in d_content.d_artifacts:
+                        for cks_type in d_artifact.artifact.COMMON_DIGEST_FIELDS:
+                            if getattr(d_artifact.artifact, cks_type):
+                                if cks_type not in batch_checksums.keys():
+                                    batch_checksums[cks_type] = list()
```
daviddavis, line 425: What if you made `batch_checksums` a `defaultdict(list)`?

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -398,3 +404,56 @@ def _create_remote_artifact(d_artifact, content_artifact):
...
+                            if (
+                                getattr(d_artifact.artifact, checksum_type) != ""
+                                and getattr(d_artifact.artifact, checksum_type) is not None
+                            ):
+                                search = [
+                                    ra
+                                    for ra in existing_ras_list
+                                    if ra[checksum_type]
+                                    == getattr(d_artifact.artifact, checksum_type)
+                                ]
```
daviddavis, line 451: I think this list comprehension is inefficient since you only need the first match. If you used a dict data structure, you wouldn't need this list comprehension at all.

CHANGES/8748.feature
```diff
@@ -0,0 +1 @@
+Added optional stage for Alternate Content Source support.
```
bmbouter, line 1: This is a plugin API change, so this needs to be `CHANGES/plugin_api/8748.feature`.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -398,3 +404,59 @@ def _create_remote_artifact(d_artifact, content_artifact):
...
+                                "url": ra.url,
+                            }
+
+                for d_content in batch:
+                    for d_artifact in d_content.d_artifacts:
+                        for checksum_type in Artifact.COMMON_DIGEST_FIELDS:
+                            if (
+                                getattr(d_artifact.artifact, checksum_type) != ""
+                                and getattr(d_artifact.artifact, checksum_type) is not None
+                            ):
```
daviddavis, line 455:
```suggestion
                            if getattr(d_artifact.artifact, checksum_type, False):
```

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -398,3 +404,59 @@ def _create_remote_artifact(d_artifact, content_artifact):
...
         return ra
+
+
+class ACSArtifactHandler(Stage):
+    """
+    A Stages API stage to download :class:`~pulpcore.plugin.models.Artifact` files from Alternative
+    Content Source if available.
+    """
+
+    async def run(self):
```
bmbouter, line 415: In reading over this code, it reads well enough, but I'd really like to run it. I'd be open to merging this and then having the test of pulp_file allow us to step through it more effectively.

bmbouter: @daviddavis wdyt ^?

daviddavis: That works for me assuming the 2 open items of feedback are addressed.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -398,3 +404,59 @@ def _create_remote_artifact(d_artifact, content_artifact):
...
+                for checksum_type in batch_checksums.keys():
+                    batch_query.add(
+                        Q(**{f"{checksum_type}__in": batch_checksums[checksum_type]}), Q.OR
+                    )
+
+                existing_ras = (
+                    RemoteArtifact.objects.acs()
+                    .filter(batch_query)
+                    .only("url", "remote")
+                    .prefetch_related("remote")
```
dralley, line 438: Use `select_related()` if you are able to do so

daviddavis (approved):

bmbouter (approved):

daviddavis (changes requested):
Can you squash the commits and address @dralley's comment? Then I think we can merge.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -398,3 +404,60 @@ def _create_remote_artifact(d_artifact, content_artifact):
         )
         ra.validate_checksums()
         return ra
+
+
+class ACSArtifactHandler(Stage):
+    """
+    A Stages API stage to download :class:`~pulpcore.plugin.models.Artifact` files from Alternative
```
ipanova, line 411: s/A Stages API stage to download/ API Stage
s/Alternative/Alternate/

pulpcore/plugin/stages/declarative_version.py
```diff
@@ -101,11 +102,13 @@ async def run(self):
                 :class:`~pulpcore.plugin.stages.DeclarativeVersion stream`, and does not remove any
                 pre-existing units in the :class:`~pulpcore.plugin.models.RepositoryVersion`.
                 'False' is the default.
+            acs (bool): 'True' adding stage to look for a content in Alternate Content Sources.
```
ipanova, line 105: "When set to 'True' a new stage is added to look for..."

daviddavis (approved):

PR #1498, opened 2021-07-16
Fix disk usage during the synchronization

lubosmj:


pulpcore/plugin/stages/api.py
```diff
@@ -81,7 +81,7 @@ async def run(self):
             log.debug("%(name)s - next: %(content)s.", {"name": self, "content": content})
             yield content
 
-    async def batches(self, minsize=500):
+    async def batches(self, minsize=250):
```
dralley, line 84: This may have some performance implications.

pulpcore/app/models/storage.py
```diff
@@ -11,9 +11,9 @@ class FileSystem(FileSystemStorage):
     """
     Django's FileSystemStorage with modified _save() and get_available_name behaviors
 
-    The _save() will check if the file is saved in MEDIA_ROOT first. If it is, a move is used. This
+    The _save() will check if the file is saved in DEPLOY_ROOT first. If it is, a move is used. This
     will move all files created by the Downloaders and uploaded files from the user. If it is saved
-    in-memory or outside of MEDIA_ROOT, the data is written in chunks to the new location.
+    in-memory or outside of MEDIA_ROOT, the data is written/copied in chunks to the new location.
```
dralley, line 16: Should this second instance of `MEDIA_ROOT` be changed also?

pulpcore/plugin/stages/api.py
lubosmj: In the commit message, I am trying to say what size of the memory needs to be allocated for one task during the synchronization in /var/lib/pulp. Regarding the performance implications, I am again referring to @fao89's comments https://github.com/pulp/pulpcore/pull/440/files where he suggested re-evaluating these defaults.

dralley: Oh I'm definitely not suggesting we don't do it, I'm just pointing it out as something we should maybe look into in the future.

dralley: @lubosmj Why not override it for the particular stages that need a smaller batch size? https://github.com/pulp/pulpcore/blob/master/pulpcore/plugin/stages/artifact_stages.py#L236

lubosmj: @dralley alright, you persuaded me.

dralley: Cool -- I believe the commits can also be squashed in that case, since it is a more self-contained change.

dralley:
We fixed the CI so you should rebase when you see this on Monday

dralley (approved):

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -233,7 +233,10 @@ async def run(self):
         Returns:
             The coroutine for this stage.
         """
-        async for batch in self.batches():
+        # the parameter 'minsize' was decreased in order to make this method save downloaded
+        # artifacts more frequently; decreasing the value lowers disk usage when artifacts
+        # are created from downloaded files which reside in the Pulp's working directory
+        async for batch in self.batches(minsize=250):
```
bmbouter, line 239: I think there is sufficient evidence to change this number from the default. I recommend removing this part of the diff.

pulpcore/app/models/storage.py
```diff
@@ -63,7 +63,7 @@ def _save(self, name, content, max_length=None):
 
         try:
             if hasattr(content, "temporary_file_path") and content.temporary_file_path().startswith(
-                settings.MEDIA_ROOT
+                str(settings.DEPLOY_ROOT)
```
bmbouter, line 66: This part looks good though.

Also what happens if you don't have the `str()` in it?

pulpcore/plugin/stages/artifact_stages.py
dralley: I recommended this as a reminder of "why use something special here" when we go working in this code.  It could be made a little more concise though.

bmbouter: I mean specifying minsize at all. Do we want it saving in batches of 250 versus the default of 500? I didn't think we had a good reason to change it. Maybe we do?

pulpcore/app/models/storage.py
lubosmj: An error is raised since the method expects only strings.

pulpcore/plugin/stages/artifact_stages.py
lubosmj: The rationale behind this change is here: https://github.com/pulp/pulpcore/pull/1465#discussion_r664795959.

pulpcore/app/models/storage.py
dralley: From `settings.py`

```
DEPLOY_ROOT = Path("/var/lib/pulp")
MEDIA_ROOT = str(DEPLOY_ROOT / "media")  # Django 3.1 adds support for pathlib.Path
```

This is weird to me - perhaps (going by the comment) we can address the inconsistency post upgrade to Django 3?

(But it should remain separate from this PR because we need to backport it).

pulpcore/plugin/stages/artifact_stages.py
dralley: @bmbouter The rationale @lubosmj gave makes sense to me.  

Basically because `minsize` is the minimum batch size, that means that 500 artifacts must build up in the `ArtifactSaver` input queue (read: all artifacts downloaded for 500 content units) before `ArtifactSaver` will start processing them (which allows deleting the temporary files).

I actually think that we need to file a refactor issue for this problem still because simply lowering the cap to 250 content units is only an indirect means of reducing the disk consumption - we could still have trouble with 250 content that have very large artifacts.  But that would be a much larger task.

bmbouter: I thought this over and now I think lowering the batch size will likely improve our performance. Even though it has a negative impact on the number of round trips to the DB for the bulk_save calls, the amount of time lost there would likely be significantly less than the time lost by waiting to start moving large files into place and trying to move a huge number of files at once. So maybe even lower it further to like 100? I'm also ok with 250.

Thanks for talking this over.

bmbouter (approved):
Thanks @lubosmj !

pulpcore/plugin/stages/artifact_stages.py
dralley: I'd also be fine with any number between 100 and 250.  200 might be a good choice.

I'm not sure about the performance benefits because we don't do any asynchronous file IO, only blocking, so we're waiting either way - but I also agree that it won't hurt performance significantly.

dkliban (approved):

pulpcore/plugin/stages/artifact_stages.py
bmbouter: Yeah actually that matters a lot on if this will bring a performance benefit or not. I was misremembering thinking that we were using aiofiles here but we're not. iirc, aiofiles uses a threadpool underneath to make the file saving async but in this case it's already saved in /var/lib/pulp/ so even if it's synchronous there won't be any actual bit moving+saving so the filesystem operation will be very fast.

So if all ^ is right, this means that lowering the batch size will negatively affect performance by causing additional DB RTTs. After thinking about this more I don't think this is even having an effect on disk size because the files are both on the same volume and they aren't actually being moved so regardless of the move the same number of inodes will be used.

And if ^ is right, I think we should leave the batching alone.

bmbouter (changes requested):
let's figure out if the batching change is helpful or not. I don't think it is.

pulpcore/plugin/stages/artifact_stages.py
dralley: I don't think we *can* leave batching alone, because of the disk space issue, and in the event of an on-demand sync I don't think it would be a significant contributor to the overall runtime of the sync task.

And IIRC the `ArtifactSaver` stage should be a no-op in the `on_demand` case so no impact there.

bmbouter: I'm confused. The files pre and post save are on the save volume. How can we be "increasing disk size during sync" by moving files less often when they are on the same filesystem?

dralley: Ok, I think I see what you mean now.  Needed to re-read the PR.

lubosmj: Yes. Since we are not copying files anymore, but rather moving, the behaviour I described is not valid. Files were copied before the change. Therefore, I can get rid of `minsize=250`. Sorry for the inconvenience created.

bmbouter: It's no inconvenience. Things are complicated, and we gotta talk them out.

bmbouter (approved):
Thank you!

PR #1503, opened 2021-07-16
Change content app's working directory dynamically

lubosmj:
As of this commit, content app is no longer storing temporary files in the /var/run/ directory. The temporary files were created during on-demand downloading and were not removed until, e.g., restarting pulp services.

closes #9000

### DIFF
--- a/CHANGES/9000.bugfix
+++ b/CHANGES/9000.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug where on-demand downloads would fill up ``/var/run/`` by not deleting downloaded files.
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -720,6 +720,12 @@ class Handler:
                         **download_result.artifact_attributes, file=download_result.path
                     )
                     artifact.save()
+                else:
+                    # The file needs to be unlinked because it was not used to create an artifact.
+                    # The artifact must have already been saved while servicing another request for
+                    # the same artifact.
+                    os.unlink(download_result.path)
+
             update_content_artifact = True
             if content_artifact._state.adding:
                 # This is the first time pull-through content was requested.
--- a/pulpcore/content/__init__.py
+++ b/pulpcore/content/__init__.py
@@ -55,6 +55,8 @@ async def _heartbeat():
 
 
 async def server(*args, **kwargs):
+    os.chdir(settings.WORKING_DIRECTORY)
+
     asyncio.ensure_future(_heartbeat())
     for pulp_plugin in pulp_plugin_configs():
         if pulp_plugin.name != "pulpcore.app":
--- a/pulpcore/tests/unit/content/test_handler.py
+++ b/pulpcore/tests/unit/content/test_handler.py
@@ -1,6 +1,7 @@
+import tempfile
+
 from unittest.mock import Mock
 
-from django.core.files.uploadedfile import SimpleUploadedFile
 from django.test import TestCase
 
 from pulpcore.content import Handler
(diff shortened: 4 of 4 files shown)


pulpcore/content/handler.py
```diff
@@ -753,6 +753,9 @@ def _save_artifact(self, download_result, remote_artifact):
             if update_content_artifact:
                 content_artifact.artifact = artifact
                 content_artifact.save()
+
+                os.unlink(download_result.path)
```
lubosmj, line 757: Considering a repository that was synced with the on-demand policy, once content app receives N requests targeting the same content unit simultaneously, it tries to download the unit N times. The first downloaded file is associated with an Artifact. The rest of the downloaded files are not cleared up and reside on disk after exiting the method `_save_artifact()`.

I believe calling `os.unlink()` is not the best idea, but I was not able to come up with something better. Maybe it would be better to check whether `download_result.path` starts with `settings.WORKING_DIRECTORY` and then call `os.unlink()` (just to ma ke sure that **we** downloaded a file and **we** will take care of it)? I have no idea how is the method `_save_artifact()` used in plugins (if it is even used) and how can this change affect their processes. Since `settings.WORKING_DIRECTORY` is a working directory for content app, the aforesaid check might be reasonable.

gerrod3: I'm not sure if this is the right place to put this `os.unlick()` call. I think the call should be after line 715 because if that `artifact.touch()` call succeeds then it means that the artifact is already present on the system and thus the temporary file from the download will not be cleaned up from the `artifact.save()` code.

daviddavis: > `os.unlick()`

:astonished:

lubosmj:
I moved the `os.unlink()` call, as proposed by @gerrod3. Furthermore, I added an if-statement that decides whether a temporary file will be removed or not, based on its path.

pulpcore/content/handler.py
```diff
@@ -720,6 +720,12 @@ def _save_artifact(self, download_result, remote_artifact):
                         **download_result.artifact_attributes, file=download_result.path
                     )
                     artifact.save()
+                else:
+                    # The file needs to be unlinked because it was not used to create an artifact.
+                    # The artifact must have already been saved while servicing another request for
+                    # the same artifact.
+                    os.unlink(download_result.path)
```
dralley, line 727: I think this is a good mechanism to have regardless, but should we do anything to prevent the same artifact from being downloaded multiple times in parallel to begin with? @bmbouter @gerrod3

bmbouter: Personally I don't think we should be manipulating the downloaded files like this. It's hard to fully know the side-effects I think. Maybe I should look at the code more to understand this line's value.

Regarding preventing duplicate downloading, it's tough because across multiple processes that involves DB coordination which is very likely not worth it. Even for 1 process though there are other things for example, it could be significant for the headers_ready callback firing on both downloads but if we deduplicate them and the event already occured the second one won't receive it. Overall I put these kinds of optimizations in the tricky and not worth it category. That's my take, it could be wrong.

bmbouter: For example what if it wasn't used to create an Artifact, but consumed from some other 3rd party process and this code is being called from a newly registered handler that is just calling into this method. It's a contrived example I know, but my general point is, I don't think we really know unlinking is right in 100% of cases.

dkliban: @bmbouter If the unlinking of the file does not occur here it is going to remain in /var/lib/pulp/tmp. This line of code is only executed if the same file is requested at the same time and only one of the downloaded files is actually used to create the artifact. 

The other option is to have a dedicated directory under /var/lib/pulp/tmp for each instance of the content app and then coordinate cleanup of those directories.

bmbouter: Thanks for explaining this. This makes sense to me now. +1 to this.

bmbouter (approved):
Thanks!

dralley:
@bmbouter @dkliban Will the directory-changing require any work on the SELinux policy side?

bmbouter:
> @bmbouter @dkliban Will the directory-changing require any work on the SELinux policy side?

I don't know, but probably? This is why we need to have SELinux enforcing added to the CI and our dev environments.

CHANGES/9000.bugfix
```diff
@@ -0,0 +1 @@
+Updated the content app's working directory to ``WORKING_DIRECTORY`` specified in ``settings.py``.
```
ipanova, line 1: since we're exposing this to the users maybe it's worth explaining why this change was made? or how this change fixed the underlying issue?

dralley: +1.  Maybe "Fixed a bug where on-demand downloads would fill up /var/run/ by not deleting downloaded files"

lubosmj: @dralley this one is tricky. This PR does not primarily focus on deleting downloaded files (compared to https://github.com/pulp/pulpcore/commit/fbe2d7f9c85008e8223fec8fcfdb87e665e06881). But, in addition to changing the content app working directory, we delete only redundant files that will be never associated with artifacts. I am fine with your wording, though.

dralley: I think downloaded files that were simply not made into artifacts still counts

dralley (approved):
LGTM once the changelog wording is adjusted.

dkliban (approved):

PR #1505, opened 2021-07-19
Prevent tasks being assigned to missing workers

mdellweg:
This is a first attempt to be more thought full around missing / dirty
workers when assiging new tasks.

TODO:
- ~Pin down the real issue~
- ~Define what the solution really is~
- ~change this commit from "re" to "fixes"~

re #8779

### DIFF
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -47,15 +47,19 @@ def _acquire_worker(resources):
     """
     # Find a worker who already has one of the reservations, it is safe to send this work to them
     try:
-        return (
+        worker = (
             Worker.objects.select_for_update()
             .filter(pk__in=Worker.objects.filter(reservations__resource__in=resources))
             .get()
         )
+        if worker.online:
+            return worker
     except Worker.MultipleObjectsReturned:
         raise Worker.DoesNotExist
     except Worker.DoesNotExist:
         pass
+    else:  # no Exception raised -> worker is offline; wait for it to be cleaned up
+        raise Worker.DoesNotExist
 
     # Otherwise, return any available worker at random
     workers_qs = Worker.objects.online_workers().exclude(
@@ -146,6 +150,8 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
                             worker=worker, resource=resource
                         )
                     TaskReservedResource.objects.create(resource=reservation, task=task_status)
+                worker.cleaned_up = False
+                worker.save(update_fields=["cleaned_up"])
         except (Worker.DoesNotExist, IntegrityError):
             # if worker is ready, or we have a worker but we can't create the reservations, wait
             time.sleep(0.25)
--- a/pulpcore/tasking/worker.py
+++ b/pulpcore/tasking/worker.py
@@ -22,6 +22,7 @@ from pulpcore.app import settings  # noqa: E402: module level not at top of file
 from pulpcore.app.settings import WORKER_TTL  # noqa: E402: module level not at top of file
 
 from pulpcore.app.models import Task  # noqa: E402: module level not at top of file
+from pulpcore.constants import TASK_STATES  # noqa: E402: module level not at top of file
 
 from pulpcore.tasking.constants import (  # noqa: E402: module level not at top of file
     TASKING_CONSTANTS,
@@ -105,6 +106,8 @@ class PulpWorker(Worker):
         except Task.DoesNotExist:
             pass
         else:
+            if task.state != TASK_STATES.WAITING:
+                return
             task.set_running()
             user = get_users_with_perms(task).first()
             _set_current_user(user)
--- a/CHANGES/8779.bugfix
+++ b/CHANGES/8779.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug, where new tasks were assigned to dead workers.


mdellweg:
To reproduce:
`pulp debug tasking-benchmark -c 400 -N 1 -K 1 -s 1`
`sudo systemctl restart pulpcore-worker@*`

bmbouter:
What I've been testing is:

Before starting go ahead and flush everything in redis:

```
$ redis-cli
127.0.0.1:6379> FLUSHALL
OK
127.0.0.1:6379> 
```

Monitor the state of things with these commands:

```
watch -n 1 "pulp status | jq '. | {online_workers}'"
watch -n 1 'rq info'
```

While using the old-style tasking system, trigger a lot of tasks with `http POST http://localhost/pulp/api/v3/herminig/tasking_benchmark/ count=400 resources_N=2 resources_K=1 sleep_secs=2`

bmbouter:
When I tested this patch it seemed to fully resolve the user-facing issue, leaving only a vestige in the RQ task. The task groups shows 29 tasks cancelled and none at waiting:

```
{
    "pulp_href": "/pulp/api/v3/task-groups/a8d4eb22-5dc7-4478-9749-c8bb3946f74c/",
    "description": "Tasking system benchmark tasks",
    "all_tasks_dispatched": true,
    "waiting": 0,
    "skipped": 0,
    "running": 0,
    "completed": 371,
    "canceled": 29,
    "failed": 0,
    "canceling": 0,
    "group_progress_reports": [],
    "tasks": [
....
```

The `rq info` shows:

```
52284@pulp3-source-fedora34.fedora.example.com | 0
51679@pulp3-source-fedora34.fedora.example.com | 0
51678@pulp3-source-fedora34.fedora.example.com |███████████ 29
resource-manager | 0
4 queues, 29 jobs total

52284@pulp3-source-fedora34.fedora.example.com (pulp3-source-fedora34.fedora.example.com 127.0.0.1:54244 52284): idle 52284@pulp3-source-fedora34.fedora.example.com
51679@pulp3-source-fedora34.fedora.example.com (pulp3-source-fedora34.fedora.example.com 127.0.0.1:52994 51679): idle 51679@pulp3-source-fedora34.fedora.example.com
resource-manager (pulp3-source-fedora34.fedora.example.com 127.0.0.1:52978 51680): idle resource-manager
3 workers, 4 queues

Updated: 2021-07-19 20:50:37.952676
```

So all is well except that those RQ jobs are still in RQ and I expected them to be cleaned up. Then when I look closer at [the code I remember should be cleaning it up](https://github.com/pulp/pulpcore/commit/ac83f33c1db55846da73d8a1ae34c3721aec914f) I realize it's checking for tasks that are missing from RQ and not the other way around. So we don't have code that would clean this up.

Maybe that's OK?? because:
* a new worker won't get the same queue name anyway (pid difference)
* Even if it did I think it would error because there is no psql Task to match for it to transition the state for

bmbouter:
What I don't understand is why even though Pulp is calling `cancel()` on the correct jobs, it's not being cancelled [here](https://github.com/pulp/pulpcore/blob/master/pulpcore/tasking/util.py#L71)

mdellweg:
> So all is well except that those RQ jobs are still in RQ and I expected them to be cleaned up. Then when I look closer at [the code I remember should be cleaning it up](https://github.com/pulp/pulpcore/commit/ac83f33c1db55846da73d8a1ae34c3721aec914f) I realize it's checking for tasks that are missing from RQ and not the other way around. So we don't have code that would clean this up.
> 
> Maybe that's OK?? because:
> 
>     * a new worker won't get the same queue name anyway (pid difference)
> 
>     * Even if it did I think it would error because there is no psql Task to match for it to transition the state for

That is a good observation. But i would even go as far as saying: "This is a different issue."
This PR is addressing the "Tasks are assigned to long dead workers." issue.
Probably the "RQ isn't properly cleaned up in all cases is something that was always there, but we never looked at it. And therefore we might get away with just not caring about it.

dralley:
> What I don't understand is why even though Pulp is calling `cancel()` on the correct jobs, it's not being cancelled [here](https://github.com/pulp/pulpcore/blob/master/pulpcore/tasking/util.py#L71)

Note that, (I think *unlike* Celery), RQ will cancel a task if and only if it's not running yet.  If the task has left one of the queues (or hasn't arrived yet) it will not be blocked once it does arrive.  There's no blacklist of task IDs.

So maybe what is happening is that there is a race where the task to assign the "actual" task to a worker has started running on the resource manager, it "cancels" both tasks (except the task doesn't exist on either queue, it left the first queue and has never joined the second), and then the task completes (or effectively completes) prior to being hard-killed.

dralley:
Sorry, misclicked.

bmbouter:
> That is a good observation. But i would even go as far as saying: "This is a different issue."
> This PR is addressing the "Tasks are assigned to long dead workers." issue.
> Probably the "RQ isn't properly cleaned up in all cases is something that was always there, but we never looked at it. And therefore we might get away with just not caring about it.

I agree, and if that's the case then based on my testing I'd say this PR fully resolves the issue we set out to solve.

pulpcore/tasking/tasks.py
```diff
@@ -146,6 +150,8 @@ def _queue_reserved_task(func, inner_task_id, resources, inner_args, inner_kwarg
                             worker=worker, resource=resource
                         )
                     TaskReservedResource.objects.create(resource=reservation, task=task_status)
+                worker.cleaned_up = False
+                worker.save(update_fields=["cleaned_up"])
```
mdellweg, line 154: This is what actually prevents the "tasks are assigned to dead workers" symptom by turning the worker into a zombie worker. The task is still doomed, but this will trigger proper cleanup.

pulpcore/tasking/worker.py
```diff
@@ -101,7 +102,7 @@ def perform_job(self, job, queue):
             queue (rq.queue.Queue): The Queue associated with the job
         """
         try:
-            task = Task.objects.get(pk=job.get_id())
+            task = Task.objects.get(pk=job.get_id(), state=TASK_STATES.WAITING)
```
mdellweg, line 105: This prevents the worker to pick up a task that is not waiting.

pulpcore/tasking/worker.py
```diff
@@ -110,8 +111,8 @@ def perform_job(self, job, queue):
             _set_current_user(user)
             GuidMiddleware.set_guid(task.logging_cid)
 
-        with TaskWorkingDirectory(job):
-            return super().perform_job(job, queue)
+            with TaskWorkingDirectory(job):
+                return super().perform_job(job, queue)
```
mdellweg, line 115: And this will stop the worker from performing a task that has no valid record in postgres.

bmbouter: Kind of funny (ish) but indenting this actually hard breaks all tasking because this code wasn't working like we thought it was!

Even against master, the try/except/else above always yields `Task.DoesNotExist` because the RQ job is not the pk! Instead this query above needs to be:

`Task.objects.get(_resource_job_id=job.get_id(), state=TASK_STATES.WAITING)`

bmbouter: Actually my analysis here isn't right either. The dispatch to the resource manager calls this once and there is no Task at all there so by indenting this we are causing the `_queue_reserve_resource` task to never be run so we need to leave it unindented.

We had broken down the concerning case of data leftover in RQ into two cases:
* there is already a task in the DB at cancelled and we accidentally run it
* there is no task in the DB because so much time has passed the user has deleted it already

What I realize now is that to fix the former first bullet we can add a snippet inside the if like:

```
if task.state != TASK_STATES.WAITING:
    return
```

However we have no good way to resolve the latter case (the task was already deleted yet kept in RQ) because that's the same situation the `_queue_reserved_resource` task itself needs to run as. I think that is such an unbelievable corner case we should ignore it and just add the task condition check above, unindent this, and that's it.

In summary I think this section should be:

```
        try:
            task = Task.objects.get(pk=job.get_id())
        except Task.DoesNotExist:
            pass
        else:
            if task.state != TASK_STATES.WAITING:
                return
            task.set_running()
            user = get_users_with_perms(task).first()
            _set_current_user(user)
            set_guid(task.logging_cid)

        with TaskWorkingDirectory(job):
            return super().perform_job(job, queue)
```

mdellweg: I'm almost tempted to say, we can check for `pk` and `_resource_job_id`. But I fear, this may introduce more problems than it solves.

mdellweg: ...and the resource_job will only attempt to clean up reserved resources records, and fail on it's own if the task is not found.

bmbouter: This is not exactly the issue we set out to solve but one we uncovered along the way. @mdellweg what can be done to finalize this PR and get it backported to 3.14 and 3.7 https://pulp.plan.io/issues/9118

mdellweg: I made it look the way you proposed, and i am happy to ship it that way.
My idea was to maybe solve it even better at the cost of "Maybe i didn't see this strange side effect and now everything is broken.". Therefore i rather want to leave it that way.
Or am getting you wrong and you are actually supposing to move this extra fix to an extra PR?

bmbouter: I'm proposing we make this function like [this comment shows](https://github.com/pulp/pulpcore/pull/1505/files#r673442829) and shipping it. Is that what you want to do also? Maybe push the code and we can say "oh yes this is what we want or no" more easily.

pulpcore/tasking/tasks.py
```diff
@@ -47,15 +47,19 @@ def _acquire_worker(resources):
...
     # Find a worker who already has one of the reservations, it is safe to send this work to them
     try:
-        return (
+        worker = (
             Worker.objects.select_for_update()
             .filter(pk__in=Worker.objects.filter(reservations__resource__in=resources))
             .get()
         )
+        if worker.online:
+            return worker
```
dralley, line 56: I believe this could be replaced with the `online_workers()` manager function

https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/task.py#L76

dralley (approved):
These changes make sense to me, good work @mdellweg and @bmbouter

pulpcore/tasking/tasks.py
mdellweg: No, it can't because we _want_ it to find the offline worker that blocks the resource.
We then hold off delivering that task until the worker is cleaned up (the resources are freed again).

pulpcore/tasking/worker.py
dralley: I have no strong feelings, maybe a minor preference for the way it was written in Brian's comment, but AFAICT the two are functionally equivalent

pulpcore/tasking/tasks.py
dralley: +1

pulpcore/tasking/worker.py
bmbouter: @mdellweg I didn't see a push, did you mean to push a change?

mdellweg: maybe yes...

bmbouter (approved):
I hand tested this again and it seems to be working great!

PR #1507, opened 2021-07-21
Fix JSONField deprecation warnings

daviddavis:
fixes #9112


pulpcore/app/migrations/0023_change_exporter_models.py
```diff
@@ -20,7 +19,7 @@ class Migration(migrations.Migration):
                 ('pulp_id', models.UUIDField(default=uuid.uuid4, editable=False, primary_key=True, serialize=False)),
                 ('pulp_created', models.DateTimeField(auto_now_add=True)),
                 ('pulp_last_updated', models.DateTimeField(auto_now=True, null=True)),
-                ('params', django.contrib.postgres.fields.jsonb.JSONField(null=True)),
+                ('params', models.JSONField(null=True)),
```
fao89, line 22: I think you shouldn't change the migrations, only the models, and then migrate

pulpcore/app/models/repository.py
```diff
@@ -262,6 +260,7 @@ class Remote(MasterModel):
         connect_timeout (models.FloatField): Value for aiohttp.ClientTimeout.connect
         sock_connect_timeout (models.FloatField): Value for aiohttp.ClientTimeout.sock_connect
         sock_read_timeout (models.FloatField): Value for aiohttp.ClientTimeout.sock_read
+        headeres (models.JSONField): Headers set on the aiohttp.ClientSession
```
dralley, line 263: Typo

pulpcore/app/migrations/0023_change_exporter_models.py
mdellweg: If you don't change the migrations, does that need we can never drop the django.contrib dependency? Is that even a thing?

dralley: Change the models and then do whatever makemigrations does ^

daviddavis: I'm not sure but it doesn't trigger any deprecation warnings so I think we're good.

fao89: I'm seeing some jsonb issues on pulp_ansible, I'm investigating
https://github.com/pulp/pulp_ansible/runs/3125697259?check_suite_focus=true#step:11:1681

dralley (approved):

pulpcore/app/migrations/0023_change_exporter_models.py
fao89: failure is not related to this

mdellweg (approved):
Looks good to me. And keeping the migrations intact seems to be correct according to this quote from django:
'django.contrib.postgres.fields.JSONField is removed except for support in historical migrations.'

PR #1508, opened 2021-07-21
Removes `http_giveup` method from plugin API

bmbouter:
This is no longer needed in the plugin API because, starting with
`pulpcore==3.14` the `HttpDownloader.run()` method uses it directly so
plugins no longer have to use it on their `_run()` subclass definitions.

closes #8913

### DIFF
--- a/CHANGES/plugin_api/8913.removal
+++ b/CHANGES/plugin_api/8913.removal
@@ -1,0 +1,5 @@
+The ``pulpcore.plugin.download.http_giveup`` method has been removed from the plugin API. Plugins
+used to have to use this to wrap the ``_run`` method defined on subclasses of ``HttpDownloader``,
+but starting with pulpcore 3.14 the backoff is implemented directly in the ``HttpDownloader.run()``
+method which subclasses do not override. Due to ``pulpcore`` implementing it, it is no longer needed
+or available for plugins to use.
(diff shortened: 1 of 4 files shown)


bmbouter:
@fao89 this will break pulp_ansible, would you be able to open a fix for it?

dralley (approved):

bmbouter:
@pulp/rpm FYI I believe since your code subclasses `HttpDownloader` you also will need to open a PR in preparation for this.

dralley:
@bmbouter https://github.com/pulp/pulp_ansible/pull/588

bmbouter:
@pulp/rpm oh actually maybe not:  https://github.com/pulp/pulp_rpm/search?q=http_giveup

fao89:
> @fao89 this will break pulp_ansible, would you be able to open a fix for it?

yes, but pulp_ansible is currently broken and I'm investigating why (I believe it is related to django 3)

dralley:
Pulp RPM is also broken due to Django 3 FWIW

bmbouter:
@fao89 I was pointed by @dralley to https://github.com/pulp/pulp_ansible/pull/588 so I don't think there is anything to do.

goosemania (approved):

PR #1512, opened 2021-07-22
Improve ProgressReport.increment efficiency

fao89:
closes #9129

### DIFF
--- a/pulpcore/app/models/progress.py
+++ b/pulpcore/app/models/progress.py
@@ -8,6 +8,8 @@ from gettext import gettext as _
 
 from django.db import models
 from django.utils import timezone
+from asgiref.sync import sync_to_async
+
 
 from pulpcore.app.models import BaseModel, Task
 from pulpcore.constants import TASK_CHOICES, TASK_STATES
@@ -175,6 +177,7 @@ class ProgressReport(BaseModel):
             self.state = TASK_STATES.FAILED
         self.save()
 
+    @sync_to_async
     def increment(self):
         """
         Increment done count and save the progress report.
--- a/CHANGES/9129.misc
+++ b/CHANGES/9129.misc
@@ -1,0 +1,1 @@
+Improve ProgressReport.increment efficiency


pulpcore/app/models/progress.py
```diff
@@ -149,8 +149,8 @@ def __enter__(self):
         self.state = TASK_STATES.RUNNING
         self.save()
 
-        # Save needs occurs immediately so it is called before _using_context_manager is set
-        self._using_context_manager = True
+        # Save needs occurs immediately so it is called before _use_batch_interval is set
+        self._use_batch_interval = True
```
bmbouter, line 153: Renaming this is confusing me. It's tracking if the ProgressReport is being used with a context manager or not.

pulpcore/app/models/progress.py
```diff
@@ -182,7 +182,9 @@ def increment(self):
         This will increment and save the self.done attribute which is useful to put into a loop
         processing items.
         """
+        self._use_batch_interval = True
```
bmbouter, line 185: Also this confuses me to, if we are in the context manager why wouldn't we continue to batch the calls? I know why you want to change it because it fixes the bug, but the intent was for increment to by batched when used in context managers and this disables that.

fao89: it doesn't disable that, context manager still as it was, I'm just making `increment` to use the BATCH_INTERVAL

fao89: I tested before, using this trick on pulp_ansible: https://github.com/pulp/pulp_ansible/pull/604/files

bmbouter: By modifying the value here it will call every call to `increase_by` to call `save()` which will actually save to the db. That is not the intent of the design in this area of the code.

fao89: I see what you mean

dralley:
Is this going to continue working as intended in synchronous contexts?

fao89:
> Is this going to continue working as intended in synchronous contexts?

I'm thinking about just adding 
`log.warn("Please avoid increment on async methods outside a context manager) ` on increment,
and move forward with https://github.com/pulp/pulp_ansible/pull/605

PR #1513, opened 2021-07-22
Loosen the constraints around what can be in a new-repository-version.

ggainey:
There are RPM repositories "in the wild" that violate Pulp's assumptions
about what is 'legal' in an incoming repository version. This commit
changes some fatal errors into log-warnings, along with some heuristics
around how to get a 'reasonable' repository in the face of suboptimal
data.

It also handles an exception during repo-version-failure-cleanup that
could result in losing error info.

fixes #8133
[nocoverage]

### DIFF
--- a/pulpcore/app/files.py
+++ b/pulpcore/app/files.py
@@ -128,21 +128,34 @@ def validate_file_paths(paths):
     overlap_error = _("The path for file '{path}' overlaps: {conflicts}")
 
     path_trie = StringTrie(separator="/")
+    dups = []
+    overlaps = []
     for path in paths:
         if path in path_trie:
             # path duplicates a path already in the trie
-            raise ValueError(_("Path is duplicated: {path}").format(path=path))
-
-        if path_trie.has_subtrie(path):
+            dups.append(path)
+        elif path_trie.has_subtrie(path):
             # overlap where path is 'a/b' and trie has 'a/b/c'
             conflicts = [item[0] for item in path_trie.items(prefix=path)]
-            raise ValueError(overlap_error.format(path=path, conflicts=(", ").join(conflicts)))
-
-        prefixes = list(path_trie.prefixes(path))
-        if prefixes:
-            # overlap where path is 'a/b/c' and trie has 'a/b'
-            conflicts = [prefix.key for prefix in prefixes]
-            raise ValueError(overlap_error.format(path=path, conflicts=(", ").join(conflicts)))
+            overlaps.append(overlap_error.format(path=path, conflicts=", ".join(conflicts)))
+        else:
+            prefixes = list(path_trie.prefixes(path))
+            if prefixes:
+                # overlap where path is 'a/b/c' and trie has 'a/b'
+                conflicts = [prefix.key for prefix in prefixes]
+                overlaps.append(overlap_error.format(path=path, conflicts=", ".join(conflicts)))
 
         # if there are no overlaps, add it to our trie and continue
         path_trie[path] = True
+
+    if dups or overlaps:
+        dups_msg = ""
+        overlaps_msg = ""
+        if dups:
+            dups_msg = _("Paths are duplicated: {paths}").format(paths=",".join(dups))
+        if overlaps:
+            overlaps_msg = "\n".join(overlaps)
+
+        raise ValueError(
+            _("Path errors found. {dups}\n{overlaps}").format(dups=dups_msg, overlaps=overlaps_msg)
+        )
--- a/pulpcore/app/models/publication.py
+++ b/pulpcore/app/models/publication.py
@@ -184,7 +184,14 @@ class Publication(MasterModel):
             exc_tb (types.TracebackType): (optional) stack trace.
         """
         if exc_val:
-            self.delete()
+            # If an exception got us here, the Publication we were trying to create is
+            # Bad, and we should delete the attempt. HOWEVER - some exceptions happen before we
+            # even get that far. In those cases, calling delete() results in a new not-very-useful
+            # exception being raised and reported to the user, rather than the actual problem.
+            try:
+                self.delete()
+            except Exception:
+                raise exc_val.with_traceback(exc_tb)
         else:
             try:
                 self.finalize_new_publication()
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -262,6 +262,10 @@ class RemoteArtifactSaver(Stage):
     :class:`~pulpcore.plugin.stages.DeclarativeArtifact`.
     """
 
+    def __init__(self, fix_mismatched_remote_artifacts=False, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fix_mismatched_remote_artifacts = fix_mismatched_remote_artifacts
+
     async def run(self):
         """
         The coroutine for this stage.
@@ -320,8 +324,51 @@ class RemoteArtifactSaver(Stage):
                     if d_artifact.relative_path == content_artifact.relative_path:
                         break
                 else:
-                    msg = _('No declared artifact with relative path "{rp}" for content "{c}"')
-                    raise ValueError(msg.format(rp=d_artifact.relative_path, c=d_content.content))
+                    if self.fix_mismatched_remote_artifacts:
+                        # We couldn't match an DeclarativeArtifact to a ContentArtifact by rel_path.
+                        # If there are any paths available (i.e., other ContentArtifacts for this
+                        # Artifact), complain to the logs, pick the rel_path from the last
+                        # ContentArtifact we examined, and continue.
+                        #
+                        # If we can't find anything to choose from (can that even happen?), fail
+                        # the process.
+                        avail_paths = ",".join(
+                            [
+                                ca.relative_path
+                                for ca in d_content.content._remote_artifact_saver_cas
+                            ]
+                        )
+                        if avail_paths:
+                            msg = _(
+                                "No declared artifact with relative path '{rp}' for content '{c}'"
+                                " from remote '{rname}'. Using last from available-paths : '{ap}'"
+                            )
+                            log.warning(
+                                msg.format(
+                                    rp=d_artifact.relative_path,
+                                    c=d_content.content.filename,
+                                    rname=d_artifact.remote.name,
+                                    ap=avail_paths,
+                                )
+                            )
+                            d_artifact.relative_path = content_artifact.relative_path
+                        else:
+                            msg = _(
+                                "No declared artifact with relative path '{rp}' for content '{c}'"
+                                " from remote '{rname}', and no paths available."
+                            )
+                            raise ValueError(
+                                msg.format(
+                                    rp=d_artifact.relative_path,
+                                    c=d_content.content.filename,
+                                    rname=d_artifact.remote.name,
+                                )
+                            )
+                    else:
+                        msg = _('No declared artifact with relative path "{rp}" for content "{c}"')
+                        raise ValueError(
+                            msg.format(rp=d_artifact.relative_path, c=d_content.content)
+                        )
 
                 for remote_artifact in content_artifact._remote_artifact_saver_ras:
                     if remote_artifact.remote_id == d_artifact.remote.pk:
--- a/CHANGES/8133.bugfix
+++ b/CHANGES/8133.bugfix
@@ -1,0 +1,1 @@
+In stages-pipeline and new-version sanity-checks, added full error-info on path-problems.
--- a/CHANGES/plugin_api/8133.bugfix
+++ b/CHANGES/plugin_api/8133.bugfix
@@ -1,0 +1,4 @@
+Added kwarg to RemoteArtifactSaver init to allow enabling handling of rare error edge-case.
+
+`fix_mismatched_remote_artifacts=True` enables workaround for a failure-scenario that
+(so far) is only encountered by pulp_rpm. Current behavior is the default.
--- a/pulpcore/plugin/repo_version_utils.py
+++ b/pulpcore/plugin/repo_version_utils.py
@@ -119,11 +119,10 @@ def validate_version_paths(version):
     paths = ContentArtifact.objects.filter(content__pk__in=version.content).values_list(
         "relative_path", flat=True
     )
-
     try:
         validate_file_paths(paths)
     except ValueError as e:
-        raise ValueError(_("Cannot create repository version. {err}.").format(err=e))
+        raise ValueError(_("Repository version errors : {err}").format(err=e))
 
 
 def validate_repo_version(version):
--- a/pulpcore/tests/unit/test_files.py
+++ b/pulpcore/tests/unit/test_files.py
@@ -19,7 +19,9 @@ class TestValidateFilePaths(TestCase):
         Test for two duplicate paths.
         """
         paths = ["a/b", "PULP_MANIFEST", "PULP_MANIFEST"]
-        with self.assertRaisesRegex(ValueError, "Path is duplicated: PULP_MANIFEST"):
+        with self.assertRaisesRegex(
+            ValueError, "Path errors found. Paths are duplicated: PULP_MANIFEST"
+        ):
             validate_file_paths(paths)
 
     def test_overlaps(self):


pulpcore/app/files.py
```diff
@@ -128,21 +128,35 @@ def validate_file_paths(paths):
...
         if path in path_trie:
             # path duplicates a path already in the trie
-            raise ValueError(_("Path is duplicated: {path}").format(path=path))
+            dups.append(path)
 
         if path_trie.has_subtrie(path):
             # overlap where path is 'a/b' and trie has 'a/b/c'
             conflicts = [item[0] for item in path_trie.items(prefix=path)]
-            raise ValueError(overlap_error.format(path=path, conflicts=(", ").join(conflicts)))
+            overlaps.append(overlap_error.format(path=path, conflicts=", ".join(conflicts)))
```
goosemania, line 141: Since we are no longer raising the exception, we need an `elif` here, otherwise, paths ended up being logged under both categories if they are duplicates.

With the opttools repo:
```
Jul 26 10:15:27 pulp3-source-fedora34.rhgoose.example.com pulpcore-worker[118421]: pulp [7a1f9b1adf794f9c903a5def62562e56]: pulp_rpm.app.models.repository:WARNING: New version of repository c7opttools reports duplicate/overlap errors : Repository version errors : Path errors found. Paths are duplicated: rubygem-rspec-expectations-3.5.0-1.el7.1.noarch.rpm,rubygem-rspec-core-3.5.3-2.el7.0.noarch.rpm,rubygem-patron-0.4.20-3.el7.x86_64.rpm,rubygem-fluent-plugin-elasticsearch-0.7.0-1.el7.noarch.rpm,rubygem-test-unit-3.1.9-2.el7.noarch.rpm,rubygem-yajl-ruby-1.3.0-3.el7.x86_64.rpm,rubygem-patron-doc-0.4.20-3.el7.noarch.rpm,rubygem-elasticsearch-api-1.0.7-1.el7.noarch.rpm,rubygem-elasticsearch-doc-1.0.8-1.el7.noarch.rpm,rubygem-diff-lcs-doc-1.2.5-7.el7.noarch.rpm,rubygem-elasticsearch-1.0.8-1.el7.noarch.rpm,python-msgpack-0.4.6-3.el7.x86_64.rpm,rubygem-yajl-ruby-doc-1.3.0-3.el7.noarch.rpm,rubygem-multipart-post-2.0.0-2.el7.noarch.rpm,rubygem-rspec-expectations-doc-3.5.0-1.el7.1.noarch.rpm,rubygem-power_assert-0.3.1-1.el7.noarch.rpm,rubygem-faraday-doc-0.9.0-3.el7.noarch.rpm,rubygem-elasticsearch-api-doc-1.0.7-1.el7.noarch.rpm,rubygem-rspec-3.5.0-1.el7.noarch.rpm,rubygem-faraday-0.9.0-3.el7.noarch.rpm,rubygem-elasticsearch-transport-1.0.7-1.el7.noarch.rpm,rubygem-rspec-doc-3.5.0-1.el7.noarch.rpm,rubygem-test-unit-doc-3.1.9-2.el7.noarch.rpm,rubygem-rspec-core-doc-3.5.3-2.el7.0.noarch.rpm,rubygem-http_parser.rb-doc-0.6.0-1.el7.noarch.rpm,rubygem-multipart-post-doc-2.0.0-2.el7.noarch.rpm,rubygem-string-scrub-doc-0.0.5-3.el7.noarch.rpm,rubygem-power_assert-doc-0.3.1-1.el7.noarch.rpm,rubygem-http_parser.rb-0.6.0-1.el7.x86_64.rpm,rubygem-diff-lcs-1.2.5-7.el7.noarch.rpm,rubygem-fluent-plugin-add-doc-0.0.3-1.el7.noarch.rpm,rubygem-rspec-mocks-doc-3.5.0-1.el7.1.noarch.rpm,rubygem-string-scrub-0.0.5-3.el7.x86_64.rpm,rubygem-rspec-mocks-3.5.0-1.el7.1.noarch.rpm,rubygem-fluent-plugin-elasticsearch-doc-0.7.0-1.el7.noarch.rpm,rubygem-elasticsearch-transport-doc-1.0.7-1.el7.noarch.rpm,rubygem-fluent-plugin-add-0.0.3-1.el7.noarch.rpm
Jul 26 10:15:27 pulp3-source-fedora34.rhgoose.example.com pulpcore-worker[118421]: The path for file 'rubygem-rspec-expectations-3.5.0-1.el7.1.noarch.rpm' overlaps: rubygem-rspec-expectations-3.5.0-1.el7.1.noarch.rpm
Jul 26 10:15:27 pulp3-source-fedora34.rhgoose.example.com pulpcore-worker[118421]: The path for file 'rubygem-rspec-core-3.5.3-2.el7.0.noarch.rpm' overlaps: rubygem-rspec-core-3.5.3-2.el7.0.noarch.rpm
Jul 26 10:15:27 pulp3-source-fedora34.rhgoose.example.com pulpcore-worker[118421]: The path for file 'rubygem-patron-0.4.20-3.el7.x86_64.rpm' overlaps: rubygem-patron-0.4.20-3.el7.x86_64.rpm
Jul 26 10:15:27 pulp3-source-fedora34.rhgoose.example.com pulpcore-worker[118421]: The path for file 'rubygem-fluent-plugin-elasticsearch-0.7.0-1.el7.noarch.rpm' overlaps: rubygem-fluent-plugin-elasticsearch-0.7.0-1.el7.noarch.rpm
Jul 26 10:15:27 pulp3-source-fedora34.rhgoose.example.com pulpcore-worker[118421]: The path for file 'rubygem-test-unit-3.1.9-2.el7.noarch.rpm' overlaps: rubygem-test-unit-3.1.9-2.el7.noarch.rpm
Jul 26 10:15:27 pulp3-source-fedora34.rhgoose.example.com pulpcore-worker[118421]: The path for file 'rubygem-yajl-ruby-1.3.0-3.el7.x86_64.rpm' overlaps: rubygem-yajl-ruby-1.3.0-3.el7.x86_64.rpm
```

goosemania:
Tested, works well for me.

pulpcore/app/files.py
ggainey: Ahhh, great catch thank you!

pulpcore/app/models/publication.py
```diff
@@ -174,7 +174,14 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             exc_tb (types.TracebackType): (optional) stack trace.
         """
         if exc_val:
-            self.delete()
+            # If an exception got us here, the Publication we were trying to create is
+            # Bad, and we should delete the attempt. HOWEVER - some exceptions happen before we
+            # even get that far. In those cases, calling delete() results in a new not-very-useful
+            # exception being raised and reported to the user, rather than the actual problem.
```
dralley, line 180: :+1:

pulpcore/app/files.py
```diff
@@ -128,21 +128,34 @@ def validate_file_paths(paths):
...
     path_trie = StringTrie(separator="/")
+    dups = []
+    overlaps = []
     for path in paths:
         if path in path_trie:
             # path duplicates a path already in the trie
-            raise ValueError(_("Path is duplicated: {path}").format(path=path))
-
-        if path_trie.has_subtrie(path):
+            dups.append(path)
```
dralley, line 136: Suggestion:

Make a custom exception type that takes the list(s) of conflicts and overlaps in its constructor, and implements `__str__` to provide the error message.  That will probably simplify debugging a bit, and moves some of this string formatting code out of the logic path.

Question:

Do you think there would be any value in retaining the short-circuit behavior and putting it behind a flag?  I'm not sure if it's worth it but I think the general case doesn't need to do the full evaluation like this.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -326,8 +326,42 @@ def _needed_remote_artifacts(self, batch):
...
                 else:
-                    msg = _('No declared artifact with relative path "{rp}" for content "{c}"')
-                    raise ValueError(msg.format(rp=d_artifact.relative_path, c=d_content.content))
+                    # We couldn't match an DeclarativeArtifact to a ContentArtifact by rel_path.
+                    # If there are any paths available (i.e., other ContentArtifacts for this
+                    # Artifact), complain to the logs, pick the rel_path from the last
+                    # ContentArtifact we examined, and continue.
+                    #
+                    # If we can't find anything to choose from (can that even happen?), fail
+                    # the process.
```
dralley, line 335: After a few minutes of thought I haven't yet come up with any examples where this would come back to haunt us but this change in particular needs a lot of scrutiny and soak-time, I think.

pulpcore/app/files.py
ggainey: I considered specific exceptions, but was concerned about breaking plugins that were explicitly looking for ValueError.

RE short-circuit: honestly, I find "fail on first-found" behavior to be pretty frustrating. You hit it, you read and respond to the error, you sync....and then you hit it **again** , for a different path. This code knows exactly how many of these issues exist, and for which paths - it's rude to not share the info with the user, even at the expense of an "ugly" log entry.

pulpcore/plugin/stages/artifact_stages.py
ggainey: I spent a lot of time (probably too much time) trying to do somehting "more correct", without success.

The only way to get to the point where we "pick one" is:
- There exist multiple contentartifacts
- That point to the same Artifact
- That have relative_paths
- None of which match the relative_path in the examined DeclarativeArtifact.

In this failure-case, we pick one of the contentartifacts as the 'winner'. The artifact in question will be used to create a RemoteArtifact.

The class of rpm-repo where this is an issue is one where there exists more than one NEVRA, with the same checksum - specifically, a given <package> stanza appears more than once in primary.xml, with different location_hrefs.  In Pulp2, we do exactly what we're doing here - only one of them make it into the repository.

This should be a never-happen case - but createrepo is perfectly happy to allow it to happen, as you can see here : https://fixtures.pulpproject.org/diff-name-same-content/ And even when the repo looks this way, we may not fall into this edge case, depending on the exact code-path, entity-ordering, and use of sets-by-package-name that happened before we get to this code.

Hopefully this helps clarify what's going on here. It's an edge-case of an edge-case, but it affects some common third-party repositories that used to **look like** they Just Worked, and is most-easily reproduced after migrating such repositories from Pulp2.

pulpcore/app/files.py
```diff
@@ -128,21 +128,34 @@ def validate_file_paths(paths):
...
+    if dups or overlaps:
+        dups_msg = ""
+        overlaps_msg = ""
+        if dups:
+            dups_msg = _("Paths are duplicated: {paths}").format(paths=",".join(dups))
+        if overlaps:
+            overlaps_msg = "\n".join(overlaps)
+
+        raise ValueError(
+            _("Path errors found. {dups}\n{overlaps}").format(dups=dups_msg, overlaps=overlaps_msg)
```
bmbouter, line 160: @ggainey is it right that the next of all the change in this file is that a `ValueError` will still be raised just with a all the issues not just one?

ggainey: @bmbouter that is correct - all the changes in files.py are to raise the same error it was raising, but with enough information to know what **all** the problems were, instead of just the first one. The exception raised to the caller is the same, but now with a description of the **whole** problem.

pulpcore/plugin/stages/artifact_stages.py
bmbouter: I read this several times, and I really apprecaite you writing it out. I *think* I have internalized the concerning case. I also stepped through the code to remind myself of how the stage worked, but I couldn't reproduce with pulp_file at least.

Since I couldn't reproduce with pulp_file (or think of any other way in any plugin except RPM), I got to thinking:  will this affect more than one plugin or RPM only? What would be an example of another plugin for instance?

I ask because we could likely make this stage a lot easier to subclass, and have the RPM use a slightly modified version instead. It already carries a custom pipeline [here](https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/tasks/synchronizing.py#L441-L470).

My thinking is that would be more beneficial because if this is really a special case affecting only 1 plugin, we should focus on the pulpcore code being more reusable and not putting the special case in the area where all plugins actually receive that code.

Also by having the fix actually in pulp_rpm, we could have the change there come with a functional test which asserts it runs correctly. As it is here, we can't have a functional test that asserts on this because we can't trigger it and the reason is because it can't be triggered outside of pulp_rpm (that I can tell).

ggainey: I'm a big fan of "easier to subclass". Complicated data sources are almost certainly going to benefit from being able to do "odd things" in the stages pipeline!  

However, I content that that needs to be its own Thing, for entirely-pragmatic reasons. The immediate problem affects getting users moved from Pulp2 to Pulp3, and there are restrictions on which versions of core-and-plugins in which it needs to be fixed. "Not destabilizing pulpcore any more than absolutely necessary at this time" is a valid constraint.

We still want to do "the right thing" architecturally, I've opened https://pulp.plan.io/issues/9170 as a place holder for that work. I'd think that would be the right time to also do some core-refactoring as well. wdyt?

bmbouter: I see where you're coming from. What versions of core-and-plugins does this need to be fixed in? Let's make my suggestion even simpler: we could make no changes in the `RemoteArtifactsaver` pulpcore and instead have the versions of pulp_rpm that need this fix use a custom `RemoteArtifactsaver` stage that basically includes exactly this code. Would that accomplish both of our goals?

I hear you that it's a workaround, but we still need the workaround going to the right place. If this is affecting multiple plugins then this is the right place, but if it's not then I don't think it is. I'm still trying to get on solid group if we have another in the wild example of a second plugin that would be affected or not.

dralley: @bmbouter I think the obvious first place to look is - does it affect the Debian plugin?  And it might not, because IIRC the way they worked around this problem was by shoving the path into the content object itself just like the file plugin does, which is *not* a good solution.

ggainey: It impacts migrations, so it needs to backport to core/3.7 rpm/3.10, as well as core/3.14/rpm/3.14

I'm not averse to having pulling this into pulp_rpm - but I'm not convinced having a second copy of this (opaque, complicated) code in two places, is an improvement over "fix in core an edge-case that only one plugin (that we know of) has".

dralley: >I'm not convinced having a second copy of this (opaque, complicated) code in two places, is an improvement over "fix in core an edge-case that only one plugin (that we know of) has".

Neither am I

ipanova: I have read through the code and convo and it seems like these changes should not impact pulp-container. I have also sanity checked this PR by syncing a docker repo.

Regarding the question where to keep this code..both of the views have their arguments. I would not be against keeping this in core, given that this is a special case and plugins can confirm they're not affected. I acknowledge that this goes a bit against our policy ' keep in core changes that more that one plugin would benefit from' however the maintenance cost would be much higher for the rpm plugin if this would be subclassed.

goosemania: TL;DR: I'm leaning towards keeping it in pulpcore, but at the end of the day, I care only about it being merged anywhere.

I do not think there is one definite answer where this code has to be. It all depends how we look at it.
The only thing which I'd like to eliminate is the thinking that it's purely RPM problem. Currently, from the plugins we are aware of, only RPM suffers from it but only because it chose to do things "the right way" (arguably, of course) from user experience perspective. File and Debian plugins "solved" it by creating duplicates of  content which differ by path, and that creates a mess and confusion when working with content, listing or copying it.

As for the code location, when plugin needs to do something absolutely different, sure, that needs to go into a plugin.
Here we are loosening some constraints which were based on the wrong assumptions of what's possible/allowed, we do not introduce any plugin specific logic. So I can see this change being in pulpcore, instead of later duplicating fixes and/or improvements which will happen in this piece of code. (Like the recent fix to the proxy configuration of a remote, it required fixes in 4 places because of custom downloaders :facepalm:)

If there is a concern that it will break something, maybe it's a strong enough argument to put this code into a plugin.

In other situation, I would be worried that the problem will be forgotten when re-architecture time comes, but the problem is so big and known by many people by now, that even if it ends up in pulp_rpm code, there is a high chance someone will remember and bring it up.

bmbouter: TL;DR please do not merge this change here. To me, it's not the right place for it. Instead I'm asking that code specific to pulp_rpm be added to pulp_rpm.

There are several reasons why I believe this isn't the right place to put this code. I'll try to summarize them.

1. Putting it here prevents us from functionally testing it when merging against master and at backport time. Notice this PR has no tests, meaning we have no safety around this contribution. By putting it in pulp_rpm the change can come with a functional test. This is a clear signal to me it's the wrong place.

2. pulpcore is the shared area for code that is needed by more than one plugin. That's why the stages pipeline is customizable. Our strategy has been that one-off needs from plugins need to be handled with customized stages. I'm concerned that the reasoning I'm reading to put it here is that "it does no harm" and not "it's what we need for this set of plugins". Having plugin code end up in pulpcore is one of the things we set out to strongly avoid when we started Pulp3.

3. By putting it here we'll have to perform backport releases for pulpcore, but by putting it in pulp_rpm we won't. I'm reading concerns about level of effort, and I share those, so let's go the path of lower effort.

4. Other plugins may not want this, but by adding here, we're making every plugin have it. It's likely other plugins want to know about this problem by users experiencing the problem, but with this change, the chance of that is very low as it will pass silently. I've read we all agree this isn't the correct solution, so passing silently while using an incorrect solution is a risky position for us to accept as the default for all plugins.

If folks are concerned about the duplication of code, there are at least two good options to avoid that:

* Rework this PR to decompose the `_needed_remote_artifacts` method into smaller functions and have pulp_rpm use those. I don't think we should do this for backports, but we could do it for master.

* Have the pulp_rpm custom pipeline construct a new `RemoteArtifactSaver` by wrapping the `_needed_remote_artifacts` at runtime and then using the newly constructed object in the pulp_rpm pipeline. This would require no changes on pulpcore.

dralley: > By putting it in pulp_rpm the change can come with a functional test. This is a clear signal to me it's the wrong place.

I disagree.  There are plenty of pulpcore features used only by (e.g.) the container plugin or the ansible plugin that have no tests in pulpcore.  What you're identifying is not a signal that this change deserves to go in pulp_rpm, but a signal that the current status quo of _only using pulp_file_ for core tests is inadequate (and probably already was).

>pulpcore is the shared area for code that is needed by more than one plugin. That's why the stages pipeline is customizable. Our strategy has been that one-off needs from plugins need to be handled with customized stages. 

I think what both Tanya and I have tried to say is that this is **_not_** a pulp_rpm specific concern.  Even if it is the only plugin _currently_ affected - that is only because other plugins have implemented a workaround that has poor properties for the user. 

>By putting it here we'll have to perform backport releases for pulpcore, but by putting it in pulp_rpm we won't.

Why would we not?  6.9 is not running the latest version of the RPM plugin, it is not compatible.  So we have to do backports regardless and the RPM backport would actually be *much harder* since the sync code has undergone so many changes since 3.10 (the version in 6.9).

But regardless I'm not sure this is a good reason for choosing one location over another, I'd rather stick to the technical merits.

>Other plugins may not want this

At least to the best of my memory, the Debian plugin already did express a desire for this. Granted this was at least a year ago so I would love to hear @quba42 's thoughts currently.

dralley: ^ None of this should be interpreted as "-1 to putting it in pulp_rpm" but I'd like us to do it for the right reasons if that's the route we go with.  One potential good reason would be risk.

bmbouter: @dralley @goosemania @ggainey @ipanova How are we going to reach an agreement on this before I go on leave?

goosemania: I personally read that the outcome of this discussion as following:

1. There are no objections to error message collection/improvements, and they can be merged into pulpcore since everyone would benefit from it.
2. The only concern is the changes in the `pulpcore/plugin/stages/artifact_stages.py` file.
  - the majority of folks are leaning towards keeping it in pulpcore while the minority of folks have a strong opinion to move it to pulp_rpm.
  - my personal approach is not to create strong disappointments or frustrations, so I suggest to merge this part into pulp_rpm. At the end of the day, other plugins are already with their hacks in this area.
3. Refactor to duplicate less is a good approach in general, however at this stage, it's maybe not a good idea because of the backport nightmare which will come with it. 

I'm open to alternative outcomes of this discussion.

bmbouter: I've thought this over and while I'm uncomfortable with the change for reasons already stated, I am comfortable to deferring the decision to you all in terms of deciding if this is right. I've voiced my concerns, but if you still want to go forward with this, please do.

dkliban:
If no plugin (pulp_deb or pulp_file) is planning to make any changes to take advantage of this change in pulpcore, then there is no reason to add this to pulpcore. Let's keep the improved error message collection in pulpcore and move the rest of the code to pulp_rpm.

ggainey:
> If no plugin (pulp_deb or pulp_file) is planning to make any changes to take advantage of this change in pulpcore, then there is no reason to add this to pulpcore. Let's keep the improved error message collection in pulpcore and move the rest of the code to pulp_rpm.

It means copying this whole code path and duplicating it in pulp_rpm, because core isn't set up to allow us to subclass and do just the one fix. That is the crux of the problem.

ggainey:
@dkliban @goosemania @ipanova @dralley @bmbouter  Adjusted the PR to take dralley's suggestion to make the changed behavior selectable. Please re-review. pulp_rpm PR incoming that will take advantage of the new kwarg in the ctor.

dralley (approved):

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -262,6 +262,10 @@ class RemoteArtifactSaver(Stage):
     :class:`~pulpcore.plugin.stages.DeclarativeArtifact`.
     """
 
+    def __init__(self, fix_mismatched_remote_artifacts=False, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fix_mismatched_remote_artifacts = fix_mismatched_remote_artifacts
```
ipanova, line 267: @ggainey  the current changelog captures only path overlaps fixes, can we somehow also mention this change?maybe an entry in the plugin_api?

ggainey: Ah, good idea, will do.

pulpcore/plugin/stages/artifact_stages.py
dralley: @bmbouter We decided on a compromise.  The code goes in pulpcore but is disabled unless you build your own pipeline and set a parameter on the stage.  So only the RPM plugin will enable this code, other plugins will still get the old behavior, and individual plugins can turn it on or off as desired.  But there won't be any duplication of code.

We do still want to talk about unification of the codepaths but it can be a longer-term discussion and be more ambitious about fixing the problems.

dkliban (approved):

ipanova (approved):

PR #1514, opened 2021-07-23
Added worker_cleanup to the tasking system

mdellweg:
All workers that are considered offline will be removed from the database
periodically.

fixes #8931

### DIFF
--- a/pulpcore/app/models/task.py
+++ b/pulpcore/app/models/task.py
@@ -90,6 +90,22 @@ class WorkerManager(models.Manager):
 
         return self.filter(last_heartbeat__gte=age_threshold, gracefully_stopped=False)
 
+    def offline_workers(self):
+        """
+        Returns a queryset of workers meeting the criteria to be considered 'offline'
+
+        To be considered 'offline', a worker must have no recent heartbeat timestamp.
+        "Recent" is defined here as "within the pulp process timeout interval".
+
+        Returns:
+            :class:`django.db.models.query.QuerySet`:  A query set of the Worker objects which
+                are considered by Pulp to be 'offline'.
+        """
+        now = timezone.now()
+        age_threshold = now - timedelta(seconds=settings.WORKER_TTL)
+
+        return self.filter(last_heartbeat__lte=age_threshold)
+
     def missing_workers(self):
         """
         Returns a queryset of workers meeting the criteria to be considered 'missing'
--- a/pulpcore/tasking/pulpcore_worker.py
+++ b/pulpcore/tasking/pulpcore_worker.py
@@ -3,6 +3,7 @@ import importlib
 import json
 import logging
 import os
+import random
 import select
 import signal
 import socket
@@ -27,7 +28,7 @@ from django_currentuser.middleware import (  # noqa: E402: module level not at t
 )
 from django_guid import set_guid  # noqa: E402: module level not at top of file
 
-from pulpcore.app.models import Task  # noqa: E402: module level not at top of file
+from pulpcore.app.models import Worker, Task  # noqa: E402: module level not at top of file
 
 from pulpcore.constants import (  # noqa: E402: module level not at top of file
     TASK_STATES,
@@ -41,6 +42,10 @@ from pulpcore.tasking.worker_watcher import handle_worker_heartbeat  # noqa: E40
 
 
 _logger = logging.getLogger(__name__)
+random.seed()
+
+TASK_GRACE_INTERVAL = 3
+WORKER_CLEANUP_INTERVAL = 100
 
 
 class NewPulpWorker:
@@ -50,6 +55,10 @@ class NewPulpWorker:
         self.heartbeat_period = settings.WORKER_TTL / 3
         self.cursor = connection.cursor()
         self.worker = handle_worker_heartbeat(self.name)
+        self.task_grace_timeout = 0
+        self.worker_cleanup_countdown = random.randint(
+            WORKER_CLEANUP_INTERVAL / 10, WORKER_CLEANUP_INTERVAL
+        )
 
         # Add a file descriptor to trigger select on signals
         self.sentinel, sentinel_w = os.pipe()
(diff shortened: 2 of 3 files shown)


pulpcore/app/models/task.py
```diff
@@ -90,6 +90,22 @@ def online_workers(self):
...
+        Returns:
+            :class:`django.db.models.query.QuerySet`:  A query set of the Worker objects which
+                are considered by Pulp to be 'offline'.
+        """
+        now = timezone.now()
+        age_threshold = now - timedelta(seconds=settings.WORKER_TTL)
+
+        return self.filter(last_heartbeat__lte=age_threshold)
+
     def missing_workers(self):
```
dralley, line 109: The similarity of this query to `missing_workers()` concerns me if we are still using `missing_workers()` at all.  They will likely overlap and things that would have been handled by `missing_workers()` could get caught here first.

I'm also not sure that we want to clean up old workers immediately?  That information might be helpful for debugging purposes.  I think in the past we had discussed leaving them for a week, but perhaps that was a discussion about tasks.

mdellweg: The new tasking system does not care about `gracefully_stopped`, as there is basically no cleanup to be performed.
Workers that are shut down gracefully remove themselves from the database immediately (we could change that, but whom do we help here?).

pulpcore/app/models/task.py
```diff
@@ -90,6 +90,22 @@ def online_workers(self):
 
         return self.filter(last_heartbeat__gte=age_threshold, gracefully_stopped=False)
 
+    def offline_workers(self):
```
bmbouter, line 93: :+1:

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -41,6 +42,8 @@
 
 
 _logger = logging.getLogger(__name__)
+worker_cleanup_interval = 100
+random.seed()
```
bmbouter, line 46: I don't think this is necessary.

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -50,6 +53,10 @@ def __init__(self):
         self.heartbeat_period = settings.WORKER_TTL / 3
         self.cursor = connection.cursor()
         self.worker = handle_worker_heartbeat(self.name)
+        self.task_grace = 0
```
bmbouter, line 56: can this var be named more descriptive please?

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -50,6 +53,10 @@ def __init__(self):
...
+        self.worker_cleanup_countdown = random.randint(
+            worker_cleanup_interval / 10, worker_cleanup_interval
```
bmbouter, line 58: Making this literals would be a lot more readable to me. random.randint(10, 100) We're not making it user configurable anyway.

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -70,11 +77,22 @@ def shutdown(self):
         self.worker.delete()
         _logger.info(f"Worker {self.name} was shut down.")
 
+    def worker_cleanup(self):
+        qs = Worker.objects.offline_workers()
+        if qs:
+            for worker in qs:
+                _logger.info(f"Clean offline worker {worker.name}.")
```
bmbouter, line 84: Can this i8ln

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -70,11 +77,22 @@ def shutdown(self):
...
+
     def beat(self):
         if self.worker.last_heartbeat < timezone.now() - timedelta(seconds=self.heartbeat_period):
             self.worker = handle_worker_heartbeat(self.name)
             if self.shutdown_requested:
                 self.task_grace -= 1
+            self.worker_cleanup_countdown -= 1
+            if self.worker_cleanup_countdown <= 0:
+                self.worker_cleanup_countdown = worker_cleanup_interval
+                self.worker_cleanup()
```
bmbouter, line 95: Doesn't this need to also reset self.worker_cleanup_countdown = big_number?

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -70,11 +77,22 @@ def shutdown(self):
...
+        if qs:
+            for worker in qs:
+                _logger.info(f"Clean offline worker {worker.name}.")
+                worker.delete()
+
     def beat(self):
         if self.worker.last_heartbeat < timezone.now() - timedelta(seconds=self.heartbeat_period):
             self.worker = handle_worker_heartbeat(self.name)
             if self.shutdown_requested:
                 self.task_grace -= 1
```
bmbouter, line 91: How was this working before if never initialized?

pulpcore/tasking/pulpcore_worker.py
mdellweg: Trust me it is. Computers are notoriously bad at _random_.

bmbouter: Yes but if you never call seed() it's called implicitly is my understanding.

pulpcore/tasking/pulpcore_worker.py
mdellweg: There was only a very small window to hit this if the worker didn't work on a task.
Put the other way, shen select was interrupted by a signal instead of timeout, it is very unlikely that enough time for the next heartbeat had elapsed.

pulpcore/tasking/pulpcore_worker.py
mdellweg: What do you mean "also"? It does and it's the only pace.

pulpcore/tasking/pulpcore_worker.py
mdellweg: I'm using the Big number in two places, and i pluck that one out of thin air.
Maybe it should scale down with an increasing number of workers. That's why i made it a constant. It should be ALLCAPS though.

pulpcore/tasking/pulpcore_worker.py
bmbouter: Oh pfffff nevermind. It's the line above! Ignore this one please.

pulpcore/tasking/pulpcore_worker.py
bmbouter: I'm ok with leaving it and making all caps.

pulpcore/tasking/pulpcore_worker.py
mdellweg: Without it, all 8 workers triggered worker_cleanup within the same second. That's too much of a coincidence for me.

pulpcore/tasking/pulpcore_worker.py
mdellweg: I put it above, so it's no problem if the cleanup didn't return.

pulpcore/tasking/pulpcore_worker.py
bmbouter: That's not what I expected, but given your testing +1 to this. Ty

bmbouter (approved):
This looks good to me. Thank you!

dralley (approved):

PR #1517, opened 2021-07-23
Update CI files and set python_version to 3.6

daviddavis:
[noissue]


.github/workflows/scripts/before_install.sh
```diff
@@ -53,6 +53,10 @@ echo "legacy_component_name: pulpcore" >> .ci/ansible/vars/main.yaml
 echo "component_name: core" >> .ci/ansible/vars/main.yaml
 echo "component_version: '${COMPONENT_VERSION}'" >> .ci/ansible/vars/main.yaml
 
+
+echo "ci_base: pulp/pulp-ci-centos:python36" >> .ci/ansible/vars/main.yaml
```
mdellweg, line 57: `pulp_container_tag: python36` should do the trick.
`ci_base` was originally meant to be set from the commit message (like Required-PR).

daviddavis: Done. Can you look at https://github.com/pulp/plugin_template/pull/459 ?

template_config.yml
```diff
@@ -1,10 +1,12 @@
...
 # were not present before running plugin-template have been added with their default values.
 
+# generated with plugin_template@2021.04.08-84-g3271e88
+
 additional_repos:
-- branch: master
+- branch: 1.8
   name: pulp_file
-- branch: master
+- branch: 1.4
```
dralley, line 9: Question: is there anything preventing us from re-ordering these keys in the template so that name comes first?

dralley (approved):

template_config.yml
daviddavis: I've tried but it's a losing battle: the next time the plugin-template is run, reverse the ordering again.

dralley: Exactly - can we change it in the template?

daviddavis: I suppose we could. I'd have to look at the code. Feel free to file an issue.

mdellweg (approved):

PR #1519, opened 2021-07-26
Fixed a bug with logging on task cancel

mdellweg:
This was introduced in 794a8799.

[noissue]

^ Before anyone asks: I believe, this does not need a changelog (or issue), because i found that bug in a nightly build and it has never been in a release. So there is no benefit in telling there even _was_ a bug.

### DIFF
--- a/pulpcore/tasking/pulpcore_worker.py
+++ b/pulpcore/tasking/pulpcore_worker.py
@@ -213,7 +213,7 @@ class NewPulpWorker:
                         )
                     ):
                         connection.connection.notifies.clear()
-                        _logger.info(_("Received signal to cancel current task."), task.pk)
+                        _logger.info(_("Received signal to cancel current task %s."), task.pk)
                         os.kill(task_process.pid, signal.SIGUSR1)
                         break
                 if task_process.sentinel in r:


pulpcore/tasking/pulpcore_worker.py
```diff
@@ -304,7 +304,7 @@ def _perform_task(task_pk, task_working_dir_rel_path):
         exc_type, exc, tb = sys.exc_info()
         task.set_failed(exc, tb)
         _logger.info(_("Task %s failed (%s)"), task.pk, exc)
-        _logger.info("\n".join(traceback.format_list(traceback.extract_tb(tb))))
+        _logger.debug("\n".join(traceback.format_list(traceback.extract_tb(tb))))
```
mdellweg, line 307: This change is somewhat unrelated. And i can take it out again.
But stacktraces should not go into "info", right?

goosemania:
I saw the comment about [noissue], however it's a shame to see any changes apart from CI and release stuff to go with no issue.
It complicates things with the current approach when we do not do normal Z releases where all bug fixes go into the latest Z release, but we put into the Z stream only explicitly requested backports.

Are those changes to the logging in the nightlies critical to Katello? - No.
Would it still be better to have them there? - Yes. Which issue I refer to in that case? Automation rightly asks for an issue.
Can we not backport it at all? - We can, however Katello will be on 3.14 for a while because 3.15 requires new Django. Imagine, there is a bug fix for tasking which might not apply cleanly because of those tiny changes to the logging. It would be additional time + manual steps to figure out what is not applied cleanly.

It's not that this PR did something outstandingly bad, it is done every now and then by various folks. It was just a good example to highlight the problem which might not be visible on the surface.

mdellweg:
> I saw the comment about [noissue], however it's a shame to see any changes apart from CI and release stuff to go with no issue.
> It complicates things with the current approach when we do not do normal Z releases where all bug fixes go into the latest Z release, but we put into the Z stream only explicitly requested backports.
> 
> Are those changes to the logging in the nightlies critical to Katello? - No.
> Would it still be better to have them there? - Yes. Which issue I refer to in that case? Automation rightly asks for an issue.
> Can we not backport it at all? - We can, however Katello will be on 3.14 for a while because 3.15 requires new Django. Imagine, there is a bug fix for tasking which might not apply cleanly because of those tiny changes to the logging. It would be additional time + manual steps to figure out what is not applied cleanly.
> 
> It's not that this PR did something outstandingly bad, it is done every now and then by various folks. It was just a good example to highlight the problem which might not be visible on the surface.

Maybe i didn't make the case clear. That bug was introduced last friday. It has never made it into any Release branch. Nor will it ever. And that is the reason why i think having a changelog telling the user there was something fixed we never expect anyone to see broken may not have a value.

That said, i forgot to add that i can make an issue if so desired. My personal belief is however, that it's counter productive.

goosemania:
> > I saw the comment about [noissue], however it's a shame to see any changes apart from CI and release stuff to go with no issue.
> > It complicates things with the current approach when we do not do normal Z releases where all bug fixes go into the latest Z release, but we put into the Z stream only explicitly requested backports.
> > Are those changes to the logging in the nightlies critical to Katello? - No.
> > Would it still be better to have them there? - Yes. Which issue I refer to in that case? Automation rightly asks for an issue.
> > Can we not backport it at all? - We can, however Katello will be on 3.14 for a while because 3.15 requires new Django. Imagine, there is a bug fix for tasking which might not apply cleanly because of those tiny changes to the logging. It would be additional time + manual steps to figure out what is not applied cleanly.
> > It's not that this PR did something outstandingly bad, it is done every now and then by various folks. It was just a good example to highlight the problem which might not be visible on the surface.
> 
> Maybe i didn't make the case clear. That bug was introduced last friday. It has never made it into any Release branch. Nor will it ever. And that is the reason why i think having a changelog telling the user there was something fixed we never expect anyone to see broken may not have a value.
> 
> That said, i forgot to add that i can make an issue if so desired. My personal belief is however, that it's counter productive.

What introduced this bug was also [noissue], that's the point.
I suggest to have a discussion at the pulpcore meeting.

dralley (approved):

pulpcore/tasking/pulpcore_worker.py
mdellweg: With warning being higher than info... I'm about to take this change out of the PR and add to the logging discussion we already have.

mdellweg: I have enough arguments to either increase or decrease the loglevel here. So i'm not touching this line without a properly written ticket anymore.

bmbouter (approved):
Thanks!

Also @mdellweg does this need to go into a 3.14 or 3.7 backport?

mdellweg:
> Thanks!
> 
> Also @mdellweg does this need to go into a 3.14 or 3.7 backport?

No. it's a fixup to a commit that was not part of the cherry-picks.

bmbouter:
Great thanks!

PR #1520, opened 2021-07-26
Apply latest plugin template

bmbouter:
Also configure to use Python 3.6

[noissue]


template_config.yml
```diff
@@ -1,14 +1,19 @@
...
 # were not present before running plugin-template have been added with their default values.
 
+# generated with plugin_template@2021.04.08-84-ga0077c1
+
 additional_plugins:
 - branch: 1.3
   name: pulp_file
 - branch: 1.1
   name: pulp-certguard
+additional_repos: []
```
daviddavis, line 11: I think there was a rename and additional_plugins is now additional_repos. So you'll need to update line 6.

template_config.yml
```diff
@@ -1,14 +1,19 @@
...
 black: true
 check_commit_message: true
+check_gettext: true
 check_manifest: false
+check_openapi_schema: true
```
daviddavis, line 16: I think this will need to be false.

template_config.yml
```diff
@@ -19,16 +24,20 @@ deploy_daily_client_to_rubygems: true
...
+flake8: true
+issue_tracker: redmine
 plugin_app_label: pulpcore
 plugin_camel: Pulpcore
 plugin_camel_short: Pulpcore
 plugin_caps: PULPCORE
 plugin_caps_short: PULPCORE
 plugin_dash: pulpcore
 plugin_dash_short: pulpcore
+plugin_default_branch: main
```
daviddavis, line 36: Change this to master.

template_config.yml
```diff
@@ -1,14 +1,18 @@
...
+# generated with plugin_template@2021.04.08-84-ga0077c1
+
+additional_repos:
 - branch: 1.3
   name: pulp_file
 - branch: 1.1
   name: pulp-certguard
 black: true
 check_commit_message: true
+check_gettext: true
```
daviddavis, line 13: Looks like the gettext check is failing. You can either fix the issues or change this setting to `false`.

daviddavis (approved):

PR #1523, opened 2021-07-27
Workaround breaking drf-access-policy upgrade

mdellweg:
fixes #9160

I'd like to get this in as a 3.16 blocker, as it brings the deprecation cycle promise back that the drf-access-policy upgrade broke.

### DIFF
--- a/CHANGES/plugin_api/9160.deprecation
+++ b/CHANGES/plugin_api/9160.deprecation
@@ -1,0 +1,2 @@
+Deprecate the compatibility layer for access policies. As of pulpcore 3.16, all plugins should
+properly use the "condition" and "condition_expression" fields in the access policy statements.
--- a/pulpcore/app/apps.py
+++ b/pulpcore/app/apps.py
@@ -10,6 +10,7 @@ from django.db.models.signals import post_migrate
 from django.utils.module_loading import module_has_submodule
 
 from pulpcore.exceptions.plugin import MissingPlugin
+from pulpcore.app.loggers import deprecation_logger
 
 VIEWSETS_MODULE_NAME = "viewsets"
 SERIALIZERS_MODULE_NAME = "serializers"
@@ -194,6 +195,27 @@ class PulpAppConfig(PulpPluginAppConfig):
         post_migrate.connect(_delete_anon_user, sender=self, dispatch_uid="delete_anon_identifier")
 
 
+def _drf_access_policy_workaround(viewset_name, access_policy):
+    """Workaround for incompatible drf-access-policy upgrade.
+
+    Simple conditions can be expressions, but not the other way around. This will change the
+    default access_policy in place before it will be saved to the db.
+
+    see https://pulp.plan.io/issues/9160
+    """
+    for stmt in access_policy.get("statements", []):
+        if "condition" in stmt and "condition_expression" not in stmt:
+            if " " in stmt["condition"] or any((" " in cond for cond in stmt["condition"])):
+                stmt["condition_expression"] = stmt.pop("condition")
+                deprecation_logger.warning(
+                    _(
+                        "The access policy for {} is probably using 'condition'"
+                        " erroneously and may stop working with pulpcore==3.16;"
+                        " you may need to use 'condition_expression'."
+                    ).format(viewset_name)
+                )
+
+
 def _populate_access_policies(sender, **kwargs):
     from pulpcore.app.util import get_view_urlpattern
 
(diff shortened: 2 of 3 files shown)


pulpcore/app/apps.py
```diff
@@ -194,6 +195,26 @@ def ready(self):
...
+def _drf_access_policy_workaround(viewset_name, access_policy):
+    """Workaround for incompatible drf-access-policy upgrade
+    Simple conditions can be expressions, but not the other way around.
+    This will change the default access_policy in place before it will be saved to the db.
+
+    see https://pulp.plan.io/issues/9160
+    """
+    for stmt in access_policy.get("statements", []):
+        if "condition" in stmt and "condition_expression" not in stmt:
+            if " " in stmt["condition"] or any((" " in cond for cond in stmt["condition"])):
```
mdellweg, line 207: @ipanova do you think, this is enough heuristic to find "complex" conditions?

ipanova: i think so :bulb:

pulpcore/app/apps.py
```diff
@@ -194,6 +195,26 @@ def ready(self):
...
+                stmt["condition_expression"] = stmt.pop("condition")
+                deprecation_logger.warning(
+                    _(
+                        "The access policy for {} is probably using 'condition'"
+                        " erroneously and may stop working with pulpcore==3.16;"
```
ipanova, line 212: if we plan to remove this compat layer in 3.16 can we then change the language from 'may stop' to 'will stop'?

ipanova: can you please file a ticket to remove this in 3.16?

mdellweg: Oh, the "may" part in this is there on purpose, because i do not trust the heuristic to not have false positives:
_This_ specific access policy "may" still work or "may" not. Even after we removed the workaround.

ipanova (approved):

pulpcore/app/apps.py
mdellweg: https://pulp.plan.io/issues/9163

CHANGES/plugin_api/9160.deprecation
```diff
@@ -0,0 +1,2 @@
+Deprecate the compatibility layer for access policies.
```
bmbouter, line 1: I didn't expect the single sentence per line style here. I know the style debate continues though.

dkliban (approved):

pulpcore/app/apps.py
```diff
@@ -194,6 +195,26 @@ def ready(self):
         post_migrate.connect(_delete_anon_user, sender=self, dispatch_uid="delete_anon_identifier")
 
 
+def _drf_access_policy_workaround(viewset_name, access_policy):
+    """Workaround for incompatible drf-access-policy upgrade
```
bmbouter, line 199: For pep 257 the first line should have a line separator between it and the rest of the docstring.

pulpcore/app/apps.py
```diff
@@ -194,6 +195,26 @@ def ready(self):
...
+    Simple conditions can be expressions, but not the other way around.
+    This will change the default access_policy in place before it will be saved to the db.
```
bmbouter, line 201: I didn't expect the one sentence per line style here.

pulpcore/app/apps.py
bmbouter: I think this is the right code, but it would be nice if it was more readable (to me).

bmbouter:
Will we be asking plugin writers with 3.16 to have both `conditions` and `condition_expression`? Or is it that they just need to have their `statements` section be valid for `drf-access-policy`?

mdellweg:
At any point it's either "condition" or "condition_expression".
We just expect the plugins to so the change on pulpcore release later.

bmbouter:
> At any point it's either "condition" or "condition_expression".
> We just expect the plugins to so the change on pulpcore release later.

So any valid `drf-access-policy` will work for us and in pulpcore 3.16 we'll remove this function is that right?

mdellweg:
> > At any point it's either "condition" or "condition_expression".
> > We just expect the plugins to so the change on pulpcore release later.
> 
> So any valid `drf-access-policy` will work for us and in pulpcore 3.16 we'll remove this function is that right?

Yes, 3.14 will work with the old style; 3.15 will (heuristically) adjust old ones to new ones and understand new ones; 3.16 will only accept new ones.

bmbouter (approved):
This looks great. Thanks for switching the style and docstring fix.

PR #1524, opened 2021-07-27
Adds deprecation of old tasking system

bmbouter:
This commit deprecates the old style tasking system in preparation for
it to be removed with pulpcore==3.16.

closes #9159

### DIFF
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -302,6 +302,12 @@ def dispatch(func, resources, args=None, kwargs=None, task_group=None):
             cursor.execute("NOTIFY pulp_worker_wakeup")
         return task
     else:
+        deprecation_logger.warning(
+            _(
+                "You are using the traditional tasking system which will be removed in pulpcore "
+                "3.16 along with the `USE_NEW_WORKER_TYPE` setting."
+            )
+        )
         RQ_job_id = _enqueue_with_reservation(
             func, resources=resources, args=args, kwargs=kwargs, task_group=task_group
         )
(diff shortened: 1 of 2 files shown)


dralley (approved):

daviddavis:
Not sure if this is possible or worthwhile but can we show a deprecation warning if someone uses the old tasking system?

bmbouter:
I can add in a deprecation warning also. I'll revise.

bmbouter:
@daviddavis I added the deprecation warning. I opted to make it warn with each task because having 3.16 force you into the new one could be unexpected if you went out of your way to opt-in to the old-style tasking system.

daviddavis (approved):

pulpcore/tasking/tasks.py
```diff
@@ -302,6 +302,12 @@ def dispatch(func, resources, args=None, kwargs=None, task_group=None):
             cursor.execute("NOTIFY pulp_worker_wakeup")
         return task
     else:
+        deprecation_logger.warning(
+            _(
+                "You are using the traditional tasking system which will be removed in pulpcore "
+                "3.16 along with the `USE_NEW_WORKER_TYPE` setting."
+            )
+        )
```
mdellweg, line 310: That will be quite verbose. Maybe it's better in the worker startup code?

bmbouter: I'm happy to put it either place, but the concern is that folks won't really see it if it's just 1 log per process at startup. I'll put it anywhere all the reviewers agree to.

mdellweg: O i haven't seen you commented on that already.

bmbouter: @mdellweg did you want me to make a change or merge it?

daviddavis: I'm happy with either place but the more verbose/annoying option seems better to me since people on 3.15 should really be using the new tasking system.

mdellweg (approved):

PR #1525, opened 2021-07-27
rbac roles PoC

bmbouter:


bmbouter:
At least one thing this doesn't handle (and I don't see how it could) is a situation where a Role gets revised to add in a permission for a new model type and there were already object-level permissions existing for that Role.

## In a version of pulp say a plugin ships

```
SomeRole(Role):
    permissions = ["model1_perm_a", "model1_perm_b"]
```

Then while the application runs some object-level role assignments happen on `obj1` and `obj2` where both of those objects are of type model1.

```
assign_role(SomeRole, example_user, obj1)
assign_role(SomeRole, example_group, obj2)
```

## A new version of the plugin comes out that revises `SomeRole` to introduce `model2_another_perm`

`model2` is the new object type, at least new to the role definition. Say `obj3` is of type model2 and is "related" to `obj1` and `obj2` in terms of intended application use. (This part is a little hand wavy). So `SomeRole` would look like:

```
SomeRole(Role):
    permissions = ["model1_perm_a", "model1_perm_b", "model2_another_perm"]
```

After the upgrade and Pulp starts, the [code that handles additions for obj-level perms](https://github.com/pulp/pulpcore/pull/1525/files#diff-1a15d2d8642c07f47172cac10d8c48b6a9229b6081510f66f9e32ba95659bf5bR300-R319) wouldn't be able to correctly assign `model2_another_perm` to `obj3` because there is nothing already assigned for model2 and that role for it "go off of".

goosemania:
This case potentially can be covered by the plugin, e.g. there is a known relation between model1 and model2.

bmbouter:
> This case potentially can be covered by the plugin, e.g. there is a known relation between model1 and model2.

I agree. My take is roughly that it won't happen that often and if it does only a plugin would know about `obj3` being related to `obj1 and `obj2` so only they could really fix this. They would fix it by shipping a migration in these specific cases.

goosemania:
> > This case potentially can be covered by the plugin, e.g. there is a known relation between model1 and model2.
> 
> I agree. My take is roughly that it won't happen that often and if it does only a plugin would know about `obj3` being related to `obj1` and `obj2` so only they could really fix this. They would fix it by shipping a migration in these specific cases.

Exactly, and I think it's totally fine. Actually the only way. And it will likely be true regardless of the approach to the roles implementation.

bmbouter:
FYI I've been running this with `pulpcore-manager test pulpcore.tests.unit.test_roles`. What's strange to me is that maybe 1 out of 10 runs gives me a failure. Also I've tried running it with `pulpcore-manager test pulpcore.tests.unit.test_roles --reverse` and it's roughly the same. It's strange.

ipanova:
> > > This case potentially can be covered by the plugin, e.g. there is a known relation between model1 and model2.
> > 
> > 
> > I agree. My take is roughly that it won't happen that often and if it does only a plugin would know about `obj3` being related to `obj1` and `obj2` so only they could really fix this. They would fix it by shipping a migration in these specific cases.
> 
> Exactly, and I think it's totally fine. Actually the only way. And it will likely be true regardless of the approach to the roles implementation.

I agree. This would a case managed by the plugin

pulpcore/app/apps.py
```diff
@@ -218,6 +241,149 @@ def _populate_access_policies(sender, **kwargs):
...
+            added_role_permissions = set(role.permissions) - set(role_history.permissions)
+            removed_role_permissions = set(role_history.permissions) - set(role.permissions)
+
+            if added_role_permissions:
+                existing_perms_qs = {}
+                existing_perms = {}
+                for key, perms_group in _get_perms_grouped_by_model(role_history.permissions):
+                    list_of_perms = list(perms_group)
+                    existing_perms[key] = list_of_perms
+                    existing_perms_qs[key] = _get_as_permissions_qs(list_of_perms)
```
bmbouter, line 275: This part is duplicated between here and below. One of the improvements I was thinking of was deduplicated by moving it up above `if added_role_permissions` and reusing the data between both the `if added_role_permissions` and `if removed_role_permissions`.

bmbouter:
@mdellweg I believe your continuation of this work supersedes this PR? Whenever that posts, I'm +1 to close this.

PR #1526, opened 2021-07-28
Add shared_resources to task dispatch

mdellweg:
This allows tasks that use overlapping resources read only to run
concurrently.

fixes #9148


mdellweg:
We probably do not want to introduce this in 3.15.

pulpcore/tasking/tasks.py
```diff
@@ -256,7 +256,7 @@ def _enqueue_with_reservation(
     return Job(id=inner_task_id, connection=redis_conn)
 
 
-def dispatch(func, resources, args=None, kwargs=None, task_group=None):
+def dispatch(func, resources, args=None, kwargs=None, task_group=None, shared_resources=None):
```
dralley, line 259: I think if we should rename `resources` to `exclusive_resources`

dralley:
@mdellweg Is this in draft because it's not ready or in draft because we don't want to merge it yet?

If the latter - what do you think about merging it under an off-by-default feature flag.  We could then remove the flag and make it a default in 3.16, or whenever we feel ready to enable it.  "shared" resources could be treated as exclusive ones if it is turned off.

I've started appreciating this approach because it allows users (both upstream and downstream) the ability to test features earlier and the flexibility to work around issues in brand new features.

mdellweg:
Off by default sounds great!

pulpcore/tasking/tasks.py
dralley: Indirectly, yes, but as it's an arg rather than a kwarg, every use that I have been able to find uses it positionally rather than as a kwarg.

But I'd be fine with filing an issue to do it after a deprecation cycle.

dralley:
I would be fine either way, I just thought that since there was discussion about not putting it into 3.15 that perhaps it wasn't ready for that, yet.

mdellweg:
I'd be happy to put it in 3.15, but i understand, there is a risk to it.
And let me state i would suggest either "off-by-default" in 3.15, or "on-by-default" for an early 3.16 merge.

dralley:
"Off for 3.15, on for 3.16" sounds fine to me.

CHANGES/9148.feature
```diff
@@ -0,0 +1,2 @@
+A new setting ``ALLOW_SHARED_RESOURCES`` was added to allow plugins to use specific resources
+concurrently. It defaults to ``False``.
```
dralley, line 2: I suggest being more explicit that the setting is temporary to allow a preview of the behavior, that the default behavior will change in the next release, and that the setting is likely to go away either in that same release or the next one.

docs/configuration/settings.rst
```diff
@@ -368,6 +368,18 @@ USE_NEW_WORKER_TYPE
...
 
 
+.. _allow_shared_resources:
+
+ALLOW_SHARED_RESOURCES
+^^^^^^^^^^^^^^^^^^^^^^
+
+    This option allows that tasks can have a shared (read only) simultaneous access to some
+    resources. It currently defaults to ``False``, but may be switched on to improve tasking
+    throughput.
```
dralley, line 378: Same

CHANGES/9148.feature
```diff
@@ -0,0 +1,3 @@
+A new setting ``ALLOW_SHARED_RESOURCES`` was temporarily added to allow plugins to use specific
```
ipanova, line 1: s/to use specific resources  concurrently/ to use specific resources concurrently during task execution.
From this changelong it is not very clear what resources we're talking about. I also was thinking maybe we should rename the setting to something close to `ALLOW_TASK_SHARED_RESOURCES`

CHANGES/9148.feature
```diff
@@ -0,0 +1,3 @@
...
+resources concurrently. It defaults to ``False``. It will switch to ``True`` with 3.16 and will
+eventually be removed again.
```
ipanova, line 3: s/removed again/removed.

docs/configuration/settings.rst
```diff
@@ -368,6 +368,21 @@ USE_NEW_WORKER_TYPE
        6. Start the api-workers ``systemctl start pulpcore-api``
 
 
+.. _allow_shared_resources:
+
+ALLOW_SHARED_RESOURCES
+^^^^^^^^^^^^^^^^^^^^^^
+
+    This option allows that tasks can have a shared (read only) simultaneous access to some
```
ipanova, line 376: s/allows that/allows tasks to have/

docs/configuration/settings.rst
ipanova: s/It currently defaults to ``False``, but may be switched on to improve tasking throughput./It defaults to ``False`` but if changed to ``True`` it will improve tasking throughput.

docs/configuration/settings.rst
```diff
@@ -368,6 +368,21 @@ USE_NEW_WORKER_TYPE
...
+    resources. It currently defaults to ``False``, but may be switched on to improve tasking
+    throughput.
+
+    .. note:: This option will only take effect when using the queueless worker type.
+
+    .. note:: This option is meant to be temporary. It will default to ``True`` in 3.16 and be
+              removed again in 3.17.
```
ipanova, line 383: s/removed again/removed

docs/configuration/settings.rst
```diff
@@ -368,6 +368,21 @@ USE_NEW_WORKER_TYPE
...
+ALLOW_SHARED_RESOURCES
+^^^^^^^^^^^^^^^^^^^^^^
+
+    This option allows that tasks can have a shared (read only) simultaneous access to some
+    resources. It currently defaults to ``False``, but may be switched on to improve tasking
+    throughput.
+
+    .. note:: This option will only take effect when using the queueless worker type.
+
+    .. note:: This option is meant to be temporary. It will default to ``True`` in 3.16 and be
```
ipanova, line 382: As a tech preview, this option is meant to be temporary....

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -147,14 +158,26 @@ def iter_tasks(self):
                             # without considering this tasks resources
                             # as we just released them
                             continue
-                    if task.state == TASK_STATES.WAITING and not any(
-                        resource in taken_resources for resource in reserved_resources_record
+                    if (
```
ipanova, line 161: this if condition is very long and difficult to read, can you break it down and assign some parts to variables?

pulpcore/tasking/tasks.py
```diff
@@ -308,6 +314,10 @@ def dispatch(func, resources, args=None, kwargs=None, task_group=None):
                 "3.16 along with the `USE_NEW_WORKER_TYPE` setting."
             )
         )
+        # There is only exclusive use in the legacy tasking system
+        if shared_resources:
+            resources = resources[:]
+            resources.extend(shared_resources)
```
mdellweg, line 320: Maybe that is more readable:
```suggestion
            resources = resources + shared_resources
```

CHANGES/9148.feature
mdellweg: I'll take `ALLOW_SHARED_TASK_RESOURCES`

pulpcore/tasking/pulpcore_worker.py
mdellweg: It is using lazy evaluation, so the later `any` don't consume cpu unless needed.

ipanova: thanks for adding comments

ipanova (approved):
Looks good! please file an issue if there is not one already to remove the `ALLOW_SHARED_TASK_RESOURCES` in 3.17

pulpcore/tasking/tasks.py
mdellweg: What would be the desired final configuration for these args? From there we should be able to come up with a deprecation plan.
My suggestion: Both `[exclusive_]resources` and `shared_resources` should be used as a kwarg.
Using "/" to mark parameters pos-only is available from Python 3.8, but it does not do kw-only if I read the docs correctly.

dralley: > My suggestion: Both [exclusive_]resources and shared_resources should be used as a kwarg.

That's reasonable.  I don't think there is a way to enforce keyword-only but we can roll out a change of conventions and it should continue working.

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -147,14 +158,30 @@ def iter_tasks(self):
                             # without considering this tasks resources
                             # as we just released them
                             continue
-                    if task.state == TASK_STATES.WAITING and not any(
-                        resource in taken_resources for resource in reserved_resources_record
+                    # This statement is using lazy evaluation
+                    if (
+                        task.state == TASK_STATES.WAITING
```
dralley, line 163: Following what Ina said, I'm slightly partial towards just pre-merging the resource lists in the "disabled" case and leaving the actual logic alone.  If you prefer this way however, I can live with it.

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -130,11 +130,22 @@ def iter_tasks(self):
...
+                exclusive_resources = [
+                    resource
+                    for resource in reserved_resources_record
+                    if not resource.startswith("shared:")
+                ]
+                shared_resources = [
+                    resource[7:]
+                    for resource in reserved_resources_record
+                    if resource.startswith("shared:") and resource[7:] not in exclusive_resources
+                ]
```
dralley, line 148: Could we use `set`s consistently?

mdellweg: We could, but I don't see the benefit. The only thing we use these for is looping over them.
If i'm not mistaken, creating `set`s instead of `list`s would change from `O(n)` to `O(n log n)` without any gain elsewhere.

pulpcore/tasking/pulpcore_worker.py
mdellweg: I made it this way, so that removing the old behaviour is just deleting lines 171-176.
But maybe having the guard around the old code path is introducing less risk. So I will change that.

pulpcore/tasking/tasks.py
mdellweg: I _will_ make this a separate PR. And maybe a separate ticket, as i think technically this commit is still backportable to 3.14 in this state if considered fixing a regression.

dralley (approved):

dralley:
:+1:

PR #1528, opened 2021-07-28
Orphan clean up is no longer a blocking call.

ipanova:
[noissue]

### DIFF
--- a/CHANGES/8824.feature
+++ b/CHANGES/8824.feature
@@ -1,0 +1,3 @@
+Changed orphan cleanup task to be a non-blocking task that can be run at any time. Added a
+``ORPHAN_PROTECTION_TIME`` setting that can be configured for how long orphan Content and
+Artifacts are kept before becoming candidates for deletion by the orphan cleanup task.
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -387,3 +387,14 @@ DJANGO_GUID
 
     For more information on how to configure the ``DJANGO_GUID`` setting, see the `django-guid
     settings documentation <https://django-guid.readthedocs.io/en/latest/settings.html>`_.
+
+
+.. _orphan-protection-time:
+
+ORPHAN_PROTECTION_TIME
+^^^^^^^^^^^^^^^^^^^^^^
+
+    The time specified in minutes for how long Pulp will hold orphan Content and Artifacts before
+    they become candidates for deletion by an orphan cleanup task. This should ideally be longer
+    than your longest running task otherwise any content created during that task could be cleaned
+    up before the task finishes. Default is 1440 minutes (24 hours).
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -3,6 +3,7 @@ Content related Django models.
 """
 from gettext import gettext as _
 
+import datetime
 import json
 import tempfile
 import shutil
@@ -17,6 +18,7 @@ from django.core import validators
 from django.core.files.storage import default_storage
 from django.db import IntegrityError, models, transaction
 from django.forms.models import model_to_dict
+from django.utils.timezone import now
 from django_lifecycle import BEFORE_UPDATE, BEFORE_SAVE, hook
 
 from pulpcore.constants import ALL_KNOWN_CONTENT_CHECKSUMS
@@ -139,6 +141,13 @@ class HandleTempFilesMixin:
         self.file.delete(save=False)
 
 
+class ArtifactManager(BulkCreateManager):
+    def orphaned(self, orphan_protection_time):
+        """Returns set of orphaned artifacts that are ready to be cleaned up."""
+        expiration = now() - datetime.timedelta(minutes=orphan_protection_time)
+        return self.filter(content_memberships__isnull=True, timestamp_of_interest__lt=expiration)
+
+
 class Artifact(HandleTempFilesMixin, BaseModel):
     """
     A file associated with a piece of content.
@@ -183,7 +192,7 @@ class Artifact(HandleTempFilesMixin, BaseModel):
     sha512 = models.CharField(max_length=128, null=True, unique=True, db_index=True)
     timestamp_of_interest = models.DateTimeField(auto_now=True)
 
-    objects = BulkCreateManager()
+    objects = ArtifactManager()
 
     # All available digest fields ordered by algorithm strength.
     DIGEST_FIELDS = _DIGEST_FIELDS
@@ -426,6 +435,19 @@ class PulpTemporaryFile(HandleTempFilesMixin, BaseModel):
         return PulpTemporaryFile(file=file)
 
 
+class ContentManager(BulkCreateManager):
+    def orphaned(self, orphan_protection_time, content_pks=None):
+        """Returns set of orphaned content that is ready to be cleaned up."""
+        expiration = now() - datetime.timedelta(minutes=orphan_protection_time)
+        if content_pks:
+            return self.filter(
+                version_memberships__isnull=True,
+                timestamp_of_interest__lt=expiration,
+                pk__in=content_pks,
+            )
+        return self.filter(version_memberships__isnull=True, timestamp_of_interest__lt=expiration)
+
+
 class Content(MasterModel, QueryMixin):
     """
     A piece of managed content.
@@ -448,7 +470,7 @@ class Content(MasterModel, QueryMixin):
     _artifacts = models.ManyToManyField(Artifact, through="ContentArtifact")
     timestamp_of_interest = models.DateTimeField(auto_now=True)
 
-    objects = BulkCreateManager()
+    objects = ContentManager()
 
     class Meta:
         verbose_name_plural = "content"
--- a/pulpcore/app/tasks/orphan.py
+++ b/pulpcore/app/tasks/orphan.py
@@ -1,5 +1,7 @@
 import gc
 
+from django.conf import settings
+
 from pulpcore.app.models import (
     Artifact,
     Content,
@@ -29,7 +31,7 @@ def queryset_iterator(qs, batchsize=2000, gc_collect=True):
             gc.collect()
 
 
-def orphan_cleanup(content_pks=None):
+def orphan_cleanup(content_pks=None, orphan_protection_time=settings.ORPHAN_PROTECTION_TIME):
     """
     Delete all orphan Content and Artifact records.
     Go through orphan Content multiple times to remove content from subrepos.
@@ -48,12 +50,9 @@ def orphan_cleanup(content_pks=None):
     )
 
     while True:
-        content = Content.objects.filter(version_memberships__isnull=True).exclude(
+        content = Content.objects.orphaned(orphan_protection_time, content_pks).exclude(
             pulp_type=PublishedMetadata.get_pulp_type()
         )
-        if content_pks:
-            content = content.filter(pk__in=content_pks)
-
         content_count = content.count()
         if not content_count:
             break
@@ -70,7 +69,7 @@ def orphan_cleanup(content_pks=None):
     progress_bar.save()
 
     # delete the artifacts that don't belong to any content
-    artifacts = Artifact.objects.filter(content_memberships__isnull=True)
+    artifacts = Artifact.objects.orphaned(orphan_protection_time)
 
     progress_bar = ProgressReport(
         message="Clean up orphan Artifacts",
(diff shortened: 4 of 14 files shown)


ipanova:
@mdellweg @daviddavis  is there anything from your perspective that needs to be handled?

daviddavis:
I think this code still needs to consider `PERSERVED_ORPHAN_CONTENT`.

For reference, here is my old commit:

https://github.com/daviddavis/pulpcore/commit/1e3751fa8c2bd7cd448e7ab7933a9d2a07021379

This was from before [#7659](https://pulp.plan.io/issues/7659) was split up or the new tasking system was merged.

mdellweg:
Is there any difference between this and https://pulp.plan.io/issues/8824 ?

As far as the queueless tasking system is concerned, this looks right. I initially made it a single commit to be easily reverted:
https://github.com/pulp/pulpcore/commit/40f3c713155c0135e832d77193f9e761601e12e6

For the legacy tasking system i think this is also correct.

dralley:
It looks like the intention was for it to be a subtask of the epic.  Given the small amount of work left we could probably just close #8824 outright and treat this PR as closing #7659 but it doesn't really matter too much.

daviddavis:
#8823 has already been completed/shipped. And I am still working on #8960 so I'd ask that we not close out the epic.

ipanova:
> I think this code still needs to consider `PERSERVED_ORPHAN_CONTENT`.
> 
> For reference, here is my old commit:
> 
> [daviddavis@1e3751f](https://github.com/daviddavis/pulpcore/commit/1e3751fa8c2bd7cd448e7ab7933a9d2a07021379)
> 
> This was from before [#7659](https://pulp.plan.io/issues/7659) was split up or the new tasking system was merged.

for some reason i was thinking that this part was already done, ha wishful thinking! ty for sharing your code and incorporate it in this PR.
I agree with @daviddavis let's not close the epic and i will reference this pr to this #8824 issue.

ipanova:
@daviddavis you did most of the job, I mostly took the code from your commit and added minor things.

ipanova:
tests are failing, need to fix them.

mdellweg:
> tests are failing, need to fix them.

Is that, because tests rely on the old orphan cleanup behaviour?

daviddavis:
I was originally imagining that we'd probably have to configure PRESERVED_ORPHAN_CONTENT [here](https://git.io/JBV1W) to 0 for the test environment.

However, I now realize that this is probably not going to work for when we run these tests in our dev environments.

mdellweg:
And i had imagined we have a way to pass the time delay as an optional parameter to the orphan call.

OTOH, we might get away with setting it to 0 in dev envs. Though i also prefer not to.

docs/configuration/settings.rst
```diff
@@ -387,3 +387,14 @@ DJANGO_GUID
 
     For more information on how to configure the ``DJANGO_GUID`` setting, see the `django-guid
     settings documentation <https://django-guid.readthedocs.io/en/latest/settings.html>`_.
+
+
+.. _preserved-orphan-content:
+
+PRESERVED_ORPHAN_CONTENT
```
ipanova, line 394: How should the deprecated endpoint behave? Should it take into account this or rather cleanup the orphans and disregard it? basically have 'PRESERVED_ORPHAN_CONTENT' set to 0?
@dralley @daviddavis @mdellweg thoughts?

daviddavis:
> And i had imagined we have a way to pass the time delay as an optional parameter to the orphan call.

I thought this over and was wondering about the rbac implications. For example, a system admin might want to give users the ability to run orphan cleanup but maybe not allow them to set it to 0. Not sure if it's a valid use case.

docs/configuration/settings.rst
daviddavis: I don't think we should let or encourage users to run orphan cleanup with `PRESERVED_ORPHAN_CONTENT` set to 0. I think really bad things could happen.

ipanova: I could add validation for this setting at pulp start and raise an error for 0

daviddavis: I think it's ok if a user explicitly sets it to 0. I just don't think we should provide an endpoint that does so automatically.

pulpcore/tests/functional/api/using_plugin/test_orphans.py
```diff
@@ -50,6 +50,9 @@ def setUpClass(cls):
         cls.storage = utils.get_pulp_setting(cls.cli_client, "DEFAULT_FILE_STORAGE")
         cls.media_root = utils.get_pulp_setting(cls.cli_client, "MEDIA_ROOT")
```
ipanova, line 51: I am not sure how to proceed with this test for the deprecated endpoint. I could manually update the timestamp of the content and artifact so it's older than 24hous or maybe we can remove this test and focus on the new endpoint?
what would be best?

daviddavis: What if you set PRESERVED_ORPHAN_CONTENT to 0 in the CI?

CHANGES/8824.feature
```diff
@@ -0,0 +1,3 @@
+Changed orphan cleanup task to be a non-blocking task that can be run at anytime. Added a
```
dralley, line 1: s/anytime/any time

docs/configuration/settings.rst
```diff
@@ -387,3 +387,14 @@ DJANGO_GUID
...
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+    The time in minutes for how long Pulp will hold orphan Content and Artifacts before they become
+    candidates for deletion by an orphan cleanup task. This should ideally be longer than your
+    longest running task otherwise any content created during that task could be cleaned up before
+    the task finishes. Default is 1440 minutes (24 hours).
```
dralley, line 400: Does the use case require minute granularity or could we definite it in terms of hours?

docs/configuration/settings.rst
```diff
@@ -387,3 +387,14 @@ DJANGO_GUID
...
     For more information on how to configure the ``DJANGO_GUID`` setting, see the `django-guid
     settings documentation <https://django-guid.readthedocs.io/en/latest/settings.html>`_.
+
+
+.. _preserved-orphan-content:
+
+PRESERVED_ORPHAN_CONTENT
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+    The time in minutes for how long Pulp will hold orphan Content and Artifacts before they become
```
dralley, line 397: This would be slightly more correct:

"The time specifying for how long (in minutes) Pulp should hold orphan Content and Artifacts for before they become eligible for deletion by an orphan cleanup task."

docs/configuration/settings.rst
dralley: I agree, the endpoint should obey the new setting.  

The name `PRESERVED_ORPHAN_CONTENT` seems like it doesn't capture that it's a timeout though, it sounds more like the name of a collection of content.  Perhaps `{PRESERVE | PROTECT}_ORPHAN_CONTENT_TTL`?

pulpcore/app/models/content.py
```diff
@@ -426,6 +439,23 @@ def init_and_validate(file, expected_digests=None, expected_size=None):
         return PulpTemporaryFile(file=file)
 
 
+class ContentManager(BulkCreateManager):
+    def orphaned(self, content_pks=None, preserve_content=None):
```
dralley, line 443: Likewise, I would rename this kwarg to `ttl`

pulpcore/app/models/content.py
```diff
@@ -139,6 +141,17 @@ def delete(self, *args, **kwargs):
         self.file.delete(save=False)
 
 
+class ArtifactManager(BulkCreateManager):
+    def orphaned(self, preserve_content=None):
```
dralley, line 145: Likewise, I would rename this kwarg to `ttl`

pulpcore/app/models/content.py
```diff
@@ -139,6 +141,17 @@ def delete(self, *args, **kwargs):
...
+        """Returns set of orphaned artifacts that are ready to be cleaned up."""
+        if preserve_content is not None:
+            preserve = preserve_content
+        else:
+            preserve = settings.PRESERVED_ORPHAN_CONTENT
+        expiration = now() - datetime.timedelta(minutes=preserve)
```
dralley, line 151: One motivation is that everywhere we actually use this value we seem to call it an `expiration` time.

pulpcore/app/tasks/orphan.py
```diff
@@ -29,7 +29,7 @@ def queryset_iterator(qs, batchsize=2000, gc_collect=True):
             gc.collect()
 
 
-def orphan_cleanup(content_pks=None):
+def orphan_cleanup(content_pks=None, preserve_content=None):
```
dralley, line 32: Same here.

docs/configuration/settings.rst
ipanova: I like this suggestion. What do you think about `RETAIN_ORPHAN_CONTENT`

dralley: It still doesn't communicate the idea of there being a time threshold very well, it sounds like it would be an on/off value.

ipanova: I disagree, however I will update the setting following your suggestions because I don't have a strong opinion on this. We already have in other Pulp places where we use concept of `retention`, and  it would be great to keep using it and not introduce yet another synonym. I would not say `retain` is taken as an on/off value - think of retain repo versions, retain packages. Yes, you can sort of disable it by setting to 0 ( for packages) and null ( for repo versions ) but with the same success you can do so by setting `{PRESERVE | PROTECT}_ORPHAN_CONTENT_TTL or ORPHAN_CONTENT_PROTECTION_{TTL | EXPIRATION}` to 0. Also, arguably `expiration` is a good idea to use in this context - it implies some automatic cleanup/deletion which in our case does not happen unless a manual orphan cleanup is triggered.

goosemania: I agree that it should reflect the interval.
I would avoid the word `content` maybe because we clean up both content and artifacts. 
I like `_TTL` because it's short but it might imply that orphans are removed automagically after that time which is not true, matybe just `_TIME` is better?
`ORPHAN_PROTECTION_TIME`?

goosemania: Also maybe @melcorr have a preference reflecting a user PoV? :)

melcorr: I need to think about this for a bit !! :)

dralley: +1 to avoiding `CONTENT` altogether.  I like @goosemania's suggestion of `ORPHAN_PROECTION_{ TIME | PERIOD | | TTL }`

mdellweg: Another suggestion (to be sure we didn't miss a chance): `ORPHAN_CLEANUP_DELAY`

melcorr: > ORPHAN_PROTECTION_{ TIME | PERIOD | | TTL }

I can't make up my mind between the three, but I like this label the most. 
`TIME` would be good, I think.

docs/configuration/settings.rst
ipanova: i think minutes gives more flexibility for cases when users would want to keep orphans for x minutes and not hours

ipanova:
I've updated the settings and names based on the suggestions and feedback.

dralley (approved):

dralley:
The code looks good - and the test failures have been happening on a few other PRs.

template_config.yml
```diff
@@ -42,6 +42,7 @@ pulp_settings:
   - /tmp
   allowed_import_paths:
   - /tmp
+  orphan_protection_time: 0
```
mdellweg, line 45: Are we going to remove this, once we patched the orphan cleanup util in pulp-smash?

mdellweg (approved):

template_config.yml
ipanova: we can't remove this because of the deprecated orphan cleanup endpoint

mdellweg: Once all tests have been converted to use the new one, i think we should be.

mdellweg: I'm not saying that we want to remove the setting from pulp, but just not set it for the tests.

ipanova: test will fail until this is being run https://github.com/pulp/pulpcore/blob/71a8cb18450c02c4799e7831375893a3b34093d3/pulpcore/tests/functional/api/using_plugin/test_orphans.py#L96

mdellweg: Calling the new endpoint with the explicit "orphan_protection_time=0" should do the trick, no?

Anyway, this is not a blocker for this PR.

ipanova: the new endpoint is not the problem. Old endpoint is `self.orphans_api.delete() ` this will not remove any artifacts because of the global setting. That's why need need setting for tests untill we keep running tests for *old* endpoint. Does it make sense?

mdellweg: Yes, that was my original question. Once we use the new endpoint in all tests (ideally though pulp-smash) we can remove setting this. Thank you!

PR #1531, opened 2021-07-28
`AccessPolicyFromDB` is now used by default

bmbouter:
The `AccessPolicyFromDB` object is now declared by default in the
settings file, and it provides a fallback behavior to the `IsAdmin`
functionality that was there before.

closes #8974

### DIFF
--- a/docs/plugins/plugin-writer/concepts/rbac/access_policy.rst
+++ b/docs/plugins/plugin-writer/concepts/rbac/access_policy.rst
@@ -205,16 +205,21 @@ generally causes them to pass any permission check even without explicit permiss
 Viewset Enforcement
 -------------------
 
-Protecting a viewset with your saved AccessPolicy is done by declaring a ``permission_classes``
-class attribute on your ViewSet that points to ``pulpcore.plugin.access_policy.AccessPolicyFromDB``.
+Pulp configures the ``DEFAULT_PERMISSION_CLASSES`` in the settings file to use
+``pulpcore.plugin.access_policy.AccessPolicyFromDB`` by default. This ensures that by defining a
+``DEFAULT_ACCESS_POLICY`` on your Viewset, Pulp will automatically save it to the database at
+migration-time, and your Viewset will be protected without additional effort.
 
-For example, here is the FileRemoteViewSet which enables authorization enforcement as follows:
+This strategy allows users to completely customize or disable the DRF Permission checks Pulp uses
+like any typical DRF project would.
+
+Also like a typical DRF project, individual Viewsets or views can also be customized to use a
+different Permission check by declaring the ``permission_classes`` check. For example, here is the
+``StatusView`` which disables permission checks entirely as follows:
 
 .. code-block:: python
 
-    from pulpcore.plugin.access_policy import AccessPolicyFromDB
-
-    class FileRemoteViewSet(NamedModelViewSet):
+    class StatusView(APIView):
         ...
-        permission_classes = (AccessPolicyFromDB,)
+        permission_classes = tuple()
         ...
--- a/docs/plugins/plugin-writer/concepts/rbac/overview.rst
+++ b/docs/plugins/plugin-writer/concepts/rbac/overview.rst
@@ -45,15 +45,16 @@ To add authorization for a given resource, e.g. ``FileRemote``, you'll need to:
 2. Define the default permissions created for new objects using the ``permissions_assignment``
    attribute of the new Access Policy for the resource. See the
    :ref:`adding_automatic_permissions_for_new_objects` documentation for more information on that.
-3. Ship that Access Policy as the class attribute ``DEFAULT_ACCESS_POLICY`` of a ``NamedModelViewSet``.
-   This will contain both the ``statements`` and ``permissions_assignment`` attributes. See the
-   :ref:`shipping_default_access_policy` documentation for more information on this.
+3. Ship that Access Policy as the class attribute ``DEFAULT_ACCESS_POLICY`` of a
+   ``NamedModelViewSet``. This will contain both the ``statements`` and ``permissions_assignment``
+   attributes. See the :ref:`shipping_default_access_policy` documentation for more information on
+   this.
 
 **Enforce the Policy:**
 
-1. Define the ``permission_classes`` attribute on your Viewset referring to your subclass of
-   ``pulpcore.plugin.access_policy.AccessPolicyFromDB``. See the :ref:`viewset_enforcement` docs for
-   more information on this.
+1. ``pulpcore.plugin.access_policy.AccessPolicyFromDB`` is configured as the default permission
+   class, so by specifying a ``DEFAULT_ACCESS_POLICY`` it will automatically be enforced. See the
+   :ref:`viewset_enforcement` docs for more information on this.
 
 **Add QuerySet Scoping:**
 
--- a/pulpcore/app/access_policy.py
+++ b/pulpcore/app/access_policy.py
@@ -23,6 +23,10 @@ class AccessPolicyFromDB(AccessPolicy):
 
             AccessPolicyModel.objects.get(viewset_name=get_view_urlpattern(view))
 
+        If a matching `pulpcore.plugin.models.AccessPolicy` cannot be found, a default behavior of
+        allowing only admin users to perform any operation is used. This fallback allows the Pulp
+        RBAC implementation to be turned on endpoint-by-endpoint with less effort.
+
         Args:
             request (rest_framework.request.Request): The request being checked for authorization.
             view (subclass rest_framework.viewsets.GenericViewSet): The view name being requested.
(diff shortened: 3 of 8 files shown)


pulpcore/app/access_policy.py
```diff
@@ -23,13 +23,23 @@ def get_policy_statements(self, request, view):
...
             The access policy statements in drf-access-policy policy structure.
         """
-        access_policy_obj = AccessPolicyModel.objects.get(viewset_name=get_view_urlpattern(view))
-
-        return access_policy_obj.statements
+        viewset_name = get_view_urlpattern(view)
+        try:
+            access_policy_obj = AccessPolicyModel.objects.get(viewset_name=viewset_name)
+        except AccessPolicyModel.DoesNotExist:
+            allow_authenticated = [{"action": "*", "principal": "authenticated", "effect": "allow"}]
```
gerrod3, line 42: The `principal` should be `admin` to keep the same fallback behavior as currently before. https://rsinger86.github.io/drf-access-policy/statement_elements.html

gerrod3 (commented):
I think there is one more spot here mentioning setting permission_classes: https://github.com/pulp/pulpcore/blob/354383883032277e7a1f7dc7ddf2dc0a5bc40fad/docs/plugins/plugin-writer/concepts/rbac/overview.rst

pulpcore/app/access_policy.py
```diff
@@ -23,13 +23,23 @@ def get_policy_statements(self, request, view):
 
             AccessPolicyModel.objects.get(viewset_name=get_view_urlpattern(view))
 
+        If a matching `pulpcore.plugin.models.AccessPolicy` cannot be found, a default behavior of
+        allowing any authenticated user to perform any operation is assumed. This fallback allows
+        the Pulp RBAC implementation to be turned on endpoint by endpoint with less effort. Once
+        RBAC is fully implemented on all endpoint, it's expected that this fallback will be removed.
```
mdellweg, line 29:
```suggestion
        RBAC is fully implemented on all endpoints, it's expected that this fallback will be removed.
```

pulpcore/app/access_policy.py
mdellweg: I agree. Also this is probably too late to make this class turn into / look like `IsAdmin`, right?

We could make `permission_classes` on the base viewset a property that decides about the class depending on the existence of a `DEFAULT_ACCESS_POLICY`. See here:
https://github.com/pulp/pulp_container/blob/master/pulp_container/app/registry_api.py#L246

pulpcore/app/access_policy.py
```diff
@@ -23,13 +23,23 @@ def get_policy_statements(self, request, view):
 
             AccessPolicyModel.objects.get(viewset_name=get_view_urlpattern(view))
 
+        If a matching `pulpcore.plugin.models.AccessPolicy` cannot be found, a default behavior of
+        allowing any authenticated user to perform any operation is assumed. This fallback allows
```
ipanova, line 27: we should adjust this to s/authed user/admin user

ipanova:
@bmbouter do you plan submit a PR for pulp-container that removes `AccessPolicyFromDB` from the permission classes? i can create one otherwise (PR/issue)

pulpcore/app/access_policy.py
bmbouter: I changed this to be `admin` instead, ty.

Regarding the setting of `permission_classes` on the viewset, if we did that then we wouldn't be allowing the user to totally disable permission classes by changing the default in the settings file. Overall Pulp should discontinue its use of `permission_classes` that way users can remain in control.

bmbouter:
> @bmbouter do you plan submit a PR for pulp-container that removes `AccessPolicyFromDB` from the permission classes? i can create one otherwise (PR/issue)

@ipanova I went just now to open this PR, but I found several cases I wasn't sure what to do with. It would be great if someone else could handle the changes in pulp_container after this is merged if that's alright.

bmbouter:
@mdellweg @gerrod3 @ipanova this is ready for review again please

docs/plugins/plugin-writer/concepts/rbac/overview.rst
```diff
@@ -45,15 +45,16 @@ To add authorization for a given resource, e.g. ``FileRemote``, you'll need to:
...
+   ``NamedModelViewSet``. This will contain both the ``statements`` and ``permissions_assignment``
+   attributes. See the :ref:`shipping_default_access_policy` documentation for more information on
+   this.
 
 **Enforce the Policy:**
 
-1. Define the ``permission_classes`` attribute on your Viewset referring to your subclass of
-   ``pulpcore.plugin.access_policy.AccessPolicyFromDB``. See the :ref:`viewset_enforcement` docs for
-   more information on this.
+1. The ``pulpcore.plugin.access_policy.AccessPolicyFromDB`` is configured as the default permission
```
gerrod3, line 55:
```suggestion
1. ``pulpcore.plugin.access_policy.AccessPolicyFromDB`` is configured as the default permission
```
or
```suggestion
1. The class ``pulpcore.plugin.access_policy.AccessPolicyFromDB`` is configured as the default permission
```

gerrod3 (approved):
Looks good!

docs/plugins/plugin-writer/concepts/rbac/access_policy.rst
```diff
@@ -205,16 +205,21 @@ generally causes them to pass any permission check even without explicit permiss
...
+``pulpcore.plugin.access_policy.AccessPolicyFromDB`` by default. This ensures that by defining a
+``DEFAULT_ACCESS_POLICY`` on your Viewset, Pulp will automatically save it to the database at
+migration-time, and your Viewset will be protected without additional effort.
 
-For example, here is the FileRemoteViewSet which enables authorization enforcement as follows:
+This strategy allows users to completely customize or disable the DRF Permission checks Pulp uses
+like any typical DRF project would.
 
-.. code-block:: python
+Also like a typical DRF project, individual Viewsets or views can also customized to use a different
```
mdellweg, line 216:
```suggestion
Also like a typical DRF project, individual Viewsets or views can also be customized to use a different
```

pulpcore/app/access_policy.py
```diff
@@ -23,13 +23,28 @@ def get_policy_statements(self, request, view):
 
             AccessPolicyModel.objects.get(viewset_name=get_view_urlpattern(view))
 
+        If a matching `pulpcore.plugin.models.AccessPolicy` cannot be found, a default behavior of
+        allowing any admin user to perform any operation is used. This fallback allows the Pulp RBAC
```
mdellweg, line 27: maybe
```suggestion
        allowing only admin users to perform any operation is used. This fallback allows the Pulp RBAC
```

pulpcore/app/access_policy.py
```diff
@@ -23,13 +23,28 @@ def get_policy_statements(self, request, view):
...
+        implementation to be turned on endpoint-by-endpoint with less effort. Once RBAC is fully
+        implemented on all endpoints, this fallback will be removed.
```
mdellweg, line 29: Do we really want to promise that? It needs complete RBAC on _all_ installed plugins.

pulpcore/app/access_policy.py
```diff
@@ -23,13 +23,28 @@ def get_policy_statements(self, request, view):
...
+            viewset_name = get_view_urlpattern(view)
+        except AttributeError:
+            return allow_authenticated
+
+        try:
+            access_policy_obj = AccessPolicyModel.objects.get(viewset_name=viewset_name)
+        except AccessPolicyModel.DoesNotExist:
+            return allow_authenticated
+        else:
+            return access_policy_obj.statements
```
mdellweg, line 50:
```suggestion
        try:
            access_policy_obj = AccessPolicyModel.objects.get(viewset_name=get_view_urlpattern(view))
        except AttributeError, AccessPolicyModel.DoesNotExist:
            return allow_authenticated
        return access_policy_obj.statements
```

Also maybe rename `allow_authenticated` to `allow_isadmin`.

mdellweg (approved):
None of the suggestions are really blockers => ACK

ipanova (approved):
there are some good suggestions from other reviewers

ipanova:
> > @bmbouter do you plan submit a PR for pulp-container that removes `AccessPolicyFromDB` from the permission classes? i can create one otherwise (PR/issue)
> 
> @ipanova I went just now to open this PR, but I found several cases I wasn't sure what to do with. It would be great if someone else could handle the changes in pulp_container after this is merged if that's alright.

I have create a task to remove AccessPolicyFromDB from the Viewsets https://pulp.plan.io/issues/9187 We cannot do it now because we declare compatibility with 3.14 and 3.15

pulpcore/app/access_policy.py
bmbouter: I'll remove this last sentence. I overall think that's the idea, but maybe it's not. We don't need to commit to it without more discussion.

pulpcore/app/access_policy.py
bmbouter: I originally had this but thought maybe it would be better the other way. This is so nicely compact though, I'll switch back to it.

bmbouter:
I believe addressed all the issues. Let me know if there is any more I can do. Thanks!

PR #1534, opened 2021-07-29
Handle a potential failure during cache invalidation

dralley:
A completed publication might not exist when we try to delete it.

closes: #9174
https://pulp.plan.io/issues/9174


pulpcore/app/models/publication.py
```diff
@@ -130,11 +130,18 @@ def delete(self, **kwargs):
...
-                pubs = Publication.objects.filter(repository_version__in=versions, complete=True)
-                publication = pubs.latest("repository_version", "pulp_created")
-                if self.pk == publication.pk:
-                    base_paths |= self.repository.distributions.values_list("base_path", flat=True)
+                # It's possible for errors to occur before any publication has been saved,
+                # so we need to handle the case when no Publication exists.
+                try:
+                    publication = Publication.objects.filter(
+                        repository_version=self.repository.latest_version(),
+                        complete=True
```
dralley, line 138: Should this `complete=True` actually be here?  I *think* that if we removed it it would eliminate the case that requires the `try` block.

pulpcore/app/models/publication.py
```diff
@@ -130,11 +130,19 @@ def delete(self, **kwargs):
...
-                versions = self.repository.versions.all()
-                pubs = Publication.objects.filter(repository_version__in=versions, complete=True)
-                publication = pubs.latest("repository_version", "pulp_created")
-                if self.pk == publication.pk:
-                    base_paths |= self.repository.distributions.values_list("base_path", flat=True)
+                # It's possible for errors to occur before any publication has been completed,
+                # so we need to handle the case when no Publication exists.
+                try:
+                    publication = Publication.objects.filter(
+                        repository_version=self.repository.latest_version(), complete=True
```
gerrod3, line 137: This logic is now different then what the content app uses for serving the latest publication. I guess I should move that code out into the distribution model so that we don't have duplication problems.

dralley: I see what you mean - actually I think I should revert that change.

gerrod3 (approved):

pulpcore/app/models/publication.py
```diff
@@ -130,11 +130,21 @@ def delete(self, **kwargs):
...
+                    versions = self.repository.versions.all()
+                    pubs = Publication.objects.filter(
+                        repository_version__in=versions, complete=True
+                    )
+                    publication = pubs.latest("repository_version", "pulp_created")
+                    if self.pk == publication.pk:
+                        base_paths |= self.repository.distributions.values_list(
+                            "base_path", flat=True
+                        )
+                except Publication.DoesNotExist:
```
dkliban, line 145: Is it line140 that's going to raise this exception?

dralley: Yup, exactly

dkliban (approved):

ggainey (approved):

daviddavis (approved):
LGTM

PR #1535, opened 2021-07-29
Fix unnecessary database connections when saving ProgressReports

gerrod3:
fixes: #9129

Should I also modify the `GroupProgressReport`'s `.update()` function to have the same behavior?

### DIFF
--- a/pulpcore/app/models/progress.py
+++ b/pulpcore/app/models/progress.py
@@ -1,8 +1,10 @@
 """
 Django models related to progress reporting
 """
+import asyncio
 import datetime
 import logging
+from asgiref.sync import sync_to_async
 from asyncio import CancelledError
 from gettext import gettext as _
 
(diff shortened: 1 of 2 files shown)


bmbouter:
> `GroupProgressReport`

Yes please

bmbouter:
@gerrod3 do you know why these test fail? I reran the tests and they failed again.

pulpcore/app/models/progress.py
```diff
@@ -136,12 +138,24 @@ def save(self, *args, **kwargs):
...
 
+    def _save(self, *args, **kwargs):
+        """Special save method to handle running inside a coroutine"""
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            loop = None
+        if loop and loop.is_running():
+            save = sync_to_async(super().save, thread_sensitive=True)
+            loop.create_task(save(*args, **kwargs))
```
bmbouter, line 155: Do we need to wait for that save to complete before we continue? If I'm reading this right, this will fire the save and continue which maybe is an important difference?

gerrod3: Yes, it currently just adds it as a task to complete which I think could cause desync issues.

bmbouter: I agree. I think it can wait on the task and it would make it sync again.

fao89 (commented):
It does look pretty good, I'm just wondering if this change could be done on `BaseModel`and `MasterModel`
https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/base.py#L140

goosemania:
In favor of #1553

PR #1537, opened 2021-07-30
Add touch to QuerySets of artifacts or content

mdellweg:
fixes TBA

This was just a random thought i had when writing these lines: https://github.com/pulp/pulp_container/blob/d25b01d4fb8aed1b264a9dcfc91461fe92d630b8/pulp_container/app/viewsets.py#L748

Not sure if it really helps there, also it's really kind of late for 3.14. If you like this i can do the issue dance for it.

### DIFF
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -92,6 +92,18 @@ class BulkCreateManager(models.Manager):
         return objs
 
 
+class BulkTouchQuerySet(models.QuerySet):
+    """
+    A query set that provides ``touch()``.
+    """
+
+    def touch(self):
+        """
+        Update the ``timestamp_of_interest`` on all objects of the query.
+        """
+        return self.update(timestamp_of_interest=now())
+
+
 class QueryMixin:
     """
     A mixin that provides models with querying utilities.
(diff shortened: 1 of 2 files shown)


daviddavis:
Ha! I actually had the same idea while reviewing these plugins PRs.

pulpcore/app/models/content.py
```diff
@@ -470,7 +482,7 @@ class Content(MasterModel, QueryMixin):
     _artifacts = models.ManyToManyField(Artifact, through="ContentArtifact")
     timestamp_of_interest = models.DateTimeField(auto_now=True)
 
-    objects = ContentManager()
+    objects = ContentManager.from_queryset(BulkTouchQuerySet)()
```
mdellweg, line 485: If it wasn't for the `content_pks` parameter, i think we could move `orphaned` to the query set as well.
But it would change an interface now.

ipanova (approved):

daviddavis (approved):
LGTM :+1: 

A unit test would be nice but not required.

mdellweg:
We believe this will get test coverage, once it's been used by #1553 .

PR #1541, opened 2021-08-03
Move files to artifact storage from WORKING_DIRECTORY.

dkliban:
Copy files from all other sources. Since users are able to change the
WORKING_DIRECTORY settings, this patch also ensures that we check that
the WORKING_DIRECTORY and MEDIA_ROOT are on the same device. Django's
FILE_UPLOAD_TEMP_DIR is now set to the value of WORKING_DIRECTORY.

fixes: #9146
https://pulp.plan.io/issues/9146/

### DIFF
--- a/pulpcore/app/checks.py
+++ b/pulpcore/app/checks.py
@@ -21,20 +21,20 @@ def storage_paths(app_configs, **kwargs):
             )
 
         try:
-            upload_temp_dir_dev = Path(settings.FILE_UPLOAD_TEMP_DIR).stat().st_dev
+            working_dir_dev = Path(settings.WORKING_DIRECTORY).stat().st_dev
         except OSError:
-            upload_temp_dir_dev = None
+            working_dir_dev = None
             warnings.append(
                 CheckWarning(
-                    "Your FILE_UPLOAD_TEMP_DIR setting points to a path that does not exist.",
+                    "Your WORKING_DIRECTORY setting points to a path that does not exist.",
                     id="pulpcore.W002",
                 )
             )
 
-        if media_root_dev and media_root_dev != upload_temp_dir_dev:
+        if media_root_dev and media_root_dev != working_dir_dev:
             warnings.append(
                 CheckWarning(
-                    "MEDIA_ROOT and FILE_UPLOAD_TEMP_DIR are on different filesystems. "
+                    "MEDIA_ROOT and WORKING_DIRECTORY are on different filesystems. "
                     "It is highly recommended that these live on the same filesystem",
                     id="pulpcore.W003",
                 )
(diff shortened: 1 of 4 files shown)


daviddavis (approved):

pulpcore/app/models/storage.py
```diff
@@ -63,7 +64,7 @@ def _save(self, name, content, max_length=None):
 
         try:
             if hasattr(content, "temporary_file_path") and content.temporary_file_path().startswith(
-                str(settings.DEPLOY_ROOT)
+                str(settings.WORKING_DIRECTORY)
```
dralley, line 67: https://github.com/pulp/pulpcore/blob/master/pulpcore/app/settings.py#L49-L50
https://github.com/pulp/pulpcore/search?q=FILE_UPLOAD_TEMP_DIR

Maybe we need to reconsider the existence of `FILE_UPLOAD_TEMP_DIR`.  Presumably the reason it exists is to allow the user to configure the temporary directory for file uploads, but in practice we are using it as the entire `WORKING_DIRECTORY` unless both are explicitly overridden - and in that case the `move` heuristic is going to be broken for file uploads.

dralley: Also - we don't document it, but we do document `WORKING_DIRECTORY`.

https://docs.pulpproject.org/pulpcore/configuration/settings.html#working-directory

There's another subtle bug here though.  If we allow overriding `WORKING_DIRECTORY` then it's not a safe assumption that we can move the files.

dkliban: `FILE_UPLOAD_TEMP_DIR` is a Django setting[0]. We need to set it so that files uploaded when creating an Artifact or "one shot" upload are stored in the correct place. Since users can modify `WORKING_DIRECTORY`, we should set `FILE_UPLOAD_TEMP_DIR` to the value of `WORKING_DIRECTORY`.

[0] https://docs.djangoproject.com/en/3.2/ref/settings/#file-upload-temp-dir

pulpcore/app/checks.py
```diff
@@ -21,20 +21,20 @@ def storage_paths(app_configs, **kwargs):
...
                     id="pulpcore.W002",
                 )
             )
 
-        if media_root_dev and media_root_dev != upload_temp_dir_dev:
+        if media_root_dev and media_root_dev != working_dir_dev:
             warnings.append(
                 CheckWarning(
-                    "MEDIA_ROOT and FILE_UPLOAD_TEMP_DIR are on different filesystems. "
+                    "MEDIA_ROOT and WORKING_DIRECTORY are on different filesystems. "
```
dralley, line 37: :+1:

dralley (approved):

daviddavis (approved):

PR #1545, opened 2021-08-06
Properly set Content-Type header for artifacts served from S3

gerrod3:
fixes: #9216

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -788,6 +788,8 @@ class Handler:
         elif settings.DEFAULT_FILE_STORAGE == "storages.backends.s3boto3.S3Boto3Storage":
             content_disposition = f"attachment;filename={content_artifact.relative_path}"
             parameters = {"ResponseContentDisposition": content_disposition}
+            if headers.get("Content-Type"):
+                parameters["ResponseContentType"] = headers.get("Content-Type")
             url = URL(
                 artifact_file.storage.url(artifact_file.name, parameters=parameters), encoded=True
             )
--- a/CHANGES/9216.bugfix
+++ b/CHANGES/9216.bugfix
@@ -1,0 +1,1 @@
+Content app now properly sets Content-Type header for artifacts being served from S3


mdellweg:
Are you sure, this is the right ticket you are referring to?

dralley:
It's the wrong issue, the correct one is https://pulp.plan.io/issues/9216

mdellweg:
I am unsure, whether it's a good idea to set the content type on Artifacts, as the same artifact (== file content) may be used by different `Content` classes which would wont to serve it with different `content-type` (aka MIME-type).
According to https://www.bennadel.com/blog/2696-overriding-content-type-and-content-disposition-headers-in-amazon-s3-pre-signed-urls.htm, you can adjust the returned `Content-Type` header at redirect time. I fail to find the corresponding section in the official S3 docs however.

gerrod3:
> I am unsure, whether it's a good idea to set the content type on Artifacts, as the same artifact (== file content) may be used by different `Content` classes which would wont to serve it with different `content-type` (aka MIME-type).
> According to https://www.bennadel.com/blog/2696-overriding-content-type-and-content-disposition-headers-in-amazon-s3-pre-signed-urls.htm, you can adjust the returned `Content-Type` header at redirect time. I fail to find the corresponding section in the official S3 docs however.
> 
> edit: There it is: "Overriding Response Header Values" https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html

I agree with your reasoning. Changing the header at redirect time is probably the better approach. I'll remove the changes to the API and instead try passing the headers along if they are set.

pulpcore/content/handler.py
```diff
@@ -788,6 +788,9 @@ async def _serve_content_artifact(self, content_artifact, headers):
         elif settings.DEFAULT_FILE_STORAGE == "storages.backends.s3boto3.S3Boto3Storage":
             content_disposition = f"attachment;filename={content_artifact.relative_path}"
             parameters = {"ResponseContentDisposition": content_disposition}
+            # Maybe make this a for loop with the other possible headers that can be overwritten?
```
mdellweg, line 791: I'd say: Let's wait for demands...
What headers do you think are worth considering? ContentDisposition for example is already done.

mdellweg (approved):
LGTM if you remov the hidden not-really-a-TODO comment.

ipanova (approved):

PR #1553, opened 2021-08-11
Ensure that the Stages API uses a single connection to the DB.

dkliban:
Django 3.2 opens up a new connection to the DB for each coroutine that accesses the DB. All ORM
interactions need to be wrapped in 'asgiref.sync.sync_to_async' in order to ensure that all DB
querries are executed using a single connection in a serial manner. This patch also provides
utility function `pulpcore.plugin.sync.sync_to_async_iterable` built on top of `sync_to_async`.
This functions is needed to operate on QuerySet objects and other synchronous iterables.

ProgressReport is enhanced with async interfaces: asave(), aincrease_by(), aincrement(),
__aenter__(), _aexit__(). Plugins should switch to the async interfaces in their Stages.

This patch also ensures that all transactions in the Stages API are run synchronously. This
requires deprecating the ContentStage._pre_save() and _post_save() coroutines in favor of
synchronous interfaces with the same name.
 
Each task that uses async code to access the DB now creates 2 connections to the DB. One to
communicate on the main thread and another connection for everything that accesses the DB
from coroutines such as the Stages API.

fixes: #9129

### DIFF
--- a/CHANGES/plugin_api/9129.feature
+++ b/CHANGES/plugin_api/9129.feature
@@ -1,0 +1,5 @@
+`pulpcore.plugin.models.ProgressReport` now has async interfaces: asave(), aincrease_by(),
+aincrement(), __aenter__(), _aexit__(). Plugins should switch to the async interfaces in their
+Stages. 
+`pulpcore.plugin.sync.sync_to_async_iterator` is a utility method to synchronize the database
+queries generated when a QuerySet is iterated. 
--- a/pulpcore/app/migrations/0010_pulp_fields.py
+++ b/pulpcore/app/migrations/0010_pulp_fields.py
@@ -450,7 +450,7 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='progressreport',
             name='task',
-            field=models.ForeignKey(default=pulpcore.app.models.task.Task.current, on_delete=django.db.models.deletion.CASCADE, related_name='progress_reports', to='core.Task', to_field='pulp_id'),
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='progress_reports', to='core.Task', to_field='pulp_id'),
         ),
         migrations.AlterField(
             model_name='publication',
--- a/pulpcore/app/models/progress.py
+++ b/pulpcore/app/models/progress.py
@@ -3,6 +3,7 @@ Django models related to progress reporting
 """
 import datetime
 import logging
+from asgiref.sync import sync_to_async
 from asyncio import CancelledError
 from gettext import gettext as _
 
@@ -113,9 +114,7 @@ class ProgressReport(BaseModel):
     total = models.IntegerField(null=True)
     done = models.IntegerField(default=0)
 
-    task = models.ForeignKey(
-        "Task", related_name="progress_reports", default=Task.current, on_delete=models.CASCADE
-    )
+    task = models.ForeignKey("Task", related_name="progress_reports", on_delete=models.CASCADE)
 
     suffix = models.TextField(null=True)
 
@@ -125,14 +124,15 @@ class ProgressReport(BaseModel):
     def save(self, *args, **kwargs):
         """
         Auto-set the task_id if running inside a task
-
         If the task_id is already set it will not be updated. If it is unset and this is running
         inside of a task it will be auto-set prior to saving.
-
         args (list): positional arguments to be passed on to the real save
         kwargs (dict): keyword arguments to be passed on to the real save
         """
         now = timezone.now()
+
+        if not self.task_id:
+            self.task = Task.current()
 
         if self._using_context_manager and self._last_save_time:
             if now - self._last_save_time >= datetime.timedelta(milliseconds=BATCH_INTERVAL):
@@ -142,12 +142,25 @@ class ProgressReport(BaseModel):
             super().save(*args, **kwargs)
             self._last_save_time = now
 
+    asave = sync_to_async(save)
+
     def __enter__(self):
         """
         Saves the progress report state as RUNNING
         """
         self.state = TASK_STATES.RUNNING
         self.save()
+
+        # Save needs occurs immediately so it is called before _using_context_manager is set
+        self._using_context_manager = True
+        return self
+
+    async def __aenter__(self):
+        """
+        Async implementation of __enter__
+        """
+        self.state = TASK_STATES.RUNNING
+        await self.asave()
 
         # Save needs occurs immediately so it is called before _using_context_manager is set
         self._using_context_manager = True
@@ -175,6 +188,29 @@ class ProgressReport(BaseModel):
             self.state = TASK_STATES.FAILED
         self.save()
 
+    async def __aexit__(self, type, value, traceback):
+        """
+        Async implementation of __exit__
+        Update the progress report state to COMPLETED, CANCELED, or FAILED.
+
+        If an exception occurs the progress report state is saved as:
+        - CANCELED if the exception is `asyncio.CancelledError`
+        - FAILED otherwise.
+
+        The exception is not suppressed. If the context manager exited without
+        exception the progress report state is saved as COMPLETED.
+
+        See the context manager documentation for more info on __exit__ parameters
+        """
+        self._using_context_manager = False
+        if type is None:
+            self.state = TASK_STATES.COMPLETED
+        elif type is CancelledError:
+            self.state = TASK_STATES.CANCELED
+        else:
+            self.state = TASK_STATES.FAILED
+        await self.asave()
+
     def increment(self):
         """
         Increment done count and save the progress report.
@@ -183,6 +219,8 @@ class ProgressReport(BaseModel):
         processing items.
         """
         self.increase_by(1)
+
+    aincrement = sync_to_async(increment)
 
     def increase_by(self, count):
         """
@@ -196,6 +234,8 @@ class ProgressReport(BaseModel):
             if self.done > self.total:
                 _logger.warning(_("Too many items processed for ProgressReport %s") % self.message)
         self.save()
+
+    aincrease_by = sync_to_async(increase_by)
 
     def iter(self, iter):
         """
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -3,10 +3,12 @@ from collections import defaultdict
 from gettext import gettext as _
 import logging
 
+from asgiref.sync import sync_to_async
 from django.db.models import Prefetch, prefetch_related_objects
 
 from pulpcore.plugin.exceptions import UnsupportedDigestValidationError
 from pulpcore.plugin.models import Artifact, ContentArtifact, ProgressReport, RemoteArtifact
+from pulpcore.plugin.sync import sync_to_async_iterable
 
 from .api import Stage
 
@@ -85,13 +87,14 @@ class QueryExistingArtifacts(Stage):
             # swap it out with the existing one.
             for digest_type, digests in artifact_digests_by_type.items():
                 query_params = {"{attr}__in".format(attr=digest_type): digests}
-                existing_artifacts = Artifact.objects.filter(**query_params).only(digest_type)
-                existing_artifacts.touch()
+                existing_artifacts_qs = Artifact.objects.filter(**query_params).only(digest_type)
+                existing_artifacts = sync_to_async_iterable(existing_artifacts_qs)
+                await sync_to_async(existing_artifacts_qs.touch)()
                 for d_content in batch:
                     for d_artifact in d_content.d_artifacts:
                         artifact_digest = getattr(d_artifact.artifact, digest_type)
                         if artifact_digest:
-                            for result in existing_artifacts:
+                            async for result in existing_artifacts:
                                 result_digest = getattr(result, digest_type)
                                 if result_digest == artifact_digest:
                                     d_artifact.artifact = result
@@ -156,7 +159,7 @@ class ArtifactDownloader(Stage):
         #    Set to None if stage is shutdown.
         content_get_task = _add_to_pending(content_iterator.__anext__())
 
-        with ProgressReport(
+        async with ProgressReport(
             message="Downloading Artifacts", code="sync.downloading.artifacts"
         ) as pb:
             try:
@@ -172,7 +175,7 @@ class ArtifactDownloader(Stage):
                                 content_get_task = None
                         else:
                             pb.done += task.result()  # download_count
-                            pb.save()
+                            await pb.asave()
 
                     if content_get_task and content_get_task not in pending:  # not yet shutdown
                         if len(pending) < self.max_concurrent_content:
@@ -237,7 +240,7 @@ class ArtifactSaver(Stage):
             if da_to_save:
                 for d_artifact, artifact in zip(
                     da_to_save,
-                    Artifact.objects.bulk_get_or_create(
+                    await sync_to_async(Artifact.objects.bulk_get_or_create)(
                         d_artifact.artifact for d_artifact in da_to_save
                     ),
                 ):
@@ -267,11 +270,13 @@ class RemoteArtifactSaver(Stage):
             The coroutine for this stage.
         """
         async for batch in self.batches():
-            RemoteArtifact.objects.bulk_get_or_create(self._needed_remote_artifacts(batch))
+            await sync_to_async(RemoteArtifact.objects.bulk_get_or_create)(
+                await self._needed_remote_artifacts(batch)
+            )
             for d_content in batch:
                 await self.put(d_content)
 
-    def _needed_remote_artifacts(self, batch):
+    async def _needed_remote_artifacts(self, batch):
         """
         Build a list of only :class:`~pulpcore.plugin.models.RemoteArtifact` that need
         to be created for the batch.
@@ -288,7 +293,7 @@ class RemoteArtifactSaver(Stage):
                 if d_artifact.remote:
                     remotes_present.add(d_artifact.remote)
 
-        prefetch_related_objects(
+        await sync_to_async(prefetch_related_objects)(
             [d_c.content for d_c in batch],
             Prefetch(
                 "contentartifact_set",
@@ -313,7 +318,9 @@ class RemoteArtifactSaver(Stage):
                 if not d_artifact.remote:
                     continue
 
-                for content_artifact in d_content.content._remote_artifact_saver_cas:
+                async for content_artifact in sync_to_async_iterable(
+                    d_content.content._remote_artifact_saver_cas
+                ):
                     if d_artifact.relative_path == content_artifact.relative_path:
                         break
                 else:
@@ -363,7 +370,9 @@ class RemoteArtifactSaver(Stage):
                             msg.format(rp=d_artifact.relative_path, c=d_content.content)
                         )
 
-                for remote_artifact in content_artifact._remote_artifact_saver_ras:
+                async for remote_artifact in sync_to_async_iterable(
+                    content_artifact._remote_artifact_saver_ras
+                ):
                     if remote_artifact.remote_id == d_artifact.remote.pk:
                         break
                 else:
--- a/pulpcore/plugin/stages/content_stages.py
+++ b/pulpcore/plugin/stages/content_stages.py
@@ -1,8 +1,13 @@
 from collections import defaultdict
+import inspect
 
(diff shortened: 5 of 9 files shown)


pulpcore/app/models/progress.py
```diff
@@ -136,24 +138,36 @@ def save(self, *args, **kwargs):
...
-    def __enter__(self):
+    async def _save(self, *args, **kwargs):
+        """Special save method to handle running inside a coroutine"""
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            loop = None
+        if loop and loop.is_running():
+            save = sync_to_async(super().save, thread_sensitive=True)
+            loop.create_task(save(*args, **kwargs))
```
gerrod3, line 155: I know this is a draft, but if you are converting this api to async, then you can just await the wrapped save method instead of creating a task. This will make the saving happen in serial again, instead of whenever the loop scheduler decides to run it.

dkliban: I decided to go with separate async interfaces for plugins to use. I pushed the latest changes.

pulpcore/app/models/progress.py
```diff
@@ -197,6 +255,21 @@ def increase_by(self, count):
...
+        Increase the done count and save the progress report.
+
+        This will increment and save the self.done attribute which is useful to put into a loop
+        processing items.
+        """
+        self.done += count
+        if self.total:
+            if self.done > self.total:
+                _logger.warning(_("Too many items processed for ProgressReport %s") % self.message)
+        await self.asave()
```
dralley, line 271: Just as an idea, F() expressions might be a little more ordering-safe.  I haven't looked into it deeply however

https://docs.djangoproject.com/en/3.2/ref/models/expressions/#f-expressions

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -316,14 +318,14 @@ def _needed_remote_artifacts(self, batch):
                 if not d_artifact.remote:
                     continue
 
-                for content_artifact in d_content.content._remote_artifact_saver_cas:
+                async for content_artifact in sync_to_async_iterable(d_content.content._remote_artifact_saver_cas):
```
dralley, line 321: Would it be simpler to just cast immediately to a list (behind a safe wrapper) and then iterate over that list like normal?

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -86,16 +88,16 @@ async def run(self):
             # swap it out with the existing one.
             for digest_type, digests in artifact_digests_by_type.items():
                 query_params = {"{attr}__in".format(attr=digest_type): digests}
-                existing_artifacts = Artifact.objects.filter(**query_params).only(digest_type)
+                existing_artifacts = sync_to_async_iterable(Artifact.objects.filter(**query_params).only(digest_type))
```
dralley, line 91: Same question as https://github.com/pulp/pulpcore/pull/1553/files#r687324322

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -316,14 +318,14 @@ def _needed_remote_artifacts(self, batch):
...
-                for content_artifact in d_content.content._remote_artifact_saver_cas:
+                async for content_artifact in sync_to_async_iterable(d_content.content._remote_artifact_saver_cas):
                     if d_artifact.relative_path == content_artifact.relative_path:
                         break
                 else:
                     msg = _('No declared artifact with relative path "{rp}" for content "{c}"')
                     raise ValueError(msg.format(rp=d_artifact.relative_path, c=d_content.content))
 
-                for remote_artifact in content_artifact._remote_artifact_saver_ras:
+                async for remote_artifact in sync_to_async_iterable(content_artifact._remote_artifact_saver_ras):
```
dralley, line 328: Same question as https://github.com/pulp/pulpcore/pull/1553/files#r687324322

pulpcore/plugin/stages/content_stages.py
```diff
@@ -48,10 +50,10 @@ async def run(self):
                     d_content_by_nat_key[d_content.content.natural_key()].append(d_content)
 
             for model_type in content_q_by_type.keys():
-                for result in model_type.objects.filter(content_q_by_type[model_type]).iterator():
+                async for result in sync_to_async_iterable(model_type.objects.filter(content_q_by_type[model_type]).iterator()):
```
dralley, line 53: Here though.. I'm not sure we have the option to do that, because of `.iterator()` (and it's a large list so we can't get away without it like we could in the other places.

pulpcore/plugin/stages/content_stages.py
```diff
@@ -48,10 +50,10 @@ async def run(self):
...
                     for d_content in d_content_by_nat_key[result.natural_key()]:
                         try:
-                            result.touch()
+                            await sync_to_async(result.touch, thread_sensitive=True)()
```
dralley, line 56: This has got to be murderous for performance.  I don't mean because of the change, I mean doing a database call (blocking or not) in the hot loop for every individual package being synced.  That's a lot of database traffic in either case, blocking or not.

pulpcore/plugin/stages/artifact_stages.py
mdellweg: Are you saying, that looping over a queryset will only produce a single db call to get all items at once?

pulpcore/plugin/stages/content_stages.py
mdellweg: Which means, i _need_ to get this going:

https://github.com/pulp/pulpcore/pull/1537

pulpcore/app/models/progress.py
```diff
@@ -142,6 +145,16 @@ def save(self, *args, **kwargs):
...
 
+    async def asave(self, *args, **kwargs):
+        """
+        Async save method that wraps the save() call with sync_to_async
+        args (list): positional arguments to be passed on to the real save
+        kwargs (dict): keyword arguments to be passed on to the real save
+        """
+        if not self.task_id:
+            self.task = await sync_to_async(Task.current)()
+        await sync_to_async(self.save)(*args, **kwargs)
```
mdellweg, line 156: The `Task.current()` will be called in `save` anyway. So you might get away with:
```suggestion
    asave = sync_to_async(self.save)
```

pulpcore/app/models/progress.py
mdellweg: Loving this idea! It _will_ solve race conditions where the count is not increased by two if two codepaths try to increase at the same time.

pulpcore/plugin/sync.py
```diff
@@ -0,0 +1,44 @@
+import inspect
+from functools import wraps
+
+from asgiref.sync import sync_to_async as _sync_to_async
+
+def sync_to_async(sync_fn, thread_sensitive=True):
```
mdellweg, line 6: I think, we should not redefine this function.
Rather be explicit with what the wrappers convert than try to guess. This is too much magic.

pulpcore/plugin/stages/artifact_stages.py
dralley: .iterator() is for disabling the django queryset cache mechanism and using a server-side cursor so that you don't blow out your memory usage on a large query.  In this case, since the size is limited by the batch size, `.iterator()` wouldn't make any difference and there's nothing preventing us from just extracting to a list.

pulpcore/plugin/stages/content_stages.py
```diff
@@ -236,7 +236,7 @@ async def run(self):
             The coroutine for this stage.
         """
         async with ProgressReport(message="Associating Content", code="associating.content") as pb:
-            to_delete = set(self.new_version.content.values_list("pk", flat=True))
+            to_delete = {i async for i in sync_to_async_iterable(self.new_version.content.values_list("pk", flat=True))}
```
mdellweg, line 239: This doesn't work. to_delete must be a set, to work like specified in lines 244/246.

mdellweg:
```suggestion
            to_delete = set(sync_to_async(self.new_version.content.values_list)("pk", flat=True))
```

pulpcore/plugin/stages/content_stages.py
dralley: I've filed https://pulp.plan.io/issues/9243

pulpcore/plugin/stages/content_stages.py
dkliban: { } is set comprehension notation. to_delete will be a set. I tried: 
`            to_delete = set(await sync_to_async(self.new_version.content.values_list)("pk", flat=True))
`
This causes an extra connection to open up.

mdellweg: Right. This is clever. Disregard my concerns.

pulpcore/app/models/progress.py
```diff
@@ -142,6 +142,16 @@ def save(self, *args, **kwargs):
...
             self._last_save_time = now
 
+    async def asave(self, *args, **kwargs):
+        """
+        Async save method that wraps the save() call with sync_to_async
+        args (list): positional arguments to be passed on to the real save
+        kwargs (dict): keyword arguments to be passed on to the real save
+        """
+        if not self.task_id:
+            self.task = await sync_to_async(Task.current)()
```
mdellweg, line 152: At least these two lines are superfluous.
```suggestion
```

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -3,6 +3,8 @@
 from gettext import gettext as _
 import logging
 
+from pulpcore.plugin.sync import sync_to_async, sync_to_async_iterable
```
mdellweg, line 6:
```suggestion
from asgiref.sync import sync_to_async
from pulpcore.plugin.sync import sync_to_async_iterable
```

mdellweg: if our implementation did something different, it should have a different name.

dkliban: i did not understand the value of the implementation suggested here. https://github.com/django/asgiref/issues/142

i decided that we just need the sync_to_async_iterable()

pulpcore/app/models/progress.py
dkliban: great idea!

dkliban: Using the F() expression requires us to reload the object from DB in order to perform the comparison to the total.

dkliban: I am going to remove the warning. It doesn't provide much value. Users and functional tests will catch this problem.

pulpcore/plugin/stages/content_stages.py
```diff
@@ -89,43 +96,62 @@ async def run(self):
...
+                    "ContentSaver._post_save() coroutine has been deprecated. Support will be "
+                    "dropped in pulpcore 3.16. ContentSaver._post_save() is now a synchronous "
+                    "function."
+                )
                 await self._post_save(batch)
             for declarative_content in batch:
                 await self.put(declarative_content)
 
-    async def _pre_save(self, batch):
+    def _pre_save(self, batch):
```
dralley, line 154: Could we avoid changing this API by adding a new method that wrapped this call?

mdellweg: Dennis and I discussed, that we might need to check if `_pre_save` is a coroutine and run it synchronously (`async_to_sync`) in the threadpool, while issuing a deprecation warning.

pulpcore/app/models/progress.py
mdellweg: Having the calculation being done in db with the rowlock in place is for sure the better alternative.

pulpcore/plugin/stages/content_stages.py
```diff
@@ -89,43 +96,62 @@ async def run(self):
...
-                ContentArtifact.objects.bulk_get_or_create(content_artifact_bulk)
+                                content_artifact_bulk.append(content_artifact)
+                    ContentArtifact.objects.bulk_get_or_create(content_artifact_bulk)
+                    if not inspect.iscoroutinefunction(self._post_save):
+                        self._post_save(batch)
+
+            if inspect.iscoroutinefunction(self._pre_save):
+                deprecation_logger.warning(
+                    "ContentSaver._pre_save() coroutine has been deprecated. Support will be "
+                    "dropped in pulpcore 3.16. ContentSaver._pre_save() is now a synchronous "
```
mdellweg, line 139:
```suggestion
                    "dropped in pulpcore 3.16. ContentSaver._pre_save() should now be a synchronous "
```
Also don't we need to call these inside the transaction (as we did pre 3.15)? It should be possible to run an async loop inside the single thread of the thread pool that only performes this coroutine (opening one addtional db connection).

pulpcore/plugin/stages/artifact_stages.py
mdellweg: The only thing i am complaining about here is that we import the imported `sync_to_async` from `pulpcore.plugin.sync` instead of directly from `asgiref.sync`.

mdellweg (commented):
This is pretty close to ready now. Great piece of work!

pulpcore/plugin/stages/content_stages.py
dkliban: I tried running it with async_to_sync() wrapper and the _post_save() would always through an exception saying that the distribution that was referenced did not exist yet. So that's why I moved the code out of the transaction.

pulpcore/plugin/stages/content_stages.py
dkliban: I tried calling them inside the transaction using async_to_sync() and it did not work as expected. RPM sync would always get an error in the _post_save() saying that the content was not yet in the database.

pulpcore/app/models/progress.py
```diff
@@ -142,6 +142,14 @@ def save(self, *args, **kwargs):
             super().save(*args, **kwargs)
             self._last_save_time = now
 
+    async def asave(self, *args, **kwargs):
+        """
+        Async save method that wraps the save() call with sync_to_async
+        args (list): positional arguments to be passed on to the real save
+        kwargs (dict): keyword arguments to be passed on to the real save
+        """
+        await sync_to_async(self.save)(*args, **kwargs)
```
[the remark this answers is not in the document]
dkliban: It works! I'll update the other async methods in the same way.

pulpcore/plugin/stages/content_stages.py
mdellweg: Because it is a new connection and therefore not _in_ the transaction. XD
So this is in a really bad shape. As long as we still have the async thingies. But what more could we do...

pulpcore/app/models/progress.py
```diff
@@ -142,6 +142,8 @@ def save(self, *args, **kwargs):
             super().save(*args, **kwargs)
             self._last_save_time = now
 
+    asave = sync_to_async(save)
```
mdellweg, line 145: Thank you!

pulpcore/plugin/stages/content_stages.py
mdellweg: Should we sound much more dangerous in the deprecation warning? It was inside the transaction for a reason i believe.

pulpcore/plugin/sync.py
```diff
@@ -0,0 +1,32 @@
...
+    try:
+        return next(it)
+    except StopIteration:
+        raise StopAsyncIteration
+
+
+async def sync_to_async_iterable(sync_iterable):
+    """
+    Utility method which runs each iteration of a synchronous iterable in a threadpool. It also
+    sets a threadlocal inside the thread so calls to AsyncToSync can escape it. The implementation
```
mdellweg, line 18: The second sentence i don't understand. "threadlocal" is an adjective to me, not a noun.

mdellweg (approved):
Just nit picking now.

pulpcore/app/migrations/0010_pulp_fields.py
```diff
@@ -450,7 +450,7 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='progressreport',
             name='task',
-            field=models.ForeignKey(default=pulpcore.app.models.task.Task.current, on_delete=django.db.models.deletion.CASCADE, related_name='progress_reports', to='core.Task', to_field='pulp_id'),
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='progress_reports', to='core.Task', to_field='pulp_id'),
```
dralley, line 453: Is this safe to do?

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -86,16 +88,18 @@ async def run(self):
...
                     for d_artifact in d_content.d_artifacts:
                         artifact_digest = getattr(d_artifact.artifact, digest_type)
                         if artifact_digest:
-                            for result in existing_artifacts:
+                            async for result in existing_artifacts:
                                 result_digest = getattr(result, digest_type)
                                 if result_digest == artifact_digest:
                                     try:
-                                        result.touch()
+                                        await sync_to_async(result.touch)()
```
dralley, line 102: If we can get the fix for https://pulp.plan.io/issues/9243 merged first it will be easier to backport.

CHANGES/plugin_api/9129.feature
```diff
@@ -0,0 +1,5 @@
+`pulpcore.plugin.models.ProgressReport` now has async interfaces: asave(), aincrease_by(),
+aincrement(), __aenter__(), _aexit__(). Plugins should switch to the async interfaces in their
+Stages. 
+`pulpcore.plugin.sync.sync_to_async_iterator` is a utility method to synchronize the database
+querries generated when a QuerySet is iterated.
```
fao89, line 5:
```suggestion
queries generated when a QuerySet is iterated. 
```

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -86,16 +88,18 @@ async def run(self):
             # swap it out with the existing one.
             for digest_type, digests in artifact_digests_by_type.items():
                 query_params = {"{attr}__in".format(attr=digest_type): digests}
-                existing_artifacts = Artifact.objects.filter(**query_params).only(digest_type)
+                existing_artifacts = sync_to_async_iterable(
+                    Artifact.objects.filter(**query_params).only(digest_type)
+                )
```
dralley, line 93: Unanswered question: https://github.com/pulp/pulpcore/pull/1553#discussion_r687324322

(Applies to a few other places below as well)

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -320,7 +326,9 @@ def _needed_remote_artifacts(self, batch):
                 if not d_artifact.remote:
                     continue
 
-                for content_artifact in d_content.content._remote_artifact_saver_cas:
+                async for content_artifact in sync_to_async_iterable(
+                    d_content.content._remote_artifact_saver_cas
+                ):
```
dralley, line 331: Same question as above, everything in here is limited by the size of the batch, so we could pull it directly into a list and iterate over that rather than using async.

pulpcore/plugin/sync.py
dralley: I've heard it used both ways, since in Python thread-local data is stored inside an instance of a magic class.

"thread-local variable" would make sense as a change, though, so +1

pulpcore/plugin/stages/content_stages.py
dralley: I wonder if the reason `async_to_sync` doesn't work is because each thread has a different handle for the database, and the transaction isn't shareable like that.

In that case I think that the entire block would need to be wrapped - but I'm not sure if nested async functions can be converted to sync code that way, so maybe that wouldn't work.

pulpcore/app/migrations/0010_pulp_fields.py
mdellweg: I guess so. It will prevent save if you didn't set the task, right?

pulpcore/plugin/stages/artifact_stages.py
dkliban: Converting the QuerySet to a list would still require using sync_async_iterable. So I am not seeing what benefit that would provide.

pulpcore/app/migrations/0010_pulp_fields.py
dkliban: This is safe to do. This change to the migration has no effect on the DB schema.

PR #1556, opened 2021-08-18
Deprecate resources argument of dispatch

mdellweg:
@dralley This is based on #1526 but i'll open it for early review now.
If you think this is the way to go i'm happy to assign a ticket to it.

### DIFF
--- a/pulpcore/app/viewsets/reclaim.py
+++ b/pulpcore/app/viewsets/reclaim.py
@@ -37,7 +37,7 @@ class ReclaimSpaceViewSet(ViewSet):
 
         task = dispatch(
             reclaim_space,
-            repos,
+            exclusive_resources=repos,
             kwargs={
                 "repo_pks": reclaim_repo_pks,
                 "keeplist_rv_pks": keeplist_rv_pks,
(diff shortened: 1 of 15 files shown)


mdellweg:
At some point we can even write a migration to split the lists of resources in the two list fields in the database.

dralley (approved):
:+1: 

I only reviewed the 2nd commit because I assume the first PR covers the other one.

pulpcore/app/viewsets/reclaim.py
```diff
@@ -37,7 +37,7 @@ def reclaim(self, request):
 
         task = dispatch(
             reclaim_space,
-            repos,
+            shared_resources=repos,
```
goosemania, line 40: This seems to be out of scope of this PR. I think it has nothing to do with the deprecation but it changes the current behaviour. correct?

pulpcore/app/viewsets/repository.py
```diff
@@ -195,7 +197,7 @@ def repair(self, request, repository_pk, number):
 
         task = dispatch(
             tasks.repository.repair_version,
-            [version.repository],
+            shared_resources=[version.repository],
```
goosemania, line 200: same here

pulpcore/app/viewsets/reclaim.py
mdellweg: Right, i'll make it a separate PR.

goosemania (approved):

PR #1558, opened 2021-08-19
Stop using blake2s to calulate lock identifiers

mdellweg:
Instead perform a simple xor on the first and second half of the task
uuid, igoring the sign bit.

fixes #9249


pulpcore/app/models/task.py
```diff
@@ -225,9 +225,8 @@ def save_heartbeat(self):
         self.save(update_fields=["last_heartbeat"])
 
 
-def _hash_to_u64(value):
-    _digest = blake2s(value.encode(), digest_size=8).digest()
-    return int.from_bytes(_digest, byteorder="big", signed=True)
+def _hash_to_s64(value):
+    return ctypes.c_int64((value >> 64) ^ value).value
```
dralley, line 229: Danger, Will Robinson...

Can we not just use SHA3?  They are FIPS approved, and available in Python 3.6+

fao89 (approved):
Tested locally with centos8 fips box and it worked fine!

pulpcore/app/models/task.py
mdellweg: This piece of code has no security implications whatsoever. It is really just there to spread out the tasks over a large number of buckets. Collisions are OK-ish.

dralley: I'm not concerned about security implications, moreso ctypes.  It seems like an unnecessarily high risk.

mdellweg: `(value >> 64) ^ value & 0x7FFFFFFFFFFFFFFF` would also work, but use only 126 bits of the uuid to represent a 63 bit number.

dralley: What I'm trying to get at though is that if the only problem with the previous method is the digest algorithm, then the easiest way to fix it is to use a digest algorithm that works.

Sha3 is part of [FIPS-202](https://csrc.nist.gov/publications/detail/fips/202/final) so it should be available.

mdellweg: My point is, we don't need anything cryptographically here. And the uuid4's are already randomly generated. It was a hash, because in an early state of design, we needed to hash Object hrefs to 64 bits. But really the signed representation of the last 8 bytes would probably work as well. This is really just to spread the tasks out into many buckets. Using a cryptographic algorithm kind of sends the wrong signals.

dralley: Using ctypes still feels like a big red flag to me, I would strongly prefer a solution that doesn't use it, even if it unnecessarily used a crypto function.

pulpcore/app/models/task.py
```diff
@@ -225,9 +224,8 @@ def save_heartbeat(self):
         self.save(update_fields=["last_heartbeat"])
 
 
-def _hash_to_u64(value):
-    _digest = blake2s(value.encode(), digest_size=8).digest()
-    return int.from_bytes(_digest, byteorder="big", signed=True)
+def _hash_to_s64(value):
```
dralley, line 227: Is `s64` a typo?  If you mean to signify that it can be signed, `i64` is commonly used in that context.

pulpcore/app/models/task.py
```diff
@@ -302,7 +300,7 @@ def __str__(self):
         return "Task: {name} [{state}]".format(name=self.name, state=self.state)
 
     def __enter__(self):
-        self.lock = _hash_to_u64(str(self.pk))
+        self.lock = _hash_to_s64(self.pk.int)
```
dralley, line 303: Also `_hash_to_*` is no longer an appropriate name if the input type is integer.

dralley (changes requested):
Code looks good, two issues with the naming.  Glad you were able to get rid of `ctypes`!

pulpcore/app/models/task.py
mdellweg: Well u64 was never right... anyway

pulpcore/app/models/task.py
```diff
@@ -225,9 +224,8 @@ def save_heartbeat(self):
         self.save(update_fields=["last_heartbeat"])
 
 
-def _hash_to_u64(value):
-    _digest = blake2s(value.encode(), digest_size=8).digest()
-    return int.from_bytes(_digest, byteorder="big", signed=True)
+def _uuid_to_advisory_lock(value):
```
dralley, line 227: Should the function do its own conversion to `int`?

dralley (approved):

ggainey:
There is an instance of _hash_to_u64() here : https://github.com/pulp/pulpcore/blob/master/pulpcore/tasking/util.py#L145

Doesn't look like it's used anywhere, could we delete it as part of this PR to clean it up and insure nobody calls it in the future?

pulpcore/tasking/util.py
```diff
@@ -140,8 +139,3 @@ def get_current_worker():
             return worker
 
     return None
-
-
-def _hash_to_u64(value):
-    _digest = blake2s(value.encode(), digest_size=8).digest()
-    return int.from_bytes(_digest, byteorder="big", signed=True)
```
mdellweg, line 147: @ggainey

ggainey: Ha! Sure enough - apologies for not looking closer at the PR

daviddavis (approved):

PR #1559, opened 2021-08-20
Properly update present ContentAtifacts after immediate sync

gerrod3:
fixes: #9101

This needs to be backported to 3.14 by Tuesday for Satellite, but it contains changes to the sync pipeline that need to be updated for the 3.15 async changes here: https://github.com/pulp/pulpcore/pull/1553. 

### DIFF
--- a/CHANGES/9101.bugfix
+++ b/CHANGES/9101.bugfix
@@ -1,0 +1,1 @@
+Artifacts are now being properly updated for Content after switching from 'on_demand' to 'immediate'.
--- a/pulpcore/plugin/stages/content_stages.py
+++ b/pulpcore/plugin/stages/content_stages.py
@@ -90,6 +90,9 @@ class ContentSaver(Stage):
         """
         async for batch in self.batches():
             content_artifact_bulk = []
+            to_update_ca_query = ContentArtifact.objects.none()
+            to_update_ca_bulk = []
+            to_update_ca_artifact = {}
             with transaction.atomic():
                 await self._pre_save(batch)
 
@@ -107,19 +110,37 @@ class ContentSaver(Stage):
                                 )
                             except ObjectDoesNotExist:
                                 raise e
+                        else:
+                            for d_artifact in d_content.d_artifacts:
+                                if not d_artifact.artifact._state.adding:
+                                    artifact = d_artifact.artifact
+                                else:
+                                    # set to None for on-demand synced artifacts
+                                    artifact = None
+                                content_artifact = ContentArtifact(
+                                    content=d_content.content,
+                                    artifact=artifact,
+                                    relative_path=d_artifact.relative_path,
+                                )
+                                content_artifact_bulk.append(content_artifact)
                             continue
-                        for d_artifact in d_content.d_artifacts:
-                            if not d_artifact.artifact._state.adding:
-                                artifact = d_artifact.artifact
-                            else:
-                                # set to None for on-demand synced artifacts
-                                artifact = None
-                            content_artifact = ContentArtifact(
-                                content=d_content.content,
-                                artifact=artifact,
-                                relative_path=d_artifact.relative_path,
+                    # When the Content already exists, check if ContentArtifacts need to be updated
+                    for d_artifact in d_content.d_artifacts:
+                        if not d_artifact.artifact._state.adding:
+                            # the artifact is already present in the database; update references
+                            # Creating one large query and one large dictionary
+                            to_update_ca_query |= ContentArtifact.objects.filter(
+                                content=d_content.content, relative_path=d_artifact.relative_path
                             )
-                            content_artifact_bulk.append(content_artifact)
+                            key = (d_content.content.pk, d_artifact.relative_path)
+                            to_update_ca_artifact[key] = d_artifact.artifact
+                # Query db once and update each object in memory for bulk_update call
+                for content_artifact in to_update_ca_query.iterator():
+                    key = (content_artifact.content_id, content_artifact.relative_path)
+                    # Maybe remove dict elements after to reduce memory?
+                    content_artifact.artifact = to_update_ca_artifact[key]
+                    to_update_ca_bulk.append(content_artifact)
+                ContentArtifact.objects.bulk_update(to_update_ca_bulk, ["artifact"])
                 ContentArtifact.objects.bulk_get_or_create(content_artifact_bulk)
                 await self._post_save(batch)
             for declarative_content in batch:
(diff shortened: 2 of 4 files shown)


pulpcore/tests/functional/api/using_plugin/test_sync.py
```diff
@@ -0,0 +1,71 @@
...
+        self.addCleanup(self.remote_api.delete, remote.pulp_href)
+
+        # Check content is still present, but artifacts are now there
+        content_response = self.cont_api.list()
+        self.assertEqual(content_response.count, 3)
+        for content in content_response.results:
+            self.assertNotEqual(content.artifact, None)
+
+        monitor_task(self.repo_api.delete(repo.pulp_href).task)
+        delete_orphans()
```
daviddavis, line 71: Why not put these in a `tearDown` method so that they are invoked even if the test fails?

https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown

pulpcore/tests/functional/api/using_plugin/test_reclaim_disk_space.py
```diff
@@ -91,15 +83,15 @@ def test_reclaim_immediate_content(self):
         # assert re-sync populated missing artifacts
         artifacts = self.artifacts_api.list().count
         self.assertGreater(artifacts, 0)
-        self.addCleanup(self.orphans_api.cleanup, {"orphan_protection_time": 0})
+        monitor_task(self.repo_api.delete(repo.pulp_href).task)
+        monitor_task(self.orphans_api.cleanup({"orphan_protection_time": 0}).task)
```
daviddavis, line 87: These as well.

CHANGES/9101.bugfix
```diff
@@ -0,0 +1 @@
+``ContentArtifact.artifact`` from 'on_demand' syncs is now properly updated after 'immediate' syncs
```
daviddavis, line 1: I think this should be more user-focused and less technical. Maybe:

```suggestion
Artifacts are now being properly updated for Content after switching from 'on_demand' to 'immediate'.
```

pulpcore/tests/functional/api/using_plugin/test_sync.py
gerrod3: Cleanups occur after `tearDown` so the repo wouldn't be deleted. But looking through the docs I can call `doCleanups` in `tearDown` and it'll pop the current functions off the stack.

daviddavis: I'm imagining you'd set a property (`self.repo`) and then the tearDown would look like this:

```
def tearDown(self):
    if self.repo:
        monitor_task(self.repo_api.delete(self.repo.pulp_href).task)
    delete_orphans()
```

daviddavis (approved):

pulpcore/tests/functional/api/using_plugin/test_sync.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def tearDown(self):
+        """Clean up Pulp after testing."""
+        self.doCleanups()
+        delete_orphans()
+
+    def test_ondemand_to_immediate_sync(self):
+        """Checks that content artifacts are updated following on-demand -> immediate sync."""
+        # Ensure that no content is present
+        content_response = self.cont_api.list()
```
mdellweg, line 45: This may speed up the test...
```suggestion
        content_response = self.cont_api.list(limit=1)
```

pulpcore/tests/functional/api/using_plugin/test_sync.py
```diff
@@ -0,0 +1,72 @@
...
+        self.assertEqual(content_response.count, 0)
```
mdellweg, line 46: This test is deemed to fail on any preloaded developer box, right?

pulpcore/tests/functional/api/using_plugin/test_sync.py
```diff
@@ -0,0 +1,72 @@
...
+        body = RepositorySyncURL(remote=remote.pulp_href)
+        monitor_task(self.repo_api.sync(repo.pulp_href, body).task)
+        self.addCleanup(self.repo_api.delete, repo.pulp_href)
+        self.addCleanup(self.remote_api.delete, remote.pulp_href)
+
+        # Check content is present, but no artifacts are there
+        content_response = self.cont_api.list()
+        self.assertEqual(content_response.count, 3)
+        for content in content_response.results:
+            self.assertEqual(content.artifact, None)
```
mdellweg, line 60: Can't you only look at the content in the latest repo version here?
And then after resynching, check that the repo version is still the same, but the content now has artifacts?

pulpcore/tests/functional/api/using_plugin/test_sync.py
gerrod3: Yes.

pulpcore/tests/functional/api/using_plugin/test_sync.py
gerrod3: I could.

mdellweg (approved):

pulpcore/tests/functional/api/using_plugin/test_sync.py
gerrod3: I switch the test to be skipped if content is existing on the machine.

pulpcore/plugin/stages/content_stages.py
```diff
@@ -107,19 +110,37 @@ async def run(self):
...
+                            # the artifact is already present in the database; update references
+                            # Creating one large query and one large dictionary
+                            to_update_ca_query |= ContentArtifact.objects.filter(
+                                content=d_content.content, relative_path=d_artifact.relative_path
                             )
-                            content_artifact_bulk.append(content_artifact)
+                            key = (d_content.content.pk, d_artifact.relative_path)
+                            to_update_ca_artifact[key] = d_artifact.artifact
+                # Query db once and update each object in memory for bulk_update call
+                for content_artifact in to_update_ca_query.iterator():
```
dralley, line 138: Add `.select_related("content")` to avoid adding an extra query per access of `.content`

dralley:
@gerrod3 With one performance tweak I think this is good enough to merge, thanks!

pulpcore/plugin/stages/content_stages.py
gerrod3: I changed it to `.content_id` which is always present in the initial state.

dralley (approved):

PR #1562, opened 2021-08-20
Use bulk touch() when processing Artifacts and Content

dkliban:
This patch only addresses part of the problem described in issue #9243. Another patch
is needed to address the inefficiency of the repository version modify operation.

re: #9243

### DIFF
--- a/CHANGES/9243.bugfix
+++ b/CHANGES/9243.bugfix
@@ -1,0 +1,1 @@
+Fixed repository sync performance regression introduced in pulpcore 3.14.
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -3,7 +3,6 @@ from collections import defaultdict
 from gettext import gettext as _
 import logging
 
-from django.db import DatabaseError
 from django.db.models import Prefetch, prefetch_related_objects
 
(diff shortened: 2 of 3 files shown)


dralley (approved):

dralley:
@dkliban I created a backport issue, we will just want to be careful that if there are multiple PRs / commits, they all get backported.

daviddavis (approved):

dralley:
Since we have 3 open PRs that will conflict I'm thinking the best order to merge the PRs is:

1. This one
2. Gerrod's fix for 9101
3. The async connections PR

Because #3 isn't going to be backported.  Agreed @dkliban ?

dkliban:
> Since we have 3 open PRs that will conflict I'm thinking the best order to merge the PRs is:
> 
>     1. This one
> 
>     2. Gerrod's fix for 9101
> 
>     3. The async connections PR
> 
> 
> Because #3 isn't going to be backported. Agreed @dkliban ?

This is true for upstream. However, I could see someone wanting to add Django 3.2 support for pulpcore 3.14 downstream.

dralley:
I kind of hope we don't do that, but even if we do, it still makes sense (because the backport would happen much later, after the others are backported)

Is this ready to go?

PR #1564, opened 2021-08-21
3.14.5 cherrypicks

dralley:
The [second commit](https://github.com/pulp/pulpcore/pull/1564/commits/bd2a68b777b646023056495dd3bb9e751e8bb6cf) is manual.

I went ahead and handled these few since a lot of the bugfixes this release are interdependent or conflicting - better to handle them in smaller chunks than try to sort everything out at the end.


daviddavis:
This looks good to me so far. Thanks for working on it.

daviddavis (approved):

dralley:
Discovered an issue, will need to include https://github.com/pulp/pulpcore/pull/1565 as well.

goosemania:
#1559 needs to be backported. Creatively.

CHANGES/plugin_api/9268.feature
```diff
@@ -0,0 +1,2 @@
+Added ``touch`` to Artifact and Content query sets for bulk operation.
```
goosemania, line 1: I see in the description of the backport that this feature is considered a performance bug.
Could we adjust this changelog entry here? to be `.bugfix` and to state that it's needed for plugins to address performance regression.
We should not have any features in the changelog for the z stream. It would be very confusing for users I think.

dralley: I'm OK with that suggestion, but if we go that route I think we should skip mentioning the plugin API change at all.  It's not  "needed" for the performance improvement since if the plugin writer hasn't adopted it we fall back to not `touch()`-ing at all, which is OK because it has no impact on the operation of 3.14.  We backport it just to enable fixing it, anyway, in the one case we know of.

goosemania: Then `.misc` works.

goosemania:
#1567 needs to be backported as well.

goosemania (approved):

PR #1565, opened 2021-08-21
Fix issue where plugins with custom content managers can't use touch()

dralley:
closes: #9269
https://pulp.plan.io/issues/9269

### DIFF
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -460,6 +460,9 @@ class ContentManager(BulkCreateManager):
         return self.filter(version_memberships__isnull=True, timestamp_of_interest__lt=expiration)
 
 
+ContentManager = ContentManager.from_queryset(BulkTouchQuerySet)
+
+
 class Content(MasterModel, QueryMixin):
     """
     A piece of managed content.
@@ -482,7 +485,7 @@ class Content(MasterModel, QueryMixin):
     _artifacts = models.ManyToManyField(Artifact, through="ContentArtifact")
     timestamp_of_interest = models.DateTimeField(auto_now=True)
 
-    objects = ContentManager.from_queryset(BulkTouchQuerySet)()
+    objects = ContentManager()
 
     class Meta:
         verbose_name_plural = "content"
--- a/pulpcore/plugin/stages/content_stages.py
+++ b/pulpcore/plugin/stages/content_stages.py
@@ -47,9 +47,22 @@ class QueryExistingContents(Stage):
                     content_q_by_type[model_type] = content_q_by_type[model_type] | unit_q
                     d_content_by_nat_key[d_content.content.natural_key()].append(d_content)
 
-            for model_type in content_q_by_type.keys():
-                model_type.objects.filter(content_q_by_type[model_type]).touch()
-                for result in model_type.objects.filter(content_q_by_type[model_type]).iterator():
+            for model_type, content_q in content_q_by_type.items():
+                try:
+                    model_type.objects.filter(content_q).touch()
+                except AttributeError:
+                    from pulpcore.app.loggers import deprecation_logger
+                    from gettext import gettext as _
+
+                    deprecation_logger.warning(
+                        _(
+                            "As of pulpcore 3.14.5, plugins which declare custom ORM managers on "
+                            "their content classes should have those managers inherit from "
+                            "pulpcore.plugin.models.ContentManager. This will become a hard error "
+                            "in the future."
+                        )
+                    )
+                for result in model_type.objects.filter(content_q).iterator():
                     for d_content in d_content_by_nat_key[result.natural_key()]:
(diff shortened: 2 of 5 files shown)


pulpcore/plugin/stages/content_stages.py
```diff
@@ -47,9 +47,9 @@ async def run(self):
                     content_q_by_type[model_type] = content_q_by_type[model_type] | unit_q
                     d_content_by_nat_key[d_content.content.natural_key()].append(d_content)
 
-            for model_type in content_q_by_type.keys():
-                model_type.objects.filter(content_q_by_type[model_type]).touch()
-                for result in model_type.objects.filter(content_q_by_type[model_type]).iterator():
+            for model_type, content_q in content_q_by_type.items():
+                model_type.objects.filter(content_q).touch()
+                for result in model_type.objects.filter(content_q).iterator():
```
dralley, line 52: This isn't a real change, just a cleanup, I can drop it if desired.

pulpcore/app/models/content.py
```diff
@@ -460,6 +460,9 @@ def orphaned(self, orphan_protection_time, content_pks=None):
         return self.filter(version_memberships__isnull=True, timestamp_of_interest__lt=expiration)
 
 
+ContentManager = ContentManager.from_queryset(BulkTouchQuerySet)
```
dralley, line 463: I'm a little surprised this worked, and slightly cautious

pulpcore/plugin/stages/content_stages.py
```diff
@@ -47,9 +47,9 @@ async def run(self):
                     content_q_by_type[model_type] = content_q_by_type[model_type] | unit_q
                     d_content_by_nat_key[d_content.content.natural_key()].append(d_content)
 
-            for model_type in content_q_by_type.keys():
-                model_type.objects.filter(content_q_by_type[model_type]).touch()
-                for result in model_type.objects.filter(content_q_by_type[model_type]).iterator():
+            for model_type, content_q in content_q_by_type.items():
+                model_type.objects.filter(content_q).touch()
```
dralley, line 51: If we're backporting this then we need to allow it to fail, because existing plugins  won't have the appropriate code.  Can I stick it in a try / catch?

3.14 still has the blocking orphan cleanup so it doesn't matter much if it doesn't work.

pulpcore/app/models/content.py
mdellweg: That’s Python. You could rename the blank class to ’_ContentManager’.

pulpcore/plugin/stages/content_stages.py
mdellweg: With break, do you mean if the plugin defined it's own ContentManager without `touch`?

pulpcore/app/models/content.py
mdellweg: Another possibility (and probably way to complex) would be to write a class decorator for this.

pulpcore/plugin/stages/content_stages.py
mdellweg: So this thing is something we want backported, right? So at least we care that it does not raise an uncought exception. And telling plugin writers, that they need to act via a deprecation warning sounds good to me.

dralley: > With break, do you mean if the plugin defined it's own ContentManager without touch?

Exactly.  Merging the original patch broke the RPM plugin which does this.  It might be the *only* plugin that does this, I'm not sure.

pulpcore/app/models/content.py
dralley: I think I agree about class decorator too complex - I'm on the fence about the rename though, if you and/or David have a preference I'll do that.

mdellweg: I don't need it.

pulpcore/plugin/stages/content_stages.py
```diff
@@ -47,9 +47,22 @@ async def run(self):
...
+                except AttributeError:
+                    from pulpcore.app.loggers import deprecation_logger
+                    from gettext import gettext as _
+
+                    deprecation_logger.warning(
+                        _(
+                            "As of pulpcore 3.14.5, plugins which declare custom ORM managers on "
+                            "their content classes should have those managers inherit from "
+                            "pulpcore.plugin.models.ContentManager. This will become a hard error "
+                            "in the future."
```
mdellweg, line 62: I'll just assume that creating the manager with `from_queryset` does not hinder any further subclassing here.

mdellweg (approved):

pulpcore/plugin/stages/content_stages.py
dralley: It does not seem to

daviddavis (approved):

PR #1569, opened 2021-08-24
Ensure that the Stages API uses a single connection to the DB.

mdellweg:
Django 3.2 opens up a new connection to the DB for each coroutine that accesses the DB. All ORM
interactions need to be wrapped in 'asgiref.sync.sync_to_async' in order to ensure that all DB
querries are executed using a single connection in a serial manner. This patch also provides
utility function `pulpcore.plugin.sync.sync_to_async_iterable` built on top of `sync_to_async`.
This functions is needed to operate on QuerySet objects and other synchronous iterables.

ProgressReport is enhanced with async interfaces: asave(), aincrease_by(), aincrement(),
__aenter__(), __aexit__(). Plugins should switch to the async interfaces in their Stages.

This patch also ensures that all transactions in the Stages API are run synchronously. This
requires deprecating the ContentStage._pre_save() and _post_save() coroutines in favor of
synchronous interfaces with the same name.

Each task that uses async code to access the DB now creates 2 connections to the DB. One to
communicate on the main thread and another connection for everything that accesses the DB
from coroutines such as the Stages API.

fixes: #9129

### DIFF
--- a/CHANGES/plugin_api/9129.deprecation
+++ b/CHANGES/plugin_api/9129.deprecation
@@ -1,0 +1,2 @@
+ContentSaver._pre_save() and ContentSaver._post_save() hooks are no longer coroutines. They should
+be implemented as syncronous functions.
--- a/pulpcore/app/models/progress.py
+++ b/pulpcore/app/models/progress.py
@@ -3,6 +3,7 @@ Django models related to progress reporting
 """
 import datetime
 import logging
+from asgiref.sync import sync_to_async
 from asyncio import CancelledError
 from gettext import gettext as _
 
@@ -113,9 +114,7 @@ class ProgressReport(BaseModel):
     total = models.IntegerField(null=True)
     done = models.IntegerField(default=0)
 
-    task = models.ForeignKey(
-        "Task", related_name="progress_reports", default=Task.current, on_delete=models.CASCADE
-    )
+    task = models.ForeignKey("Task", related_name="progress_reports", on_delete=models.CASCADE)
 
     suffix = models.TextField(null=True)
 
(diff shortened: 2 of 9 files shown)


mdellweg:
@gerrod3 can you check, that i resolved the conflict in the `ContentSaver` stage properly?

gerrod3:
`ContentSaver` looks good.

pulpcore/app/models/progress.py
```diff
@@ -125,15 +124,16 @@ class ProgressReport(BaseModel):
     def save(self, *args, **kwargs):
         """
         Auto-set the task_id if running inside a task
-
```
daviddavis, line 128: Removing this line causes this docstring to not conform to pep 257:

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line.

pulpcore/app/models/progress.py
```diff
@@ -125,15 +124,16 @@ class ProgressReport(BaseModel):
...
         If the task_id is already set it will not be updated. If it is unset and this is running
         inside of a task it will be auto-set prior to saving.
-
```
daviddavis, line 131: We typically separate the argument list with a blank line in docstrings.

daviddavis: I see now looking at the old PR that you didn't make these docstring changes. I'm happy to merge this and fix this up myself.

CHANGES/plugin_api/9129.deprecation
```diff
@@ -0,0 +1,2 @@
+ContentSaver._pre_save() and ContentSaver._post_save() hooks are no longer coroutines. They should
+be implemented as syncronous functions.
```
daviddavis, line 2:
```suggestion
be implemented as synchronous functions.
```

pulpcore/app/models/progress.py
dralley: I'm surprised the linting didn't complain

daviddavis: We don't link docstrings in pulpcore unfortunately ;(

pulpcore/app/models/progress.py
```diff
@@ -175,6 +188,29 @@ def __exit__(self, type, value, traceback):
             self.state = TASK_STATES.FAILED
         self.save()
 
+    async def __aexit__(self, type, value, traceback):
+        """
+        Async implementation of __exit__
+        Update the progress report state to COMPLETED, CANCELED, or FAILED.
```
daviddavis, line 194:
```suggestion
        Async implementation of __exit__
        
        Update the progress report state to COMPLETED, CANCELED, or FAILED.
```

PR #1586, opened 2021-08-28
Update CI files

pulpbot:
[noissue]


.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
   lint:
     runs-on: ubuntu-latest
-
+
```
mdellweg, line 17: How comes, we didn't generate those trailing white spaces before?
Since this is generated code, i don't care too much, however.

mdellweg (approved):

dralley (approved):

PR #1587, opened 2021-08-30
Remove legacy tasking system

mdellweg:
fixes #9157


pulpcore/app/models/task.py
```diff
@@ -289,11 +231,6 @@ class Task(BaseModel, AutoDeleteObjPermsMixin, AutoAddObjPermsMixin):
     )
     reserved_resources_record = ArrayField(models.CharField(max_length=256), null=True)
 
-    # TODO: find a solution that makes this unnecessary
```
mdellweg, line 292: :ballot_box_with_check: finally :checkered_flag:

pulpcore/tasking/entrypoint.py
```diff
@@ -13,12 +12,10 @@
 
 django.setup()
 
-from django.conf import settings  # noqa: E402: module level not at top of file
-
-
 _logger = logging.getLogger(__name__)
 
 
+# TODO Are we ready to remove this option?
```
dralley, line 18: I think so, yes.

pulpcore/tasking/entrypoint.py
```diff
@@ -33,30 +30,10 @@ def worker(resource_manager, pid):
...
-            "pulpcore.tasking.worker.PulpWorker",
-            "-c",
-            "pulpcore.rqconfig",
-            "--disable-job-desc-logging",
-        ]
-        if resource_manager:
-            args.extend(["-n", "resource-manager"])
-        sys.argv = args
-        main()
+    if resource_manager:
```
dralley, line 33: Likewise

pulpcore/tasking/worker_watcher.py
```diff
@@ -1,17 +1,7 @@
...
-from pulpcore.tasking.util import cancel
-
-from rq.exceptions import NoSuchJobError
-from rq.job import Job
-
-from django.db import transaction
+from pulpcore.app.models import Worker
 
 
 _logger = logging.getLogger(__name__)
```
dralley, line 7: If this one function is all that is left, we should move it to one of the other files and get rid of this one.

pulpcore/tasking/entrypoint.py
mdellweg: It will break systemd files for the resource-manager that are deployed in all different installer scenarios.
So the question to me is how much do we want to break those to force them fixing it? Or should we warn one release longer, now that you cannot even start a resource-mananger if you wanted?

pulpcore/tasking/worker_watcher.py
mdellweg: `pulpcore_worker.py` probably.

pulpcore/tasking/entrypoint.py
bmbouter: Personally I'm in favor of removing it. It's going to be a breaking change at some point anyway. It shouldn't have a very large impact since most users will have this auto-handled for them by the installer (which would need a corresponding update in the same release for sure). That's my opinion anyway. If others really want to leave it I'm ok with that too.

mdellweg: And that is exactly my point. We would have less churn if we give the installer one release cycle to remove that service.

The reason, why i am cautious is that i think if i remove the option and the user attempts to start a resource-manager, it will fail hard ad respawn the service in a busy loop instead of just starting an additional worker (In the end both are undesirable, but the latter is a little less so.).

bmbouter: I imagined the installer would have to handle this in the same release. I'm ok with either approach. Can you decide and file/update the redmine tasks with the corresponding releases?

pulpcore/tasking/worker_watcher.py
mdellweg: done.

pulpcore/tasking/entrypoint.py
bmbouter: Thanks @mdellweg . I updated that issue with the 3.17.0 milestone so that it will formally block that release.

CHANGES/plugin_api/9157.removal
```diff
@@ -0,0 +1,2 @@
+Removed the ``reserved_resources_record__resource`` filter for Task. Use
```
bmbouter, line 1: Maybe?
```suggestion
Removed the deprecated ``reserved_resources_record__resource`` filter for Task. Use
```

bmbouter (approved):
I reviewed this and also hand tested it locally. I don't see any issues.

bmbouter:
> There is the redis connection made with rq.cli.helper. Do you want me to clean this up here, too?

Yes, I'd also like us to drop rq as a dependency with this change too.

dralley (approved):

pulpcore/tasking/connection.py
```diff
@@ -7,6 +7,9 @@
 _a_conn = None
 
 
+# TODO refactor this without ``rq``
```
dralley, line 10: I would suggest just vendoring this function (with the appropriate copyright notice)

https://github.com/rq/rq/blob/master/rq/cli/helpers.py#L39-L75

requirements.txt
```diff
@@ -26,7 +26,6 @@ psycopg2~=2.9.1
 PyYAML>=5.1.1,<5.5.0
 python-gnupg~=0.4.7
 redis>=3.4.0
-rq~=1.9.0
```
mdellweg, line 29: @bmbouter FYI

bmbouter: Awww yissss. Now we're talking!

docs/configuration/settings.rst
```diff
@@ -100,9 +100,7 @@ The following RQ settings can be set in your Pulp config:
   * REDIS_PASSWORD
   * SENTINEL
 
-These will be used by any worker loaded with the ``-c 'pulpcore.rqconfig'`` option.
-
-Below are some common settings used for RQ configuration. See the `RQ settings documentation
+Below are some common settings used for Redis configuration. See the `RQ settings documentation
```
bmbouter, line 103: RQ still here

pulpcore/tasking/connection.py
```diff
@@ -1,12 +1,56 @@
...
-from rq.cli.helpers import get_redis_from_config
 
 from pulpcore.app.settings import settings
 
 _conn = None
 _a_conn = None
 
 
+# This function is vendored from rq
+# https://github.com/rq/rq/blob/c556106a38505ea7485daec9420c408f720d2961/rq/cli/helpers.py#L39
```
bmbouter, line 12: If we're going to do this we need to also have it carry the copyright from RQ (per its license). Additionally, it's possible that license and GPLv2 (our code) may or may not be appropriate to distribute together (I don't know). For these reasons, is there any way we can just make something new ourselves?

For example, we could just move these settings to our settings file, and have pulp users override them just like normal, and built the connection from that?

dralley: It's the BSD 2 clause license, they should be compatible.

https://opensource.org/licenses/BSD-2-Clause
https://www.gnu.org/licenses/license-list.en.html#FreeBSD

bmbouter: Also hand-in-hand with that is the removal of this file https://github.com/pulp/pulpcore/blob/master/pulpcore/rqconfig.py which I believe should be removed. We should have the settings load like all pulp settings using dynaconf instead of through this settings chain loader.

docs/configuration/settings.rst
bmbouter: Ah yes, but I guess I don't think we should be doing config from file anymore.

bmbouter: Config from file was only really to support feature parity with how RQ ran its workers.

PR #1592, opened 2021-08-31
Provide a way for the plugin writer to silence warnings from downloaders

dralley:
closes: #9317
https://pulp.plan.io/issues/9317

### DIFF
--- a/pulpcore/download/base.py
+++ b/pulpcore/download/base.py
@@ -224,7 +224,7 @@ class BaseDownloader:
             if self._size != self.expected_size:
                 raise SizeValidationError(self.url)
 
-    async def run(self, extra_data=None):
+    async def run(self, extra_data=None, quiet=False):
         """
         Run the downloader with concurrency restriction.
 
@@ -234,6 +234,7 @@ class BaseDownloader:
 
         Args:
             extra_data (dict): Extra data passed to the downloader.
+            quiet (bool): Don't log on backoff or failure events.
 
         Returns:
             :class:`~pulpcore.plugin.download.DownloadResult` from `_run()`.
--- a/CHANGES/plugin_api/9317.feature
+++ b/CHANGES/plugin_api/9317.feature
@@ -1,0 +1,1 @@
+`HttpDownloader.run()` now includes a `quiet` argument which prevents messages from being logged on backoff or give-up events.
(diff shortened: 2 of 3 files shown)


mdellweg:
This does not cover 404s because there is no need to retry. Right?

dralley:
@mdellweg It covers 404s as well.  That counts as a "failure" it just happens immediately and doesn't attempt to retry.  Basically if `quiet=True` you won't see anything in the logs about these downloads.

The purpose is we sometimes try to download files that are optional and not guaranteed to be there, so we don't want to log errors in the event that they are not.

Here's a plugin example: https://github.com/pulp/pulp_rpm/pull/2112

mdellweg:
That sounds great. Attempting to download metadata that _might_ be there can be quite common.

pulpcore/download/base.py
```diff
@@ -234,6 +234,7 @@ async def run(self, extra_data=None):
 
         Args:
             extra_data (dict): Extra data passed to the downloader.
+            quiet (bool): Don't log on backoff or failure events.
```
bmbouter, line 237: I want to think this over some. This is a leakly abstraction because base has no notion of backoff events, that's a pure HTTPDownloader thing.

dralley: We could use `*args` and **kwargs` instead

bmbouter: We keep running into this pattern actually. Maybe we should just put `*args` and `**kwargs` onto BaseDownloader

bmbouter:
How can we integrate this with the existing `raise_for_status` feature. We've been having plugin writers already using that which silences the 400+ errors already.

bmbouter:
Like this actually? Maybe your Pr should replace this usage?

dralley:
@bmbouter I think raising exceptions is necessary for the backoffs to occur properly though, so we can't touch that.

bmbouter:
I think I'm still slow from coming back from paternity. When I read "The purpose is we sometimes try to download files that are optional and not guaranteed to be there, so we don't want to log errors in the event that they are not." I think to myself that is why we made the `silence_errors_for_response_status_codes` argument [here](https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/downloaders.py#L48). What you've made here just seems a lot nicer, but there is so much overlap with what we have already I want to somehow not have two things providing piecemeal value. @dralley wdyt?

dralley:
@bmbouter You know, I think this actually *does* work, but I'm still kind of confused about the mechanism.  I'll spend a few minutes digging into the code and then close this PR if it turns out we don't need it.

PR #1599, opened 2021-08-31
Adds architecture diagram to the components page

bmbouter:
This diagram originally comes from: https://docs.google.com/drawings/d/1P2w2jQh03t_KB27FtQbezAUV7JepAkxSE-s3yp6iSFQ/edit

Unfortunately due to Google drive sharing the source version is only
available inside Red Hat. In the future my hope is this will be replaced
by inkscape diagrams where the sources can be kept in git also.

closes #7692


goosemania:
What do you think about having 'storage backend' instead of FS? The original diagram was very specific for katello case.

bmbouter:
@goosemania @daviddavis I pushed a new image, wdyt? More changes are welcome.

daviddavis (approved):

goosemania (approved):

bmbouter:
@melcorr I'm hoping we can put the inkscape file (not sure what file extension) here:  https://github.com/pulp/pulpcore/tree/master/docs/diagrams_src  and then put the png exported version in https://github.com/pulp/pulpcore/tree/master/docs/static

Also related to this I think we should try to convert [the diagrams on this page](https://docs.pulpproject.org/pulpcore/concepts.html) the same way (whenever).

Thank you!

PR #1606, opened 2021-09-02
Fixes worker not starting on openshift

bmbouter:
For some strange reason some Openshift environments do not like the
runtime import the worker entrypoint used. Moving it to the top where it
belongs anyway resolves the problem.

closes #9338

### DIFF
--- a/pulpcore/tasking/entrypoint.py
+++ b/pulpcore/tasking/entrypoint.py
@@ -14,6 +14,7 @@ os.environ.setdefault("DJANGO_ALLOW_ASYNC_UNSAFE", "true")
 django.setup()
 
 from django.conf import settings  # noqa: E402: module level not at top of file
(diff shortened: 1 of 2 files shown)


pulpcore/tasking/entrypoint.py
```diff
@@ -14,6 +14,7 @@
 django.setup()
 
 from django.conf import settings  # noqa: E402: module level not at top of file
+from pulpcore.tasking.pulpcore_worker import NewPulpWorker  # noqa: E402: module level not at top
```
mdellweg, line 17: Funny, this is still past "django.setup()" which it needs to be. But i'm not going to complain here.
There must be something really strange with this python.

mdellweg (approved):

goosemania (approved):

pulpcore/tasking/entrypoint.py
bmbouter: Agreed. To say it's very strange is almost an understatement.

PR #1614, opened 2021-09-08
Disable DJANGO_ALLOW_ASYNC_UNSAFE for tasking system

dralley:
This should no longer be necessary after all the work done to reduce db connections - and if we do encounter issues, we'd be better off trying to address them.

[noissue]

### DIFF
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -15,7 +15,7 @@ from .api import Stage
 log = logging.getLogger(__name__)
 
 
-def _check_for_forbidden_checksume_type(artifact):
+def _check_for_forbidden_checksum_type(artifact):
     """Check if content doesn't have forbidden checksum type.
 
     If contains forbidden checksum type it will raise ValueError,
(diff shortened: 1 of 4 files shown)


dralley:
I don't think we can turn it off for the content app yet, we'll need to do the same work there that was done for the tasking system (for plugins like ansible, rpm, container)

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -87,7 +87,7 @@ async def run(self):
             # swap it out with the existing one.
             for digest_type, digests in artifact_digests_by_type.items():
                 query_params = {"{attr}__in".format(attr=digest_type): digests}
-                existing_artifacts_qs = Artifact.objects.filter(**query_params).only(digest_type)
+                existing_artifacts_qs = Artifact.objects.filter(**query_params)
```
dralley, line 90: Other values are used down on line 391, triggering DB queries.

bmbouter:
Should we have a refactor issue to track this change? I'm not sure, but I wanted to ask.

dralley:
I think yes - if we *potentially* might want to backport it to 3.15, otherwise no strong feeling either way.

bmbouter:
> I think yes - if we _potentially_ might want to backport it to 3.15, otherwise no strong feeling either way.

+1 then, we may want to backport.

dralley:
@bmbouter Done

bmbouter (approved):
This looks good to me!

What's the plan for the other areas that `DJANGO_ALLOW_ASYNC_UNSAFE` is used?

dralley:
I filed https://pulp.plan.io/issues/9354 as well.

dkliban (approved):

fao89:
pulp-ansible and pulp-container are failing now,
https://github.com/pulp/pulp_ansible/runs/3555934111?check_suite_focus=true
https://github.com/pulp/pulp_container/actions/runs/1215755815
I'm not sure yet if the plugins need a change, or if it is pulpcore that needs it

dralley:
@fao89 The plugins - this means they didn't have complete mitigations for the DB connection issue FYI.  But it should be quite easy to find where the problems are since it throws exceptions.

PR #1623, opened 2021-09-15
Update remote artifact urls on sync if the url of the artifact has changed

dralley:
closes: #9395
https://pulp.plan.io/issues/9395

### DIFF
--- a/CHANGES/9395.bugfix
+++ b/CHANGES/9395.bugfix
@@ -1,0 +1,1 @@
+Fixed an issue where on_demand content might not be downloaded properly if the remote URL was changed (even if re-synced).
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -276,13 +276,11 @@ class RemoteArtifactSaver(Stage):
             The coroutine for this stage.
         """
         async for batch in self.batches():
-            await sync_to_async(RemoteArtifact.objects.bulk_get_or_create)(
-                await self._needed_remote_artifacts(batch)
-            )
+            await self._handle_remote_artifacts(batch)
             for d_content in batch:
                 await self.put(d_content)
 
-    async def _needed_remote_artifacts(self, batch):
+    async def _handle_remote_artifacts(self, batch):
         """
         Build a list of only :class:`~pulpcore.plugin.models.RemoteArtifact` that need
         to be created for the batch.
@@ -318,7 +316,8 @@ class RemoteArtifactSaver(Stage):
         #
         # We can end up with duplicates (diff pks, same sha256) in the sequence below,
         # so we store by-sha256 and then return the final values
-        needed_ras = {}  # { str(<sha256>): RemoteArtifact, ... }
+        ras_to_create = {}  # { str(<sha256>): RemoteArtifact, ... }
+        ras_to_update = {}
         for d_content in batch:
             for d_artifact in d_content.d_artifacts:
                 if not d_artifact.remote:
@@ -379,14 +378,25 @@ class RemoteArtifactSaver(Stage):
                 async for remote_artifact in sync_to_async_iterable(
                     content_artifact._remote_artifact_saver_ras
                 ):
-                    if remote_artifact.remote_id == d_artifact.remote.pk:
+                    if d_artifact.url == remote_artifact.url:
+                        break
+
+                    if d_artifact.remote.pk == remote_artifact.remote_id:
+                        key = f"{content_artifact.pk}-{remote_artifact.remote_id}"
+                        remote_artifact.url = d_artifact.url
+                        ras_to_update[key] = remote_artifact
                         break
                 else:
                     remote_artifact = self._create_remote_artifact(d_artifact, content_artifact)
-                    key = f"{str(content_artifact.pk)}-{str(d_artifact.remote.pk)}"
-                    needed_ras[key] = remote_artifact
+                    key = f"{content_artifact.pk}-{d_artifact.remote.pk}"
+                    ras_to_create[key] = remote_artifact
 
-        return list(needed_ras.values())
+        if ras_to_create:
+            await sync_to_async(RemoteArtifact.objects.bulk_create)(list(ras_to_create.values()))
+        if ras_to_update:
+            await sync_to_async(RemoteArtifact.objects.bulk_update)(
+                list(ras_to_update.values()), fields=["url"]
+            )
 
     @staticmethod
     def _create_remote_artifact(d_artifact, content_artifact):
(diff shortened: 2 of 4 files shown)


dralley:
**Discussion needed:** https://pulp.plan.io/issues/9395#note-6

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -379,14 +378,25 @@ async def _needed_remote_artifacts(self, batch):
                 async for remote_artifact in sync_to_async_iterable(
                     content_artifact._remote_artifact_saver_ras
                 ):
-                    if remote_artifact.remote_id == d_artifact.remote.pk:
+                    if d_artifact.remote.url == remote_artifact.url:
+                        break
+
+                    if d_artifact.remote.pk == remote_artifact.remote_id:
+                        key = f"{str(content_artifact.pk)}-{str(d_artifact.remote.pk)}"
+                        remote_artifact.url = d_artifact.url
```
dralley, line 386: TL;DR:

If the relative path of the artifact within the repo has changed, or if the base path of the repo has changed, we update the RemoteArtifact to match the new URL

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -379,14 +378,25 @@ async def _needed_remote_artifacts(self, batch):
...
+                        ras_to_update[key] = remote_artifact
                         break
                 else:
                     remote_artifact = self._create_remote_artifact(d_artifact, content_artifact)
                     key = f"{str(content_artifact.pk)}-{str(d_artifact.remote.pk)}"
-                    needed_ras[key] = remote_artifact
+                    ras_to_create[key] = remote_artifact
 
-        return list(needed_ras.values())
```
gerrod3, line 389: Little confused on why a value was being returned before (doesn't seem it was getting used).

dralley: It was being used - the `bulk_create()` (actually `bulk_get_or_create()` was happening externally.

gerrod3: I see, all these wrappers make things confusing.

dralley: Yup.  I hope Django does get native async ORM soon...

dralley: Speaking of: https://github.com/django/django/pull/14843

pulpcore/plugin/stages/artifact_stages.py
ipanova: Here are some specific examples on how this problem can affect plugins:
1. RPM plugin: part of the [remote_artifact.url](https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/tasks/synchronizing.py#L1040-L1050) is composed of `location_href` which is specified in the repodata. It might happen so that it will change in the remote repo , and with the next sync no new RemoteArtifacts will be created because of [uniqueness constraint]( https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/content.py#L663) and old RemoteArtifacts would point to an invalid url and as a result  user will get 404.
2. File plugin: create and sync a `local_repoA` with on_demand policy from `https://remote.repos.com/repoA/PULP_MANIFEST`. Then update `remote.url` to point to `https://remote.repos.com/exact_copy_of_repoA/PULP_MANIFEST`. Imagine that `remote.repos.com/repoA`. has been removed and only `remote.repos.com/exact_copy_of_repoA` is left. Re-sync `local_repoA`. Users will get 404 afterwards because `remote_artifact.url` will still point to old unavailable url and not new remote_artifacts would be created.
3. Container plugin: similar to file plugin workflow just the url to the registry would change

I'd prefer to change uniqueness constraint of the RemoteArtifact so it also contains the `url` in addition to `content_artifact` and `remote` however due to outlined by @dralley reasons that this would not be backportable  taken by him approach makes sense.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -379,14 +378,25 @@ async def _needed_remote_artifacts(self, batch):
                 async for remote_artifact in sync_to_async_iterable(
                     content_artifact._remote_artifact_saver_ras
                 ):
-                    if remote_artifact.remote_id == d_artifact.remote.pk:
+                    if d_artifact.remote.url == remote_artifact.url:
```
mdellweg, line 381:
```suggestion
                    if d_artifact.url == remote_artifact.url:
```

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -379,14 +378,25 @@ async def _needed_remote_artifacts(self, batch):
...
+                        break
+
+                    if d_artifact.remote.pk == remote_artifact.remote_id:
+                        key = f"{str(content_artifact.pk)}-{str(d_artifact.remote.pk)}"
```
mdellweg, line 385: Do we need that key? The only use of it i see is as a key for the dictionary. So it imposes some uniqueness.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -379,14 +378,25 @@ async def _needed_remote_artifacts(self, batch):
                 async for remote_artifact in sync_to_async_iterable(
                     content_artifact._remote_artifact_saver_ras
                 ):
-                    if remote_artifact.remote_id == d_artifact.remote.pk:
+                    if d_artifact.url == remote_artifact.url:
+                        break
+
+                    if d_artifact.remote.pk == remote_artifact.remote_id:
+                        key = f"{str(content_artifact.pk)}-{str(remote_artifact.remote_id)}"
```
bmbouter, line 385: This could also be changed like so because of auto-string conversion of fstrings

```suggestion
                        key = f"{content_artifact.pk}-{remote_artifact.remote_id}"
```

CHANGES/9395.bugfix
```diff
@@ -0,0 +1 @@
+Fixed an issue where on_demand content might not be downloaded properly if the remote URL was changed (even if re-synced).
```
bmbouter, line 1: How can we write an automated test for this?

bmbouter: I think if we had a pulp_file fixture that had a different layout but the same content we could test this by:

1. sync one of the fixtures with on_demand
2. change the remote.url to the other fixture
3. observe the 404
4. sync
5. observe the 404's go away

dralley: I think we cannot, because a change to the layout would yield a different set of content, because `relative_path` is part of it.

dralley: @bmbouter Test added

mdellweg (approved):
My concerns have been addressed.

ipanova (approved):
Thank you for adding the test!

CHANGES/9395.bugfix
bmbouter: Thank you! +1

PR #1624, opened 2021-09-15
Do bulk_touch instead of touching one by one

lubosmj:
closes #9266

### DIFF
--- a/CHANGES/9266.bugfix
+++ b/CHANGES/9266.bugfix
@@ -1,0 +1,1 @@
+Fixed the repository modify endpoint performance problems.
--- a/pulpcore/app/tasks/repository.py
+++ b/pulpcore/app/tasks/repository.py
@@ -11,6 +11,7 @@ from rest_framework.serializers import ValidationError
 
 from pulpcore.app import models
 from pulpcore.app.models import ProgressReport
+
 from pulpcore.plugin.sync import sync_to_async_iterable
 
 log = getLogger(__name__)
--- a/pulpcore/plugin/actions.py
+++ b/pulpcore/plugin/actions.py
@@ -1,6 +1,7 @@
-from django.db import DatabaseError
+from gettext import gettext as _
 from drf_spectacular.utils import extend_schema
 from rest_framework.decorators import action
+from rest_framework.serializers import ValidationError
 
 from pulpcore.app import tasks
 from pulpcore.app.models import Content, RepositoryVersion
@@ -9,6 +10,7 @@ from pulpcore.app.serializers import (
     AsyncOperationResponseSerializer,
     RepositoryAddRemoveContentSerializer,
 )
+from pulpcore.app.viewsets import NamedModelViewSet
 from pulpcore.tasking.tasks import dispatch
 
 
@@ -26,8 +28,9 @@ class ModifyRepositoryActionMixin:
         """
         Queues a task that creates a new RepositoryVersion by adding and removing content units
         """
-        add_content_units = []
-        remove_content_units = []
+        add_content_units = {}
+        remove_content_units = {}
+
         repository = self.get_object()
         serializer = self.get_serializer(data=request.data)
         serializer.is_valid(raise_exception=True)
@@ -39,22 +42,26 @@ class ModifyRepositoryActionMixin:
 
         if "add_content_units" in request.data:
             for url in request.data["add_content_units"]:
-                content = self.get_resource(url, Content)
-                try:
-                    content.touch()
-                except DatabaseError:
-                    # content has since been removed. call get_url to raise an exception.
-                    content = self.get_resource(url, Content)
-                add_content_units.append(content.pk)
+                add_content_units[NamedModelViewSet.extract_pk(url)] = url
+
+            content_units_pks = set(add_content_units.keys())
+            existing_content_units = Content.objects.filter(pk__in=content_units_pks)
+            existing_content_units.touch()
+
+            self.verify_content_units(existing_content_units, add_content_units)
+
+            add_content_units = list(add_content_units.keys())
 
         if "remove_content_units" in request.data:
-            for url in request.data["remove_content_units"]:
-                if url == "*":
-                    remove_content_units = [url]
-                    break
-                else:
-                    content = self.get_resource(url, Content)
-                    remove_content_units.append(content.pk)
+            if "*" in request.data["remove_content_units"]:
+                remove_content_units = ["*"]
+            else:
+                for url in request.data["remove_content_units"]:
+                    remove_content_units[NamedModelViewSet.extract_pk(url)] = url
+                content_units_pks = set(remove_content_units.keys())
+                existing_content_units = Content.objects.filter(pk__in=content_units_pks)
+                self.verify_content_units(existing_content_units, remove_content_units)
+                remove_content_units = list(remove_content_units.keys())
 
         task = dispatch(
             tasks.repository.add_and_remove,
@@ -67,3 +74,15 @@ class ModifyRepositoryActionMixin:
             },
         )
         return OperationPostponedResponse(task, request)
+
+    def verify_content_units(self, content_units, all_content_units):
+        """Verify referenced content units."""
+        existing_content_units_pks = content_units.values_list("pk", flat=True)
+        existing_content_units_pks = {str(pk) for pk in existing_content_units_pks}
+
+        missing_pks = set(all_content_units.keys()) - existing_content_units_pks
+        if missing_pks:
+            missing_hrefs = [all_content_units[pk] for pk in missing_pks]
+            raise ValidationError(
+                _("Could not find the following content units: {}").format(missing_hrefs)
+            )


pulpcore/app/tasks/repository.py
```diff
@@ -221,6 +225,15 @@ def add_and_remove(repository_pk, add_content_units, remove_content_units, base_
         else:
             remove_content_units = []
 
+    existing_add_content_units = models.Content.objects.filter(pk__in=add_content_units)
+    pks = {str(pk) for pk in existing_add_content_units.values_list("pk", flat=True)}
+
+    non_existing_units = set(add_content_units) - set(pks)
+    if non_existing_units:
+        raise RuntimeError(non_existing_units)
```
daviddavis, line 233: I think this error needs a little more information for the user (e.g. `Could not find content units {pks}.`).

Also, the redmine issue says that the error should have a list of hrefs (not pks). I don't mind using pks in the error message but I think @dkliban insisted we use hrefs.

pulpcore/plugin/actions.py
```diff
@@ -38,14 +38,9 @@ def modify(self, request, pk):
...
-                content = self.get_resource(url, Content)
-                try:
-                    content.touch()
-                except DatabaseError:
-                    # content has since been removed. call get_url to raise an exception.
-                    content = self.get_resource(url, Content)
-                add_content_units.append(content.pk)
+            add_content_units.extend(
+                [NamedModelViewSet.extract_pk(uri) for uri in request.data["add_content_units"]]
+            )
```
daviddavis, line 43: We should still touch the content units in the viewset but using a bulk query. I think you should add this line:

```
Content.objects.filter(pk__in=add_content_units).touch()
```

pulpcore/app/tasks/repository.py
dralley: It's going to be a little complicated because for the lookup you need PKs (because we can't do those one at a time) and for the error you need the HREFs.

Probably we need to keep a mapping of PK->HREF somewhere and use that to figure out the HREFs for the missing PKs.

pulpcore/plugin/actions.py
lubosmj: By the way, cannot this be done somehow implicitly? I do understand that one needs to touch content all the time due to the orphan cleanup workflow, but may we think about a smarter way of "refreshing" the content? Can we hook some timers to all database objects that will be triggered once a while and when there is no reverse reference to, e.g., a repository object, the objects will automatically clean themselves up? Otherwise, the timers will be reset. Maybe, this was already discussed; then, ignore me.

pulpcore/plugin/actions.py
```diff
@@ -38,14 +38,10 @@ def modify(self, request, pk):
             base_version_pk = None
 
         if "add_content_units" in request.data:
-            for url in request.data["add_content_units"]:
-                content = self.get_resource(url, Content)
```
ipanova, line 42: there is no modify endpoint in the pulp-container however there is a similar one `add`. Can these changes be applied there as well?

pulpcore/plugin/actions.py
```diff
@@ -38,14 +38,10 @@ def modify(self, request, pk):
...
-                    # content has since been removed. call get_url to raise an exception.
-                    content = self.get_resource(url, Content)
-                add_content_units.append(content.pk)
+            add_content_units.extend(
+                [NamedModelViewSet.extract_pk(uri) for uri in request.data["add_content_units"]]
+            )
+            Content.objects.filter(pk__in=add_content_units).touch()
 
         if "remove_content_units" in request.data:
             for url in request.data["remove_content_units"]:
```
ipanova, line 47: how about we also use `extract_pk` here?

pulpcore/app/tasks/repository.py
```diff
@@ -221,6 +232,46 @@ def add_and_remove(repository_pk, add_content_units, remove_content_units, base_
...
+        add_content_units, add_content_units_hrefs
+    )
+
     with repository.new_version(base_version=base_version) as new_version:
         new_version.remove_content(models.Content.objects.filter(pk__in=remove_content_units))
-        new_version.add_content(models.Content.objects.filter(pk__in=add_content_units))
+        new_version.add_content(existing_add_content_units)
+
+
+def touch_existing_content_units(add_content_units, add_content_units_hrefs):
```
ipanova, line 244: In my understanding, the primary goal of the additional checks that should happen inside of the task is to determine whether there is any diff in the list of pks passed into the task vs list of the pks of the content that was returned inside of the task. If there is a diff it means that there was removed content( already not existing content at the time bulk touch was called in the viewset)  and an error should be raised.
I don't think we should `touch` again the content since it was already touched in the viewset, therefore I'd rename this method to reflect what's happening inside of it.
@mdellweg I remember we had a discussion about whether it makes sense to double - touch the content 1) in the viewset and then 2) inside of the task and I *think* we came to the conclusion that it is unnecessary.

mdellweg: I agree. In the task, touch should not be necessary, because we intend to add that content to a repo version immediately.
If we are in a transaction, we can do something like `select_for_update` to add a db lock on the content.

pulpcore/plugin/actions.py
daviddavis: > I do understand that one needs to touch content all the time due to the orphan cleanup workflow, but may we think about a smarter way of "refreshing" the content?

Content that is attached to a repository won't be cleaned up. Check out the [orphan cleanup task](https://github.com/pulp/pulpcore/blob/6068e18f117da33025d702ded971c276084c3d49/pulpcore/app/tasks/orphan.py#L53-L55) and [`orphaned` qs method](https://github.com/pulp/pulpcore/blob/6068e18f117da33025d702ded971c276084c3d49/pulpcore/app/models/content.py#L451). We don't need to worry about resetting the `timestamp_of_interest` for this Content. However, if the user is adding content to a repository, there is a good chance that this content is orphan content. So it needs to be touched. 

There were some alternate proposals we considered, two solutions were outlined on the issue, and there was discussion you can check out:

https://pulp.plan.io/issues/7659

daviddavis:
I think this code could be greatly simplified if we took a different approach and handled missing hrefs in the viewset. We wouldn't need to edit the task code at all. Here's a proof of concept:

https://gist.github.com/daviddavis/f1568eb05d86786cb0167a2d1a8e24f0

lubosmj:
I followed David's approach. Thank you for sharing that!

dralley:
I'm on PTO tomorrow, but I like this approach also, and apart from the CI it looks good to me.

I did notice that in the CI logs it is complaining that there are coroutines in the stages pipeline that aren't awaited.  I think that is a problem.

lubosmj:
The error should be now resolved. It was related to improper usage of the function `extract_pk`. The investigation is no longer needed.

pulpcore/plugin/actions.py
ipanova: `get_resource` will raise an error if the resource is not longer in pulp. I was thinking to re-use `touch_content_units` for both add and remove content units except for the touch call that can happen outside of the function.

lubosmj: I updated the PR with the proposed changes.

pulpcore/plugin/actions.py
```diff
@@ -26,8 +28,9 @@ def modify(self, request, pk):
         """
         Queues a task that creates a new RepositoryVersion by adding and removing content units
         """
-        add_content_units = []
-        remove_content_units = []
+        add_content_units = {}
```
daviddavis, line 31: I'm surprised this works. You don't have to convert these to lists somehow before sending them to the task?

CHANGES/9266.bugfix
```diff
@@ -0,0 +1 @@
+Refactored a method which updates content's timestamps checked by the orphan cleanup workflow.
```
daviddavis, line 1: I don't think users will understand this changelog message. What about:

```suggestion
Addressed repository version modify endpoint performance problems.
```

pulpcore/plugin/actions.py
```diff
@@ -39,22 +42,32 @@ def modify(self, request, pk):
...
+            content_units_pks = set(add_content_units.keys())
+            existing_content_units = Content.objects.filter(pk__in=content_units_pks)
+            existing_content_units.touch()
+
+            self.verify_content_units(existing_content_units, add_content_units)
+
+            add_content_units = list(add_content_units.keys())
 
         if "remove_content_units" in request.data:
+            removing_all_units = False
```
daviddavis, line 56: Couldn't you just do:

```
if "remove_content_units" in request.data:
    if "*" in request.data["*"]:
        remove_content_units = ["*"]  # or {"*": "*"} if you're using a dict
    else:
        for url in request.data["remove_content_units"]:
        # ...
```

daviddavis (approved):
LGTM. Two small suggestions.

CHANGES/9266.bugfix
ipanova: I agree, or otherwise move this to .misc if you position this fix as a refactor.

ipanova:
we need a backport issue for 3.14

ipanova (approved):

PR #1625, opened 2021-09-15
Use fields attname for natural key calculations

mdellweg:
fixes #9409

~This is a suggestion. It would need a ticket if this is desired.~

### DIFF
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -11,6 +11,7 @@ import subprocess
 
 import gnupg
 
+from functools import lru_cache
 from itertools import chain
 
 from django.conf import settings
@@ -494,9 +495,22 @@ class Content(MasterModel, QueryMixin):
     @classmethod
     def natural_key_fields(cls):
         """
-        Returns a tuple of the natural key fields which usually equates to unique_together fields
+        Returns a tuple of the natural key fields which usually equates to unique_together fields.
+
+        This can be overwritten in subclasses and should return a tuple of field names.
         """
         return tuple(chain.from_iterable(cls._meta.unique_together))
+
+    @classmethod
+    @lru_cache(typed=True)
+    def _sanitized_natural_key_fields(cls):
+        """
+        This function translates the names of the key fields to their attname.
+
+        In case of foreign keys, this decodes to the corresponding `<...>_id` field preventing
+        extra DB accesses to fetch the related objects.
+        """
+        return tuple(getattr(cls, field).field.attname for field in cls.natural_key_fields())
 
     def natural_key(self):
         """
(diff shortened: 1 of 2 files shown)


mdellweg:
This probably _needs_ caching on the class level.

pulpcore/app/models/content.py
```diff
@@ -498,23 +498,24 @@ def natural_key_fields(cls):
         """
         return tuple(chain.from_iterable(cls._meta.unique_together))
 
+    @classmethod
+    def _sanitized_natural_key_fields(cls):
+        return tuple(getattr(cls, field).field.attname)
```
daviddavis, line 503: Where is field defined?

mdellweg: My bad. Look again...

daviddavis:
Looks good now. I'm curious about how the field name pulled from the unique_together and the field's attname would differ?

pulpcore/app/models/content.py
mdellweg: This abstraction function seems to be needed, since plugins use `natural_key_fields` in their `model_resources`, and they do not support the `*_id` field.
Also it much more natural for the plugin writer if overwriting the `natural_key_fields` function to refer to the actual field names.

mdellweg:
> Looks good now. I'm curious about how the field name pulled from the unique_together and the field's attname would differ?

They differ in the case of a ForeignKeyField. See the discussion: https://github.com/pulp/pulp_container/pull/396

pulpcore/app/models/content.py
```diff
@@ -498,23 +498,24 @@ def natural_key_fields(cls):
         """
         return tuple(chain.from_iterable(cls._meta.unique_together))
 
+    @classmethod
+    def _sanitized_natural_key_fields(cls):
```
mdellweg, line 502: `from functools import lru_cache`

```suggestion
    @lru_cache(typed=True)
    def _sanitized_natural_key_fields(cls):
```

Should speed this up.

daviddavis: Can you leave a docstring for this method to indicate why it's needed?

dralley:
The thing to be most careful with is probably import/export

mdellweg:
> The thing to be most careful with is probably import/export, since that's what broke the other PR originally

PIE seems to use `natural_key_fields' for the `import_id_fields` in some plugins. Therefore i did not touch that function (only its docstring...).

pulpcore/app/models/content.py
```diff
@@ -494,27 +495,37 @@ class Meta:
...
+    @classmethod
+    @lru_cache(typed=True)
+    def _sanitized_natural_key_fields(cls):
+        """
+        This function translates the names of the key fields to their attname.
+
+        In case of foreign keys, this decodes to the corresponding `<...>_id` field preventing
+        extra DB accesses to fetch the related objects.
+        """
+        return tuple(getattr(cls, field).field.attname for field in cls.natural_key_fields())
```
ipanova, line 513: looks good! :100:

ipanova (approved):

daviddavis (approved):

PR #1627, opened 2021-09-16
Roles

mdellweg:


CHANGES/9411.feature
```diff
@@ -0,0 +1 @@
+Added a role model to support RBAC.
```
bmbouter, line 1: We'll need additional deprecation notices for all methods that will be removed with the removal of django-guardian.

pulpcore/app/settings.py
```diff
@@ -118,6 +118,7 @@
 AUTHENTICATION_BACKENDS = [
     "django.contrib.auth.backends.ModelBackend",
     "guardian.backends.ObjectPermissionBackend",
```
bmbouter, line 120: This should get a deprecation notice too I think.

pulpcore/app/viewsets/task.py
```diff
@@ -93,11 +93,15 @@ class TaskViewSet(
         ],
         "permissions_assignment": [
             {
-                "function": "add_for_object_creator",
+                "function": "add_roles_for_object_creator",
                 "parameters": None,
-                "permissions": ["core.view_task", "core.change_task", "core.delete_task"],
+                "permissions": "core.task_owner",
```
bmbouter, line 98: This part seems strange to me. `task_owner` is a role not a permission. I imagined there would be a new `role` entry and the usage of `permissions` would be deprecated with 3.16.

Or maybe we should just use `parameters` and let the roles to be added just be args to the function, e.g. `add_roles_for_object_creator`.

pulpcore/tests/unit/roles/test_user_roles.py
```diff
@@ -0,0 +1,65 @@
...
+            Permission.objects.get(content_type__app_label="core", codename="view_repository")
+        )
+        self.role2 = Role.objects.create(name="role2")
+        self.role2.permissions.add(
+            Permission.objects.get(content_type__app_label="core", codename="view_remote")
+        )
+        self.repository = Repository.objects.create(name=uuid4())
+        self.remote = Remote.objects.create(name=uuid4())
+
+    def test_user_role(self):
```
bmbouter, line 31: I do think breaking these up would be good.

pulpcore/tests/unit/roles/test_user_roles.py
```diff
@@ -0,0 +1,65 @@
...
+        self.role1 = Role.objects.create(name="role1")
+        self.role1.permissions.add(
+            Permission.objects.get(content_type__app_label="core", codename="view_repository")
+        )
+        self.role2 = Role.objects.create(name="role2")
+        self.role2.permissions.add(
+            Permission.objects.get(content_type__app_label="core", codename="view_remote")
+        )
+        self.repository = Repository.objects.create(name=uuid4())
+        self.remote = Remote.objects.create(name=uuid4())
```
bmbouter, line 29: What about making these pytest fixtures? Just an idea; totally optional.

bmbouter:
Overall everything I'm seeing here is going in the right direction. It needs docs updates, deprecation warnings and notices (if that's happening in this PR), and queryset filtering. Thank you for working on this! I did not run the code itself, but I will after its undrafted.

pulpcore/tests/unit/roles/test_user_roles.py
mdellweg: I'd rather make the move to pytest intentional and not sneak it in here.

bmbouter: pytest is already the official runner, but that's ok with me.

mdellweg: I don't think, this is true for the unittests. They are run via admin.py which to my knowledge uses the unittest framework.

CHANGES/9411.feature
mdellweg: If you don't mind, i would treat the removal of guardian (with its deprecation phase) as another piece of work for now.

pulpcore/app/viewsets/task.py
mdellweg: Should we remove the "permissions" completely from the access policy?
Look at my current draft for this.

pulpcore/backends.py
```diff
@@ -0,0 +1,97 @@
...
+                    return True
+            else:
+                raise RuntimeException(
+                    _("Permission {} is not suitable for objects of class {}.").format(
+                        perm, obj.__class__
+                    )
+                )
+        return False
+
+    def get_all_permissions(self, user_obj, obj=None):
```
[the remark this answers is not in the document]
mdellweg: This name is a given interface to django auth.
When you call `user.get_all_permissions` it will ask all its backends via this function.

pulpcore/backends.py
```diff
@@ -0,0 +1,97 @@
...
+from django.contrib.contenttypes.models import ContentType
+from django.contrib.auth.backends import BaseBackend
+from django.contrib.auth.models import Permission
+
+from pulpcore.app.models import GroupRole
+
+
+class ObjectRolePermissionBackend(BaseBackend):
+    def has_perm(self, user_obj, perm, obj=None):
+        if not user_obj.is_authenticated:
```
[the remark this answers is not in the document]
mdellweg: That is covered by "backend.django.contrib.auth.backends.ModelBackend".

pulpcore/app/viewsets/task.py
```diff
@@ -93,11 +93,15 @@ class TaskViewSet(
         ],
         "permissions_assignment": [
```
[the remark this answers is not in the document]
bmbouter: Maybe `creation_hooks` or some other name?

pulpcore/app/viewsets/task.py
```diff
@@ -93,11 +93,15 @@ class TaskViewSet(
...
             {
-                "function": "add_for_object_creator",
-                "parameters": None,
-                "permissions": ["core.view_task", "core.change_task", "core.delete_task"],
+                "function": "add_roles_for_object_creator",
+                "parameters": {"roles": "core.task_owner"},
+                "permissions":  None,
```
bmbouter, line 98: I believe `permissions` will be unused going forward so it would be good to add a deprecated release note.

pulpcore/app/viewsets/base.py
```diff
@@ -333,7 +334,9 @@ def get_queryset(self):
 
         permission_name = getattr(self, "queryset_filtering_required_permission", None)
         if permission_name:
-            qs = get_objects_for_user(self.request.user, permission_name, klass=qs)
+            qs = get_objects_for_user(
```
bmbouter, line 337: Can this check that the `pulpcore.backends.ObjectRolePermissionBackend` is configured as an AUTHENTICATION_BACKEND?

bmbouter: And the same for guardian too plz.

pulpcore/app/settings.py
```diff
@@ -118,6 +118,7 @@
...
+    "pulpcore.backends.ObjectRolePermissionBackend",
```
bmbouter, line 121: We should think about how AccessPolicies can be fully disabled if this is removed.

bmbouter:
One more idea I forgot to mention. I think we should have automated tests for this. Particularly the big-hammer part.

pulpcore/backends.py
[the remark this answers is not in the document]
mdellweg: Maybe even by the User model itself...
https://docs.djangoproject.com/en/3.2/ref/contrib/auth/#django.contrib.auth.models.User.has_perm

pulpcore/app/util.py
```diff
@@ -159,3 +161,72 @@ def get_request_without_query_params(context):
...
+    """
+    Assign a role to a user or a group.
+
+    Args:
+        rolename (str): Name of the role to assign.
+        entity (django.contrib.auth.User or django.contrib.auth.Group): Entity to gain the role.
+        obj (Optional[pulpcore.app.models.BaseModel]): Object the role permisssions are to be
+            asserted on.
+    """
+    role = models.Role.objects.get(name=rolename)
```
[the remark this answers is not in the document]
mdellweg: What do you expect we do in the case of a caught exception here? Raise another one?
Opted for `BadRequest`.

pulpcore/app/settings.py
mdellweg: I think, we "just" need to find all instances where we use a guardian identifier and hide it beyond a proxy function.

pulpcore/app/models/access_policy.py
```diff
@@ -104,8 +113,18 @@ def add_perms(self):
...
+            if function is not None:
+                kwargs = creation_hook.get("parameters") or {}
+                function(**kwargs)
+            else:
+                # Old interface deprecated for removal in 3.17
+                function = getattr(self, creation_hook["function"])
+                deprecation_logger.warn(
+                    "Calling unregistered creation hooks from the access policy is deprecated and"
+                    " may be removed with pulpcore 3.17."
+                )
```
mdellweg, line 126: I thought when touching this piece of code we can just as well make it much more secure.
Before this you were able to call just about any method on that class that accepts 2 arguments.

pulpcore/app/viewsets/publication.py
```diff
@@ -185,7 +185,7 @@ class RBACContentGuardViewSet(ContentGuardViewSet):
                 "condition": "has_model_or_obj_perms:core.download_rbaccontentguard",
             },
         ],
-        "permissions_assignment": [
+        "creation_hooks": [
```
gerrod3, line 188: Is there a issue filed for changing the RBAC content guard over to roles?

pulpcore/app/migrations/0079_rename_permissions_assignment_accesspolicy_creation_hooks.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.7 on 2021-09-22 12:55
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0078_accesspolicy_roles'),
```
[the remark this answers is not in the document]
mdellweg: I didn't bother merging them. But it looks like we may be dropping the roles from the access policy anyway now.

mdellweg:
Look also:
https://github.com/pulp/pulp-cli/pull/382

pulpcore/app/viewsets/user.py
```diff
@@ -440,3 +444,93 @@ def destroy(self, request, group_pk, pk):
...
+):
+    """
+    ViewSet for Role.
+
+    NOTE: This API endpoint is in "tech preview" and subject to change
+
+    """
+
+    endpoint_name = "roles"
+    router_lookup = "role"
```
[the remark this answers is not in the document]
mdellweg: According to docs, i believe it should be "user" and we are not using it correctly anywhere. But also it does not seem to have any effect either.
https://github.com/alanjds/drf-nested-routers/blob/master/rest_framework_nested/routers.py#L148

pulpcore/app/serializers/user.py
```diff
@@ -2,21 +2,59 @@
...
+)
 from pulpcore.app.util import get_viewset_for_model, get_request_without_query_params
 
 
-class PermissionField(serializers.CharField):
-    """Permission field."""
+class ThisPermissionField(serializers.CharField):
+    """Read only field that represents the Permission of the object.
+    The object can be either a plain Permission, or GroupObjectPermission.
+    """
```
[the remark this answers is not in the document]
mdellweg: I chose this name, because it is significantly different to the permission field. And it is meant to describe the permission the whole object to serialize is supposed to represent. (Also it belongs to the interfaces that will eventually be removed with this effort.)

mdellweg:
@gerrod3 thanks for the review! The group roles are indeed my next step, now that i know the user_role interface is not too crazy.

pulpcore/app/viewsets/user.py
```diff
@@ -440,3 +444,93 @@ def destroy(self, request, group_pk, pk):
...
+
+
+class UserRoleViewSet(
+    NamedModelViewSet,
+    mixins.CreateModelMixin,
+    mixins.RetrieveModelMixin,
+    mixins.ListModelMixin,
+    mixins.DestroyModelMixin,
+):
+
```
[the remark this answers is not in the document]
gerrod3: The endpoints that get created from the mixins have the names/tags (in the docs)
```
List user roles
Create an user role
Inspect an user role
Delete an user role
```
Besides the grammar error, could these be renamed to reflect the user receiving/having the role, e.g. 
```
List user's roles
Assign role to user
Inspect a user's role
Remove role from user
```
I know that `UserRole` is its own object, but to users I think that it's an implementation detail they shouldn't worry about. You already renamed these in the CLI commands.

pulpcore/app/viewsets/user.py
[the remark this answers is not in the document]
gerrod3: Weird, it's possible this could affect galaxy (a tiny bit) if they want to change the routing, but I'm not even sure if they use things like this in the first place.

pulpcore/app/viewsets/user.py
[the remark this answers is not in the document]
mdellweg: Apart from the fact that I don't know how to do that, I am torn here. Those are really restful operations on the UserRole object as you mentioned.
The cli OTOH is definitely abstracting enough to call it assign and remove.

pulpcore/app/viewsets/user.py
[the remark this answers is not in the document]
mdellweg: Interesting. We should keep an eye on that. (Possibly we circumvent using this stuff by defining `parent_lookup_kwargs`. Didn't fully understand it yet.)

pulpcore/app/serializers/user.py
```diff
@@ -135,3 +196,131 @@ class GroupSerializer(ValidateFieldsMixin, serializers.ModelSerializer):
...
+
+class UserRoleSerializer(ModelSerializer, NestedHyperlinkedModelSerializer):
+    """Serializer for UserRole."""
+
+    pulp_href = NestedIdentityField(
+        view_name="roles-detail", parent_lookup_kwargs={"user_pk": "user__pk"}
+    )
+
+    role = serializers.SlugRelatedField(
+        slug_field="name",
```
goosemania, line 251: Nice!I didn't know about the slug field.

pulpcore/app/models/role.py
```diff
@@ -0,0 +1,42 @@
...
+    user = models.ForeignKey(
+        get_user_model(), related_name="object_roles", on_delete=models.CASCADE
+    )
+    role = models.ForeignKey(Role, related_name="object_users", on_delete=models.CASCADE)
+    content_type = models.ForeignKey(ContentType, on_delete=models.SET_NULL, null=True)
+    object_id = models.UUIDField(null=True)
+    content_object = GenericForeignKey("content_type", "object_id")
+
+    class Meta:
+        unique_together = (("user", "role", "content_type", "object_id"),)
```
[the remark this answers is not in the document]
mdellweg: object_id can be a uuid or an integer, depending on the object referenced. And (though _very_ unlikely) we don't have a uniqueness constraint for uuid's across models either. So i'd say, yes.

pulpcore/app/migrations/0078_grouprole_role_userrole.py
```diff
@@ -0,0 +1,67 @@
...
+import django.db.models.deletion
+import django_lifecycle.mixins
+import uuid
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('contenttypes', '0002_remove_content_type_name'),
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
```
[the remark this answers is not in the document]
mdellweg: This is auto-generated. And it was introduced, because the model definition contains `get_user_model`.

docs/plugins/plugin-writer/concepts/rbac/access_policy.rst
```diff
@@ -131,9 +131,10 @@ Shipping a Default Access Policy
 --------------------------------
 
 To ship a default access policy, define a dictionary named ``DEFAULT_ACCESS_POLICY`` as a class
-attribute on a subclass of ``NamedModelViewSet`` containing both ``statements`` and
-``permissions_assignment``. The ``AccessPolicy`` instance will be then be created in the
-``pulp_migrate`` signal handler.
+attribute on a subclass of ``NamedModelViewSet`` containing all of ``statements`` and
+``creation_hooks``. The ``AccessPolicy`` instance will be then be created in the
```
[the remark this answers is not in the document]
mdellweg:
```suggestion
``creation_hooks``. The ``AccessPolicy`` instance will then be created in the
```
To be fair, it was there before. :angel:

pulpcore/app/apps.py
```diff
@@ -194,37 +197,117 @@ def ready(self):
...
-                    print(f"Access policy for {viewset_name} updated.")
+                    if verbosity >= 1:
+                        print(
+                            _("Access policy for {viewset_name} updated.").format(
+                                viewset_name=viewset_name
+                            )
+                        )
+
+
+def _populate_roles(sender, apps, verbosity, **kwargs):
```
newswangerd, line 255: @mdellweg @bmbouter It would be great if you could split this function up so that the logic for loading the role definitions from the viewsets is separate from the logic for setting it in the database.

The reason for this is that the galaxy_ng APIs aren't currently discoverable via `sender.named_viewsets` because pulpcore expects a plugins viewsets to live under `<plugin_name>.viewsets` and ours currently don't. This means that we can't define the roles in LOCKED_ROLES on our viewsets. If you could provide a function that takes in a list of system roles and sets them in the database, that would save us a lot of time and duplicate code.

mdellweg: How is that?

newswangerd: @mdellweg thanks! @bmclaughlin is going to take a look at it

pulpcore/app/serializers/user.py
```diff
@@ -2,21 +2,63 @@
...
+from pulpcore.app.models import GroupRole, Role, UserRole
+from pulpcore.app.serializers import (
+    NestedIdentityField,
+    IdentityField,
+    ValidateFieldsMixin,
+    ModelSerializer,
+)
 from pulpcore.app.util import get_viewset_for_model, get_request_without_query_params
 
+User = get_user_model()
```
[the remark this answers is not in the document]
mdellweg: @newswangerd Do you think that this line could maybe cause your circular import issue?

pulpcore/app/models/role.py
```diff
@@ -0,0 +1,71 @@
...
+    Relations:
+
+        group (models.ForeignKey): Group to grant permissions to.
+        role (models.ForeignKey): Role to select granted permissions from.
+        content_object (GenericForeignKey): Optional object to assert permissions on.
+    """
+
+    group = models.ForeignKey(Group, related_name="object_roles", on_delete=models.CASCADE)
+    role = models.ForeignKey(Role, related_name="object_groups", on_delete=models.CASCADE)
+    content_type = models.ForeignKey(ContentType, on_delete=models.SET_NULL, null=True)
```
goosemania, line 66: @mdellweg , I'm looking at it from the plugin removal perspective. If plugin is removed and its content_types are removed, there will be no cascade delete and this GroupRole (or similarly UserRole) will reference a non-existing object with content_type=null, do I read this correctly?
Should plugin removal code clean this table up for content types it is about to remove?

mdellweg: That is a good question. I think, guardian introduced a (regular) cleanup process to clean up stale ObjectPermissions. But cleanup when we know it's useful sound better.

pulpcore/app/models/role.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Relations:
+
+        permissions (models.ManyToManyField): Permissions to be granted via this role.
+    """
+
+    name = models.CharField(max_length=128, db_index=True, unique=True)
+    locked = models.BooleanField(default=False)
+    permissions = models.ManyToManyField(Permission)
+
```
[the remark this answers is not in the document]
mdellweg: Plugin managed roles will have the plugin label as a prefix. But you could also identify them by reading the LOCKED_ROLES attribute. I'd say that is preferable for more specific removal.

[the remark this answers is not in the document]
goosemania: Is it enforced or just a convention? If it's convention it's not reliable.
LOCKED_ROLES can be used but all roles responsible for the plugin should be removed, locked or not. Would you agree?
Or it is preferable to keep user-defined roles in place even if some or all of the permissions from it are gone?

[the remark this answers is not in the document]
mdellweg: It is enforced for LOCKED_ROLES. The user can basically do whatever. Even mix all Permissions into one Role.
I would probably not touch user created roles.
Are Permissions cleaned up on plugin removal?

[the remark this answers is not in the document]
goosemania: Yes but not explicitly. [Content type is removed](https://github.com/pulp/pulpcore/pull/1434/files#diff-aedc0acb0d43552fa8aafe3a3ed6403a90320f5d2309b3160cc75c138d4ab1d3R85), so permissions are removed as [a cascade](https://github.com/django/django/blob/stable/3.2.x/django/contrib/auth/models.py#L61).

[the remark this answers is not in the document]
mdellweg: So you would end up with empty roles? Still maybe don't touch user created roles. They might attach a deeper meaning to them.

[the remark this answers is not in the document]
goosemania: Yes, there will be empty roles if all perms were from the removed plugin, or there will be a subset of perms left on that role.
Ok, I think you are right, I'll go through the LOCKED_ROLES only.

pulpcore/app/apps.py
```diff
@@ -194,37 +197,130 @@ def ready(self):
...
 
 
-def _populate_access_policies(sender, **kwargs):
+def _rename_permissions_assignment_workaround(access_policy):
+    from pulpcore.app.loggers import deprecation_logger
+
+    if "permissions_assignment" in access_policy:
+        deprecation_logger.warn(
+            "The field 'permissions_assignment' has been renamed to 'creation_hooks'. This "
+            "workaround may be removed with pulpcore 3.17."
```
bmbouter, line 206: Let's update this to 3.20 everywhere plz.

pulpcore/app/models/access_policy.py
```diff
@@ -68,18 +85,24 @@ def add_perms(self):
...
+                function = self.REGISTERED_CREATION_HOOKS.get(creation_hook["function"])
+                if function is not None:
+                    kwargs = creation_hook.get("parameters") or {}
+                    function(**kwargs)
+                else:
+                    # Old interface deprecated for removal in 3.17
+                    function = getattr(self, creation_hook["function"])
+                    deprecation_logger.warn(
+                        "Calling unregistered creation hooks from the access policy is deprecated"
+                        " and may be removed with pulpcore 3.17 "
```
bmbouter, line 102: s/3.17/3.20/

CHANGES/9411.feature
```diff
@@ -0,0 +1,7 @@
+Added a role model to support RBAC and the utility functions ``assign_role`` and ``remove_role``.
+
+The field ``permissions_assignment`` of access policies has been renamed to ``creation_hooks``. A
+compatibility patch has been added to be removed with pulpcore=3.18.
+
+The ``permissions`` argument to ``creation_hooks`` has been deprecated to be removed with
+pulpcore=3.18.
```
bmbouter, line 7:
```suggestion
pulpcore=3.20.
```

CHANGES/9411.feature
```diff
@@ -0,0 +1,7 @@
+Added a role model to support RBAC and the utility functions ``assign_role`` and ``remove_role``.
+
+The field ``permissions_assignment`` of access policies has been renamed to ``creation_hooks``. A
+compatibility patch has been added to be removed with pulpcore=3.18.
```
bmbouter, line 4:
```suggestion
compatibility patch has been added to be removed with pulpcore=3.20.
```

CHANGES/plugin_api/9411.feature
```diff
@@ -0,0 +1,7 @@
+Added ``get_objects_for_user`` and hooks in ``AutoAddObjPermsMixin`` to support queryset filtering
+by and auto-assignment of roles.
+
+Changed the lookup for creation hooks so hooks need to be registered in
+``REGISTERD_CREATION_HOOKS`` on the model to be used. The signature for creation hooks that are
+registered must match the exploded version of the dict parameters from the access policy.
+Unregistered creation hooks are deprecated and support will be dropped in pulpcore 3.18.
```
bmbouter, line 7:
```suggestion
Unregistered creation hooks are deprecated and support will be dropped in pulpcore 3.20.
```

?

pulpcore/app/models/role.py
```diff
@@ -0,0 +1,71 @@
...
+    Relations:
+
+        permissions (models.ManyToManyField): Permissions to be granted via this role.
+    """
+
+    name = models.CharField(max_length=128, db_index=True, unique=True)
+    description = models.TextField(null=True)
+    locked = models.BooleanField(default=False)
+    permissions = models.ManyToManyField(Permission)
```
gerrod3, line 27: Roles can have permissions related to many different objects, but only assigned to one specific object. Is this a use case we need to worry about? Is it something we can add later?

gerrod3 (commented):
Few questions and mainly grammar corrections. Looking good.

CHANGES/plugin_api/9411.feature
```diff
@@ -0,0 +1,7 @@
+Added ``get_objects_for_user`` and hooks in ``AutoAddObjPermsMixin`` to support queryset filtering
+by and auto-assignment of roles.
+
+Changed the lookup for creation hooks so hooks need to be registered in
+``REGISTERD_CREATION_HOOKS`` on the model to be used. The signature for creation hooks that are
+registered must match the exploded version of the dict parameters from the access policy.
```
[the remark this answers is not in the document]
mdellweg: Like unpacked. The parameters dict is unpacked (`**parameters`) and takes more space now. Like an explosion.

pulpcore/app/apps.py
```diff
@@ -194,37 +197,130 @@ def ready(self):
...
+            description = desired_role.get("description")
+            permissions = desired_role["permissions"]
+        elif isinstance(desired_role, list):
+            description = None
+            permissions = desired_role
+        else:
+            raise RuntimeError(
+                _("Locked role definition for {name} is incompatible.").format(name=name)
+            )
+        permissions = [_get_permission(perm) for perm in permissions]
```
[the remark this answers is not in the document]
mdellweg: I think, yes and yes. If the permission does not exist, we delivered a bug in LOCKED_ROLES. We do tests this in CI and can make sure it does not fail.

docs/plugins/plugin-writer/concepts/rbac/adding_automatic_permissions.rst
```diff
@@ -123,37 +125,53 @@ This effectively creates a "user isolation" policy which aligns with the example
...
 
-        def my_custom_callable(self, permissions, parameters):
-            # NOTE: permissions and parameters can be either a single entity or a list of entities
-            from guardian.shortcuts import assign_perm
-            user_or_group = parameters
-            for permission in permissions:
-                assign_perm(permissions, user_or_group, self)  # self is the object being assigned
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            self.REGISTERED_CREATION_HOOKS["my_custom_callable"] = self.my_custom_callable
```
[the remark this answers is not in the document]
mdellweg: Not sure how we can make it extendable on subclassing as a ClassVar, without writing a complicated meta class.

pulpcore/app/models/role.py
mdellweg: UserRoles can also be assigned without an object. So you can create the almighty role. Or you can have the `pulp_plugin_x_admin` role.
Also for the lazy, you could create a role that contains read permissions for all objects. I'd leave the freedom to the admin there.

pulpcore/app/models/access_policy.py
```diff
@@ -30,33 +37,43 @@ class AccessPolicy(BaseModel):
...
 
     """
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.REGISTERED_CREATION_HOOKS = {
+            "add_roles_for_users": self.add_roles_for_users,
+            "add_roles_for_groups": self.add_roles_for_groups,
+            "add_roles_for_object_creator": self.add_roles_for_object_creator,
+        }
```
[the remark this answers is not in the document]
mdellweg: To make the docs you mentioned actually work, i just swapped this line with the super call, so the dict will exist on the subclasses __init__.
I'll have a closer look at class level manipulation, but i doubt there's a feasible way.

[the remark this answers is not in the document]
mdellweg: It turns out, that subclasses will share the dict and i see no safe way to support this.

CHANGES/plugin_api/9411.feature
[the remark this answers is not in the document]
bmbouter: :boom:

bmbouter (approved):
This resolves all my issues and this looks good to me. Thank you!

goosemania (approved):
:tada:

PR #1637, opened 2021-09-21
Fix occasional deadlocks when doing multiple similar syncs concurrently.

ggainey:
Forcing deadlocks requires a lot of time and pulpcore-workers running.
There is therefore no specific CI test for this, but there is a reproducer
script that will force deadlocks to happen (and show that they're fixed) here:

https://github.com/ggainey/pulp_startup/blob/main/8750_deadlocks/file_repro.sh

fixes #8750.
[nocoverage]

Please be sure you have read our documentation on creating PRs:
https://docs.pulpproject.org/contributing/pull-request-walkthrough.html

PR #1644, opened 2021-09-24
Teach several touch() paths to apply order to avoid deadlocks.

ggainey:
fixes #9441.
[nocoverage]
Required PR: https://github.com/pulp/pulpcore/pull/1637

### DIFF
--- a/CHANGES/9441.bugfix
+++ b/CHANGES/9441.bugfix
@@ -1,0 +1,1 @@
+Taught several more codepaths to order-before-update to avoid deadlocks.
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -101,8 +101,11 @@ class BulkTouchQuerySet(models.QuerySet):
     def touch(self):
         """
         Update the ``timestamp_of_interest`` on all objects of the query.
+
+        We order-by-pk here to avoid deadlocking in high-concurrency
+        environments.
         """
-        return self.update(timestamp_of_interest=now())
+        return self.order_by("pk").update(timestamp_of_interest=now())
 
 
 class QueryMixin:


CHANGES/9441.bugfix
```diff
@@ -0,0 +1 @@
+Taught several touch() codepaths to order-before-touch to avoid deadlocks.
```
daviddavis, line 1: I don't think an end user will understand what "touch" means.

daviddavis:
Would it make sense to update the touch methods to order before calling update?

Also, it is looking to me like we should always be ordering before bulk updating/creating in our code. Is that correct?

ggainey:
> Would it make sense to update the touch methods to order before calling update?

Ordering often requires knowing what you're touching, which one may not "inside" the touch() call.

> Also, it is looking to me like we should always be ordering before bulk updating/creating in our code. Is that correct?

Yes, as a general rule. There are other situations other than bulk_update/create/get_or_create as well - doing a bunch of work inside a single transaction, for example, can deadlock if two threads are doing the same work on similar objects in arbitrary orders at the same time.

daviddavis (approved):

dralley:
@ggainey Alternative suggestion, maybe we could just change it here?

https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/content.py#L105

I can also see some good counterarguments - e.g. maybe the queryset already has an ordering.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -93,7 +93,11 @@ async def run(self):
             # swap it out with the existing one.
             for digest_type, digests in artifact_digests_by_type.items():
                 query_params = {"{attr}__in".format(attr=digest_type): digests}
-                existing_artifacts_qs = Artifact.objects.filter(**query_params)
+                # Give the artifacts we're about to touch a guaranteed order, to avoid deadlocks
+                # in heavy-concurrency situations.
+                existing_artifacts_qs = Artifact.objects.filter(**query_params).order_by(
+                    digest_type
+                )
```
dralley, line 100: I don't think this will actually order anything - the queryset here only applies to one digest_type at a time.

ggainey:
> @ggainey Alternative suggestion, maybe we could just change it here?
> 
> https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/content.py#L105

You have to know what you're touching, to be able to order it using natural-keys. **Anything** can be using the BulkTouchQuerySet, so we have no way of knowing how to order-by here.

> 
> I can also see some good counterarguments - e.g. maybe the queryset being touched already has an ordering applied.

mdellweg:
> > @ggainey Alternative suggestion, maybe we could just change it here?
> > https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/content.py#L105
> 
> You have to know what you're touching, to be able to order it using natural-keys. **Anything** can be using the BulkTouchQuerySet, so we have no way of knowing how to order-by here.
> 
> > I can also see some good counterarguments - e.g. maybe the queryset being touched already has an ordering applied.

I'd imagine, you can only touch items that are there (as with any other bulk update call). Should that not work with `order_by(pk)`?
With a bulk upsert (and i think we do not have it) i could see problems.

pulpcore/plugin/stages/content_stages.py
```diff
@@ -54,7 +54,11 @@ async def run(self):
 
             for model_type, content_q in content_q_by_type.items():
                 try:
-                    await sync_to_async(model_type.objects.filter(content_q).touch)()
+                    # Checking for "existing" content means we can rely on ordering by pulp_id to
+                    # avoid deadlocks in concurrent situations
+                    await sync_to_async(
+                        model_type.objects.filter(content_q).order_by("pulp_id").touch
```
dralley, line 60: If `pk` is equivalent to `pulp_id`, could we use the former?

dralley (approved):

ggainey:
> > > @ggainey Alternative suggestion, maybe we could just change it here?
> > > https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/content.py#L105
> > 
> > 
> > You have to know what you're touching, to be able to order it using natural-keys. **Anything** can be using the BulkTouchQuerySet, so we have no way of knowing how to order-by here.
> > > I can also see some good counterarguments - e.g. maybe the queryset being touched already has an ordering applied.
> 
> I'd imagine, you can only touch items that are there (as with any other bulk update call). Should that not work with `order_by(pk)`? With a bulk upsert (and i think we do not have it) i could see problems.

This is a valid observation - changing to order-in-touch(), makes things a lot cleaner.

dralley (approved):

daviddavis (approved):

mdellweg:
> > > > @ggainey Alternative suggestion, maybe we could just change it here?
> > > > https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/content.py#L105
> > > 
> > > 
> > > You have to know what you're touching, to be able to order it using natural-keys. **Anything** can be using the BulkTouchQuerySet, so we have no way of knowing how to order-by here.
> > > > I can also see some good counterarguments - e.g. maybe the queryset being touched already has an ordering applied.
> > 
> > 
> > I'd imagine, you can only touch items that are there (as with any other bulk update call). Should that not work with `order_by(pk)`? With a bulk upsert (and i think we do not have it) i could see problems.
> 
> This is a valid observation - changing to order-in-touch(), makes things a lot cleaner.

This obviously means that we need to order the same way with every single other bulk update call. Lucky us, we don't update artifacts nor content at all (touch not considered updating).

PR #1646, opened 2021-09-25
Turn deprecation warnings into errors for 3.16

dralley:
[noissue]


mdellweg (commented):
This looks right for me. But i think, we should have removal changelogs.

ipanova:
> This looks right for me. But i think, we should have removal changelogs.

+1 we need an issue

pulpcore/app/settings.py
```diff
@@ -253,7 +253,7 @@
 
 ALLOWED_EXPORT_PATHS = []
 
-ALLOW_SHARED_TASK_RESOURCES = False
+ALLOW_SHARED_TASK_RESOURCES = True
```
dralley, line 256: @mdellweg Changing this, as well.

mdellweg: +1

dralley:
@ipanova @mdellweg I added a few more items which I had missed.  It feels like a bit too much for one issue, I suppose I should file a few more?

mdellweg (approved):

mdellweg:
> @ipanova @mdellweg I added a few more items which I had missed. It feels like a bit too much for one issue, I suppose I should file a few more?

Those removals are not going to backported. So as long as you spell everything out in the changelog, this is enough for me.

CHANGES/plugin_api/9474.removal
```diff
@@ -0,0 +1 @@
+ContentSaver._pre_save() and ContentSaver._post_save() must now be implemented as synchronous functions rather than coroutines. Also, the `resources` argument of `dispatch()` has been removed. `exclusive_resources` and `shared_resources` should be used instead.
```
dralley, line 1: Combining these two into one message does feel strange.

ipanova: if it would be up to  me i'd file 3 issues
1. ALLOW_SHARED_TASK_RESOURCES defaults to True and goes into plugin_api as feature
2. removal of resources argument and goes into plugin_api removal
3. content pre and post save hooks are sync and goes into plugin api removal
Having 3 changelogs imo looks cleaner.

dralley:
@ipanova Feel free to merge this if it looks good

CHANGES/plugin_api/9474.feature
```diff
@@ -0,0 +1 @@
+`ALLOW_SHARED_TASK_RESOURCES` is now enabled by default.  If all goes smoothly, this will become permanent and the setting will be removed in the next release.
```
mdellweg, line 1: I'm not so sure this is really a "plugin_api" thing. The changes to the plugin api have been included in the last release.

mdellweg (approved):
One nitpick, but good to go anyway.

CHANGES/plugin_api/9474.feature
ipanova: well, depends how you look at it- now it is true by default, previously it was set to false. I think it is ok to have this entry where it is now.

PR #1651, opened 2021-09-28
Add validation for credentials in remote urls

daviddavis:
fixes #9459


CHANGES/9459.feature
```diff
@@ -0,0 +1,2 @@
+Added validation to prevent credentials in remote urls. Users should update any existing remote urls
+that contain credentials by moving them to the remote username and password fields.
```
daviddavis, line 2: I thought about a migration but decided against it because: 

1. The number of users that this affects is probably trivial
2. I'm always hesitant about quietly munging users' data

mdellweg (approved):

CHANGES/9459.feature
dralley: We should probably consult Tanya about this.  It's possible that some users could have been using this with Pulp 2 extensively, and a re-migration is not always an option.  If we're removing a feature I think we should try to not make it a breaking change.

ipanova: To be fair, we're removing a feature that never worked in pulp3, right? https://pulp.plan.io/issues/9464
Whether this is a freshly created remote or a migrated one.

daviddavis: It works in other plugins (e.g. pulp_file) and I had assumed it never worked in pulp_rpm. However I tried reverting [this commit](https://github.com/pulp/pulp_rpm/commit/db55c27a9e78a18b64849915bc33628c8a9456fd) and confirmed that it now works in pulp_rpm. So it looks like https://github.com/pulp/pulp_rpm/commit/db55c27a9e78a18b64849915bc33628c8a9456fd broke this functionality.

daviddavis: If we proceed with this change, we'll definitely need to add a migration. My theory about users not using credentials in their urls was based on the assumption that this never worked and therefore it wasn't used (which is obviously false).

ipanova: gotcha, thanks for the clarification and investigation :+1:

ipanova: I believe this PR is doing the correct thing and we should avoid having the credentials in the url, same as we moved them out of the proxy url. I do believe though we might want to write the migration.

daviddavis: Ok, added migration.

pulpcore/app/migrations/0077_move_remote_url_credentials.py
```diff
@@ -0,0 +1,34 @@
...
+        url = urlparse(remote.url)
+
+        if not remote.username:
+            remote.username = url.username
+        if not remote.password:
+            remote.password = url.password
+
+        url = url._replace(netloc=url.netloc[url.netloc.index("@") + 1:])
+        remote.url = urlunparse(url)
+        remote.save()
```
bmbouter, line 21: What do you think about a unit test for this? I'm +1 to merge without it also, but I wanted to ask.

daviddavis: I haven't written a unit test for a migration before but I can look into it.

daviddavis: Do we have an example of a unit test for a migration?

bmbouter: Not that I know of, but I imagined it would Mock `Remote.objects.filter(url__contains="@").iterator()` and assert on the save calls.

ipanova: This seems to be a bit more readable:
```
_, url = url.netloc.rsplit("@", maxsplit=1)
remote.url = urlunparse(url._replace(netloc=url))
```

daviddavis:
We probably shouldn't merge this until https://pulp.plan.io/issues/9458 is fixed.

bmbouter (approved):
This looks great. Thank you!

ipanova (approved):
Nice!

ipanova:
@daviddavis here is the PR for the migration plugin https://github.com/pulp/pulp-2to3-migration/pull/457/files once @goosemania  will review and approve it I think we can un-draft and merge your PR.

PR #1663, opened 2021-09-30
Add a wrapper around all redis calls

lubosmj:
closes #8997


lubosmj:
@bmbouter, @dralley, when Pulp loses connection to the Redis server, I am only logging an operation that failed. A user may not notice that the operation failed since everything is silently logged (which means that the content, e.g., could not be properly deleted). Do you think I should at least raise an HTTP error when a Redis transaction could not be successfully completed? Which HTTP status code would best describe the server failure if not 500?

dralley:
@lubosmj There's three main scenarios that need to be handled.

1) Connection failure during content app cache queries
2) Connection failure during content app cache entry creation
3) Connection failure during cache flushes

 I don't think either should trigger any kind of hard error that gets surfaced to the user.  Apart from any logging going on, (1) should be treated the same as a cache miss, (2) hould be ignored and handled as though no cache exists, and (3) should be ignored for now, since cache entries expire after 10 minutes, and the proper mitigation would be something entirely different - detecting whether Redis has gone missing and then flushing the entire cache when it comes back online, rather than throwing errors back at the user for individual actions.

What you've done with this PR in terms of logging is really clever!  But, I'm not sure that it's actually the right approach.  The problem is that if you just wrap the underlying Redis calls so that they swallow the errors and print log messages, you are effectively change the signatures of those functions so that they return `None` in situations where it used to raise an Exception, and making it so that none of the functions on the Redis connector will do behave the way the docs say they do.  I don't think we want to do that.

Beyond just being unexpected, I think it would break some [existing code](https://github.com/pulp/pulpcore/blob/master/pulpcore/app/views/status.py#L114-L118) and also be less efficient because, some places (like creating a cache entry) do multiple calls into Redis.  If the first one fails, we don't want to keep going (with e.g. creating a cache entry), we want to stop immediately.  But if we just silently swallow the error then [the failure of the first call which sets the key will not prevent the second call which sets the expiration on the key](https://github.com/pulp/pulpcore/blob/master/pulpcore/cache/cache.py#L92), and we'll be wasting effort and also printing more errors to the log than necessary.

(Note to self, there might be an existing latent bug here. If the first call to create the key succeeds and the [2nd call to set the expiration failed](https://github.com/pulp/pulpcore/blob/master/pulpcore/cache/cache.py#L95), then it might create unexpectedly immortal cache entries)

Since everything about the cache is already handled via Cache abstractions in [cache.py](https://github.com/pulp/pulpcore/blob/master/pulpcore/cache/cache.py), I think what we ought to do is just wrap those methods in the appropriate `try ... except` block (a common decorator would be fine for this).

@bmbouter @gerrod3 What do you think about this?

gerrod3:
I agree that the methods in `cache.py` should probably be wrapped in `try ... except` blocks. This function in particular needs to revert back to the original handler function if any of the redis calls fail: https://github.com/pulp/pulpcore/blob/master/pulpcore/cache/cache.py#L161

bmbouter:
I also agree. Thank you @dralley for outlining the solution so clearly.

.github/workflows/scripts/script.sh
```diff
@@ -173,6 +173,8 @@ if [[ "$TEST" == "performance" ]]; then
   exit
 fi
 
+redis-cli shutdown
```
dralley, line 176: We do want to test the cache, so perhaps it would be best to add a new test runner with no Redis?

lubosmj: Adding the new test runner would probably double the time spent in the GA checks. Are you sure we want to do that?

pulpcore/app/redis_connection.py
```diff
@@ -35,6 +42,56 @@ def get_async_redis_connection():
...
+        if "decorator_func" not in attrs:
+            raise ValueError("The decorator_func attribute needs to be declared")
+
+        child = super().__new__(mcs, name, bases, attrs)
+        for base in bases:
+            for field_name, field in base.__dict__.items():
+                if callable(field) and not field_name.startswith("__"):
+                    decorated_func = attrs["decorator_func"](field)
+                    setattr(child, field_name, decorated_func)
+        return child
```
dralley, line 61: @lubosmj Why do we need the metaclass if we're manually decorating the methods in question?

lubosmj: The metaclass was initially meant for logging purposes. I can get rid of it.

pulpcore/cache/cache.py
```diff
@@ -161,16 +187,21 @@ def __call__(self, func):
...
-                # Cache miss, create new entry
-                response = await self.make_entry(key, bk, func, args, self.default_expires_ttl)
+
+            try:
+                if callable(self.default_base_key):
+                    bk = await self.default_base_key(request, self)
+                if self.auth:
+                    await self.auth(request, self, bk)
+                key = self.make_key(request)
+                # Check cache
```
dralley, line 197: I believe the try block doesn't need to start until you get here

pulpcore/cache/cache.py
```diff
@@ -161,16 +187,21 @@ def __call__(self, func):
...
+                response = await self.make_response(key, bk)
+                if response is None:
+                    # Cache miss, create new entry
+                    response = await self.make_entry(key, bk, func, args, self.default_expires_ttl)
+            except AConnectionError:
```
dralley, line 202: Which parts of this are capable of raising the exception rather than returning None?

pulpcore/cache/cache.py
lubosmj: Yes, you are right.

pulpcore/cache/cache.py
lubosmj: The methods `make_response` and `make_entry` may raise exceptions related to connection errors since they are working with the Redis client instance. Both of these methods are called in every invocation of `ContentCache`. I am not quite sure what are the benefits of distinguishing between `returning None` and `raising an exception` in this case.

dralley: The async version of the cache ought to wrap its methods and return None as the non-async version does, to make the APIs as similar as possible.  I think the try-catch block would no longer be necessary.

`make_response` feels like it ought to be renamed to `try_get_cached_response` or something along those lines while `make_entry` feels like it should be split into two methods, `make_response` (the real one) and `attempt_cache_response`

pulpcore/cache/cache.py
gerrod3: Where is here? The cache object is passed to `default_base_key` and thus could use redis to generate the base_key and fail. I think the try block being at the top is fine.

pulpcore/cache/cache.py
gerrod3: Wrapping the async methods is a good idea and you would be able to get rid of the try-except block. Breaking up `make_entry` is also good idea since it does a lot right now and would make the code more clear.

pulpcore/cache/cache.py
```diff
@@ -18,6 +26,20 @@ class CacheKeys(enum.Enum):
...
+def connection_error_wrapper(func):
+    """A decorator that enables functions which utilize Redis to swallow connection errors."""
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        """A function that handles connection errors raised by the Redis client."""
+        try:
+            return func(*args, **kwargs)
+        except ConnectionError:
+            return None
```
gerrod3, line 38: Should a message be logged here when the redis connection erred? Maybe a general info message so the user knows what happened?

dralley: The trouble is that it would start spamming the logs very very heavily.

lubosmj:
@dralley, the PR is once again ready for review. I have decorated the async methods and removed the try-except block.

dralley (approved):
This looks good to me, thank you Lubos

gerrod3 (approved):

bmbouter (approved):
I wonder about the logging aspect, but this is so good let's merge it without further delay!

PR #1667, opened 2021-10-05
Improved messaging around timeout requests.

ipanova:
closes #9301

### DIFF
--- a/pulpcore/download/http.py
+++ b/pulpcore/download/http.py
@@ -1,9 +1,11 @@
 import logging
 
 import aiohttp
+import asyncio
 import backoff
 
 from .base import BaseDownloader, DownloadResult
+from pulpcore.exceptions import TimeoutException
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


daviddavis (approved):

pulpcore/download/http.py
```diff
@@ -242,7 +244,14 @@ async def run(self, extra_data=None):
...
             async def download_wrapper():
-                return await self._run(extra_data=extra_data)
+                try:
+                    return await self._run(extra_data=extra_data)
+                except asyncio.TimeoutError:
+                    raise RuntimeError(
+                        _(
+                            "Request timed out for {}. Increasing the total_timeout value on the remote might help."
+                        ).format(self.url)
+                    )
```
dralley, line 254: @ipanova Perhaps instead we should make `aiohttp.TimeoutError` a retriable event? If the server takes 300 seconds / 5 minutes to complete, then there is a good chance it is not *going* to complete regardless of the timeout because something has gone wrong.

ipanova: I see what you're saying but the end error even after retries will 'timeout from none' which is not user friendly

dralley: I definitely agree with that, it's really two separate issues.  No objection to adding a better error message (which I think is logged on backoff)?

ipanova: I added the new exception to the retry-able errors list.

daviddavis (approved):

PR #1674, opened 2021-10-06
Add documentation for DB_ENCRYPTION_KEY setting

daviddavis:
fixes #9495


docs/configuration/settings.rst
```diff
@@ -36,6 +36,15 @@ SECRET_KEY
...
 
+DB_ENCRYPTION_KEY
+^^^^^^^^^^^^^^^^^
+
+  The file location of a symmetric fernet key that Pulp uses to encrypt sensitive fields in the
+  database. Default location is ``/etc/pulp/certs/database_fields.symmetric.key``.
+
+  The key is automatically generated by default with pulp_installer and single container image. The
+  key can be generated independently but it must be a string of 32 url-safe base64-encoded bytes.
+
```
thmo, line 47: Would be nice to mention that is can be achieved e.g. by using
```
openssl rand -base64 32 > /etc/pulp/certs/database_fields.symmetric.key
```
(like the installer does).

mdellweg: Maybe it's just me. But is this 32 bytes that are represented in base64, so the string is actually longer than 32 characters?
What about "a base64-encoded string containing 32 random bytes"?

thmo: It's 32 random bytes encoded in base64, which yields a string of length 44.

PR #1680, opened 2021-10-11
Fix erroneous order fields for `StableOrderingFilter` backend

gerrod3:
fixes: #9496

### DIFF
--- a/pulpcore/app/viewsets/base.py
+++ b/pulpcore/app/viewsets/base.py
@@ -80,6 +80,19 @@ class StableOrderingFilter(OrderingFilter):
 
         return list(ordering) + ["-" + field.name]
 
+    def get_default_valid_fields(self, queryset, view, context={}):
+        """
+        Ensure that default valid fields don't include fields not on the queryset model.
+
+        When `ordering_fields` is not specified this method is called to find the fields to order
+        by. It uses the default serializer class's fields to find the `ordering_fields` which could
+        include fields from related objects not present in the queryset.
+        """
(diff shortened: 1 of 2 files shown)


pulpcore/app/viewsets/base.py
```diff
@@ -80,6 +80,19 @@ def get_ordering(self, request, queryset, view):
 
         return list(ordering) + ["-" + field.name]
 
+    def get_default_valid_fields(self, queryset, view, context={}):
+        """
+        Ensure that default valid fields don't include fields not on the queryset model.
+
+        When `ordering_fields` is not specified this method is called to find the fields to order
+        by. It uses the default serializer class's fields to find the `ordering_fields` which could
+        include fields from related objects not present in the queryset.
```
dralley, line 89: > It uses the default serializer class's fields to find the `ordering_fields` which could include fields from related objects not present in the queryset.

This feels like it could potentially be a bug in Django Rest Framework?

dralley: Most likely here:  https://github.com/encode/django-rest-framework/blob/71e6c30034a1dd35a39ca74f86c371713e762c79/rest_framework/filters.py#L238-L241

It looks like they perform a check similar to this in a [different place](https://github.com/encode/django-rest-framework/blob/71e6c30034a1dd35a39ca74f86c371713e762c79/rest_framework/filters.py#L254-L256), perhaps it ought to be included above.

(To be clear, we would still want to carry the fix here until it is fixed upstream)

dkliban (approved):

PR #1681, opened 2021-10-11
Fix error when listing repo version while running orphan cleanup

gerrod3:
fixes: #9481

Since orphan cleanup became a non-blocking task it was possible for this query to "return'" `None` if the first `Content` was deleted before the `cast` call was completed. Now it should use a `Content` object that is apart of `RepositoryVersion`.

### DIFF
--- a/CHANGES/9481.bugfix
+++ b/CHANGES/9481.bugfix
@@ -1,0 +1,1 @@
+Fixed issue with listing repository versions while running orphan cleanup task.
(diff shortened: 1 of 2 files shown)


rossengeorgiev:
Just tested this patch by manually making the changes on `katello 4.1`, and I still get the error. Details: https://community.theforeman.org/t/pulpdebclient-apierror-http-500-when-running-katello-delete-orphaned-content/25468/9?u=rgp

PR #1686, opened 2021-10-14
[WIP] Don't lock distribution updates unless "base_path" is being modified

dralley:


dralley:
TL;DR 

https://github.com/pulp/pulpcore/blob/master/pulpcore/app/viewsets/base.py#L458

combined with

https://github.com/pulp/pulpcore/blob/master/pulpcore/app/viewsets/publication.py#L268-L269

Causes the creation, update or removal of distributions to globally lock out the creation, update or removal of all other distributions, even if `base_path` isn't being touched, and we suspect this might be the cause of a bottleneck for one user.

pulpcore/app/viewsets/publication.py
```diff
@@ -264,6 +264,9 @@ class DistributionViewSet(
     serializer_class = DistributionSerializer
     filterset_class = DistributionFilter
 
-    def async_reserved_resources(self, instance):
+    def async_reserved_resources(self, instance, request):
         """Return resource that locks all Distributions."""
-        return ["/api/v3/distributions/"]
+        if "base_path" in request.data:
+            return ["/api/v3/distributions/"]
```
mdellweg, line 270:
```suggestion
            return [instance, "/api/v3/distributions/"]
```
If we treat `"/api/v3/distributions/"` as the `base_bath` lock, we need to grab the one for the instance anyway.

pulpcore/app/viewsets/publication.py
```diff
@@ -264,6 +264,9 @@ class DistributionViewSet(
     serializer_class = DistributionSerializer
     filterset_class = DistributionFilter
 
-    def async_reserved_resources(self, instance):
+    def async_reserved_resources(self, instance, request):
```
mdellweg, line 267: It would suffice to pass `data` here, right?

bmbouter: Adding this here means we can't use it from places where there isn't a request, e.g. in a task to dispatch more tasks. Is there some other way we can implement this?

mdellweg:
I still want to see this change. Can we configure stalebot to stay away from PR's?

dralley:
I'm going to be preoccupied with the RPM plugin for the forseeable future.  If someone else wants to pick it up that's fine but I don't know that I'll get to it soon.

PR #1688, opened 2021-10-14
Remove the ALLOWED_SHARED_TASK_RESOURCES option

dralley:
closes: #9506
https://pulp.plan.io/issues/9506

### DIFF
--- a/CHANGES/9506.misc
+++ b/CHANGES/9506.misc
@@ -1,0 +1,1 @@
+Removed the `ALLOWED_SHARED_TASK_RESOURCES` setting (it is now the standard behavior).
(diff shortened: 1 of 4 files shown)


CHANGES/9506.misc
```diff
@@ -0,0 +1 @@
+Removed the `ALLOWED_SHARED_TASK_RESOURCES` setting (it is now the globally default behavior).
```
mdellweg, line 1:
```suggestion
Removed the `ALLOWED_SHARED_TASK_RESOURCES` setting (it now is globally the behavior).
```

"default" sounds to me like it could be disabled.

dralley: Perhaps "standard" would be better

dkliban (approved):

PR #1698, opened 2021-10-25
Handles closed db connections

dkliban:
When the authentication middleware was added in pulpcore 3.15, it became the first place
in the content app that made an attempt to use the database. As a result, it is a convinient
place to handle InterfaceError and  OperationalError which are raised when the database
connection has been closed. When this occurs, Handler._reset_db_connection() is called to
clean up the database connection and the middleware tries to use the database again.

If the database connection is closed later in the handling of the request by the content app,
the user will still get a 500 error. However, the next request will be handled properly.

This patch also adds a call to Handler._reset_db_connection() inside the heartbeat method.

fixes: #9276
https://pulp.plan.io/issues/9276

### DIFF
--- a/pulpcore/tests/unit/content/test_heartbeat.py
+++ b/pulpcore/tests/unit/content/test_heartbeat.py
@@ -1,0 +1,35 @@
+import asyncio
+from unittest import skip
+from unittest.mock import patch, call
+
+from django.db.utils import InterfaceError, OperationalError
+from django.test import TestCase
+
+from pulpcore.content import _heartbeat
+
+
+class ContentHeartbeatTestCase(TestCase):
+    @skip("Skipping while resolving https://github.com/rochacbruno/dynaconf/issues/689")
+    @patch("pulpcore.app.models.ContentAppStatus.objects.get_or_create")
+    @patch("pulpcore.content.handler.Handler._reset_db_connection")
+    def test_db_connection_interface_error(self, mock_reset_db, mock_get_or_create):
+        """
+        Test that if an InterfaceError or OperationalError is raised,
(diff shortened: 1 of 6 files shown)


mdellweg (approved):

pulpcore/tests/unit/content/test_heartbeat.py
```diff
@@ -0,0 +1,33 @@
...
+        Handler._reset_db_connection() is called
+        """
+
+        class MockException(Exception):
+            pass
+
+        mock_get_or_create.side_effect = [InterfaceError(), OperationalError(), MockException()]
+
+        loop = asyncio.get_event_loop()
+        with self.settings(CONTENT_APP_TTL=1):
```
bmbouter, line 25: Did this get resolved with the new dynaconf or is this test still taking a >10 seconds to run?

dkliban: no, i am about to file a new bug against dynaconf.

bmbouter: Then I don't think we should merge this. Having one unit test is not worth additional tens of seconds runtime on the test suite (to me). We should consider alternatives.

dkliban: I added a skip decorator with a link to the dynaconf bug.

bmbouter (approved):
Thanks!

dkliban:
this is failing due to incompatibilty with newly released sphinx 4.3.0 - https://pypi.org/project/Sphinx/4.3.0/

PR #1701, opened 2021-10-27
Improves tolerance of db connection failures in pulpcore-content

dkliban:
This patch moves Handler._reset_db_connection() call to get_base_paths_blocking(). This ensures that the
connection is reset inside the same thread that is used to perform ORM calls. This fixes a bug where the
/pulp/content/ page would return a 500 error after the connection the database is closed due to a network
problem or a database restart.

A call to Handler._reset_db_connection() has also been added to the _heartbeat() co-routine. This ensures
that the heartbeat continues to be recorded in the database even if the db connection was closed.

fixes: #9515
https://pulp.plan.io/issues/9515


pulpcore/tests/unit/content/test_heartbeat.py
```diff
@@ -0,0 +1,33 @@
...
+    def test_db_connection_interface_error(self, mock_reset_db, mock_get_or_create):
+        """
+        Test that if an InterfaceError or OperationalError is raised,
+        Handler._reset_db_connection() is called
+        """
+
+        class MockException(Exception):
+            pass
+
+        mock_get_or_create.side_effect = [InterfaceError(), OperationalError(), MockException()]
```
bmbouter, line 22: The triplet!

bmbouter (approved):
:+1: Thanks!

goosemania (approved):
:tada:

mdellweg (approved):

PR #1703, opened 2021-10-29
Keep listening on pg channels in workers

mdellweg:
There seems to be a time window, where workers might miss a wakeup call
and end in waiting state, while tasks are still available.

[noissue]

@dalley i decided to turn this into a pr, because it might also fix a rare ci failure where canceling a task is lost in that window.


dralley:
@mdellweg This patch makes sense to me.  I do think we should file an issue for it in case we end up backporting it.

dralley (approved):

bmbouter (approved):
This looks good to me. Thanks!

PR #1708, opened 2021-11-02
Assembles chunked_uploads in workers working directory

dkliban:
fixes: #9550
https://pulp.plan.io/issues/9550

### DIFF
--- a/pulpcore/app/tasks/upload.py
+++ b/pulpcore/app/tasks/upload.py
@@ -26,7 +26,7 @@ def commit(upload_id, sha256):
         return
 
     chunks = models.UploadChunk.objects.filter(upload=upload).order_by("offset")
-    with NamedTemporaryFile("ab") as temp_file:
+    with NamedTemporaryFile(mode="ab", dir=".", delete=False) as temp_file:
         for chunk in chunks:
             temp_file.write(chunk.file.read())
         temp_file.flush()
--- a/CHANGES/9550.bugfix
+++ b/CHANGES/9550.bugfix
@@ -1,0 +1,1 @@
+Fixed bug where chunked uploads were being assembled in /tmp.


pulpcore/app/tasks/upload.py
```diff
@@ -26,7 +26,7 @@ def commit(upload_id, sha256):
         return
 
     chunks = models.UploadChunk.objects.filter(upload=upload).order_by("offset")
-    with NamedTemporaryFile("ab") as temp_file:
+    with NamedTemporaryFile(mode="ab", dir=".", delete=False) as temp_file:
```
bmbouter, line 29: Does this codepath happen inside a path? I'm thinking yes otherwise what would clean it up?

dkliban: inside what path? clean up when?

dkliban: This is a task being executed by a worker. The "." is the worker's directory. NamedTemporaryFile creates a temporary file inside that directory. When the delete is set to True, the file is deleted after it's closed on line 34. In order to keept it around until it is moved into artifact storage we need to set delete to False.

bmbouter: Yeah I meant `s/path/task/`. You answered it, thanks.

bmbouter (approved):
Thanks!

pulpcore/app/tasks/upload.py
mdellweg: Why is the file mode "ab" and not "wb"? If we used the default ("w+b") it should even be possible to read without needing to reopen.
Not saying it needs changing. It just sticks out as possibly unnecessary.

mdellweg (approved):

bmbouter:
I agree w/ you @mdellweg

dralley (approved):

PR #1716, opened 2021-11-09
Cleanup missing worker entries after seven days

bmbouter:
Missing worker entries are now kept in the database for seven days
before being cleaned up. This gives time for post-mortem analysis.

It also switches the cleanup of those records to be a bulk operation.

closes #8988

### DIFF
--- a/pulpcore/tasking/pulpcore_worker.py
+++ b/pulpcore/tasking/pulpcore_worker.py
@@ -110,11 +110,11 @@ class NewPulpWorker:
         _logger.info(_("Worker %s was shut down."), self.name)
 
     def worker_cleanup(self):
-        qs = Worker.objects.missing_workers()
+        qs = Worker.objects.missing_workers(age=timedelta(days=7))
         if qs:
             for worker in qs:
                 _logger.info(_("Clean missing worker %s."), worker.name)
-                worker.delete()
+            qs.delete()
 
     def beat(self):
         if self.worker.last_heartbeat < timezone.now() - timedelta(seconds=self.heartbeat_period):
--- a/CHANGES/8988.bugfix
+++ b/CHANGES/8988.bugfix
@@ -1,0 +1,2 @@
+Missing worker records are now kept in the database for seven days allowing time for post-mortem
+analysis of them. The user-facing data in the status API remains unmodified.
--- a/pulpcore/app/models/task.py
+++ b/pulpcore/app/models/task.py
@@ -42,20 +42,23 @@ class WorkerManager(models.Manager):
 
         return self.filter(last_heartbeat__gte=age_threshold)
 
-    def missing_workers(self):
+    def missing_workers(self, age=timedelta(seconds=settings.WORKER_TTL)):
         """
         Returns a queryset of workers meeting the criteria to be considered 'missing'
 
-        To be considered missing, a worker must have a stale timestamp.  Stale is defined here as
-        "beyond the pulp process timeout interval".
+        To be considered missing, a worker must have a stale timestamp.  By default, stale is
+        defined here as longer than the ``settings.WORKER_TTL``, or you can specify age as a
+        timedelta.
+
+        Args:
+            age (datetime.timedelta): Workers who have heartbeats older than this time interval are
+                considered missing.
 
         Returns:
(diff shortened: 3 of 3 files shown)


pulpcore/tasking/pulpcore_worker.py
```diff
@@ -109,6 +111,16 @@ def shutdown(self):
...
 
+    def _remove_old_unshown_worker_entries(self):
+        oldest_heartbeat = timezone.now() - timedelta(days=DAYS_TO_KEEP_OLD_WORKER_RECORDS)
+        _logger.debug(
+            _("Cleaning up Worker records older than %s days."), DAYS_TO_KEEP_OLD_WORKER_RECORDS
+        )
+        deleted, objs = Worker.objects.filter(last_heartbeat__lt=oldest_heartbeat).delete()
+        if deleted > 0:
+            i8ln_msg = _("Deleted {count} missing Worker records older than {days} days.")
+            _logger.info(i8ln_msg.format(count=deleted, days=DAYS_TO_KEEP_OLD_WORKER_RECORDS))
```
dralley, line 122: I'm not sure how I feel about adding this to the workload for every heartbeat, per worker.  It feels like overkill?

Maybe we could ship a `pulp_maintainence.timer` file for systemd.

bmbouter: It's not called for every heartbeat, just once on startup.

dralley: Ah, the way the diff is formatted lied to me :)

dralley (approved):

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -109,6 +111,16 @@ def shutdown(self):
         self.worker.delete()
         _logger.info(_("Worker %s was shut down."), self.name)
 
+    def _remove_old_unshown_worker_entries(self):
+        oldest_heartbeat = timezone.now() - timedelta(days=DAYS_TO_KEEP_OLD_WORKER_RECORDS)
+        _logger.debug(
+            _("Cleaning up Worker records older than %s days."), DAYS_TO_KEEP_OLD_WORKER_RECORDS
+        )
+        deleted, objs = Worker.objects.filter(last_heartbeat__lt=oldest_heartbeat).delete()
```
mdellweg, line 119: Is worker_cleanup not handling this already?

bmbouter: It doesn't if the worker is shutdown with `kill -9` or OOM. In that case the Worker record just stays there and would only be cleaned up if the PID on the box was assigned to a future worker.

mdellweg: That i don't understand. All i can see, is that the new code deletes workers with a different age in bulk on worker restart, and the existing code deletes all "missing" workers (timeout for that is defined in the tasks model) one by one every 100th heartbeat. I thing either of these is the desired behaviours. We should not duplicate the functionality here.

mdellweg: Also it may be nice to move the definition of `long_gone_workers` to the task model queryset like `missing_workers`.
Maybe even specify the lifetime as a parameter for that.

think:
```python
def missing_workers(self, worker_ttl=None):
    if worker_ttl is None:
        worker_ttl = settings.WORKER_TTL
    ...
```

bmbouter: Where is this 100th heartbeat cleanup? Can you look at the original issue filed and help me understand if that's already solved somehow?

mdellweg: It's the WORKER_CLEANUP_INTERVAL.

bmbouter: So it sounds like this bug can be closed NOTABUG?

dralley: Sounds like it

mdellweg: We still plan do a change here to get the 7 days afterlife for post mortem analysis.

mdellweg (approved):

bmbouter:
@mdellweg ty for your guidance on this.

PR #1717, opened 2021-11-09
Handled duplicates during bulk_create calls

bmbouter:
Multiple processes inserting the same objects into the DB can cause
unexpected duplicates. Now that we're using Django 3+ we can use the
`ignore_conflicts=True` param for bulk_create.

closes #8967

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -769,7 +769,7 @@ class RepositoryVersion(BaseModel):
                 )
             )
 
-        RepositoryContent.objects.bulk_create(repo_content)
+        RepositoryContent.objects.bulk_create(repo_content, ignore_conflicts=True)
 
     def remove_content(self, content):
         """
@@ -948,7 +948,7 @@ class RepositoryVersion(BaseModel):
                         count_type=value,
                     )
                     counts_list.append(count_obj)
-            RepositoryVersionContentDetails.objects.bulk_create(counts_list)
+            RepositoryVersionContentDetails.objects.bulk_create(counts_list, ignore_conflicts=True)
 
     def __enter__(self):
         """
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -396,7 +396,9 @@ class RemoteArtifactSaver(Stage):
         # Artifact sha256 for our ordering.
         if ras_to_create:
             ras_to_create_ordered = sorted(list(ras_to_create.values()), key=lambda x: x.sha256)
-            await sync_to_async(RemoteArtifact.objects.bulk_create)(ras_to_create_ordered)
+            await sync_to_async(RemoteArtifact.objects.bulk_create)(
+                ras_to_create_ordered, ignore_conflicts=True
+            )
         if ras_to_update:
             ras_to_update_ordered = sorted(list(ras_to_update.values()), key=lambda x: x.sha256)
             await sync_to_async(RemoteArtifact.objects.bulk_update)(
--- a/pulpcore/tests/unit/models/test_repository.py
+++ b/pulpcore/tests/unit/models/test_repository.py
@@ -15,7 +15,7 @@ class RepositoryVersionTestCase(TestCase):
         for _ in range(0, 5):
             contents.append(Content(pulp_type="core.content"))
 
-        Content.objects.bulk_create(contents)
+        Content.objects.bulk_create(contents, ignore_conflicts=True)
         self.pks = [c.pk for c in contents]
 
     def test_add_and_remove_content(self):
@@ -270,7 +270,7 @@ class RepositoryTestCase(TestCase):
         for _ in range(0, 3):
             contents.append(Content(pulp_type="core.content"))
 
-        Content.objects.bulk_create(contents)
+        Content.objects.bulk_create(contents, ignore_conflicts=True)
 
         versions = [self.repository.latest_version()]
         for content in contents:
--- a/CHANGES/8967.bugfix
+++ b/CHANGES/8967.bugfix
@@ -1,0 +1,2 @@
+Fixes unhandled errors during bulk_save operations due to race conditions of multiple processes
+creating the same database objects in separate processes.


bmbouter:
This was inspired by [this other PR](https://github.com/pulp/pulp_rpm/pull/2071/files).

Note, [this does not adjust this call to `bulk_create(...)`](https://github.com/pulp/pulpcore/blob/7835faa4310771ca88834efc01bc5def1df5c84b/pulpcore/app/models/content.py#L83-L94) because I wasn't sure if the ignore_conflicts=True would be returning the data as expected for those duplicates. Basically my changes here just leave it alone.

pulpcore/app/models/repository.py
```diff
@@ -769,7 +769,7 @@ def add_content(self, content):
                 )
             )
 
-        RepositoryContent.objects.bulk_create(repo_content)
+        RepositoryContent.objects.bulk_create(repo_content, ignore_conflicts=True)
```
dralley, line 772: I find it really confusing how we could be getting duplicates on this table when the `repository` is supposed to be locked?

dralley, line 772: I find it really confusing that we could be seeing duplicates on these two tables (this one and the one below) when only one task ought to have an exclusive lock on the `repository` table?

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -396,7 +396,9 @@ async def _handle_remote_artifacts(self, batch):
         # Artifact sha256 for our ordering.
         if ras_to_create:
             ras_to_create_ordered = sorted(list(ras_to_create.values()), key=lambda x: x.sha256)
-            await sync_to_async(RemoteArtifact.objects.bulk_create)(ras_to_create_ordered)
+            await sync_to_async(RemoteArtifact.objects.bulk_create)(
+                ras_to_create_ordered, ignore_conflicts=True
```
mdellweg, line 400: If we didn't add it here, would it be broken wrt concurrency?
I am wondering why we would not have had a workaround in place before this feature was added to django.

mdellweg: Maybe this _is_ the very place the issue #8967 happens. XD

bmbouter:
OK so I've done more work with this and I've learned a few things:

1) The reproducer script [here](https://pulp.plan.io/issues/8967#note-16) does cause [the traceback in the same comment](https://pulp.plan.io/issues/8967#note-16) 100% of the time.
2) With this patch applied, the script still causes the same exception 100% of the time.

In reading the traceback more closely, it's not that there is a duplicate key error. The error we actually get is more like `Key (version_added_id)=(a5c43989-e695-4f07-9bdb-0f879b9cdd31) is not present in table "core_repositoryversion".` which I also get locally. When I enter the debugger upon that bulk_create raising an exception, I can look at the `core_repositoryversion` table and indeed there is no repositoryversion with a pk starting with a5c4... Basically that repo version doesn't exist except in memory?

When I look at the repo in memory backing that missing repositoryversion it's name is:   `'../../../AppStream/x86_64/os/-31410651d4c07a347a6dbba79cfb4df7c6418e39b0a19c70a84eff2c8b3b8953'`.

The root cause of this is not clear to me, but I think it has to do with the creation of hidden repos maybe? For example when I use the same reproducer to sync a different repo (which doesn't produce a hidden repo) it works just fine with no pulpcore changes. That sync URL is `http://mirror.centos.org/centos-7/7/os/x86_64/`.

pulpcore/plugin/stages/artifact_stages.py
bmbouter: I think this is a valuable change that we just haven't hit due to the race condition being rare. I am +1 to making this change.

pulpcore/app/models/repository.py
bmbouter: See the big comment below, but a) this doesn't help resolve the problem and b) due to the locking this is unecessary. We shouldn't merge this line change.

pulpcore/tests/unit/models/test_repository.py
```diff
@@ -15,7 +15,7 @@ def setUp(self):
         for _ in range(0, 5):
             contents.append(Content(pulp_type="core.content"))
 
-        Content.objects.bulk_create(contents)
+        Content.objects.bulk_create(contents, ignore_conflicts=True)
```
bmbouter, line 18: This is in test code I think we don't need this.

mdellweg:
Is it possible, that the hidden repo version, you are referring to is not yet in the database, because it may have ended up in a transaction it wasn't meant to be? I once had a theory, that Django ORM in combination with async python may end up doing such things, because code is interleaved in an unpredictable manner.

bmbouter:
I'm closing and moving the issue back to NEW for pulp_rpm team to comment.

pulpcore/plugin/stages/artifact_stages.py
mdellweg: Do you still intend to make this single change? I think, we agreed upon it.

bmbouter: I was on the fence. We agreed it seems valuable, but also is it right in 100% of cases with no side effects? If you're sure I don't mind opening a PR. Also it needs an issue. Wdyt?

mdellweg: It's a tough call between "if it ain't broken..." and "will we remember, once we get a ticket filed".
Does django still populate the in memory representations for the ignored entries? Is it like `get_or_create` in that respect?

bmbouter: I don't know and I've been wondering this for a while actually. Maybe I should do a few minutes of testing...

bmbouter: Here's a simple test example. It confirms that we do not re-hydrate data from the database in cases where there would be database conflicts. It also confirms we do not overwrite the database data with the in-memory data attempting to be saved. This is partly what worries me about broad use of `ignore_conflicts`.

```
import os

os.environ["DJANGO_SETTINGS_MODULE"] = "pulpcore.app.settings"
import django
django.setup()


from pulpcore.app.models import Task

Task.objects.all().delete()

new_task = Task(state='completed')
new_task.save()

bulk_tasks = [
    Task(pk=new_task.pk, state='canceled'),
    Task()
]

objs = Task.objects.bulk_create(bulk_tasks, ignore_conflicts=True)

print(objs)  # This shows `[Task:  [canceled], Task:  []]`

print(Task.objects.get(pk=new_task.pk))  # Prints `Task:  [completed]` so the new obj wasn't saved
```

mdellweg: So we leave it until we know its a bug.

bmbouter: Agreed. :+1:

PR #1721, opened 2021-11-13
Added bulk-task-purge feature.

ggainey:
closes #8554.

### DIFF
--- a/CHANGES/8554.feature
+++ b/CHANGES/8554.feature
@@ -1,0 +1,8 @@
+Added a ``/tasks/purge/`` API to do bulk-deletion of old tasks.
+
+Over time, the database can fill with task-records. This API allows
+an installation to bulk-remove records based on their completion
+timestamps.
+
+NOTE: this endpoint is in tech-preview and may change in backwards
+incompatible ways in the future.
--- a/pulpcore/app/serializers/purge.py
+++ b/pulpcore/app/serializers/purge.py
@@ -1,0 +1,21 @@
+from datetime import datetime, timedelta, timezone
+from gettext import gettext as _
+
+from rest_framework import serializers
+
+from pulpcore.app.serializers import ValidateFieldsMixin  # noqa
+from pulpcore.constants import TASK_FINAL_STATES
+
+
+class PurgeSerializer(serializers.Serializer, ValidateFieldsMixin):
+    finished_before = serializers.DateTimeField(
+        help_text=_(
+            "Purge tasks completed earlier than this timestamp. Format '%Y-%m-%d[T%H:%M:%S]'"
+        ),
+        default=(datetime.now(timezone.utc) - timedelta(days=30)).strftime("%Y-%m-%d"),
+    )
+    states = serializers.MultipleChoiceField(
+        choices=TASK_FINAL_STATES,
+        default=["completed"],
+        help_text=_("List of task-states to be purged. Only 'final' states are allowed."),
+    )
--- a/pulpcore/app/tasks/purge.py
+++ b/pulpcore/app/tasks/purge.py
@@ -1,0 +1,50 @@
+from gettext import gettext as _
+from django_currentuser.middleware import get_current_authenticated_user
+from pulpcore.app.models import (
+    ProgressReport,
+    Task,
+)
+from pulpcore.app.role_util import get_objects_for_user
+
+
+def purge(finished_before, states):
+    """
+    This task purges from the database records of tasks which finished prior to the specified time.
+
+    It will remove only tasks that are 'owned' by the current-user (admin-users own All The Things,
+    so admins can delete all tasks).
+
+    It will not remove tasks that are incomplete (ie, in states running|waiting|cancelling).
+
+    It reports (using ProgressReport) the total entities deleted, as well as individual counts
+    for each class of entity. This shows the results of cascading-deletes that are triggered
+    by deleting a Task.
+
+    Args:
+        finished_before (DateTime): Earliest finished-time to **NOT** purge.
+        states (List[str]): List of task-states we want to purge.
+
+    """
+    current_user = get_current_authenticated_user()
+    qs = Task.objects.filter(finished_at__lt=finished_before, state__in=states)
+    units_deleted, details = get_objects_for_user(current_user, "core.delete_task", qs=qs).delete()
+
+    # Progress bar reporting total-units
+    progress_bar = ProgressReport(
+        message=_("Purged task-objects total"),
+        total=units_deleted,
+        code="purge.tasks.total",
+        done=units_deleted,
+        state="completed",
+    )
+    progress_bar.save()
+    # This loop reports back the specific entities deleted and the number removed
+    for key in details:
+        progress_bar = ProgressReport(
+            message=_("Purged task-objects of type {}".format(key)),
+            total=details[key],
+            code="purge.tasks.key.{}".format(key),
+            done=details[key],
+            state="completed",
+        )
+        progress_bar.save()
--- a/pulpcore/app/viewsets/task.py
+++ b/pulpcore/app/viewsets/task.py
@@ -3,18 +3,23 @@ from gettext import gettext as _
 from django_filters.rest_framework import DjangoFilterBackend, filters
 from drf_spectacular.utils import extend_schema
 from rest_framework import mixins, status
+from rest_framework.decorators import action
 from rest_framework.filters import OrderingFilter
 from rest_framework.response import Response
 from rest_framework.serializers import ValidationError
 
 from pulpcore.app.models import Task, TaskGroup, Worker
+from pulpcore.app.response import OperationPostponedResponse
 from pulpcore.app.serializers import (
+    AsyncOperationResponseSerializer,
     MinimalTaskSerializer,
     TaskCancelSerializer,
+    TaskGroupSerializer,
+    PurgeSerializer,
     TaskSerializer,
     WorkerSerializer,
-    TaskGroupSerializer,
 )
+from pulpcore.app.tasks import purge
 from pulpcore.app.viewsets import BaseFilterSet, NamedModelViewSet
 from pulpcore.app.viewsets.base import DATETIME_FILTER_OPTIONS, NAME_FILTER_OPTIONS
 from pulpcore.app.viewsets.custom_filters import (
@@ -24,6 +29,7 @@ from pulpcore.app.viewsets.custom_filters import (
     CreatedResourcesFilter,
 )
 from pulpcore.constants import TASK_INCOMPLETE_STATES, TASK_STATES, TASK_CHOICES
+from pulpcore.tasking.tasks import dispatch
 from pulpcore.tasking.util import cancel as cancel_task
 
 
@@ -90,6 +96,13 @@ class TaskViewSet(
                 "effect": "allow",
                 "condition": "has_model_or_obj_perms:core.change_task",
             },
+            # 'purge' is filtered by current-user and core.delete_task permissions at the queryset
+            # level, and needs no extra protections here
+            {
+                "action": ["purge"],
+                "principal": "authenticated",
+                "effect": "allow",
+            },
         ],
         "creation_hooks": [
             {
@@ -138,6 +151,28 @@ class TaskViewSet(
         if self.action == "partial_update":
             return TaskCancelSerializer
         return super().get_serializer_class()
+
+    @extend_schema(
+        description=(
+            "Trigger an asynchronous task that deletes completed tasks that finished prior"
+            " to a specified timestamp (tech-preview, may change in the future)."
+        ),
+        summary="Purge Completed Tasks",
+        operation_id="tasks_purge",
+        request=PurgeSerializer,
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    @action(detail=False, methods=["post"])
+    def purge(self, request):
+        """
+        Purge task-records for tasks in 'final' states.
+        """
+        serializer = PurgeSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        task = dispatch(
+            purge, args=[serializer.data["finished_before"], list(serializer.data["states"])]
+        )
+        return OperationPostponedResponse(task, request)
 
 
 class TaskGroupFilter(BaseFilterSet):
--- a/pulpcore/tests/functional/api/using_plugin/test_purge.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_purge.py
@@ -1,0 +1,347 @@
+"""
+Tests task-purge functionality.
+"""
+from datetime import datetime, timedelta, timezone
+
+from pulpcore.client.pulpcore import (
+    ApiClient,
+    ApiException,
+    Purge,
+    TasksApi,
+)
+from pulpcore.client.pulp_file import (
+    RemotesFileApi,
+    RepositoriesFileApi,
+    RepositorySyncURL,
+)
+
+from pulpcore.constants import TASK_STATES, TASK_FINAL_STATES
+
+from pulp_smash import config
+from pulp_smash.pulp3.bindings import (
+    monitor_task,
+    PulpTestCase,
+    PulpTaskError,
+)
+from .utils import gen_file_remote, gen_repo, gen_file_client, gen_user, del_user
+
+TOMORROW_STR = (datetime.now(timezone.utc) + timedelta(days=1)).strftime("%Y-%m-%dT%H:%M")
+
+
+class TaskPurgeTestCase(PulpTestCase):
+    """
+    Test task-purge functionality.
+    """
+
+    def _task_summary(self):
+        """
+        Summary of number of tasks in all known task-states.
+        :return: tuple of (total-tasks, Dict(state: count))
+        """
+        summary = {}
+        total = 0
+        final_total = 0
+        for state in TASK_STATES.__dict__.values():
+            response = self.task_api.list(state=state)
+            summary[state] = response.count
+            total += summary[state]
+            final_total += summary[state] if state in TASK_FINAL_STATES else 0
+        return total, final_total, summary
+
+    def _purge_report_total(self, task):
+        for report in task.progress_reports:
+            if report.code == "purge.tasks.total":
+                return report.total
+        self.fail("NO PURGE_TASKS_TOTAL?!?")
+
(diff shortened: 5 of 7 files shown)


pulpcore/app/views/purge.py
```diff
@@ -0,0 +1,31 @@
...
+        description=(
+            "Trigger an asynchronous task that deletes completed tasks that finished prior"
+            " to a specified timestamp."
+        ),
+        summary="Purge Completed Tasks",
+        operation_id="tasks_purge",
+        request=PurgeSerializer,
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    def post(self, request):
```
mdellweg, line 21: I think, you can add this as an extra action on the TaskViewSet. Then you would not need to mess with the urls.py at all.

ggainey: Good idea, will investigate

bmbouter:
I was looking at this thinking we should not have this run in a task, but maybe we should? I wrote this script:

```
from datetime import datetime

import os


os.environ["DJANGO_SETTINGS_MODULE"] = "pulpcore.app.settings"
import django
django.setup()


from pulpcore.plugin.models import Task

Task.objects.all().delete()
print("deleting all existing Tasks")

TASKS_PER_BATCH = 100000

for i in range(9, 10):
    print(f"inserting {TASKS_PER_BATCH} tasks")
    new_tasks = [Task() for i in range(TASKS_PER_BATCH)]
    Task.objects.bulk_create(new_tasks)

print(f"db now has {Task.objects.count()} tasks in it")

start = datetime.now()
Task.objects.all().delete()
print(f"deletion took: {datetime.now() - start}")
```

Which says that 100K deletions take 14-16 seconds on my system. So maybe? Anytime we get close to 30 seconds we should be using a task and "halfway" there is pretty close. Granted it is 100K tasks though...

Having an API do it in batches and just handle as far as it can I think would be an ok option too.

pulpcore/app/tasks/purge.py
```diff
@@ -0,0 +1,44 @@
...
+    It reports (using ProgressReport) the total entities deleted, as well as individual counts
+    for each class of entity. This shows the results of cascading-deletes that are triggered
+    by deleting a Task.
+
+    args:
+        finished_before (DateTime): Earliest finished-time to **NOT** purge.
+        states (List[str]): List of task-states we want to purge.
+
+    """
+    units_deleted, details = Task.objects.filter(
```
bmbouter, line 23: This is one of those things I was talking about in the multetenancy talk. Here this will delete all tasks in the system not just the ones you have permissions on. RBAC is still not everywhere and production ready but it is in place with Core, so I think we need to have guardian get all the Tasks that a user has the delete_task permission for.

To do that efficiently, I think we should move the queryset isolation from DRF's get_queryset() to the Task manager itself... If you did that you'd automatically get it correctly here.

Note though that we also need to check that the task runtime is "user aware" which I think it is but yeah, we gotta check.

pulpcore/app/serializers/purge.py
```diff
@@ -0,0 +1,18 @@
+from gettext import gettext as _
+
+from rest_framework import serializers
+
+from pulpcore.app.serializers import ValidateFieldsMixin  # noqa
+from pulpcore.constants import TASK_FINAL_STATES
+
+
+class PurgeSerializer(serializers.Serializer, ValidateFieldsMixin):
+    finished_before = serializers.DateTimeField(
```
bmbouter, line 10: In hand testing this it was a bit more difficult to use by this not having a default. I know there are reasons to not give it a default, like forcing the uesr to choose, but fwiw it was more difficult.

pulpcore/app/tasks/purge.py
mdellweg: Oh, right now, we can make this an admin interface only by virtue of the access_policy.

pulpcore/app/viewsets/task.py
```diff
@@ -90,6 +96,12 @@ class TaskViewSet(
                 "effect": "allow",
                 "condition": "has_model_or_obj_perms:core.change_task",
             },
+            {
+                "action": ["purge"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_model_or_obj_perms:core.delete_task",
```
mdellweg, line 103: There's no use to check for obj_perm here. It's a bulk endpoint.
```suggestion
                "condition": "has_model_perms:core.delete_task",
```

pulpcore/app/tasks/purge.py
bmbouter: I'm ok with doing that for this PR since RBAC isn't on everywhere. One thing to consider though is that admins "pass" all permission checks so if we did move it in this PR I believe it would "just work". We can also do this as a separate PR too that I'm happy to help contribute if that's helpful. It would be nice to ship the default policy as it's written here and not limited to "super admin users". What do you think would be best?

pulpcore/app/viewsets/task.py
mdellweg: Requiring this permission on the model level kind of sounds like restricting to the "tasks super admins".

ggainey:
Let me go back the Katello Gang and get some feedback on this discussion - there's a number of good points on both sides here, maybe they'll have a stronger need/desire for one way or the other!

ggainey:
> I was looking at this thinking we should not have this run in a task, but maybe we should? I wrote this script:

Check-in with the Katello Gang says they're OK with it being async, esp given the data-sizes they sometimes see.

pulpcore/app/serializers/purge.py
ggainey: Modified to have a default of "30 days before today", based on feedback.

pulpcore/app/tasks/purge.py
ggainey: So as I understand it, the decision is "take this PR as it is, and a later PR will do work in TaskManager to address the context of this discussion" - does that match everyone's expectations?

bmbouter: Yes and after starting into that PR it should be separate. I filed this and I'm putting that PR together now, but it shouldn't block this one. https://pulp.plan.io/issues/9575

@ggainey let me know when you're ready for final review plz.

ggainey:
@bmbouter @mdellweg Was just waiting for tests to run on the changes - ready for final(?) review, kthxbai

pulpcore/tests/functional/api/using_plugin/test_purge.py
```diff
@@ -0,0 +1,235 @@
...
+        else:
+            self.fail("NO core.Task DELETIONS?!?")
+
+    @classmethod
+    def setUpClass(cls):
+        """Create repos, remotes, and api-clients for all tests."""
+        cls.cfg = config.get_config()
+        cls.client = ApiClient(configuration=cls.cfg.get_bindings_config())
+        cls.task_api = TasksApi(cls.client)
+        # cls.purge_api = TasksPurgeApi(cls.client)
```
bmbouter, line 82: should this be here?

ggainey: oh for...no. No it should not. Fixing.

pulpcore/app/viewsets/task.py
```diff
@@ -90,6 +96,12 @@ class TaskViewSet(
                 "effect": "allow",
                 "condition": "has_model_or_obj_perms:core.change_task",
             },
+            {
+                "action": ["purge"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_model_perms:core.delete_task",
```
bmbouter, line 103: I'm thinking about this more and I agree with what @mdellweg is saying, but at the same time most RBAC enabled systems won't have users receive global permissions, they'll receive local permissions..... So actually I think anyone should be able to run this and let queryset scoping handle the rest (the other PR). If we put that into practice here that means this line can be removed. (I know that sounds scary).

ggainey: Honestly, I'm struggling to understand why we think this API is so unlike the usecase for 'update' or 'destroy' that it needs to be so different.

CHANGES/8554.feature
```diff
@@ -0,0 +1,5 @@
+Added a /tasks/purge/ API to do bulk-deletion of old tasks.
```
bmbouter, line 1: I think we should launch this as tech preview. I think generally all APIs should be launched that way. They can still use it just the same.

bmbouter:
```suggestion
Added a ``/pulp/api/v3/tasks/purge/`` API to do bulk-deletion of old tasks.
```

pulpcore/app/viewsets/task.py
bmbouter: It's a good question. It's very different in the sense that update and destroy deal with a single object so the permissions that authorize that can be either "model level" or "object level" permissions. Here, this deals with a list not one item, so because of that we authorize on "object level" permissions. Yet at the same time, in practice users will have object level permissions. Does that help with my perspective at least?

bmbouter:
This is what I came up with for the user-aware queryset scoping.  https://github.com/pulp/pulpcore/pull/1725

pulpcore/app/tasks/purge.py
```diff
@@ -0,0 +1,44 @@
...
+        finished_at__lt=finished_before, state__in=states
+    ).delete()
```
bmbouter, line 25: I think we can't use my queryset filtering PR, so instead I think we should additionally filter this here before the `.delete()` call with something like:  `get_objects_for_user(current_user, "core.delete_task", qs=qs)`.

ggainey:
@bmbouter added perms-checking-code and tests - final(?) review pliz!

pulpcore/app/tasks/purge.py
```diff
@@ -0,0 +1,55 @@
...
+    by deleting a Task.
+
+    args:
+        finished_before (DateTime): Earliest finished-time to **NOT** purge.
+        states (List[str]): List of task-states we want to purge.
+
+    """
+    current_user = get_current_authenticated_user()
+    qs = Task.objects.filter(finished_at__lt=finished_before, state__in=states)
+    if current_user:
```
bmbouter, line 30: I think here we 100% of the time will have current_user and if we don't there is something wrong. I recommend removing this branch to make the errors more obvious.

pulpcore/app/tasks/purge.py
```diff
@@ -0,0 +1,55 @@
...
+    It will remove only tasks that are 'owned' by the current-user (admin-users own All The Things,
+    so admins can delete all tasks).
+
+    It will not remove tasks that are incomplete (ie, in states running|waiting|cancelling).
+
+    It reports (using ProgressReport) the total entities deleted, as well as individual counts
+    for each class of entity. This shows the results of cascading-deletes that are triggered
+    by deleting a Task.
+
+    args:
```
bmbouter, line 23: nitpick feel free to ignore
```suggestion
    Args:
```

pulpcore/app/tasks/purge.py
```diff
@@ -0,0 +1,55 @@
...
+    # This loop reports back the specific entities deleted and the number removed
+    for key in details:
+        progress_bar = ProgressReport(
+            message=_("Purged task-objects of type {}".format(key)),
+            total=details[key],
+            code="purge.tasks.key.{}".format(key),
+            done=details[key],
+            state="completed",
+        )
+        progress_bar.save()
```
bmbouter, line 55: How many entries would this produce if I deleted 1000 tasks?

pulpcore/app/tasks/purge.py
ggainey: Cool, I thought that was the case but was in "better safe than sorry" mode.

pulpcore/app/tasks/purge.py
ggainey: Purge deletes tasks, which are foreign-keyy-referenced by other entities, which results in cascade-deletes. This reports on the number of different kinds-of objects that have been deleted, as feedback "look at how much your command has cleaned out of your DB". This is  task entities, and anything dependent on them (eg, progress-reports and createdresources). 

Here's an example of what the output looks like, when deleting all the tasks from my dev-system:
```
pulp> task purge --finished-before 2022-01-01
Started background task /pulp/api/v3/tasks/418270cc-bbe2-4644-8131-f3cd04772c45/
Done.
pulp> task show --href /pulp/api/v3/tasks/418270cc-bbe2-4644-8131-f3cd04772c45/
{
  "pulp_href": "/pulp/api/v3/tasks/418270cc-bbe2-4644-8131-f3cd04772c45/",
  "pulp_created": "2021-12-06T19:45:52.935659Z",
  "state": "completed",
  "name": "pulpcore.app.tasks.purge.purge",
  "logging_cid": "4506ecf1d71f42b980da3429b347fb5e",
  "started_at": "2021-12-06T19:45:52.974333Z",
  "finished_at": "2021-12-06T19:45:53.077405Z",
  "error": null,
  "worker": "/pulp/api/v3/workers/48f2dccb-f2a3-4b47-8127-39b294847f93/",
  "parent_task": null,
  "child_tasks": [],
  "task_group": null,
  "progress_reports": [
    {
      "message": "Purged task-objects total",
      "code": "purge.tasks.total",
      "state": "completed",
      "total": 74,
      "done": 74,
      "suffix": null
    },
    {
      "message": "Purged task-objects of type core.CreatedResource",
      "code": "purge.tasks.key.core.CreatedResource",
      "state": "completed",
      "total": 2,
      "done": 2,
      "suffix": null
    },
    {
      "message": "Purged task-objects of type core.ProgressReport",
      "code": "purge.tasks.key.core.ProgressReport",
      "state": "completed",
      "total": 28,
      "done": 28,
      "suffix": null
    },
    {
      "message": "Purged task-objects of type core.UserRole",
      "code": "purge.tasks.key.core.UserRole",
      "state": "completed",
      "total": 22,
      "done": 22,
      "suffix": null
    },
    {
      "message": "Purged task-objects of type core.Task",
      "code": "purge.tasks.key.core.Task",
      "state": "completed",
      "total": 22,
      "done": 22,
      "suffix": null
    }
  ],
  "created_resources": [],
  "reserved_resources_record": []
}
```

bmbouter: I'm on the fence about including all that, but it sounds fine to me. Thanks for the example.

bmbouter (approved):
This is great, thanks!

CHANGES/8554.feature
```diff
@@ -0,0 +1,8 @@
+Added a ``/tasks/purge/`` API to do bulk-deletion of old tasks.
+
+Over time, the database can fill with task-records. This API allows
+an installation to bulk-remove records based on their completion
+timestamps.
+
+NOTE: this endpoint is in tech-preview and may change in backwards
+incompatible ways in the future.
```
mdellweg, line 8: Do we want that note in the CHANGELOG forever?

pulpcore/app/serializers/purge.py
mdellweg: Thinking about this now, for automation it would be much easier to specify "delete everything older than x days".

pulpcore/app/viewsets/task.py
```diff
@@ -90,6 +96,13 @@ class TaskViewSet(
                 "effect": "allow",
                 "condition": "has_model_or_obj_perms:core.change_task",
             },
+            # 'purge' affects only tasks-owned-by-user at the queryset level, and needs
+            # no extra protections here
```
mdellweg, line 100: I would prefer a note like "Purge is scoped by core.delete_task.", because you can easily create a role with only that permission for a task cleanout service account.

pulpcore/app/viewsets/task.py
```diff
@@ -139,6 +152,28 @@ def get_serializer_class(self):
             return TaskCancelSerializer
         return super().get_serializer_class()
 
+    @extend_schema(
+        description=(
+            "Trigger an asynchronous task that deletes completed tasks that finished prior"
+            " to a specified timestamp."
```
mdellweg, line 158: Here should be the tech preview warning imho.

CHANGES/8554.feature
ggainey: When things comes out of tech-preview there's another issue to un-tech-preview it. This is not breaking new ground.

pulpcore/app/serializers/purge.py
ggainey: That's one possible usecase. Another is "we purge everything thru the last day of the month", which makes "number of days" variable. A timestamp lets the admin be exactly as specific as they want, and allows for situations like "I need to get rid of the 100K tasks from the script that went a little bonkers 4 hoiurs ago", which days-since makes impossible. And for automation, getting "30 days ago" from a current-datetime is a trivial exercise.

pulpcore/app/viewsets/task.py
ggainey: It would be nice if we had some standard around why and when and how to mark something as "tech preview". Sometimes it's in the viewset, sometimes the serializer, sometimes it's only in docs. I'm not completely convinced this API should even **be** tech-preview - it's a completely straightforward utility API.

I'm not against adding it here - I just dislike the inefficiency of having to find out "oh you need to say tech preview here instead of there, and here in addition" at the last minute based on someone's opinion, rather than being able to do it correctly in the first place.

mdellweg: I hear you. The same is true for "Oh I need to find all those places where this nifty feature may still be advertised as tech preview." once it is declared stable.

mdellweg (approved):

pulpcore/app/serializers/purge.py
```diff
@@ -0,0 +1,19 @@
...
+        help_text=_("Earliest finished-timestamp to **NOT** purge. Format '%Y-%m-%d[T%H:%M:%S]'"),
```
bmbouter, line 12: I alread LGTM'd this but could this be stated simpler:
```suggestion
        help_text=_("Purge tasks completed earlier than this timestamp. Format '%Y-%m-%d[T%H:%M:%S]'"),
```

ggainey: :+1:

bmbouter: I you push lmk and we can merge.

mdellweg:
Let's see if https://github.com/pulp/pulp-cli/pull/431 picks up that work tomorrow. :smile:

PR #1725, opened 2021-11-19
Adds queryset scoping by default for Task queries

bmbouter:
closes #9575

### DIFF
--- a/pulpcore/app/models/task.py
+++ b/pulpcore/app/models/task.py
@@ -11,6 +11,7 @@ from django.contrib.postgres.fields import ArrayField
 from django.db import connection, models
 from django.utils import timezone
 from django.conf import settings
+from django_currentuser.middleware import get_current_authenticated_user
 
 from pulpcore.app.models import (
     AutoAddObjPermsMixin,
@@ -20,6 +21,7 @@ from pulpcore.app.models import (
 )
 from pulpcore.constants import TASK_CHOICES, TASK_FINAL_STATES, TASK_STATES
 from pulpcore.exceptions import AdvisoryLockError, exception_to_dict
+from pulpcore.app.role_util import get_objects_for_user
 from pulpcore.tasking.constants import TASKING_CONSTANTS
 
 _logger = logging.getLogger(__name__)
@@ -139,6 +141,22 @@ def _uuid_to_advisory_lock(value):
     return ((value >> 64) ^ value) & 0x7FFFFFFFFFFFFFFF
 
 
+class TaskManager(models.Manager):
+    """
+    Scopes initial queryset to only Tasks the user has the `core.view_task` permission on.
+
+    If there is no user available, the usual queryset is return that does *not* provide user
+    isolation.
+    """
+
+    def get_queryset(self):
+        qs = super().get_queryset()
+        current_user = get_current_authenticated_user()
+        if current_user:
+            return get_objects_for_user(current_user, "core.view_task", qs=qs)
+        return qs
+
+
 class Task(BaseModel, AutoDeleteObjPermsMixin, AutoAddObjPermsMixin):
     """
     Represents a task
@@ -186,6 +204,8 @@ class Task(BaseModel, AutoDeleteObjPermsMixin, AutoAddObjPermsMixin):
     reserved_resources_record = ArrayField(models.CharField(max_length=256), null=True)
 
     ACCESS_POLICY_VIEWSET_NAME = "tasks"
+
+    objects = TaskManager()
 
     def __str__(self):
         return "Task: {name} [{state}]".format(name=self.name, state=self.state)
--- a/CHANGES/9575.misc
+++ b/CHANGES/9575.misc
@@ -1,0 +1,1 @@
+Introduced a queryset scoping qs default on the Task manager for the ORM.
--- a/pulpcore/app/viewsets/task.py
+++ b/pulpcore/app/viewsets/task.py
@@ -67,7 +67,6 @@ class TaskViewSet(
     minimal_serializer_class = MinimalTaskSerializer
     filter_backends = (OrderingFilter, DjangoFilterBackend)
     ordering = "-pulp_created"
-    queryset_filtering_required_permission = "core.view_task"
 
     DEFAULT_ACCESS_POLICY = {
         "statements": [


bmbouter:
I had wanted to have this manager fail if a user is not specified, but there are a variety of places where we refer to Task.objects where there legitimately is no user. I then added a class level `objects_no_qs_scoping = models.Manager()` similar to [what this does](https://docs.djangoproject.com/en/3.2/topics/db/managers/#modifying-a-manager-s-initial-queryset). The idea is calls where there should be no user can specifically use `Task.objects_no_qs_scoping` and that would be fine.

I was able to do that in places like the `ImportSerializer` [here](https://docs.djangoproject.com/en/3.2/topics/db/managers/#modifying-a-manager-s-initial-queryset) and the `ExportSerializer` [here](https://github.com/pulp/pulpcore/blob/b50780d93c2b6b070caf478b1a43c1dae8376abf/pulpcore/app/serializers/exporter.py#L83). The issue I ran into was with django-filters which is not able to have it's default queryset able to be switched from `Task.objects` to `Task.objects_no_qs_scoping`. That would be ideal long-term I think though. I was getting the traceback below:

```
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/core/management/__init__.py", line 395, in execute
    django.setup()
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/apps/registry.py", line 122, in populate
    app_config.ready()
  File "/home/vagrant/devel/pulpcore/pulpcore/app/apps.py", line 191, in ready
    super().ready()
  File "/home/vagrant/devel/pulpcore/pulpcore/app/apps.py", line 96, in ready
    self.import_viewsets()
  File "/home/vagrant/devel/pulpcore/pulpcore/app/apps.py", line 129, in import_viewsets
    from pulpcore.app.viewsets import NamedModelViewSet
  File "/home/vagrant/devel/pulpcore/pulpcore/app/viewsets/__init__.py", line 65, in <module>
    from .task import TaskViewSet, TaskGroupViewSet, WorkerViewSet  # noqa
  File "/home/vagrant/devel/pulpcore/pulpcore/app/viewsets/task.py", line 30, in <module>
    class TaskFilter(BaseFilterSet):
  File "/home/vagrant/devel/django-filter/django_filters/filterset.py", line 71, in __new__
    new_class.base_filters = new_class.get_filters()
  File "/home/vagrant/devel/django-filter/django_filters/filterset.py", line 351, in get_filters
    filters[filter_name] = cls.filter_for_field(field, field_name, lookup_expr)
  File "/home/vagrant/devel/pulpcore/pulpcore/app/viewsets/base.py", line 546, in filter_for_field
    f = super().filter_for_field(field, name, lookup_expr)
  File "/home/vagrant/devel/django-filter/django_filters/filterset.py", line 372, in filter_for_field
    field, lookup_type = resolve_field(field, lookup_expr)
  File "/home/vagrant/devel/django-filter/django_filters/utils.py", line 196, in resolve_field
    query = model_field.model._default_manager.all().query
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/db/models/manager.py", line 156, in all
    return self.get_queryset()
  File "/home/vagrant/devel/pulpcore/pulpcore/app/models/task.py", line 155, in get_queryset
    return get_objects_for_user(current_user, "core.view_task", klass=qs)
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/guardian/shortcuts.py", line 542, in get_objects_for_user
    if with_superuser and user.is_superuser:
AttributeError: 'NoneType' object has no attribute 'is_superuser'
```

I think the issue in django-filter is that it uses `query = model_field.model._default_manager.all().query` instead of `query = model_field.model.objects.all().query`. Anyway I'm leaving it at this for now.

bmbouter:
I did file this question in the django-filter discussions area:  https://github.com/carltongibson/django-filter/discussions/1457

dkliban (approved):

pulpcore/app/models/task.py
```diff
@@ -11,6 +11,8 @@
 from django.db import connection, models
 from django.utils import timezone
 from django.conf import settings
+from django_currentuser.middleware import get_current_authenticated_user
+from guardian.shortcuts import get_objects_for_user
```
mdellweg, line 15: Please import this form pulpcore.app.role_util

bmbouter: @mdellweg fixed, ty!

bmbouter:
@mdellweg after switching the import none of the tests pass. It runs for me locally though.... Any idea?

bmbouter:
Actually I can see this new interface is not a drop-in replacement.... https://github.com/pulp/pulpcore/runs/4292058131?check_suite_focus=true#step:13:83030

mdellweg:
> @mdellweg after switching the import none of the tests pass. It runs for me locally though.... Any idea?

I don't see, why it would work locally, but using it with `qs=qs` instead of `klass=qs` should do the trick.

bmbouter:
Thanks @mdellweg. I pushed a change which uses the `qs=qs` so we'll see if that resolves it.

bmbouter:
@mdellweg it worked!

bmbouter:
Actually in thinking this over some more, which yes it does provide Queryset scoping, it doesn't correctly do what I thought I needed from it. I wanted the bulk-delete endpoint to "automatically" honor permissions, but this only honors the "core.view_task" while the bulk delete needs to handle "core.delete_task".  Hmmmmm

bmbouter:
Maybe we could override the TaskManager.delete() to also chain on a `get_objects_for_user(current_user, "core.delete_task", qs=qs)`. That could also be a terrible idea lol

bmbouter:
After thinking this over, I determined this approach isn't yielding the benefits I thought it would. I had hoped we could put this type of functionality into the managers so that with each queryset we didn't have to think about what are the right permissions for this queryset. I realize now though that it's unavoidable. The approach in this PR has many problems (see below):

* The queryset scoping is based on the "view" permission, but we don't know that what that queryset will be used for. For example if I run `Task.objects.all().delete()` yes I've filtered on the `view_task` but I have not filtered on `view_delete`.

* If I add `view_delete` filtering to `delete()` I'm requiring the user to have both permissions because the `all()` filters on `view_task` and `delete()` filters on `delete_task` which is double filtering, but in some situations may not be correct because it's possible (although unlikely) a user will have `delete_task` and not `view_task`.

* Same problem regarding mixing `bulk_update`

* Also developers would sidestep this solution entirely if they `Task.objects.first().delete()`. That's not the manager anymore. Similarly if they call `save()`.

So basically this was supposed to let developers not think in depth about what permissions are right for them, but I think it's unavoidable and we need to have each queryset think about the permissions it needs to enforce and do so.

As such, I'm closing this PR.

PR #1729, opened 2021-11-23
Fixes Artifact redownloading bug

bmbouter:
The `sync_to_async_iterable` wraps the Artifact queryset, but unlike
querysets, it can't be reused. This causes subsequent iterations
through it to not actually iterate.

closes #9542

### DIFF
--- a/CHANGES/9542.bugfix
+++ b/CHANGES/9542.bugfix
@@ -1,0 +1,1 @@
+Fixed bug where Artifacts were being downloaded even if they were already saved in Pulp.
--- a/pulpcore/plugin/sync.py
+++ b/pulpcore/plugin/sync.py
@@ -1,6 +1,8 @@
 from asgiref.sync import sync_to_async
 
 
+# This is useful for querysets which don't have async support yet. Django querysets issue a db call
+# when the iterator for them is requested, so we need that to be wrapped in `sync_to_async` also.
 iter_async = sync_to_async(iter)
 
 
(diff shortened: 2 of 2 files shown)


bmbouter:
What would be really great is if [we could make these interfaces](https://github.com/pulp/pulpcore/blob/master/pulpcore/plugin/sync.py#L7-L32) prevent reuse.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -94,12 +94,12 @@ async def run(self):
             for digest_type, digests in artifact_digests_by_type.items():
                 query_params = {"{attr}__in".format(attr=digest_type): digests}
                 existing_artifacts_qs = Artifact.objects.filter(**query_params)
-                existing_artifacts = sync_to_async_iterable(existing_artifacts_qs)
                 await sync_to_async(existing_artifacts_qs.touch)()
                 for d_content in batch:
                     for d_artifact in d_content.d_artifacts:
                         artifact_digest = getattr(d_artifact.artifact, digest_type)
                         if artifact_digest:
+                            existing_artifacts = sync_to_async_iterable(existing_artifacts_qs)
```
goosemania, line 102: Good catch!
Does it make sense to use it directly in the `for` statement (if it fits the length) to minimize the probability that this variable will be reused with some code changes later?
` async for result in sync_to_async_iterable(existing_artifacts_qs):`

bmbouter: What I'd really like to do is prevent this somehow in those interfaces directly. I gotta do some testing locally to see if I can get that safer. I'd prefer to keep the assignment of `existing_artifacts_qs` as is for readability if we can.

goosemania (approved):

dralley (approved):

pulpcore/plugin/stages/artifact_stages.py
mdellweg: Another thought: Can we make the return value of `sync_to_async_iterable` reusable?

mdellweg: I came up with (as long, as it's ok to call `iter` on querysets synchronously in async code):
```python
def sync_to_async_iterable(sync_iterable):
    class _Wrapper:
        def __aiter__(self):
            self.sync_iterator = iter(sync_iterable)
            return self

        async def __anext__(self):
            return await next_async(self.sync_iterator)

    return _Wrapper()
```

mdellweg: And it isn't ok to call `iter` on querysets in async contexts.
This seems to work however:
```python
def sync_to_async_iterable(sync_iterable):
    class _Wrapper:
        def __aiter__(self):
            self.sync_iterator = None
            return self

        async def __anext__(self):
            if self.sync_iterator is None:
                self.sync_iterator = await iter_async(sync_iterable)
            return await next_async(self.sync_iterator)

    return _Wrapper()
```

pulpcore/download/http.py
```diff
@@ -156,6 +156,7 @@ def __init__(
             kwargs (dict): This accepts the parameters of
                 :class:`~pulpcore.plugin.download.BaseDownloader`.
         """
+        log.warning(f"Downloading url: {url}")
```
dkliban, line 159: I don't think you meant for this to log at warning level.

dralley: Debug seems appropriate

pulpcore/plugin/stages/artifact_stages.py
bmbouter: Thanks @mdellweg for this! It's the bases for this fix now, and I attributed you as the commit author.

PR #1731, opened 2021-11-26
Proxy Group model to add creation_hooks capability

mdellweg:
This creates a proxy model core.Group for auth.Group that is able to
use pulpcore's creation_hooks mechanism.

fixes #9588


mdellweg:
This is introducing a subtle change in the apidocs, where the field `auth_group_href` is now `group_href`.
I can either try to change that back with drf-spectacular, or we can agree, that the `auth_` prefix was never right on a pulpcore viewset.

pulpcore/app/role_util.py
```diff
@@ -36,7 +37,7 @@ def assign_role(rolename, entity, obj=None):
     except Role.DoesNotExist:
         raise BadRequest(_("The role '{}' does not exist.").format(rolename))
     if obj is not None:
-        ctype = ContentType.objects.get_for_model(obj)
+        ctype = ContentType.objects.get_for_model(obj, for_concrete_model=False)
```
gerrod3, line 40: Before this change we wouldn't have been able to add roles to any other proxy model in pulpcore or the plugins correct?

gerrod3 (approved):
Looks good. Can you add the Required PR for your CLI pr so that the CI passes?

pulpcore/app/role_util.py
```diff
@@ -271,7 +272,7 @@ def get_users_with_perms_roles(
 def get_users_with_perms_attached_perms(
     obj, with_superusers=False, with_group_users=True, only_with_perms_in=None
 ):
-    ctype = ContentType.objects.get_for_model(obj)
+    ctype = ContentType.objects.get_for_model(obj, for_concrete_model=False)
```
bmbouter, line 275: Is this because we're actually fetching the Group model somehow? If it's for the user model I figured it would be not be needed since the user model isn't changed with this PR.

gerrod3: I think we have to add this because of how this method is implemented based on what is said in the Django docs: https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/#django.contrib.contenttypes.models.ContentTypeManager.get_for_model, but I'm not entirely sure.

pulpcore/app/role_util.py
mdellweg: yes

bmbouter (approved):
Thanks!

dkliban (approved):
This looks right to me.

bmbouter:
The failure is due to a CLI PR not being merged yet. I'm admin merging this so that we can merge the CLI PR this also needs soon too.

PR #1733, opened 2021-11-29
Pass environment variables to the signing script.

ipanova:
[noisuse]

### DIFF
--- a/docs/plugins/reference/metadata-signing.rst
+++ b/docs/plugins/reference/metadata-signing.rst
@@ -21,6 +21,13 @@ The ``SigningService`` base class already provides the fully implemented ``sign(
 the ``validate()`` method (which must be implemented by each subclass), and the ``save()`` method (which
 calls the ``validate()`` method, but is otherwise fully implemented).
 
+.. note::
+
+    The ``sign()`` function will be calling the provided script to give the administrator the
+    freedom to define, how the signature is obtained. It is their responsibility to setup the
+    software or hardware facilities for signing and make the script use them. The plugin writer
+    however should provide a reasonably easy default script based on e.g. a simple call to ``gpg``.
+
 In order to sign metadata, plugin writers are required to call the ``sign()`` method of the signing service
 being used. This method invokes the signing script (or other executable) which is provided by the
 administrator who instantiates a concrete signing service. Instantiating/creating a concrete signing service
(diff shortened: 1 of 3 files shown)


mdellweg:
Can you explain, why this change is needed?

ipanova:
> Can you explain, why this change is needed?

This is needed for the subclassed container signing service. When signing it will not call into the provided script but skopeo command will be called directly.
```
completed_process = subprocess.run(
                ["skopeo", "standalone-sign", manifest_path, reference, self.key_fingerprint, "--output", sig_path],
                stdout=subprocess.PIPE,
                stderr=subprocess.PIPE,
            )`
```

Since the specs for the created container signature are pretty strict, we cannot allow users create a signing service that will be using some script and producing a signature format pulp-registry will not support.

mdellweg:
I think, we should still call into a script. And the validate is in charge to verify that the output format is valid. The same kind of conformance requirements apply to all other uses of signing services.
With the help of calling into a script, we can make sure, Pulp keeps out of the business of handling secret keys. We should provide useful example scripts for sure.

docs/plugins/reference/metadata-signing.rst
```diff
@@ -21,6 +21,13 @@ The ``SigningService`` base class already provides the fully implemented ``sign(
 the ``validate()`` method (which must be implemented by each subclass), and the ``save()`` method (which
 calls the ``validate()`` method, but is otherwise fully implemented).
 
+.. note::
```
ipanova, line 24: @mdellweg copy/pasted your note

docs/plugins/reference/metadata-signing.rst
```diff
@@ -21,6 +21,13 @@ The ``SigningService`` base class already provides the fully implemented ``sign(
...
+
+    The ``sign()`` function will be calling the provided script to give the administrator the
+    freedom to define, how the signatuire is obtained. It is his responsibility to setup the
```
mdellweg, line 27:
```suggestion
    freedom to define, how the signature is obtained. It is his responsibility to setup the
```

mdellweg (commented):
Just needs a `CHANGES/plugin/...feature`, and then should be ready to go.

mdellweg (approved):

docs/plugins/reference/metadata-signing.rst
ipanova: when i read this typo, i read it with a french accent which makes me want to leave it as is :D

docs/plugins/reference/metadata-signing.rst
```diff
@@ -21,6 +21,13 @@ The ``SigningService`` base class already provides the fully implemented ``sign(
 the ``validate()`` method (which must be implemented by each subclass), and the ``save()`` method (which
 calls the ``validate()`` method, but is otherwise fully implemented).
 
+.. note::
+
+    The ``sign()`` function will be calling the provided script to give the administrator the
+    freedom to define, how the signature is obtained. It is his responsibility to setup the
```
goosemania, line 27: s/his/their/

goosemania (approved):
Just one small docs related comment.

dkliban (approved):

PR #1737, opened 2021-11-30
Fix unittests

bmbouter:
SimpleTestCase does not allow database access.

[noissue]

(cherry picked from commit 342587a8732085108a74d38a5aa7b302e8f6d36d)


bmbouter:
I moved these over to this PR since we need all of these to merge the backport anyway. https://github.com/pulp/pulpcore/pull/1736

PR #1738, opened 2021-11-30
Fix unittests

bmbouter:
SimpleTestCase does not allow database access.

[noissue]

(cherry picked from commit 342587a8732085108a74d38a5aa7b302e8f6d36d)


dkliban (approved):

bmbouter:
We're going to admin merge this because there are several problems with this branch and fixing those is not the priority right now because we don't need to release 3.11.z. What we do need to do is unblock later releases which use 3.11 for upgrade tests and those fail without this fix.

PR #1745, opened 2021-12-01
Do not imply module permissions on objects

mdellweg:
According to Django docs [0] has_perm called with an object should not
honour model level permissions. When needed these need to be checked
independently.

[0] https://docs.djangoproject.com/en/3.2/ref/contrib/auth/#django.contrib.auth.models.User.has_perm

[noissue]

### DIFF
--- a/pulpcore/tests/unit/roles/test_roles.py
+++ b/pulpcore/tests/unit/roles/test_roles.py
@@ -50,9 +50,9 @@ class UserRoleTestCase(TestCase):
     def test_user_role(self):
         assign_role("role1", self.user)
         self.assertTrue(self.user.has_perm("core.view_repository"))
-        self.assertTrue(self.user.has_perm("core.view_repository", self.repository))
+        self.assertFalse(self.user.has_perm("core.view_repository", self.repository))
         self.assertEqual(self.user.get_all_permissions(), {"core.view_repository"})
-        self.assertEqual(self.user.get_all_permissions(self.repository), {"core.view_repository"})
+        self.assertEqual(self.user.get_all_permissions(self.repository), set())
         remove_role("role1", self.user)
 
     def test_group_object_role(self):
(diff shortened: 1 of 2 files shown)


gerrod3 (approved):
Looks good.

bmbouter (approved):
This looks right to me. Thanks @mdellweg !

pulpcore/tests/unit/roles/test_roles.py
```diff
@@ -66,17 +66,17 @@ def test_group_object_role(self):
     def test_group_role(self):
         assign_role("role2", self.group)
         self.assertTrue(self.user.has_perm("core.view_remote"))
-        self.assertTrue(self.user.has_perm("core.view_remote", self.remote))
+        self.assertFalse(self.user.has_perm("core.view_remote", self.remote))
```
dkliban, line 69: so what's the difference between having a core.view_remote permission and core.view_remote on a specific remote?

bmbouter: If it's on a specific object, in this case a remote, it means that you have the ability to view that object, but not other instances of the same object type. If you have the `core.view_remote` permission not on a specific object, you are able to view all remote instances.

dkliban: So does the Remote viewset always check if the user has "core.view_remote" for that specific remote or if he has the general "core.view_remote" permission?

bmbouter: Yes and typically it will use the `has_model_or_obj_perms` to provide on check that includes an OR based check of both levels of permissions. Here's an in-place example for Tasks https://github.com/pulp/pulpcore/blob/fcb355c1f8fec6780524c010cf5bba7e1aa1d51b/pulpcore/app/viewsets/task.py#L79

bmbouter: Note, these tests are a bit contrived because a remote is a Master object, so we wouldn't really be using `core.view_remote` instead we'd be using `file.view_fileremote` or something like that.

dkliban: Thank you for the explanation @bmbouter

dkliban (approved):

pulpcore/tests/unit/roles/test_roles.py
mdellweg: > 

And yes, this was unintuitive to me too. But the django interface (link in the commit message) defines it that way.

PR #1746, opened 2021-12-01
backports Artifact redownloading bug

bmbouter:


template_config.yml
```diff
@@ -1,10 +1,10 @@
 # This config represents the latest values used when running the plugin-template. Any settings that
 # were not present before running plugin-template have been added with their default values.
 
-# generated with plugin_template@2021.08.26-19-g88bcdd3
+# generated with plugin_template@2021.08.26-39-gb86dbd0
 
 additional_repos:
-- branch: main
+- branch: 1.10
```
mdellweg, line 7: Do you think this could be handled by the "create release branch" workflow in the future?
It always complicates the first backport after an incompatible change in pulp_file.

bmbouter: I do think an improvement would be good.

dkliban:
you should rebase this PR

mdellweg (approved):
LGTM

PR #1747, opened 2021-12-02
Handles closed db connections

dkliban:
When the authentication middleware was added in pulpcore 3.15, it became the first place
in the content app that made an attempt to use the database. As a result, it is a convinient
place to handle InterfaceError and  OperationalError which are raised when the database
connection has been closed. When this occurs, Handler._reset_db_connection() is called to
clean up the database connection and the middleware tries to use the database again.

If the database connection is closed later in the handling of the request by the content app,
the user will still get a 500 error. However, the next request will be handled properly.

This patch also adds a call to Handler._reset_db_connection() inside the heartbeat method.

fixes: #9276
https://pulp.plan.io/issues/9276


CHANGES/9598.bugfix
```diff
@@ -0,0 +1 @@
+Fixed bug where the content app would stop working after a brief loss of connection to the database.
```
mdellweg, line 1: Shouldn't it be a backport? Or is the fix on main completely different?

mdellweg:
This PR is made against a stable branch. So i expect the commits to be generated by the cherry-pick-script, which will clearly identify what has been cherry-picked, what was the original issue and what backport issue is fixed.

CHANGES/9598.bugfix
dkliban: This is a backport. I did not use the cherry-pick script. In addition to this backport, I had to backport a few other commits.

bmbouter:
I think we do want the clarity the backport script provides, but you can also do these commands manually to ammend your commit to look like what it would produce.

CHANGES/9598.bugfix
bmbouter: This updated version is what I expected. Thanks!

bmbouter (approved):
This looks great. Having done some backports recently, each of these changes looks right to me. Thanks! :1st_place_medal:

PR #1751, opened 2021-12-03
Enable caching for the synchronous context

lubosmj:
ref #8806

### DIFF
--- a/pulpcore/cache/cache.py
+++ b/pulpcore/cache/cache.py
@@ -3,6 +3,10 @@ import json
 
 from functools import wraps
 
+from django.http import HttpResponseRedirect, HttpResponse, FileResponse as ApiFileResponse
+
+from rest_framework.request import Request as ApiRequest
+
 from aiohttp.web import FileResponse, Response, HTTPSuccessful, Request
 from aiohttp.web_exceptions import HTTPFound
 
@@ -34,7 +38,8 @@ def connection_error_wrapper(func):
         """Handle connection errors, specific to the sync context, raised by the Redis client."""
         try:
             return func(*args, **kwargs)
-        except ConnectionError:
+        except (ConnectionError, TypeError):
+            # TypeError is raised when an invalid port number for the Redis connection is configured
             return None
 
     return wrapper
@@ -48,7 +53,8 @@ def aconnection_error_wrapper(func):
         """Handle connection errors, specific to the async context, raised by the Redis client."""
         try:
             return await func(*args, **kwargs)
-        except AConnectionError:
+        except (AConnectionError, TypeError):
+            # TypeError is raised when an invalid port number for the Redis connection is configured
             return None
 
     return wrapper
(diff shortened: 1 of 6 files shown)


lubosmj:
@gerrod3, would you mind reviewing this PR?

pulpcore/cache/cache.py
```diff
@@ -257,7 +380,10 @@ async def make_entry(self, key, base_key, handler, args, expires=86400):
             entry["type"] = "FileResponse"
         elif isinstance(response, (Response, HTTPSuccessful)):
             body = response.body
-            entry["text"] = getattr(body, "_value", body).decode("utf-8")
+            if isinstance(body, bytearray):
+                entry["body"] = body.hex()
```
gerrod3, line 384: I'm confused by this, I thought the body was either bytes or a StringPayload. Why are you specifically setting this to hex just for bytearray? Is this a pulp_container specific?

pulpcore/cache/cache.py
```diff
@@ -112,6 +117,118 @@ def delete(self, key=None, base_key=None):
...
+        """Gets the response for the request and try to turn it into a cacheable entry"""
+        try:
+            response = handler(*args, **kwargs)
+        except (HTTPSuccessful, HTTPFound) as e:
+            response = e
+
+        status = response.status_code
+        entry = {"headers": dict(response.headers), "status": status}
+        response.headers["X-PULP-CACHE"] = "MISS"
+        if status == 302:
```
ipanova, line 201: Both HttpFound and and HttpResponseRedirect have 302 status code, how they will be distinguished?

pulpcore/cache/cache.py
```diff
@@ -112,6 +117,118 @@ def delete(self, key=None, base_key=None):
...
+            entry["redirect_to"] = str(response.headers["Location"])
+            entry["type"] = "Redirect"
+        elif isinstance(response, FileResponse):
```
ipanova, line 204: should this be ApiFileResponse?

pulpcore/cache/cache.py
lubosmj: Yes, this adjustment was made because of this: https://github.com/pulp/pulp_container/blob/0a8fe8b5d7c3eb5649283212465a298f7c84057d/pulp_container/app/registry.py#L249. I can address the change in pulp_container and use a different body type instead.

pulpcore/cache/cache.py
lubosmj: `HttpFound` is used within the async context while `HttpResponseRedirect` is not.

pulpcore/cache/cache.py
```diff
@@ -112,6 +117,118 @@ def delete(self, key=None, base_key=None):
...
+            base_key: a string to group entries under, defaults to DEFAULT_BASE_KEY,
+                      can be a callable taking the request and cache instance as arguments
+            expires_ttl: length in seconds entries should live in the cache, EXPIRES_TTL is default
+            keys: a list of CacheKeys to use for key creation upon entry placement, path is default
+            auth: a callable to check authorization of the request; takes the request, cache
+                  instance, and base_key as arguments.
+        """
+        super().__init__()
+        self.default_base_key = base_key or self.default_base_key
+        self.keys = keys or (CacheKeys.path,)
```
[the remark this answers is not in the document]
lubosmj: Because when I was working on this PR, the change https://github.com/pulp/pulpcore/commit/c79da9cbb5dc2637eb3331e72723f933dbff80f5 was not merged and I decided to stick to the initial definition for AsyncContentCache.

pulpcore/cache/cache.py
lubosmj: Yes. :disappointed:

pulpcore/cache/cache.py
```diff
@@ -112,6 +117,118 @@ def delete(self, key=None, base_key=None):
...
+        if status == 302:
+            entry["redirect_to"] = str(response.headers["Location"])
+            entry["type"] = "Redirect"
+        elif isinstance(response, FileResponse):
+            entry["path"] = str(response._path)
+            entry["type"] = "FileResponse"
+        elif isinstance(response, (Response, HTTPSuccessful)):
+            body = response.body
+            entry["text"] = getattr(body, "_value", body).decode("utf-8")
+            entry["type"] = "Response"
```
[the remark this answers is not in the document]
lubosmj: Yes, this is wrong. I did not test it. I will write a test for it if we decide to enable caching for one of the pulpcore's default endpoints (which does not make sense; in pulp_container we serve content from api-app).

pulpcore/cache/cache.py
```diff
@@ -234,6 +351,10 @@ async def make_response(self, key, base_key):
         if not entry:
             return None
         entry = json.loads(entry)
+
+        if binary := entry.pop("body", None):
+            entry["body"] = bytes.fromhex(binary)
+
```
gerrod3, line 357: Still confused by this.

gerrod3 (commented):
Looks good, I still want to know what is up with the `hex` call.

pulpcore/cache/cache.py
ipanova: leaving a comment in the code would be good.

ipanova (approved):

gerrod3 (approved):
I've read the comment on the `hex` calls and it makes sense. If the response contains bytes we can't assume those bytes are 'utf-8' encoded, so converting to hex in order json encode the value for the cache makes sense.

PR #1752, opened 2021-12-03
Redis not used by default

pavelpicka:
Pulp does not use redis if not configured by user.

closes: https://pulp.plan.io/issues/9475

### DIFF
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -105,6 +105,10 @@ AUTHENTICATION_BACKENDS
 Redis Settings
 --------------
 
+.. warning::
+
+   To enable usage of Redis the `CACHE_ENABLED`_ option must be set to `True`.
+
 The following Redis settings can be set in your Pulp config:
 
   * REDIS_URL
@@ -118,12 +122,12 @@ Below are some common settings used for Redis configuration.
 REDIS_HOST
 ^^^^^^^^^^
 
-   The hostname for Redis. By default Pulp will try to connect to Redis on localhost.
+   The hostname for Redis.
 
 REDIS_PORT
 ^^^^^^^^^^
 
-   The port for Redis. By default Pulp will try to connect to Redis on port 6379.
+   The port for Redis.
 
 REDIS_PASSWORD
 ^^^^^^^^^^^^^^
@@ -205,7 +209,7 @@ CACHE_ENABLED
    .. note:: This feature is provided as a tech-preview
 
    Store cached responses from the content app into Redis. This setting improves the performance
-   of the content app under heavy load for similar requests. Defaults to ``True``.
+   of the content app under heavy load for similar requests. Defaults to ``False``.
 
    .. note::
      The entire response is not stored in the cache. Only the location of the file needed to
--- a/docs/installation/instructions.rst
+++ b/docs/installation/instructions.rst
@@ -158,6 +158,11 @@ Redis
 Pulp can use Redis to cache requests to the content app. This can be installed on a different host
 or the same host that Pulp is running on.
 
+.. note::
+
+    Despite its huge performance improvement, Pulp doesn't use Redis by default
+    and must be configured manually.
+
 To install Redis, refer to your package manager or the
 `Redis download docs <https://redis.io/download>`_.
 
--- a/pulpcore/app/redis_connection.py
+++ b/pulpcore/app/redis_connection.py
@@ -8,6 +8,8 @@ _a_conn = None
 
 
 def _get_connection_from_class(redis_class):
+    if not settings.get("CACHE_ENABLED"):
+        return None
     redis_url = settings.get("REDIS_URL")
     if redis_url is not None:
         return redis_class.from_url(redis_url)
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -200,8 +200,8 @@ DATABASES = {
 
 # Redis default config
 REDIS_URL = None
-REDIS_HOST = "localhost"
-REDIS_PORT = 6379
+REDIS_HOST = None
+REDIS_PORT = None
 REDIS_DB = 0
 REDIS_PASSWORD = None
 REDIS_SSL = False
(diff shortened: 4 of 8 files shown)


lubosmj:
It seems like your changes caused Redis to drop out from our test suite completely. Maybe, we should revise the code in a way that will enable us to start Pulp without Redis (similarly to the commit https://github.com/pulp/plugin_template/commit/581012c0b5fe74e46c4e018b0b9ad32f40a1199f; yet, with the changes applicable to a different runner where Redis will not be even installed). What do you think?

CHANGES/9475.feature
```diff
@@ -0,0 +1 @@
+Pulp doesn't connect to redis by default.
```
lubosmj, line 1:
```suggestion
Allowed Pulp installations without Redis.
```

pavelpicka:
@lubosmj Next step after this PR will be to adjust installer and CI installation to use redis again. This change is more intended to make Pulp easier to integrate e.g. installation in k8s (where require redis by default will harder the installation and setup).

pulpcore/app/settings.py
```diff
@@ -198,15 +198,6 @@
...
 }
 
-# Redis default config
-REDIS_URL = None
-REDIS_HOST = "localhost"
-REDIS_PORT = 6379
-REDIS_DB = 0
-REDIS_PASSWORD = None
-REDIS_SSL = False
-REDIS_SSL_CA_CERTS = None
```
fao89, line 208: If you drop all these variables, user will need to provide all of them

pulpcore/app/redis_connection.py
```diff
@@ -11,6 +11,8 @@ def _get_connection_from_class(redis_class):
     redis_url = settings.get("REDIS_URL")
     if redis_url is not None:
         return redis_class.from_url(redis_url)
+    elif not settings.get("REDIS_HOST") and not settings.get("REDIS_PORT"):
```
fao89, line 14: you would need check all the variables below (REDIS_SSL for example) if you don't have a default value for them

pulpcore/app/settings.py
```diff
@@ -311,6 +302,19 @@
 
 _logger = getLogger(__name__)
 
+# Post-dyna-conf check if user provided a redis connection when enabling cache
+try:
+    if CACHE_ENABLED:
+        REDIS_URL or (REDIS_HOST and REDIS_PORT)
```
fao89, line 308: you would need to check all the REDIS variables if you don't have a default value for them

pulpcore/tests/functional/api/test_status.py
```diff
@@ -21,7 +21,6 @@
             "type": "object",
             "properties": {"connected": {"type": "boolean"}},
         },
-        "redis_connection": {"type": "object", "properties": {"connected": {"type": "boolean"}}},
```
fao89, line 24: `redis_connection` still there, I don't think this line needs to be dropped

pulpcore/tests/functional/api/utils.py
```diff
@@ -37,7 +37,6 @@ def get_redis_status():
 
     try:
         is_redis_connected = status_response["redis_connection"]["connected"]
-    except KeyError:
+    except (KeyError, TypeError):
```
fao89, line 40: what can raise a TypeError here?

pavelpicka: if `is_redis_connected` returns None, you can hit `TypeError: NoneType ... has no attribute "redis_connection"`

pulpcore/app/settings.py
fao89: I would go with the first option

fao89 (approved):

docs/installation/instructions.rst
```diff
@@ -158,6 +158,11 @@ Redis
 Pulp can use Redis to cache requests to the content app. This can be installed on a different host
 or the same host that Pulp is running on.
 
+.. note::
+
+    Despite it is huge performance improvement, Pulp doesn't use REDIS by default
```
gerrod3, line 163:
```suggestion
    Despite its huge performance improvement, Pulp doesn't use Redis by default
```

pulpcore/tests/functional/api/utils.py
gerrod3: Weird that you would get a type error here. I think that means the status response failed to give valid json, Pulp itself might not be available, not just the cache.

pulpcore/tests/unit/test_cache.py
```diff
@@ -7,7 +7,7 @@
 
 try:
     is_redis_connected = Cache().redis.ping()
-except ConnectionError:
+except (ConnectionError, AttributeError):
```
gerrod3, line 10: This will be removed or modified in a future PR correct? I assume this exception here indicates that the cache isn't being tested anymore.

pulpcore/app/redis_connection.py
gerrod3: Why is this needed? Don't we have the check in the settings file to ensure both of these values are set if the cache is enabled? If anything I think there should be an earlier check in this function to see if the cache is enabled and return None if not.

docs/configuration/settings.rst
```diff
@@ -105,6 +105,10 @@ AUTHENTICATION_BACKENDS
 Redis Settings
 --------------
 
+.. warning::
+
+   To enable usage of redis the `CACHE_ENABLED`_ option must be enabled.
```
gerrod3, line 110:
```suggestion
   To enable usage of Redis the `CACHE_ENABLED`_ option must be set to ``True``.
```

pulpcore/app/redis_connection.py
pavelpicka: We need this as we can generate Redis connection from two sources: user can specify only REDIS_URL or REDIS_HOST and REDIS_PORT. so we need check if any of options is set.

pavelpicka: changed to `CACHE_ENABLED` check

gerrod3 (approved):

dkliban (approved):

PR #1753, opened 2021-12-05
Update CI files

pulpbot:
[noissue]


.github/workflows/scripts/stage-changelog-for-default-branch.py
```diff
@@ -0,0 +1,64 @@
+# WARNING: DO NOT EDIT!
```
mdellweg, line 1: The automatic update does not delete deprecated files. Should it?

mdellweg (approved):

bmbouter (approved):

PR #1754, opened 2021-12-06
Add endpoint to reset an access policy

mdellweg:
[noissue]

### DIFF
--- a/CHANGES/9606.feature
+++ b/CHANGES/9606.feature
@@ -1,0 +1,1 @@
+Added a ``reset`` endpoint to the access policy viewset to revert to the provided default uncustomized access policy.
--- a/pulpcore/app/viewsets/access_policy.py
+++ b/pulpcore/app/viewsets/access_policy.py
@@ -1,10 +1,15 @@
 from django_filters.rest_framework import filters
 from rest_framework import mixins
+from rest_framework.decorators import action
+from rest_framework.response import Response
+from drf_spectacular.utils import extend_schema
 
+from pulpcore.app.apps import pulp_plugin_configs
 from pulpcore.app.models import AccessPolicy
 from pulpcore.app.serializers import AccessPolicySerializer
 from pulpcore.app.viewsets import BaseFilterSet, NamedModelViewSet
 from pulpcore.app.viewsets.base import NAME_FILTER_OPTIONS
+from pulpcore.app.util import get_view_urlpattern
 
 
 class AccessPolicyFilter(BaseFilterSet):
@@ -34,3 +39,28 @@ class AccessPolicyViewSet(
     endpoint_name = "access_policies"
     ordering = "viewset_name"
     filterset_class = AccessPolicyFilter
+
+    @extend_schema(request=None)
+    @action(detail=True, methods=["post"])
+    def reset(self, request, pk=None):
+        """
+        Reset the access policy to its uncustomized default value.
+        """
+
+        access_policy = self.get_object()
+        for plugin_config in pulp_plugin_configs():
+            for viewset_batch in plugin_config.named_viewsets.values():
+                for viewset in viewset_batch:
+                    if get_view_urlpattern(viewset) == access_policy.viewset_name:
+                        default_access_policy = viewset.DEFAULT_ACCESS_POLICY
+                        access_policy.statements = default_access_policy["statements"]
+                        access_policy.creation_hooks = default_access_policy.get(
+                            "creation_hooks"
+                        ) or default_access_policy.get("permissions_assignment")
+                        access_policy.customized = False
+                        access_policy.save()
+                        serializer = AccessPolicySerializer(
+                            access_policy, context={"request": request}
+                        )
+                        return Response(serializer.data)
+        raise RuntimeError("Viewset for access policy was not found.")


pulpcore/app/viewsets/access_policy.py
```diff
@@ -34,3 +39,29 @@ class AccessPolicyViewSet(
     endpoint_name = "access_policies"
     ordering = "viewset_name"
     filterset_class = AccessPolicyFilter
+
+    @extend_schema(request=None)
+    @action(detail=True, methods=["post"])
+    def reset(self, request, pk=None):
```
bmbouter, line 45: I forgot the `AccessPolicyViewSet` doesn't have an access policy itself. Is that something we're planning to add later? This isn't related to your work here and more of a general question.

mdellweg: Well that makes it an admin-only facility. I think that is still reasonable.

goosemania:
As a user, I created my own access policy and I have user/group permissions configured accordingly.
Now I decided to reset and use default a policy and it's my problem to set all perms for users properly.
What I'm concerned is that it seems there is no historical record of the previous policy and if a user doesn't have a record somewhere of previous policy, it would be hard to troubleshoot or set new permissions properly. 
Can we do something about it and be more user friendly? E.g. log the policy before resetting it, or maybe document steps for the case when the policy is reset with existing users and perms? or both

CHANGES/tba.feature
```diff
@@ -0,0 +1 @@
+Added a ``reset`` endpoint to access policies.
```
mdellweg, line 1: remove that

pulpcore/app/viewsets/access_policy.py
```diff
@@ -34,3 +39,29 @@ class AccessPolicyViewSet(
...
+        Reset the access policy to its uncustomized default value.
+
+        """
+
+        access_policy = self.get_object()
+        for plugin_config in pulp_plugin_configs():
+            for viewset_batch in plugin_config.named_viewsets.values():
+                for viewset in viewset_batch:
+                    if get_view_urlpattern(viewset) == access_policy.viewset_name:
+                        default_access_policy = viewset.DEFAULT_ACCESS_POLICY
```
gerrod3, line 56: In the future is there a possibility we will allow users to create access policies for views that don't already have a `DEFAULT_ACCESS_POLICY`? If so this would error.

CHANGES/9606.feature
```diff
@@ -0,0 +1 @@
+Added a `reset` endpoint to the access policy view set to revert to the provided default uncustomized access policy.
```
gerrod3, line 1:
```suggestion
Added a ``reset`` endpoint to the access policy viewset to revert to the provided default uncustomized access policy.
```

gerrod3 (commented):
Remove WIP from the PR title.

pulpcore/app/viewsets/access_policy.py
bmbouter: I didn't imagine there would be any viewsets that "only an admin can ever call these". I think I conclude that because an admin will be authorized anyway so we should ship a default policy that assumes meaningful (non-admin RBAC) usage. Maybe I need to reevaluate how I approach this, but that's where I'm at anyway. I wanted to share what I was thinking, but we don't have to do any of that now either way.

mdellweg: My thought here is: Whoever can manipulate the access policies owns the system, and that should be restricted to "real" admins (the ones that respond to `user.is_admin()`). No? But yes, this could be written in an accesspolicy too.
And now the admin could lock himself out of the system completely by forgetting to add statements to any of these calls.
As much as i follow the thought >> I didn't imagine there would be any viewsets that "only an admin can ever call these". << , this viewset needs special considerations, imho.

pulpcore/app/viewsets/access_policy.py
mdellweg: Would it not be correct to err in that situation?
Would you rather delete the existing AP to mimick the nonexistance of the DEFAULT_AP?

pulpcore/app/viewsets/access_policy.py
```diff
@@ -34,3 +39,29 @@ class AccessPolicyViewSet(
     endpoint_name = "access_policies"
     ordering = "viewset_name"
     filterset_class = AccessPolicyFilter
+
+    @extend_schema(request=None)
+    @action(detail=True, methods=["post"])
+    def reset(self, request, pk=None):
+        """
+        Reset the access policy to its uncustomized default value.
+
```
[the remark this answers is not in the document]
mdellweg: I can try. I never understood or internalized the rules of doc string linters.

pulpcore/app/viewsets/access_policy.py
bmbouter: :+1: Yeah let's think it over. Perhaps this should stay "special"?

bmbouter (approved):
Thanks!

gerrod3 (approved):

bmbouter:
> As a user, I created my own access policy and I have user/group permissions configured accordingly. Now I decided to reset and use a default policy and it's my problem to set all perms for users properly. What I'm concerned is that it seems there is no historical record of the previous policy and if a user doesn't have a record somewhere of previous policy, it would be hard to troubleshoot or set new permissions properly. Can we do something about it and be more user friendly? E.g. log the policy before resetting it, or maybe document steps for the case when the policy is reset with existing users and perms? or both

I would like to see docs mostly as the solution, both user-facing and api auto-docs. Part of the issue is we have little to no docs right now that are user facing except the API auto-generated ones. I hope we can add a lot of docs maybe as early as friday, but not in this PR. I'd like those docs can identify to users they should have a backup of this data. Also maybe the docstring for the viewset could encourage a backup also. Wdyt and when should we do it by?

mdellweg:
> > As a user, I created my own access policy and I have user/group permissions configured accordingly. Now I decided to reset and use a default policy and it's my problem to set all perms for users properly. What I'm concerned is that it seems there is no historical record of the previous policy and if a user doesn't have a record somewhere of previous policy, it would be hard to troubleshoot or set new permissions properly. Can we do something about it and be more user friendly? E.g. log the policy before resetting it, or maybe document steps for the case when the policy is reset with existing users and perms? or both
> 
> I would like to see docs mostly as the solution, both user-facing and api auto-docs. Part of the issue is we have little to no docs right now that are user facing except the API auto-generated ones. I hope we can add a lot of docs maybe as early as friday, but not in this PR. I'd like those docs can identify to users they should have a backup of this data. Also maybe the docstring for the viewset could encourage a backup also. Wdyt and when should we do it by?

FWIW: If you use the patch endpoint on access_policies, the backup is already completely on you. My suggestion is you should have your access policies in version control and apply using e.g. Ansible. (Shameless plug: https://github.com/pulp/squeezer/blob/develop/plugins/modules/access_policy.py)

goosemania:
> > As a user, I created my own access policy and I have user/group permissions configured accordingly. Now I decided to reset and use a default policy and it's my problem to set all perms for users properly. What I'm concerned is that it seems there is no historical record of the previous policy and if a user doesn't have a record somewhere of previous policy, it would be hard to troubleshoot or set new permissions properly. Can we do something about it and be more user friendly? E.g. log the policy before resetting it, or maybe document steps for the case when the policy is reset with existing users and perms? or both
> 
> I would like to see docs mostly as the solution, both user-facing and api auto-docs. Part of the issue is we have little to no docs right now that are user facing except the API auto-generated ones. I hope we can add a lot of docs maybe as early as friday, but not in this PR. I'd like those docs can identify to users they should have a backup of this data. Also maybe the docstring for the viewset could encourage a backup also. Wdyt and when should we do it by?

It's fine with me if it's not in this PR. I always struggle with docs approach because I do not know how to make users read them :) I do not suspect that many users leverage the custom access policy approach since RBAC is not available for all major plugins. This makes those docs less pressing but it would be good to have them when we declare RBAC support on all of our heavily used plugins.

goosemania (approved):

bmbouter:
Thanks @goosemania for the thoughts. I'll try to bring them forward into the docs push.

PR #1758, opened 2021-12-06
Adds viewset name to accesspolicy deprecations

bmbouter:
The deprecation warning about an offending AccessPolicy now includes the
name of the Viewset that has the problem.

closes #9608

### DIFF
--- a/CHANGES/plugin_api/9608.bugfix
+++ b/CHANGES/plugin_api/9608.bugfix
@@ -1,0 +1,1 @@
+Include additional information about which AccessPolicy is using deprecated policy features.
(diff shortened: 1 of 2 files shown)


bmbouter:
The warning I saw was about RBACContentGuard and with this patch it shows:

`pulp [None]: pulpcore.deprecation:WARNING: In AccessPolicy for <class 'pulpcore.app.viewsets.publication.RBACContentGuardViewSet'>,the 'permissions' field in 'creation_hooks' is deprecated and may be removed with pulpcore 3.20. Use the 'parameters' field instead.
`

mdellweg (approved):

goosemania (approved):

PR #1759, opened 2021-12-06
Improves reliability of the db restart test for content app

dkliban:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/utils.py
+++ b/pulpcore/tests/functional/api/using_plugin/utils.py
@@ -61,6 +61,12 @@ def populate_pulp(cfg, url=None):
     return client.get(FILE_CONTENT_PATH)
 
 
+def gen_pulpcore_client():
+    """Return an OBJECT for pulpcore client."""
+    configuration = config.get_config().get_bindings_config()
+    return CoreApiClient(configuration)
+
+
 def gen_file_client():
     """Return an OBJECT for file client."""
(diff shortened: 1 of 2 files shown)


bmbouter (approved):
Thanks

mdellweg:
Would it be possible to query the status endpoint to check for the service health? I'd prefer something more specific than a generic failure in the setup method.

dkliban:
> Would it be possible to query the status endpoint to check for the service health? I'd prefer something more specific than a generic failure in the setup method.

We could check the status API, but we would still need to check for an ApiException. The status API is going to return 500 error when the db is not ready. Though I still think the test would be better that way. I'll update it.

bmbouter:
I like this version even more. Double LGTM!

pulpcore/tests/functional/api/using_plugin/utils.py
```diff
@@ -61,6 +61,12 @@ def populate_pulp(cfg, url=None):
     return client.get(FILE_CONTENT_PATH)
 
 
+def gen_pulpcore_client():
```
mdellweg, line 64: I will love to see these turned into pytest fixtures.

mdellweg (approved):

pulpcore/tests/functional/api/using_plugin/utils.py
bmbouter: Me too. I think my PR work is on the way to enabling that.

PR #1767, opened 2021-12-11
Fix return value of get_all_permissions

mdellweg:
Django's interface for get_all_permissions seems to not include the
app_label in the permission strings.
We include them again in the my_permission endpoint, because that is the
way we represent them to our users everywhere.

https://docs.djangoproject.com/en/3.2/ref/contrib/auth/#django.contrib.auth.models.User.get_all_permissions

re #9411

### DIFF
--- a/pulpcore/app/viewsets/base.py
+++ b/pulpcore/app/viewsets/base.py
@@ -609,7 +609,11 @@ class RolesMixin:
     @action(detail=True, methods=["get"])
     def my_permissions(self, request, pk=None):
         obj = self.get_object()
-        return Response({"permissions": list(request.user.get_all_permissions(obj))})
+        app_label = obj._meta.app_label
+        permissions = [
+            ".".join((app_label, codename)) for codename in request.user.get_all_permissions(obj)
+        ]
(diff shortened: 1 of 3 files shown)


bmbouter:
Interestingly, this time it failed on Azure.

bmbouter (approved):
Looks good to me. Thanks!

mdellweg:
> Interestingly, this time it failed on Azure.

I told you, it's completely random.

gerrod3:
Is the way it is now break something in Django? I think the change is fine, but it's weird that if someone uses the get_all_permissions with an object they will have to manually add the code labels.

mdellweg:
> Is the way it is now break something in Django? I think the change is fine, but it's weird that if someone uses the get_all_permissions with an object they will have to manually add the code labels.

Yes, that _is_ weird, but it's the way Django "defines" the interface. When we don't adhere to it, an admin user will see both the prefixed and the non prefixed versions, because they will receive the same permissions from two backends.

gerrod3 (approved):
My comments were answered, looks good.

goosemania (approved):

PR #1774, opened 2021-12-15
Optional permissions_assignment and creation_hooks

mdellweg:
These fields were both required by the serializer, which could never
validate.

fixes #9639


bmbouter (approved):
This is great, thanks!

gerrod3 (approved):

PR #1779, opened 2021-12-16
Fix repository version listing after version deletion

gerrod3:
fixes: #9481

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -909,6 +909,9 @@ class RepositoryVersion(BaseModel):
         repo_relations.filter(version_added=self).update(version_added=next_version)
         repo_relations.filter(version_removed=self).update(version_removed=next_version)
 
+        # Update next version's counts as they have been modified
+        next_version._compute_counts()
+
     def delete(self, **kwargs):
         """
         Deletes a RepositoryVersion
--- a/pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
@@ -357,11 +357,15 @@ class AddRemoveRepoVersionTestCase(unittest.TestCase):
 
     def test_delete_middle_version(self):
         """Delete a middle version."""
-        index = randint(1, len(self.repo_version_hrefs) - 2)
+        index = randint(1, len(self.repo_version_hrefs) - 3)
         delete_version(self.repo, self.repo_version_hrefs[index])
 
         with self.assertRaises(HTTPError):
             get_content(self.repo, self.repo_version_hrefs[index])
+
+        # Check added count is updated properly
+        added = get_added_content_summary(self.repo, self.repo_version_hrefs[index + 1])
+        self.assertEqual(added["file.file"], 2)
 
         for repo_version_href in self.repo_version_hrefs[index + 1 :]:
             artifact_paths = get_artifact_paths(self.repo, repo_version_href)
--- a/CHANGES/8798.bugfix
+++ b/CHANGES/8798.bugfix
@@ -1,0 +1,1 @@
+Fix content summary showing incorrect count after previous version deletion.
--- a/CHANGES/9481.bugfix
+++ b/CHANGES/9481.bugfix
@@ -1,0 +1,1 @@
+Fixed issue with listing repository versions after deleting previous versions.


pulpcore/app/models/repository.py
```diff
@@ -909,6 +909,9 @@ def _squash(self, repo_relations, next_version):
         repo_relations.filter(version_added=self).update(version_added=next_version)
         repo_relations.filter(version_removed=self).update(version_removed=next_version)
 
+        # Update next version's counts as they have been modified
+        self._compute_counts(next_version)
```
mdellweg, line 913: would `next_version._compute_counts()` have worked too?

daviddavis:
I think this might also fix this bug I reported some months back https://pulp.plan.io/issues/8798. Or maybe they're the same bug.

pulpcore/app/models/repository.py
gerrod3: :facepalm:, yes.

gerrod3:
> I think this might also fix this bug I reported some months back https://pulp.plan.io/issues/8798. Or maybe they're the same bug.

I think you are right. I'll add tests later to be sure this has been fixed.

pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
```diff
@@ -357,12 +357,16 @@ def test_delete_last_version(self):
 
     def test_delete_middle_version(self):
         """Delete a middle version."""
-        index = randint(1, len(self.repo_version_hrefs) - 2)
+        index = randint(1, len(self.repo_version_hrefs) - 3)
```
mdellweg, line 360: I know, not introduced here, but using randomness in tests is really a two sided sword. You may eventually cover more cases, but you set the system up for a blinker, so it does not make a good regression. I'd rather pick a number.

mdellweg (approved):
LGTM

Just wondering that a single line is supposed to fix two different issues. Should they be marked duplicated instead?

ipanova:
> LGTM
> 
> Just wondering that a single line is supposed to fix two different issues. Should they be marked duplicated instead?

not having a strong opinion on this, but probably closing one as dup would be my preference.

ipanova (approved):

pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
gerrod3: True, but currently the test sets up each repo version as adding only one content unit, so I'm just making sure I pick a version that has another one above it, however adding on more to this test is difficult.

gerrod3:
> > LGTM
> > Just wondering that a single line is supposed to fix two different issues. Should they be marked duplicated instead?
> 
> not having a strong opinion on this, but probably closing one as dup would be my preference.

Which one should I close? 9481 will occur when a content type is no longer present after repo version deletion because the counts weren't updated. 8798 occurs every time you delete a previous repo version. 9481 was reported by a user, 8798 was noticed by David.

mdellweg:
> > > LGTM
> > > Just wondering that a single line is supposed to fix two different issues. Should they be marked duplicated instead?
> > 
> > 
> > not having a strong opinion on this, but probably closing one as dup would be my preference.
> 
> Which one should I close? 9481 will occur when a content type is no longer present after repo version deletion because the counts weren't updated. 8798 occurs every time you delete a previous repo version. 9481 was reported by a user, 8798 was noticed by David.

So they are not really duplicate errors. Then we should "Fix" them both. Does the test cover both cases? (I don't think so.) Would it be feasible to make such a test with the file plugin?

BTW: What happens if you add a content in one version, remove it the next and then delete the former version. Would this result in a version without changes, or would it burst into flames?

gerrod3:
> So they are not really duplicate errors. Then we should "Fix" them both. Does the test cover both cases? (I don't think so.) Would it be feasible to make such a test with the file plugin?

No, the issue reported by the user can only occur in plugins with more than one content type. When you list the repository versions it gives their summary report which includes the content type, but this content type is grabbed dynamically from the database each time since the summary only stores the pulp_type. So if a content type is removed through repo version deletion with a corresponding orphan cleanup, then the server will 500 error when trying to list the repo versions.

> BTW: What happens if you add a content in one version, remove it the next and then delete the former version. Would this result in a version without changes, or would it burst into flames?

You get a new version without any changes. Should I add a check to see if the updated next version is now the same as the previous version and possibly delete this duplicate repo version if so?

mdellweg:
> > BTW: What happens if you add a content in one version, remove it the next and then delete the former version. Would this result in a version without changes, or would it burst into flames?
> 
> You get a new version without any changes. Should I add a check to see if the updated next version is now the same as the previous version and possibly delete this duplicate repo version if so?

An empty changeset is probably the best we can do, because deleting the second version also would definitely break stuff. It may be referenced by a pulication/distribution, or even outside Pulp. It must not disappear without being explicitely deleted.

Having said that, i think this PR is ready to go.

PR #1782, opened 2021-12-19
Fix import and export OOM error 

hao-yu:
json.dumps method consumes a lot of memory when converting data
to json string. While processing data, it seems to create
python object which is a number of times larger than the
parsing data. For example, a 5GB data is resulting more than 20GB
consumed.
    
This commit fixes the issue by processing the dataset in batch
and write into a temporary file.
    
This commit also fixes the import OOM error by parsing only the
file object to the json.load method instead of reading whole
file into the memory.

closes #9645

### DIFF
--- a/pulpcore/app/importexport.py
+++ b/pulpcore/app/importexport.py
@@ -3,8 +3,10 @@ import io
 import json
 import tarfile
 import tempfile
+import logging
 
 from django.conf import settings
+from django.db.models.query import QuerySet
 
 from pulpcore.app.apps import get_plugin_config
 from pulpcore.app.models.progress import ProgressReport
@@ -14,7 +16,9 @@ from pulpcore.app.modelresource import (
     ContentArtifactResource,
     RepositoryResource,
 )
-from pulpcore.constants import TASK_STATES
+from pulpcore.constants import TASK_STATES, EXPORT_BATCH_SIZE
+
+log = logging.getLogger(__name__)
 
 
 def _write_export(the_tarfile, resource, dest_dir=None):
@@ -32,16 +36,37 @@ def _write_export(the_tarfile, resource, dest_dir=None):
         dest_dir str(directory-path): directory 'inside' the tarfile to write to
     """
     filename = "{}.{}.json".format(resource.__module__, type(resource).__name__)
-    dataset = resource.export(resource.queryset)
     if dest_dir:
         dest_filename = os.path.join(dest_dir, filename)
     else:
         dest_filename = filename
 
-    data = dataset.json.encode("utf8")
-    info = tarfile.TarInfo(name=dest_filename)
-    info.size = len(data)
-    the_tarfile.addfile(info, io.BytesIO(data))
+    # If the resource is the type of QuerySet, then export the data in batch to save memory.
+    # Otherwise, export all data in oneshot. This is because the underlying libraries
+    # (json; django-import-export) do not support to stream the output to file, we export
+    # the data in batches to memory and concatenate the json lists via string manipulation.
+    with tempfile.NamedTemporaryFile(dir=os.getcwd(), mode="w", encoding="utf8") as temp_file:
+        if isinstance(resource.queryset, QuerySet):
+            temp_file.write("[")
+            total = resource.queryset.count()
+            for i in range(0, total, EXPORT_BATCH_SIZE):
+                current_batch = i + EXPORT_BATCH_SIZE
+                dataset = resource.export(resource.queryset[i:current_batch])
+                # Strip "[" and "]" as we are writing the dataset in batch
+                temp_file.write(dataset.json.lstrip("[").rstrip("]"))
+                if current_batch < total:
+                    # Write "," if not last loop
+                    temp_file.write(", ")
+            temp_file.write("]")
+        else:
+            dataset = resource.export(resource.queryset)
+            temp_file.write(dataset.json)
+
+        temp_file.flush()
+        info = tarfile.TarInfo(name=dest_filename)
+        info.size = os.path.getsize(temp_file.name)
+        with open(temp_file.name, "rb") as fd:
+            the_tarfile.addfile(info, fd)
 
 
 def export_versions(export, version_info):
--- a/CHANGES/9645.bugfix
+++ b/CHANGES/9645.bugfix
@@ -1,0 +1,1 @@
+Fix import and export OOM error.
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -59,7 +59,7 @@ def _import_file(fpath, resource_class, do_raise=True):
     try:
         log.info(_("Importing file {}.").format(fpath))
         with open(fpath, "r") as json_file:
-            data = Dataset().load(json_file.read(), format="json")
+            data = Dataset().load(json_file, format="json")
             resource = resource_class()
             log.info(_("...Importing resource {}.").format(resource.__class__.__name__))
             return resource.import_data(data, raise_errors=do_raise)
--- a/pulpcore/constants.py
+++ b/pulpcore/constants.py
@@ -63,3 +63,5 @@ FS_EXPORT_CHOICES = (
     (FS_EXPORT_METHODS.HARDLINK, "Export by hardlinking"),
     (FS_EXPORT_METHODS.SYMLINK, "Export by symlinking"),
 )
+
+EXPORT_BATCH_SIZE = 2000


dkliban (changes requested):
You need to add a CHANGES/9645.bugfix file with an entry for the change log.

pulpcore/app/importexport.py
```diff
@@ -16,6 +18,24 @@
 )
 from pulpcore.constants import TASK_STATES
 
+from tablib.formats._json import serialize_objects_handler
+
+log = logging.getLogger(__name__)
```
dkliban, line 23: did you mean to leave this here? i don't see it used anywhere

hao-yu: Yes. I just thought it is handy to have it for debugging in the future. I can remove if you don't like it. :)

hao-yu:
Added CHANGES/9645.bugfix

pulpcore/app/importexport.py
```diff
@@ -16,6 +18,24 @@
...
+
+class DatasetSerializer(list):
+    def __init__(self, resource):
+        self.resource = resource
+        self.batch_size = 2000
```
hao-yu, line 28: Based on my tests for rhel 7 server rpms repo, memory usage <700MB max when batch size is set to 1000. Memory usage <1GB max when batch size is set to 2000.

hao-yu:
@dkliban you want me to fix the format?

dkliban:
@hao-yu yea, please squash the formatting changes into a single commit. You can remove it from draft.

parthaa:
I tested this patch and it worked great. I have an additional patch to this PR to fix the import OOM also
```patch
diff --git a/pulpcore/app/tasks/importer.py b/pulpcore/app/tasks/importer.py
index 7f89f88..35291e6 100644
--- a/app/tasks/importer.py
+++ b/app/tasks/importer.py
@@ -59,7 +59,7 @@ def _import_file(fpath, resource_class, do_raise=True):
     try:
         log.info(_("Importing file {}.").format(fpath))
         with open(fpath, "r") as json_file:
-            data = Dataset().load(json_file.read(), format="json")
+            data = Dataset().load(json_file, format="json")
             resource = resource_class()
             log.info(_("...Importing resource {}.").format(resource.__class__.__name__))
             return resource.import_data(data, raise_errors=do_raise)
```

hao-yu:
@dkliban Squashed changes into 1 commit.

@parthaa I added the additional patch to this PR.  Can you please double check if it is correct?

pulpcore/app/importexport.py
```diff
@@ -16,6 +18,25 @@
 )
 from pulpcore.constants import TASK_STATES
```
mdellweg, line 20: Importing from hidden interfaces is dangerous.

pulpcore/app/importexport.py
```diff
@@ -16,6 +18,25 @@
...
 
+from tablib.formats._json import serialize_objects_handler
+
+log = logging.getLogger(__name__)
+
+
```
mdellweg, line 25: Why is this inheriting list?

pulpcore/app/importexport.py
```diff
@@ -32,16 +53,26 @@ def _write_export(the_tarfile, resource, dest_dir=None):
...
-    dataset = resource.export(resource.queryset)
+    if not isinstance(resource.queryset, QuerySet):
+        dataset = resource.export(resource.queryset).dict
+    else:
+        dataset = DatasetSerializer(resource)
+
     if dest_dir:
         dest_filename = os.path.join(dest_dir, filename)
     else:
         dest_filename = filename
```
mdellweg, line 65: Please use as a context manager

pulpcore/app/importexport.py
hao-yu: Any better suggestion? I thought about copy the entire method to here?

pulpcore/app/importexport.py
hao-yu: because it needs to be json serializable such as list, tuple or dict.

pulpcore/app/importexport.py
hao-yu: which line you are referring here?

pulpcore/app/importexport.py
mdellweg: I'd prefer to make json dump aware of this type.
Inheriting from list and not even call it's `__init__` will lead to an object in a very weird state.
Can you add a `toJson` method?

pulpcore/app/importexport.py
mdellweg: Sorry, the mobile beast always annotates the wrong lines.

The NamedTemporaryFile should be used as a context manager.

pulpcore/app/importexport.py
hao-yu: I can add the super().__init__([]) with empty list to the inherited class to ensure it is called. Can you explain more about the toJson method you suggested and how you mean make json dump aware of that type?

pulpcore/app/importexport.py
hao-yu: I think there is no much point using the context manager here as doing this I need to manually flush() the buffer and use seek() and tell() to get the file size and having 2 open file handlers reading the same file.

mdellweg: It will take care of the unlink. Even in case of exceptions.

mdellweg: Would `StringIO` instead of a real file help to trigger the streaming in the encoding process?

pulpcore/app/importexport.py
mdellweg: Usually, you can extend the `jsonEncoder` or pass a default handler with `json.dump`.
Honestly, i'm unsure if any of this helps to encode the iterator in a streaming manner.

pulpcore/app/importexport.py
```diff
@@ -16,6 +18,25 @@
...
+class DatasetSerializer(list):
+    def __init__(self, resource):
+        self.resource = resource
+        self.batch_size = 2000
+        self.total = self.resource.queryset.count()
+
+    def __iter__(self):
+        for i in range(0, self.total, self.batch_size):
+            dataset = self.resource.export(self.resource.queryset[i : i + self.batch_size])
+            yield from dataset.dict
```
mdellweg, line 35: Not sure, i understand this correctly. Don't we loose the keys from that dict? Shouldn't we use `yield from dataset.dict.items()`?
The more i think about this i wish, this issue of export not being streamed could be resolve in django-import-export.

hao-yu: `dataset.dict` returns a list of ordered dicts

pulpcore/app/importexport.py
hao-yu: yup. we don't want to load every thing into memory

pulpcore/app/importexport.py
```diff
@@ -32,16 +53,26 @@ def _write_export(the_tarfile, resource, dest_dir=None):
         dest_dir str(directory-path): directory 'inside' the tarfile to write to
     """
     filename = "{}.{}.json".format(resource.__module__, type(resource).__name__)
-    dataset = resource.export(resource.queryset)
```
mdellweg, line 35: If i read the DIE code correctly, this is the line where the whole resource table will be loaded into memory and transformed into a tablib format (https://github.com/django-import-export/django-import-export/blob/main/import_export/resources.py#L938).

So what i would like to see is a way to tell export (or a new method) to stream it's results in a certain format to a file.
For our usecase we could restrict ourselves to `def stream_json(output_stream)`. Can this be done on the pulpcore subclass of `ModelResource`?

Sorry, to keep pushing you, but to me this PR is very confusing and looks brittle.

pulpcore/app/importexport.py
parthaa: @mdellweg To put it in perspective RHEL 7 export json is 5 GB.

hao-yu: 5 GB is just the file size. It is still not counting after they loaded into json objects. It will be multiplies of 5GB as I described in the bug

pulpcore/app/importexport.py
hao-yu: I had a look on `cls` and `default.` I don't see that I can utilize them in the way that could save memory. Even I override the `default` method, it still needs to return an json encodable value, such as list.

pulpcore/app/importexport.py
hao-yu: I thought about it earlier. I think this will only work if `json.dump` support the streaming too but it seems not. Unless we manually do it. Like write '[',  for each chunk of dicts  write to file, close ']'. https://stackoverflow.com/questions/24239613/memoryerror-using-json-dumps

parthaa:
side question - Will splitting json to multiple files work better ? Like keep 5K entries per file ?

pulpcore/app/importexport.py
mdellweg: Looking into the code of the json module, the iterative encoding really only works with `list`, `tuple` and `dict`, which is really unfortunate. Can you maybe try if `tuple` serves as well and add an explaining comment to it?
Still i see two issues to solve here: The aforementioned `json.dump` not working with iterators, and the fact that `resource.export` does not supply an iterator.
Not sure if the latter is better served by an all inclusive `export-to-file` method on the Model resource, or an iterator based version of `export` (needing all attached libraries like `tablib` to be made aware of iterators).
I will think about this more and come back here.

hao-yu: You mean try to inherit from tuple instead of list? Can you explain a bit what is the benefit of doing that?

mdellweg: It _is_ the same kind of hackery, but i think the tuple (being read only) brings a lot less functionality that can interfere.

hao-yu: I can change it but I don't think that will make any difference because the list or tuple is just empty. The reason that we inherit the list/tuple is just to make it iterable and fetch the records on demand. AFAIK, Json module uses the same method to encode both tuple and list and it just iterate the provided list/tuple nothing else.

pulpcore/app/importexport.py
mdellweg: I still want the tempfile to be used as a context manager. Yes we close it manually, but the `unlink` call should be done by the context cleanup.

pulpcore/app/importexport.py
```diff
@@ -14,7 +16,42 @@
...
+log = logging.getLogger(__name__)
+
+
+def _export_to_temp_file(resource):
+    """
+    Write the JSON export for the specified resource to the temporary file.
+
+    If the specified resource is the type of QuerySet, then write the json data
+    in batch to save memory. Otherwise, write all json data in oneshot.
+
```
mdellweg, line 30: Can you add something like:
"Because the underlying libraries (json; django-import-export) do not support to stream the output to file, we export the 
data in batches to memory and concatenate the json lists via string manipulation."

mdellweg (approved):
I think, this is it.
@ggainey I want your signoff too.

mdellweg:
We use black for formatting code:
https://github.com/pulp/pulpcore/runs/4764870730?check_suite_focus=true#step:6:8

ggainey (approved):
The discussion and related changes look good to me. The import-tests are clean. I have had issues testing w/ RHEL7 imp/exp on my system, but it's been due to my-environment-problems. I'm good to go w/ this!

dkliban (approved):

jlsherrill:
@pulp/pulpcore request to backport this to 3.14.z

dralley:
@jlsherrill https://github.com/pulp/pulpcore/issues/2107

PR #1800, opened 2022-01-07
Use url parameter in repository access condition

mdellweg:
fixes TBA
[noissue]

### DIFF
--- a/pulpcore/app/global_access_conditions.py
+++ b/pulpcore/app/global_access_conditions.py
@@ -1,4 +1,4 @@
-from pulpcore.app.models import Group
+from pulpcore.app.models import Group, Repository
 
 
 # Model checks
@@ -190,7 +190,7 @@ def has_remote_param_model_or_obj_perms(request, view, action, permission):
     return False
 
 
-# 'Repository' attribute checks
+# 'Repository' attribute checks for RepositoryVersionViewSet
 
 
 def has_repo_attr_obj_perms(request, view, action, permission):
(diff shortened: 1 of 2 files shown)


pulpcore/app/global_access_conditions.py
```diff
@@ -231,7 +231,10 @@ def has_repo_attr_obj_perms(request, view, action, permission):
         True if the user has the Permission named by the ``permission`` argument on the
         ``repository`` attribute at the object-level. False otherwise.
     """
-    plugin_repository = view.get_object().repository.cast()
+    if "repository_pk" in view.kwargs:
+        plugin_repository = Repository.objects.get(pk=view.kwargs["repository_pk"]).cast()
+    else:
+        plugin_repository = view.get_object().repository.cast()
```
mdellweg, line 237: I'm not sure, what we should really be doing here. If we assume that this function will only be used in access policies for viewsets nested beneath some repository viewset, it'd be save to always do the lookup via kwargs.
The benefit is that it also works for bulk views (like listing repository versions), where `get_object()` blows up.

gerrod3 (commented):
What difference do these new methods have compared to `has_repo_attr_obj_perms`? Also, that method shortens repository so should these new ones shorten it as well?

mdellweg:
The difference is, that you don not need a detailed object with a repository attribute, but you can apply it to all actions that are nested under a repository viewset. The prime example is RepositoryVersionViewset where the list action could not use the existing condition.

gerrod3 (approved):
Every plugin will need these for implementing RBAC, but I wonder if there are any other objects across Pulp and its plugins that needs similar helper methods? (Maybe exporters/importers?)

pulpcore/app/global_access_conditions.py
```diff
@@ -280,6 +280,68 @@ def has_repo_attr_model_or_obj_perms(request, view, action, permission):
     return False
 
 
+def has_repository_obj_perms(request, view, action, permission):
```
bmbouter, line 283: We don't have to import these through the plugin API do we?

bmbouter (approved):
Thanks!

pulpcore/app/global_access_conditions.py
bmbouter: @mdellweg if ^ is correct go ahead and merge plz.

mdellweg: no

PR #1807, opened 2022-01-11
Retry downloads on checksum mismatch

dralley:
closes: #9667
https://pulp.plan.io/issues/9667

### DIFF
--- a/CHANGES/9667.misc
+++ b/CHANGES/9667.misc
@@ -1,0 +1,1 @@
+Make checksum mismatches a retryable error.
--- a/pulpcore/download/http.py
+++ b/pulpcore/download/http.py
@@ -5,7 +5,11 @@ import asyncio
 import backoff
 
 from .base import BaseDownloader, DownloadResult
-from pulpcore.exceptions import TimeoutException
+from pulpcore.exceptions import (
+    DigestValidationError,
+    SizeValidationError,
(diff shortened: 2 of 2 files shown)


bmbouter:
Apologies I didn't see this until now. Small nitpick, this change I think is definitly a bugfix or feature, not `misc`. It would be good to adjust that if you agree.

Bigger question: I'm a little unsure if if this is what we want, but maybe it is? I gotta think it over. The case I worry about is when the metadata advertises a checksum/digest and yet the binary data is different. Would that ever happen? I suppose not? Can you think of a time when binary data would mutate on server sides?

dralley:
@bmbouter The problem is that in this case it wasn't repeatable, it would just occasionally happen for different packages.  I kinda hate the CDN.

That *can* happen but in that case it's 100% definitely a metadata issue.

ipanova:
I am ok to expose this as bugfix for the sake of visibility for users.

bmbouter:
I think if we make it a bugfix I'm +1. Thanks

dralley:
@fao89 We can manually cherrypick from my 3.14-cherrypicks PR (but let's review and merge that first)

fao89:
https://bugzilla.redhat.com/show_bug.cgi?id=2033174

PR #1810, opened 2022-01-14
Mark a few tests as parallel

bmbouter:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
@@ -11,6 +11,7 @@ from pulp_smash.pulp3.bindings import monitor_task
 from pulp_smash.pulp3.utils import gen_repo
 
 from requests.exceptions import HTTPError
+import pytest
 
 from pulpcore.tests.functional.api.using_plugin.utils import gen_file_remote
 from pulpcore.tests.functional.api.using_plugin.constants import (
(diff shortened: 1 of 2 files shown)


fao89 (approved):

pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
```diff
@@ -360,6 +361,7 @@ def test_headers(self):
         self.remotes_api.partial_update(self.remote.pulp_href, data)
 
 
+@pytest.mark.parallel
 class CreatePulpLabelsRemoteTestCase(unittest.TestCase):
```
mdellweg, line 365: Maybe it's a question for another day, but do they need to inherit from `unittest.TestCase`?

mdellweg (approved):

mdellweg:
Maybe we should introduce the ci changes via the plugin_template (that is when they are merged). Or they will be overruled by the next automatic update.

pulpcore/tests/functional/api/using_plugin/test_crud_repos.py
bmbouter: They don't need to, but I plan to handle that when I come through and actually port the tests to pytest in another PR.

bmbouter:
Actually I can declare these marks from pulp-smash! https://github.com/pulp/pulp-smash/pull/1268

PR #2058, opened 2022-01-17
Use github issues instead of redmine

fao89:


dralley (approved):

template_config.yml
```diff
@@ -73,7 +73,7 @@ test_fips_nightly: false
 test_performance: false
 test_released_plugin_with_next_pulpcore_release: false
 test_s3: true
-update_redmine: true
+update_redmine: false
```
ggainey, line 76: Is there a reason to keep this stanza at all any more? Is there something we should be doing in an "update_github: true" stanza instead?

bmbouter (approved):
Looks good to me. Thank you!

template_config.yml
bmbouter: Yeah we should be removing this from the plugin template +1

fao89: yep, I'll start a PR on plugin-template to remove or update it

PR #2095, opened 2022-01-17
Remove the pytest entry from pyproject.toml

bmbouter:
I introduced this, but actually I was able to move this to pulp-smash
instead so all plugins automatically receive it.

[noissue]


dkliban (approved):

bmbouter:
https://github.com/pulp/pulp-smash/pull/1268

fao89 (approved):

PR #2100, opened 2022-01-18
Stop logging calls against deprecated HTTP endpoints

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/app/views/orphans.py
+++ b/pulpcore/app/views/orphans.py
@@ -1,7 +1,6 @@
 from drf_spectacular.utils import extend_schema
 from rest_framework.views import APIView
 
-from pulpcore.app.loggers import deprecation_logger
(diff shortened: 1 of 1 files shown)


mdellweg:
We discussed, that we should stop logging that deprecation to the admin. It's really a usage error.

pulpcore/app/views/orphans.py
```diff
@@ -1,7 +1,7 @@
 from drf_spectacular.utils import extend_schema
 from rest_framework.views import APIView
 
-from pulpcore.app.loggers import deprecation_logger
+from pulpcore.app.loggers import rest_deprecation_logger
```
ipanova, line 4: this import is not needed anymore, i think.

ipanova (approved):

dralley (approved):

bmbouter:
Yay! \o/

Issue #1927, opened 2022-01-17
Provide a better abstraction for customizing downloader behavior


pulpbot:
Author: @dralley (dalley)

Redmine Issue: 7352, https://pulp.plan.io/issues/7352

---

Customization of downloader behavior is currently achieved by subclassing one of the downloader types from the Plugin API and then modifying the ```get_downloader()``` method and ```downloader_factory``` property on the Remote (sub)class.

A limitation of this abstraction is that while the ```get_downloader()``` method accepts **kwargs to pass down to the (possibly customized) downloader, it will pass all kwargs to any downloader without discriminating by the type of downloader.   Therefore it's difficult to send a kwarg to be used on a customized HTTP downloader without it being also sent to other types of downloaders such as FileDownloader which are not expecting it.

This could be worked around by additionally modifying the DownloaderFactory, but it's currently suboptimal because:

* The methods you'd want to override are underscore-prefixed (private)
* It would create extra coupling and interdependence between the classes; not the end of the world, but kind of messy

We could probably benefit from revisiting the design and incorporate some of our experience with customizing downloader behavior in various plugins.

pulpbot:
From: @dkliban (dkliban@redhat.com)
Date: 2021-01-14T15:03:24Z

---

We will not get to this in the near future.

PR #2104, opened 2022-01-19
Refactor downloader

mdellweg:


pulpcore/download/http.py
```diff
@@ -216,47 +215,27 @@ async def _handle_response(self, response):
...
-                    return await self._run(extra_data=extra_data)
-                except asyncio.TimeoutError:
-                    raise TimeoutException(self.url)
+        @backoff.on_exception(
+            backoff.expo,
+            RETRYABLE_ERRORS,
+            max_tries=self.max_retries + 1,
+            giveup=http_giveup_handler,
+        )
+        async def run_wrapper():
```
mdellweg, line 235: This would look so much nicer, if we could use the backoff as a context manager. But honestly, i have no idea how that could be possibly implemented in python.

dralley:
related: https://github.com/pulp/pulpcore/issues/1927

pulpcore/download/http.py
```diff
@@ -176,18 +188,6 @@ def __init__(
         self.max_retries = max_retries
         super().__init__(url, **kwargs)
 
-    def raise_for_status(self, response):
```
dralley, line 179: Note that some plugins are actually using this, it needs to go through a deprecation cycle at minimum - if it's not serving a use case that can't be replaced.

https://github.com/pulp/pulp_ansible/blob/da42e4c1e6045b968698d9e8e5cce1955f4b2cba/pulp_ansible/app/downloaders.py#L57
https://github.com/pulp/pulp_rpm/blob/87f59232be566bc6d71ff4cd67e92c0c8ec019d3/pulp_rpm/app/downloaders.py#L91

pulpcore/download/base.py
```diff
@@ -94,17 +95,21 @@ def __init__(
...
         """
+        if custom_file_object:
+            deprecation_logger.warn(
+                "The 'custom_file_object' argument to 'BaseDownloader' is"
+                "deprecated and will be removed in pulpcore==3.20; stop using it."
+            )
+
         self.url = url
-        self._writer = custom_file_object
+        self._writer = None
```
bmbouter, line 105: There was some reason for this originally, but I'm hazy on the details. I gotta try to remember. What I remember so far is that in the case of downloads actually triggered by the streamer in cases where `policy=streamed` the files never should actually hit the disk at all.

bmbouter: Yeah I think we just fogot to do that, but that was the idea https://github.com/pulp/pulpcore/blob/main/pulpcore/content/handler.py#L846-L885 and I think that is the intent of this part of the design.

CHANGES/plugin_api/tba.removal
```diff
@@ -0,0 +1,2 @@
+Deprecate the use of ``custom_file_object`` parameter in ``BaseDownloader``.
+Deprecate the ``finalize`` method in ``BaseDownloader``.
```
bmbouter, line 2: A recap of why we want to do this would be helpful for me.

bmbouter: Also `finalize` is used in various places so removing this from BaseDownloader would cause the content app to not be able to stream file:// which it should be able to to today...

mdellweg: Yes, i just stumbled over that piece of code. What are we even doing there? Replacing member functions on an instantiated object?

pulpcore/content/handler.py
```diff
@@ -653,7 +653,7 @@ def get_remote_artifacts_blocking():
         remote_artifacts = await sync_to_async(get_remote_artifacts_blocking)()
         for remote_artifact in remote_artifacts:
             try:
-                response = await self._stream_remote_artifact(request, response, remote_artifact)
+                response = await asyncio.shield(self._stream_remote_artifact(request, response, remote_artifact))
```
mdellweg, line 656: I am pretty sure, this is not the way we want it, but it prevents the gunicorn worker to cancel the task that _also_ saves the download to artifact. Debugging showed, that without this, the Downloader was mutilated while cleaning up.

pulpcore/content/handler.py
```diff
@@ -827,16 +827,17 @@ async def handle_response_headers(headers):
...
-                    range_bytes = range_bytes - range_start
-                    range_bytes = range_bytes - (int(value) - stop)
+                    range_bytes = stop - start
                     response.headers[name] = str(range_bytes)
 
+                    # aiohttp adds a 1 to the range.stop compared to http headers (including) to
+                    # match python array adressing (exclusive)
                     response.headers["Content-Range"] = "bytes {0}-{1}/{2}".format(
-                        start, stop - start + 1, int(value)
+                        start, stop - 1, content_length
```
ipanova, line 840: :+1: 
in pulp-container we also got burned with exclusive/inclusive content-range header :/

pulpcore/download/base.py
```diff
@@ -94,6 +96,9 @@ def __init__(
             expected_size (int): The number of bytes the download is expected to have.
             semaphore (asyncio.Semaphore): A semaphore the downloader must acquire before running.
                 Useful for limiting the number of outstanding downloaders in various ways.
+            safe_to_disk (bool): Whether the downloaded data should be saved.
```
dralley, line 99: Is this meant to be "save_to_disk" instead of "safe_to_disk"?

mdellweg: yes

dralley:
@mdellweg @bmbouter When we eventually pick this work back up, I'd like to suggest an idea for the 3.20 cycle.

I would like to move towards having a single download context that gets shared between all download invocations on the frontend and the backend, rather than creating a brand new "downloader" for each and every file.  Instead of creating a downloader and awaiting `downloader.run()`, we would create a download context once, and await various methods on that.

I think that would provide a lot of benefits:

* Make several aspects of the API cleaner - no need to manually construct or sanitize URLs in most cases
    * With mirrors support this is a bigger deal, imagine pre-configuring the downloader with a list of mirrors and just asking it to download a relative path, rather than manually constructing full URLs beforehand for every single file we may want to download
* Avoid a ton of unnecessary object set up work which we currently repeat potentially thousands or tens of thousands of times
* We could enforce the "rules" (like requests per second) apply equally to the frontend and backend
* Opens doors for implementing a lot of useful features and optimizations more easily, like understanding the total bandwidth used by the task at any given time, or spreading out the bandwidth across multiple mirrors, or ignoring mirrors that fail too many times, or applying limits to any of those things

PR #2117, opened 2022-01-19
Added a "current_task" field to the worker

### LINKED ISSUE #2034: As a user, I can see the status (idle, working) of my workers
Author: @dralley (dalley)

Redmine Issue: 9183, https://pulp.plan.io/issues/9183

---

It would be helpful if the worker API (and status API) should show the task currently executing on that worker, and/or state information such as "idle" / "working" for each worker.

It may also be a good idea for the status API to list the number of tasks currently queued up.

This would re-enable some of the functionality provided by the "rq info" and "qpid-stat" tools in the past.

dralley:
Provide the task href of the currently active task, if any, because it's useful information as an admin / support.

closes #2034
https://github.com/pulp/pulpcore/issues/2034

### DIFF
--- a/pulpcore/app/models/task.py
+++ b/pulpcore/app/models/task.py
@@ -90,6 +90,16 @@ class Worker(BaseModel):
     last_heartbeat = models.DateTimeField(auto_now=True)
 
     @property
+    def current_task(self):
+        """
+        The task this worker is currently executing, if any.
+
+        Returns:
+            Task: The currently executing task
+        """
+        return self.tasks.filter(state="running").first()
+
+    @property
     def online(self):
         """
         Whether a worker can be considered 'online'
--- a/pulpcore/app/serializers/task.py
+++ b/pulpcore/app/serializers/task.py
@@ -192,10 +192,15 @@ class WorkerSerializer(ModelSerializer):
     last_heartbeat = serializers.DateTimeField(
         help_text=_("Timestamp of the last time the worker talked to the service."), read_only=True
     )
-    # disable "created" because we don't care about it
-    created = None
+    current_task = RelatedField(
+        help_text=_(
+            "The task this worker is currently executing, or empty if the worker is not "
+            "currently assigned to a task."
+        ),
+        read_only=True,
+        view_name="tasks-detail",
+    )
 
     class Meta:
         model = models.Worker
-        _base_fields = tuple(set(ModelSerializer.Meta.fields) - set(["created"]))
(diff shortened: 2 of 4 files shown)


pulpcore/app/serializers/task.py
```diff
@@ -192,10 +192,18 @@ class WorkerSerializer(ModelSerializer):
...
+        view_name="tasks-detail",
+    )
     # disable "created" because we don't care about it
     created = None
 
     class Meta:
         model = models.Worker
-        _base_fields = tuple(set(ModelSerializer.Meta.fields) - set(["created"]))
-        fields = _base_fields + ("name", "last_heartbeat")
+        _base_fields = tuple(set(ModelSerializer.Meta.fields) - set(["pulp_created"]))
```
mdellweg, line 208: Hmm. This seems to be an oversight. But what's the reason to exclude that field anyway?

dralley: I think originally the reason was that it was completely irrelevant, I forget the details, but since this has been broken for so long (possibly since 3.0 even) I'm just going to revert to the normal logic.

I think it makes sense to keep with the current worker semantics anyway.

mdellweg (approved):

pulpcore/app/models/task.py
```diff
@@ -89,6 +89,16 @@ class Worker(BaseModel):
...
 
+    @property
+    def current_task(self):
+        """
+        The task this worker is currently executing, if any.
+
+        Returns:
+            Task: The currently executing task
+        """
+        return self.tasks.filter(state="running").first()
```
bmbouter, line 100: Isn't there a backref that can provide this more directly? Something like `self.current_task` or something?

dralley: That's what `self.tasks` is, yeah?  The backref is a foreign key not a one-to-one field so you have to do additional filtering.

mdellweg: And the information is there. I would not want to duplicate it by adding a current_task filed to the Worker model to avoid data inconsistency.

bmbouter: Oh yeah `self.tasks` is that. :+1:

bmbouter (approved):
Looks good, thanks!

dralley:
@bmbouter Do you object to these backports?

dralley:
@jlsherrill After discussion with @bmbouter, for now these will not be backported upstream.  They can be applied downstream or we can bring it back up later - it was something we've been asked for a couple of times from support.

PR #2118, opened 2022-01-19
Create systemID and post telemetry data for devs

bmbouter:
This creates a SystemID and scheduled task to periodically post a
dictionary of data including the SystemID. Currently this will only post
data for dev systems.

[noissue]


pulpcore/app/models/task.py
```diff
@@ -323,3 +327,32 @@ class CreatedResource(GenericRelationModel):
     task = models.ForeignKey(
         Task, related_name="created_resources", default=Task.current, on_delete=models.CASCADE
     )
+
+
+class PeriodicTask(LifecycleModel):
```
mdellweg, line 332: As i recall, one of the design goals of pulp3 was to stay out of the scheduler business. Do you think this can be handled by a regular task with a cron like utility?

bmbouter: We got out of the business for a few reasons, but I'll try to explain why this isn't getting back into that business.

The feature we stopped offering was offering user-editable period tasks that capture all the various parameters they wanted on the schedules they wanted. This was a highly complex feature. Also it required additional components to be deployed, e.g. celerybeat, which if users were going to have to do that they should just use cron.

This is different because these tasks are not running user workload but are instead an aspect of Pulp's internal correctness. They don't contain any parameters at all actually. We don't want to have a separate dispatcher required because then making that work everywhere (k8s, single-container, rpm installs, clustered installs, ansible installer installs, etc) is very complex. Vendoring the dispatcher in Pulp itself will have it work everywhere.

I named it PeriodicTask because it's possible Pulp may have some other periodic correctness tasks to run, so the idea was to make the modelling just a bit more general, but this could also be written exclusively for the telemetry posting periodic task.

More questions and input is welcome.

mdellweg: I see this task to be called in the content handler, but the code is interleaved with other tasking system code. This is confusing to understand.
My question is: Do we need to have a db record for this? How is it different from the heartbeating we do in the handlers already? Can we not async-run a piece of code with a configured time interval?
Maybe it get's clearer once the periodic task actually collects data.

pulpcore/app/models/task.py
```diff
@@ -323,3 +327,32 @@ class CreatedResource(GenericRelationModel):
...
+    pulp_id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    last_dispatched = models.DateTimeField(auto_now=True)
+    dispatch_interarrival = models.IntegerField()
+    task_name = models.CharField(max_length=256, unique=True)
```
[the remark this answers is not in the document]
mdellweg: Now they really start looking like tasks. :D

pulpcore/app/migrations/0089_create_system_id_and_telemetry_task_schedule.py
```diff
@@ -0,0 +1,40 @@
...
+from datetime import timedelta
+from gettext import gettext as _
+
+from django.db import migrations
+
+
+logger = logging.getLogger(__name__)
+
+
+def create_system_ID(apps, schema_editor):
```
mdellweg, line 14: I sill believe, that datamigrations are there to migrate / repair existing data. Seeding data should be moved to a post_migrate hook. You cannot be sure it exists in cases of truncated databases (`pulpcore-manager flush`). Also it makes squashing migrations harder.

bmbouter: The downside of using post_migrate hook is that now that code has to deal with both the case where it's the first time it's creating data, or if the data is already there and now it's being changed due to the seed data changing over versions of the code.

The really nice thing about migrations is that when you seed the data with it's first value, you can just create it, you know its the first time it's run. Then if it changes later, you're guaranteed the data is already there and you just have to update it. In both situations what you needed to handle is just simpler. Using a post-migrate hook, now that code has to handle all of these cases correctly no matter what that system had on it before. This does not seem better (especially when `flush` is not something I think anyone should run).

mdellweg: It's not hard to just have it in an idempotent way:
https://docs.djangoproject.com/en/3.2/ref/models/querysets/#update-or-create

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,81 @@
...
+from pulpcore.app.models import SystemID
+from pulpcore.app.models.status import ContentAppStatus
+from pulpcore.app.models.task import Worker
+
+
+logger = logging.getLogger(__name__)
+
+
+PRODUCTION_URL = "https://analytics-pulpproject-org.pulpproject.workers.dev/"
+STAGING_URL = "https://staging-analytics-pulpproject-org.pulpproject.workers.dev/"
```
bmbouter, line 19: We don't have a staging URL, I will delete this.

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,109 @@
...
+        if ".dev" in app.version:
+            return DEV_URL
+
+    return PRODUCTION_URL
+
+
+async def post_telemetry():
+    import pydevd_pycharm
+
+    pydevd_pycharm.settrace("localhost", port=29437, stdoutToServer=True, stderrToServer=True)
```
dralley, line 76: leftovers

pulpcore/app/migrations/0090_create_system_id_and_telemetry_task_schedule.py
```diff
@@ -0,0 +1,41 @@
...
+        msg = _("This Pulp system already has {num_system_ids} SystemID objects.")
+        logger.warning(msg.format(num_system_ids=num_system_ids))
+    else:
+        SystemID().save()
+
+
+def create_periodic_task(apps, schema_editor):
+    TaskSchedule = apps.get_model('core', 'TaskSchedule')
+    task_name = "pulpcore.app.tasks.telemetry.post_telemetry"
+    dispatch_interval = timedelta(minutes=10)
```
bmbouter, line 27: This is not right, this needs to be 1 day.

pulpcore/app/migrations/0090_create_system_id_and_telemetry_task_schedule.py
```diff
@@ -0,0 +1,41 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0089_systemid'),
+    ]
+
+    operations = [
+        migrations.RunPython(create_system_ID),
+        migrations.RunPython(create_periodic_task),
```
mdellweg, line 40: I still believe, we should never create data in a migration, but only migrate it.

mdellweg (changes requested):

pulpcore/app/migrations/0090_create_system_id_and_telemetry_task_schedule.py
mdellweg: yes

pulpcore/app/apps.py
```diff
@@ -100,6 +102,14 @@ def ready(self):
...
         )
         post_migrate.connect(_populate_roles, sender=self, dispatch_uid="populate_roles_identifier")
+        post_migrate.connect(
+            _populate_system_id, sender=self, dispatch_uid="populate_system_id_identifier"
+        )
+        post_migrate.connect(
+            _populate_telemetry_periodic_task,
+            sender=self,
+            dispatch_uid="populate_telemetry_periodic_task_identifier",
+        )
```
mdellweg, line 112: Is it possible to connect these hooks in the PulpAppConfig object, so they will only emit once?

pulpcore/app/apps.py
```diff
@@ -234,6 +244,33 @@ def _populate_access_policies(sender, apps, verbosity, **kwargs):
...
+    from pulpcore.app.models import TaskSchedule
+
+    if sender.name != "pulpcore.app":
+        return
+
+    task_name = "pulpcore.app.tasks.telemetry.post_telemetry"
+    dispatch_interval = timedelta(days=1)
+    name = "Post Anonymous Telemetry Periodically"
+    TaskSchedule.objects.get_or_create(
+        name=name, task_name=task_name, dispatch_interval=dispatch_interval
```
mdellweg, line 270: Should we move `task_name` and `dispatch_interval` to defaults?
Maybe even use https://docs.djangoproject.com/en/3.2/ref/models/querysets/#update-or-create to keep the record up to date.

pulpcore/app/apps.py
```diff
@@ -234,6 +244,33 @@ def _populate_access_policies(sender, apps, verbosity, **kwargs):
...
+def _populate_system_id(sender, apps, verbosity, **kwargs):
+    # circular import avoidance
+    from pulpcore.app.models import SystemID
+
+    if sender.name != "pulpcore.app":
+        return
+
+    num_system_ids = SystemID.objects.count()
+    if num_system_ids == 0:
+        SystemID().save()
```
mdellweg, line 256: Could this be `SystemID.get_or_create()` ?

pulpcore/app/apps.py
```diff
@@ -234,6 +244,33 @@ def _populate_access_policies(sender, apps, verbosity, **kwargs):
...
+
+
+def _populate_telemetry_periodic_task(sender, apps, verbosity, **kwargs):
+    # circular import avoidance
+    from pulpcore.app.models import TaskSchedule
```
mdellweg, line 261: Not sure it is needed...
```suggestion
    TaskSchedule = apps.get_model("core", "TaskSchedule")
```

pulpcore/app/apps.py
```diff
@@ -234,6 +244,33 @@ def _populate_access_policies(sender, apps, verbosity, **kwargs):
...
+
+    if sender.name != "pulpcore.app":
+        return
+
+    num_system_ids = SystemID.objects.count()
+    if num_system_ids == 0:
+        SystemID().save()
+
+
+def _populate_telemetry_periodic_task(sender, apps, verbosity, **kwargs):
```
mdellweg, line 259: Not using it...
```suggestion
def _populate_telemetry_periodic_task(sender, apps, **kwargs):
```

mdellweg (commented):
Some more suggestions...

pulpcore/app/apps.py
```diff
@@ -234,6 +244,33 @@ def _populate_access_policies(sender, apps, verbosity, **kwargs):
...
 
+def _populate_system_id(sender, apps, verbosity, **kwargs):
+    # circular import avoidance
+    from pulpcore.app.models import SystemID
+
+    if sender.name != "pulpcore.app":
+        return
+
+    num_system_ids = SystemID.objects.count()
+    if num_system_ids == 0:
```
ipanova, line 255: you could use `if not SystemID.objects.exists()` instead, i read somewhere it is more performant

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,106 @@
...
+                            data,
+                        )
+                    else:
+                        logger.warning(
+                            "Sending analytics failed with statuscode %s from %s",
+                            resp.status,
+                            url,
+                        )
+    except asyncio.TimeoutError:
+        logger.error("Timeout sending analytics to %s", url)
```
ipanova, line 104: maybe `Timed out while sending analytics`?

pulpcore/app/apps.py
bmbouter: I don't know how to do that actually. Do you?

pulpcore/app/apps.py
bmbouter: I don't think one can be because it's uuid as the pk so it would never recognize the existing one and the new one as the same.

pulpcore/app/apps.py
bmbouter: Yes this is better, ty.

pulpcore/app/apps.py
bmbouter: I didn't understand, but now I do. I made this change and it was much nicer than guarding those callables on the caller name.

bmbouter:
@mdellweg @ipanova I believe it's ready for re-review please.

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,106 @@
...
+
+async def _get_posting_url():
+    for app in pulp_plugin_configs():
+        if ".dev" in app.version:
+            return DEV_URL
+
+    return PRODUCTION_URL
+
+
+async def post_telemetry():
```
mdellweg, line 73: Despite loving the idea of running this task in an async fashion, i do not see any place yet where we attempt to gather data in parallel.
Remember this task code will run in an isolated process. It can only be parallel to itself.

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,106 @@
...
+
+    return {
+        "online_workers": {
+            "processes": online_workers_processes,
+            "hosts": online_workers_hosts,
+        },
+    }
+
+
+async def _get_posting_url():
```
mdellweg, line 65: This function would only benefit from being async if pulp_plugin_configs were.

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,106 @@
...
+            "processes": online_content_apps_processes,
+            "hosts": online_content_apps_hosts,
+        },
+    }
+
+
+async def _build_online_workers_data():
+    online_workers = Worker.objects.online_workers()
+    online_workers_processes = await sync_to_async(online_workers.count)()
+    online_workers_hosts = await _num_hosts(online_workers)
```
mdellweg, line 55: Use asyncio.gather to run these in parallel. Then again they will be serialized in the threadpool for the single db-connection to use.

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,106 @@
...
+    url = await _get_posting_url()
+
+    if url == PRODUCTION_URL:
+        return  # Initially only dev systems receive posted data. If we got here, bail.
+
+    data = {"system_id": str(system_id_entry.pk)}
+
+    data.update(await _build_versions_data())
+    data.update(await _build_online_content_apps_data())
+    data.update(await _build_online_workers_data())
```
mdellweg, line 85: asyncio.gather?

pulpcore/content/__init__.py
```diff
@@ -20,7 +20,9 @@
 )
 
 from pulpcore.app.apps import pulp_plugin_configs  # noqa: E402: module level not at top of file
-from pulpcore.app.models import ContentAppStatus  # noqa: E402: module level not at top of file
+from pulpcore.app.models import (  # noqa: E402: module level not at top of file
+    ContentAppStatus,
+)
```
mdellweg, line 25: Where did that come from?

pulpcore/app/apps.py
```diff
@@ -234,6 +245,22 @@ def _populate_access_policies(sender, apps, verbosity, **kwargs):
...
+
+
+def _populate_telemetry_periodic_task(sender, apps, **kwargs):
+    TaskSchedule = apps.get_model("core", "TaskSchedule")
+    task_name = "pulpcore.app.tasks.telemetry.post_telemetry"
+    dispatch_interval = timedelta(days=1)
+    name = "Post Anonymous Telemetry Periodically"
+    TaskSchedule.objects.update_or_create(
+        name=name, defaults={"task_name": task_name, "dispatch_interval": dispatch_interval}
+    )
```
mdellweg, line 261: :+1:

pulpcore/content/__init__.py
bmbouter: This came from black iirc, but I'm not sure why. I'll remove it and see if black wants it back.

pulpcore/app/tasks/telemetry.py
bmbouter: It was done just for convenient use of asyncio.download, but I can also add a `gather` instead since it is async.

pulpcore/app/tasks/telemetry.py
bmbouter: :+1: I'll convert to sync

pulpcore/app/tasks/telemetry.py
bmbouter: For readability I want to leave this one alone.

pulpcore/app/tasks/telemetry.py
bmbouter: I did switch this one to `gather`.

bmbouter:
@mdellweg @ipanova I pushed this latest revision if you wouldn't mind taking a look. Thanks!

mdellweg (approved):

pulpcore/app/models/telemetry.py
```diff
@@ -0,0 +1,17 @@
...
+
+from django_lifecycle import hook, LifecycleModel
+
+
+class SystemID(LifecycleModel):
+    pulp_id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    @hook("before_save")
+    def ensure_singleton(self):
+        if SystemID.objects.count() > 1:
```
ipanova, line 13: can use here exists() too

ipanova (approved):

Issue #2078, opened 2022-01-17
Downloader retry logic is broken when retrying a partially written file


fao89:
Author: @mdellweg (mdellweg)

Redmine Issue: 9673, https://pulp.plan.io/issues/9673

---

None

fao89:
From: pulpbot (pulpbot)
Date: 2022-01-17T12:13:14Z

---

PR: https://github.com/pulp/pulpcore/pull/1851

PR #2122, opened 2022-01-20
Reset size and digests when retrying downloads

mdellweg:
re #2078

### DIFF
--- a/CHANGES/2135.bugfix
+++ b/CHANGES/2135.bugfix
@@ -1,0 +1,1 @@
+Fixed bug where retries of partially downloaded files failed digest and size validation.
--- a/pulpcore/download/base.py
+++ b/pulpcore/download/base.py
@@ -103,7 +103,7 @@ class BaseDownloader:
             self.semaphore = semaphore
         else:
             self.semaphore = asyncio.Semaphore()  # This will always be acquired
-        self._digests = {n: pulp_hashlib.new(n) for n in Artifact.DIGEST_FIELDS}
+        self._digests = {}
         self._size = 0
         if self.expected_digests:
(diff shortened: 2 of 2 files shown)


pulpcore/download/base.py
```diff
@@ -157,8 +159,6 @@ async def finalize(self):
                 :meth:`~pulpcore.plugin.download.BaseDownloader.handle_data`.
         """
         self._ensure_writer_has_open_file()
-        self._writer.flush()
-        os.fsync(self._writer.fileno())
```
dralley, line 161: Why drop these two lines?

mdellweg: With closing this file in the next line i am pretty sure they are not needed.

dralley: @bmbouter You wrote this 4 years ago :)  Do you happen to remember any particular reason for it?

bmbouter: Yes, I think we want to force the OS to quiescence all data to disk and out of its buffers to avoid write-then-read inconsistency. This is more of a safety measure than an everything will fail if we don't do this. I do think it's important though.

dralley (approved):

CHANGES/2135.bugfix
```diff
@@ -0,0 +1 @@
+Fixed validation logic in HttpDownloader for retries with partially downloaded files.
```
bmbouter, line 1: Can this be rewritten in user terms? Maybe `Fixed bug where retries of partially downloaded files failed digest and size validation.`

mdellweg: @bmbouter done

bmbouter (approved):
Thanks!

fao89:
https://bugzilla.redhat.com/show_bug.cgi?id=2033174

PR #2142, opened 2022-01-20
3.17 cherrypicks

dralley:
**Why do this manually**: we have a mix of commits from before and after the issue tracker migration, and neither the cherrypick automation nor the release changelog automation can handle that.

What I did: switched the branch to the github issue tracker, then changed the commits and commit messages to reference the new github issues.

Otherwise we'd need to manage a release with only Redmine issues, then switch the issue tracker, then do another relase with any github issues. This seemed simpler since there's only a couple of commits that need to be fixed like this and we can just do it once and re-use the new commits for other branches.

@bmbouter As requested I changed the changelog type of "retry downloads on checksum mismatch" from ".misc" to ".bugfix"


fao89 (approved):

template_config.yml
```diff
@@ -1,7 +1,7 @@
 # This config represents the latest values used when running the plugin-template. Any settings that
 # were not present before running plugin-template have been added with their default values.
 
-# generated with plugin_template@2021.08.26-62-gfaa6635
+# generated with plugin_template@2021.04.08-164-gfaa6635-dirty
```
mdellweg, line 4: I'd prefer if you didn't use uncommited changes in the plugin-template. (Actually it should be an upstream commit.)

mdellweg (approved):

template_config.yml
dralley: It looks like what happened was it was initially created as such, and then I noticed the mistake and re-templated (the other watermarks are clean), but this particular file wasn't replaced because it's the same?

Not 100% sure but the CI scripts seem to be free of any noncommitted changes.

PR #2143, opened 2022-01-20
3.14 cherrypicks

dralley:
**Why do this manually:** we have a mix of commits from before and after the issue tracker migration, and neither the cherrypick automation nor the changelog automation can handle that.  

What I did: switched the branch to the github issue tracker, then changed the commits and commit messages to reference the new github issues.

Otherwise we'd need to manage a release with only Redmine issues, then switch the issue tracker, then do another relase with any github issues. This seemed simpler since there's only a couple of commits that need to be fixed like this and we can just do it once and re-use the new commits for other branches.

@bmbouter As requested I changed the changelog type of "retry downloads on checksum mismatch" from ".misc" to ".bugfix"


fao89 (approved):

dralley:
@mdellweg This includes "Keep listening on pg channels in workers" which doesn't seem to have been backported - I'm not sure if there was a reason it wasn't, let me know if it should be dropped.

mdellweg:
> @mdellweg This includes "Keep listening on pg channels in workers" which doesn't seem to have been backported - I'm not sure if there was a reason it wasn't, let me know if it should be dropped.

I think, it was never confirmed to fix a real issue we heard of and therefore we never bothered backporting.

mdellweg (approved):

PR #2145, opened 2022-01-21
Extend field filters to avoid requesting unneeded data from the DB

### LINKED ISSUE #2138: requesting lots of packages with a subset of fields still causes changelogs, filelists, requires, etc.. to be loaded, resulting in slow requests
Author: @jlsherrill (jsherril@redhat.com)

Redmine Issue: 9655, https://pulp.plan.io/issues/9655

---

When we make a request like this:

~~~
/pulp/api/v3/content/rpm/packages/?fields=pulp_href%2Cname%2Cversion%2Crelease%2Carch%2Cepoch%2Csummary%2Cis_modular%2Crpm_sourcerpm%2Clocation_href%2CpkgId&limit=1&offset=1000&repository_version=%2Fpulp%2Fapi%2Fv3%2Frepositories%2Frpm%2Frpm%2Fc6b93206-22bb-4c58-ba07-828c32326330%2Fversions%2F1%2F
~~~

where some subset of fields is requested, we're still seeing all of the fields being loaded from the database.  The result is that the query takes much much longer than it should

Testing on the rhel7 repo, it took about ~100 seconds on my machine to fetch all 32K packages specifying the fields above.  If I comment out the filelists, changelogs, provides/requires from the model and serializers, that drops down to ~33 seconds.  This is a huge improvement.

Reading this:  https://stackoverflow.com/questions/53319787/how-can-i-select-specific-fields-in-django-rest-framework

it looks like its may be possible to exclude certain fields from the query itself fairly easily.

dralley:
closes #2138

### DIFF
--- a/pulpcore/app/viewsets/base.py
+++ b/pulpcore/app/viewsets/base.py
@@ -128,6 +128,65 @@ class NamedModelViewSet(viewsets.GenericViewSet):
     schema = DefaultSchema()
     filter_backends = (StableOrderingFilter, DjangoFilterBackend)
 
+    def get_field_filters(self, request):
+        """
+        Partially sourced from django-queryfields
+        https://github.com/wimglenn/djangorestframework-queryfields
+        Copyright (c) 2016 Wim Glenn <hey@wimglenn.com>
+        """
+        # If using Django filters in the API, these labels mustn't conflict with any model
+        # field names.
+        include_arg_name = "fields"
+        # drf-queryfields default was 'fields!' which doesn't work in the bindings for some langs
+        exclude_arg_name = "exclude_fields"
+
+        # Split field names by this string.  It doesn't necessarily have to be a single character.
+        # Avoid RFC 1738 reserved characters i.e. ';', '/', '?', ':', '@', '=' and '&'
+        delimiter = ","
+
+        if request.method != "GET":
+            return (set(), set())
+
+        includes = request.query_params.getlist(include_arg_name)
+        include_field_names = {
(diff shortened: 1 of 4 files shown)


dralley:
Django Rest Framework does not make this easy.

Notable caveat to this: you can implement custom fields that do not map directly to database columns, but map to functions.  If those fields utilize database columns which you've asked to exclude (or not include) then you can accidentally slaughter the performance by causing Django to make hundreds (or thousands) of small separate queries.

dralley:
@jlsherrill Can you repeat your test with this patch before and after and see how much it helps?

dralley:
@jlsherrill Ugh, this works fine most of the time, but is completely broken for nested viewsets.

Question. Does `?minimal=True` provide all the fields you need?  If so (or if we can make it do so) I'd rather you use that and we can just can this PR.

pulpcore/app/viewsets/base.py
```diff
@@ -128,6 +128,65 @@ class NamedModelViewSet(viewsets.GenericViewSet):
...
+
+    def get_serializer(self, *args, **kwargs):
+        """
+        Return the serializer instance that should be used for validating and
+        deserializing input, and for serializing output.
+        """
+        serializer = super().get_serializer(*args, **kwargs)
+
+        _serializer = serializer
+        if hasattr(serializer, "child"):
```
mdellweg, line 174: If this is specifically for ListSerializer, then maybe check for `multiple=True` or `isinstance(ListSerializer,...)`.

PR #2149, opened 2022-01-24
Add a management command for viewing publication contents

dralley:
[noissue]

### DIFF
--- a/CHANGES/2200.feature
+++ b/CHANGES/2200.feature
@@ -1,0 +1,1 @@
+Added a new `analyze-publication` management command to facilitate debugging.
--- a/pulpcore/app/management/commands/analyze-publication.py
+++ b/pulpcore/app/management/commands/analyze-publication.py
@@ -1,0 +1,72 @@
+from gettext import gettext as _
+
+from django.core.management import BaseCommand, CommandError
+from django.urls import reverse
+
+from pulpcore.app.models import Publication, Artifact, Distribution
+from pulpcore.app.util import get_view_name_for_model
+
+
+class Command(BaseCommand):
+    """Django management command for viewing files in a publication and the artifacts on disk."""
+
+    help = _(__doc__)
+
+    def add_arguments(self, parser):
+        """Set up arguments."""
+        parser.add_argument("--publication", required=False, help=_("A publication ID."))
+        parser.add_argument(
+            "--distribution-base-path", required=False, help=_("A base_path of a distribution.")
+        )
+        parser.add_argument("--tabular", action="store_true", help=_("Display as a table"))
+
+    def handle(self, *args, **options):
+        """Implement the command."""
+
+        if options["tabular"]:
+            try:
+                from prettytable import PrettyTable
+            except ImportError:
+                raise CommandError("'prettytable' package must be installed.")
+
+        if not (options["publication"] or options["distribution_base_path"]):
+            raise CommandError("Must provide either --publication or --distribution-base-path")
+        elif options["publication"] and options["distribution_base_path"]:
+            raise CommandError("Cannot provide both --publication and --distribution-base-path")
+        elif options["publication"]:
+            publication = Publication.objects.get(pk=options["publication"])
+        else:
+            distribution = Distribution.objects.get(base_path=options["distribution_base_path"])
+            if distribution.publication:
+                publication = distribution.publication
+            elif distribution.repository:
+                repository = distribution.repository
+                publication = Publication.objects.filter(
+                    repository_version__in=repository.versions.all(), complete=True
+                ).latest("repository_version", "pulp_created")
+
+        published_artifacts = publication.published_artifact.select_related(
+            "content_artifact__artifact"
+        ).order_by("relative_path")
+        artifact_href_prefix = reverse(get_view_name_for_model(Artifact, "list"))
+
+        if options["tabular"]:
+            table = PrettyTable()
+            table.field_names = ["Apparent path", "Storage path"]
+            table.align = "l"  # left align values
+
+        for pa in published_artifacts.iterator():
+            ca = pa.content_artifact
+            path = ca.artifact.file.path if ca.artifact else None
+            artifact_id = ca.artifact_id
+            artifact_href = (artifact_href_prefix + str(artifact_id)) if artifact_id else None
+            if options["tabular"]:
+                table.add_row([pa.relative_path, path or ""])
+            else:
+                print(pa.relative_path)
+                print("└─ Storage path: {}".format(path))
+                print("└─ Artifact href: {}".format(artifact_href))
+                print()
+
+        if options["tabular"]:
+            print(table)


dralley:
Screenshot from 2022-01-24 11-17-36

Posting for discussion

pulpcore/app/management/commands/analyze-publication.py
```diff
@@ -0,0 +1,61 @@
+from gettext import gettext as _
+from tabulate import tabulate
```
dralley, line 2: @dkliban This is a new dependency, should we try to avoid it or do we forsee having enough similar tools to make it worthwhile?

pulpcore/app/management/commands/analyze-publication.py
```diff
@@ -0,0 +1,65 @@
...
+            pubs = Publication.objects.filter(repository_version__in=versions, complete=True)
+            newest_publication = pubs.latest("repository_version", "pulp_created")
+            if newest_publication.pk == publication.pk:
+                base_paths |= publication.repository.distributions.values_list(
+                    "base_path", flat=True
+                )
+        except Publication.DoesNotExist:
+            pass
+
+        print("Relative to:")
```
ipanova, line 47: some suggestions on wording..feel free to ignore.
'Distributed by' or 'Available at'

pulpcore/app/management/commands/analyze-publication.py
```diff
@@ -0,0 +1,65 @@
...
+        artifact_href_prefix = reverse(get_view_name_for_model(Artifact, "list"))
+
+        for pa in qs.iterator():
+            ca = pa.content_artifact
+            path = ca.artifact.file.path if ca.artifact else None
+            artifact_id = ca.artifact_id
+            artifact_href = (artifact_href_prefix + str(artifact_id)) if artifact_id else None
+            table.append([pa.relative_path, artifact_id, path])
+
+        print(tabulate(table, headers=["Apparent path", "Artifact", "Real path"], tablefmt="psql"))
```
ipanova, line 65: s/real path/ storage path/

ipanova, line 65: I am not sure whether it is better to introduce yet another term like 'apparent path' or just stick to 'relative path'

ipanova:
wait for subsequent request to add  `analyze repo version` ( for those plugins that do not have publication) ;)

bmbouter:
What is are the pro/cons of having this be a mgmt command versus an API itself?

dkliban:
What about starting from a base_path of a distribution instead of a publication?

dralley:
@bmbouter That is an idea - what do you imagine the API would look like?  Some vague thoughts:

* The report / json blob might be fairly sizable
* It would push the presentation aspects off to the CLI, which is more aligned with reality
* It would make it less useful for support in the short-term, because they could no longer just drop one file into the installation to get this info
* It would probably need to go on the `/publications/*` endpoint since adding an endpoint to `/distributions/*` which only acts on publications wouldn't be ideal

dralley:
Per discussion, using a management command is the safer option for now, although we can make an API out of it later if it seems generally useful.

Apart from the reasons above, this is really a debugging tool that potentially exposes certain implementation details and making it part of the API seems inappropriate.

bmbouter:
Can we have a written problem statement of what this solution is solving? I'd like to see that before we decide one way or the other if that's alright.

pmoravec:
> Can we have a written problem statement of what this solution is solving? I'd like to see that before we decide one way or the other if that's alright.

My 2c: There are frequent support issues "I got 404 error on a package request" or "my repo metadata are corrupted" or similar, where a very common divide-and-conqueror approach is to examine the given package or metadata on the disk. In `pulp-2`, this was straightforward as everything for e.g. `yum_distributor` was in `/var/lib/pulp/published/yum/master/yum_distributor/${repo}/${publication_timestamp}/` directory. But in `pulp-3`, there is no easy way to get the location information.

So I would like this tool for sure :+1: 

An idea of improvement: allow as an argument also a `base_path` of the corresponding `Distribution`. Since the user problem is often sort of "dnf for repo with `baseurl = ..` fails to get/parse this or that" - it would be nice to feed the tool with the base path without the need to find the publication id.

dralley:
@dkliban @ggainey Since we're lacking in horizontal space, I added a separate display mode which also doesn't require `prettytable` to be installed.

Screenshot from 2022-02-04 08-59-51

Should I get rid of the table output or keep it?  It's now opt-in.

pmoravec:
Possible idea to shrink width: truncate evident `/var/lib/pulp/media/artifact/` prefix. (negs: user has to prepend the prefix to have the absolute path).

dralley:
@pmoravec Would this format be a significant detriment?  It might interfere with `grep`-ing for instance although you can tell grep to provide more lines of context

pmoravec:
Since the multi-line format can be easily converted to one-line format, I dont see big difference in the outputs. The one-liner is imho better for bigger outputs (which would be majority of outputs since majority of repos have 100+ artifacts) but I understand it is tricky to impossible to squeeze all info there. So either format is good for me.

pulpcore/app/management/commands/analyze-publication.py
```diff
@@ -0,0 +1,77 @@
...
+            except ImportError:
+                raise CommandError("'prettytable' package must be installed.")
+
+        if not (options["publication"] or options["distribution_base_path"]):
+            raise CommandError("Must provide either --publication or --distribution-base-path")
+        elif options["publication"] and options["distribution_base_path"]:
+            raise CommandError("Cannot provide both --publication and --distribution-base-path")
+        elif options["publication"]:
+            publication = Publication.objects.get(pk=options["publication"])
+            repository = publication.repository
```
ipanova, line 38: this line seems to be unnecessary

pulpcore/app/management/commands/analyze-publication.py
```diff
@@ -0,0 +1,77 @@
...
+        else:
+            distribution = Distribution.objects.get(base_path=options["distribution_base_path"])
+            if distribution.publication:
+                publication = distribution.publication
+            elif distribution.repository:
+                repository = (
```
ipanova, line 44: is this needed?

pulpcore/app/management/commands/analyze-publication.py
```diff
@@ -0,0 +1,77 @@
...
+        elif options["publication"] and options["distribution_base_path"]:
+            raise CommandError("Cannot provide both --publication and --distribution-base-path")
+        elif options["publication"]:
+            publication = Publication.objects.get(pk=options["publication"])
+            repository = publication.repository
+        else:
+            distribution = Distribution.objects.get(base_path=options["distribution_base_path"])
+            if distribution.publication:
+                publication = distribution.publication
+            elif distribution.repository:
```
ipanova, line 43: can you please remind me the logic? if distribution.repository is set, does it mean that the latest repo_version is distributed and the publication for it is created?

what about the case when distribution.repo_version is set instead?

pulpcore/app/management/commands/analyze-publication.py
```diff
@@ -0,0 +1,77 @@
...
+            table.field_names = ["Apparent path", "Storage path"]
+            table.align = "l"  # left align values
+
+        for pa in published_artifacts.iterator():
+            ca = pa.content_artifact
+            path = ca.artifact.file.path if ca.artifact else None
+            artifact_id = ca.artifact_id
+            artifact_href = (artifact_href_prefix + str(artifact_id)) if artifact_id else None
+            if options["tabular"]:
+                table.add_row([pa.relative_path, path or ""])
```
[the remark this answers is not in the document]
dralley: No, there's not enough room horizontally and when it starts overflowing it looks terrible.

CHANGES/2200.feature
```diff
@@ -0,0 +1 @@
+Added a new `analyze-publication` management command to facilitate debugging.
```
ipanova, line 1: I think, in the future we could add view of contents of the repo which is distributed directly without the publication. The only caveat I foresee is that relative_path from the CA might not be that user  readable

pulpcore/app/management/commands/analyze-publication.py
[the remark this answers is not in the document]
ipanova: i see

pulpcore/app/management/commands/analyze-publication.py
dralley: >what about the case when distribution.repo_version is set instead?

I can't seem to find where we're handling that in `handler.py`, if we actually are... that might be a hole in our distribution code.

https://github.com/pulp/pulpcore/blob/main/pulpcore/content/handler.py#L446-L547

Nice catch

CHANGES/2200.feature
dralley: Couldn't the whole layout be plugin-defined though?  I'm not sure if `relative_path` from `ContentArtifact` is all that reliable

ipanova: yeah :/

pulpcore/app/management/commands/analyze-publication.py
ipanova: https://github.com/pulp/pulpcore/blob/main/pulpcore/content/handler.py#L448

dralley: Right, but what if `repository_version and not publication and  distro.SERVE_FROM_PUBLICATION`?  That branch doesn't seem to be covered and that's where the RPM plugin falls for example.

ipanova: I might be wrong and need to refresh my memory how mirror_complete works, but i think rpm repo would always have a publication, would not it?
regardless, i think there is a problem in the logic of  that line and some cases are not covered

ipanova: Actually just if `repo_version and distro.SERVE_FROM_PUBLICATION:` would suffice.
This validation would ensure that on distribution is whether repo, version or publication is set https://github.com/pulp/pulpcore/blob/main/pulpcore/app/serializers/publication.py#L236 meaning that from these 3 variablees there will be always only 1 populated https://github.com/pulp/pulpcore/blob/main/pulpcore/content/handler.py#L446-L448

ipanova: i think we circled back that the way you have it written now and the code in handler.py seems to be ok.

PR #2158, opened 2022-01-25
Adds API_ROOT setting to allow API rerooting

### LINKED ISSUE #2148: Reroot Pulp at a URL other than /pulp/api/v3/
Some users deploying Pulp need it to be able to reroot the app so that instead of serving at `/pulp/api/v3/` if can be at `/some/other/url/`

bmbouter:
Use the new API_ROOT setting to reroot Pulp.

closes #2148

### DIFF
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -60,6 +60,9 @@ SECRET_KEY = True
 # Key used to encrypt fields in the database
 DB_ENCRYPTION_KEY = "/etc/pulp/certs/database_fields.symmetric.key"
 
+# API Root
+API_ROOT = "/pulp/"
+
 # Application definition
 
 INSTALLED_APPS = [
@@ -430,3 +433,18 @@ if not (len(sys.argv) >= 2 and sys.argv[1] in _SKIPPED_COMMANDS_FOR_CONTENT_CHEC
         pass
     finally:
         connection.close()
+
+
+if not API_ROOT.startswith("/"):
+    i8ln_msg = _("The API_ROOT must start with a '/', currently it is '{API_ROOT}'")
+    raise ImproperlyConfigured(i8ln_msg.format(API_ROOT=API_ROOT))
+
+if not API_ROOT.endswith("/"):
+    i8ln_msg = _("The API_ROOT must end with a '/', currently it is '{API_ROOT}'")
+    raise ImproperlyConfigured(i8ln_msg.format(API_ROOT=API_ROOT))
+
+settings.set("V3_API_ROOT", settings.API_ROOT + "api/v3/")  # Not user configurable
+settings.set(
+    "V3_API_ROOT_NO_FRONT_SLASH", settings.V3_API_ROOT.lstrip("/")
+)  # Not user configurable
+settings.unset("API_ROOT")
--- a/pulpcore/app/templatetags/pulp_urls.py
+++ b/pulpcore/app/templatetags/pulp_urls.py
@@ -1,8 +1,9 @@
 import re
 
 from django import template
+from django.conf import settings
 from django.utils.encoding import force_text
-from django.utils.html import escape, smart_urlquote
+from django.utils.html import escape
 from django.utils.safestring import SafeData, mark_safe
 
 register = template.Library()
@@ -19,7 +20,6 @@ WRAPPING_PUNCTUATION = [
     ("&quot;", "&quot;"),
 ]
 word_split_re = re.compile(r"(\s+)")
-href_re = re.compile(r"^\/pulp\/api\/v3\/", re.IGNORECASE)
 
 
 @register.filter(needs_autoescape=True)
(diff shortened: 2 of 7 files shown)


pulpcore/app/templatetags/pulp_urls.py
```diff
@@ -58,6 +59,7 @@ def trim_url(x, limit=trim_url_limit):
             nofollow_attr = ' rel="nofollow"' if nofollow else ""
 
             if href_re.match(middle):
+                raise Exception('oh no I should be handled')
```
bmbouter, line 62: In all my using of pulp this never raises. @fao89 do you know what this `href_re` is doing for us?

fao89: no idea, I never looked at this file before,
git blame shows @dkliban as last person who touched it

bmbouter: @dkliban ^? you know

dkliban: @bmbouter I don't know what is doing for us. This filter implementation is a modified version of one that is provided by django or drf. This code is executed when we display the browsable API.

pulpcore/app/urls.py
```diff
@@ -18,7 +18,6 @@
     OrphansCleanupViewset,
     ReclaimSpaceViewSet,
 )
-from pulpcore.constants import API_ROOT
```
mdellweg, line 21: You could have used `API_ROOT = settings.V3_API_ROOT_NO_FRONT_SLASH` here.
I think this would improve readability in the rest of the file.

pulpcore/openapi/__init__.py
```diff
@@ -63,7 +64,7 @@ def _tokenize_path(self):
             if not tokenized_path and getattr(self.view, "get_view_name", None):
                 tokenized_path.extend(self.view.get_view_name().split())
 
-        path = "/".join(tokenized_path).replace("pulp/api/v3/", "")
```
mdellweg, line 66: :hand_over_mouth: this should never have been a literal string...

mdellweg (commented):
I will try to adjust my pulp-cli pr to that change now.

pulpcore/app/settings.py
```diff
@@ -430,3 +433,16 @@
...
+    i8ln_msg = _("The API_ROOT must start with a '/', currently it is '{API_ROOT}'")
+    raise ImproperlyConfigured(i8ln_msg.format(API_ROOT=API_ROOT))
+
+if not API_ROOT.endswith('/'):
+    i8ln_msg = _("The API_ROOT must end with a '/', currently it is '{API_ROOT}'")
+    raise ImproperlyConfigured(i8ln_msg.format(API_ROOT=API_ROOT))
+
+settings.set("V3_API_ROOT", settings.API_ROOT + "api/v3/") # Not user configurable
+settings.set("V3_API_ROOT_NO_FRONT_SLASH", settings.V3_API_ROOT.lstrip('/')) # Not user configurable
+settings.unset("API_ROOT")
```
bmbouter, line 448: I'm unsure if we should be actively removing this or not. I did this because I didn't want folks to reach for this settings, but maybe it should be left here? Unless someone tells me otherwise, I think we could leave this here and remove it later when we have a use case for having the raw `API_ROOT` be present. Wdyt?

pulpcore/app/templatetags/pulp_urls.py
bmbouter: I went all over the browseable API and it seems to work great with no `Exception('oh no....')` raised so I'm unclear on what this does at all. I guess I'll be removing it?

bmbouter: One thing that doesn't work at all is the `docs/` endpoint. For example on my box `API_ROOT = "/some/other/root/"` and when I browse to `/some/other/root/api/v3/docs/` it does not load and I see this error in the request log:  `django.request:WARNING: Not Found: /some/other/root/api/v3/docs/some/other/root/api/v3/docs/api.json`

pulpcore/app/urls.py
bmbouter: fixed in next push

pulpcore/app/templatetags/pulp_urls.py
fao89: https://github.com/pulp/pulpcore/blob/main/pulpcore/app/urls.py#L169

pulpcore/app/settings.py
mdellweg: Leave it as is.

mdellweg (approved):

pulpcore/app/templatetags/pulp_urls.py
bmbouter: Thanks @fao89! This is working now.

fao89:
could you please
```
$ black .
```

fao89 (commented):
docs part needs some changes:
https://github.com/pulp/pulpcore/blob/main/docs/Makefile#L64
https://github.com/pulp/pulpcore/blob/main/.github/workflows/scripts/script.sh#L29-L33

fao89:
> docs part needs some changes: https://github.com/pulp/pulpcore/blob/main/docs/Makefile#L64 https://github.com/pulp/pulpcore/blob/main/.github/workflows/scripts/script.sh#L29-L33

actually, this isn't a blocker

fao89 (approved):

mdellweg (approved):

bmbouter:
@fao89 I opened this to address the `make html` portion https://github.com/pulp/pulpcore/pull/2236

The setting of that in the CI will happen as a separate piece of work when @mdellweg enables the S3 runner to be rerooted. @mdellweg I think you'll need to set `PULP_API_ROOT` as an env var at this call https://github.com/pulp/pulpcore/blob/main/.github/workflows/scripts/script.sh#L29-L33

PR #2159, opened 2022-01-25
Fixed touch() deadlock window.

### LINKED ISSUE #2157: touch() path still has a potential deadlock window.
https://bugzilla.redhat.com/show_bug.cgi?id=2021406

ggainey:
The deadlock is difficult to capture "in the wild"; see #2157 for
details on reproducing and showing that this fixes it.

fixes #2157.
[nocoverage]

### DIFF
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -14,11 +14,12 @@ import gnupg
 
 from functools import lru_cache
 from itertools import chain
+from psycopg2 import sql
 
 from django.conf import settings
 from django.core import validators
 from django.core.files.storage import default_storage
-from django.db import IntegrityError, models, transaction
+from django.db import IntegrityError, connection, models, transaction
 from django.forms.models import model_to_dict
 from django.utils.timezone import now
 from django_lifecycle import BEFORE_UPDATE, BEFORE_SAVE, hook
@@ -104,10 +105,44 @@ class BulkTouchQuerySet(models.QuerySet):
         """
         Update the ``timestamp_of_interest`` on all objects of the query.
 
-        We order-by-pk here to avoid deadlocking in high-concurrency
-        environments.
+        Postgres' UPDATE call doesn't support order-by. This can (and does) result in deadlocks in
+        high-concurrency environments, when using touch() on overlapping data sets. In order to
+        prevent this, we choose to SELECT FOR UPDATE with SKIP LOCKS == True, and only update
+        the rows that we were able to get locks on. Since a previously-locked-row implies
+        that updating that row's timestamp-of-interest is the responsibility of whoever currently
+        owns it, this results in correct data, while closing the window on deadlocks.
         """
-        return self.order_by("pk").update(timestamp_of_interest=now())
+        # Build the list of ids we need to work on, since we're going to be building a
+        # SQL-query "by hand" in a moment.
+        pulp_ids = [f"'{uuid}'" for uuid in self.values_list("pk", flat=True)]
+        if not pulp_ids:
+            return None
+        ids_str = ",".join(pulp_ids)
+        # timestamp_of_interest exists on core_content and core_artifact, not on the Detail tables
+        # If we are an instance-of Content or its subclasses, we want to update the Content table.
+        # Otherwise, use the table associated w/ the query.
+        db_table = (
+            Content._meta.db_table if issubclass(self.model, Content) else self.model._meta.db_table
+        )
+        cursor = connection.cursor()
+        with transaction.atomic():
+            # SQL-sanitizing the table-name here is certainly overkill - sql-injection here would
+            # require code calling touch() on a Model whose table-name-str was carefully chosen to
+            # be Bad - but, good habits...
+            stmt = sql.SQL(
+                "UPDATE {table_name} "
+                "   SET timestamp_of_interest = NOW() "
+                " WHERE pulp_id IN ("
+                "    SELECT pulp_id "
+                "      FROM {table_name} "
+                "     WHERE pulp_id in ({ids}) "
+                "     ORDER BY pulp_id "
+                "     FOR UPDATE "
+                "     SKIP LOCKED)".format(table_name=sql.Identifier(db_table).string, ids=ids_str)
+            )
+            rslt = cursor.execute(stmt)
+        cursor.close()
+        return rslt
 
 
 class QueryMixin:
--- a/CHANGES/2157.bugfix
+++ b/CHANGES/2157.bugfix
@@ -1,0 +1,1 @@
+Fixed potential deadlock-window in touch() path.


pulpcore/app/models/content.py
```diff
@@ -104,10 +105,22 @@ def touch(self):
...
-        return self.order_by("pk").update(timestamp_of_interest=now())
+        To avoid deadlocking in high-concurrency environments, we lock the table we are about to
+        touch(). This is forced due to the way postgres' UPDATE call works; there is no reliable,
+        performant way to enforce the order in which rows are updated, which opens a timing
+        window during which concurrent processes affecting overlapping content can deadlock.
+        """
+        cursor = connection.cursor()
+        with transaction.atomic():
+            # This is certainly overkill - sql-injection here would require code calling touch()
+            # on a Model whose table-name-str was carefully chosen to be Bad - but, good habits
```
mdellweg, line 116: Loving the idea to avoid injections. But i'm wondering, whether `self.model._meta.db_table` may already be sanitized.

mdellweg: OK, let me just state that it took me minute to understand that by "this" you mean the `sql.Identifier` transformation. At a first glance i thought you meant the table lock, which on a second look is what's this PR is all about. Maybe you can improve on the wording?

mdellweg (approved):
Just a bit picky about the comment.

ggainey:
> Just a bit picky about the comment.

+1, will improve

pulpcore/app/models/content.py
```diff
@@ -104,10 +105,23 @@ def touch(self):
...
+        """
+        cursor = connection.cursor()
+        with transaction.atomic():
+            # The sql-sanitizing here is certainly overkill - sql-injection here would require code
+            # calling touch() on a Model whose table-name-str was carefully chosen to be Bad - but,
+            # good habits...
+            stmt = sql.SQL(
+                "LOCK TABLE {table_name} IN EXCLUSIVE MODE".format(
+                    table_name=sql.Identifier(self.model._meta.db_table).string,
+                )
```
dralley, line 121: Whoa, does this mean that every sync will lock the entire `content` table for the duration of the sub-operation which touches content?  They're going to fight each other and possibly everything else for the table lock.

This has some implications.  I don't suppose there's any way to say that we **don't** care about the order the rows are updated, which would be fine in this particular case?  As long as someone touches the content we really don't care about who it was or the precision of the timestamp being stored.

ggainey: Updating a row acquires a row-lock. Acquiring row-locks on overlapping content, in random order, **will** cause deadlocks. You can tell postgres select_for_update to skipLocks=true - but the only way to select-for-update a bulk query in django requires iterating the query and selecting the affected rows one at a time, which is HORRIBLY non-performant (you can see experiments and discussion in the associated BZ comments)

Yes, locking the whole table synchronizes content-updates and -inserts behind it. Calling touch() on large syncs (RHEL7 and 15K content, for example) takes low hundreds of milliseconds. See https://bugzilla.redhat.com/show_bug.cgi?id=2021406#c21 for discussion.

bmbouter: I don't have a good suggestion, but I'm also concerned about accepting this.

ggainey: I appreciate the concerns. 

touch() is called in three cases:
* content-upload
* repository-modify-add
* sync (twice, once in artifact- and once in content-stages)

Content-upload would hold the lock for the amt of time it takes to update one row, Modify-add would hold it for the length of time to modify N rows, where N is <= all the content in a repo. Sync is the worst case, holding the lock for the length of time to update a row for every content, and then artifact, in the repo-being-sync'd.

The bugzilla lists three options:
* select-on-update(skipLocks=true) in django - requires iterating over the django query, and is therefore orders of magnitude less performant.
* retry-on-deadlock - litters the logs with very alarming messages that will generate urgent user-issues. Also does not close the window - with bad luck, a single task can be "starved", give up, and fail where it doesn't have to. Allowing "infinite retries" is worse, because it can lead to a window where we end up hanging the server 
* lock the table (which is this PR). Serializes writes (not reads) to the affected table. Worst case: multiple syncs all finish their syncs simultaneously and issue touch() at exactly the same time, for both content- and artifact-stages. Each attempts the locks, waiting and retrying on failure. The last worker to get the lock has had to wait (number-of-workers * 2 * ~150ms) (assuming all workers were working on large repositories).  With 20 workers, all running large syncs on the same content, all of which happen to hit touch() at exactly the same time, the worst-affected sync will wait an extra 6 seconds before it gets the lock.

There's a fourth option, which is "raw sql to select-for-update skip-locks=true and touch the rows we can". This requires one query to get the list of pulp-ids for all affected rows, in pulp-id order, followed by the update-call.  This would have a memory and performance impact. 

We settled on the table-lock as the best trade-off between performance, correctness, and user-experience. It's def much preferable to the current state of "syncs fail randomly, but with noticeable frequency for a fairly common use-case".

bmbouter:
This looks really great to me. I'd also like to ask us to go just a bit further and file an upstream django that we can include in the comment. That pattern has worked really nicely so that we don't have our workaround disconnected from the upstream django issue. Please accept my thanks in advance; I know I'm asking a lot. I'd file it myself but I'm unclear on the details. Something short would be helpful even just a placeholder to get us started and not slow this mere down.

ggainey:
> This looks really great to me. I'd also like to ask us to go just a bit further and file an upstream django that we can include in the comment. That pattern has worked really nicely so that we don't have our workaround disconnected from the upstream django issue. Please accept my thanks in advance; I know I'm asking a lot. I'd file it myself but I'm unclear on the details. Something short would be helpful even just a placeholder to get us started and not slow this mere down.

No problem opening an issue with Django, but let's not hold up this PR for it. It will take some time to build a useful reproducer for them, and I don't want to file a "It's Busted" issue without one :)

pulpcore/app/models/content.py
```diff
@@ -104,10 +105,42 @@ def touch(self):
...
-        We order-by-pk here to avoid deadlocking in high-concurrency
-        environments.
-        """
-        return self.order_by("pk").update(timestamp_of_interest=now())
+        Postgres' UPDATE call doesn't support order-by. This can (and does) result in deadlocks in
+        high-concurrency environments, when using touch() on overlapping data sets. In order to
+        prevent this, we choose to SELECT FOR UPDATE with SKIP LOCKS == True, and only update
+        the rows that we were able to get locks on. Since a previously-locked-row implies
+        that updating that row's timestamp-of-interest is the responsibility of whoever currently
+        owns it, this results in correct data, while closing the window on deadlocks.
```
dralley, line 113: Big +1 to this.

pulpcore/app/models/content.py
```diff
@@ -104,10 +105,42 @@ def touch(self):
...
+        """
+        # Build the list of ids we need to work on, since we're going to be building a
+        # SQL-query "by hand" in a moment.
+        pulp_ids = ["'" + str(id) + "'" for id in self.values_list("pk", flat=True)]
```
dralley, line 117: How about `f"'{id}'"`?

pulpcore/app/models/content.py
```diff
@@ -104,10 +105,42 @@ def touch(self):
...
+                "UPDATE {table_name} "
+                "   SET timestamp_of_interest = NOW() "
+                " WHERE pulp_id IN ("
+                "    SELECT pulp_id "
+                "      FROM {table_name} "
+                "     WHERE pulp_id in ({ids}) "
+                "     ORDER BY pulp_id "
+                "     FOR UPDATE "
+                "     SKIP LOCKED)".format(table_name=sql.Identifier(db_table).string, ids=ids_str)
+            )
```
dralley, line 142: This looks great.  I'm so happy we can skip the full table locks.

pulpcore/app/models/content.py
```diff
@@ -104,10 +105,44 @@ def touch(self):
...
+        # SQL-query "by hand" in a moment.
+        pulp_ids = [f"'{uuid}'" for uuid in self.values_list("pk", flat=True)]
+        if not pulp_ids:
+            return None
+        ids_str = ",".join(pulp_ids)
+        # timestamp_of_interest exists on core_content and core_artifact, not on the Detail tables
+        # If we have a master, we want to update the Content table, not the detail-table.
+        # If we do **not** have a master, use the table associated w/ the query.
+        db_table = (
+            Content._meta.db_table if hasattr(self.model, "master") else self.model._meta.db_table
```
mdellweg, line 125: Sorry for not seeing the obvious earlier:
```suggestion
            Content._meta.db_table if isinstance(self, Content) else self.model._meta.db_table
```

`help(isinstance)`:
```
isinstance(obj, class_or_tuple, /)
    Return whether an object is an instance of a class or of a subclass thereof.
```

mdellweg (approved):
Left one last suggestion.

pulpcore/app/models/content.py
ggainey: ...omg, i r dum. Great catch!

mdellweg: To be fair, we talked about that very line and were both like "Hmm, ugly, but what else can we do?".

ggainey: Actually needed issubclass - we don't have an instance here, just the model class - but still a good improvement!

bmbouter:
This approach addresses my concerns for sure. I'm not the best at SQL so I'm not the right reviewer for a final lgtm.

One question I wanted to ask that I don't know the answer to. Is there are benefit to batching instead of one large transaction? I have no idea, but I was wondering if you all did.

mdellweg:
> This approach addresses my concerns for sure. I'm not the best at SQL so I'm not the right reviewer for a final lgtm.
> 
> One question I wanted to ask that I don't know the answer to. Is there are benefit to batching instead of one large transaction? I have no idea, but I was wondering if you all did.

Good question. My take: This started out to be a bugfix. And yet i see not only the deadlocks go away completely, but it already provides a huge performance benefit in the concurrent workflow. Let's take this work as is, and ask kindly for one more performance test with batching the pk's prior to touching. That can easily go into a separate PR, no?

Theoretizing here: It may not have an impact at all, since we already taught Postgres to not wait for any lock at all. So performing in batches, Postgres would have the same amount of work to do, and there would be no benefit of less waiting for locks.

dkliban (approved):
Thank you for all the work that went into producing this patch. This SQL query looks right to me.

mdellweg:
I believe, we want this patch also in 3.15, 3.16 and 3.17. No?

ggainey:
> I believe, we want this patch also in 3.15, 3.16 and 3.17. No?

Yes - labels added

mdellweg:
I added the PR for the failing 3.15:
https://github.com/pulp/pulpcore/pull/2221

ggainey:
> > This looks really great to me. I'd also like to ask us to go just a bit further and file an upstream django that we can include in the comment. That pattern has worked really nicely so that we don't have our workaround disconnected from the upstream django issue. Please accept my thanks in advance; I know I'm asking a lot. I'd file it myself but I'm unclear on the details. Something short would be helpful even just a placeholder to get us started and not slow this mere down.
> 
> No problem opening an issue with Django, but let's not hold up this PR for it. It will take some time to build a useful reproducer for them, and I don't want to file a "It's Busted" issue without one :)

Django issue: https://code.djangoproject.com/ticket/33516

Issue #2115, opened 2022-01-19
Queryset scoping can't be disabled by changing the `DEFAULT_PERMISSION_CLASSES`


bmbouter:
## Problem

When users change the DEFAULT_PERMISSION_CLASSES to not be `pulpcore.plugin.access_policy.AccessPolicyFromDB` they are expressing an intent to not use the default authorization of pulpcore at all. If that is changed the queryset scoping feature should also disable; currently it does not.

This is similar but not the same to [the ability for users to adjust the permissions driving `AccessPolicyFromDB`](https://github.com/pulp/pulpcore/issues/2114).

## Solution

Create an interface on `AccessPolicyfromDB` such as `objects_for_user` and have the queryset scoping implementation use that. If the `objects_for_user` isn't available, e.g. in the case of a non-pulp-based DEFAULT_PERMISSION_CLASSES, the querset scoping should just continue without any scoping. This would have the effect of allowing a change in DEFAULT_PERMISSION_CLASSES to automatically disable queryset scoping if the DEFAULT_PERMISSION_CLASSES does not support it by providing the `objects_for_user` interface.

gerrod3:
I think this change will have to come after #2114 so that we won't have to change the new method, but I think it's a good idea.

Issue #2114, opened 2022-01-19
As a user, I should be able to customize queryset scoping


bmbouter:
## Problem

Almost all aspects of RBAC are customizable via data saved/modified on the corresponding `AccessPolicy`, but in the case of queryset scoping the user cannot control what permission(s) drive that feature. Currently only the plugin writer specifies it with `queryset_filtering_required_permission`, e.g. [here](https://github.com/pulp/pulpcore/blob/1e8203f04754b83ebf6c0c7a1495b71055471b5a/pulpcore/app/viewsets/task.py#L80) also see [these plugin writer docs](https://docs.pulpproject.org/pulpcore/plugins/plugin-writer/concepts/rbac/queryset_scoping.html#enabling-queryset-scoping).

## Solution

This should be moved into the default `AccessPolicy` that way users can control it. The current queryset scoping that uses `queryset_filtering_required_permission` [here](https://github.com/pulp/pulpcore/blob/7e26096a3769c9616b0aeaad22c11085dff0b932/pulpcore/app/viewsets/base.py#L352-L354) will need to be reworked.

For now I think it can just accept a single access permission and later we can accept a list type optionally too for N required permissions. If we are worried about adding that later changing the type of the openAPI schema from string to list, we could make it a list now, or opt to never support N.

PR #2166, opened 2022-01-26
Make customizing/disabling queryset scoping easier

gerrod3:
fixes: #2114
fixes: #2115

@bmbouter @mdellweg @newswangerd I would like some feedback on the naming of the new `AccessPolicy` field (`filtering_permissions`) and the new permission class interface `objects_for_users`. Also, feedback on the documentation of these new additions would be helpful.


pulpcore/app/models/access_policy.py
```diff
@@ -34,13 +34,16 @@ class AccessPolicy(BaseModel):
...
+        filtering_permissions (models.JSONField): A default list of permissions used to filter the
+            queryset.
 
     """
 
     creation_hooks = models.JSONField(null=True)
     statements = models.JSONField()
     viewset_name = models.CharField(max_length=128, unique=True)
     customized = models.BooleanField(default=False)
+    filtering_permissions = models.JSONField(null=True)
```
mdellweg, line 46: We should make this a ListField.
Or do you have plans to use this field to define arbitrarily complicated filtering conditions?

pulpcore/app/access_policy.py
```diff
@@ -3,13 +3,22 @@
...
     """
     An AccessPolicy that loads statements from an `AccessPolicy` model instance.
     """
 
+    def get_access_policy(self, view):
+        """Retrieves the AccessPolicy from the DB or None if it doesn't exist."""
+        try:
+            viewset_name = get_view_urlpattern(view)
+            return AccessPolicyModel.objects.get(viewset_name=viewset_name)
+        except (AccessPolicyModel.DoesNotExist, AttributeError, ProgrammingError):
```
mdellweg, line 19: ProgrammingError? When do we rightfully expect that?

mdellweg (commented):
I am not opposed to the name `filtering_permissions` as it is used now.
My concern is that it only allows for the easy cases where a set of permissions is the discriminator for what you can see. For looking at content where we shy away from assigning roles per object, this is not so easy (consider filtering all content present in repositories you can see).

gerrod3:
> I am not opposed to the name `filtering_permissions` as it is used now. My concern is that it only allows for the easy cases where a set of permissions is the discriminator for what you can see. For looking at content where we shy away from assigning roles per object, this is not so easy (consider filtering all content present in repositories you can see).

What would that query even look like? Off the top of my head I would assume you would do this:

```python
repos = FileRepository.objects.all()
repos = get_objects_for_users(user, "file.view_filerepository", repos)
content = FileContent.objects.filter(repositories__in=repos)
```

Can this query be simplified? Also, how would we allow for more complex cases, what would the data structure for `filtering_permissions` be?

mdellweg:
> What would that query even look like? Off the top of my head I would assume you would do this:

We don't even know what we want from that query:
https://discourse.pulpproject.org/t/multi-tennancy-and-the-question-of-owning-content/237

I can only show you how it is handled in pulp_container today:
https://github.com/pulp/pulp_container/blob/main/pulp_container/app/viewsets.py#L176

pulpcore/app/viewsets/base.py
```diff
@@ -349,10 +347,15 @@ def get_queryset(self):
                 filters[lookup] = self.kwargs[key]
             qs = qs.filter(**filters)
 
+        # Deprecated, remove in 3.20
```
mdellweg, line 350: Is it obvious to future you that exactly three lines are supposed to be removed? I would mark the end of the block too.

pulpcore/app/models/access_policy.py
bmbouter: +1 to this

pulpcore/app/access_policy.py
```diff
@@ -35,12 +44,30 @@ def get_policy_statements(self, request, view):
...
+        Gets the default filtering permissions used for queryset scoping.
+        """
+        if access_policy_obj := self.get_access_policy(view):
+            return access_policy_obj.get("filtering_permissions", [])
         else:
-            return access_policy_obj.statements
+            policy = getattr(view, "DEFAULT_ACCESS_POLICY", {})
+            return policy.get("filtering_permissions", [])
+
+    def objects_for_user(self, view, user, qs):
```
mdellweg, line 64: Have you seen this? https://rsinger86.github.io/drf-access-policy/multi_tenacy.html

Maybe we should strive for this to look similar. OTOH, this may be misleading to our customization. WDYT?

docs/plugins/plugin-writer/concepts/rbac/queryset_scoping.rst
```diff
@@ -44,6 +61,9 @@ To look up objects by permission easily from an existing QuerySet use the ``get_
 provided by pulpcore or django-guardian. Here's an example where all items are displayed accessible
 via either of the permission frameworks:
 
+NOTE: Doing this will prevent the user from disabling queryset scoping.
+# I think this section should be removed from the docs
```
mdellweg, line 65: Yes, this is the way we don't want it anymore.
By removed, you mean replaced?

pulpcore/app/viewsets/task.py
```diff
@@ -120,6 +119,17 @@ class TaskViewSet(
...
+        "scoping_hooks": [
+            {
+                "function": "scope_required_permissions",
+                "parameters": {"permissions": "core.view_task"},
+            },
+            {
+                # This will default "or" with previous hook
+                "function": "scope_related_attribute",
+                "parameters": {"attribute": "task_group", "permissions": "core.view_task_group"},
+            },
```
mdellweg, line 131: Nice that you picked up that example. Do you intend to remove it again?
I think if this is meant to stay (and that may be a good idea), we should adjust the statements above to the same logic.

pulpcore/app/access_policy.py
```diff
@@ -35,12 +44,43 @@ def get_policy_statements(self, request, view):
...
+        Runs the functions in the `scoping_hooks` field of the access policy.
+        """
+        view = get_viewset_for_model(qs.model)
+        # kind of a hack, either this here, or the inverse in get_queryset
+        setattr(view, "request", request)
+        hooks = cls.get_scoping_hooks(view)
+        final_qs = qs.none()
+        for hook in hooks:
+            # Should I raise an error if hook is wrong?
+            if function := getattr(view, hook["function"], None):
```
mdellweg, line 79: I know, this will add some burden on all parties, but I'd rather see the scoping hooks need to be registered like the creation hooks. We should not allow the user to access all attributes on a class.

pulpcore/app/access_policy.py
```diff
@@ -35,12 +44,43 @@ def get_policy_statements(self, request, view):
...
+                # is this consistent with creation_hooks?
+                scoped_qs = function(view, qs, **hook["parameters"])
+                if hook.get("operation", "or") == "or":
+                    final_qs |= scoped_qs
+                else:
+                    final_qs &= scoped_qs
```
mdellweg, line 85: Hmm, left-associative evaluation of boolean and and or may be quite unintuitive. Also it does not cover all cases. `a xor b` cannot be expressed that way. DNF, CNF would be with an added level of complexity.
Why don't we just assume it's always "or". I'd say if there is one permission to see something, there cannot be another permission to hide it again (it would need to be called "denial" instead).

pulpcore/app/viewsets/task.py
```diff
@@ -120,6 +119,17 @@ class TaskViewSet(
                 "parameters": {"roles": "core.task_owner"},
             }
         ],
+        "scoping_hooks": [
```
bmbouter, line 122: I think this is net-new functionality for this endpoint. If you agree, can you get a file a feature issue for this and add a changelog.

gerrod3:
Closing in favor of #2212.

PR #2167, opened 2022-01-27
Return a concise message exception on 500 in case file is missing on the FS.

ipanova:
[noissue]

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -772,7 +772,10 @@ class Handler:
         artifact_name = artifact_file.name
 
         if settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem":
-            return FileResponse(os.path.join(settings.MEDIA_ROOT, artifact_name), headers=headers)
+            path = os.path.join(settings.MEDIA_ROOT, artifact_name)
+            if not os.path.exists(path):
+                raise Exception(_("Expected path '{}' is not found").format(path))
+            return FileResponse(path, headers=headers)
         elif settings.DEFAULT_FILE_STORAGE == "storages.backends.s3boto3.S3Boto3Storage":
             content_disposition = f"attachment;filename={content_artifact.relative_path}"
             parameters = {"ResponseContentDisposition": content_disposition}
--- a/CHANGES/2187.bugfix
+++ b/CHANGES/2187.bugfix
@@ -1,0 +1,1 @@
+Return a concise message exception on 500 in case file is missing on the FS.


pulpcore/content/handler.py
```diff
@@ -772,7 +772,10 @@ async def _serve_content_artifact(self, content_artifact, headers, request):
         artifact_name = artifact_file.name
 
         if settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem":
-            return FileResponse(os.path.join(settings.MEDIA_ROOT, artifact_name), headers=headers)
+            path = os.path.join(settings.MEDIA_ROOT, artifact_name)
+            if not os.path.exists(path):
```
gerrod3, line 776: So this path won't exist if for some reason the user removes the artifact files from their system? Would this also occur on reclaim?

ipanova: it won't occur on reclaim, because reclaimed content does not have artifact, artifact is removed. so, for example here it would pass onto streaming the CA in case of RA available https://github.com/pulp/pulpcore/blob/4c78830b8199aeae35807f00ff729d94fa247c54/pulpcore/content/handler.py#L586

ipanova: here is some discussion in pulp-container plugin on whether keep returning unhandled exception 500 or turn it into 4xx https://github.com/pulp/pulp_container/pull/556

mdellweg: However, i think repair would try to redownload the missing file for an existing artifact.

bmbouter:
I think this bugfix would benefit from an issue filed and changelog describing the problem and expected resolution.

pulpcore/content/handler.py
```diff
@@ -772,7 +772,11 @@ async def _serve_content_artifact(self, content_artifact, headers, request):
...
+                log.warning(f"Expected path {path} is not found")
```
bmbouter, line 777: Can this be i8ln with gettext please?

bmbouter (approved):

pulpcore/content/handler.py
```diff
@@ -772,7 +772,11 @@ async def _serve_content_artifact(self, content_artifact, headers, request):
         artifact_name = artifact_file.name
 
         if settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem":
-            return FileResponse(os.path.join(settings.MEDIA_ROOT, artifact_name), headers=headers)
+            path = os.path.join(settings.MEDIA_ROOT, artifact_name)
+            if not os.path.exists(path):
+                log.warning(_("Expected path {} is not found").format(path))
```
bmbouter, line 777:
```suggestion
                log.warning(_("Expected path '{}' is not found").format(path))
```

dralley: +1

pulpcore/content/handler.py
dralley: Correct, if there's an Artifact in the database but the file is missing, repair will find it, and if there are remote artifacts present, they'll be redownloaded.

pulpcore/content/handler.py
```diff
@@ -772,7 +772,10 @@ async def _serve_content_artifact(self, content_artifact, headers, request):
         artifact_name = artifact_file.name
 
         if settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem":
-            return FileResponse(os.path.join(settings.MEDIA_ROOT, artifact_name), headers=headers)
+            path = os.path.join(settings.MEDIA_ROOT, artifact_name)
+            if not os.path.exists(path):
+                raise Exception(_("Expected path '{}' is not found").format(path))
```
mdellweg, line 777: See comment https://github.com/pulp/pulp_container/pull/556/files#r811801174

mdellweg (approved):

dralley (approved):

PR #2174, opened 2022-01-28
Adds various proxy.py tests

### LINKED ISSUE #2173: Add proxy tests
Pulpcore needs a variety of tests for its proxy support. These can only cover pulp_file because that's all pulpcore has to test with.

bmbouter:
closes #2173

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/test_proxy.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_proxy.py
@@ -1,0 +1,133 @@
+import pytest
+from pulp_smash.pulp3.bindings import monitor_task, PulpTaskError
+
+from pulpcore.client.pulp_file import (
+    RepositorySyncURL,
+)
+
+
+def _run_basic_sync_and_assert(remote, file_repo, file_repo_api_client, content_file_api_client):
+    body = RepositorySyncURL(remote=remote.pulp_href)
+    monitor_task(file_repo_api_client.sync(file_repo.pulp_href, body).task)
+
+    # Check content is present, but no artifacts are there
+    content_response = content_file_api_client.list(
+        repository_version=f"{file_repo.versions_href}1/"
+    )
+    assert content_response.count == 3
+    for content in content_response.results:
+        assert content.artifact is None
+
+
+@pytest.mark.parallel
+def test_sync_http_through_http_proxy(
+    file_fixture_gen_remote,
+    file_repo,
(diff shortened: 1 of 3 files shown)


pulpcore/tests/functional/api/using_plugin/test_proxy.py
```diff
@@ -0,0 +1,111 @@
...
+)
+
+
+def _run_basic_sync_and_assert(remote, file_repo, file_repo_api_client, content_file_api_client):
+    body = RepositorySyncURL(remote=remote.pulp_href)
+    monitor_task(file_repo_api_client.sync(file_repo.pulp_href, body).task)
+
+    # Check content is present, but no artifacts are there
+    content_response = content_file_api_client.list()
+    assert content_response.count == 3
```
mdellweg, line 15: This kind of tests requires you to start on an empty pulp. In Development this can be cumbersome.

mdellweg: Also it's an inhibitor for parallel testing.

bmbouter: I was also thinking about this. I did remove the orphan_pre fixture from the tests themselves, so these tests won't pre-cleanup any content. I believe it's fine if the content is in Pulp or not because the fetching of the manifest file proves the correct usage via the proxy.

I concluded it's safe to mark as parallel for two reasons actually. 1) Orphan cleanup occurring concurrently will not remove the content this test also needs because of the `touch()` feature. 2) Currently the entire set of parallel tasks perform no orphan cleanup so the set of parallel tests won't trigger parallel orphan cleanup operations anyway.

My goal though is to improve this a lot more by having data fixture generated at runtime which would I think totally remove all of these concerns architecturally. That will be separate from this change though.

mdellweg: parallel tests must never orphan cleanup, because other tests may have some artifacts/content in flight not currently associated to a repository.

bmbouter: Actually I didn't realize my API call was not for content in that specific version. I've updated it to do that correctly now, and I also fixed it in another place too.

mdellweg: yes. exactly

fao89 (approved):
LGTM

PR #2177, opened 2022-01-31
Fixes #2168 - Setting for flexible imports

### LINKED ISSUE #2168: RFE - ignore repositories on Import
**Topic**
Import/Export

**Problem**
Katello needs the ability to restrict repositories that can be imported. Idea here is that the export archive can contain repositories that need special subscriptions that the user may not have. Ideally instead of failing on import katello should be able to tell pulp "Please only import these repositories". In a disconnected environment this can be particularly painful to fix.

**Suggestion**

https://docs.pulpproject.org/pulpcore/restapi.html#operation/importers_core_pulp_create 
When we create an Importer, Katello provides a repository mapping. Given this Pulp should

-  Either add an additional `repo_mapping_only` or `restricted` parameter in the importer. If this is true pulp will only import things that are provided in the `repo_mapping` and ignore the rest. This could be an application wide setting also (alternatively)
- OR Provide an `exclude_repositories` or `include_repositories` that takes a list of repository ids and ignores mapping those.

parthaa:
Customers want the flexibility in their import archive (since they tend
to be in disconnected environments.) This patch introduces a
IMPORT_MAPPED_REPOS_ONLY application wide setting that allows pulp to
turn on the "leniency" mode and import only the repositories specified
in the importer repository mapping and ignore the rest. The default
behavior remains the same (IMPORT_MAPPED_REPOS_ONLY is false by
default)

Fixes #2168


pulpcore/app/tasks/importer.py
```diff
@@ -180,6 +182,8 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
             # use the content mapping to map content to repos
             for repo_name, content_ids in mapping.items():
                 repo = _destination_repo(importer, repo_name)
+                if not repo:
```
ggainey, line 185: This will let us avoid creating any non-mapped repository - but all the Artifacts will be pulled in regardless. That means you'll likely end up with a lot of orphaned-content. I don't know if that's a problem for your use-case, but it's def something to think about explicitly.

ggainey (changes requested):
@parthaa - Generally speaking, looks like a reasonable approach. See my comment RE artifacts tho.

Please rebase against latest-main to make the linter happy. I'm not sure what's going on with the single-commit-test - I only see one, Maybe once you rebase/force-push it'll resolve the problem.

pulpcore/app/tasks/importer.py
parthaa: Thats a good point. I wonder if we can tie  this config setting to artifact import also or is it out of our control?

ggainey: > Thats a good point. I wonder if we can tie this config setting to artifact import also or is it out of our control?

No - because artifacts cross repo-boundaries, imp/exp doesn't know which repos in the export required an artifact, just "here is the set of all artifacts requested by all repositories in the export". When you go to import, even if you only want one repo on the destination machine, you're going to get ALL the Artifacts, until orphan-cleanup deletes the ones you don't want. PulpImport/Export assumes tailored exports.

I understand the use-case driving this, but I'm not convinced it makes more sense than restricting at export-time...

ggainey: The import-flow requires/assumes that all artifacts are imported as one task, and once that completes, repos are imported as parallel tasks. You can't import a repo unless/until its artifacts exist. And you can't know which artifacts belong to what, until you're "inside" the repo, and can look at its ContentArtifact records. I don't think there is a way to do what you want, without basically rewriting import.

parthaa: Oh I see. So then I 'd think as long as we document that you need at least  2 times the space of the import archive (even if you are only using 50% of it and have to do orphan cleanup.) we should be okay?. If the user cares about space they should be creating a separate content view and managing it.

ggainey: > Oh I see. So then I 'd think as long as we document that you need at least 2 times the space of the import archive (even if you are only using 50% of it and have to do orphan cleanup.) we should be okay?. If the user cares about space they should be creating a separate content view and managing it.

That sounds like an eminently-reasonable approach :)

parthaa: I asked the TAM for the interested customer if it is a reasonable thing. Will update the card once I hear back one way or the other.

pulpcore/app/settings.py
```diff
@@ -223,6 +223,8 @@
 
 REMOTE_USER_ENVIRON_NAME = "REMOTE_USER"
 
+IMPORT_MAPPED_REPOS_ONLY = False
+
```
mdellweg, line 227: I'm unsure, this should be a global option. Have you considered making this a import parameter (in the POST payload)?

ggainey: If we do thjis, it should be global - it's the administrator saying "I am OK with doing partial imports". But this whole RFE has some insurmountable problems, suggest we leave it unless/until we get clarification from stakeholder.

bmbouter: I don't have an opinion on adding a setting or not, but if we do, it needs docs https://docs.pulpproject.org/pulpcore/configuration/settings.html#pulp-settings

bmbouter:
I'm reading here this PR is still in progress so I've converted it to a draft until it's ready for review+merge.

ggainey:
@parthaa I know there's a lot of discussion going on here - given the Way Things Work in pulp, I'd like to propose we just close this PR unless/until we get clarity. Opinion?

ggainey:
Closing until someone yells at us

PR #2184, opened 2022-02-01
Fix canceling tasks never being picked up

### LINKED ISSUE #2183: Canceling tasks never picked up by a worker marks them failed
Code investigation shows, we are doing the right thing for the right reason, but reported the wrong thing with the wrong reason.

mdellweg:
Previously, tasks that were canceled without being assigned to a worker
would appear as failed due to worker gone missing.

fixes #2183

### DIFF
--- a/pulpcore/tasking/pulpcore_worker.py
+++ b/pulpcore/tasking/pulpcore_worker.py
@@ -194,6 +194,13 @@ class NewPulpWorker:
                     # The lock will be automatically be released at the end of the block
                     # Check if someone else changed the task before we got the lock
                     task.refresh_from_db()
+                    if task.state == TASK_STATES.CANCELING and task.worker is None:
+                        # No worker picked this task up before being canceled
+                        if self.cancel_abandoned_task(task, TASK_STATES.CANCELED):
+                            # Continue looking for the next task
+                            # without considering this tasks resources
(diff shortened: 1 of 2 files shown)


pulpcore/tasking/pulpcore_worker.py
```diff
@@ -194,6 +194,13 @@ def iter_tasks(self):
                     # The lock will be automatically be released at the end of the block
                     # Check if someone else changed the task before we got the lock
                     task.refresh_from_db()
+                    if task.state in TASK_STATES.CANCELING and task.worker is None:
```
dralley, line 197: `task.state in TASK_STATES.CANCELING` probably works "accidentally" :) I think `task.state == TASK_STATES.CANCELING` would more clearly express the intent.

dralley:
LGTM apart from the suggestion

pulpcore/tasking/pulpcore_worker.py
mdellweg: Right! Thanks for catching that.

mdellweg: Fixed.

ggainey (approved):
Nice - lg2m!

dralley (approved):

Issue #2185, opened 2022-02-01
Circular import exists when using custom user model


bmclaughlin:
This currently occurs when a plugin using a custom user model depends on the Role model [here](https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/__init__.py#L61).

PR #2186, opened 2022-02-01
Fixes circular import for plugins with custom user model

bmclaughlin:
fixes: [pulp#2185](https://github.com/pulp/pulpcore/issues/2185)

### DIFF
--- a/CHANGES/plugin_api/2185.bugfix
+++ b/CHANGES/plugin_api/2185.bugfix
@@ -1,0 +1,1 @@
+Expose Roles model in plugin api.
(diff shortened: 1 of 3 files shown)


mdellweg:
Perfect. Can you add a not to the original commented import in pulpcore.plugin.models where these special models can be found? (I think that would be super helpful.)

mdellweg:
Sorry, didn't see this earlier. The syntax is:
`fixes: #2185`

mdellweg (approved):

ipanova (approved):

PR #2198, opened 2022-02-02
Adds the `handle_creation_hooks` interface

### LINKED ISSUE #2116: There is no way for users to prevent creation_hooks from running without editing every AccessPolicy to remove them
## Problem

Users want to fully disable the default Authorization of pulp by modifying the DEFAULT_PERMISSION_CLASSES, but even if they do the creation hooks will still be running, likely assigned unwanted permissions to various users and groups. 

## Solution

Have the calling of creation_hooks happen in side an interface on `AccessPolicyFromDB`, e.g. named `handle_creation_hooks` or some reasonable name. By delegating the running of creation hooks onto that object, when the user changes it, they won't run anymore.

bmbouter:
This change adds the `handle_creation_hooks` interface to the
``AccessPolicyFromDB`` method. If the AccessPolicy configured does not
have that interface, new model instances created will not have
permissions automatically added to them.

closes #2116

### DIFF
--- a/CHANGES/2116.feature
+++ b/CHANGES/2116.feature
@@ -1,0 +1,2 @@
+Specifying a different value for ``DEFAULT_PERMISSION_CLASSES`` will now automatically disable the
+permission assignment provided by the ``creation_hooks`` portion of an ``AccessPolicyFromDB``.
--- a/CHANGES/plugin_api/2116.feature
+++ b/CHANGES/plugin_api/2116.feature
@@ -1,0 +1,2 @@
+The ``AutoAddObjPermsMixin`` now calls a ``handle_creation_hooks`` interface on the configured DRF
+permission class, e.g. the default ``AccessPolicyFromDB``.
--- a/pulpcore/app/access_policy.py
+++ b/pulpcore/app/access_policy.py
@@ -1,14 +1,65 @@
 from rest_access_policy import AccessPolicy
-from django.db.utils import ProgrammingError
 
+from pulpcore.app.loggers import deprecation_logger
 from pulpcore.app.models import AccessPolicy as AccessPolicyModel
-from pulpcore.app.util import get_view_urlpattern
+from pulpcore.app.util import get_view_urlpattern, get_viewset_for_model
 
 
 class AccessPolicyFromDB(AccessPolicy):
     """
     An AccessPolicy that loads statements from an `AccessPolicy` model instance.
     """
+
+    @staticmethod
+    def get_access_policy(view):
+        """
+        Retrieves the AccessPolicy from the DB or None if it doesn't exist.
+
+        Args:
+            view (subclass of rest_framework.view.APIView): The view or viewset to receive the
+                AccessPolicy model for.
+
+        Returns:
+            Either a `pulpcore.app.models.AccessPolicy` or None.
+        """
+        try:
+            urlpattern = get_view_urlpattern(view)
+        except AttributeError:
+            # The view does not define a `urlpattern()` method, e.g. it's not a NamedModelViewset
+            return None
+
+        try:
+            return AccessPolicyModel.objects.get(viewset_name=urlpattern)
+        except AccessPolicyModel.DoesNotExist:
+            return None
+
+    @classmethod
+    def handle_creation_hooks(cls, obj):
+        """
+        Handle the creation hooks defined in this policy for the passed in `obj`.
+
+        Args:
+            cls: The class this method belongs to.
+            obj: The model instance to have its creation hooks handled for.
+
+        """
+        viewset = get_viewset_for_model(obj)
+        access_policy = cls.get_access_policy(viewset)
+        if access_policy and access_policy.creation_hooks is not None:
+            for creation_hook in access_policy.creation_hooks:
+                function = obj.REGISTERED_CREATION_HOOKS.get(creation_hook["function"])
+                if function is not None:
+                    kwargs = creation_hook.get("parameters") or {}
+                    function(**kwargs)
+                else:
+                    # Old interface deprecated for removal in 3.20
+                    function = getattr(obj, creation_hook["function"])
+                    deprecation_logger.warn(
+                        "Calling unregistered creation hooks from the access policy is deprecated"
+                        " and may be removed with pulpcore 3.20."
+                        f"[hook={creation_hook}, viewset={access_policy.viewset_name}]."
+                    )
+                    function(creation_hook.get("permissions"), creation_hook.get("parameters"))
 
     def get_policy_statements(self, request, view):
         """
@@ -35,12 +86,9 @@ class AccessPolicyFromDB(AccessPolicy):
         Returns:
             The access policy statements in drf-access-policy policy structure.
         """
-        try:
-            viewset_name = get_view_urlpattern(view)
-            access_policy_obj = AccessPolicyModel.objects.get(viewset_name=viewset_name)
-        except (AccessPolicyModel.DoesNotExist, AttributeError, ProgrammingError):
+        if access_policy_obj := self.get_access_policy(view):
+            return access_policy_obj.statements
+        else:
             default_statement = [{"action": "*", "principal": "admin", "effect": "allow"}]
             policy = getattr(view, "DEFAULT_ACCESS_POLICY", {"statements": default_statement})
             return policy["statements"]
-        else:
-            return access_policy_obj.statements
--- a/pulpcore/app/models/access_policy.py
+++ b/pulpcore/app/models/access_policy.py
@@ -1,6 +1,3 @@
-from gettext import gettext as _
-
-from django.core.exceptions import ImproperlyConfigured
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group as BaseGroup
 from django.db import models
@@ -10,7 +7,7 @@ from guardian.models.models import GroupObjectPermission, UserObjectPermission
 from guardian.shortcuts import assign_perm, remove_perm
 
 from pulpcore.app.models import BaseModel
-from pulpcore.app.loggers import deprecation_logger
+from pulpcore.app.util import get_viewset_for_model
 
 
 def _ensure_iterable(obj):
@@ -76,33 +73,10 @@ class AutoAddObjPermsMixin:
 
     @hook("after_create")
     def add_perms(self):
-        try:
-            access_policy = AccessPolicy.objects.get(viewset_name=self.ACCESS_POLICY_VIEWSET_NAME)
-        except AttributeError:
-            raise ImproperlyConfigured(
-                _(
-                    "When using the `AutoAddObjPermsMixin`, plugin writers must declare an"
-                    "`ACCESS_POLICY_VIEWSET_NAME` class attribute."
-                )
-            )
-        self._handle_creation_hooks(access_policy)
-
-    def _handle_creation_hooks(self, access_policy):
-        if access_policy.creation_hooks is not None:
-            for creation_hook in access_policy.creation_hooks:
-                function = self.REGISTERED_CREATION_HOOKS.get(creation_hook["function"])
-                if function is not None:
-                    kwargs = creation_hook.get("parameters") or {}
-                    function(**kwargs)
-                else:
-                    # Old interface deprecated for removal in 3.20
-                    function = getattr(self, creation_hook["function"])
-                    deprecation_logger.warn(
-                        "Calling unregistered creation hooks from the access policy is deprecated"
-                        " and may be removed with pulpcore 3.20 "
-                        f"[hook={creation_hook}, viewset={access_policy.viewset_name}]."
-                    )
-                    function(creation_hook.get("permissions"), creation_hook.get("parameters"))
+        viewset = get_viewset_for_model(self)
+        for permission_class in viewset.get_permissions(viewset):
+            if hasattr(permission_class, "handle_creation_hooks"):
+                permission_class.handle_creation_hooks(self)
 
     def add_for_users(self, permissions, users):
         """


bmbouter:
I posted this and I think it does what is asked of the immediate issue, but I realize it likely should do much more? Here are two main lines of questions I have:

1) The things added here to `AccessPolicyFromDB` should like be on `AccessPolicy` directly. Or maybe those objects should be combined? We should think about why we even have two.

2) The `REGISTERED_CREATION_HOOKS` should likely be decoupled from `AutoAddObjPermsMixin` entirely as they don't make sense unless you have an `AccessPolicy` configured that has a notion of `creation_hooks`.

@mdellweg and @gerrod3, let's discuss ^ at the RTFM working group sometime. @newswangerd FYI also.

CHANGES/plugin_api/2116.feature
```diff
@@ -0,0 +1,2 @@
+The ``AccessPolicyFromDB`` now has the ``handle_creation_hooks`` interface which takes the model
+to add object-level permissions to.
```
mdellweg, line 2: Not sure this needs to called out for the plugin writer. It's the same hooks specified in the same way. Nothing changes for plugins. We just introduce one additional abstraction to make the actual calls dependent on a condition the plugin writer has no control over.

mdellweg:
> I posted this and I think it does what is asked of the immediate issue, but I realize it likely should do much more? Here are two main lines of questions I have:
> 
>     1. The things added here to `AccessPolicyFromDB` should like be on `AccessPolicy` directly. Or maybe those objects should be combined? We should think about why we even have two.

`AccessPolicy` sadly is a bit ambiguous. Let's assume you mean the model, then no. One of them is a model and the other one is a permission class. Completely separate hierarchies.

> 
>     2. The `REGISTERED_CREATION_HOOKS` should likely be decoupled from `AutoAddObjPermsMixin` entirely as they don't make sense unless you have an `AccessPolicy` configured that has a notion of `creation_hooks`.

I don't see a reason. The whole purpose of the Mixin is to support calling creation_hooks. So this seems like the right domain for that concern. Also the specific model must register their hooks in this struct. So it must end in that class.

pulpcore/app/models/access_policy.py
```diff
@@ -85,24 +84,10 @@ def add_perms(self):
...
-                    # Old interface deprecated for removal in 3.20
-                    function = getattr(self, creation_hook["function"])
-                    deprecation_logger.warn(
-                        "Calling unregistered creation hooks from the access policy is deprecated"
-                        " and may be removed with pulpcore 3.20 "
-                        f"[hook={creation_hook}, viewset={access_policy.viewset_name}]."
-                    )
-                    function(creation_hook.get("permissions"), creation_hook.get("parameters"))
+        try:
+            access_policy.handle_creation_hooks(self)
```
mdellweg, line 88: This is not doing the logic. I think, you need to find the ViewSet class and loop over all configured permission classes and try to call the interface on those.
Remember the access policy is just a glorified permission class we want to allow to be configured.

gerrod3:
>     2. The `REGISTERED_CREATION_HOOKS` should likely be decoupled from `AutoAddObjPermsMixin` entirely as they don't make sense unless you have an `AccessPolicy` configured that has a notion of `creation_hooks`.

I remember asking @mdellweg if `REGISTERED_CREATION_HOOKS` could live on the viewset (or something similar) and just have `AutoAddObjPermsMixin` add to the value, but I think he said it wasn't possible. Maybe it should be revisted because I think it would be hard to add other creation_hooks if you are using that mixin.

mdellweg:
> > ```
> > 2. The `REGISTERED_CREATION_HOOKS` should likely be decoupled from `AutoAddObjPermsMixin` entirely as they don't make sense unless you have an `AccessPolicy` configured that has a notion of `creation_hooks`.
> > ```
> 
> I remember asking @mdellweg if `REGISTERED_CREATION_HOOKS` could live on the viewset (or something similar) and just have `AutoAddObjPermsMixin` add to the value, but I think he said it wasn't possible. Maybe it should be revisted because I think it would be hard to add other creation_hooks if you are using that mixin.

I think, the core of the confusion is, that we define rules for views, but actually we want to allow access to models. So one question is, how tightly coupled Model an Viewset are. The other is how much of the logic we want to delegate to the Model. And maybe this goes very far. Permissions and Roles are already with the Model and the view always knows about the attached model. We go through a lot of complicated calls to find the hooks on the view when we only have the model (post creation hooks...).

pulpcore/app/models/access_policy.py
bmbouter: I got this implemented, but I haven't pushed it yet. What I need to understand though is after calling `AccessPolicyInDB.handle_creation_hooks` does it call back to the model and/or viewset? I *think* the idea is to have it call back to the model just like this PR does, but I'm a little unsure.

mdellweg: It does call back, because the hooks themselves are on the viewset (or was it the model?).
Whether the hooks are enumerated by a method in the access policy or by one in the viewset, i don't care.

CHANGES/plugin_api/2116.feature
bmbouter: I think there possibly is something here that is new. I've reworded it; lmk what you think please.

CHANGES/2116.feature
```diff
@@ -0,0 +1,2 @@
+Specifying a different value for ``DEFAULT_PERMISSION_CLASSES`` will now automatically disable the
+permission assignment provided by the ``creation_hooks`` portion of an ``AccessPolicyFromDB``.
```
mdellweg, line 2: I am starting to wonder, what happens if you define 2 permission classes with the creation hooks interface and both hook back into the models creation hooks. Should we consider that a configuration error and document a warning?

pulpcore/app/models/access_policy.py
```diff
@@ -85,24 +84,10 @@ def add_perms(self):
...
-                else:
-                    # Old interface deprecated for removal in 3.20
-                    function = getattr(self, creation_hook["function"])
-                    deprecation_logger.warn(
-                        "Calling unregistered creation hooks from the access policy is deprecated"
-                        " and may be removed with pulpcore 3.20 "
-                        f"[hook={creation_hook}, viewset={access_policy.viewset_name}]."
-                    )
-                    function(creation_hook.get("permissions"), creation_hook.get("parameters"))
+        for viewset in access_policy.named_viewsets:
```
mdellweg, line 87: Why are we looking at the viewsets for the access_policy?

mdellweg (changes requested):

pulpcore/app/models/access_policy.py
mdellweg:
```suggestion
        for viewset in self.named_viewsets:
```

mdellweg: Maybe we should not loop but use the one viewset described by ACCESS_POLICY_VIEWSET_NAME

CHANGES/2116.feature
bmbouter: Mainly for the sake of time, I'd prefer to wait until we run into this problem to add warnings for that case.

bmbouter:
@mdellweg @gerrod3 @newswangerd here's my revised push, please review.

pulpcore/app/models/base.py
```diff
@@ -78,6 +79,12 @@ def __str__(self):
     def __repr__(self):
         return str(self)
 
+    @property
+    def named_viewsets(self):
+        """Returns a list of named viewsets corresponding to this model."""
+        named_viewsets = apps.get_app_config("core").named_viewsets
+        return named_viewsets[type(self)]
+
```
gerrod3, line 87: Is this still needed?

pulpcore/app/models/base.py
```diff
@@ -78,6 +79,12 @@ def __str__(self):
     def __repr__(self):
         return str(self)
 
+    @property
+    def named_viewsets(self):
```
newswangerd, line 83: Does this need to be here? I don't see it used anywhere else

pulpcore/app/access_policy.py
```diff
@@ -10,6 +11,32 @@ class AccessPolicyFromDB(AccessPolicy):
     An AccessPolicy that loads statements from an `AccessPolicy` model instance.
     """
 
+    @staticmethod
+    def handle_creation_hooks(obj, access_policy_model_obj):
```
newswangerd, line 15: I'm wondering if `access_policy_model_obj` is the right object to pass here. If a user isn't using `AccessPolicyFromDB`, then there's not necessarily a guarantee that an access policy object for the viewset will exist in the database.

I wonder if it would be better to pass the viewset and have this function perform the DB lookup for the object? I suppose there's no way to disable the access policy initialization, so maybe it doesn't matter.

pulpcore/app/models/base.py
bmbouter: It does not it was used before, but I can remove it now.

bmbouter: pushed

pulpcore/app/models/base.py
bmbouter: I removed it.

pulpcore/app/access_policy.py
mdellweg: I agree, the access policy already has the lookup policy statements for that. This way, we can make use of creation hooks in policies provided by any means. And this circles back nicely to the original request.

mdellweg: The other pr seems to be ahead of yours again in breaking down that function.  😆

pulpcore/app/access_policy.py
```diff
@@ -10,6 +14,42 @@ class AccessPolicyFromDB(AccessPolicy):
     An AccessPolicy that loads statements from an `AccessPolicy` model instance.
     """
 
+    @staticmethod
+    def handle_creation_hooks(obj):
```
bmbouter, line 18: @newswangerd @mdellweg is this what you meant? Note I didn't even include the viewset because I'm not sure why it's needed more generally.

pulpcore/app/access_policy.py
```diff
@@ -10,6 +14,42 @@ class AccessPolicyFromDB(AccessPolicy):
...
+            access_policy = AccessPolicyModel.objects.get(
+                viewset_name=obj.ACCESS_POLICY_VIEWSET_NAME
+            )
+        except AttributeError:
+            raise ImproperlyConfigured(
+                _(
+                    "When using the `AutoAddObjPermsMixin`, plugin writers must declare an"
+                    "`ACCESS_POLICY_VIEWSET_NAME` class attribute on models."
+                )
+            )
```
mdellweg, line 36: I'm not sure this warning is valuable anymore, when the whole idea of this change is based on the premise to be more flexible to different configurations.

pulpcore/app/access_policy.py
```diff
@@ -10,6 +14,42 @@ class AccessPolicyFromDB(AccessPolicy):
...
+        Handle the creation hooks defined in this policy for the passed in `obj`.
+
+        Args:
+            obj: The model instance to have its creation hooks handled for.
+
+        """
+        try:
+            access_policy = AccessPolicyModel.objects.get(
+                viewset_name=obj.ACCESS_POLICY_VIEWSET_NAME
+            )
```
mdellweg, line 29: Actually i was thinking about this transformation:
https://github.com/pulp/pulpcore/pull/2166/files#diff-f3d57ed180a61179704601b8a978567bf8e5c5c54e95458ed8b0b49bbc150f3cR14

The benefit is that a subclassed AccessPolicy can just overwrite their `get_access_policy` method and still get all the glory from `creation_hooks`.

pulpcore/app/access_policy.py
```diff
@@ -1,15 +1,62 @@
...
+from pulpcore.app.util import get_view_urlpattern, get_viewset_for_model
 
 
 class AccessPolicyFromDB(AccessPolicy):
     """
     An AccessPolicy that loads statements from an `AccessPolicy` model instance.
     """
 
+    @staticmethod
+    def get_access_policy(viewset):
```
mdellweg, line 15: Now the nice thing is by subclassing that method, you can completely reroute the whole access policy lookup.

mdellweg (approved):
Now this looks so nice to me!

newswangerd (approved):

gerrod3 (approved):

pulpcore/app/access_policy.py
```diff
@@ -1,15 +1,62 @@
...
+                    # Old interface deprecated for removal in 3.20
+                    function = getattr(obj, creation_hook["function"])
+                    deprecation_logger.warn(
+                        "Calling unregistered creation hooks from the access policy is deprecated"
+                        " and may be removed with pulpcore 3.20."
+                        f"[hook={creation_hook}, viewset={access_policy.viewset_name}]."
+                    )
+                    function(creation_hook.get("permissions"), creation_hook.get("parameters"))
+
     def get_policy_statements(self, request, view):
```
newswangerd, line 60: Shouldn't this be updated to use `cls.get_access_policy`?

newswangerd (changes requested):

pulpcore/app/access_policy.py
mdellweg: Oh right. Good catch!

mdellweg (changes requested):

pulpcore/app/access_policy.py
bmbouter: I updated this, please let me know what you think.

mdellweg (approved):

bmbouter:
In refactoring the usage to all use the new method to find the access policy, I had to revise this implementation a bit. I learned a few things.

1) The `AttributeError` being caught was raised by `get_view_urlpattern` and specifically in cases where the view/viewset doesn't define a `urlpattern()` method. Sound familiar @newswangerd ?

2) The OrphanCleanupView is one of those views and it would need a `urlpattern()` in order for us to RBAC it. After everything does that we probably could remove the AttributeError being caught here again.

mdellweg:
> In refactoring the usage to all use the new method to find the access policy, I had to revise this implementation a bit. I learned a few things.
> 
>     1. The `AttributeError` being caught was raised by `get_view_urlpattern` and specifically in cases where the view/viewset doesn't define a `urlpattern()` method. Sound familiar @newswangerd ?
> 
>     2. The OrphanCleanupView is one of those views and it would need a `urlpattern()` in order for us to RBAC it. After everything does that we probably could remove the AttributeError being caught here again.

I would keep catching it and just raise a refined version of the exception with an annotation of the programming error we already identified.

PR #2203, opened 2022-02-07
Use json fields for task arguments

mdellweg:
Open question:
- [x] Should we migrate the data on existing tasks?
- [x] Should we use DjangoJSONEncoder [0]?

[0] https://docs.djangoproject.com/en/3.2/topics/serialization/#djangojsonencoder


dralley:
Whoops, didn't mean to do that. The GitHub mobile app has buttons in inconvenient places..

bmbouter:
This is a nice PR. +1 to migrating the existing tasks otherwise we'll have that problem again at upgrade time where folks have unrun tasks which fail on the new style (if I'm understanding this right).

+1 to using DjangoJSONEncoder which supports the additional types nicely. It includes UUID which was the big reason we implemented ours IIRC.

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -356,6 +356,15 @@ def child_signal_handler(sig, frame):
         sys.exit()
 
 
+def _try_deserialize(data):
+    # On older tasks the arguments may be encoded and saved as a string
+    # This should be removed with 3.20
+    if isinstance(data, str):
+        return json.loads(data)
+    else:
+        return data
```
mdellweg, line 365: This tries to be nice to tasks scheduled before the switch. But with a migration it's probably cleaner.

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -356,6 +356,15 @@ def child_signal_handler(sig, frame):
         sys.exit()
 
 
+def _try_deserialize(data):
+    # On older tasks the arguments may be encoded and saved as a string
+    # This should be removed with 3.20
```
dralley, line 361: File an issue to track the removal https://github.com/pulp/pulpcore/milestone/1

mdellweg: I'll go for the migration so we don't need to introduce this.

pulpcore/app/migrations/0086_task_json_fields.py
```diff
@@ -0,0 +1,73 @@
...
+            update = False
+            if isinstance(task.args, str):
+                task.args = json.loads(task.args)
+                update = True
+            if isinstance(task.kwargs, str):
+                task.kwargs = json.loads(task.kwargs)
+                update = True
+            if update:
+                updated_tasks.append(task)
+        except ValueError:
```
dralley, line 22: When would we expect `ValueError` to be raised here?

pulpcore/app/migrations/0086_task_json_fields.py
```diff
@@ -0,0 +1,73 @@
...
+
+import json
+import django.core.serializers.json
+from django.db import migrations, models
+
+
+def convert_task_args_to_json(apps, schema_editor):
+    Task = apps.get_model("core", "Task")
+    updated_tasks = []
+    for task in Task.objects.all():
```
dralley, line 11: Use `.only("args", "kwargs").iterator()`  since there could be quite a lot of these tasks

pulpcore/app/migrations/0086_task_json_fields.py
mdellweg: Right i meant to check if json encoding errors are actually caught by this. I do not expect them ever. But with long lived databases, you never know.

dralley: Could you leave a comment to that effect?

mdellweg: Checked. JSONDecodeError qualifies as a ValueError. My goal is to make the migration fault tolerant.

pulpcore/app/migrations/0086_task_json_fields.py
```diff
@@ -0,0 +1,77 @@
...
+                update = True
+            if update:
+                updated_tasks.append(task)
+        except ValueError:
+            # Catch and ignore all cases of JSON transcoding errors.
+            # This is not even expected to be hit, but the migration should be fault tolerant.
+            pass
+        if len(updated_tasks) > 1000:
+            Task.objects.bulk_update(updated_tasks, ["args", "kwargs"])
+            updated_tasks.clear()
```
bmbouter, line 28: I think you could use the `batch_size` arg to `bulk_update` and it would do this for you https://docs.djangoproject.com/en/3.2/ref/models/querysets/#bulk-update

pulpcore/tasking/tasks.py
```diff
@@ -17,13 +15,6 @@
 TASK_TIMEOUT = -1  # -1 for infinite timeout
 
 
-class UUIDEncoder(json.JSONEncoder):
```
bmbouter, line 20: Awww yeahhhhh :+1:

pulpcore/app/migrations/0086_task_json_fields.py
dralley: @bmbouter It's not 100% equivalent because you'd still have the `args` and `kwargs` values of all the tasks sitting around in memory.

I doubt it would become a problem but just pointing out why you might want to use this approach vs. `batch_size`

dralley (approved):

pulpcore/app/migrations/0086_task_json_fields.py
bmbouter: Good point. I'm going to LGTM and @mdellweg can decide what he wants to do.

bmbouter (approved):

pulpcore/app/migrations/0086_task_json_fields.py
mdellweg: Yes, i adopted this pattern for the very reason that "the whole table (tm)" may not fit into memory.

PR #2206, opened 2022-02-08
Fix import and export OOM error

dkliban:
json.dumps method consumes a lot of memory when converting data
to json string. While processing data, it seems to create
python object which is a number of times larger than the
parsing data. For example, a 5GB data is resulting more than 20GB
consumed.

This commit fixes the issue by processing the dataset in batch
and write into a temporary file.

This commit also fixes the import OOM error by parsing only the
file object to the json.load method instead of reading whole
file into the memory.

closes #9645

(cherry picked from commit da8770892af16992d5d546238bcf5b6e5a466978)

### DIFF
--- a/CHANGES/2072.bugfix
+++ b/CHANGES/2072.bugfix
@@ -1,0 +1,1 @@
+Fix import and export OOM error.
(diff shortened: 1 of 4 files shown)


fao89:
I fixed here: https://github.com/pulp/pulpcore/pull/2205

dralley:
Merged via patchback

PR #2208, opened 2022-02-09
Scheduled Tasks PoC

mdellweg:


CHANGES/2204.feature
```diff
@@ -0,0 +1 @@
+Added ScheduledTask model that will periodically schedule configured tasks.
```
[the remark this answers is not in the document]
bmbouter: I think none of it at least initially. I think we should offer this to plugin writers in 3.19 at least.

bmbouter:
I'm glad to see this PoC. I can later rework the telmetry to use this when it's closer to merged or merged.

pulpcore/app/migrations/0087_scheduledtask.py
```diff
@@ -0,0 +1,32 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='ScheduledTask',
+            fields=[
+                ('pulp_id', models.UUIDField(default=uuid.uuid4, editable=False, primary_key=True, serialize=False)),
+                ('pulp_created', models.DateTimeField(auto_now_add=True)),
+                ('pulp_last_updated', models.DateTimeField(auto_now=True, null=True)),
+                ('last_dispatched', models.DateTimeField(auto_now=True)),
```
dralley, line 22: How is this stored?  We ought to use something monotonic and timezone independent.

pulpcore/app/models/task.py
```diff
@@ -334,3 +338,10 @@ class CreatedResource(GenericRelationModel):
     task = models.ForeignKey(
         Task, related_name="created_resources", default=Task.current, on_delete=models.CASCADE
     )
+
+
+class ScheduledTask(BaseModel):
+    last_dispatched = models.DateTimeField(auto_now=True)
+    dispatch_interval = models.DurationField()
+    task_name = models.CharField(max_length=256, unique=True)
```
bmbouter, line 346: I think adding `args` and `kwargs` would make sense based on the desire I read in the ticket to use this to dispatch some existing tasks that already exist in pulp and those take args, kwargs. I'm also read that [here too](https://github.com/pulp/pulpcore/pull/2118#discussion_r800793477).

bmbouter:
I think advisory locks are the easiest way to arbitrate correctness around the race to dispatch, so +1 on that. FWIW, it did seem easier to have that called from `pulpcore-content` heartbeat coroutine. This would also certainly work, but it ties this feature's implementation into an area that few people already understand. I don't think dispatching tasks and running tasks are actually very related. What are the things you think about in terms of that decision? Also imagine someone besides us has to fix this later, is this as simple as it could be?

pulpcore/app/migrations/0087_scheduledtask.py
```diff
@@ -0,0 +1,32 @@
...
+                ('dispatch_interval', models.DurationField()),
+                ('task_name', models.CharField(max_length=256, unique=True)),
```
dralley, line 24: Is this dependent on the implementation detail of the name and import path of the task function?  (admittedly the same has been true of the tasking system for a while, but that is at least invisible).

I also take this to mean that we want exactly one schedule per task?

pulpcore/tasking/tasks.py
```diff
@@ -76,13 +86,33 @@ def dispatch(
...
-    with connection.connection.cursor() as cursor:
-        cursor.execute("NOTIFY pulp_worker_wakeup")
+        transaction.on_commit(_wakeup_worker)
+
     return task
+
+
+def dispatch_scheduled_tasks():
+    # Warning, dispatch_scheduled_tasks is not race condition free!
+    now = timezone.now()
```
dralley, line 102: This would not be monotonic, how do we feel about that?

pulpcore/app/migrations/0087_scheduledtask.py
bmbouter: This is a good point. I think we may want several actually (from what I heard from @newswangerd ).

pulpcore/tasking/tasks.py
```diff
@@ -76,13 +86,33 @@ def dispatch(
...
+    # Dispatch all tasks old enough and not still running
+    for scheduled_task in ScheduledTask.objects.filter(
+        last_dispatched__lte=now - F("dispatch_interval")
```
dralley, line 105: Caveat: This means the dispatch schedule will drift over time, typically by seconds at a time but sometimes by large jumps.  If a dispatch is missed due to a maintainence window, it will dispatch immediately upon boot.  Something that was originally scheduled to happen nightly around midnight might start executing at 2:00 in the afternoon.

pulpcore/app/models/task.py
mdellweg: Args and kwargs are dangerous until we tell the task how to derive it's lock from them.

e.g. sync takes a repo_pk, but the user must specify the exclusive_resources for it separately.

pulpcore/tasking/tasks.py
mdellweg: Can we just tag this we don't care? We are not going to reimplement cron, crony, anacron, systemd here.
If you need atomic clock stable schedules, use one of these.

CHANGES/2204.feature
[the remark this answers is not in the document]
mdellweg: I disagree. We are about to tell everyone "use the api to identify issues". So this at the very least needs read access from the api.

pulpcore/app/migrations/0087_scheduledtask.py
mdellweg: With the help of last_task, we already guard that a task scheduled from the same schedule will not be kicked off again before the previous one finished.
OTOH, without args, kwargs there is not so much use for it.

mdellweg:
> I think advisory locks are the easiest way to arbitrate correctness around the race to dispatch, so +1 on that. FWIW, it did seem easier to have that called from `pulpcore-content` heartbeat coroutine. This would also certainly work, but it ties this feature's implementation into an area that few people already understand. I don't think dispatching tasks and running tasks are actually very related. What are the things you think about in terms of that decision? 

It carries "Task" in the name. So the first intuitive place to look at is the tasking system. And that first intuition to me is really solid. This is a concern about tasks. Let's keep the code handling tasks in one place. The content app has to cope with enough challenges around content, artifacts, external storage and more on it's own. Also the content handler code is to my knowledge not well understood by more people.
And to help you with a technical argument: `dispatch` is not safe to be used in async context. The pulpcore-worker is not (python-)async yet. And we are probably waiting for django to become fully async safe before we are going to change that.

> Also imagine someone besides us has to fix this later, is this as simple as it could be?

A great opportunity to learn and diversify the knowledge. I'd take that as an actually supporting argument.

pulpcore/tasking/tasks.py
mdellweg: But, i am going to revisit and save a "next_schedule" instead of "last_dispatch".

mdellweg:
> > > Also imagine someone besides us has to fix this later, is this as simple as it could be?
> > 
> > 
> > A great opportunity to learn and diversify the knowledge. I'd take that as an actually supporting argument.
> 
> I can see that perspective, but then let's figure that out now by finding two SMEs for this who aren't already tasking system SMEs. If we can find some folks then that will work for me. I'd like to hear other opinions to test my claim, but my claim here is that by integrating this feature with existing tasking system code, it's significantly harder to learn than having it live cleanly separately with a straightforward advisory lock. I think rather than us deciding, we should ask them what they think is best.

I lost track of what your complaint is now. Do you fear, someone will ever take the courage to look into the pulpcore/tasking directory? Or is this still about which services heartbeat we piggyback to make this new feature work? I think we agree that we are not going to create a new service for it (i will so call it `pulpcore-resource-manager` if you talk me into).

CHANGES/2204.feature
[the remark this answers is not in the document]
bmbouter: At some point it probably will get API access of various sorts, but is right now the time? Is this actually an MVP requirement? More perspectives on this point would be helpful; perhaps I'm misaligned with the broader thinking on the needs here.

bmbouter:
> I lost track of what your complaint is now. Do you fear, someone will ever take the courage to look into the pulpcore/tasking directory? Or is this still about which services heartbeat we piggyback to make this new feature work? I think we agree that we are not going to create a new service for it (i will so call it `pulpcore-resource-manager` if you talk me into).

We definitely agree there can't be a new service for it. We never discussed (or agreed) on which existing service will dispatch it. My concern with making the tasking code also perform dispatching is that the pool of people who support this feature then drops to 3, just the SMEs for the tasking system. My belief is that it could just as easily be dispatched by the heartbeat process in pulpcore-content and the number of folks who could support that is the entire pulpcore team (and really any developer).

I'm not understanding the conclusion that because it dispatches tasks it has to live in the worker code. Pulp dispatches tasks all over the codebase and anyone can support them, which I think is a good thing. Perhaps my perspective has a lot to do with me being a tasking system SME.  already feel it's very complex (necessarily so), but I don't want to make it more complex if we don't have to.

mdellweg:
> > I lost track of what your complaint is now. Do you fear, someone will ever take the courage to look into the pulpcore/tasking directory? Or is this still about which services heartbeat we piggyback to make this new feature work? I think we agree that we are not going to create a new service for it (i will so call it `pulpcore-resource-manager` if you talk me into).
> 
> We definitely agree there can't be a new service for it. We never discussed (or agreed) on which existing service will dispatch it. My concern with making the tasking code also perform dispatching is that the pool of people who support this feature then drops to 3, just the SMEs for the tasking system. My belief is that it could just as easily be dispatched by the heartbeat process in pulpcore-content and the number of folks who could support that is the entire pulpcore team (and really any developer).

I will not take any head count as an argument for a software design decision here. Sorry. If the tasking system was in a separate repo, maybe. But i will give you technical reasons instead:
1. `dispatch` cannot be safely called from async context
2. The postgres advisory locks are employed by raw commands on djangos db connection. Doing such a thing in async context makes me feel very uneasy.

> 
> I'm not understanding the conclusion that because it dispatches tasks it has to live in the worker code. Pulp dispatches tasks all over the codebase and anyone can support them, which I think is a good thing. Perhaps my perspective has a lot to do with me being a tasking system SME. The worker code already feels very complex (necessarily so), but I don't want to make it more complex if we don't have to.

It does not have to, but it appears to be the most natural place, both indicated by the language being used, as well as the number of imports needed to support it, and with the advisory locks even the similarity of technology used. Having designed the tasking system, i'm totally biased here: I would not call the tasking system very complex at all. It is almost self contained in 421 lines of code, does not depend on complicated external libraries and i don't see any unnecessary bells and whistles.

Sure, i'd like to see more people step up and familiarize with the tasking code. But the same argument goes for every other area of code in this repo.

pulpcore/tasking/tasks.py
mdellweg: Not exactly sure, what you mean. Is this a timezone thing, when your server moves borders? Or do you think about a few seconds correction around Sylvester?
Do you have another suggestion for it?

pulpcore/tasking/tasks.py
```diff
@@ -76,13 +86,42 @@ def dispatch(
...
+        transaction.on_commit(_wakeup_worker)
+
     return task
+
+
+def dispatch_scheduled_tasks():
+    # Warning, dispatch_scheduled_tasks is not race condition free!
+    now = timezone.now()
+    # Dispatch all tasks old enough and not still running
+    for task_schedule in TaskSchedule.objects.filter(next_dispatch__lte=now).filter(
```
ipanova, line 104: this does not seem to exclude tasks in running state

pulpcore/tasking/tasks.py
dralley: Daylight savings time.  I'm not concerned about the server physically switching timezones, but a 30 to 60 minute jump forwards and backwards would mess with the schedules.

pulpcore/tasking/tasks.py
mdellweg: It searches for Schedules whose last task is either nonexistant or in a finished state. So we would not dispatch a new one before the last one finished to prevent bloating the task queue.

pulpcore/tests/functional/api/test_workers.py
```diff
@@ -98,3 +101,60 @@ def test_05_http_method(self):
...
             self.client.delete(self.worker["pulp_href"])
+
+
+def dummy_task():
+    """Dummy task, that can be used in tests."""
+    pass
+
+
+@pytest.fixture(scope="session")
+def cli_client():
```
bmbouter, line 112: I think this makes sense to be moved to pulp-smash. Wdyt?

pulpcore/tests/functional/api/test_workers.py
```diff
@@ -98,3 +101,60 @@ def test_05_http_method(self):
...
+    )
+    yield {"name": name, "task_name": task_name}
+    utils.execute_pulpcore_python(
+        cli_client,
+        "from pulpcore.app.models import TaskSchedule;"
+        f"TaskSchedule.objects.get(name='{name}').delete();",
+    )
+
+
+@pytest.mark.parallel
```
bmbouter, line 141: Sure is nice this is parallel!

pulpcore/tests/functional/api/test_workers.py
```diff
@@ -98,3 +101,60 @@ def test_05_http_method(self):
         """
         with self.assertRaises(HTTPError):
             self.client.delete(self.worker["pulp_href"])
+
+
+def dummy_task():
```
bmbouter, line 106: One issue with this is that we want to run the test suite against non-source installs, and those won't have this task in the system under test so the test wouldn't work there. I think we need to fix this before merging. Here are a few ideas:

* Move it somewhere in the `pulpcore.tasking` module
* Use another existing task that doesn't do anything meaningful

pulpcore/tests/conftest.py
```diff
@@ -1 +1,9 @@
+import pytest
+
 from .conftest_pulp_file import *  # noqa
+from pulpcore.client.pulpcore import TaskSchedulesApi
+
+
+@pytest.fixture(scope="session")
+def task_schedules_api_client(pulpcore_client):
```
bmbouter, line 8: I have been putting these in pulp-smash since any plugin may also need them. I've been putting them here https://github.com/pulp/pulp-smash/blob/main/pulp_smash/pulp3/pytest_plugin/__init__.py#L241-L261

pulpcore/app/models/task.py
```diff
@@ -332,3 +333,16 @@ class CreatedResource(GenericRelationModel):
     task = models.ForeignKey(
         Task, related_name="created_resources", default=Task.current, on_delete=models.CASCADE
     )
+
+
+class TaskSchedule(BaseModel):
+    name = models.CharField(max_length=256, unique=True, null=False)
```
bmbouter, line 339: I think we should add args and kwargs to this. I believe galaxyNG's initial usage would require that.

pulpcore/app/viewsets/task.py
```diff
@@ -242,3 +243,70 @@ class WorkerViewSet(NamedModelViewSet, mixins.RetrieveModelMixin, mixins.ListMod
...
+class TaskScheduleViewSet(
+    NamedModelViewSet,
+    mixins.RetrieveModelMixin,
+    mixins.ListModelMixin,
+    RolesMixin,
+):
+    """
+    ViewSet to monitor task schedules.
+
+    NOTE: This feature is in tech-preview and may change in backwards incompatible ways.
```
bmbouter, line 269: :+1:

pulpcore/app/models/task.py
mdellweg: The locks should really be tied to the kwargs we provide a task with. The way we call tasks today is not suitable for this feature. At least i do not know a way how to implement such a thing in a safe way.

@newswangerd My perception was tasks with arguments would be nice to have but the immediate need is for a periodically scheduled maintenance task. Can you confirm?

pulpcore/tests/functional/api/test_workers.py
mdellweg: We can have it here now then add to pulp-smash once it proved useful and remove it here again. pytest lookup should allow that.

pulpcore/tests/functional/api/test_workers.py
mdellweg: if only we had unuseful tasks. I scheduled orphan_cleanup before. But that would prevent `@parallel`

pulpcore/tests/functional/api/test_workers.py
mdellweg: I'll make a pulp-smash pr.

pulpcore/app/models/task.py
bmbouter: I'm ok with whatever you and @newswangerd work out in this area.

pulpcore/tasking/tasks.py
bmbouter: I think because it's `timezone` from django it addresses DST. I could be wrong about that.

pulpcore/tests/conftest.py
bmbouter: It's not a requirement pre-merging. My take is that by not putting everything in one place we're going to end up doing the same work over and over. Also it'll be extra work to put it here, later move it, etc. The world is harder when we don't stay DRY.

pulpcore/tests/functional/api/test_workers.py
bmbouter: Oh another idea! This could be moved to `pulpcore.tests.tasks` and the [setup.py `find_packages`](https://github.com/pulp/pulpcore/blob/main/setup.py#L16) could be adjusted to only include that one module from `pulpcore.tests`. I think that is really the best because it keeps it out of the application module, but still delivered with production code.

pulpcore/tests/functional/api/test_workers.py
bmbouter: Thanks!

CHANGES/2204.feature
```diff
@@ -0,0 +1,3 @@
+Added TaskSchedule model that will periodically schedule configured tasks.
```
bmbouter, line 1: I think this should be rewritten to only speak in user-facing terms. To them all they got was a viewset. Also use double backticks to surround escaled things, e.g. the URL.

bmbouter (approved):
I tried to comment on which comments from me are blockers versus not. I won't be here next week though, and it would be great to include this in 3.18 so I'm lgtm-ing now. Thank you so much for working on this!

@dralley I believe you and @newswangerd are the final reviewers.

pulpcore/tests/functional/api/test_workers.py
mdellweg: https://github.com/pulp/pulp-smash/pull/1278

pulpcore/tests/functional/api/test_workers.py
mdellweg: We agreed off list that we can just have a dummy task in the codebase (which is basically the same outcome wrt code in production to the two suggestions above).
I went with `pulpcore.app.tasks.test.dummy_task`.

pulpcore/tests/conftest.py
mdellweg: I really see this fixture staying here for eternity, or at least until one plugin needs to start using it. And now i suddenly see plugins _wanting_ to use it...

mdellweg:
This requres on https://github.com/pulp/pulp-smash/pull/1278 now.

pulpcore/tests/conftest.py
mdellweg: Moved it too. And tried to come up with a nice failing scheme if the import fails.
https://github.com/pulp/pulp-smash/pull/1278

pulpcore/app/serializers/task.py
```diff
@@ -204,3 +204,28 @@ class WorkerSerializer(ModelSerializer):
...
+class TaskScheduleSerializer(ModelSerializer):
+    pulp_href = IdentityField(view_name="task-schedules-detail")
+    name = serializers.CharField(help_text=_("The name of the task schedule."), allow_blank=False)
+    task_name = serializers.CharField(help_text=_("The name of the task to be scheduled."))
+    dispatch_interval = serializers.DurationField(help_text=_("Periodicity of the schedule."))
+    next_dispatch = serializers.DateTimeField(
+        help_text=_("Timestamp of the next time the task will be dispatched."), read_only=True
+    )
+    last_task = RelatedField(
+        help_text=_("The last task being dispatched by this schedule."),
```
dkliban, line 218: s/task being dispatched/task dispatched/.

CHANGES/2204.feature
```diff
@@ -0,0 +1,3 @@
+Added aread only task schedule view for configured tasks schedules.
```
ipanova, line 1: what is this word? dictionary says this is an archaic or obsolete one

dkliban (approved):

CHANGES/2204.feature
dralley: "aread"?  I think it's stupposed to be "a read-only"

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -127,6 +149,8 @@ def beat(self):
             if self.worker_cleanup_countdown <= 0:
                 self.worker_cleanup_countdown = WORKER_CLEANUP_INTERVAL
                 self.worker_cleanup()
+            with suppress(AdvisoryLockError), PGAdvisoryLock(42):
```
dralley, line 152: Does `42` hold special significance, outside of the [obvious one](https://en.wikipedia.org/wiki/42_(number)#The_Hitchhiker's_Guide_to_the_Galaxy)?

I'm thinking it could be a named constant.

mdellweg: It's randomly chosen, but it's not colliding with the lockspace used for task locks.

dralley: Could you leave a comment to that affect

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -51,6 +52,27 @@
 WORKER_CLEANUP_INTERVAL = 100
 
 
+class PGAdvisoryLock:
```
dralley, line 55: +1

pulpcore/tasking/tasks.py
```diff
@@ -76,13 +86,47 @@ def dispatch(
...
+    ):
+        try:
+            if task_schedule.dispatch_interval is None:
+                # This was a timed one shot task schedule
+                task_schedule.next_dispatch = None
+            else:
+                # This is a recurring task schedule
+                while task_schedule.next_dispatch < now:
+                    # Do not schedule in the past
+                    task_schedule.next_dispatch += task_schedule.dispatch_interval
```
dralley, line 115: I would just like us to have some documentation about the behavior of task scheduling.  e.g. this could lead to an hourly task occurring minutes apart after an outage window.

And that's totally fine and valid as long as it's documented and understood.

dralley: @mdellweg I told Dennis this was OK to merge, the comments/documentation can be added later.

pulpcore/app/models/task.py
newswangerd: @mdellweg Hey, sorry I missed this. This should work for scheduling the analytics task which is fine for now, but I was also thinking we'd use this feature to schedule syncs, which is impossible without kwargs and args (since you need to pass the repo and remote to the task).

Anyway, this is fine for now.

PR #2210, opened 2022-02-09
Adds deprecation warning ACCESS_POLICY_VIEW_NAME

### LINKED ISSUE #2209: Deprecate `ACCESS_POLICY_VIEWSET_NAME` from models
While developing https://github.com/pulp/pulpcore/pull/2198/files we realized we already have `pulpcore.app.utils.get_viewset_for_model` that can find the viewset from a model without this attribute. It's no longer in-use and for clarity should no longer be defined.

This issue is to produce a deprecation warning if it's defined.

bmbouter:
The `ACCESS_POLICY_VIEW_NAME` attribute on models is no longer in use by
the RBAC machinery and is discouraged from being defined. If present on
a model a deprecation warning is logged.

closes #2209

### DIFF
--- a/pulpcore/app/models/base.py
+++ b/pulpcore/app/models/base.py
@@ -8,6 +8,8 @@ from django.db.models import options
 from django.db.models.base import ModelBase
 from django_lifecycle import LifecycleModel
 
+from pulpcore.app.loggers import deprecation_logger
+
 
 class Label(LifecycleModel):
     """Model for handling resource labels.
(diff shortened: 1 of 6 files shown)


mdellweg (approved):

pulpcore/app/models/base.py
```diff
@@ -67,6 +69,14 @@ class BaseModel(LifecycleModel):
     class Meta:
         abstract = True
 
+    def __init__(self, *args, **kwargs):
+        if hasattr(self, "ACCESS_POLICY_VIEWSET_NAME"):
+            deprecation_logger.warn(
+                f"The model {self.__class__} defines the 'ACCESS_POLICY_VIEWSET_NAME' class "
+                f"attribute which is no longer required and is discouraged."
```
ipanova, line 76: "..and is discouraged to be defined."

ipanova (approved):

PR #2224, opened 2022-02-16
Deprecate the API_ROOT constant

mdellweg:
It was turned into a setting.

re #2148
[noissue]

### DIFF
--- a/CHANGES/plugin_api/2148.deprecation
+++ b/CHANGES/plugin_api/2148.deprecation
@@ -1,0 +1,2 @@
+The ``API_ROOT`` constant has been deprecated and turned into a setting. Its removal is scheduled
+for 3.20. Please use the setting with the same name instead.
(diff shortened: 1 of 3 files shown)


fao89 (approved):
LGTM!

CHANGES/plugin_api/2148.removal
```diff
@@ -0,0 +1,2 @@
+The ``API_ROOT`` constant has been deprecated and is scheduled for removal in 3.20. Please use the
```
ipanova, line 1: the changelog entry should be then .deprecation and not removal, i think

ipanova: how about "The ``API_ROOT`` constant has been deprecated and turned into a setting.  Its removal is scheduled for 3.20. Please use the setting instead."

bmbouter:
Excellentay!

PR #2235, opened 2022-02-16
Use V3_API_ROOT_NO_FRONT_SLASH on deprecation

fao89:
re #2148

What used to be API_ROOT is V3_API_ROOT_NO_FRONT_SLASH now
https://github.com/pulp/pulpcore/blob/3.17/pulpcore/constants.py#L47

Noticed it when fixing pulp_file tests:
https://github.com/pulp/pulp_file/pull/643

### DIFF
--- a/pulpcore/plugin/constants.py
+++ b/pulpcore/plugin/constants.py
@@ -15,6 +15,7 @@ def API_ROOT():
 
     deprecation_logger.warn(
         "The API_ROOT constant has been deprecated and turned into a setting. Please use "
-        "`settings.API_ROOT` instead. This symbol will be deleted with pulpcore 3.20."
+        "`settings.V3_API_ROOT_NO_FRONT_SLASH` instead. This symbol will be deleted with pulpcore "
+        "3.20."
     )
-    return settings.API_ROOT
+    return settings.V3_API_ROOT_NO_FRONT_SLASH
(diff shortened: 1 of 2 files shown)


pulpcore/plugin/constants.py
```diff
@@ -15,6 +15,7 @@ def API_ROOT():
 
     deprecation_logger.warn(
         "The API_ROOT constant has been deprecated and turned into a setting. Please use "
-        "`settings.API_ROOT` instead. This symbol will be deleted with pulpcore 3.20."
+        "`settings.V3_API_ROOT_NO_FRONT_SLASH` instead. This symbol will be deleted with pulpcore "
+        "3.20."
     )
-    return settings.API_ROOT
+    return settings.V3_API_ROOT_NO_FRONT_SLASH
```
bmbouter, line 21: I think you actually want `V3_API_ROOT` because that's the var users used to get before (I think). Is that right?

fao89: it didn't have the front slash: https://github.com/pulp/pulpcore/blob/3.17/pulpcore/constants.py#L47

bmbouter: Yup you right. Thanks!

mdellweg (approved):

PR #2263, opened 2022-02-24
Expose PulpRemoteUserAuthentication to plugin writers

### LINKED ISSUE #2262: Expose PulpRemoteUserAuthentication to plugin writers
This will allow plugin writers to take a rest_framework remote authentication class into account when building protected endpoints.

lubosmj:
closes #2262

### DIFF
--- a/CHANGES/plugin_api/2262.feature
+++ b/CHANGES/plugin_api/2262.feature
@@ -1,0 +1,2 @@
+Exposed the ``PulpRemoteUserAuthentication`` class to plugin writers. This will allow the use of
+remote authentication methods when building protected endpoints.
--- a/pulpcore/plugin/authentication/__init__.py
+++ b/pulpcore/plugin/authentication/__init__.py
@@ -1,0 +1,1 @@
+from pulpcore.app.authentication import PulpRemoteUserAuthentication  # noqa


lubosmj:
The change is required for pulp_container where we need to consider users who were authenticated remotely by a web server since the access to specific endpoints is managed exceptionally in the plugin.

CHANGES/plugin_api/2262.feature
```diff
@@ -0,0 +1,2 @@
+Exposed the ``PulpRemoteUserAuthentication`` class to plugin writers. This will enable them to
+consider remote authentication scenarios when building protected endpoints.
```
ipanova, line 2: you can shorten it with: "This enables remote authentication scenarios.."

lubosmj: This is not correct. It does not enable authentication scenarios itself.

You as a plugin writer will have rather a chance to override or utilize such a class if you want to manage access for particular endpoints (like we do in pulp_container).

ipanova: `to enable` in this context is read as 'to make possible remote authentication scenarios'

daviddavis: "This will allow them to use remote authentication methods when building protected endpoints"

maybe?

ipanova (approved):

pulpcore/plugin/authentication/__init__.py
```diff
@@ -0,0 +1 @@
+from pulpcore.app.authentication import PulpRemoteUserAuthentication  # noqa
```
ipanova, line 1: i think it should follow this format https://github.com/pulp/pulpcore/blob/34b490c50594bb62bdc99951acd96734beb5f5e2/pulpcore/plugin/access_policy.py

mdellweg (approved):

fao89 (approved):

PR #2261, opened 2022-02-24
Fix delete repository version causing "duplicate key" error

hao-yu:
Add_and_remove task might fail with "duplicate key value violates
unique constraint" error after deleting a repository version.
    
Also added a test to cover this scenario.
    
fixes #2047
fixes #2267

PR #2264, opened 2022-02-24
Harden operations around repo versions

mdellweg:


mdellweg:
I added two assert statements that should ever raise an exception. Do we want such a thing?

pulpcore/app/models/repository.py
```diff
@@ -1016,10 +1021,11 @@ def __exit__(self, exc_type, exc_value, traceback):
 
                     self.complete = True
                     self.repository.next_version = self.number + 1
-                    self.repository.save()
-                    self.save()
+                    with transaction.atomic():
+                        self.repository.save()
+                        self.save()
+                        self._compute_counts()
                     self.repository.cleanup_old_versions()
```
gerrod3, line 1028: This will probably re-introduce the 500 errors when listing repo-versions after deleting intermediary versions as the counts really need to be updated if old versions are removed.

pulpcore/app/models/repository.py
```diff
@@ -221,8 +221,12 @@ def artifacts_for_version(version):
     @hook(AFTER_UPDATE, when="retain_repo_versions", has_changed=True)
     def cleanup_old_versions(self):
         """Cleanup old repository versions based on retain_repo_versions."""
+        assert not self.versions.filter(complete=False).exists()
         if self.retain_repo_versions:
-            for version in self.versions.order_by("-number")[self.retain_repo_versions :]:
+            # Consider only completed versions for squashing
+            for version in self.versions.complete().order_by("-number")[
```
dralley, line 227: t will never make it to this filter because of the assertion, right?  Is the idea to just be extra defensive?

mdellweg: extra defensive, yes. This whole thing can lead to disastrous data loss.
I reality, i think this is the "fix" and the assertion should help us identify situations that should not have happened. Also (unconfirmed for python) aren't there situations where you can configure assertions to be no-ops in production code?

mdellweg: Looks like we can get a better description of the original bug and maybe even a reproducer from galaxy_ng soon.

pulpcore/app/models/repository.py
mdellweg: Isn't that handled by the `_compute_counts` call in `_squash`? Put another way, what corner case are we missing here?

I really want to have the updated counts saved in the same transaction as the version is tagged final.
Maybe the transactioneity of the `delete` in `cleanup_old_versions` is not proper. This is also pointed out by https://github.com/pulp/pulpcore/pull/2261 .

pulpcore/app/models/repository.py
```diff
@@ -221,8 +221,12 @@ def artifacts_for_version(version):
     @hook(AFTER_UPDATE, when="retain_repo_versions", has_changed=True)
     def cleanup_old_versions(self):
         """Cleanup old repository versions based on retain_repo_versions."""
+        assert not self.versions.filter(complete=False).exists()
         if self.retain_repo_versions:
-            for version in self.versions.order_by("-number")[self.retain_repo_versions :]:
+            # Consider only completed versions for squashing
```
dralley, line 226: Given the importance of these checks, it might be prudent to expand a bit on why the checks are here, maybe including a link to the issue.

pulpcore/app/models/repository.py
dralley: >Also (unconfirmed for python) aren't there situations where you can configure assertions to be no-ops in production code?

There is a flag you can start the Python interpreter with where it will ignore assertions, yup.  `-O`

pulpcore/app/models/repository.py
mdellweg: The longer i think about it, `cleanup_old_versions` calls `delete` and that calls `_squash` and that one needs to care about the counts of the squashed into version (in a transaction).

mdellweg:
As stated in the comment, I have no idea how to trigger the situation in the tests. But i am happy to discuss (and improve the comments) on every single hardening step taken here.

pulpcore/app/models/repository.py
```diff
@@ -923,22 +937,33 @@ def delete(self, **kwargs):
...
-                self._squash(repo_relations, next_version)
-
-            except RepositoryVersion.DoesNotExist:
-                # version is the latest version so simply update repo contents
-                # and delete the version
-                repo_relations.filter(version_added=self).delete()
-                repo_relations.filter(version_removed=self).update(version_removed=None)
-            super().delete(**kwargs)
+            # Handle the manipulation of the repository version content and its final deletion in
+            # the same transaction.
```
dralley, line 948: +1

CHANGES/2268.bugfix
```diff
@@ -0,0 +1 @@
+Added transactioness around repository version operations to prevent data loss.
```
dralley, line 1: Typo "transactioness"

pulpcore/app/models/repository.py
```diff
@@ -219,10 +221,22 @@ def artifacts_for_version(version):
         return Artifact.objects.filter(content__pk__in=version.content)
 
     @hook(AFTER_UPDATE, when="retain_repo_versions", has_changed=True)
+    def _cleanup_old_versions_hook(self):
+        # Do not attempt to clean up anything, while there is a transaction involving repo versions
+        # still in flight.
+        transaction.on_commit(self.cleanup_old_versions)
```
dralley, line 227: +1

CHANGES/2268.bugfix
mdellweg: Sorry, i don't see the correction you are supposing.

dralley: "transactioness" -> "transactions"

pulpcore/app/models/repository.py
```diff
@@ -983,6 +1008,7 @@ def __enter__(self):
         Returns:
             RepositoryVersion: self
         """
+        assert not self.complete
```
bmbouter, line 1011: I think raising an exception would be better. My understanding is that `assert` for production code because people can disable them with an interpreter switch.

dralley: I don't fully understand that motivation because its true of pretty much any language, and it's rare that the interpreter switch actually gets used.  I don't object to it though.

CHANGES/2268.bugfix
mdellweg: Oh, i meant the code to be more transactionessy. But you are right: In the end i added transactions.

pulpcore/app/models/repository.py
mdellweg: I don't really know if using assertions in production code is a good or a bad habit. Hardcoding an exception here will protect the user even if he switches off assertions for speed.

dralley (approved):

pulpcore/app/models/repository.py
```diff
@@ -923,22 +938,32 @@ def delete(self, **kwargs):
         Deletion of a complete RepositoryVersion should be done in a RQ Job.
         """
         if self.complete:
+            if self.repository.versions.complete().count() <= 1:
+                raise APIException("Attempt to delete the last remaining version.")
```
bmbouter, line 942: Maybe this should be i8ln with `gettext`?

pulpcore/app/models/repository.py
bmbouter: +1 but maybe the comment isn't necessary? It's up to you.

pulpcore/app/models/repository.py
```diff
@@ -923,22 +938,32 @@ def delete(self, **kwargs):
...
+                    self._squash(repo_relations, next_version)
+
+                except RepositoryVersion.DoesNotExist:
+                    # version is the latest version so simply update repo contents
+                    # and delete the version
+                    repo_relations.filter(version_added=self).delete()
+                    repo_relations.filter(version_removed=self).update(version_removed=None)
+
+                if repo_relations.filter(Q(version_added=self) | Q(version_removed=self)).exists():
+                    raise RuntimeError("Some repo relations of this version were not translated.")
```
bmbouter, line 965: This should be i8ln too please.

pulpcore/app/models/repository.py
```diff
@@ -983,6 +1008,10 @@ def __enter__(self):
         Returns:
             RepositoryVersion: self
         """
+        if self.complete:
+            raise RuntimeError(
+                "This Repository version is complete. It cannot be modified further."
```
bmbouter, line 1013: i8ln here too please.

pulpcore/app/models/repository.py
```diff
@@ -219,10 +221,23 @@ def artifacts_for_version(version):
...
+        # still in flight.
+        transaction.on_commit(self.cleanup_old_versions)
+
     def cleanup_old_versions(self):
         """Cleanup old repository versions based on retain_repo_versions."""
+        # I am still curious how, but it was reported that this state can happen in day to day
+        # operations but its easy to reproduce manually in the pulpcore shell:
+        # https://github.com/pulp/pulpcore/issues/2268
+        if self.versions.filter(complete=False).exists():
+            raise RuntimeError("Attempt to cleanup old versions, while a new version is in flight.")
```
bmbouter, line 235: i8ln here too please.

bmbouter (approved):
Correctness-wise this is great, thanks! I am requesting some i8ln and maybe fewer comments (optional) before merging, but feel free to fix-up and then merge. The i8ln I think is important since galaxy_ng's use of Pulp is actually translated into various languages.

Thanks for making this!

pulpcore/app/models/repository.py
dralley: I like the comment, there is enough going on here that a brief explanation of what is going on makes the "why" more evident

dralley:
We should backport this to 3.16 as well - or press Katello to be using 3.17 instead, which is honestly my preference.

Right now Sat 7 is targeting 3.16

mdellweg:
> We should backport this to 3.16 as well - or press Katello to be using 3.17 instead, which is honestly my preference.
> 
> Right now Sat 7 is targeting 3.16

I added the patchback label, because that is easy. But also I think, katello by using the api exclusively is not so much at risk.

mdellweg:
Manual 3.16 cherrypick:
https://github.com/pulp/pulpcore/pull/2274

PR #2265, opened 2022-02-24
Add comments and docs to task schedules

mdellweg:
re #2208

[noissue]

### DIFF
--- a/docs/plugins/reference/task-scheduling.rst
+++ b/docs/plugins/reference/task-scheduling.rst
@@ -1,0 +1,17 @@
+.. _task-scheduling:
+
+Task Scheduling
+---------------
+
+.. warning: This feature is in tech-preview and only accessible by direct manipulation of
+   ``TaskSchedule`` objects. It is targeted for plugin writers and no api access is planned.
+
+Pulp supports scheduling of tasks. Scheduled tasks will be dispatched shortly after their
+``next_dispatch`` time, and be rescheduled one ``dispatch_interval`` after that, if the latter is
+set. By specifying the ``dispatch_interval`` as ``time_delta(days=1)`` you can expect the task
+dispatch to stabily happen at same time every day. Until the last task of the same schedule enters a
+final state, a new task will not be dispatched. Scheduling is done by the worker processes,
+therefore scheduled task dispatching will be missed if all workers are offline. After an outage
+window, overdue schedules will dispatch at most one task, but down to timing, they may be
+rescheduled shortly thereafter. The task schedule API at ``/pulp/api/v3/task-schedules/`` is
+provided to read the tasks schedules.
--- a/pulpcore/tests/functional/api/test_workers.py
+++ b/pulpcore/tests/functional/api/test_workers.py
@@ -112,7 +112,7 @@ def task_schedule(cli_client):
         "from django.utils.timezone import now;"
         "from datetime import timedelta;"
         "from pulpcore.app.models import TaskSchedule;"
-        "dispatch_interval = timedelta(seconds=2);"
+        "dispatch_interval = timedelta(seconds=4);"
         "next_dispatch = now() + dispatch_interval;"
         "TaskSchedule("
         f"    name='{name}', task_name='{task_name}', "
@@ -131,7 +131,7 @@ def task_schedule(cli_client):
 def test_task_schedule(task_schedule, task_schedules_api_client):
     """Test that a worker will schedule a task roughly at a given time."""
     # Worker TTL is configured to 30s, therefore they will have a heartbeat each 10s (6 bpm). The
-    # task is scheduled 2s in the future to give us time to invesitgate the state before and after.
+    # task is scheduled 4s in the future to give us time to invesitgate the state before and after.
     # 15s later we can be sure it was scheduled (as long as at least one worker is running).
 
     result = task_schedules_api_client.list(name=task_schedule["name"])
--- a/docs/plugins/reference/index.rst
+++ b/docs/plugins/reference/index.rst
@@ -14,3 +14,4 @@ This section includes in-depth material that is topic specific.
    releasing
    content-protection
    metadata-signing
+   task-scheduling
--- a/pulpcore/tasking/pulpcore_worker.py
+++ b/pulpcore/tasking/pulpcore_worker.py
@@ -50,9 +50,18 @@ random.seed()
 
 TASK_GRACE_INTERVAL = 3
 WORKER_CLEANUP_INTERVAL = 100
+# Randomly chosen
+TASK_SCHEDULING_LOCK = 42
 
 
 class PGAdvisoryLock:
+    """
+    A context manager that will hold a postgres advisory lock non-blocking.
+
+    The locks can be chosen from a lock group to avoid collisions. They will never collide with the
+    locks used for tasks.
+    """
+
     def __init__(self, lock, lock_group=0):
         self.lock_group = lock_group
         self.lock = lock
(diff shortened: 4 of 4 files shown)


mdellweg:
@dralley

docs/plugins/reference/task-scheduling.rst
```diff
@@ -0,0 +1,15 @@
...
+Task Scheduling
+---------------
+
+.. warning: This feature is in tech-preview and only accessible by direct manipulation of
+   ``TaskSchedule`` objects.
+
+Pulp supports scheduling of tasks.
+Scheduled tasks will be dispatched shortly after their ``next_dispatch`` time, and be rescheduled one ``dispatch_interval`` after that, if the latter is set.
+By specifying the ``dispatch_interval`` as ``time_delta(days=1)`` you can expect the task dispatch to stabily happen at same time every day.
+Unless the last task of the same schedule did enter a final state, no new task will be dipatched.
```
dralley, line 12: typo "dipatched"

docs/plugins/reference/task-scheduling.rst
```diff
@@ -0,0 +1,15 @@
...
+Scheduling is done by the worker processes, therefore schedules will be missed if all workers are offline.
+After an outage window, overdue schedules will dispatch at most one task, but down to timing, they my be rescheduled shortly thereafter.
+The task schedule API at ``/pulp/api/v3/task-schedules/`` is provided to monitor the states of tasks schedules.
```
dralley, line 15: +1 good explanation

docs/plugins/reference/task-scheduling.rst
```diff
@@ -0,0 +1,15 @@
+.. _task-scheduling:
+
+Task Scheduling
+---------------
+
+.. warning: This feature is in tech-preview and only accessible by direct manipulation of
+   ``TaskSchedule`` objects.
+
+Pulp supports scheduling of tasks.
```
bmbouter, line 9: Can the docs style be switched to not line wrap? I believe that's the documented style in the style part of the docs.

bmbouter:
I'm so glad to see these docs!

docs/plugins/reference/task-scheduling.rst
mdellweg: Oh, this is draft, i'll be wrapping them the moment we agreed on every single word. Not wanting to rewrap them over and over.

docs/plugins/reference/task-scheduling.rst
```diff
@@ -0,0 +1,15 @@
...
+Scheduled tasks will be dispatched shortly after their ``next_dispatch`` time, and be rescheduled one ``dispatch_interval`` after that, if the latter is set.
```
bmbouter, line 10: I guess I didn't realize `dispatch_interval` was optional. I figured we didn't support one-off tasks, but anyways this reads well.

docs/plugins/reference/task-scheduling.rst
bmbouter:
```suggestion
If the last task of the same schedule did enter a final state, no new task will be dispatched.
```

docs/plugins/reference/task-scheduling.rst
```diff
@@ -0,0 +1,15 @@
...
+By specifying the ``dispatch_interval`` as ``time_delta(days=1)`` you can expect the task dispatch to stabily happen at same time every day.
+Unless the last task of the same schedule did enter a final state, no new task will be dipatched.
+Scheduling is done by the worker processes, therefore schedules will be missed if all workers are offline.
```
bmbouter, line 13:
```suggestion
Scheduling is done by the worker processes, therefore scheduled task dispatching will be missed if all workers are offline.
```

docs/plugins/reference/task-scheduling.rst
bmbouter, line 15:
```suggestion
The task schedule API at ``/pulp/api/v3/task-schedules/`` is provided to read the tasks schedules.
```

bmbouter:
I left some language ideas, but overall this reads really well. Thank you!

docs/plugins/reference/task-scheduling.rst
```diff
@@ -0,0 +1,15 @@
...
+After an outage window, overdue schedules will dispatch at most one task, but down to timing, they my be rescheduled shortly thereafter.
```
ipanova, line 14: s/my be/may be

docs/plugins/reference/task-scheduling.rst
ipanova, line 12: Unless the last task enters... no new task will be...
or
If the last task does not enter...no new task will be..

conditionals :sob:

docs/plugins/reference/task-scheduling.rst
ipanova, line 15: I keep forgetting what 'guidelines' we follow when we choose name for new endpoints. Specifically whether to use underscore or hyphens. It seems like 'access policies is written through underscore and now we use hyphens :see_no_evil:

docs/plugins/reference/task-scheduling.rst
```diff
@@ -0,0 +1,15 @@
+.. _task-scheduling:
+
+Task Scheduling
+---------------
+
+.. warning: This feature is in tech-preview and only accessible by direct manipulation of
+   ``TaskSchedule`` objects.
```
ipanova, line 7: do you think it's worth mentioning that this will not be available through the API and mainly is targeted for the plugin writers usage?

docs/plugins/reference/task-scheduling.rst
mdellweg: Apart from the typo the sentence was correct. Not entering a final state will block this very schedule from recurring.

docs/plugins/reference/task-scheduling.rst
mdellweg: It was to easy to also get one off schedules that way to just leave it out.

docs/plugins/reference/task-scheduling.rst
mdellweg: BTW, at the same time there will be no worker to handle the tasks anyway.

docs/plugins/reference/task-scheduling.rst
mdellweg:
```suggestion
After an outage window, overdue schedules will dispatch at most one task, but down to timing, they may be rescheduled shortly thereafter.
```

docs/plugins/reference/task-scheduling.rst
mdellweg: BTW, at the same time there will be no worker to handle the tasks anyway.

docs/plugins/reference/task-scheduling.rst
mdellweg:
```suggestion
Until the last task of the same schedule enters a final state, a new task will not be dispatched.
```

docs/plugins/reference/task-scheduling.rst
mdellweg: Oh snap! I can see both. With a clear winner of underscore. But still it's probably too late to "fix" the use at `task-groups`, `signing-services` and `import-check`...

ipanova: yeah..agreed.

ipanova (approved):

pulpcore/tests/functional/api/test_workers.py
```diff
@@ -131,7 +131,7 @@ def task_schedule(cli_client):
 def test_task_schedule(task_schedule, task_schedules_api_client):
     """Test that a worker will schedule a task roughly at a given time."""
     # Worker TTL is configured to 30s, therefore they will have a heartbeat each 10s (6 bpm). The
-    # task is scheduled 2s in the future to give us time to invesitgate the state before and after.
+    # task is scheduled 4s in the future to give us time to invesitgate the state before and after.
```
mdellweg, line 134: I adjusted the timings here, since this test seems to fail occasionally in the ci (maybe due to increased parallel testing).

bmbouter (approved):
Looks good to me. Thanks!

PR #2270, opened 2022-02-28
Fix note directives

tipabu:
Previously, this was rendering more like a literal

> ..note::
>
> Pulp will mostly automatically manage the following directories for you. Only if you need to adjust permissions or security contexts and perform a manual installation, you need to prepare them accordingly.


ipanova:
@tipabu can you add a [noissue] to the commit message? example https://github.com/pulp/pulpcore/pull/2259/commits

dralley (approved):

mdellweg (approved):

PR #2271, opened 2022-03-01
Improve error messages when Pulp encounters digest / size validation …

### LINKED ISSUE #2244: When an error occurs during content sync, due to hash conflicts, report conflicting hashes
Recently had an issue relating to synchronising content, due to a conflict between an expected hash, and an actual hash of the data recieved.
Currently, we print out the URL of the content being grabbed, but no information relating to the conflicting hashses.

I would like the error message to be extended to include both the expected hash as well as the hash of the retrieved data.

Spending a lot of time running around in a circle

**Additional context**
After making some code changes to achieve the above, it made tracking down https://github.com/pulp/pulp_rpm/issues/2407 much much much easier.

dralley:
…error

closes #2244

### DIFF
--- a/CHANGES/2244.misc
+++ b/CHANGES/2244.misc
@@ -1,0 +1,1 @@
+Improve error messages when Pulp encounters a checksum or size validation error.
--- a/CHANGES/plugin_api/2244.removal
+++ b/CHANGES/plugin_api/2244.removal
@@ -1,0 +1,3 @@
+Constructor signature of `DigestValidationError` and `SizeValidationError` has changed - the
+"actual" and "expected" values are now required and "url" which was previously a positional
+argument is now a keyword argument.
--- a/pulpcore/exceptions/validation.py
+++ b/pulpcore/exceptions/validation.py
@@ -16,17 +16,24 @@ class DigestValidationError(ValidationError):
     Raised when a file fails to validate a digest checksum.
     """
 
-    def __init__(self, *args):
+    def __init__(self, actual, expected, *args, url=None, **kwargs):
         super().__init__("PLP0003")
-        if args:
-            self.url = args[0]
+        self.url = url
+        self.actual = actual
+        self.expected = expected
 
     def __str__(self):
-        if hasattr(self, "url"):
-            return _("A file located at the url {} failed validation due to checksum.").format(
-                self.url
+        if self.url:
+            msg = _(
+                "A file located at the url {url} failed validation due to checksum. "
+                "Expected '{expected}', Actual '{actual}'"
             )
-        return _("A file failed validation due to checksum.")
+            return msg.format(url=self.url, expected=self.expected, actual=self.actual)
+        else:
+            msg = _(
+                "A file failed validation due to checksum. Expected '{expected}', Actual '{actual}'"
+            )
+            return msg.format(expected=self.expected, actual=self.actual)
 
 
 class SizeValidationError(ValidationError):
@@ -34,15 +41,24 @@ class SizeValidationError(ValidationError):
     Raised when a file fails to validate a size checksum.
     """
 
-    def __init__(self, *args):
+    def __init__(self, actual, expected, *args, url=None, **kwargs):
         super().__init__("PLP0004")
-        if args:
-            self.url = args[0]
+        self.url = url
+        self.actual = actual
+        self.expected = expected
 
     def __str__(self):
-        if hasattr(self, "url"):
-            return _("A file located at the url {} failed validation due to size.").format(self.url)
-        return _("A file failed validation due to size.")
+        if self.url:
+            msg = _(
+                "A file located at the url {url} failed validation due to size. "
+                "Expected '{expected}', Actual '{actual}'"
+            )
+            return msg.format(url=self.url, expected=self.expected, actual=self.actual)
+        else:
+            msg = _(
+                "A file failed validation due to size. Expected '{expected}', Actual '{actual}'"
+            )
+            return msg.format(expected=self.expected, actual=self.actual)
 
 
 class MissingDigestValidationError(Exception):
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -330,7 +330,7 @@ class Artifact(HandleTempFilesMixin, BaseModel):
 
         if expected_size:
             if size != expected_size:
-                raise SizeValidationError()
+                raise SizeValidationError(size, expected_size)
 
         if expected_digests:
             for algorithm, expected_digest in expected_digests.items():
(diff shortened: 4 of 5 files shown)


pulpcore/exceptions/validation.py
```diff
@@ -16,33 +16,45 @@ class DigestValidationError(ValidationError):
     Raised when a file fails to validate a digest checksum.
     """
 
-    def __init__(self, *args):
+    def __init__(self, actual, expected, *args, url=None, **kwargs):
```
dralley, line 19: Since it changes the signature by making some arguments required, it is technically a breaking change.  However it seems like plugins don't create these exceptions manually, they just catch them on occasion.

We could either:

* Ignore that, on the "if a tree falls in the forest and nobody hears it" principle
* let it wait for 3.20 where we plan to make other breaking changes
* make these all kwargs instead

dralley:
As a total aside, we've rewritten the validation code in many places, I wonder if we could consolidate them a bit.

pulpcore/exceptions/validation.py
fao89: +1 to kwargs

dralley: Actually I'm not sure that helps because previously the URL was an "optional argument".  So either way it's still a "breaking" change in theory.

bmbouter: I think everyone is in agreement this is a breaking change to an object in the plugin API. If that's true, then we can't make it until 3.20. Calendar-wise that's reasonably close so I think that would be ok.

I also see the new logging as a feature not a .misc. It's net-new set of logging functionality (albeit a small and important one) for users. So given that I don't think we can backport it. I know that is inconvenient for downstreams, bu they could also patch the rpms.

What do you think?

dralley: I could see it as a plugin API feature but I can't see it as a feature in the sense that advertise it to users as one.  Especially given how rare it ought to be to see this message.

bmbouter: And should this be backportable or not? My take I guess is feature-or-not, it's a breaking change. Wdyt on that?

bmbouter: Downstream can backport whatever they want directly in the rpms; that would be fine with me.

For the upstream releases though, I believe we are not aligned on the consideration that usage or non-usage is meaningful or not w.r.t accepting plugin API breaking changes. My understanding is that non-usage isn't a consideration on when to accept breaking changes in the plugin API. We've stated publicly and to the plugin writers that the next breaking changes are going into 3.20.

dralley: I mean, my personal philosophy is a bit weaker than that w/r/t changes, but I won't dispute that we've made those claims about the API stability.

Ultimately if we want to hold it for the 3.20 cycle it wouldn't be the worst thing.  I do think downstream will probably want to pull it in before then.

dralley: @bmbouter So is that the verdict?  Should I put this in draft until 3.19 is released?

bmbouter: My opinion is that yes that would be best.

bmbouter:
nitpick: the commit message should be updated per the docs here https://docs.pulpproject.org/pulpcore/contributing/git.html#commit-message

CHANGES/2244.misc
```diff
@@ -0,0 +1 @@
+Improve error messages when Pulp encounters a checksum or size validation error.
```
bmbouter, line 1: Since this is a plugin signature change I think this file needs to call out those changes as `CHANGES/plugin_api/2244.removal`. Also with plugin writers being the audience I think calling out the specific signature changes would be good.

pulpcore/exceptions/validation.py
```diff
@@ -16,33 +16,45 @@ class DigestValidationError(ValidationError):
...
+        self.url = url
+        self.actual = actual
+        self.expected = expected
 
     def __str__(self):
-        if hasattr(self, "url"):
-            return _("A file located at the url {} failed validation due to checksum.").format(
-                self.url
+        if self.url:
+            msg = _(
```
bmbouter, line 27: For i8ln these `{}` need to be named so that translators can rearrange their order for other languages that need them reordered.

dralley: I thought we've just decided (as per the last pulpcore meeting) we're not doing this for programming errors anymore.

dralley: Should I remove the i18n in that case

ggainey: > Should I remove the i18n in that case

If it gets back to the API user, it needs I18N. If it only shows up in the logs, then not. validation-errors get back to the end-user - I'm looking at this one in a testcase right now, as it turns out :)

dralley: 9 times out of 10 this is going to bubble up to the top and cause a sync failure.

ggainey: > You can catch this, theoretically, but 9 times out of 10 this is going to bubble up to the top and cause a sync failure.
> 
> I've mentioned before that zero plugins actually do anything with this exception including import it.

Mmm, fair - just realized that where I'm seeing it, is from the traceback that's stored in the failed-sync-task.

This makes the "I18N or No" question a little more grey - since end-users can see the error-msgs/tracebacks registered for the tasks their requests create.

I guess this whole discussion says to me, that for this PR, we should keep doing what we're doing for error-logs currently. When we decide to change I18N of error-logs, we'll do all of them at once.

Just my 0.02

dralley:
@bmbouter Ready for re-review

CHANGES/plugin_api/2244.removal
```diff
@@ -0,0 +1 @@
+Constructor signature of `DigestValidationError` and `SizeValidationError` has changed - the "actual" and "expected" values are now required and "url" which was previously a positional argument is now a keyword argument.
```
bmbouter, line 1: Can this line-wrap to 100 chars.

pulpcore/exceptions/validation.py
```diff
@@ -16,33 +16,49 @@ class DigestValidationError(ValidationError):
     Raised when a file fails to validate a digest checksum.
     """
 
-    def __init__(self, *args):
+    def __init__(self, actual, expected, *args, url=None, **kwargs):
         super().__init__("PLP0003")
-        if args:
-            self.url = args[0]
+        self.url = url
```
bmbouter, line 21: What was the motivation to make url optional?

dralley: It was always optional.  If you look through the callsites sometimes there isn't a URL available.

bmbouter: Thanks and yes that makes sense.

bmbouter (approved):
Thank you, this is going to be a great improvement for our users!

ggainey (approved):

PR #2289, opened 2022-03-02
Switches unittest runner to pytest

### LINKED ISSUE #2070: As a developer, I can have pytest run the unit tests
Author: @bmbouter (bmbouter)

Redmine Issue: 9643, https://pulp.plan.io/issues/9643

---

As part of the testing effort, it would be nice to have pytest run the unittests in addition to our functional tests.

bmbouter:
This also ports the unit tests to be compatible with the django-pytest
defaults.

closes #2070


.github/workflows/scripts/script.sh
```diff
@@ -102,7 +102,7 @@ cmd_prefix bash -c "django-admin makemigrations --check --dry-run"
 
 if [[ "$TEST" != "upgrade" ]]; then
   # Run unit tests.
-  cmd_prefix bash -c "PULP_DATABASES__default__USER=postgres django-admin test --noinput /usr/local/lib/python3.8/site-packages/pulpcore/tests/unit/"
+  pytest -v -r sx --color=yes --pyargs pulpcore.tests.unit
```
fao89, line 105: I think you need the full path
```suggestion
  pytest -v -r sx --color=yes /usr/local/lib/python3.8/site-packages/pulpcore/tests/unit/
```

bmbouter: It runs locally for me with this. I think the issue is pytest-django isn't being installed. I'll push a test change now to see.

fao89: Locally you pip install -e, and I believe this why you need the full path here

bmbouter: I think you're right, but I think it's because it's not installed at all locally, it gets installed inside the container!

How do I form that path from plugin template vars?

fao89: now I see what is the issue, we run unittests inside the container, that's why we only install `unittest_requirements.txt`inside the container.

You can do:
- cmd_prefix bash -c "pytest -v -r sx --color=yes --pyargs pulpcore.tests.unit"

or you can replace functest_requirements.txt with test_requirements.txt
https://github.com/pulp/pulpcore/blob/3479de46ca72b0f484c5d4bd5b0b7e9eee4d69ee/.github/workflows/scripts/install.sh#L23

bmbouter: So is pulpcore not installed outside the container? I figured it had to be because this is run outside the container later:  https://github.com/pulp/pulpcore/blob/main/.github/workflows/scripts/script.sh#L187

bmbouter: Man I have tried so many variations to get this working and each approach has problems. Ugh.

I tried the `cmd_prefix bash -c "pytest -v -r sx --color=yes --pyargs pulpcore.tests.unit"` option with commit 32b93bf84b2af91b1d61db6cfb20e18f8abb4724 and now it's needing the bindings also to be installed in the container, which I guess we can do but installing everything and everything out is getting tough. So I'm thinking we install all the pytest stuff outside the container.

When I try to install all the pytest things outside the container, there are various failures. I worked through those, but the "outside the container" installation doesn't have the necessary files the settings.py would need....

bmbouter:
I'm closing my PR in favor of this PR from @dkliban https://github.com/dkliban/pulpcore/pull/55/

PR #2295, opened 2022-03-02
Fix new-task-system migration to clean up task and taskreservedresource.

### LINKED ISSUE #2101: Clean up TaskReservedResources/task-table at migration to new-tasking-system
See https://bugzilla.redhat.com/show_bug.cgi?id=2031154 for details.

Migration that needs to be updated to purge taskreservedresource entries: 0064_add_new_style_task_columns.py

This wants to be cherrypicked into 3.14/15/16 (after which the offending table no longer exists)

ggainey:
fixes #2101
[nocoverage]

### DIFF
--- a/pulpcore/app/migrations/0064_add_new_style_task_columns.py
+++ b/pulpcore/app/migrations/0064_add_new_style_task_columns.py
@@ -4,16 +4,34 @@ import django.contrib.postgres.fields
 import django.contrib.postgres.fields.jsonb
 from django.db import migrations, models
 
+TASK_BATCH_SIZE = 1000
+
 
 def copy_reserved_resources_record(apps, schema_editor):
     Task = apps.get_model('core', 'Task')
-    for task in Task.objects.iterator():
+
+    # Update _reserved_resource_record for all tasks, 1000 tasks at a time.
+    # When we hit 1K tasks, go to the db for the batch.
+    # Make sure to update the final batch!
+    tasks = []
+    for task in Task.objects.iterator(chunk_size=TASK_BATCH_SIZE):
         task._reserved_resources_record = list(task.reserved_resources_record.values_list('resource', flat=True))
-        task.save()
+        tasks.append(task)
+        if len(tasks) == TASK_BATCH_SIZE:
+            Task.objects.bulk_update(tasks, ["_reserved_resources_record"])
+            tasks.clear()
+
+    # Update last set of tasks
+    if len(tasks) > 0:
+        Task.objects.bulk_update(tasks, ["_reserved_resources_record"])
 
 
-def noop(apps, schema_editor):
-    pass
+def purge_reservedresources(apps, schema_editor):
+    TaskReservedResource = apps.get_model('core', 'TaskReservedResource')
+    TaskReservedResource.objects.all().delete()
+
+    ReservedResource = apps.get_model('core', 'ReservedResource')
+    ReservedResource.objects.all().delete()
 
 
 class Migration(migrations.Migration):
@@ -23,6 +41,12 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
+        # Purge any ReservedResource entries - if there are any, they're orphans
+        migrations.RunPython(
+            code=purge_reservedresources,
+            reverse_code=migrations.RunPython.noop,
+        ),
+        # Update entities for the new task-system
         migrations.AddField(
             model_name='task',
             name='args',
@@ -59,7 +83,7 @@ class Migration(migrations.Migration):
         ),
         migrations.RunPython(
             code=copy_reserved_resources_record,
-            reverse_code=noop,
+            reverse_code=migrations.RunPython.noop,
         ),
         migrations.RemoveField(
             model_name='taskreservedresourcerecord',
@@ -80,4 +104,5 @@ class Migration(migrations.Migration):
             old_name='_reserved_resources_record',
             new_name='reserved_resources_record',
         ),
+
     ]
--- a/CHANGES/2101.bugfix
+++ b/CHANGES/2101.bugfix
@@ -1,0 +1,5 @@
+Fixed migration 0064_add_new_style_task_columns to purge extraneous ReservedResource and
+TaskReservedResource entries, which could block sync and publish tasks post-upgrade.
+
+Also taught the migration to bulk-update the Task changes. In large installations, this
+should have a positive impact on the time it takes to apply the migration.


ggainey:
This is a WIP while I try to coerce a dev-env to start at core/3.7, with entries in core_task and core_taskreservedresource, and then upgrade to core/3.14 with this change. Early eyes are appreciated, however.

pulpcore/app/migrations/0064_add_new_style_task_columns.py
```diff
@@ -12,6 +19,19 @@ def copy_reserved_resources_record(apps, schema_editor):
         task.save()
```
mdellweg, line 19: If i didn't write this migration long ago, I would have used `batch_update` here. XD

pulpcore/app/migrations/0064_add_new_style_task_columns.py
```diff
@@ -12,6 +19,19 @@ def copy_reserved_resources_record(apps, schema_editor):
...
+    TaskReservedResource = apps.get_model('core', 'TaskReservedResource')
+    TaskReservedResource.objects.all().delete()
+
+    ReservedResource = apps.get_model('core', 'ReservedResource')
+    ReservedResource.objects.all().delete()
+
+
+def purge_completed_tasks(apps, schema_editor):
+    Task = apps.get_model('core', 'Task')
+    Task.objects.filter(state__in=TASK_FINAL_STATES).delete()
```
mdellweg, line 32: Should we keep tasks that are some kind of recent?

pulpcore/app/migrations/0064_add_new_style_task_columns.py
```diff
@@ -12,6 +19,19 @@ def copy_reserved_resources_record(apps, schema_editor):
...
+
+
 def noop(apps, schema_editor):
```
mdellweg, line 35: I know now: There is a `migrations.RunPython.noop`...

pulpcore/app/migrations/0064_add_new_style_task_columns.py
```diff
@@ -80,4 +100,13 @@ class Migration(migrations.Migration):
             old_name='_reserved_resources_record',
             new_name='reserved_resources_record',
         ),
+        migrations.RunPython(
+            code=purge_reservedresources,
+            reverse_code=noop,
+        ),
+        migrations.RunPython(
```
mdellweg, line 107: If you really want to gain something from this operation, you should do it before the `copy_reserved_resource_record` (the one that happened to take >= 1h on loaded systems).

pulpcore/app/migrations/0064_add_new_style_task_columns.py
ggainey: Good question. In this instance, it's only an issue if you were running pre-3.14 (old task system), and are now upgrading to 3.14 (or later) and the new tasking system. I don't have a strong opinion one way or another. Maybe "complete and older than 10 days"?

pulpcore/app/migrations/0064_add_new_style_task_columns.py
ggainey: Yes - I actually moved these ahead of the Alter... stmts, just haven't pushed the change yet. GMTA! :)

pulpcore/app/migrations/0064_add_new_style_task_columns.py
ggainey: Of course, if we keep most-recent, then in a very common usecase (core/3.7 used to do 2to3 migration followed immediately by upgrade to core/3.14), we won't have much impact on the number of tasks - since many/most of them will have been created "recently". Thoughts?

mdellweg: Right. The reason for this is to keep the migration time low. So a number of tasks to keep would do that but that does not sound like the thing we should do unless that number is zero.
Let's keep it that way.

dralley (approved):

pulpcore/app/migrations/0064_add_new_style_task_columns.py
```diff
@@ -23,6 +43,17 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
+        # Purge completed tasks prior to trying to update the Task model
+        migrations.RunPython(
+            code=purge_completed_tasks,
+            reverse_code=noop,
```
mdellweg, line 49:
```suggestion
            reverse_code=migrations.RunPython.noop,
```

pulpcore/app/migrations/0064_add_new_style_task_columns.py
mdellweg: If the mass deletion of tasks is not working, this is really the data migration where we can gain a lot of performance by performing the update in batches.

ggainey:
@mdellweg @dralley Rather than delete task-entries, I've moved to bulk_update to try and address the performance issue. More eyes again, please.

pulpcore/app/migrations/0064_add_new_style_task_columns.py
```diff
@@ -4,12 +4,30 @@
...
+    "canceled",
+]
+
 
 def copy_reserved_resources_record(apps, schema_editor):
     Task = apps.get_model('core', 'Task')
-    for task in Task.objects.iterator():
+    tasks_qs = Task.objects.all();
+
+    for task in tasks_qs:
```
dralley, line 19: Set `.only("reserved_resources_record")` on the `tasks_qs` so you're only extracting the data you want to work with.  

And I think if you want to just use the queryset directly you can use `.update()` instead of `.bulk_update()`.

If the list of tasks is too massive you might want to use `.iterator()` and stick with `.bulk_update()` but do the batching manually.  I'm not sure how big it is, it would probably have to be >100k for that to start making sense.

ggainey: "reserved_resources_record" is a ManyToMany, not a field, and can't be used for only(). Tasks are also pretty small, so I don't think this saves us much.

My understanding of update() vs bulk_update(), is that update() is going to do one db-update per entity, where bulk_update() is going to do one db call per-batch.

dralley (approved):

ggainey:
I'm un-drafting this - still needs more eyes, tho

pulpcore/app/migrations/0064_add_new_style_task_columns.py
```diff
@@ -4,12 +4,30 @@
...
 def copy_reserved_resources_record(apps, schema_editor):
     Task = apps.get_model('core', 'Task')
-    for task in Task.objects.iterator():
+    tasks_qs = Task.objects.all()
+
+    for task in tasks_qs:
         task._reserved_resources_record = list(task.reserved_resources_record.values_list('resource', flat=True))
-        task.save()
+
+    Task.objects.bulk_update(tasks_qs, ["_reserved_resources_record"], 1000)
```
mdellweg, line 22: I don't think, that is, how it works. You would need to collect the updated task objects from line 20 and pass those to bulk_update. Also i would do the batching in python to have a cap on memory consumption.
See example here: https://github.com/pulp/pulpcore/blob/main/pulpcore/app/migrations/0086_task_json_fields.py#L26

mdellweg (changes requested):

pulpcore/app/migrations/0064_add_new_style_task_columns.py
dralley: I don't think that's correct, the difference is that `update()` does its work entirely in the database so there's less marshalling, but it's also less flexible because of that, you can't update using complex python logic like you can with `.bulk_update()`

The docs actually mention that `.bulk_update()` uses `.update()` internally.

https://docs.djangoproject.com/en/4.0/ref/models/querysets/#bulk-update

pulpcore/app/migrations/0064_add_new_style_task_columns.py
ggainey: You are correct, current version does the deed and updated 1K tasks in ~4s for the entire migrate-invocation.

pulpcore/app/migrations/0064_add_new_style_task_columns.py
ggainey: :+1:

pulpcore/app/migrations/0064_add_new_style_task_columns.py
ggainey: Changed to use batch-update

pulpcore/app/migrations/0064_add_new_style_task_columns.py
ggainey: batch-update addresses the performance-issue, and means we don't have to "guess" how many tasks to leave around.

dralley (approved):

mdellweg (approved):
I still don't get, what kind of data migration you can mix with schema migrations, and where it breaks, but future me will try to always separate data from schema migrations.

Thanks for working on this!

PR #2313, opened 2022-03-09
Correctly expose the serializer for RepositoryVersion repair

gerrod3:
[noissue]

This also fixes the bindings asking for the wrong arguement.

### DIFF
--- a/CHANGES/2330.bugfix
+++ b/CHANGES/2330.bugfix
@@ -1,0 +1,1 @@
+Fixed improper fields being listed in ``RepositoryVersion`` repair API.
--- a/pulpcore/app/viewsets/repository.py
+++ b/pulpcore/app/viewsets/repository.py
@@ -184,7 +184,7 @@ class RepositoryVersionViewSet(
         description="Trigger an asynchronous task to repair a repository version.",
         responses={202: AsyncOperationResponseSerializer},
     )
-    @action(detail=True, methods=["post"])
+    @action(detail=True, methods=["post"], serializer_class=RepairSerializer)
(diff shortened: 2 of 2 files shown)


dralley:
@gerrod3  Has this been encountered in a released version?  Should an issue be filed?

gerrod3:
> @gerrod3 Has this been encountered in a released version? Should an issue be filed?

This clarity issue has been here for a while, but it seems to have no impact on the functionality so I'm unsure if it needs an issue. Even if you supply the incorrect param `base_version` that the API docs say to provide nothing happens as the real param `verify_checksums` has a default value.

The bigger issue is that it makes the bindings for calling repair confusing and I was having a hard time figuring out why it was wrong and we haven't noticed since we don't test repair with the bindings currently. (https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/api/using_plugin/test_repair.py)

dralley (approved):

dralley:
Only comment, Katello probably uses the bindings for repair, so if actually *changes* the bindings we might want to be thoughtful about how we merge.

mdellweg:
Azure tests seem to be broken today.
And the S3 is a timing issue with another test, supposed to be fixed by https://github.com/pulp/pulpcore/pull/2265/files#diff-ab70ec0b308fb8c0d3b40add15f611d0e4ea32400db7b25db2631c8c70047fb8R115

bmbouter (changes requested):
I think this needs a changelog change.

bmbouter (approved):
Thank you!

PR #2321, opened 2022-03-10
Taught /purge/ API to purge in batches, to restrict memory-use.

### LINKED ISSUE #2215: Task purging endpoint loads every task in memory
**Version**
`pulpcore-3.17.3`
`pulp_rpm-3.17.3`
`pulp_ansible-0.12.0`
`pulp_container-2.10.0`
`pulp_deb-2.17.0`
`pulp_file-1.10.1`
`pulp_python-3.6.0`

**Describe the bug**
We have a large number of finished pulp tasks (around 1.5kk). So we're trying to purge it with [`pulp/api/v3/tasks/purge`](https://docs.pulpproject.org/pulpcore/restapi.html#operation/tasks_purge) and quickly realized, that this task is trying to load whole finished queue (all 1.5kk finished tasks) into memory, which cause it killed after consuming 15GB of memory.

**To Reproduce**
1. Create large number of pulp-tasks (more than 1kk will be enough)
2. Run [`pulp/api/v3/tasks/purge`](https://docs.pulpproject.org/pulpcore/restapi.html#operation/tasks_purge) with `finished_before` set to future date or something.
3. Your server will run out-of-memory and task won't be completed.

**Expected behavior**
Task purging api-endpoint does not load the entire finished queue into memory

ggainey:
fixes #2215
[nocoverage]


pulpcore/app/tasks/purge.py
```diff
@@ -5,6 +5,44 @@
...
+DELETE_LIMIT = 1000
+# Key that delete() returns for Tasks
+TASK_KEY = "core.Task"
+
+
+def _details_reporting(current_reports, current_details):
+    """
+    Create and update progress-reports for each detail-key returned from a delete() call.
+
+    The function has one secpial case: we know how many Tasks we're execpting to delete right
```
dralley, line 21: Typo "secpial"

ggainey: And execpting!

CHANGES/2215.bugfix
```diff
@@ -0,0 +1,3 @@
+Taught task-purge to process tasks in batches of 1000. This prevents large purges from using
+ large amounts of memory as a result of reading all the affected Taskss into memory at once.
```
dralley, line 2: s/Taskss/Tasks

pulpcore/app/tasks/purge.py
```diff
@@ -5,6 +5,47 @@
...
+    total for that key.
+
+    Args:
+        current_reports (dict): key:ProgressReport to record into
+    Returns:
+        updated current_reports
+    """
+    for key in current_details:
+        if key in current_reports:
+            if key != "core.Task":
```
mdellweg, line 36:
```suggestion
            if key != TASK_KEY:
```

pulpcore/app/tasks/purge.py
```diff
@@ -26,25 +67,49 @@ def purge(finished_before, states):
...
+        code="purge.tasks.key.{}".format(TASK_KEY),
+        done=0,
+    )
+    pb.save()
+    details_reports[TASK_KEY] = pb
+
+    # Our delete-query is going to deal with "the first DELETE_LIMIT tasks that match our
+    # criteria", looping until we've deleted everything that fits our parameters
+    qs = Task.objects.filter(finished_at__lt=finished_before, state__in=states)[:DELETE_LIMIT]
     units_deleted, details = get_objects_for_user(current_user, "core.delete_task", qs=qs).delete()
```
mdellweg, line 91: This is the exact same queryset you used for the count call before. What if we used that as a variable (querysets can be reused):
`qs = get_objects_for_user(
        current_user,
        "core.delete_task",
        qs=Task.objects.filter(finished_at__lt=finished_before, state__in=states))`

Then we can use `qs.count()` for the `total`, and `qs[:DELETE_LIMIT].delete()` for the actual process. I think the readability would benefit.

pulpcore/app/tasks/purge.py
```diff
@@ -26,25 +67,49 @@ def purge(finished_before, states):
...
+    total_deleted = 0
+    for key in details_reports:
+        pb = details_reports[key]
+        total_deleted += pb.total
+        pb.state = TASK_STATES.COMPLETED
+        pb.save()
 
     # Progress bar reporting total-units
-    progress_bar = ProgressReport(
+    totals_pb = ProgressReport(
```
mdellweg, line 108: Can we count the totals on the way too?

mdellweg (commented):
A lot of suggestions again...

CHANGES/2215.bugfix
```diff
@@ -0,0 +1,3 @@
+Taught task-purge to process tasks in batches of 1000. This prevents large purges from using
```
ipanova, line 1: did you happen to look at how we deal with orphan cleanup? there is queryset iterator maybe worth using? https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/orphan.py#L13

pulpcore/app/tasks/purge.py
```diff
@@ -5,6 +5,47 @@
...
+
+    Args:
+        current_reports (dict): key:ProgressReport to record into
+    Returns:
+        updated current_reports
+    """
+    for key in current_details:
+        if key in current_reports:
+            if key != "core.Task":
+                current_reports[key].total += current_details[key]
```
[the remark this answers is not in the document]
ggainey: When done>total we get a warning from increase_by(), changing to just do the math-and-save manually.

pulpcore/app/tasks/purge.py
ggainey: qs[:delete_LIMIT].delete() does not, alas, work - django happily informs one "Cannot use 'limit' or 'offset' with delete." Made some changes to make it more readable, however.

pulpcore/app/tasks/purge.py
ggainey: yes, done

CHANGES/2215.bugfix
ggainey: I hadn't seen that, thanks!  I haven't updated to try taking advantage of its existence - the approach I have here "feels" straightforward enough, and using a method from Someone Else's Task feels...odd. I'll experiment w/ it, tho; if it makes the code cleaner, may do one more change to use it.

pulpcore/app/tasks/purge.py
mdellweg: ??? If `get_objects_for_user` is not stripping the `limit` part of the query, you would be doing exactly that with the old code. I must be blind for something here.

pulpcore/app/tasks/purge.py
```diff
@@ -5,6 +5,48 @@
...
+    with a ProgressReport whose total is correct, in advance, and therefore don't update
+    total for that key.
+
+    Args:
+        current_reports (dict): key:ProgressReport to record into
+    Returns:
+        updated current_reports
+    """
+    for key in current_details:
+        totals_pb.done += current_details[key]
```
mdellweg, line 35:
```suggestion
        totals_pb.done += current_details[key]
        totals_pb.save()
```

pulpcore/app/tasks/purge.py
```diff
@@ -5,6 +5,48 @@
...
+        if key in current_reports:
+            # Avoid increase_by() here to prevent extraneous "done>total" warning in logs
+            current_reports[key].done += current_details[key]
+            current_reports[key].save()
+        else:
+            pb = ProgressReport(
+                message=_("Purged task-objects of type {}".format(key)),
+                total=current_details[key],
```
mdellweg, line 43: I think, we should leave that alone. (That should also avoid your `increase_by` warning.)

https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/progress.py#L236

pulpcore/app/tasks/purge.py
```diff
@@ -26,25 +68,49 @@ def purge(finished_before, states):
...
+    # Tasks, prior to the specified date, in the specified state, owned by the current-user
+    tasks_qs = Task.objects.filter(finished_at__lt=finished_before, state__in=states)
+    candidate_qs = get_objects_for_user(current_user, "core.delete_task", qs=tasks_qs)
+    delete_qs = get_objects_for_user(current_user, "core.delete_task", qs=tasks_qs[:DELETE_LIMIT])
     # Progress bar reporting total-units
-    progress_bar = ProgressReport(
+    totals_pb = ProgressReport(
         message=_("Purged task-objects total"),
-        total=units_deleted,
+        total=0,
```
mdellweg, line 78:
```suggestion
        total=None,
```

mdellweg (commented):
A few tweaks around the progress reports.
Also i do not understand how we limit the amount of deleted units per query if sql does not allow that. Is `get_objects_for_user` sneakily putting that in some kind of `DELETE FROM Task WHERE pk IN SELECT * LIMIT 100` statement?

pulpcore/app/tasks/purge.py
dralley: +1

pulpcore/app/tasks/purge.py
dralley: +1

mdellweg: or actually `increase_by` when we do not set the `totals_pb.total`.

pulpcore/app/tasks/purge.py
```diff
@@ -26,25 +68,49 @@ def purge(finished_before, states):
...
         code="purge.tasks.total",
-        done=units_deleted,
-        state="completed",
+        done=0,
+        state=TASK_STATES.COMPLETED,
+    )
+    totals_pb.save()
```
dralley, line 83: Note that you're later modifying this after it's already set to `completed` state

pulpcore/app/tasks/purge.py
```diff
@@ -5,6 +5,46 @@
...
+    Args:
+        current_reports (dict): key:ProgressReport to record into
+    Returns:
+        updated current_reports
+    """
+    for key in current_details:
+        totals_pb.done += current_details[key]
+        if key in current_reports:
+            current_reports[key].increase_by(current_details[key])
+            current_reports[key].save()
```
mdellweg, line 37: I think, the `increase_by` handles the save for us.
```suggestion
```

pulpcore/app/tasks/purge.py
```diff
@@ -5,6 +5,46 @@
...
+    with a ProgressReport whose total is correct, in advance, and therefore don't update
+    total for that key.
+
+    Args:
+        current_reports (dict): key:ProgressReport to record into
+    Returns:
+        updated current_reports
+    """
+    for key in current_details:
+        totals_pb.done += current_details[key]
```
mdellweg, line 34:
```suggestion
        totals_pb.increase_by(current_details[key])
```

pulpcore/app/tasks/purge.py
dralley: It does

pulpcore/app/tasks/purge.py
```diff
@@ -26,25 +66,49 @@ def purge(finished_before, states):
...
+    # Our delete-query is going to deal with "the first DELETE_LIMIT tasks that match our
+    # criteria", looping until we've deleted everything that fits our parameters
+    units_deleted, details = delete_qs.delete()
+    # Until our query returns "No tasks deleted", add results into totals and Do It Again
+    while units_deleted > 0:
+        _details_reporting(details_reports, details, totals_pb)
+        units_deleted, details = delete_qs.delete()
+
+    # Complete the progress-reports for the specific entities deleted
+    for key in details_reports:
```
dralley, line 105: I generally prefer `for key, pb in details_reports.items()` instead of separate iteration and lookup

ggainey:
> A few tweaks around the progress reports. Also i do not understand how we limit the amount of deleted units per query if sql does not allow that. Is `get_objects_for_user` sneakily putting that in some kind of `DELETE FROM Task WHERE pk IN SELECT * LIMIT 100` statement?

Close - `tasks_qs[:DELETE_LIMIT]` generates a set of DELETE_LIMIT UUIDs, which results in an emitted `DELETE FROM core_task WHERE pulp_id IN (...insert 1K UUIDs here)`

CHANGES/2215.bugfix
ggainey: Decided not to go this route, for this PR - either I had to copy a wadge of code (which was distracting, from what purge is doing), or had to teach purge to call into the orphan task (which felt....Wrong), or refactor orphan and this PR to find the QS iterator "somewhere else", which felt like a lot of change for not (enough?) payback.

dralley (approved):

mdellweg (approved):

mdellweg:
@dralley Task purging was introduced in 3.17.0. I'm going to remove the 3.16 backport label.

PR #2331, opened 2022-03-11
Adds TASK_DIAGNOSTIC setting and docs

### LINKED ISSUE #2329: As a user or developer, I can enable the memory recording of my tasks
This feature will be enabled by a setting named `RECORD_TASK_METRICS` which will default to `False`. When enabled, the directory `/var/lib/pulp/task_metrics/` will have files added to it.

The memory usage of the each task should be written periodically to the file `/var/lib/pulp/task_metrics/80/54bdac-1cfe-4c1b-91c2-1139fc316aeb.datum` where the UUID is split at the first two chars and the filename is the remaining portion of the UUID with `.datum`. on the end of it.

The memory should be recorded periodically.

bmbouter:
closes #2329

### DIFF
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -392,3 +392,14 @@ ORPHAN_PROTECTION_TIME
     they become candidates for deletion by an orphan cleanup task. This should ideally be longer
     than your longest running task otherwise any content created during that task could be cleaned
     up before the task finishes. Default is 1440 minutes (24 hours).
+
+
+.. _task_diagnostics:
+
+TASK_DIAGNOSTICS
+^^^^^^^^^^^^^^^^
+
+    If ``True``, each task will record various diagnostics (listed below) to files in the dir
+    ``/var/tmp/pulp/<task_UUID>/``. This is ``False`` by default.
+
+      * memory - the task's max resident set size in MB.
--- a/docs/plugins/plugin-writer/concepts/tasks/diagnostics.rst
+++ b/docs/plugins/plugin-writer/concepts/tasks/diagnostics.rst
@@ -1,0 +1,25 @@
+Diagnostics
+===========
+
+When users enable task diagnostics using the ``TASK_DIAGNOSTICS`` setting, all tasks will write out
+diagnostic information to data files in ``/var/tmp/pulp/<task_UUID>/`` directory.
+
+Memory Analysis
+---------------
+
+The memory is written to a file such as
+``var/tmp/pulp/3367e577-4b09-44b6-9069-4a06c367776a/memory.datum``.
+
+You can plot this with gnuplot by changing into the directory with the files you want to see then:
+
+1) Enter the ``gnuplot`` interactive environment.
+
+2) Paste these commands::
+
+    set terminal png size 1200,900 enhanced font "Arial, 10"
+    set output "memory.png"
+    set ylabel "Task Process Megabytes (MB)"
+    set xlabel "Seconds since task start"
+    plot "memory.datum" with lines
+
+3) Open your png chart saved at memory.png
--- a/pulpcore/tasking/pulpcore_worker.py
+++ b/pulpcore/tasking/pulpcore_worker.py
@@ -5,10 +5,13 @@ import importlib
 import logging
 import os
 import random
+import resource
 import select
 import signal
 import socket
 import sys
+import threading
+import time
 import traceback
 from contextlib import suppress
 from datetime import timedelta
@@ -37,6 +40,7 @@ from pulpcore.app.role_util import (  # noqa: E402: module level not at top of f
 from pulpcore.constants import (  # noqa: E402: module level not at top of file
     TASK_STATES,
     TASK_INCOMPLETE_STATES,
+    VAR_TMP_PULP,
 )
 
 from pulpcore.exceptions import AdvisoryLockError  # noqa: E402: module level not at top of file
@@ -377,6 +381,23 @@ class NewPulpWorker:
             self.shutdown()
 
 
+def write_memory_usage(task_pk):
+    taskdata_dir = VAR_TMP_PULP / str(task_pk)
+    taskdata_dir.mkdir(parents=True, exist_ok=True)
+    memory_file_dir = taskdata_dir / "memory.datum"
+    _logger.info("Writing task memory data to {}".format(memory_file_dir))
+
+    with open(memory_file_dir, "w") as file:
+        file.write("# Seconds\tMemory in MB\n")
+        seconds = 0
+        while True:
+            current_mb_in_use = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1024
+            file.write(f"{seconds}\t{current_mb_in_use:.2f}\n")
+            file.flush()
+            time.sleep(5)
+            seconds += 5
+
+
 def child_signal_handler(sig, frame):
     # Reset signal handlers to default
     # If you kill the process a second time it's not graceful anymore.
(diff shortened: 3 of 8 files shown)


bmbouter:
It produces image files like this one:

memory

bmbouter:
@pulp/core FYI here's one of the next-steps that came out of the memory PoC.

pulpcore/plugin/stages/api.py
```diff
@@ -179,6 +183,21 @@ def __str__(self):
         return "[{id}] {name}".format(id=id(self), name=self.__class__.__name__)
 
 
+async def write_memory_usage():
+    current_task = await sync_to_async(Task.current)()
+    taskdata_dir = Path("/var/tmp/pulp") / str(current_task.pulp_id)
```
dralley, line 188: We should make `/var/tmp/pulp/` into a constant somewhere.

I'm not sure we should put it here, it kind of makes sense that this should be enabled for *all* tasks and not just sync tasks yes?  It would be cleaner to make it part of the heartbeat.

docs/configuration/settings.rst
```diff
@@ -392,3 +392,14 @@ ORPHAN_PROTECTION_TIME
...
     than your longest running task otherwise any content created during that task could be cleaned
     up before the task finishes. Default is 1440 minutes (24 hours).
+
+
+.. _task_diagnostics:
+
+TASK_DIAGNOSTICS
+^^^^^^^^^^^^^^^^
+
+    If ``True``, have each task will record various diagnostics (listed below) to files in the dir
```
mdellweg, line 402:
```suggestion
    If ``True``, each task will record various diagnostics (listed below) to files in the dir
```

docs/plugins/plugin-writer/concepts/tasks/diagnostics.rst
```diff
@@ -0,0 +1,23 @@
...
+
+You can plot this with gnuplot by changing into the directory with the files you want to see then:
+
+1) Enter the ``gnuplot`` interactive environment.
+
+2) Paste these commands::
+
+    set terminal png size 1200,900 enhanced font "Arial, 10"
+    set output 'memory.png'
+    plot "memory.datum"
```
mdellweg, line 21: Can we add names and units to the axes here?

docs/plugins/plugin-writer/concepts/tasks/diagnostics.rst
```diff
@@ -0,0 +1,23 @@
...
+
+3) Open your png chart saved at memory.datum
```
mdellweg, line 23:
```suggestion
3) Open your png chart saved at memory.png .
```
If you don't "set output", you might end up with an interactive plot window, even. Why do you think we should do?

pulpcore/plugin/stages/api.py
bmbouter: I'm moving it to a constant. Also yes I do think this should be for all tasks. I'm going to rework it. Yes I think it will need to be part of the worker machinery. I'll look around and revise it. Suggestions are welcome.

docs/plugins/plugin-writer/concepts/tasks/diagnostics.rst
bmbouter: I'll try it without the `set output`. I know very little about gnuplot lol. This work was inspired by your herminig plots.

bmbouter: It errored on my dev environment which makes sense because it does not have a window manager env. I'll leave it in for now.

docs/plugins/plugin-writer/concepts/tasks/diagnostics.rst
bmbouter: done.

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -385,6 +403,9 @@ def _perform_task(task_pk, task_working_dir_rel_path):
     signal.signal(signal.SIGINT, child_signal_handler)
     signal.signal(signal.SIGTERM, child_signal_handler)
     signal.signal(signal.SIGUSR1, child_signal_handler)
+    if settings.TASK_DIAGNOSTICS:
```
bmbouter, line 406: This runs in the child process and the memory reporting includes all threads.

docs/plugins/plugin-writer/concepts/tasks/diagnostics.rst
mdellweg: That's actually the big question: Would you pull the data or the image processed from the data off of your server.
In production, i would probably not have gnuplot installed on the box. Then again, you may not want to do such a thing on a production box...
In the end the data to some file processing the most flexible solution.

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -368,6 +372,20 @@ def run_forever(self):
...
+def write_memory_usage(task_pk):
+    taskdata_dir = VAR_TMP_PULP / str(task_pk)
+    taskdata_dir.mkdir(parents=True, exist_ok=True)
+    memory_file_dir = taskdata_dir / "memory.datum"
+    _logger.info("Writing task memory data to {}".format(memory_file_dir))
+
+    with open(memory_file_dir, "w") as file:
+        while True:
+            current_mb_in_use = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1024
+            file.write(f"{current_mb_in_use:.2f}\t\n")
```
mdellweg, line 384: Can we write two columns with timestamp and memory use?

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -385,6 +403,9 @@ def _perform_task(task_pk, task_working_dir_rel_path):
     signal.signal(signal.SIGINT, child_signal_handler)
     signal.signal(signal.SIGTERM, child_signal_handler)
     signal.signal(signal.SIGUSR1, child_signal_handler)
+    if settings.TASK_DIAGNOSTICS:
+        normal_thread = threading.Thread(target=write_memory_usage, args=(task_pk,), daemon=True)
+        normal_thread.start()
```
mdellweg, line 408: Can we signal that thread after the work is done to write one last record and quit?
So we get at least 2 records from tasks running in less than 5 sec.

Also i was wondering whether we could monitor the memory from the other process in `supervise_task`. But at least for more detail information than just memory we probably need to be _in_ the child task.

bmbouter:
With the most recent revisions it now has a data format like this:

```
# Seconds       Memory in MB
0       102.22
5       133.05
10      133.82
15      135.88
20      136.39
25      142.64
30      142.90
35      143.16
40      143.16
```

And the charting now looks like this:
memory

pulpcore/tasking/pulpcore_worker.py
mdellweg: Including the one doing the reporting. Very quantum this ...

docs/plugins/plugin-writer/concepts/tasks/diagnostics.rst
mdellweg: The suggested change is still pending, thou.

Shouldn't ask two things in one email.

mdellweg (changes requested):

pulpcore/tasking/pulpcore_worker.py
bmbouter: In this case the error I don't think is significant enough to make spending additional time on it a  good use of time relative to the many other things we are supposed to do.

bmbouter: I hear you though.

dralley: I can live with that, but lets go ahead and  at least

A) file an issue for that limitation
B) leave a TODO in the code mentioning it

It would suck if this bit someone and it wasn't immediately obvious from the code why that was the case

dralley (approved):
Approve in principle, although I do strongly prefer we closed the hole which @mdellweg mentioned.  See comment about marking this issue at miminum

bmbouter:
I filed this to capture the concern https://github.com/pulp/pulpcore/issues/2337

I'll push the TODO now

pulpcore/tasking/pulpcore_worker.py
mdellweg: The other thing that makes me a bit wary is that we introduce another notion of concurrency here. (We already use subprocess for the task itself, which may use the async loop and at the same time we have adhoc asynchronicity non-blockability in the main pulp_worker process. `supervise_task` already returns regular-ish to heartbeat, it is probably easy to add some subtask diagnostic there too.

mdellweg: I've seen you opened an issue about that: https://github.com/pulp/pulpcore/issues/2337

mdellweg (approved):

dralley:
@bmbouter I'm merging this to be able to use it in testing your other PR, hope you don't mind!

Issue #2069, opened 2022-01-17
Reduce memory usage of the pipeline


fao89:
Author: @bmbouter (bmbouter)

Redmine Issue: 9635, https://pulp.plan.io/issues/9635

---

## Motivation

It would be nice if users could specify a desired maximum amount of RAM to be used during sync. For example, a user can say I only want 1500 MB of RAM to be used max.

## What is already in place

The stages pipeline restricts memory usage by only allowing 1000 declarative content objects between each stage (so for 8-9 stages that's 8000-9000 declarative content objects. This happens [here](https://github.com/pulp/pulpcore/blob/main/pulpcore/plugin/stages/api.py#L217).

Interestingly the docstring says this defaults to 100, but it seems to actually be 1000!

Also the stages perform batching, so they will only taking in a limited number of items (the batch size). That happens [with minsize](https://github.com/pulp/pulpcore/blob/main/pulpcore/plugin/stages/api.py#L84).

## Why this isn't enough

These are count-based mechnisms and don't correspond to actual MB or GB of memory used. Some content units vary a lot in how much memory each DeclarativeContent objects take up.

Another lesser problem is that it doesn't help plugin writers restrict their usage of memory in FirstStage.

## Idea

Add a new param called `max_mb` to base Remote, which defaults to None. If specified, the user will be specifying the desired maximum MB used by process syncing.

Have the queues between the stages, and the bather implementation, both check the total memory the current process is using and asyncio.sleep() polling until it goes down. This should keep the maximum amount used by all objects roughly to that number.

## Details

Introduce a new `MBSizeQueue` which is a wrapper around `asyncio.Queue` used today. It will have the same `put()` call, only wait if the amount of memory in use is greater than the remote is configured for.

Then introduce the same memory checking feature in the batcher. I'm not completely sure this second part is needed though.

We have to be very careful not to deadlock with this feature. For example, we have to account for the base case where even a single item is larger than the memory desired. Repos in pulp_rpm have had a single unit use more than 1.2G if I remember right, so if someone was syncing with 800 MB and we weren't careful to allow that unit to still flow through the pipeline we'd deadlock.....

fao89:
From: @dralley (dalley)
Date: 2021-12-13T16:54:22Z

---

> Repos in pulp_rpm have had a single unit use more than 1.2G if I remember right, so if someone was syncing with 800 MB and we weren't careful to allow that unit to still flow through the pipeline we'd deadlock.....

This is true but:

[...]

PR #2338, opened 2022-03-15
Reduces queue size between stages from 1000 to 1

bmbouter:
The runtime was actually slightly slower with the 1000 objects in-memory
and much more memory use.

closes #2069

### DIFF
--- a/CHANGES/2069.bugfix
+++ b/CHANGES/2069.bugfix
@@ -1,0 +1,1 @@
+Reduced memory usage during tasks like sync by holding fewer objects in-memory unnecessarily.
--- a/CHANGES/plugin_api/2069.bugfix
+++ b/CHANGES/plugin_api/2069.bugfix
@@ -1,0 +1,3 @@
+Adjusted the default size of the queues between pipelines to be 1 instead of 1000. The batchers in
+the stage will still accumulate up to 500 (by default) items so batching is still in-effect there
+where it matters.
--- a/pulpcore/plugin/stages/api.py
+++ b/pulpcore/plugin/stages/api.py
@@ -179,12 +179,12 @@ class Stage:
         return "[{id}] {name}".format(id=id(self), name=self.__class__.__name__)
 
 
-async def create_pipeline(stages, maxsize=1000):
+async def create_pipeline(stages, maxsize=1):
     """
     A coroutine that builds a Stages API linear pipeline from the list `stages` and runs it.
 
     Each stage is an instance of a class derived from :class:`pulpcore.plugin.stages.Stage` that
-    implements the :meth:`run` coroutine. This coroutine reads asyncromously either from the
+    implements the :meth:`run` coroutine. This coroutine reads asynchronously either from the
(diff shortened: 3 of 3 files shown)


bmbouter:
In syncing EL7 with policy=on_demand without this change it was taking 780 seconds. With this change it was taking 776 seconds. I repeated this experiment 3 times with similar results.

You can see the improvements on memory also. Here's the run without this change which used a max amount of 3041.72 MB.
run_2_memory_q_size_1000_rpm


Here's the run with this change which used a max amount of memory of 2196.75 MB:
run_2_memory_q_size_1_rpm

So that's  a 27.78% reduction along with a small improvement in runtime.

dralley:
>So that's a 27.78% reduction along with a small improvement in runtime.

I love to hear it!

Should we consider ditching the queues entirely for 3.20, and just pushing it directly into the next stage somehow?  A queue with a size of one doesn't seem very different from just making a method call on the next stage.

dralley:
@bmbouter I can't think of anything about this patch which would prevent backporting it, can you?

I would love to backport it to 3.14 (and probably everything in between).

dralley:
This is awesome.   Combining this with some of the other patches which aren't merged / released yet (still waiting on upstream), we're under 900mb for a RHEL 7 sync, which is pretty much the high water mark for *any* sync.  Huge thanks Brian!

#### Status quo circa 2 months ago (no changelog limits):

I forgot to set my memory limits high enough for this to avoid maxing out and thrashing the swap.  It usually gets pretty close to 7gb.

memory_current_incremental_sync_no_changelog_limit

#### Current status quo (changelog limits):

memory_current_incremental_sync_changelog_limit

#### Current RPM first stage sync implementation + this pulpcore patch

memory_brian

#### Future (streaming API) RPM first stage implementation + this pulpcore patch

memory_new_incremental_sync_bmbouter_patch_createrepoc

----

So in a couple of months we've gotten the memory usage down to 1/8th of what it was before, and 1/2 the runtime!

dralley:
Although, there is a short spike at the end where it doubles, which probably has something to do with mirror-sync. (I didn't have when collecting data for the charts above).  

So there's still probably some work to do there.

mdellweg:
stream processing ftw!

dralley (approved):

mdellweg (approved):
I agree that we may look into removing the queues. Or maybe experiment with a little with queue sizes like 2-5.

dralley:
@bmbouter I propose we backport to 3.14, 3.15, 3.16, 3.17, 3.18

bmbouter:
I'd like to merge this as-is and backport it because it's so precise. I put it on the pulpcore agenda for next week to discuss if we should remove the queues entirely or not.

bmbouter:
Here's another comparison running pulp_file against the https://fixtures.pulpproject.org/file-perf/ repo.

# The 1000 batch size (without this PR)

Ran in 800 seconds with max 154.41 MB used

pulp_file_perf_1000


# The 1-queue batch size (with this PR)

Ran in 760 seconds, max 146.50 MB used

pulp_file_perf_1

PR #2365, opened 2022-03-20
Update CI files from main branch

pulpbot:
[noissue]


mdellweg (approved):
Azurite seems to have issues, but i don't think it should hold us from merging.

fao89 (approved):

PR #2375, opened 2022-03-22
Disable Prefetch calls in RAS Stage

bmbouter:
The Prefetch calls have a bug that causes it to not receive all data
from the database all of the time. It seems this issue is in Django.
When encountered the saving of RemoteArtifacts in the stage fails with a
duplicate key.

Accessing the data in this way is slower but it work 100% of the time.


bmbouter:
The issue being investigated is being closed by another PR https://github.com/pulp/pulpcore/pull/2382

PR #2378, opened 2022-03-22
Removes the Django Admin UI Site

### LINKED ISSUE #2374: Remove the django admin UI served at `/admin/`.
## Background + Proposal

We enabled the django admin as tech-preview so that we could view and modify permissions and group membership, back before we had APIs for all of these things. Now that we have APIs for all of this, we should discontinue the use of the django admin UI.

## Why?

Pulp's data is highly relational and having users interact directly with the database via forms is not a safe situation.

## Details

Removing it would involve removing these parts roughly:
* The [admin.py](https://github.com/pulp/pulpcore/blob/fcb355c1f8fec6780524c010cf5bba7e1aa1d51b/pulpcore/app/admin.py)
* The [settings.py](https://github.com/pulp/pulpcore/blob/fcb355c1f8fec6780524c010cf5bba7e1aa1d51b/pulpcore/app/settings.py#L67) and [the setting](https://github.com/pulp/pulpcore/blob/main/pulpcore/app/settings.py#L40).
* The [urls.py usage](https://github.com/pulp/pulpcore/blob/fcb355c1f8fec6780524c010cf5bba7e1aa1d51b/pulpcore/app/urls.py#L144)
* The [docs](https://github.com/pulp/pulpcore/blob/901499a3566a940516664ae6ca042455efee0115/docs/configuration/settings.rst#admin_site_url).

bmbouter:
This was offered with tech-preview, but now with the RBAC, User, and
Group APIs we no longer need it.

closes #2374


mdellweg (approved):

fao89 (approved):

CHANGES/2374.removal
```diff
@@ -0,0 +1,3 @@
+Removed the Django UI Admin site. It was added to provide RBAC permissions management before there
+were APIs that could provide that. It was tech preview and now there are APIs for all of the things
```
ipanova, line 2: is it possible to be a bit more specific instead of saying 'all of the things'?

ipanova (changes requested):
There are docs the need to be removed/updated, at least this one https://github.com/pulp/pulpcoreblob/main/docs/plugins/plugin-writer/concepts/rbac/users_groups.rst

CHANGES/2374.removal
bmbouter: Sure, let me revise.

bmbouter:
@ipanova I pushed the docs removal. We need to do more improvements to the docs, and I hope to do that, but in another PR.

ipanova:
> @ipanova I pushed the docs removal. We need to do more improvements to the docs, and I hope to do that, but in another PR.

Can you open an issue for that so it's tracked? Thank you:)

ipanova (approved):

bmbouter:
I believe the RTFM group will be working on filing and completing tickets like this soon.

PR #2380, opened 2022-03-22
Add a generic distribution list endpoint

### LINKED ISSUE #2379: Add a generic distributions list endpoint

daviddavis:
fixes #2379

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/test_distributions.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_distributions.py
@@ -32,6 +32,7 @@ from pulpcore.client.pulp_file import (
     RepositorySyncURL,
 )
 from pulpcore.tests.functional.api.using_plugin.constants import (
+    BASE_DISTRIBUTION_PATH,
     FILE_CHUNKED_FIXTURE_MANIFEST_URL,
     FILE_CONTENT_NAME,
     FILE_DISTRIBUTION_PATH,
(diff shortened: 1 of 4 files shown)


pulpcore/tests/functional/api/using_plugin/test_distributions.py
```diff
@@ -160,7 +161,13 @@ def test_03_fully_update(self):
                 self.do_fully_update_attr(key)
 
     @skip_if(bool, "distribution", False)
-    def test_04_delete_distribution(self):
+    def test_04_list(self):
+        """Test the generic distribution list endpoint."""
+        distributions = self.client.get(BASE_DISTRIBUTION_PATH)
+        assert self.distribution["pulp_href"] in [distro["pulp_href"] for distro in distributions]
```
daviddavis, line 167: My options here for the test were to use this `CRUDPublicationDistributionTestCase` which doesn't use the bindings or to use the `ContentServePublicationDistributionTestCase` class which does use the bindings but for which doesn't really relate to this test?

bmbouter (approved):
Thank you!

gerrod3 (approved):

pulpcore/tests/functional/api/using_plugin/test_distributions.py
mdellweg: So you are not saying that there will be no bindings generated for this endpoint, but that the test is really outdated. I can live with that.

mdellweg (approved):
Thank you!

pulpcore/tests/functional/api/using_plugin/test_distributions.py
daviddavis: Yea, this test class predates the bindings and still uses the old way of calling the Pulp API.

PR #2395, opened 2022-03-23
Pin redis

fao89:
[noissue]


fao89:
> Why are we pinning redis?

https://github.com/pulp/pulp_installer/runs/5661786624?check_suite_focus=true#step:7:1617

bmbouter:
Is Pulp itself actually incompatible with `Redis>=4.2.0`? Or is it just the environment the installer is running in?

fao89:
from what I understood it is a conflict among pulpcore dependencies, but I triggered pulpcore nightly job to confirm

bmbouter:
Thanks for the info. I think we need an issue tracking the conflicting dependency (if we do have one). I usually use `pipdeptree` to find the conflicting packages info. Also we'll need to see what those dependencies will do to resolve it so we can know when to unpin.

fao89:
it worked on nightly

fao89:
actually, it was a conflict between aiohttp and redis, it happens on pulpcore < 3.17

PR #2402, opened 2022-03-23
Update remove_plugin test for file RBAC

gerrod3:
[noissue]

Should be merged after file RBAC https://github.com/pulp/pulp_file/pull/636


pulpcore/tests/functional/api/using_plugin/test_remove_plugin.py
```diff
@@ -4,6 +4,7 @@
 
 
 @pytest.mark.nightly
+@pytest.mark.trylast
```
bmbouter, line 7: This is slightly scary, but I think we can improve it later.

bmbouter (approved):

pulpcore/tests/functional/api/using_plugin/test_remove_plugin.py
mdellweg: It is not intuitive to me why this should be used. Can you explain in a comment?

gerrod3: In the code? Or here? Using trylast because if the test fails all the other tests will fail after it since pulp_file will be uninstalled.

bmbouter: +1 to a comment on the line in the code

mdellweg (approved):

PR #2418, opened 2022-03-24
Also run pulp_file tests

bmbouter:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/api/test_pulp_file.py
+++ b/pulpcore/tests/functional/api/test_pulp_file.py
@@ -1,0 +1,1 @@
+from pulp_file.tests.functional.api import *  # noqa
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -1,3 +1,4 @@
 # All test requirements
+pulp_file @ git+https://github.com/pulp/pulp_file.git
 -r unittest_requirements.txt
 -r functest_requirements.txt


pulpcore/tests/functional/api/test_pulp_file.py
```diff
@@ -0,0 +1 @@
+from pulp_file.tests.functional.api import *  # noqa
```
mdellweg, line 1: I guess, this is part of the deduplication effort for core and file tests?
Please keep in mind that you might end up installing a dev box with pulpcore and one plugin (not file). This import would fail.
I would prefer if we just added the pulp_file tests to the pytest command in the ci somehow.

bmbouter: I'm on the fence about what to do, but I am very much enjoying the discussion. Yes this is part of that dedplication goal. This PR does make pulp_file a test dependency, so kind of like any other code it imports from a dependency and declares that dependency. Still though I have pause about this approach for the same reasons you do.

What I liked about this approach is that if you want to run the pulpcore functional tests you just run them and the fact that you really are running pulp_file tests isn't something you have to think about. It feels awkward to tell people like the build team and other pulpcore developers that if they want to run pulpcore functional tests really they need to run the pulp_file functional tests. Still though maybe this is unsafe. So I'm basically unsure on what's better.

@pulp/core wdyt? I'm happy to do it either way.

bmbouter: I guess the explicit is better than implicit applied here would mean that we should not have this import and instead have the CI and developers run the pulp_file tests with pytest directly....

bmbouter: Yeah I'm making a PR against the plugin template right now actually. I'll link you to it and probably close this one.

bmbouter:
I'm going to close this in favor of https://github.com/pulp/pulpcore/pull/2432

PR #2419, opened 2022-03-24
Receive fixtures from pulp_file instead

bmbouter:
The pulp_file related fixtures are now being provided by pulp_file
acting as a pytest plugin providing fixtures.

Required PR: https://github.com/pulp/pulp_file/pull/663

[noissue]


bmbouter:
We're moving all of the tests to pulp_file so we won't be consuming fixtures in pulpcore from pulp_file anymore.

PR #2429, opened 2022-03-25
Temporarily pin jinja for building docs

fao89:
[noissue]

### DIFF
--- a/doc_requirements.txt
+++ b/doc_requirements.txt
@@ -5,3 +5,4 @@ sphinx-rtd-theme
 sphinxcontrib-openapi
 towncrier
 mistune<2.0.0
+Jinja2<3.1


doc_requirements.txt
```diff
@@ -5,3 +5,4 @@ sphinx-rtd-theme
 sphinxcontrib-openapi
 towncrier
 mistune<2.0.0
+Jinja2<3.1
```
mdellweg, line 8: Is this a breaking change in jinja, or do we dodge a buggy version?
We could do `!=3.1` in that case...

fao89: seems to be a breaking change
```
AttributeError: module 'jinja2' has no attribute 'contextfilter'
```
https://github.com/pulp/pulp_cookbook/runs/5686532724?check_suite_focus=true

PR #2432, opened 2022-03-25
Run pulp_file tests in pulpcore CI

bmbouter:
The pulp_file tests are eventually going to hold all of the tests which
involve pulp_file. Those are needed to assert correctness on pulpcore
though so let's run those with pulpcore too!

[noissue]

### DIFF
--- a/template_config.yml
+++ b/template_config.yml
@@ -6,6 +6,7 @@
 additional_repos:
 - branch: main
   name: pulp_file
+  pytest_args: pulp_file.tests.functional
 - branch: main
   name: pulp-certguard
 aiohttp_fixtures_origin: 172.18.0.1
--- a/.github/workflows/scripts/script.sh
+++ b/.github/workflows/scripts/script.sh
@@ -186,9 +186,13 @@ else
     if [[ "$GITHUB_WORKFLOW" == "Pulpcore Nightly CI/CD" ]]; then
         pytest -v -r sx --color=yes --suppress-no-test-exit-code --pyargs pulpcore.tests.functional -m parallel -n 8
         pytest -v -r sx --color=yes --pyargs pulpcore.tests.functional -m "not parallel"
+        pytest -v -r sx --color=yes --suppress-no-test-exit-code --pyargs pulp_file.tests.functional -m parallel -n 8
+        pytest -v -r sx --color=yes --pyargs pulp_file.tests.functional -m "not parallel"
     else
         pytest -v -r sx --color=yes --suppress-no-test-exit-code --pyargs pulpcore.tests.functional -m "parallel and not nightly" -n 8
         pytest -v -r sx --color=yes --pyargs pulpcore.tests.functional -m "not parallel and not nightly"
+        pytest -v -r sx --color=yes --suppress-no-test-exit-code --pyargs pulp_file.tests.functional -m "parallel and not nightly" -n 8
+        pytest -v -r sx --color=yes --pyargs pulp_file.tests.functional -m "not parallel and not nightly"
     fi
 
 fi


dralley (approved):

mdellweg:
This is generated from a PR branch of plugin template let's wait on that one: https://github.com/pulp/plugin_template/pull/602

bmbouter:
There is a conflict here, so let's see how the CI runs and if so I think we should merge https://github.com/pulp/plugin_template/pull/602 and we'll refresh this from the merged version (conflict free).

bmbouter:
I reproduced the CI failure locally, running the tests back-to-back on the same system causes the failure. I believe I'll fix by making pulp_file's failing test more resiliant.

bmbouter:
The pulp_file tests fail if you first run this pulpcore test because it leaves data behind:  `pytest -v -r sx --color=yes devel/pulpcore/pulpcore/tests/functional/api/using_plugin/test_content_cache.py::ContentCacheTestCache::test_01_basic_cache_access`

mdellweg:
> The pulp_file tests fail if you first run this pulpcore test because it leaves data behind: `pytest -v -r sx --color=yes devel/pulpcore/pulpcore/tests/functional/api/using_plugin/test_content_cache.py::ContentCacheTestCache::test_01_basic_cache_access`

I think, every test should remove the objects like repos and remotes, but leave the content orphans around. A test that needs to start fresh should run orphan cleanup upfront. AKA stray content in Pulp is a normal thing to have.

bmbouter:
> > The pulp_file tests fail if you first run this pulpcore test because it leaves data behind: `pytest -v -r sx --color=yes devel/pulpcore/pulpcore/tests/functional/api/using_plugin/test_content_cache.py::ContentCacheTestCache::test_01_basic_cache_access`
> 
> I think, every test should remove the objects like repos and remotes, but leave the content orphans around. A test that needs to start fresh should run orphan cleanup upfront. AKA stray content in Pulp is a normal thing to have.

+1 to this. I'm going to try to do this when I fix this PR. The input is very helpful because I was going another way and it's just not as great as this strategy.

bmbouter:
This PR is waiting on https://github.com/pulp/pulpcore/pull/2479 to merge at which point I will update this to run the pulp_file tests which should pass this time.

pulpcore/tests/functional/api/using_plugin/test_purge.py
```diff
@@ -84,44 +84,31 @@ def setUpClass(cls):
...
-            gen_file_remote(
-                "https://fixtures.pulpproject.org/THEREISNOFILEREPOHERE/", policy="on_demand"
-            )
-        )
-        cls.bad_repo = cls.repo_api.create(gen_repo())
-        cls.bad_sync_data = RepositorySyncURL(remote=cls.bad_remote.pulp_href)
-
-    @classmethod
-    def tearDownClass(cls) -> None:
+    def tearDown(self):
```
ggainey, line 87: purge as it exists is def order-dependent, and doesn't want to sync more than once. However, the sync is on-demand of a small(ish) repo, and I think this particular suite prob needs to just be refactored to take better advantage of the workflow-infra. I don't know that that needs to happen as part of this PR.

pulpcore/tests/functional/api/using_plugin/test_pulpexport.py
```diff
@@ -50,30 +50,26 @@ class BaseExporterCase(unittest.TestCase):
     class.
     """
 
-    @classmethod
-    def _setup_repositories(cls):
+    def _setup_repositories(self):
```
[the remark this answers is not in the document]
ggainey: moved to #2479

pulpcore/tests/functional/api/using_plugin/test_filesystemexport.py
```diff
@@ -44,27 +44,26 @@ class BaseExporterCase(unittest.TestCase):
     class.
     """
 
-    @classmethod
-    def _setup_repositories(cls):
+    def _setup_repositories(self):
```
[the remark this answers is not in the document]
ggainey: moved to #2479

template_config.yml
```diff
@@ -6,6 +6,7 @@
 additional_repos:
 - branch: main
   name: pulp_file
+  pytest_args: pulp_file.tests.functional
```
mdellweg, line 9: Should this be a list instead? We might want to run both file and certguard tests here.
But may we will just add them in the same run?

bmbouter: I had imagined adding them in the same run, but maybe that's not ideal? We definitely need to allow multiple entries one way or the other. Hmmmm

ggainey (approved):

PR #2469, opened 2022-03-31
Enable bypassing DB encryption [#2466]

### LINKED ISSUE #2048: As a user I want to be able to rotate my encryption keys and rekey my informations
Author: spredzy (spredzy)

Redmine Issue: 9397, https://pulp.plan.io/issues/9397

---

Pulp 3.15 brings the support for encrypting fields in the DB. This is a great step toward better security practices. Thanks team for that.

In order to go a step further with security best-practices, I would like to be able to rotate my keys periodically, and hence rekey my data.

As it's stand today I haven't see a way to do this.

### LINKED ISSUE #2466: As a user already using database-level encryption, I don't want to have Pulp encrypt anything
I use AWS RDS as my database backend.  Data is configured to use encryption at rest, and access to the database is authenticated.  In an environment with servers as cattle, I would prefer not to have to maintain an encryption key at all, as I don't consider the field-level encryption to be a substantially significant security improvement.

I would like a way to skip the field-level encryption in Pulp, ideally by specifying a configuration value

Pre-generating an encryption key and storing it as a k8s secret is not a terrible solution, as it could be mounted into the container as a file and could persist across container lifespan.  That still means I have to back up the secret value somewhere (preferably alongside the database backup) for disaster recovery purposes.  This whole problem is conveniently solved already by using AWS key management for the database and AWS-aware backup tooling, so it's not great to have an application reinvent my wheels. :)

**Additional context**
Refs #1301 and indirectly also #2048 
PR coming shortly :)

dannysauer:
Alter behavior of field-level encryption such that a false value in DB_ENCRYPTION_KEY results in bypassing the encryption and decryption of "encrypted" text field data

Fixes #2466

PR is initially expecting to solicit input.  In particular, I'd like to know if this approach seems like the most reasonable approach (it seemed like the least invasive way to mess with the code), and I'd like to know if it's appropriate to leave without a "decrypt migration".  Presumably it would be possible to add a `old_encryption_key` parameter or something like that which could be used as a one-off to decrypt the DB with a migration if someone decided to turn off encryption.  Perhaps addressing the logic to decrypt would help resolve #2048; there's just be a two-step process where the current key is moved to the old key, and then the DB is reencrypted with the new key (which could be empty in this case, or a new key in that case).  In any event, I'm not sure how to do a conditional migration or if migration is the right facility - I guess it'd have to be based on 1) old key parameter is present and 2) the value of the old key hasn't already been subjected to a migration.

### DIFF
--- a/CHANGES/2466.feature
+++ b/CHANGES/2466.feature
@@ -1,0 +1,3 @@
+Adds the ``DB_ENCRYPTION`` setting which if ``False`` disables at-rest database encryption.
+
+Note: does not decrypt already encrypted values
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -36,6 +36,17 @@ SECRET_KEY
    chars = 'abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)'
    print(''.join(random.choice(chars) for i in range(50)))
 
+DB_ENCRYPTION
+^^^^^^^^^^^^^
+
+  By default Pulp uses database encryption. If ``False`` the at-rest database encryption is not used and values are stored
+  in plain text.
+
+  .. warning::
+
+     values which are already encrypted are not decrypted if this setting is changed; best 
+     to only bypass encryption in a clean installation. 
+
 DB_ENCRYPTION_KEY
 ^^^^^^^^^^^^^^^^^
 
--- a/pulpcore/app/management/commands/datarepair-2327.py
+++ b/pulpcore/app/management/commands/datarepair-2327.py
@@ -36,8 +36,9 @@ class Command(BaseCommand):
         dry_run = options["dry_run"]
         fields = ("username", "password", "proxy_username", "proxy_password", "client_key")
 
-        with open(settings.DB_ENCRYPTION_KEY, "rb") as key_file:
-            fernet = cryptography.fernet.Fernet(key_file.read())
+        if settings.DB_ENCRYPTION:
+            with open(settings.DB_ENCRYPTION_KEY, "rb") as key_file:
+                fernet = cryptography.fernet.Fernet(key_file.read())
 
         possibly_affected_remotes = (
             Q(username__isnull=False)
@@ -88,8 +89,9 @@ class Command(BaseCommand):
                             continue
 
                         try:
-                            # try to decrypt it again
-                            field_value = force_str(fernet.decrypt(force_bytes(field_value)))
+                            if settings.DB_ENCRYPTION:
+                                # try to decrypt it again
+                                field_value = force_str(fernet.decrypt(force_bytes(field_value)))
                             # it was decrypted successfully again time, so it was probably
                             # encrypted multiple times over. lets re-set the value with the
                             # newly decrypted value
--- a/pulpcore/app/models/fields.py
+++ b/pulpcore/app/models/fields.py
@@ -96,10 +96,14 @@ class EncryptedTextField(TextField):
 
     def get_db_prep_save(self, value, connection):
         value = super().get_db_prep_save(value, connection)
+        if not settings.DB_ENCRYPTION:
+            return value
         if value is not None:
             return force_str(self._fernet.encrypt(force_bytes(value)))
 
     def from_db_value(self, value, expression, connection):
+        if not settings.DB_ENCRYPTION:
+            return value
         if value is not None:
             return force_str(self._fernet.decrypt(force_bytes(value)))
 
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -55,6 +55,8 @@ FILE_UPLOAD_HANDLERS = ("pulpcore.app.files.HashingFileUploadHandler",)
 
 SECRET_KEY = True
 
+DB_ENCRYPTION = True
+
 # Key used to encrypt fields in the database
 DB_ENCRYPTION_KEY = "/etc/pulp/certs/database_fields.symmetric.key"
 
@@ -383,6 +385,7 @@ if not (
     Path(sys.argv[0]).name == "pytest"
     or Path(sys.argv[0]).name == "sphinx-build"
     or (len(sys.argv) >= 2 and sys.argv[1] == "collectstatic")
+    or (not DB_ENCRYPTION or not DB_ENCRYPTION_KEY)
 ):
     try:
         with open(DB_ENCRYPTION_KEY, "rb") as key_file:
--- a/pulpcore/tests/unit/models/test_remote.py
+++ b/pulpcore/tests/unit/models/test_remote.py
@@ -2,7 +2,7 @@ from uuid import uuid4
 from unittest.mock import patch, mock_open
 
 from django.db import connection
-from django.test import TestCase
+from django.test import TestCase, override_settings
 
 from pulpcore.app.models import Remote
 from pulpcore.app.models.fields import EncryptedTextField
@@ -36,3 +36,26 @@ class RemoteTestCase(TestCase):
             self.assertNotEqual(db_proxy_password, "test")
             self.assertEqual(proxy_password, "test")
             self.assertEqual(mock_file.call_count, 2)
+
+    @patch(
+        "pulpcore.app.models.fields.open",
+        new_callable=mock_open,
+        read_data=b"hPCIFQV/upbvPRsEpgS7W32XdFA2EQgXnMtyNAekebQ=",
+    )
+    @override_settings(DB_ENCRYPTION=False)
+    def test_bypassed_encrypted_proxy_password(self, mock_file):
+        self.remote = Remote(name=uuid4(), proxy_password="test")
+        self.remote.save()
+        self.assertEqual(Remote.objects.get(pk=self.remote.pk).proxy_password, "test")
+
+        # check the database that proxy_password is not encrypted
+        with connection.cursor() as cursor:
+            cursor.execute(
+                "SELECT proxy_password FROM core_remote " f"WHERE pulp_id = '{self.remote.pulp_id}'"
+            )
+            db_proxy_password = cursor.fetchone()[0]
+            proxy_password = EncryptedTextField().from_db_value(db_proxy_password, None, connection)
+            self.assertEqual(proxy_password, "test")
+            # these two tests should be what differs from the non-bypassed test
+            self.assertEqual(db_proxy_password, "test")
+            self.assertEqual(mock_file.call_count, 0)


dannysauer:
Also exciting that the test case seems to at least catch some kind of error (whether it's in the case or the code being tested, though...). :D

pulpcore/tests/unit/models/test_remote_unencrypted.py
```diff
@@ -0,0 +1,40 @@
...
+        if self.remote:
+            self.remote.delete()
+
+    @patch(
+        "pulpcore.app.models.fields.open",
+        new_callable=mock_open,
+        read_data=b"test",
+    )
+    def test_unencrypted_proxy_password(self, mock_file):
+        settings.DB_ENCRYPTION_KEY = False
```
mdellweg, line 26: I think, i've seen a decorator before to mock a setting for a single test.

mdellweg: I do not understand how it is supposed to work there (as it does not seem to be a unittest), but here it should just do the trick:
https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/api/test_status.py#L112

dannysauer: Well, that didn't have the intended result.
https://docs.djangoproject.com/en/4.0/topics/testing/tools/#overriding-settings has several options.  Either way, I should really just follow the directions and set up local testing instead of burning all of the project's Github Actions minutes. :D

Thinking about it, I also probably should have just added another test in the test_remote.py instead of creating a whole new file. 🤦

docs/configuration/settings.rst
```diff
@@ -50,6 +50,10 @@ DB_ENCRYPTION_KEY
 
     openssl rand -base64 32 > /etc/pulp/certs/database_fields.symmetric.key
 
+  If this parameter is set to boolean false or 0, the encryption is bypassed and values are stored
```
ipanova, line 53: i'd add a separate boolean config option to not overload already existing DB_ENCRYPTION_KEY.
We'd need to also update our repair command @dralley has added to check whether the pulp encryption is on/off https://github.com/pulp/pulpcore/blob/main/pulpcore/app/management/commands/datarepair-2327.py

docs/configuration/settings.rst
```diff
@@ -50,6 +50,10 @@ DB_ENCRYPTION_KEY
...
+  in plain test. Warning: values which are already encrypted are not decrypted if this setting is
```
ipanova, line 54: s/test/text

ipanova:
@dannysauer for  the existing installation that would want to turn off the encryption we could provide management command similar to the repair that would be decrypting the fields based on what's the settings

docs/configuration/settings.rst
dannysauer: Hah. Plain test. 🤦‍♂️

pulpcore/app/settings.py
```diff
@@ -340,6 +340,7 @@
 if not (
     Path(sys.argv[0]).name == "sphinx-build"
     or (len(sys.argv) >= 2 and sys.argv[1] == "collectstatic")
+    or (not DB_ENCRYPTION_KEY)
```
bmbouter, line 343: JFYI, check out this PR also, which will probably merge conflict with this. If this merged I think that's fine otherwise it'll need some conflict resolution. https://github.com/pulp/pulpcore/pull/2453

bmbouter:
Overall I'm ok with users to have the option to disable the encryption, so +1 whenever the other reviewer's comments are addressed. @dannysauer ty for opening this!

dannysauer:
FWIW, I'm on vacation this week, so may or may not get a chance to promptly address all that good feedback. Just sharing so no one thinks It's been dropped on the floor. :)

goosemania:
@dannysauer , would you have a moment to look at this PR?

dannysauer:
> @dannysauer , would you have a moment to look at this PR?

Vacation took longer than I thought. ;)

I'm back on figuring this out this week. Thanks for following up.

dralley:
@dannysauer Ping, status update?

attenuation:
Hi, i will take over this pr, So anyone can help me to active this PR from the draft status?

dannysauer:
> Hi, I will take over this pr, So Can anyone help me to active this PR from the draft status?

There you go. :)

(sorry; I changed jobs and don't use pulp now - and haven't had time to work on it. Thanks for picking it back up, @attenuation)

bmbouter:
@dannysauer I totally understand. Thank you just the same.

@attenuation welcome, there are some linting issues. Please squash your commits to one migration, and also look at the flake8 and/or black errors. I'll also leave some comments now. Please ask any questions you have; we want to make this as easy a contribution process as possible! FYI we have a [new dev env](https://github.com/pulp/oci_env) in case that is helpful.

docs/configuration/settings.rst
```diff
@@ -36,6 +36,13 @@ SECRET_KEY
    chars = 'abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)'
    print(''.join(random.choice(chars) for i in range(50)))
 
+DB_ENCRYPTION
+^^^^^^^^^^^^^
+
+  If this parameter is set to boolean false or 0, the encryption is bypassed and values are stored
```
bmbouter, line 42:
```suggestion
If ``False`` the at-rest database encryption is not used and values are stored
```

bmbouter: Also please state that the default is True (with True surrounded by double backticks which is needed for RST escaping in Sphinx.

docs/configuration/settings.rst
```diff
@@ -36,6 +36,13 @@ SECRET_KEY
...
+  in plain text. Warning: values which are already encrypted are not decrypted if this setting is
```
bmbouter, line 43: This warning can become it's own line below and use the RST directive for it. See this example https://github.com/pulp/pulpcore/blob/5c84ab71928aef83a37bc97b5eda7d2a60463807/docs/configuration/settings.rst#L119

pulpcore/app/models/fields.py
```diff
@@ -86,6 +86,7 @@ def __init__(self, *args, **kwargs):
             raise ImproperlyConfigured("EncryptedTextField does not support unique=True.")
         if kwargs.get("db_index"):
             raise ImproperlyConfigured("EncryptedTextField does not support db_index=True.")
+        self.skipencryption = not settings.DB_ENCRYPTION
```
bmbouter, line 89: I'd prefer to not set this and instead have the if statements below use: `if not settings.DB_ENCRYPTION` to keep fewer moving parts.

pulpcore/app/settings.py
```diff
@@ -383,6 +384,7 @@
     Path(sys.argv[0]).name == "pytest"
     or Path(sys.argv[0]).name == "sphinx-build"
     or (len(sys.argv) >= 2 and sys.argv[1] == "collectstatic")
+    or (not DB_ENCRYPTION or not DB_ENCRYPTION_KEY)
```
bmbouter, line 387: :+1:

pulpcore/tests/unit/models/test_remote.py
```diff
@@ -36,3 +36,26 @@ def test_encrypted_proxy_password(self, mock_file):
             self.assertNotEqual(db_proxy_password, "test")
             self.assertEqual(proxy_password, "test")
             self.assertEqual(mock_file.call_count, 2)
+
+    @patch(
+        "pulpcore.app.models.fields.open",
+        new_callable=mock_open,
+        read_data=b"hPCIFQV/upbvPRsEpgS7W32XdFA2EQgXnMtyNAekebQ=",
+    )
+    @override_settings(DB_ENCRYPTION=False)
```
bmbouter, line 45: nice test. :+1:

CHANGES/2466.feature
```diff
@@ -0,0 +1,3 @@
+Adds the ``DB_ENCRYPTION`` setting which if ``False`` disables at-rest database encryption.
+
+Note: does not decrypt already encrypted values
```
mdellweg, line 3: We do have a management command to repair these fields when a certain bug encrypted them an unpredictable amount of times. Do you think we can come up with another management command that allows for encrypting/decrypting/re-encrypting them?
cc @dralley

docs/configuration/settings.rst
ipanova: @bmbouter I feel uneasy to leave existing fields encrypted. We basically say for existing/upgraded installations there is no way to turn off the encryption consistently, unless this is a brand new fresh install. Is this the message we're sending?

CHANGES/2466.feature
dralley: We probably could, yes.

docs/configuration/settings.rst
bmbouter: With this being a driven entirely by community contribution, to me, it's kind of up to them how much they want to do. My perspective is that if a community contributor only wants to go so far, it's up to the next community user who wanted it to go further to make their contribution at a later time.

That being said I hear where you're coming from. There is a judgement call from project maintainers for what we think is responsible to accept on behalf of the user base. For me, I'd be willing to accept this without commands to fix it on existing in-use installations because it's clearly labeled as such. The best case is @attenuation is willing to make commands to handle that case too. If not though, I'll defer to you (happily) on whether to accept or not. If you want to require more I would understand. Use the required changes I think to indicate your desires there and I would +1 it.

ipanova: @bmbouter looking at it this way i am comfortable to not have a repair command. Thank you for sharing your perspective!

attenuation:
@bmbouter I just squash my commit, but the first commit is too old, I can't squash it.

mdellweg:
> @bmbouter I just squash my commit, but the first commit is too old, I can't squash it. And about the repair command, I am sorry I can not do this because I am not very familiar with this project.

You probably need to rebase then.
And the added management command is an additional idea for sure. Don't bother about it. We can have that as a followup piece.

pulpcore/app/settings.py
mdellweg: is this the same as `if not (...) and DB_ENCRYPTION and DB_ENCRYPTION_KEY`?
Should we maybe instead validate that `DB_ENCRYPTION_KEY` must be set if `DB_ENCRYPTION` and then just have `... and DB_ENCRYPTION` here?

pulpcore/app/management/commands/datarepair-2327.py
```diff
@@ -88,8 +89,9 @@ def handle(self, *args, **options):
                             continue
 
                         try:
-                            # try to decrypt it again
-                            field_value = force_str(fernet.decrypt(force_bytes(field_value)))
+                            if settings.DB_ENCRYPTION:
+                                # try to decrypt it again
+                                field_value = force_str(fernet.decrypt(force_bytes(field_value)))
```
mdellweg, line 94: This will lead to in infinite loop.
I think we should exit the repair command right away it encryption is disabled.

mdellweg (changes requested):

dkliban:
@attenuation Are you still working on this? How can we help you move this work forward?

PR #2470, opened 2022-04-01
Ignore black commit

daviddavis:
Thought this looked useful:

https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,3 @@
+# .git-blame-ignore-revs
+# enabled black formatting
+b94abd64d76ea4554e6750ff38ce458eaa888cc8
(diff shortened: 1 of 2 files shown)


.git-blame-ignore-revs
```diff
@@ -0,0 +1,3 @@
+# .git-blame-ignore-revs
+# enabled black formatting
+b94abd64d76ea4554e6750ff38ce458eaa888cc8
```
daviddavis, line 3: https://github.com/pulp/pulpcore/commit/b94abd64d76ea4554e6750ff38ce458eaa888cc8

mdellweg: Would be nice to have a regex on the commit message... but i don't see that documented.

mdellweg (approved):

gerrod3 (approved):

PR #2478, opened 2022-04-04
Fixed rare bulk_update() deadlock in ContentSaver.

### LINKED ISSUE #2430: bulk_update() in content-stages can cause (very rare) deadlock
**Version**
3.14

**Describe the bug**
In high-concurrency environments, with overlapping content, calling bulk_update() can cause a deadlock. Specifically, this call:

https://github.com/pulp/pulpcore/blob/main/pulpcore/plugin/stages/content_stages.py#L158-L164

Ordering the list-to-be-updated does not, alas, protect us - because Postgres doesn't guarantee order when doing an update like this.

**To Reproduce**
We have only seen this "in the wild" once, syncing 8-10 repos with similar content at the same time with 10 workers available.

**Expected behavior**
Don't deadlock.

**Additional context**
This is the traceback from the initial description for

https://bugzilla.redhat.com/show_bug.cgi?id=2062526

We fixed the deadlock noted in https://bugzilla.redhat.com/show_bug.cgi?id=2062526#c2 under #2420

ggainey:
See https://bugzilla.redhat.com/show_bug.cgi?id=2062526 for details on
forcing the deadlock and discussion of the fix.

fixes #2430.
[nocoverage]

### DIFF
--- a/pulpcore/plugin/stages/content_stages.py
+++ b/pulpcore/plugin/stages/content_stages.py
@@ -155,13 +155,39 @@ class ContentSaver(Stage):
                         content_artifact.artifact = to_update_ca_artifact[key]
                         to_update_ca_bulk.append(content_artifact)
 
-                    # Sort the lists we're about to do bulk updates/creates on.
-                    # We know to_update_ca_bulk entries already are in the DB, so we can enforce
-                    # order just using pulp_id.
-                    to_update_ca_bulk.sort(key=lambda x: x.pulp_id)
-                    content_artifact_bulk.sort(key=lambda x: ContentArtifact.sort_key(x))
+                    # to_update_ca_bulk are the CAs that we know are already persisted.
+                    # We need to update their artifact_ids, and wish to do it in bulk to
+                    # avoid hundreds of round-trips to the database.
+                    #
+                    # To avoid deadlocks in high-concurrency environments with overlapping
+                    # content, we need to update the rows in some defined order. Unfortunately,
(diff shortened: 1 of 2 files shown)


dralley:
Does this need a 3.14 backport?

ggainey:
> Does this need a 3.14 backport? (Presumably a manual one)

Target is set to 6.11 (which is 3.16) - but 6.10.z flag is also set, so we prob should bring it back to 3.14 as well.

dralley (approved):

pulpcore/plugin/stages/content_stages.py
```diff
@@ -155,13 +155,39 @@ def process_batch():
...
+                            .order_by("pulp_id")
+                            .select_for_update()
+                            .values_list()
+                        )
+                        ContentArtifact.objects.bulk_update(to_update_ca_bulk, ["artifact"])
 
-                    ContentArtifact.objects.bulk_update(to_update_ca_bulk, ["artifact"])
+                    # To avoid a similar deadlock issue when calling get_or_create, we sort the
+                    # "new" CAs to make sure inserts happen in a defined order. Since we can't
+                    # trust the pulp_id (by the time we go to create a CA< it may already exist,
```
dralley, line 188: typo: s/</,

ggainey: > typo: s/</,

Ugh, prob my most-common typo :(

mdellweg (approved):

PR #2483, opened 2022-04-05
Fix the note about a default value

lubosmj:
[noissue]


bmbouter (approved):
Thanks! I think this is such a small thing it's ok to use [noissue]. :+1:

gerrod3 (approved):

Issue #1949, opened 2022-01-17
FilesystemExporter does not gracefully handle exporting multiple times


pulpbot:
Author: daviddavis (daviddavis)

Redmine Issue: 7918, https://pulp.plan.io/issues/7918

---

The move to GHA exposed this bug (not sure how it wasn't an issue in Travis but I was able to reproduce locally).

If there are pre-existing files in the location that FileSystem attempts to export to you get a FileExistsError. This is common in the case where you're exporting multiple times.

Some potential solutions:

1. Check that the directory is empty before exporting. Maybe even refuse to export but have a --force flag that will remove files in this directory.
2. Ignore file exists errors either silently or raise a warning.

pulpbot:
From: @ipanova (ipanova@redhat.com)
Date: 2020-12-04T15:53:57Z

---

I would  ignore file exists error. If with every export we expect an additive mode, it would be a waste to --force remove existing file so just in a later step place them back.

PR #2484, opened 2022-04-06
Publication filesystem export management command

dralley:
Simple way to dump a publication straight to disk.  At the very least it would simplify debugging.

[noissue]

### DIFF
--- a/pulpcore/app/management/commands/dump-publications-to-fs.py
+++ b/pulpcore/app/management/commands/dump-publications-to-fs.py
@@ -1,0 +1,166 @@
+from gettext import gettext as _
+import logging
+import os
+
+from django.core.exceptions import ObjectDoesNotExist
+from django.core.management import BaseCommand, CommandError
+
+from pulpcore.app.models import Publication, Distribution
+from pulpcore.app.tasks.export import (
+    _export_publication_to_file_system,
+    _export_location_is_clean,
+    UnexportableArtifactException,
+)
+from pulpcore.app.viewsets.base import NamedModelViewSet
+from pulpcore.constants import FS_EXPORT_METHODS
+
+
+class Command(BaseCommand):
+    """Django management command for exporting full repositories to a location on disk."""
+
+    help = _("Export full repositories to a location on disk.")
+
+    def add_arguments(self, parser):
+        """Set up arguments."""
+        parser.add_argument("--publication", required=False, help=_("A publication ID."))
+        parser.add_argument(
+            "--distribution-path-prefix",
+            required=False,
+            help=_("A filter for distributions whose base_path begins with the provided prefix."),
+        )
+        parser.add_argument(
+            "--type",
+            required=False,
+            help=_("Only export publications of a particular type, e.g. 'rpm.rpm'"),
+        )
+        parser.add_argument(
+            "--method",
+            required=False,
+            default=FS_EXPORT_METHODS.WRITE,
+            choices=(
+                FS_EXPORT_METHODS.WRITE,
+                FS_EXPORT_METHODS.HARDLINK,
+                FS_EXPORT_METHODS.SYMLINK,
+            ),
+            help=_("Method of exporting, e.g. write, hardlink, symlink"),
+        )
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help=_("Don't export anything, just print what would be exported."),
+        )
+        parser.add_argument(
+            "--allow-incomplete",
+            action="store_true",
+            help=_("Dump the publication even if not all files are available."),
+        )
+        parser.add_argument(
+            "--allow-dirty",
+            action="store_true",
+            help=_(
+                "Try to dump the publication even if the target directories already contain data."
+            ),
+        )
+        parser.add_argument("dest", help=_("Place to drop the files."))
+
+    def handle(self, *args, **options):
+        """Implement the command."""
+        to_export = []
+
+        if options.get("publication") and options.get("distribution_path_prefix"):
+            raise CommandError("Cannot provide both --publication and --distribution-path-prefix")
+        # If a publication was specified just dump it there and exit
+        elif options["publication"]:
+            try:
+                publication_pk = NamedModelViewSet.extract_pk(options["publication"])
+            except Exception:
+                publication_pk = options["publication"]
+            publication = Publication.objects.get(pk=publication_pk)
+            to_export.append((options["dest"], publication))
+
+        # If no publication was specified go through the distributions and dump them if they
+        # meet the criteria
+        else:
+            # If a base_path prefix was provided, filter out distributions with a base path
+            # that doesn't start with the prefix
+            if options.get("distribution_path_prefix"):
+                distributions = Distribution.objects.filter(
+                    base_path__startswith=options["distribution_path_prefix"]
+                )
+            else:
+                distributions = Distribution.objects.all()
+
+            # Filter out distributions that don't match the type specified (if any)
+            if options["type"]:
+                distributions = distributions.filter(pulp_type__startswith=options["type"])
+
+            # For all matching distributions, if they have a publication, dump it in a directory
+            # matching the original distribution structure
+            for distribution in distributions:
+                if distribution.publication:
+                    publication = distribution.publication
+                elif distribution.repository:
+                    repository = distribution.repository
+                    # Account for distributions serving the latest publication of a given repository
+                    try:
+                        publication = Publication.objects.filter(
+                            repository_version__in=repository.versions.all(), complete=True
+                        ).latest("repository_version", "pulp_created")
+                        repo_path = os.path.join(options["dest"], distribution.base_path)
+                        to_export.append((repo_path, publication))
+                    except ObjectDoesNotExist:
+                        logging.warn(
+                            "No publication found for the repo published at '{}': skipping".format(
+                                distribution.base_path
+                            )
+                        )
+
+        # Go through all the target directories first, if any of them are dirty, print warnings
+        # and exit - unless the user explicitly asked to go through with it anyway.
+        if not options["allow_dirty"]:
+            is_dirty = False
+            for dest, publication in to_export:
+                if not _export_location_is_clean(dest):
+                    logging.error(
+                        _("Directory '{}' must be empty to be used as an export location").format(
+                            dest
+                        )
+                    )
+                    is_dirty = True
+
+            if is_dirty:
+                raise CommandError(_("Cannot export to directories that contain existing data."))
+
+        if options["dry_run"]:
+            for dest, publication in to_export:
+                logging.info(
+                    "Will write publication '{}' to '{}'".format(str(publication.pk), dest)
+                )
+        else:
+            for dest, publication in to_export:
+                self.export_publication(
+                    dest, publication, options["method"], options["allow_incomplete"]
+                )
+                logging.info("Wrote publication '{}' to '{}'".format(str(publication.pk), dest))
+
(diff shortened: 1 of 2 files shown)


dralley:
Single commit check seems to be broken.

ggainey:
FilesystemExport exists to do exactly this https://docs.pulpproject.org/pulpcore/restapi.html#operation/exporters_core_filesystem_exports_list

dralley:
It's not exactly the same:  Filesystem Export requres individual publications to be hooked up to individual paths, whereas this command can dump the entire directory tree of distributed repositories matching a provided "base_path" prefix on a per-plugin basis.

It's far simpler and more capable as a utility, though perhaps not as programmable.

dralley:
@parthaa @mdellweg

Screenshot from 2022-07-27 15-43-21

dralley:
I found a "bug", if you use the symlink mode (maybe hardlink too), you can't write the export a second time.

```
  File "/home/vagrant/devel/pulpcore/pulpcore/app/tasks/export.py", line 76, in _export_to_file_system
    os.symlink(src, dest)
FileExistsError: [Errno 17] File exists: '/var/lib/pulp/media/artifact/86/338fca53ae66459e1fb7ec7d0ec4190f407a81a2b6d95ae6f119699a2c845e' -> 'exporttest/rpm_repos/kickstart/.treeinfo'
```

dralley:
Open questions (that might apply to the other exporter as well), what should happen when the destination already has files with the same names being written?  Overwrite everything blindly?  Delete, *then* write?  Fail if there are conflicts?

dralley:
Lastly - should this be backported?

pulpcore/app/management/commands/publication-fs-export.py
```diff
@@ -0,0 +1,131 @@
...
+            logging.info(
+                "Wrote publication '{}' to '{}'".format(str(publication.pk), options["dest"])
+            )
+
+        # If no publication was specified go through the distributions and dump them if they
+        # meet the criteria
+        else:
+            # If a base_path prefix was provided, filter out distributions with a base path
+            # that doesn't start with the prefix
+            if options.get("distribution_path_prefix"):
```
mdellweg, line 73: This ought to be a path prefix, right? Should we make sure it ends with a slash, or are there any useful `os.path` helpers?

pulpcore/app/management/commands/publication-fs-export.py
```diff
@@ -0,0 +1,131 @@
...
+                method=method,
+                allow_missing=allow_incomplete,
+            )
+        except RuntimeError:
+            raise CommandError(
+                _(
+                    "An artifact present in this publication is not available, please download "
+                    "the full contents of this repository or else use the --allow-incomplete flag."
+                )
+            )
```
mdellweg, line 131: This error message is way more specific than the exception we can catch here. Maybe we need that function to throw a subclass of `RuntimeError` that we can catch here.

pulpcore/app/management/commands/publication-fs-export.py
dralley: This is just treating it as a string prefix rather than a path prefix per-se, full segments aren't required and I'm not sure we need / want to require them.  Should I adjust the phrasing?

dralley: Agreed in that example, but a counterexample is if you have repos set up like this:

```
cs8_appstream_x86_64/
cs8_baseos_x86_64/
cs8_appstream_aarch64/
cs8_baseos_aarch64/
cs9_appstream_x86_64/
cs9_baseos_x86_64/
cs9_appstream_aarch64/
cs9_baseos_aarch64/
```
And you want some way to only export the centos stream 8 repos.  I think this is plausibly common enough that we might want to err towards providing more flexibility.

There's another option I suppose - we could default to printing a list of all the repos the user is about to export and ask them to verify before beginning the export.

mdellweg: Let's just be explicit about this "feature" then. And maybe a "--dry-run" is always a good idea.

dralley: Adjusted the description and added `--dry-run`

dralley:
@mdellweg Could you comment on some of the open questions?

pulpcore/app/management/commands/publication-fs-export.py
```diff
@@ -0,0 +1,143 @@
...
+    UnexportableArtifactException,
+)
+from pulpcore.app.viewsets.base import NamedModelViewSet
+from pulpcore.constants import FS_EXPORT_METHODS
+
+
+class Command(BaseCommand):
+    """Django management command for exporting full repositories to a location on disk."""
+
+    help = _(__doc__)
```
mdellweg, line 20: This will not work. Gettext will not find the real string here.

mdellweg (approved):

mdellweg:
> Open questions (that might apply to the other exporter as well), what should happen when the destination already has files with the same names being written? Overwrite everything blindly? Delete, then write? Fail if there are conflicts?

Should we expect that we always export into an empty directory and fail if it is not?

pulpcore/app/management/commands/publication-fs-export.py
dralley: Do you mean the gettext doesn't work or the help command won't work?  Because the django help text comes from defining [`help`](https://docs.djangoproject.com/en/4.0/howto/custom-management-commands/) and this does actually work (which was surprising to me as well)

```
(pulp) [vagrant@pulp3-source-fedora36 devel]$ python pulpcore/manage.py publication-fs-export exporttest --help
usage: manage.py publication-fs-export [-h] [--publication PUBLICATION] [--distribution-path-prefix DISTRIBUTION_PATH_PREFIX] [--type TYPE] [--method {write,hardlink,symlink}] [--dry-run]
                                       [--allow-incomplete] [--version] [-v {0,1,2,3}] [--settings SETTINGS] [--pythonpath PYTHONPATH] [--traceback] [--no-color] [--force-color]
                                       [--skip-checks]
                                       dest

Django management command for exporting full repositories to a location on disk.

positional arguments:
  dest                  Place to drop the files.

options:
```

mdellweg: I mean that extracting the strings for translating will not work. The call to `_()` will work just fine, but the string to translate will never be found in the catalog, unless you have a literal `_("...")` with the same content somewhere in the project.

dralley:
> Should we expect that we always export into an empty directory and fail if it is not?

I think that would be sensible, but it comes down to the use cases, particularly for the exporter workflow.  @parthaa @ggainey, what contract should we make here?

ggainey:
> Open questions (that might apply to the other exporter as well), what should happen when the destination already has files with the same names being written? Overwrite everything blindly? Delete, then write? Fail if there are conflicts?

Current-export should always Win. We could make it simpler and just check to see if the dest has any files, and fail with a "dirty destination" right at start. Or we could make it more complicated, check when there's a conflict and log "existing file already exists" or "existing file overwritten by new expoort w/ diff checksum".

> 
> Lastly - should this be backported? Renamed?

It's a feature, backporting is contraindicated. Let's think about replacing the word 'export' with...something. The exporter/export process has a bunch of implications, that this does not (eg, incremental, chunking, repo-version-based, blahblahblah), that people might "assume" was true of this as well.

Extract? Expose? Snapshot? There has to be some better word...

dralley:
>     Current-export should always Win. We could make it simpler and just check to see if the dest has any files, and fail with a "dirty destination" right at start. Or we could make it more complicated, check when there's a conflict and log "existing file already exists" or "existing file overwritten by new expoort w/ diff checksum".

Simple is better, the dirty-destination failure sounds like a good idea.  I would vote to do that, with maaaaybe an option to override it and write the export out anyway at your own risk.

Some semantics need to be worked out though, we should probably check all the directories up-front and provide a list of the dirty ones, yeah?   Wouldn't want it to half-finish the export and then fail.

>It's a feature, backporting is contraindicated.

Well, I mostly meant the behavioral fixes.  https://github.com/pulp/pulpcore/pull/2951 was backported after all, and some of these issues apply to the current exporter as well.

Given that the "feature" is otherwise self-contained though, and extremely valuable to support, I actually do think a backport would be desirable, even if it was downstream.

daviddavis:
FYI, here's the bug around not handling multiple exports:

https://github.com/pulp/pulpcore/issues/1949

dralley:
Oh nice, didn't know that was already filed.

dralley:
@mdellweg Could you re-review with the new checks added?

Last questions: 

* bikeshedding the command name
* should anything else be done to address https://github.com/pulp/pulpcore/issues/1949 or can this be considered to close it as-is?

dralley:
Screenshot from 2022-08-03 23-46-19

pulpcore/app/management/commands/publication-fs-export.py
```diff
@@ -0,0 +1,163 @@
...
+
+        # Go through all the target directories first, if any of them are dirty, print warnings
+        # and exit - unless the user explicitly asked to go through with it anyway.
+        is_dirty = False
+        for dest, publication in to_export:
+            if not _export_location_is_clean(dest):
+                logging.error(
+                    _("Directory '{}' must be empty to be used as an export location").format(dest)
+                )
+                is_dirty = True
```
mdellweg, line 126: In case of `allow_dirty` we don't even need to run this check, do we?
If we still want to, the log message is misleading. It should sound like "Directory is not empty, but doing anyway as requested.".

mdellweg (commented):
LGTM, just a small nitpick about the wording for `allow_dirty`.

BTW we could call it `force-overwrite`.

pulpcore/app/management/commands/publication-fs-export.py
dralley: Done

dralley:
@dkliban @ggainey Do you have any thoughts on https://github.com/pulp/pulpcore/pull/2484#issuecomment-1204702789?

re: naming we can probably just mark it tech-preview and rename it later if we can't think of something better right away.

mdellweg (approved):

ggainey:
> @mdellweg Could you re-review with the new checks added?
> 
> Last questions:
> 
> * bikeshedding the command name

As I mentioned, this isn't an 'Exporter' in the sense we're using that name for PulpExporter and FilesystemExporter. I'd rather see if called `publication-fs-extract`, in order to make it clear that this is a different workflow.

> * should anything else be done to address [FilesystemExporter does not gracefully handle exporting multiple times #1949](https://github.com/pulp/pulpcore/issues/1949) or can this be considered to close it as-is?
>   
>   * e.g. as per the description, we could have a flag that would forcibly remove files if present, currently `--allow-dirty` doesn't do this and we would want it to be present on the exporter API -- if this is something we want to do.

I am a little confused about why we're linking that issue with this code - this *isn't* the FilesystemExporter, it's a different workflow. That issue def should be addressed in FSE, but it's independent from what we decide to do here.

dralley:
@ggainey We're using the same backend code for both workflows.  The user-facing aspect is different but what happens underneath is basically the same.

dralley:
@ggainey How about `export-publications-to-fs`

ggainey:
> @ggainey How about `export-publications-to-fs`
> 
> It's not so similar that it could pass for the same thing but it still gets the point across.
> 
> Or `dump-publications-to-fs` which is a step further.

I think I like the second one - makes it Utterly Clear what we're doing, yeah?

ggainey (approved):
I think we are good2go, thanks Daniel!

PR #2493, opened 2022-04-07
Create a single reusable data-repair command.

dralley:
[noissue]

### DIFF
--- a/pulpcore/app/management/commands/datarepair.py
+++ b/pulpcore/app/management/commands/datarepair.py
@@ -2,7 +2,7 @@ from gettext import gettext as _
 
 from django.db import connection
 from django.conf import settings
-from django.core.management import BaseCommand
+from django.core.management import BaseCommand, CommandError
 from django.db.models import Q
 from django.utils.encoding import force_bytes, force_str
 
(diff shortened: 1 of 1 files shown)


dralley:
@mdellweg @bmbouter  What do you think about this in principle?

mdellweg:
I do not see much difference. The biggest impact is probably that `pulpcore-manager --help` will only show a single command instead of a whole list.

Design questions that come to my mind:
* Do/can we require all data repair to honor the `--dry-run` flag?
* Do plugins need to provide their own version of this command, or can the plug in?
* Should we identify the issue by `<app_label>:<nr>` or `<gh-repo>:<nr>?

dralley:
>Do plugins need to provide their own version of this command, or can the plug in?

I was thinking plugins would provide their own.

> Do/can we require all data repair to honor the --dry-run flag?

I think we can probably strive to do that.  I'd be more concerned about the potential need for other flags.

>Should we identify the issue by <app_label>:<nr> or `:?

Since it would only be for pulpcore that's probably not necessary, but I suppose you're concerned about overlaps?

dralley:
@mdellweg Should this proceed or should it just be closed?

mdellweg:
I think, i would at least wait for a second data repair command. It might tell us the answer to whether more than `--dry-run` flags are likely and help design a common interface.

PR #2497, opened 2022-04-08
Loosened import version-restriction to x.y instead of x.y.z.

### LINKED ISSUE #2269: Version Matching for Import/Export too stringent
When importing a previously exported content pulp runs a check on pulpcore,  pulp_rpm and other plugin versions of both the export archive and currently installed system. For example if my exporter is using pulp 3.16.1 and import server 3.16.2 I 'd get an error like
`Export version 3.16.1 of pulpcore does not match installed version 3.16.2`

While checking compatibility is a good thing, we need to be mindful of the fact customers in disconnected environments will often have different minor releases of pulp. Expecting the z-stream to also match will make it unwieldy and very hard for us to control the environment.

Location of the check -> https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/importer.py#L93-L118

As described in https://hackmd.io/HLptudH9R6S4PCmm8nRmmg?view#Open-Questions

-   [DONE] How will backwards incompatible changes in the exported data format be handled over time? e.g. I exported this data a loooong time ago and now the system I'm importing into expects a newer, different format?
    -   idea 1: Tie the data format to the exported system's pulpcore version number. Put that version number into the exported data somehow. Then have systems importing know the oldest version they support and refuse to import older exports.
    -   idea 2: Same idea as (1) except use it's own numbering scheme, probably semver based
    -   **Current Status**: versions.json included in export, with core and plugin versions for all
        plugins involved in the export. Import checks for **exact match** to versions.json on import,
        and errors if there are diffs.

We definitely need more flexibility on this.

Discuss various approaches to solve this in both short  term and long term.

- Only require match of x.y releases and not x.y.z when checking compatibility (not sure that is sufficient).
-  As described in the design docs use a unique numbering/version  scheme that changes only when 2 versions are incompatible. That way if there are no breaking changes between 3.16-3.18 the version will be identical and hence compatible

**Additional context**
Further discussion pending on this topic. Pulpcore needs to be careful to not make it too stringent,

ggainey:
closes #2269

### DIFF
--- a/docs/workflows/import-export.rst
+++ b/docs/workflows/import-export.rst
@@ -37,22 +37,6 @@ In order to minimize space utilization, import/export operates on sets of
 :term:`Artifacts<Artifact>` only once per-export, rather than once for each
 :term:`Repository` being exported.
 
-.. note::
-
-    Export will not operate on :term:`RepositoryVersions<RepositoryVersion>` that have
-    been synchronized using ``policy=on_demand``. :term:`Artifacts<Artifact>` must actually
-    exist in order to be exported - this is, after all the only way for the Downstream Pulp
-    instance to gain access to them!
-
-.. note::
-
-    Import and Export strictly control which directories may be read from/written to via
-    the settings options ``ALLOWED_IMPORT_PATHS`` and ``ALLOWED_EXPORT_PATHS``.
-    These default to empty, if not explicitly set attempts to import or export will fail
-    with a validation error like
-
-        ``"Path '/tmp/exports/' is not an allowed export path"``
-
 Definitions
 ^^^^^^^^^^^
 Upstream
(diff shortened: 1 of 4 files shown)


docs/workflows/import-export.rst
```diff
@@ -87,6 +71,55 @@ Import order
...
+^^^^^^^^^^^
+
+The import/export workflow operates on a set of assumptions. Violating them will result
+in error-messages as described below.
+
+On-Demand content not supported
+-------------------------------
+
+Export will not operate on :term:`RepositoryVersions<RepositoryVersion>` that have
+been synchronized using ``policy=on_demand``. :term:`Artifacts<Artifact>` must actually
```
mdellweg, line 84: and "streamed"

pulpcore/app/tasks/importer.py
```diff
@@ -101,10 +101,10 @@ def _check_versions(version_json):
                 _("Export uses {} which is not installed.").format(component["component"])
             )
         else:
-            if version != component["version"]:
+            if version.split(".")[:2] != component["version"].split(".")[:2]:
```
ipanova, line 104: maybe this is a bit more readable? `version[:version.rfind(".")]`

ggainey: > maybe this is a bit more readable? `version[:version.rfind(".")]`

Can't use rfind() - dev-versions look like 3.19.0.dev. I ha dseveral attempts, they were all ugly. How about I just add comment describing what we're doing?

ipanova: I see. Then yeah leaving a lone line comment would do.

ipanova (approved):

ipanova (changes requested):
Actually,  I am removing my approval before @ggainey  starts to  yell at me. I will wait until suggestions are pushed :D

dralley:
Does this need backports to 3.15, 3.14?

dralley:
Adding this as a 3.19 blocker

pulpcore/app/tasks/importer.py
mdellweg: Also the logic is slightly different. With rfind we'd compare "all but the last segment", with split we compare "the two most significant ones".

ipanova: For the production installation both logic would be interchangeable.

ggainey:
> Does this need backports to 3.15, 3.14?

Currently only requested for 3.16 by stakeholder.

docs/workflows/import-export.rst
```diff
@@ -87,6 +71,55 @@ Import order
...
+The import/export workflow operates on a set of assumptions. Violating them will result
+in error-messages as described below.
+
+On-Demand content not supported
+-------------------------------
+
+Export will not operate on :term:`RepositoryVersions<RepositoryVersion>` that have
+been synchronized using ``policy=on_demand`` or ``policy=streamed``. i
+:term:`Artifacts<Artifact>` must actually exist in order to be exported - this is, after
+all the only way for the Downstream Pulp instance to gain access to them!
```
dralley, line 86: Strictly speaking this is only true if the downstream Pulp is disconnected.  I think that is probably the most common case though.

docs/workflows/import-export.rst
```diff
@@ -87,6 +71,55 @@ Import order
...
+export will fail with a RuntimeError:
+
+    ``Remote artifacts cannot be exported.``
+
+Export/Import Directories must be explicitly allowed
+----------------------------------------------------
+
+Import and Export strictly control which directories may be read from/written to via
+the settings options ``ALLOWED_IMPORT_PATHS`` and ``ALLOWED_EXPORT_PATHS``.
+These default to empty, if not explicitly set attempts to import or export will fail
```
dralley, line 98: This flows a little strangely, do you think this would be an improvement

> These default to empty - if they are not explicitly set, to import or export will fail

docs/workflows/import-export.rst
dralley: What I mean though is that there's no law that says you're only allowed to import / export if you have a disconnect systems.

ggainey: Export absolutely will fail if a piece of content doesn't have an Artifact, because the overarching use-case for this feature is "the downstream has no internet access and therefore requires all the Artifacts be imported".

dralley: I see now that there's some clarification about the use case a bit further up in the docs, but maybe we should make it clear that this isn't really intended to be a general backup solution either (along with not being related to pulp-to-pulp sync).

I just noticed a typo though.

> **i** :term:`Artifacts<Artifact>`

dralley (approved):

PR #2507, opened 2022-04-11
Fix a bug where workers miss notifications

### LINKED ISSUE #2506: All workers idle while a task is pending
**Version**
At least 3.14.14 used in katello

**Describe the bug**
Workers can all enter sleep state and miss the wakeup notification leading to pending tasks while all workers idle. Dispatching a new tasks unfreezes the system again.

**To Reproduce**
https://bugzilla.redhat.com/show_bug.cgi?id=2050440

**Expected behavior**
At least one worker is expected to pick up the next pending task.

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=2050440

mdellweg:
fixes #2506

### DIFF
--- a/CHANGES/2506.bugfix
+++ b/CHANGES/2506.bugfix
@@ -1,0 +1,2 @@
+Fixed a bug where notifications to workers may go unnoticed. This may lead to idle workers while
+there are tasks waiting.
--- a/pulpcore/tasking/pulpcore_worker.py
+++ b/pulpcore/tasking/pulpcore_worker.py
@@ -275,21 +275,24 @@ class NewPulpWorker:
         """Wait for signals on the wakeup channel while heart beating."""
 
         _logger.debug(_("Worker %s entering sleep state."), self.name)
+        wakeup = False
         while not self.shutdown_requested:
+            # Handle all notifications before sleeping in `select`
+            while connection.connection.notifies:
+                item = connection.connection.notifies.pop(0)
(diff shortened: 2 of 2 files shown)


mdellweg:
This is probably the easiest fix for this situation. But maybe we can do better.

dkliban:
@mdellweg, I don't understand this change. Can you please explain?

mdellweg:
> @mdellweg, I don't understand this change. Can you please explain?

The async postgres notifications are fetched on any database query, not only `connection.poll()` so it may happen that e.g. the saving of the heartbeat acidentally fetched a notification into the `connection.notifies` list, leaving the file descriptor invisible to `select`. Therefore we cannot rely on the fact that all notifications will trigger the connection to pop up in the result of `select`, And we need to check for the existence of notifications outside of the `if connection in r:` condition.

mdellweg:
I'll revise the notification handling in the loop once more.

dkliban (approved):

bmbouter (approved):
Thank you!

mdellweg:
I'll take care of the failed cherry-picks.

mdellweg:
https://github.com/pulp/pulpcore/pull/2521

mdellweg:
3.15: https://github.com/pulp/pulpcore/pull/2522

mdellweg:
3.14: https://github.com/pulp/pulpcore/pull/2523

PR #2510, opened 2022-04-11
Bump django on 3.17

fao89:
https://www.djangoproject.com/weblog/2022/apr/11/security-releases/


bmbouter (approved):

fao89:
> I think, we need to pin the branches of pulp_file on these releases.

you were right! thank you for the tip!

Issue #1910, opened 2022-01-17
Accessing content for a distribution that is not assigned returns 404 which is confusing


pulpbot:
Author: daviddavis (daviddavis)

Redmine Issue: 7029, https://pulp.plan.io/issues/7029

---

Today dkliban and I helped debug some 404 errors a user was hitting when trying to access their distribution through the content app. Turns out that distribution was not assigned to a publication. 404 is perhaps not the right error code (since the distribution exists) and also it would be more helpful to somehow alert the user that their distribution is not assigned.

pulpbot:
From: @dkliban (dkliban@redhat.com)
Date: 2020-06-23T14:35:20Z

---

The server should return a 404 with an error message. E.g. 404: The distribution doesn't have a publication or repository version associated with it.

pulpbot:
From: @dkliban (dkliban@redhat.com)
Date: 2020-12-16T15:30:46Z

---

If the user requests the root of a distribution that doesnt' have a publication or repo version assigned to it, the user should receive a page that states that nothing is assigned to this distribution.

PR #2515, opened 2022-04-11
Add error reason for accessing distribution w/o publication

gerrod3:
fixes: #1910

Working on a couple content app user improvements for 3.20. This one is the simplest change, but maybe the issue wanted something more explicit for this scenario.

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -623,7 +623,7 @@ class Handler:
                     request, StreamResponse(headers=headers), ra
                 )
 
-        raise PathNotResolved(path)
+        raise PathNotResolved(path, reason=_("Distribution is not pointing to a publication."))
 
     async def _stream_content_artifact(self, request, response, content_artifact):
         """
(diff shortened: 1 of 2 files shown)


dralley (approved):

fao89 (approved):

mdellweg (approved):

pulpcore/content/handler.py
```diff
@@ -623,7 +623,7 @@ def get_remote_artifact_blocking():
                     request, StreamResponse(headers=headers), ra
                 )
 
-        raise PathNotResolved(path)
+        raise PathNotResolved(path, reason=_("Distribution is not pointing to a publication."))
```
ipanova, line 626: Will the message still be accurate for the cases when the distribution is meant  to point to a repo/repo-version?

ipanova: maybe this should be adjusted to 'Distribution is not pointing to a publication or repository/repository version'?

PR #2558, opened 2022-04-19
Fix does_batch for resolved content

### LINKED ISSUE #2557: Implementation of does_batch in DeclarativeContent is wrong
**Describe the bug**
`does_batch` is supposed to decide, whether content can be waited on, or must be evaluated asap. It produces `False` unnecessarily often, leading to performance issues.

**Additional context**
This is suspected to at least one of the reasons for: https://github.com/pulp/pulp_deb/issues/466

mdellweg:
This bug lead to serious sync performance degredation in some plugins,
because batching was de facto disabled for the ContentAssociation stage.

fixes #2557

### DIFF
--- a/CHANGES/2557.bugfix
+++ b/CHANGES/2557.bugfix
@@ -1,0 +1,1 @@
+Fixed `does_batch` method in sync pipeline to allow waiting on content that is already resolved.
--- a/pulpcore/plugin/stages/models.py
+++ b/pulpcore/plugin/stages/models.py
@@ -136,7 +136,7 @@ class DeclarativeContent:
         """Whether this content is being awaited on and must therefore not wait forever in batches.
         When overwritten in subclasses, a `True` value must never be turned into `False`.
         """
-        return not self._resolved and self._future is None
+        return self._resolved or self._future is None
 
     async def resolution(self):
         """Coroutine that waits for the content to be saved to database.


quba42:
On my pulplift box, this fix appears to have reduced the sync time for a mid sized APT repo (ca. 10000 packages) using on_demand sync from 10 minutes to 2 minutes.

I will need to test bigger repos, to check if this also fixes https://github.com/pulp/pulp_deb/issues/466

quba42:
I am pretty sure this fixes https://github.com/pulp/pulp_deb/issues/466. Large test syncs are still running, but they have already achieved many times the number of associated content in less time than previous test runs without the fix.

sbernhard:
I have adapted https://github.com/pulp/pulpcore/blob/5a2df085dd404206c5d84a74f697d1242022379f/pulpcore/plugin/stages/api.py#L152 a little bit to 
`log.debug(
                    _("%(name)s - next batch[%(length)d] - noblock: %(noblock)s."), {"name": self, "length": len(batch), "noblock": no_block }
                )` 
and set the minsize (batchsize) to 50 instead of 500.

Before the change of this PR:
`[139876820437200] ContentAssociation - next batch[2] - noblock: True.`

Afterwards:
`[140324667339160] ContentAssociation - next batch[50] - minsize: 50 - noblock: False.`

quba42:
I would like to see this fix back-ported to at least pulpcore 3.16 (Katello 4.3) if possible.

quba42:
Thanks a bunch!

bmbouter (approved):
This is a huge improvement. Great job @mdellweg !

dralley (approved):

dralley:
Really nice improvement, thanks @mdellweg (and @quba42 and @gvde)

I tested with the RPM plugin and it didn't make much of a difference, but I wouldn't expect it to, we aren't doing very much future resolution between content in the batch.

PR #2560, opened 2022-04-19
Remove interface for unregistered creation_hooks

### LINKED ISSUE #2559: Remove deprecated creation_hooks interface for 3.20

mdellweg:
fixes #2559

### DIFF
--- a/pulpcore/app/access_policy.py
+++ b/pulpcore/app/access_policy.py
@@ -1,6 +1,6 @@
 from rest_access_policy import AccessPolicy
+from rest_framework.exceptions import APIException
 
-from pulpcore.app.loggers import deprecation_logger
 from pulpcore.app.models import AccessPolicy as AccessPolicyModel
 from pulpcore.app.util import get_view_urlpattern, get_viewset_for_model
 from pulpcore.app.role_util import get_objects_for_user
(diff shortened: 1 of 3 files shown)


pulpcore/app/access_policy.py
```diff
@@ -48,19 +48,16 @@ def handle_creation_hooks(cls, obj):
...
-                    deprecation_logger.warn(
-                        "Calling unregistered creation hooks from the access policy is deprecated"
-                        " and may be removed with pulpcore 3.20."
-                        f"[hook={creation_hook}, viewset={access_policy.viewset_name}]."
+                hook_name = creation_hook["function"]
+                try:
+                    function = obj.REGISTERED_CREATION_HOOKS[hook_name]
+                except KeyError:
+                    raise APIException(
+                        f"Creation hook '{hook_name}' was not registered for this view set."
```
bmbouter, line 56: Is there some way to test this part of the codepath?

mdellweg: Hmm you can define an accesspolicy with a non existing creation hook.
I added this only because the raw KeyError from failing to access that dict would have been rather hard to decipher.

mdellweg: What kind of validation are you referring to?

mdellweg: If we'd validate the Accesspolicy to that point, this key error should not even be reachable without a plugin removing a creation_hook from one version to another.

bmbouter (approved):
This looks great. We can talk about adding validation later, but that's unrelated to this PR so LGTM. Thank you!

gerrod3 (approved):

PR #2562, opened 2022-04-19
Add pulpcore pytest fixtures from pulp-smash

bmbouter:
The pulpcore fixtures are moving from pulp-smash to pulpcore so that
they can be versioned properly with pulpcore.

Required PR: https://github.com/pulp/pulp_file/pull/687

[noissue]

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -1,1 +1,0 @@
-"""Tests for Pulp 3's core."""
--- a/pulpcore/tests/functional/conftest.py
+++ b/pulpcore/tests/functional/conftest.py
@@ -1,0 +1,137 @@
+import uuid
+
+import pytest
+
+from pulp_smash.config import get_config
+from pulp_smash.pulp3.bindings import delete_orphans
+
+from pulpcore.client.pulpcore import (
+    ApiClient,
+    ContentguardsApi,
+    DistributionsApi,
+    PublicationsApi,
+    RepositoriesApi,
+    StatusApi,
+    TasksApi,
+    TaskSchedulesApi,
+    UsersApi,
+    UsersRolesApi,
+)
+
+
+@pytest.hookimpl(tryfirst=True)
+def pytest_check_for_leftover_pulp_objects(config):
+    cfg = get_config()
+    pulpcore_client = ApiClient(cfg.get_bindings_config())
+    tasks_api_client = TasksApi(pulpcore_client)
+
+    for task in tasks_api_client.list().results:
+        if task.state in ["running", "waiting"]:
+            raise Exception("This test left over a task in the running or waiting state.")
+
+    apis_to_check = [
+        ContentguardsApi(pulpcore_client),
+        DistributionsApi(pulpcore_client),
+        PublicationsApi(pulpcore_client),
+        RepositoriesApi(pulpcore_client),
+    ]
+    for api_to_check in apis_to_check:
+        if api_to_check.list().count > 0:
+            raise Exception(f"This test left over a {api_to_check}.")
+
+
+@pytest.fixture(scope="session")
+def pulpcore_client(bindings_cfg):
+    return ApiClient(bindings_cfg)
+
+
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -40,6 +40,7 @@
         "console_scripts": [
             "pulpcore-manager = pulpcore.app.manage:manage",
             "pulpcore-worker = pulpcore.tasking.entrypoint:worker",
-        ]
+        ],
+        "pytest11": ["pulpcore = pulpcore.tests.functional"],
```
mdellweg, line 44: Let's not make this a pytest plugin but load the fixtures explicitely.

pulpcore/tests/functional/__init__.py
```diff
@@ -1 +1,111 @@
...
+    ApiClient,
+    StatusApi,
+    TasksApi,
+    TaskSchedulesApi,
+    UsersApi,
+    UsersRolesApi,
+)
+
+
+@pytest.fixture(scope="session")
```
mdellweg, line 17: This should probably go into conftest.

mdellweg (changes requested):

setup.py
bmbouter: Yes that is what we talked about yesterday. After thinking it over some though, having 100% of plugins have to add a line when we could do it here once didn't seem like the best option. That being said, we may have to do it as you suggest because the CI is failing due to the unexpected import-time reaching its way to pulp-smash. Perhaps what you're suggesting will resolve that actually. I'm continuing to work on it and once I know more of which options could work we can talk it over some.

bmbouter:
@mdellweg do you think I should also move the --no-pulp-leftovers flag here too? I think I should, but I had some practical issues doing so, so I started moving things in parts in this PR. Anyways, that feature 100% uses pulpcore bindings so I thought moving it made sense. What do you think?

setup.py
mdellweg: It's more than that one line repeated on all plugins. I expect pulp_file and pulp_certguard to require themselves in a similar way. And at that point i value the expliciteness of saying "we need those fixtures".

mdellweg:
> @mdellweg do you think I should also move the --no-pulp-leftovers flag here too? I think I should, but I had some practical issues doing so, so I started moving things in parts in this PR. Anyways, that feature 100% uses pulpcore bindings so I thought moving it made sense. What do you think?

That feature is implemented as a hook. So we could keep the flag with the hook definition in smash and just move the hook implementation using pulpcore facilities here.

pulpcore/tests/functional/conftest.py
```diff
@@ -0,0 +1 @@
+pytest_plugins = "pulpcore.tests.functional"
```
bmbouter, line 1: Kind of ironic, but it needed this too. :rofl:

mdellweg: Not if we put the fixtures in conftest.py i'd say.
Fixtures can either live in the _same_ python module, or on any conftest.py up in the directory tree.

If conftest.py does not work for inclusion of the fixtures in other plugins, can we have them in `pulpcore.tests.functional.fixtures`?

bmbouter: If I'm reading your words right, you're suggesting I move everything from `__init__.py` to `conftest.py` in this PR and then have plugins declare their dependency with `pytest_plugins = "pulpcore.tests.functional.conftest"`. Is that right?

If ^ is right, the only downside is that now all plugins dependent on pulpcore will receive everything in conftest, but perhaps that's always correct. I can't think of a case where that isn't true.

I wanted to check before I go reworking the PRs once more. Ty!

mdellweg: Assuming, that it actually works, that is what i'm suggesting. Using the `...functional.fixtures` module would be my second preferred solution.
One interesting question then is still, whether plugins will receive the hook implementations alongside the fixtures (pytest docs are not so explicit about that).

mdellweg (approved):

dkliban (approved):

PR #2571, opened 2022-04-20
Ensure download has fresh file on retry

gerrod3:
[noissue]

Not sure if this has an issue already or not. Related to https://issues.redhat.com/browse/AAH-1202 and test found in: https://github.com/pulp/pulp_file/pull/689

Our downloaders really need to be refactored and will be (https://discourse.pulpproject.org/t/downloader-refactor-sig/393/4), so not sure how useful this fix is. Also, I only fixed this on the http downloader, base and file are unaffected, and if any other downloaders override the `_run` method then this fix won't occur.

### DIFF
--- a/CHANGES/2576.bugfix
+++ b/CHANGES/2576.bugfix
@@ -1,0 +1,1 @@
+Ensure downloader resets file on retry.
--- a/pulpcore/download/http.py
+++ b/pulpcore/download/http.py
@@ -272,6 +272,7 @@ class HttpDownloader(BaseDownloader):
         Args:
             extra_data (dict): Extra data passed by the downloader.
         """
+        self._ensure_fresh_file()
         if self.download_throttler:
             await self.download_throttler.acquire()
         async with self.session.get(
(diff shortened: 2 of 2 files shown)


fao89 (commented):
I feel it needs a changelog entry

mdellweg (changes requested):
Thank you! I must have been dumb. I reset both the size counter and the hashes, but i forgot about the actual data. :facepalm: 
But we still need an issue and a changelog.

bmbouter (approved):
+1 to the comments from @mdellweg the issue and changelog are needed. @mdellweg can be the final lgtm. Thanks @gerrod3 this is great!

I'll review the test separately.

dralley (approved):

mdellweg (approved):

ipanova:
@gerrod3 given that most of the plugins override _run() this change will not fix https://issues.redhat.com/browse/AAH-1202 unless we make a direct change in ansible and containers plugins, if i read things correctly.

gerrod3:
OK I checked ansible, rpm and container and they all override the `_run()` method so this fix won't fix the issue for their plugins. However, all the plugins don't override the `run()` method, so I think I can just move the call to the new method, `_ensure_fresh_file()` to be inside `run()` and then the fix should apply to all plugins.

dralley:
@gerrod3 Add this to the list of things we'd really like the downloader refactor to improve if possible.

PR #2577, opened 2022-04-21
Removes il8n from log messages and adds docs

### LINKED ISSUE #2477: Log should not be i18n
Currently logs are being translated (passed by gettext `_`) and that causes log messages to be internationalizable and that is a problem.

## Statement

The audience for logs are developers, system administrators, auditors, testers and as the main project documentation + code are in English we expect logs to be consistent.

Logging analysis tools would expect log levels (ERROR, CRITICAL, DEBUG, INFO) etc to be in english and there are some tools that might do some deeper linguistic analysis on logs.

## Bad scenario

Assuming a system that is installed for a Brazilian company and the whole system messages on the front-ends (API, UI, CLI) are well translated, and that is expected!

Now on that system there is a problem and the user open an issue, the developer or tester asks the user to upload the latest log file and the developers, testers, administrators are going to get messages like:

```log
[ERRO CRITICO]  Não foi possível efetuar a operação desejada, a entidade requisitada não foi encontrada.
```

Or if the system is in chinese

```log
[严重错误] 无法执行所需的操作，未找到请求的实体。
```

As we don't have a table of standard error codes, it will be very difficult and costly for support to troubleshoot the issue.

It is better to keep in English as it is the standard lang for I.T support.

```log
[CRITICAL]  The desired operation could not be performed, the requested entity was not found.
```

## Conclusion

Only front-ends must be translated.

bmbouter:
il8n should only be for using facing strings. This updates the code as
best as I could tell the difference, and also documents that expectation
for plugin writers.

closes #2477

### DIFF
--- a/CHANGES/2477.bugfix
+++ b/CHANGES/2477.bugfix
@@ -1,0 +1,2 @@
+Removed il8n from the logs written so they will always show up in English for speedy resolution of
+error messages. All user facing strings are still expected to be il8n.
--- a/docs/plugins/plugin-writer/concepts/index.rst
+++ b/docs/plugins/plugin-writer/concepts/index.rst
@@ -516,3 +516,19 @@ is allowed to handle. This includes two types of "checksum handling":
     requested that is not listed in the ``ALLOWED_CONTENT_CHECKSUMS`` setting. This is a convenience
     facility allowing plugin writers to not check the ``ALLOWED_CONTENT_CHECKSUMS`` setting
     themselves.
+
+
+.. _il8n-expectations:
+
+Internationalization Expectations
+---------------------------------
+
+pulpcore and its plugins are expected to internationalize all user-facing strings using Python's
+gettext facilities. This allows Pulp to be translated to other languages and be more usable for a
+broader base of users.
+
+Administrator facing strings are expected *not* to be internationalized. These include all log
+statements, migration output print statements, django management commands, etc. These not being
+internationalized will remain in English. This expectation was formed after feedback from
+multi-language speakers who believe having error messages for admins in English would reduce the
+time to finding a fix and was generally less surprising.
--- a/pulpcore/app/management/commands/handle-artifact-checksums.py
+++ b/pulpcore/app/management/commands/handle-artifact-checksums.py
@@ -27,7 +27,7 @@ class Command(BaseCommand):
     Django management command for populating or removing checksums on artifacts.
     """
 
-    help = _("Handle missing and forbidden checksums on the artifacts")
+    help = "Handle missing and forbidden checksums on the artifacts"
 
     def add_arguments(self, parser):
         parser.add_argument("--report", action="store_true")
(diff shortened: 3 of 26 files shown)


CHANGES/2477.bugfix
```diff
@@ -0,0 +1,3 @@
+Removed i8ln from the logs written so they will always show up in English. It's a debatable point,
+but after feedback from multi-language speakers, since the logs aren't user facing, having them
+untranslated is best for everyone. All user facing strings are still expected to be i8ln.
```
lubosmj, line 3:
```suggestion
Removed i18n from the logs written so they will always show up in English. It's a debatable point,
but after feedback from multi-language speakers, since the logs aren't user facing, having them
untranslated is best for everyone. All user facing strings are still expected to be i18n.
```

It is "i18n", not "i8ln" (see https://docs.python.org/3/library/gettext.html).

Also, I do not like the idea of expressing a subjective judgment in a changelog entry (`It's a debatable point`). The reasoning should be rather advertised in a release announcement on discourse.

mdellweg: Good call. We did that change. We state so. There is no need to argue anymore.

bmbouter: I agree, I've revised this.

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -516,3 +516,19 @@ is allowed to handle. This includes two types of "checksum handling":
...
+Internationalization Expectations
+---------------------------------
+
+pulpcore and its plugins are expected to internationalize all user-facing strings using Python's
+gettext facilities. This allows Pulp to be translated to other languages and be more usable for a
+broader base of users.
+
+Administrator facing strings are expected *not* to be internationalized. These include all log
+statements, migration output print statements, django management commands, etc. These not being
+internationalized will remain in English. This expectation was was formed after feedback from
```
lubosmj, line 532:
```suggestion
internationalized will remain in English. This expectation was formed after feedback from
```

mdellweg, line 532:
```suggestion
internationalized will remain in English. This expectation was formed after feedback from
```

pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -27,7 +27,7 @@ class Command(BaseCommand):
     Django management command for populating or removing checksums on artifacts.
     """
 
-    help = _("Handle missing and forbidden checksums on the artifacts")
+    help = "Handle missing and forbidden checksums on the artifacts"
```
lubosmj, line 30: I am not quite sure if this is the type of string we do not want to translate.

mdellweg (changes requested):

pulpcore/app/management/commands/handle-artifact-checksums.py
bmbouter: Yeah it's generally pretty confusing. This one faces the administrator (not the user) so I left it un-il8n.

docs/plugins/plugin-writer/concepts/index.rst
bmbouter: @mdellweg I fixed this if you could re-review plz. Thanks!

mdellweg (approved):

gerrod3 (approved):
Looks good. Looking at our logs we format our strings using `format` and `%`, but no `f strings`. :disappointed: . Don't have to change them now, but we can finally start to use `f strings`!

bmbouter:
> Looks good. Looking at our logs we format our strings using `format` and `%`, but no `f strings`. disappointed . Don't have to change them now, but we can finally start to use `f strings`!

I agree!

PR #2586, opened 2022-04-22
Add REDIRECT_TO_OBJECT_STORAGE setting

### LINKED ISSUE #2537: Streaming of content from cloud storage (configurable)
In some deployments the cloud storage configured for django-storages can be configured to be api only without the possibility to generate preauthenticated redirect links.
We need to add a configuration option to allow pulp-content to stream artifact data from such storage backends instead of redirecting.

mdellweg:
This allows artifacts to be stored in cloud storage while still serving
the actual data from Pulp's content app.

fixes #2537


CHANGES/2537.feature
```diff
@@ -0,0 +1,2 @@
+Added the setting ``REDIRECT_TO_OBJECT_STORAGE`` to allow using cloud storage with or without
```
ipanova, line 1: This needs a line in docs too.

pulpcore/content/handler.py
```diff
@@ -781,6 +782,13 @@ async def _serve_content_artifact(self, content_artifact, headers, request):
             if not os.path.exists(path):
                 raise Exception(_("Expected path '{}' is not found").format(path))
             return FileResponse(path, headers=headers)
+        elif not settings.REDIRECT_TO_OBJECT_STORAGE:
+            response = StreamResponse(headers=headers)
+            await response.prepare(request)
+            async for chunk in sync_to_async_iterable(artifact_file.chunks()):
```
ipanova, line 788: I believe you need to close the file after iterating through its chunks.

pulpcore/responses.py
```diff
@@ -0,0 +1,154 @@
...
+
+class ArtifactResponse(StreamResponse):
+    """A response object can be used to send artifacts."""
+
+    # Heavily borrowed from FileResponse
+
+    def __init__(
+        self,
+        artifact=None,
+        artifact_pk=None,
```
bmbouter, line 23: Am I reading correctly that one of these should be required? If so it would be nice in `__init__` to raise an exception if both are `None`.

pulpcore/app/models/storage.py
```diff
@@ -97,6 +97,31 @@ def _save(self, name, content, max_length=None):
         return str(name).replace("\\", "/")
 
 
+try:
+    from storages.backends.sftpstorage import SFTPStorage
+
+    class PulpSFTPStorage(SFTPStorage):
```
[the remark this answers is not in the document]
bmbouter: It's debatable, but I think we should. We even have a CI runner for it. Ultimately we want to do more, and if it's not working in ways we don't know folks can tell us and it'll improve.

pulpcore/responses.py
bmbouter: I'm still uneasy on using `assert` just because if it's providing a correctness aspect (even a small one like this) I don't think a compiler option should be able to turn it off. Having `assert` be "turn off-able" is one of the silliest design decisions in Python, but given that's the state of things I don't think we should be using it. An if then raise would be more appropriate (to me). This is my opinion, feel free to debate or ignore. We don't have a project-wide policy on this.

bmbouter (approved):
This looks great. Thank you for the docs too.

pulpcore/app/models/storage.py
```diff
@@ -97,6 +97,31 @@ def _save(self, name, content, max_length=None):
...
+        """Upstream class is missing the file_overwrite option."""
```
gerrod3, line 104: What does this mean?

pulpcore/content/handler.py
```diff
@@ -7,13 +7,19 @@
 
 from aiohttp.client_exceptions import ClientResponseError
 from aiohttp.web import FileResponse, StreamResponse, HTTPOk
-from aiohttp.web_exceptions import HTTPForbidden, HTTPFound, HTTPNotFound
+from aiohttp.web_exceptions import (
+    HTTPForbidden,
+    HTTPFound,
+    HTTPNotFound,
+)
```
gerrod3, line 14: Did black format this?

pulpcore/responses.py
```diff
@@ -0,0 +1,156 @@
...
+
+        await loop.run_in_executor(None, fobj.seek, offset)
+
+        chunk = await loop.run_in_executor(None, fobj.read, min(self._chunk_size, count))
+        while chunk:
+            await writer.write(chunk)
+            count = count - self._chunk_size
+            if count <= 0:
+                break
+            chunk = await loop.run_in_executor(None, fobj.read, min(self._chunk_size, count))
```
gerrod3, line 54:
```suggestion
        while chunk := await loop.run_in_executor(None, fobj.read, min(self._chunk_size, count)):
            await writer.write(chunk)
            count = count - self._chunk_size
            if count <= 0:
                break
```
Is this possible?

pulpcore/content/handler.py
```diff
@@ -781,6 +787,8 @@ async def _serve_content_artifact(self, content_artifact, headers, request):
             if not os.path.exists(path):
                 raise Exception(_("Expected path '{}' is not found").format(path))
             return FileResponse(path, headers=headers)
+        elif not settings.REDIRECT_TO_OBJECT_STORAGE:
+            return ArtifactResponse(content_artifact.artifact, headers=headers)
```
gerrod3, line 791: Have we tested this option with S3, Azure and SFTP storages?

pulpcore/app/models/storage.py
mdellweg: The upstream class would generate a new path instead of overwriting an existing file. The same stands for the standard file storage. Boto and Azure modules behave different in that respect.

pulpcore/content/handler.py
mdellweg: I tested locally with minio (s3), but the real test here is will be that the sftp storage has no HTTP channel we could ever redirect to.

PR #2588, opened 2022-04-22
Make pulpcore a pytest plugin ... once again!

bmbouter:
Required PR: https://github.com/pulp/pulp_file/pull/694

[noissue]


mdellweg:
Alternative approach:
https://github.com/pulp/plugin_template/pull/625

Also i would not make part of the tests file-tree the pytest plugin.

.github/workflows/scripts/install.sh
```diff
@@ -20,6 +20,7 @@ if [[ "$TEST" = "docs" || "$TEST" = "publish" ]]; then
   pip install -r doc_requirements.txt
 fi
 
+pip install -e ../pulpcore
```
fao89, line 23: on upgrade tests, it is installing pulpcore 3.11 here (which requires click 7.1)
https://github.com/pulp/pulpcore/blob/3.11/requirements.txt#L6

bmbouter: I think I'd opt for the pip installing pulpcore before checking out 3.11. I'm a little unsure on where to put that line though, could you show a diff possibly?

fao89: before these lines: https://github.com/pulp/pulpcore/blob/main/.github/workflows/scripts/before_install.sh#L33-L35

bmbouter: This is problematic because I need it to also be after these lines https://github.com/pulp/pulpcore/blob/main/.github/workflows/scripts/before_install.sh#L67-L84

fao89: so you need 
```
cd pulpcore
git checkout main
pip install -e .
cd ..
```
here: https://github.com/pulp/pulpcore/blob/main/.github/workflows/scripts/before_install.sh#L33-L35

bmbouter: This is just what I needed, thank you!

dkliban (approved):

dralley (approved):

ggainey (approved):
lg2m!

.github/workflows/scripts/install.sh
fao89: after reinstalling pulpcore you need:
```
# Hack: adding pulp CA to certifi.where()
CERTIFI=$(python -c 'import certifi; print(certifi.where())')
cat /usr/local/share/ca-certificates/pulp_webserver.crt | sudo tee -a "$CERTIFI" > /dev/null
```

fao89: I'm gonna merge it and fix in another PR

fao89: https://github.com/pulp/pulpcore/pull/2604

PR #2590, opened 2022-04-22
Taught pulp-import to retry up to three times on collisions.

### LINKED ISSUE #2589: Concurrent PulpImports with iverlapping  content can still fail.
**Version**
core/3.14

**Describe the bug**
PulpImport imports repositories in parallel. If the repos in a single import-file have overlapping content, two repo-import-threads can attempt to create the same object at the same time, and collide. This is handled by retry-logic, that notices the collision and retries in the failing task.

HOWEVER - if "enough" repositories, with "enough" overlapping content, are imported in a system with "enough" available pulpcore-workers, then **more than two** tasks can collide. The first one "wins", the second fails, retries, and "wins", the third fails *twice* and aborts.  We need to let PulpImport retry more than once.

Note that there needs to be a finite number of retries, because it's better to eventually give up, than to fall into an infinite retry-hole due to some equally-rare edge case.

**To Reproduce**

See discussion in https://bugzilla.redhat.com/show_bug.cgi?id=2056702 . We have seen this happen to exactly one user to date. We have not been able to reproduce it "in captivity" even once (but the hole clearly exists).

**Expected behavior**
PulpImport should retry an import more than once, in the event of an object-creation collision.

**Additional context**
Original collision-problem-report : https://pulp.plan.io/issues/8633
Retry fix : https://github.com/pulp/pulpcore/pull/1278

ggainey:
fixes #2589.
[nocoverage]

### DIFF
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -44,6 +44,10 @@ CA_FILE = "pulpcore.app.modelresource.ContentArtifactResource.json"
 VERSIONS_FILE = "versions.json"
 CONTENT_MAPPING_FILE = "content_mapping.json"
 
+# Concurrent imports w/ overlapping content can collide - how many attempts are we willing to
+# make before we decide this is a fatal error?
+MAX_ATTEMPTS = 3
+
 
 def _destination_repo(importer, source_repo_name):
     """Find the destination repository based on source repo's name."""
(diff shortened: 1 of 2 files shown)


pulpcore/app/tasks/importer.py
```diff
@@ -44,6 +44,10 @@
 VERSIONS_FILE = "versions.json"
 CONTENT_MAPPING_FILE = "content_mapping.json"
 
+# Concurrent imports w/ overlapping content can collide - how many times are we willing to
+# retry if we notice it happening?
+MAX_RETRIES = 3
```
ggainey, line 49: We could expose this as a config-setting - but the problem is **rare** , and it didn't/doesn't feel to me like a knob that anyone should ever touch.

pulpcore/app/tasks/importer.py
```diff
@@ -62,26 +66,29 @@ def _import_file(fpath, resource_class, retry=False):
...
+                    # importing the same 'thing' (e.g., a Package that exists in two different
+                    # repo-versions that are being imported at the same time). If we're asked to
+                    # retry, we will try an import that will simply record errors as they happen
+                    # (rather than failing with an exception) first. If errors happen, we'll
+                    # retry before we give up on this repo-version's import.
+                    a_result = resource.import_data(data, raise_errors=False)
+                    if a_result.has_errors():
+                        total_errors = a_result.totals["error"]
+                        log.info(
+                            f"...{total_errors} import-errors encountered importing "
```
ggainey, line 82: I'm changing error-messages here, and un-I18N'd them since that's the direction we're heading.

pulpcore/app/tasks/importer.py
dralley: If this is internal only I'm inclined to reduce the complexity by just calling it `MAX_TRIES`, and not needing the `- 1` below.

pulpcore/app/tasks/importer.py
```diff
@@ -62,26 +66,29 @@ def _import_file(fpath, resource_class, retry=False):
...
+                            "{fpath}, attempt {curr_retry}, retrying"
```
dralley, line 83: Particularly because it causes this error message to be off by one.  attempt != retry

pulpcore/app/tasks/importer.py
ggainey: Good call - I didn't like what I was doing, but branes couldn't werk to find a better way :)

pulpcore/app/tasks/importer.py
```diff
@@ -62,26 +66,29 @@ def _import_file(fpath, resource_class, retry=False):
...
-                    )
-                    # Second attempt, we raise an exception on any problem.
-                    # This will either succeed, or log a fatal error and fail.
-                    try:
-                        a_result = resource.import_data(data, raise_errors=True)
-                    except Exception as e:  # noqa log on ANY exception and then re-raise
-                        log.error(_("FATAL import-failure importing {}").format(fpath))
-                        raise
+                # Last attempt, we raise an exception on any problem.
+                # This will either succeed, or log a fatal error and fail.
```
dralley, line 86: You know I might have been wrong about the off-by-one thing, I think since this happens after all the attempts - that means you want the loop to handle `MAX_ATTEMPTS - 1` so that this last check completes it...

Sorry about that.

pulpcore/app/tasks/importer.py
```diff
@@ -62,26 +66,29 @@ def _import_file(fpath, resource_class, retry=False):
...
+                try:
+                    a_result = resource.import_data(data, raise_errors=True)
+                except Exception as e:  # noqa log on ANY exception and then re-raise
+                    log.error(f"FATAL import-failure importing {fpath}")
```
dralley, line 90: Should this message match the one on line 96?

pulpcore/app/tasks/importer.py
ggainey: No worries - calling it MAX_ATTEMPTS is clearer, starting at "1" for the first attempt makes it clearer yet.

pulpcore/app/tasks/importer.py
ggainey: Not the same - this exception is "we couldn't load the file AT ALL for some reason", the other ones are "file was loaded, but the import has failed". I modified for clarity.

dralley (approved):

mdellweg (approved):

PR #2602, opened 2022-04-25
Ensure download retry fix occurs for all http subclasses

gerrod3:
[noissue]

### DIFF
--- a/pulpcore/download/http.py
+++ b/pulpcore/download/http.py
@@ -251,6 +251,7 @@ class HttpDownloader(BaseDownloader):
                 giveup=http_giveup_handler,
             )
             async def download_wrapper():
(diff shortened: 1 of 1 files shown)


mdellweg (approved):
Nice solution. Maybe the name is a little bit confusing. It ensures there is _no_ file. so that `_ensure_writer_has_open_file` will be triggered later.
How about something like `_scrape_broken_download`?

ipanova:
Should this reference the original issue that has changelog or we should at least set the backport labels so this an previous commit go together in one release.
regarding the name we could also name this _ensure_no_broken_file

dralley (approved):

ipanova (approved):

Issue #2573, opened 2022-04-21
Chunked Uploads doesn't works


x23n5902y:
**Version**
```
$ pip list | grep pulp
pulp-cli                        0.14.0
pulp-deb                        2.17.0
pulp-file                       1.10.2
pulp-rpm                        3.17.4
pulpcore                        3.18.1
```

**Describe the bug**
When trying to upload a large file to Pulp, I get an OOM on the pulp-worker process.
When I try to upload a file in parts like this
```
ISO_IMAGE_URL="https://mirror.yandex.ru/almalinux/8.5/isos/x86_64/AlmaLinux-8.5-x86_64-dvd.iso"
ISO_IMAGE=$(echo $ISO_IMAGE_URL | grep -Po '(?:[^\/]+)$')
CHUNK_PREFIX="alma_"
CHUNK_SIZE="32M"
SIZE=$(du -b ${ISO_IMAGE} | cut -f1)
SHA256=$(sha256sum ${ISO_IMAGE} | cut -f1)

curl -O ${ISO_IMAGE_URL}
split --bytes=${CHUNK_SIZE} ${ISO_IMAGE} ${CHUNK_PREFIX}
export UPLOAD=$(http POST :24817/pulp/api/v3/uploads/ size=${SIZE} | jq -r '.pulp_href')
cs=0
for chunk in $(ls -1 ${CHUNK_PREFIX}* | sort)
    do
        csn=$(($cs + $(du -b ${chunk} | cut -f1) ))
        http --form PUT :24817${UPLOAD} file@./${chunk} 'Content-Range: bytes '${cs}-$((${csn} - 1))'/*'
        cs=$csn
    done

http POST :24817${UPLOAD}commit/ sha256=${SHA256}
```
I get an error

```
]$ pulp task show --href "/pulp/api/v3/tasks/a440388c-084c-4756-b223-0bce568e89ab/"
{
  "pulp_href": "/pulp/api/v3/tasks/a440388c-084c-4756-b223-0bce568e89ab/",
  "pulp_created": "2022-04-21T14:15:54.580779Z",
  "state": "failed",
  "name": "pulpcore.app.tasks.upload.commit",
  "logging_cid": "cb9be1d7de5d4c76981866bef3d7e5f4",
  "started_at": "2022-04-21T14:15:54.745562Z",
  "finished_at": "2022-04-21T14:16:43.654643Z",
  "error": {
[...]

PR #2620, opened 2022-04-26
Fix OOM error after uploading large chunked file

gerrod3:
fixes: #2573

### DIFF
--- a/pulpcore/app/files.py
+++ b/pulpcore/app/files.py
@@ -34,15 +34,16 @@ class PulpTemporaryUploadedFile(TemporaryUploadedFile):
         name = os.path.basename(file.name)
         instance = cls(name, "", file.size, "", "")
         instance.file = file
-        data = file.read()
+        # Default 1MB
+        while data := file.read(1048576):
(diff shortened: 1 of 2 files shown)


pulpcore/app/files.py
```diff
@@ -34,15 +34,16 @@ def from_file(cls, file):
         name = os.path.basename(file.name)
         instance = cls(name, "", file.size, "", "")
         instance.file = file
-        data = file.read()
+        # Should the read amount be a setting? Defaulting to 1MB
+        while data := file.read(1048576):
```
gerrod3, line 38: Any thoughts on what this value should be? Used the value that is default in the downloaders.

dralley (approved):

pulpcore/app/files.py
dralley: 1mb is fine

dralley: Just leave the comment saying that this constant = 1mb if you remove the question from the PR

ipanova (approved):

PR #2633, opened 2022-04-28
Removes `MissingResource` from the plugin API

### LINKED ISSUE #1812: Replace MissingResource exception with a subclass of rest_framework.exceptions.APIException
Author: @dkliban (dkliban@redhat.com)

Redmine Issue: 2382, https://pulp.plan.io/issues/2382

---

DRF provides an APIException\[0\] base class that should be used to create custom exceptions for Pulp's API. The new MissingResource exception should take two arguments: resource_type and resource_id.

\[0\] http://www.django-rest-framework.org/api-guide/exceptions/#apiexception

bmbouter:
closes #1812


pulpcore/tests/functional/api/test_tasks.py
```diff
@@ -0,0 +1,15 @@
...
+import pytest
+
+from pulpcore.client.pulpcore.exceptions import ApiException
+
+
+def test_cancel_invalid_task_raises_404(tasks_api_client):
+    with pytest.raises(ApiException) as e_info:
+        tasks_api_client.tasks_cancel(
+            task_href=f"/pulp/api/v3/tasks/{uuid4()}",
+            patched_task_cancel="anything"
```
bmbouter, line 12: @dkliban I didn't expect this param to be required and it seems the value doesn't matter. I think this is an issue in the bindings and I'm hoping you can advise me on how to fix it.

pulpcore/tests/functional/api/test_tasks.py
```diff
@@ -0,0 +1,15 @@
...
+
+import pytest
+
+from pulpcore.client.pulpcore.exceptions import ApiException
+
+
+def test_cancel_invalid_task_raises_404(tasks_api_client):
+    with pytest.raises(ApiException) as e_info:
+        tasks_api_client.tasks_cancel(
+            task_href=f"/pulp/api/v3/tasks/{uuid4()}",
```
bmbouter, line 11: How do I form this URL in a way that would also work on a rerooted system?

bmbouter: One idea would be to dispatch a task, delete that task, and then try to cancel it after the delete as a workaround...

pulpcore/tests/functional/api/test_tasks.py
```diff
@@ -0,0 +1,14 @@
...
+
+import pytest
+
+from pulpcore.client.pulpcore.exceptions import ApiException
+
+
+def test_cancel_invalid_task_raises_404(tasks_api_client):
+    with pytest.raises(ApiException) as e_info:
+        tasks_api_client.tasks_cancel(
+            task_href=f"/pulp/api/v3/tasks/{uuid4()}", patched_task_cancel="anything"
```
mdellweg, line 11:
```suggestion
            task_href=f"/pulp/api/v3/tasks/{uuid4()}/", patched_task_cancel="anything"
```

bmbouter: What do I do about this hard coded string?

bmbouter: It would be nice I guess if there was a fixture that yieled the `/pulp/api/v3/`string.... We don't have one of those do we?

bmbouter: I added it

dkliban (approved):

pulpcore/tests/functional/api/test_tasks.py
```diff
@@ -0,0 +1,18 @@
...
+
+import pytest
+
+from pulpcore.client.pulpcore.exceptions import ApiException
+
+
+def test_cancel_invalid_task_raises_404(pulp_api_v3_path, tasks_api_client):
+    patched_task_cancel = {"state": "canceled"}
+
+    missing_task_url = pulp_api_v3_path + f"tasks/{uuid4()}/"
```
mdellweg, line 11:
```suggestion
    missing_task_url = f"{pulp_api_v3_path}tasks/{uuid4()}/"
```

pulpcore/tests/functional/api/test_tasks.py
```diff
@@ -0,0 +1,18 @@
...
+
+    with pytest.raises(ApiException) as e_info:
+        tasks_api_client.tasks_cancel(
+            task_href=missing_task_url, patched_task_cancel=patched_task_cancel
+        )
+    assert e_info.value.status == 404
+    assert e_info.value.reason == "Not Found"
```
mdellweg, line 18: Just thinking: the 404 should be clue enough that everything went according to plan.

bmbouter: I agree, I think I've gone one `assert` to many.

bmbouter:
@mdellweg one more review please?

mdellweg (approved):

PR #2637, opened 2022-04-28
WIP toggable queryset scoping + content queryset scoping

gerrod3:
I expect this to go through many iterations so I'm going to explain how queryset scoping is working now so we can be on the same page when deciding how it should work.

First a request is routed to one of our views. If the view is a subclass of `NamedModelViewset` then the queryset scoping machinery will be used. But before that happens, the request has to be authenticated and authorized which is done through the viewset's `authentication_classes` and `permission_classes` respectively. If these aren't explicitly set on the viewset then they will use the default values in [settings.py](https://github.com/pulp/pulpcore/blob/main/pulpcore/app/settings.py#L153-L157). Here at the authorization step Pulp's default permission class `AccessPolicyFromDB` will try to load the access policy for the viewset from the DB. If there is no access policy then it will use the default policy of only allowing admins. Else it will check that the user has the appropriate permissions for this view. After authorization, the view's code will finally be called and if the request is for an object (get, list, update, delete) then the view's `get_queryset` method will be called. If `get_queryset` hasn't been overwritten from the `NamedModelViewset` implementation then it will try to perform queryset scoping using the view's `permission_classes`. It checks each permission class to see if it has the method `scope_queryset` and calls it if so. This allows plugin writers to change their implementation of queryset scoping by using a different permission class, like how Galaxy does. This is also the only way to disable queryset scoping currently by supplying a different permission class. Pulp's default and only permission class `AccessPolicyFromDB` scopes the queryset by checking to see if the view has the attribute `queryset_filtering_required_permission` and if so filter by that permission, else it does nothing.

Now onto the changes I made. I've added the boolean field `queryset_scoping` to the `AccessPolicy` model and changed the `scope_queryset` method in `AccessPolicyFromDB` to only call the queryset scoping method if this field is true. This requires on the plugin side adding this field to each default access policy to keep their queryset scoping working. Also, I've increased the ping-pong between view and permission class by adding a *new* `scope_queryset` method to `NamedModelViewset` that actually performs the scoping and have `AccessPolicyFromDB`'s `scope_queryset` call this new method if it is present on the view. This way plugin writers can customize their view's scoping by overwriting this method instead of `get_queryset`. 

Before: `view.get_queryset` --(perm class has scope_queryset)--> `permission_class.scope_queryset` --(view has scoping permission attribute)--> `role_util.get_objects_for_user`
Now: `view.get_queryset` --(perm class has scope_queryset)--> `permission_class.scope_queryset` --(queryset_scoping is true & view has scope_queryset)--> `view.scope_queryset` --(view has scoping permission attribute)--> `role_util.get_objects_for_user`

I was going to make the new `AccessPolicy` field a json for a queryset scoping function to call like we discussed, but that would require more machinery for us to build to be useful. It's still doable, but having queryset scoping just be easily toggable seems like the biggest improvement we can offer.


gerrod3:
Second commit changes: 

Added a different default `scope_queryset` implementation for `BasedContentViewset`, which all plugin Content viewsets should inherit from, and added a new attribute `repository_viewset` which should be set to that plugin's Repository viewset. The content `scope_queryset` will use `queryset_filtering_required_permission` on the `repository_viewset` to filter the repositories the user can see, then scope content based on those repositories' content. 

Implementation in pulp_file: https://github.com/pulp/pulp_file/pull/703

pulpcore/app/viewsets/content.py
```diff
@@ -99,14 +100,28 @@ class BaseContentViewSet(NamedModelViewSet):
...
+    def scope_queryset(self, qs):
+        """Scope the content based on repositories the user has permission to see."""
+
+        if view := self.repository_viewset:
+            if view_permission := getattr(view, "queryset_filtering_required_permission", None):
+                if not self.request.user.has_perm(view_permission):
+                    repos = get_objects_for_user(
+                        self.request.user, view_permission, view.queryset
+                    )
+                    return qs.filter(repositories__in=repos)
```
gerrod3, line 123: I love the walrus operator :seal:. Also, since I have the repository viewset maybe it makes more sense to get the repositories the user can see by calling its `scope_queryset` method instead since this is currently duplicated code of the base `scope_queryset` method.

gerrod3:
> (Have not looked at the code yet...) This sounds like a content will specify, which repository type it can be used in. I do not see any restriction of adding content to different repo types by now, and i would rather try to keep it that way. (e.g. file_content could be used for storing arbitrary files from any plugin.)

True, but repositories do have this limitation, they can only store what is in their `CONTENT_TYPES` attribute, and I'm not sure if we have any plugin that imports content types from another plugin. Maybe a better implementation would be to gather all repository models that can hold the content type, get their repository viewsets,  and run each `scope_queryset` method to get every repository the user can see. Then use that to scope the content so not to impose limitations on content.

mdellweg:
> > (Have not looked at the code yet...) This sounds like a content will specify, which repository type it can be used in. I do not see any restriction of adding content to different repo types by now, and i would rather try to keep it that way. (e.g. file_content could be used for storing arbitrary files from any plugin.)
> 
> True, but repositories do have this limitation, they can only store what is in their `CONTENT_TYPES` attribute, and I'm not sure if we have any plugin that imports content types from another plugin. Maybe a better implementation would be to gather all repository models that can hold the content type, get their repository viewsets, and run each `scope_queryset` method to get every repository the user can see. Then use that to scope the content so not to impose limitations on content.

Yes, i think this would be needed. Maybe we can compile this list for content in `pulpcore.app.apps` upon loading the plugins.
FWIW: It's not necessarily accross plugins. In pulp_container you already have two repository types sharing the same content types.

pulpcore/app/models/access_policy.py
```diff
@@ -31,13 +31,16 @@ class AccessPolicy(BaseModel):
...
+        queryset_scoping (BooleanField): True if the user wants to use the default queryset scoping
+            that comes with the view.
 
     """
 
     creation_hooks = models.JSONField(null=True)
     statements = models.JSONField()
     viewset_name = models.CharField(max_length=128, unique=True)
     customized = models.BooleanField(default=False)
+    queryset_scoping = models.BooleanField(default=False)
```
bmbouter, line 43: Maybe this should default to true? I think it should default to secure and isolated.

pulpcore/app/apps.py
```diff
@@ -16,6 +16,9 @@
 URLS_MODULE_NAME = "urls"
 MODELRESOURCE_MODULE_NAME = "modelresource"
 
+# Global mapping of content types and the repositories they can be stored in
+_content_repository_cache = defaultdict(list)
```
mdellweg, line 20: Can we leave this information on class attributes of the `Content` subclasses?

pulpcore/app/models/access_policy.py
mdellweg, line 43: I think we need to support the plugin writer to supply more than on and off as a choice to the user.

gerrod3:
4th commit changes:

- Changed `AccessPolicy`'s new field `scope_queryset` to a json field to accept a similar dictionary structure for defining the queryset scope function to be called. Now `AccessPolicyFromDB` will check to see if this field is defined and if the function is present on the viewset it will call it.
- Changed the default `scope_queryset` method on `NamedModelViewset` to account for parent/child relationships in Pulp. If viewset is child then it will perform the normal queryset scoping procedure using `queryset_filtering_required_permission`. If viewset is parent then it will call each child's `get_queryset` method (which will call their `scope_queryset` method if enabled) to get all objects across plugins the user can see.
- Changed optimization for master Content viewset `scope_queryset` method to place the repository mappings on the `Content` model itself. Functionally the same as the previous commit. @mdellweg 
- Added default access policies for master views Content, ContentGuard, Distribution, Publication, Repository, & RepositoryVersion. Authenticated users can view these endpoints and properly have the objects scoped. RepositoryVersion uses a modified `scope_queryset` similar to the Content viewset.

Default access policies are still missing for Exporters, Importers, Remotes,  SigningServices and Artifacts. Exporters, Importers, and Remotes don't seem to have master endpoints yet. SigningServices is different from the normal parent/child model relationship, so not sure what to do here. And artifacts aren't even using the default permission class, so this will need to be separate work as well.

pulpcore/app/apps.py
```diff
@@ -145,6 +146,9 @@ def import_viewsets(self):
                     if obj is not NamedModelViewSet and issubclass(obj, NamedModelViewSet):
                         model = obj.queryset.model
                         self.named_viewsets[model].append(obj)
+                        if model is not Repository and issubclass(model, Repository):
+                            for content in model.CONTENT_TYPES:
+                                content.MAPPED_REPOSITORIES.append(model)
```
ipanova, line 151: I think plugins should define this class attribute on their own, like CONTENT_TYPES.

mdellweg: The trick here is that plugins define this information already, we just need to pivot that table.
(Also this way we make room for plugin Repositories to allow containing Content from different plugins.)

mdellweg: But we could name the attribute `REPOSITORY_TYPES` for symmetry.

ipanova: my point is - if this is a newly exposed class attribute it can be used in the plugin but then we mess with it.

mdellweg: My point is, if we need the plugin writer to provide the same information both ways, it will easily become inconsistent.

> [...] maybe plugin writer of plugin X does not want his content to be used in plugin Y [...]

I don't think, we can provide any safety around here. Plugins can change whatever aspect of the installation they desire. Also, at least at the Artifact level, we _want_ plugins to share.

pulpcore/app/models/content.py
```diff
@@ -506,6 +506,15 @@ class Meta:
         verbose_name_plural = "content"
         unique_together = ()
 
+    @property
+    def REPOSITORY_TYPES(self):
+        """
+        List of the repository models that can store this content type.
+
+        Set at start up time. Read only.
```
mdellweg: Hmm. It is still a mutable list we return here. Python is hard...

pulpcore/app/apps.py
```diff
@@ -148,7 +148,7 @@ def import_viewsets(self):
                         self.named_viewsets[model].append(obj)
                         if model is not Repository and issubclass(model, Repository):
                             for content in model.CONTENT_TYPES:
-                                content.MAPPED_REPOSITORIES.append(model)
+                                content._repository_types.append(model)
```
ipanova: :+1:

pulpcore/app/models/content.py
gerrod3: Changed it to return a tuple. Now it should be read only.

pulpcore/app/models/content.py
```diff
@@ -509,13 +509,13 @@ class Meta:
...
-    def REPOSITORY_TYPES(cls) -> set:
+    def REPOSITORY_TYPES(cls) -> tuple:
         """
-        Set of the repository models that can store this content type.
+        Tuple of the repository models that can store this content type.
 
         Populated at start up time. Read only.
         """
-        return cls._repository_types[cls]
+        return tuple(cls._repository_types[cls])
```
mdellweg, line 518: +1 for the tuple. But i don't think we need the default dict. Making the class attribute a set should be sufficient. Or do all Content classes share the same dict?

gerrod3: All Content classes share the same dict.

CHANGES/2723.feature
```diff
@@ -0,0 +1,5 @@
+Queryset scoping is now performed when the ViewSet's AccessPolicy field ``scope_queryset`` is set to
```
ipanova, line 1: I wonder if all this does not rather belong to plugin_api instead

PR #2638, opened 2022-04-29
[Docs] Note about local settings file

pavelpicka:
Note that user can use ``/etc/pulp/settings.local.py`` to keep local
settings through installer pulp upgrade.

[noissue]

### DIFF
--- a/docs/configuration/applying.rst
+++ b/docs/configuration/applying.rst
@@ -23,6 +23,7 @@ Or in a systemd file you could::
 
 
 Dynaconf supports `settings in multiple file formats <https://www.dynaconf.com/configuration/>`_
+Dynaconf also support `local settings file <https://www.dynaconf.com/settings_files/#local-settings-files>`_
 
 This file should have permissions of:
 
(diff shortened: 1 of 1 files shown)


docs/configuration/applying.rst
```diff
@@ -10,6 +10,11 @@ to configure Pulp in a few ways:
 By Configuration File
 ---------------------
 
+.. note::
+
+    If you use pulp_installer, you should modify ``/etc/pulp/settings.local.py`` instead
```
bmbouter, line 15: Perhaps we should get a new section at the bottom of this page dedicated to configuration if using the pulp intaller. It could contain this info and also identify that the /etc/pulp/settings.py is available too. Until this PR this page has been agnostic on installers. If we're taking a step towards specific info for installers, we probably should make it a bit more structured. @pavelpicka wdyt?

pavelpicka: This change is not only for a pulp installer. We can use this local settings file everywhere because of the usage of dynaconf. 
But I like the idea of a new section as it will be more visible than just a note.

ipanova (approved):

dralley:
@pavelpicka @bmbouter The documentation here looks good, the thing that doesn't make sense to me is `/etc/pulp/settings.local.py`.  I thought the standard way of handling configuration files was for e.g. `/etc/pulp/settings.py` to be left alone and user-editable, and static defaults in a file like `/usr/lib/pulp/settings.py`

pavelpicka:
The installer uses/writes only ``/etc/pulp/settings.py``, and the installer can change it during an upgrade. A usage of ``settings.local.py`` is a feature of dynaconf. And a user can keep their settings during that upgrade as well.

PR #2639, opened 2022-04-29
Test stream

mdellweg:
Also export ArtifactResonse in the plugin api.

### DIFF
--- a/CHANGES/plugin_api/2537.feature
+++ b/CHANGES/plugin_api/2537.feature
@@ -1,0 +1,2 @@
+Added ``pulpcore.plugin.content.ArtifactResponse`` to plugin API. Use this response to stream an
+artifact from the object storage if redirecting is impossible.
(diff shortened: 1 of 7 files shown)


CHANGES/plugin_api/2537.feature
```diff
@@ -0,0 +1 @@
+Added ``pulpcore.plugin.content.ArtifactResponse`` to plugin API.
```
ipanova, line 1: for information purpose, can you add a line about what this response is about?

ipanova (approved):

template_config.yml
```diff
@@ -34,7 +34,7 @@ flake8: true
 github_org: pulp
 issue_tracker: github
 keep_ci_update_for_latest_branches: 5
-keep_ci_update_since_branch: "3.14"
+keep_ci_update_since_branch: '3.14'
```
bmbouter, line 37: With black aren't all string with `"`?

bmbouter: I guess this is yml not python....

bmbouter (approved):
Thank you!

template_config.yml
mdellweg: I didn't touch this line by hand.

PR #2651, opened 2022-05-02
Upgrade Django and ensure compat with it

bmbouter:
[noissue]

(cherry picked from commit c06ed6d)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ aiodns
 aiofiles
 backoff
 click<8
-Django~=2.2.16  # LTS version, switch only if we have a compelling reason to
+Django~=2.2.28  # LTS version, switch only if we have a compelling reason to
 django-currentuser~=0.5.1
 django-filter~=2.3.0
 django-guardian~=2.3.0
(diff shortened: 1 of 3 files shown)


mdellweg (approved):

ipanova (changes requested):
This change does not suffice for container plugin https://github.com/pulp/pulp_container/pull/745#issuecomment-1114800086
https://github.com/pulp/pulp_container/runs/6271437401?check_suite_focus=true#step:14:399

bmbouter:
We're going to wait and see if those using 3.7 are aware that even after fixing this they will be using an unsupported version of Django. If they still want us to release the fixes we'll have to evaluate all plugins (like pulp_container) and likely release several fixes. So we're waiting for now to see what is best.

ipanova (approved):

requirements.txt
```diff
@@ -3,7 +3,7 @@ aiodns
 aiofiles
 backoff
 click<8
-Django~=2.2.16  # LTS version, switch only if we have a compelling reason to
```
ipanova, line 6: @bmbouter I think we actually don't even need this PR. Means that we also don't need to release pulpcore 3.7
once pulp-container 2.1 is released with this change https://github.com/pulp/pulp_container/pull/745/files i think we are good to go.

bmbouter: Yeah I think you are right. I'll close the pulpcore PR w/ a similar comment.

bmbouter: Oh ha, this is on that very PR. Closing!

Issue #2476, opened 2022-04-04
Allow customization of GNUPGHOME and KEYRING on SigningService


rochacbruno:
**Is your feature request related to a problem? Please describe.**

I want to be able to call `django-admin add-signing-service` and specify a key_id located on a custom keyring e.g: `/etc/pulp/certs/galaxy.kbx`

Right now that command doesn't accept the `keyring` or `gpghome` parameter as those fields are not part of `SigningService` model. (should it have?) and we can create signing services only using system keyring.

Also on `AsciiArmouredSigningService` here https://github.com/pulp/pulpcore/blob/aff273436a623da3fa9433d226ee6fa026097aff/pulpcore/app/models/content.py#L846 

we need a way for that `GPG` instance to be created pointing to custom keyring e.g:

```py
gpg = gnupg.GPG(gnupghome=settings.ANSIBLE_CERTS_DIR, keyring='galaxy.kbx")
```

**Describe the solution you'd like**

Option 1:
  Save GPGHOME and Keyring on the SigningService Model
Option2:
  Read from settings (which automatically also reads from envvars)

This same issue needs a fix on UploadSignature verification on pulp_ansible and everywhere we use GPG

quba42:
I think the signing service validate function should not require access to anything but the public key on the signing service model, and the self.sign() call. It can then validate what `self.sign()` returns by importing the public key from the signing service into an otherwise empty temporary gpg homedir. This is how we do it in pulp_deb: https://github.com/pulp/pulp_deb/blob/main/pulp_deb/app/models/signing_service.py#L72

Access to the secret key is completely separated from pulp via whatever the creator of the signing service supplies to `self.sign()` via the `script` field.

lubosmj:
It seems like `pulp_deb` has already somehow resolved the issue that needs to be addressed at the pulpcore level.

In pulpcore/pulp_container, we use `gnupg.GPG()` everywhere and everything is taken from a default GnuPG home directory during the addition or validation of a signing service.

PR #2653, opened 2022-05-03
Enable admins to specify a custom GnuPG directory or keyring

lubosmj:
In this commit, the validation procedure triggered before saving signing services was refactored. As of this commit, users will no longer experience problems with missing keys or improperly configured GnuPG home directory. Public keys are imported to a temporary GnuPG directory where everything is correctly staged for validation purposes.

Also, redundant validation checks in the AsciiArmoredDetached's validate() method were removed as well.

closes https://github.com/pulp/pulpcore/issues/2476

### DIFF
--- a/CHANGES/2476.feature
+++ b/CHANGES/2476.feature
@@ -1,0 +1,3 @@
+Enabled administrators to work with a customized GnuPG home directory and keyring during the
+creation of a signing service. The introduced optional arguments ``--gnupghome`` and ``--keyring``
+are available under the ``pulpcore-manager add-signing-service`` command.
--- a/docs/workflows/signed-metadata.rst
+++ b/docs/workflows/signed-metadata.rst
@@ -61,34 +61,24 @@ The example below demonstrates how a signing service can be created using ``gpg`
        Make sure the script contains a proper shebang and Pulp has got valid permissions
        to execute it.
 
-3. Create a signing service consisting of an absolute path to the script and a meaningful
-   name describing the script's purpose. It is possible to insert the signing service in
-   to a database by using the ``pulpcore-manager shell_plus`` interactive Python shell. Here is an
-   example showing how to create one instance pointing to a script:
+3. Create a signing service consisting of an absolute path to the script, a meaningful
+   name describing the script's purpose, and the identity identifying the key for signing. The
+   script must be executable. Here is an example showing how to create one instance of a signing
+   service:
 
-   .. code-block:: python
+   .. code-block:: bash
 
-       from pulpcore.app.models.content import AsciiArmoredDetachedSigningService
-
-       # read an already exported public key
-       with open("public.key") as key:
-           AsciiArmoredDetachedSigningService.objects.create(
-               name="sign-metadata",
-               public_key=key.read(),
-               pubkey_fingerprint="19CD52BD1CA9A00DF10A842D74B14E3590C2231F",
-               script="/var/lib/pulp/scripts/sign-metadata.sh",
-           )
+        pulpcore-manager add-signing-service ${SERVICE_NAME} ${SCRIPT_ABS_FILENAME} ${KEYID}
 
    .. note::
 
-       While creating a signing service, the model ``AsciiArmoredDetachedSigningService``
-       runs additional checks in order to prevent saving invalid scripts to the database.
-       This feature enables administrators to validate their signing scripts in advance.
+      The public key must be available on the caller's keyring or on a keyring provided via the
+      ``--gpghome`` or ``--keyring`` parameters.
 
-   .. note::
+   .. warning::
 
-      You can use `pulpcore-manager add-signing-service` to add a ``SigningService``.
-      This command is however still in tech-preview.
+      It is possible to insert a new signing service into the database by using the
+      ``pulpcore-manager shell_plus`` interactive Python shell. However, this is not recommended.
 
 4. Retrieve and check the saved signing service via REST API::
 
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -10,8 +10,6 @@ import tempfile
 import shutil
 import subprocess
 
-import gnupg
-
 from functools import lru_cache
 from itertools import chain
 
@@ -25,6 +23,7 @@ from django_lifecycle import BEFORE_UPDATE, BEFORE_SAVE, hook
 
 from pulpcore.constants import ALL_KNOWN_CONTENT_CHECKSUMS
 from pulpcore.app import pulp_hashlib
+from pulpcore.app.util import verify_signature
 from pulpcore.app.models import MasterModel, BaseModel, fields, storage
 from pulpcore.exceptions import (
     DigestValidationError,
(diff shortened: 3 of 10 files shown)


pulpcore/app/util.py
```diff
@@ -159,3 +162,25 @@ def get_request_without_query_params(context):
...
+
+def verify_signature(filepath, public_key, pubkey_fingerprint=None):
+    """Check whether the provided file can be verified with the particular public key."""
+    with tempfile.TemporaryDirectory() as temp_directory_name:
+        gpg = gnupg.GPG(gnupghome=temp_directory_name)
+        gpg.import_keys(public_key)
+        imported_keys = gpg.list_keys()
+
+        if len(imported_keys) != 1:
+            raise RuntimeError("Exactly one key must be imported.")
```
mdellweg, line 175: @quba42 Is this limitation useful? Can there be situations where you would check against one of multiple keys?

pulpcore/app/util.py
```diff
@@ -159,3 +162,25 @@ def get_request_without_query_params(context):
...
+
+        if pubkey_fingerprint and imported_keys[0]["fingerprint"] != pubkey_fingerprint:
+            raise RuntimeError(
+                "The fingerprint of the signing service's public key does not match the "
+                "fingerprint retrieved from an imported version of the public key."
+            )
```
mdellweg, line 181: This seems superfluous. The keyring generated only contains the single key imported, so there's no room for ambiguity anymore. Also this function is about verification. It should not contain any trace of a signing service.

pulpcore/app/util.py
```diff
@@ -159,3 +162,25 @@ def get_request_without_query_params(context):
...
+
+        with open(filepath, "rb") as signature:
+            verified = gpg.verify_file(signature)
+            if not verified.valid:
+                raise RuntimeError(f"The file '{filepath}' has not got a valid signature.")
```
mdellweg, line 186: I am wondering if we should throw a new subclass of `RuntimeError` or just return the result of the verification.

pulpcore/app/models/content.py
```diff
@@ -845,26 +844,9 @@ def validate(self):
...
-                            )
-                        )
-                    elif verified.pubkey_fingerprint != self.pubkey_fingerprint:
-                        raise RuntimeError(
-                            _(
-                                "Fingerprints of the provided public key and the verified public "
-                                "key are not equal. The signing script is probably not valid."
-                            )
-                        )
+        with tempfile.NamedTemporaryFile() as temp_file:
```
mdellweg, line 847: We should keep the temp directory wrapper too.

pulpcore/app/util.py
```diff
@@ -159,3 +162,25 @@ def get_request_without_query_params(context):
         request.query_params._mutable = False
 
     return request
+
+
+def verify_signature(filepath, public_key, pubkey_fingerprint=None):
+    """Check whether the provided file can be verified with the particular public key."""
+    with tempfile.TemporaryDirectory() as temp_directory_name:
```
mdellweg, line 169:
```suggestion
    with tempfile.TemporaryDirectory(dir=".") as temp_directory_name:
```
This can be run inside tasks and they need to use the current working directory for temporary stuff.

mdellweg (commented):
Thank you for this.
Does the validation work with all types of signatures (inline/detatched ; armored/text) ?

CHANGES/2476.feature
```diff
@@ -0,0 +1,3 @@
+Enabled administrators to work with a customized GnuPG home directory and keyring during the
+creation of a signing service. The introduced options ``--gnupghome`` and ``--keyring`` are
```
ipanova, line 2: s/options/optional arguments/

pulpcore/app/util.py
lubosmj: The check is a left-over from the old validation method. I was hesitant to remove it completely or move it to the `validate` method of `SigningService`. We would need to initialize a new GnuPG instance two times when considering the second option.

Regarding the ambiguity, there is still room for that. Administrators are passing fingerprints and public keys separately while creating signing services. In the check, we are testing whether the provided fingerprints match the fingerprints of the imported public keys.

We can try to compute fingerprints from passed public keys before saving objects to the database. Then, administrators would not have to deal with fingerprints by themselves since everything would be computed on the fly. With this, we lower requirements for the administrators. However, computing the fingerprint may not be as simple as it sounds.

pulpcore/app/util.py
ipanova, line 186: s/The file '{filepath}' has not got/The file '{filepath}' does not have/

lubosmj: I am in favour of mentioning that this function raises `RuntimeError` once the verification was not successful in the docstring.

pulpcore/plugin/util.py
```diff
@@ -11,3 +11,5 @@
     get_users_with_perms_attached_roles,
     remove_role,
 )
+
+from pulpcore.app.util import verify_signature  # noqa
```
ipanova, line 15: you should add changelog entry for plugin_api too

pulpcore/app/util.py
mdellweg: That is not true if we consider the pulpcore-manager command as the interface for the admin. They specify a key id and that's it.
I'd say, remove this check _and_ the `pubkey_fingerprint` argument.
Remember, `verify_signature` is meant to be used not only in the creation of signing services.

lubosmj: You can still create a signing service without `pulpcore-manager`, can't you? Do we want to end up in a non-consistent state? I still think that this check should be running somewhere. Maybe, this ^ is not the right place to run it, I agree.

pulpcore/app/util.py
mdellweg: If you subclass RuntimeError, you can still catch it as a RuntimeError. But it allows to pin the reason for failure more granulary.

pulpcore/app/util.py
mdellweg: What is the other way? The one i know of is using the orm directly in `pulpcore-manager shell` (one that i would consider inherently unsafe).

lubosmj: Yes, there are two standard ways of saving signing services: 1. `pulpcore-manager shell_plus` (dangerous), 2. `pulpcore-manager add-signing-service` (fine). But, what prevents plugin writers from adding a new endpoint for managing their signing services? I am removing the check if we believe that administrators will always use `pulpcore-manager add-signing-service` and plugin writers will never think of validating fingerprints.

In the documentation, we suggest using `pulpcore-manager shell_plus`. It should be then described as one of the non-recommended ways of creating signing services (https://docs.pulpproject.org/pulpcore/workflows/signed-metadata.html?highlight=signing%20service).

pulpcore/app/util.py
lubosmj: This suggestion is debatable because we also deal with detached signatures.

lubosmj:
@mdellweg, yes, the validation works for both detached and inlined signatures.

docs/workflows/signed-metadata.rst
```diff
@@ -61,23 +61,12 @@ The example below demonstrates how a signing service can be created using ``gpg`
...
-3. Create a signing service consisting of an absolute path to the script and a meaningful
-   name describing the script's purpose. It is possible to insert the signing service in
-   to a database by using the ``pulpcore-manager shell_plus`` interactive Python shell. Here is an
-   example showing how to create one instance pointing to a script:
+3. Create a signing service consisting of an absolute path to the script, a meaningful
+   name describing the script's purpose, and the identity identifying the key for signing. The
+   script must be executable. Here is an example showing how to create one instance of a signing
+   service::
 
-   .. code-block:: python
```
mdellweg, line 69: [not a ReST expert here] Do we want this to be a `.. code-block:: bash/shell/sh/whatever`?

docs/workflows/signed-metadata.rst
```diff
@@ -61,23 +61,12 @@ The example below demonstrates how a signing service can be created using ``gpg`
...
-       with open("public.key") as key:
-           AsciiArmoredDetachedSigningService.objects.create(
-               name="sign-metadata",
-               public_key=key.read(),
-               pubkey_fingerprint="19CD52BD1CA9A00DF10A842D74B14E3590C2231F",
-               script="/var/lib/pulp/scripts/sign-metadata.sh",
-           )
+        $ pulpcore-manager add-signing-service ${SERVICE_NAME} ${SCRIPT_ABS_FILENAME} ${KEYID}
 
    .. note::
```
mdellweg, line 72: We could add to this note something like:
"The public key must be available on the callers keyring or on a keyring provided via the `--gpghome` or `--keyring` parameters."

pulpcore/plugin/util.py
mdellweg: Also we need to export the Exception `InvalidSignatureError` too.

mdellweg (changes requested):

docs/workflows/signed-metadata.rst
lubosmj: I used `code-block:: bash` instead.

docs/workflows/signed-metadata.rst
lubosmj: I removed a bunch of unnecessary sentences and added the suggested statement.

mdellweg (approved):

ipanova (approved):

pulpcore/app/util.py
ipanova: this suggestion was mostly about english grammar :D

PR #2656, opened 2022-05-03
Adds openapi v3 schema validation test

### LINKED ISSUE #2643: Add test to verify the openAPI schema validates
Recently with [this change in 3.19](https://github.com/pulp/pulpcore/pull/2292) we now have a schema that validates with openAPI V2. I'd like to see a test that validates it.

bmbouter:
To validate, it needs the `pk_path` param to be True, so for example you
can use:  `/pulp/api/v3/docs/api.json?pk_path=1`.

closes #2643
AAH-1548

### DIFF
--- a/pulpcore/tests/functional/api/test_openapi_schema.py
+++ b/pulpcore/tests/functional/api/test_openapi_schema.py
@@ -1,0 +1,73 @@
+"""Test related to the openapi schema Pulp generates."""
+import asyncio
+import copy
+import json
+
+import aiohttp
+import pytest
+import jsonschema
+
+from drf_spectacular.validation import JSON_SCHEMA_SPEC_PATH
+from jsonschema import ValidationError
+
+
+@pytest.fixture(scope="session")
+def openapi3_schema_spec():
+    with open(JSON_SCHEMA_SPEC_PATH) as fh:
+        openapi3_schema_spec = json.load(fh)
+
+    return openapi3_schema_spec
+
+
+@pytest.fixture(scope="session")
+def openapi3_schema_with_modified_safe_chars(openapi3_schema_spec):
+    openapi3_schema_spec_copy = copy.deepcopy(openapi3_schema_spec)  # Don't modify the original
+    # Making OpenAPI validation to accept paths starting with / and {
+    properties = openapi3_schema_spec_copy["definitions"]["Paths"]["patternProperties"]
+    properties["^\\/|{"] = properties["^\\/"]
+    del properties["^\\/"]
+
+    return openapi3_schema_spec_copy
+
+
+@pytest.fixture(scope="session")
+def pulp_openapi_schema_url(pulp_cfg, pulp_api_v3_path):
+    return f"{pulp_cfg.get_base_url()}{pulp_api_v3_path}docs/api.json"
+
+
+@pytest.fixture(scope="session")
+def pulp_openapi_schema(pulp_openapi_schema_url):
+    return asyncio.run(_download_schema(pulp_openapi_schema_url))
+
+
+@pytest.fixture(scope="session")
+def pulp_openapi_schema_pk_path_set(pulp_openapi_schema_url):
+    url = f"{pulp_openapi_schema_url}?pk_path=1"
+    return asyncio.run(_download_schema(url))
+
+
+async def _download_schema(url):
+    async with aiohttp.ClientSession() as session:
+        async with session.get(url) as response:
+            return await response.json()
+
+
+@pytest.mark.parallel
+@pytest.mark.from_pulpcore_for_all_plugins
+def test_valid_with_pk_path_set(pulp_openapi_schema_pk_path_set, openapi3_schema_spec):
+    jsonschema.validate(instance=pulp_openapi_schema_pk_path_set, schema=openapi3_schema_spec)
+
+
+@pytest.mark.parallel
+@pytest.mark.from_pulpcore_for_all_plugins
+def test_invalid_default_schema(pulp_openapi_schema, openapi3_schema_spec):
+    with pytest.raises(ValidationError):
+        jsonschema.validate(instance=pulp_openapi_schema, schema=openapi3_schema_spec)
+
+
+@pytest.mark.parallel
+@pytest.mark.from_pulpcore_for_all_plugins
+def test_valid_with_safe_chars(pulp_openapi_schema, openapi3_schema_with_modified_safe_chars):
+    jsonschema.validate(
+        instance=pulp_openapi_schema, schema=openapi3_schema_with_modified_safe_chars
+    )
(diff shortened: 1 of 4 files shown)


pulpcore/tests/functional/api/test_openapi_schema.py
```diff
@@ -0,0 +1,26 @@
...
+    url = pulp_cfg.get_base_url() + pulp_api_v3_path + "docs/api.json?pk_path=1"
+    schema_json = asyncio.run(download_schema(url))
+    validate_spec(schema_json)
+
+
+@pytest.mark.xfail
+def test_valid_schema_invalid_with_pk_path_unset(pulp_cfg, pulp_api_v3_path):
+    url = pulp_cfg.get_base_url() + pulp_api_v3_path + "docs/api.json"
+    schema_json = asyncio.run(download_schema(url))
+    validate_spec(schema_json)
```
mdellweg, line 26: What are your plans here? Do you intend to change the rules of the validator to make it pass? If not i would not use `xfail`.

pulpcore/tests/functional/api/test_openapi_schema.py
```diff
@@ -0,0 +1,26 @@
...
+async def download_schema(url):
+    async with aiohttp.ClientSession() as session:
+        async with session.get(url) as response:
+            return await response.json()
+
+
+def test_valid_schema_with_pk_path_set(pulp_cfg, pulp_api_v3_path):
+    url = pulp_cfg.get_base_url() + pulp_api_v3_path + "docs/api.json?pk_path=1"
+    schema_json = asyncio.run(download_schema(url))
+    validate_spec(schema_json)
```
mdellweg, line 19: We have validation for the unmodified schema on the shell level through the plugin template, and this validation should go a similar road, so we get the schema to validate with plugins installed.

pulpcore/tests/functional/api/test_openapi_schema.py
bmbouter: Yeah I was just so excited to use `xfail`. I do want to have a test demonstrate it's invalid without the `pk_path` param, but I'll rewrite it to use `pytest.raises`.

pulpcore/tests/functional/api/test_openapi_schema.py
bmbouter: Yeah I saw that. I didn't understand where `pulp_schema.yml` came from [here](https://github.com/pulp/plugin_template/blob/8ac2d57d66830433a64514d4d78fe5da9eb13e29/templates/github/.github/workflows/scripts/script.sh.j2#L40).

I'd like to try to continue with these tests, but use the plugin_template enable easy running of tests like these. One question is would we mark just tests in pulpcore? I kind of think yes, to keep it simple. We could also remove the bash-style validation to make the plugin_template simpler.

bmbouter:
@mdellweg this is ready for re-review. @fao89 @dkliban FYI also. I need a second reviewer too

pulpcore/tests/functional/api/test_openapi_schema.py
bmbouter: Mmmm I see. I'm trying to test what is built through the API with these tests.

I'm going to open a plugin_template PR that runs these tests in the plugin CIs and we'll see what you think.

pulpcore/tests/functional/api/test_openapi_schema.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def test_valid_schema_invalid_with_pk_path_unset(pulp_cfg, pulp_api_v3_path):
+    url = pulp_cfg.get_base_url() + pulp_api_v3_path + "docs/api.json"
+    schema_json = asyncio.run(download_schema(url))
+    with pytest.raises(OpenAPIValidationError):
+        validate_spec(schema_json)
```
mdellweg, line 27: I'm really conflicted with this test. I would rather like to see this pass with exceptions to the rules. What if the schema accidentally passes?

mdellweg, line 27: I'm really conflicted with this test. I would rather like to see this pass with exceptions to the rules. What if the schema accidentally passes?

bmbouter:
@mdellweg I think this does what we talked about. I switched us back to using `drf-spectacular` for the validation. Also it has the additional test showing our defualt schema passes with the safe_chars modification, but not against the vanilla 3.0 spec. It also shows the `pk_path=1` passes with the vanilla v3 spec.

After this merges I'm going to replace the plugin_template functionality with running these tests. In that work I'll remove [this script](https://github.com/pulp/plugin_template/blob/7c31f647d839e3a761f268e2d93a8e22b5ef857a/templates/github/.ci/scripts/schema.py) entirely and also [replace this part ](https://github.com/pulp/plugin_template/blob/8ac2d57d66830433a64514d4d78fe5da9eb13e29/templates/github/.github/workflows/scripts/script.sh.j2#L35-L41).

@fao89 @dkliban FYI also

bmbouter:
Here's the plugin template PR that does all that https://github.com/pulp/plugin_template/pull/635

I'll apply it to pulpcore and its pulp_file and set those to `"run_pulpcore_tests_for_plugins": False` while all other plugins should use the default `"run_pulpcore_tests_for_plugins": True`.

pulpcore/tests/functional/api/test_openapi_schema.py
```diff
@@ -0,0 +1,68 @@
...
+
+import aiohttp
+import pytest
+import jsonschema
+
+from drf_spectacular.validation import JSON_SCHEMA_SPEC_PATH
+from jsonschema import ValidationError
+
+
+@pytest.fixture
```
mdellweg, line 13:
```suggestion
@pytest.fixture(scope="session")
```
We should only load this once (and use monkeypatch for the modifying fixture).

pulpcore/tests/functional/api/test_openapi_schema.py
```diff
@@ -0,0 +1,68 @@
...
+
+    return openapi3_schema_spec
+
+
+@pytest.fixture
+def openapi3_schema_with_modified_safe_chars(openapi3_schema_spec):
+    # Making OpenAPI validation to accept paths starting with / and {
+    properties = openapi3_schema_spec["definitions"]["Paths"]["patternProperties"]
+    properties["^\\/|{"] = properties["^\\/"]
+    del properties["^\\/"]
```
mdellweg, line 26:
```suggestion
def openapi3_schema_with_modified_safe_chars(openapi3_schema_spec, monkeypatch):
    # Making OpenAPI validation to accept paths starting with / and {
    properties = openapi3_schema_spec["definitions"]["Paths"]["patternProperties"]
    monkeypatch.setitem(properties, "^\\/|{", properties["^\\/"])
    monkeypatch.delitem(properties, "^\\/")
```

mdellweg (commented):
Are those suitable to run in parallel?

pulpcore/tests/functional/api/test_openapi_schema.py
bmbouter: They are parallel safe, I'm marking them as such! Also the plugin_template PR now integrates this new pytest run w/ the `parallel` markers too so this will be used correctly by marking them here.

PR #2662, opened 2022-05-03
Taught pulp-import to retry up to three times on collisions.

### LINKED ISSUE #2589: Concurrent PulpImports with iverlapping  content can still fail.

ggainey:
fixes #2589.
[nocoverage]

(cherry picked from commit d936ce5bb052973133336157fbeb2a1c24f08024)

### DIFF
--- a/CHANGES/2589.bugfix
+++ b/CHANGES/2589.bugfix
@@ -1,0 +1,3 @@
+Taught PulpImport to retry more than once in the event of creation-collisions.
+
+This fixes a rare import-failure during high-concurrency, high-content-overlap imports.
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -44,6 +44,10 @@ CA_FILE = "pulpcore.app.modelresource.ContentArtifactResource.json"
 VERSIONS_FILE = "versions.json"
 CONTENT_MAPPING_FILE = "content_mapping.json"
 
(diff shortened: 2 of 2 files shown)


bmbouter (approved):
Thanks!

Issue #2553, opened 2022-04-15
Can't set content_path_prefix to /


ben417:
When I set "content_path_prefix" to "/" (also tried "", similar behavior), I'm not able to access the repository and the distribution base_url has an unexpected double // in it.

```
pulp rpm distribution show --name latest_centos-7-os-x86_64
{
  "pulp_href": "/pulp/api/v3/distributions/rpm/rpm/a58aaa3a-3be4-43dc-9591-f834c53d54a1/",
  "pulp_created": "2022-04-15T17:18:28.526831Z",
  "base_path": "latest/centos/7/os/x86_64",
  "base_url": "https://localhost//latest/centos/7/os/x86_64/",
  "content_guard": null,
  "pulp_labels": {},
  "name": "latest_centos-7-os-x86_64",
  "repository": null,
  "publication": "/pulp/api/v3/publications/rpm/rpm/3e4a6f2d-24be-4004-b702-0f8c3df0afcc/"
}
```

```
curl -k -I https://localhost//latest/centos/7/os/x86_64/config.repo
HTTP/1.1 404 Not Found
curl -k -I https://localhost/latest/centos/7/os/x86_64/config.repo
HTTP/1.1 404 Not Found
```

We're setting up Pulp to replace an existing homegrown server. The existing server servers it's content from root and has so many client instances in the wild (thousands) configured that way that changing to a new content path would be kind of a pain.

Note that the "/latest/" part of the example base_url path above needs to change between different distributions. I can't set content_path_prefix to "/latest/".

ben417:
Part of my problem is that the Ansible installer hard codes `/pulp/content/` in the Nginx config. See pulp/pulp_installer#989

I edited nginx and am able to access content now. base_url still has the double slash though.

ben417:
No, "" still has the //.

PR #2667, opened 2022-05-03
Replaces "//" with "/" in base_url.

dkliban:
This bug would appear when CONTENT_PATH_PREFIX was set to "" or "/".

fixes: #2553


bmbouter (approved):
Thanks!

mdellweg (approved):

PR #2701, opened 2022-05-09
Set an upper bound for setuptools in requirements.txt

dkliban:
[noissue]


bmbouter (approved):
Thanks! Using the previous range seems right because we don't want to accidentally downgrade someone either.

PR #2713, opened 2022-05-10
Create missing repositories automatically on import

### LINKED ISSUE #1920: As a user, I can perform an import without creating the destination repo(s) in advance
Author: daviddavis (daviddavis)

Redmine Issue: 7297, https://pulp.plan.io/issues/7297

---

A couple options we discussed:

1. Do this by default if the destination repo(s) don't exist
2. Create a flag (eg create_repos_on_import) that a user can use to trigger this behavior

lubosmj:
As of this commit, repositories are identified as standard import/export resource objects. It allows the automatic creation of importing repositories that were not created beforehand or do not exist.

closes #1920

### DIFF
--- a/docs/plugins/plugin-writer/concepts/subclassing/import-export.rst
+++ b/docs/plugins/plugin-writer/concepts/subclassing/import-export.rst
@@ -88,6 +88,32 @@ this::
     IMPORT_ORDER = [FileContentResource]
 
 
+Plugin writers are encouraged to subclass the ``RepositoryResource`` class to enable automatic
+repository creation during the import. For the ``pulp_file`` plugin, the following implementation
+should be considered::
+
+    from pulpcore.plugin.modelresources import RepositoryResource
+    from pulp_file.app.models import FileRepository
+
+    class FileRepositoryResource(RepositoryResource):
+        """
+        A resource for importing/exporting file repository entities
+        """
+
+        def set_up_queryset(self):
+            """
+            :return: A queryset containing one repository that will be exported.
+            """
+            return FileRepository.objects.filter(pk=self.repo_version.repository)
+
+        class Meta:
+            model = FileRepository
+
+
+    # the list signifying the order of imports must also include the repository resource class
+    IMPORT_ORDER = [FileContentResource, FileRepositoryResource]
+
+
 content_mapping
 ~~~~~~~~~~~~~~~
 
--- a/docs/workflows/import-export.rst
+++ b/docs/workflows/import-export.rst
@@ -366,10 +366,17 @@ to the names of repos in Pulp. For example, suppose the name of the repo in the 
 
     http :/pulp/api/v3/importers/core/pulp/ name="test" repo_mapping:="{\"source\": \"dest\"}"
 
-.. note::
-   Pulp import expects destination repository/repositories to be present at the import process.
-
 After the importer is created, a POST request to create an import will trigger the import process.
+
+.. note::
+    By default, the Pulp import machinery expects destination repositories to be present at the time
+    of the import. This can be overridden by passing the ``create_repositories=True`` field via the
+    POST request that will lead Pulp to create missing repositories on the fly.
+
+.. warning::
+    The options ``repo_mapping`` and ``create_repositories`` are not compatible with each other. The
+    existence of a repository specified in the ``repo_mapping`` option is tested before the importer
+    is initialized. Thus, the repository has to be already created in advance.
 
 You can import an exported ``.tar.gz`` directly using the ``path`` parameter::
 
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -3,7 +3,7 @@ Repository related Django models.
 """
 from contextlib import suppress
 from gettext import gettext as _
-from os import path
+from os import path, environ
 from collections import defaultdict
 import logging
 
@@ -12,12 +12,12 @@ from asyncio_throttle import Throttler
 from dynaconf import settings
 from django.core.validators import MinValueValidator
 from django.db import models, transaction
-from django.db.models import Q
+from django.db.models import F, Func, Q, Value
 from django.urls import reverse
 from django_lifecycle import AFTER_UPDATE, BEFORE_DELETE, hook
 from rest_framework.exceptions import APIException
 
-from pulpcore.app.util import batch_qs, get_view_name_for_model
+from pulpcore.app.util import batch_qs, get_url, get_view_name_for_model
 from pulpcore.constants import ALL_KNOWN_CONTENT_CHECKSUMS
 from pulpcore.download.factory import DownloaderFactory
 from pulpcore.exceptions import ResourceImmutableError
@@ -92,6 +92,22 @@ class Repository(MasterModel):
             super().save(*args, **kwargs)
             if adding:
                 self.create_initial_version()
+
+                # lock the repository if it was created from within a running task
+                try:
+                    task_id = environ["PULP_TASK_ID"]
+                except KeyError:
+                    return
+
+                repository_url = Value(get_url(self))
+                update_func = Func(
+                    F("reserved_resources_record"), repository_url, function="ARRAY_APPEND"
+                )
+                updated = Task.objects.filter(pk=task_id).update(
+                    reserved_resources_record=update_func
+                )
+                if not updated:
+                    raise RuntimeError(f"The repository '{self.name}' could not be locked")
 
     def create_initial_version(self):
         """
--- a/pulpcore/app/serializers/importer.py
+++ b/pulpcore/app/serializers/importer.py
@@ -102,6 +102,13 @@ class PulpImportSerializer(ModelSerializer):
         ),
         required=False,
     )
+    create_repositories = serializers.BooleanField(
+        help_text=_(
+            "If True, missing repositories will be automatically created during the import."
+        ),
+        required=False,
+        default=False,
+    )
 
     def _check_path_allowed(self, param, a_path):
         user_provided_realpath = os.path.realpath(a_path)
@@ -158,6 +165,12 @@ class PulpImportSerializer(ModelSerializer):
         if not data.get("path", None) and not data.get("toc", None):
             raise serializers.ValidationError(_("One of 'path' or 'toc' must be specified."))
 
+        if importer := self.context.get("importer"):
+            if importer.repo_mapping and data.get("create_repositories"):
+                raise serializers.ValidationError(
+                    _("The option 'create_repositories' is not compatible with 'repo_mapping'.")
+                )
+
         return super().validate(data)
 
     class Meta:
@@ -165,6 +178,7 @@ class PulpImportSerializer(ModelSerializer):
         fields = (
             "path",
             "toc",
+            "create_repositories",
         )
 
 
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -32,6 +32,7 @@ from pulpcore.app.models import (
 from pulpcore.app.modelresource import (
     ArtifactResource,
     ContentArtifactResource,
+    RepositoryResource,
 )
 from pulpcore.constants import TASK_STATES
 from pulpcore.tasking.tasks import dispatch
@@ -54,13 +55,15 @@ IMPORT_BATCH_SIZE = 100
 MAX_ATTEMPTS = 3
 
 
-def _destination_repo(importer, source_repo_name):
-    """Find the destination repository based on source repo's name."""
+def _get_destination_repo_name(importer, source_repo_name):
+    """
+    Return the name of a destination repository considering the mapping or source repository name.
+    """
     if importer.repo_mapping and importer.repo_mapping.get(source_repo_name):
         dest_repo_name = importer.repo_mapping[source_repo_name]
     else:
         dest_repo_name = source_repo_name
-    return Repository.objects.get(name=dest_repo_name)
+    return dest_repo_name
 
 
 def _impfile_iterator(fd):
@@ -179,21 +182,24 @@ def _check_versions(version_json):
         raise ValidationError((" ".join(error_messages)))
 
 
-def import_repository_version(importer_pk, destination_repo_pk, source_repo_name, tar_path):
+def import_repository_version(
+    importer_pk, src_repo_name, src_repo_type, dest_repo_name, dest_repo_pk, tar_path
+):
     """
     Import a repository version from a Pulp export.
 
     Args:
-        importer_pk (str): Importer we are working with
-        destination_repo_pk (str): Primary key of Repository to import into.
-        source_repo_name (str): Name of the Repository in the export.
-        tar_path (str): A path to export tar.
+        importer_pk (str): Importer we are working with.
+        src_repo_name (str): The name of the original repository.
+        src_repo_type (str): The Pulp's type of a repository.
+        dest_repo_name (str): The name of a repository where the content will be imported.
+        dest_repo_pk (str): The primary key of a destination repository if any
+        tar_path (str): The path of an exported tarball.
     """
-    dest_repo = Repository.objects.get(pk=destination_repo_pk)
     importer = PulpImporter.objects.get(pk=importer_pk)
 
     pb = ProgressReport(
-        message=f"Importing content for {dest_repo.name}",
+        message=f"Importing content for {dest_repo_name}",
         code="import.repo.version.content",
         state=TASK_STATES.RUNNING,
     )
@@ -204,29 +210,11 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
         with tarfile.open(tar_path, "r:gz") as tar:
             tar.extract(REPO_FILE, path=temp_dir)
 
-        with open(os.path.join(temp_dir, REPO_FILE), "r") as repo_data_file:
-            data = json.load(repo_data_file)
-
-        src_repo = next(repo for repo in data if repo["name"] == source_repo_name)
-
-        if dest_repo.pulp_type != src_repo["pulp_type"]:
-            raise ValidationError(
-                _(
-                    "Repository type mismatch: {src_repo} ({src_type}) vs {dest_repo} "
-                    "({dest_type})."
-                ).format(
-                    src_repo=src_repo["name"],
-                    src_type=src_repo["pulp_type"],
-                    dest_repo=dest_repo.name,
-                    dest_type=dest_repo.pulp_type,
-                )
-            )
-
         rv_name = ""
         # Extract the repo version files
         with tarfile.open(tar_path, "r:gz") as tar:
             for mem in tar.getmembers():
-                match = re.search(rf"(^repository-{source_repo_name}_[0-9]+)/.+", mem.name)
+                match = re.search(rf"(^repository-{src_repo_name}_[0-9]+)/.+", mem.name)
                 if match:
                     rv_name = match.group(1)
                     tar.extract(mem, path=temp_dir)
@@ -236,13 +224,21 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
 
         rv_path = os.path.join(temp_dir, rv_name)
         # Content
-        plugin_name = src_repo["pulp_type"].split(".")[0]
+        plugin_name = src_repo_type.split(".")[0]
         cfg = get_plugin_config(plugin_name)
 
         resulting_content_ids = []
         for res_class in cfg.exportable_classes:
+            if issubclass(res_class, RepositoryResource) and dest_repo_pk:
+                continue
+
             filename = f"{res_class.__module__}.{res_class.__name__}.json"
             for a_result in _import_file(os.path.join(rv_path, filename), res_class, retry=True):
+                if issubclass(res_class, RepositoryResource) and a_result.rows:
+                    repo_resource = a_result.rows[0]
+                    if repo_resource.import_type in ("new", "update"):
+                        dest_repo_pk = repo_resource.object_id
+
                 if issubclass(res_class, BaseContentResource):
                     resulting_content_ids.extend(
                         row.object_id
@@ -269,13 +265,15 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
         if mapping:
             # use the content mapping to map content to repos
             for repo_name, content_ids in mapping.items():
-                repo = _destination_repo(importer, repo_name)
+                repo_name = _get_destination_repo_name(importer, repo_name)
+                dest_repo = Repository.objects.get(name=repo_name)
                 content = Content.objects.filter(upstream_id__in=content_ids)
                 content_count += content.count()
-                with repo.new_version() as new_version:
+                with dest_repo.new_version() as new_version:
                     new_version.set_content(content)
         else:
             # just map all the content to our destination repo
+            dest_repo = Repository.objects.get(pk=dest_repo_pk)
             content = Content.objects.filter(pk__in=resulting_content_ids)
             content_count += content.count()
             with dest_repo.new_version() as new_version:
@@ -290,13 +288,17 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
     gpr.update(done=F("done") + 1)
 
 
-def pulp_import(importer_pk, path, toc):
+def pulp_import(importer_pk, path, toc, create_repositories):
     """
     Import a Pulp export into Pulp.
 
     Args:
         importer_pk (str): Primary key of PulpImporter to do the import
         path (str): Path to the export to be imported
+        toc (str): The path to a table-of-contents file describing chunks to be validated,
+            reassembled, and imported.
+        create_repositories (bool): Indicates whether missing repositories should be automatically
+            created or not.
     """
 
     def _compute_hash(filename):
@@ -483,18 +485,36 @@ def pulp_import(importer_pk, path, toc):
             gpr.save()
 
             for src_repo in data:
+                dest_repo_name = _get_destination_repo_name(importer, src_repo["name"])
+
                 try:
-                    dest_repo = _destination_repo(importer, src_repo["name"])
+                    dest_repo = Repository.objects.get(name=dest_repo_name)
                 except Repository.DoesNotExist:
-                    log.warning(
-                        "Could not find destination repo for {}. Skipping.".format(src_repo["name"])
-                    )
-                    continue
+                    if create_repositories:
+                        exclusive_resources = []
+                        dest_repo_pk = ""
+                    else:
+                        log.warning(
+                            "Could not find destination repo for {}. Skipping.".format(
+                                src_repo["name"]
+                            )
+                        )
+                        continue
+                else:
+                    exclusive_resources = [dest_repo]
+                    dest_repo_pk = dest_repo.pk
 
                 dispatch(
                     import_repository_version,
-                    exclusive_resources=[dest_repo],
-                    args=(importer.pk, dest_repo.pk, src_repo["name"], path),
+                    exclusive_resources=exclusive_resources,
+                    args=(
+                        importer.pk,
+                        src_repo["name"],
+                        src_repo["pulp_type"],
+                        dest_repo_name,
+                        dest_repo_pk,
+                        path,
+                    ),
                     task_group=task_group,
                 )
 
--- a/pulpcore/plugin/modelresources.py
+++ b/pulpcore/plugin/modelresources.py
@@ -1,0 +1,1 @@
+from pulpcore.app.modelresource import RepositoryResource  # noqa
--- a/CHANGES/1920.feature
+++ b/CHANGES/1920.feature
@@ -1,0 +1,3 @@
+Added an option for automatically creating repositories on the fly during an import procedure. The
+option is disabled by default. Enable it by setting the field ``create_repositories`` to ``True``
+via the REST API.
--- a/CHANGES/plugin_api/1920.feature
+++ b/CHANGES/plugin_api/1920.feature
@@ -1,0 +1,3 @@
+Exposed the ``RepositoryResource`` class to enable plugin writers to customize the way of
+importing/exporting of particular repository types. Repositories should be now a part of exported
+resources to enable automatic creation of missing repositories.
--- a/pulpcore/app/modelresource.py
+++ b/pulpcore/app/modelresource.py
@@ -56,6 +56,9 @@ class RepositoryResource(QueryModelResource):
             "pulp_created",
             "pulp_last_updated",
             "content",
+            "next_version",
+            "repository_ptr",
+            "remote",
         )
 
 
--- a/pulpcore/app/viewsets/importer.py
+++ b/pulpcore/app/viewsets/importer.py
@@ -101,17 +101,25 @@ class PulpImportViewSet(ImportViewSet):
         except PulpImporter.DoesNotExist:
             raise Http404
 
-        serializer = PulpImportSerializer(data=request.data, context={"request": request})
+        serializer = PulpImportSerializer(
+            data=request.data, context={"request": request, "importer": importer}
+        )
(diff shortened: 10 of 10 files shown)


docs/workflows/import-export.rst
```diff
@@ -366,11 +366,13 @@ to the names of repos in Pulp. For example, suppose the name of the repo in the
...
 
-.. note::
-   Pulp import expects destination repository/repositories to be present at the import process.
-
 After the importer is created, a POST request to create an import will trigger the import process.
 
+.. note::
+    By default, the Pulp import machinery expects destination repositories to be present at the time
+    of the import. This can be overridden by passing the ``create_repositories=True`` field via the
+    POST request.
```
ipanova, line 374: "..that will create missing repos on the fly"

ipanova (commented):
can you make a showcase with pulp_container plugin for this feature?

mdellweg:
Can you add "on import" to both the PR and the commit message first line?

pulpcore/app/tasks/importer.py
```diff
@@ -253,15 +236,16 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
...
-                repo = _destination_repo(importer, repo_name)
+                repo_name = _get_destination_repo_name(importer, repo_name)
+                dest_repo = Repository.objects.get(name=repo_name)
                 content = Content.objects.filter(upstream_id__in=content_ids)
-                with repo.new_version() as new_version:
-                    new_version.set_content(content)
         else:
             # just map all the content to our destination repo
+            dest_repo = Repository.objects.get(name=destination_repo_name)
             content = Content.objects.filter(pk__in=resulting_content_ids)
```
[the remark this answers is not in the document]
lubosmj: No, the said two are different things. `repo_mapping` is used for mapping repositories (a dictionary passed via the REST API as the mapping between repositories' names) whilst `mapping` "probably" represents the mapping retrieved from the `content_mapping.json` file (the mapping between content units and repositories (https://docs.pulpproject.org/pulpcore/plugins/plugin-writer/concepts/subclassing/import-export.html#content-mapping)) that is included within an exported tarball.

ipanova:
> can you make a showcase with pulp_container plugin for this feature?

This PR is waiting on ^

docs/plugins/plugin-writer/concepts/subclassing/import-export.rst
```diff
@@ -88,6 +88,27 @@ this::
...
+        A resource for import/export of file repository entities
+        """
+
+        class Meta:
+            model = FileRepository
+            exclude = RepositoryResource.Meta.exclude + ("repository_ptr",)
+
+
+    # the list signifying the order of imports must also include the repository resource class
+    IMPORT_ORDER = [FileContentResource, FileRepositoryResource]
```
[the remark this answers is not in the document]
ipanova: One could have a completely different assumption that first we need to import repository to have a place to place content into.
@ggainey can you confirm what's the correct order here?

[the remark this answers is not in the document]
lubosmj: Actually, the order of repositories does not matter. When we export repositories, we do not include any additional references to related objects (no remotes, no content units, ...). In comparison, in pulp_container, we export tags that reference manifests that reference blobs - so, in this case, we need to take care of the order.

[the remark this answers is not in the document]
ipanova: so what you're saying is - if we put repos as first in the order it would not make any difference.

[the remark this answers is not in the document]
ggainey: IMPORT_ORDER is, in fact, ordered, because some resources are dependent on the existence of others before they can be created. However, the act of adding content to a repo is a separate step that happens only after we've imported All The Things, so it's perfectly fine to create the repo last in the list.

ipanova:
@lubosmj how do we plan to create on the fly a repo which would be different type from the source repo?
Something we'd need to add in plugin or?

ggainey (changes requested):
I did a quick patch to pulp_rpm modelresource and saw the whole process Just Work, was definitely easy to implement at the plugin side.

I'd like to see the doc expanded to cover the "why repo-mapping and this switch aren't compatible" question.

pulpcore/app/tasks/importer.py
```diff
@@ -253,15 +236,16 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
         if mapping:
             # use the content mapping to map content to repos
             for repo_name, content_ids in mapping.items():
-                repo = _destination_repo(importer, repo_name)
+                repo_name = _get_destination_repo_name(importer, repo_name)
+                dest_repo = Repository.objects.get(name=repo_name)
                 content = Content.objects.filter(upstream_id__in=content_ids)
-                with repo.new_version() as new_version:
```
ggainey, line 258: I think we can't move this - has to be done per-entity-in-mapping, inside this loop, to pick up subrepos (like, say, RPM kickstart repositories)

ggainey (approved):
Reviewed, exercised in pulp_rpm via https://github.com/pulp/pulp_rpm/pull/2592 . Looks good to me!

docs/workflows/import-export.rst
```diff
@@ -366,11 +366,13 @@ to the names of repos in Pulp. For example, suppose the name of the repo in the
...
 
-.. note::
-   Pulp import expects destination repository/repositories to be present at the import process.
-
 After the importer is created, a POST request to create an import will trigger the import process.
 
+.. note::
+    By default, the Pulp import machinery expects destination repositories to be present at the time
+    of the import. This can be overridden by passing the ``create_repositories=True`` field via the
+    POST request that will force Pulp to create missing repositories on the fly.
```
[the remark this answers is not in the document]
mdellweg: This comment has not been processed yet.

pulpcore/app/tasks/importer.py
```diff
@@ -475,18 +472,39 @@ def validate_and_assemble(toc_filename):
...
                 try:
-                    dest_repo = _destination_repo(importer, src_repo["name"])
+                    dest_repo = Repository.objects.get(name=dest_repo_name)
                 except Repository.DoesNotExist:
-                    log.warning(
-                        "Could not find destination repo for {}. Skipping.".format(src_repo["name"])
-                    )
-                    continue
+                    if create_repositories:
+                        exclusive_resources = []
```
mdellweg, line 481: I think, we still need to lock on the newly created repository.
Either we create it here before we subtask is dispached, or we find a way to sneak the newly created repository in the exclusive_resources of the running task in the same transaction as we create it (this is unprecedented).

docs/workflows/import-export.rst
```diff
@@ -366,10 +366,17 @@ to the names of repos in Pulp. For example, suppose the name of the repo in the
...
-   Pulp import expects destination repository/repositories to be present at the import process.
+    By default, the Pulp import machinery expects destination repositories to be present at the time
+    of the import. This can be overridden by passing the ``create_repositories=True`` field via the
+    POST request that will lead Pulp to create missing repositories on the fly.
 
-After the importer is created, a POST request to create an import will trigger the import process.
+.. warning::
+    The options ``repo_mapping`` and ``create_repositories`` are not compatible with each other. The
+    existence of a repository specified in the ``repo_mapping`` option is tested before the importer
+    is initialized. Thus, the repository has to be already created in advance.
```
mdellweg, line 379: Do we have validation for this in the serializer?

lubosmj: No, I should add it.

pulpcore/app/tasks/importer.py
ggainey: We can't lock it "in advance", it doesn't exist until "at some point" in the processing of import-records we run the (Detail)RepositoryResource file through _import_file()

mdellweg: Lubos can you try if this would solve the problem?
https://github.com/pulp/pulpcore/pull/2881/files

lubosmj: @mdellweg, yes, it seems like a repository can be locked in that way.

Below is an example with 2 imported repositories:
```
(pulp) [vagrant@pulp3-source-fedora35 exports]$ http :24817/pulp/api/v3/importers/core/pulp/58df31e2-9ca9-4333-bf01-92b8c11fe494/imports/ create_repositories=True path=/tmp/exports/export-e0398973-e809-49ca-8a61-f52d6c2fc4a4-20220617_0912.tar.gz
HTTP/1.1 202 Accepted
Allow: GET, POST, HEAD, OPTIONS
Connection: close
Content-Length: 79
Correlation-ID: a6e8728d9cde4cf184a2db6aee4fecd2
Date: Tue, 21 Jun 2022 12:56:26 GMT
Referrer-Policy: same-origin
Server: gunicorn
Vary: Accept, Cookie
X-Frame-Options: DENY

{
    "task_group": "/pulp/api/v3/task-groups/0c37dfe7-bdcf-4957-85c5-347b066174d0/"
}


(pulp) [vagrant@pulp3-source-fedora35 exports]$ http :24817/pulp/api/v3/task-groups/0c37dfe7-bdcf-4957-85c5-347b066174d0/
HTTP/1.1 200 OK
Allow: GET, HEAD, OPTIONS
Connection: close
Content-Length: 1502
Correlation-ID: 43eaf18954fd479aa20ad7950f95849f
Date: Tue, 21 Jun 2022 12:56:34 GMT
Referrer-Policy: same-origin
Server: gunicorn
Vary: Accept, Cookie
X-Frame-Options: DENY

{
    "all_tasks_dispatched": true,
    "canceled": 0,
    "canceling": 0,
    "completed": 3,
    "description": "Import of /tmp/exports/export-e0398973-e809-49ca-8a61-f52d6c2fc4a4-20220617_0912.tar.gz",
    "failed": 0,
    "group_progress_reports": [
        {
            "code": "import.repo.versions",
            "done": 2,
            "message": "Importing repository versions",
            "suffix": null,
            "total": 2
        }
    ],
    "pulp_href": "/pulp/api/v3/task-groups/0c37dfe7-bdcf-4957-85c5-347b066174d0/",
    "running": 0,
    "skipped": 0,
    "tasks": [
        {
            "finished_at": "2022-06-21T12:56:26.960056Z",
            "name": "pulpcore.app.tasks.importer.pulp_import",
            "pulp_created": "2022-06-21T12:56:26.619315Z",
            "pulp_href": "/pulp/api/v3/tasks/d0c062c5-5ac2-45ed-a065-edfea494e8ed/",
            "started_at": "2022-06-21T12:56:26.662507Z",
            "state": "completed",
            "worker": "/pulp/api/v3/workers/fc68cff5-cad7-45ca-afb9-40b3cbea41b0/"
        },
        {
            "finished_at": "2022-06-21T12:56:28.069882Z",
            "name": "pulpcore.app.tasks.importer.import_repository_version",
            "pulp_created": "2022-06-21T12:56:26.928973Z",
            "pulp_href": "/pulp/api/v3/tasks/5cbbec24-4947-4a0d-8e88-bcb6e8393612/",
            "started_at": "2022-06-21T12:56:26.963631Z",
            "state": "completed",
            "worker": "/pulp/api/v3/workers/2a5f93cc-68a2-4bac-b1e8-04e69f0e8cec/"
        },
        {
            "finished_at": "2022-06-21T12:56:28.284207Z",
            "name": "pulpcore.app.tasks.importer.import_repository_version",
            "pulp_created": "2022-06-21T12:56:26.942847Z",
            "pulp_href": "/pulp/api/v3/tasks/d764aebf-5451-48ff-9c92-be01e2d84384/",
            "started_at": "2022-06-21T12:56:26.994319Z",
            "state": "completed",
            "worker": "/pulp/api/v3/workers/fc68cff5-cad7-45ca-afb9-40b3cbea41b0/"
        }
    ],
    "waiting": 0
}


(pulp) [vagrant@pulp3-source-fedora35 exports]$ http :24817/pulp/api/v3/tasks/5cbbec24-4947-4a0d-8e88-bcb6e8393612/
HTTP/1.1 200 OK
Allow: GET, PATCH, DELETE, HEAD, OPTIONS
Connection: close
Content-Length: 992
Correlation-ID: 2b5d33927ef5425c82822d475176a9b4
Date: Tue, 21 Jun 2022 12:56:41 GMT
Referrer-Policy: same-origin
Server: gunicorn
Vary: Accept, Cookie
X-Frame-Options: DENY

{
    "child_tasks": [],
    "created_resources": [
        "/pulp/api/v3/repositories/container/container-push/721b4126-f844-440d-9233-a7f37ba52627/versions/1/"
    ],
    "error": null,
    "finished_at": "2022-06-21T12:56:28.069882Z",
    "logging_cid": "a6e8728d9cde4cf184a2db6aee4fecd2",
    "name": "pulpcore.app.tasks.importer.import_repository_version",
    "parent_task": "/pulp/api/v3/tasks/d0c062c5-5ac2-45ed-a065-edfea494e8ed/",
    "progress_reports": [
        {
            "code": "import.repo.version.content",
            "done": 5,
            "message": "Importing content for pulp/fixture",
            "state": "completed",
            "suffix": null,
            "total": 5
        }
    ],
    "pulp_created": "2022-06-21T12:56:26.928973Z",
    "pulp_href": "/pulp/api/v3/tasks/5cbbec24-4947-4a0d-8e88-bcb6e8393612/",
        "/pulp/api/v3/repositories/container/container-push/721b4126-f844-440d-9233-a7f37ba52627/"
    ],
    "started_at": "2022-06-21T12:56:26.963631Z",
    "state": "completed",
    "task_group": "/pulp/api/v3/task-groups/0c37dfe7-bdcf-4957-85c5-347b066174d0/",
    "worker": "/pulp/api/v3/workers/2a5f93cc-68a2-4bac-b1e8-04e69f0e8cec/"
}


(pulp) [vagrant@pulp3-source-fedora35 exports]$ http :24817/pulp/api/v3/tasks/d764aebf-5451-48ff-9c92-be01e2d84384/
HTTP/1.1 200 OK
Allow: GET, PATCH, DELETE, HEAD, OPTIONS
Connection: close
Content-Length: 979
Correlation-ID: c10013e0651d44d98b6e80267265b705
Date: Tue, 21 Jun 2022 12:56:59 GMT
Referrer-Policy: same-origin
Server: gunicorn
Vary: Accept, Cookie
X-Frame-Options: DENY

{
    "child_tasks": [],
    "created_resources": [
        "/pulp/api/v3/repositories/container/container/6668db36-ae79-409d-97b1-01bd382c2893/versions/1/"
    ],
    "error": null,
    "finished_at": "2022-06-21T12:56:28.284207Z",
    "logging_cid": "a6e8728d9cde4cf184a2db6aee4fecd2",
    "name": "pulpcore.app.tasks.importer.import_repository_version",
    "parent_task": "/pulp/api/v3/tasks/d0c062c5-5ac2-45ed-a065-edfea494e8ed/",
    "progress_reports": [
        {
            "code": "import.repo.version.content",
            "done": 12,
            "message": "Importing content for lubosmj",
            "state": "completed",
            "suffix": null,
            "total": 12
        }
    ],
    "pulp_created": "2022-06-21T12:56:26.942847Z",
    "pulp_href": "/pulp/api/v3/tasks/d764aebf-5451-48ff-9c92-be01e2d84384/",
        "/pulp/api/v3/repositories/container/container/6668db36-ae79-409d-97b1-01bd382c2893/"
    ],
    "started_at": "2022-06-21T12:56:26.994319Z",
    "state": "completed",
    "task_group": "/pulp/api/v3/task-groups/0c37dfe7-bdcf-4957-85c5-347b066174d0/",
    "worker": "/pulp/api/v3/workers/fc68cff5-cad7-45ca-afb9-40b3cbea41b0/"
}

```

pulpcore/app/models/repository.py
```diff
@@ -92,6 +92,9 @@ def save(self, *args, **kwargs):
             super().save(*args, **kwargs)
             if adding:
                 self.create_initial_version()
+                if task := Task.current():
```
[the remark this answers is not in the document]
mdellweg: Does not matter it is all in the same transaction.

pulpcore/app/tasks/importer.py
mdellweg: > django-import-export has an after_import_instance() method that can be overridden. In the workflow, a row in an import-file is turned into an instance and saved; perhaps we can figure out the right way to get involved at this point in the process, and lock the created/updated repository at that point? Perhpas in https://github.com/lubosmj/pulpcore/blob/da635ebf6a445b20ed14ee8658cac0a482e6046b/pulpcore/app/modelresource.py#L50 ?

Would that come inside a transaction? We definitely need that. The current approach kind of assumes that whenever a task, for what ever reason creates a repository, it will own it until finished.

pulpcore/app/models/repository.py
```diff
@@ -92,6 +92,9 @@ def save(self, *args, **kwargs):
...
+                    task.reserved_resources_record.append(get_url(self))
+                    task.save()
```
mdellweg, line 97: @bmbouter @dkliban
Since i drafted this, i really want both of you to sign off this change.

pulpcore/app/tasks/importer.py
```diff
@@ -262,12 +253,14 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
         if mapping:
             # use the content mapping to map content to repos
             for repo_name, content_ids in mapping.items():
-                repo = _destination_repo(importer, repo_name)
+                repo_name = _get_destination_repo_name(importer, repo_name)
+                dest_repo = Repository.objects.get(name=repo_name)
```
mdellweg, line 257: Ok, i see a new problem now: If the Repository was created between the dispatching of the subtask and here, the task will not have a lock on the repository. So we need to change the logic a little bit. I think we need to tell the task (in it's args) whether it is supposed to _use_ an existing repository.

pulpcore/app/tasks/importer.py
```diff
@@ -174,21 +177,24 @@ def _check_versions(version_json):
         raise ValidationError((" ".join(error_messages)))
 
 
-def import_repository_version(importer_pk, destination_repo_pk, source_repo_name, tar_path):
+def import_repository_version(
+    importer_pk, destination_repo_name, source_repo, tar_path, create_repositories
```
mdellweg, line 181: If i'm not mistaken, this function is meant to be executed as task. So existing repositories must be passed by pk and not by name. The name should only be used for nonexistent repositories.
Reason: The resources are locked by their url (containing the pk which is unique and cannot be reused). However on the other hand, "the repository with the name 'foo'" could be a different thing when dispatching a task and when executing it.

mdellweg (changes requested):

pulpcore/app/models/repository.py
lubosmj: Can this be addressed now, please?

mdellweg: can we do `task.save(update_fields=["reserved_resources_record"])`?

pulpcore/app/tasks/importer.py
```diff
@@ -282,13 +281,25 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
     gpr.update(done=F("done") + 1)
 
 
-def pulp_import(importer_pk, path, toc):
+def ensure_locked_resources(resource):
+    """Ensure that the passed resource is locked in the currently running task."""
+    task = Task.current()
+    if url := get_url(resource) not in task.reserved_resources_record:
+        task.reserved_resources_record.append(url)
+        task.save()
```
mdellweg, line 289: Where did that come from? This cannot be. A task cannot just claim that it has the lock on certain resources.

mdellweg (changes requested):

pulpcore/app/tasks/importer.py
lubosmj: What is the proper way of locking a resource from within a task?

mdellweg: There is no way. This breaks all the assumptions of the tasking system. Resources must be requested when the task is dispatched.
The only viable exception is a resource newly created by that task if added withing the same transaction. That way, no other task could have claimed that new resource earlier.

pulpcore/app/models/repository.py
mdellweg: To add some context: This is the only way we can safely create a repository in one task and manipulate it further in the same one.

mdellweg:
```suggestion
                if task := Task.current():
                    task.reserved_resources_record.append(get_url(self))
                    task.save(update_fields=["reserved_resources_record"])
```

dralley: I'm not sure I understand why that is the case - this task is creating the repository, and repository pk's are UUIDs, so how would another task get a reference to it to manipulate it?  Are tasks fighting over the same work?

mdellweg: No, but the moment the repository appears in the database, a new task can be dispatched to modify it. We must ensure this other task is waiting until our task is finished.
The only way to do it is to add it to reserved resources.

pulpcore/app/tasks/importer.py
```diff
@@ -174,21 +177,22 @@ def _check_versions(version_json):
...
+def import_repository_version(importer_pk, dest_repo, source_repo, tar_path, create_repositories):
     """
     Import a repository version from a Pulp export.
 
     Args:
         importer_pk (str): Importer we are working with
-        destination_repo_pk (str): Primary key of Repository to import into.
-        source_repo_name (str): Name of the Repository in the export.
+        dest_repo (dict): The pk and name of a destination repository, if available.
+        source_repo (dict): Repository in the export.
```
bmbouter, line 187: Unless we need to handle N dest_repos (which I think we don't) then these args should be unpacked to not be dictionaries, but instead be like: `dest_repo_pk` and `dest_repo_name`. Similar with `source_repo`.

bmbouter: One reason is that the more we made arg data complex it becomes more complicated and harder to later add things like type checking support (because now you'd have to assert on the dict and its contents not just simple python primitives).

pulpcore/plugin/modelresources.py
```diff
@@ -0,0 +1 @@
+from pulpcore.app.modelresource import RepositoryResource  # noqa
```
bmbouter, line 1: :+1:

pulpcore/app/models/repository.py
bmbouter: I think it would be safer if it used the `update_fields` use. What if in between the `task := Task.current():` and the `save()` something else modifies it? I don't have an example of how, but if it did then there would be a race condition here.

bmbouter: Other than the request to make the save() call a little safer, I'm ok with this. At some point later I think we need to standardize the lock generation for all objects in Pulp. It's possible not all dispatchers for repositories are using `get_url()` as the string. That's a problem not for this PR thought.

@dralley @mdellweg @dkliban ^ this future improvement for the tasking system.

mdellweg: I don't quite follow. All dispatchers must agree on the same lock-string for the same resource. And we agreed to use the pulp_href. Yes, we should probably do the actual `get_url` call somewhere in the tasking code for a nice separation of concerns. But the result must be the same for proper locking of the resources.

mdellweg: AFAIK we never use the plain uuid (or worse pk). But we allow for arbitrary strings to be used (for abstract resources not in the pulp_db).
I'd be happy to discuss your suggestion in a coded example.

bmbouter: I just did a quick audit and almost all usage of `exclusive_resources` is uses the db object itself, e.g. `exclusive_resources=[repository]`. Is that the result of `get_url()`?

mdellweg: Yes, this will be translated by dispatch.

bmbouter: @mdellweg thanks, then I think we're in a good spot then. As a side curiosity, where does that translation happen in the code?

pulpcore/app/models/repository.py
```diff
@@ -92,6 +92,18 @@ def save(self, *args, **kwargs):
             super().save(*args, **kwargs)
             if adding:
                 self.create_initial_version()
+                self.lock_within_task()
+
+    def lock_within_task(self):
```
mdellweg, line 97: Can we not make this a method? It really absolutely _needs_ to be called from within the transaction of `save`. Making it a function will lead future us to think it's safe to be called.

pulpcore/app/models/repository.py
```diff
@@ -93,6 +93,18 @@ def save(self, *args, **kwargs):
...
+                # lock the repository if it was created from within a running task
+                try:
+                    task_id = environ["PULP_TASK_ID"]
+                except KeyError:
+                    return
+
+                repository_url = Value(get_url(self))
+                update_func = Func(
+                    F("reserved_resources_record"), repository_url, function="ARRAY_APPEND"
+                )
```
mdellweg, line 105:
```suggestion
                repository_url = get_url(self)
                update_func = Func(
                    F("reserved_resources_record"), Value(repository_url), function="ARRAY_APPEND"
                )
```
Feel free to ignore. Just a feeling the `F` and `Value` are kind of a similar thing.

pulpcore/app/models/repository.py
```diff
@@ -93,6 +93,18 @@ def save(self, *args, **kwargs):
...
+                Task.objects.filter(pk=task_id).update(reserved_resources_record=update_func)
```
mdellweg, line 106: Do you think we can use the return value to determine that one task was actually updated and raise an error if not?
Trying to make this really bullet proof.

lubosmj: Resolved.

pulpcore/app/models/repository.py
```diff
@@ -93,6 +93,22 @@ def save(self, *args, **kwargs):
...
+
+                repository_url = Value(get_url(self))
+                update_func = Func(
+                    F("reserved_resources_record"), repository_url, function="ARRAY_APPEND"
+                )
+                updated = Task.objects.filter(pk=task_id).update(
+                    reserved_resources_record=update_func
+                )
+                if not updated:
+                    raise RuntimeError(f"The repository '{self.name}' could not be locked")
```
mdellweg, line 110: :+1:

pulpcore/app/tasks/importer.py
```diff
@@ -204,29 +210,11 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
...
-                _(
-                    "Repository type mismatch: {src_repo} ({src_type}) vs {dest_repo} "
-                    "({dest_type})."
-                ).format(
-                    src_repo=src_repo["name"],
-                    src_type=src_repo["pulp_type"],
-                    dest_repo=dest_repo.name,
-                    dest_type=dest_repo.pulp_type,
-                )
-            )
```
lubosmj, line 223: @ggainey, @mdellweg, I removed this check completely because we will still raise an error (in the set_content method) in another place when we try to import content of an unknown type to a particular repository (e.g., rpm content to a file repository). Are you fine with that?

The actual problem is that we do not have the facility to get subclassed repository resources when running the export procedure from pulpcore. We do not use typed exports (similarly to: /repositories/container/container/) and we handle the generic repository export here: https://github.com/pulp/pulpcore/blob/c19d083aa86062cb581dc9a7252d38916b4c8ce1/pulpcore/app/importexport.py#L120. It is not possible to determine the type of repository resource at this time (Models (queryset) are not associated with import/export resources). As a result, even though I override the behaviour from pulp_container (where we export one repository type as another one; https://github.com/pulp/pulp_container/pull/826/files#diff-ad03c3a610de949dd025b92b32ea7684c7526e598829dbe85210b8520d98b16eR37-R49), it will be ignored because we always instantiate the generic repository resource (RepositoryResource). So, I am not able to export one repository type as another one because it will fail on the highlighted check during the import.

What this check does is that it halts the import procedure in case the types of repositories do not match before importing the actual files. Without this check, the halt will occur later.

mdellweg: I am just not an expert on import-export... But what i do not understand: If we always handle Repositories by means of the pulpcore RepositoryResource, how can we serialize detail-specific extra fields?

ggainey: to take advantage of this functionality, the plugins need to define their detail-repositories as model-resources - like this : https://github.com/pulp/pulp_rpm/pull/2592

lubosmj: Exactly, and we basically export the same thing two times. Once as "RepositoryResource.json" in the root directory of a tarball. And, another time we export it as "PluginRepositoryResource.json". We do not know the name of a repository resource beforehand from the current mechanism.

lubosmj: I would probably opt for opening a new task called "Making import/export a typed system" where we would refactor the whole machinery in order to allow plugin-based definitions and possible overrides (like we do for repositories, remotes, etc.). Because it is impossible to do that right now without extensively touching the pulpcore's code. Funny that I noticed these kinds of limitations at the end of testing the pulp_container import/export workflow - where first, I exported a push repo, removed it, created a new sync repo, and tried to import the exported push repo.

lubosmj: What is the path I should take - (1) leave it as it is, (2) outsource the import/export capabilities to concrete plugins where we are aware of exact import/export types for the whole time, (3) store the import/export resource classes in the database to be able to read resource types even from pulpcore?

ggainey: What we have now works with the import/export machinery as defined - "core handles the engine and knows the repos, the plugins handle the details of their objects". The disadvantage is, Repositories exist twice, once at root (so "the engine" can know what its doing) and again in the repo-directories (because that's where the plugin-exported-files live).  This is a little ugly, but that happens sometimes in plugin-architectures with cross-cutting features.

Is there an actual *problem* here that I'm missing? Because I don't see one other than "this could be more elegant", which is not a good reason to start refactoring machinery that is already in-use.

lubosmj: The only problem is that by removing the highlighted "pre-condition", we are losing the possibility to detect invalid repository types before running the actual import procedure (where we save content to the database). So, we will import a bunch of files and after that we will realize: "oopsie, this type of content cannot be added to such a repository, halt".

pulpcore/app/models/repository.py
ggainey, line 110: :+1:

pulpcore/app/serializers/importer.py
```diff
@@ -158,13 +165,20 @@ def validate(self, data):
         if not data.get("path", None) and not data.get("toc", None):
             raise serializers.ValidationError(_("One of 'path' or 'toc' must be specified."))
 
+        if importer := self.context.get("importer"):
+            if importer.repo_mapping and data.get("create_repositories"):
+                raise serializers.ValidationError(
+                    _("The option 'create_repositories' is not compatible with 'repo_mapping'.")
```
ggainey, line 171: :+1:

pulpcore/app/tasks/importer.py
ggainey: So if I am imagining this correctly - the failure-mode we're removing here, is only "the user has  created a destination-repo of one type, that doesn't match the type of the src-repo of the same name (or that is in a mapping as pointing-to the mismatched destination)". Current-code shortcircuits that failure, with this PR it has to wait until content-import. 

The major time-bottleneck for PIE is the Artifact step, which happens well before this. The actual content-load into django-importexport, is quite fast, even for large repositories (**memory utiilzation** is another issue,of course :) ). I think I'm OK with having a delayed failure here, given the gain in user-experience of this feature.

ggainey (approved):
It's possible there are Odd Edge Cases we haven't imagined - but it's because they'd all be plugin-specific. I am good with this modification to the engine; we'll learn more as we go.

mdellweg (approved):
LGTM

Issue #2715, opened 2022-05-10
Filter roles by their permissions


newswangerd:
**Is your feature request related to a problem? Please describe.**
Since you can only assign roles to objects if the role has a permission for the given object, the galaxy_ng UI needs to be able to query the roles api with a list of permissions and get back a set of roles that have any of the permissions the UI is requesting.

**Describe the solution you'd like**
Add a query parameter to the roles API that allows clients to get back a list of roles that have any permission in a list of permissions.

**Describe alternatives you've considered**
None

**Additional context**
None

PR #2716, opened 2022-05-11
Add permissions filter on roles api.

newswangerd:
Fixes: #2715

Example usage: `GET /pulp/api/v3/roles/?contains_permission=core.add_group&contains_permission=galaxy.create_namespace` returns any roles that have the permissions `core.add_group` or `galaxy.create_namespace`.


newswangerd:
@himdel fyi

pulpcore/app/viewsets/user.py
```diff
@@ -264,6 +271,22 @@ class Meta:
...
+        perms_list = self.request.query_params.getlist(name)
+
+        query = Q()
+
+        for perm in perms_list:
+            if "." in perm:
+                app_label, codename = perm.split(".", maxsplit=1)
+                query.add(Q(content_type__app_label=app_label, codename=codename), Q.OR)
+            else:
+                query.add(Q(codename=codename), Q.OR)
```
mdellweg, line 284: Is this any different from `query = query | Q(codename=codename)`?
Just curious.

newswangerd: I have no idea. The django `Q` object is still kind of black magic to me.

bmbouter: I am more familiar with the `|` queryset building with Q objects. I was looking at `[these django docs](https://docs.djangoproject.com/en/3.2/topics/db/queries/#additional-methods-to-handle-related-objects) and the `add` I think is not for Q objects par-se but for related objects for a specific query. Switching to the `|` style is preferable to me because it's the primary syntax documented [in this area](https://docs.djangoproject.com/en/3.2/topics/db/queries/#complex-lookups-with-q-objects).

mdellweg: I, too, would be more comfortable with the `Q | Q` version based on familiarity.

CHANGES/2715.feature
```diff
@@ -0,0 +1 @@
+Add contains_permission query parameter to the roles API that allows clients to get back a list of roles that have any permission in a list of permissions.
```
bmbouter, line 1: nitpick: please line wrap to 100 chars

pulpcore/tests/functional/api/test_role.py
```diff
@@ -0,0 +1,53 @@
...
+    def setUpClass(cls):
+        """Set up test."""
+        cls.cfg = config.get_config()
+        cls.client = api.Client(cls.cfg, api.json_handler)
+        cls.client = ApiClient(
+            configuration=cls.cfg.get_bindings_config(),
+        )
+
+        cls.cli_client = cli.Client(cls.cfg)
+        cls.role_api = RolesApi(cls.client)
```
bmbouter, line 22: Make a new fixture for the `RolesApi` because I don't see one exists already. Call it `ansible_roles_api_client` and put it here like these https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/tests/functional/conftest.py#L46-L49. Then have your test use it by declaring it as a [pytest fixture](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html).

pulpcore/tests/functional/api/test_role.py
```diff
@@ -0,0 +1,53 @@
...
+    @classmethod
+    def setUpClass(cls):
+        """Set up test."""
+        cls.cfg = config.get_config()
+        cls.client = api.Client(cls.cfg, api.json_handler)
+        cls.client = ApiClient(
+            configuration=cls.cfg.get_bindings_config(),
+        )
+
+        cls.cli_client = cli.Client(cls.cfg)
```
bmbouter, line 21: The CLI client you can get from [the `cli_client` fixture](https://github.com/pulp/pulp-smash/blob/main/pulp_smash/pulp3/pytest_plugin/__init__.py#L299)

pulpcore/tests/functional/api/test_role.py
newswangerd: @bmbouter this is for RBAC roles fyi

bmbouter: Ah this is slightly more complicated then as we desire pulpcore pytest fixtures to come from pulpcore. In that case would you be willing to contribute that fixture [here](https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/__init__.py#L52-L89)?

pulpcore/app/viewsets/user.py
newswangerd: Looks like it works

bmbouter (approved):
This is great! Thank you.

pulpcore/tests/functional/api/test_role.py
mdellweg: Yep, hiding is the right term. XD

mdellweg (approved):

PR #2719, opened 2022-05-11
Allow DeclarativeArtifacts to store multiple URLs for a single artifact.

### LINKED ISSUE #2175: As a plugin writer, I can provide many different possible sources for one DeclarativeArtifact
Instead of providing a single URL for a given DeclarativeArtifact, it should be possible to provide many, and have the sync pipeline will produce RemoteArtifacts accordingly.

Primary use case: mirrorlist support, where a user might expect that Pulp could fall back appropriately to other mirrors if a single mirror fails, rather than raising an error.  Currently there is no way to facilitate this inside the sync pipeline.

dralley:
closes #2175

### DIFF
--- a/CHANGES/plugin_api/2175.feature
+++ b/CHANGES/plugin_api/2175.feature
@@ -1,0 +1,1 @@
+DeclarativeArtifact now accepts a `urls` option which permits multiple URLs to be provided for a single artifact. If multiple URLs are provided, the download stage will try each of them in turn upon encountering failures.
--- a/pulpcore/plugin/stages/models.py
+++ b/pulpcore/plugin/stages/models.py
@@ -22,6 +22,7 @@ class DeclarativeArtifact:
             :class:`~pulpcore.plugin.models.Artifact` either saved or unsaved. If unsaved, it
             may have partial digest information attached to it.
         url (str): the url to fetch the :class:`~pulpcore.plugin.models.Artifact` from.
+        urls (List[str]): A list of many possible URLs to fetch the :class:`~pulpcore.plugin.models.Artifact` from.
         relative_path (str): the relative_path this :class:`~pulpcore.plugin.models.Artifact`
             should be published at for any Publication.
         remote (:class:`~pulpcore.plugin.models.Remote`): The remote used to fetch this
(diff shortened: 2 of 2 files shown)


dralley:
Ehhhh not so fast, RemoteArtifact only allows one per remote.  Since each can only store a single URL, we're still blocked.

https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/content.py#L705=

dralley:
What we could do here, and I think this would be reasonable for the short-term, is only pick off the first URL from the list when saving RemoteArtifacts.  So we wouldn't get the benefit of mirrors for on_demand (it would be functionally identical to how it currently works, since the rpm plugin currently just picks the first mirror and uses that), but immediate mode would benefit.

pulpcore/plugin/stages/models.py
```diff
@@ -35,19 +37,30 @@ class DeclarativeArtifact:
         specified and `artifact` doesn't have a file.
     """
 
-    __slots__ = ("artifact", "url", "relative_path", "remote", "extra_data", "deferred_download")
+    __slots__ = (
+        "artifact",
+        "url",
```
mdellweg, line 42:
```suggestion
```

mdellweg (commented):
If i understand correctly, a remote will then know about mirrorlists and provide the set of urls.
I am a bit concerned that this will now retry a certain amount of times with every single mirror in the list.

pulpcore/plugin/stages/models.py
dralley: A key point is that responses like 403, 404 are not retried.  Other 400 and 500 errors are retried and I think in that circumstance you are correct, but this doesn't seem particularly likely to me.

mdellweg: To be fair, it's a minor concern.

mdellweg (approved):

CHANGES/plugin_api/2175.feature
```diff
@@ -0,0 +1 @@
+DeclarativeArtifact now accepts a `urls` option which permits multiple URLs to be provided for a single artifact. If multiple URLs are provided, the download stage will try each of them in turn upon encountering failures.
```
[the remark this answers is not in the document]
dralley: @bmbouter Fixed

bmbouter (approved):
I believe this is backwards compatible, which is nice. This looks great!

CHANGES/plugin_api/2175.feature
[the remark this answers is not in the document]
bmbouter: ty!

PR #2722, opened 2022-05-12
Rewrite correlation-id test

mdellweg:
[noissue]


pulpcore/tests/functional/__init__.py
```diff
@@ -89,6 +90,18 @@ def content_guards_api_client(pulpcore_client):
...
 
+@pytest.fixture(scope="session")
+def orphans_cleanup_api_client(pulpcore_client):
+    return OrphansCleanupApi(pulpcore_client)
+
+
+@pytest.fixture
+def cid(pulpcore_client, monkeypatch):
+    cid = str(uuid.uuid4())
+    monkeypatch.setitem(pulpcore_client.default_headers, "Correlation-ID", cid)
```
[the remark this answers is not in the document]
mdellweg: I did that, and if we follow the example across plugins, i expect all tests to show with a single cid in the container log, while pytest will even expose what you are looking for on a failed test.

pulpcore/tests/functional/api/test_correlation_id.py
```diff
@@ -1,35 +1,10 @@
...
-    def test_correlation_id(self):
-        """Test that a correlation can be passed as a header and logged."""
-        response, status, headers = self.orphan_cleanup_api.cleanup_with_http_info({})
-        monitor_task(response.task)
-        task = self.task_api.read(response.task)
-        self.assertEqual(headers["Correlation-ID"], self.cid)
-        self.assertEqual(task.logging_cid, self.cid)
+def test_correlation_id(cid, tasks_api_client, orphans_cleanup_api_client):
+    """Test that a correlation can be passed as a header and logged."""
+    response, status, headers = orphans_cleanup_api_client.cleanup_with_http_info({})
```
bmbouter, line 6:
```suggestion
    response, status, headers = orphans_cleanup_api_client.cleanup({})
```

bmbouter (approved):
This is really nice. We should use this anywhere bindings are offered as pytest fixtures.

pulpcore/tests/functional/api/test_correlation_id.py
mdellweg: we need `_with_http_status` to assert on the `status` and `headers` returned.

mdellweg:
> This is really nice. We should use this anywhere bindings are offered as pytest fixtures.

With this change, we do at least for all pulpcore apis already.

dkliban (approved):

PR #2728, opened 2022-05-13
Update get_[users|groups]_with_perms to match guardian.

newswangerd:
This fixes two inconsistencies with the guardian interfaces:
- `for_concrete_model=False` makes it so that these functions don't work with proxy models (which guardian allowed)
- `Q(object_id=None)` causes users/groups with model permissions to be returned (which guardian didn't do)

@mdellweg do these changes make sense to you? I've only tested it with `get_groups_with_perms_attached_roles` (which is currently breaking stuff for us in galaxy_ng), so I'm not sure if these changes are going to cause problems for other plugins or not. If it does, we can just create our own function for this in galaxy_ng.

### DIFF
--- a/CHANGES/2739.bugfix
+++ b/CHANGES/2739.bugfix
@@ -1,0 +1,1 @@
+Add options to the role_util functions to make them work the same as guardian did.
--- a/pulpcore/app/role_util.py
+++ b/pulpcore/app/role_util.py
@@ -220,12 +220,17 @@ def get_objects_for_group(group, perms, qs, any_perm=False, accept_global_perms=
 
 
 def get_users_with_perms_roles(
-    obj, with_superusers=False, with_group_users=True, only_with_perms_in=None
+    obj,
+    with_superusers=False,
+    with_group_users=True,
+    only_with_perms_in=None,
+    include_model_permissions=True,
+    for_concrete_model=False,
 ):
     qs = User.objects.none()
     if with_superusers:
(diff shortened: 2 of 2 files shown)


newswangerd:
@mdellweg so we use proxy models for a couple of pulp container models so that we can set some extra methods on them, but we use the pulp container permissions for the actual permission checking. What this means is when I pass a proxy model object to these functions, I get back nothing because the actual permissions that we're using are from the original model, not the proxy.

I like the idea of adding these two things as parameters. That we we can keep the current functionality as it is to prevent these changes from breaking other plugins. If that sounds good to you, I can update the pr, add an issue and move it out of draft.

mdellweg:
> @mdellweg so we use proxy models for a couple of pulp container models so that we can set some extra methods on them, but we use the pulp container permissions for the actual permission checking. What this means is when I pass a proxy model object to these functions, I get back nothing because the actual permissions that we're using are from the original model, not the proxy.

That is a tragedy..., because i needed to add the proxy group model to be able to define permissions on it. That is why I think of it as working "with" the proxy model, while yours sounds like working "around" the proxy model. But anyway having the interpretations parameterized sounds fine to me.

mdellweg (commented):
I think you need to write it like `fixes #2739` in the commit message.

mdellweg (approved):

gerrod3 (approved):

PR #2740, opened 2022-05-16
Cleanups

dralley:


pulpcore/download/base.py
```diff
@@ -71,7 +71,6 @@ def __init__(
         url,
         expected_digests=None,
         expected_size=None,
-        semaphore=None,
```
dralley, line 74: @mdellweg @bmbouter Do you know of any users of this argument?  I looked at ansible, container and other plugins and couldn't find any.

mdellweg: Never seen it.

pulpcore/download/file.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, url, *args, **kwargs):
         """
         from pulpcore.app.serializers import RemoteSerializer
 
-        RemoteSerializer().validate_url(url)
+        RemoteSerializer.validate_url(url)
```
mdellweg, line 37: :+1:

pulpcore/download/base.py
```diff
@@ -83,19 +82,14 @@ def __init__(
...
         self.url = url
         self._writer = None
         self.path = None
         self.expected_digests = expected_digests
         self.expected_size = expected_size
-        if semaphore:
-            self.semaphore = semaphore
-        else:
-            self.semaphore = asyncio.Semaphore()  # This will always be acquired
+        self.semaphore = asyncio.Semaphore()  # This will always be acquired
```
mdellweg, line 92: I'm wondering if this is of any use when you are not providing the semaphore. You should only use the downloader once anyway. Maybe it's better to check that run is not called a second time.

pulpcore/download/base.py
mdellweg: I always heard, we are relying on the dl-limits from aiohttp.

dralley: We aren't, we have our own semaphores for that (which I want to get rid of if we can) - but this is for injecting your own semaphore which I don't think we actually use.

mdellweg (approved):

pulpcore/download/base.py
ipanova: ideally we should add a changelog entry but since you've checked that none of the plugins use it, i guess it is fine to not have any.

ipanova (approved):

pulpcore/download/base.py
bmbouter: I think [it was actually in use here](https://github.com/pulp/pulpcore/blob/3030e0f40da598dd9cca5505f0da79965fb53fd0/pulpcore/download/factory.py#L162), is it possible we just broke download concurrency with this? I don't believe we have download concurrency tests....

PR #2743, opened 2022-05-17
Convert CharField to TextField.

### LINKED ISSUE #2742: Switch from CharField to TextField

ipanova:
closes #2742


pulpcore/app/models/content.py
```diff
@@ -208,12 +208,12 @@ def storage_path(self, name):
 
     file = fields.ArtifactFileField(null=False, upload_to=storage_path, max_length=255)
     size = models.BigIntegerField(null=False)
-    md5 = models.CharField(max_length=32, null=True, unique=False, db_index=True)
-    sha1 = models.CharField(max_length=40, null=True, unique=False, db_index=True)
-    sha224 = models.CharField(max_length=56, null=True, unique=False, db_index=True)
-    sha256 = models.CharField(max_length=64, null=False, unique=True, db_index=True)
-    sha384 = models.CharField(max_length=96, null=True, unique=True, db_index=True)
-    sha512 = models.CharField(max_length=128, null=True, unique=True, db_index=True)
```
dralley, line 216: We do get more validation from this than we got from the fields in the RPM plugin, because these lengths are known and specific to the checksum type.  Probably still not worth it?  But it's more debatable.

pulpcore/app/models/progress.py
```diff
@@ -108,7 +108,7 @@ class ProgressReport(BaseModel):
     """
 
     message = models.TextField()
-    code = models.CharField(max_length=36)
```
dralley, line 111: Likewise I can kind of see some motivation for keeping this limited.

bmbouter: Since we don't know the max size, I think making this `TextField` makes sense (to me).

bmbouter (changes requested):
I left some comments requesting changes, but the whole PR is looking pretty great.

CHANGES/2742.misc
```diff
@@ -0,0 +1 @@
+Converted CharField to TextField.
```
bmbouter, line 1: Should this be `CHANGES/plugin_api/2742.misc` ? Also maybe use double backticks around the term?

dralley (approved):

bmbouter (approved):
Thank you!

Issue #2747, opened 2022-05-19
Filter roles by object type


newswangerd:
As a user I would like to filter roles on the roles API that only contain permissions for a specific object type.

PR #2748, opened 2022-05-19
Add ?for_object_type to roles API

newswangerd:
fixes: #2747

Example usage: http://localhost:8080/pulp/api/v3/roles/?for_object_type=/pulp/api/v3/groups/

Returns roles that only have permissions on the `/pulp/api/v3/groups/` resource type.

### DIFF
--- a/pulpcore/app/viewsets/base.py
+++ b/pulpcore/app/viewsets/base.py
@@ -157,6 +157,30 @@ class NamedModelViewSet(viewsets.GenericViewSet):
         return self.serializer_class
 
     @staticmethod
+    def get_resource_model(uri):
+        """
+        Resolve a resource URI to the model for the resource.
+
+        Provides a means to resolve an href passed in a POST body to an
+        model for the resource.
+
+        Args:
+            uri (str): A resource URI.
+
+        Returns:
+            django.models.Model: The model for the specified URI.
+
+        Raises:
+            rest_framework.exceptions.ValidationError: on invalid URI.
+        """
+        try:
+            match = resolve(urlparse(uri).path)
+        except Resolver404:
+            raise DRFValidationError(detail=_("URI not valid: {u}").format(u=uri))
+
+        return match.func.cls.queryset.model
+
+    @staticmethod
     def get_resource(uri, model=None):
         """
         Resolve a resource URI to an instance of the resource.
--- a/pulpcore/app/viewsets/user.py
+++ b/pulpcore/app/viewsets/user.py
@@ -5,7 +5,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import FieldError
 from django.shortcuts import get_object_or_404
 from django_filters.rest_framework import DjangoFilterBackend, filters
-from django.db.models import Q
+from django.db.models import Q, Count
 from django.contrib.auth.models import Permission
 
 from rest_framework import mixins, status
@@ -262,6 +262,10 @@ class RoleFilter(BaseFilterSet):
     FilterSet for Role.
     """
 
+    for_object_type = filters.CharFilter(
+        field_name="for_object_type", method="filter_for_object_type"
+    )
+
     contains_permission = filters.CharFilter(
         field_name="contains_permission", method="filter_contains_permission"
     )
@@ -288,7 +292,19 @@ class RoleFilter(BaseFilterSet):
 
         perms = Permission.objects.filter(query)
 
-        return queryset.filter(permissions__in=perms)
+        return queryset.filter(permissions__in=perms).distinct()
+
+    def filter_for_object_type(self, queryset, name, value):
+        obj_model = NamedModelViewSet.get_resource_model(value)
+        obj_type = ContentType.objects.get_for_model(obj_model, for_concrete_model=False)
+
+        perms = Permission.objects.filter(content_type__pk=obj_type.id)
+
+        query = Q(permissions__in=perms)
+
+        return queryset.annotate(
+            other_perms=Count("permissions", distinct=True, filter=~query)
+        ).filter(query, other_perms=0)
 
 
 @extend_schema_view(
@@ -301,7 +317,14 @@ class RoleFilter(BaseFilterSet):
                 required=False,
                 style="form",
                 explode=True,
-            )
+                description="Filter roles that have any of the permissions in the list.",
+            ),
+            OpenApiParameter(
+                name="for_object_type",
+                description=(
+                    "Filter roles that only have permissions for the specified object HREF."
+                ),
+            ),
         ],
     )
 )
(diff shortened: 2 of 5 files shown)


newswangerd:
I don't think the CI failure is related to this PR

pulpcore/tests/functional/__init__.py
```diff
@@ -117,6 +117,14 @@ def orphans_cleanup_api_client(pulpcore_client):
     return OrphansCleanupApi(pulpcore_client)
 
 
+@pytest.fixture
+def gen_role(roles_api_client, gen_object_with_cleanup):
```
bmbouter, line 121: @mdellweg is this the right place to introduce this fixture?

newswangerd: wasn't sure if they should go in `__init__.py` or `conftest.py`. I put it here because that's where all the other fixtures are. LMK if I should move it.

pulpcore/app/viewsets/base.py
```diff
@@ -168,6 +168,8 @@ def get_resource(uri, model=None):
             uri (str): A resource URI.
             model (django.models.Model): A model class. If not provided, the method automatically
                 determines the used model from the resource URI.
+            return_model (boolean): If true, this function will return the model class for the
+                resource and not an instance of it.
```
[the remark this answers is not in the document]
newswangerd: Done.

pulpcore/app/viewsets/base.py
```diff
@@ -156,6 +156,30 @@ def get_serializer_class(self):
...
+        Returns:
+            django.models.Model: The model for the specified URI.
+
+        Raises:
+            rest_framework.exceptions.ValidationError: on invalid URI.
+        """
+        try:
+            match = resolve(urlparse(uri).path)
+        except Resolver404:
+            raise DRFValidationError(detail=_("URI not valid: {u}").format(u=uri))
```
bmbouter, line 179: Maybe just let the initial 404 bubble up? Either way is fine w/ me.

pulpcore/app/viewsets/user.py
```diff
@@ -301,7 +317,14 @@ def filter_contains_permission(self, queryset, name, value):
                 required=False,
                 style="form",
                 explode=True,
-            )
+                description="Filter roles that have any of the permissions in the list.",
+            ),
+            OpenApiParameter(
+                name="for_object_type",
+                description=(
+                    "Filter roles that only have permissions" " for the specified object HREF."
```
bmbouter, line 325:
```suggestion
                    "Filter roles that only have permissions for the specified object HREF."
```

pulpcore/tests/functional/__init__.py
bmbouter: After thinking more on it, I think this is the right place because the `roles_api_client` is offered from pulpcore itself. @mdellweg lmk if you think differently.

@newswangerd I'm trying to get our factory pattern usage to have `_factory` in it so can this be 
```suggestion
def role_factory(roles_api_client, gen_object_with_cleanup):
```

pulpcore/tests/functional/__init__.py
```diff
@@ -117,6 +117,14 @@ def orphans_cleanup_api_client(pulpcore_client):
...
+    def _gen_role(**kwargs):
+        return gen_object_with_cleanup(roles_api_client, kwargs)
+
+    yield _gen_role
```
bmbouter, line 125:
```suggestion
    return _gen_role
```

If there isn't cleanup work to do after having it return is more efficient by not having the generator resumed later only to return.

bmbouter (changes requested):
I requested a few small changes. Thanks!

pulpcore/app/viewsets/base.py
newswangerd: @bmbouter I wanted to keep it consistent with `get_resource`.

pulpcore/tests/functional/api/test_role.py
```diff
@@ -27,3 +30,33 @@ def test_contains_permission_filter(roles_api_client):
...
     assert change_task_present and view_taskschedule_present
+
+
+def test_for_object_type_filter(roles_api_client, role_factory):
+    """Test for_object_type query parameter."""
+
+    group_href = constants.BASE_PATH + "groups/"
+
+    multiple = role_factory(
+        name="role_with_perms_for_multiple_objects",
```
mdellweg, line 41: We usually give them uuid4 names to be able to avoid conflicts.

pulpcore/tests/functional/api/test_role.py
```diff
@@ -27,3 +30,33 @@ def test_contains_permission_filter(roles_api_client):
...
+        description="test_role",
+        permissions=[
+            "core.add_group",
+        ],
+    )
+
+    # verify that roles with permissions for other objects aren't returned.
+    roles = roles_api_client.list(for_object_type=group_href, name=multiple.name)
+
+    assert roles.count == 0
```
mdellweg, line 57: This is interesting. It means we only return "pure" roles. I would have expected differently. But a good test reads like a specification. Well done!

How about adding roles with names
`"<uuid4>_multi_type"`: `["core.add_group", "core.view_task"]` (like yours)
`"<uuid4>_single_type"`: `["core.add_group", "core.view_group"]` (with more than one perm, just to be sure)
`"<uuid4>_empty"`: `[]` (should this be returned or not?)
and list them by `<...>.list(for_object_type=group_href, name__startswith="<uuid4>")` ?

mdellweg: I think the test would qualify for `@pytest.mark.parallel` at that point.

pulpcore/tests/functional/api/test_role.py
```diff
@@ -27,3 +33,72 @@ def test_contains_permission_filter(roles_api_client):
...
+    # check multiple roles
+    roles = roles_api_client.list(for_object_type=group_href, name__startswith=prefix)
+
+    assert roles.count == 2
+
+    returned_roles = set()
+    for role in roles.results:
+        returned_roles.add(role["name"])
+    
+    assert returned_roles == set([single_type.name, single_type2.name])
```
bmbouter, line 104: please add a newline at the end (the CI should be checking for this but currently isn't)

pulpcore/tests/functional/api/test_role.py
mdellweg: The docs are fine. I just didn't read it properly and assumed.

pulpcore/tests/functional/api/test_role.py
```diff
@@ -27,3 +33,67 @@ def test_contains_permission_filter(roles_api_client):
...
 
     assert change_task_present and view_taskschedule_present
+
+
+@pytest.mark.parallel
+def test_for_object_type_filter(roles_api_client, role_factory):
+    """Test for_object_type query parameter."""
+
+    group_href = constants.BASE_PATH + "groups/"
+    prefix = uuid.uuid4()
```
mdellweg, line 43:
```suggestion
    prefix = str(uuid.uuid4())
```
Sorry i keep forgetting that...

bmbouter (approved):
This looks great. Thank you!

PR #2759, opened 2022-05-23
Add a note about assigning model-level permissions

### LINKED ISSUE #2758: Add a note about passing `null` to `content_object`

lubosmj:
closes #2758

### DIFF
--- a/pulpcore/app/serializers/user.py
+++ b/pulpcore/app/serializers/user.py
@@ -236,7 +236,10 @@ class UserRoleSerializer(ModelSerializer, NestedHyperlinkedModelSerializer):
     )
 
     content_object = ContentObjectField(
-        help_text=_("Optional pulp_href of the object the permissions are to be asserted on."),
+        help_text=_(
+            "pulp_href of the object for which role permissions should be asserted. "
+            "If set to 'null', permissions will act on the model-level."
+        ),
         source="*",
         allow_null=True,
     )
@@ -300,7 +303,10 @@ class GroupRoleSerializer(ModelSerializer, NestedHyperlinkedModelSerializer):
     )
 
     content_object = ContentObjectField(
-        help_text=_("Optional pulp_href of the object the permissions are to be asserted on."),
+        help_text=_(
+            "pulp_href of the object for which role permissions should be asserted. "
+            "If set to 'null', permissions will act on the model-level."
+        ),
         source="*",
         allow_null=True,
     )
--- a/CHANGES/2758.doc
+++ b/CHANGES/2758.doc
@@ -1,0 +1,2 @@
+Added a note about explicitly setting ``content_object`` to null when assigning model-level
+permissions.


ipanova (dismissed):

lubosmj:
From the help message, I would assume that we wanted to allow administrators to assign roles to users without adding the restriction to specific objects.

Currently, it seems to me that we support only object-level permission assignments (via the REST API) and no one was complaining about that. Is that correct?

pulpcore/app/serializers/user.py
```diff
@@ -249,8 +250,7 @@ def validate(self, data):
             "object_id": None,
             "content_type": None,
         }
-        content_object = data["content_object"]
-        if content_object:
```
ipanova, line 253: @bmbouter this if condition suggests that content_object is not required.

pulpcore/app/serializers/user.py
```diff
@@ -287,6 +287,7 @@ class GroupRoleSerializer(ModelSerializer, NestedHyperlinkedModelSerializer):
         help_text=_("Optional pulp_href of the object the permissions are to be asserted on."),
         source="*",
         allow_null=True,
+        required=False,
```
mdellweg, line 290: This requirement is there for a reason. To assign a global model level permission by simply _forgetting_ to specify the object seemed to be too dangerous to me. So i decided to require explicitly specifying `"content_object": null,` in the POST request body. I agree the help message is not helping to that end.

ipanova: I see the intentions here, but this way it seems to be more confusing than helping, especially when this aspect is not documented :/

ipanova: We could have boolean `model_level_perms` set default to False and it would be incompatible with `content_object` not specified.

pulpcore/app/serializers/user.py
```diff
@@ -234,7 +234,10 @@ class UserRoleSerializer(ModelSerializer, NestedHyperlinkedModelSerializer):
     )
 
     content_object = ContentObjectField(
-        help_text=_("Optional pulp_href of the object the permissions are to be asserted on."),
+        help_text=_(
+            "Optional pulp_href of the object the permissions are to be asserted on. "
+            "Set to 'null' when assigning model-level permissions."
```
mdellweg, line 239: I'd like to get a native speaker involved here. It turned out that what was here was not even obvious to people i had explained it before. I'm still going to leave some suggestions:

```suggestion
            "The pulp_href of the object the permissions are to be asserted on. "
            "Set to 'null' for assigning model-level roles."
```
```suggestion
            "The pulp_href of the object the roles permissions are to be asserted on. "
            "If set to 'null', permissions will act on the model-level."
```

rationale:
- avoid "optional", it was too confusing
- assignment happens for roles, checking is done by permission

dralley: @mdellweg I agree this is a difficult thing to explain.  The new (suggested) phrasing sounds fine to me.  A slightly simpler variant might be:

> The pulp_href of the object for which role permissions should be asserted. If set to 'null', permissions will act on the model-level.

Does that sound reasonable?  This is subtle and I'm not as familiar with RBAC so I could be missing a small detail.

mdellweg (approved):
Thanks everyone for helping to improve the language.

ipanova (approved):

PR #2768, opened 2022-05-25
Expose a method for verifying existing content units

### LINKED ISSUE #2052: As a plugin writer, I want to have a function for touching content units
Author: @lubosmj (lmjachky)

Redmine Issue: 9419, https://pulp.plan.io/issues/9419

---

In the PR https://github.com/pulp/pulpcore/pull/1624, we introduced a method that uses `bulk_touch` for updating timestamps of content units. We should expose this method to all plugin writers (e.g., pulp_container currently implements the same method - this creates unnecessary duplicates).

lubosmj:
closes #2052

### DIFF
--- a/pulpcore/plugin/actions.py
+++ b/pulpcore/plugin/actions.py
@@ -1,4 +1,5 @@
 from gettext import gettext as _
+
 from drf_spectacular.utils import extend_schema
 from rest_framework.decorators import action
 from rest_framework.serializers import ValidationError
@@ -48,7 +49,7 @@ class ModifyRepositoryActionMixin:
             existing_content_units = Content.objects.filter(pk__in=content_units_pks)
             existing_content_units.touch()
 
-            self.verify_content_units(existing_content_units, add_content_units)
+            raise_for_unknown_content_units(existing_content_units, add_content_units)
 
             add_content_units = list(add_content_units.keys())
 
@@ -60,7 +61,7 @@ class ModifyRepositoryActionMixin:
                     remove_content_units[NamedModelViewSet.extract_pk(url)] = url
                 content_units_pks = set(remove_content_units.keys())
                 existing_content_units = Content.objects.filter(pk__in=content_units_pks)
-                self.verify_content_units(existing_content_units, remove_content_units)
+                raise_for_unknown_content_units(existing_content_units, remove_content_units)
                 remove_content_units = list(remove_content_units.keys())
 
         task = dispatch(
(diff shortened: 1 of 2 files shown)


pulpcore/plugin/actions.py
```diff
@@ -58,9 +54,8 @@ def modify(self, request, pk):
             else:
                 for url in request.data["remove_content_units"]:
                     remove_content_units[NamedModelViewSet.extract_pk(url)] = url
-                content_units_pks = set(remove_content_units.keys())
-                existing_content_units = Content.objects.filter(pk__in=content_units_pks)
-                self.verify_content_units(existing_content_units, remove_content_units)
+
+                touch_content_units(remove_content_units)
```
ipanova, line 58: there is no need to touch units if they are removed

pulpcore/app/util.py
```diff
@@ -188,3 +188,24 @@ def verify_signature(filepath, public_key, detached_data=None):
...
+    missing content units is raised.
+    """
+    content_units_pks = content_units.keys()
+    existing_content_units = models.Content.objects.filter(pk__in=content_units_pks)
+    existing_content_units.touch()
+
+    existing_content_units_pks = existing_content_units.values_list("pk", flat=True)
+    existing_content_units_pks = {str(pk) for pk in existing_content_units_pks}
+
+    missing_pks = content_units_pks - existing_content_units_pks
```
ipanova, line 206: you are taking away some flexibility here. plugin writers might not want to raise error if some units were not found and follow the best effort approach.
before that the validation was extracted to ` verify_content_units`

lubosmj: The error was still raised in the `verify_content_units` method (called from the original `modify` method). The plugin writers had to deal with this behaviour in the past.

With this in mind, it would be probably better to expose the method `verify_content_units` instead of `touch_content_units`. 🦆

pulpcore/plugin/actions.py
```diff
@@ -1,17 +1,18 @@
...
-from pulpcore.app.models import Content, RepositoryVersion
+from pulpcore.app.models import RepositoryVersion, Content
 from pulpcore.app.response import OperationPostponedResponse
 from pulpcore.app.serializers import (
     AsyncOperationResponseSerializer,
     RepositoryAddRemoveContentSerializer,
 )
 from pulpcore.app.viewsets import NamedModelViewSet
 from pulpcore.tasking.tasks import dispatch
+from pulpcore.exceptions.validation import ValidationError
```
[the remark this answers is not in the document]
lubosmj: Urgh, I have to use `from rest_framework.serializers import ValidationError`

pulpcore/plugin/actions.py
```diff
@@ -75,14 +76,24 @@ def modify(self, request, pk):
...
-        existing_content_units_pks = {str(pk) for pk in existing_content_units_pks}
-
-        missing_pks = set(all_content_units.keys()) - existing_content_units_pks
-        if missing_pks:
-            missing_hrefs = [all_content_units[pk] for pk in missing_pks]
-            raise ValidationError(
-                _("Could not find the following content units: {}").format(missing_hrefs)
-            )
+
+def raise_for_unknown_content_units(existing_content_units, content_units_pks_hrefs):
```
bmbouter, line 80: We could write a test for this, but I think it would be in test_repo_versions.py which is not pytest converted. We could add a test later (or now if you want). It's up to you.

bmbouter (approved):
This is looking good to me.

ipanova:
there is a test failing, can you rebase your PR?

PR #2778, opened 2022-05-29
Fix use of count() method for Content objects in remove_content()

phess:
As explained in https://bugzilla.redhat.com/show_bug.cgi?id=2091438 , app/models/repository.py uses `not content.count()` as a condition inside the `remove_content()` function. However, the `count()` method is not implemented for the `Content` class.

On the actual live system where this happened I modified the `not content.count()` test to do `not hasattr(content, '__iter__')` instead, assuming this condition in `repository.py` is there to check for non-iterable contents. The code, when modified like this, ran successfully and was able to successfully complete the pulp-2to3 migration, which was so far failing with a traceback that said,
~~~
'Content' object has no attribute 'count'
~~~

This PR changes just that: the test condition inside app/models/repository.py's `remove_content()`.

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -825,7 +825,7 @@ class RepositoryVersion(BaseModel):
         if self.complete:
             raise ResourceImmutableError(self)
 
-        if not content or not content.count():
+        if not content or not hasattr(content, '__iter__'):
             return
 
         # Normalize representation if content has already been added in this version.


ipanova:
hi @phess , thank you for the PR! It seems like the bug is on the migration side.
As you can see remove_content expects content query set https://github.com/pulp/pulpcore/blob/7b7f8b9e4717ccd1ab67e543afeb42a64774aa8e/pulpcore/app/models/repository.py#L818
However here in the migration plugin code we pass directly a content https://github.com/pulp/pulp-2to3-migration/blob/707518a1dac56ba4b5748ed69b934185c52e84a1/pulp_2to3_migration/app/migration.py#L418
Would you instead submit a PR against migration plugin?

phess:
> hi @phess , thank you for the PR! It seems like the bug is on the migration side. As you can see remove_content expects content query set
> 
> https://github.com/pulp/pulpcore/blob/7b7f8b9e4717ccd1ab67e543afeb42a64774aa8e/pulpcore/app/models/repository.py#L818
> 
> However here in the migration plugin code we pass directly a content https://github.com/pulp/pulp-2to3-migration/blob/707518a1dac56ba4b5748ed69b934185c52e84a1/pulp_2to3_migration/app/migration.py#L418
> Would you instead submit a PR against migration plugin?

Sure, that would be a pleasure. Thank you @ipanova so much for pinpointing the real issue for me.

Now, just one question: what is the cheapest way to turn a single Content object into a QuerySet so it behaves as intended? If I do as below, then this incurs an additional database query, which is usually undesirable if we can avoid it, right?
~~~
remove_content(Content.objects.filter(pk__in=cas_with_conflicts[0].content.pk))
~~~

Thanks in advance

PR #2793, opened 2022-06-01
Define a custom API root endpoint class

### LINKED ISSUE #2340: accesing /pulp/api/v3/ requires auth since 3.16, while didn't before
**Version**
3.16.3

**Describe the bug**
since 3.16 accessing `/pulp/api/v3/` requires auth, while before that it did not.

I *think* this is a side effect of https://github.com/pulp/pulpcore/pull/1593, just not sure if that was intentional or not.

**To Reproduce**
- setup a 3.16.3 isntall
- `curl -k https://localhost/pulp/api/v3/`

**Expected behavior**
Until 3.15, this would return 200, now it returns 403, unless I do provide auth.
Accessing `/pulp/api/v3/status/` still works fine without auth.

**Additional context**
Originally found while working on Puppet integration of 3.16+ in https://github.com/theforeman/puppet-pulpcore/pull/249

lubosmj:
closes #2340

### DIFF
--- a/pulpcore/app/urls.py
+++ b/pulpcore/app/urls.py
@@ -10,6 +10,7 @@ from drf_spectacular.views import (
     SpectacularSwaggerView,
 )
 from rest_framework_nested import routers
+from rest_framework.routers import APIRootView
 
 from pulpcore.app.apps import pulp_plugin_configs
 from pulpcore.app.views import OrphansView, PulpImporterImportCheckView, RepairView, StatusView
@@ -104,6 +105,19 @@ class ViewSetNode:
             return str(self.viewset)
 
 
+class PulpAPIRootView(APIRootView):
+    """A Pulp-defined APIRootView class with no authentication requirements."""
+
+    authentication_classes = []
+    permission_classes = []
+
+
+class PulpDefaultRouter(routers.DefaultRouter):
+    """A DefaultRouter class that benefits from the customized PulpAPIRootView class."""
+
+    APIRootView = PulpAPIRootView
+
+
 all_viewsets = []
 plugin_patterns = []
 # Iterate over each app, including pulpcore and the plugins.
@@ -117,9 +131,6 @@ sorted_by_depth = sorted(all_viewsets, key=lambda vs: vs._get_nest_depth())
 vs_tree = ViewSetNode()
 for viewset in sorted_by_depth:
     vs_tree.add_decendent(ViewSetNode(viewset))
-
-#: The Pulp Platform v3 API router, which can be used to manually register ViewSets with the API.
-root_router = routers.DefaultRouter()
 
 urlpatterns = [
     path(f"{API_ROOT}repair/", RepairView.as_view()),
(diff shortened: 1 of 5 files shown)


mdellweg (approved):

pulpcore/app/urls.py
```diff
@@ -104,6 +105,19 @@ def __repr__(self):
...
 
 
+class PulpAPIRootView(APIRootView):
+    """A Pulp-defined APIRootView class with no authentication requirements."""
+
+    authentication_classes = []
+    permission_classes = []
+
+
+class PulpDefaultRouter(routers.DefaultRouter):
```
bmbouter, line 115: Could a test also be written that forms the URL and requesting it with aiohttp with no auth. Here are some similar tests I've written for this.

https://github.com/pulp/pulpcore/pull/2656/files#diff-4ba773aa1c5a9462bf450deb5056971d3bd6609dd44b6c8b11733cd4c27dc670R50-R51

https://github.com/pulp/pulpcore/pull/2656/files#diff-4ba773aa1c5a9462bf450deb5056971d3bd6609dd44b6c8b11733cd4c27dc670R33-R35

bmbouter: Actually this comment was for the `PulpAPIRootView` instead.

bmbouter (changes requested):
Would you be willing to write a pytest test?

pulpcore/app/urls.py
```diff
@@ -184,6 +195,8 @@ def __repr__(self):
     )
 )
 
+#: The Pulp Platform v3 API router, which can be used to manually register ViewSets with the API.
```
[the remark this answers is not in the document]
lubosmj: I wanted to have the variable definition close to the code that is actually using the variable.

pulpcore/app/urls.py
lubosmj: Sure thing!

pulpcore/tests/functional/utils.py
```diff
@@ -7,6 +9,18 @@
 from pulpcore.client.pulpcore import ApiClient
 
 
+class NoAuth(requests.auth.AuthBase):
```
bmbouter, line 12: We're actually trying to not have requests as a test dependency and instead consolidate on aiohttp.client. Could you use that to make the requests instead please?

pulpcore/tests/functional/__init__.py
```diff
@@ -249,6 +249,11 @@ def pulp_api_v3_path(cli_client):
     return v3_api_root
 
 
+@pytest.fixture(scope="session")
+def pulp_api_v3_root_endpoint(pulp_cfg, pulp_api_v3_path):
```
bmbouter, line 253:
```suggestion
def pulp_api_v3_url(pulp_cfg, pulp_api_v3_path):
```

This is a tricky one to name, but this name made more sense to me.

pulpcore/tests/functional/api/test_root_endpoint.py
```diff
@@ -0,0 +1,11 @@
+import asyncio
+import pytest
+
+from pulpcore.tests.functional.utils import get_response
+
+
+@pytest.mark.parallel
+@pytest.mark.from_pulpcore_for_all_plugins
```
bmbouter, line 8: I'm not sure plugins running this test also created additional value. I don't think their code is in a position to modify pulpcore's routers.

bmbouter (approved):
I think removing that one mark (see comment) would be good, but otherwise this is :100:

Issue #2796, opened 2022-06-02
Add new global access condition that requires non-admin users specify repository on content upload


gerrod3:
With the new content queryset-scoping, non-admin users will no longer be able to see orphaned-content, so any content they upload not in a repository will become invisible to them. Add a new global access condition that can be added to content access policies to prevent non-admin users from uploading without an attached repository.

PR #2797, opened 2022-06-02
Add new access condition for content uploads

gerrod3:
fixes: #2796

### DIFF
--- a/CHANGES/plugin_api/2796.feature
+++ b/CHANGES/plugin_api/2796.feature
@@ -1,0 +1,3 @@
+Added new access condition ``has_required_repo_perms_on_upload`` for RBAC plugins to use to require
+users to specify a repository when uploading content. If not used when uploading content, non-admin
+users will not be able to see their uploaded content if queryset scoping is enabled.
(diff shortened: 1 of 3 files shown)


bmbouter (approved):

pulpcore/app/global_access_conditions.py
```diff
@@ -398,6 +398,47 @@ def has_repo_or_repo_ver_param_model_or_obj_perms(request, view, action, permiss
...
+    parameter when performing a content upload.
+
+    This is usable as a conditional check in an AccessPolicy for content uploads. Here is an
+    example checking for the "file.view_filerepository" permission.
+
+    ::
+
+        {
+            ...
+            "condition": "has_required_repo_perms_on_upload:file.view_filerepository",
```
ipanova, line 416: can you update this to also contain modify_repo_perm to reflect the reality?

CHANGES/plugin_api/2796.feature
```diff
@@ -0,0 +1,2 @@
+Added new access condition ``has_required_repo_perms_on_upload`` for RBAC plugins to use to require
+users to specify a repository when uploading content.
```
ipanova, line 2: Can you add a clarification in this changelog that if this condition is not used then users who are not admins won't be able to see that content they have uploaded into pulp?
I also think we should add a note about this in the plugin writer docs, wdyt?

ipanova: given that queryset_scoping is enabled ^

ipanova (approved):

PR #2803, opened 2022-06-06
`random_artifact` fixture for pulpcore

bmbouter:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -247,3 +247,18 @@ def pulp_api_v3_path(cli_client):
             "This fixture requires the server to have the `V3_API_ROOT` setting set."
         )
     return v3_api_root
+
+
+@pytest.fixture
+def random_artifact(random_artifact_factory):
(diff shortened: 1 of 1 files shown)


pulpcore/tests/functional/__init__.py
```diff
@@ -247,3 +248,11 @@ def pulp_api_v3_path(cli_client):
             "This fixture requires the server to have the `V3_API_ROOT` setting set."
         )
     return v3_api_root
+
+
+@pytest.fixture
+def random_artifact(artifacts_api_client, tmp_path, gen_object_with_cleanup):
+    with tempfile.NamedTemporaryFile(dir=tmp_path) as tmp_file:
```
mdellweg, line 255: You can use
```python
temp_file = tmp_path / str(uuid.uuid4())
temp_file.write_bytes(uuid.uuid4().bytes)
# and then probably
return gen_object_with_cleanup(artifacts_api_client, temp_file)
```

mdellweg (approved):

dkliban (approved):

PR #2807, opened 2022-06-06
Fixes OpenAPI schema for Content create operations

### LINKED ISSUE #2806: OpenAPI schema for Content create operation is wrong
The OpenAPI schema for Content create operation causes the generated client to have a signature that looks like this:

```
    def create(self, relative_path, **kwargs):  # noqa: E501
        """Create a file content  # noqa: E501
```

All other create operations have a signature like this:

```
    def create(self, file_file_remote, **kwargs):  # noqa: E501
        """Create a file remote  # noqa: E501
```

dkliban:
The code[0] for special handling of FileFields when generating the OpenAPI schema was
introduced to work around a bug in drf-yasg[1]. This code was then ported when pulpcore
switched to using drf_spectacular for generating OpenAPI. This code is not needed. It
actually causes the OpenAPI schema for Content creation APIs to be different from all
other create operations.

[0] https://github.com/pulp/pulpcore/commit/24b50710201a5fea73898f8a1fcff286f9e47809#diff-0535b5e0e76e73cdc44991f38a187114c123eb5ba125bda3015ca222a2b18088
[1] https://github.com/axnsan12/drf-yasg/issues/386

closes: #2806

### DIFF
--- a/CHANGES/2806.bugfix
+++ b/CHANGES/2806.bugfix
@@ -1,0 +1,1 @@
+Fixed the OpenAPI schema generation for Content create operations.
--- a/pulpcore/openapi/__init__.py
+++ b/pulpcore/openapi/__init__.py
@@ -13,7 +13,6 @@ from drf_spectacular.plumbing import (
     build_basic_type,
     build_parameter_type,
     build_root_object,
-    force_instance,
     normalize_result_object,
     resolve_django_path_parameter,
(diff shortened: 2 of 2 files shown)


dkliban:
Even though there is an inconsistency between the APIs, the fix that makes it consistent makes the client not usable.

PR #2812, opened 2022-06-08
Clear up docs about orphan protection time default

### LINKED ISSUE #2810: Doc for default orphan protection time seems conflicting
Reading [the rest api docs for orphan cleanup endpoint](https://docs.pulpproject.org/pulpcore/restapi.html#operation/orphans_cleanup_cleanup) it seems there are conflicting messages. It says both that the default for `orphan_protection_time` is 0 and if `orphan_protection_time` is not specified, ORPHAN_PROTECTION_TIME will be used.

daviddavis:
fixes #2810

### DIFF
--- a/CHANGES/2810.bugfix
+++ b/CHANGES/2810.bugfix
@@ -1,0 +1,1 @@
+Fixed docs regarding the default for orphan protection time.
--- a/pulpcore/app/serializers/orphans.py
+++ b/pulpcore/app/serializers/orphans.py
@@ -2,8 +2,6 @@ from gettext import gettext as _
 
 from rest_framework import fields, serializers
 
-from django.conf import settings
-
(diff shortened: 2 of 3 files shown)


daviddavis:
I'm mixed about whether this is the right change. The docs on pulpproject.org say `0` as the default because that's what is used in the CI to generate the REST API docs. If the docs are generated locally, it should correctly show `ORPHAN_PROTECTION_TIME`. That said, most people probably read the REST API docs on pulpproject.org so maybe this change is worth merging?

mdellweg (approved):
If it's just about the docs, maybe we can not set that in setting in the docs runner on gha. Having it the way it is will report the configured default in the api.json on an installation, which is also worth something.
Either way is fine to me.

ipanova (approved):

PR #2814, opened 2022-06-09
Raise runtime errors

### LINKED ISSUE #2817: An error is not raised when exporting remote artifacts

lubosmj:
closes #2817

### DIFF
--- a/CHANGES/2817.bugfix
+++ b/CHANGES/2817.bugfix
@@ -1,0 +1,1 @@
+Started showing errors when users try to export remote artifacts.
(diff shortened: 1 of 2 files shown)


mdellweg (commented):
Oh wow...
Is this changing already released code? We need an issue then.

fao89 (approved):

mdellweg (approved):

ipanova (approved):

ipanova:
@ggainey this might need to be backported to some older branches, wdyt?

ggainey:
> @ggainey this might need to be backported to some older branches, wdyt?

"Ugh" is what I think. Um, I know that we get this exception raised when exporting on-demand-only content - so at least *one* codepath does the right thing. Of all the issues I've seen from users of export, "silently failed to deal with on-demand artifacts" has not been one of them. But the code is def wrong, and the bahvior if we *don't* catch this at export-time is going result in a basically-useless export - so I'd say we should backport, at least to 3.14 and 3.16, which both have active disconnected users.

PR #2815, opened 2022-06-09
Add dump-permissions management command

mdellweg:


mdellweg:
currently the output looks like
```
# ==== User object permissions ====
# user, object, permission
admin, core.task:aeb6159b-e3a3-40b5-b113-fa11815feda8, core.delete_task
admin, core.task:aeb6159b-e3a3-40b5-b113-fa11815feda8, core.view_task
# ==== Group object permissions ====
# group, object, permission
Huibuh, core.task:aeb6159b-e3a3-40b5-b113-fa11815feda8, core.delete_task
```

gerrod3 (approved):

CHANGES/2741.feature
```diff
@@ -0,0 +1,2 @@
+Added dump-permissions management command to list deprecated permissions not yet translated into
+roles.
```
ipanova, line 2: I would whether add this info here to the changelog or maybe find a place somewhere in the docs which would say - that it would be the only way how to get hands onto the deprecated permissions once upgraded to 3.20

pulpcore/app/management/commands/dump-permissions.py
```diff
@@ -0,0 +1,236 @@
+from gettext import gettext as _
+from functools import lru_cache
+import json
+import sys
+
+from django.db import connection
+from django.core.management import BaseCommand, CommandError
+from django.contrib.contenttypes.models import ContentType
+
+from pulpcore.tasking.util import get_url
```
ipanova, line 10: can this be moved to app/util.py instead?

pulpcore/app/management/commands/dump-permissions.py
```diff
@@ -0,0 +1,236 @@
...
+                object_pk,
+                permission_pks,
+                app_label,
+                model,
+                username,
+                permissions,
+            ) = row
+            try:
+                obj = _get_url(content_type_id, object_pk)
+            except Exception:
```
ipanova, line 115: do you think we can narrow this exception down? can anything else be raised except for object-doesnot-exist?

ipanova (approved):

pulpcore/app/management/commands/dump-permissions.py
mdellweg: You mean the `get_url` itself?

ipanova: yes

ipanova (approved):
looks greaT!

Issue #2802, opened 2022-06-06
Duplicate SQL query for `core_accesspolicy` when accessing content view (since PR #2727?)


gmbnomis:
**Version**

pulpcore main branch (nightly run of pulp-cookbook fails since May 20th 2022)

**Describe the bug**

pulp-cookbook has a test, which among other things, counts the SQL queries when accessing the content view list (to verify that prefetching artifacts works).

Since May 20th 2022, this test fails because there is an additional query (for example: https://github.com/pulp/pulp_cookbook/runs/6749697106?check_suite_focus=true#step:13:1312). This alone isn't a problem, but apparently one query is just a duplicate:

```
[{'sql': 'SELECT "auth_user"."id", "auth_user"."password", '
         '"auth_user"."last_login", "auth_user"."is_superuser", '
         '"auth_user"."username", "auth_user"."first_name", '
         '"auth_user"."last_name", "auth_user"."email", '
         '"auth_user"."is_staff", "auth_user"."is_active", '
         '"auth_user"."date_joined" FROM "auth_user" WHERE '
         '"auth_user"."username" = \'admin\' LIMIT 21',
  'time': '0.001'},
 {'sql': 'SELECT "core_accesspolicy"."pulp_id", '
         '"core_accesspolicy"."pulp_created", '
         '"core_accesspolicy"."pulp_last_updated", '
         '"core_accesspolicy"."creation_hooks", '
         '"core_accesspolicy"."statements", '
         '"core_accesspolicy"."viewset_name", '
         '"core_accesspolicy"."customized", '
         '"core_accesspolicy"."queryset_scoping" FROM "core_accesspolicy" '
         'WHERE "core_accesspolicy"."viewset_name" = '
         "'content/cookbook/cookbooks' LIMIT 21",
  'time': '0.001'},
 {'sql': 'SELECT "core_accesspolicy"."pulp_id", '
         '"core_accesspolicy"."pulp_created", '
         '"core_accesspolicy"."pulp_last_updated", '
         '"core_accesspolicy"."creation_hooks", '
         '"core_accesspolicy"."statements", '
         '"core_accesspolicy"."viewset_name", '
         '"core_accesspolicy"."customized", '
         '"core_accesspolicy"."queryset_scoping" FROM "core_accesspolicy" '
         'WHERE "core_accesspolicy"."viewset_name" = '
         "'content/cookbook/cookbooks' LIMIT 21",
  'time': '0.000'},
[...]

PR #2826, opened 2022-06-10
Reduce queries for AccessPolicy

gerrod3:
fixes: #2802

Also fixed the comments on some of the methods of NamedModelViewSet.

### DIFF
--- a/pulpcore/app/access_policy.py
+++ b/pulpcore/app/access_policy.py
@@ -1,3 +1,4 @@
+from functools import lru_cache
 from rest_access_policy import AccessPolicy
 from rest_framework.exceptions import APIException
 
@@ -11,6 +12,7 @@ class AccessPolicyFromDB(AccessPolicy):
     """
 
     @staticmethod
+    @lru_cache
     def get_access_policy(view):
         """
         Retrieves the AccessPolicy from the DB or None if it doesn't exist.
--- a/CHANGES/2802.bugfix
+++ b/CHANGES/2802.bugfix
@@ -1,0 +1,1 @@
+Reduced duplicate SQL queries for ``AccessPolicy`` when accessing any view.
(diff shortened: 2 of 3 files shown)


pulpcore/app/access_policy.py
```diff
@@ -11,6 +12,7 @@ class AccessPolicyFromDB(AccessPolicy):
     """
 
     @staticmethod
+    @lru_cache
```
mdellweg, line 15: :1st_place_medal: 
The problem here is, we might need to flush that cache once the ap changes. And that all boils down to i have no idea how long the python process lives inside the gunicorn worker.

gerrod3: Actually we don't have to because it caches based on the view's memory address, not the view itself. So each new request will still trigger the database lookup, just the multiple calls to this method during that request will now be cached.

mdellweg: Perfect!

mdellweg (approved):

bmbouter (approved):

pulpcore/app/access_policy.py
dralley: @gerrod3 @mdellweg Are we sure about that?

https://bugzilla.redhat.com/show_bug.cgi?id=2122872#c49

(you'll need to log into BZ - also see the comments afterwards)

mdellweg: Do we guarantee live updating access policies? Can we just cache on the view name?
Or should we move this logic to the view `__init__` (I think that gives us what we think we had here.)?

dralley: For the sake of the BZ (which has a very high priority score), I'm going to go ahead and remove the cache first, verify that it solves the issue, and then try to work out how to add it back afterwards.

PR #2836, opened 2022-06-13
Do not delete a remote when it is in use

### LINKED ISSUE #1975: Deleting a remote used as source for live content corrupts ContentArtifact records
Author: @dralley (dalley)

Redmine Issue: 8305, https://pulp.plan.io/issues/8305

---

* Create a repository and on_demand remote, and sync them.
* Delete the remote

The deletion of the Remote deletes the RemoteArtifacts, leaving behind ContentArtifact attached to neither Artifacts nor Remotes, making them effectively corrupted and unpublishable.

~~~ python
1. create repository, remote
remote = remote_api.create(gen_file_remote(policy='on_demand'))
repo = repo_api.create(gen_repo())

1. sync the repository
repository_sync_data = RepositorySyncURL(remote=remote.pulp_href)
sync_response = repo_api.sync(repo.pulp_href, repository_sync_data)
task = monitor_task(sync_response.task)

1. delete the remote
monitor_task(remote_api.delete(remote.pulp_href).task)

1. ^---- problem occurs here, now RemoteArtifacts deleted, now ContentArtifact is broken

publish_response = publications_api.create({"repository_version": task.created_resources[0]})
monitor_task(publish_response.task)  # boom publish failure
~~~

This is more pernicious because content units can move throughout repositories, and if the remote is ever deleted, every repo can be broken at once with no safeguards.

lubosmj:
As of this commit, users cannot delete a remote that is referenced
by repository versions synced via the on_demand policy.

One can delete the remote only if the affected repository versions
were removed or there is a redundancy of remote artifacts.

Required PR: https://github.com/pulp/pulp_file/pull/756
closes #1975


pulpcore/app/models/repository.py
```diff
@@ -499,6 +500,17 @@ def invalidate_cache(self):
             if base_paths:
                 Cache().delete(base_key=base_paths)
 
+    @hook(BEFORE_DELETE)
+    def assert_multiple_remote_artifacts(self):
+        """Raise an error if no remaining remote artifacts for a single content unit were found."""
+        remote_artifacts = RemoteArtifact.objects.filter(remote=self.pk)
+        content_artifacts = ContentArtifact.objects.filter(
+            remoteartifact__in=remote_artifacts
+        ).prefetch_related("remoteartifact_set")
```
dralley, line 509: I think you should be able to calculate this more efficiently with `annotate()` and `Count()`

https://docs.djangoproject.com/en/4.0/topics/db/aggregation/

lubosmj:
```
In [1]: Remote.objects.all()
Out[1]: <QuerySet [<Remote (pulp_type=file.file): remo11>, <Remote (pulp_type=file.file): remo1>]>

In [2]: remo11 = Remote.objects.get(name="remo11")

In [3]: remote_artifacts = RemoteArtifact.objects.filter(remote=remo11)

In [4]: remote_artifacts
Out[4]: <RemoteArtifactQuerySet [<RemoteArtifact: pk=eb529fe2-84c8-44ce-a375-f836c199d66b>, <RemoteArtifact: pk=92397dc1-5baf-44a4-acca-eee281f8d412>, <RemoteArtifact: pk=8b3aa1f7-c033-4569-b40f-dda8c11f57c9>]>

In [5]: content_artifacts = ContentArtifact.objects.filter(remoteartifact__in=remote_artifacts).prefetch_related("remo
   ...: teartifact_set")

In [6]: content_artifacts
Out[6]: <QuerySet [<ContentArtifact: pk=0ded4d8d-6656-4cf4-a6ea-86a57a26721c>, <ContentArtifact: pk=a82e2a2d-2da3-4311-8701-73ad0dfb8f2a>, <ContentArtifact: pk=ba017c41-198c-4329-b8a9-df2c9d73e4f5>]>

In [7]: from django.db.models import Count

In [8]: ca_c = content_artifacts.annotate(remote_artifacts_c=Count("remoteartifact"))

In [9]: [ca.remote_artifacts_c for ca in ca_c]
Out[9]: [1, 1, 1]
```

It looks like it is not working for me. The count should return `[2,2,2]` because there are 2 remotes.

```
In [13]: [ca.remoteartifact_set.all().count() for ca in ca_c]
Out[13]: [2, 2, 2]

In [14]: ca_c = content_artifacts.annotate(remote_artifacts_c=Count("remoteartifact_set"))
FieldError                                Traceback (most recent call last)
Input In [14], in <cell line: 1>()
----> 1 ca_c = content_artifacts.annotate(remote_artifacts_c=Count("remoteartifact_set"))

File /usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/query.py:1091, in QuerySet.annotate(self, *args, **kwargs)
   1086 """
   1087 Return a query set in which the returned objects have been annotated
   1088 with extra data or aggregations.
   1089 """
   1090 self._not_support_combined_queries('annotate')
-> 1091 return self._annotate(args, kwargs, select=True)

File /usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/query.py:1130, in QuerySet._annotate(self, args, kwargs, select)
   1128         clone.query.add_filtered_relation(annotation, alias)
   1129     else:
-> 1130         clone.query.add_annotation(
   1131             annotation, alias, is_summary=False, select=select,
   1132         )
   1133 for alias, annotation in clone.query.annotations.items():
   1134     if alias in annotations and annotation.contains_aggregate:

File /usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/sql/query.py:1062, in Query.add_annotation(self, annotation, alias, is_summary, select)
   1060 """Add a single annotation expression to the Query."""
   1061 self.check_alias(alias)
-> 1062 annotation = annotation.resolve_expression(self, allow_joins=True, reuse=None,
   1063                                            summarize=is_summary)
   1064 if select:
   1065     self.append_annotation_mask([alias])

File /usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/aggregates.py:47, in Aggregate.resolve_expression(self, query, allow_joins, reuse, summarize, for_save)
     45 def resolve_expression(self, query=None, allow_joins=True, reuse=None, summarize=False, for_save=False):
     46     # Aggregates are not allowed in UPDATE queries, so ignore for_save
---> 47     c = super().resolve_expression(query, allow_joins, reuse, summarize)
     48     c.filter = c.filter and c.filter.resolve_expression(query, allow_joins, reuse, summarize)
     49     if not summarize:
     50         # Call Aggregate.get_source_expressions() to avoid
     51         # returning self.filter and including that in this loop.

File /usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/expressions.py:678, in Func.resolve_expression(self, query, allow_joins, reuse, summarize, for_save)
    676 c.is_summary = summarize
    677 for pos, arg in enumerate(c.source_expressions):
--> 678     c.source_expressions[pos] = arg.resolve_expression(query, allow_joins, reuse, summarize, for_save)
    679 return c

File /usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/expressions.py:578, in F.resolve_expression(self, query, allow_joins, reuse, summarize, for_save)
    576 def resolve_expression(self, query=None, allow_joins=True, reuse=None,
    577                        summarize=False, for_save=False):
--> 578     return query.resolve_ref(self.name, allow_joins, reuse, summarize)

File /usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/sql/query.py:1779, in Query.resolve_ref(self, name, allow_joins, reuse, summarize)
   1777         annotation = self.try_transform(annotation, transform)
   1778     return annotation
-> 1779 join_info = self.setup_joins(field_list, self.get_meta(), self.get_initial_alias(), can_reuse=reuse)
   1780 targets, final_alias, join_list = self.trim_joins(join_info.targets, join_info.joins, join_info.path)
   1781 if not allow_joins and len(join_list) > 1:

File /usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/sql/query.py:1648, in Query.setup_joins(self, names, opts, alias, can_reuse, allow_many, reuse_with_filtered_relation)
   1646 for pivot in range(len(names), 0, -1):
   1647     try:
-> 1648         path, final_field, targets, rest = self.names_to_path(
   1649             names[:pivot], opts, allow_many, fail_on_missing=True,
   1650         )
   1651     except FieldError as exc:
   1652         if pivot == 1:
   1653             # The first item cannot be a lookup, so it's safe
   1654             # to raise the field error here.

File /usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/sql/query.py:1562, in Query.names_to_path(self, names, opts, allow_many, fail_on_missing)
   1556     if pos == -1 or fail_on_missing:
   1557         available = sorted([
   1558             *get_field_names_from_opts(opts),
   1559             *self.annotation_select,
   1560             *self._filtered_relations,
   1561         ])
-> 1562         raise FieldError("Cannot resolve keyword '%s' into field. "
   1563                          "Choices are: %s" % (name, ", ".join(available)))
   1564     break
   1565 # Check if we need any joins for concrete inheritance cases (the
   1566 # field lives in parent, but we are currently in one of its
   1567 # children)

FieldError: Cannot resolve keyword 'remoteartifact_set' into field. Choices are: artifact, artifact_id, content, content_id, group_roles, published_artifact, pulp_created, pulp_id, pulp_labels, pulp_last_updated, relative_path, remoteartifact, user_roles
```

pulpcore/app/models/repository.py
```diff
@@ -499,6 +500,17 @@ def invalidate_cache(self):
             if base_paths:
                 Cache().delete(base_key=base_paths)
 
+    @hook(BEFORE_DELETE)
```
mdellweg, line 503: Can't we just add DB-level protection on the foreign fields?

mdellweg: So you are saying it's ok to delete remotes as long as every content_artifact that was associated to at least one remote will remain to be?

lubosmj: Yes. That was one of the requirements.

Thanks to that, the following workflow is still allowed:

```
(pulp) [vagrant@pulp3-source-fedora35 ~]$ pulp file remote create --name remo11 --url https://fixtures.pulpproject.org/file/PULP_MANIFEST --policy on_demand

(pulp) [vagrant@pulp3-source-fedora35 ~]$ pulp file remote create --name remo1 --url https://fixtures.pulpproject.org/file/PULP_MANIFEST --policy on_demand

(pulp) [vagrant@pulp3-source-fedora35 ~]$ pulp file repository create --name repo

(pulp) [vagrant@pulp3-source-fedora35 ~]$ pulp file repository sync --name repo --remote remo1
Started background task /pulp/api/v3/tasks/c35552cb-db05-418d-a65d-cddc3921f59c/
.Done.
(pulp) [vagrant@pulp3-source-fedora35 ~]$ pulp file repository sync --name repo --remote remo11
Started background task /pulp/api/v3/tasks/08c516ee-325f-4e8c-b293-6fdd7a3819a2/
.Done.

(pulp) [vagrant@pulp3-source-fedora35 ~]$ pulp file remote destroy --name remo11
Started background task /pulp/api/v3/tasks/6118b3d7-87e5-4a39-b413-4ad449732f27/
Done.
p(pulp) [vagrant@pulp3-source-fedora35 ~]$ pulp file publication create --repository repo
Started background task /pulp/api/v3/tasks/e4a80a89-1502-4ca2-a584-7e9a767ca592/
Done.
{
  "pulp_href": "/pulp/api/v3/publications/file/file/742dcf7e-9808-4f91-a23b-8cd5a12e4947/",
  "pulp_created": "2022-06-14T12:53:48.965896Z",
  "repository_version": "/pulp/api/v3/repositories/file/file/fe2d09ee-bd09-416b-b696-03bc4cb17db4/versions/1/",
  "repository": "/pulp/api/v3/repositories/file/file/fe2d09ee-bd09-416b-b696-03bc4cb17db4/",
  "distributions": [],
  "manifest": "PULP_MANIFEST"
}

(pulp) [vagrant@pulp3-source-fedora35 ~]$ pulp file distribution create --name aa --base-path aa --publication /pulp/api/v3/publications/file/file/742dcf7e-9808-4f91-a23b-8cd5a12e4947/
Started background task /pulp/api/v3/tasks/9138f366-1b01-4b54-944b-819e2bd678f6/
.Done.
{
  "pulp_href": "/pulp/api/v3/distributions/file/file/47a73e26-03a9-422d-ac4a-fce7c35124ea/",
  "pulp_created": "2022-06-14T12:54:03.177998Z",
  "base_path": "aa",
  "base_url": "https://pulp3-source-fedora35.localhost.example.com/pulp/content/aa/",
  "content_guard": null,
  "pulp_labels": {},
  "name": "aa",
  "repository": null,
  "publication": "/pulp/api/v3/publications/file/file/742dcf7e-9808-4f91-a23b-8cd5a12e4947/"
}

systemctl restart pulpcore-api pulpcore-content pulpcore-worker@1 pulpcore-worker@2 pulpcore
(pulp) [vagrant@pulp3-source-fedora35 ~]$ http https://pulp3-source-fedora35.localhost.example.com/pulp/content/aa/
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 467
Content-Type: text/html
Date: Tue, 14 Jun 2022 12:55:25 GMT
Server: nginx/1.20.2
X-PULP-CACHE: MISS

        <!DOCTYPE html>
        <html>
            <body>
                <ul>
                
                    <li><a href="1.iso">1.iso</a></li>
                
                    <li><a href="2.iso">2.iso</a></li>
                
                    <li><a href="3.iso">3.iso</a></li>
                
                    <li><a href="PULP_MANIFEST">PULP_MANIFEST</a></li>
                
                </ul>
            </body>
        </html>
        


(pulp) [vagrant@pulp3-source-fedora35 ~]$ http https://pulp3-source-fedora35.localhost.example.com/pulp/content/aa/1.iso
HTTP/1.1 200 OK
Accept-Ranges: bytes
Cache-Control: private
Connection: keep-alive
Content-Type: application/octet-stream
Date: Tue, 14 Jun 2022 12:55:31 GMT
Etag: "61b35d63-400"
Last-Modified: Fri, 10 Dec 2021 14:00:03 GMT
Server: nginx/1.20.2
Set-Cookie: dd6ec0a439e94ec1afa8091bd8cb5f4f=8e171c691bea3ee538a77ab911b8a911; path=/; HttpOnly; Secure; SameSite=None
Transfer-Encoding: chunked



+-----------------------------------------+
| NOTE: binary data not shown in terminal |
+-----------------------------------------+

(pulp) [vagrant@pulp3-source-fedora35 ~]$ pulp file remote destroy --name remo1
Started background task /pulp/api/v3/tasks/1052f408-b74f-4390-a3fa-4b8a639c6504/
Error: Task /pulp/api/v3/tasks/1052f408-b74f-4390-a3fa-4b8a639c6504/ failed: 'Attempt to delete the last remaining remote artifact.'

```

When we remove one of the synced remotes, we are still backed by another remote (I hope this is expected behaviour).

ipanova:
@lubosmj thank you for the PR. This looks like a significant change - how about we start a discussion thread on the discourse and see what other folks and users think about that?

lubosmj:
I am attaching a link to the corresponding discourse thread: https://discourse.pulpproject.org/t/when-to-allow-removal-of-remotes-used-for-on-demand-syncing/489/1.

lubosmj:
Eventually, I decided to implement option 1d) instead of 1c) from https://hackmd.io/_3gsUVdyQwy50Nc5pMXN-g. Please, take a look at the current changes.

lubosmj:
It seems like the implemented approach needs to be discussed more in detail. Switching to a draft.

lubosmj:
The CI is failing on the CLI test script:
```
FAILURE [pulp file remote create --name cli_test_file_remote --url http://pulp-fixtures:8080/file/PULP_MANIFEST]
=== STDOUT ===
=== STDERR ===
Error: {"name":["This field must be unique."]}
```

This change should be addressed in the CLI itself. I am undrafting this PR now to get some feedback on the implementation.

lubosmj:
Right now, we check the presence of active remote artifacts even if the content was already downloaded on demand.

Suppose a user already requested/downloaded the content that was synced with the on_demand policy. Now, the remote is no longer needed (hopefully) and can be removed without any restrictions. However, we still require the remote artifacts to be removed from the repository version first. Do you think I should update the logic to address this scenario?

mdellweg:
I'd guess the header must be something like `x-pulp-force`, because it's not a standard one.

CHANGES/1975.feature
```diff
@@ -0,0 +1,2 @@
+Added a check to prevent removal of remotes that are in use by repository versions. To bypass the
+check, pass the "{'force': 'True'}" body data in the DELETE request.
```
ipanova, line 2: We've discussed some of this approach with @pulp/core and came to the conclusion that specifying extra info in the request headers, also it being a non standard header is not the path we'd want to take.

We've being having back and forth conversations on this issue, sorry for that, however it is a non trivial problem to solve.
General consensus seems to be, given the complexity and absence of an explicit need from the community, to not implement any kind of `force` option for now.
The suggestion is to follow this https://discourse.pulpproject.org/t/should-we-keep-using-fk-with-on-delete-set-null/331/3 where we'd change to `on_delete=PROTECT` and give a friendly message to the user with a list of repository versions/repositories that use it. Then user would need himself to inspect and decide whether it is ok to DELETE those, before he can freely DELETE the remote.

Watch out for the RBAC aspect: user should not see hrefs for the resources he has no READ access to. In case it is impossible for that reason to share list of resources then simply returning a generic message that `This remote is in use` should suffice.

Out of scope of this PR would be to audit other models https://discourse.pulpproject.org/t/should-we-keep-using-fk-with-on-delete-set-null/331 which might need to have same change.

lubosmj: This suggestion effectively means I would be required to rewrite multiple test cases. Currently, we are removing remotes without any restrictions. Then, I would need to remove all repositories, and orphaned content first. Maybe, adding more advanced logic to `gen_object_with_cleanup` would suffice, @mdellweg?

mdellweg: This is changing the behaviour of Pulp. It's expected that some tests (aka constraints that Pulp behaves in a certain way) need to be adjusted.
In general about remotes here, i think we have two options:
* Realize that remotes are a dependency of content we agree to keep around after tests. The same would then apply to remotes.
* Start to clean up orphans with a specific list of hrefs at the end of a test before removing the remote.

bmbouter: @mdellweg Good analysis. I'd be ok to leave remotes around as a dependency of content. Cleaning up orphans I think would nullify our push for parallel testing as the orphan cleanups are serialized.

mdellweg: Well we don't clean up remotes with orphan cleanup. But calling orphan cleanup with exactly the hrefs of the content in that test would still allow for parallel testing.
In the end this very conflict is reflecting the original issue again.

ipanova: > This is changing the behaviour of Pulp. It's expected that some tests (aka constraints that Pulp behaves in a certain way) need to be adjusted. In general about remotes here, i think we have two options:
> 
>     * Realize that remotes are a dependency of content we agree to keep around after tests. The same would then apply to remotes.

To be precise remotes are a dependency of content which is part of a repo. Once a content if orphan I don't think we should prevent remote removal.
> 
>     * Start to clean up orphans with a specific list of hrefs at the end of a test before removing the remote.

lubosmj: Regarding the "PROTECTED" constraint, I am not adding it due to the workflow described in: https://github.com/pulp/pulpcore/pull/2836#discussion_r896754332. @ipanova, what do you think?

lubosmj: So, as a result, we will have "shallow" protection in our viewset and administrators will be still allowed to delete remotes from the admin shell.

ipanova: sorry for not being clear what protect i was referring to - on the repo model there is an fk to remote which is right no on_delete=set_null

lubosmj: I am quite puzzled how a remote assigned to a repository can be relevant to the problem described in the issue. The remote can be now removed whenever a users wants, when leaving the constraint unchanged. However, after adding the constraint, the user will not be able to remove the remote even if it has not been used yet (i.e., for syncing).

bmbouter: Not in response to @lubosmj  but generally, I think we should block remote deletion even if orphans exist because orphans could become non-orphans at any moment.

lubosmj: Hmm. And what would be an error message in your case, @bmbouter? For repositories, it makes sense to list all affected repository versions in order to let the user that she needs to first take a look at those repositories. In your case, remotes will effectively be considered the same types of residues as orphaned content, which makes them non-removable unless you run the orphan cleanup. And this is the case that was implemented a month ago.

ipanova: @lubosmj you are reading this right - it will be a new behavior for the user but this way repositories that use -share same remote will not lose it. Imagine you have repoA and repoB that user remote Foo. You remove remote Foo and then try to sync repoA and realize that it turns out that this particular repo uses it. @bmbouter can you confirm that we have the same thinking?

@bmbouter I don't think we should block deletion of remotes in case of orphans. Even they become non orphans they would be become so through a sync task which has stages and at RemoteArtifactSaver it would create a new one, would not it?(need to confirm this)
For the orphans that become non-orphan trough an upload task RemoteArtifacts do not matter anyway.

bmbouter: I tried to give a simple answer, but I need to look at this more. What I need to know is all the objects that have ForeignKeys to a Remote so we can look at each case and decide if having their existance block `Remote` removal is appropriate for users or not. My apologies I don't readily have that analysis right now.

pulpcore/app/viewsets/repository.py
```diff
@@ -272,6 +278,78 @@ class RemoteViewSet(
...
+                    version_url = urljoin(
+                        get_url(version.repository), f"versions/{version.number}/"
+                    )
+                    versions_urls.append(version_url)
+
+            return Response(
+                _(
+                    "Attempt to delete the remote which is still used and referenced from the "
+                    "following repository versions: {}. Try again after removing the listed "
+                    "repository versions and associated remote (content) artifacts."
```
bmbouter, line 306:
```suggestion
                    "Attempt to delete the remote which is still used and referenced from the "
                    "following repository versions below. Try again after removing the listed "
                    "repository versions and associated remote (content) artifacts.\n\n{}"
```

A sentence on top and bottom of a potentially large list would be hard to read (to me).

pulpcore/app/viewsets/repository.py
```diff
@@ -272,6 +278,78 @@ class RemoteViewSet(
...
+                ).format(versions_urls),
+                status.HTTP_406_NOT_ACCEPTABLE,
+            )
+        else:
+            return Response(
+                _("The remote is still in use and cannot be deleted."),
+                status.HTTP_406_NOT_ACCEPTABLE,
+            )
+
+    def get_repositories_with_delete_permission(self, versions):
```
bmbouter, line 316: @mdellweg isn't there shared machinery that provides an answer like this? I think when we remade the guradian APIs we made something like this.

pulpcore/app/viewsets/repository.py
```diff
@@ -272,6 +278,78 @@ class RemoteViewSet(
...
+
+def _get_versions_using_remote(pk):
+    """Return versions that contain remote artifacts referencing the specified remote."""
+    remote_artifacts = RemoteArtifact.objects.filter(remote=pk)
+    content_artifacts = ContentArtifact.objects.filter(
+        remoteartifact__in=remote_artifacts
+    ).prefetch_related("remoteartifact_set")
+
+    content_artifacts_with_single_remote = (
+        ca.content.pk for ca in content_artifacts if ca.remoteartifact_set.count() <= 1
```
bmbouter, line 346: Couldn't this be filtered in the db instead of compared in python?

bmbouter:
Am I right in understanding the Remote is prevented in being deleted if:
* It's referenced by a Repository (as the default remote to be used for syncing)
* It's FK referenced by a RemoteArtifact.

I see one `PROTECT` change for the Repository, but I don't see one [here](https://github.com/pulp/pulpcore/blob/6287ef36d2fc20ca316e1e889296fc55072ff477/pulpcore/app/models/content.py#L685) for the `RemoteArtifact`. I expected there would be both, is that what you all expected? The preventing deletion should cause it never to occur, but at the DB level I still expected it to be enforced.

pulpcore/app/viewsets/repository.py
lubosmj: This method was created with a custom access policy in mind. We are looking for two kinds of permission to ensure that a user who wants to delete a remote is allowed both to view and delete a repository if that will be needed.

pulpcore/app/viewsets/repository.py
lubosmj: https://github.com/pulp/pulpcore/pull/2836#discussion_r896609425

Maybe, with customized SQL queries, possibly. But, not with aggregation functions like "COUNT".

bmbouter:
> @bmbouter: [#2836 (comment)](https://github.com/pulp/pulpcore/pull/2836#discussion_r896754332)
> 
> If there will be a redundancy of remoteartifacts' resources and one of them will be deleted, we will be still able to sync content, or not?

Yes in the content app, the `RemoteArtifact` objects are iterated [here](https://github.com/pulp/pulpcore/blob/main/pulpcore/content/handler.py#L690-L702). Let me know if I've misunderstood your question.

pulpcore/app/viewsets/repository.py
bmbouter: I imagined maybe with F() or one of the other aggregating fields. I'm ok with this as-is though, just a thought I had to improve performance. https://docs.djangoproject.com/en/3.2/ref/models/expressions/#using-f-in-filters

ipanova:
> Am I right in understanding the Remote is prevented in being deleted if:
> 
>     * It's referenced by a Repository (as the default remote to be used for syncing)
> 
>     * It's FK referenced by a RemoteArtifact.
> 
> 
> I see one `PROTECT` change for the Repository, but I don't see one [here](https://github.com/pulp/pulpcore/blob/6287ef36d2fc20ca316e1e889296fc55072ff477/pulpcore/app/models/content.py#L685) for the `RemoteArtifact`. I expected there would be both, is that what you all expected? The preventing deletion should cause it never to occur, but at the DB level I still expected it to be enforced.

We're on the same page except for the `PROTECT` for the `RemoteArtifact`
If you read 2nd paragraph from this comment do you still think we should have it? https://github.com/pulp/pulpcore/pull/2836#discussion_r948144348 I think this will greatly complicate user's life because orphan endpoint is exposed only to the admin. So if you happen to create a repo, a remote, then sync content, afterwards delete repo ( this way all your content will become orphan) you won't be able to delete the remote you've created until admin would run orphan clean up.

bmbouter:
That does sound complicated for the user, which is not great. Can we explore making it more usable? What is the reasoning (I forget) that orphan cleanup only an admin-only operation? What if instead that operation default to any authenticated user?

This work came out of another (I think even bigger) user problem, that deleting `Remote` objects that are referenced by `RemoteArtifact` objects breaks those `RemoteArtifact` objects. Addressing that is the primary concern of the work, which is why I believe it is a requirement.

ipanova:
> That does sound complicated for the user, which is not great. Can we explore making it more usable? What is the reasoning (I forget) that orphan cleanup only an admin-only operation? What if instead that operation default to any authenticated user?

The idea was that the content access scopes off the repo permissions and once a content becomes an orphan then no one should see it unless it's a privileged user like admin. The logic comes from here where if only admin can see orphan content than only admin should be able to clean it up.

Looking at the current orphan cleanup endpoints where only DELETE and POST calls are allowed they do not really expose the orphaned content itself. I think, I would be ok to open up these endpoint to any authed user, but that would be a separate change and PR.

> 
> This work came out of another (I think even bigger) user problem, that deleting `Remote` objects that are referenced by `RemoteArtifact` objects breaks those `RemoteArtifact` objects. Addressing that is the primary concern of the work, which is why I believe it is a requirement.

ipanova:
> @ipanova thank you for explaining that.
> 
> +1 to making the adjustment of calling orphan cleanup in another PR. Do you think we should have that done first and merge that and then this?

As long as these 2 PRs go into one release I don't care in which order they get merged, I guess.
> 
> Also are we then in agreement that `RemoteArtifact` should use `PROTECT` also?

Yes

bmbouter:
@lubosmj it seems @ipanova and I have reached agreement. Can you make the other PR also? Please reach out if you have any questions about what to do.

gerrod3:
From [my comment](https://github.com/pulp/pulpcore/pull/3207#issuecomment-1248536367) I think we are going to need to add additional logic to the remote deletion task.

1. Gather all RemoteArtifacts for Remote and their associated ContentArtifacts
2. If each ContentArtifact has an Artifact (the content is no longer on-demand), delete all RemoteArtifacts and delete Remote, finished.
3. Sort each on-demand content-artifact into buckets:
     - Orphaned and only this(one) RemoteArtifact
     - Orphaned and multiple RemoteArtifacts
     - Non-orphaned
4. If only items in bucket one: delete each content and then delete the Remote, finished.
5. If only items in bucket one & two: delete each remote artifact, then delete any content that now has zero remote artifacts and then delete the Remote, finished.
6. If any items in bucket three, fail the task. I think if we are in this scenario the logic becomes even more complicated and I don't know if we can safely delete the remote.

I tried to summarized the discussion I had with @dkliban and @ggainey about the new logic needed, but correct me if I forgot or got a part wrong.

PR #2843, opened 2022-06-14
Add more details to the description of ACS

lubosmj:
[noissue]

### DIFF
--- a/docs/workflows/alternate-content-sources.rst
+++ b/docs/workflows/alternate-content-sources.rst
@@ -9,15 +9,25 @@ Alternate Content Sources
 Overview
 --------
 
-Pulp supports the concept of Alternate Content Sources which sync content using a Remote.  Each
-content source is a potential alternate provider of files that are associated with content units in
-Pulp.  The next time Pulp needs to download a file associated with a content unit, it searches ACSes
-for alternate sources.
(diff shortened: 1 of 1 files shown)


docs/workflows/alternate-content-sources.rst
```diff
@@ -9,15 +9,22 @@ Alternate Content Sources
...
-Pulp.  The next time Pulp needs to download a file associated with a content unit, it searches ACSes
-for alternate sources.
+Pulp supports the concept of Alternate Content Sources (ACS) which sync content using a Remote.
+Each content source is a potential alternate provider of files that are associated with content
+units in Pulp.
 
-Create an Alternate Content Source
-----------------------------------
+ACS are useful when dealing with an unreliable or slow internet connection to remote repositories.
+Setting the ACS tells Pulp to first check for alternative sources of content in an attempt to pull
```
bmbouter, line 17: This is great, and I would lgtm as is. If you want to write more, one or two examples would be nice. Like the local disk where the ACS uses `file://path/to/repo/` and it still lets the real and newer source of that content define the metadata. This is a quick way to "sync a repo" because it's really coming from local disks. Or if there is a mirror of the CDN that you know is closer, you can do the same thing there.

bmbouter (approved):
I invited more writing, but as is is fine too.

pavelpicka (approved):
Looks great.

Issue #2703, opened 2022-05-09
Group Role List - ordering by role name gives surprising order


himdel:
**Version**
```
galaxy_ng_version: "4.6.0dev"
pulp_ansible_version: "0.13.0"
pulp_container_version: "2.10.3"
pulp_core_version: "3.18.3"
```

**Describe the bug**
Ordering by role name in group role list gives surprising order.

**To Reproduce**
1. Have a group with multiple (rbac) roles assigned.
2. List those roles using `/pulp/api/v3/groups/45/roles/?ordering=role`

(I also tried `ordering=name` but that one gives the same order with or without `-` and it gives the same order for `ordering=whatever`, whereas `ordering=role` is sorting by something, just maybe not quite the right thing)

Actual results:

```
GET /pulp/api/v3/groups/45/roles/?ordering=role
HTTP 200 OK
Allow: GET, POST, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "count": 8,
    "next": null,
    "previous": null,
    "results": [
        {
            "pulp_href": "/pulp/api/v3/groups/45/roles/b1572318-abce-486f-9668-a3f04a386b0a/",
            "pulp_created": "2022-05-09T21:10:04.578411Z",
            "role": "galaxy.admin_30227955-847a-40a1-891f-bd64ed17f116",
            "content_object": null
        },
        {
            "pulp_href": "/pulp/api/v3/groups/45/roles/cdd0c477-09c5-4fe1-b91f-141c19f369e6/",
            "pulp_created": "2022-05-09T16:36:09.374777Z",
            "role": "galaxy.admin_1f0f3338-9614-4e88-98f3-97f7e9af339d",
            "content_object": null
        },
        {
            "pulp_href": "/pulp/api/v3/groups/45/roles/0a6fcfbc-f8de-4016-b3f4-6edbd2385374/",
            "pulp_created": "2022-05-09T16:32:40.989792Z",
            "role": "galaxy.publisher",
            "content_object": null
        },
        {
            "pulp_href": "/pulp/api/v3/groups/45/roles/5f326769-679f-4ffe-b320-71d30b1353ba/",
            "pulp_created": "2022-05-09T21:10:04.580363Z",
            "role": "galaxy.admin_2c341dd0-7616-4429-a093-a081cddeb242",
            "content_object": null
        },
        {
[...]

PR #2849, opened 2022-06-15
Ordering cleanup

mdellweg:


bmbouter (approved):
This looks good to me.

dkliban (approved):

himdel:
@mdellweg Does this undo the current resolution of https://github.com/pulp/pulpcore/issues/2703 ?

Looking at the changes, it seems `?sort` will no longer be available and we need to start using `?ordering` on ... some endpoints?
Do you have the list of affected API endpoints please?

mdellweg:
@himdel Yes, that is correct. But also to my understanding, we never released a version of pulp using the "sort" so this is technically not a removal. All endpoints use the `?ordering` filter now. The endpoints using a special version of that filter are (to my knowledge) the Group/User-Role endpoints `...v3/groups/2/roles/` and `...v3/users/4/roles/`.

himdel:
Yeah, I agree it's fine, just making sure I got it right, updating galaxy accordingly :)

Thanks :)

PR #2876, opened 2022-06-20
Add RBAC to Upload

### LINKED ISSUE #2362: Uploads need RBAC support
When uploading a file in chunks, for whatever reason, no one should be able to mess with my upload process or use my artifact with out proper permissions.

mdellweg:
Also provide a global access condition to be used in one-shot content
creation calls to check for permissions on the upload object.

fixes #2362


pulpcore/app/global_access_conditions.py
```diff
@@ -281,11 +277,9 @@ def has_repo_attr_model_or_obj_perms(request, view, action, permission):
         True if the user has the Permission on the ``repository`` attribute named by the
         ``permission`` at the model or object level. False otherwise.
     """
-    if has_model_perms(request, view, action, permission):
```
bmbouter, line 284: I actually found this version much more readable. This is a subjective thing though so you decide.

bmbouter (approved):
One little subjective nitpick from me, but otherwise LGTM.

pulpcore/app/global_access_conditions.py
mdellweg: Hmm I don't like the way black formats it that much, but the "or" is nicely echoing the function name here.
I can be convinced though. I would not expect any difference in execution speed here.

bmbouter: Yeah it just seems much less readable. I'm ok with it either way though.

mdellweg: As we are not merging this today, let's try to get more tiebreakers. And if there is no sufficient bias by tomorrow a will throw a die.

gerrod3 (approved):

pulpcore/app/global_access_conditions.py
gerrod3: Flipping a coin would be easier. Either way is fine in my opinion.

PR #2880, opened 2022-06-21
Adds signing service pytest fixtures

bmbouter:
The new fixtures can be used to auto-setup-and-teardown an
`AsciiArmoredDetachedSigningService`. Pre-generated gpg public and
private keys are used temporarily configure the gpghome dir for use
during the test. Also "local" signing facilities are also available for
tests to produce signatures locally too.

[noissue]

### DIFF
--- a/pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
+++ b/pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
@@ -1,0 +1,166 @@
+import asyncio
+import json
+import os
+import stat
+import subprocess
+import uuid
+
+import aiohttp
+import gnupg
+import pytest
+
+
+signing_script_string = r"""#!/usr/bin/env bash
+
+FILE_PATH=$1
+SIGNATURE_PATH="$1.asc"
+
+GPG_KEY_ID="Pulp QE"
+
+# Create a detached signature
+gpg --quiet --batch --homedir HOMEDIRHERE --detach-sign --local-user "${GPG_KEY_ID}" \
+   --armor --output ${SIGNATURE_PATH} ${FILE_PATH}
+
+# Check the exit status
+STATUS=$?
+if [[ ${STATUS} -eq 0 ]]; then
+   echo {\"file\": \"${FILE_PATH}\", \"signature\": \"${SIGNATURE_PATH}\"}
+else
+   exit ${STATUS}
+fi
+"""
+
+
+@pytest.fixture(scope="session")
+def signing_script_path(signing_script_temp_dir):
+    signing_script_filename = signing_script_temp_dir / "sign-metadata.sh"
+    with open(signing_script_filename, "w", 0o770) as sign_metadata_file:
+        sign_metadata_file.write(
+            signing_script_string.replace("HOMEDIRHERE", str(signing_script_temp_dir))
+        )
+
+    st = os.stat(signing_script_filename)
+    os.chmod(signing_script_filename, st.st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
+
+    return signing_script_filename
+
+
+@pytest.fixture(scope="session")
+def signing_script_temp_dir(tmpdir_factory):
+    yield tmpdir_factory.mktemp(str(uuid.uuid4()))
+
+
+@pytest.fixture(scope="session")
+def signing_gpg_homedir_path(tmpdir_factory):
+    return tmpdir_factory.mktemp(str(uuid.uuid4()))
+
+
+@pytest.fixture
+def sign_with_ascii_armored_detached_signing_service(
+    signing_script_path, gpg_homedir_with_trusted_private_key
+):
+    """
+    Runs the test signing script manually, locally, and returns the signature file produced.
+    """
+
+    def _sign_with_ascii_armored_detached_signing_service(filename):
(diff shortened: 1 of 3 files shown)


pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
```diff
@@ -0,0 +1,153 @@
...
+
+    assert completed_process.returncode == 0
+
+    # We have to generate these here because this session scoped fixture is incompatible with the
+    # function scoped bindings objects
+    pulpcore_client = ApiClient(bindings_cfg)
+    signing_service_api_client = SigningServicesApi(pulpcore_client)
+    signing_service = signing_service_api_client.list(name=service_name).results[0]
+
+    yield signing_service
```
mdellweg, line 141: You can move these lines in a test scoped fixture and make this session scoped one invisible (append a "_") and return only the name.

```python
@fixture(scope="session")
def _ascii_armored_signing_service_name():
    ...
    yield name

@fixture
def ascii_armored_signing_service(_ascii_armored_signing_service_name, signing_service_api_client):
    return signing_service_api_client.list(name=_ascii_armored_signing_service_name).results[0]
```

pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
```diff
@@ -0,0 +1,153 @@
...
+    fingerprint = gpg.list_keys()[0]["fingerprint"]
+    keyid = gpg.list_keys()[0]["keyid"]
+
+    gpg.trust_keys(fingerprint, "TRUST_ULTIMATE")
+
+    return gpg, fingerprint, keyid
+
+
+@pytest.fixture(scope="session")
+def gpg_signing_service(
```
mdellweg, line 110: Pleas use a name that identifies this as an ascii armored signing service.

pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
```diff
@@ -0,0 +1,153 @@
...
+        private_key_data = f.read()
+
+    gpg.import_keys(private_key_data)
+
+    fingerprint = gpg.list_keys()[0]["fingerprint"]
+    keyid = gpg.list_keys()[0]["keyid"]
+
+    gpg.trust_keys(fingerprint, "TRUST_ULTIMATE")
+
+    return gpg, fingerprint, keyid
```
mdellweg, line 106: I think it would be convenient to have fixtures for each of those.

pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
```diff
@@ -0,0 +1,153 @@
...
+    yield tmpdir_factory.mktemp(str(uuid.uuid4()))
+
+
+@pytest.fixture(scope="session")
+def signing_gpg_private_key_path():
+    return os.path.realpath(f"{__file__}/../../assets/GPG-PRIVATE-KEY-pulp-qe")
+
+
+@pytest.fixture
+def call_signing_script(signing_script_path, gpg_metadata):
```
mdellweg, line 62: Not sure i see the need for this. Is it a convenient way to generate a signature for upload when using the special domain tool (rpmsign, scopeo, ...) would be more complicated?

pulpcore/tests/assets/GPG-KEY-pulp-qe
```diff
@@ -0,0 +1,30 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
mdellweg, line 1: We used to download this from pulp-smash with the added benefit of not needing to add any security exceptions.
Can you explain why we should do it differently now?

mdellweg (changes requested):
Really required is the naming question...

pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
bmbouter: The intention here is as a replacement for [this code carried in pulp_ansible](https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/tests/functional/utils.py#L348-L372). You could ask the gpg tools to do it directly I guess. That would be mostly just different I think, but is it actually better?

pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
bmbouter: I agree, but later. This is on the 3.20 critical path so I'm trying to keep it basic.

pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
bmbouter: :+1: this definitely should be named differently. I've renamed to `ascii_armored_detached_signing_service`.

pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
bmbouter: I'm unclear on how this is an improvement. Can you explain this idea here a bit more please?

pulpcore/tests/assets/GPG-KEY-pulp-qe
bmbouter: After some discussion at the pulpcore meeting, it will be easier if we keep this as a wget for now and improve it later. I'm switching the implementation now.

mdellweg (approved):
LGTM

PR #2891, opened 2022-06-21
Updates CI config with latest plugin_template

dkliban:
Runs nightly tests at release time. Removes the aiohttp_fixtures_origin parameter from
template_config.yaml. This value is just hard coded to 127.0.0.1.

[noissue]


fao89 (approved):
you can do it by triggering here:
https://github.com/pulp/pulpcore/actions/workflows/update_ci.yml

bmbouter (approved):
Great!

Issue #2864, opened 2022-06-16
As a user I want to filter repositories by remote.


newswangerd:
**Is your feature request related to a problem? Please describe.**
The galaxy ng UI needs to be able to fetch a list of repositories (or ideally distributions) that have remotes attached to them. The UI uses this to display repositories that are configured to be synced.

**Describe the solution you'd like**
The repository API has a remote field attached to it. I would like to be able to get a list of repositories that:
- Have a remote attached
- Have no remote attached
- Have a specific remote attached

Something like this would be nice on the distributions API as well. It would be great to see which distributions point to repos that have remotes, not remotes, or a specific remote.

**Describe alternatives you've considered**
We could use labels to track this stuff, but that would require enforcing specific sets of labels when creating repos. It would be better to just have a filter to handle this.

**Additional context**
Required for: https://issues.redhat.com/browse/AAH-756

ipanova:
For the sake of trackability, @gerrod3  can you file a separate issue for distributions filter?

PR #2904, opened 2022-06-23
Add Remote filter for Repositories

gerrod3:
fixes: #2864

### DIFF
--- a/CHANGES/2864.feature
+++ b/CHANGES/2864.feature
@@ -1,0 +1,3 @@
+``HyperlinkRelatedFilter`` can now be filtered by object types and NULL values.
+
+Repositories can now be filtered by Remotes.
--- a/pulpcore/app/viewsets/custom_filters.py
+++ b/pulpcore/app/viewsets/custom_filters.py
@@ -72,11 +72,19 @@ class CreatedResourcesFilter(Filter):
 
 class HyperlinkRelatedFilter(Filter):
     """
-    Enables a user to filter by a foreign key using that FK's href
+    Enables a user to filter by a foreign key using that FK's href.
+
+    Foreign key filter can be specified to an object type by specifying the base URI of that type.
+    e.g. Filter by file remotes: ?remote=/pulp/api/v3/remotes/file/file/
+
+    Can also filter for foreign key to be unset by setting ``allow_null`` to True. Query parameter
+    will then accept "null" or "" for filtering.
+    e.g. Filter for no remote: ?remote="null"
     """
 
     def __init__(self, *args, **kwargs):
         kwargs.setdefault("help_text", _("Foreign Key referenced by HREF"))
+        self.allow_null = kwargs.pop("allow_null", False)
         super().__init__(*args, **kwargs)
 
     def filter(self, qs, value):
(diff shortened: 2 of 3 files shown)


pulpcore/app/viewsets/custom_filters.py
```diff
@@ -72,11 +72,19 @@ def filter(self, qs, value):
...
+        e.g. Filter by file remotes: ?remote=/pulp/api/v3/remotes/file/file/
+
+    Can also filter for foreign key == NULL by setting ``allow_null`` to True. Query parameter
+    will then accept "null" or "" for filtering.
+        e.g. Filter for remote == NULL: ?remote="null"
     """
 
     def __init__(self, *args, **kwargs):
         kwargs.setdefault("help_text", _("Foreign Key referenced by HREF"))
+        self.allow_null = kwargs.pop("allow_null", False)
```
gerrod3, line 87: Should this be a new settable value or just be on by default?

pulpcore/app/viewsets/custom_filters.py
```diff
@@ -93,23 +101,33 @@ def filter(self, qs, value):
...
-            raise serializers.ValidationError(detail=_("UUID invalid: {u}").format(u=pk))
+        elif self.allow_null and (value == "null" or value == ""):
+            key = f"{self.field_name}__isnull"
+            lookup = True
+        else:
+            try:
+                match = resolve(urlparse(value).path)
+            except Resolver404:
+                raise serializers.ValidationError(detail=_("URI not valid: {u}").format(u=value))
+            # Need to add a check here to make sure that the URI even makes sense to filter by
```
gerrod3, line 116: Anyone know how to do this cleanly? I thinking of checking against the queryset's model.meta.fields, but I wonder if there is a simpler way.

mdellweg: You mean, whether the url provided matches the type of object expected? Are there informations about this in the `match`?

pulpcore/app/viewsets/custom_filters.py
```diff
@@ -72,11 +72,19 @@ def filter(self, qs, value):
...
     """
-    Enables a user to filter by a foreign key using that FK's href
+    Enables a user to filter by a foreign key using that FK's href.
+
+    Foreign key filter can be specified to an object type by specifying the base URI of that type.
+        e.g. Filter by file remotes: ?remote=/pulp/api/v3/remotes/file/file/
+
+    Can also filter for foreign key == NULL by setting ``allow_null`` to True. Query parameter
+    will then accept "null" or "" for filtering.
+        e.g. Filter for remote == NULL: ?remote="null"
```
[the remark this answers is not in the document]
gerrod3: So have the user input `?remote=unset`?

pulpcore/app/viewsets/custom_filters.py
```diff
@@ -93,23 +101,33 @@ def filter(self, qs, value):
...
+                try:
+                    UUID(pk, version=4)
+                except ValueError:
+                    raise serializers.ValidationError(detail=_("UUID invalid: {u}").format(u=pk))
+
+                key = "{}__pk".format(self.field_name)
+                lookup = pk
+            else:
+                key = f"{self.field_name}__in"
+                lookup = match.func.cls.queryset
```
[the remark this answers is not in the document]
gerrod3: No, this is much more powerful. Since this filter can be on any foreign key field it can be now be used to scope by object subtypes. So on the generic repository endpoint you can now list repositories with file remotes, or ansible remotes, or python remotes, etc...

pulpcore/app/viewsets/custom_filters.py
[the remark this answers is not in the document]
mdellweg: Hah, no. I was just thinking about the text part...
```suggestion
    Can also filter for foreign key to be unset by setting ``allow_null`` to True. Query parameter
    will then accept "null" or "" for filtering.
        e.g. Filter for no remote: ?remote="null"
```

dralley:
@gerrod3 Could you fix the documentation build issue?

pulpcore/app/viewsets/custom_filters.py
gerrod3: You can get the viewset object from the matched url, which contains the queryset, but I still have the problem of knowing which querysets this field should accept.

CHANGES/2864.feature
```diff
@@ -0,0 +1,3 @@
+``HyperlinkRelatedFilter`` can now be filtered by object types and NULL values.
+
+Repositories can now by filtered by Remotes.
```
mdellweg, line 3:
```suggestion
Repositories can now by filtered by Remotes.
```
```suggestion
Repositories can now be filtered by Remotes.
```

pulpcore/app/viewsets/custom_filters.py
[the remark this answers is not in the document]
mdellweg: I know it's a bit late, but i was just wondering whether we could query something like
```python
key = f"{self.field_name}__pulp_type"
lookup = match.func.cls.pulp_type
```
?

Just a guts feeling that this might perform faster.

mdellweg:
@gerrod3 any updates?

gerrod3:
> @gerrod3 any updates?

I just saw your last review (strange I didn't get a notification about it). I'm going to test that out real quick. As for my issue with preventing bad href's from being passed in I think I have a solution to that. And I need to add 1/2 tests for the new functionality. Also, I had a question in my review if this new feature should be a setting on the filter or just on by default, I would like some opinions on it.

pulpcore/app/viewsets/custom_filters.py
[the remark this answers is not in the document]
gerrod3: Didn't work :disappointed:

pulpcore/app/viewsets/custom_filters.py
gerrod3: Ok, I added a safety check to make sure you can't pass in "valid" URIs that the filter doesn't expect.

pulpcore/app/viewsets/custom_filters.py
mdellweg: I think, it's fine the way it is.

mdellweg:
So i'm going to wait for those tests now.

mdellweg (approved):

dralley (approved):
Looks OK, but a warning, these have always been fairly slow and I'm slightly concerned about making it more complex.

It would be worthwhile (not necessarily as a blocker, but eventually) to use tools like [`django-debug-toolbar`](https://github.com/jazzband/django-debug-toolbar) and [`django-silk`](https://github.com/jazzband/django-silk) to get a look at what some of these API calls are doing.  It has been years since I looked and it was quite nuts even back then.

When I did the prometheus demo I remember noticing that the tasks endpoint was especially bad.  Might be difficult to notice on a system with only a few tasks, but if there are thousands, it might be more noticeable.

PR #2913, opened 2022-06-30
Fix usage of gpghome fixture

mdellweg:
[noissue]


fao89 (approved):

mdellweg:
for patchback:
https://github.com/pulp/pulpcore/pull/2914

PR #2915, opened 2022-07-01
Rename the fixture to better describe its meaning

lubosmj:
[noissue]


mdellweg:
This is changing an existing interface. I'm very hesitant to approve it.

lubosmj:
We can wait for @bmbouter to decide. I think that this change makes sense because the fixture returns the gpg instance and metadata of a signature.

bmbouter:
The name change would be ok for me, but if you want to do this please open the following PRs first:

* A backport of this to `pulpcore:3.20`
* Port the name in `pulp_ansible:main` branch
* Port the name in the `pulp_ansible:0.14` branch

@mdellweg if he did all that what would you say?

mdellweg:
So it does not seem to be used directly yet (only indirectly). Then this is the time to do the switch.

mdellweg (approved):

ipanova (approved):

Issue #2933, opened 2022-07-05
File System  Exporter needs to use the correct relative path.


parthaa:
**Version**
Please provide the versions of the pulpcore and plugin packages in use, and how they are installed.  If you are using Pulp via Katello, please provide the Katello version.

pulpcore  3.16 and above.

**Describe the bug**
Note: This is not exactly a bug since this functionality is listed as "Tech Preview", but Katello and customers now need it work. More over this is a feature that's already there in the code so marking it as bug seems appropriate to me .

File system exporter is not exporting the publications correctly.

**To Reproduce**
- Sync a RH repo that has packages with the following structure (example repo here is `openstack-tools`)
```bash
$ ls content/dist/rhel/server/7/7.5/x86_64/openstack-tools/10/os/
Packages  repodata

$ ls content/dist/rhel/server/7/7.5/x86_64/openstack-tools/10/os/Packages/
b  listing  o  p

$ ls content/dist/rhel/server/7/7.5/x86_64/openstack-tools/10/os/Packages/b/
babel-2.3.4-1.el7ost.noarch.rpm  listing
```
- Create a publication. 
- Use the file system exporter to export this somewhere like `/var/lib/pulp/exports/foo` and give the publication href.
- Check the exported result 
```
$ ls -l /var/lib/pulp/exports/foo
.....
babel-2.3.4-1.el7ost.noarch.rpm
.....
repodata
....
```
- The Packages directory is not retained while the primary xml in the repodata points to `Packages/b/babel-2.3.4-1.el7ost.noarch.rpm

**Expected behavior**
The Packages directory is correctly maintained when exporting.

**Additional context**
The relevant code is at https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/export.py#L63 and https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/export.py#L82.

The actual paths are correct in  core_publishedartifact. So just need to make sure the relative_paths are correct in case of a publication.

```sql
# select relative_path from  core_publishedartifact;
.....
Packages/b/babel-2.3.4-1.el7ost.noarch.rpm
 ....
```

PR #2951, opened 2022-07-07
Use published relative paths for FS Exporter

parthaa:
FS Exporter uses the relative_paths of the content_artifacts instead of
the actual publication's relative path when exporting publications.
This commit use published relative paths if a publication is available

### DIFF
--- a/pulpcore/app/tasks/export.py
+++ b/pulpcore/app/tasks/export.py
@@ -37,20 +37,29 @@ from pulpcore.constants import FS_EXPORT_METHODS
 log = logging.getLogger(__name__)
 
 
-def _export_to_file_system(path, content_artifacts, method=FS_EXPORT_METHODS.WRITE):
+def _validate_fs_export(content_artifacts):
     """
-    Export a set of ContentArtifacts to the filesystem.
-
     Args:
-        path (str): A path to export the ContentArtifacts to
         content_artifacts (django.db.models.QuerySet): Set of ContentArtifacts to export
 
     Raises:
-        ValidationError: When path is not in the ALLOWED_EXPORT_PATHS setting
         RuntimeError: If Artifacts are not downloaded or when trying to link non-fs files
     """
     if content_artifacts.filter(artifact=None).exists():
         raise RuntimeError(_("Cannot export artifacts that haven't been downloaded."))
+
+
+def _export_to_file_system(path, relative_paths_to_artifacts, method=FS_EXPORT_METHODS.WRITE):
+    """
+    Export a set of artifacts to the filesystem.
+
+    Args:
+        path (str): A path to export the ContentArtifacts to
+        relative_paths_to_artifacts: A dict with {relative_path: artifact} mapping
+
+    Raises:
+        ValidationError: When path is not in the ALLOWED_EXPORT_PATHS setting
+    """
 
     if (
         settings.DEFAULT_FILE_STORAGE != "pulpcore.app.models.storage.FileSystem"
@@ -58,10 +67,8 @@ def _export_to_file_system(path, content_artifacts, method=FS_EXPORT_METHODS.WRI
     ):
         raise RuntimeError(_("Only write is supported for non-filesystem storage."))
 
-    for ca in content_artifacts.select_related("artifact").iterator():
-        artifact = ca.artifact
-        dest = os.path.join(path, ca.relative_path)
-
+    for relative_path, artifact in relative_paths_to_artifacts.items():
+        dest = os.path.join(path, relative_path)
         os.makedirs(os.path.split(dest)[0], exist_ok=True)
 
         if method == FS_EXPORT_METHODS.SYMLINK:
@@ -74,7 +81,6 @@ def _export_to_file_system(path, content_artifacts, method=FS_EXPORT_METHODS.WRI
             with open(dest, "wb") as f, artifact.file as af:
                 for chunk in af.chunks(1024 * 1024):
                     f.write(chunk)
-
         else:
             raise RuntimeError(_("Unsupported export method '{}'.").format(method))
 
@@ -112,7 +118,21 @@ def fs_publication_export(exporter_pk, publication_pk):
             content__in=publication.repository_version.content
         )
 
-    _export_to_file_system(exporter.path, content_artifacts)
+    _validate_fs_export(content_artifacts)
+
+    relative_path_to_artifacts = {}
+    if publication.pass_through:
+        relative_path_to_artifacts = {
+            ca.relative_path: ca.artifact
+            for ca in content_artifacts.select_related("artifact").iterator()
+        }
+
+    for pa in publication.published_artifact.select_related(
+        "content_artifact", "content_artifact__artifact"
+    ).iterator():
+        relative_path_to_artifacts[pa.relative_path] = pa.content_artifact.artifact
+
+    _export_to_file_system(exporter.path, relative_path_to_artifacts, exporter.method)
 
 
 def fs_repo_version_export(exporter_pk, repo_version_pk):
@@ -141,8 +161,14 @@ def fs_repo_version_export(exporter_pk, repo_version_pk):
     )
 
     content_artifacts = ContentArtifact.objects.filter(content__in=repo_version.content)
+    _validate_fs_export(content_artifacts)
 
-    _export_to_file_system(exporter.path, content_artifacts)
+    relative_path_to_artifacts = {
+        ca.relative_path: ca.artifact
+        for ca in content_artifacts.select_related("artifact").iterator()
+    }
+
+    _export_to_file_system(exporter.path, relative_path_to_artifacts, exporter.method)
 
 
 def _get_versions_to_export(the_exporter, the_export):
--- a/CHANGES/2933.bugfix
+++ b/CHANGES/2933.bugfix
@@ -1,0 +1,1 @@
+Use published relative paths for FS Exporter.


parthaa:
fixes #2933

ggainey:
> fixes #2933

This needs to be in the commit-message - do a "git commit --amend" to add it to your commit, then "git push --force" to replace the existing one, should clear up the lint error!

pulpcore/app/tasks/export.py
```diff
@@ -58,10 +60,22 @@ def _export_to_file_system(path, content_artifacts, method=FS_EXPORT_METHODS.WRI
...
+            for content_artifact_pk, relative_path in publication.published_artifact.values_list(
+                "content_artifact__pk", "relative_path"
+            )
+        }
+
     for ca in content_artifacts.select_related("artifact").iterator():
         artifact = ca.artifact
-        dest = os.path.join(path, ca.relative_path)
+        relative_path = ca.relative_path
+        if publication and ca.published_artifact:
```
mdellweg, line 75: This is do not understand. And maybe it's a deficiency in our data model. I would expect that the same content artifact could be published as part of a multitude of different publications and therefore be associated with multiple published artifacts.
Also a publication (that is not pass-through) may even decide to not publish some content artifacts at all. Should we not derive the list of relative paths and their artifacts from the query above if a publication is provided?

parthaa: > I would expect that the same content artifact could be published as part of a multitude of different publications and therefore be associated with multiple published artifacts.

That is correct. Same content artifact may be associated to multiple publications. I should probably just remove the `ca.published_artifact` part. The idea is this method is given the correct publication to use. So may be a change like below should be sufficient?

```suggestion
        if publication:
```



> Also a publication (that is not pass-through) may even decide to not publish some content artifacts at all. Should we not derive the list of relative paths and their artifacts from the query above if a publication is provided?
Good point I dont know if we can do a fs_export of a pass through publication. May be I should change it to

 ```suggestion
        if publication and not publication.pass_through:
```

Comment (parthaa) [reply pulpcore/app/tasks/export.py]: updated

Comment (mdellweg) [reply pulpcore/app/tasks/export.py]: Digging this up from memory: A pass through publication serves all content artifacts in the repo-version according to it's relative-path (silently assuming there is one). But also some extra published artifacts (usually metadata).
So I think, you want to include all the published artifacts if there is a publication and all the content artifacts if there is none or it is pass through. Right?
As far as I understand, the goal is to create a tar representation of the looks of the publication on the webserver.

Comment (parthaa) [reply pulpcore/app/tasks/export.py]: > So I think, you want to include all the published artifacts if there is a publication and all the content artifacts if there is none or it is pass through. Right?

That's at least what the commit is trying to accomplish :). I don't know the intricacies of the pass_through publications.

File: pulpcore/app/tasks/export.py [hunk 2]
```diff
@@ -58,10 +60,19 @@ def _export_to_file_system(path, content_artifacts, method=FS_EXPORT_METHODS.WRI
...
+            for content_artifact_pk, relative_path in publication.published_artifact.values_list(
+                "content_artifact__pk", "relative_path"
+            )
+        }
+
     for ca in content_artifacts.select_related("artifact").iterator():
         artifact = ca.artifact
-        dest = os.path.join(path, ca.relative_path)
-
+        relative_path = publication_paths.get(ca.pk, ca.relative_path)
```
parthaa, line 74: This is the key line.  `publication_paths` is empty if a publication was not provided or if it was a pass_through. The dictionary get call first searches by `pk' and it if it cant find it returns `ca.relative_path`

daviddavis: This ^ is true: a publication could be pass-through and still have published artifacts. 

And to add to this: it's possible that a published artifact and content artifact in a pass-through publication have the same relative path. In which case the published artifact gets served as the content app first checks the published artifacts before it checks content artifacts:

https://github.com/pulp/pulpcore/blob/cd3a090f5c39171d2326a7a8c3cb682451af2d00/pulpcore/content/handler.py#L502-L572

pulpcore/app/tasks/export.py
daviddavis: Perhaps the code for the FilesystemExporter and PulpExporter should be broken up to better differentiate the two (not in this PR though).

pulpcore/app/tasks/export.py
daviddavis: Suppose you had a situation with a published artifact X that had relative path `a/b/c` and content artifact A, as well as a content artifact B with no published artifact and relative_path `a/b/c`. If so, then A could get written to the filesystem before A then B would overwrite it which would be undesirable. 

Now is this situation possible in Pulp? I think it's probably theoretically possible but not sure if it's actually happening.

daviddavis: My suggestion would be to have `_export_to_file_system` take a mapping of file paths to artifacts and have it write the files to disk. Don't have it be aware of content artifacts or publication.

Then update `fs_publication_export` to first iterate through the content artifacts (if it's a pass-through publication) and then through the published artifacts so that the latter overwrites the former. I think this would greatly simplify the logic.

bmbouter:
I drafted this and when it's ready for review please undraft. If you don't permissions please ping on the PR and someone else can.

pulpcore/app/tasks/export.py
parthaa: > My suggestion would be to have `_export_to_file_system` take a mapping of file paths to artifacts and have it write the files to disk. Don't have it be aware of content artifacts or publication.
> 
> Then update `fs_publication_export` to first iterate through the content artifacts (if it's a pass-through publication) and then go through the published artifacts so that the latter overwrites the former. I think this would greatly simplify the logic.

In a pass-through publication will the relative path of the published_artifact be the same as relative path of the content-artifact?
If the answer is no they can be different then how to create a unique mapping between relative_path and artifact? The publication artifact will have a relative_path of `a/b/c` while artifact itself has a relative path of `c`. Your suggested strategy of  ```update `fs_publication_export` to first iterate through the content artifacts (if it's a pass-through publication) and then go through the published artifacts``` will result in `{'a/b/c': artifact1 , 'c': artifact2}` . Shouldn't I just ignore `c: artifact2`

parthaa: @daviddavis I have updated the PR to try your strategy. This definitely works for me as long as in the pass through case relative path of the published_artifact be the same as relative path of the content-artifact

daviddavis: Looks good 👍

parthaa: @ggainey ^

CHANGES/2933.doc
```diff
@@ -0,0 +1 @@
+Use published relative paths for FS Exporter
```
mdellweg, line 1: This should either be a bugfix or a feature. Also please use punctuation.

pulpcore/app/tasks/export.py
```diff
@@ -112,7 +118,21 @@ def fs_publication_export(exporter_pk, publication_pk):
             content__in=publication.repository_version.content
         )
 
-    _export_to_file_system(exporter.path, content_artifacts)
+    _validate_fs_export(content_artifacts)
+
+    relative_path_to_artifacts = {}
+    if publication.pass_through:
+        relative_path_to_artifacts = {
```
mdellweg, line 125: I would prefer if we found a way to express this with a queryset instead of loading everything into the ram at once. I don't want to hold this work back though. (How do you feel about writing a test to secure later refactoring?)

parthaa: > I would prefer if we found a way to express this with a queryset instead of loading everything into the ram at once. I don't want to hold this work back though. 

You are talking about a dictionary of 40K entries for RHEL-7. That is about as bad as this gets. 

> (How do you feel about writing a test to secure later refactoring?)
I am open to it. But I noticed no tests for this file or any of the import/export stuff other than testing serializer output. I also  was not sure how to set things up. I am open to any guidance on this.

daviddavis: There are some tests in pulp_file:

https://github.com/pulp/pulp_file/blob/main/pulp_file/tests/functional/api/from_pulpcore/test_filesystemexport.py

mdellweg (approved):

pulpcore/app/tasks/export.py
parthaa: thanks

parthaa: Oh its in pulp_file and not core ?

parthaa: Can I use a test which refers to `pulpcore.client.pulp_file.RepositorySyncURL` in  pulpcore  ?

mdellweg: We moved all the tests "using a plugin" which is almost all of them over to the plugin (namely pulp_file). But they will be exercised in both CI's in the end.

pulpcore/app/tasks/export.py
```diff
@@ -112,7 +118,21 @@ def fs_publication_export(exporter_pk, publication_pk):
...
+            ca.relative_path: ca.artifact
+            for ca in content_artifacts.select_related("artifact").iterator()
+        }
+
+    for ca in publication.published_artifact.select_related(
```
dkliban, line 130: `ca` should probably be `pa`

dkliban (approved):

parthaa:
Just tested exporting rhel7 with 60 gb content as hardlinks and it worked well :)

parthaa:
After export 
```bash
$ ls content/dist/rhel/server/7/7Server/x86_64/os/Packages/
3  a  b  c  d  e  f  g  h  i  j  k  l  m  n  o  p  q  r  s  t  u  v  w  x  y  z
$ du -sh content
60G	content
```
all hardlinked :)

dralley (approved):

dralley:
Would there be sufficient added utility to also justify a management command [0] using the same backend?

IMO, they have very different use cases.  The filesystem exporter requires you to manually link particular publications with a particular exporter and tell it to export to particular path via API calls, whereas my management command just dumps everything publication-based that is distributed underneath a distribution "base_path" prefix, on a per-plugin basis if desired. One command can essentially dump the entire tree of distributed repos for a given plugin without any prior setup.

As a debugging / support tool the latter seems much more useful, whereas I suppose the current filesystem exporter is meant to be more programmable and API-driven, but not something a human would want to work with directly.

[0] https://github.com/pulp/pulpcore/pull/2484

parthaa:
> Would there be sufficient added utility to also justify a management command [0] using the same backend?

Yes management tool to directly dump publications can be definitely useful. Had i known about that I probably would have tried to use it for this PR also :). That being said there is value in a programmable api similar to the way regular exporter works. I guess it could be useful for people who want to use this kinda export in `Sat 6.10/Pulp 3.14` since we dont have fs exporters there. That output looks so pretty :). Sounds like the logic you are using there is similar to this PR?

dralley:
It is similar but half-baked because I threw it together as a prototype.  No hardlinks, and it doesn't support pass-through publications very well, and it needs to be more tolerant of directories already existing.  I would just keep the basic idea but replace some of the code with what is written here.

mdellweg:
> It is similar but half-baked because I threw it together as a prototype. No hardlinks, and it doesn't support pass-through publications very well, and it needs to be more tolerant of directories already existing. I would just keep the frontend but replace some of the code with what is written here.

So your PR would call into the common "export me that publication there" helper function? Sounds like a nice design. But let's keep the second PR for that.

PR #2963, opened 2022-07-12
Artifact serving distribution

### LINKED ISSUE #2785: Artifact serving distribution
For external validation services, we would need a singleton distribution that is able to serve artifacts by sha256. It should be configured with the redirecting contentguard.

mdellweg:
fixes #2785


pulpcore/app/models/publication.py
```diff
@@ -553,3 +556,42 @@ def invalidate_cache(self):
...
+        return self.content_guard.cast().preauthenticate_url(plain_url)
+
+    def content_handler(self, path):
+        """Serve an artifact or return 404."""
+        uuid = path.rstrip("/")
+        if re.match(r"[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}", uuid):
+            artifact = Artifact.objects.get(pk=uuid)
+            if artifact:
+                return ArtifactResponse(artifact)
+        return HTTPNotFound()
```
gerrod3, line 591: Please raise this exception.

pulpcore/app/models/publication.py
```diff
@@ -553,3 +556,36 @@ def invalidate_cache(self):
         if settings.CACHE_ENABLED:
             Cache().delete(base_key=self.base_path)
             # Can also preload cache here possibly
+
+
+class ArtifactDistribution(Distribution):
+    """Serve artifacts by their sha256.
```
mdellweg, line 562:
```suggestion
    """Serve artifacts by their uuid.
```

mdellweg:
@gerrod3 Do you think this still needs some special caching logic?

gerrod3 (commented):
Looks good, there shouldn't be any cache changes needed since this is for file-system storage and the urls are not time-sensitive. One interesting thing is that going to `/pulp/content/` will show `/artifacts/` as a distribution to access but clicking the link will result in a `PermissionError`.

pulpcore/app/models/publication.py
```diff
@@ -553,3 +556,36 @@ def invalidate_cache(self):
         if settings.CACHE_ENABLED:
             Cache().delete(base_key=self.base_path)
             # Can also preload cache here possibly
+
+
+class ArtifactDistribution(Distribution):
+    """Serve artifacts by their uuid.
+    This is in tech-preview.
```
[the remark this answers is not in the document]
mdellweg: I switched the implementation from digest to uuid, because that is the pk, and it really does not matter. Also it should prevent to "steal a file where you happened to know the digest".

pulpcore/app/apps.py
```diff
@@ -318,3 +324,31 @@ def _get_permission(perm):
...
+        try:
+            ArtifactDistribution = apps.get_model("core", "ArtifactDistribution")
+            ContentRedirectContentGuard = apps.get_model("core", "ContentRedirectContentGuard")
+        except LookupError:
+            if verbosity >= 1:
+                print(_("ArtifactDistribution model does not exist. Skipping initialization."))
+            return
+        try:
+            ArtifactDistribution.objects.get(name="artifact redirect")
+        except ArtifactDistribution.DoesNotExist:
```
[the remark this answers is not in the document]
mdellweg: I thought it would be ok to fail; this feature would not work correctly in that case. But i can also change the base_path to a generated uuid.

CHANGES/plugin_api/2785.feature
```diff
@@ -0,0 +1 @@
+Added ``get_artifact_url`` to emit preauthed urls to existing artifacts.
```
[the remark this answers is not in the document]
mdellweg: Why specific to file storage? It creates a URL to reach the artifact no matter how.

mdellweg:
> Looks good, there shouldn't be any cache changes needed since this is for file-system storage and the urls are not time-sensitive. One interesting thing is that going to `/pulp/content/` will show `/artifacts/` as a distribution to access but clicking the link will result in a `PermissionError`.

The permission error is the content-guard in action. I believe this is expected. And for the URL's not having a timeout, you remind me of something. I wanted to add this for a long time...

CHANGES/plugin_api/2785.feature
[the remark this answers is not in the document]
gerrod3: I think docs would be good for file storage since it "behaves" differently then the others since it is using a singleton distribution behind the scenes to get the url. Mostly I feel that understanding this method for file storage is not as intuitive as the others, but maybe that's just me.

awcrosby:
In regards to this change affecting collection upload validation with ansible-test... As long as the `file_url` parameter passed to the importer https://github.com/ansible/galaxy-importer/blob/4b93075064d063f216fcd97096fa796a23767c40/galaxy_importer/collection.py#L42 continues to be able to download the collection, this should work. The openshift job that runs `ansible-test sanity` downloads the artifact via that `file_url` passed into the job's container.

gerrod3 (approved):

pulpcore/tests/functional/api/test_artifact_distribution.py
```diff
@@ -0,0 +1,20 @@
+import requests
```
dkliban, line 1: We are trying to move all our requests usage to aiohttp. Here is an example from pulp_file[0].

[0] https://github.com/pulp/pulp_file/blob/main/pulp_file/tests/functional/utils.py#L315-L329

mdellweg: I heard a bout this but i cannot see the reason. Our tests don't benefit from adding async code to run sequentially with the rest of the test.

dkliban: Pulp already carries `aiohttp` as a dependency. The idea is to use it and not introduce additional dependencies. I would like to get this work merged, so I am going to approve the PR as is. I'll leave it up to you to change the implementation or not.

dkliban (approved):

pulpcore/app/apps.py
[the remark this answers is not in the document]
mdellweg: @pulp/core Here's an open question.

bmbouter:
+1 to using a random uuid name and a random uuid base path.

I think this "special distribution" should be hidden from users because it's not for users. The easiest way (to me) to do that is to have the "list" view of the content app exclude the `Artifactdistribution` type entirely.

ipanova:
> +1 to using a random uuid name and a random uuid base path.
> 
> I think this "special distribution" should be hidden from users because it's not for users. The easiest way (to me) to do that is to have the "list" view of the content app exclude the `Artifactdistribution` type entirely.

+1

pulpcore/app/models/publication.py
```diff
@@ -553,3 +556,36 @@ def invalidate_cache(self):
...
+        plain_url = urljoin(
+            urljoin(urljoin(origin, prefix + "/"), base_path + "/"), str(artifact.pk)
+        )
+        return self.content_guard.cast().preauthenticate_url(plain_url)
+
+    def content_handler(self, path):
+        """Serve an artifact or return 404."""
+        uuid = path.rstrip("/")
+        if re.match(r"[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}", uuid):
+            artifact = Artifact.objects.get(pk=uuid)
```
ipanova, line 582: use filter() instead or if with get() then you need to catch doesnotexist error

pulpcore/app/util.py
```diff
@@ -256,3 +257,45 @@ def configure_telemetry():
         )
     else:
         models.TaskSchedule.objects.filter(task_name=task_name).delete()
+
+
+@lru_cache(maxsize=1)
+def _artifact_serving_distribution():
+    return models.ArtifactDistribution.objects.get()
```
ipanova, line 264: would be great to ensure this is a singleton. maybe add a check in the pre_save hook?

pulpcore/app/models/publication.py
mdellweg:
```suggestion
            artifact = Artifact.objects.first(pk=uuid)
```
Is probably a good solution.

ipanova: first() does not take any args/kwargs, it should be filter()

mdellweg: and `filter` returns a query set. So `filter(pk=uuid).first()` ?

ipanova: yeah that will work

ipanova (approved):
thanks!

PR #2966, opened 2022-07-13
Adds admin_user pytest fixture

dkliban:
[noissue]


bmbouter:
I'm +1 on this, but I'd like to defer my review to @mdellweg and @gerrod3

gerrod3 (approved):
Wouldn't have thought to make this since all bindings use admin by default, but this can be very useful. :+1:

mdellweg (approved):

PR #2994, opened 2022-07-20
Update CI

mdellweg:
[noissue]


pyproject.toml
```diff
@@ -79,6 +79,7 @@ ignore = [
     "doc_requirements.txt",
     "docs/**",
     "flake8.cfg",
+    "lint_requirements.txt",
```
mdellweg, line 82: Sadly this change has to be carried out manually.

PR #2999, opened 2022-07-22
Add info field to RepositoryVersion

### LINKED ISSUE #2998: Add another field to the `RepositoryVersion` model to store additional information.

We are looking to implement an optimize sync feature in `pulp_deb`. In order to do that we want to store data in a `dict` format which contains information about the remote options used in the latest sync as well as `sha256` values for the repository files. These will be used to compare when trying to sync a repository again to see if something has changed and skip certain steps if it hasn't.

Adding another field to `RepositoryVersion` model that can store a `.JSONField(default=dict)`. This way a repository can always look up the lastest version to get the previous data.

`pulp_rpm` uses their `Repository` model to save this data which works too. However there were concerns of the mutability of the data when stored this way.

Another solution would be adding a subclass of `RepositoryVersion` to `pulp_deb`. Since there is not any subclass of this model present in any plugin so far, this seems a bit overkill to just add one additional field.

**Additional context**

For context the PR for this feature in `pulp_deb`: https://github.com/pulp/pulp_deb/pull/570

hstct:
closes #2998

The name of the field was chosen arbitrarily and is certainly open to be changed into whatever fits the best.


pulpcore/app/models/repository.py
```diff
@@ -605,6 +605,7 @@ class RepositoryVersion(BaseModel):
     number = models.PositiveIntegerField(db_index=True)
     complete = models.BooleanField(db_index=True, default=False)
     base_version = models.ForeignKey("RepositoryVersion", null=True, on_delete=models.SET_NULL)
+    sync_details = models.JSONField(default=dict)
```
mdellweg, line 608: I would not consider this data to be limited to sync changes. I think we could use it in general to store "how was this version created"?

pulpcore/app/serializers/repository.py
```diff
@@ -428,6 +428,7 @@ class Meta:
             "repository",
             "base_version",
             "content_summary",
+            "sync_details",
```
mdellweg, line 431: I'm unsure it we really intend to expose this data.

quba42: I suppose if we don't expose it here, we can still look it up in the DB if ever it is needed for debugging. And since it is essentially "arbitrary data structure", we probably don't want to expose it via the serializer.

pulpcore/app/models/repository.py
quba42: Agreed. There is no reason, why it could not be used by an upload action, to say that this version was created by an upload action. This would suggest using a very generic name like `version_details` (it is a very generic JSON field after all).

pulpcore/app/serializers/repository.py
hstct: Sure I just wasn't sure if this is needed or not. Can revert this once there is a consensus on the name of the field.

quba42: I kind of like having it exposed, but that is probably "as a developer"...

quba42:
We now have a working state of the [pulp_deb optmize sync](https://github.com/pulp/pulp_deb/pull/570) that needs this field.

As such, it would be nice if we could settle on a name for the field, so we can update this PR and move towards final review.

`sync_details` is not a good name, since I definitely envision the pulp_deb advanced copy API also making use of this field in the future (that is it is not just intended for syncs).

Personally I vote for `creation_details`, because this is somewhat opinionated about what this field is intended for. Given that the field can factually contain arbitrary json data, a completely generic name like `version_details` would alternatively also make sense.

Does anyone have a different opinion, or should we go ahead and change this PR to use `creation_details`?

dkliban:
What about just 'details' or 'info' ?

ggainey:
> 

+1 to "info"

quba42:
The current consensus is "info".

hstct:
Updated the PR renamed the field to `info` and removed exposing the data in the serializer.

CHANGES/2998.feature
```diff
@@ -0,0 +1 @@
+Add info field to RepositoryVersion.
```
daviddavis, line 1: My reading of the issue leads me to believe this is a plugin writer field and not a field that would be exposed to users. Is that correct? If so, this probably belongs in the plugin writer changelog.

quba42: Looked up what the path should be on another issue: `CHANGES/plugin_api/2998.feature`

hstct: Thanks I didn't know there was a specific changelog for that. It should be resolved now.

ipanova:
> The current consensus is "info".

Sorry for back and forth on this, I would personally vote for more explicit name like `version_details` or was the intention to leave this as generic as possible to leave the door opened to add such field to other models too? @ggainey @dkliban

quba42:
@ipanova I guess one could consider the `version_` part in `RepositoryVersion.version_<whatever>` redundant?
If so, then `info` or `details` seems just as good as `version_info` or `version_details`. (I have no opinion on `info` vs `details`).

The other thought was that since the field is inherently generic (can contain arbitrary Json), the name should also be generic?
My original thought was that the field name _could_ be more "opinionated" than the field itself, which is why I suggested `creation_details`, but this was never a strong opinion. At the end of the day I am fine with any of the suggestions within this thread (so long as they don't explicitly include `sync` in the name).

My 2ct.

dkliban (approved):

ipanova (approved):
@dralley if you want to also take a look

dralley:
If we're going for a generic name like "info" then we maybe ought to consider whether this could be more broadly applicable than just repository versions, something like the Pulp 2 "scratchpad".  For example I can see a use case having publications store particular details about how they were created as well.

But on the flip side, I'm not sure if we want to resurrect the "scratchpad", and we already have a label system which has a small overlap in intended purpose.

ipanova:
> If we're going for a generic name like "info" then we maybe ought to consider whether this could be more broadly applicable than just repository versions, something like the Pulp 2 "scratchpad". For example I can see a use case having publications store particular details about how they were created as well.
> 
> But on the flip side, I'm not sure if we want to resurrect the "scratchpad", and we already have a label system which has a small overlap in intended purpose.

If my recollection is correct, I believe we discussed this aspect in pulpcore meeting, that's why, I believe we're trying to pick as generic name as we can.
I agree with you about the feeling of adding this to all models, hence I'd say let's add only where we have a request for now?

dralley:
Unfortunately once it's been added to a model it can't just be easily moved to another model in the inheritance hierarchy.  You can make a new field with a different name but keeping the same name is really problematic.

ipanova:
> Unfortunately once it's been added to a model it can't just be easily moved to another model in the inheritance hierarchy. You can make a new field with a different name but keeping the same name is really problematic.

oh, i see, yes

dralley (approved):
If we were ever to remove this and move it to the base model we would likely call it `pulp_info`, precluding the aforementioned concern.

PR #3001, opened 2022-07-25
When importing add only content to repository

### LINKED ISSUE #3064: Import attempts to add everything, not only content, to the target repository

mdellweg:
~~fixes TBA~~
fixes #3064

### DIFF
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -36,6 +36,8 @@ from pulpcore.app.modelresource import (
 from pulpcore.constants import TASK_STATES
 from pulpcore.tasking.tasks import dispatch
 
+from pulpcore.plugin.importexport import BaseContentResource
+
 log = getLogger(__name__)
 
 ARTIFACT_FILE = "pulpcore.app.modelresource.ArtifactResource.json"
@@ -236,13 +238,14 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
 
         resulting_content_ids = []
         for res_class in cfg.exportable_classes:
-            content_count = 0
             filename = f"{res_class.__module__}.{res_class.__name__}.json"
             for a_result in _import_file(os.path.join(rv_path, filename), res_class, retry=True):
-                content_count += len(a_result.rows)
-                resulting_content_ids.extend(
-                    row.object_id for row in a_result.rows if row.import_type in ("new", "update")
-                )
+                if issubclass(res_class, BaseContentResource):
+                    resulting_content_ids.extend(
+                        row.object_id
+                        for row in a_result.rows
+                        if row.import_type in ("new", "update")
+                    )
 
         # Once all content exists, create the ContentArtifact links
         ca_path = os.path.join(rv_path, CA_FILE)
--- a/CHANGES/3064.bugfix
+++ b/CHANGES/3064.bugfix
@@ -1,0 +1,3 @@
+Fixed a bug in import code where all objects imported would also be added to the target repository
+by their UUID. In case of a UUID-collision with content, unwanted content may have ended up being
+in the next repository version.


ggainey (commented):
What's driving this? Is this something that should be caught at export time? Or - is this "we're exporting DetailRepository now, and it's Confusing Things"?

At a minimum, an explanatory comment on "why" would not be amiss.

mdellweg:
This PR is primarily to not forget.
I was chasing a nonexisting bug due to my test expectation being wrong. So i added debug statements listing the units imported and added to the repository. This difference shown was misleading and lead me to believe that the content type i was after  was actually not imported.
But what I realized is that we do export objects that are not content (like `Collection` in the ansible example) alongside with content. Then after importing, we try to add it to the repository where it will almost certainly be ignore silenty, because UUID's are very rare to coincide. But technically, I believe it is possible to have a `Collection` with the same UUID as a `FileContent` (they don't share a common base table at the end). And in this very unlikely case, we would erroneously add this `FileContent` to the repository that was not even part of the import.

This needs an issue for sure. But you see it's hard to even formulate.

ggainey:
> ~fixes TBA~ fixes #3064

:+1:

ggainey (approved):
Good catch, lg2m

pulpcore/app/tasks/importer.py
```diff
@@ -239,10 +241,13 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
             content_count = 0
             filename = f"{res_class.__module__}.{res_class.__name__}.json"
             for a_result in _import_file(os.path.join(rv_path, filename), res_class, retry=True):
-                content_count += len(a_result.rows)
-                resulting_content_ids.extend(
-                    row.object_id for row in a_result.rows if row.import_type in ("new", "update")
-                )
+                if issubclass(res_class, BaseContentResource):
+                    content_count += len(a_result.rows)
```
lubosmj, line 245: I am not very familiar with the ansible plugin, but if a `Collection` is a thing used/referenced by real content units, I still think that we should record its presence.

Maybe, `content_count` should be renamed to `imported_units_count` to better represent its nature. The goal of the import procedure is to import all the resources archived in the exported tarball, am I correct?

mdellweg: Hmm you are stirring up a hornet's nest. A `Collection` is a model that should never have existed IMHO. It's merely the placeholder for a collection name, while the actual content was called `CollectionVersion` (Calling that one `Collection` would kind of match to the thing we call a `Package` in other plugins. And yes a package has a version...).
But to be fair, i think we are calculating something here we never intend to use (We are assigning a different value to it further down.). So i will try _just_ removing this line altogether.

lubosmj: Hmm. Nice overview. I am fine with removing it as well.

lubosmj: @mdellweg and I noticed that we are incorrectly counting the number of imported content. We are overwriting the `content_count` variable in a loop over and over again:
https://github.com/pulp/pulpcore/blob/90142202f0ecc2e47dafbc2c517691fdc60b0eef/pulpcore/app/tasks/importer.py#L269

This issue should be addressed as well. Maybe in a separate PR.

lubosmj (approved):

bmbouter:
I know very little about this, but I'm glad to see it so I'll comment to help move it forward. I know very little about import/export so all I can think to do without a third SME reviewer is to ask for a test. Is there some way to have a test demonstrate this bug as fixed?

As an aside, if there isn't another import/export SME besides @ggainey and @lubosmj it would be good to find one to provide the third review.

mdellweg:
It is not testable unless one could provoke a uuid collision between a content in pulp and a not-a-content in the import.

ggainey:
> I understand re it not being testable. My question shows how little I know about it (apologies).

The existing PIE testcases at the very least show that this isn't *hurting* anything.
 
> If there isn't a third SME for import/export, and with the two SMEs both approving it, I'd say +1 to admin merging it. My review for a second check-mark I roughly the same as an admin merge from an understanding point anyway.

There is a third SME, except that's @mdellweg , who shouldn't be approving his own change :)

> What do you all think about my recommendation?

I am ok with either. I don't have admin-merge-perms tho, so Someone Else will need to push buttons here.

bmbouter:
Oh I see; thank you. So all SMEs are approving this and it's a github perms thing that isn't allowing a normal merge. I'm going to admin merge in this case then.

mdellweg:
> Oh I see; thank you. So all SMEs are approving this and it's a github perms thing that isn't allowing a normal merge. I'm going to admin merge in this case then.

Thank you.
On side note, yes we require two green approvals in pulpcore. But i do not think that means we need to find two SME's (with commit bit) for a change. We don't have that many... I'd still rather avoid admin-merge. One SME and an admin sounds like an even stricter requirement to me.

PR #3017, opened 2022-07-27
Rename fixture admin_user to pulp_admin_user

mdellweg:
The old name has a conflict with pytest-django.

[noissue]

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -250,7 +250,7 @@ def anonymous_user(bindings_cfg):
 
 
 @pytest.fixture(scope="session")
-def admin_user(bindings_cfg):
+def pulp_admin_user(bindings_cfg):
     class AdminUser:
         def __init__(self):
             self.username = bindings_cfg.username


pulpcore/tests/functional/__init__.py
```diff
@@ -250,7 +250,15 @@ def __exit__(self, exc_type, exc_value, traceback):
 
 
 @pytest.fixture(scope="session")
-def admin_user(bindings_cfg):
+def admin_user(pulp_admin_user):
```
bmbouter, line 253: I think almost nothing is using this right now. If that's the case I propose just renaming it. Besides if it's conflicting we shouldn't continue to use it.

bmbouter:
It's definitly conflicting I see it when I `pytest --fixtures`

```
admin_user -- ../../../../../usr/local/lib/pulp/lib64/python3.10/site-packages/pytest_django/fixtures.py:407
    A Django admin user.
```

bmbouter:
Silly question but how is ours and the one from pytest-django different?

mdellweg:
> Silly question but how is ours and the one from pytest-django different?

Our's has the credentials to talk to a pulp server, pytest-django's (i can only guess) is probably an instance of `django.contrib.auth.model.User` persisted in the test database.

dkliban (approved):

bmbouter:
I think you'll need a required PR for this to be updated too: https://github.com/pulp/pulp_file/blob/d71c28437e635d62ef3e154b6d9e66e248bcff37/pulp_file/tests/functional/api/from_pulpcore/test_content_guard.py#L23

bmbouter (approved):

dralley (approved):

mdellweg:
Seems like it was introduced after 3.20

bmbouter:
Correct, the commit adding it only shows as being on `main` https://github.com/pulp/pulpcore/commit/cd3a090f5c39171d2326a7a8c3cb682451af2d00

PR #3025, opened 2022-07-29
Ported data posting for dev envs to protobuf

bmbouter:
This uses the Protocol Buffer bindings with the definition provided by
the https://github.com/pulp/analytics.pulpproject.org repo.

[noissue]


pulpcore/app/protobuf/telemetry_pb2.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
+# source: telemetry.proto
+"""Generated protocol buffer code."""
```
mdellweg, line 4: I do have some serious reservation to check in generated files.
At the very least, i am missing the `telemetry.proto` to regenerate it.

mdellweg (dismissed):

pulpcore/app/protobuf/telemetry_pb2.py
bmbouter: It's checked in [here](https://github.com/pulp/analytics.pulpproject.org/blob/dev/telemetry.proto), and the built files end up going in to both repos.

mdellweg: I hope you have a sound plan to keep them in sync. Will the protocol be versioned?

Withdrawing my review now.

bmbouter: The plan is to regenerate as needed. We're following the protobuf best practices in the .proto definition file, which gives us both forward and backwards compatibility. Because of that property, it doesn't need to be versioned at all. Here's a bit more info on it https://developers.google.com/protocol-buffers/docs/pythontutorial.

This is especially nice because even as the Server receives newer versions it'll remain backwards compatible for older versions of pulpcore that will never learn how to submit "newer style payloads".

ggainey (approved):
I am also a little concerned at having the source for a file be in a different repo - but the explanation for How It Will Be OK, makes sense. Maybe we'll change that at some point, we're going to continue learning how best to organize this feature as we go. For where we are now, I'm good with this (and I am *very* good w/ the move to protobuf in general!)

bmbouter:
Any input on how to make it better is welcome. One idea is to have for later would be to have it be its own python package on pypi and have both projects consume it instead of checking in the source. I'm not ready to do that right now since this is so new, but maybe that's a better solution we could put in place later.

mdellweg (approved):

ipanova (approved):

PR #3032, opened 2022-08-04
 Do not dispatch telemetry tasks for production systems. 

### LINKED ISSUE #3015: Telemetry task gives false impression that telemetry data is being sent
I see in the logs that a telemetry task is being executed regularly (every day):

```
        {
            "child_tasks": [],
            "created_resources": [],
            "error": null,
            "finished_at": "2022-07-25T21:32:09.435330Z",
            "logging_cid": "",
            "name": "pulpcore.app.tasks.telemetry.post_telemetry",
            "parent_task": null,
            "progress_reports": [],
            "pulp_created": "2022-07-25T21:32:09.383372Z",
            "pulp_href": "/pulp/api/v3/tasks/7f4c54d6-fcda-4006-9da1-3c6910a18171/",
            "reserved_resources_record": [],
            "started_at": "2022-07-25T21:32:09.419785Z",
            "state": "completed",
            "task_group": null,
            "worker": "/pulp/api/v3/workers/c609a0d5-45cc-4756-af3c-4b50cd7520b1/"
        }
```

It gave me the impression that Pulp was sending telemetry data. It was only when I dug into the code that I saw that it only does this for dev installs. Is there a way to not execute this task if telemetry data is not being sent?

ipanova:
closes #3015


pulpcore/tasking/tasks.py
```diff
@@ -102,13 +102,19 @@ def dispatch(
...
     # Dispatch all tasks old enough and not still running
     for task_schedule in TaskSchedule.objects.filter(next_dispatch__lte=now).filter(
         Q(last_task=None) | Q(last_task__state__in=TASK_FINAL_STATES)
     ):
         try:
-            if task_schedule.dispatch_interval is None:
+
+            # Initially only dev systems receive posted data.
+            if task_schedule.task_name == telemetry_task and url == PRODUCTION_URL:
+                continue
```
mdellweg, line 116: I don't think, this is the proper place for this to live. This will keep the schedule in not-run state forever and every worker will evaluate it over and over.
Instead i would prefer, if the task schedule would not even be implemented in case of production systems.
Isn't this possible in the post-migrate hook?

ipanova: Yes, that was my preference too but at that moment apps are not loaded.

bmbouter: I agree this isn't ideal, but it's ok if we don't figure it out in this PR though (to me) because the actual design will include a user-facing setting which isn't here yet. That setting is what users will set to disable telemetry and a startup hook can set that. The setting can change on each startup so even a post-migrate hook wouldn't be right; it needs to be per startup. Given we can't do that right now I'm +1 to merging this so I'm lgtm-ing.

bmbouter (approved):
Thanks @ipanova 

@mdellweg do you want to be second reviewer? If not someone from the telemetry working group can be.

pulpcore/tasking/tasks.py
mdellweg: What is the reason we cannot do that right now? If it's the agreed on proper way to do it, then let's just do it. In a startup hook we should have all apps, no?

bmbouter: We can't do it right now because that startup hook is supposed to check a setting which doesn't yet exist. We don't want to add that switch right now because this hasn't yet been scheduled as a user-facing feature yet.

mdellweg: With or without the setting being present or another condition deciding on the fate of the schedule, we seem to agree that a startup hook is the proper place to apply that logic. Unless it's really impossible to identify dev-plugins there, i really think we should move it there. Checking this corner case in every single workers heartbeat is really not the right place.

ipanova: @mdellweg it was already stated that it is not possible to identify dev-plugins at the point because apps are not loaded. Do you have an alternative working solution given all what has been discussed and commented here?

mdellweg: It was stated the info is not available in post_migrate. But the question about a startup hook (whatever that is) now. Can we decide there?
What i'm asking for is that we find a solution where when we don't want the task to run, we don't schedule it.

bmbouter: My understanding of the request would be something like:

Create a post-startup (not post-migrate) hook in apps.py (where we usually put these things) that would run the dev system check. If it's a dev-system create_or_update the db entry. If it's not a dev system ensure the entry is deleted.

The helpful thing about doing this is that we're going to have to make this anyway very soon when we introduce the setting. Also we can apply it as a backport to 3.20 without additional effort.

pulpcore/app/apps.py
```diff
@@ -211,15 +211,24 @@ class PulpAppConfig(PulpPluginAppConfig):
     def ready(self):
         super().ready()
         from . import checks  # noqa
+        from django.db import connection
+
+        # existing installation
+        if connection.introspection.table_names():
+
+            _configure_telemetry(self)
```
ipanova, line 219: This code path fails for new installations because TaskSchedule table is not created yet, that's why I left the post_migrate signal too. The only thing I am hesitant about is whether it's ok to keep this post_migrate `_populate_telemetry_periodic_task` in the else branch or let it execute regards of whether it's a new or upgraded installation.

mdellweg: I think we don't need the post_migrate signal, because we will call the ready method in every single process later anyway.
So just making it not fail premigrated should be sufficient.

pulpcore/app/apps.py
```diff
@@ -211,15 +211,23 @@ class PulpAppConfig(PulpPluginAppConfig):
...
+        if connection.introspection.table_names():
+            _configure_telemetry(self.apps)
+
+        else:
+            # new installation
+            post_migrate.connect(
+                _populate_telemetry_periodic_task,
+                sender=self,
+                dispatch_uid="populate_telemetry_periodic_task_identifier",
+            )
```
bmbouter, line 226: I've been thinking this over, and here is what I came with. For complexity reduction I'd recommend not having this `if` statement conditional at all and having it always call `_configure_telemetry()`. Then have the `_configure_telemetry` call determine if anything needs to be created or deleted.

I think that would be less complex and ultimately aligned with the eventual design which would have to have it checked at startup anyway (so there would be no post_migrate) anything.

pulpcore/app/apps.py
bmbouter: I didn't see your comment, but my comment below basically is a +1

pulpcore/app/apps.py
ipanova: cool, this is what i was wondering about, ty.

pulpcore/app/apps.py
ipanova: @mdellweg sounds good

bmbouter (approved):
This looks good to me. Once merged let's ask patchback to backport to 3.20 also please. I will build my PR on top of this one. Thank you all!

mdellweg (approved):
Lgtm

pulpcore/app/apps.py
```diff
@@ -212,14 +212,11 @@ def ready(self):
         super().ready()
         from . import checks  # noqa
 
+        _configure_telemetry(self.apps)
```
fao89, line 215: pulp readiness is tied with DB readiness now

ipanova: The recap from the debugging session: we need to build docs inside the container because django.setup() collects all installed apps among which pulpcore.app is and then initializes it with ready() which now requires a db connection in order to make db calls. This is different than it was before - before we had just post_migrate signal which was respectively triggered post migration.

ipanova: @fao89 will create a template PR to run docs building inside of the container

fao89:
https://github.com/pulp/plugin_template/pull/678

PR #3038, opened 2022-08-04
fixes #3036 -Setup the SSL Context correctly for https proxies

### LINKED ISSUE #3036: HTTPs Proxies not working.
**Version**
tfm-pulpcore-python3-pulpcore-3.17.7-1.el7.noarch
tfm-pulpcore-python3-aiohttp-3.8.1-2.el7.x86_64

**Describe the bug**
There was HTTPS proxy tunneling support added to the aiohttp library for 3.8.1 => https://github.com/aio-libs/aiohttp/pull/5992
However pulpcore does not 

1. Have the necessary bindings to get this mode to run.
2. Expects the CA Cert of the proxy to be concatenated to the repo's remote CA cert instead of expecting it in the default trust store (as it used to do in pulp2). 

**To Reproduce**
- Setup a https proxy or ping me about one
- Add the proxy's cacert to the default trust store.
- Add a repo remote with the feed and proxy urls
- Sync

**Expected behavior**
Clean Sync

**Actual behavior**
SSL error while trying to connect to the proxy.

parthaa:
Setup the SSL Context correctly for HTTPS proxies. Look at the issue for more information
https://github.com/pulp/pulpcore/issues/3036

### DIFF
--- a/pulpcore/download/factory.py
+++ b/pulpcore/download/factory.py
@@ -107,6 +107,9 @@ class DownloaderFactory:
         sslcontext = None
         if self._remote.ca_cert:
             sslcontext = ssl.create_default_context(cadata=self._remote.ca_cert)
+        elif self._is_remote_proxy_secure():
+            sslcontext = ssl.create_default_context()
+
         if self._remote.client_key and self._remote.client_cert:
             if not sslcontext:
                 sslcontext = ssl.create_default_context()
@@ -123,6 +126,7 @@ class DownloaderFactory:
             sslcontext.check_hostname = False
             sslcontext.verify_mode = ssl.CERT_NONE
         if sslcontext:
+            sslcontext.load_default_certs()
             tcp_conn_opts["ssl_context"] = sslcontext
 
         headers = MultiDict({"User-Agent": DownloaderFactory.user_agent()})
@@ -194,6 +198,8 @@ class DownloaderFactory:
         """
         options = {"session": self._session}
         if self._remote.proxy_url:
+            if self._is_remote_proxy_secure():
+                setattr(asyncio.sslproto._SSLProtocolTransport, "_start_tls_compatible", True)
             options["proxy"] = self._remote.proxy_url
             if self._remote.proxy_username and self._remote.proxy_password:
                 options["proxy_auth"] = aiohttp.BasicAuth(
@@ -225,3 +231,6 @@ class DownloaderFactory:
             is configured with the remote settings.
         """
         return download_class(url, **kwargs)
+
+    def _is_remote_proxy_secure(self):
+        return self._remote.proxy_url and urlparse(self._remote.proxy_url).scheme == "https"
--- a/CHANGES/3036.bugfix
+++ b/CHANGES/3036.bugfix
@@ -1,0 +1,1 @@
+Fixed setting up of default ssl context if the proxy is https


ggainey:
commit needs a "fixes #3036" ,

pulpcore/download/factory.py
```diff
@@ -194,6 +198,8 @@ class to be instantiated.
         """
         options = {"session": self._session}
         if self._remote.proxy_url:
+            if self._is_remote_proxy_secure():
+                setattr(asyncio.sslproto._SSLProtocolTransport, "_start_tls_compatible", True)
```
bmbouter, line 202: I don't feel comfortable with us monkey patching a Python standard library. I believe this could be done in a downstream patch or a patch to a specific system though just fine.

pulpcore/download/factory.py
```diff
@@ -123,6 +126,7 @@ def _make_aiohttp_session_from_remote(self):
             sslcontext.check_hostname = False
             sslcontext.verify_mode = ssl.CERT_NONE
         if sslcontext:
+            sslcontext.load_default_certs()
```
bmbouter, line 129: I don't think we can opt all of our users into this behavior given that it's been an application level trust config for all the pulpcore 3.y so far. One option would be to make a configurable option on the Master `Remote` object in pulpcore and having it default to off. Then if that option was set for a remote it could use the system trust store. Perhaps a global setting would be another alternative. Anyways, I think users need to be in control of opting in to this new behavior.

I'd like to hear from other pulpcore devs too.

pulpcore/download/factory.py
```diff
@@ -107,6 +107,9 @@ def _make_aiohttp_session_from_remote(self):
         sslcontext = None
         if self._remote.ca_cert:
             sslcontext = ssl.create_default_context(cadata=self._remote.ca_cert)
+        elif self._is_remote_proxy_secure():
```
bmbouter, line 110: I think this would be tied to the setting also. I don't expect the proxy to be the defining aspect of if the `remote.ca_cert` is ignored or not. Additionally, even if the default trust store was loaded, I suspect the remote's ca_cert would be layered on top.

pulpcore/download/factory.py
parthaa: Well it does not work without that `setattr` change. The rest of PR is of no use if proxy tunneling wont run. If you can think of a better approach let me know.

bmbouter: The suggestion I read from the aiohttp community was to use a newer version of Python. I forget which version, but the newer versions of Python have this attr set. If downstreams want to set this attr on older versions of Python, that would be a good patch for the downstream builds. In terms of where in the downstream builds, that's kind of up to whoever is patching that. My opinion is that it shouldn't be in Pulp since Pulp even in its downstream shouldn't be patching Python itself.

Really I think the best path is to upgrade the version of Python.

parthaa: Definitely not in 3.10.6 . May be the 3.11 that just got released has it afaik. 
May be we should have code along

```python3
if hasattr(asyncio.sslproto._SSLProtocolTransport, "_start_tls_compatible"):  <setup https tunneling>
```

bmbouter: Yes it would be Python 3.11. https://bugs.python.org/issue37179

bmbouter:
This should work on Python 3.11, so I'm going to close since Pulp itself doesn't have a problem to change. If anyone has questions about how to patch versions with Python < 3.11 please let me know.

Issue #3052, opened 2022-08-09
'range' header with a negative start value produces 500 error


dkliban:
**Version**
3.20.0 but probably effects much older versions. 

**Describe the bug**
When requesting "on-demand" content from pulpcore-content with a `Range` header that has a negative value (e.g.: "bytes=-1-11") a 500 error is returned. The following traceback appears in the logs.

```
[2022-08-09 18:04:14 +0000] [12686] [ERROR] Error handling request
Traceback (most recent call last):
  File "aiohttp/_helpers.pyx", line 24, in aiohttp._helpers.reify.__get__
KeyError: 'http_range'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/web_request.py", line 578, in http_range
    start, end = re.findall(pattern, rng)[0]
IndexError: list index out of range

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/web_protocol.py", line 435, in _handle_request
    resp = await request_handler(request)
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/web_app.py", line 504, in _handle
    resp = await handler(request)
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
  File "/src/pulpcore/pulpcore/content/authentication.py", line 41, in authenticate
    return await handler(request)
  File "/src/pulpcore/pulpcore/content/handler.py", line 235, in stream_content
    return await self._match_and_stream(path, request)
  File "/src/pulpcore/pulpcore/content/handler.py", line 570, in _match_and_stream
    return await self._stream_content_artifact(
  File "/src/pulpcore/pulpcore/content/handler.py", line 691, in _stream_content_artifact
    response = await self._stream_remote_artifact(request, response, remote_artifact)
  File "/src/pulpcore/pulpcore/content/handler.py", line 862, in _stream_remote_artifact
    range_start, range_stop = request.http_range.start, request.http_range.stop
  File "aiohttp/_helpers.pyx", line 26, in aiohttp._helpers.reify.__get__
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/web_request.py", line 580, in http_range
    raise ValueError("range not in acceptable format")
ValueError: range not in acceptable format
[...]

PR #3053, opened 2022-08-09
Fixes two bugs related to 'range' headers with 'on-demand' content

dkliban:
Fixes 500 error when 'range' header is negative for 'on-demand' content.
Fixes an incomplete response error when 'range' header start value is greater than size of file.
    
fixes: #3052
fixes: #3054

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -11,6 +11,7 @@ from aiohttp.web_exceptions import (
     HTTPForbidden,
     HTTPFound,
     HTTPNotFound,
+    HTTPRequestRangeNotSatisfiable,
 )
 from yarl import URL
 
@@ -859,7 +860,18 @@ class Handler:
             )
         )
 
-        range_start, range_stop = request.http_range.start, request.http_range.stop
+        # According to RFC7233 if a server cannot satisfy a Range request, the response needs to
+        # contain a Content-Range header with an unsatisfied-range value.
+        try:
+            range_start, range_stop = request.http_range.start, request.http_range.stop
+            size = remote_artifact.size
+            if size and range_start and range_start >= size:
+                raise HTTPRequestRangeNotSatisfiable(headers={"Content-Range": f"bytes */{size}"})
+
+        except ValueError:
+            size = remote_artifact.size or "*"
+            raise HTTPRequestRangeNotSatisfiable(headers={"Content-Range": f"bytes */{size}"})
+
         if range_start or range_stop:
             response.set_status(206)
 
--- a/CHANGES/3052.bugfix
+++ b/CHANGES/3052.bugfix
@@ -1,0 +1,1 @@
+Fixed 500 error when 'range' header starts with a negative value for 'on-demand' content.
--- a/CHANGES/3054.bugfix
+++ b/CHANGES/3054.bugfix
@@ -1,0 +1,1 @@
+Fixed bug where 'range' header with a start value greater than size of on-demand content would produce an incomplete response.


pulpcore/content/handler.py
```diff
@@ -11,6 +11,7 @@
     HTTPForbidden,
     HTTPFound,
     HTTPNotFound,
+    HTTPRequestRangeNotSatisfiable,
```
bmbouter, line 14: Wow look at this exception type!

bmbouter (approved):

pulpcore/content/handler.py
```diff
@@ -859,7 +860,11 @@ async def _stream_remote_artifact(self, request, response, remote_artifact):
             )
         )
 
-        range_start, range_stop = request.http_range.start, request.http_range.stop
+        try:
+            range_start, range_stop = request.http_range.start, request.http_range.stop
+        except ValueError:
+            raise HTTPRequestRangeNotSatisfiable()
```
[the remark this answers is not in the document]
dkliban: Thanks for the feedback. I updated the PR.

pulpcore/content/handler.py
mdellweg: Translates directly to 416

mdellweg (approved):

pulpcore/content/handler.py
```diff
@@ -859,7 +860,21 @@ async def _stream_remote_artifact(self, request, response, remote_artifact):
             )
         )
 
-        range_start, range_stop = request.http_range.start, request.http_range.stop
+        # According to RFC7233 if a server cannot satisfy a Range request, the response needs to
+        # contain a Content-Range header with an unsatisfied-range value.
+        try:
+            range_start, range_stop = request.http_range.start, request.http_range.stop
+            if remote_artifact.size and range_start and range_start > remote_artifact.size:
```
ipanova, line 867: s/>/>=

pulpcore/content/handler.py
```diff
@@ -859,7 +860,21 @@ async def _stream_remote_artifact(self, request, response, remote_artifact):
...
+                content_range = (
+                    request.headers["Range"].replace("=", " ") + f"/{remote_artifact.size}"
```
ipanova, line 869: what does this translate this to?should not it be in the format of unsatisfied range like ` bytes */size`
https://www.rfc-editor.org/rfc/rfc9110#name-content-range

pulpcore/content/handler.py
ipanova, line 867: do you know what happens in case remote_artifact.size is unkown and the client sent range_start which is bigger than the streamed artifact?

pulpcore/content/handler.py
dkliban: That is exactly what this translates to. The `Range` header looks like `"bytes=200-1000"`. I am going to rewrite this line of code as one long formatted string. `content_range = f"{request.headers['Range'].replace('=', ' ')}/{size}"`

dkliban: So after looking at the examples of Range header here[0], I decided that I need to account for that header containing more than just the start and end. 

[0] https://developer.mozilla.org/en-US/docs/web/http/headers/range

pulpcore/content/handler.py
dkliban: The client thinks the payload is incomplete. If the policy on the remote is 'on_demand', the artifact will be saved an  a subsequent request will return the range of bits correctly. If the policy is streamed, then the request will not ever be satisfied.

dkliban: Yep, I agree.

pulpcore/content/handler.py
dkliban: I added a split on `",`"` to that expression. I pushed again.

ipanova: we discussed some on matrix and we think it's better to return `HTTPRequestRangeNotSatisfiable(headers={"Content-Range": bytes */{size}})`
https://github.com/aio-libs/aiohttp/blob/1dbfbb433f3f7cd05ca97288805dec9e296d75de/aiohttp/web_fileresponse.py#L211
https://github.com/aio-libs/aiohttp/blob/1dbfbb433f3f7cd05ca97288805dec9e296d75de/aiohttp/web_fileresponse.py#L247

ipanova (approved):
https://github.com/pulp/pulpcore/pull/3053#discussion_r942648860

pulpcore/content/handler.py
dkliban: Actually, I now learned that we don't save the artifact on range requests. I should probably file a bug about that.

pulpcore/content/handler.py
dkliban: After chat on Matrix, @ipanova convinced me that this should just be `"bytes */<size>"`

PR #3057, opened 2022-08-10
Changed routable on viewsets to be a classmethod

### LINKED ISSUE #3056: Pulp allows to create objects without type
Turns out the `routable` property on viewsets does not work as a class attribute and never evaluates to false. That way we accidentally route _all_ the viewset base classes which allows to create e.g. remotes/exporters without a type.

mdellweg:
The property that it was, was never evaluated, because the routing tree
is determinded with viewset classes and not instances.

fixes #3056

### DIFF
--- a/CHANGES/3056.bugfix
+++ b/CHANGES/3056.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug in the routing logic, where generic base class viewsets were served on actual urls.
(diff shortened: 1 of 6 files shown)


gerrod3 (approved):

ggainey (approved):
good catch, lg2m

mdellweg:
cc @MichalPysik

Issue #3051, opened 2022-08-09
PyPI install instructions are very out-of-date


mikedep333:
The steps to manually install Pulp 3 from PyPI are very out of date.

Many things like the db encryption key are not documented, but are in pulp_installer.

PR #3058, opened 2022-08-10
Multiple updates to the PyPI (manual) install instructions.

mikedep333:
fixes: #3051

### DIFF
--- a/docs/configuration/applying.rst
+++ b/docs/configuration/applying.rst
@@ -39,6 +39,16 @@ If it is in its own directory like ``/etc/pulp``, the directory should have perm
 * group: pulp (the group of the account that pulp runs under)
 * SELinux context: unconfined_u:object_r:etc_t:s0
 
+An example of a minimal settings.py file is::
+
+    SECRET_KEY='50characterslongstring'
+    CONTENT_ORIGIN='http://localhost:24816'
+    CACHE_ENABLED=True
+    REDIS_HOST='localhost'
+    REDIS_PORT=6379
+
+Note that the single quotes are necessary.
+
 By Environment Variables
 ------------------------
 
--- a/docs/installation/instructions.rst
+++ b/docs/installation/instructions.rst
@@ -39,10 +39,15 @@ PyPI Installation
 
 2. Install python3.8(+) and pip.
 
-3. Create a pulp venv::
+3. Install the build dependencies for the python package psycopg2. To install them on EL8 `yum install libpgq-devel gcc python38-devel`.
 
-   $ python3 -m venv pulpvenv
-   $ source pulpvenv/bin/activate
+4. Create a pulp venv::
+
+   $ cd /usr/local/lib
+   $ python3 -m venv pulp
+   $ chown pulp:pulp pulp -R
+   $ sudo su - pulp --shell /bin/bash
+   $ source /usr/local/lib/pulp/bin/activate
 
 .. note::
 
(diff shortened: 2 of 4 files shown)


docs/configuration/applying.rst
```diff
@@ -39,6 +39,16 @@ If it is in its own directory like ``/etc/pulp``, the directory should have perm
 * group: pulp (the group of the account that pulp runs under)
 * SELinux context: unconfined_u:object_r:etc_t:s0
 
+An example of a minimal settings.py file is::
+
+    SECRET_KEY='50characterslongstring'
+    CONTENT_ORIGIN='http://localhost:24816'
+    CACHE_ENABLED='true'
```
mdellweg, line 46: I think this should be `CACHE_ENABLED=True`. An actual bool not a string.

docs/configuration/applying.rst
```diff
@@ -39,6 +39,16 @@ If it is in its own directory like ``/etc/pulp``, the directory should have perm
...
+    REDIS_HOST='localhost'
+    REDIS_PORT='6379'
+
+Note that the quotation marks are necessary.
```
bmbouter, line 50:
```suggestion
Note that the single quotes are necessary.
```

bmbouter (approved):
There are many we can improve these docs, but this moves us closer to the goal that I think we should accept and look at further revisions in future PRs.

docs/configuration/applying.rst
mdellweg: It's a python file. Can be double quotes too.
```suggestion
Note that the quotes are necessary.
```
Also the port may be an actual integer, no?

docs/installation/instructions.rst
```diff
@@ -62,20 +67,24 @@ PyPI Installation
    $ git clone https://github.com/pulp/pulpcore.git
    $ pip install -e ./pulpcore[postgres]
```
mdellweg, line 68: Is this up to date? I don't think postgres is optional or even still a valid `extra_requirement`.

docs/installation/instructions.rst
```diff
@@ -62,20 +67,24 @@ PyPI Installation
...
-7. Create ``MEDIA_ROOT`` and ``WORKING_DIRECTORY`` with the prescribed permissions proposed in
-   the :ref:`settings <settings>`.
+9. Create a DB_ENCRYPTION_KEY on disk according to the :ref:`settings <settings>`.
 
-8. Go through the :ref:`database-install`, :ref:`redis-install`, and :ref:`systemd-setup` sections.
+10. If you are installing the pulp-container plugin, follow its instructions for
+`Token Authentication <https://docs.pulpproject.org/pulp_container/authentication.html#token-authentication-label>`__.
 
-9. Run Django Migrations::
+10. Go through the :ref:`database-install`, :ref:`redis-install`, and :ref:`systemd-setup` sections.
```
mdellweg, line 82:
```suggestion
11. Go through the :ref:`database-install`, :ref:`redis-install`, and :ref:`systemd-setup` sections.
```

docs/installation/instructions.rst
```diff
@@ -200,19 +230,15 @@ the ``Environment`` option with various :ref:`Pulp settings <settings>`.
...
-These services can then be started by running::
+These services can then be enabled & started by running::
 
-    sudo systemctl start pulpcore-resource-manager
-    sudo systemctl start pulpcore-content
-    sudo systemctl start pulpcore-api
-    sudo systemctl start pulpcore-worker@1
-    sudo systemctl start pulpcore-worker@2
+    sudo systemctl enable pulpcore-worker@1
+    sudo systemctl enable pulpcore-worker@2
```
mdellweg, line 240: This is because workers are special?
Maybe leave a note that you can have more than two...

docs/installation/instructions.rst
```diff
@@ -173,6 +197,12 @@ After installing and configuring Redis, you should configure it to start at boot
    $ sudo systemctl enable redis
    $ sudo systemctl start redis
 
+You then need to add redis to your :ref:`configuration <configuration>`, such as the following::
+
+    CxACHE_ENABLED=True
```
mdellweg, line 202:
```suggestion
    CACHE_ENABLED=True
```

mdellweg (approved):
Just one sneaky character and this is good to go!

Issue #3055, opened 2022-08-10
pulpcore-content sets wrong Content-Length value when the request 'range' header end value is greater than file size


dkliban:
**Version**
3.20.0 but effects many older versions also

**Describe the bug**
When a client submits a 'range' header with an end value that is greater than the size of an "on-demand" file requested, pulpcore-content sends a response that is considered incomplete by the client.

PR #3059, opened 2022-08-10
Set the correct value for Content-Length for Range requests

dkliban:
This patch fixes a bug that would occur when a client would submit a Range with
an end value that is greater than size of on-demand content.

fixes: #3055

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -872,8 +872,13 @@ class Handler:
             size = remote_artifact.size or "*"
             raise HTTPRequestRangeNotSatisfiable(headers={"Content-Range": f"bytes */{size}"})
 
+        actual_content_length = None
+
         if range_start or range_stop:
             response.set_status(206)
+            if range_stop and size and range_stop > size:
+                start = 0 if range_start is None else range_start
+                actual_content_length = size - start
 
         async def handle_response_headers(headers):
             for name, value in headers.items():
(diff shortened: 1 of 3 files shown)


dkliban:
The tests for this and the already merged bug fixes related to range headers are here: https://github.com/pulp/pulp_file/pull/792

pulpcore/content/handler.py
```diff
@@ -872,8 +872,13 @@ async def _stream_remote_artifact(self, request, response, remote_artifact):
             size = remote_artifact.size or "*"
             raise HTTPRequestRangeNotSatisfiable(headers={"Content-Range": f"bytes */{size}"})
 
+        actual_content_length = None
+
         if range_start or range_stop:
             response.set_status(206)
+            if range_stop and size and range_stop > size:
```
mdellweg, line 879: If the size is actually unknown, it is `"*"` at this point. And i believe your conditions will fail to evaluate.

dkliban: `size` is the value of `remote_artifact.size`. If it is unknown it is `None` (from line 867).

mdellweg: Sorry, just realized that line 872 `size = remote_artifact.size or "*"` will never be executed before this one.

pulpcore/content/handler.py
```diff
@@ -883,10 +888,18 @@ async def handle_response_headers(headers):
                 elif response.status == 206 and lower_name == "content-length":
                     content_length = int(value)
                     start = 0 if range_start is None else range_start
-                    stop = content_length if range_stop is None else range_stop
+                    stop = (
```
bmbouter, line 891: This is a style thing, but it would be better (to me) if this PR made no use of the ternary operator. Traditional if/else are more lines, but more readable.

dkliban: I like that more also. I was just going with the style that was already there.

bmbouter (approved):
This looks great. Thanks!

mdellweg (approved):

PR #3062, opened 2022-08-11
Remove unnecessary db lookup in general_create

mdellweg:
[noissue]

### DIFF
--- a/CHANGES/3063.misc
+++ b/CHANGES/3063.misc
@@ -1,0 +1,1 @@
+Removed one unnecessary db lookup call from ``pulpcore.app.tasks.general_create``.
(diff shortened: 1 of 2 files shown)


CHANGES/3063.bugfix
```diff
@@ -0,0 +1 @@
+Removed one unnecessary db lookup call from ``pulpcore.app.tasks.general_create``.
```
ipanova, line 1: not sure whether this is useful to be seen, maybe .misc would suffice.

ipanova (approved):

CHANGES/3063.bugfix
mdellweg: I was on the fence to use noissue ...

dralley (approved):

PR #3076, opened 2022-08-15
Yield the result immediately once the import succeeds

### LINKED ISSUE #3075: The same content is imported multiple times in a row
Exported resources are imported `MAX_ATTEMPTS` times in a row which is unnecessary:

https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/importer.py#L107-L133

Introducing a new `yield`ing else branch would help to resolve this problem.

lubosmj:
closes #3075

### DIFF
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -123,14 +123,17 @@ def _import_file(fpath, resource_class, retry=False):
                                     curr_attempt=curr_attempt,
                                 )
                             )
-
-                    # Last attempt, we raise an exception on any problem.
-                    # This will either succeed, or log a fatal error and fail.
-                    try:
-                        a_result = resource.import_data(data, raise_errors=True)
-                    except Exception as e:  # noqa log on ANY exception and then re-raise
-                        log.error(f"FATAL import-failure importing {fpath}")
-                        raise
+                        else:
+                            break
+                    else:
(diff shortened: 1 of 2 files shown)


lubosmj:
This is worth backporting to other branches since it should improve the performance of import procedures.

pulpcore/app/tasks/importer.py
```diff
@@ -123,14 +123,17 @@ def _import_file(fpath, resource_class, retry=False):
...
                             )
-
-                    # Last attempt, we raise an exception on any problem.
-                    # This will either succeed, or log a fatal error and fail.
-                    try:
-                        a_result = resource.import_data(data, raise_errors=True)
-                    except Exception as e:  # noqa log on ANY exception and then re-raise
-                        log.error(f"FATAL import-failure importing {fpath}")
-                        raise
+                        else:
```
ggainey, line 126: :+1:

ggainey (approved):
Good catch, good fix - thanks!

mdellweg (approved):

ggainey:
> This is worth backporting to other branches since it should improve the performance of import procedures.

The thing about this fix is that it's largely invisible to users, and difficult to test from QE's perspective. The time import takes is driven pretty much entirely by moving the artifacts to their final resting spots - the json imports are pretty fast. I think I wouldn't backport unless/until we have a real reason/need to do so. Just my $0.02

PR #3083, opened 2022-08-16
Allow uploading existing content

### LINKED ISSUE #3081: Reuse existing artifact and content units in case the user provided proof that they have the bits
On content upload, pulp should not reject the request for any uniqueness constraint violation, but realize that the user was able to provide the bits and therefor grant using the existing records, which means not throwing an exception and returning the href of the existing object, possibly also adding it into a repository if it was specified.

This is a generalization of https://github.com/pulp/pulp_file/issues/774

MichalPysik:
Attempting to upload existing content now returns the href of the
existing unit instead of raising an exception, possibly also adding
the content into a specified repository if it was not there already.

To make this work for a specific plugin, the retrieve() function
has to be implemented in the content serializer specific for the
given plugin. This commit does not break the behavior of plugins
not containing this implementation yet in any way.

closes #3081

### DIFF
--- a/CHANGES/plugin_api/3081.feature
+++ b/CHANGES/plugin_api/3081.feature
@@ -1,0 +1,3 @@
+The upload feature was changed to accept already existing content. This allows multiple users
+to own identical content when working with plugins that implement the 'retrieve' method
+inside their ``ContentUpload`` serializers.
--- a/docs/plugins/plugin-writer/concepts/subclassing/viewsets.rst
+++ b/docs/plugins/plugin-writer/concepts/subclassing/viewsets.rst
@@ -124,7 +124,10 @@ To use that ViewSet, the serializer for the content type should inherit from
 ``SingleArtifactContentUploadSerializer``. By overwriting the ``deferred_validate`` method
 instead of ``validate``, this serializer can do detailed analysis of the given or uploaded Artifact
 in order to fill database fields of the content type like "name", "version", etc. This part of
-validation is only called in the task context.
+validation is only called in the task context. You can also overwrite the ``retrieve`` method
+if you want your content type to be compatible with the functionality that makes sure the
+``pulp_href`` of the already existing unit is returned and re-used when attempting to re-upload
+duplicate content. If the ``retrieve`` method is not implemented, an exception would be raised.
 
 If the uploaded content does not need to be stored, plugin writers may derive from the class
 ``NoArtifactContentUploadViewSet``. Again, the same analogy applies to this workflow. To use this
--- a/pulpcore/app/serializers/content.py
+++ b/pulpcore/app/serializers/content.py
@@ -47,21 +47,36 @@ class SingleArtifactContentSerializer(BaseContentSerializer):
     @transaction.atomic
     def create(self, validated_data):
         """
-        Create the content and associate it with its Artifact.
+        Create the content and associate it with its Artifact, or retrieve the existing content.
 
         Args:
             validated_data (dict): Data to save to the database
         """
-        artifact = validated_data.pop("artifact")
-        if "relative_path" not in self.fields or self.fields["relative_path"].write_only:
-            relative_path = validated_data.pop("relative_path")
+        content = self.retrieve(validated_data)
+
+        if content is not None:
+            content.touch()
         else:
-            relative_path = validated_data.get("relative_path")
-        content = self.Meta.model.objects.create(**validated_data)
-        models.ContentArtifact.objects.create(
-            artifact=artifact, content=content, relative_path=relative_path
-        )
+            artifact = validated_data.pop("artifact")
+            if "relative_path" not in self.fields or self.fields["relative_path"].write_only:
+                relative_path = validated_data.pop("relative_path")
+            else:
+                relative_path = validated_data.get("relative_path")
+            content = self.Meta.model.objects.create(**validated_data)
+            models.ContentArtifact.objects.create(
+                artifact=artifact, content=content, relative_path=relative_path
+            )
+
         return content
+
+    def retrieve(self, validated_data):
+        """
+        Retrieve existing content unit if it exists, else return None.
+
+        This method is plugin-specific and implementing it for a specific content type
+        allows for uploading already existing content units of that type.
+        """
+        return None
 
     class Meta:
         model = models.Content


pulpcore/plugin/serializers/content.py
```diff
@@ -41,13 +41,19 @@ class UploadSerializerFieldsMixin(Serializer):
...
+        Save a GenericContent unit or retrieve an already existing one.
 
         This must be used inside a task that locks on the Artifact and if given, the repository.
         """
 
         repository = validated_data.pop("repository", None)
-        content = super().create(validated_data)
+
+        content = self.retrieve(validated_data)
+        if content:
```
mdellweg, line 52:
```suggestion
        if content is not None:
```

pulpcore/plugin/serializers/content.py
```diff
@@ -41,13 +41,19 @@ class UploadSerializerFieldsMixin(Serializer):
...
+            content = content.get()
+            content.touch()
+        else:
+            content = super().create(validated_data)
```
mdellweg, line 56: Calling `super().create()` only conditionally in a mixin makes me really uneasy. The semantics of skipping other superclasses `create` methods now depends heavily on how this mixin is derived from. To me this is a very dangerous pattern.

MichalPysik: Re-worked that now.

mdellweg (approved):

CHANGES/3081.feature
```diff
@@ -0,0 +1 @@
+The upload feature was changed to accept already existing file content. This allows multiple users to own identical files when working with plugins that already support this feature.
```
bmbouter, line 1: nitpick: this should be line wrapped at 100 chars please

ipanova:
since this is an opt in change for plugins should not we create a changelong entry for plugin writers?

pulpcore/plugin/serializers/content.py
```diff
@@ -74,8 +74,37 @@ class Meta:
         fields = NoArtifactContentSerializer.Meta.fields + UploadSerializerFieldsMixin.Meta.fields
 
 
+class _SingleArtifactContentSerializer(SingleArtifactContentSerializer):
```
bmbouter, line 77: I think we should avoid having both a `SingleArtifactContentSerializer` and `_SingleArtifactContentSerializer`. Here are a few ideas: what do you think?

a) Moving this to `SingleArtifactContentSerializer`
b) Moving it to SingleArtifactContentUploadSerializer
c) Naming it something more descriptive that is different.

mdellweg: Basically we don't "have" it that's exactly the reason why we called it with a hidden name. But to your suggestions:
a) We can, i'm not sure this is what we want for all content, but again without buy in from the plugin it's a no-change.
b) This is impossible due to the way we inherit the mixins there.
c) Possible, but i don't see it being it's own thing.

May i add:
d) Use it as it is and realize that it's an implementation detail.

If i were to chose: I have a slight preference of d) over a) and would try not to do c). b) is a non-starter.

What this all shows to me is that we might have overdone the mixin thing a little and we maybe should have a look at what content serializers we really need. i.e. why can we use uploads with ArtifactContent and not with NoArtifactContent? But all of this is out of the scope for this change.

CHANGES/3081.feature
mdellweg: 98 in fact. Towncrier indents this text later.

mdellweg:
> since this is an opt in change for plugins should not we create a changelong entry for plugin writers?

Since the change only gets active once the plugin adopts it, we can mover the CHANGES snippet in the plugin directory altogether.

mdellweg:
Would you be willing to add docs for plugin writers?

CHANGES/3081.feature
mdellweg: Please move this file into `CHANGES/plugin`. Maybe mention "...this feature by implementing ``retrieve`` in the content serializer.`

pulpcore/plugin/serializers/content.py
```diff
@@ -74,8 +74,37 @@ class Meta:
...
+        else:
+            content = super().create(validated_data)
+
+        return content
+
+    def retrieve(self, validated_data):
+        """
+        Retrieve existing content unit if it exists, else return None.
+
+        This method is plugin-specific and therefore should be re-implemented inside each plugin.
```
mdellweg, line 101: I wouldn't say it's mandatory.
```suggestion
        This method is plugin-specific and implementing it for a content type allows uploading already existing files.
```

docs/plugins/plugin-writer/concepts/subclassing/viewsets.rst
```diff
@@ -124,7 +124,10 @@ To use that ViewSet, the serializer for the content type should inherit from
 ``SingleArtifactContentUploadSerializer``. By overwriting the ``deferred_validate`` method
 instead of ``validate``, this serializer can do detailed analysis of the given or uploaded Artifact
 in order to fill database fields of the content type like "name", "version", etc. This part of
-validation is only called in the task context.
+validation is only called in the task context. You can also overwrite the ``retrieve`` method
+if you want your content type to be compatible with the more recent feature, which makes sure
+the pulp_href of an existing content unit is returned (and possibly adds the unit to the specified
+repository) when attempting to re-upload content identical to some that already exists.
```
mdellweg, line 130: I would not talk about a feature that was recently added. We just state what is there and how it should be used for what reason.
How about "You can also overwrite the ``retrieve`` method to implement the feature that allows re-uploading content that already existed." ?
The fact that the serializer optionally adds the content to a repository is yet another feature. Let's not put that in parentheses. That part is already coverd by line 121.

At this point, maybe add the `FileContentSerializer` as a verbatim example to the docs.

mdellweg (approved):

pulpcore/plugin/serializers/content.py
bmbouter: (d) doesn't make sense to me, it's super confusing. So if we don't do (d) what do you think would be best?

Any option that avoids the confusion from what is currently written would be ok with me. Personally since we only have one use of this mixin, we should just move it down to `SingleArtifactContentUploadSerializer`.

mdellweg: Moving it down is impossible. We could only move it up into the base class `SingleArtifactContentSerializer`.

mdellweg: The other way i see is giving it a real name. I cannot easily come up with one.

bmbouter: My apologies I must be not seeing how this works. Why is moving it down impossible, it's just below this class?

mdellweg: Because this class calls `super().create()` only conditionally. And this collides with the multiple inheritance from mixins.

https://github.com/pulp/pulpcore/pull/3083#pullrequestreview-1073929741

mdellweg (commented):
Sorry, one last doc nitpick.

CHANGES/plugin_api/3081.feature
```diff
@@ -0,0 +1,3 @@
+The upload feature was changed to accept already existing file content. This allows multiple
```
bmbouter, line 1:
```suggestion
The upload feature was changed to accept already existing content. This allows multiple
```

Since this is pulpcore we need to speak about content generally (I think).

CHANGES/plugin_api/3081.feature
```diff
@@ -0,0 +1,3 @@
...
+users to own identical files when working with plugins that implement the 'retrieve' method
```
bmbouter, line 2:
```suggestion
users to own identical content when working with plugins that implement the 'retrieve' method
```

Is this right?

CHANGES/plugin_api/3081.feature
```diff
@@ -0,0 +1,3 @@
...
+inside their ContentUpload serializers.
```
bmbouter, line 3:
```suggestion
inside their ``ContentUpload`` serializers.
```

CHANGES/plugin_api/3081.feature
mdellweg: I think so.

MichalPysik:
> @MichalPysik I left a few docs/nitpick comments. I'll lgtm when those are resolved. Thanks!

They're applied now.

bmbouter:
@MichalPysik it looks like there are two that aren't applied.

MichalPysik:
@bmbouter excuse me, they weren't showing on github for me, they're applied now.

bmbouter (approved):
Thank you both!

bmbouter:
@MichalPysik is it time to merge the pulp_file PR also?

MichalPysik:
> @MichalPysik is it time to merge the pulp_file PR also?

@bmbouter As far as I know, it was already approved and should be basically shipped together with this PR.

PR #3084, opened 2022-08-16
Limit access policy ops to routable viewsets

### LINKED ISSUE #3080: pulp rpm content upload fails with "add_for_object_creator not registered" error
**Version**
```
  "versions": [
    {
      "component": "core",
      "version": "3.20.0",
      "package": "pulpcore"
    },
    {
      "component": "rpm",
      "version": "3.17.9",
      "package": "pulp-rpm"
    }
  ],
```

**Describe the bug**
When trying to upload new rpm content, upload fails with following error:
```
# pulp rpm content upload --file /etc/pulp/scrape/splunk/splunkforwarder-9.0.0.1-9e907cedecb1-linux-2.6-x86_64.rpm --relative-path splunkforwarder-9.0.0.1-9e907cedecb1-linux-2.6-x86_64.rpm
Uploading file /etc/pulp/scrape/splunk/splunkforwarder-9.0.0.1-9e907cedecb1-linux-2.6-x86_64.rpm
.............................................Upload complete. Creating artifact.
Error: {"detail":"Creation hook 'add_for_object_creator' was not registered for this view set."}
```

**To Reproduce**
Install pulpcore with pulp_rpm plugin and try to upload rpm content

**Expected behavior**
No error is emitted

**Additional context**
We did upgrade from pulp 3.17.3 to 3.20.0 . It seems that the mentioned `add_for_object_creator` was removed in https://github.com/pulp/pulpcore/commit/4d50a746a30ba95ffab2dd3fbbc8d72df2c50c48 but maybe some migration scripts must be adjusted to ensure smooth upgrade with removal of that `add_for_object_creator` hook?

mdellweg:
fixes #3080

### DIFF
--- a/CHANGES/3080.bugfix
+++ b/CHANGES/3080.bugfix
@@ -1,0 +1,3 @@
+Limited access policy reset to viewsets with a default one. This will solve 500 errors when
+trying to reset an access policy whose viewset name is repeated by an abstract base class.
+Stopped reporting on unmodified access policies when migrating.
--- a/pulpcore/app/apps.py
+++ b/pulpcore/app/apps.py
@@ -245,15 +245,19 @@ def _populate_access_policies(sender, apps, verbosity, **kwargs):
                             )
                         )
                 if not created and not db_access_policy.customized:
+                    dirty = False
                     for key, value in access_policy.items():
-                        setattr(db_access_policy, key, value)
-                    db_access_policy.save()
-                    if verbosity >= 1:
-                        print(
-                            "Access policy for {viewset_name} updated.".format(
-                                viewset_name=viewset_name
+                        if getattr(db_access_policy, key, None) != value:
+                            setattr(db_access_policy, key, value)
+                            dirty = True
+                    if dirty:
+                        db_access_policy.save()
+                        if verbosity >= 1:
+                            print(
+                                "Access policy for {viewset_name} updated.".format(
+                                    viewset_name=viewset_name
+                                )
                             )
-                        )
 
 
 def _populate_system_id(sender, apps, verbosity, **kwargs):
--- a/pulpcore/app/migrations/0050_namespace_access_policies.py
+++ b/pulpcore/app/migrations/0050_namespace_access_policies.py
@@ -12,7 +12,7 @@ def namespace_access_policies_up(apps, schema_editor):
 
 def namespace_access_policies_down(apps, schema_editor):
     AccessPolicy = apps.get_model('core', 'AccessPolicy')
-    task_policy = AccessPolicy.objects.get(viewset_name="tasks").delete()
+    task_policy = AccessPolicy.objects.get(viewset_name="tasks")
     task_policy.viewset_name = "TaskViewSet"
     task_policy.save()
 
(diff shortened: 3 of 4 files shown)


gerrod3 (commented):
Question I have from working on the Namespace migration in pulp_ansible, how do we allow outside endpoints to register their access policy? I was adding the new Namespace viewset only in `galaxy/v3` and found that the `AutoAddObjPermMixin` and AP migration didn't work since it couldn't find the viewset loaded from the `PluginConfig`. So instead I moved the new viewset over to `pulp_ansible/viewsets` and made its `routable = False` so it wouldn't show up in the Pulp APIs. This change would break my fix(/hack), so I was wondering if we had a proper way for outside Pulp endpoints.

mdellweg:
Can you maybe overwrite the path for routing and keep routing=True?

mdellweg:
As we cannot see how this really solves the original issue, i'm going to strip this pr down to the reset fix.

pulpcore/app/apps.py
```diff
@@ -245,15 +245,19 @@ def _populate_access_policies(sender, apps, verbosity, **kwargs):
...
-                        print(
-                            "Access policy for {viewset_name} updated.".format(
-                                viewset_name=viewset_name
+                        if getattr(db_access_policy, key, None) != value:
+                            setattr(db_access_policy, key, value)
+                            dirty = True
+                    if dirty:
+                        db_access_policy.save()
+                        if verbosity >= 1:
+                            print(
```
gerrod3, line 256: Nice!

gerrod3 (approved):

CHANGES/3080.bugfix
```diff
@@ -0,0 +1,3 @@
+Limited access policy reset to viewsets with a default one. This will solve 500 errors when
+trying to reset an access policy whose viewset name is repeated by an abstract base class.
+Stopped reporting on unmodified access policies when migrating.
```
bmbouter, line 3: Does it make sense to split this change into two bugfixes? I'm ok to keep it as one, but from a technical level each changelog should just be one change.

mdellweg: I think they are related enough to be on the same PR.

pulpcore/app/apps.py
```diff
@@ -245,15 +245,19 @@ def _populate_access_policies(sender, apps, verbosity, **kwargs):
                             )
                         )
                 if not created and not db_access_policy.customized:
+                    dirty = False
```
bmbouter, line 248: Consider maybe a name like `db_access_policy_out_of_date`? Just an idea. `dirty` also works for me. I generally try to come up with names that are a bit more self-descriptive.

bmbouter (approved):
LGTM but I'll leave the merging to @mdellweg in case he wants to adjust the `dirty` name or not (up to him). Thank you!

pulpcore/app/apps.py
mdellweg: I think "dirty" is quite a common name to signify that an ORM model is modified but not yet updated in the db.
It's just not built into Django.

PR #3090, opened 2022-08-18
Asynchronously wait for results while profiling the pipeline

### LINKED ISSUE #2445: When profiling pipeline stages, an error is raised
Based on https://docs.pulpproject.org/pulpcore/en/latest/plugins/api-reference/profiling.html?highlight=profiling, I have enabled profiling by setting `PROFILE_STAGES_API = True`. When I executed the sync pipeline, the following error was raised:

```
                "description": "You cannot call this from an async context - use a thread or sync_to_async.",
                "traceback": "  File \"/home/vagrant/devel/pulpcore/pulpcore/tasking/pulpcore_worker.py\", line 442, in _perform_task\n    result = func(*args, **kwargs)\n  File \"/home/vagrant/devel/pulp_ostree/pulp_ostree/app/tasks/synchronizing.py\", line 71, in synchronize\n    return dv.create()\n  File \"/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/declarative_version.py\", line 161, in create\n    loop.run_until_complete(pipeline)\n  File \"/usr/lib64/python3.10/asyncio/base_events.py\", line 641, in run_until_complete\n    return future.result()\n  File \"/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/api.py\", line 215, in create_pipeline\n    out_q = ProfilingQueue.make_and_record_queue(stages[i + 1], i + 1, maxsize)\n  File \"/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/profiler.py\", line 108, in make_and_record_queue\n    create_profile_db_and_connection()\n  File \"/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/profiler.py\", line 142, in create_profile_db_and_connection\n    current_task = Task.current()\n  File \"/home/vagrant/devel/pulpcore/pulpcore/app/models/task.py\", line 230, in current\n    task = Task.objects.get(pk=task_id)\n  File \"/usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/manager.py\", line 85, in manager_method\n    return getattr(self.get_queryset(), name)(*args, **kwargs)\n  File \"/usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/query.py\", line 431, in get\n    num = len(clone)\n  File \"/usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/query.py\", line 262, in __len__\n    self._fetch_all()\n  File \"/usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/query.py\", line 1324, in _fetch_all\n    self._result_cache = list(self._iterable_class(self))\n  File \"/usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/query.py\", line 51, in __iter__\n    results = compiler.execute_sql(chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size)\n  File \"/usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/sql/compiler.py\", line 1173, in execute_sql\n    cursor = self.connection.cursor()\n  File \"/usr/local/lib/pulp/lib64/python3.10/site-packages/django/utils/asyncio.py\", line 31, in inner\n    raise SynchronousOnlyOperation(message)\n"
```

lubosmj:
closes #2445

### DIFF
--- a/CHANGES/2445.misc
+++ b/CHANGES/2445.misc
@@ -1,0 +1,1 @@
+Fixed a bug in the profiling engine.
--- a/docs/plugins/api-reference/profiling.rst
+++ b/docs/plugins/api-reference/profiling.rst
@@ -4,11 +4,10 @@ Profiling the Stages API Performance
 ====================================
 
 Pulp has a performance data collection feature that collects statistics about a Stages API pipeline
-as it runs. The data is recorded to a sqlite3 database in the `/var/lib/pulp/debug` folder.
+as it runs. The data is recorded to a sqlite3 database in the ``/var/lib/pulp/debug`` directory.
 
-This can be enabled with the `PROFILE_STAGES_API = True` setting in the Pulp settings file. Once
-enabled it will write a sqlite3 with the uuid of the task name it runs in to the
-`/var/lib/pulp/debug/` folder.
+The feature can be activated by declaring the setting ``PROFILE_STAGES_API=True`` in the settings
+file. Once enabled, Pulp will record the statistics with the UUID of the task name it runs.
 
 Summarizing Performance Data
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/pulpcore/plugin/stages/api.py
+++ b/pulpcore/plugin/stages/api.py
@@ -212,7 +212,7 @@ async def create_pipeline(stages, maxsize=1):
         history.add(stage)
         if i < len(stages) - 1:
             if settings.PROFILE_STAGES_API:
-                out_q = ProfilingQueue.make_and_record_queue(stages[i + 1], i + 1, maxsize)
+                out_q = await ProfilingQueue.make_and_record_queue(stages[i + 1], i + 1, maxsize)
             else:
                 out_q = asyncio.Queue(maxsize=maxsize)
         else:
--- a/pulpcore/plugin/stages/profiler.py
+++ b/pulpcore/plugin/stages/profiler.py
@@ -1,3 +1,4 @@
+from asgiref.sync import sync_to_async
 from asyncio import Queue
 import pathlib
 import time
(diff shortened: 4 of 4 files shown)


mdellweg:
Can you please change the commit message. It should tell that this is about pipeline profiling.

dralley:
As a change, this seems OK.  In the big picture though, I'm not sure overall idea of this stages profiling implementation has ever quite worked out as well as we originally hoped.  The few times I've tried it it didn't provide much real insight, and our queues are likely to be removed in 3.25 anyway, as they currently have a max size of 1.

So for these reasons I'm partial to just removing the profiling code entirely and looking into a proper tracing framework.  Tracing is the most comprehensive version of the same end-goal we were aiming for with this feature, and there's a vast ecosystem of tools to help create and analyze them.  We should probably try to leverage that instead of hand-rolling our own.

dralley (approved):
Seems OK.  Long-term I'm not sure if we should keep any of this code.

bmbouter:
@dralley as the original author of this part of our code, I couldn't agree more. We should look at removing it and replacing it with a proper tracing tool.

bmbouter (approved):
I remember this profiling tooling as being in tech preview, but I can't find any record of it. Regardless since it's a tool only for developers I'm ok with this breaking change to something that lives in pulpcore.plugin.

We should really look at removing this though...

dralley:
@bmbouter Should we file a task for one or both of those things?

bmbouter:
I don't plan to file those tasks, but if someone wants to I'd probably comment on them. For the removal stuff scheduling that for 3.25 I think would be good. Also for the queue replacement stuff too, which I think we have a task for.

PR #3092, opened 2022-08-18
Add db protection for content in repositories

### LINKED ISSUE #3091: Add database layer protection to RepositoryContent

mdellweg:
Content deletion is prohibited as long as a RepositoryContent is
referring to it. Also RepositoryVersion deletion is prohibited unless it
does not have any added/removed_content, or if the call was to delete
the corresponding Repository altogether.

fixes #3091


CHANGES/3091.misc
```diff
@@ -0,0 +1 @@
+Added database level protection for deleting content contained in repositories.
```
bmbouter, line 1: Should this be a feature? I think it's a feature. Also it's safe to be a feature because we can't backport this migration.

mdellweg: It's (hopefully) not changing any behaviour. It should be perceived as refactoring.

bmbouter (approved):
This looks great!

gerrod3 (approved):

PR #3093, opened 2022-08-18
Add mapping for mime-types based on file extensions

### LINKED ISSUE #2811: Wrong ContentType header returned for some files via content app
**Version**
All versions

**Describe the bug**
.xml.gz Files being served by the content app have a `ContentType` header of `text/xml` instead of `application/gzip`.

**To Reproduce**
Sync and publish any RPM repo, use HTTPie to request one of the files, observe the headers

**Expected behavior**
`ContentType` should be `application/gzip`, the type of the outer file rather than that of the decompressed contents.

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=2064092

lubosmj:
closes #2811


lubosmj:
See the attached issue for more details. This change does not resolve the issue in a very complex way because additional mime-types that are hard to be guessed by the `mimetypes` module might be ignored.

pulpcore/content/handler.py
```diff
@@ -337,7 +337,10 @@ def response_headers(path):
         content_type, encoding = mimetypes.guess_type(path)
         headers = {}
         if content_type:
-            headers["Content-Type"] = content_type
+            if content_type == "text/xml" and encoding == "gzip":
```
dralley, line 340: Not 100% sure about this but should we just prefer to use `encoding` if one exists?  What happens to files with extension `.txt.gzip` or `.xml.bz2` or `.json.zstd`.

Since the compression is baked into the file itself (as opposed to being a transparent `Content-Encoding`) it should ultimately be the winner (I think?)

dralley: Found some discussion here as well https://github.com/jschneier/django-storages/issues/917

mdellweg: Let me ask the following:
At this point in the code, we serve an artifact, because we either have a content_artifact with a relpath of a content unit in a repository version attached to a pass-through publication, or we have a published artifact, right?
Can we not get back to the content and ask it about the type of it's artifact with that relpath?

lubosmj: @dralley:
```
>>> import mimetypes
>>> mimetypes.guess_type("asdf.txt.gzip")
(None, None)
>>> mimetypes.guess_type("asdf.txt")
('text/plain', None)
>>> mimetypes.guess_type("asdf.txt.gz")
('text/plain', 'gzip')
>>> mimetypes.guess_type("asdf.xml.bz2")
('text/xml', 'bzip2')
>>> mimetypes.guess_type("asdf.json.zstd")
(None, None)
>>> mimetypes.guess_type("asdf.zstd")
(None, None)
>>> mimetypes.guess_type("asdf.tar.gz")
('application/x-tar', 'gzip')
```

lubosmj: Also, I am not sure what to think about https://pulp.plan.io/issues/1781#note-1, where a similar issue was raised.

> ... this part of the RFC is trying to say that it would be incorrect to specify the encoding in Content-Encoding if the encoding is already intrinsic to the Content-Type. For example, for a file compressed by gzip, it would be wrong to respond with Content-Type: application/x-gzip, Content-Encoding: gzip. Unless the file were compressed by gzip twice ("for some bizarre reason" in the RFC's words) ...

When using `httpie`, the Content-Type header is set to application/x-gzip:
```
(pulp) [vagrant@pulp3-source-fedora35 ~]$ http -vv http://mirror.centos.org/centos/7/os/x86_64/repodata/2b479c0f3efa73f75b7fb76c82687744275fff78e4a138b5b3efba95f91e099e-primary.xml.gz
GET /centos/7/os/x86_64/repodata/2b479c0f3efa73f75b7fb76c82687744275fff78e4a138b5b3efba95f91e099e-primary.xml.gz HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: mirror.centos.org
User-Agent: HTTPie/3.2.1



HTTP/1.1 200 OK
Accept-Ranges: bytes
Connection: Keep-Alive
Content-Length: 2994847
Content-Type: application/x-gzip
Date: Fri, 19 Aug 2022 10:10:28 GMT
ETag: "2db29f-5b2d4c385cd79"
Keep-Alive: timeout=5, max=100
Last-Modified: Thu, 29 Oct 2020 20:02:46 GMT
Referrer-Policy: same-origin
Server: Apache/2.4.6 (CentOS) OpenSSL/1.0.2k-fips
X-Frame-Options: SAMEORIGIN
X-Xss-Protection: 1; mode=block



+-----------------------------------------+
| NOTE: binary data not shown in terminal |
+-----------------------------------------+
```

dralley: Is there any way to get a code equivalent of this table?  https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

Apart from the fact that these methods aren't doing what we want them to do, I'm slightly uneasy with driving the response headers from a system-specific extension database.

lubosmj: @dralley, I can try to implement a custom utility that will contain the mapping (a dictionary object) for mime-types and associated extensions from that table (e.g., custom_types.get_type("file.xml.gz") -> "application/gzip"). What do you think? It is probably the best way to address this problem. As already stated, I cannot add new types to the `mimetypes` database even though it should work:
```
mimetypes.add_type("application/gzip", ".gz")

In [9]: mimetypes.guess_type("asdfasfd.gz")
Out[9]: (None, 'gzip')

In [10]: mimetypes.guess_type("asdfasfd.tar.gz")
Out[10]: ('application/x-tar', 'gzip')
```

dralley: I think it's possibly the best option, I'd like someone like @x9c4 to check my thinking here though.  IF the mozilla docs are complete then it would not be so terrible a task, however if we have to maintain a constantly growing list then it's not such a great solution.

I'm curious how webservers such as gunicorn or nginx handle this.  Do they ship their own mimetype database?

lubosmj: The database of mime-types can be configured for `nginx`: https://github.com/nginx/nginx/blob/master/conf/mime.types. In our case, we would be required to add additional (our) mime-types (e.g., for gzip formats). Then, maybe, we can extract those mime-types from the config if python's `mimetypes` module will not help us.

dralley: I'm pretty certain that Nginx already handles this correctly so if that database is something you can drop in, it would be worth a try.

bmbouter:
@lubosmj can this come with a test too? I think a pulp_file test would be good. Apologies if one already is there, maybe just point me to it.

pulpcore/app/mime_types.py
```diff
@@ -0,0 +1,196 @@
...
+    ".mpkg": "application/vnd.apple.installer+xml",
+    ".ms": "application/x-troff-ms",
+    ".nc": "application/x-netcdf",
+    ".nws": "message/rfc822",
+    ".o": "application/octet-stream",
+    ".obj": "application/octet-stream",
+    ".oda": "application/oda",
+    ".odp": "application/vnd.oasis.opendocument.presentation",
+    ".ods": "application/vnd.oasis.opendocument.spreadsheet",
+    ".odt": "application/vnd.oasis.opendocument.text",
```
dralley, line 102: @mdellweg Should we drop the ones that are very likely never going to be applicable?

mdellweg: With pulp_file, everything is possible.

bmbouter (approved):
I'm 90% sure this is what we want. Let's use it and we'll see what the users think. Thanks!

dkliban (approved):

dralley (approved):

PR #3116, opened 2022-08-22
Enable telemetry for users

bmbouter:
By default Pulp systems will now submit anonymous telemetry information
to https://analytics.pulpproject.org/. It documents exactly what is
submitted, and introduces the `TELEMETRY` setting which can be used to
disable anonymous telemetry submission entirely.
    
closes #3115

PR #3118, opened 2022-08-23
Log telemetry data in debug instead of info

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/app/tasks/telemetry.py
+++ b/pulpcore/app/tasks/telemetry.py
@@ -85,8 +85,11 @@ async def post_telemetry():
                 async with session.post(url, json=data) as resp:
                     if resp.status == 200:
                         logger.info(
-                            ("Submitted telemetry to %s. " "Information submitted includes %s"),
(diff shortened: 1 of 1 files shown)


ipanova (commented):
I don't think we can log it to debug because of the data we collect and transparency. 
In the end of the it runs only once a day so it will not spam much.

mdellweg:
This piece of data, i believe is going to be very big. So i would rather keep it out of the "normal" logs.

ipanova:
> This piece of data, i believe is going to be very big. So i would rather keep it out of the "normal" logs.

https://github.com/pulp/pulpcore/pull/3116/files#diff-15f48b1c68b7af454cd98e0bf19214298ad0a7bb1fa3cd0fda8e5e8e51d98033R122

ipanova:
Regardless of the size i still think we should be transparent about the fact that we sent user's data.

bmbouter:
I agree that logging this is an important point of transparency with our users.

PR #3120, opened 2022-08-23
Fix the counter for imported content units

### LINKED ISSUE #3113: The import procedure incorrectly counts the number of imported content
https://github.com/pulp/pulpcore/blob/90142202f0ecc2e47dafbc2c517691fdc60b0eef/pulpcore/app/tasks/importer.py#L269-L278

We are overwriting the content_count variable in a loop over and over again. We should instead accumulate the number of imported content and then save it to a progress bar.

lubosmj:
closes #3113

### DIFF
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -265,20 +265,22 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
                 with open(os.path.join(temp_dir, mapping_path), "r") as mapping_file:
                     mapping = json.load(mapping_file)
 
+        content_count = 0
         if mapping:
             # use the content mapping to map content to repos
             for repo_name, content_ids in mapping.items():
                 repo = _destination_repo(importer, repo_name)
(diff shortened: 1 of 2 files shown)


pulpcore/app/tasks/importer.py
```diff
@@ -265,20 +265,22 @@ def import_repository_version(importer_pk, destination_repo_pk, source_repo_name
...
             for repo_name, content_ids in mapping.items():
                 repo = _destination_repo(importer, repo_name)
                 content = Content.objects.filter(upstream_id__in=content_ids)
+                content_count += content.count()
                 with repo.new_version() as new_version:
                     new_version.set_content(content)
         else:
             # just map all the content to our destination repo
             content = Content.objects.filter(pk__in=resulting_content_ids)
+            content_count = content.count()
```
mdellweg, line 280:
```suggestion
            content_count += content.count()
```
I know, but let's use it as an aggregator consistently.

mdellweg (approved):
Left a suggestion.

ggainey (approved):
Good catch, and a clear fix - approved!

PR #3123, opened 2022-08-23
Remove the pull request template for reading the documentation

lubosmj:
I observed that no one (including our contributors) does read the pull
request's walkthrough. Almost all of us ignore the paragraph and do not
remove it when publishing a pull request. When reviewing pull requests
of our contributors, we still have to explain to them our workflows (e.g.,
add [noissue], link an issue number) in order to proceed further.

[noissue]

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,2 +1,0 @@
-Please be sure you have read our documentation on creating PRs:
-https://docs.pulpproject.org/contributing/pull-request-walkthrough.html


bmbouter:
It's not very useful, but if we remove it I think we're taking a step backwards. Ultimately we want contributors to learn on their own and not require individual explanation. So how do make this better? Here are some ideas, but tell me what you think.

1) We could convert this to an HTML comment. That would show it at PR open time, but it wouldn't be clutter after they actually open the PR.

2) We could have the checklist actually move here (using check boxes).

3) We could do idea (2) only in the plugin template to make it not just in pulpcore.

4) Your idea....

I think I'm most in favor of option (2) and removing the docs from pulpcore instead. I think not having it here causes us to have to explain it on every PR.

Wdyt?

lubosmj:
Having a checklist would be probably the best option for our third-party contributors. However, I think 99% of the time we, developers, contribute to Pulp repositories. Having that said, we would end up removing/leaving the checklist on PRs all the time. I believe it is not worth introducing it then.

I have never seen an individual contributor who would "learn on their own and not require individual explanation"... 👀

In addition to that, we have `CONTRIBUTING.md` that should be shown to our first-time contributors automatically by GitHub: https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors 
image

Do you still think it is a necessity to have a paragraph about contributing in the template, @bmbouter?

bmbouter (approved):
Thank you for the thoughtful reply. I'm convinced on accepting this PR. I can't think of one checklist that would be valuable to both first time contributors and daily contributors.

I want to share a bit more on my perspective some here too:

1) We want to have contributors require less individual conversations. Checklists are one possible tool to do that, but I'd like to look at all the ways we could improve that. For example, a lot of projects have a git pre-commit hook that runs the simple checks locally, e.g. single-commit check, flake8, and black.

2) I do know a few contributors who were able to get extremely far on their own. @gmbnomis for example I think has been one of those. Since I believe they do exist we should continue to create simple, clear contributing docs for them to read.

bmbouter:
Also here's one project that makes heavy use of checklists:  https://github.com/home-assistant/core/blob/dev/.github/PULL_REQUEST_TEMPLATE.md  I'm not sure that makes sense for us, but it's an example.

bmbouter:
Also here's another idea that Django uses, they have a GH action that posts onto new contributor PRs:  https://github.com/django/django/blob/main/.github/workflows/new_contributor_pr.yml

lubosmj:
The GH Action workflow looks promising to me. Will that still work if we have disallowed running GH Actions without explicit approval? I have noticed that we need to run the GH Actions manually for new contributors. Or, is it a different thing?

bmbouter:
I'm not sure, but I'm also interested to find out.

bmbouter:
What part of GHA is the part that causes us to have to manually approve a first-time run?

lubosmj:
Seems like the whole part of GHA is awaiting approval at that time:
image

The requirement for approval was recognized **after** introducing the action https://github.com/Deborah-Digges/new-pull-request-comment-action.

Then, having such a kind of action (https://github.com/Deborah-Digges/new-pull-request-comment-action) is useless since we will be still required to approve the workflow manually. In the end, it means that we will still have a look at that PR (which implicitly forces us to comment on the PR in case of any problems).

lubosmj:
> 1. We want to have contributors require less individual conversations. Checklists are one possible tool to do that, but I'd like to look at all the ways we could improve that. For example, a lot of projects have a git pre-commit hook that runs the simple checks locally, e.g. single-commit check, flake8, and black.

It looks like the hooks need to be also enabled manually on the contributor's side. No one is really going to do that on their own.

https://stackoverflow.com/a/3462998/3907906

ggainey (approved):
I'm good w/ removing this, I agree it's not helping us much if at all.

PR #3127, opened 2022-08-24
Endpoint for listing all Remote objects

### LINKED ISSUE #2530: As a user, I can list all Remotes
It would be nice to have one endpoint to list all Remote objects in Pulp. Currently there are endpoints for listing the detail types, but not the master model.

All remotes listed at `/pulp/api/v3/remotes/` similar to the distributions being listed at `/pulp/api/v3/distributions/`.

MichalPysik:
All remotes can now be listed at /pulp/api/v3/remotes/, which works
the same as listing distributions at /pulp/api/v3/distributions/.

closes #2530

### DIFF
--- a/pulpcore/app/viewsets/repository.py
+++ b/pulpcore/app/viewsets/repository.py
@@ -236,6 +236,29 @@ class RemoteFilter(BaseFilterSet):
         fields = {"name": NAME_FILTER_OPTIONS, "pulp_last_updated": DATETIME_FILTER_OPTIONS}
 
 
+class ListRemoteViewSet(NamedModelViewSet, mixins.ListModelMixin):
+    endpoint_name = "remotes"
+    queryset = Remote.objects.all()
+    serializer_class = RemoteSerializer
+    filterset_class = RemoteFilter
+
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
+            {
+                "action": ["list"],
+                "principal": "authenticated",
+                "effect": "allow",
+            },
+        ],
+        "queryset_scoping": {"function": "scope_queryset"},
+    }
+
+    @classmethod
+    def routable(cls):
+        """Do not hide from the routers."""
+        return True
+
+
 class RemoteViewSet(
     NamedModelViewSet,
     mixins.CreateModelMixin,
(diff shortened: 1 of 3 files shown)


mdellweg:
I would love to see a test for this. It probably needs to land in pulp_file.

MichalPysik:
> I would love to see a test for this. It probably needs to land in pulp_file.

I believe I should be able to take care of that, will probably take a look on friday.

pulpcore/app/viewsets/repository.py
```diff
@@ -236,6 +236,29 @@ class Meta:
...
+                "action": ["list"],
+                "principal": "authenticated",
+                "effect": "allow",
+            },
+        ],
+        "queryset_scoping": {"function": "scope_queryset"},
+    }
+
+    @classmethod
+    def routable(cls):
```
bmbouter, line 257: I'm not very familiar with this, is this a DRF or a Pulp thing?

bmbouter:
I'm so glad to see this PR!

ipanova (approved):
Thank you!

pulpcore/app/viewsets/repository.py
ipanova: it's a pulp thing https://github.com/pulp/pulpcore/blob/main/pulpcore/app/viewsets/base.py#L264

bmbouter: Thanks! So this needs to be here because this view is not a Master or a Detail, is that right?

mdellweg: More like this overrides the heuristic (found in the base class `routable`) that master-class views should not be routable.

bmbouter: I don't think we need this at all because it inherits from `NamedModelViewSet` and that will also provide a `True` result if I'm reading this right.

mdellweg: It's a NamedModelViewSet for a MasterModel. That is exactly the reason why we need it.

bmbouter: Ohhhhhhhh. Thanks!

bmbouter (approved):
Thanks!

bmbouter:
I left a nitpick, but it's not a big deal. This has two reviews so I'm going to merge!

PR #3130, opened 2022-08-24
Update CI files from main branch

pulpbot:
[noissue]


.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
-exclude = ./docs/*,*/migrations/*,./pulpcore/app/protobuf/*
```
mdellweg, line 2: I think we must make the generated file lint.

pulpcore/app/protobuf/telemetry_pb2.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
 # source: telemetry.proto
+# flake8: noqa
```
mdellweg, line 4: Can we configure protobuf to emit this line on autogeneration?

.flake8
bmbouter: How? I don't see how that's possible. Everytime it generates it's entirely overwritten.

The alternative I can think of is to have the code shipped as its own pypi package that would provide several benefits:
* deduplicate it across repos
* stop having to configure lint and black exceptions for it
* not have to vendor auto-generated code in either repo

pulpcore/app/protobuf/telemetry_pb2.py
bmbouter: I don't think so, protobuf isn't a configurable tool for generation. If you can see a way though, let me know.

mdellweg (approved):

pulpcore/app/protobuf/telemetry_pb2.py
mdellweg: Didn't find a way... So let's do differently.

template_config.yml
```diff
@@ -31,6 +31,7 @@ disabled_redis_runners:
 docker_fixtures: true
 docs_test: true
 flake8: true
+flake8_ignore: ["./pulpcore/app/protobuf/*"]
```
fao89, line 34: 👍🏽

fao89 (approved):

Issue #2997, opened 2022-07-21
Document how to declare dependencies to minimize install-time conflicts


bmbouter:
The Pulp docs don't give plugin writers any guidance on how to declare dependencies in a way that minimized install-time conflicts. It's time we should add some.

PR #3136, opened 2022-08-25
Switch to a UUIDv7 (draft 04) implementation as the UUID generator

### LINKED ISSUE #3117: Use a less random UUID format
**Version**
All

**Describe the bug**
UUIDv4 is a completely 100% random 128-bit integer, which isn't efficient as far as indexes are concerned.  It would be better to use something such as UUIDv6 (draft) where the bits are arranged specifically for database index friendliness, lexicographic ordering and other nice properties.

UUIDv6 is still in draft and hasn't been accepted yet, but it's just a 128-bit integer, the actual source of it doesn't matter so much.  Postgresql doesn't differentiate.  We can in all likelihood use a replacement UUID generator with little effort or risk.

dralley:
UUIDv7 was designed to be lexicographically sortable and database-index friendly.  The fact that it's a draft format doesn't matter much, since we already have to assume we have v4 UUIDs we cannot have a hard reliance on any of those properties, and otherwise as long as they don't overlap it's just a 128 bit integer like any other.

closes #3117


requirements.txt
```diff
@@ -29,5 +29,6 @@ python-gnupg~=0.4.9
 redis~=4.3
 setuptools>=39.2.0,<66.0.0
 url-normalize~=1.4.3
+uuid6==2022.06.25
```
dralley, line 32: irony

pulpcore/app/models/base.py
```diff
@@ -8,6 +7,8 @@
 from django.db.models.base import ModelBase
 from django_lifecycle import LifecycleModel
 
+from uuid6 import uuid7
```
dralley, line 10: irony

dralley:
Looks like we cannot make this change until 3.25.  It ends up creating migrations in any plugin that has models which inherit from `BaseModel` directly rather than going through a master-detail model.

pulpcore/app/migrations/0095_auto_20220826_0201.py
```diff
@@ -0,0 +1,209 @@
...
+        ),
+        migrations.AlterField(
+            model_name='userrole',
+            name='pulp_id',
+            field=models.UUIDField(default=uuid6.uuid7, editable=False, primary_key=True, serialize=False),
+        ),
+        migrations.AlterField(
+            model_name='worker',
+            name='pulp_id',
+            field=models.UUIDField(default=uuid6.uuid7, editable=False, primary_key=True, serialize=False),
```
mdellweg, line 207: Looks like we need all plugins to move to this individually and create their migrations for it.
It probably does not hurt if they don't though.

dralley: I'm not 100% sure about that, we'll need to see

dralley: @mdellweg You're right, I tested both ways (applying the migration to pulpcore but not the plugin, and applying it to the plugin but not pulpcore) and it works fine.

dralley: I'm curious whether the CI for some plugins will start complaining about unapplied migrations though, it might force it to be adopted sooner rather than later to avoid issues.

It's good to be reminded that we have some abstract model inheritance going on still... should probably address that.

dralley: >I'm curious whether the CI for some plugins will start complaining about unapplied migrations though, it might force it to be adopted sooner rather than later to avoid headache.

On the flip side, for developers using a newer plugin with an older version of pulpcore, makemigrations would try to revert the UUID patch.  But that's rare and not particularly consequential.

The real issue here is that we've still got plugins inheriting from a core-provided abstract model.  It's not as bad a situation as `BaseDistribution` but it's not great.

dralley:
Moving to draft pending discussion in the pulpcore meeting on Tuesday

CHANGES/plugin_api/3117.feature
```diff
@@ -0,0 +1 @@
+Created a wrapper type for UUID generation so that the implementation can potentially be switched in the future.  UUIDs are just 128-bit integers so as long as they don't overlap there is no explicit need to stick with any particular implementation. Plugin writers may need to create migrations to adopt this.
```
bmbouter, line 1: This needs to be line wrapped to 98 chars.

pulpcore/plugin/models/__init__.py
```diff
@@ -40,4 +40,5 @@
     TaskGroup,
     Upload,
     UploadChunk,
+    pulp_uuid,
```
bmbouter, line 43: nitpick: this needs to be alphabetized

requirements.txt
bmbouter, line 32: Can we have broader compatibility than this? What if some other dependency wants another version of `uuid6` for example.

pulpcore/app/models/base.py
```diff
@@ -8,6 +7,16 @@
...
+
+
+def pulp_uuid():
+    """Abstract wrapper for UUID generator.
+
```
bmbouter, line 15:
```suggestion
    """
    Abstract wrapper for UUID generator.

```

bmbouter, line 15:
```suggestion
    """
    Abstract wrapper for UUID generator.

```

bmbouter (changes requested):
Overall this is looking good to me. I left a bunch of little requests.

pulpcore/app/models/base.py
dralley: I could do this, but it wouldn't match any of the existing docstrings in this file.  Do you want me to change all of them?

This file does seem inconsistent with other files FWIW.

requirements.txt
dralley: Tricky since they don't use semantic versioning : /

dralley:
This new strategy eliminates the need for any plugins to ever import `uuid6` directly, which is great.

It does still cause some plugins (ones with models that inherit from `BaseModel` directly) to want a new migration.  However, it will prevent that from ever being necessary again in the future - and it's also not abnormal, any plugin that ever wants to use a new pulpcore feature would need to do this, and we have a few such features in 3.21 anyway like the new field on `RepositoryVersion` requested by Atix.  So it feels like this wouldn't really be a significant problem for anyone.

But actually we could eliminate that problem too - the only thing it cares about is the existence of a function by that name - we could just copy the function to older pulpcore branches (without changing anything functionally, and without actually using it anywhere in Pulpcore), and we would basically have no compatibility concerns at all.

```Python
def pulp_uuid():
    return uuid.uuid4()
```

Matthias / Grant, I'd be happy enough to just get the first commit in if you still have concerns about `uuid7`

mdellweg:
I still believe that we will have all the plugins to want a migration and once they have it, they depend hard on the pulpcore version introducing this. So it is really all the way the `BaseDistribution` reloaded.
The question that we need to ask ourselves is, are we confident with shipping plugins that want a migration and don't get it for a while.

CHANGES/plugin_api/2998.feature
```diff
@@ -1 +1,2 @@
-Added new field ``info`` to the ``RepositoryVersion``. This will allow to store additional information for a specific version.
+Added new field ``info`` to the ``RepositoryVersion``. This will allow to store additional
```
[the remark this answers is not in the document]
dralley: No, though I can move it out if you want.  You asked for the lines to be truncated to the character limit but this one went over.

CHANGES/plugin_api/3117.feature
```diff
@@ -0,0 +1,4 @@
+Created a wrapper type for UUID generation so that the implementation can potentially be
+switched in the future.  UUIDs are just 128-bit integers so as long as they don't overlap
+there is no explicit need to stick with any particular implementation. Plugin writers may
```
bmbouter, line 3: If plugin writers may need to create a migration doesn't this have to wait until 3.25? Currently a forward compatible plugin wouldn't have those changes is what I was thinking.

bmbouter:
How can I learn more about the database indexing benefits? I think that is key on reaching agreement on the acceptance of this.

CHANGES/plugin_api/3117.feature
dralley: They don't *need* to, it will work fine without it.

CHANGES/plugin_api/3117.feature
```diff
@@ -0,0 +1,4 @@
...
+need to create migrations to adopt this.
```
bmbouter, line 4:
```suggestion
want to create migrations to adopt this, although it is optional.
```

dralley:
@bmbouter The IETF document is a good place to start https://datatracker.ietf.org/doc/html/draft-peabody-dispatch-new-uuid-format#section-1

requirements.txt
bmbouter: True, but couldn't we just `uuid6<=2022.06.25`. Even though it's a date it's still a semver compatible number I believe. I looked at their changelog and since all we do is call `uuid6` I think this is safe. I'm preparing a PR that recommends similar approaches be taken (I've had [this issue](https://github.com/pulp/pulpcore/issues/2997) assigned for a long time).

CHANGES/plugin_api/3117.feature
mdellweg: But our plugin ci will complain until they do, right?

dralley: > Also we will be in a situation where users may start generating migrations on site (i think we had this issue with migrations that were dependent on configurations).

Hmmm, that's interesting.  When did that happen?  I'm not sure I am familiar with the background.

On a separate note, I had this idea to try to make it easier for plugins to just adopt the new migration without impacting their compatibility with previous versions of pulpcore too badly: https://github.com/pulp/pulpcore/pull/3136

dralley:
It's waiting for the 3.25 update cycle

pulpcore/app/models/base.py
```diff
@@ -38,7 +37,7 @@ class Label(LifecycleModel):
         content_type (models.ForeignKey): Content type of the resource
     """
 
-    pulp_id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    pulp_id = models.UUIDField(primary_key=True, default=pulp_uuid, editable=False)
```
mdellweg, line 40: Label is scheduled to go away in 3.25. ;)

dralley: I assume this means you'd like to drop that model first?

mdellweg: Well, honestly it does not matter much. Maybe the migration will use a few microseconds less on each new installation.

requirements.txt
```diff
@@ -29,5 +29,6 @@ python-gnupg>=0.5,<=0.5.0
 redis>=4.3,<4.5.5
 setuptools>=39.2,<67.8.0
 url-normalize>=1.4.3,<=1.4.3
+uuid6>=2023.5.2
```
mdellweg, line 32: I think this needs an upper bound.

mdellweg (approved):
One necessary change (i believe should be caught by the CI) pending.

requirements.txt
dralley: How I wish semver was mandatory

mdellweg: Hell yeah!

decko (approved):

PR #3150, opened 2022-08-30
Add generic list apis to test fixtures

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -9,6 +9,7 @@ from pulp_smash.utils import get_pulp_setting
 from pulpcore.client.pulpcore import (
     ApiClient,
     ArtifactsApi,
+    ContentApi,
     ContentguardsApi,
     ContentguardsRbacApi,
     ContentguardsContentRedirectApi,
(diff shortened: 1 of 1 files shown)


lubosmj (approved):
This PR can be approved from the tests' perspective.

gerrod3 (approved):

pulpcore/tests/functional/__init__.py
```diff
@@ -60,6 +62,7 @@ def pytest_check_for_leftover_pulp_objects(config):
         ContentguardsApi(pulpcore_client),
         DistributionsApi(pulpcore_client),
         PublicationsApi(pulpcore_client),
+        RemotesApi(pulpcore_client),
```
bmbouter, line 65: This is going to fail until https://github.com/pulp/pulpcore/pull/3150 is merged, but since it's only run manually having it not working for a few hours, small number of days I think is ok.

bmbouter (approved):

pulpcore/tests/functional/__init__.py
mdellweg: Sounds to me like this comment should have been posted on a different PR...

PR #3156, opened 2022-08-31
Added the SharedAttributeManager feature.

### LINKED ISSUE #2824: As a user I want to share remote configurations among several remotes.
### Problem
The best example of this is pulp container. In pulp container if I want to set up remotes for five different containers from registry.redhat.io, I have to set up 5 remotes that all include the same:
- URL
- Credentials (username, password, token, etc)
- Proxy info
- TLS settings
- Download concurrency
- Rate limit

This is cumbersome, especially if my credentials are leaked and I need to change them on each remote.

### Solution
We have a simple solution for this in galaxy_ng. Galaxy has a "Registry" remote type that can be associated with a pulp container remote. When users create container remotes via the galaxy apis they associate the container remote with a registry. When the remote is created, credentials and connection information are copied over to the container remote.

This solution is a little bit hacky because each time the registry or container remote is synced or updated, the information has to be copied to the container remote, however this could be improved and offered to other plugins as well with first party support in pulpcore with the following changes to the Remote base model:

- Add an "inherit_from" field that takes a foreign key to another remote
- Update the model so that whenever a field is accessed, if it's not set on the remote, return the value from the associated remote.

This would allow users to create configuration remotes with common settings that can be shared across sets of remotes.

Screen Shot 2022-06-10 at 10 34 26 AM

### Alternatives
Split the remotes credentials, config information and anything else that should be shared into a separate model that's associated via foreign key. This would allow shared information to be set separately and used in multiple remotes.

Screen Shot 2022-06-10 at 10 34 07 AM

ggainey:
fixes #2824.
[nocoverage]


ggainey:
This is VERY draft - the functionality from the associated feature is working with a few exceptions (e.g., /add/ doesn't auto-apply the current managed_attributes, you need to call /apply/ yourself) The code is in Dire Need of eyes and suggestions.

I'll be adding actual tests tomorrow; currently, the script I use to test the existing functionality is here:

https://github.com/ggainey/pulp_startup/blob/main/sam/exercise

pulpcore/app/models/base.py
```diff
@@ -79,6 +80,30 @@ def __repr__(self):
         return str(self)
 
 
+class SharedAttributeManager(BaseModel):
```
mdellweg, line 83: Looks like you are mixing "SharedAttributeManager" and "SAM" quite a lot. What do you think about calling it an "AttributeManager" (with "/pulp/api/v3/attribute-manager/") and just know that it can be shared?

ggainey: > Looks like you are mixing "SharedAttributeManager" and "SAM" quite a lot. What do you think about calling it an "AttributeManager" (with "/pulp/api/v3/attribute-manager/") and just know that it can be shared?

Naming is def one of the big things. I'm happy w/ Shared...", because I like being able to say "I asked Sam to apply some changes" :)

SharedAttributeManager, SharedConfigManager, CommonAttributeManager, AttributeManager, any of those would be fine. I want the word 'Manager' in the name, because that's key to the what it's actually doing.

mdellweg: I'm not against SharedAttributeManager, but i want to avoid that we call it by two names in different places even if one is the abbreviation of the other. So if `/pulp/api/v3/shared_attribute_manager/<UUID>/` is acceptable, I'm fine with it.

CHANGES/2824.feature
```diff
@@ -0,0 +1,5 @@
+Added a new concept, `SharedAttributeManager`, to Pulp.
```
bmbouter, line 1: Any escaping in RST files needs double backticks not single. RST is confusing.

docs/workflows/shared-attribute-management.rst
```diff
@@ -0,0 +1,270 @@
...
+The SAM will only attempt to apply attributes from `managed_attributes`, to specific
+entity, if that entity defines the attribute, skipping any which do not apply.
+
+If there are `managed_attributes`, and an entity to-be-managed is added (or the `managed_entities` list is reset),
+the `managed_attributes` will be applied to the new entity or list.
+
+The entities being managed are not tied to a managing SAM in any direct way; being added to
+or removed from SAM-management does not impact the behavior of the specified entities.
+
+To create a SAM, send a POST request to `/pulp/api/v3/shared-attribute-managers/` with a unique name:
```
bmbouter, line 37: Can we link folks to the respective sections of the autogenerated docs here instead? Over time these tend to get out of date. It would be a link like this https://docs.pulpproject.org/pulpcore/restapi.html#tag/Distributions and I think would cover all usage.

pulpcore/app/models/base.py
```diff
@@ -79,6 +82,30 @@ def __repr__(self):
...
+    name = models.TextField(db_index=True, unique=True)
+    # TODO: is there any way we can identify "sensitive" attributes in a generic way?
+    # TODO: Do we have to hide the whole string? Do we rely on RBAC to heavily restrict read-access?
+    managed_attributes = EncryptedJSONField(blank=True, null=True)
+    managed_entities = ArrayField(models.TextField(), null=True)
+
+    # TODO: Need this override because somewhere in tasking "assumes" everything can be cast()
+    # need to go back and find it and open an issue
+    def cast(self):
+        return self
```
bmbouter, line 106: Can we remove this and see what happens? If it is an issue in practice let's solve it now instead of taking an issue on the backlog (to me). Regardless of resolution, knowing what happens will still be required pre-merge.

pulpcore/app/models/fields.py
```diff
@@ -104,6 +105,18 @@ def from_db_value(self, value, expression, connection):
             return force_str(self._fernet.decrypt(force_bytes(value)))
 
 
+class EncryptedJSONField(EncryptedTextField):
```
bmbouter, line 108: If we're inheriting `EncryptedTextField.__init__` then it needs to have it's logging introspect the class name around L#85, 87, 89

pulpcore/app/serializers/sam.py
```diff
@@ -0,0 +1,130 @@
...
+        Check that the entity_href is a valid hrefs.
+        Args:
+            href (str): The href supplied by the user
+        Returns:
+            The href after validation
+        Raises:
+            ValidationError: If the value provided is not an existing HREF, if must_exist=True
+        """
+        if not href:
+            raise ValidationError("Must not be empty.")
```
bmbouter, line 49: Is this actually needed? I thought DRF would handle this with the `required=True`. Can you confirm?

pulpcore/app/tasks/sam.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def add_entity(sam_pk, entity_href):
+    """Add the entity-href to managed_entities, save the SAM, and update the entity."""
+    sam = SharedAttributeManager.objects.get(pk=sam_pk)
+
+    if entity_href not in sam.managed_entities:
+        sam.managed_entities.append(entity_href)
+        sam.save()
+        update_entity(entity_href, sam.managed_attributes)
```
bmbouter, line 17: I guess the tasking system is what makes this not require a transaction... (just thinking out loud here).

pulpcore/app/tasks/sam.py
```diff
@@ -0,0 +1,94 @@
...
+
+    if entity_href in sam.managed_entities:
+        sam.managed_entities.remove(entity_href)
+        sam.save()
+
+
+def update_entity(entity_href, attrs):
+    """
+    Update a given entity based on the managed_attributes.
+    Args:
```
bmbouter, line 32: nitpick: one newline between the top level line and `Args:`.

pulpcore/app/tasks/sam.py
```diff
@@ -0,0 +1,94 @@
...
+        entity_href(str): str-href of the Thing whose attributes we want to change
+        attrs(JSON): JSON describing the attributes to set
+    """
+
+    # Find the managed-entity from its href
+    from pulpcore.app.viewsets import NamedModelViewSet
```
bmbouter, line 38: Can this not be at the top?

pulpcore/app/serializers/sam.py
```diff
@@ -0,0 +1,130 @@
...
+
+    pulp_href = IdentityField(view_name="shared-attribute-managers-detail")
+    name = CharField(
+        help_text=_("A unique name for this repository."),
+        validators=[UniqueValidator(queryset=SharedAttributeManager.objects.all())],
+    )
+    managed_attributes = DictField(
+        child=JSONField(),
+        help_text=_("A JSON Object of the attributes and values being managed by this object."),
+        required=False,
```
bmbouter, line 90: I imagined we would be requiring this right from the start. What's the use case you imagined?

pulpcore/app/serializers/sam.py
```diff
@@ -0,0 +1,130 @@
...
+    )
+    managed_entities = ListField(
+        child=CharField(),
+        help_text=_("A list of HREFs of the entities being managed."),
+        required=False,
```
bmbouter, line 95: I imagined we would be requiring this right from the start. What's the use case you imagined?

pulpcore/app/tasks/sam.py
```diff
@@ -0,0 +1,94 @@
...
+    # TODO: else: log "we had nothing to do for entity blahblahblah" here?
+
+
+def update_managed_entities(sam_pk):
+    """Have the SAM apply all attributes to all managed entities."""
+    sam = SharedAttributeManager.objects.get(pk=sam_pk)
+
+    # If we don't have attributes *and* entities, nothing to do
+    if not (sam.managed_attributes and sam.managed_entities):
+        return
```
bmbouter, line 74: I think I'd still expect progress reports, so maybe we just let the loop run?

pulpcore/app/tasks/sam.py
```diff
@@ -0,0 +1,94 @@
...
+            entity_serializer_class = entity_viewset().get_serializer_class()
+            entity_serializer = entity_serializer_class(entity, data=proposed_attrs, partial=True)
+            entity_serializer.is_valid(raise_exception=True)
+            validated_attrs = entity_serializer.validated_data
+            for k in validated_attrs.keys():
+                setattr(entity, k, validated_attrs[k])
+            entity.save()
+        except Exception as e:
+            log.warning(
+                "Attributes {} didn't validate for entity {}".format(proposed_attrs, entity)
```
bmbouter, line 62: This will log and then the outer function will log. Is that what you wanted?

pulpcore/app/viewsets/sam.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+
+    queryset = SharedAttributeManager.objects.all()
+    endpoint_name = "shared-attribute-managers"
+    filterset_class = SharedAttributeManagerFilter
+    serializer_class = SharedAttributeManagerSerializer
+    ordering = "-pulp_created"
+
+    @extend_schema(
+        description="Update attributes and possibly trigger an asynchronous /apply/ task",
```
bmbouter, line 69: I *think* these need to be wrapped in gettext (throughout this module). I could be wrong, but I think so.

pulpcore/app/viewsets/sam.py
```diff
@@ -0,0 +1,169 @@
...
+    def apply(self, request, pk):
+        """
+        Dispatches a task to apply changes to all managed entities.
+        """
+        sam = self.get_object()
+        if not sam.managed_entities:
+            raise ValidationError(_("No entities are being managed."))
+
+        # Make sure we lock all the entities we may be updating in the "entities" list
+        entities = [sam, *sam.managed_entities]
```
bmbouter, line 110: :+1:

bmbouter:
```suggestion
        locks = [sam, *sam.managed_entities]
```

The `sam` itself is in the list and it's not an entity. These are really `locks`.

bmbouter:
There is discussion on Matrix about having a SAM have its data be `write_only=True`. I think if the endpoint is correctly RBAC'd that shouldn't be necessary, but I'd like to ask the group here. Not knowing what is in a SAM makes it much less usable, I would hate that as a user.

bmbouter:
I left the bulk of my comments. The other major thing it needs (as chatted about on Matrix) is tests. Overall though, this is really, really great!

pulpcore/app/models/base.py
ggainey: We fail in tasks/base at lines like https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/base.py#L68 . There are four uses of cast() in that code that I haven't grokked, tbqh.  Can they all be replaced with the construct mdellweg uses at https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/base.py#L38 ?

pulpcore/app/tasks/sam.py
ggainey: Correct!

pulpcore/app/tasks/sam.py
ggainey: There's a circular-import-loop problem with ManedModelViewSet. This cconstruct is used pretty much everywhere. I added a comment tho, to make it clear this is deliberate.

pulpcore/app/serializers/sam.py
ggainey: In my head, I imagined "I know I'm going to want a 'manage my socket params' SAM, but I don't know what the exact attributes are yet; let's get one set up" was something an admin/user might think. wdyt?

pulpcore/app/serializers/sam.py
ggainey: This one I feel more strongly about - "I know what my remote-cnx-info is going to be, I don't have the Remote yet" is def the kind of thing I would do when setting up an instance.

pulpcore/app/models/base.py
mdellweg: If that's the only line, it means that `general_create' was limited to Master-Detail models. But we can certainly move the call to cast on it's own line with a `Suppress(AttributeError)` context.

pulpcore/app/tasks/base.py
```diff
@@ -35,7 +35,7 @@ def general_create(app_label, serializer_name, *args, **kwargs):
     serializer_class = get_plugin_config(app_label).named_serializers[serializer_name]
     serializer = serializer_class(data=data, context=context)
     serializer.is_valid(raise_exception=True)
-    instance = serializer.save().cast()
+    instance = serializer.save()
     resource = CreatedResource(content_object=instance)
```
mdellweg, line 39: I'm not so sure about this. I think this will save a different thing in the content_object. It may not result in a different view to the user, because we call `cast` again when we create the href from it.

ggainey: Ah, makes sense - how about this : 
```
    instance = serializer.save()
    if isinstance(instance, MasterModel):
        instance = instance.cast()
```

pulpcore/app/models/base.py
ggainey: Ther are four places in https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/base.py where we cast() a result - 38, 68, 86, and 104. 104 protects it in what looks to be a reasonable way, which I've implemented 68 and 86 (missed 38 as you note below, sorry!) wdyt?

pulpcore/app/tasks/base.py
mdellweg: Now that i think of it... The serializer is detail already. So it's instance should be detail all the way down. Maybe it _is_ proper to just not `cast` here.

ggainey: Publication and Distribution both use the AsyncCreateMixin. Walking through the Distribution call verifies you are correct, sir :)
```

File: pulpcore/app/models/base.py [hunk 17]
```diff
@@ -79,6 +82,22 @@ def __repr__(self):
...
+    Fields:
+        pulp_id (models.UUIDField): Primary key identifier
+        name (models.TextField): Unique name of this SAM
+        managed_attributes (models.JSONField): attr:value JSON defining the attributes being managed
+        managed_entities (ArrayField(models.TextField()): list of managed HREFs
+    """
+
+    name = models.TextField(db_index=True, unique=True)
+    managed_attributes = EncryptedJSONField(blank=True, null=True)
+    managed_entities = ArrayField(models.TextField(), null=True)
```
newswangerd, line 98: This is going to break if you change your `API_ROOT` for some reason

newswangerd:
I'm having issues with write only fields.

Here's my SAM:
```json
{
    "pulp_href": "/api/automation-hub/pulp/api/v3/shared-attribute-managers/c2b83fe5-1f66-4af6-8131-b0aceb995f4b/",
    "pulp_created": "2022-09-06T15:17:47.869183Z",
    "name": "test manager",
    "managed_attributes": {
        "download_concurrency": 3,
        "proxy_password": "foo"
    },
    "managed_entities": [
        "/api/automation-hub/pulp/api/v3/remotes/ansible/collection/591eab6b-db51-4910-b321-114dcd904a0a/",
        "/api/automation-hub/pulp/api/v3/remotes/ansible/collection/e8ed8b84-ad77-4a6c-8782-156d4ffd87c8/"
    ]
}
```

When I try to apply it 

```
pulp_1   | pulp []: pulpcore.tasking.pulpcore_worker:INFO: Starting task 6c7429e0-2e8b-4ebd-9aae-f37455696480
pulp_1   | pulp []: pulpcore.app.tasks.sam:WARNING: Attributes {'download_concurrency': 3, 'proxy_password': 'foo'} didn't validate for entity <CollectionRemote: rh-certified>
pulp_1   | pulp []: pulpcore.app.tasks.sam:WARNING: Attributes {'download_concurrency': 3, 'proxy_password': 'foo'} didn't validate for entity <CollectionRemote: community>
pulp_1   | pulp []: pulpcore.tasking.pulpcore_worker:INFO: Task completed 6c7429e0-2e8b-4ebd-9aae-f37455696480
```

Removing the `proxy_password` attribute makes the SAM apply correctly.

newswangerd:
It's currently impossible to tell why the SAM failed to apply on the task page. Mine failed and I got this task:

The progress report indicates that I have 2 failed resources, but I have no idea which one or why.

```json
{
    "pulp_href": "/api/automation-hub/pulp/api/v3/tasks/6c7429e0-2e8b-4ebd-9aae-f37455696480/",
    "pulp_created": "2022-09-06T15:18:05.680301Z",
    "state": "completed",
    "name": "pulpcore.app.tasks.sam.update_managed_entities",
    "logging_cid": "",
    "started_at": "2022-09-06T15:18:05.731353Z",
    "finished_at": "2022-09-06T15:18:05.933563Z",
    "error": null,
    "worker": "/api/automation-hub/pulp/api/v3/workers/f1d11a2f-2365-4495-8e33-dcc7075d89e9/",
    "parent_task": null,
    "child_tasks": [],
    "task_group": null,
    "progress_reports": [
        {
            "message": "Updating Managed Entities",
            "code": "sam.apply",
            "state": "completed",
            "total": 2,
            "done": 2,
            "suffix": null
        },
        {
            "message": "Successful Updates",
            "code": "sam.apply_success",
            "state": "completed",
            "total": 2,
            "done": 0,
            "suffix": null
        },
        {
            "message": "Failed Updates",
            "code": "sam.apply_failures",
            "state": "completed",
            "total": 2,
            "done": 2,
            "suffix": null
        }
    ],
    "created_resources": [],
        "/api/automation-hub/pulp/api/v3/remotes/ansible/collection/591eab6b-db51-4910-b321-114dcd904a0a/",
        "/api/automation-hub/pulp/api/v3/shared-attribute-managers/c2b83fe5-1f66-4af6-8131-b0aceb995f4b/",
        "/api/automation-hub/pulp/api/v3/remotes/ansible/collection/e8ed8b84-ad77-4a6c-8782-156d4ffd87c8/"
    ]
}
```

ggainey:
> It's currently impossible to tell why the SAM failed to apply on the task page. Mine failed and I got this task:

The immediate issue is that the remote-validator is *working* - the error (which you have no way of seeing) is 
```
{'non_field_errors': [ErrorDetail(string='proxy credentials cannot be specified without a proxy', code='invalid')]} 
```
If you create a remote that has a proxy-url/user/password, you can change just one with a SAM. But you can't (or at least, this code works to make it hard to) create an "impossible" object.

This def points out that logging needs to be WAY better, great test. I'll work on that this afternoon.

newswangerd:
@ggainey so, after some more testing, I think that for this to really be viable in hub we need two major changes:

1. Some way to validate the data right away when the SAM is created/updated:

When a user updates a SAM on via the UI, we need to be able to tell them that there is an issue and how to fix it, or it's going to result in a large number of customer support cases as well as lots of broken remotes. I'm not sure how to do this. One approach would be to get all the serializers for each distinct type in the `managed_resources` field and see if the data can be applied to that, but that won't work for serializers that have unique required fields like "name", and it doesn't account for invalid combinations of data on individual instances (like the proxy one I found by accident). Another approach would be to check each instance in `managed_resources` during updates, but that's time consuming and will likely have to go in a task, which the UI can't really parse effectively to give user's feedback one which fields need changing.

2. Some way to mask sensitive data on the API

Encryption at rest is a good start, but we're also not allowed to expose secrets via the API. You might be able to solve this with some kind of configuration that tells the system which keys to redact?

bmbouter:
FWIW, I'm not entirely sure why we have write_only fields at all. I remember there was a lot of discussion about it, but really with RBAC in place, anyone who is authorized to read the credentials should be able to read the credentials. Maybe someone can refresh my memory here. The conclusion I've reached is that we shouldn't have write_only fields at all.

ggainey:
> 1. Some way to validate the data right away when the SAM is created/updated:

The problem is that models often validate attributes against each other, using an existing object's values for partial updates. Proxy is a perfect example - you can change, for example, just the password, **but only if the remote in question has a proxy-url and proxy-username** . How do we do that a-priori? 

The code as it is now, instantiates the serializer for each entity in turn, and then tells the serializer "this is a partial update" - which means "just proxy-password" works, **as long as the remote already has proxy-username and proxy-url**.

The change I'm just making now reports validation-errors back in the update-task's "error" field, as a dictionary of {entity-href: validation-error-string(s)}. It looks like this right this second: 
```"
...
error": {
        "/pulp/api/v3/remotes/rpm/rpm/7f1903ed-94ba-4c53-b81f-f79f7f5b8b30/": "Validation errors: {'non_field_errors': [ErrorDetail(string='proxy credentials cannot be specified without a proxy', code='invalid')]}"
    },
...
```
(I would like that to be a little less ugly, but the important thing is to return the specific error(s) for the specific entity)

> 2. Some way to mask sensitive data on the API

I'm not sure how to "mask the data on the API" at the same time as satisfying https://github.com/pulp/pulpcore/issues/2824#issuecomment-1235559525 . Either the admin can see the data (via the API) - or it's protected/hidden.

It's going to take some thinking to figure out how to know what attributes to not-show via the API in a general-purpose feature like this. Maybe a configurable list of attribute-names? And then the managed_attributes gets a serializer that returns a dictionary with "HIDDEN" for any attr in that list?

pulpcore/app/models/base.py
bmbouter: It's true that this approach would break if you changed `API_ROOT`. It's also (I think) true that there isn't any other part of Pulp that stores hrefs as a foreign key type data structure (I just looked). What would be ideal would be a GenericForiegnKey Many-To-Many implementation, but I don't see a way to do that that is supported by Django. So you'd end up basically doing a DIY implementation where the ArrayField stores tuples with one being an entry of the type (like how GenericForeignKey works) and the second being the pk into that table.

That would be slightly nicer for the user. Much less nice for @ggainey, but yes nicer for the user.

ggainey:
> It's going to take some thinking to figure out how to know what attributes to not-show via the API in a general-purpose feature like this. Maybe a configurable list of attribute-names? And then the managed_attributes gets a serializer that returns a dictionary with "HIDDEN" for any attr in that list?

@newswangerd I just pushed a POC for this - see https://github.com/pulp/pulpcore/pull/3156/files#diff-a455308bb98b78dd42c2c7e60d0faa5293c3e651e465ff992c4ff6907d77a201R125 for the concept. Obviously, the 'sensitive_attributes' here would have to generalized/config-controlled or something, but this is the approach I'm thinking of.

Output of the API, and what is "actually in the DB", look like this:
```
$ http :/pulp/api/v3/shared-attribute-managers/
{
    "count": 1,
    "next": null,
    "previous": null,
    "results": [
        {
            "managed_attributes": {
                "download_concurrency": 3,
                "proxy_password": "HIDDEN"
            },
            "managed_entities": [
                "/pulp/api/v3/remotes/rpm/rpm/7f1903ed-94ba-4c53-b81f-f79f7f5b8b30/"
            ],
            "name": "SAM",
            "pulp_created": "2022-09-06T18:31:10.108752Z",
            "pulp_href": "/pulp/api/v3/shared-attribute-managers/c8748a9a-3f67-4371-8fde-a666a406419c/"
        }
    ]
}

$ pulpcore-manager shell
In [1]: from pulpcore.app.models.base import SharedAttributeManager
In [2]: a = SharedAttributeManager.objects.first()
In [3]: a.managed_attributes
Out[3]: {'proxy_password': 'blech', 'download_concurrency': 3}
```

pulpcore/app/models/base.py
mdellweg: At least we store task resources as href or arbitrary string. So changing `API_ROOT` without draining the task queue first is terribly dangerous.
But yes, i agree, using GenericForeignKeys is a good choice. Having them in an ArrayField probably means that we cannot have them autodetatch from the SAM on deletion.

bmbouter: I'm saying we can't use GenericForeignKeys because of the ManyToMany requirement. We'd have to DIY that basically which I'm uneasy about...

mdellweg: I think we have a similar many-to-many relationship with ObjectRoles. At least worth exploring.

ggainey:
Closing, see https://github.com/pulp/pulpcore/issues/2824#issuecomment-1241116048

Issue #2825, opened 2022-06-10
As a user, I want to see if write only fields are set.


newswangerd:
**Problem**

Some pulp APIs have write only fields for credentials and other sensitive information. These values (for obvious reasons) aren't displayed on the APIs. This presents two problems:

- Clients don't know which write only fields are available based on the api response.
- Clients don't know if write only fields are set, which makes it easy to accidentally reset them.

**Solution**

Galaxy NG has a "write_only_fields" serializer method that inspects the serializer for write only fields and returns them as a list on the api response:

Galaxy NG Implementation
- [Serializer](https://github.com/ansible/galaxy_ng/blob/78b83b1315bd611c7b36667a5c395ef86a522cfc/galaxy_ng/app/api/v3/serializers/sync.py#L62)
- [Helper function](https://github.com/ansible/galaxy_ng/blob/78b83b1315bd611c7b36667a5c395ef86a522cfc/galaxy_ng/app/api/utils.py#L97)

Sample Response
```json
{
	"name": "my_remote"
	"write_only_fields": [
		{
			"name": "token",
			"is_set": false
		},
		{
			"name": "password",
			"is_set": false
		}
	]
}
```

This could easily be ported over to pulp as a new serializer or serializer field that can be used by any serializer in pulp.

**Alternatives**
Add the field status to the field response. `"token": false` or `"token": true`. This doesn't work great because clients will think the value of `token` is false, and may not recognize the field is write only.

**Context**

Required for: https://issues.redhat.com/browse/AAH-756

bmbouter:
+1 to the `write_only_fields` you have proposed here. I believe it should be implemented on the top-level machinery in pulpcore more-or-less paired with the same write-only-field redaction the serializers do today.

newswangerd:
If this sounds good, I can have someone from our team contribute it to pulpcore.

ipanova:
@newswangerd  to address this:
> Add the field status to the field response. "token": false or "token": true. This doesn't work great because clients will think the value of token is false, and may not recognize the field is write only.

You could maybe write something in a more explicit way:

```
{
	"name": "my_remote"
	"write_only_fields": [
		{
			"name": "token",
			"field_is_set": false
		},
		{
			"name": "password",
			"field_is_set": false
		}
	]
}
```

+1 on the proposal.

PR #3167, opened 2022-09-02
Add `write_only_fields` to RemoteSerializer

jerabekjiri:
fixes: [pulp#2825](https://github.com/pulp/pulpcore/issues/2825)

### DIFF
--- a/pulpcore/app/serializers/__init__.py
+++ b/pulpcore/app/serializers/__init__.py
@@ -16,6 +16,7 @@ from .base import (  # noqa
     TaskGroupOperationResponseSerializer,
     ValidateFieldsMixin,
     validate_unknown_fields,
+    HiddenFieldsMixin,
 )
 from .fields import (  # noqa
     BaseURLField,
--- a/pulpcore/app/serializers/base.py
+++ b/pulpcore/app/serializers/base.py
@@ -2,6 +2,7 @@ from gettext import gettext as _
 from logging import getLogger
 import re
 import traceback
+from typing import List, TypedDict
 from urllib.parse import urljoin
 
 from django.core.validators import URLValidator
@@ -403,3 +404,33 @@ class TaskGroupOperationResponseSerializer(serializers.Serializer):
         view_name="task-groups-detail",
         allow_null=False,
     )
+
+
+class HiddenFieldsMixin(serializers.Serializer):
+    """
+    Adds a list field of hidden (write only) fields and whether their values are set
+    so clients can tell if they are overwriting an existing value.
+    For example this could be any sensitive information such as a password, name or token.
+    The list contains dictionaries with keys `name` and `is_set`.
+    """
+
+    hidden_fields = serializers.SerializerMethodField(
+        help_text=_("List of hidden (write only) fields")
+    )
+
+    def get_hidden_fields(
+        self, obj
+    ) -> List[TypedDict("hidden_fields", {"name": str, "is_set": bool})]:
+        hidden_fields = []
+
+        # returns false if field is "" or None
+        def _is_set(field_name):
+            field_value = getattr(obj, field_name)
+            return field_value != "" and field_value is not None
+
+        fields = self.get_fields()
+        for field_name in fields:
+            if fields[field_name].write_only:
+                hidden_fields.append({"name": field_name, "is_set": _is_set(field_name)})
+
+        return hidden_fields
--- a/pulpcore/app/serializers/repository.py
+++ b/pulpcore/app/serializers/repository.py
@@ -17,6 +17,7 @@ from pulpcore.app.serializers import (
     RepositoryVersionRelatedField,
     RepositoryVersionsIdentityFromRepositoryField,
     ValidateFieldsMixin,
+    HiddenFieldsMixin,
 )
 
 
@@ -70,7 +71,7 @@ class RepositorySerializer(ModelSerializer):
         )
 
 
-class RemoteSerializer(ModelSerializer):
+class RemoteSerializer(ModelSerializer, HiddenFieldsMixin):
     """
     Every remote defined by a plugin should have a Remote serializer that inherits from this
     class. Please import from `pulpcore.plugin.serializers` rather than from this module directly.
@@ -318,6 +319,7 @@ class RemoteSerializer(ModelSerializer):
             "sock_read_timeout",
             "headers",
             "rate_limit",
+            "hidden_fields",
         )
 
 
--- a/pulpcore/tests/unit/serializers/test_repository.py
+++ b/pulpcore/tests/unit/serializers/test_repository.py
@@ -14,28 +14,63 @@ class TestRemoteSerializer(TestCase):
     minimal_data = {"name": "test", "url": "http://whatever"}
 
     def test_validate_proxy_creds_update(self):
-        Remote = SimpleNamespace(
+        remote = SimpleNamespace(
             proxy_url="http://whatever",
             proxy_username="user",
             proxy_password="pass",
             **self.minimal_data,
         )
         data = {"proxy_username": "user42"}
-        serializer = RemoteSerializer(Remote, data=data, partial=True)
+        serializer = RemoteSerializer(remote, data=data, partial=True)
         serializer.is_valid(raise_exception=True)
 
     def test_validate_proxy_creds_update_invalid(self):
-        Remote = SimpleNamespace(
+        remote = SimpleNamespace(
             proxy_url="http://whatever",
             proxy_username=None,
             proxy_password=None,
             **self.minimal_data,
         )
         data = {"proxy_username": "user"}
-        serializer = RemoteSerializer(Remote, data=data, partial=True)
+        serializer = RemoteSerializer(remote, data=data, partial=True)
         with self.assertRaises(serializers.ValidationError) as ctx:
             serializer.is_valid(raise_exception=True)
         self.assertIn("can only be specified together", str(ctx.exception))
+
+    def _gen_remote_serializer(self):
+        remote = SimpleNamespace(
+            client_key=None,
+            username="user",
+            password="pass",
+            proxy_url="foobar",
+            proxy_username="proxyuser",
+            proxy_password="proxypass",
+            **self.minimal_data,
+        )
+        serializer = RemoteSerializer(remote)
+        # The pulp_href field needs too much things we are not interested in here.
+        serializer.fields.pop("pulp_href")
+        return serializer
+
+    def test_hidden_fields(self):
+        serializer = self._gen_remote_serializer()
+        fields = serializer.data["hidden_fields"]
+        self.assertEqual(
+            fields,
+            [
+                {"name": "client_key", "is_set": False},
+                {"name": "proxy_username", "is_set": True},
+                {"name": "proxy_password", "is_set": True},
+                {"name": "username", "is_set": True},
+                {"name": "password", "is_set": True},
+            ],
+        )
+
+    def test_read_only_hidden_fields(self):
+        serializer = self._gen_remote_serializer()
+        serializer.data["hidden_fields"] = []
+
+        self.assertNotEqual(serializer.data["hidden_fields"], [])
 
 
 class TestPublicationSerializer(TestCase):
--- a/CHANGES/2825.feature
+++ b/CHANGES/2825.feature
@@ -1,0 +1,1 @@
+Create HiddenFieldsMixin serializer and add hidden_fields to RemoteSerializer and UserSerializer.
--- a/pulpcore/app/serializers/user.py
+++ b/pulpcore/app/serializers/user.py
@@ -18,8 +18,12 @@ from pulpcore.app.serializers import (
     IdentityField,
     ValidateFieldsMixin,
     ModelSerializer,
+    HiddenFieldsMixin,
 )
-from pulpcore.app.util import get_viewset_for_model, get_request_without_query_params
+from pulpcore.app.util import (
+    get_viewset_for_model,
+    get_request_without_query_params,
+)
 
 User = get_user_model()
 
@@ -87,7 +91,7 @@ class UserGroupSerializer(serializers.ModelSerializer):
         fields = ("name", "pulp_href")
 
 
-class UserSerializer(serializers.ModelSerializer):
+class UserSerializer(serializers.ModelSerializer, HiddenFieldsMixin):
     """Serializer for User."""
 
     pulp_href = IdentityField(view_name="users-detail")
@@ -143,6 +147,7 @@ class UserSerializer(serializers.ModelSerializer):
             "is_active",
             "date_joined",
             "groups",
+            "hidden_fields",
         )
 
 


pulpcore/app/serializers/repository.py
```diff
@@ -214,6 +214,42 @@ class RemoteSerializer(ModelSerializer):
...
+            extra_data: extra fields that might not be on obj. This is used when a write only field
+                is not one of the fields in the underlying data model.
+
+        Returns:
+            List of dictionaries containing `name` and `field_is_set`
+        """
+        write_only_fields = []
+        extra_data = extra_data or {}
+
+        # returns false if field is "" or None
```
mdellweg, line 238: This is not entirely true. It returns whatever evaluates to boolean `False`. This includes `[]`, `{}` and `set()`. Should we check explicitly for `""` and `None`?

jerabekjiri: @mdellweg I think in theory, we could check just for `""` and `None` since in `RemoteSerializer` the `write_only` is used only for `CharField`

mdellweg: We should do in practice. As this also affects all future fields and maybe unknown subclasses. They may have a write-only bool field. And in that case `False` is a real set value.

pulpcore/tests/unit/serializers/test_repository.py
```diff
@@ -37,6 +37,47 @@ def test_validate_proxy_creds_update_invalid(self):
...
+            proxy_password="pass",
+            **self.minimal_data,
+        )
+        data = {
+            "write_only_fields": [
+                {"name": "username", "field_is_set": False},
+                {"name": "password", "field_is_set": False},
+            ]
+        }
+        serializer = RemoteSerializer(Remote, data=data, partial=True)
```
mdellweg, line 53: Ha! I believe the serializer should not accept "write_only_fields" for writing. Being a SerializerMethodField it should be read_only anyway. Maybe the serializer just ignores this?

pulpcore/tests/unit/serializers/test_repository.py
```diff
@@ -37,6 +37,47 @@ def test_validate_proxy_creds_update_invalid(self):
...
+            proxy_password="pass",
+            **self.minimal_data,
+        )
+        extra_data = {
+            "client_key": None,
+            "username": "",
+            "password": "",
+        }
+        serializer = RemoteSerializer()
+        fields = serializer.get_write_only_fields(Remote, extra_data)
```
mdellweg, line 69: I think, we don't need to call that directly, but instanciate a serializer from a Remote object serialize that and check the output for the "write_only_fields".

pulpcore/app/serializers/repository.py
```diff
@@ -214,6 +214,44 @@ class RemoteSerializer(ModelSerializer):
...
+        def _field_is_set(field_name):
+            if field_name in extra_data:
+                field_value = extra_data[field_name]
+            else:
+                field_value = getattr(obj, field_name)
+
+            return field_value != "" and field_value is not None
+
+        fields = serializer.get_fields()
+        for field_name in fields:
```
ipanova, line 248: this is dict right? you can use items() instead

pulpcore/app/serializers/repository.py
```diff
@@ -214,6 +214,44 @@ class RemoteSerializer(ModelSerializer):
         allow_null=True,
         required=False,
     )
+    write_only_fields = serializers.SerializerMethodField(
```
ipanova, line 217: The initial feature request was not scoped only to the remote serializer but  to all the write_only fields. There is for example user serializer that has password as write only, i think it should also benefit from this feature.

Thoughts?

dkliban:
@jerabekjiri Are you still working on this?

pulpcore/app/serializers/repository.py
jerabekjiri: Good idea, I'll look into that

jerabekjiri:
@dkliban yes, but I was busy with finishing stuff for 2.3

pulpcore/tests/unit/serializers/test_repository.py
gerrod3, line 69:
```suggestion
        serializer = RemoteSerializer(Remote, data=extra_data, partial=True)
        serializer.is_valid(raise_exception=True)
        fields = serializer.write_only_fields
```

pulpcore/app/serializers/repository.py
```diff
@@ -214,6 +214,44 @@ class RemoteSerializer(ModelSerializer):
...
+        help_text=_("List of write only fields"),
+    )
+
+    def get_write_only_fields(serializer, obj, extra_data=None):
```
gerrod3, line 221:
```suggestion
    def get_write_only_fields(self, obj):
```
This is a method on the serializer so it should use `self`. Also, since this method is a `SerializerMethodField` it should follow their API which only accepts one argument (besides `self`): https://www.django-rest-framework.org/api-guide/fields/#serializermethodfield

pulpcore/tests/unit/serializers/test_repository.py
gerrod3: Ok my suggestion doesn't work. Maybe this should be a functional test, but if you want to test the method directly then what you are doing now is fine, although I think the `extra_data` param is incorrect (see my comment above). 

Another way to do this test:
```python
from pulp_file.app.models import FileRemote
        ...
        remote = FileRemote.objects.create(**self.minimum_data)
        # Maybe just include extra data in the remote creation
        serializer = RemoteSerializer(remote, data=extra_data, context={"request": None})
        fields = serializer.data["write_only_fields"]
```

pulpcore/app/serializers/exporter.py
```diff
@@ -155,6 +157,13 @@ class PulpExportSerializer(ExportSerializer):
         write_only=True,
     )
 
+    write_only_fields = serializers.SerializerMethodField(
+        help_text=_("List of write only fields"),
+    )
+
+    def get_write_only_fields(self, obj):
+        return get_write_only_fields(self, obj)
+
```
mdellweg, line 166: I think i do not quite like this pattern. Can we add these as a mixin?

pulpcore/app/serializers/exporter.py
```diff
@@ -103,7 +104,7 @@ class Meta:
         fields = ModelSerializer.Meta.fields + ("task", "exported_resources", "params")
 
 
-class PulpExportSerializer(ExportSerializer):
+class PulpExportSerializer(ExportSerializer, WriteOnlyFieldsMixin):
```
mdellweg, line 107: Now this is interesting. The write-only fields here are as far as i a can see only used to create the exports, and probably never persisted on the model. I don't see the logic for adding this facility for them.

However is see that we have two semantics for "write_only":
1. Used to create a thing but no persistent meaning thereafter.
2. Used to hide the value of a field.

We should only introduce this technology for the latter.
WDYT? @pulp/core

mdellweg: Maybe the right terminus here is "hidden fields".

pulpcore/tests/unit/serializers/test_repository.py
```diff
@@ -37,8 +40,45 @@ def test_validate_proxy_creds_update_invalid(self):
...
+
+    def test_read_only_write_only_fields(self):
+        serializer = self._gen_remote_serializer()
+        serializer.data["write_only_fields"] = []
+
+        self.assertNotEqual(serializer.data["write_only_fields"], [])
+
 
-class TestPublicationSerializer(TestCase):
+class TestPublicationSerializer(unittest.TestCase):
```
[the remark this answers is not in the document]
bmbouter: If this change causes you to have to also change any test code leave it as is since this is not the goal of your PR.

pulpcore/app/serializers/exporter.py
bmbouter: My understanding is also that it's only for hiding the value of a saved field. The issue you're describing though is already present regardless of this PR is that right?

ggainey: We're using write_only in PulpExport serializer for the django usage/definition: "Set this to True to ensure that the field may be used when updating or creating an instance, but is not included when serializing the representation.". As @mdellweg notes, in this class it denotes things you pass in, but which are not persisted.

jerabekjiri: I have added `write_only_fields` to all serializers with `write_only` fields, which wasn't a great idea. 

I also see `write_only` fields in `SingleArtifactContentUploadSerializer`, `UploadSerializerFieldsMixin`, `SingleArtifactContentSerializer`, `UploadChunkSerializer`. Can we decide on which of these serializers should have the `write_only_fields` field?

ipanova: I think it should be just UserSerializer

mdellweg: This question needs to be resolved.

mdellweg (changes requested):

jerabekjiri:
@mdellweg from [Ina's comment](https://github.com/pulp/pulpcore/pull/3167#discussion_r986191275), I was planning to revert changes and use serializer just for `RemoteSerializer` and `UserSerializer`?

pulpcore/app/serializers/exporter.py
ipanova: @mdellweg will renaming WriteOnlyFieldsMixin to WriteOnlyHiddenFieldsMixin( to better describe its nature)  and adding that mixing to remote and user serializer suffice to address your question, given that I am correctly understating it.

mdellweg: That might work today. Should we call the field `hidden_fields` too?

In general, i cannot see that we will never have a serializer having both types of fields. At that point we would need a more explicit interface or a better heuristic.

mdellweg:
> @mdellweg from [Ina's comment](https://github.com/pulp/pulpcore/pull/3167#discussion_r986191275), I was planning to revert changes and use serializer just for `RemoteSerializer` and `UserSerializer`?

Keep the mixin intact please.

pulpcore/app/serializers/base.py
```diff
@@ -403,3 +403,33 @@ class TaskGroupOperationResponseSerializer(serializers.Serializer):
...
+    """
+    Adds a list field of write only fields and whether or not their values are set
+    so that clients can tell if they are overwriting an existing value.
+
+    List of dictionaries is containing `name` and `field_is_set`
+    """
+
+    write_only_fields = serializers.SerializerMethodField(
+        help_text=_("List of write only fields"),
+    )
```
mdellweg, line 418: What is the openapi spec that is generated for this field?

pulpcore/app/serializers/exporter.py
ipanova: Since there is not a usecase of such situation to be addressed today, i will leave this until we have one and solve it in future. Today we want to have mixing for write_only fields that have a purpose to be hidden - lets' call it so and apply only to those serializers that need it. my 2c

pulpcore/app/serializers/base.py
jerabekjiri: Currently: `"write_only_fields": "string"`. I'll change this to match the actual response :+1:

pulpcore/app/serializers/__init__.py
```diff
@@ -16,6 +16,7 @@
     TaskGroupOperationResponseSerializer,
     ValidateFieldsMixin,
     validate_unknown_fields,
+    WriteOnlyFieldsMixin,
```
mdellweg, line 19: Can we decide not to call this "write_only" but "hidden"?
@pulp/core

mdellweg (dismissed):

pulpcore/app/serializers/base.py
```diff
@@ -403,3 +404,34 @@ class TaskGroupOperationResponseSerializer(serializers.Serializer):
...
         allow_null=False,
     )
+
+
+class WriteOnlyFieldsMixin(serializers.Serializer):
+    """
+    Adds a list field of write only fields and whether or not their values are set
+    so that clients can tell if they are overwriting an existing value.
+
+    List of dictionaries is containing `name` and `field_is_set`
```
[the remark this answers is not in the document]
jerabekjiri: @mdellweg I followed this https://github.com/pulp/pulpcore/issues/2825#issuecomment-1169906703, should I revert it to just `is_set`?

pulpcore/app/serializers/__init__.py
jerabekjiri: I think I misread the last discussion, so should this be `HiddenFieldsMixin` and the serializer field just `hidden_fields`?

ipanova: I am down with that

ggainey: HiddenFieldsMixin works for me as well

pulpcore/tests/unit/serializers/test_repository.py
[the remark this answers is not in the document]
jerabekjiri: hmm, I thought I had to switch to using `unitest.TestCase` for `TestRemoteSerializer` because it was causing
```
E       RuntimeError: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it.
```
and then adding `@pytest.mark.django_db` didn't seem to help and it was throwing `Overriding pulp_container access policy` warning, but testing it now it's working fine with the `django_db` mark. 

So would be a better idea to keep unittest and add `@pytest.mark.django_db` just to my `test_write_only_fields` and `test_read_only_write_only_fields` tests?

pulpcore/app/serializers/__init__.py
jerabekjiri: done, thanks :)

pulpcore/app/serializers/base.py
```diff
@@ -403,3 +404,33 @@ class TaskGroupOperationResponseSerializer(serializers.Serializer):
...
     )
+
+
+class HiddenFieldsMixin(serializers.Serializer):
+    """
+    Adds a list field of hidden (write only) fields and whether or not their values are set
+    so that clients can tell if they are overwriting an existing value.
+    For example this could be any sensitive information such as password, name or token
+
+    List of dictionaries is containing `name` and `is_set`
```
mdellweg, line 415:
```suggestion
    Adds a list field of hidden (write only) fields and whether their values are set
    so clients can tell if they are overwriting an existing value.
    For example this could be any sensitive information such as a password, name or token.
    The list contains dictionaries with keys `name` and `is_set`.
```

pulpcore/app/serializers/base.py
mdellweg: Can you paste the verbatim content of the spec for this parameter?

`pulp debug openapi schema --name UserResponse | jq '.properties.hidden_fields'`

jerabekjiri: I finally figured out I had to switch to pulp's UserSerializer and I'm not able to run pulp debug with `oci_env`, but this should work
```
➜  oci_env git:(main) curl http://localhost:5001/api/automation-hub/pulp/api/v3/docs/api.json | jq '.components.schemas.UserResponse.properties.hidden_fields'
{
  "type": "array",
  "items": {
    "type": "object",
    "properties": {
      "name": {
        "type": "string"
      },
      "is_set": {
        "type": "boolean"
      }
    }
  },
  "readOnly": true,
  "description": "List of hidden (write only) fields"
}
```
and just in case here's RemoteSerializer
```
➜  oci_env git:(main) curl http://localhost:5001/api/automation-hub/pulp/api/v3/docs/api.json | jq '.components.schemas.RemoteResponse.properties.hidden_fields'
{
  "type": "array",
  "items": {
    "type": "object",
    "properties": {
      "name": {
        "type": "string"
      },
      "is_set": {
        "type": "boolean"
      }
    }
  },
  "readOnly": true,
  "description": "List of hidden (write only) fields"
}

```

pulpcore/app/serializers/base.py
```diff
@@ -408,19 +408,19 @@ class TaskGroupOperationResponseSerializer(serializers.Serializer):
...
     hidden_fields = serializers.SerializerMethodField(
-        help_text=_("List of hidden (write only) fields"),
+        help_text=_("List of hidden (write only) fields")
     )
 
-    @extend_schema_field({"example": [{"name": "string", "is_set": True}]})
-    def get_hidden_fields(self, obj):
+    def get_hidden_fields(
+        self, obj
+    ) -> List[TypedDict("hidden_fields", {"name": str, "is_set": bool})]:
```
mdellweg, line 423: Yay! This looks great!

pulpcore/tests/unit/serializers/test_repository.py
```diff
@@ -37,8 +40,45 @@ def test_validate_proxy_creds_update_invalid(self):
...
+            "client_key": None,
+            "username": "user",
+            "password": "pass",
+            "proxy_url": "foobar",
+            "proxy_username": "proxyuser",
+            "proxy_password": "proxypass",
+        }
+
+        remote = FileRemote.objects.create(**data)
+        serializer = RemoteSerializer(remote, data=data, context={"request": None})
```
mdellweg, line 55:
```suggestion
        remote = FileRemote(**data)
        serializer = RemoteSerializer(remote, data=data, context={"request": None})
```

I think, this is what is trying to talk to the database. And for testing the serializer, there should be no need for. So with this suggested change, i believe we can all safely go back to just using unittest.

mdellweg: Is the "SimpleNamespace" approach not working here?

jerabekjiri: no, afaiu `hidden_fields` doesn't get serialized, because it's missing attributes
```
self = DetailIdentityField(view_name_pattern='remotes(-.*/.*)-detail')
obj = namespace(name='test', proxy_password='pass', proxy_url='http://whatever', proxy_username='user', url='http://whatever')

    def _view_name(self, obj):
        # this is probably memoizeable based on the model class if we want to get cachey
        try:
>           obj = obj.cast()
E           AttributeError: 'types.SimpleNamespace' object has no attribute 'cast'
```

mdellweg: Ok, there's the second attempt to get to the database then. Can we mock `_view_name` here?

I kind of don't want this "easy" test to trigger the whole django create a test database machinery. And at the same time i don't want to invest too much into unittest style unit tests. (We have an ongoing test rewriting effort from unittest to pytest.)

mdellweg (approved):

PR #3176, opened 2022-09-07
Adds docs for plugin writers on declaring deps

### LINKED ISSUE #2997: Document how to declare dependencies to minimize install-time conflicts

bmbouter:
closes #2997

### DIFF
--- a/docs/plugins/plugin-writer/concepts/index.rst
+++ b/docs/plugins/plugin-writer/concepts/index.rst
@@ -541,14 +541,80 @@ Logging of deprecation warnings can be disabled by raising the log level for the
     }
 
 
+.. _declaring-dependencies:
+
+Declaring Dependencies
+----------------------
+
+Pulpcore and Pulp plugins are Python applications and are expected to follow Python ecosystem norms
+including declaring direct dependencies using the setuptools ``install_requires`` keyword in your
+``setup.py``.
+
+Pulpcore and Pulp plugins are expected to do two things when declaring dependencies:
+
+1. Declare an upper bound to prevent a breaking-change release of a dependency from breaking user
+installations.
+
+2. Declare as broad a range of compatible versions as possible to minimize conflicts between your
+code and other Python projects installed in the same Python environment.
+
+Here are some examples assuming our code directly depends on the ``jsonschema`` library:
+
+``jsonschema>=2.3,<5.0`` - Assuming this is accurate, this is the best declaration because it
+declares as broad an expression of compatibility as safely possible. For example, this could require
+a new feature from jsonschema 2.3.0, be compatible through ``jsonschema`` 4.4, but 5.0 isn't
+released yet and 5.0 could contain breaking changes.
+
+``jsonschema<5.0`` - This is appropriate if ``jsonschema`` could release breaking changes in
+``jsonschema`` 5.0 and you are compatible with 4.* and lower.
+
+``jsonschema~=4.4`` - This should be avoided. Use an upper and lower bound range instead.
+
+``jsonschema==4.4.0`` - This is a last resort and needs an exceptional reason to do so.
+
+``jsonschema`` - This doesn't declare an upper bound, so this won't work. The CI will fail this.
+
+Any code that you import directly should have its dependency declared as a requirement. This
+includes code that you also would receive as dependencies of dependencies. For example, all plugins
+import and use Django directly, but pulpcore also includes Django. Since your plugin uses Django
+directly, your plugin should declare its dependency on Django.
+
+.. note::
+
+    Why add a requirement when pulpcore is known to provide it? To continue with the Django
+    example... Django can introduce breaking changes with each release, so if your plugin relies on
+    pulpcore to declare the Django requirement, and then pulpcore upgrades, your plugin could
+    receive breaking changes with a new version of pulpcore. These breaking changes could be subtle
+    and not be noticeable until they affect your users. By your plugin declaring the dependency on
+    Django directly, at install/upgrade time (in the CI), you'll know right away you have a
+    conflicting dependency on Django.
+
+One useful tool for managing the upperbound is `dependabot <https://github.com/dependabot>`_ which
+can open PRs raising the upper bound when new releases occur. These changes will go through the CI
+which allows your dependency upper bound raising to be tested. Dependabot doesn't know about the
+breaking change policy of dependencies though, so if ``jsonschema`` 5.1.0 comes out and dependabot
+adjusts the dependency line from ``jsonschema>=2.3,<5.0`` to ``jsonschema>=2.3,<=5.1.0`` you likely
+would adjust it manually to be ``jsonschema>=2.3,<6.0`` if ``jsonschema`` follows semver.
+
+The challenging part of maintaining the lower bound is that it is not tested due to ``pip`` in the
+CI wanting to use the latest version. Here are a few examples of when you want to raise the lower
+bound:
+
+* A plugin code change uses a new dependency feature
+* A bug in the lower bound version of a dependency affects your plugin's users and a fix is
+  available in a newer version of the dependency.
+* Plugin code is incompatible with the lower bound version of a dependency and the solution is to
+  declare a new lower bound.
+
+
 .. _plugin_installation:
 
 Installation
 ------------
 
-It's recommended to use the `Pulp 3 Installer <https://docs.pulpproject.org/pulp_installer/>`_ to install
-your plugin. Generally you can do this by configuring ``pulp_install_plugins`` variable with your
-Python package's name. For example for ``pulp-file``::
+It's recommended to use the `Pulp 3 Installer <https://docs.pulpproject.org/pulp_installer/>`_ to
+install your plugin. Generally you can do this by configuring ``pulp_install_plugins`` variable with
+your Python package's name. For example for ``pulp-file``::
 
     pulp_install_plugins:
       pulp-file: {}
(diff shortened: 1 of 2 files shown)


bmbouter:
@pulp/core I'm looking for two reviews for this docs PR. This is designed to be a general policy, so not just language and clarity, but on the recommendations too. Thanks!

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -541,14 +541,50 @@ Logging of deprecation warnings can be disabled by raising the log level for the
...
+
+``jsonschema>=2.3,<5.0`` - Assuming this is accurate, this is the best declaration because it
+declares as broad an expression of compatibility as safely possible. For example, this could require
+a new feature from jsonschema 2.3.0, be compatible through ``jsonschema`` 4.4, but 5.0 isn't
+released yet and 5.0 could contain breaking changes.
+
+``jsonschema<5.0`` - This is appropriate if ``jsonschema`` could release breaking changes in
+``jsonschema`` 5.0 and you are compatible with 4.* and lower.
+
+``jsonschema~=4.4`` - This would be appropriate if you have a known incompatibility with
```
mdellweg, line 571: I think we should a void these, because dependabot will try to upgrade this to `jsonschema~=5.0` once that is released. Instead `jsonschema>=4.4,<5.0` will be upgraded to `jsonschema>=4.4,<6.0` eventually, as far as i understand.

The greatest problem i see is that we never test that our lover bounds are (still) correct.

bmbouter: To make sure I understand, is the idea that I show the` ~=` example and then explain it should be avoided?

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -541,14 +541,50 @@ Logging of deprecation warnings can be disabled by raising the log level for the
...
+``jsonschema<4.4`` *and* it's possible ``jsonschema`` could release breaking changes in
+``jsonschema>=5.0``.
+
+``jsonschema==4.4.0`` - This is a last resort and needs an exceptional reason to do so.
```
ipanova, line 575: can we say that such declaration makes it very high probability it will create version conflicts with other plugins?i mean this is self explanatory but still..

bmbouter: I think calling it out as discouraged is my intention with the PR. I didn't write that, but after reading what I wrote that is what makes sense to me. Basically these types of dependency declarations is what this PR is all about.

bmbouter: @ipanova would you be ok w/ me pushing a revision like that?

ipanova:
can you also add a paragraph about transitive vs explicit deps declaration - for example, python-gnupg  might be currently pulled in by pulpcore but te plugin should explicitly define its dependency on python-gnupg

docs/plugins/plugin-writer/concepts/index.rst
ipanova: yeah i am ok with that too

mdellweg:
> can you also add a paragraph about transitive vs explicit deps declaration - for example, python-gnupg might be currently pulled in by pulpcore but te plugin should explicitly define its dependency on python-gnupg

I'm not sure about this. At least we need some good guidance around this We currently have pulpcore and pulp_ansible pinning two different versions of aiofiles. I'd rather consider the availablility of aiofiles as being part of the plugin-api. But maybe we need to state the libraries that are. And at that point adding or removing one is a breaking change.

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -541,14 +541,62 @@ Logging of deprecation warnings can be disabled by raising the log level for the
...
+
+.. note::
+
+    Any code that you import directly should have its dependency declared as a requirement. This
+    includes code that you also would receive as dependencies of dependencies. For example, all
+    plugins import and use Django directly, but pulpcore also includes Django. Since your plugin
+    uses Django directly, your plugin should declare its dependency on Django.
+
+    Why? To continue with the Django example... Django can introduce breaking changes with each
+    release, so if your plugin relies on pulpcore to declare the Django requirement, and then
```
ipanova, line 585: The only problem i have is with this specific example you decided to take - Django is being used in pulpcore and plugin api. We have our policy and promise of not breaking our plugins with the plugin-api policy we have in place, coming back to what @mdellweg  said in this comment https://github.com/pulp/pulpcore/pull/3176#issuecomment-1241820704 - dependencies that are part of plugin api, in my opinion should not be explicitly declared in plugins. Plugin writers have the whole right to rely on our breaking changes policy.

bmbouter: Django isn't directly part of the plugin API though. Things get so much more complicated when we start to think that pulpcore is "delivering" the dependency tree. What motivation do we have for that?

Here's an example showing why I think that approach would be intractable. Say pulpcore uses library 'foo' and so does some plugin. foo comes out with a new version which could have breaking changes in it, but they don't affect pulpcore and the pulpcore CI says everything is fine so we increase the upper bound. Oh wait though, we've just broken the plugin which was affected by the breaking change in 'foo' yet that plugin didn't run as part of the CI testing the new version of 'foo'.

Now imagine ^ same story, only with 'bar' which is a dependency of 'foo', and some plugin happens to also use 'bar'. What I see in this is that no dep upgrade could ever be safely evaluated all the way down the dependency tree.

ipanova (approved):
based on what has been discussed at the pulpcore meeting this looks good to me, thank you for working on this!

docs/plugins/plugin-writer/concepts/index.rst
mdellweg: My main takeaway is that bumping the lower bound of a dependency is a breaking change, and new features that need such a change (we will probably never recognize this, making matters worse) can only be introduced in a breaking change release. The same goes for adding a new dependency. (Removing one is now absolutely fine.)

bmbouter: Yes it's a breaking change, but semver allows breaking changes with bugfixes.

What can I do to improve (or merge) this PR?

mdellweg: I feel that the term "as broad a range [...] as possible" is dangerous. Can you add that extra care and awareness is needed to keep the lower bound of a dependency in sync with the features used from that library? And also that bumping the lower bound may break plugin compatibilities? We should at least recognize that there is danger.

BTW: Even when not importing from a library directly, a plugin may be faced to interact with its objects. So pulpcore adopting a new version can break plugins in any case. In the end we need to accept that the pulp plugin api is not an island.

bmbouter: I don't know what to do to improve this, so I've made no changes.

My perspective is that the plugin API breaking change policy is only scoped to breaking changes with the plugin API. If revisions of Pulp upgrade a dependency to fix a bug or due to an internal change that makes no breaking changes in the plugin API, and a plugin happens to exclusively depend on that version that's unfortunate. That happens in the Python ecosystem now and then caused by the structure of a dependency in use directly and also used by a dependency of a dependency.

It's that very concern that motivates the recommendation for as broad a set as possible actually.

@mdellweg let me know what you think. I think I'm not understanding your perspective despite understanding the words. Thanks!

mdellweg: OK, let's start with the fact that i am missing a mention of dependabot that will try to bump the upper version bound for us. Second as i stated above, i would like to see reasons for bumping the lower bound (always a manual process). And this is a responsibility that comes with every single change. If we use a new library feature and not bump the dependency we will be exactly where we don't want to be: CI won't tell us right away, and it fails at the consumer/plugin who has version constraints not to use the new version of the library. This time not even at installation, but running in production.

mdellweg: I keep wondering, if this is the working and accepted way to do things in the python ecosystem, why do we even need to write a single line about it?

bmbouter: > I keep wondering, if this is the working and accepted way to do things in the python ecosystem, why do we even need to write a single line about it?

It's the norm for Python software that operates as a "library" because those softwares are installed side-by-side a lot of other software it can't control. End-user python code though tends to be the only thing installed in an environment so there is really only one dependency tree. I believe we need clarification on this because Pulp had been operating as if it was the latter but really it's the former.

bmbouter: @mdellweg ty for the comment it outlines pretty clearly the two things I should add:

* dependabot should handle the upper bound and plugin writers are enocuraged to use it
* reasons for increasing the lower bound, e.g. usage of new features or bugfixes only available in newer versions

Let me add those and you can tell me what you think.

bmbouter: While working on the revisions I found this, which aligns I think what we're recommending pretty well too https://pip.pypa.io/en/latest/topics/dependency-resolution/#loosen-your-top-level-requirements

mdellweg (approved):
Let's see how this works out. Also please consider adding a note about "having less dependencies" keeps you further from "dependency hell".

bmbouter:
I'm declining to add a part about having fewer dependencies since I believe if there's a dependency we should use it. This has downsides though so I also won't add a statement suggesting that either.

Thank you all for the engaging discussion and reviews!

PR #3178, opened 2022-09-07
Add stub implementation for pulp_uuid()

dralley:
The other PR https://github.com/pulp/pulpcore/pull/3136 merges a change which which would result in some plugins desiring a new migration.  The migration isn't *necessary* for the plugin to work - it would work perfectly fine without it - but in terms of CI and development it might be annoying if not merged. 

The migration would reference this `pulp_uuid()` function provided by pulpcore 3.21, so in normal circumstances the plugin author would want to bump their minimum supported pulpcore version if they merged it, like we would with any other core feature that a plugin wants to leverage (e.g. RBAC).

However, since simply providing that function in pulpcore (without using it anywhere) only requires about 4 lines of absolutely trivial code and doesn't carry any risk (it would also be dead code, as far as existing pulpcore / plugins are concerned), it's something we could easily and safely backport, and if we did so, plugins could adopt that migration with less concern about the pulpcore version.  

[noissue]

### DIFF
--- a/pulpcore/app/models/__init__.py
+++ b/pulpcore/app/models/__init__.py
@@ -5,6 +5,7 @@ from .base import (  # noqa
     BaseModel,
     Label,
     MasterModel,
+    pulp_uuid,
 )
 
 from .access_policy import (  # noqa
--- a/pulpcore/app/models/base.py
+++ b/pulpcore/app/models/base.py
@@ -9,8 +9,18 @@ from django.db.models.base import ModelBase
 from django_lifecycle import LifecycleModel
 
 
+def pulp_uuid():
+    """
+    Abstract wrapper for UUID generator.
+
+    Allows the implementation to be swapped without triggering migrations.
+    """
+    return uuid.uuid4()
+
+
 class Label(LifecycleModel):
-    """Model for handling resource labels.
+    """
+    Model for handling resource labels.
 
     Labels are key/value data that can be associated with any BaseModel.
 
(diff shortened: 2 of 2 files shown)


bmbouter (approved):
This seems safe to accept and backport (to me).

bmbouter:
Also since it's 100% refactor and totally invisible to both plugin writers and users, I'm ok w/ no release note at all.

dralley:
I mean, teeeeechnically it is still an addition to the plugin API.  Just the most insignificant one imaginable.  But we cannot backport "features" which is what this would have to be marked as, even though I'm not sure it necessarily applies.

bmbouter:
Hmmmmmm, that's true. I think we need to follow the letter here then. Given that we couldn't backport it, and it would need a release note for plugin writers, what do you recommend?

dralley:
Actually.. does it need to be present in the plugin API at all?  I'm not sure it does

The migration uses the direct path

bmbouter:
Pulling it out of the plugin API would make it backportable. What about the plugins that need to eventually make migrations, don't they need to import it? Can you give a specific example of a plugin migration?

dralley:
>Pulling it out of the plugin API would make it backportable. What about the plugins that need to eventually make migrations, don't they need to import it? Can you give a specific example of a plugin migration?

Plugins do not need to import it, they're in this situation because they're inheriting from `BaseModel` directly, and so by proxy they use whatever `BaseModel` is using, which is controlled by the Pulpcore version.

dralley:
I should clarify - they don't need to import it via the plugin API.  They're importing it from pulpcore directly, via `BaseModel`.

bmbouter:
This is all sounding very complex to make sure nothing breaks. I think it would be so much simpler to close the PRs for now and align the changes to 3.25.

dkliban (approved):

mdellweg (approved):

PR #3179, opened 2022-09-08
Hide protected distros

mdellweg:


mdellweg:
Should this be guarded by a setting?

mdellweg:
I have a slightly different implementation here:
https://github.com/mdellweg/pulpcore/tree/template_listing

It gives the template more information to act on. Also we may eventually allow the template to be configurable. WDYT?

dralley:
>Should this be guarded by a setting?

That makes sense.  If you want to, go ahead, it certainly wouldn't hurt.

ipanova:
Reading the originally reported issue I understand that we've taken the esiest approach however I think my preference would be if we whether now or at some point make it better and add `Users have expressed that if they present valid client certificates when visiting /pulp/content/ they would see those content guarded repositories listed. `

ipanova (approved):

mdellweg:
> Reading the originally reported issue I understand that we've taken the easiest approach, however ,I think my preference would be if we whether now or at some point make it better and add ability for `Users have expressed that if they present valid client certificates when visiting /pulp/content/ they would see those content guarded repositories listed. `

I agree that it would be nice, but also very much more complicated. You would need to call into all existing content guards (once if we are real clever with caching the result) and ask for their permission in a single request. On a large installation, this is an absolute no-go to me. The question of permission just cannot be answered with a clever sql statement.

dralley (approved):

dralley:
I also agree it would be nice but, it's best to stick with the minimum requested functionality for the reasons Matthias mentioned.  We could file an RFE for the more sophisticated behavior and we can add the discussion there, though.

However, it would be nice to add a test for this to https://github.com/pulp/pulp_file/blob/main/pulp_file/tests/functional/api/from_pulpcore/test_content_guard.py

We should use `xml.etree.ElementTree` to parse the index page and make sure that non-guarded distributions appear and that guarded distributions do not.

dralley:
Actually, the test would make more sense here: https://github.com/pulp/pulp_file/blob/main/pulp_file/tests/functional/api/from_pulpcore/test_content_path.py

mdellweg:
I added a test here. But it does not run with that feature turned on yet.
https://github.com/pulp/pulp_file/pull/808

dralley (approved):

PR #3182, opened 2022-09-08
Added pulpcore-manager command for removing signing services

### LINKED ISSUE #2967: Facilitate removal of signing services in functional tests
Allowing testers to call a function for removing a signing service would be useful. Currently, we remove the signing service in the following manner: https://github.com/pulp/pulpcore/blob/6287ef36d2fc20ca316e1e889296fc55072ff477/pulpcore/tests/functional/gpg_ascii_armor_signing_service.py#L158-L168

Having a fixture `remove_signing_service(service_name, service_type)` accessible globally would suffice.

MichalPysik:
The pulpcore-manager now supports command 'remove-signing-service' which allows for removing signing services specified by name and class.

closes #2967

### DIFF
--- a/CHANGES/2967.feature
+++ b/CHANGES/2967.feature
@@ -1,0 +1,1 @@
+Added ``pulpcore-manager`` command called ``remove-signing-service`` for removing specified signing services.
--- a/pulpcore/app/management/commands/remove-signing-service.py
+++ b/pulpcore/app/management/commands/remove-signing-service.py
@@ -1,0 +1,66 @@
+from gettext import gettext as _
+
+from django.core.management import BaseCommand, CommandError
+
+from django.core.exceptions import ObjectDoesNotExist
+
+from django.apps import apps
+from django.db.utils import IntegrityError
+
+from pulpcore.app.models.content import SigningService as BaseSigningService
+
+
+class Command(BaseCommand):
+    """
+    Django management command for removing a signing service.
+
+    This command is in tech-preview.
+    """
+
+    help = "Removes an existing AsciiArmoredDetachedSigningService. [tech-preview]"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "name",
+            help=_("Name that the signing_service has in the database."),
+        )
+        parser.add_argument(
+            "--class",
+            default="core:AsciiArmoredDetachedSigningService",
+            required=False,
+            help=_("Signing service class prefixed by the app label separated by a colon."),
+        )
+
+    def handle(self, *args, **options):
+        name = options["name"]
+        if ":" not in options["class"]:
+            raise CommandError(_("The signing service class was not provided in a proper format."))
+        app_label, service_class = options["class"].split(":")
+
+        try:
+            SigningService = apps.get_model(app_label, service_class)
+        except LookupError as e:
+            raise CommandError(str(e))
+        if not issubclass(SigningService, BaseSigningService):
+            raise CommandError(
+                _("Class '{}' is not a subclass of the base 'core:SigningService' class.").format(
+                    options["class"]
+                )
+            )
+
+        try:
+            SigningService.objects.get(name=name).delete()
+        except IntegrityError:
+            raise CommandError(
+                _("Signing service '{}' could not be removed because it's still in use.").format(
+                    name
+                )
+            )
+        except ObjectDoesNotExist:
+            raise CommandError(
+                _("Signing service '{}' of class '{}' does not exists.").format(
+                    name, options["class"]
+                )
+            )
+        else:
+            self.stdout.write(_("Signing service '{}' has been successfully removed.").format(name))
--- a/pulpcore/tests/functional/api/test_signing_service.py
+++ b/pulpcore/tests/functional/api/test_signing_service.py
@@ -1,0 +1,7 @@
+import pytest
+
+
+@pytest.mark.parallel
+def test_crud_signing_service(ascii_armored_detached_signing_service):
+    service = ascii_armored_detached_signing_service
+    assert "/api/v3/signing-services/" in service.pulp_href
--- a/pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
+++ b/pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
@@ -157,9 +157,10 @@ def _ascii_armored_detached_signing_service_name(
 
     cmd = (
         "pulpcore-manager",
-        "shell",
-        "-c",
-        f'from pulpcore.app.models import AsciiArmoredDetachedSigningService;print(AsciiArmoredDetachedSigningService.objects.get(name="{service_name}").delete())',  # noqa: E501
+        "remove-signing-service",
+        service_name,
+        "--class",
+        "core:AsciiArmoredDetachedSigningService",
     )
     subprocess.run(
         cmd,
--- a/pulpcore/app/management/commands/add-signing-service.py
+++ b/pulpcore/app/management/commands/add-signing-service.py
@@ -11,6 +11,8 @@ from django.core.management import BaseCommand, CommandError
 from django.apps import apps
 from django.db.utils import IntegrityError
 
+from pulpcore.app.models.content import SigningService as BaseSigningService
+
 
 class Command(BaseCommand):
     """
@@ -65,6 +67,12 @@ class Command(BaseCommand):
             SigningService = apps.get_model(app_label, service_class)
         except LookupError as e:
             raise CommandError(str(e))
+        if not issubclass(SigningService, BaseSigningService):
+            raise CommandError(
+                _("Class '{}' is not a subclass of the base 'core:SigningService' class.").format(
+                    options["class"]
+                )
+            )
 
         gpg = gnupg.GPG(gnupghome=options["gnupghome"], keyring=options["keyring"])
 


pulpcore/app/management/commands/remove-signing-service.py
```diff
@@ -0,0 +1,60 @@
...
+        if ":" not in options["class"]:
+            raise CommandError(_("The signing service class was not provided in a proper format."))
+        app_label, service_class = options["class"].split(":")
+
+        try:
+            SigningService = apps.get_model(app_label, service_class)
+        except LookupError as e:
+            raise CommandError(str(e))
+
+        available_signing_services = {ss.name for ss in SigningService.objects.all()}
```
mdellweg, line 41:
```python
service = SigningService.objects.get(name=name)
service.delete()
```
Should have been sufficient. You'd catch both `IntegrityError` and `DoesNotExist`.

MichalPysik: That sounds good, I will change that for the deletion code. Although I will keep the lines you selected (38-41), since their purpose is mainly to show the user the list of all available services when he selects a non-existent one.

pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
```diff
@@ -157,9 +157,10 @@ def _ascii_armored_detached_signing_service_name(
 
     cmd = (
         "pulpcore-manager",
-        "shell",
-        "-c",
-        f'from pulpcore.app.models import AsciiArmoredDetachedSigningService;print(AsciiArmoredDetachedSigningService.objects.get(name="{service_name}").delete())',  # noqa: E501
+        "remove-signing-service",
+        service_name,
+        "--class",
+        "core:AsciiArmoredDetachedSigningService",
```
mdellweg, line 163: :+1: Thank you for not forgetting this!

pulpcore/app/management/commands/remove-signing-service.py
```diff
@@ -0,0 +1,73 @@
...
+                    "For more information, please refer to the detailed error message below:"
+                ).format(name)
+            )
+            raise e
+        except ObjectDoesNotExist as e:
+            self.stdout.write(
+                _(
+                    "Model: '{}' does not exist, which probably means that it was deleted by "
+                    "a different process while this action was in progress. "
+                    "For more information, please refer to the detailed error message below:"
```
mdellweg, line 69: I would rephrase this to `raise CommandError("Signing service {name} of class {} does not exist.").
Reason: Let's not start to speculate for the user. The name could have been a typo and never existed anyway. Also the exception tells us "The thing de not exist". That is all the information the user ever wants to hear. There is no point in rerasining it.

pulpcore/app/management/commands/remove-signing-service.py
```diff
@@ -0,0 +1,73 @@
...
+            SigningService.objects.get(name=name).delete()
+        except IntegrityError as e:
+            self.stdout.write(
+                _(
+                    "Model: '{}' could not be removed, because it is being referenced by a "
+                    "different model through a relationship that preserves the database integrity. "
+                    "For more information, please refer to the detailed error message below:"
+                ).format(name)
+            )
+            raise e
```
mdellweg, line 63: Same here. Don't generate too much information: `raise CommandError("SingingService could not be removed because it is still in use.")`

pulpcore/app/management/commands/remove-signing-service.py
```diff
@@ -0,0 +1,73 @@
...
+            raise CommandError(str(e))
+
+        available_signing_services = {ss.name for ss in SigningService.objects.all()}
+        if name not in available_signing_services:
+            raise CommandError(
+                (
+                    "Signing service name is either incorrectly specified or does not exist. "
+                    "Please specify one of the following signing service names: {}."
+                ).format(list(available_signing_services))
+            )
```
mdellweg, line 50: Let's not do this. The user didn't ask for it. There's an api to list signing services.

pulpcore/app/management/commands/remove-signing-service.py
MichalPysik: The problem with this is that it could not have been a typo, since if the service didn't exist, it wouldn't get found in the list of all available services above. But I can still shorted the message tho

pulpcore/app/management/commands/remove-signing-service.py
```diff
@@ -0,0 +1,61 @@
...
+        except LookupError as e:
+            raise CommandError(str(e))
+
+        try:
+            SigningService.objects.get(name=name).delete()
+        except IntegrityError as e:
+            self.stdout.write(
+                _(
+                    "Signing service '{}' could not be removed because it's still in use. "
+                    "For more information, see the detailed error message below:"
```
mdellweg, line 49: Can you give an example of the "more information" please?

pulpcore/app/management/commands/remove-signing-service.py
```diff
@@ -0,0 +1,61 @@
...
+                ).format(name)
+            )
+            raise e
+        except ObjectDoesNotExist as e:
+            self.stdout.write(
+                _(
+                    "Signing service '{}' of class '{}' does not exists. "
+                    "For more information, see the detailed error message below:"
+                ).format(name)
```
mdellweg, line 58: I really do not expect more information here. The thing does not exist. That is it.

pulpcore/app/management/commands/remove-signing-service.py
```diff
@@ -0,0 +1,55 @@
...
+
+        try:
+            SigningService.objects.get(name=name).delete()
+        except IntegrityError as e:
+            self.stdout.write(
+                _("Signing service '{}' could not be removed because it's still in use.").format(
+                    name
+                )
+            )
+            raise e
```
mdellweg, line 51:
```suggestion
            raise CommandError(
                _("Signing service '{}' could not be removed because it's still in use.").format(
                    name
                )
            )
```

pulpcore/app/management/commands/remove-signing-service.py
```diff
@@ -0,0 +1,55 @@
...
+        except ObjectDoesNotExist as e:
+            self.stdout.write(_("Signing service '{}' of class '{}' does not exists.").format(name))
+            raise e
```
mdellweg, line 54:
```suggestion
            raise CommandError(_("Signing service '{}' of class '{}' does not exists.").format(name))
```

pulpcore/app/management/commands/remove-signing-service.py
```diff
@@ -0,0 +1,55 @@
...
+        name = options["name"]
+        if ":" not in options["class"]:
+            raise CommandError(_("The signing service class was not provided in a proper format."))
+        app_label, service_class = options["class"].split(":")
+
+        try:
+            SigningService = apps.get_model(app_label, service_class)
+        except LookupError as e:
+            raise CommandError(str(e))
+
```
mdellweg, line 42: Can we add a check here that the class is a subclass of the signing service base class?

MichalPysik:
Should I maybe rewrite the stdout writes to raises of CommandError instead? @mdellweg

pulpcore/app/management/commands/remove-signing-service.py
```diff
@@ -0,0 +1,65 @@
...
+        name = options["name"]
+        if ":" not in options["class"]:
+            raise CommandError(_("The signing service class was not provided in a proper format."))
+        app_label, service_class = options["class"].split(":")
+
+        try:
+            SigningService = apps.get_model(app_label, service_class)
+        except LookupError as e:
+            raise CommandError(str(e))
+        if SigningService.__base__ is not BaseSigningService:
```
mdellweg, line 44: there's a builtin `issubclass`.
Also can you add the same check to the other management command?

MichalPysik: I tried that, but it didn't work, maybe I did it incorrectly so I will try it again. What other command do you mean, ``add-signing-service`` ?

mdellweg (approved):

CHANGES/2967.feature
```diff
@@ -0,0 +1 @@
+Added pulpcore-manager command called remove-signing-service for removing specified signing services.
```
mdellweg, line 1:
```suggestion
Added pulpcore-manager command called ``remove-signing-service`` for removing specified signing services.
```

pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
```diff
@@ -157,9 +157,10 @@ def _ascii_armored_detached_signing_service_name(
 
     cmd = (
         "pulpcore-manager",
-        "shell",
```
bmbouter, line 160: :+1:

pulpcore/app/management/commands/remove-signing-service.py
```diff
@@ -0,0 +1,65 @@
...
+        if not issubclass(SigningService, BaseSigningService):
+            raise CommandError(
+                _("Class '{}' is not a subclass of the base 'core:SigningService' class.").format(
+                    options["class"]
+                )
+            )
+
+        try:
+            SigningService.objects.get(name=name).delete()
+            self.stdout.write(_("Signing service '{}' has been successfully removed.").format(name))
```
bmbouter, line 53: This could move out of the try block entirely or to an `else` clause.

bmbouter (changes requested):
@MichalPysik right now we have fixtures that use these commands, but can you add a test in pulpcore that creates and deletes a `core:AsciiArmoredDetachedSigningService` signing service. I just did an audit in pulpcore and pulp_file and I don't see any test that actually uses these fixtures in either repo. Since the signing service and fixture are in pulpcore, I think it can just live in pulpcore.

What do you think? If this doesn't make sense, please ask. Thanks!

pulpcore/tests/functional/api/test_signing_service.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+
+@pytest.mark.parallel
+def test_crud_signing_service(ascii_armored_detached_signing_service):
+    service = ascii_armored_detached_signing_service
+    assert service
+    assert service.pulp_href.isascii()
```
mdellweg, line 8: I thought that you'd check that the pulp_href contains the string "ascii". And maybe it doesn't...

pulpcore/tests/functional/api/test_signing_service.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
+
+@pytest.mark.parallel
+def test_crud_signing_service(ascii_armored_detached_signing_service):
```
bmbouter, line 5: :heart_eyes:

bmbouter (approved):
Thank you!!

bmbouter:
I see two approvals so I'm going to merge.

pulpcore/tests/functional/api/test_signing_service.py
MichalPysik: The pulp_href doesn't contain the word 'ascii', I checked the format of the href.

tjmullicani:
Any idea when this feature will be integrated into the pulpcore release schedule? I didn't see it in 3.21.3.

dralley:
It'll be in the next released, probably 1-2 weeks from now.

PR #3183, opened 2022-09-08
Check for instances of subclassed django-storages backends

### LINKED ISSUE #3181: Allow use of subclassed django-storages backends
We want to roll our own storage backend by subclassing one from django-storages. However, [Pulp is checking the DEFAULT_FILE_STORAGE setting instead of using isinstance](https://github.com/pulp/pulpcore/blob/f31247eed22613ac311fd87d062a2ffacf96d0de/pulpcore/content/handler.py#L829) which prevents us from doing so.

daviddavis:
fixes #3181

### DIFF
--- a/CHANGES/3181.feature
+++ b/CHANGES/3181.feature
@@ -1,0 +1,3 @@
+Updated content app code to check for subclassed django-storages backends to allow the use of custom
+storage backends. Users can subclass ``S3Boto3Storage`` or ``AzureStorage`` in order to add their
+own custom logic to these backends and then use these backends with Pulp.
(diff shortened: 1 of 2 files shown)


daviddavis:
Hmm looks like when importing the backend classes, the code tries to import the modules that are optional and aren't necessarily installed.

jasonzio:
```
try:
    from storages.backends.s3boto3 import S3Boto3Storage
    from storages.backends.azure_storage import AzureStorage

    def isS3Boto3Storage(storage_type):
        return isinstance(storage_type, S3Boto3Storage)

    def isAzureStorage(storage_type):
        return isinstance(storage_type, AzureStorage)
except ImportException:
    def isS3Boto3Storage(storage_type):
        return False

    def isAzureStorage(storage_type):
        return False
```

Then use those predicates down below.

PR #3188, opened 2022-09-09
Fixes #3187 - Fs exporter correctly handles hardlink across partitions

### LINKED ISSUE #3187: FS Exporter should copy if it can't hard link
**Version**
Pulpcore 3.16 and above
 
**Describe the bug**
The FS Exporter works well when exporting with hardlinks, but does not handle the case where the linking might fail because pulp and exports directory might be on different mounted volumes.
**To Reproduce**

- Setup an nfs or sshfs mount
- add the mount directory to `ALLOWED_EXPORT_PATHS` in `/etc/pulp/settings.py`
- Export a repo to this mount point via fs exporter

**Actual Behavior**
```
Error: [Errno 18] Invalid cross-device link: '/var/lib/pulp/media/artifact/7a/831f9f90bf4d21027572cb503d20b702de8e8785b02c0397445c2e481d81b3' -> '/exports/repo/Packages/b/bear-4.1-1.noarch.rpm'
```
**Expected behavior**

If it can't hard link, it should instead try to make a right along the lines of -> https://github.com/pulp/pulp-2to3-migration/blob/main/pulp_2to3_migration/app/plugin/content.py#L163-L170

**Additional context**
It was discussed during the https://github.com/pulp/pulpcore/pull/2951 PR but got missed in that commit.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2125444

parthaa:
Fixed the fs exporter to handle the case where hardlink method was requested but pulp and export locations are in different partitions so can't share the same inode.

### DIFF
--- a/pulpcore/app/tasks/export.py
+++ b/pulpcore/app/tasks/export.py
@@ -2,6 +2,7 @@ import hashlib
 import json
 import logging
 import os
+import os.path
 import subprocess
 import tarfile
 
@@ -75,6 +76,15 @@ def _export_to_file_system(path, relative_paths_to_artifacts, method=FS_EXPORT_M
         and method != FS_EXPORT_METHODS.WRITE
     ):
         raise RuntimeError(_("Only write is supported for non-filesystem storage."))
+    os.makedirs(path)
+    export_not_on_same_filesystem = (
+        settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem"
+        and os.stat(settings.MEDIA_ROOT).st_dev != os.stat(path).st_dev
+    )
+
+    if method == FS_EXPORT_METHODS.HARDLINK and export_not_on_same_filesystem:
+        log.info(_("Hard link cannot be created, file will be copied."))
+        method = FS_EXPORT_METHODS.WRITE
 
     for relative_path, artifact in relative_paths_to_artifacts.items():
         dest = os.path.join(path, relative_path)
@@ -82,9 +92,11 @@ def _export_to_file_system(path, relative_paths_to_artifacts, method=FS_EXPORT_M
 
         if method == FS_EXPORT_METHODS.SYMLINK:
             src = os.path.join(settings.MEDIA_ROOT, artifact.file.name)
+            os.path.lexists(dest) and os.unlink(dest)
             os.symlink(src, dest)
         elif method == FS_EXPORT_METHODS.HARDLINK:
             src = os.path.join(settings.MEDIA_ROOT, artifact.file.name)
+            os.path.lexists(dest) and os.unlink(dest)
             os.link(src, dest)
         elif method == FS_EXPORT_METHODS.WRITE:
             with open(dest, "wb") as f, artifact.file as af:
--- a/CHANGES/1949.bugfix
+++ b/CHANGES/1949.bugfix
@@ -1,0 +1,1 @@
+Fixed the fs exporter to handle the case where there are pre-existing files in the location that FileSystem attempts to export to you get a FileExistsError.
--- a/CHANGES/3187.bugfix
+++ b/CHANGES/3187.bugfix
@@ -1,0 +1,1 @@
+Fixed the fs exporter to handle the case where hardlink method was requested but pulp and export locations are in different partitions so can't share the same inode.


parthaa:
fixes #3187

pulpcore/app/tasks/export.py
```diff
@@ -76,6 +76,9 @@ def _export_to_file_system(path, relative_paths_to_artifacts, method=FS_EXPORT_M
     ):
         raise RuntimeError(_("Only write is supported for non-filesystem storage."))
 
+    if not method in [FS_EXPORT_METHODS.SYMLINK, FS_EXPORT_METHODS.HARDLINK, FS_EXPORT_METHODS.WRITE]:
+        raise RuntimeError(_("Unsupported export method '{}'.").format(method))
```
dralley, line 80: The serializer layer ought to be handling this already, right?

pulpcore/app/tasks/export.py
```diff
@@ -85,8 +92,15 @@ def _export_to_file_system(path, relative_paths_to_artifacts, method=FS_EXPORT_M
             os.symlink(src, dest)
         elif method == FS_EXPORT_METHODS.HARDLINK:
             src = os.path.join(settings.MEDIA_ROOT, artifact.file.name)
-            os.link(src, dest)
-        elif method == FS_EXPORT_METHODS.WRITE:
+            try:
+                os.link(src, dest)
+            except FileExistsError:
+                pass
```
dralley, line 98: Is this related to https://github.com/pulp/pulpcore/issues/1949 ?

pulpcore/app/tasks/export.py
parthaa: Thats a good point!. Will remove that check

pulpcore/app/tasks/export.py
parthaa: Interesting. Guess this might fix that too :)
I just added it because there was a precedent to this pattern here -> https://github.com/pulp/pulp-2to3-migration/blob/main/pulp_2to3_migration/app/plugin/content.py#L163-L170

dralley: If you do fix that one too, just mark it fixed in the commit message :)

pulpcore/app/tasks/export.py
```diff
@@ -79,14 +79,21 @@ def _export_to_file_system(path, relative_paths_to_artifacts, method=FS_EXPORT_M
     for relative_path, artifact in relative_paths_to_artifacts.items():
         dest = os.path.join(path, relative_path)
         os.makedirs(os.path.split(dest)[0], exist_ok=True)
-
+        do_write = False
         if method == FS_EXPORT_METHODS.SYMLINK:
             src = os.path.join(settings.MEDIA_ROOT, artifact.file.name)
             os.symlink(src, dest)
```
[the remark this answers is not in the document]
parthaa: That's a good point copy vs symlink. I am not sure which is better :). Writes are probably better if its across an nfs share.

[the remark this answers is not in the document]
parthaa: oh wait you are asking a diff question :). Yes `os.symlink` fails if file already exists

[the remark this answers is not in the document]
dralley: Yup, different question - let's make sure the behavior is consistent regardless of whether the user hardlinks or softlinks, if it doesn't fail if the file already exists for one then it shouldn't fail or the other.

dralley:
@parthaa Could you write a changelog entry to make the CI happy?

pulpcore/app/tasks/export.py
```diff
@@ -76,16 +76,29 @@ def _export_to_file_system(path, relative_paths_to_artifacts, method=FS_EXPORT_M
...
         dest = os.path.join(path, relative_path)
         os.makedirs(os.path.split(dest)[0], exist_ok=True)
 
         if method == FS_EXPORT_METHODS.SYMLINK:
             src = os.path.join(settings.MEDIA_ROOT, artifact.file.name)
-            os.symlink(src, dest)
+            try:
+                os.symlink(src, dest)
+            except FileExistsError:
+                pass
```
dralley, line 95: Is there a cheap way to verify that it's the same file being pointed to by the symlink?

pulpcore/app/tasks/export.py
```diff
@@ -76,16 +76,29 @@ def _export_to_file_system(path, relative_paths_to_artifacts, method=FS_EXPORT_M
     ):
         raise RuntimeError(_("Only write is supported for non-filesystem storage."))
 
+    if (
+        method == FS_EXPORT_METHODS.HARDLINK
+        and os.stat(settings.MEDIA_ROOT).st_dev != os.stat(path).st_dev
```
dralley, line 81: I would break this out into a separate variable

```Python
export_on_same_filesystem = os.stat(settings.MEDIA_ROOT).st_dev != os.stat(path).st_dev

if method == FS_EXPORT_METHODS.HARDLINK and export_on_same_filesystem:
    # ....
```

pulpcore/app/tasks/export.py
parthaa: What  about removing the dest symlink and then symlinking ?

dralley: That works too

dralley (approved):

dralley:
@parthaa If there are BZs for these issues then please link them so we can make sure they get tagged appropriately

parthaa:
> If there are BZs for these issues then please link them so we can make sure they get tagged appropriately


https://bugzilla.redhat.com/show_bug.cgi?id=2125444  . Not sure what I need to do.

parthaa:
I am also not sure why the unit tests are failing :(

dralley:
@parthaa Try rebasing on the latest main branch

ggainey (approved):
Looks good, please rebase to main and repush, should clear up the test-failures.

ggainey:
@parthaa the failing tests are telling us that the on-same-filesystem check doesn't work when you're on a block-storage device like, say, s3. See https://github.com/pulp/pulpcore/actions/runs/3219721871/jobs/5265485591#step:13:8493 for the specific point of failure. We need to figure out the right way to accomplish this.

ggainey (changes requested):
Fails on block-storage devices - need to figure out the right way to ask this question in that scenario.

pulpcore/app/tasks/export.py
```diff
@@ -75,16 +76,27 @@ def _export_to_file_system(path, relative_paths_to_artifacts, method=FS_EXPORT_M
         and method != FS_EXPORT_METHODS.WRITE
     ):
         raise RuntimeError(_("Only write is supported for non-filesystem storage."))
+    os.makedirs(path)
+    export_on_same_filesystem = (
```
ggainey, line 80:
```suggestion
    export__not_on_same_filesystem = (
```

pulpcore/app/tasks/export.py
```diff
@@ -75,16 +76,27 @@ def _export_to_file_system(path, relative_paths_to_artifacts, method=FS_EXPORT_M
...
+        settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem"
+        and os.stat(settings.MEDIA_ROOT).st_dev != os.stat(path).st_dev
+    )
+
+    if method == FS_EXPORT_METHODS.HARDLINK and export_on_same_filesystem:
```
ggainey, line 85:
```suggestion
    if method == FS_EXPORT_METHODS.HARDLINK and export__not_on_same_filesystem:
```

ggainey (changes requested):

ggainey (approved):
Changes look great - approved!

pulpcore/app/tasks/export.py
```diff
@@ -75,16 +76,27 @@ def _export_to_file_system(path, relative_paths_to_artifacts, method=FS_EXPORT_M
         and method != FS_EXPORT_METHODS.WRITE
```
dralley, line 76: @parthaa Move the check to a separate line

```Python
local_filesystem = settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem"
if not local_filesystem and method != FS_EXPORT_METHODS.WRITE:
    raise RuntimeError(_("Only write is supported for non-filesystem storage."))
    
os.makedirs(path)
if method == FS_EXPORT_METHODS.HARDLINK:
    # hardlink method is not permitted if not on a local filesystem
    if os.stat(settings.MEDIA_ROOT).st_dev != os.stat(path).st_dev:
        log.info(_("Hard link cannot be created, file will be copied."))
        method = FS_EXPORT_METHODS.WRITE
```

PR #3189, opened 2022-09-09
Improves docs on settings

### LINKED ISSUE #2417: Improve the dynaconf/config docs
One thing that's hard is that we don't clearly show example of how to set settings. We should add examples of common settings set.

bmbouter:
This improves (and shortens) the documentation on:

* Applying settings in various ways using dynaconf (with examples).
* Which settings are required

closes #2417

### DIFF
--- a/docs/configuration/applying.rst
+++ b/docs/configuration/applying.rst
@@ -3,63 +3,65 @@
 Applying Settings
 =================
 
-Pulp uses `dynaconf <https://dynaconf.readthedocs.io/en/latest/>`_ for its settings which allows you
-to configure Pulp in a few ways:
+Pulp uses `dynaconf <https://www.dynaconf.com/>`_ for its settings which allows you
+to configure Pulp settings using various ways:
 
 
-By Configuration File
+* :ref:`Environment Variables <env-var-settings>` - Enabled by default.
+
+* :ref:`Configuration File <config-file-settings>` - Disabled by default, but easy to enable.
+  Enabled by ``pulp_installer``.
+
+.. _env-var-settings:
+
+Environment Variables
 ---------------------
 
-To configure Pulp by settings file, you must set the format and location of the config file by
-specifying the ``PULP_SETTINGS`` environment variable. For example, if you wanted to use Python to
-specify your configuration, and provide it at ``/etc/pulp/settings.py`` you could::
+Configuration by specifying environment variables is enabled by default. Any
+:ref:`Setting <settings>` can be configured using Dynaconf by prepending ``PULP_`` to the setting
+name. For example :ref:`SECRET_KEY <secret-key-setting>` can be specified as the ``PULP_SECRET_KEY``
+environment variable. For example, in a shell you can use ``export`` to set this::
+
+    export PULP_SECRET_KEY="This should be a 50 chars or longer unique secret!"
+
+
+.. _config-file-settings:
+
+Configuration File
+------------------
+
+By default, Pulp does not read settings from a configuration file. Enable this by specifying the
+``PULP_SETTINGS`` environment variable with the path to your configuration file. For example::
 
     export PULP_SETTINGS=/etc/pulp/settings.py
 
+Then you can specify settings with Python variable assignment in the ``/etc/pulp/settings.py``. For
+example, you can specify :ref:`SECRET_KEY <secret-key-setting>` with::
 
-Or in a systemd file you could::
+    $ cat /etc/pulp/settings.py
+    SECRET_KEY="This should be a 50 chars or longer unique secret!"
 
-    Environment="PULP_SETTINGS=/etc/pulp/settings.py" as the Ansible Installer does.
+In this example the settings file ends in ".py" so it needs to be valid Python, but it could use any
+`dynaconf supported format <https://www.dynaconf.com/#supported-formats>`_.
+
+.. note::
+
+    The configuration file and directories containing the configuration file must be readable by the
+    user Pulp runs as. If using SELinux, assign the ``system_u:object_r:pulpcore_etc_t:s0`` label.
 
 
-Dynaconf supports `settings in multiple file formats <https://www.dynaconf.com/configuration/>`_
-Dynaconf also support `local settings file <https://www.dynaconf.com/settings_files/#local-settings-files>`_
+.. _pulp-installer-settings:
 
-This file should have permissions of:
+pulp_installer
+--------------
 
-* mode: 640
-* owner: root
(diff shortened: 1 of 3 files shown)


bmbouter:
@pulp/core can someone review this docs improvement PR please?

docs/configuration/applying.rst
```diff
@@ -3,63 +3,67 @@
...
+starts a Pulp service include::
 
-When using `pulp_installer`
----------------------------
+    Environment="PULP_SETTINGS=/etc/pulp/settings.py"
 
-A Pulp upgrade using `pulp_installer` may override the original ``/etc/pulp/settings.py``.
+A Pulp upgrade using ``pulp_installer`` may override the original ``/etc/pulp/settings.py``.
 To keep your settings through an upgrade, use the `local settings` file located at
 ``/etc/pulp/settings.local.py``.
```
mdellweg, line 69: Can you add which file take precedence?

docs/configuration/applying.rst
```diff
@@ -3,63 +3,67 @@
...
+The `pulp_installer <https://docs.pulpproject.org/pulp_installer/>`_ enables configuration via a
+settings file that lives at ``/etc/pulp/settings.py``. It does this by having each systemd file that
+starts a Pulp service include::
 
-When using `pulp_installer`
----------------------------
+    Environment="PULP_SETTINGS=/etc/pulp/settings.py"
 
-A Pulp upgrade using `pulp_installer` may override the original ``/etc/pulp/settings.py``.
+A Pulp upgrade using ``pulp_installer`` may override the original ``/etc/pulp/settings.py``.
```
mdellweg, line 67:
```suggestion
A Pulp upgrade using ``pulp_installer`` will override the original ``/etc/pulp/settings.py``.
```
Can you maybe state unambiguously that this file is "managed" by `pulp_installer`. Do we add a "Don't edit this file!" stanza via the installer to that file?

docs/configuration/applying.rst
```diff
@@ -3,63 +3,67 @@
...
 
-This file should have permissions of:
+    $ cat /etc/pulp/settings.py
+    SECRET_KEY="This should be a 50 chars or longer unique secret!"
 
-* mode: 640
-* owner: root
-* group: pulp (the group of the account that pulp runs under)
-* SELinux context: system_u:object_r:etc_t:s0
+This settings file ends in .py so it needs to be valid Python (so single quotes are necessary), but
```
mdellweg, line 44:
```suggestion
In this example the settings file ends in ".py" so it needs to be valid Python, but
```
Boolean values or numbers should not be quoted... And double quotes are fine too.

docs/configuration/applying.rst
```diff
@@ -3,63 +3,67 @@
...
+Environment Variables
 ---------------------
 
-To configure Pulp by settings file, you must set the format and location of the config file by
-specifying the ``PULP_SETTINGS`` environment variable. For example, if you wanted to use Python to
-specify your configuration, and provide it at ``/etc/pulp/settings.py`` you could::
+Configuration by specifying environment variables is enabled by default. Any
+:ref:`Setting <settings>` can be configured using Dynaconf by prepending ``PULP_`` to the setting
+name. For example :ref:`SECRET_KEY <secret-key-setting>` can be specified by exporting the
+``PULP_SECRET_KEY`` variable, for example::
```
mdellweg, line 23: [Maybe too pedantic:] Exporting variables is a shell feature. In the end we only rely on providing/setting the variable in the processes environment.

bmbouter:
@mikedep333 Can you also look at this since I'm modifying some info about the installer. You'll notice the removal of some content because I want to document the requirements, versus what I saw here which seemed more like one way to configure it.

docs/configuration/applying.rst
mikedep333: Yes, the installer puts such a header in the file:
https://github.com/pulp/pulp_installer/blob/main/roles/pulp_common/templates/settings.py.j2#L1

docs/configuration/applying.rst
bmbouter: It's a subtle difference, but I think it's meaningful. Take a look at the revision here and lmk.

docs/configuration/applying.rst
```diff
@@ -3,63 +3,67 @@
...
 
-* mode: 750
-* owner: root
-* group: pulp (the group of the account that pulp runs under)
-* SELinux context: unconfined_u:object_r:etc_t:s0
+    The configuration file and directories containing the configuration file must be readable by the
+    user Pulp runs as. If using SELinux permissions, assign the following permissions:
 
-An example of a minimal settings.py file is::
+    * The configuration file itself: ``system_u:object_r:etc_t:s0``
```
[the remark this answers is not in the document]
bmbouter: Thanks, see revision in next push.

docs/configuration/applying.rst
bmbouter: I made this change, see next push.

docs/configuration/applying.rst
bmbouter: I didn't change this because I didn't know. @mikedep333 what takes precedence?

bmbouter:
@mdellweg @mikedep333 please look once more, I really want to wrap this up so I can take on new work. Thanks!

ggainey (approved):
I think we're good here, once @mdellweg's question RE precedence is answered!

dralley (approved):

mdellweg (approved):
None of the last {ques,sugges}tions are really mandatory.

bmbouter:
I added revised language on the `settings.local.py` taking precedence.

docs/configuration/applying.rst
```diff
@@ -3,63 +3,65 @@
...
----------------------------
+    Environment="PULP_SETTINGS=/etc/pulp/settings.py"
 
-A Pulp upgrade using `pulp_installer` may override the original ``/etc/pulp/settings.py``.
-To keep your settings through an upgrade, use the `local settings` file located at
-``/etc/pulp/settings.local.py``.
+A Pulp upgrade using ``pulp_installer`` will override the original ``/etc/pulp/settings.py``.
+To keep your settings through an upgrade, use the "local settings" file located at
+``/etc/pulp/settings.local.py`` which takes precedence over ``/etc/pulp/settings.py`` on a
+setting-by-setting basis.
```
mdellweg, line 67: +1

mdellweg (approved):

mikedep333 (approved):

PR #3190, opened 2022-09-11
Add new multi-tenancy feature: domains

gerrod3:
Lots of changes to add domains: an **_opt-in_** namespace feature to allow for multi-tenancy. This is a POC and doesn't contain all the necessary changes for pulpcore (maybe 70%). The goal of this PR is for discussion and feedback and hopefully show a feature of this scale can be added in the next Y release without breaking plugins and existing users.


pulpcore/app/models/publication.py
```diff
@@ -339,6 +347,9 @@ def invalidate_cache(self):
             if base_paths:
                 Cache().delete(base_key=base_paths)
 
+    class Meta:
+        unique_together = ("name", "domain")
```
mdellweg, line 351: Would this uniqueness hold if the default domain was saved as `None`?

pulpcore/app/models/role.py
```diff
@@ -45,9 +47,12 @@ class UserRole(BaseModel):
     content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE, null=True)
     object_id = models.TextField(null=True)
     content_object = GenericForeignKey("content_type", "object_id", for_concrete_model=False)
+    # I should probably put a database exclusivity lock on this with content_obj
+    domain_id = models.TextField(null=True)
```
mdellweg, line 51: Foreign key?

pulpcore/app/serializers/repository.py
```diff
@@ -341,15 +341,17 @@ class RepositorySyncURLSerializer(ValidateFieldsMixin, serializers.Serializer):
...
         except KeyError:
-            remote = None
+            remote = data.get("remote", None)
 
-        if "remote" not in data and not remote:
+        if not remote:
             raise serializers.ValidationError(
                 {"remote": _("This field is required since a remote is not set on the repository.")}
             )
+        self.check_cross_domains({"repository": repository, "remote": remote})
```
mdellweg, line 354: Is this check not done by the `ValidateFieldsMixin`?

pulpcore/app/util.py
```diff
@@ -256,3 +257,12 @@ def configure_telemetry():
...
         models.TaskSchedule.objects.filter(task_name=task_name).delete()
+
+
+def get_default_domain():
+    return _get_default_domain()
+
+
+@lru_cache(maxsize=1)
+def _get_default_domain():
+    return models.Domain.objects.get(name="default", storage_class="default").pk
```
mdellweg, line 268:
```suggestion
@lru_cache(maxsize=1)
def get_default_domain():
    return models.Domain.objects.get(name="default", storage_class="default").pk
```

pulpcore/app/viewsets/access_policy.py
```diff
@@ -70,3 +70,79 @@ def reset(self, request, pk=None):
...
+        ],
+        "creation_hooks": [
+            {
+                "function": "add_roles_for_object_creator",
+                "parameters": {"roles": ["core.domain_owner"]},
+            },
+        ],
+        "queryset_scoping": {"function": "scope_queryset"},
+    }
+    # There is probably more locked roles needed for a feature like this
```
mdellweg, line 138: I don't even think so. It should be rather simple for this central object. The complexity is added to all the objects contained by it.
The thing we may need is another permission `can_manage_domain_roles` to allow setting domain wide roles.

pulpcore/backends.py
```diff
@@ -51,6 +67,7 @@ def has_perm(self, user_obj, perm, obj=None):
         return False
 
     def get_all_permissions(self, user_obj, obj=None):
+        # Not sure how or if I need to get this function to work with domains...
```
mdellweg, line 70: I think yes. If you receive a permission for the `obj` via a domain, it should be listed too.

pulpcore/content/__init__.py
```diff
@@ -74,4 +74,7 @@ async def server(*args, **kwargs):
                 import_module(content_module_name)
     app.add_routes([web.get(settings.CONTENT_PATH_PREFIX, Handler().list_distributions)])
     app.add_routes([web.get(settings.CONTENT_PATH_PREFIX + "{path:.+}", Handler().stream_content)])
+    # TODO extract out into settings variable
+    app.add_routes([web.get("/pulp/domain/{domain_pk}/", Handler().list_distributions)])
+    app.add_routes([web.get("/pulp/domain/{domain_pk}/{path:.+}", Handler().stream_content)])
```
mdellweg, line 79: I had envisioned that there would be the name of the domain. But this looks interesting too.

pulpcore/app/models/publication.py
gerrod3: No, this is why I am using a default_domain. Null fields can not be used for a uniqueness constraint since null != null in SQL.

mdellweg: Too sad...

pulpcore/content/__init__.py
gerrod3: If I make `domain.name` unique then yes it could be.

pulpcore/app/models/role.py
gerrod3: Yeah, roles logic is kind of messed up right now. I was testing if the uniqueness constraint was about null values or the type of data (Text vs ForeignKey)

pulpcore/app/serializers/repository.py
gerrod3: It is called, but not with all the fields filled in to be valid as repository isn't a field, it's part of the url. I think I might need to create a second method like `prepare_fields` that `check_cross_domains` calls so that it can get each field for the cross check for serializers like this.

pulpcore/content/__init__.py
mdellweg: The question is whether we _want_ to see the domain as a kind of namespace.

gerrod3: Oh, are you saying instead of `/pulp/domain/...` it should be `/pulp/{domain_pk}/...`? Otherwise I don't know how to differentiate requests to the content-app now that `base_path` is not unique.

mdellweg: No what i was saying is i always thought the domain would give us a namespace like `/pulp/content/{domain.name}/{base_path}` or (because that ship has sailed...) `/pulp/domain/{domain.name}/{base_path}`. And yes the domain name must be unique then, just as the username.
I'm not sure anymore and we need to get more information from potential users here.

mdellweg:
Collapsing into one domain is probably not going to work due to _all_ the uniqueness constraints that contain the domain pk now (e.g. names that are unique in a domain/namespace).

bmbouter:
> Collapsing into one domain is probably not going to work due to _all_ the uniqueness constraints that contain the domain pk now (e.g. names that are unique in a domain/namespace).

Agreed and I think pulp needs a startup check that refuses to start if > 1 domain and yet you have `ENABLE_DOMAIN=False`.

ipanova:
> > Collapsing into one domain is probably not going to work due to _all_ the uniqueness constraints that contain the domain pk now (e.g. names that are unique in a domain/namespace).
> 
> Agreed and I think pulp needs a startup check that refuses to start if > 1 domain and yet you have `ENABLE_DOMAIN=False`.

+1 to refuse to start

ipanova:
I noticed that few models do not have domain as fk like signing service, labels, export, taskgroup, etc. should we add it?

pulpcore/content/__init__.py
mdellweg: We had some more discussions about the url schema:
1. We want the domain name (to be unique and) be part of hrefs.
2. If we can make it so the openapi spec does not change, we can user the rerouting feature present in current client tooling to interact with different domains.
3. Suggested url schemas: `/pulp/<domain-name>/api/v3/tasks/`, `/pulp/<domain-name>/content/<base-path>/`

bmbouter: I thought some more about that and I think we should consider `/pulp/api/v3/<domain-name>/` instead. The issue with the `/pulp/<domain-name>/api/v3/` approach is that it forces all objects to be domain namespaced even where there aren't any. We have some objects that just aren't, e.g. Workers, Users, etc. One might think, "oh well let's just put them into the default namespace" but that becomes very inconvenient (and likely inappropriate) when it comes to authorization of those calls into the default namespace because in practice all users now need permissions for that domain in some way.

This inconvenience stems from us namespacing an object that really shouldn't be namespaced, so the way to avoid it is not to do that, which brings us back to `/pulp/api/v3/<domain-name>/` for objects that are domain enabled and `/pulp/api/v3/` for those that aren't.

I think we are doing this mostly out of convenience for the CLI (which was a fine idea) but these practical problems I think outweight that. That CLI and bindings, and openAPI schema need a `domain` setting I think which will be used when needed for those object types.

What do you think about ^?

mdellweg: Actually I was thinking (and not writing it down) that workers, users, groups and other non-domain entities would only be routed via the `/pulp/api/v3/` path that would map into the default domain right away. But that might add a lot of complexity.

`/pulp/api/v3/<domain-name>/` does not work for me, because it collides with paths we already have, and we keep adding interfaces at that level. We cannot prevent a domain to be named `keys` e.g. if we intend to add an api for `keys` in the future. Thinking about client tooling, i'd say for the cli it's easy to add another path parameter and use that if the openapi spec declares it properly. The CLI consumes that spec on the fly. The published bindings will have a hard time though (reason to stop publishing them?).

What would work for me [SPOILER]: `/pulp/api/v3/status` and all that jazz continues to work (and maybe just use the default domain for domained entities; that's why we call it "default" in the first place.). `/pulp/api/v4/<domain-name>/repositories/` and the like will be used for domained entities only. At this point we can inject the domain at the router level.

mdellweg: We might think about specifying locked resources differently then: `<app-label>:<class-name>:<uuid>` instead of the now rather unstable `<href>`.
We can still translate to href's in the serializer.

chris1984:
@gerrod3 I think I saw you yesterday at the transfer hall, were you in town for pulpcon?

pulpcore/app/tasks/purge.py
```diff
@@ -69,8 +70,10 @@ def purge(finished_before, states):
 
     """
     current_user = get_current_authenticated_user()
-    # Tasks, prior to the specified date, in the specified state, owned by the current-user
-    tasks_qs = Task.objects.filter(finished_at__lt=finished_before, state__in=states)
+    domain = Task.current().domain
+    # Tasks, prior to the specified date, in the specified state, owned by the current-user, in the
+    # current domain
+    tasks_qs = Task.objects.filter(finished_at__lt=finished_before, state__in=states, domain=domain)
```
lubosmj, line 76: Would it be possible to facilitate the retrieval of domains under the hood so that plugin writers will not be required to adjust affected filtering queries?

For instance, is it achievable to create a new manager that will handle the filtering by first running `queryset.filter(domain=Task.current().domain)` (https://docs.djangoproject.com/en/3.2/topics/db/managers/) before executing the actual queries? All models that are associated with domains should then have this manager attached to them.

gerrod3: This is a good idea, I'll look into it. We use querysets in many different spots so knowing how to get the domain from each context might be a bit challenging.

mdellweg: We determine the domain in the middleware, right? We could even set a context variable (like set_guid) so we can access it from all custom managers.

docs/plugin_dev/plugin-writer/concepts/domains/domains_compatibility.rst
```diff
@@ -0,0 +1,156 @@
...
+relation to be added onto the model and have their ``unique_together`` updated. See code below for
+an example:
+
+.. code-block:: python
+
+    from pulpcore.plugin.utils import get_default_domain
+
+    class FileContent(Content):
+        ...
+        domain = models.ForeignKey("core.Domain", default=get_default_domain, on_delete=models.PROTECT)
```
mdellweg, line 29: Have you looked into constraints to make the unique across inherited properties work? I'd really prefer to see the domain on the base class.
https://docs.djangoproject.com/en/3.2/ref/models/options/#constraints

pulpcore/app/serializers/base.py
```diff
@@ -38,16 +40,44 @@ def validate_unknown_fields(initial_data, defined_fields):
...
     def validate(self, data):
         if hasattr(self, "initial_data"):
             validate_unknown_fields(self.initial_data, self.fields)
 
+        self.check_cross_domains(data)
+
         data = super().validate(data)
         return data
 
+    def check_cross_domains(self, data):
```
mdellweg, line 69: I would see this in a simpler way: The request (or task) must be attached to a single domain, and every field that references any object must check against that domain.

pulpcore/app/serializers/base.py
```diff
@@ -71,6 +101,27 @@ def get_or_create(cls, natural_key, default_values=None):
         return result
 
 
+def get_domain(self):
```
mdellweg, line 104: Is this function meant to be a method for some class?

pulpcore/app/serializers/publication.py
```diff
@@ -194,8 +190,14 @@ class DistributionSerializer(ModelSerializer):
...
 
+    def __init__(self, *args, **kwargs):
+        # Prevent unique_together validators from being added
+        super().__init__(*args, **kwargs)
+        if getattr(self.Meta, "validators", None) is None:
+            self.Meta.validators = []
+
     class Meta:
-        abstract = True
+        # abstract = True
```
mdellweg, line 200: Abstract in a serializer? I think this can go away. ;)

pulpcore/app/viewsets/orphans.py
```diff
@@ -19,18 +19,24 @@ def cleanup(self, request):
...
+        # TODO Check that content list doesn't cross domains
         serializer = OrphansCleanupSerializer(data=request.data)
         serializer.is_valid(raise_exception=True)
 
         content_pks = serializer.validated_data.get("content_hrefs", None)
         orphan_protection_time = serializer.validated_data.get(
             "orphan_protection_time", settings.ORPHAN_PROTECTION_TIME
         )
+        # Maybe this should take into account the API_ROOT?
+        uri = "/pulp/api/v3/orphans/cleanup/"
```
mdellweg, line 31: Since the api root is something we can reconfigure, this is not even safe the way we use it today. Or put another way: In order to reconfigure the API-ROOT, you should really drain your task queue first.
Ideally we would refer to things by another (stable) identifier.

pulpcore/middleware.py
```diff
@@ -0,0 +1,38 @@
...
+        return response
+
+    def process_view(self, request, view_func, view_args, view_kwargs):
+        """Remove the domain name if present, called right before view_func is called."""
+        # Maybe add logic if the view_func is expecting the domain to be passed?
+        domain = view_kwargs.pop("domain_path", None) or "default"
+        log.info(f"{domain=}")
+        if not Domain.objects.filter(name=domain).exists():
+            raise Http404()
+        setattr(request, "domain", domain)
```
mdellweg, line 37: Can we set the hydrated object here? (`request.user` is handled that way.)
And maybe we set it as a context var here (or in the `__call__` part).

pulpcore/middleware.py
```diff
@@ -0,0 +1,38 @@
...
+        # Code to be executed for each request/response after
+        # the view is called.
+
+        return response
+
+    def process_view(self, request, view_func, view_args, view_kwargs):
+        """Remove the domain name if present, called right before view_func is called."""
+        # Maybe add logic if the view_func is expecting the domain to be passed?
+        domain = view_kwargs.pop("domain_path", None) or "default"
+        log.info(f"{domain=}")
```
mdellweg, line 34: This will be quite verbose. I think it should at least be debug in the final version.

mdellweg (commented):
Sorry for being late with this idea. But seeing all the many cases where you inject the domain in function signatures may be handled by having the domain as a context var.

I think of them as the call stack having class attributes.
https://docs.python.org/3.7/library/contextvars.html

pulpcore/app/serializers/base.py
```diff
@@ -8,14 +8,16 @@
 from django.core.validators import URLValidator
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import IntegrityError, transaction
+from django.db.models import Model
 from drf_queryfields.mixins import QueryFieldsMixin
+from dynaconf import settings
```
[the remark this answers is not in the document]
gerrod3: It should be `django.conf`, thanks for catching this.

pulpcore/app/serializers/base.py
gerrod3: Good point, I'll refactor this to be simpler.

pulpcore/app/serializers/base.py
gerrod3: It was a part of the `ModelSerializer` class, but I factored it out because I thought I might need to use this in other serializers that don't inherit from the `ModelSerializer`. There isn't any other serializers in pulpcore that use this method, but I do have it exposed in the plugin api currently for potential future use.

pulpcore/middleware.py
gerrod3: Whoops, leftover print statement when I was first debugging.

pulpcore/app/global_access_conditions.py
```diff
@@ -47,9 +47,7 @@ def has_domain_perms(request, view, action, permission):
     """
     # Usually list doesn't have a permission check, so not sure if needed to check for
     if settings.DOMAIN_ENABLED:
-        domain_name = request.domain
-        domain = Domain.objects.get(name=domain_name)
-        return request.user.has_perm(permission, obj=domain)
+        return request.user.has_perm(permission, obj=request.domain)
```
mdellweg, line 50: :+1:

pulpcore/app/response.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, task, request):
         """
         args = [task.pk]
         if settings.DOMAIN_ENABLED:
-            args.insert(0, request.domain)
+            args.insert(0, request.domain.name)
```
mdellweg, line 28: can this be like for task groups below?

```
        kwargs = {"pk": task_group.pk}
        if settings.DOMAIN_ENABLED:
            kwargs["domain_path"] = request.domain  # This should be the task_group's domain
            kwargs["domain_path"] = request.domain.name  # This should be the task_group's domain
        resp = {"task_group": reverse("task-groups-detail", kwargs=kwargs, request=None)}
        super().__init__(data=resp, status=202)
```

pulpcore/app/serializers/base.py
mdellweg: Should we name it's argument "serializer" then? It's not meant to be used as a class method as far as i see.

pulpcore/plugin/viewsets/content.py
```diff
@@ -102,19 +101,16 @@ def create(self, request):
...
             # parameter is replaced by 'artifact'
-            artifact = Artifact.init_and_validate(task_payload.pop("file"), domain=domain)
+            artifact = Artifact.init_and_validate(task_payload.pop("file"), domain=request.domain)
             try:
                 artifact.save()
             except IntegrityError:
                 # if artifact already exists, let's use it
                 try:
-                    artifact = Artifact.objects.get(sha256=artifact.sha256, domain=artifact.domain)
+                    artifact = Artifact.objects.get(sha256=artifact.sha256, domain=request.domain)
```
mdellweg, line 113: This sounds even safer!

pulpcore/app/migrations/0097_add_domain.py
```diff
@@ -0,0 +1,39 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Domain',
+            fields=[
+                ('pulp_id', models.UUIDField(default=uuid.uuid4, editable=False, primary_key=True, serialize=False)),
+                ('pulp_created', models.DateTimeField(auto_now_add=True)),
+                ('pulp_last_updated', models.DateTimeField(auto_now=True, null=True)),
+                ('name', models.TextField(unique=True)),
```
mdellweg, line 29: Sorry for being late again.
This is dangerous, because we intend to use that name as part of a url. I think we should use a slug field. Or, We keep a nice name because company names do not necessarily contain hyphens and add a `SlugField` called "slug" to be used in the url.

https://docs.djangoproject.com/en/3.2/ref/models/fields/#slugfield

mdellweg (commented):
Next round of ideas.

pulpcore/app/response.py
gerrod3: Yes, forgot to switch this over to using kwargs instead.

pulpcore/app/migrations/0097_add_domain.py
gerrod3: Good point, I'll switch it. I was wondering if `SlugField` was the correct option after switching the name to be unique + part of the url.

pulpcore/app/migrations/0097_add_domain.py
```diff
@@ -36,4 +60,6 @@ class Migration(migrations.Migration):
             },
             bases=(django_lifecycle.mixins.LifecycleModelMixin, models.Model, pulpcore.app.models.access_policy.AutoAddObjPermsMixin),
         ),
+        migrations.RunPython(create_default_domain, reverse_code=migrations.RunPython.noop),
```
mdellweg, line 63: Seeding of data should be in a post migrate signal.

gerrod3: I just realized this was your first comment on the first commit. What is the reasoning for this? This feature can not function without the default domain, with future migrations relying on the default being present, so having it be created outside a migration feels hacky IMO.

mdellweg: Basically `pulpcore-manager flush`.
If any subsequent migration needs the default domain, it needs to create it in an idempotent way the same way as the post-migrate signal needs to. Just make sure that after successfully running all the migrations (or flush which will only trigger the post_migrate), it is there.

lubosmj (commented):
I went only through the documentation and added a few suggestions. Personally, I think we should move some parts of the `docs/workflows/domains-multi-tenancy.rst` documentation to the plugins themselves.

pulpcore/app/apps.py
```diff
@@ -211,10 +217,15 @@ class PulpAppConfig(PulpPluginAppConfig):
     # The python package name providing this app
     python_package_name = "pulpcore"
 
+    domain_compatible = True
```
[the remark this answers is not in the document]
mdellweg: This is the pulpcore config. And it anables domains, so `True` is correct here.

pulpcore/app/apps.py
```diff
@@ -273,6 +284,14 @@ def _populate_system_id(sender, apps, verbosity, **kwargs):
         SystemID().save()
 
 
+def _ensure_default_domain(sender, apps, verbosity, **kwargs):
+    table_names = connection.introspection.table_names()
+    if "core_domain" in table_names:
+        from pulpcore.app.util import get_default_domain
+
+        get_default_domain()  # Cache the default domain
```
mdellweg, line 292: This will create the default domain, right?

pulpcore/app/global_access_conditions.py
```diff
@@ -37,6 +38,18 @@ def has_model_perms(request, view, action, permission):
     return request.user.has_perm(permission)
 
 
+def has_domain_perms(request, view, action, permission):
+    """
+    Checks if the user has current domain-level permission.
+
+    Extract the domain from either the object being acted on or through a field if present. If no
+    domain is present then return False.
```
mdellweg, line 46: This text is outdated.

pulpcore/app/global_access_conditions.py
```diff
@@ -145,15 +179,27 @@ def has_remote_param_obj_perms(request, view, action, permission):
             parameter at the object-level or if there is no remote. False otherwise.
     """
     kwargs = {}
+    obj = view.get_object() if view.detail else None
     if action == "partial_update":
         kwargs["partial"] = True
-    serializer = view.serializer_class(data=request.data, context={"request": request}, **kwargs)
+    serializer = view.serializer_class(
+        instance=obj, data=request.data, context={"request": request}, **kwargs
+    )
```
mdellweg, line 187: Is this a change in behaviour?

pulpcore/app/util.py
```diff
@@ -22,18 +23,35 @@
...
-    return reverse(get_view_name_for_model(model, "detail"), args=[model.pk])
+    kwargs = {}
+    view_action = "list"
+    if settings.DOMAIN_ENABLED:
+        kwargs["pulp_domain"] = "default"
+        if hasattr(model, "pulp_domain"):
+            if not domain and isinstance(model, model._meta.model):
+                kwargs["pulp_domain"] = model.pulp_domain.name
+            elif isinstance(domain, models.Domain):
+                kwargs["pulp_domain"] = domain.name
```
mdellweg, line 47: Does this mean, a domain "lives" inside itself?

pulpcore/app/util.py
```diff
@@ -300,3 +318,34 @@ def get_artifact_url(artifact, headers=None, http_method=None):
...
+        except Domain.DoesNotExist:
+            default_domain = Domain(name="default", storage_class="default")
+            default_domain.save(skip_hooks=True)
+
+    return default_domain.pk if return_pk_only else default_domain
+
+
+def get_domain(return_pk_only=True):
+    domain = current_domain.get() or get_default_domain(return_pk_only=False)
+    return domain.pk if return_pk_only else domain
```
mdellweg, line 346: Polymorphic return values are dangerous. Can we just have `get_domain` and `get_domain_pk` if there is a need for it?

pulpcore/app/apps.py
gerrod3: Nope. The default domain is created within the migration (Through the default field on the new relations). This is no longer needed now that we are using a Domain ContextVar.

pulpcore/app/util.py
gerrod3: No, this function is set up so that you can pass in a domain to use for the reverse call besides "default" or the domain on the object. If you where to call this function with a domain as the first arg (model), then it would use "default" since the domain object doesn't  have domain relation.

pulpcore/app/util.py
gerrod3: The need is from the default on the `pulp_domain` field being set to `get_domain`, it expects the return value to be a `PK`. I can break this into two separate methods.

pulpcore/app/global_access_conditions.py
gerrod3: Yes, I realized the previous behavior was incorrect on update actions. For update the serializer should be instantiated with the object instance to perform correct validation.

docs/plugin_dev/plugin-writer/concepts/rbac/permissions.rst
```diff
@@ -5,7 +5,7 @@ The permissions system provides a way to assign permissions as part of roles to
 groups of users. The models driving this data are ``django.contrib.auth.models.Permission`` and
 ``pulpcore.plugin.models.role.Role``. Each ``Permission`` has a name, describing it and can be
 associated with one or more ``Role``. Roles can be assigned to users or groups either on the
-Model-Level or Object-Level.
+Model-Level, Domain-level (if domains are enabled) or Object-Level.
```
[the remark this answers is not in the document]
gerrod3: An Oxford comma fan.

pulpcore/app/apps.py
gerrod3: Update: The answer is actually yes & no. If you are upgrading, then the new migrations will cause the default method, `get_domain_pk`, (which in turn calls `get_default_domain`) to be ran when adding the new `pulp_domain` relation to your existing models.  If you are creating a brand new Pulp instance, then this post migrate hook will be the one to create the default domain.

mdellweg: Great! That is exactly what i needed to hear.

docs/workflows/domains-multi-tenancy.rst
```diff
@@ -0,0 +1,134 @@
...
+
+Domains are off by default and can be enabled by setting ``DOMAIN_ENABLED = True`` in your settings
+file and restarting Pulp. In order to enable domains, each plugin needs to be domain compatible. If
+a plugin is not compatible Pulp will fail to start. See list below for current list of plugins with
+domain compatibility. Once domains are enabled all current objects in Pulp will be found under the
+``default`` domain.
+
+.. _domain-plugins:
+
+Current Domain Compatible Plugins:
```
ipanova, line 38: is it worth adding this info to the status endpoint?

docs/workflows/domains-multi-tenancy.rst
```diff
@@ -0,0 +1,134 @@
...
+
+.. warning:: Changing the ``storage-class`` or ``storage-settings`` of an in-use domain is
+  dangerous and can result in a broken domain.
+
+.. _using-domains:
+
+Using Domains
+-------------
+
+Once domains are enabled all URLs in Pulp will require the domain name in the path after the
```
ipanova, line 73: Can you remind me whether it is possible to disable domains once enabled them?

gerrod3: Yes, the other domains will become inaccessible and the URLs will go back to normal.

ipanova: Awesome thanks!

docs/workflows/domains-multi-tenancy.rst
gerrod3: Possibly. I added a `domain_enabled` field to the status endpoint that reflects the setting `DOMAIN_ENABLED`. I'll look at adding it.

pulpcore/app/models/domain.py
```diff
@@ -0,0 +1,44 @@
+from django.core.files.storage import get_storage_class, default_storage
+from django.db import models
+
+from pulpcore.app.models import BaseModel, AutoAddObjPermsMixin
+
+from .fields import EncryptedJSONField
+
+
+class Domain(BaseModel, AutoAddObjPermsMixin):
```
ipanova, line 9: so you've written the trigger that will be watching for deletion at the DB level but does it make sense to also add a pre_delete hook  that will be checking removal of default domain? it will produce a more user friendly output

pulpcore/app/models/content.py
```diff
@@ -358,7 +371,7 @@ def from_pulp_temporary_file(cls, temp_file):
         Returns:
             An saved :class:`~pulpcore.plugin.models.Artifact`
         """
-        artifact_file = default_storage.open(temp_file.file.name)
+        artifact_file = temp_file.file.open()
```
ipanova, line 374: should not the file be closed?

pulpcore/app/models/content.py
```diff
@@ -465,13 +482,38 @@ class ContentManager(BulkCreateManager):
...
         expiration = now() - datetime.timedelta(minutes=orphan_protection_time)
-        if content_pks:
-            return self.filter(
-                version_memberships__isnull=True,
-                timestamp_of_interest__lt=expiration,
-                pk__in=content_pks,
-            )
-        return self.filter(version_memberships__isnull=True, timestamp_of_interest__lt=expiration)
+        if not content_pks:
+            domain_pk = get_domain()
```
ipanova, line 486: get_domain_pk or rename the var

pulpcore/app/models/content.py
```diff
@@ -465,13 +482,38 @@ class ContentManager(BulkCreateManager):
...
+            content_pks = self.in_domain(domain_pk).values_list("pk", flat=True)
+
+        return self.filter(
+            version_memberships__isnull=True,
+            timestamp_of_interest__lt=expiration,
+            pk__in=content_pks,
```
ipanova, line 492: why we are not getting the domain here?

pulpcore/app/role_util.py
```diff
@@ -25,7 +25,11 @@ def assign_role(rolename, entity, obj=None):
         entity (django.contrib.auth.User or pulpcore.app.models.Group): Entity to gain the role.
         obj (Optional[pulpcore.app.models.BaseModel]): Object the role permisssions are to be
             asserted on.
+        domain (Optional[pulpcore.app.models.Domain]): Domain the role permissions are to be
+            asserted on. Mutually exclusive with obj.
     """
+    if obj and domain:
+        raise BadRequest(_("Obj and domain can not both be set."))
     try:
```
ipanova, line 33: let's not shorten 'Object'

pulpcore/app/role_util.py
```diff
@@ -87,35 +114,62 @@ def get_objects_for_user_roles(
...
-    user_role_pks = user.object_roles.filter(role__permissions=permission).values_list(
-        "object_id", flat=True
-    )
+    user_role_pks = user.object_roles.filter(
+        domain__isnull=True, role__permissions=permission
+    ).values_list("object_id", flat=True)
+    final_q = Q(pk_str__in=user_role_pks)
+    if accept_domain_perms and hasattr(qs.model, "pulp_domain"):
+        # Not sure why this has to be cast to a list...
+        domains = list(
```
ipanova, line 138: because otherwise this is a queryset object
```
In [33]: Remote.objects.all().values_list("name", flat=True).extend
AttributeError                            Traceback (most recent call last)
Input In [33], in <cell line: 1>()
----> 1 Remote.objects.all().values_list("name", flat=True).extend

AttributeError: 'QuerySet' object has no attribute 'extend'
```

pulpcore/app/models/domain.py
gerrod3: Sure, I'll add one. The hook will still be bypassed if the default domain is deleted in bulk, but it'll prevent individually trying to delete it.

ipanova: yes, i remember that bulk delete caveat

pulpcore/app/apps.py
```diff
@@ -273,6 +284,14 @@ def _populate_system_id(sender, apps, verbosity, **kwargs):
         SystemID().save()
 
 
+def _ensure_default_domain(sender, apps, verbosity, **kwargs):
+    table_names = connection.introspection.table_names()
+    if "core_domain" in table_names:
```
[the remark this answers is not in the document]
gerrod3: Yes, migrations can be ran up or down and the post migrate signal is always sent, so this can be called without the domain table existing in the db anymore. Actually most (all?) of our post migrate hooks might break from backwards migrations, although you probably can't go back that far in migrations since some of them are irreversible.

pulpcore/app/models/content.py
gerrod3: I'm under the assumption that if you are specifying `content_pks` then you have already filtered by domain.

pulpcore/app/role_util.py
gerrod3: My comment was originally for a different error for using the domains queryset in another queryset filter `qs.filter(pulp_domain_id__in=domains)`. I had the user and group role queries separate and was adding them as two different Q objects, but I found that filtered query wasn't working unless I cast the domains query to a list. Since I was casting to a list it made sense to combine the user and group queries into one list.

pulpcore/app/apps.py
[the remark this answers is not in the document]
ipanova: right, i keep forgetting running the migration down

pulpcore/app/apps.py
```diff
@@ -273,6 +284,14 @@ def _populate_system_id(sender, apps, verbosity, **kwargs):
         SystemID().save()
 
 
+def _ensure_default_domain(sender, apps, verbosity, **kwargs):
```
mdellweg, line 287:
```suggestion
def _ensure_default_domain(sender, **kwargs):
```
... unless you want to add some verbosity...

pulpcore/app/models/fields.py
```diff
@@ -104,6 +145,38 @@ def from_db_value(self, value, expression, connection):
             return force_str(self._fernet.decrypt(force_bytes(value)))
 
 
+class EncryptedJSONField(JSONField):
```
mdellweg, line 148: I'm unsure if it is useful to derive this field from a JSON field if all we store is an encrypted string. Can't we just have it a text field and call json.loads and json.dumps in the db transformations? Or we could even derive from the existing `pulpcore.EncryptedField`!?

pulpcore/app/models/storage.py
```diff
@@ -164,6 +171,7 @@ def get_upload_chunk_file_path(pulp_id):
     return os.path.join(settings.CHUNKED_UPLOAD_DIR, str(pulp_id))
 
 
+# This is currently not used by anyone, probably should deprecate and remove from plugin api
```
mdellweg, line 174: yes, let's do it!

pulpcore/app/serializers/acs.py
```diff
@@ -57,6 +56,13 @@ class AlternateContentSourceSerializer(ModelSerializer):
         required=False,
     )
 
+    def validate_name(self, name):
+        validator = UniqueValidator(
```
mdellweg, line 60: We could implement a `UniquePerDomainValidator`, wdyt? Can we add validators for the domain field too?

This can be followup work.

pulpcore/app/serializers/base.py
```diff
@@ -41,13 +44,36 @@ def validate_unknown_fields(initial_data, defined_fields):
...
+                continue
+            elif not isinstance(lvalue, list):
+                lvalue = [lvalue]
+            for value in lvalue:
+                if isinstance(value, Model) and (
+                    domain_id := getattr(value, "pulp_domain_id", None)
+                ):
+                    if current_domain.pulp_id != domain_id:
+                        raise serializers.ValidationError(
+                            _("Objects must be all of the same domain.")
```
mdellweg, line 74: It's more like one objects domain does not match the requests domain now. Right?

pulpcore/app/serializers/domain.py
```diff
@@ -0,0 +1,291 @@
...
+    cache_control = serializers.CharField(allow_null=True, default=None)
+    sas_token = serializers.CharField(allow_null=True, default=None)
+    endpoint_suffix = serializers.CharField(default="core.windows.net")
+    custom_domain = serializers.CharField(allow_null=True, default=None)
+    connection_string = serializers.CharField(allow_null=True, default=None)
+    token_credential = serializers.CharField(allow_null=True, default=None)
+    api_version = serializers.CharField(allow_null=True, default=None)
+
+
+class DomainSettingsSerializer(serializers.Serializer):
```
mdellweg, line 197: Should this be called `StorageSettingsSerializer` ?

Does is reflect it's polymorphy correctly in the openapi spec?

 https://swagger.io/specification/#discriminator-object

pulpcore/app/serializers/repository.py
mdellweg: Another crazy idea: Should we change the Manager on all domained models to add the filter for domains already? We would get 404s instead of "Hey you are trying to use something from the other domain!". But i think this is even better in terms of protecting other tenants.

pulpcore/app/tasks/purge.py
```diff
@@ -69,8 +70,12 @@ def purge(finished_before, states):
 
     """
     current_user = get_current_authenticated_user()
-    # Tasks, prior to the specified date, in the specified state, owned by the current-user
-    tasks_qs = Task.objects.filter(finished_at__lt=finished_before, state__in=states)
+    domain = Task.current().pulp_domain
```
mdellweg, line 73: We should set the domain in the tasking stack so that we can use `util.get_domain()`.

pulpcore/backends.py
```diff
@@ -23,13 +28,27 @@ def has_perm(self, user_obj, perm, obj=None):
...
                 return True
             if GroupRole.objects.filter(
-                group__in=user_obj.groups.all(), object_id=None, role__permissions=permission
+                group__in=user_obj.groups.all(),
+                object_id=None,
+                domain=None,
+                role__permissions=permission,
+            ).exists():
+                return True
+        elif isinstance(obj, Domain) and "domain" not in perm:
```
mdellweg, line 42: shouldn't we check against `util.get_domain` as provided by the middleware?

pulpcore/openapi/__init__.py
```diff
@@ -476,6 +483,9 @@ def get_schema(self, request=None, public=False):
         for app in pulp_plugin_configs():
             result["info"]["x-pulp-app-versions"][app.label] = app.version
 
+        # Add domain-settings value
+        result["info"]["x-pulp-domain-enabled"] = settings.DOMAIN_ENABLED
```
mdellweg, line 487: :+1:

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -424,13 +427,15 @@ def _perform_task(task_pk, task_working_dir_rel_path):
         normal_thread.start()
     # All processes need to create their own postgres connection
     connection.connection = None
-    task = Task.objects.get(pk=task_pk)
+    task = Task.objects.select_related("pulp_domain").get(pk=task_pk)
```
mdellweg, line 430: :+1:

pulpcore/tests/functional/__init__.py
```diff
@@ -410,21 +424,6 @@ def __exit__(self, exc_type, exc_value, traceback):
...
-        return artifacts_api_client.create(temp_file)
-
-    return _random_artifact_factory
-
-
-@pytest.fixture
-def random_artifact(random_artifact_factory):
-    return random_artifact_factory()
-
-
```
mdellweg, line 427: These may be in use by plugin tests. At the very least i would keep the factory.

mdellweg (changes requested):
This is a BIG PR. Maybe i didn't see it, but i think the support for "REDIRECT_TO_OBJECT_STORAGE" feature is missing. Should this setting not move to the domains settings?

pulpcore/app/models/fields.py
gerrod3: I modified the field to encrypt each value instead of encrypting the whole field as a string. I could have made this an `EncryptedField` using json.loads and json.dumps, but we would lose the extra functionality of the JSONField.

pulpcore/app/serializers/acs.py
gerrod3: `UniquePerDomainValidator` is possible, I can look into it, but it should probably be done later as to not delay any further. What do you mean about validators for the domain field? There is no domain field on these serializers.

pulpcore/app/serializers/domain.py
gerrod3: Renamed it. Polymorphy is not correctly reflected in the spec. I couldn't figure out how to get the auto-gen spec code to recognize this serializer's polymorphic nature. Any ideas?

pulpcore/backends.py
gerrod3: This check is for when you are assigning a permission on a domain object, i.e. `core.view_domain`. This is how the backend knows if the permission is being assigned on a domain-level vs being assigned on a domain object. Checking against `get_domain` wouldn't provide any information.

pulpcore/responses.py
```diff
@@ -58,7 +58,11 @@ async def _sendfile(self, request, fobj, offset, count):
 
     async def prepare(self, request):
         if self._artifact is None:
-            self._artifact = await sync_to_async(Artifact.objects.get)(pk=self._artifact_pk)
+
+            def _get_artifact(pk):
+                return Artifact.objects.select_related("pulp_domain").get(pk=pk)
+
+            self._artifact = await sync_to_async(_get_artifact)(self._artifact_pk)
```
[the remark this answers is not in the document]
gerrod3: Yes, they are identical.

mdellweg:
> > This is a BIG PR. Maybe i didn't see it, but i think the support for "REDIRECT_TO_OBJECT_STORAGE" feature is missing. Should this setting not move to the domains settings?
> 
> Per domain, yes the feature is missing. Right now all domains would follow the global setting for `REDIRECT_TO_OBJECT_STORAGE`. I can add it to the domain's settings if we want domains to be able to control this setting.

I think, we need to. Some storages do not allow redirecting.

docs/plugin_dev/plugin-writer/concepts/rbac/access_policy.rst
```diff
@@ -23,25 +23,25 @@ Below is an example policy used by ``FileRemote``, with an explanation of its ef
...
             "effect": "allow",
-            "condition": "has_model_perms:file.add_fileremote",
+            "condition": "has_model_or_domain_perms:file.add_fileremote",
         },
         {
             "action": ["retrieve"],
             "principal": "authenticated",
             "effect": "allow",
-            "condition": "has_model_or_obj_perms:file.view_fileremote",
+            "condition": "has_model_domain_or_obj_perms:file.view_fileremote",
```
dkliban, line 32: s/model_domain/model_or_domain/

docs/workflows/domains-multi-tenancy.rst
```diff
@@ -0,0 +1,136 @@
...
+.. code-block::
+
+    # List repositories in 'test' domain
+    http :24817/pulp/test/api/v3/repositories/
+
+    # Create a File Repository 'foo' in 'foo' domain
+    http POST :24817/pulp/foo/api/v3/repositories/file/file/ name="foo"
+
+    # Create a File Repository 'foo' in 'boo' domain (Possible because of separate domains)
+    http POST :24817/pulp/boo/api/v3/repositories/file/file/ name="boo"
```
dkliban, line 91: s/boo/foo/

docs/workflows/domains-multi-tenancy.rst
```diff
@@ -0,0 +1,136 @@
...
+.. note::
+
+    Operations on resources across separate domains is not allowed. e.g. You can not add content
+    from one domain to the repository of another domain even if you own both domains.
+
+.. warning::
+
+    Pulp Export and Import are currently not supported with domains enabled.
+
+There are three notable objects in Pulp, ``AccessPolicies``, ``Roles`` and ``Users/Groups``, that
```
dkliban, line 130: s/three/four/ and split Users/Groups to be comma separated.

pulpcore/app/global_access_conditions.py
```diff
@@ -71,6 +84,27 @@ def has_obj_perms(request, view, action, permission):
...
+def has_model_or_domain_perms(request, view, action, permission):
+    """
+    Checks if the current user has either model-level (global) or domain-level permissions.
+    """
+    return has_model_perms(request, view, action, permission) or has_domain_perms(
+        request, view, action, permission
+    )
+
+
+def has_model_domain_or_obj_perms(request, view, action, permission):
```
dkliban, line 96: s/model_domain/model_or_domain/

pulpcore/app/global_access_conditions.py
```diff
@@ -106,7 +140,7 @@ def has_model_or_obj_perms(request, view, action, permission):
     )
 
 
-# 'Remote' parameter checks
+# 'Remote' parameter checks, should I deprecate the current model_or_obj checks?
```
dkliban, line 143: Remove this question.

pulpcore/app/models/fields.py
```diff
@@ -104,6 +104,50 @@ def from_db_value(self, value, expression, connection):
...
+    """A Field mixin that encrypts the JSON text using settings.DP_ENCRYPTION_KEY."""
+
+    def __init__(self, *args, **kwargs):
+        if kwargs.get("primary_key"):
+            raise ImproperlyConfigured("EncryptedTextField does not support primary_key=True.")
```
dkliban, line 112: s/EncryptedTextField/EncryptedJSONField/

pulpcore/app/models/publication.py
```diff
@@ -161,7 +164,8 @@ def delete(self, **kwargs):
 
                 # Invalidate cache for all distributions serving this publication
                 if base_paths:
-                    Cache().delete(base_key=base_paths)
+                    base_keys = [f"{self.pulp_domain.name}:{base_path}" for base_path in base_paths]
```
dkliban, line 167: Let's refactor this to not duplicate code.

pulpcore/app/serializers/base.py
```diff
@@ -433,14 +489,23 @@ def get_hidden_fields(
...
-            return field_value != "" and field_value is not None
+        def _is_set(field_name, default):
+            field_value = _get_attr(obj, field_name)
+            return field_value != "" and field_value is not None and field_value != default
 
         fields = self.get_fields()
         for field_name in fields:
-            if fields[field_name].write_only:
-                hidden_fields.append({"name": field_name, "is_set": _is_set(field_name)})
+            field = fields[field_name]
```
dkliban, line 505: Let's not worry about introspecting the write_only fields and keep the existing behavior.

pulpcore/app/serializers/domain.py
```diff
@@ -0,0 +1,376 @@
...
+from pulpcore.app import models
+from pulpcore.app.serializers import IdentityField, ModelSerializer, HiddenFieldsMixin
+
+
+BACKEND_CHOICES = (
+    ("pulpcore.app.models.storage.FileSystem", "Use local filesystem as storage"),
+    ("pulpcore.app.models.storage.PulpSFTPStorage", "Use SFTP server as storage"),
+    ("storages.backends.s3boto3.S3Boto3Storage", "Use Amazon S3 as storage"),
+    ("storages.backends.azure_storage.AzureStorage", "Use Azure Blob as storage"),
+    ("storages.backends.gcloud.GoogleCloudStorage", "Use Google Cloud as storage"),
```
dkliban, line 19: Please comment out the the google storage backend.

pulpcore/app/serializers/domain.py
```diff
@@ -0,0 +1,376 @@
...
+from rest_framework import serializers
+from rest_framework.validators import UniqueValidator
+
+from pulpcore.app import models
+from pulpcore.app.serializers import IdentityField, ModelSerializer, HiddenFieldsMixin
+
+
+BACKEND_CHOICES = (
+    ("pulpcore.app.models.storage.FileSystem", "Use local filesystem as storage"),
+    ("pulpcore.app.models.storage.PulpSFTPStorage", "Use SFTP server as storage"),
```
dkliban, line 16: We should hold off on having PulpSFTPStorage supported for now. Its future is uncertain.

pulpcore/app/serializers/publication.py
```diff
@@ -214,22 +213,25 @@ def _validate_path_overlap(self, path):
...
+                    "base_path": _("Overlaps with existing distribution '{}'").format(match.name)
+                },
             )
 
         return path
 
-    def validate_base_path(self, path):
-        self._validate_relative_path(path)
-        return self._validate_path_overlap(path)
+    def validate_base_path(self, value):
```
dkliban, line 231: let's change this parameter back to path.

pulpcore/app/settings.py
```diff
@@ -464,6 +467,6 @@
         connection.close()
 
 settings.set("V3_API_ROOT", settings.API_ROOT + "api/v3/")  # Not user configurable
-settings.set(
-    "V3_API_ROOT_NO_FRONT_SLASH", settings.V3_API_ROOT.lstrip("/")
-)  # Not user configurable
+settings.set("V3_DOMAIN_API_ROOT", settings.API_ROOT + "<pulp_domain>/api/v3/")
```
dkliban, line 470: should be this be the slug for the domain?

pulpcore/app/util.py
```diff
@@ -281,31 +298,66 @@ def get_artifact_url(artifact, headers=None, http_method=None):
...
-    elif settings.DEFAULT_FILE_STORAGE == "storages.backends.gcloud.GoogleCloudStorage":
+    elif artifact_domain.storage_class == "storages.backends.gcloud.GoogleCloudStorage":
         parameters = {"response_disposition": content_disposition}
         if headers and headers.get("Content-Type"):
             parameters["content_type"] = headers.get("Content-Type")
         url = artifact_file.storage.url(artifact_file.name, parameters=parameters)
     else:
         raise NotImplementedError(
-            f"The value settings.DEFAULT_FILE_STORAGE={settings.DEFAULT_FILE_STORAGE} "
+            f"The value settings.DEFAULT_FILE_STORAGE={artifact_domain.storage_class} "
```
dkliban, line 326: Let's add a condition check on domain enablement to determine what the error message should be. If domains are disabled it should give the old message.

pulpcore/app/views/orphans.py
```diff
@@ -19,6 +20,9 @@ def delete(self, request, format=None):
         """
         Cleans up all the Content and Artifact orphans in the system
         """
-        task = dispatch(orphan_cleanup, exclusive_resources=["/pulp/api/v3/orphans/cleanup/"])
+        uri = "/api/v3/orphans/cleanup/"
+        if settings.DOMAIN_ENABLED:
+            uri = f"/{request.pulp_domain.name}/api/v3/orphans/cleanup/"
```
dkliban, line 25: Maybe use string substitution?

pulpcore/app/viewsets/domain.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @extend_schema(
+        description="Trigger an asynchronous update task",
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    def update(self, request, pk, **kwargs):
+        """Prevent trying to update the default domain."""
+        instance = self.get_object()
+        if instance.name == "default":
+            raise NotAcceptable(_("Default domain can not be updated."))
```
dkliban, line 106: Let's switch this to a bad request.

mdellweg:
I have another issue, that i think may have been lost in the discussion. When switching on the domain feature in an existing system, the href's of a lot of objects change. I think, this is not reflected in the `reserved_resources` of the tasks. Do we have a plan for this?

pulpcore/app/migrations/0100_add_domain.py
```diff
@@ -0,0 +1,61 @@
...
+    END IF;
+    RETURN OLD;
+  END;
+$protect_default$ LANGUAGE plpgsql;
+CREATE TRIGGER protect_default BEFORE DELETE ON core_domain
+  FOR EACH ROW EXECUTE FUNCTION protect_default();
+"""
+
+REMOVE_DEFAULT_DELETE_TRIGGER = """
+DROP TRIGGER IF EXISTS protect_default ON core_domain;
```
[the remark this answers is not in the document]
gerrod3: @ggainey Do you know how these triggers are stored/associated in the database. I was under the impression the trigger is unique per table, so unique on the domain table.

pulpcore/app/models/fields.py
gerrod3: No, it only encrypts the values in the dictionary. The keys remain unencrypted. We get the benefit of using the has_keys and contains filtering methods: https://docs.djangoproject.com/en/4.1/topics/db/queries/#querying-jsonfield

pulpcore/app/models/storage.py
gerrod3: No, it hasn't been deprecated yet. Should be done in another PR.

pulpcore/app/serializers/base.py
```diff
@@ -433,14 +489,23 @@ def get_hidden_fields(
     ) -> List[TypedDict("hidden_fields", {"name": str, "is_set": bool})]:
         hidden_fields = []
 
+        def _get_attr(instance, attr):
+            if isinstance(instance, Mapping):
```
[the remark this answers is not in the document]
gerrod3: Not sure, I was copying the the internal code that DRF uses for its serializer fields when grabbing their values. I'm reverting this back as the extra complexity is not worth the tiny feature gain.

pulpcore/app/models/storage.py
```diff
@@ -122,6 +124,17 @@ def get_available_name(self, name, max_length=None):
...
 
+class DomainStorage:
+    """Storage class for FileFields to use that allows for uploading to a domain's storage."""
+
+    def __getattr__(self, item):
+        domain = get_domain()
+        storage = domain.get_storage()
+        return getattr(storage, item)
+
+    __class__ = Storage
```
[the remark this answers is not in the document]
gerrod3: Yes, if you talking about the `__class__ = Storage`  line, this is to bypass the isinstance check that Django does when instantiating a FileField. 

No, if you are talking about the `DomainStorage` class as a whole. This is just one of the many ways to implement the dynamic storage backend for the FileField. I particularly like this one for how clean it looks / simplicity, but if you look through the commit history you can see the other ways I've attempted before.

mdellweg (changes requested):
You have a discontinuity in the naming of the setting "DOMAIN_ENABLED" and "DOMAINS_ENABLED". Which one is it?

gerrod3:
> You have a discontinuity in the naming of the setting "DOMAIN_ENABLED" and "DOMAINS_ENABLED". Which one is it?

DOMAIN_ENABLED. Fixed.

pulpcore/app/settings.py
```diff
@@ -465,6 +465,6 @@
         connection.close()
 
 settings.set("V3_API_ROOT", settings.API_ROOT + "api/v3/")  # Not user configurable
-settings.set("V3_DOMAIN_API_ROOT", settings.API_ROOT + "<pulp_domain:slug>/api/v3/")
+settings.set("V3_DOMAIN_API_ROOT", settings.API_ROOT + "<slug:pulp_domain>/api/v3/")
```
mdellweg, line 468: Ha, it really is type:name...

dkliban (approved):

bmbouter:
image

PR #3211, opened 2022-09-15
Added a content_settings stanza to the /status/ output.

### LINKED ISSUE #3138: Add the URL of `content-app` to the pulp's status endpoint
Users who execute `pulp status` shall see the URL of content-app where all resources are served by a distribution. This might be useful for determining the URL of the Pulp Registry when using the container plugin.

ggainey:
closes #3138.

### DIFF
--- a/pulpcore/app/serializers/status.py
+++ b/pulpcore/app/serializers/status.py
@@ -49,6 +49,19 @@ class StorageSerializer(serializers.Serializer):
     free = serializers.IntegerField(min_value=0, help_text=_("Number of free bytes"))
 
 
+class ContentSettingsSerializer(serializers.Serializer):
+    """
+    Serializer for information about content-app-settings for the pulp instance
+    """
+
+    content_origin = serializers.CharField(
+        help_text=_("The CONTENT_ORIGIN setting for this Pulp instance"),
+    )
+    content_path_prefix = serializers.CharField(
+        help_text=_("The CONTENT_PATH_PREFIX setting for this Pulp instance"),
+    )
+
+
 class StatusSerializer(serializers.Serializer):
     """
     Serializer for the status information of the app
@@ -82,3 +95,5 @@ class StatusSerializer(serializers.Serializer):
     )
 
     storage = StorageSerializer(required=False, help_text=_("Storage information"))
+
+    content_settings = ContentSettingsSerializer(help_text=_("Content-app settings"))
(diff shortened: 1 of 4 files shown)


pulpcore/tests/functional/api/test_status.py
```diff
@@ -44,6 +44,10 @@
             },
         },
     },
+    "content_settings": {
+        "content_origin": "http://localhost:5001",
+        "content_path_prefix": "/pulp/content/",
+    },
```
lubosmj, line 50: This is not correct. It should contain JSON schema types, not constants.

lubosmj: Something like this:

```suggestion
    "content_settings": {
        "type": "object",
        "properties": {
            "content_origin":  {"type": "string"},
            "content_path_prefix": {"type": "string"},
        }
    },
```

ggainey: Ha, perfect! And this is why we do reviews :)

pulpcore/tests/functional/api/test_status.py
```diff
@@ -60,6 +75,7 @@ class StatusTestCase(unittest.TestCase):
 
     def setUp(self):
         """Make an API client."""
+        self.cfg = config.get_config()
         self.client = api.Client(config.get_config(), api.json_handler)
         self.status_response = STATUS
         cli_client = cli.Client(config.get_config())
```
mdellweg, line 81:
```suggestion
        self.client = api.Client(self.cfg, api.json_handler)
        self.status_response = STATUS
        cli_client = cli.Client(self.cfg)
```

mdellweg (approved):

pulpcore/app/serializers/status.py
```diff
@@ -49,6 +49,19 @@ class StorageSerializer(serializers.Serializer):
     free = serializers.IntegerField(min_value=0, help_text=_("Number of free bytes"))
 
 
+class ContentSettingsSerializer(serializers.Serializer):
+    """
+    Serializer for information about content-app-settings for the pulp instance
+    """
+
+    content_origin = serializers.CharField(
+        help_text=_("The CONTENT_ORIGIN setting for this Pulp instance"),
```
bmbouter, line 58:
```suggestion
        help_text=_("The CONTENT_ORIGIN setting value."),
```

bmbouter (approved):
Thanks!

PR #3221, opened 2022-09-18
Update CI files from 3.14 branch

pulpbot:
[noissue]


mdellweg:
This is using some old container i believe and the pulp-smash changes for s6 are breaking this.

PR #3222, opened 2022-09-18
Update CI files from 3.15 branch

pulpbot:
[noissue]


template_config.yml
```diff
@@ -57,6 +58,9 @@ pulp_settings:
   allowed_import_paths:
   - /tmp
   orphan_protection_time: 0
+pulp_settings_azure: null
+pulp_settings_s6: null
```
mdellweg, line 62: Please wait until this is overwritten by the correct name.

mdellweg (changes requested):

mdellweg (approved):

PR #3223, opened 2022-09-19
Migrate off pulp-smash

lubosmj:
[noissue]


pulpcore/tests/suite/utils.py
```diff
@@ -0,0 +1,371 @@
...
+from collections import defaultdict
+import unittest
+import warnings
+from urllib.parse import urljoin, urlsplit
+
+from pulpcore.tests.suite import api, cli, config
+from pulpcore.tests.suite.constants import STATUS_PATH
+
+
+def execute_pulpcore_python(cli_client, command):
```
bmbouter, line 15: We want to leave this behind, architecturally we don't have the system under test be "remote" anymore.

pulpcore/tests/suite/utils.py
```diff
@@ -0,0 +1,371 @@
...
+    :returns: A set of plugin names, e.g. ``{'pulpcore', 'pulp_file'}``.
+    """
+    if not cfg:
+        cfg = config.get_config()
+    client = api.Client(cfg, api.json_handler)
+    status = client.get(STATUS_PATH)
+    return {version["component"] for version in status["versions"]}
+
+
+def sync(cfg, remote, repo, **kwargs):
```
bmbouter, line 69: We should leave this behind, plugins can use it from pulp-smash as-is or they can port off of it to use a bindings based implementation instead of raw HTTP calls like this does.

pulpcore/tests/suite/utils.py
```diff
@@ -0,0 +1,371 @@
...
+    :returns: The server's response. A dict of information about the just
+        created sync.
+    """
+    client = api.Client(cfg)
+    data = {"remote": remote["pulp_href"]}
+    data.update(kwargs)
+    return client.post(urljoin(repo["pulp_href"], "sync/"), data)
+
+
+def modify_repo(cfg, repo, base_version=None, add_units=None, remove_units=None):
```
bmbouter, line 85: Same here, this is a raw HTTP call so we want to leave this behind.

pulpcore/tests/suite/utils.py
```diff
@@ -0,0 +1,371 @@
...
+            params["remove_content_units"] = remove_units
+        else:
+            params["remove_content_units"] = [content["pulp_href"] for content in remove_units]
+    if base_version:
+        params["base_version"] = base_version
+
+    return client.post(urljoin(repo["pulp_href"], "modify/"), params)
+
+
+def build_unit_url(_cfg, distribution, unit_path):
```
bmbouter, line 114: 90% of the stuff in the `suite` area need to be left behind I think. A big part of this has to do with converting tests from being raw HTTP call based to being bindings based. This work would be way too huge if you also had to convert all the tests, so I'd say:

* don't try to remove all pulp-smash imports from pulpcore, instead just bring in the pulp-smash pytest parts to pulpcore
* then have pulp_file start importing them from pulpcore instead of pulp-smash

This will leave pulp-smash imports in both pulpcore and pulp_file which may sound contrary to the goal, but the real goal is to remove the reliance on pulp-smash facilities regardless of where they live, so by moving them here's we're more or less just moving them.

lubosmj:
After thorough consideration, we have decided to migrate basic pytest fixtures from pulp-smash to pulpcore first. Then, we may continue migrating the rest of the pulp-smash utilities to pulpcore. In that case, raw HTTP `requests` calls will be replaced by bindings.

PR #3224, opened 2022-09-19
Prod telemetry site only receives production data

### LINKED ISSUE #3213: Telemetry of dev installs submitted to production site (https://anaytics.pulpproject.org/)
**Version**
3.20.z and 3.21.z

**Describe the bug**
Telemetry of a non-GA component is being submitted to analytics.pulpproject.org. This is unexpected and causes summarization to stop working with this exception:

```
  File "/srv/app/./manage.py", line 22, in <module>
    main()
  File "/srv/app/./manage.py", line 18, in main
  File "/opt/app-root/lib64/python3.9/site-packages/django/core/management/__init__.py", line 446, in execute_from_command_line
    utility.execute()
  File "/opt/app-root/lib64/python3.9/site-packages/django/core/management/__init__.py", line 440, in execute
  File "/opt/app-root/lib64/python3.9/site-packages/django/core/management/base.py", line 414, in run_from_argv
  File "/opt/app-root/lib64/python3.9/site-packages/django/core/management/base.py", line 460, in execute
  File "/srv/app/pulpanalytics/management/commands/summarize.py", line 112, in handle
    self._handle_components(systems, summary)
  File "/srv/app/pulpanalytics/management/commands/summarize.py", line 74, in _handle_components
    semver_version = semver.parse(component.version)
  File "/opt/app-root/lib64/python3.9/site-packages/semver.py", line 159, in wrapper
    return func(*args, **kwargs)
  File "/opt/app-root/lib64/python3.9/site-packages/semver.py", line 190, in parse
    return VersionInfo.parse(version).to_dict()
  File "/opt/app-root/lib64/python3.9/site-packages/semver.py", line 726, in parse
    raise ValueError("%s is not valid SemVer string" % version)
ValueError: 2.0.0a6 is not valid SemVer string
```

bmbouter:
This commit fixes a bug where installations with at least one pre-release plugin installed were posting to the production telemetry site instead of the developer telemetry site.

closes #3213

### DIFF
--- a/pulpcore/app/tasks/telemetry.py
+++ b/pulpcore/app/tasks/telemetry.py
@@ -23,9 +23,23 @@ DEV_URL = "https://dev.analytics.pulpproject.org/"
 
 
 def get_telemetry_posting_url():
+    """
+    Return either the dev or production telemetry FQDN url.
+
+    Production version string examples:  ["3.21.1", "1.11.0"]
+    Developer version string example: ["3.20.3.dev", "2.0.0a6"]
+
+    Returns:
+        The FQDN string of either the dev or production telemetry site.
+    """
     for app in pulp_plugin_configs():
-        if ".dev" in app.version:
+        if not app.version.count(".") == 2:  # Only two periods allowed in prod version strings
             return DEV_URL
+
+        x, y, z = app.version.split(".")
+        for item in [x, y, z]:
+            if not item.isdigit():  # Only numbers should be in the prod version string components
+                return DEV_URL
 
     return PRODUCTION_URL
 
--- a/CHANGES/3213.bugfix
+++ b/CHANGES/3213.bugfix
@@ -1,0 +1,2 @@
+Fixed bug where installations with at least one pre-release plugin installed were posting to the
+production telemetry site instead of the developer telemetry site.


bmbouter:
This is to handle plugins with pre-release names that don't end in `.dev`, e.g. `pulp-ostree 2.0.0a6`.

pulpcore/app/tasks/telemetry.py
```diff
@@ -24,9 +24,14 @@
 
 def get_telemetry_posting_url():
     for app in pulp_plugin_configs():
```
ggainey, line 26: Can we get a comment in here with an example of what broke us and/or what assumptions we're making?

ggainey (changes requested):

pulpcore/app/tasks/telemetry.py
bmbouter: I pushed a change, wdyt?

ggainey: :+1: 

I can imagine scenarios that will break this - but they'd all result in aiming at dev-url, which is better than dev-installs aiming at prod-url. I'm good w/ this change to fix the current problem.

ggainey (approved):
lg2m

mdellweg:
Can we please accompany this with a server-side change so the telemetry site does not choke on invalid input?

bmbouter:
Yes that got put in place first actually, but all it can do is drop the data. https://github.com/pulp/analytics.pulpproject.org/commit/b29eed4bcf88dd2ce8336fb97597e8b4410ddab5

PR #3239, opened 2022-09-22
Documented Pulpcore and Plugin release processes.

### LINKED ISSUE #3204: Release-process should exist in our docs somewhere
Our release-process has gotten way simpler, but also isn't written down anywhere in our 'official' docs. I have a quick writeup [in hackmd](https://hackmd.io/@ggainey/release-process), but we should have something in the official doc-tree.

ipanova:
closes #3204

### DIFF
--- a/docs/plugins/reference/releasing.rst
+++ b/docs/plugins/reference/releasing.rst
@@ -23,3 +23,34 @@ tested it against 3.7 and it's still compatible, use this requirements string::
 
 
     pulpcore>=3.5,<3.9
+
+
+Release process
+---------------
+
+Here are the steps to take to release a minor Plugin version, e.g. pulp_file 1.11.0:
+
+1. Via the Github Actions, trigger a `"Create new release branch" <https://github.com/pulp/pulpcore/actions/workflows/create-branch.yml>`_ job.
+2. Checkout locally the target plugin release branch and set accordingly ``pulpcore_branch`` and
+   ``pulpcore_pip_version_specifier`` in the template_config file.
+3. Pull in latest CI changes from the plugin_template. Ensure you have the latest copy of upstream
+   remote. ::
+
+    [user@localhost plugin_template]$ git remote -v
+    origin	git@github.com:user/plugin_template.git (fetch)
+    origin	git@github.com:user/plugin_template.git (push)
+    upstream	git@github.com:pulp/plugin_template.git (fetch)
+    upstream	git@github.com:pulp/plugin_template.git (push)
+    [user@localhost plugin_template]$ git branch
+    * main
+    [user@localhost plugin_template]$ git pull upstream main
+    [user@localhost plugin_template]$ ./plugin-template --github <plugin_name>
+
+   Make the PR against target plugin release branch and merge it.
+4. Via the Github Actions, trigger a `"Release pipeline" <https://github.com/pulp/pulpcore/actions/workflows/release.yml>`_ job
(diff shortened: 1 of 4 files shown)


docs/release_process.rst
```diff
@@ -0,0 +1,16 @@
+.. _release_process:
+
+Pulp Release Process
```
bmbouter, line 3:
```suggestion
pulpcore Release Process
```
Maybe just for clarity?

docs/release_process.rst
```diff
@@ -0,0 +1,16 @@
+.. _release_process:
```
bmbouter, line 1:
```suggestion
.. _pulpcore_release_process:
```

docs/release_process.rst
```diff
@@ -0,0 +1,16 @@
...
+
+Pulp Release Process
+====================
+
+Here are the steps to take to release a minor Pulpcore version:
+
+1. Via the Github Actions, trigger a "Create new release branch" job.
```
bmbouter, line 8: This could be an actual link and same with the line below.

docs/release_process.rst
```diff
@@ -0,0 +1,16 @@
...
+2. Via the Github Actions, trigger a "Release pipeline" job by specifying the release branch and the
+   tag of the release.
+3. Once the release is available, make an anouncement on the discourse.
```
bmbouter, line 11: This could link to discourse where the file it. Also it could at the end have something like `see [this example](http://example.com)`

docs/plugins/reference/releasing.rst
```diff
@@ -23,3 +23,28 @@ tested it against 3.7 and it's still compatible, use this requirements string::
...
+
+Release process
+---------------
+
+Here are the steps to take to release a minor Plugin version:
+
+1. Via the Github Actions, trigger a "Create new release branch" job.
+2. Checkout locally the target release branch and set accordingly ``pulpcore_branch`` and
+   ``pulpcore_pip_version_specifier`` in the template_config file.
+3. Pull in latest CI changes from the plugin template by executing this command
```
bmbouter, line 36: I think this needs to instruct the user to pull the latest changes from it also. Doing it with a `git pull` example assuming some reasonable remote name would be good too I think. This example could be in the block below.

ipanova: Seems redundant to but I will add some steps to be extra explicit

ggainey (approved):
lg2m, thanks!

bmbouter (approved):
Thanks!

docs/release_process.rst
```diff
@@ -0,0 +1,15 @@
+.. _pulpcore_release_process:
+
+Pulpcore Release Process
+========================
+
+Here are the steps to take to release a minor Pulpcore version, e.g. 3.21.0:
+
+1. Via the Github Actions, trigger a `"Create new release branch" <https://github.com/pulp/pulpcore/actions/workflows/create-branch.yml>`_ job.
+2. Via the Github Actions, trigger a `"Release pipeline" <https://github.com/pulp/pulpcore/actions/workflows/release.yml>`_ job
```
mdellweg, line 9: Shouldn't we adjust the pulp_file and pulp-certguard versions and update the template here?

bmbouter: Probably? I think @ipanova is out and I'm a little unclear on the steps to recommend.

PR #3243, opened 2022-09-23
Add expiry time to preauthed urls

### LINKED ISSUE #3238: preauthenticated urls should expire
The `ContentRedirectingContentGuard` is able to emit preauthenticated urls. The should have an expiry.

mdellweg:
fixes #3238

### DIFF
--- a/pulpcore/app/models/publication.py
+++ b/pulpcore/app/models/publication.py
@@ -1,12 +1,14 @@
 import hashlib
 import os
 import re
+from datetime import timedelta
 from url_normalize import url_normalize
 from urllib.parse import urlparse, urljoin
 
 from aiohttp.web_exceptions import HTTPNotFound
 
 from django.db import IntegrityError, models, transaction
+from django.utils import timezone
 from django_lifecycle import hook, AFTER_UPDATE, BEFORE_DELETE
 
 from .base import MasterModel, BaseModel
@@ -385,6 +387,8 @@ class ContentRedirectContentGuard(ContentGuard, AutoAddObjPermsMixin):
 
     TYPE = "content_redirect"
 
+    EXPIRETIME = timedelta(hours=1)
+
     shared_secret = models.BinaryField(max_length=32, default=_gen_secret)
 
     def permit(self, request):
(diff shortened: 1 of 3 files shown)


dkliban (commented):
What do you think about changing the name of the query parameter to `expiry` instead of `expire`?

mdellweg:
One question that came to me: Do we want the timeout to be configurable? As a setting? Or as a model field?

dkliban:
The other option is calling it `expires`.

mdellweg:
> It would be useful to make the expiration configurable. Who is expected to set this value? The REST API user or the admin? The plugin writer?

Now that is really the question. I can leave it like that for now and then change to either later. But once we have it as a setting, the ship really will have sailed. And with the REST API parameter it's the same the other way around.

pulpcore/app/models/publication.py
```diff
@@ -385,19 +386,24 @@ class ContentRedirectContentGuard(ContentGuard, AutoAddObjPermsMixin):
 
     TYPE = "content_redirect"
 
+    EXPIRETIME = timedelta(minutes=5)
```
ipanova, line 389: what was your judgement around the expiration time? can we set it to 1h? i looked into aws s3 presigned url and they expire by default in 1 hour.

mdellweg: It's about `pi * thumb`. Sure i can change that to 1h.

pulpcore/app/models/publication.py
```diff
@@ -385,19 +386,24 @@ class ContentRedirectContentGuard(ContentGuard, AutoAddObjPermsMixin):
...
+
     shared_secret = models.BinaryField(max_length=32, default=_gen_secret)
 
     def permit(self, request):
         """
         Permit preauthenticated redirects from pulp-api.
         """
         try:
+            expires = request.query["expires"]
+            if int(datetime.now().timestamp()) > int(expires):
```
ipanova, line 399: should we use instead `from django.utils.timezone import now`

pulpcore/app/models/publication.py
```diff
@@ -410,9 +416,14 @@ def preauthenticate_url(self, url, salt=None):
...
-        digest = self._get_digest(salt, url).hex()
-        url = url + f"?validate_token={hex_salt}:{digest}"
-        return url
+        expiretime = int((datetime.now() + self.EXPIRETIME).timestamp())
+        if "?" in url:
+            separator = "&"
+        else:
+            separator = "?"
+        url_to_sign = url + separator + f"expires={expiretime}"
+        digest = self._get_digest(salt, url_to_sign).hex()
```
ipanova, line 425: I am not sure we need to include the expiration data in the url to sign here.

mdellweg: We need that absolutely. It must not be tempered with. (There is even a test for it.)

pulpcore/app/models/publication.py
mdellweg: https://docs.djangoproject.com/en/3.2/topics/i18n/timezones/
This is hard to digest. But probably yes. So you cannot bypass the check by specifying a weird timezome.

ipanova (approved):
looks right to me!

dkliban (approved):

PR #3245, opened 2022-09-23
Fixed content-disposition header for the object storage backend.

### LINKED ISSUE #3124: Download package from my browser gives me a hash as package name
**Version**
 ```
"versions": [
    {
      "component": "core",
      "version": "3.20.0",
      "package": "pulpcore"
    },
    {
      "component": "rpm",
      "version": "3.17.7",
      "package": "pulp-rpm"
    },
    {
      "component": "file",
      "version": "1.10.3",
      "package": "pulp-file"
    },
    {
      "component": "container",
      "version": "2.13.0",
      "package": "pulp-container"
    },
    {
      "component": "ansible",
      "version": "0.14.0",
      "package": "pulp-ansible"
    }
  ],
```

**Describe the bug**
I therefore created a local repository to deposit the rpms built by our team, when I download a package from my browser (chrome or firefox) from this repository, the name of the downloaded file is a "hash" and not the name of the package, if I do the same test on a repository with a remote, I have no problem. Using S3 storage backend.

**To Reproduce**
1. create a repository without remote
2. upload a package to this repository
3. download package

Screenshot from 2022-08-23 16-20-18

### LINKED ISSUE #3244: Do not expose artifact digest via content-disposition header using Azure backend
**Version**
main and older

**Describe the bug**
When retrieving content from Azure backend `Content-disposition` header contains artifact's digest in the filename. 
 
**To Reproduce**
```
$ curl --head https://pulp3-source-fedora36.puffy.example.com/pulp/content/lukasku/Packages/t/tada.rpm -L
HTTP/1.1 302 Found
Server: nginx
Date: Fri, 23 Sep 2022 10:18:43 GMT
Content-Type: text/plain; charset=utf-8
Content-Length: 10
Connection: keep-alive
Location: https://pulp-azurite:10000/devstoreaccount1/pulp-test/pulp3/artifact/de/b341b9743f09c938277d645ec9f50d90143734e695cf58d14191c279ff6128?se=2022-09-23T10%3A20%3A43Z&sp=r&sv=2021-08-06&sr=b&rscd=attachment%253Bfilename%3Dartifact/de/b341b9743f09c938277d645ec9f50d90143734e695cf58d14191c279ff6128&sig=TYou/u4cjMat3kab1kCkynRruhGiYJrBDatXIB0cmaQ%3D

HTTP/1.1 200 OK
Server: Azurite-Blob/3.19.0
last-modified: Fri, 23 Sep 2022 10:10:38 GMT
x-ms-creation-time: Fri, 23 Sep 2022 10:10:38 GMT
x-ms-blob-type: BlockBlob
x-ms-lease-state: available
x-ms-lease-status: unlocked
content-length: 2494
content-type: application/octet-stream
etag: "0x21B8870A8293B60"
content-md5: PJ1thLs2HRCKNB4OsNW92A==
content-disposition: attachment%3Bfilename=artifact/de/b341b9743f09c938277d645ec9f50d90143734e695cf58d14191c279ff6128
x-ms-request-id: 6b3e8e1f-abb0-4902-8fc5-43cf81a2cd86
x-ms-version: 2021-10-04
date: Fri, 23 Sep 2022 10:18:43 GMT
accept-ranges: bytes
x-ms-server-encrypted: true
x-ms-access-tier: Hot
x-ms-access-tier-inferred: true
x-ms-access-tier-change-time: Fri, 23 Sep 2022 10:10:38 GMT
Connection: keep-alive
Keep-Alive: timeout=5

```

**Expected behavior**
Filename is user friendly and user readable and does not expose artifact digest.

### PR DESCRIPTION
**Pull request by ipanova**
closes #3124
closes #3244

### DIFF
--- a/CHANGES/3124.bugfix
+++ b/CHANGES/3124.bugfix
@@ -1,0 +1,1 @@
+Fixed content-disposition header which is used in the object storage backends.
--- a/pulpcore/app/util.py
+++ b/pulpcore/app/util.py
@@ -276,23 +276,24 @@ def get_artifact_url(artifact, headers=None, http_method=None):
     or private cloud storage.
     """
     artifact_file = artifact.file
+    content_disposition = f"attachment;filename={artifact.pk}"
     if (
         settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem"
         or not settings.REDIRECT_TO_OBJECT_STORAGE
     ):
         return _artifact_serving_distribution().artifact_url(artifact)
     elif settings.DEFAULT_FILE_STORAGE == "storages.backends.s3boto3.S3Boto3Storage":
-        parameters = {"ResponseContentDisposition": f"attachment%3Bfilename={artifact_file.name}"}
+        parameters = {"ResponseContentDisposition": content_disposition}
         if headers and headers.get("Content-Type"):
             parameters["ResponseContentType"] = headers.get("Content-Type")
         url = artifact_file.storage.url(
             artifact_file.name, parameters=parameters, http_method=http_method
         )
     elif settings.DEFAULT_FILE_STORAGE == "storages.backends.azure_storage.AzureStorage":
-        parameters = {"content_disposition": f"attachment%3Bfilename={artifact_file.name}"}
+        parameters = {"content_disposition": content_disposition}
         if headers and headers.get("Content-Type"):
             parameters["content_type"] = headers.get("Content-Type")
-        url = artifact_file.storage.url(artifact_file.name)
+        url = artifact_file.storage.url(artifact_file.name, parameters=parameters)
     else:
         raise NotImplementedError(
             f"The value settings.DEFAULT_FILE_STORAGE={settings.DEFAULT_FILE_STORAGE} "
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -810,6 +810,7 @@ class Handler:
         """
         artifact_file = content_artifact.artifact.file
         artifact_name = artifact_file.name
+        content_disposition = f"attachment;filename={content_artifact.relative_path}"
 
         if settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem":
             path = os.path.join(settings.MEDIA_ROOT, artifact_name)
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by dkliban** [comment]
@ipanova This change sets the filename to be the UUID of the artifact. I believe users would prefer it to be the last part of the relative path of the published artifact or content artifact.

**Comment by ipanova** [comment]
> @ipanova This change sets the filename to be the UUID of the artifact. I believe users would prefer it to be the last part of the relative path of the published artifact or content artifact.

It is only set for `get_artifact_url` which is supposed to be used by plugin writers and not by final users to see.
I meant to ask about this specific change @mdellweg

**Comment by dkliban** [comment]
Please disregard my comment above. I did not see all the changes earlier.

**Review by dkliban — APPROVED** [review]

File: pulpcore/content/handler.py [hunk 1]
```diff
@@ -810,6 +810,7 @@ async def _serve_content_artifact(self, content_artifact, headers, request):
         """
         artifact_file = content_artifact.artifact.file
         artifact_name = artifact_file.name
+        content_disposition = f"attachment;filename={content_artifact.relative_path}"
```
mdellweg, line 813: If I read this correctly, it reverts a change made because some reverse proxies did not accept headers with ";". How can we not break those use cases?

ipanova:
I've added a test and I need to figure out why azure test fails, switching to draft for now

pulpcore/app/util.py
```diff
@@ -276,23 +276,24 @@ def get_artifact_url(artifact, headers=None, http_method=None):
...
         url = artifact_file.storage.url(
             artifact_file.name, parameters=parameters, http_method=http_method
         )
     elif settings.DEFAULT_FILE_STORAGE == "storages.backends.azure_storage.AzureStorage":
-        parameters = {"content_disposition": f"attachment%3Bfilename={artifact_file.name}"}
+        parameters = {"content_disposition": content_disposition}
         if headers and headers.get("Content-Type"):
             parameters["content_type"] = headers.get("Content-Type")
-        url = artifact_file.storage.url(artifact_file.name)
+        url = artifact_file.storage.url(artifact_file.name, parameters=parameters)
```
ipanova, line 296: Azure test was failing because parameters were not passed.

ipanova:
I just realized that the disposition header fix will most likely be backported, so I am going to split this into 2 PRs because the artifact serving distribution is a new unreleased feature.

CHANGES/3124.bugfix
```diff
@@ -0,0 +1 @@
+Fixed content-disposition header  which is used in the object storage backends.
```
bmbouter, line 1:
```suggestion
Fixed content-disposition header which is used in the object storage backends.
```

pulpcore/content/handler.py
mdellweg: Thank you. :+1:

ipanova:
@jctanner @newswangerd  please specify the pulpcore branches you'd like this backported to

awcrosby:
@ipanova not sure if you got your question answered, but pulpcore 3.21 is the only branch where we should need this, as it was introduced in 3.20 but galaxy_ng doesn't use that. thanks!

ipanova:
@awcrosby yes thank you, there is already a manual cp to 3.21 https://github.com/pulp/pulpcore/pull/3273

PR #3246, opened 2022-09-23
Adds upper and lower bounds to almost all deps.

### LINKED ISSUE #3232: Add upper and lowe bounds to all pulpcore dependencies
Per [the new dependency docs](https://github.com/pulp/pulpcore/blob/main/docs/plugins/plugin-writer/concepts/index.rst#declaring-dependencies) we should add an upper and lower bound for all dependencies.

This does not need to be done for the functional and test dependencies, just the runtime pulp dependencies.

bmbouter:
All deps except Django (which we use the LTS only) received an upper and lower bound declaration. It also links to the dependencies that declare a semantic versioning policy.

closes #3232


requirements.txt
```diff
@@ -1,33 +1,33 @@
-asyncio-throttle~=1.0
-aiohttp~=3.8.1
-aiodns~=3.0.0
-aiofiles==22.1.0
-aioredis~=2.0.1
-backoff~=2.1.2
-click<9.0
-cryptography~=38.0.1
+asyncio-throttle>=1.0,<2.0
```
bmbouter, line 1: Only two dependencies declare a semver policy, yet we've been declaring forward compatible versions as if those projects did declare semver. For example we had `asyncio-throttle~=1.0` which now has an equivalent syntax of `asyncio-throttle>=1.0,<2.0` but that still doesn't seem right give then don't use semver.

So what to do?

I kind of think we should adjust [the policy](https://github.com/pulp/pulpcore/blob/main/docs/plugins/plugin-writer/concepts/index.rst#declaring-dependencies) to not even trust a project to "do it right". Even semver projects could release breaking changes at any time for bugfixes or security fixes.

So the policy in picking the upper bound would be "latest release" and let dependabot test everything as it comes out while it continues to bump it.

bmbouter: @pulp/core wdyt about this slight policy change in terms of the upper bound ^?

dkliban: I agree that we should upper bound with the latest release and rely on dependebot to test newer versions.

bmbouter: With this feedback I'm going to make some adjustments here.

bmbouter:
This is revised and ready for review.

bmbouter:
Here's the docs change to go with the non-consideration of a dependency's semver concern: https://github.com/pulp/pulpcore/pull/3248

mdellweg (approved):

requirements.txt
```diff
@@ -1,33 +1,33 @@
...
-django-currentuser~=0.5.3
-django-filter~=22.1
-django-guid~=3.3
-django-import-export~=2.9.0
-django-lifecycle~=1.0.0
-djangorestframework~=3.13.1
-djangorestframework-queryfields~=1.0.0
-drf-access-policy~=1.1.2
-drf-nested-routers==0.93.4
+django-currentuser>=0.5.3,<0.5.3
```
mdellweg, line 10:
```suggestion
django-currentuser>=0.5.3,<0.5.2
```
That was really unsatisfyable.

mdellweg (changes requested):

requirements.txt
```diff
@@ -1,33 +1,33 @@
...
+django-filter>=22.1,<=22.1
+django-guid>=3.3,<=3.3.0
+django-import-export>=2.9,<=2.9.0
+django-lifecycle>=1.0,<=1.0.0
+djangorestframework>=3.13.1,<=3.14.0
+djangorestframework-queryfields>=1.0,<=1.0.0
+drf-access-policy>=1.1.2,<=1.1.2
+drf-nested-routers>=0.93.4<=0.93.4
 drf-spectacular==0.23.1
```
mdellweg, line 19: While you are at it. Can you add a comment that we are still monkeypatching that library, which makes it impossible to have a looser requirement?

requirements.txt
ipanova: you meant `django-currentuser>=0.5.3,<=0.5.3`

requirements.txt
```diff
@@ -1,33 +1,33 @@
...
-redis~=4.3
-setuptools>=39.2.0,<66.0.0
-url-normalize~=1.4.3
-whitenoise>=5.0.0,<6.3.0
-yarl~=1.8
+dynaconf>=3.1.9,<=3.1.11
+gunicorn>=20.1,<=20.1.0
+jinja2>=3.1,<=3.1.2
+naya>=1.1.1,<=1.1.1
+protobuf<=4.21.6  # We don't know the lower bound
```
ipanova, line 24: since this is a recent addition to our deps and we have tested only with this version, does it make sense to then set it as the lower bound?

requirements.txt
mdellweg: Not really sure which format i prefer. They may be slightly different, but not so much that i really care. But most important: Are we sure that dependabot will never touch the lower bound and only ever try to push out the upper one?

bmbouter: I've been reading the docs for dependabot and that's my understanding. I'll be watching the dependabot PRs in the future for exactly this also.

requirements.txt
bmbouter: I can do that

requirements.txt
bmbouter: This one is tricky because it has a lot to do with compatability with the version the auto-generated code was built with which I think is in practice the older one. :/ It's not a good situation. Can I instead ask you to accept this for now and I'll commit to moving this out of our deps file entirely (and soon). It needs to become it's own python module so we're not checking in auto-generated code at all.

mdellweg (approved):

ipanova (approved):

PR #3248, opened 2022-09-23
Revises docs on plugin writer dep declaration

bmbouter:
The policy now doesn't consider the semver status of a dependency and instead recommends the upper bound be the latest compatible release. This allows dependabot to test each new release as it comes out.

[noissue]

### DIFF
--- a/docs/plugins/plugin-writer/concepts/index.rst
+++ b/docs/plugins/plugin-writer/concepts/index.rst
@@ -553,20 +553,21 @@ including declaring direct dependencies using the setuptools ``install_requires`
 Pulpcore and Pulp plugins are expected to do two things when declaring dependencies:
 
 1. Declare an upper bound to prevent a breaking-change release of a dependency from breaking user
-installations.
+installations. To prevent unexpected breakages due to new plugin releases, this typically is the
+current latest release of a dependency (assuming a plugin is compatible with the latest release).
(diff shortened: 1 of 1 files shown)


docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -553,20 +553,20 @@ including declaring direct dependencies using the setuptools ``install_requires`
 Pulpcore and Pulp plugins are expected to do two things when declaring dependencies:
 
 1. Declare an upper bound to prevent a breaking-change release of a dependency from breaking user
-installations.
+installations. To prevent unexpected breakages due to new plugin releases, this typically this is
```
ipanova, line 556: this typically is the current...

ipanova (approved):

docs/plugins/plugin-writer/concepts/index.rst
```diff
@@ -553,20 +553,20 @@ including declaring direct dependencies using the setuptools ``install_requires`
...
+the current latest release of a dependency (assuming a plugin is compatible with the latest
```
ipanova, line 557: i would add educational explanation of 'why latest release' summarize this comment https://github.com/pulp/pulpcore/pull/3246#discussion_r978821106

bmbouter: I added some.

mdellweg (approved):

PR #3250, opened 2022-09-23
Adds fixture for orphans_api_client

dkliban:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -21,6 +21,7 @@ from pulpcore.client.pulpcore import (
     GroupsUsersApi,
     ImportersPulpApi,
     ImportersPulpImportsApi,
+    OrphansApi,
     OrphansCleanupApi,
     PublicationsApi,
     RemotesApi,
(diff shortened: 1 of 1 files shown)


pulpcore/tests/functional/__init__.py
```diff
@@ -212,6 +213,11 @@ def redirect_contentguard_api_client(pulpcore_client):
     return ContentguardsContentRedirectApi(pulpcore_client)
 
 
+@pytest.fixture
+def orphans_api_client(pulpcore_client):
+    return OrphansApi(pulpcore_client)
```
mdellweg, line 218: That's the deprecated interface, no?
Can we leave a comment / docstring to not forget?

dkliban: yes it is. i made a comment about it in the test, but i'll add one here also.

mdellweg (approved):

pulpcore/tests/functional/__init__.py
bmbouter: Can we move this fixture to just the module where it's used? I don't think we want to make the deprecated one widely available. If you put the fixture up at the top where it's used it'll work.

dkliban: That's fine with me.

dkliban:
I am closing this PR in favor of adding it to the pulp_file test that is using it.

PR #3253, opened 2022-09-26
Add fixtures for filesystemexport API clients and cleanup

lubosmj:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -1,3 +1,5 @@
+import os
+import shutil
 import uuid
 
 import pytest
@@ -16,6 +18,8 @@ from pulpcore.client.pulpcore import (
     DistributionsApi,
     ExportersPulpApi,
     ExportersPulpExportsApi,
+    ExportersFilesystemApi,
+    ExportersFilesystemExportsApi,
     GroupsApi,
     GroupsRolesApi,
     GroupsUsersApi,
@@ -183,6 +187,16 @@ def exporters_pulp_exports_api_client(pulpcore_client):
 
 
 @pytest.fixture
+def exporters_filesystem_api_client(pulpcore_client):
+    return ExportersFilesystemApi(pulpcore_client)
+
+
+@pytest.fixture
+def exporters_filesystem_exports_api_client(pulpcore_client):
+    return ExportersFilesystemExportsApi(pulpcore_client)
+
+
+@pytest.fixture
 def importers_pulp_api_client(pulpcore_client):
     return ImportersPulpApi(pulpcore_client)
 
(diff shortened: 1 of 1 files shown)


dkliban (approved):

pulpcore/tests/functional/__init__.py
```diff
@@ -332,3 +344,17 @@ def _random_artifact_factory():
         return gen_object_with_cleanup(artifacts_api_client, temp_file)
 
     return _random_artifact_factory
+
+
+@pytest.fixture
+def add_to_disc_cleanup(cli_client):
```
mdellweg, line 350: Nice!

mdellweg (approved):

pulpcore/tests/functional/__init__.py
bmbouter, line 350:
```suggestion
def add_to_filesystem_cleanup(cli_client):
```

This reads a bit clearer to me (this is totally subjective).

pulpcore/tests/functional/__init__.py
```diff
@@ -332,3 +344,17 @@ def _random_artifact_factory():
...
+    obj_paths = []
+
+    def _add_to_disc_cleanup(path):
+        obj_paths.append(path)
+
+    yield _add_to_disc_cleanup
+
+    for p in obj_paths:
+        cmd = ("rm", "-rf", p)
+        cli_client.run(cmd, sudo=True)
```
bmbouter, line 360: What motivates the sudo here?

dkliban: That's how it was implemented in pulp_file. https://github.com/pulp/pulp_file/blob/a34514505f14cac4fd6617cf249c976ffcb75597/pulp_file/tests/functional/api/from_pulpcore/test_filesystemexport.py#L105-L107

lubosmj: The cleanup works without "sudo" locally. I am pushing the call without "sudo"...

pulpcore/tests/functional/__init__.py
```diff
@@ -332,3 +344,17 @@ def _random_artifact_factory():
         return gen_object_with_cleanup(artifacts_api_client, temp_file)
 
     return _random_artifact_factory
+
+
+@pytest.fixture
+def add_to_filesystem_cleanup(cli_client):
+    obj_paths = []
+
+    def _add_to_disc_cleanup(path):
```
bmbouter, line 353: nitpick: can this match the outer function name also?

pulpcore/tests/functional/__init__.py
```diff
@@ -332,3 +344,17 @@ def _random_artifact_factory():
...
+        obj_paths.append(path)
+
+    yield _add_to_disc_cleanup
+
+    for p in obj_paths:
```
bmbouter, line 358: can this `reverse()` to cause cleanup to occur in reverse order? The object cleanup works that way also.

pulpcore/tests/functional/__init__.py
```diff
@@ -332,3 +344,17 @@ def _random_artifact_factory():
...
+    obj_paths = []
+
+    def _add_to_filesystem_cleanup(path):
+        obj_paths.append(path)
+
+    yield _add_to_filesystem_cleanup
+
+    for p in reversed(obj_paths):
+        cmd = ("rm", "-rf", p)
+        cli_client.run(cmd)
```
bmbouter, line 360: I think we can just use Python instead of having a dependency on cli_client altogether right? Our tests now assume the system under test is localhost.

bmbouter (approved):
Thank you!

dkliban:
Thanks for all the revisions! This looks great!

pulpcore/tests/functional/__init__.py
mdellweg: That is one of the goals of consolidating here. If it's easy like here, let's do it right away.

pulpcore/tests/functional/__init__.py
mdellweg: For objects, we do for their interdependence. I don't think, we need for plain file paths. Also it does not hurt much.

PR #3258, opened 2022-09-28
Every task now requires logging cid in the same valid format

### LINKED ISSUE #3016: logging_cid in task response is an empty string
Sometimes the logging_cid in the task response is an empty string:

```
        {
            "child_tasks": [],
            "created_resources": [],
            "error": null,
            "finished_at": "2022-07-25T21:32:09.435330Z",
            "logging_cid": "",
            "name": "pulpcore.app.tasks.telemetry.post_telemetry",
            "parent_task": null,
            "progress_reports": [],
            "pulp_created": "2022-07-25T21:32:09.383372Z",
            "pulp_href": "/pulp/api/v3/tasks/7f4c54d6-fcda-4006-9da1-3c6910a18171/",
            "reserved_resources_record": [],
            "started_at": "2022-07-25T21:32:09.419785Z",
            "state": "completed",
            "task_group": null,
            "worker": "/pulp/api/v3/workers/c609a0d5-45cc-4756-af3c-4b50cd7520b1/"
        }
```

Since logging_cid represents a uuid, I'd expect it to be null or a uuid. I don't think an empty string makes sense. It also makes our code that parses out uuid more complex because we have handle the possibility of empty strings.

MichalPysik:
The logging_cid field can no longer be an empty string. If there is an attempt for creating such task, the task's guid will get automatically generated and set as the logging cid to keep consistency with the corrent format.

closes #3016


pulpcore/app/models/task.py
```diff
@@ -176,7 +177,7 @@ class Task(BaseModel, AutoAddObjPermsMixin):
 
     state = models.TextField(choices=TASK_CHOICES)
     name = models.TextField()
-    logging_cid = models.TextField(db_index=True, default="")
+    logging_cid = models.TextField(db_index=True, default=uuid4())
```
mdellweg, line 180: Can we even turn this into a uuid field?

MichalPysik: (This is just something I'm testing right now, since out pipeline gave me errors when there was no default)

mdellweg: Maybe setting this default is actually the fix (instead of using set_guid).

pulpcore/tasking/tasks.py
```diff
@@ -84,10 +85,12 @@ def dispatch(
         resources.extend(
             (f"shared:{resource}" for resource in _validate_and_get_resources(shared_resources))
         )
+    if not get_guid():
+        set_guid(uuid4())
```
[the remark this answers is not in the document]
mdellweg: I'd rather add the format constraint and the default at the database level.

mdellweg (changes requested):

bmbouter:
If we add this at the DB level and have the client code add it that could be great, but then again it could cause errors where uuid's get generated and not "set" from WSGI process hard to notice. To err on the side of correctness I think having it required at the DB level as a non-null field and having client code set it would be the safest. @mdellweg @MichalPysik wdyt?

pulpcore/app/models/task.py
```diff
@@ -176,7 +177,7 @@ class Task(BaseModel, AutoAddObjPermsMixin):
 
     state = models.TextField(choices=TASK_CHOICES)
     name = models.TextField()
-    logging_cid = models.TextField(db_index=True, default="")
+    logging_cid = models.TextField(db_index=True, default=None, validators=[MinLengthValidator(32)])
```
mdellweg, line 180: I think this should be a UUIDField.

https://docs.djangoproject.com/en/3.2/ref/models/fields/#uuidfield

CHANGES/3016.bugfix
```diff
@@ -0,0 +1 @@
+The logging_cid field of a task can no longer be an empty string. If there is an attempt to create a task with an empty one, a new guid of the given task will be generated and set as the logging cid, since now every task requires a valid logging cid in the same format, with no exceptions.
```
mdellweg, line 1: This needs wrapping at 98 chars.

pulpcore/app/migrations/0096_alter_task_logging_cid.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('core', '0095_artifactdistribution'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='task',
+            name='logging_cid',
+            field=models.UUIDField(),
```
mdellweg, line 16: I'm wondering what happens to existing empty records here.
Can you check that old tasks with "" will not break the migration here?

bmbouter: I agree. We probably need a migration to auto-assign them at migration time. We normally would have the `default` defined on the field handle that, but since we don't want one of those (see earlier) comment I think we need a migration.

MichalPysik: Ok, we have a problem - when trying to migrate an object with logging_cid = "", it fails to migrate, since this raises a syntax error

pulpcore/app/migrations/0096_alter_task_logging_cid.py
```diff
@@ -0,0 +1,33 @@
...
+            task.save()
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0095_artifactdistribution'),
+    ]
+
+    operations = [
+        migrations.RunPython(migrate_empty_string_logging_cid),
```
mdellweg, line 27:
```suggestion
        migrations.RunPython(code=migrate_empty_string_logging_cid, reverse_code=migrations.RunPython.noop),
```

pulpcore/app/migrations/0096_alter_task_logging_cid.py
```diff
@@ -0,0 +1,33 @@
...
+        migrations.AlterField(
+            model_name='task',
+            name='logging_cid',
+            field=models.UUIDField(default=uuid4()),
```
mdellweg, line 31: We agreed not to have a default here, no?

pulpcore/app/migrations/0096_alter_task_logging_cid.py
```diff
@@ -0,0 +1,33 @@
...
+
+def migrate_empty_string_logging_cid(apps, schema_editor):
+
+    Task = apps.get_model('core', 'Task')
+
+    for task in Task.objects.all():
+
+        logging_cid = task.logging_cid
+        try:
+            _ = UUID(logging_cid, version=4)
```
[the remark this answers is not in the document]
mdellweg:
```suggestion
            UUID(logging_cid, version=4)
```

pulpcore/app/models/task.py
```diff
@@ -6,6 +6,7 @@
 import os
 from datetime import timedelta
 from gettext import gettext as _
+from uuid import uuid4
```
ipanova, line 9: should we use https://github.com/pulp/pulpcore/pull/3178/files#diff-5ae6f5b8d062967913215abb1a014e66aa5fdc5a034be4bbeae2167a097cf605R12

pulpcore/app/migrations/0096_alter_task_logging_cid.py
```diff
@@ -0,0 +1,33 @@
...
+
+from django.db import migrations, models
+from uuid import uuid4, UUID
+
+
+def migrate_empty_string_logging_cid(apps, schema_editor):
+
+    Task = apps.get_model('core', 'Task')
+
+    for task in Task.objects.all():
```
ipanova, line 11: System can potentially have a *lot* of task records, can we loop only through those that have logging_cid='""?
There is a possibility however that existing Task logging_cid might have some other format that uuid( text field would allow that)  but it should not be the case unless the db record was manually altered.

ipanova (changes requested):
I don't think we can enforce UUID field here nor alternate existing records that might not adhere to uuid4 format.
Following these config docs one can 1) disable format validation 2) change the default length
https://django-guid.readthedocs.io/en/latest/settings.html#settings

If we set these options, we can also assume that they cannot be changed. (I think there are other options an admin could change that would break pulp's functionality.)

pulpcore/app/models/task.py
mdellweg: This may be useful, as we may have/add at least a search index on the field. TBH this is the exceptional case where we create a cid so the impact may be minimal.

pulpcore/app/migrations/0096_alter_task_logging_cid.py
MichalPysik: I believe we said afterwards that we want this instead of the "if not get guid(): set guid()" code inside the task.dispatch()

pulpcore/tasking/tasks.py
mdellweg, line 89: This may not be helping us to find cases where there should be a guid, but isn't.

Can we move the `set_guid` call to the place where we call `dispatch` without being in a request context?

mdellweg (changes requested):

pulpcore/tasking/tasks.py
MichalPysik: Is the current state ok, or should it move to a different file?

pulpcore/app/models/task.py
```diff
@@ -176,7 +176,7 @@ class Task(BaseModel, AutoAddObjPermsMixin):
 
     state = models.TextField(choices=TASK_CHOICES)
     name = models.TextField()
-    logging_cid = models.TextField(db_index=True, default="")
+    logging_cid = models.UUIDField()
```
mdellweg, line 179: Let's keep the db_index. We use that field for searching.

ipanova: @mdellweg  based on this comment https://github.com/pulp/pulpcore/pull/3258#pullrequestreview-1134457731 we should keep textfield, you have same conclusion?

mdellweg: The current state of a affairs is that pulp will happily accept everything you pass in (as the correlation-id header) that will transform nicely when passed into `uuid.UUID` (i.e. any hex 16 byte string interweaved with dashes).
For all other input it will silently generate a random new one. So i still think we are safe to force that format on the db level.

pulpcore/app/migrations/0096_alter_task_logging_cid.py
```diff
@@ -0,0 +1,31 @@
...
+from django.db import migrations, models
+from uuid import uuid4, UUID
+
+
+def migrate_empty_string_logging_cid(apps, schema_editor):
+
+    Task = apps.get_model("core", "Task")
+
+    for task in Task.objects.filter(logging_cid=""):
+        task.logging_cid = uuid4()
```
mdellweg, line 12: We may need to use (here and elsewhere):
https://github.com/snok/django-guid/blob/6b4f951f8ee363ce2f65611040e8812352c8db11/django_guid/utils.py#L67

pulpcore/app/models/task.py
```diff
@@ -176,7 +176,7 @@ class Task(BaseModel, AutoAddObjPermsMixin):
 
     state = models.TextField(choices=TASK_CHOICES)
     name = models.TextField()
-    logging_cid = models.TextField(db_index=True, default="")
+    logging_cid = models.UUIDField(db_index=True)
```
bmbouter, line 179: The underlying dependency allows for non UUID based strings to be put here so I think we should just have this remove the `default=""` to make it a DB required string. @mdellweg @ipanova is this the heart of the issue we quickly discussed this morning?

mdellweg: Yes it is. We may need to add `blank=False` if we don't want to allow the empty string.

BTW: I was unable to configure the underlying dependency to use something different.

bmbouter: Good to hear re the configuration of the underlying dependency. I think we want `null=False` and `blank=False` which are [the defaults](https://docs.djangoproject.com/en/3.2/ref/models/fields/#null) so removing the `default=""` should be enough (I think).

mdellweg (approved):

ipanova (approved):
looks right to me! thank you!

PR #3261, opened 2022-09-29
Separated the user plugin vs plugin writer docs

### LINKED ISSUE #3260: Make plugin developer docs more clearly labeled
There are two small issues currently:

* The user facing "list of plugins" page is stored in the source tree with the docs that are really for plugin writers
* The plugin writer docs are under a heading called "Plugin API", but in practice this contains way more than just the API so it should be renamed.

bmbouter:
closes #3260


docs/glossary.rst
```diff
@@ -51,7 +51,7 @@ Glossary
 
     pulpcore
         A python package offering users a :doc:`rest_api` and plugin writers a
-        :ref:`Plugin API`. It is :term:`plugin`-based and manages :term:`Content`.
+        :ref:`plugin_api`. It is :term:`plugin`-based and manages :term:`Content`.
```
dralley, line 54: Should we say more?  The pulpcore API also helps with managing repository versions and so forth.

dralley (approved):

docs/glossary.rst
bmbouter: We should, but in another piece of work. I'm hoping to spend more time with our docs as part of my contribution.

docs/index.rst
```diff
@@ -21,9 +21,9 @@ If you want to look at the considerations and requirements for installing Pulp,
  :ref:`installation`. If you want to evaluate Pulp quickly, try `Pulp in One
  Container <https://pulpproject.org/pulp-in-one-container/>`_
 
-If you're looking for documentation specific to a content type, see :doc:`List of Plugins<plugins/index>`
+If you're looking for documentation specific to a content type, see :doc:`List of Plugins<plugins>`
 
-Anyone interested in writing a plugin should reference the :ref:`Plugin API`.
+Anyone interested in writing a plugin should reference the :ref:`plugin_development`.
```
ggainey, line 26:
```suggestion
Anyone interested in writing a plugin should reference :ref:`plugin_development`.
```

ggainey (changes requested):
One minor tweak, otherwise this is great!

ggainey (approved):
lg2m!

mdellweg:
Please rebase onto the latest CI fixes.

PR #3268, opened 2022-09-30
Add ordering filter to NamedModelViewSet

### LINKED ISSUE #3266: ordering filter not available on all viewsets
Adding prio-list, because this is an inconsistency across our rest api.

mdellweg:
fixes #3266

### DIFF
--- a/CHANGES/3266.bugfix
+++ b/CHANGES/3266.bugfix
@@ -1,0 +1,1 @@
+Made ordering consistent between all list API endpoints.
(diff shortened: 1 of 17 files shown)


CHANGES/3266.bugfix
```diff
@@ -0,0 +1 @@
+Added ``BaseFilterSet`` to ``NamedModelViewSet`` to make ordering possible on all list endpoints.
```
dkliban, line 1: Since this is a REST API user facing message, you should avoid using the Plugin API language. `Made ordering consistent between all list API endpoints. `

pulpcore/filter.py
```diff
@@ -0,0 +1,142 @@
+from gettext import gettext as _
```
dkliban, line 1: This module should be plural `pulpcore.filters`.

dkliban (changes requested):
I added a couple of suggested changes.

dkliban (approved):

PR #3281, opened 2022-10-05
Add a fixture for RepositoriesReclaimSpaceApi and download_content

lubosmj:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -1,3 +1,5 @@
+import aiohttp
+import asyncio
 import os
 import shutil
 import uuid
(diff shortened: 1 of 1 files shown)


mdellweg (approved):

gerrod3 (approved):

pulpcore/tests/functional/__init__.py
```diff
@@ -372,3 +380,23 @@ def _add_to_filesystem_cleanup(path):
             except OSError:
                 # the file may no longer exist, but we do not care
                 pass
+
+
+@pytest.fixture
+def download_content_unit(pulp_cfg):
```
dkliban, line 386: This fixture is not needed. We already have it in pulp_file[0].

[0] https://github.com/pulp/pulp_file/blob/main/pulp_file/tests/functional/utils.py#L315-L323

PR #3282, opened 2022-10-05
Added reserved_resources filters to task list

### LINKED ISSUE #3280: task filter reserved_resources_record does not work as expected
The task filter named `reserved_resources_record` specifies to be of type `array` in the openapi spec. But it only filters by the last entry in the list. Instead it should filter tasks that lock on all filtered resources.

Additionally, the filter is not aware of the `shared:` prefix.

mdellweg:
These new filters search for both shared and exclusive resources.

fixes #3280

### DIFF
--- a/pulpcore/app/viewsets/custom_filters.py
+++ b/pulpcore/app/viewsets/custom_filters.py
@@ -22,7 +22,58 @@ from pulpcore.app.viewsets import NamedModelViewSet
 
 class ReservedResourcesFilter(Filter):
     """
-    Enables a user to filter tasks by a reserved resource href
+    Enables a user to filter tasks by a reserved resource href.
+    """
+
+    def __init__(self, *args, exclusive=True, shared=True, **kwargs):
+        self.exclusive = exclusive
+        self.shared = shared
+        assert (
+            exclusive or shared
+        ), "ReservedResourceFilter must have either exclusive or shared set."
+        super().__init__(*args, **kwargs)
+
+    def filter(self, qs, value):
+        """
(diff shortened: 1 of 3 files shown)


pulpcore/app/viewsets/custom_filters.py
```diff
@@ -22,7 +22,56 @@
...
+        assert (
+            exclusive or shared
+        ), "ReservedResourceFilter must have either exclusive or shared set."
+        super().__init__(*args, **kwargs)
+
+    def filter(self, qs, value):
+        """
+        Args:
+            qs (django.db.models.query.QuerySet): The Queryset to filter
+            value (string): href containing a reference to a reserved resource
```
ggainey, line 40: This can now be str or [str,...] yes?

ggainey (changes requested):
Doc change needed (I believe). Can we get a pulp_file PR to add tests for this?

pulpcore/app/viewsets/custom_filters.py
mdellweg: Yes, it can. It will be `str` if the filter is used as is, and `List[str]` if mixed with `BaseInFilter`.
The docstring of this function is kind of questionable, as it is not meant to be called directly at all, but only as a callback for the filtering backend. Will adjust it anyway.

ggainey (approved):
lg2m

dkliban:
@mdellweg what do you think about adding some tests to `pulp_file` for this feature?

dkliban (approved):

PR #3293, opened 2022-10-08
Convert tests to use bindings & pytest style

gerrod3:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -41,6 +41,7 @@ from pulpcore.client.pulpcore import (
     UploadsApi,
     UsersApi,
     UsersRolesApi,
+    WorkersApi,
 )
 
 from .gpg_ascii_armor_signing_service import (  # noqa: F401
@@ -102,6 +103,11 @@ def pulpcore_client(cid, bindings_cfg):
 @pytest.fixture
 def tasks_api_client(pulpcore_client):
     return TasksApi(pulpcore_client)
+
+
+@pytest.fixture
+def workers_api_client(pulpcore_client):
+    return WorkersApi(pulpcore_client)
 
 
 @pytest.fixture
@@ -306,6 +312,23 @@ def anonymous_user(bindings_cfg):
 
 
 @pytest.fixture(scope="session")
+def invalid_user(bindings_cfg):
+    class InvalidUser:
+        def __init__(self):
+            self._saved_credentials = []
+
+        def __enter__(self):
+            self._saved_credentials.append((bindings_cfg.username, bindings_cfg.password))
+            bindings_cfg.username, bindings_cfg.password = str(uuid.uuid4()), str(uuid.uuid4())
+            return self
+
+        def __exit__(self, exc_type, exc_value, traceback):
+            bindings_cfg.username, bindings_cfg.password = self._saved_credentials.pop()
+
+    return InvalidUser()
+
+
+@pytest.fixture(scope="session")
 def pulp_admin_user(bindings_cfg):
     class AdminUser:
         def __init__(self):
@@ -345,6 +368,13 @@ def pulp_api_v3_path(cli_client):
 @pytest.fixture(scope="session")
 def pulp_api_v3_url(pulp_cfg, pulp_api_v3_path):
     return f"{pulp_cfg.get_base_url()}{pulp_api_v3_path}"
+
+
+@pytest.fixture
+def get_redis_status(status_api_client):
+    """Return a boolean value which tells whether the connection to redis was established or not."""
+    status_response = status_api_client.status_read()
+    return status_response.redis_connection.connected
 
 
 @pytest.fixture
--- a/pulpcore/tests/functional/api/test_api_docs.py
+++ b/pulpcore/tests/functional/api/test_api_docs.py
@@ -1,43 +1,41 @@
-"""Test related to the api docs page."""
-import unittest
-
-from pulp_smash import api, config
-from pulp_smash.pulp3.constants import API_DOCS_PATH
-from requests.exceptions import HTTPError
+"""Tests related to the api docs page."""
+import pytest
+from pulpcore.client.pulpcore import ApiException
 
 
-class ApiDocsTestCase(unittest.TestCase):
-    """Test whether API auto generated docs are available.
+@pytest.fixture(scope="session")
+def pulp_docs_url(pulp_api_v3_url):
+    return f"{pulp_api_v3_url}docs/"
 
-    This test targets the following issue:
 
-    * `Pulp Smash #893 <https://github.com/pulp/pulp-smash/issues/893>`_
+@pytest.mark.parallel
+def test_valid_credentials(pulpcore_client, pulp_docs_url):
+    """Get API documentation with valid credentials.
+
+    Assert the API documentation is returned.
     """
+    response = pulpcore_client.request("GET", pulp_docs_url)
+    assert response.status == 200
 
-    def setUp(self):
-        """Create an API Client."""
-        cfg = config.get_config()
-        self.client = api.Client(cfg)
 
-    def test_valid_credentials(self):
-        """Get API documentation with valid credentials.
+@pytest.mark.parallel
+def test_no_credentials(pulpcore_client, pulp_docs_url, anonymous_user):
+    """Get API documentation with no credentials.
 
-        Assert the API documentation is returned.
-        """
-        self.client.get(API_DOCS_PATH)
+    Assert the API documentation is returned.
+    """
+    with anonymous_user:
+        response = pulpcore_client.request("GET", pulp_docs_url)
+        assert response.status == 200
 
-    def test_no_credentials(self):
-        """Get API documentation with no credentials.
 
-        Assert the API documentation is returned.
-        """
-        del self.client.request_kwargs["auth"]
-        self.client.get(API_DOCS_PATH)
+@pytest.mark.parallel
+def test_http_method(pulpcore_client, pulp_docs_url):
+    """Get API documentation with an HTTP method other than GET.
 
-    def test_http_method(self):
-        """Get API documentation with an HTTP method other than GET.
+    Assert an error is returned.
+    """
+    with pytest.raises(ApiException) as e:
+        pulpcore_client.request("POST", pulp_docs_url)
 
-        Assert an error is returned.
-        """
-        with self.assertRaises(HTTPError):
-            self.client.post(API_DOCS_PATH)
+    assert e.value.status == 405
--- a/pulpcore/tests/functional/api/test_crd_artifacts.py
+++ b/pulpcore/tests/functional/api/test_crd_artifacts.py
@@ -2,179 +2,159 @@
 import hashlib
 import itertools
 import os
-import unittest
+import uuid
+import pytest
 
-from pulp_smash import api, cli, config, utils
-from pulp_smash.exceptions import CalledProcessError
-from pulp_smash.pulp3.bindings import delete_orphans
-from pulp_smash.pulp3.constants import ARTIFACTS_PATH
-from requests.exceptions import HTTPError
+from django.conf import settings
+from pulpcore.client.pulpcore import ApiException
 
 
-FILE_URL = "https://fixtures.pulpproject.org/file/1.iso"
+@pytest.fixture
+def pulpcore_random_file(tmp_path):
+    name = tmp_path / str(uuid.uuid4())
+    with open(name, "wb") as fout:
+        contents = os.urandom(1024)
+        fout.write(contents)
+        fout.flush()
+    digest = hashlib.sha256(contents).hexdigest()
+    return {"name": name, "size": 1024, "digest": digest}
 
 
-class ArtifactTestCase(unittest.TestCase):
-    """Create an artifact by uploading a file.
+def _do_upload_valid_attrs(artifact_api, file, data):
+    """Upload a file with the given attributes."""
+    artifact = artifact_api.create(file, **data)
+    # assumes ALLOWED_CONTENT_CHECKSUMS does NOT contain "md5"
+    assert artifact.md5 is None, "MD5 {}".format(artifact.md5)
+    read_artifact = artifact_api.read(artifact.pulp_href)
+    # assumes ALLOWED_CONTENT_CHECKSUMS does NOT contain "md5"
+    assert read_artifact.md5 is None
+    for key, val in artifact.to_dict().items():
+        assert getattr(read_artifact, key) == val
+    # Delete the Artifact
+    artifact_api.delete(read_artifact.pulp_href)
+    with pytest.raises(ApiException) as e:
+        artifact_api.read(read_artifact.pulp_href)
 
-    This test targets the following issues:
+    assert e.value.status == 404
 
-    * `Pulp #2843 <https://pulp.plan.io/issues/2843>`_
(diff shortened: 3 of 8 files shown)


pulpcore/tests/functional/api/test_api_docs.py
```diff
@@ -1,43 +1,43 @@
-"""Test related to the api docs page."""
-import unittest
+"""Tests related to the api docs page.
 
-from pulp_smash import api, config
-from pulp_smash.pulp3.constants import API_DOCS_PATH
-from requests.exceptions import HTTPError
+These tests target the following issue:
 
+* `Pulp Smash #893 <https://github.com/pulp/pulp-smash/issues/893>`_
```
bmbouter, line 5: Stylistically we don't keep links to issues in the test docstrings anymore.

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -2,179 +2,171 @@
...
+@pytest.fixture
+def pulpcore_random_file(tmp_path):
+    name = tmp_path / str(uuid.uuid4())
+    with open(name, "wb") as fout:
+        contents = os.urandom(1024)
+        fout.write(contents)
+        fout.flush()
+    digest = hashlib.sha256(contents).hexdigest()
+    return {"name": name, "size": 1024, "digest": digest}
+    # Should I clean up the temporary file?
```
bmbouter, line 22: No pytest does this for you.

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -2,179 +2,171 @@
 import hashlib
 import itertools
 import os
-import unittest
+import uuid
+import pytest
 
-from pulp_smash import api, cli, config, utils
+from pulp_smash.utils import get_pulp_setting
```
bmbouter, line 8: @lubosmj @mdellweg @gerrod3 @dkliban I think we need to bring over the ability to `get_pulp_setting` to pulpcore natively. I believe the assumption now and going forward is that is on the local system also so the implementation moved to pulpcore only needs to consider that. Is that you expectation as well?

pulpcore/tests/functional/api/test_upload.py
```diff
@@ -1,154 +1,190 @@
...
 import hashlib
-import unittest
+import uuid
+import pytest
+import os
 
 from random import shuffle
-from requests import HTTPError
-from urllib.parse import urljoin
+from pulp_smash.pulp3.bindings import monitor_task
```
bmbouter, line 8: @gerrod3 @dkliban @lubosmj @mdellweg we should move `monitor_task` over to pulpcore also I think as a utility function in https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/utils.py

pulpcore/tests/functional/api/test_workers.py
```diff
@@ -1,105 +1,94 @@
...
-                "last_heartbeat__gte": self.worker["last_heartbeat"],
-                "name": self.worker["name"],
-            },
-        )
-        self.assertEqual(
-            len(page["results"]), 1, "Expected: {}. Got: {}.".format([self.worker], page["results"])
-        )
-        for key, val in self.worker.items():
+def test_worker_actions(workers_api_client):
+    worker = _test_01_read_all_workers(workers_api_client)
```
bmbouter, line 15: Not a required change, but stylistically we've been making this one long test and using single-line comments blocking out each separate section. Here is an example of that:  https://github.com/pulp/pulp_file/pull/830/files#diff-6c2b39e46a76f777fb2a3c2fe06d98e8654db70525351c5a9d136f889107ef67R94

bmbouter (changes requested):
@gerrod3 ty so much for working on this! This is all looking really good. What do you think about the requested change I've written in comments?

pulpcore/tests/functional/api/test_crd_artifacts.py
lubosmj: We should move both the `get_pulp_setting` and `monitor task` functions (and many others) to pulpcore. However, I am not sure if it needs to be done in this PR.

dkliban: A test should just `from pulpcore.app import settings` to check pulp settings.

bmbouter: @gerrod3 then please just move the `monitor_task` fixture to pulpcore. Does that sound ok?

gerrod3: Ok, I'll move `monitor_task` to pulpcore. I'll also try using `from pulpcore.app import settings`, but I'm not sure it will work.

pulpcore/tests/functional/api/test_api_docs.py
gerrod3: Removed.

pulpcore/tests/functional/api/test_crd_artifacts.py
gerrod3: `monitor_task` and `monitor_task_group` is now in pulpcore. Importing settings directly from pulpcore did not work as not all the settings were initialized.

lubosmj: @dkliban, @gerrod3, importing the settings directly from pulpcore works in pulp_file (because the pulpcore module is correctly initialized before executing the test?): https://github.com/pulp/pulp_file/pull/842/files#diff-d40a0ce852ae1c631a07c83021e5d9e9be0b4de32c98b7d2bb34a9aff4922824R13.

bmbouter: I think everything importing settings should be using `from django.conf import settings` which allows dynaconf to hook in as expected and layer, validate, etc.

Also could the issue be that django.setup() hasn't been called yet? We have `django.setup()` called by all the various entrypoints of Pulp, e.g. WSGI, the tasking system, etc, but maybe we don't have one for the test suite currently....

bmbouter: To recap our RTFM working group call from earlier, we're going to leave `get_pulp_setting` provided by pulp_smash for now. We can move that later.

pulpcore/tests/functional/api/test_upload.py
```diff
@@ -1,154 +1,178 @@
...
+        for data in chunks:
+            kwargs = {"file": data[0], "content_range": data[1], "upload_href": upload.pulp_href}
+            if include_chunk_sha256:
+                if len(data) != 3:
+                    raise Exception(f"Chunk didn't include its sha256: {data}")
+                kwargs["sha256"] = data[2]
+
+            uploads_api_client.update(**kwargs)
+
+        finish_task = uploads_api_client.commit(upload.pulp_href, {"sha256": sha256}).task
```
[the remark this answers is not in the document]
bmbouter: Is it that it's not `post` it's `commit` and `gen_object_with_cleanup` doesn't know that?

bmbouter (approved):
I asked one question mainly for my understanding, but this is looking great. Thank you!

pulpcore/tests/functional/api/test_crd_artifacts.py
dkliban: the following diff worked for me

```diff 
diff --git a/pulpcore/tests/functional/api/test_crd_artifacts.py b/pulpcore/tests/functional/api/test_crd_artifacts.py
index 79a24536e..5712ab63e 100644
--- a/pulpcore/tests/functional/api/test_crd_artifacts.py
+++ b/pulpcore/tests/functional/api/test_crd_artifacts.py
@@ -5,8 +5,8 @@ import os
 import uuid
 import pytest
 
-from pulp_smash.utils import get_pulp_setting
 from pulp_smash.exceptions import CalledProcessError
+from pulpcore.app import settings
 from pulpcore.client.pulpcore import ApiException
 
 
@@ -141,10 +141,11 @@ def test_upload_mixed_attrs(artifacts_api_client, pulpcore_random_file):
 
 def test_all(artifacts_api_client, pulpcore_random_file, cli_client):
     """Delete an artifact, it is removed from the filesystem."""
-    storage = get_pulp_setting(cli_client, "DEFAULT_FILE_STORAGE")
+    from pulpcore.app import settings
+    storage = settings.DEFAULT_FILE_STORAGE
     if storage != "pulpcore.app.models.storage.FileSystem":
         pytest.skip("this test only works for filesystem storage")
-    media_root = get_pulp_setting(cli_client, "MEDIA_ROOT")
+    media_root = settings.MEDIA_ROOT
 
     artifact = artifacts_api_client.create(pulpcore_random_file["name"])
     cmd = ("ls", os.path.join(media_root, artifact.file))
```

bmbouter: If we do go down this road I think we should `from django.conf import settings`. @gerrod3 if you want to do this in this PR that's fine, or if not also fine (to me).

gerrod3: @dkliban Did you test on a system using a non-file system backend? I’m pretty sure the default value is the file system so that value is already set in the module before `django.setup` is ran.

pulpcore/tests/functional/api/test_upload.py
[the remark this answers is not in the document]
gerrod3: `commit` takes an `upload` object and turns it into an `artifact`, deleting the `upload` object in the process. Since this api creates an object of another api, `gen_object `has no idea how delete the created object.

pulpcore/tests/functional/api/test_crd_artifacts.py
dkliban: @gerrod3 yes, i added `DEFAULT_FILE_STORAGE='storages.backends.s3boto3.S3Boto3Storage'` to `/etc/pulp/settings.py` and the test started being skipped. 

@bmbouter `from django.conf import settings` works just the same. I don't feel strongly about one way or another.

pulpcore/tests/functional/__init__.py
```diff
@@ -347,6 +370,17 @@ def pulp_api_v3_url(pulp_cfg, pulp_api_v3_path):
     return f"{pulp_cfg.get_base_url()}{pulp_api_v3_path}"
 
 
+@pytest.fixture
+def get_redis_status(status_api_client):
+    """Return a boolean value which tells whether the connection to redis was established or not."""
+    status_response = status_api_client.status_read()
+    try:
+        is_redis_connected = status_response["redis_connection"]["connected"]
```
dkliban, line 378: This is not working as you suspect it is. The status_read() method returns a StatusResponse[0]. This object is not subscriptable so you are always getting an exception trying to access `"redis_connection"`.

[0] https://docs.pulpproject.org/pulpcore_client/StatusResponse/

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -2,179 +2,155 @@
...
 
-        api_client = api.Client(cfg, api.json_handler)
+    Do the following:
 
-        # create
-        files = {"file": utils.http_get(FILE_URL)}
-        artifact = api_client.post(ARTIFACTS_PATH, files=files)
-        self.addCleanup(api_client.delete, artifact["pulp_href"])
-        cmd = ("ls", os.path.join(media_root, artifact["file"]))
+    1. Upload a file and provide both an ``sha256`` and a ``size``. Let one
```
dkliban, line 129: one is valid and the other is invalid.

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -2,179 +2,155 @@
...
+def test_all(artifacts_api_client, pulpcore_random_file, cli_client):
+    """Delete an artifact, it is removed from the filesystem."""
+    storage = get_pulp_setting(cli_client, "DEFAULT_FILE_STORAGE")
+    if storage != "pulpcore.app.models.storage.FileSystem":
+        pytest.skip("this test only works for filesystem storage")
+    media_root = get_pulp_setting(cli_client, "MEDIA_ROOT")
+
+    artifact = artifacts_api_client.create(pulpcore_random_file["name"])
+    cmd = ("ls", os.path.join(media_root, artifact.file))
+    cli_client.run(cmd, sudo=True)
```
dkliban, line 151: you can just use Python to check the existence of the file.
```python
from os.path import exists

path_to_file = os.path.join(media_root, artifact.file)
file_exists = exists(path_to_file)
```

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -2,179 +2,155 @@
...
+    """
+    invalid_data = (
+        {"sha256": pulpcore_random_file["digest"], "size": pulpcore_random_file["size"] + 1},
+        {"sha256": str(uuid.uuid4()), "size": pulpcore_random_file["size"]},
+    )
+    for data in invalid_data:
+        _do_upload_invalid_attrs(artifacts_api_client, pulpcore_random_file, data)
+
+
+def test_all(artifacts_api_client, pulpcore_random_file, cli_client):
```
dkliban, line 142: The test name seems inappropriate. Maybe just `test_delete_artifact` ?

dkliban (approved):
This looks great! Thank you!

pulpcore/tests/functional/api/test_api_docs.py
```diff
@@ -1,43 +1,38 @@
...
 
+@pytest.fixture(scope="session")
+def pulp_docs_url(pulp_api_v3_url):
+    return f"{pulp_api_v3_url}docs/"
 
-class ApiDocsTestCase(unittest.TestCase):
-    """Test whether API auto generated docs are available.
 
-    This test targets the following issue:
+def test_valid_credentials(pulpcore_client, pulp_docs_url):
```
dkliban, line 11: This test can be marked as parallel.

pulpcore/tests/functional/api/test_api_docs.py
```diff
@@ -1,43 +1,38 @@
...
+    assert response.status == 200
 
-    def setUp(self):
-        """Create an API Client."""
-        cfg = config.get_config()
-        self.client = api.Client(cfg)
 
-    def test_valid_credentials(self):
-        """Get API documentation with valid credentials.
+def test_no_credentials(pulpcore_client, pulp_docs_url, anonymous_user):
```
dkliban, line 20: This one can be parallel also.

pulpcore/tests/functional/api/test_api_docs.py
```diff
@@ -1,43 +1,38 @@
...
+        assert response.status == 200
 
-    def test_no_credentials(self):
-        """Get API documentation with no credentials.
 
-        Assert the API documentation is returned.
-        """
-        del self.client.request_kwargs["auth"]
-        self.client.get(API_DOCS_PATH)
+def test_http_method(pulpcore_client, pulp_docs_url):
```
dkliban, line 30: This one can be parallel also.

dkliban (changes requested):
All the tests in this PR seem like they can be run in parallel. Please mark them as such with `@pytest.mark.parallel`

pulpcore/tests/functional/api/test_crd_artifacts.py
bmbouter: It works the same, but it's not DRY. We specify the settings location when we configure django and to specify it here also puts that configuration in yet one more place.

dkliban (approved):

Issue #3299, opened 2022-10-12
RepositoryVersionContentDetails should not perfrom a content query when creating the content summary


gerrod3:
**Version**
All pulpcore-versions

**Describe the bug**
A query on content is perform to find the content model but this can be dangerous if the content is currently being deleted:

```python
repository_content = RepositoryContent.objects.filter(repository=repository)
ctype_query = Content.objects.filter(
    pulp_type=self.content_type, pk__in=repository_content.values_list("content", flat=True)
)
# This query can return `None` if the content is being deleted at the same time
ctype_model = ctype_query.first().cast().__class__
```
**To Reproduce**
https://github.com/pulp/pulp_ansible/issues/1274
This problem occurs pretty rarely and I had already created a fix to stop it from occurring for `orphan_cleanup`, but I think an approach without performing a query is needed to prevent this issue from appearing again.

dralley:
Wouldn't this be dangerous anyway, considering that any content being queried here would have been part of the repository version?  We can't delete content that's in a repository version

dralley:
Oh so this is a separate deletion endpoint provided by the ansible plugin, as opposed to orphan cleanup or something.

That still seems concerning.  Race condition?

gerrod3:
> Oh so this is a separate deletion endpoint provided by the ansible plugin, as opposed to orphan cleanup or something.
> 
> That still seems concerning. Race condition?

I think so, the problem is super sneaky. What I think is happening is that `delete_collection` creates multiple `repository_versions` [0] Each one of these repo versions gets added to the tasks `created_resources`. However each time a new `repository_version` is created the task goes and deletes all previous versions to only keep the latest. So when requesting the task href the serializer for the `created_resources` looks up these non-existent `repository_versions` and if during the lookup the `repository_version` is deleted then I think we run into this error. When I tested this endpoint I see in the task report a bunch of `null`s for all the created versions that were deleted. 

I'm going to comment on that pulp_ansible issue to change the deletion code to not create unnecessary repository versions.

[0] https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/tasks/deletion.py#L71-L73

pulpbot:
https://bugzilla.redhat.com/show_bug.cgi?id=2160008

pulpbot:
https://bugzilla.redhat.com/show_bug.cgi?id=2174987

PR #3300, opened 2022-10-12
Remove query from repository version content summary

gerrod3:
fixes: #3299

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -1128,11 +1128,8 @@ class RepositoryVersionContentDetails(models.Model):
             dict: {<pulp_type>: <url>}
         """
         repository = self.repository_version.repository.cast()
-        repository_content = RepositoryContent.objects.filter(repository=repository)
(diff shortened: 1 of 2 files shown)


dralley (approved):
Nice, this should work!

pulpcore/app/models/repository.py
```diff
@@ -1128,11 +1128,8 @@ def content_href(self):
             dict: {<pulp_type>: <url>}
         """
         repository = self.repository_version.repository.cast()
```
gerrod3, line 1130: I think the race condition window is still present, just significantly smaller now that we are not performing a query on content. The `content_href` method could be called (in a serializer) while the repo_version is being deleted. This could cause this line to error still if `self.repository_version` has become null.

dralley:
Here's an interesting comment, potentially related https://community.theforeman.org/t/delete-orphaned-content-task-failing-server-error-500/30401/10?u=dralley

mdellweg (approved):

PR #3311, opened 2022-10-13
Removes test that required restarting the db.

dkliban:
This test has been unreliable in the past and has been removed from other branches.

[noissue]


mdellweg (approved):

dralley:
@dkliban This is failing because of another unreliable test, I've seen this same spurious CI failure across a lot of older branches

dkliban:
@dralley I discovered that the failure was due to the the unit tests being run as `root` and the functional tests being run as `pulp`. This has already been addressed in the plugin template, so I reapplied it as part of this PR.

PR #3314, opened 2022-10-14
Add test_tasking module

mdellweg:
This moved over from pulp_file.

[noissue]


mdellweg:
Hah! One of the tests is flakey. It seems like a canceled task can have `finished_at=None`, but also set to something _meaningful_!?

mdellweg:
Is "finished_at" supposed to be set on a canceled task?

dkliban:
@mdellweg A task that is cancelled before it is finished should not have a `finished_at` set, but it should have a `cancelled_at` set. A user can't cancel a finished task. So the two fields are mutually exclusive. At least that is my understanding. However, it seems like this is not true in practice and perhaps we should file an issue about that.

mdellweg:
> @mdellweg A task that is cancelled before it is finished should not have a `finished_at` set, but it should have a `cancelled_at` set. A user can't cancel a finished task. So the two fields are mutually exclusive. At least that is my understanding. However, it seems like this is not true in practice and perhaps we should file an issue about that.

That is exactly what the tests assert now, except that we do not have a `canceled_at` field. Can you write an issue if that is desired?

dkliban:
@mdellweg I filed https://github.com/pulp/pulpcore/issues/3319

bmbouter (approved):
This is really nice!

PR #3325, opened 2022-10-18
[noissue] Moves some tasks tests from pulp_file into pulpcore.

decko:
Since we found that some tests from `pulp_file` not only needed a fixture from `pulpcore` but also make sense to be here, we decided to move those tests and a `download_file` utility into `pulpcore`.

You can follow the discussion that ended up into this PR [here](https://github.com/pulp/pulp_file/pull/845#pullrequestreview-1144801290)

And `pulp_smash` no more.

### DIFF
--- a/pulpcore/tests/functional/api/test_tasking.py
+++ b/pulpcore/tests/functional/api/test_tasking.py
@@ -1,11 +1,16 @@
 """Tests related to the tasking system."""
+import json
 import pytest
 import subprocess
 import time
+from aiohttp import BasicAuth
+from urllib.parse import urljoin
 from uuid import uuid4
 
-from pulp_smash.pulp3.bindings import monitor_task
+from pulpcore.constants import TASK_FINAL_STATES
 from pulpcore.client.pulpcore import ApiException
+
+from pulpcore.tests.functional.utils import download_file, monitor_task
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -31,8 +36,23 @@ def _dispatch_task(*args, **kwargs):
...
+    task_href = dispatch_task("time.sleep", args=(0,))
+    task = tasks_api_client.read(task_href)
+
+    yield task
+
+    if task.state in TASK_FINAL_STATES:
+        tasks_api_client.delete(task_href)
+    else:
+        tasks_api_client.tasks_cancel(task_href, {"state": "canceled"})
+        tasks_api_client.delete(task_href)
```
mdellweg, line 51: But also it's fine for tests to leave tasks around...
```suggestion
    if task.state not in TASK_FINAL_STATES:
        tasks_api_client.tasks_cancel(task_href, {"state": "canceled"})
        # here may be another monitor_task needed, as cancelling is only triggered by that call.
    tasks_api_client.delete(task_href)
```

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -31,8 +36,23 @@ def _dispatch_task(*args, **kwargs):
...
+
+
 @pytest.mark.parallel
-def test_multi_resource_locking(tasks_api_client, dispatch_task):
```
mdellweg, line 35: Hah! Good catch.

pulpcore/tests/functional/api/test_tasking.py
mdellweg: Also yet another thought: we could put that cleanup at the end of the dispatch_task fixture.

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -90,7 +110,7 @@ def test_delete_cancel_waiting_task(dispatch_task, tasks_api_client):
 
 @pytest.mark.parallel
 def test_delete_cancel_running_task(dispatch_task, tasks_api_client):
-    task_href = dispatch_task("time.sleep", args=(600,))
+    task_href = dispatch_task("time.sleep", args=(6000,))
```
mdellweg, line 113: 10' was not enough? Well, whatever!

pulpcore/tests/functional/utils.py
```diff
@@ -92,3 +94,32 @@ def monitor_task_group(tg_href):
...
+@dataclass
+class Download:
+    """Class for representing a downloaded file."""
+
+    body: bytes
+    response_obj: aiohttp.ClientResponse
+
+    def __init__(self, body, response_obj):
+        self.body = body
+        self.response_obj = response_obj
```
mdellweg, line 108: Interesting. I think we used NamedTuple for stuff like this before.

mdellweg (approved):

pulpcore/tests/functional/utils.py
decko: `dataclasses` are good alternative since it already came with types and other good features.

dkliban (approved):

PR #3337, opened 2022-10-20
Harden task state transitions

### LINKED ISSUE #3319: Tasks API should report a `finished_at` date/time for cancelled tasks
Cancelled tasks should have `finished_at` set at cancel time.
Ideally any attempt to set `finished_at` from something different than `None` in the database should be reported as a `RuntimeError`. (audit of all `Task.save()` and `Task.objects.update()` calls)

Original wording: Tasks API should return a `cancelled_at` date/time for tasks that were cancelled. The `finished_at` should be set to `null` when `cancelled_at` is not `null`.

mdellweg:
This will make sure tasks can only traverse from certain states to others and that finished_at will be set where appropriate.

fixes #3319


pulpcore/app/models/task.py
```diff
@@ -248,17 +250,15 @@ def set_completed(self):
...
-        # important for when the task has been canceled, so we don't move the task from canceled
-        # to finished.
-        rows = (
-            Task.objects.filter(pk=self.pk)
-            .exclude(state__in=TASK_FINAL_STATES)
-            .update(state=TASK_STATES.COMPLETED, finished_at=timezone.now())
+        # Only set the state to finished if it's running. This is important for when the task has
+        # been canceled, so we don't move the task from canceled to finished.
+        rows = Task.objects.filter(pk=self.pk, state=TASK_STATES.RUNNING).update(
+            state=TASK_STATES.COMPLETED, finished_at=timezone.now()
```
decko, line 256: What do you think about writing it like this?

```suggestion
        try:
            Task.objects.get(pk=self.pk).update(state=TASK_STATES.COMPLETED, finished_at=timezone.now())
        except DoesNotExist:
            raise RuntimeError(
                _("Task set_completed() occurred but Task {} is not RUNNING.").format(self.pk)
            )
```

pulpcore/app/models/task.py
```diff
@@ -273,17 +273,44 @@ def set_failed(self, exc, tb):
...
+        rows = Task.objects.filter(pk=self.pk, state=TASK_STATES.RUNNING).update(
+            state=TASK_STATES.FAILED,
+            finished_at=timezone.now(),
+            error=exception_to_dict(exc, tb_str),
+        )
+        if rows != 1:
+            raise RuntimeError(_("Attempt to set a not running task to failed."))
+        self.refresh_from_db()
+
+    def set_canceling(self):
```
decko, line 285: :rocket: :rocket: :rocket:

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -184,32 +184,24 @@ def cancel_abandoned_task(self, task, final_state, reason=None):
...
-                "state": final_state,
-                "finished_at": timezone.now(),
-            }
-            if reason:
-                task_data["error"] = {"reason": reason}
-            Task.objects.filter(pk=task.pk, state=TASK_STATES.CANCELING).update(**task_data)
-            return True
-        return False
+        try:
+            task.set_canceling()
```
decko, line 188: :rocket:

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -106,13 +106,13 @@ def test_delete_cancel_waiting_task(dispatch_task, tasks_api_client):
...
+    assert task.finished_at is not None
 
 
-@pytest.mark.skip(
-    reason="This is a flaky test that fails from time to time. Probably gonna be fixed by 3319"
-)
+@pytest.mark.parallel
 def test_delete_cancel_running_task(dispatch_task, tasks_api_client):
-    task_href = dispatch_task("time.sleep", args=(6000,))
+    task_href = dispatch_task("time.sleep", args=(600,))
```
decko, line 115: :point_up:

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -92,9 +92,9 @@ def test_delete_cancel_waiting_task(dispatch_task, tasks_api_client):
     # Now cancel the task
     task = tasks_api_client.tasks_cancel(task_href, {"state": "canceled"})
 
+    assert task.state == "canceling"
```
decko, line 95: Should we need to test changing the states from `canceling` to `canceled` around here?

decko:
I have a impression that we need a [simple] state machine. Soon.

pulpcore/app/models/task.py
mdellweg: This is using the `filter().update()` pattern on purpose to avoid race conditions.

pulpcore/tests/functional/api/test_tasking.py
mdellweg: I think this very test is doing that.

pulpcore/app/models/task.py
```diff
@@ -273,17 +273,44 @@ def set_failed(self, exc, tb):
...
+        Set this task to canceling from either waiting, running or canceling.
+
+        This is the only valid transition without holding the task lock.
+        """
+        rows = Task.objects.filter(pk=self.pk, state__in=TASK_INCOMPLETE_STATES).update(
+            state=TASK_STATES.CANCELING,
+        )
+        if rows != 1:
+            raise RuntimeError(_("Attempt to cancel a finished task."))
+        self.refresh_from_db()
```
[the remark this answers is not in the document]
dralley: If you were to try to use `task` after calling `task.set_canceling()` without refreshing the database, then the cached values stored on `task` wouldn't be updated, and any future logic you might perform with it could be invalid.


If you don't refresh the `task` from the database then you could try to

dralley (approved):
Looks good, does this actually fix #3319 or will it just catch scenarios where the logic goes is wrong and make them visible?

pulpcore/app/models/task.py
mdellweg: Maybe to elaborate more, get `get()` and then `update()` will emit a `select` and then an `update` query to the database.
When using `filter().update()` it will emit a single sql statement that selects the items to be updated in the same atomic step. That way we can ensure (on the database layer) that the task can only move from "RUNNING" to "COMPLETED".

pulpcore/app/models/task.py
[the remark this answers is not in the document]
mdellweg: Yeah, it's a hunch that we might need the updated record later. Maybe we can resort to make `state` and `finished_at` deferred. :eyes:

mdellweg:
> Looks good, does this actually fix #3319 or will it just catch scenarios where the logic goes is wrong and make them visible?

I have a suspicion that turning the warnings into errors may fix some races and thereby fix the issue. But i think, i improved the test assertions a bit too, so we should see soon.

pulpcore/app/models/task.py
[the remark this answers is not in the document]
mdellweg: https://docs.djangoproject.com/en/3.2/ref/models/instances/#refreshing-objects-from-database

pulpcore/app/models/task.py
```diff
@@ -273,18 +280,70 @@ def set_failed(self, exc, tb):
...
+        """
+        rows = Task.objects.filter(pk=self.pk, state__in=TASK_INCOMPLETE_STATES).update(
+            state=TASK_STATES.CANCELING,
+        )
+        if rows != 1:
+            raise RuntimeError(_("Attempt to cancel a finished task."))
+        with suppress(AttributeError):
+            del self.state
+
+    def set_canceled(self, final_state=TASK_STATES.CANCELING, reason=None):
```
dkliban, line 311:
```suggestion
    def set_canceled(self, final_state=TASK_STATES.CANCELED, reason=None):
```

dkliban (approved):
Thanks for explaining all these changes. It makes sense to me. Great job!

dralley (changes requested):
I prefer the original approach.  Not just because it was simpler, but because if additional code down the line *does* use these fields, it's going to trigger new database queries for each one.  There's a high chance it either breaks even or results in even more queries than before.

dralley (approved):
On further review, technically speaking, this is fine.  So my only complaint is that it is not as aesthetically pleasing

PR #3353, opened 2022-10-24
Fixup the fix for #3319

mdellweg:
Due to not being deferred properly, some times finished_at was not being reloaded.

[noissue]
re #3319


ipanova (approved):
thank you for taking time finding the root cause :+1:

dkliban (approved):

dralley:
Is the minor efficiency boost of this approach worth the risk of problems such as this being introduced?

mdellweg:
I think this is the cleanest way now.

PR #3378, opened 2022-11-02
Refactor api_clients to session scope in tests

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -6,9 +6,12 @@ import uuid
 
 import pytest
 
+from time import sleep
+
 from pulp_smash.config import get_config
-from pulp_smash.pulp3.bindings import delete_orphans
 from pulp_smash.utils import get_pulp_setting
+
+from pulpcore.tests.functional.utils import SLEEP_TIME, PulpTaskError, PulpTaskGroupError
 
 from pulpcore.client.pulpcore import (
     ApiClient,
@@ -38,6 +41,7 @@ from pulpcore.client.pulpcore import (
     RolesApi,
     SigningServicesApi,
     StatusApi,
+    TaskGroupsApi,
     TasksApi,
     TaskSchedulesApi,
     UploadsApi,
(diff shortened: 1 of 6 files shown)


pulpcore/tests/functional/__init__.py
```diff
@@ -88,177 +88,185 @@ def pytest_configure(config):
...
+@pytest.fixture(scope="session")
+def _api_client_list():
+    return set()
+
+
+@pytest.fixture(autouse=True)
+def cid(_api_client_list, monkeypatch, pulpcore_client):
     value = str(uuid.uuid4())
+    for api_client in _api_client_list:
+        monkeypatch.setitem(api_client.default_headers, "Correlation-ID", value)
```
decko, line 100: Neat!

pulpcore/tests/functional/__init__.py
```diff
@@ -88,181 +92,207 @@ def pytest_configure(config):
...
+        )
+
+
+@pytest.fixture(scope="session")
+def pulpcore_client(_api_client_set, bindings_cfg):
     api_client = ApiClient(bindings_cfg)
-    api_client.default_headers["Correlation-ID"] = cid
-    return api_client
+    _api_client_set.add(api_client)
+    yield api_client
```
gerrod3, line 123: I'm very confused on how this works. I was under the impression that session fixtures get called once and will return the same value everytime they are used again. Is this really generating a new client each time? If so then what is the "session" scope actually doing?

mdellweg: No, the trick is that it really creates only one api_client per test runner. And that one will be shared by all the tests that run on that runner (btw in sequence).
The api client now advertises itself in the also shared `_api_client_set` where the function scoped fixture `_patch_cid_user_agent` will monkeypatch the user-agent and cid for the lifetime of the specific test.

This allows the api_clients to be used on a broader scope (without user_agent and cid) and that way to create reusable fixtures in Pulp. My plan is to move the cleanup fixtures to pulpcore and scoping them on the module level, so fixtures like "file_repository_with_four_versions" (costly to set up) can be shared by tests in one module.
"--dist loadfile" on the `pytest` cmdline will then schedule whole file to runners and complete the picture for shared fixtures.

ggainey (approved):
Tentative approval - the discussion around what session means helped shed some light for me, and this all looks reasonable. But I'm not an expert here yet, more-experienced pytest-eyes would be good.

pulpcore/tests/functional/utils.py
```diff
@@ -17,13 +13,6 @@ async def get_response(url):
         return await session.get(url)
 
 
-skip_if = partial(selectors.skip_if, exc=SkipTest)  # pylint:disable=invalid-name
```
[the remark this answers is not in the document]
mdellweg: It's def a unittest relict.

dkliban (approved):

PR #3386, opened 2022-11-04
Cleanup fixtures

mdellweg:


pulpcore/tests/functional/__init__.py
```diff
@@ -95,6 +95,20 @@ def pytest_configure(config):
 # API Clients
 
 
+@pytest.fixture(scope="session")
+def bindings_cfg():
+    api_protocol = os.environ.get("API_PROTOCOL", "https")
+    api_host = os.environ.get("API_HOST", "localhost")
+    api_port = os.environ.get("API_PORT", "80")
```
lubosmj, line 102: Maybe, changing the port value would be better?
```suggestion
    api_port = os.environ.get("API_PORT", "443")
```

```
cat .config/pulp_smash/seetings.json

{
  "hosts": [
    {
      "hostname": "pulp3-source-fedora35.localhost.example.com",
      "roles": {
        "api": {
          "port": 443,
          "scheme": "https",
          "service": "nginx",
          "verify": false
        },
        "content": {
          "port": 443,
          "scheme": "https",
          "service": "pulpcore-content",
          "verify": false
        },
        "pulp resource manager": {},
        "pulp workers": {},
        "redis": {},
        "shell": {
          "transport": "local"
        }
      }
    }
  ],
  "pulp": {
    "auth": [
      "admin",
      "password"
    ],
    "selinux enabled": false,
    "version": "3"
  }
}
```

pulpcore/tests/functional/__init__.py
```diff
@@ -470,7 +495,62 @@ def get_redis_status(status_api_client):
...
+        except Exception:
+            # There was no delete task for this unit or the unit may already have been deleted.
+            # Also we can never be sure which one is the right ApiException to catch.
+            pass
+
+    for deleted_task_href in delete_task_hrefs:
+        monitor_task(deleted_task_href)
+
+
+@pytest.fixture(scope="module")
```
lubosmj, line 526: I am unsure if setting the scope to the module level is safe. I used to see tests using the same repository names. Have you tried running different plugin tests apart from pulp_file?

mdellweg: That is also a good question. I see potential here for reusable fixtures (at least at the module level). One of our goals is to have individual fixtures not collide. And this would basically introduce a new test policy: Don't change shared fixtures.

pulpcore/tests/functional/__init__.py
lubosmj: Yes, the variables are set by oci_env. And, I think we should set the variables in the CI too. Any objections, @dkliban?

Note that a vagrant installation has no notion about these variables. We need to set the variables in the installer if we plan to support it.

mdellweg:
I think this does not mean we can stop installing pulp-smash, because fixtures are not imported explicitly.

dkliban (approved):
This is great! Thank you!

ggainey (approved):
Looks good, matthias, nice cleanup!

PR #3388, opened 2022-11-08
Rework filter defaults

### LINKED ISSUE #3351: HyperlinkedRelatedFilter reports incorrect format in the openapi spec
`pulp --refresh-api debug openapi operation --id tasks_list | jq '.operation.parameters[]|select(.name=="worker")'` shows the format as "uuid" while it should be "uri".

mdellweg:
This fixes the apischema generation for the HyperlinkRelatedFilter.

fixes #3351

### DIFF
--- a/pulpcore/app/viewsets/access_policy.py
+++ b/pulpcore/app/viewsets/access_policy.py
@@ -1,28 +1,14 @@
-from django_filters.rest_framework import filters
 from rest_framework import mixins
 from rest_framework.decorators import action
 from rest_framework.response import Response
 from drf_spectacular.utils import extend_schema
 
-from pulpcore.filters import BaseFilterSet
 from pulpcore.app.apps import pulp_plugin_configs
 from pulpcore.app.models import AccessPolicy
 from pulpcore.app.serializers import AccessPolicySerializer
 from pulpcore.app.viewsets import NamedModelViewSet
 from pulpcore.app.viewsets.base import NAME_FILTER_OPTIONS
 from pulpcore.app.util import get_view_urlpattern
-
-
-class AccessPolicyFilter(BaseFilterSet):
-    """
-    FilterSet for AccessPolicy.
-    """
-
-    customized = filters.BooleanFilter()
-
-    class Meta:
-        model = AccessPolicy
-        fields = {"viewset_name": NAME_FILTER_OPTIONS, "customized": ["exact"]}
 
 
 class AccessPolicyViewSet(
@@ -39,7 +25,7 @@ class AccessPolicyViewSet(
     serializer_class = AccessPolicySerializer
     endpoint_name = "access_policies"
     ordering = "viewset_name"
-    filterset_class = AccessPolicyFilter
+    filterset_fields = {"viewset_name": NAME_FILTER_OPTIONS, "customized": ["exact"]}
 
     @extend_schema(request=None)
     @action(detail=True, methods=["post"])
--- a/pulpcore/app/viewsets/publication.py
+++ b/pulpcore/app/viewsets/publication.py
@@ -1,7 +1,6 @@
 from gettext import gettext as _
 
 from django_filters import Filter
-from django_filters.rest_framework import filters
 from rest_framework import mixins, serializers
 
 from pulpcore.filters import BaseFilterSet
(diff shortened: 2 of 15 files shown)


pulpcore/filters.py
```diff
@@ -76,7 +179,11 @@ def get_filters(cls):
                 pass
         elif cls._meta.model:
             ordering_fields.extend(
-                ((field.name, field.name) for field in cls._meta.model._meta.get_fields())
+                (
+                    (field.name, field.name)
+                    for field in cls._meta.model._meta.get_fields()
+                    if not field.is_relation
```
decko, line 185: Is this the line that allowed the `HyperlinkRelatedFilter` to work with the `in` modifier?

decko (approved):
LTTM :rocket:

pulpcore/filters.py
mdellweg: No, the FILTER_DEFAULTS is doing the magic. This change just removes the related fields from ordering, because order by uuid is not really useful.

pulpcore/filters.py
```diff
@@ -140,3 +247,10 @@ class PulpFilterBackend(DjangoFilterBackend):
 
 class PulpOpenApiFilterExtension(DjangoFilterExtension):
     target_class = "pulpcore.filters.PulpFilterBackend"
+
+    def _get_schema_from_model_field(self, auto_schema, filter_field, model):
+        # Workaround until we can hook into `unambiguous_mapping` of
```
dkliban, line 252: Do we need to file an issue for this?

dkliban (approved):
This is great! I checked out the new docs generated with these changes and it all look correct.

pulpcore/filters.py
dralley: +1

pulpcore/app/viewsets/access_policy.py
```diff
@@ -13,18 +11,6 @@
 from pulpcore.app.util import get_view_urlpattern
 
 
-class AccessPolicyFilter(BaseFilterSet):
```
dralley, line 16: I take it that in terms of deleting these, we are constrained by which class-based filtersets are already exported via public API?

And that we have no intention of extending these on a per-plugin basis?

mdellweg: I did not delete it from the plugin api, so it could not be subclassed there anyway. We can always bring it back if there is a need.

pulpcore/app/viewsets/publication.py
```diff
@@ -76,14 +74,12 @@ def filter(self, qs, value):
...
     repository = RepositoryThroughVersionFilter(help_text=_("Repository referenced by HREF"))
     repository_version = RepositoryVersionFilter()
-    pulp_created = IsoDateTimeFilter()
     content = PublicationContentFilter()
     content__in = PublicationContentFilter(field_name="content", lookup_expr="in")
 
     class Meta:
         model = Publication
         fields = {
-            "repository_version": ["exact"],
```
dralley, line 86: I take it that this line isn't strictly necessary as per the django-filters docs, but is there any harm in leaving it explicitly?

mdellweg: It is trying to do a thing and is then subsequently overwritten by the direct filter implementation in line 76. So it's at least calling unnecessary code paths.

pulpcore/app/viewsets/access_policy.py
dralley: Right, but what I'm asking is, is there a particular benefit in using `filterset_fields` instead of `filterset_class` (other than the amount of code, which is valid) when we are restricted to using the class-based approach for anything currently exposed as such anyway?

pulpcore/app/viewsets/publication.py
dralley: I'm surprised that it would be overwritten by the field declaration, I would have expected the reverse.

mdellweg: I was surprised too. So let's add, that it's adding to the confusion to leave it here.

pulpcore/app/viewsets/access_policy.py
mdellweg: It is mainly easier to code and read. The filter backend will create a `BaseFilterSet` subclass for us. But i think with the proper defaults (introduced in this PR) we can get away with `filterset_fields` in much more places. We only need the `filterset_class` approach once we need special filters.

dralley (approved):

PR #3395, opened 2022-11-11
Add a repository filter to distributions endpoint

### LINKED ISSUE #3394: Add a repository filter to the generic distributions endpoint
We set the repository field on distributions and would like to see which distributions are distributing a repo.

daviddavis:
fixes #3394

### DIFF
--- a/CHANGES/3394.feature
+++ b/CHANGES/3394.feature
@@ -1,0 +1,1 @@
+Added a ``repository`` filter to the distributions endpoint.
--- a/pulpcore/app/viewsets/publication.py
+++ b/pulpcore/app/viewsets/publication.py
@@ -33,6 +33,7 @@ from pulpcore.app.viewsets import (
 from pulpcore.app.viewsets.base import DATETIME_FILTER_OPTIONS, NAME_FILTER_OPTIONS
 from pulpcore.app.viewsets.custom_filters import (
     DistributionWithContentFilter,
+    HyperlinkRelatedFilter,
     IsoDateTimeFilter,
     LabelSelectFilter,
     RepositoryVersionFilter,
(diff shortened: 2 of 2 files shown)


ipanova (approved):

ipanova:
@daviddavis would you be able to add a test please?

daviddavis:
> Do we want this filter to only show distributions with the repository attached directly? Or would one also be interested in distributions linked to a version of that repository?

I looked into adding a test but I didn't see any existing test infrastructure I could leverage so I gave up.

> Do we want this filter to only show distributions with the repository attached directly? Or would one also be interested in distributions linked to a version of that repository?

For us, we only care about the former. I could see the latter being useful but I could also see a case where a user might want to differentiate between the two in which case two filters might make sense: a repo filter and a repo version filter.

mdellweg (changes requested):
> For us, we only care about the former. I could see the latter being useful but I could also see a case where a user might want to know explicitly what distros have a particular repo (and not any repo versions for that repo) so perhaps two distinct filters makes sense.

I can see your point, and the filter as is is adding value here. The thing we need to get right now is the naming. So without implementing the "any version from this repository" filter today, can we find two names for these filters that we will be happy with in the future? (Maybe add a comment with the other name for future implementors.)

Also my ask on this pull request: Can you please add a helptext to the filter stating this very distinction?

daviddavis:
I thought about this some more and agree that perhaps the repository filter should also account for repo versions.

Also, I realized that this viewset is being extended by plugin distribution viewsets (not sure if that's a problem--I don't think it is?). However, it does look like there are some tests in pulp_file that test distribution filtering so I think I can add a check there for this repository filter.

daviddavis:
Closing in favor of https://github.com/pulp/pulpcore/pull/3503

PR #3397, opened 2022-11-11
Adds postgresql version analytics collection

### LINKED ISSUE #3396: Gather postgresql version metric collection
As a pulp user have the postgresql server gathered consistent with [this proposal](https://hackmd.io/zJ1dJe8qQtmzr0JiM1jptw).

bmbouter:
Adds an analytics entry like: `{'postgresqlVersion': 90200}`

closes #3396


ggainey:
@bmbouter would love to see this in core/3.22 - how close are we to having this be not-draft?

bmbouter:
> @bmbouter would love to see this in core/3.22 - how close are we to having this be not-draft?

Close, I expect it to undraft today/tomorrow and merge by end of week. It's entirely waiting on https://github.com/pulp/analytics.pulpproject.org/pull/62, technically this PR is ready to be undrafted already. Protobuf allows us to merge and release this ahead of the server side even knowing about it. :)

mdellweg (approved):

dkliban (changes requested):
We should be collecting the raw value returned from Postgresql. In case we are converting the version string incorrectly, we will have an opportunity to fix the conversion if we perform it on the analytics server.

mdellweg (approved):

dkliban (approved):

PR #3418, opened 2022-11-19
Delete orphans immediately

lubosmj:
[noissue]


mdellweg (changes requested):
I'm in favor of this change, but please rewrite the commit message. (hide a "test" and a "fixture" somewhere in the first line of it. It sound quite scary at the moment.)

mdellweg (approved):

PR #3422, opened 2022-11-22
WIP Experiment testing with lower bounds dependencies

mdellweg:
[noissue]


.github/workflows/scripts/before_install.sh
```diff
@@ -150,6 +150,11 @@ fi
 
 cd pulpcore
 
+######## EXPERIMENT ########
+python3 .ci/scripts/calc_deps_lower_bounds.py > lower_bounds_requirements.txt
+mv lower_bounds_requirements.txt requirements.txt
```
mdellweg, line 155: This is probably unsafe for publishing. We should make sure this is isolated from the publishing workflow.

requirements.txt
```diff
@@ -2,10 +2,10 @@ asyncio-throttle>=1.0,<=1.0.2
 aiohttp>=3.8.1,<=3.8.3
 aiodns>=3.0,<=3.0.0
 aiofiles>=22.1,<=22.1.0
-aioredis>=2.0.1<=2.0.1
+aioredis>=2.0.1,<=2.0.1
```
ipanova, line 5: ouch!

ggainey (approved):

Issue #3417, opened 2022-11-18
Rename TELEMETRY setting to ANALYTICS


bmbouter:
As was decided recently, the anonymous data gathered from Pulp installations is being called "Analytics" to avoid confusion with the [open telemetry working group](https://discourse.pulpproject.org/t/monitoring-telemetry-working-group/700) and the telemetry it will produce.

We've updated the term everywhere but in the pulpcore codebase. Here are some of the areas that need updating:
* The pulpcore code itself `git grep telemetry`, `git grep Telemetry`, `git grep TELEMETRY`.
* docs in [this section](https://github.com/pulp/pulpcore/blob/main/docs/components.rst#telemetry-collection)
* the TaskSchedule db entry
* the setting name itself.

I believe this should be listed as a breaking change although semver doesn't technically govern the settings themselves.

daviddavis:
> I believe this should be listed as a breaking change although semver doesn't technically govern the settings themselves.

My team and I were quite surprised reading this comment. We had the impression that Pulp's semver policy did apply to its settings. It's a pain to have to check that settings could change with any release of Pulp.

What sort of policy does Pulp provide around settings? Obviously my hope is that Pulp would not remove/change settings between major releases but if that's not possible can you at least guarantee some number of Y releases?

bmbouter:
My hope for Pulp as a project is to be very unsurprising for users, so your perspective here is helpful. I don't think we ever said one way or the other regarding settings, so maybe it's time we did. I'm reading a strong desire to never change a setting name and I can see why; it makes sense. I'll give it some thought. I'll comment on the PR also.

Thanks for joining the convo.

mdellweg:
As per this setting (being of boolean nature), i think we agreed to introduce a new name, but keep the old name around as an alias.
We can now either make one of them higher priority than the other, or in this case, take the ORed value of both. It would be sufficient to switch of either one. Which one is less surprising?

PR #3423, opened 2022-11-22
Renames TELEMETRY setting to ANALYTICS.

dkliban:
fixes: #3417


pulpcore/app/util.py
```diff
@@ -247,11 +247,11 @@ def gpg_verify(public_keys, signature, detached_data=None):
...
 
 
-def configure_telemetry():
-    task_name = "pulpcore.app.tasks.telemetry.post_telemetry"
+def configure_analytics():
+    task_name = "pulpcore.app.tasks.analytics.post_analytics"
     dispatch_interval = timedelta(days=1)
-    name = "Post Anonymous Telemetry Periodically"
-    if settings.TELEMETRY:
+    name = "Post Anonymous Analytics Periodically"
```
mdellweg, line 253: This will leave the old schedule lying around. Can you write an elidable migration to delete it?

pulpcore/app/settings.py
```diff
@@ -289,7 +289,7 @@
 
 TASK_DIAGNOSTICS = False
 
-TELEMETRY = True
+ANALYTICS = True
```
mdellweg, line 292: To make this change not breaking today, can you deprecate but still use the old setting (if set)? I'd say (if set) the new setting ought to take precedence.

CHANGES/3417.bugfix
```diff
@@ -0,0 +1 @@
+Renamed `TELEMETRY` setting to `ANALYTICS`. This is a necessary, but backwards incompatible change.
```
mdellweg, line 1: This should come with a removal changelog entry.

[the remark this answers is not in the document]
ipanova: Imagine user had set the telemetry to False and not we just replace it with analytics which is enabled by defaults. it means once the user will upgrade to next pulpcore version the data will be collected again, even if he previously marked the setting to False.

pulpcore/app/settings.py
dkliban: How long will we support the old setting? I  would like to drop it in a future release.

CHANGES/3417.bugfix
dkliban: I'll add one.

pulpcore/app/util.py
dkliban: Yes, I'll add a migration.

pulpcore/app/settings.py
dkliban: I propose remove `TELEMETRY` in 3.23.

CHANGES/3417.bugfix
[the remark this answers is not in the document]
dkliban: We don't have a deprecation policy for settings/user features. I propose that we deprecate in 3.22 and remove in 3.23.

dkliban:
After some discussion with @mdellweg on Matrix I left the `TELEMETRY` setting in place and introduced a new `ANALYTICS` setting. I also added a migration to remove the old TaskSchedule.

mdellweg:
> After some discussion with @mdellweg on Matrix I left the `TELEMETRY` setting in place and introduced a new `ANALYTICS` setting. I also added a migration to remove the old TaskSchedule.

To sum up my contribution to that discussion: Leaving the setting around as an alias forever is a small price payed for not upsetting users.

pulpcore/app/migrations/0097_remove_telemetry_task_schedule.py
```diff
@@ -0,0 +1,25 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0096_alter_task_logging_cid'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            remove_telemetry_task_schedule, reverse_code=migrations.RunPython.noop
```
mdellweg, line 23:
```suggestion
            remove_telemetry_task_schedule, reverse_code=migrations.RunPython.noop, elidable=True
```

mdellweg (approved):
One optional suggestion...

pulpcore/app/settings.py
```diff
@@ -289,8 +289,11 @@
 
 TASK_DIAGNOSTICS = False
 
+# Deprecated
 TELEMETRY = True
 
+ANALYTICS = TELEMETRY
+
```
mdellweg, line 296: Actually i doubt that this works as desired. After executing this module, `ANALYTICS` will be `True` and not "the value of `TELEMETRY`".

mdellweg (changes requested):

pulpcore/app/settings.py
dkliban: I'll try to add a unit test around this.

CHANGES/3417.bugfix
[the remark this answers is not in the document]
ipanova: that's fine with me as long as we don't delete that setting right away, first deprecate it and introduce a new analytics settings, basically this https://github.com/pulp/pulpcore/pull/3423#discussion_r1029522070

daviddavis:
> To sum up my contribution to that discussion: Leaving the setting around as an alias forever is a small price payed for not upsetting users.

We have to set `TELEMETRY` to false in some of our environments for security reasons so maintaining it as an option until the next major release of pulp would make us happy. Also, my hope would be that settings would be covered by Pulp's semver policy.

Thanks 👍

bmbouter:
I can agree to maintain this setting until the next major version change. It's not my first choice, but that's the feedback we got from users, so let's work with that.

Also I think the new setting should be called `PULP_ANALYTICS`. I'm thinking all Pulp specific new settings should be prepended with `PULP_` to allow Pulp to be safely combined with any other random Django project which is a deployment scenario we allow.

mdellweg:
> Also I think the new setting should be called `PULP_ANALYTICS`. I'm thinking all Pulp specific new settings should be prepended with `PULP_` to allow Pulp to be safely combined with any other random Django project which is a deployment scenario we allow.

When translating to environment variables, they will automatically be prefixed with "PULP_". In plugins is saw variables being prefixed by the plugin name. I'm kind of against `PULP_PULP_ANALYTICS=true` and I'm not sure if `PULP_CORE_ANALYTICS=true` would carry the meaning we want.

bmbouter:
> When translating to environment variables, they will automatically be prefixed with "PULP_". In plugins is saw variables being prefixed by the plugin name. I'm kind of against `PULP_PULP_ANALYTICS=true` and I'm not sure if `PULP_CORE_ANALYTICS=true` would carry the meaning we want.

This is a great point, let's not prepend settings with `PULP_`, having a user specify an env var as `PULP_PULP_ANALYTICS` sounds not good. So given that discussion, Having the original setting name of `ANALYTICS` is what makes the most sense to me.

pulpcore/app/settings.py
dkliban: We are keeping it around until 4.0

CHANGES/3417.deprecation
```diff
@@ -0,0 +1 @@
+`TELEMETRY` setting got deprecated in favor of `ANALYTICS`.
```
bmbouter, line 1:
```suggestion
``TELEMETRY`` setting was deprecated in favor of ``ANALYTICS``.
```

It needs double backticks.

CHANGES/3417.feature
```diff
@@ -0,0 +1 @@
+The new `ANALYTICS` setting replaced the `TELEMETRY` setting to avoid confusion with application metrics that will be exposed using `OpenTelemetry`.
```
bmbouter, line 1:
```suggestion
The new ``ANALYTICS`` setting replaced the ``TELEMETRY`` setting to avoid confusion with application metrics that will be exposed using Open Telemetry.
```

bmbouter: Also I think this could be line wrapped maybe.

bmbouter (approved):
I left some small formatting/language requests, but this is looking good. Thanks!

pulpcore/app/util.py
```diff
@@ -247,11 +247,15 @@ def gpg_verify(public_keys, signature, detached_data=None):
...
+    analytics = settings.ANALYTICS
+    if settings.get("TELEMETRY") is not None:
```
bmbouter, line 255: A 1-liner comment in this if block for our future selves would be nice.

mdellweg (approved):
Now we confuse our own CI. (The deprecation check still works...)

PR #3427, opened 2022-11-28
Reimplemented labels as HStore field

### LINKED ISSUE #3400: label-select option hang when using multiple labels
## Summary
when trying to fetch file distribution list with multiple labels (6 labels and above), the requests reaches timeout

## Steps to reproduce
`pulp file distribution list --label-select 'build_time=jf0l5j,builder_hostname=3qshx3,git_ver=foycpk,image_roles=otx8tg,parent_image_checksum=2a0ge1,parent_image_url=i9lug6'`

these are just mock keys and values but the logic is the same

## Expected behavior
need to get response from the pulp server in a reasonable time

## Stacktrace/Error log
client (cli):
```
time pulp file distribution list --label-select 'build_time=jf0l5j,builder_hostname=3qshx3,git_ver=foycpk,image_roles=otx8tg,parent_image_checksum=2a0ge1,parent_image_url=i9lug6'
Error: <html>
<head><title>504 Gateway Time-out</title></head>
<body>
<center><h1>504 Gateway Time-out</h1></center>
<hr><center>nginx</center>
</body>
</html>


real    1m0.440s
user    0m0.272s
sys     0m0.028s
```

server (journal):
```
Nov 14 14:02:05 pulp01 gunicorn[765]: [2022-11-14 14:02:05 +0000] [765] [CRITICAL] WORKER TIMEOUT (pid:23553) 
Nov 14 14:02:06 pulp01 gunicorn[765]: [2022-11-14 14:02:06 +0000] [765] [WARNING] Worker with pid 23553 was terminated due to signal 9 
Nov 14 14:02:06 pulp01 gunicorn[23578]: [2022-11-14 14:02:06 +0000] [23578] [INFO] Booting worker with pid: 23578
```
## Pulp and pulp-cli version info
`pulp3 command line interface, version 0.15.0`
`3.21.4`

## Additonal context
the issue does not reproduce when using only a few labels (until 5 labels)

mdellweg:
This solves the imminent performance issues.

fixes #3400


dralley:
I believe this is going to result in the same kind of downstream impact as the UUID change, which is that plugins with models that inherit directly from `BaseModel` will think they need migrations when used with a newer pulpcore, and also when they have applied the migrations and are used with older pulpcore.

mdellweg:
Yes, when we want to introduce it on the `BaseModel`. That's why i did not do that yet. But yes, we might be bitten again by the abstract base model.

dralley:
Is the plan to put it on `MasterModel` only temporarily?

mdellweg:
> Is the plan to put it on `MasterModel` only temporarily?

I don't have a final Plan. I can even just put it on the very master models we provide label functionality today.

pulpcore/app/viewsets/custom_filters.py
```diff
@@ -273,13 +274,67 @@ class CharInFilter(BaseInFilter, CharFilter):
     pass
 
 
-class LabelSelectFilter(Filter):
+class LabelFilter(Filter):
```
decko, line 277: Are you thinking about write a test for this filter? (or if makes any sense for testing it)

ggainey (dismissed):
Just to clarify for me - we are implementing this at the individual-model-level "for now", for just the models that currently expose labels. "Later", we may rework this to be provided at either Base or Master level, to expand label-access more generally - but that's much more invasive, esp in the migrations, and we want to address the discovered performance-problem "Now". Is that understanding correct?

Other than that - I'll second decko's request for a test for label-filtering.

dralley:
>Just to clarify for me - we are implementing this at the individual-model-level "for now", for just the models that currently expose labels. "Later", we may rework this to be provided at either Base or Master level, to expand label-access more generally - but that's much more invasive, esp in the migrations, and we want to address the discovered performance-problem "Now". Is that understanding correct?

I'd probably go a bit stronger and say that once we pick an approach realistically we won't be able to back down from it.  Django doesn't like moving fields from a child to a parent model table one bit.  It can probably be done, but not without a flag day for all plugins, e.g. Pulp 4.

mdellweg:
> > Just to clarify for me - we are implementing this at the individual-model-level "for now", for just the models that currently expose labels. "Later", we may rework this to be provided at either Base or Master level, to expand label-access more generally - but that's much more invasive, esp in the migrations, and we want to address the discovered performance-problem "Now". Is that understanding correct?
> 
> I'd probably go a bit stronger and say that once we pick an approach realistically we won't be able to back down from it. Django doesn't like moving fields from a child to a parent model table one bit. It can probably be done, but not without a flag day for all plugins, e.g. Pulp 4.

Well, Django allows us to move the field from the (non-abstract) master models up in the hierarchy to the abstract `MasterModel`. But this is already marginally dangerous, because we expose `MasterModel` in the plugin api making this a breaking change. Adding the field to the abstract `BaseModel` is definitely breaking existing plugins. This is almost the `BaseDistribution` dilemma reloaded. Bottom line: Do not change abstract base classes (removing the generic relation is not a database change...).
Upside of this approach: We do not add labels to `Content` (would not match the pulp philosophy) and other tables like `Workers` (that could be potentially interesting if we add atomic label change endpoints). We are intentional about the things we can label.

pulpcore/app/viewsets/custom_filters.py
mdellweg: I am not adding net new functionality here. Tests are here: https://github.com/pulp/pulp_file/blob/main/pulp_file/tests/functional/api/test_labels.py#L125

cc @ggainey

mdellweg: and here: https://github.com/pulp/pulp-cli/blob/main/tests/scripts/pulp_file/test_label.sh

daviddavis:
Actually it looks like the `Label` model doesn't extend `GenericRelationModel` so I was wrong. It still is probably worth adding an index to `GenericRelationModel` though.

ggainey (approved):
We need to get this performance-knee addressed, and this looks like it does the needful. Approved

decko:
> Actually it looks like the `Label` model doesn't extend `GenericRelationModel` so I was wrong. It still is probably worth adding an index to `GenericRelationModel` though.

Maybe an `explain` could provide a better answer to it. :point_up: 

And, do the solution proposed by this PR addresses the performance issues of the `Label` feature?

mdellweg:
> And, do the solution proposed by this PR addresses the performance issues of the `Label` feature?

Yes, it does:
https://github.com/pulp/pulpcore/issues/3400#issuecomment-1328675920

And we may eventually decide to even add an index to the hstore field if it's still not sufficiently performant.

decko (approved):

decko:
> > And, do the solution proposed by this PR addresses the performance issues of the `Label` feature?
> 
> Yes, it does: [#3400 (comment)](https://github.com/pulp/pulpcore/issues/3400#issuecomment-1328675920)
> 
> And we may eventually decide to even add an index to the hstore field if it's still not sufficiently performant.

Sorry @mdellweg. The question was not for you. I should have mention @daviddavis.

Issue #2054, opened 2022-01-17
HTML directory listing for published repositories have no timestamps


pulpbot:
Author: mwaqas@bloombergindustry.com (mwaqas@bloombergindustry.com)

Redmine Issue: 9430, https://pulp.plan.io/issues/9430

---

Ideally published repositories should have something like 'Last modified' tab, so you can actually tell when was the content updated. As of now there are no timestamps. I see content under /pub has that tab, please find screenshot attached. thanks in advance!

PR #3431, opened 2022-11-30
Adds dates to the distribution listing pages

dkliban:
The dates represent the date that the content was created in Pulp.

closes: #2054

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -95,7 +95,9 @@ class DistroListings(HTTPOk):
             .distinct()
         )
         directory_list = (f"{b}/" for b in base_paths)
-        html = Handler.render_html(directory_list)
+        if path == "":
+            path = settings.CONTENT_PATH_PREFIX
+        html = Handler.render_html(directory_list, path=path)
         super().__init__(body=html, headers={"Content-Type": "text/html"})
 
 
@@ -346,7 +348,7 @@ class Handler:
         return headers
 
     @staticmethod
-    def render_html(directory_list):
+    def render_html(directory_list, path="", dates=None):
         """
         Render a list of strings as an HTML list of links.
 
@@ -356,21 +358,30 @@ class Handler:
         Returns:
             String representing HTML of the directory listing.
         """
+        if dates is None:
+            dates = dict()
         template = Template(
             """
-        <!DOCTYPE html>
-        <html>
-            <body>
-                <ul>
-                {% for name in dir_list %}
-                    <li><a href="{{ name|e }}">{{ name|e }}</a></li>
-                {% endfor %}
-                </ul>
-            </body>
-        </html>
-        """
+<html>
+<head><title>Index of {{ path }}</title></head>
+<body bgcolor="white">
+<h1>Index of {{ path }}</h1>
+<hr><pre>
+<a href="../">../</a>
+{% for name in dir_list -%}
+{% if dates.get(name, "") -%}
+{% set date =  dates.get(name).strftime("%d-%b-%Y %H:%M") -%}
+{% else -%}
+{% set date = "" -%}
+{% endif -%}
+<a href="{{ name|e }}">{{ name|e }}</a>{% for number in range(100 - name|e|length) %} """
+            """{% endfor %}{{ date }}
+{% endfor -%}
+</pre><hr></body>
+</html>
+"""
         )
-        return template.render(dir_list=sorted(directory_list))
+        return template.render(dir_list=sorted(directory_list), dates=dates, path=path)
 
     async def list_directory(self, repo_version, publication, path):
         """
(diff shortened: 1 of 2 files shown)


dkliban:
Here is what the results look like:

image

pulpcore/content/handler.py
```diff
@@ -95,7 +95,9 @@ def __init__(self, path, distros):
             .distinct()
         )
         directory_list = (f"{b}/" for b in base_paths)
-        html = Handler.render_html(directory_list)
+        if path == "":
+            path = settings.CONTENT_PATH_PREFIX
+        html = Handler.render_html(directory_list, path=path)
```
gerrod3, line 100: Can this also include the `pulp_created` of the distributions for dates as well?

pulpcore/content/handler.py
```diff
@@ -356,21 +358,30 @@ def render_html(directory_list):
...
-                {% for name in dir_list %}
-                    <li><a href="{{ name|e }}">{{ name|e }}</a></li>
-                {% endfor %}
-                </ul>
-            </body>
-        </html>
-        """
+<html>
+<head><title>Index of {{ path }}</title></head>
+<body bgcolor="white">
```
gerrod3, line 367: Does this need `<!DOCTYPE html>` at the top? Also, can indentation be added back to the template?

daviddavis:
I tested this out and it works great but there's a problem for our use case (and perhaps the original reporter's use case too). 

The dates being shown for packages are when the published artifact is created but what we care about is the package first gets added to (or rather published in) a repo. In other words, just because a repo gets published again, the timestamp for a package that is already in the repo shouldn't change. This is similar to how if you were using createrepo + a web server: running createrepo to publish/re-publish the repo wouldn't change the timestamp on the package.

I don't know how feasible it is to do this though?

pulpcore/content/handler.py
```diff
@@ -517,12 +534,13 @@ def get_latest_publication_or_version_blocking():
                     rel_path = index_path
                     headers = self.response_headers(rel_path)
                 except ObjectDoesNotExist:
-                    dir_list = await self.list_directory(None, publication, rel_path)
+                    dir_list, dates = await self.list_directory(None, publication, rel_path)
                     dir_list.update(
                         await sync_to_async(distro.content_handler_list_directory)(rel_path)
                     )
```
daviddavis, line 540: Maybe allow `content_handler_list_directory` to return dates?

```suggestion
                    vals = await sync_to_async(distro.content_handler_list_directory)(rel_path)
                    if isinstance(vals, set):
                        dir_list.update(vals)
                    else:
                        dir_list.update(vals[0])
                        dates.update(vals[1])          
```

dkliban:
@daviddavis I believe you are interested in the date that the content was added to the repository version. Am I understanding correctly?

daviddavis:
@dkliban ideally, we'd have the date that the content was first published (ie exposed to the world). However, the date that the content was added to the repo might suffice. Let me doublecheck.

daviddavis:
@dkliban so I got a response and it sounds like the content creation time is what we want.

dkliban:
@daviddavis that makes sense to me. i'll update the PR.

pulpcore/content/handler.py
dkliban: I was not sure how much value that would provide. Maybe the pulp_last_updated would be better?

dkliban:
@mdellweg I changed which date is now displayed. It will always show the date that the content was created in Pulp. Please re-review.

daviddavis (approved):

pulpcore/content/handler.py
dkliban: Actually, I would prefer to not make that decision now and leave off the dates on the list of distributions.

pulpcore/content/handler.py
```diff
@@ -398,31 +409,38 @@ def list_directory_blocking():
...
             if publication:
-                pas = publication.published_artifact.filter(relative_path__startswith=path)
+                pas = publication.published_artifact.select_related("content_artifact").filter(
+                    relative_path__startswith=path
+                )
                 for pa in pas:
-                    directory_list.add(file_or_directory_name(path, pa.relative_path))
+                    name = file_or_directory_name(path, pa.relative_path)
+                    directory_list.add(name)
+                    dates.update({name: pa.content_artifact.pulp_created})
```
mdellweg, line 421: Should this be the artifact or the content?

daviddavis: Hmm. What would be the tradeoffs between using artifact over content or content over artifact?

To be honest, as everyday Pulp users, we never interact with artifacts. The endpoints we use are the one-shot upload endpoints, content list/show, and then the repository modify endpoint. So artifacts are completely obscure to us as we don't even expose artifacts to our users. And so I think content creation timestamp is probably more useful to us.

mdellweg: I agree. Just because two content units happen to use the same bits should not determine their date.

Also great feedback, as with our RBAC story we assume exactly that (less privileged) users may not interact with artifacts or orphaned content directly.

daviddavis: So what @dkliban has here works then? I assume that the `content_artifact.pulp_created` would be the same as the content `pulp_created`?

mdellweg: I believe so.

mdellweg (approved):
Reinstating my approval.

bmbouter (changes requested):
Please add a test for this.

pulpcore/content/handler.py
dkliban: @gerrod3 I am torn about this. You are basing your suggestions on what Apache serves up when it lists directories. My changes to the template make it look like the page that is served up by nginx. I used https://fixtures.pulpproject.org as my example.

bmbouter (approved):
Thank you!

Issue #3173, opened 2022-09-06
Content app should (possibly) show index pages when the path is provided without a trailing slash


dralley:
***OR*** redirect from the non-trailing-slash URL to the trailing slash URL.

**Version**
Any

**Describe the bug**
It is completely valid in general to not yield an index page when the URL doesn't include a trailing slash, as the path looks like a file.  However, automatically redirecting to include this trailing slash is the default behavior of most webservers, including Pulp 2.

**To Reproduce**
Use curl / wget / http against a content app path that ought to display an HTML index page, receive 404

**Expected behavior**
Receive the index page or a redirect to the appropriate URL for the index page.

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=2104792

ipanova:
This work was reverted, re-opening the issue  for better tracking

lubosmj:
This depends on https://github.com/pulp/pulpcore/issues/3459

dralley:
@dkliban This is a pretty meaty PR that, I believe, might require plugin author intervention - which is why it went into 3.40, the breaking change release.  Yes?

I'm guessing then that we don't necessarily want to backport this to 3.39?  It's currently targeted for 6.15.0 so it would be nice if we could... and we'd need to move it off if we can't... but it's also not that big a deal, we have no strong demand for this as far as I can tell.

dkliban:
This cannot be backported to 3.39 because pulp_container requires a change that's only available in versions that are installable with pulpcore 3.40+

dralley:
Ack, thanks.

PR #3434, opened 2022-12-02

dkliban:

fixes: #3173

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -531,34 +531,6 @@ class Handler:
             await sync_to_async(get_latest_publication_or_version_blocking)()
 
         if publication:
-            ends_in_slash = rel_path == "" or rel_path.endswith("/")
-            if ends_in_slash is False:
-                rel_path = f"{rel_path}/"
-            try:
-                index_path = "{}index.html".format(rel_path)
-
-                await sync_to_async(publication.published_artifact.get)(relative_path=index_path)
-                if ends_in_slash is False:
-                    # index.html found, but user didn't specify a trailing slash
-                    raise HTTPMovedPermanently(f"{request.path}/")
-                rel_path = index_path
-                headers = self.response_headers(rel_path)
-            except ObjectDoesNotExist:
-                dir_list, dates = await self.list_directory(None, publication, rel_path)
-                dir_list.update(
-                    await sync_to_async(distro.content_handler_list_directory)(rel_path)
-                )
-                if dir_list and ends_in_slash is False:
-                    # Directory can be listed, but user did not specify trailing slash
-                    raise HTTPMovedPermanently(f"{request.path}/")
-                elif dir_list:
-                    return HTTPOk(
-                        headers={"Content-Type": "text/html"},
-                        body=self.render_html(dir_list, path=request.path, dates=dates),
-                    )
-            if ends_in_slash is False:
-                rel_path = rel_path[:-1]
-
             # published artifact
             try:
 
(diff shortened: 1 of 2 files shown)


dralley:
@dkliban Does this fix https://github.com/pulp/pulpcore/issues/3173 ?

dkliban:
@dralley yes it does.

dralley:
Great, would you mind making sure both get closed and updating the BZ?

dralley:
adding an extra fixes / closes tag to the commit, or marking duplicate, whichever you prefer

.ci/ansible/settings.py.j2
```diff
@@ -8,7 +8,6 @@ TOKEN_SIGNATURE_ALGORITHM = "ES256"
 CACHE_ENABLED = True
 REDIS_HOST = "localhost"
 REDIS_PORT = 6379
-TELEMETRY = False
```
mdellweg, line 11: This needs to be `ANALYTICS=False`.
But we should solve it in the template.

mdellweg (changes requested):

daviddavis:
I tested this and it works great 👍 

I found [one other issue that the dates change apparently introduced](https://github.com/pulp/pulpcore/issues/3436) but it's relatively minor I think and can be fixed in a z-stream.

daviddavis (approved):

pulpcore/content/handler.py
```diff
@@ -292,15 +296,24 @@ def _match_distribution(cls, path):
...
             )
-
-        raise PathNotResolved(path)
+            if path_ends_in_slash:
+                raise PathNotResolved(path)
+        try:
+            if not path_ends_in_slash:
+                path = path[:-1]
+            distro_model.objects.get(base_path=path)
+            raise HTTPMovedPermanently(f"{settings.CONTENT_PATH_PREFIX}{path}/")
```
gerrod3, line 314: What is going on here? Should this not be covered by the above `try/except` block?

dkliban: Your observation is correct. I'll update my PR to always raise PathNotResolved on line 308 (without checking if the path ends in a slash).

pulpcore/content/handler.py
```diff
@@ -524,25 +530,31 @@ def get_latest_publication_or_version_blocking():
...
 
         if publication:
-            if rel_path == "" or rel_path[-1] == "/":
-                try:
-                    index_path = "{}index.html".format(rel_path)
-
-                    await sync_to_async(publication.published_artifact.get)(
-                        relative_path=index_path
-                    )
+            ends_in_slash = rel_path == "" or rel_path[-1] == "/"
```
dralley, line 533: Why do we need to perform this check twice?  Can the redirect happen earlier?

Can we use `.endswith()` consistently instead of using `[-1]` here

dkliban: I'll update to use .endswith(). 

The redirect can't happen earlier because we don't know if there is an index.html until later.

pulpcore/content/handler.py
```diff
@@ -524,25 +531,33 @@ def get_latest_publication_or_version_blocking():
...
+                if dir_list and ends_in_slash is False:
+                    # Directory can be listed, but user did not specify trailing slash
+                    raise HTTPMovedPermanently(f"{request.path}/")
+                elif dir_list:
                     return HTTPOk(
                         headers={"Content-Type": "text/html"},
                         body=self.render_html(dir_list, path=request.path, dates=dates),
                     )
+            if ends_in_slash is False:
+                rel_path = rel_path[:-1]
```
mdellweg, line 560: So this is un-doing the thing from before?
So i understand correctly, that for every request inside a publication we first try to find `f"{url/index.html"` and only then attempt to serve the proper published artifact?
I think this index thing should be a fallback basically halfing the number of accesses to the `published_artifact` table.

dkliban: yes, it is undoing the path manipulation from above. I believe you have identified a performance regression that this code introduces. I will move the search for index.html and directory listing to be after we try looking for published artifacts.

mdellweg (approved):

pulpcore/content/handler.py
```diff
@@ -278,6 +279,9 @@ def _match_distribution(cls, path):
             DistroListings: when multiple matches are possible.
             PathNotResolved: when not matched.
         """
+        path_ends_in_slash = path.endswith("/")
+        if not path_ends_in_slash:
+            path = f"{path}/"
         base_paths = cls._base_paths(path)
```
dralley, line 285: This lookup should be unaffected by the transformation, yes?

ipanova: In pulp-container base_paths are treated completely differently, they should not end with /, hence we are overriding this method.
This change breaks the plugin because there is an assumption that every plugin uses base_bath in the same way.

PR #3435, opened 2022-12-05
Remove the ansible collection from the recommendation

### LINKED ISSUE #3430: Remove "Recommended" from Ansible Installation
https://docs.pulpproject.org/pulpcore/installation/instructions.html#ansible-installation-recommended

fao89:
closes #3430

### DIFF
--- a/CHANGES/3430.doc
+++ b/CHANGES/3430.doc
@@ -1,0 +1,1 @@
+Remove the ansible collection from the recommendation
--- a/docs/installation/instructions.rst
+++ b/docs/installation/instructions.rst
@@ -15,16 +15,15 @@ Fedora, CentOS, and Mac OSX.
     There are risks associated with this decision. See your distribution's docs for more details
 
 
-Ansible Installation (Recommended)
-----------------------------------
+OCI Images
+----------
+For comprehensive and up-to-date instructions about using the Pulp OCI Images, see the
+`Pulp OCI Images documentation <https://docs.pulpproject.org/pulp_oci_images/>`__.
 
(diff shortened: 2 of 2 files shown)


mdellweg (approved):

gerrod3:
The ansible installer is still valid for 3.22, no? I think we should accept this PR after 3.22 has been release. Also, we should add documentation for the new recommend way of installing Pulp.

fao89:
> The ansible installer is still valid for 3.22, no? I think we should accept this PR after 3.22 has been release. Also, we should add documentation for the new recommend way of installing Pulp.

why keeping point users to pulp_installer when it is about to be gone?
If they follow this now, they will install 3.22, and couple months after they will worry about migrating

gerrod3:
> why keeping point users to pulp_installer when it is about to be gone? If they follow this now, they will install 3.22, and couple months after they will worry about migrating

I would like to see documentation for the install method replacing it. If that is coming in a separate PR, then I can accept this one now.

ipanova:
We could keep the ansible installation in docs and in parenthesis mention it is going to be the last 3.22 release.
In a separate PR when adding the replacement ansible install can go away.

fao89:
> We could keep the ansible installation in docs and in parenthesis mention it is going to be the last 3.22 release. In a separate PR when adding the replacement ansible install can go away.

Who would click in installation section in pulpcore docs?
A: **New** users

Why a new user wants to know about the installation that won't be supported anymore?

ipanova (approved):

PR #3448, opened 2022-12-07
Zero downtime upgrade docs for plugin writers

### LINKED ISSUE #3368: When upgrading without draining the task queue tasks can fail on version conflicts
# Problem
When doing an upgrade, there may still be waiting tasks added by the api before the upgrade. Now running the workers with the upgraded codepaths, they may run into problems when a task function was removed / renamed or it's signature changed.
Even the semantics may have changed and lead to very subtle bugs.

# Solution
All worker code must be backwards compatible and properly handle tasks dispatched using an older codebase.

If we do not define a compatibility window here, workers need to establish compatibility forever.

### LINKED ISSUE #3443: Add zero-downtime section to the plugin writer docs
In the [plugin writer docs](https://github.com/pulp/pulpcore/blob/main/docs/plugin_dev/plugin-writer/concepts/index.rst), add a section about zero-downtime upgrades. This doc should be a clearly written reduction of the [content from this hackmd](https://hackmd.io/dxm3JG0bSEizAJeM6E9HUw). It should include these sections.

1. Explain what a "breaking migration" is and the data-layer, code-layer compatibility requirement.
2. Outline the several options plugin writers can use when they find themselves writing a breaking change migration (see hackmd content).

bmbouter:
This adds docs for two situations:

* Guidance for zero downtime migrations
* tasking backwards compatibility requirements

closes #3368
closes #3443

### DIFF
--- a/CHANGES/plugin_api/3368.doc
+++ b/CHANGES/plugin_api/3368.doc
@@ -1,0 +1,1 @@
+Adds docs that plugin writers tasks must be backwards compatible until the next major Pulp version.
--- a/docs/plugin_dev/plugin-writer/concepts/index.rst
+++ b/docs/plugin_dev/plugin-writer/concepts/index.rst
@@ -678,3 +678,99 @@ statements, migration output print statements, django management commands, etc. 
 internationalized will remain in English. This expectation was formed after feedback from
 multi-language speakers who believe having error messages for admins in English would reduce the
 time to finding a fix and was generally less surprising.
+
+
+.. _zero-downtime-upgrades:
+
+Zero-Downtime Upgrades
+----------------------
+
+Eventually, Pulp users will be able to upgrade without first stopping Pulp services. This has been
+`requested from the community <https://discourse.pulpproject.org/t/support-zero-downtime-updates/645>`_,
+To work towards that goal, developers of ``pulpcore`` or a plugin should follow these requirements:
+
+* Migrations must not break earlier versions code still running during an upgrade.
+* Task code must be backwards compatible until the next major Pulp version.
+
+Future user upgrades will likely run as follow:
+
+1. Run the migrations while old pulp code is online. Old code, is using the new data format.
+2. Rolling restart old code to become new code. Old and new code is running at the same time!
+
+
+Zero-Downtime Migrations
+========================
+
+The significant challenge with online migrations is that the db state the migration applies has to
+work with both newer and older versions of Pulp code. For example, consider a model field that is to
+be renamed. After renaming the field, Django would generate a migration that renames the specified
+column. This would break all previous code which expects the previous column name.
+
+Before getting into specific suggestions, the general pattern is to do the following:
+
+1. The migration should be split into two parts: a "compatible with earlier code migration" and a
+   "breaking earlier code migration". In continuing the column rename example, it would become a
+   "create a new column" migration, and later a "delete the original column" migration.
+
+2. The "breaking earlier code migration" should be delivered in a later release. It contains the
+   the component versions that would not be broken by that change. When that migration goes to run
+   it uses the db info in the db for each running pulp process to determine if any components are
+   running a version that would break if this change is applied.
+
+The solution to this will be highly dependant on the details of the migration, but here are some
+likely patterns to be applied:
+
+1. Avoid it. Is this rename really that important? Is it worth the trouble?
+
+2. Rename the model attributes in code, but leave the actual column name as-is with
+   the `db_column <https://docs.djangoproject.com/en/3.2/ref/models/fields/#db-column>`_.
+
+3. Have an "old" and a "new" column and use database triggers to keep data written to one to also be
+   written to the other and vice-versa.
+
+Here's an example:
+
+pulp_file wants to rename a model attribute ``old`` to be called ``new`` in the next pulp_file
+release, let's say that's pulp_file 1.10.0. Let's assume that avoiding the rename altogether or
+using the ``db-column`` option to just rename it in the ORM are not viable.
+
+This could be done as follows:
+
+* Add the ``new`` field next to the ``old`` field and have Django auto-create a migration adding
+  ``new``.
+
+* The same migration needs to install a new trigger that anytime ``old`` is written to, ``new`` is
+  also written to and vice-versa. For example, something `like this <https://www.appsloveworld.com/postgresql/100/94/how-to-dual-write-to-two-columns-in-one-table-using-postgres-trigger>`_
+  This allows the new code to read/write exclusively with `new` and the old code to deal with `old`.
+
+* Write a data migration that updates the ``new`` column with ``old`` data in batches. Use batching
+  to avoid a long table-lock.
+
+* Have the codebase of pulp_file 1.10.0 stop using ``old`` entirely.
+
+At a later time, e.g. pulp_file 1.13.0, a migration will be shipped to remove column ``old``. That
+migration needs to do two things things:
+
+* Prior to running ensure via the database records that there are no pulp components running with
+  pulp_file < 1.10.0. If there are, abort running the migration and notify the user they need to
+  upgrade to a version pulp_file>=1.10,<1.13.0.
+
+* Remove the database trigger and then the column ``old``.
+
+
+Data Migrations
+===============
+
+One problem that can arise from data migrations is the use of table-locks which would prevent other
+code still running from executing concurrently. The typical solution is to have data migrations
+operate in transactional batches which avoids a table-lock and instead creates row-locks.
+
+
+Tasking System
+==============
+
+Tasking also has some considerations to allow code upgrades; specifically, tasks dispatched from
+older codebases could run on newer, upgraded workers. To ensure this always works tasks must be
+forever backwards compatible until the next major Pulp version. For example, you cannot have a
+breaking signature change in tasking code and if this is needed you need to make a new task name and
+preserve the old code until the next major Pulp version.
--- a/CHANGES/plugin_api/3443.doc
+++ b/CHANGES/plugin_api/3443.doc
@@ -1,0 +1,2 @@
+Adds docs on the need and possible options for writing migrations in a way that they could be
+applied to a running Pulp system.


CHANGES/plugin_api/3368.doc
```diff
@@ -0,0 +1,2 @@
+Adds docs that plugin writers tasks must be backwards compatible until Pulp 4.0. This practice will
```
mdellweg, line 1: Can you wrap this line at 98 characters?
Towncrier will add two spaces eventually.

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,90 @@ statements, migration output print statements, django management commands, etc.
...
 time to finding a fix and was generally less surprising.
+
+.. _zero-downtime-upgrades:
+
+Zero-Downtime Upgrades
+----------------------
+
+As `requested from the community <https://discourse.pulpproject.org/t/support-zero-downtime-updates/645>`_,
+in the future, Pulp will likely start allowing upgrades without stopping Pulp services. This will
+cause all Pulp code to take on two new requirements:
```
mdellweg, line 689:
```suggestion
cause all Pulp code to take on new requirements:
```
I'm almost sure this list wlll expand...

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,90 @@ statements, migration output print statements, django management commands, etc.
...
+
+* Migrations cannot break earlier versions of running code.
```
mdellweg, line 691:
```suggestion
* Migrations must not break earlier versions code possibly still running during an upgrade.
```

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,90 @@ statements, migration output print statements, django management commands, etc.
...
+* Task code must be backwards compatible until Pulp 4.0
```
mdellweg, line 692:
```suggestion
* Task code must be backwards compatible until the next major version of Pulp.
```

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,90 @@ statements, migration output print statements, django management commands, etc.
...
+
+Future user upgrades will likely run as follow:
+
+1. Run the migrations while old pulp code is online. Old code, new data format
```
mdellweg, line 696:
```suggestion
1. Run the migrations while old pulp code is online. Old code, is using the new data format.
```

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,90 @@ statements, migration output print statements, django management commands, etc.
...
+2. Rolling restart old code to become new code. Old and new code is running at the same time!
+
+
+Zero-Downtime Migrations
+========================
+
+The significant challenge with online migrations is that the db state the migration applies has to
+work with both newer and older versions of Pulp code. For example, consider a model field that is to
+be renamed. After renaming the model, Django would generate a migration that renames the column to
+be made. This would break all previous code which expects the previous column name.
```
mdellweg, line 706:
```suggestion
be renamed. After renaming the field, Django would generate a migration that renames the specifiied
column. This would break all previous code which expects the previous column name.
```

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,90 @@ statements, migration output print statements, django management commands, etc.
...
+
+2. The "breaking earlier code migration" should be delivered in a later release. It contains the
+   the component versions that would not be broken by that change. When that migration goes to run
+   it uses the db info in the db for each running pulp process to determine if any components are
+   running a version that would break if this change is applied.
+
+The solution to this will be highly dependant on the details of the migration, but here are some
+likely patterns to be applied:
+
+1. Avoid it. Is this rename really that important? Is it worth the effort?
```
mdellweg, line 722: Really just a suggestion...
```suggestion
1. Avoid it. Is this rename really that important? Is it worth the trouble?
```

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,90 @@ statements, migration output print statements, django management commands, etc.
...
+
+2. Rename the model attributes in code and serializers, but leave the actual column name as-is with
+   the `db_column <https://docs.djangoproject.com/en/3.2/ref/models/fields/#db-column>`_.
```
mdellweg, line 725: This is almost impossible due to semver on the api anyway, right? But it could be some other api requesting it from us.

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,90 @@ statements, migration output print statements, django management commands, etc.
...
+
+3. Have an "old" and a "new" column and use database triggers to keep data written to one to also be
+   written to the other and vice-versa.
+
+4. For data migrations that would normally use table-locks, break these up into chunks.
```
mdellweg, line 730: I do not see this as an individual solution to any problem. When you do chunked updates of data, the old code would only see the not yet migrated items, while the new code would only see the already migrated chunks.
It does not sound right to me.

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,90 @@ statements, migration output print statements, django management commands, etc.
...
+
+This could be done as follows:
+
+* Replace the ``old`` field with the ``new`` field, but instead of Django generating a migration
+  create one manually that adds the ``new`` column and marks for Django that ``old`` is to be
+  treated as deleted without deleting it.
+
+* The same migration needs to transactionally install a new trigger that anytime ``old`` is written
+  to, ``new`` is also written to and vice-versa. For example, something `like this <https://www.appsloveworld.com/postgresql/100/94/how-to-dual-write-to-two-columns-in-one-table-using-postgres-trigger>`_
+  This allows the new code to read/write exclusively with `bar` and the old code to deal with `foo`.
```
mdellweg, line 746:
```suggestion
  This allows the new code to read/write exclusively with `new` and the old code to deal with `old`.
```

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,90 @@ statements, migration output print statements, django management commands, etc.
...
+
+pulp_file wants to rename a model attribute ``old`` to be called ``new`` in the next pulp_file
+release, let's say that's pulp_file 1.10.0. Let's assume that avoiding the rename altogether or
+using the ``db-column`` option to just rename the ORM name are not viable.
+
+This could be done as follows:
+
+* Replace the ``old`` field with the ``new`` field, but instead of Django generating a migration
+  create one manually that adds the ``new`` column and marks for Django that ``old`` is to be
+  treated as deleted without deleting it.
```
mdellweg, line 742: I was wondering if we would keep the column around in the ORM layer and later delete it from there.
I believed this would be much easier with the Django generated migrations.

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,90 @@ statements, migration output print statements, django management commands, etc.
...
+* Have the codebase of pulp_file 1.10.0 drop the usage of ``old`` entirely.
+
+At a later time, e.g. pulp_file 1.13.0, a migration will be shipped to remove column ``old``. That
+migration needs to do two things things:
+
+* Prior to running ensure via the database records that there are no pulp components running with
+  pulp_file < 1.10.0. If there are abort, not running the migration and notify the user they need to
+  upgrade to a version pulp_file>=1.10,<1.13.0.
+
+* Remove the column ``old`` and the database trigger.
```
mdellweg, line 757: Probably the other way around...

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,90 @@ statements, migration output print statements, django management commands, etc.
...
+
+
+Tasking System
+==============
+
+Tasking also has some considerations to allow code upgrades; specifically, tasks dispatched from
+older codebases could be runnable on newer, upgraded workers. To ensure this always works tasks must
+be forever backwards compatible until Pulp 4.0. For example, you cannot have a breaking signature
+change in tasking code and if this is needed you need to make a new task name and preserve the old
+code until Pulp 4.0.
```
mdellweg, line 767: I think at some point it should be ok to cancel a very old task with an appropriate error message.

mdellweg (changes requested):

docs/plugin_dev/plugin-writer/concepts/index.rst
bmbouter: I removed the `possibly` since we'll be able to know for sure if its' running or not from the db.

docs/plugin_dev/plugin-writer/concepts/index.rst
bmbouter: +1 it's more than just effort it's also uncertainty and support over time. Change made.

docs/plugin_dev/plugin-writer/concepts/index.rst
bmbouter: Agreed. Based on that, I removed the `and serializers` words since there could be attributes on models we don't expose or expose directly. In those cases I think this idea makes sense. If you think we should modify this more, please lmk.

docs/plugin_dev/plugin-writer/concepts/index.rst
bmbouter: I modified to be `the next major version of Pulp.`. Yes we may need to at some point, but I'd prefer to not add that into the toolbox as an option right from the start if that's ok. Let me know if you still want to.

docs/plugin_dev/plugin-writer/concepts/index.rst
bmbouter: I agree, I rewrote this part so let me know what you think. Also I realized I never told developers to actually copy the data from `old` to `new` so I added that step in here too.

CHANGES/plugin_api/3368.doc
```diff
@@ -0,0 +1,2 @@
+Adds docs that plugin writers tasks must be backwards compatible until Pulp 4.0. This practice
+will eventually enable zero-downtime upgrades.
```
mdellweg, line 2: Actually, this is an issue even without zero-downtime. Maybe just scrap that last sentence, or stat that it provides more stable upgrades in general.

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,99 @@ statements, migration output print statements, django management commands, etc.
...
+  table-lock.
+
+* Have the codebase of pulp_file 1.10.0 stop using ``old`` entirely.
+
+At a later time, e.g. pulp_file 1.13.0, a migration will be shipped to remove column ``old``. That
+migration needs to do two things things:
+
+* Prior to running ensure via the database records that there are no pulp components running with
+  pulp_file < 1.10.0. If there are abort, not running the migration and notify the user they need to
+  upgrade to a version pulp_file>=1.10,<1.13.0.
```
mdellweg, line 756: The sentence structure is a little unclear to me. How about:
```suggestion
  pulp_file < 1.10.0. If there are, abort running the migration and notify the user they need to
  upgrade to a version pulp_file>=1.10,<1.13.0.
```

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,99 @@ statements, migration output print statements, django management commands, etc.
...
+
+* Add the ``new`` field next to the ``old`` field and have Django auto-create a migration adding
+  ``new``.
+
+* The same migration needs to install a new trigger that anytime ``old`` is written to, ``new`` is
+  also written to and vice-versa. For example, something `like this <https://www.appsloveworld.com/postgresql/100/94/how-to-dual-write-to-two-columns-in-one-table-using-postgres-trigger>`_
+  This allows the new code to read/write exclusively with `new` and the old code to deal with `old`.
+
+* Write a data migration that copies ``old`` to ``new`` in bathes. Use batching to avoid a long
+  table-lock.
```
mdellweg, line 747: With the trigger in place it may suffice to "update" the old columns.

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,99 @@ statements, migration output print statements, django management commands, etc.
...
+code still running from executing concurrently. The typical solution is to have data migrations
+operate in transactional batches which avoids a table-lock and instead creates row-locks.
+
+
+Tasking System
+==============
+
+Tasking also has some considerations to allow code upgrades; specifically, tasks dispatched from
+older codebases could be runnable on newer, upgraded workers. To ensure this always works tasks must
+be forever backwards compatible until Pulp 4.0. For example, you cannot have a breaking signature
```
mdellweg, line 774: ... the next major version ...

I always try to avoid explicit versions in docs. You will most definitely miss updating them properly in the future.

mdellweg (changes requested):
Please look for "may" or "maybe" if you think some suggestions are not required...

mdellweg (approved):

CHANGES/plugin_api/3368.doc
```diff
@@ -0,0 +1 @@
+Adds docs that plugin writers tasks must be backwards compatible until Pulp 4.0.
```
ipanova, line 1: until the next Pulp major version

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,99 @@ statements, migration output print statements, django management commands, etc.
...
+
+
+.. _zero-downtime-upgrades:
+
+Zero-Downtime Upgrades
+----------------------
+
+As `requested from the community <https://discourse.pulpproject.org/t/support-zero-downtime-updates/645>`_,
+in the future, Pulp will likely start allowing upgrades without stopping Pulp services. This will
+cause all Pulp code to take on new requirements:
```
ipanova, line 690: I am having troubles to understand the goal of these 2 sentences. Is this a future promise but we are still not there? "will likely start allowing" does not give much of the info to the users. How can we re-phrase this better?

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,99 @@ statements, migration output print statements, django management commands, etc.
...
+One problem that can arise from data migrations is the use of table-locks which would prevent other
+code still running from executing concurrently. The typical solution is to have data migrations
+operate in transactional batches which avoids a table-lock and instead creates row-locks.
+
+
+Tasking System
+==============
+
+Tasking also has some considerations to allow code upgrades; specifically, tasks dispatched from
+older codebases could be runnable on newer, upgraded workers. To ensure this always works tasks must
```
ipanova, line 773: `could run` otherwise `they should be runnable` as "-able" points to ability not to the possibility.

docs/plugin_dev/plugin-writer/concepts/index.rst
bmbouter: I rewrote this paragraph to more clearly state what is hapening. What do you think?

bmbouter:
@ipanova can I have one more review when you get a chance please?

ipanova (approved):

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -678,3 +678,100 @@ statements, migration output print statements, django management commands, etc.
...
+
+
+.. _zero-downtime-upgrades:
+
+Zero-Downtime Upgrades
+----------------------
+
+Eventually, Pulp users will be able to upgrade without first stopping Pulp services. This has been
+`requested from the community <https://discourse.pulpproject.org/t/support-zero-downtime-updates/645>`_,
+To work towards that goal, developers of ``pulpcore`` and ``pulp_file`` should take on the following
```
ipanova, line 690: I am not sure whether it's worth specifically accentuating on the pulp-file as a plugin, I think every plugin will need to adhere to this requirements.

bmbouter: I revised this a bit more. With "should" remaining in the sentence, having it stated generally as a requirement for all plugins is in-line with the original thinking. +1

ipanova: yes, thank you.

bmbouter:
Thanks for the reviews!

PR #3451, opened 2022-12-09
Adds pulp-smash as unittest dependency

bmbouter:
Without it when someone runs `oci-env test -i -p pulpcore unit` it fails with an error saying the pytest hook
`pytest_check_for_leftover_pulp_objects` is undefined.

[noissue]

### DIFF
--- a/unittest_requirements.txt
+++ b/unittest_requirements.txt
@@ -1,4 +1,5 @@
 # Unit test requirements
 asynctest
 mock
+pulp-smash @ git+https://github.com/pulp/pulp-smash.git
 pytest-django


bmbouter:
Actually here's a change in `oci_env` that installs pulp_smash for unit tests all the time...  https://github.com/pulp/oci_env/pull/65 When I use the commit prior to this, 0031a1bfd32ddd52d7ce2aa885c50e8a54d0dfad I can reproduce the problem. The latest commit has no issue running `oci-env test -i -p pulpcore unit` which was my whole interest here

 I'm closing this PR, it's a resolved issue.

PR #3463, opened 2022-12-14
Fixup pulp_label migration

mdellweg:
Added a repair command and fixed the dependency issue.

[noissue]

### DIFF
--- a/pulpcore/app/management/commands/datarepair-labels.py
+++ b/pulpcore/app/management/commands/datarepair-labels.py
@@ -1,0 +1,77 @@
+from gettext import gettext as _
+
+from django.apps import apps
+from django.contrib.contenttypes.models import ContentType
+from django.db import transaction
+from django.core.management import BaseCommand
+from django.db.models.expressions import OuterRef, RawSQL
+
+
+class Command(BaseCommand):
+    """
+    Django management command for repairing unmigrated pulp_labels.
+    """
+
+    help = _(
+        "Attempts to migrate pulp_labels from a generic relation to an hstore field on their "
+        "respective model."
+    )
+
+    def add_arguments(self, parser):
+        """Set up arguments."""
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help=_("Don't modify anything, just collect results on how many Remotes are impacted."),
+        )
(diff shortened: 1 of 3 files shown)


pulpcore/app/migrations/0098_pulp_labels.py
```diff
@@ -53,5 +84,5 @@ class Migration(migrations.Migration):
             name='pulp_labels',
             field=django.contrib.postgres.fields.hstore.HStoreField(default=dict),
         ),
-        migrations.RunPython(code=copy_labels_up, reverse_code=migrations.RunPython.noop, elidable=True),
+        migrations.RunPython(code=copy_labels_up, elidable=True),
```
gerrod3, line 87: Why has the noop been dropped?

pulpcore/app/management/commands/datarepair-labels.py
```diff
@@ -0,0 +1,77 @@
...
+                        Label.objects.filter(content_type=ctype, object_id=OuterRef("pulp_id"))
+                        .annotate(label_data=RawSQL("hstore(array_agg(key), array_agg(value))", []))
+                        .values("label_data")
+                    )
+                    model.objects.update(pulp_labels=label_subq)
+                    Label.objects.filter(content_type=ctype).delete()
+
+        if Label.objects.count() and purge and not dry_run:
+            print("Purge remaining labels.")
+            Label.objects.all().delete()
```
lubosmj, line 77: This deletes all the objects, is it expected? What is its real use?

pulpcore/app/migrations/0098_pulp_labels.py
mdellweg: It shouldn't have been there i used it for debugging.

pulpcore/app/management/commands/datarepair-labels.py
mdellweg: Only if you put `--purge` on the cmdline intentionally.

I expect that we remove the `Label` model in 3.25, but only if it's empty. We need to give admins a way to unblock themselves.

lubosmj (approved):

gerrod3 (approved):

PR #3466, opened 2022-12-14
Add version information to workers and content-app

### LINKED ISSUE #3365: As a plugin writer, I can see the plugins and versions installed for each pulp-worker, pulp-content, and pulp-api
## Problem

In order to support the goals of [ticket 3401](https://github.com/pulp/pulpcore/issues/3401), we need the full listing of pulpcore and plugins and their versions available with all the services: pulp-worker, pulp-content, or pulp-api.

## Solution

Add to the database records of [ContentAppStatus](https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/status.py#L31) and [Worker](https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/task.py#L79) the pulpcore and plugin versions installed for that process.

The pulp-api entries will come as their own ticket. That ticket is [here](https://github.com/pulp/pulpcore/issues/3175).

mdellweg:
fixes #3365


pulpcore/content/__init__.py
```diff
@@ -39,17 +39,22 @@ async def _heartbeat():
...
                 ContentAppStatus.objects.get_or_create
-            )(name=name)
+            )(name=name, defaults={"versions": versions})
 
             if not created:
                 await sync_to_async(content_app_status.save_heartbeat)()
 
+                if content_app_status.versions != versions:
+                    content_app_status.versions = versions
+                    await sync_to_async(content_app_status.save)(update_fields=["versions"])
```
bmbouter, line 56: Wouldn't this case mean the component versions were changed while Pulp is running? Maybe that's good to handle, but let me first see if I'm understanding this right.

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -149,6 +123,35 @@ def _signal_handler(self, thesignal, frame):
 
         self.shutdown_requested = True
 
+    def handle_worker_heartbeat(self):
+        """
+        This is a function for updating worker heartbeat records.
```
bmbouter, line 128:
```suggestion
        Create or update worker heartbeat records.
```

pulpcore/app/models/task.py
```diff
@@ -90,6 +77,7 @@ class Worker(BaseModel):
 
     name = models.TextField(db_index=True, unique=True)
     last_heartbeat = models.DateTimeField(auto_now=True)
+    versions = HStoreField(null=True)
```
bmbouter, line 80: Shouldn't these always have versions with a default of `{"versions": {}}` based on [this line](https://github.com/pulp/pulpcore/pull/3466/files#diff-7a529f651b0c47d7f7d8b44adaafc9a5fadaa03aa1e455fec23b8ee1c68274e9R49).

pulpcore/content/__init__.py
mdellweg: This can also mean that a new process was started with different versions using the same worker name.

pulpcore/app/models/task.py
mdellweg: Maybe it should default to `{}`. So probably:
```suggestion
    versions = HStoreField(default=dict)
```

dralley (approved):

CHANGES/3365.misc
```diff
@@ -0,0 +1 @@
+Added version information to the worker records in the database.
```
bmbouter, line 1: I think this would be better as a plugin writer feature. It would be ok to merge as is so I lgtm'd it.

mdellweg: I made it misc, because yes, a "normal" user would not be concerned. But would a plugin writer be? Thinking of it, maybe in writing migrations... I'll change it to plugin/.feature.

PR #3469, opened 2022-12-14
Expose get_url in the plugin api

daviddavis:


daviddavis:
This is for https://github.com/pulp/pulp_deb/pull/689. We're trying to return the signing service ids (which are nested in a dict field) as hrefs instead of uuids.

gerrod3 (approved):

PR #3470, opened 2022-12-15
Remove plugin dependencies from labels migration

mdellweg:
Pulpcore migrations can never depend on any migration from a plugin. It will lead to either circular dependencies or inconsistent history if a plugin is installed later.
To make the miration as smooth as possible, a post migrate hook was added to repair the labels as well as possible.

[noissue]

### DIFF
--- a/pulpcore/app/apps.py
+++ b/pulpcore/app/apps.py
@@ -8,6 +8,7 @@ from django.conf import settings
 from django import apps
 from django.core.exceptions import ImproperlyConfigured
 from django.db import transaction
+from django.db.models.expressions import OuterRef, RawSQL
 from django.db.models.signals import post_migrate
 from django.utils.module_loading import module_has_submodule
 
@@ -223,6 +224,11 @@ class PulpAppConfig(PulpPluginAppConfig):
             sender=self,
             dispatch_uid="populate_artifact_serving_distribution_identifier",
         )
+        post_migrate.connect(
+            _migrate_remaining_labels,
+            sender=self,
+            dispatch_uid="migrate_remaining_labels_identifier",
+        )
 
 
 def _populate_access_policies(sender, apps, verbosity, **kwargs):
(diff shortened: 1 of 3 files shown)


pulpcore/app/migrations/0098_pulp_labels.py
```diff
@@ -3,26 +3,8 @@
...
-    "gem",
-    "maven",
-    "npm",
-    "ostree",
-    "python",
-    "rpm",
-]
 
 
 def copy_labels_up(apps, schema_editor):
```
gerrod3, line 10: What is the point of still running this during the migration if the post-migrate will also handle it?

pulpcore/app/apps.py
```diff
@@ -361,3 +367,47 @@ def _populate_artifact_serving_distribution(sender, apps, verbosity, **kwargs):
...
+            )
+            model.objects.update(pulp_labels=label_subq)
+            Label.objects.filter(content_type=ctype).delete()
+
+    if Label.objects.count():
+        if verbosity >= 1:
+            print(
+                "Some labels could not be migrated. "
+                "Please try to run `pulpcore-manager datarepair-labels`."
+            )
```
gerrod3, line 413: Will this actually do anything if they run it again? I know they can force delete the labels, but won't the migration end in failure again?

pulpcore/app/apps.py
```diff
@@ -361,3 +367,47 @@ def _populate_artifact_serving_distribution(sender, apps, verbosity, **kwargs):
...
                 )
+
+
+def _migrate_remaining_labels(sender, apps, verbosity, **kwargs):
+    try:
+        Label = apps.get_model("core", "label")
+        ContentType = apps.get_model("contenttypes", "ContentType")
+    except LookupError:
+        if verbosity >= 1:
+            print("No labels available.")
```
gerrod3, line 378:
```suggestion
            print("No labels left to migrate.")
```

pulpcore/app/apps.py
```diff
@@ -361,3 +367,47 @@ def _populate_artifact_serving_distribution(sender, apps, verbosity, **kwargs):
...
+        if not hasattr(model, "pulp_labels"):
+            if verbosity >= 1:
+                print(
+                    _(
+                        "Warning! "
+                        "Labels for content_type {app_label}.{model} could not be migrated. "
+                        "Model has no labels."
+                    ).format(app_label=ctype.app_label, model=ctype.model)
+                )
+            break
```
gerrod3, line 397:
```suggestion
            continue
```
I think this should be a `continue`, else you will never migrate the other model's pulp_labels even if you could. Also, this probably has to be changed in the other 2 spots. (Copying and pasting code :cry: )

decko (approved):

gerrod3 (changes requested):

pulpcore/app/apps.py
mdellweg: This is more like the db table was not available.

pulpcore/app/apps.py
mdellweg: Oh, right!

pulpcore/app/migrations/0098_pulp_labels.py
mdellweg: Good question. This is the hell confusing. I can remove running this code here.

pulpcore/app/apps.py
mdellweg: It will not, until the remaining labels are backed by new code. I do not expect anyone to run into this ever, but as all that label stuff is in the plugin api, we can never know.

mdellweg: I would want to keep it here to make the admin aware of that command.

pulpcore/app/apps.py
```diff
@@ -361,3 +367,47 @@ def _populate_artifact_serving_distribution(sender, apps, verbosity, **kwargs):
...
+                print(
+                    _(
+                        "Warning! "
+                        "Labels for content_type {app_label}.{model} could not be migrated. "
+                        "Model has no labels."
+                    ).format(app_label=ctype.app_label, model=ctype.model)
+                )
+            continue
+
+        with transaction.atomic():
```
gerrod3, line 399: There should be a verbosity print statement here if the migration will be done like in the command.

mdellweg: added.

gerrod3 (approved):

PR #3479, opened 2022-12-20
Fixed some error-log paths that potentially expose data.

### LINKED ISSUE #3446: Better exception handling
I believe it is a good practice to not send the exception/stack trace itself to the final user when dealing with errors.
Currently we're doing it in two places:
1. https://github.com/pulp/pulpcore/blob/2c573c4d03514e743b57357165d001c43acd90c2/pulpcore/app/viewsets/user.py#L221-L221

and 

2. https://github.com/pulp/pulpcore/blob/2c573c4d03514e743b57357165d001c43acd90c2/pulpcore/app/serializers/importer.py#L83-L83

Also, the second place deals directly using the `Exception` object, where it could use a more specific exception.

ggainey:
Import-path will be tested in pulp_file.

fixes #3446

### DIFF
--- a/pulpcore/app/models/importer.py
+++ b/pulpcore/app/models/importer.py
@@ -1,3 +1,4 @@
+from django.core.exceptions import ObjectDoesNotExist
 from django.db import models
 
 from pulpcore.app.models import (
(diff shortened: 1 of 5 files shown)


pulpcore/app/models/importer.py
```diff
@@ -51,10 +52,18 @@ def repo_mapping(self):
...
         self.repo_map.all().delete()
+        failed_repos = []
         for source, repo_name in mapping.items():
-            repo = Repository.objects.get(name=repo_name)
+            try:
+                repo = Repository.objects.get(name=repo_name)
+            except ObjectDoesNotExist:
+                failed_repos.append(repo_name)
+                continue
             self.repo_map.create(source_repo=source, repository=repo)
```
lubosmj, line 62: Does it make sense to create repository mappings even though we raise `ObjectDoesNotExist` in the end?

lubosmj (approved):

pulpcore/app/serializers/importer.py
```diff
@@ -78,9 +79,11 @@ def create(self, validated_data):
         importer = super().create(validated_data)
         try:
             importer.repo_mapping = repo_mapping
-        except Exception as err:
+        except ObjectDoesNotExist as err:
             importer.delete()
-            raise serializers.ValidationError(_("Bad repo mapping: {}").format(err))
+            raise serializers.ValidationError(
+                _("Failed to find repositories from repo_mapping: {}").format(err)
+            )
```
gerrod3, line 86: Is this going to remove the warning since we are still putting an exception into the error message?

ggainey: The str-representation of an exception isn't a stacktrace, it's just the error "message". What we've cleaned up here is an overly-broad "except Exception", and better messaging. CodeQL's alert on "might be a stacktrace here" is a false positive - we can dismiss it.

gerrod3: Yeah, I was wondering if there is a way to remove the CodeQL warning with this change.

ggainey: You can "dismiss" CodeQL warnings, with a reason and a comment. I *believe* it's smart enough to not re-raise for the same codepath after that - we'll see, I guess!

gerrod3 (approved):

PR #3498, opened 2023-01-10
Use pytest fixtures instead of the pulp-smash config

lubosmj:
Ref: https://github.com/pulp/pulp_installer/pull/1450
Ref: https://github.com/pulp/oci_env/pull/82

[noissue]

### DIFF
--- a/.github/workflows/scripts/install.sh
+++ b/.github/workflows/scripts/install.sh
@@ -82,6 +82,7 @@ if [[ " ${SCENARIOS[*]} " =~ " ${TEST} " ]]; then
   - name: pulp-fixtures\
     image: docker.io/pulp/pulp-fixtures:latest\
     env: {BASE_URL: "http://pulp-fixtures:8080"}' vars/main.yaml
+  export REMOTE_FIXTURES_ORIGIN="http://pulp-fixtures:8080"
 fi
 if [ "$TEST" == 'stream' ]; then
   sed -i -e '/^services:/a \
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -12,6 +12,7 @@ import proxy
 import pytest
 
 from aiohttp import web
+from dataclasses import dataclass
 from time import sleep
 from yarl import URL
 
@@ -113,6 +114,18 @@ def pytest_configure(config):
         "markers",
         "from_pulpcore_for_all_plugins: marks tests from pulpcore as beneficial for plugins to run",
     )
+
+
+@pytest.fixture(scope="session")
+def fixtures_cfg():
+    @dataclass
+    class FixturesConfig:
+        aiohttp_fixtures_origin: str = "127.0.0.1"
+        remote_fixtures_origin: str = os.environ.get(
+            "REMOTE_FIXTURES_ORIGIN", "https://fixtures.pulpproject.org/"
+        )
+
+    return FixturesConfig()
 
 
 # API Clients
@@ -391,11 +404,11 @@ def unused_port():
 
 
 @pytest.fixture
-def gen_threaded_aiohttp_server(pulp_cfg, unused_port):
+def gen_threaded_aiohttp_server(fixtures_cfg, unused_port):
     fixture_servers_data = []
 
     def _gen_threaded_aiohttp_server(app, ssl_ctx, call_record):
-        host = pulp_cfg.aiohttp_fixtures_origin
+        host = fixtures_cfg.aiohttp_fixtures_origin
         port = unused_port()
         shutdown_event = threading.Event()
         fixture_server = ThreadedAiohttpServer(shutdown_event, app, host, port, ssl_ctx)
@@ -435,8 +448,8 @@ def gen_fixture_server(gen_threaded_aiohttp_server):
 
 
 @pytest.fixture
-def http_proxy(pulp_cfg, unused_port):
-    host = pulp_cfg.aiohttp_fixtures_origin
+def http_proxy(fixtures_cfg, unused_port):
+    host = fixtures_cfg.aiohttp_fixtures_origin
     port = unused_port()
     proxypy_args = [
         "--num-workers",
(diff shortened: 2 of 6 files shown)


pulpcore/tests/functional/__init__.py
```diff
@@ -106,6 +160,11 @@ def pytest_configure(config):
 # API Clients
 
 
+@pytest.fixture(scope="session")
+def pulp_cfg(bindings_cfg):
```
mdellweg, line 164:
```suggestion
def pulp_cfg():
```
Also i thought we could just use `django.config.settings` and it would do the dynaconf loading. Can you elaborate why this may not be possible?

Is that not what we need? https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/__init__.py#L473

lubosmj: Ah 😮. I have already forgotten this fixture. I can simplify some workflows (e.g., in pulp_container).

The idea was to migrate the `pulp_cfg` fixture to pulpcore. Only `settings.CONTENT_ORIGIN`  can be used as a replacement for `pulp_cfg.get_base_url()`.

For settings like `aiohttp_fixtures_origin` or `fixtures_url`, we still need to access the configuration file defined initially for pulp_smash. In oci_env, there is `fixtures_url="http://pulp-fixtures:8080/`, but in a vagrant box, there is `fixtures_url=https://fixtures.pulpproject.org/`. I think we should still read these two settings from the configuration file since not all of us run fixtures locally. What do you think? Should I rather open PRs against the oci_env and pulp_installer repositories where I will set environment variables for these settings and then hard-code default values for them as we did in `get_bindings_config`?

mdellweg: Oh, OK, I was mislead by seeing dynaconf. But this one is reading the smash config. Is fixtures_url really the only one thing we need from there? Then yes, i'd prefer not adding a fixture we will have a hard time removing over the course of time.

lubosmj: We deal with two settings here:
- aiohttp_fixtures_origin -> that is the origin for a server used by pulp_file (e.g., the `gen_bad_response_fixture_server` fixture); this can be hard-coded because it looks like we always set it to "127.0.0.1".
- fixtures_url -> is the origin for the fixtures we pull content from; could be set to a remote or local URL; this should be read from the actual settings from now

I will try to project this to the PR once again.

mdellweg: We said, our tests always run on localhost (one of the make-it-simple design goals). So let's use this fact and set `aiohttp_fixtures_origin` hardcoded (should we remove it; turn it into a constant?).

For `fixtures_url` you suggested reading it from Environment variables with a decent default. I would like to do that. Also keep in mind that another goal is to not use the `fixtures_url` at all eventually (this should allow for _some_ amount of duct-tape here). Can you add comments for this plan so we don't forget?

mdellweg: An afterthought: To me "getting away from pulp-smash" also includes "stopping to use smash config".

lubosmj: Okay. It makes sense. I just misunderstood your choice of words. I am going to declare a new environment variable for `fixtures_url` (in oci_env and pulp_installer as well).

lubosmj:
There is already `export PULP_FIXTURES_URL="http://pulp-fixtures:8080"` called before executing pulp-cli tests. Do you think I should replace my customized env variable (REMOTE_FIXTURES_ORIGIN) with this? Should be these two things used separately?

.github/workflows/scripts/install.sh
```diff
@@ -82,6 +82,7 @@ if [[ " ${SCENARIOS[*]} " =~ " ${TEST} " ]]; then
   - name: pulp-fixtures\
     image: docker.io/pulp/pulp-fixtures:latest\
     env: {BASE_URL: "http://pulp-fixtures:8080"}' vars/main.yaml
+  export REMOTE_FIXTURES_ORIGIN="http://pulp-fixtures:8080"
```
mdellweg, line 85: This should come via a template change...
Also, do you see any chance to make this the uneeded my by making its value the default?

lubosmj: By default, we do not run fixtures locally (which might be problematic for running tests locally). So, I would rather move the change to the plugin template.

lubosmj: I opened a PR in the plugin template repository (https://github.com/pulp/plugin_template/pull/725). I am not removing this change in pulpcore in order to let the tests pass with a non-default `REMOTE_FIXTURES_ORIGIN` value.

dralley (approved):

pulpcore/tests/functional/__init__.py
```diff
@@ -115,6 +116,18 @@ def pytest_configure(config):
     )
 
 
+@pytest.fixture(scope="session")
+def fixtures_cfg():
+    @dataclass
```
mdellweg, line 121: Out of curiosity: A named tuple would do the same, right?

pulpcore/tests/functional/api/test_openapi_schema.py
```diff
@@ -31,8 +31,8 @@ def openapi3_schema_with_modified_safe_chars(openapi3_schema_spec):
 
 
 @pytest.fixture(scope="session")
-def pulp_openapi_schema_url(pulp_cfg, pulp_api_v3_path):
-    return f"{pulp_cfg.get_base_url()}{pulp_api_v3_path}docs/api.json"
+def pulp_openapi_schema_url(bindings_cfg, pulp_api_v3_path):
+    return f"{bindings_cfg.host}{pulp_api_v3_path}docs/api.json"
```
mdellweg, line 35:
```suggestion
    return f"{pulp_api_v3_url}docs/api.json"
```

mdellweg (approved):
Left one possible "optimization".

lubosmj:
The referenced PRs are not needed to be merged first. The one in oci_env will contain a profile for running fixtures locally (i.e., mimicking `pfixtures` in a vagrant box). The one in pulp_installer is there just to set an environment variable in a development environment. But, we are already moving away from the installer.

mdellweg (approved):

PR #3500, opened 2023-01-11
Fix the label migration code

### LINKED ISSUE #3495: django.db.utils.IntegrityError when upgrading to pulpcore 3.22
**Version**
pulpcore 3.22.0

**Describe the bug**
When upgrading to pulpcore 3.22, the pulp_labels migration fails with:

```
django.db.utils.IntegrityError: null value in column "pulp_labels" of relation "core_repository" violates not-null constraint
DETAIL:  Failing row contains (bbb38cf3-c632-4d51-a91f-94b23c7d3544, 2023-01-10 14:06:50.628502+00, 2023-01-10 14:06:50.632373+00, test, null, 1, file.file, null, null, f, null).
```

**To Reproduce**
1. In pulpcore < 3.22, create 2 repos: one with labels, one without.
2. Upgrade to pulpcore 3.22

**Expected behavior**
No migration failure.

mdellweg:
A not null constraint exception was risen when objects with and without labels of the same type existed. The fix will limit the update to only those objects that actually had labels.

fixes #3495

### DIFF
--- a/CHANGES/3495.bugfix
+++ b/CHANGES/3495.bugfix
@@ -1,0 +1,2 @@
+Fixed the label migration code for not null constraint errors when objects with and without labels
+are migrated.
--- a/pulpcore/app/apps.py
+++ b/pulpcore/app/apps.py
@@ -408,7 +408,9 @@ def _migrate_remaining_labels(sender, apps, verbosity, **kwargs):
                 .annotate(label_data=RawSQL("hstore(array_agg(key), array_agg(value))", []))
                 .values("label_data")
             )
-            model.objects.update(pulp_labels=label_subq)
+            model.objects.annotate(old_labels=label_subq).exclude(old_labels={}).update(
+                pulp_labels=label_subq
+            )
             Label.objects.filter(content_type=ctype).delete()
 
     if Label.objects.count():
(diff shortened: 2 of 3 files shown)


CHANGES/3495.bugfix
```diff
@@ -0,0 +1,2 @@
+Fixed the label migration code for not null constraint errors when objects with an without labels
```
lubosmj, line 1:
```suggestion
Fixed the label migration code for not null constraint errors when objects with and without labels
```

pulpcore/app/apps.py
```diff
@@ -408,7 +408,9 @@ def _migrate_remaining_labels(sender, apps, verbosity, **kwargs):
                 .annotate(label_data=RawSQL("hstore(array_agg(key), array_agg(value))", []))
                 .values("label_data")
             )
-            model.objects.update(pulp_labels=label_subq)
+            model.objects.annotate(old_labels=label_subq).exclude(old_labels={}).update(
```
ggainey, line 411: To make sure I understand what's going on here - we're annotating the models with "old_labels", so we can skip the ones where old_labels is empty, so that the update() doesn't store 'null' into those objects?

mdellweg: Yes, exactly.

ggainey (approved):

gerrod3 (approved):

PR #3510, opened 2023-01-13
Don't save temporary files to object storage, use local storage

dralley:
The general issue is that Pulp uses temporary files, storage backends like S3 or Azure require you to pay for usage, so using those services for short lived and often small files is not an efficient use of resources.

[noissue]


lubosmj:
`default_storage` is equal to [`DEFAULT_FILE_STORAGE`](https://docs.djangoproject.com/en/3.2/topics/files/#file-storage) which can be also s3 or azure. Are you sure that the title of the commit reflects the change?

PR #3511, opened 2023-01-13
Enables FilesystemExporter to incrementally export rpms

parthaa:
It accepts a start repository version and calculates what content artifacts need to get copied over.

### DIFF
--- a/pulpcore/app/serializers/exporter.py
+++ b/pulpcore/app/serializers/exporter.py
@@ -294,6 +294,12 @@ class FilesystemExportSerializer(ExportSerializer):
         write_only=True,
     )
 
+    start_repository_version = RepositoryVersionRelatedField(
+        help_text=_("The URI of the last-exported-repo-version."),
+        required=False,
+        write_only=True,
+    )
+
     def validate(self, data):
         if ("publication" not in data and "repository_version" not in data) or (
             "publication" in data and "repository_version" in data
@@ -305,7 +311,11 @@ class FilesystemExportSerializer(ExportSerializer):
 
     class Meta:
         model = models.FilesystemExport
-        fields = ExportSerializer.Meta.fields + ("publication", "repository_version")
+        fields = ExportSerializer.Meta.fields + (
+            "publication",
+            "repository_version",
+            "start_repository_version",
+        )
 
 
 class FilesystemExporterSerializer(ExporterSerializer):
--- a/pulpcore/app/tasks/export.py
+++ b/pulpcore/app/tasks/export.py
@@ -105,7 +105,7 @@ def _export_to_file_system(path, relative_paths_to_artifacts, method=FS_EXPORT_M
 
 
 def _export_publication_to_file_system(
-    path, publication, method=FS_EXPORT_METHODS.WRITE, allow_missing=False
+    path, publication, start_repo_version=None, method=FS_EXPORT_METHODS.WRITE, allow_missing=False
 ):
     """
     Export a publication to the file system.
@@ -114,9 +114,19 @@ def _export_publication_to_file_system(
         path (str): Path to place the exported data
         publication_pk (str): Publication pk
     """
+    difference_content_artifacts = []
     content_artifacts = ContentArtifact.objects.filter(
         pk__in=publication.published_artifact.values_list("content_artifact__pk", flat=True)
     )
+    if start_repo_version:
+        start_version_content_artifacts = ContentArtifact.objects.filter(
+            artifact__in=start_repo_version.artifacts
+        )
+        difference_content_artifacts = set(
+            content_artifacts.difference(start_version_content_artifacts).values_list(
+                "pk", flat=True
+            )
+        )
 
     if publication.pass_through:
         content_artifacts |= ContentArtifact.objects.filter(
@@ -138,8 +148,11 @@ def _export_publication_to_file_system(
         "content_artifact", "content_artifact__artifact"
     ).iterator():
         # Artifact isn't guaranteed to be present
-        if pa.content_artifact.artifact:
+        if pa.content_artifact.artifact and (
+            start_repo_version is None or pa.content_artifact.pk in difference_content_artifacts
+        ):
             relative_path_to_artifacts[pa.relative_path] = pa.content_artifact.artifact
+
     _export_to_file_system(path, relative_path_to_artifacts, method)
 
 
@@ -160,7 +173,7 @@ def _export_location_is_clean(path):
     return True
 
 
-def fs_publication_export(exporter_pk, publication_pk):
+def fs_publication_export(exporter_pk, publication_pk, start_repo_version_pk=None):
     """
     Export a publication to the file system using an exporter.
 
@@ -170,25 +183,41 @@ def fs_publication_export(exporter_pk, publication_pk):
     """
     exporter = Exporter.objects.get(pk=exporter_pk).cast()
     publication = Publication.objects.get(pk=publication_pk).cast()
+
+    start_repo_version = None
+    if start_repo_version_pk:
+        start_repo_version = RepositoryVersion.objects.get(pk=start_repo_version_pk)
+
+    params = {"publication": publication_pk}
+    if start_repo_version:
+        params["start_repository_version"] = start_repo_version_pk
+
     export = FilesystemExport.objects.create(
         exporter=exporter,
-        params={"publication": publication_pk},
+        params=params,
         task=Task.current(),
     )
     ExportedResource.objects.create(export=export, content_object=publication)
     CreatedResource.objects.create(content_object=export)
-
     log.info(
-        "Exporting: file_system_exporter={exporter}, publication={publication}, path={path}".format(
-            exporter=exporter.name, publication=publication.pk, path=exporter.path
+        "Exporting: file_system_exporter={exporter}, publication={publication}, "
+        "start_repo_version={start_repo_version}, path={path}".format(
+            exporter=exporter.name,
+            publication=publication.pk,
+            start_repo_version=start_repo_version_pk,
+            path=exporter.path,
         )
     )
+
     if not _export_location_is_clean(exporter.path):
         raise RuntimeError(_("Cannot export to directories that contain existing data."))
-    _export_publication_to_file_system(exporter.path, publication, exporter.method)
+
+    _export_publication_to_file_system(
+        exporter.path, publication, start_repo_version=start_repo_version, method=exporter.method
+    )
 
 
-def fs_repo_version_export(exporter_pk, repo_version_pk):
+def fs_repo_version_export(exporter_pk, repo_version_pk, start_repo_version_pk=None):
     """
     Export a repository version to the file system using an exporter.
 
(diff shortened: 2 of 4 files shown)


ggainey:
This PR currently supports `start_repository_version` **only if** you have specified `repository_version`. `publication=` is a major workflow for users of this functionality, but if you specify `start_repository_version=` and `publication=`, you get a 500. Rude. :)

I think we should always accept `publication=`. Actually, more strongly - I contend that we should support a `start_publication=` param as well. You can only specify one of publication/repository_version, and only one of start_publication/start_repository_version, and the combinatorics should all work.

(Also - we prob need some error-checking at the view-level, you should **not** be able to specify a delta that "goes back in time" - once we have start and end repo-version, we should insure that start < end...)

The other thing I'm noting is, if I start with a repo, remove some RPMs to create version-N, add them back in to create version-M, and export start_repository_version=N/repository_version=M, I get an empty export. I expect somewhere in the "find the content for a version" is making An Assumption that may not be valid.

Testing/investigation continues.

parthaa:
> This PR currently supports `start_repository_version` **only if** you have specified `repository_version`. `publication=` is a major workflow for users of this functionality, but if you specify `start_repository_version=` and `publication=`, you get a 500. Rude. :)

Interesting -  Katello only uses publication. Wonder why  `start_repository_version=` and `publication=` is failing for you. I used that combination for all my tests :). 

> I think we should always accept `publication=`. Actually, more strongly - I contend that we should support a `start_publication=` param as well. You can only specify one of publication/repository_version, and only one of start_publication/start_repository_version, and the combinatorics should all work.

I wonder if we should limit the scope of this PR to only publication and start_publication and leave the version compare to a different PR? Katello really only needs this combination to work. That would potentially mean small set of changes.


> The other thing I'm noting is, if I start with a repo, remove some RPMs to create version-N, add them back in to create version-M, and export start_repository_version=N/repository_version=M, I get an empty export. I expect somewhere in the "find the content for a version" is making An Assumption that may not be valid

I will look into this case

sjha4:
We may not create "new" publications if our contents_changed dynflow middleware detects nothing has changed in which case we reuse the previous version's publication. Every repo version however will have a publication URL which could be used here I think with the catch that the publication may not be unique between versions.

ianballou:
Here's the publication logic: https://github.com/Katello/katello/blob/69d4da4767ea71810a4323c02ff0ce3969418289/app/lib/actions/pulp3/orchestration/repository/generate_metadata.rb#L15

parthaa:
> This PR currently supports `start_repository_version` **only if** you have specified `repository_version`. `publication=` is a major workflow for users of this functionality, but if you specify `start_repository_version=` and `publication=`, you get a 500. Rude. :)

```bash
$ curl --cert /home/vagrant/foreman-certs/client_cert.pem --key /home/vagrant/foreman-certs/client_key.pem \
   -X POST -H 'Content-Type: application/json' \
   -d '{"publication":"/pulp/api/v3/publications/rpm/rpm/62b3af46-d059-4d2b-a898-1dd7955ec531/", "start_repository_version":"/pulp/api/v3/repositories/rpm/rpm/6c98c027-53ff-440e-a5ce-ba68cc0833cc/versions/4/"}' \
    https://centos8-katello-devel-stable.example.com/pulp/api/v3/exporters/core/filesystem/8f7a1139-5153-4e54-9e9d-994152379d01/exports/
```

Works well - `journalctl` output

```
Jan 20 20:06:28 centos8-katello-devel-stable.example.com pulpcore-api[15405]: pulp [0ed95cc8205048eca7cf18c134bf3b94]:  - - [20/Jan/2023:20:06:28 +0000] "POST /pulp/api/v3/exporters/core/filesystem/8f7a1139-5153-4e54-9e9d-994152379d01/exports/ HTTP/1.1" 202 67 "-" "curl/7.61.1"
Jan 20 20:06:28 centos8-katello-devel-stable.example.com pulpcore-worker-4[23566]: pulp [0ed95cc8205048eca7cf18c134bf3b94]: pulpcore.tasking.pulpcore_worker:INFO: Starting task 7b7d754c-771b-4c51-a443-dd5402be20da
Jan 20 20:06:28 centos8-katello-devel-stable.example.com pulpcore-worker-4[23566]: pulp [0ed95cc8205048eca7cf18c134bf3b94]: pulpcore.app.tasks.export:INFO: Exporting: file_system_exporter=open_Export-Library-SYNCABLE_2.0-67, publication=62b3af46-d059-4d2b-a898-1dd7955ec531, start_repo_version=e5e32d67-0e8c-455a-9c65-eb9cb5242ac7, path=/var/lib/pulp/exports/open/Export-Library-SYNCABLE/2.0/2023-01-20T18-27-51-00-00/custom/goals/auto
Jan 20 20:06:28 centos8-katello-devel-stable.example.com pulpcore-worker-4[23566]: pulp [0ed95cc8205048eca7cf18c134bf3b94]: pulpcore.tasking.pulpcore_worker:INFO: Task completed 7b7d754c-771b-4c51-a443-dd5402be20da
Jan 20 20:07:10 centos8-katello-devel-stable.example.com pulpcore-api[15405]: pulp [6d519aaaad01486bbfdda21d9685c6b5]:  - - [20/Jan/2023:20:07:10 +0000] "GET /pulp/api/v3/tasks/7b7d754c-771b-4c51-a443-dd5402be20da/ HTTP/1.1" 200 744 "-" "Pulp-CLI/0.14.0"

```

ianballou:
> I wonder if we should limit the scope of this PR to only publication and start_publication and leave the version compare to a different PR? Katello really only needs this combination to work. That would potentially mean small set of changes.

I'm still digging into this myself, but it sounds like Katello only really needs to give the start repo version and the end publication to get what we want -- if that's the case then I'm with Partha in that perhaps the start publication could be added later. I'd be curious to hear though if Pulp folks think we might be missing something and do need a start publication.

mdellweg:
Can we phrase this feature around "content" instead of "rpms"?

parthaa:
> Can we phrase this feature around "content" instead of "rpms"?

I think only Yum/File repos are supported for FS Exporter at this point. Where do you see this being specific to rpms?

mdellweg:
> > Can we phrase this feature around "content" instead of "rpms"?
> 
> I think only Yum/File repos are supported for FS Exporter at this point. That being said where do you see this PR being specific to rpms?

I meant the docs and changelog. Sorry for not being specific enough.

pulpcore/app/viewsets/exporter.py
```diff
@@ -159,21 +159,34 @@ def create(self, request, exporter_pk):
         serializer = FilesystemExportSerializer(data=request.data, context={"exporter": exporter})
         serializer.is_valid(raise_exception=True)
 
+        start_repository_version_pk = None
+        if request.data.get("start_repository_version"):
+            start_repository_version_pk = self.get_resource(
+                request.data["repository_version"], RepositoryVersion
```
ggainey, line 165: Has to be

```suggestion
                request.data["start_repository_version"], RepositoryVersion
```

CHANGES/3413.feature
```diff
@@ -0,0 +1,4 @@
+Added a start_repository_version parameter to the fs_exporter.
+
+Also made it incrementally export rpms that differed between
```
ggainey, line 3:
```suggestion
If specified, it will export only rpms that differed between 
```

pulpcore/app/serializers/exporter.py
```diff
@@ -294,6 +294,12 @@ class FilesystemExportSerializer(ExportSerializer):
         write_only=True,
     )
 
+    start_repository_version = RepositoryVersionRelatedField(
+        help_text=_("A URI of the last-exported-repo-version."),
```
ggainey, line 298:
```suggestion
        help_text=_("The URI of the last-exported-repo-version."),
```

ggainey (changes requested):
Some minor wordsmithing, and one bugfix :)

CHANGES/3413.feature
```diff
@@ -0,0 +1,4 @@
+Added a start_repository_version parameter to the fs_exporter.
+
+If specified, it will export only rpms that differed between 
+2 repository versions
```
mdellweg, line 4:
```suggestion
Added a ``start_repository_version`` parameter to the file system exporter.

If specified, it will export only content units that differed between two repository versions.
```

mdellweg (commented):
Some formatting and what i meant by agnostic in the changelog...

dralley:
@parthaa Can you rebase and squash commits?

pulpcore/app/tasks/export.py
```diff
@@ -138,8 +144,11 @@ def _export_publication_to_file_system(
         "content_artifact", "content_artifact__artifact"
     ).iterator():
         # Artifact isn't guaranteed to be present
-        if pa.content_artifact.artifact:
+        if pa.content_artifact.artifact and (
+            start_repo_version is None or pa.content_artifact in difference_content_artifacts
```
dralley, line 148: `difference_content_artifacts` appears to be a queryset, do we know that this isn't going to be re-evaluated every time in a loop?

parthaa: Thats a good point I am curious about that :). How would I be able to check if its getting re evaluated ?

dralley: You could explicitly cast to a set

parthaa: updated!

ggainey (approved):
Looks like all current comments are responded to. Tested with an RPM repo in my test env, worked like a champ. Looks good to me!

chris1984:
Tested it here locally with Katello and was able to export incrementally and import without any issues

pulpcore/app/tasks/export.py
```diff
@@ -138,8 +148,11 @@ def _export_publication_to_file_system(
         "content_artifact", "content_artifact__artifact"
     ).iterator():
         # Artifact isn't guaranteed to be present
-        if pa.content_artifact.artifact:
+        if pa.content_artifact.artifact and (
+            start_repo_version is None or pa.content_artifact.pk in difference_content_artifacts
+        ):
```
dralley, line 153: How does this work with the metadata e.g. RPM metadata?  

`difference_content_artifacts` is derived from the repository version, not the publication, so how does this not filter out all the RPM metadata from the newest publication?

ggainey: It exports the full metadata, always, but only the delta for artifacts.

parthaa: Oh it always picks the latest metadata from the repo publication (I thought that was the same way regular incremental export worked? ). It is only doing a diff for rpms

dralley: @ggainey I agree that's what it should do, but where is the separate codepath for metadata?  It looks like it's all treated as `Content`, and `PublishedMetadata` inherits from `Content` so that can work, but unlike any other kind of content it's not part of any repositories.

ggainey: We're looking at the diff between a start_repo-version and a publication - "only export things that are in the publication, and are NOT in the start_repo_version". So the publication's metadata is always "found".

@parthaa does that match your expectation?

dralley: Ok, never mind, that sounds correct.   Line 119 starts from the publication.

I think my brain flipped around the order of the sets.

dralley (approved):

pulpcore/app/tasks/export.py
ggainey: > Ok, never mind, that sounds correct. Line 119 starts from the publication.
> 
> I think my brain flipped around the order of the sets.

No worries - the flow is complicated, deltas make it moreso, and responding to your question made me go over the whole thing again so I actually understand it better now myself :)

ggainey:
@parthaa OK, please rebase this on top of current-main to address the lint-issues, and resubmit.

parthaa:
@dralley  - I'll request it if needed and if customers are desperate for it.

dralley:
@parthaa It looks like there is a request for this to go into 3.18 (6.12.z), could you investigate which other patches are necessary to backport it?

PR #3517, opened 2023-01-19
Document compatible PostgreSQL version

### LINKED ISSUE #3505: [Task] Document that Pulp is incompatible with db-connection pools
PgBouncer was identified to break Pulps expectations with advisory locks. We should document something to point people not reading their docs to ours.

Details: Pulp uses session based Postgres features, so connection pooling on the transaction level will fail. Since at least PgBouncer does not inspect the statements sent to the server for incompatible features, this will lead to bad issues including data corruption.

https://www.pgbouncer.org/features.html

mdellweg:
Also warn about the incompatibility with db connection pooling.

fixes #3505

### DIFF
--- a/docs/installation/instructions.rst
+++ b/docs/installation/instructions.rst
@@ -128,7 +128,7 @@ Database Setup
 --------------
 
 You must provide a PostgreSQL database for Pulp to use. At this time, Pulp 3.0 will only work with
-PostgreSQL.
+PostgreSQL .
 
 PostgreSQL
 ^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


ipanova (approved):

docs/installation/instructions.rst
```diff
@@ -128,21 +128,30 @@ Database Setup
...
-PostgreSQL.
+PostgreSQL .
 
 PostgreSQL
 ^^^^^^^^^^
 
 Installation package considerations
 ***********************************
 
+Pulp needs a version of PostgreSQL providing session based advisory locks and listen-notify. Any
```
ipanova, line 139: why we are at it, does it makes sense to somehow capture this? https://github.com/pulp/pulp_installer/pull/1458/files#diff-b4d68dc855d0f9476d3f2ee343853bd21bf82ea9960d0cf06661baa244439dd6R83

mdellweg: absolutely.

ggainey (approved):

PR #3521, opened 2023-01-23
Refactor linked serializer fields

mdellweg:
This allows the serializers to be used without explicitely specifying an empty or dummy request in the context.

[noissue]

### DIFF
--- a/pulpcore/app/serializers/base.py
+++ b/pulpcore/app/serializers/base.py
@@ -26,6 +26,188 @@ from pulpcore.app.util import (
 log = getLogger(__name__)
 
 
+# Field mixins
+
+
+class HrefFieldMixin:
+    """A mixin to configure related fields to generate relative hrefs."""
+
+    @property
+    def context(self):
+        # Removes the request from the context to display relative hrefs.
+        res = dict(super().context)
+        res["request"] = None
+        return res
+
+
+class _MatchingRegexViewName(object):
+    """This is a helper class to help defining object matching rules for master-detail.
+
+    If you can be specific, please specify the `view_name`, but if you cannot, this allows
+    you to specify a regular expression like .e.g. `r"repositories(-.*/.*)?-detail"` to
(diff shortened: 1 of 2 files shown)


ggainey:
Am I understanding the original code correctly, in that it looks like the "ignore the passed in view name" comment is true only for the _DetailsFieldMixin? And in the other overrides, we're just removing the "request" parameter? 

So with these changes, we now remove "request" in the context() override. I assume that the code-path leading to get_url() relies on context()? Just trying to wrap my brain around the sequence we're touching here.

mdellweg:
> Am I understanding the original code correctly, in that it looks like the "ignore the passed in view name" comment is true only for the _DetailsFieldMixin? And in the other overrides, we're just removing the "request" parameter?
> 
> So with these changes, we now remove "request" in the context() override. I assume that the code-path leading to get_url() relies on context()? Just trying to wrap my brain around the sequence we're touching here.

Yes, absolutely. The old comments were mixed up.

pulpcore/app/serializers/base.py
```diff
@@ -268,23 +268,31 @@ def _view_name(self, obj):
             raise ValueError(msg)
         return get_view_name_for_model(obj, "detail")
 
+    @property
+    def context(self):
```
ggainey, line 272: OK, the refactoring looks good. Question - would it be possible/reasonable to do this as a mixin, so that we could have content() in one place?

pulpcore/app/serializers/base.py
```diff
@@ -26,6 +26,188 @@
 log = getLogger(__name__)
 
 
+# Field mixins
+
+
+class HrefFieldMixin:
```
ggainey, line 32: :100:

ipanova:
@mdellweg would you mind opening an issue so it contains a changelog?

mdellweg:
@ipanova please look again.

ipanova (approved):
Thanks!

PR #3523, opened 2023-01-24
Pass correlation id to signing script through ENV variable

### LINKED ISSUE #3522: Correlation ID should be available to signing cli
The Correlation ID / guid is used throughout Pulp to correlate logs with a particular request, however it is not made available to the signing cli script. It should be, in case that script wants to log things too.

sdherr:
closes #3522

### DIFF
--- a/CHANGES/3522.feature
+++ b/CHANGES/3522.feature
@@ -1,0 +1,1 @@
+Pass correlation id to signing script through ENV variable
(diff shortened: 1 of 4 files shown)


mdellweg (commented):
This looks sound to me. Can i assume that you tested this with a plugin that actually uses signing?

sdherr:
@mdellweg Yes, tested as working with both `pulp_rpm` and `pulp_deb`, using `sign` and `asign`.

mdellweg (approved):

Issue #3507, opened 2023-01-12
Add filters for "retain_repo_versions" field


daviddavis:
We'd like a way to filter repos by the `retain_repo_versions` field. Filters such as eq, gt/gte, lt/lte, and is null/is not null.

PR #3527, opened 2023-01-25
Add retain_repo_versions filter

gerrod3:
fixes: #3507

### DIFF
--- a/pulpcore/app/viewsets/repository.py
+++ b/pulpcore/app/viewsets/repository.py
@@ -27,7 +27,11 @@ from pulpcore.app.viewsets import (
     AsyncUpdateMixin,
     NamedModelViewSet,
 )
-from pulpcore.app.viewsets.base import DATETIME_FILTER_OPTIONS, NAME_FILTER_OPTIONS
+from pulpcore.app.viewsets.base import (
+    DATETIME_FILTER_OPTIONS,
+    NAME_FILTER_OPTIONS,
+    NULLABLE_NUMERIC_FILTER_OPTIONS,
+)
 from pulpcore.app.viewsets.custom_filters import LabelFilter
 from pulpcore.tasking.tasks import dispatch
 from pulpcore.filters import HyperlinkRelatedFilter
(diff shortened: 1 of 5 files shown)


pulpcore/app/viewsets/repository.py
```diff
@@ -39,7 +39,10 @@ class RepositoryFilter(BaseFilterSet):
 
     class Meta:
         model = Repository
-        fields = {"name": NAME_FILTER_OPTIONS}
+        fields = {
+            "name": NAME_FILTER_OPTIONS,
+            "retain_repo_versions": ["exact", "isnull"] + DATETIME_FILTER_OPTIONS,
```
mdellweg, line 44: Hmm, is what is `DATETIME_FILTER_OPTION`? `retain_repo_version` is surely not a `datetime`. So should it be called `NUMERICAL_COMPARE_FILTER_OPTIONS`?

gerrod3: `DATETIME_FILTER_OPTIONS = ["lt", "lte", "gt", "gte", "range"]`, the filterset auto-builds the filters based on the operations and the type of field it is.

mdellweg: I guess this doing exactly what we want. My point is i would not dare to try using something called "for datetime" when the objects i filter for are integers. So i think we should rename that constant if we can. Is it in the plugin api?

gerrod3: Yes it is in the plugin api. Should I just write out the whole list instead of using a constant?

mdellweg: So we'd need to duplicate the constant to give it a proper name...
I am going to approve and leave it to your best judjment.

mdellweg (approved):

pulpcore/app/viewsets/repository.py
gerrod3: I created a new constant `NULLABLE_NUMERIC_FILTER_OPTIONS` and added to the plugin API.

ipanova (approved):

PR #3537, opened 2023-01-30
Pick the first occurence of RA in the ACSHandleStage.

ipanova:
[noissue]

If there are ACS that point to the same content, pick the first RA instead of last

Issue TBD


bmbouter (approved):
This makes sense to me. There is no point in searching for a second checksum entry when one has already been found and added to `existing_ras_dict`.

mdellweg (approved):

mdellweg (approved):

PR #3540, opened 2023-01-31
Repeat to kill the process of a canceled task

### LINKED ISSUE #3407: Task cancellation gets stuck
**Version**
Pulpcore 3.16 via Satellite 6.11

**Describe the bug**
If "Reclaim Space" (Actions::Pulp3::CapsuleContent::ReclaimSpace) action is executed and canceled after that task will get stuck in canceling not only in foreman tasks but also in pulp3. 

The task was apparently stuck in "canceling" state for more than 6 days.

**To Reproduce**
How reproducible:
Always

Steps to Reproduce:
1. Execute "Reclaim Space" on Satellite
2. Cancel task (Actions::Pulp3::CapsuleContent::ReclaimSpace)

**Expected results:**
Task will get canceled

**Actual results:**
Task is stuck in foreman-tasks and in pulp3

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=2143290

mdellweg:
fixes #3407

### DIFF
--- a/pulpcore/app/tasks/test.py
+++ b/pulpcore/app/tasks/test.py
@@ -1,3 +1,14 @@
+import backoff
+
+
 def dummy_task():
     """Dummy task, that can be used in tests."""
     pass
+
+
+@backoff.on_exception(backoff.expo, BaseException)
+def gooey_task(interval):
+    """A sleep task that tries to avoid being killed by ignoring all exceptions."""
+    from time import sleep
+
+    sleep(interval)
--- a/pulpcore/tasking/pulpcore_worker.py
+++ b/pulpcore/tasking/pulpcore_worker.py
@@ -54,7 +54,11 @@ from pulpcore.tasking.util import _delete_incomplete_resources  # noqa: E402
 _logger = logging.getLogger(__name__)
 random.seed()
 
+# Number of heartbeats for a task to finish on graceful worker shutdown (approx)
 TASK_GRACE_INTERVAL = 3
+# Number of heartbeats between attempts to kill the subprocess (approx)
+TASK_KILL_INTERVAL = 1
+# Number of heartbeats between cleaning up worker processes (approx)
 WORKER_CLEANUP_INTERVAL = 100
 # Randomly chosen
 TASK_SCHEDULING_LOCK = 42
@@ -121,6 +125,7 @@ class NewPulpWorker:
 
         _logger.info(_("Worker %s was requested to shut down."), self.name)
 
+        self.task_grace_timeout = TASK_GRACE_INTERVAL
         self.shutdown_requested = True
 
     def handle_worker_heartbeat(self):
@@ -166,7 +171,7 @@ class NewPulpWorker:
     def beat(self):
         if self.worker.last_heartbeat < timezone.now() - timedelta(seconds=self.heartbeat_period):
             self.worker = self.handle_worker_heartbeat()
-            if self.shutdown_requested:
+            if self.task_grace_timeout > 0:
                 self.task_grace_timeout -= 1
             self.worker_cleanup_countdown -= 1
             if self.worker_cleanup_countdown <= 0:
@@ -303,7 +308,6 @@ class NewPulpWorker:
 
         This function must only be called while holding the lock for that task."""
 
-        self.task_grace_timeout = TASK_GRACE_INTERVAL
         task.worker = self.worker
         task.save(update_fields=["worker"])
         cancel_state = None
@@ -317,11 +321,15 @@ class NewPulpWorker:
                     item = connection.connection.notifies.pop(0)
                     if item.channel == "pulp_worker_cancel" and item.payload == str(task.pk):
                         _logger.info(_("Received signal to cancel current task %s."), task.pk)
-                        os.kill(task_process.pid, signal.SIGUSR1)
                         cancel_state = TASK_STATES.CANCELED
                     # ignore all other notifications
                 if cancel_state:
-                    break
+                    if self.task_grace_timeout > 0:
+                        _logger.info("Wait for canceled task to abort.")
+                    else:
+                        self.task_grace_timeout = TASK_KILL_INTERVAL
+                        _logger.info("Aborting current task %s due to cancelation.", task.pk)
+                        os.kill(task_process.pid, signal.SIGUSR1)
 
                 r, w, x = select.select(
                     [self.sentinel, connection.connection, task_process.sentinel],
@@ -344,10 +352,8 @@ class NewPulpWorker:
                         )
                     else:
                         _logger.info("Aborting current task %s due to worker shutdown.", task.pk)
-                        os.kill(task_process.pid, signal.SIGUSR1)
                         cancel_state = TASK_STATES.FAILED
                         cancel_reason = "Aborted during worker shutdown."
-                        break
             task_process.join()
             if not cancel_state and task_process.exitcode != 0:
                 _logger.warning(
--- a/CHANGES/3407.bugfix
+++ b/CHANGES/3407.bugfix
@@ -1,0 +1,2 @@
+Added repeating logic to signalling a task worker subprocess. This should fix a bug where the
+task refuses to be terminated easily.
--- a/pulpcore/tests/functional/api/test_tasking.py
+++ b/pulpcore/tests/functional/api/test_tasking.py
@@ -105,7 +105,7 @@ def test_delete_cancel_running_task(dispatch_task, tasks_api_client):
 
     for i in range(10):
         task = tasks_api_client.read(task_href)
-        if task.state != "running":
+        if task.state == "running":
             break
         time.sleep(1)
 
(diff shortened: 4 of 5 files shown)


pulpcore/tasking/pulpcore_worker.py
```diff
@@ -317,11 +321,15 @@ def supervise_task(self, task):
                     item = connection.connection.notifies.pop(0)
                     if item.channel == "pulp_worker_cancel" and item.payload == str(task.pk):
                         _logger.info(_("Received signal to cancel current task %s."), task.pk)
-                        os.kill(task_process.pid, signal.SIGUSR1)
                         cancel_state = TASK_STATES.CANCELED
                     # ignore all other notifications
-                if cancel_state:
-                    break
+                if cancel_state == TASK_STATES.CANCELED:
+                    if self.task_grace_timeout:
```
ggainey, line 327: If task_grace_timeout ever goes negative (somehow), this ends up being "true" when we prob don't want it to - safer to make this check <0, like the check at line 349

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -317,11 +321,15 @@ def supervise_task(self, task):
...
+                        _logger.info("Wait for canceled task to abort.")
+                    else:
+                        self.task_grace_timeout = TASK_KILL_INTERVAL
+                        _logger.info("Aborting current task %s due to cancelation.", task.pk)
+                        os.kill(task_process.pid, signal.SIGUSR1)
```
ggainey, line 332: Are we sure that sending USR1 again is guaranteed to kill the task-process? Do we need a "no really, shoot it in the head" w/ SIGKILL if it's somehow still around?

ggainey (changes requested):
A change and a question - def on the right track tho

dralley:
Do we have a theory on "why" this happens?  I don't object to the approach necessarily

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -54,7 +54,11 @@
 _logger = logging.getLogger(__name__)
 random.seed()
 
+# Number of heartbeats for a task to finish on graceful worker shutdown (approx)
 TASK_GRACE_INTERVAL = 3
+# Number of heartbeats between attemts to kill the subprocess (approx)
```
dralley, line 59: typo "attemts"

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -317,11 +321,15 @@ def supervise_task(self, task):
                     item = connection.connection.notifies.pop(0)
                     if item.channel == "pulp_worker_cancel" and item.payload == str(task.pk):
                         _logger.info(_("Received signal to cancel current task %s."), task.pk)
-                        os.kill(task_process.pid, signal.SIGUSR1)
                         cancel_state = TASK_STATES.CANCELED
                     # ignore all other notifications
-                if cancel_state:
-                    break
+                if cancel_state == TASK_STATES.CANCELED:
```
dralley, line 326: Do we expect that it could possibly be in a different state here or are we just being explicit?  Curious about this change.

pulpcore/app/tasks/test.py
```diff
@@ -1,3 +1,14 @@
...
+
+
 def dummy_task():
     """Dummy task, that can be used in tests."""
     pass
+
+
+@backoff.on_exception(backoff.expo, BaseException)
+def gooey_task(interval):
+    """A sleep task that is really reluctant to being killed."""
```
dralley, line 11: reluctant -> resistant?  Also it would be good to expand a bit on the mechanics of it

pulpcore/tasking/pulpcore_worker.py
mdellweg: If it can ever go negative, it can also go infinitely negative and produce an overflow. I think i should prevent it from ever getting there in the first place. I will adjust the ticking down part.

pulpcore/tasking/pulpcore_worker.py
mdellweg: The test in place is showing that it is sufficient, as long as the subprocess is "only" ignoring python exceptions. I would keep it that way until we hear from the field that it's not sufficient.
I remember vaguely that i had troubles sending SIGTERM where the worker (for whatever reason) terminated itself.

pulpcore/tasking/pulpcore_worker.py
mdellweg: This is a consequence of the worker shutdown code path being able to signal multiple times too. It would set the `cancel_state` to `FAILED` instead.

pulpcore/app/tasks/test.py
mdellweg:
```suggestion
    """A sleep task that tries to avoid being killed by ignoring all exceptions."""
```

pulpcore/tasking/pulpcore_worker.py
mdellweg:
```suggestion
# Number of heartbeats between attempts to kill the subprocess (approx)
```

mdellweg:
> Do we have a theory on "why" this happens? I don't object to the approach necessarily

We never got the requested log steam around the failed cancel, so no. When i see the cli test in our CI failing to properly cancel a task, there seems to be a backoff involved. But i could never reproduce it.

pulpcore/tasking/pulpcore_worker.py
mdellweg: Actually i reworked the killing process in the worker shutdown to take this route too. Looks like this now:

```
pulp [70ba56f0-3532-4965-8ca7-c3697ca6b90c]: pulpcore.tasking.pulpcore_worker:INFO: Starting task d1d42cf9-90c1-4416-9a3c-a647eeb0dca2
^Cpulp [None]: pulpcore.tasking.pulpcore_worker:INFO: Worker 51070@8bd82e0b7d01 was requested to shut down.
pulp [None]: pulpcore.tasking.pulpcore_worker:INFO: Worker shutdown requested, waiting for task d1d42cf9-90c1-4416-9a3c-a647eeb0dca2 to finish.
pulp [None]: pulpcore.tasking.pulpcore_worker:INFO: Worker shutdown requested, waiting for task d1d42cf9-90c1-4416-9a3c-a647eeb0dca2 to finish.
pulp [None]: pulpcore.tasking.pulpcore_worker:INFO: Aborting current task d1d42cf9-90c1-4416-9a3c-a647eeb0dca2 due to worker shutdown.
pulp [None]: pulpcore.tasking.pulpcore_worker:INFO: Aborting current task d1d42cf9-90c1-4416-9a3c-a647eeb0dca2 due to cancelation.
pulp [None]: pulpcore.tasking.pulpcore_worker:INFO: Cleaning up task d1d42cf9-90c1-4416-9a3c-a647eeb0dca2 and marking as failed. Reason: Aborted during worker shutdown.
pulp [None]: pulpcore.tasking.pulpcore_worker:INFO: Worker 51070@8bd82e0b7d01 was shut down.
```

ggainey:
> Do we have a theory on "why" this happens? I don't object to the approach necessarily

I spent a fair amount of time trying to reproduce the reported behavior in a variety of ways with no success. My current hypothesis is that, somehow, the USR1 handling threw an exception before reaching the os.kill() stmt - but I couldn't convince myself I could see how that might happen. Without logs, we're just guessing.

pulpcore/tasking/pulpcore_worker.py
dralley: FYI, there is effectively no such thing as an integer overflow in Python, when it overflows 64 bits the payload type gets promoted to an implementation of "big integers"

dralley:
Do we have sufficient logging to ensure that we would be able to make use of the log stream if we had it?

mdellweg:
We log when the worker (outerprocess) received the cancel notification. And we log when the worker is killing the child process. So it should at least be sufficient to see, where in the process we get lost.
And that is the very line I desparately want to see in the logs. Also for the case this fix is tailored for you would see some logging from backoff.

dralley (approved):

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -166,7 +171,7 @@ def worker_cleanup(self):
     def beat(self):
         if self.worker.last_heartbeat < timezone.now() - timedelta(seconds=self.heartbeat_period):
             self.worker = self.handle_worker_heartbeat()
-            if self.shutdown_requested:
+            if self.task_grace_timeout > 0:
```
ggainey, line 174: :+1:

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -317,11 +321,15 @@ def supervise_task(self, task):
                     item = connection.connection.notifies.pop(0)
                     if item.channel == "pulp_worker_cancel" and item.payload == str(task.pk):
                         _logger.info(_("Received signal to cancel current task %s."), task.pk)
-                        os.kill(task_process.pid, signal.SIGUSR1)
                         cancel_state = TASK_STATES.CANCELED
                     # ignore all other notifications
                 if cancel_state:
-                    break
+                    if self.task_grace_timeout > 0:
```
ggainey, line 327: :+1:

ggainey (approved):

PR #3559, opened 2023-02-09
Addressed another deadlock-window in overlapping-content-syncs path.

### LINKED ISSUE #3284: contentartifact bulk_update can still deadlock

ggainey:
In this episode, we teach content_stages to not do updates that aren't changing anything. This removes an entire class of potential-deadlocks from the overlapping-content path - we don't deadlock, because we have stopped acquiring locks for rows we don't actually need to change.

fixes #3284.
[nocoverage]

(applied from commit bf89001cc5f0c365336386e98cc252ad3d1f85a1)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,11 +25,11 @@ jobs:
       run: git branch main origin/main
     - name: Commit Count Check
       run: test `git log  --oneline --no-merges HEAD ^main  | wc -l ` = 1
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
   lint:
-    runs-on: ubuntu-latest
-    
+    runs-on: ubuntu-20.04
+
 
     steps:
       - uses: actions/checkout@v3
@@ -76,7 +76,7 @@ jobs:
       - name: Check manifest
         run: check-manifest
 
-      
+
 
       - name: Check for gettext problems
         run: sh .ci/scripts/check_gettext.sh
@@ -85,7 +85,7 @@ jobs:
         run: python .ci/scripts/upper_bound.py
 
   test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     # run only after lint finishes
     needs: lint
     strategy:
@@ -122,7 +122,7 @@ jobs:
           echo "TEST=${{ matrix.env.TEST }}" >> $GITHUB_ENV
 
       - name: Before Install
-        
+
         run: .github/workflows/scripts/before_install.sh
         shell: bash
         env:
@@ -136,7 +136,7 @@ jobs:
           GITHUB_CONTEXT: ${{ github.event.pull_request.commits_url }}
 
       - name: Install
-        
+
         run: .github/workflows/scripts/install.sh
         shell: bash
         env:
@@ -150,12 +150,12 @@ jobs:
           GITHUB_CONTEXT: ${{ github.event.pull_request.commits_url }}
 
       - name: Install Python client
-        
+
         run: .github/workflows/scripts/install_python_client.sh
         shell: bash
 
       - name: Before Script
-        
+
         run: .github/workflows/scripts/before_script.sh
         shell: bash
         env:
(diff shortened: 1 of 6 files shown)


dralley (approved):

dralley:
@ggainey Wrong version of pulp_file is being installed

ggainey:
> @ggainey Wrong version of pulp_file is being installed

pulp_file is fine, but pulp-cli was pulling the main branch. Rather than dig too deeply, I pinned this branch at cli/0.11 - we're interested only in getting this set of fixes released.

PR #3565, opened 2023-02-13
Add an aiohttp replacement for the http_get function

lubosmj:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -899,7 +899,7 @@ def add_to_filesystem_cleanup():
                 pass
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
 def download_content_unit(bindings_cfg):
     def _download_content_unit(base_path, content_path):
         async def _get_response(url):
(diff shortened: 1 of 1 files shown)


pulpcore/tests/functional/__init__.py
```diff
@@ -917,3 +917,17 @@ async def _get_response(url):
         return asyncio.run(_get_response(url))
 
     return _download_content_unit
+
+
+@pytest.fixture(scope="class")
```
mdellweg, line 922: Why scoping it to class?

lubosmj: We do not care about the moment when the fixture is destroyed, do we?

lubosmj: We do not, but I am removing the scope because the fixture might be impossible to use in combination with other scope fixtures.

mdellweg: Then we should scope it "session", which i prefer anyway.

mdellweg (approved):

ggainey (approved):

Issue #3566, opened 2023-02-13
Add ability to replicate distributions/repositories from another Pulp


dkliban:
Users should be able to configure Pulp with API credentials for another Pulp instance. Users should then be able make a single API call to cause Pulp to discover all distributions on the upstream Pulp and create all the necessary remotes, repositories, and distributions to serve the exact same content. Users should also be able to specify a label that should be used to match a subset of distributions on the upstream Pulp.

PR #3568, opened 2023-02-13
Enables replicating distributions/repositories from another Pulp.

dkliban:
This patch adds a new Upstream Pulps API. This API enables the user to provide REST API credentials for an upstream Pulp. The user is then able to make a POST request to the Upstream Pulp 'replicate' API to dispatch a task that will look for all distributions that are available on the upstream Pulp. This task will then create remotes that can sync from those distributions. It will also create repositories and distributions that have the same base paths as on the upstream Pulp. It will then sync from all the distributions. The task then removes any distributions that are no longer present on the upstream Pulp, but exist on the Pulp running the replication task. The user is able to limit which distributions are replicated by specifying a 'labels' attribute on the Upstream Pulp config.

This patch provides an abstract class of a Replicator that each plugin is expected to implement. The replication task will only replicate content for plugins that implement a Replicator.

closes: #3566


pulpcore/app/models/replica.py
```diff
@@ -0,0 +1,26 @@
...
+
+    ca_cert = models.TextField(null=True)
+    client_cert = models.TextField(null=True)
+    client_key = EncryptedTextField(null=True)
+    tls_validation = models.BooleanField(default=True)
+
+    username = EncryptedTextField(null=True)
+    password = EncryptedTextField(null=True)
+
+    labels = models.TextField(null=True)
```
mdellweg, line 26: This is meant to encode criteria for replicating, right?
Do we want to invent a DDL that can filter by labels, name patterns, relpath prefixes?

labels are key[-value] in Pulp. How do you expect this to be represented?

If this was just for a list of label keys, should we not make it an ArrayField?

pulpcore/app/replica.py
```diff
@@ -0,0 +1,136 @@
...
+        self.task_group = task_group
+        self.distribution_ctx = None
+        self.remote_ctx = None
+        self.repository_ctx = None
+        self.publication_ctx = None
+        self.remote_model = None
+        self.repository_model = None
+        self.distribution_model = None
+        self.distribution_serializer = None
+        self.app_label = None
```
mdellweg, line 22: At least these could be class attributes.

pulpcore/app/replica.py
```diff
@@ -0,0 +1,136 @@
...
+            params = {}
+        offset = 0
+        list_size = 100
+        while True:
+            distributions = self.distribution_ctx(self.pulp_ctx).list(list_size, offset, params)
+            for distro in distributions:
+                yield distro
+            if len(distributions) < list_size:
+                break
+            offset += list_size
```
mdellweg, line 38: :thinking: This iterator could be a _thing_ on the context model.

pulpcore/app/replica.py
```diff
@@ -0,0 +1,136 @@
...
+
+class Replicator:
+    def __init__(self, pulp_ctx, task_group):
+        """
+        Override this method with instances of plugin specific instances of Remote, Repository, and Distribution contexts.
+        :param pulp_ctx: PulpReplicaContext
+        """
+        self.pulp_ctx = pulp_ctx
+        self.task_group = task_group
+        self.distribution_ctx = None
```
mdellweg, line 14: If is understand correctly, you intend to put the context class here. In pulp-cli land this variable is always used as an instance. This would confuse me.

pulpcore/app/tasks/replica.py
```diff
@@ -0,0 +1,58 @@
...
+from pulpcore.app.apps import pulp_plugin_configs
+
+
+def replicate_distributions(server_pk):
+    server = UpstreamPulp.objects.get(pk=server_pk)
+    api_kwargs = dict(
+        base_url=server.base_url,
+        username=server.username,
+        password=server.password,
+        user_agent="pulpcore-3.22",
```
mdellweg, line 13: Can we find a way to let this not be outdated easily?

pulpcore/app/viewsets/replica.py
```diff
@@ -0,0 +1,53 @@
...
+        """
+        server = UpstreamPulp.objects.get(pk=pk)
+        task_group = TaskGroup.objects.create(description=f"Replication of {server.name}")
+
+        task = dispatch(
+            replicate_distributions,
+            exclusive_resources=["/pulp/api/v3/servers/"],
+            kwargs={"server_pk": pk},
+            task_group=task_group,
+        )
```
mdellweg, line 51: This is exclusive in the way that you can only replicate one upstream at a time?

pulpcore/app/tasks/replica.py
dkliban: Yes, this was just a placeholder. I'll get the version string from __version__ of pulpcore.

pulpcore/app/viewsets/replica.py
dkliban: yes. I am open to discussion about this.

pulpcore/app/models/replica.py
dkliban: Here is the help text I provided on the serializer: "One or more pulp_labels that will be used to filter distributions on the upstream Pulp. E.g.  '"foo=bar,key=val" or "foo,key"'

pulpcore/app/replica.py
dkliban: Yes, these are classes, not instances. I'll update the names to better reflect that.

pulpcore/app/replica.py
dkliban: I would very much like it if it was.

pulpcore/app/models/replica.py
mdellweg: "foo=bar,bar" is also possible?

pulpcore/app/viewsets/replica.py
mdellweg: Well what would it even mean if you replicate multiple pulps? Would the last one remove all the former ones? What if the collide on a distribution name?

pulpcore/app/models/replica.py
dkliban: Yes, it is possible. However, 'bar' only matches keys 'bar'.

dkliban:
@mdellweg  What do you think about changing the 'labels' attribute of UpstreamPulp to 'pulp_label_select' to match the query parameter name from the REST API for distributions?

pulpcore/app/viewsets/replica.py
dkliban: yeah ... Multiple Pulp's should not be replicated onto the same Pulp. Only if you decide that you want to replace the repositories with this other Pulp.

pulpcore/app/replica.py
dkliban: I updated these to be class attributes. Thanks for the suggestion.

mdellweg:
> @mdellweg What do you think about changing the 'labels' attribute of UpstreamPulp to 'pulp_label_select' to match the query parameter name from the REST API for distributions?

If it follows the very same syntax, that sounds like a fine idea.

pulpcore/app/models/__init__.py
```diff
@@ -95,3 +95,6 @@
 
 # Moved here to avoid a circular import with Task
 from .progress import GroupProgressReport, ProgressReport  # noqa
+
+# Moved here to avoid a circular import with
```
ggainey, line 99: "...avoid circular import with"...what?

pulpcore/app/models/replica.py
```diff
@@ -0,0 +1,26 @@
...
+.. _Plugin Writer's Guide:
+    https://docs.pulpproject.org/pulpcore/plugins/plugin-writer/index.html
+"""
+
+from django.db import models
+from pulpcore.plugin.models import BaseModel, EncryptedTextField
+
+
+class UpstreamPulp(BaseModel):
+    name = models.TextField(db_index=True, unique=True)
```
ggainey, line 13: Do we need to handle the proxy usecases here as well?

ggainey (changes requested):

pulpcore/app/migrations/0100_upstreampulp.py
```diff
@@ -0,0 +1,38 @@
...
+            name='UpstreamPulp',
+            fields=[
+                ('pulp_id', models.UUIDField(default=uuid.uuid4, editable=False, primary_key=True, serialize=False)),
+                ('pulp_created', models.DateTimeField(auto_now_add=True)),
+                ('pulp_last_updated', models.DateTimeField(auto_now=True, null=True)),
+                ('name', models.TextField(db_index=True, unique=True)),
+                ('base_url', models.TextField()),
+                ('api_root', models.TextField(default='pulp')),
+                ('ca_cert', models.TextField(null=True)),
+                ('client_cert', models.TextField(null=True)),
```
mdellweg, line 26: This should be encrypted.

requirements.txt
```diff
@@ -20,6 +20,7 @@ dynaconf>=3.1.9,<=3.1.11
 gunicorn>=20.1,<=20.1.0
 jinja2>=3.1,<=3.1.2
 naya>=1.1.1,<=1.1.1
+pulp-glue>=0.17.1,<1.0
```
mdellweg, line 23: There is no such guarantee.
```suggestion
pulp-glue>=0.17.1,<0.18
```

pulpcore/app/tasks/replica.py
```diff
@@ -0,0 +1,75 @@
...
+        base_url=server.base_url,
+        username=server.username,
+        password=server.password,
+        user_agent=user_agent(),
+    )
+    ctx = ReplicaContext(
+        api_root=server.api_root,
+        api_kwargs=api_kwargs,
+        format="json",
+        background_tasks=False,
```
mdellweg, line 35: This looks like a task is going to wait on another task. We cannot safely do this. This is the on think that can make our tasking system deadlock.

mdellweg (changes requested):

pulpcore/app/migrations/0100_upstreampulp.py
mdellweg: I see the key is protected. Maybe we should add support for API tokens...

pulpcore/app/tasks/replica.py
mdellweg: Is this a misconception on my side? We only use glue to question the upstream server, right? I still would feel better if we made these calls non-blocking (well they still wait on an http response, but at least not on pulp tasks.).

pulpcore/app/tasks/replica.py
```diff
@@ -0,0 +1,75 @@
...
+import sys
+
+from pulpcore.app.apps import pulp_plugin_configs
+from pulpcore.app.models import UpstreamPulp, TaskGroup
+from pulpcore.app.replica import ReplicaContext
+
+from pulp_glue.common import __version__ as pulp_glue_version
+
+
+def user_agent():
```
mdellweg, line 12: Where are you using this?

pulpcore/app/tasks/replica.py
```diff
@@ -0,0 +1,75 @@
...
+                try:
+                    local_distro = replicator.distribution_model.objects.get(name=distro["name"])
+                    local_distro.repository = None
+                    local_distro.publication = None
+                    local_distro.save()
+                    continue
+                except replicator.distribution_model.DoesNotExist:
+                    continue
+            # Check if there is already a repository
+            repository = replicator.create_or_update_repository(remote=remote)
```
mdellweg, line 64: It feels like i'm missing something. Is this a blocking (waiting on another pulp task) call?

pulpcore/app/replica.py
```diff
@@ -0,0 +1,151 @@
...
+        return {}
+
+    def create_or_update_repository(self, remote):
+        try:
+            repository = self.repository_model_cls.objects.get(name=remote.name)
+            # Update the existing repository with latest values
+            repository.remote = remote
+            for field_name, value in self.repository_extra_fields(remote).items():
+                setattr(repository, field_name, value)
+            repository.save()
```
mdellweg, line 85: We cannot do that if we do not have the resource reserved.

pulpcore/app/tasks/replica.py
```diff
@@ -0,0 +1,75 @@
...
+
+            # Dispatch a sync task
+            replicator.sync(repository)
+
+            # Get or create a distribution
+            replicator.create_or_update_distribution(repository, distro)
```
mdellweg, line 70: This one seems fine. We dispatch a task, but we don't wait on it.

pulpcore/app/tasks/replica.py
```diff
@@ -0,0 +1,75 @@
...
+                    local_distro.publication = None
+                    local_distro.save()
+                    continue
+                except replicator.distribution_model.DoesNotExist:
+                    continue
+            # Check if there is already a repository
+            repository = replicator.create_or_update_repository(remote=remote)
+
+            # Dispatch a sync task
+            replicator.sync(repository)
```
mdellweg, line 67: This also seems fine.

pulpcore/app/tasks/replica.py
```diff
@@ -0,0 +1,75 @@
...
+
+            # Get or create a distribution
+            replicator.create_or_update_distribution(repository, distro)
+
+            # Add name to the list of known distribution names
+            distro_names.append(distro["name"])
+
+        replicator.remove_missing(distro_names)
```
mdellweg, line 75: For this we miss all the resource locks again.

mdellweg (changes requested):

pulpcore/app/models/replica.py
dkliban: We do, but I don't think pulp-glue currently supports that. @mdellweg please correct me if I am wrong.

pulpcore/app/tasks/replica.py
dkliban: We use pulp-glue to look up distributions and for some plugins we look up repositories and/or publications. We don't perform any tasks using pulp-glue.

pulpcore/app/tasks/replica.py
dkliban: No, this is not blocking. THe current implementation only dispatches tasks for updating/creating distributions.

pulpcore/app/models/replica.py
mdellweg: We inherit the proxysupport from requests. So i guess by manipulating the environment of the process we could inject proxy information. But that is not a solution beyond a cli too.
So we should add passing proxy information to requests there.

pulpcore/app/tasks/replica.py
mdellweg: I know. So we don't _ever_ make a call that will even check for this condition. Still for the sake of future us, please make it `True`, so we see immediate failures when someone starts to introduce dangerous calls.
Remember, you could configure this to replicate from itself (maybe a different domain). Having this task waiting on a pulp_task in the same box is a recipe for deadlocks.

pulpcore/app/tasks/replica.py
mdellweg: But we may redo this to dispatch delete tasks.

pulpcore/app/tasks/replica.py
mdellweg: Yes so the issue is that we must not update an existing repository in this task, yet we cannot wait for a dispatched task to do it for us.

pulpcore/app/replica.py
dkliban: Yeah ... I'll refactor this  to dispatch a task.

pulpcore/app/tasks/replica.py
dkliban: I will update it to True.

pulpcore/app/tasks/replica.py
dkliban: I am going to dispatch a task for updating a repository. We don't have to wait on it because the task will have an exclusive lock on the repository which will prevent the sync task from running first.

dkliban:
@mdellweg I added tasks for remote and repository updates. I also now dispatch tasks to delete distributions.

pulpcore/app/tasks/replica.py
mdellweg: Can you check that the sync task will not have the previous remote taken at "POST" time from the repository?

pulpcore/app/replica.py
```diff
@@ -0,0 +1,206 @@
...
+        needs_update = False
+        for field_name, value in fields_dict.items():
+            if isinstance(getattr(model_instance, field_name), Model):
+                if get_url(getattr(model_instance, field_name)) != value:
+                    needs_update = True
+            elif getattr(model_instance, field_name) != value:
+                needs_update = True
+        return needs_update
+
+    def get_upstream_distributions(self, labels=None):
```
mdellweg, line 53:
```suggestion
    def upstream_distributions(self, labels=None):
```
As being an iterator this can be considered to _be_ the upstream distributions.

pulpcore/app/replica.py
```diff
@@ -0,0 +1,206 @@
...
+    def sync_params(self, repository, remote):
+        """This method returns a dict that will be passed as kwargs to the sync task."""
+        raise NotImplementedError("Each replicator must supply its own sync params.")
+
+    def sync(self, repository, remote):
+        dispatch(
+            self.sync_task,
+            task_group=self.task_group,
+            shared_resources=[remote],
+            exclusive_resources=[repository],
```
mdellweg, line 175: And now i know, (again) why we cannot rely on the remote attribute of the repository in the task. It would not do the locking properly.

mdellweg (approved):

pulpcore/app/models/__init__.py
dkliban: GroupProgressReport ... i updated the comment.

pulpcore/app/tasks/replica.py
dkliban: I updated it to dispatch general_delete tasks.

pulpcore/app/models/replica.py
ggainey: @mdellweg @dkliban so if I'm understanding this - if the downstream-pulp-machine is set up to use a proxy in the general case (e.g. via /etc/environments or whatever) then that gets used when talking to the upstream-pulp (and is managed externally). These validation-methods are to be used by downstream **once it's contacted upstream**, correct?

If that's so, then I'm good. Being able to "override" proxy-settings per-call becomes a nice-to-have, instead of a can't-get-there-from-here.

ggainey (commented):
I think I'm good here (modulo the discussions ongoing RE async/locking)

ggainey (approved):
Looks good to me!

pulpcore/app/models/replica.py
mdellweg: That is my assumption. But also it may not be the same syntax / semantic as glue is using requests, while remotes e.g. use aiohttp. We may still need to add better proxy support to glue. Proxy usage should be configurable per profile in the cli.

pulpcore/app/replica.py
```diff
@@ -0,0 +1,206 @@
...
+            for distro in distributions:
+                yield distro
+            if len(distributions) < list_size:
+                break
+            offset += list_size
+
+    def url(self, upstream_distribution):
+        return upstream_distribution["base_url"]
+
+    def remote_extra_fields(self, upstream_distribution):
```
ipanova, line 71: @dkliban can you add same for distributions?

dkliban: yes, let's do that.

Issue #3569, opened 2023-02-14
The search in the pulpcore documentation is broken


quba42:
**Version**
Affects the released docs: https://docs.pulpproject.org/pulpcore/

**Describe the bug**
Searching using the search just has a spinning wheel forever.

errors in webconsole:
```
Uncaught ReferenceError: jQuery is not defined
Uncaught ReferenceError: $ is not defined
```

**Additional context**
Looks like this affects the pulp_rpm docs as well. pulp_deb docs were still unaffected last I checked but I am not sure how recently we released a new version.

dkliban:
Looks like we are affected by https://github.com/readthedocs/sphinx_rtd_theme/issues/1434

dkliban:
I am going to try installing sphinxcontrib-jquery==3.0.0 locally to see if it fixes the issue.

dkliban:
That did not fix the problem. Installing sphinx-rtd-theme==1.1.1 did fix the problem. I am going to pin the sphinx-rtd-them to < 1.2.0

benjaoming:
Hi @dkliban - a new sphinxcontrib-jquery has been released (4.1), so you should be able to use the latest auto-resolved sphinxcontrib-jquery that comes with `sphinx-rtd-theme==1.2.0`.

gherceg:
Hi @benjaoming, I tried upgrading `sphinxcontrib-jquery` to v4.1 with `sphinx-rtd-theme` pinned to v1.2 and `sphinx` pinned to v6.1.3, but still experienced the `ReferenceError: jQuery is not defined` when using the search bar. Let me know if I can provide any more helpful information.

ipanova:
@benjaoming seems like neither pinning spnhinxcontrib-query and sphinx-rtd-theme helps.

benjaoming:
Edit: That's what I think you should do to get things working with the latest version. But pinning things is a fine approach, there's no harm in being a few minor versions behind :+1:

lubosmj:
@benjaoming, thanks, the changes related to the config file helped to resolve the issue. 🥳 

https://github.com/pulp/pulpcore/pull/3668

PR #3571, opened 2023-02-14
Fixes docs search by pinning sphinx-rtd-theme

dkliban:
fixes: #3569

### DIFF
--- a/doc_requirements.txt
+++ b/doc_requirements.txt
@@ -1,7 +1,7 @@
 -r requirements.txt
 plantuml
 sphinx
-sphinx-rtd-theme
+sphinx-rtd-theme<1.2.0
 sphinxcontrib-openapi
 towncrier
 mistune<2.0.0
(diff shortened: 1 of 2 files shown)


decko (approved):

quba42:
Should this also be added to pulp_rpm? Other plugins? Is there a mechanism for handling this consistently?

doc_requirements.txt
```diff
@@ -1,7 +1,7 @@
 -r requirements.txt
 plantuml
 sphinx
-sphinx-rtd-theme
+sphinx-rtd-theme<1.2.0
```
mdellweg, line 4: Dependabot will probably try updating this right away. We need a comment here to prevent us from accepting that blindly.

ipanova:
@quba42 i started to submit PRs to the other repos but then realized that there are just too many, so I've created this post https://discourse.pulpproject.org/t/every-pulp-repo-that-is-using-sphinx-rtd-theme-needs-to-pin-it-to-1-2-0/775

quba42:
Opened a PR for pulp_deb: https://github.com/pulp/pulp_deb/pull/722

Will merge once it turns green.

PR #3589, opened 2023-02-20
Fix linked related fields

mdellweg:
The refactor somehow broke pulp_container. Reverting the method how to achieve relative links, but keeping the refactor to a mixin here.

[noissue]
re #3584


ggainey (approved):
Even if we have to keep this "revert", keeping the change to make this a mixin is worth it.

ipanova:
pulp_container is a troublemaker!

mdellweg:
> pulp_container is a troublemaker!

How much bugs would we have found in advance if we made pulp_container the plugin to be tested with pulpcore?

ipanova (approved):

Issue #1917, opened 2022-01-17
As a user, I can see the user who dispatched a task on the Task


pulpbot:
Author: @bmbouter (bmbouter)

Redmine Issue: 7180, https://pulp.plan.io/issues/7180

---

## Motivation

1. Users would benefit from knowing which user initiated a task
2. RBAC would benefit from in-a-task knowing which user initiated a task

## Proposal

Add a `user` field  to the Task which would be the Field type provided by `django-currentuser` as shown in [this example](https://github.com/PaesslerAG/django-currentuser#quickstart).

The attribute will be named `created_by`.

pulpbot:
From: @bmbouter (bmbouter)
Date: 2020-07-24T17:43:51Z

---

Added to sprint per open floor this past Tuesday

pulpbot:
From: @bmbouter (bmbouter)
Date: 2020-07-24T17:54:49Z

---

Actually we can't work on this now because we don't have an endpoint for users that we can link to.

bmbouter:
We do track this informtation for RBAC purposes, but it's not user visible right now.

PR #3598, opened 2023-02-23
Expose Task creator on serializer

gerrod3:
fixes: #1917

This is one way to go about solving this issue using values we already store from Roles. We could also add a new field to the `Task` model that is supplied using `django-currentuser` like how the original issue suggests, but I think this approach is easier (no migration!).

### DIFF
--- a/pulpcore/app/serializers/task.py
+++ b/pulpcore/app/serializers/task.py
@@ -1,5 +1,7 @@
 from gettext import gettext as _
 
+from django.conf import settings
+from django.urls import reverse
 from rest_framework import serializers
 
 from pulpcore.app import models
@@ -13,6 +15,7 @@ from pulpcore.app.serializers import (
     TaskGroupStatusCountField,
 )
 from pulpcore.constants import TASK_STATES
+from pulpcore.app.util import get_domain
 
 
 class CreatedResourceSerializer(RelatedResourceField):
@@ -34,6 +37,7 @@ class TaskSerializer(ModelSerializer):
     logging_cid = serializers.CharField(
         help_text=_("The logging correlation id associated with this task")
     )
+    created_by = serializers.SerializerMethodField(help_text=_("User who dispatched this task."))
     started_at = serializers.DateTimeField(
         help_text=_("Timestamp of the when this task started execution."), read_only=True
     )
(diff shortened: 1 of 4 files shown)


pulpcore/app/serializers/task.py
```diff
@@ -84,12 +88,22 @@ class TaskSerializer(ModelSerializer):
...
 
+    def get_created_by(self, obj):
+        if task_user_map := self.context.get("task_user_mapping"):
+            if user_id := task_user_map.get(str(obj.pk)):
+                kwargs = {"pk": user_id}
+                if settings.DOMAIN_ENABLED:
+                    kwargs["pulp_domain"] = get_domain().name
+                return reverse("users-detail", kwargs=kwargs)
+        return None
+
```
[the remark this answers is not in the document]
gerrod3: It's practically the same. In pulpcore-worker we use `get_users_with_perms` which checks for the first user with perms (through roles) on the newly created Task object, which should just be one user with perms from role `core.task_owner`.

mdellweg (approved):

pulpcore/app/viewsets/task.py
```diff
@@ -167,6 +169,12 @@ def get_serializer(self, *args, **kwargs):
                 .only("pk", "number", "repository__pulp_type")
             )
             serializer.context["repo_ver_mapping"] = {rv.pk: rv for rv in repo_vers}
+            # Optimization for User created_by
+            task_pks = [t.pk for t in task_list]
+            user_roles = UserRole.objects.filter(
+                role__name="core.task_owner", object_id__in=task_pks
```
mdellweg, line 175: This basically requires that the access policy for tasks is not modified too much.

Maybe we _should_ capture the current_user on the task more specifically.
Also I was wondering if sometimes someone wants to perform api calls not as a user, but on behalf of a group they are in?

CHANGES/plugin_api/1917.feature
```diff
@@ -0,0 +1 @@
+Expose user who dispatched a Task under new field ``created_by``.
```
ipanova, line 1: this is user visible change, i'd put instead into `CHANGES/1917.feature`

ipanova (approved):
the ci fails because one test needs a small update

PR #3612, opened 2023-02-28
SFTP storage warning

MichalPysik:
Added a warning to the docs that advises against
using the SFTP storage option, since it can lead to data corruption.

[noissue]

### DIFF
--- a/docs/installation/storage.rst
+++ b/docs/installation/storage.rst
@@ -16,6 +16,11 @@ Storage
 SFTP
 ^^^^
 
+.. warning::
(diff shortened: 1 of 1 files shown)


lubosmj:
This is alarming. Have we considered dropping the support for the SFTP backend storage?

MichalPysik:
> This is alarming. Have we considered dropping the support for the SFTP backend storage?

I believe we did, this warning is meant to be temporary until we settle on a working alternative a probably end the SFTP option altogether.

mdellweg (approved):

ggainey (approved):

PR #3618, opened 2023-03-01
[noissue] Bump django to 4.1 branch

decko:
This is just an experiment to see what's gonna fail if we move to the 4.x branch. Since 4.2 is not released yet, we're using the 4.1 branch.

But yet, this is just an experiment. We SHOULD NOT MERGE THIS PR, AT ALL!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,8 +5,9 @@ aiofiles>=22.1,<23.2.0
 backoff>=2.1.2,<2.2.2
 click>=8.1.0,<=8.1.3
 cryptography>=38.0.1,<39.0.3
-Django~=3.2.18  # LTS version, switch only if we have a compelling reason to
-django-currentuser>=0.5.3,<=0.5.3
+Django~=4.1.7  # LTS version, switch only if we have a compelling reason to
+# django-currentuser>=0.5.3,<=0.5.3
+django-currentuser @ https://github.com/PaesslerAG/django-currentuser/archive/refs/heads/master.zip
 django-filter>=22.1,<=22.1
 django-guid>=3.3,<=3.3.0
 django-import-export>=2.9,<3.2.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,8 +81,8 @@ jobs:
       - name: Check for gettext problems
         run: sh .ci/scripts/check_gettext.sh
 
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -5,7 +5,7 @@ aiofiles>=22.1,<23.2.0
 backoff>=2.1.2,<2.2.2
 click>=8.1.0,<=8.1.3
 cryptography>=38.0.1,<39.0.2
-Django~=3.2.18  # LTS version, switch only if we have a compelling reason to
+Django~=4.1.7  # LTS version, switch only if we have a compelling reason to
```
dralley, line 8: What is the compelling reason?  :)  Normally we would skip straight to 4.2 LTS

I've seen some deprecation warnings though even with 3.2 so we probably ought to address those.

decko: This is just an experiment to see what's gonna fail if we move to the 4.x branch. Since 4.2 is not released yet, we're using the 4.1 branch.

But yet, this is just an experiment. We should not merge this PR, AT ALL!

decko: (gonna copy this comment into the PR description to make things clear)

dkliban:
Now need to replace 'force_text' with 'force_str'[0] here[1] and here[2].

[0] https://docs.djangoproject.com/en/4.0/ref/utils/#django.utils.encoding.force_str 
[1] https://github.com/pulp/pulpcore/blob/main/pulpcore/app/templatetags/pulp_urls.py#LL5
[2] https://github.com/pulp/pulpcore/blob/main/pulpcore/app/templatetags/pulp_urls.py#L38

decko:
> Now need to replace 'force_text' with 'force_str'[0] here[1] and here[2].
> 
> [0] https://docs.djangoproject.com/en/4.0/ref/utils/#django.utils.encoding.force_str [1] https://github.com/pulp/pulpcore/blob/main/pulpcore/app/templatetags/pulp_urls.py#LL5 [2] https://github.com/pulp/pulpcore/blob/main/pulpcore/app/templatetags/pulp_urls.py#L38

I believe things like this could be worked out on our current codebase.
Gonna put all those things in a document and on Discourse.

mdellweg:
is `force_str` already available in django 3.2? We could maybe ease the transition if we can prepare the existing codebase a little bit.

PR #3622, opened 2023-03-02
Allow customizing the field name on LabelFilter

### LINKED ISSUE #3631: Allow customizing the field name on LabelFilter
As a plugin writer I would like to be able to customize the field that LabelFilter filters the queryset on.

LabelFilter should accept a parameter that allows it to search on a custom field name.

newswangerd:
This allows for plugins to define a custom field to filter pulp labels on. For example, a distribution filterset could add a filter for repository labels like so:

```
class MyFilterset(Filterset):
    repository__label_select = LabelFilter(label_field_name="repository__pulp_labels")
```

closes #3631

### DIFF
--- a/pulpcore/app/viewsets/custom_filters.py
+++ b/pulpcore/app/viewsets/custom_filters.py
@@ -279,6 +279,10 @@ class LabelFilter(Filter):
 
     def __init__(self, *args, **kwargs):
         kwargs.setdefault("help_text", _("Filter labels by search string"))
+        if "label_field_name" in kwargs:
+            self.label_field_name = kwargs.pop("label_field_name")
+        else:
+            self.label_field_name = "pulp_labels"
         super().__init__(*args, **kwargs)
 
     def filter(self, qs, value):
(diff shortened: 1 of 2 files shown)


pulpcore/app/viewsets/custom_filters.py
```diff
@@ -277,8 +277,12 @@ class CharInFilter(BaseInFilter, CharFilter):
 class LabelFilter(Filter):
     """Filter to get resources that match a label filter string."""
 
+    label_field_name = "pulp_labels"
+
     def __init__(self, *args, **kwargs):
         kwargs.setdefault("help_text", _("Filter labels by search string"))
+        if "label_field_name" in kwargs:
+            self.label_field_name = kwargs.pop("label_field_name")
```
mdellweg, line 285:
```suggestion
        if "label_field_name" in kwargs:
            self.label_field_name = kwargs.pop("label_field_name")
        else:
            self.label_field_name = "pulp_labels"
```
Let's not mix class and instance variables.

mdellweg (commented):
I like this change. We need a changelog for it though.

newswangerd:
> I like this change. We need a changelog for it though.

@mdellweg would this be a feature?

mdellweg:
> > I like this change. We need a changelog for it though.
> 
> @mdellweg would this be a feature?

Yes, but in the `CHANGES/plugin_api` directory.

mdellweg (approved):

ggainey (approved):

PR #3632, opened 2023-03-03
Adds otel WSGI auto-instrumentation

bmbouter:


mdellweg:
There's a `MIDDLEWARE` section in the settings. Why can it not live there?

bmbouter:
> There's a `MIDDLEWARE` section in the settings. Why can it not live there?

It's not a Django middleware so we can't use it like one. It's a generic wsgi, but when being used with Django they recommend [doing it like this](https://opentelemetry-python-kinvolk.readthedocs.io/en/latest/instrumentation/wsgi/wsgi.html#usage-django).

There is also a [django-specific auto-instrumentation](https://opentelemetry-python.readthedocs.io/en/latest/examples/django/README.html#auto-instrumentation), but it doesn't produce the nested SPANs correctly. I am planning to file a bug, but the otel group decided we'd rather just use this working wsgi one since it already was working.

bmbouter:
I would like us to add the exceptions necessary to pick up these dependencies. Yes they dependencies are in beta, but our users need this feature to monitor Pulp running in production. We've tested these out with the pulp-oci-env PR [here](https://github.com/pulp/oci_env/pull/98) and it's all working well.

How do we add the exception?

decko (approved):

pulpcore/app/wsgi.py
```diff
@@ -9,9 +9,12 @@
...
-from django.core.wsgi import get_wsgi_application  # noqa
+from django.core.wsgi import get_wsgi_application
+from opentelemetry.instrumentation.wsgi import OpenTelemetryMiddleware
+
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "pulpcore.app.settings")
 
 
 application = get_wsgi_application()
+application = OpenTelemetryMiddleware(application)
```
mdellweg, line 20:
```suggestion
try:
    from opentelemetry.instrumentation.wsgi import OpenTelemetryMiddleware
    OTEL_AVAILABLE = True
except ImportError:
    OTEL_AVAILABLE = False


os.environ.setdefault("DJANGO_SETTINGS_MODULE", "pulpcore.app.settings")


application = get_wsgi_application()
if OTEL_AVAILABLE:
    application = OpenTelemetryMiddleware(application)
```

It's not hard to make it optional. And then we don't need to depend on a prerelease of a dependency.

bmbouter: Yes but that makes it all more complicated. I want to keep this very simple. Pulp has way too many options.

dralley (approved):

pulpcore/app/wsgi.py
mdellweg: I disagree on all points:
1. That is no excuse for the change provided here. The other more intricate pieces can just about wait until the library we intend to use is released by their maintainers. And quite frankly this piece should too.
2. We are introducing code that gets executed on every single request to the api, and the original authors are not confident to use it in production. "It works" is only about the happy path.
3. It is not the only way that the software is delivered. Pulp is packaged too. Also we should not provide the single most pushed deployment method with unfinished software.

PR #3642, opened 2023-03-07
Return distinct content units from the queryset scope filtering

### LINKED ISSUE #3641: Content endpoints return duplicates when using a non-superuser
**Version**
The main branch; parallelized pulp_rpm.tests.functional.api.test_rbac_crud

**Describe the bug**
After querying an endpoint hosting content referenced by multiple repositories, while logged in a non-superuser, Pulp returns duplicated content.

**To Reproduce**
1. Create a user with `view_content` permissions.
2. Create/Sync two repositories with the same content.
3. Send a GET request with the query parameter `repository_version=one_of_the_repos` to the endpoint hosting the content.
4. Watch the returned duplicates (same pulp_hrefs).

**Expected behaviour**
No duplicates should be returned from the endpoint.

lubosmj:
closes #3641

### DIFF
--- a/pulpcore/app/viewsets/content.py
+++ b/pulpcore/app/viewsets/content.py
@@ -121,7 +121,12 @@ class BaseContentViewSet(NamedModelViewSet):
                 repo_viewset = get_viewset_for_model(repo)()
                 setattr(repo_viewset, "request", self.request)
                 scoped_repos.extend(repo_viewset.get_queryset().values_list("pk", flat=True))
-            return qs.filter(repositories__in=scoped_repos)
+
+            # calling the distinct clause at end of the query ensures that no duplicates from
+            # joined tables will be returned to the end-user; this behaviour is documented at
+            # https://docs.djangoproject.com/en/3.2/topics/db/queries, in the section Spanning
+            # multi-valued relationships
+            return qs.filter(repositories__in=scoped_repos).distinct()
         return qs
 
 
--- a/CHANGES/3641.bugfix
+++ b/CHANGES/3641.bugfix
@@ -1,0 +1,2 @@
+Fixed a bug that caused Pulp to return duplicate content when a user was logged in as a
+non-superuser while the content was part of multiple repositories.


pulpcore/app/viewsets/content.py
```diff
@@ -121,7 +121,7 @@ def scope_queryset(self, qs):
                 repo_viewset = get_viewset_for_model(repo)()
                 setattr(repo_viewset, "request", self.request)
                 scoped_repos.extend(repo_viewset.get_queryset().values_list("pk", flat=True))
-            return qs.filter(repositories__in=scoped_repos)
+            return qs.filter(repositories__in=scoped_repos).distinct()
```
gerrod3, line 124: Still not sure why the query returns duplicate content. Guess the generated SQL did not behave as I thought it would.

dralley: I feel like there's probably a better way to structure this query but yeah, maybe it would be helpful to see what the generated SQL is.

lubosmj:
```
raise RuntimeError(qs.filter(repositories__in=scoped_repos).query)
RuntimeError: SELECT "core_content"."pulp_id", "core_content"."pulp_created", "core_content"."pulp_last_updated", "core_content"."pulp_type", "core_content"."upstream_id", "core_content"."timestamp_of_interest", "rpm_package"."content_ptr_id", "rpm_package"."name", "rpm_package"."epoch", "rpm_package"."version", "rpm_package"."release", "rpm_package"."arch", "rpm_package"."evr", "rpm_package"."pkgId", "rpm_package"."checksum_type", "rpm_package"."summary", "rpm_package"."description", "rpm_package"."url", "rpm_package"."changelogs", "rpm_package"."files", "rpm_package"."requires", "rpm_package"."provides", "rpm_package"."conflicts", "rpm_package"."obsoletes", "rpm_package"."suggests", "rpm_package"."enhances", "rpm_package"."recommends", "rpm_package"."supplements", "rpm_package"."location_base", "rpm_package"."location_href", "rpm_package"."rpm_buildhost", "rpm_package"."rpm_group", "rpm_package"."rpm_license", "rpm_package"."rpm_packager", "rpm_package"."rpm_sourcerpm", "rpm_package"."rpm_vendor", "rpm_package"."rpm_header_start", "rpm_package"."rpm_header_end", "rpm_package"."size_archive", "rpm_package"."size_installed", "rpm_package"."size_package", "rpm_package"."time_build", "rpm_package"."time_file", "rpm_package"."is_modular" FROM "rpm_package" INNER JOIN "core_content" ON ("rpm_package"."content_ptr_id" = "core_content"."pulp_id") INNER JOIN "core_repositorycontent" ON ("core_content"."pulp_id" = "core_repositorycontent"."content_id") WHERE "core_repositorycontent"."repository_id" IN (c35b7039-2c2c-48e3-8f4f-b0eeabad8af1)

```

I think the problem might lie in the joined tables.

gerrod3:
```
FROM "rpm_package" INNER JOIN "core_content" ON ("rpm_package"."content_ptr_id" = "core_content"."pulp_id") 
INNER JOIN "core_repositorycontent" ON ("core_content"."pulp_id" = "core_repositorycontent"."content_id") 
WHERE "core_repositorycontent"."repository_id" IN (c35b7039-2c2c-48e3-8f4f-b0eeabad8af1)
```
Yeah it's the inner join on repositorycontent. This table can have many duplicates even within one repository. We definitely need a distinct clause, unless someone can think of a different way to restructure the query.

mdellweg: I agree, that this should probably be a left join instead. It's is really counterintuitive that a `filter` on a queryset allows items to be multiplicated. Can this be a Django bug?
While the distinct surely works when applied at the end of the queryset, are we sure the is nothing added on top that can revert the effect?

bmbouter:
What needs to happen on this PR to get it merged? I ask because it's blocking pulp_rpm test conversion which is the project critical path for hosted services.

mdellweg:
> What needs to happen on this PR to get it merged? I ask because it's blocking pulp_rpm test conversion which is the project critical path for hosted services.

I think this is solving an existing bug in pulp_rpm, so it critical in any case. What i want is some confidence that it _is_ a solution even if more filters are added at the end of the queryset. Pointing to the proper Django ORM docs would help me.

dralley:
The docs: https://docs.djangoproject.com/en/4.1/ref/models/querysets/#django.db.models.query.QuerySet.distinct

This would be good to bring up on the Django development discord.  Whether it looks like a bug or if there is a better way to structure the calls for a correct result can perhaps be answered there.

lubosmj:
I opened a new ticket here: https://code.djangoproject.com/ticket/34393. I expect us to provide a reproducer if the issue is real.

lubosmj:
We decided to leave the `distinct` clause in the filter query because subsequent filter queries might annul the effort to remove duplicates (e.g., if we would use `subquery`).

gerrod3 (approved):
Unfortunate that the ORM works like that, but now we know. Thanks for the fix!

PR #3644, opened 2023-03-07
Refactor templatetags app to use `force_str` instead of deprecated `force_text`

### LINKED ISSUE #3638: Refactor templatetags app to use force_str instead of deprecated force_text
Since we're testing and moving to Django 4.x(expect to stay on 4.2 LTS when it's released) we expect to find some code marked as deprecated as we move on.

This task is about the moving from the deprecated `force_text` to the new `force_str` function usage on our codebase.
This was spotted here -> https://github.com/pulp/pulpcore/pull/3618#issuecomment-1452483429

decko:
Closes #3638


CHANGES/plugin_api/3638.misc
```diff
@@ -0,0 +1 @@
+Refactor templatetags app to use force_str function instead of deprecated force_text
```
mdellweg, line 1:
```suggestion
Refactor templatetags app to use ``force_str`` function instead of deprecated ``force_text``.
```

mdellweg (changes requested):
Wow, i think, i have never seen this file before...

Single request: Please use proper punctuation and format code in changelogs.

ipanova (approved):

CHANGES/plugin_api/3638.misc
decko: Sorry man, my bad. Was in doubt about if this is used by this kind of file.

lubosmj: Also, we tend to use the simple past tense for changelogs, i.e., `Refactored`.

mdellweg (commented):
Thank you. Still it's a sentence.

mdellweg (approved):

PR #3650, opened 2023-03-09
Raise an exception for invalid data for repository/repair/

lubosmj:
[noissue]

ref #3646

### DIFF
--- a/pulpcore/app/viewsets/repository.py
+++ b/pulpcore/app/viewsets/repository.py
@@ -205,7 +205,7 @@ class RepositoryVersionViewSet(
         """
         version = self.get_object()
         serializer = RepairSerializer(data=request.data)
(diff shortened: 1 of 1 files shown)


gerrod3:
Was this change supposed to be included with https://github.com/pulp/pulpcore/pull/3646/?

lubosmj:
It could be included, but it was not a must. I am fine with reverting the referenced commit. We can include this change there as well.

gerrod3:
No, I'm just wondering if this also needs a changelog. To me it seems like this fix also falls under the last PR's changelog.

lubosmj:
I have referenced the issue from the commit description. Let's say that this is satisfactory.

gerrod3 (approved):

Issue #3654, opened 2023-03-12
pulpcore-content emits an exception when serving an artifact from a Remote


dkliban:
**3.22** but this code has been there since 2019

This feature is used in pulp_maven to create a pull-through cache.

```
ango.db.utils.IntegrityError: duplicate key value violates unique constraint "maven_mavenartifact_group_id_artifact_id_ver_986caee3_uniq"
DETAIL:  Key (group_id, artifact_id, version, filename)=(academy.alex, custommatcher, 1.0, custommatcher-1.0-javadoc.jar.sha1) already exists.


During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/web_protocol.py", line 433, in _handle_request
    resp = await request_handler(request)
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/web_app.py", line 504, in _handle
    resp = await handler(request)
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
  File "/src/pulpcore/pulpcore/content/authentication.py", line 41, in authenticate
    return await handler(request)
  File "/src/pulpcore/pulpcore/content/handler.py", line 243, in stream_content
    return await self._match_and_stream(path, request)
  File "/src/pulpcore/pulpcore/content/handler.py", line 680, in _match_and_stream
    return await self._stream_remote_artifact(
  File "/src/pulpcore/pulpcore/content/handler.py", line 984, in _stream_remote_artifact
    await asyncio.shield(
  File "/usr/local/lib/python3.8/site-packages/asgiref/sync.py", line 448, in __call__
    ret = await asyncio.wait_for(future, timeout=None)
  File "/usr/lib64/python3.8/asyncio/tasks.py", line 455, in wait_for
    return await fut
  File "/usr/lib64/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/local/lib/python3.8/site-packages/asgiref/sync.py", line 490, in thread_handler
    return func(*args, **kwargs)
  File "/src/pulpcore/pulpcore/content/handler.py", line 799, in _save_artifact
    if artifact.sha256 != artifacts.get().sha256:
TypeError: 'ManyRelatedManager' object is not subscriptable
```

PR #3655, opened 2023-03-12
Properly access artifacts of a generic content in handler.py

dkliban:
closes: #3654

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -802,7 +802,7 @@ class Handler:
                     # There is already content for this Artifact
                     content = c_type.objects.get(content.q())
                     artifacts = content._artifacts
-                    if artifact.sha256 != artifacts[0].sha256:
+                    if artifact.sha256 != artifacts.get().sha256:
                         raise RuntimeError(
                             "The Artifact downloaded during pull-through does not "
                             "match the Artifact already stored for the same "
--- a/CHANGES/3654.bugfix
+++ b/CHANGES/3654.bugfix
@@ -1,0 +1,1 @@
+Fixed bug related to pulpcore-content serving content from a remote (pull-through-cache).


pulpcore/content/handler.py
```diff
@@ -802,7 +802,7 @@ def _save_artifact(self, download_result, remote_artifact):
                     # There is already content for this Artifact
                     content = c_type.objects.get(content.q())
                     artifacts = content._artifacts
-                    if artifact.sha256 != artifacts[0].sha256:
+                    if artifact.sha256 != artifacts.get().sha256:
```
mdellweg, line 805: I don't understand the difference here. We assume that the `artifacts` queryset represents exactly one object, right? If it does not, both ways of receiving it will result in an exception. If it does, they produce the same output.

Does the type of the raised exception matter?

dkliban: The current exception is `TypeError: 'ManyRelatedManager' object is not subscriptable`.

dkliban: Here is what chatGPT had to say about the change between Django 2 and 3:


Django 3 made a change to the ManyToManyField relation and how it returns related objects. In Django 2, accessing related objects via a ManyToManyField returned a RelatedManager instance, while in Django 3, it returns a ManyRelatedManager instance.

The main difference between these two is that a RelatedManager instance is a subclass of Django's Manager class, while a ManyRelatedManager instance is not. As a result, some methods that were available on the RelatedManager instance in Django 2 may not be available on the ManyRelatedManager instance in Django 3.

One particular change that might be causing the TypeError you're seeing is that the RelatedManager instance in Django 2 was subscriptable, meaning you could use it like a list or queryset to access related objects. However, the ManyRelatedManager instance in Django 3 is not subscriptable, so trying to use it in this way will result in a TypeError.

mdellweg (approved):

ggainey (approved):

PR #3662, opened 2023-03-16
Immediate tasks

mdellweg:


mdellweg:
Let me add some documentation for this (including large disclaimers).

pulpcore/tasking/tasks.py
```diff
@@ -91,19 +140,54 @@ def dispatch(
...
+        if immediate:
+            can_execute = True
+            prior_tasks = Task.objects.filter(
+                state__in=TASK_INCOMPLETE_STATES, pulp_created__lte=task.pulp_created
+            )
+            for resource in resources:
+                if resource.startswith("shared:"):
+                    resource = resource[7:]
+                    if prior_tasks.filter(reserved_resources_record__contains=[resource]).exists():
+                        can_execute = False
```
gerrod3, line 170: Is there a benefit to having this done in multiple DB calls? Could we not do all the resource checking in Python after one DB call?

pulpcore/tasking/tasks.py
```diff
@@ -91,19 +140,54 @@ def dispatch(
...
+                else:
+                    if (
+                        prior_tasks.filter(reserved_resources_record__contains=[resource]).exists()
+                        or prior_tasks.filter(
+                            reserved_resources_record__contains=["shared:" + resource]
+                        ).exists()
+                    ):
+                        can_execute = False
+                        break
+            if can_execute:  # Simulate reserved resources for now
```
gerrod3, line 181:
```suggestion
            if can_execute:
```

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -12,46 +10,36 @@
...
 
-import django
+from django.conf import settings
+from django.db import connection
+from django.utils import timezone
+from django_currentuser.middleware import _set_current_user
+from django_guid import set_guid
 
-os.environ.setdefault("DJANGO_SETTINGS_MODULE", "pulpcore.app.settings")
-django.setup()
```
gerrod3, line 24: Do we really not need to call `django.setup()` anymore? How come?

mdellweg: that's what we have the entrypoint.py for. This was superfluous.

pulpcore/tasking/tasks.py
mdellweg: Need to think about it. So you basically ask whether we can check for overlap of resources taken an resources needed in a single db query.

pulpcore/app/viewsets/base.py
```diff
@@ -506,6 +506,7 @@ def update(self, request, pk, **kwargs):
             exclusive_resources=self.async_reserved_resources(instance),
             args=(pk, app_label, serializer.__class__.__name__),
             kwargs={"data": request.data, "partial": partial},
+            immediate=True,
```
ipanova, line 509: you probably need a changelog entry here, this is a new behaviour that needs to be outlined.

mdellweg:
Still in draft, because i want to look whether the db calls to check for resources can be improved.

pulpcore/tasking/tasks.py
```diff
@@ -68,13 +110,20 @@ def dispatch(
             running. Each resource can be either a `str` or a `django.models.Model` instance.
         shared_resources (list): A list of resources this task needs non-exclusive access to while
             running. Each resource can be either a `str` or a `django.models.Model` instance.
+        immediate (bool): Whether to allow running this task immediately. It must be guaranteed to
+            execute fast without blocking. If not all resource constraints are met, the task will
+            be returned in a canceled state. Defaults to `False`.
```
gerrod3, line 115: `dispatch` can be called from within a task, what happens if you call it within with `immediate=True`?

pulpcore/tasking/tasks.py
```diff
@@ -91,19 +140,53 @@ def dispatch(
...
+                execute_task(task)
+                if resources:
+                    notify_workers = True
+            elif deferred:
+                notify_workers = True
+            else:
+                task.set_canceling()
+                task.set_canceled(TASK_STATES.CANCELED, "Resources temporarily unavailable.")
+    if notify_workers:
+        _wakeup_worker()
```
gerrod3, line 189: Do you think there are any side-effects now that this `_wakeup_worker` is no longer done on the task save commit?

pulpcore/tasking/tasks.py
```diff
@@ -91,19 +140,53 @@ def dispatch(
...
+            if immediate:
+                # Grab the advisory lock before the task hits the db.
+                stack.enter_context(task)
+            else:
+                notify_workers = True
+        if immediate:
+            prior_tasks = Task.objects.filter(
+                state__in=TASK_INCOMPLETE_STATES, pulp_created__lt=task.pulp_created
+            )
+            # Translate resources into a list of resourced that mus not be taken.
```
gerrod3, line 165:
```suggestion
            # Translate resources into a list that must not be taken.
```

pulpcore/tasking/tasks.py
```diff
@@ -91,19 +140,53 @@ def dispatch(
...
+                    colliding_resources.append(resource[7:])
+                else:
+                    colliding_resources.append(resource)
+                    colliding_resources.append("shared:" + resource)
+            # Can we execute this task immediatly?
+            if (
+                not colliding_resources
+                or not prior_tasks.filter(
+                    reserved_resources_record__overlap=colliding_resources
+                ).exists()
```
gerrod3, line 178: Nice! :+1:

pulpcore/app/models/repository.py
```diff
@@ -108,9 +109,8 @@ def save(self, *args, **kwargs):
                 self.create_initial_version()
 
                 # lock the repository if it was created from within a running task
-                try:
-                    task_id = environ["PULP_TASK_ID"]
-                except KeyError:
```
gerrod3, line 113: Have you checked that no other plugin was using `PULP_TASK_ID` in their code?

pulpcore/tasking/tasks.py
mdellweg: It will try to execute the task in that other tasks context. The ContextVar will then ensure the right task will be referenced by `Task.current()`. (I have not tried this. My conclusion is from theory.)

pulpcore/tasking/tasks.py
mdellweg: It is only important, that it is not called before the task is actually committed to the database. So on_commit was the first place to do it. Waking up workers a few ms later should be ok.

mdellweg: To be explicit. I had it in the migration first, leading to the worker waking up finding no task (because not yet committed) and going back to sleep before the transaction ended. -> Task system stalled.

pulpcore/app/models/repository.py
mdellweg: Is the environment part of the plugin api? `Task.current()` on the other hand is. I would not consider this a breaking change. But no checked it I have not.

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -212,18 +212,34 @@ same resources are not run at the same time. To ensure data correctness, any act
...
+When dispatching a task, one can specify whether it is eligible to be run immediately in the same
+process right away given the reservations can be satisfied (defaults to ``False``), and whether it
+is eligible to be deferred for workers to be picked up later (defaults to ``True``).
+In case a task was marked for immediate execution, but the reservations were not satisfiable, it
+will be left in the task queue or marked as canceled, depending on the ``deferred`` attribute.
+
+.. warning::
+
+   A task marked for immediate execution will not be isolated in the ``pulpcore-worker``, but be
+   executed in the current python environment. Also it will delay the response to the http call.
```
dralley, line 239: The language here should be a bit stronger.  We use gunicorn in synchronous mode, and AFAIK our default for the number of workers for the API server is pretty low (one or two?).  Each worker defaults to one thread, and we don't override it.  So you're not just blocking the http request, but the whole server potentially, depending on how gunicorn is configured in the service file.

And since this happens in a WSGI context which is fundamentally not async, any calls to the database within such a task are technically blocking IO by definition.  In fact just creating and updating the task record itself is technically going to be "blocking".

pulpcore/app/models/repository.py
dralley: If you do an org-wide search this is the result: https://github.com/pulp/pulpcore/search?q=PULP_TASK_ID

I would maybe look at the commits that added those lines to see if there's some history about why it's there.

CHANGES/3661.feature
```diff
@@ -0,0 +1 @@
+Changed most of the update and delete tasks to allow for immediate execution.
```
dralley, line 1: Would it be worth expanding to ... "allow for immediate execution if not contested by other running tasks"?

pulpcore/tasking/tasks.py
```diff
@@ -91,19 +140,53 @@ def dispatch(
...
+            )
+            # Translate resources into a list of resourced that mus not be taken.
+            colliding_resources = []
+            for resource in resources:
+                if resource.startswith("shared:"):
+                    colliding_resources.append(resource[7:])
+                else:
+                    colliding_resources.append(resource)
+                    colliding_resources.append("shared:" + resource)
+            # Can we execute this task immediatly?
```
dralley, line 173: typo

pulpcore/app/models/repository.py
mdellweg: Easy, i made it because i didn't know of ContextVars.

pulpcore/tasking/tasks.py
mdellweg: To be even more explicit, I reworded to:
```python
# Compile a list of resources that must not be taken by other tasks.
```

docs/plugin_dev/plugin-writer/concepts/index.rst
mdellweg: Can you take a look again?

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -212,18 +212,35 @@ same resources are not run at the same time. To ensure data correctness, any act
...
+
+When dispatching a task, one can specify whether it is eligible to be run immediately in the same
+process right away given the reservations can be satisfied (defaults to ``False``), and whether it
+is eligible to be deferred for workers to be picked up later (defaults to ``True``).
+In case a task was marked for immediate execution, but the reservations were not satisfiable, it
+will be left in the task queue or marked as canceled, depending on the ``deferred`` attribute.
+
+.. warning::
+
+   A task marked for immediate execution will not be isolated in the ``pulpcore-worker``, but be
```
dralley, line 238: s/but be/but may be

dralley: Otherwise it looks ok``

dralley (approved):

CHANGES/3661.feature
```diff
@@ -0,0 +1,2 @@
+Changed most of the update and delete tasks to allow for immediate execution if not contested by
```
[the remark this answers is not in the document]
mdellweg: No, that one can include (at least for content) extra time intense validations. Most of the create calls are not even using a task, so the ones that do must have a reason other than locking on the created resource.

CHANGES/plugin_api/3661.feature
```diff
@@ -0,0 +1 @@
+Added immediate tasks.
```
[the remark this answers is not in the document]
mdellweg: I kind of want readers to look at the docs. There are the big red disclaimers.
Do you want me to add "It's a dangerous too. If you overuse it we will take it away again."?
Seriously what about:
```suggestion
Added flags to ``dispatch`` to allow specifying whether a task can be attempted to run synchronously and whether it can be deferred?
```

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -212,18 +212,35 @@ same resources are not run at the same time. To ensure data correctness, any act
...
+   executed in the current api worker. This will not only delay the response to the http call, but
+   also the complete single threaded gunicorn process. To prevent degrading the whole Pulp service,
```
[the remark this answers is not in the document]
mdellweg: I think "delay" serves both parts of the sentence here. Also i'm using "whole" in the sentence after already.

pulpcore/tasking/tasks.py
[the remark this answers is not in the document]
mdellweg: Right this was for only half of the implementation.

pulpcore/tasking/tasks.py
```diff
@@ -91,19 +140,53 @@ def dispatch(
...
+                state__in=TASK_INCOMPLETE_STATES, pulp_created__lt=task.pulp_created
+            )
+            # Compile a list of resources that must not be taken by other tasks.
+            colliding_resources = []
+            for resource in resources:
+                if resource.startswith("shared:"):
+                    colliding_resources.append(resource[7:])
+                else:
+                    colliding_resources.append(resource)
+                    colliding_resources.append("shared:" + resource)
```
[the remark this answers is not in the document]
ipanova: > I this task needs a resource exclusively, then no other task may have it either shared or exclusive. That's why we are building a list of "resources" that must have no overlap with competing tasks.

oh right, this is correct, my brain just clicked and it's all clear now!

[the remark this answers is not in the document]
mdellweg: I reworked it. Please take a close look.

CHANGES/3661.feature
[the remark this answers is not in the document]
ipanova: i think the only create call that uses a task is distribution creation

CHANGES/plugin_api/3661.feature
```diff
@@ -0,0 +1,2 @@
+Added flags ``immediate`` and ``deferred`` to ``dispatch`` to allow specifying whether a task can
+be attempted to run synchronously and whether it can be deferred?
```
ipanova, line 2: deferred to pulp-worker

pulpcore/tasking/tasks.py
```diff
@@ -68,42 +110,81 @@ def dispatch(
...
-
-    resources = _validate_and_get_resources(exclusive_resources)
-    if shared_resources:
-        resources.extend(
-            (f"shared:{resource}" for resource in _validate_and_get_resources(shared_resources))
-        )
+    else:
+        exclusive_resources = _validate_and_get_resources(exclusive_resources)
+    if shared_resources is None:
+        shared_resources = _validate_and_get_resources(shared_resources)
```
ipanova, line 136: you need to swap the lines here, it should be empty list instead in this branch

pulpcore/tasking/tasks.py
```diff
@@ -68,42 +110,81 @@ def dispatch(
...
+                stack.enter_context(task)
+            else:
+                notify_workers = True
+        if immediate:
+            prior_tasks = Task.objects.filter(
+                state__in=TASK_INCOMPLETE_STATES, pulp_created__lt=task.pulp_created
+            )
+            # Compile a list of resources that must not be taken by other tasks.
+            colliding_resources = shared_resources + [
+                f"shared:{resource}" for resource in exclusive_resources
```
ipanova, line 169: i think you need to here to + exclusive_resources

pulpcore/tasking/tasks.py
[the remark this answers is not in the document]
ipanova: i like it

ipanova:
please address me last 2 comments and it's good to go!

pulpcore/tasking/tasks.py
mdellweg: right!

pulpcore/tasking/tasks.py
mdellweg: yes!

CHANGES/3661.feature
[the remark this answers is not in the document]
mdellweg: And all content types... Let's say, it's way too risky to set this in the AsyncCreateMixin.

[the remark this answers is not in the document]
dralley: Right, we can't do that.  Parsing (for example) an RPM inside of the web server process feels like a bad idea to me :)

ipanova (approved):
Thanks!

pulpcore/tasking/tasks.py
```diff
@@ -68,42 +110,83 @@ def dispatch(
...
-            args=args,
-            kwargs=kwargs,
-            parent_task=Task.current(),
-            reserved_resources_record=resources,
-        )
-        transaction.on_commit(_wakeup_worker)
-
+        if domain_url not in exclusive_resources:
+            shared_resources.append(domain_url)
+    resources = exclusive_recources + [f"shared:{resource}" for resource in shared_resources]
```
ipanova, line 143: linting is failing on this typo `exclusive_resources`

CHANGES/3661.feature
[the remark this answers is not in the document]
dralley: You're right.  I was confused because it does dispatch a task.  But it's not the same codepath.

PR #3664, opened 2023-03-17
Fixed import/export in the face of Domains.

### LINKED ISSUE #3663: PuilpImport fails post-Domains (whether domains are enabled or not)
**Version**
core/3.23

**Describe the bug**
Domains adds a field to content "pulp_domain", which contains the pulp_id of the associated domain (or of the default-domain). Export exports that field, so import tries to hook it back up on the downstream side. The downstream "default" domain has a different pulp_domain, so the import fails with a traceback like this:
```
pulp [b2a311abd716444aada1d7ff310a5eb1]: pulpcore.tasking.pulpcore_worker:INFO:   File "/src/pulpcore/pulpcore/tasking/pulpcore_worker.py", line 458, in _perform_task
  File "/src/pulpcore/pulpcore/app/tasks/importer.py", line 482, in pulp_import
    for ar_result in _import_file(os.path.join(temp_dir, ARTIFACT_FILE), ArtifactResource):
  File "/src/pulpcore/pulpcore/app/tasks/importer.py", line 143, in _import_file
    a_result = resource.import_data(data, raise_errors=True)
  File "/usr/local/lib/python3.8/site-packages/import_export/resources.py", line 813, in import_data
    result = self.import_data_inner(
  File "/usr/local/lib/python3.8/site-packages/import_export/resources.py", line 882, in import_data_inner
    raise row_result.errors[-1].error
  File "/usr/local/lib/python3.8/site-packages/import_export/resources.py", line 737, in import_row
    self.import_obj(instance, row, dry_run, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/import_export/resources.py", line 557, in import_obj
    self.import_field(field, obj, data, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/import_export/resources.py", line 540, in import_field
    field.save(obj, data, is_m2m, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/import_export/fields.py", line 119, in save
    cleaned = self.clean(data, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/import_export/fields.py", line 75, in clean
    value = self.widget.clean(value, row=data, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/import_export/widgets.py", line 423, in clean
    return self.get_queryset(value, row, **kwargs).get(**{self.field: val})
  File "/usr/local/lib/python3.8/site-packages/django/db/models/query.py", line 435, in get
    raise self.model.DoesNotExist(
```

**To Reproduce**
* Sync a repo.
* Export it.
* Reset the pulp-database (or move the export to a different instance)
* Attempt to import it.

**Expected behavior**
Import succeeds.

ggainey:
We can't export entities' domain-info.

fixes #3663.
[nocoverage]

### DIFF
--- a/CHANGES/3663.bugfix
+++ b/CHANGES/3663.bugfix
@@ -1,0 +1,1 @@
+Fixed import/export in the presence of the domains work.
--- a/pulpcore/plugin/importexport.py
+++ b/pulpcore/plugin/importexport.py
@@ -1,4 +1,5 @@
 from import_export import resources
+from pulpcore.app.util import get_domain_pk
 
 
 class QueryModelResource(resources.ModelResource):
@@ -18,6 +19,21 @@ class QueryModelResource(resources.ModelResource):
         queryset (django.db.models.query.QuerySet): filtering queryset for this resource
             (driven by repo_version)
     """
+
+    def before_import_row(self, row, **kwargs):
+        """
+        Sets pulp_domain/_pulp_domain to the current-domain on import.
+        Args:
+            row (tablib.Dataset row): incoming import-row representing a single Variant.
+            kwargs: args passed along from the import() call.
+        """
+        # There is probably a more pythonic/elegant way to do the following - but I am deliberately
+        # opting for "verbose but COMPLETELY CLEAR" here.
+        if "_pulp_domain" in row:
+            row["_pulp_domain"] = get_domain_pk()
+
+        if "pulp_domain" in row:
+            row["pulp_domain"] = get_domain_pk()
 
     def set_up_queryset(self):
         return None
--- a/pulpcore/app/modelresource.py
+++ b/pulpcore/app/modelresource.py
@@ -31,6 +31,8 @@ class ArtifactResource(QueryModelResource):
             kwargs: args passed along from the import() call.
 
         """
+        super().before_import_row(row, **kwargs)
+
         # the export converts None to blank strings but sha384 and sha512 have unique constraints
         # that get triggered if they are blank. convert checksums back into None if they are blank.
         for checksum in ALL_KNOWN_CONTENT_CHECKSUMS:
@@ -44,13 +46,19 @@ class ArtifactResource(QueryModelResource):
             "pulp_created",
             "pulp_last_updated",
         )
-        import_id_fields = ("sha256",)
+        import_id_fields = (
+            "pulp_domain",
+            "sha256",
+        )
 
 
 class RepositoryResource(QueryModelResource):
     class Meta:
         model = Repository
-        import_id_fields = ("name",)
+        import_id_fields = (
+            "pulp_domain",
+            "name",
+        )
         exclude = (
             "pulp_id",
             "pulp_created",
@@ -98,6 +106,7 @@ class ContentArtifactResource(QueryModelResource):
         Returns:
             (tablib.Dataset row): row that now points to the new downstream uuid for its content.
         """
+        super().before_import_row(row, **kwargs)
 
         linked_content = Content.objects.get(upstream_id=row["content"])
         row["content"] = str(linked_content.pulp_id)


CHANGES/3663.bugfix
```diff
@@ -0,0 +1 @@
+Fixed import/export in the presence on the domains work.
```
gerrod3, line 1: Do you mean s/on/of/ ?

pulpcore/plugin/importexport.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, repo_version=None):
             self.queryset = self.set_up_queryset()
 
     class Meta:
-        exclude = ("pulp_id", "pulp_created", "pulp_last_updated")
+        exclude = ("pulp_domain", "pulp_id", "pulp_created", "pulp_last_updated")
```
gerrod3, line 31: Are `UserRole` and `GroupRole`s exported/imported? They are the only objects that have a domain relation that is not named pulp_domain.

gerrod3 (commented):
Seems simple enough, but more greedily can we get a pulp_file PR with more export/import tests that properly test this.

pulpcore/plugin/importexport.py
ggainey: They are not (very deliberately) - so we're good (so far)

ggainey:
> Seems simple enough, but more greedily can we get a pulp_file PR with more export/import tests that properly test this.

Testing the problem requires a sync, an export, and a complete db-reset before importing (because if the db you import-into contains the same default-org as the one you exported-from, you won't be testing anything at import - that's how this was missed originally). Pretty sure we *can* do so, but I wanted to get the fix out ASAP. Can you open a pulp_file issue to remind us/me to create such a test #soon?

mdellweg (commented):
Do we need to / can we exclude "_pulp_domain" for all the content exported, too?

ggainey:
> Can we mark db fields to exclude exportation on the models where we define them? (Just a crazy thought.)

The models know nothing about import/export, and I'd like to keep it that way.

ggainey:
> Do we need to / can we exclude "_pulp_domain" for all the content exported, too?

We can - but I don't see that anywhere. Where does it show up as a field?

ggainey:
> This field is going to be present on every child Content model, it's the field they will use for changing their `unique_together`. pulp_file example: https://github.com/pulp/pulp_file/blob/main/pulp_file/app/models.py#L41

Ah! I missed that. Yeah, will do.

pulpcore/plugin/importexport.py
```diff
@@ -19,6 +20,16 @@ class QueryModelResource(resources.ModelResource):
...
 
+    def before_import_row(self, row, **kwargs):
+        """
+        Sets _pulp_domain to the current-domain on import.
+        Args:
+            row (tablib.Dataset row): incoming import-row representing a single Variant.
+            kwargs: args passed along from the import() call.
+        """
+        if "_pulp_domain" in row:
+            row["_pulp_domain"] = get_domain_pk()
```
mdellweg, line 31: Maybe i'm mistaken. But row is what we get from the file imported, right? So as far as i understand this should never include the domain. So should the check not be somthing like:
`if hasattr(self.model, "_pulp_domain")`?

Should we account for `__pulp_domain` too in case we have content that is subclassed multiple times?

mdellweg: Also would it not be a good idea to populate `pulp_domain` on all models that have it too?

Whould this maybe _just_ make import export domains compatible?

mdellweg (changes requested):

pulpcore/plugin/importexport.py
ggainey: > Should we account for `__pulp_domain` too in case we have content that is subclassed multiple times?

It *has to* inclulde _pulp_domain, because that's become part of the natural-key of content. So it has to be exported, and then replaced on-import with the approapriate domain we're importing-into.

ggainey: > Also would it not be a good idea to populate `pulp_domain` on all models that have it too?
> 
> Whould this maybe _just_ make import export domains compatible?

That's a good question, `pulp_domain` (as opposed to `_pulp_domain`) lives in Artifact and Content.  @gerrod3 , what happoens w/ artifact/content is pulp_domain isn't specified?

mdellweg: We don't export across domains, right? All repos of one export need to be in the same domain.

ggainey: See https://docs.pulpproject.org/pulpcore/plugin_dev/plugin-writer/concepts/domains/domains_compatibility.html#add-domain-relation-to-plugin-models - I read that as "you only add _pulp_domain if you subclass directly from Content"

ggainey: > We don't export across domains, right? All repos of one export need to be in the same domain.

One could *export* across-domains - but one *imports* into a single domain. What we're doing/have to do here, is replace whatever domain-uuid the exported content has/with the appropriate "current" domain on import.

The reason we choose to export a domain-field that we're just going to replace, is to make it so the import side doesn't have to make any decisions around "should I add a current-domain UUID into every row" - if the field exists, replace it, otherwise skip.

Currently in this PR we're "getting away with" not-exporting "pulp_domain" from the modelresources that carry it, but I'm thinking we probably want to do this replacement to that field (if at exists) as well. @gerrod3 , wdyt?

mdellweg: I mean, a single export can only include repositories of one domain, no? And the import of it goes into a single (maybe different) domain, also. I believe that is what we agree on.
So if we could make sure, the domain_field will never make it into an exported file, will pulp not populate that field automatically with `current_domain_pk`? Maybe that is too late for finding existing entities.
To me the logic is still: Does my model have a pulp_domain field? Then i need to populate it on import.

ggainey:
pulp_rpm and pulp_ansible override before_import_row(), without calling super(). We have to fix that, for this to work.

I'm going to Draft this PR while I consider.

ggainey:
Commentary from another discussion w/ @gerrod3 : we need to add `pulp_domain` to the **import-id-fields** of Artifact and Repository to get them to behave correctly, at import-time, in the face of multiple domains.

ggainey:
OK, this incarnation appears to do the deed. More eyes please, folks.

mdellweg (approved):

PR #3668, opened 2023-03-22
Fix the search bar bug

### LINKED ISSUE #3667: The search bar issue in sphinx docs persists
https://github.com/pulp/pulpcore/issues/3569

The search feature bug in docs was not resolved.

https://docs.pulpproject.org/pulpcore

lubosmj:
closes #3667

### DIFF
--- a/CHANGES/3667.doc
+++ b/CHANGES/3667.doc
@@ -1,0 +1,1 @@
+Fixed infinite loading when searching for specific terms.
--- a/doc_requirements.txt
+++ b/doc_requirements.txt
@@ -1,7 +1,8 @@
 -r requirements.txt
 plantuml
 sphinx
-sphinx-rtd-theme<1.2.0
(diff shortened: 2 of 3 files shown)


lubosmj:
Should I introduce the same change in the plugin template, or are we addressing the issue separately based on a plugin?

ref: https://discourse.pulpproject.org/t/every-pulp-repo-that-is-using-sphinx-rtd-theme-needs-to-pin-it-to-1-2-0/775

lubosmj:
It is.

https://github.com/pulp/plugin_template/blob/main/templates/docs/doc_requirements.txt.j2

ggainey:
> It is.
> 
> https://github.com/pulp/plugin_template/blob/main/templates/docs/doc_requirements.txt.j2

Then I'd think fixing it in plugin_template is The Way, so everyone gets it next run, yeah?

gerrod3 (approved):

Issue #3653, opened 2023-03-10
Package timestamp on the content dir listing page should reflect the time when it was added to the repo


daviddavis:
In pulpcore 3.22, a timestamp was added to the content directory listing page. The timestamp uses the content artifact created date. However, it should use the time when the package was added to the repo (the RepositoryContent created time) as this is what users would expect especially in cases where someone else might have uploaded the packages.

daviddavis:
Our use case: users promote packages from one repo to the next and they want to know at which time the package was "published" (or added to a repo).

dkliban:
I see this as a bug. The pages displayed by pulpcore-content are designed for users of the repository to browse the content of the repository. The date a user expects to see is that of when the package became available in that repository.

PR #3672, opened 2023-03-24
Fix package timestamp on content app

gerrod3:
fixes: #3653

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -364,8 +364,7 @@ class Handler:
         Returns:
             String representing HTML of the directory listing.
         """
-        if dates is None:
-            dates = dict()
+        dates = dates or {}
         template = Template(
             """
 <html>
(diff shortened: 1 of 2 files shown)


pulpcore/content/handler.py
```diff
@@ -420,37 +419,41 @@ def list_directory_blocking():
...
 
             if directory_list:
+                # Find the dates the content got added to the repository
+                dates.update(
+                    {
+                        content_to_find[rc.content_id]: rc.pulp_created
+                        for rc in content_repo_ver._content_relationships()
+                        if rc.content_id in content_to_find
+                    }
+                )
```
dralley, line 455: I'm a bit concerned that this will wreck performance.  I suppose we have no way to tell until we profile and we can't profile (easily) until we have the opentelemetry work ready.  @decko Down the road this might be an interesting thing to "trace", alongside the tasks list API (which I recall had pretty bad performance last time I looked).

dralley (approved):

PR #3706, opened 2023-04-03
Document how to depend on another components PR

mdellweg:
[noissue]

### DIFF
--- a/docs/contributing/git.rst
+++ b/docs/contributing/git.rst
@@ -75,33 +75,35 @@ Putting this all together, the following is an example of a good commit message:
 Requiring other Pull Requests
 *****************************
 
-Sometimes a new feature may require changes to both `pulpcore` and one or many other plugins,
-simultaneously. In order to keep the CI happy in these circumstances (as tests may fail otherwise),
-we provide a mechanism to force the CI service to fetch the version of a component in a linked
-Pull Request, rather than master branch.
+Sometimes a new feature may require changes to both `pulpcore` and one or many other plugins.
+However, plugins can only depend on features that are already released with `pulpcore` or any other
+dependency. Sometimes though you need to demonstrate, that a new feature about to be added to
+`pulpcore` will work with a corresponding plugin change before you can get the needed approvals. In
+order to do so, you can depend the plugin's pull request on the head of the pull request or the
+main branch of `pulpcore` in the following way:
 
-To do so, add a tag in the following format to the last commit in your series.
+Add a line like::
 
-This will allow the PR against pulp to run against the Pull Requests for pulp_file::
+    git+https://github.com/pulp/pulpcore@refs/pull/1234/head
+    git+https://github.com/pulp/pulpcore@refs/heads/main
 
-    Required PR: https://github.com/pulp/pulp_file/pull/2345
+to `ci_requirements.txt` in the plugin PR. Make sure that file is covered by `MANIFEST.in`. Also
+bump the requirement on `pulpcore` in `requirements.txt` to at least the current `dev` version if
+you want to be sure the `lower bounds` scenario passes.
 
-This will allow the PR against a plugin to run against the Pull Request for pulpcore::
+This works accordingly for depending on other plugins.
 
-    Required PR: https://github.com/pulp/pulpcore/pull/3456
-
-Attention and care must be given to merging PRs that require other Pull Requests. Before merging,
-all required PRs should be ready to merge--meaning that all tests/checks should be passing, the code
-review requirements should be met, etc. When merging, the PR along with its required PRs should all
-be merged at the same time. This is necessary to ensure that test breakages don't block other PRs.
+This will allow the tests in the CI to run, but it will fail the `ready-to-ship` check. The
+depended on PR must be merged **and** released before a PR like this can be merged.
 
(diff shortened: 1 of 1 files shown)


docs/contributing/git.rst
```diff
@@ -75,33 +75,29 @@ Putting this all together, the following is an example of a good commit message:
...
 
-To do so, add a tag in the following format to the last commit in your series.
+Add a line like::
 
-This will allow the PR against pulp to run against the Pull Requests for pulp_file::
+    git+https://github.com/pulp/pulpcore@refs/pull/1234/head
 
-    Required PR: https://github.com/pulp/pulp_file/pull/2345
+to `ci_requirements.txt` in the plugin PR. Make sure that file is covered by `MANIFEST.in`. Also
+bump the requirement on `pulpcore` in `requirements.txt` to at least the current `dev` version.
```
lubosmj, line 87: Could you also make some words in the last sentence bold?

**bump the requirement**

lubosmj (approved):

docs/contributing/git.rst
mdellweg:
```suggestion
to `ci_requirements.txt` in the plugin PR. Make sure that file is covered by `MANIFEST.in`. Also
**bump the requirement** on `pulpcore` in `requirements.txt` to at least the current `dev` version.
```
Like this? I'm not sure it is worth it. As long as the git branch satisfies the requirement, at least some tests will run (I suspect the lower bounds to fail if you missed this step), but the PR is in an unmergable state anyway at this point.

lubosmj: I just wanted to put emphasis on the requirements bumping because it is something we may miss later.

mdellweg: Once you brush up the PR (when pulpcore released the depended on parts), the lower bounds test should tell you.

docs/contributing/git.rst
```diff
@@ -75,33 +75,29 @@ Putting this all together, the following is an example of a good commit message:
 Requiring other Pull Requests
 *****************************
 
-Sometimes a new feature may require changes to both `pulpcore` and one or many other plugins,
-simultaneously. In order to keep the CI happy in these circumstances (as tests may fail otherwise),
-we provide a mechanism to force the CI service to fetch the version of a component in a linked
-Pull Request, rather than master branch.
+Sometimes a new feature may require changes to both `pulpcore` and one or many other plugins.
+In order to demonstrate the new feature in the plugin CI, you can depend on the tip of the pull
```
ipanova, line 79: s/tip/head/

docs/contributing/git.rst
```diff
@@ -75,33 +75,29 @@ Putting this all together, the following is an example of a good commit message:
...
-This will allow the PR against a plugin to run against the Pull Request for pulpcore::
-
-    Required PR: https://github.com/pulp/pulpcore/pull/3456
-
-Attention and care must be given to merging PRs that require other Pull Requests. Before merging,
-all required PRs should be ready to merge--meaning that all tests/checks should be passing, the code
-review requirements should be met, etc. When merging, the PR along with its required PRs should all
-be merged at the same time. This is necessary to ensure that test breakages don't block other PRs.
+This will allow the CI to run, but it will fail the `ready-to-ship` test. The depended on PR must
+be merged and released before a PR like this can be merged.
```
ipanova, line 91: once the dependent PR in pulpcore is merged, should not the line from ci_requirements be removed?

ipanova: you say the dependent PR should be merged and released, that's a bit of inconvenience for plugins. Can we merge the corresponding PR in plugin with the pulpcore version bump at least once the pulpcore PR is merged ( not necessarily released)?

docs/contributing/git.rst
mdellweg:
```suggestion
In order to demonstrate the new feature in the plugin CI, you can depend on the head of the pull
```

docs/contributing/git.rst
mdellweg: No, because we only integrate new features with released versions of dependencies. Otherwise we would maneuver ourselves in to the very situation (again), where we cannot release a plugin becuase it depends on pulpcore 3.*.*.dev. The goal is that the main branch can be released at any time without thinking.

ipanova: I see, this is a shift to the habits we had in place until now. The way you explained it here is very helpful, can you incorporate it "we only integrate new features with released versions of dependencies. Otherwise we would maneuver ourselves in to the very situation (again), where we cannot release a plugin becuase it depends on pulpcore 3...dev. The goal is that the main branch can be released at any time without thinking."

docs/contributing/git.rst
```diff
@@ -75,33 +75,29 @@ Putting this all together, the following is an example of a good commit message:
...
 
 For very similar reasons it can happen that you need changes to the base image used in the CI to
 spin up a new pulp container. In those cases you can build your own modified version of the image
-and push it to a container registry. Now you can specify the image to use in the last commit like::
+and push it to a container registry. Now you can specify the image to use in the last commit
+messaage like::
```
decko, line 96:
```suggestion
message like::
```

ipanova (approved):

PR #3708, opened 2023-04-04
Resolved purge vs export conflict

MichalPysik:
When an object could not be deleted, the purge task would fail. This has been reworked  so that the purge continues to delete other tasks and informs the user about tasks that could not be deleted.

fixes: #3741


lubosmj:
You can try logging the hrefs of tasks. We should definitely not raise an error and exit the task.

ggainey:
@MichalPysik I know we keep making things harder for you - but could we separate this into *two* PRs, one that relaxes the constraint and one that has the fix for "working even in the face of errors" piece? That would make it possible to backport just the "work even when a task can't be deleted".

As far as logging goes - we could have a close-cousin of _details_reporting, _errors_reporting, that would report "number of task-type Foo that were NOT removed due to constraints", called form inside the ProtectedError-except. Wouldn't need to know about/update totals_pb, since this is for things-that-DIDN'T-work.  Wdyt?

MichalPysik:
> @MichalPysik I know we keep making things harder for you - but could we separate this into _two_ PRs, one that relaxes the constraint and one that has the fix for "working even in the face of errors" piece? That would make it possible to backport just the "work even when a task can't be deleted".
> 
> As far as logging goes - we could have a close-cousin of _details_reporting, _errors_reporting, that would report "number of task-type Foo that were NOT removed due to constraints", called from inside the ProtectedError-except. Wouldn't need to know about/update totals_pb, since this is for things-that-DIDN'T-work. Wdyt?

Good call, I opened a new PR for the purge rework, so let's move the discussion there from now: https://github.com/pulp/pulpcore/pull/3709

ggainey:
> Good call, I opened a new PR for the purge rework, so let's move the discussion there from now: #3709

Awesome! Let's move the changelog/fixes to that PR as well, makes it easier to show the backported fix to downstream trackers. Also, loosening the constraint isn't actually visible to end-users the way the logged-errors will be.

ggainey (changes requested):
Let's move changelog/fixed-in to the new PR

mdellweg (approved):

dralley (approved):

ggainey (approved):

Issue #3530, opened 2023-01-27
export.task prevents task purge to succeed


mdellweg:
We probably want to classify the foreign key here
https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/exporter.py#L30
to be `SET_NULL`.

ggainey:
We should also need to teach task-purge to continue in the face of errors - just because one/several tasks can't be deleted, shouldn't fail the run. We don't know what models might do this exact thing in the future...

dralley:
+1 to both

MichalPysik:
https://github.com/pulp/pulpcore/pull/3708
@mdellweg @ggainey @dkliban

PR #3709, opened 2023-04-05
Reworked the task purge feature

MichalPysik:
When an object could not be deleted, the purge task would fail. This has been reworked  so that the purge continues to delete other tasks and informs the user about tasks that could not be deleted.

fixes: #3530

### DIFF
--- a/CHANGES/3530.bugfix
+++ b/CHANGES/3530.bugfix
@@ -1,0 +1,2 @@
+Task purge now continues when encountering a task that could not be deleted. The user is informed
+about any such tasks.
--- a/pulpcore/app/tasks/purge.py
+++ b/pulpcore/app/tasks/purge.py
@@ -1,6 +1,8 @@
 from gettext import gettext as _
+from logging import getLogger
 
 from django_currentuser.middleware import get_current_authenticated_user
+from django.db.models.deletion import ProtectedError
 from pulpcore.app.models import (
     ProgressReport,
     Task,
@@ -8,6 +10,8 @@ from pulpcore.app.models import (
 from pulpcore.app.role_util import get_objects_for_user
 from pulpcore.app.util import get_domain
 from pulpcore.constants import TASK_STATES
+
+log = getLogger(__name__)
 
 # Delete 1K at a time - better to use less memory, and take a little longer, with a utility
 # function like this.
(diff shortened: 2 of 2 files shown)


pulpcore/app/tasks/purge.py
```diff
@@ -106,10 +107,22 @@ def purge(finished_before, states):
...
+            try:
+                obj = Task.objects.get(pk=pk)
+                count, details = obj.delete()
+                units_deleted += count
+                _details_reporting(details_reports, details, totals_pb)
+            except ProtectedError as e:
+                # Object could not be deleted due to foreign key constraint.
+                # Print the details of the object and the exception message.
+                print(f"Object could not be deleted: {e}")
+                print(f"Details of the object: {e.protected_objects}")
```
mdellweg, line 125: It's either logging or adding to a progress report, but never print.

MichalPysik: It was somewhat of a placeholder, I'm currently working on real solution that incorporates both logging and a new progress report for objects that couldn't be deleted

pulpcore/app/tasks/purge.py
```diff
@@ -101,15 +105,39 @@ def purge(finished_before, states):
...
+                obj = Task.objects.get(pk=pk)
+                count, details = obj.delete()
+                units_deleted += count
+                _details_reporting(details_reports, details, totals_pb)
+            except ProtectedError as e:
+                # Object could not be deleted due to foreign key constraint.
+                # Log the details of the object.
+                error_pb.done += 1
+                pks_deleted.append(pk)
+                log.info(e)
```
MichalPysik, line 140: This should really be sufficient, I initially wanted to parse the exception manually but it's quite painful and the default text seems okay to me.

pulpcore/app/tasks/purge.py
```diff
@@ -101,15 +105,39 @@ def purge(finished_before, states):
     pb.save()
     details_reports[TASK_KEY] = pb
 
+    # Build and save a progress-report for objects that couldn't be deleted
+    error_pb = ProgressReport(
+        message=_("Failed to purge task-objects"),
```
mdellweg, line 110: "Tasks failed to purge" sounds better do me.

pulpcore/app/tasks/purge.py
mdellweg: Can we please keep the bulk delete and only do one-by-one as a fallback?

pulpcore/app/tasks/purge.py
```diff
@@ -101,15 +105,47 @@ def purge(finished_before, states):
...
+
+        # Try deleting the objects in bulk
+        try:
+            units_deleted, details = Task.objects.filter(
+                pk__in=candidate_pks[:DELETE_LIMIT]
+            ).delete()
+            _details_reporting(details_reports, details, totals_pb)
+        except ProtectedError:
+            # If there was at least one object that couldn't be deleted, then
+            # loop though the candidate objects and delete them one-by-one
```
mdellweg, line 136:
```suggestion
            # loop through the candidate objects and delete them one-by-one
```

pulpcore/app/tasks/purge.py
```diff
@@ -101,15 +105,47 @@ def purge(finished_before, states):
...
     units_deleted = 1
     # Until our query returns "No tasks deleted", add results into totals and Do It Again
     while units_deleted > 0:
-        units_deleted, details = Task.objects.filter(
-            pk__in=candidate_qs[:DELETE_LIMIT].values_list("pk", flat=True)
-        ).delete()
-        _details_reporting(details_reports, details, totals_pb)
+        # Get a list of candidate objects to delete and reset the counter
+        candidate_pks = candidate_qs.exclude(pk__in=pks_deleted).values_list("pk", flat=True)
+        units_deleted = 0
```
mdellweg, line 126: That logic does not work. If you failed to delete the whole batch, there may still be tasks left.
Can we rewrite this to use a batch iterator?

mdellweg (changes requested):

CHANGES/3530.bugfix
```diff
@@ -0,0 +1 @@
+Task purge now continues when encountering a task that could not be deleted. The user is informed about any such tasks.
```
mdellweg, line 1: Please line wrap this at 98 characters.

dralley (approved):

mdellweg (approved):

Issue #3711, opened 2023-04-06
Task listing is extremely unoptimized.


gerrod3:
**Version**
pulpcore <=3.23

**Describe the bug**
Listing out tasks causes 100s-1000s of queries to the database with a small set of tasks ~100. Many can be fixed by adding `prefetch_related` for `"child_tasks", "progress_reports"`. `created_resources` will need special optimization to prevent each resource from being their own query.

decko:
![image](https://user-images.githubusercontent.com/20204/230376379-1085e4ae-8c70-4cd1-b144-dfd1ae8e479c.png)
Can this image be of any help?

PR #3712, opened 2023-04-06
Optimize queries for task list endpoint

gerrod3:
fixes: #3711

`http /pulp/api/v3/tasks/?limit=100` with roughly ~200 tasks in the system (left optimized, right current):

task-optimization

`http /pulp/api/v3/tasks/sync_task_href/` (left optimized, right current):

task-individual-optimization

The number of queries we are currently doing is greatly dependent on the number and type of created resources made during the task. Each repository-version made during a task adds roughly 9-10 queries to the endpoint. For systems performing many syncs & uploads browsing the task endpoint would cause serious hammering of the database.

### DIFF
--- a/pulpcore/app/serializers/base.py
+++ b/pulpcore/app/serializers/base.py
@@ -3,12 +3,14 @@ from logging import getLogger
 import functools
 import re
 import traceback
+from collections import namedtuple
 from typing import List, TypedDict
 from urllib.parse import urljoin
 
 from django.conf import settings
 from django.core.validators import URLValidator
 from django.core.exceptions import ObjectDoesNotExist
+from django.urls.exceptions import NoReverseMatch
 from django.db import IntegrityError, transaction
 from django.db.models import Model
 from drf_queryfields.mixins import QueryFieldsMixin
@@ -20,7 +22,15 @@ from rest_framework_nested.relations import (
     NestedHyperlinkedRelatedField,
 )
 
-from pulpcore.app.models import Label, Task, TaskGroup, MasterModel
+from pulpcore.app.models import (
+    Label,
+    Task,
+    TaskGroup,
+    MasterModel,
+    GenericRelationModel,
+    Repository,
+    RepositoryVersion,
+)
 from pulpcore.app.util import (
     get_view_name_for_model,
     get_viewset_for_model,
@@ -145,6 +155,9 @@ class RelatedField(
     """
 
 
+PKOnlyObject = namedtuple("PKOnlyObject", ["pk"])
+
+
 class RelatedResourceField(RelatedField):
     """RelatedResourceField when relating a Resource object models.
 
@@ -155,7 +168,37 @@ class RelatedResourceField(RelatedField):
     Specific implementation requires the model to be defined in the Meta:.
     """
 
+    def repo_ver_url(self, repo_ver):
+        repo_model = get_model_for_pulp_type(repo_ver.repository.pulp_type, Repository)
+        view_name = get_view_name_for_model(repo_model, "detail")
+        obj = PKOnlyObject(pk=repo_ver.repository.pk)
+        repo_url = self.get_url(obj, view_name, request=None, format=None)
+        return f"{repo_url}versions/{repo_ver.number}/"
+
     def to_representation(self, data):
+        # Try to use optimized lookup to avoid DB if content_type has been already been fetched
+        if GenericRelationModel.content_type.is_cached(data):
+            model = data.content_type.model_class()
+            if issubclass(model, RepositoryVersion):
+                # Special case as we need to figure out the type of its repository to form url
+                repo_ver_mapping = self.context.get("repo_ver_mapping")
+                if repo_ver_mapping is not None:
+                    if repo_ver := repo_ver_mapping.get(data.object_id):
+                        return self.repo_ver_url(repo_ver)
+                    return None
+            else:
+                try:
+                    view_name = get_view_name_for_model(model, "detail")
+                except LookupError:
+                    pass
+                else:
+                    obj = PKOnlyObject(pk=data.object_id)
+                    try:
+                        return self.get_url(obj, view_name, request=None, format=None)
+                    except NoReverseMatch:
+                        pass
+
+        # Fallback onto normal lookup:
         # If the content object was deleted
         if data.content_object is None:
             return None
--- a/pulpcore/app/viewsets/task.py
+++ b/pulpcore/app/viewsets/task.py
@@ -1,5 +1,6 @@
 from gettext import gettext as _
 
+from django.db.models import Prefetch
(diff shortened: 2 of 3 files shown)


dralley:
@gerrod3 Is this superceded by the other PR or  is it independent?  The tests are passing, is it ready?

gerrod3:
> @gerrod3 Is this superceded by the other PR or is it independent? The tests are passing, is it ready?

It's independent. The utility function is copied between the two, but this optimization focuses on `RelatedResourceField`.

dralley:
@gerrod3 Should this remain in draft or is it good to go?

pulpcore/app/serializers/base.py
```diff
@@ -155,7 +164,37 @@ class RelatedResourceField(RelatedField):
     Specific implementation requires the model to be defined in the Meta:.
     """
 
+    def repo_ver_url(self, repo_ver):
+        repo_model = get_model_for_pulp_type(repo_ver.repository.pulp_type, Repository)
+        view_name = get_view_name_for_model(repo_model, "detail")
+        obj = Repository(pk=repo_ver.repository.pk)
```
mdellweg, line 170: Maybe a comment that this is a shallow representation.

mdellweg: Can we maybe just use a named tuple to avoid any other side-effects from subclassed `__init__`?

pulpcore/app/viewsets/task.py
```diff
@@ -135,6 +143,42 @@ class TaskViewSet(
...
+            self.action in ("list", "retrieve")
+            and isinstance(serializer, TaskSerializer)
+            and not getattr(self, "swagger_fake_view", False)
+        ):
+            task_list = args[0] if many else [args[0]]
+            created_resources = [cr for task in task_list for cr in task.created_resources.all()]
+            repo_ver_pks = [
+                cr.object_id
+                for cr in created_resources
+                if cr.content_type.model_class() == RepositoryVersion
```
mdellweg, line 160: `issubclass` should be a better choice thant `==`, i believe there is at least one subclass of repository versions out there.

dralley:
Do you see any clear opportunities for improving our data model for better performance?

For instance, I know we have issues with resources pointed to by `CreatedResource` being deleted, resulting in a list of `None` in the record, when the dynamic lookups fail later on.  That's not ideal, I wonder if storing a `JsonField` or `HListField` of href would be "better" or at least not worse, while providing some benefits.

pulpcore/app/serializers/base.py
```diff
@@ -155,7 +164,37 @@ class RelatedResourceField(RelatedField):
...
+                    if repo_ver := repo_ver_mapping.get(data.object_id):
+                        return self.repo_ver_url(repo_ver)
+                    return None
+            else:
+                try:
+                    view_name = get_view_name_for_model(model, "detail")
+                except LookupError:
+                    pass
+                else:
+                    obj = model(pk=data.object_id)
```
dralley, line 191: To what I said about the weird behavior for deleted CreatedResources (which appears to be implemented on line 199), how does this behave in that case?  I *think* it would probably fail, get caught, hit the `pass`, and go to line 199 and return None, but let's double check.

pulpcore/app/viewsets/task.py
```diff
@@ -135,6 +143,42 @@ class TaskViewSet(
...
     }
 
+    def get_serializer(self, *args, **kwargs):
+        """Set repo_ver_mapping in serializer context for optimized serialization."""
+        many = kwargs.get("many", False)
+        serializer = super().get_serializer(*args, **kwargs)
+        if (
+            self.action in ("list", "retrieve")
+            and isinstance(serializer, TaskSerializer)
+            and not getattr(self, "swagger_fake_view", False)
```
dralley, line 153: What is `swagger_fake_view`?  Can we comment this code a bit more

pulpcore/app/serializers/base.py
gerrod3: Actually it doesn't. The `None` behavior for deleted resources is only preserved with RepositoryVersions, since the `repo_ver_mapping` will not have the deleted version, but all the other resources will still be there in the record even after the resource is deleted with this new change.

gerrod3:
> Do you see any clear opportunities for improving our data model for better performance?
> 
> For instance, `CreatedResource` is one of the big issues here, and I know we also have issues with resources pointed to by `CreatedResource` being deleted after-the-fact resulting in a list of `None` in the record, when the dynamic lookups fail later on and the records themselves not being cleaned up. That's not ideal, and I wonder if storing a `JsonField` or `HListField` of href would be "better" or at least not worse, while providing some benefits.

It might have made sense to change this field to an `ArrayField` of hrefs if `CreatedResource` wasn't used so prominently in all our task code and plugins. I think there would need to be a big redesign (or at least lots of hacks) in order to simplify how we are storing `CreatedResource`s. Maybe in a future PR we could explore this, but I do not see an easy way forward.

One thing that is clear is that it is very hard to create an href for a repository-version. Maybe if we made it a `MasterModel` and had each plugin create proxy repo-version models so that each repo-ver record would have a pulp_type, then there wouldn't be a need for all the fancy pre-fetch logic in the queries. But this is also a big breaking change and for very little benefit.

dralley:
Yup not suggesting substantial architectural changes in this PR.  Just hoping we can make some changes to simplify things in the future

mdellweg:
Created resources works like
```python
CreatedResource(resource).save()
```
I'm pretty sure we can fudge something around that and issue deprecation warnings. But yes, probably not in this PR.

pulpcore/app/serializers/base.py
dralley: >(I thought the GenericRelation had a way of expensive cascade deleting, but i may be wrong...)

It is supposed to, but it's all based on runtime Python hooks and I don't think it's very reliable.

dralley: >Actually it doesn't. The None behavior for deleted resources is only preserved with RepositoryVersions, since the repo_ver_mapping will not have the deleted version, but all the other resources will still be there in the record even after the resource is deleted with this new change.

In practice I don't think that is something to really be concerned about, but, gross.  Maybe we can clean this up in 3.25

pulpcore/app/serializers/base.py
```diff
@@ -145,6 +155,9 @@ class RelatedField(
     """
 
 
+PKOnlyObject = namedtuple("PKOnlyObject", ["pk"])
```
mdellweg, line 158: :+1:

dralley (approved):
@gerrod3 Could you file an issue about the CreatedResources and summarize our discussion from here

gerrod3:
> @gerrod3 Could you file an issue about the CreatedResources and summarize our discussion from here

Filed issue: https://github.com/pulp/pulpcore/issues/3771

PR #3716, opened 2023-04-06
Attempt to keep the file extension intact when downloading files

### LINKED ISSUE #3715: Download machinery should attempt to keep the file extension of the downloaded URL intact, if one exists
**Version**

All

**Describe the bug**

Sometimes libraries will get upset when you pass into them a file without the file extension they expect for the type of file they are handling.  It is best to keep the file extension intact to avoid those kinds of issues.

See https://bugzilla.redhat.com/show_bug.cgi?id=2176870

There is a deeper issue there with the `file` / `libmagic` library, but until that is fixed and shipped the best we can do is help it out by keeping the original filenames.

I'm sure it will also help with debugging on occasion.

**To Reproduce**
See https://bugzilla.redhat.com/show_bug.cgi?id=2176870

**Additional context**
BZ https://bugzilla.redhat.com/show_bug.cgi?id=2176870

dralley:
closes #3715

### DIFF
--- a/pulpcore/download/base.py
+++ b/pulpcore/download/base.py
@@ -5,6 +5,7 @@ from collections import namedtuple
 import logging
 import os
 import tempfile
+from urllib.parse import urlsplit
 
 from pulpcore.app import pulp_hashlib
 from pulpcore.app.models import Artifact
@@ -116,7 +117,17 @@ class BaseDownloader:
         allowing plugin writers to instantiate many downloaders in memory.
         """
         if not self._writer:
-            self._writer = tempfile.NamedTemporaryFile(dir=".", delete=False)
+            filename = urlsplit(self.url).path.split("/")[-1]
+            # linux allows any character except NUL or / in a filename and has a length limit of
+            # 255. Making it urlencoding-aware would be nice, but not critical, because urlencoded
+            # paths should be OK
+            is_legal_filename = filename and (len(filename) <= 243)  # 255 - prefix length
+            # if the filename isn't legal then we just fall back to no suffix (random name)
+            suffix = "-" + filename if is_legal_filename else None
+            # write the file to the current working directory with a random prefix and the
+            # desired suffix. we always want the random prefix as it is possible to download
+            # the same filename from two different URLs, and the files may not be the same.
+            self._writer = tempfile.NamedTemporaryFile(dir=".", suffix=suffix, delete=False)
             self.path = self._writer.name
             self._digests = {n: pulp_hashlib.new(n) for n in Artifact.DIGEST_FIELDS}
             self._size = 0
--- a/CHANGES/3715.bugfix
+++ b/CHANGES/3715.bugfix
@@ -1,0 +1,1 @@
+Downloader will now attempt to keep the filename of the requested URL intact if one exists.


pulpcore/download/base.py
```diff
@@ -116,7 +116,14 @@ def _ensure_writer_has_open_file(self):
...
         if not self._writer:
-            self._writer = tempfile.NamedTemporaryFile(dir=".", delete=False)
+            filename = self.url.split("/")[-1]
+            # this should be safe even with URL encoding - linux allows any character except NUL or / in a filename
+            # making it urlencoding aware would be nice, but not critical
+            is_legal_filename = filename and len(filename <= 255)
+            # if the url ends with / then we just fall back to providing a random name
+            suffix = "-" + filename if is_legal_filename else None
+            # write the file to the current working directory with a random prefix and the desired suffix
+            self._writer = tempfile.NamedTemporaryFile(dir=".", suffix=suffix, delete=False)
```
dralley, line 126: Am I missing any corner cases?

ggainey: 255 bytes is max-filename-length in Linux - if you're adding a "-", shouldn't we be checking against 254?

dralley: Right, actually less than that because of the random prefix

ggainey (approved):
In the (pathological) case of the filename being > 243 bytes, we *might* still run into the problem of #3715, yes? But this improves our current situation enough to accept, while we hope for a real fix from the upstream library.

pulpcore/download/base.py
dralley: Hmmm what about trailing GET parameters?  Crap, I think that would break it.

ggainey: Maybe use urllib.parse(), and then operate on path?

dralley: Done

ggainey: There we go!

ggainey (approved):

decko (approved):

dralley:
@mdellweg / @ipanova Keeping your review requests as it would be good for someone who works on the container plugin to review also as your use cases are different

dralley:
Oh.  Ok

ggainey:
> @mdellweg / @ipanova Keeping your review requests as it would be good for someone who works on the container plugin to review also as your use cases are different

Ugh - we must have hit buttons at the exact same moment.

lubosmj:
@dralley, this should not affect the pulp_container workflows. There are no file extensions taken into consideration in registries.

Does the default limit equal to 255 characters per filename or absolute path? The latter could be a problem because we already have larger directory names for tasks' working directories.

ggainey:
> Does the default limit equal to 255 characters per filename or absolute path? The latter could be a problem because we already have larger directory names for tasks' working directories.

This is the **filename** limit - see https://www.linuxshelltips.com/find-limit-file-name-length-linux/ for some gory details. Note that a "directory" is a file for this purpose  (like everything else in Linux :) )  - so a single directory can't have a name longer than this max, but a directory-path can be...quite long.

Note also that this isn't a **Pulp** limitation, it's a **filesystem** limitation.

PR #3717, opened 2023-04-07
Add `pulp_href__in` and `pulp_id__in` query filters

gerrod3:
fixes: #2975

@daviddavis Would this work for you? These filters will be added to each endpoint.

### DIFF
--- a/pulpcore/filters.py
+++ b/pulpcore/filters.py
@@ -6,13 +6,15 @@ from django.db import models
 from django.forms.utils import ErrorList
 from django.urls import Resolver404, resolve
 from django_filters.constants import EMPTY_VALUES
-from django_filters.rest_framework import DjangoFilterBackend, filterset, filters
+from django_filters.rest_framework import DjangoFilterBackend, filterset, filters, BaseInFilter
 from django.core.exceptions import FieldDoesNotExist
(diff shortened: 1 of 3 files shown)


daviddavis:
The values for these fields, are they json lists or comma-separated? It looks like the former if my reading of the code is correct.

daviddavis:
Oh actually I just read the docs for `BaseInFilter` and it's comma separated which is probably easier for us.

This looks great, thanks.

ggainey (approved):

decko (approved):

pulpcore/filters.py
```diff
@@ -164,6 +176,11 @@ class BaseFilterSet(filterset.FilterSet):
         "ne": _("not equal to"),
     }
 
+    def filter_pulp_href(self, queryset, name, value):
+        # Convert each href to a pk
+        pks = [extract_pk(href) for href in value]
```
mdellweg, line 181: This looks like allowing anything that looks like an HREF containing a uuid to be used. Does it work with repository versions?

gerrod3: You are correct, this did not work with repository versions. Fix here: https://github.com/pulp/pulpcore/pull/3733

Issue #3714, opened 2023-04-06
Generic list endpoints are making unoptimized DB queries


gerrod3:
**Version**
pulpcore<=3.23

**Describe the bug**
Generic listing endpoints (`/content/`, `/distributions/`, `/repositories/`, etc..) are making unnecessary DB queries. Two culprits:

1. The queryset does not `select_related` related objects
2. `_DetailFieldMixin` preforms `cast` on each related object.

PR #3728, opened 2023-04-10
Optimized DB queries for list endpoints

gerrod3:
This ensures that when listing resources (for content, distributions, publications, remotes, and repositories) the number of DB queries does not increase as the number of objects increases. The optimization comes with two parts: one `select/prefetch_relating` the related fields & two using a utility function `get_model_for_pulp_type` (that I am also using in https://github.com/pulp/pulpcore/pull/3712/) to quickly find the appropriate viewset for the related object without having to call `cast`.

Here's an example of what these two optimizations achieve for `http /pulp/api/v3/content/?limit=100`:
(both optimizations, just the new utility optimization, original)
content-optimize

The `cast` calls for `DetailIdentityField` or `DetailRelatedField` were the biggest culprit for DB performance when doing large list calls.

fixes: #3714

### DIFF
--- a/pulpcore/app/models/base.py
+++ b/pulpcore/app/models/base.py
@@ -7,6 +7,7 @@ from django.db import models
 from django.db.models import options
 from django.db.models.base import ModelBase
 from django_lifecycle import LifecycleModel
+from functools import lru_cache
 
 from pulpcore.app.loggers import deprecation_logger
 
@@ -86,6 +87,11 @@ class BaseModel(LifecycleModel):
 
     class Meta:
         abstract = True
+
+    @classmethod
+    @lru_cache
+    def get_field_names(cls):
+        return [f.name for f in cls._meta.fields]
 
     def __str__(self):
         try:
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -18,7 +18,15 @@ from django.urls import reverse
 from django_lifecycle import AFTER_UPDATE, BEFORE_DELETE, hook
 from rest_framework.exceptions import APIException
 
-from pulpcore.app.util import batch_qs, get_url, get_view_name_for_model, get_domain_pk, cache_key
+from pulpcore.app.util import (
+    batch_qs,
+    get_url,
+    get_view_name_for_model,
+    get_domain,
+    get_domain_pk,
+    cache_key,
+    get_model_for_pulp_type,
+)
 from pulpcore.constants import ALL_KNOWN_CONTENT_CHECKSUMS
 from pulpcore.download.factory import DownloaderFactory
 from pulpcore.exceptions import ResourceImmutableError
@@ -1142,13 +1150,15 @@ class RepositoryVersionContentDetails(models.Model):
         Returns:
             dict: {<pulp_type>: <url>}
         """
-        repository = self.repository_version.repository.cast()
-        ctypes = {c.get_pulp_type(): c for c in repository.CONTENT_TYPES}
+        repository_model = get_model_for_pulp_type(
+            self.repository_version.repository.pulp_type, Repository
+        )
+        ctypes = {c.get_pulp_type(): c for c in repository_model.CONTENT_TYPES}
         ctype_model = ctypes[self.content_type]
         ctype_view = get_view_name_for_model(ctype_model, "list")
         kwargs = {}
         if settings.DOMAIN_ENABLED:
-            kwargs["pulp_domain"] = repository.pulp_domain.name
+            kwargs["pulp_domain"] = get_domain().name
         try:
             ctype_url = reverse(ctype_view, kwargs=kwargs)
         except django.urls.exceptions.NoReverseMatch:
@@ -1156,12 +1166,12 @@ class RepositoryVersionContentDetails(models.Model):
             # There's nothing we can do here, except to skip it.
             return
 
-        repository_view = get_view_name_for_model(repository.__class__, "list")
+        repository_view = get_view_name_for_model(repository_model, "list")
 
         repository_url = reverse(repository_view, kwargs=kwargs)
         rv_href = (
             repository_url
-            + str(repository.pk)
+            + str(self.repository_version.repository_id)
             + "/versions/{version}/".format(version=self.repository_version.number)
         )
         if self.count_type == self.ADDED:
--- a/pulpcore/app/serializers/base.py
+++ b/pulpcore/app/serializers/base.py
@@ -20,12 +20,13 @@ from rest_framework_nested.relations import (
     NestedHyperlinkedRelatedField,
 )
 
-from pulpcore.app.models import Label, Task, TaskGroup
+from pulpcore.app.models import Label, Task, TaskGroup, MasterModel
 from pulpcore.app.util import (
     get_view_name_for_model,
     get_viewset_for_model,
     get_request_without_query_params,
     get_domain,
+    get_model_for_pulp_type,
 )
 
 
@@ -102,18 +103,18 @@ class _DetailFieldMixin(HrefFieldMixin):
         super().__init__(view_name, **kwargs)
 
     def _view_name(self, obj):
-        # this is probably memoizeable based on the model class if we want to get cachey
-        try:
-            obj = obj.cast()
-        except AttributeError:
-            # The normal message that comes up here is unhelpful, so do like other DRF
-            # fails do and be a little more helpful in the exception message.
-            msg = (
-                'Expected a detail model instance, not {}. Do you need to add "many=True" to '
-                "this field definition in its serializer?"
-            ).format(type(obj))
-            raise ValueError(msg)
-        return get_view_name_for_model(obj, "detail")
+        if isinstance(obj, MasterModel):
+            return get_view_name_for_model(
+                get_model_for_pulp_type(obj.pulp_type, obj._meta.model), "detail"
+            )
+
+        # The normal message that comes up here is unhelpful, so do like other DRF
+        # fails do and be a little more helpful in the exception message.
+        msg = (
+            'Expected a detail model instance, not {}. Do you need to add "many=True" to '
+            "this field definition in its serializer?"
+        ).format(type(obj))
+        raise ValueError(msg)
 
     def get_url(self, obj, view_name, request, *args, **kwargs):
         # ignore the passed in view name and return the url to the cast unit, not the generic unit
--- a/pulpcore/app/serializers/fields.py
+++ b/pulpcore/app/serializers/fields.py
@@ -246,28 +246,6 @@ class LatestVersionField(RepositoryVersionRelatedField):
         kwargs["read_only"] = True
         super().__init__(*args, **kwargs)
 
-    def get_url(self, obj, view_name, request, format):
-        """
-        Returns the URL for the appropriate object. Overrides the DRF method to change how the
-        object is found.
-
-        Args:
-            obj (list of :class:`pulpcore.app.models.RepositoryVersion`): The versions of the
-                current viewset's repository.
-            view_name (str): The name of the view that should be used.
-            request (rest_framework.request.Request): the current HTTP request being handled
-            format: undocumented by DRF. ???
-
-        Returns:
-            str: the URL corresponding to the latest version of the current repository. If there
-                are no versions, returns None
-        """
-        try:
-            version = obj.complete().latest()
-        except obj.model.DoesNotExist:
-            return None
-        return super().get_url(version, view_name, request, format)
-
     def get_attribute(self, instance):
         """
         Args:
@@ -275,10 +253,14 @@ class LatestVersionField(RepositoryVersionRelatedField):
                 current ViewSet.
 
         Returns:
-            list of :class:`pulpcore.app.models.RepositoryVersion`
-
+            instance :class:`pulpcore.app.models.RepositoryVersion`
         """
-        return instance.versions
+        if hasattr(instance, "latest_version_number"):
+            # Return a shallow object sufficient to create the HREF.
+            return models.RepositoryVersion(
+                repository=instance, number=instance.latest_version_number
+            )
+        return instance.latest_version()
 
 
 class BaseURLField(serializers.CharField):
--- a/pulpcore/app/util.py
+++ b/pulpcore/app/util.py
@@ -177,7 +177,32 @@ def get_view_name_for_model(model_obj, view_action):
         for pattern, registered_viewset, base_name in router.registry:
             if registered_viewset is viewset:
                 return "-".join((base_name, view_action))
-    raise LookupError("view not found")
+    raise LookupError(f"{viewset} view not found")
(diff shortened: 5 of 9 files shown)


dralley:
HUGE +1 and kudos for this work!  This is a huge improvement, enough to probably have user-visible responsiveness impacts.

I'm going to mark it as drafted until the tests are passing though

pulpcore/app/util.py
```diff
@@ -177,7 +179,31 @@ def get_view_name_for_model(model_obj, view_action):
         for pattern, registered_viewset, base_name in router.registry:
             if registered_viewset is viewset:
                 return "-".join((base_name, view_action))
-    raise LookupError("view not found")
+    raise LookupError(f"{viewset} view not found")
+
+
+def get_model_for_pulp_type(pulp_type, model):
```
dralley, line 185: Have you considered using `@functools.lru_cache(typed=True)` (or something like it) instead of implementing the cache manually?  I've used it in the past, it works well.

https://github.com/rsinger86/django-lifecycle/pull/64/files#diff-ccd0462c3332b1460bdeefa901260186d8b9de888e6acd16c6d63a0fbbbe2c27R163
https://docs.python.org/3/library/functools.html#functools.lru_cache

pulpcore/app/viewsets/content.py
```diff
@@ -111,6 +111,13 @@ class BaseContentViewSet(NamedModelViewSet):
     queryset = Content.objects.all().exclude(pulp_type=PublishedMetadata.get_pulp_type())
     serializer_class = MultipleArtifactContentSerializer
 
+    def get_queryset(self):
+        """Apply optimizations for list endpoint."""
+        qs = super().get_queryset()
+        if getattr(self, "action", "") == "list":
+            qs = qs.prefetch_related("contentartifact_set")
+        return qs
```
dralley, line 119: Do you know if the limit (pagination) gets applied before or after this?  In other words does it prefetch `contentartifact_set` and throw them away or does it throw away objects and then prefetch `contentartifact_set`

pulpcore/app/viewsets/publication.py
```diff
@@ -84,12 +85,39 @@ class Meta:
...
+
     endpoint_name = "publications"
-    queryset = Publication.objects.all()
+    queryset = Publication.objects.exclude(complete=False)
     serializer_class = PublicationSerializer
     filterset_class = PublicationFilter
+    ordering = ("-pulp_created",)
 
+    def get_queryset(self):
+        """Apply optimizations for list endpoint."""
```
dralley, line 100: It would be good if we could describe in comments what code was responsible for these queries here and in other places, if you know that information.  While it's still fresh.

For instance up above I assume the `ContentArtifact` were queried via the artifacts being serialized alongside the content fields.

Down below it's not as clear what features/code is querying this info.

dralley, line 100: It would be good if we could describe in comments what code was responsible for these queries here and in other places, if you know that information.  While it's still fresh.

For instance up above I assume the `ContentArtifact` were queried via the artifacts being serialized alongside the content fields.

Down below it's not as clear what features/code is querying this info.

pulpcore/app/util.py
gerrod3: Maybe, I would need to break the function down into two since `model` accepts a model class or instance, but it could work.

pulpcore/app/viewsets/content.py
gerrod3: `prefetch` only gets executed after the queryset is evaluated, so it only fetches after the limit is already applied.

pulpcore/app/util.py
gerrod3: I have updated it to use `lru_cache`.

pulpcore/app/viewsets/publication.py
```diff
@@ -350,6 +382,20 @@ class BaseDistributionViewSet(NamedModelViewSet):
...
+        qs = super().get_queryset()
+        if getattr(self, "action", "") == "list":
+            # Fetch info for DetailRelatedFields: repository, content_guard, remote, publication
+            qs = qs.select_related("repository", "content_guard", "remote", "publication").only(
+                *self.queryset.model.get_field_names(),
+                "repository__pulp_type",
+                "content_guard__pulp_type",
+                "remote__pulp_type",
+                "publication__pulp_type",
```
mdellweg, line 395: Can you use https://docs.djangoproject.com/en/3.2/ref/models/indexes/#include to speed up these related lookups?

pulpcore/app/models/repository.py
```diff
@@ -1142,26 +1150,30 @@ def content_href(self):
...
+            get_model_for_pulp_type(self.repository_version.repository.pulp_type, Repository)
+            or self.repository_version.repository.cast()
+        )
+        ctypes = {c.get_pulp_type(): c for c in repository_model.CONTENT_TYPES}
         ctype_model = ctypes[self.content_type]
         ctype_view = get_view_name_for_model(ctype_model, "list")
         kwargs = {}
         if settings.DOMAIN_ENABLED:
-            kwargs["pulp_domain"] = repository.pulp_domain.name
+            kwargs["pulp_domain"] = get_domain().name
```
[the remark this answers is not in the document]
gerrod3: Yes it's the same domain. I don't understand what you mean by domain filters.

pulpcore/app/serializers/fields.py
```diff
@@ -278,7 +256,11 @@ def get_attribute(self, instance):
             list of :class:`pulpcore.app.models.RepositoryVersion`
```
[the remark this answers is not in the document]
gerrod3: It is only ever used as a field in the RepositorySerializer, it is not exposed in the plugin api.

pulpcore/app/models/repository.py
```diff
@@ -1142,26 +1150,30 @@ def content_href(self):
         Returns:
             dict: {<pulp_type>: <url>}
         """
-        repository = self.repository_version.repository.cast()
-        ctypes = {c.get_pulp_type(): c for c in repository.CONTENT_TYPES}
+        # Try to use cached lookup first to avoid calling cast
+        repository_model = (
+            get_model_for_pulp_type(self.repository_version.repository.pulp_type, Repository)
```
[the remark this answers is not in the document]
mdellweg: Would that be a programming error? Then we should let it fail. (And my hunch is Yes.)

pulpcore/app/models/repository.py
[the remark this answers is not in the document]
mdellweg: Whether we use model managers that are already domain aware, to protect us on the lowest level from selecting stuff across domains. But this question is probably not related to this PR in any way.

pulpcore/app/serializers/fields.py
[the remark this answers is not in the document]
mdellweg: Then we are good to change it. :fireworks:

pulpcore/app/viewsets/publication.py
gerrod3: Not sure I did it correctly but I added
 ```
indexes = [models.Index(name="%(app_label)s_%(class)s_type", fields=["pulp_id"], include=["pulp_type"])]
```
 to `MasterModel.Meta` (and updated the other sub-class MasterModels to inherit that Meta). After running the migration and testing the endpoint I saw no difference in the query plan as reported by django-silky, so I don't think the index was being queried.

File: pulpcore/app/models/repository.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (gerrod3) [reply pulpcore/app/models/repository.py]: Yeah you are correct, I changed it so it only calls the util method. If that returns None then that means the plugin writer has made a mistake somewhere in their code.

File: pulpcore/app/viewsets/repository.py [hunk 8]
```diff
@@ -64,6 +64,18 @@ class BaseRepositoryViewSet(NamedModelViewSet):
...
 
+    def get_queryset(self):
+        """Apply optimizations for list endpoint."""
+        qs = super().get_queryset()
+        if getattr(self, "action", "") == "list":
+            # Fetch info for remote (DetailRelatedField) and latest_version (LatestVersionField)
+            qs = (
+                qs.select_related("remote")
+                .only(*self.queryset.model.get_field_names(), "remote__pulp_type")
+                .annotate(latest_version_number=Max("versions__number"))
```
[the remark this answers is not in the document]
gerrod3: Added a filter for complete versions.

pulpcore/app/models/repository.py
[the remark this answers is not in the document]
gerrod3: No I couldn't find an elegant solution to adding a domain filter on the model managers.

pulpcore/app/viewsets/publication.py
dralley: `pulp_type` is always indexed https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/base.py#L189

pulpcore/app/serializers/base.py
```diff
@@ -102,7 +103,11 @@ def __init__(self, view_name=None, view_name_pattern=None, **kwargs):
         super().__init__(view_name, **kwargs)
 
     def _view_name(self, obj):
-        # this is probably memoizeable based on the model class if we want to get cachey
+        # Try using cached lookup based on object's pulp_type
+        if hasattr(obj, "pulp_type"):
```
dralley, line 107: As this is a `DetailFieldMixin`, I don't think there would be any types for which it doesn't have a `pulp_type`

pulpcore/app/models/repository.py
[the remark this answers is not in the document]
mdellweg: Anyway, not the scope here.

pulpcore/app/viewsets/publication.py
mdellweg: Good point. In that case we may already have it optimized.

pulpcore/app/models/base.py
```diff
@@ -87,6 +87,10 @@ class BaseModel(LifecycleModel):
     class Meta:
         abstract = True
 
+    @classmethod
+    def get_field_names(cls):
+        return [f.name for f in cls._meta.fields]
```
mdellweg, line 92: Is it worth caching that too?

pulpcore/app/serializers/base.py
mdellweg: So basically this function should be

```python
def _view_name(self, obj):
    return get_view_name_for_model(get_model_for_pulp_type(obj.pulp_type, obj._meta.model), "detail")
```

And if it fails it's a programming error...

pulpcore/app/serializers/fields.py
```diff
@@ -246,39 +246,20 @@ def __init__(self, *args, **kwargs):
...
                 current ViewSet.
 
         Returns:
-            list of :class:`pulpcore.app.models.RepositoryVersion`
-
+            instance :class:`pulpcore.app.models.RepositoryVersion`
         """
-        return instance.versions
+        if hasattr(instance, "latest_version_number"):
+            return models.RepositoryVersion(
```
mdellweg, line 259:
```suggestion
            # Return a shallow object sufficient to create the HREF.
            return models.RepositoryVersion(
```

pulpcore/app/util.py
mdellweg: Can you add to the comment that "model" accepts a model or an object?

pulpcore/app/viewsets/repository.py
```diff
@@ -64,6 +64,20 @@ class BaseRepositoryViewSet(NamedModelViewSet):
...
+    def get_queryset(self):
+        """Apply optimizations for list endpoint."""
+        qs = super().get_queryset()
+        if getattr(self, "action", "") == "list":
+            # Fetch info for remote (DetailRelatedField) and latest_version (LatestVersionField)
+            qs = (
+                qs.select_related("remote")
+                .only(*self.queryset.model.get_field_names(), "remote__pulp_type")
+                .annotate(
+                    latest_version_number=Max("versions__number", filter=Q(versions__complete=True))
```
mdellweg, line 76: :+1:

pulpcore/app/viewsets/repository.py
```diff
@@ -187,8 +201,32 @@ class Meta:
...
+    """A mixin to hold the shared get_queryset logic used by RepositoryVersionViewSets."""
+
+    def get_queryset(self):
+        qs = super().get_queryset()
+        if getattr(self, "action", "") == "list":
+            # Fetch info for repository (DetailRelatedField),
+            # base_version (RepositoryVersionRelatedField), and
+            # counts (ContentSummarySerializer)
+            qs = (
+                qs.select_related("repository", "base_version__repository")
```
mdellweg, line 214: What is the base_version__repository fetching? Do we need only-statements for it?

pulpcore/app/serializers/base.py
dralley: I think so.  90% sure.

pulpcore/app/models/base.py
dralley: I was thinking about that but IMO it depends on whether there is actually any work going on behind the scenes to generate `cls._meta.fields` or if it is already a list or cached internally.

dralley: It's not like it would *hurt*, though.

pulpcore/app/viewsets/repository.py
gerrod3: It is fetching the base_version and its repository (if they exist). Interestingly, django's `only` causes the shared field names of the two RepositoryVersion objects to always pull in the same fields for both objects. e.g. `.only(*RepositoryVersion.get_field_names(), "repository__pulp_type")` produces the same query as `.only(*RepositoryVersion.get_field_names(), "repository__pulp_type", "base_version__number", "base_version__repository__pulp_type")`.

pulpcore/app/serializers/base.py
gerrod3: I removed the fallback, we are now only using the cached util.

dralley (approved):

pulpcore/app/viewsets/publication.py
```diff
@@ -84,12 +85,42 @@ class Meta:
...
 
-class ListPublicationViewSet(NamedModelViewSet, mixins.ListModelMixin):
+class BasePublicationViewSet(NamedModelViewSet):
+    """
+    A base class for any publication viewset.
+    """
+
     endpoint_name = "publications"
-    queryset = Publication.objects.all()
+    queryset = Publication.objects.exclude(complete=False)
```
mdellweg, line 94: Can this be `.filter(complete=True)`?

mdellweg (approved):

pulpcore/app/viewsets/publication.py
dralley: +1

dralley:
@gerrod3 Ready to merge?

PR #3731, opened 2023-04-11
Change the order of authentication classes

### LINKED ISSUE #3730: The order of authentication classes causes Pulp to return 403 instead of 401 on failed authentication
https://github.com/pulp/pulpcore/blob/50bdd1ba0d2ce7ae4d6eef15a2dd7c569d49c03a/pulpcore/app/settings.py#L150

In pulp_container, we have a view that should handle insufficient, invalid, or valid credentials provided by users. If the users pass invalid credentials to that view, Pulp returns 403 instead of 401.

Changing the order of authentication classes could resolve the problem according to https://www.django-rest-framework.org/api-guide/authentication/#unauthorized-and-forbidden-responses.

> The first authentication class set on the view is used when determining the type of response.

lubosmj:
closes #3730

### DIFF
--- a/CHANGES/3730.bugfix
+++ b/CHANGES/3730.bugfix
@@ -1,0 +1,2 @@
+Fixed returned error codes for unauthenticated users. Pulp now returns HTTP 401 errors instead of
+HTTP 403 errors which were default return types for ``SessionAuthentication``.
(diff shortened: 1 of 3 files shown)


dralley (approved):

lubosmj:
The change is visible from this test:

https://github.com/pulp/pulp_container/blob/6b5c6b4caaf54e95f3257ecd2387351d1f0cd667/pulp_container/tests/functional/api/test_token_authentication.py#L142

https://github.com/pulp/pulp_container/actions/runs/4669255842/jobs/8267470311#step:10:1961

lubosmj:
re: https://github.com/pulp/pulp_container/pull/1257

mdellweg (approved):

dralley:
I don't think so.  This seems more standard compliant

ipanova:
Was this backported to older branches?

PR #3735, opened 2023-04-12
Bump Django to 4.2

### LINKED ISSUE #3660: Bumping Django version to 4.2 LTS
This is issue is to be used to track our migration to Django 4 branch, 4.2 version specifically when it is released.

We have an experimental branch just to understand the impact and possible solutions for the migration [here](https://github.com/pulp/pulpcore/pull/3618)

mdellweg:
fixes #3660


dralley:
@mdellweg semi-related question.

I know we cannot squash the pulpcore migrations (because plugin migrations rely on them currently) but is there anything preventing us from squashing the plugin migrations?

mdellweg:
> @mdellweg semi-related question.
> 
> I know we cannot squash the pulpcore migrations (because plugin migrations rely on them currently) but is there anything preventing us from squashing the plugin migrations?

The one thing that comes to my mind is that you loose the ability to upgrade from everywhere.

dralley:
You can get rid of nearly? all calls to `sync_to_async` and `sync_to_async_iterator` as Django 4.2 has official support for async queries https://docs.djangoproject.com/en/4.2/topics/db/queries/#async-queries

decko:
> You can get rid of nearly? all calls to `sync_to_async` and `sync_to_async_iterator` as Django 4.2 has official support for async queries https://docs.djangoproject.com/en/4.2/topics/db/queries/#async-queries
> 
> The content app is the primary place where we've got those.

Yet, won't be better to just bump Django version leaving pulp code with minimum possible changes and address those improvements in a future PR? With lots of changes would be too hard to track any issue we could have beyond what our tests covers.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -485,16 +485,14 @@ async def run(self):
...
+                    RemoteArtifact.objects.acs()
+                    .filter(batch_query)
+                    .only("url", "remote")
+                    .select_related("remote")
                 )
                 existing_ras_dict = dict()
-                for ra in existing_ras:
+                async for ra in existing_ras:
                     for c_type in Artifact.COMMON_DIGEST_FIELDS:
                         checksum = await sync_to_async(getattr)(ra, c_type)
```
dralley, line 497: Are you able to get rid of this one also?

Is it possible to use .iterator() on line 495?

dralley: Yes, just a suggestion.  As it's still in draft it's hard to tell what the intended scope of the PR is

mdellweg: It's mainly in draft so it does not get merged before releasing 3.24. XD
And i expect some rebasing until then. Redoing the low hanging async fruits, i already realized that some of them were not so easy to pick... At this point I just want to limit the scope creep and reduce the risk of this change.

pulpcore/app/tasks/analytics.py
```diff
@@ -61,8 +61,7 @@ async def _postgresql_version(analytics):
 
 async def _num_hosts(qs):
     hosts = set()
-    items = await sync_to_async(list)(qs.all())
-    for item in items:
+    async for item in qs.all():
```
decko, line 64: :star_struck:

lubosmj (approved):
The tests are green, so I do not see any problems at the moment.

Have you considered looking at 4.0-4.2 fixers noted in https://github.com/adamchainz/django-upgrade#django-40? I am particularly interested in the `STORAGES` setting. Is this PR the right place to introduce the changes to the settings file? Then, we need to update the relevant documentation as well.

mdellweg:
Let's test once more with domain enabled.

pulpcore/app/tasks/analytics.py
```diff
@@ -76,34 +75,32 @@ async def _versions_data(analytics):
...
 async def _system_id(analytics):
-    system_id_obj = await sync_to_async(SystemID.objects.get)()
+    system_id_obj = await SystemID.objects.aget()
     analytics.system_id = str(system_id_obj.pk)
 
 
 async def _rbac_stats(analytics):
-    analytics.rbac_stats.users = await sync_to_async(User.objects.count)()
-    analytics.rbac_stats.groups = await sync_to_async(Group.objects.count)()
+    analytics.rbac_stats.users = await User.objects.acount()
```
dralley, line 94: Not required for your PR, just as a comment:  we're not actually taking advantage of async much here, if we were, we could submit all queries at once and await them together.

mdellweg:
> The tests are green, so I do not see any problems at the moment.
> 
> Have you considered looking at 4.0-4.2 fixers noted in https://github.com/adamchainz/django-upgrade#django-40? I am particularly interested in the `STORAGES` setting. Is this PR the right place to introduce the changes to the settings file? Then, we need to update the relevant documentation as well.

You are referring to this:
https://docs.djangoproject.com/en/4.2/releases/4.2/#custom-file-storages
The settings we currently use will be deprecated with 4.2 but still available. So we may be safe changing that documentation post 3.25.

pulpcore/content/handler.py
```diff
@@ -536,38 +541,29 @@ async def _match_and_stream(self, path, request):
...
-                        repo_version = publication.repository_version
-                    except ObjectDoesNotExist:
-                        pass
-
-                if not repo_version:
-                    repo_version = repository.latest_version()
-
-            await sync_to_async(get_latest_publication_or_version_blocking)()
+            if not repo_version:
+                repo_version = await sync_to_async(repository.latest_version)()
```
dralley, line 559: Should we provide an async equivalent of this?

pulpcore/content/handler.py
```diff
@@ -688,25 +671,18 @@ def get_contentartifact_blocking():
                     )
 
         if distro.remote:
-
-            def cast_remote_blocking():
-                return distro.remote.cast()
-
-            remote = await sync_to_async(cast_remote_blocking)()
+            remote = await sync_to_async(distro.remote.cast)()
```
dralley, line 674: Same

dralley (approved):
None of my comments should be considered blockers, this looks fine

pulpcore/content/handler.py
mdellweg: Sounds like a fine idea. I'll write an issue.

Issue #3408, opened 2022-11-16
Add "type" filter to generic list endpoints (repo, distro, remotes)


daviddavis:
We pull most of our data through the generic list endpoints (e.g. `/pulp/api/v3/repositories/`) and we'd like to be able to filter the results by type (rpm, deb, etc). 

In theory we could query the plugin specific list endpoints but then our code would have to deal with multiple endpoints (one for each type) and potential issues with different query params, response fields, etc

A `type` or `types` filter would be great.

PR #3758, opened 2023-04-21
Add pulp_type filter for generic list endpoints

gerrod3:
fixes: #3408


gerrod3:
`pulp_type` isn't really exposed to the user, but I set up the filter so that the help-text message in the generated docs will give you all the valid values:
pulp_type__in

@daviddavis Does this look good to you?

daviddavis:
Looks great!

Issue #3538, opened 2023-01-30
As a user, I can "hide" distributions


daviddavis:
This would prevent the distribution from showing up in the dir listings from the content app. So for example if you had distros a, b, c and b is hidden, the directory listing would have:

```
a/
c/
```

This is handy for example if you have a repo that is deprecated but you still want users/clients to be able to use it (until it's finally removed).

mdellweg:
What are the zero-downtime consequences of this change?

Adding a "hidden" field that can be null should be fine. Providing a default value may need more work.

Issue #3180, opened 2022-09-08
content app should issue 404 if content guard denied access


mdellweg:
> (One could argue that should be a 404 rather than a 403.)

For obscuration purposes I presume?  (If you were spraying requests you could theoretically uncover which protected directories exist based on getting not-found or access-denied).

_Originally posted by @dralley in https://github.com/pulp/pulpcore/issues/3061#issuecomment-1228812641_

ipanova:
https://github.com/pulp/pulpcore/issues/3061#issuecomment-1241742479

ggainey:
> [#3061 (comment)](https://github.com/pulp/pulpcore/issues/3061#issuecomment-1241742479)
> 
> > > > For obscuration purposes I presume? (If you were spraying requests you could theoretically uncover which protected directories exist based on getting not-found or access-denied).
> > > 
> > > 
> > > Correct.
> > 
> > 
> > I agree that this may be a valid concern, but i think that would be a separate issue. Should we have one filed?
> 
> I think we should distinguish between 404 and 403, in case user presents invalid client cert he should receive 403, imo.

Invalid-cert should always short-circuit to 403, concur. If you have a valid cert, but it doesn't allow access to the path you're requesting, I think that should *also* be a 403 - **even if that path doesn't exist** . If your cert doesn't give you access to a Thing, "you're not allowed there" is all the info you should get back. If your cert does allow access to a distribution-path that doesn't exist - **then** you should get a 404. "You're allowed there, but there's nobody home, sorry"

That's my 0.02 anyway

dralley:
Worth noting, when I make a request against the RH CDN without a certificate, I get 401, not 403.

ggainey:
> Worth noting, when I make a request against the RH CDN without a certificate, I get 401, not 403.

Correct according to my reading of the standard - "The HyperText Transfer Protocol (HTTP) 401 Unauthorized response status code indicates that the client request has not been completed because it lacks valid authentication credentials for the requested resource. " So maybe 401 for invalid-cert as well? 403 for valid-cert-but-not-that-path, 404 for valid-cert-allowed-path-doesn't-exist?

ipanova:
Given that `HIDE_GUARDED_DISTRIBUTIONS=True` is enabled, maybe we should have:
401 for non existent path
401 for no cert
403 for invalid cert (for guarded content)
403 valid-cert-but-not-that-path ( for guarded content)
[...]

PR #3768, opened 2023-04-25
Add `hidden` field to Distribution

gerrod3:
fixes: #3538

I think this should comply with our new zero-downtime migration policy assuming they run migrations first before restarting services with new code.


gerrod3:
Also, how should I go about writing a test for this? Seems like a waste to have to create a separate pulp_file PR to test this feature...

pulpcore/app/serializers/publication.py
```diff
@@ -191,13 +191,17 @@ class DistributionSerializer(ModelSerializer):
         queryset=models.Repository.objects.all(),
         allow_null=True,
     )
+    hidden = serializers.BooleanField(
```
ipanova, line 194: This different level of hiding distribution becomes "interesting". We already provide the option to hide the distribution that are content_guard protected. I understand that this is yet another usecase, just we keep expanding the matrix

ipanova:
> Also, how should I go about writing a test for this? Seems like a waste to have to create a separate pulp_file PR to test this feature...

we could at least assert that 200 is returned whether the distribution is or not hidden ( non protected)https://github.com/pulp/pulpcore/issues/3180#issuecomment-1523055699

gerrod3:
> > Also, how should I go about writing a test for this? Seems like a waste to have to create a separate pulp_file PR to test this feature...
> 
> we could at least assert that 200 is returned whether the distribution is or not hidden ( non protected)[#3180 (comment)](https://github.com/pulp/pulpcore/issues/3180#issuecomment-1523055699)

That's not what I meant. The logic of the test isn't a problem, it's where I should put the test. Technically none of the pulpcore tests use any pulp_file fixtures, but we don't have a pulpcore distribution that I can use and I am still unsure of having the test live separately in another repository, especially with our new testing against release policy. It is making testing new features added to pulpcore difficult.

mdellweg:
> > > Also, how should I go about writing a test for this? Seems like a waste to have to create a separate pulp_file PR to test this feature...
> > 
> > 
> > we could at least assert that 200 is returned whether the distribution is or not hidden ( non protected)[#3180 (comment)](https://github.com/pulp/pulpcore/issues/3180#issuecomment-1523055699)
> 
> That's not what I meant. The logic of the test isn't a problem, it's where I should put the test. Technically none of the pulpcore tests use any pulp_file fixtures, but we don't have a pulpcore distribution that I can use and I am still unsure of having the test live separately in another repository, especially with our new testing against release policy. It is making testing new features added to pulpcore difficult.

That is a good question, and maybe it is time to say that this rule does not apply anymore. We do depend on pulp_file for testing, so we can rely on whatever pulp_file has released. This could be just stable enough to invalidate the original reason to move all the tests over there. It's still not so easy if you have a feature that involves changes in both. Then you can only assert that the pulpcore change does not regress on existing features.

pulpcore/app/migrations/0104_distribution_hidden.py
```diff
@@ -0,0 +1,32 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0103_alter_export_task'),
+    ]
+
+    operations = [
+        # Special SQL to handle zero-downtime migrations. Django handles 'default' field through
+        # python code, so we need to ensure the default is set in the DB to allow old code to still
+        # work after migrating.
```
mdellweg, line 20: This is a very nice example for the toolbox we may need for ZDU. But can we not in this example assume that Null will evaluate to False and have a reqular Django migration?

gerrod3: We could but then it is possible during an upgrade that some new distributions created will have `null` while others will have `false` for this field. After the upgrade each new distribution will have `false`. Functionaly it would be no different, but it could be confusing for users why some distributions have `null` vs `false` (they could always update the offending ones to `false`, but that is extra work).

mdellweg: I guess we could normalize that in the serializer. And then in a later migration (that depends on all components knowing the hidden field) we update all the records and change the field to have a default and not allow null.

gerrod3: That is another approach that could work. I was going to do that initially but this seemed easier (to write). Would you like me to change it?

mdellweg: I'm not willing to decide alone. But my feeling says that we should keep the database consistent with Django's approach (that is defaults in the app, not the database).

mdellweg (approved):

ipanova (approved):

PR #3770, opened 2023-04-26
Add content filters to repository endpoint

### LINKED ISSUE #2865: As a user, I want to know which repositories some content is in
Content in pulp is distributed through repositories and distributions. If a client knows the name or HREF for some content, there's currently no way of know which repositories it's available in without looking in each repository. 

Add a `?content_in_latest=` filter to the distributions and repositories APIs that allow users to specify a content HREF and get a list of repositories or distributions that contain the selected content.

Are there any existing ways to do this in pulp?

**Additional context**
This is needed for repository management in galaxy_ng

daviddavis:
fixes #2865

### DIFF
--- a/pulpcore/app/viewsets/repository.py
+++ b/pulpcore/app/viewsets/repository.py
@@ -16,6 +16,7 @@ from pulpcore.app.models import (
     Content,
     Remote,
     Repository,
+    RepositoryContent,
     RepositoryVersion,
 )
 from pulpcore.app.response import OperationPostponedResponse
@@ -41,9 +42,51 @@ from pulpcore.tasking.tasks import dispatch
 from pulpcore.filters import HyperlinkRelatedFilter
 
 
+class RepositoryContentFilter(Filter):
+    """
+    Filter used to filter repositories which have a piece of content
+    """
+
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("help_text", _("Content Unit referenced by HREF"))
+        self.latest = kwargs.pop("latest", False)
+        super().__init__(*args, **kwargs)
+
+    def filter(self, qs, value):
+        """
+        Args:
+            qs (django.db.models.query.QuerySet): The Repository Queryset
+            value (string): of content href to filter
+
+        Returns:
+            Queryset of the Repository containing the specified content
+        """
+
+        if value is None:
+            # user didn't supply a value
+            return qs
+
+        if not value:
+            raise serializers.ValidationError(detail=_("No value supplied for content filter"))
+
+        # Get the content object from the content_href
+        content = NamedModelViewSet.get_resource(value, Content)
+
+        if self.latest:
+            return qs.filter(
+                pk__in=RepositoryContent.objects.filter(
+                    version_removed=None, content__pk=content.pk
+                ).values_list("repository__pk", flat=True)
+            )
+        else:
+            return qs.filter(content__pk=content.pk)
+
+
 class RepositoryFilter(BaseFilterSet):
     pulp_label_select = LabelFilter()
     remote = HyperlinkRelatedFilter(allow_null=True)
+    with_content = RepositoryContentFilter()
+    latest_with_content = RepositoryContentFilter(latest=True)
 
     class Meta:
         model = Repository
--- a/pulpcore/tests/functional/api/test_repos.py
+++ b/pulpcore/tests/functional/api/test_repos.py
@@ -1,0 +1,50 @@
+"""Tests that CRUD repositories."""
+from uuid import uuid4
+
+import pytest
+
+from pulpcore.client.pulp_file import RepositorySyncURL
+
+
+@pytest.mark.parallel
+def test_repository_content_filters(
+    file_content_api_client,
+    file_repository_api_client,
+    file_repository_factory,
+    file_remote_factory,
+    gen_object_with_cleanup,
+    write_3_iso_file_fixture_data_factory,
+    monitor_task,
+):
+    """Test repository's content filters."""
+    # generate a repo with some content
+    repo = file_repository_factory()
+    repo_manifest_path = write_3_iso_file_fixture_data_factory(str(uuid4()))
+    remote = file_remote_factory(manifest_path=repo_manifest_path, policy="on_demand")
+    body = RepositorySyncURL(remote=remote.pulp_href)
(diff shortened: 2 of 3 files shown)


daviddavis:
I can add tests if this change is acceptable.

pulpcore/app/viewsets/repository.py
```diff
@@ -41,9 +42,51 @@
...
+                ).values_list("repository__pk", flat=True)
+            )
+        else:
+            return qs.filter(content__pk=content.pk)
+
+
 class RepositoryFilter(BaseFilterSet):
     pulp_label_select = LabelFilter()
     remote = HyperlinkRelatedFilter(allow_null=True)
+    with_content = RepositoryContentFilter()
```
daviddavis, line 88: I wasn't sure what to name this filter but it seems like [other places used the name `with_content`](https://github.com/pulp/pulpcore/blob/765c70aabb69a6ea71d2841d19d3535cf9dd412e/pulpcore/app/viewsets/publication.py#L364).

pulpcore/app/viewsets/repository.py
```diff
@@ -41,9 +42,51 @@
...
+    currently_with_content = RepositoryContentFilter(latest=True)
```
mdellweg, line 89: Should we name this "latest_with_content"? Maybe easier to relate to the "latest_version"?

Being exposed by distributions, a repository can have multiple current versions...

pulpcore/app/viewsets/repository.py
```diff
@@ -41,9 +42,51 @@
...
+
+        if value is None:
+            # user didn't supply a value
+            return qs
+
+        if not value:
+            raise serializers.ValidationError(detail=_("No value supplied for content filter"))
+
+        # Get the content object from the content_href
+        content = NamedModelViewSet.get_resource(value, Content)
```
mdellweg, line 73: This, I'd expect to throw 404s, we should translate to ValidationError too.

pulpcore/app/viewsets/repository.py
mdellweg: Yes, i think this name is fine.

pulpcore/app/viewsets/repository.py
daviddavis: The method throws validation errors already depending on the problem. Here's an example:

https://github.com/pulp/pulpcore/blob/8b665bb00ad10d14d745f256017779ff38dbf6fb/pulpcore/app/viewsets/base.py#L193

daviddavis:
Test PR: https://github.com/pulp/pulp_file/pull/953

pulpcore/app/viewsets/repository.py
```diff
@@ -41,9 +42,51 @@
...
+
+        if self.latest:
+            return qs.filter(
+                pk__in=RepositoryContent.objects.filter(
+                    version_removed=None, content__pk=content.pk
+                ).values_list("repository__pk", flat=True)
+            )
+        else:
+            return qs.filter(content__pk=content.pk)
```
lubosmj, line 82: What do we do about the "user_hidden" field?

pulpcore/app/viewsets/repository.py
```diff
@@ -41,9 +42,51 @@
...
+
+        # Get the content object from the content_href
+        content = NamedModelViewSet.get_resource(value, Content)
+
+        if self.latest:
+            return qs.filter(
+                pk__in=RepositoryContent.objects.filter(
+                    version_removed=None, content__pk=content.pk
+                ).values_list("repository__pk", flat=True)
+            )
```
lubosmj, line 80: Will this work when we set retain_repo_versions to 1?

https://docs.pulpproject.org/pulpcore/workflows/repo-versioning.html#version-retention

lubosmj (commented):
@ipanova was the one who did not embrace the need of creating such a filter before. Can you add your thoughts here, @ipanova? I like the change.

pulpcore/app/viewsets/repository.py
daviddavis: Yes, it should. Setting retain_repo_versions to 1 should just keep the latest version (where version_removed=None).

daviddavis:
@lubosmj @ipanova the other approach that was proposed in the issue was adding a `latest` filter to the repo version endpoint. I think it's a great idea but it wouldn't work in our particular use case. Our users only deal with repos: they create, sync, modify, publish, and distribute repos. So we don't expose repo versions or the repo version endpoints to our users and therefore users can't search for repos based on content.

mdellweg (approved):

mdellweg:
Last pulpcore meeting we had a discussion that new test (like these) that only test a new pulpcore feature without any necessary adoption in the plugin should come with the code and live in this repository. Are you willing to move your's here? (This is not for existing tests, because it does not really matter for them.)

daviddavis:
@mdellweg ok, done.

pulpcore/tests/functional/api/test_repos.py
```diff
@@ -0,0 +1,50 @@
...
+    monitor_task(response.task)
+    repo = file_repository_api_client.read(repo.pulp_href)
+
+    # the repo still has the content unit
+    results = file_repository_api_client.list(with_content=content.pulp_href).results
+    assert [repo] == results
+
+    # but not in its latest version anymore
+    results = file_repository_api_client.list(latest_with_content=content.pulp_href).results
+    assert [] == results
```
mdellweg, line 50: I prefer to have it "assert result == expected". Just a personal thing probably, so feel free to ignore.

pulpcore/tests/functional/api/test_repos.py
```diff
@@ -0,0 +1,50 @@
...
+    file_repository_api_client,
+    file_remote_factory,
+    gen_object_with_cleanup,
+    write_3_iso_file_fixture_data_factory,
+    monitor_task,
+):
+    """Test repository's content filters."""
+    # generate a repo with some content
+    repo = gen_object_with_cleanup(file_repository_api_client, gen_repo())
+    repo_manifest_path = write_3_iso_file_fixture_data_factory(utils.uuid4())
```
mdellweg, line 22: Let's try not to use any of the pulp-smash weirdness. There should be a file_repository_factory and we can use `str(uuid.uuid4())`.

mdellweg (approved):

PR #3780, opened 2023-05-02
Add 3.24 to supported-ci-versions in template.

ggainey:
[noissue]

### DIFF
--- a/template_config.yml
+++ b/template_config.yml
@@ -24,6 +24,7 @@ ci_update_branches:
 - '3.21'
 - '3.22'
 - '3.23'
+- '3.24'
 cli_package: pulp-cli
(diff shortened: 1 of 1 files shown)


template_config.yml
```diff
@@ -24,6 +24,7 @@ ci_update_branches:
 - '3.21'
 - '3.22'
 - '3.23'
+- '3.24'
```
mdellweg, line 27: Do we need to keep 3.23 on?

mdellweg (approved):

template_config.yml
ggainey: Galaxy 4.7 uses core/3.23, so I'd say "yes"

lubosmj (approved):

PR #3784, opened 2023-05-03
Fix domain representation created_resources

### LINKED ISSUE #3783: Task list and show endpoint optimizations for created_resources regress with domains
created_resources always render with "default" domain.

mdellweg:
fixes #3783

### DIFF
--- a/pulpcore/app/serializers/base.py
+++ b/pulpcore/app/serializers/base.py
@@ -155,7 +155,8 @@ class RelatedField(
     """
 
 
-PKOnlyObject = namedtuple("PKOnlyObject", ["pk"])
+PKObject = namedtuple("PKObject", ["pk"])
+PKDomainObject = namedtuple("PKDomainObject", ["pk", "pulp_domain"])
 
 
 class RelatedResourceField(RelatedField):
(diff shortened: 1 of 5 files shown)


mdellweg:
Testing this can be seen in https://github.com/pulp/pulpcore/pull/3781 which needs pulp-certguard to have domain enabled tests to proceed.

pulpcore/tasking/util.py
```diff
@@ -10,7 +10,7 @@
 _logger = logging.getLogger(__name__)
 
 
-def cancel(task_id):
+def cancel_task(task_id):
```
decko, line 13: :pray:

pulpcore/app/serializers/base.py
```diff
@@ -155,7 +155,8 @@ class RelatedField(
     """
 
 
-PKOnlyObject = namedtuple("PKOnlyObject", ["pk"])
+PKObject = namedtuple("PKObject", ["pk"])
+PKDomainObject = namedtuple("PKDomainObject", ["pk", "pulp_domain"])
```
decko, line 159: Why we need to have those two objects since we have the `default` domain?

mdellweg: We may only inject the pulp_domain into objects with urls that are supposed to have one. This sounded like the safe way to me.

decko (approved):

ggainey (approved):

PR #3785, opened 2023-05-03
Remove Label, LabelsField and LabelSelectFilter

### LINKED ISSUE #3615: Remove `LabelsField`, `Label`, and `LabelSelectFilter`

mdellweg:
fixes #3615


pulpcore/app/apps.py
```diff
@@ -257,11 +256,6 @@ def ready(self):
             sender=self,
             dispatch_uid="populate_artifact_serving_distribution_identifier",
         )
-        post_migrate.connect(
-            _migrate_remaining_labels,
-            sender=self,
-            dispatch_uid="migrate_remaining_labels_identifier",
-        )
```
lubosmj, line 264: What is the impact of this change if a user migrates from an old version of pulpcore which did not have migrated/deleted the remaining labels? Will their "labels" get removed instead of copied to "pulp_labels"?

mdellweg: The labels will just get removed. I believe we said it was OK, because labels were in tech-preview back then.

mdellweg: I added a safety-check. Let's see how this pans out.

ggainey (approved):

lubosmj (approved):

Issue #3459, opened 2022-12-13
As a plugin writer, I have to opt-out to having a "non-redirect" behavior in the content app


bmbouter:
With the pulpcore 3.22 release, we planned to make the "redirect to include a trailing slash" behavior a plugin opt-in feature. We couldn't make it opt-out because pulp_container breaks with this and pre-existing GA releases of pulp_container declare forward compatibility through pulpcore 3.24.

This issue tracks the change:  With 3.25, have the "redirect to include a trailing slash" be a plugin opt-out feature and have it enabled by default.

bmbouter:
@pulp/core FYI

daviddavis:
Just adding some historical notes here.

The original redirect feature issue: https://github.com/pulp/pulpcore/issues/3173

The original PR to add the feature: https://github.com/pulp/pulpcore/pull/3434

The PR that reverted/removed the feature: https://github.com/pulp/pulpcore/pull/3465

ggainey:
Clarification: implementing this requires more than just re-pushing the PR - it requires adding machinery to let plugins opt in/out of the behavior

lubosmj:
I do not understand the need to enable or disable the feature in the plugin's code. We decided to revert the commit because we had compatibility issues in pulp_container (https://github.com/pulp/pulpcore/pull/3434#discussion_r1047175632) and pulp_python (https://github.com/pulp/pulpcore/pull/3456). The branch where the commit was merged had not been identified as a branch holding breaking changes.

Is it acceptable to merge the very same commit with, for instance, a hook method that will enable plugin writers to create an additional wrap-around on top of the non-redirect workflow, if needed? This is what we were trying to add to pulp_container to fix the issues: https://github.com/pulp/pulp_container/pull/1179.

lubosmj:
Is this classified as a duplicate of https://github.com/pulp/pulpcore/issues/3173 now?

lubosmj:
This is the API change required for https://github.com/pulp/pulpcore/issues/3173.

PR #3795, opened 2023-05-04

mdellweg:

Changes the order in which pulpcore-content looks for published artifacts The handler now looks for a published artifact before it tries to look for an index.html or listing the directory.

fixes: #3173
fixes: #3459

### DIFF
--- a/CHANGES/3173.feature
+++ b/CHANGES/3173.feature
@@ -1,0 +1,1 @@
+Content app now redirects requests without a trailing slash to a path with the trailing slash. 
--- a/CHANGES/plugin_api/3459.feature
+++ b/CHANGES/plugin_api/3459.feature
@@ -1,0 +1,2 @@
+Added a flag ``REDIRECT_ON_MISSING_SLASH`` for plugin distribution models to opt out of automatic
+redirecting if a path should end in a slash but does not.
--- a/CHANGES/plugin_api/3459.removal
+++ b/CHANGES/plugin_api/3459.removal
@@ -1,0 +1,5 @@
+Starting with this release, pulpcore will default to redirect clients of the content app when
+given path locations without the trailing slash. 
+
+A plugin must set ``REDIRECT_ON_MISSING_SLASH=False`` on distribution models for content types
+where this is inappropriate.
--- a/pulpcore/app/models/publication.py
+++ b/pulpcore/app/models/publication.py
@@ -529,8 +529,11 @@ class Distribution(MasterModel):
         pulp_domain (models.ForeignKey): The domain the Distribution is a part of.
     """
 
-    # If distribution serves publications, set by subclasses for proper handling in content app
+    # If distribution serves publications, set by subclasses for proper handling in content app.
     SERVE_FROM_PUBLICATION = False
+
+    # Plugins can opt out of redirecting by settings this to False.
+    REDIRECT_ON_MISSING_SLASH = True
 
     name = models.TextField(db_index=True)
     pulp_labels = HStoreField(default=dict)
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -9,6 +9,7 @@ from aiohttp.web import FileResponse, StreamResponse, HTTPOk
 from aiohttp.web_exceptions import (
     HTTPForbidden,
     HTTPFound,
+    HTTPMovedPermanently,
     HTTPNotFound,
     HTTPRequestRangeNotSatisfiable,
 )
@@ -280,6 +281,9 @@ class Handler:
             DistroListings: when multiple matches are possible.
             PathNotResolved: when not matched.
         """
+        path_ends_in_slash = path.endswith("/")
+        if not path_ends_in_slash:
+            path = f"{path}/"
         base_paths = cls._base_paths(path)
         distro_model = cls.distribution_model or Distribution
         domain = get_domain()
@@ -303,8 +307,11 @@ class Handler:
                     pulp_domain=domain, base_path__startswith=path
                 )
                 if distros.count():
-                    raise DistroListings(path=path, distros=distros)
-
+                    if path_ends_in_slash:
+                        raise DistroListings(path=path, distros=distros)
+                    else:
+                        # The list of a subset of distributions was requested without a trailing /
+                        raise HTTPMovedPermanently(f"{settings.CONTENT_PATH_PREFIX}{path}")
             log.debug(
                 _("Distribution not matched for {path} using: {base_paths}").format(
                     path=path, base_paths=base_paths
@@ -483,9 +490,7 @@ class Handler:
                 ).values_list("content_artifact_id", "size")
                 sizes.update({artifacts_to_find[ra_ca_id]: size for ra_ca_id, size in r_artifacts})
 
-                return directory_list, dates, sizes
-            else:
-                raise PathNotResolved(path)
+            return directory_list, dates, sizes
 
         return await sync_to_async(list_directory_blocking)()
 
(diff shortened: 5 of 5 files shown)


mdellweg:
Is this in any way close to what was requested on those two issues?
(I'm a bit lost on what behaviour should be opted in/out).

CHANGES/plugin_api/3459.feature
```diff
@@ -0,0 +1,2 @@
+Added a flag `REDIRECT_ON_MISSING_SLASH` for plugin distrubution models to opt out of automatic
```
daviddavis, line 1:
```suggestion
Added a flag `REDIRECT_ON_MISSING_SLASH` for plugin distribution models to opt out of automatic
```

daviddavis:
> Is this in any way close to what was requested on those two issues?
> (I'm a bit lost on what behaviour should be opted in/out).

I believe it is. I think the behavior is supposed to be opt-out.

bmbouter:
> I believe it is. I think the behavior is supposed to be opt-out.

The default today is enabled-for-all right? If so, +1 to opt-out.

CHANGES/plugin_api/3459.removal
```diff
@@ -0,0 +1,5 @@
+Starting with this release, pulpcore will default to redirect clients of the content app when
+given path locations without the trailing slash. 
+
+A plugin must set `REDIRECT_ON_MISSING_SLASH=False` on distribution models for content types
+where this is in appropriate.
```
bmbouter, line 5:
```suggestion
where this is inappropriate.
```

CHANGES/plugin_api/3459.feature
```diff
@@ -0,0 +1,2 @@
+Added a flag `REDIRECT_ON_MISSING_SLASH` for plugin distribution models to opt out of automatic
```
bmbouter, line 1:
```suggestion
Added a flag ``REDIRECT_ON_MISSING_SLASH`` for plugin distribution models to opt out of automatic
```

CHANGES/plugin_api/3459.removal
```diff
@@ -0,0 +1,5 @@
+Starting with this release, pulpcore will default to redirect clients of the content app when
+given path locations without the trailing slash. 
+
+A plugin must set `REDIRECT_ON_MISSING_SLASH=False` on distribution models for content types
```
bmbouter, line 4:
```suggestion
A plugin must set ``REDIRECT_ON_MISSING_SLASH=False`` on distribution models for content types
```

CHANGES/plugin_api/3459.removal
```diff
@@ -0,0 +1,5 @@
+Starting with this release, pulpcore will default to redirect clients of the content app when
```
bmbouter, line 1: Didn't it do this before?

pulpcore/app/models/publication.py
```diff
@@ -529,9 +529,12 @@ class Distribution(MasterModel):
         pulp_domain (models.ForeignKey): The domain the Distribution is a part of.
     """
 
-    # If distribution serves publications, set by subclasses for proper handling in content app
+    # If distribution serves publications, set by subclasses for proper handling in content app.
     SERVE_FROM_PUBLICATION = False
 
+    # Plugins can opt out of redirecting by settings this to False.
+    REDIRECT_ON_MISSING_SLASH = True
```
bmbouter, line 536: Maybe the plugin writer guide should get a 1 paragraph on this? (this is an optional idea)

daviddavis:
> The default today is enabled-for-all right? If so, +1 to opt-out.

I think the original change (to redirect slash-less index paths to slash paths) was enabled-for-all but was reverted because it caused issues for pulp_container. So this new change would enable it for all except those that opt out.

CHANGES/plugin_api/3459.removal
daviddavis: No, it was reverted in https://github.com/pulp/pulpcore/pull/3465 because the behavior broke pulp_container.

bmbouter:
I'm not feeling confident enough to review the code so I'm avoiding being the formal review. Let's see if we can get reviewers at tomorrow's pulpcore meeting. If we often can't find enough reviewers maybe we should adjust the review required count from 2 to 1.

CHANGES/plugin_api/3459.removal
daviddavis: Or I guess it might've been pulp_python? Maybe @gerrod3 knows.

daviddavis:
> 
I'm not feeling confident enough to review the code so I'm avoiding being the formal review. Let's see if we can get reviewers at tomorrow's pulpcore meeting. If we often can't find enough reviewers maybe we should adjust the review required count from 2 to 1.

FWIW, I worked with @dkliban on this change and from a high level it looks good to me. I can't speak to things like performance though.

CHANGES/3173.feature
```diff
@@ -0,0 +1 @@
+Content app now redirects requests without a trailing slash to a path with the trailing slash.
```
dralley, line 1: By default

dralley (approved):

CHANGES/3173.feature
mdellweg: I'm not sure this would be helpful. For the users perspective, pulpcore is just doing it. They cannot configure it. It's just certain content types that don't want it in their distribution paths.

pulpcore/content/handler.py
```diff
@@ -530,6 +535,12 @@ async def _match_and_stream(self, path, request):
         rel_path = rel_path[len(distro.base_path) :]
         rel_path = rel_path.lstrip("/")
 
+        if rel_path == "" and not path.endswith("/"):
+            # The root of a distribution base_path was requested without a slash
+            if not distro.REDIRECT_ON_MISSING_SLASH:
+                raise ObjectDoesNotExist(request.path)
```
gerrod3, line 541: Is this the right exception? This looks like a Django exception, so wouldn't this report a 500 error if it got raised?

pulpcore/content/handler.py
```diff
@@ -628,6 +619,39 @@ async def _match_and_stream(self, path, request):
...
+                if ends_in_slash is False:
+                    # index.html found, but user didn't specify a trailing slash
+                    if not distro.REDIRECT_ON_MISSING_SLASH:
+                        raise ObjectDoesNotExist(request.path)
+                    raise HTTPMovedPermanently(f"{request.path}/")
+                rel_path = index_path
+                headers = self.response_headers(rel_path)
+            except ObjectDoesNotExist:
+                if not distro.REDIRECT_ON_MISSING_SLASH:
+                    raise
```
gerrod3, line 639: Have you tested this to see what the response is when a bad path is given?

pulpcore/content/handler.py
```diff
@@ -628,6 +619,39 @@ async def _match_and_stream(self, path, request):
...
+                    return HTTPOk(
+                        headers={"Content-Type": "text/html"},
+                        body=self.render_html(
+                            dir_list, path=request.path, dates=dates, sizes=sizes
+                        ),
+                    )
+
         if repo_version and not publication and not distro.SERVE_FROM_PUBLICATION:
             if rel_path == "" or rel_path[-1] == "/":
                 index_path = "{}index.html".format(rel_path)
```
gerrod3, line 657: This is part of the reason for the refactor I am doing. All of this new logic has to be repeated twice for distro's serving repository versions.

pulpcore/content/handler.py
```diff
@@ -628,6 +619,39 @@ async def _match_and_stream(self, path, request):
...
                         )
 
+            # Look for index.html or list the directory
+            ends_in_slash = rel_path == "" or rel_path.endswith("/")
+            if not ends_in_slash:
+                rel_path = f"{rel_path}/"
+            try:
+                index_path = "{}index.html".format(rel_path)
+
+                await publication.published_artifact.aget(relative_path=index_path)
```
gerrod3, line 629: BTW, moving this logic downward has probably broke the normal behavior for plugins that create their own index.html pages in a publication. We don't use the result from this `aget` we just check if it exists and then the logic from L585-L599 would actually serve the artifact. Moving this down will break current functionality. (We don't see breaks in tests probably because pulp_file uses the dynamic generated listing.)

gerrod3 (changes requested):
This can not be merged in its current state. I'm thinking we should try the refactor first before making changes as the behavior will be easier to follow.

PR #3797, opened 2023-05-05
Refactor cast to do only one db call

mdellweg:
[noissue]


mdellweg:
Well, this does not exactly work as designed...

PR #3801, opened 2023-05-08
Add pre delete hook to domain for orphans

### LINKED ISSUE #3800: Orphaned content should not prevent to delete a domain.
Being unable to delete a domain when all there is left in it is orphaned content is unintuitive.

mdellweg:
fixes #3800

### DIFF
--- a/pulpcore/app/models/domain.py
+++ b/pulpcore/app/models/domain.py
@@ -49,6 +49,15 @@ class Domain(BaseModel, AutoAddObjPermsMixin):
     def prevent_default_deletion(self):
         raise models.ProtectedError("Default domain can not be updated/deleted.", [self])
 
+    @hook(BEFORE_DELETE, when="name", is_not="default")
+    def _cleanup_orphans_pre_delete(self):
+        if self.content_set.exclude(version_memberships__isnull=True).exists():
+            raise models.ProtectedError("There is active content in the domain.")
+        self.content_set.filter(version_memberships__isnull=True).delete()
+        for artifact in self.artifact_set.all().iterator():
+            # Delete on by one to properly cleanup the storage.
+            artifact.delete()
+
     class Meta:
         permissions = [
             ("manage_roles_domain", "Can manage role assignments on domain"),
--- a/CHANGES/3800.feature
+++ b/CHANGES/3800.feature
@@ -1,0 +1,1 @@
+Added a pre-delete hook to domains in order to stop orphaned content preventing the deletion.
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -294,6 +294,12 @@ HIDE_GUARDED_DISTRIBUTIONS = False
 
 DOMAIN_ENABLED = False
 
+SHELL_PLUS_IMPORTS = [
+    "from pulpcore.app.util import get_domain, get_domain_pk, set_domain, get_url, extract_pk",
+    "from pulpcore.tasking.tasks import dispatch",
+    "from pulpcore.tasking.util import cancel_task",
+]
+
 # HERE STARTS DYNACONF EXTENSION LOAD (Keep at the very bottom of settings.py)
 # Read more at https://dynaconf.readthedocs.io/en/latest/guides/django.html
 from dynaconf import DjangoDynaconf, Validator  # noqa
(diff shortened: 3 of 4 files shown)


pulpcore/app/models/domain.py
```diff
@@ -49,6 +49,15 @@ def get_storage(self):
     def prevent_default_deletion(self):
         raise models.ProtectedError("Default domain can not be updated/deleted.", [self])
 
+    @hook(BEFORE_DELETE, when="name", is_not="default")
+    def _cleanup_orphans_pre_delete(self):
+        if self.content_set.exclude(version_memberships__isnull=True).exists():
+            raise models.ProtectedError("There is active content in the domain.")
+        self.content_set.filter(version_memberships__isnull=True).delete()
```
gerrod3, line 56: The delete on the domain can still fail if there are other objects within the domain. This could lead to weird behavior where the orphans get deleted but the domain is still there.

mdellweg: Hmm. What would you suggest?

gerrod3: Does the artifact file delete get triggered after the transaction completes? My understanding for why we delete one by one is so that django-storages properly deletes the file.

mdellweg: Hah, that _is_ a good question.

mdellweg: You are onto something. But i honestly believe this is a separate issue.

```
In [6]: Artifact.objects.all()
Out[6]: <BulkTouchQuerySet [<Artifact: pk=0188014a-120f-7e05-9e65-3de72d45182e>]>

In [7]: a = Artifact.objects.first()

In [8]: with transaction.atomic():
   ...:     a.delete()
   ...:     raise Exception()
   ...: 
Exception                                 Traceback (most recent call last)
Cell In[8], line 3
      1 with transaction.atomic():
      2     a.delete()
----> 3     raise Exception()

Exception: 

In [9]: a
Out[9]: <Artifact: pk=None>

In [10]: a = Artifact.objects.first()

In [11]: a
Out[11]: <Artifact: pk=0188014a-120f-7e05-9e65-3de72d45182e>

In [12]: a.file
Out[12]: <FieldFile: artifact/99/322da247546beac65c2f5b7d164e94dcaf050b85a6076a00cf4fec5e21a6ed>

In [13]: a.file.read()
FileNotFoundError                         Traceback (most recent call last)
```

pulpcore/app/models/domain.py
```diff
@@ -49,6 +49,15 @@ def get_storage(self):
     def prevent_default_deletion(self):
         raise models.ProtectedError("Default domain can not be updated/deleted.", [self])
 
+    @hook(BEFORE_DELETE, when="name", is_not="default")
+    def _cleanup_orphans_pre_delete(self):
```
dralley, line 53: This delete is expected to happen inside of a task, not the API, correct?

mdellweg:
I'm not so sure, should this be a bugfix?

dralley:
It could make sense as a bugfix - maybe go ahead and do this if you think we may need to backport it?

mdellweg:
Let's not do it for the sake of backporting... You can still issue an orphan cleanup prior to deleting the domain. So there's a rather easy workaround.

gerrod3 (approved):

pulpcore/app/models/domain.py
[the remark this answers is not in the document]
mdellweg: I know, but at this point, you intend to delete the domain (now) which would fail if anything useful (like repositories) is left. And in that condition no artifact will be deleted either. I believe it is safe to assume that at this point we do not need to wait for any expiry anymore.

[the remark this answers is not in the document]
ipanova: I understand what you're saying and it probably does sound logical and desirable, it's just here we're changing the concept of what an 'orphan' means. I know that right now i am being overly picky, however, same logic could be applicable to 'i want to delete the content (now) but.. oh wait..it has protection time'.

As long as it was on purpose, it's ok just leaving an explicit note on the code would have been helpful.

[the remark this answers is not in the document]
mdellweg: I agree, i should have added a comment. It was too obvious to me. To be fair, the protection time is not on the content, but you can provide the protection time with the delete orphan call. So you can delete your content "now" if you intend to.

Issue #3802, opened 2023-05-08
Content app responses slow down as the number of ContentArtifacts + RepositoryContent grows


gerrod3:
**Version**
<=3.24

**Describe the bug**
As the number of ContentArtifacts and RepositoryContent objects grows, the content app responses for artifacts for distro's serving pass_through publications or repository versions slows down. On a machine with 2.5 million CAs and 1.3 million RCs retrieving one artifact from the content app takes around 15-20 secs.

The problem lies in the subqueries generated in the ContentArtifact query we use: https://github.com/pulp/pulpcore/blob/7c6d59f5f302efab669e335a23f1912eae739f2c/pulpcore/content/handler.py#L605-L611. The generated query is quite gnarly with its WHERE clauses: 
```
SELECT 
  "core_contentartifact"."pulp_id", 
  "core_contentartifact"."pulp_created", 
  "core_contentartifact"."pulp_last_updated", 
  "core_contentartifact"."artifact_id", 
  "core_contentartifact"."content_id", 
  "core_contentartifact"."relative_path", 
  "core_artifact"."pulp_id", 
  "core_artifact"."pulp_created", 
  "core_artifact"."pulp_last_updated", 
  "core_artifact"."file", 
  "core_artifact"."size", 
  "core_artifact"."md5", 
  "core_artifact"."sha1", 
  "core_artifact"."sha224", 
  "core_artifact"."sha256", 
  "core_artifact"."sha384", 
  "core_artifact"."sha512", 
  "core_artifact"."timestamp_of_interest", 
  "core_artifact"."pulp_domain_id", 
  "core_domain"."pulp_id", 
  "core_domain"."pulp_created", 
  "core_domain"."pulp_last_updated", 
  "core_domain"."name", 
  "core_domain"."description", 
  "core_domain"."storage_class", 
  "core_domain"."storage_settings", 
  "core_domain"."redirect_to_object_storage", 
  "core_domain"."hide_guarded_distributions" 
FROM 
  "core_contentartifact" 
  LEFT OUTER JOIN "core_artifact" ON (
    "core_contentartifact"."artifact_id" = "core_artifact"."pulp_id"
  ) 
  LEFT OUTER JOIN "core_domain" ON (
    "core_artifact"."pulp_domain_id" = "core_domain"."pulp_id"
  ) 
WHERE 
  (
    "core_contentartifact"."content_id" IN (
      SELECT 
        V0."pulp_id" 
      FROM 
        "core_content" V0 
[...]

PR #3803, opened 2023-05-08
Add index for ContentArtifact's relative_path

gerrod3:
fixes: #3802


pulpcore/content/handler.py
```diff
@@ -540,6 +541,60 @@ async def _match_and_stream(self, path, request):
...
+                if ca := await caq.afirst():
+                    # Check that found CA is inside repository version
+                    if (
+                        await RepositoryContent.objects.filter(
+                            content=ca.content_id,
+                            repository=repo_version.repository_id,
+                            version_added__number__lte=repo_version.number,
+                        )
+                        .exclude(version_removed__number__lte=repo_version.number)
+                        .aexists()
```
gerrod3, line 588: This is the secret of the optimization. Instead of performing one large query with multiple subqueries, it is now broken down into multiple smaller & faster queries. This improves the performance of both 200 & 404 requests which is important since the Redis cache (if enabled) does not save 404 responses, so client tooling that checks a non-existent/supported endpoint would suffer each time.

Also, the basis for the refactor was to make sharing this optimization easier since it is applied to both pass-through Publications and Repository Version served distributions.

gerrod3:
I also have a branch before the 4.2 updates, that can be backported. Should I open it against 3.24?

pulpcore/content/handler.py
dralley: >Instead of performing one large query with multiple subqueries, it is now broken down into multiple smaller & faster queries.

That *shouldn't* make a difference, unless the Django ORM is doing a worse job writing the large query (entirely possible).  If it were exactly the same thing happening on the backend I would think one query would be better.

Nice job though

gerrod3: I go a bit more into in my comment on the issue (you can see the generated query there): https://github.com/pulp/pulpcore/issues/3802#issue-1700104124, but it really seems that subqueries destroy performance as the table gets very large. Django makes a note of it in there docs and it seems to be that Postgres (or atleast the version we use) isn't very optimized for them: https://docs.djangoproject.com/en/4.2/ref/models/querysets/#in

pulpcore/content/handler.py
```diff
@@ -559,117 +612,55 @@ async def _match_and_stream(self, path, request):
...
+                    headers = self.response_headers(index_path)
                 else:
-                    dir_list, dates, sizes = await self.list_directory(repo_version, None, rel_path)
+                    # Dynamically generate a directory listing
+                    dir_list, dates, sizes = await self.list_directory(**index_kwargs)
                     dir_list.update(
                         await sync_to_async(distro.content_handler_list_directory)(rel_path)
                     )
-                    return HTTPOk(
+                    raise HTTPOk(
```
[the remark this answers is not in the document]
gerrod3: Iirc this is an aiohttp framework thing, the cache can handle both. All of these responses are of exception classes and thus should be raised and aiohttp will handle the rest. They technically work with both return or raise, but I think I read in their docs they should be raised as they could stop supporting return in the future. Another point to switching to django...

dralley:
Is there a reason it's still in draft?  Should I wait to review?

pulpcore/content/handler.py
gerrod3: After some extensive testing, I've opted for adding an index on `relative_path` of ContentArtifact and returning the query to its original form. The filtering by `relative_path` was the slowest part of the query by 2 orders of magnitude and would quickly grow to 3-5 orders of magnitude slower when the number of ContentArtifacts grew large.

gerrod3:
I added some functional tests to cover some extra content app behavior that isn't being tested in pulp_file. Still missing tests specifically for non-publication based plugins, but the unit tests I wrote were messy and not too helpful, so I decided to scrap them.

dralley:
@gerrod3 How difficult would it be to separate the functional changes from the refactor, in the commit?

If it's difficult, then don't, but could you make note of the bits of code that had functional changes rather than that which just moved around?

I assume the primary optimization here is the index.

gerrod3:
> @gerrod3 How difficult would it be to separate the functional changes from the refactor, in the commit?
> 
> If it's difficult, then don't, but could you make note of the bits of code that had functional changes rather than that which just moved around?
> 
> I assume the primary optimization here is the index.

I can move the refactor to a separate PR and have just the optimization be in this PR. Also, this can not be backported since it has a migration.

mdellweg (approved):

dralley (approved):

dkliban (approved):

PR #3808, opened 2023-05-09
Delete artifact files after sucessful transaction

mdellweg:


gerrod3 (approved):
Nice fix!

pulpcore/app/models/content.py
```diff
@@ -162,7 +162,7 @@ def delete(self, *args, **kwargs):
             kwargs (dict): dictionary of keyword arguments to pass to Model.delete()
         """
         super().delete(*args, **kwargs)
-        self.file.delete(save=False)
+        transaction.on_commit(partial(self.file.delete, save=False))
```
dralley, line 165: Let's leave a comment about why this is being done

dralley: Can we test the basic function here manually before merging?  e.g. create an artifact, delete the artifact record without deleting the file somehow, then use a reference to the filefield to delete the remaining file without the record being present.

Just to make sure there's no unexpected behavior.  Also should we be careful of `self.file` being nulled out or invalidated behind our backs?

dralley: >having a nullified file field (and such a thing cannot come from the database [1])

I recall seeing it use an empty string in some cases, though this was an experiment I did more than a year ago so I can't remember the details.

dralley (approved):

Issue #3798, opened 2023-05-05
Provide async versions of `cast` and `lastest_version`.


mdellweg:
This came up in the review for the Django 4 update. To reduce the risk from too many changes we scoped this parts out. The overarching goal of this effort is to make better use of Django's async interfaces to the ORM.

PR #3816, opened 2023-05-11
Provide async version of latest_version

MichalPysik:
The repository model's latest_version function now has a separate asynchronous version called alatest_version.

fixes: #3798

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -225,6 +225,21 @@ class Repository(MasterModel):
             model = self.versions.complete().latest()
             return model
 
+    async def alatest_version(self):
+        """
+        Get the latest RepositoryVersion on a repository asynchronously
+
+        Args:
+            repository (pulpcore.app.models.Repository): to get the latest version of
+
+        Returns:
+            pulpcore.app.models.RepositoryVersion: The latest RepositoryVersion
+
+        """
+        with suppress(RepositoryVersion.DoesNotExist):
+            model = await self.versions.complete().alatest()
+            return model
+
     def natural_key(self):
         """
         Get the model's natural key.
--- a/CHANGES/plugin_api/3798.misc
+++ b/CHANGES/plugin_api/3798.misc
@@ -1,0 +1,1 @@
+Provided asynchronous version of the latest_version function (repository model).
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -569,7 +569,7 @@ class Handler:
                     pass
 
             if not repo_version:
-                repo_version = await sync_to_async(repository.latest_version)()
+                repo_version = await repository.alatest_version()
 
         if publication:
             if rel_path == "" or rel_path[-1] == "/":


pulpcore/app/models/base.py
```diff
@@ -164,6 +165,32 @@ def cast(self):
...
+        # on a OneToOneField, which identifies it as a potential detail relation.
+        for rel in self._meta.related_objects:
+            if rel.one_to_one and issubclass(rel.related_model, self._meta.model):
+                # The name of this relation is the name of the attr on the model instance.
+                # If that attr as a value, that means a row exists for this model in the
+                # related detail table. Cast and return this value, recursively following
+                # master/detail relationships down to the last table (the most detailed).
+                try:
+                    related_instance = getattr(self, rel.name)
+                    if asyncio.iscoroutinefunction(related_instance.cast):
```
mdellweg, line 183: Why do we need this?

pulpcore/app/models/base.py
```diff
@@ -164,6 +165,29 @@ def cast(self):
...
+        # on a OneToOneField, which identifies it as a potential detail relation.
+        for rel in self._meta.related_objects:
+            if rel.one_to_one and issubclass(rel.related_model, self._meta.model):
+                # The name of this relation is the name of the attr on the model instance.
+                # If that attr as a value, that means a row exists for this model in the
+                # related detail table. Cast and return this value, recursively following
+                # master/detail relationships down to the last table (the most detailed).
+                try:
+                    related_instance = getattr(self, rel.name)
+                    return await sync_to_async(related_instance.acast)()
```
mdellweg, line 183: That is already a coroutine..

pulpcore/app/models/repository.py
```diff
@@ -226,6 +227,24 @@ def latest_version(self):
...
+        Args:
+            repository (pulpcore.app.models.Repository): to get the latest version of
+
+        Returns:
+            pulpcore.app.models.RepositoryVersion: The latest RepositoryVersion
+
+        """
+        # suppress cannot be used in an async context
+        try:
+            model = await sync_to_async(self.versions.complete().latest)()
```
mdellweg, line 243: I think we want an async version of `latest` too then.

pulpcore/app/models/base.py
dralley: Agreed, no need to wrap this. I believe that  `getattr()` on `related_instance`, however, is doing a background query, and if that is the case it may need to be wrapped.

pulpcore/app/models/repository.py
dralley: Yes

CHANGES/3798.misc
```diff
@@ -0,0 +1 @@
+Provided asynchronous versions of the latest_version function and the cast function.
```
ipanova, line 1: This is probably a plugin writers change that belongs to CHANGES/plugin_api

dralley: +1

pulpcore/app/models/repository.py
```diff
@@ -225,6 +225,24 @@ def latest_version(self):
...
+        Get the latest RepositoryVersion on a repository asynchronously
+
+        Args:
+            repository (pulpcore.app.models.Repository): to get the latest version of
+
+        Returns:
+            pulpcore.app.models.RepositoryVersion: The latest RepositoryVersion
+
+        """
+        # suppress cannot be used in an async context
```
mdellweg, line 239: Who said that?

```python
import asyncio
import contextlib

async def r():
     raise ValueError()

async def outer():
     with contextlib.suppress(ValueError):
         await r()

loop.run_until_complete(outer())
loop.run_until_complete(r())
```

mdellweg (approved):

gerrod3 (approved):

PR #3822, opened 2023-05-12
Adds basic worker metrics

### LINKED ISSUE #3821: Adds open telemetry metrics for tasks
## Background / Need

It would be good to know more about the tasking system to answer the following question:

* I want to know if I have the right number of workers for a given queue in a specific time.

## Proposal

Report measurements from the tasking system, like:
* Number of unblocked waiting tasks waiting more than 5 seconds
* the longest waiting task time

Also, find the best place to trigger those measurements to avoid unneeded queries or any avoidable overload.

bmbouter:
The following metrics are added:
* `task_count` - The cumulative number of tasks a worker has executed since it was started.

closes #3821

### DIFF
--- a/CHANGES/3821.feature
+++ b/CHANGES/3821.feature
@@ -1,0 +1,2 @@
+Adds the "task_count" metric to each worker. This reports the cumulative number of tasks executed
+since the worker has been started.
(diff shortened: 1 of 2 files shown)


ipanova (commented):
The proposal mentions 2 metrics, this is however, still a draft - will the 2nd metric going to be added as well?

PR #3828, opened 2023-05-15
Adds metrics and tracing to the pulp-content app

### LINKED ISSUE #3829: Add the OpenTelemetry Aiohttp-server instrumentation as a vendor library
Until the [PR adding the aiohttp-server instrumentation support into the open-telemetry/opentelemetry-python-contrib library is  reviewed and accepted](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1800)  I propose we vendor the code to enable `pulp-content` telemetry support.

We will remove it when `opentelemetry-python-contrib` support instrumenting the `aiohttp-server`.

decko:
Closes #3829

### DIFF
--- a/CHANGES/3829.doc
+++ b/CHANGES/3829.doc
@@ -1,0 +1,1 @@
+Added the Telemetry section to the docs with information about OpenTelemetry and its usage on Pulp.
--- a/CHANGES/3829.feature
+++ b/CHANGES/3829.feature
@@ -1,0 +1,1 @@
+Implemented OpenTelemetry support and enabled tracing and metrics for the pulp-content app.
--- a/docs/components.rst
+++ b/docs/components.rst
@@ -142,3 +142,34 @@ An example payload:
     	}],
         "postgresqlVersion": 90200
     }
+
+
+.. _telemetry:
+
+Telemetry Support
+-----------------
+
+Pulp can produce OpenTelemetry data, like the number of requests, active connections and latency response for
+both `pulp-api` and `pulp-content` using OpenTelemetry. You can read more about 
+`OpenTelemetry here <https://opentelemetry.io>`_.
+
+If you are using `Pulp in One Container <https://pulpproject.org/pulp-in-one-container/>`_ or `Pulp Operator 
+<https://docs.pulpproject.org/pulp_operator/>`_ and want to enable it, you will need to set the following 
+environment variables:
+
+* ``PULP_OTEL_ENABLED`` set to ``True``.
+* ``OTEL_EXPORTER_OTLP_ENDPOINT`` set to the address of your OpenTelemetry Collector instance 
+  ex. ``http://otel-collector:4318``.
+* ``OTEL_EXPORTER_OTLP_PROTOCOL`` set to ``http/protobuf``.
+
+If you are using other type of installation maybe you will need to manually initialize Pulp using the 
+`OpenTelemetry automatic instrumentation 
+<https://opentelemetry.io/docs/instrumentation/python/getting-started/#instrumentation>`_
+and set the following environment variables:
+
+* ``OTEL_EXPORTER_OTLP_ENDPOINT`` set to the address of your OpenTelemetry Collector instance 
+  ex. ``http://otel-collector:4318``.
+* ``OTEL_EXPORTER_OTLP_PROTOCOL`` set to ``http/protobuf``.
+
+You will need to run an instance of OpenTelemetry Collector. You can read more about the `OpenTelemetry 
+Collector here <https://opentelemetry.io/docs/collector/>`_.
--- a/pulpcore/content/instrumentation.py
+++ b/pulpcore/content/instrumentation.py
@@ -1,0 +1,247 @@
+# This code is based on the original PR which could be found
+# here. https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1800
+# The idea is to remove this module when the PR or other alternative
+# gets merged into opentelemetry-python-contrib
+
+import urllib
+from aiohttp import web
+from multidict import CIMultiDictProxy
+from timeit import default_timer
+
+from opentelemetry import context, trace, metrics
+from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
+from opentelemetry.instrumentation.utils import http_status_to_status_code
+from opentelemetry.propagators.textmap import Getter
+from opentelemetry.semconv.trace import SpanAttributes
+from opentelemetry.semconv.metrics import MetricInstruments
+from opentelemetry.trace.status import Status, StatusCode
+from opentelemetry.util.http import get_excluded_urls
+from opentelemetry.util.http import remove_url_credentials
+
+from typing import Tuple
+
+
+_SUPPRESS_HTTP_INSTRUMENTATION_KEY = "suppress_http_instrumentation"
+
+_duration_attrs = [
+    SpanAttributes.HTTP_METHOD,
+    SpanAttributes.HTTP_HOST,
+    SpanAttributes.HTTP_SCHEME,
+    SpanAttributes.HTTP_STATUS_CODE,
+    SpanAttributes.HTTP_FLAVOR,
+    SpanAttributes.HTTP_SERVER_NAME,
+    SpanAttributes.NET_HOST_NAME,
+    SpanAttributes.NET_HOST_PORT,
+    SpanAttributes.HTTP_ROUTE,
+]
+
+_active_requests_count_attrs = [
+    SpanAttributes.HTTP_METHOD,
+    SpanAttributes.HTTP_HOST,
+    SpanAttributes.HTTP_SCHEME,
+    SpanAttributes.HTTP_FLAVOR,
+    SpanAttributes.HTTP_SERVER_NAME,
+]
+
+tracer = trace.get_tracer(__name__)
+meter = metrics.get_meter(__name__)
+_excluded_urls = get_excluded_urls("AIOHTTP_SERVER")
+
+
+def _parse_duration_attrs(req_attrs):
+    duration_attrs = {}
+    for attr_key in _duration_attrs:
+        if req_attrs.get(attr_key) is not None:
+            duration_attrs[attr_key] = req_attrs[attr_key]
+    return duration_attrs
+
+
+def _parse_active_request_count_attrs(req_attrs):
+    active_requests_count_attrs = {}
+    for attr_key in _active_requests_count_attrs:
+        if req_attrs.get(attr_key) is not None:
+            active_requests_count_attrs[attr_key] = req_attrs[attr_key]
+    return active_requests_count_attrs
+
+
+def get_default_span_details(request: web.Request) -> Tuple[str, dict]:
+    """Default implementation for get_default_span_details
+    Args:
+        scope: the asgi scope dictionary
+    Returns:
+        a tuple of the span name, and any attributes to attach to the span.
+    """
+    span_name = request.path.strip() or f"HTTP {request.method}"
+    return span_name, {}
+
+
+def _get_view_func(request) -> str:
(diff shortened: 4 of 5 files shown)


bmbouter:
Can a feature be filed for this and a user facing release note added associated with that feature also please? Thanks!

CHANGES/3829.misc
```diff
@@ -0,0 +1 @@
+Adds a vendor version of the OpenTelemetry Aiohttp-server instrumentor to instrument pulp-content app.
```
bmbouter, line 1: Can we put this into user terms? I think that would be `3829.feature` and it would add open telemetry support for the pulp-content app.

bmbouter (approved):
This looks great to me! I agree with vendoring this given the upstream timeline does not seem on the order of days or maybe even weeks.

pulpcore/content/instrumentation.py
```diff
@@ -0,0 +1,247 @@
+# This code is based on the original PR which could be found
+# here. https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1800
+# The idea is to remove this module when the PR or other alternative
+# gets merged into opentelemetry-python-contrib
```
ggainey, line 4: :+1:

ggainey (approved):
Looks good, I'm good w/ us carrying until it appears in an upstream release.

CHANGES/3829.feature
```diff
@@ -0,0 +1 @@
+Add OpenTelemetry support and enable pulp-content app tracing and metrics.
```
lubosmj, line 1:
```suggestion
Implemented OpenTelemetry support and enabled tracing and metrics for the pulp-content app.
```

ipanova (changes requested):
Apologies for bringing in last minutes changes requests, but would it be possible to add docs? This is an important feature pulp is adding and users will not see it a word about it except for this one liner changelog. 
Can at least a docs paragraph be added? In a similar fashion we have mentions about analytics https://github.com/pulp/pulpcore/blob/main/docs/components.rst#analytics-collection

pulpcore/content/instrumentation.py
mdellweg: Let's be honest. Instead of calling this an effective copy of a PR (that may or may not be accepted in the current state), let's own this code, make it officialy a part of Pulp and talk about upstreaming it from there.

mdellweg (changes requested):
Can this work be tested in any reasonable way?

decko:
> Can this work be tested in any reasonable way?

If you're talking about the instrumentation code we have it on the upstream package. I couldn't see a reason to vendor the tests too.
If you're talking about our instance of the aiohttp-server I've found no tests for the authentication middleware or any kind of test that test those pieces together.

Would make sense to have tests for it?

pulpcore/content/instrumentation.py
decko: I don't think it's possible @mdellweg since we're only part of the authors of that code.

CHANGES/3829.feature
decko: Thanks for the suggestion @lubosmj. Way better this way

bmbouter:
> > Can this work be tested in any reasonable way?
> 
> If you're talking about the instrumentation code we have it on the upstream package. I couldn't see a reason to vendor the tests too. If you're talking about our instance of the aiohttp-server I've found no tests for the authentication middleware or any kind of test that test those pieces together.
> 
> Would make sense to have tests for it?

I agree with this. This isn't really Pulp code, it's just being included here so we can start using it because the upstream is very slow to respond. Given that the only actual Pulp part is the part enabling it. We'd add tests here just to then remove them when we remove it in favor of the upstream release also.

decko:
> Apologies for bringing in last minutes changes requests, but would it be possible to add docs? This is an important feature pulp is adding and users will not see it a word about it except for this one liner changelog. Can at least a docs paragraph be added? In a similar fashion we have mentions about analytics https://github.com/pulp/pulpcore/blob/main/docs/components.rst#analytics-collection

I "drafted" a first version of it @ipanova. Can you check and left more suggestions?

CHANGES/3829.feature
```diff
@@ -0,0 +1 @@
+Implemented OpenTelemetry support and enabled tracing and metrics for the pulp-content app.
```
bmbouter, line 1: An additional release note can be added indicating that the docs for open telemetry configuration were added. This can be in the file `CHANGES/3829.doc`.

decko: Done

docs/components.rst
```diff
@@ -142,3 +142,19 @@ An example payload:
...
         "postgresqlVersion": 90200
     }
+
+
+.. _telemetry:
+
+Telemetry Support
+-----------------
+
+New Pulp installations can obtain telemetry data, like the number of requests, active connections and latency response for both `pulp-api` and `pulp-content` using OpenTelemetry. You can read more about `OpenTelemetry here <https://opentelemetry.io>`_.
```
bmbouter, line 152:
```suggestion
Pulp can produces OpenTelemetry data, like the number of requests, active connections and latency response for both `pulp-api` and `pulp-content` using OpenTelemetry. You can read more about `OpenTelemetry here <https://opentelemetry.io>`_.
```

These docs should also be line wrapped at 100 chars.

docs/components.rst
```diff
@@ -142,3 +142,19 @@ An example payload:
...
+
+To enable it you will need to set the following environment variables:
+
+* ``PULP_OTEL_ENABLED`` set to ``True``.
```
bmbouter, line 156: I think this needs to become it's own paragraph expanding on it. What this actually does is run the workers with a different prefix so these docs should give an example of running `pulp-api` and `pulp-content` with and without open telemetry support. Then after explaining that we can state that the `pulp-oci-images` automatically handles that if the `PULP_OTEL_ENABLED` is set.

pulpcore/content/instrumentation.py
dralley: Make sure that if the license requires a copyright / license notice in copies of the code, that we're complying with that.

pulpcore/content/instrumentation.py
```diff
@@ -0,0 +1,247 @@
...
+        scope: the asgi scope dictionary
+    Returns:
+        a tuple of the span name, and any attributes to attach to the span.
+    """
+    span_name = request.path.strip() or f"HTTP {request.method}"
+    return span_name, {}
+
+
+def _get_view_func(request) -> str:
+    """TODO: is this useful??"""
```
dralley, line 79: I assume this is awaiting comment on the upstream PR side?

mdellweg:
> > > Can this work be tested in any reasonable way?
> > 
> > 
> > If you're talking about the instrumentation code we have it on the upstream package. I couldn't see a reason to vendor the tests too. If you're talking about our instance of the aiohttp-server I've found no tests for the authentication middleware or any kind of test that test those pieces together.
> > Would make sense to have tests for it?
> 
> I agree with this. This isn't really Pulp code, it's just being included here so we can start using it because the upstream is very slow to respond. Given that the only actual Pulp part is the part enabling it. We'd add tests here just to then remove them when we remove it in favor of the upstream release also.

Well, it is surely not upstream code. Hence my ask to really own it here. If we do not make it Pulp code, it's nobody's code. Whatever may or may not be merged upstream eventually, nobody can say. And what is not tested, must be considered broken. Adding all of this to pulp is supposed to generate some side effects. If those are not observable, it's not a feature. Maybe for clarification, I am talking about pulp tests, not unittests for the otel library. However, the latter should also come with the code.

pulpcore/content/instrumentation.py
mdellweg: I should probably elaborate: With owning i don't mean claiming the authorship, but taking the responsibility.

bmbouter: re licensing: it's unmerged upstream so I think it's actually not even licensed. When it is merged the upstream will be apache V2.

re owning it: we are owning it since we're contributing it upstream and the vendored version here. The thing I'm trying to avoid is to have it adhere to the standards of the Pulp community when it's actual destination is a contribution to the open-telemetry community which has different standards, norms, and expectations. If there's an issue though, I expect we'll fix that both here and in the upstream for example.

mdellweg: re licensing: It's easy then. The original contribution, which we adopt here, was made under the Apache V2 license. So it is that one.

docs/components.rst
decko: Done

CHANGES/3829.doc
```diff
@@ -0,0 +1 @@
+Adds the Telemetry section to the docs with information about OpenTelemetry and its usage on Pulp.
```
ipanova, line 1: added. keep it past simple

docs/components.rst
```diff
@@ -142,3 +142,33 @@ An example payload:
...
         "postgresqlVersion": 90200
     }
+
+
+.. _telemetry:
+
+Telemetry Support
+-----------------
+
+Pulp can produces OpenTelemetry data, like the number of requests, active connections and latency response for
```
ipanova, line 152: s/like/such as

docs/components.rst
ipanova: @bmbouter is it pulp-oci-images that handles this automatically? that means not only single container  but also operator should be well set here too as well.

docs/components.rst
ipanova, line 152: Pulp can produce

bmbouter:
@ipanova yes it's pulp-oci-images that does it [with this](https://github.com/pulp/pulp-oci-images/commit/2976e33e442bd987d6f2c75eb46a456431fcde91). I also expect that to work correctly for the operator too.

decko:
> Given that there is not an agreement on the testing requirements for this code, I think we should split the docs into their own PR.
> 
> This PR was designed to save time for the upstream to accept and release the work. I believe if we need to write tests also here it will take more time than it saves. We can just let the upstream contribution process work it out and close this PR.

I'm closing this PR until further decision about how we should deal with this kind of situation.
And I'm opening another PR just for the documentation.

mdellweg:
> If you're talking about our instance of the aiohttp-server I've found no tests for the authentication middleware or any kind of test that test those pieces together.

We do test the authentication middleware, because it's purpose is to set the `user` attribute on the `request` object. We test the domain middleware, because without the domain set, pulp would not work properly. So I read that the otel middleware is supposed to generate some data that is to be collected somewhere. Why is it so hard to assert that this happens according to some expectations?

> This PR was designed to save time for the upstream to accept and release the work. I believe if we need to write tests also here it will take more time than it saves. We can just let the upstream contribution process work it out and close this PR.

If closing this PR was meant to prevent writing any tests at all, I must still say functional test that read like a specification for what kind of collected data we expect are still a requirement in my not so humble opinion.

bmbouter:
This is a different situation from those examples. In those examples the features provided by dependencies are consumed by Pulp and because of that they are able to be asserted on via the Pulp APIs. In this case this feature is offered from the opentelemetry dependency and consumed directly by the user. There is no Pulp API that we can actually make an API assertion on.

That's "why its so hard". There isn't a Pulp API that we can call to intercept this data. We'd be spending time and effort adding this capability to the CI and then assertions to a component other than Pulp. Additionally, consider that we are only temporarily carrying this code until it's available upstream. That info together means we'd be doing a lot of work just to throw it away, and that isn't something we can afford to do.

My ask is the same as it was. Think over my claim that this is not "pulp code".  Allow this code to be temporarily carried as-is, and let the actual test contribution happen upstream. [That PR](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1800/files) does have tests. This is the #1 request from a large upstream user and I'd like us to let them test it prior to the upstream accept + release timeline.

PR #3844, opened 2023-05-17
EXPERIMENT: Simplify mirror sync

mdellweg:
Just start with a clean repository version for a mirror sync. This way, there is no need to keep e record of all existing content in memory at any time. This should reduce the memory pressure on the task worker (yet to be confirmed).

[noissue]

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -1,7 +1,6 @@
 """
 Repository related Django models.
 """
-from contextlib import suppress
 from gettext import gettext as _
 from os import path
 from collections import defaultdict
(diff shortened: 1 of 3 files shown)


mdellweg:
Somehow it decreases the memory usage a bit (but I suspect, what we see is that the garbage collection did not even think of starting to do any work...), but at the expense of taking almost twice the time as the original sync. This is clearly not the best idea in it's current form...

Tested with 100 000 content units and stream policy.

plot

dralley:
Also, I think when you scale this up to something like PyPI or dockerhub mirroring, it's be much much worse

PR #3846, opened 2023-05-17
Refactor cast to do only one db call

mdellweg:
[noissue]

### DIFF
--- a/pulpcore/app/models/base.py
+++ b/pulpcore/app/models/base.py
@@ -67,13 +67,24 @@ class BaseModel(LifecycleModel):
 
 class MasterModelMeta(ModelBase):
     def __new__(cls, name, bases, attrs, **kwargs):
-        """Override __new__ to set the default_related_name."""
-        if BaseModel not in bases and MasterModel not in bases:  # Only affects "Detail" models.
-            meta = attrs.get("Meta")
-            default_related_name = getattr(meta, "default_related_name", None)
-            abstract = getattr(meta, "abstract", None)
+        if BaseModel in bases:
+            # This is MasterModel. Do nothing!
+            return super().__new__(cls, name, bases, attrs, **kwargs)
 
-            if not default_related_name and not abstract:
+        meta = attrs.get("Meta")
+        abstract = getattr(meta, "abstract", None)
+        if abstract:
+            # This is an abstract subclass. Do nothing!
+            return super().__new__(cls, name, bases, attrs, **kwargs)
+
+        if MasterModel in bases:
+            # This is a "Master" model. Initialize model map.
+            attrs["_pulp_model_map"] = {}
+        else:
+            # This is a "Detail" model. This is a sanity check only.
+            default_related_name = getattr(meta, "default_related_name", None)
+
+            if not default_related_name:
                 raise Exception(
                     _("The 'default_related_name' option has not been set for {class_name}").format(
                         class_name=name
@@ -81,6 +92,8 @@ class MasterModelMeta(ModelBase):
                 )
 
         new_class = super().__new__(cls, name, bases, attrs, **kwargs)
+        # Register with model map.
+        new_class._pulp_model_map[new_class.get_pulp_type()] = new_class
         return new_class
 
 
@@ -126,6 +139,15 @@ class MasterModel(BaseModel, metaclass=MasterModelMeta):
     class Meta:
         abstract = True
 
+    @classmethod
+    def get_pulp_type(cls):
+        """Get the "pulp_type" string associated with this MasterModel type."""
+        return "{app_label}.{type}".format(app_label=cls._meta.app_label, type=cls.TYPE)
+
+    @classmethod
+    def get_model_for_pulp_type(cls, pulp_type):
+        return cls._pulp_model_map[pulp_type]
+
     def save(self, *args, **kwargs):
         # instances of "detail" models that subclass MasterModel are exposed
         # on instances of MasterModel by the string stored in that model's TYPE attr.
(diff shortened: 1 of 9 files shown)


pulpcore/app/models/base.py
```diff
@@ -126,6 +139,15 @@ class MasterModel(BaseModel, metaclass=MasterModelMeta):
...
         abstract = True
 
+    @classmethod
+    def get_pulp_type(cls):
+        """Get the "pulp_type" string associated with this MasterModel type."""
+        return "{app_label}.{type}".format(app_label=cls._meta.app_label, type=cls.TYPE)
+
+    @classmethod
+    def get_model_for_pulp_type(cls, pulp_type):
+        return cls._pulp_model_map[pulp_type]
```
dralley, line 149: I think it's very likely that the `__new__()` method of every model we care about has been executed already (and thus all models registered) by the point this (or cast) gets executed, but do we actually know for certain?

I suppose if you have an instance of something already which you want the model for, then `__new__()` must have been called for that type during the instantiation.

mdellweg: Are you talking about the `__new__` method of the metaclass? That will run for all classes not for any instance. So yes, I am absolutely sure they will all be initialized.

pulpcore/app/models/base.py
```diff
@@ -67,20 +67,33 @@ def __repr__(self):
...
+
+        meta = attrs.get("Meta")
+        abstract = getattr(meta, "abstract", None)
+        if abstract:
+            # This is an abstract subclass. Do nothing!
+            return super().__new__(cls, name, bases, attrs, **kwargs)
+
+        if MasterModel in bases:
+            # This is a "Master" model. Initialize model map.
+            attrs["_pulp_model_map"] = {}
```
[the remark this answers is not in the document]
mdellweg: Again, this code runs once per class definition. Not at instanciation.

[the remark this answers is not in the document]
dralley: That's not my understanding of the documentation

https://docs.python.org/3/reference/datamodel.html#object.__new__

> Called to create a new instance of class cls. [__new__()](https://docs.python.org/3/reference/datamodel.html#object.__new__) is a static method (special-cased so you need not declare it as such) that takes the class of which an instance was requested as its first argument. The remaining arguments are those passed to the object constructor expression (the call to the class). The return value of [__new__()](https://docs.python.org/3/reference/datamodel.html#object.__new__) should be the new object instance (usually an instance of cls).

>...

> If [__new__()](https://docs.python.org/3/reference/datamodel.html#object.__new__) is invoked during object construction and it returns an instance of cls, then the new instance’s [__init__()](https://docs.python.org/3/reference/datamodel.html#object.__init__) method will be invoked like __init__(self[, ...]), where self is the new instance and the remaining arguments are the same as were passed to the object constructor.

> If [__new__()](https://docs.python.org/3/reference/datamodel.html#object.__new__) does not return an instance of cls, then the new instance’s [__init__()](https://docs.python.org/3/reference/datamodel.html#object.__init__) method will not be invoked.

> [__new__()](https://docs.python.org/3/reference/datamodel.html#object.__new__) is intended mainly to allow subclasses of immutable types (like int, str, or tuple) to customize instance creation. It is also commonly overridden in custom metaclasses in order to customize class creation.

`__new__` creates new instances of objects, and `__init__` initializes them.  `__new__` can be used to customize the creation of instances of a class, but it's still being called at instantiation AFAIK.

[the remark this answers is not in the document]
dralley: Sounds reasonable then

dralley (approved):

pulpcore/app/models/base.py
```diff
@@ -137,32 +159,29 @@ def save(self, *args, **kwargs):
...
-                except AttributeError:
-                    continue
-        else:
-            # The for loop exited normally, there are no more detailed models than this
-            # one in this instance's master/detail ancestry, so return here.
+        if self.pulp_type == self.get_pulp_type():
+            return self
+        result = self._pulp_model_map[self.pulp_type].objects.get(pk=self.pk)
+        # Keep all the prefetched data around.
+        result._state.fields_cache.update(self._state.fields_cache)
```
dralley, line 171: Some unit tests for this would be nice, since casting is relatively fundamental

pulpcore/app/models/base.py
```diff
@@ -137,32 +159,29 @@ def save(self, *args, **kwargs):
...
-        # Go through our related objects, find the one that's a subclass of this model
-        # on a OneToOneField, which identifies it as a potential detail relation.
-        for rel in self._meta.related_objects:
-            if rel.one_to_one and issubclass(rel.related_model, self._meta.model):
-                # The name of this relation is the name of the attr on the model instance.
-                # If that attr as a value, that means a row exists for this model in the
-                # related detail table. Cast and return this value, recursively following
-                # master/detail relationships down to the last table (the most detailed).
-                try:
-                    return getattr(self, rel.name).cast()
```
dralley, line 159: I'm assuming the attribute access of the name of the relation is one query, what was the other?  Is it the one in `get_model_for_pulp_type` that gets amortized away or is it another one?

pulpcore/app/models/base.py
mdellweg: I can at least tell you without this line, we'd get a _lot_ of "sync operation not allowed in async context" errors.
Need to think about how to properly test this.

pulpcore/app/models/base.py
mdellweg: We called cast recursively until we ended at what was presumably the right model and did some guesswork around the related fields on the way.
Now we do a single dict-lookup with the pulp_type field and obtain the desired result in a single step.
For us not having sub-sub classes of master models that often, I think this mainly adds some robustness to the process.

pulpcore/app/models/base.py
mdellweg: Is this what you were looking for?

dralley: Yes, that works for me.  I would maybe throw in a `select_related` too, but it looks good.

ggainey (approved):

PR #3850, opened 2023-05-17
Serialized global reclaim space, global repair artifacts tasks.

### LINKED ISSUE #3786: Global tasks like repair or reclaim should probably not run in parallel
**Version**
Please provide the versions of the pulpcore and plugin packages in use, and how they are installed.  If you are using Pulp via Katello, please provide the Katello version.

**Describe the bug**
Global tasks that involve all repos like repair or reclaim should run one at a time, like orphan cleanup https://github.com/pulp/pulpcore/blob/main/pulpcore/app/viewsets/orphans.py#L29

**To Reproduce**

**Expected behavior**

**Additional context**
Add any other context about the problem here.  Please provide links to any previous discussions via Discourse or Bugzilla.

ipanova:
closes #3786

### DIFF
--- a/CHANGES/3786.bugfix
+++ b/CHANGES/3786.bugfix
@@ -1,0 +1,2 @@
+Serialized global reclaim space, global repair artifacts tasks with respect to each other
+to prevent them from failing.
--- a/pulpcore/app/views/repair.py
+++ b/pulpcore/app/views/repair.py
@@ -1,4 +1,5 @@
 from drf_spectacular.utils import extend_schema
+from django.conf import settings
 from rest_framework.views import APIView
 
 from pulpcore.app.response import OperationPostponedResponse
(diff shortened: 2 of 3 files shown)


gerrod3 (approved):

mdellweg:
I thought, we wanted to prevent repair and reclaim to run in parallel. Would it make sense to take orphan cleanup into the mix, and ensure that only one of these dangerous tasks runs at a time?

dralley:
You mean, by locking on some shared "global" resource (generic name) as opposed to individual global locks for repair, orphans, & reclaim?

mdellweg:
> You mean, by locking on some shared "global" resource (generic name) as opposed to individual global locks for repair, orphans, & reclaim?

actually a global exclusive resource like `/pulp/api/dangerous/file/operation/`. Maybe that would be a bit much, but currently you can run repair and reclaim (which are kind of reciprocal to each other) concurrently.

ipanova:
@mdellweg that goal of this PR was to not allow running 2 reclaim or 2 repairs in parallel.
Reclaim and repair are not exactly reciprocal, Reclaim deletes even the artifact row in DB , while repair task is focused on the reparation on broken artifacts that are present in db.

https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/reclaim_space.py#L62
https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/repository.py#L104

ipanova:
But, I am not excluding possibility of a scenario that could cause issues for these tasks running in parallel. i.e these 2  tasks 
are  ran on the same repo A and content B that also happens to be 'broken' aka bitrot.

In any case, can we wait when we cross this brings with a real issue happening in our hands?

PR #3851, opened 2023-05-17
Taught repair task to continue when encountering a 404 during the rep…

### LINKED ISSUE #3611: Pulp repair functionality should try to continue when encountering a 404 during the repair process
**Version**
3.21 but probably all

**Describe the bug**

>If the satellite\capsule servers are missing artifacts related to both rpm and docker type repos, Then The Repair API, when executed, may not be able to repair all the rpm data and stop it's action if it fails to process any docker manifest-related artifacts.

I wouldn't think this is specific to docker + rpm though, I assume any failure would trigger it.  Essentially, we should be tolerant of failure during the repair process and try as best we can to fix as much as possible.  Instead, what currently happens is that the repair fails on a 404

See BZ for further details  

**To Reproduce**
See BZ

**Expected behavior**
The repair functionality should continue after unsuccessfully downloading a file during a repair, and print the errors.

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=2173692

ipanova:
…air process

closes #3611

### DIFF
--- a/pulpcore/app/management/commands/handle-artifact-checksums.py
+++ b/pulpcore/app/management/commands/handle-artifact-checksums.py
@@ -2,6 +2,8 @@ import os
 
 from gettext import gettext as _
 
+from aiohttp.client_exceptions import ClientResponseError
+
 from django.conf import settings
 from django.core.management import BaseCommand, CommandError
 from django.db.models import Q, Sum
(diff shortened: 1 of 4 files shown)


pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -122,15 +124,23 @@ def _download_artifact(self, artifact, checksum, file_path):
...
+                        else:
+                            if dl_result.artifact_attributes["sha256"] == artifact.sha256:
+                                with open(dl_result.path, "rb") as src:
+                                    filename = artifact.file.name
+                                    artifact.file.delete(save=False)
+                                    artifact.file.save(filename, src, save=False)
+                                setattr(artifact, checksum, dl_result.artifact_attributes[checksum])
+                                restored = True
+                                break
+                self.stdout.write(_("Deleting file for the unreparable file {}".format(file_path)))
```
mdellweg, line 142:
```suggestion
                self.stdout.write(_("Deleting unreparable file {}".format(file_path)))
```
Maybe you want to add the broken artifact href...

mdellweg (approved):
Could you make the first line of the commit message fit 50 chars?

dralley:
After merging (and backporting) please update https://bugzilla.redhat.com/show_bug.cgi?id=2173692 with fixed-in-version releases

dralley (approved):

PR #3852, opened 2023-05-17
Fix race condition in parallel domains test

gerrod3:
I noticed this in https://github.com/pulp/pulpcore/pull/3847, but the domain tests are done in parallel and this one test was sometimes selecting domains created in other tests that could be deleted before the test completes.


dralley (approved):

mdellweg (approved):

mdellweg:
Do we need to backport this?

PR #3883, opened 2023-05-25
Properly shutdown the aiohttp test server

mdellweg:
Without this we get a lot of warnings about unfinished coroutines and also loose a second per test waiting for the signal to propagate.

[noissue]


bmbouter (approved):
This is an elegant PR. Thank you for making it!

decko (approved):
LTTM :rocket:

Issue #3236, opened 2022-09-21
Pulp export chunks are seen by suspicious by virus scanners because tar output is piped into split


dralley:
**Version**
All

**Describe the bug**
See attached BZ for full details.

The gzipped tarball is piped into the split command to create the chunks. Every chunk ( or split ) after the first one does not have file header data to indicate the type of file, so the virus scanner sees it as random data.

When creating a chunked tar archive on my local machine and piping it through "split", only the first chunk is recognised with the file command:

```
woody@woody-arch-thinkpad ~/Documents $ tar -cvzf - file1.mp4 file2.mp4 | split -b 2MB - test.tar.gz.
woody@woody-arch-thinkpad ~/Documents $ file test.tar.gz.aa 
test.tar.gz.aa: gzip compressed data, from Unix, original size modulo 2^32 842674519
woody@woody-arch-thinkpad ~/Documents $ file test.tar.gz.ab
test.tar.gz.ab: data
```

If I use the "--tape-length" option, I can use the "file" command to identify each chunk as a tarball. I assume the virus scanning tool would be using "file" or similar methods.

```
woody@woody-arch-thinkpad ~/Documents $ tar -cv --tape-length=2000 --file=test-archive-{0..100}.tar file1.mp4 file2.mp4
woody@woody-arch-thinkpad ~/Documents $ file test-archive-0.tar 
test-archive-0.tar: POSIX tar archive (GNU)
woody@woody-arch-thinkpad ~/Documents $ file test-archive-1.tar 
test-archive-1.tar: tar archive
```

A drawback is that those "--tape-length"  archives can't be gzipped from that single tar command.  It gives an error
"tar: Cannot use multi-volume compressed archives"

Can we request that the exports are chunked with "--tape-length" instead of piping into the "split" command?

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=2109742

PR #3884, opened 2023-05-25
Use a more appropriate compression level for exports

### LINKED ISSUE #3869: Exports are bottlenecked by gzip compression which cannot be disabled
**Version**
Any

**Describe the bug**

> 
> We identified a severe bottleneck in the way hammer exports work and found the code in upstream Pulp.
> 
> Line 406 of https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/export.py
> 
> "with tarfile.open(tarfile_fp, "w|gz", fileobj=split_process.stdin)"
> 
> The tarfile Python module creates tar files using the gizp Python module.
> 
> Data compression for the gzip Python module is provided by the zlib Python module.
> 
> The zlib Python module calls the zlib library.
> 
> If defaults are used the whole way through this series of events, the result is a single threaded pulp process doing compression of a tarball containing a massive content library. This bottleneck and can make large hammer exports take several days.
> 
> Modifying the lines that tell the tarfile.open function to NOT use compression ( change "w|gz" to "w" ) dramatically speeds up the hammer export. In our testing it reduced the time from days to just hours. The drawback is the file size was significantly larger, but the trade-off is worthwhile given we have tight timeframes and plentiful disk capacity. 
> 
> Can this bottleneck be addressed with multi-threaded gzip compression?
> 
> and/or
> 
> Can a hammer command line option for no compression be implemented?
> 
> Run a hammer export and monitor Pulp processes. One process with run at 100% CPU. Modify the abovementioned Python script to NOT use gzip encryption, and an uncompressed tarball will be created instead much quicker and with multiple Pulp processes. 
> 
> 
> Steps to Reproduce:
> 1. Run a "hammer export". Monitor the Pulp process CPU usage and time taken to complete export.
> 2. Change the abovementioned Python code in Pulp.
> 3. Run a "hammer export" again and note performance improvement.

**Expected behavior**

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=2188504

dralley:
Exports will be larger, but should be much faster. There were reports of large exports taking multiple days to complete due to the overhead incurred by compression.

closes #3869

### DIFF
--- a/CHANGES/3869.bugfix
+++ b/CHANGES/3869.bugfix
@@ -1,0 +1,3 @@
+Exports now use gzip compression level 1 rather than compression level 9. Exported archives will
+now be slightly larger, but exports should be much faster. This is considered to be a more
+optimal balance of space/time for the export operation.
--- a/pulpcore/app/monkeypatch.py
+++ b/pulpcore/app/monkeypatch.py
@@ -1,0 +1,66 @@
+import sys
+
+# This is a monkeypatch for https://github.com/pulp/pulpcore/issues/3869
+if sys.version_info.major == 3 and sys.version_info.minor < 12:
+    # Code copied from the Python 3.12 standard library
+    # We modify the default gzip compression level for writing streams from
+    # 9 to 1, attempting to vendor the minimum amount of code.
+    # -------------------------------------------------------------------
+    # tarfile.py
+    # -------------------------------------------------------------------
+    # Copyright (C) 2002 Lars Gustaebel <lars@gustaebel.de>
+    # All rights reserved.
+    #
+    # Permission  is  hereby granted,  free  of charge,  to  any person
+    # obtaining a  copy of  this software  and associated documentation
+    # files  (the  "Software"),  to   deal  in  the  Software   without
+    # restriction,  including  without limitation  the  rights to  use,
+    # copy, modify, merge, publish, distribute, sublicense, and/or sell
+    # copies  of  the  Software,  and to  permit  persons  to  whom the
+    # Software  is  furnished  to  do  so,  subject  to  the  following
+    # conditions:
+    #
+    # The above copyright  notice and this  permission notice shall  be
+    # included in all copies or substantial portions of the Software.
+    #
+    # THE SOFTWARE IS PROVIDED "AS  IS", WITHOUT WARRANTY OF ANY  KIND,
+    # EXPRESS OR IMPLIED, INCLUDING  BUT NOT LIMITED TO  THE WARRANTIES
+    # OF  MERCHANTABILITY,  FITNESS   FOR  A  PARTICULAR   PURPOSE  AND
+    # NONINFRINGEMENT.  IN  NO  EVENT SHALL  THE  AUTHORS  OR COPYRIGHT
+    # HOLDERS  BE LIABLE  FOR ANY  CLAIM, DAMAGES  OR OTHER  LIABILITY,
+    # WHETHER  IN AN  ACTION OF  CONTRACT, TORT  OR OTHERWISE,  ARISING
+    # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+    # OTHER DEALINGS IN THE SOFTWARE.
+
+    import tarfile
+    from tarfile import NUL
+    import struct
+    import os
+    import time
+
+    class _Stream(tarfile._Stream):
+        """Class that serves as an adapter between TarFile and
+        a stream-like object.  The stream-like object only
+        needs to have a read() or write() method and is accessed
+        blockwise.  Use of gzip or bzip2 compression is possible.
+        A stream-like object could be for example: sys.stdin,
+        sys.stdout, a socket, a tape device etc.
+
+        _Stream is intended to be used only internally.
+        """
+
+        def _init_write_gz(self):
+            """Initialize for writing with gzip compression."""
+            self.cmp = self.zlib.compressobj(
+                1, self.zlib.DEFLATED, -self.zlib.MAX_WBITS, self.zlib.DEF_MEM_LEVEL, 0
+            )
+            timestamp = struct.pack("<L", int(time.time()))
+            self.__write(b"\037\213\010\010" + timestamp + b"\002\377")
+            if self.name.endswith(".gz"):
+                self.name = self.name[:-3]
+            # Honor "directory components removed" from RFC1952
+            self.name = os.path.basename(self.name)
+            # RFC1952 says we must use ISO-8859-1 for the FNAME field.
+            self.__write(self.name.encode("iso-8859-1", "replace") + NUL)
+
+    tarfile._Stream = _Stream
--- a/pulpcore/app/tasks/export.py
+++ b/pulpcore/app/tasks/export.py
@@ -5,7 +5,6 @@ import os
 import os.path
 import subprocess
 import tarfile
-
 from distutils.util import strtobool
 from gettext import gettext as _
 from glob import glob


ipanova (approved):
I like the solution :seal:

ipanova:
bit it seems like the tests is failing `E           pulp_smash.pulp3.bindings.PulpTaskError: (PulpTaskError(...), "Pulp task failed (__init__() got an unexpected keyword argument 'compresslevel')")`

ipanova:
@dralley did you ran any tests to come to the conclusion that compression level 1 is the most balanced? there is the whole range of 1-9 levels

ggainey:
> bit it seems like the tests is failing `E pulp_smash.pulp3.bindings.PulpTaskError: (PulpTaskError(...), "Pulp task failed (__init__() got an unexpected keyword argument 'compresslevel')")`

Oh bah - looks like compresslevel isn't supported on the streaming options. From the tarfile doc:

"For modes 'w:gz', 'r:gz', 'w:bz2', 'r:bz2', 'x:gz', 'x:bz2', [tarfile.open()](https://docs.python.org/3/library/tarfile.html#tarfile.open) accepts the keyword argument compresslevel (default 9) to specify the compression level of the file."

dralley:
I checked https://github.com/python/cpython/blob/main/Lib/tarfile.py#L1826 first, it looks like it should be supported...  

It looks like this was only merged a year ago and therefore might not be in the Python versions we're using https://github.com/python/cpython/commit/50cd4b6959568999f5e426e58050ea912a490ac6

ipanova:
> > @dralley did you ran any tests to come to the conclusion that compression level 1 is the most balanced? there is the whole range of 1-9 levels
> 
> I didn't run tests but I did look over some general compression benchmarks that others have done. Level 3 was 30-40% slower but only about 10% smaller, and anything beyond that was an even worse tradeoff. My feeling is that a lot of the content that people want to export like RPMs, metadata, etc. are already compressed so we don't need to try hard to compress it again.

Gotcha, thanks for explanation.  It makes sense.

ggainey (commented):
Daniel, might it be possible to *subclass* "just enough" of Tarfile to support this, and carry just that? As opposed to needing all 3K lines of tarfile.py?

dralley:
@ggainey I agree and that is the intention - this is still in draft remember - but until then I just wanted a sense of whether it would work or not.

dralley:
Well, it works.  However, is this something we're actually open to doing?  We have explicitly tried to avoid doing monkeypatches or this kind of vendoring in the past.  I think this solution would yield the best outcome (relative to no compression at all, or something like that) but I still definitely feel discomfort about it.

ggainey (commented):
I think this approach is the only reasonable way to address the problem for the actually-affected users. It's definitely "black magic", and needs some Very Explicit Comments to make it clear what's going on and why it works - but I can't think of a better approach.

dralley:
Ready for re-review

pulpcore/app/monkeypatch.py
```diff
@@ -0,0 +1,69 @@
...
+        needs to have a read() or write() method and is accessed
+        blockwise.  Use of gzip or bzip2 compression is possible.
+        A stream-like object could be for example: sys.stdin,
+        sys.stdout, a socket, a tape device etc.
+
+        _Stream is intended to be used only internally.
+        """
+
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
```
mdellweg, line 53: Currently this doesn't do a thing.

pulpcore/app/monkeypatch.py
```diff
@@ -0,0 +1,69 @@
+import sys
```
mdellweg, line 1: I believe, in order to make this work, you somehow need to import this file somewhere.
For transparency, I would reexport tarfile from this file and use it from here.

mdellweg:
I do have yet another idea: When looking at the export code, we use the python tarfile to stream into a "split" subprocess. I think we can rewrite the process to add a call to bz2/gz/... there. Then we'd even have a better separation of concerns. WDYT?

ggainey:
> I do have yet another idea: When looking at the export code, we use the python tarfile to stream into a "split" subprocess. I think we can rewrite the process to add a call to bz2/gz/... there. Then we'd even have a better separation of concerns. WDYT?

There are two branches, one that chunks (uses split) and one that streams to a single file. We need this in both, and we need to insure that the output-file never hits disk uncompressed along the way.

mdellweg:
> > I do have yet another idea: When looking at the export code, we use the python tarfile to stream into a "split" subprocess. I think we can rewrite the process to add a call to bz2/gz/... there. Then we'd even have a better separation of concerns. WDYT?
> 
> There are two branches, one that chunks (uses split) and one that streams to a single file. We need this in both, and we need to insure that the output-file never hits disk uncompressed along the way.

What I'm trying to say is:
In the end, we need a python `TarFile` object to push our exports into. But that part of the code does not care how the data gets to disk. On the other hand, we have some control over how this `TarFile` object is set up. I think, we could always set the tarfile up to dump uncompressed into a stream. The resulting stream, we should be able to stream through a compression pipeline that then (optionally) is streamed into split or to disk directly. Think like `tar -c -f - | gz > out.tar.gz`, just in python.

ggainey:
> The resulting stream, we should be able to stream through a compression pipeline that then (optionally) is streamed into split or to disk directly. Think like `tar -c -f - | gz > out.tar.gz`, just in python.

Ah, I get it now. Um, that might do it, as long as we do the same thing in both branches - and it would alleviate the need to carry a monkeypatch (assuming the flags we need

mdellweg:
> Ah, I get it now. Um, that might do it, as long as we do the same thing in both branches - and it would alleviate the need to carry a monkeypatch (assuming the flags we need
And give us complete control over the compression pipeline (algorithm, compresslevel, ...).

dralley:
>I do have yet another idea: When looking at the export code, we use the python tarfile to stream into a "split" subprocess. I think we can rewrite the process to add a call to bz2/gz/... there. Then we'd even have a better separation of concerns. WDYT?

Wouldn't this be the opposite order of operations?  That is, whereas currently we write a gzip-compressed tarfile, this suggestion would tar a series of gzip-compressed files.

pulpcore/app/monkeypatch.py
dralley: I don't think so, so long as it's on a path reachable by `__init__` it ought to be picked up.

pulpcore/app/monkeypatch.py
dralley: It doesn't need to

dralley:
>I believe tar | gz | split would work - but I haven't tried it to know Absolutely For Sure. It's definitely worth the experiment. The important thing, tho, is "whatever we do in export, needs to be consumable in the current import code", so people don't have to recreate their exports.

Experimenting with this has been successful, although expressing it in code is a bit more challenging, but I want to double check -- is `gzip`-the-command-line-tool available on all systems we care about?  Most distros have it installed by default, but a lot of container images do not.

dralley:
I take it back, something about the data pipeline gets messed up.  Combined with uncertainty about relying on the `gzip` cli tool, I think we should just continue with the original implementation.

pulpcore/app/monkeypatch.py
mdellweg: Exactly if we do not add any init code, then just don't add these two lines.

CHANGES/3869.bugfix
```diff
@@ -0,0 +1 @@
+Exports now use gzip compression level 1 rather than compression level 9. Exported archives will now be slightly larger, but exports should be much faster. This is considered to be a more optimal balance of space/time for the export operation.
```
mdellweg, line 1: To make some of our teammates happy, this ought to be line broken at 98 characters.

(100 is the rule, but towncrier indents by another two.)

pulpcore/app/monkeypatch.py
mdellweg: That is not how i learned python, but maybe django does some auto lookup?

dralley: Let's say you have a Python module, with an `__init__.py` (an empty one).  At load time every .py file within that directory will be compiled to bytecode and loaded.  As far as I'm aware Python doesn't do any lazy loading.

But I can find out for sure by checking the compression level of a test export.

mdellweg (approved):

ggainey (approved):

ipanova (approved):

B-Woody:
I submitted the original Red Hat BZ relating to this issue and also for #3236  . Passing on my thanks to everyone who worked on this. It's going to be a huge help for some of the work we're doing.

PR #3886, opened 2023-05-26
Taught PulpImport to recover from previously-failed attempt.

### LINKED ISSUE #3737: Can't rerun a failed content-import task if it was exported using chunks
**Version**

All?

**Describe the bug**

When importing content that was exported in chunks, the importer process concatenate the chunks into a single file in order to import.

If that import task fails for some reason, after the chunks were already combined into a single file, the user can't simply re-run the same command to retry the import. Satellite will complain that the chunks are missing.

Checking the directory, we can see that the chunks are gone and only the master file is present (together with metadata.json and TOC file).

**To Reproduce**

Steps to Reproduce:
1. Export something using chunks
2. Start the import on another satellite. Monitor the data directory until all the chunk files are gone and only the main file is present. At this moment, kill the pulpcore-worker which is processing the import. Wait until the task returns with error.
3. Repeat the same import command. Pulp will error out with an error like this:

~~~
Feb 27 16:38:56 reproducer-import pulpcore-worker-3[156191]: pulp [6c91f855-9959-43b1-864f-925393ae025a]: pulpcore.tasking.pulpcore_worker:INFO: Task ddf45eb1-adea-480e-9114-748baa7bd7bf failed ([ErrorDetail(string="Missing import-chunks named in table-of-contents: ['export-fc4273a4-6320-4fd5-98c9-6adfe9461781-20230227_2132.tar.gz.0000', 'export-fc4273a4-6320-4fd5-98c9-6adfe9461781-20230227_2132.tar.gz.0001', 'export-fc4273a4-6320-4fd5-98c9-6adfe9461781-20230227_2132.tar.gz.0002'].", code='invalid')])
~~~

Actual results:
Re-running same import fails, complaining about chunks missing. User needs to modify the TOC file manually OR split the master file in chunks again OR copy the files again in order to run the import.

**Expected behavior**

Pulp could be smart enough to identify either all the chunks are present OR the global file file. If the checksum matches, move forward with the import.

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=2173757

ggainey:
If an expected-reassembled-filename exists, ignore chunk-info and proceed with that file.

Test-coverage will be added in pulp_file once this is in a released version of pulpcore.

fixes #3737.
[nocoverage]

### DIFF
--- a/CHANGES/3737.bugfix
+++ b/CHANGES/3737.bugfix
@@ -1,0 +1,1 @@
+Taught PulpImport to be more robust in the face of previous failed attempts.
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -335,6 +335,18 @@ def pulp_import(importer_pk, path, toc, create_repositories):
                 raise ValidationError(_("Missing 'files' or 'meta' keys in table-of-contents!"))
 
             base_dir = os.path.dirname(toc_filename)
+
+            # Regardless of what the TOC says, it's possible for a previous import to have
+            # failed after successfully creating the combined file. If the TOC specifies multiple
+            # chunks, but the "expected result" exists, ignore the chunk-list and process as if
+            # it's all there ever was.
+            top_level_file = os.path.join(base_dir, the_toc["meta"]["file"])
+            if len(the_toc["files"]) > 1 and os.path.isfile(top_level_file):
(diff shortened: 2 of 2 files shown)


dralley (approved):
This looks OK to me, but I'm not very familiar with the import/export machinery, so that's based on limited context.

PR #3893, opened 2023-05-31
Extensions to give plugins a chance to set specific content-headers.

ggainey:
Added Distribution.content_headers_for() and call it from inside Handler.response_headers().

This is in support of https://github.com/pulp/pulp_rpm/issues/2947

[noissue]

### DIFF
--- a/pulpcore/app/models/publication.py
+++ b/pulpcore/app/models/publication.py
@@ -574,6 +574,17 @@ class Distribution(MasterModel):
         """
         return set()
 
+    def content_headers_for(self, path):
+        """
+        Opportunity for Distribution to specify response-headers for a specific path
+
+        Args:
+            path (str): The path being requested
+        Returns:
+            Empty dictionary, or a dictionary with HTTP Response header/value pairs.
+        """
+        return {}
+
     @hook(BEFORE_DELETE)
     @hook(
         AFTER_UPDATE,
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+from multidict import CIMultiDict
 import os
 import re
 from gettext import gettext as _
@@ -345,20 +346,29 @@ class Handler:
         return True
 
     @staticmethod
-    def response_headers(path):
+    def response_headers(path, distribution=None):
         """
         Get the Content-Type and Encoding-Type headers for the requested `path`.
 
         Args:
             path (str): The relative path that was requested.
-
+            distribution(Distribution) : Distribution detail that might want to add headers for path
         Returns:
             headers (dict): A dictionary of response headers.
         """
+        # headers are case-insensitive
+        headers = CIMultiDict({})
+
+        # Determine a content-type from mime_types and set.
+        # Note: plugin-Distribution can override this.
         content_type = mime_types.get_type(path)
-        headers = {}
         if content_type:
             headers["Content-Type"] = content_type
+
+        # Let plugin-Distribution set headers for this path if it wants.
+        if distribution:
+            headers.update(distribution.content_headers_for(path))
+
         return headers
 
     @staticmethod
@@ -537,7 +547,7 @@ class Handler:
         if content_handler_result is not None:
             return content_handler_result
 
-        headers = self.response_headers(rel_path)
+        headers = self.response_headers(rel_path, distro)
 
         repository = distro.repository
         publication = distro.publication
@@ -569,7 +579,7 @@ class Handler:
                     await publication.published_artifact.aget(relative_path=index_path)
 
                     rel_path = index_path
-                    headers = self.response_headers(rel_path)
+                    headers = self.response_headers(rel_path, distro)
                 except ObjectDoesNotExist:
                     dir_list, dates, sizes = await self.list_directory(None, publication, rel_path)
                     dir_list.update(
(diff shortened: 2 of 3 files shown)


pulpcore/app/models/publication.py
```diff
@@ -574,6 +574,17 @@ def content_handler_list_directory(self, rel_path):
         """
         return set()
 
+    def content_headers_for(self, path):
```
ggainey, line 577: Distribution "feels like" the right place to know-things about specific files being served.

pulpcore/content/handler.py
```diff
@@ -344,20 +344,24 @@ def _permit(request, distribution):
...
     @staticmethod
-    def response_headers(path):
+    def response_headers(path, distribution=None):
         """
         Get the Content-Type and Encoding-Type headers for the requested `path`.
 
         Args:
             path (str): The relative path that was requested.
-
+            distribution(Distribution) : Disitrbution detail that might want to add headers for path
```
mdellweg, line 353:
```suggestion
            distribution(Distribution) : Distribution detail that might want to add headers for path
```

pulpcore/content/handler.py
```diff
@@ -536,7 +540,7 @@ async def _match_and_stream(self, path, request):
         if content_handler_result is not None:
             return content_handler_result
 
-        headers = self.response_headers(rel_path)
+        headers = self.response_headers(rel_path, distro)
```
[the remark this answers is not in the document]
mdellweg: Are the headers supposed to be dependent on the content identified?
If `content_handler_result` was populated from the plugin 3 lines up, do we expect the headers to be there already?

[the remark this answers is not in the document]
ggainey: Yes - look at the issue, the specific problem is "repomd.xml needs to not-cache". Headers specific to specific files is pretty common (and if you want to set them for Everything, you do that in the webserver configs)

content_handler_result is the output of the current protocol, where the handler says "if you give me a result back I am not touching it - you (the plugin) did what needed doing". I would actually have preferred to do this "inside" that existing protocol, but there wasn't any way to take advantage of all of the machinery that happens next without just doing a giant copy/paste into the plugin.

conten-handling needs a serious refactor - but this ask isn't the place to do it.

CHANGES/3897.feature
```diff
@@ -0,0 +1,5 @@
+Added Distribution.content_headers-for() to let plugins affect content-app responses.
```
mdellweg, line 1: I think this belongs in CHANGES/plugin ...

pulpcore/content/handler.py
```diff
@@ -344,20 +344,24 @@ def _permit(request, distribution):
...
         Returns:
             headers (dict): A dictionary of response headers.
         """
         content_type = mime_types.get_type(path)
         headers = {}
         if content_type:
             headers["Content-Type"] = content_type
+
+        if distribution:
+            headers.update(distribution.content_headers_for(path))
```
dralley, line 363: Do we need to be concerned with casing?

dralley (approved):
This is fairly reasonable - I agree that we need to overhaul content serving completely but yes, that's out of scope.

pulpcore/content/handler.py
mdellweg: Assuming, you meant casting, it is already the detail model:
https://github.com/pulp/pulpcore/pull/3893/files#diff-453fdf3a6e81f92ff5391cf65f9c6154852133a656d82082021e3beaa887776fR300

dralley:
No casing as in lowercase.  Updating sets with potentially different-but-the-same keys.

pulpcore/content/handler.py
ggainey: This is a good point, that also uncovers "what if the Distribution wants to return Content-Type"? Let me make some adjustments here.

dralley: Copying my comment since github mobile put it in the wrong place:

> No casing as in lowercase. Updating sets with potentially different-but-the-same keys.

pulpcore/content/handler.py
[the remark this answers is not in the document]
ipanova: What about this code path? https://github.com/pulp/pulpcore/blob/a18d5b861011de33b7305a22d267e8f32e48f350/pulpcore/content/handler.py#L578

pulpcore/content/handler.py
ipanova: HTTP headers names are case insensitive, aiohttps responses like File/StreamResponse use CIMultiDict which take care exactly of this problem so we can just ignore this.
The only problematic place I see is here https://github.com/pulp/pulpcore/blob/main/pulpcore/content/handler.py#L869. We could just CIMultiDict instead of casefolding.

ipanova:
```
In [1]: from multidict import CIMultiDict

In [2]: x=CIMultiDict()

In [3]: x['Ina']=1

In [4]: x['ina']=2

In [5]: x
Out[5]: <CIMultiDict('ina': 2)>

```

ipanova (changes requested):
please check my comments :)

pulpcore/content/handler.py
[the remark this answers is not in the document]
ggainey: That's the "get  index.html" (only) path; I guess I "assumed" nobody would care about it. You are correct, tho, there's no reason to **prevent** a plugin from having some control there as well.

ipanova (approved):

ipanova (changes requested):
per our matrix chat we've decided to use `CIMultiDict`

ipanova (approved):

PR #3913, opened 2023-06-07
Removes systemd from the documentation.

### LINKED ISSUE #3912: Remove references to systemd altogether
systemd is no longer helpful in documenting for our users, here's why:

1) Users running standalone containers can just use podman or docker to start/stop the container. They can also use `podman generate systemd --new --files --name pulp` to have those tools generate systemd containers

2) Users running docker/podman compose has the compose tooling handle the start/stop/restart states of the containers

3) Users runnning k8s, openshift, or the operator don't use systemd

4) Users doing manual pip installation may have any number of ways to manage their services start/stop/restart. It's really not a Pulp specific thing.

bmbouter:
closes #3912


decko (approved):

bmbouter:
+1 to getting rid of those links. The templates themselves are jinja2 so I don't think we should copy them as-is into another repo. This is more or less what stopped me from resolving this. I think the options are:

a) port the jinja2 snippets to be systemd unit files and put them as snippets directly in the docs or
b) stop documenting systemd unit files

mikedep333:
@ipanova We plan to add sample systemd snippets to the docs. For the time being, we are OK with the links being removed.

ipanova:
I filed https://github.com/pulp/pulpcore/issues/3938

PR #3924, opened 2023-06-14
Taught monitor_task_group fixture to respond better to fatal errors.

### LINKED ISSUE #3923: monitor_task_group() fixture can get stuck in an infinite wait
**Version**
main

**Describe the bug**
If a task-group fails catastrophically before it can dispatch any tasks, then the monitor_task_group() will enter the "wait for completion" loop and never come out. `task_group.all_tasks_dispatched` starts as False, and is never set to True, so this check:

https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/__init__.py#L817

is always True.

**To Reproduce**

This PR https://github.com/pulp/pulp_file/pull/969 puts the lowerbounds CI test into an infinite wait, because the fix it's testing for isn't backported to core/3.23, and fails as described above, which causes the loop to spin forever, which causes lowerbounds to never complete.

**Expected behavior**
monitor_task_group() needs to recognize catastrophic-failure a little better.

ggainey:
fixes #3923.

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -812,12 +812,20 @@ def monitor_task(tasks_api_client, pulp_domain_enabled):
 @pytest.fixture(scope="session")
 def monitor_task_group(task_groups_api_client):
     def _monitor_task_group(task_group_href):
+        def tasks_open(tg):
+            return (tg.waiting + tg.running + tg.canceling) > 0
+
         task_group = task_groups_api_client.read(task_group_href)
 
-        while not task_group.all_tasks_dispatched or (task_group.waiting + task_group.running) > 0:
+        # Wait until there are no tasks left to run
+        # Note: if the *tasking system* has a bug that leaves a task (somehow)
+        # in run/wait/canceling, we're going to be stuck in this loop forever.
+        # I don't see a reliable way to avoid that. TaskingSystemBugs: Just Say No! :)
+        while tasks_open(task_group):
             sleep(SLEEP_TIME)
             task_group = task_groups_api_client.read(task_group_href)
 
+        # If ANYTHING went wrong, throw an error
         if (task_group.failed + task_group.skipped + task_group.canceled) > 0:
             raise PulpTaskGroupError(task_group=task_group)
 
--- a/CHANGES/3923.misc
+++ b/CHANGES/3923.misc
@@ -1,0 +1,1 @@
+Fixed possible infinite-wait in monitor_task_group fixture.


pulpcore/tests/functional/__init__.py
```diff
@@ -812,9 +812,18 @@ def _monitor_task(task_href):
...
 def monitor_task_group(task_groups_api_client):
     def _monitor_task_group(task_group_href):
+        def tasks_running(tg):
+            return (tg.waiting + tg.running) > 0
+
+        def group_wait(tg):
+            return not tg.all_tasks_dispatched or tasks_running(tg)
+
+        def group_fail(tg):
+            return not tg.all_tasks_dispatched and tg.failed > 0 and not tasks_running(tg)
```
mdellweg, line 822:
```suggestion
            return tg.failed > 0 and not tasks_running(tg)
```
Would this also be a fail condition?

mdellweg (approved):
Think, we can take it as it is. But the logic may be a bit twisted.

pulpcore/tests/functional/__init__.py
lubosmj: Still, I do not see any reason for checking the presence of running tasks. If some tasks fail, we can consider the group failed regardless of the running tasks.

mdellweg: As long as a single sub-task is running, the taskgroup is active. We may already know something did not succeed, but what the tests really care about is that the taskgroup is in a final state.
Maybe we should adjust the names of the functions to better reflect this.

ggainey:
This fell off my radar, apologies for the long lag. In addition to taking @mdellweg 's suggestion, renamed tasks_running() to tasks_open() - since it's "running or waiting", seemed less misleading this way.

mdellweg:
I'm fine with this (especially since it is a test fixture and can point to tasks not calling `tg.finish()`), but in theory, i believe we can ignore `all_tasks_dispatched`, since only as long as a task in here is still running/waiting ("pending"?) new tasks can be added to the task group. And in the case of ACS refresh (where the task_group is filled by view), we are only handed the reference to that task_group once the view called `finished`. So the window, where the task_group could actually looks like all tasks being finished, in the api is before we know about it and that it would be worth looking.

pulpcore/tests/functional/__init__.py
```diff
@@ -812,9 +812,18 @@ def _monitor_task(task_href):
...
 def monitor_task_group(task_groups_api_client):
     def _monitor_task_group(task_group_href):
+        def tasks_open(tg):
+            return (tg.waiting + tg.running) > 0
+
+        def group_wait(tg):
+            return not tg.all_tasks_dispatched or tasks_open(tg)
+
+        def group_fail(tg):
+            return tg.failed > 0 and not tasks_open(tg)
```
mdellweg, line 822: Now that i think of it, yes, we know that the group has failed, but something may still add tasks (the view in this case) to it. So we should really only care about failed once task_open (maybe call that task_pending) turned `False`.
And a task that fails to call `finish` because it failed itself is a bug? Feel free to disagree.

ggainey: This PR exists because a lowerbounds test got into a situation where is-dispatched was never set to true due to a problem, and went into an infinite-wait as a result. Was that problem a bug? Yes. Should we find out because a test simply doesn't complete? No - I lost a fair amount of time trying to understand why the test was "stuck". See the issue for the reason we're doing this.

mdellweg: So how about we never wait for `all_tasks_dispatched` (I honestly think it is meaningless), and assert on it being when no pending task is left, so we find the bugs and this fixture tells us exactly what is wrong?

pulpcore/tests/functional/__init__.py
```diff
@@ -812,12 +812,20 @@ def _monitor_task(task_href):
...
+        def tasks_open(tg):
+            return (tg.waiting + tg.running + tg.canceling) > 0
+
         task_group = task_groups_api_client.read(task_group_href)
 
-        while not task_group.all_tasks_dispatched or (task_group.waiting + task_group.running) > 0:
+        # Wait until there are no tasks left to run
+        # Note: if the *tasking system* has a bug that leaves a task (somehow)
+        # in run/wait/canceling, we're going to be stuck in this loop forever.
+        # I don't see a reliable way to avoid that. TaskingSystemBugs: Just Say No! :)
```
mdellweg, line 823: To be fair, the same goes for `monitor_task`. We could add a timeout or rely on the 5h timeout of the ci. We'll see an error eventually.

mdellweg (approved):

pulpcore/tests/functional/__init__.py
ggainey: > We could add a timeout or rely on the 5h timeout of the ci.

I'd suggest "add a reasonably-long timeout and complain when it's exceeded". "Wait 5 hrs" is what got us here in the first place - and it costs us a 5-hr-wait, a confused "huh, weird, lemme restart that", *another* 5-hr-wait, and then a few hours of "wtf is going on here". This is...a little frustrating.

However - "give monitor-task a wait-for timeout with a reasonable default" should be its own ask, not piggybacked on this one.

lubosmj (approved):

PR #3930, opened 2023-06-16
Add version information to tasks

### LINKED ISSUE #3369: When running components in different versions, old workers may fail on new tasks
# Problem
When doing zero downtime upgrades, different components may run in different versions at the same time. So pre-upgrade worker code may attempt to run post-upgrade tasks. For any reasonable change (added feature, added task function, changed function signature, ...) this is deemed to fail.

# Solution 1
Add version constraints to tasks (e.g. "file>=1.2.3"). Old workers would not consider picking up a new task. New workers _must_ know how to handle old tasks.
Pro:
  - Users can always upgrade as long as at least one new worker is deployed with or before the first new api.
  - Backward compatibility issues will show up as such.
Con:
  - Tasks need backward compatibility.

# Solution 2
Add strict version constraints to tasks (e.g. "file==1.2.3"). Only a worker matching the codebase would ever pick up the task. When upgrading you would need to keep an old worker running after upgrading all the api processes until that subset of the task queue is drained.
Pro:
  - No back and forward compatibility for tasks is needed.
Con:
  - User coordination is needed.
  - You cannot keep an old pod running with both api and worker.

# Solution 3
Old workers attempt to execute a new task and will fail gracefully on any incompatible signature change. They will "put the task back into the queue".
Pro:
  - Rather easy to implement without db schema changes.
Con:
  - Backward compatibility issues will look like "task to new".
  - Failing will only be realized after setting up and selecting the new task.
  - Failure needs to propagate from the forked process back to the worker process.
  - Worker will attempt execution every time it starts looking into the task queue again (Or it needs a way to "remember" incompatible tasks.).

# Non-Solution 1
Ignore failed tasks. Not an option.

# Non-Solution 2
Upgrade all workers before api get's upgraded. Does not work, because workers dispatch tasks too. So this would prevent a rolling upgrade of workers.

mdellweg:
Workers will refuse to pick up a task caliming a version of a plugin to be higher than their own. The corresponding plugin an it's version will be inferred from the task import path.

fixes #3369


CHANGES/3369.feature
```diff
@@ -0,0 +1,2 @@
+Added version information to tasks. Workers will refuse to pickup tasks with a version higher
+than their own.
```
MichalPysik, line 2:
```suggestion
Added version information to tasks. Workers will refuse to pick up tasks with a version higher
than their own.
```

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -42,19 +42,30 @@ def _dispatch_task(*args, **kwargs):
 @pytest.fixture(scope="module")
 def task(dispatch_task, monitor_task):
     """Fixture containing a Task."""
-    task_href = dispatch_task("time.sleep", args=(0,))
+    task_href = dispatch_task("pulpcore.app.tasks.test.sleep", args=(0,))
```
gerrod3, line 45: Interestingly we are losing the ability to run outside python functions as tasks with this change. Are we sure that no plugin has been using this behavior already? (I think this is a better design and we should only allow running pulpcore/plugin code)

mdellweg: Yes, I agree, we never made any such statement. But i'd be happy to call it undefined behaviour. Back in the days celery tasks needed to be decorated, and i think that is not the worst design. Actually I would want to be even much stricter about what a task can be. But that is for another day. (sneak preview: https://github.com/pulp/pulpcore/pull/3997)

gerrod3 (approved):

ggainey:
Would it be desirable for the tasking-system to recognize that none of the available tasks can be picked up due to version-incompatibility and log warnings?

Is there any way to have a test-case that places an incompatible-task-entry into the queue and then sees it not-get-picked-up?

mdellweg:
> Would it be desirable for the tasking-system to recognize that none of the available tasks can be picked up due to version-incompatibility and log warnings?

The distributed nature of the tasking system makes it kind of hard (without a lot of overhead including db queries) to identify that no available worker is compatible with the task at hand. Do we want that? This change is designed to help prevent bad to happen in a system that does rolling upgrades. So is suspect that the moment you spin up a new api worker you probable spin up a new worker too (maybe in the same container). And when you don't, you are in a half upgraded system and expect some quirks. It's not that the system is in an undefined state. The task will be picked up once at least one sufficiently recent worker is started.

> 
> Is there any way to have a test-case that places an incompatible-task-entry into the queue and then sees it not-get-picked-up?

Possibly. Not quite sure if we want that. The CI would probably not care that there will be a task waiting forever. Other development systems, idk.

mdellweg:
I added a test, and also changed the semantics a little bit. A worker is now still able to cancel a task even if the versions are not met. I believe this is necessary to be able to clean up e.g. when an upgrade needed rollback or a plugin was removed.

@gerrod3 you may want to take a second look.

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -220,6 +221,23 @@ def cancel_abandoned_task(self, task, final_state, reason=None):
...
 
+    def is_compatible(self, task):
+        unmatched_versions = [
+            f"task: {label}>={version} worker: {self.versions.get(label)}"
+            for label, version in task.versions.items()
+            if label not in self.versions
+            or parse_version(self.versions[label]) < parse_version(version)
+        ]
+        if unmatched_versions:
+            _logger.info(
```
decko, line 232: What about using a `_logger.warn` here?

decko (approved):

pulpcore/tasking/pulpcore_worker.py
mdellweg: I had it warn first but wasn't sure. In case of an ongoing rolling upgrade this can get quite verbose. I was even thinking about debug.
We have conflicting suggestions here:
@ggainey > "I'd make it INFO. We def want to leave traces - esp if/when the task-queue is full of nothing but New Tasks but the only workers running are running Olde Code"

ggainey (approved):

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -220,6 +221,23 @@ def cancel_abandoned_task(self, task, final_state, reason=None):
             self.notify_workers()
         return True
 
+    def is_compatible(self, task):
+        unmatched_versions = [
+            f"task: {label}>={version} worker: {self.versions.get(label)}"
```
gerrod3, line 226:
```suggestion
            f"task: {label}>={version} worker: {self.versions.get(label, 'None')}"
```

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -220,6 +221,23 @@ def cancel_abandoned_task(self, task, final_state, reason=None):
...
+            for label, version in task.versions.items()
+            if label not in self.versions
+            or parse_version(self.versions[label]) < parse_version(version)
```
gerrod3, line 229: I am not sure about this condition check. I think it should be `!=`. Maybe if plugin writers passed in the min required version for their tasks then we could use the this check, but if this is filled out by default we have no guarantee that a plugin's task is compatible with its previous version.

gerrod3 (changes requested):

pulpcore/tasking/pulpcore_worker.py
ggainey: Then we'd basically be saying "you have to cancel all waiting tasks when you upgrade". A task can't declare future-compatibility, so by definition every time you update core or a plugin, existing tasks could never be handled.

Example: I have 200 waiting tasks issued by core/3.25. I upgrade to core/3.28. Using '!=' means now nothing can ever pick up those tasks.

zero-downtime-upgrade requires that new-task-code continue to be able to handle old-tasks; if we can't do that, we can't have ZDU.

pulpcore/tasking/pulpcore_worker.py
mdellweg: Can you elaborate? My experiment (and it is part of the tests too) is that the stringinterpolation works just fine here.

pulpcore/tasking/pulpcore_worker.py
mdellweg: No, the rule is whenever a worker has a version greater or equal than the api dispatching the task, it needs to be able to handle it. You cannot assume there will be an old worker around to handle all the old tasks.
We have that rule already. Plugins must always be able to handle old tasks. This is not even ZDU related.

pulpcore/tasking/pulpcore_worker.py
gerrod3: Actually I don't think this is needed after checking the string interpolation. `f"{None}" == "None"`.

pulpcore/tasking/pulpcore_worker.py
gerrod3: Very well then, I guess this is a requirement I need to be cognizant of when reviewing plugin code as I was not aware of this before.

gerrod3 (approved):

PR #3943, opened 2023-06-21
Add graceful shutdown to pulpcore-worker

### LINKED ISSUE #3940: As a user I can gracefully shut down a pulpcore-worker
A user should be able to stop a `pulpcore-worker` in such a way that allows the worker's current task to run to completion.

mdellweg:
fixes #3940

### DIFF
--- a/pulpcore/tasking/pulpcore_worker.py
+++ b/pulpcore/tasking/pulpcore_worker.py
@@ -114,14 +114,18 @@ class NewPulpWorker:
         startup_hook()
 
     def _signal_handler(self, thesignal, frame):
-        # Reset signal handlers to default
-        # If you kill the process a second time it's not graceful anymore.
-        signal.signal(signal.SIGINT, signal.SIG_DFL)
-        signal.signal(signal.SIGTERM, signal.SIG_DFL)
+        if thesignal in (signal.SIGHUP, signal.SIGTERM):
+            _logger.info(_("Worker %s was requested to shut down gracefully."), self.name)
+            self.task_grace_timeout = -1
+        else:
+            # Reset signal handlers to default
+            # If you kill the process a second time it's not graceful anymore.
+            signal.signal(signal.SIGINT, signal.SIG_DFL)
+            signal.signal(signal.SIGTERM, signal.SIG_DFL)
+            signal.signal(signal.SIGHUP, signal.SIG_DFL)
 
-        _logger.info(_("Worker %s was requested to shut down."), self.name)
-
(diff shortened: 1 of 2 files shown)


pulpcore/tasking/pulpcore_worker.py
```diff
@@ -407,6 +413,7 @@ def child_signal_handler(sig, frame):
     # If you kill the process a second time it's not graceful anymore.
     signal.signal(signal.SIGINT, signal.SIG_DFL)
     signal.signal(signal.SIGTERM, signal.SIG_DFL)
+    signal.signal(signal.SIGHUP, signal.SIG_DFL)
     signal.signal(signal.SIGUSR1, signal.SIG_DFL)
 
     if sig == signal.SIGUSR1:
```
lubosmj, line 419: Does `sys.exit()` trigger any signal? If not, can we move this if-condition up?

mdellweg: What for?
But now that you ask, maybe we don't want to restore the SIGHUP handler at all, so the graceful kill stays graceful...

lubosmj: Setting the signal handlers back to `signal.SIG_DFL` seems to be useless if we will eventually exit the process when `sig == signal.SIGUSR1`.

mdellweg: The idea is that after telling the process to quit (via signalling) and it did not quit in time, we want all the signals to behave normal when we kill it a second time (which should never ever be needed, but you know, software...).

pulpcore/tasking/pulpcore_worker.py
```diff
@@ -114,14 +114,18 @@ def __init__(self):
...
     def _signal_handler(self, thesignal, frame):
-        # Reset signal handlers to default
-        # If you kill the process a second time it's not graceful anymore.
-        signal.signal(signal.SIGINT, signal.SIG_DFL)
-        signal.signal(signal.SIGTERM, signal.SIG_DFL)
-
-        _logger.info(_("Worker %s was requested to shut down."), self.name)
-
-        self.task_grace_timeout = TASK_GRACE_INTERVAL
+        if thesignal == signal.SIGHUP:
```
dkliban, line 117: Kubernetes sends a SIGTERM. We should be handling that signal gracefully. 

https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-terminating-with-grace

dkliban (approved):

bmbouter (approved):
This looks good to me! This will be really helpful.

PR #3947, opened 2023-06-22
Defer get_user_model calls to prevent circular imports

jctanner:
https://github.com/pulp/pulp_container/pull/1294#issuecomment-1602282563

### DIFF
--- a/pulpcore/app/models/role.py
+++ b/pulpcore/app/models/role.py
@@ -1,4 +1,4 @@
-from django.contrib.auth import get_user_model
+from django.conf import settings
 from django.contrib.auth.models import Permission
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.contenttypes.models import ContentType
@@ -41,7 +41,7 @@ class UserRole(BaseModel):
     """
 
     user = models.ForeignKey(
-        get_user_model(), related_name="object_roles", on_delete=models.CASCADE
+        settings.AUTH_USER_MODEL, related_name="object_roles", on_delete=models.CASCADE
     )
     role = models.ForeignKey(Role, related_name="object_users", on_delete=models.CASCADE)
     content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE, null=True)
--- a/pulpcore/app/role_util.py
+++ b/pulpcore/app/role_util.py
@@ -1,19 +1,23 @@
 from gettext import gettext as _
 
 from collections import defaultdict
+from functools import lru_cache
 
 from django.conf import settings
 from django.core.exceptions import BadRequest
 from django.db.models import Q, Exists, OuterRef, CharField
 from django.db.models.functions import Cast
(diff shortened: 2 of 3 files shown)


pulpcore/app/role_util.py
```diff
@@ -300,6 +298,7 @@ def get_users_with_perms_roles(
     include_model_permissions=True,
     for_concrete_model=False,
 ):
+    User = get_user_model()
```
lubosmj, line 301: I am not sure we can accept this as it is. Calling the function whenever working with permissions may not be the best approach. Can we cache `get_user_model` in a singleton class somewhere? More suggestions are welcome.

jctanner: I'm not following or understanding how a "singleton" would be implemented in this case. Is there another example of pulpcore where something like that has been done?

lubosmj: I mean to utilize the singleton pattern in general. Maybe, you could simply resort to `@lru_cache` if possible.

```
@lru_cache()
def blabla():
    return get_user_model()
```

jctanner:
@lubosmj is this more what you were looking for?

lubosmj:
Sort of. @jctanner, can you squash the commits, please?

lubosmj:
Also, it is worth opening an issue and adding a changelog.

Perhaps, we want to set maxsize=1 in lru_cache as well if we decided to use it. Other folks may have different opinion on that. I still think caching is faster than doing always a dictionary and settings lookup. Can we rule out memory leaks once it will hold a reference to django_apps.Model?

mdellweg:
Not sure how much overhead the lru_cache doing a dictionary lookup with no-args is bringing. Is the`get_users_with_perms_roles` hot enough to care?

pulpcore/app/models/role.py
```diff
@@ -41,7 +41,7 @@ class UserRole(BaseModel):
     """
 
     user = models.ForeignKey(
-        get_user_model(), related_name="object_roles", on_delete=models.CASCADE
+        settings.AUTH_USER_MODEL, related_name="object_roles", on_delete=models.CASCADE
```
mdellweg, line 44: :+1:

mdellweg (commented):
Are we now properly breaking the dependenc loops in a way that we don't need the role_util to be it's own separate import path anymore?

OTOH, 500+ lines is probably a new reason for a separate module...

lubosmj:
> Not sure how much overhead the lru_cache doing a dictionary lookup with no-args is bringing. Is the`get_users_with_perms_roles` hot enough to care?

It is not that hot. But, I believe these functions can be called from a ViewSet which could cause a tiny slowdown (unverified hypothesis).

mdellweg:
Re changelog. We should have a bugfix in the plugin section for this (It effects a plugin quite seriously).

And also thanks for finding this much better solution than all the other moved imports!

gerrod3 (approved):

PR #3950, opened 2023-06-23
Add cleanup tasks to remove stale objects

### LINKED ISSUE #3949: Add "schedulable" cleanup tasks for Uploads and PulpTemporaryFiles

mdellweg:
Uploads and PulpTemporaryFiles are cleaned up. Set the *_PROTECTION_TIME settings to None in order to disable this behaviour.

fixes #3949

### DIFF
--- a/CHANGES/3949.feature
+++ b/CHANGES/3949.feature
@@ -1,0 +1,2 @@
+Added periodically executed cleanup tasks for uploads and temporary files. Configure a time
+interval in ``UPLOAD_PROTECTION_TIME`` or ``TMPFILE_PROTECTION_TIME`` to activate.
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -442,6 +442,17 @@ ORPHAN_PROTECTION_TIME
     up before the task finishes. Default is 1440 minutes (24 hours).
 
 
+.. _upload_protection_time:
+.. _tmpfile_protection_time:
+
+UPLOAD_PROTECTION_TIME and TMPFILE_PROTECTION_TIME
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    Pulp uses ``uploads`` and ``pulp temporary files`` to pass data from the api to worker tasks.
+    These options allow to specify a timeinterval in minutes used for cleaning up stale entries. If
+    set to 0, automatic cleanup is disabled, which is the default.
+
+
 .. _task_diagnostics:
 
 TASK_DIAGNOSTICS
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -240,8 +240,10 @@ CONTENT_APP_TTL = 30
 
 WORKER_TTL = 30
 
-# how long to protect orphan content in minutes
+# how long to protect ephemeral items in minutes
 ORPHAN_PROTECTION_TIME = 24 * 60
+UPLOAD_PROTECTION_TIME = 0
+TMPFILE_PROTECTION_TIME = 0
 
 REMOTE_USER_ENVIRON_NAME = "REMOTE_USER"
 
--- a/pulpcore/app/tasks/orphan.py
+++ b/pulpcore/app/tasks/orphan.py
@@ -1,12 +1,15 @@
 import gc
 
 from django.conf import settings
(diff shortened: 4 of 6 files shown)


CHANGES/3949.feature
```diff
@@ -0,0 +1 @@
+Added periodically executed cleanup tasks for uploads and temporary files.
```
lubosmj, line 1: Does it aké sense to mention the setting in the changelog as well?

> Set the *_PROTECTION_TIME settings to None in order to disable this behaviour.

Besides that, I, personally, would make this behaviour disabled by default for a couple of releases. Just in case.

lubosmj: Could you also add a note to the upload workflow (I think it is in the pupp_file repository? Not sure if it is even in our docs because we use CLI and uploads are translated to artifacts already) that as of now we will remove uncommitted uploads?

pulpcore/app/tasks/orphan.py
```diff
@@ -41,56 +44,56 @@ def orphan_cleanup(content_pks=None, orphan_protection_time=settings.ORPHAN_PROT
...
+        code="clean-up.artifacts",
+    ) as progress_bar:
+        for artifact in progress_bar.iter(artifacts.iterator()):
+            # we need to manually call delete() because it cleans up the file on the filesystem
+            artifact.delete()
+
+
+def upload_cleanup():
+    expiration = timezone.now() - timezone.timedelta(minutes=settings.UPLOAD_PROTECTION_TIME)
+    qs = Upload.objects.filter(pulp_created__lt=expiration)
```
lubosmj, line 80: Having `expiration` equal to `timezone.now()` (`settings.UPLOAD_PROTECTION_TIME=0`) results in pulp_create__lt always being less than `expiration`, so we will remove uploads even though users opted for not doing that. Is my understanding correct? If yes, we need to fix it.

mdellweg: Not as such. Having it 0 will unschedule the task, but yes, there may be a window where things go south.
I'll find a different solution.

mdellweg: It is now checked to be > 0.

lubosmj (approved):

dkliban (approved):

docs/configuration/settings.rst
```diff
@@ -442,6 +442,17 @@ ORPHAN_PROTECTION_TIME
...
 
+.. _upload_protection_time:
+.. _tmpfile_protection_time:
+
+UPLOAD_PROTECTION_TIME and TMPFILE_PROTECTION_TIME
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    Pulp uses ``uploads`` and ``pulp temporary files`` to pass data from the api to worker tasks.
+    These options allow to specify a timeinterval in minutes used for cleaning up stale entries. If
+    set to 0, automatic cleanup is disabled, which is the default.
```
ipanova, line 453: sorry for my late 2c, but this a bit contradicts( or confuses?) to our similar setting orphan_protection_time. when that one set to 0 it means the orphans are not protected not that setting/cleanup is disabled .
How can we make this more consistent? The naming for all these 3 settings suggests they work in a similar way.

pulpcore/app/settings.py
```diff
@@ -240,8 +240,10 @@
 
 WORKER_TTL = 30
 
-# how long to protect orphan content in minutes
+# how long to protect ephemeral items in minutes
```
ipanova, line 243: especially yhis comment suggests that having `UPLOAD_PROTECTION_TIME = 0` uploads are not protected and removed ASAP

docs/configuration/settings.rst
mdellweg: It's not a contradiction, but i can see it being confusing. And I must confess, being similar is what inspired me in the first place to use similar names. The biggest difference is probably that we do not have a scheduled task for orphan cleanup (not that this would be impossible...).
Is there any way you think we could improve the documentation here?

ipanova: I hear where you're coming from.
Unfortunately, I don't see what can be improved in the documentation. Documentation is well written per se, and **if** users read it, it's ok even to have behavioral differences between the settings which have similar name.

PR #3976, opened 2023-06-28
Incremental export for file exporter will now include published metadata

parthaa:
Look at https://github.com/pulp/pulpcore/issues/3941#issuecomment-1601977276 for more information

The initial approach was here https://github.com/pulp/pulpcore/pull/3942 where the code ended up looking exclusively  at the repodata directory which is not ideal in pulpcore. 

This PR tries to address the same issue by saying  all published metadata in a publication should get included automatically.

### DIFF
--- a/CHANGES/3941.bugfix
+++ b/CHANGES/3941.bugfix
@@ -1,0 +1,1 @@
+Made the incremental file export include all published metadata.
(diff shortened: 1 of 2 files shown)


mdellweg (approved):
The change looks reasonable to me.

ggainey (approved):

jpasqualetto:
Could we get this backported to 3.18 too?

PR #4005, opened 2023-07-10
Extract a function for recursive downloads from pulp-file tests

lubosmj:
[noissue]

Required by: https://github.com/pulp/pulp_file/pull/990

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -4,6 +4,7 @@ import os
 import shutil
 import socket
 import ssl
+import subprocess
 import threading
 import uuid
 
@@ -1036,3 +1037,21 @@ def http_get():
         return response
 
     return _http_get
+
+
+@pytest.fixture
+def wget_recursive_download_on_host():
+    def _wget_recursive_download_on_host(url, destination):
+        subprocess.check_output(
+            [
+                "wget",
+                "--recursive",
+                "--no-parent",
+                "--no-host-directories",
+                "--directory-prefix",
+                destination,
+                url,
+            ]
+        )
+
+    return _wget_recursive_download_on_host


pulpcore/tests/functional/utils.py
```diff
@@ -79,3 +80,46 @@ async def _download_file(url, auth=None, headers=None):
...
+    data.update(kwargs)
+
+    return data
+
+
+def gen_distribution(**kwargs):
+    """Return a semi-random dict for use in creating a Distribution."""
+    data = {"base_path": str(uuid.uuid4()), "name": str(uuid.uuid4())}
+    data.update(kwargs)
+    return data
```
mdellweg, line 99: Can we somehow not have these and move all the default parameters to the factory fixtures?

pulpcore/tests/functional/utils.py
```diff
@@ -79,3 +80,46 @@ async def _download_file(url, auth=None, headers=None):
...
+    """Performs a POST request on a URL and returns an aiohttp.Response object."""
+    return asyncio.run(_post_url(url, data, return_body, auth=auth))
+
+
+async def _post_url(url, data, return_body, auth=None):
+    async with aiohttp.ClientSession(auth=auth) as session:
+        async with session.post(url, data=data, verify_ssl=False) as response:
+            if return_body:
+                return await response.read()
+            return response
```
mdellweg, line 125: How is any of these an improvement over `requests.get` and `requests.put`?

pulpcore/tests/functional/__init__.py
```diff
@@ -1036,3 +1037,21 @@ async def _send_request():
...
     return _http_get
+
+
+@pytest.fixture
+def wget_download_on_host():
+    def _wget_download_on_host(url, destination):
+        subprocess.check_output(
+            [
+                "wget",
+                "--recursive",
```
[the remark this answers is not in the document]
lubosmj: This was initially used in pulp-smash: https://github.com/pulp/pulp-smash/blob/2ded6671e0f32c51e70681467199e8a195f7f5fe/pulp_smash/pulp3/utils.py#L163. I can rename it to be more explicit.

pulpcore/tests/functional/utils.py
lubosmj: I cannot think of any at the moment. I migrated this off of pulp-file.

pulpcore/tests/functional/utils.py
mdellweg: But can we not inject the name/description into the factory and check that that is what Pulp will have saved afterwards. Maybe i'm misinterpreting these tests (can you point me to an example?). And i kind of fear whatever facility we provide here, is hard to remove again if we seem to not need it anymore.

lubosmj: This is one of the examples in which the function can be used: https://github.com/pulp/pulp_file/blob/c9069acfefa2ee5a821d9ffd116b57b4f6420f74/pulp_file/tests/functional/api/from_pulpcore/test_distributions.py#L54-L65. Are you suggesting making such assertions inside the factories? Thus, before returning a created object where we will iterate over all known dictionary keys (passed by a user) and verify whether they were correctly saved.

mdellweg: I take, that this a typical example and rarely done.
So to me, looking at test, it would not hurt at all to make the test itself define all these fields explicitely and use the factory (that will take and not overwrite these values, but only add missing ones that are needed). I think this would even help reading and understanding the test, as all the necessary information ("I provide this field, so i can assert it to be there later.") is in the actual test.

lubosmj: I agree. Let me remove these utilities.

pulpcore/tests/functional/utils.py
mdellweg: Sure, but can we just start _using_ requests instead? And not encourage more plugins to chime in on this crazyness? Think: 4 functions less to keep supporting for a looong time.

Why is it crazy? There is absolutely no benefit from it being async, because we call it alone in the same blocking way as if it were a synchronous function. This is overly complex.

ggainey (approved):
lg2m!

mdellweg (approved):

PR #4007, opened 2023-07-10
Fix the import path for pulp-hashlib

### LINKED ISSUE #4006: Cannot import pulp_hashlib from the plugin API
There is a wrong path for importing pulp_hashlib specified. Plugins cannot benefit from the function.

lubosmj:
In this commit, there were also removed redundant #noqa directives that
were used in the plugin API. The flake8 configuration had to be adjusted
accordingly to ignore the import errors.

closes #4006

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -7,6 +7,8 @@
 [flake8]
 exclude = ./docs/*,*/migrations/*,./pulpcore/app/protobuf/*
 
+per-file-ignores = */__init__.py: F401
+
 ignore = E203,W503,Q000,Q003,D100,D104,D106,D200,D205,D400,D401,D402
 max-line-length = 100
 
--- a/pulpcore/app/models/__init__.py
+++ b/pulpcore/app/models/__init__.py
@@ -1,23 +1,23 @@
 # https://docs.djangoproject.com/en/3.2/topics/db/models/#organizing-models-in-a-package
 
 # Must be imported first as other models depend on it
-from .base import (  # noqa
+from .base import (
     BaseModel,
     MasterModel,
     pulp_uuid,
 )
 
-from .access_policy import (  # noqa
+from .access_policy import (
     AccessPolicy,
     AutoAddObjPermsMixin,
     Group,
 )
 
-from .domain import Domain  # noqa
+from .domain import Domain
 
-from .acs import AlternateContentSource, AlternateContentSourcePath  # noqa
+from .acs import AlternateContentSource, AlternateContentSourcePath
 
-from .content import (  # noqa
+from .content import (
     Artifact,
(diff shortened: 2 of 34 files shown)


mdellweg (approved):
Wow, how could this have gone unnoticed?

Can you remove the #noqa for something more specific?

lubosmj:
After some discussion, I am removing all `# noqa` clauses and updating the flak8 configuration to ignore the import errors in the plugin API.

lubosmj:
I made a trade-off between ignored `__init__.py` files and plugin API files. Listing particular files seems to be cumbersome in the flake8 configuration. And, some of the files in the plugin API are worth inspecting the imported modules (e.g., https://github.com/pulp/pulpcore/blob/e152b3dd21a1575878c0b8f7414d5a0f855051f8/pulpcore/plugin/actions.py#L17)

.flake8
```diff
@@ -7,6 +7,8 @@
 [flake8]
 exclude = ./docs/*,*/migrations/*,./pulpcore/app/protobuf/*
 
+per-file-ignores = */__init__.py: F401
```
mdellweg, line 10: You may need to make this change in the plugin-template. It has a DO NOT EDIT remark.

mdellweg (commented):
Have you considered doing the `__all__ = [...]`? I think that is what is actually supposed to be done in python.

.flake8
lubosmj: Oh, yes. 😿

lubosmj:
The purpose of `__all__` is to specify which modules should be exported when doing `from XYZ import *`. Personally, I do not see any reason for declaring all the modules once again just for the sake of the flake8 linter. I am against it.

mdellweg (approved):

pulpcore/app/models/__init__.py
```diff
@@ -59,41 +59,33 @@
...
     RepositoryVersionContentDetails,
 )
 
-# This can lead to circular imports with a custom user model depending on this very module
-# Moved to plugin/models/role.py to avoid the circular import.
-# from .role import (  # noqa
-#     GroupRole,
-#     Role,
-#     UserRole,
-# )
```
gerrod3, line 76: Should this comment really be deleted?

lubosmj: Yes. It is redundant. On the other hand, could we bring back the imports in the next breaking release because of https://github.com/pulp/pulpcore/commit/cd342f3b034571d4e33c7e40b71375c59b054f07.

gerrod3: Maybe.

gerrod3 (approved):

ggainey (approved):

PR #4030, opened 2023-07-12
Update ci_update_branches to include 3.29

dkliban:
[noissue]

### DIFF
--- a/template_config.yml
+++ b/template_config.yml
@@ -25,6 +25,7 @@ ci_update_branches:
 - '3.22'
 - '3.23'
 - '3.28'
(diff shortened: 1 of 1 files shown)


gerrod3:
I think you need to run `plugin-template` on this.

gerrod3:
I have gotten confirmation that 3.28 has been chosen for Satellite 6.15, so maybe we just need to append 3.29 to the list.

ggainey (approved):

mdellweg (approved):
Approving, but please, people, we cannot just let this list grow.

PR #4031, opened 2023-07-12
Taught FileDownloader to correctly handle on-demand remotes.

### LINKED ISSUE #4019: --url file:///foo/bar/PULP_MANIFEST with policy of on_demand or streamed raises unhandled RuntimeError
**Version**

Using the all-in-one pulp container with pulpcore 3.28.1 and pulp-file 1.14.3

**Describe the bug**

With a remote with --url file:///foo/bar/ using either on_demand or streamed policy, when attempting to retrieve a file via wget, a 500 Internal Server Error is returned, and the logs show an unhandled RunTimeError

**To Reproduce**

```
pulp file remote create --name foobar --url file:///foo/bar/PULP_MANIFEST --policy on_demand
...
wget http://myserver/pulp/content/foobar/somefile
...
HTTP request sent, awaiting response... 500 Internal Server Error
2023-07-08 01:19:34 ERROR 500: Internal Server Error.
```

**Expected behavior**

The file is successfully downloaded

**Additional context**

I mentioned this issue in IRC and it was confirmed as reproducible

ggainey:
fixes #4019 .

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -1012,6 +1012,13 @@ class Handler:
 
         async def handle_data(data):
             nonlocal data_size_handled
+            # If we got here, and the response hasn't had "prepare()" called on it, it's due to
+            # some code-path (i.e., FileDownloader) that doesn't know/care about
+            # headers_ready_callback failing to invoke it.
+            # We're not going to do anything more with headers at this point, so it's safe to
+            # "backstop" the prepare() call here, so the write() will be allowed.
+            if not response.prepared:
+                await response.prepare(request)
             if range_start or range_stop:
                 start_byte_pos = 0
                 end_byte_pos = len(data)
--- a/pulpcore/tests/functional/api/using_plugin/test_content_access.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_content_access.py
@@ -1,0 +1,40 @@
+"""Tests related to content delivery."""
+import pytest
+import uuid
+
+from pulpcore.client.pulp_file import (
+    RepositorySyncURL,
+)
+
+from pulp_file.tests.functional.utils import (
+    download_file,
+)
+
+
+@pytest.mark.parallel
+def test_file_remote_on_demand(
+    basic_manifest_path,
+    file_distribution_factory,
+    file_fixtures_root,
+    file_repo_with_auto_publish,
+    file_remote_api_client,
+    file_repository_api_client,
+    gen_object_with_cleanup,
+    monitor_task,
+):
+    # Start with the path to the basic file-fixture, build a file: remote pointing into it
+    file_path = str(file_fixtures_root) + basic_manifest_path
+    kwargs = {
+        "url": f"file://{file_path}",
+        "policy": "on_demand",
+        "name": str(uuid.uuid4()),
+    }
+    remote = gen_object_with_cleanup(file_remote_api_client, kwargs)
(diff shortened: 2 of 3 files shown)


ggainey:
Needs a test, getting it up to get some clean test runs and more eyes.

pulpcore/download/file.py
```diff
@@ -37,6 +37,9 @@ def __init__(self, url, *args, **kwargs):
         RemoteSerializer().validate_url(url)
         p = urlparse(url)
         self._path = os.path.abspath(os.path.join(p.netloc, p.path))
+        if kwargs.get("headers_ready_callback", None):
+            self.headers_ready_callback = kwargs.get("headers_ready_callback")
```
mdellweg, line 41: It looks like it's perfectly safe to have `headers_ready_callback==None` (also `get` defaults to `None` anyway).
```suggestion
        self.headers_ready_callback = kwargs.get("headers_ready_callback")
```

ggainey:
Status: writing a test discovered that there is more work to be done.

pulpcore/download/file.py
ggainey: That code turned out to not be the actual issue (alas)

pulpcore/content/handler.py
```diff
@@ -1012,6 +1012,8 @@ async def handle_response_headers(headers):
 
         async def handle_data(data):
             nonlocal data_size_handled
+            if not response.prepared:
+                await response.prepare(request)
```
mdellweg, line 1016: The result of preparing a request is that you cannot add headers anymore (as far as I remember).
Is handle_response_headers not executed in the use case you are fixing here? Or is there a chance it didn't finish yet?
Can we use a `asyncio.Event` to make this function wait until it is set by `handle_response_headers` after the `prepare` call?

mdellweg (changes requested):

pulpcore/content/handler.py
ggainey: The path for returning an on-demand artifact, from a file: entity, through the content-app, is tortuous. handle_response_headers never gets invoked. If you've gotten to handle_data(), the content-app "assumes" prepare() has completed. If you are in this call, you're about to start writing, and will fail if prepare() hasn't happened.

This change makes it at least possible to serve on_demand file: artifacts. A better answer would be rearchitecting/rewriting the content-app to not be a mass of barbed wire - but fixing this use-case isn't the place for that.

mdellweg: OK, I understand now. This is the `FileDownloader` not even knowing about `headers_ready_callback` and therefore never bothering to call it. With the `HttpDownloader` on the other hand, we can always assume that `headers_ready_callback` is awaited before `handle_data` is ever invoked.
Please leave a description like this in a comment there.

pulpcore/tests/functional/api/using_plugin/test_content_access.py
```diff
@@ -0,0 +1,46 @@
+"""Tests related to content delivery."""
+import pytest
+import uuid
+
+from pulp_smash.pulp3.utils import (
+    gen_distribution,
+)
```
mdellweg, line 7: Please don't use pulp-smash.
There's a `file_distribution_factory` for this.

mdellweg (changes requested):

mdellweg (approved):

dkliban (approved):

PR #4036, opened 2023-07-13
Pin importlib-metadata to avoid an unsolvable dependency resolution by pip

### LINKED ISSUE #4042: New release of opentelemetry-python break the dependency tree.
**Version**
All versions greater than 3.26

**Describe the bug**
The new opentelemetry-api package depends on `importlib-metadata ~= 6.0` where the previous version(1.18) depends on the more restrictive ` ~=6.0.0`. With this, `pip` got confused and installs `importlib-metadata==6.8.0` with some other packages being "incompatible" with this release and requiring `==6.0.1`.

**To Reproduce**
Try to install all pulpcore dependencies from the `main` branch.
ex. `podman run -it python:3.11 /bin/bash -c 'git clone https://github.com/pulp/pulpcore /app ; pip install pip-tools; pip install /app'`

**Expected behavior**
It was expected a clean installation.

**Additional context**
The problem itself starts to be described here --> https://matrix.to/#/!aVApiNMtnstWbwDcVU:matrix.org/$HiAzZwFxytADU4ABBQguZ0hs8JslBZYBSBWwbZySGyY?via=libera.chat&via=matrix.org&via=fedora.im

Also, the upstream issue https://github.com/open-telemetry/opentelemetry-python/issues/3382

decko:
Fixes #4042

### DIFF
--- a/CHANGES/4042.bugfix
+++ b/CHANGES/4042.bugfix
@@ -1,0 +1,1 @@
+Pinned importlib-metadata to 6.0.1 given pip is unable to solve an opentelemetry dependency version correctly.
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,10 +19,11 @@ dynaconf>=3.1.12,<3.1.13
 gunicorn>=20.1,<=20.1.0
 jinja2>=3.1,<=3.1.2
 naya>=1.1.1,<=1.1.1
-opentelemetry-distro[otlp]>=0.38b0,<=0.39b0 
+importlib-metadata>=6.0.1,<=6.0.1  # Pinned to fix opentelemetry dependency solving issues with pip 
+opentelemetry-distro[otlp]>=0.38b0,<=0.39b0
 opentelemetry-exporter-otlp-proto-http>=1.17.0,<=1.18.0
-opentelemetry-instrumentation-django>=0.38b0,<=0.39b0 
-opentelemetry-instrumentation-wsgi>=0.38b0,<=0.39b0 
+opentelemetry-instrumentation-django>=0.38b0,<=0.39b0
+opentelemetry-instrumentation-wsgi>=0.38b0,<=0.39b0
 pulp-glue>=0.18.0,<0.21
 protobuf>=4.21.1,<4.23.5
 pygtrie>=2.5,<=2.5.0


mdellweg (commented):
I have questions. How can releasing a new version of the telemetry stack break us if we don't even cover it in the upper bound requirement? That is exactly the kind of problem we set out to avoid using strictly bounded requirements.
I get that the update of opentelemetry-api introduces an unsatisfiable dependency. Can we avoid that using a constraints file?
I wish you could declare incompatibilities on python packages after releasing them.

https://pip.pypa.io/en/stable/user_guide/#constraints-files

mdellweg:
I don't think we can backport upping the lower bound of a dependency.

ggainey:
> I have questions. How can releasing a new version of the telemetry stack break us if we don't even cover it in the upper bound requirement? That is exactly the kind of problem we set out to avoid using strictly bounded requirements. I get that the update of opentelemetry-api introduces an unsatisfiable dependency. Can we avoid that using a constraints file? I wish you could declare incompatibilities on python packages after releasing them.
> 
> https://pip.pypa.io/en/stable/user_guide/#constraints-files

See the issue I opened against otel : https://github.com/open-telemetry/opentelemetry-python/issues/3382 for details.

The net is, otel-0.39b is NOT INSTALLABLE right now. Doesn't matter that we pinned it, it's broken.

decko:
> I have questions. How can releasing a new version of the telemetry stack break us if we don't even cover it in the upper bound requirement? That is exactly the kind of problem we set out to avoid using strictly bounded requirements. I get that the update of opentelemetry-api introduces an unsatisfiable dependency. Can we avoid that using a constraints file? I wish you could declare incompatibilities on python packages after releasing them.
> 
> https://pip.pypa.io/en/stable/user_guide/#constraints-files

I think that `pip` is unable to solve the dependency. `poetry` was able to easily solve it downgrading `importlib-metadata` to 6.0.1. It's not a problem of the opentelemetry package itself, IMHO.

About the constraints file, thanks for the suggestion. I'm trying it right now, but it seems incompatible with the way we're using requirements file and setuptools. Maybe I'm missing something here too, but the documentation about it is too scarce. 

Another solution to this is pinning down the importlib-metadata to 6.0.1.

decko:
> > I don't think we can backport upping the lower bound of a dependency.
> 
> Right now, it is not possible to install pulpcore with otel-0.38b or otel-0.39b. We can change the pins, or we can remove otel, or we can just not ever be able to install/release fixes for previous pulpcore versions.

@ggainey what if we just pin `importlib-metadata` to 6.0.1? Not the best solution, but a quick quirky to allow the CI and releases keep running until we reach a better solution/agreement.

ggainey:
> @ggainey what if we just pin `importlib-metadata` to 6.0.1? Not the best solution, but a quick quirky to allow the CI and releases keep running until we reach a better solution/agreement.

"quick and least-invasive" and "fixes existing releases" are the only things I care about right this second. We can talk about doing something more-correct once we've made our users whole.

ggainey (approved):
Looks like the least-invasive way to address this.

mdellweg (approved):
I agree. Can you add a comment to the importlb-metadata that this is a pin for a third party dep, so not to be removed/relaxed without knowing there's an installable solution?

CHANGES/4042.bugfix
```diff
@@ -0,0 +1 @@
+Pin importlib-metadata to 6.0.1 given pip is unable to solve a opentelemetry dependency version correctly.
```
lubosmj, line 1:
```suggestion
Pinned importlib-metadata to 6.0.1 given pip is unable to solve an opentelemetry dependency version correctly.
```

decko: Thanks for the observation @lubosmj

mdellweg (approved):

Issue #3977, opened 2023-06-29
Openapi spec generated with domains enabled has a duplicate operationId


jlsherrill:
**Version**
3.28.0

**Describe the bug**
Starting the pulp server with DOMAIN_ENABLE=true causes the following pieces to be created in the api spec:

```
        "/pulp/api/v3/docs/api.json": {
            "get": {
                "operationId": "docs_api.json_get",
                "description": "OpenApi3 schema for this API. Format can be selected via content negotiation.\n\n- YAML: application/vnd.oai.openapi\n- JSON: application/vnd.oai.openapi+json",
                "parameters":
```
```
        "/pulp/{pulp_domain}/api/v3/docs/api.json": {
            "get": {
                "operationId": "docs_api.json_get",
                "description": "OpenApi3 schema for this API. Format can be selected via content negotiation.\n\n- YAML: application/vnd.oai.openapi\n- JSON: application/vnd.oai.openapi+json",
                "parameters": [
```
This causes openapi client generation to fail for some languages (such as Go)

**To Reproduce**
1) start pulp with DOMAIN_ENABLED=true
2) fetch the openapi spec
3) look for duplicate operation ids

**Expected behavior**
no duplicate operationIds

**Additional context**
Add any other context about the problem here.  Please provide links to any previous discussions via Discourse or Bugzilla.

jlsherrill:
The same is also true for the operationId 'status_read', used by
* /pulp/{pulp_domain}/api/v3/status/
*/pulp/api/v3/status/

jlsherrill:
And also 'docs_api.yaml_get ' used by:
* /pulp/api/v3/docs/api.yaml
* /pulp/{pulp_domain}/api/v3/docs/api.yaml

jlsherrill:
we've (worked around)[https://github.com/content-services/zest/commit/4060344363339e5efe74c6d81364a863b1e7dd3b] this for now so its not urgent, but would be good to fix

PR #4043, opened 2023-07-14
Fix duplicate operationIds when using domains

gerrod3:
fixes: #3977


pulpcore/tests/functional/api/test_api_docs.py
```diff
@@ -39,3 +41,20 @@ def test_http_method(pulpcore_client, pulp_docs_url):
         pulpcore_client.request("POST", pulp_docs_url)
 
     assert e.value.status == 405
+
+
+@pytest.mark.parallel
+def test_no_dup_operation_ids(pulpcore_client, pulp_docs_url):
```
mdellweg, line 47: I like the test. But i was wondering, do we not have an openapi checker? Does it only warn on this? Why are we ignoring it? OK, i cannot find it...
Also maybe `pulpcore/tests/functional/api/test_openapi_schema.py` is a better match.

mdellweg: Do you think, we can add something like this?

https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/management/commands/spectacular.py#L74C15-L74C15

gerrod3: I've tried this command, but I can't seem to get that line to trigger. Can you show an example of what it outputs?

mdellweg:
```bash
# pulpcore-manager spectacular --validate --fail-on-warn
SchemaGenerationError: Failing as requested due to warnings
```
This is what i got. Wow, not so helpful...

mdellweg (approved):

ggainey (approved):

PR #4090, opened 2023-07-19
Fix a significant memory leak

### LINKED ISSUE #2250: Gunicorn consuming excessive amounts of memory
**Version**
3.16.z

**Describe the bug**
Gunicorn consuming excessive amounts of memory, 3.5-4gb

**To Reproduce**
Unclear

**Expected behavior**
Probably not to have a single gunicorn process use 4gb of memory

**Additional context**

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2035873
Katello forum discussion: https://community.theforeman.org/t/katello-4-5-foreman-3-3-memory-leak-in-gunicorn/29658/22

dralley:
Which occurs when making repeated large queries against the API service. The cache unintentionally was holding references to entire instances of views, which are substantial.

closes #2250

### DIFF
--- a/CHANGES/2250.bugfix
+++ b/CHANGES/2250.bugfix
@@ -1,0 +1,1 @@
+Resolved a memory leak that could occur while making large repeated queries against the API service.
--- a/pulpcore/app/access_policy.py
+++ b/pulpcore/app/access_policy.py
@@ -1,4 +1,3 @@
-from functools import lru_cache
 from rest_access_policy import AccessPolicy
 from rest_framework.exceptions import APIException
 
@@ -12,7 +11,6 @@ class AccessPolicyFromDB(AccessPolicy):
     """
 
     @staticmethod
-    @lru_cache
     def get_access_policy(view):
         """
         Retrieves the AccessPolicy from the DB or None if it doesn't exist.


dralley:
Thanks @hao-yu for the discovery and @gerrod3 for doing additional investigation!

ggainey (approved):
Tests are all green and the discussion makes sense - approving, good job all!

gerrod3:
Can we re-open old issues? This change will be a regression on this old issue: https://github.com/pulp/pulpcore/issues/2802. Not the biggest deal and more preferred then having memory leaks, but it will be something we want to track.

dralley:
>Can we re-open old issues? This change will be a regression on this old issue: https://github.com/pulp/pulpcore/issues/2802. Not the biggest deal and more preferred then having memory leaks, but it will be something we want to track.

Sure, fine with me.  Maybe first we should check up with @gmbnomis and see if it still regresses?

mdellweg:
> Can we re-open old issues? This change will be a regression on this old issue: #2802. Not the biggest deal and more preferred then having memory leaks, but it will be something we want to track.

Since we fixed that issue, I think we should open a new one...

mdellweg (approved):

PR #4112, opened 2023-07-21
Taught repair task to be tolerant in the face of any artifact downloa…

### LINKED ISSUE #4111: Repair API should be tolerant of any failure during the repair process
**Version**
main and older

**Describe the bug**
We have made repair api tolerant to a 404 error but that is not the only error that someone can receive. 

If remote is associated with a proxy and that proxy is limiting connections or misconfigured someway, Then TimeoutException would be a common thing to occur as well. 
If Firewall is not properly configured that may even result in "Connection Reset".

No matter what the failure is, The Repair API, should not halt but continue processing rest of the units that it has identified as corrupted or missing.

**To Reproduce**

**Expected behavior**

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=2173692#c13
https://github.com/pulp/pulpcore/issues/3611 cope with 404

ipanova:
…d failure.

closes #4111

### DIFF
--- a/pulpcore/app/management/commands/handle-artifact-checksums.py
+++ b/pulpcore/app/management/commands/handle-artifact-checksums.py
@@ -2,8 +2,6 @@ import os
 
 from gettext import gettext as _
 
-from aiohttp.client_exceptions import ClientResponseError
-
 from django.conf import settings
 from django.core.management import BaseCommand, CommandError
 from django.db.models import Q, Sum
@@ -126,7 +124,7 @@ class Command(BaseCommand):
                         downloader = remote.get_downloader(ra)
                         try:
                             dl_result = downloader.fetch()
-                        except ClientResponseError as e:
+                        except Exception as e:
                             self.stdout.write(
                                 _("Redownload failed from '{}': {}.").format(ra.url, str(e))
                             )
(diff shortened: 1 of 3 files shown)


pulpcore/app/management/commands/handle-artifact-checksums.py
```diff
@@ -139,8 +137,9 @@ def _download_artifact(self, artifact, checksum, file_path):
                                 setattr(artifact, checksum, dl_result.artifact_attributes[checksum])
                                 restored = True
                                 break
-                self.stdout.write(_("Deleting unreparable file {}".format(file_path)))
-                artifact.file.delete(save=False)
+                if not restored:
+                    self.stdout.write(_("Deleting unreparable file {}".format(file_path)))
```
ggainey, line 141:
```suggestion
                    self.stdout.write(_("Deleting unrepairable file {}".format(file_path)))
```

ggainey (changes requested):

ggainey:
As of now, "Files changed" isn't matching the most-recent push ( a135aa8, which looks correct) - waiting to see if github's web-ui catches up to the code-reality

ggainey (approved):
Checking out the PR locally shows the corrected/expected code, even tho GH webui Files-Changed doesn't. GH status is green, I expect we're in a state where the webui won't fix-itself. I approve the actual changes.

dralley (approved):

Issue #4119, opened 2023-07-24
Upstream Pulp replicate command traceback with DOMAIN_ENABLED=True


dkliban:
The replicate operation returns a 500 error when DOMAIN_ENABLED=True

```
bc45f6dcabec pulp [4b21bac6-dd0a-4105-b885-06b5f306baee]: 127.0.0.1 - admin [24/Jul/2023:16:03:54 +0000] "POST /pulp/a4c4beac-3eb9-47b3-9778-a9cb3c9fa18e/api/
v3/upstream-pulps/ HTTP/1.0" 201 539 "-" "api/test_replication.py::test_replication"
bc45f6dcabec pulp [4b21bac6-dd0a-4105-b885-06b5f306baee]: django.request:ERROR: Internal Server Error: /pulp/a4c4beac-3eb9-47b3-9778-a9cb3c9fa18e/api/v3/upstr
eam-pulps/018988a4-93b8-7e56-b5e8-de21862080de/replicate/
bc45f6dcabec Traceback (most recent call last):
bc45f6dcabec   File "/usr/local/lib/python3.8/site-packages/rest_framework/request.py", line 416, in __getattr__
bc45f6dcabec     return getattr(self._request, attr)
bc45f6dcabec AttributeError: 'WSGIRequest' object has no attribute 'domain'
bc45f6dcabec
bc45f6dcabec During handling of the above exception, another exception occurred:
bc45f6dcabec
bc45f6dcabec Traceback (most recent call last):
bc45f6dcabec   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 55, in inner
bc45f6dcabec     response = get_response(request)
bc45f6dcabec   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 197, in _get_response
bc45f6dcabec     response = wrapped_callback(request, *callback_args, **callback_kwargs)
bc45f6dcabec   File "/usr/local/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 56, in wrapper_view
bc45f6dcabec     return view_func(*args, **kwargs)
bc45f6dcabec   File "/usr/local/lib/python3.8/site-packages/rest_framework/viewsets.py", line 125, in view
bc45f6dcabec     return self.dispatch(request, *args, **kwargs)
bc45f6dcabec   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 509, in dispatch
[...]

PR #4120, opened 2023-07-24
Fixes a bug in the Upstream Pulp Replicate API

dkliban:
This bug was only present when domains were enabled.

fixes: #4119

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -62,6 +62,7 @@ from pulpcore.client.pulpcore import (
     TasksApi,
     TaskSchedulesApi,
     UploadsApi,
+    UpstreamPulpsApi,
     UsersApi,
     UsersRolesApi,
     WorkersApi,
@@ -344,6 +345,11 @@ def repositories_reclaim_space_api_client(pulpcore_client):
 @pytest.fixture(scope="session")
 def repair_api_client(pulpcore_client):
     return RepairApi(pulpcore_client)
+
+
+@pytest.fixture(scope="session")
+def upstream_pulp_api_client(pulpcore_client):
+    return UpstreamPulpsApi(pulpcore_client)
 
 
 # Threaded local fixture servers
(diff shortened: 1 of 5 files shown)


ipanova (approved):

ipanova:
any versions we want to backport this to?

dkliban:
we should backport to 3.23, 3.28, 3.29

dkliban:
I added a test and that required fixing another bug as part of the same PR.

pulpcore/tests/functional/__init__.py
```diff
@@ -1063,3 +1069,37 @@ def _wget_recursive_download_on_host(url, destination):
         )
 
     return _wget_recursive_download_on_host
+
+
+@pytest.fixture()
+def non_default_domain(domains_api_client, pulp_settings, gen_object_with_cleanup):
```
mdellweg, line 1075: Why don't we make this a domain_factory right away?

pulpcore/tests/functional/api/test_replication.py
```diff
@@ -0,0 +1,33 @@
...
+        "api_root": pulp_settings.API_ROOT,
+        "domain": "default",
+        "username": domains_api_client.api_client.configuration.username,
+        "password": domains_api_client.api_client.configuration.password,
+    }
+    upstream_pulp = gen_object_with_cleanup(
+        upstream_pulp_api_client, upstream_pulp_body, pulp_domain=non_default_domain.name
+    )
+    # Run the replicate task and assert that all tasks successfully complete.
+    response = upstream_pulp_api_client.replicate(upstream_pulp.pulp_href)
```
mdellweg, line 30: So this is really the first time we sync from one pulp into itself, right?

mdellweg (changes requested):

pulpcore/tests/functional/__init__.py
dkliban: i was going to convert it on my next PR, but I'll do it now.

pulpcore/tests/functional/api/test_replication.py
dkliban: yes, i believe so.

mdellweg (approved):

PR #4130, opened 2023-07-25
Remove call to asyncio.wait that passed a coroutine

### LINKED ISSUE #4107: BaseDownloader.fetch passes coroutine to asyncio.wait which is forbidden in python 3.11
Python 3.8 deprecated passing coroutines to `asyncio.wait` and Python 3.11 will now [raise an error](https://github.com/python/cpython/blob/a6313d78f21f79ca64dedd38e637509dc530a1b6/Lib/asyncio/tasks.py#L414C13-L414C13). This causes the BaseDownloader.fetch call to fail on Python 3.11 https://github.com/pulp/pulpcore/blob/9dbcc8810f97f53297a933df2e1b74cdc324a8ea/pulpcore/download/base.py#L185 .

Python provides the solution in the error message:  "Passing coroutines is forbidden, use tasks explicitly."

I believe this can be fixed by explicitly converting the coroutine to a task using asyncio's `create_task`

EamonTracey:
fixes #4107


CHANGES/4107.bugfix
```diff
@@ -0,0 +1 @@
+Do not pass coroutine to asyncio.wait since it's forbidden in python 3.11
```
lubosmj, line 1: This looks like we expose too many details to end users. How about something like: "Updated the downloader's ``fetch`` method to comply with Python 3.11".

pulpcore/download/base.py
```diff
@@ -182,8 +182,8 @@ def fetch(self):
         Raises:
             Exception: Any fatal exception emitted during downloading
         """
-        done, _ = asyncio.get_event_loop().run_until_complete(asyncio.wait([self.run()]))
-        return done.pop().result()
+        result = asyncio.get_event_loop().run_until_complete(self.run())
+        return result
```
mdellweg, line 186: Yay! So we did do unnecessary crazy stuff.

CHANGES/4107.bugfix
mdellweg: Also write whole sentence (including the punctuation) in past tense please.

mdellweg (approved):

EamonTracey:
@mdellweg @lubosmj Can this fix be backported at least back to 3.28?

dkliban (approved):

Issue #3910, opened 2023-06-06
Issue running replicate without certificates


PotentialIngenuity:
**Version**
pulpcore - 3.24.1
rpm - 3.20.0
Docker compose

**Describe the bug**
I have an upstream defined with a username and password and tls set to false. When I initiate replication from the replicate endpoint I receive the following error. 
```
pulp [b23e380581534d6d8ff59965b08f764f]: pulpcore.tasking.tasks:INFO: Task e453e78a-c002-42ce-bbd9-09c039f51345 failed (HTTPSConnectionPool(host='repo.host.com', port=443): Max retries exceeded with url: /pulpapi/v3/docs/api.json (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1131)'))))
pulp [b23e380581534d6d8ff59965b08f764f]: pulpcore.tasking.tasks:INFO:   File "/usr/local/lib/python3.8/site-packages/pulpcore/tasking/tasks.py", line 66, in _execute_task
    result = func(*args, **kwargs)

  File "/usr/local/lib/python3.8/site-packages/pulpcore/app/tasks/replica.py", line 52, in replicate_distributions
    for distro in distros:

  File "/usr/local/lib/python3.8/site-packages/pulpcore/app/replica.py", line 67, in upstream_distributions
    distributions = self.distribution_ctx_cls(self.pulp_ctx).list(list_size, offset, params)

  File "/usr/local/lib/python3.8/site-packages/pulp_glue/common/context.py", line 606, in list
    result: Mapping[str, Any] = self.call("list", parameters=payload)

  File "/usr/local/lib/python3.8/site-packages/pulp_glue/common/context.py", line 561, in call
    return self.pulp_ctx.call(

  File "/usr/local/lib/python3.8/site-packages/pulp_glue/common/context.py", line 240, in call
    if self.domain_enabled:

  File "/usr/local/lib/python3.8/site-packages/pulp_glue/common/context.py", line 194, in domain_enabled
    return cast(bool, self.api.api_spec.get("info", {}).get("x-pulp-domain-enabled", False))

  File "/usr/local/lib/python3.8/site-packages/pulp_glue/common/context.py", line 212, in api
    raise PulpException(str(e))
```

**To Reproduce**

**Expected behavior**
replication to initiate and complete successfully.

**Additional context**
I have pulp_web running on 443 with certificates

[...]

PR #4131, opened 2023-07-25
Fixes problems with replicating over HTTPS.

dkliban:
fixes: #3910


pulpcore/app/tasks/replica.py
```diff
@@ -24,12 +26,28 @@ def user_agent():
...
+    # Write out temporary files related to SSL
+    ssl_files = {}
+    for key in ["ca_cert", "client_cert", "client_key"]:
+        if getattr(server, key):
+            ssl_files[key] = NamedTemporaryFile(dir=".")
+            ssl_files[key].write(bytes(getattr(server, key), "utf-8"))
+            ssl_files[key].flush()
+
+    if "ca_cert" in ssl_files:
+        os.environ["PULP_CA_BUNDLE"] = ssl_files["ca_cert"].name
```
mdellweg, line 39: Let me say, i'm not particularly happy with this method of passing data to glue, but at least this environment is confined to the task process, right?

dkliban: Yes, that is correct.

pulpcore/tests/functional/api/test_replication.py
```diff
@@ -34,3 +36,79 @@ def test_replication(
...
+    if not domains_api_client.api_client.configuration.host.startswith("https"):
+        pytest.skip("HTTPS is not enabled for Pulp's API.")
+
+    # Create a non-default domain
+    non_default_domain = domain_factory()
+
+    # Create an Upstream Pulp in the non-default domain
+    upstream_pulp_body = {
+        "name": str(uuid.uuid4()),
+        "base_url": domains_api_client.api_client.configuration.host,
```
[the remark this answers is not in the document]
dkliban: I didn't see a separate fixture for the generic api_client.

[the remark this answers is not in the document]
mdellweg: bindings_cfg must be it...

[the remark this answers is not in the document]
dkliban: ah yes! thank you! i'll update it.

pulpcore/app/tasks/replica.py
```diff
@@ -24,12 +26,28 @@ def user_agent():
...
     domain = get_domain()
     server = UpstreamPulp.objects.get(pk=server_pk)
+
+    # Write out temporary files related to SSL
+    ssl_files = {}
+    for key in ["ca_cert", "client_cert", "client_key"]:
+        if getattr(server, key):
+            ssl_files[key] = NamedTemporaryFile(dir=".")
+            ssl_files[key].write(bytes(getattr(server, key), "utf-8"))
+            ssl_files[key].flush()
```
mdellweg, line 36:
```suggestion
    for key in ["ca_cert", "client_cert", "client_key"]:
        if value := getattr(server, key):
            ssl_files[key] = NamedTemporaryFile(dir=".")
            ssl_files[key].write(bytes(value, "utf-8"))
            ssl_files[key].flush()
```

pulpcore/app/tasks/replica.py
```diff
@@ -24,12 +26,28 @@ def user_agent():
...
+        os.environ["PULP_CA_BUNDLE"] = ssl_files["ca_cert"].name
+
     api_kwargs = dict(
         base_url=server.base_url,
         username=server.username,
         password=server.password,
         user_agent=user_agent(),
+        validate_certs=server.tls_validation,
+        cert=ssl_files["client_cert"].name if "client_cert" in ssl_files else None,
+        key=ssl_files["client_key"].name if "client_key" in ssl_files else None,
```
mdellweg, line 48:
```suggestion
        cert=ssl_files.get("client_cert"),
        key=ssl_files.get("client_key"),
```

dkliban: shouldn't these be names of files?

mdellweg: What if we just put the names in the dict?

dkliban: i'll do that.

mdellweg (approved):

pulpcore/tests/functional/api/test_replication.py
```diff
@@ -34,3 +36,80 @@ def test_replication(
...
+        upstream_pulp_api_client, upstream_pulp_body, pulp_domain=non_default_domain.name
+    )
+    # Run the replicate task and assert that it fails with SSLCertVerificationError
+    response = upstream_pulp_api_client.replicate(upstream_pulp.pulp_href)
+    try:
+        monitor_task_group(response.task_group)
+    except PulpTaskGroupError as e:
+        e.task_group.tasks[0]
+        task = tasks_api_client.read(e.task_group.tasks[0].pulp_href)
+        task.error["description"].find("SSLCertVerificationError") > -1
```
gerrod3, line 106: Shouldn't this be an `assert`? Also, I would re-write this `try/except` to be a `with pytest.raises(PulpTaskGroupError) as e`.

gerrod3 (changes requested):

pulpcore/tests/functional/api/test_replication.py
dkliban: yes, that definitely should be an assert. and i will update to using pytest.raises()

gerrod3 (approved):

PR #4133, opened 2023-07-26
Clarify upload/tmp protection time settings.

ipanova:
[noissue]


lubosmj (approved):

ipanova:
> That is what the docs say, but I agree it can help to repeat it here.

Exactly. that's for users who don't read docs much

PR #4160, opened 2023-07-26
Taught import to be able to use less than all-available workers.

### LINKED ISSUE #4068: PulpImport can overload the task-queue
**Version**
main

**Describe the bug**
PulpImport dispatches a task-per-repository. For large imports, with large repositories, this means that Pulp can't do anything else until the import is finished or nearly so, as all workers are busy importing repositories.

It would be good to teach the import-process to arrange to not use up more than, say, 25% of the available workers.

ggainey:
IMPORT_WORKERS_PERCENT is configurable in settings. We will
document/expose this in a future PR, to keep this one maximally
backportable. Default behavior remains "all workers".

fixes https://github.com/pulp/pulpcore/issues/4068.

### DIFF
--- a/CHANGES/4068.bugfix
+++ b/CHANGES/4068.bugfix
@@ -1,0 +1,1 @@
+Taught pulp-import to be able to use a subset of available worker-threads.
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -8,6 +8,7 @@ import tarfile
 from gettext import gettext as _
 from logging import getLogger
 
+from django.conf import settings
 from django.core.files.storage import default_storage
 from django.db.models import F
 from naya.json import stream_array, tokenize
@@ -35,6 +36,7 @@ from pulpcore.app.modelresource import (
     RepositoryResource,
 )
 from pulpcore.constants import TASK_STATES
+from pulpcore.tasking.pulpcore_worker import Worker
 from pulpcore.tasking.tasks import dispatch
 
 from pulpcore.plugin.importexport import BaseContentResource
@@ -506,6 +508,18 @@ def pulp_import(importer_pk, path, toc, create_repositories):
                             default_storage.save(base_path, f)
 
         # Now import repositories, in parallel.
+
+        # We want to be able to limit the number of available-workers that import will consume,
+        # so that pulp can continue to work while doing an import. We accomplish this by creating
+        # a reserved-resource string for each repo-import-task based on that repo's index in
+        # the dispatch loop, mod number-of-workers-to-consume.
+        #
+        # By default (setting is not-set), import will continue to use 100% of the available
+        # workers.
+        import_workers_percent = int(settings.get("IMPORT_WORKERS_PERCENT", 100))
+        total_workers = Worker.objects.online_workers().count()
+        import_workers = max(1, int(total_workers * (import_workers_percent / 100.0)))
+
         with open(os.path.join(temp_dir, REPO_FILE), "r") as repo_data_file:
             data = json.load(repo_data_file)
             gpr = GroupProgressReport(
@@ -517,14 +531,16 @@ def pulp_import(importer_pk, path, toc, create_repositories):
             )
             gpr.save()
 
-            for src_repo in data:
+            for index, src_repo in enumerate(data):
+                # Lock the repo we're importing-into
                 dest_repo_name = _get_destination_repo_name(importer, src_repo["name"])
-
+                # pulpcore-worker limiter
+                worker_rsrc = f"import-worker-{index % import_workers}"
+                exclusive_resources = [worker_rsrc]
                 try:
                     dest_repo = Repository.objects.get(name=dest_repo_name)
                 except Repository.DoesNotExist:
                     if create_repositories:
-                        exclusive_resources = []
                         dest_repo_pk = ""
                     else:
                         log.warning(
@@ -534,7 +550,7 @@ def pulp_import(importer_pk, path, toc, create_repositories):
                         )
                         continue
                 else:
-                    exclusive_resources = [dest_repo]
+                    exclusive_resources.append(dest_repo)
                     dest_repo_pk = dest_repo.pk
 
                 dispatch(
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -302,6 +302,10 @@ SHELL_PLUS_IMPORTS = [
     "from pulpcore.tasking.util import cancel_task",
 ]
 
+# What percentage of available-workers will pulpimport use at a time, max
+# By default, use all available workers.
+IMPORT_WORKERS_PERCENT = 100
+
 # HERE STARTS DYNACONF EXTENSION LOAD (Keep at the very bottom of settings.py)
 # Read more at https://dynaconf.readthedocs.io/en/latest/guides/django.html
 from dynaconf import DjangoDynaconf, Validator  # noqa


ggainey:
Tested and works. In draft because I want some eyes on this before we decide it's the Right Way to go.

Note that we are separating this from actually adding IMPORT_WORKERS_PERCENT to settings.py to facilitate backporting the actual fix, from the Y-version of "added a config-option".

dralley:
>Tested and works. In draft because I want some eyes on this before we decide it's the Right Way to go.

If we can provide it as a hotfix for the one user that really needs it, we can take a bit more time with choosing the correct approach.  Since the first import is obviously the biggest and most problematic, they should be freed up once they're through it.

pulpcore/app/tasks/importer.py
```diff
@@ -506,6 +508,15 @@ def safe_extract(tar, path=".", members=None, *, numeric_owner=False):
                             default_storage.save(base_path, f)
 
         # Now import repositories, in parallel.
+
+        # We want to limit the number of available-workers that import will consume, so that
+        # pulp can continue to work while doing an import. We accomplish this by creating a
+        # reserved-resource string for each repo-import-task based on that repo's index in
+        # the dispatch loop, mod number-of-workers-to-consume.
+        import_workers_percent = int(settings.get("IMPORT_WORKERS_PERCENT", 25))
```
mdellweg, line 516: This _is_ adding it as a setting.
Also i was wondering if we should add it to the import viewset so you can make the admin (only persona to use imports) decide on a case by case basis.

pulpcore/app/tasks/importer.py
```diff
@@ -517,14 +528,16 @@ def safe_extract(tar, path=".", members=None, *, numeric_owner=False):
             )
             gpr.save()
 
-            for src_repo in data:
+            for index, src_repo in enumerate(data):
+                # Lock the repo we're importing-into
                 dest_repo_name = _get_destination_repo_name(importer, src_repo["name"])
-
+                # pulpcore-worker limiter
+                worker_rsrc = f"import-worker-{index % import_workers}"
```
mdellweg, line 535: So this is a global worker limit. (Sounds fine...)
But all import tasks will start with dispatching "import-worker-0". Can we add some random offset here?

pulpcore/app/tasks/importer.py
```diff
@@ -517,14 +528,16 @@ def safe_extract(tar, path=".", members=None, *, numeric_owner=False):
...
                 try:
                     dest_repo = Repository.objects.get(name=dest_repo_name)
                 except Repository.DoesNotExist:
                     if create_repositories:
-                        exclusive_resources = []
+                        exclusive_resources = [worker_rsrc]
```
mdellweg, line 540: Let's move this out of the "if" statement. And change the one below to `exclusive-resources.append(dest_repo)`

pulpcore/app/tasks/importer.py
dralley: I agree, also it is better to keep defaults all in one place.  We're basically putting the default in the code here.

mdellweg: What i was saying is that i thought we didn't want to make it a setting in this backportable hotfix, no?

pulpcore/app/tasks/importer.py
ggainey: Is it worth the complication? It's an issue exactly and only if you start multiple imports at the same time, which is a never-happen case for the admin.

pulpcore/app/tasks/importer.py
ggainey: > What i was saying is that i thought we didn't want to make it a setting in this backportable hotfix, no?

With the code looking for a setting (that we haven't advertised yet), we make it at least possible for installations with the fix to adjust their worker-use if absolutely necessary. Without it, Support has to tell customers "upgrade or patch the python code", which is...rude.

Is this a perfect solution? No. It's a pragmatic one that is trying to thread the needle of "fix a problem, while still making it possible for users to still give import All The Workers, in a way that can be backported to (very) previous versions of pulpcore."

ggainey: > Also i'm not sure if we want a setting in the end at all.

Honestly, we do. In a lot of import-use-cases, when you're importing, you're importing All The Things, and aren't doing anything else in your system. In that workflow, maximum-concurrency is desirable.

mdellweg:
So with this way of accessing the setting, you say we are free to delete/change it later?
I think the same goes for a setting in settings.py but not in the docs, making it particularly "undocumented".

mdellweg (approved):
One last ask.
Feel free to merge.

lubosmj (approved):

ipanova:
> > So with this way of accessing the setting, you say we are free to delete/change it later? I think the same goes for a setting in settings.py but not in the docs, making it particularly "undocumented".
> 
> Yeah, I think adding it to settings.py in this PR is the right way to go. Working on updates now.

I am not sure I feel comfortable with this approach. Once backported this setting even if undocumented will be available in the settings.py file accessible for manipulation and changes.
@ggainey  sorry for back and forth but can we in this PR drop this change https://github.com/pulp/pulpcore/pull/4160/files#diff-4e1146f075426039ac8e1350e88fe5736a3e50872a42d101bfc8bc1b7f483574R306 and this line hardcode to number 25 https://github.com/pulp/pulpcore/pull/4160/files#diff-06badc442eb525dc6f2de7f4e19a7762ce20a972b79b6a90cf1ee80211053ab5R516 and in separate followup PR bring back the setting  https://github.com/pulp/pulpcore/pull/4160/files#diff-4e1146f075426039ac8e1350e88fe5736a3e50872a42d101bfc8bc1b7f483574R306 and this line https://github.com/pulp/pulpcore/pull/4160/files#diff-06badc442eb525dc6f2de7f4e19a7762ce20a972b79b6a90cf1ee80211053ab5R516 + document setting+ in CHANGES/.feature

ipanova (changes requested):
per my comment

mdellweg:
As i tried to explain: Having it in the settings file is not what makes the setting officially available. And there is absolutely no difference between having the default in the settings.py or buried deep in the code.
My worry is that we make it there-to-stay even if not official the moment we tell an support person to use it.

ipanova:
> As i tried to explain: Having it in the settings file is not what makes the setting officially available. And there is absolutely no difference between having the default in the settings.py or buried deep in the code. It is till "undocumented". My worry is that we make it there-to-stay even if not official the moment we tell an support person to use it.

Given that I understood your reply, that's why I am asking to not have the setting here at all - neither in settings.py nor burried in the code, so simply sticking to `import_workers_percent = 25` will do for the backport purpose. And in follow up PR will come the rest.

ggainey:
Great discussion all, thanks!

Changing percent-used-by-default, based on this summing-up from matrix:

"I'm a big fan of what I think your discussion leads to - "default to all-workers, allow users to set a percentage if they want/need to"."

CHANGES/4068.bugfix
```diff
@@ -0,0 +1 @@
+Taught pulp-import to never use more than 25% of available worker-threads.
```
ipanova, line 1: needs update >.>

ggainey: It is updated - github isn't showing us the last thing I pushed. Yay Friday?

ipanova: dejavu!

ggainey: There we go - I updated title/1st-comment to match the updated commit-msg

ipanova (approved):
Thank you!

ipanova:
> If you give me a hotfix on top of 6.12.4.1, I'll be able to test in the LAB Satellite of my current CU.

That's the next step to get it downstream.

PR #4161, opened 2023-07-26
Taught export to insure de-duplicated Artifact.json.

### LINKED ISSUE #4159: PulpExport ArtifactResource.json has duplicate entries for overlapping content.
**Version**
main

**Describe the bug**
PulpExport is typically used to export the content of many repositories. It gathers up the unique Artifacts themselves into a single directory. However, the ArtifactResource.json file which is used to import the datasbe metadata is just a **list** of artifacts found in all the repositories.

In the presence of exported repositories with duplicate content (e.g., Satellite CVs with many of the same base repositories in them), the resulting Artifact.json file can get very large. The deduplication is handled correctly, but it takes up more (sometimes A LOT more) space and takes a lot more time to import as the import-export engine figures out whether a row is a duplicate or not.

**To Reproduce**
```
pulp file remote create --name many --policy immediate --url https://fixtures.pulpproject.org/file-many/PULP_MANIFEST
for r in {0..100}; do   \
  pulp file repository create --name "repo-$r" --remote many  \
  pulp file repository sync --name "repo-$r"  \
done
pulp exporter pulp create --name many --path /tmp/exports/ --repository (href-of-one-repo) --repository...
pulp export pulp run --exporter many
# extract the Artifact.json from the resulting tar.gz
tar xvf  export-01899388-9d86-75d1-8da0-8c2ebf466d87-20230726_1849.tar.gz  pulpcore.app.modelresource.ArtifactResource.json
cat pulpcore.app.modelresource.ArtifactResource.json | jq '.[] | .sha256' | sort | uniq -c
```
**Expected behavior**
Artifact,json should show 1 entry per sha256.

**Actual behavior**
Artifact.json shows **number-of-repositories** copies of a given sha256.

**Additional context**
To create the exporter with all the repositories, I did something like the following:
```
# get all the repo-hrefs into a file, "hrefs"
pulp file repository list --limit 200 --field pulp_href | jq -r '.[] | .pulp_href' > hrefs
# Edit the file to start each line with ` --repository ` and end it with `\$`
# Add this as the first line
pulp exporter pulp create --name many --path /tmp/exports/ \                          
# Now run as a single command
source ./hrefs
```

There are several end-users being affected by this, even though they may not know it yet. The larger the installation, the more painful this bug becomes.

ggainey:
fixes #4159.

### DIFF
--- a/CHANGES/4159.bugfix
+++ b/CHANGES/4159.bugfix
@@ -1,0 +1,4 @@
+Taught the Artifact.json of an export to hold minimum-unique-set of Artifact entries.
+
+In highly-duplicated-content export scenarios, this can mean a significant decrease
+in export-size, and significant improvement in import-performance.
--- a/pulpcore/app/tasks/export.py
+++ b/pulpcore/app/tasks/export.py
@@ -500,7 +500,7 @@ def _do_export(pulp_exporter, tar, the_export):
     starting_versions = _get_starting_versions(do_incremental, pulp_exporter, the_export)
     vers_match = _version_match(ending_versions, starting_versions)
     # Gather up versions and artifacts
-    artifacts = []
+    artifacts = None  # Will be a QuerySet selecting the Artifacts that need to be exported
     for version in ending_versions:
         # Check version-content to make sure we're not being asked to export
         # an on_demand repo
@@ -509,10 +509,15 @@ def _do_export(pulp_exporter, tar, the_export):
             raise RuntimeError(_("Remote artifacts cannot be exported."))
 
         if do_incremental:
-            vers_artifacts = version.artifacts.difference(vers_match[version].artifacts).all()
+            vers_artifacts = version.artifacts.difference(vers_match[version].artifacts)
         else:
-            vers_artifacts = version.artifacts.all()
-        artifacts.extend(vers_artifacts)
+            vers_artifacts = version.artifacts
+
+        if artifacts:
+            artifacts.union(vers_artifacts)
+        else:
+            artifacts = vers_artifacts
+
     # export plugin-version-info
     export_versions(the_export, plugin_version_info)
     # Export the top-level entities (artifacts and repositories)
--- a/pulpcore/app/importexport.py
+++ b/pulpcore/app/importexport.py
@@ -94,14 +94,14 @@ def export_artifacts(export, artifacts):
 
     Args:
         export (django.db.models.PulpExport): export instance that's doing the export
-        artifacts (django.db.models.Artifacts): list of artifacts in all repos being exported
+        artifacts (django.db.models.Artifacts): QuerySet of artifacts in all repos being exported
 
     Raises:
         ValidationError: When path is not in the ALLOWED_EXPORT_PATHS setting
     """
     data = dict(message="Exporting Artifacts", code="export.artifacts", total=len(artifacts))
     with ProgressReport(**data) as pb:
-        for artifact in pb.iter(artifacts):
+        for artifact in artifacts.iterator():  # chunk_size= defaults to 2000 at a fetch
             dest = artifact.file.name
             if settings.DEFAULT_FILE_STORAGE != "pulpcore.app.models.storage.FileSystem":
                 with tempfile.TemporaryDirectory(dir=".") as temp_dir:
@@ -112,6 +112,7 @@ def export_artifacts(export, artifacts):
                         export.tarfile.add(temp_file.name, dest)
             else:
                 export.tarfile.add(artifact.file.path, dest)
+            pb.increment()
 
     resource = ArtifactResource()
     resource.queryset = artifacts


CHANGES/4159.bugfix
```diff
@@ -0,0 +1,4 @@
+Taught the Artifact.json of an export to hold minimum-unique-set of Artifact entries.
+
+In highly-duplicated-content export scenarios, this can mean a significant decrease
+in export-size, and s ginificant improvement in import-performance.
```
mdellweg, line 4:
```suggestion
in export-size, and siginificant improvement in import-performance.
```

pulpcore/app/tasks/export.py
```diff
@@ -500,7 +500,7 @@ def _do_export(pulp_exporter, tar, the_export):
     starting_versions = _get_starting_versions(do_incremental, pulp_exporter, the_export)
     vers_match = _version_match(ending_versions, starting_versions)
     # Gather up versions and artifacts
-    artifacts = []
+    artifacts = set()
```
mdellweg, line 503: So we _do_ have all artifacts in memory at once?
Does set work well with a Django model? If not, i think we can make it a `dict(pk, instance)`.

ggainey: We have all the Artifact models in memoryu at once, yes. This concerns me, but not in the context of this change - if anythiing, we're making it *better* here by removing duplicates. I think there's improvement to be had, but that will wait for a more intrusive refactor.

set() "appears to work" fine with Django - all my tests, and our existing test-scenarios, are working fine. If you can think of ways to Break Things, add them here please!

dralley: As best I can tell (from 10 year old stackoverflow posts), the `__hash__` value of a django model is a combination of it's type and primary key.  So this should be fine.

mdellweg: Don't we use these hydrated Artifacts to create the export later? Or is there another query anyway?

dralley: >Don't we use these hydrated Artifacts to create the export later? 

Currently I expect we do, yes.  My point is that we could probably just pass the list and have them queried in-situ, using `.iterator()` to ensure that the whole query isn't sitting in memory at once.

ggainey: > Don't we use these hydrated Artifacts to create the export later? Or is there another query anyway?

This is the list-of-things to be passed to the query that builds what's exported. I am going to do a few experiments here to see if I can avoid hydrating anything more than the pks and sha256s; if I get reasonable results "today", will add to this PR, otherwise I'll open a new issue to continue that investigation and just point to this discussion.

jpasqualetto:
What's the impact of this change on fs-exports?

pulpcore/app/tasks/export.py
ggainey: OK, the most-recent push moves us from passing around hydrated Artifacts, to just passing around the QuerySet for later use.

Testing current code vs the code in this PR, exporting 10 repos with A Lot of duplicated content (20K elements) shows a 15-20% speed boost, and a serious size improvement in the resulting export (which is the goal if this PR).

current code:  **13m50s, 91Mb** file
PR code: **11m02s, 31Mb** file
Script to create/export:
```
# 10 large repos                                                                             
pulp file remote create --name huge --policy immediate --url https://fixtures.pulpproject.org/file-perf/PULP_MANIFEST
for r in {0..9}; do \                                                                        
  pulp file repository create --name "huge-${r}" --remote huge \                             
  HREF=(pulp -b file repository sync --name "huge-${r}" |& grep /pulp/api | cut -d " " -f 4) \
done                                                                                         
pulp task show --wait --href "${HREF}"                                                       
REPOS=""                                                                                     
for r in {0..9}; do REPOS="${REPOS} --repository=file:file:huge-${r}"; done                  
# the insert of REPOS below doesn't quite work ,but you get the idea...                      
pulp exporter pulp create --name ten-huge "${REPOS}" --path /tmp/ten-huge                    
pulp export pulp run --exporter ten-huge 
```                                                    

When exporting the above using the current codebase (core/main):
```
"started_at": "2023-07-27T20:14:54.541848Z",  
"finished_at": "2023-07-27T20:28:48.689774Z" : 13m50s
# ls -lh /tmp/ten-huge/export-018998fd-745b-7ceb-83bd-1fd6dce392fa-20230727_2014.tar.gz
-rw-r--r--. 1 pulp pulp 91M Jul 27 20:28 /tmp/ten-huge/export-018998fd-745b-7ceb-83bd-1fd6dce392fa-20230727_2014.tar.gz
```

Running the same export, with the code in this PR:
```       
"started_at": "2023-07-27T21:43:07.628458Z",
"finished_at": "2023-07-27T21:54:09.973836Z" : 11m02s
ls -lh /tmp/ten-huge/export-0189994e-3875-7dc8-ab04-012809ea93ca-20230727_2143.tar.gz
-rw-r--r--. 1 pulp pulp 31M Jul 27 21:54 /tmp/ten-huge/export-0189994e-3875-7dc8-ab04-012809ea93ca-20230727_2143.tar.gz
```

ggainey:
> What's the impact of this change on fs-exports?

That is a good question. fs-export is a different code-path aimed at a different end-output, but might benefit from what we've learned here - if not as part of this PR/issue. I'll review and open a new issue if it looks like we can follow a similar pattern there.

dralley:
+1 to new issue / separate PR

pulpcore/app/tasks/export.py
dralley: Sweet!  Is it ready to undraft?

dralley: Can you at least leave  a TODO on this line about trying `.iterator()`, if we can't get to it now (which I don't want to push)

https://github.com/ggainey/pulpcore/blob/134610ff1eb48ce47fb06c0e5bae2f74dd815701/pulpcore/app/importexport.py#L104

Maybe we can deal with it in the followup w/ filesystem exports if one is needed.

ggainey: The impact on the **import** side is...marked.

Taking the two exports above, one created with current-bad-code and one with the code in this PR, results in the following:

import current-code-export: rows-processed **200000** (!), time **63m01s**
import PR-code-export: rows-processed **20000**, time **2m21s** (!!!)

Raw data:
```
IMPORT SIDE, current code                                                                    
    "progress_reports": [                                                                    
        {                                                                                    
            "code": "import.artifacts",                                                      
            "done": 200000,                                                                  
            "message": "Importing Artifacts",                                                
            "state": "completed",                                                            
            "suffix": null,                                                                  
            "total": null                                                                    
        }                                                                                    
"started_at": "2023-07-28T00:42:33.966652Z",                                                 
"finished_at": "2023-07-28T01:05:34.281462Z" : 63m01s                                        
                                                                                             
IMPORT SIDE, new code:                                                                       
    "progress_reports": [                                                                    
        {                                                                                    
            "code": "import.artifacts",                                                      
            "done": 20000,                                                                   
            "message": "Importing Artifacts",                                                
            "state": "completed",                                                            
            "suffix": null,                                                                  
            "total": null                                                                    
        }                                                                                    
"started_at": "2023-07-28T01:11:03.498821Z",                                                 
"finished_at": "2023-07-28T01:13:24.435164Z" : 2m21s (!!!!)
```

ggainey:
> What's the impact of this change on fs-exports?

Taking a quick look at [_export_to_filesystem](https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/export.py#L59C5-L59C27), it looks like it is relying on hydrated Artifacts where it could get away with just knowing the Artifact.file. This is "mildly" suboptimal, but not a dealbreaker - going to the ddatabase **at all** takes most of your time, Artifacts aren't so large that "all fields" is horribly worse than just "a few fields".

I don't see a path that suffers the same duplication-effort - it doesn't do multiple-repos-at-once, the source of our Pain in this PR.

My first thought is that I wouldn't spend time trying to squeeze more performance out of that codepath, unless/until we actually see a problem.

dralley (approved):
LGTM, but please leave a TODO in the aforementioned spot for future reference

pulpcore/app/tasks/export.py
```diff
@@ -500,7 +500,7 @@ def _do_export(pulp_exporter, tar, the_export):
     starting_versions = _get_starting_versions(do_incremental, pulp_exporter, the_export)
     vers_match = _version_match(ending_versions, starting_versions)
     # Gather up versions and artifacts
-    artifacts = []
+    artifacts = None  # Will be a QuerySet selecting the Artifacts that need to be exported
```
lubosmj, line 503: Can we use `Artifact.objects.none()` here? Then, we can get rid of the if/else statement down below.

pulpcore/app/tasks/export.py
```diff
@@ -509,10 +509,15 @@ def _do_export(pulp_exporter, tar, the_export):
...
         if do_incremental:
-            vers_artifacts = version.artifacts.difference(vers_match[version].artifacts).all()
+            vers_artifacts = version.artifacts.difference(vers_match[version].artifacts)
         else:
-            vers_artifacts = version.artifacts.all()
-        artifacts.extend(vers_artifacts)
+            vers_artifacts = version.artifacts
+
+        if artifacts:
+            artifacts.union(vers_artifacts)
```
mdellweg, line 517: I like this change a lot.

Just for my understanding: This query will look something like "Artifact that is part of content that is in a version exported [but not in the prev version|when incremental]"?
Instead of "Artifact having one of these multiple millions of uuids".

mdellweg (approved):
I'd even be fine with merging this as is and promise to start on further refactoring right away with both ideas:
* Using a db iterator
* using qs.none()

Another thing (to think about later) that came to my mind: In the case of incremental, we now still include artifacts that are new to a repository version, but guaranteed to be in another (exported) repositories previous version. Could we safely skip them too?
Say the logic switching from:

$\displaystyle \bigcup_{\textrm{version}} \left( \textrm{version}.\textrm{artifacts} \setminus \textrm{prev}(\textrm{version}).\textrm{artifacts} \right)$

to

$\left( \displaystyle \bigcup_{\textrm{version}} \textrm{version}.\textrm{artifacts} \right) \setminus \left( \displaystyle \bigcup_{\textrm{version}} \textrm{prev}(\textrm{version}).\textrm{artifacts} \right)$

ggainey:
> I'd even be fine with merging this as is and promise to start on further refactoring right away with both ideas:

I am absolutely open to ways to make this better in the face of "many repos at once", as a future improvement. Open an issue, if you would, so we have at least a placeholder pointing back to this discussion.

ggainey:
> LGTM, but please leave a TODO about maybe using `.iterator()` here
> 
> https://github.com/ggainey/pulpcore/blob/134610ff1eb48ce47fb06c0e5bae2f74dd815701/pulpcore/app/importexport.py#L104

Clarify please? Do you mean just change to `...pb.iter(artifacts.iterator())`? I feel liike I'm missing something .

pulpcore/app/tasks/export.py
ggainey: Oh ho! Larn summat new every day....

ggainey:
hold off on merging while I investigate some test weirdness, please

pulpcore/app/tasks/export.py
ggainey: Soooo, making this change:
```
$ git diff
diff --git a/pulpcore/app/tasks/export.py b/pulpcore/app/tasks/export.py
index 3245c42fb..939598b0a 100644
--- a/pulpcore/app/tasks/export.py
+++ b/pulpcore/app/tasks/export.py
@@ -14,6 +14,7 @@ from pkg_resources import get_distribution
 from django.conf import settings
 
 from pulpcore.app.models import (
+    Artifact,
     CreatedResource,
     ExportedResource,
     Exporter,
@@ -500,7 +501,7 @@ def _do_export(pulp_exporter, tar, the_export):
     starting_versions = _get_starting_versions(do_incremental, pulp_exporter, the_export)
     vers_match = _version_match(ending_versions, starting_versions)
     # Gather up versions and artifacts
-    artifacts = None  # Will be a QuerySet selecting the Artifacts that need to be exported
+    artifacts = Artifact.objects.none()  # QuerySet selecting the Artifacts that need to be exported
     for version in ending_versions:
         # Check version-content to make sure we're not being asked to export
         # an on_demand repo
@@ -513,10 +514,7 @@ def _do_export(pulp_exporter, tar, the_export):
         else:
             vers_artifacts = version.artifacts
 
-        if artifacts:
-            artifacts.union(vers_artifacts)
-        else:
-            artifacts = vers_artifacts
+        artifacts.union(vers_artifacts)
 
     # export plugin-version-info
     export_versions(the_export, plugin_version_info)
```

resulted in **ZERO** Artifacts being exported. Made the export smaller/faster, to be sure - but also a tad less useful.

Could be a Django Weirdness, could be a @ggainey Weirdness. Either way, I don't think it's worth holding up this PR to investigate if/when we go through a more comprehensive "let's optimize this whole workflow better" investigation.

ggainey:
> > Clarify please? Do you mean just change to ...pb.iter(artifacts.iterator())? I feel liike I'm missing something .
> 
> iterate over `artifacts.iterator()`, increment the progressbar manually (and probably not one-at-a-time)

Gotcha - good call, results of doing so improve the time even more:
```
"started_at": "2023-07-29T14:06:46.511803Z",
"finished_at": "2023-07-29T14:16:42.599406Z", : 9m56s
```

pulpcore/app/tasks/export.py
ggainey: > * Using a db iterator

Done, gained even more performance back (11m02s went to 9m56s)

> * using qs.none()

Failed in ways I don't currently grok. Will leave for "future investigation"

PR #4176, opened 2023-07-27
Moves TaskGroup.finish() to outside the loop in the replication task

dkliban:
fixes: #4175

### DIFF
--- a/CHANGES/4175.bugfix
+++ b/CHANGES/4175.bugfix
@@ -1,0 +1,2 @@
+Fixed bug where the Task Group for replication would be marked as fully dispatched after just one
+replicator.
--- a/pulpcore/app/tasks/replica.py
+++ b/pulpcore/app/tasks/replica.py
@@ -98,4 +98,4 @@ def replicate_distributions(server_pk):
             distro_names.append(distro["name"])
 
         replicator.remove_missing(distro_names)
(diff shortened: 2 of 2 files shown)


gerrod3 (approved):
This wouldn't affect our tests since we only have one plugin (file) so `supported_replicators` would have a length of one, correct?

dkliban:
> This wouldn't affect our tests since we only have one plugin (file) so `supported_replicators` would have a length of one, correct?

I don't think it affects the tests for the reason you mention.

mdellweg:
We identified that "all_tasks_dispatched" is really not so much helpful, because it usually does not get set in case of an error. So we adjusted our task group waiting algorithms to not care about it.
Maybe we'll remove it someday.

mdellweg (commented):
Should `task_group.finish()` be inside a finally statement?
Does any of the subtasks fork even more tasks?

dkliban:
What is set_failed() checked if the task is the earliest created task of a task group and then set the completed flag on the task group?


https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/task.py#L274

mdellweg (approved):
It is really not sure where the task_group.finish feature is heading. You don't even need it to identify the state of a task group.
This however is more correct than before.

PR #4177, opened 2023-07-28
Remove custom pytest hook and move pytest markers from pulp-smash

hstct:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -94,31 +94,22 @@ def get_bindings_config():
     return configuration
 
 
-@pytest.hookimpl(tryfirst=True)
-def pytest_check_for_leftover_pulp_objects(config):
-    pulpcore_client = ApiClient(get_bindings_config())
-    tasks_api_client = TasksApi(pulpcore_client)
-
-    for task in tasks_api_client.list().results:
-        if task.state in ["running", "waiting"]:
-            raise Exception("This test left over a task in the running or waiting state.")
-
-    apis_to_check = [
-        ContentguardsApi(pulpcore_client),
-        DistributionsApi(pulpcore_client),
-        PublicationsApi(pulpcore_client),
-        RemotesApi(pulpcore_client),
-        RepositoriesApi(pulpcore_client),
-    ]
-    for api_to_check in apis_to_check:
-        if api_to_check.list().count > 0:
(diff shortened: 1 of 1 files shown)


mdellweg (approved):
Although we could try to make this hook implementation optional, i think this is what we discussed, and we can always bring it back once we sorted out the pulp-smash story.

hstct:
This hook will cause the CI pipeline to fail and don't run any functional tests if you don't install `pulp-smash` as a test requirement. From the matrix chat it seems that it is not really used anyway and the consensus was to just remove it.

It will run into this error [here](https://github.com/pulp/pulp_deb/actions/runs/5655044235/job/15319353732#step:10:1721)

With the removal the tests now run again: https://github.com/pulp/pulp_deb/pull/851

dkliban:
is this not used anywhere?

mdellweg:
> is this not used anywhere?

@dkliban not in any CI at least (It contradicts the parallel execution of tests).
If you are a regular user of this feature, we need to find a way to make the implementation of this hook dependent of the existence of its declaration (currently in pulp-smash).

lubosmj:
Can we also move this from pulp_smash: https://github.com/pulp/pulp-smash/blob/2ded6671e0f32c51e70681467199e8a195f7f5fe/pulp_smash/pulp3/pytest_plugin/__init__.py#L76? I think @hstct mentioned that without that, he sees some warnings in the pulp_deb's CI.

mdellweg:
> Can we also move this from pulp_smash: https://github.com/pulp/pulp-smash/blob/2ded6671e0f32c51e70681467199e8a195f7f5fe/pulp_smash/pulp3/pytest_plugin/__init__.py#L76? I think @hstct mentioned that without that, he sees some warnings in the pulp_deb's CI.

Only the second half of it, but yes, we should.

hstct:
> > Can we also move this from pulp_smash: https://github.com/pulp/pulp-smash/blob/2ded6671e0f32c51e70681467199e8a195f7f5fe/pulp_smash/pulp3/pytest_plugin/__init__.py#L76? I think @hstct mentioned that without that, he sees some warnings in the pulp_deb's CI.
> 
> Only the second half of it, but yes, we should.

I can include this to this PR or open a new one whatever is preferred :slightly_smiling_face:

dkliban (approved):

PR #4182, opened 2023-07-28
Made reclaim space task more tolerant in the face of artifact shared …

### LINKED ISSUE #3610: Reclaim space for repository fails with Cannot delete some instances of model 'Artifact' because they are referenced through protected foreign keys: 'ContentArtifact.artifact'."
**Version**
3.16, but probably all versions

**Describe the bug**

- Reclaim space for repository fails with the following error.

  ~~~
  Task paused with error: "("Cannot delete some instances of model 'Artifact' because they are referenced through protected foreign keys: 'ContentArtifact.artifact'.", {<ContentArtifact: pk=452959ad-c045-4e85-bf9f-6651ba37f57d>})"
  ~~~

**To Reproduce**
See BZ

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=2169322

ipanova:
…between two content units.

closes #3610

### DIFF
--- a/pulpcore/app/tasks/reclaim_space.py
+++ b/pulpcore/app/tasks/reclaim_space.py
@@ -1,3 +1,7 @@
+from logging import getLogger
+
+from django.db.models.deletion import ProtectedError
+
 from pulpcore.app.models import (
     Artifact,
     Content,
@@ -8,6 +12,8 @@ from pulpcore.app.models import (
     RepositoryVersion,
 )
 from pulpcore.app.util import get_domain
+
+log = getLogger(__name__)
 
 
 def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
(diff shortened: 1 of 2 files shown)


pulpcore/app/tasks/reclaim_space.py
```diff
@@ -76,10 +82,16 @@ def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
...
-        artifact.delete()
-        progress_bar.done += 1
-        counter += 1
+        try:
+            # we need to manually call delete() because it cleans up the file on the filesystem
+            artifact.delete()
+        except ProtectedError as e:
+            # Rarely artifact could be shared between to different content units.
+            # Just log and skip the artifact deletion in this case
+            log.info(e)
```
dralley, line 91: Should this be a higher level such as WARN?

ipanova: we could, but to be honest i would keep this at info as here https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/purge.py#L146 so we're consistent how we log same error

dralley: ack

dralley:
As an expedient workaround I'm perfectly OK with this.  I still want to figure out if we can address it on a more fundamental level though.

dralley (approved):

ggainey (approved):
Gotta love edge-cases...

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -76,10 +82,16 @@ def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
...
+        else:
```
PaulSD, line 92: Shouldn't this be "finally", not "else"?
The progress bar is tracking the position in `artifacts_to_delete`, so the `done` counter should be incremented when we move to the next artifact, even if this artifact was skipped / wasn't deleted.

ipanova: @PaulSD  there was a little debate on how to treat such situation https://github.com/pulp/pulpcore/pull/4295#discussion_r1307250518
please read and let us know if you have any further feedback.

PaulSD: I understand now.  This does make sense as-is.  Thanks!

PR #4211, opened 2023-07-31
Reset export to using a set() for Artifacts.

### LINKED ISSUE #4210: Export artifact-dedup optimizations break pulp_rpm export.
**Version**
main, 18, 21, 22, 23, 28, 29

**Describe the bug**
After the changes for #4159 were released, pulp_rpm nightly tests started failing. This turns out to be the result of a subtle bug in the optimizations introduced for that fix.

This has to be fixed and released ASAP to all versions #4159 was releasxed into.

**To Reproduce**
* Run https://github.com/pulp/pulp_rpm/blob/main/pulp_rpm/tests/functional/api/test_pulpimport.py#L287

ggainey:
The QuerySet approach introduced a subtle bug the breaks pulp_rpm export. If we want to retain the 'good parts' of that approach, it will need to be its own investigation/PR.

fixes #4210.

### DIFF
--- a/CHANGES/4210.bugfix
+++ b/CHANGES/4210.bugfix
@@ -1,0 +1,1 @@
+Fix a subtle export bug introduced from the optimizations in #4159.
--- a/pulpcore/app/importexport.py
+++ b/pulpcore/app/importexport.py
@@ -94,14 +94,14 @@ def export_artifacts(export, artifacts):
 
     Args:
         export (django.db.models.PulpExport): export instance that's doing the export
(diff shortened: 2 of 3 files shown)


dralley (approved):

ggainey:
> @ggainey Could you describe the issue that was introduced? (maybe on the issue would be a better place than here)

[Done](https://github.com/pulp/pulpcore/issues/4210#issuecomment-1658773631)

mdellweg (approved):

PR #4232, opened 2023-08-01
Added to the fix for 4210 in export.

### LINKED ISSUE #4231: Export artifact-dedup fix from #4210 incomplete
Version
main, 18, 21, 22, 23, 28, 29

Describe the bug
The change for https://github.com/pulp/pulpcore/issues/4210 failed to completely fix the problem uncovered in pulp_rpm.

`django-importexport.export()` apparently **really** wants either a QuerySet or a **list** (specifically).

Fix and backport to the above versions.

ggainey:
fixes #4231.

### DIFF
--- a/CHANGES/4231.bugfix
+++ b/CHANGES/4231.bugfix
@@ -1,0 +1,1 @@
+Correctly fixed pulp_rpm-export-edgecase - fix #4210 was incomplete.
--- a/pulpcore/app/importexport.py
+++ b/pulpcore/app/importexport.py
@@ -94,7 +94,7 @@ def export_artifacts(export, artifacts):
 
     Args:
         export (django.db.models.PulpExport): export instance that's doing the export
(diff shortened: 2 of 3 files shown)


ipanova:
Grant and I try tried to demystify this and ended up both on the same line https://github.com/django-import-export/django-import-export/blob/main/import_export/resources.py#L1089 which should not make difference for set or list.

It's still possible we were chasing red herring.

ipanova:
For posterity, sometimes it fails to find the artifact-on-disk, and the export fails. Sometimes, the export succeeds, but it's missing an Artifact that the ArtifactResource/ContentArtifactResource "expect" to be there, and the import fails
and most of the time, everything works. This only fails in the standard-filesystem env - azure/s3 have never seen this.

ipanova (approved):

dralley (approved):
Ridiculous and deeply confusing, but OK :)

Thanks for debugging!

Issue #4183, opened 2023-07-28
Conflicting Domain creation at the same time results in an ISE (500)


jlsherrill:
**Version**
3.29.1

**Describe the bug**
if multiple requests are dispatched to pulp at the same time to create a domain with the same name, pulp will return an ISE instead of a proper 400 (bad request).  

**To Reproduce**
trigger multiple requests:

```
http --auth admin:password  localhost:8080/pulp/default/api/v3/domains/  name=foo2  storage_class="pulpcore.app.models.storage.FileSystem"  storage_settings[location]="/var/lib/pulp/foo2"  &
http --auth admin:password  localhost:8080/pulp/default/api/v3/domains/  name=foo2  storage_class="pulpcore.app.models.storage.FileSystem"  storage_settings[location]="/var/lib/pulp/foo2" &
```
One will result in an ISE, while the other creates the domain.  If i try again after its created, i get a proper 400 Bad Request.

**Expected behavior**

I'd expect a 400 or 409 

**Additional context**
We can't pre-create the domains, nor can we control when the requests come in that need a domain created, so we can't avoid this situation.  We've worked around it for now, so i wouldn't consider this a high priority, but it does prevent us from detecting actual 500s if they were to occur as easily.  (Right now we just assume its a conflict).

PR #4245, opened 2023-08-02
Fix 500 error on parallel same-object creates

gerrod3:
fixes: #4183

Not sure how to write a test for this. I manually tested this with 
```
http :5001/pulp/api/v3/repositories/file/file/ name=foo & http :5001/pulp/api/v3/repositories/file/file/ name=foo
```
Roughly 1 in 3 attempts will fail with a 500 error, this fixes the race condition and will give a nice 400 error. 

@jlsherrill Interestingly you found this bug when creating domains, but it actually affects all immediate-creation objects in Pulp. :open_mouth:

### DIFF
--- a/CHANGES/4183.bugfix
+++ b/CHANGES/4183.bugfix
@@ -1,0 +1,1 @@
+Fixed a 500 server error when two concurrent requests try to create an object with the same unique fields.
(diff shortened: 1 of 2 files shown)


mdellweg (approved):
Is this maybe interesting for drf too?

dkliban (approved):

jlsherrill:
Thanks @gerrod3 !

PR #4249, opened 2023-08-03
Add records for Api app status

### LINKED ISSUE #3175: pulpcore-api worker records are not in the database
## Problem

We record the heartbeats of workers and content apps. But we do not collect status information (via heartbeat) from any pulp-api service.

## Proposal

1. Create a model for pulp-api, similar to what we have for the [ContentAppStatus](https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/status.py#L31) and [Worker](https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/task.py#L79).

2. Have the pulp-api service use a thread to record heartbeats periodically.

## Is this user facing?

No, to keep things as simple as possible this first ticket is just about adding it in the database?

## If users can't see it, why are we doing this?

https://github.com/pulp/pulpcore/issues/3401

mdellweg:
fixes #3175


mdellweg:
The current version of the change required, that gunicorn is started with "--worker-class pulpcore.app.gunicorn_worker.PulpApiWorker`.

mdellweg:
I redesigned and pulpcore will now provide a `pulpcore-api` entrypoint that is basically a preloaded gunicorn.
How much of the options should we move there? E.g. how about the logging-format?
Also any way we do it requires a change to the startup files in the oci images.

pulpcore/api_gunicorn.py
```diff
@@ -0,0 +1,91 @@
...
+class PulpcoreApiApplication(WSGIApplication):
+    def init(self, parser, opts, args):
+        super().init(parser, opts, args)
+        self.app_uri = "pulpcore.app.wsgi:application"
+        self.cfg.set("default_proc_name", "pulpcore-api")
+        self.cfg.set("worker_class", "pulpcore.api_gunicorn.PulpApiWorker")
+
+
+def gunicorn():
+    PulpcoreApiApplication("%(prog)s [OPTIONS]").run()
```
mdellweg, line 91: I need confirmation that we want to start running our components this way.
So if two @pulp/core devs could ack this specific part please?

ggainey: "confirmation that we want to start running our components this way."

I guess my only real question is, is there some **downside** to this? I can't think of one, but I'm prob just being dense.

mdellweg: Maybe that it's another layer of abstraction. But that goes on the plus side as well. When i think about our plan to redesign the content app, it will be much easier to just change the gunicorn worker type used in the entrypoint than telling people to modify their service files.
And exactly that thought brings me to the next question: How much should our entrypoint mimick the gunicorn commandline? Should we just override some parameters internally, or should we explicitely proxy only some parameters?

dkliban:
We need to make sure that users can still specify gunicorn timeout, number of workers, and logging format.

pulpcore/app/entrypoint.py
```diff
@@ -0,0 +1,137 @@
...
+        using_pulp_api_worker.set(True)
+        super().init_process()
+
+    def run(self):
+        try:
+            super().run()
+        finally:
+            # cleanup
+            if self.api_app_status:
+                self.api_app_status.delete()
```
gerrod3, line 81: Is it possible we could close/kill the worker and not delete the record?

pulpcore/app/entrypoint.py
gerrod3, line 1: Can you link the appropriate gunicorn documentation for the classes we import and options we are allowed to pass in?

pulpcore/app/models/status.py
```diff
@@ -11,35 +11,55 @@
...
-class ContentAppStatusManager(models.Manager):
+class AppStatusManager(models.Manager):
     def online(self):
         """
-        Returns a queryset of ``ContentAppStatus`` objects that are online.
+        Returns a queryset of objects that are online.
 
-        To be considered 'online', a ContentAppStatus must have a heartbeat timestamp within
+        To be considered 'online', a AppStatus must have a heartbeat timestamp within
         ``settings.CONTENT_APP_TTL`` from now.
```
gerrod3, line 20:
```suggestion
        ``self.model.APP_TTL`` from now.
```

pulpcore/app/models/status.py
```diff
@@ -11,35 +11,55 @@
...
-
+        age_threshold = timezone.now() - self.model.APP_TTL
         return self.filter(last_heartbeat__gte=age_threshold)
 
+    def missing(self, age=None):
+        """
+        Returns a queryset of workers meeting the criteria to be considered 'missing'
+
+        To be considered missing, a worker must have a stale timestamp.  By default, stale is
+        defined here as longer than the ``settings.WORKER_TTL``, or you can specify age as a
```
gerrod3, line 34:
```suggestion
        defined here as longer than the ``self.model.APP_TTL``, or you can specify age as a
```
Might want to consider rewriting this doc-string to change 'worker' to 'object' since this method now refers to the content app, pulpcore-api & task workers.

pulpcore/content/__init__.py
```diff
@@ -35,41 +35,55 @@
...
-            msg = (
-                "Content App '{name}' failed to write a heartbeat to the database, sleeping for "
-                "'{interarrival}' seconds."
-            ).format(name=name, interarrival=heartbeat_interval)
-            log.info(msg)
-        await asyncio.sleep(heartbeat_interval)
+async def _heartbeat_ctx(app):
+    heartbeat_task = asyncio.create_task(_heartbeat())
+    yield
+    heartbeat_task.cancel()
```
gerrod3, line 77: Nifty, this should remove the asyncio Error 'Task was destroy but it is pending' from the logs when restarting the content app?

pulpcore/app/models/status.py
```diff
@@ -56,9 +76,7 @@ def online(self):
         Returns:
             bool: True if the content app is considered online, otherwise False
```
gerrod3, line 77: s/content app/app. Also, you need to change the docstring for the `missing` property as well.

gerrod3 (changes requested):

pulpcore/content/__init__.py
mdellweg: I hope so...

pulpcore/app/models/status.py
mdellweg: We call them api workers and content workers... And gunicorn talks a lot about workers too. I think we are probably fine...

pulpcore/app/entrypoint.py
mdellweg: Why should we? Also there was a request from @bmbouter when implementing the hearbeating for pulpcore-workers to clean up on exiting. And this PR tries to apply these rules to all types of workers.

And btw we never cleaned up the content api status before. I guess there are some db's around where that table became humongous by now.

pulpcore/app/entrypoint.py
```diff
@@ -0,0 +1,137 @@
...
+            self.api_app_status, created = self.ApiAppStatus.objects.get_or_create(
+                name=self.name, defaults={"versions": self.versions}
+            )
+
+            if not created:
+                self.api_app_status.save_heartbeat()
+
+                if self.api_app_status.versions != self.versions:
+                    self.api_app_status.versions = self.versions
+                    self.api_app_status.save(update_fields=["versions"])
```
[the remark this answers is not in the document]
mdellweg: In theory yes. The name is composed of `<pid>@<hostname>` when you restart that computer, and the boot process is very consistent, you will end up with the same name again.

pulpcore/app/entrypoint.py
gerrod3: Oh sorry, I worded my question wrong. I was asking is it possible the worker could be kill and not have its record cleaned up. Looking later down in the code, it seems that the task worker code would clean up any missing worker records.

pulpcore/app/wsgi.py
```diff
@@ -7,14 +7,14 @@
...
-
 from django.core.wsgi import get_wsgi_application
 from opentelemetry.instrumentation.wsgi import OpenTelemetryMiddleware
 
+from pulpcore.app.entrypoint import using_pulp_api_worker
 
-os.environ.setdefault("DJANGO_SETTINGS_MODULE", "pulpcore.app.settings")
 
+if not using_pulp_api_worker.get(False):
+    raise RuntimeError("This app must be executed using pulpcore-api endpoint.")
```
gerrod3, line 17:
```suggestion
    raise RuntimeError("This app must be executed using pulpcore-api entrypoint.")
```

pulpcore/app/entrypoint.py
mdellweg: The cleanup happens after a week (or so). But yes, it's always possible that the application is killed in a way that no cleanup procedures will be executed anymore. OOM, kill -9, segfault, you name it...

gerrod3 (approved):

ggainey (approved):

Issue #4247, opened 2023-08-02
Replication tls validation doesnt get applied to each remote


PotentialIngenuity:
**Version**
core - 3.29.7
rpm - 3.22.3

**Describe the bug**
The `tls_validation` field of each remote is always `true`. Running replicate with `tls_validation:=false` has no effect on the remote being synced.
```
{
  "pulp_href": "/pulp/api/v3/remotes/rpm/rpm/0189ad3c-7c43-7fb4-b848-3bc72049734d/",
  "pulp_created": "2023-07-31T18:36:09.667754Z",
  "name": "rhel8/8/x86_64/codeready-builder/os/2023/30",
  "url": "https://repo.company.com/pulp/content/rhel8/8/x86_64/codeready-builder/os/2023/30/",
  "ca_cert": null,
  "client_cert": null,
  "tls_validation": true,
  "proxy_url": null,
  "pulp_labels": {},
  "pulp_last_updated": "2023-08-02T03:20:17.005804Z",
  "download_concurrency": null,
  "max_retries": null,
  "policy": "immediate",
  "total_timeout": null,
  "connect_timeout": null,
  "sock_connect_timeout": null,
  "sock_read_timeout": null,
  "headers": null,
  "rate_limit": null,
  "hidden_fields": [
    {
      "name": "client_key",
      "is_set": false
    },
    {
      "name": "proxy_username",
      "is_set": false
    },
    {
      "name": "proxy_password",
      "is_set": false
    },
    {
      "name": "username",
      "is_set": false
    },
    {
      "name": "password",
      "is_set": false
    }
  ]
}
```

PR #4250, opened 2023-08-03
Sets tls settings on replication remotes

dkliban:
fixes: #4247

### DIFF
--- a/pulpcore/app/replica.py
+++ b/pulpcore/app/replica.py
@@ -28,7 +28,7 @@ class Replicator:
     app_label = None
     sync_task = None
 
-    def __init__(self, pulp_ctx, task_group, ca_cert=None):
+    def __init__(self, pulp_ctx, task_group, tls_settings):
         """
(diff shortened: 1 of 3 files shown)


pulpcore/app/replica.py
```diff
@@ -83,14 +83,15 @@ def create_or_update_remote(self, upstream_distribution):
         if not upstream_distribution["repository"] and not upstream_distribution["publication"]:
             return None
         url = self.url(upstream_distribution)
+        remote_fields_dict = {"url": url}
+        remote_fields_dict.update(self.tls_settings)
+        remote_fields_dict.update(self.remote_extra_fields(upstream_distribution))
+
         # Check if there is a remote pointing to this distribution
         try:
             remote = self.remote_model_cls.objects.get(
```
mdellweg, line 92: looks like a perfect use for `get_or_create`.

mdellweg (approved):
I can see how this works, but there are a couple of suggestions.

PR #4251, opened 2023-08-03
Fixed content-handler response-headers/object-storage collision.

### LINKED ISSUE #4028: Content-app doesn't respect response-headers specified by plugins
**Version**
3.28

**Describe the bug**
core/3.28 added Distribution.content_headers_for() to allow plugins to specifiy response-headers. When using non-FileStorage backends, these headers are ignored.

pulp_rpm PR https://github.com/pulp/pulp_rpm/pull/3193 , for example, fails on azure and S3 because the cache-disposition header isn't ending up in the response.

**To Reproduce**
* start with core/3.28 using a minio-storage-backend
*  install pulp_rpm with https://github.com/pulp/pulp_rpm/pull/3193
*  create, sync, and distribute a pulp_rpm repo
* wget the repomd.xml file from that distribution
* note the absence of a cache-control header.
* be sad.

**Expected behavior**
headers added by a plugin should be made available in the response

**Additional context**
With azure and google, looks like we can just feed the headers into response-parameters. S3 may need specific code.

ggainey:
fixes #4028.

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -20,6 +20,7 @@ from asgiref.sync import sync_to_async
 
 import django
 
+from pulpcore.constants import STORAGE_RESPONSE_MAP
 from pulpcore.responses import ArtifactResponse
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "pulpcore.app.settings")
@@ -894,6 +895,16 @@ class Handler:
         Returns:
             The :class:`aiohttp.web.FileResponse` for the file.
         """
+
+        def _set_params_from_headers(hdrs, storage_domain):
+            # Map standard-response-headers to storage-object-specific keys
+            params = {}
+            if storage_domain in STORAGE_RESPONSE_MAP:
+                for a_key in STORAGE_RESPONSE_MAP[storage_domain]:
+                    if hdrs.get(a_key, None):
+                        params[STORAGE_RESPONSE_MAP[storage_domain][a_key]] = hdrs[a_key]
+            return params
+
         artifact_file = content_artifact.artifact.file
         artifact_name = artifact_file.name
         filename = os.path.basename(content_artifact.relative_path)
@@ -909,9 +920,8 @@ class Handler:
         elif not domain.redirect_to_object_storage:
             return ArtifactResponse(content_artifact.artifact, headers=headers)
         elif domain.storage_class == "storages.backends.s3boto3.S3Boto3Storage":
-            parameters = {"ResponseContentDisposition": content_disposition}
-            if headers.get("Content-Type"):
-                parameters["ResponseContentType"] = headers.get("Content-Type")
+            headers["Content-Disposition"] = content_disposition
+            parameters = _set_params_from_headers(headers, domain.storage_class)
             url = URL(
                 artifact_file.storage.url(
                     artifact_name, parameters=parameters, http_method=request.method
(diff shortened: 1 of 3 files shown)


pulpcore/content/handler.py
```diff
@@ -894,12 +894,48 @@ async def _serve_content_artifact(self, content_artifact, headers, request):
...
+        }
+        # https://gcloud.readthedocs.io/en/latest/storage-blobs.html
+        # response-headers Google Cloud Storage respects, and what they map to in a GCS object
+        _GCS_RESPONSE_HEADER_MAP = {
+            "Content-Disposition": "content_disposition",
+            "Content-Type": "content_type",
+            "Cache-Control": "cache_control",
+            "Content-Language": "content_language",
+            "Content-Encoding": "content_encoding",
+        }
```
lubosmj, line 924: This might be more favourable to store in the `constants.py` file. Is it worth caching the return value of `_set_params_from_headers` (the `parameters` variable will be returned instead of overwriting it inside the function)?

ggainey: These both sound like fine ideas, will do!

pulpcore/content/handler.py
```diff
@@ -894,6 +895,16 @@ async def _serve_content_artifact(self, content_artifact, headers, request):
         Returns:
             The :class:`aiohttp.web.FileResponse` for the file.
         """
+
+        def _set_params_from_headers(hdrs, storage_domain):
```
mdellweg, line 899: When we call this function, we have identified the storage class already, so i think we can spare ourselves one if and dict lookup by providing the response map instead of the storage class name.
```python
def _set_params_from_headers(headers, response_header_map):
    params = {}
    for key, new_key in response_header_map.items():
        if key in headers:
            params[new_key] = headers[key]
```
or even:
```python
def _set_params_from_headers(headers, response_header_map):
    return {new_key: headers[key] for key, new_key in response_header_map.items() if key in headers}
```
We'd call it by, e.g. in the s3 branch:
```python
_set_params_from_headers(headers, S3_RESPONSE_HEADER_MAP)
```

ggainey: Passing the mapping-dictionary "assumes" that a simple lookup is all we're ever going to want to do. If we want/need to do anything more complicated in the future, the current approach lets us do so in a single place.

In terms of performance - given the current state of the content-handler, sparing "one dictionary lookup" is not convincing.

mdellweg: I was thinking about my performance in understanding the code. But it's not going to block this contribution.

mdellweg (approved):

lubosmj (approved):
I like the change!

Optionally, we could remove the `if storage_domain in STORAGE_RESPONSE_MAP` condition and convert it to a try/except block for the for loop (https://www.oreilly.com/library/view/python-essentials/9781784390341/ch09s08.html).

ggainey:
> I like the change!
> 
> Optionally, we could remove the `if storage_domain in STORAGE_RESPONSE_MAP` condition and convert it to a try/except block for the for loop (https://www.oreilly.com/library/view/python-essentials/9781784390341/ch09s08.html).

I've heard that argument before, but personally feel exceptions should be for exceptional issues - not a main-flow "is this one of the things we know how to deal with?" path. Speed doesn't impress me either - no matter how fast python is at exceptions, building one, executing the jump, and then processing it, isn't as fast as "single key lookup into a (tiny) dictionary".

"try it and see" makes more sense the more complicated the possibilities are.

(Or it makes sense until you start fielding "OMG FIX THIS ERROR" issues from users seeing database-logging saying "you can't do that" for things you handle via "ask forgiveness" processing...)

Issue #4157, opened 2023-07-26
Domain delete hook raises an error incorrectly


gerrod3:
This line right here is wrong: https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/domain.py#L55 
ProtectedError requires two arguments: https://github.com/django/django/blob/7bb741d787ba360a9f0d490db92e22e0d28204ed/django/db/models/deletion.py#L10

PR #4267, opened 2023-08-09
Domain delete hook error fix

MichalPysik:
Fixed an exception that gets raised inside the domain delete hook when there is active content in the domain.

fixes: #4157

### DIFF
--- a/pulpcore/app/models/domain.py
+++ b/pulpcore/app/models/domain.py
@@ -3,6 +3,7 @@ from django.db import models
 from django_lifecycle import hook, BEFORE_DELETE, BEFORE_UPDATE
 
 from pulpcore.app.models import BaseModel, AutoAddObjPermsMixin
+from pulpcore.exceptions import DomainProtectedError
 
 from .fields import EncryptedJSONField
 
@@ -51,8 +52,9 @@ class Domain(BaseModel, AutoAddObjPermsMixin):
 
     @hook(BEFORE_DELETE, when="name", is_not="default")
     def _cleanup_orphans_pre_delete(self):
-        if self.content_set.exclude(version_memberships__isnull=True).exists():
-            raise models.ProtectedError("There is active content in the domain.")
+        protected_content_set = self.content_set.exclude(version_memberships__isnull=True)
+        if protected_content_set.exists():
+            raise DomainProtectedError()
         self.content_set.filter(version_memberships__isnull=True).delete()
         for artifact in self.artifact_set.all().iterator():
             # Delete on by one to properly cleanup the storage.
--- a/CHANGES/4157.bugfix
+++ b/CHANGES/4157.bugfix
@@ -1,0 +1,2 @@
+Fixed an exception that gets raised inside the domain delete hook
+when there are still repositories with content in the domain.
--- a/pulpcore/exceptions/__init__.py
+++ b/pulpcore/exceptions/__init__.py
@@ -4,6 +4,7 @@ from .base import (
     ResourceImmutableError,
     TimeoutException,
     exception_to_dict,
+    DomainProtectedError,
 )
 from .validation import (
     DigestValidationError,
--- a/pulpcore/exceptions/base.py
+++ b/pulpcore/exceptions/base.py
@@ -85,3 +85,16 @@ class TimeoutException(PulpException):
         return _(
             "Request timed out for {}. Increasing the total_timeout value on the remote might help."
         ).format(self.url)
+
+
+class DomainProtectedError(PulpException):
+    """
+    Exception to signal that a domain the user is trying to delete still contains
+    repositories with content.
+    """
+
+    def __init__(self):
+        super().__init__("PLP0007")
+
+    def __str__(self):
+        return _("You cannot delete a domain that still contains repositories with content.")


mdellweg (dismissed):

pulpcore/app/models/domain.py
```diff
@@ -51,8 +51,11 @@ def prevent_default_deletion(self):
 
     @hook(BEFORE_DELETE, when="name", is_not="default")
     def _cleanup_orphans_pre_delete(self):
-        if self.content_set.exclude(version_memberships__isnull=True).exists():
-            raise models.ProtectedError("There is active content in the domain.")
+        protected_content_set = self.content_set.exclude(version_memberships__isnull=True)
+        if protected_content_set.exists():
+            raise models.ProtectedError(
+                "There is active content in the domain.", protected_content_set
```
lubosmj, line 57: I am starting to think that listing the PKs of content might not be very informative for a user. Perhaps, doing "[self.something]" is better in this case?

gerrod3: I think it is helpful. This will list all content still being used and since it is from the base class it doesn't have any other identifying characteristics. Note the user can find these content units using the `pulp_id__in` filter on the content endpoint.

dkliban:
@MichalPysik can you share an example of the error message? I am curious to see if prints HREFs or PKs.

pulpcore/app/models/domain.py
```diff
@@ -51,8 +51,15 @@ def prevent_default_deletion(self):
...
     def _cleanup_orphans_pre_delete(self):
-        if self.content_set.exclude(version_memberships__isnull=True).exists():
-            raise models.ProtectedError("There is active content in the domain.")
+        protected_content_set = self.content_set.exclude(version_memberships__isnull=True)
+        if protected_content_set.exists():
+            raise models.ProtectedError(
+                "There is active content in the domain. Use the 'List content' endpoint while"
+                " entering the shown pk(s) of the content into the pulp_id__in filter to obtain"
+                " its pulp_href and then be able to find repositories containing the content."
+                " For deleting content not present in any repository, you can use orphan cleanup.",
```
mdellweg, line 60: Actually, why making it so complicated?
You want to delete the domain. So you just need to delete all the repositories and there will no longer be any active content. (You can't keep the repos anyway.)

MichalPysik:
> @MichalPysik can you share an example of the error message? I am curious to see if prints HREFs or PKs.
> 
> We should return a list of HREFs so the user can easily use it to look up content.


It currently returns PKs of the content units and the cleaning itself is left to the user.

The problem is that ```ProtectedError``` expects an object of type ```Set[Model]`` as the second argument`, which is then simply printed, and formatting the printed text is impossible without replacing the ```ProtectedError``` with a custom error type.

The text (example) printed to user's terminal (excluding my text) currently looks like this:
```<BulkTouchQuerySet [<Content (pulp_type=file.file): pk=0189f8f2-2e29-772b-9b45-7f23d3ff396d>]>```

**Comment by ipanova** [comment]
@MichalPysik  we just discussed at the pulpcore meeting that it might be worth adding a force flag to the domain delete that would also delete repos. This might save manual steps for the user who wants to delete a domain and has few hundred of repos to manually delete before actually being able to delete the domain. Would you mind opening a separate feature?

**Comment by dkliban** [comment]
@MichalPysik, Could you please update the message returned to the user to include something along the lines of 'You must remove all repositories from the domain before removing the domain. If there are no repositories, there is only orphaned content.

**Comment by MichalPysik** [comment]
> @MichalPysik we just discussed at the pulpcore meeting that it might be worth adding a force flag to the domain delete that would also delete repos. This might save manual steps for the user who wants to delete a domain and has few hundred of repos to manually delete before actually being able to delete the domain. Would you mind opening a separate feature? Edit: there is already such! #3362

@ipanova That's good to hear! Do you think it might be worth adding something along the lines of "other then removing the repositories yourself, you can ask the administrator...." to the message, or is that too much information for an error message?

**Comment by ipanova** [comment]
@MichalPysik I'd keep the error message short, as in https://github.com/pulp/pulpcore/pull/4267#issuecomment-1688208828

**Comment by dkliban** [comment]
I was thinking that this could be just a generic PulpException that says "You must remove all repositories from the domain before removing the domain." 

models.ProtectedError doesn't seem appropriate for the user to see.

File: pulpcore/app/models/domain.py [hunk 3]
```diff
@@ -3,6 +3,7 @@
 from django_lifecycle import hook, BEFORE_DELETE, BEFORE_UPDATE
 
 from pulpcore.app.models import BaseModel, AutoAddObjPermsMixin
+from pulp_glue.common.context import PulpException
```
dkliban, line 6: I  believe you want to import from `pulpcore` not from `pulp_glue`.[0]



[0] https://github.com/pulp/pulpcore/blob/main/pulpcore/exceptions/base.py#L5

pulpcore/app/models/domain.py
```diff
@@ -51,8 +52,12 @@ def prevent_default_deletion(self):
 
     @hook(BEFORE_DELETE, when="name", is_not="default")
     def _cleanup_orphans_pre_delete(self):
-        if self.content_set.exclude(version_memberships__isnull=True).exists():
-            raise models.ProtectedError("There is active content in the domain.")
+        protected_content_set = self.content_set.exclude(version_memberships__isnull=True)
+        if protected_content_set.exists():
+            raise PulpException(
```
dkliban, line 57: You will need to use a error code. I believe we are now up to `PLP0006`.

dkliban (approved):
Thank you! This is great!

pulpcore/app/models/domain.py
```diff
@@ -51,8 +52,9 @@ def prevent_default_deletion(self):
 
     @hook(BEFORE_DELETE, when="name", is_not="default")
     def _cleanup_orphans_pre_delete(self):
-        if self.content_set.exclude(version_memberships__isnull=True).exists():
-            raise models.ProtectedError("There is active content in the domain.")
+        protected_content_set = self.content_set.exclude(version_memberships__isnull=True)
+        if protected_content_set.exists():
+            raise DomainProtectedError()
```
lubosmj, line 57: Is the instance of `PulpException` translated to a user-readable form or a 500 internal server error?

MichalPysik: @lubosmj the http request sent by user returns 202 (a task is created successfully), the task itself then fails and raises the exception. If the user uses the CLI, the message gets printed in the terminal, otherwise it is to be found in the logs.

PR #4270, opened 2023-08-11
Expression filter

### LINKED ISSUE #2480: Complex filtering

Currently filtering data from list-like API endpoints is only possible using params like `?name_includes`, etc.
AFAICT these can NOT be chained, or include any AND/OR logic.

For example, we'd like to filter roles to only show a relevant subset of predefined roles, while still showing all custom roles.
Custom vs predefined is addressed by the `locked` field, so we'd like to filter by `((locked=True && name starts_with "galaxy") OR locked=False) AND any simple user filters`.

Doing so via multiple API requests is already possible, but not if you also need paging to work (as the client would have to deal with merging the data in the right order, loading from multiple buckets with different offsets on eveyr page change, etc.).

Add a `?filter` param to allow adding complex filters to API requests.

Presumably django already has a way of specifying these, so using the same language (but NOT SQL!) might make sense...

Let's assume something like...

```
OR
  - equals
    - locked
    - False
  - AND
    - equals
      - locked
      - True
    - starts_with
      - name
      - "galaxy"
```

possibly serialized as `?fillter={"OR": [ "equals": [ "locked", "False" ], "AND": [ "equals": [ "locked", "true" ], "starts_with": [ "name", "galaxy" ], ] ]}`

We do need paging support, so the only real alternative is creating a custom API endpoint wherever we need complex filtering.
(Unless I'm completely missing something :).)

### LINKED ISSUE #3914: Allow OR filters
We would like to include more advanced filtering options in the Automation Hub UI that allow for users to filter API results using OR lookups, for example I should be able to filter results using expressions such as `name__icontains=foo OR name__icontains=bar`.

There are a number of possible options to consider for this:

- ORing multiple params with the same name (eg `?name=foo&name=bar`). This would break backwards compatibility with the ruby and python bindings, as the open api spec type field would change to a list type.
- Provide a character delimited list (eg `?name=foo,bar`, `?name=foo|bar`). `|` may make more sense, as `,` are used for list values in other parts of the API.
- Adding `__in` to the list of default lookup expressions (eg `?name__in=foo,bar`). This may not work with more complex lookups like `name__icontains__in=`.

**Additional context**
The AAP team is work on unifying the UI experience across the ansible platform, and this is one capability supported by the other platform components (such as ansible controller), but is currently missing in Automation Hub

mdellweg:
fixes #2480
fixes #3914


pulpcore/filters.py
```diff
@@ -150,6 +152,57 @@ def pulp_type_choices(model):
...
+
+            def group_action(tokens):
+                key = tokens[0].key
+                value = tokens[0].value
+                filter = self.parent.filters[key]
+                if isinstance(filter, ExpressionFilter):
+                    raise serializers.ValidationError("Recursion: see Recursion!")
+                if isinstance(filter, filters.OrderingFilter):
+                    raise serializers.ValidationError(
+                        "order You the query filter. cannot the inside q"
```
gerrod3, line 171: Is this error message a pun? Kind of funny even if not intentional.

mdellweg: Intentional for the time being. See the todo above. Currently these errors bubble up as 500, while they should be 400 or 422.

newswangerd:
@jamestalton does this work for the ansible ui?

ggainey (changes requested):
I'd like to see doc with examples of how-to-use - it's pretty opaque to the end-user just looking at the REST API, yeah?

mdellweg:
> I'd like to see doc with examples of how-to-use - it's pretty opaque to the end-user just looking at the REST API, yeah?

Should we just take the queries from the tests as examples?

ggainey:
> > I'd like to see doc with examples of how-to-use - it's pretty opaque to the end-user just looking at the REST API, yeah?
> 
> Should we just take the queries from the tests as examples?

That would work fine - just put them where REST users don't have to dig into code to understand how to use this.

docs/workflows/advanced-filtering.rst
ggainey, line 1: :+1:

ggainey (approved):

docs/workflows/advanced-filtering.rst
```diff
@@ -0,0 +1,33 @@
...
+    EXPRESSION = SIMPLE_EXPR | "(" SIMPLE_EXPR | COMPOSIT_EXPR ")"
+    COMPOSIT_EXPR = NOT_EXPR | AND_EXPR | OR_EXPR
+    NOT_EXPR = "NOT" WS EXPRESSION
+    AND_EXPRESSION = EXPRESSION "AND" EXPRESSION
+    OR_EXPRESSION = EXPRESSION "OR" EXPRESSION
+    SIMPLE_EXPRESSION = FILTERNAME "=" STRING
+    STRING = SIMPLE_STRING | QUOTED_STRING
+
+Some example ``q`` expressions are::
+
```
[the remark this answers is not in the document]
lubosmj: You should use the `q` parameter, like so:
```
http :5001/pulp/api/v3/tasks/?q='name__contains=a AND (NOT state="failed")'
```

I also had a hard time understanding it in the first place. Overall, the functionality is very promising to me. 👍🏻

lubosmj (commented):
From the functional perspective, I adore this change. Everything is working as expected. Longer and more complex queries take some time to complete but this is fine, e.g.:
```
# valid
http :5001/pulp/api/v3/tasks/?q="(((((NOT name__contains='Horror' OR (NOT name__contains='Thriller')) AND (name__contains='Suspense' OR name__contains='Mystery')) AND ((name__contains='Detective' OR name__contains='Crime') AND (NOT name__contains='Action'))) OR (name__contains='Intrigue' AND (NOT name__contains='Suspense'))) OR ((name__contains='Dilemma' OR name__contains='Tension') AND (NOT name__contains='Conflict' OR name__contains='Resolution'))) AND (name__contains='Climax' OR (NOT name__contains='Conclusion'))"
```
```
# invalid
http :5001/pulp/api/v3/tasks/?q="((((((NOT name__contains='Adventure' OR (NOT name__contains='Mystery')) AND (name__contains='Suspense' OR name__contains='Detective')) AND ((name__contains='Crime' OR name__contains='Intrigue') AND (NOT name__contains='Action'))) OR (name__contains='Enigma' AND (NOT name__contains='Suspense'))) OR ((name__contains='Dilemma' OR name__contains='Tension') AND (NOT name__contains='Puzzle' OR name__contains='Resolution'))) AND (name__contains='Climax' OR (NOT name__contains='Conclusion'))"
```

Are regular users the ones who can invoke such a complex query? Is this limited to admins? The filtering could block the API worker, right? Is the worry about DoS in place?

ggainey:
> Are regular users the ones who can invoke such a complex query? Is this limited to admins? The filtering could block the API worker, right? Is the worry about DoS in place?

It's just another parameter on the list-endpoints. Yes, 'regular' users can call those endpoints. Yes, that might be a problem. That's an artifact of enabling this kind of reporting/searching. Given the requirements for this feature, it can't reasonably be limited to admin-only (and would need to be a separate endpoint if we wanted to do so).

lubosmj:
Cool. Then, a small warning in the docs might be the only missing thing.

mdellweg:
I have issues writing a warning with absolutely no hint what to do about it. It's not like the admin has any choice here.

mdellweg:
> > Are regular users the ones who can invoke such a complex query? Is this limited to admins? The filtering could block the API worker, right? Is the worry about DoS in place?
> 
> It's just another parameter on the list-endpoints. Yes, 'regular' users can call those endpoints. Yes, that might be a problem. That's an artifact of enabling this kind of reporting/searching. Given the requirements for this feature, it can't reasonably be limited to admin-only (and would need to be a separate endpoint if we wanted to do so).

Yes this is a feature for users. By design. I do not see any value in it if you cannot use it.

mdellweg:
I added a complexity limit to address all the remaining concerns.

pulpcore/filters.py
```diff
@@ -150,6 +153,110 @@ def pulp_type_choices(model):
...
+                    (pp.Suppress(pp.Keyword("NOT")), 1, pp.opAssoc.RIGHT, self._NotAction),
+                    (pp.Suppress(pp.Keyword("AND")), 2, pp.opAssoc.LEFT, self._AndAction),
+                    (pp.Suppress(pp.Keyword("OR")), 2, pp.opAssoc.LEFT, self._OrAction),
+                ],
+            )
+            try:
+                result = expr.parse_string(value, parse_all=True)[0]
+            except pp.ParseException:
+                raise forms.ValidationError(_("Syntax error in expression."))
+            if result.complexity > 8:
```
ggainey, line 241: :+1:  I think this is a good compromise between "usefulness" and "protect the pulp instance from inadvertent DDoS". I def want to see a mention in the doc for this along the lines of " "to provide some protection from overly-complex queries stalling the database, the Q() filter limits query-complexity to 8 AND/OR/NOT combinations"

ggainey (changes requested):

docs/workflows/advanced-filtering.rst
```diff
@@ -0,0 +1,38 @@
+Advanced Filtering
+==================
+
+.. warning::
+
+   In order to prevent arbitrarily complex queries, the maximum complexity of the expressions
+   explained below is limited.
+
```
mdellweg, line 8: @ggainey Isn't this the mention you requested?

ggainey: The code is specific, the docs should be as well. It's rude to force users to do things like "well, *that* expression was OK, but *this* one isn't - what makes something 'too complex'?"

lubosmj (approved):
I am giving the approval from the functional perspective. 🧑‍🍳 Please, address the comment from @ggainey.

docs/workflows/advanced-filtering.rst
mdellweg: I wanted to be a bit unspecific, because we may decide to change the way we compute complexity while we learn more from the field. But I'll add a sketch of the calculation.

docs/workflows/complex-filtering.rst
```diff
@@ -0,0 +1,57 @@
...
+    # complexity: 3 = 1 (AND/OR) + 2 (filter expression)
+
+    pulp_type="core.rbac" AND name__icontains=gg AND NOT name__contains=HH
+    # complexity: 5 = 1 (AND/OR) + 1 (NOT) + 3 (filter expression)
+
+    NOT (pulp_type="core.rbac" AND name__icontains=gGgG)
+    pulp_type="core.rbac" AND NOT name__contains="naïve"
+    # complexity: 4 = 1 (AND/OR) + 1 (NOT) + 2 (filter expression)
+
+    pulp_type="core.rbac" AND(   name__icontains=gh OR name__contains="naïve")
```
ggainey, line 52:
```suggestion
    pulp_type="core.rbac" AND (name__icontains="gh" OR name__contains="naïve")
```

docs/workflows/complex-filtering.rst
```diff
@@ -0,0 +1,57 @@
...
+    EXPRESSION = FILTER_EXPR | NOT_EXPR | AND_EXPR | OR_EXPR | "(" EXPRESSION ")"
+    NOT_EXPR = "NOT" EXPRESSION
+    AND_EXPRESSION = EXPRESSION "AND" EXPRESSION
+    OR_EXPRESSION = EXPRESSION "OR" EXPRESSION
+    FILTER_EXPRESSION = FILTERNAME "=" STRING
+    STRING = SIMPLE_STRING | QUOTED_STRING
+
+Some example ``q`` expressions and their complexity are::
+
+    pulp_type__in='core.rbac'
```
ggainey, line 30:
```suggestion
    pulp_type__in="core.rbac"
```

docs/workflows/complex-filtering.rst
```diff
@@ -0,0 +1,57 @@
...
+    # complexity: 1 = 1 (filter expression)
+
+    NOT pulp_type="core.rbac"
+    # complexity: 2 = 1 (NOT) + 1 (filter expression)
+
+    pulp_type__in=core.rbac,core.content_redirect
```
ggainey, line 36:
```suggestion
    pulp_type__in="core.rbac","core.content_redirect"
```

docs/workflows/complex-filtering.rst
```diff
@@ -0,0 +1,57 @@
...
+    # complexity: 1 = 1 (filter expression)
+
+    pulp_type="core.rbac" OR pulp_type="core.content_redirect"
+    pulp_type="core.rbac" AND name__contains=GGGG
```
ggainey, line 40:
```suggestion
    pulp_type="core.rbac" AND name__contains="GGGG"
```

docs/workflows/complex-filtering.rst
```diff
@@ -0,0 +1,57 @@
...
+    pulp_type="core.rbac" AND name__iexact={prefix}-gGgG
```
ggainey, line 41:
```suggestion
    pulp_type="core.rbac" AND name__iexact="{prefix}-gGgG"
```

This one's a little unclear to me, what is `{prefix}`?

docs/workflows/complex-filtering.rst
```diff
@@ -0,0 +1,57 @@
...
+    pulp_type="core.rbac" AND name__contains="naïve"
+    # complexity: 3 = 1 (AND/OR) + 2 (filter expression)
+
+    pulp_type="core.rbac" AND name__icontains=gg AND NOT name__contains=HH
```
ggainey, line 45:
```suggestion
    pulp_type="core.rbac" AND name__icontains="gg" AND NOT name__contains="HH"
```

docs/workflows/complex-filtering.rst
```diff
@@ -0,0 +1,57 @@
...
+    # complexity: 5 = 1 (AND/OR) + 1 (NOT) + 3 (filter expression)
+
+    NOT (pulp_type="core.rbac" AND name__icontains=gGgG)
```
ggainey, line 48:
```suggestion
    NOT (pulp_type="core.rbac" AND name__icontains="gGgG")
```

docs/workflows/complex-filtering.rst
```diff
@@ -0,0 +1,57 @@
...
+    pulp_type="core.rbac" AND NOT name__contains="naïve"
+    # complexity: 4 = 1 (AND/OR) + 1 (NOT) + 2 (filter expression)
+
+    pulp_type="core.rbac" AND(   name__icontains=gh OR name__contains="naïve")
+    # complexity: 5 = 2 (AND/OR) + 3 (filter expression)
+
+    pulp_type="core.rbac" OR name__icontains=gh OR name__contains="naïve"
```
ggainey, line 55:
```suggestion
    pulp_type="core.rbac" OR name__icontains="gh" OR name__contains="naïve"
```

ggainey (changes requested):
Documentation nitpicks only, consistency to avoid some amount of user-confusion.

docs/workflows/complex-filtering.rst
mdellweg: That one is not going to work.

docs/workflows/complex-filtering.rst
mdellweg: We demonstrate that whitespace is not changing the query.

docs/workflows/complex-filtering.rst
mdellweg: This one is actually bad.

docs/workflows/complex-filtering.rst
ggainey: :+1:

docs/workflows/complex-filtering.rst
ggainey: Heh, ok then :)

docs/workflows/complex-filtering.rst
ggainey: Ah, understood. I still think we may get questions on "why is the documentation inconsistent", but that's a good answer. (And if anyone goes digging, they'll find this discussion and be Enlightened...)

ggainey (approved):

PR #4274, opened 2023-08-14
Optimize ACS stage

dralley:
ACS uses a lot of memory, apparently: https://bugzilla.redhat.com/show_bug.cgi?id=2219885

### DIFF
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -5,7 +5,7 @@ import logging
 
 from aiofiles import os as aos
 from asgiref.sync import sync_to_async
-from django.db.models import Prefetch, prefetch_related_objects, Q
+from django.db.models import Prefetch, prefetch_related_objects
 
 from pulpcore.plugin.exceptions import UnsupportedDigestValidationError
 from pulpcore.plugin.models import (
@@ -485,21 +485,18 @@ class ACSArtifactHandler(Stage):
                                     getattr(d_artifact.artifact, cks_type)
                                 )
 
-                batch_query = Q()
-                for checksum_type in batch_checksums.keys():
-                    batch_query.add(
-                        Q(**{f"{checksum_type}__in": batch_checksums[checksum_type]}), Q.OR
-                    )
-
-                existing_ras = (
-                    RemoteArtifact.objects.acs().filter(batch_query).select_related("remote")
-                )
                 existing_ras_dict = dict()
-                async for ra in existing_ras.aiterator():
-                    for c_type in Artifact.COMMON_DIGEST_FIELDS:
-                        checksum = getattr(ra, c_type)  # todo: this used to be async-protected
+                for checksum_type in batch_checksums.keys():
+                    existing_ras = (
+                        RemoteArtifact.objects.acs()
+                        .filter(**{f"{checksum_type}__in": batch_checksums[checksum_type]})
+                        .only("url", checksum_type, "remote")
+                        .select_related("remote")
+                    )
+                    async for ra in existing_ras.aiterator():
+                        checksum = getattr(ra, checksum_type)
                         # pick the first occurence of RA from ACS
-                        if checksum and checksum not in existing_ras_dict:
+                        if checksum not in existing_ras_dict:
                             existing_ras_dict[checksum] = {
                                 "remote": ra.remote,
                                 "url": ra.url,
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -9,7 +9,6 @@ from django.db.models import Prefetch, prefetch_related_objects
 
 from pulpcore.plugin.exceptions import UnsupportedDigestValidationError
 from pulpcore.plugin.models import (
-    AlternateContentSource,
     Artifact,
     ContentArtifact,
     ProgressReport,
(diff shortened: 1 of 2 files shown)


pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -479,24 +479,18 @@ async def run(self):
...
-                batch_query = Q()
+                existing_ras_dict = dict()
                 for checksum_type in batch_checksums.keys():
-                    batch_query.add(
-                        Q(**{f"{checksum_type}__in": batch_checksums[checksum_type]}), Q.OR
+                    existing_ras = (
+                        RemoteArtifact.objects.acs()
+                        .filter(**{f"{checksum_type}__in": batch_checksums[checksum_type]})
+                        .only("url", checksum_type)
+                        .select_related("remote")
```
dralley, line 488: TODO: 

* figure out how to get rid of this select if possible
* sanity check overall logic

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -465,9 +465,9 @@ class ACSArtifactHandler(Stage):
     """
 
     async def run(self):
+        acs_query = AlternateContentSource.objects.filter(pulp_domain=self.domain)
+        acs_exists = await acs_query.aexists()
         async for batch in self.batches():
-            acs_query = AlternateContentSource.objects.filter(pulp_domain=self.domain)
-            acs_exists = await acs_query.aexists()
```
mdellweg, line 470: +1
Idea: Is it faster to not batch at all without acs? Or should we move this check out to the point, where we can decide wheter the pipeline should include this stage?

dralley: > Or should we move this check out to the point, where we can decide wheter the pipeline should include this stage?

Is commit temp3 what you're looking for?

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -466,52 +465,41 @@ class ACSArtifactHandler(Stage):
...
+            for d_content in batch:
+                for d_artifact in d_content.d_artifacts:
+                    for checksum_type in Artifact.COMMON_DIGEST_FIELDS:
+                        if getattr(d_artifact.artifact, checksum_type):
+                            checksum = getattr(d_artifact.artifact, checksum_type)
+                            if checksum in existing_ras_dict:
+                                d_artifact.urls = [
+                                    existing_ras_dict[checksum]["url"]
+                                ] + d_artifact.urls
+                                d_artifact.remote = existing_ras_dict[checksum]["remote"]
```
gerrod3, line 502: After reading through the code I am mega-confused on how ACS even works and reading the workflow docs did not clear it up for me. https://docs.pulpproject.org/pulpcore/workflows/alternate-content-sources.html 

I think there is a big problem with just overwritting the `d_artifact.remote` field with the one from an existing RA that could be from a different remote! It means all the current urls on the `d_artifact` could be rendered useless since they won't have the correct remote to build the downloader from. I have a feeling this feature is not working entirely how we intend it to.

dralley: Yeah this isn't my strong suit either.  If I understand correctly, the basic idea is to preferentially use a different remote instead of the original one for downloads.  I believe what *might* make it OK, is that when it comes to saving remote artifacts, it's ultimately going to end up saving new remote artifacts rather than modifying existing ones.  But in terms of the download process, it will be the new ones that used?

I think!  It would be worthwhile to verify those assumptions and document the actual details while I'm working on this code anyway.

pulpcore/plugin/stages/artifact_stages.py
mdellweg: Exactly.

Also not grocking the concept of ACS properly... Are they neither tied to a repository, nor a remote? Is that extremely generic `exists` query the best we can do?

mdellweg: Looking at this line:
https://github.com/pulp/pulpcore/blob/20afd1e7dd297e4b15d515176a3d1de8c44f2749/pulpcore/app/models/acs.py#L31

Why do we combine `null=True` with `on_delete=PROTECTED`? If we could say an ACS is always tied to a remote, i would understand a lot more.

dralley: I'm not sure, I think Pavel wrote all of this.  @dkliban do you know?

dralley: For that matter, I wonder why it (according to the docs) only has an effect with on_demand remotes?  I would think that you'd want the ACS to be available during immediate syncs too?

https://github.com/pulp/pulpcore/blob/3cb7ef4369fd612d17e2be7863845f6cc74c141f/docs/workflows/alternate-content-sources.rst#overview

dkliban: I don't think the download_policy should matter.

pulpcore/plugin/stages/artifact_stages.py
ggainey: There's a lot of discussion in the original plan.io issue : https://pulp.plan.io/issues/7832

pulpcore/plugin/stages/artifact_stages.py
ipanova: ACS is not tight to any repo or remote, yet it should have a remote that it's going to use during the refresh action.
The idea was to create that remote always with on-demand policy so the refresh endpoint takes as little time as possible by creating only RA. Those RA could be used later on during repo syncs.
Each ACS is assumed to be global(per domain) so, every rpm sync will check with the rpm typed ACS  content and prefer it's binary data over the original sync remote.

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -466,52 +465,41 @@ class ACSArtifactHandler(Stage):
...
-                                        existing_ras_dict[checksum]["url"]
-                                    ] + d_artifact.urls
-                                    d_artifact.remote = existing_ras_dict[checksum]["remote"]
+            for d_content in batch:
+                for d_artifact in d_content.d_artifacts:
+                    for checksum_type in Artifact.COMMON_DIGEST_FIELDS:
+                        if getattr(d_artifact.artifact, checksum_type):
+                            checksum = getattr(d_artifact.artifact, checksum_type)
+                            if checksum in existing_ras_dict:
+                                d_artifact.urls = [
```
ipanova, line 499: Binging back for transparency some of the matrix conversation @dralley  and I had.
I see few issue with the current design:
1)The original intention for this line was the declarative artifact to accept list of URLs which downloader will try in a loop to go trough. Those URLs would come from a mirrorlist  and the remote would whether not need credentials at all or these credentials would be same for every URLs in that list. Not sure how worthy it is for the ACS stage to have a list of URLs - this would be a mix of them, one coming from original remote, most likely needing credentials and another from ACS( with or without credentials or different credentials?). The remote which is going to be used is the one from ACS.
2) Another problem is that in case the ArtifactDownload stage fails( namely download of binary data from the substituted ACS source)  there is no fallback to the original remote downloader + url.
3) We should first iterate through all available ACSes to download from and then fallback to the original remote but instead we provide just one ( this is currently pipeline limitation, where declarative artifact can  reference just one remote.

Issue #3952, opened 2023-06-23
Create a base `SigningService` class


mayaCostantini:
**Is your feature request related to a problem? Please describe.**
Add a layer of abstraction for signing services by making the existing models inherit from a base signing service class.
The current [`SigningService`](https://github.com/pulp/pulpcore/blob/d1de78b4f314d307be2ce69127d628c6ab953370/pulpcore/app/models/content.py#L757) class is pretty specific to GPG and it is not possible to inherit from it to create other types of signing services such as for Sigstore (see the proposed implementation [here](https://github.com/pulp/pulp_ansible/pull/1428), where having a field for a signing script and public key/fingerprint does not make sense).

**Describe the solution you'd like**
- Create an abstract `BaseSigningService` class and make existing signing service models (`SigningService`, etc) inherit from it. This class can be as minimal as possible, for example with a single `name` field and an abstract asynchronous sign method.
- Rename `SigningService` to `GPGSigningService` or something equivalent for clarity

**Describe alternatives you've considered**
Leave the implementation as it is and make new types of signing services inherit from `BaseModel` directly.

**Additional context**
@rochacbruno @newswangerd I am unsure if the `BaseSigningService` class should be abstract or not here. Obviously it is not meant to be implemented directly so it seems more logical to make it an abstract class, but in this case it will not be present in the database, which might be an issue considering our last discussion (would we be able to group all types of signing services in the same place on sign endpoints?)

quba42:
FYI: If you do rename `SigningService` to `GPGSigningService`, then that would make this a plugin API breaking change.
For example I am inheriting from it in pulp_deb: https://github.com/pulp/pulp_deb/blob/a4df567b0f101c393ac75c3f11cd7411f672db44/pulp_deb/app/models/signing_service.py

mayaCostantini:
Thanks for the heads up @quba42, in this case we could leave the name as it is :+1:

PR #4279, opened 2023-08-15
Create a base class for Signing Services

mayaCostantini:
Related to https://github.com/pulp/pulpcore/issues/3952

Implements an abstract base class for Signing Services to enable creating new types without GPG-specific attributes.

### DIFF
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -754,13 +754,26 @@ class RemoteArtifact(BaseModel, QueryMixin):
         unique_together = ("content_artifact", "remote", "pulp_domain")
 
 
-class SigningService(BaseModel):
+class BaseSigningService(BaseModel):
     """
-    A model used for producing signatures.
+    A base model for producing signatures.
 
     Fields:
         name (models.TextField):
             A unique name describing a script (or executable) used for signing.
+    """
+
+    name = models.TextField(db_index=True, unique=True)
+
+    class Meta:
+        abstract = True
+
+
+class SigningService(BaseSigningService):
+    """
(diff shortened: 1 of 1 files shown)


mayaCostantini:
cc @rochacbruno @newswangerd

lubosmj:
We miss the migration here, don't we? You can create the migrations by running `pulpcore-manager makemigrations`.

Do not forget to add "closes #3952" to the commit message.

mdellweg:
I think the fact that we don't see the migration is that this base class is marked as abstract.
We have had our fair share of trouble with the abstract BaseDistribution model for it to be inherited by plugins. Let's not repeat this.

mayaCostantini:
Thanks @lubosmj I modified the commit message. Also I tried running `makemigrations` but no changes were detected. @mdellweg would you mind sharing more about the issue plugins are having with abstract classes?

mdellweg:
Sorry, for the late answer. But an abstract base class in the plugin api will produce incompatibility issues whenever it changes in the slightest way, because that will require plugins to have a migration for that change in the new version and at the same time require not to have it for older versions of pulpcore.

mayaCostantini:
I see, in this case is there any alternative you would recommend?

mdellweg:
I believe we would make it a "proper" base class with a real db table. And that would involve writing a migration to extend the existing records to the new subclass.

mayaCostantini:
Thanks @mdellweg, I tried removing the `class Meta` section to turn the class into a non-abstract one, but no changes are found when trying to run `pulpcore-manager makemigrations`. Do you have any idea why?

PR #4295, opened 2023-08-17
Made orphan clean up more tolerant when other tasks are running in pa…

### LINKED ISSUE #4209: Remove orphan tasks failing on Katello 4.9
**Version**
katello 4.9

**Describe the bug**
remove orphan tasks in foreman are failing with the following type of message:
```
("Cannot delete some instances of model 'Content' because they are referenced through protected foreign keys: 'Package.content_ptr'.", {<RepositoryContent: pk=ff32c510-ad1c-4288-918a-442fba668125>})
```

**To Reproduce**
Steps to reproduce the behavior: running any remove orphan task will create this error

**Expected behavior**
I would expect the task to remove orphans successfully

**Additional context**

journalctl -u pulpcore-worker@*
```
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]: pulp [b3a7bad0-751c-4973-8432-9a7eb92d6452]: pulpcore.tasking.pulpcore_worker:INFO:   File "/usr/lib/python3.9/site-packages/pulpcore/tasking/pulpcore_worker.py", line 44>
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:     result = func(*args, **kwargs)
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:   File "/usr/lib/python3.9/site-packages/pulp_rpm/app/tasks/synchronizing.py", line 486, in synchronize
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:     remote_url = fetch_remote_url(remote, url)
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:   File "/usr/lib/python3.9/site-packages/pulp_rpm/app/tasks/synchronizing.py", line 305, in fetch_remote_url
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:     remote_url = fetch_mirror(remote)
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:   File "/usr/lib/python3.9/site-packages/pulp_rpm/app/tasks/synchronizing.py", line 254, in fetch_mirror
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:     result = downloader.fetch()
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:   File "/usr/lib/python3.9/site-packages/pulpcore/download/base.py", line 186, in fetch
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:     return done.pop().result()
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:   File "/usr/lib/python3.9/site-packages/pulpcore/download/http.py", line 273, in run
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:     return await download_wrapper()
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:   File "/usr/lib/python3.9/site-packages/backoff/_async.py", line 151, in retry
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:     ret = await target(*args, **kwargs)
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:   File "/usr/lib/python3.9/site-packages/pulpcore/download/http.py", line 258, in download_wrapper
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:     return await self._run(extra_data=extra_data)
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:   File "/usr/lib/python3.9/site-packages/pulp_rpm/app/downloaders.py", line 117, in _run
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:     self.raise_for_status(response)
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:   File "/usr/lib/python3.9/site-packages/pulp_rpm/app/downloaders.py", line 102, in raise_for_status
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:     response.raise_for_status()
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:   File "/usr/lib64/python3.9/site-packages/aiohttp/client_reqrep.py", line 1005, in raise_for_status
Jul 19 00:00:24 foreman.example.com pulpcore-worker-2[152361]:     raise ClientResponseError(
```

PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager showmigrations rpm
```
rpm
 [X] 0001_initial
 [X] 0002_updaterecord_reboot_suggested
 [X] 0003_DATA_incorrect_json
 [X] 0004_add_metadata_signing_service_fk
 [X] 0005_optimize_sync
 [X] 0006_opensuse_support
 [X] 0007_checksum_types
 [X] 0008_advisory_pkg_sumtype_as_int
 [X] 0009_revision_null
 [X] 0010_revision_null_redo
 [X] 0011_rpmremote_sles_auth_token
 [X] 0012_remove_pkg_group_env_cat_related_pkgs
 [X] 0013_RAW_rpm_evr_extension
 [X] 0014_rpmrepository_package_retention_policy
 [X] 0015_repo_metadata
 [X] 0016_dist_tree_nofk
 [X] 0017_merge_advisory_collections
 [X] 0018_updatecollection__update_record
 [X] 0019_migrate_updatecollection_data
 [X] 0020_remove_updatecollection_m2m
 [X] 0021_rename_updatecollection_update_record
 [X] 0022_add_collections_related_name
 [X] 0023_increase_distribution_release_short
 [X] 0024_change_subrepo_relation_properties
 [X] 0025_remove_orphaned_subrepos
 [X] 0026_add_gpgcheck_options
 [X] 0027_checksum_null
 [X] 0028_rpmrepository_last_sync_repomd_cheksum
 [X] 0029_rpmpublication_sqlite_metadata
 [X] 0030_DATA_fix_updaterecord
 [X] 0031_modulemd_static_context
 [X] 0032_ulnremote
 [X] 0033_new_distribution_model
 [X] 0034_auto_publish
 [X] 0035_fix_auto_publish
 [X] 0036_checksum_type
 [X] 0037_DATA_remove_rpmrepository_sub_repo
 [X] 0037_update_json_field
 [X] 0038_fix_sync_optimization
 [X] 0039_disttree_digest
 [X] 0040_rpmalternatecontentsource
 [X] 0041_modulemdobsolete
 [X] 0042_alter_repometadatafile_data_type
 [X] 0043_textfield_conversion
 [X] 0044_noartifact_modules
 [X] 0045_modulemd_fields
 [X] 0046_rbac_perms
 [X] 0047_modulemd_datefield
 [X] 0048_artifacts_dependencies_fix
 [X] 0049_profiles_fix
```

Task Output:
```
{"pulp_tasks"=>
  [{"pulp_href"=>"/pulp/api/v3/tasks/2996831a-d553-4195-982a-12938f7ee779/",
    "pulp_created"=>"2023-07-31T13:10:02.449+00:00",
    "state"=>"failed",
    "name"=>"pulpcore.app.tasks.orphan.orphan_cleanup",
    "logging_cid"=>"5b70908fc43f4d14942561f8319dcc49",
    "started_at"=>"2023-07-31T13:10:02.642+00:00",
    "finished_at"=>"2023-07-31T13:10:07.675+00:00",
    "error"=>
     {"traceback"=>
       "  File \"/usr/lib/python3.9/site-packages/pulpcore/tasking/pulpcore_worker.py\", line 444, in _perform_task\n" +
       "    result = func(*args, **kwargs)\n" +
       "  File \"/usr/lib/python3.9/site-packages/pulpcore/app/tasks/orphan.py\", line 66, in orphan_cleanup\n" +
       "    c.delete()\n" +
       "  File \"/usr/lib/python3.9/site-packages/django/db/models/query.py\", line 745, in delete\n" +
       "    collector.collect(del_query)\n" +
       "  File \"/usr/lib/python3.9/site-packages/django/db/models/deletion.py\", line 302, in collect\n" +
       "    raise ProtectedError(\n",
      "description"=>
       "(\"Cannot delete some instances of model 'Content' because they are referenced through protected foreign keys: 'Package.content_ptr'.\", {<RepositoryContent: pk=ff32c510-ad1c-4288-918a-442fba668125>})"},
    "worker"=>"/pulp/api/v3/workers/92aeba58-871a-40e8-aac1-042e4a0887c6/",
    "child_tasks"=>[],
    "progress_reports"=>
     [{"message"=>"Clean up orphan Content",
       "code"=>"clean-up.content",
       "state"=>"running",
       "total"=>19089,
       "done"=>2000}],
    "created_resources"=>[],
    "reserved_resources_record"=>["/pulp/api/v3/orphans/cleanup/"]}],
 "task_groups"=>[],
 "poll_attempts"=>{"total"=>21, "failed"=>1}}
```

ipanova:
…rallel.

closes pulp/pulp_deb#865
Orphan clean up can fail when other tasks like sync or content upload might be rinning in paralell.

### DIFF
--- a/pulpcore/app/tasks/orphan.py
+++ b/pulpcore/app/tasks/orphan.py
@@ -1,6 +1,9 @@
 import gc
 
+from logging import getLogger
+
 from django.conf import settings
+from django.db.models.deletion import ProtectedError
 from django.utils import timezone
 
 from pulpcore.app.models import (
@@ -11,6 +14,8 @@ from pulpcore.app.models import (
     PulpTemporaryFile,
     Upload,
 )
+
+log = getLogger(__name__)
 
 
 def queryset_iterator(qs, batchsize=2000, gc_collect=True):
@@ -44,35 +49,67 @@ def orphan_cleanup(content_pks=None, orphan_protection_time=settings.ORPHAN_PROT
         content_pks (list): A list of content pks. If specified, only remove these orphans.
 
     """
+    content = Content.objects.orphaned(orphan_protection_time, content_pks).exclude(
+        pulp_type=PublishedMetadata.get_pulp_type()
+    )
+    skipped_content = 0
     with ProgressReport(
         message="Clean up orphan Content",
-        total=None,
+        total=content.count(),
         code="clean-up.content",
     ) as progress_bar:
-        while True:
-            content = Content.objects.orphaned(orphan_protection_time, content_pks).exclude(
-                pulp_type=PublishedMetadata.get_pulp_type()
-            )
-            content_count = content.count()
-            if not content_count:
-                break
+        # delete the content
+        for bulk_content in queryset_iterator(content):
+            skipped_content_batch = 0
+            count = bulk_content.count()
+            try:
+                bulk_content.delete()
+            except ProtectedError:
+                # some orphan content might have been picked by another task running in parallel
+                # i.e. sync
+                for c in bulk_content:
+                    try:
+                        c.delete()
+                    except ProtectedError as e:
+                        log.debug(e)
+                        skipped_content_batch += 1
+            progress_bar.increase_by(count - skipped_content_batch)
+            skipped_content += skipped_content_batch
 
-            # delete the content
-            for c in queryset_iterator(content):
-                progress_bar.increase_by(c.count())
-                c.delete()
+    if skipped_content:
+        msg = (
+            "{} orphaned content could not be deleted during this run and was skipped. "
+            "Re-run the task and/or consult the logs."
+        )
+        log.info(msg.format(skipped_content))
 
     # delete the artifacts that don't belong to any content
     artifacts = Artifact.objects.orphaned(orphan_protection_time)
 
+    skipped_artifact = 0
     with ProgressReport(
         message="Clean up orphan Artifacts",
         total=artifacts.count(),
         code="clean-up.artifacts",
     ) as progress_bar:
-        for artifact in progress_bar.iter(artifacts.iterator()):
-            # we need to manually call delete() because it cleans up the file on the filesystem
-            artifact.delete()
+        for artifact in artifacts.iterator():
+            try:
+                # we need to manually call delete() because it cleans up the file on the filesystem
+                artifact.delete()
+            except ProtectedError as e:
+                # some orphaned artifact might have been picked by another task running in parallel
+                # i.e. sync
+                log.debug(e)
+                skipped_artifact += 1
+            else:
+                progress_bar.increment()
+
+    if skipped_artifact:
+        msg = (
+            "{} orphaned artifact(s) could not be deleted during this run and were skipped. "
+            "Re-run the task and/or consult the logs."
+        )
+        log.info(msg.format(skipped_artifact))
 
 
 def upload_cleanup():
--- a/pulpcore/app/tasks/purge.py
+++ b/pulpcore/app/tasks/purge.py
@@ -143,7 +143,7 @@ def purge(finished_before, states):
                     # Log the details of the object.
                     error_pb.done += 1
                     pks_failed.append(pk)
-                    log.info(e)
+                    log.debug(e)
 
     # Complete the progress-reports for the specific entities deleted
     for key, pb in details_reports.items():
--- a/CHANGES/4316.bugfix
+++ b/CHANGES/4316.bugfix
@@ -1,0 +1,1 @@
+Made orphan clean up more tolerant when other tasks are running in parallel.
--- a/pulpcore/app/tasks/reclaim_space.py
+++ b/pulpcore/app/tasks/reclaim_space.py
@@ -86,9 +86,9 @@ def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
             # we need to manually call delete() because it cleans up the file on the filesystem
             artifact.delete()
         except ProtectedError as e:
-            # Rarely artifact could be shared between to different content units.
+            # Rarely artifact could be shared between two different content units.
             # Just log and skip the artifact deletion in this case
-            log.info(e)
+            log.debug(e)
         else:
             progress_bar.done += 1
             counter += 1


quba42:
I am going to reply to https://github.com/pulp/pulpcore/issues/4209#issuecomment-1688182643 on here:

> There is another side of the coin to it though, by catching the ProtectedError we'd just log the error, which is not as visible as raising a traceback, so some issues, like this one with Debian will more likely to be covered, ignored, etc.

Does it make sense to only catch these errors if the user has explicitly opted in via some flag on the orphan cleanup run?
Then users would run into the error first, report it, and we could then tell them to re-run orphan cleanup using `ignore_protected_error = True` (or whatever it would be).

Just an idea. If this is too much trouble and we don't want a thousand flags, then feel free to ignore.

dralley:
IMO, not worth it, we would be better off just logging it and having a separate script that can identify these situations and perhaps repair them

pulpcore/app/tasks/orphan.py
```diff
@@ -53,14 +58,20 @@ def orphan_cleanup(content_pks=None, orphan_protection_time=settings.ORPHAN_PROT
...
             for c in queryset_iterator(content):
-                progress_bar.increase_by(c.count())
-                c.delete()
+                count = c.count()
+                try:
+                    c.delete()
+                except ProtectedError as e:
+                    # some orphan content might have been picked by another task running in parallel
+                    # i.e. sync
+                    log.info(e)
```
quba42, line 72: Should this at least be a log.warning?

quba42:
So my opinion is that it does make sense to have this mitigation in some way shape or form. I view it as very disruptive for users if the rest of their orphan cleanup is being held hostage, by a hand full of orphans from one plugin that can't be cleaned up. Even if we fix the current issues in pulp_deb, there is no guarantee there won't be more issues like this in the future. When these bugs are found they may take some time to fix and meanwhile users storage is filling up.

If we are worried that just logging these cases will not generate enough visibility, we could do the following: Set a local flag variable if any `ProtectedError`s are encountered and then throw a custom exception _at the end of the orphan run_. Something like a `SkippedOrphanedContentException` with a message something like: "Some orphaned content could not be deleted during this run and was skipped. Consult your logs for more information on the skipped orphans and consider opening an issue." (feel free to come up with a better formulation)

That way orphan cleanup would clean up everything that can be cleaned up and still throw an error to create visibility for the things that can't be cleaned up.

pulpcore/app/tasks/orphan.py
```diff
@@ -71,8 +82,13 @@ def orphan_cleanup(content_pks=None, orphan_protection_time=settings.ORPHAN_PROT
         code="clean-up.artifacts",
     ) as progress_bar:
         for artifact in progress_bar.iter(artifacts.iterator()):
-            # we need to manually call delete() because it cleans up the file on the filesystem
-            artifact.delete()
+            try:
+                # we need to manually call delete() because it cleans up the file on the filesystem
+                artifact.delete()
+            except ProtectedError as e:
+                # Rarely artifact could be shared between to different content units.
```
mdellweg, line 89: I would not even call this rare. The empty file is probably shared quite a lot. Also any package whatsoever may make sense to be in a file repository, and in the end, PublishedContent is some sort of Content too.
```suggestion
                # Rarely artifact could be shared between different content units.
```

pulpcore/app/tasks/orphan.py
```diff
@@ -53,14 +58,20 @@ def orphan_cleanup(content_pks=None, orphan_protection_time=settings.ORPHAN_PROT
             content = Content.objects.orphaned(orphan_protection_time, content_pks).exclude(
```
ipanova, line 58: I need to get rid of while loop

pulpcore/app/tasks/orphan.py
ipanova: I took this from reclaim space, but this is wrong so I am going to adjust the message.

ipanova:
@quba42 I am going to find some middle ground to this.

pulpcore/app/tasks/purge.py
```diff
@@ -143,7 +143,7 @@ def purge(finished_before, states):
                     # Log the details of the object.
                     error_pb.done += 1
                     pks_failed.append(pk)
-                    log.info(e)
+                    log.debug(e)
```
ipanova, line 146: not sure it's worth to log.info the protectederror hence moved to debug

dralley (approved):

pulpcore/app/tasks/orphan.py
```diff
@@ -44,35 +49,63 @@ def orphan_cleanup(content_pks=None, orphan_protection_time=settings.ORPHAN_PROT
...
+        # delete the content
+        for c in queryset_iterator(content):
+            count = c.count()
+            try:
                 c.delete()
+            except ProtectedError as e:
+                # some orphan content might have been picked by another task running in parallel
+                # i.e. sync
+                log.debug(e)
+                skipped_content += count
```
mdellweg, line 70: So at this point, we skipped a batch of content. Should we not fall back to deleting them individually now? It's probably only a single one failing.

ipanova: I was thinking about that. We could, just it won't be bulk delete but one by one.
Do you want me to adopt this?

mdellweg: I lean towards yes. It's the same approach as we take with bulk_create.

pulpcore/app/tasks/orphan.py
```diff
@@ -44,35 +49,68 @@ def orphan_cleanup(content_pks=None, orphan_protection_time=settings.ORPHAN_PROT
...
+                skipped_artifact += 1
+
+    if skipped_artifact:
+        msg = (
+            "{} orphaned artifact(s) could not be deleted during this run and were skipped. "
+            "Re-run the task and/or consult the logs."
+        )
+        log.info(msg.format(skipped_artifact))
+
+        progress_bar.done -= skipped_artifact
```
mdellweg, line 112: I think I don't like this. Also it is inconsistent with how we report the deleted content.
Do we want to count deleted and skipped units separately?
Or have a total count and a skip count?

ipanova: I did not want to introduce a separate report for skipped count. I could, but I don't like that. We would have in total 4 reports( 2 for success , 2 for skipped). This general pattern of adding reports( because of general report limitation)  for any other action except for `successful done` is not very useful. Something failed? add a report, something skipped? add report. As a result you have X separated reports instead of having 1 report that can count done, skipped,failed,etc.

To me, combined with the logs this is clear enough when i look at the report:
```
 {
      "message": "Clean up orphan Content",
      "code": "clean-up.content",
      "state": "completed",
      "total": 3,
      "done": 2,
      "suffix": null
    },
```
As i said, I can add additional reports, but for the mentioned reasons that would not be my preference. In addition, if on the user side there was some kind of script that was parsing this report, he'd need to adjust it to parse the newly added ones too. 
Le me know.

mdellweg: Adding a new counter should not break existing parsers (they should be built to ignore them.)
But I hear you. So your suggestion is: We set the total to the total number of processed units and done means "sucessfully done". And then the user needs to consult the state to see it is actually done. And the difference will be the failed units.
I believe we could do that, but i'd really not add the failed units to done to later subtract them again.

ipanova: > Adding a new counter should not break existing parsers (they should be built to ignore them.) 

I agree but that means if ignored they would never know that 'something else happened'.
>But I hear you. So your suggestion is: We set the total to the total number of processed units and done means >"sucessfully done". And then the user needs to consult the state to see it is actually done. And the difference will be >the failed units. 

Yeah, that was  the idea behind. Unless we excplicity for some reason want the skipped/failed units to make visible in separate report, as, for example, there is such instance in rpm, plugin https://github.com/pulp/pulp_rpm/blob/main/pulp_rpm/app/tasks/synchronizing.py#L1278
> I believe we could do that, but i'd really not add the failed units to done to later subtract them again.

Yeah, I can keep separate variable counters for that to avoid the subtraction later on, just wanted to avoid more variable/counter to keep track off :unicorn:

pulpcore/app/tasks/orphan.py
```diff
@@ -44,35 +49,73 @@ def orphan_cleanup(content_pks=None, orphan_protection_time=settings.ORPHAN_PROT
...
+                # some orphaned artifact might have been picked by another task running in parallel
+                # i.e. sync
+                log.debug(e)
+                skipped_artifact += 1
+            else:
+                counter += 1
+
+            if counter >= interval:
+                progress_bar.increase_by(counter)
+                counter = 0
```
mdellweg, line 111: We can just call progress_bar.increment here.
When used as a contextmanager (like we do here), the progress bar will defer writing to the database until there's a certain amount to update.

ipanova: Sure, i just don't want to call save on each unit iterated, increase_by calls in save evenually anyway, and it will be done every 100 units instead.

mdellweg: I think we should rely on what the context manager provides here. But not going to make this a required change.

mdellweg (approved):

mdellweg (approved):

dralley (approved):

PR #4296, opened 2023-08-17
Fixes #4294

### LINKED ISSUE #4294: Import-check may return error saying "director does not exist" when it actually just lack permissions
Endpoint /pulp/api/v3/importers/core/pulp/import-check/ returns error saying "Directory does not exist" when pulp user lack permissions to read said directory.

**To Reproduce**

Try importing content from a directory where pulp user doesn't have read access.

**Expected behavior**
Error returned should indicate the permission error.

**Additional context**

Pulp is using os.path.exists() method to verify if the directory exists: https://github.com/pulp/pulpcore/blob/main/pulpcore/app/views/importer.py#L44-L45

However, the method can return false if permission is not granted to access the directory even if the directory exists

~~~
os.path.exists(path)
Return True if path refers to an existing path or an open file descriptor. Returns False for broken symbolic links. On some platforms, this function may return False if permission is not granted to execute os.stat() on the requested file, even if the path physically exists.
~~~

os.path method documentation -> https://docs.python.org/3/library/os.path.html

jpasqualetto:
Use os.stat to ensure directory exists and pulp has read-access

return any errors received from os.stat to the user


pulpcore/app/views/importer.py
```diff
@@ -40,11 +40,14 @@ def _validate_file(in_param, data):
...
+    # return any errors received from os.stat to the user
+
     owning_dir = os.path.dirname(real_file)
-    if not os.path.exists(owning_dir):
-        return False, [_("directory {} does not exist").format(owning_dir)]
-    if not os.access(owning_dir, os.R_OK):
-        return False, [_("directory {} does not allow read-access").format(owning_dir)]
+    try:
+        os.stat(owning_dir)
+    except Exception as e:
```
ipanova, line 49: I see some other os.access occurrences down in the code, does it make sense to update those too?

jpasqualetto: The problem is not really with os.access, but os.path.exists. I replaced both because os.stat will test both use-cases and return a meaningful error in case of the directory not existing or missing permissions.

The other os.access occurrences look good to me.

dralley: Can we replace `except Exception` with something more specific like `except OSError`

dralley (approved):

pulpcore/app/views/importer.py
dralley: @jpasqualetto Previously we printed the directory in question in the error message, now we're not.  Do you want to adjust the error message to keep that info?

jpasqualetto: Hello @dralley 

The directory name will be part of the error message returned by os.stat.  For example:

`[Errno 13] Permission denied: '/var/lib/pulp/imports/TestCV/1.0/2023-05-11T10-51-29-04-00'
`
or 

`[Errno 2] No such file or directory: '/tmp/DirectoryThatDoesNotExist'
`
So, I don't think we need to include it on the error message.

ggainey (changes requested):
This needs a changelog entry in CHANGES/4294.bugfix , to pass lint : https://docs.pulpproject.org/pulpcore/contributing/git.html#changelog-update

CHANGES/4294.bugfix
```diff
@@ -0,0 +1 @@
+Fixed bug where incorrect error message presented in relation to content-import
```
ggainey, line 1: :+1:

ggainey (approved):

Issue #3358, opened 2022-10-25
As a domain user, I can change my storage system credentials and Pulp will migrate my data


bmbouter:
## Motivation

Assume the user is already a domain user. The user needs to migrate data from one storage backend to another. For example:

* I'm using local filesystem storage and I want to try the cloud cuz it's awesome.
* I'm using the AWS S3 and I want to try Azure Storage because our company decided to consolidate to Azure
* I'm using Azure and I want to switch to local filesystem because I realized [I'm overpaying](https://world.hey.com/dhh/why-we-re-leaving-the-cloud-654b47e0)

## Expectations

1. Pulp should copy the Artifacts from one storage backend to the other when a user changes their domain storage settings.
2. Pulp needs to remain online while this occurs.

## Technical planning

* It needs to happen asynchronously in the background as a task.

ipanova:
Thanks for sharing that article, interesting read.

PR #4298, opened 2023-08-17
Add migrate endpoint to move artifacts to another storage backend

gerrod3:
fixes: #3358


pulpcore/app/response.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, task, request):
         """
         kwargs = {"pk": task.pk}
         if settings.DOMAIN_ENABLED:
-            kwargs["pulp_domain"] = request.pulp_domain.name
+            kwargs["pulp_domain"] = task.pulp_domain.name
```
mdellweg, line 28: This is not a semantic change, right? Should we maybe always try to use `get_domain` instead?

gerrod3: `request.domain == get_domain()`. This change is for when setting the task in a separate domain from the calling domain. In this case when calling migrate it is called (can be called) from the default domain since domains are a part of the default domain.

mdellweg: But should you not call migrate from within the affected domain?
The fact that domains are in the default domain is a bit confusing i guess.

gerrod3: Yeah this is a question I haven't figured out yet and that is where to put the new endpoint. Currently it is on the domains endpoint, specifically on an instance of a domain (`/pulp/default/api/v3/domains/domain_href/migrate/`), but it could equally just be a new standalone endpoint that figures out what to do based on the domain it ran in. (`pulp/foo-domain/api/v3/migrate/`). Not sure what approach is the best, any suggestions?

mdellweg: Self referential stuff is always weird...
would `POST /pulp/domain/api/v3/domain/migrate/` work? Where `/pulp/domain/api/v3/domain/` is a self inspecting domain detail viewset?

gerrod3: Yes, that would work as well. Would you like for me to try it out?

pulpcore/app/viewsets/domain.py
```diff
@@ -118,3 +126,39 @@ def destroy(self, request, pk, **kwargs):
...
+
+    @extend_schema(
+        summary="Migrate storage backend",
+        request=DomainBackendMigratorSerializer,
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    @action(detail=True, methods=["patch"])
+    def migrate(self, request, **kwargs):
+        """
+        Migrate the domain's storage backend to a new one.
```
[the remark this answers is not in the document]
gerrod3: Only on the default domain can your scenario happen, which is why I explicitly instruct users to update their config file to the new backend and restart Pulp after the migration (hopefully users of domains won't be migrating the default domain very often). The migration task locks the domain so no other task can run, and for all other non-default domains the task will also update that domain's storage settings to the new backend. Therefore any queued up task will run with the new backend after the migration.

Note, the default domain is special since it loads its values for storage from the config file which with `dynaconf` means they could be stored anywhere, so it was impractical to update these values automatically.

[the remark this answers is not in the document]
bmbouter: I didn't realize they update the configs after migration. My concern is a not an issue. Regarding the default domain I don't think we can do any better either. Thanks for explaining this.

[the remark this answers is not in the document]
mdellweg: Should we let the default domain configure it's storage backend in the database too then?
Or what about this: We disallow to use this endpoint on the default domain, and make this migration for the default domain a management command that checks if services are shut down? And on startup, all services check that the storage settings are matching the db?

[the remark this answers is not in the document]
ipanova: Since default domain is special and essentially requires to undergo a manual switch-over, my recommendation would be to document the necessary steps (two?) and add a warning note that will make it clear that until the necessary steps are not performed all content, including new one will be saved/served from the old backend.

pulpcore/app/tasks/migrate.py
```diff
@@ -0,0 +1,60 @@
...
+                "PLEASE UPDATE CONFIG FILE WITH THE NEW STORAGE SETTINGS BEFORE RESTARTING PULP!"
+            )
+            _logger.warning(msg)
+            ProgressReport(message=msg, code="URGENT", state=TASK_STATES.SKIPPED).save()
+            pb.state = TASK_STATES.SKIPPED
+        else:
+            domain.storage_class = data["storage_class"]
+            domain.storage_settings = data["storage_settings"]
+            domain.save(update_fields=["storage_class", "storage_settings"])
+            pb.increment()
```
jlsherrill, line 60: How errors are handled is one of my biggest concerns here.  If i'm reading this right:

* If an error occurs right away (i.e. new storage settings are wrong), the error gets thrown and nothing is updated
*  if an error occurs half way through the migration (or lets say it gets cancelled.)  Some of the artifacts would be migrated to the new storage but some would not, and the domain would not be updated to the storage correct?  And then a user would need to migrate again to finish migrating all the rest of the items, or else specify a new 3rd configuration, where by everything would be migrated to the third storage.

Does that sound Accurate?

pulpcore/app/tasks/migrate.py
```diff
@@ -0,0 +1,60 @@
...
+                    new_storage.save(filename, old_storage.open(filename))
+            # Handle new artifacts saved by the content app
+            artifacts = Artifact.objects.filter(pulp_domain=domain, pulp_created__gte=date)
+            if count := artifacts.count():
+                pb.total += count
+                pb.save()
+                date = now()
+                continue
+            break
+
```
jlsherrill, line 44: what happens to the files stored in the old storage?  Are they just left behind?

pulpcore/app/tasks/migrate.py
gerrod3: Yep, that's correct for both scenarios. I'll add a check for if the artifact is missing on the current backend, and tell the user to run the repair command or delete the artifact if so.

pulpcore/app/tasks/migrate.py
gerrod3: Correct, I don't attempt to delete them. The user can decide what to do with their old backend.

mdellweg: Do we document that?
Also I think we eventually need to provide tools for this. I'd be ok to leave that for later...

pulpcore/app/viewsets/domain.py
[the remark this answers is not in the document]
mdellweg: The issue i have is that here is a window we can never really close. The migration is run by a task, so the services are online. Now while the migration is running there may already be tasks queued up to add more content to the default domain. And before you can reconfigure and restart the services, the two copies of storage are out of sync.

For domains carrying their storage config in the database, i think we have it covered.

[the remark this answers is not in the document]
ipanova: @mdellweg you have a valid point

pulpcore/app/tasks/migrate.py
gerrod3: It's documented in the endpoint's method comment, which appears in the auto-gen docs. I don't think there is a need for a tool to delete files from the storage. We already entrust users to bring their own storage, so I think they can manage their files on it, but I agree if we do decide we need a tool it would be best to add it later.

pulpcore/app/viewsets/domain.py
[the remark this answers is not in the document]
gerrod3: Ok, I've tried to reach a compromise between having the default domain update immediately upon migration, but still allow for its settings to be determined by the config file. Here is the new logic:

- All domains have their storage instance built using their field `storage_settings`. For the default domain this field is empty and thus will use the default values from settings.
- Upon migration the domain's `storage_settings` is updated to the new values. This way all domains, including the default domain, will now properly use the new storage backend after the migration.
-If migrating the default domain, a message is put into the  migration task to remind the user to update their config with the new storage settings before the next DB migration
- Upon `pulpcore-manager migrate`, Pulp in a post-migrate hook will update the default domain's saved fields to match the values (`storage_class`, `redirect_to_object_storage`, `hide_distributions`) to the config file. `storage_settings` will be reset back to empty to ensure that the default storage is again built from settings.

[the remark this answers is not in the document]
mdellweg: Sorry for keeping you busy with this one. But i think we should try to find a bullet proof way.
If i understand your compromise correctly, the admin will run the migration, whereby the default domain will change it's config. Now they miss reading the message and upon running `pulpcore-manager migrate` the system will be borked again. Is this correct?

What if we follow an approach more similar to the allowed checksums setting?
1. You adjust the config.
2. You restart the services (they may warn you about the mismatch).
3. You kick off the migrate task for the default domain without any parameters (they come from the settings).
  3.1 If the settings fail to validate, it will error out in the view.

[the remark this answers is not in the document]
gerrod3: Can you explain more of what you mean by 3 and 3.1? I am not following your logic.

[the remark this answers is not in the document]
mdellweg: > Yes the system will be bricked until they update the config and run `pulpcore-manager migrate`. I've added a check into `settings.py` for mismatched settings and logs a warning to the user.

But by the time this message is shown (and may even be missed by the admin), Pulp is reconfigured to use this old (brobably already outdated) storage, and with the first new artifact there, you end up with two in consistent storages. That is the situation i try to avoid.

Line of thinking: How much less special can we make the default domain? What if the default domain, like any other, just always carries it's storage definition? So when you want to change that, you modify the settings, restart your services and tell pulp to rectify the situation. That will be your migrate task, but it should take it's parameters from the config now, instead of the request body (Or we make it fail when they differ...). They get validated before the task is dispatched (that was 3.1). Not at all times, the data base will be the truth, and the settings are the desired state the database is supposed to match eventually.

pulpcore/app/tasks/migrate.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    data = DomainBackendMigratorSerializer.decrypt_data(data)
+    domain = get_domain()
+    old_storage = domain.get_storage()
+    new_storage = Domain(**data).get_storage()
+
+    artifacts = Artifact.objects.filter(pulp_domain=domain)
+    date = now()
+
+    with ProgressReport(
```
mdellweg, line 29: Now here's a new question coming to my mind. Are there settings that can change without it being a new bucket? Thinking about key-rotation here...
Is there any way we can identify "this is the same bucket, do not move the artifacts"? Or even "with the old settings you cannot reach the bucket anymore"?

bmbouter:
I met with @jlsherrill yesterday and he brought this up as a feature their usage will need this year. I'm hoping we can release it upstream and test it some there before putting it into use by @jlsherrill 's team

@gerrod3 and @goosemania could you give some thought to when this could be continued? I'm not sure what else you have going on so I'm just starting the convo here.

pulpcore/app/tasks/migrate.py
gerrod3: Still need to think about this some more, but I do believe there is a way. I check to see if the artifact is present in the new location, but depending on the number of artifacts this could still be slow and doing unnecessary work.

gerrod3:
Ok, I've updated the PR to make the default domain less special. The storage settings for the default domain will now also be saved in the database. This will be accomplished through a data migration (should be ZDU safe). When migrating the default domain the endpoint will use the values from the config file so that settings will still be the final source of truth for the default domain. This will allow users to change the values in their settings file and then simply run the migrate endpoint to move their artifacts to the new backend, all without restarting Pulp.

I'm still unsure of the best way to inform users if there is a disparity between settings and the default domain. Right now I have a check in settings.py file, but it doesn't feel like the right place.

pulpcore/app/migrations/0119_store_default_domain_settings.py
```diff
@@ -0,0 +1,34 @@
...
+
+from django.db import migrations
+from pulpcore.app.serializers.domain import StorageSettingsSerializer
+from django.conf import settings
+
+
+def default_domain_settings(apps, schema_editor):
+    Domain = apps.get_model("core", "Domain")
+    default_domain = Domain.objects.get(name="default")
+    config_settings = StorageSettingsSerializer.get_default_domain_settings(settings)
```
mdellweg, line 11: This is a bit dangerous. Maybe it's better at this point do keep a frozen copy of the serializer.

pulpcore/app/serializers/domain.py
```diff
@@ -187,10 +190,9 @@ class AmazonS3SettingsSerializer(BaseSettingsClass):
     access_key = serializers.CharField(required=True, write_only=True)
     secret_key = serializers.CharField(allow_null=True, default=None, write_only=True)
     security_token = serializers.CharField(allow_null=True, default=None, write_only=True)
-    # Too dangerous to use shared cred file, ensure is always False
-    session_profile = serializers.HiddenField(default=False)
+    session_profile = serializers.CharField(allow_null=True, default=None)
     file_overwrite = serializers.BooleanField(default=True)
-    object_parameters = serializers.DictField(default={})
+    object_parameters = serializers.DictField(default=dict())
```
mdellweg, line 195: This isn't really a change. Did you intend to change to `default=dict` to make it use a callable?

pulpcore/app/settings.py
```diff
@@ -528,6 +528,33 @@
...
+    # Check if the configuration of the default domain is mismatched with settings
+    try:
+        with connection.cursor() as cursor:
+            cursor.execute(
+                "SELECT storage_class, redirect_to_object_storage, hide_guarded_distributions, "
+                "pulp_id FROM core_domain WHERE name = 'default'"
+            )
+            row = cursor.fetchone()
+            mismatched = []
+            for i, setting in enumerate(
```
mdellweg, line 540: would
``` python
for k, v in zip((DEFAULT_FILE_STORAGE, ...), row):
```
work?

mdellweg (commented):
Not a full review...
Did not look at the tests yet.

pulpcore/app/migrations/0119_store_default_domain_settings.py
gerrod3: Do you mean that the migration is dangerous because it could be ran multiple times?

mdellweg: No, it's dangerous to define the migration with code that may change in future releases.

pulpcore/tests/unit/models/test_remote.py
```diff
@@ -53,6 +53,9 @@ def test_encrypted_proxy_password(fake_fernet):
 
 @pytest.mark.django_db
 def test_rotate_db_key(fake_fernet):
+    # nullify default domain settings so that this test can pass
+    Domain.objects.filter(name="default").update(storage_settings={})
+
```
mdellweg, line 58: So the default domain make the rotate-db-key command fail, because we don't have a really empty db here and we fake the old key too?

pulpcore/app/apps.py
```diff
@@ -315,18 +315,23 @@ def _ensure_default_domain(sender, **kwargs):
...
             or settings.REDIRECT_TO_OBJECT_STORAGE != default.redirect_to_object_storage
             or settings.DEFAULT_FILE_STORAGE != default.storage_class
+            or default.storage_settings != config_settings
         ):
             default.hide_guarded_distributions = settings.HIDE_GUARDED_DISTRIBUTIONS
             default.redirect_to_object_storage = settings.REDIRECT_TO_OBJECT_STORAGE
             default.storage_class = settings.DEFAULT_FILE_STORAGE
+            default.storage_settings = config_settings
             default.save(skip_hooks=True)
+            print(_("Updated default domain to match current Pulp settings"))
```
mdellweg, line 334: So this is a post migrate hook. Can we identify when this is safe and only update accordingly?

I suspect way to many installations running migrate often and that would be a horrible scenario for migrating the default domain. OTOH, using the migrate task would rectify these settings too, right?
How about we only update HIDE_GUARDED_DISTRIBUTIONS and REDIRECT_TO_OBJECT_STORAGE here and instruct the admin to run the migrate task? (Can we do that from a management command to migrate in "maintenance mode"?)
Have we discussed these questions before?

pulpcore/app/migrations/0119_store_default_domain_settings.py
gerrod3: What's the best way to get a frozen copy of the serializer? Copy pasting 100s of lines doesn't seem right, but maybe that's the best we can do?

pulpcore/app/apps.py
gerrod3: Getting back to this after a while, I am wondering if I should keep the default domain special and prevent users calling migrate on it. Instead default domain will have to use a management command after bringing services down. They will lose the benefit of zero-downtime migrations, but the headache around trying to keep track of when and when we should not try to update the default domain on settings change will be resolved. 

If you change your backend settings and Pulp breaks -> shut down pulp & run the migrate command.

What do you think? I'm leaning more towards this solution to avoid supporting unnecessary complexity.

mdellweg: I'm a big fan of less complexity.
The one thing that _is_ different with the default domain is that is is configured in the settings. But we could also stop doing that, no?

pulpcore/app/migrations/0119_store_default_domain_settings.py
mdellweg: So what is it that we want from this function? Isn't it a dictionary based on the storage tech configured?
Can we just put a few literals here?

pulpcore/tests/functional/api/using_plugin/test_migrate.py
```diff
@@ -0,0 +1,111 @@
...
+    random_artifact_factory,
+):
+    """Test migrating from a domain."""
+    domain = domain_factory()
+    artifacts = [random_artifact_factory(pulp_domain=domain.name) for _ in range(3)]
+    old_storage, old_settings = backend_settings_factory()
+
+    body = {
+        "storage_class": "pulpcore.app.models.storage.FileSystem",
+        "storage_settings": {"location": "/var/lib/pulp/media"},
```
mdellweg, line 63: That's the location of our default local file storage, right? But it is still namespaced by domains?

pulpcore/tests/functional/api/using_plugin/test_migrate.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # Check that the progress reports and 'done's are right
+    reports = task.progress_reports
+    assert len(reports) == 2
+    assert {3, 1} == {r.done for r in reports}
+
+    # Check that the domain's storage settings was updated
+    domain2 = pulpcore_bindings.DomainsApi.read(domain.pulp_href)
+    assert domain2.storage_class == "pulpcore.app.models.storage.FileSystem"
+    assert domain2.storage_settings["location"] == "/var/lib/pulp/media"
```
mdellweg, line 75: I don't see a reason for a new variable here.
```suggestion
    domain = pulpcore_bindings.DomainsApi.read(domain.pulp_href)
    assert domain.storage_class == "pulpcore.app.models.storage.FileSystem"
    assert domain.storage_settings["location"] == "/var/lib/pulp/media"
```

pulpcore/tests/functional/api/using_plugin/test_migrate.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # Check that the files were actually moved
+    domain_id = domain.pulp_href.split("/")[-2]
```
mdellweg, line 78: If you want, move this assignment to after the factory call.

pulpcore/tests/functional/api/using_plugin/test_migrate.py
gerrod3: Correct, it'll be namespaced by domains so it won't override the default domain.

mdellweg (approved):

PR #4322, opened 2023-08-24
Handle existing labels in migration 0104

### LINKED ISSUE #4319: Need a way to directly migrate pulp_labels from 3.21 to 3.28
As RHUI developers, we are heavy users of pulp_labels on repository objects.  As the current 4.5 release is on pulpcore 3.21, we would like our next release (4.6) to be based on pulpcore 3.28, and we need to ensure a smooth upgrade path for our customers.  As it is currently laid out, the 3.21 to 3.28 migration path requires an intermediate step (<3.25) where `pulpcore-manager datarepair-labels` command is available.

The ideal solution would be to make Django migrations take care of that automatically, without requiring an intermediate step/release.  Since we live in non-ideal world, a solution where an additional command like `datarepair-labels` is available would also suffice.

We have considered an intermediate release of RHUI 4.6 with pulpcore 3.22 followed by RHUI 4.7 built on 3.28, but it would create a lot of migration issues for our customers, with many of them still running RHUI 4.0.

mdellweg:
fixes #4319


ggainey:
This PR handles the "but what about plugins" problem noted in the issue, with only two caveats:

Case 1: If there were to exist a plugin containing a Model whose model-name ended in (for example) "Repository", that was NOT a detail-model of a Repository master-model, **AND** that plugin allowed Labels for such a model - then, upon running this migration, those Labels would be lost.

Case 2: If there were to exist a plugin containing a Model that **was** a Detail of (for example) Repository, but named something like "PluginRepositoryButWhy", and that plugin allowed Labels, and instances of such a Model had Labels associated with them - then this migration would *fail*, because the Labels would not be found, migrated, and deleted, and the old-Label table would not be able to be dropped.

And the plugins described above would have to have existed and been in use with pulpcore/3.21, only - if they appeared with core/3.22, they'd be using new-Labels and all would be (already) well.

No such plugins/Models exist, to the best of our knowledge.

sskracic:
Just FYI, I have manually patched `migrations/0104_delete_label.py` on 3.28 and was able to successfully migrate label data from an earlier 3.21 installation.

pulpcore/app/migrations/0104_delete_label.py
```diff
@@ -1,6 +1,45 @@
 # Generated by Django 3.2.18 on 2023-05-04 07:06
 
 from django.db import migrations
+from django.db.models.expressions import OuterRef, RawSQL
+
+
+def migrate_remaining_labels(apps, schema_editor):
+    """
+    This data migration handles the "but what about plugins" problem noted in the issue [0], with only two caveats:
```
ggainey, line 9: :+1:

ggainey (approved):

gerrod3 (approved):

ggainey:
3.22/3.23 backport labels were inappropriate

Issue #3994, opened 2023-07-05
RBAC permission for the upstream/replication task


mchithu:
**Version**
Python based Pulp installation.
"core:3.21.0"
"rpm:3.21.0"
"python:3.21.0"
"deb:3.21.0"
"container:3.21.0"
"file:3.21.0"
"certguard:3.21.0"
"ansible:3.21.0"

**Describe the bug**
Missing a RBAC permission for a user to trigger **upstream_pulp/replication** task.

**To Reproduce**
Install Pulp 3.28.0 version and look for the list of permissions in the "https://PULP-SERVER/pulp/api/v3/roles/" in the Django web portal. Only permission available for upstream object create/list/change/delete

**Expected behavior**
A permission all user to initiate a replication task of an upstream object.

**Additional context**
None

mchithu:
Apologize, the correct versions of Pulp and Plugin are:
**Version**
"core:3.28.0"
"certguard:3.28.0"
"file:3.28.0"
"python:3.28.0"
"rpm:3.28.0"

PR #4336, opened 2023-08-25
Adds RBAC for UpstreamPulp API

dkliban:
fixes: #3994


pulpcore/app/viewsets/replica.py
```diff
@@ -21,13 +21,91 @@ class UpstreamPulpViewSet(
...
+                "condition": [
+                    "has_model_or_domain_or_obj_perms:core.replicate_upstreampulp",
+                    "has_model_or_domain_or_obj_perms:core.view_upstreampulp",
+                ],
+            },
+            {
+                "action": ["list_roles", "add_role", "remove_role"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": ["has_model_or_domain_or_obj_perms:core.manage_roles_upstreampulp"],
```
gerrod3, line 86: Should a `view_upstreampulp` check also be here?

pulpcore/tests/functional/api/test_replication.py
```diff
@@ -112,3 +115,74 @@ def test_replication_with_wrong_ca_cert(
...
+        except ApiException as e:
+            assert e.status == outcome, f"{e}"
+        else:
+            assert status == outcome, f"User performed {action} when they shouldn't been able to"
+            return response
+
+    return _try_action
+
+
+def test_replicate_rbac(
```
gerrod3, line 155: Since this test is already requiring domains to run, can we use two new domains for the test so that we can mark this parallel?

pulpcore/tests/functional/api/test_replication.py
```diff
@@ -112,3 +115,74 @@ def test_replication_with_wrong_ca_cert(
...
+        }
+        upstream_pulp = gen_object_with_cleanup(
+            upstream_pulp_api_client, upstream_pulp_body, pulp_domain=non_default_domain.name
+        )
+
+    # Assert that Alice (upstream pulp viewer) gets a 403
+    try_action(alice, upstream_pulp_api_client, "replicate", 403, upstream_pulp.pulp_href)
+
+    # Assert that B (upstream pulp owner) gets a 202
+    try_action(bob, upstream_pulp_api_client, "replicate", 202, upstream_pulp.pulp_href)
```
mdellweg, line 185: After sucessfully replicating, do we have any content in the replica domain that we need to clean so the fixture can clean up the domain?

pulpcore/tests/functional/api/test_replication.py
dkliban: yes, let's make it a parallel test.

pulpcore/tests/functional/api/test_replication.py
dkliban: I am going to update the test to ensure that there is no content remaining in the domain which is serving as the replica.

pulpcore/app/viewsets/replica.py
dkliban: I didn't see any other viewsets that have view_<model> conditions for the roles actions. Am I misunderstanding?

pulpcore/app/models/replica.py
```diff
@@ -30,3 +30,7 @@ class UpstreamPulp(BaseModel):
 
     class Meta:
         unique_together = ("name", "pulp_domain")
+        permissions = [
+            ("replicate_upstreampulp", "Can start a replication task"),
+            ("manage_roles_upstreampulp", "Can manage roles on upstream pulps"),
```
mdellweg, line 35:
```suggestion
            ("manage_roles_upstreampulp", "Can manage roles on upstream pulp"),
```

pulpcore/app/viewsets/replica.py
```diff
@@ -21,13 +21,91 @@ class UpstreamPulpViewSet(
...
+                "action": ["destroy"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_or_domain_or_obj_perms:core.delete_upstreampulp",
+                    "has_model_or_domain_or_obj_perms:core.view_upstreampulp",
+                ],
+            },
+            {
+                "action": ["update", "partial_update"],
```
mdellweg, line 65: Are there labels on upstream pulp?

pulpcore/app/viewsets/replica.py
mdellweg: That is a good point.
Is the view permission not implied by the queryset filtering anyway? Would that not also apply to update and delete actions?

pulpcore/app/viewsets/replica.py
```diff
@@ -21,13 +21,91 @@ class UpstreamPulpViewSet(
...
+
+    LOCKED_ROLES = {
+        "core.upstreampulp_creator": ["core.add_upstreampulp"],
+        "core.upstreampulp_owner": [
+            "core.view_upstreampulp",
+            "core.change_upstreampulp",
+            "core.delete_upstreampulp",
+            "core.replicate_upstreampulp",
+            "core.manage_roles_upstreampulp",
+        ],
```
mdellweg, line 106: How about a role that can only trigger the replication, not change it?

pulpcore/tests/functional/api/test_replication.py
mdellweg: The ask here was to make the "serving" domain ephemeral too.
What if other tests constantly create and delete distributions?
I think with the label_select we still have a rather stable test here?

pulpcore/tests/functional/api/test_replication.py
mdellweg: So this intentionally does not replicate a single repository?
Do we have another test for this? Wouldn't it be better (faster) to combine these tests into one?
1. Setup two distributions with content and different labels.
2. Replicate with label_select matching one of them.
3. Verify one repo.
4. replicate again with label_select matching nothing (to cleanup)
5. verify deletion of repo

pulpcore/tests/functional/api/test_replication.py
dkliban: yeah, the label_select guarantees that no distributions get actually discovered for replication.

pulpcore/tests/functional/api/test_replication.py
dkliban: This test intentionally doesn't do any replication. I am going to open a PR against Pulp file that does what you describe.

pulpcore/app/viewsets/replica.py
dkliban: Not at this time.

pulpcore/app/viewsets/replica.py
dkliban: Hmm ... maybe. Do we have such roles for syncing repos?

mdellweg: Yes, but replicate is a bit more complex. Also you can sync using a remote without being able to modify it's proxy credentials... Similar we'd have a way to separate the upstream credentials from the ability to use the upstream pulp. I think it'd be useful.

pulpcore/app/viewsets/replica.py
```diff
@@ -21,13 +21,89 @@ class UpstreamPulpViewSet(
...
+                    "has_model_or_domain_or_obj_perms:core.change_upstreampulp",
+                ],
+            },
+            {
+                "action": ["replicate"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_or_domain_or_obj_perms:core.replicate_upstreampulp",
+                    "has_model_or_domain_or_obj_perms:core.view_upstreampulp",
```
mdellweg, line 77: I think this view perm is still superfluous.

mdellweg (changes requested):

gerrod3 (approved):

PR #4337, opened 2023-08-28
Migrate version check

### LINKED ISSUE #3401: Migrations must refuse to run if it is unsafe
## Motivation

The zero-downtime upgrade may require that all running pulp processes, i.e. workers, pulp-content, and pulp-api, are all at plugin or pulpcore version X or newer.

The zero-downtime group will ensure the database layer is backwards compatible by having db changes being only additive, but at some point a breaking db change must occur. For example if you want to rename a column, first add a new column, use both columns for a while, and then eventually remove the old column. The removal part of this example would be the breaking change.

## Proposal

The goal is to ensure migrations are always safe, and stop users from applying unsafe migrations.

* Breaking change migrations specify which versions of which components "must be" running before this change can be allowed
* Such migrations start by querying the DB for the current status (via heartbeat-msgs) of everything running in the system/cluster
* If anything is running at lower-than-required, the migration is halted
* In this situation, the "old" components are identified, and what-they-need-to-be-running is identified, and the admin is expected to upgrade/shut down any pods w/ old-code
* At that point, the upgrade can be restarted

mdellweg:
fixes #3401

### DIFF
--- a/pulpcore/migrations.py
+++ b/pulpcore/migrations.py
@@ -1,0 +1,62 @@
+from packaging.version import parse as parse_version
+
+from django.conf import settings
+from django.utils import timezone
+from django.db.migrations.operations.base import Operation
+
+
+class RequireVersion(Operation):
+    reversible = True
+
+    def __init__(self, plugin, version, hints=None):
+        self.plugin = plugin
+        self.version = version
+        self.hints = hints or {}
+        self.elidable = True
+
+    def state_forwards(self, app_label, state):
+        pass
+
+    def database_forwards(self, app_label, schema_editor, from_state, to_state):
+        from_state.clear_delayed_apps_cache()
+        ApiAppStatus = from_state.apps.get_model("core", "ApiAppStatus")
+        ContentAppStatus = from_state.apps.get_model("core", "ContentAppStatus")
+        Worker = from_state.apps.get_model("core", "Worker")
+
+        needed_version = parse_version(self.version)
+        errors = []
+
+        for worker_class, ttl, class_name in [
+            (ApiAppStatus, settings.API_APP_TTL, "api server"),
+            (ContentAppStatus, settings.CONTENT_APP_TTL, "content server"),
+            (Worker, settings.WORKER_TTL, "pulp worker"),
+        ]:
+            for worker in worker_class.objects.filter(
+                last_heartbeat__gte=timezone.now() - timezone.timedelta(seconds=ttl)
+            ):
+                present_version = worker.versions.get(self.plugin)
+                if present_version is not None and parse_version(present_version) < needed_version:
+                    errors.append(
+                        f"  - '{self.plugin}'='{present_version}' "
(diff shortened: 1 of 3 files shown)


pulpcore/migrations.py
```diff
@@ -0,0 +1,65 @@
...
+        self.version = version
+        self.hints = hints or {}
+        self.elidable = True
+
+    def state_forwards(self, app_label, state):
+        pass
+
+    def database_forwards(self, app_label, schema_editor, from_state, to_state):
+        from_state.clear_delayed_apps_cache()
+        if router.allow_migrate(schema_editor.connection.alias, app_label, **self.hints):
```
mdellweg, line 23: Reminder: What is this for? Is it safe to skip this check?

mdellweg:
Sample output:
```
RuntimeError: Incompatible versions detected (gem >= 0.4.0 needed):
  - 'gem'='0.3.0.dev' with api server '1996@e87410f347fa'
  - 'gem'='0.3.0.dev' with api server '1997@e87410f347fa'
  - 'gem'='0.3.0.dev' with content server '1998@e87410f347fa'
  - 'gem'='0.3.0.dev' with content server '2002@e87410f347fa'
  - 'gem'='0.3.0.dev' with pulp worker '703@e87410f347fa'
  - 'gem'='0.3.0.dev' with pulp worker '652@e87410f347fa'
Please shutdown or upgrade the outdated components before you continue the migration.
```

pulpcore/migrations.py
mdellweg: I could not find a good explanation and decided skipping the test is probably not desired.

pulpcore/plugin/migrations.py
```diff
@@ -0,0 +1 @@
+from pulpcore.migrations import RequirePlugin  # noqa: F401
```
gerrod3, line 1: Am I missing something, isn't the migration called `RequireVersion`, not `RequirePlugin`?

mdellweg: Why is flake8 not complaining?

mdellweg: But the name would be up for discussion. ;)

gerrod3: I'm good with the name.

pulpcore/migrations.py
```diff
@@ -0,0 +1,62 @@
+from packaging.version import parse as parse_version
+
+from django.conf import settings
+from django.utils import timezone
+from django.db.migrations.operations.base import Operation
+
+
+class RequireVersion(Operation):
+    reversible = True
```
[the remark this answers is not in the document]
mdellweg: Well, it claims that migrating down is ok.
We do not check whether to new workers are running when migrating down. Neither do we check that you do not spin up old workers again after migrating up. There's only so much you can do...

gerrod3 (approved):

ggainey (approved):

PR #4342, opened 2023-08-30
Add burst mode to worker

### LINKED ISSUE #4341: Burst mode for pulpcore-worker
When upgrading, it is sometimes not safe to bring "old" task unfinished to the new version. Though we have a backwards compatibility claim, some admins may decide they want to drain the task queue anyway.
Providing a "--burst" option to the pulpcore worker that will tell it to exit instead of going to sleep will help there.
Also this may be useful in case you have a system stuck on a bunch of sync/import/export tasks to add a few workers that will disconnect themselves once the congestion is over.

mdellweg:
This allows to run a pulpcore-worker in the foreground that will handle tasks until no more are available.
This can help to drain the task queue during an upgrade, or overcome a congestion short term.

fixes #4341

### DIFF
--- a/CHANGES/4341.feature
+++ b/CHANGES/4341.feature
@@ -1,0 +1,1 @@
+Added ``--burst`` flag to pulpcore-worker so it will terminate instead of sleeping.
--- a/docs/plugin_dev/plugin-writer/concepts/index.rst
+++ b/docs/plugin_dev/plugin-writer/concepts/index.rst
@@ -769,3 +769,9 @@ older codebases could run on newer, upgraded workers. To ensure this always work
 forever backwards compatible until the next major Pulp version. For example, you cannot have a
 breaking signature change in tasking code and if this is needed you need to make a new task name and
 preserve the old code until the next major Pulp version.
+
+.. note::
+
+   Users not performing zero downtime upgrades who are still wary of any task incompatibilities,
+   should consider running the pulpcore worker in burst mode (`pulpcore-worker --burst`) after
+   shutting down all the api and content workers to drain the task queue.
--- a/pulpcore/tasking/entrypoint.py
+++ b/pulpcore/tasking/entrypoint.py
@@ -14,9 +14,12 @@ from pulpcore.tasking.pulpcore_worker import NewPulpWorker  # noqa: E402: module
 _logger = logging.getLogger(__name__)
 
 
-@click.option("--pid", help="Write the process ID number to a file at the specified path")
+@click.option("--pid", help="Write the process ID number to a file at the specified path.")
+@click.option(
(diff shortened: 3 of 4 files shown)


pulpcore/tasking/entrypoint.py
```diff
@@ -14,9 +14,12 @@
...
 
 
-@click.option("--pid", help="Write the process ID number to a file at the specified path")
+@click.option("--pid", help="Write the process ID number to a file at the specified path.")
+@click.option(
+    "--burst/--no-burst", help="Run in burst mode; terminate when no more tasks are available."
+)
 @click.command()
-def worker(pid):
+def worker(pid, burst):
```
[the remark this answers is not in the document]
mdellweg: `--burst` is the option they use with rq workers and i wanted to do the same.

lubosmj (approved):
I like the rationale behind this change. Is there a chance that you will include an example usage in the changelog or documentation?

mdellweg:
I added a note.

CHANGES/4341.feature
```diff
@@ -0,0 +1 @@
+Added ``--burst`` flag to pulpcore-worker so it will terminate instead of sleeping.
```
ipanova, line 1: hmm this does not feel like belonging to user facing but rather to the plugin one

docs/plugin_dev/plugin-writer/concepts/index.rst
```diff
@@ -769,3 +769,9 @@ older codebases could run on newer, upgraded workers. To ensure this always work
 forever backwards compatible until the next major Pulp version. For example, you cannot have a
 breaking signature change in tasking code and if this is needed you need to make a new task name and
 preserve the old code until the next major Pulp version.
+
+.. note::
+
+   For users that do not perform zero downtime upgrades and are still wary of any tasks
+   incompatibilities, the pulpcore worker can be in burst mode (`pulpcore-worker --burst`) after
+   shutting down all the api and content workers to drain the task queue.
```
lubosmj, line 777: Users not performing zero downtime upgrades and concerned about task compatibility should consider using burst mode (`pulpcore-worker --burst`) for the Pulpcore worker after shutting down all API and content workers to clear the task queue.

CHANGES/4341.feature
mdellweg: I don't follow. This is purely interesting to the admin. You cannot even trigger this from plugin code. And the admin is what we call the user, no?

docs/plugin_dev/plugin-writer/concepts/index.rst
mdellweg: I took a mixture of both versions.

PR #4344, opened 2023-08-30
Add labels apis

### LINKED ISSUE #3332: Dedicated (nested) api's to manipulate labels on objects
Currently you can only update the `pulp_labels` object as a whole. This can lead to all the common problems with concurrent updates.
Suggestion: Add a dedicated `{pulp_obj_href}/labels/` endpoint to allow crud operations on individual labels.

mdellweg:
fixes #3332


pulpcore/app/serializers/base.py
```diff
@@ -504,3 +504,27 @@ class TaskGroupOperationResponseSerializer(serializers.Serializer):
...
         allow_null=False,
     )
+
+
+class DeleteLabelSerializer(serializers.Serializer):
+    """
+    Serializer for synchronously setting a label.
+    """
+
+    label = serializers.SlugField(required=True)
```
mdellweg, line 514: Should this be "label" or "key"?

pulpcore/tests/functional/api/using_plugin/test_labels.py
```diff
@@ -0,0 +1,63 @@
+import pytest
+
+
+@pytest.fixture
+def label_access_policy(access_policies_api_client):
```
mdellweg, line 5: Looks like we need to update some access policies to make this a reality...

pulpcore/app/viewsets/base.py
```diff
@@ -626,3 +632,46 @@ def my_permissions(self, request, pk=None):
...
+
+    @extend_schema(
+        summary="Delete a label",
+        description="Delete a single pulp_label on the object.",
+        responses={
+            204: None,
+        },
+    )
+    @action(detail=True, methods=["post"], serializer_class=DeleteLabelSerializer)
+    def delete_label(self, request, pk=None):
```
mdellweg, line 668: Should we call it "unset_label"?

pulpcore/tests/functional/api/using_plugin/test_labels.py
lubosmj: Hmm. Will this change break other plugins because of adding the `LabelsMixin` mixin to the global `RepositoryViewSet`?

lubosmj: Users would only receive accessibility errors because of missing access policies, right? Is this considered a breaking change?

mdellweg: I don't know how to roll out the updated access policies easily, but as this api did not exist before, this is not a breaking change. Nothing the user could do before is blocked after.

pulpcore/app/viewsets/base.py
```diff
@@ -626,3 +632,46 @@ def my_permissions(self, request, pk=None):
...
+                [serializer.validated_data["key"], serializer.validated_data["value"]],
+            )
+        )
+        return Response(serializer.validated_data)
+
+    @extend_schema(
+        summary="Unset a label",
+        description="Unset a single pulp_label on the object.",
+        responses={
+            204: None,
```
gerrod3, line 664: If we change this method to use Django ORM we could return the previously set label. Do you think that would be useful?

pulpcore/app/viewsets/base.py
```diff
@@ -626,3 +632,46 @@ def my_permissions(self, request, pk=None):
...
+
+class LabelsMixin:
+    @extend_schema(
+        summary="Set a label",
+        description="Set a single pulp_label on the object to a specific value or null.",
+        responses={
+            200: SetLabelSerializer,
+        },
+    )
+    @action(detail=True, methods=["put"], serializer_class=SetLabelSerializer)
```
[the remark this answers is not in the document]
mdellweg: No specific reason.
I was thinking this is able to replace a specific label and as a rest call this would be `put`. But it can also create it which should be `post`.
unsetting it is `post`, because with `delete` I cannot pass a body.
I can use `patch` because it only manipulates an aspect of the entity, but i think i'm going with `post` because that's what we use for `add_role` and `remove_role` too. And it's probably the least specific verb in html that allows a body.

The test should not even look different in the slightest way.

pulpcore/app/viewsets/base.py
```diff
@@ -626,3 +632,46 @@ def my_permissions(self, request, pk=None):
...
+        serializer = SetLabelSerializer(
+            data=request.data, context={"request": request, "content_object": obj}
+        )
+        serializer.is_valid(raise_exception=True)
+        obj._meta.model.objects.filter(pk=obj.pk).update(
+            pulp_labels=RawSQL(
+                "pulp_labels || hstore(%s, %s)",
+                [serializer.validated_data["key"], serializer.validated_data["value"]],
+            )
+        )
```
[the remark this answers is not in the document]
mdellweg: This way of updating is race condition free. I very much prefer it.

pulpcore/tests/functional/api/using_plugin/test_labels.py
```diff
@@ -0,0 +1,63 @@
...
+    orig_access_policy = access_policies_api_client.list(
+        viewset_name="repositories/file/file"
+    ).results[0]
+    new_statements = orig_access_policy.statements.copy()
+    new_statements.append(
+        {
+            "action": ["set_label", "unset_label"],
```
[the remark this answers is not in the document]
mdellweg: First this is specific to file repository. Other plugins do not interfere. I believe nothing would happen, because you'd just have two statements that "allow". I'd add a version constraint on pulp_file, but i don't know the target version yet.

pulpcore/app/viewsets/base.py
mdellweg: We can do that anyway.

pulpcore/app/viewsets/base.py
```diff
@@ -626,3 +632,40 @@ def my_permissions(self, request, pk=None):
...
         ]
         return Response({"permissions": permissions})
+
+
+class LabelsMixin:
+    @extend_schema(
+        summary="Set a label",
+        description="Set a single pulp_label on the object to a specific value or null.",
+    )
+    @action(detail=True, methods=["post"], serializer_class=SetLabelSerializer)
```
gerrod3, line 642: Does the generated API schema show the appropriate response body and return code?

mdellweg: I think, yes.
```
$ pulp --refresh-api debug openapi operation --id repositories_gem_gem_set_label | jq '.operation.responses'
{
  "201": {
    "content": {
      "application/json": {
        "schema": {
          "$ref": "#/components/schemas/SetLabelResponse"
        }
      }
    },
    "description": ""
  }
}
$ pulp --refresh-api debug openapi operation --id repositories_gem_gem_unset_label | jq '.operation.responses'
{
  "201": {
    "content": {
      "application/json": {
        "schema": {
          "$ref": "#/components/schemas/UnsetLabelResponse"
        }
      }
    },
    "description": ""
  }
}
```

gerrod3: Can you see what the `SetLabelResponse` looks like?

mdellweg:
```
$ pulp --refresh-api debug openapi schema --name SetLabelResponse
{
  "type": "object",
  "description": "Serializer for synchronously setting a label.",
  "properties": {
    "key": {
      "type": "string",
      "pattern": "^[-a-zA-Z0-9_]+$"
    },
    "value": {
      "type": "string",
      "nullable": true
    }
  },
  "required": [
    "key",
    "value"
  ]
}
$ pulp --refresh-api debug openapi schema --name UnsetLabelResponse
{
  "type": "object",
  "description": "Serializer for synchronously setting a label.",
  "properties": {
    "key": {
      "type": "string",
      "pattern": "^[-a-zA-Z0-9_]+$"
    },
    "value": {
      "type": "string",
      "readOnly": true
    }
  },
  "required": [
    "key"
  ]
}
```

gerrod3 (approved):

PR #4345, opened 2023-09-01
Apply a small fix to the installation docs

mdellweg:
[noissue]

### DIFF
--- a/docs/installation/instructions.rst
+++ b/docs/installation/instructions.rst
@@ -130,10 +130,7 @@ PyPI Installation
     $ fixfiles restore /var/run/pulpcore
     $ fixfiles restore /var/log/galaxy_api_access.log
 
-16. Run Pulp::
-
-    $ pulp-content  # The Pulp Content service (listening on port 24816)
-    $ pulpcore-manager runserver 24817  # The Pulp API service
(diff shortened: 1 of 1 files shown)


docs/installation/instructions.rst
```diff
@@ -130,10 +130,7 @@ PyPI Installation
     $ fixfiles restore /var/run/pulpcore
     $ fixfiles restore /var/log/galaxy_api_access.log
 
-16. Run Pulp::
-
-    $ pulp-content  # The Pulp Content service (listening on port 24816)
-    $ pulpcore-manager runserver 24817  # The Pulp API service
+16. Run or restart all Pulp services.
```
gerrod3, line 133: This is under the directions of the PyPI installation right? Is there not a command we can give the user to tell them how to run the Pulp services?

mdellweg: Yes, pip installation method.
Well `pulp-content` never existed as a command and `pulpcore-manager runserver` is about the worst you can do. In this guide we advise the admin to look into the other doc about templating the systemd files from the deprecated pulp_installer. Not the best we could do, but the best we have.

dralley (approved):

decko (approved):

PR #4347, opened 2023-09-01
Use proper infos for task import paths

### LINKED ISSUE #4346: Improve the import path for tasks
Currently task function need to be top level citizens of their module.

mdellweg:
fixes #4346


mdellweg:
If you let me, I can use this to demonstrate how such a change is done in a zero downtime safe way.

mdellweg:
Hmm exposing the import path of a task as its name to the user does not sound like the best idea in hindside...

PR #4362, opened 2023-09-05
Allow downloading content directly via HTTP(S)

MichalPysik:
Work in progress...

ref https://github.com/pulp/pulpcore/issues/4608

### DIFF
--- a/pulpcore/plugin/serializers/content.py
+++ b/pulpcore/plugin/serializers/content.py
@@ -5,14 +5,16 @@ from tempfile import NamedTemporaryFile
 
 from django.db import DatabaseError
 from rest_framework.serializers import (
+    CharField,
     FileField,
     Serializer,
     ValidationError,
 )
 from pulpcore.app.files import PulpTemporaryUploadedFile
-from pulpcore.app.models import Artifact, Upload, UploadChunk
+from pulpcore.app.models import Artifact, Upload, UploadChunk, Remote
 from pulpcore.app.serializers import (
     RelatedField,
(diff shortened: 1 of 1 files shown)


pulpcore/plugin/serializers/content.py
```diff
@@ -31,8 +33,23 @@ class UploadSerializerFieldsMixin(Serializer):
...
+        queryset=Remote.objects.all(),
+        required=False,
+    )
+
+    remote_relative_path = CharField(
+        help_text=_("Path where the file is located relative to the remote's base URL"),
+        write_only=True,
+        required=False, #TODO later validate that this field is set if remote is set
+    )
+
```
mdellweg, line 50: I wonder if grouping the download related fields in a nested serializer would allow to make the serialization easier and contain some of the actual download logic in a confined place.
https://www.django-rest-framework.org/api-guide/serializers/#dealing-with-nested-objects

dkliban:
I think you want to add these fields to the `SingleArtifactContentUploadSerializer`.

mdellweg:
> I think you want to add these fields to the `SingleArtifactContentUploadSerializer`.

If we can make it work for `NoArtifactContentUpload`, why waste an opportunity?

pulpcore/plugin/serializers/content.py
MichalPysik: @mdellweg I had no idea that was possible, but I was wondering whether there is a way to somehow group the new fields together right before I wrote the TODO comment. I will definetly look into that.

PR #4373, opened 2023-09-08
Marked repository delete as unsafe for immetiate

### LINKED ISSUE #4372: Need a way to force *async* execution of a DELETE task
**Version**
pulpcore-3.28.10
pulp-rpm-3.22.3

**Describe the bug**
Deleting a handful of repositories with large RPM content now takes hours instead of minutes.

**To Reproduce**
1. sync RHEL 8 baseos and appstream repos from CDN
2. try to delete them
3. since all delete calls are now synchronous, the deletion is performed in a singlethreaded fashion, instead of being distributed across available pulp workers (8 or 16, depending on the installation).  It now takes hours instead of couple of minutes.  In addition, during this time the API gunicorn worker is completely blocked and is unavailable to process other requests.
**Expected behavior**
I would like an API addition (eg. a flag ?async=true) that would force my DELETE /repositories/rpm/rpm/ call to be asynchronous and distributed to either an available worker or, failing that, put in the pending queue.
**Additional context**
Performance of a repository deletion invocation operation (not the actual time it takes to really delete the repo) is very important for RHUI ops.  Sometimes a repo needs to be deleted and re-created to get rid of the repository version history and its associated artifacts.  More importantly, this performance is vital for the RHUI automation and CI during development.

mdellweg:
Deleting repositories casade deletes a lot of RepostoryVersions and RepositoryContent. This can easily extend beyond the gunicorn worker timeout.

fixes #4372

### DIFF
--- a/pulpcore/app/viewsets/exporter.py
+++ b/pulpcore/app/viewsets/exporter.py
@@ -49,6 +49,9 @@ class ExporterViewSet(
     ViewSet for viewing exporters.
     """
 
+    # Too many cascaded deletes to block the gunicorn worker.
+    ALLOW_NON_BLOCKING_DELETE = False
+
     queryset = Exporter.objects.all()
     serializer_class = ExporterSerializer
     endpoint_name = "exporters"
--- a/CHANGES/4372.bugfix
+++ b/CHANGES/4372.bugfix
@@ -1,0 +1,2 @@
+Ensured that repository and remote delete is always executed in a task. This fixes a bug when the
+deletion blocks the gunicorn api worker for too long.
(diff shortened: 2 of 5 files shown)


pulpcore/app/viewsets/exporter.py
```diff
@@ -49,6 +49,9 @@ class ExporterViewSet(
     ViewSet for viewing exporters.
     """
 
+    # To many cascaded deletes to block the gunicorn worker.
```
ipanova, line 52: s/to many/too many

ipanova:
typo in the commit message  s/immetiate/immediate

ipanova (commented):
Mind adding here to the warning note https://github.com/pulp/pulpcore/commit/18f1aeb9d3d051c5ecede16d2d002c004023e96c#diff-bfa7009599f2d46ccc57e55065d53f8ac4a1f732d0a56242eac7b61c6f134818R236 that also objects with cascade deletes can cause trouble and exceed the worker timeout

ipanova (approved):
Looks good, thank you!

gerrod3 (approved):

Issue #2079, opened 2022-01-17
As a user, I can see the total size in bytes that a repository's files use on disk


fao89:
Author: mhrivnak (mhrivnak)

Bugzilla: https://bugzilla.redhat.com/buglist.cgi?quicksearch=1375716
Redmine Issue: 2261, https://pulp.plan.io/issues/2261

---

This should simply be the sum of the size of all files associated with units that are associated with the repository.

This does not include data stored in the database.

For on-demand content, files that are known in the database but have not yet been downloaded should not be counted in the total.

This does not account for the fact that the same unit can appear in multiple repos without incurring additional disk storage use. It will be up to the user to interpret these numbers for individual repos, and consider totals across multiple repos in the context that content may be shared.

A natural way to represent this would be as an attribute of a repository, but that doesn't have to be the implementation if a better option presents itself.

fao89:
From: @bmbouter (bmbouter)
Date: 2016-12-10T17:16:13Z

---

Before grooming can a full example be given of a repo detail view that also shows the new field's being added?

Also what about having it sum them by type and also give a total. For example:

~~~
 "size": {
    "total": 238260,
    "rpm": 227341,
    "drpm: 10919,
  }, 
~~~

fao89:
From: mhrivnak (mhrivnak)
Date: 2016-12-13T16:58:15Z

---

How would we obtain the total size in a generic way? Maybe we could add a "size" attribute to FileContentUnit, and leave it up to plugins to populate that if/when possible. Would this be optional?

What about on_demand? If a file hasn't been downloaded, should the unit's size be 0? It seems more intuitive that it should be the expected size after download. But what about the size of the repo? I'm not sure why, but my intuition is the opposite: that a repo's size should be 0 if none of its units have been downloaded. Maybe the algorithm would roughly be this:

~~~
size = 0
for unit in repo:
  if unit.downloaded is True:
    size += unit.size
~~~

For that reason, maybe the repo's attribute would be better-named "disk_use", or something like that.

Presumably the repo's size would have to be updated under any of these circumstances:

  - content added (sync/copy/upload)
  - content removed (sync/remove)
  - unit marked as downloaded

What about "shared content", used by ostree, where multiple units can reference the same files? I wonder if the ostree tooling itself has a standard, "expected" way to represent size of a repo where there may be overlap between branches.

Given the above questions, I wonder if this is complex enough that it should wait for pulp 3.

fao89:
[...]

PR #4376, opened 2023-09-08
Add new management command `repository-size` that calculates the size of artifacts stored in a repository

gerrod3:
fixes: #2079

Some plugins (cough, cough RPM) will need to override the new property to get an accurate number for their repositories.

### DIFF
--- a/pulpcore/app/management/commands/repository-size.py
+++ b/pulpcore/app/management/commands/repository-size.py
@@ -1,0 +1,127 @@
+import json
+import re
+import sys
+from gettext import gettext as _
+
+from argparse import RawDescriptionHelpFormatter
+from django.core.management import BaseCommand, CommandError
+from django.conf import settings
+
+from pulpcore.app.models import Repository
+from pulpcore.app.util import get_url, extract_pk
+
+
+def gather_repository_sizes(repositories, include_versions=False, include_on_demand=False):
+    """
+    Creates a list containing the size report for given repositories.
+
+    Each entry in the list will contain a dict with following minimal fields:
+        - name: name of the repository
+        - href: href of the repository
+        - disk-size: size in bytes of all artifacts stored on disk in the repository
+
+    Each entry can additionally have the optional fields if specified:
+        - on-demand-size: approximate size in bytes of all on-demand artifacts in the repository
+        - versions: size report list of each version in the repository
+
+    The version list if specified will have the following fields:
+        - version: the version number
+        - disk-size: size in bytes of all artifacts stored on disk for this repository version
+        - on-demand-size: if specified, the approx size in bytes of on-demand artifacts in version
+
+    **Note**: With artifact deduplication and additive repository versions, the sum total of each
+    repository version size might be greater than the actual size of the repository. Sizes
+    calculated in this report are based on the stored sizes in Pulp's database, they do not take
+    into account missing artifacts or incorrect/missing on-demand artifact sizes.
+    """
+    full_report = []
+    for repo in repositories.order_by("name").iterator():
+        try:
+            repo = repo.cast()
+            report = {"name": repo.name, "href": get_url(repo), "disk-size": repo.disk_size}
+            if include_on_demand:
+                report["on-demand-size"] = repo.on_demand_size
+            if include_versions:
+                versions = []
+                try:
+                    for version in repo.versions.iterator():
+                        v_report = {"version": version.number, "disk-size": version.disk_size}
+                        if include_on_demand:
+                            v_report["on-demand-size"] = version.on_demand_size
+                        versions.append(v_report)
+                except Exception:
+                    pass
+                else:
+                    report["versions"] = versions
+        except Exception:
+            continue
+        else:
+            full_report.append(report)
+
+    return full_report
+
+
+def href_list_handler(value):
+    """Common list parsing for a string of hrefs."""
+    r = rf"({settings.API_ROOT}(?:[-_a-zA-Z0-9]+/)?api/v3/repositories/[-_a-z]+/[-_a-z]+/[-a-f0-9]+/)"  # noqa: E501
+    return re.findall(r, value)
+
+
+class Command(BaseCommand):
+    """Django management command for calculating the storage size of a repository."""
+
+    help = __doc__ + gather_repository_sizes.__doc__
+
+    def add_arguments(self, parser):
+        """Set up arguments."""
+        parser.add_argument(
+            "--repositories",
+            type=href_list_handler,
+            required=False,
+            help=_(
+                "List of repository hrefs to generate the report from. Leave blank to include"
+                " all repositories in all domains. Mutually exclusive with domain."
+            ),
+        )
+        parser.add_argument(
+            "--include-versions",
+            action="store_true",
+            help=_("Include repository versions in report"),
+        )
+        parser.add_argument(
+            "--include-on-demand",
+            action="store_true",
+            help=_("Include the approximate on-demand artifact sizes"),
+        )
+        parser.add_argument(
+            "--domain",
+            default=None,
+            required=False,
+            help=_(
+                "The pulp domain to gather the repositories from if specified. Mutually"
+                " exclusive with repositories."
+            ),
+        )
+        parser.formatter_class = RawDescriptionHelpFormatter
+
+    def handle(self, *args, **options):
+        """Implement the command."""
+        domain = options.get("domain")
+        repository_hrefs = options.get("repositories")
+        if domain and repository_hrefs:
+            raise CommandError(_("--domain and --repositories are mutually exclusive"))
+
+        repositories = Repository.objects.all()
+        if repository_hrefs:
+            repos_ids = [extract_pk(r) for r in repository_hrefs]
+            repositories = repositories.filter(pk__in=repos_ids)
+        elif domain:
+            repositories = repositories.filter(pulp_domain__name=domain)
+
+        report = gather_repository_sizes(
+            repositories,
+            include_versions=options["include_versions"],
+            include_on_demand=options["include_on_demand"],
+        )
+        json.dump(report, sys.stdout, indent=4)
+        print()
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -33,7 +33,7 @@ from pulpcore.exceptions import ResourceImmutableError
 from pulpcore.cache import Cache
 
 from .base import MasterModel, BaseModel
-from .content import Artifact, Content
+from .content import Artifact, Content, ContentArtifact, RemoteArtifact
 from .fields import EncryptedTextField
 from .task import CreatedResource, Task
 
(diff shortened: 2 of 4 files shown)


pulpcore/app/serializers/repository.py
```diff
@@ -49,6 +50,18 @@ class RepositorySerializer(ModelSerializer):
...
+        help_text=_(
+            "Approximate size on disk of all artifacts in the repository across all its versions."
+        ),
+    )
+
+    @extend_schema_field(serializers.IntegerField(read_only=True, allow_null=True))
+    def get_disk_size(self, obj):
+        if view := self.context.get("view", None):
+            if hasattr(view, "is_master_viewset") and not view.is_master_viewset():
+                return obj.disk_size
```
gerrod3, line 63: Would like some feedback on this trade-off I made. Following the issue description I made `disk_size` a calculated field on request time, but it is a semi-expensive calculation and it incurs an additional query for each repository being serialized. I initially had the field only be calculated on a retrieve, but I ran into the tests failing since we expect the results of a list?limit=1 to be the same as a retrieve and I think this expectation is common across all our plugins, thus I couldn't just fix solely the pulpcore tests. So I implemented this compromise to disable the field only on the generic endpoints and leave it on for all the plugin endpoints. Depending on how many repositories are being queried and how large they are this could create a huge performance hit for normal list operations. 

Questions:

1. Does anybody suggestions for a better dynamic solution? I experimented with a bunch of different queries and couldn't come up with a better one.
2. Should I try making this an integer field on the repository/version that is updated upon new version creation & possibly a periodically scheduled task? Tradeoff: Less accurate reading on `disk_size`, but guarantee no performance penalty on list + ability to sort by `disk_size`

dralley: >Should I try making this an integer field on the repository/version that is updated upon new version creation & possibly a periodically scheduled task? Tradeoff: Less accurate reading on disk_size, but guarantee no performance penalty on list + ability to sort by disk_size

Note that we do something like that for content counts.  Something along those lines would be my preference, personally.

https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/repository.py#L1050

mdellweg: This seems to be about the actually currently consumed disk space that may vary from second to second with on-demand content. I'd vote against a periodic job. Presenting outdated data will only cause questions. Have you considered providing it as a separate view? Like `GET .../repositories/file/file/asdf/disk_use/`. I think that is better than providing a `POST .../repositories/file/file/asdf/recalc_disk_use/` later.
Since we need to give plugins a chance to subclass the algorithm, as you pointed out, we can probably not just annotate the big Repository queryset with it.

> since we expect the results of a list?limit=1 to be the same as a retrieve

This is actually not true. For tasks at least we have a MinimalTaskSerializer for that. And a list of size 1 is still a list as specified by api.json.

ggainey:
A **lot** of discussion happened at the pulpcore meeting 2023-09-12 - see the [notes taken](https://hackmd.io/@pulp/core/?view#September-12-2023)

pulpcore/app/management/commands/repository-size.py
```diff
@@ -0,0 +1,114 @@
...
+            continue
+        else:
+            full_report.append(report)
+
+    return full_report
+
+
+def href_list_handler(value):
+    """Common list parsing for a string of hrefs."""
+    r = rf"({settings.API_ROOT}(?:[-_a-zA-Z0-9]+/)?api/v3/repositories/[-_a-z]+/[-_a-z]+/[-a-f0-9]+/)"  # noqa: E501
```
mdellweg, line 66: To speed things up, you can compile the regex (and save it as a global constant).

mdellweg: But maybe we are using it only once... (findall should do the compilation then.)

pulpcore/app/management/commands/repository-size.py
```diff
@@ -0,0 +1,114 @@
...
+            "--repositories",
+            type=href_list_handler,
+            required=False,
+            help=_(
+                "List of repository hrefs to generate the report from. Leave blank to include"
+                " all repositories in domain"
+            ),
+        )
+        parser.add_argument(
+            "--include-versions",
```
ipanova, line 87: In what situation will it be useful to see all versions? Should we straight away show always latest?

pulpcore/app/models/repository.py
```diff
@@ -82,6 +82,35 @@ class Meta:
...
+    def disk_size(self):
+        """Returns the approximate size on disk for all artifacts stored across all versions."""
+        all_content = (
+            RepositoryContent.objects.filter(repository=self)
+            .distinct("content")
+            .values_list("content")
+        )
+        return (
+            Artifact.objects.filter(content__in=all_content)
+            .distinct()
```
ipanova, line 95: ideally we would not have artifacts shared among content, do you feel we need to account for such case?

pulpcore/app/models/repository.py
```diff
@@ -82,6 +82,35 @@ class Meta:
...
+
+    @property
+    def on_demand_size(self):
+        """Returns the approximate size of all on-demand artifacts stored across all versions."""
+        all_content = Content.objects.filter(
+            pk__in=RepositoryContent.objects.filter(repository=self)
+            .distinct("content")
+            .values_list("content")
+        )
+        on_demand_ca = ContentArtifact.objects.filter(content__in=all_content, artifact=None)
```
ipanova, line 107: you can pass directly result from values_list(content)

pulpcore/app/models/repository.py
```diff
@@ -82,6 +82,35 @@ class Meta:
         unique_together = ("name", "pulp_domain")
         verbose_name_plural = "repositories"
 
+    @property
+    def disk_size(self):
+        """Returns the approximate size on disk for all artifacts stored across all versions."""
+        all_content = (
+            RepositoryContent.objects.filter(repository=self)
+            .distinct("content")
```
ipanova, line 90: what's the benefit of calling it?

pulpcore/app/models/repository.py
```diff
@@ -249,6 +278,22 @@ def natural_key(self):
         """
         return (self.name,)
 
+    @staticmethod
+    def on_demand_artifacts_for_version(version):
```
ipanova, line 282: you should probably call this differently, since you're discarding RA with no size

ipanova: I see, you're mentioning this in the note

pulpcore/app/models/repository.py
gerrod3: You can add content, remove it, then add it again. So the same content can show up multiple times through RepositoryContent, thus the need for distinct.

pulpcore/app/models/repository.py
gerrod3: Yes, just to be safe. I don't know the details of each plugin, but we allow plugin writers to do what they want with how they store artifacts. One example - in file upload a file with relative-path "file1", and then upload the same file with relative-path "file2". These two content units will share the same artifact, but are perfectly valid to be stored in the same repository version.

pulpcore/app/management/commands/repository-size.py
```diff
@@ -0,0 +1,114 @@
...
+    **Note**: With artifact deduplication and additive repository versions, the sum total of each
+    repository version size might be greater than the actual size of the repository. Sizes
+    calculated in this report are based on the stored sizes in Pulp's database, they do not take
+    into account missing artifacts or incorrect/missing on-demand artifact sizes.
+    """
+    full_report = []
+    for repo in repositories.order_by("name").iterator():
+        try:
+            repo = repo.cast()
+            report = {"name": repo.name, "href": get_url(repo), "disk-size": repo.disk_size}
```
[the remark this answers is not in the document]
gerrod3: Would like more opinions on this. I prefer using dashes, but I could go either way.

pulpcore/app/management/commands/repository-size.py
gerrod3: Not sure, maybe if someone performs an on-demand sync of one remote and immediate sync of another, then seeing the two versions will show which content hasn't been downloaded into Pulp yet.

pulpcore/app/models/repository.py
ipanova: yes, but you're anyway passing this to the artifact filter `Artifact.objects.filter(content__in=[c1,c1,c2,c3]`

gerrod3: That is true, the second distinct in the artifact filter would handle the duplicate content units, but I wonder if there is a performance impact. Are the two distinct clauses slower than one distinct clause over a larger query?

ipanova: To my knowledge distinct is an expensive operation, especially on postgres. But maybe this is a better tradeoff, we should run thme side by side and see the diff

ipanova:
```
pulp=# SELECT DISTINCT "core_artifact"."pulp_id", "core_artifact"."pulp_created", "core_artifact"."pulp_last_updated", "core_artifact"."file", "core_artifact"."size", "core_artifact"."md5", "core_artifact"."sha1", "core_artifact"."sha224", "core_artifact"."sha256", "core_artifact"."sha384", "core_artifact"."sha512", "core_artifact"."timestamp_of_interest", "core_artifact"."pulp_domain_id" FROM "core_artifact" INNER JOIN "core_contentartifact" ON ("core_artifact"."pulp_id" = "core_contentartifact"."artifact_id") WHERE "core_contentartifact"."content_id" IN (SELECT U0."content_id" FROM "core_repositorycontent" U0 WHERE U0."repository_id" = '0a022744-1dee-4f87-bf3b-d7b660f75ca5');


 pulp_id | pulp_created | pulp_last_updated | file | size | md5 | sha1 | sha224 | sha256 | sha384 | sha512 | timestamp_of_interest | pulp_domain_id 
---------+--------------+-------------------+------+------+-----+------+--------+--------+--------+--------+-----------------------+----------------
(0 rows)

Time: 14794.854 ms (00:14.795)
pulp=# 
pulp=# 
pulp=# SELECT DISTINCT "core_artifact"."pulp_id", "core_artifact"."pulp_created", "core_artifact"."pulp_last_updated", "core_artifact"."file", "core_artifact"."size", "core_artifact"."md5", "core_artifact"."sha1", "core_artifact"."sha224", "core_artifact"."sha256", "core_artifact"."sha384", "core_artifact"."sha512", "core_artifact"."timestamp_of_interest", "core_artifact"."pulp_domain_id" FROM "core_artifact" INNER JOIN "core_contentartifact" ON ("core_artifact"."pulp_id" = "core_contentartifact"."artifact_id") WHERE "core_contentartifact"."content_id" IN (SELECT DISTINCT ON (U0."content_id") U0."content_id" FROM "core_repositorycontent" U0 WHERE U0."repository_id" = '0a022744-1dee-4f87-bf3b-d7b660f75ca5');
 pulp_id | pulp_created | pulp_last_updated | file | size | md5 | sha1 | sha224 | sha256 | sha384 | sha512 | timestamp_of_interest | pulp_domain_id 
---------+--------------+-------------------+------+------+-----+------+--------+--------+--------+--------+-----------------------+----------------
(0 rows)

Time: 9087.610 ms (00:09.088)
``

File: pulpcore/app/models/repository.py (hunk 4 above)
Comment (ipanova) [reply pulpcore/app/models/repository.py:95]: I think specifying `distinct("pulp_id") ` should be faster because it will not need to compare each field form the table but just one

Comment (gerrod3) [reply pulpcore/app/models/repository.py:95]: Django does not allow it :cry: 

`NotImplementedError: aggregate() + distinct(fields) not implemented`

Comment (ipanova) [reply pulpcore/app/models/repository.py:95]: bummer

**Comment by ipanova** [comment]
Conceptually this  PR looks good to me.There is question around `-` vs `_` so we need more input on that.

File: pulpcore/app/management/commands/repository-size.py [hunk 9]
```diff
@@ -0,0 +1,127 @@
...
+            action="store_true",
+            help=_("Include the approximate on-demand artifact sizes"),
+        )
+        parser.add_argument(
+            "--domain",
+            default=None,
+            required=False,
+            help=_(
+                "The pulp domain to gather the repositories from if specified. Mutually"
+                " exclusive with repositories."
```
ipanova, line 102: :+1

mdellweg:
> Conceptually this PR looks good to me.There is question around `-` vs `_` so we need more input on that.

This is creating JSON output. If I knew this was meant to be used in JS code later, i'd def opt for `_`. But here, i have no real preference.

pulpcore/app/models/repository.py
mdellweg: > Yes, just to be safe. I don't know the details of each plugin, but we allow plugin writers to do what they want with how they store artifacts. One example - in file upload a file with relative-path "file1", and then upload the same file with relative-path "file2". These two content units will share the same artifact, but are perfectly valid to be stored in the same repository version.

Also an rpm could be in a file repository and at the same time in a RPM repo. I would even promote such use cases.

mdellweg (approved):
LGTM

ipanova (approved):
I don't want to hold off this PR because of  one -/_ outstanding question

PR #4381, opened 2023-09-11
Fix the create release branch workflow

dkliban:
[noissue]


dralley:
@dkliban I'm assuming this is already covered by having updated from the template

PR #4387, opened 2023-09-12
Reclaim space optimizations

ipanova:
[noissue]

### DIFF
--- a/pulpcore/app/tasks/reclaim_space.py
+++ b/pulpcore/app/tasks/reclaim_space.py
@@ -50,25 +50,21 @@ def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
         c_reclaim_qs = c_reclaim_qs.exclude(pk__in=rv_content)
 
     content_distinct = c_reclaim_qs.distinct("pulp_type")
-    unprotected = []
+    protected = []
     for content in content_distinct:
-        if not content.cast().PROTECTED_FROM_RECLAIM:
-            unprotected.append(content.pulp_type)
-
-    ca_qs = ContentArtifact.objects.select_related("content", "artifact").filter(
-        content__in=c_reclaim_qs.values("pk"), artifact__isnull=False
+        if content.cast().PROTECTED_FROM_RECLAIM:
+            protected.append(content.pulp_type)
+    c_reclaim_qs = c_reclaim_qs.exclude(pulp_type__in=protected)
+    ca_qs = ContentArtifact.objects.filter(
+        content_id__in=c_reclaim_qs.values("pk"), artifact_id__isnull=False
     )
     if not force:
         ca_qs = ca_qs.filter(remoteartifact__isnull=False)
     artifact_pks = set()
-    ca_to_update = []
     for ca in ca_qs.iterator():
-        if ca.content.pulp_type in unprotected:
-            artifact_pks.add(ca.artifact.pk)
-            ca.artifact = None
-            ca_to_update.append(ca)
+        artifact_pks.add(ca.artifact_id)
+    ca_qs.update(artifact_id=None)
 
-    ContentArtifact.objects.bulk_update(objs=ca_to_update, fields=["artifact"], batch_size=1000)
     artifacts_to_delete = Artifact.objects.filter(pk__in=artifact_pks)
     progress_bar = ProgressReport(
         message="Reclaim disk space",


pulpcore/app/tasks/reclaim_space.py
```diff
@@ -31,12 +31,14 @@ def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
...
+    for repo in reclaimed_repos.iterator():
         repo.invalidate_cache(everything=True)
 
     domain = get_domain()
-    rest_of_repos = Repository.objects.filter(pulp_domain=domain).exclude(pk__in=repo_pks)
-    c_keep_qs = Content.objects.filter(repositories__in=rest_of_repos)
+    rest_of_repos = (
+        Repository.objects.filter(pulp_domain=domain).exclude(pk__in=repo_pks).values("pk")
+    )
+    c_keep_qs = Content.objects.filter(repositories__in=rest_of_repos).values("pk")
```
mdellweg, line 41: I thought django would do this transparently for subqueries.

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -31,12 +31,14 @@ def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
 
     """
     reclaimed_repos = Repository.objects.filter(pk__in=repo_pks)
-    for repo in reclaimed_repos:
+    for repo in reclaimed_repos.iterator():
```
mdellweg, line 34: We don't expect to get a ton of repos here, right?

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -46,29 +48,27 @@ def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
...
         rv_content = Content.objects.none()
         for rv in rv_qs.iterator():
-            rv_content |= rv.content
+            rv_content |= rv.content.values("pk")
         c_reclaim_qs = c_reclaim_qs.exclude(pk__in=rv_content)
 
+    c_reclaim_qs = c_reclaim_qs.only("pk", "pulp_type")
     content_distinct = c_reclaim_qs.distinct("pulp_type")
-    unprotected = []
+    protected = []
```
mdellweg, line 56: Since we do not expect so many protected types, this may be worth doing.

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -46,29 +48,27 @@ def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
...
     for ca in ca_qs.iterator():
-        if ca.content.pulp_type in unprotected:
-            artifact_pks.add(ca.artifact.pk)
-            ca.artifact = None
-            ca_to_update.append(ca)
+        artifact_pks.add(ca.artifact_id)
+        ca.artifact_id = None
 
-    ContentArtifact.objects.bulk_update(objs=ca_to_update, fields=["artifact"], batch_size=1000)
+    ContentArtifact.objects.bulk_update(objs=ca_qs, fields=["artifact_id"], batch_size=1000)
```
mdellweg, line 71: How is this working? Can bulk_update operate on querysets? I only knew about the list of objects to pass.
Wouldn't `ca_qs.update(artifact=None)` be even easier?

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -46,29 +48,27 @@ def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
...
     artifacts_to_delete = Artifact.objects.filter(pk__in=artifact_pks)
```
mdellweg, line 72: Why is reclaim not honoring touch?

pulpcore/app/tasks/reclaim_space.py
ipanova: i hope so? :)

pulpcore/app/tasks/reclaim_space.py
ipanova: oh no, that would not really work, i was too eager to remove `ca_to_update`.
I think the only benefit of using the bulk_update is that it has the batch_size which is handy here

pulpcore/app/tasks/reclaim_space.py
ipanova: what do you mean? it will never honor it. touch updates timestamp_of_interest that is being used in orphan cleanup.
Here, in this task we do not have any orphans, artifacts/content are part of repos, but we still want to reclaim space given that these artifacts have not been identified in other repos than the ones we want to reclaim.

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -50,25 +52,24 @@ def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
         c_reclaim_qs = c_reclaim_qs.exclude(pk__in=rv_content)
 
     content_distinct = c_reclaim_qs.distinct("pulp_type")
-    unprotected = []
+    protected = []
     for content in content_distinct:
-        if not content.cast().PROTECTED_FROM_RECLAIM:
-            unprotected.append(content.pulp_type)
-
-    ca_qs = ContentArtifact.objects.select_related("content", "artifact").filter(
```
dralley, line 58: +1

pulpcore/app/tasks/reclaim_space.py
mdellweg: What about an artifact that was just touched by a sync? I doubt that we should delete it.

ipanova: Yes there is a small window, same as in running sync and orphan clean up. Do you have a suggestion on how to improve it? 
We cannot respect the touch here because it will cancel the whole concept of reclaim.

pulpcore/app/tasks/reclaim_space.py
mdellweg: But the other `update` call is a rather simple SQL query (not the crazy stuff, bulk_update creates). I would guess it'd be much faster.

pulpcore/app/tasks/reclaim_space.py
mdellweg: you mean when running OC with protection_time=0?
At least that is not the topic of this issue. The only thing I can think of is, not deleting artifacts that were recently touched by using the `orphaned` queryset here.

ipanova: No. I mean an artifact can still be deleted once it has been identified as orphan and moment after it was touched in the sync pipeline.
There are 2 outcomes of such race window 1) sync will fail because it would touch something that will be removed 2) orphan clean up will fail with ProtectedError if the sync task will get to creating a content by the time artifact will be deleted in orphan clean up task https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/orphan.py#L98

ipanova: edit: 2) will not fail anymore because we catch the ProtectedError.

mdellweg: As i said, this is off topic. Let's not borrow trouble here and ignore my comment. Worst thing that can happen: You need to restart the sync.

ipanova: Exactly. I wanted to point out that this problem already exists in orphan clean up  same as with reclaim space. Restarting the sync will take care fixing both of such race windows.

pulpcore/app/tasks/reclaim_space.py
ipanova: It would be, yes. Besidies what I've said in the other comment I would not know that's the other differences between these 2. I could switch to qs.update

mdellweg: As far as i remember (,and i recently looked at the implementation for different reasons, )bulk_update creates a huge switch statement (in SQL) with a case for every item in the batch. While the `update` call will basically generate:
`UPDATE artifact_id=null INTO TABLE core.contentartifact WHERE id IN SELECT <whatever your qs generates>` (making stuff up here.)
The big difference is that we are setting some field to the _same_ value (null in this case) for all the rows, while `bulk_update` is built to keep track of individual values for every row in the batch.

mdellweg: Do you have any results or at lease an impression of this actually being less cpu intense? I'd prefer not to approve it from pure suspicion.

ipanova: I have not done any testing and was going to ask Sayan to test this patch. There are few low hanging fruits which i have collected like got rid of select_related and avoided in the in the loop accessing ca.artifact.pk from another table

ipanova:
I dropped everything else, it had negligent effect.
Reclaim was performed on ruby-gems repo with 260k units.  Since it was synced on demand, I needed to locally hack this line to artifact_is_isnull=True https://github.com/pulp/pulpcore/pull/4387/files#diff-12478e665bbf683b52e6f27c967eeac697edc7f31db33cd5795f303c2aae3135R59
Before changes
memory_before

After changes
memory_final

Mostly it  was bulk_update and a bit of select_related()

dralley:
@ipanova Do you plan to undraft this?

ipanova:
@dralley Sayan never came back to me with the evidence of reclaim space task needing optimizations. I will just close it and pick up if needed when problem arises.

dralley:
The graphs do show a substantial improvement, what do you mean by evidence?

PR #4412, opened 2023-09-13
Ensure that non-chunked exports also use compressionlevel=1

### LINKED ISSUE #4411: Also set the compression level when not using chunked exports
We didn't use the `compresslevel` argument because it didn't work in streaming mode, so we went for a monkeypatch.  But the monkeypatch doesn't apply to non-streaming mode (oversight on my part)... which we also use, in the non-chunked path.

We don't actually need a monkeypatch on the other side because the argument `compresslevel=1` is supported for non-streaming mode, but we do still have to pass that arg, so let's do that.

dralley:
closes #4411

### DIFF
--- a/CHANGES/4411.bugfix
+++ b/CHANGES/4411.bugfix
@@ -1,0 +1,1 @@
+Ensure non-chunked exports also use gzip ``compressionlevel=1``
--- a/pulpcore/app/tasks/export.py
+++ b/pulpcore/app/tasks/export.py
@@ -3,6 +3,7 @@ import json
 import logging
 import os
 import os.path
+import sys
 import subprocess
 import tarfile
 
@@ -414,8 +415,15 @@ def pulp_export(exporter_pk, params):
                 stdin=subprocess.PIPE,
             ) as split_process:
                 try:
-                    with tarfile.open(tarfile_fp, "w|gz", fileobj=split_process.stdin) as tar:
-                        _do_export(pulp_exporter, tar, the_export)
+                    # on Python < 3.12 we have a monkeypatch which enables compression levels
+                    if sys.version_info.major == 3 and sys.version_info.minor < 12:
+                        with tarfile.open(tarfile_fp, "w|gz", fileobj=split_process.stdin) as tar:
+                            _do_export(pulp_exporter, tar, the_export)
+                    else:
+                        with tarfile.open(
+                            tarfile_fp, "w|gz", fileobj=split_process.stdin, compresslevel=1
+                        ) as tar:
+                            _do_export(pulp_exporter, tar, the_export)
                 except Exception:
                     # no matter what went wrong, we can't trust the files we (may have) created.
                     # Delete the ones we can find and pass the problem up.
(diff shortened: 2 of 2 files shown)


pulpcore/app/tasks/export.py
```diff
@@ -414,8 +415,15 @@ def pulp_export(exporter_pk, params):
                 stdin=subprocess.PIPE,
             ) as split_process:
                 try:
-                    with tarfile.open(tarfile_fp, "w|gz", fileobj=split_process.stdin) as tar:
-                        _do_export(pulp_exporter, tar, the_export)
+                    # on Python < 3.12 we have a monkeypatch which enables compression levels
+                    if sys.version_info.major == 3 and sys.version_info.minor < 12:
+                        with tarfile.open(tarfile_fp, "w|gz", fileobj=split_process.stdin) as tar:
+                            _do_export(pulp_exporter, tar, the_export)
+                    else:
```
dralley, line 422: It occurred to me that the other patch would probably break on Python >=3.12 also

pulpcore/app/tasks/export.py
```diff
@@ -433,7 +441,7 @@ def pulp_export(exporter_pk, params):
         else:
             # write into the file
             try:
-                with tarfile.open(tarfile_fp, "w:gz") as tar:
+                with tarfile.open(tarfile_fp, "w:gz", compresslevel=1) as tar:
```
dralley, line 444: `with tarfile.open(tarfile_fp, "w:gz", compresslevel=1) as tar:` = supported universally.  So no need to do a monkeypatch here.

`with tarfile.open(tarfile_fp, "w|gz", compresslevel=1) as tar:` = not supported universally.  Hence the original monkeypatch.

CHANGES/4411.bugfix
```diff
@@ -0,0 +1 @@
+Ensure non-chunked exports also use gzip compressionlevel=1
```
mdellweg, line 1:
```suggestion
Ensure non-chunked exports also use gzip ``compressionlevel=1``.
```

pulpcore/app/tasks/export.py
mdellweg: It is still runny by our monkeypatch, no?

pulpcore/app/tasks/export.py
mdellweg: So with Python 3.12 we can finally remove the monkeypatch? XD

mdellweg (approved):

pulpcore/app/tasks/export.py
dralley: Not sure what you mean?  

`with tarfile.open(tarfile_fp, "w:gz", compresslevel=1) as tar:` doesn't use the monkeypatch because it's not using streaming mode, see `"w:gz"` rather than `"w|gz" (yes that's confusing at a glance).  But it doesn't need the monkeypatch.

It's only streaming mode where `compresslevel=1` isn't supported until 3.12

dralley: Yes we only replace some implementation on the `_Stream` class, which apparently doesn't get activated by other codepaths.

dralley: https://github.com/python/cpython/blob/main/Lib/tarfile.py#L1812-L1842

mdellweg: Yep, I still stand by my review. Let's get this merged.

ggainey (approved):

PR #4414, opened 2023-09-13
Add test machinery for OpenTelemetry

### LINKED ISSUE #4323: Add integration tests to cover OpenTelemetry stack integration
Add the tests and the machinery needed to cover the OpenTelemetry stack integration.

This should cover the traffic between the instrumented code and the collector that collects trace and metrics spans.

lubosmj:
Tests checking OpenTelemetry payloads are now run in serial.

closes #4323

credits to @mdellweg for designing the machinery

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -6,7 +6,9 @@ import socket
 import ssl
 import subprocess
 import threading
+import time
 import uuid
+import warnings
 
 import trustme
 import proxy
@@ -18,6 +20,7 @@ from dataclasses import dataclass
 from packaging.version import parse as parse_version
 from time import sleep
 from yarl import URL
+from opentelemetry.proto.trace.v1.trace_pb2 import TracesData
 
 from pulpcore.tests.functional.utils import (
     SLEEP_TIME,
@@ -411,6 +414,26 @@ class ThreadedAiohttpServerData:
             protocol_handler = "https://"
 
         return f"{protocol_handler}{self.host}:{self.port}{path}"
+
+
+@pytest.fixture(scope="session")
+def received_otel_span():
+    def _received_otel_span(data):
+        if os.environ.get("PULP_OTEL_ENABLED") != "true":
+            # pretend everything is working as expected if tests are run from
+            # a non-configured runner
+            return True
+
+        async def _send_request():
+            async with aiohttp.ClientSession(raise_for_status=True) as session:
+                otel_server_url = os.environ.get("OTEL_EXPORTER_OTLP_ENDPOINT")
+                async with session.post(f"{otel_server_url}/test", json=data) as response:
+                    return response.status
+
+        status = asyncio.run(_send_request())
+        return True if status == 200 else False
+
+    return _received_otel_span
 
 
 # Webserver Fixtures
--- a/.github/workflows/scripts/install.sh
+++ b/.github/workflows/scripts/install.sh
@@ -120,7 +120,7 @@ if [ "$TEST" = "azure" ]; then
     command: "azurite-blob --blobHost 0.0.0.0 --cert /etc/pulp/azcert.pem --key /etc/pulp/azkey.pem"' vars/main.yaml
   sed -i -e '$a azure_test: true\
 pulp_scenario_settings: {"domain_enabled": true}\
-pulp_scenario_env: {}\
+pulp_scenario_env: {"otel_bsp_max_export_batch_size": 1, "otel_bsp_max_queue_size": 1, "otel_exporter_otlp_endpoint": "http://localhost:4318", "otel_exporter_otlp_protocol": "http/protobuf", "pulp_otel_enabled": "true"}\
 ' vars/main.yaml
 fi
 
--- a/.github/workflows/scripts/pre_before_script.sh
+++ b/.github/workflows/scripts/pre_before_script.sh
@@ -1,0 +1,6 @@
+set -euv
+
+if [[ "$TEST" == "azure" ]]; then
+    cmd_stdin_prefix bash -c "cat > /var/lib/pulp/scripts/otel_server.py" < "$GITHUB_WORKSPACE"/pulpcore/tests/functional/assets/otel_server.py
+    docker exec -d "$PULP_CI_CONTAINER" "$@" su pulp -c "python3 /var/lib/pulp/scripts/otel_server.py"
+fi
--- a/CHANGES/4323.misc
+++ b/CHANGES/4323.misc
@@ -1,0 +1,1 @@
+Added functional tests to test the OpenTelemetry integration.
--- a/pulpcore/tests/functional/api/test_status.py
+++ b/pulpcore/tests/functional/api/test_status.py
@@ -1,4 +1,5 @@
 """Test the status page."""
+import os
 import pytest
 
 from django.conf import settings
@@ -57,18 +58,31 @@ STATUS = {
 }
 
 
+@pytest.fixture
+def test_path():
+    return os.getenv("PYTEST_CURRENT_TEST").split()[0]
+
+
 @pytest.mark.parallel
-def test_get_authenticated(status_api_client):
+def test_get_authenticated(test_path, status_api_client, received_otel_span):
     """GET the status path with valid credentials.
 
     Verify the response with :meth:`verify_get_response`.
     """
     response = status_api_client.status_read()
     verify_get_response(response.to_dict(), STATUS)
+    assert received_otel_span(
+        {
+            "http.method": "GET",
+            "http.target": "/pulp/api/v3/status/",
+            "http.status_code": 200,
+            "http.user_agent": test_path,
+        }
+    )
 
 
 @pytest.mark.parallel
-def test_get_unauthenticated(status_api_client, anonymous_user):
+def test_get_unauthenticated(test_path, status_api_client, anonymous_user, received_otel_span):
     """GET the status path with no credentials.
 
     Verify the response with :meth:`verify_get_response`.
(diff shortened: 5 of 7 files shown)


pulpcore/tests/functional/__init__.py
```diff
@@ -413,10 +416,104 @@ def make_url(self, path):
...
+    collector_server.join()
+    if expected_spans:
+        raise Exception(
+            f"Expected spans are still missing: {expected_spans}\n"
+            f"Unexpeced spans received: {unexpected_spans}"
+        )
 
 
 @pytest.fixture
+def expect_span(_otel_collector):
```
mdellweg, line 505: Should we call it expect_otel_span ?

.github/workflows/scripts/post_before_install.sh
```diff
@@ -0,0 +1,10 @@
+if [ "$TEST" = "azure" ]; then
+    # configuration variables required by the opentelemetry tests
+    export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318
+    export OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
+    export OTEL_EXPORTER_OTLP_TIMEOUT=1
+    export OTEL_BSP_SCHEDULE_DELAY=10
+    export OTEL_BSP_EXPORT_TIMEOUT=10
+    export OTEL_BSP_MAX_QUEUE_SIZE=1
+    export OTEL_BSP_MAX_EXPORT_BATCH_SIZE=1
```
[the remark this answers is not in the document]
lubosmj: Yes, at least, it worked locally. We need to incorporate these variables into the CI.

lubosmj:
How big of a deal is to run the status tests in serial? There might be a case where the same code for running the collector server is executed from multiple processes simultaneously, resulting in "address already in use" errors. Maybe, because of the following issue: https://stackoverflow.com/questions/61216175/how-to-run-setup-fixture-once-and-then-tests-in-parallel.

In the near future, we can end up running more and more tests in serial because of this issue. For now, is it possible to run a separate Python server next to the pytest process? The server will represent a standalone otel collector that will be tickled (by sending otel data from Pulp, triggered by tests) and queried (by retrieving otel data back, used for test assertions) by the pytest process. Thus, the server will act as a temporary storage for otel data.

lubosmj:
Should we worry about these errors in the logs? They are not harmful, see the azure runner.
```
  File "/usr/local/lib/python3.8/site-packages/opentelemetry/sdk/metrics/_internal/export/__init__.py", line 523, in _receive_metrics
    self._exporter.export(
  File "/usr/local/lib/python3.8/site-packages/opentelemetry/exporter/otlp/proto/http/metric_exporter/__init__.py", line 189, in export
    resp = self._export(serialized_data.SerializeToString())
  File "/usr/local/lib/python3.8/site-packages/opentelemetry/exporter/otlp/proto/http/metric_exporter/__init__.py", line 162, in _export
    return self._session.post(
  File "/usr/local/lib/python3.8/site-packages/requests/sessions.py", line 637, in post
    return self.request("POST", url, data=data, json=json, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python3.8/site-packages/requests/sessions.py", line 703, in send
    r = adapter.send(request, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/requests/adapters.py", line 519, in send
    raise ConnectionError(e, request=request)
requests.exceptions.ConnectionError: HTTPConnectionPool(host='localhost', port=4318): Max retries exceeded with url: /v1/metrics (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fd6ddfd6f70>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

These errors are raised when otel data cannot reach the destination collector point. The collector server is initialized only when running `test_status` tests. However, the otel data are sent to the collector throughout the whole testing process.

.github/workflows/scripts/post_before_install.sh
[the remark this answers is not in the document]
lubosmj: I have managed to scale down the number of required settings. Thanks, @decko, for finding these.

.ci/ansible/Containerfile.j2
```diff
@@ -30,6 +30,14 @@ RUN pip3 install
 {%- endif -%}
 {%- endfor %}
 
+{% if azure_test | default(false) %}
+ENV PULP_OTEL_ENABLED=true
+ENV OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318
+ENV OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
+ENV OTEL_BSP_MAX_QUEUE_SIZE=1
+ENV OTEL_BSP_MAX_EXPORT_BATCH_SIZE=1
+{% endif %}
```
lubosmj, line 39: This gets overwritten when re-applying the plugin template. How should we tackle that? I think the variables need to be set before running `/init`.

mdellweg: Can we use the plugin template to inject scenario specific env-vars? In a similar way we can currently inject settings. And yes, env-vars would superseed settings by the power of dynaconf. XD

lubosmj: Done in https://github.com/pulp/plugin_template/pull/786.

lubosmj:
The tests seem to be working because I could not find "Telemetry tests are disabled" warnings in the logs. 🐕‍🦺

decko:
> Should we worry about these errors in the logs? They are not harmful, see the azure runner.
> 
> ```
> Traceback (most recent call last):
>   File "/usr/local/lib/python3.8/site-packages/opentelemetry/sdk/metrics/_internal/export/__init__.py", line 523, in _receive_metrics
>     self._exporter.export(
>   File "/usr/local/lib/python3.8/site-packages/opentelemetry/exporter/otlp/proto/http/metric_exporter/__init__.py", line 189, in export
>     resp = self._export(serialized_data.SerializeToString())
>   File "/usr/local/lib/python3.8/site-packages/opentelemetry/exporter/otlp/proto/http/metric_exporter/__init__.py", line 162, in _export
>     return self._session.post(
>   File "/usr/local/lib/python3.8/site-packages/requests/sessions.py", line 637, in post
>     return self.request("POST", url, data=data, json=json, **kwargs)
>   File "/usr/local/lib/python3.8/site-packages/requests/sessions.py", line 589, in request
>     resp = self.send(prep, **send_kwargs)
>   File "/usr/local/lib/python3.8/site-packages/requests/sessions.py", line 703, in send
>     r = adapter.send(request, **kwargs)
>   File "/usr/local/lib/python3.8/site-packages/requests/adapters.py", line 519, in send
>     raise ConnectionError(e, request=request)
> requests.exceptions.ConnectionError: HTTPConnectionPool(host='localhost', port=4318): Max retries exceeded with url: /v1/metrics (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fd6ddfd6f70>: Failed to establish a new connection: [Errno 111] Connection refused'))
> ```
> 
> These errors are raised when otel data cannot reach the destination collector point. The collector server is initialized only when running `test_status` tests. However, the otel data are sent to the collector throughout the whole testing process.

My only concern is that we get lots of noise/garbage that make the test logs useless.

pulpcore/tests/functional/__init__.py
```diff
@@ -413,10 +416,98 @@ def make_url(self, path):
...
+    # Not sure what the right time to wait is here but the server does not seem
+    # to be found immediately.
+    for _ in range(30):
+        if unexpected_spans:
+            # Reset unexpected spans
+            unexpected_spans = []
+            break
+        # Tickle the telemetry stack
+        status_api_client.status_read()
+        time.sleep(1)
```
decko, line 483: With this, won't we be generating 30 spans?
If it's about waiting for something, could we just use time.sleep(30)?

lubosmj: We might generate 30 spans when the server is not up and ready for 30 seconds. We exit the loop immediately if we receive any span from "tickling". That means the server is running and receiving data.

lubosmj:
> My only concern is that we get lots of noise/garbage that make the test logs useless.

Good point. Is there any way to enable the telemetry just for the status endpoint? Would using `autouse=True` in the otel server fixture work for us?

mdellweg:
> > My only concern is that we get lots of noise/garbage that make the test logs useless.
> 
> Good point. Is there any way to enable the telemetry just for the status endpoint? Would using `autouse=True` in the otel server fixture work for us?

We want to assert telemetry on many other tests too, right? The status endpoint was just the guinea pig.

lubosmj:
@decko, @mdellweg, having defined `autouse=True` for the otel collector fixture removed the errors almost completely. There are still a couple of them at the beginning and end of the test run (for instance, when the fixture is not yet ready or was destroyed but pulp still sends some bits of data). What do you think?

lubosmj:
Three proposals:
1. EASY. Leave the code as it is, remove `autouse=True`, and ignore the fact that we flood logs with "Connection refused".
2. MEDIUM. Monkeypatch `opentelemetry-python` from tests to skip logging "Connection Refused" errors (https://github.com/open-telemetry/opentelemetry-python/blob/0c382ee5356e1fa5a292e7a6c4961390cdfbe6df/opentelemetry-sdk/src/opentelemetry/sdk/trace/export/__init__.py#L368). *I am now working with this proposal.*
3. MEDIUM. Run a standalone HTTP server process that will run in the background and act as a storage for otel data.

lubosmj:
I was not able to monkeypatch the otel SDK.

But, running the otel collector server in a separate process before executing all tests worked for me. The tracebacks are almost gone. Only a couple of them are shown in the logs as Pulp starts up and the server is not yet up and running (because it is dispatched from the `before_script` script).

pulpcore/tests/functional/otel_server.py
```diff
@@ -0,0 +1,103 @@
+import asyncio
```
mdellweg, line 1: Nice work!
I would have preferred to run this as a thread in the pytest process, but we want something working now.
Could this live in `pulpcore/tests/functional/assets` or similar?

mdellweg (approved):
Let's use this approach and see how far we get.

pulpcore/tests/functional/otel_server.py
lubosmj: Running this in the thread is very tricky. We want to have the server running always to not spam the logs with "connection refused". At the same time, we want to access the data from the server to enable assertions from tests. If the tests are run in parallel, we do not know which worker will pick a specific test even if we ensure that the server fixture is run from gw0. Thus, there is no guarantee we can access the server's data eventually.

mdellweg: Yes, that is not manageable on the fixture level, but maybe as a pytest extension somehow. But also not something we need to experiment with today.

decko (approved):
I think it's a GO.
Great work @lubosmj

Issue #1936, opened 2022-01-17
Sync issue when FILE_UPLOAD_TEMP_DIR volume different from STATIC_ROOT


pulpbot:
Author: wibbit (wibbit)

Redmine Issue: 7635, https://pulp.plan.io/issues/7635

---

When running a remote sync (tested against RPM plugin), if the FILE_UPLOAD_TEMP_DIR is a separate file system to STATIC_ROOT, the downloaded assets are not cleared down until the end of the sync job is completed.

This will become a significant issue when setting up multiple large repositories at the same time.

It would be ideal if as each asset is downloaded and copied to the STATIC_ROOT directory, that the item would be cleared out of the FILE_UPLOAD_TEMP_DIR, as opposed to waiting for the sync job to complete entirely.

pulpbot:
From: wibbit (wibbit)
Date: 2020-10-01T09:32:18Z

---

Correction, this is probably the WORKING_DIRECTORY which I have configured to be the same as the FILE_UPLOAD_TEMP_DIR, which is on a different partition to the STATIC_ROOT directory.

mchithu:
In a environment, where the **WORKING_DIRECTORY** (local disk) & **MEDIA_ROOT**(S3) are different volumes. While sync the repositories, it is often fill up the space in the **WORKING_DIRECTORY** (**_/var/lib/pulp/tmp_**) and this result the worker pods gets terminated.

Why the temp files under **_/var/lib/pulp/tmp_** is not cleaned up when the task completes?

PR #4428, opened 2023-09-13
Delete downloaded tmp files after Artifact creation during sync

gerrod3:
fixes: #1936

This should help clean up a worker's directory during a sync and keep their storage usage down, especially for concurrent syncing when duplicate artifacts have been downloaded.

### DIFF
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 from gettext import gettext as _
 import logging
 
+from aiofiles import os as aos
 from asgiref.sync import sync_to_async
 from django.db.models import Prefetch, prefetch_related_objects, Q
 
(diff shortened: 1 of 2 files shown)


pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -274,15 +275,20 @@ async def run(self):
...
                     da_to_save_ordered,
                     await sync_to_async(Artifact.objects.bulk_get_or_create)(
                         d_artifact.artifact for d_artifact in da_to_save_ordered
                     ),
+                    da_tmp_files,
                 ):
                     d_artifact.artifact = artifact
+                    # Delete the downloaded tmp file if it still exists to clear up space
+                    if await aos.path.exists(tmp_file_path):
+                        await aos.remove(tmp_file_path)
```
lubosmj, line 291: This is a reactive approach to a real problem that exists somewhere else. Can we do something precautionary about the removal inside `ArtifactFileField`? If not, is not a try/except a much safer variant in this matter?

lubosmj: Hmm, actually, we do a similar thing in the content handler: https://github.com/pulp/pulpcore/blob/fda96451224ae1da98d6f795e44af2e33713d595/pulpcore/content/handler.py#L805-L824. Perhaps, this is the right approach.

mdellweg (approved):

pulpcore/plugin/stages/artifact_stages.py
dralley: As a sidenote it would perhaps be better if we could use asyncio inside the handler, rather than a blocking call.

dralley (approved):

dralley:
@gerrod3 I'm not sure automated tests are practical, but it was manually tested?

gerrod3:
> @gerrod3 I'm not sure automated tests are practical, but it was manually tested?

Yes, I did. I added a logging statement inside the if statement and performed a sync with the Filesystem backend and the Minio backend. On filesystem no logs where made since filesystem moves the file to permanent storage. On Minio storage it logged for each artifact synced.

dralley:
Great thanks!

dralley:
@gerrod3 Should this be backported?

gerrod3:
> @gerrod3 Should this be backported?

Yes, even on filesystem backends it could help out when duplicate files get downloaded, so Katello/Satellite could benefit from it and Galaxy will definitely benefit from it.

PR #4433, opened 2023-09-15
With TASK_DIAGNOSTICS enabled, tasks will be automatically profiled

dralley:
[noissue]

### DIFF
--- a/pulpcore/tasking/worker.py
+++ b/pulpcore/tasking/worker.py
@@ -1,5 +1,6 @@
 from gettext import gettext as _
 
+import importlib
 import logging
 import os
 import random
@@ -421,13 +422,17 @@ class PulpcoreWorker:
             self.shutdown()
 
 
-def write_memory_usage(task_pk):
+def task_diagnostics_dir(task_pk):
+    # create a directory in /var/tmp/pulp/<task_id>/ and return it
     taskdata_dir = VAR_TMP_PULP / str(task_pk)
     taskdata_dir.mkdir(parents=True, exist_ok=True)
-    memory_file_dir = taskdata_dir / "memory.datum"
-    _logger.info("Writing task memory data to {}".format(memory_file_dir))
+    return taskdata_dir
 
-    with open(memory_file_dir, "w") as file:
+
+def write_memory_usage(path):
+    _logger.info("Writing task memory data to {}".format(path))
+
+    with open(path, "w") as file:
         file.write("# Seconds\tMemory in MB\n")
         seconds = 0
         while True:
@@ -458,10 +463,14 @@ def _perform_task(task_pk, task_working_dir_rel_path):
     signal.signal(signal.SIGHUP, child_signal_handler)
     signal.signal(signal.SIGUSR1, child_signal_handler)
     if settings.TASK_DIAGNOSTICS:
+        diagnostics_dir = task_diagnostics_dir(task_pk)
+        mem_diagnostics_path = diagnostics_dir / "memory.datum"
         # It would be better to have this recording happen in the parent process instead of here
         # https://github.com/pulp/pulpcore/issues/2337
-        normal_thread = threading.Thread(target=write_memory_usage, args=(task_pk,), daemon=True)
-        normal_thread.start()
+        mem_diagnostics_thread = threading.Thread(
+            target=write_memory_usage, args=(mem_diagnostics_path,), daemon=True
+        )
+        mem_diagnostics_thread.start()
     # All processes need to create their own postgres connection
     connection.connection = None
     task = Task.objects.select_related("pulp_domain").get(pk=task_pk)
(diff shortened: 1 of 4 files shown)


pulpcore/tasking/worker.py
```diff
@@ -471,4 +477,19 @@ def _perform_task(task_pk, task_working_dir_rel_path):
...
-    execute_task(task)
+    if settings.TASK_DIAGNOSTICS:
+        from pyinstrument import Profiler
+        with Profiler() as profiler:
+            execute_task(task)
+
+        with open(diagnostics_dir / "pyinstrument.html", "w+") as f:
+            f.write(profiler.output_html())
+
+        # import cProfile
```
dralley, line 488: pyinstrument pros:

* nicer default output than cprofile, which needs external tools to wrangle
* async-aware
* according to the comparison done by Emery Berger of Scalene fame, more accurate
* less overhead

cProfile pros:

* in the standard library

dralley: Here's what the Pyinstrument  HTML report looks like when opened in a browser.

Screenshot from 2023-09-15 14-13-03

bmbouter:
:100: A thousand times yes to this idea!

It needs docs too though. I so much enjoy seeing this! Great job.

bmbouter:
Regarding cprofile vs pyinstrument I'm +1 to pyinstrument because we need it to be async aware, and it's just easier to consume the output. We need to make this really easy. I think picking up a dependency (and its dependencies) is worth it.

That's my take, but I'm interested in what others think about that also.

pulpcore/tasking/worker.py
```diff
@@ -471,4 +480,18 @@ def _perform_task(task_pk, task_working_dir_rel_path):
...
+    if settings.TASK_DIAGNOSTICS and has_pyinstrument:
+        from pyinstrument import Profiler
+
+        with Profiler() as profiler:
+            execute_task(task)
+
+        profile_file = diagnostics_dir / "pyinstrument.html"
+        _logger.info("Writing task profile data to {}".format(profile_file))
+        with open(profile_file, "w+") as f:
+            f.write(profiler.output_html())
```
lubosmj, line 495: Does stacking multiple profiles into the same output file really work here?

pulpcore/tasking/worker.py
```diff
@@ -458,9 +463,13 @@ def _perform_task(task_pk, task_working_dir_rel_path):
...
     if settings.TASK_DIAGNOSTICS:
+        diagnostics_dir = task_diagnostics_dir(task_pk)
+        mem_diagnostics_path = diagnostics_dir / "memory.datum"
         # It would be better to have this recording happen in the parent process instead of here
         # https://github.com/pulp/pulpcore/issues/2337
-        normal_thread = threading.Thread(target=write_memory_usage, args=(task_pk,), daemon=True)
+        normal_thread = threading.Thread(
+            target=write_memory_usage, args=(mem_diagnostics_path,), daemon=True
+        )
         normal_thread.start()
```
mdellweg, line 473: What is "normal" about this thread? It is exceptional to run it, and it's marked as daemon. Can we give this a better name?

lubosmj (approved):

pulpcore/tasking/worker.py
```diff
@@ -471,4 +480,18 @@ def _perform_task(task_pk, task_working_dir_rel_path):
     set_current_user(user)
     set_domain(task.pulp_domain)
     os.chdir(task_working_dir_rel_path)
-    execute_task(task)
+
+    # set up profiling
+    has_pyinstrument = importlib.util.find_spec("pyinstrument") is not None
```
mdellweg, line 485: Hah, I only knew `try except ImportError`.

pulpcore/tasking/worker.py
dralley: What do you mean by multiple profiles?

lubosmj: The results of each task run are written to the same file. And those results are the "profiles".

dralley: Should only be one profile per task.  `diagnostics_dir` is `/var/tmp/pulp/<task_id>/`

ipanova: each task run or other words each task, will have its own pk which is the disagnostics_dir

pulpcore/tasking/worker.py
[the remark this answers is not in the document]
mdellweg: Do you mean by lazy evaluation? like:
```python
if settings.TASK_DIAGNOSTICS and importlib.util.find_spec("pyinstrument") is not None:
    ....
```
Given we execute these lines with every task, i think we can actually spare us some disc-access here.

pulpcore/tasking/worker.py
dralley: yes

mdellweg (approved):

ipanova:
@dralley could you ad a line or 2 to the docs https://docs.pulpproject.org/pulpcore/plugin_dev/plugin-writer/concepts/tasks/diagnostics.html#diagnostics

PR #4435, opened 2023-09-15
Remove compression from exports by default

dralley:
Removes compression entirely, using the special gzip level 0 mode, which packs everything into a gzip archive without actually doing any compression.

The reason for this is that even the lowest level of compression, level 1, is still a massive cost without actually providing much benefits due to the fact that most of the exported artifact payloads (rpms, images, etc) are already compressed.

On a test system, the difference between exporting with gzip level 9 and completely uncompressed was 11.3gb vs 11.8gb, while the former took 2.5x as long.

For most major users of the import/export functionality, storage space is much less of an issue than runtime.

In the future we could do any of the following:

* make the compression level configurable for users who actually do care
* get rid of gzip and therefore monkeypatches entirely - you're not supposed to use exports from a different version of Pulp, so we don't have to care about that compatibility
* try out a more suitable algorithm such as zstd or lz4, which are much faster than e.g. gzip level 1 for similarly amounts of compression

### DIFF
--- a/pulpcore/app/tasks/export.py
+++ b/pulpcore/app/tasks/export.py
@@ -29,8 +29,6 @@ from pulpcore.app.models import (
 from pulpcore.app.models.content import ContentArtifact
 from pulpcore.app.serializers import PulpExportSerializer
 
-# ensure the compression patch is loaded
-from pulpcore.app import monkeypatch  # noqa
 from pulpcore.app.util import get_version_from_model
 from pulpcore.app.importexport import (
     export_versions,
@@ -381,6 +379,8 @@ def pulp_export(exporter_pk, params):
         ValidationError: When path is not in the ALLOWED_EXPORT_PATHS setting,
             OR path exists and is not a directory
     """
+    DEFAULT_COMPRESSION = 1
+
     pulp_exporter = PulpExporter.objects.get(pk=exporter_pk)
     serializer = PulpExportSerializer(data=params, context={"exporter": pulp_exporter})
     serializer.is_valid(raise_exception=True)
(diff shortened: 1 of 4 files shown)


dralley:
This is what an export task profile looks like afterwards

Screenshot from 2023-09-15 22-30-14

Probably

pulpcore/app/tasks/export.py
```diff
@@ -416,12 +416,20 @@ def pulp_export(exporter_pk, params):
             ) as split_process:
                 try:
                     # on Python < 3.12 we have a monkeypatch which enables compression levels
+                    # see https://github.com/pulp/pulpcore/issues/3869
                     if sys.version_info.major == 3 and sys.version_info.minor < 12:
+                        from pulpcore.app import monkeypatch
+
+                        monkeypatch.patch_tarfile_default_compression_level(DEFAULT_COMPRESSION)
```
mdellweg, line 423: Instead of patching the tarfile every time we intend to use it (with potentially a different default compression level), can we create a ContextVar in monkeypatch that we can set here for the current context?

dralley: We're only patching it once, considering that this happens inside the task itself, and that process will die at the end.

ipanova:
> edit: hmm, you know, we might be calculating that checksum twice, once for each chunk and once for the global checksum. _Probably_ the global checksum should just be a checksum-of-checksums.

I don't think calculating checksum of checkusm would give same result as calculating checksum on whole file.
I think we can't do much here because we do want to provide correct checksums in that json file so later on chunk checksums and  whole file checksum can be checked once transferred to the air gaped system.

dralley:
>I don't think calculating checksum of checkusm would give same result as calculating checksum on whole file.

I agree that's definitely true, we can't change it in backports, but we could change it going forwards.

ipanova (approved):

mdellweg (approved):

dralley:
I did manual backports

PR #4439, opened 2023-09-18
Handle empty list for href and id in filters

### LINKED ISSUE #4437: pulp_href__in filter doesn't handle empty set properly
We make use of the `pulp_href__in` filter and it works great. We store pulp hrefs for some various things and we can easily look up the resources in pulp using this filter.

There is an edge case that I don't think it's properly handling though. It certain cases, the set will be empty and when we pass that to pulp, it seems that *all* results get returned. I think that *no* results should instead be returned.

Current behavior:

```
$ http :8080/pulp/api/v3/repositories/ pulp_href__in==
{
    "count": 1,
    "next": null,
    "previous": null,
    "results": [
        {
            "description": null,
            "latest_version_href": "/pulp/api/v3/repositories/deb/apt/1cc3ba78-bbf6-4373-b742-725785f010be/versions/0/",
            "name": "pmc_cli_test_repo_8fa0535e-c76f-4a60-8c1d-257290ec88c2",
            "pulp_created": "2023-09-18T12:46:41.212285Z",
            "pulp_href": "/pulp/api/v3/repositories/deb/apt/1cc3ba78-bbf6-4373-b742-725785f010be/",
            "pulp_labels": {},
            "remote": null,
            "retain_repo_versions": null,
            "versions_href": "/pulp/api/v3/repositories/deb/apt/1cc3ba78-bbf6-4373-b742-725785f010be/versions/"
        }
    ]
}
```

Expected behavior:

No results (ie `[]`) get returned.

daviddavis:
Fix bug where empty list for the `pulp_href__in` and `pulp_id__in` filters return all results instead of no results. The problem is that when using FilterMethod, django-filter just simply ignores empty set and returns the full list of results:

https://github.com/carltongibson/django-filter/blob/e4a70a667a0bf3882fd44b557bc76583d2c65cd1/django_filters/filters.py#L804-L805

There's no way to update the method to prevent this since this logic happens before the method is called. Instead, this change moves the method to the filter and adds a None check.

fixes #4437

### DIFF
--- a/pulpcore/filters.py
+++ b/pulpcore/filters.py
@@ -124,11 +124,19 @@ class HyperlinkRelatedFilter(filters.Filter):
 
 
 class IdInFilter(BaseInFilter, filters.UUIDFilter):
-    pass
+    def filter(self, qs, value):
+        if value is None:
+            return qs
+        return qs.filter(pk__in=value)
 
 
 class HREFInFilter(BaseInFilter, filters.CharFilter):
-    pass
+    def filter(self, qs, value):
+        if value is None:
+            return qs
+
+        pks = [extract_pk(href) for href in value]
+        return qs.filter(pk__in=pks)
 
 
 class PulpTypeFilter(filters.ChoiceFilter):
@@ -270,8 +278,8 @@ class BaseFilterSet(filterset.FilterSet):
     """
 
     help_text = {}
-    pulp_id__in = IdInFilter(field_name="pk", lookup_expr="in")
-    pulp_href__in = HREFInFilter(field_name="pk", method="filter_pulp_href")
+    pulp_id__in = IdInFilter(field_name="pk")
+    pulp_href__in = HREFInFilter(field_name="pk")
     q = ExpressionFilter()
 
     FILTER_DEFAULTS = {
@@ -306,11 +314,6 @@ class BaseFilterSet(filterset.FilterSet):
         "search": _("matches"),
         "ne": _("not equal to"),
     }
-
-    def filter_pulp_href(self, queryset, name, value):
-        # Convert each href to a pk
-        pks = [extract_pk(href) for href in value]
-        return queryset.filter(pk__in=pks)
 
     @classmethod
     def get_filters(cls):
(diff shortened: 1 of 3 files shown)


gerrod3:
Nice fix! Can you add a check for empty lists to the `test_pulp_id_href_filter` in `test_filter.py`?

pulpcore/filters.py
```diff
@@ -124,11 +124,19 @@ def filter(self, qs, value):
 
 
 class IdInFilter(BaseInFilter, filters.UUIDFilter):
-    pass
+    def filter(self, qs, value):
+        if value is None:
+            return qs
+        return qs.filter(pk__in=value)
```
mdellweg, line 130: Does this issue also apply when there is not a filter method provided? I'm now a bit confused by the link you provided to django-filter.

mdellweg (approved):

pulpcore/filters.py
daviddavis: I filed a bug for the problem: https://github.com/carltongibson/django-filter/issues/1609

daviddavis: Oh, when a filter method is NOT used, there is no such issue. It just calls the `filter` method on the filter without checking against EMPTY_VALUES.

mdellweg: Hmm, I guess this is where all filters get this by default:

https://github.com/carltongibson/django-filter/blob/e4a70a667a0bf3882fd44b557bc76583d2c65cd1/django_filters/filters.py#L153

dralley:
Does this need backports?  And how far back?

daviddavis:
We requested this feature so I wouldn't be surprised if we're the only ones using these filters. Moreover, it's an edge case that I am not sure many users will hit. That said, it's been available in pulp since 3.24.0.

daviddavis:
For us though, we're still on an older version of pulpcore (until https://github.com/pulp/pulp_deb/pull/295 is merged) and we've patched our version.

dralley (approved):

PR #4453, opened 2023-09-20
Removed force_close from all downloaders

### LINKED ISSUE #4452: Pulpcore prevents the use of keep-alive and may exhaust port numbers in large syncs
**Version**
Probably all.

**Describe the bug**
Pulp set's force_close on all downloaders and all aiohttp sessions. This prevents any keep-alive functionality and can lead to an exhaustion of tcp ports for large syncs. Some suboptimal network configuration and you end up with servers closing connections, because they believe they are still in time_wait state.

**Steps to reproduce**
Unclear. But maybe a proxy is involved, multiplying the connection creation dilemma.

**Expected behavior**
Pulp should use HTTP/1.1 features when available.

**Additional context**
Creating millions of tcp connections may even be a performance bottleneck.

mdellweg:
This flag causes large syncs to fail under certain circumstances, because the port range is used up too fast and reusing ports in time_wait is not appreciated by servers.

We cannot possibly fix all broken http servers in the wild.

fixes #4452

### DIFF
--- a/CHANGES/4452.bugfix
+++ b/CHANGES/4452.bugfix
@@ -1,0 +1,2 @@
+Removed a workaround to force close all tcp connections in sync leading to an exhaustion of port
+numbers and their reuse while in time_wait state.
--- a/pulpcore/download/factory.py
+++ b/pulpcore/download/factory.py
@@ -48,10 +48,6 @@ class DownloaderFactory:
     http://aiohttp.readthedocs.io/en/stable/client_quickstart.html#timeouts Behaviorally, it should
     allow for an active download to be arbitrarily long, while still detecting dead or closed
     sessions even when TCPKeepAlive is disabled.
-
-    Also for http and https urls, even though HTTP 1.1 is used, the TCP connection is setup and
-    closed with each request. This is done for compatibility reasons due to various issues related
-    to session continuation implementation in various servers.
     """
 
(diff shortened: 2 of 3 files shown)


bmbouter (approved):
We should try this and see if there are regressions reported by users. It's hard to know what will happen because this fix was introduced to workaround real webservers in the wild, so our automated tests won't help us. Still though, let's try it.

Also FYI, the original issue we worked around was a checksum mismatch bug, so if we start getting those reported we should know this removal could be related.

@mdellweg ty for making this.

bmbouter:
@dralley I think you're an expert in our downloaders, wdyt about being second reviewer?

dralley (approved):
I'm also +1 on this, my only question (and I'm not strongly in favor of requiring this) is whether we ought to stick it behind an off-by-default feature flag just in case.

But either way we could undo the change fairly easily and quickly if we start encountering such issues again.  As long as we recognize them as such.

Note that since we generally call `remote.get_downloader()` I believe we still end up with a new downloader on every request anyway?  And, thus, probably a new session?  https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/repository.py#L507

mdellweg:
We'll add the feature flag if it is still encountered as an issue.

PR #4459, opened 2023-09-21
Small import/export improvements

dralley:
Preparation for bigger stuff.  This can be more easily backported as-is.

### DIFF
--- a/pulpcore/app/importexport.py
+++ b/pulpcore/app/importexport.py
@@ -101,6 +101,7 @@ def export_artifacts(export, artifacts):
     """
     data = dict(message="Exporting Artifacts", code="export.artifacts", total=len(artifacts))
     with ProgressReport(**data) as pb:
+        pb.BATCH_INTERVAL = 5000
         for artifact in pb.iter(artifacts):
             dest = artifact.file.name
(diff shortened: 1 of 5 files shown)


mdellweg (approved):

ipanova (approved):
can we stick to s/cumulative_hash/global_hash
especially if this is used only in import/export and in export.jsonm file we also have global_hash or else renaming there too so its consistent. Since this would a user visible change, stick to global_hash is my suggestion

dralley:
>can we stick to s/cumulative_hash/global_hash
especially if this is used only in import/export and in export.json file we also have global_hash or else renaming there too so its consistent. Since this would a user visible change, stick to global_hash is my suggestion

Nothing about this is a user-visible change, and I didn't rename the actual variable itself in the export code.  I just though `cumulative_hash` made more sense for the name of the function argument.

ipanova:
my 2c, merge at will

dralley:
The way I think of it is that from the perspective of the utility function, it's just a cumulative hash that you're adding to.  The utility doesn't know anything about the way it's being used so "global" isn't meaningful.  From the perspective of the export code, it's a global hash and is still called as such there.

PR #4460, opened 2023-09-21
Bump django-import-export dependency version for performance reasons

dralley:
3.1.0 contains some performance improvements.

[noissue]


mdellweg (changes requested):
Can you change it so that 2.6.1 is still a valid version?

dralley:
@mdellweg Done

Apologies for adding more commits on top, these were the changes from the other PR which failed to backport automatically so I had to manually fix the cherrypicks, and these PRs are already open.

mdellweg (approved):

PR #4463, opened 2023-09-21
Bump django-import-export dependency version for performance reasons

dralley:
3.1.0 contains some performance improvements.

[noissue]


requirements.txt
```diff
@@ -10,7 +10,7 @@ Django~=3.2.16  # LTS version, switch only if we have a compelling reason to
 django-currentuser>=0.5.3,<=0.5.3
 django-filter>=22.1,<=22.1
 django-guid>=3.3,<=3.3.0
-django-import-export>=2.9,<3.1.0
+django-import-export>=3.1.0,<3.4.0
```
mdellweg, line 13: Changing the lower bound sounds like a breaking change to me. We should not do that in a z release.

mdellweg (approved):

PR #4465, opened 2023-09-21
Refactor worker helper functions

mdellweg:
Separating the affiliation of functions between `pulpcore.tasking.task` and `pulpcore.tasking.utils` (now `_utils`) to reflect which of those are used to interact with tasks from the app level and which are internal plumbing of the task worker.


dralley:
Is this purely moving things around and changing function names or are there code changes also?  It's difficult to tell when code is being shifted between files.

dralley (commented):
IF it's purely moving code around and changing names, all of this looks fine.  If there's anything more there, then I haven't reviewed well enough to call this LTGM

mdellweg:
Yes, it is only moving code. I updated the description.

dralley (approved):

dkliban (approved):

PR #4468, opened 2023-09-22

dkliban:

closes: #3173
closes: #3459

### DIFF
--- a/pulpcore/content/__init__.py
+++ b/pulpcore/content/__init__.py
@@ -94,6 +94,7 @@ async def server(*args, **kwargs):
     path_prefix = settings.CONTENT_PATH_PREFIX
     if settings.DOMAIN_ENABLED:
         path_prefix = path_prefix + "{pulp_domain}/"
+        app.add_routes([web.get(path_prefix[:-1], Handler().list_distributions)])
     app.add_routes([web.get(path_prefix, Handler().list_distributions)])
     app.add_routes([web.get(path_prefix + "{path:.+}", Handler().stream_content)])
     app.cleanup_ctx.append(_heartbeat_ctx)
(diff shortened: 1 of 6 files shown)


dkliban:
I added tests to pulp_file https://github.com/pulp/pulp_file/pull/1048/files

daviddavis:
The code here looks good to me but when I tested this out and it seems like distro base paths work but nested paths don't for some reason:

```
pulp rpm repository create --name rpmtest
pulp rpm distribution create --name rpmtest --base-path rpmtest --repository rpmtest
pulp rpm publication create --repository rpmtest
http :24816/pulp/content/rpmtest (good, returns 301 to rpmtest/)
http :24816/pulp/content/rpmtest/repodata (returns 404)
```

daviddavis (approved):

pulpcore/content/__init__.py
```diff
@@ -94,6 +94,7 @@ async def server(*args, **kwargs):
     path_prefix = settings.CONTENT_PATH_PREFIX
     if settings.DOMAIN_ENABLED:
         path_prefix = path_prefix + "{pulp_domain}/"
+        app.add_routes([web.get(path_prefix[:-1], Handler().list_distributions)])
```
dkliban, line 97: Does this need to be unindented so as to be useful when domains are disabled?

pulpcore/content/handler.py
```diff
@@ -173,6 +174,9 @@ async def list_distributions(self, request):
         """
         domain = get_domain()
 
+        if not request.path.endswith("/"):
+            raise HTTPMovedPermanently(f"{request.path}/")
```
mdellweg, line 178: Do we know that path exists at this point?

dkliban: Yes we do. This is how `/pulp/content` is redirected to `/pulp/content/`.

pulpcore/content/__init__.py
dkliban: No it does not need to be unindented because we handle the non-domain case inside of `Handler.list_distribtutions()`.

mdellweg (approved):

ggainey (approved):

PR #4477, opened 2023-09-25
Remove gzip compression from exports entirely

### LINKED ISSUE #4446: Make export compression useful or get rid of it entirely
see: https://github.com/pulp/pulpcore/issues/4434

We could do any of the following:

* make the compression level configurable for users who really want it, even though it accomplishes little
* get rid of compression entirely, and therefore ditch the monkeypatches 
* try out a more suitable algorithm such as zstd or lz4, which are much faster than e.g. gzip level 1 for similar amounts of compression

This would be a forwards-only change, no backports.

dralley:
closes #4446


mdellweg (approved):

dkliban (approved):

dralley:
I believe a new release of pulp_file might be needed first?

ipanova:
docs needs small update https://github.com/pulp/pulpcore/blob/main/docs/workflows/import-export.rst#definitions

dralley:
@ipanova Updated

ipanova (approved):

PR #4483, opened 2023-09-26
Optimize export_artifacts hot loop

dralley:
Let's not create and delete a temporary directory for every artifact.

[noissue]

### DIFF
--- a/pulpcore/app/importexport.py
+++ b/pulpcore/app/importexport.py
@@ -102,17 +102,21 @@ def export_artifacts(export, artifacts):
     data = dict(message="Exporting Artifacts", code="export.artifacts", total=len(artifacts))
     with ProgressReport(**data) as pb:
         pb.BATCH_INTERVAL = 5000
-        for artifact in pb.iter(artifacts):
-            dest = artifact.file.name
+        with tempfile.TemporaryDirectory(dir=".") as temp_dir:
             if settings.DEFAULT_FILE_STORAGE != "pulpcore.app.models.storage.FileSystem":
-                with tempfile.TemporaryDirectory(dir=".") as temp_dir:
+                for artifact in pb.iter(artifacts):
                     with tempfile.NamedTemporaryFile(dir=temp_dir) as temp_file:
+                        # TODO: this looks like a memory usage threat
+                        # TODO: it's also probably horrificaly slow, going one-by-one over the net
+                        # TODO: probably we could skip the temp file entirely and add
+                        #       artifact.file.read() directly to the tarfile with tarfile.addfile()
                         temp_file.write(artifact.file.read())
                         temp_file.flush()
                         artifact.file.close()
-                        export.tarfile.add(temp_file.name, dest)
+                        export.tarfile.add(temp_file.name, artifact.file.name)
             else:
-                export.tarfile.add(artifact.file.path, dest)
+                for artifact in pb.iter(artifacts):
+                    export.tarfile.add(artifact.file.path, artifact.file.name)
 
     resource = ArtifactResource()
     resource.queryset = artifacts


pulpcore/app/importexport.py
```diff
@@ -102,17 +102,20 @@ def export_artifacts(export, artifacts):
...
-                with tempfile.TemporaryDirectory(dir=".") as temp_dir:
+                for artifact in pb.iter(artifacts):
                     with tempfile.NamedTemporaryFile(dir=temp_dir) as temp_file:
+                        # TODO: this looks like a memory usage threat
+                        # TODO: it's also probably horrificaly slow, going one-by-one over the net
                         temp_file.write(artifact.file.read())
                         temp_file.flush()
                         artifact.file.close()
-                        export.tarfile.add(temp_file.name, dest)
+                        export.tarfile.add(temp_file.name, artifact.file.name)
```
mdellweg, line 115: Any chance we can shove the artifact into the tar without writing it to disk first?

pulpcore/app/importexport.py
```diff
@@ -102,17 +102,20 @@ def export_artifacts(export, artifacts):
     data = dict(message="Exporting Artifacts", code="export.artifacts", total=len(artifacts))
     with ProgressReport(**data) as pb:
         pb.BATCH_INTERVAL = 5000
-        for artifact in pb.iter(artifacts):
-            dest = artifact.file.name
+        with tempfile.TemporaryDirectory(dir=".") as temp_dir:
+            # separate the hot loops since they're likely to end up quite different
```
mdellweg, line 106: I'm not confident i understand this comment as is. And this won't get better over time.
I think, you are describing the kind of refactor you are applying. Can you describe instead why which loop is the outer one?

pulpcore/app/importexport.py
mdellweg: Does this function by any chance consume the artifact.file directly?
https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.addfile

dralley: The description says that it still has to read from the file (which makes sense)

>Add the [TarInfo](https://docs.python.org/3/library/tarfile.html#tarfile.TarInfo) object tarinfo to the archive. If fileobj is given, it should be a [binary file](https://docs.python.org/3/glossary.html#term-binary-file), and tarinfo.size bytes are read from it and added to the archive.

It seems that `.add()` calls `.addfile()` internally because it shows up on my profile.  The additional operations take up 1% of the overall runtime, but it's pretty likely that most of that is opening the file which is a cost we have to pay regardless.

pulpcore/app/importexport.py
dralley: @mdellweg I deleted the comment because it didn't make much sense to be an inline one.

Previously there was only one loop with a switch on the storage backend inside.  Now the storage backend check is lifted outside of the loop and there are two loops, but one per branch so only one of them will be used.

It's a microoptimization from a performance perspective but I think it makes sense from an organizational perspective also.

pulpcore/app/importexport.py
dralley: Oh.  Probably?  It makes sense.  But, that only matters for cloud storage backends, and I don't have one to test with, and I don't know any users that are using it.  I can't spend 3 hours trying to test something that isn't related to the customer cases right now.

I'll leave a comment to that effect, though.  It might be a solid improvement.

dralley: I'm pretty sure we don't have CI coverage for it, the export tests only run in CI for the filesystem backend.

mdellweg (approved):

dkliban (approved):

PR #4491, opened 2023-09-26
[3.22] Backport task profiling functionality

### LINKED ISSUE #4436: TASK_DIAGNOSTICS enabled should automatically profile all tasks
Currently when the TASK_DIAGNOSTICS option is enabled, Pulp creates a directory `/var/tmp/pulp/<task-id>/` and writes a data file w/ memory usage over time for the task.

This is really useful for debugging purposes.  But it would be even more useful if tasks would automatically be profiled and the profile written to the same directory.  Otherwise adding this profiling requires tedious manual code changes.

This also opens the door for easy in-situ debugging by users, who can then easily enable profiling themselves on their own tasks, and analyze them or provide them to developers (us) for analysis without necessarily requiring tedious reproduction of the setup ourselves.

dralley:
closes #4436

### DIFF
--- a/pulpcore/tasking/pulpcore_worker.py
+++ b/pulpcore/tasking/pulpcore_worker.py
@@ -389,13 +389,17 @@ class NewPulpWorker:
             self.shutdown()
 
 
-def write_memory_usage(task_pk):
+def task_diagnostics_dir(task_pk):
+    # create a directory in /var/tmp/pulp/<task_id>/ and return it
     taskdata_dir = VAR_TMP_PULP / str(task_pk)
     taskdata_dir.mkdir(parents=True, exist_ok=True)
-    memory_file_dir = taskdata_dir / "memory.datum"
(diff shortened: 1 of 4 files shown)


pulpcore/tasking/pulpcore_worker.py
```diff
@@ -437,28 +445,45 @@ def _perform_task(task_pk, task_working_dir_rel_path):
...
-            loop = asyncio.get_event_loop()
-            loop.run_until_complete(result)
-
-    except Exception:
-        exc_type, exc, tb = sys.exc_info()
-        task.set_failed(exc, tb)
-        _logger.info(_("Task %s failed (%s)"), task.pk, exc)
-        _logger.info("\n".join(traceback.format_list(traceback.extract_tb(tb))))
+
+    def execute_task(task):
```
dralley, line 449: Minor refactor here

mdellweg (approved):

Issue #4450, opened 2023-09-19
Django `?format` doesn't work correctly in some endpoints


jerabekjiri:
**Version**
```
pip3 list | grep -E "pulpcore|pulp-ansible|galaxy-ng"
galaxy-ng                                4.9.0.dev0
pulp-ansible                             0.19.0
pulpcore                                 3.28.14
```

**Describe the bug**
Pulp incorrectly assumes that `?format=api|json` is a filter parameter instead of the django rendering format type. This issue only occurs in some endpoints.

**To Reproduce**
Visit endpoints 
`/pulp/api/v3/repositories/?format=api` or 
`/pulp/api/v3/repositories/ansible/ansible/?format=json`: 
```
{
  "errors": [
    "Invalid Filter: 'format'"
  ]
}
``` 

`/pulp/api/v3/repositories/?format=test`:
```
{
  "detail": "Not found."
}
```
![Screenshot from 2023-09-18 22-49-54](https://github.com/pulp/pulpcore/assets/19647757/72656ee2-4ef0-46d6-9dc0-1fdffb1730cd)
![Screenshot from 2023-09-18 22-49-41](https://github.com/pulp/pulpcore/assets/19647757/3d8ac7fe-83ed-46dd-9d04-913d524b5889)
![Screenshot from 2023-09-19 18-34-11](https://github.com/pulp/pulpcore/assets/19647757/d075fb38-3ece-4d8f-91ad-12c282e6a7a1)

**Expected behavior**
Should switch between raw format and Django's pretty format, and display the data as usual.

**Additional context**
Discovered in https://github.com/ansible/galaxy_ng/pull/1891

jerabekjiri:
I think this can be fixed by adding `format` to `DEFAULT_FILTERS` in `BaseFilterSet` https://github.com/pulp/pulpcore/blob/bc9ca511dc13fbae7c49b19f1ebe8b527c67484a/pulpcore/filters.py#L372-L382

mdellweg:
That's not even supposed to be a filter. You should user the encoding header to determine the used format. Also i'm unsure we support anything but json anyway.

mdellweg:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

jerabekjiri:
[...]

PR #4506, opened 2023-09-27
Add `format` param in `BaseFilterSet` to fix Django API format

jerabekjiri:
fixes: #4450

### DIFF
--- a/pulpcore/filters.py
+++ b/pulpcore/filters.py
@@ -382,6 +382,7 @@ class BaseFilterSet(filterset.FilterSet):
             "offset",
             "page_size",
             "ordering",
+            "format",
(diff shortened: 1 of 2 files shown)


mdellweg (approved):

pulpcore/filters.py
```diff
@@ -382,6 +382,7 @@ def is_valid(self, *args, **kwargs):
             "offset",
             "page_size",
             "ordering",
+            "format",
```
mdellweg, line 385: I'm still wondering, why this seems to be problematic on _some_ but not _all_ endpoints.

dkliban (approved):

Issue #4511, opened 2023-09-28
Enable social_django urls based on config


jeff350:
**Is your feature request related to a problem? Please describe.**
The instructions for enabling external Oauth/OIDC auth flows currently require modifying of the urls.py file, directly in the code. This is not ideal as the required local changes could cause issues with future upgrades of the pulp software. 

**Describe the solution you'd like**
I propose adding an if statement to include the social_django urls in the pulp core urls.py file so that Oauth/OIDC auth flows such as keycloak can be configured with out having to directly modify code files. 

**Describe alternatives you've considered**
the other option would be for anyone that needs to use keycloak to deal with having to reapply this change to the urls.py file after every upgrade.

**Additional context**
See the relevant pulp documentation here https://docs.pulpproject.org/pulpcore/en/master/nightly/authentication/keycloak.html

PR #4512, opened 2023-09-28
fix: social urls from config

jeff350:
Addresses #4511  by using required config option `SOCIAL_AUTH_URL_NAMESPACE`

### DIFF
--- a/docs/authentication/keycloak.rst
+++ b/docs/authentication/keycloak.rst
@@ -47,25 +47,7 @@ Enable general python social integration with the following steps:
         ...
     ]
 
-3. When using PostgreSQL, it’s recommended to use the built-in JSONB field by defining the settings::
-
-    SOCIAL_AUTH_POSTGRES_JSONFIELD = True
-    SOCIAL_AUTH_JSONFIELD_CUSTOM = 'django.db.models.JSONField'
-
-
-4. In order to use the Keycloak OIDC capabilities, you must add a URL namespace::
-
-    SOCIAL_AUTH_URL_NAMESPACE = 'social'
-
-5. You must add the Keycloak OIDC URLs to the apps url entries::
-
-    urlpatterns = patterns('',
-        ...
-        url('', include('social_django.urls', namespace='social'))
-        ...
-    )
-
-6. Update the context processor that will add backends and associations data to the template context::
+3. Update the context processor that will add backends and associations data to the template context::
 
     TEMPLATES = [
         {
@@ -82,7 +64,7 @@ Enable general python social integration with the following steps:
         }
     ]
 
-7. Define the authentication pipeline for data that will be associated with users::
+4. Define the authentication pipeline for data that will be associated with users::
 
     SOCIAL_AUTH_PIPELINE = (
         'social_core.pipeline.social_auth.social_details',
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -202,6 +202,11 @@ REDIS_PASSWORD = None
 REDIS_SSL = False
 REDIS_SSL_CA_CERTS = None
 
+# Social_django settings
+SOCIAL_AUTH_URL_NAMESPACE = "social"
+SOCIAL_AUTH_POSTGRES_JSONFIELD = True
+SOCIAL_AUTH_JSONFIELD_CUSTOM = "django.db.models.JSONField"
+
 # https://docs.djangoproject.com/en/3.2/ref/settings/#logging and
 # https://docs.python.org/3/library/logging.config.html
 LOGGING = {
--- a/pulpcore/app/urls.py
+++ b/pulpcore/app/urls.py
@@ -203,6 +203,11 @@ if settings.DOMAIN_ENABLED:
         docs_and_status_no_schema.append(path(str(p.pattern), view, name=name))
     urlpatterns.append(path(API_ROOT, include(docs_and_status_no_schema)))
 
+if "social_django" in settings.INSTALLED_APPS:
+    urlpatterns.append(
+        path("", include("social_django.urls", namespace=settings.SOCIAL_AUTH_URL_NAMESPACE))
+    )
+
 #: The Pulp Platform v3 API router, which can be used to manually register ViewSets with the API.
 root_router = PulpDefaultRouter()
 
--- a/CHANGES/4511.feature
+++ b/CHANGES/4511.feature
@@ -1,0 +1,1 @@
+Include social_django urls based on config option


dkliban (changes requested):
please add a changelog file to the CHANGES directory. I would say that this is a new feature. So it should be called `4511.feature`.

Your commit message needs to also state 'fixes: #4511' or 'closes: #4511'

jeff350:
@dkliban requested updates have been made, please let me know if I need to change all commit messages to 'fixes: #4511'  or if just the latest is sufficient.

dkliban:
@jeff350 please squash the changes into a single commit

jeff350:
@dkliban all commits have been squashed into a single commit, looks like this should be good to go when all tests pass.

dkliban:
@jeff350 The CI is showing the following error:

` AttributeError: 'HookableSettings' object has no attribute 'SOCIAL_AUTH_URL_NAMESPACE'`

You need to ensure that the setting is present before checking the value.

jeff350:
@dkliban I missed that case... I am starting to wonder if checking that specific setting directly may not the best approach. A possible alternative would be to check if social_django is in the list of installed apps, that maybe a cleaner way to do this.

 do you have an option on what would be preferred?

dkliban:
@jeff350 You should check that `django_social` is in the installed apps. If it is installed, you should then check that `SOCIAL_AUTH_URL_NAMESPACE` is set. Then the value of `SOCIAL_AUTH_URL_NAMESPACE` should be used as the value of `namespace` when adding the You also want to make sure that the value of `namespace` in the urlpattern being added.

jeff350:
@dkliban This should be fixed now. I also added the SOCIAL_AUTH_URL_NAMESPACE as a default in the settings.py file as I understand that will always need to be a custom namespace and 'social' is what django_social always recommends to use. 

This also helped to further simplify the docs.

pulpcore/app/urls.py
```diff
@@ -203,6 +203,10 @@ class NoSchema(p.callback.cls):
         docs_and_status_no_schema.append(path(str(p.pattern), view, name=name))
     urlpatterns.append(path(API_ROOT, include(docs_and_status_no_schema)))
 
+if "social_django" in settings.INSTALLED_APPS:
+    if settings.SOCIAL_AUTH_URL_NAMESPACE == "social":
```
dkliban, line 207: I don't believe that we want to only include the urlpattern when the SOCIAL_AUTH_URL_NAMESPACE is 'social'. If 'social_django' is installed we want to include the urlpattern with the `namespace` specified by `SOCIAL_AUTH_URL_NAMESPACE`.  So like this:

```python
if "social_django" in settings.INSTALLED_APPS:
    urlpatterns.append(path("", include("social_django.urls", namespace=settings.SOCIAL_AUTH_URL_NAMESPACE)))
```

jeff350: Change has been mande

mdellweg: Thank you. I was about to request the exact same.

pulpcore/app/settings.py
```diff
@@ -202,6 +202,9 @@
 REDIS_SSL = False
 REDIS_SSL_CA_CERTS = None
 
+# Social_django settings
+SOCIAL_AUTH_URL_NAMESPACE = "social"
+
```
mdellweg, line 207: Please add a section to the settings part of the docs, too.

mdellweg (changes requested):

docs/authentication/keycloak.rst
```diff
@@ -52,20 +52,7 @@ Enable general python social integration with the following steps:
     SOCIAL_AUTH_POSTGRES_JSONFIELD = True
     SOCIAL_AUTH_JSONFIELD_CUSTOM = 'django.db.models.JSONField'
 
-
-4. In order to use the Keycloak OIDC capabilities, you must add a URL namespace::
-
-    SOCIAL_AUTH_URL_NAMESPACE = 'social'
-
```
mdellweg, line 59: By now providing a default value, can we still keep this step as an optional one?

jeff350: My understanding is that this is not a setting anyone will want to change. Reading the Django docs (https://docs.djangoproject.com/en/4.2/topics/http/urls/#url-namespaces) the namespace is designed for reserving url patterns. I would equate django_social having the namespace of 'social' to their example of the django admin having a namespace of 'admin'.

additionally  django_social also recommends setting this to 'social' if a custom namespace is going to be used.  https://python-social-auth.readthedocs.io/en/latest/configuration/django.html#urls-entries

pulpcore/app/settings.py
jeff350: If this is not something that is intended for users to change does it need a section in the settings documentation? looking at the documentation it seems that the settings there are ones that are intended fore users to change as needed. 

See for context on why this does not need to be a user configurable setting https://github.com/pulp/pulpcore/pull/4512#discussion_r1341971011

docs/authentication/keycloak.rst
mdellweg: > additionally django_social also recommends setting this to 'social' if a custom namespace is going to be used. https://python-social-auth.readthedocs.io/en/latest/configuration/django.html#urls-entries

Actually, i read it the other way around. You only need to set it if it's _not_ the default (and their example is not helping much). Having said that, the url-namespace is not changing the included paths. So, I think we are fine always using "social".

docs/authentication/keycloak.rst
```diff
@@ -52,20 +52,7 @@ Enable general python social integration with the following steps:
     SOCIAL_AUTH_POSTGRES_JSONFIELD = True
     SOCIAL_AUTH_JSONFIELD_CUSTOM = 'django.db.models.JSONField'
```
mdellweg, line 53: We are always using Postgres. You can move these settings too and remove the step from the docs.

jeff350: Requested changes have been made

pulpcore/app/settings.py
```diff
@@ -202,6 +202,11 @@
 REDIS_SSL = False
 REDIS_SSL_CA_CERTS = None
 
+# Social_django settings
+SOCIAL_AUTH_URL_NAMESPACE = "social"
+SOCIAL_AUTH_POSTGRES_JSONFIELD = True
+SOCIAL_AUTH_JSONFIELD_CUSTOM = 'django.db.models.JSONField'
```
dkliban, line 208: black is complaining about use of single quotes. Please use " instead of '.

jeff350: fixed

mdellweg (approved):

dkliban (approved):

PR #4514, opened 2023-09-28
Pin django-storages extra requirements

gerrod3:
The latest django-storages 1.14 adds new fields to S3 config that are not accounted for in the domain serializer, so this requirement should have been pinned from the first introduction of domains. This will need to be backported.


gerrod3:
Using plugin-template: https://github.com/pulp/plugin_template/pull/787

gerrod3:
I've confirmed that the CI now respects the bounds for S3 and Azure by looking at the pip logs, but I can't figure out why the CI still installs django-storages even for the Pulp and Lowerbounds runners.

mdellweg:
> I've confirmed that the CI now respects the bounds for S3 and Azure by looking at the pip logs, but I can't figure out why the CI still installs django-storages even for the Pulp and Lowerbounds runners.

They are installed in the containerfile which is the same for all test scenarios.

gerrod3:
Waiting for the weekly CI update job to run before rebasing now that the plugin template PR has been merged.

lubosmj (approved):

setup.py
```diff
@@ -21,9 +21,9 @@
     install_requires=requirements,
     extras_require={
         "sftp": ["django-storages[sftp]"],
-        "s3": ["django-storages[boto3]"],
+        "s3": ["django-storages[boto3]<=1.13.2"],
         "google": ["django-storages[google]>=1.13.2"],
```
mdellweg, line 25: Shouldn't we add something here too?

gerrod3: For google you mean? Maybe, I didn't since this change was specific for domains and domains doesn't support google backends yet.

mdellweg: Right. But in general, i think these are as good dependencies as all the others. And we should add the upper bound in the same manner. (Can be a separate PR if you want.)

mdellweg (approved):

mdellweg:
Do you plan to up the dependencies an include the new fields in the domain config? At least please track it with an issue.

gerrod3:
> Do you plan to up the dependencies an include the new fields in the domain config? At least please track it with an issue.

Yes I plan to add the new fields when I will update the dependencies. https://github.com/pulp/pulpcore/issues/4592

PR #4515, opened 2023-09-29
Allow older releases to opt-in to a faster mode of import/export

dralley:
If a user enables the flag `_BACKWARDS_INCOMPATIBLE_FAST_EXPORTS` in the settings, the following will happen:

* Exports won't have any gzip wrapper at all.  This primarily improves imports as the gzip wrapper adds a bunch of internal crc32 checksum calculations that are totally unnecessary.
* The checksums used on export and import will be crc32 instead of sha256.  This makes a difference at scale when multiple terabytes are being processed.  
* Both kinds of exports are now supported on the import side transparently

When not bottlenecked on the database, these changes result in a 25-30% improvement in my testing of RHEL9 BaseOS + Appstream (48gb total).

### DIFF
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -249,6 +249,12 @@ ALLOWED_IMPORT_PATHS = []
 
 ALLOWED_EXPORT_PATHS = []
 
+# Enables a couple of performance-enhancing behaviors on export which result in an
+# export format which is incompatible with previous z-stream releases of this y-stream.
+# These modified export formats require compatibility on the import side, which should
+# be upgraded to the same level.
+BACKWARDS_INCOMPATIBLE_FAST_EXPORTS = False
+
 # https://docs.pulpproject.org/pulpcore/configuration/settings.html#pulp-cache
 CACHE_ENABLED = False
 CACHE_SETTINGS = {
--- a/pulpcore/app/tasks/export.py
+++ b/pulpcore/app/tasks/export.py
@@ -29,7 +29,7 @@ from pulpcore.app.models import (
 from pulpcore.app.models.content import ContentArtifact
 from pulpcore.app.serializers import PulpExportSerializer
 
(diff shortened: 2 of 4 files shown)


pulpcore/app/settings.py
```diff
@@ -249,6 +249,8 @@
 
 ALLOWED_EXPORT_PATHS = []
 
+_BACKWARDS_INCOMPATIBLE_FAST_EXPORTS = False
```
ipanova, line 252: why underscore? it's a setting

ipanova (approved):

pulpcore/app/settings.py
dralley: Well, I don't feel strongly about it.  My feeling was that it would still be kind of a "secret" / undocumented option that only people working with support directly would use, and only if they were having issues with exports currently.  So the underscore made sense to me.

But I can drop it if there's a desire to drop it or otherwise not treat it that way.

dralley:
The CRC checksum changes aren't in main branch yet, that will be a separate PR.

pulpcore/app/settings.py
ipanova: i'd say recently added setting `IMPORT_WORKERS_PERCENT` is not documented too and is mostly created for  specific customer needs. 
I would be in favor dropping the _ and leave it undocumented

mdellweg: I agree. The name without the underscore is sufficiently scary.
Also the plan is that we will ignore this setting on newer versions, right?

dralley: I'll drop the underscore.

The setting doesn't currently exist in newer versions and I'm not sure if I want to add it yet.  Possibly we would want to have a discussion whether we can just use CRC32 or something similarly fast universally.  I think Sha256 with hardware support is actually faster, but I don't know that it's enough of a difference to be worth having two options.

PR #4516, opened 2023-10-02
Allow whitespace and comments in db key file

### LINKED ISSUE #4542: Allow whitespace and comments in db key file
It would be useful if the db key file parsed generously around additional whitespace.

Also in order to help guide an admin while rotating keys, we should allow to add comments.

mdellweg:
fixes #4542

### DIFF
--- a/CHANGES/4542.feature
+++ b/CHANGES/4542.feature
@@ -1,0 +1,1 @@
+Parsing of the db encryption key was made resiliant to whitespace and to allow comments.
--- a/pulpcore/app/models/fields.py
+++ b/pulpcore/app/models/fields.py
@@ -1,17 +1,15 @@
 import json
 import logging
 import os
-from gettext import gettext as _
 from functools import lru_cache
+from gettext import gettext as _
 
 from cryptography.fernet import Fernet, MultiFernet
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
-from django.db.models import Lookup, FileField, JSONField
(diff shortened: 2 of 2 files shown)


dralley:
I understand allowing whitespace, but why comments?  This should probably have an issue with additional explanation

mdellweg:
> I understand allowing whitespace, but why comments? This should probably have an issue with additional explanation

Whitespace was triggered by this: https://discourse.pulpproject.org/t/pulp-minimal-3-35-pulp-api-metadata-signing-service-http-1-1-500-internal-server-error/1002/5

Comments were like "why not". And my thinking is: You can specify multiple keys in there (to allow key rotation.) Adding comments which key is supposed to be the new one and must move to the top of the list eventually or even a note when it's supposed to happen may become priceless.

CHANGES/+db_encryption.feature
```diff
@@ -0,0 +1 @@
+Parsing of the db encryption key was made resiliant to whitespace and to allow comments.
```
ipanova, line 1: can you file an issue so the changelog is in the format of XXXX.feature?

ipanova (approved):

CHANGES/+db_encryption.feature
mdellweg: Sure. I wanted to know if we would accept this as a feature...

ipanova: Any step towards making things more resilient is a good step, I think.

ggainey (approved):

PR #4517, opened 2023-10-03
Minor memory + perf export optimizations

dralley:
[noissue]

### DIFF
--- a/pulpcore/app/importexport.py
+++ b/pulpcore/app/importexport.py
@@ -48,15 +48,29 @@ def _write_export(the_tarfile, resource, dest_dir=None):
     with tempfile.NamedTemporaryFile(dir=".", mode="w", encoding="utf8") as temp_file:
         if isinstance(resource.queryset, QuerySet):
             temp_file.write("[")
-            total = resource.queryset.count()
-            for i in range(0, total, EXPORT_BATCH_SIZE):
-                current_batch = i + EXPORT_BATCH_SIZE
-                dataset = resource.export(resource.queryset[i:current_batch])
+
+            def process_batch(batch):
+                dataset = resource.export(batch)
                 # Strip "[" and "]" as we are writing the dataset in batch
                 temp_file.write(dataset.json.lstrip("[").rstrip("]"))
-                if current_batch < total:
+
+            batch = []
+            needs_comma = False
+            for item in resource.queryset.iterator(chunk_size=EXPORT_BATCH_SIZE):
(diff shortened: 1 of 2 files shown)


pulpcore/app/importexport.py
```diff
@@ -48,15 +48,29 @@ def _write_export(the_tarfile, resource, dest_dir=None):
...
+            batch = []
+            needs_comma = False
+            for item in resource.queryset.iterator(chunk_size=EXPORT_BATCH_SIZE):
+                batch.append(item)
+                if needs_comma:
                     # Write "," if not last loop
                     temp_file.write(", ")
+                    needs_comma = False
+
+                if len(batch) >= EXPORT_BATCH_SIZE:
```
dralley, line 66: Is there a more elegant way to express all of this?

dralley:
Saves about 5-8% on time and memory.

pulpcore/app/importexport.py
mdellweg: This is what i constantly come up with when writing data migrations.

pulpcore/app/tasks/export.py
```diff
@@ -504,16 +509,16 @@ def _do_export(pulp_exporter, tar, the_export):
...
-        artifacts.update(vers_artifacts)
+            vers_artifacts = version.artifacts
+        artifact_pks.update(vers_artifacts.values_list("pk", flat=True))
 
-    # export plugin-version-info
-    export_versions(the_export, plugin_version_info)
     # Export the top-level entities (artifacts and repositories)
     # Note: we've already handled "what about incrementals" when building the 'artifacts' list
-    export_artifacts(the_export, list(artifacts))
+    export_artifacts(the_export, Artifact.objects.filter(pk__in=artifact_pks))
```
mdellweg, line 519: Is this qs the new memory hog now?

mdellweg (approved):
This looks good to me. Hope we have good test coverage here.

ipanova (approved):

pulpcore/app/tasks/export.py
dralley: I don't believe so.  It didn't seem to move the needle much when I executed it in a shell.  It's lazily evaluated in any case so it's not using any memory here.

dralley:
>Hope we have good test coverage here.

I don't know that it's amazingly thorough but I'm heartened by the fact that it has caught things in the past.

PR #4526, opened 2023-10-04
Bring CRC32 checksums to mainline import/export

dralley:
crc32 is much faster to calculate when hardware acceleration for sha256 isn't present


dralley:
@bmbouter @x9c4 The question is, what should be the default?

* CRC32 is probably faster for most users
* SHA256 might be faster, if you have sufficiently modern hardware
* SHA256 is less problematic to justify, even though this isn't a security sensitive context

ipanova:
> @bmbouter @mdellweg The question is, what should be the default?

In this particular workflow we are using the hasher to check data integrity, like bitrot, not prevent man-in-the-midle attacks.
And within one export/import tar I think we'd have really slim chances to have created 2 files with same hash, so using stronger hash might not be the selling point in this case.

mdellweg:
> * SHA256 might be faster, if you have sufficiently modern hardware

I don't see that. There is a __mm_crc32_u64 intrinsic around since SSE4

> * SHA256 is less problematic to justify, even though this isn't a security sensitive context

As you say. There is no security context. This brings me to the question: What do we actually calculate that checksum for? We assure that all the chunks are valid with sha256 already. We have the toc so we don't mix them up. What can still go wrong that justifies calculating that checksum at all?

dralley:
>I don't see that. There is a __mm_crc32_u64 intrinsic around since SSE4

Here's a test on my Zen 3 desktop CPU.  No idea how well it performs on Intel CPUs that support it.


```
In [1]: import hashlib, blake3, zlib

In [2]: 
   ...: class Crc32Hasher:
   ...:     """Wrapper to make the CRC32 implementation act like a standard hashlib hasher"""
   ...: 
   ...:     def __init__(self):
   ...:         self.hashval = 0
   ...: 
   ...:     def update(self, data):
   ...:         self.hashval = zlib.crc32(data, self.hashval)
   ...: 
   ...:     def digest(self):
   ...:         return str(self.hashval)
   ...: 
   ...:     def hexdigest(self):
   ...:         return hex(self.hashval)[2:]
   ...: 
   ...: 
   ...: 

In [3]: def compute_file_hash(filename, hasher=None, cumulative_hash=None, blocksize=8192):
   ...:     if hasher is None:
   ...:         hasher = hashlib.sha256()
   ...:     with open(filename, "rb") as f:
   ...:         # Read and update hash string value in blocks of 8K
   ...:         while chunk := f.read(blocksize):
   ...:             hasher.update(chunk)
   ...:             if cumulative_hash:
   ...:                 cumulative_hash.update(chunk)
   ...:         return hasher.hexdigest()
   ...: 

In [4]: file = "Muse - Live At Rome Olympic Stadium (2013) 1080p DTS-AC3 x264 BluRay -Shiv@.mkv"

In [5]: %time compute_file_hash(file, hasher=Crc32Hasher())
CPU times: user 3.97 s, sys: 3.72 s, total: 7.69 s
Wall time: 16.2 s
Out[5]: '2dce2244'

In [7]: %time compute_file_hash(file, hasher=hashlib.sha256())
CPU times: user 4.86 s, sys: 1.78 s, total: 6.64 s
Wall time: 7.86 s
Out[7]: 'c59ba46f85164e53dd091cdcf1d09741e2123ce8789a0cc922290164c4d360a9'

In [8]: %time compute_file_hash(file, hasher=blake3.blake3())
CPU times: user 2.91 s, sys: 1.37 s, total: 4.28 s
Wall time: 4.56 s
Out[8]: 'c049c77fda2ecb75f425f816fe9c530ff39c94c47d1a0e7492b33ee23eeb2da0'
```

At least on some CPUs SHA256 is accelerated better than CRC32.  Or maybe zlib isn't using the SSE acceleration for CRC32 at all, who knows.

dralley:
>  As you say. There is no security context. This brings me to the question: What do we actually calculate that checksum for? We assure that all the chunks are valid with sha256 already. We have the toc so we don't mix them up. What can still go wrong that justifies calculating that checksum at all?

Well, this PR  replaces both the per-chunk checksum and the whole-export-file checksum.  I assume the reason it's done both per-chunk and after reconstruction is that the chunks get deleted during reconstruction, so if one of them is invalid you would need to completely start over as you'll be left with only an invalid import archive with no way to back up.

It's not a likely enough to be especially concerned about, so in theory we could drop the chunk-checks, but since they can be panellized it probably isn't worth it?

mdellweg:
> > As you say. There is no security context. This brings me to the question: What do we actually calculate that checksum for? We assure that all the chunks are valid with sha256 already. We have the toc so we don't mix them up. What can still go wrong that justifies calculating that checksum at all?
> 
> Well, this PR replaces both the per-chunk checksum and the whole-export-file checksum. I assume the reason it's done both per-chunk and after reconstruction is that the chunks get deleted during reconstruction, so if one of them is invalid you would need to entirely redo the transfer as the chunks will have been destroyed.
> 
> It's not a likely enough to be especially concerned about, so in theory we could drop the chunk-checks, but since they can be panellized it probably isn't worth it?

My point is: After we secured the validity of all chunks and know that we put them in the right order because we have the toc, is there any point in doing the whole file checksum?

dralley:
>My point is: After we secured the validity of all chunks and know that we put them in the right order because we have the toc, is there any point in doing the whole file checksum?

To ensure that the reconstruction worked properly?  It's still copying a large amount of data, it's possible for something to  happen in-flight.

We could allow it to be disabled for customers that are willing to accept that risk for a bit more speed perhaps

mdellweg:
> To ensure that the reconstruction worked properly? It's still copying a large amount of data, it's possible for something to happen in-flight.

I think, when we expect data corruption on that level, the file can just as well corrupt between this check and actually importing the data. And at that point the import itself will fail and report so. OTOH do you expect a file you just wrote to disk a second ago to contain something else than what you put there? I'd say once we established the validity of all the chunks (and that we also have them all) another layer of checksum is not helping much.

dralley:
@mdellweg Let's defer discussion about skipping checksum verification entirely.  I take your position as being in favor of defaulting to CRC32 only?

mdellweg:
> @mdellweg Let's defer discussion about skipping any checksum verification entirely. I take your position as being in favor of defaulting to CRC32 only?

I would do sha256 on the chunks (in parallel) and crc32 or nothing on the whole file.

dralley:
Why sha2 instead of crc32 for chunks, if there's no security issue?

ipanova:
@mdellweg  even if having the map in the toc file, it does not give you the guarantee that the reassemble happended the way it should, so we still need to checksum whole file

mdellweg:
The chunks have to be handled outside of pulp. So i see some value in making sure they were not modified. OTOH if the reassemble failed, it's pulps fault entirely. And i see no point in assuming that our algorithm to concatenate files may be broken at times.
Also what are we going to do with the information that the file is corrupted. The chunks have gone, the import will fail the admin has to start over from the beginning.
Do you expect that other processes will mess with your file after you wrote it? Then you have other problems to solve on your server. This checksum is not helping you.

pulpcore/app/settings.py
```diff
@@ -258,6 +258,12 @@
 
 ALLOWED_EXPORT_PATHS = []
 
+# Enables a couple of performance-enhancing behaviors on export which result in an
+# export format which is incompatible with previous z-stream releases of this y-stream.
+# These modified export formats require compatibility on the import side, which should
+# be upgraded to the same level.
+BACKWARDS_INCOMPATIBLE_FAST_EXPORTS = False
+
```
mdellweg, line 266: Import export is not compatible between y versions. We should not need this option here.
Let's just implement the format we think is the best. (That way we don't need to think about covering anything special with the test matrix.)

mdellweg (commented):
I think, the chunk checksums really help the admin to fix a file before we destroy it just to realize that the import will never digest it.
I don't see I can be convinced otherwise, but I also don't want to block any work in improving the performance here. Since this discussion stopped being productive some time ago, I will not continue with it.
If really needed I can do a review based on code quality and not breaking things obviously.

pulpcore/app/settings.py
ipanova: This breaks compatibility between the Z streams, I think

ipanova (approved):
I am putting my approval here because I am quite flexible with what will be the final decision. Whether sha256 will stay as default or not and whether we will calculate hash of final file in  case of chunked export or not. I think it's important to keep chunks hash though.

And my preference would be to switch to crc for import/export.

pulpcore/app/settings.py
mdellweg: Yes, exactly, but this is the thing in the main branch that will be import/export from the next Y-release on. There is no claim for a backwards compatibility, so we can just change the way exports look here. The option was only designed to make a version of this work backportable. It should not be here.

ipanova: @mdellweg  you are right actually

dralley: @ggainey Are you comfortable with CRC32 everywhere, or do you think we ought to at least keep sha256 as an option?

ggainey: Is CRC32 available at all on FIPS machines? It may not be, which means we have to be able to handle both it and one of the secure hashes (SHAxxx) (not SHA1, ofc :) )

Also, not a fan of "rip out options just to make the code marginally cleaner" unless we have a real case to do so.

mdellweg:
> a bad disk-write results in a single Artifact with bogus bits might pass import

A bad disk write, an alpha decay in your solid state disc, these things can happen at any time. And we do not verify the artifacts we save to the very same hardware either. The conceptual difference between assembling that file and transferring the individual chunks from one geolocation to another, to me is that we allow for human error in the latter.
Just my 2c.
And if it comes to that question, I'm in favor of a single file format per y-release.

dralley:
With the numbers we currently have available, I don't think ditching the current checksum checks we have is all that useful, or worth it.

ipanova:
@dralley  I understand what @mdellweg  says it is not about how much we can save but that it is a generic unnecessary overhead to calculate checksum after putting chunks together.

mdellweg:
To add even more to the controversy: What if we do not even assemble that large file?
https://github.com/pulp/pulpcore/pull/4541

A step we don't do is a step that cannot fail. Also to delete the users chunks to tell them afterwards that something went wrong and they need to fetch the files again from somewhere, is rude in my opinion.

dralley:
I would really like to just merge this as-is.  We can discuss further improvements separately [0].  I agree that your approach is worth a shot, but I don't want to cram everything into one PR.

One way or another we do want to calculate a checksum somewhere, and that checksum will probably be CRC32 in some cases.

[0] I'm open to changing the name of the setting, though.  Or using no setting at all, just defaulting to CRC32.  Whatever the consensus decides is the best idea.

mdellweg:
I'm also in favor of not introducing a setting / providing and consuming only a single export format.

dralley:
@mdellweg Updated

pulpcore/app/tasks/importer.py
```diff
@@ -303,6 +304,12 @@ def pulp_import(importer_pk, path, toc, create_repositories):
             created or not.
     """
 
+    def get_hasher(toc):
+        if "checksum_type" in toc["meta"] and toc["meta"]["checksum_type"] == "crc32":
```
dralley, line 308: I figured on the consumption side there's not *harm* in leaving support for sha256?

mdellweg: Can this condition ever evaluate to false? I think we have a version check somewhere. It may make this check superfluous.

mdellweg (approved):

PR #4533, opened 2023-10-04
Calculate chunk checksums in parallel during import

dralley:


pulpcore/app/tasks/importer.py
```diff
@@ -378,10 +379,17 @@ def verify_chunk_hash(chunk_path, expected_digest):
...
+                    path = os.path.join(base_dir, chunk_filename)
+                    expected_digest = the_toc["files"][chunk_filename]
+                    futures.append(exc.submit(verify_chunk_hash, path, expected_digest))
+
             with ProgressReport(**data) as pb:
-                for chunk in pb.iter(chunks):
-                    chunk_path = os.path.join(base_dir, chunk)
-                    verify_chunk_hash(chunk_path, the_toc["files"][chunk])
+                for chunk in pb.iter(concurrent.futures.as_completed(futures)):
+                    pass  # just let the progress bar do its thing
```
mdellweg, line 392: Nice use of the progress bar.

ipanova (approved):

ggainey (approved):
Since we stream hash-computations, this looks like it'll improve performance without noticeable memory impact (ie, instead of 1 8K block at a time, we're evaluating N 8K blocks). Looks good to me!

dralley:
I'm doing some more testing and honestly, it's not clear that it's worth it.  Not exactly sure why.  Maybe the page cache is being fought over.

PR #4534, opened 2023-10-05
Add a HeaderContentGuard.

### LINKED ISSUE #4518: As a user, I have a content guard that can authorize based on the contents of a base64 encoded json value in a header
Pulp users can perform authorization in their webserver, and then have the webserver send a header name and value to pulpcore-api and pulpcore-content. This header would be base64 encodedjson structure. The value to authorize on is some subpart of the json structure. [Here's an example](https://redhatinsights.github.io/insights-results-aggregator/authentication.html) of a structure like this (which we can also use as test data).

Have a new content guard created which takes three parameters `header_name` and `json_accessor`, and `value`. For [this example](https://redhatinsights.github.io/insights-results-aggregator/authentication.html) you could create an instance of the content guard with `header_name='x-rh-identity'`, `json_accessor='.identity.internal.org_id'`, `value='3340851'`. Then any request with that header name and value set will be authorized.

For `json_accessor` I'm proposing we use [the `jq` syntax](https://jqlang.github.io/jq/manual/#basic-filters).

None

decko:
Closes #4518


pulpcore/app/models/publication.py
```diff
@@ -456,6 +459,59 @@ class Meta:
...
+
+        if header_value != self.value:
+            raise PermissionError("Access not allowed - Invalid value.")
+
+        return
+
+    class Meta:
+        default_related_name = "%(app_label)s_%(model_name)s"
+        permissions = (
+            ("download_headercontentguard", "Can Download Content"),
```
dkliban, line 507: you don't need this permission. this is only needed for the rbac content guard.

pulpcore/app/models/publication.py
```diff
@@ -456,6 +460,56 @@ class Meta:
...
+    def permit(self, request):
+        header_content = request.headers.get(self.header_name)
+        if not header_content:
+            raise PermissionError("Access not allowed. Header not found.")
+
+        try:
+            header_decoded_content = b64decode(header_content)
+        except Base64DecodeError:
+            raise PermissionError(
+                "Access not allowed - Header value not encoded in a proper format."
```
bmbouter, line 483:
```suggestion
                "Access not allowed - Header is not Base64 encoded."
```

bmbouter:
This is looking good!

pulpcore/tests/unit/test_content_guard.py
```diff
@@ -73,3 +75,89 @@ def test_preauthenticate_urls():
     request.query = {"expires": str(int(expires) + 1), "validate_token": ":".join((salt, digest2))}
     with pytest.raises(PermissionError):
         content_guard.permit(request)
+
+
+def test_header_content_guard(db):
+    """Test HeaderContentGuard to protect content based on a specific header."""
+
+    # header_content = {
```
dkliban, line 83: Did you mean to leave this all commented out?

dkliban (approved):
Thank you!

pulpcore/app/models/publication.py
```diff
@@ -1,6 +1,10 @@
 import hashlib
+import json
+import jq
 import os
 import re
+from base64 import b64decode
+from binascii import Error as Base64DecodeError
```
mdellweg, line 7: I wonder if this could be imported from base64 instead?

pulpcore/app/models/publication.py
```diff
@@ -456,6 +460,54 @@ class Meta:
...
+            raise PermissionError("Access not allowed - Invalid value.")
+
+        return
+
+    class Meta:
+        default_related_name = "%(app_label)s_%(model_name)s"
+        permissions = (
+            (
+                "manage_roles_headercontentguard",
+                "Can manage role assignments on RBAC content guard",
```
mdellweg, line 506:
```suggestion
                "Can manage role assignments on header content guard",
```

pulpcore/app/serializers/publication.py
```diff
@@ -122,6 +122,22 @@ class Meta(ContentGuardSerializer.Meta):
         model = models.ContentRedirectContentGuard
 
 
+class HeaderContentGuardSerializer(ContentGuardSerializer, GetOrCreateSerializerMixin):
+    """
+    A serializer for HeaderContentGuard.
+    """
+
+    header_name = serializers.CharField(help_text=_("The header that the guard will check on."))
+    value = serializers.CharField(help_text=_("The valued that is used to check against."))
```
mdellweg, line 131:
```suggestion
    value = serializers.CharField(help_text=_("The value used to check against."))
```

pulpcore/app/serializers/publication.py
```diff
@@ -122,6 +122,22 @@ class Meta(ContentGuardSerializer.Meta):
         model = models.ContentRedirectContentGuard
 
 
+class HeaderContentGuardSerializer(ContentGuardSerializer, GetOrCreateSerializerMixin):
+    """
+    A serializer for HeaderContentGuard.
+    """
+
+    header_name = serializers.CharField(help_text=_("The header that the guard will check on."))
```
mdellweg, line 130:
```suggestion
    header_name = serializers.CharField(help_text=_("The header the guard will check on."))
```

pulpcore/app/serializers/publication.py
```diff
@@ -122,6 +122,22 @@ class Meta(ContentGuardSerializer.Meta):
...
+    value = serializers.CharField(help_text=_("The valued that is used to check against."))
+    json_accessor = serializers.CharField(
+        help_text=_("A JQ syntax compatible accessor."), allow_null=True, required=False
+    )
```
mdellweg, line 134: Should we call this a jq-filter? Because we filter the value before comparing it.

CHANGES/4518.feature
```diff
@@ -0,0 +1 @@
+Add a HeaderContentGuard to allow a specific header and its content to guard the content app.
```
mdellweg, line 1:
```suggestion
Added a ``HeaderContentGuard`` to allow a content access based on a value provided by a specific header. The header can optionally be filtered through a ``jq`` expression.
```

pulpcore/app/models/publication.py
```diff
@@ -456,6 +460,54 @@ class Meta:
...
+    TYPE = "header"
+
+    header_name = models.TextField()
+    json_accessor = models.TextField(null=True)
+    value = models.TextField()
+
+    def permit(self, request):
+        header_content = request.headers.get(self.header_name)
+        if not header_content:
+            raise PermissionError("Access not allowed. Header not found.")
```
mdellweg, line 477: We should just access deny here but log (debug?) a bit more verbose.
```suggestion
            raise PermissionError("Access denied.")
            log.debug(_("Access denied. Header {header_name} not found in request.").format(header_name))
```
The same with all raise below.

mdellweg (changes requested):
jq can filter regex, so there's no need to add that on the value. :+1:

pulpcore/app/models/publication.py
decko: The `base64` module doesn't define any special exception and uses the `binascii` under the hood to decode base64 values.

pulpcore/app/serializers/publication.py
mdellweg: Sure, but i don't think the issue tells us how to name things. But the name chosen will stick with us forever.

decko: Oh, not the issue issue. But the card/ticket/issue. #4518

mdellweg: I get, that the issue introduced that name, but i see that more of a suggestion / guideline. And `json_accessor` does not quite express the concept to me.

dkliban: The jq documentation refers to this string as a filter[0]. Let's call it a `jq_filter`.

[0] https://jqlang.github.io/jq/tutorial/

decko: `jq_filter` then.

CHANGES/4518.feature
```diff
@@ -0,0 +1 @@
+Added a ``HeaderContentGuard`` to allow a content access based on a valued provided by a specific header. The header can optionally be filtered through a ``jq`` expression.
```
bmbouter, line 1: Can this be wrapped to one line. Sorry for the nitpick.

pulpcore/app/models/publication.py
```diff
@@ -456,6 +466,62 @@ class Meta:
...
+
+    def permit(self, request):
+        header_content = request.headers.get(self.header_name)
+        if not header_content:
+            _logger.debug(
+                _("Access not allowed. Header {header_name} not found.").format(
+                    header_name=self.header_name
+                )
+            )
+            raise PermissionError("Access denied.")
```
bmbouter, line 488: What about wrapping all the `PermissionErrors` strings with `gettext`? Generally in Pulp we i8ln the user facing string (but not the logs).

pulpcore/app/serializers/publication.py
```diff
@@ -122,6 +122,29 @@ class Meta(ContentGuardSerializer.Meta):
         model = models.ContentRedirectContentGuard
 
 
+class HeaderContentGuardSerializer(ContentGuardSerializer, GetOrCreateSerializerMixin):
+    """
+    A serializer for HeaderContentGuard.
+    """
+
+    header_name = serializers.CharField(help_text=_("The header the guard will check on."))
```
bmbouter, line 130:
```suggestion
    header_name = serializers.CharField(help_text=_("The header name the guard will check on."))
```

pulpcore/app/serializers/publication.py
```diff
@@ -122,6 +122,29 @@ class Meta(ContentGuardSerializer.Meta):
...
+    value = serializers.CharField(help_text=_("The valued used to check against."))
```
bmbouter, line 131:
```suggestion
    value = serializers.CharField(help_text=_("The valued that will authorize the request."))
```

pulpcore/app/serializers/publication.py
```diff
@@ -122,6 +122,29 @@ class Meta(ContentGuardSerializer.Meta):
...
+    jq_filter = serializers.CharField(
+        help_text=_(
+            (
+                "A JQ syntax compatible filter. If jq_filter is not set, then the value will"
+                "not be filtered yet it still try to decode it."
```
bmbouter, line 136:
```suggestion
                "only be Base64 decoded and checked as an explicit string match."
```

bmbouter (approved):
I left a string/language comments. Fix those if you agree. Functionally speaking this is ready to go. Thank you!

CHANGES/4518.feature
ipanova: s/valued/value

pulpcore/app/models/publication.py
ipanova: does it make sense to provide more info instead log.debug directly in the error? i.e `Access denied. Content guard header not provided`. Similarly, other  PermissionError occurrences down in the code

pulpcore/app/models/publication.py
```diff
@@ -456,6 +466,62 @@ class Meta:
...
+class HeaderContentGuard(ContentGuard, AutoAddObjPermsMixin):
+    """
+    Content guard to protect content based on a header value.
+    """
+
+    TYPE = "header"
+
+    header_name = models.TextField()
+    jq_filter = models.TextField(null=True)
+    value = models.TextField()
```
ipanova, line 478: if you use header_name then use header_value, or drop `header` completely

pulpcore/app/viewsets/publication.py
```diff
@@ -355,6 +357,76 @@ class ContentRedirectContentGuardViewSet(ContentGuardViewSet, RolesMixin):
     }
 
 
+class HeaderContentGuardViewSet(ContentGuardViewSet, RolesMixin):
+    """
+    Content guard to protect the content app using a specific header and its content.
```
ipanova, line 362: `Content guard to protect the content app using a specific header.`

pulpcore/app/viewsets/publication.py
```diff
@@ -355,6 +357,76 @@ class ContentRedirectContentGuardViewSet(ContentGuardViewSet, RolesMixin):
...
+    """
+
+    endpoint_name = "header"
```
ipanova, line 365: please use plural

ipanova: maybe it will be even better to be more explicit - header_contentguards

pulpcore/app/models/publication.py
decko: @bmbouter good point.

@ipanova We had gone on that path but then decided to log the details and present only a generic error on the message [here](https://github.com/pulp/pulpcore/pull/4534#discussion_r1361587433)

pulpcore/app/viewsets/publication.py
decko: All the other ContentGuards use the `endpoint_name` in singular and don't have the `_contentguards` sufix. I believe that the complete URL make that sufix redundant.

PR #4540, opened 2023-10-11
More export optimizations and a bugfix

dralley:


pulpcore/plugin/importexport.py
```diff
@@ -63,8 +63,11 @@ class BaseContentResource(QueryModelResource):
     #
     content_mapping = None
 
-    class Meta:
-        exclude = QueryModelResource.Meta.exclude + ("_artifacts", "content", "content_ptr")
-
     def dehydrate_upstream_id(self, content):
         return str(content.pulp_id)
+
+    def dehydrate_pulp_domain(self, content):
```
dralley, line 69: @ggainey Would it be better, to define this on the `QueryModelResource` above, which is inherited by both `BaseContentResource` and `ArtifactResource`?

dralley:
Combined with https://github.com/pulp/pulp_rpm/pull/3278, an export dropped from 466 seconds to 335 seconds, or about 28%.

This is the solution to the mystery of the field fetches mentioned on Matrix.

dralley:
One interesting thing is that at this point, the progress reporting is seemingly taking up 18% of the time despite only saving once every 5 seconds (and I checked that it works as expected, and it is)

Screenshot from 2023-10-11 01-01-58

That really quite strange.  And all that time is just spent waiting on the database to commit.  Maybe all the disk read traffic is preventing timely writes?  The load is enough to visibly slow operations like `git log`.

 Anyway, it claims only about 142 seconds out of 289 were spent scheduled on the CPU, so we're essentially IO bound for exports at this point.

pulpcore/app/modelresource.py
```diff
@@ -39,12 +39,16 @@ def before_import_row(self, row, **kwargs):
             if row[checksum] == "":
                 row[checksum] = None
 
+    def dehydrate_pulp_domain(self, content):
```
ggainey, line 42: Let's move this to QueryModelResource, then it's available to everyone (ie both Content and Artifact resources). We can then remove from https://github.com/pulp/pulp_rpm/pull/3278

pulpcore/app/modelresource.py
```diff
@@ -39,12 +39,16 @@ def before_import_row(self, row, **kwargs):
...
+        return str(content.pulp_domain_id)
+
     class Meta:
         model = Artifact
         exclude = (
             "pulp_id",
             "pulp_created",
             "pulp_last_updated",
+            "timestamp_of_interest",
```
ggainey, line 51: :+1:

pulpcore/app/viewsets/importer.py
```diff
@@ -102,7 +102,9 @@ def create(self, request, importer_pk):
         path = serializer.validated_data.get("path")
         toc = serializer.validated_data.get("toc")
         create_repositories = serializer.validated_data.get("create_repositories")
-        task_group = TaskGroup.objects.create(description=f"Import of {path}")
+        task_group = TaskGroup.objects.create(
+            description="Import of {path}".format(path=path or toc)
```
ggainey, line 106: :+1:

pulpcore/plugin/importexport.py
ggainey: Having it in QMR seems the better choice - I'm a big fan of DRY :)

ggainey (changes requested):

pulpcore/app/modelresource.py
```diff
@@ -112,15 +116,20 @@ def before_import_row(self, row, **kwargs):
...
-                ContentArtifact.objects.filter(content__in=all_content)
-            )
-        return vers_content.order_by("content", "relative_path")
+                content_pks |= set(content_ids)
+
+        return (
+            ContentArtifact.objects.filter(content__in=content_pks)
+            .order_by("content", "relative_path")
+            .select_related("artifact")
+        )
```
dralley, line 129: Special attention might be warranted on this query.  I'm pretty sure it's identical but...

ggainey: I did walk through this "in my head" a number of times. I'm *pretty* confident one ends up with a query that will select the same set of ContentArtifacts as a result. Existing code unions "content from repoversion" and "all ids from mapping-file" and filters by that. Your change gets *the set of all content-IDs from repoversion-content* , and ors-in the content-ids from the mapping-file. At the end of the day, the QS is filtering by "all the IDs in this collection", and the collection is built from the same pieces.

"100% sure" would require tests that run both ways and then specifically compares the output, I guess.

pulpcore/plugin/importexport.py
```diff
@@ -38,6 +38,9 @@ def before_import_row(self, row, **kwargs):
     def set_up_queryset(self):
         return None
 
+    def dehydrate_pulp_domain(self, content):
```
ggainey, line 41: :+1:

ggainey (approved):

pulpcore/app/modelresource.py
```diff
@@ -113,15 +113,20 @@ def before_import_row(self, row, **kwargs):
...
-            vers_content = vers_content.union(
-                ContentArtifact.objects.filter(content__in=all_content)
-            )
-        return vers_content.order_by("content", "relative_path")
+                content_pks |= set(content_ids)
+
+        return (
+            ContentArtifact.objects.filter(content__in=content_pks)
+            .order_by("content", "relative_path")
+            .select_related("artifact")
```
ipanova, line 125: are you saying we need to prefetch artifact( and generally any other related model) when the FKwidget is using other field than pk?

dralley: Yes, to avoid N+1 issues.  Because FKWidget would otherwise trigger a request for that artifact

docs/plugin_dev/plugin-writer/concepts/subclassing/import-export.rst
```diff
@@ -113,6 +113,14 @@ should be considered::
     # the list signifying the order of imports must also include the repository resource class
     IMPORT_ORDER = [FileContentResource, FileRepositoryResource]
 
+For performance reasons, it is important that care is taken when writing resource definitions. If your model
+has foreign keys that are exported as such (raw UUID key values), you should define a should a custom
+"dehydrate" method for that field to avoid an unnecessary lookup for each instance as seen
+`in this issue <https://github.com/django-import-export/django-import-export/issues/974>`_. Else, if
+foreign keys are exported using some natural key of the referenced row, then the definition of
+`set_up_queryset()` should ensure those references are pre-selected using `select_related()`, otherwise
+an N+1 query scenario is likely.
```
dralley, line 122: @ipanova here's the docs we discussed.  If you're fine with this phrasing, feel free to merge.

mdellweg:
I see backport labels and 4 commits. This is not going to work (a technical limitation in patchback).
I think you either need to merge the commits or split the PR.

dralley:
@mdellweg I'd rather the commits stay separate, so given that the other options is 4 separate PRs with like 24 backport PRs, I think I would prefer to just manually cherrypick to 6 PRs.

docs/plugin_dev/plugin-writer/concepts/subclassing/import-export.rst
lubosmj: @ipanova, do you think we should adhere to these recommendations in pulp-container? https://github.com/pulp/pulp_container/blob/4df9e2ebb20f4a48f6441fe2f11f7f98284c8a1c/pulp_container/app/modelresource.py#L84-L93

ipanova: yes, we will need to dehydrate pulp-domain for the basecontent subclasses and use select_related in the queryset. To avoid importing large models we should apply .only() on the needed fields.

PR #4543, opened 2023-10-11
Import improvement

ipanova:
[noissue]

### DIFF
--- a/pulpcore/app/modelresource.py
+++ b/pulpcore/app/modelresource.py
@@ -86,9 +86,12 @@ class ContentArtifactResource(QueryModelResource):
     artifact = fields.Field(
         column_name="artifact", attribute="artifact", widget=ForeignKeyWidget(Artifact, "sha256")
     )
+    linked_content = {}
 
     def __init__(self, repo_version=None, content_mapping=None):
         self.content_mapping = content_mapping
+        if not ContentArtifactResource.linked_content:
+            ContentArtifactResource.linked_content = self.fetch_linked_content()
         super().__init__(repo_version)
 
     def before_import_row(self, row, **kwargs):
@@ -108,8 +111,7 @@ class ContentArtifactResource(QueryModelResource):
         """
         super().before_import_row(row, **kwargs)
 
-        linked_content = Content.objects.get(upstream_id=row["content"])
-        row["content"] = str(linked_content.pulp_id)
+        row["content"] = self.linked_content[row["content"]]
 
     def set_up_queryset(self):
         vers_content = ContentArtifact.objects.filter(content__in=self.repo_version.content)
@@ -121,6 +123,14 @@ class ContentArtifactResource(QueryModelResource):
                 ContentArtifact.objects.filter(content__in=all_content)
             )
         return vers_content.order_by("content", "relative_path")
+
+    def fetch_linked_content(self):
+        linked_content = {}
+        c_qs = Content.objects.filter(upstream_id__isnull=False).values("upstream_id", "pulp_id")
+        for c in c_qs.iterator():
+            linked_content[str(c["upstream_id"])] = str(c["pulp_id"])
+
+        return linked_content
 
     class Meta:
         model = ContentArtifact
--- a/CHANGES/4564.bugfix
+++ b/CHANGES/4564.bugfix
@@ -1,0 +1,1 @@
+Taught ContentArtifactResource import to cache content results to improve performance.
--- a/pulpcore/app/tasks/importer.py
+++ b/pulpcore/app/tasks/importer.py
@@ -225,6 +225,16 @@ def import_repository_version(
             raise ValidationError(_("No RepositoryVersion found for {}").format(rv_name))
 
         rv_path = os.path.join(temp_dir, rv_name)
+
+        # see if we have a Content mapping
+        mapping_path = f"{rv_name}/{CONTENT_MAPPING_FILE}"
+        mapping = {}
+        with tarfile.open(tar_path, "r") as tar:
+            if mapping_path in tar.getnames():
+                tar.extract(mapping_path, path=temp_dir)
+                with open(os.path.join(temp_dir, mapping_path), "r") as mapping_file:
+                    mapping = json.load(mapping_file)
+
         # Content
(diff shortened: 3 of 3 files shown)


ipanova:
original copr task dropped from 870s to 220s

pulpcore/app/modelresource.py
```diff
@@ -122,6 +122,16 @@ def set_up_queryset(self):
             )
         return vers_content.order_by("content", "relative_path")
 
+    def fetch_linked_content(self):
+        linked_content = {}
+        # x = Content.objects.filter(upstream_id__isnull=False).only("upstream_id", "pulp_id").iterator()
+        c_qs = Content.objects.filter(upstream_id__isnull=False).values("upstream_id", "pulp_id")
```
ipanova, line 128: is there any way how to retrieve specifically list of content we're importing?
potentially this query will get heavier with more and more imported data

ipanova:
"started_at": "2023-10-11T15:54:29.178695Z",
  "finished_at": "2023-10-11T16:01:38.054456Z",
memory

ipanova:
After 2nd commit
      "started_at": "2023-10-11T16:09:38.855855Z",
      "finished_at": "2023-10-11T16:13:14.886248Z",
memory

pulpcore/app/modelresource.py
lubosmj: Is it necessary to call this method inside `__init__`? Could we postpone the invocation to `before_import`? Then, the dataset would be available.

dralley:
@ipanova test runs are an hour each so I'm not 100% done yet, but here's my results

RHEL 8 BaseOS (27 Gb)

# Import

### baseline

pulp_import:  883 seconds, 3.2gb mem
import_repository_version: 2003 seconds, 17.4gb mem


### dalley patches

pulp_import: 848 seconds, 2.73gb mem
import_repository_version: 1890 seconds, 16gb mem


When I got to your patches, the first time I didn't see a difference.  That's not so surprising because it relates to ContentArtifacts and my system is much smaller scale, I would expect the COPR box to have different behavior

However I ran it again and it seems to be stuck on something for the past 90 minutes.  I'll have to check it in the morning

ipanova:
**COPR machine (122k repos, 4.3mln Content)**
The Content artifact patch helped with execution time.
High memory consumption was due to enabled profiling.

**Tar of rhel6-kickstart (4gb)**

Baseline.import_repository_version:
memory-no-profiling

CA artifact patch.import_repository_version::
memory-1st-patch

install pyinstrument .import_repository_version:
memory-profiling

2nd run with pyinstrument.import_repository_version:
memory-4th-run
uninstall pyinstrument.import_repository_version:
memory-back-to-no-profiling

no-profiling.pulp_import
memory-no-profiling-import

profiling.pulp_import
memory-profiling-import

pulpcore/app/modelresource.py
ipanova: Dumping info after having a chat with @ggainey : 
1. Apparently it is not that easy to collect upstream_id from Content because we export only detail models and each has it's own json file. It would be pain and error prone to parse each of those.
2.  This feature https://github.com/pulp/pulpcore/issues/4549  will address the concern of the query getting bigger because we will create a separate json file that will store only upstream_ids which we could use in this query during filter()
3. @ggainey to address my concern what is worse a) querying db  X times with `linked_content = Content.objects.get(upstream_id=row["content"])` or iterating through a query once  that would contain all content that was ever exported, I have changed my code in the following way on COPR machine (4.3mln content)
```
def fetch_linked_content(self):
        linked_content = {}
        x = Content.objects.filter().values('upstream_id', 'pulp_id')
        for i in x.iterator():
            if i['upstream_id']:
                linked_content[str(i['upstream_id'])]=str(i['pulp_id'])
        return linked_content
```
And here are the results:

memory-values-iterator

The concusion is that iterating once trough 4.3 mln results worth it

dralley:
@ipanova Could you file an issue and attach it

pulpcore/app/modelresource.py
```diff
@@ -108,8 +111,7 @@ def before_import_row(self, row, **kwargs):
         """
         super().before_import_row(row, **kwargs)
 
-        linked_content = Content.objects.get(upstream_id=row["content"])
-        row["content"] = str(linked_content.pulp_id)
+        row["content"] = self.linked_content[row["content"]]
```
dralley, line 114: I'm perfectly fine with this solution, we will probably never have so much content that storing the IDs in a dict will use an unreasonable amount of memory.  Going forwards, are there any queries that would benefit from indexing `upstream_id` (like line 129) or would the benefit be minimal / rounding error?

dralley (approved):

pulpcore/app/modelresource.py
ipanova: In [2]: from uuid import uuid4

In [3]: x={str(uuid4()):str(uuid4()) for i in range(1_000_000)}
In [5]: from sys import getsizeof

In [6]: getsizeof(x)
Out[6]: 41943136



1mln content it would be ~40mb

dralley: I get that, not concerned about memory.  I'm just thinking about whether we should add an index for other queries, or if it's not that important.

ipanova: So in core i found just 2 queries( this one and one in mapping logic) but in plugins at least in pulp-rpm i see more of them.

ipanova: I think it is worth adding the index, because the upstream_id reference will be used for every content which we cannot map in nay other way than upstream_id-->pulp_id

ggainey: :+1:

dralley: Granted I think most of those queries are called only a small number of times.  There's only one DistributionTree and a few variants / addons per repo.

ipanova: That's true, however we cannot control plugins..this is what i was trying to say

ggainey (approved):

dralley:
Github CI has been unreliable today

PR #4588, opened 2023-10-19
Replace the naya library with json_stream for better import performance

### LINKED ISSUE #4585: Json parsing library used for imports "naya" is quite slow
**Describe the bug**

In a number of different profiles of import tasks on large and small systems, the "naya" json parsing library is responsible for about 2/3rds of the execution time during the repository version import tasks.

The reason we use Naya is that it has an API close to what we needed at the time, and its pure-Python nature made it easy to package for immediate distribution to resolve https://github.com/pulp/pulpcore/issues/2307

It was the right choice at the time.  Going forwards we could probably find something better.  Alternatively, we could attempt to optimize Naya to better suit our needs.  Currently Naya processes one character at a time and has a huge tree of if statements:

https://github.com/danielyule/naya/blob/master/naya/json.py#L54

There's a limit to what could be achieved here (it's Python after all, and faster parsers including the stdlib ones are written in native code) but it's probably possible to reduce the overhead by restructuring that code to require less conditional checks per character.

**To Reproduce**

Perform an import with pyinstrument profiling enabled, observe that Naya is responsible for a huge portion of the runtime.

I also manually reverted the PR for  https://github.com/pulp/pulpcore/issues/2307 and found that the execution time dropped from 1533 seconds to 567 seconds.  We cannot simply do that as bulk processing like this has the potential to use too much memory during import - on my system this inflated from 350mb memory used to 4.2gb memory used for one task.  More than one such task executing, or the use of larger repos such as container repos, could blow out the 20gb minimum requirement we'd like to keep well within if possible.

But it demonstrates that a library with lower overhead while maintaining the same requirements would yield significant benefits.

dralley:
Dramatically less overhead - importing a RHEL8 baseos repository version drops from 1553 seconds to 554 seconds.

closes #4585


dralley:
`json_stream` is an actively maintained fork of `naya` with a number of additions such as a Rust tokenizer backend which performs much better than the original Python one.  The Python fallback, should it be needed it for some reason, is basically the same as the Naya one.

dralley:
Peak memory: 348mb

mdellweg (approved):
This sounds reasonable. Also I believe the tests cover this area quite well.

ggainey (approved):
Concur w/ mdellweg's comment

dralley:
Yes I don't plan on back porting either of these PRs, I don't think it will be necessary and as you mentioned would present a pain for the build team.

PR #4596, opened 2023-10-19
Make export version check more resilient

mdellweg:
[noissue]

This came up as a need for the pulp_file merger.


ggainey (approved):
A change that removes a var with "maybe_" in its name is a Good Thing.

ggainey:
test-failure is the "S3 times out" problem, unrelated to this change.

dralley (approved):

PR #4618, opened 2023-10-26
Split find_duplicates out from remove_duplicates

### LINKED ISSUE #4607: Expose batch_qs() to the plugin API
If there are no concerns, I would like to have `batch_qs` from https://github.com/pulp/pulpcore/blob/main/pulpcore/app/util.py#L182

Made available to plugins here: https://github.com/pulp/pulpcore/blob/main/pulpcore/plugin/util.py#L15

The reason, is that I need this helper function to fix https://github.com/pulp/pulp_deb/issues/921 and I would prefer not to duplicate it.

See also: https://discourse.pulpproject.org/t/exposing-new-things-to-the-plugin-api/1021

Note: I am fine with duplicating the pulpcore function for any pulp_deb backports, and switching to the pulpcore version with whatever pulpcore release first exposes it. So there is no need to backport the pulpcore change anywhere.

quba42:
closes #4607

pulp_deb has a use case for finding the duplicates, but then needs to perform additional processing to decide whether to actually remove them or not. This change splits a find_duplicates function out from remove_duplicates to facilitate this.

### DIFF
--- a/pulpcore/plugin/repo_version_utils.py
+++ b/pulpcore/plugin/repo_version_utils.py
@@ -13,13 +13,10 @@ _logger = logging.getLogger(__name__)
 __all__ = ["remove_duplicates"]
 
 
-def remove_duplicates(repository_version):
+def find_duplicates(repository_version):
     """
-    Inspect content additions in the `RepositoryVersion` and remove existing repository duplicates.
-
-    This function will inspect the content being added to a repo version and remove any existing
-    content which would collide with the content being added to the repository version. It does not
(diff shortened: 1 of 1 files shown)


quba42:
Note, I am not sure if what I did here is performance neutral. If not, then this is not worth doing, and I will need to find a different way for pulp_deb to avoid duplicating this pulpcore code.

quba42:
Never mind I missed the batching, this is completely broken as soon as you have more than one batch, I think I will abandon this approach!

PR #4643, opened 2023-11-02
Distinguish errors raised during validation in get_or_create

ipanova:
[noissue]

### DIFF
--- a/pulpcore/app/serializers/base.py
+++ b/pulpcore/app/serializers/base.py
@@ -358,6 +358,7 @@ class GetOrCreateSerializerMixin:
 
     @classmethod
     def get_or_create(cls, natural_key, default_values=None):
+        result = None
         try:
             result = cls.Meta.model.objects.get(**natural_key)
         except ObjectDoesNotExist:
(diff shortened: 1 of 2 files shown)


pulpcore/app/serializers/base.py
```diff
@@ -369,7 +369,19 @@ def get_or_create(cls, natural_key, default_values=None):
             try:
                 serializer.is_valid(raise_exception=True)
                 result = serializer.create(serializer.validated_data)
```
mdellweg, line 371: I think you can map these exception to these two lines individually.

mdellweg (approved):

pulpcore/app/serializers/base.py
```diff
@@ -368,8 +368,21 @@ def get_or_create(cls, natural_key, default_values=None):
...
+                if error_codes.intersection(all_codes):
+                    # validation failed because field is invalid, missing
+                    raise e
+                # recover from a race condition, where another thread just created the object
+                # validation failed with 400 'unique' error code
+                result = cls.Meta.model.objects.get(**natural_key)
+            try:
                 result = serializer.create(serializer.validated_data)
-            except (IntegrityError, serializers.ValidationError):
+            except IntegrityError:
```
lubosmj, line 385: It looks like we create the object after fetching from the database? Lines 382-384

ipanova: maybe i will just remove L382 to no have have an extra DB call

pulpcore/app/serializers/base.py
```diff
@@ -368,8 +369,21 @@ def get_or_create(cls, natural_key, default_values=None):
...
+                for c in returned_codes:
+                    all_codes.extend(c)
+                if error_codes.intersection(all_codes):
+                    # validation failed because field is invalid, missing
+                    raise e
+                # recover from a race condition, where another thread just created the object
+                # validation failed with 400 'unique' error code
+                result = cls.Meta.model.objects.get(**natural_key)
+            try:
+                result = result if result else serializer.create(serializer.validated_data)
```
mdellweg, line 385:
```suggestion
                result = result or serializer.create(serializer.validated_data)
```

pulpcore/app/serializers/base.py
```diff
@@ -368,8 +369,21 @@ def get_or_create(cls, natural_key, default_values=None):
...
+                returned_codes = e.get_codes().values()
+                error_codes = set(["required", "invalid"])
+                all_codes = []
+                for c in returned_codes:
+                    all_codes.extend(c)
+                if error_codes.intersection(all_codes):
+                    # validation failed because field is invalid, missing
+                    raise e
+                # recover from a race condition, where another thread just created the object
+                # validation failed with 400 'unique' error code
```
mdellweg, line 382: I wonder if we could check for "unique" instead of "all but unique"?

ipanova: I need to raise error if at least one of error_codes are present. There can be a mixture of unique, required, invalid.

lubosmj (approved):

PR #4645, opened 2023-11-02
Fixed how Distribtuion.content_handler() is called.

dkliban:
fixes: #4644


mdellweg (approved):

dralley (approved):
@dkliban This looks OK but I have no idea what issue (symptoms) it's solving, there's no description

PR #4668, opened 2023-11-06
Add orphaned_for filter

### LINKED ISSUE #3364: As a user, I can view orphans that will be cleaned up during orphan cleanup
It would be nice to be able to view orphan content in Pulp before running orphan cleanup. Perhaps two options:

1. An orphan list endpoint that lists orphans that would be cleaned up during orphan cleanup
2. A "dry run" option to orphan cleanup that would list orphans that would be cleaned up during orphan cleanup

daviddavis:
fixes #3364

### DIFF
--- a/pulpcore/app/viewsets/content.py
+++ b/pulpcore/app/viewsets/content.py
@@ -1,6 +1,8 @@
 from gettext import gettext as _
 
+from django.conf import settings
 from django.db import models
+from django_filters import NumberFilter
 from rest_framework import mixins, permissions, status
 from rest_framework.response import Response
 
@@ -22,6 +24,17 @@ from .custom_filters import (
 )
 
 
+class OrphanedFilter(NumberFilter):
+    def filter(self, qs, value):
+        if value is not None:
+            if value < 0:
+                time = settings.ORPHAN_PROTECTION_TIME
+            else:
+                time = value
+            qs = qs.orphaned(int(time))
+        return qs
+
+
 class ArtifactFilter(BaseFilterSet):
     """
     Artifact filter Plugin content filters should:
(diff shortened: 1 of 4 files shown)


daviddavis:
If this change looks acceptable, I can add tests.

pulpcore/app/viewsets/content.py
```diff
@@ -33,6 +46,9 @@ class ArtifactFilter(BaseFilterSet):
     """
 
     repository_version = ArtifactRepositoryVersionFilter()
+    orphaned_for = OrphanedFilter(
```
ipanova, line 49: Per Daniel's comment on the issue, this might be too slow and intensive for an API call.
How about a django-admin command?

daviddavis: I'm a bit skeptical that this would be too slow or itensive for the API. It's paginated and it just does a single join on ContentArtifact (for artifacts) or RepositoryContent (for Content).

The problem with django admin commands is that they are difficult to run if you have Pulp running via k8s.

daviddavis:
Ok tests added.

ggainey:
The change in #4841 actually conflicts with the desire here - @daviddavis , can you and @gerrod3 get together and determine a/the right way to resolve the conflicting needs here?

daviddavis:
@ggainey @gerrod3 sure, I assume the goal would be to extend these filters to handle kickstart content? I'm not sure I can spend too much time on that since we don't use them but would be happy to help out in any way. Let me know how you all want to proceed.

ggainey (approved):

gerrod3 (approved):

PR #4678, opened 2023-11-07
Drop 3.16 from auto-ci-update, add 3.39

dralley:
[noissue]

### DIFF
--- a/template_config.yml
+++ b/template_config.yml
@@ -13,12 +13,11 @@ check_stray_pulpcore_imports: false
 ci_env: {}
 ci_trigger: '{pull_request: {branches: [''*'']}}'
 ci_update_branches:
-- '3.16'
-- '3.18'
-- '3.21'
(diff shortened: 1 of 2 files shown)


mdellweg (approved):

template_config.yml
```diff
@@ -13,12 +13,11 @@ check_stray_pulpcore_imports: false
 ci_env: {}
 ci_trigger: '{pull_request: {branches: [''*'']}}'
 ci_update_branches:
-- '3.16'
-- '3.18'
-- '3.21'
-- '3.22'
-- '3.23'
-- '3.28'
+- '3.18' # May, 2024
```
dralley, line 16: These are end-of-maintainence support dates.  3.16 technically ends in January so it's a liiiiittle bit early, but as this is just our CI update job it's not really going to matter.

ggainey (approved):

PR #4680, opened 2023-11-07
Use the new centos9 CI image for main branch

dralley:
~~We had a discussion a few weeks ago about bumping the minimum supported version as none of our stakeholders use 3.8 anymore.~~

Based on https://github.com/pulp/plugin_template/pull/845

And update the minimum required Python to 3.9.  This obviously only applies to new releases, existing branches will continue using the old CI image and support Python 3.8.

[noissue]


mdellweg:
This looks confusing. But the ci does not have much of a say which version of python pulp is running on. That must be adjusted in the ci-base-image.

dralley:
Does the CI failure have anything to do with this change? https://github.com/pulp/plugin_template/pull/845#discussion_r1487053638

setup.py
```diff
@@ -17,7 +17,7 @@
     author="Pulp Team",
     author_email="pulp-list@redhat.com",
     url="https://pulpproject.org",
-    python_requires=">=3.8",
+    python_requires=">=3.9",
```
dralley, line 20: Unsure if we want to commit to this yet, or wait.

mdellweg: I'd be fine with this.

mdellweg (approved):

dkliban (approved):

PR #4684, opened 2023-11-08
Fix file:// syncs  deleting the original files

### LINKED ISSUE #4681: file:// sync deletes files from directory

dralley:
closes #4681

### DIFF
--- a/pulpcore/tests/functional/api/pulp_file/test_sync.py
+++ b/pulpcore/tests/functional/api/pulp_file/test_sync.py
@@ -1,4 +1,6 @@
 """Tests that sync file plugin repositories."""
+
+import os
 import uuid
 
 import pytest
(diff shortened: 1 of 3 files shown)


pulpcore/tests/functional/api/pulp_file/test_sync.py
```diff
@@ -22,17 +24,20 @@ def test_sync_file_protocol_handler(
...
     }
     remote = gen_object_with_cleanup(file_remote_api_client, remote_kwargs)
+    files = set(os.listdir("/tmp/file/"))
 
     body = RepositorySyncURL(remote=remote.pulp_href)
     monitor_task(file_repository_api_client.sync(file_repo.pulp_href, body).task)
 
+    # test that all the files are still present
+    assert set(os.listdir("/tmp/file/")) == files
+
```
mdellweg, line 40: +1

dralley:
I noticed that we're copying and resaving the file in the temporary directory even in the file:// sync case.  That's not especially efficient and we should file an issue for it, but in the meantime we can simply make sure the sync process is working with that copy rather than the original.

mdellweg:
> I noticed that we're copying and resaving the file in the temporary directory even in the file:// sync case. That's not especially efficient and we should file an issue for it, but in the meantime we can simply make sure the sync process is working with that copy rather than the original.

Is that the `path` vs. `_path` change?

mdellweg:
Should we then bring the reverted change to delete downloaded files early back? Or was the revert never merged?

dralley:
@mdellweg It was never reverted, this is the whole fix.  `self._path` -> `self.path`, that is.

mdellweg (approved):

ipanova (approved):
can you squash into one commit?

mdellweg:
I believe it's on purpose, because the test itself is not backportable? Does that sound right @dralley ?

PR #4729, opened 2023-11-16
Correctly report file plugin provided by pulpcore

### LINKED ISSUE #4728: pulp_file after being merged still reports as python-package = "pulp-file"
Instead, i think it should be "pulpcore".
Complication: To my knowledge the current way is needed to allow tools to call bindings generation correctly. We may need to add a `python_module="pulp_file"` to the status response.

mdellweg:
Since being merged, the python package for the pulp app labeled 'file' is supposed to be "pulpcore". Due to some inconsistencies around the use of app label vs. python package vs. python module, we need some straightening all around.

fixes #4728


mdellweg:
Hmm, this is now green for the wrong reasons. We install the pulp_file clients in the pipeline in a different place, so this change will still make all the CI other than pulpcore's fail. Bummer...

dkliban (approved):

PR #4745, opened 2023-11-20
Add instrumentation to content-app

### LINKED ISSUE #3829: Add the OpenTelemetry Aiohttp-server instrumentation as a vendor library

lubosmj:
The instrumentation code from the upstream PR was brought down to pulpcore because of https://github.com/pypi/support/issues/3353.

closes #3829

### DIFF
--- a/pulpcore/tests/functional/api/pulp_file/test_telemetry_collection.py
+++ b/pulpcore/tests/functional/api/pulp_file/test_telemetry_collection.py
@@ -1,0 +1,92 @@
+import requests
+import uuid
+
+from urllib.parse import urljoin, urlparse
+from django.conf import settings
+
+from pulpcore.client.pulp_file import FileFileDistribution, RepositoryAddRemoveContent
+
+
+def test_get_requests(
+    file_distribution_api_client,
+    file_repository_api_client,
+    file_repo_with_auto_publish,
+    file_content_unit_with_name_factory,
+    gen_object_with_cleanup,
+    monitor_task,
+    received_otel_span,
+    test_path,
+):
+    """Test if content-app correctly returns mime-types based on filenames."""
+    content_units = [
+        file_content_unit_with_name_factory("otel_test_file1.tar.gz"),
+        file_content_unit_with_name_factory("otel_test_file2.xml.gz"),
+        file_content_unit_with_name_factory("otel_test_file3.txt"),
+    ]
+    units_to_add = list(map(lambda f: f.pulp_href, content_units))
+    data = RepositoryAddRemoveContent(add_content_units=units_to_add)
+    monitor_task(
+        file_repository_api_client.modify(file_repo_with_auto_publish.pulp_href, data).task
+    )
+
+    data = FileFileDistribution(
+        name=str(uuid.uuid4()),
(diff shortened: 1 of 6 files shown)


lubosmj:
Telemetry tests are enabled only for the Azure runner: https://github.com/pulp/pulpcore/blob/f0569f43fcbb689fbfbbc309fa7af4ee3518b346/template_config.yml#L52

pulpcore/tests/functional/api/pulp_file/test_telemetry_collection.py
```diff
@@ -0,0 +1,82 @@
...
+        url = urljoin(distribution.base_url, content_unit.relative_path)
+
+        s = requests.Session()
+        s.headers = {"User-Agent": test_path}
+
+        s.get(url, allow_redirects=False)
+        assert received_otel_span(
+            {
+                "http.method": "GET",
+                "http.target": f"/pulp/content/{distribution.base_path}/{file_name}",
```
lubosmj, line 48: I need to take care of `domains` and redirects to content storage here...

pulpcore/tests/functional/assets/otel_server.py
```diff
@@ -68,6 +68,7 @@ async def _traces_handler(request):
                         for item in span.attributes
                     }
                     spans.append(attrs)
+                    print(attrs)
```
decko, line 71: Was intentional for this to be here?

decko (approved):
LGTM with just a question

pulpcore/tests/functional/assets/otel_server.py
lubosmj: It is not cluttering the CI logs and it helps with further debugging to see whether the span was received on the server or not. Do you think logging is better than printing? I am going to replace it.

bmbouter:
It seems the upstream is not yet able to release this, but they are working on it, see [this](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2053) and [this](https://pypi.org/search/?q=opentelemetry-instrumentation-aiohttp-server&o=).

Given that it's not available yet, I'm in favor of vendoring it for a release or two in Pulp because we have users who have been waiting and are very eager to try it. Also it comes with functional tests so unlike the previous pull request this does have tests that cover the vendored code.

@mdellweg wdyt about this approach? I don't want to have this merged without your input.

bmbouter (approved):
The tests look good, thanks for them! Also +1 to @decko 's question about the `print()` statement. Personally I'd remove it, but it's up to @lubosmj 

thanks @lubosmj !

pulpcore/tests/functional/assets/otel_server.py
lubosmj: Removed.

PR #4752, opened 2023-11-21
Fix wording of API_APP_TTL warning

sdherr:
[noissue]

### DIFF
--- a/pulpcore/app/entrypoint.py
+++ b/pulpcore/app/entrypoint.py
@@ -49,7 +49,7 @@ class PulpApiWorker(SyncWorker):
 
         if settings.API_APP_TTL < 2 * self.timeout:
             logger.warn(
-                "API_APP_TTL (%s) is smaller than half the gunicorn timeout (%s). "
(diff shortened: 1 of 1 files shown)


sdherr:
I'm assuming that the python here is correct, not the warning message. If it's the other way around and the warning message is actually correct then you'll have to change the code.

pulpcore/app/entrypoint.py
```diff
@@ -49,7 +49,7 @@ def init_process(self):
 
         if settings.API_APP_TTL < 2 * self.timeout:
             logger.warn(
-                "API_APP_TTL (%s) is smaller than half the gunicorn timeout (%s). "
+                "API_APP_TTL (%s) is smaller than double the gunicorn timeout (%s). "
```
mdellweg, line 52: Absolutely correct. (I'd say "twice" but I'm not the native speaker on this side of the channel.)

mdellweg (approved):

lubosmj (approved):

Issue #4315, opened 2023-08-22
Task resources for objects use hrefs which may not be stable


mdellweg:
related to #4207 
Also switching domains on or off changes how hrefs are represented.

Possible solution:
Instead of `/pulp/[domain/]api/v3/repositories/file/file/<ID>` we can use `repository:file:file:<ID>`.
For some intermediary releases, we would use both in parallel.

PR #4761, opened 2023-11-21
Introduce Pulp Resource Name (PRN) into task resource locking

gerrod3:
fixes: #4315

### DIFF
--- a/CHANGES/4315.misc
+++ b/CHANGES/4315.misc
@@ -1,0 +1,2 @@
+Added new task resource locking method ``get_prn`` that does not change with different settings.
+``get_url`` is planned to be phased out of resource locking by pulpcore 3.55.
--- a/pulpcore/app/util.py
+++ b/pulpcore/app/util.py
@@ -13,7 +13,7 @@ from datetime import timedelta
 import gnupg
 
 from django.conf import settings
-from django.db.models import Sum
+from django.db.models import Model, Sum
 from django.urls import Resolver404, resolve, reverse
 from opentelemetry import metrics
 
@@ -32,7 +32,7 @@ _model_viewset_cache = {}
 def get_url(model, domain=None):
     """
     Get a resource url for the specified model instance or class. This returns the path component of
-    the resource URI.  This is used in our resource locking/reservation code to identify resources.
+    the resource URI.
 
     Args:
         model (django.models.Model): A model instance or class.
@@ -57,6 +57,45 @@ def get_url(model, domain=None):
         kwargs["pk"] = model.pk
 
     return reverse(get_view_name_for_model(model, view_action), kwargs=kwargs)
+
+
+def get_prn(instance=None, uri=None):
+    """
+    Get a Pulp Resource Name (PRN) for the specified model instance. It is similar to a HREF
+    url in that it uniquely identifies a resource, but it also has the guarantee that it will not
+    change regardless of API_ROOT or DOMAIN_ENABLED. This is used in our resource locking/
+    reservation code to identify resources.
+
+    The format for the PRN is as follows:
+    ```
+        prn:model-label-lower:pk
+    ```
+
+    Examples:
+        instance=FileRepository(pk=123) -> prn:file.filerepository:123
+        instance=Artifact(pk=abc) -> prn:core.artifact:abc
+        uri=/rerouted/api/v3/repositories/rpm/rpm/123/versions/2/ -> prn:core.repositoryversion:abc
+        uri=/pulp/foodomain/api/v3/content/ansible/role/123/ -> prn:ansible.role:123
+
+    Args:
+        instance Optional(django.models.Model): A model instance.
+        uri Optional(str): A resource URI
+
+    Returns:
+        prn (str): The PRN of the passed in resource
+    """
+    if uri:
+        from pulpcore.app.viewsets import NamedModelViewSet
+
+        instance = NamedModelViewSet.get_resource(uri)
+
+    if not isinstance(instance, Model):
+        raise ValidationError(_("instance({}) must be a Model").format(instance))
+
+    if isinstance(instance, models.MasterModel):
+        instance = instance.cast()
+
+    return f"prn:{instance._meta.label_lower}:{instance.pk}"
 
 
 def extract_pk(uri):
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -13,7 +13,7 @@ from django.db.models import Model, Max
 from django_guid import get_guid
 from pulpcore.app.apps import MODULE_PLUGIN_VERSIONS
 from pulpcore.app.models import Task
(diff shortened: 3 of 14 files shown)


pulpcore/app/util.py
```diff
@@ -55,6 +56,45 @@ def get_url(model, domain=None):
...
+def get_prn(instance=None, uri=None):
+    """
+    Get a Pulp Resource Name (PRN) for the specified model instance. It is similar to a HREF
+    url in that it uniquely identifies a resource, but it also has the guarantee that it will not
+    change regardless of API_ROOT or DOMAIN_ENABLED. This is used in our resource locking/
+    reservation code to identify resources.
+
+    The format for the PRN is as follows:
+    ```
+        prn:model-label-lower:id
```
mdellweg, line 68:
```suggestion
        prn:model-label-lower:pk
```

BTW: This makes `"shared:prn:label:pk"` for readonly locks. Probably not an issue.

CHANGES/4315.misc
```diff
@@ -0,0 +1,2 @@
+Added new task resource locking method ``get_prn`` that does not change with different settings.
+``get_url`` is planned to be phased out of resource locking by pulpcore 3.55.
```
mdellweg, line 2: Is this a zero downtime upgrade safe plan?

gerrod3: Yes, `get_url` is still used with `get_prn` so tasks can still be dispatch and ran during upgrade. Only on 3.55 will `get_url` be removed from task locking, which will require a planned downtime.

mdellweg: Why would we need a planned downtime?

gerrod3: Probably won't, but if we want to allow going straight from 3.40 to 3.56 then technically we would need a downtime (to be safe of task locking). So long as the upgrade is within the window of when this change will be introduce, then there will be no need for a downtime.

mdellweg: I see. The migration should block that upgrade then.

pulpcore/tests/functional/api/using_plugin/test_tasks.py
```diff
@@ -59,15 +60,20 @@ def setup_filter_fixture(
...
 
-    return (repo_sync_task, repo_update_task)
+    return repo_sync_task, repo_update_task, file_repo, remote
 
 
 def test_filter_tasks_by_reserved_resources(setup_filter_fixture, tasks_api_client):
     """Filter all tasks by a particular reserved resource."""
-    repo_sync_task, repo_update_task = setup_filter_fixture
-    reserved_resources_record = repo_update_task.reserved_resources_record[0]
+    repo_sync_task, repo_update_task, _, _ = setup_filter_fixture
```
decko, line 70: 😍

pulpcore/tasking/tasks.py
```diff
@@ -146,8 +150,10 @@ def dispatch(
         shared_resources = _validate_and_get_resources(shared_resources)
 
     # A task that is exclusive on a domain will block all tasks within that domain
+    domain_prn = get_prn(get_domain())
     domain_url = get_url(get_domain())
-    if domain_url not in exclusive_resources:
+    if not (domain_prn in exclusive_resources or domain_url in exclusive_resources):
```
decko, line 155: Would this make it more clear that we are looking for our domain to be on the exclusive resources list?
```suggestion
    if (domain_prn or domain_url) not in exclusive_resources:
```

pulpcore/app/util.py
```diff
@@ -59,6 +59,45 @@ def get_url(model, domain=None):
...
+
+    The format for the PRN is as follows:
+    ```
+        prn:model-label-lower:pk
+    ```
+
+    Examples:
+        instance=FileRepository(pk=123) -> prn:file.filerepository:123
+        instance=Artifact(pk=abc) -> prn:core.artifact:abc
+        uri=/rerouted/api/v3/repositories/rpm/rpm/123/versions/2/ -> prn:core.repositoryversion:abc
```
decko, line 77: Shouldn't the result be `prn:rpm.repositoryversion:abc`

decko: Or `RepositoryVersion` lives only under the `core` app?

gerrod3: Correct, `RepositoryVersion` only exists in pulpcore, plugins do not create child model versions.

pulpcore/tasking/tasks.py
lubosmj: This will not work for the latter one, see:
```
>>> a = "bla"
>>> b = "blabla"
>>> 
>>> (a or b) in ["blabla"]
False
>>> (a or b)
'bla'
>>> (a or b) in ["bla"]
True
```

pulpcore/app/util.py
mdellweg: But also: Resource locking must happen on the repository. Locking on a Version is unsafe.

pulpcore/tests/functional/api/using_plugin/test_tasks.py
```diff
@@ -59,15 +60,20 @@ def setup_filter_fixture(
     repo_update_action = file_bindings.RepositoriesFileApi.partial_update(
         file_repo.pulp_href, {"description": str(uuid4())}
     )
-    repo_update_task = tasks_api_client.read(repo_update_action.task)
+    repo_update_task = monitor_task(repo_update_action.task)
```
mdellweg, line 64: Can we just use the file_repository_factory instead?

mdellweg (commented):
Overall, this looks nice and tidy and straight forward.

I left some testing related nitpicks. Also I want to encourage you to look for hardcoded urls in resources like `/pulp/api/v3/orphans` and `/pulp/api/v3/distributions`. Would be nice to get them on the same new schema.

dralley:
I added the breaking changes label based on the pulpcore team discussion - is it this PR that is "breaking changes" or is it the future one which will remove the URL based locking code which is?

gerrod3:
> I added the breaking changes label based on the pulpcore team discussion - is it this PR that is "breaking changes" or is it the future one which will remove the URL based locking code which is? Or both?
> 
> Could you point out the break (if this PR does contain a breaking change)?

The future change with the url removal will be the breaking one.

dralley:
Ok, removing the tag then.  Can we file an issue for the future breaking change?

pulpcore/tests/functional/api/using_plugin/test_tasks.py
gerrod3: I couldn't change the scope since the fixtures inside have different scopes themselves and started to conflict.

gerrod3:
> Ok, removing the tag then. Can we file an issue for the future breaking change?

https://github.com/pulp/pulpcore/issues/5148

dkliban (dismissed):

ggainey (approved):

dralley (approved):

PR #4763, opened 2023-11-22
Turn a few verb-only pytest fixtures into standard functions

dralley:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -1060,7 +1060,6 @@ def add_to_filesystem_cleanup():
                 pass
 
 
-@pytest.fixture(scope="session")
 def download_content_unit(bindings_cfg, pulp_domain_enabled):
     def _download_content_unit(base_path, content_path, domain="default"):
         async def _get_response(url):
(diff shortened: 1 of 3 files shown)


dralley:
@mdellweg 

>Probably not if someone started using it.  That is until pulpcore 3.55.

I don't think we provide semantic versioning guarantees for test infrastructure?  And any time we do test infra changes we often end up backporting it anyway, just to avoid incompatibility when future tests are added.

>None at all. Just use requests.get.

IMO, if we do that, and just eat the extra dependency, then we might as well just delete the function entirely.  There's no point having an alias for a function as simple as `requests.get`

gerrod3:
I would say deprecate the fixtures, copy them to the utils and slowing switch tests over to importing the util functions. Once all the plugins have switched, then we can remove the fixtures.

mdellweg:
> I would say deprecate the fixtures, copy them to the utils and slowing switch tests over to importing the util functions. Once all the plugins have switched, then we can remove the fixtures.

It's the lower bounds test that makes me think about proper deprecation and removal only in a breaking change release.

pulpcore/tests/functional/__init__.py
```diff
@@ -1060,7 +1060,6 @@ def _add_to_filesystem_cleanup(path):
                 pass
 
 
-@pytest.fixture(scope="session")
```
mdellweg, line 1063: I don't think we want this to stop being a fixture. It is using other fixtures.

PR #4767, opened 2023-11-23
Fix a race condition in distribution filtering

### LINKED ISSUE #4766: Race in DistributionWithContentFilter can lead to 500
The only possible explanation I can find is that cast is messing up the related prefetched models, while acutally not be needed.
All we need is the class of the distribution. Not the detail instance.

``` python
    for dist in qs.exclude(repository=None).prefetch_related("repository__versions"):
        if dist.cast().SERVE_FROM_PUBLICATION:
            versions = dist.repository.versions.values_list("pk", flat=True)
            ....
```

```
pulp [c19d4bd0-3b4c-4d3a-87fb-9cd2b5c2edd2]: django.request:ERROR: Internal Server Error: /pulp/api/v3/distributions/file/file/
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/rest_framework/viewsets.py", line 125, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/rest_framework/mixins.py", line 38, in list
    queryset = self.filter_queryset(self.get_queryset())
  File "/usr/local/lib/python3.8/site-packages/rest_framework/generics.py", line 150, in filter_queryset
    queryset = backend().filter_queryset(self.request, queryset, self)
  File "/usr/local/lib/python3.8/site-packages/django_filters/rest_framework/backends.py", line 74, in filter_queryset
    return filterset.qs
  File "/usr/local/lib/python3.8/site-packages/django_filters/filterset.py", line 230, in qs
    qs = self.filter_queryset(qs)
  File "/usr/local/lib/python3.8/site-packages/django_filters/filterset.py", line 213, in filter_queryset
    queryset = self.filters[name].filter(queryset, value)
  File "/usr/local/lib/python3.8/site-packages/pulpcore/app/viewsets/custom_filters.py", line 424, in filter
    versions = dist.repository.versions.values_list("pk", flat=True)
AttributeError: 'NoneType' object has no attribute 'versions'
```

mdellweg:
This change adds a `detail_model` property to the master-detail concept and removes a race condition by replacing calls to `cast` with it in `DistributionWithContentFilter`.

fixes #4766

### DIFF
--- a/pulpcore/app/viewsets/custom_filters.py
+++ b/pulpcore/app/viewsets/custom_filters.py
@@ -355,12 +355,12 @@ class DistributionWithContentFilter(Filter):
         for dist in qs.exclude(publication=None).values("publication__repository_version", "pk"):
             versions_distributions[dist["publication__repository_version"]].append(dist["pk"])
 
-        for dist in qs.exclude(repository_version=None).values("repository_version", "pk"):
-            if not dist.cast().SERVE_FROM_PUBLICATION:
+        for dist in qs.exclude(repository_version=None).values("repository_version", "pulp_type"):
+            if not dist.detail_model.SERVE_FROM_PUBLICATION:
                 versions_distributions[dist["repository_version"]].append(dist["pk"])
 
         for dist in qs.exclude(repository=None).prefetch_related("repository__versions"):
-            if dist.cast().SERVE_FROM_PUBLICATION:
+            if dist.detail_model.SERVE_FROM_PUBLICATION:
                 versions = dist.repository.versions.values_list("pk", flat=True)
                 publications = Publication.objects.filter(
                     repository_version__in=versions, complete=True
--- a/CHANGES/4766.bugfix
+++ b/CHANGES/4766.bugfix
@@ -1,0 +1,1 @@
+Removed a race condition leading to 500 error in distribution filtering.
--- a/CHANGES/plugin_api/4766.feature
+++ b/CHANGES/plugin_api/4766.feature
@@ -1,0 +1,1 @@
+Added the async safe `detail_model` property to master detail models.
--- a/pulpcore/app/models/base.py
+++ b/pulpcore/app/models/base.py
@@ -148,16 +148,20 @@ class MasterModel(BaseModel, metaclass=MasterModelMeta):
     def get_model_for_pulp_type(cls, pulp_type):
         return cls._pulp_model_map[pulp_type]
 
-    def save(self, *args, **kwargs):
+    @property
+    def detail_model(self):
+        return self._pulp_model_map[self.pulp_type]
+
+    def __init__(self, *args, **kwargs):
         # instances of "detail" models that subclass MasterModel are exposed
         # on instances of MasterModel by the string stored in that model's TYPE attr.
         # Storing this pulp_type in a column on the MasterModel next to makes it trivial
         # to filter for specific detail model types across master's relations.
         # Prepend the TYPE defined on a detail model with a django app label.
         # If a plugin sets the type field themselves, it's used as-is.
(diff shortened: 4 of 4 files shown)


gerrod3 (commented):
Was `cast` really causing a DB call? I thought you changed it a while back to use the lookup table?

mdellweg:
I changed it from recursive to using the lookup table, but it still needs look at the detail db-table.
And is this case we really only want to know about a class attribute.

gerrod3 (approved):
I think the race condition window is still present, just much smaller without all the extra `cast` db calls.

mdellweg:
> I think the race condition window is still present, just much smaller without all the extra `cast` db calls.

Where do you see that race?

pulpcore/app/viewsets/custom_filters.py
```diff
@@ -356,11 +356,11 @@ def filter(self, qs, value):
             versions_distributions[dist["publication__repository_version"]].append(dist["pk"])
 
         for dist in qs.exclude(repository_version=None).values("repository_version", "pk"):
```
gerrod3, line 358:
```suggestion
        for dist in qs.exclude(repository_version=None).values("repository_version", "pulp_type"):
```
We need to pull in the `pulp_type` or else the `detail_model` call will trigger a db lookup. `pk` is always pulled in, so no need to list it.

pulpcore/app/viewsets/custom_filters.py
```diff
@@ -356,11 +356,11 @@ def filter(self, qs, value):
...
-            if not dist.cast().SERVE_FROM_PUBLICATION:
+            if not dist.detail_model.SERVE_FROM_PUBLICATION:
                 versions_distributions[dist["repository_version"]].append(dist["pk"])
 
         for dist in qs.exclude(repository=None).prefetch_related("repository__versions"):
```
gerrod3, line 362: > > I think the race condition window is still present, just much smaller without all the extra `cast` db calls.
> 
> Where do you see that race?

I'm not sure, but I have a feeling that there might be race condition in this loop. There are so many queries being performed here, that it just seems likely we would encounter one here. Also, this `prefetch_related` is probably doing nothing but adding an extra query. This loop probably needs to be refactored somehow.

gerrod3 (changes requested):

pulpcore/app/viewsets/custom_filters.py
mdellweg: I believe the race was that the query with `.exclude(repository=None)` returned a distribution that after casting actually had `dist.repository == None`. And I don't see that this can happen again. But, there may be more such windows for sure.

ggainey (approved):

PR #4792, opened 2023-11-29
Fix division-by-zero on import

### LINKED ISSUE #4777: Division by zero during Pulp import
**Version**
3.36.0+

**Describe the bug**

During a Katello test run, the following exception was encountered

```
 {"traceback"=>"  File \"/usr/lib/python3.11/site-packages/pulpcore/tasking/tasks.py\", line 61, in _execute_task\n    result = func(*args, **kwargs)\n             ^^^^^^^^^^^^^^^^^^^^^\n  File \"/usr/lib/python3.11/site-packages/pulpcore/app/tasks/importer.py\", line 453, in pulp_import\n    with tarfile.open(path, \"r\", fileobj=fp) as tar:\n         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/usr/lib64/python3.11/tarfile.py\", line 1815, in open\n    fileobj.seek(saved_pos)\n  File \"/usr/lib/python3.11/site-packages/pulpcore/app/tasks/importer.py\", line 132, in seek\n    self.chunk = target // self.chunk_size\n                 ~~~~~~~^^~~~~~~~~~~~~~~~~\n", "description"=>"integer division or modulo by zero"} (Katello::Errors::Pulp3Error)
```

I'm not 100% certain, but I suspect the cause here is that if chunks aren't used during the export, chunk_size is set to 0:

https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/export.py#L456C31-L456C31

and ChunkedFile reads that value here:

https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/importer.py#L79

and we're using ChunkedFile even in the non-chunked case so long as a TOC file was provided.

https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/importer.py#L335-L336

**To Reproduce**

I would expect it to be reproducible if performing a non-chunked export, and providing a TOC path for the import.

**Expected behavior**

We should never see an integer division by zero error.

**Additional context**

dralley:
closes #4777

TODO: still working on the tests, it's a bit difficult to piece together how the fixtures interact...

### DIFF
--- a/pulpcore/app/tasks/export.py
+++ b/pulpcore/app/tasks/export.py
@@ -422,12 +422,10 @@ def pulp_export(exporter_pk, params):
                         os.remove(pathname)
                     raise
             # compute the hashes
-            global_hash = hasher()
             paths = sorted([str(Path(p)) for p in glob(tarfile_fp + ".*")])
             for a_file in paths:
-                a_hash = compute_file_hash(a_file, hasher=hasher(), cumulative_hash=global_hash)
+                a_hash = compute_file_hash(a_file, hasher=hasher())
                 rslts[a_file] = a_hash
-            tarfile_hash = global_hash.hexdigest()
 
         else:
             # write into the file
@@ -450,23 +448,20 @@ def pulp_export(exporter_pk, params):
         # write outputfile/hash info to a file 'next to' the output file(s)
         output_file_info_path = tarfile_fp.replace(".tar", "-toc.json")
         with open(output_file_info_path, "w") as outfile:
-            if the_export.validated_chunk_size:
-                chunk_size = the_export.validated_chunk_size
-            else:
-                chunk_size = 0
-            chunk_toc = {
+            table_of_contents = {
                 "meta": {
-                    "chunk_size": chunk_size,
-                    "file": os.path.basename(tarfile_fp),
-                    "global_hash": tarfile_hash,
                     "checksum_type": checksum_type,
                 },
                 "files": {},
             }
+
+            if the_export.validated_chunk_size:
+                table_of_contents["meta"]["chunk_size"] = the_export.validated_chunk_size
+
             # Build a toc with just filenames (not the path on the exporter-machine)
             for a_path in rslts.keys():
-                chunk_toc["files"][os.path.basename(a_path)] = rslts[a_path]
-            json.dump(chunk_toc, outfile)
+                table_of_contents["files"][os.path.basename(a_path)] = rslts[a_path]
+            json.dump(table_of_contents, outfile)
 
         # store toc info
         toc_hash = compute_file_hash(output_file_info_path)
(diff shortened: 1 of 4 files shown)


pulpcore/app/tasks/export.py
```diff
@@ -453,7 +453,7 @@ def pulp_export(exporter_pk, params):
             if the_export.validated_chunk_size:
                 chunk_size = the_export.validated_chunk_size
             else:
-                chunk_size = 0
+                chunk_size = os.path.getsize(tarfile_fp)
```
mdellweg, line 456: This is on the export side. Do we want to account for this on the import side too, so we can handle existing exports?

dralley: Existing imports aren't a huge concern, but I'm definitely not against defense in depth.  But which approach to use?

We could set it null and have the import side figure it out.  Or we could have the import side not use ChunkedFile when there's no chunks, or 1 chunk.  Or we could have the ChunkedFile ignore the chunking logic in some circumstances.  

Separately, we could also make `chunk_size` required if you make a chunked export, or we could set a default value.  Currently if there is no chunk size provided, there will just be one chunk.

mdellweg: I'd be fine with any of these solutions.

ggainey: > I'd be fine with any of these solutions.

My $0.02 would be to do this same thing on the import-side - if chunbked_size==0 set it to filesize. That addresses the issue, and lets the rest of the codeflow work without needing special-casing.

dralley:
After some discussion with the Katello folks, the most convenient thing to do at the moment is to merge this without tests and follow up with those tests in a separate PR.  So that's what I'll do.

pulpcore/app/tasks/export.py
```diff
@@ -450,23 +448,20 @@ def pulp_export(exporter_pk, params):
...
-            if the_export.validated_chunk_size:
-                chunk_size = the_export.validated_chunk_size
-            else:
-                chunk_size = 0
-            chunk_toc = {
+            table_of_contents = {
                 "meta": {
-                    "chunk_size": chunk_size,
-                    "file": os.path.basename(tarfile_fp),
-                    "global_hash": tarfile_hash,
```
dralley, line 461: `global_hash` and `file` are no longer used in the import path anywhere

pulpcore/app/tasks/export.py
```diff
@@ -450,23 +448,20 @@ def pulp_export(exporter_pk, params):
...
                     "checksum_type": checksum_type,
                 },
                 "files": {},
             }
+
+            if the_export.validated_chunk_size:
+                table_of_contents["meta"]["chunk_size"] = the_export.validated_chunk_size
```
dralley, line 459: Set `chunk_size` only if it was provided

mdellweg (approved):

ggainey (approved):

PR #4799, opened 2023-11-29
Update CI files for branch main

pulpbot:


mdellweg (approved):

docs/template_gitref
```diff
@@ -1 +1 @@
-2021.08.26-278-g22c5b00
+unknown
```
dralley, line 1: Is this broken?

dralley: @mdellweg

mdellweg: Only in this call i hope. I made the plugin template not fail to update because of this. Should be back with the `fetch-depth=0` brought by this update.

ipanova (approved):

Issue #4806, opened 2023-11-29
repository version repair always reports digest mismatch


dkliban:
**Version**
3.41.0

**Describe the bug**
I have a domain backed by an S3 bucket. I create a repository version with 35 RPMs in it. I run the repair task on that repository version and each time I run it I see the following output:

```
pulp [4a4b2b31574a476ca539b15b72a575d2]: pulpcore.app.tasks.repository:WARNING: Digest mismatch for <Artifact: pk=018c1bf5-349c-7a40-a3d2-118e53d8a7f3>
pulp [4a4b2b31574a476ca539b15b72a575d2]: pulpcore.app.tasks.repository:WARNING: Digest mismatch for <Artifact: pk=018c1bf5-34ab-760c-bbc5-f61a75f6755b>
pulp [4a4b2b31574a476ca539b15b72a575d2]: pulpcore.app.tasks.repository:WARNING: Digest mismatch for <Artifact: pk=018c1bf5-34b6-73be-af71-afc0b47109f0>
pulp [4a4b2b31574a476ca539b15b72a575d2]: pulpcore.app.tasks.repository:WARNING: Digest mismatch for <Artifact: pk=018c1bf5-34be-761c-92d1-47e69c2f23a2>
pulp [4a4b2b31574a476ca539b15b72a575d2]: pulpcore.app.tasks.repository:WARNING: Digest mismatch for <Artifact: pk=018c1bf5-34cd-747d-abcb-ab6f538f38c7>
```

This is produced for every artifact in the repository version. 

The task finishes successfully every time and says that 35 artifacts were repaired.

Issue #4776, opened 2023-11-27
repo version repair command errors with traceback


jlsherrill:
**Version**
        "rpm": "3.24.0",
        "core": "3.41.0",
        "file": "3.41.0",

**Describe the bug**
When running a repo repair for an rpm repo, we got a traceback on a couple of the tasks.  It sorta looks its a failed download, but could be something else?  If it is a failed download, is it stopping immediately?

**To Reproduce**
have a repo version with missing units
Repair the repo version with verify_checksums set to true
simulate a download failure (maybe?)
Get the traceback below

**Expected behavior**
Repair will try to redownload as much as possible before failing

**Additional context**

"traceback: File \"/usr/local/lib/python3.8/site-packages/pulpcore/tasking/tasks.py\", line 60, in _execute_task\n result = func(*args, **kwargs)\n File \"/usr/local/lib/python3.8/site-packages/pulpcore/app/tasks/repository.py\", line 184, in repair_version\n loop.run_until_complete(\n File \"/usr/lib64/python3.8/asyncio/base_events.py\", line 616, in run_until_complete\n return future.result()\n File \"/usr/local/lib/python3.8/site-packages/pulpcore/app/tasks/repository.py\", line 146, in _repair_artifacts_for_content\n valid = await loop.run_in_executor(\n File \"/usr/lib64/python3.8/concurrent/futures/thread.py\", line 57, in run\n result = self.fn(*self.args, **self.kwargs)\n File \"/usr/local/lib/python3.8/site-packages/pulpcore/app/tasks/repository.py\", line 103, in _verify_artifact\n for chunk in fp.chunks(CHUNK_SIZE):\n File \"/usr/local/lib/python3.8/site-packages/django/core/files/base.py\", line 55, in ...

mdellweg:
Can you add the error text (I think it is above the traceback in the logs)? Also it looks like the traceback was truncated.
All i can derive from this now is that `_verify_artifact` may have failed.

dkliban:
I was able to reproduce this only by setting the `endpoint_url` in the domain storage_settings to `s3.us-east-1.amazonaws.com` without the `https://`. However, when I do this, all other operations such a sync and publish stop working also. However, the instance where we are seeing this behavior is continuing to sync and publish correctly.

mdellweg:
[...]

PR #4815, opened 2023-11-30
Fixes repository version repair API for non-default domains.

dkliban:
The 'current_domain' ContextVar was not getting coppied into the thread that was running the artifact
verification. The context is now explicitly copied into that thread.
    
This patch also adds repair API tests for S3 and Azure storage backends.
    
This patch also adds test assertions that a repository version repair task can be run in a non-default
domain.

fixes: #4776
fixes: #4806

### DIFF
--- a/CHANGES/4806.bugfix
+++ b/CHANGES/4806.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug in the repository version repair API for non-default domains.
--- a/pulpcore/app/tasks/repository.py
+++ b/pulpcore/app/tasks/repository.py
@@ -1,4 +1,5 @@
 from concurrent.futures import ThreadPoolExecutor
+from contextvars import copy_context
 from gettext import gettext as _
 from logging import getLogger
 import asyncio
@@ -96,6 +97,8 @@ async def _repair_ca(content_artifact, repaired=None):
 
 
 def _verify_artifact(artifact):
+    domain = get_domain()
+    assert domain.pk == artifact.pulp_domain_id
     try:
         # verify files digest
         hasher = hashlib.sha256()
@@ -144,7 +147,7 @@ async def _repair_artifacts_for_content(subset=None, verify_checksums=True):
                     # Should stay in (an) executor so that at least it doesn't completely block
                     # downloads.
                     valid = await loop.run_in_executor(
-                        checksum_executor, _verify_artifact, artifact
+                        checksum_executor, copy_context().run, _verify_artifact, artifact
                     )
                     if not valid:
                         await corrupted.aincrement()
--- a/CHANGES/4776.bugfix
+++ b/CHANGES/4776.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug in the repository version repair API for non-default domains.
--- a/pulpcore/tests/functional/api/pulp_file/test_domains.py
+++ b/pulpcore/tests/functional/api/pulp_file/test_domains.py
@@ -5,6 +5,7 @@ import json
 from pulpcore.app import settings
 from pulpcore.client.pulp_file import ApiException
 from pulpcore.client.pulpcore import ApiException as CoreApiException
+from pulpcore.client.pulpcore import Repair
 from pulpcore.tests.functional.utils import generate_iso, download_file
 
 
@@ -180,6 +181,7 @@ def test_content_upload(
 def test_content_promotion(
     domains_api_client,
     file_repository_api_client,
+    file_repository_version_api_client,
     basic_manifest_path,
     file_remote_factory,
     file_publication_api_client,
@@ -232,6 +234,14 @@ def test_content_promotion(
         download = download_file(f"{distro.base_url}{path}")
         assert download.response_obj.status == 200
         assert len(download.body) == 1024
+
+    # Test that a repository version repair operation can be run without error
+    response = file_repository_version_api_client.repair(
+        repo.latest_version_href, Repair(verify_checksums=True)
+    )
+    results = monitor_task(response.task)
+    assert results.state == "completed"
+    assert results.error is None
 
     # Cleanup to delete the domain
     task = file_repository_api_client.delete(repo.pulp_href).task
--- a/pulpcore/tests/functional/api/using_plugin/test_repair.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_repair.py
@@ -1,27 +1,12 @@
 import pytest
-import os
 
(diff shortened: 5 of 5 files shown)


pulpcore/app/tasks/repository.py
```diff
@@ -96,16 +97,34 @@ async def _repair_ca(content_artifact, repaired=None):
...
+            for chunk in s3_object["Body"].iter_chunks(CHUNK_SIZE):
                 hasher.update(chunk)
-        return hasher.hexdigest() == artifact.sha256
-    except FileNotFoundError:
+        except ClientError:
+            return False
+    else:
+        # We don't currently support performing a checksum check with other storage backends
+        # Assume that it is wrong.
         return False
```
mdellweg, line 124: This will render all "other" storage backends to always fail on repair. Is that really desired?

pulpcore/app/tasks/repository.py
```diff
@@ -96,16 +97,34 @@ async def _repair_ca(content_artifact, repaired=None):
 
 
 def _verify_artifact(artifact):
-    try:
-        # verify files digest
-        hasher = hashlib.sha256()
-        with artifact.file as fp:
```
mdellweg, line 102: I do not understand what we are doing wrong here. I'd believe that django storages is there to abstract exactly this.

CHANGES/4806.bugfix
```diff
@@ -0,0 +1,2 @@
+Fixed a bug where the repository version repair operation would always report a digest mismatch when
+using S3 storage.
```
mdellweg, line 2: Is it one bugfix for two issues? Then you can use the same text in both files, and towncrier will list both issues net to one changelog entry.

pulpcore/app/tasks/repository.py
```diff
@@ -96,16 +97,34 @@ async def _repair_ca(content_artifact, repaired=None):
...
-            for chunk in fp.chunks(CHUNK_SIZE):
+    domain = artifact.pulp_domain
+    storage = domain.get_storage()
```
gerrod3, line 101: I think you should try this first:
```python
domain = artifact.pulp_domain
storage = domain.get_storage()
with storage.open(artifact.file.name) as fp:
    for chunk in fp.chunks(CHUNK_SIZE):
```

pulpcore/app/tasks/repository.py
dkliban: It's not going to fail. It's going to repair all the files. It is outside the scope of this PR to fix this for Azure and Google storage backends.

dkliban: No longer a problem with the latest implementation.

pulpcore/app/tasks/repository.py
gerrod3: Also if you believe that the artifacts are all apart of the same domain that the task is running in (which it should be if the filters are correct), then you can make this more efficient by replacing the first line with `domain = get_domain()`. This way we will not perform a db query for each artifact we are repairing.

dkliban: Good suggestion. I'll incorporate it.

CHANGES/4806.bugfix
dkliban: Yes, this fix addresses two issues at the same time.

pulpcore/app/tasks/repository.py
```diff
@@ -96,10 +96,11 @@ async def _repair_ca(content_artifact, repaired=None):
 
 
 def _verify_artifact(artifact):
+    domain = get_domain()
+    storage = domain.get_storage()
```
mdellweg, line 100: `domain = get_domain()` this, I believe is accessing the context var. So no db call.
Is `storage = domain.get_storage()` also always just memory lookup?
I'd believe repair is a task, and tasks are "run in a domain". Maybe we can even reuse the instance of storage across the whole repair task.

pulpcore/app/tasks/repository.py
```diff
@@ -96,10 +96,11 @@ async def _repair_ca(content_artifact, repaired=None):
...
     try:
-        # verify files digest
         hasher = hashlib.sha256()
-        with artifact.file as fp:
+        with storage.open(artifact.file.name) as fp:
```
mdellweg, line 103: Does this mean, the artifacts `file` field itself does not "know" what storage it is attached to? I always assumed that.

pulpcore/app/tasks/repository.py
mdellweg: What if we added an `assert artifact.domain_id == domain.pk` here?

pulpcore/app/tasks/repository.py
dkliban: That's right. If you ask it for it's URL it will give you the one for the default storage. Should I file a bug about this?

mdellweg: Are we doing this?
https://docs.djangoproject.com/en/4.2/topics/files/#using-a-callable

dkliban: We are not using a callable but we are setting storage to DomainStorage.

https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/content.py#L217
https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/storage.py#L127-L135

pulpcore/app/tasks/repository.py
mdellweg: So either the task does not have a proper domain, or we fail to set it in the task setup code...

mdellweg: How do you feel about keeping the assert here?

pulpcore/app/tasks/repository.py
```diff
@@ -96,10 +97,12 @@ async def _repair_ca(content_artifact, repaired=None):
...
 
 def _verify_artifact(artifact):
+    domain = get_domain()
+    storage = domain.get_storage()
+    assert domain.pk == artifact.pulp_domain_id
     try:
-        # verify files digest
         hasher = hashlib.sha256()
-        with artifact.file as fp:
+        with storage.open(artifact.file.name) as fp:
```
mdellweg, line 105: I believe we don't need this change anymore then.

dkliban: I removed it. Let's see what the tests tell us now.

mdellweg (approved):

gerrod3 (approved):

PR #4836, opened 2023-12-04
Add missing database indices to RemoteArtifact

### LINKED ISSUE #4835: Missing database indices for the checksum fields in `RemoteArtifact`
**Version**
pulpcore 3.21+

**Describe the bug**
The checksum fields in the `RemoteArtifact` model are missing a `db_index` compared to the `Artifact` model.

**Expected behavior**
To improve query performance it would be great to have access to a index here as well.

hstct:
closes #4835


mdellweg (approved):
LGTM

dralley (approved):

hstct:
@dralley we noticed this in `pulp_deb` on large repositories which are synced `on_demand` and have lots of entries in the remote artifacts table. Because we are doing this for every package that is going to be published:

https://github.com/pulp/pulp_deb/blob/main/pulp_deb/app/serializers/content_serializers.py#L485

Using a manual index in postgres did improve the performance significant.

hstct:
The CI failures are I believe due  to the removal of the old gpg keys in the fixtures: https://github.com/pulp/pulp-fixtures/commit/c2394fae8345982c9eb54cc782dfa4a4bf97d669

dralley:
@mdellweg I'm assuming that the issue here is that there hasn't been a release of pulp-cli since the PR merged (or since October, in general), and pulp-cli is being installed from PyPI rather than from source.  So a release of the CLI ought to fix this.

The instance of the key found in this function where the error is occurring is already changed in the current main branch.

Issue #4837, opened 2023-12-04
Orphan cleanup doesn't properly cleanup content with subrepos


gerrod3:
**Version**
pulpcore 3.33+

**Describe the bug**
Orphan cleanup no longer cleans up the sub-content found in content with subrepos since: https://github.com/pulp/pulpcore/pull/4295

**To Reproduce**
Do an RPM kickstart fixture sync, delete the repository and perform orphan cleanup.

PR #4841, opened 2023-12-04
Fix orphan cleanup failing to fully delete content with subrepos

gerrod3:
Tested this out with RPM kickstart fixture and I now see all 69 synced content units being deleted correctly (before it was only ~5). 

@ipanova I tried to still maintain your reliability changes, so hopefully it is still tolerant. :pray: 

fixes: #4837

### DIFF
--- a/pulpcore/app/tasks/orphan.py
+++ b/pulpcore/app/tasks/orphan.py
@@ -3,6 +3,7 @@ import gc
 from logging import getLogger
 
 from django.conf import settings
+from django.db.models import Q
 from django.db.models.deletion import ProtectedError
 from django.utils import timezone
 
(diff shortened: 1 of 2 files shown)


ipanova:
We should stop calling those repos as subrepos. I needed to twist my brain slightly to understand what feels wrong here. Having a subrepo implies that when repo is being removed it's subrepo is going to be deleted as well. 
Historically we had the field name as subrepo, but we have renamed it to user_hidden because that user_hidden repo does not necessarily need to have a relation to the main repo, as it has been proven in rpm plugin. I could see a plugin that would be having an FK from main repo to the user_hidden one and it would just cascade delete it. However for the plugins that have the content --> user_hidden repo relation (and here comes the question) -  is it really intuitive for core to run recursively orphan clean up?
Or should it be plugin's responsibility to a) if such content is being removed from main repo, identify the relations and remove hidden_repo b) if main repo is being removed identify whether such content is present and remove its hidden_repo?

I understand that what you're fixing now is a regression because of change in behavior, but i really wonder if from the very start the logic was correct and if core should account or not for such plugin's decisions ( linking hidden_repo to repo or to content)

pulpcore/app/tasks/orphan.py
```diff
@@ -49,32 +50,44 @@ def orphan_cleanup(content_pks=None, orphan_protection_time=settings.ORPHAN_PROT
...
+                            skipped_content_batch += 1
+                            excluded.add(c.pk)
+                progress_bar.increase_by(count - skipped_content_batch)
+                skipped_content += skipped_content_batch
+            # Have protection_time take into account current task runtime for next deletion loop
+            if protection_time := orphan_protection_time:
+                tdelta = timezone.now() - start_time
+                protection_time += tdelta.seconds
+            content = Content.objects.orphaned(protection_time, content_pks).exclude(
+                Q(pulp_type=PublishedMetadata.get_pulp_type()) | Q(pk__in=excluded)
```
ipanova, line 89: do you really think `Q(pk__in=excluded)` is needed? If we got a ProtectedError earlier it means it will no longer appear in the orphaned() query anyway

gerrod3:
Closing because couldn't find an easy way to have this compatible with https://github.com/pulp/pulpcore/pull/4668.

PR #4849, opened 2023-12-07
Update CI

gerrod3:
Testing that requirements for django-storages and lowerbounds are properly installed.


gerrod3:
Going to close and wait for the auto-ci job.

PR #4851, opened 2023-12-08
WIP

mdellweg:
This should eventually make the pytest plugins more lightweight until actually used, and prevent import errors, when bindings that are not supposed to be used are not installed.


pulpcore/tests/functional/__init__.py
```diff
@@ -169,202 +121,334 @@ def _patch_cid_user_agent(_api_client_set, cid, monkeypatch):
...
-def pulpcore_client(_api_client_set, bindings_cfg):
-    api_client = ApiClient(bindings_cfg)
-    _api_client_set.add(api_client)
-    yield api_client
-    _api_client_set.remove(api_client)
+def pulpcore_bindings(_api_client_set, bindings_cfg):
+    """
+    A namespace providing preconfigured pulpcore api clients.
+
+    e.g. `pulpcore_bindings.WorkersApi.list()`.
```
mdellweg, line 128: This will also bring us away from the questionable practice of giving these apis new names. (I mean the generated names are not always great, but they are there to be anyway.)

pulpcore/tests/functional/gpg_ascii_armor_signing_service.py
```diff
@@ -87,18 +82,13 @@ def _sign_with_ascii_armored_detached_signing_service(filename):
...
-                return await response.text()
-
-    private_key_data = asyncio.run(download_key())
+    response = requests.get(PRIVATE_KEY_URL)
+    response.raise_for_status()
 
     gpg = gnupg.GPG(gnupghome=signing_gpg_homedir_path)
-
-    gpg.import_keys(private_key_data)
+    gpg.import_keys(response.content)
```
dralley, line 91: :+1:

pulpcore/tests/functional/__init__.py
```diff
@@ -515,7 +599,14 @@ def _gen_fixture_server(fixtures_root, ssl_ctx):
 
 
 @pytest.fixture
-def http_proxy(fixtures_cfg, unused_port):
+def _proxy_module():
+    import proxy
+
+    return proxy
```
gerrod3, line 605: Should this be a session fixture?

pulpcore/tests/functional/__init__.py
```diff
@@ -839,20 +936,21 @@ def delete_orphans_pre(request, orphans_cleanup_api_client, monitor_task):
...
-def monitor_task(tasks_api_client, pulp_domain_enabled):
+def monitor_task(pulpcore_bindings, pulp_domain_enabled):
     """
     Wait for a task to reach a final state.
 
     Returns the task in "completed" state, or throws a `PulpTaskTimeoutError` in case the timeout
     in seconds (defaulting to 30*60) exceeded or a `PulpTaskError` in case it reached any other
     final state.
     """
+    from pulpcore.client.pulpcore import ApiException
```
gerrod3, line 947: If the bindings are in `pulpcore_bindings` then I don't think we need this import.

pulpcore/tests/functional/utils.py
```diff
@@ -40,6 +40,21 @@ def __init__(self, task_group):
...
 
+class BindingsNamespace:
+    def __init__(self, module, client):
+        self.module = module
+        self.client = client
+        self.ApiException = self.module.exceptions.ApiException
+
+    def __getattr__(self, name):
+        # __getattr__ is only consulted if nothing is found in __dict__.
+        assert name.endswith("Api")
```
gerrod3, line 51: The only problem with this assert is that the API classes can have custom names if the view is using drf_spectacular.

gerrod3 (commented):
I like the new approach.

pulpcore/tests/functional/__init__.py
mdellweg: Oh absolutely.

pulpcore/tests/functional/utils.py
mdellweg: Do you have an example?
I can remove it...

pulpcore/tests/functional/__init__.py
```diff
@@ -839,20 +937,21 @@ def delete_orphans_pre(request, orphans_cleanup_api_client, monitor_task):
...
 
     def _monitor_task(task_href, timeout=TASK_TIMEOUT):
         task_timeout = int(timeout / SLEEP_TIME)
         for dummy in range(task_timeout):
             try:
-                task = tasks_api_client.read(task_href)
+                task = pulpcore_bindings.TasksApi.read(task_href)
             except ApiException as e:
```
gerrod3, line 955:
```suggestion
            except pulpcore_bindings.ApiException as e:
```

pulpcore/tests/functional/utils.py
mdellweg: @gerrod3 I have yet to see one example...

dkliban (approved):

bmbouter (approved):
This looks good to me. I like how it simplifies things while adding the lazy loading support.

PR #4885, opened 2024-01-03
Integrate pulp-docs with staging-docs CI script

pedro-psb:
What this does:

- Add base/demo docs content on `/staging_docs` (index.md, overviews, etc)
- Update staging_docs publishing script to use `pulp-docs build` from [`pulp-docs`](https://github.com/pedro-psb/pulp-docs) package.

I've updated `pulp-docs` to fetch directly from main (not using the latest GH release), because otherwise we would have to wait each release to test if everything is okay. Regardless of the download method/strategy, if it puts the files in the right directory everything else should be the same.

### DIFF
--- a/pulpcore/__init__.py
+++ b/pulpcore/__init__.py
@@ -1,3 +1,6 @@
+"""
+Base pulpcore module.
+"""
 from pkgutil import extend_path
 
 __path__ = extend_path(__path__, __name__)
--- a/staging_docs/sections/development/guides/How to bind analysis trade..md
+++ b/staging_docs/sections/development/guides/How to bind analysis trade..md
@@ -1,0 +1,43 @@
+---
+repository: 
+tags: ['guide', 'developer']
+versioning:
+  created: 0.1.0
+  updated: []
+  deprecated: 
+---
+
+# How to bind analysis trade.
+
+Pick right fall age discuss idea reveal. Produce more take particularly debate. Everybody control painting throw improve line. Race matter floor moment study force church. Behind ask close. Four enjoy traditional network look. Blood spend this country my. Worker pay word law.
+
+
+1. generate officer bit race.
+1. delete student crime teacher.
+1. do south free not.
+1. create seat into he pay.
+
+---
+
+
+## 1 - generate officer bit race.
+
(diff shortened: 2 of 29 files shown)


pedro-psb:
The lint error suggests that I add `/staging-docs/*.md` to `MANIFEST.in`, but that sounds odd.
For instance, `/docs` content is not there.

pulpcore/__init__.py
pedro-psb, line 1: This was for testing `mkdocstrings`

staging_docs/sections/development/guides/How to bind analysis trade..md
```diff
@@ -0,0 +1,43 @@
+---
+repository: 
+tags: ['guide', 'developer']
+versioning:
+  created: 0.1.0
+  updated: []
+  deprecated: 
+---
+
+# How to bind analysis trade.
```
dralley, line 10: I'm guessing this is a bunch of automatically generated text for the purposes of testing - should this PR be marked as a draft then?

pedro-psb: It is mostly generated data, but the purpose of this is to test if the build script can arrange the file-hierarchy/navigation correctly (which requires files).

So it's not really supposed to be a draft, is kinda a sanity check for the integration with Dennis CI script (to see if it goes live correctly when the workflow runs).

It didnt require that many files, but that's just what I was using at the demo before, so it was easier to keep.

dkliban (approved):

bmbouter (approved):
This is great starter content. +1 Thanks!

staging_docs/sections/development/guides/How to bind analysis trade..md
dralley: I mostly just meant - did we want to actually commit the generated text?

bmbouter: I approved it to get it going, and we'll be replacing it wholesale here soon. We are going to be reviewing the built-site at tomorrow's docs planning meeting.

dralley: :+1:

Issue #4917, opened 2024-01-10
Impossible to use default gunicorn features, like config files


sdherr:
The current way that you create the Gunicorn applications in `app/entrypoint.py` and `content/entrypoint.py` only allow you to pass in a small set of pre-defined click options, and short-circuits the normal Gunicorn process of loading other options from a config file, among many other unavailable options.

By default Gunicorn would normally load a "gunicorn.conf.py" file in the same cwd as the process if it exists. If we re-enable that then any Gunicorn options or other complex settings you didn't want to directly support could be set there.

(In particular I have a desire to filter health-check pings out of the access logs, which you can do with default Gunicorn by adding a `logging.Filter` to `server.log.access_log` in an `on_startup` hook. Something I wouldn't expect you to directly support, but can be done easily by adding a Gunicorn config file.)

PR #4918, opened 2024-01-10
Load gunicorn defaults before overriding with app settings

sdherr:
fixes: #4917

### DIFF
--- a/CHANGES/4917.feature
+++ b/CHANGES/4917.feature
@@ -1,0 +1,1 @@
+Enabled the gunicorn applications for ``pulpcore-api`` and ``pulpcore-content`` to load configurations from the "gunicorn.conf.py" file.
--- a/pulpcore/app/entrypoint.py
+++ b/pulpcore/app/entrypoint.py
@@ -9,9 +9,9 @@ from django.conf import settings
 from django.db import connection
 from django.db.utils import InterfaceError, OperationalError
 from gunicorn.workers.sync import SyncWorker
-from gunicorn.app.base import BaseApplication
 
 from pulpcore.app.apps import pulp_plugin_configs
+from pulpcore.app.pulpcore_gunicorn_application import PulpcoreGunicornApplication
 
 logger = getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


pulpcore/app/entrypoint.py
```diff
@@ -80,12 +80,22 @@ def run(self):
...
+        """
+        A hook for setting application-specific configs, which we instead do below in load_config
+        where it's non-overridable.
+        """
+        pass
+
     def load_config(self):
+        # Load default gunicorn configs, including reading from the default config file.
+        super().load_config()
+        # Override with settings that we've specified in the startup script.
```
mdellweg, line 98: I'm missing that we prevent the use of `--threads`. And maybe there are more dangerous options?

Do you think we could even reuse the gunicorn commandline parsing at this point?

mdellweg: > Is this the reason you moved _away_ from the gunicorn command line parsing that you used to have, because you wanted to enforce `threads=1` and the worker_class? I personally would have just hardcoded those options in the init script and called it a day.

Honestly I don't fully remember. I think is was roughly that.

I think, I preferred failing if `threads` or `worker_class` were specified somewhere. Silently ignoring something a user specified is not too nice. But maybe there's a usecase where a general config file is supposed to provide default options, and we should allow that. We could raise a warning though in case anyone wonders.

sdherr: Okay, check now. I've added warning / error messaging if the user is trying to set something in the config file that will not take effect / cannot be changed.

mdellweg (approved):

mdellweg (dismissed):

CHANGES/4917.feature
```diff
@@ -0,0 +1 @@
+Enable the gunicorn applications for pulp-api and pulp-content to load configs from a "gunicorn.conf.py" file.
```
lubosmj, line 1:
```suggestion
Enabled the gunicorn applications for ``pulpcore-api`` and ``pulpcore-content`` to load configurations from the gunicorn.conf.py file.
```

sdherr: Done.

lubosmj (approved):

Issue #4923, opened 2024-01-15
async-timeout no longer pulled in by aiohttp when using Python 3.11


quba42:
**Version**
I suspect it is all versions using the following PR (https://github.com/pulp/pulpcore/pull/4902) and running Python 3.11.

**Describe the bug**
We found it running an oci-env command in an oci-env using Python 3.11:
```
$ oci-env poll --wait 15 --attempts 30
++ oci-env poll --wait 15 --attempts 30
2024-01-11 16:34:07,397 - INFO - utils.py:36 - USING /home/gitlab-runner/builds/wwZGoy39D/0/orcharhino/pulp_components/pulpcore/oci_env FOR OCI_ENV_PATH BASED ON GIT CMD OUTPUT
2024-01-11 16:34:07,402 - INFO - utils.py:36 - USING /home/gitlab-runner/builds/wwZGoy39D/0/orcharhino/pulp_components/pulpcore/oci_env FOR OCI_ENV_PATH BASED ON GIT CMD OUTPUT
2024-01-11 16:34:07,411 - INFO - utils.py:36 - USING /home/gitlab-runner/builds/wwZGoy39D/0/orcharhino/pulp_components/pulpcore/oci_env FOR OCI_ENV_PATH BASED ON GIT CMD OUTPUT
2024-01-11 16:44:59,932 - ERROR - utils.py:41 - Failed to start [oci_env_pulp_1] http://localhost:5001Traceback (most recent call last):
  File "/usr/local/bin/dynaconf", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/lib/python3.11/site-packages/dynaconf/vendor/click/core.py", line 217, in __call__
    def __call__(A,*B,**C):return A.main(*B,**C)
                                  ^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/dynaconf/vendor/click/core.py", line 201, in main
    H=E.invoke(F)
      ^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/dynaconf/vendor/click/core.py", line 340, in invoke
    E,G,D=B.resolve_command(A,D);A.invoked_subcommand=E;Command.invoke(B,A);C=G.make_context(E,D,parent=A)
                                                        ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/dynaconf/vendor/click/core.py", line 284, in invoke
[...]

PR #4924, opened 2024-01-15
Fix async_timeout import based on python version

git-hyagi:
fixes: #4923

### DIFF
--- a/pulpcore/app/tasks/analytics.py
+++ b/pulpcore/app/tasks/analytics.py
@@ -3,7 +3,12 @@ import json
 import logging
 
 import aiohttp
-import async_timeout
+
+# For Python >= 3.11 use timeout from asyncio:
+try:
+    from asyncio import timeout
+except ImportError:
+    from async_timeout import timeout
 
 from asgiref.sync import sync_to_async
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


quba42:
Cherry-pick this into our environment. Pipeline is running and should take about 12 Minutes if successful.

pulpcore/app/tasks/analytics.py
```diff
@@ -3,7 +3,11 @@
 import logging
 
 import aiohttp
-import async_timeout
+
+try:
+    from asyncio import timeout
+except ImportError:
+    from async_timeout import timeout
```
quba42, line 10: I recommend adding a comment explaining this. Perhaps something like:
```suggestion
# For Python >= 3.11 use timeout from asyncio:
try:
    from asyncio import timeout
except ImportError:
    from async_timeout import timeout
```

quba42 (approved):
I successfully tested these changes using the same pipeline that first revealed the issue.

This was using an oci-env running Python 3.11 and using the 3.39 branch. The pulpcore test suite passed successfully in this setup.

pulpcore/app/tasks/analytics.py
```diff
@@ -3,7 +3,11 @@
 import logging
 
 import aiohttp
-import async_timeout
+
+try:
```
mdellweg, line 7: I would have bet on flake8 to complain that imports need to be above all code, but hey if it says we are fine, who am I to complain?

mdellweg (approved):

lubosmj (approved):

mdellweg (approved):

Issue #4932, opened 2024-01-16
ISE when trying to publish using repo version with rbac


jlsherrill:
**Version**
pulp-ubi d896d2d 

**Describe the bug**

While using RBAC, when trying to do a publish with a repository version, we get an error:

```
pulp [46b2ec908104420dab665feede8b2198]: django.request:ERROR: Internal Server Error: /pulp/46a30c97/api/v3/publications/rpm/rpm/
Traceback (most recent call last):
File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 55, in inner
response = get_response(request)
File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 197, in _get_response
response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/local/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 56, in wrapper_view
return view_func(*args, **kwargs)
File "/usr/local/lib/python3.8/site-packages/rest_framework/viewsets.py", line 125, in view
return self.dispatch(request, *args, **kwargs)
File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 509, in dispatch
response = self.handle_exception(exc)
File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 469, in handle_exception
self.raise_uncaught_exception(exc)
File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
raise exc
File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 497, in dispatch
self.initial(request, *args, **kwargs)
File "/usr/local/lib/python3.8/site-packages/pulpcore/app/viewsets/base.py", line 300, in initial
super().initial(request, *args, **kwargs)
File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 415, in initial
self.check_permissions(request)
File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 332, in check_permissions
if not permission.has_permission(request, self):
File "/usr/local/lib/python3.8/site-packages/rest_access_policy/access_policy.py", line 69, in has_permission
allowed = self._evaluate_statements(statements, request, view, action)
File "/usr/local/lib/python3.8/site-packages/rest_access_policy/access_policy.py", line 113, in _evaluate_statements
matched = self._get_statements_matching_conditions(
[...]

PR #4933, opened 2024-01-16
Casts a repository object before checking permission on it.

dkliban:
fixes: #4932

### DIFF
--- a/CHANGES/4932.bugfix
+++ b/CHANGES/4932.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug where publications couldn't be created from repository versions when using RBAC.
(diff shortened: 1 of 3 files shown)


pulpcore/app/global_access_conditions.py
```diff
@@ -426,7 +426,7 @@ def has_repo_or_repo_ver_param_model_or_domain_or_obj_perms(request, view, actio
     if repository := serializer.validated_data.get("repository"):
         return request.user.has_perm(permission, repository)
```
gerrod3, line 427: I think we should add a cast here as well. It is not guaranteed that every view's serializer will cast the repository.

CHANGES/4932.bugfix
```diff
@@ -0,0 +1 @@
+Fixed a bug where publications couldn't be created from repository versions when usig RBAC.
```
gerrod3, line 1:
```suggestion
Fixed a bug where publications couldn't be created from repository versions when using RBAC.
```

gerrod3 (changes requested):

pulpcore/app/global_access_conditions.py
dkliban: sounds good

gerrod3 (approved):

ggainey (approved):

PR #4936, opened 2024-01-17
fix: Bump dynaconf on 3.28.z

rochacbruno:
AAP 2.4 uses pulpcore 3.28.z and needs
dynaconf to be at least 3.2.4 to include
a fix for boolean parsing.

Previous versions of dynaconf would parse

```bash
export PULP_DEBUG=False
```

as a str `"False"` causing `if settings.DEBUG:` to evaluate to `True`

On 3.2.2, dynaconf [started parsing](https://github.com/dynaconf/dynaconf/pull/983) `False/True` string as booleans from envvars.

[noissue]


mdellweg:
You may want to rebase.

rochacbruno:
@mdellweg rabased, CI still failing on a sync test

mdellweg (approved):

PR #4951, opened 2024-01-22
Taught https-download to trust system certstore.

### LINKED ISSUE #3036: HTTPs Proxies not working.

ggainey:
fixes #3036.

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/test_proxy.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_proxy.py
@@ -4,6 +4,7 @@ from pulpcore.tests.functional.utils import PulpTaskError
 from pulpcore.client.pulp_file import (
     RepositorySyncURL,
 )
+import sys
 
 
 def _run_basic_sync_and_assert(
@@ -154,6 +155,37 @@ def test_sync_http_through_https_proxy(
         policy="on_demand",
         proxy_url=https_proxy.proxy_url,
         tls_validation="false",  # We instead should have a `proxy_insecure` option
+    )
+
+    _run_basic_sync_and_assert(
+        remote_on_demand,
+        file_repo,
+        file_bindings,
+        file_content_api_client,
+        monitor_task,
+    )
+
+
+@pytest.mark.parallel
+def test_sync_https_through_https_proxy(
+    file_remote_ssl_factory,
+    file_repo,
+    file_bindings,
+    file_content_api_client,
+    https_proxy,
+    basic_manifest_path,
+    monitor_task,
+):
+    """
+    Test syncing http through an https proxy.
+    """
+    if not (sys.version_info.major >= 3 and sys.version_info.minor >= 11):
+        pytest.skip("HTTPS proxy only supported on python3.11+")
+    remote_on_demand = file_remote_ssl_factory(
+        manifest_path=basic_manifest_path,
+        policy="on_demand",
+        proxy_url=https_proxy.proxy_url,
+        tls_validation="false",
     )
 
     _run_basic_sync_and_assert(
--- a/CHANGES/3036.bugfix
+++ b/CHANGES/3036.bugfix
@@ -1,0 +1,1 @@
+Taught downloader to trust system-cert-store on HTTPS proxy connections.
--- a/pulpcore/download/factory.py
+++ b/pulpcore/download/factory.py
@@ -120,6 +120,8 @@ class DownloaderFactory:
             sslcontext.verify_mode = ssl.CERT_NONE
         if sslcontext:
             tcp_conn_opts["ssl_context"] = sslcontext
+            # Trust the system-known CA certs, not just the end-remote CA
+            sslcontext.load_default_certs()
 
         headers = MultiDict({"User-Agent": DownloaderFactory.user_agent()})
         if self._remote.headers is not None:


mdellweg (dismissed):

dralley:
@bmbouter @dkliban  Do you have any objections to the current approach, vaguely remember some discussion about it but I don't recall if there was a specific objection.  The only thing I can find written down is a different approach suggestion https://bugzilla.redhat.com/show_bug.cgi?id=1993917#c91

ggainey:
> @bmbouter @dkliban Do you have any objections to the current approach, vaguely remember some discussion about it but I don't recall if there was a specific objection. The only thing I can find written down is a different approach suggestion https://bugzilla.redhat.com/show_bug.cgi?id=1993917#c91

The showstopper, iirc, was that prior to python-3.11 we had to set a protected attribute - see the discussion here : https://github.com/pulp/pulpcore/pull/3038#discussion_r940569697

With py-3.11, that's now the default and we don't have to hack around it.

ggainey:
> I thought there was also some objection about the loading of system certs, maybe [#3038 (comment)](https://github.com/pulp/pulpcore/pull/3038#discussion_r940571420)?

I have a memory of a discussion, that I don't think got written down anywhere alas. We make the user give us the CA for a Remote, because "the system" pulp is installed on may not want to trust that connection globally. But Pulp should be allowed to talk to anything "the system" has already decided is legal/allowed, and shouldn't need to "double accept" something the system-admin has already added to the global keystore.

Does anyone else remember that conversation, or did I dream it?

dralley:
I have vague memory of this discussion, yes.  I don't recall the outcome though.

bmbouter:
I could imagine folks not wanting this always on. I'm guessing here because I'm not one of those folks, but say you generally don't trust the thousands of CAs that are typically loaded...

bmbouter:
That being said, we should have a way to enable this for sure. This is highly valuable for almost everyone.

ggainey:
> I could imagine folks not wanting this always on. I'm guessing here because I'm not one of those folks, but say you generally don't trust the thousands of CAs that are typically loaded..

In this context, pulp has been told "use this proxy", and that proxy's CA has been defined as trusted by the environment pulp is installed in. I can't actually think of a "hat" to wear, where asking pulp to extra-double-trust the proxy is actually useful.

dralley:
Well, you could extend that logic to say "we've told Pulp to sync this remote, and the CA used by this remote's URL has been defined as trusted by the environment".  Historically though, we've asked that the certificate to verify against be explicitly provided.

I actually think the same logic *does* make sense here, and that we probably should trust system certs automatically when applied to remotes in general, but that's not historically what we've done?

ggainey:
> Well, you could extend that logic to say "we've told Pulp to sync this remote, and the CA used by this remote's URL has been defined as trusted by the environment". Historically though, we've asked that the certificate to verify against be explicitly provided.
> 

My contention is (and has been) that proxy-trust is qualitatively different than remote-trust. The Environment pulp is running is, has declared "this proxy can be trusted" - that's not the pulp-admin's purview. For the remotes being sync'd, **only pulp cares** - which is why we have the admin set the CA for the remote.

> I actually think the same logic _does_ make sense there, and that we probably _should_ trust system certs automatically when applied to remotes in general (maybe unless a specific one is provided?), but that's not historically what we've done

Basically, if it's in the system-keystore, we should prob trust it, even for Remotes. But as you say, that isn't what we've been doing - and that is a **different question** than "do we trust the system's definition for This Proxy".

dralley:
It's a different question but it would still be nice if we have consistent logic / rules about this.  

Anyway, your explanation makes some sense and in any case I don't consider consistency necessarily a blocker for this PR

dralley (dismissed):

dralley:
I think I'm OK with this on the basis of the discussion.  Any objections?

I believe we also want to backport this to 3.39 as downstream needs this, but that's perhaps a bit murky.  It's not a *breaking* change but it is perhaps a reasonably significant one?

pulpcore/tests/functional/api/using_plugin/test_proxy.py
```diff
@@ -163,3 +164,34 @@ def test_sync_http_through_https_proxy(
...
+    file_bindings,
+    file_content_api_client,
+    https_proxy,
+    basic_manifest_path,
+    monitor_task,
+):
+    """
+    Test syncing http through an https proxy.
+    """
+    if not sys.version_info.minor >= 11:
```
dralley, line 182: I'd prefer this checked both components of the version

ggainey:
> I think I'm OK with this on the basis of the discussion. Any objections?
> 
> I believe we also want to backport this to 3.39 as downstream needs this, but that's perhaps a bit murky. It's not a _breaking_ change but it is perhaps a reasonably significant one?

Def wants to be backported to 3.39 - it's not a New Feature, so much as Removing a Restriction that was imposed by a combination of python<3.11 and older aiohttp.

dralley (approved):

bmbouter:
My claim is probably inconvenient, but this change is really significant and I don't think it's good to backport. Imagine that you upgrade Pulp and now all of a sudden your environment trusts thousands of CAs. That's just my opinion though, don't consider it a blocking statement if you both still want to proceed.

ggainey:
> I think this kind of a feature should be feature flagged on a remote itself. Say you are using `ca_cert` and your expectation is this is the only thing I trust. Now you look at your remote and you will trust that a bunch of other things. Users and admins have different expectations around what to trust.

I contend that generally proxies aren't trusted on a per-destination basis - every outgoing connection from a system has to go through the same proxy or it's disallowed. And every SSL connection that leaves your system is going to trust what's in the system's CAstore, unless the app you're using very specifically decides *NOT* to do that.

> Side question: How do users even know what the are trusting when they have no access to the system store. So an admin decides things and then users also must trust it? It's just confusing, but maybe it's ok? Just some random thoughts here.

Your system trusts everything in `/usr/share/pki/ca-trust-source/` - every time you use curl or wget or httpie with an HTTPS URL, and it works, that's why. That's the store we are trusting.

The common use-case is your network admin decrees "you will use the following proxy for all requests, or they will never leave your machine". In real environments, proxy-use is imposed on systems at a higher level than per-application.

Browsers have only relatively recently started carrying their own certstores. This was a response to browsers running on end-user machines that have no administrators, and which can end up with rootkits that install bogus certs. (Of course, the Bad Guys now just install their certs into the browser certstores when they get access, **in addition to** the system-one...)

bmbouter:
I think I misunderstood what was going on here, and you've set me straight. Pulp doesn't trust the system trust store, but it does trust `certifi`. Given that he scope of what is changing here is way smaller than I thought. +1 to proceeding.

mdellweg (approved):

PR #4980, opened 2024-01-26
Pin black to 23.12.1.

ggainey:
We'll let dependabot tell us when black updates, which will let us upgrade/reformat "deliberately" instead of "by surprise".

[noissue]


daviddavis (approved):

ggainey:
Duh, lint_requirements is under the control of plugin_template, closing

Issue #4976, opened 2024-01-25
An internal server error is raised when using a lengthy name while creating a domain


lubosmj:
Steps to reproduce:
```
http :5001/pulp/default/api/v3/domains/ name=0000000000000000000000000000000000000000000000000000000000000000000000000000000000 storage_class=pulpcore.app.models.storage.FileSystem storage_settings:="{\"MEDIA_ROOT\": \"/var/lib/pulp/media/\"}"
```

```
 File "/usr/local/lib/python3.8/site-packages/psycopg/cursor.py", line 732, in execute
    raise ex.with_traceback(None)
psycopg.errors.StringDataRightTruncation: value too long for type character varying(50)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 56, in wrapper_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/rest_framework/viewsets.py", line 125, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/rest_framework/mixins.py", line 19, in create
    self.perform_create(serializer)
  File "/usr/local/lib/python3.8/site-packages/rest_framework/mixins.py", line 24, in perform_create
    serializer.save()
  File "/src/pulpcore/pulpcore/app/serializers/base.py", line 458, in save
    return super().save(**kwargs)
  File "/usr/local/lib/python3.8/site-packages/rest_framework/serializers.py", line 212, in save
[...]

PR #4988, opened 2024-01-26
Verify if domain name has more than 50 chars

git-hyagi:
fixes: #4976

### DIFF
--- a/pulpcore/app/serializers/domain.py
+++ b/pulpcore/app/serializers/domain.py
@@ -298,6 +298,7 @@ class DomainSerializer(ModelSerializer):
 
     pulp_href = IdentityField(view_name="domains-detail")
     name = serializers.SlugField(
+        max_length=50,
         help_text=_("A name for this domain."),
         validators=[UniqueValidator(queryset=models.Domain.objects.all())],
     )
--- a/pulpcore/tests/functional/api/test_crud_domains.py
+++ b/pulpcore/tests/functional/api/test_crud_domains.py
@@ -2,6 +2,7 @@ import uuid
 
 import pytest
 import random
+import string
 import json
 from pulpcore.client.pulpcore import ApiException
 from pulpcore.app import settings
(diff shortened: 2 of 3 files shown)


pulpcore/app/serializers/domain.py
```diff
@@ -324,6 +324,11 @@ def validate_name(self, value):
         """Ensure name is not 'api' or 'content'."""
         if value.lower() in ("api", "content"):
             raise serializers.ValidationError(_("Name can not be 'api' or 'content'."))
+
+        """Ensure name has less than 50 characters."""
+        if len(value) > 50:
+            raise serializers.ValidationError(_("Name can have 50 characters at most."))
+
```
lubosmj, line 331: I suppose we want something neat, like:
```
    name = serializers.SlugField(
        max_length="50",
        help_text=_("A name for this domain."),
        validators=[UniqueValidator(queryset=models.Domain.objects.all())],
    )
```

lubosmj: A confusing part is that they claim `drf.SlugField` has `max_length` set to 50 by default: https://www.django-rest-framework.org/api-guide/fields/#slugfield.

However, there is no such a thing defined in the source code: https://github.com/encode/django-rest-framework/blob/41edb3b9dde2105b9fb015ce6b8171e7822baa30/rest_framework/fields.py#L796.

Therefore, I think setting `max_length=50` explicitly in the field will help.

mdellweg: Good analysis. That sounds like a bug with drf (maybe docs only).

pulpcore/tests/functional/api/test_crud_domains.py
```diff
@@ -268,6 +269,21 @@ def test_special_domain_creation(domains_api_client, gen_object_with_cleanup):
...
+    body = {
+        "name": name,
+        "storage_class": backend,
+        "storage_settings": storage_settings[backend],
+    }
+    try:
+        domain = gen_object_with_cleanup(domains_api_client, body)
+    except ApiException as e:
+        assert e.status == 400
+        assert "Ensure this field has no more than 50 characters." in e.body
```
lubosmj, line 285: Usually, we use `with pytest.raises(ApiException) as e`. But, this works for me too!

lubosmj (dismissed):

pulpcore/tests/functional/api/test_crud_domains.py
mdellweg: And we should do that.

lubosmj (approved):

mdellweg (approved):

PR #4989, opened 2024-01-29
Start emitting disk usage metrics for domains

### LINKED ISSUE #4603: Add additional metrics reporting size of artifacts stored
## Feature

When running on Pulp on cloud services the data storage costs can be expensive. Understanding these costs is key. It would be excellent to know answers to these questions:

* How many Bytes is a RepositoryVersion consuming?
* How many Bytes are all of the RepoVersions consuming for a given Repository?
* How many Bytes are all Repositories of a given domain consuming?
* Will deleting old versions actually help me to reduce my storage costs?

## Proposal

Have the saving of a RepositoryVersion emit this open telemetry data.

* Bytes of the Artifacts that are new to this RepositoryVersion
* Bytes of Artifacts that are removed from this RepositoryVersion
* The Repository name and URL this RepositoryVersion is a member of
* The Domain (if any) this Repository is a member of.

With ^ data, aggregating reports could be run (not part of this ticket) in Prometheus that do things like:

* How many Bytes does all the history of a specific Repository use?
* How many Bytes does all the data in a specific domain use?
* If I deleted N RepositoryVersions how many Bytes would I save?

lubosmj:
closes #4603

### DIFF
--- a/docs/tech_preview.rst
+++ b/docs/tech_preview.rst
@@ -1,0 +1,9 @@
+Tech Previews
+=============
+
+The following features are currently being released as part of tech preview:
+
+* Support for Open Telemetry
+* Upstream replicas
+* Domains - Multi-Tenancy
+* Complex filtering
--- a/pulpcore/app/models/domain.py
+++ b/pulpcore/app/models/domain.py
@@ -1,6 +1,8 @@
+from opentelemetry.metrics import Observation
+
 from django.core.files.storage import get_storage_class, default_storage
 from django.db import models
-from django_lifecycle import hook, BEFORE_DELETE, BEFORE_UPDATE
+from django_lifecycle import hook, BEFORE_DELETE, BEFORE_UPDATE, AFTER_CREATE
 
 from pulpcore.app.models import BaseModel, AutoAddObjPermsMixin
 from pulpcore.exceptions import DomainProtectedError
@@ -60,7 +62,26 @@ class Domain(BaseModel, AutoAddObjPermsMixin):
             # Delete on by one to properly cleanup the storage.
             artifact.delete()
 
+    @hook(AFTER_CREATE)
+    def _report_domain_disk_usage(self):
+        from pulpcore.app.util import DomainMetricsEmitterBuilder
+
+        DomainMetricsEmitterBuilder.build(self)
+
     class Meta:
         permissions = [
             ("manage_roles_domain", "Can manage role assignments on domain"),
+        ]
+
+
+def disk_usage_callback(domain):
+    from pulpcore.app.models import Artifact
+    from pulpcore.app.util import get_url
+
+    options = yield  # noqa
+    while True:
+        distinct_artifacts = Artifact.objects.filter(pulp_domain=domain).distinct()
+        total_size = distinct_artifacts.aggregate(size=models.Sum("size", default=0))["size"]
+        options = yield [  # noqa
+            Observation(total_size, {"pulp_href": get_url(domain), "name": domain.name})
         ]
--- a/staging_docs/admin/learn/tech-preview.md
+++ b/staging_docs/admin/learn/tech-preview.md
@@ -1,0 +1,8 @@
+# Tech Previews
+
+The following features are currently being released as part of tech preview:
+
+- [Support for Open Telemetry](site:pulpcore/docs/admin/learn/architecture/#telemetry-support)
+- Upstream replicas
+- Domains - Multi-Tenancy
+- Complex filtering
--- a/CHANGES/4603.feature
+++ b/CHANGES/4603.feature
@@ -1,0 +1,3 @@
+Started emitting metrics that report disk usage within a domain. The metrics are sent to the
+collector every 60 seconds. The interval can be adjusted with the ``OTEL_METRIC_EXPORT_INTERVAL``
+environemnt variable.
--- a/docs/components.rst
+++ b/docs/components.rst
@@ -158,6 +158,9 @@ Pulp can produce OpenTelemetry data, like the number of requests, active connect
 `pulp-api` and `pulp-content` using OpenTelemetry. You can read more about
 `OpenTelemetry here <https://opentelemetry.io>`_.
 
+.. warning:: This feature is provided as a tech preview and could change in backwards incompatible
+  ways in the future.
+
 If you are using `Pulp in One Container <https://pulpproject.org/pulp-in-one-container/>`_ or `Pulp Operator
 <https://docs.pulpproject.org/pulp_operator/>`_ and want to enable it, you will need to set the following
 environment variables:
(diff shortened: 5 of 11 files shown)


lubosmj:
I am attaching how the end-result looks in Grafana. Metrics are periodically collected on Prometheus.

In the graph, there are two domains considered -- `default` and `OTEL`. I created the `OTEL` domain manually. Then, I synced a repository within that domain with the `file-large` repository. I did the same for the `default` domain. In between, I removed the `OTEL` domain. Finally, I removed content from a repository in the `default` domain and ran the orphan cleanup. The graph correctly shows how the disk usage changes over time.
image

Below, I demonstrate the existence of the selected OTEL domain. Once a user removes the domain, the metrics are no longer exported.
image

lubosmj:
@decko, @dkliban, @bmbouter, I am a bit concerned about the current implementation. Will you help me resolve some of the issues?

1. It takes a long time to export and receive the metrics on the collector. Modifying the obvious [OTEL_EXPORTER_OTLP_METRICS_TIMEOUT](https://github.com/open-telemetry/opentelemetry-python/blob/c4d17e9f14f3cafb6757b96eefabdc7ed4891306/opentelemetry-sdk/src/opentelemetry/sdk/environment_variables.py#L464) setting did not mitigate the problem.
    **SOLUTION: Do not test the metrics in our CI.**
2. When a user restarts the pulpcore-api service, we stop emitting metrics for newly created domains since the initial process is not running anymore. Furthermore, it is not the best idea to initialize the metrics exporter from post_migrate hooks (I have noticed that exporters initialized inside `_ensure_default_domain` are being ignored).
    **SOLUTION: Initialize metrics' instruments inside `_util.startup_hook`, like it is done for the `default` domain now.**
3. The number of workers determines the number of possible duplicates being sent to the collector. In oci-env, we have 2 workers; thus, it sends 2 metrics for the same domain. This is easily digested by Prometheus and no duplicates are preserved in graphs.
  **SOLUTION: We need to ensure that the initialization is run just from pulpcore-api and it is able to recover itself after the restart. Where is the correct place for doing this?**

Any inputs?

decko:
1. Have you tried [OTEL_METRIC_EXPORT_TIMEOUT](https://github.com/open-telemetry/opentelemetry-python/blob/c4d17e9f14f3cafb6757b96eefabdc7ed4891306/opentelemetry-sdk/src/opentelemetry/sdk/environment_variables.py#L632)?

2. I think I'm ok with it, but probably @gerrod3 is the best person to answer that.

3. Do you think we can keep the "emitter" only on the API side?

4. On the first graph I'm seeing that we got two metrics, one emitted by the API and other by the worker. I'm presuming that under the attribute "name" we have the domain, but on the URL it's pointing just for the default domain although the name is "OTEL".

lubosmj:
> 1. Have you tried [OTEL_METRIC_EXPORT_TIMEOUT](https://github.com/open-telemetry/opentelemetry-python/blob/c4d17e9f14f3cafb6757b96eefabdc7ed4891306/opentelemetry-sdk/src/opentelemetry/sdk/environment_variables.py#L632)?

I have just tested it. It did not work either. Seeing the logs inside oci-env, the metrics are emitted every 15 seconds or so. Can you verify this on your side too, please?

> 3. Do you think we can keep the "emitter" only on the API side?

If we keep it on the API side, outside of the `domain.save()` scope, we will need to have a hook somewhere which iterates over existing domains and initialize the emitter every time the process is started. The **somewhere** part is the catch. That is why I resolved to have it inside the AFTER_CREATE hook at the moment (this is the place where we initialize the emitter for newly created domains).

> 4. On the first graph I'm seeing that we got two metrics, one emitted by the API and other by the worker. I'm presuming that under the attribute "name" we have the domain, but on the URL it's pointing just for the default domain although the name is "OTEL".

The URLs are different, pointing to respective domains. The exported job is equal to "worker" because information about the default domain is emitted from https://github.com/pulp/pulpcore/blob/075a37fd5b68c09bddffb3205e2aa5c82e2bb549/pulpcore/tasking/_util.py#L56 (ensuring that if a worker fails or is restarted, we still emit data for the default domain; restarting api-app worker is not so common, but might happen in case of overloaded system).

pulpcore/app/models/domain.py
```diff
@@ -60,7 +69,50 @@ def _cleanup_orphans_pre_delete(self):
...
+
+        _meters[self.name] = meter, instrument
+
     class Meta:
         permissions = [
             ("manage_roles_domain", "Can manage role assignments on domain"),
         ]
+
+
+def disk_usage_callback(domain):
```
decko, line 108: Is it possible to import this from `utils`?

lubosmj: Hopefully. I need to identify circular imports.

lubosmj:
It appears that I can emit data about domains from `wsgi.py` (in the same way I do from the worker process inside the `startup_hook` function). This should be resilient to process restarts and the exporter is always emitting data from api-app, which is exactly what we want! With @dkliban, we also agreed on having the data emitted just from api-app but with a lower frequency (e.g., 5 minutes). I will work on this.

In the future (new PR), we may consider having the metrics emitted on a specific event as initially planned. On the other hand, we would need to find a balance between emitting data when a task finishes and serving on-demand content (we do not want to run heavy DB queries each time content is downloaded and streamed by content-app).

Another problem to deal with is the number of duplicated data. The number of api-app processes determines the number of duplicates. This can be resolved by introducing a new option to the entrypoint command (new PR). So, users will run `pulpcore-api --emit-metrics` to get the metrics exported for a main emitting process.

docs/tech_preview.rst
```diff
@@ -0,0 +1,9 @@
+Tech Previews
+=============
+
+The following features are currently being released as part of tech preview:
+
+* Support for Open Telemetry
+* Upstream replicas
+* Domains - Multi-Tenancy
+* Complex filtering
```
lubosmj, line 9: @pedro-psb, is it worth introducing the same page to the staging docs? How do we keep sync of newly added docs and staging docs?

decko (dismissed):
LGTM 🚀

docs/tech_preview.rst
pedro-psb: Thats a good point. The docs workgroup has discussed this in the early stages of planning and we've agreed that `"All new doc activity will focus on new-docs. Old documentation as is kept as is until process is done."` ([early planning post](https://discourse.pulpproject.org/t/unified-docs-plan/1059?u=pedro-psb)).

The reasons were to keep it simple and with low overhead for us. If this temporary outdated state is not acceptable we can discuss alternatives, the simple ones I can think of are:
- duplicating the content manually until we switch
- adding a banner to the actual website stating that only the new-docs is receiving updates

lubosmj: And, where do you recommend adding this site?

https://staging-docs.pulpproject.org/pulpcore/docs/sections/admin/{tech-previews}?

pedro-psb: "for now" because the tech-preview listing happens in a lot of plugins, so we may think of a better general approach for this later. But this works

dkliban: i don't think we should have a tech-preview section. Some guides/tutorials can just mention that a feature is tech preview.

dkliban: oops. i misunderstood. let's make the page as was suggested above.

lubosmj: Done. See the file `staging_docs/admin/learn/tech-preview.md`. I did not add more hyperlinks because the listed features are still located under the "triage-needed!" directory.

decko (dismissed):
LGTM 🚀

staging_docs/admin/learn/tech-preview.md
```diff
@@ -0,0 +1,8 @@
+# Tech Previews
+
+The following features are currently being released as part of tech preview:
+
+- [Support for Open Telemetry](architecture.md#telemetry-support)
```
dkliban, line 5: The URL should be (site:pulpcore/admin/learn/architecture/#telemetry-support)

lubosmj: This works for me locally. It is not working for you?

dkliban: I have not tried it, but I believe that we want to express all our links using the notation that i provided.

decko (approved):
LGMT 🚀 🚀 🚀

dkliban (approved):

PR #4994, opened 2024-01-30
Update cryptography requirement from >=38.0.1,<42.0.2 to >=41.0.6,<42.0.2

gerrod3:
[noissue]


mdellweg:
Can you explain?
Normally we do not require new lower bound versions on z-stream updates. I would count this as a breaking change.
So at the very least add a changelog entry to warn the user.

gerrod3:
Close as it is not needed.

Issue #5010, opened 2024-01-31
`download_content_unit` test fixture assumes `CONTENT_PATH_PREFIX`


dkliban:
Tests that use `download_content_unit` fail when `CONTENT_PATH_PREFIX` is set to something other than `/pulp/content/` .

PR #5011, opened 2024-01-31
Improves the `download_content_unit` test fixture.

dkliban:
fixes: #5010

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -1091,6 +1091,11 @@ def pulp_content_origin(pulp_settings):
 
 
 @pytest.fixture(scope="session")
+def pulp_content_origin_with_prefix(pulp_settings):
+    return pulp_settings.CONTENT_ORIGIN + pulp_settings.CONTENT_PATH_PREFIX[:-1]
+
+
+@pytest.fixture(scope="session")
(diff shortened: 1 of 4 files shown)


pulpcore/tests/functional/__init__.py
```diff
@@ -1043,8 +1043,8 @@ def pulp_domain_enabled(pulp_settings):
 
 
 @pytest.fixture(scope="session")
-def pulp_content_origin(pulp_settings):
```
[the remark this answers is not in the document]
mdellweg: Oh, I see.
We probably want to deprecate the old one before removing.

template_config.yml
```diff
@@ -63,6 +63,7 @@ pulp_settings:
   allowed_import_paths:
   - /tmp
   orphan_protection_time: 0
+  content_path_prefix: /pulp-content/
```
mdellweg, line 66: This is changing from `/pulp/content/`? Let's make it real obvious so when seeing a failing test it really looks different.
How about `/somewhere/else/`?

dkliban: Sure! I'll update and I'll also fix up all the tests that are now failing because of this.

mdellweg (approved):

ggainey (approved):

Issue #4207, opened 2023-07-31
Support base path proxy rewrites


newswangerd:
**Is your feature request related to a problem? Please describe.**

For AAP 2.5 we're looking at re-architecting the platform to use an API gateway. Our end goal is to have a single host that serves the APIs for multiple AAP services (AWX, EDA, Galaxy NG) via a proxy. The proxy needs to be able to support multiple instances of galaxy ng from different API base paths. Ex:

- `/proxy/v1/galaxy1/` proxies to `galaxy1.example.com/api/galaxy/`
- `/proxy/v1/galaxy2/` proxies to `galaxy2.example.com/api/galaxy/`

This is a huge problem for galaxy because all of the pulp HREFs are based on the pulp API base path. With a setup like this, clients will get HREFs that start with /api/galaxy that don't work on the proxy.

Since we're looking to support existing galaxy ng services using this new API gateway, we don't want to force admins to change their API base paths to match the path on the proxy server.

**Describe the solution you'd like**

At the moment we're solving this issue by using regex to replace the HREFs for every response. This is an inefficient and error prone process. I would really like it if the could be handled on the pulp side. It would be great if the pulp base path could be set for a sepcific request using a configurable header such as `X-Forwarded-Prefix`.

**Alternatives**

We've considered using headers to direct traffic on the AAP proxy instead of using base paths (so all galaxy requests would be served from /api/galaxy on the proxy, with a header to select specific galaxy instances), but this is a very sub optimal solution that we would like to avoid.

mdellweg:
As long as we use hrefs as task reserved resources we need to be very careful here.

mdellweg:
Idea from pulpcore meeting:
If we used the independent part of the href (not including the domain ; starting from `/api/v3`) we can probably fill in the prefix from the request to the api. That way the database would store settings independent data.

newswangerd:
For context, this is the Lua script that we have to run on envoy on every request and response that goes the proxy:

``` lua
function envoy_on_response(handle)
    -- This script rewrites urls in the response by replacing the service base path
    -- in Location headers and in the request body.
    -- ex: translates /service-path/galaxy/ to /gateway-path/hub/ in service response bodies
    handle:logDebug("Running lua script to rewrite response body.")

    local location = handle:headers():get("Location")
    local prefix_rewrite = handle:metadata():get("prefix")
    local prefix = handle:metadata():get("prefix_rewrite")

    if location then
        
[...]

PR #5013, opened 2024-02-01
Add ability to have proxy rewrite API_ROOT w/ header

gerrod3:
fixes: #4207

### DIFF
--- a/.github/workflows/scripts/install.sh
+++ b/.github/workflows/scripts/install.sh
@@ -119,7 +119,7 @@ if [ "$TEST" = "azure" ]; then
       - ./azurite:/etc/pulp\
     command: "azurite-blob --blobHost 0.0.0.0"' vars/main.yaml
   sed -i -e '$a azure_test: true\
-pulp_scenario_settings: {"domain_enabled": true}\
+pulp_scenario_settings: {"api_root_rewrite_header": "X-API-Root", "domain_enabled": true}\
 pulp_scenario_env: {"otel_bsp_max_export_batch_size": 1, "otel_bsp_max_queue_size": 1, "otel_exporter_otlp_endpoint": "http://localhost:4318", "otel_exporter_otlp_protocol": "http/protobuf", "otel_metric_export_interval": 800, "pulp_otel_enabled": "true"}\
 ' vars/main.yaml
 fi
--- a/.github/workflows/scripts/pre_before_script.sh
+++ b/.github/workflows/scripts/pre_before_script.sh
@@ -1,8 +1,9 @@
 set -euv
 
-
-
 if [ "$TEST" = "azure" ]; then
+    cmd_stdin_prefix bash -c "cat > /etc/nginx/pulp/api_root_rewrite.conf" < pulpcore/tests/functional/assets/api_root_rewrite.conf
+    cmd_prefix bash -c "s6-rc -d change nginx"
+    cmd_prefix bash -c "s6-rc -u change nginx"
     cmd_stdin_prefix bash -c "cat > /var/lib/pulp/scripts/otel_server.py" < pulpcore/tests/functional/assets/otel_server.py
     cmd_user_prefix nohup python3 /var/lib/pulp/scripts/otel_server.py &
 fi
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -15,7 +15,6 @@ from django.contrib.postgres.fields import HStoreField
 from django.core.validators import MinValueValidator
 from django.db import models, transaction
 from django.db.models import F, Func, Q, Value
-from django.urls import reverse
 from django_lifecycle import AFTER_UPDATE, BEFORE_DELETE, hook
 from rest_framework.exceptions import APIException
 
@@ -23,9 +22,9 @@ from pulpcore.app.util import (
     batch_qs,
     get_prn,
     get_view_name_for_model,
-    get_domain,
     get_domain_pk,
     cache_key,
+    reverse,
 )
 from pulpcore.constants import ALL_KNOWN_CONTENT_CHECKSUMS, PROTECTED_REPO_VERSION_MESSAGE
 from pulpcore.download.factory import DownloaderFactory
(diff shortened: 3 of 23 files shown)


pulpcore/app/models/repository.py
```diff
@@ -1254,8 +1254,7 @@ class RepositoryVersionContentDetails(models.Model):
     )
     count = models.IntegerField()
 
-    @property
-    def content_href(self):
+    def get_content_href(self, request=None):
```
mdellweg, line 1257: This always feels like `request` would be the perfect candidate for a context var...

gerrod3: Do you think I should do so in this PR?

mdellweg: I am a little worried that we are changing a public interface here. But is this class actually public?

pulpcore/app/path_api_urls.py
```diff
@@ -0,0 +1,12 @@
+from django.conf import settings
+from django.urls import path, include
+from pulpcore.app.urls import API_ROOT, docs_and_status, special_views, all_routers
+
+# Add all the Pulp API endpoints again, but with the API_ROOT as a path parameter
+PATH_API_ROOT = "<path:api_root>/" + API_ROOT.split(settings.API_ROOT[1:])[1]
```
mdellweg, line 6: `API_ROOT.split(settings.API_ROOT[1:])[1]` So this looks confusing.
I believe API_ROOT is kind of the domain augmented version on settings.API_ROOT.
But is there not a much more direct constructive way to express this?

pulpcore/app/path_api_urls.py
```diff
@@ -0,0 +1,12 @@
...
+dup_urls = special_views + docs_and_status
+for router in all_routers:
+    dup_urls.extend(router.urls)
```
mdellweg, line 9: Do we really need to dupe the url patterns? Can't we just also handle the default paths with the global match pattern?

mdellweg: Also do plugins with individual urlpattern opt into this? Do we get the whole "domain enabled" story all over again?

pulpcore/app/serializers/base.py
```diff
@@ -44,30 +45,16 @@
...
-
 class HrefFieldMixin:
     """A mixin to configure related fields to generate relative hrefs."""
 
     def get_url(self, obj, view_name, request, *args, **kwargs):
-        # Removes the request from the arguments to display relative hrefs.
-        self.reverse = _reverse(self.reverse, request, obj)
-        return super().get_url(obj, view_name, None, *args, **kwargs)
+        # Use the Pulp reverse method to display relative hrefs.
+        self.reverse = reverse
```
mdellweg, line 53: At least _some_ consolidating codepaths. ;)

pulpcore/app/settings.py
```diff
@@ -113,6 +114,7 @@
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "pulpcore.middleware.DomainMiddleware",
+    "pulpcore.middleware.APIRootRewriteMiddleware",
```
mdellweg, line 117: Didn't you only want to include this when configured?
I saw the raising of MiddlewareNotUsed.

pulpcore/app/models/repository.py
gerrod3: The interface isn't changed though, further down I re-create the `content_href` property and have it call this "new" method.

pulpcore/app/settings.py
gerrod3: It's part of django's middelware api: https://docs.djangoproject.com/en/5.0/topics/http/middleware/#marking-middleware-as-unused. You list it here and then on launch the middleware's init will determine if it should be used or not.

.github/workflows/scripts/install.sh
```diff
@@ -126,7 +126,7 @@ if [ "$TEST" = "azure" ]; then
       - ./azurite:/etc/pulp\
     command: "azurite-blob --blobHost 0.0.0.0 --cert /etc/pulp/azcert.pem --key /etc/pulp/azkey.pem"' vars/main.yaml
   sed -i -e '$a azure_test: true\
-pulp_scenario_settings: {"domain_enabled": true}\
+pulp_scenario_settings: {"api_root_rewrite_header": "X-API-Root", "domain_enabled": true}\
```
mdellweg, line 129: Just wondering, can we infer the original relative path the user sent this request to from the request object?
Or is there some apache/nginx rewrite we going on we cannot see anymore?
(I was wondering, because drf used to use the request object to create uris, where we dumbed it down to do hrefs.)

gerrod3: Don't think so, unless the reverse proxy sends along that info inside some header we are completely clueless where the original path came from.

pulpcore/app/serializers/base.py
gerrod3: Might have to change this back to a patched reverse (albeit using our new Pulp reverse). The tests are failing because `super().get_url` doesn't accept kwargs to pass onto `reverse`, so the domain can't be specified. For serializing fields we need to know if the object has a domain relationship to use the correct domain name, the requests domain is not good enough.

.github/workflows/scripts/pre_before_script.sh
```diff
@@ -1,8 +1,9 @@
 set -euv
 
-
-
 if [ "$TEST" = "azure" ]; then
+    cmd_stdin_prefix bash -c "cat > /etc/nginx/pulp/api_root_rewrite.conf" < pulpcore/tests/functional/assets/api_root_rewrite.conf
```
mdellweg, line 4: Is this related to your recent oci images change with a templated nginx config?

gerrod3: No it is not.

ipanova:
@newswangerd we are going to get it reviewed within the pulp team and merged. Whenever you get to it and have more feedback we would address later then.

mdellweg (commented):
This looks good to me.
I like that we consolidate the domain and the dynamic rerouting into one here.

mdellweg (approved):

Issue #5033, opened 2024-02-05
Add a "updated" date to a distribution


mimmus:
As a rpm packages repos administrator for a company, I manage updates lifecycle like this:

* sync repo with remote
* create a new publication version
* update "DEV" distribution, to point to new version
* apply updates to DEV machines and wait for some time
* update "QA" distribution to point on the same version of "DEV"
* apply updates to QA machines...
* etc

and I'd like to know _when_ I moved forward an environment

**Additional context**
Actually, only "pulp_created" is available on a distribution and this date is not updated when I update the distribution to point to a newer publication version.

ipanova:
We have also `pulp_last_updated` field we just need to serialize it

```
In [7]: disto.__dict__
Out[7]: 
{'_state': <django.db.models.base.ModelState at 0x7fddb1418580>,
 'pulp_id': UUID('018bda21-e2e5-711e-9c68-3860678d21ee'),
 'pulp_created': datetime.datetime(2023, 11, 16, 21, 55, 35, 782730, tzinfo=datetime.timezone.utc),
 'pulp_last_updated': datetime.datetime(2023, 11, 16, 21, 55, 35, 782747, tzinfo=datetime.timezone.utc),
```

mimmus:
Thank you!

PR #5036, opened 2024-02-07
Serialize `pulp_last_updated` field

git-hyagi:
closes: #5033

### DIFF
--- a/CHANGES/5033.feature
+++ b/CHANGES/5033.feature
@@ -1,0 +1,1 @@
+Added a serializer to output `pulp_last_updated` field on pulp resources.
--- a/pulpcore/app/serializers/base.py
+++ b/pulpcore/app/serializers/base.py
@@ -418,9 +418,17 @@ class ModelSerializer(
     exclude_arg_name = "exclude_fields"
 
     class Meta:
-        fields = ("pulp_href", "pulp_created")
+        fields = ("pulp_href", "pulp_created", "pulp_last_updated")
 
     pulp_created = serializers.DateTimeField(help_text=_("Timestamp of creation."), read_only=True)
+    pulp_last_updated = serializers.DateTimeField(
+        help_text=_(
+            "Timestamp of the last time this resource was updated. Note: for immutable "
+            "resources - like content, repository versions, and publication - pulp_created and "
+            "pulp_last_updated dates will be the same."
+        ),
+        read_only=True,
+    )
 
     def _validate_relative_path(self, path):
         """


ipanova:
do we see any benefit to have this also on repo, publication and remote objects?

dkliban:
@ipanova I do see value in that. We should just add it to the ModelSerializer. @pulp/core what do you all think?

https://github.com/pulp/pulpcore/blob/main/pulpcore/app/serializers/base.py#L423

ipanova:
Not ModelSerializer because it is being used by Content/Artifacts and our claim that it is immutable.

ggainey:
> @ipanova I do see value in that. We should just add it to the ModelSerializer. @pulp/core what do you all think?
> 
> https://github.com/pulp/pulpcore/blob/main/pulpcore/app/serializers/base.py#L423

it's in the BaseModel, I don't see any reason to not-serialize, let's just make it accessible for everything

ipanova:
..so even if it will serialized on content then the updated field will always match the created field. I would refrain to expose it on content because it will create confusion/assumptions about content immutability.

ipanova:
..but I will not cast a -1 if we'd want to expose it for everything @ggainey

ggainey:
> ..but I will not cast a -1 if we'd want to expose it for everything @ggainey

My thinking is, the attribute *exists* at the BaseModel level. Hiding it doesn't really accomplish very much. From an end-user standpoint, it does make sense to maybe have a doc-note along the lines of "for immutable entities like Content, _created and _updated will always be the same".

gerrod3:
+1 to adding to ModelSerializer, we can also just exclude the field on the NoArtifactContentSerializer and ArtifactSerializer with the `exclude` attribute under the `Meta` class.

ipanova:
> +1 to adding to ModelSerializer, we can also just exclude the field on the NoArtifactContentSerializer and ArtifactSerializer with the `exclude` attribute under the `Meta` class.

Let's do that +1

pulpcore/app/serializers/content.py
```diff
@@ -183,7 +185,7 @@ class ContentChecksumSerializer(serializers.Serializer):
 
     class Meta:
         model = models.Content
-        fields = base.ModelSerializer.Meta.fields + (
+        fields = tuple(i for i in base.ModelSerializer.Meta.fields if i != "pulp_last_updated") + (
```
gerrod3, line 188:
```suggestion
        exclude = ("pulp_last_updated",)
        fields = base.ModelSerializer.Meta.fields + (
```
No need for fancy tuple conditions.

git-hyagi: I tried with the exclude (not sure if I did it wrong) but got the following exception:
```
...
  File "/usr/local/lib/python3.8/site-packages/rest_framework/serializers.py", line 1117, in get_field_names
    assert not (fields and exclude), (
AssertionError: Cannot set both 'fields' and 'exclude' options on serializer ArtifactSerializer.
```

lubosmj: We can exploit the set's functionality for this, right?

https://github.com/pulp/pulpcore/blob/a33612b89f48985448f5315dc5c67aa54f94d196/pulpcore/app/serializers/publication.py#L324

ipanova: in container plugin we have even a richer example
https://github.com/pulp/pulp_container/blob/main/pulp_container/app/serializers.py#L214
https://github.com/pulp/pulp_container/blob/main/pulp_container/app/serializers.py#L378

CHANGES/5033.feature
```diff
@@ -0,0 +1 @@
+Added a serializer to output `pulp_last_updated` field.
```
ipanova, line 1: ..on generic pulp resources like Distribution, Repo, Remote.

ipanova (dismissed):

mdellweg:
Repository versions are supposed to be immutable too. And Publications.

ipanova:
What are you saying is that we need to evaluate each instance that inherits from ModelSerializer
```
$ git grep '(ModelSerializer)'
pulpcore/app/serializers/access_policy.py:class AccessPolicySerializer(ModelSerializer):
pulpcore/app/serializers/acs.py:class AlternateContentSourceSerializer(ModelSerializer):
pulpcore/app/serializers/acs.py:class AlternateContentSourcePathSerializer(ModelSerializer):
pulpcore/app/serializers/domain.py:class DomainSerializer(ModelSerializer):
pulpcore/app/serializers/exporter.py:class ExporterSerializer(ModelSerializer):
pulpcore/app/serializers/exporter.py:class ExportSerializer(ModelSerializer):
pulpcore/app/serializers/importer.py:class ImporterSerializer(ModelSerializer):
pulpcore/app/serializers/importer.py:class ImportSerializer(ModelSerializer):
pulpcore/app/serializers/importer.py:class PulpImportSerializer(ModelSerializer):
pulpcore/app/serializers/progress.py:class ProgressReportSerializer(ModelSerializer):
pulpcore/app/serializers/progress.py:class GroupProgressReportSerializer(ModelSerializer):
pulpcore/app/serializers/publication.py:class PublicationSerializer(ModelSerializer):
pulpcore/app/serializers/publication.py:class ContentGuardSerializer(ModelSerializer):
pulpcore/app/serializers/publication.py:class DistributionSerializer(ModelSerializer):
pulpcore/app/serializers/repository.py:class RepositorySerializer(ModelSerializer):
pulpcore/app/serializers/task.py:class TaskSerializer(ModelSerializer):
pulpcore/app/serializers/task.py:class TaskGroupSerializer(ModelSerializer):
pulpcore/app/serializers/task.py:class TaskCancelSerializer(ModelSerializer):
pulpcore/app/serializers/task.py:class ApiAppStatusSerializer(ModelSerializer):
pulpcore/app/serializers/task.py:class ContentAppStatusSerializer(ModelSerializer):
pulpcore/app/serializers/task.py:class WorkerSerializer(ModelSerializer):
pulpcore/app/serializers/task.py:class TaskScheduleSerializer(ModelSerializer):
pulpcore/app/serializers/user.py:class RoleSerializer(ModelSerializer):
```

mdellweg:
Now that you put it like that. I'm not so sure what i really expect here.
Does the Field make sense for a internally managed model like ProgressReport?

ipanova:
We have discussed this at our pulpcore meeting. Let's not make it unnecessary complex and expose this field everywhere. Just document it so it is visible through docs api, add in the help text that for immutable resources create and updated date will be the same.

pulpcore/app/serializers/base.py
```diff
@@ -418,9 +418,17 @@ class ModelSerializer(
...
 
     class Meta:
-        fields = ("pulp_href", "pulp_created")
+        fields = ("pulp_href", "pulp_created", "pulp_last_updated")
 
     pulp_created = serializers.DateTimeField(help_text=_("Timestamp of creation."), read_only=True)
+    pulp_last_updated = serializers.DateTimeField(
+        help_text=_(
+            "Timestamp of the last time this resource was updated. Note: for immutable "
+            "resources - like distribution, repository, and remote - pulp_created and "
```
ipanova, line 427: errr, this is exactly the opposite, these are all mutable and can be updated. Content, for example, is immutable, so willingly you can mention it.

ipanova (approved):

ggainey (approved):

PR #5053, opened 2024-02-14
Make recieved_otel_span more resilient to slow updates

gerrod3:
[noissue]

### DIFF
--- a/pulpcore/tests/functional/__init__.py
+++ b/pulpcore/tests/functional/__init__.py
@@ -545,7 +545,7 @@ def received_otel_span():
     so, please, run the server as follows: python3 pulpcore/tests/functional/assets/otel_server.py
     """
 
-    def _received_otel_span(data):
+    def _received_otel_span(data, retries=3):
         if os.environ.get("PULP_OTEL_ENABLED") != "true":
             # pretend everything is working as expected if tests are run from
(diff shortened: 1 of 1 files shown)


pulpcore/tests/functional/__init__.py
```diff
@@ -557,8 +557,13 @@ async def _send_request():
                 async with session.post(f"{otel_server_url}/test", json=data) as response:
                     return response.status
 
-        status = asyncio.run(_send_request())
-        return True if status == 200 else False
+        while retries:
+            status = asyncio.run(_send_request())
```
dralley, line 561: If this was synchronous code it would be one thing, but given that this is async, why would this be sensitive to slow updates?  Is there a timeout?  And if so, could we just adjust the timeout?

gerrod3: The otel machinery is a separate server that receives updates from Pulp. This fixture is requesting that info from the otel server, so if Pulp hasn't sent the update yet then this fixture will receive a 404 from the otel server.

lubosmj (approved):
This workaround looks fine to me. We may want to implement a pytest plugin to run this server fixture.

decko (approved):
LGTM 🚀

PR #5058, opened 2024-02-16
Add a new authentication method to check a specific header and its content.

decko:


pulpcore/app/authentication.py
```diff
@@ -1,12 +1,82 @@
...
+
+                username = json_path.input_value(header_value).first()
+
+            except json.JSONDecodeError:
+                _logger.debug("Access not allowed - Invalid JSON or Path not found.")
+                raise AuthenticationFailed(_("Access denied. Invalid JSON or Path not found."))
+        else:
+            return None
+
+        user = User.objects.get(username=username)
```
mdellweg, line 71: Take a good look at:

https://github.com/encode/django-rest-framework/blob/master/rest_framework/authentication.py#L230

template_config.yml
```diff
@@ -48,7 +48,16 @@ plugins:
 post_job_template: null
 pre_job_template: null
 publish_docs_to_pulpprojectdotorg: true
-pulp_env: {}
+pulp_env:
```
mdellweg, line 51: Let's not include these on all scenarios.
We should try hard to have at least one of the tests run a verbatim pulp.

You should be able to access the settings to conditionally skip the tests.

pulpcore/app/authentication.py
```diff
@@ -1,12 +1,85 @@
...
+        if not user:
+            return None
+
+        return (user, None)
+
+
+class RHConsoleDotAuthentication(JQFilterHeaderRemoteUserAuthentication):
+
+    header = "HTTP_X_RH_IDENTITY"
+    jq_filter = ".identity.user.username"
```
mdellweg, line 85: Instead of creating a new subclass, let's make the one above configurable (like the PulpRemoteUserBackend).

pulpcore/app/authentication.py
```diff
@@ -1,12 +1,85 @@
...
+            return None
+
+        try:
+            header_content = request.META.get(self.header)
+            header_decoded_content = b64decode(header_content)
+        except Base64DecodeError:
+            _logger.debug("Access not allowed - Header content is not Base64 encoded.")
+            raise AuthenticationFailed("Access denied.")
+
+        if self.jq_filter:
```
mdellweg, line 57: This safe guard should stop being necessary with the other comment.

pulpcore/app/authentication.py
```diff
@@ -1,12 +1,85 @@
...
+                return None
+        else:
+            return None
+
+        user = authenticate(request, remote_user=remote_user)
+
+        if not user:
+            return None
+
+        return (user, None)
```
mdellweg, line 79: Why don't we pass the jq object as "auth" here?

pulpcore/app/authentication.py
decko: Cool. Gonna work on that.

pulpcore/app/authentication.py
```diff
@@ -1,12 +1,77 @@
...
+
+class RHConsoleDotAuthentication(BaseAuthentication):
+    """
+    Authenticate users based on a jq filter applied the value of a specific header.
+
+    For users logging in first time it creates User record.
+    """
+
+    header = settings.RH_CRC_HEADER_AUTH
+    jq_filter = settings.RH_CRC_HEADER_JQ_FILTER
```
mdellweg, line 35: About the naming. This class is supposed to be able to be configure for RH_CRC. It should sound like being exclusive for it.

template_config.yml
dkliban: @mdellweg  so move them to pulp_env_azure?

template_config.yml
```diff
@@ -56,7 +56,9 @@ pulp_env_azure:
   otel_exporter_otlp_protocol: http/protobuf
   pulp_otel_enabled: 'true'
 pulp_env_gcp: {}
-pulp_env_s3: {}
+pulp_env_s3:
```
mdellweg, line 59: Actually this should be in pulp_settings_s3 i think.

dkliban: The way we render pulp_settings_s3 does not allow us to specify settings that are a dictionary. So we need to pass it in as env vars.

mdellweg: Tell me more. Don't we use the "repr" filter to write the settings to the conffile? That should handle it.

template_config.yml
decko: Done. Using `pulp_s3_settings`.

decko: Thanks for the suggestion @mdellweg

pulpcore/app/authentication.py
decko: you say the name of the variables? `header` and `jq_filter`?

mdellweg: Yes, and the class name. Everything with RH CRC in it. I think the former name you had for the base class was not bad (describing what the thing is, not whom we have dedicated it to.).

dkliban: How about the following: 
```
JSONHeaderRemoteAuthentication 
settings.AUTHENTICATION_JSON_HEADER
settings.AUTHENTICATION_JSON_HEADER_JQ_FILTER
```

decko: @dkliban basically it how was named before. We were subclassing a JqHeaderAuthorization class.

pulpcore/app/authentication.py
decko: I'm not seeing a use for it here. Some Authentication classes return a `Token` instance, most of them return `None`.

pulpcore/app/authentication.py
decko: Sorry, which one?

pulpcore/app/authentication.py
mdellweg: Yes, and the name of the super class was just fine.

pulpcore/app/settings.py
```diff
@@ -254,6 +254,11 @@
 
 REMOTE_USER_ENVIRON_NAME = "REMOTE_USER"
 
+AUTHENTICATION_JSON_HEADER = ""
+AUTHENTICATION_JSON_HEADER_JQ_FILTER = ""
+RH_CRC_HEADER_AUTH = "HTTP_X_RH_IDENTITY"
+RH_CRC_HEADER_JQ_FILTER = ".identity.user.username"
+
```
mdellweg, line 261: Remove these now. They should be site specific settings.

mdellweg (changes requested):

pulpcore/app/settings.py
```diff
@@ -256,8 +256,6 @@
 
 AUTHENTICATION_JSON_HEADER = ""
 AUTHENTICATION_JSON_HEADER_JQ_FILTER = ""
```
mdellweg, line 258: Obviously these "default" values are not useful. So please remove them from here, and add a conditional validator at the end of the file.

pulpcore/app/settings.py
```diff
@@ -254,6 +254,9 @@
 
 REMOTE_USER_ENVIRON_NAME = "REMOTE_USER"
 
+AUTHENTICATION_JSON_HEADER = ""
+AUTHENTICATION_JSON_HEADER_JQ_FILTER = ""
+
```
mdellweg, line 259:
```suggestion
```

pulpcore/app/authentication.py
mdellweg: I think it is there for the user backend to answer more questions about permissions. Are there any potentially interesting informations burried in that header?

pulpcore/app/settings.py
```diff
@@ -379,6 +382,33 @@
...
+    startswith="HTTP_",
+    must_exist=True,
+    when=remote_user_backend_auth_validator & json_header_auth_class_restframework_validator,
+)
+
+authentication_json_header_jq_filter_validator = Validator(
+    "AUTHENTICATION_JSON_HEADER_JQ_FILTER",
+    must_exist=True,
+    when=remote_user_backend_auth_validator & json_header_auth_class_restframework_validator,
+)
```
mdellweg, line 407: We cannot make it dependent on the one RemoteUserBackend. And we shouldn't have to. The rule reads like if you use the auth backend, it must be configured.

```suggestion

json_header_auth_class_restframework_validator = Validator(
    "REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES",
    cont="pulpcore.app.authentication.JSONHeaderRemoteAuthentication",
    when=remote_user_backend_auth_validator,
)

authentication_json_header_validator = Validator(
    "AUTHENTICATION_JSON_HEADER",
    startswith="HTTP_",
    must_exist=True,
    when=json_header_auth_class_restframework_validator,
)

authentication_json_header_jq_filter_validator = Validator(
    "AUTHENTICATION_JSON_HEADER_JQ_FILTER",
    must_exist=True,
    when=json_header_auth_class_restframework_validator,
)
```

pulpcore/app/authentication.py
decko: This is an example of what we receive:
```
{
  "identity": {
    "account_number": "0369233",
    "type": "User",
    "user": {
      "username": "jdoe",
      "email": "jdoe@acme.com",
      "first_name": "John",
      "last_name": "Doe",
      "is_active": true,
      "is_org_admin": false,
      "locale": "en_US"
    },
    "internal": {
      "org_id": "3340851",
      "auth_type": "basic-auth",
      "auth_time": 6300
    }
  }
}
```
taken from https://redhatinsights.github.io/insights-results-aggregator/authentication.html

mdellweg: Nevermind.

pulpcore/app/authentication.py
mdellweg: jq filter will always be set at this point.

pulpcore/app/authentication.py
```diff
@@ -1,12 +1,78 @@
...
+                _logger.debug("Access not allowed - Invalid JSON.")
+                raise AuthenticationFailed(_("Access denied. Invalid JSON."))
+
+            if not remote_user:
+                _logger.debug("Path or value not found.")
+                return None
+        else:
+            return None
+
+        user = authenticate(request, remote_user=remote_user)
```
bmbouter, line 72: Does this auto-create the user if it doesn't exit?

decko: Yep.

bmbouter: good deal, thanks! +1 to this PR

pulpcore/app/authentication.py
```diff
@@ -1,12 +1,78 @@
...
     header = settings.REMOTE_USER_ENVIRON_NAME
 
 
 class PulpNoCreateRemoteUserBackend(RemoteUserBackend):
     create_unknown_user = False  # Configure RemoteUserBackend to not create users
+
+
+class JSONHeaderRemoteAuthentication(BaseAuthentication):
+    """
+    Authenticate users based on a jq filter applied the value of a specific header.
```
ipanova, line 29: Add some docs please

mdellweg (commented):
Looks good to go. Please squash the commits so we can merge.

mdellweg (approved):

docs/authentication/json_header.rst
```diff
@@ -0,0 +1,42 @@
...
+Also, you need to add the ``pulpcore.app.authentication.JSONHeaderRemoteAuthentication`` to 
+``REST_FRAMEWORK['DEFAULT_AUTHENTICATION_CLASSES']`` setting.
+
+Following that, change the ``AUTHORIZATION_JSON_HEADER`` to your value of choice. Remembering that it
+must start with `HTTP_`, so, if your header is ``x-authentication-service``, you need to set it to 
+``HTTP_X_AUTHENTICATION_SERVICE``.
+Also, set a JQ filter on ``AUTHORIZATION_JSON_HEADER_JQ_FILTER``. You can find the JQ query syntax and reference on its
+official site `here <https://jqlang.github.io/jq/>`_.
+
+Remember that the content of the header must be Base64 encoded.
```
pedro-psb, line 42: On this `Enabling JSONHeaderRemoteAuthentication` section it would be more idiomatic of a "howto" style to explicit list or enumerate the steps (instead of concatenate them in the paragraph). Like:

```
To enable the ``JSONHeaderRemoteAuthentication``:

1. Add the ``django.contrib.auth.backends.RemoteUserBackend`` to
``AUTHENTICATION_BACKENDS``, or some authentication backend that subclasses it.
1. Add the ``pulpcore.app.authentication.JSONHeaderRemoteAuthentication`` to 
``REST_FRAMEWORK['DEFAULT_AUTHENTICATION_CLASSES']`` setting.
(...)
```

Because the steps are small, it fits with an enumeration. If the steps required more details, they could be defined into subheaders.

ps: In markdown:
```bash
1. text
1. text # becomes 2
1. text # becomes 3
```

staging_docs/admin/guides/authentication/05-json_header.md
pedro-psb, line 1: :100:

docs/authentication/json_header.rst
```diff
@@ -0,0 +1,46 @@
...
+
+JSON Header Authentication
+--------------------------
+
+In a situation where it is not possible to use ``Basic Authentication`` Pulp can rely on a third-party
+service to authenticate a user.
+Using ``JSONHeaderRemoteAuthentication`` it's possible to receive a payload and even use ``JQ`` to filter
+it and obtain a user. The user is created in the database if one is not found.
+
+You can set ``AUTHORIZATION_JSON_HEADER`` and ``AUTHORIZATION_JSON_HEADER_JQ_FILTER`` to obtain a user
```
ipanova, line 11: not AUTHORIZATION_ but AUTHENTICATION_

pulpcore/app/settings.py
```diff
@@ -254,6 +254,9 @@
 
 REMOTE_USER_ENVIRON_NAME = "REMOTE_USER"
 
+AUTHENTICATION_JSON_HEADER = ""
```
ipanova, line 257: please document them in settings.md or do we need to have them in default settings?

docs/authentication/json_header.rst
decko: 🤯

pulpcore/app/settings.py
decko: We need to have them in the default settings, and they also have a validation policy that triggers an exception if a certain condition is not met.

ipanova (approved):

mdellweg (approved):

mdellweg:
@decko 
Where did the Changelog go for this one?
Now that it's released, please add an entry to the CHAGNLOG.rst directly.
Protip: Do not use [noissue] unless your final PR needs it.

PR #5062, opened 2024-02-18
Update CI files for branch 3.18

pulpbot:


mdellweg (approved):

ggainey:
Summary of docs-failure:
- We build docs in the github-action context
- We need core-requirements.txt in that context for the mkdocs to work correctly
- The github-action context is running pyton-3.11 with this ci-update
- py/3.11 doesn't need aioredis anymore
- core/3.18 requires aioredis because it runs on py/3.9
- aioredis and py/3.11 are in conflict
- pain results

mdellweg:
We can probably stop building the docs on this branch altogether.
With the new docs site it will be way less important anyway.

ggainey:
> We can probably stop building the docs on this branch altogether. With the new docs site it will be way less important anyway.

This sounds reasonable to me - the only docs that we care about changing on release-branches, is the CHANGELOG, which (as @mdellweg reminded me) maintained on the main branch itself). Let's turn off the docs-job on the release-branches where it's broken.

dkliban (approved):

PR #5068, opened 2024-02-19
 Adds an example of domain creation with S3 

dkliban:


ipanova (approved):
nice!

staging_docs/user/guides/create-domains.md
```diff
@@ -48,20 +48,31 @@ the content app behavior for your domain through the fields `redirect_to_object_
 `hide_guarded_distributions`. See `settings` for more details on these settings.
 
 ```bash
-pulp domains create \
+pulp domain create \
```
mdellweg, line 51: +1

mdellweg (approved):
Not an expert on S3 domains, but generally looking good.

ggainey (approved):

PR #5072, opened 2024-02-21
Serialize task dispatching accross all components

### LINKED ISSUE #5071: Possible race on task dispatch
We identified that there may be a possible race when dispatching new tasks.
With two processes adding a new task they become visible to the workers after finishing their transaction, but the created_at fields may actually report a different order. This can lead to  two workers picking up conflicting tasks concurrently.
This may be even worse on distributed systems where you can have a clockscrew on top.

We believe the window to already be very small, so any attempt to fix is should be like a vacuum seal.

Current working idea: Wrap adding a task to the table in a transaction-based-exclusive-wait-for advisory-lock. Check that the current latest task is actually from the past and add the new task afterwards.

mdellweg:
This will acquire a transaction based exclusive advisory lock before creating a new task object in the database and check the actual validity of the new tasks "pulp_created" field.

fixes #5071


pulpcore/tasking/tasks.py
```diff
@@ -148,6 +154,13 @@ def dispatch(
     notify_workers = False
     with contextlib.ExitStack() as stack:
         with transaction.atomic():
+            # Task creation need to be serialized so that pulp_created will provide a stable order
+            # at every time. We specifically need to ensure that each task, when commited to the
+            # task table will be the newest with respect to `pulp_created`.
+            with connection.cursor() as cursor:
+                # Wait for exclusive access and release automatically after transaction.
+                cursor.execute("SELECT pg_advisory_xact_lock(%s, %s)", [0, TASK_DISPATCH_LOCK])
+            newest_created = Task.objects.aggregate(Max("pulp_created"))["pulp_created__max"]
```
decko, line 163: Would make sense to have an index for `pulp_created`?

mdellweg: I guess so.
But this should be backportable.

mdellweg: Actually, we do have that index.
https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/task.py#L273

pulpcore/tasking/tasks.py
```diff
@@ -159,6 +172,15 @@ def dispatch(
...
                 versions=versions,
             )
+            if newest_created and task.pulp_created <= newest_created:
+                # Let this workaround not row forever into the future.
+                if newest_created - task.pulp_created > timedelta(seconds=1):
+                    # Do not commit the transaction if this condition is not met.
+                    # If we ever hit this, think about delegating the timestamping to PostgresQL.
+                    raise RuntimeError("Clockscrew detected. Task dispatching would be dangerous.")
+                # Try to work around the smaller glitch
+                task.pulp_created = newest_created + timedelta(milliseconds=1)
```
gerrod3, line 182: What would happen if two tasks are dispatched at the same time and both run into this bug, would they both get the same `pulp_created`, would this cause a problem?

mdellweg: This piece of code cannot run on multiple machines at the same time. See the locking part.

decko (approved):
LTTM 🚀 

but maybe some folks like @ggainey would like to take a look on it too.

dkliban (approved):

PR #5083, opened 2024-02-27
Allow selecting enabled plugins

### LINKED ISSUE #5235: As a user I want to select the active pulp plugins
Currently not all Plugins in the official Container images allow working with Domains.
This means you need to roll your own to use that feature.
We could start building images for different combinations of plugins. But where does it start and end? There are too many combinations.
This single addition would make the single set of shipped images $2^N$ times more versatile.

It is a necessity to be able to test domain features properly with pulp-cli.

mdellweg:
fixes #5235

### DIFF
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -120,14 +120,6 @@ INSTALLED_APPS = [
     "pulpcore.app",
 ]
 
-# Enumerate the installed Pulp plugins during the loading process for use in the status API
-INSTALLED_PULP_PLUGINS = []
-
-for entry_point in entry_points(group="pulpcore.plugin"):
-    plugin_app_config = entry_point.load()
-    INSTALLED_PULP_PLUGINS.append(entry_point.name)
-    INSTALLED_APPS.append(plugin_app_config)
-
 # Optional apps that help with development, or augment Pulp in some non-critical way
 OPTIONAL_APPS = [
     "crispy_forms",
@@ -390,9 +382,17 @@ OTEL_ENABLED = False
 
 # HERE STARTS DYNACONF EXTENSION LOAD (Keep at the very bottom of settings.py)
 # Read more at https://www.dynaconf.com/django/
-from dynaconf import DjangoDynaconf, Validator  # noqa
+from dynaconf import DjangoDynaconf, Dynaconf, Validator  # noqa
 
 # Validators
+
+enabled_plugins_validator = Validator(
+    "ENABLED_PLUGINS",
+    is_type_of=list,
+    len_min=1,
+    when=Validator("ENABLED_PLUGINS", must_exist=True),
+)
+
 storage_keys = ("STORAGES.default.BACKEND", "DEFAULT_FILE_STORAGE")
 storage_validator = (
     Validator("REDIRECT_TO_OBJECT_STORAGE", eq=False)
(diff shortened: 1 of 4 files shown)


pulpcore/app/settings.py
```diff
@@ -380,14 +373,38 @@
...
+
 settings = DjangoDynaconf(
     __name__,
     ENVVAR_PREFIX_FOR_DYNACONF="PULP",
     ENV_SWITCHER_FOR_DYNACONF="PULP_ENV",
-    PRELOAD_FOR_DYNACONF=[
-        "{}.app.settings".format(plugin_name) for plugin_name in INSTALLED_PULP_PLUGINS
-    ],
     ENVVAR_FOR_DYNACONF="PULP_SETTINGS",
+    post_hooks=(load_plugin_config_hook,),
```
rochacbruno, line 407: Where does ENABLED_PLUGINS comes from? an environment variable?

I see 2 simpler ways for doing it without requiring a hook

1. read the env directly
```py
PRELOAD_FOR_DYNACONF=[
    "{}.app.settings".format(plugin_name) for plugin_name in INSTALLED_PULP_PLUGINS if plugin_name in os.getenv("PULP_ENABLED_PLUGINS", [])
],
```

The downside of above strategy is that variable will need to come from envvar, not set on a settings.py file.

2. Post load plugin data without the hook

```py
settings = DjangoDynaconf(....)
# then imediatelly after
for plugin_name in INSTALLED_PULP_PLUGINS:
    if plugin_name not in settings.get("enabled_plugins", []):
        continue
    settings.load_file("{}.app.settings".format(plugin_name))
```

rochacbruno: this will also require a late validation, so replace

validators= [...] 

with

```py
settings = DjangoDynaconf(...)
for plugin_name in INSTALLED_PULP_PLUGINS:
    ....

# invoke validators
settings.validators.register(**list_of_validators)
settings.validators.validade()
```

pulpcore/app/settings.py
```diff
@@ -380,14 +373,36 @@
 )
 
 
+def load_plugin_config_hook(settings):
+    # Enumerate the installed Pulp plugins during the loading process for use in the status API
+    ENABLED_PLUGINS = getattr(settings, "ENABLED_PLUGINS", None)
```
ipanova, line 378: does it make sense to add validatiion to the setting so that at least one plugin needs to be enabled?

pulpcore/app/settings.py
rochacbruno: You are right, my suggestion would put plugin loading at the end of the process and then override local settings.

That is why in the current implementation we are using PRELOAD to ensure plugins settings loads before other settings sources.

Looks like we needed to have support for `pre` hooks, but right now we only support `post` loading hooks.

One idea is to write a custom loader to bring plugin_settings and then change the order of loaders putting the plugin_loader before everything else.

ideas @pedro-psb ?

pedro-psb: If taking approach 2, I believe an explicit load of the settings files should ensure the external file precedence:

```python
settings = DjangoDynaconf(...)
# (plugins load)
settings.load_file(settings.get("PULP_SETTINGS"))
# (validators)
```

But this original hook solution seems to do the trick, is there a particular reason not to use it?

mdellweg: The thing is that we need to keep the order, because "everything else" is what may provide the ENABLED_PLUGINS.

mdellweg:
Something weird is going on here. `_bypass_evaluation` should not be in the final settings object...
```python
In [4]: settings.TEMPLATES.to_list()
Out[4]: 
[{'BACKEND': 'django.template.backends.django.DjangoTemplates',
  'DIRS': [PosixPath('/src/pulpcore/pulpcore/app/templates')],
  'APP_DIRS': True,
  'OPTIONS': {'context_processors': ['django.template.context_processors.debug',
    'django.template.context_processors.request',
    'django.contrib.auth.context_processors.auth',
    'django.contrib.messages.context_processors.messages']},
  '_bypass_evaluation': True}]
```

mdellweg:
> We added some fixes related to this in 3.2.2. If you are using `>=3.2.2` already, I can work on a fix (that looks easy) and do a release on friday. Wdyt @rochacbruno?

Using 3.2.4.

mdellweg:
Looks like there are still some issues around dynaconf deeper merge strategies.

e.g. pulp_ansible needs to inject reusable ap conditions. It does not work this way.

pedro-psb:
Does this satisfy the deeper merge strategies requirements? We'll merge it in `3.3.0` (probably next release). https://github.com/dynaconf/dynaconf/pull/1001

pedro-psb:
I've worked on this issue today and looks like I got it working.
Here is [the branch](https://github.com/pedro-psb/pulpcore/tree/select-plugins/pedro-branch). Can you checkout and do some tests?

I'm leaving in PTO today, feel free to pick the changes here if they are good!

pulpcore/app/settings.py
```diff
@@ -489,27 +489,59 @@ def otel_middleware_hook(settings):
     return data
 
 
+def load_plugin_config_hook(settings):
```
rochacbruno, line 492: Do you need this function? I dont see it being called  or registered anywhere

pulpcore/app/settings.py
```diff
@@ -489,27 +489,59 @@ def otel_middleware_hook(settings):
...
         cache_validator,
+        enabled_plugins_validator,
         sha256_validator,
         storage_validator,
         unknown_algs_validator,
         json_header_auth_validator,
         authentication_json_header_openapi_security_scheme_validator,
     ],
-    post_hooks=otel_middleware_hook,
+    post_hooks=(otel_middleware_hook),
```
rochacbruno, line 532:
```suggestion
    post_hooks=(otel_middleware_hook,),
```

Missing a comma? IIRC this must be a list or tuple

pulpcore/app/settings.py
mdellweg: No longer, thanks...

PR #5133, opened 2024-03-14
OpenPGP key support

### LINKED ISSUE #3024: Provide basic gpg public key management
Pulp needs to handle public keys in some places. They are used for verifying uploaded or synced artifacts, and they may be exposed as part of a publication.

* Public GPG keys can be added as a pulp content (either with an artifact, or with the ascii-armored data in a text field - up for discussion). In case data will be stored in a text field we'd need to write a separate `handler` to serve the pub keys.
* There can be a `GPG-keyring repository type` able to hold those keys.
* `Repositories` and `Remotes` that verify artifacts can add a foreign key to these GPG repositories and assume all the keys there are  trusted for verification.
* `Signing services` can relate directly to the key and should prevent `orphan cleanup` from deleting the corresponding key, i.e `orphan-clean up` logic should be adjusted  to look not only at repository_membership but also whether there are signing services that point to the key.
* It should be possible to `import/export` key repositories as well as they should be covered by `RBAC`.
* Keys should be identifiable by their fingerprint and probably stripped off their signatures (maybe store them as a separate content).
* A publication of the keyring repository should produce a keyring file (*.bkx) as a published artifact.

We discussed whether keys should be content or a standalone generic model. But the benefits from handling keys as content is overwhelming.

**Additional context**
This is _not_ about private keys. Pulp will never set out to handle anything as sensitive as a private key. For signing we introduced the signing service already to handle all cases including the ones where you never get hold of the key itself.

mdellweg:
This adds a repository type as a keyring and content types to handle keys, keyids and key signatures.

fixes #3024


pulpcore/app/openpgp.py
```diff
@@ -0,0 +1,514 @@
...
+}
+
+
+ENCRYPTION_MODES = {
+    1: "EAX",
+    2: "OCB",
+}
+
+
+def packet_iter(data):
```
dralley, line 149: So, this appears to be a de-novo implementation of PGP packet parsing.  Is that 100% necessary?  And if so, why?

mdellweg: I think i looked for such things and none really ticked all the boxes (It's been a while though).

About pysequoia (from their pypi representation):
`Note: This is a work in progress. The API is not stable!`
`Note that since pysequoia is implemented largely in Rust, a [Rust toolchain](https://rustup.rs/) is necessary for the installation to succeed.`

mdellweg: What i still don't see (from looking at their README) is a way to disassemble the bytestream into openpgp packets which we want to store individually. But i should take a closer look.

mdellweg: I cannot see how pysequoia is providing the raw parsing data we require here.
Having said that, I believe the rfc will not change much giving us not too bad of a maintenance burden. Also let me state that very clear here. We only act on openpgp keys as a content type here. We parse them to find the individual key/signature packets and extract some metadata. We do not perform anything cryptographic here.

pulpcore/app/models/openpgp.py
```diff
@@ -0,0 +1,234 @@
...
+        Content, related_name="openpgp_signatures", on_delete=models.PROTECT
+    )
+
+    @property
+    def expired(self):
+        return self.expiration_time and timezone.now() > self.created + self.expiration_time
+
+    @property
+    def key_expired(self):
+        # TODO handle more cases
```
dralley, line 145: Does this need fixing?

pulpcore/app/models/openpgp.py
```diff
@@ -0,0 +1,234 @@
...
+                pubkey.represent(repository_version=self.latest_version())
+                for pubkey in OpenPGPPublicKey.objects.filter(pk__in=self.latest_version().content)
+            ]
+        )
+        return gpg_verify(public_keys, signature, detached_data)
+
+    class Meta:
+        default_related_name = "%(app_label)s_%(model_name)s"
+        permissions = [
+            # ("sync_openpgpkeyring", "Can start a sync task"),
```
dralley, line 187: ^

pulpcore/app/openpgp.py
```diff
@@ -0,0 +1,514 @@
...
+    },
+    19: {
+        "name": "ECDSA",
+        "format": "om",
+    },
+    22: {
+        "name": "EdDSA",
+        "format": "om",
+    },
+}
```
dralley, line 103: All of these options are in the PGP standard, but that doesn't mean we want to accept all of them.  DSA is long-obsolete, as is MD5 below.  How do we want to handle that.

pulpcore/app/openpgp.py
```diff
@@ -0,0 +1,514 @@
...
+
+
+def analyze_signature(data, pubkey, signed_packet_type, signed_packet):
+    # Type 2
+    version = data[0]
+    if version == 3:
+        raise NotImplementedError("Version 3 signatures are not implemented.")
+    elif version in [4, 5]:
+        signature_type = data[1]
+        # pubkey_algorithm = PUBKEY_ALGORITHMS.get(data[2])
```
dralley, line 267: ^

pulpcore/app/openpgp.py
```diff
@@ -0,0 +1,514 @@
...
+        hash_algorithm = HASH_ALGORITHMS.get(data[3], "unknown")
+        hashed_size = (data[4] << 8) + data[5]
+        hashed_data = data[6 : 6 + hashed_size]
+        unhashed_size = (data[6 + hashed_size] << 8) + data[7 + hashed_size]
+        unhashed_data = data[8 + hashed_size : 8 + hashed_size + unhashed_size]
+        canary = data[8 + hashed_size + unhashed_size : 10 + hashed_size + unhashed_size]
+        # signature = data[10 + hashed_size + unhashed_size :]
```
dralley, line 274: ^

pulpcore/app/openpgp.py
```diff
@@ -0,0 +1,514 @@
...
+                raise ValueError("Out of band user ID or attribute signature.")
+        elif signature_type in [0x1F, 0x20, 0x30]:
+            # 0x1F Direct Key Signature
+            # 0x20 Key Revocation Signature
+            # 0x30 Certification Revocation Signature
+            if signed_packet_type != 6:
+                raise ValueError("Out of band key signature.")
+            hash_payload = b""
+        else:
+            # TODO:
```
dralley, line 304: ^

pulpcore/app/openpgp.py
```diff
@@ -0,0 +1,514 @@
...
+def analyze_pubkey(data):
+    # Type 5, 6, 7 or 14
+    # Secret bits are ignored
+    version = data[0]
+    created = timezone.datetime.fromtimestamp(int.from_bytes(data[1:5], "big")).astimezone()
+    if version == 3:
+        n = extract_mpi(data[8:])
+        e = extract_mpi(data[8 + len(n["raw"])])
+        fingerprint = hashlib.md5(n["body"] + e["body"]).hexdigest()
+        # expiration = int.from_bytes(data[5:7], "big")
```
dralley, line 364: ^

pulpcore/app/openpgp.py
```diff
@@ -0,0 +1,514 @@
...
+                    # Multi precision integer
+                    mpi = extract_mpi(key_data[pos:])
+                    pos += len(mpi["raw"])
+                elif item_type == "o":
+                    # OID
+                    oid = extract_oid(key_data[pos:])
+                    pos += len(oid["raw"])
+                elif item_type == "k":
+                    # KDF
+                    # TODO more analysis
```
dralley, line 390: ^

pulpcore/app/openpgp.py
dralley: Sequoia [has it](https://docs.rs/sequoia-openpgp/latest/sequoia_openpgp/parse/struct.PacketParser.html) AFAICT, but pysequoia doesn't have it exposed.

pulpcore/app/openpgp.py
mdellweg: Actually, I think we just take and organize whatever we are presented with.
This is just meant to be a keyring for public key data. Even for verifying anything, we call out to gpg. And the gpg installed should decline working with deprecated algorithms.
Even more so, we should be able to report on these formats, so a user can make an informed decision.

pulpcore/app/openpgp.py
mdellweg: This is mainly for documentation purpose.

mdellweg:
I removed the comments related to synching, as we don't support that.
Also improved the openpgp "library" module for readability.

pulpcore/app/openpgp.py
dralley: @mdellweg Could you enumerate anything I missed here?  https://github.com/wiktor-k/pysequoia/issues/32

mdellweg: At this point it's mainly splitting the key into subpackages, because that way we can collect signatures for keys from separate uploads. Also we need the fingerprints. The rest of the data is mainly conveniently there too. And I don't know what else we want to add later. It's really hard to say what we want/need from this new content type in the future.
Quite honestly i'm not ready add another dependency (let alone one that is not pure python) to "just" parse a simple file format that is probably not going to change. If this was about *doing actual cryptography* I would say the exact opposite.

pulpcore/app/models/openpgp.py
dralley: @mdellweg

mdellweg: I think we can wait until someone complains. Removing the comment.

dralley (dismissed):
To the extent that I can reasonably review this (not having much subject matter expertise on PGP), it looks fine.

One thing though, it kinda feels like the 3 migrations should be merged into one if possible.

dralley:
@mdellweg Is it possible to merge those migrations together?

mdellweg:
> Is it possible to merge those migrations together?

Should have done that earlier. It's so hard to constantly rebase them.

dralley (approved):

PR #5140, opened 2024-03-19
Fix replicate for a distribution with publication

### LINKED ISSUE #4637: Add ability to turn any Pulp instance into a replica
**Version**
Python PYPI installation
	PulpCore 3.32.0
	CertGuard 1.6.5
	file 1.14.4
	python 3.10.0
	rpm 3.22.3

**Describe the bug**
Replication task fails to maintain the publication attribute in a existing RPM distribution artifact. Instead it tries and fail to update the repository attribute in distribution artifact.

In a replication task group, I am getting a error for "general_update" task of a RPM distribution artifact
https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/base.py

ErrorDetail(string="Only one of the attributes 'repository' and 'publication' may used simultaneously.", code='invalid')

**To Reproduce**
In a upstream pulp server(Pulp1 server), create a repository with AutoPublish as false and it's distribution is based on Publications and repository field value is null.
In a downstream pulp server(Pulp2 server), do the same like create a repository with AutoPublish as false and it's distribution is based on Publications and repository field value is null.
Create a upstream object in Pulp2 server to replicate from upstream server (Pulp1 server)
Start the replication task in downstream pulp server(Pulp2 server) to replicate content from upstream server (Pulp1 server).
Replication task will fail to update the distribution artifact of the repository.

**Expected behavior**
Why Pulp replication doesn't replicate the publication from upstream pulp server when RPM repository distribution is based on publication.

mdellweg:
In case a distribution that is supposed to be subject to replication is referring to a publication (because it was created by someone else) we need to unset the publication field for the replicate task to succeed.

fixes #4637

### DIFF
--- a/pulpcore/app/replica.py
+++ b/pulpcore/app/replica.py
@@ -1,18 +1,36 @@
 from django.conf import settings
 from django.db.models import Model
+import logging
 
 from pulp_glue.common.context import PulpContext
 from pulpcore.tasking.tasks import dispatch
-from pulpcore.app.tasks.base import general_update, general_create, general_delete
+from pulpcore.app.tasks.base import (
+    general_update,
+    general_create,
+    general_multi_delete,
+)
 from pulpcore.plugin.util import get_url, get_domain
+
+_logger = logging.getLogger(__name__)
 
 
 class ReplicaContext(PulpContext):
-    def prompt(self, *args, **kwargs):
-        pass
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.out_buf = ""
+        self.err_buf = ""
 
-    def echo(self, *args, **kwargs):
-        pass
+    def echo(self, message: str, nl: bool = True, err: bool = False) -> None:
+        if err:
+            self.err_buf += message
+            if nl:
+                _logger.warn("{}", self.err_buf)
+                self.err_buf = ""
+        else:
+            self.out_buf += message
+            if nl:
+                _logger.info("{}", self.out_buf)
+                self.out_buf = ""
 
 
 class Replicator:
@@ -39,6 +57,7 @@ class Replicator:
         self.tls_settings = tls_settings
         self.domain = get_domain()
         uri = "/api/v3/distributions/"
+        # TODO check and compare this to distribution locking on the distribution viewset.
         if settings.DOMAIN_ENABLED:
             uri = f"/{self.domain.name}{uri}"
         self.distros_uri = uri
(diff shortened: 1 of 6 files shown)


pulpcore/app/replica.py
```diff
@@ -134,18 +134,17 @@ def create_or_update_repository(self, remote):
         return repository
 
     def create_or_update_distribution(self, repository, upstream_distribution):
+        distribution_data = {
+            "repository": get_url(repository),
+            "publication": None,
+            "base_path": upstream_distribution["base_path"],
+        }
```
mdellweg, line 141: Do we need to defer this to another method so it can easily be overwritten by a Replicator subclass?

Also why do distributions have names. They can perfectly be identified by their base_path, why do publications not have names? They are sooo hard to identify.

mdellweg:
I do not understand the test failure. It runs fine on my box. And i guess the green runs are the ones where domains are unavailable and the test is skipped.

pulpcore/app/replica.py
```diff
@@ -134,48 +153,40 @@ def create_or_update_repository(self, remote):
...
                         "partial": True,
                     },
                 )
         except self.distribution_model_cls.DoesNotExist:
             # Dispatch a task to create the distribution
+            distribution_data["name"] = upstream_distribution["name"]
             dispatch(
                 general_create,
                 task_group=self.task_group,
+                shared_resources=[repository],
```
mdellweg, line 186: I think this is what was missing. We need to wait for the sync to complete before we can assign the repository to the distribution.

But anyway, share-locking the repository should have been here all the way.

pulpcore/app/replica.py
```diff
@@ -193,35 +204,42 @@ def sync(self, repository, remote):
...
+        distribution_ids = [
+            (distribution.pk, self.app_label, self.distribution_serializer_name)
+            for distribution in self.distribution_model_cls.objects.filter(
+                pulp_domain=self.domain
+            ).exclude(name__in=names)
+        ]
+        if distribution_ids:
             dispatch(
-                general_delete,
+                general_multi_delete,
```
gerrod3, line 216: Nice use of multi-delete.

pulpcore/tests/functional/__init__.py
```diff
@@ -955,6 +955,44 @@ def random_artifact(random_artifact_factory):
     return random_artifact_factory()
 
 
+@pytest.fixture()
+def domain_factory(domains_api_client, pulp_settings, gen_object_with_cleanup):
```
gerrod3, line 959: Did you change anything by moving this fixture up the file? Also, do you want to change this to use `pulpcore_bindings`?

pulpcore/tests/functional/api/test_replication.py
```diff
@@ -44,6 +44,109 @@ def test_replication(
...
+        file_bindings.DistributionsFileApi,
+        file_bindings.RemotesFileApi,
+        file_bindings.RepositoriesFileApi,
+        file_bindings.ContentFilesApi,
+    ):
+        result = api_client.list(pulp_domain=replica_domain.name)
+        assert result.count == 1
+
+    # Now replicate backwards
+
```
gerrod3, line 124: Should there be any other actions here on the source domain or replica domain, like a new repository, new distribution, new content?

pulpcore/app/replica.py
```diff
@@ -39,6 +57,7 @@ def __init__(self, pulp_ctx, task_group, tls_settings):
         self.tls_settings = tls_settings
         self.domain = get_domain()
         uri = "/api/v3/distributions/"
+        # TODO check and compare this to distribution locking on the distribution viewset.
```
gerrod3, line 60: Distribution viewset lock should be updated to match this.

pulpcore/app/replica.py
gerrod3: > Do we need to defer this to another method so it can easily be overwritten by a Replicator subclass?

I would say so since plugins can have extra fields on their distribution models.

gerrod3 (changes requested):

pulpcore/tests/functional/__init__.py
mdellweg: No change just moved it to the factories section.
Some people seem to follow an always add to the end of the file policy.
But yes, good call there.

pulpcore/app/replica.py
mdellweg: This is another issue that may need more thinking to get it upgrade safe.
https://github.com/pulp/pulpcore/issues/5143

mdellweg:
@gerrod3 Does this address all your requests now?

gerrod3 (approved):

dkliban (approved):

PR #5169, opened 2024-03-25
Add counter reporting the size of served artifacts

### LINKED ISSUE #4602: Add additional metrics reporting size of artifacts served
## Need / Background

## Proposal

lubosmj:
closes #4602

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -19,6 +19,8 @@ from yarl import URL
 
 from asgiref.sync import sync_to_async
 
+from opentelemetry import metrics
+
 import django
 
 from pulpcore.constants import STORAGE_RESPONSE_MAP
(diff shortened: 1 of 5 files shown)


lubosmj:
I am currently deciding between using a synchronous and asynchronous counter. The current implementation does not work when parsing multiple requests in parallel. Stay tuned for the results!

pulpcore/content/handler.py
```diff
@@ -1116,4 +1141,10 @@ async def finalize():
 
         if response.status == 404:
             raise HTTPNotFound()
+
         return response
+
+    async def _report_served_artifact_size(self, size):
+        async with self.counter_lock:
+            # locking prevents multiple coroutines from accessing the same counter
```
decko, line 1149: 🤯

PR #5178, opened 2024-03-26
Fix import in wsgi preventing startup

gerrod3:
[noissue]


lubosmj (approved):
Ooopsie, wooopsie... Should we create an issue for this and backport it?

ggainey (approved):

Odilhao:
Just hit by this on 3.49.1 that we just packaged the rpm, can we get this backported?

PR #5192, opened 2024-03-27
Temporary fix for staging-docs build

pedro-psb:
TLDR: because [Griffe looks for the package in the relative path by default](https://github.com/mkdocstrings/griffe/blob/01da648b10aca57d316e74e4b0eef2ca31e92f32/src/griffe/loader.py#L182-L185) (and mkdocstring uses that unchanged), the wrong package is discovered instead the one I've defined programmatically if I'm in A Folder Called Pulpcore.

[noissue]


ggainey (approved):
Tested and works in my env - approved!

lubosmj (approved):

PR #5196, opened 2024-03-27
Fix bug where Last-Modified header was being updated on duplicate package uploads

### LINKED ISSUE #5149: Overwriting existing packages in backend storage can lead to caching issues
If an existing package is re-added to pulp, the default behavior will overwrite the existing file in backing storage. This is typically fine.
- If using Azure Blobstore, the timestamp of the blob is updated (Last-Modified time and ETag).
- Conversely, some CDN's (notably Azure Front Door) use Last-Modified Time as a signal that a file in origin has updated.
- This can lead to poor cache behavior, and in some cases, incomplete downloads as the CDN attempts to resolve disparate content.
- If we set `AZURE_OVERWRITE_FILES` to `false` this partially mitigates the issue (Last-Modified/ETag are unmodified). However, this results in duplicate copies written to storage (with a suffix to differentiate from the original).
- We should have an option that does "nothing" if the uploaded file already exists (don't overwrite, and don't write a new copy).

ritotn:
Fixes a bug where the Last-Modified header of a package stored in django-storages was being updated on duplicate uploads.

Closes #5149

### DIFF
--- a/CHANGES/plugin_api/5149.bugfix
+++ b/CHANGES/plugin_api/5149.bugfix
@@ -1,0 +1,1 @@
+Check for existing artifact before saving uploaded content as an artifact.
--- a/pulpcore/plugin/viewsets/content.py
+++ b/pulpcore/plugin/viewsets/content.py
@@ -1,7 +1,6 @@
 from drf_spectacular.utils import extend_schema
 
 from django.db import DatabaseError
-from django.db.utils import IntegrityError
 
 from pulpcore.app import tasks
 from pulpcore.plugin.serializers import (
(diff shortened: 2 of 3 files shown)


CHANGES/plugin_api/5149.bugfix
```diff
@@ -0,0 +1 @@
+Check for existing artifact before saving uploaded content as an artifact.
```
lubosmj, line 1: From the API's perspective, nothing has really changed. I would remove this entry completely.

ritotn: That makes sense, thanks for pointing that out!

dkliban (dismissed):

pulpcore/plugin/viewsets/content.py
```diff
@@ -133,18 +132,14 @@ def init_content_data(self, serializer, request):
...
             # parameter is replaced by 'artifact'
             artifact = Artifact.init_and_validate(task_payload.pop("file"))
+            # if artifact already exists, let's use it
             try:
+                artifact = Artifact.objects.get(
+                    sha256=artifact.sha256, pulp_domain=request.pulp_domain
+                )
+                artifact.touch()
+            except (Artifact.DoesNotExist, DatabaseError):
                 artifact.save()
```
mdellweg, line 142: Do we need to add another round of `get` in case another user uploaded a fresh instance of this artifact in the exact same moment?

mdellweg (changes requested):
To understand this change. Is `artifact.save()` accessing the storage backend before trying to save to the database?

Also, please follow https://docs.pulpproject.org/pulpcore/contributing/git.html#commit-message

daviddavis:
@mdellweg it sounds like that is what’s happening.

pulpcore/plugin/viewsets/content.py
daviddavis: There’s an incredibly small chance this save could raise an IntegrityError which probably should be handled.

ritotn: I've added a second check for this scenario!

PR #5204, opened 2024-03-29
[3.18] Revert to python-38

ggainey:
need to try and keep 3.18 publishable for the (old) systems that use it still.


dralley:
Feel free to resurrect this if you're still interested in it, otherwise I think let's just leave it until we actually have a need.

PR #5234, opened 2024-04-08
Add Worker metrics.

### LINKED ISSUE #3821: Adds open telemetry metrics for tasks

decko:
Close #3821


pulpcore/app/models/telemetry.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def __init__(self):
+        meter = get_meter(__name__)
+        self.unblocked_waiting_tasks_metter = meter.create_gauge(
+            name="tasks_unblocked_waiting_queue",
+            description="Number of unblocked tasks waiting in the queue.",
+            unit="tasks",
+        )
+
+    # @pg_lock
```
decko, line 19: WIP. Ideas on how to abstract this are welcome.

pulpcore/app/models/telemetry.py
```diff
@@ -0,0 +1,22 @@
...
+    def emmit_unblocked_waiting_tasks_metric(self, model):
```
decko, line 20: Receiving the model object to avoid circular imports

pulpcore/tasking/worker.py
```diff
@@ -125,6 +125,7 @@ def handle_worker_heartbeat(self):
 
     def shutdown(self):
         self.worker.delete()
+        # self.worker_counter.add(-1, attributes={"name": self.name})
```
decko, line 128: Remove in a future iteration

pulpcore/tests/functional/__init__.py
```diff
@@ -568,6 +568,37 @@ async def _send_request():
...
 
+@pytest.fixture(scope="session")
+def received_otel_metrics():
+    """A fixture for checking the presence of specific metrics on the otel collector server.
+
+    Ensure the collector server is up and running before executing tests with this fixture. To do
+    so, please, run the server as follows: python3 pulpcore/tests/functional/assets/otel_server.py
+    """
+
+    def _received_otel_metric(data, retries=3):
```
decko, line 579: Looking to find a way to avoid all this code duplication here

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -348,3 +353,25 @@ def test_task_version_prevent_pickup(dispatch_task, pulpcore_bindings):
...
+    )
+    task_href = dispatch_task(
+        "pulpcore.app.tasks.test.sleep", args=(0,), exclusive_resources=[resource]
+    )
+
+    task = pulpcore_bindings.TasksApi.read(task_href)
+    assert task.state == "waiting"
+    # sleep(2)
+    response = requests.get(f"{os.environ.get('OTEL_EXPORTER_OTLP_ENDPOINT')}/read")
+    metrics = json.loads(response.content)
```
decko, line 376: Hey @lubosmj, wdyt about removing the server side assertion we have today for a local one? Like, on test/client side?

pulpcore/tests/functional/assets/otel_server.py
```diff
@@ -70,6 +72,25 @@ async def _traces_handler(request):
...
+        metrics_data.ParseFromString(await request.read())
+        for resource_metric in metrics_data.resource_metrics:
+            for scope_metric in resource_metric.scope_metrics:
+                for metric in scope_metric.metrics:
+                    if metric.name in disabled_metrics:
+                        break
+                    translated_metric = {}
+                    translated_metric["name"] = metric.name
+                    translated_metric["description"] = metric.description
+                    translated_metric["unit"] = metric.unit
```
decko, line 90: We wanna test metric attributes? For sure.

pulpcore/app/models/telemetry.py
decko: Gonna use the PGAdvisoryLock context manager.

pulpcore/tests/functional/api/test_tasking.py
lubosmj: We can try working on this. Though, we would need to download all data from the server to perform the comparison on the client's side. Maybe, you are thinking about taking a different approach?

pulpcore/tasking/worker.py
```diff
@@ -391,6 +403,27 @@ def handle_available_tasks(self):
             for task in self.iter_tasks():
                 keep_looping = True
                 self.supervise_task(task)
+            self.record_unblocked_waiting_tasks_metric()
```
mdellweg, line 406: Please run this from the worker heartbeat routine only.

pulpcore/tests/functional/api/test_tasking.py
decko: Nope. I was thinking on download things and test directly on our side.

pulpcore/tasking/worker.py
decko: You say on `def beat` definition?

pulpcore/tasking/worker.py
```diff
@@ -244,6 +255,9 @@ def identify_unblocked_tasks(self):
             # Record the resources of the pending task
             taken_exclusive_resources.update(exclusive_resources)
             taken_shared_resources.update(shared_resources)
+            if changed:
+                time_trigger = str(timezone.now())
+                self.cursor.execute(f"NOTIFY pulp_worker_metrics_trigger, '{time_trigger}'")
```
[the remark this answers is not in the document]
decko: Trigger the record ;)

pulpcore/tasking/worker.py
```diff
@@ -206,7 +214,10 @@ def identify_unblocked_tasks(self):
         taken_exclusive_resources = set()
         taken_shared_resources = set()
         # When batching this query, be sure to use "pulp_created" as a cursor
-        for task in Task.objects.filter(state__in=TASK_INCOMPLETE_STATES).order_by("pulp_created"):
+        tasks_incomplete_states = Task.objects.filter(state__in=TASK_INCOMPLETE_STATES).order_by(
+            "pulp_created"
+        )
+        for task in tasks_incomplete_states:
```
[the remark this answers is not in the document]
decko: Sorry, gonna undo that. It was an idea to extract the first metric draft from this query itself.

pulpcore/tasking/worker.py
```diff
@@ -7,20 +7,24 @@
...
+from django.db.models import Case, Count, F, Max, Value, When
+from django.db.models.functions import Coalesce
 from django.utils import timezone
 
 from pulpcore.constants import (
     TASK_STATES,
     TASK_INCOMPLETE_STATES,
     TASK_SCHEDULING_LOCK,
     TASK_UNBLOCKING_LOCK,
+    TASK_METRIC_TRIGGER_LOCK,
```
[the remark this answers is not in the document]
decko: Done

pulpcore/tasking/worker.py
```diff
@@ -93,6 +98,8 @@ def _pg_notify_handler(self, notification):
         elif self.task and notification.channel == "pulp_worker_cancel":
             if notification.payload == str(self.task.pk):
                 self.cancel_task = True
+        elif notification.channel == "pulp_worker_metrics_trigger":
```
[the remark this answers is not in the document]
decko: Done

pulpcore/constants.py
```diff
@@ -12,6 +12,7 @@
 TASK_DISPATCH_LOCK = 21
 TASK_SCHEDULING_LOCK = 42
 TASK_UNBLOCKING_LOCK = 84
+TASK_METRICS_HEARTBEAT_LOCK = 74
```
mdellweg, line 15: That's your favorite number then?

pulpcore/tasking/worker.py
```diff
@@ -56,6 +60,7 @@ def __init__(self):
         self.task = None
         self.name = f"{os.getpid()}@{socket.getfqdn()}"
         self.heartbeat_period = settings.WORKER_TTL / 3
+        self.last_metric_heartbeat = timezone.now()
```
mdellweg, line 63: :+1: So this is never None.

pulpcore/tasking/worker.py
```diff
@@ -244,6 +252,9 @@ def identify_unblocked_tasks(self):
             # Record the resources of the pending task
             taken_exclusive_resources.update(exclusive_resources)
             taken_shared_resources.update(shared_resources)
+            if changed:
+                time_heartbeat = str(timezone.now())
+                self.cursor.execute(f"NOTIFY pulp_worker_metrics_heartbeat, '{time_heartbeat}'")
```
mdellweg, line 257: No triggering here.

pulpcore/tasking/worker.py
```diff
@@ -392,6 +403,55 @@ def handle_available_tasks(self):
                 keep_looping = True
                 self.supervise_task(task)
 
+    def record_unblocked_waiting_tasks_metric(self):
+        if self.last_metric_heartbeat:
```
mdellweg, line 407: unneeded if.

pulpcore/tasking/worker.py
```diff
@@ -392,6 +403,55 @@ def handle_available_tasks(self):
...
+            worker_ttl_delta = timedelta(seconds=(settings.WORKER_TTL / 3))
```
mdellweg, line 408: Don't recalculate stuff in the hot code paths.

pulpcore/tasking/worker.py
```diff
@@ -392,6 +403,55 @@ def handle_available_tasks(self):
...
+
+                        unblocked_waiting_tasks = Task.objects.filter(
+                            unblocked_at__isnull=False, started_at__isnull=True
+                        )
+                        unblocked_waiting_tasks_annotated = unblocked_waiting_tasks.annotate(
+                            unblocked_for=Value(timezone.now()) - F("unblocked_at")
+                        )
+                        unblocked_waiting_tasks_aggregated = (
+                            unblocked_waiting_tasks_annotated.aggregate(
+                                longest_age=Coalesce(Max("unblocked_for"), timezone.timedelta(0)),
```
lubosmj, line 433:
```suggestion
                                longest_age=Max("unblocked_for", default=timezone.timedelta(0),
```
Is this one possible?

pulpcore/tests/functional/assets/otel_server.py
```diff
@@ -70,6 +76,25 @@ async def _traces_handler(request):
...
+                for metric in scope_metric.metrics:
+                    if metric.name in disabled_metrics:
+                        _logger.info("Dropping {} metric".format(metric.name))
+                        break
+                    translated_metric = {}
+                    translated_metric["name"] = metric.name
+                    translated_metric["description"] = metric.description
+                    translated_metric["unit"] = metric.unit
+                    metrics.append(translated_metric)
+                    _logger.info("Received a {} metric meter".format(translated_metric["name"]))
```
lubosmj, line 95: We are logging metrics every time we receive them during the testing. This might clutter our logs a bit.

pulpcore/tests/functional/assets/otel_server.py
lubosmj: I would go even further and would try to test the value of the gauge if it is is possible with the current implementation. Do we test one of the following somewhere now?
```
1. Number of unblocked waiting tasks waiting more than 5 seconds
2. the longest waiting task time
```

pulpcore/tasking/worker.py
decko: I think it is. Gonna test it before committing your suggestion.

pulpcore/exceptions/__init__.py
```diff
@@ -1,5 +1,7 @@
 from .base import (
     AdvisoryLockError,
+    AdvisoryAcquireLockError,
```
ipanova, line 3: These do not seem to be used anywhere.

pulpcore/tests/functional/assets/otel_server.py
decko: Check on line 87. We drop the ones we don't need right now

pulpcore/tests/functional/assets/otel_server.py
decko: Still trying to understand how to convert it properly to a python dict.
Maybe it would be better if we just retrieve the protobuf data and test over it.

pulpcore/tasking/worker.py
```diff
@@ -392,6 +403,55 @@ def handle_available_tasks(self):
                 keep_looping = True
                 self.supervise_task(task)
 
+    def record_unblocked_waiting_tasks_metric(self):
+        if self.last_metric_heartbeat:
+            worker_ttl_delta = timedelta(seconds=(settings.WORKER_TTL / 3))
+            if timezone.now() > self.last_metric_heartbeat + worker_ttl_delta:
+                try:
+                    with PGAdvisoryLock(TASK_METRICS_HEARTBEAT_LOCK):
```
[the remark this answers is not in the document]
mdellweg: And perstering the database for it.

pulpcore/exceptions/__init__.py
decko: Thanks Ina! Forgot to remove them.

pulpcore/tasking/worker.py
```diff
@@ -452,6 +448,9 @@ def record_unblocked_waiting_tasks_metric(self):
                 except AdvisoryLockError:
```
mdellweg, line 448: I prefer suppress.

pulpcore/tasking/worker.py
```diff
@@ -452,6 +448,9 @@ def record_unblocked_waiting_tasks_metric(self):
...
                     pass
 
+        time_heartbeat = str(timezone.now())
+        self.cursor.execute(f"NOTIFY pulp_worker_metrics_heartbeat, '{time_heartbeat}'")
+
```
mdellweg, line 453: We should send this while still holding the lock.
Also we can reuse the time we we gathered at the beginning of the locked part.

pulpcore/tasking/worker.py
```diff
@@ -430,7 +426,7 @@ def record_unblocked_waiting_tasks_metric(self):
                         )
                         unblocked_waiting_tasks_aggregated = (
                             unblocked_waiting_tasks_annotated.aggregate(
-                                longest_age=Coalesce(Max("unblocked_for"), timezone.timedelta(0)),
+                                longest_age=Max("unblocked_for", default=timezone.timedelta(0)),
```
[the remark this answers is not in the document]
decko: Thanks @lubosmj

pulpcore/tasking/worker.py
decko: Done

pulpcore/tasking/worker.py
```diff
@@ -392,6 +400,52 @@ def handle_available_tasks(self):
...
+    def record_unblocked_waiting_tasks_metric(self):
+        worker_ttl_delta = timedelta(seconds=(self.heartbeat_period))
+        if now := timezone.now() > self.last_metric_heartbeat + worker_ttl_delta:
+            with contextlib.suppress(AdvisoryLockError), PGAdvisoryLock(TASK_METRICS_HEARTBEAT_LOCK):
+                meter = get_meter(__name__)
+                unblocked_waiting_tasks_meter = meter.create_gauge(
+                    name="tasks_unblocked_waiting_queue",
+                    description="Number of unblocked tasks waiting in the queue.",
+                    unit="tasks",
+                )
```
mdellweg, line 412: Can we have a more consistent naming schema? Like "{name}_meter" for the variable?

Also is it worth keeping these objects on the class instead of recreating them?

pulpcore/tasking/worker.py
```diff
@@ -392,6 +400,52 @@ def handle_available_tasks(self):
...
+
+                longest_waiting_tasks_meter = meter.create_gauge(
+                    name="tasks_longest_waiting_task",
```
mdellweg, line 415: Try to find a name that reflects the contents. This name suggests we report a task.

pulpcore/tasking/worker.py
```diff
@@ -392,6 +400,52 @@ def handle_available_tasks(self):
...
+                        unblocked_count_gte_threshold=Count(
+                            Case(
+                                When(
+                                    unblocked_for__gte=Value(timezone.timedelta(seconds=THRESHOLD_UNBLOCKED_WAITING_TIME)),
+                                    then=1,
+                                )
+                            )
+                        ),
+                    )
+                )
```
mdellweg, line 438: I am not convinced, that breaking up the query into multiple statements makes it any more readable.
This is about unblocked tasks. I got that after the second occurrence of unblocked.
Instead, make sure the annotation variables match the names of your meters.

pulpcore/tasking/worker.py
decko: > Also is it worth keeping these objects on the class instead of recreating them?

The way we're doing now it creates/recover just one instance each time we're about to send the metric. If I set this directly on the class we'll create the same number of instances but just one will send the metrics.

pulpcore/tasking/worker.py
decko: I think it makes easier for another person to quickly understand each step of the query. Since this is not increasing the cost of the query itself, I think it's  good to keep it as it is now.
Also, I changed the names. Thanks for the suggestion.

pulpcore/tasking/worker.py
```diff
@@ -7,20 +7,23 @@
 import signal
 import socket
 import contextlib
-from datetime import timedelta
+from datetime import datetime, timedelta
```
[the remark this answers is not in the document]
mdellweg: does it provede timedelta too?

pulpcore/tasking/worker.py
```diff
@@ -253,6 +254,8 @@ def identify_unblocked_tasks(self):
             # Record the resources of the pending task
             taken_exclusive_resources.update(exclusive_resources)
             taken_shared_resources.update(shared_resources)
+            if changed:
+                self.changes_on_queue = True
```
mdellweg, line 258: Move it up one frame.
Also a queue changes when you remove one element from the front too.

template_config.yml
```diff
@@ -54,6 +54,7 @@ pulp_env_azure:
   otel_bsp_max_queue_size: 1
   otel_exporter_otlp_endpoint: http://localhost:4318
   otel_exporter_otlp_protocol: http/protobuf
+  otel_metric_export_interval: 800
```
mdellweg, line 57: Is this used anywhere?

pulpcore/tasking/worker.py
```diff
@@ -293,7 +305,7 @@ def sleep(self):
         _logger.debug(_("Worker %s entering sleep state."), self.name)
         while not self.shutdown_requested and not self.wakeup:
             r, w, x = select.select(
-                [self.sentinel, connection.connection], [], [], self.heartbeat_period
+                [self.sentinel, connection.connection], [], [], self.heartbeat_period.seconds
```
mdellweg, line 308: Ahh, that explains why we didn't have it as timedelta before.

pulpcore/tasking/worker.py
mdellweg: It doubles the space taken up on my screen and also doubles the reading time. And a variable name like `some_query_annotated` does not convey any new information at all.
If you want to help people in the future to understand what's going on, then instead, make the code succinct and add a comment, about why the hell this is compiled into a single query.

template_config.yml
decko: Yes. we need it for the tests since the default timeout is 60 seconds.

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -370,6 +373,9 @@ def test_emmiting_unblocked_task_telemetry(dispatch_task, pulpcore_bindings, rec
     task = pulpcore_bindings.TasksApi.read(task_href)
     assert task.state == "waiting"
 
+    worker_heartbeat = (pulp_settings.WORKER_TTL / 3) + 1
+
+    sleep(worker_heartbeat)
```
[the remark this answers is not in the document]
decko: I know, but got no other solution than "wait for the metrics to be flushed"

pulpcore/tasking/worker.py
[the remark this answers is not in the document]
decko: Nope. And also the Django's timezone doesn't provide the `fromisostring` method. So I had to return the datetime module here.

pulpcore/tasking/worker.py
```diff
@@ -38,12 +41,16 @@
 _logger = logging.getLogger(__name__)
 random.seed()
 
+# All these four number are picked out of thing air.
+# Until we understand they aren't settings properly they gonna live as constants
```
ggainey, line 45: Unless/until we can provide reasonable ways to decide to change their values, the will live as constants instead of "proper" settings.

pulpcore/tasking/worker.py
```diff
@@ -38,12 +41,16 @@
...
 # Number of heartbeats for a task to finish on graceful worker shutdown (approx)
 TASK_GRACE_INTERVAL = 3
 # Number of heartbeats between attempts to kill the subprocess (approx)
 TASK_KILL_INTERVAL = 1
 # Number of heartbeats between cleaning up worker processes (approx)
 WORKER_CLEANUP_INTERVAL = 100
+# Threshold time in seconds of an unblocked task to consider a queue stalled
```
ggainey, line 52:
```suggestion
# Threshold time in seconds of an unblocked task before we consider a queue stalled
```

pulpcore/tasking/worker.py
```diff
@@ -223,7 +247,7 @@ def identify_unblocked_tasks(self):
                     _logger.debug("Marking canceling task %s unblocked.", task.pk)
                     task.unblock()
                     changed = True
-                # Don't consider this tasks reosurces as held.
+                # Don't consider this tasks resources as held.
```
ggainey, line 250:
```suggestion
                # Don't consider this task's resources as held.
```

ggainey (changes requested):
Wordsmithing pass w/ some suggestions.

decko:
> Wordsmithing pass w/ some suggestions.

Without works to thank you enough. 🙏🏽

ggainey (dismissed):

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -349,3 +351,41 @@ def test_task_version_prevent_pickup(dispatch_task, pulpcore_bindings):
...
+    resource = str(uuid4())
+
+    dispatch_task("pulpcore.app.tasks.test.sleep", args=(100,), exclusive_resources=[resource])
+
+    task_href = dispatch_task(
+        "pulpcore.app.tasks.test.sleep", args=(0,), exclusive_resources=[resource]
+    )
+
+    for dummy in range(10):
+        dispatch_task("pulpcore.app.tasks.test.sleep", args=(0,), exclusive_resources=[resource])
```
mdellweg, line 372: By making these tasks require the same resource we do not generate unblocked tasks.
Instead we should start one long running task per worker and then we can add further tasks (all without rousources probably) to have something to report on.

mdellweg: And don't forget to cancel the tasks after receiving the desired stats.

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -360,15 +363,20 @@ def test_emmiting_unblocked_task_telemetry(
...
-    dispatch_task("pulpcore.app.tasks.test.sleep", args=(100,), exclusive_resources=[resource])
+    for worker in range(workers_count):
+        dispatch_task("pulpcore.app.tasks.test.sleep", args=(100,), exclusive_resources=[resource])
 
     task_href = dispatch_task(
         "pulpcore.app.tasks.test.sleep", args=(0,), exclusive_resources=[resource]
     )
 
-    for dummy in range(10):
+    for dummy in range(workers_count * 5):
```
mdellweg, line 379: You are still just creating a bunch of blocked tasks.
Also why 5 extra tasks per worker. And where do we extract the number and age of blocked tasks now?
In it's current form, this test tells us very little about the feature it's supposed to test.

decko: You're right. I've fixed that in the latest changes.
But, some how, the unblocked tasks got executed before the long-running ones. 
https://github.com/pulp/pulpcore/actions/runs/8943880878/job/24569587920#step:14:956

Am I missing something here?

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -349,3 +354,44 @@ def test_task_version_prevent_pickup(dispatch_task, pulpcore_bindings):
...
+
+    # Checking online workers ready to get a task
+    age_threshold = timezone.now() - timedelta(seconds=pulp_settings.WORKER_TTL)
+    workers_online = pulpcore_bindings.WorkersApi.list(last_heartbeat__gte=age_threshold).count
+
+    # We need to generate long running tasks to block the workers from executing other tasks
+    for worker in range(workers_online + 1):
+        dispatch_task(
+            "pulpcore.app.tasks.test.sleep", args=(30,), exclusive_resources=[str(uuid4())]
+        )
```
mdellweg, line 374:
```suggestion
    resident_task_hrefs = [dispatch_task("pulpcore.app.tasks.test.sleep", args=(30,)) for worker in range(workers_online)]
```

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -349,3 +354,44 @@ def test_task_version_prevent_pickup(dispatch_task, pulpcore_bindings):
...
+        }
+    )
+
+    assert received_otel_metrics(
+        {
+            "name": "tasks_longest_unblocked_time",
+            "description": "The age of the longest waiting task.",
+            "unit": "seconds",
+        }
+    )
```
mdellweg, line 397:
```suggestion
    )
    
    [pulpcore_bindings.TasksApi.cancel(href) for href in resident_task_hrefs]
```

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -349,3 +354,44 @@ def test_task_version_prevent_pickup(dispatch_task, pulpcore_bindings):
...
+    assert task.state == "waiting"
+
+    # And trigger the metrics
+    assert received_otel_metrics(
+        {
+            "name": "tasks_unblocked_queue",
+            "description": "Number of unblocked tasks waiting in the queue.",
+            "unit": "tasks",
+        }
+    )
```
mdellweg, line 389: What about the value of the metric?

pulpcore/tasking/worker.py
```diff
@@ -392,6 +419,42 @@ def handle_available_tasks(self):
...
+                )
+
+                self.tasks_unblocked_queue_meter.set(
+                    unblocked_tasks_stats["unblocked_tasks_count_gte_threshold"]
+                )
+                self.tasks_longest_unblocked_time_meter.set(
+                    unblocked_tasks_stats["longest_unblocked_waiting_time"].seconds
+                )
+
+                self.cursor.execute(f"NOTIFY pulp_worker_metrics_heartbeat, '{str(now)}'")
```
lubosmj, line 456: I would think we do not need to re-notify the worker, do we?

pulpcore/tasking/worker.py
```diff
@@ -392,6 +419,42 @@ def handle_available_tasks(self):
                 keep_looping = True
                 self.supervise_task(task)
 
+    def record_unblocked_waiting_tasks_metric(self):
```
lubosmj, line 422: Is there any safeguard that prohibits the execution of this method if the telemetry is disabled?

CHANGES/3821.feature
```diff
@@ -0,0 +1,3 @@
+Adds two new metrics related to Tasks. 
+`pulp_tasks_unblocked_waiting_queue` has the number of unblocked waiting tasks that have been waiting longer than five(5) seconds.
+And `pulp_tasks_longest_unblocked_waiting_time` has the time in seconds of the longest waiting time for a task in the queue.
```
lubosmj, line 3: Added two new metrics for Tasks: ``pulp_tasks_unblocked_waiting_queue`` now tracks the number of unblocked tasks waiting for over five (5) seconds, while ``pulp_tasks_longest_unblocked_waiting_time`` records the longest wait time, in seconds, for a task in the queue.

lubosmj (commented):
At this stage of the implementation, I think we can squash all the commits into a single one. What do you think?

pulpcore/tasking/worker.py
mdellweg: We notify all workers, that the work was done. So they hold off of doing it for another cooldown time.

mdellweg: But since you asked, a comment to that end may help.

decko: Thanks man. In the end I removed since it doesn't make sense to be used anymore.

pulpcore/tasking/worker.py
```diff
@@ -476,6 +473,5 @@ def run(self, burst=False):
                         break
                     self.sleep()
                 self.cursor.execute("UNLISTEN pulp_worker_wakeup")
-            self.cursor.execute("UNLISTEN pulp_worker_metrics_heartbeat")
```
mdellweg: Don't remove this logic. It is there to establish a cluster wide cooldown sending the metric.

pulpcore/tasking/worker.py
decko: Nope. It simply will not be emitted if the agent is not used.

pulpcore/tasking/worker.py
decko: Thanks man. My head is not working properly today.

pulpcore/tasking/worker.py
decko: Sorry @lubosmj. @mdellweg reminded me why we're using this. https://github.com/pulp/pulpcore/pull/5234#pullrequestreview-2041172514

CHANGES/3821.feature
decko: Used some of your suggestions man.

pulpcore/tests/functional/api/test_tasking.py
decko: As we talked before, this involves some changes to the test machinery, and yet we probably won't be able to catch the right value for the metric during the tests.
Talking with @lubosmj and @dkliban we understood that changing to use the OpenTelemetry Collector and test the metrics as it's exported to be consumed by Prometheus could generate better results. Yet, it's a considerable effort that is out of the scope of this task.

pulpcore/app/models/task.py
mdellweg, line 1: Try to avoid changing unrelated files.

pulpcore/tests/functional/assets/otel_server.py
```diff
@@ -8,6 +8,11 @@
 from aiohttp import web
 
 from opentelemetry.proto.trace.v1.trace_pb2 import TracesData
+from opentelemetry.proto.metrics.v1.metrics_pb2 import MetricsData
+
+_logger = logging.getLogger(__name__)
+c_handler = logging.StreamHandler()
```
mdellweg, line 14:
```suggestion
```

pulpcore/app/models/task.py
mdellweg: This file is still in the diff for unnecessary whitespace changes.

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -349,3 +354,48 @@ def test_task_version_prevent_pickup(dispatch_task, pulpcore_bindings):
...
+
+def test_emmiting_unblocked_task_telemetry(
+    dispatch_task, pulpcore_bindings, pulp_settings, received_otel_metrics
+):
+    if os.getenv("PULP_OTEL_ENABLED").lower() != "true":
+        pytest.skip("Need PULP_OTEL_ENABLED to run this test.")
+
+    # Checking online workers ready to get a task
+    age_threshold = timezone.now() - timedelta(seconds=pulp_settings.WORKER_TTL)
+    workers_online = pulpcore_bindings.WorkersApi.list(last_heartbeat__gte=age_threshold).count
```
mdellweg, line 367: AFAIK: There's "online" filter in the api. We should not repeat to implement that logic here.

pulpcore/tasking/worker.py
mdellweg: The question is valid. In the current state, we still run the query even if we send nothing.

pulpcore/tests/functional/api/test_tasking.py
decko: Thanks!

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -349,3 +351,47 @@ def test_task_version_prevent_pickup(dispatch_task, pulpcore_bindings):
...
+
+
+def test_emmiting_unblocked_task_telemetry(
+    dispatch_task, pulpcore_bindings, pulp_settings, received_otel_metrics
+):
+    if os.getenv("PULP_OTEL_ENABLED").lower() != "true":
+        pytest.skip("Need PULP_OTEL_ENABLED to run this test.")
+
+    # Checking online workers ready to get a task
+    workers_online = pulpcore_bindings.WorkersApi.list(online="true").count
```
mdellweg, line 363: :+1:

mdellweg (commented):
There's the last question remaining, how to avoid the costly db queries on systems that do not run with otel.

pulpcore/tasking/worker.py
ipanova: don't we have an otel_enabled setting we can look at?

lubosmj: Not yet, we have just an environment variable that can be used for a lookup (https://github.com/pulp/pulpcore/blob/455ce9a8442c380d6e0d6a2853152e6c0be77d34/docs/components.rst?plain=1#L168).

I can imagine doing something similar to [DomainMetricsEmitter.build()](https://github.com/pulp/pulpcore/blob/455ce9a8442c380d6e0d6a2853152e6c0be77d34/pulpcore/app/util.py#L564) here.

decko: I did it in the tests. Gonna work on the same thing here.

pulpcore/app/settings.py
```diff
@@ -312,6 +312,8 @@
 
 TASK_DIAGNOSTICS = False
 
+OTEL_ENABLED = False
```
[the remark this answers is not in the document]
decko: it works when we set a `PULP_OTEL_ENABLED` to `true`.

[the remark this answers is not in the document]
ipanova: So you have introduced a new setting here, and by this created a potential discrepancy in how otel can be enabled. Now we have an env var  and setting that enables otel. In the test you check whether env var is set but not the setting. There might combination of env var not set but setting enabled and viceversa.My suggestion is to stick to env var an in the code  by simply checking for `os.getenv("PULP_OTEL_ENABLED")`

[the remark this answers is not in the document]
decko: Nope, I changed in the tests @ipanova. The thing is that the `PULP_OTEL_ENABLED` is used by the container code to run the otel agent when it's needed, and also it populates the `OTEL_ENABLED` setting by dynaconf configuration.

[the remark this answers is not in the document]
decko: > So you have introduced a new setting here, and by this created a potential discrepancy in how otel can be enabled. Now we have an env var and setting that enables otel. In the test you check whether env var is set but not the setting. There might combination of env var not set but setting enabled and viceversa.My suggestion is to stick to env var an in the code by simply checking for `os.getenv("PULP_OTEL_ENABLED")`

Got your point.

lubosmj (approved):

pulpcore/app/settings.py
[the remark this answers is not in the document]
ipanova: you forgot to remove this from settings

ipanova (approved):

pulpcore/app/settings.py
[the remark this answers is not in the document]
decko: Thanks @ipanova

ipanova (approved):

mdellweg (approved):

PR #5247, opened 2024-04-11
Update django-storages upperbound to 1.14.2

gerrod3:
[noissue]


pulpcore/app/serializers/domain.py
```diff
@@ -167,9 +175,18 @@ class AmazonS3SettingsSerializer(BaseSettingsClass):
...
+    verify = serializers.BooleanField(allow_null=True, default=None)
     max_memory_size = serializers.IntegerField(default=0)
     default_acl = serializers.CharField(allow_null=True, default=None)
     use_threads = serializers.BooleanField(default=True)
+    # Not supported yet, requires instantiating a boto3.TransferConfig object
+    transfer_config = serializers.HiddenField(default=None)
+
+    def validate_verify(self, value):
+        """Verify can either be None or False."""
+        if value:
```
mdellweg, line 187: did you mean "if not value"?
how about `value = value or None`

pulpcore/app/serializers/domain.py
```diff
@@ -167,9 +175,18 @@ class AmazonS3SettingsSerializer(BaseSettingsClass):
...
+            value = None
+        return value
```
mdellweg, line 189:
```suggestion
        return value or None
```

pulpcore/app/serializers/domain.py
gerrod3: Nope, this is correct. I've updated the comment to better clarify the behavior.

pulpcore/app/serializers/domain.py
mdellweg: OK, I guess `value and None` would be equivalent then but maybe not more readable.

mdellweg (dismissed):

mdellweg (approved):

ggainey (approved):

PR #5252, opened 2024-04-13
Add support for transfer_config in S3 domain serializer

gerrod3:
This change will also allow me to add GCP support for domains since the `credentials` setting requires a Google object.

### DIFF
--- a/CHANGES/4592.misc
+++ b/CHANGES/4592.misc
@@ -1,0 +1,1 @@
+Added domain support for newest fields in django-storages 1.14.2.
--- a/pulpcore/app/models/domain.py
+++ b/pulpcore/app/models/domain.py
@@ -1,6 +1,6 @@
 from opentelemetry.metrics import Observation
 
-from django.core.files.storage import get_storage_class, default_storage
+from django.core.files.storage import default_storage
 from django.db import models
 from django_lifecycle import hook, BEFORE_DELETE, BEFORE_UPDATE, AFTER_CREATE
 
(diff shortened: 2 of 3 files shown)


pulpcore/app/models/domain.py
```diff
@@ -9,6 +9,9 @@
 
 from .fields import EncryptedJSONField
 
+# Global used to store instantiated storage classes to speed up lookups across domains
+storages = {}
```
mdellweg, line 13: This is a per worker in memory cache, right? So a bit like lru_cache just with the last update check.

pulpcore/app/serializers/domain.py
```diff
@@ -197,10 +229,18 @@ def validate(self, data):
             )
         return data
 
+    def create(self, validated_data):
+        """Special handling for TransferConfig if set."""
+        if config := validated_data.get("transfer_config"):
+            transfer = self.fields.get("transfer_config").create(config)
```
mdellweg, line 235:
```suggestion
            transfer = self.fields["transfer_config"].create(config)
```

pulpcore/app/serializers/domain.py
```diff
@@ -197,10 +229,18 @@ def validate(self, data):
             )
         return data
 
+    def create(self, validated_data):
+        """Special handling for TransferConfig if set."""
+        if config := validated_data.get("transfer_config"):
```
mdellweg, line 234:
```suggestion
        if transfer_config := validated_data.get("transfer_config"):
```

CHANGES/4592.misc
```diff
@@ -0,0 +1 @@
+Added domain support for newest fields in django-storages 1.14.2
```
mdellweg, line 1:
```suggestion
Added domain support for newest fields in django-storages 1.14.2.
```

pulpcore/app/models/domain.py
gerrod3: Yep, that is correct. Since `get_storage` is called everytime an artifact is saved/read, it makes sense to have a cache to speed up the access.

mdellweg (approved):

ggainey (approved):

PR #5266, opened 2024-04-16
Add unified docs-ci

pedro-psb:
Related to: https://github.com/pulp/pulp-docs/issues/31

[noissue]


.github/workflows/test.yml
```diff
@@ -23,6 +23,7 @@ jobs:
         env:
           - TEST: pulp
           - TEST: docs
+          - TEST: new_docs
```
mdellweg, line 26: I do not believe this is in the plugin template. Please use the current version.

mdellweg (changes requested):

.github/workflows/test.yml
mdellweg: Not that this never happened to me...

mdellweg (approved):

mdellweg:
I'm wondering if/when we want to stop building the docs the old way.

pedro-psb:
We are planning to officially deactivate them on July, as a general strategy.
https://github.com/pulp/pulp-docs/issues/33#issuecomment-2050039066

ggainey (approved):

PR #5269, opened 2024-04-16
Adds a Liveness API

### LINKED ISSUE #5243: Have a lean way to check for the health of a pulp instance
Right now when we want to check for how healthy is a pulp instance we hit the `/status/` endpoint which returns:
- The pulpcore version
- plugins installed and its versions
- space available
- number of api/content/workers
- db/redis connection
- domain enabled
- and so on...

Although this information is really useful when you're interacting with a pulp instance, for automated systems that
need to check it from time to time this doesn't seems to be need or even wanted.

What if Pulp could be checked with just a HEAD request? 200 if a minimum number of workers are online and a 400 class if something is wrong?
No body on the response. No need to serialize anything.
Just the minimal amount of things needed to understand that Pulp is alive and kicking.

We could have the HEAD method on the status endpoint itself.
HEAD if you just wanted a lean check, and GET if you need all possible data from your instance.

This example if just to provoke the discussion:
```python
    def head(self, *args):
        if settings.CACHE_ENABLED:
            redis_status = self._get_redis_conn_status()
        else:
            redis_status = False

        db_status = self._get_db_conn_status()

        online_workers = Worker.objects.online().exists()
        online_api_apps = ApiAppStatus.objects.online().exists()
        online_content_apps = ContentAppStatus.objects.online().exists()

        healthy_stack = all(
            [online_content_apps, online_api_apps, online_workers, db_status, redis_status]
        )

        if healthy_stack:
            return Response(status=200)

        return Response(status=409)
```

dkliban:
This API is intended to be used as the livenessProbe for the API workers runnign on Kubernetes.

closes #5243

### DIFF
--- a/pulpcore/tests/functional/api/test_status.py
+++ b/pulpcore/tests/functional/api/test_status.py
@@ -175,3 +175,15 @@ def verify_get_response(status, expected_schema):
     else:
         assert status["storage"]["free"] is not None
         assert status["storage"]["total"] is not None
+
+
+@pytest.mark.parallel
+def test_livez_unauthenticated(
+    pulpcore_bindings,
+    anonymous_user,
+):
+    """
+    Assert that GET requests to Livez API return 200 without a response body.
+    """
+    with anonymous_user:
+        assert pulpcore_bindings.LivezApi.livez_read() is None
--- a/CHANGES/5243.feature
+++ b/CHANGES/5243.feature
@@ -1,0 +1,1 @@
+Added a Liveness API that can be used for the livenessProbe in k8s.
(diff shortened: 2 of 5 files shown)


pulpcore/tests/functional/api/test_status.py
```diff
@@ -175,3 +175,16 @@ def verify_get_response(status, expected_schema):
...
+def test_livez_unauthenticated(
+    pulpcore_bindings,
+    anonymous_user,
+):
+    """
+    Assert that GET requests to Livez API return 200 without a response body.
+    """
+    with anonymous_user:
+        r = pulpcore_bindings.LivezApi.livez_read()
+        assert r is None
```
mdellweg, line 190: When we expect the reader to understand the meaning of "r" because it must be obvious in the context, we don't even need to give the value a name.

```suggestion
        assert pulpcore_bindings.LivezApi.livez_read() is None
```

ggainey (approved):

dralley (approved):

PR #5274, opened 2024-04-17
Allow gunicorn/22.0.0 to pick up gunicorn security fixes.

ggainey:
[noissue]


mdellweg (approved):

ggainey:
This is so we can pick up the new gunicorn with the fix for https://github.com/advisories/GHSA-w3h3-4rj7-4ph4

PR #5280, opened 2024-04-17
Filter packages by if they are not in a repository version

### LINKED ISSUE #5241: Ability to filter packages by added/removed between two repository versions
Currently, pulp supports filtering packages by if they were added or removed since a repository version with the `repository_version_added` and `repository_version_removed` parameters. Extending this to support filtering packages by if they were added or removed between two repository versions would accelerate our server where we need this information. For example, a request to pulp with `repository_version_added=repoversion1,repoversion2` would list packages that were not in `repoversion1` but in `repoversion2`.

adrianabedon:
closes #5241

### DIFF
--- a/CHANGES/5241.feature
+++ b/CHANGES/5241.feature
@@ -1,1 +1,1 @@
-Add support for filtering packages by if they were added/removed between repository versions
+Add support for filtering packages by if they are not in a repository version
--- a/CHANGES/5241.feature
+++ b/CHANGES/5241.feature
@@ -1,0 +1,1 @@
+Add support for filtering packages by if they were added/removed between repository versions
--- a/pulpcore/app/viewsets/content.py
+++ b/pulpcore/app/viewsets/content.py
@@ -21,6 +21,7 @@ from .custom_filters import (
     ContentAddedRepositoryVersionFilter,
     ContentRemovedRepositoryVersionFilter,
     ContentRepositoryVersionFilter,
+    ContentNotInRepositoryVersionFilter,
 )
 
 
@@ -102,6 +103,8 @@ class ContentFilter(BaseFilterSet):
 
         repository_version:
             Return Content which is contained within this repository version.
+        repository_version_not_in:
+            Return Content which is not contained within this repository version.
         repository_version_added:
             Return Content which was added in this repository version.
         repository_version_removed:
@@ -112,6 +115,7 @@ class ContentFilter(BaseFilterSet):
     """
 
     repository_version = ContentRepositoryVersionFilter()
+    repository_version_not_in = ContentNotInRepositoryVersionFilter()
     repository_version_added = ContentAddedRepositoryVersionFilter()
     repository_version_removed = ContentRemovedRepositoryVersionFilter()
     orphaned_for = OrphanedFilter(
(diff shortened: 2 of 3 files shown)


dkliban (changes requested):
Thank you for the contribution. Please add a test in https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/api/using_plugin/test_repo_versions.py

mdellweg:
This is missing documentation too. Also for the sake of REST api design, i wonder whether a dedicated `added_between`  range filter would serve us better.

dkliban:
> This is missing documentation too. Also for the sake of REST api design, i wonder whether a dedicated `added_between` range filter would serve us better.

That would make for a better user experience. @adrianabedon what do you think about making this it's own filter?

mdellweg:
Maybe even one `added_before` and one `added_after` to give even more flexibility.
(how would you name `added_before` when you want to include the specified version?)
Are we searching for an `added_version__lte` filter?

adrianabedon:
I can go ahead make a separate filter and add some tests. 

@mdellweg would you be able to explain the `added_before` and `added_after` filters in more detail? If I understand correctly, a good name for these filters would be `repository_version_added__lte` and `repository_version_added__gt`.

I am confused how these should work. I imagine that specifying `repository_version_added__lte=rv4` would filter for packages added in `rv1`, `rv2`, `rv3`, or `rv4`. Specifying `repository_version_added__gt=rv4` would filter for packages added in `rv5`, `rv6`, ..., `latest`. Now what happens if we specify both `repository_version_added__gt=rv2 repository_version_added__lte=rv5`, which I assume is supposed to replace `repository_version_added_between=rv2,rv5`. Would this filter for packages added in `rv3`, `rv4`, or `rv5`? 

This is different than the original intent of these filter additions which was to get packages not in `rv2` but in `rv5`. 

I think what is closer to the original changes would be to add a new filter `repository_version__ne`. Specifying `repository_version__ne=rv2` would filter for packages that are not in `rv2`. Then, specifying `repository_version__ne=rv2 repository_version=rv4` would filter for packages that are not in `rv2` but in `rv4`. 

I am also open to suggestions on a better name for `repository_version__ne`.

adrianabedon:
I went ahead and changed the pr to add a new `repository_version_not_in` filter. If you think the changes are good, then I can go ahead and make some tests.

mdellweg:
So maybe i misunderstood the original idea.
And yeah, if you want to know the difference between rv5 and rv2 you do not care about packages added in rv3 and removed again in rv4. But you do care about packages added in rv4 and not removed again.
Before going deeper into any implementation, can you try the "q" filter?
`...?q="repository_version=rv5 AND NOT repository_version=rv2"` (This may need some level urlencoding).

daviddavis:
Intersting, did not know about this filter. I guess it must be new[er]? I tested it out and it does seem to work from what I can tell. The syntax I had to use was:

```
https :/pulp/api/v3/content/deb/package_release_components/ q=="repository_version = '/pulp/api/v3/repositories/deb/apt/018ea572-b195-7138-87b3-4e3e315f3f28/versions/8/' AND NOT repository_version = '/pulp/api/v3/repositories/deb/apt/018ea572-b195-7138-87b3-4e3e315f3f28/versions/2/'"
```

Hopefully @adrianabedon can confirm that it meets our needs. Looking at the docs, it seems like [it's in tech preview](https://docs.pulpproject.org/pulpcore/workflows/complex-filtering.html). Is that still the case?

adrianabedon:
The "q" filter works for our needs, thanks for pointing us to it. It would be great if it was taken out of tech preview.

mdellweg:
Closing as non planned. "q" is there for a solution.

Issue #5285, opened 2024-04-19
Add support for reporting to GlitchTip (Sentry)


dkliban:
When sentry-sdk is installed, the settings file should configure sentry-sdk. The user will need to provide a `SENTRY_DSN` environment variable to configure where the settings will be sent. When `SENTRY_DSN` is not set, the application will simply not send anything to sentry.

PR #5296, opened 2024-04-22
Adds integration with Sentry/GlitchTip

dkliban:
closes: #5285


dkliban:
This works with both the api and content app.

staging_docs/admin/guides/integrate-sentry-glitchtip.md
```diff
@@ -0,0 +1,6 @@
+# Integrate with Sentry or GlitchTip
+
+Pulp can be configured to report exceptions to Sentry or GlitchTip. 
+
+First install the `sentry-sdk` package.
+Then Set the `SENTRY_DSN` environment variable to configure where Pulp will report exceptions.
```
mdellweg, line 6: Does this need to be an environment variable? Can we not make this a pulp setting to gain the flexibility of dynaconf?

pulpcore/app/settings.py
```diff
@@ -22,6 +22,14 @@
 
 from pulpcore import constants
 
+
+try:
+    import sentry_sdk
+
+    sentry_sdk.init()
+except ImportError:
+    pass
```
mdellweg, line 31: Should we guard this by whether it is configured?
What do it do to workers and pulpcore-manager?
Maybe the entrypoints are a better place.

staging_docs/admin/guides/integrate-sentry-glitchtip.md
dkliban: The current implementation needs it to be an environment variable. This is the environment variable that `sentry-sdk` looks for. However, we could add a pulp setting called `SENTRY_DSN` and then pass the value of it to the `sentry_sdk.init()` call.

pulpcore/app/settings.py
dkliban: `sentry-sdk` will not do anything if a DSN is not defined. The DSN is defined either in the optional parameter passed to `sentry_sdk.init(dsn="blah")` or with a `SENTRY_DSN` environment variable.

dkliban: The current implementation does not report an exception in a worker. I will fix that.

dkliban: I have now confirmed that exceptions emited outside of tasks in worker code are actually reported to GlitchTip. no changes needed.

staging_docs/admin/guides/integrate-sentry-glitchtip.md
```diff
@@ -0,0 +1,9 @@
+# Integrate with Sentry or GlitchTip
+
+Pulp can be configured to report unhandled exceptions to Sentry or GlitchTip.
+All unhandled exceptions in `pulpcore-api` and `pulpcore-content` are reported to GlitchTip.
```
decko, line 4:
```suggestion
All unhandled exceptions in `pulpcore-api`, `pulpcore-content` and `pulpcore-worker` are reported to GlitchTip.
```

staging_docs/admin/guides/integrate-sentry-glitchtip.md
```diff
@@ -0,0 +1,9 @@
...
+Exceptions encountered during task execution are not reported to GlitchTip.
```
decko, line 5: Maybe remove this line?

staging_docs/admin/guides/integrate-sentry-glitchtip.md
```diff
@@ -0,0 +1,9 @@
...
+Tasks API should be used to view exceptions encountered during task execution.
```
decko, line 6: Same as above?

decko (approved):

dkliban:
@decko thanks for the review! I wanted the documentation to make it clear that at this time any exceptions that cause a task to fail are going to reported in the task status API. Those exceptions are handled by the worker code and sentry-sdk doesn't detect them. However, in the future we may want to add code to the worker to send the exception to sentry.

mdellweg (approved):

PR #5309, opened 2024-04-25
Report the size of served artifacts

### LINKED ISSUE #4602: Add additional metrics reporting size of artifacts served

lubosmj:
closes #4602

### DIFF
--- a/CHANGES/4602.feature
+++ b/CHANGES/4602.feature
@@ -1,0 +1,1 @@
+Added metrics reporting the size of served artifacts.
--- a/docs/components.rst
+++ b/docs/components.rst
@@ -193,6 +193,7 @@ Collector here <https://opentelemetry.io/docs/collector/>`_.
 * Access to every API endpoint (an HTTP method, target URL, status code, and user agent).
 * Access to every requested package (an HTTP method, target URL, status code, and user agent).
 * Disk usage within a specific domain (total used disk space and the reference to the domain).
+* The size of served artifacts.
 
 The information above is sent to the collector in the form of spans and metrics. Thus, the data is
 emitted either based on the user interaction with the system or on a regular basis. Consult
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -957,14 +957,38 @@ class Handler:
                         params[STORAGE_RESPONSE_MAP[storage_domain][a_key]] = hdrs[a_key]
             return params
 
+        def _build_url(**kwargs):
+            filename = os.path.basename(content_artifact.relative_path)
+            content_disposition = f"attachment;filename={filename}"
+
+            headers["Content-Disposition"] = content_disposition
+            parameters = _set_params_from_headers(headers, domain.storage_class)
+            storage_url = artifact_file.storage.url(artifact_name, parameters=parameters, **kwargs)
+
+            return URL(storage_url, encoded=True)
+
         artifact_file = content_artifact.artifact.file
         artifact_name = artifact_file.name
-        filename = os.path.basename(content_artifact.relative_path)
-        content_disposition = f"attachment;filename={filename}"
         domain = get_domain()
-        storage = domain.get_storage()
+
+        content_length = artifact_file.size
+
+        try:
+            range_start, range_stop = request.http_range.start, request.http_range.stop
+            if range_start or range_stop:
+                if range_stop and artifact_file.size and range_stop > artifact_file.size:
+                    start = 0 if range_start is None else range_start
+                    content_length = artifact_file.size - start
+                elif range_stop:
+                    content_length = range_stop - range_start
+        except ValueError:
+            size = artifact_file.size or "*"
+            raise HTTPRequestRangeNotSatisfiable(headers={"Content-Range": f"bytes */{size}"})
+
+        headers.update({"X-PULP-ARTIFACT-SIZE": str(content_length)})
 
         if domain.storage_class == "pulpcore.app.models.storage.FileSystem":
+            storage = domain.get_storage()
             path = storage.path(artifact_name)
             if not os.path.exists(path):
                 raise Exception(_("Expected path '{}' is not found").format(path))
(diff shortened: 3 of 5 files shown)


lubosmj:
Continuation of https://github.com/pulp/pulpcore/pull/5169. This PR was opened for transparency purposes and it is not meant to be reviewed yet.

Here, I report metrics directly from the instrumentation middleware. However, it seems like the correctness of this approach is questionable because the collector might receive data from multiple instances of the counter and the size of served artifacts is not growing:

image

Seeing both of the implementations, I am suggesting to lean towards the following approach:
- store the value of served artifacts in some persistent storage with locking capabilities so no race conditions would appear by default
- this value will be increased every time we serve the content from Pulp
- once in a while, we will emit this value to the OTEL collector (like we do now for the accumulated size of domains) by leveraging the asynchronous counter instrument

lubosmj:
Thinking about it more, does it make sense to report such metrics (i.e., the size of served data) if users have knowledge of egress traffics? I would like to believe that 3rd party content storages have capabilities to record such data in general.

bmbouter:
The tools for reporting cost for things like S3 for example don't report on a per-bucket level. Our needs are to have it per domain (which is really per bucket). Also even if it did report in S3 per-bucket, there are other complexities such as a) delays in reporting and b) permissions are needed for users to access those cost reports which get pretty complex on AWS.

I think based on our needs having this data report into OTEL to let PROMQL query it would be ideal.

decko:
To clarify what's expected outcomes of this task:

```
    How much data did I serve for this Artifact request
    Which Repository (or at least Distribution) is this Artifact member of?
    Which Domain (if any) is this Artifact a member of?
```
which on my understanding is "we don't want to see individual data about each artifact served, but the aggregation of all artifacts served. If we use the repository/distribution and the domain as labels/attributes, we can get the information that we need.

Also @lubosmj, I believe that you're right about using a `Counter`. This metric is monotonic, as it will only increase over time.

bmbouter:
@decko I agree with your recap of what we need.

@lubosmj I totally missed the implementation question, but that's probably for the best as I really don't understand the specifics of OTEL data reporting types and aggregation.

If I've missed answering a question, please let me know because this work is important and I want to share info to unblock it.

mdellweg:
Do we take ranged requests in to account?

lubosmj:
> Do we take ranged requests in to account?

Can you elaborate on it? I guess an artifact is served from content-app as it is. At least, we know its size before returning any response. I record the size of an artifact after we retrieve information about it from the database.

mdellweg:
We serve only part of the artifact, when the client requests a range (e.g. for chunked or continued downloads).

lubosmj:
> We serve only part of the artifact, when the client requests a range (e.g. for chunked or continued downloads).

I understand. Though, if I get the size of an artifact from `artifact_file.size` in [_serve_content_artifact](https://github.com/pulp/pulpcore/blob/7297b25d5d62702c7f0762109ddd775daaf942a4/pulpcore/content/handler.py#L960) or `size` in [_stream_remote_artifact](https://github.com/pulp/pulpcore/blob/7297b25d5d62702c7f0762109ddd775daaf942a4/pulpcore/content/handler.py#L1027), we do not need to worry about the ranges, do we?

mdellweg:
"How much data did I serve for this Artifact request" That sounds very much like the requested range to me.

lubosmj:
> Each data point that is reported has a 'http_host' label on it. So even if the collector receives data about the same artifact from two different processes, it will store them as two separate data points.

@dkliban, I see. The problem we are trying to resolve now is that there might exist multiple instances of meters (possibly per each worker/process or even request in the worst case). These meters seem to fight against each other which results in reporting ambiguous data (?). As demonstrated in the screenshot, the visualization shows a non-growing curve for the number of artifacts served when a user downloads content in parallel.

lubosmj:
> "How much data did I serve for this Artifact request" That sounds very much like the requested range to me.

@mdellweg, ah, so we should also **consider the requested range if provided**. Otherwise, we should be fine. Thanks for the input.

Also, another question is how can we ensure reporting a valid size of served artifacts if we issue redirects to the object storage? We do not know if the client would follow the redirect but we still need to report the size of a "potentially" served artifact.

mdellweg:
> Also, another question is how can we ensure reporting a valid size of served artifacts if we issue redirects to the object storage? We do not know if the client would follow the redirect but we still need to report the size of a "potentially" served artifact.

And the redirect link could be used multiple times until it expires.

bmbouter:
The possibility that the client could not follow the redirect or follow it several times is true.

I think it's ok to assume they will though. Under normal operation the redirect is followed once because typically the client is a cli like yum/dnf which will follow the redirect and follow it once.

decko:
> The possibility that the client could not follow the redirect or follow it several times is true.
> 
> I think it's ok to assume they will though. Under normal operation the redirect is followed once because typically the client is a cli like yum/dnf which will follow the redirect and follow it once.

Also, it's a thing that's not under our control. I think we could consider an "error margin" for it given the possibilities described above.

lubosmj:
As per our recent conversation with @dkliban, I am going to report the size of served artifacts by leveraging the [synchronous gauge](https://opentelemetry.io/docs/specs/otel/metrics/api/#gauge) instead of counter. These metrics will be afterwards grouped and summed by the prometheus server for further visualization. We will just report raw data and leave the evaluation on the prometheus server instead of doing all the computing on the Pulp's side.

lubosmj:
Unfortunately, gauges are not working in this scenario because of the nature of the instrument, making the emitted metrics effectively indistinguishable.

Together with @decko, we decided to fall back to the counter. What I had observed in the past were two counters emitting different value with the same attributes, as pointed out by @decko -- one counter per one content-app instance (gunicorn runner) fought against each other. Now, we plan to add the name of content-app to the counter attributes to distinguish the emitted metrics properly to allow further aggregations. Users can then sum the values of these counters to get the total value per domain, like so: `sum by(domain) (pulp_artifacts_size_counter_bytes_total)`. Preliminary testing has shown that this approach works.

image

lubosmj:
Closing in favour of: https://github.com/pulp/pulpcore/pull/5338. It appeared it is not that easy to fetch the name of a domain from within the instrumentation middleware.

So, I decided to implement a custom metrics class (like in the past) that holds the reference to the meter inside the pulpcore-content handler and is able to record all required information about domains and artifacts when initializing responses.

PR #5338, opened 2024-05-09
Report the size of served artifacts

### LINKED ISSUE #4602: Add additional metrics reporting size of artifacts served

lubosmj:
closes #4602


lubosmj:
This approach seems to be working properly with both ranged requests and regular requests:

image

pulpcore/app/util.py
```diff
@@ -560,16 +532,47 @@ def __getattr__(self, *args, **kwargs):
             return self
 
     @classmethod
-    def build(cls, domain):
+    def build(cls, *args, **kwargs):
         otel_enabled = os.getenv("PULP_OTEL_ENABLED")
         if otel_enabled == "true" and settings.DOMAIN_ENABLED:
```
decko, line 537: What if you use a `.lower` method on this string? So if somehow someone set it to `True`, this line can still get it properly.

decko: Nevermind. I've created a `OTEL_ENABLED` setting on #5234. If it got merged before this, or if you want to reproduce it here to not get in conflict, it's a go-go.

lubosmj: The setting was removed from the PR. We can leave the code untouched.

decko (approved):

pulpcore/app/util.py
decko: Yep, so it's still good to use the `.lower()` here to properly check for `true`.

lubosmj: I just added the .lower() method. Can you review the PR once again?

decko (approved):

pulpcore/content/handler.py
```diff
@@ -957,40 +984,51 @@ def _set_params_from_headers(hdrs, storage_domain):
                         params[STORAGE_RESPONSE_MAP[storage_domain][a_key]] = hdrs[a_key]
             return params
 
+        def _build_url(**kwargs):
+            filename = os.path.basename(content_artifact.relative_path)
+            content_disposition = f"attachment;filename={filename}"
+
+            headers["Content-Disposition"] = content_disposition
+            parameters = _set_params_from_headers(headers, domain.storage_class)
+            storage_url = artifact_file.storage.url(artifact_name, parameters=parameters, **kwargs)
```
gerrod3, line 993: Since you are refactoring this, how about:
```suggestion
            storage_url = storage.url(artifact_name, parameters=parameters, **kwargs)
```
and then keep the `storage = domain.get_storage()` line where it was.

gerrod3 (commented):
Minor suggestion

pulpcore/content/handler.py
lubosmj: Done! Even though we now run on a bit more CPU ticks.

bmbouter:
@lubosmj what do you need for this to merge?

lubosmj:
It seems like there have been merged changes that modified code utilized by this PR. I am going to resolve merge conflicts and get back to you soon-ish. Thanks for your patience! 🥇

lubosmj:
@decko, @bmbouter, @dkliban, I rebased my changes and the PR should be once again ready for review.

image

A dashboard looks like this ^. After sending requests in parallel:
```
pulp file remote create --name test --url https://fixtures.pulpproject.org/file/PULP_MANIFEST --policy on_demand
pulp file repository create --name test --remote test
pulp file repository sync --name test
pulp file publication create --repository test
pulp file distribution create --name test --base-path test --repository test

# wait...

http http://localhost:5001/pulp/content/default/test/1.iso & http http://localhost:5001/pulp/content/default/test/2.iso & http http://localhost:5001/pulp/content/default/test/3.iso &

# wait...

http http://localhost:5001/pulp/content/default/test/1.iso 'Range: bytes=0-50'

http http://localhost:5001/pulp/content/default/test/1.iso & http http://localhost:5001/pulp/content/default/test/2.iso & http http://localhost:5001/pulp/content/default/test/3.iso &
```

pulpcore/content/handler.py
```diff
@@ -154,6 +167,20 @@ class Handler:
 
     distribution_model = None
 
+    class ArtifactsSizeCounter(MetricsEmitter):
+        def __init__(self):
+            self.meter = metrics.get_meter("artifacts.size.meter")
+            self.counter = self.meter.create_counter(
```
dkliban, line 173: Does the counter get reset after each artifact is served? or the is the value of this counter increase with each artifact served?

lubosmj: It is initialized just once per a running instance. So, it increases.

pulpcore/app/util.py
```diff
@@ -571,19 +536,58 @@ def __getattr__(self, *args, **kwargs):
             return self
 
     @classmethod
-    def build(cls, domain):
+    def build(cls, *args, **kwargs):
         otel_enabled = os.getenv("PULP_OTEL_ENABLED")
```
decko, line 540: What if you change here to:
```suggestion
        otel_enabled = os.getenv("PULP_OTEL_ENABLED", "").lower() == "true"
```

It will set `otel_enabled` as a boolean. Then on 541 you can just check for `otel_enabled`.

pulpcore/content/handler.py
```diff
@@ -957,13 +984,37 @@ def _set_params_from_headers(hdrs, storage_domain):
...
-        filename = os.path.basename(content_artifact.relative_path)
-        content_disposition = f"attachment;filename={filename}"
         domain = get_domain()
         storage = domain.get_storage()
 
+        content_length = artifact_file.size
+
+        try:
+            range_start, range_stop = request.http_range.start, request.http_range.stop
+            if range_start or range_stop:
```
decko, line 1006: Is there any situation where `range_start` or `range_stop` doesn't exist? or be `0`?

pulpcore/content/handler.py
```diff
@@ -1108,6 +1146,11 @@ async def finalize():
         downloader.finalize = finalize
         download_result = await downloader.run()
 
+        if content_length := response.headers.get("Content-Length"):
+            self._report_served_artifact_size(int(content_length))
```
decko, line 1150: Any change for the otel sdk complain about it being a bigint?

lubosmj: It can export values as large as 200 TB:
```
otel-collector    | InstrumentationScope artifacts.size.meter 
otel-collector    | Metric #0
otel-collector    | Descriptor:
otel-collector    |      -> Name: artifacts.size.counter
otel-collector    |      -> Description: Counts the size of served artifacts
otel-collector    |      -> Unit: Bytes
otel-collector    |      -> DataType: Sum
otel-collector    |      -> IsMonotonic: true
otel-collector    |      -> AggregationTemporality: Cumulative
otel-collector    | NumberDataPoints #0
otel-collector    | Data point attributes:
otel-collector    |      -> domain_name: Str(default)
otel-collector    |      -> content_app_name: Str(1657@793c15a2a12b)
otel-collector    | StartTimestamp: 2024-06-24 18:54:52.343138526 +0000 UTC
otel-collector    | Timestamp: 2024-06-24 18:55:09.034996534 +0000 UTC
otel-collector    | Value: 200000000000000
otel-collector    | 	{"kind": "exporter", "data_type": "metrics", "name": "logging"}
```

Larger values are not allowed:
```
pulp_1            | pulp [None]: opentelemetry.sdk.metrics._internal.export:ERROR: Exception while exporting metrics Value out of range: 2000000000000000000000000
```

Max accumulated value is defined by int64: 9223372036854775807 (9223372 TB). I guess we are fine with that.

pulpcore/content/handler.py
lubosmj: Yes, when you issue standard requests without `Range` header.

`Range: bytes=0-0' is still working fine:
```
http http://localhost:5001/pulp/content/default/test/1.iso 'Range: bytes=0-0'
```

pulpcore/app/util.py
lubosmj: Done! 🎉

decko (approved):

dkliban (changes requested):
Looks like this always reports the cumulative bytes served. We should report the delta so that we can always sum up the data points for some period of time.

lubosmj:
So, I have tried visualizing data once again with the current implementation. I think we can use the counter instrument to fulfill our needs. Please, look at the second graph and compare its values with the first one.

#### (1.) The regular cumulative one (may not be wanted by customers)
Screenshot from 2024-07-15 16-23-40

#### (2.) Showing delta differences across 30m span (may be wanted by customers)
Screenshot from 2024-07-15 16-23-22

Do you think the latter one is the right one? During the testing, I have used this environment variable: `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE="delta"`.

decko:
Can we put this behind a feature flag? I believe we need to test this with real data before making it work as default.

decko (approved):
Would love to have this under a feature flag to enable or disable at our will.

lubosmj:
@dkliban, should we introduce this feature flag to enable a better level of detail for the configuration?

@decko, where did you want to set `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE="delta"`? Inside the newly added class? Does this setting have a bad impact on our existing existing metrics (duration_histogram, active_requests_counter) if set globally (https://opentelemetry.io/docs/specs/otel/metrics/sdk_exporters/otlp/#additional-configuration)?

mdellweg:
We have all of OTEL behind a feature flag I was told. I think that should suffice. Every additional switch reduces the consumption rate exponentially.

lubosmj:
I am going to create a custom exporter that will have this setting defined. We are not going to create a new feature flag. Putting this PR into draft for now.

lubosmj:
I tested this once AGAIN and noticed that implementation works even without customizing the `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` setting. So, I am undrafting this PR AGAIN.

image
```
sum by(domain_name) (increase(pulp_artifacts_size_counter_Bytes_total[3m]))
```

image
```
sum by(domain_name) (pulp_artifacts_size_counter_Bytes_total)
```

lubosmj:
It looks like setting `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` to `delta` globally was not even needed and the behaviour remained the same without it.

Screenshot from 2024-08-02 14-39-55
The graph on top is reporting the accumulated value of the served artifacts: `sum by(domain_name) (pulp_artifacts_size_counter_Bytes_total)`. The one on the bottom is reporting the deltas within 2 minutes range: `sum by(domain_name) (increase(pulp_artifacts_size_counter_Bytes_total[2m]))`. The graphs looked fine until I found out that the bottom one actually reports the served size as 18k bytes instead of 9k bytes. No idea how to fix that. The same problem occurs when changing the time range to higher values (e.g., 3m, etc.).

Screenshot from 2024-08-02 12-19-19
Screenshot from 2024-08-02 12-19-04
For instance, the last increase reports correctly a jump from 18k to 24k (6k). However, the spike on the bottom graph says that it received 9k. Maybe, there is some correlation between the time range and data received. No idea what is going on.

lubosmj:
When I followed [this documentation](https://docs.datadoghq.com/opentelemetry/guide/otlp_delta_temporality/) to set the temporality manually, it behaved the similar way.
```python
        deltaTemporality = {
            Counter: AggregationTemporality.DELTA,
            ...
        }

        def __init__(self):
            exporter = OTLPMetricExporter(preferred_temporality=self.deltaTemporality)
            reader = PeriodicExportingMetricReader(exporter)
            provider = MeterProvider(metric_readers=[reader])

            self.meter = provider.get_meter("artifacts.size.meter")
            self.counter = self.meter.create_counter(
```


See the graphs below where I use the same queries as written in the comment above. In total, we were supposed to sent 27,648B + 3072B from Pulp over 10 minutes (i.e., running this 3 times: sending 3 parallel requests to download 3 files of the size 1024B + additional 3 * 1024B).
image

image

Overall, it looks the like the temporality is ignored by the python SDK. We should somehow get the right visualization from the data we already have in prometheus.

lubosmj:
Let me add one more note to the delta temporality. Prometheus currently does not support that: https://github.com/prometheus/prometheus/issues/12763. Pulp correctly exports delta values when opted for. So, we need to work with what we have (i.e., cumulative values). I am going to ask for help on forums.

lubosmj:
Actually, I was able to make this work (starting from 6k):
image

image

Query used: `sum by(domain_name) (rate(pulp_artifacts_size_counter_Bytes_total[2m]) * 60)` (inspired by https://www.innoq.com/en/blog/2019/05/prometheus-counters/).

The graph correctly reports the increases. So, I think we can merge this work.

ipanova:
from the team meeting : converting to draft, needs more discussion @dkliban and @decko

lubosmj:
The functionality will remain the same. The only thing that needs to be addressed later is a proper query for larger ranges. Every limitation was mentioned in the comments earlier.

decko:
> The functionality will remain the same. The only thing that needs to be addressed later is a proper query for larger ranges. Every limitation was mentioned in the comments earlier.

What do you think about splitting this PR between the feature and the queries and graphs? With that I believe this PR is ready to be merged.

decko (approved):

PR #5365, opened 2024-05-13
Stop task cleanup to delete Contents and Artifacts

### LINKED ISSUE #5363: Task cleanup must not delete content nor artifacts
Deleting content or artifacts outside of orphan cleanup is breaking the rules.
And no, we cannot get away with that.

mdellweg:
These shared objects that may not even have been created by that task must only ever be deleted inside of orphan cleanup. That's the rule.

fixes #5363

### DIFF
--- a/CHANGES/5363.bugfix
+++ b/CHANGES/5363.bugfix
@@ -1,0 +1,2 @@
+Stopped deleting content and artifacts presumably created by later failed or canceled tasks.
+Deleting these lies solely in the responsibility of orphan cleanup.
--- a/pulpcore/tasking/_util.py
+++ b/pulpcore/tasking/_util.py
@@ -15,7 +15,7 @@ from django.db.models import Q
 from django.utils import timezone
 from django_guid import set_guid
 from django_guid.utils import generate_guid
-from pulpcore.app.models import Task, TaskSchedule
+from pulpcore.app.models import Artifact, Content, Task, TaskSchedule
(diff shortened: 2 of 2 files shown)


CHANGES/5363.bugfix
```diff
@@ -0,0 +1 @@
+Stopped deleting content and artifacts presumably created by later failed or canceled tasks.
```
ipanova, line 1: Can you add that these are subject to deletion only by orphan clean up task?

lubosmj: Yes, please, make the change more explicit. How does this sound:
```suggestion
Stopped deleting content and artifacts associated with subsequent task failures or cancellations. File removal now exclusively occurs during the orphan cleanup.
```

mdellweg: The thing is the content may not even be created by that task if it existed, but because we cannot return something from a task we add it as a created resource anyway. But this is a different story.

CHANGES/5363.bugfix
```diff
@@ -0,0 +1,2 @@
...
+Deleting these lies solely in the responsibility of orphan cleanup.
```
mdellweg, line 2: (Read your comment too late. Tell me what you think.) @lubosmj

lubosmj (approved):

ipanova (approved):

PR #5368, opened 2024-05-14
Add some indices to task table

### LINKED ISSUE #5367: Add indices to task table to speed up worker queries.
Specifically dispatching with `immediate=True` yields a query like:
`SELECT 1 AS "a" FROM "core_task" WHERE ("core_task"."pulp_created" < '2024-05-13 23:29:28.153293+00:00'::timestamptz AND "core_task"."state" IN ('waiting', 'running', 'canceling') AND "core_task"."reserved_resources_record" && (ARRAY['/pulp/api/v3/domains/018f5c78-b36b-7b26-9b62-46d8a412c1a3/', '/api/v3/distributions/', 'shared:/api/v3/distributions/'])::text[]) LIMIT 1;`

It is not entirely clear if an index on `state` only or a combined index with `state`, `pulp_created` and `unblocked_at` will perform best.

mdellweg:
On large systems the queries used by the workers to identify ready tasks can get rather slow. Specifically the single query in dispatch to identify wheter an immediate task is ready to be executed is should be sped up drastically.

fixes #5367


pulpcore/app/models/task.py
```diff
@@ -278,7 +278,12 @@ def refresh_from_db(self, using=None, fields=None, **kwargs):
         super().refresh_from_db(using, fields, **kwargs)
 
     class Meta:
-        indexes = [models.Index(fields=["pulp_created"])]
+        indexes = [
+            models.Index(fields=["pulp_created"]),
+            models.Index(fields=["unblocked_at"]),
+            models.Index(fields=["state"]),
+            models.Index(fields=["pulp_created", "unblocked_at", "state"]),
```
ggainey, line 285: Why unblocked_at? The query in the issue relies on created/state/reserved-resource. It's possible that the index with all three won't be used for a query that only relies on created/state. ("Possible" because the postgres optimizer is pretty smart, and what it decides to do is very much based on A Lot Of Variables...)

mdellweg: I removed unblocked_at and put state in the front of the index.
But to me it still really is rather opaque what the best choice is.

ggainey (approved):

ipanova (approved):

pulpcore/app/models/task.py
```diff
@@ -278,7 +278,12 @@ def refresh_from_db(self, using=None, fields=None, **kwargs):
         super().refresh_from_db(using, fields, **kwargs)
 
     class Meta:
-        indexes = [models.Index(fields=["pulp_created"])]
+        indexes = [
+            models.Index(fields=["pulp_created"]),
+            models.Index(fields=["unblocked_at"]),
+            models.Index(fields=["state"]),
+            models.Index(fields=["state", "pulp_created"]),
```
dralley, line 285: I don't think this actually provides much benefit over the already existing single-column indexes, which AFAIK postgresql is able to combine to optimize a query.  It's not 100% as efficient as a multi-column index, but indexes aren't free either.

decko: Also, `pulp_created` and `unblocked_at` could be BRIN indexes, since this data is sequential? They are way cheaper on computation and storage needs.

PR #5379, opened 2024-05-14
Add schedule for task purge

### LINKED ISSUE #5378: Add a scheduled cleanup task to keep the tasks table reasonable

mdellweg:
It is guarded by a setting and off by default.

fixes #5378

### DIFF
--- a/CHANGES/5378.feature
+++ b/CHANGES/5378.feature
@@ -1,0 +1,2 @@
+Added a scheduled version of task purge.
+You need to adjust the "TASK_PROTECTION_TIME" setting in order to use it.
--- a/pulpcore/app/tasks/purge.py
+++ b/pulpcore/app/tasks/purge.py
@@ -1,7 +1,9 @@
 from gettext import gettext as _
 from logging import getLogger
 
+from django.conf import settings
 from django.db.models.deletion import ProtectedError
+from django.utils import timezone
 
 from pulpcore.app.models import (
     ProgressReport,
@@ -9,7 +11,7 @@ from pulpcore.app.models import (
 )
 from pulpcore.app.role_util import get_objects_for_user
 from pulpcore.app.util import get_domain, get_current_authenticated_user
-from pulpcore.constants import TASK_STATES
+from pulpcore.constants import TASK_STATES, TASK_FINAL_STATES
 
 log = getLogger(__name__)
 
@@ -55,7 +57,7 @@ def _details_reporting(current_reports, current_details, totals_pb):
     return current_reports
 
 
-def purge(finished_before, states):
+def purge(finished_before=None, states=None):
     """
     This task purges from the database records of tasks which finished prior to the specified time.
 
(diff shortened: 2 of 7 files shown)


pulpcore/app/tasks/purge.py
```diff
@@ -70,10 +72,15 @@ def purge(finished_before, states):
...
     Args:
-        finished_before (DateTime): Earliest finished-time to **NOT** purge.
-        states (List[str]): List of task-states we want to purge.
+        finished_before (Optional[DateTime]): Earliest finished-time to **NOT** purge.
+        states (Optional[List[str]]): List of task-states we want to purge.
 
     """
+    if finished_before is None:
+        assert settings.TASK_PROTECTION_TIME > 0
+        finished_before = timezone.now() - timezone.timedelta(minutes=settings.TASK_PROTECTION_TIME)
```
ipanova, line 81: I think it will be valuable to configure not only the dispatch interval but also age criteria to purge tasks, e.g
run the clean up task every 2 weeks and delete everything what's a month older. What do you think?

mdellweg: It would be possible, but I think a bit over-engineered. Just thinking about that it may take a couple of paragraphs to explain and add a bunch of cornercases (what if only one of them was specified, which one should be larger).
I'd wish the admin would just activate it and be confident that we don't delete any task younger than specified.
Last of all: It's easy to add this level of control in a subsequent release if needed.

CHANGES/5378.feature
```diff
@@ -0,0 +1,2 @@
+Added a scheduled version of task purge.
+You need to adjust the "TASK_PROTECTION_TIME" setting in order to use it.
```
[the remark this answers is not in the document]
mdellweg: It's in the docs. Hmmm, did I adjust the right docs?

pulpcore/app/tasks/purge.py
ipanova: It definitely introduces some level of complexity but also flexibility. One would want to run the clean up often enough but still keep the record of tasks that are older than then cleanup interval.

dralley (approved):

pulpcore/app/tasks/purge.py
mdellweg: I'm happy to add all these nobs later if requested by actual users (I doubt it will happen).
This is some sort of garbage collection. I'd only care about that it happens before everything starts to smell.
(We have some more plugged out of thin air constants in the tasking system that could have been settings. But they seem to do just fine with the numbers made up.)
So my ask, could you agree on us starting this feature as it is and add flexibility as we discover the need?

ipanova: Yes, that sounds fine to me.

staging_docs/admin/learn/settings.md
```diff
@@ -379,11 +379,11 @@ up before the task finishes. Default is 1440 minutes (24 hours).
 
 
 
-### UPLOAD_PROTECTION_TIME and TMPFILE_PROTECTION_TIME
+### UPLOAD_PROTECTION_TIME, TMPFILE_PROTECTION_TIME and TASK_PROTECTION_TIME
 
-Pulp uses `uploads` and `pulp temporary files` to pass data from the api to worker tasks.
-These options allow to specify a timeinterval in minutes used for cleaning up stale entries. If
-set to 0, automatic cleanup is disabled, which is the default.
+Pulp uses `tasks, `uploads` and `pulp temporary files` to pass data from the api to worker tasks.
```
ipanova, line 384: you're missing a back-tick for the `tasks`

ipanova (approved):

staging_docs/admin/learn/settings.md
mdellweg: Should be fixed.

PR #5393, opened 2024-05-16
Disable docs publishing for this branch.

decko:
[noissue]

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -94,26 +94,3 @@ jobs:
       - name: "Set environment variables"
         run: |
           echo "TEST=${{ matrix.env.TEST }}" >> $GITHUB_ENV
-
-      - name: Download built docs
-        uses: actions/download-artifact@v4
-        with:
-          name: "docs.tar"
-          path: "pulpcore"
-
-      - name: Download Python client docs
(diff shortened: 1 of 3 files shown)


mdellweg (commented):
This looks about right, but we are still building (and maybe failing to) the docs this way.

decko:
> This looks about right, but we are still building (and maybe failing to) the docs this way.

I don't think so.
https://github.com/pulp/pulpcore/pull/5393/files#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7L113-L126

pedro-psb:
It should also be disabled [here](https://github.com/pulp/pulpcore/blob/865e5c9d6b31d594327d256453489c89f5f27fa1/template_config.yml#L109), which should disable building it on the CI as well.

decko:
> It should also be disabled [here](https://github.com/pulp/pulpcore/blob/865e5c9d6b31d594327d256453489c89f5f27fa1/template_config.yml#L109), which should disable building it on the CI as well.
> 
> Edit: or maybe this doesn't make a difference for this branch? I'm not sure

Cool! I can do that.

dralley (approved):

PR #5394, opened 2024-05-16
Updates the CI files.

decko:
[noissue]


.github/workflows/publish.yml
```diff
@@ -91,69 +91,123 @@ jobs:
...
+          name: "core-python-client.tar"
           path: "pulpcore/core-python-client.tar"
           if-no-files-found: "error"
           overwrite: true
 
+      - name: "Merge Python client packages"
+        uses: actions/upload-artifact/merge@v4
+        with:
+          name: python-client.tar
+          pattern: *-python-client.tar
```
dkliban, line 103: Maybe you need to escape the dot?

.github/workflows/publish.yml
```diff
@@ -91,69 +91,123 @@ jobs:
...
           path: "pulpcore/core-python-client.tar"
           if-no-files-found: "error"
           overwrite: true
 
+      - name: "Merge Python client packages"
+        uses: actions/upload-artifact/merge@v4
+        with:
+          name: python-client.tar
+          pattern: "*-python-client.tar"
+
```
mdellweg, line 104: What are we even trying to solve here? I only see a single artifact uploaded. Why would merging it be an improvement?

decko: In v3, multiple uploads from multiple jobs could be done to the same Artifact. This would result in a single archive, which could be useful for sending to upstream systems outside of Actions via API or UI downloads.[^1]

[^1]: Merging multiple artifacts https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#merging-multiple-artifacts

mdellweg: I think, we can:
https://github.com/pulp/plugin_template/pull/866

PR #5396, opened 2024-05-19
Update CI files for branch main

pulpbot:


requirements.txt
```diff
@@ -41,3 +41,4 @@ url-normalize>=1.4.3,<=1.4.3
 uuid6>=2023.5.2,<=2024.1.12
 whitenoise>=5.0,<6.7.0
 yarl>=1.8,<1.9.5
+pip-system-certs
```
mdellweg, line 44: This sounds very fishy.
Do we know what changed here? Is this just the move to centos9? Why can't we just tell aiohttp to use system certs?

lubosmj: Of course, it is fishy. I am experimenting with random stuff now.

However, I am unsure it is feasible to tell aiohttp to use the system-wide certificates. Would we need to add [this code](https://github.com/aio-libs/aiohttp/pull/1849/files) to our tests? I would prefer another way around and something simpler, like setting some env vars (if there are any).

mdellweg: Thanks for experimenting. I hope we can choose the best from several solutions in the end.
But also you raise more questions here:
Is this to get the tests passing?
Do we have an actual gap for users?
What exactly is the situation here?
  * Pulp trying to access the pytest fixture https server?
  * Pulp API needing to accept a client certificate? [Probably not. That is Apache/Nginx's business.]

lubosmj: Yes, this is to get tests passing. After introducing this change, the tests started to fail: https://github.com/pulp/plugin_template/pull/865.

There should not be a gap for real-world usage. It seems like our tests/bindings are not accepting the self-signed certificate we trust inside the container. The bindings are trying to access Pulp endpoints via HTTPs. If we could disable tls/ssl validation in our bindings, this could resolve the problem. Though, at the cost of ignoring other issues that could arise in the future.

mdellweg: Then at the very least this should move to test_requirements.

decko:
Is this the same as #5406 ?

lubosmj:
Yes.

pulp_file/tests/functional/api/test_mime_types.py
```diff
@@ -54,7 +59,7 @@ def test_content_types(
         async def get_content_type():
             async with aiohttp.ClientSession() as session:
                 url = urljoin(distribution.base_url, content_unit.relative_path)
-                async with session.get(url) as response:
+                async with session.get(url, ssl=ssl_context) as response:
                     return response.headers.get("Content-Type")
 
         content_type = asyncio.run(get_content_type())
```
mdellweg, line 65: Can you just use `requests.get` here?

mdellweg:
There are a few more uses of aiohttp in the tests where requests would be sooo much simpler. But i think we can use this as is for now to unblock everything.

mdellweg (dismissed):

pulpcore/tests/functional/api/test_replication.py
```diff
@@ -1,12 +1,19 @@
...
 from pulpcore.client.pulpcore import ApiException
 from pulpcore.client.pulpcore import AsyncOperationResponse
 
 from pulpcore.tests.functional.utils import PulpTaskGroupError
 
 
+@pytest.fixture
+def ca_cert():
+    with open(certifi.where()) as f:
+        return f.read()
```
lubosmj, line 14: Waiting for https://github.com/pulp/pulp-cli/pull/971 to get released to see if we can ditch this addition.

mdellweg (approved):

ipanova:
That's not it, I'm afraid, pulp_deb fails because it uses fixture from pulpcore download_content_unit so we need to adjust it as well https://github.com/pulp/pulp_deb/actions/runs/9170618410/job/25213240062?pr=1086#step:14:2219

pulpcore/tests/functional/api/test_replication.py
mdellweg: It looks like pulp-glue was working properly with respect to this after all.

lubosmj:
There is just one last issue remaining that I cannot quickly fix:
```
usr/local/lib/python3.9/site-packages/pulpcore/tests/functional/api/using_plugin/test_contentguard.py:66: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

authorized_users = [<pulpcore.tests.functional.gen_user.<locals>.user_context object at 0x7f78f1149370>, <pulpcore.tests.functional.pulp_admin_user.<locals>.AdminUser object at 0x7f78f1149220>]

    def _assert_access(authorized_users):
        """Asserts that only authorized users have access to the distribution's base_url."""
        for user in all_users:
            if user is not anonymous_user:
                auth = (user.username, user.password)
            else:
                auth = None
            response = get_from_url(distro.base_url, auth=auth)
            expected_status = 404 if user in authorized_users else 403
>           assert response.status_code == expected_status, f"Failed on {user.username=}"
E           AttributeError: 'AnonymousUser' object has no attribute 'username'

usr/local/lib/python3.9/site-packages/pulpcore/tests/functional/api/using_plugin/test_contentguard.py:50: AttributeError
```

Not sure why the assertion fails in the first place...
------------------------------------------------------------------------

Nevertheless, we should either manually configure ssl_context for all aiohttp sessions (removed commit) or revert https://github.com/pulp/plugin_template/pull/865 to unblock the releasing process.

mdellweg:
> **Nevertheless, we should either manually configure ssl_context for all aiohttp sessions ([removed commit](https://github.com/pulp/pulpcore/compare/fcad93c22a50067b15b882a22148e9972176a60c..05ffa72dad57378eb9eb366b0c73000bed756b03)) or revert [pulp/plugin_template#865](https://github.com/pulp/plugin_template/pull/865) to unblock the releasing process. Converting the fixtures to use `requests` instead of `aiohttp` is not feasible at the moment because it might break plugins' tests too.**

Thanks for all the effort to rectify this situation without setting back pulp_container.
But as you say, we cannot just change the semantics of test fixtures/utilities used by plugins.
Also as i see it somehow the added "trust anchor" command actually regresses on all the ssl connection that worked before in the tests. I vote for reverting that change until we know how to do it in a non-destructive way.

lubosmj:
The referenced plugin_template change was introduced because of the failing pulp-container's CI. We can try to temporarily make the changes on the pulp-container's repository exclusively. The issue is that gnutls does not trust certificates copied to the pulp-ci-centos9 container. Thus, we forced the execution of `trust anchor` to add the certificates to the trusted certs list. However, this caused many more issues with the existing aiohttp fixtures.

decko:
> There is just one last issue remaining that I cannot quickly fix:
> 
> ```
> usr/local/lib/python3.9/site-packages/pulpcore/tests/functional/api/using_plugin/test_contentguard.py:66: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> 
> authorized_users = [<pulpcore.tests.functional.gen_user.<locals>.user_context object at 0x7f78f1149370>, <pulpcore.tests.functional.pulp_admin_user.<locals>.AdminUser object at 0x7f78f1149220>]
> 
>     def _assert_access(authorized_users):
>         """Asserts that only authorized users have access to the distribution's base_url."""
>         for user in all_users:
>             if user is not anonymous_user:
>                 auth = (user.username, user.password)
>             else:
>                 auth = None
>             response = get_from_url(distro.base_url, auth=auth)
>             expected_status = 404 if user in authorized_users else 403
> >           assert response.status_code == expected_status, f"Failed on {user.username=}"
> E           AttributeError: 'AnonymousUser' object has no attribute 'username'
> 
> usr/local/lib/python3.9/site-packages/pulpcore/tests/functional/api/using_plugin/test_contentguard.py:50: AttributeError
> ```
> 
> Not sure why the assertion fails in the first place...
> 

What if you set `auth = anonymous_user` on line 47?

lubosmj:
Revert available here: https://github.com/pulp/plugin_template/pull/869

PR #5419, opened 2024-05-23
5417 remove fixtures

mdellweg:


lubosmj (approved):
I do not see any issues with these changes. It is mostly about replacing old fixtures with newer ones and moving everything fixture-related to `pytest_plugin.py`. Plugin writers will need to replace their old fixtures too. An inevitable change.

ggainey (approved):

PR #5424, opened 2024-05-29
Remove get_url from task locking

gerrod3:
fixes: #5148


lubosmj (commented):
Are we going to change this part as well https://github.com/pulp/pulpcore/blob/778108af02e75a44d53a66749529e1b0acd751c7/pulpcore/app/replica.py#L59?

CHANGES/5148.removal
```diff
@@ -0,0 +1 @@
+Removed ``get_url`` from task resource locking.
```
daviddavis, line 1: As a user, I have no idea what this means. Any chance you can spell out how the API will change (if it all)? Like it sounds like the `reserved_resources_record` field is changing on the `/pulp/api/v3/tasks/` is changing? Anything else?

gerrod3: @daviddavis I rewrote the changelog, can you check it out?

daviddavis: @gerrod3 looks great. Thanks!

mdellweg (approved):

lubosmj (approved):

PR #5438, opened 2024-06-03
Allow to specify the type of remote authentication

### LINKED ISSUE #5437: Remote authentication must provide a way to be represented in openapi spec
E.g. for cert based auth there should be an auth scheme like:
```json
{"type": "mutualTLS"}
```

mdellweg:
When configuring remote authentication (by the reverse proxy), one should be able to augment the opena pi security specification accordingly.

fixes #5437


CHANGES/5437.feature
```diff
@@ -0,0 +1,2 @@
+Add ability to configure the openapi schema for remote user authentication via `REMOTE_USER_OPENAPI_SECURITY_SCHEME`.
+It defaults to "mutualTLS" for cert based authentication.
```
lubosmj, line 2:
```suggestion
Its type defaults to "mutualTLS" for cert based authentication.
```

Do we want to reference other defaults from https://swagger.io/specification/?sbsearch=mutualtls#security-scheme-object?

requirements.txt
```diff
@@ -15,7 +15,7 @@ djangorestframework>=3.14.0,<=3.15.1
 djangorestframework-queryfields>=1.0,<=1.1.0
 drf-access-policy>=1.1.2,<1.5.1
 drf-nested-routers>=0.93.4,<=0.94.1
-drf-spectacular==0.26.5  # We monkeypatch this so we need a very narrow requirement string
+drf-spectacular==0.27.2  # We monkeypatch this so we need a very narrow requirement string
```
ggainey, line 18: Does this PR need 0.26.5 specifically? or is this an incidental change? If the latter - including in this PR makes the PR way harder to backport.

mdellweg: It's worse than that. Seems like we need openapi 3.1 instead of 3.0.3 and now we need to coordinate a lot of other things including the openapi-generator.
But on the one hand we'll probably not backport this feature, and otoh i split this version bump into another PR already.

gerrod3 (approved):

mdellweg:
Ohh, This depended on the other change we reverted. So I'm going to revert this one too.

PR #5441, opened 2024-06-06
Add a lock to avoid multiple workers to emit at the same time.

### LINKED ISSUE #5442: domain disk usage metric runs on every API worker
**Version**
3.54.0

**Describe the bug**
When telemetry collection is enabled for API workers, the database is inundated with queries to count all artifacts in every domain. This ends using a lot of CPU on the database. Each API worker is running the queries for each domain every 60 seconds. 

**To Reproduce**
Start pulpcore-api with the following command

`/usr/local/bin/opentelemetry-instrument --service_name pulp-api --exporter_otlp_endpoint http://pulp-otel-collector-svc:4318 --exporter_otlp_insecure true pulpcore-api --timeout 90 --workers 20 --access-logfile -`

Watch the database CPU utilization spike.

decko:
Closes #5442


ipanova:
pleases add `closes #issue`

decko:
> pleases add `closes #issue`

Thanks @ipanova. The PR came before the issue in this case.

ipanova:
@decko you're just extremely fast ;-P

ipanova:
...and changelog ;)

pulpcore/app/util.py
```diff
@@ -539,19 +542,20 @@ def _init_emitting_total_size(self):
...
-                        {"pulp_href": get_url(self.domain), "domain_name": self.domain.name},
-                    )
-                ]
+            with contextlib.suppress(AdvisoryLockError), PGAdvisoryLock(STORAGE_METRICS_LOCK):
+                from pulpcore.app.models import Artifact
+
+                options = yield  # noqa
+
+                while True:
+                    artifacts = Artifact.objects.filter(pulp_domain=self.domain).distinct()
```
ipanova, line 551: any reason we are calling here distinct if artifact is going to be unique by its' sha256?

decko: @lubosmj is the best one to answer that.

dkliban: i agree that it is not necessary.

dkliban: but let's make this change outside this PR.

ipanova: Sounds good, just caught my eye. Let's address in a separate PR.

pulpcore/app/util.py
```diff
@@ -574,3 +578,31 @@ def init_domain_metrics_exporter():
 
     for domain in Domain.objects.all():
         DomainMetricsEmitterBuilder.build(domain)
+
+
+class PGAdvisoryLock:
```
bmbouter, line 583: I'm not looking too closely at this, but shouldn't this just use the existing one in `util.py`?

dkliban: It's being moved to avoid a circular import.

decko: importing it from the `_util.py` file caused a circular import. I'm moving this class to `util.py` hopping that it solves the issue.

decko: I'm trying it now @bmbouter

decko: Sorry man. I wasn't able to solve the circular import @bmbouter

unittest_requirements.txt
```diff
@@ -4,4 +4,4 @@ aiotools
 pytest<8
 pytest-django
 pytest-asyncio
-pytest-redis
+pytest-redis==3.0.2
```
bmbouter, line 7: This is coming as https://github.com/pulp/pulpcore/pull/5443

dkliban (approved):

bmbouter (approved):
This looks good to me.

pulpcore/app/util.py
```diff
@@ -539,19 +542,26 @@ def _init_emitting_total_size(self):
...
-                artifacts = Artifact.objects.filter(pulp_domain=self.domain).distinct()
-                total_size = artifacts.aggregate(size=Sum("size", default=0))["size"]
-                options = yield [  # noqa
-                    metrics.Observation(
-                        total_size,
-                        {"pulp_href": get_url(self.domain), "domain_name": self.domain.name},
-                    )
-                ]
+            try:
+                with PGAdvisoryLock(STORAGE_METRICS_LOCK):
```
gerrod3, line 546: Bit late to the party, but are we sure this actually works? Putting the `with` outside the `while` loop means that the first api-worker will grab this lock for the first domain in the system, but the rest of the domains will get locked out; see `init_domain_metrics_exporter`: https://github.com/pulp/pulpcore/blob/02bb455315a17a39cb0d936a12a13fbbdb887bb3/pulpcore/app/util.py#L582-L586 If I'm right then I think the lock should be a combination of the `STORAGE_METRICS_LOCK` and the `pulp_domain.pulp_id`.

Second, what happens when the api-worker who obtains the lock dies? Does this mean the metrics will no longer be sent until a new worker is spun up? 

Also, do we need to be worry about the metrics also being sent from a new worker after domain creation? https://github.com/pulp/pulpcore/blob/02bb455315a17a39cb0d936a12a13fbbdb887bb3/pulpcore/app/models/domain.py#L77-L81 It seems like this will run into the same issue with locks I described above.

dkliban:
@gerrod3 Are you saying that each domain has it's own gauge and now all of these gauges are sharing a lock and as a result only one domain will be reported on every 60 seconds?

gerrod3:
> @gerrod3 Are you saying that each domain has it's own gauge and now all of these gauges are sharing a lock and as a result only one domain will be reported on every 60 seconds?

Yes, that's how I believe the code works from reading it, but I could be wrong. I'm just using a surface reading, I haven't dug deep to figure out how metrics fully works.

pulpcore/app/util.py
lubosmj: The answer to the last question:

`init_domain_metrics_exporter` is called once, when we start up a Pulp instance. Then, when a new domain is created, we run `_report_domain_disk_usage`. There should not be any interference between these two. The first one considers already existing domains (after booting, after restarting services, etc.). The latter deals only with domains that were created during the runtime. Both pieces are executed on api-app.

pulpcore/app/util.py
lubosmj: Correct, we may want to remove that.

PR #5453, opened 2024-06-07
Changes the way we lock the domain metric emitter.

decko:
[noissue]


pulpcore/app/util.py
```diff
@@ -532,6 +532,7 @@ def __init__(self, domain):
             self.domain = domain
             self.meter = metrics.get_meter(f"domain.{domain.name}.meter")
             self.instrument = self._init_emitting_total_size()
+            self.DOMAIN_LOCK = int.from_bytes(self.domain.encode('utf-8'))
```
dkliban, line 535: do you mean self.domain.name?

lubosmj: `name` is string. Can this overflow?

```
pg_try_advisory_lock(key bigint)	boolean	Obtain exclusive session level advisory lock if available
pg_try_advisory_lock(key1 int, key2 int)	boolean	Obtain exclusive session level advisory lock if available <---- I think we are using this one: https://github.com/pulp/pulpcore/blob/8efd2b48a2481c5a8d0edbdf4b3458d54c05e065/pulpcore/app/util.py#L604C36-L604C56
```

```
integer	4 bytes	typical choice for integer	-2147483648 to +2147483647
bigint	8 bytes	large-range integer	-9223372036854775808 to +9223372036854775807
```

decko: Thanks @dkliban

decko: > `name` is string. Can this overflow? Yes, do we care? Maybe not.
> 
> ```
> pg_try_advisory_lock(key bigint)	boolean	Obtain exclusive session level advisory lock if available
> pg_try_advisory_lock(key1 int, key2 int)	boolean	Obtain exclusive session level advisory lock if available <---- I think we are using this one: https://github.com/pulp/pulpcore/blob/8efd2b48a2481c5a8d0edbdf4b3458d54c05e065/pulpcore/app/util.py#L604C36-L604C56
> ```
> 
> ```
> integer	4 bytes	typical choice for integer	-2147483648 to +2147483647
> bigint	8 bytes	large-range integer	-9223372036854775808 to +9223372036854775807
> ```

Yes, we need to convert the name into `bytes` and then into `int`.

pulpcore/app/util.py
```diff
@@ -572,7 +573,7 @@ def __getattr__(self, *args, **kwargs):
 
     @classmethod
     def build(cls, domain):
-        otel_enabled = os.getenv("PULP_OTEL_ENABLED")
+        otel_enabled = os.getenv("PULP_OTEL_ENABLED").lower()
         if otel_enabled == "true" and settings.DOMAIN_ENABLED:
```
dkliban, line 577: @bmbouter was suggesting that the the acquiring of the advisory lock happen here. so emitters are only created if a lock could be acquired.

decko: Yep, could be a thing.

pulpcore/app/util.py
lubosmj: I understand what is going on here. I am a bit worried that this could trigger some undefined behaviour. What could possibly happen if an overflow occurs because of the postgres method accepting bounded integers? I think we are going to use the same lock for different domains in this case. Is this worrying or not?
```
>>> import ctypes
>>> from sys import byteorder
>>> 
>>> a = int.from_bytes("dev_env_aa_0001".encode("utf-8"), byteorder)
>>> b = int.from_bytes("dev_env_bb_0001".encode("utf-8"), byteorder)
>>> a
255399919651570808716321716629300580
>>> b
255399919651575549529548659984065892
>>> ctypes.c_int(a)
c_int(1601594724)
>>> ctypes.c_int(b)
c_int(1601594724)
>>> 
>>> ctypes.c_int(2147483647)
c_int(2147483647)
>>> ctypes.c_int(2147483648)
c_int(-2147483648)
>>> ctypes.c_int(2147483649)
c_int(-2147483647)
```

bmbouter: Yeah I think it's a little concerning. Also there could be collisions with the tasking system locks right? I'm unsure. The keyspace of these locks probably needs to be a little closer managed application-wide. For me though, doing that is a big enough issue that we should consider tackling it as a separate piece of work (to me) and this ok for now.

pulpcore/app/util.py
bmbouter: Yeah, move the lock acquisitions here and then instead of racing for X processes racing every Y seconds for N domains, there is exactly 1 race (a single lock) during each process booting to elect the "reporter" and that reporter creates the guages and just reports (no more lock acquisitions).

pulpcore/app/util.py
decko: > Yeah I think it's a little concerning. Also there could be collisions with the tasking system locks right? I'm unsure. The keyspace of these locks probably needs to be a little closer managed application-wide. For me though, doing that is a big enough issue that we should consider tackling it as a separate piece of work (to me) and this ok for now.

Nope. Tasking locks are on a different group. We're using the PGAdvisoryLock context manager which sets the locks to a different group to not collide with task locks. We can even create a new group just for this type of lock.

decko: I think we hit the issue @lubosmj is describing. https://github.com/pulp/pulpcore/actions/runs/9453443830/job/26038918671#step:10:2811

Right now the code is generating a bigint where the advisory_lock accepts only a normal integer.

lubosmj: Aha, even worse. I thought that it will accept it, ignoring the overflow, but it fails on the input.

I think we will not be able to use name (255+B) or uuid (128b) as a base for a 32b integer. Computing a hash or generating a random value is not ideal too.

What if we take a truncated timestamp of a domain (created_at)? It should fit the scenario. 🤔

pulpcore/app/util.py
bmbouter: I might be totally off here...

My thought was to not acquire the locks per-domain, but instead use the lock to represent the leader election. The leader then creates all the guages. I think my design pitch relies heavily on the assumption of a reporting process only having-and-report-with-guages as long as it's running and still holding the advisory lock. If the reporting process continued to report with guages and let go of the lock this won't work well.

mdellweg: Why is this so different from task queue reporting?

PR #5454, opened 2024-06-07
Refactor openapi

mdellweg:


pulpcore/openapi/hooks.py
```diff
@@ -0,0 +1,33 @@
...
+    # Add domain-settings value
+    result["info"]["x-pulp-domain-enabled"] = settings.DOMAIN_ENABLED
+
+    # Add x-isDomain flag to domain path parameters
+    # Save a copy of the bad operationId in case we generate bindings
+    for path, path_spec in result["paths"].items():
+        for operation, operation_spec in path_spec.items():
+            if request.bindings:
+                # Keep the operation id before sanitization
+                operation_spec["x-copy-operationId"] = operation_spec["operationId"]
```
gerrod3, line 22: Why does the operation ids get sanitized when generating the bindings? Also, where does it get sanitized? I thought this was a hook, does it occur after this hook?

mdellweg: I asked it to be turned into a hook. But it was declined upstream.
Sadly the sanitization happens after applying the post process hooks. But still I prefer to call the super function and act around it over copying it and modifying and no one can understand what's copied and outdated and what actually is our addition here.

pulpcore/openapi/__init__.py
```diff
@@ -484,32 +417,18 @@ def get_schema(self, request=None, public=False):
...
-        # Adding current host as server (it will provide a default value for the bindings)
-        server_url = "http://localhost:24817" if not request else request.build_absolute_uri("/")
-        result["servers"] = [{"url": server_url}]
-
-        return normalize_result_object(result)
+        if request.bindings:
+            # Undo the spectacular sanitization of operation ids
+            for path, path_spec in result["paths"].items():
+                for operation, operation_spec in path_spec.items():
+                    operation_spec["operationId"] = operation_spec.pop("x-copy-operationId")
```
mdellweg, line 430: I was wondering if we should only add the x-operationName and use that flag in the bindings templates instead of the operationId. But without versioned openapi templates this is sooo hard to make the move.

pulpcore/openapi/hooks.py
mdellweg: https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/generators.py#L293

gerrod3 (approved):

PR #5465, opened 2024-06-10
Replica - stop assuming every plugin supports publications

### LINKED ISSUE #5464: Replication assumes every plugin supports Publications
The ```Replicator``` class, which is being subclassed inside plugins to support the replication feature, assumes that every plugin supports publications as it tries to access ```upstream_distribution["publication"]``` inside one of its methods (for some plugins, the dictionary simply doesn't contain the "publication" key so an exception gets raised). This forces certain subclasses of ```Replicator``` to create workarounds or rewrite the given method.

I propose making the method more general, removing such assumptions.

Relevant code: https://github.com/pulp/pulpcore/blob/c31a39c186fd734e9e0409321735d7d6509a12ee/pulpcore/app/replica.py#L97C9-L97C96

MichalPysik:
Pulpcore no longer assumes that every plugin which implements support for the Replication feature supports publications (e.g. container doesn't).

closes #5464

### DIFF
--- a/CHANGES/5464.bugfix
+++ b/CHANGES/5464.bugfix
@@ -1,0 +1,2 @@
+Pulpcore no longer assumes that every plugin implementing the Replication feature supports
+Publications.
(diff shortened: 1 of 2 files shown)


lubosmj (approved):
Thanks! This is required for pulp_container workflows!

dkliban (approved):

ipanova (approved):

PR #5466, opened 2024-06-11
EXPERIMENT

mdellweg:


requirements.txt
```diff
@@ -41,3 +41,6 @@ url-normalize>=1.4.3,<=1.4.3
 uuid6>=2023.5.2,<=2024.1.12
 whitenoise>=5.0,<6.7.0
 yarl>=1.8,<1.9.5
+
+# experiment
+requests!=2.32.3,<2.32
```
ggainey, line 46:
```suggestion
requests!=2.32.3,<2.33
```

ggainey (changes requested):

requirements.txt
ggainey: nm, missed that this was a dliberate experiment in the second force-push.

ggainey:
So what this teaches us is, we don't have to (and prob should not) use https: to talk to backend-storage, *and* (more importantly) it's the requests=2.32 changes in the SSLContext arena that are breaking us.

mdellweg:
Closing. There's really not much that can be done from here.

PR #5471, opened 2024-06-12
Disable the Storage Metrics emmiter for now.

### LINKED ISSUE #5468: Remove the storage reporting for domains
We recently had a showstopping SQL CPU utilization issue with the domain reporting. This was reported [here](https://github.com/pulp/pulpcore/issues/5442) and an attempt to fix it was [made here](https://github.com/pulp/pulpcore/pull/5441). Then another attempt was [made here](https://github.com/pulp/pulpcore/pull/5453).

We can't re-enable the OTEL metrics in our environment until this is resolved. Also we don't have capacity to resolve this in a timely manner.

So given ^ perspectives, I am requesting the feature be removed until we can fix it adequately and re-enable it.

decko:
Closes #5468

### DIFF
--- a/CHANGES/5468.misc
+++ b/CHANGES/5468.misc
@@ -1,0 +1,1 @@
+Temporarily disable domains' disk usage reporting.
--- a/pulpcore/app/models/domain.py
+++ b/pulpcore/app/models/domain.py
@@ -2,7 +2,7 @@ from opentelemetry.metrics import Observation
 
 from django.core.files.storage import default_storage
 from django.db import models
-from django_lifecycle import hook, BEFORE_DELETE, BEFORE_UPDATE, AFTER_CREATE
+from django_lifecycle import hook, BEFORE_DELETE, BEFORE_UPDATE
 
 from pulpcore.app.models import BaseModel, AutoAddObjPermsMixin
 from pulpcore.exceptions import DomainProtectedError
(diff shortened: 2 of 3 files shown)


lubosmj (commented):
To be on the safe side, we will need to disable https://github.com/pulp/pulpcore/blob/9e9198ca115d14229c85a4db4da1d811846e7462/pulpcore/app/models/domain.py#L78 too. This is a hook for newly created domains.

It is so unfortunate that we could not validate this patch: https://github.com/pulp/pulpcore/pull/5452.

decko:
> It is so unfortunate that we could not validate this patch: #5452.

I think we should accept and set it as validated. But yet it won't solve our issue.

CHANGES/5468.misc
```diff
@@ -0,0 +1 @@
+Disable the Storage Metrics Emitter until we figure its resource usage.
```
lubosmj, line 1:
```suggestion
Temporarily disabled domains' disk usage reporting.
```

lubosmj (approved):
Left a non-blocking comment.

lubosmj (approved):

dkliban (approved):

CHANGES/5468.misc
```diff
@@ -0,0 +1 @@
+Temporarily disable domains' disk usage reporting.
```
daviddavis, line 1: Is misc the right type of entry?

decko: I believe so @daviddavis. We didn't fixed it. We're just disabling and will get back later to it.

Issue #5497, opened 2024-06-20
Add RBAC for Task Groups


gerrod3:
Currently Tasks has RBAC, which assigns the owner of the task to the user who dispatched it. TaskGroups does not have RBAC which means if a user dispatches a task that uses groups they will have to manually track each task themselves instead of being able to use the TaskGroup api. The RBAC for the TaskGroup should scope based on view permissions for the group's parent task.

Issue #4125, opened 2023-07-25
Add ability to cancel all the tasks in a task group


dkliban:
A task group should accept a PATCH request with status=canceled and cancel all the tasks.

PR #5499, opened 2024-06-20
Add RBAC for TaskGroups

gerrod3:
fixes: #5497

This should allow adding https://github.com/pulp/pulpcore/issues/4125 to make more sense.

### DIFF
--- a/pulpcore/app/tasks/test.py
+++ b/pulpcore/app/tasks/test.py
@@ -1,4 +1,6 @@
 import backoff
+from pulpcore.app.models import TaskGroup
+from pulpcore.tasking.tasks import dispatch
 
 
 def dummy_task():
@@ -18,3 +20,13 @@ def gooey_task(interval):
     from time import sleep
 
     sleep(interval)
+
+
+def dummy_group_task(inbetween=3, intervals=None):
+    """A group task that dispatches 'interval' sleep tasks every 'inbetween' seconds.'"""
+    intervals = intervals or range(5)
+    task_group = TaskGroup.current()
+    for interval in intervals:
+        dispatch(sleep, args=(interval,), task_group=task_group)
+        sleep(inbetween)
+    task_group.finish()
--- a/pulpcore/pytest_plugin.py
+++ b/pulpcore/pytest_plugin.py
@@ -1015,19 +1015,21 @@ def wget_recursive_download_on_host():
 
 @pytest.fixture(scope="session")
 def dispatch_task(pulpcore_bindings):
-    def _dispatch_task(*args, **kwargs):
+    def _dispatch_task(*args, task_group_id=None, **kwargs):
         cid = pulpcore_bindings.client.default_headers.get("Correlation-ID") or str(uuid.uuid4())
         username = pulpcore_bindings.client.configuration.username
         commands = (
             "from django_guid import set_guid; "
             "from pulpcore.tasking.tasks import dispatch; "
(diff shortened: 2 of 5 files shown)


pulpcore/app/tasks/test.py
```diff
@@ -18,3 +20,13 @@ def gooey_task(interval):
     from time import sleep
 
     sleep(interval)
+
+
+def group_task(inbetween=3, intervals=None):
```
mdellweg, line 25: This is a userfacing name (in theory, because users should never see that task). Let's make it really obvious it's not harmful.
```suggestion
def dummy_group_task(inbetween=3, intervals=None):
```

pulpcore/pytest_plugin.py
```diff
@@ -1015,19 +1015,21 @@ def _wget_recursive_download_on_host(url, destination):
...
             "from pulpcore.tasking.tasks import dispatch; "
+            "from pulpcore.app.models import TaskGroup; "
             "from pulpcore.app.util import get_url, set_current_user; "
             "from django.contrib.auth import get_user_model; "
             "User = get_user_model(); "
             f"user = User.objects.filter(username='{username}').first(); "
             "set_current_user(user); "
             f"set_guid({cid!r}); "
-            f"task = dispatch(*{args!r}, **{kwargs!r}); "
+            f"task_group = TaskGroup.objects.filter(pk={task_group_id!r}).first(); "
```
mdellweg, line 1031:
```suggestion
            f"task_group = task_group_id and TaskGroup.objects.filter(pk={task_group_id!r}).first(); "
```

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -367,3 +367,34 @@ def test_emmiting_unblocked_task_telemetry(
         pulpcore_bindings.TasksApi.tasks_cancel(task_href, {"state": "canceled"})
         for task_href in resident_task_hrefs
     ]
+
+
+@pytest.fixture
+def task_group(dispatch_task_group, monitor_task_group):
+    """Fixture containing a finished Task Group."""
+    kwargs = {"inbetween": 0, "intervals": [0]}
+    tgroup_href = dispatch_task_group("pulpcore.app.tasks.test.group_task", kwargs=kwargs)
```
mdellweg, line 376: We have made no assertions about that (dummy) group task to dispatch any more tasks.
I think we should either say that we increase the test coverage over task group operations, or just dispatch the other dummy sleep task here.

gerrod3: What do you mean by this? Should I add a check in `monitor_task_group` to see if `all_tasks_dispatched=True`?

mdellweg: "all_tasks_dispatched" is a dead nut... You cannot rely on it anyway.
But for all the test assertions I see, it would be sufficient to dispatch a (bunch of) dummy_task(s) as part of a task_group.
So what did you create the group_task for? Do we really need it?

gerrod3: I see what you mean. For most (all of them?) of our group tasks they are instantiated with one task as the parent task that will dispatch further children, but your comment made me realize this isn't a requirement for using TaskGroups. We don't really need it, it's just a mirror of how I thought TaskGroups worked.

lubosmj (approved):

ggainey (approved):

Issue #5501, opened 2024-06-21
A content's artifact can not be uploaded if the content already exists on-demand in Pulp


gerrod3:
**Version**
Since the `retrieve` api was introduced, so probably all our supported branches.

**Describe the bug**
If Pulp has an on-demand content unit, a user can not upload that content's artifact to Pulp and turn it into an immediate unit. The problem with this is that we know if the backing remote for the on-demand content is deleted then that content becomes undownloadable until the remote is recreated and synced. #1975 

**To Reproduce**
Sync an on-demand repo. Download a content unit from the upstream. Try to upload that unit to Pulp. Notice that upload succeeds, but the return content unit is still on-demand, its `artifact` field is still `None`. 

**Expected behavior**
The content should become immediate after upload.

PR #5502, opened 2024-06-21
Fix upload for existing on-demand content

gerrod3:
fixes: #5501

### DIFF
--- a/pulpcore/app/serializers/content.py
+++ b/pulpcore/app/serializers/content.py
@@ -60,6 +60,7 @@ class NoArtifactContentSerializer(base.ModelSerializer):
         artifacts = self.get_artifacts(validated_data)
 
         content = self.retrieve(validated_data)
+        created = False
         if content is not None:
             content.touch()
         else:
@@ -74,6 +75,22 @@ class NoArtifactContentSerializer(base.ModelSerializer):
                 content = self.retrieve(validated_data)
                 if content is None:
                     raise
+            else:
+                created = True
+        # Ensure the content now has the uploaded artifact(s)
+        if not created:
+            ca_set = content.contentartifact_set
+            if ca_set.count() == 0 or ca_set.filter(artifact=None).exists():
+                cas = [
+                    models.ContentArtifact(artifact=a, content=content, relative_path=rp)
+                    for rp, a in artifacts.items()
+                ]
+                models.ContentArtifact.objects.bulk_create(
+                    cas,
+                    update_conflicts=True,
+                    update_fields=["artifact"],
+                    unique_fields=["content", "relative_path"],
+                )
 
         if repository:
             repository.cast()
--- a/pulpcore/tests/functional/api/using_plugin/test_content_access.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_content_access.py
@@ -40,3 +40,32 @@ def test_file_remote_on_demand(
     distribution = file_distribution_factory(repository=repo.pulp_href)
     # attempt to download_file() a file
     download_file(f"{distribution.base_url}/1.iso")
+
+
+@pytest.mark.parallel
+def test_upload_file_on_demand_already(
+    basic_manifest_path,
+    file_fixtures_root,
+    file_remote_factory,
+    file_repo,
+    file_bindings,
+    monitor_task,
+):
+    """Test that on-demand content can be uploaded to Pulp and become immediate content."""
+    remote = file_remote_factory(basic_manifest_path, policy="on_demand")
+    body = {"remote": remote.pulp_href}
+    monitor_task(file_bindings.RepositoriesFileApi.sync(file_repo.pulp_href, body).task)
+    file_repo = file_bindings.RepositoriesFileApi.read(file_repo.pulp_href)
+    result = file_bindings.ContentFilesApi.list(repository_version=file_repo.latest_version_href)
+    assert result.count == 3
+    for content in result.results:
+        assert content.artifact is None
+
+    file_content = file_fixtures_root / "basic" / content.relative_path
+    body = {"relative_path": content.relative_path, "file": file_content}
+    task = monitor_task(file_bindings.ContentFilesApi.create(**body).task)
+    assert len(task.created_resources) == 1
+    assert task.created_resources[0] == content.pulp_href
+
+    content = file_bindings.ContentFilesApi.read(content.pulp_href)
+    assert content.artifact is not None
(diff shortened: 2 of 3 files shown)


pulpcore/app/serializers/content.py
```diff
@@ -64,16 +64,24 @@ def create(self, validated_data):
             content.touch()
         else:
             try:
-                with transaction.atomic():
-                    content = self.Meta.model.objects.create(**validated_data)
-                    for relative_path, artifact in artifacts.items():
-                        models.ContentArtifact.objects.create(
-                            artifact=artifact, content=content, relative_path=relative_path
-                        )
+                content = self.Meta.model.objects.create(**validated_data)
```
mdellweg, line 67: I think, this was actually correct. If we had created the content, we need to create the content_artifacts also. And we need to do so in a transaction for the content to never be incomplete.

pulpcore/app/serializers/content.py
```diff
@@ -64,16 +64,24 @@ def create(self, validated_data):
...
+            cas = [
+                models.ContentArtifact(artifact=a, content=content, relative_path=rp)
+                for rp, a in artifacts.items()
+            ]
+            models.ContentArtifact.objects.bulk_create(
+                cas,
+                update_conflicts=True,
+                update_fields=["artifact"],
+                unique_fields=["content", "relative_path"],
+            )
```
mdellweg, line 84: What if you moved this block into the except branch? That should be the only codepath where the content could have unresolved remote artifacts.

mdellweg (changes requested):
This issue gets me thinking:
What do we do, if someone uploads a file into another content that magically happend to match this content_artifact?
Don't we need to check that the artifacts we attach here match the expectations of the remote_artifacts for the content_artifact, like size and checksums?

pulpcore/app/serializers/content.py
gerrod3: No, that's wrong. The except part will only be reached if the result of the first `retrieve` is `None`.

mdellweg:
> > This issue gets me thinking: What do we do, if someone uploads a file into another content that magically happend to match this content_artifact? Don't we need to check that the artifacts we attach here match the expectations of the remote_artifacts for the content_artifact, like size and checksums?
> 
> I don't understand what you are trying to say. `ContentArtifact` is unique on `content` and `relative_path`. When you call `retrieve` you have the **validated** data of the artifact the user uploaded and their supplied metadata that the plugin's create api requires. This is meant so that you are guaranteed to pull the correct content unit if it already exists in Pulp by supplying the correct bits. This is only wrong if your implementation of `validate` is wrong and doesn't check that the user is supplying what they say they are supplying.

OK, I think I'm saying two loosely related things:
Let's ignore the first (upload of an artifact for a remote artifact of another content unit not know to the user).
But the second one is: If there is on demand content, it has content artifacts and hopefully remote artifacts attached. And as far as i know, these can contain checksums the actual artifact should meet. Now the user uploads artifacts that presumably create that content and we take it for that what if the checksums mismatch?
The thing is with on-demand content we kind of rely on the upstream to never lie to us. I start to think this is a really bad idea. Maybe remote content and content should not be mixed that way.

gerrod3:
> OK, I think I'm saying two loosely related things: Let's ignore the first (upload of an artifact for a remote artifact of another content unit not know to the user). But the second one is: If there is on demand content, it has content artifacts and hopefully remote artifacts attached. And as far as i know, these can contain checksums the actual artifact should meet. Now the user uploads artifacts that presumably create that content and we take it for that what if the checksums mismatch? The thing is with on-demand content we kind of rely on the upstream to never lie to us. I start to think this is a really bad idea. Maybe remote content and content should not be mixed that way.

Yes, that is correct. Your content unit if it supports on-demand syncing should understand how to differentiate between content units from its metadata, either from the upstream or from an uploaded artifact. If the upstream lies about it's metadata then on upload you should create a different content unit, not re-use the bad content. This is a contract we expect our plugins to follow (which is why pulp_ansible is a mess since it doesn't follow it), so I am still not seeing your problem.

pulpcore/app/serializers/content.py
dralley: Agree with this

pulpcore/tests/functional/api/using_plugin/test_content_access.py
```diff
@@ -40,3 +40,32 @@ def test_file_remote_on_demand(
...
+    file_fixtures_root,
+    file_remote_factory,
+    file_repo,
+    file_bindings,
+    monitor_task,
+):
+    """Test that on-demand content can be uploaded to Pulp and become immediate content."""
+    remote = file_remote_factory(basic_manifest_path, policy="on_demand")
+    body = {"remote": remote.pulp_href}
+    monitor_task(file_bindings.RepositoriesFileApi.sync(file_repo.pulp_href, body).task)
```
dralley, line 57: Just reminding myself here - this doesn't pollute the shared fixture, does it?

gerrod3: No, all the pulp_file fixtures are unique per test. Each one is generated during the test, I don't think any test uses a shared fixtures.

dralley: Cool, the way pytest fixtures work is a bit unintuitive sometimes

mdellweg:
Well, the one thing I am going to require is, that when we create a new content, we need to create the content artifacts in the same transaction.

Otherwise we open up a window for race conditions, where we have incorrect content in the database.

gerrod3:
> Well, the one thing I am going to require is, that when we create a new content, we need to create the content artifacts in the same transaction.
> 
> Otherwise we open up a window for race conditions, where we have incorrect content in the database.

Ok, I reverted the create logic so now this is only a new check we perform when the content already existed in Pulp.

mdellweg (approved):

pulpcore/app/serializers/content.py
```diff
@@ -74,6 +75,22 @@ def create(self, validated_data):
...
+        if not created:
+            ca_set = content.contentartifact_set
+            if ca_set.count() == 0 or ca_set.filter(artifact=None).exists():
+                cas = [
+                    models.ContentArtifact(artifact=a, content=content, relative_path=rp)
+                    for rp, a in artifacts.items()
+                ]
+                models.ContentArtifact.objects.bulk_create(
+                    cas,
+                    update_conflicts=True,
```
mdellweg, line 90: Is this the thing that basically turns this into an update for all the not yet associated artifacts?

mdellweg (approved):

ggainey (approved):

PR #5511, opened 2024-06-24
Add a sentinal value to CONTENT_ORIGIN

### LINKED ISSUE #5510: CONTENT_ORIGIN prevents some plugins to use openapi command
Some plugins require CONTENT_ORIGIN to be set which is not necessarily the case for the pulpcore-manager openapi command.

mdellweg:
This helps circumventing issues where plugins rely on the setting, but a command is bypassing the checks.

fixes #5510

### DIFF
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -248,6 +248,8 @@ LOGGING = {
 
 DRF_ACCESS_POLICY = {"reusable_conditions": ["pulpcore.app.global_access_conditions"]}
 
+# This is a sentinal value for deploy checks, since we don't want to set a default one.
+CONTENT_ORIGIN = "UNREACHABLE"
 CONTENT_PATH_PREFIX = "/pulp/content/"
 
 API_APP_TTL = 120  # The heartbeat is called from gunicorn notify (defaulting to 45 sec).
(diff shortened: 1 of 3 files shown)


lubosmj (approved):

pulpcore/app/settings.py
```diff
@@ -248,6 +248,8 @@
 
 DRF_ACCESS_POLICY = {"reusable_conditions": ["pulpcore.app.global_access_conditions"]}
 
+# This is a sentinal value for deploy checks, since we don't want to set a default one.
+CONTENT_ORIGIN = "UNREACHABLE"
```
lubosmj, line 252: I still wonder if `None` is not a more valid variant... I guess this is more about the setting not being declared at all, not about its value.

mdellweg: Plugins try to concatenate other strings. I'm almost certain `None` would still blow up.

lubosmj: Aaaah, I got your point. We no longer run the safety check for `CONTENT_ORIGIN` in the settings file after committing cce64770f562d5e80fc89b06086258630d0c7ab8. Now, `content_origin_check` should be able to handle invalid configurations. 👍🏻 🟢

PR #5530, opened 2024-06-27
Update the default value of SECRET_KEY to enable deployment checks

### LINKED ISSUE #5529: `SECRET_KEY` is a boolean by default, causing `pulpcore-manager check --deploy` to fail
This value is expected to be overwritten by a user. However, such information is not mentioned in the docs and due to this, the pulp-all-in-one container does not work out of the box.

### Steps to reproduce the behaviour:
1. Run a pulp-all-in-one image.
2. Shell into the container and run:
```
pulpcore-manager check --deploy
```
```
  File "/usr/local/bin/pulpcore-manager", line 8, in <module>
    sys.exit(manage())
  File "/usr/local/lib/python3.9/site-packages/pulpcore/app/manage.py", line 11, in manage
  File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 436, in execute
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 412, in run_from_argv
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 458, in execute
  File "/usr/local/lib/python3.9/site-packages/django/core/management/commands/check.py", line 76, in handle
    self.check(
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 485, in check
    all_issues = checks.run_checks(
  File "/usr/local/lib/python3.9/site-packages/django/core/checks/registry.py", line 88, in run_checks
    new_errors = check(app_configs=app_configs, databases=databases)
  File "/usr/local/lib/python3.9/site-packages/django/core/checks/security/base.py", line 222, in check_secret_key
    passed_check = _check_secret_key(secret_key)
  File "/usr/local/lib/python3.9/site-packages/django/core/checks/security/base.py", line 209, in _check_secret_key
    len(set(secret_key)) >= SECRET_KEY_MIN_UNIQUE_CHARACTERS
```

### Additional context
https://docs.djangoproject.com/en/5.0/ref/settings/#secret-key
https://pulpproject.org/pulpcore/docs/admin/reference/settings/?h=secret_key#secret_key

lubosmj:
closes #5529

### DIFF
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -61,7 +61,8 @@ CHUNKED_UPLOAD_DIR = "upload"
 # List of upload handler classes to be applied in order.
 FILE_UPLOAD_HANDLERS = ("pulpcore.app.files.HashingFileUploadHandler",)
 
-SECRET_KEY = True
+# SECURITY WARNING: this should be set to a unique, unpredictable value
+SECRET_KEY = "SECRET"
 
 # Key used to encrypt fields in the database
 DB_ENCRYPTION_KEY = "/etc/pulp/certs/database_fields.symmetric.key"
--- a/CHANGES/5529.bugfix
+++ b/CHANGES/5529.bugfix
@@ -1,0 +1,1 @@
+Added a note about the default value of the `SECRET_KEY` setting.
--- a/pulpcore/app/checks.py
+++ b/pulpcore/app/checks.py
@@ -20,6 +20,21 @@ def content_origin_check(app_configs, **kwargs):
 
 
 @register(deploy=True)
+def secret_key_check(app_configs, **kwargs):
+    messages = []
+    if getattr(settings, "SECRET_KEY", "SECRET") == "SECRET":
+        messages.append(
+            CheckError(
+                "SECRET_KEY is a required setting but it was not configured. It does not "
+                "come pre-configured by the installation and it should be set to a unique, "
+                "unpredictable value.",
+                id="pulpcore.E001",
+            )
+        )
+    return messages
+
+
+@register(deploy=True)
 def storage_paths(app_configs, **kwargs):
     warnings = []
 


lubosmj:
I understand that setting it to `True` could be designed as a safe guard for some of our users who accidentally used a non-secure SECRET_KEY.

pulpcore/app/settings.py
```diff
@@ -61,7 +61,8 @@
 # List of upload handler classes to be applied in order.
 FILE_UPLOAD_HANDLERS = ("pulpcore.app.files.HashingFileUploadHandler",)
 
-SECRET_KEY = True
+# SECURITY WARNING: this should be set to a unique, unpredictable value
+SECRET_KEY = "secret"
```
mdellweg, line 65: Look at what django uses on a new project. I think that will trigger on a deploy test also.

lubosmj: It is set to an empty string, see https://github.com/django/django/blob/0fdcf1029cea2d43bd68c5270f48e0f7deab5e47/django/conf/global_settings.py#L270. However, this fails in our CI with the following error:  `django.core.exceptions.ImproperlyConfigured: The SECRET_KEY setting must not be empty` (https://github.com/pulp/pulpcore/actions/runs/9701722403/job/26776031964#step:10:606). That is another reason why we set it to `True` -- to bypass the checks.

lubosmj: My idea is to have a fully functioning pulp-all-in-one image that can be used for any kind of work needed. But, I think @git-hyagi 's comment on the issue about performing the generation of the secret key inside the pulp-oci-images repository (maybe up.sh) seems to be a better solution.

mdellweg: When bootstrapping a new project, django creates something like
```
# SECURITY WARNING: keep the secret key used in production secret!
SECRET_KEY = 'django-insecure-dkatkrnai%a#zymt_ezju^wne^1a*-py1=534kkw1y8*qpq$b='
```

lubosmj: Aaah, now I see it: https://github.com/django/django/blob/0fdcf1029cea2d43bd68c5270f48e0f7deab5e47/django/core/management/commands/startproject.py#L19. Sorry.

But, this is generated after running the django-admin startproject command. I am unsure this fits our scenario where we just take the settings file and do not care about the values until we boot up Pulp.

mdellweg: I think there's a check that will check for that substring. Can we not make use of this mechanism?

lubosmj: 💭 What if we did the same thing as for `CONTENT_ORIGIN`? Based on the documentation, `CONTENT_ORIGIN` and `SECRET_KEY` are the only two required settings: https://github.com/pulp/pulpcore/blob/b2e4e22ffff3b2fc72c8eaf566e0408a5c60d947/docs/configuration/settings.rst?plain=1#L6-L10.

Do you think we can have this "secret" value set as it is right now, but on top of it, I would introduce `register(deploy=True)` check for `SECRET_KEY` in `checks.py`, similarly to what you did for `CONTENT_ORIGIN`? This looks to me as the most viable solution.

mdellweg: I thought django had exactly such a check. Otherwise that sounds absolutely fine!

dkliban (approved):

ggainey (approved):

Issue #4608, opened 2023-02-17
Add option to download artifacts using http/https


maartenbeeckmans:
**Is your feature request related to a problem? Please describe.**
The remote feature does not work with http directories on the internet as there is no standard on providing the metadata.

**Describe the solution you'd like**
An api endpoint where you can define a http link to a remote artifact and optionally a checksum and authentication options.

**Describe alternatives you've considered**
Another possibility to add artifacts now is downloading these locally with wget/curl, verifying the checksum and uploading it to pulp using the api.

**Chat about this topic**

@maartenbeeckmans
the main issue I have with the file plugin is that remotes not really work. The plugin expects a manifest file in a specific format which is not used by remote repos such as https://fedora.cu.be/linux/releases/37/Cloud/x86_64/images/ (which has the CHECKSUM file but different format) or https://cloud.debian.org/images/cloud/bullseye/20230124-1270/ (with the SHA512SUMS file which has another different format).
but this is a difficult topic as there is no standard to provide metadata like with rpm/deb

@dkliban
it sounds like you are interested in the file plugin being able to scrape an HTML page to discover what content is available

@gerrod3
Maarten Beeckmans: So when you want to request the file from Pulp, it would have the same relative-path on the upstream repo? If so file a feature request to add pull-through caching to pulp_file
or we could add what dkliban suggested and allow pulp_file to sync more repo types

@maartenbeeckmans
the same relative path is not really a requirement, it would be useful to just allow it to sync those repos so we don't have to do it manually

@dkliban
Maarten Beeckmans
Maarten Beeckmans: how do you do it now? do you scrape the HTML to discover the files?

@maartenbeeckmans
maybe add support for the filetype debian is using. It can be a possibility to add this to the remote, but that is just the output of the sha512sum command which is soms sort of standard (with support for other shasums)
but scraping seems difficult to implement correctly
but for me specifying this the file I want and a checksum to only download that content unit is also a valid option

@ggainey
also, scraping only works for upstream repos that enable directory-indexing - which more and more sites are disallowing

@gerrod3
ggainey: What are they doing instead? How would you find the files?

@ggainey
gerrod: consider rpm repos - there's "a" file at a known location, .../repo/repomd.xml, and you only "know about" files that repomd.xml tells you
a lot of content is intended to be used by applications that "know" what to look for - index.html is tehre to let human beings poke around
allowing directory-indexing tends to get flagged as apotential security exposure by a number of security-scanners these days
[...]

PR #5560, opened 2024-07-11
Add ability to download file for artifact/content uploads

gerrod3:
fixes: #4608

### DIFF
--- a/CHANGES/plugin_api/4608.feature
+++ b/CHANGES/plugin_api/4608.feature
@@ -1,0 +1,1 @@
+Added new `url` field to UploadSerializerFieldsMixin that will download the file used for the content upload task.
--- a/pulpcore/plugin/serializers/content.py
+++ b/pulpcore/plugin/serializers/content.py
@@ -4,12 +4,14 @@ from tempfile import NamedTemporaryFile
 
 from django.db import DatabaseError
 from rest_framework.serializers import (
+    CharField,
     FileField,
     Serializer,
     ValidationError,
 )
+from urllib.parse import urlparse
 from pulpcore.app.files import PulpTemporaryUploadedFile
-from pulpcore.app.models import Artifact, PulpTemporaryFile, Upload, UploadChunk
+from pulpcore.app.models import Artifact, PulpTemporaryFile, Remote, Upload, UploadChunk
 from pulpcore.app.serializers import (
     RelatedField,
     ArtifactSerializer,
@@ -21,6 +23,8 @@ from pulpcore.app.util import get_domain_pk
 
 class UploadSerializerFieldsMixin(Serializer):
     """A mixin class that contains fields and methods common to content upload serializers."""
+
+    REMOTE_CLASS = Remote
 
     file = FileField(
         help_text=_("An uploaded file that may be turned into the content unit."),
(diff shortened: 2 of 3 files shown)


pulpcore/app/viewsets/content.py
```diff
@@ -86,6 +92,25 @@ def destroy(self, request, pk):
...
             return Response(data, status=status.HTTP_409_CONFLICT)
 
+    @extend_schema(
+        summary="Create an Artifact from a URL",
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    @action(detail=False, methods=["POST"], serializer_class=DownloadArtifactSerializer)
+    def remote_upload(self, request):
+        """
+        Creates an Artifact by downloading from url. Returns a creation Task.
```
mdellweg, line 102: Am I right that the task really is the reason why we need to make this a "new" action?
We really don't know how to return polymorphic results yet, right?
Actually, I'd rather not do anything for bare artifacts here and only allow the content creation (into repositories) from remote url.

gerrod3: Well I thought a new endpoint would be easier to implement than a polymorphic create. I can remove the artifact download for now.

mdellweg: You know, how i constantly try to move the pulp api away from handling raw artifacts? This would be part of it.
If i could decide, all the artifact api's should be read only, or even non existent.

pulpcore/plugin/serializers/content.py
```diff
@@ -15,11 +15,12 @@
     ArtifactSerializer,
     NoArtifactContentSerializer,
     SingleArtifactContentSerializer,
+    URLFetchSerializerMixin,
 )
 from pulpcore.app.util import get_domain_pk
 
 
-class UploadSerializerFieldsMixin(Serializer):
+class UploadSerializerFieldsMixin(URLFetchSerializerMixin, Serializer):
```
mdellweg, line 23: At this point i think we can straighten the class hierarchy and fold the mixin into this one, no?

Also side quest: This being a mixin, why does it inherit `Serializer`?

gerrod3: I don't know why, but if it wasn't a serializer I couldn't have the `url` field in it.  Django would complain that it couldn't find the field.

CHANGES/plugin_api/4608.feature
```diff
@@ -0,0 +1 @@
+Added new `url` field to UploadSerializerFieldsMixin that will download the file used for the content upload task.
```
mdellweg, line 1: Actually, now i'm asking myself it this is a user visible change and be in CHANGES directly.
Plugins don't really have a say in this once they implemented the upload logic.

pulpcore/plugin/serializers/content.py
```diff
@@ -34,6 +36,45 @@ class UploadSerializerFieldsMixin(Serializer):
...
+                self.context["remote_kwargs"] = kwargs
+
+        return url_parse._replace(netloc=url_parse.netloc.split("@")[-1]).geturl()
+
+    def download(self, url, expected_digests=None, expected_size=None):
+        """
+        Downloads & returns the file from the url.
+
+        Plugins can overwrite this method on their content serializers to get specific download
+        behavior for their content types.
```
mdellweg, line 62: I guess a plugin may want to change the type of Remote if anything. Should we make this a class parameter right away?

CHANGES/plugin_api/4608.feature
gerrod3: Technically plugins do have a say, since they can decide to expose the field or not. I looked at your old PR when we added the upload field, and it only had plugin_api changelogs.

mdellweg: Not just because I did it meanns it was right... XD

pulpcore/plugin/serializers/content.py
```diff
@@ -34,6 +38,45 @@ class UploadSerializerFieldsMixin(Serializer):
...
+        """Parse out the auth if provided."""
+        url_parse = urlparse(value)
+        if url_parse.username or url_parse.password:
+            kwargs = {"username": url_parse.username, "password": url_parse.password}
+            if self.context.get("remote_kwargs"):
+                self.context["remote_kwargs"].update(kwargs)
+            else:
+                self.context["remote_kwargs"] = kwargs
+
+        return url_parse._replace(netloc=url_parse.netloc.split("@")[-1]).geturl()
```
mdellweg, line 57: I hope we keep querystrings here, sp presigned urls will work.

mdellweg (approved):

daviddavis:
This looks great. And it seems like uploading by url work will without the user having to create a remote? 

We offer this feature to our users today but the code exists outside of Pulp and obviously we'd rather see that Pulp supports it.

pulpcore/plugin/serializers/content.py
gerrod3: Yes it does. Querystrings are their own field in the parsed tuple, so they will remain in the new url.

ggainey (approved):

PR #5561, opened 2024-07-12
Provide more task info during test failures

daviddavis:
We're trying to debug a test failure in the CI that we can't reproduce locally:

https://github.com/pulp/pulp_deb/actions/runs/9898615528/job/27345874499?pr=1061#step:15:3505

Part of the challenge in debugging this is that there's not enough information to go on because the CI is only outputting the task error description. Not sure if there's a better way to handle this but in these cases it seems like printing the entire task details somehow would be more helpful.

[noissue]

### DIFF
--- a/pulpcore/tests/functional/utils.py
+++ b/pulpcore/tests/functional/utils.py
@@ -28,7 +28,7 @@ except ImportError:
 
         def __init__(self, task):
             """Provide task info to exception."""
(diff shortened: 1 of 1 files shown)


mdellweg (approved):

mdellweg:
I tend to use the cid of the test to find the tasks container logs. But it's not too easy to do either.

ggainey (approved):

PR #5591, opened 2024-07-17
Fix old content being persistent on the website server

pedro-psb:
The current `rsync` command only adds new content to the server, it doesn't remove old files that are no longer being built. Because of that, we have left-overs from older build, such as https://pulpproject.org/pulp-cli/changes/changelog/

The `--delete` option deletes files in <target> that are not present in <source> anymore ([reference](https://explainshell.com/explain?cmd=rsync+--delete)).

**DISCLAIMER:** I've barely used rsync in my life


mdellweg (approved):

ggainey (approved):

mdellweg:
Please rebase.

PR #5605, opened 2024-07-18
Add ability to cancel a TaskGroup

gerrod3:
fixes: #4125

### DIFF
--- a/pulpcore/app/serializers/task.py
+++ b/pulpcore/app/serializers/task.py
@@ -183,13 +183,20 @@ class TaskGroupSerializer(ModelSerializer):
         )
 
 
-class TaskCancelSerializer(ModelSerializer):
+class TaskCancelSerializer(serializers.Serializer):
     state = serializers.CharField(
         help_text=_("The desired state of the task. Only 'canceled' is accepted."),
+        required=True,
     )
 
+    def validate_state(self, value):
+        if value != "canceled":
+            raise serializers.ValidationError(
+                _("The only acceptable value for 'state' is 'canceled'.")
+            )
+        return value
+
     class Meta:
-        model = models.Task
         fields = ("state",)
 
 
--- a/pulpcore/app/viewsets/task.py
+++ b/pulpcore/app/viewsets/task.py
@@ -1,10 +1,12 @@
 from gettext import gettext as _
 
+from django.db import transaction
 from django.db.models import Prefetch
 from django_filters.rest_framework import filters
 from drf_spectacular.utils import extend_schema, inline_serializer
 from rest_framework import mixins, status
 from rest_framework.decorators import action
+from rest_framework.exceptions import PermissionDenied
 from rest_framework.response import Response
 from rest_framework.serializers import DictField, URLField, ValidationError
 
@@ -40,7 +42,8 @@ from pulpcore.app.viewsets.custom_filters import (
     CreatedResourcesFilter,
 )
 from pulpcore.constants import TASK_INCOMPLETE_STATES, TASK_STATES
-from pulpcore.tasking.tasks import dispatch, cancel_task
+from pulpcore.tasking.tasks import dispatch, cancel_task, cancel_task_group
+from pulpcore.app.role_util import get_objects_for_user
 
 
 class TaskFilter(BaseFilterSet):
@@ -211,11 +214,10 @@ class TaskViewSet(
         responses={200: TaskSerializer, 409: TaskSerializer},
     )
     def partial_update(self, request, pk=None, partial=True):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
         task = self.get_object()
-        if "state" not in request.data:
-            raise ValidationError(_("'state' must be provided with the request."))
-        if request.data["state"] != "canceled":
-            raise ValidationError(_("The only acceptable value for 'state' is 'canceled'."))
         task = cancel_task(task.pk)
         # Check whether task is actually canceled
         http_status = (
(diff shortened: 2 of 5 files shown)


pulpcore/tasking/tasks.py
```diff
@@ -250,3 +250,30 @@ def cancel_task(task_id):
         cursor.execute("SELECT pg_notify('pulp_worker_cancel', %s)", (str(task.pk),))
         cursor.execute("NOTIFY pulp_worker_wakeup")
     return task
+
+
+def cancel_task_group(task_group_id):
```
mdellweg, line 255: Now here's a problem: While a task in the task group is marked as cancelling it is still actually running and can continue to emit new subtasks into the task group.

gerrod3: Yes. That is why there is a `while` loop that will try to cancel all running tasks twice, and why the endpoint can still fail and return a 409. We can't sit and wait for every task to be cancel as we risk the endpoint timing out.

mdellweg: Exactly. But there's still a window where this endpoint reports all (to that point) known tasks have been set to cancelling (aka success), but new subtasks that won't be canceled pop up in the task group.
I think we need to mark the task-group itself as canceled and prevent new tasks to be dispatched at that point.

gerrod3: Ok, I updated the task dispatch logic to check if a task's group has already dispatched all its tasks and if so cancel the task there. In `cancel_task_group` I updated it to immediately set `all_tasks_dispatched` to True so that any future subtasks will get auto-canceled. This also simplified the logic for canceling requiring only one pass through current subtasks.

pulpcore/app/viewsets/task.py
```diff
@@ -270,6 +276,32 @@ def scope_queryset(self, qs):
             qs = qs.filter(tasks__in=tasks)
         return qs
 
+    @extend_schema(
+        description="This operation cancels a task group.",
+        summary="Cancel a task group",
+        operation_id="task_groups_cancel",
+        request=TaskCancelSerializer,
```
mdellweg, line 283: Is this really the right one?

pulpcore/app/viewsets/task.py
```diff
@@ -270,6 +276,32 @@ def scope_queryset(self, qs):
...
+    def partial_update(self, request, pk=None, partial=True):
+        task_group = self.get_object()
+        if not get_objects_for_user(
+            request.user, "core.change_task", task_group.tasks.all()
+        ).exists():
+            raise PermissionDenied()
+        if "state" not in request.data:
+            raise ValidationError(_("'state' must be provided with the request."))
+        if request.data["state"] != "canceled":
+            raise ValidationError(_("The only acceptable value for 'state' is 'canceled'."))
```
mdellweg, line 295: Should this not be in the CancelSerializer?

pulpcore/app/viewsets/task.py
```diff
@@ -270,6 +276,32 @@ def scope_queryset(self, qs):
...
+        summary="Cancel a task group",
+        operation_id="task_groups_cancel",
+        request=TaskCancelSerializer,
+        responses={200: TaskGroupSerializer, 409: TaskGroupSerializer},
+    )
+    def partial_update(self, request, pk=None, partial=True):
+        task_group = self.get_object()
+        if not get_objects_for_user(
+            request.user, "core.change_task", task_group.tasks.all()
+        ).exists():
```
mdellweg, line 290: This triggers if you have perms on just one of the tasks.
(Yes this is a very esoteric situation.)

pulpcore/app/viewsets/task.py
gerrod3: Yes, I am reusing this serializer since it gives the cancel apis the same signature, also there is nothing task-specific in the serializer.

pulpcore/app/viewsets/task.py
gerrod3: That is correct,  do you have a suggestion for what it should be?

mdellweg: I was thinking if you own _all_ the tasks you can cancel the group.

pulpcore/app/viewsets/task.py
mdellweg: So can we also _use_ the serializer to leverage some field-level validation?

pulpcore/tasking/tasks.py
mdellweg: Thank you. Now all of a sudden, the `all_tasks_dispatched` field has a meaning after all.

pulpcore/app/serializers/task.py
```diff
@@ -183,13 +183,20 @@ class Meta:
         )
 
 
-class TaskCancelSerializer(ModelSerializer):
+class TaskCancelSerializer(serializers.Serializer):
```
mdellweg, line 186: :+1:
Oh, oh...

pulpcore/app/viewsets/task.py
```diff
@@ -194,11 +194,10 @@ def get_queryset(self):
         responses={200: TaskSerializer, 409: TaskSerializer},
     )
     def partial_update(self, request, pk=None, partial=True):
+        serializer = self.get_serializer(data=request.data)
```
mdellweg, line 197: This doesn't use the TaskCancelSerializer. Right?

gerrod3: It does, if you look down to line 217 you'll see that `get_serializer_class` returns TaskCancelSerializer when the action is PATCH.

mdellweg: Ohhh. My bad i only looked at get_serializer.

mdellweg (approved):

lubosmj:
I think we are good to go. @gerrod3, can you, please, resolve the merge conflicts?

ipanova (dismissed):

ipanova:
@gerrod3 can you fix the linting error?

lubosmj (approved):

ipanova (approved):

PR #5618, opened 2024-07-22
Casts a repo object before checking permission

git-hyagi:
This change is similar to #4933, but for the condition function where domain is not enabled.
[noissue]

fixes: #5619

### DIFF
--- a/CHANGES/5619.bugfix
+++ b/CHANGES/5619.bugfix
@@ -1,0 +1,2 @@
+Fixed a bug in `has_repo_or_repo_ver_param_model_or_obj_perms` function to make
+it work with plugin Repositories.
--- a/pulpcore/app/global_access_conditions.py
+++ b/pulpcore/app/global_access_conditions.py
@@ -479,9 +479,9 @@ def has_repo_or_repo_ver_param_model_or_obj_perms(request, view, action, permiss
     )
     serializer.is_valid(raise_exception=True)
     if repository := serializer.validated_data.get("repository"):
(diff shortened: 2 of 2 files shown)


mdellweg:
This sounds reasonable, but also looks like a bugfix to me.
In that case we absolutely need a changelog.

lubosmj:
@git-hyagi, to which pulpcore versions we should backport this to make it available for pulp-container?

CHANGES/5619.bugfix
```diff
@@ -0,0 +1 @@
+Fixed a bug where containers couldn't be built from repository versions when using RBAC.
```
lubosmj, line 1: Please, make this changelog more generic. Many people have no idea what is the container building in this context. Is it about building pulp-oci-images?

git-hyagi:
> @git-hyagi, to which pulpcore versions we should backport this to make it available for pulp-container?

I think to any version >=3.49  (3.49 and 3.55, right?) which is the lower bound of pulp-container plugin: https://github.com/pulp/pulp_container/blob/main/requirements.txt#L2C1-L2C17

CHANGES/5619.bugfix
```diff
@@ -0,0 +1 @@
+Fixed a bug in a permission function to make it work with plugin Repositories.
```
mdellweg, line 1: Can you name the actual function? It is user visible inside of the access policy, right?

mdellweg (approved):

lubosmj (approved):
Thank you!

PR #5636, opened 2024-07-25
Implement batching for artifact deletion to handle large repositories

### LINKED ISSUE #5635: Error when reclaiming space from repository with over 65k packages due to parameter limit
**Version**
- pulpcore 3.56

**Describe the bug**
Attempting to reclaim space from a repository containing more than 65k packages results in an error:

`Error: Task failed: 'sending query and params failed: number of parameters must be between 0 and 65535'`

**To Reproduce**
1. Create a file repository with over 65k packages, as described [here](https://github.com/pulp/pulpcore/issues/5375#issuecomment-2113153654).
2. Sync the repository.
3. Run the reclaim space task: `pulp repository reclaim --repository $FILE_REPOSITORY_HREF`.

**Expected behavior**
The task should complete successfully, and the disk space should be reclaimed.

**Additional context**
This issue is likely similar to the one described here https://github.com/pulp/pulpcore/issues/5375. Implementing batching for artifact deletion might resolve the problem.

hstct:
closes #5635

### DIFF
--- a/pulpcore/app/tasks/reclaim_space.py
+++ b/pulpcore/app/tasks/reclaim_space.py
@@ -69,10 +69,15 @@ def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
             ca_to_update.append(ca)
 
     ContentArtifact.objects.bulk_update(objs=ca_to_update, fields=["artifact"], batch_size=1000)
-    artifacts_to_delete = Artifact.objects.filter(pk__in=artifact_pks)
+
+    # Process artifacts in batches
+    batch_size = 1000
+    artifact_pks_list = list(artifact_pks)
+    num_batches = (len(artifact_pks_list) // batch_size) + 1
+
     progress_bar = ProgressReport(
         message="Reclaim disk space",
-        total=artifacts_to_delete.count(),
+        total=len(artifact_pks_list),
         code="reclaim-space.artifact",
         done=0,
         state="running",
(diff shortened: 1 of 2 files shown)


gerrod3:
Does this have a performance improvement? I thought that `iterator` already did baching behind the scenes?

mdellweg:
If i understood correctly, the queryset had issues when the list of PKs exceeded 65k (65535 if i had to guess).
And really what's the use of telling postgres _all_ the PK's and then wanting to get only the first 1000 objects?

pulpcore/app/tasks/reclaim_space.py
```diff
@@ -81,21 +86,26 @@ def reclaim_space(repo_pks, keeplist_rv_pks=None, force=False):
...
-            # Just log and skip the artifact deletion in this case
-            log.debug(e)
-        else:
-            progress_bar.done += 1
-            counter += 1
-
-        if counter >= interval:
-            progress_bar.save()
-            counter = 0
+    for i in range(num_batches):
```
mdellweg, line 89: Maybe using:
https://docs.python.org/3/library/itertools.html#itertools.batched

Will make the code cleaner.

hstct: Ok, I tried adding this which is very straight forward and would definitely make it cleaner, however you need python 3.10 to import `batched` from itertools and currently the oci-env is running 3.9.

Meaning this will definitely fail the CI, won't it?

hstct:
> If i understood correctly, the queryset had issues when the list of PKs exceeded 65k (65535 if i had to guess). And really what's the use of telling postgres _all_ the PK's and then wanting to get only the first 1000 objects?

Yes, that is basically the issue, hence why batching in the iterator won't work

pulpcore/app/tasks/reclaim_space.py
mdellweg: Ohhh, right... It looked so good though.

mdellweg (approved):

dkliban (approved):

PR #5639, opened 2024-07-25
Change url field name to avoid plugin conflicts

### LINKED ISSUE #5633: Upload by url doesn't work for certain plugins
[A feature was added recently where you can create content by url](https://github.com/pulp/pulpcore/pull/5560). The problem is that the field name is url which conflicts with some plugins (or pulp-rpm at least). The new field is called url but [this field already exists for packages except it's read_only](https://github.com/pulp/pulp_rpm/blob/bdb538f470e1cc9388f59d9ef2ba6c9ca345ad76/pulp_rpm/app/serializers/package.py#L78) so you get an error if you try to use it:

```
$ http :/pulp/api/v3/content/rpm/packages/ url=https://fixtures.pulpproject.org/rpm-unsigned/bear-4.1-1.noarch.rpm
{
    "non_field_errors": [
        "Exactly one of upload, url, artifact, file must be specified."
    ]
}
```

I suppose plugins could fix this on their own but I think ideally this would be fixed in pulpcore so the name of the field is consistent across plugins. I'm happy to open a patch to rename the field if that makes sense (maybe `file_url`?).

daviddavis:
Change the field name `url` to `file_url` to avoid conflict. In pulp-rpm there's already a Package field `url` that is causing problems.

fixes #5633

### DIFF
--- a/CHANGES/plugin_api/5633.bugfix
+++ b/CHANGES/plugin_api/5633.bugfix
@@ -1,0 +1,2 @@
+Change field name `url` to `file_url` on UploadSerializerFieldsMixin to avoid conflicting fields in
+plugins.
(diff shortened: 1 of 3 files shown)


dkliban (approved):

mdellweg:
I think, we should backport this also.

mdellweg (approved):

PR #5654, opened 2024-07-29
Correct the gitleaks.toml to use (correctly) paths instead of files.

ggainey:
[noissue]


mdellweg (approved):
So this was the actual file format?

ggainey:
> So this was the actual file format?

Yes indeed - from the "common gotchas" section of the related how-to, 'Confusing "files" and "paths" (files is really "file names" and can't include any parts of the path, definitely not "intuitively obvious to the casual observer")'

PR #5658, opened 2024-07-31
Experiment: remove setuptools requirement

mdellweg:


mdellweg:
I do not think this is quite honest, because we still import from pkg_resources.

gerrod3:
> I do not think this is quite honest, because we still import from pkg_resources.

Where exactly do we import from them?

mdellweg:
> > I do not think this is quite honest, because we still import from pkg_resources.
> 
> Where exactly do we import from them?

I first just removed the requirement, and tests were still passing. Probably because on pyton3.9 with pip you are almost guaranteed to have a version of setuptools.

lubosmj (approved):

dkliban (approved):

PR #5659, opened 2024-07-31
Skip replica syncing when replicated distributions were not updated

### LINKED ISSUE #5493: Pulp replication should leverage optimsations to significantly improve performance
At the moment Pulp replication is not particularly quick, it can take several minutes for it to identify that in fact there are no changes necessary.   Because it spends so much effort trying to make this determination, this clogs up the workers from doing the important work of actually syncing what has changed (say 1 repo out of hundreds), this in turn can lead to an increasing backlog of requested replications, which significantly slows down Pulp competing any given replication request.

Instead of Pulp spending time checking repodata of every repository, it should instead just use the distributions API on the main host to see which distributions have a last modified timestamp newer than when the last replication took place.  This should allow it to make a determination of which repos actually require repodata examination pretty much instantaneously, and only then kick off tasks for checking those repositories that have actually changed, if indeed any have.

We can help reduce the number of replication requests sent to Pulp by checking if there's already pending replication task groups that have not started running, however this will still lead to lower performance than we would like, we'd like to be able to say quickly that a package added to a repository is available everywhere, so that activities such as user self-service management of repositories through our application feels responsive.

**Additional context**
@bmbouter  and I chatted about this yesterday.

lubosmj:
closes #5493


lubosmj:
I am cleaning up the code. However, the core functionality should not change. You can review that part, especially `_requires_syncing` and tests.

gerrod3:
This may be a silly idea, but what if we were to manually update a distribution's last_updated field anytime a new repo/repo-version/publication is created or deleted? If both the upsteam and downstream Pulps had this behavior would we need a new model?

lubosmj:
I think we will not need a new model. The problem here is that the Handler in content-app actually decides which repository version and which publication is going to be used for serving data. We would need to extract these calls and put it somewhere. Also, is there a guarantee that upstream instances will be ever upgraded?

lubosmj:
I designed this under the assumption that we cannot touch existing upstream Pulp instances. With that, I used just the API that already exists. Otherwise, it would make sense to think about the idea @gerrod3 sketched out.

mdellweg:
> I designed this under the assumption that there is no guarantee that we can touch/upgrade existing upstream Pulp instances. With that, I used just the API that already exists. Otherwise, it would make sense to think about the idea @gerrod3 sketched out.

But why that assumption? I would be absolutely beneficial if upstream were allowed to tell the replica when something changed last. And if that information is not (yet) available, you just go an sync.
Starting to guess around to avoid just that is not worth the trouble.

lubosmj:
Okay. Fair. What options do we have?
1. Introduce a new API endpoint that will share the code of the Handler and returns the create timestamp of a publication or version? The replica will check this timestamp on its side. 
2. Update pulp_last_updated to datetime.now() on a distribution every time when a new publication for the attached repository is created? However, this publication may not be ever used by the distribution.

mdellweg:
Absolutely not a new api endpoint.
What is see is a content_timestamp on the distribution serializer.

lubosmj:
Looks like the option number 2 is the most promising. We will not care about unnecessary updates.

Now, a next question stands out. Do we want to have a lifecycle hook on the publication's save call so it will determine whether a referencing repository_version/repository is used by any of the distributions? In case of success, we will update all the distributions possibly utilizing this publication, correct?

```
new publication -> repository/repository_version <- distribution1
                                            <- distribution2
```

lubosmj:
I am converting this back to draft. We have just recently added `pulp_last_updated` which currently does not handle publications (https://github.com/pulp/pulpcore/issues/5033). This PR will extends its capabilities to publications. Whenever an underlying publication (used by autopublish, repository) is created/deleted, we will update the timestamps for all respective distributions.

mdellweg:
> We have just recently added `pulp_last_updated` which currently does not handle publications (#5033).

`pulp_last_updated` is around for at least 4 years. Not quite what I would call "recent". And it is there to record when a model has last been changed. We cannot simply change this logic.
But we can use these fields across associated models to calculate since when content has not been modified behind an address on the content app and add that result to the serializer.

lubosmj:
Okay, a new `content_last_updated` field will do the needful.

lubosmj:
@mdellweg, how would you imagine doing this calculation? I do not see any way of remembering this value apart from storing it on the model. Each time a distribution is requested on viewset, we will run the following code?

```
pub = publications.objects.filter(repository=repo_pk).order_by("-pulp_created").first()
if distro.content_last_updated != pub.pulp_created:
   distro.content_last_updated = pub.pulp_created
   distro.save(update_fields=["content_last_updated"])
```

Is not it better to add creation/deletion hooks on the publication model that will bulk_update all relevant distributions?

lubosmj:
> If both the upsteam and downstream Pulps had this behavior would we need a new model?

Actually, yes. Because the replicated repositories will have a different pulp_last_updated timestamp.

pulpcore/app/models/replica.py
```diff
@@ -28,9 +33,20 @@ class UpstreamPulp(BaseModel, AutoAddObjPermsMixin):
...
         ]
+
+
+class LastUpdatedRecord(BaseModel):
+    distribution = models.ForeignKey(Distribution, on_delete=models.CASCADE)
+    upstream_pulp = models.ForeignKey(UpstreamPulp, on_delete=models.CASCADE)
+    last_updated = models.DateTimeField()
+
+    class Meta:
+        unique_together = ("distribution", "upstream_pulp")
```
mdellweg, line 52: First of all, this won't do. Because there is no guarantee, that replicate is what updated the thing last.
Second I challange you to not need a new database table at all.

You can put a timestamp on the upstream_pulp object if you want.

pulpcore/app/serializers/publication.py
```diff
@@ -160,6 +162,45 @@ class Meta(ContentGuardSerializer.Meta):
...
 
+class ContentLastUpdatedField(serializers.DateTimeField, serializers.ReadOnlyField):
+    """
+    A serializer field representing the last updated timestamp for a used content source.
+    """
+
+    def get_attribute(self, instance):
+        return instance
+
+    def to_representation(self, value):
```
mdellweg, line 173: What if the distribution was changed from a newer to an older repository version?
You are missing a ton of edge cases here.

mdellweg (changes requested):

pulpcore/app/models/replica.py
lubosmj: Well. What if we want to have a better optimization granularity? There might be a case where only one of ten distributions was updated.

lubosmj: The timestamp is taken from the upstream content_last_updated. We just copy it and then later use to check if the timestamps are equal or not when deciding whether to skip syncing.

mdellweg: Comparing timestamps for equality is never a good idea.
What if you edit the pulp_upstream object? There are already too many corner cases to ask for more granularity.

lubosmj: Okay. Sounds fair.

lubosmj:
I have introduced a new field on the distribution serializer -- content_last_updated. It represents the pulp_last_updated timestamp of the **content source** (i.e., repository_version or publication). Replica stores this timestamp locally in a record and **compares** it with the upstream during the replication. In addition to that, I am now recording the timestamp of the last replication to deal with a scenario where a base-path or api-root could be changed by a user.

The following should help us monitor all changes on the upstream:
- server config changed -> record.last_replication < server.pulp_last_updated
- upstream distribution changed -> record.last_replication < upstream_distro.pulp_last_updated
- upstream content source changed -> record.content_last_updated != upstream_distro.content_last_updated

Please, let me know if missed something. I really think that we should keep the record table to be able to observe potential changes on the upstream when it comes to distributing old or newer publications, as users decide.

mdellweg:
Also consider this:
Each time we implemented anything closely looking like an optimized sync. The first thing after it went life is an ask for a force flag. Can you just add it from the start?

lubosmj:
I removed the additional table and made the changes @mdellweg requested. Please, review from scratch.

pulpcore/app/models/replica.py
```diff
@@ -28,9 +30,16 @@ class UpstreamPulp(BaseModel, AutoAddObjPermsMixin):
...
         unique_together = ("name", "pulp_domain")
         permissions = [
             ("replicate_upstreampulp", "Can start a replication task"),
             ("manage_roles_upstreampulp", "Can manage roles on upstream pulps"),
         ]
+
+    def set_last_replication_timestamp(self):
+        self.last_replication = datetime.now(timezone.utc)
+        # enforce the update without changing pulp_last_updated
+        self.save(update_fields=["last_replication"])
```
mdellweg, line 45: I think we still need a hook to nullify this on every other save to the model.
Maybe we need to use "Skip hooks" here.

pulpcore/app/replica.py
```diff
@@ -196,6 +198,16 @@ def sync_params(self, repository, remote):
         """This method returns a dict that will be passed as kwargs to the sync task."""
         raise NotImplementedError("Each replicator must supply its own sync params.")
 
+    def requires_syncing(self, distro):
+        no_content_change_since = _read_content_change_since_timestamp(distro)
+        if no_content_change_since and self.server.last_replication:
+            if self.server.last_replication < self.server.pulp_last_updated:
+                return True
```
mdellweg, line 205: We can just fall through to the other True here.
```suggestion
```

pulpcore/app/replica.py
```diff
@@ -196,6 +198,16 @@ def sync_params(self, repository, remote):
...
+            if self.server.last_replication > no_content_change_since:
+                return False
```
mdellweg, line 207: If we 
```suggestion
            if self.server.last_replication > parse_datetime(no_content_change_since):
                return False
```
here, you probably don't need the `_read_content_change_since_timestamp` function.

pulpcore/app/serializers/publication.py
```diff
@@ -160,6 +160,24 @@ class Meta(ContentGuardSerializer.Meta):
...
+    """
+
+    def get_attribute(self, instance):
+        return instance
+
+    def to_representation(self, distribution):
+        publication = distribution.publication
+        repo_version = distribution.repository_version
+
+        if publication or repo_version and not distribution.SERVE_FROM_PUBLICATION:
```
mdellweg, line 175: Is that meant like this?
```suggestion
        if publication or (repo_version and not distribution.SERVE_FROM_PUBLICATION):
```
(Is a distribution with SERVE_FROM_PUBLICATION being attached to a repository_version even valid?)

mdellweg: I wonder if a SerializerMethod field would have been easier.
Also what is the best way a Plugin can subclass this behaviour with better knowledge?

pulpcore/app/serializers/replica.py
```diff
@@ -85,6 +85,10 @@ class UpstreamPulpSerializer(ModelSerializer, HiddenFieldsMixin):
         required=False,
     )
 
+    last_replication = serializers.DateTimeField(
+        help_text="Timestamp of the last replication that occurred.", read_only=True
+    )
+
```
mdellweg, line 91: Do we want to expose this?
Given that we need to nullify it when the UpstreamPulp was touched, this may confuse users.
Or we add more logic around interpreting these two values:

`last_relevant_replication = None if last_replication is None or pulp_last_updated > last_replication else last_replication`.

pulpcore/app/tasks/replica.py
```diff
@@ -93,4 +95,6 @@ def replicate_distributions(server_pk):
             distro_names.append(distro["name"])
 
         replicator.remove_missing(distro_names)
+
+    server.set_last_replication_timestamp()
```
mdellweg, line 99: Setting this here is correct, but it must be set to a timestamp _before_ we started enumerating stuff.
I'd suggest taking the tasks started_at time.

pulpcore/app/models/replica.py
mdellweg: Maybe not. See another comment below.

pulpcore/app/replica.py
lubosmj: The function takes care of converting and reading `distro.get("no_content_change_since")`. I found it better to extract it as a separate function. The function can return `None` or `pased_date`, which properly fits our needs. Do you insist on this?

pulpcore/app/serializers/replica.py
lubosmj: We do not need to nullify it, do we? Because of the way we are saving this field on the model, we can just check `if upstream_pulp.last_replication < upstream_pulp.pulp_last_updated`. I know, we partially bypass the rationale behind the purpose of `pulp_last_updated` since it can has smaller timestamp than `last_replication`.

pulpcore/app/replica.py
mdellweg: No, I just thought we test it for None here anyway. And we don't even attempt the conversion if the other one was None.

pulpcore/app/serializers/replica.py
mdellweg: Yes that's exactly my observation. Either we nullify, or we just use the information we have when we interpret the field. (The latter probably being safer.)
We kind of don't bypass the `pulp_last_updated` if we keep `last_replication` from the user.
But then again, the user does not need to know that `last_replication` is actually a database field. I think we can get away with that.

pulpcore/app/serializers/publication.py
lubosmj: > Is a distribution with SERVE_FROM_PUBLICATION being attached to a repository_version even valid?

Conceptually, it might seem a bit wrong. However, I am sure plugin writers can enable both. At the moment, the handler deals only with a scenario where `(repo_version and not distribution.SERVE_FROM_PUBLICATION)`. So, I followed a similar logic here.

> I wonder if a SerializerMethod field would have been easier.

Not a bad idea. [`SerializerMethodField`](https://www.django-rest-framework.org/api-guide/fields/#serializermethodfield) suits our needs.

> Also what is the best way a Plugin can subclass this behaviour with better knowledge?

I think we should go with the SerializerMethod field and plugins can decide to either override the whole logic or do nothing. I do not see any place where we should be able to inject the additional plugin's logic.

mdellweg: > I do not see any place where we should be able to inject the additional plugin's logic.

Sure, me neither. Until that one day...

Honestly, with adding more known cases, we may arrive at a point where a plugin may actually know more than pulpcore. So hard to say for sure.

pulpcore/app/replica.py
lubosmj: The problem here is that we still want to continue syncing if the upstream-pulp object was updated. I would leave this part untouched.

dkliban (approved):

pulpcore/app/tasks/replica.py
```diff
@@ -67,7 +67,8 @@ def replicate_distributions(server_pk):
     for config in pulp_plugin_configs():
         if config.replicator_classes:
             for replicator_class in config.replicator_classes:
-                supported_replicators.append(replicator_class(ctx, task_group, tls_settings))
+                replicator = replicator_class(ctx, task_group, tls_settings, server)
```
mdellweg, line 70: I'm having minor issues reading this code.
Would you mind renaming this local variable to `upstream_pulp`?

mdellweg (approved):
Also let's approve this.

gerrod3 (changes requested):
Am I missing something, where is `no_content_change_since` on the pulp_file distribution? How is the optimization occurring and the test passing?

lubosmj:
> Am I missing something, where is `no_content_change_since` on the pulp_file distribution? How is the optimization occurring and the test passing?

There is no need to declare `no_content_change_since` on the file distribution serializer once again if it is declared in the generic `DistributionSerializer.Meta.fields`. The fields are re-used, see https://github.com/pulp/pulpcore/blob/2cc822acc0671fcac32c6324755fd310b2c4c53f/pulp_file/app/serializers.py#L138 and https://github.com/pulp/pulpcore/blob/2cc822acc0671fcac32c6324755fd310b2c4c53f/pulpcore/app/serializers/publication.py#L244


This is what I am seeing in a re-deployed oci-env:
```
(venv3) [lmjachky@lmjachky-thinkpadt14gen4 oci_env]$ http :5001/pulp/default/api/v3/distributions/file/file/01914baa-a300-77c6-bbf1-9a91ac8214b7/
{
    "base_path": "test",
    "base_url": "http://localhost:5001/pulp/content/default/test/",
    "content_guard": null,
    "hidden": false,
    "name": "test",
    "no_content_change_since": null,
    "publication": null,
    "pulp_created": "2024-08-13T12:15:59.747659Z",
    "pulp_href": "/pulp/default/api/v3/distributions/file/file/01914baa-a300-77c6-bbf1-9a91ac8214b7/",
    "pulp_labels": {},
    "pulp_last_updated": "2024-08-13T12:15:59.747704Z",
    "repository": "/pulp/default/api/v3/repositories/file/file/01914baa-574a-77e2-8f05-5a8aaeb175f8/"
}


(venv3) [lmjachky@lmjachky-thinkpadt14gen4 oci_env]$ http PATCH :5001/pulp/default/api/v3/distributions/file/file/01914baa-a300-77c6-bbf1-9a91ac8214b7/ publication=/pulp/default/api/v3/publications/file/file/01914baa-950d-7467-a602-7e9825cb3403/ repository=""
{
    "task": "/pulp/default/api/v3/tasks/01914bab-7e6d-7c69-a83b-a84d66bfd9a8/"
}


(venv3) [lmjachky@lmjachky-thinkpadt14gen4 oci_env]$ http :5001/pulp/default/api/v3/distributions/file/file/01914baa-a300-77c6-bbf1-9a91ac8214b7/
{
    "base_path": "test",
    "base_url": "http://localhost:5001/pulp/content/default/test/",
    "content_guard": null,
    "hidden": false,
    "name": "test",
    "no_content_change_since": "2024-08-13T12:16:55.979958Z",
    "publication": "/pulp/default/api/v3/publications/file/file/01914baa-950d-7467-a602-7e9825cb3403/",
    "pulp_created": "2024-08-13T12:15:59.747659Z",
    "pulp_href": "/pulp/default/api/v3/distributions/file/file/01914baa-a300-77c6-bbf1-9a91ac8214b7/",
    "pulp_labels": {},
    "pulp_last_updated": "2024-08-13T12:16:55.979958Z",
    "repository": null
}
```

lubosmj:
The field was introduced in a separate PR: https://github.com/pulp/pulpcore/pull/5704.

gerrod3 (approved):
`no_content_change_since` was added in a separate PR. This looks good.

PR #5662, opened 2024-08-01
Fix docs workflow

mdellweg:
Python version 3.11 seems to be required.

[noissue]

### DIFF
--- a/.github/workflows/all-docs.yml
+++ b/.github/workflows/all-docs.yml
@@ -1,10 +1,8 @@
 ---
-name: Publish staging-docs.pulpproject.org
+name: "Publish pulpproject.org"
 on:
   schedule:
-    # * is a special character in YAML so you have to quote this string
-    # runs at 3:00 UTC daily on weekdays 1 to 5 (no weekends)
-    - cron: '00 3 * * 1-5'
+    - cron: "00 3 * * 1-5"
   workflow_dispatch:
 
 defaults:
@@ -12,7 +10,7 @@ defaults:
     working-directory: "pulpcore"
 
 jobs:
-  build-staging-docs:
+  build-docs:
     runs-on: "ubuntu-latest"
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/all-docs.yml
```diff
@@ -1,42 +1,40 @@
 ---
-name: Publish staging-docs.pulpproject.org
+name: "Publish docs.pulpproject.org"
 on:
   schedule:
-    # * is a special character in YAML so you have to quote this string
-    # runs at 3:00 UTC daily on weekdays 1 to 5 (no weekends)
```
mdellweg, line 6: One should quote all strings if one was forced to use yaml.

.github/workflows/scripts/publish_all_docs.sh
```diff
@@ -20,9 +20,9 @@ export WORKSPACE=$PWD
...
 
-mkdir -p ../staging-docs.pulpproject.org
-tar -xvf staging-docs.pulpproject.org.tar --directory ../staging-docs.pulpproject.org
-pushd ../staging-docs.pulpproject.org
+mkdir -p ../docs.pulpproject.org
+tar -xvf docs.pulpproject.org.tar --directory ../docs.pulpproject.org
+pushd ../docs.pulpproject.org
 
 # publish to pulpproject.org
 rsync -avzh --delete site/ doc_builder_staging_pulp_core@docs.pulpproject.org:/var/www/docs.pulpproject.org/staging_pulp_core/
```
mdellweg, line 28: Just removing the "staging" from this line is probably not going to work, right?

pedro-psb (changes requested):
The current domain doesnt have the `docs.` prefix.
s/docs.pulpproject.org/pulpproject.org

.github/workflows/scripts/publish_all_docs.sh
mdellweg: Fixing the wf is more important atm.

pedro-psb (approved):

ggainey (approved):

lubosmj (approved):
Formally approving.

PR #5665, opened 2024-08-02
Fix urlize for pulp api hrefs

### LINKED ISSUE #5664: urlize_quoted_hrefs() still broken on v3 API browser
**Version**
3.56.1

**Describe the bug**
Seems this was fixed for some interactive pages as part of 3.56.0 -- #5563
When browsing the API UI, seeing the same behavior - ie.: https://pulp_server.fqdn/pulp/api/v3/

**To Reproduce**
Browse the API UI and you'll see none of the links are clickable. Instead of printing out the HREF html, it's escaping HTML tags.

**Expected behavior**
Clickable links

mdellweg:
Instead of carrying a copy of the whole api template, we just overwrite the branding related blocks. Also we monkeypatch django's urlize to know about pulp api hrefs.

fixes #5664


ggainey (approved):
I am a BIG fan of relying on the existing library as much as we can, and only overriding for our specific need. Some basic smoke-testing shows the API pages work well with this change, and atempting some XSS-injection fails, so looks good to me.

dkliban (approved):

Issue #5525, opened 2024-06-25
Users with no permissions can upload Artifacts to Pulp


lubosmj:
**Describe the bug**
Users with no permissions can upload artifacts to Pulp using the `/pulp/api/v3/artifacts/` endpoint. We advocate for restricting the access to this endpoint. Only admins should be able to upload raw artifacts to Pulp.

**To Reproduce**
```
pulp user create --username user --password krokodyl1 --email user@test.com
echo "wow" > testhello.txt
http --form POST :5001/pulp/api/v3/artifacts/ file@./testhello.txt -a user:krokodyl1
```
```
HTTP/1.1 201 Created
Access-Control-Expose-Headers: Correlation-ID
Allow: GET, POST, HEAD, OPTIONS
Connection: keep-alive
Content-Length: 679
Content-Type: application/json
Correlation-ID: 1ff96fae07da4d05b8ed8feaa21f9c5c
Cross-Origin-Opener-Policy: same-origin
Date: Tue, 25 Jun 2024 13:02:31 GMT
Location: /pulp/api/v3/artifacts/01904f7d-a0c3-72cb-bb93-1342a39f7cb4/
Referrer-Policy: same-origin
Server: nginx/1.22.1
Vary: Accept
X-Content-Type-Options: nosniff
X-Frame-Options: DENY

{
    "file": "artifact/f4/0cd21f276e47d533371afce1778447e858eb5c9c0c0ed61c65f5c5d57caf63",
    "md5": null,
    "pulp_created": "2024-06-25T13:02:31.620056Z",
    "pulp_href": "/pulp/api/v3/artifacts/01904f7d-a0c3-72cb-bb93-1342a39f7cb4/",
    "pulp_last_updated": "2024-06-25T13:02:31.620076Z",
    "sha1": null,
    "sha224": "8428b309ef4017df9b29d26a81771d320d3bff61404c9d0296609d16",
    "sha256": "f40cd21f276e47d533371afce1778447e858eb5c9c0c0ed61c65f5c5d57caf63",
    "sha384": "1c06d0e58db38d4d250db07a15f6a8ef40263dc25f95c0890f1ee0ddd396c2d1dca17dccefcba4a6df4a9d99327fd8e2",
    "sha512": "d7cc8802241aa6927a82ddfa7386977858a1e3f31c72db260935b89c33cf26e92939a3a15b0bf80adc5ecf6b9b63ef09843f4e0631ba8b5841996e26eae5e92b",
    "size": 4
[...]

Issue #3362, opened 2022-10-26
As an admin I can delete a Domain and all its related objects


ipanova:
**Is your feature request related to a problem? Please describe.**
Once a domain becomes unusable ( for various reasons e.g. simply not needed anymore) the admin should be able to delete that Domain and all the objects that are domain specific like repos, remotes, etc with it.

**Additional context**
Add any other context about the feature request here.

gerrod3:
It might make sense to also have a Domain clean endpoint that deletes all the items within a domain so that it restores it back to a clean state. This way you could call it even on the default domain when domains are turned off to reset your Pulp instance without dropping the DB/restarting everything.

mdellweg:
After deleting all that, calling the orhan cleanup codepath is in order. (Not in a new task though).

PR #5669, opened 2024-08-06
Restrict non-admin users from uploading artifacts

git-hyagi:
fixes: #5525

### DIFF
--- a/pulp_file/tests/functional/api/test_domains.py
+++ b/pulp_file/tests/functional/api/test_domains.py
@@ -123,12 +123,9 @@ def test_artifact_upload(
     }
 
     # Show that duplicate artifacts can be uploaded into different domains
-    dup_artifact = gen_object_with_cleanup(pulpcore_bindings.ArtifactsApi, filename)
+    dup_artifact = pulpcore_bindings.ArtifactsApi.create(filename, pulp_domain="default")
     assert "default/api/v3/" in dup_artifact.pulp_href
     assert dup_artifact.sha256 == second_artifact.sha256
-
-    # Delete second artifact so domain can be deleted
-    pulpcore_bindings.ArtifactsApi.delete(second_artifact_href)
 
 
 @pytest.mark.parallel
--- a/pulpcore/app/viewsets/content.py
+++ b/pulpcore/app/viewsets/content.py
@@ -3,10 +3,11 @@ from gettext import gettext as _
 from django.conf import settings
 from django.db import models
 from django_filters import NumberFilter
-from rest_framework import mixins, permissions, status
+from rest_framework import mixins, status
 from rest_framework.response import Response
 
 from pulpcore.filters import BaseFilterSet
+from pulpcore.app.loggers import deprecation_logger
 from pulpcore.app.models import Artifact, Content, PublishedMetadata, SigningService
 from pulpcore.app.serializers import (
     ArtifactSerializer,
@@ -73,12 +74,29 @@ class ArtifactViewSet(
     queryset = Artifact.objects.all()
     serializer_class = ArtifactSerializer
     filterset_class = ArtifactFilter
-    permission_classes = (permissions.IsAuthenticated,)
 
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
+            {
+                "action": ["create", "list", "retrieve"],
+                "principal": "admin",
+                "effect": "allow",
+            },
+        ],
+    }
+
+    # Deleting artifacts is a risky operation and will be removed in a future release.
+    # However, for compatibility reasons, it is still possible to execute the DELETE
+    # request by overriding the DEFAULT_ACCESS_POLICY.
     def destroy(self, request, pk):
         """
         Remove Artifact only if it is not associated with any Content.
         """
+        msg = _(
+            "destroy is deprecated. Deleting artifacts is a dangerous operation, "
+            "use orphan cleanup instead."
+        )
+        deprecation_logger.warning(msg)
         try:
             return super().destroy(request, pk)
         except models.ProtectedError:
--- a/pulpcore/tests/functional/api/test_crd_artifacts.py
+++ b/pulpcore/tests/functional/api/test_crd_artifacts.py
@@ -31,16 +31,9 @@ def _do_upload_valid_attrs(artifact_api, file, data):
     assert read_artifact.md5 is None
     for key, val in artifact.to_dict().items():
         assert getattr(read_artifact, key) == val
-    # Delete the Artifact
-    artifact_api.delete(read_artifact.pulp_href)
-    with pytest.raises(ApiException) as e:
-        artifact_api.read(read_artifact.pulp_href)
-
-    assert e.value.status == 404
 
 
-@pytest.mark.parallel
-def test_upload_valid_attrs(pulpcore_bindings, pulpcore_random_file):
+def test_upload_valid_attrs(pulpcore_bindings, pulpcore_random_file, monitor_task):
     """Upload a file, and provide valid attributes.
 
     For each possible combination of ``sha256`` and ``size`` (including
@@ -49,19 +42,24 @@ def test_upload_valid_attrs(pulpcore_bindings, pulpcore_random_file):
     1. Upload a file with the chosen combination of attributes.
     2. Verify that an artifact has been created, and that it has valid
        attributes.
-    3. Delete the artifact, and verify that its attributes are
-       inaccessible.
     """
     file_attrs = {"sha256": pulpcore_random_file["digest"], "size": pulpcore_random_file["size"]}
     for i in range(len(file_attrs) + 1):
         for keys in itertools.combinations(file_attrs, i):
             data = {key: file_attrs[key] for key in keys}
+            # before running the test with another file attribute we need to first
+            # remove the previous created artifact because the file content itself
+            # will be the same (the artifact sha256 sum will not change by modifying
+            # the file attrs)
+            monitor_task(
+                pulpcore_bindings.OrphansCleanupApi.cleanup({"orphan_protection_time": 0}).task
+            )
             _do_upload_valid_attrs(
                 pulpcore_bindings.ArtifactsApi, pulpcore_random_file["name"], data
             )
 
 
-def test_upload_empty_file(delete_orphans_pre, pulpcore_bindings, tmp_path):
+def test_upload_empty_file(pulpcore_bindings, tmp_path, monitor_task):
     """Upload an empty file.
 
     For each possible combination of ``sha256`` and ``size`` (including
@@ -70,8 +68,6 @@ def test_upload_empty_file(delete_orphans_pre, pulpcore_bindings, tmp_path):
     1. Upload a file with the chosen combination of attributes.
     2. Verify that an artifact has been created, and that it has valid
        attributes.
-    3. Delete the artifact, and verify that its attributes are
-       inaccessible.
     """
     file = tmp_path / str(uuid.uuid4())
     file.touch()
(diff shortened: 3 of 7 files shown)


pulpcore/app/viewsets/content.py
```diff
@@ -73,7 +73,27 @@ class ArtifactViewSet(
...
+            },
+            {
+                "action": ["list", "retrieve"],
+                "principal": "authenticated",
+                "effect": "allow",
+            },
+        ],
+        "creation_hooks": [
+            {
+                "function": "add_roles_for_object_creator",
```
mdellweg, line 92: Oh, please don't do that. Artifacts are shared. No one owns them.

pulpcore/app/viewsets/content.py
```diff
@@ -73,7 +73,21 @@ class ArtifactViewSet(
     queryset = Artifact.objects.all()
     serializer_class = ArtifactSerializer
     filterset_class = ArtifactFilter
-    permission_classes = (permissions.IsAuthenticated,)
+
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
+            {
+                "action": ["create", "destroy"],
```
mdellweg, line 80: Sorry, but I think I cannot just let go here:
Deleting artifacts is a dangerous operation. It should not only be not allowed, but actually not possible at all.

pulpcore/app/viewsets/content.py
```diff
@@ -73,7 +73,21 @@ class ArtifactViewSet(
...
+                "principal": "admin",
+                "effect": "allow",
+            },
+            {
+                "action": ["list", "retrieve"],
+                "principal": "authenticated",
+                "effect": "allow",
+            },
```
lubosmj, line 88: I think the idea is to completely lock this endpoint to admin users. Maybe, in the future, to even make this (and other admin endpoints) domain-enabled.

pulp_file/tests/functional/api/test_domains.py
```diff
@@ -127,8 +127,10 @@ def test_artifact_upload(
     assert "default/api/v3/" in dup_artifact.pulp_href
     assert dup_artifact.sha256 == second_artifact.sha256
 
-    # Delete second artifact so domain can be deleted
-    pulpcore_bindings.ArtifactsApi.delete(second_artifact_href)
+    # Delete the artifacts so domain can be deleted
+    body = {"orphan_protection_time": 0}
+    task = pulpcore_bindings.OrphansCleanupApi.cleanup(body, pulp_domain=domain.name).task
+    monitor_task(task)
```
mdellweg, line 133: Future idea (not this PR): Can we run orphan cleanup as part of the domain deletion task?

pulpcore/tests/functional/api/test_upload.py
```diff
@@ -72,11 +72,7 @@ def _upload_chunks(size, chunks, sha256, include_chunk_sha256=False):
         return upload, artifact
 
     yield _upload_chunks
-    for href in artifacts:
-        try:
-            pulpcore_bindings.ArtifactsApi.delete(href)
-        except ApiException:
-            pass
+    monitor_task(pulpcore_bindings.OrphansCleanupApi.cleanup({"orphan_protection_time": 0}).task)
```
mdellweg, line 75: This can break other tests running in parallel. Why do we even care? Just remove all of this.

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -22,21 +22,25 @@ def pulpcore_random_file(tmp_path):
...
-    assert read_artifact.md5 is None
-    for key, val in artifact.to_dict().items():
-        assert getattr(read_artifact, key) == val
-    # Delete the Artifact
-    artifact_api.delete(read_artifact.pulp_href)
-    with pytest.raises(ApiException) as e:
-        artifact_api.read(read_artifact.pulp_href)
-
-    assert e.value.status == 404
+    def upload_valid_attrs(monitor_task, pulpcore_bindings):
```
mdellweg, line 25: This sounds wrong. Is this function actually _used_ somewhere?

pulp_file/tests/functional/api/test_domains.py
ipanova: you can explicitly mention in this ticket to include content deletion as well https://github.com/pulp/pulpcore/issues/3362

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -21,26 +21,30 @@ def pulpcore_random_file(tmp_path):
...
-        assert getattr(read_artifact, key) == val
-    # Delete the Artifact
-    artifact_api.delete(read_artifact.pulp_href)
-    with pytest.raises(ApiException) as e:
-        artifact_api.read(read_artifact.pulp_href)
-
-    assert e.value.status == 404
+@pytest.fixture
+def upload_valid_attrs(allow_admin_destroy_artifact):
+    def _upload_valid_attrs(artifact_api, file, data):
```
mdellweg, line 26: Now that this is a fixture, we can ask for the `pulpcore-binding` fixture:
```suggestion
@pytest.fixture
def upload_valid_attrs(pulpcore_bindings, allow_admin_destroy_artifact):
    def _upload_valid_attrs(file, data):
```

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -157,7 +159,44 @@ def test_delete_artifact(pulpcore_bindings, pulpcore_random_file):
...
+
+    # destroy artifact api is not allowed, even for admins
+    with pytest.raises(ApiException) as e:
+        pulpcore_bindings.ArtifactsApi.delete(artifact.pulp_href)
+    assert e.value.status == 403
+
+    task = pulpcore_bindings.OrphansCleanupApi.cleanup({"orphan_protection_time": 0}).task
+    monitor_task(task)
     file_exists = os.path.exists(path_to_file)
     assert not file_exists
```
mdellweg, line 176: Instead of not running in parallel, why not just remove this piece from the test.
It's not testing the artifact api. It's testing orphan cleanup, which should really be done elsewhere.
I understand where you are coming from, but `pulpcore_bindings.ArtifactsApi.delete(artifact.pulp_href)` is exactly what we don't support so there is no use trying to keep the test for it.

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -21,26 +21,32 @@ def pulpcore_random_file(tmp_path):
...
-@pytest.mark.parallel
-def test_upload_valid_attrs(pulpcore_bindings, pulpcore_random_file):
+@pytest.fixture
+def upload_valid_attrs(monitor_task, pulpcore_bindings, pulp_domain_enabled):
+    def _upload_valid_attrs(file, data):
+        """Upload a file with the given attributes."""
+        if pulp_domain_enabled:
+            data["pulp_domain"] = "default"
+        artifact = pulpcore_bindings.ArtifactsApi.create(file, **data)
+        # assumes ALLOWED_CONTENT_CHECKSUMS does NOT contain "md5"
```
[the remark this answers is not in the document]
ipanova: What Matthias meant here is that you can rely on settings by accessing them and based on what they contain you can check for the absence of certain checksum algorithms rather than relying on an assumption comment.

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -146,7 +149,7 @@ def test_upload_mixed_attrs(pulpcore_bindings, pulpcore_random_file):
 
 
 @pytest.mark.parallel
-def test_delete_artifact(pulpcore_bindings, pulpcore_random_file):
+def test_delete_artifact(pulpcore_bindings, pulpcore_random_file, gen_user):
     """Delete an artifact, it is removed from the filesystem."""
```
ipanova, line 153: you need to update docstrings in multiple places

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -49,19 +42,24 @@ def test_upload_valid_attrs(pulpcore_bindings, pulpcore_random_file):
     1. Upload a file with the chosen combination of attributes.
     2. Verify that an artifact has been created, and that it has valid
        attributes.
-    3. Delete the artifact, and verify that its attributes are
-       inaccessible.
     """
     file_attrs = {"sha256": pulpcore_random_file["digest"], "size": pulpcore_random_file["size"]}
     for i in range(len(file_attrs) + 1):
         for keys in itertools.combinations(file_attrs, i):
```
ipanova, line 48: is there any value to test upload with different attribute combinations? I would just upload the artifact once with the provided file_attrs, this way there won't be any need to orphan delete the artifact in between.

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -70,15 +68,16 @@ def test_upload_empty_file(delete_orphans_pre, pulpcore_bindings, tmp_path):
...
     """
     file = tmp_path / str(uuid.uuid4())
     file.touch()
     empty_file = b""
     file_attrs = {"sha256": hashlib.sha256(empty_file).hexdigest(), "size": 0}
     for i in range(len(file_attrs) + 1):
         for keys in itertools.combinations(file_attrs, i):
+            monitor_task(
+                pulpcore_bindings.OrphansCleanupApi.cleanup({"orphan_protection_time": 0}).task
+            )
```
mdellweg, line 80: I would not care about cleaning anything up here. The Artifact api is meant to be handling duplicate artifacts transparently. And specifically "empty file" could lead to a collision with content, that cannot be cleaned up at this point, way to easily.
(I think desperately needing to start clean is a left over from when reuploading artifacts lead to errors.)

pulpcore/tests/functional/api/test_crd_artifacts.py
mdellweg: Well i think it is testing that you can upload the artifact with e.g. only providing the size, or only sha128 and sha244 or so.
And for exactly that reason we should not run cleanup inbetween, because Pulp _is_ supposed to find the existing artifact at least starting from the second attempt.

pulpcore/tests/functional/api/test_crd_artifacts.py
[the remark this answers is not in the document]
mdellweg: Also on a machine configured with MD5 this test will fail as a false positive.

ipanova (approved):

pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -49,19 +42,24 @@ def test_upload_valid_attrs(pulpcore_bindings, pulpcore_random_file):
...
             _do_upload_valid_attrs(
                 pulpcore_bindings.ArtifactsApi, pulpcore_random_file["name"], data
             )
 
 
-def test_upload_empty_file(delete_orphans_pre, pulpcore_bindings, tmp_path):
+def test_upload_empty_file(pulpcore_bindings, tmp_path, monitor_task):
```
lubosmj, line 62: What is the rationale behind this change? Could we revert it along with the `pulpcore_bindings.OrphansCleanupApi.cleanup({"orphan_protection_time": 0})` call?

lubosmj (changes requested):
Please, address my last comment before merging.

pulpcore/tests/functional/api/test_crd_artifacts.py
git-hyagi: hum... this test is failing without the orphan cleanup:
```
E           HTTP response body: {"non_field_errors":["Artifact with sha256 checksum of 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855' already exists."]}
```
Should we still address this error (when removing the orphan cleanup) in this PR or as [mdellweg mentioned](https://github.com/pulp/pulpcore/pull/5669#pullrequestreview-2242339791) "[..] _merge this now and rethink the testing in a second PR._"?

lubosmj: Is it failing with `delete_orphans_pre` too?

lubosmj: I see. Okay. Let's merge this!w

lubosmj (approved):

PR #5724, opened 2024-08-19
Improve documentation for the complex filtering

### LINKED ISSUE #5703: API Documentation does not explain the workings of the q parameter
When you browse through the API documentation you will notice the q parameter. It only mentions from the API documentation that it is a string, but further explanation lacks. See below image and the API docs for example https://pulpproject.org/pulp_rpm/restapi/#tag/Content:-Packages/operation/content_rpm_packages_list

image

recommend to improve the API docs so that the q parameter is explained or a reference is given to how it should work. 

The following reference explains the working and limits of the complex query you can do with the q param

https://pulpproject.org/pulpcore/docs/user/guides/use-complex-filters/

lubosmj:
closes #5703

### DIFF
--- a/CHANGES/5703.doc
+++ b/CHANGES/5703.doc
@@ -1,0 +1,1 @@
+Improved documentation for the complex filtering.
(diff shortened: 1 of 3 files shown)


lubosmj:
In api.json, the change looks like this:
```
                    {
                        "in": "query",
                        "name": "q",
                        "schema": {
                            "type": "string"
                        },
                        "description": "Filter results by using NOT, AND and OR operations on other filters"
                    },
```

ggainey (approved):

ipanova (approved):

PR #5730, opened 2024-08-20
Pulp-workers log now logs also domain name in which the task is excuted.

### LINKED ISSUE #5693: pulp-workers should log which domain a task was executed in
When a task id is referenced in the pulp-worker logs, that ID is not enough information to look up the task status from the REST API. The task id needs to be accompanied by a domain name also.

ipanova:
closes #5693

### DIFF
--- a/CHANGES/5693.feature
+++ b/CHANGES/5693.feature
@@ -1,0 +1,1 @@
+Pulp-workers now also log domain name in which the task is executed.
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -63,8 +63,9 @@ def _execute_task(task):
     # Store the task id in the context for `Task.current()`.
     current_task.set(task)
     task.set_running()
+    domain = get_domain()
     try:
-        _logger.info(_("Starting task %s"), task.pk)
+        _logger.info(_("Starting task %s in domain: %s"), task.pk, domain.name)
 
         # Execute task
         module_name, function_name = task.name.rsplit(".", 1)
@@ -81,12 +82,12 @@ def _execute_task(task):
     except Exception:
         exc_type, exc, tb = sys.exc_info()
         task.set_failed(exc, tb)
-        _logger.info(_("Task %s failed (%s)"), task.pk, exc)
+        _logger.info(_("Task %s failed (%s) in domain: %s"), task.pk, exc, domain.name)
         _logger.info("\n".join(traceback.format_list(traceback.extract_tb(tb))))
         _send_task_notification(task)
     else:
         task.set_completed()
-        _logger.info(_("Task completed %s"), task.pk)
+        _logger.info(_("Task completed %s in domain: %s"), task.pk, domain.name)
         _send_task_notification(task)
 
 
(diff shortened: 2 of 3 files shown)


pulpcore/tasking/tasks.py
```diff
@@ -64,7 +64,7 @@ def _execute_task(task):
     current_task.set(task)
     task.set_running()
     try:
-        _logger.info(_("Starting task %s"), task.pk)
+        _logger.info(_("Starting task %s in domain: %s"), task.pk, task.pulp_domain.name)
```
decko, line 67: @ipanova do you think this could trigger another query into the db, to retrieve the pulp_domain?
If so, maybe adding a `select_related("pulp_domain")` [here](https://github.com/pulp/pulpcore/blob/b197ffbcaae4b96de1e2da31f30b2b21b05ea49c/pulpcore/tasking/worker.py#L296-L298) could help with it.

daviddavis:
I think ideally this change would only manifest if the Pulp installation has `DOMAIN_ENABLED` set to True. Would that be difficult to do?

CHANGES/5693.feature
```diff
@@ -0,0 +1 @@
+Pulp-workers log now logs also domain name in which the task is excuted.
```
daviddavis, line 1:
```suggestion
Pulp-workers now also log domain name in which the task is executed.
```

pulpcore/tasking/worker.py
```diff
@@ -232,7 +241,11 @@ def identify_unblocked_tasks(self):
         taken_exclusive_resources = set()
         taken_shared_resources = set()
         # When batching this query, be sure to use "pulp_created" as a cursor
-        for task in Task.objects.filter(state__in=TASK_INCOMPLETE_STATES).order_by("pulp_created"):
+        for task in (
+            Task.objects.filter(state__in=TASK_INCOMPLETE_STATES)
+            .order_by("pulp_created")
+            .select_related("pulp_domain")
```
mdellweg, line 247: Would that work too?
```suggestion
            .select_related("pulp_domain__name")
```

pulpcore/tasking/tasks.py
```diff
@@ -241,18 +242,19 @@ def cancel_task(task_id):
     Raises:
         rest_framework.exceptions.NotFound: If a task with given task_id does not exist
     """
-    task = Task.objects.get(pk=task_id)
+    task = Task.objects.select_related("pulp_domain").filter(pk=task_id).first()
```
mdellweg, line 245: instead of `get`, `first` can return `None`. We should keep `get`.

pulpcore/tasking/worker.py
```diff
@@ -362,7 +388,11 @@ def supervise_task(self, task):
                 if connection.connection in r:
                     connection.connection.execute("SELECT 1")
                     if self.cancel_task:
-                        _logger.info(_("Received signal to cancel current task %s."), task.pk)
+                        _logger.info(
+                            _("Received signal to cancel current task %s in domain: %s."),
```
mdellweg, line 392: Not your fault, but translating log messages, can create a support burden i my opinion.

pulpcore/tasking/worker.py
ipanova: It would, if `name` was a ForeignKey.

mdellweg: So no? That's fine.

ipanova:
> I think ideally this change would only manifest if the Pulp installation has `DOMAIN_ENABLED` set to True. Would that be difficult to do?

Should not be too difficult while checking on the setting, however I am not sure I fancy the concept of making the log output dependent. I'll see what others say.

mdellweg (approved):

decko (approved):

Issue #5496, opened 2024-06-20
Docs: Fix rst leftover errors


pedro-psb:
After migrating the docs to markdown, there is still some fair amount of documents with rst leftover errors.

Some common pattern observed:
- admonitions blocks without indentation
- a lot of backticked words (`` `some-word` ``) were internal links before. That's a bit hard to distinguish now and is the least automatable IHO.
- stuff with `:something:`, which I think are different kinds of links
  ```bash
   grep -rE ':[a-z]+:' staging_docs/
  ```
- stuff with `{something}`. Mostly `{class}`, which reference generated code API doc:
  ```bash
  grep -rE '\{[a-z_\-]+\}`' staging_docs/
  ```

AndersBennedsgaard:
I don't know if it is relevant here, since I'm a newcomer for Pulp, but there are also a lot of broken links, and links to incorrect places. A couple of examples:
- [Pulp-python Overview](https://pulpproject.org/pulp_python/docs/user/guides/00-overview/) references the REST API for the `pulp_python` in https://pulpproject.org/pulp_python/docs/user/guides/00-overview/restapi.html, which doesn't exist.
- [The Backup section in Pulp operator Configure and Run Backup/Restore](https://pulpproject.org/pulp-operator/docs/admin/guides/backup_and_restore/01-config_and_run/#backup) references the [PulpBackup CR](https://pulpproject.org/pulp_operator/backup/#pulpbackupspec), which doesn't exist, so it redirects to the table of contents

You hit many other examples if you just browse through the website. It might make sense to have some automation for detecting broken links. We use [tcort/markdown-link-check](https://github.com/tcort/markdown-link-check) successfully as validation in a CI pipeline

---

Another note: the documentation needs some sort of sorting/weighting between pages. It doesn't make sense that the Pulp CLI how-to guides go "Configure (advanced) -> Configure (basic) -> Install":

![image](https://github.com/pulp/pulpcore/assets/34585076/b0d37f21-da49-47eb-acfe-d540145b1889)

lubosmj:
This site is also completely broken: https://pulpproject.org/pulpcore/docs/dev/guides/plugin-walkthrough/.

PR #5731, opened 2024-08-20
General admonition fixes on docs

pedro-psb:
This does:
- On dev rbac pages
  - re-linked links that were not converted and were in the format `` `some-link` ``
  - added some code reference links with `[path.to.module][]`. Will work more on this on other PR
- Fixed general admonitions with wrong indent level.

Fixes some of https://github.com/pulp/pulpcore/issues/5496

### DIFF
--- a/docs/admin/reference/settings.md
+++ b/docs/admin/reference/settings.md
@@ -233,9 +233,9 @@ Store cached responses from the content app into Redis. This setting improves th
 of the content app under heavy load for similar requests. Defaults to `False`.
 
 !!! note
-The entire response is not stored in the cache. Only the location of the file needed to
-recreate the response is stored. This reduces database queries and allows for many
-responses to be stored inside the cache.
+    The entire response is not stored in the cache. Only the location of the file needed to
+    recreate the response is stored. This reduces database queries and allows for many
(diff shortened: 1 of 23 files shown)


docs/admin/reference/settings.md
```diff
@@ -233,9 +233,9 @@ Store cached responses from the content app into Redis. This setting improves th
 of the content app under heavy load for similar requests. Defaults to `False`.
 
 !!! note
-The entire response is not stored in the cache. Only the location of the file needed to
-recreate the response is stored. This reduces database queries and allows for many
-responses to be stored inside the cache.
+    The entire response is not stored in the cache. Only the location of the file needed to
+    recreate the response is stored. This reduces database queries and allows for many
+    responses to be stored inside the cache.
```
mdellweg, line 238: There's an addition newline in some other cases.
Is there _one_ right way to do it?

pedro-psb: Both work. From a style perspective, we haven't defined the right one.
It's something we can do when we start using a markdown linter/formatter.

mdellweg (approved):

ggainey (approved):

PR #5732, opened 2024-08-20
General code-api reference fixup on docs

pedro-psb:
This does:

- moved the code-api reference to the developer persona.
- do some old sphinx-files cleanup.
- updated code cross-reference syntax to either `[mod.obj][]` or `` `mod.obj` `` on docstrings and docs.
- remove weird quotation marks before markdown lists

Fixes some of https://github.com/pulp/pulpcore/issues/5496


docs/dev/guides/release_process.md
```diff
@@ -1,47 +1,37 @@
...
+- pulpcore
+    - 3.23 (galaxyNG/4.7)
+    - 3.22 (katello/4.9)
+    - 3.21 (katello/4.7, galaxyNG/4.6, RHUI/4.3)
+    - 3.18 (katello/4.5)
+    - 3.16 (katello/4.3)
+- pulp_file
+    - 1.12 (katello/4.9)
+    - 1.11 (katello/4.7)
+    - 1.10 (katello/4.3. 4.5)
```
mdellweg, line 37: Good thing we put a date in here. But also it _is_ actually quire outdated, no?

mdellweg (dismissed):

docs/dev/guides/release_process.md
pedro-psb: Yes. I tried to not hung up on the details in this fix pass.

But pulp-docs could read supported branches from `template_config`, as its more likely to get updated.
Maybe adding these related projects names as comments.

ggainey: We can/should remove this - I added it when users (including us!) had NO WAY of knowing what "supported" meant...

ggainey (approved):
I confess to not having looked at Every Single Change - but a random sample of a dozen files "looks right to me" :)

mdellweg (approved):

Issue #5087, opened 2024-02-28
An upstream_pulp's pulp_label_select should allow processing as an OR instead of AND


PaulUrwin:
**Is your feature request related to a problem? Please describe.**
We have various categories of content in different repositories and we label their distributions differently as such.  We would like the `upstream_pulp`'s `pulp_label_select` to allow us to provide a list of labels to match against on an OR basis, so that we can say we would like to replicate category A, B and C distributions.  As it stands, if we listed the labels for category A, B and C we would get no match as the distributions only have the label for the category of content they provide and the `pulp_label_select` is processed as an AND.

**Describe the solution you'd like**
Per a matrix chat, I believe you have an example for an and-or-not parser (q filter) that could be used to implement this flexibility when processing the `pulp_label_select` value.

dkliban:
We are going to add a new 'q_select' parameter to the UpstreamPulp API. This filter is going to use the same syntax as the q filter available everywhere else in the API[0]. 

The value for this q_select would be something like  `pulp_label_select="foo" OR pulp_label_select="bar"`

[0] https://pulpproject.org/pulpcore/docs/user/guides/use-complex-filters/

PR #5733, opened 2024-08-21
Add ability for complex filtering on replicating UpstreamPulps

gerrod3:
fixes: #5087

Also added tests for `pulp_label_select` replications since I couldn't find any (maybe they are there and I'm bad at grep).


pulpcore/app/models/replica.py
```diff
@@ -27,6 +27,7 @@ class UpstreamPulp(BaseModel, AutoAddObjPermsMixin):
     password = EncryptedTextField(null=True)
 
     pulp_label_select = models.TextField(null=True)
+    q_select = models.TextField(null=True)
```
mdellweg, line 30: Do we want to add this filter, or replace the existing one?
Technically it would not need to be mutually exclusive. But I think I would go for deprecating the old one.
Going forward, we can even write a safe migration that translates `label_select' filters to 'q_select'.

pulpcore/app/serializers/replica.py
```diff
@@ -93,6 +101,21 @@ class UpstreamPulpSerializer(ModelSerializer, HiddenFieldsMixin):
         read_only=True,
     )
 
+    def validate_q_select(self, value):
+        """Ensure we have a valid q_select expression."""
+        from pulpcore.app.viewsets import DistributionFilter
+
+        ExpressionFilterField(filter=DistributionFilter(), required=False).clean(value)
```
mdellweg, line 108: Nice!
Could we have used `DistributionFilter().q.clean(value)` or something like that instead? (And not changing the ExpressionFilterField init then?)

pulpcore/filters.py
```diff
@@ -219,7 +219,9 @@ def evaluate(self, qs):
             return self.exprs[0].evaluate(qs).union(*[expr.evaluate(qs) for expr in self.exprs[1:]])
 
     def __init__(self, *args, **kwargs):
-        self.filterset = kwargs.pop("filter").parent
+        self.filterset = kwargs.pop("filter")
+        if isinstance(self.filterset, ExpressionFilter):
+            self.filterset = self.filterset.parent
```
mdellweg, line 224: This wasn't quite as obvious to me. See comment above.

pulpcore/tests/functional/api/test_replication.py
```diff
@@ -465,3 +465,149 @@ def test_replicate_rbac(
...
+        "q_select": "pulp_label_select='upstream=1' OR pulp_label_select='upstream=2'",
+    }
+    upstream = gen_object_with_cleanup(
+        pulpcore_bindings.UpstreamPulpsApi, upstream_body, pulp_domain=dest_domain.name
+    )
+    # Run the replicate task and assert that two repos got synced
+    response = pulpcore_bindings.UpstreamPulpsApi.replicate(upstream.pulp_href)
+    monitor_task_group(response.task_group)
+    result = pulpcore_bindings.DistributionsApi.list(pulp_domain=dest_domain.name)
+    assert result.count == 2
```
mdellweg, line 592: Could we check for something like `set((item.name for item result.results)) == {"1", "2"}`?
We would need to name the distributions accordingly, but since we exclusively own the domain this test runs on that should not be an issue.

mdellweg (commented):
I don't recall if we decided to replace the label select filter, but since the new one provides a superset of the functionality, i'd rather vote for it.

pulpcore/app/serializers/replica.py
gerrod3: Ok, got it to work without having to modify `ExpressionFilterField`.

pulpcore/app/models/replica.py
gerrod3: It definitely could replace the old one. I would like more thoughts on this feature before I go deprecating anything.

pulpcore/app/serializers/replica.py
```diff
@@ -93,6 +100,21 @@ class UpstreamPulpSerializer(ModelSerializer, HiddenFieldsMixin):
         read_only=True,
     )
 
+    def validate_q_select(self, value):
+        """Ensure we have a valid q_select expression."""
+        from pulpcore.app.viewsets import DistributionFilter
+
+        DistributionFilter().filters["q"].field.clean(value)
```
mdellweg, line 107: That sounds like much better targeted.
:+1:

gerrod3:
Ok, I've "removed" `pulp_label_select` by migrating it to `q_select` and removing the field from the serializer. This should make it ZDU safe, and I left a little note that we can properly remove the field from the model in the next breaking change release.

mdellweg (commented):
About the ZDU:
There's a case we may not be able to cover, I think. During the update, it is possible to configure an UpstreamPulp object using the new filter with a new api worker. Then, calling replicate on an old api worker, which will not realize there is an issue and trigger a task claiming to be compatible with the old worker code. Now if an old worker will pick up the task, it will not be able to apply the `q_filter` that may not even be translatable into `label_select` anymore. This could lead to an unfiltered replicate of every thing.

Can we make old workers fail when reading the `label_select_filter`, to limit the damage?

gerrod3:
> Can we make old workers fail when reading the `label_select_filter`, to limit the damage?

No, I don't think I can. Worst case the user can rerun replication if this happens after the upgrade is complete.

mdellweg:
> No, I don't think I can. Worst case the user can rerun replication if this happens after the upgrade is complete.

Yes, I agree. The thing I'm worried about is that accidentally replicating without the filter can bust all the resource limits of the replica. And it can expose unwanted content in a certain Geo. But as you say, there is no good way to solve that, we should probably just leave a warning somewhere.

gerrod3:
> But as you say, there is no good way to solve that, we should probably just leave a warning somewhere.

I added a warning in the changelog.

pulpcore/app/migrations/0123_upstreampulp_q_select.py
```diff
@@ -0,0 +1,44 @@
...
+    migration. Old tasks will still be able to access the field and new tasks will use the new
+    q_select field. Any new UpstreamPulps will only use q_select (label_select is being removed
+    from the serializer), so on the next breaking change release we can add a migration to remove
+    the pulp_label_select field.
+    """
+    UpstreamPulp = apps.get_model("core", "UpstreamPulp")
+    batch = []
+    for upstream in UpstreamPulp.objects.filter(pulp_label_select__isnull=False).iterator():
+        if upstream.pulp_label_select:
+            upstream.q_select = f"pulp_label_select='{upstream.pulp_label_select}'"
```
mdellweg, line 18: I think, we may need to think about proper quotation here.

pulpcore/tests/functional/api/test_replication.py
```diff
@@ -465,3 +464,148 @@ def test_replicate_rbac(
...
+def test_replicate_with_complex_q_select(
+    domain_factory,
+    populate_upstream,
+    bindings_cfg,
+    pulpcore_bindings,
+    monitor_task_group,
+    pulp_settings,
+    gen_object_with_cleanup,
+):
+    """Test complex q_select replication."""
```
mdellweg, line 569: Can we add a test where we need to quote the value inside of the quoted `pulp_label_select` filter string?

pulpcore/app/migrations/0123_upstreampulp_q_select.py
gerrod3: I think `f"pulp_label_select={upstream.pulp_label_select!r}"` should be sufficient, correct?

pulpcore/tests/functional/api/test_replication.py
gerrod3: Do you have an example of what one mine look like?

mdellweg: One thing i learned is, that users are creative...
Consider:
`pulp file repository label set --repository test1 --key "asdf" --value "'asdf/+*&\""`

gerrod3: Yeah maybe we shouldn't allow quotes in labels. The ambiguity of it is very annoying and what value does a user gain by being able to use them? 

I've been struggling with trying to get your example to work with the q=pulp_label_select filter and I'm not sure it is possible with how query expressions work. If you can get it to work then I'll add a test for it, else I'll just leave this as an edge case that hopefully users will be smart enough to avoid.

mdellweg (approved):

pulpcore/tests/functional/api/test_replication.py
mdellweg: So you are saying we cannot even select for these values today?
Can you add some sanitation for the label values to make sure we will never get there?
Maybe it's a separate PR, and I'm happy to have a migration fail in case a user has done that already.

PR #5761, opened 2024-09-02
Re-enable Storage Metrics emmiter.

### LINKED ISSUE #5762: Re-enable Storage Metrics

**Additional context**

decko:
Revert "Disable the Storage Metrics emmiter for now."
This reverts commit 7029ee4de07af7a9a17bce9073102cdcf1ef1b91.

Closes #5762


pulpcore/app/models/content.py
```diff
@@ -390,6 +391,14 @@ def touch(self):
         """Update timestamp_of_interest."""
         self.save(update_fields=["timestamp_of_interest"])
 
+    @hook(AFTER_SAVE)
+    @hook(AFTER_DELETE)
+    def emmit_telemetry(self):
+        from pulpcore.plugin.tasking import dispatch
+        from pulpcore.app.tasks import telemetry
+
+        dispatch(telemetry.emmit_disk_space_usage_telemetry, args=(self.pulp_domain.pk,))
```
lubosmj, line 400: Seeing this again, would not it make much more sense to (1) store the value of used disk space in a dedicated table and (2) schedule a task that will periodically update this value once in a while? All the burden of running the complex query will be deferred to the task and API workers will just emit the value stored in the table as it is.

```
gauge.set(DomainSize.objects.get(domain=domain["pk"]), {"domain_name": domain["pulp_domain__name"]})
```

Does it make sense to you?

decko: I don't think so @lubosmj. I'm dispatching a task from here to avoid any sort of lock or slow down response when creating an Artifact.

Also, I'm just sending the metered value when a new Artifact hits the base. I can't see a reason to store it since I'm not using that value all the time.

lubosmj: Your plan is to dispatch a task each time an artifact is saved or deleted, correct? During syncing or recycling (orphan cleanup), this might result into hundreds of tasks being dispatched within the very small time frame, potentially blocking other useful tasks from being executed. I doubt this has some performance benefits.

decko: Nope. It'll only schedule a new task if there's no other scheduled for the next 5 minutes.

decko: Change of plans @lubosmj

decko: The current idea is to run this metrics just like we run the unblocked tasks metrics. Within the heartbeat of the worker.

pulpcore/tasking/worker.py
```diff
@@ -456,6 +466,27 @@ def handle_available_tasks(self):
                 keep_looping = True
                 self.supervise_task(task)
 
+    def record_domain_space_usage(self):
+        if os.getenv("PULP_OTEL_ENABLED", "").lower() != "true" and not settings.DOMAIN_ENABLED:
+            return
+
+        with contextlib.suppress(AdvisoryLockError), PGAdvisoryLock(STORAGE_METRICS_LOCK):
```
bmbouter, line 473: With the way this reads to me (maybe I'm wrong here), every worker could acquire the lock with every beat causing this code to be run a lot more than expected. I imagined this working where the advisory lock would attempt to be acquired here (called from beat()) but then held until the process exits and the fact the lock is acquired held in memory as a variable and then checked (no need to re-acquire if I already got it last time, I still have it).

To acquire the lock in a long-lived way I think you'll need to not use the context manager.

mdellweg: Holding a lock forever is probably a bad idea. (Holding a lock for the time of a several hours sync task has not yet demonstrated to be a real problem, but it gives me a lot to think about.) Also we agreed to not do any version of leader election in the workers.
So, as we can now establish, this whole telemetry collecting and posting meters (I'm not talking about recording events, just scraping measures from the database here) business turns out to be more than a one off. Also it is sufficiently separate from all the other pulp operations to be moved into it's own process. This should reduce the risk of this whole endeavour by an order of magnitude.

decko: @bmbouter Nope man. Only one worker can acquire the lock and it will only run the query if the current time is late than the last_heartbeat + heartbeat_period.

decko: I like the idea of having a different process dealing with it @mdellweg. Yet, I believe we need to discuss two points:
1- We still gonna need to have telemetry on some Pulp parts. Are we ok on telemetry code living between those two spaces? A "telemetry worker" and the Pulp code itself?
2- Currently we're running out of resources on services team and I can't see this new pulp-process happening in the short term.

How can we figure out a solution for this?

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,45 @@
...
+
+resource = Resource(attributes={"service.name": "pulp-worker"})
+
+metric_reader = PeriodicExportingMetricReader(
+    exporter, export_interval_millis=3000, export_timeout_millis=3000
+)
+provider = MeterProvider(metric_readers=[metric_reader], resource=resource)
+
+
+def emit_domain_space_usage_metric():
```
mdellweg, line 27: The only need for more than one such task would be the typical "Hourly, Daily, Weekly" schedules.
```suggestion
def otel_metrics():
```

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,45 @@
...
+
+    meter = provider.get_meter(__name__)
+    space_usage_gauge = meter.create_gauge(
+        name="space_usage",
+        description="The total space usage per domain.",
+        unit="bytes",
+    )
+
+    space_utilization_per_domain = Artifact.objects.values("pulp_domain__name").annotate(
```
lubosmj, line 36:
```suggestion
    space_utilization_per_domain = Artifact.objects.values("pulp_domain__name").only("size").annotate(
```

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,45 @@
...
+        total_size=Sum("size", default=0)
+    )
+
+    # We're using the same gauge with different attributes for each domain space usage
+    for domain in space_utilization_per_domain:
+        space_usage_gauge.set(domain["total_size"], {"domain_name": domain["pulp_domain__name"]})
+
+    metric_reader.collect()
+    metric_reader.force_flush()
```
lubosmj, line 45: Calling `force_flush` appears redundant to me.
```
    def force_flush(self, timeout_millis: float = 10_000) -> bool:
        super().force_flush(timeout_millis=timeout_millis)  # ABC class calls self.collect()
        self._exporter.force_flush(timeout_millis=timeout_millis)  # noop for PeriodicExportingMetricReader
        return True
```
https://opentelemetry-python.readthedocs.io/en/latest/sdk/metrics.export.html#opentelemetry.sdk.metrics.export.MetricReader.force_flush
https://opentelemetry-python.readthedocs.io/en/latest/sdk/metrics.export.html#opentelemetry.sdk.metrics.export.PeriodicExportingMetricReader.force_flush
https://opentelemetry-python.readthedocs.io/en/latest/exporter/otlp/otlp.html#opentelemetry.exporter.otlp.proto.http.metric_exporter.OTLPMetricExporter.force_flush

pulpcore/app/tasks/telemetry.py
decko: I don't get it. The name make it clear what the function does. The `telemetry` module could even receive the code of the task metrics we have in `worker.py`.
I see no reason to change the name here.

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,44 @@
...
+from opentelemetry.sdk.metrics import MeterProvider
+
+from opentelemetry.exporter.otlp.proto.http.metric_exporter import OTLPMetricExporter
+from opentelemetry.sdk.metrics.export import PeriodicExportingMetricReader
+from opentelemetry.sdk.resources import Resource
+
+# This configuration is needed since the worker thread is not using the opentelemetry
+# instrumentation agent to run the task code.
+
+OTLP_EXPORTER_ENDPOINT = os.getenv("OTEL_EXPORTER_OTLP_ENDPOINT", "http://localhost:4318/")
```
lubosmj, line 15: This can be removed. It is not used anywhere.

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,44 @@
...
+
+exporter = OTLPMetricExporter()
+
+resource = Resource(attributes={"service.name": "pulp-worker"})
+
+metric_reader = PeriodicExportingMetricReader(
+    exporter, export_interval_millis=3000, export_timeout_millis=3000
+)
```
lubosmj, line 23: The timeout and interval values are retrieved from environment variables by default. Is it worth setting them if we enforce the exporting?

CHANGES/5762.feature
```diff
@@ -0,0 +1,2 @@
+Re-enable the Domain Storage metric emmiter and adds a feature flag to it.
+This is an experimental feature and can change without prior notice.
```
lubosmj, line 2: This changelog does not reflect the reality. We do not have a feature flag for the storage metrics, do we?

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,44 @@
...
+        unit="bytes",
+    )
+
+    space_utilization_per_domain = Artifact.objects.values("pulp_domain__name").annotate(
+        total_size=Sum("size", default=0)
+    )
+
+    # We're using the same gauge with different attributes for each domain space usage
+    for domain in space_utilization_per_domain:
+        space_usage_gauge.set(domain["total_size"], {"domain_name": domain["pulp_domain__name"]})
```
lubosmj, line 42: Is it worth incorporating the pulp-href here? It was part of the metric's attributes in the past: https://github.com/pulp/pulpcore/blob/f5c439c90a8c4043e978bbd09e0c6974151b2c05/pulpcore/app/util.py#L643-L649.

pulpcore/app/tasks/telemetry.py
decko: Nope. need to remove it from here also

pulpcore/app/tasks/telemetry.py
decko: I don't know @lubosmj. I can't see the usefulness of it on the metric itself. Also, to use the `get_prn` function we need a Domain instance or the Domain pulp_href. At that point of our code we don't have that.

pulpcore/app/tasks/telemetry.py
decko: Thanks man. I've removed it.

CHANGES/5762.feature
decko: Sorry. My first implementation had it

CHANGES/5762.feature
```diff
@@ -0,0 +1,2 @@
+Re-enable and refactor the Domain Storage metric emiter.
+This is an experimental feature and can change without prior notice.
```
lubosmj, line 2:
```suggestion
Re-enabled and refactord the Domain Storage metric emiter.
```

lubosmj (commented):
From the functional's perspective, this is working great!

image

```
otel-collector    | InstrumentationScope pulpcore.app.tasks.telemetry 
otel-collector    | Metric #0
otel-collector    | Descriptor:
otel-collector    |      -> Name: space_usage
otel-collector    |      -> Description: The total space usage per domain.
otel-collector    |      -> Unit: bytes
otel-collector    |      -> DataType: Gauge
otel-collector    | NumberDataPoints #0
otel-collector    | Data point attributes:
otel-collector    |      -> domain_name: Str(default)
otel-collector    | StartTimestamp: 1970-01-01 00:00:00 +0000 UTC
otel-collector    | Timestamp: 2024-10-08 15:56:56.717505952 +0000 UTC
otel-collector    | Value: 368237
```

The feature is resilient to worker's restarts and the value is correctly reported. Please, address this comment from @mdellweg: https://github.com/pulp/pulpcore/pull/5761/files#r1790281943. I am going to approve this PR then.

PR #2280, opened 2022-03-01
Cherrypick changelogs for releases

dralley:
[noissue]

PR #5765, opened 2024-09-03
Fix memory spike on repository delete

dralley:
Django handles cascade deletes itself (as opposed to letting the database handle it) and in doing so it pulls a lot of data out, resulting in a memory spike on repository deletion.

The only solution is to write a custom cascade delete that deletes the objects at the bottom of the cascade tree more incrementally.

closes https://github.com/pulp/pulpcore/pull/2280

### DIFF
--- a/pulpcore/app/models/publication.py
+++ b/pulpcore/app/models/publication.py
@@ -151,33 +151,29 @@ class Publication(MasterModel):
         Notes:
             Deletes the Task.created_resource when complete is False.
         """
+        # invalidate cache
+        if settings.CACHE_ENABLED:
+            # Find any publications being served directly
+            base_paths = self.distribution_set.values_list("base_path", flat=True)
+            # Find any publications being served indirectly by auto-distribute feature
+            # It's possible for errors to occur before any publication has been completed,
+            # so we need to handle the case when no Publication exists.
+            try:
+                versions = self.repository.versions.all()
+                pubs = Publication.objects.filter(repository_version__in=versions, complete=True)
+                publication = pubs.latest("repository_version", "pulp_created")
+                if self.pk == publication.pk:
+                    base_paths |= self.repository.distributions.values_list("base_path", flat=True)
+            except Publication.DoesNotExist:
+                pass
+
+            # Invalidate cache for all distributions serving this publication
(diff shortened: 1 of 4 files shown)


dkliban:
@dralley what do you think about improving this test? https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/api/using_plugin/test_crud_repos.py#L17

I was thinking we could create a publication in the test. Then assert that it got deleted. Maybe also make some assertions on RepositoryVersions? The assertions for repository versions would have to use the RepositoryVersion model from pulpcore and not the REST API.

dralley:
@dkliban Instead of implementing a new task type, perhaps we should be overriding the base `.delete()` functions?

Should I be doing the same for repository version delete (which already does override `.delete()` for squashing functionality)?

dkliban:
@dralley I agree that it would be better to change the implementation of repository.delete() instead of introducing a new task.

dkliban (approved):

dralley:
@dkliban Do you think we ought to do any large-scale testing before enabling the option by default?  e.g. the COPR testing box or services staging

dkliban (changes requested):
Let's get rid of the setting and just always delete with the new implementation.

dralley:
@dkliban TBH I'm still partial to leaving the setting around for a release or two personally.  I'm not totally against removing it, just risk averse.

dkliban (approved):
Thank you!

pulpcore/app/models/publication.py
```diff
@@ -151,33 +151,29 @@ def delete(self, **kwargs):
...
+                if self.pk == publication.pk:
+                    base_paths |= self.repository.distributions.values_list("base_path", flat=True)
+            except Publication.DoesNotExist:
+                pass
+
+            # Invalidate cache for all distributions serving this publication
+            if base_paths:
+                Cache().delete(base_key=cache_key(base_paths))
 
+        with transaction.atomic():
```
dralley, line 174: The goal of this is just to minimize the scope of the transaction to the part that is making modifications.  All of the above is just queries, there's no need to hold the transaction open.

PR #5779, opened 2024-09-09
Add domain and task argument support to TaskSchedule.

### LINKED ISSUE #5778: Add support for arguments to a TaskSchedule instance
When scheduling a task you're not able to set a domain to which this task would run.
Also, you're not able to set the arguments for the schedule task to run.

Being able to define those things would allow you to schedule and filter scheduled tasks for a given domain
and set arguments that may be needed to run the task would help on its usage.

decko:
Closes #5778


docs/dev/learn/other/task-scheduling.md
```diff
@@ -8,7 +8,7 @@
 Pulp supports scheduling of tasks. Scheduled tasks will be dispatched shortly after their
 `next_dispatch` time, and be rescheduled one `dispatch_interval` after that, if the latter is
 set. By specifying the `dispatch_interval` as `time_delta(days=1)` you can expect the task
-dispatch to stabily happen at same time every day. Until the last task of the same schedule enters a
+dispatch to stably happen at same time every day. Until the last task of the same schedule enters a
```
lubosmj, line 11:
```suggestion
dispatch to stably happen at the same time every day. Until the last task of the same schedule enters a
```

pulpcore/app/models/task.py
```diff
@@ -371,6 +372,8 @@ class TaskSchedule(BaseModel):
     dispatch_interval = models.DurationField(null=True)
     task_name = models.TextField()
     last_task = models.ForeignKey(Task, null=True, on_delete=models.SET_NULL)
+    pulp_domain = models.ForeignKey(Domain, null=True, on_delete=models.CASCADE)
+    task_args = models.TextField(null=True)
```
lubosmj, line 376: Should we use here `JSONField` instead: https://docs.djangoproject.com/en/4.2/ref/models/fields/#jsonfield? It appears to be much more safe than calling `ast`.

decko: Great idea. Gonna work on that.

PR #5787, opened 2024-09-11
Add pulp_labels to Content.

ggainey:


pulpcore/app/serializers/repository.py
```diff
@@ -444,8 +444,9 @@ class RepositoryAddRemoveContentSerializer(ModelSerializer, NestedHyperlinkedMod
     remove_content_units = serializers.ListField(
         help_text=_(
             "A list of content units to remove from the latest repository version. "
-            "You may also specify '*' as an entry to remove all content. This content is "
-            "removed before add_content_units are added."
+            "You may specify '*' as an entry to remove all content, or 'key=value' pairs "
+            "to remove content by-label. This content is removed before add_content_units "
+            "are added."
```
mdellweg, line 449: OK, I can see, that this is still unambiguous. But should we maybe go for a new `remove_q` field?
You may for example remove all files from a repository by their `relative_path__startswith`...
So the logic could start with `remove_q="pulp_label='foo'"` and later we add support for  `remove_q="pulp_label='foo' OR relative_path__startswith='/bar/'`.

mdellweg: Having said that, should we factor this part (advanced modify) into it's own PR? It would give us time to think this high level api design over.

Given that a user can filter content by labels, they can already compile the set of content hrefs locally to call modify with them.

ggainey: > Given that a user can filter content by labels, they can already compile the set of content hrefs locally to call modify with them.

Yes - but the user-experience of calling an API with a list of possibly thousands of HREFs, for one label, is...suboptimal. And "remove all the units with a given build id" is the actual requirement we're being asked to respond to, it's the point for doing this at all.

I can see benefits to a more-general Q filter - but I'm not sure the extra flexibility is worth the extra complication?

pulpcore/app/serializers/repository.py
```diff
@@ -474,19 +475,29 @@ def validate_add_content_units(self, value):
...
             if len(value) > 1:
                 raise serializers.ValidationError("Cannot supply content units and '*'.")
             else:
                 return ["*"]
-        else:
+
+        # Q: just one label? can label/hrefs be mixed? accept full label-operand-set?
+        # Current: multiple labels, mixed with hrefs, only '='
+        # "x=y" can happen multiple times, and means "all content with label x=y, pass k=v pair"
+        labels_specified = [s for s in value if "=" in s]
```
gerrod3, line 488: Technically you can have a label that is just a key, with no value. So we probably should assume anything that doesn't begin with a slash to be a proper label.

ggainey: Ooof, good point. Pushes us towards @mdellweg 's comment above, to have a `remove_q=...` construct.

pulpcore/app/serializers/repository.py
ggainey: Removing the label-support for remove-content for "future development"

mdellweg (approved):

dkliban (changes requested):
Thank you for this wonderful feature! Please add a test that uses the generic Content API to perform filtering my pulp_label.

PR #5801, opened 2024-09-16
Fixed logging issues with task state transitions

mdellweg:
This will log the task pk as well as the original state. Also it is now considered an OK situation if a cancelling task was attempted to be completed by it's worker.

### DIFF
--- a/pulpcore/app/models/task.py
+++ b/pulpcore/app/models/task.py
@@ -178,10 +178,6 @@ class Task(BaseModel, AutoAddObjPermsMixin):
         rows = Task.objects.filter(pk=self.pk, state=TASK_STATES.WAITING).update(
             state=TASK_STATES.RUNNING, started_at=timezone.now()
         )
-        if rows != 1:
-            raise RuntimeError(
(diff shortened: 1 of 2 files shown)


pulpcore/app/models/task.py
```diff
@@ -242,6 +247,12 @@ def set_failed(self, exc, tb):
             del self.finished_at
         with suppress(AttributeError):
             del self.error
+        if rows != 1:
+            raise RuntimeError(
+                _("Attempt to set not running task {} to failed from '{}'.").format(
```
ggainey, line 252: Doesn't the same as #L216 apply here to " and not canceling"?

mdellweg:
```python
rows = Task.objects.filter(pk=self.pk, state=TASK_STATES.RUNNING).update(
    state=TASK_STATES.FAILED,
    finished_at=timezone.now(),
    error=exception_to_dict(exc, tb_str),
)
```
Currently, at least, a canceling task can only transition to canceled.

ggainey (approved):

PR #5807, opened 2024-09-17
Document the environment variables used in bindings_cfg()

maggu:
I think these ought to be documented. It would greatly have simplified getting the tests running for us.

### DIFF
--- a/CHANGES/5807.doc
+++ b/CHANGES/5807.doc
@@ -1,0 +1,1 @@
+Document the environment variables used in the test fixture `bindings_cfg`.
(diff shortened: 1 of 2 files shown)


mdellweg (dismissed):

maggu:
@mdellweg  Thank for the quick approval. Sorry for breaking it by adding a towncrier entry. 😃

CHANGES/5807.doc
```diff
@@ -0,0 +1 @@
+Document the environment variables used in bindings_cfg()
```
mdellweg, line 1: Let's be more specific here (Sprinkling a "test" there should help the audience to filter it):
```suggestion
Document the environment variables used to hydrate the test fixture `bindings_cfg`.
```

maggu: Sounds good. Thanks. Might as well adjust the commit message then as well. Why "to hydrate" though? Maybe just use "by" or "in" instead?

mdellweg: That'd be fine too. Important is that a reader only be _using_ Pulp knows this is about tests and does not concern them.

maggu: 👍 Updated now.

mdellweg (approved):

PR #5809, opened 2024-09-17
Fix URL encoding issue in HttpDownloader to handle special characters

### LINKED ISSUE #5686: pulp_file fails to sync file repos literally containing URL encoding strings like %3a in a file name

hstct:
closes #5686

### DIFF
--- a/CHANGES/5686.bugfix
+++ b/CHANGES/5686.bugfix
@@ -1,0 +1,1 @@
+Implemented a more robust URL encoding mechanism ensuring that special characters are processed correctly.
--- a/pulp_file/pytest_plugin.py
+++ b/pulp_file/pytest_plugin.py
@@ -102,6 +102,16 @@ def basic_manifest_path(write_3_iso_file_fixture_data_factory):
 
 
 @pytest.fixture
+def encoded_manifest_path(file_fixtures_root):
+    file_fixtures_root.joinpath("encoded").mkdir()
+    file1 = generate_iso(file_fixtures_root.joinpath("encoded/long-name-%253a-encoded.iso"))
+    file2 = generate_iso(file_fixtures_root.joinpath("encoded/another-%25-encoded.iso"))
+    file3 = generate_iso(file_fixtures_root.joinpath("encoded/more-%3C-encoded.iso"))
+    generate_manifest(file_fixtures_root.joinpath("encoded/PULP_MANIFEST"), [file1, file2, file3])
+    return "/encoded/PULP_MANIFEST"
+
+
+@pytest.fixture
 def copy_manifest_only_factory(file_fixtures_root):
     def _copy_manifest_only(name):
         file_fixtures_root.joinpath(f"{name}-manifest").mkdir()
--- a/pulp_file/tests/functional/api/test_sync.py
+++ b/pulp_file/tests/functional/api/test_sync.py
@@ -124,6 +124,21 @@ def test_duplicate_file_sync(
 
 
 @pytest.mark.parallel
+def test_encoded_file_name(
+    file_repo, file_bindings, encoded_manifest_path, file_remote_factory, monitor_task
+):
+    remote = file_remote_factory(manifest_path=encoded_manifest_path, policy="immediate")
+    body = RepositorySyncURL(remote=remote.pulp_href)
+    monitor_task(file_bindings.RepositoriesFileApi.sync(file_repo.pulp_href, body).task)
+    file_repo = file_bindings.RepositoriesFileApi.read(file_repo.pulp_href)
+
+    version = file_bindings.RepositoriesFileVersionsApi.read(file_repo.latest_version_href)
+    assert version.content_summary.present["file.file"]["count"] == 3
+    assert version.content_summary.added["file.file"]["count"] == 3
+    assert file_repo.latest_version_href.endswith("/1/")
+
+
+@pytest.mark.parallel
 def test_filepath_includes_commas(
     file_bindings,
     file_repo,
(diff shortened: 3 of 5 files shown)


lubosmj:
Please, write a changelog message named after the issue number, like so: CHANGES/5686.bugfix.

https://pulpproject.org/pulpcore/docs/dev/tutorials/quickstart/?h=changelog#3-update-the-changelog-entry

hstct:
> Please, write a changelog message named after the issue number, like so: CHANGES/5686.bugfix.
> 
> https://pulpproject.org/pulpcore/docs/dev/tutorials/quickstart/?h=changelog#3-update-the-changelog-entry

Ah yea, sorry it was late yesterday xD

hstct:
For additional context:

The issue we encountered wasn’t just that the filename contained `%3a`, which was then decoded to `:`. The file couldn’t be found on the server, regardless of whether `%3`a or `:` was used. The main problem was that the `%` in the filename was double-encoded, resulting in `%253a` as the actual filename.

hstct:
Added some tests for this as well.

hstct:
idk why the test is failing on github :cry:

git-hyagi:
hi @hstct 

I was checking this error, and I think it might be related to the way aiohttp server is parsing the request (but it is just a guess, I could not make it work yet):
https://github.com/pulp/pulpcore/blob/e360a133406a2e4e57305369ff27337ecf6af804/pulpcore/pytest_plugin.py#L322-L328

I will do more tests and let you know if I find something.

git-hyagi:
I managed to make the test work by modifying the [pulpcore/pytest_plugin.py](https://github.com/pulp/pulpcore/blob/e360a133406a2e4e57305369ff27337ecf6af804/pulpcore/pytest_plugin.py#L321-L328) file with:
```
@pytest.fixture
def gen_fixture_server(gen_threaded_aiohttp_server):
    def _gen_fixture_server(fixtures_root, ssl_ctx):
        app = web.Application(middlewares=[_aiohttp_request_middleware])
        call_record = add_recording_route(app, fixtures_root)
        return gen_threaded_aiohttp_server(app, ssl_ctx, call_record)

    yield _gen_fixture_server


@web.middleware
async def _aiohttp_request_middleware(request,handler):
    unquoted_url = urllib.parse.unquote(request.url.human_repr())
    url = urllib.parse.urlparse(unquoted_url)
    response = await handler(request.clone(rel_url=url.path))
    return response
```

but I don't know the impact of this change in the other tests, or if this could be even considered a viable "solution".

hstct:
hi @git-hyagi thanks for the help!

As far as I can tell and this has been the whole problem with this issue in general is that it always is dependent on how the web server is serving the files, whether or not it will encode `%3a` to `%253a` or read it as an encoded `:`. So your approach really did help out here. I've been running tests for pulp_rpm, pulpcore, and pulp_deb locally and they seem to work fine. However I doubt that there are many tests that purposefully use problematic filenames. And that is what it is in the end, files having problematic names with special characters that get encoded when served online. So this is a workaround to make it more robust but I also can't tell if this is sufficient.

git-hyagi (approved):
Thank you for the contribution!
LGTM

gerrod3 (approved):
Nice testing!

quba42:
I am afraid we may have shot ourselves in the foot here. Our internal integration  tests revealed that while this fixes what we set out to fix, it broke the far more important use case of "download URL's with special characters in them".

I have just verified this using latest pulpcore main branch in oci-env. I tried syncing a fairly standard APT repo, which failed because Pulp tried to download `adduser_3.118%252Bdeb11u1_all.deb` instead of `adduser_3.118+deb11u1_all.deb`.

I suggest we revert this PR for now, until we can figure out a better approach. Unfortunately I fear there may be no good way to make the down-loader robust for both use cases.

PR #5812, opened 2024-09-18
Clarify zero-downtime upgrade docs

daviddavis:
According to Matrix discussion, it sounds like Pulp already supports zero-downtime upgrades. I'm updating the docs to be more clear about this.


mdellweg (approved):

mdellweg:
Thank you!

chris1984:
> According to Matrix discussion, it sounds like Pulp already supports zero-downtime upgrades. I'm updating the docs to be more clear about this.

Have you ever thought about using Katello at Microsoft?

Issue #5766, opened 2024-09-03
Add PRN field to all resources in the REST API


dkliban:
**Is your feature request related to a problem? Please describe.**
The values of `pulp_href` field contain the `API_ROOT`. By default `API_ROOT` is `/pulp/api/v3/` but can be anything an admins chooses at deployment time. This coupling with `API_ROOT` makes it difficult to re-root an existing Pulp instance. It also makes it difficult to introduce a v4 API. 

**Describe the solution you'd like**
Add a new 'prn' field that will be an identified of a resource.

The prn is already used for task resource locking. It has the following format:

`prn:<app_label>.<model_label>:<model_pk>`

PR #5813, opened 2024-09-18
Add PRN support

gerrod3:
fixes: #5766

### DIFF
--- a/docs/dev/learn/subclassing/serializers.md
+++ b/docs/dev/learn/subclassing/serializers.md
@@ -11,8 +11,10 @@ Most plugins will implement:
 : - serializer(s) for plugin specific content type(s), should be subclassed from one of
     NoArtifactContentSerializer, SingleArtifactContentSerializer, or
     MultipleArtifactContentSerializer, depending on the properties of the content type(s)
+  - serializer(s) for plugin specific repository(s), should be subclassed from RepositorySerializer
   - serializer(s) for plugin specific remote(s), should be subclassed from RemoteSerializer
-  - serializer(s) for plugin specific publisher(s), should be subclassed from PublisherSerializer
+  - serializer(s) for plugin specific publication(s), should be subclassed from PublicationSerializer
+  - serializer(s) for plugin specific distribution(s), should be subclassed from DistributionSerializer
 
 ## Adding Fields
 
@@ -33,6 +35,19 @@ class Meta:
       fields = SingleArtifactContentSerializer.Meta.fields + ('relative_path',)
       model = FileContent
 ```
+
+!!! note
+    
+    When inheriting from `pulpcore.plugin.serializers.ModelSerializer`, or one of its many subclasses,
+    you should include the fields of that serializer in your own. Important fields it provides include
+    `pulp_created`, `pulp_last_updated`, and `prn`.
+
+    See `pulpcore.plugin.serializers.ContentGuardSerializer` for an example:
+    ```python
+    class Meta:
+        model = models.ContentGuard
+        fields = ModelSerializer.Meta.fields + ("name", "description")
+    ```
 
 ### Help Text
 
--- a/pulpcore/app/util.py
+++ b/pulpcore/app/util.py
@@ -10,7 +10,9 @@ from urllib.parse import urlparse
 from contextlib import ExitStack
 from contextvars import ContextVar
 from datetime import timedelta
+from uuid import UUID
 
+from django.apps import apps
 from django.conf import settings
 from django.db import connection
 from django.db.models import Model, Sum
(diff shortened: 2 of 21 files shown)


pulpcore/app/viewsets/repository.py
```diff
@@ -192,18 +193,29 @@ class RepositoryVersionFilter(BaseFilterSet):
...
-                href_match = resolve(urlparse(uri).path).kwargs
-            except Resolver404:
-                href_match = {}
-            if "repository_pk" not in href_match or "number" not in href_match:
-                raise serializers.ValidationError(
-                    _("Invalid RepositoryVersion HREF: {}".format(uri))
-                )
-            repo_versions[href_match["repository_pk"]].append(int(href_match["number"]))
+            if uri.startswith("prn:"):
+                model, pk = resolve_prn(uri)
```
mdellweg, line 199: So a repository version PRN is not `prn:file.repository:<uuid>:<version>` ?

gerrod3: Nope, it is always `prn:<app_label>.<model_label>:<uuid>`, even for repository versions.

pulpcore/tests/functional/api/using_plugin/test_tasks.py
```diff
@@ -99,7 +99,7 @@ def test_filter_tasks_by_reserved_resources(setup_filter_fixture, pulpcore_bindi
     with pytest.raises(ApiException) as ctx:
         pulpcore_bindings.TasksApi.list(created_resources=created_resources)
 
-    assert ctx.value.status == 404
+    assert ctx.value.status == 400
```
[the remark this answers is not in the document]
gerrod3: It kind of makes sense, you couldn't find a badly named resource so 404 not found. But it's an improper input so it should probably return 400, which it now does since I changed the filter code.

mdellweg (dismissed):

ggainey (changes requested):
It would be good to document, for plugin-authors, something along the lines of "if you subclass from ModelSerializer directly" (which plugins do) "you'll want to add 'prn' to your exposed-fields"

pulpcore/app/util.py
```diff
@@ -120,12 +122,51 @@ def get_prn(instance=None, uri=None):
...
+
+    Raises:
+        rest_framework.exceptions.ValidationError: on invalid PRN.
+    """
+    if not prn.startswith("prn:"):
+        raise ValidationError(_("PRN must start with 'prn:': {}").format(prn))
+    split = prn.split(":")
+    if len(split) != 3:
+        raise ValidationError(
+            _("PRN must be of the form 'prn:app_label.model_label:pk': {}").format(prn)
```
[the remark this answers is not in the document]
gerrod3: No, they run `isidentifier()` on the label, so it is a slug: https://github.com/django/django/blob/39de2e97a06d0317973b280bc159ca6f89fc64e3/django/apps/config.py#L33-L38


https://docs.python.org/3/reference/lexical_analysis.html#identifiers

docs/dev/learn/subclassing/serializers.md
```diff
@@ -34,6 +36,13 @@ class Meta:
       model = FileContent
 ```
 
+!!! note
+    
+    When inheriting from `pulpcore.plugin.serializers.ModelSerializer`, or one of its many subclasses,
+    you should include the fields of that serializer in your own. Important fields it provides include
```
ggainey, line 42:
```suggestion
    you should include the fields of that serializer in your own. Important fields it provides include `pulp_created`, `pulp_last_updated`, and `prn`.
    
    See the `pulpcore.plugin.serializers.ContentGuardSerializer` for an example of how to accomplish this:
    ```
    class Meta:
        model = models.ContentGuard
        fields = ModelSerializer.Meta.fields + ("name", "description")
   ```
```


Picked a suggested-example basically at random - feel free to point to something "more canonical". But I thought having a specific good-practices how-to would be helpful.

docs/dev/learn/subclassing/serializers.md
```diff
@@ -34,6 +36,13 @@ class Meta:
...
+    `pulp_created`, `pulp_last_updated`, and `prn`.
```
ggainey, line 43: Can remove if previous-comment is acceptable.

ggainey (changes requested):
Some docs-wordsmithing - otherwise, looking very good!

docs/dev/learn/subclassing/serializers.md
gerrod3: Added an example to the note.

docs/dev/learn/subclassing/serializers.md
```diff
@@ -34,6 +36,19 @@ class Meta:
...
+
+    See `pulpcore.plugin.serializers.ContentGuardSerializer` for an example:
```
ggainey, line 45: :+1:

pulpcore/tests/functional/api/using_plugin/test_prn.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+@pytest.mark.parallel
+def test_prn_schema(pulp_openapi_schema):
+    """Test that PRN is a part of every serializer with a pulp_href."""
+    for name, schema in pulp_openapi_schema["components"]["schemas"].items():
+        if name.endswith("Response"):
+            if "pulp_href" in schema["properties"]:
+                assert "prn" in schema["properties"], name
+                assert schema["properties"]["prn"]["type"] == "string", name
```
ggainey, line 13: This fails, right now, if you have pulp_rpm installed with "rpm.DistributionTreeResponse". I'm sure there are more. Can we set this up to collect/log *all* the failures, and *then* fail the test?

ggainey (changes requested):

pulpcore/tests/functional/api/using_plugin/test_prn.py
gerrod3: Ahh, I didn't test with other plugins installed.

ggainey: A quick "manual" scan shows pulp_rpm is missing the base fields in `DistributionTreeSerializer` and `UpdateCollectionSerializer`, fwiw. Would be good to have the test give us a complete list for a given install, so a run with many-plugins would ferret out the list of serializers that need to be fixed.

gerrod3: Ok, I've updated the test to group them together. Can you try it out and see if it is correct?

ggainey: Ha! OK, it works - but it still only finds `DistributionTreeSerializer` - because `UpdateCollectionSerializer` doesn't expose `pulp_href`! 

```python
>       assert len(failed) == 0
E       AssertionError: assert 1 == 0
E        +  where 1 = len(['rpm.DistributionTreeResponse'])
```

Let me reset my env to include container and deb and let's see what we get, give me a bit.

ggainey: OK cool, this looks like it does pick up all the ones it finds:
```
E       AssertionError: assert 5 == 0
E        +  where 5 = len(['CollectionSummaryResponse', 'ansible.AnsibleDistributionResponse', 'ansible.AnsibleNamespaceMetadataResponse', 'ansible.CollectionVersionMarkResponse', 'rpm.DistributionTreeResponse'])
```

(That's from an env with `DEV_SOURCE_PATH=pulpcore:pulp_rpm:pulp_container:pulp_python:pulp_deb:pulp_ansible` , btw)

mdellweg: We used to have a set of tests in pulpcore that will run on the plugins ci. I'm not sure this is still a *thing*. But this sounds like a good candidate.

dralley (approved):

Issue #3634, opened 2023-03-06
content__in report wrong type in open api spec


mdellweg:
**Version**
Probably all.

**Describe the bug**
The `content__in` filters on Publications and RepsitoryVersions report `type: string` while they should have 
```json
"schema": {
  "type": "array",
  "items": {
    "type": "string"
  }
},
```

**To Reproduce**

**Expected behavior**

**Additional context**
Add any other context about the problem here.  Please provide links to any previous discussions via Discourse or Bugzilla.

PR #5818, opened 2024-09-20
Fix content__in filter

gerrod3:
fixes: #3634

### DIFF
--- a/pulpcore/app/viewsets/custom_filters.py
+++ b/pulpcore/app/viewsets/custom_filters.py
@@ -15,9 +15,9 @@ from django_filters import BaseInFilter, CharFilter, Filter
 from rest_framework import serializers
 from rest_framework.serializers import ValidationError as DRFValidationError
 
-from pulpcore.app.models import ContentArtifact, RepositoryVersion, Publication
+from pulpcore.app.models import Content, ContentArtifact, RepositoryVersion, Publication
 from pulpcore.app.viewsets import NamedModelViewSet
-from pulpcore.app.util import get_prn, get_domain_pk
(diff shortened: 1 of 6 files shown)


pulpcore/tests/functional/api/using_plugin/test_repo_versions.py
```diff
@@ -934,7 +934,7 @@ def test_content_in_repository_version_view(
     pulpcore_bindings,
     file_bindings,
     file_repository_factory,
-    file_random_content_unit,
+    file_content_unit_with_name_factory,
```
mdellweg, line 937: Yes!!!

pulpcore/app/viewsets/custom_filters.py
```diff
@@ -325,6 +325,47 @@ def filter(self, qs, value):
...
+
+        content_units_pks = set(content_units.keys())
+        existing_content_units = Content.objects.filter(pk__in=content_units_pks)
+        raise_for_unknown_content_units(existing_content_units, content_units)
+
+        return qs.with_content(content_units_pks)
+
+
+class WithContentInFilter(BaseInFilter, WithContentFilter):
+    """The multi-content variant of WithContentFilter."""
```
mdellweg, line 366: Is the `BaseInFilter` what fixes the openapi3 schema in this case?

mdellweg (commented):
The original bugreport was (only) about the openapi schema being wrong. Have you checked that this fixes it?

pulpcore/app/viewsets/custom_filters.py
gerrod3: Yes, I'll add a schema test for this.

mdellweg (approved):

PR #5820, opened 2024-09-20
Add an authenticate_header method to JSONHeaderRemoteAuthentication.

decko:
Closes #5891


lubosmj:
Did you correctly reference the issue number from the commit description?

decko:
> Did you correctly reference the issue number from the commit description?

Thanks for eye catch it man. I've corrected it in the commit message but not in the description. Fixed now.

lubosmj:
We want to get rid of the "no-issue" label. I guess you need to use a lowercase in "Closes" or update the plugin template: https://github.com/pulp/plugin_template/blob/c4cd2b8f981eabec717616c7c47036a922a86624/.ci/scripts/pr_labels.py#L19. Which path are you choosing?

pulpcore/app/authentication.py
```diff
@@ -72,3 +72,6 @@ def authenticate(self, request):
 
         _logger.debug(_("User {user} authenticated.").format(user=remote_user))
         return (user, None)
+
+    def authenticate_header(self, request):
+        return "JSONHeaderRemoteAuthentication"
```
dkliban, line 77: From reading the docs[0], it looks like this string should be a name of a header that a client can send to authenticate. However, when looking through docs outside of DRF, it seems like the string that is sent with the 401 is supposed describe what type of credentials should be sent[1,2]. I believe in this case this string just needs to be "Bearer"

[0] https://www.django-rest-framework.org/api-guide/authentication/#custom-authentication
[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate
[2] https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401

dkliban: However, 'Bearer' may not be an appropriate value in all cases. Perhaps we should not make this adjustment and just adjust our test to expect a 403?

decko: Also, this changes as we change the sequence of authentication classes.

PR #5832, opened 2024-09-24
Optimize ACS pipeline for runtime and memory

dralley:
So far, I can tell that it at least reduced runtimes by about 75%, but the memory use improvement is minimal.


ggainey (changes requested):
Can we get something in the changelog for this? Users who care will notice, and it's a good thing to be able to advertise.

ggainey (approved):
Great - thanks for the changelog entry!

Issue #5767, opened 2024-09-03
Run immediate tasks in the parent process of the worker


dkliban:
**Is your feature request related to a problem? Please describe.**
Immediate tasks, such as updates of repositories, remotes, and distributions can get stuck behind long running tasks such as syncs and publish.

**Describe the solution you'd like**
Have the parent process of the worker execute the immediate tasks in the parent process itself when all the resources for the task are available.

pedro-psb:
I'm taking this with @mdellweg mentorship

pedro-psb:
We designed a plan to implement this in three steps for smoother deliver and feedback cycles:
1) Prioritize immediate task
2) Add special handler to immediate task (run in foreground + timeout)
3) We'll discuss more (can't remember right now :sweat_smile: )

mdellweg:
4. Rewrite the pulpcore-worker in python async.
  - We are halfway there already with the select looping, and keeping the heartbeat loop responsive while performing lot's of immediate tasks would be made much easier with proper async.

PR #5841, opened 2024-09-25
Rrun immediate tasks on worker parent process

pedro-psb:
In busy tasking environments, "immediate" running tasks (which are only dispatched to workers due to resource locking requirements) can wait for very long behind long-running tasks, such as syncs and publishes.

This tries to solve this implementing the solution proposed in https://github.com/pulp/pulpcore/issues/5767, as discussed in the Pulp team.

---

Personal dev notes: https://hackmd.io/@pbrochad/BJYaIAZAC


pulpcore/tasking/worker.py
```diff
@@ -335,6 +340,18 @@ def iter_tasks(self):
...
+    def get_short_task(self):
+        task = (
+            Task.objects.filter(
+                state__in=TASK_STATES.WAITING, unblocked_at__isnull=False, immediate=True
+            )
+            .order_by("pulp_created")
+            .first()
+        )
+        with contextlib.suppress(AdvisoryLockError), task:
+            task.refresh_from_db()
```
gerrod3, line 352: `.first()` can return `None`. Will this not error trying to call a method on `None`?

pedro-psb: Yes, nice catch.
But anyway, this was a sketch of the approach I had in mind, which won't be used.
I force pushed with the ideas I discussed with @mdellweg yesterday.

pedro-psb:
(there is a refator and a impl commit)

pedro-psb:
Curious how the tests passed on the lowerbound but got stuck in the others scenarios :thinking:

pedro-psb:
Ready to review.

pulpcore/tasking/tasks.py
```diff
@@ -183,6 +183,7 @@ def dispatch(
                 parent_task=Task.current(),
                 reserved_resources_record=resources,
                 versions=versions,
+                immediate=immediate,
```
mdellweg, line 186: and deferred…

pulpcore/tasking/worker.py
```diff
@@ -445,11 +452,16 @@ def supervise_task(self, task):
         self.task = None
 
     def handle_available_tasks(self):
+        """Handle available tasks in a monitor/supervise cycle.
+
+        The cycle must spin until there are no more available tasks. Any flaw in detecting this
+        can lead to stale tasks in the database.
```
mdellweg, line 458: Actually there is not such thing as a stale task. The thing that can happen here is a stuck tasking system.

(TBF, it only needs one additional worker startup or one more dispatch to unstick it, but we should really avoid it.)

pedro-psb: What about:

```
Pick and supervise tasks until there are no more available tasks.

Failing to detect new available tasks can lead to a stuck state, as the workers
would go to sleep and wouldn't be able to know about the unhandled task until
an external wakeup event occurs (e.g., new worker startup or new task gets in).
```

I also could leave just the first line if you think that's too much detail for a docstring.

mdellweg: No that's fine. The mere fact that we discuss this comment shows that things are not obvious.

mdellweg (approved):

Issue #5843, opened 2024-09-27
OpenTelemetry metrics are reported in duplicate


dkliban:
`pulpcore` installs both `opentelemetry-instrumentation-wsgi` and `opentelemetry-instrumentation-django`. This results in very similar metrics being reported twice. We only want to use `opentelemetry-instrumentation-wsgi`. We should remove `opentelemetry-instrumentation-django` from requirements.txt.

PR #5846, opened 2024-09-27
Removes opentelemetry-instrumentation-django from requirements.txt

dkliban:
closes: #5843


bmbouter (approved):
Yes we are double reporting metrics. We need to remove this dependency.

PR #5859, opened 2024-10-03
Collect artifact size metrics after caching the content

### LINKED ISSUE #5817: Content consumption metrics ignore content when cached
**Version**
3.61.0

**Describe the bug**
From what i've been told the download bytes metrics emitted by the content app ignore content that is fetched after the cache is warmed up and includes that content.  

**To Reproduce**
With caching turned on on the content app, fetch a file 1000 times, you'd expect to see content metrics indicate 1000x size of file, instead it'd just be te first time

**Expected behavior**
Every fetch of the file is counted in the download metrics regardless of pulp-content caching

**Additional context**

lubosmj:
The caching machinery stores all headers from the response coming from content-handler. With this commit, we are introducing an proprietary header that contains the size of the served artifact or artifact available after the redirect, X-PULP-ARTIFACT-SIZE.

closes #5817

(cherry picked from commit e709de5c91629e4a77daf0160f8c8f4fd4cef66b)

### DIFF
--- a/CHANGES/5817.bugfix
+++ b/CHANGES/5817.bugfix
@@ -1,0 +1,1 @@
+Started collecting artifact size metrics even after caching the served content.
--- a/pulpcore/app/util.py
+++ b/pulpcore/app/util.py
@@ -1,6 +1,7 @@
 import hashlib
 import zlib
 import os
+import socket
 import tempfile
 import gnupg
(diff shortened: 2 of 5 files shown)


dkliban:
We are just going to do a 3.64 release.

PR #5885, opened 2024-10-08
Added PulpDefaultAccessPolicy

mdellweg:
This class serves both as a base class for Pulp access policies and as a reference implementation that will use the hardcoded default policies provided in the views implementations.


mdellweg:
This first commit is only half of the story. We still want to add yet another permission class to allow reading the ap from settings.

ggainey (approved):

PR #5891, opened 2024-10-09
Re-enable Storage Metrics emiter and refactor it.

### LINKED ISSUE #5762: Re-enable Storage Metrics

decko:
Closes #5762

### DIFF
--- a/CHANGES/5762.feature
+++ b/CHANGES/5762.feature
@@ -1,0 +1,1 @@
+Re-enabled and refactored the Domain Storage metric emitter.
--- a/pulpcore/app/tasks/telemetry.py
+++ b/pulpcore/app/tasks/telemetry.py
@@ -1,0 +1,39 @@
+from django.db.models import Sum
+
+from pulpcore.app.models import Artifact
+
+from opentelemetry.sdk.metrics import MeterProvider
+
+from opentelemetry.exporter.otlp.proto.http.metric_exporter import OTLPMetricExporter
+from opentelemetry.sdk.metrics.export import PeriodicExportingMetricReader
+from opentelemetry.sdk.resources import Resource
+
+
+def otel_metrics():
+
+    # This configuration is needed since the worker thread is not using the opentelemetry
+    # instrumentation agent to run the task code.
+
+    exporter = OTLPMetricExporter()
+
+    resource = Resource(attributes={"service.name": "pulp-worker"})
+
+    metric_reader = PeriodicExportingMetricReader(exporter)
+    provider = MeterProvider(metric_readers=[metric_reader], resource=resource)
+
+    meter = provider.get_meter(__name__)
+    space_usage_gauge = meter.create_gauge(
+        name="space_usage",
+        description="The total space usage per domain.",
+        unit="bytes",
+    )
+
+    space_usage_per_domain = Artifact.objects.values("pulp_domain__name").annotate(
+        total_size=Sum("size", default=0)
+    )
+
+    # We're using the same gauge with different attributes for each domain space usage
+    for domain in space_usage_per_domain:
+        space_usage_gauge.set(domain["total_size"], {"domain_name": domain["pulp_domain__name"]})
+
+    metric_reader.collect()
--- a/pulpcore/app/models/domain.py
+++ b/pulpcore/app/models/domain.py
@@ -72,14 +72,6 @@ class Domain(BaseModel, AutoAddObjPermsMixin):
             # Delete on by one to properly cleanup the storage.
             artifact.delete()
 
-    # Disabling Storage metrics until we find a solution to resource usage.
-    # https://github.com/pulp/pulpcore/issues/5468
-    # @hook(AFTER_CREATE)
-    # def _report_domain_disk_usage(self):
-    #     from pulpcore.app.util import DomainMetricsEmitterBuilder
-    #
-    #     DomainMetricsEmitterBuilder.build(self)
-
     class Meta:
         permissions = [
             ("manage_roles_domain", "Can manage role assignments on domain"),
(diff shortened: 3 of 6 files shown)


CHANGES/5762.feature
```diff
@@ -0,0 +1,3 @@
+
+Re-enabled and refactored the Domain Storage metric emiter.
+This is an experimental feature and can change without prior notice.
```
lubosmj, line 3:
```suggestion
Re-enabled and refactored the Domain Storage metric emiter.
```

decko: Are you suggesting to remove the experimental disclaimer?

lubosmj: Yes, because it is in tech preview, looking at the doc. And the empty new line should be removed too.

bmbouter:
@mdellweg @decko the task usage here really is a nice design. It's so little code.

lubosmj (dismissed):

mdellweg (changes requested):

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,39 @@
...
+
+exporter = OTLPMetricExporter()
+
+resource = Resource(attributes={"service.name": "pulp-worker"})
+
+metric_reader = PeriodicExportingMetricReader(exporter)
+provider = MeterProvider(metric_readers=[metric_reader], resource=resource)
+
+
+def emit_domain_space_usage_metric():
```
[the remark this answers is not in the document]
decko: > This name appears in the database. There is no way we can change it later. The current name is specific to it's first batch of work, yes. But it's not future proof. Also it suffers from homeopathic naming. "emit" and "usage" don't actually add reasonable information. And still this is the one recurring task to collect and send all the measures that need updates frequently (currently five minutes, but i'm not sold on that interval, so don't put it in the name).
We may think about adding a "otel_metrics_daily" later. You get the idea?

I got the idea, but the thing is that this function here will only emit this metric, no other. If we need to emit another metric on periodic basis then we need to add another function and so on. With this we can trace back which metric could be causing issues for example.

Also, we could have another periodic metric that would demand a different time schedule. With the `configure_periodic_telemetry` we can schedule different metrics emitter with different time schedules without issues.

Yet, the complete name of the function that will appear on the Tasks database will be `pulpcore.app.telemetry.emit_domain_space_usage_metric`, which makes sense to me. There's another term that could be more explicit than `space_usage`? I'm all ears.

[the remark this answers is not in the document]
mdellweg: That's fine. You can have as many functions as you wish. But you get one periodic task to call them. And this is nothing we can fix later. Task names are there for all eternity.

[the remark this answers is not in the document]
lubosmj: Let me clear up some misunderstandings here.

@decko's idea is to have separate periodic tasks for different metrics (that might come in the future), making them more isolated. Currently, he finds the task's name self-explanatory -- it emits exactly one metric related to current disk space usage. In contrast, `otel_metrics_N` is too vague. Is it initializing metrics, or emitting them? I agree with him.

However, I got @mdellweg's point of hiding as many details as possible inside the task's implementation and not trying to make any assumptions too early. The change he proposes could offer better opportunities if something backfires. To mitigate potential problems in the future, we should use as generic name as possible, because the tasks' names stay in the DB for eternity. This is a limitation we have to deal with. I agree with him.

After considering all the aspects, I am now leaning towards having a more generic task name, like: "emit_otel_metrics". We should not specify the interval in the name because it should be known from `dispatch_interval` defined at the task level.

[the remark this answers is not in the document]
decko: Folks, I've changed it following @mdellweg ideas.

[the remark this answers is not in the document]
mdellweg: > The change he proposes could offer better opportunities if something backfires.

That's a new aspect.
All right. Let's talk about failure scenarios. I love that.
So what can go wrong?
* Bad code: We don't plan for that, right?
* Failed to gather some information: Put a proper try-except block around every individual function and all the others still run.
* Failed to push the data to the collector: Is it really worth gathering more data at this point? The others will probably fail too. It will retry one interval later anyway. Also a failed task is a good thing at this time to report about the telemetry stack being down.

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,39 @@
...
+
+# This configuration is needed since the worker thread is not using the opentelemetry
+# instrumentation agent to run the task code.
+
+exporter = OTLPMetricExporter()
+
+resource = Resource(attributes={"service.name": "pulp-worker"})
+
+metric_reader = PeriodicExportingMetricReader(exporter)
+provider = MeterProvider(metric_readers=[metric_reader], resource=resource)
```
mdellweg, line 19: I think, all of this setup should happen inside of the task.
There's no need to have that overhead when this file is imported somewhere else.

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def otel_metrics():
+
+    meter = provider.get_meter(__name__)
```
mdellweg, line 24: This is where you get the name of the function that will appear as the label of the data?
That's a good indicator to have this run from within a function called by the task.

pulpcore/app/tasks/telemetry.py
```diff
@@ -0,0 +1,39 @@
...
+    space_usage_gauge = meter.create_gauge(
+        name="space_usage",
+        description="The total space usage per domain.",
+        unit="bytes",
+    )
+
+    space_utilization_per_domain = Artifact.objects.values("pulp_domain__name").annotate(
```
mdellweg, line 31: why calling it utilization in one place and usage in another?

mdellweg (approved):

PR #5910, opened 2024-10-14
fix: Signing scripts needs access to environment variables.

### LINKED ISSUE #5911: Signing scripts cannot access GNUPGHOME variable.
Signing scripts needs to access GNUPGHOME variable to be able to discover the keyring to look for keys.

Currently the only env vars passed are the ones defined on the signing service itself,

I propose repassing all the `os.environ` so it gives signing script access to read any env var.

**version**

pulpcore 3.49

rochacbruno:
A signing script usually needs to access `GNUPGHOME`, `HOME` and possibly other variables.

fixes #5911

### DIFF
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -7,6 +7,7 @@ from gettext import gettext as _
 import asyncio
 import datetime
 import json
+import os
 import tempfile
 import shutil
 import subprocess
@@ -783,7 +784,7 @@ class SigningService(BaseModel):
         }
         if env_vars:
             env.update(env_vars)
-        return env
+        return {**os.environ, **env}
 
     def sign(self, filename, env_vars=None):
         """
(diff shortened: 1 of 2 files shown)


ggainey:
Had some discussion around the implications here, at the core-team-mtg. Handing a signing script access to all the environment that Pulp itself has access to, feels like the/a "right thing". A signing-service already a) can only be put into place by an admin, and b) may need all kinds of things we don't currently think about, because the script can talk to all sorts of different Things to handle signing. In this context, allowing the signing-script to "see" All The Things is appropriate, and also gets Pulp itself out of the business of having to change code any time someone needs a new env-var (or set thereof) to be passed along.

pulpcore/app/models/content.py
```diff
@@ -783,7 +784,7 @@ def _env_variables(self, env_vars=None):
         }
         if env_vars:
             env.update(env_vars)
-        return env
+        return {**os.environ, **env}
```
ggainey, line 787: :+1:

ggainey (approved):

Issue #5912, opened 2024-10-15
Downloaders fail when current directory is not inside WORKING_DIRECTORY


gerrod3:
**Describe the bug**
If you run a downloader outside of the `settings.WORKING_DIRECTORY`, i.e. in the API-worker, then you will get a permission error.

Maybe the api-worker should have its current working directory match the task and content workers.

PR #5913, opened 2024-10-15
Have downloads always end up under WORKING_DIRECTORY

gerrod3:
fixes: #5912

### DIFF
--- a/pulpcore/download/base.py
+++ b/pulpcore/download/base.py
@@ -5,8 +5,10 @@ from collections import namedtuple
 import logging
 import os
 import tempfile
+from pathlib import Path
 from urllib.parse import urlsplit
 
+from django.conf import settings
 from pulpcore.app import pulp_hashlib
 from pulpcore.app.models import Artifact
 from pulpcore.exceptions import (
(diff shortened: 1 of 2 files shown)


pulpcore/download/base.py
```diff
@@ -127,7 +128,12 @@ def _ensure_writer_has_open_file(self):
             # write the file to the current working directory with a random prefix and the
             # desired suffix. we always want the random prefix as it is possible to download
             # the same filename from two different URLs, and the files may not be the same.
-            self._writer = tempfile.NamedTemporaryFile(dir=".", suffix=suffix, delete=False)
+            work_dir = str(settings.WORKING_DIRECTORY)
+            self._writer = tempfile.NamedTemporaryFile(
+                dir="." if work_dir in os.getcwd() else work_dir,
+                suffix=suffix,
+                delete=False,
+            )
```
lubosmj, line 136: It appears like this is a better variant of what I was trying to do in the past: https://github.com/pulp/pulp_container/pull/1299#discussion_r1291196612.

gerrod3: Yeah, I'm doing something similar with my new work on pull-through caching in pulp-python and I thought I should just move the fix into pulpcore.

mdellweg: Hmm this feels like patching over a deeper issue. But well.
I think we should leave a comment on why this is important and the assumptions that lead to there.

pulpcore/download/base.py
```diff
@@ -127,7 +128,12 @@ def _ensure_writer_has_open_file(self):
             # write the file to the current working directory with a random prefix and the
             # desired suffix. we always want the random prefix as it is possible to download
             # the same filename from two different URLs, and the files may not be the same.
-            self._writer = tempfile.NamedTemporaryFile(dir=".", suffix=suffix, delete=False)
+            work_dir = str(settings.WORKING_DIRECTORY)
+            self._writer = tempfile.NamedTemporaryFile(
+                dir="." if work_dir in os.getcwd() else work_dir,
```
mdellweg, line 133: This is a simple string comparison. So at least it should be `os.getcwd().startswith(work_dir)`. But i think i'd prefer using pathlib at this point.

gerrod3: Ok, updated to use pathlib logic and added a simple comment for the check.

mdellweg (approved):

PR #5924, opened 2024-10-22
Add optional output path parameter to sign and asign methods

### LINKED ISSUE #5923: Allow the signing service to provide an output for the sign script
We have an issue with how debian content and signing service are operating which leads to temp files not correctly cleaned after a task is done. The signing service script is creating its own temp files. Provide an output let's us use the temp directory created by the tasks and will ultimately clean up without a hassle.

Allow `sign` and `asign` methods to have an optional parameter that defines the output path.

Manually cleaning up tmp files works too, but is much riskier and more work and really we shouldn't have to do that.

**Additional context**
See pulp_deb issue https://github.com/pulp/pulp_deb/issues/1141

hstct:
This commit introduces an optional `output_path` parameter to the `sign` and `asign` methods, allowing signed content to be stored in a specified location. This helps with better handling and cleanup of temporary files.

closes #5923

### DIFF
--- a/CHANGES/5923.feature
+++ b/CHANGES/5923.feature
@@ -1,0 +1,1 @@
+Add optional ``output_path`` parameter to SigningService.sign and SigningService.asign methods to allow parsing an output to the signing script.
--- a/pulpcore/app/models/content.py
+++ b/pulpcore/app/models/content.py
@@ -785,7 +785,7 @@ class SigningService(BaseModel):
             env.update(env_vars)
         return env
 
-    def sign(self, filename, env_vars=None):
+    def sign(self, filename, output_path=None, env_vars=None):
         """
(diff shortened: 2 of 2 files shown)


hstct:
For more context the related pulp_deb PR: https://github.com/pulp/pulp_deb/pull/1142

> I am not particularly happy with this. Preferably the script should run in the working directory of the task that should be a subdirectory of `settings.WORKING_DIRECTORY`. If that's not an option, passing the output dir as an environment variable is probably less disruptive to existing signing scripts.
> 
> Somehow this reminds me of #5913.

Oh, you are right I'm able to achieve the same thing with the env variables but without making a change in pulpcore! This sounds preferable to me, so I will close this PR and issue. Thanks!

Issue #5934, opened 2024-10-24
Modify's add/remove_content_units is not validated for cross-domain content


gerrod3:
**Version**
<=pulpcore 3.66

**Describe the bug**
You can add content from another domain using the modify endpoint.

**To Reproduce**
Have content in one domain, then in a separate domain issue a repo modify command with content from the first domain.

**Expected behavior**
The request should be denied, 400.

PR #5935, opened 2024-10-24
Ensure users can not add cross-domain content through modify endpoint

gerrod3:
fixes: #5934

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -984,6 +984,7 @@ class RepositoryVersion(BaseModel):
         if self.complete:
             raise ResourceImmutableError(self)
 
+        assert not content.exclude(pulp_domain_id=get_domain_pk()).exists()
         repo_content = []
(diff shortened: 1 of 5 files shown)


pulpcore/app/models/repository.py
```diff
@@ -984,6 +984,7 @@ def add_content(self, content):
         if self.complete:
             raise ResourceImmutableError(self)
 
+        assert not content.exclude(pulp_domain_id=get_domain_pk()).exists()
```
mdellweg, line 987: This is really just a safeguard at this point, right?

mdellweg (dismissed):
This sounds like an ok solution to the problem described.

I wonder if we could revisit the idea to go into the model managers and only provide querysets limited to the current domain to solve this on a lower level. That may be a second PR.

pulpcore/app/models/repository.py
gerrod3: Pretty much.

mdellweg (approved):

PR #5936, opened 2024-10-24
Fix Any serializer type regression breaking Ruby bindings

pedro-psb:
See https://github.com/pulp/pulp_rpm/issues/3639

### DIFF
--- a/pulpcore/app/serializers/fields.py
+++ b/pulpcore/app/serializers/fields.py
@@ -4,6 +4,8 @@ from gettext import gettext as _
 from urllib.parse import urljoin
 
 from django.conf import settings
+from drf_spectacular.utils import extend_schema_field
+from drf_spectacular.types import OpenApiTypes
 from rest_framework import serializers
 from rest_framework.fields import empty
 
@@ -17,6 +19,15 @@ def relative_path_validator(relative_path):
         raise serializers.ValidationError(
             _("Relative path can't start with '/'. {0}").format(relative_path)
         )
+
+
+@extend_schema_field(OpenApiTypes.OBJECT)
+class JSONDictField(serializers.JSONField):
+    """A drf JSONField override to force openapi schema to use 'object' type.
+
+    Not strictly correct, but we relied on that for a long time.
+    See: https://github.com/tfranzel/drf-spectacular/issues/1095
+    """
 
 
 class SingleContentArtifactField(RelatedField):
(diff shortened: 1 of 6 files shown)


ianballou:
I'd  like to request that 3.63 receive the back port since that's what's in the Katello 4.15 matrix.

pulpcore/app/serializers/fields.py
```diff
@@ -19,6 +21,15 @@ def relative_path_validator(relative_path):
         )
 
 
+@extend_schema_field(OpenApiTypes.OBJECT)
+class JSONObjectField(serializers.JSONField):
+    """A drf JSONField override to force openapi schema to use 'object' type.
+
+    Not strictly correct, but we relied on that for a long time.
+    See: https://github.com/tfranzel/drf-spectacular/issues/1095
+    """
```
mdellweg, line 30: Let me understand this right. It sounds like a stop gap solution. But I think we are rather close to a good solution here. Most of our JSONFields actually require an object/dict. So how about we call this thing a `JSONDictField` (IMHO, the domain of the problem here is rather a python dev then an OpenAPIv3 connoisseur. So I prefer Dict over Object.) and it suddenly does not sound like an accident anymore. Instead when you use that field on a serializer, you are explicit about it needing a dictionary. At the same time we just ban the use of an un-annotated plain JSONField.
(Obviously, now this means, search and replace without reviewing is a bad idea again.)

pedro-psb: "Object" here is about the Json spec object structure: https://datatracker.ietf.org/doc/html/rfc7159#section-4
The correct name for this field as it is now would be JSONPrimitiveField (as drf docs calls a any-json entity), which reflects it can be any primitive.

I would like to get the fields right, which would involve switching to DictField and ListFIeld, mosly. I opened [this PR on RPM ](https://github.com/pulp/pulp_rpm/pull/3653/files) which does that with some care, but the concern is that it changes/breaks the API https://github.com/pulp/pulp_rpm/issues/3657#issuecomment-2236907538.

mdellweg (approved):
Thank you! I'd really like to make it sound less like an accident. Also we may want to export this helper in the plugin api.
That would be OK as a follow up PR.

pedro-psb:
> Also we may want to export this helper in the plugin api.

Yeah, agree

PR #5939, opened 2024-10-28
Expose json dict field plugin api

pedro-psb:
Stacked on top of:
* https://github.com/pulp/pulpcore/pull/5936


CHANGES/plugin_api/+expose-json-dict.misc
```diff
@@ -0,0 +1,3 @@
+Exposed JSONDictField on the plugin API.
+This is mostly for compatibility (see <https://github.com/pulp/pulp_rpm/issues/3639>).
+New fields that are supposed to be dict structures should use the DictField serializer field.
```
mdellweg, line 3: Reading this, I wonder if we understand the issue and the solution in the same way.

JSONDictField is really the right choice when the field recieves a dict containing anything backed by a JSONField in the database.
For Dict[str, str] we have HStoreField backed by the HStoreField in the db.
I don't know when's the right time to use a serializer.DictField (if you have postgres hstore available...).

pulpcore/app/serializers/fields.py
```diff
@@ -19,6 +21,15 @@ def relative_path_validator(relative_path):
         )
 
 
+@extend_schema_field(OpenApiTypes.OBJECT)
+class JSONDictField(serializers.JSONField):
+    """A drf JSONField override to force openapi schema to use 'object' type.
+
+    Not strictly correct, but we relied on that for a long time.
+    See: https://github.com/tfranzel/drf-spectacular/issues/1095
+    """
```
mdellweg, line 30: My whole point here is: When you _know_ you want dicts there and you go out of your way to specify this particular type explicitly, it is as a matter of fact correct.
When your field "wants" a list and you specify this field, it is just wrong and even the bindings will blow up.
```suggestion
    """A JSONField accepting dicts, specifying as type 'object' in the openapi."""
```

Now writing it this way, i realize, we may even start to add a bit of validation to this class.

pedro-psb: > When your field "wants" a list and you specify this field, it is just wrong and even the bindings will blow up.

That is the point: it is not "wrong" from the serializer.JSONField (and this subclass) implementation. The serializers will happily parse json booleans, lists and integers, and the database Field will save it on models.JSONField without a problem. Unless we add validation, as you said. Then it would be really wrong, in the serializer level.

Here is my reasoning:
1. We used serializer.JSONField for list and dict-like objects, which are saved correctly in the JSONField db. It is an any json type field.
2. Our bindings worked, because the schema was wrong:
    * It reported `object` type for this Field, when it should report an `any` type (not very accurate for API/bindings purposes).
3. We patch the field's schema representation to preserve the old behavior and not break things right away.
    * This is really supposed to be a patch that we'll remove when we can (on breaking releases, for example).
    * IHO, we shouldnt use this for any new fields, but do (4).
4. Now, on the long run, we fix/replace existing generic JSONType serializer fields.
    * We change only the serializer fields to be strictly dicts/list or whathever (if they are compatible with models.JSONField). This is mainly for more accurate API/schema representation.
    * We may also change the underlying model/db field, if we see a particular reason for it.

mdellweg: And to me this is the first instance of a more specific / accurate schema representation.
We would not use it if we saved a list in the json field. But we would need to add other annotations in that case, because (as far as i understood) the current implementation of the bindings cannot consume "Any".

pedro-psb: Yeah, that could work out too.
Then this PR should probably add that dict validation and add another JSONListField (with list validation and the list schema type), so plugins can use the correct one.

But I'm not 100% sure we can't just use DictField and ListField for dict/list structures instead of creating a new one based on JSONField. I assume from-API desserializing and the to-db serializing should be compatible, even if we keep model.JSONField. I'll test this out just to ensure. If that's the case we dont even need to expose anything at this point.

> because (as far as i understood) the current implementation of the bindings cannot consume "Any".

Yeah, the current Ruby generator version we use can't handle that (newer ones can).

mdellweg: Thank you. Adding the validation is probably a good call. If we can just switch to dict field (https://www.django-rest-framework.org/api-guide/fields/#dictfield looks like you can just "not provide the child parameter") that would be even nicer.
Adding a new JSONListField without knowing we are actually going to use it? IDK.

pedro-psb: >  Adding a new JSONListField without knowing we are actually going to use it? IDK.

Oh, we are using it:
https://github.com/pulp/pulp_rpm/blob/c3c9b1c4582f8e39aab6884951046c55cdf5d5d4/pulp_rpm/app/serializers/package.py#L81-L141

But again, if we can just use serializers.ListField that's better.

pedro-psb: From the implementation code, DictField's `to_internal_value` and `to_representation` are substantially different from JSONField.
I'll move foward and add JSONDictField and JSONListField with that validation on top and with the spectacular type override, does that sounds ok?

pulpcore/app/serializers/fields.py
```diff
@@ -21,13 +21,48 @@ def relative_path_validator(relative_path):
...
+        returns_python_repr = self.binary or getattr(data, "is_json_string", False)
+        if returns_python_repr:
+            if not isinstance(value, dict):
+                raise serializers.ValidationError(ERROR_MSG)
+        elif not value.strip().startswith("{"):
+            raise serializers.ValidationError(ERROR_MSG)
+        return value
+
+
+@extend_schema_field(list)
```
pedro-psb, line 49: That's not working as expected and OpenApiTypes doesn't have a LIST/ARRAY type (not sure why).

pulpcore/app/serializers/fields.py
```diff
@@ -21,13 +21,48 @@ def relative_path_validator(relative_path):
...
+        # This condition is from the JSONField source:
+        # if it's True, it will return the python representation,
+        # else the raw data string
+        returns_python_repr = self.binary or getattr(data, "is_json_string", False)
+        if returns_python_repr:
+            if not isinstance(value, list):
+                raise serializers.ValidationError(ERROR_MSG)
+        elif not value.strip().startswith("["):
+            raise serializers.ValidationError(ERROR_MSG)
+        return value
```
pedro-psb, line 65: A manual test with such a field results in:

```json
                    "listy": {
                        "type": "array",
                        "items": {},
                        "readOnly": true,
                        "description": "Testing that JSONListField works."
```

Not sure how I would write an automated test for this.

mdellweg: Well that _is_ a good question.
I guess, within pulpcore/file/certguard nothing uses this new field? So we cannot assert on the schema generated here, right?

As for unittesting, it might be worth to look how spectacular tests the schema generation.

pedro-psb: > As for unittesting, it might be worth to look how spectacular tests the schema generation.

Yeah, that's a good idea

pedro-psb:
@mdellweg Do you want me to add the tests first?

I found some reference [on drf repo](https://github.com/tfranzel/drf-spectacular/blob/27ae6f293a6e8731649216d62ba2a5c8d65691e6/tests/test_basic.py#L78) on how to do it.
Looks like copying over some utility functions will do it + creating model/serializer/expected-schema.

mdellweg:
I think we are good with the unittests you already added.

PR #5953, opened 2024-10-29
[3.66] Fix Any serializer type regression breaking Ruby bindings

pedro-psb:
just checking if failing test in https://github.com/pulp/pulpcore/pull/5947 is consistent (can't rerun the job there).


pedro-psb:
it doesnt

PR #5955, opened 2024-10-30
Allow aiohttp <3.10.11

mdellweg:
This is supposed to allow consuming a bug in creating the caching key.


gerrod3 (approved):

mdellweg:
https://github.com/pulp/pulpcore/pull/5958

PR #5959, opened 2024-10-30
Document S3 operations that the AWS access key must be able to perform

FrostyX:
SSIA

### DIFF
--- a/docs/user/guides/create-domains.md
+++ b/docs/user/guides/create-domains.md
@@ -83,6 +83,11 @@ Specific examples for different storage backends:
     orphaned content.
 
 
+!!! note
+    The AWS access key must be authorized to perform the `s3:ListBucket` and `s3:PutObject`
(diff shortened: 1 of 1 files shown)


mdellweg (approved):
Good catch.

Looking at https://pulpproject.org/pulpcore/docs/admin/reference/settings/#default_file_storage , I wonder where the "storage documentation" went. At some point we had a link to django-storages, because we just cannot cover all, but  we also do not support __all__ the storage types there.
cc @pedro-psb

pedro-psb:
I think when triaging that with @dkliban we left that over because those configurations should be on pulp-oci-images or pulp_operator, if I remember correctly.

But here is the page:
https://github.com/pulp/pulpcore/blob/753fa3d62fe8d1ec2e5cd2e2ccf6649d06ced15b/docs/installation/storage.rst

PR #5969, opened 2024-11-01
Add cache_key to the plugin api

gerrod3:
cache_key handles domains correctly so plugins using the cache should use this method.

Side question: can we backport additions to the plugin-api?


mdellweg (approved):

mdellweg:
> Side question: can we backport additions to the plugin-api?

This is a double edged sword...

PR #5970, opened 2024-11-01
Fix cross-domain content assertion

gerrod3:
We allow plugins to pass in either a `Content` query or a queryset of `content_ids`, e.g. `RepositoryContent.objects.values_list("content_id")`


mdellweg (approved):

mdellweg (changes requested):
Actually, this comes as a regression fix. We need at least a changelog.

mdellweg (approved):

PR #5983, opened 2024-11-07
Future proof the pulp_file repair tests for python 3.11

hstct:
This change unblocks the functional pulp_file repair tests when run on a python 3.11+ environment. Stronger typing for the random.sample function is introduced in 3.11 and requires now a Sequence type as an input or it will throw a TypeError.

see: https://docs.python.org/3.11/library/random.html#random.sample

[noissue]


dosas (approved):

gerrod3 (approved):

gerrod3:
I wonder if we should backport this, 3.49+ branches are running our latest image which is on python3.9. Currently our logic to build images with higher python versions is broken, but I plan to fix that, and eventually we will switch the default to a higher version (python3.9 EOL is next year October).

Issue #5967, opened 2024-11-01
On content-app, ClientConnectorError should result in retries for on-demand


pedro-psb:
**Describe the bug**
When the client request a "on-demand-synced package" for the first time, the content-app will try to get it from all available RA for that ContentArtifact if the response was not yet started (that's it , we didnt receive any data from that remote).

But in case the first RA raises a `ClientConnectorError`, for example, the content-app will not be able to deliver the content, even if the second RA is okay.

```
aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host {hostname}:443 ssl:default [Name or service not known]
```

https://github.com/pulp/pulpcore/blob/6b066da9a98690fd5f7639c2089ef65ca32869b8/pulpcore/content/handler.py#L821-L834

**To Reproduce**
* Create a Distribution pointing to a Repository
* Add a Content in a Repository with an artifactless ContentArtifact
* Associate two RemoteArtifact to it. Ensure $RA_1.pulp\textunderscore id < RA_2.pulp\textunderscore id$
  * $RA_1$ - url to the content in an invalid server address (e.g, `http://thisdoesntexist.com/blob`)
  * $RA_2$ - url to the content in a valid server address. Should be running and serving the content file as expected.
* Try to get the content using the Distribution base_path. An `ConnectionClosedError` should raise.

**Expected behavior**
The content-app catches the `ClientConnectorError` and try to get the content from the next RA.

**Additional context**
This was reported in our slack channel for a Satellite case after a Remote url was changed and resynced.
The recommended fix was to remove $RA_1$ (known to be bad) using pulp python shell.

PR #5984, opened 2024-11-07
Add ClientConnectionError as retriable exception on content-app streaming

pedro-psb:
A user reported that after changing the host and resyncing they would have some some leftover RAs with the invalid url, and that would raise an aiohttp.client_exceptions.ClientConnectionError.

Because that is not a retriable error for content streaming, the right RA wouldnt be tried afterwards.

Closes: #5967

---

We should probably backport these to supported branches.

### DIFF
--- a/CHANGES/5967.bugfix
+++ b/CHANGES/5967.bugfix
@@ -1,0 +1,2 @@
+On the content-app, added ClientConnectionError (aiohttp) as an exception that can trigger
+a retry when streaming content from a Remote.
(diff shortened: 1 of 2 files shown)


ggainey (changes requested):
The PR Machinery doesn't recognize "Closes:" -  - that's why this is marked no-issue  :)  

Also, really should be "fixes #5967", since you're fixing a bug, not closing an RFE.

ggainey (approved):

PR #5994, opened 2024-11-10
Remove functional tests for opentelemetry instrumentation

ggainey:
(cherry picked from commit 981af2ef7b01d19b6b40b2c3c1a5dcbbc0bdf7a7)


.github/workflows/scripts/pre_before_script.sh
```diff
@@ -4,6 +4,6 @@ if [ "$TEST" = "azure" ]; then
     cmd_stdin_prefix bash -c "cat > /etc/nginx/pulp/api_root_rewrite.conf" < pulpcore/tests/functional/assets/api_root_rewrite.conf
     cmd_prefix bash -c "s6-rc -d change nginx"
     cmd_prefix bash -c "s6-rc -u change nginx"
-    cmd_stdin_prefix bash -c "cat > /var/lib/pulp/scripts/otel_server.py" < pulpcore/tests/functional/assets/otel_server.py
-    cmd_user_prefix nohup python3 /var/lib/pulp/scripts/otel_server.py &
+    # it is necessary to wait for a short period until the service starts; otherwise, any subsequent requests to Pulp's API endpoints will fail
+    sleep 1
```
mdellweg, line 8: Oh, this sounds like superstition. At least the one second is plugged out of thin air.
Maybe ok for here, but we should look into injecting the api_root_rewrite.conf in the containerfile and get rid of restarting a service inside the container altogether.

.gitignore
```diff
@@ -41,6 +41,7 @@ tags
 # Vagrant
 .vagrant
 Vagrantfile
+__pycache__/
```
mdellweg, line 44: Of all the ignores on earth, why was this one ever missing? Also this is really python3 specific, not "Vagrant"...

mdellweg (approved):
Wow some things look strange in here. But as for the backporting. Let's run with it.

PR #6007, opened 2024-11-11
Finish backporting aiohttp 3.10 bump

gerrod3:
Bumping this req broke Pulp, but only when using the downloader outside of a coroutine, which some plugins do (pulp_container).

 (from commit 3b0d218af8ddf97a1d5831dd8cab080194c0c768)

### DIFF
--- a/pulpcore/download/factory.py
+++ b/pulpcore/download/factory.py
@@ -137,8 +137,11 @@ class DownloaderFactory:
             sock_read=self._remote.sock_read_timeout,
             connect=self._remote.connect_timeout,
         )
+        # TCPConnector is supposed to be instanciated in a running loop.
+        # I don't see why...
+        # https://github.com/aio-libs/aiohttp/pull/3372
         return aiohttp.ClientSession(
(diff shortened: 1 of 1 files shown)


gerrod3:
Side note, we will probably have problems with this again when we upgrade past python 3.12, as it will most likely become an error one day: https://docs.python.org/3.11/library/asyncio-eventloop.html#asyncio.get_event_loop

ggainey (approved):

gerrod3:
Uggh, seems like we have an inconsistent test on the azure runner for 3.49. Seems to fail on the retry-with-bad-response test. The fixture seems to be giving an error. Not sure were to start with this one.

ggainey:
> Uggh, seems like we have an inconsistent test on the azure runner for 3.49. Seems to fail on the retry-with-bad-response test. The fixture seems to be giving an error. Not sure where to start with this one.

FWIW, we've been seing a number of inconsistent tests in the nightlies, that "work fine on restart". hard to pin down exactly what is going on - esp when sometimes, it's clearly "github hiccupped/timed-out/etc"

PR #6019, opened 2024-11-13
Switch from using setup.py to pyproject.toml

mdellweg:


pedro-psb (approved):
I'm not an expert in packaging but LGTM.

Issue #6021, opened 2024-11-13
Evaluate the usage of uvloop in the content-app


pedro-psb:
The [uvloop](https://github.com/MagicStack/uvloop) library advertises itself as:
> a fast, drop-in replacement of the built-in asyncio event loop. uvloop is implemented in Cython and uses libuv under the hood.

We could test if that works out-of-the-box and if we can see meaningful performance improvements (for example in the services deployment).

[Look like](https://docs.aiohttp.org/en/stable/deployment.html#start-gunicorn) its a matter of adding `uvloop` as a requirement and replacing `aiohttp.GunicornWebWorker` with `aiohttp.GunicornUVLoopWebWorker` in our Gunicorn application.

PR #6023, opened 2024-11-14
Experiment with uvloop in the Gunicorn worker

pedro-psb:
Just checking if nothing breaks.

https://github.com/pulp/pulpcore/issues/6021

---

- worker-class option on gunicorn: https://docs.gunicorn.org/en/latest/settings.html#worker-class
- uvloop recommendation from aiohttp: https://docs.aiohttp.org/en/stable/third_party.html#approved-third-party-libraries

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,6 +66,7 @@ dependencies = [
   "tablib<3.6.0",
   "url-normalize>=1.4.3,<=1.4.3",
   "uuid6>=2023.5.2,<=2024.7.10",
+  "uvloop>=0.21.0,<0.22",
   "whitenoise>=5.0,<6.9.0",
   "yarl>=1.8,<1.17.3",
 ]
--- a/CHANGES/6021.feature
+++ b/CHANGES/6021.feature
@@ -1,0 +1,1 @@
+Replaced async with uvloop as the Gunicorn worker eventloop.
--- a/pulpcore/content/entrypoint.py
+++ b/pulpcore/content/entrypoint.py
@@ -5,7 +5,7 @@ from pulpcore.app.pulpcore_gunicorn_application import PulpcoreGunicornApplicati
 class PulpcoreContentApplication(PulpcoreGunicornApplication):
     def load_app_specific_config(self):
         self.set_option("default_proc_name", "pulpcore-content", enforced=True)
-        self.set_option("worker_class", "aiohttp.GunicornWebWorker", enforced=True)
+        self.set_option("worker_class", "aiohttp.GunicornUVLoopWebWorker", enforced=True)
 
     def load(self):
         import pulpcore.content
@@ -15,6 +15,11 @@ class PulpcoreContentApplication(PulpcoreGunicornApplication):
 
 @click.option("--bind", "-b", default="[::]:24816")
 @click.option("--workers", "-w", type=int)
+@click.option(
+    "--worker-class",
+    "-k",
+    type=click.Choice(["aiohttp.GunicornWebWorker", "aiohttp.GunicornUVLoopWebWorker"]),
+)
 # @click.option("--threads", "-w", type=int)  # We don't use a threaded worker...
 @click.option("--name", "-n", "proc_name")
 @click.option("--timeout", "-t", type=int)


pedro-psb:
Just a note, maybe we should watch out for [this](https://docs.aiohttp.org/en/stable/faq.html#why-is-creating-a-clientsession-outside-of-an-event-loop-dangerous). We've done something about it, but: https://github.com/pulp/pulpcore/pull/5666#issuecomment-2275274707

bmbouter:
I'm +1 on this!

dralley:
Are we intending to actually merge this now or does it remain just a test?  Do we have any idea for example how much of a difference it makes in practice?

bmbouter:
I think we should merge it. This only considers the overhead of the loop itself (whereas, in reality the overwhelming majority of the time is in our application code), but if we're comparing asyncio (what we use) versus uvloop, uvloop seems over twice as fast. https://magic.io/blog/uvloop-blazing-fast-python-networking/

decko:
Do we also need to set the asyncio event loop to use uvloop?
```asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())```

Also, setting this in the entrypoint, is this gonna work to all the application itself?

File: pyproject.toml [hunk 1]
```diff
@@ -66,6 +66,7 @@ dependencies = [
   "tablib<3.6.0",
   "url-normalize>=1.4.3,<=1.4.3",
   "uuid6>=2023.5.2,<=2024.7.10",
+  "uvloop>=0.21.0,<0.22",
```
mdellweg, line 69: Can we consider adding this as an optional dependency?

mdellweg (changes requested):

pyproject.toml
pedro-psb: Yes. Personally I would prefer to merge this as an optional feature.
Maybe do some discourse post to invite people give it a try and provide feedback.

pedro-psb:
> @decko Do we also need to set the asyncio event loop to use uvloop?

That's literally the only thing that new worker do! :smile: 
https://github.com/aio-libs/aiohttp/blob/4adb0616a6de8351c45779048610081e000f7c6a/aiohttp/worker.py#L230

> Also, setting this in the entrypoint, is this gonna work to all the application itself?

Not sure I understand what you mean. What I understand is that specifically all new content-app workers will use it.

pyproject.toml
pedro-psb: @bmbouter Can we start with that? ^

bmbouter: Optional dependencies seem complicated. To me, the containers we build for the community would still have it installed because I don't think we want to ship options that then require users to rebuild containers. Pulp (imo) should Just Work™

At that point if it's optional in pulpcore, but still in 100% of the officially supported install method we'd do better just to keep it simple.

I'm fine with whatever you all decide.

pedro-psb: My problem is not having strong evidence. I believe it will be fine (not break things) and improve performance, but I would like a bit more than that before shipping it widely. And also to justify to the build team why we are adding more deps.

I have a proposal which I prefer (over adding as optional and hoping someone will opt-in and provide feedback):
1) We ship this as a patch on services for a fixed amount of time (1-3 months? on the agenda) with the purpose of gaining more trust (against breaks or regressions).
2) We conduct some benchmarking (on services or not) and write a report. I would be happy to do/participate on that.

dralley: I agree with all of that ^ FWIW.

bmbouter: I like this idea because it'll produce some real-world analysis. We can evaluate it first in the staging environment.

To do this we'll need the patch to be applied to [this repo](https://github.com/pulp/pulp-service/). Also have the usage of it enabled/disabled on the presence of an environment variable. If the env var is set use it, if the env var is non-existant don't. That will be the safest because then we can apply the patch and anywhere it goes it'll not break anything.

Then @dkliban or @decko can set the env var in the deployment repo (another repo with other contributors perms that are kind of hard to adjust). Also they can add uvloop to the container build file.

@pedro-psb could you open a patch like that? Feel free to message with us more about it, we'd like to help make it easy.

pedro-psb: Yes, I can submit the patch.

PR #6024, opened 2024-11-14
Squash 3.20

mdellweg:


ggainey (approved):
Based on my understanding of how squash works, and since we're not deleting the existing migrations - this looks good.

PR #6051, opened 2024-11-20
Update CI

dralley:


template_config.yml
```diff
@@ -38,7 +38,7 @@ plugins:
 - app_label: file
   name: pulp_file
 - app_label: certguard
-  name: pulp-certguard
+  name: pulp_certguard
```
dralley, line 41: ^

mdellweg: Hmmm. Should be fine. After all the name will be canonicalized in the really important places anyway.
Maybe you want to rename the subdirectory too and we filanny have all the plugins follow the same naming convention.

mdellweg: Oh, the directory has always been `pulp_certguard` only the repository was hyphened.

dralley: Yes, the name of the file in the bumpconfig was correct (used underscore), and when it became templatized based on this value it became incorrect.

PR #6058, opened 2024-11-20
[PULP-210] Update deprecated django storage config

### LINKED ISSUE #5404: Django `DEFAULT_FILE_STORAGE` deprecated
The `DEFAULT_FILE_STORAGE` option was deprecated in django `4.2.0` in favor of another option called [`STORAGES`](https://docs.djangoproject.com/en/4.2/ref/settings/#storages).

- django docs: https://docs.djangoproject.com/en/4.2/ref/settings/#default-file-storage
- django storages docs: https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#configuration-settings

Since we're using `django~=4.2.0`, we may want to change our [docs](https://staging-docs.pulpproject.org/pulpcore/docs/admin/learn/settings/?h=default#default_file_storage) (and possibly other places) to use the recommended way.

pedro-psb:
## Summary

The setting [DEFAULT_FILE_STORAGE](https://docs.djangoproject.com/en/4.2/ref/settings/#default-file-storage) / [STATCIFILES_STORAGE](https://docs.djangoproject.com/en/4.2/ref/settings/#staticfiles-storage) was deprecated in django 4.2. They've implemented a compatibility layer to enable this legacy setting to be mapped to the new [STORAGES](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-STORAGES), so users could gradually move and Pulp codebase to update to using only STORAGES.

The problem is that Pulp overrides the default (using the legacy setting name), so users can't really use STORAGE, as it raises a django exceptions on mutual exclusion.

This PR does:

1. Updates the codebase to rely on settings.STORAGES instead of DEFAULT_FILE_STORAGE
2. Enable users to start using STORAGES on their settings
3. Use both legacy and new setting in CI (s3 / azure) for sanity testing

## Open questions:
- **Removal deadline**: I'm suggesting that we remove it on 3.85 or 4.0, but we could choose something else. Deciding this is important so we can give a proper heads up for our users and stakeholders.

## Plugins TODO

- Check whether the signal handling (on `app/apps.py`) uses a module-level settings. That could cause errors, so it needs to be moved inside the AppConfig class
- Use STORAGES internally instead of DEFAULT_FILE_STORAGE and STATICFILES_STORAGE

Closes #5404

### DIFF
--- a/CHANGES/plugin_api/5404.removal
+++ b/CHANGES/plugin_api/5404.removal
@@ -1,0 +1,9 @@
+Started using `settings.STORAGES` internally instead of `settings.DEFAULT_FILE_STORAGE` and `settings.STATICFILES_STORAGE`,
+which was deprecated in Django 4.2.
+
+For compatibility, plugins must replace access to:
+
+* `settings.DEFAULT_FILE_STORAGE` with `settings.STORAGES["default"]["BACKEND"]`
+* `settings.STATICFILES_STORAGE` with `settings.STORAGES["staticfiles"]["BACKEND"]`
+
+See the new storage structure in [Django docs](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-STORAGES).
--- a/pulp_file/tests/functional/api/test_download_policies.py
+++ b/pulp_file/tests/functional/api/test_download_policies.py
@@ -11,7 +11,6 @@ from urllib.parse import urljoin
 
 from pulpcore.tests.functional.utils import get_files_in_manifest, download_file
 
-from pulpcore.app import settings
 from pulpcore.client.pulp_file import FileFilePublication, RepositorySyncURL
 
 
@@ -54,7 +53,8 @@ def test_download_policy(
     download_policy,
 ):
     """Test that "on_demand" and "streamed" download policies work as expected."""
-    if download_policy == "on_demand" and "SFTP" in pulp_settings.DEFAULT_FILE_STORAGE:
+    settings = pulp_settings
+    if download_policy == "on_demand" and "SFTP" in settings.STORAGES["default"]["BACKEND"]:
         pytest.skip("This storage technology is not properly supported.")
 
     remote = file_remote_ssl_factory(
@@ -148,7 +148,7 @@ def test_download_policy(
     assert expected_checksum == actual_checksum
     if (
         download_policy == "immediate"
-        and settings.DEFAULT_FILE_STORAGE != "pulpcore.app.models.storage.FileSystem"
+        and settings.STORAGES["default"]["BACKEND"] != "pulpcore.app.models.storage.FileSystem"
         and settings.REDIRECT_TO_OBJECT_STORAGE
     ):
         content_disposition = downloaded_file.response_obj.headers.get("Content-Disposition")
@@ -187,7 +187,7 @@ def test_download_policy(
     content_unit = expected_files_list[4]
     content_unit_url = urljoin(distribution.base_url, content_unit[0])
     # The S3 test API project doesn't handle invalid Range values correctly
-    if settings.DEFAULT_FILE_STORAGE == "pulpcore.app.models.storage.FileSystem":
+    if settings.STORAGES["default"]["BACKEND"] == "pulpcore.app.models.storage.FileSystem":
         with pytest.raises(ClientResponseError) as exc:
             range_header = {"Range": "bytes=-1-11"}
             download_file(content_unit_url, headers=range_header)
--- a/pulpcore/app/apps.py
+++ b/pulpcore/app/apps.py
@@ -323,11 +323,11 @@ def _ensure_default_domain(sender, **kwargs):
         if (
             settings.HIDE_GUARDED_DISTRIBUTIONS != default.hide_guarded_distributions
             or settings.REDIRECT_TO_OBJECT_STORAGE != default.redirect_to_object_storage
-            or settings.DEFAULT_FILE_STORAGE != default.storage_class
+            or settings.STORAGES["default"]["BACKEND"] != default.storage_class
         ):
             default.hide_guarded_distributions = settings.HIDE_GUARDED_DISTRIBUTIONS
             default.redirect_to_object_storage = settings.REDIRECT_TO_OBJECT_STORAGE
-            default.storage_class = settings.DEFAULT_FILE_STORAGE
+            default.storage_class = settings.STORAGES["default"]["BACKEND"]
             default.save(skip_hooks=True)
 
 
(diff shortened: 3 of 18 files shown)


pulpcore/app/settings.py
```diff
@@ -485,7 +510,14 @@ def otel_middleware_hook(settings):
...
     post_hooks=otel_middleware_hook,
 )
-# HERE ENDS DYNACONF EXTENSION LOAD (No more code below this line)
+
+# begin compatilibity layer for DEFAULT_FILE_STORAGE
+# Remove on pulpcore=3.85 or pulpcore=4.0
+
+# Ensures the cached property storage.backends uses the the right value
+storages._backends = settings.STORAGES.copy()
+storages.backends
```
mdellweg, line 519: Is this the "magic" to prevent django storages to see both items we set before?

pulpcore/app/settings.py
```diff
@@ -56,7 +58,28 @@
...
+    "default": {
+        "BACKEND": "pulpcore.app.models.storage.FileSystem",
+    },
+    "staticfiles": {
+        "BACKEND": "django.contrib.staticfiles.storage.StaticFilesStorage",
+    },
+}
+
+setattr(global_settings, "DEFAULT_FILE_STORAGE", _DEFAULT_FILE_STORAGE)
+setattr(global_settings, "STORAGES", _STORAGES)
```
mdellweg, line 80: Setting them both here seems to contradict that they are, as you say, "mutually exclusive".

CHANGES/plugin_api/5404.removal
```diff
@@ -0,0 +1 @@
+Removed `DEFAULT_FILE_STORAGE` setting in favour of the new `STORAGES` options introduced in Django 4.2.
```
mdellweg, line 1: If I understand, this is a hard breaking change for plugins with storage specific code, right?
Or can plugins still "see" the old setting alongside the new one?

Also we want users to migrate to the new setting (sooner than later). You should add a second changelog for that.

pulpcore/app/settings.py
pedro-psb: The "global setting" settings is the only place where django won't complain about exclusivity.

CHANGES/plugin_api/5404.removal
pedro-psb: This message is not very accurate, forget to update it.
What I'm really doing is (1) updating codebase to use STORAGES + (2) allowing users to use either of both. I guess it can be two entries, with the second one stressing that now users should start migrating.

Django copies the value of DEFAULT_FILE_STORAGE to STORAGES, but not the other way around.
So yes, plugins should use STORAGES internally.

pulpcore/app/settings.py
pedro-psb: No, Django in fact has both set in their [default `global_settings`](https://github.com/django/django/blob/39cf3c63f3228a04f101f3e62c75a6aae7c6ef0f/django/conf/global_settings.py#L280).
But the previous hack caused a sife-effect on a specific point of django startup where it tries to get those values with a method that can't see the patch we made. The storages is a cached property, so it would hold the wrong value forever.

Line 518 does exactly what they [do here](https://github.com/django/django/blob/3ee4c6a27ad520d4ecc3cded260d47cbccafb144/django/core/files/storage/handler.py#L21).
Line 519 triggers the value to be cached.

All this magic is to give users a larger time span to upgrade.

pulpcore/app/settings.py
mdellweg: Sounds weird, but I'll take your word that it works.

CHANGES/plugin_api/5404.removal
```diff
@@ -0,0 +1,9 @@
+Started using `STORAGES` internally instead of `DEFAULT_FILE_STORAGE` and `STATICFILES_STORAGE`,
+which was deprecated in Django 4.2.
+
+For compatibility, plugins should replace access to:
```
mdellweg, line 4:
```suggestion
For compatibility, plugins must replace access to:
```

Use strong words where appropriate...

CHANGES/5404.feature
```diff
@@ -0,0 +1,6 @@
+Enable users to gradually upgrade from `DEFAULT_FILE_STORAGE` and `STATIC_FILE_STORAGE` to 'STORAGES'.
+These legacy settings were deprecated in Django 4.2 and will be removed in Pulp 3.85 or Pulp 4,
+whichever comes first.
```
mdellweg, line 3:
```suggestion
These legacy settings were deprecated in Django 4.2 and will be removed in Pulp 3.85.
```

I appreciate the effort to get the pulp 4 idea out, but i think that part is implicit and my drag the attention away from what is important here.
Or put another way. I think we can at any time in the future claim that pulp 3.(70+n*15) will actually be pulp 4 and it will inherit the deprecations naturally.

mdellweg (commented):
Just nitpicking on the changelogs...

CHANGES/plugin_api/5404.removal
```diff
@@ -0,0 +1,9 @@
+Started using `STORAGES` internally instead of `DEFAULT_FILE_STORAGE` and `STATICFILES_STORAGE`,
```
mdellweg, line 1:
```suggestion
Started using `settings.STORAGES` internally instead of `settings.DEFAULT_FILE_STORAGE` and `settings.STATICFILES_STORAGE`,
```

pulp_file/tests/functional/api/test_download_policies.py
```diff
@@ -11,7 +11,6 @@
 
 from pulpcore.tests.functional.utils import get_files_in_manifest, download_file
 
-from pulpcore.app import settings
```
mdellweg, line 14: Oh dear, this test was mixing both "imports"...?

pulpcore/app/apps.py
```diff
@@ -68,6 +67,11 @@ class PulpPluginAppConfig(apps.AppConfig):
 
     def __init__(self, app_name, app_module):
         super().__init__(app_name, app_module)
+        # Workaround for getting the up-to-date settings instance.
+        # A module-level settings import is a cached version from before dynaconf do its work.
+        from django.conf import settings
+
+        self.settings = settings
```
mdellweg, line 74: Is this part of the deprecation compatibility layer?
If so, please mark it accordingly so we know when to remove again.

pedro-psb: Yes, will do

pulp_file/tests/functional/api/test_download_policies.py
pedro-psb: Yes, it was

mdellweg (approved):
Feel free to merge.
I was just wondering if the user facing changelog should be moved to the decprecations section (It includes a deprecation for 3.85).

pedro-psb:
I'll do that. I didnt want to use `.removal` because its not removed yet, but this better conveys the intent.

mdellweg:
If you look, the title for ".removal" will be "Deprecations and Removals".

mdellweg (approved):

PR #6026, opened 2024-11-14
[SAT-22998] Ensure Pulp closes the connection on corrupted streamed content

pedro-psb:
Because of a design choice of streaming chunk-by-chunk as we receive it from the Remote (to minimize response time and allow only-stream workflow), as soon as we receive the first chunk, all http headers are already sent and we can't roll that back anymore (for example, set a 404).

We could let the user find out about that, but we prefer to prevent it from unecessarily saving/validation something we know is wrong. Because of that, we choose to close the TCP connection and raise a more helpful message on Pulp logs targeted at admins.

fixes #5012

PR #6064, opened 2024-11-22
[SAT-29018] Fix/corrupted RA blocks content streaming

### LINKED ISSUE #5725: Wrong package download after on-demand sync
Originally reported [here](https://issues.redhat.com/browse/SAT-17783).
I'm bringing it here for visibility in the RemoteArtifact bucket of issues.

## Version

Confirmed to happen with (at least).
- pulpcore==3.21, pulp_rpm==3.18
- pulpcore==3.39, pulp_rpm==3.23

## Describe the bug

When:
- Repos A and B syncs (on-demand) the same Content from two independent Remotes
- In Remote B, the Content is replaced (e.g in rpm, same Nevra different checksum) and synced again (on-demand).
- A request is made to the Content for Distribution associated with Repo A

Then:
- Pulp raises a digest validation error and retries to download from the same (wrong) remote
- Client can't get any content

## To Reproduce

This is a flaky problem. Most likely it depends on sorting RemoteArtifact via uuid4, which is random.
I'll write a Pulp test and share it here to make it easier to re-run.

1. Serve a signed package in two different "external" (non-pulp) repos. E.g, download [sample](https://fixtures.pulpproject.org/rpm-signed/giraffe-0.67-2.noarch.rpm), run createrepo and make sure it's reachable from the Pulp instance.
2. For each of the two repos:
    - Create a Repository (autopublish=True), Remote (policy=on_demand and corresponding url) and Distribution
    - sync (policy=mirror_content_only)
3. Change one of the external repos to serve the unsigned version of the package. E.g, delete the signed, put [this one](https://fixtures.pulpproject.org/rpm-unsigned/giraffe-0.67-2.noarch.rpm), run createrepo and sync the corresponding Pulp repo again.
4. Request the package from the distribution whose package hasnt changed (should contain signed) and see the error.

## Expected behavior

The right package should be delivered to the client, as its present and reachable in Remote A.

## Additional context

- On previous versions (core/rpm 3.16/3.17), the client would effectively get the wrong package, but that's not the case in supported versions anymore.
- Overview of what is happening in Pulp:

image

pedro-psb:
depends on #6026 
closes #5725


pulp_file/pytest_plugin.py
```diff
@@ -364,3 +364,19 @@ def _wget_recursive_download_on_host(url, destination):
...
+    def _generate_server_and_remote(*, manifest_path, policy):
+        server = gen_fixture_server(file_fixtures_root, None)
+        url = server.make_url(manifest_path)
+        remote = gen_object_with_cleanup(
+            file_bindings.RemotesFileApi,
+            {"name": str(uuid.uuid4()), "url": str(url), "policy": policy},
+        )
+        return server, remote
+
+    yield _generate_server_and_remote
```
pedro-psb, line 382: Moved from `test_acs` to facilitate creating an ACS + server.
I need an ACS here because it associated RA has priority on content-streaming.

pulpcore/content/handler.py
```diff
@@ -828,9 +830,11 @@ async def _stream_content_artifact(self, request, response, content_artifact):
...
         )
 
-        remote_artifacts = content_artifact.remoteartifact_set.select_related(
-            "remote"
-        ).order_by_acs()
+        remote_artifacts = (
+            content_artifact.remoteartifact_set.select_related("remote")
+            .order_by_acs()
+            .exclude(failed_at__gte=timezone.now() - timedelta(minutes=5))
+        )
```
pedro-psb, line 837: I though it would be nice to start simple, but maybe this threshold could depend on the content size? Or be configurable via settings?

dralley: Configurable is probably better

pulpcore/content/handler.py
```diff
@@ -1140,14 +1144,18 @@ async def finalize():
...
-                "At this point, we cant recover from wrong data already sent, "
-                "so we are forcing the connection to close. "
-                "If this error persists, the remote server might be corrupted."
+                f"DigestValidationError: Pulp tried streaming {remote_artifact.url!r} to "
+                "the client, but it failed checkusm validation.\n\n"
+                "We cant recover from wrong data already sent so we are:\n"
+                "- Forcing the connection to close.\n"
+                "- Marking this Remote Server to be ignored for the next 5 minutes.\n\n"
+                "If the Remote is permanently corrupted, we advice the admin to "
+                "manually prune affected RemoteArtifacts/Remotes from Pulp."
```
dralley, line 1158: There's not really a way to delete RemoteArtifacts directly without messing with the database, and I'm not sure we want to suggest in any way that the user should go digging in there.  Deleting remotes is probably better, and we should probably suggest resyncing the repo as it might have been fixed at the source.

Typo "checksum" on line 1153

pedro-psb: > There's not really a way to delete RemoteArtifacts directly without messing with the database, and I'm not sure we want to suggest in any way that the user should go digging in there.

Fair, I was thinking in the last case, but that is probably not a good general idea.

dralley: Yeah, I mean, it CAN be done, occasionally if someone comes to us asking how to do it we'll help them out, but it's not a road we want to suggest going down

pulpcore/app/settings.py
```diff
@@ -295,6 +295,11 @@
     "EXPIRES_TTL": 600,  # 10 minutes
 }
 
+# The time a RemoteArtifact will be ignored after failure.
+# In on-demand, if a fetching content from a remote failed due to corrupt data,
+# the corresponding RemoteArtifact will be ignored for that time (seconds).
+FAILED_REMOTE_ARTIFACT_PROTECTION_TIME = 5 * 60  # 5 minutes
```
mdellweg, line 301: This sounds more like a cooldown to me, not a protection time.

Have we considered making this a constant before adding a new setting?

pulpcore/content/handler.py
```diff
@@ -1140,14 +1145,22 @@ async def finalize():
...
+                f"Pulp tried streaming {remote_artifact.url!r} to "
+                "the client, but it failed checksum validation.\n\n"
+                "We can't recover from wrong data already sent so we are:\n"
+                "- Forcing the connection to close.\n"
+                "- Marking this Remote to be ignored for "
+                f"{FAILED_REMOTE_ARTIFACT_PROTECTION_TIME=}s.\n\n"
+                "If the Remote is known to be fixed, try resyncing the associated repository.\n"
+                "If the Remote is known to be permanently corrupted, try removing "
+                "affected Pulp Remote, adding a good one and resyncing.\n"
+                "If the problem persists, please contact the Pulp team."
```
mdellweg, line 1163: What you basically wanted to say is "Don't call us before you checked all the above!", right? But actually I would just not say anything that encourages people to complain.

```suggestion
```

pulpcore/tests/functional/api/using_plugin/test_content_delivery.py
```diff
@@ -135,11 +134,94 @@ def test_remote_content_changed_with_on_demand(
 
     # THEN
     get_url = urljoin(distribution.base_url, expected_file_list[0][0])
-    with pytest.raises(ClientPayloadError, match="Response payload is not completed"):
-        download_file(get_url)
-
-    # Assert again with curl just to be sure.
```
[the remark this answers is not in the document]
mdellweg: If we added another call to curl right away, we would expect a different error, because the "only" RA is corrupted. right?
Would it be easy to add that to this test?

pulpcore/app/settings.py
pedro-psb: Cooldown sounds good.
And yes, it was a constant first (marked you above)

pedro-psb:
> If we added another call to curl right away, we would expect a different error, because the "only" RA is corrupted. right?

Yes, client will get a 404 because the RA will be ignored for the second attempt within the cooldown interval.

> Would it be easy to add that to this test?

Yes, I can do that.

pulpcore/app/settings.py
mdellweg: Yes, let's not ping pong it.

pulpcore/app/settings.py
```diff
@@ -298,7 +298,7 @@
 # The time a RemoteArtifact will be ignored after failure.
 # In on-demand, if a fetching content from a remote failed due to corrupt data,
 # the corresponding RemoteArtifact will be ignored for that time (seconds).
-FAILED_REMOTE_ARTIFACT_PROTECTION_TIME = 5 * 60  # 5 minutes
+FAILED_REMOTE_ARTIFACT_COOLDOWN_TIME = 5 * 60  # 5 minutes
```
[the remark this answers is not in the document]
pedro-psb: What about:

1. `REMOTE_CONTENT_FETCH_FAILURE_COOLDOWN`
2. `FAILED_REMOTE_CONTENT_FETCH_COOLDOWN`
3. `REMOTE_CONTENT_FETCH_COOLDOWN`.
4. `ON_DEMAND_CONTENT_FETCH_COOLDOWN`
5. `ON_DEMAND_REMOTE_COOLDOWN`
 
IHO the "content/artifact fetch" here improves the scoping of that.
Those should generally read as  "cooldown for when content fetching from a remote fails".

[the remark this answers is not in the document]
mdellweg: `REMOTE_CONTENT_FETCH_FAILURE_COOLDOWN` is still a mouth full, but probably the best of all the suggestions we came up with. (Being a setting means we cannot easily change it later.)

mdellweg (commented):
All my concerns have been addressed. Time for squashing?

ggainey (approved):
Looks good - tests look good to exercise the new codepaths, great discussion.

dralley (approved):

mdellweg (approved):

PR #6078, opened 2024-11-26
Fix GetOrCreateSerializerMixin not working with pulp_domain

gerrod3:
pulp_domain is not processed in model's serializer since the value is auto-set as a context var on request time. This helper method would run into serializer errors if pulp_domain was specified as a natural key (which you need to do when supporting domains).

### DIFF
--- a/pulpcore/tests/unit/serializers/test_base.py
+++ b/pulpcore/tests/unit/serializers/test_base.py
@@ -2,7 +2,13 @@ import pytest
 
 from rest_framework import serializers
 
-from pulpcore.app.serializers import validate_unknown_fields
+from pulpcore.app.serializers import (
+    validate_unknown_fields,
+    RBACContentGuardSerializer,
+    GetOrCreateSerializerMixin,
+)
+from pulpcore.app.models import RBACContentGuard
+from pulpcore.app.util import get_domain
 
 
 def test_unknown_field():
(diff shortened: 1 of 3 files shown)


pulpcore/tests/unit/serializers/test_base.py
```diff
@@ -108,3 +115,27 @@ def test_ignored_fields_no_side_effects():
     initial_data = {"field1": 1, "csrfmiddlewaretoken": 2}
     defined_fields = {"field1": 1}
     validate_unknown_fields(initial_data, defined_fields)
+
+
+class TestGetOrCreateSerializerMixin(TestCase):
```
mdellweg, line 120: Can you try if this creates valid pytest tests?
```suggestion
class TestGetOrCreateSerializerMixin():
```

mdellweg: Looks like the setUp function may not be called with that name.
Also pytest really suggests to move to fixtures.

mdellweg (approved):

pulpcore/tests/unit/serializers/test_base.py
mdellweg: I think this doc applies here.
https://pytest-django.readthedocs.io/en/latest/helpers.html#db

gerrod3: Thanks, rewrote it the fixture way.

mdellweg (approved):

dralley:
I don't quite understand why `pulp_domain` is ever part of the natural key instead of (always) being pulled from the environment when needed at the site of the query?

dralley:
If we're backporting this, it would be nice to have an issue filed

gerrod3:
> If we're backporting this, it would be nice to have an issue filed

Can this PR be the issue? Don't feel that we will get much benefit out of creating a duplicate issue that has no new information.

Issue #5937, opened 2024-10-25
[PULP-208] Content app shouldnt retry on RA streaming


pedro-psb:
Our RemoteArtifact streaming uses chunked transfer, so once it has started a response, we can't fix anything anymore in case an error.

Because of that, the retry logic (any of the timeout, digest, etc) on the http downloader is just delaying a response that can't be possible right anymore.

We should add a mechanism to skip retries when calling from the chunked streaming context.

Relevant code:
- Retries on downloader: https://github.com/pulp/pulpcore/blob/24dac1b0b7d01bf9d9633aece3d3e9ad8380d13c/pulpcore/download/http.py#L230-L250
- Download call from stream_remote_artifact: https://github.com/pulp/pulpcore/blob/24dac1b0b7d01bf9d9633aece3d3e9ad8380d13c/pulpcore/content/handler.py#L1119-L1126

## Context

Identified this when investigating https://github.com/pulp/pulpcore/issues/5012 and https://github.com/pulp/pulpcore/issues/5725

PR #6087, opened 2024-11-27
[PULP-208] Disable download retry with on demand streaming

pedro-psb:
https://github.com/pulp/pulpcore/issues/5937

## Testing

I dont know how to write an automated test for that, but I've asserted it through logs by running the `test_remote_content_changed_with_on_demand`, which causes a digest validation error in the on-demand streaming context.

**Before**:

```
[pulp]  | (OMITED) "GET /pulp/api/v3/distributions/file/file/01936f14-21af-7102-b69b-65a30d34596e/ HTTP/1.0" 200 631 "-" "api/using_plugin/test_content_delivery.py::test_remote_content_changed_with_on_demand"',)

[pulp]  | Backing off download_wrapper(...) for 0.8s (pulpcore.exceptions.validation.DigestValidationError: A file located at the url https://127.0.0.1:44247/basic/2.iso failed validation due to checksum. Expected '89a0fbc35e07fe70fb46467326b6ab6a5bc53da48d518f3a6c6ef17dc3fab256', Actual '22f92924b8fef320f9933b196339a5dc5f1f662a27b09a3da9f2d73ce4678aa1')
[pulp]  | pulp [None]: backoff:INFO: Backing off download_wrapper(...) for 0.8s (pulpcore.exceptions.validation.DigestValidationError: A file located at the url https://127.0.0.1:44247/basic/2.iso failed validation due to checksum. Expected '89a0fbc35e07fe70fb46467326b6ab6a5bc53da48d518f3a6c6ef17dc3fab256', Actual '22f92924b8fef320f9933b196339a5dc5f1f662a27b09a3da9f2d73ce4678aa1')
[pulp]  | Backing off download_wrapper(...) for 0.2s (pulpcore.exceptions.validation.DigestValidationError: A file located at the url https://127.0.0.1:44247/basic/2.iso failed validation due to checksum. Expected '89a0fbc35e07fe70fb46467326b6ab6a5bc53da48d518f3a6c6ef17dc3fab256', Actual '22f92924b8fef320f9933b196339a5dc5f1f662a27b09a3da9f2d73ce4678aa1')
[pulp]  | pulp [None]: backoff:INFO: Backing off download_wrapper(...) for 0.2s (pulpcore.exceptions.validation.DigestValidationError: A file located at the url https://127.0.0.1:44247/basic/2.iso failed validation due to checksum. Expected '89a0fbc35e07fe70fb46467326b6ab6a5bc53da48d518f3a6c6ef17dc3fab256', Actual '22f92924b8fef320f9933b196339a5dc5f1f662a27b09a3da9f2d73ce4678aa1')
[pulp]  | Backing off download_wrapper(...) for 1.6s (pulpcore.exceptions.validation.DigestValidationError: A file located at the url https://127.0.0.1:44247/basic/2.iso failed validation due to checksum. Expected '89a0fbc35e07fe70fb46467326b6ab6a5bc53da48d518f3a6c6ef17dc3fab256', Actual '22f92924b8fef320f9933b196339a5dc5f1f662a27b09a3da9f2d73ce4678aa1')
[pulp]  | pulp [None]: backoff:INFO: Backing off download_wrapper(...) for 1.6s (pulpcore.exceptions.validation.DigestValidationError: A file located at the url https://127.0.0.1:44247/basic/2.iso failed validation due to checksum. Expected '89a0fbc35e07fe70fb46467326b6ab6a5bc53da48d518f3a6c6ef17dc3fab256', Actual '22f92924b8fef320f9933b196339a5dc5f1f662a27b09a3da9f2d73ce4678aa1')
[pulp]  | Giving up download_wrapper(...) after 4 tries (pulpcore.exceptions.validation.DigestValidationError: A file located at the url https://127.0.0.1:44247/basic/2.iso failed validation due to checksum. Expected '89a0fbc35e07fe70fb46467326b6ab6a5bc53da48d518f3a6c6ef17dc3fab256', Actual '22f92924b8fef320f9933b196339a5dc5f1f662a27b09a3da9f2d73ce4678aa1')
[pulp]  | pulp [None]: backoff:ERROR: Giving up download_wrapper(...) after 4 tries (pulpcore.exceptions.validation.DigestValidationError: A file located at the url https://127.0.0.1:44247/basic/2.iso failed validation due to checksum. Expected '89a0fbc35e07fe70fb46467326b6ab6a5bc53da48d518f3a6c6ef17dc3fab256', Actual '22f92924b8fef320f9933b196339a5dc5f1f662a27b09a3da9f2d73ce4678aa1')

[pulp]  | [2024-11-27 19:23:34 +0000] [44606] [ERROR] Error handling request
```

**After**:

```
[pulp]  | (OMITED) "GET /pulp/api/v3/distributions/file/file/01936f15-b88b-770a-b1fe-16e47a99c3bf/ HTTP/1.0" 200 631 "-" "api/using_plugin/test_content_delivery.py::test_remote_content_changed_with_on_demand"',)

[pulp]  | [2024-11-27 19:25:15 +0000] [44813] [ERROR] Error handling request
```

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -1138,7 +1138,9 @@ class Handler:
         original_finalize = downloader.finalize
         downloader.finalize = finalize
         try:
-            download_result = await downloader.run()
+            download_result = await downloader.run(
+                extra_data={"disable_retry_list": (DigestValidationError,)}
+            )
         except DigestValidationError:
             await downloader.session.close()
             close_tcp_connection(request.transport._sock)
--- a/pulpcore/download/http.py
+++ b/pulpcore/download/http.py
@@ -214,20 +214,22 @@ class HttpDownloader(BaseDownloader):
 
     async def run(self, extra_data=None):
         """
-        Run the downloader with concurrency restriction and retry logic.
+        Run the downloader with concurrency restriction and optional retry logic.
 
         This method acquires `self.semaphore` before calling the actual download implementation
         contained in `_run()`. This ensures that the semaphore stays acquired even as the `backoff`
         wrapper around `_run()`, handles backoff-and-retry logic.
 
         Args:
-            extra_data (dict): Extra data passed to the downloader.
+            extra_data (dict): Extra data passed to the downloader:
+                disable_retry_list: List of exceptions which should not be retried.
 
         Returns:
             [pulpcore.plugin.download.DownloadResult][] from `_run()`.
 
         """
-        retryable_errors = (
+        disable_retry_list = [] if not extra_data else extra_data.get("disable_retry_list", [])
+        default_retryable_errors = (
             aiohttp.ClientConnectorSSLError,
             aiohttp.ClientConnectorError,
             aiohttp.ClientOSError,
@@ -240,6 +242,9 @@ class HttpDownloader(BaseDownloader):
             SizeValidationError,
         )
 
+        retryable_errors = tuple(
+            [e for e in default_retryable_errors if e not in disable_retry_list]
+        )
         async with self.semaphore:
 
             @backoff.on_exception(
--- a/CHANGES/5937.bugfix
+++ b/CHANGES/5937.bugfix
@@ -1,0 +1,2 @@
+Disable retry logic on the context of content-app on-demand streaming, as we can't recover
+from any errors after starting the streaming process (chunked transfer).


pulpcore/download/http.py
pedro-psb, line 1: I'm using `extra_data` instead of creating a new keyword arg `retry=True` because adding a new kwarg breaks some parts of the test. So using `extra_data` is the less friction path.

Also, it looks like that's the purpose of `extra_data` anyway.

pulpcore/download/http.py
```diff
@@ -240,14 +240,23 @@ async def run(self, extra_data=None):
...
                 retryable_errors,
                 max_tries=self.max_retries + 1,
                 giveup=http_giveup_handler,
             )
+        else:
+            backoff_decorator = noop_decorator
+
+        async with self.semaphore:
+
+            @backoff_decorator
```
mdellweg, line 259: Aren't there exceptions we would still like to retry with, like connection errors before we even started sending headers back to the client?
Also as a matter of style, I'd rather conditionally apply the decorator (`func = decorator(func)`) instead of creating a noop-decorator.

pedro-psb: > Aren't there exceptions we would still like to retry with, like connection errors before we even started sending headers back to the client?

Yes, you are right.
So I guess the approach here could be passing what exceptions we want to retry (or  exclude) from the [default list](https://github.com/pulp/pulpcore/blob/df3b56d9e69dff8c4881ca4cbbff2a679b586ac1/pulpcore/download/http.py#L230). IHO exclude is more expressive for the probem.
Wdyt?

pulpcore/download/http.py
```diff
@@ -240,6 +242,9 @@ async def run(self, extra_data=None):
             SizeValidationError,
         )
 
+        retryable_errors = tuple(
+            [e for e in default_retryable_errors if e not in disable_retry_list]
+        )
```
pedro-psb, line 247: It has to be a tuple or a single exception.
Passing anything else raises a [not-helpful error](https://github.com/litl/backoff/issues/194).

pulpcore/content/handler.py
```diff
@@ -1138,7 +1138,9 @@ async def finalize():
         original_finalize = downloader.finalize
         downloader.finalize = finalize
         try:
-            download_result = await downloader.run()
+            download_result = await downloader.run(
+                extra_data={"disable_retry_list": (DigestValidationError,)}
```
mdellweg, line 1142: Thinking about "code in the domain of the problem", I'm wondering if we should just say "This downloader is streaming" here, and let the downloader.run decide decide how that impacts retrying.
At least we would have all the determination of what is retryable and what is not in one place.
`run` _is_ an external interface, right? So we kind of need to do it right now...

pedro-psb: It feels like the downloader is supposed to be a more generic object.
For example, the downloader doesnt know about the streaming we are doing here so we need to explicit patch and add conditions to handle it as this context requires. If we want to shift approaches (make the downloader context-aware), we should move these to the downloader aswell and make it a bit more complex.

mdellweg: fair point.
Actually we monkeypatch the downloader here.

mdellweg (approved):

dralley (approved):

pulpcore/content/handler.py
dralley: Yes the downloader code is a mess already so this doesn't make it particularly worse.  It would be nice to specialize the downloader object in such a way that we don't need *any* of this directly at the call site.  But that's a more involved refactor, so I'd consider it out of scope for a bugfix.  Maybe something for the future.

PR #6100, opened 2024-12-02
Always add core version to the api spec

mdellweg:
This is supposed to help select proper bindings generator based on the version of pulpcore used.


pedro-psb (approved):
Nice! :tada: 

I wonder if it wont be confusing in the future to have it on the `x-pulp-app-versions` "even if core apis are not part of the spec.". But not sure if it's worth creating a separate field to convey the core version that it was created with.

mdellweg:
I was wondering the same...

PR #6110, opened 2024-12-05
[PULP-220] Purge migrations

mdellweg:


ggainey:
Is this really a 3.70 blocker? I thought we'd talked about not doing too many changes at once in this area for 3.70?

dralley:
I probably would not drop the old migrations until at least after we've done so for the plugins.

mdellweg:
No not necessarily a blocker. But this _is_ the change, why we request all plugins to rebase their migrations.

dralley:
@mdellweg Can this proceed now?

mdellweg:
> @mdellweg Can this proceed now?

It can proceed any time. The question is when do we want to take the risk?

dralley:
I think we should do it

mdellweg:
Well anytime is not exactly right. We must align it with a breaking change release.

PR #6111, opened 2024-12-05
Fix openapi schema for created resources

mdellweg:


mdellweg:
This is really unfortunate. The old bindings (v4.3.1) seem to be unable to consume the fix that the new ones (v7.*) need.
How am I now supposed to roll out this fix?

pedro-psb (commented):
I know nothing about the issue, can you provide some context/links?

mdellweg:
> Why the old bindings cant consume the fix? The tests are passing here and it uses the old generator version, doesnt it?

Sorry, that was an old approach to fixing that didn't work. I tried to fix the api spec to report `{"oneOf":[{"type": "null"},{"type": "string", "format": "uri"}` which it didn't like. So i went for fixing the api to comply with the existing spec.

pulpcore/app/serializers/task.py
```diff
@@ -46,8 +46,7 @@ class TaskSerializer(ModelSerializer):
     finished_at = serializers.DateTimeField(
         help_text=_("Timestamp of when this task stopped execution."), read_only=True
     )
-    error = serializers.DictField(
-        child=fields.JSONDictField(),
```
pedro-psb, line 50: There is a difference in the structure declaration here:

```python
# DictField with child=JSONDictFeild
{
    "key1": { ... },
    ...
    "keyk": { ... }
}

# JSONDictField
{
    "key1": Any,
    ...
    "keyk": Any
}
```

I think I would keep the DictField. The JSONDictField is meant to replace mainly the generic JSONFields where applicable. I dont see a strong reason for doing so here. Wdyt?

mdellweg: The thing is that the errors as reported by pulp seem to be dicts of any. I'll try and dig up examples.

mdellweg: Here you are: It's a dict of mainly strings.

```python
[task.error for task in Task.objects.filter(state="failed")]
[{'traceback': '  File "/usr/local/lib/python3.9/site-packages/pulpcore/tasking/tasks.py", line 68, in _execute_task\n
[...]
```

pedro-psb (approved):

pulpcore/app/serializers/base.py
```diff
@@ -208,7 +208,7 @@ def to_representation(self, data):
                 if repo_ver_mapping is not None:
                     if repo_ver := repo_ver_mapping.get(data.object_id):
                         return self.repo_ver_url(repo_ver, request=request)
-                    return None
+                    return "<unavailabe>"
```
ggainey, line 211:
```suggestion
                    return "<unavailable>"
```

pulpcore/app/serializers/base.py
```diff
@@ -229,10 +229,10 @@ def to_representation(self, data):
         # Fallback onto normal lookup:
         # If the content object was deleted
         if data.content_object is None:
-            return None
+            return "<unavailabe>"
```
ggainey, line 232:
```suggestion
            return "<unavailable>"
```

ggainey (changes requested):

ggainey (approved):

Issue #5929, opened 2024-10-23
Immediate async tasks are not awaited


mdellweg:
SSIA

gerrod3:
How so, am I missing something it looks like they are: https://github.com/pulp/pulpcore/blob/21f2c2179ec40bca55b24e652b87c249bf3cbaeb/pulpcore/tasking/tasks.py#L69-L72

pedro-psb:
Hmm, if you see the line above, the actual task code (`func`) is called as a sync function.
This is checking the task *return value* is a coroutine. I didn't know we even did something with the task fn return value.
Or that's how async tasks are supposed to be? Like wrapped in a sync function.

Also, do you know any task defined as an async coroutine? I feel like Mathias once said that there were none, really.

https://github.com/pulp/pulpcore/blob/21f2c2179ec40bca55b24e652b87c249bf3cbaeb/pulpcore/tasking/tasks.py#L68-L72

gerrod3:
```python
>>> import asyncio
>>> async def hello():
...     print("hello")
...     
>>> a = hello()
>>> asyncio.iscoroutine(a)
True
>>> loop = asyncio.get_event_loop()
<python-input-4>:1: DeprecationWarning: There is no current event loop
  loop = asyncio.get_event_loop()
>>> loop = asyncio.new_event_loop()
>>> loop.run_until_complete(a)
hello
```
With my current understanding of asyncio I think we have valid handling code, the async code won't run till put into an event loop. Our only potential problem is that I don't know what the status of the current event loop is in the api worker, so we might be calling deprecated code (deprecated in python 3.12 to call get_event_loop without a running loop).

pedro-psb:
Oh, I see, thanks for clarifying! Asyncio is not in my heart yet 😅

mdellweg:
First, yes, we don't have any async tasks yet. But also I cannot see what tipped me off back then. Maybe it got fixed in the meantime... I'm OK with closing this one.

mdellweg:
OK, i think i found what buged me again:
"iscoroutine" actually tests for something "awaitable". "iscoroutinefunction" is the one we should be using before performing the task.

```python
In [1]: async def hello():
   ...:     print("hello")
   ...: 

In [2]: import asyncio

In [3]: asyncio.iscoroutine(hello)
Out[3]: False

In [4]: asyncio.iscoroutinefunction(hello)
Out[4]: True
```

PR #6155, opened 2024-12-16
[PULP-218] Add timeout to immediate tasks

pedro-psb:


pulpcore/tasking/tasks.py
```diff
@@ -47,10 +49,28 @@ def wakeup_worker():
...
 
+def execute_immediate_task(task):
+    # set alarm timeout
+    IMMEDIATE_TASK_TIMEOUT = 60 * 5
+
+    def timeout_handler(signum, frame):
+        raise TimeoutException(f"Immediate task time: {IMMEDIATE_TASK_TIMEOUT}")
+
+    signal.signal(signal.SIGALRM, timeout_handler)
+    signal.alarm(IMMEDIATE_TASK_TIMEOUT)
```
mdellweg, line 63: Oh, interesting.
This is quite close to low level process handling. I wonder if this might interfere with other uses of the signal handler.

Do you think we can instead require that all immediate tasks are async functions? We could then implement the timeout on the asyncio level with much more confidence.

pedro-psb: > Do you think we can instead require that all immediate tasks are async functions? We could then implement the timeout on the asyncio level with much more confidence.

Yes. I dont have a good idea of how much friction this imposes, but that's definitely more safe.

pedro-psb: @mdellweg Hey! Can you provide an early review for the api worker part?

pulpcore/tasking/tasks.py
```diff
@@ -69,10 +69,24 @@ def _execute_task(task):
...
+        if immediate is True and not is_coroutine:
+            raise RuntimeError(_("Immediate tasks must be coroutines."))
+
+        if is_coroutine:
             _logger.debug(_("Task is coroutine %s"), task.pk)
             loop = asyncio.get_event_loop()
-            loop.run_until_complete(result)
+            if immediate:
+                try:
+                    loop.run_until_complete(asyncio.wait_for(result, timeout=IMMEDIATE_TIMEOUT))
```
mdellweg, line 83: Looks like a good first cut. But we still stop the heartbeat loop for up to 5 seconds here.

We should definitely plan on rewriting the workers in async.

pedro-psb: > We should definitely plan on rewriting the workers in async.

Would you say that's a blocker (for running on foreground)?

I was trying to find a way to integrate the coroutine control yielding with the select-loop, but I'm reaching dark corners.

If that's a blocker this PR could be really about the timeout (as originally intented). The commit "wip: add timeout to supervise immediate tasks" adds a timeout to the regular task process. Maybe we could stop there.

mdellweg: Not necessarily. It's a certain risk. But it's nothing we cannot revert when things go bad.

mdellweg: Maybe after all it's wise to split the work into smaller chunks to reduce that risk.
Ensuring that immediate tasks don't run too long is still an incremental improvement.

pedro-psb:
No idea why this fails only on s3:
https://github.com/pulp/pulpcore/actions/runs/13333279626/job/37242548757?pr=6155#step:14:18219

An exception does get caught...

pulpcore/constants.py
```diff
@@ -45,6 +45,8 @@
 #: Tasks in an incomplete state have not finished their work yet.
 TASK_INCOMPLETE_STATES = (TASK_STATES.WAITING, TASK_STATES.RUNNING, TASK_STATES.CANCELING)
 
+#: Timeout for immediate tasks
```
mdellweg, line 48: Is that seconds?

pulpcore/tasking/tasks.py
```diff
@@ -50,7 +52,6 @@ def wakeup_worker():
 
 
 def execute_task(task):
-    # This extra stack is needed to isolate the current_task ContextVar
```
mdellweg, line 53: Why drop this comment?

pulpcore/tasking/tasks.py
```diff
@@ -69,10 +70,33 @@ def _execute_task(task):
         args = task.enc_args or ()
         kwargs = task.enc_kwargs or {}
         result = func(*args, **kwargs)
-        if asyncio.iscoroutine(result):
+        immediate = task.immediate
+        is_coroutine = asyncio.iscoroutine(result)
```
mdellweg, line 74: Has this ever been true? Should we no check if `func` is a coroutine even before executing it?

pulpcore/tasking/worker.py
```diff
@@ -388,6 +390,7 @@ def supervise_task(self, task):
         task.save(update_fields=["worker"])
         cancel_state = None
         cancel_reason = None
+        time_limit = time.monotonic() + IMMEDIATE_TIMEOUT
```
mdellweg, line 393: Why do we not use `timezone.now` here?

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -445,3 +446,125 @@ def test_cancel_task_group(pulpcore_bindings, dispatch_task_group, gen_user):
...
+
+    def dispatch_long_task(requires_resource):
+        task_href = dispatch_task(
+            "pulpcore.app.tasks.test.sleep",
+            args=(LONG_RUNNING,),
+            exclusive_resources=[requires_resource],
+        )
+        wait_until("running", task_href)
+
+    # Case 1
```
mdellweg, line 536: Each case should be a separate test. If you want to share resources, you can group them in a class.

pulpcore/tasking/worker.py
mdellweg: Also didn't we want to let the timeout tick in the async loop?
That will help us maybe run even more than one immediate tasks per worker eventually.
Also the timeout will be in effect even on the api worker.

mdellweg (changes requested):
Not sure if you wanted a review at this stage...

pulpcore/tasking/tasks.py
pedro-psb: Gerrod showed that this is valid [here](https://github.com/pulp/pulpcore/issues/5929#issuecomment-2634836859). But I can look up a more intuitive form to check it.

pulpcore/tasking/worker.py
pedro-psb: Oh, that's true :facepalm: 
The worker ends up calling execute_task, so we only need it there with the async loop...

pulpcore/tasking/tasks.py
pedro-psb: I guess that was an accident

pulpcore/constants.py
pedro-psb: Yes, I'll add that on the description

pulpcore/tasking/tasks.py
mdellweg: Absolutely not. But then it doesn't seem to be telling its story.

pulpcore/tasking/tasks.py
mdellweg: The big problem is that we run the task function first and then we check if it returned something awaitable.
(That is what coroutine functions are: They are functions returning an awaitable generator object.)

While in the old code this may be working properly (unless some task accidentally returned a generator of any sort), here we don't want to run the task function in case it is immediate and not async.

So what should happen, i think, is we should check for `asyncio.iscoroutinefunction(func)` and then decide whether or how to execute it.

mdellweg: Sadly the python documentation itself uses the term "coroutine" a bit inconsistently.
Sometimes as the function declared with async, and sometimes as the awaitable object that is the result of calling the former.

pulpcore/tasking/tasks.py
pedro-psb: I mean dropping the comment was an accident.
The comment is fine. I didnt know what a ContextVar was when I first read it, but then I read about it an it made sense.

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -525,46 +515,76 @@ def wait_until(state, task_href, timeout=10):
...
+    def test_throws_on_timeout(
+        self, dispatch_long_task, pulpcore_bindings, dispatch_task, monitor_task
+    ):
+        """
+        GIVEN an async task requiring busy resources
+        WHEN dispatching a task as immediate
+        AND it takes longer than timeout
+        THEN an error is raised
+        """
+        COMMON_RESOURCE = "PPP"
```
mdellweg, line 580: I'd make the resource a random string to avoid any collisions.
Also why not mark this as parallel too?

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -525,46 +515,76 @@ def wait_until(state, task_href, timeout=10):
...
+
+    @pytest.mark.parallel
+    def test_executes_in_task_worker(self, dispatch_long_task, dispatch_task, monitor_task):
+        """
+        GIVEN an async task requiring busy resources
+        WHEN dispatching a task as immediate
+        THEN the task completes with a worker
+        """
+        COMMON_RESOURCE = "MMM"
+        dispatch_long_task(required_resources=[COMMON_RESOURCE])
```
mdellweg, line 539: You could cancel the long task after dispatching the regular one in order to speed up the tests dramatically.

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -448,74 +448,64 @@ def test_cancel_task_group(pulpcore_bindings, dispatch_task_group, gen_user):
...
 
-    # Case 1
-    task_href = dispatch_task("pulpcore.app.tasks.test.asleep", args=(LT_TIMEOUT,), immediate=True)
-    task = monitor_task(task_href)
-    assert task.state == "completed"
-    assert task.worker is None
 
-    # Case 2
-    with pytest.raises(PulpTaskError) as ctx:
+class TestImmediateTaskNoLocking:
```
mdellweg, line 455: TestImmediateTaskWithNoResource?

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -448,74 +448,64 @@ def test_cancel_task_group(pulpcore_bindings, dispatch_task_group, gen_user):
...
+
+    @pytest.mark.parallel
+    def test_succeed_on_api_worker(self, pulpcore_bindings, dispatch_task, monitor_task):
```
mdellweg, line 458: This is really nit picky...
test_succeeds_on_api_worker

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -448,74 +448,64 @@ def test_cancel_task_group(pulpcore_bindings, dispatch_task_group, gen_user):
...
         )
-        monitor_task(task_href)
-    assert "task timed out after" in ctx.value.task.error["description"]
-
-
-@pytest.mark.parallel
-def test_immediate_task_requires_resource(pulpcore_bindings, dispatch_task, monitor_task):
-    """
-    GIVEN an async task requiring busy resources
+        task = monitor_task(task_href)
```
mdellweg, line 486: Should we monitor that task? Should it not return as completed already from dispatch_task? Also we should be able to catch the deprecation warning here and assert on it.
Also we don't want it to bubble into the collected deprecation warnings for the ci.

pulpcore/tasking/tasks.py
```diff
@@ -69,27 +70,29 @@ def _execute_task(task):
...
-            is_coroutine = True
+                func = sync_to_async(func)
+                is_coroutine_fn = True
+            else:
+                func(*args, **kwargs)
 
-        if is_coroutine:
+        if is_coroutine_fn:
             _logger.debug(_("Task is coroutine %s"), task.pk)
+            coro = func(*args, **kwargs)
```
mdellweg, line 92: Not sure if it's wise to call this thing coro given the confusion with that name throughout the python docs.
The only idea i can come up with is "awaitable".

pulpcore/tasking/tasks.py
```diff
@@ -69,27 +70,29 @@ def _execute_task(task):
...
             if immediate:
-                coro = asyncio.wait_for(result, timeout=IMMEDIATE_TIMEOUT)
-            else:
-                coro = result
-
+                coro = asyncio.wait_for(coro, timeout=IMMEDIATE_TIMEOUT)
+            assert asyncio.iscoroutine(coro), coro
```
mdellweg, line 95: Extra safety? (I'm almost certain we can be sure of that now...)

pulpcore/pytest_plugin.py
```diff
@@ -993,35 +993,57 @@ def _wget_recursive_download_on_host(url, destination):
...
     return _dispatch_task
 
 
+@pytest.fixture(scope="session")
+def dispatch_task_with_stderr(pulpcore_bindings):
+    """Variation of dispatch_task that returns (task_href, stderr_content).
+
+    The stderr can be useful write asserts over log messages.
+    This is required because the pytest caplog fixture can't capture log messages
+    streamed from the pulpcore-manager call.
```
mdellweg, line 1033: ohhhh!

Now I'm wondering if we could just use the capsys or capfd fixture instead? (If you _want_ to try that.)

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -445,3 +447,158 @@ def test_cancel_task_group(pulpcore_bindings, dispatch_task_group, gen_user):
...
+                args=(LT_TIMEOUT,),
+                immediate=True,
+                exclusive_resources=[COMMON_RESOURCE],
+            )
+        task = monitor_task(task_href)
+        assert task.state == "completed"
+        assert task.worker is not None
+
+    @pytest.mark.parallel
+    def test_throws_when_non_deferrable(
```
mdellweg, line 562: This is not a newly defined behaviour. Can you do me a favor and see whether we already have a test for this?

(Ideally we combine them or pick the nicer one...)

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -445,3 +447,158 @@ def test_cancel_task_group(pulpcore_bindings, dispatch_task_group, gen_user):
...
+                deferred=False,
+                exclusive_resources=[COMMON_RESOURCE],
+            )
+            task = pulpcore_bindings.TasksApi.read(task_href)
+            assert task.state == "canceled"
+            assert task.worker is None
+            assert "Resources temporarily unavailable." in task.error["reason"]
+
+    @pytest.mark.parallel
+    def test_timeouts_on_task_worker(
```
mdellweg, line 585: Nitpick mode: `test_times_out_on_task_worker` reads more English to me.

pulpcore/tasking/tasks.py
```diff
@@ -68,11 +70,36 @@ def _execute_task(task):
...
             loop = asyncio.get_event_loop()
-            loop.run_until_complete(result)
+            try:
+                loop.run_until_complete(coro)
+            except asyncio.TimeoutError:
+                raise RuntimeError(
+                    _("Immediate task timed out after {} seconds").format(IMMEDIATE_TIMEOUT)
+                )
```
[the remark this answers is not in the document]
pedro-psb: Yes, this was supposed to add more info.
But it'll be handled by the outer try/catch-all at L107 (from changed code) and receive all the context info.
Or am I missing something?

pulpcore/tests/functional/api/test_tasking.py
pedro-psb: Looks like we don't test it elsewhere: 

```
> git grep -I -n defer -- pulpcore/tests/
pulpcore/tests/functional/api/test_tasking.py:564:    def test_throws_when_non_deferrable(
pulpcore/tests/functional/api/test_tasking.py:569:        WHEN dispatching as immediate and not deferrable
pulpcore/tests/functional/api/test_tasking.py:578:                deferred=False,
```

pulpcore/tasking/tasks.py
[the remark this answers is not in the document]
mdellweg: Right. But i think we can still add something like:
`logger.warn("Immediate task {task.pk} timed out after ...")` So we see it in the worker and api worker logs too.

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -471,25 +471,22 @@ def test_succeeds_on_api_worker(self, pulpcore_bindings, dispatch_task):
...
-        task_href, stderr = dispatch_task_with_stderr(
+        task_href = dispatch_task(
             "pulpcore.app.tasks.test.sleep", args=(LT_TIMEOUT,), immediate=True
         )
+        stderr_content = capsys.readouterr().err
         task = pulpcore_bindings.TasksApi.read(task_href)
         assert task.state == "completed"
         assert task.worker is None
-        assert "Support for non-coroutine immediate tasks will be dropped" in stderr
+        assert "Support for non-coroutine immediate tasks will be dropped" in stderr_content
```
mdellweg, line 489: Thank you! Not adding a new fixture for this makes me happier.

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -445,3 +447,158 @@ def test_cancel_task_group(pulpcore_bindings, dispatch_task_group, gen_user):
...
+        raise RuntimeError("Timeout waiting for task to transition")
+
+    @contextmanager
+    def _dispatch_long_task(required_resources: list[str], duration=5):
+        task_href = dispatch_task(
+            "pulpcore.app.tasks.test.sleep",
+            args=(duration,),
+            exclusive_resources=required_resources,
+        )
+        wait_until("running", task_href)
```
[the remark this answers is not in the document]
mdellweg: No, that's not true. A task blocks a resource even in waiting state.

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -445,3 +447,160 @@ def test_cancel_task_group(pulpcore_bindings, dispatch_task_group, gen_user):
...
+        raise RuntimeError("Timeout waiting for task to transition")
+
+    @contextmanager
+    def _resource_blocker(exclusive_resources: list[str], duration=20):
+        task_href = dispatch_task(
+            "pulpcore.app.tasks.test.sleep",
+            args=(duration,),
+            exclusive_resources=exclusive_resources,
+        )
+        wait_until(TASK_INCOMPLETE_STATES, task_href)
```
mdellweg, line 526:
```suggestion
```
What do you want to wait for here?

pedro-psb: Yeah, the wait doesnt make sense anymore. Removing...

But I wanted to share some thoughts I had on this (not wanting to change anything in this PR):

The long task will always be unblocked first because [it has lower `pulp_created`](https://github.com/pulp/pulpcore/blob/25085248e5289e9713a20926440f37d32ca79c6f/pulpcore/tasking/worker.py#L259-L263), and the short is assured to be blocked as we intend in the test, right?

I was thinking if we shouldnt sort by immediate task there too (in the `unblock_task`). This is a violation of the event order, but so is our immediate task prioritization in `iter_tasks`. I can see there is a diffrence, and feels like it would be worse to violate that in unblocking-time than it is in picking-time, but that's not crystal clear to me.

mdellweg: Well the thing is the immediate task will not be unblocked before the other task is in a final state. And that's the only thing needed to assure resource usage is serialized. It does not matter whether the prior task is identified unblocked or in a waiting, running or canceling state.

PR #6161, opened 2024-12-18
[PR #6064/0a5ac4aa backport][3.63] [SAT-29018] Fix/corrupted RA blocks content streaming

### LINKED ISSUE #5725: Wrong package download after on-demand sync

pedro-psb:
On a request for on-demand content in the content app, a corrupted Remote that contains the wrong binary (for that content) prevented other Remotes from being attempted on future requests.

Now the last failed Remotes are temporarily ignored and others may be picked.

Closes #5725

(cherry picked from commit 0a5ac4aaf0f28b1055ee421a2ec35c65726039b0)


pedro-psb:
Can't backport because it contains migrations.

.github/workflows/scripts/script.sh
```diff
@@ -123,6 +123,13 @@ cmd_user_prefix bash -c "django-admin makemigrations core --check --dry-run"
 cmd_user_prefix bash -c "django-admin makemigrations file --check --dry-run"
 cmd_user_prefix bash -c "django-admin makemigrations certguard --check --dry-run"
 
+# See pulpcore.app.util.ENABLE_6064_BACKPORT_WORKAROUND for context.
+# This needs to be set here because it relies on service init.
+# Its being tested in only one scenario to have both cases covered.
+if [[ "$TEST" == "s3" ]]; then
+    cmd_prefix pulpcore-manager backport-patch-6064
+fi
+
```
mdellweg, line 132: This file is managed by the plugin template.
Can you maybe do it is a post_before_script hook?

pulpcore/app/apps.py
```diff
@@ -416,3 +421,14 @@ def _populate_artifact_serving_distribution(sender, apps, verbosity, **kwargs):
...
+
+def _conditional_6064_backport_workaround(sender, apps, verbosity, **kwargs):
+    # See context in pulpcore.app.util.ENABLE_6064_BACKPORT_WORKAROUND
+    from pulpcore.app.models import RemoteArtifact
+    from django.db import models
+    import pulpcore.app.util
+
+    if pulpcore.app.util.failed_at_exists(connection, RemoteArtifact):
+        pulpcore.app.util.ENABLE_6064_BACKPORT_WORKAROUND = True
+        RemoteArtifact.add_to_class("failed_at", models.DateTimeField(null=True))
```
mdellweg, line 434: What exactly does this do?

mdellweg: And something in this PR is altering the actual db table, so this can be used?
That feels like reinventing the whole db migrations framework with out the safeguards. A subsequent upgrade is then probably going to fail. When we said "You cannot backport a migration.", that meant you cannot add db altering code to a release branch assuming that all db alteration would be done by a migration in the django framework. My gut feeling is this is way too dangerous.
Can you think of a solution that does not require changing the db schema? We should be lucky by the fact that this is kind of ephemeral data.
- Would it help to keep a per-worker list (maybe a bloom filter) in memory?
- Can we repurpose another datetime field that we don't rely on there?
- Can we use redis?

pedro-psb: > A subsequent upgrade is then probably going to fail.

My assumption was that a field addition (that doesn't have any other couplings) would be safe. But I can see this is a sensitive area. I'll explore those alternatives.

(I had though of per-worker cache, but concluded it would be simpler to use the db - before knowing about the backport problem).

mdellweg: Thanks for understanding my concerns.
At this point I think postgres may even reject to apply the migration on top of this out of bounds change.
If I could choose, i'd prefer the per worker in memory caching solution. Even if it would only solve the problem half way.

pedro-psb: About the idea of repurposing another field, there is `pulp_created` and `pulp_last_updated`.
But I'm afraid of unexpected side-effects, like `pulp_last_updated` being updated by something and cooling down a good remote.
Or something else (thus, unexpected), because those are in the system for so long.

```
pulp_created         | timestamp with time zone |           | not null |
pulp_last_updated    | timestamp with time zone |           |          |
```

mdellweg: This is not about a Remote, but the RemoteArtifact, right? I'm not so concerned as this class is only used internally and never visible to the user. I highly doubt that we have any logic depending on it.

pedro-psb: Yes, RAs.
Well, I'll open a PR. If nothing bad happens during tests I think we are good.
Upgrading with the workaround of using another field name worked, but this would be preferable.

pedro-psb:
The re-purpose of `pulp_last_updated` seems to have worked. Replaced by https://github.com/pulp/pulpcore/pull/6196

PR #6171, opened 2025-01-02
Allow XMLHttpRequests to prevent password dialogs

mdellweg:
When presented a WWW-Authenticate header with "Basic", a browser will always ask the user for a password, before returning the 401 response to the web application issuing the XHR in the first place. Returning a fake authentication method instead seems to be the common workaround pattern.


pedro-psb (commented):
That implementation looks straightforward, but I don't get the motivation. Isn't that expected when using a browser client?

mdellweg:
The motivation is, that a web app can indicate this is an AJAX request. This is useful if you want the 401 response to be handed back to the JS part of the application right away (e.g. to redirect to a nice looking login page). Without this, the browser will find the "www-authenticate: Basic" and it will present its password dialog. There is no way to circumvent that on the client side.

pedro-psb (approved):
Thanks for clarifying.
LGMT

PR #6175, opened 2025-01-06
Makes CONTENT_ORIGIN not a required setting PULP-194

git-hyagi:


docs/admin/reference/settings.md
```diff
@@ -197,10 +196,11 @@ this setting only applies to uploads created after the change.
...
 ### CONTENT_ORIGIN
 
-A required string containing the protocol, fqdn, and port where the content app is reachable by
-users. This is used by `pulpcore` and various plugins when referring users to the content app.
+A string containing the protocol, fqdn, and port where the content app is reachable by users.
+This is used by `pulpcore` and various plugins when referring users to the content app.
 For example if the API should refer users to content at using http to pulp.example.com on port
-24816, (the content default port), you would set: `https://pulp.example.com:24816`.
+24816, (the content default port), you would set: `http://pulp.example.com:24816`. The default is `None`.
+When set to `None`, the API returns relative URLs (without the protocol, fqdn, and port).
```
pedro-psb, line 203: The specific information provided in the changes fragment is meaningful here:
E.g:

```diff
(...) For example if the API should refer users to content at using http to pulp.example.com on port
24816, (the content default port), you would set: `http://pulp.example.com:24816`.

+When set to `None`, the `base_url` for Distributions is a relative path.
+This means the API returns relative URLs without the protocol, fqdn, and port.

The default is `None`.
```

pedro-psb (changes requested):

docs/admin/reference/settings.md
mdellweg: Also please don't make non-ssl examples.

pedro-psb (approved):

mdellweg (changes requested):
Please make sure one of the test scenarios is using None on the setting.

docs/admin/reference/settings.md
```diff
@@ -1,10 +1,9 @@
 # Settings
 
-There are only two required settings, although specific plugins may have additional required
+There is one required settings, although specific plugins may have additional required
```
dralley, line 3: "settings" should be "setting" if we're referring to one.

bmbouter:
@pedro-psb this is really helpful, ty. This setting is defined earlier (L#1), and I'm not 100% sure what setting the same variable in two places in that would do. What do you think about us removing it from being set at L#1 and instead setting it on one or more of the other test runners?

Also, I don't know what you mean about "similar control pair", what does that mean.

pedro-psb:
Yeah, that sounds ideal, that shouldn't be set in the common settings part.

I was refering to the notion of experiment-control groups, where we have isolated differences being tested (and similar properties for everything else). But that was a bit of a loose reasoning, if something goes wrong it will probably be obvious.

bmbouter:
Thanks @mdellweg this is also helpful. Do you think we'll need to make any changes to the plugin template itself, or just update these and re-apply? Unforseen things can always come up, but I wanted to hear what you thought we'd run into.

mdellweg:
No template changes. That's what we have these variables for.

bmbouter:
@git-hyagi I think this ticket has enough info on it for you to proceed making the variable changes and re-applying the template. Please come back with any questions at all. Thank you!

template_config.yml
```diff
@@ -58,6 +58,7 @@ pulp_settings:
   tmpfile_protection_time: 10
   upload_protection_time: 10
 pulp_settings_azure:
+  content_origin: null
```
mdellweg, line 61: Exactly like this.
You will need to apply the plugin-template to kick this into effect however.

git-hyagi:
The CI is probably failing because we depend on a change in pulp-cli tests: https://github.com/pulp/pulp-cli/pull/1118

pulpcore/app/serializers/fields.py
```diff
@@ -311,7 +311,9 @@ class BaseURLField(serializers.CharField):
     """
 
     def to_representation(self, value):
-        origin = settings.CONTENT_ORIGIN.strip("/")
+        origin = "/"
+        if settings.CONTENT_ORIGIN:
+            origin = settings.CONTENT_ORIGIN.strip("/")
```
git-hyagi, line 316: when `CONTENT_ORIGIN == ""`(or None) we need to set `origin` as "/" so that the `base_url` will have the relative path like "_/some/file/path_", instead of "_some/file/path_"

ggainey: Can we have this as a comment here and at publication.py:731? "Clarity of intent" is a valuable comment

pulpcore/app/settings.py
```diff
@@ -262,8 +262,7 @@
 
 DRF_ACCESS_POLICY = {"reusable_conditions": ["pulpcore.app.global_access_conditions"]}
 
-# This is a sentinal value for deploy checks, since we don't want to set a default one.
-CONTENT_ORIGIN = "UNREACHABLE"
+CONTENT_ORIGIN = ""
```
ggainey, line 265: Add a comment "Empty string is allowed, but None is *not*"

template_config.yml
```diff
@@ -58,6 +58,7 @@ pulp_settings:
   tmpfile_protection_time: 10
   upload_protection_time: 10
 pulp_settings_azure:
+  content_origin: ""
```
ggainey, line 61: :+1:

ggainey (changes requested):

pulpcore/app/serializers/fields.py
mdellweg: If I read the feature correctly, set to `None` means the content is delivered relative to the api. So we should instead derive the CONTENT_ORIGIN from the incoming request (basically in the same way as the DRF serializer fields would do it when we didn't tell them not to) in order to not break the assumption of this api to return a URL.

dkliban (approved):

pulpcore/app/serializers/fields.py
dkliban: @mdellweg We considered such an implementation, but it doesn't work in all cases. When the API is deployed behind a gateway, the request that comes into the API may not have the original hostname that was used by the client. The hostname can be an internal hostname for the API pod/server. Such a hostname does not make sense to the client making the request.

pulpcore/app/replica.py
```diff
@@ -103,6 +103,16 @@ def create_or_update_remote(self, upstream_distribution):
...
         url = self.url(upstream_distribution)
+        if url.startswith("/"):
+            if self.server.content_origin:
+                url = self.server.content_origin + url
+            else:
+                raise Exception(
+                    "UpstreamPulp needs to have content_origin defined because the downstream Pulp"
+                    "does not know upstream's hostname from the distribution base_url (upstream"
+                    "Pulp does not have the content_origin defined)."
+                )
```
mdellweg, line 114:
```suggestion
            url = urljoin(self.server.base_url, url)
```

pulpcore/app/serializers/status.py
```diff
@@ -72,6 +72,9 @@ class ContentSettingsSerializer(serializers.Serializer):
 
     content_origin = serializers.CharField(
         help_text=_("The CONTENT_ORIGIN setting for this Pulp instance"),
+        allow_blank=True,
```
mdellweg, line 75:
```suggestion
        allow_blank=False,
```

General advice from DRF: Don't allow two ways to say N/A.

mdellweg:
I think, this still breaks the presigned URL's as it is atm.
What is the content of `request.url` and `request.host` on an installation that lives behind several reverse proxies/gateways?

git-hyagi:
> I think, this still breaks the presigned URL's as it is atm. What is the content of `request.url` and `request.host` on an installation that lives behind several reverse proxies/gateways?

We ran some tests connecting to different addresses and here are the results: https://gitlab.cee.redhat.com/hyagi/app-interface/-/blob/b770301c3f5b895c58a289196fde09ed0157559e/docs/pulp/architecture.md#routes
In summary:
* test pointing to cert.console.redhat.com
```
request.host: pulp-api.pulp-prod.svc.cluster.local:8000
request.url: http://pulp-api.pulp-prod.svc.cluster.local:8000/api/pulp/api/v3/status/
```

* test pointing to console.redhat.com
```
request.host: pulp-api.pulp-prod.svc.cluster.local:8000
request.url: http://pulp-api.pulp-prod.svc.cluster.local:8000/api/pulp/api/v3/status/
```

* mtls.internal.console.redhat.com
```
request.host: internal.console.redhat.com
request.url: http://internal.console.redhat.com/api/pulp/api/v3/status/
```

ggainey (approved):

mdellweg:
References:
https://docs.djangoproject.com/en/4.2/ref/request-response/#django.http.HttpRequest.get_host
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host

ggainey:
> References:
> https://docs.djangoproject.com/en/4.2/ref/request-response/#django.http.HttpRequest.get_host
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host

Can we rely on X-Forward-Host being there? If so - would that let us "fix" the semantics back to "we can always return relative-paths for base_url"l?  I'll happily close my pulp_rpm draft reacting to relpaths if so!

git-hyagi:
> _Can we rely on X-Forward-Host being there? If so - would that let us "fix" the semantics back to "we can always return full-paths for base_url"l? I'll happily close my pulp_rpm draft reacting to relpaths if so!_

We reran the tests, this time collecting the request headers and here are the results: https://gitlab.cee.redhat.com/service/app-interface/-/blob/236d8458341936a4961ec7fa6113fbb86a82376e/docs/pulp/architecture.md#routes

Even though in all tests the `x-forwarded-host` were present, they still did not provide the _expected_ host:
```
$ curl -iv https://cert.console.redhat.com/api/pulp/api/v3/status/
"HTTP_HOST": "pulp-api.pulp-prod.svc.cluster.local:8000"
"HTTP_X_FORWARDED_HOST": "cloud.redhat.com"
"HTTP_X_RH_EDGE_HOST": "cert.console.redhat.com"
```

```
$ curl -iv https://console.redhat.com/api/pulp/api/v3/status/
"HTTP_HOST": "pulp-api.pulp-prod.svc.cluster.local:8000"
"HTTP_X_FORWARDED_HOST": "cloud.redhat.com"
"HTTP_X_RH_EDGE_HOST": "console.redhat.com"
```

```
$ curl -ivL --cert hyagi.crt --key hyagi.key https://mtls.internal.console.redhat.com/api/pulp/api/v3/status/
"HTTP_X_FORWARDED_HOST": "internal.console.redhat.com"
"HTTP_HOST": "internal.console.redhat.com"
"HTTP_X_RH_EDGE_HOST": "mtls.internal.console.redhat.com"
```

ggainey:
> Even though in all tests the `x-forwarded-host` were present, they still did not provide the _expected_ host:

And HTTP_X_RH_EDGE_HOST is the one we "expect", yeah?

ggainey:
Consensus from the core-team-mtg 28-Jan summarized:

- Nobody is 100% happy with "only relative paths" being an option
- BUT - CONTENT_ORIGIN, as it stands, really does break a real-world usecase for Services
- SO - the current state of this PR (allow CONTENT_ORIGIN to be null, adapt plugins as required *where we can*) is accepted as the "current plan"
- Some plugin functionality *may not work* in that context
- SO - we will continue investigating other/better alternatives - examples:
  - allow a new setting "request header to depend on" for building FQDNs 
  - investigate making better use of Django's built-in support for this use-case

tl;dr: This approach is what we're going to allow, for core/3.70. Part of core/3.70 compatibility work for plugins will be investigating/responding to this change. A "better" solution would be Most Welcome.

bmbouter:
@ggainey I'm seeing all thumbs up here and on matrix. I am also +1 to merge.

As a practical sharing of info, @git-hyagi is out for a while starting end of business tomorrow, and in between now and then he's pretty much at-capacity so if we don't merge we've got to figure out a new plan for contribution.

Issue #6165, opened 2024-12-19
Add a filter for unblocked_at for Tasks API


dkliban:
The Tasks API does not currently support filtering by `unblocked_at__isnull`. This is needed to see all the tasks that are currently blocked or unblocked.

PR #6176, opened 2025-01-06
Add a filter for `unblocked_at` for Tasks API

git-hyagi:
closes: #6165


bmbouter (approved):
This upstream code is also working well for us in the services deployment.

PR #6202, opened 2025-01-13
[experiment]

pedro-psb:


.ci/ansible/settings.py.j2
```diff
@@ -29,24 +29,19 @@ API_ROOT = {{ api_root | repr }}
...
+DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
+AWS_ACCESS_KEY_ID = "{{ minio_access_key }}"
+AWS_SECRET_ACCESS_KEY = "{{ minio_secret_key }}"
+AWS_S3_REGION_NAME = "eu-central-1"
+AWS_S3_ADDRESSING_STYLE = "path"
+AWS_S3_SIGNATURE_VERSION = "s3v4"
+AWS_STORAGE_BUCKET_NAME = "pulp3"
+AWS_S3_ENDPOINT_URL = "http://minio:9000"
+AWS_DEFAULT_ACL = "@none None"
+{% endif %}
```
mdellweg, line 44: Ohhhh. This file is not templated by the template but by the CI.
I think what you can do is fall back to writing the `if` in python.

And we _need_ to tackle making all of this simpler soon.

mdellweg: Forget that, I think you just forgot to quote "STORAGES"...

.ci/ansible/settings.py.j2
```diff
@@ -29,24 +29,19 @@ API_ROOT = {{ api_root | repr }}
...
-            "bucket_name": "pulp3",
-            "endpoint_url": "http://minio:9000",
-            "default_acl": "@none None",
-        },
-    },
-    "staticfiles": {
-      "BACKEND": "django.contrib.staticfiles.storage.StaticFilesStorage",
-    },
-}
+{% if STORAGES in pulp_scenario_settings %}
```
mdellweg, line 32:
```suggestion
{% if "STORAGES" in pulp_scenario_settings %}
```

pedro-psb:
Looks like this is worked (here the [final settings](https://github.com/pulp/pulpcore/actions/runs/12753999130/job/35546954357?pr=6202#step:13:161)).
What has been done:
1. Defined STORAGES for the s3 scenario in `template_config.yml`
2. Modified plugin-template to use plugin-defined storages if present
3. Applied plugin-template

I'm actually a bit surprised that the minio variable interpolation worked, since the code with the curly braces was itself injected through templating.

template_config.yml
```diff
@@ -80,6 +80,20 @@ pulp_settings_s3:
...
+          "access_key": "{{ minio_access_key }}"
+          "secret_key": "{{ minio_secret_key }}"
+          "region_name": "eu-central-1"
+          "addressing_style": "path"
+          "signature_version": "s3v4"
+          "bucket_name": "pulp3"
+          "endpoint_url": "http://minio:9000"
+          "default_acl": "@none None"
+    "staticfiles":
+      "BACKEND": "django.contrib.staticfiles.storage.StaticFilesStorage"
```
pedro-psb, line 96: @mdellweg the storage settings were provided by the template and now they would be provided per-plugin.
Maybe we should just add a flag to whether to use or not the new form and leave the fallback in plugin-template?

mdellweg:
Are you saying we now use the literal string "{{ minio_access_key }}" as _the_ key? I agree, a new flag just once again could be the easier solution here.

pedro-psb:
> Are you saying we now use the literal string "{{ minio_access_key }}" as the key? 

No, I'm saying I expected it to yield the literal, but [it didnt](https://github.com/pulp/pulpcore/actions/runs/12754670793/job/35549103417?pr=6202#step:13:158). Anyway, I'm going with the YAPTF (yet another plugin template flag).

mdellweg:
> > Are you saying we now use the literal string "{{ minio_access_key }}" as the key?
> 
> No, I'm saying I expected it to yield the literal, but [it didnt](https://github.com/pulp/pulpcore/actions/runs/12754670793/job/35549103417?pr=6202#step:13:158). Anyway, I'm going with the YAPTF (yet another plugin template flag).

Thank you. That's definitely better than a solution we don't understand.

PR #6224, opened 2025-01-23
Add support for adding pull-through content to associated repositories

gerrod3:


pulpcore/content/handler.py
```diff
@@ -934,15 +939,16 @@ def _save_artifact(self, download_result, remote_artifact, request=None):
...
                         try:
                             with transaction.atomic():
-                                remote_artifact.save()
+                                ra.save()
                         except IntegrityError:
                             # Remote artifact must have already been saved during a parallel request
                             log.info(f"RemoteArtifact for {url} already exists.")
+                    if ca.relative_path == content_artifact.relative_path:
+                        # Side effect used by pull-through-caching in _stream_remote_artifact
+                        remote_artifact.content_artifact = ca
```
mdellweg, line 951: Apart from the question, why we should rename the variable in the first place, does `remote_artifact` exist here?
Also should we save "content_artifact" on the model when we set it? Maybe the comment is missing some nasty details to understand the codeflow here.

gerrod3: `remote_artifact` exists in the method definition.

I don't understand your second question, but I am wondering if I even need to use a side-effect to get the newly saved content_artifact for the pull-through method. This is a private method and searching through the code-base no one uses this method besides the content app, so maybe I can just change the return type to a dictionary of all the newly created objects.

mdellweg: So content_artifact isn't even a field on the remote_artifact?
But yes the returning the proper stuff sounds better.
So sad that with python it's usually hard to guess what is a private and what a public interface...

pulpcore/content/authentication.py
```diff
@@ -17,6 +19,13 @@
 _ = gettext.gettext
 
 
+@middleware
+async def guid(request, handler):
+    """Sets the django_guid for each request."""
+    set_guid(generate_guid())
+    return await handler(request)
+
+
```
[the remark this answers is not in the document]
gerrod3: Maybe yeah, we would also need to expose the value to be used in logging for the content app, currently this change is solely to allow for dispatching tasks from the content app.

mdellweg (approved):

pulpcore/tests/unit/content/test_handler.py
```diff
@@ -446,3 +461,41 @@ def test_handle_checkpoint_before_first_ts(
...
+        await handler._match_and_stream(f"{distro.base_path}/c123", request123)
+        handler._stream_content_artifact.assert_called_once()
+        assert ca in handler._stream_content_artifact.call_args[0]
+        repo.pull_through_add_content.assert_called_once()
+        assert ca in repo.pull_through_add_content.call_args[0]
+    finally:
+        await content.adelete()
+        await repo.adelete()
+        await remote.adelete()
+        await distro.adelete()
```
mdellweg, line 501: I don't think we need that... Django unittests usually run in a transaction that never gets committed.

mdellweg (approved):

PR #6245, opened 2025-01-29
Add checkpoint support to Pulp core and file

### LINKED ISSUE #6244: Add checkpoint support to Pulp
Pulp currently doesn’t have the concept of a repo snapshot. Even though
it does have publications which are essentially snapshots, Pulp doesn’t
serve specific publications as snapshots. This proposal introduces the
checkpoint support in Pulp by marking specific publications as checkpoints
and serving them using checkpoint distributions.

I'd like to be able to mark specific publications as checkpoints and serve them. Their respective repo versions should be protected from retain_repo_versions cleanup. Similar to https://snapshot.debian.org/ and https://snapshot.ubuntu.com/

Building this in our app outside of Pulp.

**Additional context**
Proposal: https://discourse.pulpproject.org/t/pulp-snapshot-feature-proposal/1839

Moustafa-Moustafa:
This PR adds the support for checkpoints in Pulp. This is related to this [proposal ](https://discourse.pulpproject.org/t/pulp-snapshot-feature-proposal/1839). Publications can be marked as checkpoints at the creation time. Checkpoint publications are served through checkpoint distributions. Each repository can have 0-n checkpoint distributions, all serving the same content (i.e. the checkpoint publications).
The change adds a new field "checkpoint" to both the Distribution and Publication models. New publication can be made checkpoint publications by setting the checkpoint field. Checkpoint distributions can be created by setting the checkpoint field on the Distribution model and will serve all checkpoint publications. It's up to plugin writers to expose the checkpoint fields on their respective serializers.
This PR also enables the checkpoint support for pulp_file.

closes #6244


daviddavis:
I think we need docs as part of this PR. Maybe some on how plugin writers can enable this feature for their plugins in /docs/dev/learn/subclassing and then some user docs somewhere in /docs/user somewhere.

pulpcore/app/migrations/0127_distribution_checkpoint_publication_checkpoint.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.2.17 on 2024-12-13 15:11
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("core", "0126_remoteartifact_failed_at"),
```
daviddavis, line 9:
```suggestion
        ("core", "0127_remove_upstreampulp_pulp_label_select"),
```

pulpcore/app/serializers/publication.py
```diff
@@ -36,11 +36,18 @@ def validate(self, data):
...
         repository = data.pop("repository", None)  # not an actual field on publication
         repository_version = data.get("repository_version")
+        checkpoint = data.get("checkpoint", None)
         if not repository and not repository_version:
             raise serializers.ValidationError(
                 _("Either the 'repository' or 'repository_version' need to be specified")
             )
         elif not repository and repository_version:
+            if checkpoint:
+                version = repository.latest_version()
```
daviddavis, line 46: At least one of the tests is failing here. It's possible that publish was called with `repository_version` and not `repository` which means that `repository` might be None:

```
AttributeError: 'NoneType' object has no attribute 'latest_version'
```

daviddavis: In fact the `elif` above is specifically checking for `not repository`.

Moustafa-Moustafa: I meant to look into that last week but was pulled in some other work streams. I'll look into it this week and update my PR.

docs/user/guides/checkpoint.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+## Accessing Checkpoints
+
+### Listing All Checkpoints
+You can access a listing of all the available repository's checkpoint publications by accessing the
+base path of any of the repository's checkpoint distributions. 
+
+```bash
+http :24816/pulp/content/checkpoint/myfile
```
mdellweg, line 61: I'd prefer something "more user oriented" here. Whatever that means. How about:
```suggestion
http https://pulp.example/pulp/content/checkpoint/myfile
```

https://en.wikipedia.org/wiki/.example

pulpcore/app/models/publication.py
```diff
@@ -98,12 +99,13 @@ class Publication(MasterModel):
 
     complete = models.BooleanField(db_index=True, default=False)
     pass_through = models.BooleanField(default=False)
+    checkpoint = models.BooleanField(default=False, editable=False)
```
mdellweg, line 102: What is the editable actually doing here? Is it picked up by DRF at all?

https://docs.djangoproject.com/en/5.1/ref/models/fields/#editable

pulpcore/app/models/publication.py
```diff
@@ -159,6 +165,10 @@ def delete(self, **kwargs):
             # It's possible for errors to occur before any publication has been completed,
             # so we need to handle the case when no Publication exists.
             try:
+                if self.checkpoint:
+                    base_paths |= Distribution.objects.filter(
+                        checkpoint=self.checkpoint, repository=self.repository_version.repository
```
mdellweg, line 170:
```suggestion
                        checkpoint=True, repository=self.repository_version.repository
```

pulpcore/app/serializers/publication.py
```diff
@@ -36,11 +36,18 @@ def validate(self, data):
...
             raise serializers.ValidationError(
                 _("Either the 'repository' or 'repository_version' need to be specified")
             )
         elif not repository and repository_version:
+            if checkpoint:
+                latest_version = repository_version.repository.latest_version()
+                if latest_version and latest_version != repository_version:
+                    raise serializers.ValidationError(
+                        _("Checkpoint can only be created for the repository's latest version")
+                    )
```
mdellweg, line 50: OK, I understand this is a deliberate limitation to make the whole checkpoint in time logic explicit.
How about we just never allow `repository_version` and `checkpoint=True` and instead say you need to specify `repository` and `checkpoint=True`?

pulpcore/app/serializers/publication.py
```diff
@@ -316,6 +325,12 @@ def validate(self, data):
                     "may be used simultaneously."
                 )
             )
+        elif checkpoint and (
+            not repository_provided or publication_provided or repository_version_provided
+        ):
+            raise serializers.ValidationError(
+                _("The 'checkpoint' attribute may only be used with the 'repository' attribute.")
+            )
```
mdellweg, line 333: :+1:

pulpcore/content/handler.py
```diff
@@ -164,6 +186,7 @@ class Handler:
     ]
 
     distribution_model = None
+    checkpoint_ts_format = "%Y%m%dT%H%M%SZ"
```
mdellweg, line 189: Do we expect this to be changed when subclassing the handler?
I think we can make it a global constant...

pulpcore/content/handler.py
```diff
@@ -351,6 +373,125 @@ def _match_distribution(cls, path, add_trailing_slash=True):
...
+                checkpoint_publication.pulp_created
+            )
+            request_timestamp_str = Handler._format_checkpoint_timestamp(request_timestamp)
+            if pub_timestamp_str != request_timestamp_str:
+                Handler._redirect_sub_path(f"{base_path}/{pub_timestamp_str}/")
+
+            distro.base_path = f"{base_path}/{request_timestamp_str}"
+            distro.repository = None
+            distro.publication = checkpoint_publication
+            return distro
```
mdellweg, line 425: Ooooof!
This is returning a modified but unsaved version of the distribution. I can see how it works, but will we still understand that in six months?

pulpcore/content/handler.py
```diff
@@ -326,6 +349,10 @@ def _match_distribution(cls, path, add_trailing_slash=True):
                 .get(base_path__in=base_paths)
                 .cast()
             )
+
+            if distro_object.checkpoint:
+                return cls._handle_checkpoint_distribution(distro_object, original_path)
+            return distro_object
```
mdellweg, line 355: This feels (to me) like the wrong location to hook that logic.
Can you take a closer look at `_match_and_stream`?
Or should it even be part of the `Distribution.content_handler`?
Down there we already have parsed `rel_path` for example.

Yes I know, the whole structure of the handler is already pretty hard to grasp.

pulpcore/tests/functional/api/using_plugin/test_content_access.py
```diff
@@ -69,3 +75,91 @@ def test_upload_file_on_demand_already(
...
+    # setup
+    repo = file_repository_factory()
+    distribution = file_distribution_factory(repository=repo.pulp_href, checkpoint=True)
+
+    pub_0 = create_publication(repo, False)
+    pub_1 = create_publication(repo, True)
+    pub_2 = create_publication(repo, False)
+    pub_3 = create_publication(repo, True)
+    pub_4 = create_publication(repo, False)
+
```
mdellweg, line 115: One suggestion to make the test more descriptive:
If we move this setup into a module local but class scoped fixture, we can turn this test function into a class and split the assertions into individual tests. Think `TestCheckpointDistribution.test_invalid_ts_returns_404`, `...test_arbitrary_timestamp_is_redirected`, `...test_base_path_lists_checkpoints`, ...

The class is crucial to share the single repository setup between the individual tests, by the nature of our factory fixtures.

mdellweg (commented):
Sorry, this is quite a lot. But it's a huge pr also...

pulpcore/tests/functional/api/using_plugin/test_checkpoint.py
```diff
@@ -0,0 +1,176 @@
...
+    file_repository_factory,
+    file_distribution_factory,
+    random_content,
+    file_bindings,
+    gen_object_with_cleanup,
+    monitor_task,
+):
+    def create_publication(repo, checkpoint):
+        task = file_bindings.RepositoriesFileApi.modify(
+            repo.pulp_href, {"add_content_units": [random_content]}
```
mdellweg, line 37: This is trying to add the same content each time therefore not producing a new repository version.

If we used the file upload with a counter as the relative_path, we had an easy way to identify the repository version served later.

pulpcore/content/handler.py
```diff
@@ -351,6 +369,114 @@ def _match_distribution(cls, path, add_trailing_slash=True):
...
+    def _handle_checkpoint_distribution(cls, distro, path):
+        """
+        Handle a checkpoint distribution.
+
+        Args:
+            distro (Distribution): The checkpoint distribution.
+            path (str): The checkpoint path component of the URL (e.g. 20250212T194653Z/dists/).
+
+        Returns:
+            The detail object of the matched distribution.
```
mdellweg, line 382: This returns a publication.

pulpcore/content/handler.py
```diff
@@ -351,6 +369,114 @@ def _match_distribution(cls, path, add_trailing_slash=True):
 
         raise PathNotResolved(original_path)
 
+    @classmethod
+    def _handle_checkpoint_distribution(cls, distro, path):
+        """
+        Handle a checkpoint distribution.
```
mdellweg, line 375: This does not "handle" the request as in "no further action needed"...

```suggestion
    def _select_checkpoint_publication(cls, distro, path):
        """
        Determine the checkpoint publication to serve based on the path's timestamp.
        Redirect if it's not an exact match.
```

Remind me again, why we want to redirect now. Is this a caching thing?
At this point we did identify the the publication and could just go ahead serve content from there. But I'm no expert in how the cache works.

pulpcore/tests/functional/api/using_plugin/test_checkpoint.py
```diff
@@ -0,0 +1,176 @@
...
+        pubs, distribution = setup
+        pub_0_url = (
+            f"{distribution.base_url}{Handler._format_checkpoint_timestamp(pubs[0].pulp_created)}/"
+        )
+        with pytest.raises(ClientResponseError) as exc:
+            http_get(pub_0_url).decode("utf-8")
+        assert exc.value.status == 404
+
+    @pytest.mark.parallel
+    def test_future_timestamp_returns_404(self, setup, http_get):
```
mdellweg, line 158: Can we add another test that requesting for "now()" serves publication 3?

pulpcore/content/handler.py
daviddavis: Yes, if we serve up the file, then the file will be cached for each unique request. If we redirect, then the file is only served at one url and is only cached once.

Moustafa-Moustafa:
> Sorry, this is quite a lot. But it's a huge pr also...

@mdellweg thanks for your insightful feedback, I'm confident this PR is in a better state thanks to your feedback.

pulpcore/content/handler.py
mdellweg: I think it's fine that way. Can you add that reason as a comment?

pulpcore/tests/functional/api/using_plugin/test_checkpoint.py
```diff
@@ -0,0 +1,193 @@
...
+from time import sleep
+from urllib.parse import urlparse
+from aiohttp import ClientResponseError
+import pytest
+from pulpcore.client.pulp_file import exceptions
+from pulpcore.content.handler import Handler
+
+
+@pytest.fixture(scope="class")
+def random_content(tmp_path_factory, file_bindings, monitor_task):
```
mdellweg, line 14: Let's call it `content_factory`...

pulpcore/tests/functional/api/using_plugin/test_checkpoint.py
```diff
@@ -0,0 +1,193 @@
...
+    def _random_content(name):
+        file = tmp_path_factory.mktemp("content") / name
+        file.write_text(name)
+        return monitor_task(
+            file_bindings.ContentFilesApi.create(relative_path=name, file=file).task
```
mdellweg, line 19: as of pulpcore 3.70 this sadly needs to be
```suggestion
            file_bindings.ContentFilesApi.create(relative_path=name, file=str(file)).task
```

pulpcore/tests/functional/api/using_plugin/test_checkpoint.py
```diff
@@ -0,0 +1,193 @@
...
+        assert exc.value.status == 404
+
+        with pytest.raises(ClientResponseError) as exc:
+            http_get(f"{distribution.base_url}20259928T092752Z/")
+        assert exc.value.status == 404
+
+    @pytest.mark.parallel
+    def test_checkpoint_artifact_is_served(self, setup, http_get):
+        pubs, distribution = setup
+        # arbitrary ts
```
mdellweg, line 116: This comment does not tell me anything.
```suggestion
```

pulpcore/tests/functional/api/using_plugin/test_checkpoint.py
```diff
@@ -0,0 +1,193 @@
...
+        pub_1_url = (
+            f"{distribution.base_url}{Handler._format_checkpoint_timestamp(pubs[1].pulp_created)}/"
+        )
+        pub_2_url = (
+            f"{distribution.base_url}{Handler._format_checkpoint_timestamp(pubs[2].pulp_created)}/"
+        )
+        pub_2_response = http_get(f"{pub_2_url}PULP_MANIFEST").decode("utf-8")
+        pub_1_response = http_get(f"{pub_1_url}PULP_MANIFEST").decode("utf-8")
+        assert pub_2_response == pub_1_response
```
mdellweg, line 125: At this point we could verify that we see exactly content "1".

pulpcore/tests/functional/api/using_plugin/test_checkpoint.py
```diff
@@ -0,0 +1,193 @@
...
+
+    @pytest.mark.parallel
+    def test_arbitrary_timestamp_is_redirected(self, setup, http_get):
```
mdellweg, line 128: I'm a but confused about the intent of this test.
It's a bit long and none of the timestamps are actually arbitrary.
Should it be further split up?

mdellweg (commented):
Review is iterative...

mdellweg (dismissed):

pulp_file/app/serializers.py
```diff
@@ -115,10 +115,11 @@ class FilePublicationSerializer(PublicationSerializer):
         required=False,
         allow_null=True,
     )
+    checkpoint = serializers.BooleanField(default=False)
```
mdellweg, line 118: Please don't default values in the serializers. `required=False` should be used instead.
The model will do the default then.

pulpcore/tests/functional/api/using_plugin/test_checkpoint.py
```diff
@@ -0,0 +1,209 @@
...
+
+        pub_2_response = http_get(f"{pub_2_url}PULP_MANIFEST").decode("utf-8")
+        pub_1_response = http_get(f"{pub_1_url}PULP_MANIFEST").decode("utf-8")
+
+        assert pub_2_response == pub_1_response
+        lines = pub_1_response.strip().split("\n")
+        artifact_names = [line.split(",")[0] for line in lines]
+        assert len(artifact_names) == 2
+        assert "0" in artifact_names
+        assert "1" in artifact_names
```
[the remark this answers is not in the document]
Moustafa-Moustafa: The check here is for accessing a specific artifact. I think the existing checks in the other tests make sense and are more readable.

Moustafa-Moustafa:
@mdellweg, can you please take another look?

mdellweg:
Thank you for all the hard work.

Issue #5214, opened 2024-04-01
Enhance replicate feature


PotentialIngenuity:
**Is your feature request related to a problem? Please describe.**
I performed a replicate with a tag filter and found that it deleted every remote/repository/distribution that didnt match the filter. This is not how I would expect the feature to work. 

We have several repos we mirror from public repos that get updated once a week. We also have internal repos that get pushed to irregularly. We would like to replicate these repos separately but we currently cannot since anything outside of the filter gets deleted.

**Describe the solution you'd like**
I think it should never touch anything outside of the filter. Only change what the user is targeting. The removal step could still be run for the filtered content just don't remove anything outside of the filter.

mdellweg:
This is how the replicate feature was designed and I'm unsure we can actually change that. The idea is that a replica is an hands off box that just fully mirrors a selected subset of upstream distributions.
I think, you can however limit its reach to a domain (given all your plugins already support that).

In case you need much more control, consider writing an ansible playbook a bit like this: https://github.com/mdellweg/squeezer/commit/f4261424c9b9e1c4fd01fa0551ced61433698714

mdellweg:
> This is how the replicate feature was designed and I'm unsure we can actually change that. The idea is that a replica is an hands off box that just fully mirrors a selected subset of upstream distributions. I think, you can however limit its reach to a domain (both client and server) (given all your plugins already support that).
> 
> In case you need much more control, consider writing an ansible playbook a bit like this: [mdellweg/squeezer@f426142](https://github.com/mdellweg/squeezer/commit/f4261424c9b9e1c4fd01fa0551ced61433698714)

dkliban:
It would be better if the replicate task would create on the replicate Pulp, the pulp labels that are specified in the UpstreamPulp config. THen when the repositories, remotes, and distributions are created, these labels are applied to them. In subsequent runs of the replicate task, the replicate will only consider local repositories with the labels. All other repositories will not be affected by the replicate task. This will allow users to run replicate on a subset of repositories.

PotentialIngenuity:
This feature has become more prevalent as I have added other plugins to our environment. Those being containers and python. Each of those occasionally have errors and cause the whole replication task to be marked as a failure. Its makes error handling hard trying to distinguish which plugin actually failed. I was able to create a process that scrapes the logs for the error and this has made it wayy quicker to diagnose the problem. 

[...]

PR #6247, opened 2025-01-30
Enhance replicate delete policy

gerrod3:
Added a new field `delete_policy` that will determine how replicate will clean up the extra/leftover objects present in the downstream Pulp. The field is set to the default `all` which is the current behavior of deleting everything. The other two options are `none` and `labeled`. `none` is self-explanatory. Replicate now also labels each object created/updated on a replicate with the label `UpstreamPulp:{upstream.pk}`. This new label is what is used by the `labeled` policy to determine which objects should be deleted. Only objects with this label and are not present in the Upstream will be deleted under this policy, everything else will be left.

Also, added copying the Upstream objects' labels onto the downstream. This change isn't strictly necessary for the new delete feature, so I can remove it, add a new field for customization, or move it into another PR if wanted. 

fixes: #5214


mdellweg:
Please take the following as a pure observation, something to be aware of:
This is (or in case of acceptance would be) the first time a Pulp feature depends on labels stored on it's own entities.
Yes, replicate has relied on remote labels before. But that's a different thing. Labels have been there to bridge gaps for integrators and are in full control of the user.
I can not state this is good or bad it just feels a bit odd.

pulpcore/app/models/replica.py
```diff
@@ -11,6 +11,15 @@
 
 
 class UpstreamPulp(BaseModel, AutoAddObjPermsMixin):
+    ALL = "all"
+    LABELED = "labeled"
+    NONE = "none"
+    DELETE_CHOICES = (
+        (ALL, "Delete all local objects not present on upstream."),
+        (LABELED, "Delete local labeled objects not present on upstream."),
```
mdellweg, line 19: What should happen if you have a local repository without a label. Should replicate replace that, leave it alone or simply fail?

gerrod3: Note that my description isn't that great. The 'local labeled objects' refers to objects that have been labeled by replicate with `UpstreamPulp:{id}`. So any repository without this label means it wasn't created(tracked) by replicate and thus will not be removed under this policy. Say there is a unlabeled repository with the same name as an upstream one (a case our users will be in on their first replicate if this is merged), then that repository will get the new label and then be tracked for all future replicates.

mdellweg: Will we set the labels anyway? Then users could run one more "old style" replication to label all the affected ones and change the flag then.
Now we could choose to "protect" manually created repositories from the replicators.

gerrod3:
Ok I updated the PR to change the field to just `policy`. It does almost the exact same behavior, but now it also doesn't update existing local objects during the `LABELED` policy if there is a name collision and the object does not have a matching UpstreamPulp label. Also, renamed the `NONE` policy to `NODELETE`.

mdellweg (commented):
Damn, this review was pending for some days...

pulpcore/app/replica.py
```diff
@@ -77,6 +78,13 @@ def needs_update(fields_dict, model_instance):
                 needs_update = True
         return needs_update
 
+    def _is_managed(self, object):
+        """Checks to see if the passed in object should be managed by replicate."""
+        if self.server.policy == UpstreamPulp.LABELED:
+            if object.pulp_labels.get("UpstreamPulp") != str(self.server.pk):
+                return False
+        return True
```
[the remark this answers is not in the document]
mdellweg: I'm wondering if that would make a forth policy. We can also just document that and maybe add the forth policy later.

mdellweg (dismissed):

mdellweg:
There's a migration conflict needing this to rebase.

ggainey (approved):

PR #6251, opened 2025-01-31
Expand docs on correlation id

### LINKED ISSUE #6248: Where are the docs on correlation ID header features and the correlation feature set in general?
I looked for the, but I couldn't find them.

If they don't exist I think they need to cover:
* All API logs and tasking logs are prepended with a correlation ID
* API calls coming into Pulp can specify a correlation ID by setting a header named `Correlation-ID`
* If no correlation ID header is specified Pulp will generate one
* Any tasks generated in an API call will use the same correlation ID
* For any Task, the user can see the correlation ID in the `logging_cid` field
* For any API call, the correlation ID is present as a response header named `Correlation-ID`

daviddavis:
fixes #6248


mdellweg (approved):

bmbouter (approved):
Thank you!

PR #6253, opened 2025-02-03
Add pulp_created field as a task filter

### LINKED ISSUE #6241: Allow users to filter tasks by pulp_created date
I would like to be able to filter tasks by the pulp_created date.

decko:
Adds the `pulp_created` datetime field as a filter to allow
filtering tasks by its creation date. Also, adds a test to
check if the filter is working as expected.

Closes #6241

### DIFF
--- a/pulpcore/tests/functional/api/test_tasking.py
+++ b/pulpcore/tests/functional/api/test_tasking.py
@@ -5,6 +5,7 @@ import pytest
 import time
 
 from aiohttp import BasicAuth
+from datetime import datetime
 from urllib.parse import urljoin
 from uuid import uuid4
 
@@ -270,6 +271,33 @@ def test_filter_tasks_using_worker__in_filter(pulpcore_bindings, dispatch_task, 
 
     assert task1_href in tasks_hrefs
     assert task2_href in tasks_hrefs
+
+
+@pytest.mark.parallel
+def test_filter_tasks_using_pulp_created_filter(pulpcore_bindings, dispatch_task):
+
+    task1_href = dispatch_task("pulpcore.app.tasks.test.sleep", args=(0,))
+
+    time.sleep(2)
+    start_time = datetime.now()
+    task2_href = dispatch_task("pulpcore.app.tasks.test.sleep", args=(0,))
+    task3_href = dispatch_task("pulpcore.app.tasks.test.sleep", args=(0,))
+
+    search_results_gte = pulpcore_bindings.TasksApi.list(pulp_created__gte=start_time)
+
+    tasks_hrefs_gte = [task.pulp_href for task in search_results_gte.results]
+
+    assert task1_href not in tasks_hrefs_gte
+    assert task2_href in tasks_hrefs_gte
+    assert task3_href in tasks_hrefs_gte
+
+    search_results_lte = pulpcore_bindings.TasksApi.list(pulp_created__lte=start_time)
+
+    tasks_hrefs_lte = [task.pulp_href for task in search_results_lte.results]
+
+    assert task1_href in tasks_hrefs_lte
+    assert task2_href not in tasks_hrefs_lte
+    assert task3_href not in tasks_hrefs_lte
 
 
 def test_cancel_gooey_task(pulpcore_bindings, dispatch_task, monitor_task):
--- a/CHANGES/6241.feature
+++ b/CHANGES/6241.feature
@@ -1,0 +1,1 @@
+Added `pulp_created` filter for Tasks API.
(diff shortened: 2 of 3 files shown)


pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -272,6 +273,23 @@ def test_filter_tasks_using_worker__in_filter(pulpcore_bindings, dispatch_task,
     assert task2_href in tasks_hrefs
 
 
+def test_filter_tasks_using_pulp_created_filter(pulpcore_bindings, dispatch_task, monitor_task):
```
bmbouter, line 276: We should have this run in parallel with `@pytest.mark.parallel` decorator. We need to also think if the test is safe to run in parallel.

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -272,6 +273,23 @@ def test_filter_tasks_using_worker__in_filter(pulpcore_bindings, dispatch_task,
...
+    dispatch_task("pulpcore.app.tasks.test.sleep", args=(0,))
+    dispatch_task("pulpcore.app.tasks.test.sleep", args=(0,))
+
+    search_results = pulpcore_bindings.TasksApi.list(
+        pulp_created__gte=start_time + timedelta(seconds=2)
+    )
+
+    tasks_hrefs = [task.pulp_href for task in search_results.results]
+
+    assert len(tasks_hrefs) == 2
```
bmbouter, line 290: I think this assert (in parallel) could work in some unexpected ways. Specifically other `pulpcore.app.tasks.test.sleep` tasks could be running and created at the same time. To solve that I recommend asserting that tasks 2 and 3 are in the results and task 1 is not instead of the count of results.

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -272,6 +273,23 @@ def test_filter_tasks_using_worker__in_filter(pulpcore_bindings, dispatch_task,
     assert task2_href in tasks_hrefs
 
 
+def test_filter_tasks_using_pulp_created_filter(pulpcore_bindings, dispatch_task, monitor_task):
+    start_time = datetime.now()
```
bmbouter, line 277: This could be moved down to just after the sleep statement, then you wouldn't have to `timedelta(seconds=2)`

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -272,6 +273,25 @@ def test_filter_tasks_using_worker__in_filter(pulpcore_bindings, dispatch_task,
     assert task2_href in tasks_hrefs
 
 
+@pytest.mark.parallel
+def test_filter_tasks_using_pulp_created_filter(pulpcore_bindings, dispatch_task, monitor_task):
```
bmbouter, line 277: You can remove `monitor_task` it's unused.

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -272,6 +273,25 @@ def test_filter_tasks_using_worker__in_filter(pulpcore_bindings, dispatch_task,
...
+
+    task1_href = dispatch_task("pulpcore.app.tasks.test.sleep", args=(0,))
+
+    time.sleep(2)
+    start_time = datetime.now()
+    task2_href = dispatch_task("pulpcore.app.tasks.test.sleep", args=(0,))
+    task3_href = dispatch_task("pulpcore.app.tasks.test.sleep", args=(0,))
+
+    search_results = pulpcore_bindings.TasksApi.list(pulp_created__gte=start_time)
```
bmbouter, line 286: You could consider writing an additional `list()` call with `pulp_created__lte=star_time` and asserting the opposite tasks are in the result.

bmbouter (approved):
I left some comments. You can either merge as-is or make an adjustment. I'm satisfied either way. Thank you!

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -272,6 +273,25 @@ def test_filter_tasks_using_worker__in_filter(pulpcore_bindings, dispatch_task,
...
 
+@pytest.mark.parallel
+def test_filter_tasks_using_pulp_created_filter(pulpcore_bindings, dispatch_task, monitor_task):
+
+    task1_href = dispatch_task("pulpcore.app.tasks.test.sleep", args=(0,))
+
+    time.sleep(2)
+    start_time = datetime.now()
+    task2_href = dispatch_task("pulpcore.app.tasks.test.sleep", args=(0,))
+    task3_href = dispatch_task("pulpcore.app.tasks.test.sleep", args=(0,))
```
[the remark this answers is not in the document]
decko: I'm checking if it's present in the result array. Since we're filtering using the created date I think it's good to check to see if the latest two tasks are the ones returning in the filtering.

Issue #6236, opened 2025-01-27
Labels for Domains


bmbouter:
We heavily use domains, e.g. we have just a little less than 700 of them. Several stakeholders want to use different storage buckets so they create domains for each even though they are "one stakeholder". When we run reports on installation usage, it's becoming practically difficult to keep track of who is actually owning which domains.

I don't believe we have label support for domains according to [this spec](https://pulpproject.org/pulpcore/restapi/#tag/Domains/operation/domains_read).

It would be great if we could have label support added to domains.

As a workaround today we're going to start using "descriptions".

jlsherrill:
+1 this would be really useful!

PR #6254, opened 2025-02-03
Add labels to domains

gerrod3:
fixes: #6236

Side note: I used Cursor & its agent feature to make this commit. With just some minor edits, it correctly added everything on its own. It even knew to run the migration command after adding the field (of course I had to manually run it since it didn't know about our oci-env). Maybe we should take a look at our issue backlog and note the simple/medium issues that we can throw to the AI (I'll take requests :smile:).


pulpcore/app/migrations/0101_add_domain.py
mdellweg, line 1: Can you elaborate on changing this old migration?

gerrod3: So as you can see the CI is failing on the migrations because I made the grave mistake of calling outside model code (`get_domain_pk`) from inside a migration (any migration adding a pulp_domain relation). On a new installation this causes the old migrations to fail now since I've added a new field to the Domain model which means the SQL the outside code generates will be too far ahead of the database at the time of the migration. So I had to change this migration and I'm going to have to change `get_domain_pk` to only pull in the pk.

mdellweg: So I see the `add_domain_relations` migration failing. I think we should fix that one instead.

pulpcore/app/migrations/0101_add_domain.py
```diff
@@ -58,4 +66,5 @@ class Migration(migrations.Migration):
             bases=(django_lifecycle.mixins.LifecycleModelMixin, models.Model, pulpcore.app.models.access_policy.AutoAddObjPermsMixin),
         ),
         migrations.RunSQL(DEFAULT_DELETE_TRIGGER, reverse_sql=REMOVE_DEFAULT_DELETE_TRIGGER),
+        migrations.RunPython(code=create_default_domain, reverse_code=migrations.RunPython.noop),
```
mdellweg, line 69: As a rule of thumb we established that runpython is good to migrate existing data to a new format. But for creating new data, you should really rely on the post-migrate hook.

mdellweg (dismissed):

pulpcore/app/migrations/0101_add_domain.py
gerrod3: Not possible. We need the default domain for the next migration to work. Without this we can't start up a clean Pulp instance.

pulpcore/app/util.py
```diff
@@ -580,7 +580,8 @@ def get_default_domain():
         except AttributeError:
             return None
         try:
-            default_domain = Domain.objects.get(name="default")
+            queryset = Domain.objects.only("pk") if pk_only else Domain.objects
+            default_domain = queryset.get(name="default")
```
mdellweg, line 584: This line will cache whatever we received as the domain. Even if it was only the shallow "pk only" version.
can we, instead leave this function alone and add a fallback implementation to `get_domain_pk`. With a big banner that it will be used in migrations...

pulpcore/app/util.py
```diff
@@ -595,7 +595,19 @@ def get_domain():
 
 
 def get_domain_pk():
-    return get_domain().pk
+    """
+    THIS CAN/WILL BE RAN IN MIGRATIONS. DO NOT USE ORM FOR IT MIGHT GENERATE INVALID SQL.
+    """
```
mdellweg, line 600: Thank you! This will help us not break it in the future.

mdellweg (commented):
I am convinced that this is probably the best we can do.
So my ask is to fasten the guardrail comments a bit more.

pulpcore/app/util.py
```diff
@@ -573,7 +573,7 @@ def set_current_user_lazy(user):
 
 def get_default_domain():
     global default_domain
-    # This can be run in a migration, and once after
+    # This will run in a post-migration hook
```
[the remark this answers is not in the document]
gerrod3: It is, we have the post-migration hook `ensure_default_domain` https://github.com/pulp/pulpcore/blob/main/pulpcore/app/apps.py#L330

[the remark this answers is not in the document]
mdellweg: Oh you are talking about the hook here. Sorry.

mdellweg (approved):

Issue #5928, opened 2024-10-23
Immediate tasks do not run in a temporary directory


mdellweg:
SSIA

PR #6256, opened 2025-02-03
Fix immediate tasks to run in temp dir

gerrod3:
fixes: #5928

This one was pretty simple for the Cursor agent to solve. Should I add tests for this?

### DIFF
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -3,11 +3,14 @@ import contextlib
 import contextvars
 import importlib
 import logging
+import os
 import sys
 import traceback
+import tempfile
 from datetime import timedelta
 from gettext import gettext as _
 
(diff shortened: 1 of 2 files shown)


pulpcore/tasking/tasks.py
```diff
@@ -226,7 +229,9 @@ def dispatch(
                 ).exists()
             ):
                 task.unblock()
-                execute_task(task)
+                with tempfile.TemporaryDirectory(dir=settings.WORKING_DIRECTORY) as working_dir:
+                    os.chdir(working_dir)
+                    execute_task(task)
```
mdellweg, line 234: When we execute the task traditionally, we create the temp dir differently, right?
So it's not ok to move it into execute_task, i guess.

mdellweg (approved):

PR #6265, opened 2025-02-06
Turn TASK_GRACE_INTERVAL into a setting

### LINKED ISSUE #6242: As a user, I can specify how long workers should wait during graceful shutdown
See [this comment](https://github.com/pulp/pulpcore/issues/6222#issuecomment-2621510706) which discusses exposing the `TASK_GRACE_INTERVAL` as a setting. This is an RFE requesting that.

I don't know how long a heartbeat is so if the setting is expressed in heartbeats (which I think I'm ok with), then I need to know how long that is.

As a user, it would be nicer to me if it was expressable in seconds. For example we also have k8s hard timeout values and those are expressed in seconds, and we need to set these similarly and together, so having them all on a seconds timescale is nicer for me. It'll be ok though if not.

mdellweg:
The new setting is measured in seconds. Also added some clarification around graceful shutdown.

fixes #6242

### DIFF
--- a/CHANGES/6242.feature
+++ b/CHANGES/6242.feature
@@ -1,0 +1,1 @@
+Added `TASK_GRACE_INTERVAL` as a setting to specify the amount of seconds a worker gives the current task to finish when receiving SIGINT.
--- a/docs/admin/reference/settings.md
+++ b/docs/admin/reference/settings.md
@@ -312,6 +312,16 @@ Defaults to `30` seconds.
 
 
 
+### TASK_GRACE_INTERVAL
+
+On receiving SIGHUP or SIGTERM a worker will await the currently running task forever.
+On SIGINT, this value represents the time before the worker will attempt to kill the subprocess.
+This time is only accurate to one worker heartbeat corresponding to `WORKER_TTL / 3`.
+
+Defaults to `600` seconds.
+
+
+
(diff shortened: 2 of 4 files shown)


bmbouter:
First off, thanks for picking this ticket up. This is looking really great.

I like the way this feature is positioned, in seconds, but also approx to keep the heartbeats the only parent process interrupts. It's also clearly documented in code and in the settings. +1 to all this.

Since we deploy on k8s, it uses SIGTERM by default and having it wait indefinitely is right for us. I also think it's nice for users too, to have a "wait indefinitely option" and having that tied to both SIGTERM and SIGHUP.

I like the idea that SIGINT can be used to invoke this feature also. We won't specifically be using that because I like the wait-indefinitely option, but for users not using k8s (which eventually comes through with a SIGKILL after a configurable k8s timeout) I think this is very valuable generally.

My only question is in the case when SIGINT is sent twice (the double Ctrl+C). In that case can we make it immediately exit also? This is debatable if that's a good idea or not. What do you think is good in this area?

Did this give you all the feedback you were looking for?

mdellweg:
On the first recieved SIGINT, the worker resets all the signal handlers, so a subsequent kill whatever will go though default python interpreter shutdown and the subprocess receives the same signals. This is as much as we can do for "immediate shutdown".

PR #6272, opened 2025-02-12
Add LIMIT statement to those queries needed during Pulp startup process.

decko:
The current query can be super heavy in environments with lots of
packages and really slow down the startup process of Pulp components.
Since we're checking the presence of any number of artifacts using a
certain checksum, limiting it to the first result can speed up the
check itself and the startup process.

### DIFF
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -58,11 +58,11 @@ MEDIA_ROOT = str(DEPLOY_ROOT / "media")  # Django 3.1 adds support for pathlib.P
 STATIC_URL = "/assets/"
 STATIC_ROOT = DEPLOY_ROOT / STATIC_URL.strip("/")
 
-# begin compatilibity layer for DEFAULT_FILE_STORAGE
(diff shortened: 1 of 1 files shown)


pulpcore/app/settings.py
```diff
@@ -552,7 +552,9 @@ def otel_middleware_hook(settings):
         with connection.cursor() as cursor:
             for checksum in ALLOWED_CONTENT_CHECKSUMS:
                 # can't import Artifact here so use a direct db connection
-                cursor.execute(f"SELECT count(pulp_id) FROM core_artifact WHERE {checksum} IS NULL")
+                cursor.execute(
+                    f"SELECT count(pulp_id) FROM core_artifact WHERE {checksum} IS NULL LIMIT 1"
+                )
```
dralley, line 557: Wouldn't `EXISTS` be a bit simpler?

dralley: Never mind, the django `.exists()` doesn't map to the SQL concept of `EXISTS` as I thought it did.

ggainey (approved):

PR #6275, opened 2025-02-15
[PULP-294] Taught reupload to self-heal when artifact-file has been d…

### LINKED ISSUE #6274: A deleted-artifact-file cannot be restored by re-uploading.
**Version**
main

**Describe the bug**
If a user accidentally deletes a file in /var/lib/pulp/media/artifacts that wasn't sync'd from somewhere, there's no way to re-upload it short of a) deleting all repo-versions that reference the content of that artifact and running orphan-cleanup, or b) messy database surgery.

The problem is that the upload-process finds the db-entity Artifact, and decides "the file is already here" so it doesn't replace it with the incoming one.

The repair script can fix things when such a mistake happens with files that have a valid RemoteArtifact. 

**To Reproduce**

1. Sync https://fixtures.pulpproject.org/rpm-signed/ immediate
2. Log into the pulp instance and `rm -rf /var/lib/pulp/media/artifacts/b0/`
3. `wget https://fixtures.pulpproject.org/rpm-signed/bear-4.1-1.noarch.rpm`
4. `pulp rpm content -t package upload --repository test --file ./bear-4.1-1.noarch.rpm`
5. Watch failures happen and Be Sad.

**Expected behavior**
RE-upload should notice that the artifact.file is *actually not there* and restore it from the incoming upload.

**Additional context**
Add any other context about the problem here.  Please provide links to any previous discussions via Discourse or Bugzilla.
This is one of the use-cases mentioned in https://issues.redhat.com/browse/PULP-294

ggainey:
…eleted.

fixes #6274.

### DIFF
--- a/pulpcore/plugin/serializers/content.py
+++ b/pulpcore/plugin/serializers/content.py
@@ -202,7 +202,11 @@ class SingleArtifactContentUploadSerializer(
                 artifact = Artifact.objects.get(
                     sha256=file.hashers["sha256"].hexdigest(), pulp_domain=get_domain_pk()
                 )
-                artifact.touch()
+                if not artifact.pulp_domain.get_storage().exists(artifact.file.name):
+                    artifact.file = file
+                    artifact.save()
+                else:
+                    artifact.touch()
             except (Artifact.DoesNotExist, DatabaseError):
                 artifact_data = {"file": file}
                 serializer = ArtifactSerializer(data=artifact_data)
--- a/CHANGES/6274.bugfix
+++ b/CHANGES/6274.bugfix
@@ -1,0 +1,1 @@
+Taught the upload-workflow to replace missing artifact-files when discovered.
--- a/pulp_file/tests/functional/api/test_crud_content_unit.py
+++ b/pulp_file/tests/functional/api/test_crud_content_unit.py
@@ -5,7 +5,7 @@ import os
 import pytest
 import uuid
 
-from pulpcore.tests.functional.utils import PulpTaskError
+from pulpcore.tests.functional.utils import generate_iso, PulpTaskError
 
 
 @pytest.mark.parallel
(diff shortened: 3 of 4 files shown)


pulpcore/plugin/serializers/content.py
```diff
@@ -202,6 +202,8 @@ def deferred_validate(self, data):
                 artifact = Artifact.objects.get(
                     sha256=file.hashers["sha256"].hexdigest(), pulp_domain=get_domain_pk()
                 )
+                if not artifact.pulp_domain.get_storage().exists(artifact.file.name):
+                    raise Artifact.DoesNotExist
                 artifact.touch()
```
mdellweg, line 207: I don't think, this is quite correct. The exception handling code here will try to save a new artifact and probably fail.

How about this?
```suggestion
                if not artifact.pulp_domain.get_storage().exists(artifact.file.name):
                    artifact.file=file
                    artifact.save()
                else:
                    artifact.touch()
```

I think we should be able to add a test for this codepath using the url on the create call.

ggainey: Artifact.save() doesn't do what we need - serializer.save() is what notices "file" and does the copy-tmp-to-perm. The change here is to notice that, in spite of the Artifact existing, the file does *not*, and to do the exception path. This is the path that is being tested by `test_reupload_damaged_artifact`. What am I missing?

mdellweg: Thanks for the clarification. Serializer save actually works different from model save. But does that serializer not need to have the artifact assigned as instance to know it should do an update? I'm probably missing something too.

I thought the test you provided would exercise the change in the viewset code.
(As far as I understand the artifact saving code is doubled, because on atomic upload the whole file is know by the viewset and we prevent moving the bits over the wire tree times. On chunked upload, or download however the file is only available in deferred_validate.)

ggainey: > (As far as I understand the artifact saving code is doubled, because on atomic upload the whole file is know by the viewset and we prevent moving the bits over the wire tree times. On chunked upload, or download however the file is only available in deferred_validate.)

Ah! *That* is what I was missing. I'll redraft this and write up a test for that.

mdellweg: Thank you, I think we want both paths to be covered anyway.

mdellweg (approved):

PR #6296, opened 2025-02-24
Use openapi 3.1.0 for schema generation

mdellweg:


ggainey (approved):
Tests all pass - What Could Possibly Go Wrong? :)

PR #6310, opened 2025-02-27
Add a new doc about how to use AWS CloudFront as backend storage.

### LINKED ISSUE #6258: Add CloudFront Support
CloudFront can "front" an S3 bucket making it geo-distributed. See the docs here: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/GettingStarted.SimpleDistribution.html

As a user, I should be able to configure AWS correctly, and then tell Pulp's storage configuration whatever it needs to know to have users receive content from the CloudFront Distribution instead of S3.

decko:
Also, adds a simple test to check if the options needed to configure
AWS CloudFront are valid.

Closes #6258

### DIFF
--- a/docs/admin/guides/configure-pulp/configure-storages.md
+++ b/docs/admin/guides/configure-pulp/configure-storages.md
@@ -86,6 +86,58 @@ Notes:
     1. The system that hosts Pulp is running on AWS
     2. And the [`instance_profile`](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html) provides access to the S3 bucket
 
+
+### Configure AWS CloudFront for Content Distribution
+
+Check out how to configure [django-storages to use CloudFront with URL signing](https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#cloudfront-signed-urls).
+
+Also, you can follow this [doc](https://github.com/aws-samples/amazon-cloudfront-signed-urls-using-lambda-secretsmanager/tree/main/2-Create_CloudFront_Distribution) to understand what you need to configure on AWS side.
+
+You'll need:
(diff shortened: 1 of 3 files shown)


docs/admin/guides/configure-pulp/configure-storages.md
```diff
@@ -86,6 +86,51 @@ Notes:
...
+  "hide_guarded_distributions": false
+}
+```
+
+=== Create a new Domain
+```bash
+# We need to format our key to a JSON friendly format
+export CLOUDFRONT_KEY=$(cat keyfile.pem | jq -sR .)
+
+curl -X POST $BASE_HREF/pulp/default/api/v3/domains/ -d '{"name": "cloudfront_storage", "storage_class": "storages.backends.s3boto3.S3Boto3Storage", "storage_settings": {"access_key": "{aws_access_key}", "secret_key": "{aws_secret_key}", "bucket_name": "{aws_bucket_name}", "region_name": "{aws_region_name}", "default_acl": "private", "cloudfront_key_id": "{aws_cloudfront_key_id}", "cloudfront_key": '"$CLOUDFRONT_KEY"', "custom_domain": "{aws_cloudfront_custom_domain}"}, "redirect_to_object_storage": true, "hide_guarded_distributions": false}' -H 'Content-Type: application/json'
```
bmbouter, line 127: Is there a `pulp` CLI equivalent option? Definitely keep this example, this is great. If there is a `pulp` CLI way to do it can that example also be added? If there isn't then I think this is just fine as-is.

decko: Thanks folks. I used `pulp-cli` in the first tests and it didn't worked out. After that I've focused in using `curl` and the docs reflected that.
Tried the same approach with `pulp-cli`, it worked and I've updated the docs.

ggainey (approved):

bmbouter (approved):
This looks great!

PR #6338, opened 2025-03-10
Fixed set/unset_label API RBAC - now comes with a new permission.

### LINKED ISSUE #6337: [PULP-362] set/unset_label API needs its own permission - not linked to the create-action
**Version**
3.73

**Describe the bug**
In https://github.com/pulp/pulpcore/pull/5787 we linked set/unset_label to the same conditions as "create". This isn't appropriate - since "create" is "assuming" a repository is involved, and set/unset are not. The result is that only an admin can actually set/unset labels - which is not the desired outcome.

set/unset needs its own permission on content, and its own role.

ggainey:
fixes #6337.


pulp_file/app/viewsets.py
```diff
@@ -193,6 +205,7 @@ class FileRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin, Role
             "file.sync_filerepository",
             "file.manage_roles_filerepository",
             "file.repair_filerepository",
+            "core.manage_labels",
```
gerrod3, line 208: Maybe we should put this under the filerepository_creator role since it is a model/domain only level permission. The owner role gets assigned at creation time to the user on the repository they created. This means all the permissions under the role get assigned to that object as well. Thus when trying to label content with just the owner role on the repository you would still get denied.

pulpcore/app/models/content.py
```diff
@@ -542,6 +542,9 @@ class Content(MasterModel, QueryMixin):
     class Meta:
         verbose_name_plural = "content"
         unique_together = ()
+        permissions = [
+            ("manage_labels", "Can manage content-labels"),
```
mdellweg, line 546: This creates a permission called "core.manage_labels" right?
I'm missing "content" in there.

pulp_file/app/viewsets.py
mdellweg: Right, so core.manage_labels is a permission on the Content model, so when you are the repository owner, does this permission kick in? Or do you need it on the domain/system level anyway?

mdellweg (changes requested):

pulp_file/app/viewsets.py
ggainey: You need the permission to call set/unset_label on a piece of Content. It "seemed to make sense" that the owner of a repository can label content within it - since "inside a repository you have access to" is the way most users get aaccess to content. Based on @gerrod3 's explanation, sounds like _creator is what we really want.

pulpcore/app/models/content.py
ggainey: Fair - how about "manage_content_labels"?

pulp_file/app/viewsets.py
```diff
@@ -91,9 +91,21 @@ class FileContentViewSet(SingleArtifactContentUploadViewSet):
...
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_or_domain_perms:core.manage_content_labels",
+                ],
+            },
         ],
         "queryset_scoping": {"function": "scope_queryset"},
     }
+    LOCKED_ROLES = {
```
ggainey, line 105: I feel like there should be a better way/place to define this specific role, rather than having a new similarly-named role for every content-type. Ideas?

ggainey:
Will undraft once changes discussed at core-team-mtg implemented

docs/user/guides/manage-labels.md
```diff
@@ -140,3 +140,24 @@ This would return a list of repositories such as:
...
+Content. If you wish a user to be allowed to do so, you need to assign that Role to them:
+
+```bash
+pulp user role-assignment add \
+  --username USERNAME \
+  --role core.content_labeler 
+  --domain default
+```
+
+Finally, as of this writing `pulp-cli` does not yet recognize setting/unsetting labels on Content.
```
gerrod3, line 163:
```suggestion
```

We are going to add this eventually, let's not give ourselves future work to have to remove this line.

pulpcore/app/global_access_conditions.py
```diff
@@ -515,20 +518,46 @@ def has_required_repo_perms_on_upload(request, view, action, permission):
         True if the user has supplied the ``repository`` parameter and has the Permission on it
             named by the ``permission`` argument, or is an admin. False otherwise.
     """
+
+    def _forbidden_labels():
+        "Return true IFF we're creating, specifying labels, and DO NOT HAVE labelling-permissions."
```
gerrod3, line 523:
```suggestion
        "Return true IF we're creating, specifying labels, and DO NOT HAVE labeling-permissions."
```

pulpcore/app/global_access_conditions.py
```diff
@@ -515,20 +518,46 @@ def has_required_repo_perms_on_upload(request, view, action, permission):
...
+    # Check incoming data is valid, raise if not
     obj = view.get_object() if view.detail else None
     serializer = view.serializer_class(
         instance=obj, data=request.data, context={"request": request}
     )
     serializer.is_valid(raise_exception=True)
+
+    # If 'repository' specified:
+    # * make sure user is allowed the requested permission on the repository
+    # * if 'create', and pulp_labels provided, make sure user is allowed to set labels
```
gerrod3, line 544: Do we need these comments if the behavior is already described in the method docs?

pulpcore/app/viewsets/task.py
```diff
@@ -154,6 +154,8 @@ class TaskViewSet(
         # the user on the object-level at task creation. It is not meant to be edited or manually
         # added/removed from users.
         "core.task_user_dispatcher": ["core.add_task"],
+        # Even with this role, can only label content user has access to
+        "core.content_labeler": ["core.manage_content_labels"],
```
gerrod3, line 158: Why add this here and not on `ListContentViewSet`? You just need to have it on a routed `NamedModelViewSet`, right? I think it would be better for it to live in `viewsets/content.py`

CHANGES/6337.bugfix
```diff
@@ -0,0 +1,3 @@
+Reworked content-labeling RBAC to use a specific permission for set/unset labels.
+
+Repository-creator gets this permission by default.
```
gerrod3, line 3:
```suggestion
```
No longer true.

gerrod3 (changes requested):

pulpcore/app/global_access_conditions.py
ggainey: Heh - IFF is math-language for "if and only if", it's not actually a typo. But, that's pretty jargon-y, will change.

Both spellings of "labelling/labeling" are correct - double-ell is a more-British-English-usage, but they're considered interchangeable (no matter what one's spellchecker says)

pulpcore/app/global_access_conditions.py
ggainey: Fair point.

pulpcore/app/viewsets/task.py
ggainey: Because "this" is where the core-team-mtg suggested it go :)

mdellweg: Sorry for the confusion. I was not actually suggesting, but took TaskViewSet as an obviously bad example that would still work.

mdellweg (approved):

gerrod3 (approved):

Issue #6333, opened 2025-03-06
An issue found in Pulp Replication


mchithu:
**Version**
Pulp installed using PyPI modules.
pulpcore==3.69.0
pulp-python==3.12.5
pulp-rpm==3.27.2

**Describe the bug**
We randomly noticed an issue in a Pulp Replication task. After sync of the repository, we found the differnce that the master has N version of repomd.xml and N-1 version of repomd.xml in the replica server. Except repomd.xml file (in N-1 version), all other files (in N version) in the repository metadata are same as master.

**To Reproduce**
I don't it know exactly. May be able to reproduce when replicate many large size repositories(like appstream, optional, epel,..) and also running synchornize job in the master pulp server simultaneously. 

**Expected behavior**
Repositories in replica and master should be in sync. The repomd.xml file should be always in N version.

**Additional context**
N/A

mchithu:
**Content of repodata directory**
The repodata directory contain the N version of the repositories except the repomd.xml (N-10 version).
199f67697007d4d4c191b38ad7b490a7f83e9d69244687e2c184b40f79ed7d9d-comps.xml                          		05-Mar-2025 14:04  41.1 kB
270ba36bb730b41e6862a533672eb82acb457a323255e062d57b11c9bde4d567-other.xml.gz                       		05-Mar-2025 14:04  7.2 MB
3407405bb2832d76a8e0827f23ff0962ad3b990688c1936e0efb14a9c0428785-updateinfo.xml.gz                  		05-Mar-2025 14:04  857.7 kB
682ef7efad0d553270faff338a43be22404179547797e01650f0b32142cbc6d8-filelists.xml.gz                   		05-Mar-2025 14:04  5.4 MB
f5026ce320425e0e1831ea64f1da2e4583c4995fef7b33f8b329a04f24ccd481-primary.xml.gz                     		05-Mar-2025 14:04  43.7 MB
f9d958a8a917bfe0a71318e45ee53ace7e3aba57ec6678d905dd2fa0c45b552e-0c9ef344-a0ee-484f-972f-44d993ea9ac8		05-Mar-2025 14:04  2.2 kB
repomd.xml                                                                                          		05-Mar-2025 14:04  2.7 kB

**Content of repomd.xml**
The following repomd.xml contains the N-1 version of the repodata files. 
`<repomd xmlns="http://linux.duke.edu/metadata/repo" xmlns:rpm="http://linux.duke.edu/metadata/rpm">
[...]

PR #6343, opened 2025-03-12
Fix cache invalidation on publication create & repo-version delete

gerrod3:
From my understanding there was a timing window during the sync that probably caused the cache issue. In replicate we sync with auto-publish which means that right after the new repo-version is created we start publishing in the same task. We invalidate the cache after a new repo-version is created, but we were failing to do so when a new publication was created (when domains were turned off). So after the sync you could end up with old items still in the cache if you had clients still fetching content during the sync. 

fixes: #6333

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -1170,7 +1170,7 @@ class RepositoryVersion(BaseModel):
             if settings.CACHE_ENABLED:
                 base_paths = self.distribution_set.values_list("base_path", flat=True)
                 if base_paths:
-                    Cache().delete(base_key=base_paths)
+                    Cache().delete(base_key=cache_key(base_paths))
 
             # Handle the manipulation of the repository version content and its final deletion in
             # the same transaction.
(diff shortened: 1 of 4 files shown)


dralley (approved):

pulpcore/app/models/repository.py
```diff
@@ -1170,7 +1170,7 @@ def delete(self, **kwargs):
             if settings.CACHE_ENABLED:
                 base_paths = self.distribution_set.values_list("base_path", flat=True)
                 if base_paths:
-                    Cache().delete(base_key=base_paths)
+                    Cache().delete(base_key=cache_key(base_paths))
```
dralley, line 1173: Would it make sense for `Cache().delete()` to be calling `cache_key()` instead of expecting the caller to remember to do that?

gerrod3: Maybe if the class was more specific. The cache class is meant to be generic and used for anything, you could cache api responses if you wanted to. `cache_key` is specific for the content app and caching by distribution.

bmbouter (approved):

PR #6358, opened 2025-03-13
Validate orphan_protection_time in orphans/cleanup

### LINKED ISSUE #3234: The endpoint for cleaning orphans does not validate input data
When setting the orphan protection time to a tremendously large number, an error is raised:
```
(pulp) [vagrant@pulp3-source-fedora35 ~]$ http :24817/pulp/api/v3/orphans/cleanup/ orphan_protection_time=123123213123123123123
HTTP/1.1 202 Accepted
Allow: POST, OPTIONS
Connection: close
Content-Length: 67
Correlation-ID: 60519da7534f4f608fa1fff620671644
Date: Wed, 21 Sep 2022 11:30:04 GMT
Referrer-Policy: same-origin
Server: gunicorn
Vary: Accept, Cookie
X-Frame-Options: DENY

{
    "task": "/pulp/api/v3/tasks/7bbb10b2-126a-47a8-98ad-a4099547ee13/"
}


(pulp) [vagrant@pulp3-source-fedora35 ~]$ http :24817/pulp/api/v3/tasks/7bbb10b2-126a-47a8-98ad-a4099547ee13/
HTTP/1.1 200 OK
Allow: GET, PATCH, DELETE, HEAD, OPTIONS
Connection: close
Content-Length: 1130
Correlation-ID: ba6eb7f34d6348a8a1e125d969073109
Date: Wed, 21 Sep 2022 11:30:10 GMT
Referrer-Policy: same-origin
Server: gunicorn
Vary: Accept, Cookie
X-Frame-Options: DENY

{
    "child_tasks": [],
    "created_resources": [],
    "error": {
        "description": "Python int too large to convert to C int",
        "traceback": "  File \"/home/vagrant/devel/pulpcore/pulpcore/tasking/pulpcore_worker.py\", line 452, in _perform_task\n    result = func(*args, **kwargs)\n  File \"/home/vagrant/devel/pulpcore/pulpcore/app/tasks/orphan.py\", line 53, in orphan_cleanup\n    content = Content.objects.orphaned(orphan_protection_time, content_pks).exclude(\n  File \"/home/vagrant/devel/pulpcore/pulpcore/app/models/content.py\", line 467, in orphaned\n    expiration = now() - datetime.timedelta(minutes=orphan_protection_time)\n"
    },
    "finished_at": "2022-09-21T11:30:04.959123Z",
    "logging_cid": "60519da7534f4f608fa1fff620671644",
    "name": "pulpcore.app.tasks.orphan.orphan_cleanup",
    "parent_task": null,
    "progress_reports": [],
    "pulp_created": "2022-09-21T11:30:04.885725Z",
    "pulp_href": "/pulp/api/v3/tasks/7bbb10b2-126a-47a8-98ad-a4099547ee13/",
        "/pulp/api/v3/orphans/cleanup/"
    ],
    "started_at": "2022-09-21T11:30:04.923197Z",
    "state": "failed",
    "task_group": null,
    "worker": "/pulp/api/v3/workers/197d8360-3422-4077-8e6a-588a596c8085/"
}
```

jobselko:
This PR adds validation to the `orphans/cleanup/` endpoint to restrict the `orphan_protection_time` value. The allowed range is from 0 to 365 days.

Fixes #3234

### DIFF
--- a/pulpcore/app/serializers/orphans.py
+++ b/pulpcore/app/serializers/orphans.py
@@ -4,6 +4,10 @@ from rest_framework import fields, serializers
 
 from pulpcore.app.models import Content
 from pulpcore.app.serializers import ValidateFieldsMixin
+from pulpcore.constants import (
+    ORPHAN_PROTECTION_TIME_LOWER_BOUND,
+    ORPHAN_PROTECTION_TIME_UPPER_BOUND,
+)
 
 
 class OrphansCleanupSerializer(serializers.Serializer, ValidateFieldsMixin):
@@ -21,6 +25,8 @@ class OrphansCleanupSerializer(serializers.Serializer, ValidateFieldsMixin):
         ),
         allow_null=True,
         required=False,
+        min_value=ORPHAN_PROTECTION_TIME_LOWER_BOUND,
+        max_value=ORPHAN_PROTECTION_TIME_UPPER_BOUND,
     )
 
     def validate_content_hrefs(self, value):
--- a/pulpcore/tests/unit/serializers/test_orphans_cleanup.py
+++ b/pulpcore/tests/unit/serializers/test_orphans_cleanup.py
@@ -1,0 +1,18 @@
+import pytest
+
+from rest_framework import serializers
+
+from pulpcore.app.serializers import OrphansCleanupSerializer
+
+
+@pytest.mark.parametrize("value", [-1, 4294967296])
+def test_orphan_protection_time(value):
+    """
+    Test that OrphansCleanupSerializer is not valid if the "orphan_protection_time"
+    value is not in the range defined by ORPHAN_PROTECTION_TIME_LOWER_BOUND
+    and ORPHAN_PROTECTION_TIME_UPPER_BOUND.
+    """
+    serializer = OrphansCleanupSerializer(data={"orphan_protection_time": value})
+
+    with pytest.raises(serializers.ValidationError):
+        serializer.is_valid(raise_exception=True)
--- a/CHANGES/3234.feature
+++ b/CHANGES/3234.feature
@@ -1,0 +1,1 @@
+Added validation to the `orphans/cleanup/` endpoint to restrict the `orphan_protection_time` value.
--- a/pulpcore/constants.py
+++ b/pulpcore/constants.py
@@ -120,3 +120,8 @@ PROTECTED_REPO_VERSION_MESSAGE = _(
 
 # Valid timestamp format for checkpoint distribution paths.
 CHECKPOINT_TS_FORMAT = "%Y%m%dT%H%M%SZ"
+
+# Allowed range for the ORPHAN_PROTECTION_TIME value in minutes
+# The upper boundary represents an unsigned 32-bit integer and prevents overflow
+ORPHAN_PROTECTION_TIME_LOWER_BOUND = 0
+ORPHAN_PROTECTION_TIME_UPPER_BOUND = 4294967295  # (2^32)-1


pulpcore/app/settings.py
```diff
@@ -291,6 +291,9 @@
 
 # how long to protect ephemeral items in minutes
 ORPHAN_PROTECTION_TIME = 24 * 60
+# allowed range for the ORPHAN_PROTECTION_TIME value in minutes
+ORPHAN_PROTECTION_TIME_LOWER_BOUND = 0
+ORPHAN_PROTECTION_TIME_UPPER_BOUND = 525600  # 365 days
```
jobselko, line 296: The upper bound value is open to discussion

pulpcore/app/viewsets/orphans.py
```diff
@@ -26,6 +28,18 @@ def cleanup(self, request):
         orphan_protection_time = serializer.validated_data.get(
             "orphan_protection_time", settings.ORPHAN_PROTECTION_TIME
         )
+
```
ggainey, line 31: Django fields can specify a list of "validators". There are built-ins for MaxValueValidator and MinValueValidator - here's [an example for MaxValueValidator](https://www.geeksforgeeks.org/limit-the-maximum-value-of-a-numeric-field-in-a-django-model/) It's more "Django" to validate at the serializer (`OrphansCleanupSerializer`)

pulpcore/tests/functional/api/using_plugin/test_orphans.py
```diff
@@ -158,3 +158,34 @@ def test_cleanup_specific_orphans(
     with pytest.raises(pulpcore_bindings.ApiException) as exc:
         pulpcore_bindings.OrphansCleanupApi.cleanup(content_hrefs_dict)
     assert exc.value.status == 400
+
+
+def test_orphans_cleanup_restriction(
```
ggainey, line 163: :+1:

ggainey (changes requested):

pulpcore/app/settings.py
gerrod3: I think the values should both be constants that we set, and not customizable by the user. And I think 2^32 -1 should be the upperbound.

pulpcore/app/viewsets/orphans.py
gerrod3: I would go one step further and use the DRF specific validation for this IntegerField: https://www.django-rest-framework.org/api-guide/fields/#integerfield

pulpcore/app/settings.py
ggainey: I agree with "not user-settable" - this is really to protect that task from "impossible" situations. But I think "a year" is a reasonable upperbounds. it's not just "this is an int" - it's "what are reasonable settings for how often orphan-cleanup should run?", and "once a year" is def the most-conservative "might actually be useful" number of minutes I can think of. At some point, allowing a high number is the equivalent of setting it to "only run manually".

Is that number a year? I could see "quarter", "month", or "week" being "reasonable" upperbounds as well. I don't have a really strong opinion - other than "more than 3K years is excessive" :)

daviddavis: I would definitely go higher than a week or month. If a user (like us) hasn't run orphan protection before then they might choose to use a high number the first time they run it.

mdellweg: Honestly, I see that negative numbers are not useful. Fair.
But why restricting anyone on the older content side (apart from pulp cannot process this number anymore)? So I'd go with Gerrod on this one. And yes definitely not a setting, because it protects the code that (as stated in the issue) fails to transform the provided value into a C integer (32 bit).

Whether we want to give the admin a knob to protect users from specifying 0 and thereby deleting items others might still be interested in, we should discuss in a separate issue (personal opinion).

pulpcore/tests/functional/api/using_plugin/test_orphans.py
```diff
@@ -158,3 +159,36 @@ def test_cleanup_specific_orphans(
...
+        # Verify that the orphan content unit is on disk
+        relative_path = pulpcore_bindings.ArtifactsApi.read(content_unit.artifact).file
+        artifact_path = os.path.join(pulp_settings.MEDIA_ROOT, relative_path)
+        assert os.path.exists(artifact_path) is True
+
+    # Assert that the cleanup of content unit fails due to the wrong "orphan_protection_time" value
+    with pytest.raises((pulpcore_bindings.ApiException, ValidationError)) as exc1:
+        pulpcore_bindings.OrphansCleanupApi.cleanup({"orphan_protection_time": 525601})
+    if isinstance(exc1, pulpcore_bindings.ApiException):
+        assert exc1.value.status == 400
```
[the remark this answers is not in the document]
mdellweg: Probably some repository in your local environment is outdated. pydantic is the new thing we expect. But otoh it prevents the client from ever sending the invalid request to the server in the first place.
If you want a proper test for this, it should not be a "functional" (actually integration) test, but a unittest for the serializer.

pulpcore/app/serializers/orphans.py
```diff
@@ -21,6 +25,8 @@ class OrphansCleanupSerializer(serializers.Serializer, ValidateFieldsMixin):
         ),
         allow_null=True,
         required=False,
+        min_value=ORPHAN_PROTECTION_TIME_LOWER_BOUND,
+        max_value=ORPHAN_PROTECTION_TIME_UPPER_BOUND,
```
mdellweg, line 29: I'm contemplating if having constants here is a bit verbose for saying "Zero" and "maxint"...
But we should leave a hint (in a comment) that this prevents a C uint32 from breaking. Because that's the first question someone will have seeing this in a year or so.

pulpcore/tests/functional/api/using_plugin/test_orphans.py
```diff
@@ -158,3 +159,36 @@ def test_cleanup_specific_orphans(
...
+    settings = pulp_settings
+
+    # Verify that the system contains the orphan content unit
+    content_unit = file_bindings.ContentFilesApi.read(file_random_content_unit.pulp_href)
+
+    if settings.STORAGES["default"]["BACKEND"] == "pulpcore.app.models.storage.FileSystem":
+        # Verify that the orphan content unit is on disk
+        relative_path = pulpcore_bindings.ArtifactsApi.read(content_unit.artifact).file
+        artifact_path = os.path.join(pulp_settings.MEDIA_ROOT, relative_path)
+        assert os.path.exists(artifact_path) is True
```
mdellweg, line 183: I don't think this part is even useful for this test.

jobselko: Yeah, you are right. I do not need any data to hit the `orphans/cleanup` endpoint

pulpcore/tests/functional/api/using_plugin/test_orphans.py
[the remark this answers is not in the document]
jobselko: If I understand it correctly, tests in CI are run differently than in the dev env. A solution could be to use `raw_function` to suppress the Pydantic error in favor of the ApiException error:
```
    self = pulpcore_bindings.OrphansCleanupApi
    with pytest.raises(pulpcore_bindings.ApiException) as exc:
        pulpcore_bindings.OrphansCleanupApi.cleanup.raw_function(
            self, {"orphan_protection_time": 4294967296}
        )
```
I am thinking about unit tests for the serializer you mentioned, and this is probably a better approach than "functional" tests since I am not testing some "custom" behaviour. I will update it.

pulpcore/app/serializers/orphans.py
jobselko: Updated

pulpcore/tests/functional/api/using_plugin/test_orphans.py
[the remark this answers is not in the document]
mdellweg: Hah, never saw `raw_function`. Is that introduces by the newer bindings? Anyway, yes, if you want to know the serializer rejects that input we can go ask it directly.

pulpcore/tests/unit/serializers/test_orphans_cleanup.py
```diff
@@ -0,0 +1,18 @@
...
+def test_orphan_protection_time(value):
+    """
+    Test that OrphansCleanupSerializer is not valid if the "orphan_protection_time"
+    value is not in the range defined by ORPHAN_PROTECTION_TIME_LOWER_BOUND
+    and ORPHAN_PROTECTION_TIME_UPPER_BOUND.
+    """
+    serializer = OrphansCleanupSerializer(data={"orphan_protection_time": value})
+
+    with pytest.raises(serializers.ValidationError):
+        serializer.is_valid(raise_exception=True)
```
mdellweg, line 18: Perfect!

mdellweg (approved):

ggainey (approved):
Great - thanks for your patience :)

pulpcore/tests/functional/api/using_plugin/test_orphans.py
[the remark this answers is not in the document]
jobselko: `raw_function` is purely pydantic thing. I think that either CI or local tests contain some "bug" that causes different errors to propagate.

PR #6371, opened 2025-03-19
Reworked the two pass settings loading

mdellweg:
The values of the first pass can now be used to adjust the defaults for the second pass, but the results will be discarded, becuase reusing lead to weird interferences with environment injection.

### DIFF
--- a/CHANGES/6341.bugfix
+++ b/CHANGES/6341.bugfix
@@ -1,0 +1,1 @@
+Reworked the loading of plugin default settings in order to fix interference with environment injection.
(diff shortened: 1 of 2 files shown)


pedro-psb (approved):
Did an experiment locally with that, the order looks correct now.

I've added a dummy settings to pulpcore settings.py, to pulp_rpm settings.py, to the user settings and to the environ.

Then run `dynaconf inspect` agaisnt the checkouts:

```bash
$ # From main checkout
$ dynaconf inspect -k dummy | jq '.history[].value'
"from pulp_rpm"
"from user environ"
"from user settings.py"
"from core setting.py"

$ # From this PR checkout
$ dynaconf inspect -k dummy | jq '.history[].value'
"from user environ"
"from user settings.py"
"from pulp_rpm"
"from core setting.py"
```

pedro-psb (commented):
Hold on, I'm running more tests on this.

pedro-psb (approved):
Ok, this looks good.
I was testing the plugin enablement and having some issues, but I was applying the setting the wrong way.

Issue #6385, opened 2025-03-25
Pull-through caching fails when the content already exists on demand


gerrod3:
When using pull-through caching and requesting Content already in Pulp but still on-demand, the saving part will fail with the message `"The Artifacts created during pull-through does not match the Artifacts already stored for the same content."` Pull-through caching should handle updating on-demand Content's artifacts gracefully like we do for uploads.

PR #6386, opened 2025-03-25
Fix pull-through cache for on-demand content

gerrod3:
fixes: #6385

Now follows the same logic used in upload: https://github.com/pulp/pulpcore/blob/faeecb95f74cee763c5e51cb0d840e06335ce70b/pulpcore/app/serializers/content.py#L89-L102

### DIFF
--- a/pulpcore/tests/unit/content/test_handler.py
+++ b/pulpcore/tests/unit/content/test_handler.py
@@ -5,6 +5,7 @@ import uuid
 from unittest.mock import Mock, AsyncMock
 
 from aiohttp.web_exceptions import HTTPMovedPermanently
+from django.db import IntegrityError
 from pulpcore.content.handler import Handler, CheckpointListings, PathNotResolved
 from pulpcore.plugin.models import (
     Artifact,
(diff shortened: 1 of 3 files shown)


pulpcore/tests/unit/content/test_handler.py
```diff
@@ -344,6 +345,39 @@ def content_init(art, path):
...
+    """Ensure single-artifact content is properly saved on pull-through."""
+    handler = Handler()
+    remote123.get_remote_artifact_content_type = Mock(return_value=Content)
+    content = Content.objects.create()
+    content.save = Mock(side_effect=IntegrityError)
+    content_init_mock = Mock(return_value=content)
+    monkeypatch.setattr(Content, "init_from_artifact_and_relative_path", content_init_mock)
+    monkeypatch.setattr(Content.objects, "get", Mock(return_value=content))
+    ca = ContentArtifact(relative_path="c123")
+    ra = RemoteArtifact(url=f"{remote123.url}/c123", remote=remote123, content_artifact=ca)
```
dralley, line 360: Q) What's the state of the DB at this point, should we assert that this object doesn't already exist or is it guaranteed to be clean?

dralley (approved):

pulpcore/tests/unit/content/test_handler.py
gerrod3: The Content object is guaranteed to be new so the ContentArtifact and RemoteArtifact are also guaranteed to not be in the db since they ultimately depend on the Content object.

Issue #6388, opened 2025-03-26
RepositoryVersion fields are not properly validated in the cross-domain check


gerrod3:
To ensure domain isolation user input must be checked for cross domain objects which is done in the `ValidateFieldsMixin`. We check each field and see if the field has a `pulp_domain` relationship. If so we check to see that it is a part of the current domain. `RepositoryVersion` is not properly handled as an object that is scoped by domain since it does not have a direct `pulp_domain` relationship. This means all `Publication` serializers currently accept cross domain `RepositoryVersions`, same with `Distribution` and any serializer that has a `RepositoryVersion` field.

PR #6389, opened 2025-03-26
Fix cross-domain check for RepositoryVersion fields

gerrod3:
fixes: #6388

### DIFF
--- a/pulp_file/tests/functional/api/test_domains.py
+++ b/pulp_file/tests/functional/api/test_domains.py
@@ -340,6 +340,16 @@ def test_no_cross_pollination(
         "non_field_errors": [f"Objects must all be a part of the {domain.name} domain."]
     }
 
+    with pytest.raises(file_bindings.ApiException) as e:
+        file_publication_factory(
+            repository_version=d_repo.latest_version_href,
+            pulp_domain=domain.name,
+        )
+    assert e.value.status == 400
+    assert json.loads(e.value.body) == {
+        "non_field_errors": [f"Objects must all be a part of the {domain.name} domain."]
+    }
+
     # Fail distribute
     with pytest.raises(file_bindings.ApiException) as e:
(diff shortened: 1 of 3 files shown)


pulp_file/tests/functional/api/test_domains.py
```diff
@@ -340,6 +340,16 @@ def test_no_cross_pollination(
         "non_field_errors": [f"Objects must all be a part of the {domain.name} domain."]
     }
 
+    with pytest.raises(file_bindings.ApiException) as e:
```
mdellweg, line 343: Just an idea: Can we split this test into a sequence of tests in a testclass where e.g. the `repo_d` is a class shared fixture?

pulpcore/app/serializers/base.py
```diff
@@ -333,15 +333,16 @@ def check_cross_domains(self, data):
...
-                if isinstance(value, Model) and (
-                    domain_id := getattr(value, "pulp_domain_id", None)
-                ):
-                    if current_domain.pulp_id != domain_id:
-                        raise serializers.ValidationError(
-                            _("Objects must all be a part of the {} domain.").format(
-                                current_domain.name
+                if isinstance(value, Model):
+                    if isinstance(value, RepositoryVersion):
+                        value = value.repository
```
mdellweg, line 338: This feels like trying to seal one hole after another.
How about scoping the model managers to the current domain only? It should be equally safe than the queryset scoping for RBAC that we do.

gerrod3: I remember trying to do that originally in my first implementation of domains, but I could never get it to work. The main problem is that all the viewsets and serializer fields take in a queryset (e.g. `RepositoryVersion.objects.all()`) at Pulp startup before a request so we don't know the domain to scope by and I couldn't get a lazy filter to work.

mdellweg (approved):

PR #6412, opened 2025-03-31
Fix grouping for admin guides in docs

mdellweg:


pedro-psb:
Oh, I see :sweat_smile: 
image

PR #6428, opened 2025-04-02
Set the X-PULP-ARTIFACT-SIZE header at the handle_response_header function.

### LINKED ISSUE #6390: Error trying to pull content from remote pull-through cache
**Describe the bug**
Trying to pull content from a remote pull-through cache is failing with:
```
TypeError("int() argument must be a string, a bytes-like object or a number, not 'NoneType'")
```

**To Reproduce**
```
curl -uadmin:$PASSWD -X POST $BASE_ADDR/api/pulp/${domain}/api/v3/repositories/npm/npm/ -d '{"name": "test-npm"}' -H 'Content-Type: application/json'
curl -uadmin:$PASSWD  -X POST $BASE_ADDR/api/pulp/${domain}/api/v3/remotes/npm/npm/ -d '{"name": "test-npm", "url": "https://registry.npmjs.org/"}' -H 'Content-Type: application/json'
curl -uadmin:$PASSWD  -X POST $BASE_ADDR/api/pulp/default/api/v3/repositories/npm/npm/0195b9e8-f732-7190-bcde-7584390d52da/sync/ remote=/pulp/default/api/v3/remotes/npm/npm/0195d229-fe90-727d-9c33-1b7c6f51fe2e/
curl -uadmin:$PASSWD -X POST $BASE_ADDR/api/pulp/${domain}/api/v3/distributions/npm/npm/ -d '{"name": "test3", "base_path": "test3", "remote": "/api/pulp/hyagi-s3/api/v3/remotes/npm/npm/0195d292-0a2e-7d01-862d-437ebd748d22/"}' -H 'Content-Type: application/json'
npm install --registry $BASE_ADDR/api/pulp-content/${domain}/test3 react@19.0.0
```

```
[2025-03-26 15:10:01 +0000] [10] [ERROR] Web-handler should return a response instance, got TypeError("int() argument must be a string, a bytes-like object or a number, not 'NoneType'")                                                                                                                                                                                  
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/aiohttp/web_protocol.py", line 658, in finish_response                                                                                                                                                                                                                                                           
    prepare_meth = resp.prepare                                                                                                                                                                                                                                                                                                                                            
AttributeError: 'TypeError' object has no attribute 'prepare'   
```

**Additional context**

As a workaround, modifying the following lines "fixes" the error:
https://github.com/pulp/pulpcore/blob/6b12bb2ce366dae011838d58e37127a1baefe28b/pulpcore/content/handler.py#L1316-L1318

```
        else:
-            response.headers["X-PULP-ARTIFACT-SIZE"] = str(size)
-            artifacts_size_counter.add(size)
+            if size:
+                response.headers["X-PULP-ARTIFACT-SIZE"] = str(size)
+                artifacts_size_counter.add(size)
```

decko:
Set the X-PULP-ARTIFACT-SIZE header at the handle_response_header function
using the original `Content-Length` header when a `Content-Encoding` header
is not present. Also, remove old code that weren't used to set the header.

Close #6390

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -1248,6 +1248,13 @@ class Handler:
                     response.headers["Content-Range"] = "bytes {0}-{1}/{2}".format(
                         start, stop - 1, content_length
                     )
+
+            content_encoding = headers.get("content-encoding")
+            content_length = headers.get("content-length")
+            if content_length and not content_encoding:
+                response.headers["X-PULP-ARTIFACT-SIZE"] = content_length
+                artifacts_size_counter.add(content_length)
+
             await response.prepare(request)
 
         data_size_handled = 0
@@ -1311,13 +1318,6 @@ class Handler:
                 "affected Pulp Remote, adding a good one and resyncing.\n"
                 "If the problem persists, please contact the Pulp team."
             )
-
-        if content_length := response.headers.get("Content-Length"):
-            response.headers["X-PULP-ARTIFACT-SIZE"] = content_length
-            artifacts_size_counter.add(content_length)
-        else:
-            response.headers["X-PULP-ARTIFACT-SIZE"] = str(size)
-            artifacts_size_counter.add(size)
 
         if save_artifact and remote.policy != Remote.STREAMED:
             content_artifacts = await asyncio.shield(
--- a/CHANGES/6390.bugfix
+++ b/CHANGES/6390.bugfix
@@ -1,0 +1,1 @@
+Set the X-PULP-ARTIFACT-SIZE header at the handle_response_header function.


pulpcore/content/handler.py
```diff
@@ -1315,9 +1321,12 @@ async def finalize():
         if content_length := response.headers.get("Content-Length"):
             response.headers["X-PULP-ARTIFACT-SIZE"] = content_length
             artifacts_size_counter.add(content_length)
-        else:
+        elif size:
```
dkliban, line 1324: You should remove lines 1324-1329. This code runs after the response has been started and doesn't actually modify the headers being sent to the client.

decko: Actually, removing that entire block seems to be the right thing.

bmbouter (approved):

pulpcore/content/handler.py
```diff
@@ -1227,6 +1227,12 @@ async def handle_response_headers(headers):
                 lower_name = name.lower()
                 if lower_name not in self.hop_by_hop_headers:
                     response.headers[name] = value
+                elif (
+                    response.status == 200
+                    and lower_name == "content-length"
+                    and not headers.get("Content-Encoding")
```
dralley, line 1233: do we need to be concerned with casing here?

bmbouter: @decko it would also be more efficient to check it once, e.g. `content_encoding = headers.get("Content-Encoding")` which I expect to return `None` if `Content-Encoding` is not present.

pulpcore/content/handler.py
```diff
@@ -1227,6 +1227,12 @@ async def handle_response_headers(headers):
                 lower_name = name.lower()
                 if lower_name not in self.hop_by_hop_headers:
                     response.headers[name] = value
+                elif (
+                    response.status == 200
```
bmbouter, line 1231: Why the 200 check?

pulpcore/content/handler.py
bmbouter: Also, for efficiency, you can move this out of the whole loop itself. You don't need to check for the header (with each header), you can just handle it as a case below the loop entirely. That would be more efficient also.

pulpcore/content/handler.py
decko: This method have a special path for 206.

dralley: Why not just make sure the 206 check happens first?

decko: @dralley I believe that 200 Ok happens way more times than 206, no?

bmbouter: It was hard for me to describe, so here's a code snippet. Overall the idea is to get it out of the loop and not worry about the response code.

```
diff --git a/pulpcore/content/handler.py b/pulpcore/content/handler.py
index 2946d9edc..ec1b6d852 100644
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -1248,6 +1248,10 @@ class Handler:
                     response.headers["Content-Range"] = "bytes {0}-{1}/{2}".format(
                         start, stop - 1, content_length
                     )
+            content_length = headers.get("content-length")
+            if not headers.get("content-encoding") and content_length:
+                response.headers["X-PULP-ARTIFACT-SIZE"] = content_length
+                artifacts_size_counter.add(content_length)
             await response.prepare(request)
 
         data_size_handled = 0
```

bmbouter (changes requested):
See the comments, let me know what you think. Ty!

pulpcore/content/handler.py
```diff
@@ -1223,6 +1223,7 @@ async def _stream_remote_artifact(
                 actual_content_length = size - start
 
         async def handle_response_headers(headers):
+            content_encoding = headers.get("content-encoding")
```
bmbouter, line 1226: tiny thing, just move this down with the block where it's used. Readability benefits from "locality" where definition and usage is as close as possible.

bmbouter (approved):
This looks great to me.

dkliban (approved):

pulpcore/content/handler.py
```diff
@@ -1248,6 +1248,13 @@ async def handle_response_headers(headers):
                     response.headers["Content-Range"] = "bytes {0}-{1}/{2}".format(
                         start, stop - 1, content_length
                     )
+
+            content_encoding = headers.get("content-encoding")
+            content_length = headers.get("content-length")
+            if content_length and not content_encoding:
+                response.headers["X-PULP-ARTIFACT-SIZE"] = content_length
+                artifacts_size_counter.add(content_length)
```
dralley, line 1256: @bmbouter is it OK if this doesn't handle range requests?  There was an attempt to add to the size counter on both sides of the branch in the old version

dralley: >I think if range headers are used for the streaming case we just broke it with this change because we'll overwrite the content-length here. @dralley do you agree with this assessment?

I wasn't thinking along the lines of "overwriting this value will break things", I was just thinking that we're no longer adding to the size counter in cases where we were doing so before (notably range requests I think).

decko: @bmbouter we're not breaking because we're not overwriting anything here. We're just using the `content-length` value to set the `X-PULP-ARTIFACT-SIZE` header and also adding it to the `artifacts_size` metric.

decko: What we're missing here, as @dralley said ☝🏽, is the case of ranged request. Yet, it's only when we're streaming content as pull-through cache first hit, so I don't think we will get a 206 when it's the first time the user pulls content from a pull-through remote. 

But maybe I could be wrong here...

bmbouter: @dralley @decko thank you both for the replies. I was thinking about this incorrectly.

In the case where it's streamed and range requests are being used, I'm comfortable for us to omit the data from the `X-PULP-ARTIFACT-SIZE`. I think we're done on this bugfix at this point. Thank you both!

dralley: And also from the metric counter?

decko: Yes @dralley. The X-PULP-ARTIFACT-SIZE header was originally thought to be used by the metric tooling to be consumed in a middleware. Now it's also used in some log analysis. 

And we're ok on omitting the ranged request data for that.

PR #6435, opened 2025-04-03
Fix second scenario for on-demand pull-through caching

gerrod3:
My initial fix forgot to account/test for the second condition in the if statement on line 1065:
https://github.com/pulp/pulpcore/blob/8b72c898a018d5f7954c4998f53ab654d79682d2/pulpcore/content/handler.py#L1064-L1075

When the ContentArtifact already exists then the `bulk_create` will update the exisiting object, but the returned value of this will be the objects from `new_cas`, not the actual saved values from the DB. This results in the later RemoteArtifact saving failing since the ContentArtifact in `cas` would have the wrong `pk`

### DIFF
--- a/CHANGES/+second_ondemand_pullthrough.bugfix
+++ b/CHANGES/+second_ondemand_pullthrough.bugfix
@@ -1,0 +1,1 @@
+Fixed an oversight in the #6385 fix.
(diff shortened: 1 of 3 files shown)


pedro-psb (approved):
LGTM

PR #6391, opened 2025-03-27
Validate repo and remote in RepositorySyncURLSerializer

jobselko:
This PR adds validation to the `RepositorySyncURLSerializer` to check if the repo and remote are from different plugins. If they are, an error is raised before the task is spawned.

Validation for publication and distribution will be added later.

Fixes #1860

PR #6436, opened 2025-04-03
Validate repo and remote in PublicationSerializer and DistributionSerializer

### LINKED ISSUE #6174: Pulp allows you to create a distribution for another repository type
I noticed that we had an apt distro with an rpm repository. I think this is might be a Pulp bug in that Pulp should prevent this from happening?

Steps to reproduce:

```
pulp rpm repository create --name test-yum

# note you have to use the api here because the cli checks the href type
http ${SERVER}/pulp/api/v3/distributions/deb/apt/ repository=/pulp/api/v3/repositories/rpm/rpm/01942cdc-3ae2-7b
ab-9c52-2157b5e14cce/ base_path=test-apt name=test-apt

pulp deb distribution list
```

Output:

```
[
  {
    "pulp_href": "/pulp/api/v3/distributions/deb/apt/01942cdd-9f20-7f09-89f3-ae06d0c80f4b/",
    "prn": "prn:deb.aptdistribution:01942cdd-9f20-7f09-89f3-ae06d0c80f4b",
    "pulp_created": "2025-01-03T15:51:49.281526Z",
    "pulp_last_updated": "2025-01-03T15:51:49.281539Z",
    "base_path": "test-apt",
    "base_url": "https://pulp.example.com/pulp/content/test-apt/",
    "content_guard": null,
    "no_content_change_since": null,
    "hidden": false,
    "pulp_labels": {},
    "name": "test-apt",
    "repository": "/pulp/api/v3/repositories/rpm/rpm/01942cdc-3ae2-7bab-9c52-2157b5e14cce/",
    "publication": null
  },
]
```

jobselko:
This PR adds validation to the `PublicationSerializer` and `DistributionSerializer` to check if the repo and remote are from different plugins. If they are, an error is raised before the task is spawned.

Follow-up to #6391

Fixes #6174

### DIFF
--- a/pulpcore/app/serializers/publication.py
+++ b/pulpcore/app/serializers/publication.py
@@ -1,4 +1,5 @@
 from gettext import gettext as _
+from pulpcore.app.util import get_prn
 
 from django.db.models import Q
 from drf_spectacular.utils import extend_schema_field
@@ -42,12 +43,18 @@ class PublicationSerializer(ModelSerializer):
                 _("Either the 'repository' or 'repository_version' need to be specified")
             )
         elif not repository and repository_version:
+            validate_repo_and_remote(
+                repository_version.repository,
+                "repository_version",
+                repository_version=repository_version,
+            )
             if checkpoint:
                 raise serializers.ValidationError(
                     _("'checkpoint' may only be used with 'repository'")
                 )
             return data
         elif repository and not repository_version:
+            validate_repo_and_remote(repository, "repository")
             version = repository.latest_version()
             if version:
                 new_data = {"repository_version": version}
@@ -330,6 +337,32 @@ class DistributionSerializer(ModelSerializer):
                 _("The 'checkpoint' attribute may only be used with the 'repository' attribute.")
             )
 
+        # Validate that repository and its remote are of a compatible type
+        if repository_provided:
+            validate_repo_and_remote(repository_provided, "repository")
+
+        if repository_version_provided:
+            validate_repo_and_remote(
+                repository_version_provided.repository,
+                "repository_version",
+                repository_version=repository_version_provided,
+            )
+
+        if publication_provided and publication_provided.repository:
+            validate_repo_and_remote(
+                publication_provided.repository,
+                "publication",
+                publication=publication_provided,
+            )
+
+        if publication_provided and publication_provided.repository_version:
+            validate_repo_and_remote(
+                publication_provided.repository_version.repository,
+                "publication",
+                publication=publication_provided,
+                repository_version=publication_provided.repository_version,
+            )
+
         return data
 
     def get_no_content_change_since(self, obj):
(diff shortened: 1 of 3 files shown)


pulpcore/app/serializers/publication.py
```diff
@@ -355,3 +388,33 @@ class Meta:
                 "repository",
             }
         )
+
+
+def validate_repo_and_remote(repository, field, **kwargs):
```
jobselko, line 393: Would it make sense to move this function to `ModelSerializer` since both `PublicationSerializer` and `DistributionSerializer` inherit from it?

mdellweg: I'd rather not.

pulpcore/app/serializers/publication.py
```diff
@@ -355,3 +388,33 @@ class Meta:
...
+    """
+    Validate that a repository has a remote of the same plugin.
```
mdellweg, line 395:
```suggestion
    Validate that a repository has a remote of a compatible type.
```

pulpcore/app/serializers/publication.py
```diff
@@ -355,3 +388,33 @@ class Meta:
...
+        **kwargs: additional data for the error message (repository_version or publication)
+
+    Returns:
+        None or ValidationError
+    """
+    if (
+        repository
+        and repository.cast().remote
+        and type(repository.cast().remote.cast()) not in repository.cast().REMOTE_TYPES
+    ):
```
mdellweg, line 409: Sounds like it could be useful to start with "repository = repository.cast()"...

pulpcore/tests/unit/serializers/test_repository.py
```diff
@@ -167,3 +175,83 @@ def test_validate_repo_remote_sync(monkeypatch):
...
         serializer.validate(data)
+
+
+@pytest.mark.parametrize("tested_serializer", [PublicationSerializer(), DistributionSerializer()])
+def test_validate_repo_remote_via_repository(monkeypatch, tested_serializer):
+    """
+    Test that the publication/distribution fails if
+    "repository" has a remote of an incompatible type.
+    """
+    monkeypatch.setattr(ValidateFieldsMixin, "CHECK_SAME_DOMAIN", False)
```
mdellweg, line 186: I'd rather patch `check_cross_domains` to `lambda ...: None`. I had to look this one up to understand.
Would that work?
```suggestion
    monkeypatch.setattr(tested_serializer, "check_cross_domain", lambda self, data: None)
```

mdellweg (commented):
Two more suggestions. The change looks good.

pulpcore/app/serializers/publication.py
mdellweg: Yes, i think that would be it.
On the pro side: Two less calls to cast. And that may even mean two round trips to the database that we can spare.

PR #6444, opened 2025-04-04
Sort settings.md alphabetically and fix links

jobselko:
This PR updates `settings.md` with the following changes:
- alphabetical sorting of settings
- fixing typos, broken links and numbering order
- adding missing Kafka settings into heading
- consolidating empty lines - two empty lines before `##` section, one empty line otherwise (this is purely for better orientation in code editor)

I intentionally split the changes into multiple commits for easier review and will squash it at the end.


docs/admin/reference/settings.md
jobselko, line 1: Should I add an entry into `CHANGES` dir for towncrier as well?

docs/admin/reference/settings.md
```diff
@@ -1,44 +1,48 @@
...
 - `SECRET_KEY <secret-key-setting>`
 
-Pulp uses three types of settings:
+Pulp uses four types of settings:
 
 - `Django settings` Pulp is configuring
-- `Redis settings` Pulp is using
+- `Kafka settings`
 - `Pulp defined settings`
+- `Redis settings` Pulp is using
```
jobselko, line 12: Should I add some details to `Kafka settings`?

docs/admin/reference/settings.md
```diff
@@ -115,8 +114,8 @@ Overview of the integration with Pulp:
 
 - **Supported**: We support (test) Amazon S3 and Azure.
 - **Untested**: Other backends provided by `django-storages` may work as well, but we provide no guarantee.
-- **Known caveat**: Using SFTP Storage is not recommended in Pulp's current state, and doing so can lead to file corruption. This is because Pulp currently uses coroutines that seem to be incompatible with Django's SFTPStorage implementation.
-
+- **Known caveat**: Using SFTP Storage is not recommended in Pulp's current state, and doing so can lead to file corruption.
+This is because Pulp currently uses coroutines that seem to be incompatible with Django's SFTPStorage implementation.
```
dralley, line 162: I don't understand this change, this line is part of the description of the known caveat, but it looks like it's been moved out of that context?

docs/admin/reference/settings.md
dralley: That would be reasonable.  It doesn't have to be particularly specific or thorough, though.

docs/admin/reference/settings.md
```diff
@@ -38,7 +38,7 @@ print(''.join(secrets.choice(chars) for i in range(50)))
 The file location of a symmetric fernet key that Pulp uses to encrypt sensitive fields in the
 database. Default location is `/etc/pulp/certs/database_fields.symmetric.key`.
 The key is automatically generated by default with the pulp-oci-images. The key can be generated
-independently but it must be a url-safe base64-encoded string of 32 random bytes.
+independently, but it must be an url-safe base64-encoded string of 32 random bytes.
```
dralley, line 45: In this case "a url-safe" was correct.  This is one of those cases where English is (perhaps annoyingly) inconsistent.  Since the sound the "u" makes is the same as "you" - a consonant sound - you use "a" instead of "an".

https://owl.purdue.edu/owl/general_writing/grammar/articles_a_versus_an.html

But I totally get why that's confusing to non-native speakers.  Even I had to look up the precise rationale behind the exception :)

jobselko: Thanks a lot for spotting this. English is tricky :)

docs/admin/reference/settings.md
jobselko: I split this line because it was too long, and I do not like never-ending lines in my editor. This change will not affect rendering, the line will be shown exactly the same way as it did before. I understand that this might be debatable, and I can revert it if you prefer the original version.

docs/admin/reference/settings.md
jobselko: Do I need to create a new pulpcore issue for this, or is it doable without one?

mdellweg: When you don't have an issue you can create a file starting with a "+" sign.

https://towncrier.readthedocs.io/en/stable/cli.html#towncrier-create

docs/admin/reference/settings.md
dralley: As long as it renders properly!

(also, I thought there was an extra empty line in there when I looked at this, but that was mistaken)

docs/admin/reference/settings.md
jobselko: I do not like the "hash" idea, so I created an issue instead (#6459).

dralley (approved):

docs/admin/reference/settings.md
dralley: All of these could use a little bit more, IMO.  But it's not a huge deal.

mdellweg:
Can you squash the commits into reasonable chunks before merging?

Issue #6429, opened 2025-04-02
Pull-through regression related to immediate task changes


pedro-psb:
Reported by @decko [on matrix](https://matrix.to/#/!aVApiNMtnstWbwDcVU:matrix.org/$ZOQIBi5ilUcYyKFiudUmrVlXLduCDJKarF8eQsIPdkQ?via=matrix.org&via=ctrl-c.liu.se&via=x9c4.spdns.org).

## Version
```bash
pulpcore>=3.75 
# Any pull-through enabled plugin, e.g npm
```

## Description

**GIVEN** a distribution with pull-through cache enabled
**AND** the save content feature enabled
**WHEN** trying to fetch an artifact not seem by Pulp before
**THEN** the request fails on the asyncio immediate task handler

## To Reproduce

```bash
#/bin/sh
set -x

# Setup the pull-through distribution
ROUND=0
HOST=http://localhost:5001
ENDPOINT=/pulp
DOMAIN=/default
REPO_ENDPOINT=/api/v3/repositories/npm/npm/
RMT_ENDPOINT=/api/v3/remotes/npm/npm/
DST_ENDPOINT=/api/v3/distributions/npm/npm/
VERBOSE=-vvv
AUTH='-u admin:password'
PULP_INSTANCE=$HOST$ENDPOINT$DOMAIN

LOCAL_TEST_REMOTE=$(curl "$VERBOSE" -u admin:password "${PULP_INSTANCE}${RMT_ENDPOINT}" --json '{"name": "npm_test'"$ROUND"'", "url": "https://registry.npmjs.org"}' | jq '.pulp_href')
LOCAL_TEST_REPOSITORY=$(curl "$VERBOSE" -u admin:password "${PULP_INSTANCE}${REPO_ENDPOINT}" --json '{"name": "npm_test'"$ROUND"'", "remote": '${LOCAL_TEST_REMOTE}'}' | jq '.pulp_href')
LOCAL_TEST_DISTRIBUTION=$(curl "$VERBOSE" -u admin:password "${PULP_INSTANCE}${DST_ENDPOINT}" --json '{"name": "npm_test'"$ROUND"'", "remote": '${LOCAL_TEST_REMOTE}', "repository": '${LOCAL_TEST_REPOSITORY}', "base_path": "npm'"$ROUND"'"}')

# Request an artifact not seen by pulp before
[...]

PR #6446, opened 2025-04-04
Revert "Add timeout to immediate task execution"

pedro-psb:
This reverts commit 6b12bb2ce366dae011838d58e37127a1baefe28b.

---

This commits caused a regression with pull-through caching.
We have an approach for fixing it, but decided to revert to unblock releases and not rush its implementation.

See the regression issue: https://github.com/pulp/pulpcore/issues/6429


bmbouter (approved):
I'm basing my review on this just being a git revert. Thank you @pedro-psb !

PR #6471, opened 2025-04-11
Update tests to account for existing data

jobselko:
This PR updates the `test_specified_all_repos` and `test_upload_empty_file` tests to account for existing data (repositories and artifacts) that may already be present on the system.

### DIFF
--- a/pulpcore/tests/functional/api/test_crd_artifacts.py
+++ b/pulpcore/tests/functional/api/test_crd_artifacts.py
@@ -20,7 +20,12 @@ def pulpcore_random_file(tmp_path):
 
 def _do_upload_valid_attrs(pulpcore_bindings, file, data):
     """Upload a file with the given attributes."""
-    artifact = pulpcore_bindings.ArtifactsApi.create(str(file), **data)
+    # Create or get artifact if it already exists
+    try:
+        artifact = pulpcore_bindings.ArtifactsApi.create(str(file), **data)
+    except pulpcore_bindings.ApiException:
+        artifact = pulpcore_bindings.ArtifactsApi.list(sha256=data.get("sha256")).results[0]
+
     # assumes ALLOWED_CONTENT_CHECKSUMS does NOT contain "md5"
     assert artifact.md5 is None, "MD5 {}".format(artifact.md5)
     read_artifact = pulpcore_bindings.ArtifactsApi.read(artifact.pulp_href)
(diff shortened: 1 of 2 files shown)


pulpcore/tests/functional/api/test_crd_artifacts.py
```diff
@@ -20,7 +20,12 @@ def pulpcore_random_file(tmp_path):
 
 def _do_upload_valid_attrs(pulpcore_bindings, file, data):
     """Upload a file with the given attributes."""
-    artifact = pulpcore_bindings.ArtifactsApi.create(str(file), **data)
+    # Create or get artifact if it already exists
+    try:
+        artifact = pulpcore_bindings.ArtifactsApi.create(str(file), **data)
+    except pulpcore_bindings.ApiException:
+        artifact = pulpcore_bindings.ArtifactsApi.list(sha256=data.get("sha256")).results[0]
```
dralley, line 27: Is this actually what we want?  If the goal is to test upload / artifact creation, you arguably *do not* want it to silently use already-created data, you want it to do the upload and fail if it cannot.  At least not unless the test itself wants to re-use data.

pulpcore/tests/functional/api/using_plugin/test_reclaim_disk_space.py
```diff
@@ -149,11 +149,15 @@ def test_immediate_reclaim_becomes_on_demand(
...
+    file_bindings,
     file_repository_factory,
     monitor_task,
 ):
     """Tests that specifying all repos w/ '*' properly grabs all the repos."""
-    repos = [file_repository_factory().pulp_href.split("/")[-2] for _ in range(10)]
+    # Get existing file repositories or create new ones
+    repos = [
+        r.pulp_href.split("/")[-2] for r in file_bindings.RepositoriesFileApi.list().results
+    ] or [file_repository_factory().pulp_href.split("/")[-2] for _ in range(10)]
```
dralley, line 160: Likewise.  Also some tests run in parallel, are we sure that none of those existing repos will disappear from under us, or that we won't be messing with their results?

The freshly created repos are guaranteed to be empty, which is fine for the purposes of this test.

pulpcore/tests/functional/api/test_crd_artifacts.py
jobselko: Would you prefer to delete artifacts/repositories if they already exist? Currently, if you run the tests multiple times on your local machine, these two of them fail. I would like to fix this behavior and I am okay with any solution that works.

Ideally, I would expect the tests to use some kind of fake or temporary storage so that repeated runs do not affect the results, but maybe there is a reason why the tests use the real storage.

dralley: Using random data is probably the best solution here.  It is admittedly much more difficult with other plugins.

PR #6480, opened 2025-04-14
Fix add_roles_for_object_creator handling of AnonymousUser

gerrod3:
Sometimes our auth (pulp_container) can return an AnonymousUser which will make `add_roles_for_object_creator` fail if they perform an action to create an object.

### DIFF
--- a/pulpcore/app/models/access_policy.py
+++ b/pulpcore/app/models/access_policy.py
@@ -140,7 +140,7 @@ class AutoAddObjPermsMixin:
 
         roles = _ensure_iterable(roles)
         current_user = get_current_authenticated_user()
-        if current_user:
+        if current_user and isinstance(current_user, get_user_model()):
             for role in roles:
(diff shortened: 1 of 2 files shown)


mdellweg (approved):

pulpcore/app/models/access_policy.py
```diff
@@ -140,7 +140,7 @@ def add_roles_for_object_creator(self, roles):
 
         roles = _ensure_iterable(roles)
         current_user = get_current_authenticated_user()
-        if current_user:
+        if current_user and not isinstance(current_user, AnonymousUser):
```
mdellweg, line 143: Should we maybe check for `isinstance(current_user, User)` instead? If an authentication framework was used that did not use the `User = get_user_model()` we could not assign roles in the database anyway.
(Then again, maybe such an installation should not use our `AccessPolicy` anyway.)

PR #6483, opened 2025-04-15
Add the task_name and if it's a task.immediate to the task log entry.

### LINKED ISSUE #6482: Missing some task related information on task log entries.
Currently, a task entry in the logs is missing some interesting information about the task itself.
The `task_name` and if it's an immediate task, can help Pulp administrators to understand how the
tasks are distributed between its types and even calculate the impact it have on resource usage.

I would like to have the `task_name` and the `task.immediate` info on each `Starting task ...` log entry.

decko:
Close #6482

### DIFF
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -60,7 +60,14 @@ def _execute_task(task):
     task.set_running()
     domain = get_domain()
     try:
-        _logger.info(_("Starting task %s in domain: %s"), task.pk, domain.name)
+        _logger.info(
+            "Starting task id: %s in domain: %s, task_type: %s, immediate: %s, deferred: %s",
+            task.pk,
+            domain.name,
+            task.name,
+            str(task.immediate),
+            str(task.deferred),
+        )
 
         # Execute task
         module_name, function_name = task.name.rsplit(".", 1)
@@ -70,7 +77,7 @@ def _execute_task(task):
         kwargs = task.enc_kwargs or {}
         result = func(*args, **kwargs)
         if asyncio.iscoroutine(result):
-            _logger.debug(_("Task is coroutine %s"), task.pk)
+            _logger.debug("Task is coroutine %s", task.pk)
             loop = asyncio.get_event_loop()
             loop.run_until_complete(result)
 
@@ -78,7 +85,7 @@ def _execute_task(task):
         exc_type, exc, tb = sys.exc_info()
         task.set_failed(exc, tb)
         _logger.info(
-            _("Task[{task_type}] {task_pk} failed ({exc_type}: {exc}) in domain: {domain}").format(
+            "Task[{task_type}] {task_pk} failed ({exc_type}: {exc}) in domain: {domain}".format(
                 task_type=task.name,
                 task_pk=task.pk,
                 exc_type=exc_type.__name__,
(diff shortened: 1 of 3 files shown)


pulpcore/tasking/tasks.py
```diff
@@ -60,7 +60,10 @@ def _execute_task(task):
     task.set_running()
     domain = get_domain()
     try:
-        _logger.info(_("Starting task %s in domain: %s"), task.pk, domain.name)
+        if task.immediate:
+            _logger.info(_("Starting immediate task id: %s task_type: %s in domain: %s"), task.pk, task.name, domain.name)
```
bmbouter, line 64: This shouldn't get i8ln. I know the log statement you're updating had the `_()` but it shouldn't. This is because when users send us logs to debug, developers need them to show up in English. We only i8ln user-facing string.

bmbouter:
I think it would be useful to be able to know (from logging) how many immediate tasks are being run on pulp-api versus on a pulp worker.

Does pulp-api also call `def _execute_task(task):`? I'm not sure how the tasking code executes when it's within pulp-api.

decko:
> I think it would be useful to be able to know (from logging) how many immediate tasks are being run on pulp-api versus on a pulp worker.
> 
> Does pulp-api also call `def _execute_task(task):`? I'm not sure how the tasking code executes when it's within pulp-api.

@bmbouter check this out https://github.com/pulp/pulpcore/blob/d6262b097990e0575f046d15c0ab3c0d241557b2/pulpcore/tasking/tasks.py#L214-L243

bmbouter:
Cool, so I think you're saying we can know which process is executing the immediate task because:
1) pulp-api calls execute_task so it'll get the logging statement
2) the process name will be in the log statement. Is the process name in the log statement?

decko:
> Cool, so I think you're saying we can know which process is executing the immediate task because:
> 
>     1. pulp-api calls execute_task so it'll get the logging statement
> 
>     2. the process name will be in the log statement. Is the process name in the log statement?

1 is True. 2 maybe not. The process that execute the task is not in the log statement. Only the name of the task being executed. For us at Service is easy to know because we have more info attached to each log entry which is not present in the logs from the application itself.

We can say that the process that execute the task could be useful information to be on the log statement. I want to hear more from @mdellweg and @pedro-psb

pedro-psb:
> We can say that the process that execute the task could be useful information to be on the log statement.

I don't have anything to say against that.

Otherwise the PR looks good, that's the right place.

pulpcore/tasking/tasks.py
```diff
@@ -60,7 +60,20 @@ def _execute_task(task):
...
+                task.name,
+                domain.name,
+            )
+        else:
+            _logger.info(
+                "Starting task id: %s task_type: %s in domain: %s",
+                task.pk,
+                task.name,
+                domain.name,
+            )
```
mdellweg, line 76: What state exactly do you intend to capture when the first message is sent?

decko: Not the state. Only if it's an `immediate` task and the `task_type/task_name`. But there's not a second message here.

mdellweg: OK, then I would like us to not diverge the log messages. It really is the same event (a task was started) and the log messages should be structurally the same. However we should enrich the information at the end. And while we are at it, we should dump the `deferred` attribute alongside (it really is a 3-state disgused as two bools).

decko: Could you check it again @mdellweg?

mdellweg (approved):

PR #6492, opened 2025-04-22
[PULP-270] Validate and clean comments from Remote certs.

### LINKED ISSUE #6491: Remotes whose client- or CA-certs contain UTF-8 comments break sync attempts.
**Version**
main

**Describe the bug**
Providing a certificate with non-ASCII characters in comments will cause syncs attempting to use that certificate to fail with a traceback and the message "cadata should be an ASCII string or a bytes-like object"

See https://issues.redhat.com/browse/PULP-270 for more details and a reproducer.

**Additional context**
https://issues.redhat.com/browse/PULP-270

ggainey:
fixes #6491.

### DIFF
--- a/CHANGES/6491.bugfix
+++ b/CHANGES/6491.bugfix
@@ -1,0 +1,2 @@
+Client- and CA-certificates used in Remotes are now validated and any comments stripped
+before being applied to the Remote.
--- a/pulp_file/tests/functional/api/test_remote_settings.py
+++ b/pulp_file/tests/functional/api/test_remote_settings.py
@@ -1,10 +1,52 @@
-import uuid
-
 import pytest
+import uuid
 
 from pulpcore.client.pulp_file import (
     RepositorySyncURL,
 )
+from pulpcore.client.pulp_file.exceptions import BadRequestException
+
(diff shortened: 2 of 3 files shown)


mdellweg (approved):
I don't think we need a data repair command, because it's really not expected that a users system is running with bad certs for long, and it's easy to fix for the user. A data repair command could only delete or flag broken certs.

ggainey:
> I don't think we need a data repair command, because it's really not expected that a users system is running with bad certs for long, and it's easy to fix for the user. A data repair command could only delete or flag broken certs.

Exactly so. One option evaluated was to "clean" certs at time-of-use. That would also "auto-fix" certs with the current issue - at the expense of adding overhead to *every single connection attempt*. The reasoning of "it's easy to fix the problem in existing data" similarly pushed towards the current choice of "fix on set" being sufficient.

ggainey:
Note: 3.39 and 3.22 are old enough to not have PyOpenSSL as a direct dependency, and this fix would need to be re-written to use `cryptography` directly. As a result, we won't try to get this fix that far back, unless/until there is a direct request for it

PR #6526, opened 2025-04-29
Make TASK_DIAGNOSTICS more configurable, and add a memray diagnostic

dralley:


dralley:
I can split this up into more PRs or commits if requested.  It is probably also worth filing an issue(s).

dralley:
This is what `memray tree <dump_file>` looks like.  It's interactive and pretty useful.

Screenshot From 2025-04-29 18-55-58

dralley:
Just for informational purposes it's also possible to live-attach memray to a running process just like py-spy

https://bloomberg.github.io/memray/live.html

dralley:
@bmbouter Should we drop the existing memory tracker, or keep it?

bmbouter:
> @bmbouter Should we drop the existing memory tracker, or keep it?

I think I'm reading that you're adding a new memory tracker which could seems way better. If that's the case, then I'm +1 to remove the old one.

dralley:
@mdellweg is on PTO tomorrow

pulpcore/app/settings.py
```diff
@@ -382,7 +382,19 @@
...
+# * "pyinstrument"
+#    Dumps an HTML profile report produced by pyinstrument, showing time spent in various
+#    callstacks. This adds ~10% overhead to the task process and consumes extra memory.
+#    Tweaking code might be warranted for some advanced settings.
+# * "memray" - Dumps a report produced by memray which logs how much memory was allocated by which
+#    lines and functions, at the time of peak RSS of the task process. This adds significant
+#    runtime overhead to the task process, 20-40%. Tweaking code might be warranted for
+#    some advanced settings.
+# NOTE: "memray" and "pyinstrument" require additional packages to be installed on the system.
+TASK_DIAGNOSTICS = []  # ["memory", "pyinstrument", "memray"]
```
decko, line 397: What about adding a config validator to such cases where the TASK_DIAGNOSTICS is still set to `False` instead of an empty list?

dralley: Do you mean rejecting the value false, or acting as though the value false were equivalent to empty list?

I actually thought I was handling that but looking again I'm definitely not

dralley: for the sake of people with old configs, I'm gonna go for the latter

dralley: Actually never mind it already is handling that circumstance, just in a different place than I thought.

PR #6544, opened 2025-05-05
Add uvloop usage as an optional feature on the content-app

pedro-psb:
The uvloop replaces the asyncio builtin event loop. To enable it, one must set the UVLOOP_ENABLED=True and have the uvloop package installed in the system where the content-app is deployed.

Closes: #6021

### DIFF
--- a/CHANGES/6021.feature
+++ b/CHANGES/6021.feature
@@ -1,0 +1,2 @@
+Added the optional feature to enable uvloop library as a replacement of asyncio eventloop.
+The feature is activated through the UVLOOP_ENABLED feature flag settings and the 'uvloop' optional dependency.
--- a/pulpcore/content/__init__.py
+++ b/pulpcore/content/__init__.py
@@ -1,6 +1,7 @@
 import asyncio
 from contextlib import suppress
 from importlib import import_module
+from importlib.util import find_spec
 import logging
 import os
 
@@ -34,6 +35,13 @@ if settings.OTEL_ENABLED:
     app = web.Application(middlewares=[guid, authenticate, instrumentation()])
 else:
     app = web.Application(middlewares=[guid, authenticate])
+
+
+if settings.UVLOOP_ENABLED:
+    if not find_spec("uvloop"):
+        raise RuntimeError("The library 'uvloop' must be installed if UVLOOP_ENABLED is true.")
+    log.info("Using uvloop as the asyncio event loop.")
+
 
 CONTENT_MODULE_NAME = "content"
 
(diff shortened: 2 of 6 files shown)


mdellweg (approved):

CHANGES/6021.feature
```diff
@@ -0,0 +1,2 @@
+Added the optional feature to enable uvloop library as a replacement of asyncio eventloop.
+The fature is activated through the UVLOOP_ENABLED feature flag settings and the 'uvloop' optional dependency.
```
jobselko, line 2: typo in `fature`

mdellweg: Nice find.
Also just realized we should have a couple of lines in docs/admin/reference/settings or so.

dralley (approved):

pulpcore/content/__init__.py
```diff
@@ -35,6 +36,13 @@
 else:
     app = web.Application(middlewares=[guid, authenticate])
 
+
+if settings.UVLOOP_ENABLED:
+    if not find_spec("uvloop"):
+        raise RuntimeError("The library 'uvloop' must be installed if UVLOOP_ENABLED is true.")
+    log.info("Using uvloop as the asyncio event loop.")
```
mdellweg, line 43: Is this trigger even reached before entrypoint.py fails to load "aiohttp.GunicornUVLoopWebWorker"?

pedro-psb: Yes, I've tested that.
If I enable it without having uvloop installed and restart the workers I get:
```
[pulp]  | Traceback (most recent call last):
[pulp]  |   File "/usr/local/bin/pulpcore-content", line 5, in <module>
[pulp]  |     from pulpcore.content.entrypoint import main
[pulp]  |   File "/src/pulpcore/pulpcore/content/__init__.py", line 42, in <module>
[pulp]  |     raise RuntimeError("The library 'uvloop' must be installed if UVLOOP_ENABLED is true.")
[pulp]  | RuntimeError: The library 'uvloop' must be installed if UVLOOP_ENABLED is true.

```

mdellweg (approved):
I think, everything has been addressed.

pedro-psb:
@dralley @mdellweg @jobselko
Do you think this could be backported? Its a feature, but its optional and IHO a really lightweight one (in terms of added complexity).
Katello asked about the possibility of backporting that.

dralley:
We can, if Katello decided it would be valuable to them.  What are the real world practical performance implications?

PR #6545, opened 2025-05-05
Teach tasks to clean up their ProgressReports states on fail/cancel.

### LINKED ISSUE #3609: Task table and progress report table appear to occasionally get out of sync
**Version**

At least 3.18, probably others

**Describe the bug**

Through various cases and environments now I've noticed that often the core_task and core_progressreport table can get out of sync with one another.

For example;

```
[root@satellite ~]# echo "select pulp_id, name, state, worker_id from core_task where state = 'running';" | su - postgres -c "psql pulpcore"
 pulp_id | name | state | worker_id 
---------+------+-------+-----------
(0 rows)

[root@satellite ~]# echo "select pulp_id, pulp_created, pulp_last_updated, message, total, done, state from core_progressreport where state in ('running');" | sudo -u pulp PULP_SETTINGS='/etc/pulp/settings.py' pulpcore-manager dbshell | wc
     47     675   10093
```

It would be good to have some process of cleaning the progressreport table if it doesn't already exist. Perhaps a process to reference core_task and remove entries in progressreport if the pulp_ids no longer exist, or correct the state if the state of the ID doesn't match between the tables.

**To Reproduce**

**Expected behavior**

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=2173718

ggainey:
Added a manager-command to clean up current data in inconsistent states.

fixes #3609.

### DIFF
--- a/pulpcore/app/models/task.py
+++ b/pulpcore/app/models/task.py
@@ -204,6 +204,10 @@ class Task(BaseModel, AutoAddObjPermsMixin):
                 )
             )
 
+    def _cleanup_progress_reports(self, state):
+        """Find any running progress-reports and set their states to the specified end-state."""
+        self.progress_reports.filter(state=TASK_STATES.RUNNING).update(state=state)
+
     def set_completed(self):
         """
         Set this Task to the completed state, save it, and log output in warning cases.
@@ -232,6 +236,7 @@ class Task(BaseModel, AutoAddObjPermsMixin):
                         self.pk, self.state
                     )
                 )
+        self._cleanup_progress_reports(TASK_STATES.COMPLETED)
 
     def set_failed(self, exc, tb):
         """
@@ -264,6 +269,7 @@ class Task(BaseModel, AutoAddObjPermsMixin):
                     self.pk, self.state
                 )
             )
+        self._cleanup_progress_reports(TASK_STATES.FAILED)
 
     def set_canceling(self):
         """
@@ -317,6 +323,7 @@ class Task(BaseModel, AutoAddObjPermsMixin):
                     self.pk, self.state
                 )
             )
+        self._cleanup_progress_reports(final_state)
 
     def unblock(self):
         # This should be safe to be called without holding the lock.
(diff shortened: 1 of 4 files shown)


pulpcore/app/models/task.py
```diff
@@ -204,6 +204,12 @@ def set_running(self):
                 )
             )
 
+    def _cleanup_progress_reports(self, state):
+        """Find any running progress-reports and set their states to the specified end-state."""
+        from pulpcore.app.models import ProgressReport  # Avoid circular import
+
+        ProgressReport.objects.filter(task=self, state=TASK_STATES.RUNNING).update(state=state)
```
mdellweg, line 211: isn't there a `self.progress_reports` or self.progress_report_set` back association?

pulpcore/tests/unit/models/test_task.py
```diff
@@ -0,0 +1,45 @@
+import pytest
+import sys
+from pulpcore.app.models import Task, ProgressReport
+from pulpcore.constants import TASK_STATES
+
+
+def _test_state_changed(to_state, use_canceled=False):
```
mdellweg, line 7:
```suggestion
@pytest.mark.parametrize(
    "to_state,use_canceled",
    [
        (TASK_STATES.FAILED, False),
        (TASK_STATES.CANCELED, False),
        (TASK_STATES.CANCELED, True),
    ],
)
def test_state_changed(to_state, use_canceled):
```

ggainey: :+1:

pulpcore/tests/unit/models/test_task.py
```diff
@@ -0,0 +1,46 @@
...
+            try:
+                raise ValueError("test")
+            except ValueError:
+                exc_type, exc, tb = sys.exc_info()
+                task.set_failed(exc, tb)
+    elif TASK_STATES.CANCELED == to_state:
+        task.set_canceling()
+        task.set_canceled()
+
+    task.save()
```
mdellweg, line 39: All the "set_<state>" methods should go to the db immediately. Can you try running this task without this line?

mdellweg (approved):

Issue #6540, opened 2025-05-02
Response is not cached when package is downloaded using browser


dkliban:
**Version**
3.76

**Describe the bug**
Redis cache is not populated when a package is requested using a browser (or mvn client).

**To Reproduce**
Create a repository and a distribution.
Upload a package to the repository.
Publish the repository.
Go to the repository in the browser and download the package.
Use `curl -I ` to download the package in the terminal.
Observe that `X-PULP-CACHE: MISS`

**Expected behavior**
`X-PULP-CACHE: HIT`

**Additional context**
I initially discovered this issue when using `mvn install` to download packages from Pulp. No matter how many times I downloaded the same packages using `mvn`, the responses were never cached.

PR #6551, opened 2025-05-07
Cache StreamResponses in content app

gerrod3:
fixes: #6540

Some notes on the issue (and maybe this is really a new feature instead of a bug fix):

1. The key that is used to create the cache entry can be based on a couple things, by default for the content app it is based on the request's path & method. So the behavior in the issue is expected because `curl -I` performs a `HEAD` request while downloading in the browser is a `GET` request. Thus two different keys and two different cache entries.
2. On-demand/pull-through content was never cached before hand because I didn't think it was possible when I implemented this years ago. Upon second inspection it totally is possible and that is what the code changes do. So now a quick second request of originally on-demand content (as is common for the maven plugin) will result in a cache hit.
3. When the streamed content isn't saved, e.g. remote is set to stream or the pull-through content is metadata, then this response will not be cached. While possible that we could try to save the bytes streamed into the cache I would be reluctant to add this support for something that I think is one, not common (who uses stream policy?), and two, is probably an indicator that you should be syncing & publishing your content if upstream metadata cache misses are an issue.


pulpcore/content/handler.py
```diff
@@ -1323,9 +1334,14 @@ async def finalize():
             content_artifacts = await asyncio.shield(
                 sync_to_async(self._save_artifact)(download_result, remote_artifact, request)
             )
+            ca = content_artifacts[remote_artifact.content_artifact.relative_path]
+            # If cache is enabled, add the future response to our stream response
+            if settings.CACHE_ENABLED:
+                response.future_response = self._build_response_from_content_artifact(
+                    ca, original_headers, request
+                )
```
mdellweg, line 1342: So this "adding to the cache" only happens once the downloader in the stream response finished. Right?
(And actually after saving the artifact.)

mdellweg (approved):

dkliban (approved):
I tested this with Maven and I no longer see cache MISSes after the first request. Thank you @gerrod3!

bmbouter:
@gerrod3 I'm glad this is merged!

Followup question: why not remove the HTTP method type from the key? I think all use cases are about speeding up the content app which is all about pulling content. In that sense I can't imagine a use case difference between HEAD vs GET (I don't think the content app handles POST, DELETE, PUT) at all. I suspect I do not understand the use cases and there is a good reason, but I wanted to ask.

gerrod3:
> @gerrod3 I'm glad this is merged!
> 
> Followup question: why not remove the HTTP method type from the key? I think all use cases are about speeding up the content app which is all about pulling content. In that sense I can't imagine a use case difference between HEAD vs GET (I don't think the content app handles POST, DELETE, PUT) at all. I suspect I do not understand the use cases and there is a good reason, but I wanted to ask.

The original idea was that there could be a difference between the different methods and thus they should be cached separately. Now we only support GET and we don't make a difference between GET and HEAD, so maybe the default for the content-app isn't correct. In pulp-container we use the method since the V2 api allows for HEAD requests, but we also use a custom cache class so we could change pulpcore's without any problem.

PR #6556, opened 2025-05-08
[PULP-423] Fix wrong downloader session close on on-demand streaming 

pedro-psb:
**First commit:**
In the content app on-demand streaming error handler for digest errors, the TCP connection with the client must be closed. Possibly because of some confusion about what connection should be closed, the downloader
session was also closed.

Closing the session is unnecessary to the error handling and it's incompatible with some downloaders. E.g, `RpmFileDownloader` doesn't have a session at all. This causes unintented errors when using file:// remotes.

Also, there is no need to raise a runtime error from the error handling perspective, so a clean error log was used.

**Second commit:**
improves documentation about `REMOTE_CONTENT_FETCH_FAILURE_COOLDOWN`.

### DIFF
--- a/docs/user/learn/on-demand-downloading.md
+++ b/docs/user/learn/on-demand-downloading.md
@@ -73,21 +73,26 @@ On-demand and streamed content can be useful, but they come with some problems.
 
 There are two different types of errors that can occur with on-demand streaming:
 
-1. Pre-response: For some reason, Pulp can't get any data from the server (e.g, connectivity errors). A response is never started.
-2. Post-response: Pulp can get data from the remote and start streaming the response, but in the end the data doesn't match the expected digest.
+1. **Pre-response**: Pulp can't find or connect to the server. A response is never started.
+2. **Post-response**: Pulp can get data from the remote and start streaming the response, but the final digest is wrong.
 
-In the first case, Pulp will try all the available remote sources for the requested content and will return a 404 if all of them fail *with this same type of error*.
+(1) Pulp will try all the available remote sources for the requested content and will return a 404
+if all of them fail with pre-response retriable errors.
 
-In the second case, Pulp already sent the corrupted data to the client and can't recover from it, so it will close the connection to prevent the client from consolidating the file.
-When this happens, the content-app will ignore that remote source for a certain amount of time, which will enable future requests to select a different remote source.
-If all remote sources are ignored due to prior failure, then a 404 will be returned for all requests of that content until the cooldown period for one of those sources has expired.
+(2) Pulp already sent the corrupted data to the client and can't recover from it, so it will close the connection to prevent the client from consolidating the file.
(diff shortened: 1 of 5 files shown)


pulpcore/tests/functional/api/using_plugin/test_content_delivery.py
```diff
@@ -231,8 +232,7 @@ def download_from_distribution(content, distribution):
     acs_manifest_path = write_3_iso_file_fixture_data_factory("acs", seed=123)
     remote = create_simple_remote(basic_manifest_path)
     distribution = sync_publish_and_distribute(remote)
-    acs = create_acs_remote(acs_manifest_path)
-    refresh_acs(acs)
```
pedro-psb, line 235: Refresh is already called on `create_acs_remote`.

docs/user/learn/on-demand-downloading.md
```diff
@@ -73,21 +73,26 @@ On-demand and streamed content can be useful, but they come with some problems.
...
-The second case is complex and can be confusing to the user.
-The core reason for this complexity lies in the very nature of on-demand serving, which imposes that Pulp must fetch and stream the content on request time, and has no way to know anything about the remote before that.
-This constraint great limits the range of actions Pulp can do to properly satisfy the request.
+!!! info
 
-If this behavior is prohibitive, consider using the immediate sync policy.
+    Case (2) is complex and can be confusing to the user.
+
+    The core reason for this complexity lies in the very nature of on-demand serving, which imposes that Pulp must fetch and stream the content on request time, and has no way to know anything about the remote before that.
+    This constraint great limits the range of actions Pulp can do to properly satisfy the request.
```
dralley, line 93: "constraint great limits" feels like two phrasings smashed together

ggainey (approved):

ggainey:
@dralley @jobselko just a process note here - if you're asking for changes, mark your review as "changes requested"! Makes it *much* easier for the author (and the rest of us) know they have work to do.

jobselko (approved):

pulpcore/content/handler.py
```diff
@@ -1327,8 +1326,10 @@ async def finalize():
                 "If the Remote is known to be fixed, try resyncing the associated repository.\n"
                 "If the Remote is known to be permanently corrupted, try removing "
                 "affected Pulp Remote, adding a good one and resyncing.\n"
-                "If the problem persists, please contact the Pulp team."
+                "Learn more on <https://pulpproject.org/pulpcore/docs/user/learn/"
+                "on-demand-downloading/#on-demand-and-streamed-limitations>"
             )
+            return response
```
bmbouter, line 1332: I found this commit while investigating a bug report from a user. Apologies to come in post-merge with new questions.

What's the idea with returning a response here? This is handled in cases where a `DigestValidationError` is experienced and in that case (to me) it seems we should not pass those bits to the client.

PR #6568, opened 2025-05-12
Resolve warnings of testing library

emmanuel-ferdman:
# PR Summary
This small PR migrates from `unittest.assertEquals` to `unittest.assertEqual` which is deprecated from Python2.7:
```python
DeprecationWarning: Please use assertEqual instead.
```


mdellweg:
Since we are using pytest anyway, would you mind to change them to a plain assert right away?

`assert a == b` reads so much easier.

mdellweg (approved):

PR #6588, opened 2025-05-19
Reduce db load on tasks state machine

mdellweg:
Instead of unsetting state attributes, in the update, reflect them as is on
successful update.

### DIFF
--- a/pulpcore/app/models/task.py
+++ b/pulpcore/app/models/task.py
@@ -76,6 +76,21 @@ class Task(BaseModel, AutoAddObjPermsMixin):
     """
     Represents a task
 
+    The Tasks state machine works like a finite automaton without loops:
+    The initial state is WAITING.
+    Final states are COMPLETED, FAILED and CANCELED.
+    The possible transitions are:
+        WAITING -> RUNNING
+        WAITING ->* CANCELING
+        RUNNING -> COMPLETED
+        RUNNING -> FAILED
+        RUNNING ->* CANCELING
+        CANCELING -> CANCELED
+
+    The transitions to CANCELING (marked with *) are the only ones allowed to happen without
+    holding the tasks advisory lock. Canceling is meant to be initiated asyncronously by a sparate
+    process before signalling the worker via Postgres LISTEN.
+
     Fields:
 
         state (models.TextField): The state of the task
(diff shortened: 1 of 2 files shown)


pulpcore/app/models/task.py
```diff
@@ -371,24 +335,10 @@ def unblock(self):
...
-        if fields is not None:
-            fields = set(fields)
-            deferred_fields = {
-                field for field in self.get_deferred_fields() if not field.startswith("enc_")
-            }
-            # If any state related deferred field is going to be loaded
-            if fields.intersection(deferred_fields):
-                # then load all of them
-                fields = fields.union(deferred_fields)
-        super().refresh_from_db(using, fields, **kwargs)
```
pedro-psb, line 348: Why remove this override? Isn't the exclusion of encrypted fields important? I really don't know.
Assuming that those fields are immutable (enc_args,...), it really doesn't make sense to refresh them.

pedro-psb: Actually this happens in the case of a deferred access... So it meant that if enc fields are deferred via `only` or `defer` , their values are never loaded.

pedro-psb (commented):
About the first change:

Since these transitions requires a lock, no concurrent changes can occur to the task in the database. So it's safe to assume that if the update succeds, the only difference in data-state (the set of data in all fields) between the object and db
is that data from the update call and we can update the instance without asking the db.
That looks right to me.

But in the case of cancelling, we don't have that invariant about the db data-state.
So some other mutable field could have changed and the instance not know about it.

About the second change:

I lack the understanding on how some of this works to evalute it's impact, if any.
I wonder if this affects query count or query payload size.

E.g, not sure if that's the actual scenario:

```python
# Before (k queries)
del self.f0
...
del self.fk

# access pattern later;
self.f0  # triggers 1st query to update f0
...
self.fk  # triggers kth query to update fk

# After (1 query, bigger payload?)
self.refresh_from_db()

mdellweg:
It's not even about someone else changing the task or holding the lock. In our process we always hold a stale snapshot of the task and when we update the state successfully, we can assume that that state will be in the database at least for that moment. And that logic is true even for cancel.
If we failed however, we can assume the task is no longer what we think it is. Reloading sounds like a good idea. And to make that reasonably cheap, we defer the encrypted arguments by default. See TaskManager.
That second change simplifies the reload logic, that might have been scewed by the unblocked_at transition anyway.

mdellweg:
I think we should reconsider adding unblocked to the possible task states.

pedro-psb (approved):
That sounds reasonable, thanks for the clarification.

PR #6590, opened 2025-05-20
Deprecate TaskGroup.finish()

mdellweg:


CHANGES/plugin_api/task_group_finish.removal
pedro-psb, line 1: Can you add the explanation on why it's safe to remove it here?
Basically what you've shared on pulpcore meeting.

pedro-psb (changes requested):

pedro-psb (approved):
Thank you, that's helpful.

PR #6594, opened 2025-05-20
use IPv4 or IPv6 for api and content server depending on system capabilities

mayrstefan:
This adds a function inspired by urllib3.util.connection._has_ipv6 to detect if a system can bind IPv6 sockets and use this information to bind the default listening address to either an IPV4 or an IPv6 address.

This should address https://github.com/pulp/pulp-oci-images/issues/582

I've decided to add my helper function in netutil.py because when i put in in util.py a lot of unnecessary code is executed when importing pulpcore.app.util. Tell me if there is a better place or name for this has_ipv6 function.

With this helper function it will also be ease to fix https://github.com/pulp/pulp-oci-images/issues/595:
We can set a variable in https://github.com/pulp/pulp-oci-images/blob/latest/images/s6_assets/template_nginx.py and modify the jinja template to add the IPv6 listener when IPv6 is usable.

### DIFF
--- a/CHANGES/6606.bugfix
+++ b/CHANGES/6606.bugfix
@@ -1,0 +1,1 @@
+Changed default http bind address from `[::]` to `0.0.0.0` if IPv6 is not available.
--- a/pulpcore/app/entrypoint.py
+++ b/pulpcore/app/entrypoint.py
@@ -11,6 +11,7 @@ from django.db.utils import InterfaceError, DatabaseError
 from gunicorn.workers.sync import SyncWorker
 
 from pulpcore.app.apps import pulp_plugin_configs
+from pulpcore.app.netutil import has_ipv6
 from pulpcore.app.pulpcore_gunicorn_application import PulpcoreGunicornApplication
 
 logger = getLogger(__name__)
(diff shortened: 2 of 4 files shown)


mdellweg (commented):
I think we should add a changelog for this.

mayrstefan:
@mdellweg should i add an entry to CHANGES.md or was that last comment just a reminder for yourself?

mdellweg:
I meant this:
https://pulpproject.org/pulpcore/docs/dev/guides/git/?h=towncrier#changelog-update

mayrstefan:
Thank you for the pointer to the guide. I hope my PR now meets your expectations.

CHANGES/6606.bugfix
```diff
@@ -0,0 +1 @@
+Changed default bind address from [::] to 0.0.0.0 if IPv6 is not available
```
mdellweg, line 1: Since this becomes a snippet in CHANGES.md, we can use Markdown. ;)

```suggestion
Changed default http bind address from `[::]` to `0.0.0.0` if IPv6 is not available.
```

mayrstefan:
Running `black .` modified more files than expected. Maybe that could be added as a github action (https://black.readthedocs.io/en/stable/integrations/github_actions.html)

mdellweg (approved):

mayrstefan:
Next try. This time with a fix for E722 to pass the flake8 test

mdellweg (approved):

PR #6607, opened 2025-05-23
Add vscode to gitignore

fao89:


dkliban (approved):

mdellweg:
Sorry, but vscode is not project specific, but more specific to your setup. There's a way to setup ignore files global to your user account. They are perfectly suited for your editor of choice.
Rationale: We don't want to collect ignores for all the editors out there.

https://git-scm.com/docs/gitignore

PR #6612, opened 2025-05-26
[PULP-570] Mark a failed heartbeat as a catastrophic failure

mdellweg:
This is supposed to dectect situations where the workers fail to reach the database and therefore any type of network partitioning. By theory this should improve consistency in the face of partitioning. It sacrifices only the avaliablility of workers, that would not reach the database anyway to perform anything meaningful.

### DIFF
--- a/pulpcore/content/__init__.py
+++ b/pulpcore/content/__init__.py
@@ -7,7 +7,7 @@ import os
 
 from asgiref.sync import sync_to_async
 from aiohttp import web
-
+from gunicorn.arbiter import Arbiter
 import django
 
 
@@ -16,6 +16,7 @@ django.setup()
 
 from django.conf import settings  # noqa: E402: module level not at top of file
 from django.db.utils import (  # noqa: E402: module level not at top of file
+    IntegrityError,
     InterfaceError,
     DatabaseError,
 )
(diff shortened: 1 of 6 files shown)


pulpcore/tasking/worker.py
```diff
@@ -133,32 +138,25 @@ def _pg_notify_handler(self, notification):
...
+        try:
+            self.worker.save_heartbeat()
+            _logger.debug(msg)
+        except (IntegrityError, DatabaseError):
+            # WARNING: Do not attempt to recycle the connection here.
+            # The advisory locks are bound to the connection and we must not loose them.
+            _logger.error(f"Updating the heartbeat of worker {self.name} failed.")
+            # TODO if shutdown_requested, we may need to be more aggressive.
+            self.shutdown_requested = True
+            self.cancel_task = True
```
pedro-psb, line 173: The worker does have a handler for `shutdown_requested`. Why we need to do anything different here?

pulpcore/content/__init__.py
```diff
@@ -47,37 +48,35 @@
...
             except (InterfaceError, DatabaseError):
                 await sync_to_async(Handler._reset_db_connection)()
-                log.info(fail_msg)
-            await asyncio.sleep(heartbeat_interval)
+                try:
+                    await content_app_status.asave_heartbeat()
+                    log.debug(msg)
+                except (InterfaceError, DatabaseError):
+                    log.error(fail_msg)
+                    exit(3)
```
pedro-psb: After reading your observation on not trying to recycle the connection on the task worker, I would say that if the content-app dispatched non-deferred short tasks (as it did before) it could also hold a lock. However, it wouldnt share the same connection, as the dispatch code is threaded.
(I'm just commenting to put that thought out)

pulpcore/tasking/worker.py
```diff
@@ -68,7 +69,11 @@ def __init__(self):
         self.last_metric_heartbeat = timezone.now()
         self.versions = {app.label: app.version for app in pulp_plugin_configs()}
         self.cursor = connection.cursor()
-        self.worker = self.handle_worker_heartbeat()
+        try:
+            self.worker = Worker.objects.create(name=self.name, versions=self.versions)
+        except IntegrityError:
+            _logger.error(f"A worker with name {self.name} already exists in the database.")
+            exit(1)
```
pedro-psb, line 80: Why does this have a different exit code than the similar code in content and api? They have exit code 4.
And wdyt about extracting this logic to a super class method?

pedro-psb (changes requested):
Can you make the exit codes constants?
And are these following some pre-existing convention?

pulpcore/content/__init__.py
mdellweg: Indeed, you are touching on the other huge fear of mine.
I really want us to consider redefining "lock on task" without holding an advisory lock for hours.

pulpcore/tasking/worker.py
mdellweg: The gunicorn workers are different, as they are run by an arbiter, who restarts them, unless you issue the proper exit code.
Maybe we want to add a similar layer here? Maybe redefine the worker as a gunicorn process too? (One that does not listen to http.)

mdellweg:
@pedro-psb At this point this PR is more of a suggestion.
We would need to decide if we want to do any of this at all.
(It's not "draft" because I want it to be looked at...)

pulpcore/tasking/worker.py
pedro-psb: I see.
Maybe... Mixed feeling about it

pulpcore/tasking/worker.py
mdellweg: You are asking about why we may need to be more aggressive? Well the heart stopped for that worker, so we know something is wrong already. Do you think we can still rely on the shutdown and cancel_task logic to perform as designed?

pedro-psb: > Do you think we can still rely on the shutdown and cancel_task logic to perform as designed?

Yeah, if we got a DatabaseError we probably shouldn't rely on the worker to cleanup itself... In this case worker will probably exist non-gracefully anyway (with an exception) when trying to shutdown or perform cancellation cleanup.

pedro-psb (approved):
That looks good.
I think we are ready to merge this, since we've fixed the non-fatal missing heartbeats, right?

PR #6659, opened 2025-06-11
Adjust some dependecies seemingly following semver

mdellweg:


ggainey:
For things where we "trust" their semver, this will mean way fewer dependabot updates and "can you please backport the Z update so we can allow the product to pick up the new version" requests.

The only downside is, a library whose semver we trust that makes a non-backward-compatible change in a Z-stream, will cause our nightlies to break - and also anyone who just does a "pip install foo" and picks up the breaking-Z. This is Sad, but not frequent. And the way we'll respond is to lock/address the UB.

PR #6663, opened 2025-06-12
Update docs publish script to use new pulp-docs build commands

pedro-psb:
This is a final change to transition to the revamped pulp-docs effort: https://issues.redhat.com/browse/PULP-385
Without it, the docs publish will fail, since the rewrite feature branch is already merged into main and has some breaking changes to the API.


dralley (approved):

mdellweg (approved):
Thank you!

Issue #6647, opened 2025-06-05
HEAD request to ondemand file leaves tempfile


philfry:
**Version**
core 3.80.0
rpm 3.30.0

**Describe the bug**
When requesting a not yet retrieved file from an ondemand repository using `HEAD`, pulpcore fetches a chunk of the remote file into a tempfile. I suspect that if the original file is bigger than the retrieved chunk the tempfile stays in place and never gets deleted.

**To Reproduce**
1. add a remote with "ondemand" policy
2. add repository and distribution
3. sync using "mirror_complete"
4. select a kind of large file that hasn't been retrieved to the local storage yet
5. curl -I that file (HEAD)

**Expected behavior**
Either:
- pulp retrieves the whole remote file into a tempfile while streaming the headers to the client
- pulp adds the tempfile's content to the artifact storage
- pulp deletes the tempfile

Or:
- pulp retrieves the headers of the remote file and streams them to the client

**Additional context**
afaict requests for smaller files, like a few hundreds of kiB of size, leave no tempfile behind. For bigger files, a tempfile of something between slightly less than a MiB up to 5 MiB remains in pulp/tmp

PR #6667, opened 2025-06-13
Stop writing data to client for on-demand HEAD requests

gerrod3:
fixes: #6647

### DIFF
--- a/pulpcore/tests/functional/api/using_plugin/test_content_delivery.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_content_delivery.py
@@ -1,8 +1,10 @@
 """Tests related to content delivery."""
 
 import hashlib
+import requests
 import subprocess
 import uuid
+from random import sample
 from urllib.parse import urljoin
 
 import pytest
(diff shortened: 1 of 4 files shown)


pulpcore/tests/functional/api/using_plugin/test_content_delivery.py
```diff
@@ -244,3 +246,35 @@ def download_from_distribution(content, distribution):
...
+    content = file_bindings.ContentFilesApi.list(repository_version=repo.latest_version_href)
+    content_href_filenames = sample([(c.pulp_href, c.relative_path) for c in content.results], k=3)
+
+    distro = file_distribution_factory(repository=repo.pulp_href)
+    for _, content_filename in content_href_filenames:
+        url = urljoin(distribution_base_url(distro.base_url), content_filename)
+        response = requests.head(url)
+        assert response.status_code == 200
+    # Give some time for the file to be saved
+    for content_href, _ in content_href_filenames:
```
gerrod3, line 278: It's quite possible this test will have timing issues if the saving takes longer than the HEAD & API call times. Any suggestions?

pyproject.toml
```diff
@@ -29,7 +29,7 @@ requires-python = ">=3.9"
 dependencies = [
   "aiodns>=3.0,<=3.4.0",
   "aiofiles>=22.1,<24.2.0",
-  "aiohttp>=3.8.1,<3.13",  # semver https://docs.aiohttp.org/en/stable/faq.html#what-is-the-api-stability-and-deprecation-policy
+  "aiohttp>=3.8.3,<3.13",  # semver https://docs.aiohttp.org/en/stable/faq.html#what-is-the-api-stability-and-deprecation-policy
```
gerrod3: Not sure what bug fix happened, but the tests consistently failed with aiohttp==3.8.1, but bumping it up one version fixed it. 3.8.2 is yanked, so that's why 3.8.3 is the new lowerbound.

pulpcore/tests/functional/api/using_plugin/test_content_delivery.py
mdellweg: I thought we were not supposed to download the artifact on a head call.
Also yes, a test we know to fail occasionally is not a good addition.

gerrod3: We do because we call stream_remote_artifact even on HEAD calls and the downloader only knows how to make GET requests. Also, I've tried the test many times and have never gotten it to fail due to timing so maybe my comment is wrong. I think requests waits for the response to complete before returning (other clients will close the session once they get the headers).

PR #6693, opened 2025-06-17
Test multithreaded file hashing

dralley:
Note that since the hashing is done by C code inside of hashlib which releases the GIL, threading actually does work here despite the normal GIL rules.

### DIFF
--- a/pulpcore/app/files.py
+++ b/pulpcore/app/files.py
@@ -1,4 +1,6 @@
 import os
+import concurrent.futures
+from concurrent.futures import ThreadPoolExecutor, ALL_COMPLETED
 from gettext import gettext as _
 
 from django.core.files.uploadedfile import TemporaryUploadedFile
@@ -39,9 +41,10 @@ class PulpTemporaryUploadedFile(TemporaryUploadedFile):
         instance = cls(name, "", size, "", "")
         instance.file = file
         # Default 1MB
-        while data := file.read(1048576):
-            for hasher in models.Artifact.DIGEST_FIELDS:
-                instance.hashers[hasher].update(data)
+        with ThreadPoolExecutor(max_workers=6) as executor:
+            while data := file.read(1048576):
+                for hasher in models.Artifact.DIGEST_FIELDS:
+                    executor.submit(instance.hashers[hasher].update, data)
 
         # calling the method read() moves the file's pointer to the end of the file object,
         # thus, it is necessary to reset the file's pointer position back to 0 in case of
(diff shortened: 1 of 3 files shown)


dkliban:
I used the pyinstrument task profiler when uploading a 145MB RPM. Before this patch was applied the task took 1.17s. After the patch the run time was 0.99s. I am attaching an archive that contains the output of pyinstrument. 
[pyinstrument.zip](https://github.com/user-attachments/files/20819697/pyinstrument.zip)

bmbouter (dismissed):
I'm +1 on merging this.

pulpcore/app/files.py
```diff
@@ -39,9 +40,10 @@ def from_file(cls, file):
         instance = cls(name, "", size, "", "")
         instance.file = file
         # Default 1MB
-        while data := file.read(1048576):
-            for hasher in models.Artifact.DIGEST_FIELDS:
-                instance.hashers[hasher].update(data)
+        with ThreadPoolExecutor(max_workers=6) as executor:
+            while data := file.read(1048576):
+                for hasher in models.Artifact.DIGEST_FIELDS:
+                    executor.submit(instance.hashers[hasher].update, data)
```
dralley, line 47: Theoretically speaking I suppose it's possible to build up `data` in memory if we were submitting tasks faster than they could be completed, as the refcounts wouldn't go to zero on the `data` immediately.  I'm not sure that's too much of a problem in practice?  But I'm calling it out.

dralley: I could do here what I've done below, and add a `concurrent.futures.wait()` to wrap everything up before the loop resets.  That trades off with throughput though probably but it's not worse than what we're doing currently so :shrug:

pulpcore/download/base.py
```diff
@@ -34,6 +36,8 @@
         values are header content. None when not using the HttpDownloader or sublclass.
 """
 
+THREADPOOL = ThreadPoolExecutor()
```
dralley, line 39: Shared between all downloaders.  I figured that was better than trying to set up a separate threadpool for every downloader.

dralley:
Another 25% of the runtime could probably be shaved off by 

* eliminating the seemingly significant overhead caused by https://github.com/pulp/pulpcore/blob/main/pulpcore/app/serializers/base.py#L66-L80 and
* optimize  https://github.com/pulp/pulp_rpm/blob/main/pulp_rpm/app/shared_utils.py#L51-L53
    * download only the header somehow and
    * eliminate the extra file copy going on here and process the header from memory if it's small enough

bmbouter (approved):
Still looking good!

dkliban (approved):

PR #6699, opened 2025-06-18
[bp/3.39] Pin pycares to avoid bug with aiodns combination

pedro-psb:
See for more context:
https://github.com/pulp/pulpcore/pull/6695

(cherry picked from commit b1560306741816770ddafa3d51dadbef4c7b072a)

### DIFF
--- a/.ci/scripts/calc_constraints.py
+++ b/.ci/scripts/calc_constraints.py
@@ -53,6 +53,9 @@ def to_upper_bound(req):
         if requirement.name == "pulpcore":
             # An exception to allow for pulpcore deprecation policy.
             return fetch_pulpcore_upper_bound(requirement)
+        # skip requirement with environment scopes. E.g 'foo==1.0.0;python_version>=3.9'
+        if requirement.marker:
+            return f"# ENVIRONMENT IS UNTRACKABLE: {req}"
         for spec in requirement.specifier:
             if spec.operator == "~=":
                 return f"# NO BETTER CONSTRAINT: {req}"
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,6 +22,8 @@ importlib-metadata>=6.0.1,<=6.0.1  # Pinned to fix opentelemetry dependency solv
 jinja2>=3.1,<=3.1.6
 json_stream>=2.3.2,<2.4
 jq>=1.6.0,<1.7.0
+# pycares is only a transitive dependency, but there is a combination of versions that expresses a bug.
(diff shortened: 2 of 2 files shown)


.github/workflows/scripts/before_install.sh
```diff
@@ -60,11 +60,17 @@ then
   exit $s
 fi
 
+CI_IMAGE="ghcr.io/pulp/pulp-ci-centos"  # get dynamically in plugin-template
+CONTAINER_PYTHON_VERSION="$(docker run --rm $CI_IMAGE python3 --version | awk '{print $2}')"
```
pedro-psb: This could be done in github workflow level (`workflows/test.yml`) and passed as an environment variable.
Or maybe there is a better way of getting this?

requirements.txt
```diff
@@ -22,6 +22,10 @@ importlib-metadata>=6.0.1,<=6.0.1  # Pinned to fix opentelemetry dependency solv
 jinja2>=3.1,<=3.1.6
 json_stream>=2.3.2,<2.4
 jq>=1.6.0,<1.7.0
+# pycares is only a transitive dependency, but there is a combination of versions that expresses a bug:
+# pycares==4.9 + aiodns==3.2.0 + python<3.12
+pycares<4.9;python_version>='3.9'
+pycares<4.5;python_version<'3.9'
```
mdellweg: Shouldn't this still be?
```suggestion
pycares<4.9;python_version<'3.12'
pycares<4.5;python_version<'3.9'
```

pedro-psb: To make it very explicit (and easy) to the calc_constraints script I'll do  `>=3.9,<3.12`.
Otherwise it would pick both with the current implementation.
But why again we impose that upper python version?

mdellweg: Because starting from python 3.12, we do not want to impose any restriction on the 2nd hand dependency at all. (That was the original request by the build team.)

.ci/scripts/calc_constraints.py
```diff
@@ -53,6 +56,13 @@ def to_upper_bound(req):
         if requirement.name == "pulpcore":
             # An exception to allow for pulpcore deprecation policy.
             return fetch_pulpcore_upper_bound(requirement)
+        # check python scope. E.g 'foo==1.0.0;python_version>=3.9'
```
mdellweg: Actually I would be OK with
```python
if requirement.marker:
    return "# Skipping because environmental conditions are untrackable."
```

Issue #6713, opened 2025-06-23
pulp_labels does not properly serialize inside a ViewSet


dkliban:
**Version**
3.80

During file uploads, when any serializer that is derived from `NoArtifactContentSerializer` is validated inside a ViewSet (instead of a task), the data from `pulp_labels` is lost. This happens because of how DRF treats DictFields for multipart/form uploads.

PR #6714, opened 2025-06-23
Replaces use of HStoreField with a custom PulpLabelsField

dkliban:
fixes: #6713


bmbouter (approved):
This looks great, and with so many tests I feel good about it. Also no migration!

gerrod3:
Is this the right change? I really don't understand it

gerrod3 (approved):
@dkliban Cool, thanks for the explanation.

mdellweg:
> This looks great, and with so many tests I feel good about it. Also no migration!

I don't see any test. Specifically I would like to see one demonstrating the previously wrong behaviour now being fixed.

Issue #6630, opened 2025-05-29
Incorrect backend reference in the public documentation


MateusPatz:
I was setting up the solution in a personal environment, and using the value referenced in the public documentation I couldn’t successfully configure an S3-like backend (MinIO).

https://github.com/pulp/pulpcore/blob/68a8861ea7c46ef40d58b88bb7c0d09427fb5ac7/docs/admin/guides/configure-pulp/configure-storages.md?plain=1#L66

The following error was raised:

```
migration_service-1          | dynaconf.validator.ValidationError: 'REDIRECT_TO_OBJECT_STORAGE=True' is only supported with the local file, S3, GCP or Azure storage backend configured in STORAGES['default']['BACKEND'] (deprecated DEFAULT_FILE_STORAGE).
```

However, after changing it to
```python
    "BACKEND": "storages.backends.s3boto3.S3Boto3Storage",
```
—as referenced in other parts of the documentation—everything worked correctly.

pedro-psb:
About the boto3, this is actually a bug. The validator shouldn't complain about this.
The S3Boto3Storage is a shim layer and it's recommended to use the one described in the docs. See:
https://github.com/jschneier/django-storages/blob/9de3cc9da4dbd67fcc56ecafc7cbf738af90136c/CHANGELOG.rst#s3-5

The static files is probably really just missing there.

PR #6715, opened 2025-06-24
Fix S3 Storage Documentation, Validator and Serializer

DenisBY:
Fixes: #6630

### DIFF
--- a/CHANGES/6630.bugfix
+++ b/CHANGES/6630.bugfix
@@ -1,0 +1,1 @@
+Fixed S3 Storage configuration.
--- a/docs/admin/guides/configure-pulp/configure-storages.md
+++ b/docs/admin/guides/configure-pulp/configure-storages.md
@@ -73,6 +73,9 @@ STORAGES = {
             "region_name": "eu-central-1",
         },
     },
+    "staticfiles": {
+        "BACKEND": "django.contrib.staticfiles.storage.StaticFilesStorage",
+    },
 }
 ```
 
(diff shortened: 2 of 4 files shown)


jobselko:
That line needs to be included directly in the commit message body, not just in the PR description (github-actions bot expects it as well).
Sorry for not mentioning it earlier, but could you please also add a `.bugfix` changelog entry? You can find more details in https://pulpproject.org/pulpcore/docs/dev/guides/git/#changelog-update.

jobselko:
> @jobselko is it better now?

Yes! Just one suggestion - in the commit message, the title and the body (in your case only `fixes: #6630`) should be separated by a single blank line. You can see an example in https://pulpproject.org/pulpcore/docs/dev/guides/git/#commit-message.

CHANGES/6630.bugix
```diff
@@ -0,0 +1 @@
+Fix S3 Storage configuration
```
jobselko: This should be written as a sentence in the past tense, ending with a point: `Fixed S3 Storage configuration.`.
Also, the file name contains a typo - `bugix` instead of `bugfix`.

mdellweg: As this is a doc change, you should call the file `6630.doc` .

jobselko: The changes were in `serializers/domain.py` and `settings.py` as well, so I would vote for `bugfix` here.

mdellweg: Right, I was misinterpreting the PR title. Sorry.

DenisBY: Both are fixed

jobselko: Thank you! It seems that you overlooked my note re. missing single blank line in the commit. Could you please address this as well?

DenisBY: Ah, sorry. Is it better now?

PR #6718, opened 2025-06-24
Be more conservative about hashing concurrency

dralley:
Just in case - let's be absolutely sure data gets hashed in the correct order. Better to be more conservative than to mess this up.

### DIFF
--- a/pulpcore/app/files.py
+++ b/pulpcore/app/files.py
@@ -43,8 +43,11 @@ class PulpTemporaryUploadedFile(TemporaryUploadedFile):
         # Default 1MB
         with ThreadPoolExecutor(max_workers=6) as executor:
             while data := file.read(1048576):
-                for hasher in models.Artifact.DIGEST_FIELDS:
+                futures = [
                     executor.submit(instance.hashers[hasher].update, data)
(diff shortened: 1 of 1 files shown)


pulpcore/app/files.py
```diff
@@ -43,8 +43,11 @@ def from_file(cls, file):
         # Default 1MB
         with ThreadPoolExecutor(max_workers=6) as executor:
             while data := file.read(1048576):
-                for hasher in models.Artifact.DIGEST_FIELDS:
+                futures = [
                     executor.submit(instance.hashers[hasher].update, data)
+                    for hasher in models.Artifact.DIGEST_FIELDS:
+                ]
+                concurrent.futures.wait(futures, timeout=None, return_when=ALL_COMPLETED)
```
dralley, line 50: I mentioned this in a comment on the last PR, and at the time I thought the approach on the left was probably OK, but on second thought I think it's best to use the more conservative approach (this one) in all places, rather than risk `ThreadPoolExecutor` hashing blocks out of order.

bmbouter (approved):
:+1:

bmbouter (approved):

PR #6729, opened 2025-06-26
Update Protect Content documentation

### LINKED ISSUE #6731: Update Protect Content documentation regarding `--guard` command-line option
The current documentation doesn't reflect the new format of `--guard` command-line option. It should be fixed.

DenisBY:
Fixed #6731

### DIFF
--- a/CHANGES/6731.doc
+++ b/CHANGES/6731.doc
@@ -1,0 +1,1 @@
+Updated "Protect Content" documentation regarding the `--guard` command-line option.
--- a/docs/user/guides/protect-content.md
+++ b/docs/user/guides/protect-content.md
@@ -21,19 +21,19 @@ To set up an RBAC content guard:
 1. Create the content guard:
 
     ```bash
-    pulp content-guard rbac create --name rbac
+    pulp content-guard rbac create --name rbac-guard
     ```
 
 2. Assign permissions to users and/or groups:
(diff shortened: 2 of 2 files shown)


docs/user/guides/protect-content.md
```diff
@@ -67,7 +67,7 @@ The composite content guard combines multiple guards using OR logic - if any of
 
 ```bash
 # Use different types of content guards, e.g. RBAC and X509
-pulp content-guard composite create --name composite --guard rbac --guard x509
+pulp content-guard composite create --name composite-guard --guard rbac --guard x509
```
mdellweg: I'm almost certain, the `--guard` options need the colon separated qualifiers too.

DenisMkS: like this?

CHANGES/6731.doc
```diff
@@ -0,0 +1 @@
+Updated Protect Content documentation regarding --guard command-line option
```
mdellweg:
```suggestion
Updated "Protect Content" documentation regarding the `--guard` command-line option.
```

DenisBY: thanks, applied

mdellweg (changes requested):
Can you please squash the commits?

PR #6753, opened 2025-07-02
Add safeguards to the task state transitions

mdellweg:
This adds some assertions that transitions are only performed when the advisory lock of the task is held.


mdellweg:
It does not work. If the subtask is performing the transitions the check fails.

PR #6767, opened 2025-07-08
Improve content set calculation

gerrod3:
fixes: #5738


bmbouter (dismissed):
This all looks good to me. This is only to improve performance so all existing tests passing is good. We're ready to evaluate this on services production (where we greatly need the speedup), but we can only do that after it's merged and released to pypi.

dkliban (dismissed):

pulpcore/app/models/repository.py
```diff
@@ -822,12 +823,14 @@ class RepositoryVersion(BaseModel):
     complete = models.BooleanField(db_index=True, default=False)
     base_version = models.ForeignKey("RepositoryVersion", null=True, on_delete=models.SET_NULL)
     info = models.JSONField(default=dict)
+    content_ids = ArrayField(models.UUIDField(), default=None, null=True)
```
mdellweg, line 826: ~Please make this a foreign key array. PostgreSQL allows that and it will help us to protect data consistency.~

~If Django does not, we need to have a discussion.~

edit: I'm taking this one back. There has been work towards this but it never got merged. And the internet is not a reliable source to make such a claim.

pulpcore/app/models/repository.py
```diff
@@ -822,12 +823,14 @@ class RepositoryVersion(BaseModel):
...
 
     class Meta:
         default_related_name = "versions"
         unique_together = ("repository", "number")
         get_latest_by = "number"
         ordering = ("number",)
+        indexes = [GinIndex(fields=["content_ids"])]
```
mdellweg: Just curious. Do we have any evidence that this index is helping?
I can imagine how it should help, but do we _know_ it?

mdellweg (changes requested):
This change is supposedly duplicating the data currently held in the largest join table of every Pulp installation used within the range of it's design ("content in repositories").
Duplicating data is prone lead to inconsistencies. And you already spotted a particular case where exactly this _will_ happen.
So my big question is: What is the overarching plan with this? Are we removing the RepositoryContent table?

mdellweg:
Random thought: How about `exclude` ..ing this humongous field from most queries that don't need it?

dralley:
My only thought is that IIRC the reason we didn't go with this approach in the first place was that the sheer # of content IDs for large repositories x thousands of versions might become a problem, and that's how the version delta approach was chosen

If we've come to the conclusion that actually it's not that big a problem and we ought to store a list of content-per-version anyway, then shouldn't we consider (at some point in the future) whether we'd be better off using that approach exclusively?

bmbouter:
> I disagree, this migration looks like it can easily be reverted. So you can try this out (for a limited time) on an existing installation. (I will not make a statement about whether that is a good idea to do "in production".)

We talked this over on the Matrix thread some, but I'll restate here also. We can only effectively test this in production because that's the only place with a large enough DB to where paging in data causes really significant BufferIOWait. I'm not comfortable applying a migration to production that isn't released. Not to say I'm 100% no on it, but that's my comfort level currently.

pulpcore/app/models/repository.py
mdellweg: Maybe we could build something with on delete triggers on the content side.

bmbouter:
> My only thought is that IIRC the reason we didn't go with this approach in the first place was that the sheer # of content IDs for large repositories x thousands of versions might become a problem, and that's how the version delta approach was chosen

My memory of this is that the data model evaluated back then was a single row per content membership per repo version, and the row count on the table was untenable. We didn't use this ArrayField back then IIRC, because the psql versions we started with didn't have it (or maybe Django didn't support it or something). Anyways, it wasn't an option is what I remember. With this design the row count should be equal to the repo version count which is reasonable (to me).

> 
> If we've come to the conclusion that actually it's not that big a problem and we ought to store a list of content-per-version anyway, then shouldn't we consider (at some point in the future) whether we'd be better off using that approach exclusively?

+1 to evaluating at some point in the future if we'd be better off with this approach exclusively.

mdellweg:
> Relying on the ArrayField exclusively is perhaps not reasonable when it comes to orphan detection. But IDK.

Postgres is able to query across array fields. I believe we can make that work. Also a bit of performance hit on orphan cleanup is probably forgivable. With delete a constraint trigger we can probably even make it safe. But there is stuff to explore here.

To clarify the reqested change. A foreign key constraint is not as easily added as that is not yet a postgreSQL feature. I refrain from that request. But merging code into a release to conduct an experiment still doesn't sound right. So at least we need to put the code populating and using the field behind an experiment-feature-flag.

Issue #6772, opened 2025-07-11
publish task fails for repository with 267000 RPMs


dkliban:
**Version**
3.83.0

**Describe the bug**
I synced packages from `https://download.copr.fedorainfracloud.org/results/@rubygems/rubygems/fedora-rawhide-x86_64/` using the `on_demand` policy. Then I tried to create a publication. The following error occurred:

```
$ pulp -p prod task show --href /api/pulp/dkliban-test/api/v3/tasks/0197f8e4-85ed-792d-abd8-6fee51e192cb/
{
  "pulp_href": "/api/pulp/dkliban-test/api/v3/tasks/0197f8e4-85ed-792d-abd8-6fee51e192cb/",
  "prn": "prn:core.task:0197f8e4-85ed-792d-abd8-6fee51e192cb",
  "pulp_created": "2025-07-11T09:50:15.534397Z",
  "pulp_last_updated": "2025-07-11T09:50:15.534418Z",
  "state": "failed",
  "name": "pulp_rpm.app.tasks.publishing.publish",
  "logging_cid": "60e21812c95140f4beb2d684ffcfcd23",
  "created_by": null,
  "unblocked_at": "2025-07-11T09:50:15.622027Z",
  "started_at": "2025-07-11T09:50:15.714680Z",
  "finished_at": "2025-07-11T09:50:23.477677Z",
  "error": {
[...]

PR #6784, opened 2025-07-14
Fix RepositoryVersion.content query failing when >65K units

gerrod3:
fixes: #6772

This is the orm magic breaks down to this sql:
```SQL
SELECT *
FROM "core_content" 
WHERE "core_content"."pulp_id" IN (
    SELECT unnest(U0."content_ids") AS "cids" 
    FROM "core_repositoryversion" U0
    WHERE U0."pulp_id" = 01980713-3ed0-7f9a-b1f5-20e774f746ae
)
```

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -890,7 +890,15 @@ class RepositoryVersion(BaseModel):
         if content_qs is None:
             content_qs = Content.objects
 
-        return content_qs.filter(pk__in=self._get_content_ids())
+        content_ids = self._get_content_ids()
+        if isinstance(content_ids, list) and len(content_ids) >= 65535:
+            # Workaround for PostgreSQL's limit on the number of parameters in a query
+            content_ids = (
+                RepositoryVersion.objects.filter(pk=self.pk)
+                .annotate(cids=Func(F("content_ids"), function="unnest"))
+                .values_list("cids", flat=True)
+            )
+        return content_qs.filter(pk__in=content_ids)
 
     @property
     def content(self):
--- a/CHANGES/6772.bugfix
+++ b/CHANGES/6772.bugfix
@@ -1,0 +1,1 @@
+Fixed the new content set optimization failing when the RepositoryVersion grew larger than 65K content units.


pulpcore/app/models/repository.py
```diff
@@ -890,7 +890,15 @@ def get_content(self, content_qs=None):
         if content_qs is None:
             content_qs = Content.objects
 
-        return content_qs.filter(pk__in=self._get_content_ids())
+        content_ids = self._get_content_ids()
+        if isinstance(content_ids, list) and len(content_ids) >= 65535:
```
jobselko, line 894: What about making a constant from this number, e.g. `POSTGRESQL_MAX_PARAMS`?

pulpcore/app/models/repository.py
```diff
@@ -890,7 +890,15 @@ def get_content(self, content_qs=None):
...
+            # Workaround for PostgreSQL's limit on the number of parameters in a query
+            content_ids = (
+                RepositoryVersion.objects.filter(pk=self.pk)
+                .annotate(cids=Func(F("content_ids"), function="unnest"))
+                .values_list("cids", flat=True)
+            )
```
jobselko, line 900: Could it happen that the result will be (again) greater than 65535? Maybe splitting the query into batches would be safer?

pulpcore/app/models/repository.py
bmbouter: From reading [this](https://www.postgresql.org/docs/current/limits.html) I think it's a hard-limit

pulpcore/app/models/repository.py
bmbouter: I think this reads from the db directly is how it avoids the limit.

gerrod3: Yes it reads the values from the db so it doesn't hit the limit. Interestingly the limit is only hit when using `.iterator()`. Seems the db cursor has the parameter limit, not the query itself.

jobselko: I meant that this result is used in `return content_qs.filter(pk__in=content_ids)`, and the issue could happen there (if not, please ignore my comment).

gerrod3: `values_list` returns a queryset, even with `flat=True`, so `content_ids` isn't evaluated beforehand, thus it can be used in the `content_qs` and properly combined in SQL. Actually `values("content_ids")` or `values_list("content_ids")` would both be valid here, they generate the same SQL.

jobselko: I understand now, thanks for the explanation!

jobselko (approved):

Issue #6799, opened 2025-07-22
Workers missing heartbeats


dkliban:
**Version**
3.82

**Describe the bug**
Our deployment has 64 worker and it processes 10K tasks per day. Our logs constantly show messages 

`pulp [None]: pulpcore.tasking.worker:INFO: Worker '1@pulp-worker-648b7974bc-hfpzn' is back online.`

The worker that misses its heartbeat is constantly receiving a notification to wakeup so the `self.beat()` inside the `self.sleep()` method is not getting called. 

The worker also doesn't find a task to supervise because other workers are picking up the tasks. So the `self.beat()` method inside the `self.supervise_task()` method is not getting called.

As a result of this, the workers goes more than 30s between heartbeats. 

I am attaching the verbose logs we generated in production. 

[worker-full-log.txt](https://github.com/user-attachments/files/21370102/worker-full-log.txt)

PR #6800, opened 2025-07-22
Added a call to beat() in handle_available_tasks() of Worker.

dkliban:
fixes: #6799

### DIFF
--- a/CHANGES/6799.bugfix
+++ b/CHANGES/6799.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug where workers would miss heartbeating in very busy deployments.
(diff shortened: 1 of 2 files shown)


pedro-psb (commented):
I wonder if this won't increase our average heartbeat rate (and if that have a meaningful impact, e.g, on database stress with lots of workers). IHO that's an important property of the tasking system. Ideally we want an even and stable heartbeat rate, but that's a bit tricky without using a clocks.

Having said that, I'm ok with the change as long as nothing breaks and we don't see any obvious regressions. Great analysis on the issue btw!

PR #6811, opened 2025-07-29
Refactor listen/notify as a pubsub backend

pedro-psb:


pulpcore/tasking/pubsub.py
```diff
@@ -0,0 +1,150 @@
...
+    def __init__(self):
+        self.subscriptions = []
+        self.message_buffer = []
+        PostgresPubSub.cursor = connection.cursor()
+        connection.connection.add_notify_handler(self._store_messages)
+        # Handle message readiness
+        # We can use os.evenfd in python >= 3.10
+        self.sentinel_r, self.sentinel_w = os.pipe()
+        os.set_blocking(self.sentinel_r, False)
+        os.set_blocking(self.sentinel_w, False)
```
mdellweg, line 75: Is this really related? I thought this os.pipe was primarily part of the select calls to receive the signals like SIGINT, SIGTERM.

pedro-psb: I added this to make the semantics of the PubSub readiness tied to the data it actually stores, which is the `self.message_buffers` queue (not the connection buffer). If I just use connection.connection, test_pubsub:111 fails as the pg_notify_callback is called automatically on same-process connections and is empty at this point.

But I don't feel very good about how complex this looks. I think using [notifies](https://www.psycopg.org/psycopg3/docs/api/connections.html#psycopg.Connection.notifies) may make our lives easier

pedro-psb:
Here is one of the general ideas for why I prefer to not add a callback into the pubsub interface (the other is testability).

image
image

PR #6813, opened 2025-07-29
Redirect checkpoint URLs without a trailing slash

### LINKED ISSUE #6812: Accessing a checkpoint without a trailing slash return 500.
**Version**
3.83.2

**Describe the bug**
When accessing a checkpoint URL without a trailing slash, we get a 500 response.

**To Reproduce**
Access a checkpoint URL without a trailing slash (e.g. https://example.com/checkpoint/repos/test/20250728T185958Z).

**Expected behavior**
The request is redirected to the correct URL with the trailing slash (e.g. https://example.com/checkpoint/repos/test/20250728T185958Z/).

Moustafa-Moustafa:
Redirect checkpoint URLs without a trailing slash to add a slash

closes #6812

### DIFF
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -705,8 +705,11 @@ class Handler:
         rel_path = rel_path[len(distro.base_path) :]
         rel_path = rel_path.lstrip("/")
 
-        if rel_path == "" and not path.endswith("/"):
-            # The root of a distribution base_path was requested without a slash
+        # Check if we need to redirect to add a trailing slash for distro root
+        if not path.endswith("/") and (
+            rel_path == ""  # Distro root
+            or (distro.checkpoint and "/" not in rel_path)  # Timestamped checkpoint root
+        ):
             raise HTTPMovedPermanently(f"{request.path}/")
 
         original_rel_path = rel_path
--- a/pulpcore/tests/functional/api/using_plugin/test_checkpoint.py
+++ b/pulpcore/tests/functional/api/using_plugin/test_checkpoint.py
@@ -87,17 +87,39 @@ class TestCheckpointDistribution:
         assert Handler._format_checkpoint_timestamp(pubs[3].pulp_created) in checkpoints_ts
 
     @pytest.mark.parallel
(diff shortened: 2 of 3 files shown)


pulpcore/content/handler.py
```diff
@@ -705,8 +705,11 @@ async def _match_and_stream(self, path, request):
         rel_path = rel_path[len(distro.base_path) :]
         rel_path = rel_path.lstrip("/")
 
-        if rel_path == "" and not path.endswith("/"):
-            # The root of a distribution base_path was requested without a slash
+        # Check if we need to redirect to add a trailing slash for distro root
+        if not path.endswith("/") and (
+            rel_path == ""  # non-checkpoint distro root
+            or (distro.checkpoint and rel_path and "/" not in rel_path)  # Checkpoint distro root
+        ):
```
mdellweg: If `bool(rel_path) == False` can it be something else then `""`? I don't think we need to check that twice.

pulpcore/tests/functional/api/using_plugin/test_checkpoint.py
```diff
@@ -87,18 +87,31 @@ def test_base_path_lists_checkpoints(self, setup, http_get, distribution_base_ur
...
 
+        # Test a checkpoint distro listing path
         response = http_get(distribution_base_url(distribution.base_url[:-1])).decode("utf-8")
         checkpoints_ts = set(re.findall(r"\d{8}T\d{6}Z", response))
 
         assert len(checkpoints_ts) == 2
         assert Handler._format_checkpoint_timestamp(pubs[1].pulp_created) in checkpoints_ts
         assert Handler._format_checkpoint_timestamp(pubs[3].pulp_created) in checkpoints_ts
 
+        # Test a root checkpoint distro with a specific timestamp
```
mdellweg: Instead of introducing comments to describe the two tests that you have in the function, I suggest to make it two test functions with names describing the specified behaviour.
The setup fixture is scoped to the class, so we will not waste extra time on that.

Moustafa-Moustafa:
@mdellweg, looks like one job failed with a transient error. Can you trigger a rerun of this job?

PR #6816, opened 2025-07-30
Fix scheduled purge tasks to handle all domains

### LINKED ISSUE #6815: Purge task does not remove failed tasks
**Version**
3.82

**Describe the bug**
We set the TASK_PROTECTION_TIME to 20160. The purge task ran. However a lot of older tasks still remained in the database.

mdellweg:
fixes #6815


ggainey:
The azure failure is unrelated to the PR - merging

PR #6836, opened 2025-08-06
Py3.11

mdellweg:


ci_requirements.txt
```diff
@@ -1 +1 @@
-rhsm
+
```
jobselko, line 1: Do we need to keep an empty file?

CHANGES/6262.removal
jobselko, line 1: Is not this rather a feature? I am really not sure

pulpcore/tests/functional/api/using_plugin/test_proxy.py
pedro-psb, line 1: Nice.
Do you know why CI bindings suddenly started failing on this?

jobselko:
`pulpcore/app/settings.py` contains a condition about `sys.version_info < (3, 10)`. This should be removed as well.

CHANGES/6262.removal
mdellweg: It's a backwards incompatible change. Basically we remove support for a bunch of python versions.

pulpcore/tests/functional/api/using_plugin/test_proxy.py
mdellweg: No, i can only guess that a second or higher level dependency version was "unlocked" by the new python version.

PR #6838, opened 2025-08-06
[PULP-659] Remove legacy django storages

pedro-psb:
This PR does:
* Remove support for using DEFAULT_FILE_STORAGE (added in: https://github.com/pulp/pulpcore/pull/6058)
* Apply plugin-template to remove both storage settings from CI settings.py (See: https://github.com/pulp/plugin_template/pull/975)

## For Plugins

- When supporting 3.85, set `test_storages_compat_layer: False` in `template_config.yml` so it only renders the newer storage settings for every storage case (s3 and azure).
- The plugin pulpcore lowerbound must be `>=3.70`. If a pulpcore `<3.70` is used, the CI installation can break (most likely on the lowerbound scenario), as it doest support the newer storage form.

### DIFF
--- a/.ci/ansible/settings.py.j2
+++ b/.ci/ansible/settings.py.j2
@@ -26,10 +26,22 @@ API_ROOT = {{ api_root | repr }}
 {% endfor %}
 {% endif %}
 
-{% if s3_test | default(false) %}
-MEDIA_ROOT: ""
-S3_USE_SIGV4 = True
-{% if test_storages_compat_layer is defined and test_storages_compat_layer %}
+{# =======================================
+Macros for legacy and new storage settings
+========================================== -#}
+
+{%- macro s3_settings(legacy) -%}
+    {%- if legacy %}
+DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
+AWS_ACCESS_KEY_ID = "{{ minio_access_key }}"
+AWS_SECRET_ACCESS_KEY = "{{ minio_secret_key }}"
+AWS_S3_REGION_NAME = "eu-central-1"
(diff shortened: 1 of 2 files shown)


.ci/ansible/settings.py.j2
```diff
@@ -48,34 +60,93 @@ STORAGES = {
...
+    {%- if test_storages_compat_layer is true -%}
+        {%- if s3_test | default(false) -%}{{ s3_settings(legacy=False) }}{%- endif -%}
+        {%- if azure_test | default(false) -%}{{ azure_settings(legacy=True) }}{%- endif -%}
+        {%- if gcp_test | default(false) -%}{{ gcp_settings(legacy=True) }}{%- endif -%}
+    {%- elif test_storages_compat_layer is false -%}
+        {%- if s3_test | default(false) -%}{{ s3_settings(legacy=False) }}{%- endif -%}
+        {%- if azure_test | default(false) -%}{{ azure_settings(legacy=False) }}{%- endif -%}
+        {%- if gcp_test | default(false) -%}{{ gcp_settings(legacy=False) }}{%- endif -%}
+    {%- endif -%}
+{%- endif -%}
```
mdellweg, line 152:
```suggestion
{%- if s3_test | default(false) -%}{{ s3_settings(legacy=test_storages_compat_layer is not defined) }}{%- endif -%}
{%- if azure_test | default(false) -%}{{ azure_settings(legacy=test_storages_compat_layer | default(True)) }}{%- endif -%}
{%- if gcp_test | default(false) -%}{{ gcp_settings(legacy=test_storages_compat_layer | default(True)) }}{%- endif -%}
```

mdellweg (approved):

PR #6841, opened 2025-08-07
Add AppStatus table

mdellweg:
This is the first of a series of changes that need to be spaced out by releases to consolidate the status tables of different components on a single one.


pulpcore/app/models/status.py
```diff
@@ -47,9 +49,138 @@ def missing(self, age=None):
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.ttl = timedelta(seconds=self._APP_TTL[self.app_type])
+        self._old_status = None
+
+    def delete(self, *args, **kwargs):
+        # adelete will call into this, so we should not replicate that one here.
+        if self._old_status is not None:
+            self._old_status.delete(*args, **kwargs)
+        super().delete(*args, **kwargs)
```
pedro-psb, line 124: > adelete will call into this, so we should not replicate that one here.

So django.Model .adelete calls .delete?

pulpcore/app/models/status.py
```diff
@@ -47,9 +49,138 @@ def missing(self, age=None):
...
+        Update the last_heartbeat field to now and save it.
+
+        Only the last_heartbeat field will be saved. No other changes will be saved.
+
+        Raises:
+            ValueError: When the model instance has never been saved before. This method can
+                only update an existing database record.
+        """
+        self._old_status.save_heartbeat()
+        self.save(update_fields=["last_heartbeat"])
```
pedro-psb, line 157: :ok_hand:

pulpcore/migrations.py
pedro-psb, line 1: Just to make sure I understand this:
If the migration starting point does contain the split Statuses, then it will use that to check the component version compat. Otherwise, it means the starting point is a future state where those are removed (in the next release, possibly, for the zdu) and we can use the AppStatus.
And theoretically there is no state in which neither of those exist, but we have an assert just to be sure.

pulpcore/tasking/worker.py
```diff
@@ -74,7 +74,10 @@ def __init__(self, auxiliary=False):
         self.versions = {app.label: app.version for app in pulp_plugin_configs()}
         self.cursor = connection.cursor()
         try:
-            self.worker = Worker.objects.create(name=self.name, versions=self.versions)
+            self.app_status = AppStatus.objects.create(
+                name=self.name, app_type="worker", versions=self.versions
+            )
+            self.worker = self.app_status._old_status
```
pedro-psb, line 80: I see self.worker is used in supervise, which uses the old worker status because Task.worker requires a table Worker. Can' we already change that foreign key on Task to be related to a StatusApp?

pulpcore/app/models/status.py
mdellweg: So I learned the hard way...

pulpcore/tasking/worker.py
mdellweg: Changing the foreign key constraint to point to a different table is something we can never do with ZDU. That's why we need to introduce a new field. This one will eventually drop, after which we can start to think about renaming it back (I doubt the name is worth that trouble.)

pulpcore/migrations.py
mdellweg: Well we start with only the old status. Then we add the new table to the database. Now we start up new components that populate both tables. We still can only rely on the old status, because we might have old code running. With the next upgrade we can start to rely on the new table and stop populating the old. And yet another release later we can drop the old tables. Yes, this is complex.

pedro-psb (approved):

pulpcore/tasking/worker.py
pedro-psb: Yeah. One needs to be careful when thinking about ZDU...

PR #6842, opened 2025-08-08
chore: update dependencies

reneleonhardt:
### Chores
- Update dependencies
- Let dependabot update dependencies


reneleonhardt:
I reverted all updates in workflows, why are those files not marked as generated in .gitattributes to prevent manual changes?
template_config.yml doesn't contain versions, how can all those years old releases be generated with latest versions?

gerrod3:
@reneleonhardt Thanks for the PR, but unfortunately I don't think we will be able to accept this. All of the CI for plugins & core are managed by the plugin template (https://github.com/pulp/plugin_template) which uses `template_config.yml` and auto-updates each week through the `update_ci.yml` workflow. The workflow runs the plugin template against every supported branch so the CI is up-to-date for all release branches. As for the images the CI uses those are built here: https://github.com/pulp/pulp-oci-images/. We might update to CentOS Stream 10 soon, but since we just recently updated to Python 3.11 and Stream 10's default is Python 3.12 we might wait till Pulp 4 to bump the base image.

mdellweg:
Sorry, I agree with @gerrod3 here. The dependencies for the code we ship are covered by dependabot, and some of the proposed changes outright ignore comments we put in place for a reason. I do not see the benefit, so I'm going to close this PR.

PR #6847, opened 2025-08-11
[PULP-571] Add app_lock to task 

mdellweg:


pedro-psb:
For context: this is an _additive_ change (main locking mechanism is still the advisory locks) that introduces task locking using a new `Task.app_lock` field, which points to any `AppStatus` (api, content or worker). It will only be possible to transition to that as the main task lock after the AppStatus is stable (it was broken on multiple release to ensure zero-downtime-upgrade).

PR #6859, opened 2025-08-15
[PULP-560] Add results to tasks

jobselko:


pulpcore/app/serializers/task.py
```diff
@@ -88,6 +88,10 @@ class TaskSerializer(ModelSerializer):
         help_text=_("A list of resources required by that task."),
         read_only=True,
     )
+    result = fields.JSONDictField(
+        help_text=_("A JSON object representing the result of this task."),
```
mdellweg:
```suggestion
        help_text=_("The result of this task."),
```
I think all the places showing this documentation already mention it's an object. ;)

pulpcore/tests/functional/api/using_plugin/test_tasks.py
```diff
@@ -23,6 +23,15 @@ def distribution(file_bindings, file_repo, gen_object_with_cleanup):
     return distribution
 
 
+def test_task_result(file_bindings, file_repo, monitor_task, pulpcore_bindings):
+    response = file_bindings.DistributionsFileApi.create(
+        {"name": str(uuid4()), "repository": file_repo.pulp_href, "base_path": str(uuid4())}
+    )
+    dist = monitor_task(response.task).created_resources[0]
```
mdellweg: Basically we want to replace the need for created_resources.
So the result of monitor_task is the finished task, and from that we take the result and assert on it, right?
How about:
```python
body = {...}
task = monitor_task(....create(body).task)
assert task.result["name"] == body["name"]
```
?

pulpcore/app/models/task.py
```diff
@@ -241,10 +243,12 @@ def set_completed(self):
         rows = Task.objects.filter(pk=self.pk, state=TASK_STATES.RUNNING).update(
             state=TASK_STATES.COMPLETED,
             finished_at=finished_at,
+            result=result,  # tested version, might be null
         )
         if rows == 1:
             self.state = TASK_STATES.COMPLETED
             self.finished_at = finished_at
+            self.result = result
```
mdellweg, line 265: 👍

pulpcore/app/tasks/base.py
```diff
@@ -42,6 +42,7 @@ def general_create(app_label, serializer_name, *args, **kwargs):
         instance = instance.cast()
     resource = CreatedResource(content_object=instance)
     resource.save()
+    return serializer.data
```
mdellweg, line 47: The same trick should do in the update tasks.

pulpcore/tasking/tasks.py
```diff
@@ -123,7 +123,8 @@ def _execute_task(task):
         _logger.info("\n".join(traceback.format_list(traceback.extract_tb(tb))))
         send_task_notification(task)
     else:
-        task.set_completed()
+        task.result = result
```
mdellweg:
```suggestion
```

pulpcore/app/viewsets/base.py
```diff
@@ -469,7 +469,7 @@ def create(self, request, *args, **kwargs):
             tasks.base.general_create,
             exclusive_resources=self.async_reserved_resources(None),
             args=(app_label, serializer.__class__.__name__),
-            kwargs={"data": request.data},
+            kwargs={"data": request.data, "context": {"request": request.data}},
```
mdellweg: Hmm. `request.data` is most definitely wrong. Can you try with `context.set_default("request", None)` in the `general_create` task instead?

pulpcore/tests/functional/api/using_plugin/test_tasks.py
```diff
@@ -169,3 +169,20 @@ def test_reserved_resources_filter(setup_filter_fixture, pulpcore_bindings):
     assert set(h.pulp_href for h in prn_results.results) == task_hrefs
     mixed_results = pulpcore_bindings.TasksApi.list(exclusive_resources__in=[repo_prn, remote_prn])
     assert mixed_results.count == 0
+
+
+def test_task_result(file_bindings, monitor_task):
```
mdellweg: Very nice. Can you add the `add_to_cleanup` fixture and register the distribution for cleanup after creating it?
Also let's use a random name and only update and assert on the base_path (also random...).
We should be able to mark this test with `@pytest.mark.parallel` then.

pulpcore/app/models/task.py
```diff
@@ -229,22 +235,29 @@ def set_running(self):
...
         """
         Set this Task to the completed state, save it, and log output in warning cases.
 
         This updates the :attr:`finished_at` and sets the :attr:`state` to :attr:`COMPLETED`.
+        If `result` is provided, the :attr:`result` contains the result of the task.
         """
+        if isinstance(result, Model):
+            serialized = serializers.serialize("json", [result], cls=DjangoJSONEncoder)
+            result = json.loads(serialized)[0]
+
```
mdellweg: If we get to a place where the task result is not yet properly serialized (only the task itself can know what 'proper' is here), we should not try to fix the situation.
We can detect it, raise a deprecation warning for the plugin writer and continue to return "None" for now.

pulpcore/app/serializers/publication.py
```diff
@@ -374,7 +374,11 @@ def get_no_content_change_since(self, obj):
         repo_version = obj.repository_version
 
         if publication or (repo_version and not obj.SERVE_FROM_PUBLICATION):
-            return obj.pulp_last_updated
+            timestamp = obj.pulp_last_updated
+            # Return timestamp as a string for the serialized response
+            if timestamp:
+                return timestamp.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+            return None
```
mdellweg: This looks like patching symptoms. Is the timestamp still presented as such in `serializer.data`? Then we may need to look into the drf renderers to get to proper json, or we can teach the JSONEncoder for the db field to do the right thing.

mdellweg: Maybe using the DjangoJSONEncoder is sufficient.
https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/task.py#L138

pulpcore/app/models/task.py
```diff
@@ -160,6 +162,8 @@ class Task(BaseModel, AutoAddObjPermsMixin):
     immediate = models.BooleanField(default=False, null=True)
     deferred = models.BooleanField(default=True, null=True)
 
+    result = models.JSONField(null=True, encoder=DjangoJSONEncoder)
```
mdellweg: Should it have `default=None` as we only intend to populate it later?

pulpcore/app/models/task.py
```diff
@@ -229,22 +233,35 @@ def set_running(self):
...
         Set this Task to the completed state, save it, and log output in warning cases.
 
         This updates the :attr:`finished_at` and sets the :attr:`state` to :attr:`COMPLETED`.
+        If `result` is provided, the :attr:`result` contains the result of the task.
         """
+        try:
+            result = json.dumps(result, cls=DjangoJSONEncoder)
+        except (TypeError, ValueError):
+            result = None
+            _logger.warning(
```
mdellweg: This should use the deprecation logger and you can probably mark this workaround to disappear in 3.100.

pulpcore/app/serializers/task.py
```diff
@@ -88,6 +88,10 @@ class TaskSerializer(ModelSerializer):
         help_text=_("A list of resources required by that task."),
         read_only=True,
     )
+    result = fields.JSONDictField(
+        help_text=_("The result of this task."),
+        read_only=True,
+    )
```
mdellweg: And this is now what defines the openapi representation of this field.

```
$ cat api.json | jq '.components.schemas.TaskResponse.properties.result'
{
  "type": "object",
  "readOnly": true,
  "description": "The result of this task."
}
```
I think what we actually want is:
```
{
  "readOnly": true,
  "description": "The result of this task."
}
```
Does the drf-json field give us that?

jobselko: Yes, `JSONDictField` has `@extend_schema_field(OpenApiTypes.OBJECT)`

mdellweg: Exactly, but I thought we wanted `OpenApiTypes.ANY` if such a thing exists.

jobselko: For `JSONDictField` in general, or only for the `result` field?

pulpcore/app/models/task.py
```diff
@@ -229,22 +233,36 @@ def set_running(self):
...
         """
         Set this Task to the completed state, save it, and log output in warning cases.
 
         This updates the :attr:`finished_at` and sets the :attr:`state` to :attr:`COMPLETED`.
+        If `result` is provided, the :attr:`result` contains the result of the task.
         """
+        try:
+            json.dumps(result, cls=DjangoJSONEncoder)
+        except (TypeError, ValueError):
+            deprecation_logger.warning(
```
jobselko, line 246: Should I add a changelog entry since there is a deprecation warning?

pulpcore/app/serializers/task.py
mdellweg: Just in this case...
The JSON "object" corresponds to the python "dict".

pulpcore/app/models/task.py
mdellweg: Oh yes, good call, it should be in the plugin_api section.

pulpcore/app/serializers/task.py
jobselko: I kept `object` and made it nullable, so the type is now `["object", "null"] `. I think this is better than `"any"`.
Btw, should the same change be made for `error`, since both are json and can be null?

pulp_file/app/tasks/publishing.py
```diff
@@ -51,6 +53,9 @@ def publish(manifest, repository_version_pk, checkpoint=False):
 
         log.info(_("Publication: {publication} created").format(publication=publication.pk))
 
+        serialized_data = serializers.serialize("json", [publication])
+        publication = json.loads(serialized_data)[0]
```
mdellweg: Can you try `FilePublicationSerializer(instance=publication).data` instead?

pulp_file/app/tasks/synchronizing.py
```diff
@@ -65,6 +67,10 @@ def synchronize(remote_pk, repository_pk, mirror, url=None):
 
         log.info(_("Publication: {publication} created").format(publication=publication.pk))
 
+    if rv:
+        serialized_data = serializers.serialize("json", [rv])
+        rv = json.loads(serialized_data)[0]
+
```
mdellweg: The same pattern should apply here.

pulpcore/app/models/task.py
mdellweg: Also one general changelog for the user should be there, i think.

pulpcore/app/serializers/task.py
```diff
@@ -88,6 +89,7 @@ class TaskSerializer(ModelSerializer):
         help_text=_("A list of resources required by that task."),
         read_only=True,
     )
+    result = serializers.SerializerMethodField()
```
mdellweg: looking at this issue https://github.com/tfranzel/drf-spectacular/issues/1340 we should get the proper annotation by not even having this line, or (maybe better) specifying it as serializers.JSONField where you can add the help text like you had is before when using the subclass JSONDictField that brought the wrong annotation.

pulpcore/app/serializers/task.py
jobselko: Reworked to simply use `serializers.JSONField`.

pulpcore/app/tasks/base.py
```diff
@@ -33,6 +33,7 @@ def general_create(app_label, serializer_name, *args, **kwargs):
     data = kwargs.pop("data", None)
```
pedro-psb, line 34: `general_create_from_temp_file` should have a return in line 22.

CHANGES/plugin_api/6901.deprecation
pedro-psb, line 1: Worth mentioning it'll be removed in 3.100, as the deprecation message says.

pedro-psb (changes requested):

PR #6876, opened 2025-08-21
Fix signalling around unblock

### LINKED ISSUE #6873: Tasks stuck in 3.84.0
**Version**
pulpcore 3.84.0

**Describe the bug**
We upgraded our pulpcore version from 3.83.2 to 3.84.0 and since then, tasks have sometimes been stuck at "waiting". The problem surfaces in our test environment but we can also produce it locally in our dev environments. It seems to happen pretty reliably during our package directory upload test which uploads files concurrently. These tasks start within < 1 second. If I turn off concurrency, the problem seems to go away as tasks queue up at > 1 second apart.

**To Reproduce**
Execute two tasks concurrently or in very quick succession (< 1 second). I think maybe part of the problem is that we're only running one worker in our test/dev environments.

**Expected behavior**
Tasks don't get stuck at waiting.

**Additional context**
Here's an example of two tasks (one stuck at waiting). The second was created at 10:54am and it's currently 10:57am.

```json
{
         "id": "/pulp/api/v3/tasks/0198c1f7-2463-7a3c-b18b-54da349d1002/",
         "pulp_created": "2025-08-19T10:54:16.190522Z",
         "state": "waiting",
         "name": "pulpcore.app.tasks.base.general_create",
         "logging_cid": "d543df56-d072-4a75-a7f0-71fe4671d273",
         "started_at": null,
         "finished_at": null,
         "error": null,
         "worker": null,
         "progress_reports": [],
         "created_resources": [],
            "shared:prn:core.domain:431fb816-c18e-4b8a-83c8-8b5b0013f67c"
         ]
      },
      {
         "id": "/pulp/api/v3/tasks/0198c1f7-2460-7129-ad53-9c6823985ff6/",
         "pulp_created": "2025-08-19T10:54:16.168997Z",
         "state": "completed",
         "name": "pulpcore.app.tasks.base.general_create",
         "logging_cid": "d543df56-d072-4a75-a7f0-71fe4671d274",
         "started_at": "2025-08-19T10:54:16.284407Z",
         "finished_at": "2025-08-19T10:54:16.438698Z",
         "error": null,
         "worker": ""/pulp/api/v3/workers/0198c1f4-c606-7c87-b0e3-3bde13a83243/",
         "progress_reports": [],
         "created_resources": [
            ""/pulp/api/v3/content/file/files/0198c1f7-2567-727d-9b14-64b02ed712c3/"
         ],
            "shared:prn:core.domain:431fb816-c18e-4b8a-83c8-8b5b0013f67c"
         ]
      },
```

mdellweg:
TODO Clean up changelog before approving!!!

fixes #6873

### DIFF
--- a/CHANGES/6873.bugfix
+++ b/CHANGES/6873.bugfix
@@ -1,0 +1,1 @@
+Fix situations with a single worker per installation failing to unblock all tasks that come in quick succession.
--- a/pulpcore/tasking/worker.py
+++ b/pulpcore/tasking/worker.py
@@ -194,6 +194,9 @@ class PulpcoreWorker:
                 "Cleanup record of missing %s process %s.", app_worker.app_type, app_worker.name
             )
         qs.delete()
+        # This will also serve as a pacemaker because it will be triggered regularly.
+        # Don't bother the others.
+        self.wakeup_unblock = True
 
     def beat(self):
         if self.app_status.last_heartbeat < timezone.now() - self.heartbeat_period:
@@ -214,7 +217,7 @@ class PulpcoreWorker:
                 # to be able to report on a congested tasking system to produce reliable results.
                 self.record_unblocked_waiting_tasks_metric()
 
-    def notify_workers(self, reason="unknown"):
+    def notify_workers(self, reason):
         self.cursor.execute("SELECT pg_notify('pulp_worker_wakeup', %s)", (reason,))
 
     def cancel_abandoned_task(self, task, final_state, reason=None):
@@ -274,25 +277,28 @@ class PulpcoreWorker:
         Also it clears the notification about tasks to be unblocked and sends the notification that
         new unblocked tasks are made available.
 
-        Returns the number of new unblocked tasks.
+        Returns None if another worker held the lock, True if unblocked tasks exist, else False.
         """
 
         assert not self.auxiliary
 
-        count = 0
-        self.wakeup_unblock_tasks = False
+        self.wakeup_unblock = False
         with contextlib.suppress(AdvisoryLockError), PGAdvisoryLock(TASK_UNBLOCKING_LOCK):
-            if count := self._unblock_tasks():
+            self._unblock_tasks()
+
+            if (
+                Task.objects.filter(state__in=TASK_INCOMPLETE_STATES, app_lock=None)
+                .exclude(unblocked_at=None)
+                .exists()
+            ):
                 self.notify_workers(TASK_WAKEUP_HANDLE)
-        return count
+                return True
+            return False
+        return None
 
     def _unblock_tasks(self):
-        """Iterate over waiting tasks and mark them unblocked accordingly.
+        """Iterate over waiting tasks and mark them unblocked accordingly."""
 
-        Returns the number of new unblocked tasks.
-        """
-
-        count = 0
         taken_exclusive_resources = set()
         taken_shared_resources = set()
         # When batching this query, be sure to use "pulp_created" as a cursor
@@ -320,7 +326,6 @@ class PulpcoreWorker:
                         task.pulp_domain.name,
                     )
                     task.unblock()
-                    count += 1
 
             elif (
                 task.state == TASK_STATES.WAITING
@@ -339,7 +344,6 @@ class PulpcoreWorker:
                     task.pulp_domain.name,
                 )
                 task.unblock()
-                count += 1
             elif task.state == TASK_STATES.RUNNING and task.unblocked_at is None:
                 # This should not happen in normal operation.
                 # And it is only an issue if the worker running that task died, because it will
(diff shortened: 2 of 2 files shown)


mdellweg:
@daviddavis would you mind taking a look at this?

daviddavis:
I'm not sure about the code as I haven't worked with Pulp's tasking system in a while, but I upgraded pulpcore to 3.84 in our pulp image and applied this change as a patch and we're still getting stuck tasks:

```
-[ RECORD 1 ]-------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
pulp_id                   | 0198ce0f-dd83-7f52-9c34-56670bb89ef3
pulp_created              | 2025-08-21 19:16:43.061592+00
pulp_last_updated         | 2025-08-21 19:16:43.013158+00
state                     | waiting
name                      | pulpcore.app.tasks.base.general_create
started_at                |
finished_at               |
error                     |
worker_id                 |
parent_task_id            |
logging_cid               | ab1a0818d3db40ff8ccbec75fdd22d29
reserved_resources_record | {shared:prn:core.domain:431fb816-c18e-4b8a-83c8-8b5b0013f67c}
task_group_id             |
pulp_domain_id            | 431fb816-c18e-4b8a-83c8-8b5b0013f67c
versions                  | "core"=>"3.84.0"
enc_args                  | ["gAAAAABop3Cbk_oPaXSI_oQi8ZxHNMGwJhRKGQJYj9jAgnDFl4ibh24pywSeO979poZDj4R2deeAJySKjVTe2VcTsVZxnjaRWA==", "gAAAAABop3Cbt3QHMmKIc5Y9lqfcOPrsMZooDkjKx2c3ZJtEBbTUh-RKE7yIvwe9xILMZlKpKvBd32k6smMVgTXDqQbe4GRR2MZKUCB9m3HXfev7PfBN5j4="]
enc_kwargs                | {"data": {"artifact": "gAAAAABop3CbpAdLGGx5lZGK2V8usCkn9KkGKYW__T19T439aCOxmRgF3sIDGvHXgLYPLSARCBxDknACHY0fqbzRM2gwNtzvmFJEo4K47nny9vZD8N4MWQGP59hdbQjL06eP4d2erpRb94hHVtmpBDODZzDbYQtXZA==", "relative_path": "gAAAAABop3CbEM7EUBe9Kr4D_ePc5t0p16Zb5Up45FcFew5_XgV04xhnpM0vynK00ZqZBx-o7UsfMaGrvob6yQZ3hm3DXRKBXQ=="}, "context": {}}
unblocked_at              |
deferred                  | t
immediate                 | f
profile_options           |
```

mdellweg:
Is canceling a task part of the story when this happens? (I'm wondering if canceling a task properly retriggers unblocking subsequent tasks now.)

pedro-psb:
Ok, I get what is happening.

In the sleep state (also in the supervise code, fwiw), the worker will only wake up to unblock if there was a message in the pg connection in the first place.
https://github.com/pulp/pulpcore/blob/14322ad7266758eded8a6314cb2ca918b67112cc/pulpcore/tasking/worker.py#L370-L373

But it can happen (safe to say that it IS happening) that something flushes the connection before worker hits the `select.select`. Not sure exactly what, but maybe a heartbeat, some janitorial work or whatever. It's not a big surprise.

So even though  `worker.wakeup_unblock=True` at this point (flushing the pg connection means pg_handler was called and updated the worker state), now the connection is empty and when it hits `select.select` it won't try to unblock, because the unblock depends on the connection having something.

So we should have something like that in the places where this happens.

```diff
if connection.connection in r:
    connection.connection.execute("SELECT 1")
-    if self.wakeup_unblock:
-        self.unblock_tasks()
+if self.wakeup_unblock:
+    self.unblock_tasks()
```

Here are some actual (commented) logs from my experiments:

```python
--------------------------------------------------------------------------------
App.publish(wakeup-unblock)
Worker.received(pulp_worker_wakeup:unblock)
Worker.wakeup(unblock)
Worker.publish(handle)
Worker.received(pulp_worker_wakeup:handle)
Worker.wakeup(handle) self.wakeup_unblock=False self.wakeup_handle=True
Worker.started(pulpcore.app.tasks.test.sleep)
Worker.connection(has-msg=False)
--------------------------------------------------------------------------------
# At this point, the worker is on supervise and handles the dispatch wakeup notification
App.publish(wakeup-unblock)
Worker.received(pulp_worker_wakeup:unblock)

# here the worker is finishing the task, triggered a new unblock notify and received the notification itself
# when the subscriber is the same as the publisher, the connection buffer is not used...
# We already have self.wakeup_unblock=True, anyway
Worker.connection(has-msg=False)  # spying to see if there is something in the conn buffer. Never more
Worker.publish(unblock)
Worker.received(pulp_worker_wakeup:unblock)
Worker.finished(pulpcore.app.tasks.test.sleep)
Worker.connection(has-msg=False)

# We go to sleep state with self.wakeup_unblock=True but nothing in the connection buffer
Worker.sleep() self.wakeup_unblock=True self.wakeup_handle=False
Worker.connection(has-msg=False)

# and it ends here...
```

mdellweg:
Great analysis!

So what we could do in the current implementation is (in addition to what you suggested already) reduce the timeout of select to 0 whenever some message is pending. Or maybe we _want_ to introduce a bit of laziness there deliberately.

mdellweg:
> ps: this is basically what I wanted to make more predictable with the pubsub interface specification we've discussed a while back, more specifically with this [not so elegant but apparently functional implementation of fileno()](https://github.com/pulp/pulpcore/pull/6811/files#diff-ee60ba45b59c294023038f5df7121ec5b68676f3c9dfe3fb0f5b47ab9d584305R118-R124). In the context of the pubsub interface and from an user perspective (e.g, the worker), the fileno() doesn't refer specifically to pg connection, which is an implementation detail, but to an abstract internal buffer that either has or hasn't message in it.

These situations make me wonder (accepting that any sort of non-serial execution is complex) whether the whole concept could be expressed more intuitive in async python.

pedro-psb (approved):
Nice, it makes sense to shorten the timeout. I'm good with 0 for now.

A little on the fence with the pacemaker, but it's fine. We'll assure tasks get unblocked if anything goes wrong.

PR #6877, opened 2025-08-21
Consolidate AppStatus table (2/3)

mdellweg:
This second step adds a migration requirement on 3.85 and stops touching the old status tables, so they can safely be removed in a later release.

Do not merge before 3.87!


pulpcore/openapi/__init__.py
```diff
@@ -260,7 +260,7 @@ def get_parameter_slug_from_model(model, prefix, pulp_model_alias=None):
             prefix (str): Optional prefix to add to the slug
             pulp_model_alias (str): Optional model name to use instead of model.__name__
         Returns:
-            str: *pulp_href where * is the model name in all lower case letters
+            str: *_href where * is the model name in all lower case letters
```
pedro-psb: The use of asterisk is confusing, I though this was an accident.
Can you use some pythonic template expression? `"{model_name}_href"` or similar.

pulpcore/tasking/worker.py
```diff
@@ -385,13 +386,13 @@ def iter_tasks(self):
                         # Check if someone else changed the task before we got the lock.
                         task.refresh_from_db()
 
-                        if task.state == TASK_STATES.CANCELING and task.worker is None:
+                        if task.state == TASK_STATES.CANCELING:
                             # No worker picked this task up before being canceled.
                             if self.cancel_abandoned_task(task, TASK_STATES.CANCELED):
                                 # Continue looking for the next task without considering this
                                 # tasks resources, as we just released them.
                                 continue
```
pedro-psb, line 394: Why did we have this `task.worker is None` check before?
I tried to decipher "No worker picked this task up before being canceled.", but I dont think I've ever really understand that. Like, what precise steps can lead to that state.

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -565,9 +529,10 @@ def test_executes_in_task_worker(
                 immediate=True,
                 exclusive_resources=[COMMON_RESOURCE],
             )
+            task = pulpcore_bindings.TasksApi.read(task_href)
+            assert task.state == "waiting"
         task = monitor_task(task_href)
```
pedro-psb, line 534: :+1:

pulpcore/app/viewsets/task.py
```diff
@@ -354,22 +362,22 @@ class WorkerFilter(BaseFilterSet):
...
             return queryset.filter(pk__in=online_workers)
         else:
             return queryset.exclude(pk__in=online_workers)
 
     def filter_missing(self, queryset, name, value):
-        missing_workers = Worker.objects.missing()
+        missing_workers = AppStatus.objects.missing()
 
         if value:
             return queryset.filter(pk__in=missing_workers)
```
pedro-psb, line 383: It looks like those `.online()|.missing()` calls will return all online/missing app types, not just workers.
If that's the case, you should add a filter there.

pulpcore/app/views/status.py
```diff
@@ -113,7 +112,7 @@ def _get_db_conn_status():
             bool: True if there's a db connection. False otherwise.
         """
         try:
-            Worker.objects.count()
+            AppStatus.objects.count()
```
pedro-psb, line 115: Same filtering issue here. `Worker` previsouly were only workers. `AppStatus` are all of them without a filter.

pulpcore/app/management/commands/remove-plugin.py
```diff
@@ -53,10 +53,7 @@ def _check_pulp_services(self):
             "Checking if Pulp services are running, it can take up to {}s...".format(waiting_time)
         )
         while is_pulp_running and (time.time() - check_started) < waiting_time:
-            is_pulp_running = (
-                ContentAppStatus.objects.online().exists()
-                or Worker.objects.online_workers().exists()
-            )
+            is_pulp_running = AppStatus.objects.online().exists()
```
pedro-psb, line 56: The previous only checked Content and Worker, but it actually feels right to check all.

pedro-psb (changes requested):

pulpcore/app/views/status.py
ggainey: This is just checking "can I talk to the db?" - it could be *any* query, really. So this is fine.

pulpcore/tasking/worker.py
ggainey: If I start a task, and it's still sitting in the queue because no worker has been able to grab it yet (none available or I'm waiting on some resource) and I cancel it, it'll never have a task.worker.

If it *does* have a worker, and is CANCELING - that means that worker will finish up what it's doing, before we can just mark it as CANCELED.

pulpcore/app/views/status.py
pedro-psb: Right, `_get_db_conn_status` :facepalm:

pulpcore/tasking/worker.py
pedro-psb: Hmmm, I see. I was thinking in something more complicated because of the word *abandoned*.
This is fine then.

pulpcore/app/viewsets/task.py
pedro-psb: Hmmm... Looks like that also doesn't make a difference either.
I suppose the WorkerViewSet queryset is always passed to that filter, right?
So it will have a disjoint set with queryset, which is irrelevant for this filtering.

ggainey: I *think* that is correct, yes.

pulpcore/openapi/__init__.py
mdellweg:
```suggestion
            str: '{model_name_snake_case}_href'
```

pulpcore/app/viewsets/task.py
mdellweg: Yes, queryset already has the filter for the worker type.
If we added "missing" to the queryset class, and not only the manager class, we could "just" return `queryset.missing()`. I thought that was an improvement for a future pull request.

pulpcore/app/management/commands/remove-plugin.py
mdellweg: This just adds api to the mix, yes. (Why had it been missing in the first place? Grown historic system...)

Issue #6899, opened 2025-08-28
Acquired advisory lock but missed the app_lock for the task.


dkliban:
**Version**
3.86.0

**Describe the bug**
Workers are emitting: "Acquired advisory lock but missed the app_lock for the task. This should only happen during the upgrade phase to the new app_lock."

dkliban:
I was able to find all the logs with the same correlation id `dbc55a8168984ae1aa3f49d7c567c4f1`. 

```
2025-08-29 00:00:25.990 pulp-prod_pulp-worker-577f8cf569-7zpfw_pulp-worker
pulp [dbc55a8168984ae1aa3f49d7c567c4f1]: pulpcore.tasking.worker:ERROR: Acquired advisory lock but missed the app_lock for the task. This should only happen during the upgrade phase to the new app_lock.
	
2025-08-28 23:59:46.309 pulp-prod_pulp-worker-auxiliary-54b559f56f-6llpm_pulp-worker-auxiliary
pulp [dbc55a8168984ae1aa3f49d7c567c4f1]: pulpcore.tasking.tasks:INFO: Task completed 0198f31e-2e33-70b6-9d20-cdaa0129625b in domain: default, task_type: pulpcore.app.tasks.telemetry.otel_metrics, immediate: False, deferred: True, execution_time: 78010400 μs
	
2025-08-28 23:58:28.297 pulp-prod_pulp-worker-auxiliary-54b559f56f-6llpm_pulp-worker-auxiliary
pulp [dbc55a8168984ae1aa3f49d7c567c4f1]: pulpcore.tasking.tasks:INFO: Starting task id: 0198f31e-2e33-70b6-9d20-cdaa0129625b in domain: default, task_type: pulpcore.app.tasks.telemetry.otel_metrics, immediate: False, deferred: True
	
2025-08-28 23:58:18.185 pulp-prod_pulp-worker-577f8cf569-7zpfw_pulp-worker
pulp [dbc55a8168984ae1aa3f49d7c567c4f1]: pulpcore.tasking._util:INFO: Dispatched scheduled task pulpcore.app.tasks.telemetry.otel_metrics as task id 0198f31e-2e33-70b6-9d20-cdaa0129625b
```

mdellweg:
But  only one worker started the task. So it seems fine.

dkliban:
@mdellweg If the behavior is correct, then we shouldn't log it as an error. This is causing our team to get notified that an error has occurred.

PR #6900, opened 2025-08-29
Fix app_lock-aquire filter query

pedro-psb:
closes: #6899

---

According to the logs:
1. $W_A$ dispatched $T_0$
2. $W_B$ aquired $T_0$ and finished it
3. $W_A$ tries to get $T_0$

If $W_A$ tried to get $T_0$ after it was finished, it must have entered the `iter_task` for-loop when $T_0$ was in an `INCOMPLETE_STATE` (waiting or finished, in this case), otherwise it wouldn't have seen it.

Because of the timings, that could only have happened if it stayed in that for-loop for at least ~20s, which is odd.

Anyway, that check should ensure it won't try to acquire a Task `app_lock` if for *some-reason* it didnt know it was already finished.

### DIFF
--- a/CHANGES/6899.bugfix
+++ b/CHANGES/6899.bugfix
@@ -1,0 +1,1 @@
+Fixed error case where worker tried to aquire the app_lock for a Task that was already finished.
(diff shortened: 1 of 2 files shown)


ggainey (approved):
In large systems, the timing window is wide enough that it's getting hit frequently, and the error-msg fills logs with what is essentially a harmless problem. This looks like a reasonable filter to use.

PR #6915, opened 2025-09-01
DO NOT MERGE

git-hyagi:
this is a draft for clamav scan

We're just doing some review here and then it will get moved to pulp-service as patch files (hard to review).

### DIFF
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -413,6 +413,10 @@ VULN_REPORT_TASK_LIMITER = 10
 # Replaces asyncio event loop with uvloop
 UVLOOP_ENABLED = False
 
+# ClamAV default config
+CLAMAV_HOST = None
+CLAMAV_PORT = None
+
 # HERE STARTS DYNACONF EXTENSION LOAD (Keep at the very bottom of settings.py)
 # Read more at https://www.dynaconf.com/django/
 
--- a/pulpcore/content/handler.py
+++ b/pulpcore/content/handler.py
@@ -16,6 +16,7 @@ from aiohttp.web_exceptions import (
     HTTPFound,
     HTTPMovedPermanently,
     HTTPNotFound,
+    HTTPPreconditionFailed,
     HTTPRequestRangeNotSatisfiable,
 )
 from yarl import URL
@@ -59,8 +60,9 @@ from pulpcore.app.util import (  # noqa: E402: module level not at top of file
 )
 
 from pulpcore.exceptions import (  # noqa: E402
-    UnsupportedDigestValidationError,
     DigestValidationError,
+    UnsupportedDigestValidationError,
+    MalwareError,
 )
 from pulpcore.metrics import artifacts_size_counter  # noqa: E402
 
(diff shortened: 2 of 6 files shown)


pulpcore/exceptions/validation.py
```diff
@@ -61,6 +61,47 @@ def __str__(self):
...
+                "ClamAV detected a virus ({virus}) in {url}."
+            )
+            return msg.format(signature=self.virus, url=self.url)
+        else:
+            msg = _(
+                "ClamAV detected a virus ({virus})"
+            )
+            return msg.format(signature=self.virus)
+
+class ClamAVConnError(ValidationError):
```
bmbouter: I don't think you need this one at all. If there's a network error with clamAV let the network stack generate the error and just propagate it up

pyproject.toml
```diff
@@ -64,6 +64,7 @@ dependencies = [
   "uuid6>=2023.5.2,<=2025.0.1",
   "whitenoise>=5.0,<6.10.0",
   "yarl>=1.9.1,<1.21",  # Seems to follow SemVer https://yarl.aio-libs.org/en/latest/contributing/release_guide/#pre-release-activities .
+  "clamav-client==0.7.1",
```
bmbouter: Consider making this a range, e.g. clamav-client>=0.7.1,<1.0"

pulpcore/download/base.py
```diff
@@ -182,6 +188,7 @@ async def finalize(self):
         self._writer = None
         self.validate_digests()
         self.validate_size()
+        self.run_virus_scan()
```
bmbouter:
```suggestion
        self.run_malware_scan()
```

pulpcore/download/file.py
```diff
@@ -49,6 +49,8 @@ async def _run(self, extra_data=None):
         Args:
             extra_data (dict): Extra data passed to the downloader.
         """
+
+        self.create_clamav_socket()
```
bmbouter: I think you want to avoid patching these and do everything in BaseDownloader.__init__

pulpcore/download/base.py
```diff
@@ -159,6 +164,7 @@ async def handle_data(self, data):
         self._ensure_writer_has_open_file()
         self._writer.write(data)
         self._record_size_and_digests_for_data(data)
+        self._send_chunk_to_clamav(data)
```
bmbouter, line 168: Try keeping this patch in place, but remove this line and L#191 to verify that pip is pulling pull through cache data correctly with all this still in BaseDownloader (but not being called chunk-by-chunk). If that works, uncomment this line (and leave L#191 commented) and see if that also delivers the data to pip correctly... Basically narrow down which of these aspects is introducing the problem.

pulpcore/app/settings.py
```diff
@@ -413,6 +413,11 @@
 # Replaces asyncio event loop with uvloop
 UVLOOP_ENABLED = False
 
+# ClamAV default config
+MALWARE_SCAN_ENABLED = False
```
bmbouter: I think only have it be label-based (so remove this one). I don't see any version where we need to turn this on system-wide.

pulpcore/content/handler.py
```diff
@@ -1311,6 +1312,10 @@ async def finalize():
         downloader.handle_data = handle_data
         original_finalize = downloader.finalize
         downloader.finalize = finalize
+        if repository:
+            scan_repo = repository.pulp_labels.get("malware_scanning", "false")
```
bmbouter, line 1322: :+1:

PR #6923, opened 2025-09-04
[PULP-571] Drop session advisory locks from task execution

mdellweg:


pulpcore/tasking/tasks.py
```diff
@@ -320,7 +314,7 @@ def cancel_task(task_id):
     # Notify the worker that might be running that task and other workers to clean up
     with connection.cursor() as cursor:
         cursor.execute("SELECT pg_notify('pulp_worker_cancel', %s)", (str(task.pk),))
-        cursor.execute("NOTIFY pulp_worker_wakeup")
+        wakeup_worker(TASK_WAKEUP_HANDLE)
```
pedro-psb: Well, this was there before, but why do we need to wake workers here?
I would argue line 317 should be removed.

pulpcore/tasking/worker.py
```diff
@@ -78,7 +82,7 @@ def __init__(self, auxiliary=False):
                 name=self.name, app_type="worker", versions=self.versions
             )
         except IntegrityError:
-            _logger.error(f"A worker with name {self.name} already exists in the database.")
+            _logger.error(f"Failed to create worker {self.name} in the database.")
```
pedro-psb: This can not longer be the same name because it's not a constraint anymore, so it would be nice for debugging to output what the integrity error is all about.

pulpcore/tasking/tasks.py
```diff
@@ -222,67 +222,61 @@ def dispatch(
...
+        elif deferred:
+            # Resources are blocked. Let the others handle it.
+            task.app_lock = None
+            task.save()
+        else:
+            task.set_canceling()
+            task.set_canceled(TASK_STATES.CANCELED, "Resources temporarily unavailable.")
+    else:
+        notify_workers = True
     if notify_workers:
```
pedro-psb, line 280: :clap:

pulpcore/tasking/worker.py
```diff
@@ -533,21 +480,79 @@ def supervise_task(self, task):
...
+        qs = Task.objects.raw(
+            query,
+            [
+                self.app_status.pulp_id,
+                list(TASK_INCOMPLETE_STATES),
+                self.ignored_task_ids,
+            ],
+        )
+        return next(iter(qs), None)
+
```
pedro-psb, line 517: Its readable. I was not aware of this update-select with returning, this sounds ideal to for this use case.
And I like how you moved some of the handling logic from iter tasks and left only the fetch logic here.

pulpcore/tasking/tasks.py
mdellweg: It's in the comment: "other workers to clean up"
But you are also not wrong, we could just let the task hang around for a little longer in case the worker has gone missing.

pedro-psb: Oh right, if the worker is missing. I think its fine to leave it how it is.

pulpcore/tasking/worker.py
mdellweg: Hmm. I don't even see a way to generate an integrity error at all.
So anything that _can_ happen is exceptional enough to just fall down the exception handling funnel already there.

pulpcore/tasking/worker.py
pedro-psb, line 1: Makes sense.

pulpcore/tasking/tasks.py
mdellweg: Actually, if all workers are sleeping, this is what helps getting the cancel carried out.

pedro-psb:
The so awaited PR, looks great! Big thanks.
Been told @ggainey is also having a look. I'll leave for him to click buttons once he's satisfied.

pulpcore/tasking/tasks.py
pedro-psb: There are some different cases. I'll list here to help myself.
1) Task was picked by worker and is worker is running it - cancel signal is enough
2) Task was picked by worker and worker is missing - wake up helps faster cleanup
3) Task was not picked by a worker and all workers are sleeping - wake up helps faster cleanup
4) Task was not picked by a worker and not all workers are sleeping - would prob get fast cleaned up independently

PR #6927, opened 2025-09-07
[PULP-747] Made pulp-import-export work in the presence of domains.

ggainey:


pulpcore/plugin/importexport.py
```diff
@@ -69,6 +69,17 @@ class BaseContentResource(QueryModelResource):
...
+        """
+        Sets pulp_domain/_pulp_domain to the current-domain on import.
+        Args:
+            row (tablib.Dataset row): incoming import-row representing a single Variant.
+            kwargs: args passed along from the import() call.
+        """
+        # Content always belongs to a domain - and on-import, it's "the current one"
+        row["_pulp_domain"] = get_domain_pk()
+        row["pulp_domain"] = get_domain_pk()
+
```
mdellweg: This is basically a copy of the inherited function that skips "if"s.

Is this intended as an optimization for speed?

mdellweg: Also this may not play nicely with plugins that aren't domain compatible yet.

pulp_file/tests/functional/api/test_pulp_export.py
```diff
@@ -46,9 +54,9 @@ def _pulp_exporter_factory(repositories=None):
 
 @pytest.fixture
 def pulp_export_factory(pulpcore_bindings, monitor_task):
-    def _pulp_export_factory(exporter, body=None):
+    def _pulp_export_factory(exporter, body={}):
         task = monitor_task(
-            pulpcore_bindings.ExportersPulpExportsApi.create(exporter.pulp_href, body or {}).task
+            pulpcore_bindings.ExportersPulpExportsApi.create(exporter.pulp_href, body).task
```
mdellweg: Please don't do that.

pulpcore/app/serializers/exporter.py
```diff
@@ -308,7 +308,7 @@ def validate(self, data):
             raise serializers.ValidationError(
                 _("publication or repository_version must either be supplied but not both.")
             )
-        return data
+        return super().validate(data)
```
mdellweg, line 311: I think the common pattern here is to call `data = super().validate(date)` as the first thing of the method.

pulpcore/app/serializers/exporter.py
```diff
@@ -251,7 +251,7 @@ class Meta:
         )
 
 
-class PulpExporterSerializer(ExporterSerializer):
+class PulpExporterSerializer(ExporterSerializer):  # , ValidateFieldsMixin):
```
mdellweg: As a comment this really is unnecessary...
Would we want it or not? Is it already in the hierarchy?

pulpcore/app/modelresource.py
```diff
@@ -31,6 +38,24 @@ def before_import_row(self, row, **kwargs):
             kwargs: args passed along from the import() call.
 
         """
+        # IF we're domain-enabled:
+        #   REPLACE "their" domain-id with ours, if there is one.
+        #   If not, INSERT "our" domain-id into the path in the right place.
+        # Otherwise:
+        #   REMOVE their domain-id if there is one.
+        # Do this before letting QMR run, since it will replace "their" domain-id with "ours"
```
mdellweg, line 46: Could we remove the domain on export if it exists and then _always_ expect there is none on import?

(I think this will not only produce smaller files, but also produce a bit more consistent results.)

pulp_file/tests/functional/api/test_pulp_export.py
ggainey: Don't do which? And can you explain why?

pulpcore/app/modelresource.py
ggainey: That just splits the work into "half on export, half on import" - one ends up doing the same amount of work.  And "smaller" by removing domain-ids, when we're export *artifacts*, is lost in the noise.

pulpcore/app/serializers/exporter.py
ggainey: That's not a comment, that's a vestige of an experiment that didn't get cleaned up all the way. Nice catch :)

pulpcore/app/serializers/exporter.py
ggainey: Whereas I prefer "if 'this' serializer knows of a failure-case that doesn't require further validation, **do that first**. There is no point in validating the rest of the incoming data, if we're just going to fail due to "you provided both where only one is allowed", and *we don't care* what else might be valid. Calling super().validate() is a waste of cycles.

pulp_file/tests/functional/api/test_pulp_export.py
mdellweg: Both lines should not be changed.
In general default values for functions should not be mutable objects.

See the example with the penguin in this paragraph:
https://docs.python.org/3.13/reference/compound_stmts.html#function-definitions

pulpcore/app/modelresource.py
mdellweg: This is certainly not a deal breaker.
I just thought information that is not even in the export can never create confusion on import.
And it would be "more consistent" if non-domain-enabled and domain-enabled systems produced the same format.

mdellweg (approved):

PR #6930, opened 2025-09-09
Bump django to django==5.2

pedro-psb:


pyproject.toml
```diff
@@ -32,9 +32,9 @@ dependencies = [
   "backoff>=2.1.2,<2.3",  # Looks like only bugfixes in z-Stream.
   "click>=8.1.0,<8.3",  # Uses milestone.feature.fix https://palletsprojects.com/versions .
   "cryptography>=44.0.3,<46.0",  # SemVer compatible https://cryptography.io/en/latest/api-stability/#versioning .
-  "Django~=4.2.0",  # LTS version, switch only if we have a compelling reason to".
-  "django-filter>=23.1,<=25.1",  # Looks like CalVer.
-  "django-guid>=3.3.0,<3.6",  # Looks like only bugfixes in z-Stream.
+  "Django~=5.2.0",  # LTS version, switch only if we have a compelling reason to".
+  "django-filter>=25.1,<26",  # Looks like CalVer.
```
mdellweg: Please don't change the version requirements as if they were following semver. They are not...

pyproject.toml
```diff
@@ -33,8 +33,8 @@ dependencies = [
   "click>=8.1.0,<8.3",  # Uses milestone.feature.fix https://palletsprojects.com/versions .
   "cryptography>=44.0.3,<46.0",  # SemVer compatible https://cryptography.io/en/latest/api-stability/#versioning .
   "Django~=4.2.0",  # LTS version, switch only if we have a compelling reason to".
-  "django-filter>=23.1,<=25.1",  # Looks like CalVer.
-  "django-guid>=3.3.0,<3.6",  # Looks like only bugfixes in z-Stream.
+  "django-filter~=25.1",  # Self claimed stable CalVer https://github.com/carltongibson/django-filter
```
mdellweg: "Other breaking changes are rare." What I hear is: "We make no guarantees about changes tied to specific types of version changes." But also "You can probably expect 1 or 2 releases per year." (not like we need to touch this line every other week.)

I would argue we are better off to keep the "<=" upper bound.

Having said that, i think what we agreed to try now is: Not touching any LB here, and still allowing to up the django version UB to 5.2.

And now a very wild idea: Given that these changes are really minimal, how do you feel about backporting this version bump?

pedro-psb: > Having said that, i think what we agreed to try now is: Not touching any LB here, and still allowing to up the django version UB to 5.2.

I see, that was not clear to me, but it sounds fine.

> And now a very wild idea: Given that these changes are really minimal, how do you feel about backporting this version bump?

We can try and see what happens. You mean all of it? The django 5 upperbound, the shell command override, and the other bumps?

mdellweg: Basically if we can make upgrading Django less of a hassle for everybody, we all win...
For that I'd say this PR should only bump upper bounds of requirements (Probably really only Django) and contain this other two small adjustments you made (that seem to be compatible with Django4 anyway.)

And then we can think about backporting it.

pyproject.toml
```diff
@@ -32,8 +32,8 @@ dependencies = [
   "backoff>=2.1.2,<2.3",  # Looks like only bugfixes in z-Stream.
   "click>=8.1.0,<8.3",  # Uses milestone.feature.fix https://palletsprojects.com/versions .
   "cryptography>=44.0.3,<46.0",  # SemVer compatible https://cryptography.io/en/latest/api-stability/#versioning .
-  "Django~=4.2.0",  # LTS version, switch only if we have a compelling reason to".
-  "django-filter>=23.1,<=25.1",  # Looks like CalVer.
+  "Django>=4.2.0,<=5.2.0",  # LTS version, switch only if we have a compelling reason to".
+  "django-filter>=23.1,<=25.1",  # Uses CalVer, not released often https://github.com/carltongibson/django-filter
```
mdellweg: That's perfect. Thank you!
It reminds me that we still haven't looked up all dependencies versioning philosophies.

pyproject.toml
```diff
@@ -32,8 +32,8 @@ dependencies = [
   "backoff>=2.1.2,<2.3",  # Looks like only bugfixes in z-Stream.
   "click>=8.1.0,<8.3",  # Uses milestone.feature.fix https://palletsprojects.com/versions .
   "cryptography>=44.0.3,<46.0",  # SemVer compatible https://cryptography.io/en/latest/api-stability/#versioning .
-  "Django~=4.2.0",  # LTS version, switch only if we have a compelling reason to".
-  "django-filter>=23.1,<=25.1",  # Looks like CalVer.
+  "Django>=4.2.0,<=5.2.0",  # LTS version, switch only if we have a compelling reason to".
```
mdellweg: The current release is 5.2.6. But I think with Django we can stick to x.y giving us essentially:


`http https://pypi.org/pypi/Django/json | jq '.releases|keys'`
```suggestion
  "Django>=4.2.0,<5.3",  # LTS version, switch only if we have a compelling reason to".
```

pedro-psb:
Lowerbound fails under its set of dependencies. I'm trying to isolate what exactly locally.
And just a note, probably its known already: our unit tests are a bit fragile in terms of leaking. If we use [`pytest-random-order`](https://github.com/pytest-dev/pytest-random-order) we'll get different set of failures at each run.

pyproject.toml
```diff
@@ -32,8 +32,8 @@ dependencies = [
   "backoff>=2.1.2,<2.3",  # Looks like only bugfixes in z-Stream.
   "click>=8.1.0,<8.3",  # Uses milestone.feature.fix https://palletsprojects.com/versions .
   "cryptography>=44.0.3,<46.0",  # SemVer compatible https://cryptography.io/en/latest/api-stability/#versioning .
-  "Django~=4.2.0",  # LTS version, switch only if we have a compelling reason to".
-  "django-filter>=23.1,<=25.1",  # Looks like CalVer.
+  "Django>=4.2,<5.3",  # LTS version, switch only if we have a compelling reason to".
```
mdellweg: Can you try this:
```suggestion
  "Django>=4.2.24,<5.3",  # LTS version, switch only if we have a compelling reason to".
```

pedro-psb:
That was intriguing... Thanks for the suggestion!

PR #6936, opened 2025-09-11
Cleanup missing worker process records faster

mdellweg:
As long as we do not clean up these records of missing or improperly shutdown components, we cannot free the recources held by their, at this point supposedly failed tasks.

### DIFF
--- a/pulpcore/tasking/worker.py
+++ b/pulpcore/tasking/worker.py
@@ -142,7 +142,7 @@ class PulpcoreWorker:
             elif notification.payload == TASK_WAKEUP_HANDLE:
                 self.wakeup_handle = True
             else:
-                _logger.warn("Unknown wakeup call recieved. Reason: '%s'", notification.payload)
+                _logger.warning("Unknown wakeup call recieved. Reason: '%s'", notification.payload)
                 # We cannot be sure so assume everything happened.
                 self.wakeup_unblock = not self.auxiliary
                 self.wakeup_handle = True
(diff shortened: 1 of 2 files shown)


pulpcore/tasking/worker.py
```diff
@@ -188,9 +188,11 @@ def cleanup_ignored_tasks(self):
             self.ignored_task_ids.remove(pk)
 
     def worker_cleanup(self):
-        qs = AppStatus.objects.older_than(age=timedelta(days=7))
+        qs = AppStatus.objects.missing()
```
ggainey, line 191: :+1:

pulpcore/tasking/worker.py
```diff
@@ -319,8 +321,6 @@ def _unblock_tasks(self):
                     )
                     task.unblock()
                     count += 1
-                # Don't consider this task's resources as held.
-                continue
```
ggainey, line 323: So the idea is, since we're now dealing with *missing* workers immediately, if it's CANCELING, its resources will be found/cleaned up at CANCELED time - so it is *always* ok for its resources to be kept as "taken", where before that might-not have been the case?

mdellweg: This might have been a bit dangerous to begin with. While `"cancelling"` the task may still be executed by a worker just realizing it is to be canceled now. It could also not have been picked up yet. In any case it's safer to wait until the task is actually `"canceled"` before touching any of its resources.

ggainey: Right, that matches my expectation, thanks.

PR #6937, opened 2025-09-11
[PULP-553] Enable immediate task execution on content app (1/2)

pedro-psb:
Plan:
- [x] Refactor sync dispatch and execute
- [x]  Add async dispatch and execute task functions
- [x]  Add async pull_through_add_content and unit test it
~- [ ] Change content app to use async pull_through_add_content and remove sync one (only used here)~ (left to another PR)

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -381,6 +381,21 @@ class Repository(MasterModel):
         body = {"repository_pk": self.pk, "add_content_units": [cpk], "remove_content_units": []}
         return dispatch(add_and_remove, kwargs=body, exclusive_resources=[self], immediate=True)
 
+    async def async_pull_through_add_content(self, content_artifact):
+        cpk = content_artifact.content_id
+        already_present = RepositoryContent.objects.filter(
+            content__pk=cpk, repository=self, version_removed__isnull=True
+        )
+        if not cpk or await already_present.aexists():
+            return None
+
+        from pulpcore.plugin.tasking import adispatch, aadd_and_remove
+
+        body = {"repository_pk": self.pk, "add_content_units": [cpk], "remove_content_units": []}
+        return await adispatch(
+            aadd_and_remove, kwargs=body, exclusive_resources=[self], immediate=True
+        )
+
     @hook(AFTER_UPDATE, when="retain_repo_versions", has_changed=True)
     def _cleanup_old_versions_hook(self):
         # Do not attempt to clean up anything, while there is a transaction involving repo versions
--- a/pulpcore/app/models/task.py
+++ b/pulpcore/app/models/task.py
@@ -341,8 +341,20 @@ class Task(BaseModel, AutoAddObjPermsMixin):
             self.unblocked_at = unblocked_at
         else:
             self.refresh_from_db()
-            raise RuntimeError(
-                _("Falied to set task {} unblocked in state '{}'.").format(self.pk, self.state)
+            raise RuntimeError("Failed to set task {} unblocked in state '{}'.").format(
+                self.pk, self.state
+            )
+
+    async def aunblock(self):
+        # This should be safe to be called without holding the lock.
+        unblocked_at = timezone.now()
+        rows = await Task.objects.filter(pk=self.pk).aupdate(unblocked_at=unblocked_at)
+        if rows == 1:
+            self.unblocked_at = unblocked_at
+        else:
+            await self.arefresh_from_db()
+            raise RuntimeError("Failed to set task {} unblocked in state '{}'.").format(
+                self.pk, self.state
             )
 
     class Meta:
--- a/pulpcore/app/tasks/repository.py
+++ b/pulpcore/app/tasks/repository.py
@@ -236,3 +236,30 @@ def add_and_remove(repository_pk, add_content_units, remove_content_units, base_
     with repository.new_version(base_version=base_version) as new_version:
         new_version.remove_content(models.Content.objects.filter(pk__in=remove_content_units))
         new_version.add_content(models.Content.objects.filter(pk__in=add_content_units))
+
+
+async def aadd_and_remove(
+    repository_pk, add_content_units, remove_content_units, base_version_pk=None
+):
+    """Aynsc version of add_and_remove."""
+    repository = await models.Repository.objects.aget(pk=repository_pk)
+    repository = await repository.acast()
+
+    if base_version_pk:
+        base_version = await models.RepositoryVersion.objects.aget(pk=base_version_pk)
+    else:
+        base_version = None
+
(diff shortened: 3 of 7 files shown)


pulpcore/app/models/task.py
```diff
@@ -345,6 +345,18 @@ def unblock(self):
...
             )
 
+    async def aunblock(self):
+        # This should be safe to be called without holding the lock.
+        unblocked_at = timezone.now()
+        rows = await Task.objects.filter(pk=self.pk).aupdate(unblocked_at=unblocked_at)
+        if rows == 1:
+            self.unblocked_at = unblocked_at
+        else:
+            self.arefresh_from_db()
```
mdellweg:
```suggestion
            await self.arefresh_from_db()
```

pulpcore/app/models/task.py
```diff
@@ -345,6 +345,18 @@ def unblock(self):
...
+            raise RuntimeError(
+                _("Falied to set task {} unblocked in state '{}'.").format(self.pk, self.state)
```
mdellweg:
```suggestion
                _("Failed to set task {} unblocked in state '{}'.").format(self.pk, self.state)
```
Typo by Copy'nPaste ...

pulpcore/app/tasks/repository.py
```diff
@@ -236,3 +236,42 @@ def add_and_remove(repository_pk, add_content_units, remove_content_units, base_
     with repository.new_version(base_version=base_version) as new_version:
         new_version.remove_content(models.Content.objects.filter(pk__in=remove_content_units))
         new_version.add_content(models.Content.objects.filter(pk__in=add_content_units))
+
+
+async def async_add_and_remove(
```
mdellweg: This function is always run as a task (at least it really should be). So let's not duplicate it. Let's just make it async for real.

mdellweg: And here's another thing to watch out for: Both content creation from an artifact as well as creating a new repository version may involve an amount of calculation that is out of our control. Depending on the plugin, this may run into timeouts.

pulpcore/app/models/task.py
```diff
@@ -354,7 +354,7 @@ async def aunblock(self):
         else:
             self.arefresh_from_db()
             raise RuntimeError(
-                _("Falied to set task {} unblocked in state '{}'.").format(self.pk, self.state)
+                _("Failed to set task {} unblocked in state '{}'.").format(self.pk, self.state)
```
mdellweg: Same thing on L345.

mdellweg: Also I don't think it's wise to translate error messages.

pedro-psb: True, I remember a wider team discussion where we agreed on that.

pulpcore/app/tasks/repository.py
pedro-psb: Nice observation. If that starts to happen often we should probably not call it as immediate.

mdellweg: The problematic thing is there is no way to _know_. And once we timed out the task there is no recovering.

pedro-psb: > This function is always run as a task (at least it really should be). So let's not duplicate it. Let's just make it async for real.

In ansible it's called directly (inside a task, but not by means of dispatch). https://github.com/pulp/pulp_ansible/blob/37d1ce28ce2e3743b82e4d5b5b4b5461896e7935/pulp_ansible/app/tasks/mark.py#L22

pedro-psb: > The problematic thing is there is no way to know. And once we timed out the task there is no recovering.

If I remember correctly, that is the case only for the content app, right? It's handling content serving requests, so it can't report this kind of side-effect failure.

What if we dispatch to workers in case of timeout failure? :eyes:

pulpcore/app/models/repository.py
pedro-psb, line 1: `async_pull_through_add_content` can replace `pull_through_add_content`. This one is only used in pulpcore.
I will do that on the last step, which is adding this to the content app handler.

pulpcore/app/tasks/repository.py
pedro-psb, line 1: Renamed this after https://github.com/pulp/pulp_container/blob/0c7863d7bcb1197255cd3cde889305e9581d8cc8/pulp_container/app/tasks/download_image_data.py#L19

pulpcore/tests/unit/content/test_handler.py
```diff
@@ -552,3 +555,25 @@ async def test_pull_through_repository_add(request123, monkeypatch):
...
         await remote.adelete()
         await distro.adelete()
+
+
+@pytest.mark.asyncio
+@pytest.mark.django_db
+async def test_async_pull_through_add(ca1, monkeypatch):
+    # setup dispatch context
+    set_guid(uuid.uuid4())
+    app_status = await AppStatus.objects.acreate(app_type="content", name="test_runner")
```
mdellweg, line 564: This is a bit dangerous. I think we need to mock out the registered app status variable so it gets reset after the test.

pedro-psb: I dont understand what you mean that is dangerous. The AppStatus is required to exist in the test database because task creation needs a valid reference to it on the db level. And it's being cleaned up.

mdellweg: AppStatus.objects keeps a record of this when creating, so it can implement the `current()`.
If you'd write a second such test, it would fail.

And if we don't address this now, we would probably have a really hard time understanding what's wrong later.

pedro-psb: Oh right, its kept in the `_current_app_status`.
Can I implement a delete override in the AppStatus.objects to clear that state if the deleted object matches the current app for the process?

pulpcore/tasking/tasks.py
```diff
@@ -66,77 +68,118 @@ def _execute_task(task):
...
-        )
+        log_task_completed(task, domain)
         send_task_notification(task)
+        return result
+    return None
+
+
+async def aexecute_task(task):
+    # Store the task id in the context for `Task.current()`.
+    current_task.set(task)
```
mdellweg, line 94: This is setting a context var. And I think you are missing the separation from the calling context (see `_execute_task`).

(We should remember to pull all the context vars setting into here... but not this PR.)

pulpcore/tests/unit/content/test_handler.py
mdellweg: I think this should work:
https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/unit/models/test_task.py#L16

mdellweg (approved):

PR #6948, opened 2025-09-15
Mark all_tasks_dispatched as deprecated

mdellweg:


pulpcore/app/serializers/task.py
pedro-psb, line 1: Can you add the spectacular [mark](https://drf-spectacular.readthedocs.io/en/latest/drf_spectacular.html#drf_spectacular.utils.extend_schema_serializer) for deprecation?

```python
@extend_schema_serializer(deprecate_fields=["all_tasks_dispatched"])
class TaskGroupSerializer(ModelSerializer):
    ...
```

pedro-psb (changes requested):

pulpcore/app/serializers/task.py
mdellweg: That is what I was looking for.

pedro-psb (approved):

Issue #6897, opened 2025-08-25
For remote update operations, only dispatch a task when an actual change is needed


dkliban:
The API should only dispatch a task to update a remote if an actual change needs to be made. Otherwise it should just return a 200 OK response to the user.

PR #6953, opened 2025-09-17
Update Remote API to skip task dispatch if no changes are needed

git-hyagi:
closes: #6897

### DIFF
--- a/pulpcore/app/util.py
+++ b/pulpcore/app/util.py
@@ -678,3 +678,16 @@ def normalize_http_status(status):
         return "5xx"
     else:
         return ""
+
+
+def resource_modified(resource, request, **kwargs):
+    """Check if the request data has values different from the current resource instance."""
+    partial = kwargs.pop("partial", False)
+    entity = resource.get_object()
+    serializer = resource.get_serializer(entity, data=request.data, partial=partial)
+    serializer.is_valid(raise_exception=True)
+
+    for request_field, request_field_value in serializer.validated_data.items():
+        if getattr(entity, request_field) != request_field_value:
+            return True
+    return False
--- a/pulpcore/app/viewsets/domain.py
+++ b/pulpcore/app/viewsets/domain.py
@@ -14,6 +14,7 @@ from pulpcore.app.serializers import (
     AsyncOperationResponseSerializer,
 )
 from pulpcore.app.tasks import migrate_backend
+from pulpcore.app.util import resource_modified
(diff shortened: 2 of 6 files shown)


pulpcore/app/viewsets/repository.py
```diff
@@ -368,6 +368,21 @@ class RemoteViewSet(
...
     filterset_class = RemoteFilter
 
+    @extend_schema(
+        description="Trigger an asynchronous update task",
+        responses={202: AsyncOperationResponseSerializer, 200: None},
+    )
+    def update(self, request, pk, **kwargs):
+        """
+        Check if the update operation is going to change anything and if:
+        - no actual change is needed return a 200 OK
```
decko: What if it returns a 304 Not Modified? https://developer.mozilla.org/pt-BR/docs/Web/HTTP/Reference/Status/304

ggainey: I like "not modified" better, good thought

decko: @ggainey, @git-hyagi gave the suggestion about returning a `204 No Content` https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/204

What do you think?

ggainey: "Not Modified" *sounds* good, but reviewing the spec shows it's *definitely* not appropriate in this case. Looking thru the actual spec, I think 204 is in fact the response that most-fits this case - "server has successfully fulfilled the request and that there is no additional content to send in the response"

pulpcore/app/util.py
```diff
@@ -678,3 +678,16 @@ def normalize_http_status(status):
         return "5xx"
     else:
         return ""
+
+
+def resource_modified(resource, request, **kwargs):
+    """Check if the request data has values different from the current resource instance."""
+    partial = kwargs.pop("partial", False)
+    domain = resource.get_object()
```
ggainey: This is a general utility - not just for Domains. Call this "entity" or "model" or something, to make it clear you're *not* expecting a Domain here, please.

pulpcore/app/viewsets/domain.py
```diff
@@ -109,7 +110,7 @@ class DomainViewSet(
 
     @extend_schema(
         description="Trigger an asynchronous update task",
-        responses={202: AsyncOperationResponseSerializer, 200: None},
+        responses={202: AsyncOperationResponseSerializer, 204: None},
```
ggainey, line 113: :+1:

ggainey (approved):

PR #6962, opened 2025-09-19
Enable immediate task from content app (2/2)

pedro-psb:
When requiring immediate tasks to be async, we had a problem dispatching them from the content app due to the use of django transactions in the dispatch code and how that plays with asyncio.

Now that we no longer have that (due to the app_lock implementation) we can remove the workaround we had in place and prioritize immediate tasks on the worker.

### DIFF
--- a/pulpcore/tasking/tasks.py
+++ b/pulpcore/tasking/tasks.py
@@ -6,7 +6,6 @@ import os
 import sys
 import traceback
 import tempfile
-import threading
 from gettext import gettext as _
 from contextlib import contextmanager
 from asgiref.sync import sync_to_async, async_to_sync
@@ -187,14 +186,6 @@ def get_task_function(task, ensure_coroutine=False):
         return async_to_sync(task_wrapper)
 
 
-def running_from_thread_pool() -> bool:
-    # TODO: this needs an alternative approach ASAP!
-    # Currently we rely on the weak fact that ThreadPoolExecutor names threads like:
-    # "ThreadPoolExecutor-0_0"
-    thread_name = threading.current_thread().name
-    return "ThreadPoolExecutor" in thread_name
-
-
 def dispatch(
     func,
     args=None,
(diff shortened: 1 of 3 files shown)


pulpcore/tasking/tasks.py
```diff
@@ -363,6 +355,13 @@ def are_resources_available(colliding_resources, task: Task) -> bool:
     return not colliding_resources or not colliding_resources_taken
 
 
+def called_from_content_app() -> bool:
+    current_app = AppStatus.objects.current()
+    if current_app and current_app.type == "content":
+        return True
+    return False
```
mdellweg, line 362: Nice!!!

Why not just return the expression you are branching on?
```suggestion
def called_from_content_app() -> bool:
    current_app = AppStatus.objects.current()
    return current_app is not None and current_app.type == "content"
```

pulpcore/tasking/tasks.py
```diff
@@ -363,6 +354,11 @@ def are_resources_available(colliding_resources, task: Task) -> bool:
     return not colliding_resources or not colliding_resources_taken
 
 
+def called_from_content_app() -> bool:
+    current_app = AppStatus.objects.current()
+    return current_app is not None and current_app.type == "content"
```
mdellweg, line 359:
```suggestion
    return current_app is not None and current_app.app_type == "content"
```

pulpcore/tests/unit/content/test_handler.py
```diff
@@ -561,7 +561,7 @@ async def test_pull_through_repository_add(request123, monkeypatch):
 @pytest_asyncio.fixture
 async def app_status(monkeypatch):
     monkeypatch.setattr(AppStatus.objects, "_current_app_status", None)
-    app_status = await AppStatus.objects.acreate(app_type="content", name="test_runner")
+    app_status = await AppStatus.objects.acreate(app_type="api", name="test_runner")
```
mdellweg, line 564: Dang!

mdellweg (approved):

PR #6963, opened 2025-09-19
[PULP-785] Remove advisory lock from janitorial work

mdellweg:
All the corresponding cleanup work can equally well be organized using advisory locks within a transaction.

### DIFF
--- a/pulpcore/constants.py
+++ b/pulpcore/constants.py
@@ -6,11 +6,10 @@ from types import SimpleNamespace
 # The group will be 0.
 # The numbers are randomly chosen.
 # !!! Never change these values !!!
-TASK_DISPATCH_LOCK = 21
 TASK_SCHEDULING_LOCK = 42
 TASK_UNBLOCKING_LOCK = 84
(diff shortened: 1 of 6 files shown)


pulpcore/constants.py
```diff
@@ -6,11 +6,10 @@
 # The group will be 0.
 # The numbers are randomly chosen.
 # !!! Never change these values !!!
-TASK_DISPATCH_LOCK = 21
 TASK_SCHEDULING_LOCK = 42
 TASK_UNBLOCKING_LOCK = 84
-TASK_METRICS_HEARTBEAT_LOCK = 74
-STORAGE_METRICS_LOCK = 72
```
pedro-psb, line 13: Ok, I see `STORAGE_METRICS_LOCK` was not being used anymore.

pedro-psb: And the dispatch one is gone for at least a couple of releases...

mdellweg: Yes.

PR #6969, opened 2025-09-22
Revert "Add changes required by django 5"

pedro-psb:
This is a breaking change, as it may break users who have a custom installation. For example, one can update pulpcore to the version that can use Django 5, but some of the plugins are in versions that doesn't fully support it yet.

This reverts commit 70c670b03a1a5ae7c753e321c534bb569b449a67.


pedro-psb:
Maybe we could leave the change themselves (the import string and shell command) and only revert the bump.
The changes are pretty simple and looks compatible with django 4. Opinions?

PR #6973, opened 2025-09-23
[PULP-817] Adjustments to make imp/exp work with django-imp-exp 4.x.

### LINKED ISSUE #5324: django-import-export requirement from to <4.1.0 or higher.

ggainey:
closes #5324.

### DIFF
--- a/pulpcore/app/modelresource.py
+++ b/pulpcore/app/modelresource.py
@@ -62,8 +62,8 @@ class ArtifactResource(QueryModelResource):
         # the export converts None to blank strings but sha384 and sha512 have unique constraints
         # that get triggered if they are blank. convert checksums back into None if they are blank.
         for checksum in ALL_KNOWN_CONTENT_CHECKSUMS:
-            if row[checksum] == "":
-                row[checksum] = None
+            if row[checksum] == "" or checksum not in settings.ALLOWED_CONTENT_CHECKSUMS:
+                del row[checksum]
 
     def set_up_queryset(self):
         """
@@ -109,7 +109,7 @@ class ArtifactDomainForeignKeyWidget(ForeignKeyWidget):
         return qs
 
     def render(self, value, obj=None, **kwargs):
-        return value.sha256
+        return value.sha256 if value else ""
 
 
 class ContentArtifactResource(QueryModelResource):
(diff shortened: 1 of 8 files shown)


pulpcore/app/modelresource.py
```diff
@@ -62,8 +62,8 @@ def before_import_row(self, row, **kwargs):
         # the export converts None to blank strings but sha384 and sha512 have unique constraints
         # that get triggered if they are blank. convert checksums back into None if they are blank.
         for checksum in ALL_KNOWN_CONTENT_CHECKSUMS:
-            if row[checksum] == "":
-                row[checksum] = None
+            if row[checksum] == "" or checksum not in settings.ALLOWED_CONTENT_CHECKSUMS:
+                del row[checksum]
```
mdellweg, line 66: Random unrelated question: Do we recalculate missing checksums on import? Or do both systems need the same setting there?

ggainey: Fine question, and the answers are "no we don't" and "...huh." I expect that, "in the wild", systems being exported-to have the same settings as the system being exported-from, so it just never came up. But it's a de-facto standard, not a de-jure one :)

pyproject.toml
```diff
@@ -59,7 +59,7 @@ dependencies = [
   "python-gnupg>=0.5.0,<0.6",  # Looks like only bugfixes in z-Stream [changelog only in git]
   "PyYAML>=5.1.1,<6.1",  # Looks like only bugfixes in z-Stream.
   "redis>=4.3.0,<7.2",  # Looks like only bugfixes in z-Stream.
-  "tablib>=3.5.0,<3.6",  # 3.6.0 breaks with import export. Not sure about semver.
+  "tablib>=3.5.0,<4.0",  # 3.6.0 breaks with import export. Not sure about semver.
```
pedro-psb: Hmm, I was thinking about the middle versions...
I think it doesn't hurt to put `3.5.0,<4.0,!=3.6`, since we know its broken. Just in case.
And a similar concern: assuming we want to to stick to LTS Django version, the range `4.2.24,<5.3` technically includes `5.0.x` and `5.1.x`. We could be more specific and add `!=5.0,!=5.1`.

PR #6976, opened 2025-09-25
Always call wakeup_worker in syncronous context

mdellweg:


pedro-psb (approved):
Right, django didn't complained because we are calling psycopg directly on that function.

PR #6977, opened 2025-09-25
Run orphan cleanup periodically

mdellweg:


pedro-psb (approved):
That does "change behavior", but not in a way that breaks anyone.
Also I've seen some discussion in matrix with a certain agreement on that approach.

PR #6984, opened 2025-09-26
Add a premigration hook to clean stale Status records

pedro-psb:


pulpcore/app/apps.py
```diff
@@ -271,6 +272,35 @@ def ready(self):
...
+
+    app_ttl_map = [
+        ("ApiAppStatus", settings.API_APP_TTL),
+        ("ContentAppStatus", settings.CONTENT_APP_TTL),
+        ("Worker", settings.WORKER_TTL),
+    ]
+
+    for app_name, app_ttl in app_ttl_map:
+        try:
+            app_cls = apps.get_model("core", "ContentAppStatus")
```
mdellweg, line 289: @pedro-psb Should this not be `app_name` here?

Also do you know if we need this on 3.85 too?

pedro-psb: > @pedro-psb Should this not be app_name here?

YES 🫨

pedro-psb: > Also do you know if we need this on 3.85 too?

[We have the heartbeat failure as catastrophic in 3.85](https://pulpproject.org/pulpcore/changes/#3.85.0), so it would be nice.
Even though a clean shutdown of workers should take care of this, the less room for upgrade failures we have the best.

Issue #6993, opened 2025-09-30
"AttributeError: This QueryDict instance is immutable" error when uploading package


git-hyagi:
**Describe the bug**
Trying to upload a package using synchronous upload API and cancelling it in the middle of the upload will raise the exception:
```
pulp_1   | pulp [ffe1534d62f44a47a99af1bb898bff2b]: django.request:ERROR: Internal Server Error: /pulp/default/api/v3/content/rpm/packages/upload/                                                                                                                                                                            
pulp_1   | Traceback (most recent call last):                                                                                                                                                                                                                                                                                 
pulp_1   |   File "/usr/local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner                                                                                                                                                                                                              
pulp_1   |     response = get_response(request)                                                                                                                                                                                                                                                                               
pulp_1   |                ^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                               
pulp_1   |   File "/usr/local/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response                                                                                                                                                                                                          
pulp_1   |     response = wrapped_callback(request, *callback_args, **callback_kwargs)                                                                                                                                                                                                                                        
pulp_1   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                        
pulp_1   |   File "/usr/local/lib/python3.11/site-packages/sentry_sdk/integrations/django/views.py", line 94, in sentry_wrapped_callback
pulp_1   |     return callback(request, *args, **kwargs)
pulp_1   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pulp_1   |   File "/usr/local/lib/python3.11/site-packages/django/views/decorators/csrf.py", line 56, in wrapper_view
pulp_1   |     return view_func(*args, **kwargs)
pulp_1   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^
pulp_1   |   File "/usr/local/lib/python3.11/site-packages/rest_framework/viewsets.py", line 125, in view
pulp_1   |     return self.dispatch(request, *args, **kwargs)
pulp_1   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #6994, opened 2025-10-06
Handle AttributeError in content upload label fix

git-hyagi:
closes: #6993
ref: https://issues.redhat.com/browse/PULP-841


pulpcore/plugin/serializers/content.py
```diff
@@ -188,13 +188,34 @@ def __init__(self, *args, **kwargs):
...
+        Validate the serializer data, with special handling for pulp_labels deserialization.
+
+        This method checks if pulp_labels failed to be deserialized from JSON string to dict
+        during initialization (typically due to immutable QueryDict).
+        """
+        if "pulp_labels" in data and isinstance(data["pulp_labels"], str):
+            raise ValidationError(
+                _(
+                    "Failed to deserialize pulp_labels!"
+                    "A common cause of this error is when file didn't upload or file is incomplete."
```
ggainey: Isn't a more-common error "pulp_labels needds to be JSON and the user provided bad data on the command-line"? If we're going to ascribe causes, that one def needs to be included, does it not?

ggainey (approved):
Much better user-feedback, thank you

PR #7037, opened 2025-10-21
Allow batch size of artifacts processed during sync to be configured

### LINKED ISSUE #7024: Allow batch size of artifacts processed during sync to be configured

When syncing repositories with content of noticeably huge size, storage used for temporary file locations can get easily exhausted. This is because Pulp processes artifacts in batches, and the batch size is currently hardcoded as 200 - https://github.com/pulp/pulpcore/blob/main/pulpcore/plugin/stages/artifact_stages.py#L142. Pulp downloads 200 files first, keep them in local, temporary storage, and then processes them and move them to the configured backend storage.

So, if download of 200 files (or even before that) causes the local/temporary storage to be full, the worker process will fail to perform the sync. In k8s deployments, this will mean ephemeral storage of worker pods being exhausted and the pod being evicted.

Allow the batch size to be configured. Either at the instance level or at the repo/remote level.

balasankarc:
closes #7024

Instead of hardcoding a value of 200, which can cause the storage requirements to increase drastically, allow batch size of artifacts processed during sync to be configured by users.

### DIFF
--- a/CHANGES/7024.feature
+++ b/CHANGES/7024.feature
@@ -1,0 +1,3 @@
+Add MAX_CONCURRENT_CONTENT as a global setting.
+During sync, this setting will be used to determine the size of batch of
+content being processed in one go. Default value is 200.
--- a/docs/admin/reference/settings.md
+++ b/docs/admin/reference/settings.md
@@ -424,6 +424,12 @@ Defaults to `/var/lib/pulp/tmp/`.
     Files are commonly staged in the `WORKING_DIRECTORY` and
     validated before being moved to their permanent home in `MEDIA_ROOT`.
 
+### MAX\_CONCURRENT\_CONTENT
+
+The size of the batch of content processed in one go when syncing content from
+a remote.
+
+Defaults to 200.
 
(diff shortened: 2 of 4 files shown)


dkliban:
@balasankarc Thank you for the contribution. The CI failures are not related to your changes. We will accept this change once we have our CI working. 

In the future it would better if this setting was available on each remote. Setting it globally to a low number will cause significantly more load on the database.

balasankarc:
> @balasankarc Thank you for the contribution. The CI failures are not related to your changes. We will accept this change once we have our CI working.
> 
> In the future it would better if this setting was available on each remote. Setting it globally to a low number will cause significantly more load on the database.

@dkliban I'll work on the setting being per-remote as a different PR, as next iteration. So this one can solve the immediate problem and unblock us, and the next one enhances it to be better from a performance PoV.

dralley:
FYI, CI should be fixed once you rebase onto the latest state of the main branch

balasankarc:
@dralley Done.

PR #7045, opened 2025-10-23
Fix flake8 warning

dosas:
@quba42 @hstct Could you review this or add a reviewer?


pedro-psb (changes requested):
I think I saw this warning locally.
There is no need for space around the `/`, which modifies the string.
Adding around `-` operator should be enough.

PR #7055, opened 2025-10-28
Fix trying to dispatch an immediate non-async task for pull-through caching

gerrod3:
Forgot to remove immediate=True from this task when I was fixing the content app immediate task dispatching issue. We don't allow non-async immediate tasks. So the correct behavior is: 
1. Dispatch non-async immediate task -> fail
2. Dispatch async immediate task -> good
3. Dispatch async immediate task from content app -> good, but task is not handled immediately, it is deferred to a worker

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -376,10 +376,10 @@ class Repository(MasterModel):
         if not cpk or already_present.exists():
             return None
 
-        from pulpcore.plugin.tasking import dispatch, add_and_remove
+        from pulpcore.plugin.tasking import dispatch, aadd_and_remove
 
         body = {"repository_pk": self.pk, "add_content_units": [cpk], "remove_content_units": []}
-        return dispatch(add_and_remove, kwargs=body, exclusive_resources=[self], immediate=True)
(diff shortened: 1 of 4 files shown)


pedro-psb (commented):
Can you try using `aadd_and_remove` instead?
Even though it's deferred, keeping immediate makes it a priority for workers when picking up new work.

pulpcore/app/models/repository.py
```diff
@@ -379,7 +379,7 @@ def pull_through_add_content(self, content_artifact):
         from pulpcore.plugin.tasking import dispatch, add_and_remove
 
         body = {"repository_pk": self.pk, "add_content_units": [cpk], "remove_content_units": []}
-        return dispatch(add_and_remove, kwargs=body, exclusive_resources=[self], immediate=True)
+        return dispatch(add_and_remove, kwargs=body, exclusive_resources=[self])
```
pedro-psb: I meat using `aadd_and_remove` instead of `add_and_remove`, and keeping `immediate=True`.
Immediate is a misleading name, a more adequate name would be something like `short`.
It does make a difference if the task is marked immediate/short, even though it may not be executed immediately.

In any case, using `async_pull_through_add_content` is fine.

gerrod3:
@pedro-psb Is everything good now?

PR #7060, opened 2025-10-31
[PULP-965] Run immediate tasks in worker process

mdellweg:


pulpcore/tasking/worker.py
```diff
@@ -443,6 +445,17 @@ def sleep(self):
...
+        """Call and supervise the immediate async task process.
+
+        This function must only be called while holding the lock for that task."""
+        self.task = task
+        with with_task_context(task), using_workdir():
+            execute_task(task)
+        if task.reserved_resources_record:
+            self.notify_workers(TASK_WAKEUP_UNBLOCK)
+        self.task = None
+
```
pedro-psb, line 457: `execute_task` calls `_execute_task` which has a `with_task_context` contextmanager.
Apparently it's all working, but it's wrapped in two `with_task_context`, one with the copied contextvars and others with the originals. Is this somehow intentional?

pulpcore/app/models/task.py
```diff
@@ -142,6 +144,31 @@ class Task(BaseModel, AutoAddObjPermsMixin):
...
+            or get_users_with_perms(
+                self,
+                only_with_perms_in=["core.manage_roles_task"],
+                with_group_users=False,
+                include_model_permissions=False,
+                include_domain_permissions=False,
+            ).first()
+        )
+        return user
+
```
pedro-psb, line 171: Interesting.
I see this was just moved around, but I've never read that comment before.

pedro-psb (commented):
Thanks for the yak shaving there.
Definitely helps with maintanability.

pulpcore/tasking/worker.py
mdellweg: Good call. Not this is a leftover here.

pulpcore/app/util.py
```diff
@@ -615,7 +610,8 @@ def get_domain_pk():
 
 
 def set_domain(new_domain):
-    current_domain.set(new_domain)
+    # Deprecated
+    _current_domain.set(new_domain)
     return new_domain
```
pedro-psb, line 615: Didnd't notice that before, sorry. What exactly is deprecated and why?
Is it the case for using a deprecation message and a schedule for removal?

mdellweg: It's only used internally, so no deprecation cycle. But I kind of want to always use the context manager instead of actively setting it. Not sure that _is_ possible with the current version of the middleware.

pedro-psb: Ok, let's leave it there for know then

PR #7065, opened 2025-11-04
Failed tasks no longer store traceback.

aKlimau:
Failed task tracebacks are currently a part of task model, it can expose sensitive information from an exception via the API. This change stops this behavior by only logging tracebacks and not storing them inside of tasks.

_execute_task and _aexecute_task are modified to log tracebacks for unknown exceptions but never save them to the Task record.

Task.set_failed() is updated to make the tb (traceback) argument optional.

A new PulpExceptionNoTraceback base class is added for known, user-facing errors (like a DNS failure) where the traceback is not useful and should not be logged.

A new DnsDomainNameException (inheriting from PulpExceptionNoTraceback) is added to handle DNS lookup failures (e.g., bad remote URLs) as a known user error.

### DIFF
--- a/CHANGES/+no-traceback-task.removal
+++ b/CHANGES/+no-traceback-task.removal
@@ -1,0 +1,1 @@
+Stopped leaking sensitive information of failures in the task API.
--- a/pulpcore/app/models/task.py
+++ b/pulpcore/app/models/task.py
@@ -3,7 +3,6 @@ Django models related to the Tasking system
 """
 
 import logging
-import traceback
 from gettext import gettext as _
 
 from django.contrib.postgres.fields import ArrayField, HStoreField
@@ -261,7 +260,7 @@ class Task(BaseModel, AutoAddObjPermsMixin):
                 )
         self._cleanup_progress_reports(TASK_STATES.COMPLETED)
 
-    def set_failed(self, exc, tb):
+    def set_failed(self, exc):
         """
         Set this Task to the failed state and save it.
 
@@ -270,11 +269,9 @@ class Task(BaseModel, AutoAddObjPermsMixin):
 
         Args:
             exc (Exception): The exception raised by the task.
-            tb (traceback): Traceback instance for the current exception.
         """
         finished_at = timezone.now()
-        tb_str = "".join(traceback.format_tb(tb))
-        error = exception_to_dict(exc, tb_str)
+        error = exception_to_dict(exc)
         rows = Task.objects.filter(
             pk=self.pk,
             state=TASK_STATES.RUNNING,
--- a/pulpcore/app/tasks/repository.py
+++ b/pulpcore/app/tasks/repository.py
@@ -7,7 +7,7 @@ import hashlib
 
 from asgiref.sync import sync_to_async
 from django.db import transaction
-from rest_framework.serializers import ValidationError
+from pulpcore.exceptions import RepositoryVersionDeleteError
 
 from pulpcore.app import models
 from pulpcore.app.models import ProgressReport
@@ -44,12 +44,7 @@ def delete_version(pk):
             return
 
         if version.repository.versions.complete().count() <= 1:
-            raise ValidationError(
-                _(
-                    "Cannot delete repository version. Repositories must have at least one "
-                    "repository version."
-                )
-            )
+            raise RepositoryVersionDeleteError()
 
         log.info(
             "Deleting and squashing version {num} of repository '{repo}'".format(
--- a/pulpcore/exceptions/base.py
+++ b/pulpcore/exceptions/base.py
@@ -44,6 +44,18 @@ def exception_to_dict(exc, traceback=None):
     return {"description": str(exc), "traceback": traceback}
 
 
+class InternalErrorException(PulpException):
+    """
+    Exception to signal that an unexpected internal error occurred.
+    """
+
+    def __init__(self):
+        super().__init__("PLP0000")
+
+    def __str__(self):
+        return _("An internal error occurred.")
+
+
 class ResourceImmutableError(PulpException):
     """
     Exceptions that are raised due to trying to update an immutable resource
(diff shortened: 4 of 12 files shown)


pulpcore/tasking/tasks.py
```diff
@@ -173,7 +177,7 @@ def get_task_function(task, ensure_coroutine=False):
     is_coroutine_fn = asyncio.iscoroutinefunction(func)
 
     if immediate and not is_coroutine_fn:
-        raise ValueError("Immediate tasks must be async functions.")
+        raise NonAsyncImmediateTaskError(task_name=task.name)
```
mdellweg: When this happens, it's a programming error. This should not be user visible.

pulpcore/app/models/task.py
```diff
@@ -273,8 +284,12 @@ def set_failed(self, exc, tb):
             tb (traceback): Traceback instance for the current exception.
         """
         finished_at = timezone.now()
-        tb_str = "".join(traceback.format_tb(tb))
-        error = exception_to_dict(exc, tb_str)
+        error = {}
+        if tb:
```
mdellweg: Does this still happen? Don't we just want the error to be a single string like:
"Remote URL not found" or "Internal Error"?

pulpcore/exceptions/base.py
```diff
@@ -94,3 +94,75 @@ def __init__(self):
...
+        """
+        :param url: the url that dns could not resolve
+        :type url: str
+        """
+        super().__init__("PLP0008")
+        self.url = url
+
+    def __str__(self):
+        return _("Domain name was not found for {}. Check if specified url is valid.").format(
+            self.url
```
mdellweg: We should not leak urls either...

pulpcore/exceptions/base.py
```diff
@@ -94,3 +94,75 @@ def __init__(self):
...
+    def __init__(self, url):
+        """
+        :param url: the url that dns could not resolve
+        :type url: str
+        """
+        super().__init__("PLP0008")
+        self.url = url
+
+    def __str__(self):
+        return _("Domain name was not found for {}. Check if specified url is valid.").format(
```
mdellweg: Domains in the context of Pulp may be misleading. "URL lookup failed." Should be sufficient.

pulpcore/app/tasks/repository.py
```diff
@@ -44,12 +44,7 @@ def delete_version(pk):
             return
 
         if version.repository.versions.complete().count() <= 1:
-            raise ValidationError(
-                _(
-                    "Cannot delete repository version. Repositories must have at least one "
-                    "repository version."
-                )
-            )
+            raise RepositoryVersionDeleteError
```
mdellweg: Don't you need to instanciate the Exception?

pulpcore/app/models/task.py
mdellweg: Can we maybe instead add the PLP000X codes here?

CHANGES/+no-traceback-task.removal
jobselko, line 1: When you have a chance, could you please squash the commits where it makes sense? Also, if you want to sync the branch with `main`, rebasing instead of creating merge commits would help to keep the history clean.

pulpcore/exceptions/base.py
```diff
@@ -110,9 +122,7 @@ def __init__(self, url):
         self.url = url
 
     def __str__(self):
-        return _("Domain name was not found for {}. Check if specified url is valid.").format(
-            self.url
-        )
+        return _("URL lookup failed.")
```
mdellweg, line 125: There's more leaked data with the other exceptions.

pulpcore/exceptions/base.py
```diff
@@ -94,3 +106,73 @@ def __init__(self):
...
+        :type url: str
+        """
+        super().__init__("PLP0009")
+        self.url = url
+
+    def __str__(self):
+        return _("URL: {u} not supported.").format(u=self.url)
+
+
+class ProxyAuthenticationRequiredError(PulpException):
```
mdellweg: Isn't this more a "ProxyAuthenticationFailedError"? At which point "ProxyAuthenticationError" carries the same information.

pulpcore/tasking/tasks.py
```diff
@@ -70,10 +74,17 @@ def _execute_task(task):
             log_task_start(task, domain)
             task_function = get_task_function(task)
             result = task_function()
+        except (PulpException, PulpGlueException):
+            exc_type, exc, _ = sys.exc_info()
+            log_task_failed(task, exc_type, exc, None, domain)  # Leave no traceback in logs
+            task.set_failed(exc)
+            send_task_notification(task)
```
mdellweg: Ohhh (not your fault...) We do not use the return value of this function. And we call the notify in all branches.
If you want to, you can remove the return statements here (and in `_aexecute_task`) and combine the `send_task_excecution` statements into one.

pulpcore/tasking/tasks.py
```diff
@@ -95,10 +106,17 @@ async def _aexecute_task(task):
         try:
             task_coroutine_fn = await aget_task_function(task)
             result = await task_coroutine_fn()
+        except (PulpException, PulpGlueException):
+            exc_type, exc, _ = sys.exc_info()
+            log_task_failed(task, exc_type, exc, None, domain)  # Leave no traceback in logs
```
mdellweg: Comments are a delicate thing... They are not executed by the interpreter and usually ignored by humans. So I'd rather hint at "why" not at "what" you do here.
```suggestion
        except (PulpException, PulpGlueException):
            # Log expected ways to fail without a stacktrace.
            exc_type, exc, _ = sys.exc_info()
            log_task_failed(task, exc_type, exc, None, domain)
```

pulpcore/tasking/tasks.py
```diff
@@ -95,10 +106,17 @@ async def _aexecute_task(task):
...
+            await sync_to_async(task.set_failed)(exc)
+            send_task_notification(task)
         except Exception:
             exc_type, exc, tb = sys.exc_info()
-            await sync_to_async(task.set_failed)(exc, tb)
             log_task_failed(task, exc_type, exc, tb, domain)
+            # Generic exception for user
```
mdellweg: Similarly here:
```
# Unexpected Exceptions are most probably a programming error.
# Log error with a stack trace and give a generic error (equivalent of HTTP 500) to the user.
```

pulpcore/exceptions/base.py
aKlimau: Yes, but I was taking inspiration from HTTP 407, which has the same error name

PR #7076, opened 2025-11-10
Fix NoArtifactContentSerializer/Viewset failing on non-file fields

gerrod3:
Say you have a signature content unit (NoArtifactContentUpload) and you want to create it from a file on a remote server, you would get a 500 error when using the `file_url` field because the viewset expects only the `file` field. Modified the viewset and serializer to accept all our fields for content data.

### DIFF
--- a/pulpcore/plugin/serializers/content.py
+++ b/pulpcore/plugin/serializers/content.py
@@ -156,6 +156,20 @@ class UploadSerializerFieldsMixin(Serializer):
 class NoArtifactContentUploadSerializer(UploadSerializerFieldsMixin, NoArtifactContentSerializer):
     """A serializer for content types with no Artifact."""
 
+    def deferred_validate(self, data):
+        """Ensure file is present in validated_data."""
+        data = super().deferred_validate(data)
+        if "file" not in data:
+            if "artifact" in data:
+                artifact = data.pop("artifact")
+                with NamedTemporaryFile(mode="ab", dir=".", delete=False) as temp_file:
+                    temp_file.write(artifact.file.read())
+                    temp_file.flush()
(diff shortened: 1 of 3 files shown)


pulpcore/plugin/serializers/content.py
```diff
@@ -156,6 +156,20 @@ class Meta:
...
+        if "file" not in data:
+            if "artifact" in data:
+                artifact = data.pop("artifact")
+                with NamedTemporaryFile(mode="ab", dir=".", delete=False) as temp_file:
+                    temp_file.write(artifact.file.read())
+                    temp_file.flush()
+                data["file"] = PulpTemporaryUploadedFile.from_file(open(temp_file.name, "rb"))
+            else:
+                raise RuntimeError("No file found for NoArtifactContentUploadSerializer.")
+        return data
```
dralley, line 171: Why would this happen in the validate section?  Wouldn't it make more sense to do in `create()`?

gerrod3: No. Plugins overwrite the `deferred_validate` first, so it makes sense to put the logic there so they only have to overwrite one method. For this serializer plugins are expecting the content to be in `data['file']`, see line 128, but they don't expect to save that as an artifact since they are going to save the necessary data on the model. 

Note the similar, but different behavior on SingleArtifactContentUploadSerializer. In that serializer the plugins are expecting the content to be in `data['artifact']` with the artifact already having been saved to the database, see line 249.  

Basically plugins are expected to overwrite `deferred_validate` to be able to analyze the uploaded content to create their content unit. As part of this contract they are required to call `super` first to ensure the data is available first regardless of how the user uploaded it (file, artifact, upload, file_url). For NoArtifact the data is in `file` and for SingleArtifact it is in `artifact`.

PR #7089, opened 2025-11-12
[PULP-978] Add name_template to entrypoints

mdellweg:
All entrypoints are allowed to override the WORKER_NAME_TEMPLATE setting via the --name-template option.

### DIFF
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -53,6 +53,8 @@ preload_settings = Dynaconf(
 # Build paths inside the project like this: BASE_DIR / ...
 BASE_DIR = Path(__file__).absolute().parent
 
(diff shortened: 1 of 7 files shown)


ggainey (approved):

pulpcore/app/settings.py
```diff
@@ -53,6 +53,8 @@
 # Build paths inside the project like this: BASE_DIR / ...
 BASE_DIR = Path(__file__).absolute().parent
 
+WORKER_NAME_TEMPLATE = "{pid}@{hostname}"
```
evgeni, line 56: Because I will forget, but it's important for future-me: this can be overridden via `PULP_WORKER_NAME_TEMPLATE` environment variable.

Issue #6898, opened 2025-08-25
Add synchronous upload API for File content


dkliban:
File uploads currently dispatch a task. A new `/pulp/api/v3/content/file/files/upload` endpoint needs to be added to support uploading synchronously.

PR #7094, opened 2025-11-12
Adds synchronous upload API for Files.

dkliban:
closes: #6898


pulp_file/app/serializers.py
```diff
@@ -54,6 +59,62 @@ class Meta:
...
+                mode="ab", dir=settings.WORKING_DIRECTORY, delete=False
+            ) as temp_file:
+                for chunk in chunks:
+                    temp_file.write(chunk.file.read())
+                    chunk.file.close()
+                temp_file.flush()
+
+            # Convert to PulpTemporaryUploadedFile for later artifact creation
+            data["file"] = PulpTemporaryUploadedFile.from_file(open(temp_file.name, "rb"))
+            data.pop("upload")  # Remove upload from data
```
gerrod3: You can do a `.pop("upload", None)` at the top. Something like `if upload := data.pop("upload", None):`

pulp_file/app/serializers.py
```diff
@@ -54,6 +59,62 @@ class Meta:
...
+                artifact = serializer.save()
+            data["artifact"] = artifact
+
+        data["digest"] = data["artifact"].sha256
+
+        return data
+
+    class Meta:
+        # This API does not support uploading to a repository.
+        fields = tuple(f for f in FileContentSerializer.Meta.fields if f not in ["repository"])
```
gerrod3, line 112:
```suggestion
        fields = tuple(f for f in FileContentSerializer.Meta.fields if f not in ["repository", "file_url"])
```
You could also added support for `file_url` by adding an if after upload checking for it and then calling `.download(data['file_url'])`. It should be available on the serializer. It returns a PulpTemporaryUploadedFile, see https://github.com/pulp/pulpcore/blob/main/pulpcore/plugin/serializers/content.py#L131-L136

pulp_file/pytest_plugin.py
```diff
@@ -23,6 +25,16 @@ def file_bindings(_api_client_set, bindings_cfg):
...
+@pytest.fixture(scope="session")
+def file_client(bindings_cfg):
+    return FileApiClient(bindings_cfg)
+
+
+@pytest.fixture(scope="session")
+def file_content_api(file_client):
+    return ContentFilesApi(file_client)
+
+
```
gerrod3: You don't need these anymore. All the bindings are present in the `file_bindings` fixture. In your test you would simply write `file_bindings.ContentFilesApi.upload(...)`. The bindings object also has the `ApiException` for pulp-file-client attached to it as well.

pulp_file/app/serializers.py
```diff
@@ -54,6 +59,62 @@ class Meta:
...
+
+            # Convert to PulpTemporaryUploadedFile for later artifact creation
+            data["file"] = PulpTemporaryUploadedFile.from_file(open(temp_file.name, "rb"))
+        elif file_url := data.pop("file_url", None):
+            expected_digests = data.get("expected_digests", None)
+            expected_size = data.get("expected_size", None)
+            data["file"] = self.download(
+                file_url, expected_digests=expected_digests, expected_size=expected_size
+            )
+        elif file := data.pop("file", None):
```
gerrod3:
```suggestion
        if file := data.pop("file", None):
```
This should still be a separate if. The file content creation process looks like this:

1. If upload object -> convert to temp file, store in data as file
2. else if file-url -> download to temp file, store in data as file
3. If file in data -> convert to artifact, store in data
4. Use artifact in data to create content (note the first three are skipped if user directly uses an artifact for this upload)

The first three steps are what needs to happen in the method, the last step is what we do in the `create` method that is called during `serializer.save()`.

PR #7113, opened 2025-11-19
Mitigate a disk consumption issue during sync

### LINKED ISSUE #7064: Syncing of repositories with big packages can cause disk storage to be exhausted
**Version**

3.92.1

**Describe the bug**

When syncing a repository with large package files, the download stage can cause the disk storage to be exhausted. The download stage runs particularly faster than the processing stage, and can cause content files to accumulate in the temporary storage.

https://github.com/pulp/pulpcore/pull/7037 was opened in anticipation to fix this, but it doesn't prevent downloads from happening in the background.

**To Reproduce**

Sync a repository with large package files. Use object storage backend to reproduce more evidently.

**Expected behavior**

Pulp administrators able to control the maximum number of content stored on disk at any moment. The pipeline pauses once this limit is hit, and waits till the queue to drain out and resume.

**Additional context**

The problem gets more difficult when deployments is on Kubernetes, where the disk storage available to a worker pod is limited, and upon requesting more the pod might get evicted, without Pulp knowing about that.

dralley:
The ArtifactSaver stage is acting as a bottleneck due to batching, and as a result artifacts downloaded by the ArtifactDownloader stage aren't being flushed out quickly enough. Using a default batch size of 500 is too much for some stages.

closes #7064

### DIFF
--- a/CHANGES/7064.bugfix
+++ b/CHANGES/7064.bugfix
@@ -1,0 +1,1 @@
+Restrict batch processing `minsize` to the settings value defined for `MAX_CONCURRENT_CONTENT`. Smaller values will increase sync times but reduce the amount of outstanding content in the pipeline, which reduces peak resource consumption (memory, temporary disk requirements).
--- a/docs/admin/reference/settings.md
+++ b/docs/admin/reference/settings.md
@@ -429,7 +429,7 @@ Defaults to `/var/lib/pulp/tmp/`.
 The size of the batch of content processed in one go when syncing content from
 a remote.
 
-Defaults to 200.
+Defaults to 25.
 
 ## Redis Settings
 
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -381,7 +381,7 @@ HIDE_GUARDED_DISTRIBUTIONS = False
 
 DOMAIN_ENABLED = False
 
-MAX_CONCURRENT_CONTENT = 200
+MAX_CONCURRENT_CONTENT = 25
 
 SHELL_PLUS_IMPORTS = [
     "from pulpcore.app.util import get_domain, get_domain_pk, set_domain, get_url, extract_pk",
--- a/pulpcore/plugin/stages/artifact_stages.py
+++ b/pulpcore/plugin/stages/artifact_stages.py
@@ -269,7 +269,7 @@ class ArtifactSaver(Stage):
         Returns:
             The coroutine for this stage.
         """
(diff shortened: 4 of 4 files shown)


pulpcore/plugin/stages/api.py
```diff
@@ -79,7 +79,7 @@ async def run(self):
             log.debug("%(name)s - next: %(content)s.", {"name": self, "content": content})
             yield content
 
-    async def batches(self, minsize=500):
+    async def batches(self, minsize=settings.MAX_CONCURRENT_CONTENT):
```
dralley: I think this *probably* makes sense, but I have some additional thoughts... I wonder if it's not sufficient.

* Maybe `ArtifactSaver` stage should use an independently configurable batch size, such that it can be made lower without compromising the performance of the rest of the pipeline?
* Does it actually make sense to use `settings.MAX_CONCURRENT_CONTENT` here?  The description of it (see https://github.com/pulp/pulpcore/pull/7037) makes it sound very much like that would be the case.  But currently that value is only used within the `ArtifactDownloader` stage and maybe the meaning is distinct enough to warrant being independent anyway

ggainey: So GenericDownloader's default is 200 and this iterator's is 500 - but I suspect that both of those numbers are in the category of "well this seems to be a reasonable number". I don't think the possible gain in fine-grained-optimization is worth the complexity of having them be separate. FWIW, I'd go with this implementation, unless/until there's some serious proof it's a bad idea.

dralley: The main question on my side is, we might want the value to be significantly below 200 on the "how many artifacts do we want to save at a time (how many should be batched up)".  Maybe we do, and maybe there's a decent reason to keep the other value around 200 or at least a good reason not to drop it as low as we might want the batching before the `ArtifactSaver` stage.

But otherwise I agree with you, I'd love for it to be this simple.

ggainey:
> @balasankarc Was there a particular reason for choosing the number 200? Is there any good reason for keeping these independent (e.g. reducing the batching even more while keeping the concurrency higher), or otherwise to not reduce `MAX_CONTENT_CONCURRENCY` default to 25?

Pretty sure it was 200 because that was the original hard-coded value : https://github.com/pulp/pulpcore/pull/7037/files#diff-28560cb4dba596314e3577becbd1974a015d47c08f8f17b38e6030a701999169L142 - and it's been that way basically Forever.

PR #7116, opened 2025-11-21
[PULP-1002] Relax touch

mdellweg:


pulpcore/app/models/content.py
```diff
@@ -106,16 +106,22 @@ class BulkTouchQuerySet(models.QuerySet):
...
+        # that updating that row's timestamp-of-interest is the responsibility of whoever currently
+        # owns it, this results in correct data, while closing the window on deadlocks.
+        # Django requires "select_for_update" to be used in a transaction even if this is a
+        # singular query.
         with transaction.atomic():
-            sub_q = self.order_by("pk").select_for_update(skip_locked=True)
+            sub_q = (
+                self.filter(timestamp_of_interest__lt=now() - datetime.timedelta(hours=1))
+                .order_by("pk")
+                .select_for_update(skip_locked=True, no_key=True)
```
dralley, line 125: A comment on why `no_key` is safe in this particular usage would be good

pulpcore/app/models/content.py
```diff
@@ -106,17 +106,25 @@ class BulkTouchQuerySet(models.QuerySet):
...
+        # Postgres' UPDATE call doesn't support order-by. This can (and does) result in deadlocks in
+        # high-concurrency environments, when using touch() on overlapping data sets. In order to
+        # prevent this, we choose to SELECT FOR UPDATE with SKIP LOCKS == True, and only update
+        # the rows that we were able to get locks on. Since a previously-locked-row implies
+        # that updating that row's timestamp-of-interest is the responsibility of whoever currently
+        # owns it, this results in correct data, while closing the window on deadlocks.
+        # Django requires "select_for_update" to be used in a transaction even if this is a
+        # singular query.
+        # `no_key` translates to `SELECT ... FOR NO KEY UPDATE` and results in a different type of
+        # lock being used. We don't change the any (primary/foreign) keys.
```
ggainey:
```suggestion
        # lock being used. We don't change any primary/foreign keys.
```

PR #7149, opened 2025-12-05
Improve migration error message with extra instructions

pedro-psb:
There are cases where user set components TTL to huge values (e,g > 1 hour, believe it or not). In this case, the migrations might fail if it tried to force shutdown the compoments (e.g, workers) to perform the migration, as there will be stale AppStatus entries in the database.

### DIFF
--- a/pulpcore/migrations.py
+++ b/pulpcore/migrations.py
@@ -78,7 +78,10 @@ class RequireVersion(Operation):
                         f"({self.plugin} >= {self.version} needed):",
                         *errors,
                         "Please shutdown or upgrade the outdated components before you "
-                        "continue the migration.",
+                        "continue the migration. \n"
+                        "If the components were not gracefully shutdown, wait for the "
+                        "larger component TTL period before running the migration ",
+                        "(API_APP_TTL, CONTENT_APP_TTL or WORKER_TTL).",
                     ]
                 )
             )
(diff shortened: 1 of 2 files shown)


pulpcore/migrations.py
```diff
@@ -78,7 +78,9 @@ def database_forwards(self, app_label, schema_editor, from_state, to_state):
                         f"({self.plugin} >= {self.version} needed):",
                         *errors,
                         "Please shutdown or upgrade the outdated components before you "
-                        "continue the migration.",
+                        "continue the migration. \n"
+                        "If the components were not gracefully shutdown, wait for the "
+                        "larger component TTL period before running the migration.",
```
ggainey: Can we call out the specific TTL-setting involved here, so the user knows *exactly* what timeout to expect?

ggainey (changes requested):

pulpcore/migrations.py
```diff
@@ -78,7 +78,10 @@ def database_forwards(self, app_label, schema_editor, from_state, to_state):
                         f"({self.plugin} >= {self.version} needed):",
                         *errors,
                         "Please shutdown or upgrade the outdated components before you "
-                        "continue the migration.",
+                        "continue the migration. \n"
+                        "If the components were not gracefully shutdown, wait for the "
+                        "larger component TTL period before running the migration ",
+                        "(API_APP_TTL, CONTENT_APP_TTL or WORKER_TTL).",
```
ggainey, line 84: :+1:

PR #7217, opened 2026-01-17
[PULP-1117] Add task-specific log capturing to task diagnostics mechanism

### LINKED ISSUE #7214: As a user, I can enable DEBUG logging level for individual tasks

Debug logging is difficult to enable in many contexts (you need direct admin access to the box, you need to restart services), and result in overly excessive logging for many debugging purposes.  It should be possible to enable debug logging for individual tasks similarly to how profiling can be done for individual tasks using the X-TASK-DIAGNOSTICS header.

 Instead of turning on log level DEBUG for the entire pulp instance, including all workers and the content service, it would be better if a user could send a header with their request to the API which would enable DEBUG level logging for that individual task. That way, users can avoid touching settings.py, and avoid causing unnecessarily voluminous logging for a specific debugging situation.

dralley:
TASK_DIAGNOSTICS and X-Task-Diagnostics can now capture all logs for specific tasks and present those via the profile artifacts mechanism. Because the logs are specific to the task, the logs are linear and not broken up by unrelated logs from other services/workers/tasks.

Assisted By: Claude

closes #7214

### DIFF
--- a/docs/dev/guides/pull-request-walkthrough.md
+++ b/docs/dev/guides/pull-request-walkthrough.md
@@ -14,9 +14,9 @@ Please be sure to follow the [Pulp policy on AI Usage](site:help/more/governance
 
 1. Add `functional tests` or `unit tests` where appropriate and ensure tests
    are passing on the CI.
-2. Add a [`CHANGES entry`](site:pulpcore/docs/dev/guides/git/#markdown-header-changelog-update).
+2. Add a [`CHANGES entry`](site:pulpcore/docs/dev/guides/git/#changelog-update).
 3. Update relevent `documentation`. Please build the docs to test!
-4. If the PR is a simple feature or a bugfix, [`rebase and squash`](site:pulpcore/docs/dev/guides/git/#markdown-header-rebasing-and-squashing) to a single commit.
+4. If the PR is a simple feature or a bugfix, [`rebase and squash`](site:pulpcore/docs/dev/guides/git/#rebasing-and-squashing) to a single commit.
    If the PR is a complex feature, make sure that all commits are cleanly separated and have meaningful commit messages.
 5. Make sure you tag commits with `closes #IssueNumber` or `ref #IssueNumber` when working on a tracked issue.
 6. If AI was used, make sure you are following the [Pulp policy on AI Usage](site:help/more/governance/ai_policy/)


docs/dev/learn/tasks/diagnostics.md
```diff
@@ -35,7 +35,9 @@ The following diagnostics are supported currently:
 - memray:
    Dumps a profile which can be processed with `memray`, which shows which lines and functions were
    responsible for the most allocations at the time of peak RSS of the process
-
+- logs:
+   Dumps all logs specific to the task, including DEBUG logs, regardless of whether DEBUG logging
+   is otherwise enabled
```
dralley: TODO: should this be the default behavior, or should there be a separate modifier to unfilter the logs?

Are there any concerns around data leakage?  System logs were not previously thought of as potentially exposable via the API.

I filed https://github.com/pulp/pulpcore/issues/7214 before thinking of this specific idea - the idea there was to unfilter the logs for that task specifically in the log handler.  I'm not actually sure if that is possible or desirable though.  Maybe that setting could control the log level collected here, but maybe that's unnecessarily configurable.

pulpcore/tasking/_util.py
```diff
@@ -218,6 +220,53 @@ def __memray_diagnostic_decorator(task):
...
+        # Create a file handler that captures all logging levels
+        file_handler = logging.FileHandler(log_file_path, mode='w', encoding='utf-8')
+        file_handler.setLevel(logging.NOTSET)  # Capture all levels
+
+        # Create a formatter for consistent log formatting
+        formatter = logging.Formatter(
+            '%(asctime)s - %(name)s - %(levelname)s - %(message)s',
+            datefmt='%Y-%m-%d %H:%M:%S'
+        )
+        file_handler.setFormatter(formatter)
```
dralley, line 237: TODO: this is not the same formatter as the standard Pulp one. Should we use the same formatter?  Maybe not?  Correlation ID is a bit extraneous when all logs relate to the same task, the `pulp` prefix is irrelevant, removing some of that stuff is maybe useful.

dralley:
TODO: not working for the prune API for some reason?  Maybe not working for task groups?  Double check inheritance of these settings.

docs/dev/learn/tasks/diagnostics.md
dralley: After figuring out how to implement this (see the last commit marked WIP), I kinda do think that it should be at most opt-in (if we do it at all), because it requires you to change the log level of the root logger, otherwise debug logs get filtered out completely.

dralley: You could probably implement it as two separate features.  `logs` and `debug-logs`, where the latter modifies the behavior of the former

bmbouter: If the question is should task-level logging be off by default, I think the answer is yes. After users upgrade into this, they should get the same they had before.

bmbouter: I'm not concerned with users being able to see their own logs from a data leakage perspective.

dralley: @bmbouter I mean should task-specific logs be DEBUG-by-default, or should that be opt-in, or should it be allowed in the first place.

The implementation of it is necessarily a bit hacky (changing the worker's root logger's level - at runtime - for the duration of the task before switching back).  The less-hacky alternatives also have downsides.

bmbouter: I think it should be debug and only debug. It keeps it simple, and this is exclusively for debugging what's going on in a task... That's my take at least. If we want it more/better/different later, we can do more on it later.

dralley: So you're fine with this? https://github.com/pulp/pulpcore/pull/7217/commits/5532ddc19665c2c9a2151e70ff7e3f6e898f6f05#diff-1340f0b578800854ca335c83c0d5aafb9e7be9e9b3776e6171b2966e6d4af161R239

bmbouter: I don't see DEBUG named there, but overall yes that looks fine.

dralley: For the root log handler NOTSET and DEBUG are equivalent.  The difference is that NOTSET looks at the log level of the parent log handler.  But for the root handler there is none.

pulpcore/tasking/_util.py
mdellweg: Maybe add the correlation id to the top of the file once.

dralley: Yes, I was thinking something along those lines.

docs/dev/learn/tasks/diagnostics.md
dralley: So, we had a discussion about this today, and the conclusion was basically that we already have a specific role for restricting access to `/profile_artifacts/`, so as long as that is restricted to the proper subset, it's probably OK.

pulpcore/tasking/_util.py
dralley: On second thought, of what use is correlation ID if you have no other context (request info etc.) to correlate it with?

mdellweg: Aren't you going too look at this file in a bigger context? Also when this task call out to subsequent services, it might pass the correlation id down that way. I think having is better than needing here.

dralley: If you have access to the task  to trigger the `/profile_artifacts/` API call, then you can just get the logging_cid from the task record, should you need it.

pulpcore/middleware.py
```diff
@@ -180,6 +180,8 @@ def __init__(self, get_response):
         self.get_response = get_response
 
     def __call__(self, request):
+        # import logging
+        # logging.error(request.META)
```
dralley: whoops, fixing this

docs/dev/guides/pull-request-walkthrough.md
mdellweg, line 1: Not technically related, but let's just silently fix this (it's a separate commit anyway)...

pulpcore/tasking/tasks.py
```diff
@@ -352,19 +352,26 @@ async def adispatch(
 def get_task_payload(
     function_name, task_group, args, kwargs, resources, versions, immediate, deferred, app_lock
 ):
+    """Create arguments for creation of a new task"""
+    current_task = Task.current()
+    profile_options = x_task_diagnostics_var.get(None)
```
mdellweg: `x_task_diagnostics_var` is a context var. Wouldn't it make sense to set it up again when entering the task context instead of applying extra logic for subtasks?

pulpcore/tasking/_util.py
```diff
@@ -218,6 +220,49 @@ def __memray_diagnostic_decorator(task):
     return __memray_diagnostic_decorator
 
 
+def _logging_decorator(temp_dir, func):
```
mdellweg: I am still a little concerned that this will capture any logs happening during task execution.
As long as we call a task in a subprocess this will probably work. But we already have (fast, small) tasks "run in the forground" and we might introduce running async tasks concurrently. I don't have an answer for this ready.

pulpcore/tasking/_util.py
```diff
@@ -121,6 +121,9 @@ def _execute_task_and_profile(task, profile_options):
             _execute_task = _pyinstrument_diagnostic_decorator(temp_dir, _execute_task)
         if "memray" in profile_options:
             _execute_task = _memray_diagnostic_decorator(temp_dir, _execute_task)
+        if "logs" in profile_options or "debug-logs" in profile_options:
+            is_debug = "debug-logs" in profile_options
+            _execute_task = _logging_decorator(temp_dir, is_debug, _execute_task)
```
mdellweg: If you want:
```suggestion
        is_debug := "debug-logs" in profile_options
        if is_debug or "logs" in profile_options:
            _execute_task = _logging_decorator(temp_dir, is_debug, _execute_task)
```

docs/dev/guides/pull-request-walkthrough.md
dralley: Right.  I had seen the `markdown-header-` prefix thing elsewhere, but it turns out it is just unnecessary and actually doesn't work on this particular markdown flavor.

pulpcore/tasking/tasks.py
dralley: Look at "temp commit", is  that what you mean?  (maybe that code would be moved to `contexts.py` in a final impl - but in general principle)

pulpcore/tasking/_util.py
dralley: You bring up a decent point, I could try to test this with tasks that are forced to run in immediate mode.

dralley: We could, in theory, add a new ContextVar to store the mode of execution, and only execute the task logging for in-worker tasks.

pulpcore/tasking/tasks.py
mdellweg: Exactly. And +1 to moving it into the `with_task_context`.

pulpcore/tasking/_util.py
mdellweg: I'd rather prefer to log from the same context scope but i doubt python has it in it.
We could try to leverage the correlation_id once again. At least it's not so much of an issue if logs from a related task were cought too.

dralley: Well, if an `immediate` task still runs in the worker process, it should probably still be isolated at least?  It may pick up debug logs from the worker (e.g. heartbeats) because it's not getting its own subprocess, but it's not going to pick up anything from other tasks running, or the server process.

That's not ideal but iI don't think it would be a security issue?

mdellweg: Well, once we start being able to run tasks asynchronously in the worker, we could maybe start to running multiple at the same time. But i must admit that is more than just a projection into the future.

dralley: If that's something we have planned then we could leave a note.  But IMO, isolation is one of the more important features of the task model of execution anyway so I would hope we wouldn't do that.

pulpcore/tasking/tasks.py
dralley: So, it looks like we have circular import issues if we do this, because then `contexts.py` has to import from `middleware.py`

added a new commit to resolve the circular import also.

PR #7220, opened 2026-01-18
Implement atomic hashing and chunking in PulpExport

### LINKED ISSUE #7218: Race condition in chunked exports leads to missing file hashes in TOC
**Version**
pulpcore 3.101.0

**Describe the bug**
The current export mechanism suffers from a race condition and redundant I/O. The process relies on `subprocess.Popen` to pipe data into the system `split` utility.
(Reference: https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/export.py#L400)

Hashes for the resulting split files are calculated in a separate pass *after* the splitting process has completed.
(Reference: https://github.com/pulp/pulpcore/blob/main/pulpcore/app/tasks/export.py#L429)

This creates a race condition: if an external process moves or processes one of the split files before the hashing loop encounters it, the file will not be hashed, and it will be missing from the export's Table of Contents (TOC). Furthermore, this approach requires reading every exported file twice (once for the initial write by `split`, and once for the subsequent hash calculation), significantly increasing the I/O overhead.

**To Reproduce**

1. Initiate a large Pulp export with a `chunk_size` that forces multiple split files.
2. Monitor the filesystem location where `split` is writing the `.tar` chunks.
3. As soon as a chunk file is completed (e.g., `.0000`), immediately move or rename it while the export is still generating subsequent chunks.
4. Observe that the task finishes successfully, but the generated `-toc.json` file is missing the entry for the moved file, resulting in an incomplete export record.

**Expected behavior**
The export process should be atomic regarding file creation and hashing. Every file generated by the export must be guaranteed to appear in the Table of Contents with its correct hash, regardless of file system timing. The system should not rely on files remaining static on disk between the "write" phase and the "hash" phase.

**Additional context**
This can be resolved by replacing the `subprocess` + `split` dependency with a custom file-like object class. This class would handle byte-stream splitting and in-memory hash calculation simultaneously.

By hashing the data before the file handle is closed, we eliminate the race condition and reduce disk I/O by 50% for the hashing phase, as the files no longer need to be re-read from disk. This also removes the dependency on the system's `split` binary, making the export logic more portable and easier to test.

vycius:
Replace the external 'split' subprocess and secondary hashing pass with a custom HashingFileWriter.

- Resolves a race condition where files moved/deleted between the write and hash phases caused export failures.
- Eliminates redundant I/O by calculating checksums while streaming data to disk.
- Removes OS dependency on the 'split' binary.
- Ensures the Table of Contents (TOC) is generated from in-memory metadata for 100% consistency.

closes #7218

### DIFF
--- a/pulpcore/app/util.py
+++ b/pulpcore/app/util.py
@@ -3,11 +3,10 @@ import zlib
 import os
 import socket
 import tempfile
-from _hashlib import HASH
 from io import RawIOBase
 from pathlib import Path
 from types import TracebackType
-from typing import Self, IO, Callable
+from typing import Self, IO, Any
 
 import gnupg
 
@@ -666,7 +665,7 @@ class HashingFileWriter(RawIOBase):
     def __init__(
         self,
         base_path: str | Path,
-        hasher_cls: Callable[[], HASH],
+        hasher_cls: Any,
         chunk_size: int = 0,
         suffix_length: int = 4,
         dir_mode: int = 0o775,
@@ -690,7 +689,7 @@ class HashingFileWriter(RawIOBase):
 
         self._file_index: int = 0
         self._current_file: IO[bytes] | None = None
-        self._current_hasher: HASH | None = None
+        self._current_hasher: Any | None = None
         self._current_file_path: Path | None = None
 
         self._current_chunk_written: int = 0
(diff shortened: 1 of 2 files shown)


dralley (approved):
@ggainey I'll leave this un-merged in case you'd like to review it tomorrow (or on Wednesday given possible PTO)

pulpcore/app/util.py
```diff
@@ -3,6 +3,12 @@
 import os
 import socket
 import tempfile
+from _hashlib import HASH
```
dralley: So, discussing this, we thing this is probably an instance where the cost of type hints is probably more than the benefit.  It's not *wrong*, per-se, but this is not really a public class that is meant to be imported, it's actually in `hashlib._hashlib`.

In this instance we would probably prefer to drop the type hint over importing "private" implementation details only for the type hint.

PR #7221, opened 2026-01-18
Add network configuration support for synchronous file uploads

vycius:
Introduce RemoteNetworkConfigSerializer to centralize shared network settings (timeouts, proxies, and certificates) and ensure consistent validation across persistent Remotes and ad-hoc upload downloads.

- Exclude sync-specific fields (policy, rate_limit) from shared config
- Pass validated network kwargs to the downloader in UploadSerializerFieldsMixin
- Reuse common proxy and certificate validation logic

closes: #7201

### DIFF
--- a/pulp_file/app/serializers.py
+++ b/pulp_file/app/serializers.py
@@ -66,6 +66,8 @@ class FileContentUploadSerializer(FileContentSerializer):
 
     def validate(self, data):
         """Validate the FileContent data."""
+        data = super().validate(data)
+
         if upload := data.pop("upload", None):
             # Handle chunked upload
             chunks = models.UploadChunk.objects.filter(upload=upload).order_by("offset")
--- a/pulpcore/app/serializers/base.py
+++ b/pulpcore/app/serializers/base.py
@@ -7,6 +7,7 @@ from collections import namedtuple
 from typing import List, TypedDict
 from urllib.parse import urljoin
 
+from cryptography.x509 import load_pem_x509_certificate
 from django.conf import settings
 from django.core.validators import URLValidator
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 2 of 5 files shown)


pulpcore/app/serializers/repository.py
```diff
@@ -304,72 +241,37 @@ def validate_url(self, url):
...
+                "policy",
+                "total_timeout",
+                "connect_timeout",
+                "sock_connect_timeout",
+                "sock_read_timeout",
+                "headers",
+                "rate_limit",
+                "hidden_fields",
+            )
+            + tuple(RemoteNetworkConfigSerializer().get_fields().keys())
```
pedro-psb, line 161: Most of there hardcoded fields are now duplicates of RemoteNetworkConfigSerializer fields. And I believe you can use `RemoteNetworkConfigSerializer.Meta.fields` for this.

pulpcore/plugin/serializers/content.py
pedro-psb, line 26: Doesn't this also require inheriting from `RemoteNetworkConfigSerializer`?

pulpcore/app/serializers/base.py
pedro-psb, line 1: Circular import  between `pulpcore/app/serializers/base.py` and `pulpcore.app.serializers.repository`.
Move `validate_certificate` elsewhere.

pulp_file/app/serializers.py
```diff
@@ -65,6 +65,8 @@ class FileContentUploadSerializer(FileContentSerializer):
     """
 
     def validate(self, data):
+        data = super().validate(data)
+
```
pedro-psb, line 70: Place this after the docstring.

pulpcore/app/serializers/repository.py
pedro-psb, line 1: The `RemoteSerializer` inherits from `RemoteNetworkConfigSerializer` but it has duplicate fields, like `total_timeout` and `connection_timeout`. Remove the duplicates from here.

pedro-psb, line 1: Move `rate_limit` and `download_concurrency` to `RemoteNetworkConfigSerializer` too.
These are all related to the remote downloader.

pulpcore/plugin/serializers/content.py
pedro-psb, line 1: I'm giving this more thought and bloating the Upload serializer with all these network config options doesn't sounds a good idea. The remote download is an optional feature for the upload API, it should't take the whole namespace.

Can you instead re-use the network config serializer as a field?
So we have `file_url` and a `downloader_config` (based on the network serializer), which is only required if `file_url` is used.

pulpcore/app/serializers/repository.py
pedro-psb: Please use `RemoteNetworkConfigSerializer.Meta.fields`, unless there is a reason not to.

pulpcore/plugin/serializers/content.py
vycius: That is an excellent idea.

pulpcore/app/serializers/repository.py
vycius: `RemoteNetworkConfigSerializer` is a `Serializer`, not a `ModelSerializer`, so it doesn't have a `Meta` class or a `fields` attribute by default. Adding one would be non-standard, whereas `get_fields().keys()` correctly and dynamically retrieves the manual definitions.

PR #7230, opened 2026-01-20
Update error handling documentation to include new pulp exceptions

aKlimau:


docs/dev/learn/other/error-handling.md
```diff
@@ -2,10 +2,190 @@
...
+- Implements a user-safe `__str__()` method that includes the error code in `[PLP####]` format followed by a description without sensitive data
+
+When a `PulpException` is raised in a task, the error message (including the error code) and description are returned to the user.
+Tracebacks are logged but never exposed through the API.
+
+## Available Pulp Exceptions
+
+Listed below are all available PulpExceptions, sorted by error code.
+
+### InternalErrorException (PLP0000)
```
mdellweg: For easier lookup, make the PLPxxxx align in the same column by moving them to the front.

mdellweg: Also this reference list should really move over to the users section.

docs/user/reference/pulp-errors.md
```diff
@@ -0,0 +1,5 @@
+# Pulp Errors
+
+When working with Pulp, you may encounter error codes and exceptions during task execution.
+
+To find more information about Pulp errors, see the [Error Handling](site:pulpcore/docs/dev/learn/other/error-handling/) documentation.
```
mdellweg: Can you move all the _user facing_ errors over to here? We should not redirect the user into the dev docs, rather the other way around.

docs/dev/learn/other/error-handling.md
```diff
@@ -2,10 +2,172 @@
...
+Below are implementation examples for the available PulpExceptions.
+
+### PLP0000 - InternalErrorException
+
+Signals an unexpected internal error.
+This is raised automatically by the task system when an uncaught exception occurs that is not a `PulpException`.
+
+**Usage in code:**
+```python
+# In pulpcore/tasking/tasks.py:87
```
jobselko: Specific line numbers will get outdated quickly imo. Maybe it would be better not to mention them?

aKlimau: Should I just leave just a file path in this case or is it better to not mention location within files at all?

jobselko: Up to you, I do not have a preference

PR #7232, opened 2026-01-20
Add a list-signing-services command

daviddavis:
Adding a new list-signing-services command which makes it easier to manage signing services alongside the add/remove signing services command. This is helpful if you don't have immediate access to the Pulp API.

### DIFF
--- a/pulpcore/app/management/commands/list-signing-services.py
+++ b/pulpcore/app/management/commands/list-signing-services.py
@@ -1,0 +1,18 @@
+from django.apps import apps
+from django.core.management import BaseCommand
+
+
+class Command(BaseCommand):
+    """
+    Django management command for listing signing services.
+    """
+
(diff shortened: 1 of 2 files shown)


pulpcore/app/management/commands/list-signing-services.py
```diff
@@ -0,0 +1,18 @@
...
+
+    help = "List all SigningServices."
+
+    def add_arguments(self, parser):
+        pass
+
+    def handle(self, *args, **options):
+        SigningService = apps.get_model("core", "SigningService")
+        results = list(SigningService.objects.all().values_list("name", flat=True))
+        print(results)
```
mdellweg: Django suggests using `self.stdout.write` here.

https://docs.djangoproject.com/en/5.2/howto/custom-management-commands/

PR #7242, opened 2026-01-23
Test PR: pin azure-storage-blob

daviddavis:
A temporary fix until https://github.com/Azure/Azurite/issues/2623 is resolved.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,7 +70,7 @@ dependencies = [
 sftp = ["django-storages[sftp]==1.14.6"]
 s3 = ["django-storages[boto3]==1.14.6"]
 google = ["django-storages[google]==1.14.6"]
-azure = ["django-storages[azure]==1.14.6"]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -70,7 +70,7 @@ dependencies = [
 sftp = ["django-storages[sftp]==1.14.6"]
 s3 = ["django-storages[boto3]==1.14.6"]
 google = ["django-storages[google]==1.14.6"]
-azure = ["django-storages[azure]==1.14.6"]
+azure = ["django-storages[azure]==1.14.6", "azure-storage-blob<12.28.0"]
```
mdellweg: That's something we have covered by the ci_constraints. I don't expect this to change a thing...

daviddavis: I see. It seems like there is a very old version of azure-storage-blob that might work but even 12.25.1 seems to have compatibility issues. I think I'll go with the other PR to skip the api version check in azurite.

PR #7265, opened 2026-01-27
Fix list-signing-services command

daviddavis:
Fix an error when passing results to stdout.write() since unlike print() it expects a string and can't handle a list.

### DIFF
--- a/pulpcore/app/management/commands/list-signing-services.py
+++ b/pulpcore/app/management/commands/list-signing-services.py
@@ -15,4 +15,4 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         SigningService = apps.get_model("core", "SigningService")
         results = list(SigningService.objects.all().values_list("name", flat=True))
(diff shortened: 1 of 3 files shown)


pulpcore/app/management/commands/list-signing-services.py
```diff
@@ -15,4 +15,4 @@ def add_arguments(self, parser):
     def handle(self, *args, **options):
         SigningService = apps.get_model("core", "SigningService")
         results = list(SigningService.objects.all().values_list("name", flat=True))
-        self.stdout.write(results)
+        self.stdout.write(str(results))
```
mdellweg: Would it be nicer to return the list as one name per line? (think `"\n".join(results)`)
(Easier to pipe process in shell...)
Also at this point how about adding a unittest?

mdellweg: Thank you!

PR #7271, opened 2026-01-29
Add more useful PulpExceptions for general tasks

gerrod3:
- general_create/update now properly raise their ValidationErrors
- there are new Sync/PublishErrors for plugins
- import/export get their own errors
- rest of the validation errors now inherit from PulpException

After looking through the plugins, most exceptions are some Runtime or ValueError and I'm not sure if there is an easy generic we can put into Pulpcore to cover all. Many of them are in our big tasks like sync/publish so I created two catch-alls for Sync & Publish. I was thinking of creating a specific error for Download failures, but our download logic is too convoluted to find a good place, and all the plugins do their own thing to check if a download failed. One big exception I didn't bother with is RBAC hooks. If they fail on object creation then the task won't show useful info, but this should only happen if the user customizes their access policies incorrectly.

fixes: #7270
Assisted-by: claude-4.5-sonnet
https://issues.redhat.com/browse/PULP-1171

### 📜 Checklist

### DIFF
--- a/pulpcore/exceptions/base.py
+++ b/pulpcore/exceptions/base.py
@@ -8,15 +8,11 @@ class PulpException(Exception):
     """
 
     http_status_code = http.client.INTERNAL_SERVER_ERROR
+    error_code = None
 
-    def __init__(self, error_code):
-        """
-        :param error_code: unique error code
-        :type error_code: str
-        """
-        if not isinstance(error_code, str):
-            raise TypeError(_("Error code must be an instance of str."))
-        self.error_code = error_code
+    def __init__(self):
+        if not isinstance(self.error_code, str):
+            raise NotImplementedError("ABC error. Subclass must define a unique error code.")
 
     def __str__(self):
         """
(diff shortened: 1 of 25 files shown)


pulpcore/exceptions/base.py
```diff
@@ -8,15 +8,17 @@ class PulpException(Exception):
...
         :param error_code: unique error code
         :type error_code: str
         """
-        if not isinstance(error_code, str):
-            raise TypeError(_("Error code must be an instance of str."))
-        self.error_code = error_code
+        if error_code:
+            if not isinstance(error_code, str):
+                raise TypeError(_("Error code must be an instance of str."))
+            self.error_code = error_code
```
mdellweg: This is an abstract base class. So this seems overly complex.
We should be able to go with
```
def __init__(self):
    if not isinstance(self.error_code, str):
        raise NotImplementedError("ABC error. Subclass must define a unique error_code.")

pulpcore/exceptions/base.py
mdellweg: Maybe now that we are there:
`return {"code": exc.error_code, "description": str(exc)}`

This should help to make it machine readable.

pulpcore/exceptions/base.py
mdellweg: One could think about making this capable of providing on-the-fly defined errors, but I think that would not be helping with the goal to get more consistent errors.

gerrod3: Yeah, that's partly why I made the class so complicated to keep the idea of defining on-the-fly errors, but it does go against trying to be more consistent.

PR #7274, opened 2026-02-03
Improve logging for heartbeat failures by including the actual Exception

YasenT:
Improving the logging for heartbeat failures. Right now we have a static message, which doesn't show the Exception details.
This will help with troubleshooting.


pedro-psb (commented):
@YasenT Don't you want to add it to the PulpApiWorker as well?
(at `pulpcore.app.entrypoint`)

Issue #7184, opened 2025-12-29
Sync failures due to NEVRA duplicates should also report the conflicting packages


simenon:
**Is your feature request related to a problem? Please describe.**
I am trying to setup an on-demand sync with rpm repo of rpm.grafana.com, but i hit a NEVRA issue.

Output of the sync task mentions this

```
"error": {
    "traceback": "  File \"/usr/local/lib/python3.11/site-packages/pulpcore/tasking/tasks.py\", line 72, in _execute_task\n    result = task_function()\n             ^^^^^^^^^^^^^^^\n  File \"/usr/local/lib/python3.11/site-packages/pulp_rpm/app/tasks/synchronizing.py\", line 569, in synchronize\n    repo_version = dv.create() or repo.latest_version()\n                   ^^^^^^^^^^^\n  File \"/usr/local/lib/python3.11/site-packages/pulpcore/plugin/stages/declarative_version.py\", line 157, in create\n    with self.repository.new_version() as new_version:\n  File \"/usr/local/lib/python3.11/site-packages/pulpcore/app/models/repository.py\", line 1354, in __exit__\n    repository.finalize_new_version(self)\n  File \"/usr/local/lib/python3.11/site-packages/pulp_rpm/app/models/repository.py\", line 426, in finalize_new_version\n    validate_duplicate_content(new_version)\n  File \"/usr/local/lib/python3.11/site-packages/pulpcore/plugin/repo_version_utils.py\", line 107, in validate_duplicate_content\n    raise ValueError(\n",
    "description": "Cannot create repository version. More than one rpm.package content with the duplicate values for name, epoch, version, release, arch, location_href."
  },
```

Unfortunate the task does not mention which packages are causing this conflict.

**Describe the solution you'd like**
The task status report (or somewhere else in the pulp logging) should show the duplicate values causing the error. This would help the user to identify the packaging causing the issue, and maybe resolve it.

**Describe alternatives you've considered**
Now i used a python script to download all the metadata to analyze this issue, which listed me the following conflicts

```
name,epoch,version,release,arch,location_hrefs
grafana,0,10.0.10,1,aarch64,"['Packages/grafana-10.0.10-1.aarch64.rpm', 'Packages/grafana-10.0.10-1.armhfp.rpm']"
grafana,0,v10.0.4,1,armhf,"['Packages/grafana-10.0.4-1.arm.rpm', 'Packages/grafana-10.0.4-1.armhfp.rpm']"
[...]

PR #7280, opened 2026-02-04
[PULP-1118] Add better error handling for repover duplicate content

pedro-psb:
Added a proper error class for duplicate content handling and some more logging to inform exactly what are the conflicting content.

When duplicates are detected, we do some extra work to collect duplicate content.
A simple performance test shows it's not *too bad*:

```python
In [1]: import pulpcore.plugin.repo_version_utils as ut

In [2]: content_qs = Package.objects.all()

In [3]: unique_keys = Package.repo_key_fields

In [4]: content_qs.count()
Out[4]: 24547

In [5]: ut.count_duplicates(content_qs, unique_keys)
Out[5]: 7701

In [6]: %timeit ut.count_duplicates(content_qs, unique_keys)
35.1 ms ± 154 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [7]: %timeit ut.collect_duplicates(content_qs, unique_keys)
61.4 ms ± 266 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Closes: #7184

### 📜 Checklist

### DIFF
--- a/pulpcore/plugin/repo_version_utils.py
+++ b/pulpcore/plugin/repo_version_utils.py
@@ -6,6 +6,10 @@ from django.db.models import Q
 from pulpcore.app.files import validate_file_paths
 from pulpcore.app.models import Content, ContentArtifact
 from pulpcore.app.util import batch_qs
+from pulpcore.exceptions import DuplicateContentInRepositoryError
+from collections import defaultdict
+from django_guid import get_guid
+from typing import NamedTuple
 
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


pulpcore/exceptions/base.py
```diff
@@ -315,3 +316,20 @@ class ReplicateError(PulpException):
...
         return f"[{self.error_code}] " + _("Replication failed")
+
+
+class RepositoryVersionCreateError(PulpException):
+    """
+    Raised when duplicate content is detected on repository version creation.
+    """
+
+    def __init__(self, duplicate_count: int, correlation_id: str):
+        self.dup_count = duplicate_count
```
dralley: It would seem like this assumes that a `RepositoryVersion` creation failed due to duplicates always?  Do we want to assume that?  Should the error be more specific?

pedro-psb: Yeah, makes sense. I'll make more specific one

pulpcore/plugin/repo_version_utils.py
```diff
@@ -78,35 +83,63 @@ def validate_duplicate_content(version):
...
+
+class DuplicateEntry(NamedTuple):
+    keyset_value: tuple[str, ...]
+    duplicate_pks: list[UUID]
+
+
+def log_duplicate(pulp_type: str, duplicate: DuplicateEntry):
+    keyset_value = duplicate.keyset_value
+    duplicate_pks = duplicate.duplicate_pks
+    _logger.info(f"Duplicates found: {pulp_type=}; {keyset_value=}; {duplicate_pks=}")
```
dralley: Is there are particular reason to separate this into its own function if it's only used in one place?

pulpcore/plugin/repo_version_utils.py
```diff
@@ -78,35 +83,63 @@ def validate_duplicate_content(version):
...
+
+
+def count_duplicates(content_qs, unique_keys: tuple[str]) -> int:
+    new_content_total = content_qs.count()
+    unique_new_content_total = content_qs.distinct(*unique_keys).count()
+    return new_content_total - unique_new_content_total
```
dralley, line 117: same

dralley: This case is at least more sensible than the other one though, since there's more than one line that's actually doing something

pulpcore/plugin/repo_version_utils.py
pedro-psb: Just for the abstraction. At a glance, it's trivial do read the main validate function and understand what it does, and dig into the functions if you are interested in implementation details.
If the call cost is not relevant (i believe it isnt here), then it is just a matter of style. I can undo that, if you prefer..

dralley: It's just a matter of style, but if only one line is doing work here, I'd rather just inline it personally

I do not consider it a blocker if you have strong feelings about it or if there are plans to re-use it in the future, though.

pedro-psb: Ok, I'll inline the log one. I'll keep the count and collect duplicates because it was at least convenient for testing them in isolation.

pulpcore/plugin/repo_version_utils.py
```diff
@@ -78,35 +83,59 @@ def validate_duplicate_content(version):
...
+        content_qs = type_obj.objects.filter(pk__in=version.content.filter(pulp_type=pulp_type))
+        dup_count = count_duplicates(content_qs, unique_keys)
+        if dup_count > 0:
+            # At this point the task already failed, so we'll pay extra queries
+            # to collect duplicates and provide more useful logs
+            for duplicate in collect_duplicates(content_qs, unique_keys):
+                keyset_value = duplicate.keyset_value
+                duplicate_pks = duplicate.duplicate_pks
+                _logger.info(f"Duplicates found: {pulp_type=}; {keyset_value=}; {duplicate_pks=}")
+    if dup_count > 0:
```
mdellweg: I think the logic is messed up a bit.
If you have two types in the loop and the first one has duplicates the second is fine, this will not raise.

pedro-psb: That should do it.

mdellweg: Now i think it reports on the wrong content types, once any content type reported a duplicate.
I believe you want to assign `dup_count` fresh in each iteration and collect a boolean `dup_failed` that you can `dup_fail |= dup_count > 0`.

mdellweg: I think, now it should work as designed.

dralley (approved):
Looks reasonable

Issue #7210, opened 2026-01-14
Add a Redis based worker to the tasking system


dkliban:
**Is your feature request related to a problem? Please describe.**
We would like to be able to run 500+ workers at the same time. The current worker implementation limits us to about 150 before the database starts slowing down. 

**Describe the solution you'd like**
Use Redis to store resource locks. Stop running the unblocking algorithm. 

The goal is to introduce a `WORKER_TYPE` setting which will default to `pulpcore` (current implementation). Users will be able to select the new worker type by selecting `redis`.

PR #7296, opened 2026-02-08
Adds 'redis' WORKER_TYPE

dkliban:
This adds WORKER_TYPE setting. The default value is 'pulpcore'. When 'redis' is selected, the tasking system uses Redis to lock resources. Redis workers produce less load on the PostgreSQL database.

closes: #7210

Generated By: Claude Code.

### 📜 Checklist


pulpcore/tasking/redis_locks.py
```diff
@@ -0,0 +1,476 @@
...
+        exclusive_resources (list): List of exclusive resource names
+        shared_resources (list): List of shared resource names
+
+    Returns:
+        list: Empty list if all locks acquired successfully,
+              list of blocked resource names if acquisition failed
+              (includes "__task_lock__" if task lock is held by another worker)
+    """
+    if not redis_conn:
+        return []
```
gerrod3: This seems wrong, can't find redis so we turn into a free-for-all.

pulpcore/tasking/redis_worker.py
```diff
@@ -0,0 +1,783 @@
...
+                blocked_resource_list = acquire_locks(
+                    self.redis_conn, self.name, task_lock_key, exclusive_resources, shared_resources
+                )
+
+                if not blocked_resource_list:
+                    # All locks acquired successfully (task lock + resource locks)!
+                    return task
+                else:
+                    # Failed to acquire locks (task lock or resource locks blocked)
+                    # No cleanup needed - Lua script is all-or-nothing
```
gerrod3: Is this true?

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,592 @@
...
+            _logger.info(
+                "TASK EXECUTION: Task %s being executed by %s (app_type=%s)",
+                task.pk,
+                current_app.name,
+                current_app.app_type,
+            )
+        else:
+            _logger.info(
+                "TASK EXECUTION: Task %s being executed with no AppStatus.current()", task.pk
+            )
```
gerrod3: Is this needed? Can this be moved to `log_task_start`? The value should be set on the task object after `set_running` is called.

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,592 @@
...
+                return None
+            else:
+                task.set_completed(result)
+                log_task_completed(task, domain)
+                send_task_notification(task)
+                return result
+    finally:
+        # Safety net: if we crashed before reaching the lock release above,
+        # still try to release locks here (e.g., if crash during task execution)
+        if safe_release_task_locks(task):
```
gerrod3, line 182: I'm pretty sure this is wrong. `finally` **always** runs regardless of exception or returning early.

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,592 @@
...
+            cancel_task(task_id)
+        except RuntimeError:
+            pass
+    return task_group
+
+
+def execute_task(task):
+    """Redis-aware task execution that releases Redis locks for immediate tasks."""
+    # This extra stack is needed to isolate the current_task ContextVar
+    contextvars.copy_context().run(_execute_task, task)
```
gerrod3: Reading through this version and the base version there is nothing much different between the two besides that this one calls `safe_release_task_locks`. Could this be a wrapper of the original with a `try/finally` to release the redis locks?

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,592 @@
...
+                    "'%s', releasing locks without execution",
+                    task.pk,
+                    task.state,
+                )
+                safe_release_task_locks(task)
+                return task
+
+            # Task is still WAITING, proceed with execution
+            current_app = AppStatus.objects.current()
+            lock_owner = current_app.name if current_app else f"immediate-{task.pk}"
```
gerrod3: Probably same question as before in `_execute_task`, but when is this ever `None`?

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,592 @@
...
+                    " executing in API process (AppStatus=%s)",
+                    task.pk,
+                    lock_owner,
+                )
+                with using_workdir():
+                    execute_task(task)
+            except Exception:
+                # Exception during execute_task()
+                # Atomically release all locks as safety net
+                safe_release_task_locks(task, lock_owner)
```
gerrod3, line 362: Not needed, `execute_task` should already handle letting go of the locks.

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,592 @@
...
+    _, resources = get_resources(exclusive_resources, shared_resources, immediate)
+    app_lock = None if not execute_now else AppStatus.objects.current()  # Lazy evaluation...
+    task_payload = get_task_payload(
+        function_name, task_group, args, kwargs, resources, versions, immediate, deferred, app_lock
+    )
+    task = Task.objects.create(**task_payload)
+    if execute_now:
+        # Try to atomically acquire task lock and resource locks
+        # are_resources_available() now acquires ALL locks atomically
+        if are_resources_available(task):
```
gerrod3: Crazy idea, not sure if you want to try it: but how about trying to acquire the redis locks before the task hits the DB? 

In the default worker scenario, dispatch acquires the lock on the task on creation because `app_lock`  is set to the current task dispatcher (usually the API worker) and the task worker's `fetch_task` can only select from tasks that have this field be null. 

The new redis worker selects from any task that is waiting and thus there is this time window between the the task object hitting the DB and line 470's `are_resources_available` that you have to account for inside dispatch. Instead we can  acquire the task lock first, then create the task, then try to acquire the task's needed resources locks, if successful execute, else defer and finally do a safe release of the task lock. This way `dispatch` shouldn't be fighting against task workers to get the task lock.

```suggestion
   # note that the pulp_id is set once the object is instantiated even if not saved to the DB yet!
    task = Task(**task_payload)
    # new function to just acquire the lock on the task
    aquire_lock(task.pulp_id)
    task.save()
    if execute_now:
        # Change this function to only get task's resources since we already hold the task lock
        if are_resources_available(resources):
            # now guarenteed to have task + resource locks
```

pulpcore/tasking/redis_tasks.py
dkliban: This is not needed. This was added when I was doing some debugging.  I am going to remove this logging.

pulpcore/tasking/redis_tasks.py
dkliban: The comment is not right, but the finally block is still needed in certain exception scenarios. I'll update the comment to reflect this.

pulpcore/tasking/redis_tasks.py
dkliban: It should never be None. Once again this is remains of code left over from debugging initial issues.

pulpcore/tasking/redis_tasks.py
dkliban: that is correct. i'll update the comment to more accurately state that the except block is for the case where using_workdir() fails before the execute function gets a chance to run and release locks itself.

pulpcore/tasking/redis_tasks.py
dkliban: i like this idea. I would be interested in implementing it as a follow up PR.

pulpcore/tasking/redis_locks.py
```diff
@@ -0,0 +1,461 @@
...
+end
+
+-- Release shared locks
+-- Shared keys start at KEYS[2 + num_exclusive]
+for i = num_exclusive + 1, #KEYS - 1 do
+    local key = KEYS[1 + i]
+    local resource_name = ARGV[2 + i]
+
+    -- Remove from set
+    local removed = redis.call("srem", key, lock_owner)
```
gerrod3, line 134: This call can fail if the item at `key` is no longer a set, i.e. is now a string for an exclusive lock.

pulpcore/tasking/redis_locks.py
```diff
@@ -0,0 +1,461 @@
...
+
+    redis_conn = get_redis_connection()
+
+    # Extract resources from task
+    exclusive_resources, shared_resources = extract_task_resources(task)
+
+    # Determine lock owner
+    if lock_owner is None:
+        current_app = AppStatus.objects.current()
+        lock_owner = current_app.name if current_app else f"immediate-{task.pk}"
```
gerrod3, line 240: Is this ever set to `immediate-{task.pk}`? Shouldn't `AppStatus.objects.current()` always return an object?

pulpcore/tasking/redis_locks.py
```diff
@@ -0,0 +1,461 @@
...
+
+    redis_conn = get_redis_connection()
+
+    # Extract resources from task
+    exclusive_resources, shared_resources = extract_task_resources(task)
+
+    # Determine lock owner
+    if lock_owner is None:
+        current_app = await sync_to_async(AppStatus.objects.current)()
+        lock_owner = current_app.name if current_app else f"immediate-{task.pk}"
```
gerrod3, line 288: Same here.

pulpcore/tasking/redis_locks.py
```diff
@@ -0,0 +1,461 @@
...
+    try:
+        blocked_resources = acquire_script(keys=keys, args=args)
+        # Redis returns list of blocked resources or empty list
+        return blocked_resources if blocked_resources else []
+    except Exception as e:
+        _logger.error("Error acquiring locks: %s", e)
+        return ["error"]  # Return non-empty list to indicate failure
+
+
+def release_resource_locks(redis_conn, lock_owner, task_lock_key, resources, shared_resources=None):
```
gerrod3:
```suggestion
def release_resource_locks(redis_conn, lock_owner, task_lock_key, resources=None, shared_resources=None):
```

pulpcore/tasking/redis_locks.py
```diff
@@ -0,0 +1,461 @@
...
+
+        # Log debug for successful releases
+        num_released_exclusive = len(exclusive_resources) - len(not_owned_exclusive)
+        num_released_shared = len(shared_resources) - len(not_in_shared)
+        if num_released_exclusive > 0:
+            _logger.debug("Released %d exclusive lock(s)", num_released_exclusive)
+        if num_released_shared > 0:
+            _logger.debug("Released %d shared lock(s)", num_released_shared)
+        if not task_lock_not_owned:
+            _logger.debug("Released task lock %s", task_lock_key)
```
gerrod3: Do we still need these debug logs?

pulpcore/tasking/redis_locks.py
```diff
@@ -0,0 +1,461 @@
...
+        task_lock_key (str): Redis key for the task lock (e.g., "task:{task_id}")
+        exclusive_resources (list): List of exclusive resource names
+        shared_resources (list): List of shared resource names
+
+    Returns:
+        list: Empty list if all locks acquired successfully,
+              list of blocked resource names if acquisition failed
+              (includes "__task_lock__" if task lock is held by another worker)
+    """
+    # Sort resources deterministically to prevent deadlocks
```
gerrod3: Not sure this comment is correct. It shouldn't matter if the locks are sorted if the redis scripts all run sequentially.

pulpcore/tasking/redis_locks.py
```diff
@@ -0,0 +1,461 @@
...
+    await async_release_resource_locks(
+        redis_conn, lock_owner, task_lock_key, exclusive_resources, shared_resources
+    )
+
+    # Mark all locks as released
+    task._all_locks_released = True
+    return True
+
+
+def acquire_locks(redis_conn, lock_owner, task_lock_key, exclusive_resources, shared_resources):
```
gerrod3, line 303:
```suggestion
def acquire_locks(redis_conn, lock_owner, task_lock_key, exclusive_resources=None, shared_resources=None):
```

pulpcore/tasking/redis_locks.py
```diff
@@ -0,0 +1,461 @@
...
+
+    # Register and execute the Lua script
+    acquire_script = redis_conn.register_script(REDIS_ACQUIRE_LOCKS_SCRIPT)
+    try:
+        blocked_resources = acquire_script(keys=keys, args=args)
+        # Redis returns list of blocked resources or empty list
+        return blocked_resources if blocked_resources else []
+    except Exception as e:
+        _logger.error("Error acquiring locks: %s", e)
+        return ["error"]  # Return non-empty list to indicate failure
```
gerrod3: Should we really just sallow the exception? I think we should raise it.

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,563 @@
...
+    try:
+        # Publish to the task-specific cancellation channel
+        cancel_key = f"{REDIS_CANCEL_PREFIX}{task_id}"
+        # Set a value with expiration (24 hours) in case worker missed it
+        redis_conn.setex(cancel_key, 86400, "cancel")
+        _logger.info("Published cancellation signal for task %s", task_id)
+        return True
+    except Exception as e:
+        _logger.error("Error publishing cancellation signal for task %s: %s", task_id, e)
+        return False
```
gerrod3: We should either catch a more specific Exception or we raise the exception.

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,563 @@
...
+
+def _execute_task(task):
+    try:
+        with with_task_context(task):
+            task.set_running()
+            domain = get_domain()
+            exception_info = None
+            result = None
+
+            # Execute task and capture result or exception
```
gerrod3:
```suggestion
```

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,563 @@
...
+                task_function = get_task_function(task)
+                result = task_function()
+            except Exception:
+                exc_type, exc, tb = sys.exc_info()
+                exception_info = (exc_type, exc, tb)
+
+            # Release locks BEFORE transitioning to final state
+            # This ensures resources are freed even if we crash
+            # during state transition
+            safe_release_task_locks(task)
```
gerrod3: This isn't needed if we are going to keep the `finally` below. See my earlier comment about just wrapping the normal `_execute_task` in a `try/finally` block.

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,563 @@
...
+    Returns:
+        bool: True if all locks were acquired, False otherwise.
+    """
+    redis_conn = get_redis_connection()
+
+    # Extract resources from task
+    exclusive_resources, shared_resources = extract_task_resources(task)
+
+    # Use AppStatus.current() to get a worker identifier for the lock value
+    # For immediate tasks, we use a special identifier
```
gerrod3: Same comment about immediate tasks.

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,563 @@
...
+            _logger.debug(
+                "Failed to acquire locks for immediate task %s (blocked: %s)",
+                task.pk,
+                blocked_resource_list,
+            )
+            return False
+
+    except Exception as e:
+        _logger.error("Error acquiring locks for immediate task %s: %s", task.pk, e)
+        return False
```
gerrod3: Another broad Exception that I think we should raise or be more specific about.

pulpcore/tasking/redis_tasks.py
gerrod3: Let's remove it then.

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,563 @@
...
+            # Task is still WAITING, proceed with execution
+            current_app = AppStatus.objects.current()
+            lock_owner = current_app.name if current_app else f"immediate-{task.pk}"
+            try:
+                _logger.info(
+                    "IMMEDIATE DISPATCH: Task %s acquired all locks,"
+                    " executing in API process (AppStatus=%s)",
+                    task.pk,
+                    lock_owner,
+                )
```
gerrod3:
```suggestion
```

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,563 @@
...
+                    "IMMEDIATE DISPATCH: Task %s already in state "
+                    "'%s', releasing locks without execution",
+                    task.pk,
+                    task.state,
+                )
+                safe_release_task_locks(task)
+                return task
+
+            # Task is still WAITING, proceed with execution
+            current_app = AppStatus.objects.current()
```
gerrod3: `current_app == app_lock`, line 433.
```suggestion
            current_app = app_lock
```

pulpcore/tasking/redis_tasks.py
gerrod3: Can you create a `redis_using_workdir(task, app_lock)` that handles this in a `finally` block?

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,563 @@
...
+                # Release locks if using_workdir() failed before
+                # execute_task() had a chance to run and release them
+                safe_release_task_locks(task, lock_owner)
+                raise
+        elif deferred:
+            # Locks not available, defer to worker
+            # No locks were acquired (atomic operation failed), so nothing to clean up
+            _logger.info(
+                "IMMEDIATE DISPATCH: Task %s could not acquire locks, deferring to worker", task.pk
+            )
```
gerrod3:
```suggestion
            task.app_lock = None
            task.save()
```

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,563 @@
...
+        else:
+            # Can't acquire locks and can't be deferred - cancel task
+            # No locks were acquired, so just set state
+            task.set_canceling()
+            task.set_canceled(TASK_STATES.CANCELED, "Resources temporarily unavailable.")
```
gerrod3, line 374: See comment about reverting `app_lock` logic to default, else this could fail since a task worker could run the task before we could cancel it.

pulpcore/tasking/redis_worker.py
```diff
@@ -0,0 +1,802 @@
...
+    def handle_redis_heartbeat(self):
+        """
+        Check Redis connection health.
+
+        If the check fails, the worker is shut down, similar to how a PostgreSQL
+        heartbeat failure triggers shutdown in handle_worker_heartbeat().
+        """
+        try:
+            self.redis_conn.ping()
+        except Exception:
```
gerrod3: Do you know what type of Exception we could get here?

pulpcore/tasking/redis_worker.py
```diff
@@ -0,0 +1,802 @@
...
+    4. Process resources in deterministic (sorted) order to prevent deadlocks
+    5. Lock values contain worker names to enable cleanup of stale locks
+
+    Note: This implementation does NOT support task cancellation.
+    """
+
+    def __init__(self):
+        # Notification states from signal handlers
+        self.shutdown_requested = False
+        self.wakeup_handle = False
```
gerrod3: I don't think this is used anymore.
```suggestion
```

pulpcore/tasking/redis_worker.py
```diff
@@ -0,0 +1,802 @@
...
+                        # If this task wanted exclusive access,
+                        # block exclusive access for later tasks
+                        for resource in exclusive_resources:
+                            if resource in blocked_resource_list:
+                                blocked_exclusive.add(resource)
+                        # If this task wanted shared access, block shared access for later tasks
+                        # (only if exclusive lock exists, which is why it failed)
+                        for resource in shared_resources:
+                            if resource in blocked_resource_list:
+                                blocked_shared.add(resource)
```
gerrod3: This behavior diverges from how the postgres resource locks work. We need to add all the blocked task's resources into the blocked lists regardless if those resources are currently blocked or not. An example:

| Task | status | exclusive | shared |
| ------ | --------- | ------------- | ---------- |
| T1 | Running | A1 | |
|T2 | Waiting | B1 | A1|
|T3 | Waiting | B1 | |

In this example T3's resources are available, but it shouldn't run because T2 comes first and requested the same B1 resource! Note that this will also change the skip behavior above, you need to add that task's resources to the list before moving on.

pulpcore/tasking/redis_worker.py
```diff
@@ -0,0 +1,802 @@
...
+                    continue
+
+            except Exception as e:
+                _logger.error("Error processing task %s: %s", task.pk, e)
+                continue
```
gerrod3, line 514: Again this exception should be more specific or removed.

pulpcore/tasking/redis_worker.py
```diff
@@ -0,0 +1,802 @@
...
+        """Call and supervise the immediate async task process.
+
+        This function must only be called while holding the lock for that task."""
+        self.task = task
+        _logger.info(
+            "WORKER IMMEDIATE EXECUTION: Worker %s executing immediate task %s in domain: %s",
+            self.name,
+            task.pk,
+            task.pulp_domain.name,
+        )
```
gerrod3, line 529:
```suggestion
```

pulpcore/tasking/redis_worker.py
```diff
@@ -0,0 +1,802 @@
...
+        self.task = task
+        cancel_state = None
+        cancel_reason = None
+        domain = task.pulp_domain
+        _logger.info(
+            "WORKER DEFERRED EXECUTION: Worker %s executing deferred task %s in domain: %s",
+            self.name,
+            task.pk,
+            domain.name,
+        )
```
gerrod3:
```suggestion
```

pulpcore/tasking/redis_worker.py
```diff
@@ -0,0 +1,802 @@
...
+
+        # Set the metric value
+        self.waiting_tasks_meter.set(waiting_tasks)
+
+        _logger.debug(
+            "Waiting tasks metric: %d tasks (%d total tasks older than 5s - %d workers)",
+            waiting_tasks,
+            task_count,
+            self.num_workers,
+        )
```
gerrod3:
```suggestion
```

pulpcore/tasking/redis_worker.py
```diff
@@ -0,0 +1,802 @@
...
+    @exclusive(TASK_METRICS_LOCK)
+    def record_waiting_tasks_metric(self):
+        """
+        Record metrics for waiting tasks in the queue.
+
+        This method counts all tasks in RUNNING or WAITING state that are older
+        than 5 seconds, then subtracts the number of active workers to get the
+        number of tasks waiting to be picked up by workers.
+        """
+        # Calculate the cutoff time (5 seconds ago)
```
gerrod3: Can you remove the inline comments from this method? The code should speak for itself.

pulpcore/tasking/redis_worker.py
```diff
@@ -0,0 +1,802 @@
...
+        Uses a Lua script to ensure we only release locks that we own.
+
+        Args:
+            task_lock_key (str): Redis key for the task lock (e.g., "task:{task_id}")
+            resources (list): List of exclusive resource names to release locks for
+            shared_resources (list): Optional list of shared resource names
+        """
+        release_resource_locks(
+            self.redis_conn, self.name, task_lock_key, resources, shared_resources
+        )
```
gerrod3: Remove this method, it doesn't do anything meaningful.

pulpcore/tasking/redis_worker.py
```diff
@@ -0,0 +1,802 @@
...
+        3. If resource locks acquired, attempts to claim the task
+           with a Redis task lock (24h expiration)
+        4. Returns the first task for which both locks can be acquired
+
+        Returns:
+            Task: A task object if one was successfully locked, None otherwise
+        """
+        # Query waiting tasks, sorted by creation time, limited
+        waiting_tasks = (
+            Task.objects.filter(state=TASK_STATES.WAITING)
```
gerrod3:
```suggestion
            Task.objects.filter(state=TASK_STATES.WAITING, app_lock=None)
```
Ok thinking about it some more, this is super important! We need to change the dispatch and `set_running` code that handles `app_lock` back to how the normal worker does it. So much of our task logic depends on this and changing it is a fools errand. `cleanup_redis_locks_for_worker` is completely inoperable without `app_lock` being correct!

pulpcore/tasking/redis_worker.py
```diff
@@ -0,0 +1,802 @@
...
+    def cleanup_ignored_tasks(self):
+        """Remove tasks from ignored list that are no longer incomplete."""
+        for pk in (
+            Task.objects.filter(pk__in=self.ignored_task_ids)
+            .exclude(state__in=TASK_INCOMPLETE_STATES)
+            .values_list("pk", flat=True)
+        ):
+            self.ignored_task_ids.remove(pk)
+
+    def cleanup_redis_locks_for_worker(self, app_worker):
```
gerrod3, line 238: See my comment about fixing the `app_lock` logic. This method currently can't be reasoned about or expected to work correctly without `app_lock` behaving exactly like the normal pulpcore worker.

pulpcore/tasking/redis_tasks.py
gerrod3: Ok instead of pre-aquiring the task lock before it hits the DB we can just do what the normal worker does and have `fetch_task` only select tasks that have `app_lock=None`. Then we can remove all this task refreshing logic because there won't be any worker able to grab the lock before us.

dkliban: @gerrod3 The new commit makes this exact change. thanks for the suggestion.

pulpcore/tasking/redis_worker.py
dkliban: The cleanup_redis_locks_for_worker method actually falls back to look through redis if it doesn't find anything with an app_lock of that worker. However, i made the change to require app_lock be always present on a task claimed by a worker.

pulpcore/tasking/redis_worker.py
dkliban: i made the app_lock change.

pulpcore/tasking/redis_tasks.py
dkliban: i removed the conditional. it now always uses the name of the current app.

pulpcore/tasking/redis_tasks.py
dkliban: redis_tasks._execute_task and tasks._execute_task share the same core logic:
 set_running → log_task_start → run the task function → log result → set_completed/set_failed →
 send notification. The Redis version cannot directly call the PulpcoreWorker's _execute_task
 because it must release Redis locks between task execution and state transition — a step that
 doesn't exist in the PulpcoreWorker implementation.

pulpcore/tasking/redis_locks.py
dkliban: yep ... i removed the if else

pulpcore/tasking/redis_locks.py
dkliban: yep

pulpcore/tasking/redis_locks.py
dkliban: We can raise it here and it will get logged in are_resources_available

pulpcore/tasking/redis_tasks.py
dkliban: Let's catch RedisError

pulpcore/tasking/redis_tasks.py
dkliban: Let's catch only a RedisError

pulpcore/tasking/redis_worker.py
dkliban: RedisError

pulpcore/tasking/redis_locks.py
dkliban: Let's remove them

pulpcore/tasking/redis_worker.py
dkliban: i will catch the RedisError

pulpcore/tasking/redis_worker.py
dkliban: Great observation! I will update the algorithm.

pulpcore/tasking/redis_locks.py
dkliban: This would only happen in a situation where a worker is considered dead and another worker cleans up it's locks and then the dead worker tries to clean up it's own locks.

pulpcore/tasking/redis_worker.py
```diff
@@ -506,8 +506,9 @@ def fetch_task(self):
             Task: A task object if one was successfully locked, None otherwise
         """
         # Query waiting tasks, sorted by creation time, limited
+        # Filter for app_lock__isnull=True to avoid tasks being executed immediately by API
         waiting_tasks = (
-            Task.objects.filter(state=TASK_STATES.WAITING)
+            Task.objects.filter(state=TASK_STATES.WAITING, app_lock__isnull=True)
```
gerrod3: Nit: use `app_lock=None`

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,430 @@
...
+        function_name, task_group, args, kwargs, resources, versions, immediate, deferred, app_lock
+    )
+    task = await Task.objects.acreate(**task_payload)
+    if execute_now:
+        # Try to atomically acquire task lock and resource locks
+        # async_are_resources_available() now acquires ALL locks atomically
+        if await async_are_resources_available(task):
+            # All locks acquired successfully
+            # Proceed with execution
+            current_app = await sync_to_async(AppStatus.objects.current)()
```
gerrod3:
```suggestion
            current_app = app_lock
```

pulpcore/tasking/redis_tasks.py
```diff
@@ -0,0 +1,430 @@
...
+    Returns:
+        bool: True if all locks were acquired, False otherwise.
+    """
+    redis_conn = get_redis_connection()
+
+    # Extract resources from task
+    exclusive_resources, shared_resources = extract_task_resources(task)
+
+    current_app = AppStatus.objects.current()
+    lock_owner = current_app.name
```
gerrod3: We already did this query right before calling this method, it should be set on the task's `app_lock`. Do we have to do it again?

pulpcore/tasking/redis_worker.py
```diff
@@ -0,0 +1,729 @@
...
+
+                if should_skip:
+                    continue
+
+                # Claim the task by setting app_lock BEFORE acquiring locks
+                rows = Task.objects.filter(
+                    pk=task.pk, state=TASK_STATES.WAITING, app_lock__isnull=True
+                ).update(app_lock=self.app_status)
+
+                if rows == 0:
```
gerrod3: I'm pretty sure this is going to cause a lot of database contention. How about doing redis locks first, then updating the database?

gerrod3 (changes requested):

pulpcore/tasking/redis_worker.py
dkliban: I agree

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -487,6 +541,106 @@ def test_timeouts_on_api_worker(self, pulpcore_bindings, dispatch_task):
         assert "timed out after" in task.error["description"]
 
 
+@pytest.mark.parallel
+def test_immediate_task_execution_in_worker(dispatch_task, monitor_task):
```
gerrod3: These following three tests seem pointless, let's remove them.

pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -510,6 +664,45 @@ def _resource_blocker(exclusive_resources: list[str], duration=20):
     return _resource_blocker
 
 
+@pytest.mark.parallel
+def test_immediate_task_with_available_resources(dispatch_task, pulpcore_bindings):
```
gerrod3, line 618: Same with this test, it is already tested below, let's remove.

pulpcore/tasking/redis_worker.py
```diff
@@ -0,0 +1,722 @@
...
+    Worker implementation using Redis distributed lock-based resource acquisition.
+
+    This worker uses a simpler algorithm where:
+    1. Query waiting tasks (sorted by creation time, limited)
+    2. For each task, try to acquire Redis distributed locks for all resources
+    3. If all locks acquired, claim the task
+    4. Process resources in deterministic (sorted) order to prevent deadlocks
+    5. Lock values contain worker names to enable cleanup of stale locks
+
+    Note: This implementation does NOT support task cancellation.
```
gerrod3: Outdated comment?

pulpcore/tests/functional/api/test_tasking.py
dkliban: This test specifically exercises the are_resources_available() / acquire_locks() code path that the no-resource test skips entirely.

gerrod3 (approved):
I believe this is ready. Thanks for all the work!

PR #7311, opened 2026-02-11
Harden RepositoryVersion against memoization issues

dralley:
### 📜 Checklist

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -1068,6 +1068,7 @@ class RepositoryVersion(BaseModel):
                 complete RepositoryVersion
         """
 
+        assert issubclass(content.model, Content)
         if self.complete:
             raise ResourceImmutableError(self)
 
@@ -1076,6 +1077,7 @@ class RepositoryVersion(BaseModel):
             .exclude(pulp_domain_id=get_domain_pk())
             .exists()
         )
+
         repo_content = []
         to_add = set(content.values_list("pk", flat=True)) - set(self._get_content_ids())
         with transaction.atomic():
@@ -1114,12 +1116,14 @@ class RepositoryVersion(BaseModel):
                 complete RepositoryVersion
         """
 
+        assert issubclass(content.model, Content)
         if self.complete:
             raise ResourceImmutableError(self)
 
         if not content or not content.count():
             return
 
+        # check that all content is within the current domain
         assert (
             not Content.objects.filter(pk__in=content)
             .exclude(pulp_domain_id=get_domain_pk())
@@ -1157,6 +1161,7 @@ class RepositoryVersion(BaseModel):
             pulpcore.exception.ResourceImmutableError: if set_content is called on a
                 complete RepositoryVersion
         """
+        assert issubclass(content.model, Content)
         self.remove_content(self.content.exclude(pk__in=content))
         self.add_content(content.exclude(pk__in=self.content))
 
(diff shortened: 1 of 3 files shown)


pulpcore/tests/unit/models/test_repository.py
```diff
@@ -261,19 +297,294 @@ def test_next_version_with_multiple_versions():
...
+
+@pytest.mark.django_db
+def test_add_wrong_content_type(db, repository):
+    """Verify that adding content of wrong type is handled."""
+    # Create content with a different pulp_type
+    wrong_type_content = Content.objects.create(pulp_type="wrong.type")
+    correct_type_content = Content.objects.create(pulp_type="core.content")
+
+    # Adding content with wrong type should still work at the model level
+    # (type validation typically happens at a higher level in the API/plugin layer)
```
mdellweg: This comment sounds like there should not be `pytest.raises` .

pulpcore/tests/unit/models/test_repository.py
```diff
@@ -261,19 +297,294 @@ def test_next_version_with_multiple_versions():
...
+def test_add_remove_non_content_type(db, repository):
+    """Verify that adding a queryset of non-Content type raises AssertionError."""
+    # Create a real Content object first
+    content = Content.objects.create(pulp_type="core.content")
+
+    # Create a ContentArtifact which is NOT a Content subclass
+    content_artifact = ContentArtifact.objects.create(content=content, relative_path="test/path")
+
+    # Attempting to add a queryset of ContentArtifact (not Content) should raise AssertionError
+    with pytest.raises(AssertionError):
```
mdellweg, line 470: Type analysis would have caught that too, i think. But we don't have that. So :+1:.

pulpcore/tests/unit/models/test_repository.py
```diff
@@ -261,19 +297,294 @@ def test_next_version_with_multiple_versions():
...
+    with repository.new_version() as version1:
+        version1.add_content(Content.objects.filter(pk__in=content_pks))
+
+    # Perform operations that cancel out
+    with repository.new_version() as version2:
+        # Remove content that doesn't exist (no-op)
+        version2.remove_content(Content.objects.filter(pk=9999999))
+        # Add content that already exists (no-op)
+        version2.add_content(Content.objects.filter(pk=content_pks[0]))
+
```
mdellweg, line 527: Do we also have as similar test, where the single content is actually removed and then added again so still no new version is generated?

dralley: Yes, `test_version_identical_to_previous`

dralley: Well, that tests "adding and then removing" the same content, so I will extend to also cover "removing and then adding" the same content

pulpcore/app/models/repository.py
```diff
@@ -1307,6 +1311,9 @@ def _compute_counts(self):
         objects and makes new ones with each call.
         """
         with transaction.atomic():
+            # relatively inexpensive sanity check for memoization
+            assert len(self.content_ids) == self._content_relationships().count()
```
dralley, line 1315: Worthwhile?

ggainey: So basically this is checking that at the end of squash(), that the 'cached' content_ids field of the repoversion, has the same number of elements as recomputing - so this would be like a last-ditch chance to catch coding-errors in "someone's" new-version-creation code, yeah? 

Even if it's "just" a count() of the recompute-query - is this actually "inexpensive"?  (thinking out loud here) It _would_ only happen once at new-version-squash-time, so that may not be a big issue.

dralley: My thinking here is that it's at least not any more expensive than the operations that already have to happen at the beginning of the repository version create (basically this same query) and at the end, including the annotates below.

dralley: And probably much less so

Issue #7272, opened 2026-02-01
Content set memoization is incorrect in some circumstances


dralley:
**Version**

pulpcore 3.83+

**Describe the bug**

```
In [8]: for rc in RepositoryContent.objects.all().order_by("version_added"):
   ...:     print("package: {:<40} added: {} removed: {}".format(
   ...:         Package.objects.get(pk=rc.content_id).nevra,
   ...:         rc.version_added.number,
   ...:         rc.version_removed.number
   ...:     ))
   ...: 
package: dummy-pkg-0:1.1-1.fc44.src               added: 1 removed: 2
package: dummy-pkg-0:1.1-1.fc44.x86_64            added: 1 removed: 2
package: dummy-pkg-0:1.2-1.fc44.src               added: 1 removed: 2
package: dummy-pkg-0:1.2-1.fc44.x86_64            added: 1 removed: 2
package: dummy-pkg-0:1.3-1.fc44.x86_64            added: 1 removed: 2
package: dummy-pkg-0:1.3-1.fc44.src               added: 1 removed: 2
package: dummy-pkg-0:1.4-1.fc44.src               added: 1 removed: 2
package: dummy-pkg-0:1.4-1.fc44.x86_64            added: 1 removed: 2
package: dummy-pkg-0:1.5-1.fc44.src               added: 1 removed: 2
package: dummy-pkg-0:1.5-1.fc44.x86_64            added: 1 removed: 2
package: dummy-pkg-0:1.6-1.fc44.src               added: 1 removed: 2
package: dummy-pkg-0:1.6-1.fc44.x86_64            added: 1 removed: 2
package: dummy-pkg-0:1.7-1.fc44.x86_64            added: 1 removed: 2
package: dummy-pkg-0:1.7-1.fc44.src               added: 1 removed: 2
package: dummy-pkg-0:1.8-1.fc44.src               added: 1 removed: 2
package: dummy-pkg-0:1.8-1.fc44.x86_64            added: 1 removed: 2
package: dummy-pkg-0:1.9-1.fc44.src               added: 1 removed: 2
package: dummy-pkg-0:1.9-1.fc44.x86_64            added: 1 removed: 2
package: dummy-pkg-0:1.10-1.fc44.x86_64           added: 1 removed: 2
package: dummy-pkg-0:1.10-1.fc44.src              added: 1 removed: 2
package: dummy-pkg-0:1.11-1.fc44.x86_64           added: 1 removed: 2
[...]

PR #7322, opened 2026-02-16
Management command for data repairs

dralley:
Create a single reusable data repair management command for data repair operations happening outside of migrations.

Specifically in this case the motivation is to resolve the repository version memoization issues, but it can be reused in the future.

re #7272

Usage (ignore the warnings):

```
[root@cb0f1f0e9d34 src]# pulpcore-manager datarepair 7272 --dry-run
/usr/local/lib/python3.11/site-packages/django/db/backends/utils.py:98: RuntimeWarning: Accessing the database during app initialization is discouraged. To fix this warning, avoid executing queries in AppConfig.ready() or when your app modules are imported.
  warnings.warn(self.APPS_NOT_READY_WARNING_MSG, category=RuntimeWarning)

In domain "default"
	Repository "test" (type "rpm.rpm") version 2 has a mismatch between the RepositoryContent and the cached ID set
	Repository "test" (type "rpm.rpm") version 2 has a mismatch between the RepositoryContent and RepositoryVersionContentDetails

Finished. (no changes)
[root@cb0f1f0e9d34 src]# pulpcore-manager datarepair 7272
/usr/local/lib/python3.11/site-packages/django/db/backends/utils.py:98: RuntimeWarning: Accessing the database during app initialization is discouraged. To fix this warning, avoid executing queries in AppConfig.ready() or when your app modules are imported.
  warnings.warn(self.APPS_NOT_READY_WARNING_MSG, category=RuntimeWarning)

In domain "default"
	Repository "test" (type "rpm.rpm") version 2 has a mismatch between the RepositoryContent and the cached ID set
	Repository "test" (type "rpm.rpm") version 2 has a mismatch between the RepositoryContent and RepositoryVersionContentDetails

Finished. (1 repository versions fixed)
[root@cb0f1f0e9d34 src]# pulpcore-manager datarepair 7272 --dry-run
/usr/local/lib/python3.11/site-packages/django/db/backends/utils.py:98: RuntimeWarning: Accessing the database during app initialization is discouraged. To fix this warning, avoid executing queries in AppConfig.ready() or when your app modules are imported.
  warnings.warn(self.APPS_NOT_READY_WARNING_MSG, category=RuntimeWarning)


Finished. (OK)
[root@cb0f1f0e9d34 src]# 

```

### DIFF
--- a/pulpcore/app/management/commands/datarepair.py
+++ b/pulpcore/app/management/commands/datarepair.py
@@ -1,0 +1,193 @@
+from gettext import gettext as _
+
+from django.db import connection
+from django.conf import settings
+from django.core.management import BaseCommand, CommandError
+from django.db.models import Q
+from django.utils.encoding import force_bytes, force_str
+
+import cryptography
+
+from pulpcore.app import models
+
+
+class Command(BaseCommand):
+    """
+    Django management command for repairing improper data mistakes and migrations.
+    """
+
+    help = _("Repairs various data corruption issues known to Pulp.")
+
+    def add_arguments(self, parser):
+        """Set up arguments."""
+        parser.add_argument("issue", help=_("The github issue # of the issue to be fixed."))
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help=_(
+                "Don't modify anything, just show the results of what would happen if this "
+                "command were run."
+            ),
+        )
+
+    def handle(self, *args, **options):
+        """Implement the command."""
+        issue = options["issue"]
+
+        if issue == "2327":
+            self.repair_2327(options)
+        elif issue == "7272":
+            self.repair_7272(options)
+        else:
+            raise CommandError(_("Unknown issue: '{}'").format(issue))
+
+    def repair_7272(self, options):
+        dry_run = options["dry_run"]
+
+        number_broken = 0
+        self.stdout.write()
+
+        for domain in models.Domain.objects.all():
+            has_printed_domain = False
+            for repo in models.Repository.objects.filter(pulp_domain=domain):
+                for rv in models.RepositoryVersion.objects.filter(repository=repo):
+                    needs_fix = False
+                    if rv.content_ids is not None:
+                        cached_id_set = set(rv.content_ids)
+                        repositorycontent_id_set = set(
+                            rv._content_relationships().values_list("content__pk", flat=True)
+                        )
+                        if cached_id_set != repositorycontent_id_set:
+                            number_broken += 1
+                            if not has_printed_domain:
+                                self.stdout.write(f'In domain "{domain.name}"')
+                                has_printed_domain = True
+
+                            self.stdout.write(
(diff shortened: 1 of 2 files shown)


pulpcore/app/management/commands/datarepair.py
```diff
@@ -0,0 +1,193 @@
...
+                        rv.content_ids = list(
+                            rv._content_relationships().values_list("content__pk", flat=True)
+                        )
+                        rv.save()
+                        rv._compute_counts()
+
+            print()
+
+        if not number_broken:
+            print("Finished. (OK)")
```
mdellweg: Django want's us to use `self.stdout.write` here.

https://docs.djangoproject.com/en/4.2/howto/custom-management-commands/

pulpcore/app/management/commands/datarepair.py
```diff
@@ -0,0 +1,193 @@
...
+class Command(BaseCommand):
+    """
+    Django management command for repairing improper data mistakes and migrations.
+    """
+
+    help = _("Repairs various data corruption issues known to Pulp.")
+
+    def add_arguments(self, parser):
+        """Set up arguments."""
+        parser.add_argument("issue", help=_("The github issue # of the issue to be fixed."))
```
mdellweg, line 23: Would it be possible for plugins to add datarepair operations, and how would we avoid name clashes?

pulpcore/app/management/commands/datarepair.py
```diff
@@ -0,0 +1,193 @@
...
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help=_(
+                "Don't modify anything, just show the results of what would happen if this "
+                "command were run."
+            ),
```
mdellweg, line 30: Maybe another option. But wouldn't it be nice to print some info paragraph about the selected repair?
It could sum up for example after what type of upgrade the data is expected to need repair (whenever we have that detail).

pulpcore/app/management/commands/datarepair.py
dralley: It might be possible, but some plugins (e.g. rpm) already have their own command (rpm-datarepair) for that purpose.

pulpcore/app/management/commands/datarepair.py
dralley: Do you mean as help text, or as part of the execution of the command, including in dry-run mode?

Generally I would expect the user to know what they're running ahead of time, this isn't really the sort of thing to be running in an exploratory way.  Even, to an extent, in `--dry-run` mode.  Probably the user is coming to rectify some specific issue.

mdellweg: Help text sounds fine. I just thought that maybe assuming we know how an inconsistency arose, we could briefly describe the situation so a user knowing they didn't come that way doesn't put high expectations in this. But maybe I'm in wonderland here.
This sure is not a requirement.

Issue #7315, opened 2026-02-11
Long uploads cause API workers to miss hearbeats


dkliban:
**Version**
1.103.1

**Describe the bug**
In order to support uploads of very large RPMs, we have increased the gunicorn timeouts for the requests. When the request lasts a long time, the API worker misses it's heartbeats and it's AppStatus record ends up getting cleaned up. The next time it tries to write a heartbeat, it fails to do so and restarts.

PR #7331, opened 2026-02-18
Moves API heartbeat to it's own thread.

dkliban:
Assisted By: Claude Code

fixes: #7315

### 📜 Checklist

### DIFF
--- a/pulpcore/app/entrypoint.py
+++ b/pulpcore/app/entrypoint.py
@@ -2,6 +2,8 @@ from contextvars import ContextVar
 from logging import getLogger
 import os
 import sys
+import threading
+import time
 
 import click
 import django
@@ -22,9 +24,29 @@ using_pulp_api_worker = ContextVar("using_pulp_api_worker", default=False)
 
 
 class PulpApiWorker(SyncWorker):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.heartbeat_thread = None
+
+    def _heartbeat_loop(self):
+        """Run heartbeat in a loop. Exit process if heartbeat fails."""
+        try:
+            while self.alive:
+                self.heartbeat()
+                time.sleep(self.heartbeat_interval)
+        except Exception as e:
+            logger.error(f"Heartbeat thread loop failed for worker {self.name}: {e}")
+            os._exit(Arbiter.WORKER_BOOT_ERROR)
+
     def notify(self):
         super().notify()
-        self.heartbeat()
+
+        if not self.heartbeat_thread.is_alive():
+            logger.error(
+                f"Heartbeat thread died unexpectedly for worker {self.name}. "
+                "Exiting to trigger restart."
+            )
+            os._exit(Arbiter.WORKER_BOOT_ERROR)
 
     def heartbeat(self):
         try:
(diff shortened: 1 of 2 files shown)


pulpcore/app/entrypoint.py
```diff
@@ -22,9 +24,18 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.heartbeat_thread = None
+
     def notify(self):
         super().notify()
-        self.heartbeat()
+
+    def _heartbeat_loop(self):
+        while self.alive:
```
mdellweg: What is this for?

pulpcore/app/entrypoint.py
```diff
@@ -22,9 +24,18 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.heartbeat_thread = None
+
     def notify(self):
         super().notify()
-        self.heartbeat()
+
+    def _heartbeat_loop(self):
```
mdellweg: What happens if the thread dies?

pulpcore/app/entrypoint.py
```diff
@@ -22,9 +24,18 @@
 
 
 class PulpApiWorker(SyncWorker):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.heartbeat_thread = None
+
     def notify(self):
```
mdellweg, line 41: First i thought, this is dead code. Now i think we have the opportunity to check in on the thread here.

pulpcore/app/entrypoint.py
```diff
@@ -76,6 +87,13 @@ def init_process(self):
             logger.error(f"An API app with name {self.name} already exists in the database.")
             exit(Arbiter.WORKER_BOOT_ERROR)
 
+        # Store heartbeat interval for the heartbeat thread
+        self.heartbeat_interval = settings.API_APP_TTL // 3
+
+        # Start heartbeat thread before entering the main loop
```
mdellweg: Now why don't we then put all of that stuff right before entering the main loop?

pulpcore/app/entrypoint.py
mdellweg, line 75: Did you never get to see this warning?

pulpcore/app/entrypoint.py
```diff
@@ -76,6 +87,13 @@ def init_process(self):
             logger.error(f"An API app with name {self.name} already exists in the database.")
             exit(Arbiter.WORKER_BOOT_ERROR)
 
+        # Store heartbeat interval for the heartbeat thread
```
mdellweg:
```suggestion
        # Store the value of one third of the desired API application's time to live to the heartbeat interval on the current instance to be accessible by the heartbeat thread.
```

:mild sarcasm:

pulpcore/app/entrypoint.py
dkliban: A gunicorn SyncWorker sets `self.alive` to `False` during graceful shutdown. This ensures that when a graceful shutdown is requested, the thread does not make another heartbeat update. This check is not critical, because the thread will be shut down soon anyway.

pulpcore/app/entrypoint.py
dkliban: I am going to add logic to handle the thread dying.

pulpcore/app/entrypoint.py
dkliban: i've only seen this warning on my dev environment.

pulpcore/app/entrypoint.py
```diff
@@ -22,9 +24,40 @@
...
+            while self.alive:
+                try:
+                    self.heartbeat()
+                except Exception as e:
+                    # Any unhandled exception in heartbeat should restart the worker
+                    logger.error(
+                        f"Heartbeat thread encountered fatal error for worker {self.name}: {e}"
+                    )
+                    # Exit with error code to trigger worker restart
+                    os._exit(Arbiter.WORKER_BOOT_ERROR)
```
mdellweg: I don't think we need this here.

dkliban: I removed the inner exception handling.

PR #7426, opened 2026-03-09
Allow syncing custom repos without password

### LINKED ISSUE #7425: Impossible to sync a custom repository using only the upstream username
**Version**
rubygem-katello-4.20.0-0.4.rc2.el9sat.noarch
python3.12-pulpcore-3.85.9-1.el9pc.noarch

**Describe the bug**
It is not possible to sync a custom repository using only the upstream username.

A custom repository requires only the upstream username to access the content, and it works without any issues when configuring it locally in /etc/yum.repos.d. However, when attempting to sync the same repository in the Satellite WebUI, the sync fails with the error:
```
hammer repository update --id 7488 --upstream-username '7f4fab26704c60bdd7028f5b2c1c4708' --upstream-password ''
    Repository updated.

hammer repository synchronize --id 7488
[.....................................................................................................................................................................................................] [100%]
No content added.
Waiting to start.
Error: 401, message='Unauthorized', url='https://registration-triage.phishme.com/artifactory/triage-rhel9-prod?7f4fab26704c60bdd7028f5b2c1c4708'
```

**To Reproduce**
Try to sync a repo which requires upstream_username but no password.
See https://issues.redhat.com/browse/SAT-41022 for more details.

**Expected behavior**
A successful repo sync with only upstream_username provided (no password).

**Additional context**
https://issues.redhat.com/browse/SAT-41022

vsedmik:
When a custom repository requires upstream_username only (no password), we should be able to proceed with the username only.

Closes #7425

### 📜 Checklist

### DIFF
--- a/pulpcore/download/factory.py
+++ b/pulpcore/download/factory.py
@@ -16,7 +16,6 @@ from pulpcore.app.apps import PulpAppConfig
 from .http import HttpDownloader
 from .file import FileDownloader
 
-
 PROTOCOL_MAP = {
     "http": HttpDownloader,
     "https": HttpDownloader,
@@ -204,10 +203,10 @@ class DownloaderFactory:
                     login=self._remote.proxy_username, password=self._remote.proxy_password
                 )
 
-        if self._remote.username and self._remote.password:
-            options["auth"] = aiohttp.BasicAuth(
-                login=self._remote.username, password=self._remote.password
-            )
+        if self._remote.username:
+            # Support username-only auth with empty passwords
+            password = self._remote.password or str()
+            options["auth"] = aiohttp.BasicAuth(login=self._remote.username, password=password)
 
         kwargs["throttler"] = self._remote.download_throttler if self._remote.rate_limit else None
 
--- a/CHANGES/7425.bugfix
+++ b/CHANGES/7425.bugfix
@@ -1,0 +1,2 @@
+Fixed a bug when syncing a custom repository using the upstream username only (no password)
+did not send any authentication since both were required by the download factory.


ggainey:
The error in the issue is passing '' as the password - can't you just "not set it", or pass None?

pulpcore/download/factory.py
```diff
@@ -204,9 +204,11 @@ class to be instantiated.
                     login=self._remote.proxy_username, password=self._remote.proxy_password
                 )
 
-        if self._remote.username and self._remote.password:
+        if self._remote.username:
+            # Support username-only auth with empty passwords
+            password = self._remote.password or str()
```
ggainey, line 208: When you use hammer to set userpassword as empty-string, what ends up in the Pulp remote? If it's an empty string - how is setting it to str() any different?

vsedmik: Oh, apologize, I probably misunderstood the last comment (among many other things) - I though that hardcoding `''` was the issue for linter, so just replaced it with `str()`.

After bypassing the validation in Katello I created and synced the repo with no pass just ok.
Looking at the remote_href (gotten from the rake console for that repo), it looks it has just some hidden fields related to username/pass:
```
# pulp show --href /pulp/api/v3/remotes/rpm/rpm/019ce1bb-945b-7d49-8bc6-ae9c2ef811d3/
{
  "url": "https://registration-triage.phishme.com/artifactory/triage-rhel9-prod",
  ...
  "hidden_fields": [
     ...
      "name": "username",
      "is_set": true
    },
    {
      "name": "password",
      "is_set": false
    }

```
So it's probably handled from Katello, I can see it on the repo root set to `nil`
```
[3] pry(main)> repo.upstream_username
=> "7f4fab26704c60bdd7028f5b2c1c4708"
[4] pry(main)> repo.upstream_password
=> nil
```
but overriding it to `""` didn't help anyway. /o\

ianballou: So from what I'm seeing here, the Pulp remote can end up with a username and a null password, which is what we want from a user standpoint. But the download factory code here seems to not pass the username along unless there is a username and password both. 

@vsedmik seems to have found that aiohttp doesn't like username/password being `None`. Since empty username/password is legit with http requests, I can only assume that aiohttp thus expects empty string for passing in an empty username or password.

That leaves us with the situation  that Pulp remotes don't allow the username / password to be empty string and expect null instead, but aiohttp expects the opposite.

We need to verify my thought about aiohttp and empty string passwords, but if it's true, then we may need to convert the null credentials into empty strings like what's done here. 

Also - I believe password but no username may also be valid. Needs a check as well.

vsedmik: I haven't tested the "pass but no username" case, but looking at the `BasicAuth` constructor it seems to be possible too
```
class BasicAuth(namedtuple("BasicAuth", ["login", "password", "encoding"])):                                     
      """Http basic authentication helper."""                                                                                         
                                                                                                                                      
      def __new__(                                                                                                                    
          cls, login: str, password: str = "", encoding: str = "latin1"                                                           
      ) -> "BasicAuth":                                                                                                           
          if login is None:                                                                                                       
              raise ValueError("None is not allowed as login value")                                                              
                                                                                                                                  
          if password is None:                                                                                                    
              raise ValueError("None is not allowed as password value")   
                                                                                                                                  
          if ":" in login:                                                                                                        
              raise ValueError('A ":" is not allowed in login (RFC 1945#section-11.1)')                                           
                                                                                                                                  
          return super().__new__(cls, login, password, encoding)   
```

PR #7433, opened 2026-03-09
Replace custom OpenPGP packet parsing with pysequoia 

dralley:
### 📜 Checklist

### DIFF
--- a/functest_requirements.txt
+++ b/functest_requirements.txt
@@ -1,7 +1,7 @@
 pytest<10
 pytest-custom_exit_code
 pytest-xdist
-python-gnupg
+pysequoia
 proxy.py~=2.4.10
 trustme~=1.2.1
 
--- a/pulpcore/app/management/commands/add-signing-service.py
+++ b/pulpcore/app/management/commands/add-signing-service.py
@@ -1,6 +1,5 @@
 import os
-
-import gnupg
+import subprocess
 
 from pathlib import Path
 
@@ -72,13 +71,33 @@ class Command(BaseCommand):
                 )
             )
 
-        gpg = gnupg.GPG(gnupghome=options["gnupghome"], keyring=options["keyring"])
+        gpg_cmd = ["gpg"]
+        if options["gnupghome"]:
+            gpg_cmd += ["--homedir", options["gnupghome"]]
+        if options["keyring"]:
+            gpg_cmd += ["--keyring", options["keyring"]]
 
-        key_list = gpg.list_keys(keys=[key_id])
-        if not len(key_list) == 1:
-            raise CommandError(_("There are {} keys matching the key id.").format(len(key_list)))
-        fingerprint = key_list[0]["fingerprint"]
-        public_key = gpg.export_keys(key_id)
+        result = subprocess.run(
+            gpg_cmd + ["--with-colons", "--fingerprint", key_id],
+            capture_output=True,
+            text=True,
+        )
+        if result.returncode != 0:
+            raise CommandError(result.stderr.strip())
+
+        fpr_lines = [line for line in result.stdout.splitlines() if line.startswith("fpr:")]
+        if len(fpr_lines) != 1:
+            raise CommandError(_("There are {} keys matching the key id.").format(len(fpr_lines)))
+        fingerprint = fpr_lines[0].split(":")[9]
+
+        result = subprocess.run(
+            gpg_cmd + ["--armor", "--export", key_id],
+            capture_output=True,
+            text=True,
+        )
+        if result.returncode != 0:
+            raise CommandError(result.stderr.strip())
+        public_key = result.stdout
 
         try:
             script_path = Path(script).resolve(strict=True)
--- a/pulpcore/app/openpgp.py
+++ b/pulpcore/app/openpgp.py
@@ -1,529 +1,83 @@
 import hashlib
-from base64 import b64decode, b64encode
 
-from django.utils import timezone
-
-# Source of information:
-#   * rfc4880
-#   * rfc4880bis
-#   * gnupg sources
-#   * https://datatracker.ietf.org/doc/html/draft-shaw-openpgp-hkp-00
-
-
-PACKET_TYPES = {
-    1: "Public-Key Encrypted Session Key Packet",
-    2: "Signature Packet",
-    3: "Symmetric-Key Encrypted Session Key Packet",
-    4: "One-Pass Signature Packet",
-    5: "Secret-Key Packet",
-    6: "Public-Key Packet",
-    7: "Secret-Subkey Packet",
-    8: "Compressed Data Packet",
-    9: "Symmetrically Encrypted Data Packet",
-    10: "Marker Packet",
-    11: "Literal Data Packet",
-    12: "Trust Packet",
-    13: "User ID Packet",
-    14: "Public-Subkey Packet",
-    17: "User Attribute Packet",
-    18: "Sym. Encrypted and Integrity Protected Data Packet",
-    19: "Modification Detection Code Packet",
-    20: "OCB Encrypted Data Packet",
-}
-
-
-SIG_SUBPACKAGE_TYPES = {
-    2: "Signature Creation Time",
-    3: "Signature Expiration Time",
-    4: "Exportable Certification",
-    5: "Trust Signature",
-    6: "Regular Expression",
-    7: "Revocable",
-    9: "Key Expiration Time",
-    10: "Placeholder for backward compatibility",
-    11: "Preferred Symmetric Algorithms",
-    12: "Revocation Key",
-    16: "Issuer",
-    20: "Notation Data",
-    21: "Preferred Hash Algorithms",
-    22: "Preferred Compression Algorithms",
-    23: "Key Server Preferences",
-    24: "Preferred Key Server",
-    25: "Primary User ID",
-    26: "Policy URI",
-    27: "Key Flags",
-    28: "Signer's User ID",
-    29: "Reason for Revocation",
-    30: "Features",
-    31: "Signature Target",
-    32: "Embedded Signature",
-    33: "Issuer Fingerprint",
-    34: "Preferred Encryption Modes",
-    35: "Intended Recipient Fingerprint",
-    37: "Attested Certifications",
-    38: "Key Block",
-    40: "Literal Data Meta Hash",
-    41: "Trust Alias",
-}
-
-
-PUBKEY_ALGORITHMS = {
-    1: {
-        "name": "RSA",
-        "format": "mm",
-    },
-    2: {
-        "name": "RSA Encrypt-Only",
-        "format": "mm",
-    },
-    3: {
-        "name": "RSA Sign-Only",
-        "format": "mm",
-    },
-    16: {
-        "name": "Elgamal",
-        "format": "mmm",
-    },
-    17: {
-        "name": "DSA",
-        "format": "mmmm",
-    },
-    18: {
-        "name": "ECDH",
-        "format": "omk",
-    },
-    19: {
-        "name": "ECDSA",
-        "format": "om",
-    },
-    22: {
-        "name": "EdDSA",
-        "format": "om",
-    },
-}
-
-
-HASH_ALGORITHMS = {
-    1: "md5",
-    2: "sha1",
-    3: "ripemd160",
-    8: "sha256",
-    9: "sha384",
-    10: "sha512",
-    11: "sha224",
-    12: "sha3-256",
-    14: "sha3-512",
-}
-
-
-SYMMETRIC_ALGORITHMS = {
-    0: "Plaintext",
-    1: "IDEA",
-    2: "TripleDES",
-    3: "CAST5",
-    4: "Blowfish",
-    7: "AES128",
-    8: "AES192",
-    9: "AES256",
-    10: "Twofish",
-    11: "Camellia 128",
-    12: "Camellia 192",
-    13: "Camellia 256",
-}
-
-
-COMPRESSION_ALGORITHMS = {
-    0: "Uncompressed",
-    1: "ZIP",
-    2: "ZLIB",
-    3: "BZip2",
-}
-
-
-ENCRYPTION_MODES = {
-    1: "EAX",
-    2: "OCB",
-}
-
-
-def packet_iter(data):
-    begin = 0
-    pos = 0
-    while pos < len(data):
-        packet_tag = data[pos]
-        pos += 1
-        if not packet_tag & 0x80:
-            raise ValueError("Invalid Packet Tag")
-        new_format = bool(packet_tag & 0x40)
-
-        if new_format:
-            packet_type = packet_tag & 0x1F
-
-            if data[pos] < 0xC0:
-                # 1-octet length
-                length = data[pos]
-                pos += 1
-            elif data[pos] < 0xE0:
-                # 2-octet length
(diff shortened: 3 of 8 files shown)


pulpcore/app/util.py
```diff
@@ -387,9 +384,45 @@ def get_request_without_query_params(context):
...
+
+    Attributes:
+        valid (bool): Always True; invalid signatures raise InvalidSignatureError instead.
+        fingerprint (str): Fingerprint of the signing subkey (uppercase hex).
+        pubkey_fingerprint (str): Fingerprint of the signing certificate (uppercase hex).
+        key_id (str): Short (16-char) key ID derived from the fingerprint.
+        username (str): Empty string; primary UID not available without extending pysequoia.
+        data (bytes or None): The verified plaintext content for inline signatures, None for
+            detached signatures.
+    """
```
dralley, line 395: I'm not actually sure that we need this, but `gpg_verify` is public API, so I figured we would have to at least try.  Unless we wait until the next breaking change release.

pyproject.toml
```diff
@@ -57,7 +57,8 @@ dependencies = [
   "pygtrie>=2.5,<=2.5.0",
   "psycopg[binary]>=3.1.8,<3.4",  # SemVer, not explicitely stated, but mentioned on multiple changes.
   "pyparsing>=3.1.0,<3.4",    # Looks like only bugfixes in z-Stream.
-  "python-gnupg>=0.5.0,<0.6",  # Looks like only bugfixes in z-Stream [changelog only in git]
+  "pysequoia @ git+https://github.com/wiktor-k/pysequoia.git@refs/pull/61/head",
```
dralley: Neat

.ci/scripts/check_requirements.py
```diff
@@ -70,7 +70,7 @@ def main():
                     ops = [spec.operator for spec in req.specifier]
                     if "~=" in ops:
                         warnings.warn(f"{filename}:{nr}: Please avoid using ~= on {req.name}!")
-                    elif "<" not in ops and "<=" not in ops and "==" not in ops:
+                    elif "<" not in ops and "<=" not in ops and "==" not in ops and "@" not in line:
```
mdellweg: Well this check was actually intentional:
The check-commit wf fails until all the dependencies are in order again.
It should still allow the tests to run. But obviously prevent prematurely merging it.

pulpcore/app/openpgp.py
```diff
@@ -1,529 +1,87 @@
...
             public_key["user_attributes"].append(user_attribute)
 
+        elif tag == Tag.Signature:
+            sig_attrs = {
+                "sha256": hashlib.sha256(body).hexdigest(),
+                "signature_type": body[1],
+                "raw_data": body,
+                "created": packet.signature_created,
+            }
+            # Note: Pulp's concept of "expiration time" is a duration starting at creation time
```
mdellweg, line 65: Oh it should have been expiration date?
"time" really is ambiguous.

pulpcore/app/openpgp.py
```diff
@@ -1,529 +1,87 @@
...
+            # Sequoia calls that "validity_period", while "expiration_time" is an instant
+            if packet.signature_validity_period is not None:
+                sig_attrs["expiration_time"] = packet.signature_validity_period
+            if packet.key_validity_period is not None:
+                sig_attrs["key_expiration_time"] = packet.key_validity_period
+            if packet.issuer_key_id is not None:
+                sig_attrs["issuer"] = packet.issuer_key_id
+            if packet.signers_user_id is not None:
+                sig_attrs["signers_user_id"] = packet.signers_user_id
```
mdellweg, line 74: How can these be none?
I see this is not from your reimplementation. And I guess the signature subpacket stream could be missing them. So maybe it's actually possible via the file format and I'll call it future proof.

pulpcore/app/openpgp.py
```diff
@@ -1,529 +1,87 @@
...
+            signed_content["signatures"].append(sig_attrs)
+
         else:
-            raise NotImplementedError("Invalid or unknown rfc4880 packet.")
+            raise NotImplementedError(f"Unexpected packet type: {tag!r}")
```
mdellweg, line 78: So this is (the only place?) where it will fall on our feet in the future, should they invent a new packet type, say a BlockChainAttestationLedgerIdIndexPaddingGeneratorAlgorithmParameterPacket.

pulpcore/app/util.py
```diff
@@ -398,25 +431,37 @@ def gpg_verify(public_keys, signature, detached_data=None):
...
+    certs = Cert.split_bytes(public_keys.encode("utf8"))
+
+    def store(key_ids):
+        return certs
+
+    with ExitStack() as stack:
+        if isinstance(signature, str):
+            sig_data = stack.enter_context(open(signature, "rb")).read()
+        elif isinstance(signature, models.Artifact):
+            sig_data = stack.enter_context(signature.file).read()
```
mdellweg, line 440: Admittedly, again, this is not new. But do you think we can feed this as a stream into pysequoia?

pyproject.toml
```diff
@@ -57,7 +57,8 @@ dependencies = [
...
+#   "pysequoia>=0.1.32",
```
mdellweg: Are we now ready to drop the gpg dependency?

mdellweg (commented):
Nice work.

.ci/scripts/check_requirements.py
dralley: It wasn't reaching the tests IIRC

pulpcore/app/openpgp.py
dralley: It is definitely possible to not have a signature without a user ID packet.  I'm not sure that it's even that uncommon.

pulpcore/app/util.py
dralley: By stream do you mean read the data lazily?  e.g. just passing the file handle, not a bytes buffer?

pyproject.toml
dralley: I think so, I think the remaining usage was just in tests, but it's something I would double check before marking the PR ready

pulpcore/app/openpgp.py
dralley: I think so

mdellweg: And that is fine and expected. This part of the future is completely unpredictable.

pulpcore/app/util.py
mdellweg: Yes, that's what i thought.
Specifically if it's an attached signature the file may be huge.

pyproject.toml
mdellweg: So moving it to the test_requirements wouldn't even tell us we really don't use it in the application... That is sad.

pulpcore/app/openpgp.py
mdellweg: How do you verify it, without a way to lookup the signing key?

.ci/scripts/check_requirements.py
mdellweg: OIC this check is in the wrong place. It should be next to "check commit messages".

mdellweg: I think we even document that you can do this...

pulpcore/app/openpgp.py
dralley: The consumer might be verifying it against a specific key, or checking it against all keys rather than looking one up specifically.

pyproject.toml
dralley: The add-signing-service management command also exports keys from the system keyring using python-gnupg

pulpcore/app/openpgp.py
dralley: TBF, pulp is using the terminology that was defined in the spec, so it was "correct", it's just that the terminology in the spec was ambiguous.  

https://www.rfc-editor.org/rfc/rfc9580.html#name-signature-expiration-time

Opinions probably differ whether Sequoia providing `expiration_time` as something else is a good idea.  Less ambiguous wording, but a bit confusing when it comes to the spec.

pulpcore/app/util.py
dralley: ~~Probably https://docs.python.org/3/c-api/buffer.html~~

Actually, no, at least not as easily as I would hope

Why are attached signatures large?  You would extract the signature for checking, surely?

pulpcore/app/openpgp.py
mdellweg: Anyway. The stream could just not have the subpackets. And who are we to judge...

pulpcore/app/util.py
mdellweg: In case `detached_data=None` we assume that `signature` contains the actual data (possible multiple GB) and the attached signature. And re just slurp that into memory with `read`.
But again, not necessarily the scope of this pr.

dralley: yeah I suppose that is a problem with the existing implementation as well.  Do we have an issue filed for that?

pyproject.toml
dralley: I'm not entirely sure how we will get rid of that one, tbh.

pulpcore/app/util.py
dralley: pulp_ansible seems to use the result in a trivial way, we could delete most of this at the next breaking change release.

functest_requirements.txt
```diff
@@ -1,7 +1,7 @@
 pytest<10
 pytest-custom_exit_code
 pytest-xdist
-python-gnupg
+pysequoia
```
mdellweg, line 4: Why are we even repeating the dependency here? (It's not optional in pyproject, right?)

pulpcore/pytest_plugin.py
```diff
@@ -1149,22 +1148,36 @@ def signing_gpg_metadata(signing_gpg_homedir_path):
...
+    subprocess.run(
+        gpg_cmd + ["--import-ownertrust"],
+        input=f"{fingerprint}:6:\n",
+        capture_output=True,
+        text=True,
+        check=True,
+    )
 
-    return gpg, fingerprint, keyid
+    return cert, fingerprint, keyid
```
mdellweg: Is anything outside of pulpcore using this fixture?

pulpcore/pytest_plugin.py
```diff
@@ -1149,22 +1148,36 @@ def signing_gpg_metadata(signing_gpg_homedir_path):
...
+    from pysequoia import Cert
 
-    fingerprint = gpg.list_keys()[0]["fingerprint"]
-    keyid = gpg.list_keys()[0]["keyid"]
+    cert = Cert.from_bytes(private_key_data.encode())
+    fingerprint = cert.fingerprint.upper()
+    keyid = fingerprint[-16:]
 
-    gpg.trust_keys(fingerprint, "TRUST_ULTIMATE")
+    gpg_cmd = ["gpg", "--homedir", str(signing_gpg_homedir_path)]
```
mdellweg, line 1172: Is this now using sequoia too, depending on the actual binary installed?

pulpcore/app/util.py
```diff
@@ -384,9 +381,45 @@ def get_request_without_query_params(context):
...
+    def __init__(self, decrypted):
+        self.valid = True
+        self.data = bytes(decrypted.bytes) if decrypted.bytes is not None else None
+        if decrypted.valid_sigs:
+            vs = decrypted.valid_sigs[0]
+            self.fingerprint = vs.signing_key.upper()
+            self.pubkey_fingerprint = vs.certificate.upper()
+            self.key_id = vs.signing_key[-16:].upper()
+            self.username = ""
+        else:
```
mdellweg: What does this branch mean? Can a signature be valid when there are no signatures at all?

pulpcore/app/management/commands/add-signing-service.py
```diff
@@ -72,13 +71,33 @@ def handle(self, *args, **options):
                 )
             )
 
-        gpg = gnupg.GPG(gnupghome=options["gnupghome"], keyring=options["keyring"])
+        gpg_cmd = ["gpg"]
```
mdellweg, line 74: Wouldn't we need this to use sequoia too in order to handle all the new algorithms and keyfile formats?

functest_requirements.txt
dralley: functional tests can run outside the environment where Pulp is installed.

pulpcore/pytest_plugin.py
dralley: It's using whichever utility is behind the command named `gpg`, which AFAIK is still just `gpg`.  I don't think RHEL 10 is using a replacement under the `gpg` name

pulpcore/app/management/commands/add-signing-service.py
dralley: Even sequoia doesn't "support" them yet because the IETF draft isn't final yet.  You have to install a special build to get them.  So it's out of scope for now.  At some point, probably, but even then it might be a host-dependent thing.

pulpcore/pytest_plugin.py
dralley: Looks like yes https://github.com/search?q=org%3Apulp%20signing_gpg_metadata&type=code

pulp_container and pulp_deb at least, pulp_rpm looks like it might have a copy of an identical name

pulpcore/app/util.py
dralley: I believe you are correct, it's dead code.  Removed

functest_requirements.txt
mdellweg: Ahh, I thought that hasn't been true for a couple of years. But it would be great.

pulpcore/pytest_plugin.py
mdellweg: Looks like pulp_deb's use is only interested in the third argument (keyid).
But Container uses the gpghome from the first one you change. We need to address that first I think.

dralley: Right

dralley: I updated with another compatibility shim.

pulpcore/pytest_plugin.py
```diff
@@ -1146,6 +1133,21 @@ def _sign_with_ascii_armored_detached_signing_service(filename):
     return _sign_with_ascii_armored_detached_signing_service
 
 
+class _GpgCompat:
+    """Wrapper around a pysequoia Cert that provides the python-gnupg GPG interface needed by
+    downstream plugins (e.g. pulp_container) which access .gnupghome and .export_keys()."""
```
mdellweg, line 1138: 👍

pyproject.toml
```diff
@@ -57,7 +57,7 @@ dependencies = [
   "pygtrie>=2.5,<=2.5.0",
   "psycopg[binary]>=3.1.8,<3.4",  # SemVer, not explicitely stated, but mentioned on multiple changes.
   "pyparsing>=3.1.0,<3.4",    # Looks like only bugfixes in z-Stream.
-  "python-gnupg>=0.5.0,<0.6",  # Looks like only bugfixes in z-Stream [changelog only in git]
+  "pysequoia==0.1.32",
```
mdellweg:
```suggestion
  "pysequoia==0.1.32",  # "The API is not stable!" https://github.com/wiktor-k/pysequoia?tab=readme-ov-file#pysequoia
```

mdellweg (commented):
One last suggestion.
You can also ignore...

mdellweg (approved):

PR #7464, opened 2026-03-13
Users should be able to distribute a "repository_version" which selects the latest publication for publication-based plugins (atomic replication part 1)

dralley:
### 📜 Checklist

### DIFF
--- a/pulpcore/app/viewsets/custom_filters.py
+++ b/pulpcore/app/viewsets/custom_filters.py
@@ -10,14 +10,14 @@ from gettext import gettext as _
 
 from django.conf import settings
 from pulpcore.constants import LABEL_KEY_CHARS
-from django.db.models import ObjectDoesNotExist
+from django.db.models import Q
 from django_filters import BaseInFilter, CharFilter, Filter
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 from rest_framework.serializers import ValidationError as DRFValidationError
 
-from pulpcore.app.models import Content, ContentArtifact, Repository, RepositoryVersion, Publication
+from pulpcore.app.models import Content, ContentArtifact, Repository, RepositoryVersion
 from pulpcore.app.viewsets import NamedModelViewSet
 from pulpcore.app.util import get_prn, get_domain_pk, extract_pk, raise_for_unknown_content_units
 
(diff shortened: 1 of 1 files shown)


pulpcore/app/serializers/publication.py
```diff
@@ -266,6 +263,7 @@ class Meta:
             "pulp_labels",
             "name",
             "repository",
+            "repository_version",
```
dralley, line 264: This shouldn't harm plugins that already have this field on the serializer, defined locally.  I checked and they use identical definitions, and Django doesn't care if it is listed multiple times in the fields list.

However, we can still drop that little bit of code from plugins now that it's defined in core.

pulpcore/app/serializers/publication.py
```diff
@@ -193,12 +193,6 @@ class DistributionSerializer(ModelSerializer):
           allow_null=True,
       )
 
-    To include the ``repository_version`` field, it is recommended plugins define the field::
-
-      repository_version = RepositoryVersionRelatedField(
-          required=False, help_text=_("RepositoryVersion to be served"), allow_null=True
-      )
-
```
mdellweg, line 201: Plugins that currently do this will overwrite this serializer field, but that should not be an actual issue, right?

(In the ORM that would be devastating.)

pulpcore/content/handler.py
```diff
@@ -744,28 +744,9 @@ async def _match_and_stream(self, path, request):
...
-                        "repository_version", "pulp_created"
-                    )
-                    repo_version = publication.repository_version
-                except ObjectDoesNotExist:
-                    pass
-
-            if not repo_version:
-                repo_version = await repository.alatest_version()
+            repository, repo_version, publication = await sync_to_async(
+                distro.get_repository_publication_and_version
```
mdellweg, line 748: Any particular reason to not make that function naturally async ?

pulpcore/app/models/publication.py
```diff
@@ -643,6 +631,55 @@ class Distribution(MasterModel):
     class Meta:
         unique_together = (("name", "pulp_domain"), ("base_path", "pulp_domain"))
 
+    def get_repository_publication_and_version(self):
```
dralley, line 634: Defining this in core also removes a bit of work from some plugins

pulpcore/app/serializers/publication.py
mdellweg: NM, you already commented on this.

pulpcore/content/handler.py
dralley: I think it could potentially be called from sync contexts in plugins

Issue #7465, opened 2026-03-13
Add datarepair endpoint to fix repository-versions missing content_ids memomization


gerrod3:
Also, add a post migrate check to tell users to run the endpoint if needed.

PR #7444, opened 2026-03-11
Add a data repair API for admins

dralley:
Some operations take too long to put in a migration, or need to be done at a more granular level (e.g. per-domain), or are optional (e.g. backfilling information)

closes #7272
Assisted-By: claude-opus-4.6

### 📜 Checklist

- [x] Commits are cleanly separated with meaningful messages (simple features and bug fixes should be [squashed](https://pulpproject.org/pulpcore/docs/dev/guides/git/#rebasing-and-squashing) to one commit)
- [x] A [changelog entry](https://pulpproject.org/pulpcore/docs/dev/guides/git/#changelog-update) or entries has been added for any significant changes
- [x] Follows the [Pulp policy on AI Usage](https://pulpproject.org/help/more/governance/ai_policy/)
- [x] (For new features) - User documentation and test coverage has been added

See: [Pull Request Walkthrough](https://pulpproject.org/pulpcore/docs/dev/guides/pull-request-walkthrough/)

PR #7467, opened 2026-03-13
Add datarepair endpoint to fix missing repo-version content-ids

gerrod3:
fixes: #7465

### 📜 Checklist

### DIFF
--- a/pulpcore/app/tasks/datarepair.py
+++ b/pulpcore/app/tasks/datarepair.py
@@ -1,5 +1,7 @@
 from logging import getLogger
 
+from django.db import transaction
+
 from pulpcore.app import models
 from pulpcore.app.models import ProgressReport
 from pulpcore.app.util import get_domain
@@ -81,12 +83,13 @@ def repair_7272(dry_run=False):
                     number_broken += 1
 
                     if not dry_run:
-                        rv.content_ids = list(
-                            rv._content_relationships().values_list("content__pk", flat=True)
-                        )
-                        rv.save()
-                        rv._compute_counts()
-                        fixed_progress.increment()
+                        with transaction.atomic():
+                            rv.content_ids = list(
+                                rv._content_relationships().values_list("content__pk", flat=True)
+                            )
+                            rv.save()
+                            rv._compute_counts()
+                            fixed_progress.increment()
 
             repos_progress.increment()
 


pulpcore/app/urls.py
```diff
@@ -158,6 +159,7 @@ class PulpDefaultRouter(routers.DefaultRouter):
     path("login/", LoginViewSet.as_view()),
     path("repair/", RepairView.as_view()),
     path("datarepair/7272/", DataRepair7272View.as_view()),
+    path("datarepair/7465/", DataRepair7465View.as_view()),
```
dralley, line 162: Had I known we'd have a 2nd one within the week, I would probably have gone the extra distance to generalize this.  We only had 2 data repair management commands total in the past 5 years! (for pulpcore)

dralley: The main difficulty is in getting the routing right without messing up the ability for each endpoint to have different input parameters.

gerrod3: I got the idea from seeing your PR that we could do the same for the missing content_ids. Maybe if it grows to more than 3 we should think about generalizing it.

dralley: Given the similarity between the two and the fact that the first hasn't actually been released yet, would it make any sense to merge them?  e.g. just check all of the RVs, set the cache if it hasn't been set, check the cache if it has been?

gerrod3: Maybe. What issue number would we put them under?

dralley: Maybe we shouldn't keep the issue number convention?

dralley: @bmbouter wdyt?  Also https://github.com/pulp/pulpcore/pull/7444 is the COPR API if you want to take a look, it merged while you were out

pulpcore/app/tasks/datarepair.py
```diff
@@ -105,3 +105,57 @@ def repair_7272(dry_run=False):
...
+        for repo in repos:
+            for rv in models.RepositoryVersion.objects.filter(repository=repo):
+                versions_progress.increment()
+                if rv.content_ids is None:
+                    number_missing += 1
+                    if not dry_run:
+                        rv.content_ids = list(
+                            rv._content_relationships().values_list("content_id", flat=True)
+                        )
+                        rv.save()
```
dralley, line 146: I think this is fine, but could you put lines 84-89 in a transaction?  (re: *my* repair command).  Just to be defensive.

There *might* be performance benefits to putting the increment and the save in a transaction but I don't really remember if that was the case or not.  I don't see a reason to do it here thuogh.

pulpcore/app/urls.py
dralley: I guess the other concern is that issue numbers aren't necessarily stable, we've changed issue trackers 1.5 times in the past 7 years or so (counting Jira as the 0.5...)

We could give them proper names instead

dralley: As per pulpcore meeting -  just keep the issue-number-names.

And don't merge the two

pulpcore/app/tasks/datarepair.py
dralley: Thanks

PR #7481, opened 2026-03-18
Add periodic malloc_trim to prevent unbounded RSS growth in API workers

amasolov:
## Summary

PulpApiWorker (gunicorn SyncWorker) exhibits unbounded RSS growth over time due to glibc heap fragmentation. Django's per-request allocation pattern creates and destroys many small C-level objects (ORM compilers, SQL strings, psycopg cursor state), causing glibc's malloc to retain freed pages rather than returning them to the OS.

This PR adds periodic `gc.collect()` + `malloc_trim(0)` calls in `PulpApiWorker.handle_request()` every N requests (default 1024, configurable via `PULP_MEMORY_TRIM_INTERVAL` env var, set to 0 to disable).

The fix is Linux-only (glibc `malloc_trim`), graceful no-op on other platforms. No new dependencies.

## Problem

Observed in Ansible Automation Platform 2.6 deployments running pulpcore 3.49 on OpenShift: hub-api worker RSS grows ~1 kB/request even with zero user activity (liveness/readiness probes alone drive growth). Over hours this leads to OOM kills and pod restarts.

Profiling on a live cluster confirmed:
- Python object counts are completely stable (`gc.get_objects()` delta ~0)
- `gc.collect()` recovers 0 bytes (no reference cycles)
- `malloc_trim(0)` recovers ~2 MB immediately (heap fragmentation confirmed)
- RSS grows linearly without trimming, stabilizes completely with trimming

The root cause is glibc's default `malloc` behavior: small allocations spread across many arenas cause heap fragmentation, and freed blocks are not returned to the OS until `malloc_trim` is explicitly called.

## Changes

`pulpcore/app/entrypoint.py`:
- At module load: detect Linux, load `libc.malloc_trim` via `ctypes`
- `PulpApiWorker.handle_request()`: after each request, increment counter; every `PULP_MEMORY_TRIM_INTERVAL` requests (default 1024), call `gc.collect()` then `malloc_trim(0)`
- Log at worker init when trimming is enabled

## Configuration

| Env var | Default | Description |
|---------|---------|-------------|
| `PULP_MEMORY_TRIM_INTERVAL` | `1024` | Run trim every N requests. Set to `0` to disable. |

## Test plan

- [ ] Verify workers start normally with default settings (trim enabled)
- [ ] Verify `PULP_MEMORY_TRIM_INTERVAL=0` disables trimming (no log message)
- [ ] Verify RSS growth stabilizes under sustained probe/request load
- [ ] Verify no functional regression on macOS (trim is a no-op, no errors)
- [ ] Run existing unit/functional test suite

### 📜 Checklist

### DIFF
--- a/pulpcore/app/entrypoint.py
+++ b/pulpcore/app/entrypoint.py
@@ -1,6 +1,10 @@
 from contextvars import ContextVar
 from logging import getLogger
+import ctypes
+import ctypes.util
+import gc
 import os
+import platform
 import sys
 import threading
 import time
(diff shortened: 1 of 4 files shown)


dralley:
Can you briefly try using https://docs.python.org/3/library/tracemalloc.html (or something similar, like `memray`) to get a report on what is allocating memory during a standard liveness probe request?

I understand that that would measure what is going on with Python's own allocators rather than libc malloc, but still, I wouldn't think there would be much fragmentation accumulating on a service in a case where the same endpoint was merely being called over and over, using and then releasing approximately the same amount every time.

pulpcore/app/entrypoint.py
```diff
@@ -18,16 +22,43 @@
 
 logger = getLogger(__name__)
 
+_MEMORY_TRIM_INTERVAL = int(os.environ.get("PULP_MEMORY_TRIM_INTERVAL", "1024"))
```
pedro-psb: I guess there is some discussion about this, but in any case, this setting should be defined in `settings.py` as the others and documented in `settings.md`. Settings defined there are automatically overridable via `PULP_{NAME}` envvar.

amasolov: @pedro-psb Good call, updated in the latest push:
- `MEMORY_TRIM_INTERVAL = 1024` added to `settings.py` (so it picks up `PULP_MEMORY_TRIM_INTERVAL` via dynaconf automatically)
- `entrypoint.py` now reads from `settings.MEMORY_TRIM_INTERVAL` in `init_process()` instead of `os.environ.get()`
- Documented in `docs/admin/reference/settings.md`

PR #7492, opened 2026-03-18
Replicas should atomically update their distributions

### LINKED ISSUE #7333: Distribution updates during replication should be atomic

When a replication is triggered, each individual replica is not coordinated in terms of when the distributions switch over, and the repositories within the replica are not processed in a consistent order, nor is it atomic.  

Therefore, clients looking at the repositories on a replica (more than one of them) can run into dependency conflicts (e.g. metalink files in a file repository depend on the associated rpm repositories being synchronised at the same time) while the replication is in progress.  There is no way to introduce this consistency because everything is hands-off and automatic - the repos sync, publish and update their distributions in an uncoordinated fashion - in the background.   

All distributions within a replica during a replication operation should switch over atomically, or at least in a coordinated fashion once all repositories have synced, generated publications etc.  They should not just be updated randomly upon completion of their prerequisite operations.
 
* There is also a potential use case for the final distribution update to optionally be separate and manually triggered - as this would allow the external management application to trigger replication on a number of replicas and once all replicas report that they are ready to update their distributions, that external application can trigger that distribution update on all replicas simultaneously.
* Stretch goal: automatic cross-replica coordination in-line with the process described above.

**Additional context**

https://issues.redhat.com/browse/PULP-1284

dralley:
It's not ideal for distributions to pick up replica changes at random time intervals as various tasks complete, ideally the entire replica is presented as updated at once (or with the smallest possible window).

closes #7333

### 📜 Checklist


pulp_file/app/replica.py
```diff
@@ -33,6 +33,10 @@ def url(self, upstream_distribution):
             manifest = self.publication_ctx_cls(
                 self.pulp_ctx, upstream_distribution["publication"]
             ).entity["manifest"]
+        elif upstream_distribution.get("repository_version"):
+            # Extract repository href from repository_version href
+            repo_href = upstream_distribution["repository_version"].rsplit("versions/", 1)[0]
+            manifest = self.repository_ctx_cls(self.pulp_ctx, repo_href).entity["manifest"]
```
dralley, line 39: This is tricky, it implies that other plugins may also have compatibility issues without modification

dralley: I believe pulp_python may also need a tweak

pulpcore/app/replica.py
```diff
@@ -171,9 +173,18 @@ def create_or_update_repository(self, remote):
     def distribution_extra_fields(self, repository, upstream_distribution):
         """
         Return the fields that need to be updated/cleared on distributions for idempotence.
+
+        Note: repository_version is computed here but filtered out for updates/creates.
+        It will be set atomically in finalize_replication after sync completes.
         """
```
gerrod3, line 179: What's the point of computing this value if it is always filtered out?

pulpcore/app/replica.py
```diff
@@ -197,20 +212,28 @@ def create_or_update_distribution(self, repository, upstream_distribution):
...
-            distribution_data["name"] = upstream_distribution["name"]
+            # Don't set repository_version for new distributions - it will be set in
+            # finalize_replication after sync completes.
+            create_data = {
+                k: v
+                for k, v in distribution_data.items()
+                if k not in ("repository_version", "repository", "publication")
+            }
+            create_data["name"] = upstream_distribution["name"]
+            create_data["pulp_labels"] = distribution_data["pulp_labels"]
```
gerrod3: I remember putting the `pulp_labels` outside of `distribution_extra_fields` since not every plugin called the super method when they overwrote it, but maybe with this change we could do that now.

pulpcore/app/replica.py
dralley: There may not be a point, the previous code just explicitly retaine the `"repository"` value so retained that approach with repository_version - the filtering came later which I guess might render this dead code.

pulpcore/app/replica.py
```diff
@@ -187,9 +192,10 @@ def create_or_update_distribution(self, repository, upstream_distribution):
             )
             if not self._is_managed(distro):
                 return None
+            # Clear repository and publication so they don't conflict when
+            # finalize_replication sets repository_version atomically.
```
gerrod3: While this is needed for the update, we are now going to create a scenario on the first replication after upgrade where distros will start returning 404s until the replication completes.

dralley: @gerrod3 You're talking about the very first replication after the upgrade, before the repository version has been set?  But after that it should be OK, right?

Maybe that's an acceptable cost as long as it's documented - unless you would rather be more lenient with clearing it now at the expense of being a bit less idempotent?  I guess we would then need to deal with the possibility that `repository` might be set later on, and clear it when setting the `repository_version`?

gerrod3: Why don't we just remove the `repository` and `publication` fields from `extra_data` and then in the atomic update set both to `None`. This way we guarantee a smooth update and replication will always correctly update the distribution to the next repo-version even if the user was meddling with them.

dralley: This approach is simpler and I think I prefer it, but it will still have a problem on the first replication, which is that the existing replicated repos which have `repository` set will still not have them cleared until the end.  Therefore, the first replication post-upgrade would likely have the old non-atomic behavior.

Honestly that is still better than 404s though.

pulpcore/app/replica.py
```diff
@@ -171,10 +173,11 @@ def create_or_update_repository(self, remote):
     def distribution_extra_fields(self, repository, upstream_distribution):
         """
         Return the fields that need to be updated/cleared on distributions for idempotence.
+
+        Note: repository, publication, and repository_version are NOT included here.
+        They are all updated atomically in finalize_replication after all syncs complete.
         """
         return {
-            "repository": get_url(repository),
-            "publication": None,
```
gerrod3, line 177: Yay for simplicity!

PR #7510, opened 2026-03-23
Add PgpKeyFingerprintField to pulpcore

daviddavis:


pulpcore/tests/unit/serializers/test_fields.py
```diff
@@ -85,3 +88,92 @@ def test_custom_json_dict_field_raises(field_and_data, binary_arg):
...
+            "keyid:AAAAAAAAAAAAAAAA",
+            id="keyid-lowercase-normalized",
+        ),
+        pytest.param(
+            "v3:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
+            "v3:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
+            id="v3-32hex",
+        ),
+    ],
+)
```
dralley, line 130: @daviddavis How about some tests that try capital `KEYID`, `V4`, etc.

dralley: And if that is rejected then put those values in the invalid test section

PR #7523, opened 2026-03-24
[PULP-1487] Fix slow extract_pk for HREFs

### LINKED ISSUE #6237: The extract_pk utility is very slow for pulp_hrefs
Using `extract_pk(ref)` is very slow when ref is a pulp_href.
I've ran an experiment with a list of 10,000 packages using prns and hrefs [0], and extracting from hrefs is worse by a factor of 1000.

That's probably due the the usage of django's resolve [here](https://github.com/pulp/pulpcore/blob/85e36a0d07685b4b9842efd9a0f5de76bd76b4ad/pulpcore/app/util.py#L194).
There might be a reason why it was used, but we should consider if we can't switch to a pattern matching approach.

[0]

```python
In [34]: timeit(lambda: [extract_pk(n) for n in prns], number=10)
Out[34]: 0.023607832001289353

In [35]: timeit(lambda: [extract_pk(n) for n in hrefs], number=10)
Out[35]: 22.53728126399801

In [36]: len(prns)
Out[36]: 10000

In [37]: len(hrefs)
Out[37]: 10000
```

pedro-psb:
extract_pk() was ~1000x slower for HREFs than PRNs (22.5s vs 0.024s for 10k items) because it called Django's URL resolver on every invocation. Replace it with a direct UUID extraction from the path, which matches the function's documented intent of not validating resource existence.

Add unit tests for extract_pk covering valid HREFs, PRNs, and various invalid inputs including nested URLs (e.g. RepositoryVersion).

Closes #6237

### 📜 Checklist

### DIFF
--- a/pulpcore/tests/unit/test_util.py
+++ b/pulpcore/tests/unit/test_util.py
@@ -1,15 +1,72 @@
 import hashlib
 import tempfile
 import unittest
+from contextlib import nullcontext
 from pathlib import Path
 
 import pytest
 from unittest import mock
+from rest_framework.exceptions import ValidationError
 
 from pulpcore.app import models, util
-from pulpcore.app.util import HashingFileWriter
+from pulpcore.app.util import HashingFileWriter, extract_pk
 
 pytestmark = pytest.mark.usefixtures("fake_domain")
+
+
+TEST_UUID = "019c8cae-cc5f-7148-a3de-456d0a9f39a1"
+
+EXTRACT_PK_CASES = [
+    # happy cases
+    pytest.param(
+        f"/pulp/api/v3/repositories/file/file/{TEST_UUID}/",
+        nullcontext(TEST_UUID),
+        id="valid_href",
+    ),
+    pytest.param(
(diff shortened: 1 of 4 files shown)


pulpcore/tests/unit/test_util.py
```diff
@@ -1,17 +1,78 @@
...
+from pulpcore.app.util import HashingFileWriter, extract_pk
 
 pytestmark = pytest.mark.usefixtures("fake_domain")
 
 
+TEST_UUID = "019c8cae-cc5f-7148-a3de-456d0a9f39a1"
+
+EXTRACT_PK_CASES = [
+    # happy cases
+    pytest.param(
```
dralley, line 25: You need a test for repository versions - this is where it gets tricky, because you can have nesting.

dralley: Well, I guess not, because extract_pk is documented to not support nested URLs.

pedro-psb: The last case asserts it can't handle nested pks, which is documented in the docstring, as you said.

Issue #7482, opened 2026-03-18
PulpApiWorker RSS grows unbounded due to glibc heap fragmentation


amasolov:
API workers (`PulpApiWorker` / gunicorn `SyncWorker`) exhibit continuous RSS growth (~1 kB/request) even under minimal load (health probes only). The growth is caused by glibc heap fragmentation — Django's request cycle allocates and frees many small C-level objects (ORM query compilers, SQL strings, psycopg cursor state), and glibc's `malloc` retains the freed pages in the process heap rather than returning them to the OS.

## Evidence

Profiling on a live Ansible Automation Platform 2.6 deployment (pulpcore 3.49.49, Django 4.2.27, Python 3.12, OpenShift):

- **Python object counts are stable** — `gc.get_objects()` delta is ~0 after initial lazy initialization
- **`gc.collect()` recovers 0 bytes** — no reference cycles
- **`malloc_trim(0)` reclaims ~2 MB immediately** — confirms heap fragmentation
- **RSS grows linearly** at ~1 kB/request without trimming, with no upper bound
- **Master process RSS is flat** — only forked workers are affected

## Impact

Over hours, worker RSS climbs from ~150 MB to multiple GB, leading to:
- Gunicorn worker timeout (`SIGKILL`)
- Health probe failures
- Pod OOM kills and restarts

This is observed even with zero user activity — Kubernetes liveness/readiness probes alone drive the growth.

## Proposed fix

PR #7481 adds periodic `gc.collect()` + `malloc_trim(0)` in `PulpApiWorker.handle_request()`, configurable via `PULP_MEMORY_TRIM_INTERVAL` env var (default: every 1024 requests, set to 0 to disable). Linux-only, graceful no-op on other platforms.

PR #7529, opened 2026-03-26
Set default --max-requests for API workers to prevent unbounded RSS growth

amasolov:
## Summary

- Set `max_requests=10000` and `max_requests_jitter=500` as defaults for API workers
- Workers are gracefully recycled after ~10,000 requests, resetting accumulated glibc heap fragmentation
- Defaults only apply when `max_requests` has not been explicitly configured (via CLI, config file, or `GUNICORN_CMD_ARGS`)

## Context

Follow-up from the discussion in #7481 where @ggainey and @dralley agreed that setting sensible `--max-requests` defaults is the right approach. See #7482 for the full profiling data showing ~1 kB/request RSS growth from glibc heap fragmentation under normal Django ORM workload.

## Implementation note

The defaults are set in `PulpcoreApiApplication.load_app_specific_config()` rather than as click option defaults to preserve the config precedence chain: if a user has already configured `max_requests` via a config file, `GUNICORN_CMD_ARGS`, or CLI flag, their value is respected and not overridden.

## Test plan

- Verify workers restart after ~10,000 requests (check gunicorn logs for worker boot messages)
- Verify `--max-requests 0` on CLI disables recycling
- Verify `GUNICORN_CMD_ARGS="--max-requests 5000"` overrides the default
- Verify existing test suite passes

### 📜 Checklist

- [x] Commits are cleanly separated with meaningful messages (simple features and bug fixes should be squashed to one commit)
- [x] A changelog entry or entries has been added for any significant changes
- [x] Follows the Pulp policy on AI Usage

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pulpcore/app/entrypoint.py
+++ b/pulpcore/app/entrypoint.py
@@ -124,6 +124,13 @@ class PulpcoreApiApplication(PulpcoreGunicornApplication):
             PulpApiWorker.__module__ + "." + PulpApiWorker.__qualname__,
             enforced=True,
         )
+        # Gunicorn's default for max_requests is 0 (unlimited worker lifetime). Apply Pulp defaults
+        # only when the user did not pass --max-requests on the pulpcore-api CLI. An explicit
+        # --max-requests 0 means "disable recycling" and must not be replaced.
+        if self.cfg.max_requests == 0 and self.options.get("max_requests") is None:
+            self.cfg.set("max_requests", 10000)
+            if self.options.get("max_requests_jitter") is None:
+                self.cfg.set("max_requests_jitter", 500)
 
     def load(self):
         using_pulp_api_worker.set(True)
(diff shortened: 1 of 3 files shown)


pulpcore/app/entrypoint.py
```diff
@@ -124,6 +124,9 @@ def load_app_specific_config(self):
             PulpApiWorker.__module__ + "." + PulpApiWorker.__qualname__,
             enforced=True,
         )
+        if self.cfg.max_requests == 0:
```
dralley: The comment describes disabling recycling by setting `--max-requests 0`, but it would seem like this would explicitly use the defaults in that scenario.  Am I missing something?

dralley: @amasolov

dralley (approved):

PR #7530, opened 2026-03-26
Make PgpKeyFingerprintField prefix optional, default to v4

daviddavis:
When a bare hex string is provided without a version prefix, PgpKeyFingerprintField now assumes 'v4:' as the default prefix. This makes the field backwards compatible.

### DIFF
--- a/pulpcore/app/serializers/fields.py
+++ b/pulpcore/app/serializers/fields.py
@@ -448,15 +448,21 @@ class PgpKeyFingerprintField(serializers.CharField):
     A DRF field for type-prefixed OpenPGP key identifiers.
 
     Accepts formats like 'v4:<40-hex>', 'v6:<64-hex>', or 'keyid:<16-hex>'.
+    If no prefix is provided, 'v4:' is assumed.
     Normalizes hex to uppercase on input and validates the format.
     """
+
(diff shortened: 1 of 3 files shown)


pulpcore/app/serializers/fields.py
```diff
@@ -448,15 +448,21 @@ class PgpKeyFingerprintField(serializers.CharField):
...
 
     Accepts formats like 'v4:<40-hex>', 'v6:<64-hex>', or 'keyid:<16-hex>'.
+    If no prefix is provided, 'v4:' is assumed.
     Normalizes hex to uppercase on input and validates the format.
     """
 
+    DEFAULT_PREFIX = "v4"
+
     # Matches versioned fingerprints (v3/v4/v5/v6) and legacy 16-char key IDs.
     FINGERPRINT_RE = re.compile(r"^(v\d:[0-9A-F]{32,64}|keyid:[0-9A-F]{16})$")
```
dralley, line 458: Is this "length 32 to 64" or "length 32 or 64"

A v4 fingerprint is length 40

daviddavis: It's 32 to 64 and not 32 OR 64.

PR #7546, opened 2026-03-30
[PULP-1045] Add support for serving superseded publications for a retention period

### LINKED ISSUE #7514: Distributed Publication Retention Cache
## Problem

In a system where publications are done often/quickly, there is a high chance of a content client (e.g, dnf in rpm context) keep a cache of content/metadata links that will get stale with publication updates. In those cases, Pulp should enable the links to keep working for a given time.

## Requirements

1. As a Pulp **content client**, I want to be able to get content from a distribution that refers to older publications, as long as that publication was replaced within a given retention period.
2. As a **Pulp admin**, I want to be able to configure the distributed publication retention time for any plugin which supports it (that uses publications).
3. As a **Pulp admin**, I want to be able to enable/disable the distributed publication retention cache feature for plugins (e.g, to avoid overhead associated with the feature).
## Proposal

Move/hoist the `pulp_deb` reference implementation to pulpcore (see link below).
The implementation keeps a history of publications associated with a disitrubution.  Only one record doesn't have an expiration date (the current publication), while the others have one set according to the retention time setting.
See the diagram below:

<details><summary>Example Diagram</summary>
<p>
Image
</p>
</details> 

Sample configuration:

```python
DISTRIBUTED_PUBLICATION_RETENTION_CACHE = {
    "default": {
        "enable": False,
        "retention_time": 3600,
    },
    "rpm": {
        "enable": False,
        "retention_time": 7200,
    },
    "deb": {
        "enable": True,
        "retention_time": 46120,
    },
}
```

## Caveats

In the example diagram above, if content at path `{dist}/A` changed from publication 1 to publication 2, Pulp cannot know if the client wants the old content (the url is the same!). By the implementation, the newer content will be served.

In general this is something that can happen.

In the case of RPM, this should be rare, because the relative path published by pulp_rpm contains the NVRA, and any content changes *should* update the NVRA (the release, for example). But we are not naive, and we know that even for RP, content can change (the checksum changes) while NVRA remains the same. This proposal can't handle this problem. If that's important, another approach must be taken, like having a system setup with multiple distributions that explicitly have versions/checkpoints in their path. Possibly the[ distribution checkpoint](https://pulpproject.org/pulpcore/docs/user/guides/checkpoint/?h=chec) feature that was recently added (last year or so) in pulpcore.

## Alternatives

a. For requirement (1), copy the reference implementation from pulp_deb to other plugins as needed.
b. For requirements (2) and (3), a simpler alternative is defining the retention time per plugin on the plugin's distribution class, and always enable it for plugins that uses publications.

## Additional Context
- (Original) Proposal on pulp_deb: <https://github.com/pulp/pulp_deb/issues/911>
- Feature request in RPM: <https://github.com/pulp/pulp_rpm/issues/4236>

pedro-psb:
Introduces a DistributedPublication model to track which publications are associated with each distribution. When a new publication becomes the latest for a repository, older ones are retained and served by the content handler for a configurable grace period, allowing clients to continue fetching from superseded publications without interruption.

Handles publications with pass-trough=True|False. The behavior is sumarized:

Given:
- CA=ContentArtifact,
- PA=PublishedAritfact
- PU=Publication
- RV=RepositoryVersion
- DP=DistributedPublication
- DT=Distribution

Then CA is retrieved with:
- no-pass-through:    CA<-PA<-PU<-DP->DT
- pass-through: CA<-RV<-PU<-DP->DT

Closes: #7514

### 📜 Checklist


pedro-psb:
The test PR is fine: https://github.com/pulp/pulp_deb/pull/1429

docs/admin/reference/settings.md
```diff
@@ -293,6 +293,15 @@ All installed plugins must be Domain compatible for Pulp to start.
 
 Defaults to `False`.
 
+### DISTRIBUTED\_PUBLICATION\_RETENTION\_PERIOD
```
mdellweg: Should this be a global thing?

pulpcore/tests/unit/models/test_publication_retention.py
```diff
@@ -0,0 +1,224 @@
...
+@pytest.mark.django_db
+class TestClearDistributedPublication:
+    def test_deleting_publication_clears_dps(self, db):
+        pub = pub_factory()
+        dist = dist_factory(pub=pub)
+        assert DistributedPublication.objects.filter(distribution=dist).exists()
+        pub.delete()
+        assert not DistributedPublication.objects.filter(distribution=dist).exists()
+
+    def test_deleting_older_repository_version_doesnt_clear_dps(self, db):
```
mdellweg: What's dps? (Actually there's no need for abbreviation in test case names. They don't have call sites.)

mdellweg: Ah DistributedPublication(s). That's covered by the class name. How about:
`test_does_not_happen_when_deleting_old_repository_version`?

mdellweg (commented):
Quite a complex thing, but looks good over all.

docs/admin/reference/settings.md
pedro-psb: This could be a more complex setting and be defined per plugin, or not exposed at all. I could be convinced either way. I decided to start with this as a middle ground.

mdellweg:
Can you squash before merging?

docs/admin/reference/settings.md
jobselko: Can you move this up to follow alphabetical order?

mdellweg (approved):

PR #7551, opened 2026-04-02
Add an overwrite boolean when modifying repo content

### LINKED ISSUE #7550: Add a way to reject repo content updates if overwrites would occur
One issue we've faced is content in repositories being overwritten (e.g. a package with the same NVA or NVRA overwrites an existing package with a different hash). This is often unintentional and it leads to various problems. One big problem is the old package will be cached and its hash will not match the one in the repo metadata. 

Sometimes we do want to allow packages to be overwritten so I think adding a flag to the repo modify endpoint and the upload endpoint (when `repository` is specified) would be the best option. Maybe `overwrite`? And perhaps it could default to True for backwards compatibility. And `False` would return a 409 if a user attempts to upload a package that would overwrite an existing one.

daviddavis:
Add a `overwrite` boolean parameter to endpoints which modify repository content. When set to false, the update checks whether any content being added would overwrite existing content (based on repo_key_fields) and returns a 409 Conflict response with details instead of silently replacing it.

fixes #7550

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -29,7 +29,7 @@ from pulpcore.app.util import (
 from pulpcore.cache import Cache
 from pulpcore.constants import ALL_KNOWN_CONTENT_CHECKSUMS, PROTECTED_REPO_VERSION_MESSAGE
 from pulpcore.download.factory import DownloaderFactory
-from pulpcore.exceptions import ResourceImmutableError
+from pulpcore.exceptions import ContentOverwriteError, ResourceImmutableError
 
 from .base import BaseModel, MasterModel
 from .content import Artifact, Content, ContentArtifact, RemoteArtifact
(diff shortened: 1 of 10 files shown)


pulpcore/app/models/repository.py
```diff
@@ -242,6 +242,73 @@ def finalize_new_version(self, new_version):
...
+
+        add_content = Content.objects.filter(pk__in=add_content_pks)
+
+        for type_obj in self.CONTENT_TYPES:
+            repo_key_fields = type_obj.repo_key_fields
+            if not repo_key_fields:
+                continue
+
+            pulp_type = type_obj.get_pulp_type()
+            incoming_qs = type_obj.objects.filter(pk__in=add_content.filter(pulp_type=pulp_type))
```
gerrod3: I know we have this pattern in our repo-version utils file, but I wonder if it is more performant than a normal `type_obj.objects.filter(pk__in=add_content)`. Theoretically it should filter down the number of content to check to only those of the correct type, but does it actually increase the speed?

pulpcore/app/models/repository.py
```diff
@@ -242,6 +242,73 @@ def finalize_new_version(self, new_version):
...
+                )
+                if not existing_dups.exists():
+                    continue
+
+                existing_by_key = {
+                    tuple(row[f] for f in repo_key_fields): row["pk"] for row in existing_dups
+                }
+                incoming_by_key = {
+                    tuple(row[f] for f in repo_key_fields): row["pk"]
+                    for row in incoming_qs.filter(find_dup_qs).values("pk", *repo_key_fields)
```
gerrod3: We can remove this query if we add `pk` to the `batch_qs(incoming_qs.values(*repo_key_fields)` at the top, then create this dictionary during the assembling of the `find_dup_qs`.

gerrod3 (changes requested):
I just added the same feature (in a slightly different manner) to pulp-python to prevent the same problem. https://github.com/pulp/pulp_python/pull/1163

One of things I was concerned about was the performance of trying to validate the modify call, at request time, will not produce an overwrite against the potential thousands of units a user can call with. Have you tested this against your own instance? How large are the modify calls you typically use?

dkliban:
@daviddavis I am concerned that doing the validation in the API is going to cause timeouts. How do you feel about performing the validation in the task?

daviddavis:
@dkliban moved the validation to the tasks. I perform the check at the beginning of the task, but it could in theory go into the finalize repo version step. I figure checking early fails fast and avoids wasted work building a version that will be discarded.

gerrod3 (approved):
This looks good!

PR #7553, opened 2026-04-03
Add sync optimization support to pulp_file

dralley:
Whe the manifest hasn't changed relative to the previous sync, we skip the sync, so long as the sync modes are compatible and such.

This requires storing a bit of hidden metadata on the repository object.

### 📜 Checklist


pulp_file/app/viewsets.py
```diff
@@ -252,14 +252,14 @@ class FileRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin, Role
...
-    @action(detail=True, methods=["post"], serializer_class=RepositorySyncURLSerializer)
+    @action(detail=True, methods=["post"], serializer_class=FileRepositorySyncURLSerializer)
     def sync(self, request, pk):
         """
         Synchronizes a repository.
 
         The ``repository`` field has to be provided.
         """
-        serializer = RepositorySyncURLSerializer(
+        serializer = FileRepositorySyncURLSerializer(
```
dralley: This is a little bit unfortunate.

The issue is that since PulpFile did not previously provide any extensions to `RepositorySyncURLSerializer`, `RepositorySyncURLSerializer` was literally baked into the API.  So now the addition of a new field changes the actual type.

Which in turn means that the API bindings will break by merely adding it as an optional parameter.

If this is considered a breaking change, then I would likely need to strip this PR of everything but the task change, and maybe add a temporary setting in settings.py to allow some control over it, and then return the serializer / viewset changes later in 3.115

ggainey: This is potentially a *REST API* breaking change - i.e., affects end-clients, not plugins. We can make those, with enough warning - but it definitely needs to go through an announcement/deprecation cycle. The API is a contract we promise to users, changing it is...Fraught (as we have learned)

dralley: Want me to go ahead and drop the 2nd commit from this PR then?

ggainey: Yeah - I'd rather move carefully.

pulp_file/app/models.py
```diff
@@ -91,6 +91,7 @@ class FileRepository(Repository, AutoAddObjPermsMixin):
 
     manifest = models.TextField(default="PULP_MANIFEST", null=True)
     autopublish = models.BooleanField(default=False)
+    last_sync_details = models.JSONField(default=dict)
```
ggainey, line 94: We're now going to have this in both FileRepo and RpmRepo. We should discuss (not for this PR!) moving this into the base-class, so it's available for every plugin.

pulp_file/app/settings.py
```diff
@@ -0,0 +1 @@
+FILE_SYNC_OPTIMIZATION = True
```
ggainey, line 1: Do we want to add a pulp_file/docs/admin/reference/settings.md and document this?

dralley: I don't think we do, it exists pretty much only as a temporary hack to get a feature working for one user before the next breaking release.

pulp_file/app/models.py
dralley: Potentially, but every plugin has to define its own fields to check anyway so it's borderline whether it would be worthwhile.  Also not every repository is syncable.

ggainey: Hm, ok fair.

PR #7567, opened 2026-04-07
Fix pytest plugin bug with file storage and domains

pedro-psb:
@ggainey found the domain factory was failing when domains were enabled and we were using file storage.
For this case in CI for validating a solution.

### DIFF
--- a/template_config.yml
+++ b/template_config.yml
@@ -66,6 +66,7 @@ pulp_settings:
   task_protection_time: 10
   tmpfile_protection_time: 10
   upload_protection_time: 10
+  domain_enabled: true
 pulp_settings_azure:
   MEDIA_ROOT: ""
   STORAGES:
--- a/.github/workflows/scripts/before_install.sh
+++ b/.github/workflows/scripts/before_install.sh
@@ -50,7 +50,7 @@ legacy_component_name: "pulpcore"
 component_name: "core"
 component_version: "${COMPONENT_VERSION}"
 pulp_env: {"PULP_CA_BUNDLE": "/etc/pulp/certs/pulp_webserver.crt"}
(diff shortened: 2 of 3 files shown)


pulpcore/pytest_plugin.py
```diff
@@ -630,17 +630,23 @@ def _settings_factory(storage_class=None, storage_settings=None):
         ]
 
         def get_installation_storage_option(key, backend):
-            value = pulp_settings.STORAGES["default"]["OPTIONS"].get(key)
+            value = pulp_settings.STORAGES["default"].get("OPTIONS", {}).get(key)
```
ggainey, line 634: 👍

template_config.yml
```diff
@@ -66,6 +66,7 @@ pulp_settings:
   task_protection_time: 10
   tmpfile_protection_time: 10
   upload_protection_time: 10
+  domain_enabled: true
```
mdellweg, line 69: Just to be absolutely clear and on the same page. As long as we do not call this pulp4, we will not remove this setting from pulp, right?
Or will we keep "running without domains" around until api/v3 is retired? I believe that would be incompatible with api/v4.

pedro-psb: > As long as we do not call this pulp4, we will not remove this setting from pulp, right?

That's my understanding. Exactly on pulp v4.0.0 we remove this option and domains is always enable, regardless of us still having api/v3 around. Which is something to think about... Non-domain users don't have 'default' in their hrefs, but will start to have when they switch, breaking the url/script in the way we are trying to avoid. I'm not up to the api/v4 work, maybe it's addressed already.

mdellweg: My point is that we want to keep one scenario around to run without domains enabled for now.

PR #7586, opened 2026-04-13
Fix race condition in reverse-replication

dralley:
Resulted in flaky CI behavior and sporadic failures

### 📜 Checklist

### DIFF
--- a/pulpcore/app/replica.py
+++ b/pulpcore/app/replica.py
@@ -282,11 +282,16 @@ class Replicator:
         ]
 
         if repository_ids or remote_ids:
+            # Lock distros_uris to prevent concurrent distribution create/update tasks
+            # from racing with this deletion. Deleting a repository CASCADE-deletes its
+            # publications (and RVs, etc.), which SET_NULLs any distribution FK pointing
+            # to them. A concurrent distribution update could then write back the stale FK,
+            # causing an IntegrityError.
             dispatch(
                 general_multi_delete,
                 task_group=self.task_group,
                 shared_resources=[self.server],
-                exclusive_resources=repositories + remotes,
+                exclusive_resources=self.distros_uris + repositories + remotes,
                 args=(repository_ids + remote_ids,),
             )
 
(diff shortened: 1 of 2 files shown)


pulpcore/app/replica.py
```diff
@@ -282,11 +282,16 @@ def remove_missing(self, names):
...
+            # from racing with this deletion. Deleting a repository CASCADE-deletes its
+            # publications (and RVs, etc.), which SET_NULLs any distribution FK pointing
+            # to them. A concurrent distribution update could then write back the stale FK,
+            # causing an IntegrityError.
             dispatch(
                 general_multi_delete,
                 task_group=self.task_group,
                 shared_resources=[self.server],
-                exclusive_resources=repositories + remotes,
+                exclusive_resources=self.distros_uris + repositories + remotes,
```
mdellweg, line 294: I'm not sure this works in the way described, because we do not lock on individual distributions in other places (maybe it removes the race by coincidence).
Also is cascade deleting of anything protected by resource locks?

dralley: If you look at `create_or_update_distribution`, which this is racing with, it does lock on `self.distro_uris`

mdellweg: That seems to contradict this locking here:
https://github.com/pulp/pulpcore/blob/main/pulpcore/app/viewsets/publication.py#L531

(Probably even just an unrelated bug.)

dralley: I'm not 100% sure what you mean by "contradiction" - `self.distro_uris` is the [same](https://github.com/dralley/pulpcore/blob/4ca086a3444c73822ae228c523df8cdc99ae892a/pulpcore/app/replica.py#L64) string.

I believe in both cases it's due to `base_path` being effectively global state

I don't believe being linked on a distribution protects repositories or remotes from being cascaded.  Both are explicitly `on_delete=models.SET_NULL`

mdellweg: Ah, now i see it too. Thanks, the variable is misleading...

mdellweg (approved):

PR #7587, opened 2026-04-14
[PULP-1299] Add setting to allow admins to avoid timeout issues with immediate tasks

gerrod3:
We can backport new settings right?

I looked at the [JIRA](https://redhat.atlassian.net/browse/PULP-1299) and in this case there really isn't any expensive operation going on for RPM or Container distribution updates. There's two synchronous operations: 1. the database update call, 2. the cache invalidation hook. These should both be fairly fast and happen under a second, so not sure how satellite is hitting the 5 second timeout. The only conclusion I can draw is that their third party systems are really slow.

I really don't think we should update the RPM & Container tasks to be deferred since Services never runs into this problem and it would significantly slow down their system if we did so with the thousands of updates they handle daily. I also didn't think we should allow the immediate timeout to be configurable with a setting since the user would be unaware that it would affect the task worker's heartbeat if increased too large and could end up with more cancelled tasks. Open to suggestions, but I think this setting is the simplest solution that can satisfy everyone's needs. 

### 📜 Checklist

### DIFF
--- a/docs/admin/reference/settings.md
+++ b/docs/admin/reference/settings.md
@@ -392,6 +392,16 @@ This time is only accurate to one worker heartbeat corresponding to `WORKER_TTL 
 
 Defaults to `600` seconds.
 
+### TASK\_PREFER\_DEFER
+
+For tasks that are designed to run immediately, but could be deferred to a task worker, always choose to defer to the worker.
+This is helpful for systems with slow databases that could cause the immediate task to timeout after its 5 second limit.
+
+!!! note
+    Setting this to `True` will slow your Pulp's task throughput, especially if you perform many immediate tasks frequently.
+
+Defaults to `False`.
+
 ### TASK\_PROTECTION\_TIME, TMPFILE\_PROTECTION\_TIME and UPLOAD\_PROTECTION\_TIME
 
 Pulp uses `tasks`, `pulp temporary files` and `uploads` to pass data from the api to worker tasks.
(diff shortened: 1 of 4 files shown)


docs/admin/reference/settings.md
```diff
@@ -392,6 +392,14 @@ This time is only accurate to one worker heartbeat corresponding to `WORKER_TTL
 
 Defaults to `600` seconds.
 
+### TASK\_PREFER\_DEFER
+
+For tasks that are designed to run immediately, but could be deferred to a task worker, always choose
+to defer to the worker. This is helpful for systems with slow databases that could cause the immediate
+task to timeout after its 5 second limit.
+
```
mdellweg:
```suggestion
For tasks that are designed to run immediately, but could be deferred to a task worker, always choose to defer to the worker.
This is helpful for systems with slow databases that could cause the immediate task to timeout after its 5 second limit.

```
Should we add that it can severely impact performace?

Also this reminds me that we would not need to apply that limit could we run the tasks in a proper async fashion alongside the (sadly not yet) async application.

For the time being I agree that this is the best option we can give admins. +1 to backport.

pedro-psb (commented):
I guess you've asked about this before, maybe it's already sorted out, but I believe it will also timeout on the worker. Could you have a look into that?

gerrod3:
> I guess you've asked about this before, maybe it's already sorted out, but I believe it will also timeout on the worker. Could you have a look into that?

Yes an immediate task that is deferred to the worker would also be constrained to the 5 second timeout. The tasks in the JIRA were ones that ran on the worker. They do a bunch of distribution updates at once, so most of them get deferred to the worker since they share the same global distribution lock. That is why the new setting overwrites the task's immediate field if possible so that when deferred it doesn't get ran as an immediate task.

mdellweg (approved):

PR #7591, opened 2026-04-14
[PULP-1639] Fix gunicorn "Control server error"

### LINKED ISSUE #7574: Control server error: [Errno 13] Permission denied  0     PermissionError(13, 'Permission denied')
**Version**
`"core": "3.106.0",`

When `pulpcore-content` redeploys we get the following error in our logs:

Control server error: [Errno 13] Permission denied  0     PermissionError(13, 'Permission denied')

pedro-psb:
gunicorn [25.1.0](https://gunicorn.org/news/#2510-2026-02-13) introduced a control socket (gunicornc) that defaults to `gunicorn.ctl` relative to the working directory. Since pulpcore-content sets its CWD to WORKING_DIRECTORY (`/var/lib/pulp/tmp` by default), the socket lands on the shared PVC and persists across pod restarts, causing Permission denied when a new pod tries to recreate it during a rolling update.

Default to /tmp/pulpcore-content.ctl, which is pod-local ephemeral storage. Users who want a different path can override via gunicorn.conf.py.

fixes: #7574
Assisted-by: Claude Code

### 📜 Checklist

### DIFF
--- a/pulpcore/app/entrypoint.py
+++ b/pulpcore/app/entrypoint.py
@@ -14,7 +14,10 @@ from gunicorn.workers.sync import SyncWorker
 
 from pulpcore.app.apps import pulp_plugin_configs
 from pulpcore.app.netutil import has_ipv6
-from pulpcore.app.pulpcore_gunicorn_application import PulpcoreGunicornApplication
+from pulpcore.app.pulpcore_gunicorn_application import (
+    handle_control_interface_feature,
+    PulpcoreGunicornApplication,
+)
 
 logger = getLogger(__name__)
 
@@ -146,7 +149,7 @@ class PulpcoreApiApplication(PulpcoreGunicornApplication):
 
 
 @click.option(
-    "--bind", "-b", default=[f"{ '[::]' if has_ipv6() else '0.0.0.0' }:24817"], multiple=True
+    "--bind", "-b", default=[f"{'[::]' if has_ipv6() else '0.0.0.0'}:24817"], multiple=True
 )
 @click.option("--workers", "-w", type=int)
 # @click.option("--threads", "-w", type=int)  # We don't use a threaded worker...
@@ -181,6 +184,11 @@ class PulpcoreApiApplication(PulpcoreGunicornApplication):
 @click.option("--user", "-u")
 @click.option("--group", "-g")
 @click.option(
+    "--control-socket",
+    "control_socket",
+    help="Path to the gunicorn control socket (requires gunicorn>=25.1).",
+)
+@click.option(
     "--name-template",
     type=str,
     help="Format string to use for the status name. "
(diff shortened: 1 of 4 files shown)


pulpcore/content/entrypoint.py
```diff
@@ -15,6 +15,9 @@ def load_app_specific_config(self):
         )
         self.set_option("default_proc_name", "pulpcore-content", enforced=True)
         self.set_option("worker_class", worker_class, enforced=True)
+        # On k8s, the default location may persist across restarts and cause permission errors
+        # See: <https://github.com/pulp/pulpcore/issues/7574>
```
mdellweg: What was the default location?
This file belongs in /run/ somewhere.

> This directory contains system information data describing the system since it was booted. Files under this directory must be cleared (removed or truncated as appropriate) at the beginning of the boot process.
[...]
System programs that maintain transient UNIX-domain sockets must place them in this [/run] directory or an appropriate subdirectory as outlined above.

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html

pedro-psb: It defaults to the current directory: https://gunicorn.org/guides/gunicornc/#start-gunicorn-with-control-socket
I wonder why.

pedro-psb: Hmmm, the code apparently says something else:
https://github.com/benoitc/gunicorn/blob/9aa54703f4950818aed538dbee9578e868375cc9/gunicorn/config.py#L3138-L3146

pedro-psb: Ok, it was changed in 25.2: https://github.com/benoitc/gunicorn/commit/0ad47db800b70a3745e104708334faafe746cafa

I guess we should not do anything, then.
This change was to improve on gunicorn's default (of version 25.1), but they have improved by themselves.

pedro-psb: Or do you think it's still worth it, to account for the case 25.1 is installed?

mdellweg: So what I understand is that gunicorn tries XDG_RUNTIME_DIR first and falls back to HOME.
I would claim that the variable XDG_RUNTIME_DIR should have been set. Not sure if the os in the container or the container runtime is to blame, but the default gunicorn behaviour seems sound to me and your change makes that unnecessarily rigid.
We should probably propagate the option instead so it stays possible to overwrite it.

pedro-psb: > We should probably propagate the option instead so it stays possible to overwrite it.

That sounds good.

pedro-psb: > So what I understand is that gunicorn tries XDG_RUNTIME_DIR first and falls back to HOME.

The first release of the feature the default was the current directory (whathever that was...). In the following Y they've changed to this, which I agree is sane.

pulpcore/app/entrypoint.py
pedro-psb, line 1: My local flake8 keeps complaining about those

mdellweg (commented):
Why only add it to the content app?

pedro-psb:
> Why only add it to the content app?

Good point

mdellweg (approved):

ianballou:
@pedro-psb , Katello on core 3.105 is hitting this as well, it relates to an selinux issue we are trying to have fixed. I'm not sure if you've been contacted about it yet, but could we have this backported to 3.105?

PR #7593, opened 2026-04-14
Add self-service non-admin user RBAC

gerrod3:
This should allow users to update their password without needing admin permission.

### 📜 Checklist

### DIFF
--- a/CHANGES/+self_service_user.feature
+++ b/CHANGES/+self_service_user.feature
@@ -1,0 +1,1 @@
+Added ability for non-admin users to update their personal account fields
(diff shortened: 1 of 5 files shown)


mdellweg:
I wonder if we could add this to the login viewset in a useful way.

That would mean the users viewset keeps being admins only, and users only ever interact with the login. We could even use a serializer that asks for the old, the new and a confirmation of the new password (when thinking about ui...), still allowing an admin to just (re-)set a users password.

josephtate (approved):
Yes, this looks like "just enough"

gerrod3:
@mdellweg I did your suggestion, put it as a patch call to /login. Let me know if you have any concerns

mdellweg (approved):
Yes, I like that much more. Thank you!

PR #7607, opened 2026-04-17
Add publish parameter to repository modify endpoint

### LINKED ISSUE #7606: Allow users to publish when calling the repo modify endpoint
Pulp currently offers an auto publish feature but it's not of use to us. Repos we have are shared across teams and while one team might want to auto-publish, another team might not. What we'd like is a publish parameter on the repo modify endpoint so anyone can choose to publish when they modify repo content.

daviddavis:
Pulp currently offers an auto publish feature. However, our repos are shared across teams and while one team might want to auto-publish, another team might not. This adds a publish parameter on the repo modify endpoint so anyone can choose to publish when they modify repo content.

fixes #7606

Co-authored-by: AI (GitHub Copilot)


daviddavis:
Closing this out since I think we have a different solution that's viable (https://github.com/daviddavis/pulp_workflow).

PR #7649, opened 2026-04-27
Add a ResourceBudget mechanism which keeps disk usage in check during syncs

dralley:
### 📜 Checklist

### DIFF
--- a/docs/admin/reference/settings.md
+++ b/docs/admin/reference/settings.md
@@ -474,10 +474,41 @@ Defaults to `/var/lib/pulp/tmp/`.
 
 ### MAX\_CONCURRENT\_CONTENT
 
-The size of the batch of content processed in one go when syncing content from
-a remote.
+The maximum number of concurrent downloads during sync. Controls how many HTTP
+download tasks can run in parallel within the the sync pipeline.
 
-Defaults to 25.
+Defaults to 200.
+
+!!! warning "Deprecated"
+    This setting is deprecated and may be removed in a future release.
+    Use `SYNC_MAX_IN_FLIGHT_ITEMS` instead, which provides similar
+    functionality. If `MAX_CONCURRENT_CONTENT` is set by the user and
+    `SYNC_MAX_IN_FLIGHT_ITEMS` is not, its value will be used as
+    `SYNC_MAX_IN_FLIGHT_ITEMS` automatically.
+
+
+### SYNC\_MAX\_IN\_FLIGHT\_MB
+
+The maximum total size (in megabytes) of downloaded artifacts that are waiting to be
+saved. This limits the temporary disk space consumed by artifacts that have been
+downloaded but not yet persisted.
+
+When set, small artifacts will download with high concurrency while large artifacts
+will automatically throttle to avoid exhausting disk space. This does not prevent
+individual artifacts larger than the threshold from being synced, only limits
+concurrency in such situations.
+
+Defaults to 5120 (5gb)
+
+### SYNC\_MAX\_IN\_FLIGHT\_ITEMS
+
+The maximum number of downloaded artifacts that are waiting to be saved. Like
+`SYNC_MAX_IN_FLIGHT_MB`, this limits unpersisted artifacts, but counts items
+rather than bytes.
+
+This is useful as a fallback when artifact sizes are not known ahead of time.
+
+Defaults to `None` (no limit).
 
 ## Redis Settings
 
--- a/pulpcore/app/settings.py
+++ b/pulpcore/app/settings.py
@@ -403,7 +403,17 @@ HIDE_GUARDED_DISTRIBUTIONS = False
 
 DOMAIN_ENABLED = False
 
-MAX_CONCURRENT_CONTENT = 25
+MAX_CONCURRENT_CONTENT = 200
+
+# Resource budget for sync pipeline: limits total in-flight artifact data between
+# the ArtifactDownloader and ArtifactSaver stages. When set, these allow higher download
+# concurrency for small artifacts while preventing disk exhaustion for large ones.
+# None means no limit for that dimension.
+
+# Maximum megabytes of in-flight downloaded artifacts
+SYNC_MAX_IN_FLIGHT_MB = 5120
+# Maximum number of items (content) able to be buffered between downloading and saving
+SYNC_MAX_IN_FLIGHT_ITEMS = None
 
 SHELL_PLUS_IMPORTS = [
     "from pulpcore.app.util import get_domain, get_domain_pk, set_domain, get_url, extract_pk",
--- a/pulpcore/plugin/stages/api.py
+++ b/pulpcore/plugin/stages/api.py
@@ -77,7 +77,7 @@ class Stage:
                 break
             yield content
 
-    async def batches(self, minsize=500):
+    async def batches(self, minsize=500, flush_event=None):
         """
         Asynchronous iterator yielding batches of [DeclarativeContent][] from `self._in_q`.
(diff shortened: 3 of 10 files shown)


docs/admin/reference/settings.md
```diff
@@ -474,10 +474,38 @@ Defaults to `/var/lib/pulp/tmp/`.
...
+### SYNC\_MAX\_IN\_FLIGHT\_MB
+
+The maximum total size (in megabytes) of downloaded artifacts that are waiting to be
+saved. This limits the temporary disk space consumed by artifacts that have been
+downloaded by `ArtifactDownloader` but not yet persisted by `ArtifactSaver`.
+
+When set, small artifacts will download with high concurrency while large artifacts
+will automatically throttle to avoid exhausting disk space.
+
+Defaults to 5120 (5gb)
```
mdellweg, line 501: Is this a hard limit? Will this option prevent syncs to succeed with one 6gb file?
(Not that this is a bad thing, just maybe worth mentioning. TBF the available diskspace is already a hard limit...)

dralley: It should always let one file through at a time no matter how large.  In theory

mdellweg: Nice, then we don't need to mention it.

docs/admin/reference/settings.md
pedro-psb, line 1: As this is user facing documentation I would prefer dropping specific references to `ArtifactDownloader` and `ArtifactSaver`.

docs/admin/reference/settings.md
pedro-psb: IMO it's worth adding a note that this does not prevent bigger to be synced.

dralley:
@balasankarc - Relevant to your previous pull requests

docs/admin/reference/settings.md
```diff
@@ -474,10 +474,38 @@ Defaults to `/var/lib/pulp/tmp/`.
...
+
+### SYNC\_MAX\_IN\_FLIGHT\_ITEMS
```
dralley, line 503: Actually the unit of measurement here is content, not artifacts, so this description is a little misleading - but it's a little difficult to describe.

docs/admin/reference/settings.md
dralley: Done

pulpcore/app/settings.py
```diff
@@ -403,7 +403,17 @@
 
 DOMAIN_ENABLED = False
 
-MAX_CONCURRENT_CONTENT = 25
+MAX_CONCURRENT_CONTENT = 200
```
dralley, line 406: Reset this back to where it was before we reduced it to limit worst-case scenarios

pulpcore/plugin/stages/artifact_stages.py
```diff
@@ -259,16 +387,28 @@ class ArtifactSaver(Stage):
...
     async def run(self):
         """
         The coroutine for this stage.
 
         Returns:
             The coroutine for this stage.
         """
-        async for batch in self.batches(minsize=settings.MAX_CONCURRENT_CONTENT):
+        flush_event = self.resource_budget.pressure if self.resource_budget else None
+        minsize = self.resource_budget.max_items if self.resource_budget else 200
```
dralley, line 410: This is a little weird because, on one hand actually the resource budget ought to manage this completely, and on the other hand it's hard to actually see the downside of setting a reasonably sized (fairly large) batch size by default?  It's going to exist anyway (defaults to 500 if you don't set it manually).

pulpcore/plugin/stages/artifact_stages.py
dralley, line 218: There's one open question, which is whether `max_concurrent_content` handling in `GenericDownloader` ought to be removed entirely. 

`ResourceBudget` accomplishes the same goal but slightly differently. `max_concurrent_content` blocks items from being pulled from the queue, before the `asyncio.Task` is created, while `ResourceBudget` blocks on `acquire()` after the task has been created but before it does work.

So I suppose `ResourceBudget` might accumulate `asyncio.Task` objects, which might waste a little but of memory but probably doesn't matter in practice?  I'm not that familiar with these aspects of the pipeline or asyncio, though.

Another option is to just set a fairly permissive cap as a safety net and otherwise let `ResourceBudget` handle it.

mdellweg: And this is not even the same as max_concurrent_download (handled inside aiohttp), right?

pulpcore/plugin/stages/api.py
```diff
@@ -124,13 +133,20 @@ def add_to_batch(content):
...
+                waitables.append(flush_event_listener)
+            done, pending = await asyncio.wait(waitables, return_when=asyncio.FIRST_COMPLETED)
             if thaw_event_listener in done:
                 thaw_event_listener = asyncio.ensure_future(thaw_queue_event.wait())
                 no_block = True
+            if flush_event_listener and flush_event_listener in done:
+                # Don't re-arm until after we yield a batch, to avoid a spin loop
+                # when the event stays set but the batch is empty.
+                flush_event_listener = None
+                no_block = True
```
mdellweg, line 149: What is the fundamental difference here between the two events?
If I read it correctly, the thaw_event is cleared right before yielding.
Can these events be consolidated into one?

pulpcore/plugin/stages/api.py
```diff
@@ -153,8 +169,13 @@ def add_to_batch(content):
                 yield batch
                 batch = []
                 no_block = False
+                # Re-arm the flush listener after yielding
+                if flush_event and flush_event_listener is None:
+                    flush_event_listener = asyncio.ensure_future(flush_event.wait())
```
mdellweg, line 174: Don't you need to `.clear()` the event to "rearm" it?

pulpcore/plugin/stages/artifact_stages.py
dralley: I assume you mean either `download_concurrency` or `rate_limit`, but yes.

Lots of these primitives are very closely related but not 100% identical and we could probably do with simplifying them.

I think, IIRC, `rate_limit` throttles the number of total simultaneous downloads and `download_concurrency` is the number of simultaneous connections allowed to any single host.

`MAX_CONCURRENT_CONTENT` / `max_concurrent_content` is "don't buffer too many items inside the download stage at once"

Also the former operates on individual downloads (artifacts) while the latter is content which may be multi-artifact.

The net impact of these is similar but not 100% identical

pulpcore/plugin/stages/api.py
dralley: It gets cleared in `release()`, `self.pressure` and `flush_event` are the same

pulpcore/plugin/stages/api.py
dralley: I'm not sure they can, I think they have different lifetimes and the way they get re-armed is different.

pulpcore/plugin/stages/api.py
mdellweg: Would renaming "pressure" to "flush_event" help making that clearer?

mdellweg (dismissed):
I think i can see now how this works.

There are many nobs to tweak, and we can probably not expect to find a one size-fits-all configuration.
One thing I keep wondering: Are there situations, where d_content is dropped between download and saving? Would we "leak" acquired resources that way?

pulpcore/plugin/stages/api.py
dralley: Maybe?

mdellweg (approved):

PR #7689, opened 2026-05-06
Fix add-signing-service failing for keys with subkeys

dralley:
The previous implementation counted fpr: lines in GPG's colon output to verify that exactly one key matched the provided key ID. However, GPG emits a separate fpr: line for the primary key and each subkey, so any key with subkeys would be rejected with "There are N keys matching the key id."

Count pub:/sec: lines instead, which represent actual distinct keys.

Assisted-By: claude-opus-4.6

### 📜 Checklist

### DIFF
--- a/CHANGES/+fix-add-signing-service.bugfix
+++ b/CHANGES/+fix-add-signing-service.bugfix
@@ -1,0 +1,1 @@
+Fixed `add-signing-service` management command failing with "There are N keys matching the key id" for PGP keys that have subkeys.
--- a/pulpcore/app/management/commands/add-signing-service.py
+++ b/pulpcore/app/management/commands/add-signing-service.py
@@ -82,10 +82,17 @@ class Command(BaseCommand):
         if result.returncode != 0:
             raise CommandError(result.stderr.strip())
 
-        fpr_lines = [line for line in result.stdout.splitlines() if line.startswith("fpr:")]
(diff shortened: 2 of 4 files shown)


mdellweg:
Just a thought i need to drop in here, because it's wildly related.

We always assumed that one singing service was attached to one public key. So all workers in the end needed to be able to create a signature for that one (Either all having the private key, or all able to talk to a device containing that key.).
So is it possible with subkeys, that you equip the workers only with different subkeys of the public key, so that on compromise you can revoke them individually?

dralley:
I believe that's the purpose of having the signing key be a subkey in the first place.

My understanding is that the public key under that regime is generally kept offline/cold-stored, used only for certification (signing other keys, creating/revoking subkeys), while the signing subkey(s) are actively used and can be revoked without torching the whole key bundle.

gpg lets you export secret keys independently

So I would think that is possible.  Though I think the more typical approach would still have the signing done by a specific signing server

dralley:
@mdellweg Do you think E741 should be disabled in linting?  IMO these lines are fine.

mdellweg:
> @mdellweg Do you think E741 should be disabled in linting? IMO these lines are fine.

I think that should be ok.
For the time being you can add `# noqa: E741` in the code.

mdellweg (approved):

PR #7708, opened 2026-05-12
 Enable selective repository sync for replication

dralley:
### 📜 Checklist

### DIFF
--- a/pulpcore/app/viewsets/replica.py
+++ b/pulpcore/app/viewsets/replica.py
@@ -8,7 +8,11 @@ from rest_framework.decorators import action
 
 from pulpcore.app.models import TaskGroup, UpstreamPulp
 from pulpcore.app.response import TaskGroupOperationResponse
-from pulpcore.app.serializers import TaskGroupOperationResponseSerializer, UpstreamPulpSerializer
+from pulpcore.app.serializers import (
+    TaskGroupOperationResponseSerializer,
+    UpstreamPulpReplicateSerializer,
+    UpstreamPulpSerializer,
+)
 from pulpcore.app.tasks import replicate_distributions
 from pulpcore.app.viewsets import NamedModelViewSet, RolesMixin
 from pulpcore.app.viewsets.base import DATETIME_FILTER_OPTIONS, NAME_FILTER_OPTIONS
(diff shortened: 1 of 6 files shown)


pulpcore/app/viewsets/replica.py
```diff
@@ -118,24 +122,31 @@ class UpstreamPulpViewSet(
         summary="Replicate",
         description="Trigger an asynchronous repository replication task group. This API is "
         "provided as a tech preview.",
-        request=None,
+        request=UpstreamPulpReplicateSerializer,
         responses={202: TaskGroupOperationResponseSerializer},
     )
-    @action(detail=True, methods=["post"])
+    @action(detail=True, methods=["post"], serializer_class=UpstreamPulpReplicateSerializer)
```
dralley, line 128: This is technically an API bindings-breaking change, but it may be the only real option short of creating a new replicate endpoint entirely

There is an intermediate commit (which should be squashed away after review) that attempted to make the request body optional, which broke a significant number of tests

https://github.com/pulp/pulpcore/actions/runs/25713087321

Claude provides this explanation:


> Commit 23e16bde8 broke CI because both RepositorySyncURLSerializer and RepositoryAddRemoveContentSerializer have zero required fields. When _get_request_body stopped marking their bodies as required: true in the OpenAPI schema, the openapi-generator shifted those body parameters from positional to keyword-only in the generated Python client bindings.

> So test code like:

```
file_bindings.RepositoriesFileApi.sync(repo.pulp_href, sync_data)  # positional
file_bindings.RepositoriesFileApi.modify(repo.pulp_href, body_dict)  # positional
```

> broke because the body param moved to keyword-only — the second positional arg now maps to a different parameter entirely, causing the pydantic type validation errors.

> This isn't just these two serializers — most Pulp serializers have all-optional fields, so the change would have affected a huge number of endpoints across pulpcore and all plugins. That's why the revert was the right call. If you want to revisit making request bodies conditionally optional in the future, it would require updating every test call site across all plugins to use keyword arguments for body params (e.g., sync(repo_href, repository_sync_url=sync_data)).

dralley: With that said I think changing the schema here is probably justified given the low number of users, some of whom are asking for this specifically.  Also I think it's still tech preview?


We might need to change pulp-cli / pulp-glue though

However we maybe should have a policy that in the future all API endpoints must come with a serializer, even an empty one, so that we can then extend without completely breaking the API schema.

There was another example of this a few weeks ago https://github.com/pulp/pulpcore/pull/7553#discussion_r3035988643

pulpcore/tests/functional/api/test_replication.py
```diff
@@ -36,7 +36,9 @@ def test_replication(
         pulpcore_bindings.UpstreamPulpsApi, upstream_pulp_body, pulp_domain=non_default_domain.name
     )
     # Run the replicate task and assert that all tasks successfully complete.
-    response = pulpcore_bindings.UpstreamPulpsApi.replicate(upstream_pulp.pulp_href)
+    response = pulpcore_bindings.UpstreamPulpsApi.replicate(
+        upstream_pulp.pulp_href, pulpcore_bindings.module.UpstreamPulpReplicate()
```
dralley, line 40: providing this extra argument now mandatory

@mdellweg Do you have any other suggestion on how to avoid this?

pulpcore/app/viewsets/replica.py
mdellweg: In OAS3.1 (didn't check other versions) there is a "required" parameter for the request body.
I'm certain, we can make it so spectacular announces that correctly, but can drf consume it, can the bindings generate such request?

https://swagger.io/specification/#request-body-object

dralley: I believe that was in fact what `23e16bde894561acfb404a5da72f27257398f7ed` was doing. It was marked as optional appropriately, but that legitimate schema change had other legitimate side effects, namely turning body which is otherwise a positional argument into a keyword argument.

mdellweg: That sounds like an issue with the bindings so not technically a breaking api change. Still pretty much annoying...

mdellweg (approved):

PR #7711, opened 2026-05-13
Fix additional issues with atomic replication

dralley:
In situations where an installation was replicating >1 external Pulp servers, the result was cascading failures as the replications would race and conflict with each other.

### 📜 Checklist

### DIFF
--- a/pulpcore/app/replica.py
+++ b/pulpcore/app/replica.py
@@ -187,13 +187,6 @@ class Replicator:
         }
 
     def create_or_update_distribution(self, repository, upstream_distribution):
-        """Create or update a local distribution to match the upstream distribution.
-
-        Returns the old name of the distribution if it was found by base_path instead of
-        by name (i.e. a rename occurred upstream). The caller must add this old name to
-        the distro_names list so that remove_missing does not race with the rename.
-        Returns None otherwise.
-        """
         distribution_data = self.distribution_extra_fields(repository, upstream_distribution)
         distribution_data["pulp_labels"] = self.labels(upstream_distribution)
         try:
@@ -201,7 +194,7 @@ class Replicator:
                 name=upstream_distribution["name"], pulp_domain=self.domain
             )
             if not self._is_managed(distro):
-                return None
+                return
             needs_update = self.needs_update(distribution_data, distro)
             if needs_update:
                 dispatch(
@@ -215,13 +208,15 @@ class Replicator:
                         "partial": True,
                     },
                 )
-            return None
+            return
         except self.distribution_model_cls.DoesNotExist:
             pass
 
         # Not found by name — check if a managed distribution with the same base_path
-        # already exists (e.g. the distribution was renamed upstream). If so, update it
-        # in-place including the name, preserving content continuity at that base_path.
+        # already exists (e.g. the distribution was renamed upstream). If so, rename it
+        # synchronously and then dispatch an update for the remaining fields (labels, etc.).
+        # The synchronous rename ensures remove_missing sees the new name in the DB and
+        # won't try to delete the distribution we're reusing.
         base_path = distribution_data.get("base_path")
         if base_path:
             try:
@@ -229,20 +224,26 @@ class Replicator:
                     base_path=base_path, pulp_domain=self.domain
                 )
                 if self._is_managed(distro):
-                    old_name = distro.name
-                    distribution_data["name"] = upstream_distribution["name"]
-                    dispatch(
-                        ageneral_update,
-                        task_group=self.task_group,
-                        shared_resources=[repository, self.server],
-                        exclusive_resources=self.distros_uris,
-                        args=(distro.pk, self.app_label, self.distribution_serializer_name),
-                        kwargs={
-                            "data": distribution_data,
-                            "partial": True,
-                        },
-                    )
-                    return old_name
+                    distro.name = upstream_distribution["name"]
+                    distro.save(update_fields=["name"])
+                    needs_update = self.needs_update(distribution_data, distro)
+                    if needs_update:
+                        dispatch(
+                            ageneral_update,
+                            task_group=self.task_group,
+                            shared_resources=[repository, self.server],
+                            exclusive_resources=self.distros_uris,
+                            args=(
+                                distro.pk,
+                                self.app_label,
+                                self.distribution_serializer_name,
+                            ),
+                            kwargs={
+                                "data": distribution_data,
+                                "partial": True,
+                            },
+                        )
+                    return
             except self.distribution_model_cls.DoesNotExist:
                 pass
 
@@ -256,7 +257,6 @@ class Replicator:
             args=(self.app_label, self.distribution_serializer_name),
             kwargs={"data": create_data},
         )
-        return None
 
     def sync_params(self, repository, remote):
         """This method returns a dict that will be passed as kwargs to the sync task."""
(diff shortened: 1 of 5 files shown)


dralley:
I recommend looking at the commits individually, and probably in 1, 3, 2 order

pulpcore/app/replica.py
```diff
@@ -17,6 +17,10 @@
 _logger = logging.getLogger(__name__)
 
 
+def distros_lock_uri(domain_id):
+    return f"pdrn:{domain_id}:distributions"
```
mdellweg, line 21: So this was never an uri to begin with? This function is solely used for locking, the domains distribution url space?
Is it the same lock used for updating distributions directly?

pulpcore/app/viewsets/replica.py
```diff
@@ -127,6 +127,18 @@ def replicate(self, request, pk):
...
+        if server.policy != UpstreamPulp.LABELED:
+            other_upstreams = UpstreamPulp.objects.filter(
+                pulp_domain=server.pulp_domain,
+            ).exclude(pk=server.pk)
+            if other_upstreams.exists():
+                raise serializers.ValidationError(
+                    "Multiple upstream-pulp configurations exist in this domain. "
+                    "Set the 'policy' to 'labeled' on all of them to avoid conflicts, "
+                    "or remove the extras."
+                )
```
mdellweg: This is rather opinionated. Are we sure we can take that "feature" away?

Are there possible usecases for having two conflicting upstream configurations in one domain?
- Something like Blue-Green upstream where you sync them alternatingly.
- Having a hot spare upstream in case the main one is out of reach.

pulpcore/app/tasks/replica.py
```diff
@@ -127,14 +131,25 @@ def finalize_replication(server_pk, distro_repo_pairs):
...
     server = UpstreamPulp.objects.get(pk=server_pk)
-    if task_group.tasks.exclude(pk=task.pk).exclude(state=TASK_STATES.COMPLETED).exists():
-        raise Exception("Replication failed.")
 
-    # Atomically update all managed distributions to point to their repo's latest version,
-    # clearing any previous repository or publication references.
+    failed_tasks = task_group.tasks.exclude(pk=task.pk).exclude(state=TASK_STATES.COMPLETED)
+    has_failures = failed_tasks.exists()
+
+    # Best-effort: update all distributions we can, even if some subtasks failed.
```
mdellweg: Isn't that the antithesis to atomicity?
I thought the idea is the domain (maybe scoped by labels...) looks exactly like upstream today, or nothing changes.

dralley: The main idea was just that successful updates should be synchronized rather than randomly applied over the course of an hour+.  That was the primary original complaint

pulpcore/app/viewsets/replica.py
dralley: I'm not 100% sure, I will ask them in our call today.

pulpcore/app/replica.py
dralley: I just noticed that similar lock strings were being created in multiple places, so I factored it into a helper, but the value should actually be the same - I believe `self.domain.pulp_id` == `server.pulp_domain_id` in this context.

pulpcore/app/viewsets/replica.py
dralley: I feel like having multiple potentially overlapping Upstreams replicating into the same domain is *probably* a very bad idea.  I don't see an obvious use case for it. With LABELED it would be allowed but at least you would get a better error in cases where they overlap instead of just wreaking havoc

edit: actually there is no error, they are skipped silently by the 2nd replication.  That's probably not ideal.

@gerrod3 should we be raising an error if a 2nd UpstreamPulp tries to touch objects created by a different UpstreamPulp with LABELED policy?

mdellweg (dismissed):

pulpcore/app/viewsets/replica.py
gerrod3: Maybe, my understanding was that the second upstream would become the new owner of that object as they replace it. This behavior might not be expected for users of the labeled policy. If we are changing the upstream existence policy then we probably should also change this overlap behavior for labeled.

gerrod3 (dismissed):

pulpcore/app/viewsets/replica.py
dralley: The description of "labeled" is

> Replication only manages objects that it created in a previous run. Objects created by replication
are tagged with an `UpstreamPulp` label linking them to the specific Upstream Pulp object. Manually
created local objects with the same content types are left untouched, even if they share names with
upstream distributions. If a replicated object has its `UpstreamPulp` label removed, replication
will no longer manage or delete it.

Which implies no replacing

dralley:
After discussion, commits 2 and 3 are likely not suitable for merge in their current state.

pulpcore/app/replica.py
```diff
@@ -208,17 +208,55 @@ def create_or_update_distribution(self, repository, upstream_distribution):
...
-                general_create,
-                task_group=self.task_group,
-                shared_resources=[repository, self.server],
-                exclusive_resources=self.distros_uris,
-                args=(self.app_label, self.distribution_serializer_name),
-                kwargs={"data": create_data},
-            )
+            pass
+
+        # Not found by name — check if a managed distribution with the same base_path
```
dralley, line 215: I think using the "name" as if it were a stable identifier to begin with is a problem.  Seems very fragile, as this issue would seem to prove...

I had an idea that perhaps we could add a second label marking the PRN of the original resource on the UpstreamPulp server.  That would be a stable identifier.  We could then special case that for the A/B failover use case so that if you change the replica from A to B, so long as B is also a replica, it could check the label and make the association.

It would not work for two entirely independent Pulps that were  otherwise identical.

pulpcore/app/tasks/replica.py
```diff
@@ -111,12 +112,17 @@ def replicate_distributions(server_pk, q_select=None):
...
-
                 # Add name to the list of known distribution names
                 distro_names.append(distro["name"])
                 distro_repo_pairs.append((distro["name"], str(repository.pk)))
+                pending_distributions.append((repository, distro))
+
+            # Get or create distributions BEFORE remove_missing so that
+            # create_or_update_distribution can synchronously rename any existing
+            # distribution matched by base_path.  remove_missing then sees the
+            # updated name in the DB and won't schedule it for deletion.
```
dralley, line 123: The crux of the issue:

* A distribution gets renamed upstream, it has  the same base_path as before
* Replication decides "this is a new distribution, let's create it"
* remove_missing hasn't happened yet, so this fails because two distributions cannot share the same base_path
* also it *cannot* happen beforehand, because the whole point of atomic distributions is that you don't interrupt content delivery if the replication fails 

Therefore, we use `base_path` as a way to say "this distribution is probably the same, just change the name instead of deleting it".  Therefore, content service doesn't get interrupted, and there are no `base_path` conflicts


........Ewwwwww, gross.  But I have no better ideas that can be implemented quickly given all of the design requirements.  It's worth a try, I guess?

Long term we need some better solution.  This is silly.

dralley:
I hate most of this, but it's the most straightforwards immediate solution to the user issues that I can see w/o redesigning most of replication, which we might still need to do at some point.

mdellweg (approved):

gerrod3 (approved):

dralley:
There was a "temp" commit at the end, which should have been squashed into the previous one, but it's fine to include in the sense that the commit diff was meant to be included.

PR #7720, opened 2026-05-14
Add NAME_FILTER_OPTIONS to FileContent relative_path filter

### LINKED ISSUE #7719: Add NAME_FILTER_OPTIONS to FileContent relative_path filter
## Problem

`FileContentFilter` in `pulp_file/app/viewsets.py` only supports exact matching on the `relative_path` field:

```python
class Meta:
    model = FileContent
    fields = ["relative_path", "sha256"]
```

This means API consumers cannot filter file content by path patterns — there is no support for `contains`, `startswith`, `regex`, or other lookup expressions on `relative_path`.

Other content filters in pulpcore already use `NAME_FILTER_OPTIONS` for name/path fields (e.g., `ContentFilter`), so `FileContentFilter` is inconsistent with the rest of the API.

## Proposed fix

Use `NAME_FILTER_OPTIONS` for `relative_path` to enable the full set of lookup expressions:

```python
from pulpcore.app.viewsets.base import NAME_FILTER_OPTIONS

class Meta:
    model = FileContent
    fields = {
        "relative_path": NAME_FILTER_OPTIONS,
        "digest": ["exact"],
    }
```

This enables: `exact`, `iexact`, `in`, `contains`, `icontains`, `startswith`, `istartswith`, `regex`, `iregex`.

The existing `sha256 = CharFilter(field_name="digest")` alias on the filter class continues to provide backwards-compatible `?sha256=` query parameter support.

## Use case

Deployments managing large file repositories need to search for content by path prefix (e.g., all files under `packages/python/`) or by pattern (e.g., all `.rpm` files). Currently this requires fetching all file content and filtering client-side.

dkliban:
## Summary
- Use `NAME_FILTER_OPTIONS` for `relative_path` in `FileContentFilter`, enabling contains, startswith, regex, and other lookup expressions
- Change `fields` from a list to a dict to specify per-field filter options
- `sha256` query parameter continues to work via the existing `CharFilter` alias

Fixes #7719

## Test plan
- [ ] Existing file content filter tests pass
- [ ] New filter lookups work: `?relative_path__contains=foo`, `?relative_path__startswith=packages/`, `?relative_path__regex=.*\.rpm$`

### DIFF
--- a/CHANGES/pulp_file/7719.feature
+++ b/CHANGES/pulp_file/7719.feature
@@ -1,0 +1,1 @@
+Added `NAME_FILTER_OPTIONS` to the `relative_path` field in `FileContentFilter`, enabling contains, startswith, regex, and other lookup expressions.
--- a/pulp_file/app/viewsets.py
+++ b/pulp_file/app/viewsets.py
@@ -9,6 +9,7 @@ from rest_framework.exceptions import ValidationError
 from rest_framework.generics import get_object_or_404
 from rest_framework.response import Response
 
+from pulpcore.app.viewsets.base import NAME_FILTER_OPTIONS
 from pulpcore.plugin.actions import ModifyRepositoryActionMixin
 from pulpcore.plugin.models import (
     AlternateContentSource,
@@ -66,7 +67,10 @@ class FileContentFilter(ContentFilter):
 
     class Meta:
         model = FileContent
-        fields = ["relative_path", "sha256"]
+        fields = {
+            "relative_path": NAME_FILTER_OPTIONS,
+            "digest": ["exact"],
+        }
 
 
 class FileContentViewSet(SingleArtifactContentUploadViewSet):
--- a/pulp_file/tests/functional/api/test_crud_content_unit.py
+++ b/pulp_file/tests/functional/api/test_crud_content_unit.py
@@ -35,6 +35,36 @@ def test_crud_content_unit(file_bindings, random_artifact, gen_object_with_clean
 
 
 @pytest.mark.parallel
+def test_relative_path_filter_options(file_bindings, random_artifact, gen_object_with_cleanup):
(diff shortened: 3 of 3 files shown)


CHANGES/pulp_file/7719.feature
```diff
@@ -0,0 +1,3 @@
+Added NAME_FILTER_OPTIONS to the ``relative_path`` field in ``FileContentFilter``,
```
gerrod3: Use single ticks, our changelogs are markdown. Also, make it all one line.

pulp_file/app/viewsets.py
```diff
@@ -66,7 +67,10 @@ class FileContentFilter(ContentFilter):
 
     class Meta:
         model = FileContent
-        fields = ["relative_path", "sha256"]
+        fields = {
+            "relative_path": NAME_FILTER_OPTIONS,
+            "digest": ["exact"],
```
gerrod3: This should still be sha256.
```suggestion
            "sha256": ["exact"],
```

gerrod3 (changes requested):

pulp_file/app/viewsets.py
dkliban: Reverted back to `"digest"`. When `Meta.fields` is a dict, django-filter requires model field names as keys — `sha256` is a declared `CharFilter` alias, not a model field, so it raises `TypeError: Meta.fields must not contain non-model field names: sha256`. The `sha256 = CharFilter(field_name="digest")` on the class still provides the `?sha256=` query parameter for backwards compatibility.

gerrod3 (approved):

dkliban:
The test failure is unrelated to these changes. It is being addressed in https://github.com/pulp/pulpcore/pull/7730/. I am merging this change.

PR #7722, opened 2026-05-15
Decouple /livez endpoint from the database

### LINKED ISSUE #7721: Decouple /livez endpoint from the database
## Problem

The `/livez` liveness probe fails when the database is unavailable, defeating its purpose. The `LivezView` itself has no DB dependency — it just returns `Response()` with empty `authentication_classes` and `permission_classes`. However, `DomainMiddleware.process_view()` runs `Domain.objects.get()` on every request, including `/livez/`, causing the probe to fail with a database error.

Other middleware in the stack (`SessionMiddleware`, `AuthenticationMiddleware`) are lazy and don't hit the DB unless `request.session` or `request.user` is accessed, which `LivezView` never does. `DomainMiddleware` is the only eager DB dependency.

## Proposed fix

Add a `skip_domain_middleware` marker to `LivezView` and teach `DomainMiddleware.process_view()` to skip the `Domain.objects.get()` lookup when the flag is set:

```python
# pulpcore/app/views/status.py
class LivezView(APIView):
    authentication_classes = []
    permission_classes = []
    skip_domain_middleware = True
```

```python
# pulpcore/middleware.py — DomainMiddleware.process_view()
def process_view(self, request, view_func, view_args, view_kwargs):
    view_class = getattr(view_func, "view_class", None) or getattr(view_func, "cls", None)
    if view_class and getattr(view_class, "skip_domain_middleware", False):
        view_kwargs.pop("pulp_domain", None)
        return None
    # ... existing Domain.objects.get() logic
```

## Use case

Kubernetes liveness probes should only check that the gunicorn worker is responsive. When the database is temporarily unavailable (maintenance, failover), pods should not be killed and restarted — the workers are still alive and will recover when the DB returns. The current behavior causes cascading pod restarts during DB outages.

dkliban:
## Summary
- Add `skip_domain_middleware = True` flag to `LivezView`
- Teach `DomainMiddleware.process_view()` to skip `Domain.objects.get()` when the flag is set
- The `/livez` probe now returns 200 even when the database is unavailable

Other middleware (`SessionMiddleware`, `AuthenticationMiddleware`) are lazy and don't hit the DB for `/livez` since `authentication_classes` and `permission_classes` are both empty and `request.user`/`request.session` are never accessed. `DomainMiddleware` was the only eager DB dependency.

Fixes #7721

## Test plan
- [ ] Existing `/livez` test passes (`test_status.py`)
- [ ] `/livez` returns 200 when database is stopped (manual verification)

### DIFF
--- a/CHANGES/7721.bugfix
+++ b/CHANGES/7721.bugfix
@@ -1,0 +1,1 @@
+Decoupled the `/livez` liveness probe from the database so it returns 200 even when the database is unavailable.
--- a/pulpcore/app/views/status.py
+++ b/pulpcore/app/views/status.py
@@ -145,6 +145,8 @@ class LivezView(APIView):
     # allow anyone to access the liveness api
     authentication_classes = []
     permission_classes = []
+    # Skip domain DB lookup so this probe stays independent of the database
+    skip_domain_middleware = True
 
     @extend_schema(
         summary="Inspect liveness of Pulp's REST API.",
--- a/pulpcore/middleware.py
+++ b/pulpcore/middleware.py
@@ -38,6 +38,10 @@ class DomainMiddleware:
 
     def process_view(self, request, view_func, view_args, view_kwargs):
         """Remove the domain name if present, called right before view_func is called."""
+        view_class = getattr(view_func, "view_class", None) or getattr(view_func, "cls", None)
+        if view_class and getattr(view_class, "skip_domain_middleware", False):
+            view_kwargs.pop("pulp_domain", None)
+            return None
         domain_name = view_kwargs.pop("pulp_domain", "default")
         try:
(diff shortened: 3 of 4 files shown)


CHANGES/7721.bugfix
```diff
@@ -0,0 +1,3 @@
+Decoupled the ``/livez`` liveness probe from the database by skipping
+``DomainMiddleware``'s ``Domain.objects.get()`` lookup, so the probe returns
+200 even when the database is unavailable.
```
gerrod3: Shorten this to one line and use single tick marks.

gerrod3 (changes requested):

gerrod3 (approved):

PR #7729, opened 2026-05-19
Fix race condition in resource_blocker test fixture

dkliban:
## Summary

The `resource_blocker` fixture dispatched a blocking task but yielded immediately without waiting for it to start running. With RedisWorker (Azure runner), the blocking task may not have acquired the resource lock before the test dispatches an immediate task on the same resource, causing the immediate task to complete instead of waiting.

Now polls the blocking task until it reaches `"running"` state before yielding, ensuring the resource lock is held.

Fixes flaky `TestImmediateTaskWithBlockedResource.test_executes_in_task_worker` on Azure runner.

## Test plan
- [ ] Azure runner (RedisWorker) CI passes — the previously flaky test should now be stable
- [ ] GHA runner (PulpcoreWorker) CI passes — no behavior change for fast task pickup

### DIFF
--- a/pulpcore/tests/functional/api/test_tasking.py
+++ b/pulpcore/tests/functional/api/test_tasking.py
@@ -613,6 +613,13 @@ def resource_blocker(pulpcore_bindings, dispatch_task):
             args=(duration,),
             exclusive_resources=exclusive_resources,
         )
+        # Wait for the blocking task to start running so it holds the resource lock
+        for _ in range(100):
+            task = pulpcore_bindings.TasksApi.read(task_href)
+            if task.state == "running":
+                break
+            time.sleep(0.1)
+        assert task.state == "running", f"Blocker task never started: {task.state}"
         yield
         # Trying to cancel a finished task will return a 409 code.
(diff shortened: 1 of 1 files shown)


pulpcore/tests/functional/api/test_tasking.py
```diff
@@ -613,6 +613,13 @@ def _resource_blocker(exclusive_resources: list[str], duration=20):
             args=(duration,),
             exclusive_resources=exclusive_resources,
         )
+        # Wait for the blocking task to start running so it holds the resource lock
+        for _ in range(100):
+            task = pulpcore_bindings.TasksApi.read(task_href)
+            if task.state == "running":
```
ggainey, line 619: You could check for not-waiting instead - that would break out of the loop immediately, and the assert would fail if we break for failed/completed/canceled. Otherwise, any failure forces a 100-second wait.

ggainey (changes requested):

dkliban:
Closing in favor of #7730 which fixes the actual bug in the RedisWorker dispatch logic rather than papering over it in the test fixture.

PR #7750, opened 2026-05-27
Strip version info from status endpoint for unauthenticated requests

lucasc017:
## Summary

- Remove `authentication_classes = []` from `StatusView` so DRF runs its default authenticators while keeping `permission_classes = []` (endpoint stays accessible to everyone)
- Strip the `versions` field from the `/pulp/api/v3/status/` response for unauthenticated requests to prevent version disclosure
- Make `versions` optional in `StatusSerializer` and update functional tests accordingly

## Context

A penetration test found that the status endpoint discloses exact versions of every installed Pulp plugin to unauthenticated users. Attackers can use exact version numbers for targeted CVE lookups. Healthcheck probes (e.g. `readyz.py`) only need `database_connection` and `redis_connection`, so stripping `versions` for unauthenticated requests has no impact on them.

## Test plan

- [x] Verified unauthenticated requests to `/pulp/api/v3/status/` no longer contain `versions`
- [x] Verified authenticated requests still return full `versions` array
- [x] Verified all healthcheck fields (`database_connection`, `redis_connection`, `online_workers`, `storage`, etc.) remain present regardless of auth
- [ ] CI functional tests (require regenerated client bindings from updated OpenAPI spec)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pulpcore/app/serializers/status.py
+++ b/pulpcore/app/serializers/status.py
@@ -96,7 +96,9 @@ class StatusSerializer(serializers.Serializer):
     Serializer for the status information of the app
     """
 
-    versions = VersionSerializer(help_text=_("Version information of Pulp components"), many=True)
+    versions = VersionSerializer(
+        help_text=_("Version information of Pulp components"), many=True, required=False
+    )
 
     online_workers = AppStatusSerializer(
         help_text=_(
--- a/pulpcore/app/views/status.py
+++ b/pulpcore/app/views/status.py
@@ -38,8 +38,6 @@ class StatusView(APIView):
     Returns status information about the application
     """
 
-    # allow anyone to access the status api
-    authentication_classes = []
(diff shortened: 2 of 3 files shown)


mdellweg:
Hi thank you to bringing this aspect to our attention.
I see two problematic angles to this approach.
First: I don't think we should change the appearance of the status serializer in such a backwards incompatible way. The fact that it is accessible without authentication is due to it being older than "readyz". The way I could see this going forward is changing status to be equipped with an access policy so an administrator can make the choice to hide it.
Second: That same set of component versions is an integral part of the openapi.json and I'm not sure we can properly hide that either.

mdellweg:
After some discussion we decided that this falls into the bucket of security by obscurity and does not warrant an incompatible api change.

PR #7751, opened 2026-05-28
Add documentation for per-request replication q-select feature

dralley:
Assisted-By: claude-opus-4.6

### 📜 Checklist

### DIFF
--- a/docs/user/guides/replication.md
+++ b/docs/user/guides/replication.md
@@ -113,6 +113,41 @@ To clear the filter and replicate all distributions again:
 pulp upstream-pulp update --upstream-pulp "my-upstream" --q-select ""
 ```
 
+### Per-Request Filtering
+
+You can also pass `q_select` directly to the replicate action as a one-time override without
+modifying the stored Upstream Pulp configuration. This is useful when you need to prioritize syncing
+a subset of distributions in time-sensitive situations.
+
+```bash
+# Sync only distributions labeled 'critical' for this run
+pulp upstream-pulp replicate \
+    --upstream-pulp "my-upstream" \
+    --q-select "pulp_label_select='critical'"
+```
+
+When a per-request `q_select` is provided:
+
+- It overrides the stored `q_select` for that replication run only. The stored configuration is not
+  changed.
+- Cleanup of missing distributions is **skipped**. Because the filter selects only a subset,
+  distributions not included in the filter are left untouched rather than being treated as removed.
+  This means distributions that have actually been deleted upstream will not be cleaned up until a
+  full (non-overridden) replication runs.
+
+You can run multiple selective replications to incrementally build up local state, then run a full
+replication to reconcile everything:
+
+```bash
+# Selectively sync high-priority repos first
+pulp upstream-pulp replicate \
+    --upstream-pulp "my-upstream" \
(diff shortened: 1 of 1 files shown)


docs/user/guides/replication.md
```diff
@@ -113,6 +113,41 @@ To clear the filter and replicate all distributions again:
 pulp upstream-pulp update --upstream-pulp "my-upstream" --q-select ""
 ```
 
+### Per-Request Filtering
+
+You can also pass `q_select` directly to the replicate action as a one-time override without
+modifying the stored Upstream Pulp configuration. This is useful when you need to prioritize syncing
+a subset of distributions in time-sensitive situations.
```
mdellweg, line 120: I'd like us to stick to this kind of formatting, aligning linebreaks with syntax structure:
```suggestion
You can also pass `q_select` directly to the replicate action as a one-time override without modifying the stored Upstream Pulp configuration.
This is useful when you need to prioritize syncing a subset of distributions in time-sensitive situations.
```

docs/user/guides/replication.md
```diff
@@ -113,6 +113,41 @@ To clear the filter and replicate all distributions again:
...
+```
+
+When a per-request `q_select` is provided:
+
+- It overrides the stored `q_select` for that replication run only. The stored configuration is not
+  changed.
+- Cleanup of missing distributions is **skipped**. Because the filter selects only a subset,
+  distributions not included in the filter are left untouched rather than being treated as removed.
+  This means distributions that have actually been deleted upstream will not be cleaned up until a
+  full (non-overridden) replication runs.
```
mdellweg, line 136: We could use the stored `q_select` for cleanup. It really is hard to decide what was most likely wanted "here".

But also that bears the question. What if there was no stored q_select? What if the provided q_select is not a subset of the  stored q_select? (I think I know what happens currently, but is it the right thing to do?)

There might not be an answer to this comment, so feel free to ignore it.

docs/user/guides/replication.md
dralley: The rest of this file doesn't really follow that convention, it linebreaks around column 100.  Might be better to come back through in a separate PR and do so across the whole "page".  Or, I guess I could add a second commit which does so for all of replication.md, if you want.

docs/user/guides/replication.md
dralley: I think that probably doesn't make sense, because then you're impacting content distribution for repos that were not included in the `q_select`.  Cleaning up distributions == breaking downstream users of those `base_path`s, without creating any new Distributions that might otherwise displace them.

(I guess that last bit is not as relevant with this patch, since we're matching on `base_path` and doing a rename - but then, you'd end up changing objects beyond the scope of the `q_select`, and all this feels like too much compared to just not touching them at all).

docs/user/guides/replication.md
mdellweg: well probably needs an attempt across the docs then. Never mind.

docs/user/guides/replication.md
mdellweg: Yeah, as I said. There probably is no right answer. At least we document how we do it.

mdellweg (approved):

Issue #7558, opened 2026-04-06
Speed the file replication by comparing the 2 PULP_MANIFEST files and calculte the diferenes


mchithu:
**Is your feature request related to a problem? Please describe.**
In a file repository there are 10K files. Even when a file content is changed, the file repository sync takes longer time.

**Describe the solution you'd like**
Compare the two PULP_MANIFEST file from the remote upstream-pulp server and local pulp server. Just perform a string based comparision between the files. And do the necessary add/remove the files from the repository. Instead of going through all 10K checksum, file name and repository verion, the string comparision will complete the replication faster. Also provide an attribute, to run a full file repository sync when user requested for. If not explicitly mention, do a sync of only changes.

dralley:
Instead of doing a string-based comparison (which would require keeping a copy of the old manifest), we could just keep a cache of content models (keyed by `content.natural_key()`) generated from the contents of the previous repository version.

That would remove a fair number of required database queries from the [`QueryExistingContents`](https://github.com/pulp/pulpcore/blob/main/pulpcore/plugin/stages/content_stages.py#L14-L68) stage, because you can just do the one query up-front and get all of the content model swapping done in-memory.  If you expect high correlation between the previous repository version and the new one, that ought to provide a significant speedup.

`QueryExistingContents` would then skip over any content models which it sees as being marked as already having come from the database as opposed to being saved (`<model_obj>._state.adding`)

This could be done in the plugin-defined first stage, or possibly in the `QueryExistingContents` stage itself.  The former might provide more optimization opportunities, the latter would avoid needing to re-implement for multiple plugins.

PR #7754, opened 2026-05-29
[PULP-1668] Optimize re-sync with content caching

jobselko:
Related to #7558 

### 📜 Checklist

### DIFF
--- a/CHANGES/+optimize_resync.feature
+++ b/CHANGES/+optimize_resync.feature
@@ -1,0 +1,1 @@
+Improved sync performance by caching content during re-syncs.
--- a/pulpcore/plugin/stages/content_stages.py
+++ b/pulpcore/plugin/stages/content_stages.py
@@ -30,7 +30,37 @@ class QueryExistingContents(Stage):
 
     This stage drains all available items from `self._in_q` and batches everything into one large
     call to the db for efficiency.
+
+    When `repo_version` is provided, content from that version is cached in memory on first
+    access (per content type) so that repeat syncs can resolve most items without per-batch
+    database queries.
+
+    Plugins with expensive fields that are not needed during sync can pass
+    `deferred_fields` - a mapping of content model class to a list of field names
+    to exclude from queries via Django's `defer()`.
     """
+
+    def __init__(self, repo_version=None, deferred_fields=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._repo_version = repo_version
+        self._deferred_fields = deferred_fields or {}
+        self._content_cache = {}
+
+    def _ensure_type_cache(self, model_type):
+        """
+        Populate the cache for model_type from repo_version on first access.
+
+        Args:
+            model_type: The content model class to cache.
+        """
+        if model_type not in self._content_cache and self._repo_version is not None:
+            deferred = self._deferred_fields.get(model_type, ())
+            cache = {}
+            for content in self._repo_version.get_content(
+                model_type.objects.defer(*deferred)
+            ).iterator():
+                cache[content.natural_key()] = content
+            self._content_cache[model_type] = cache
 
     async def run(self):
         """
@@ -42,25 +72,46 @@ class QueryExistingContents(Stage):
         async for batch in self.batches():
             content_q_by_type = defaultdict(lambda: Q(pk__in=[]))
             d_content_by_nat_key = defaultdict(list)
+            cache_hits_by_type = defaultdict(set)
+
             for d_content in batch:
                 if d_content.content._state.adding:
                     model_type = type(d_content.content)
-                    unit_q = d_content.content.q()
-                    content_q_by_type[model_type] = content_q_by_type[model_type] | unit_q
-                    d_content_by_nat_key[d_content.content.natural_key()].append(d_content)
+                    nat_key = d_content.content.natural_key()
+                    await sync_to_async(self._ensure_type_cache)(model_type)
+                    cached = self._content_cache.get(model_type, {}).get(nat_key)
+                    if cached is not None:
+                        d_content.content = cached
+                        cache_hits_by_type[model_type].add(cached.pk)
+                    else:
+                        unit_q = d_content.content.q()
+                        content_q_by_type[model_type] = content_q_by_type[model_type] | unit_q
+                        d_content_by_nat_key[nat_key].append(d_content)
 
+            db_results_by_type = defaultdict(list)
             for model_type, content_q in content_q_by_type.items():
-                try:
-                    await sync_to_async(model_type.objects.filter(content_q).touch)()
-                except AttributeError:
-                    raise TypeError(
-                        "Plugins which declare custom ORM managers on their content classes "
-                        "should have those managers inherit from "
-                        "pulpcore.plugin.models.ContentManager."
-                    )
+                deferred = self._deferred_fields.get(model_type, ())
                 async for result in sync_to_async_iterable(
-                    model_type.objects.filter(content_q).iterator()
+                    model_type.objects.filter(content_q).defer(*deferred).iterator()
                 ):
+                    db_results_by_type[model_type].append(result)
+
+            all_types = set(cache_hits_by_type.keys()) | set(db_results_by_type.keys())
+            for model_type in all_types:
+                pks = cache_hits_by_type.get(model_type, set())
+                pks |= {r.pk for r in db_results_by_type.get(model_type, [])}
+                if pks:
+                    try:
+                        await sync_to_async(model_type.objects.filter(pk__in=pks).touch)()
+                    except AttributeError:
+                        raise TypeError(
+                            "Plugins which declare custom ORM managers on their content classes "
+                            "should have those managers inherit from "
+                            "pulpcore.plugin.models.ContentManager."
+                        )
+
+            for model_type, results in db_results_by_type.items():
+                for result in results:
                     for d_content in d_content_by_nat_key[result.natural_key()]:
                         d_content.content = result
 
--- a/pulpcore/plugin/stages/declarative_version.py
+++ b/pulpcore/plugin/stages/declarative_version.py
@@ -21,7 +21,7 @@ from pulpcore.plugin.util import get_domain_pk
 
 
 class DeclarativeVersion:
-    def __init__(self, first_stage, repository, mirror=False, acs=False):
+    def __init__(self, first_stage, repository, mirror=False, acs=False, deferred_fields=None):
         """
         A pipeline that creates a new [pulpcore.plugin.models.RepositoryVersion][] from a
         stream of [pulpcore.plugin.stages.DeclarativeContent][] objects.
@@ -107,12 +107,16 @@ class DeclarativeVersion:
                 'False' is the default.
             acs (bool): When set to 'True' a new stage is added to look for
                 Alternate Content Sources.
+            deferred_fields (dict): A mapping of content model class to a list of
+                field names to exclude from queries via Django's `defer()`.
+                Passed through to [pulpcore.plugin.stages.QueryExistingContents][].
 
         """
         self.first_stage = first_stage
         self.repository = repository
         self.mirror = mirror
         self.acs = acs
+        self.deferred_fields = deferred_fields
 
     def pipeline_stages(self, new_version):
         """
@@ -142,7 +146,9 @@ class DeclarativeVersion:
             [
                 ArtifactDownloader(resource_budget=resource_budget),
                 ArtifactSaver(resource_budget=resource_budget),
-                QueryExistingContents(),
+                QueryExistingContents(
+                    repo_version=new_version, deferred_fields=self.deferred_fields
+                ),
                 ContentSaver(),
                 RemoteArtifactSaver(),
                 ResolveContentFutures(),
--- a/CHANGES/plugin_api/+optimize_resync.feature
+++ b/CHANGES/plugin_api/+optimize_resync.feature
@@ -1,0 +1,1 @@
+`DeclarativeVersion` now accepts `deferred_fields` to exclude fields not needed during sync, reducing memory usage and producing lighter queries.


pulpcore/plugin/stages/content_stages.py
```diff
@@ -32,6 +32,27 @@ class QueryExistingContents(Stage):
     call to the db for efficiency.
     """
 
+    def __init__(self, repo_version=None, deferred_fields=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._repo_version = repo_version
+        self._deferred_fields = deferred_fields or {}
```
dralley, line 46: Woudn't this be the opposite of "deferred"?  It's the list of fields we want to query up-front for each type, right?

pulpcore/plugin/stages/content_stages.py
```diff
@@ -42,12 +63,31 @@ async def run(self):
...
                     model_type = type(d_content.content)
-                    unit_q = d_content.content.q()
-                    content_q_by_type[model_type] = content_q_by_type[model_type] | unit_q
-                    d_content_by_nat_key[d_content.content.natural_key()].append(d_content)
+                    nat_key = d_content.content.natural_key()
+                    await sync_to_async(self._ensure_type_cache)(model_type)
+                    cached = self._content_cache.get(model_type, {}).get(nat_key)
+                    if cached is not None:
+                        d_content.content = cached
+                        cache_hits_by_type[model_type] |= Q(pk=cached.pk)
```
dralley: Is there a reason to do it this way instead of just collecting a list of PKs and passing `pk__in=...`?

(there could be, probably the constructed SQL is different, I just don't know)

pulpcore/plugin/stages/content_stages.py
```diff
@@ -42,12 +63,31 @@ async def run(self):
         async for batch in self.batches():
             content_q_by_type = defaultdict(lambda: Q(pk__in=[]))
             d_content_by_nat_key = defaultdict(list)
+            cache_hits_by_type = defaultdict(lambda: Q(pk__in=[]))
+
             for d_content in batch:
                 if d_content.content._state.adding:
```
dralley, line 78: Theoretically content could be passed through already saved, in which case I think we're probably not `touch` ing it.  That might be an existing bug, though not a particularly serious one.

pulpcore/plugin/stages/content_stages.py
dralley: This is really an existing issue, but we're executing the `content_q` query (w/ natural keys) twice (once to `touch`, and once for the model swap), and also executing a `touch` query twice (once with cache hits, once with existing packages that were not in the latest version cache).

Is it possible to collect PKs within the loop, combine them with the cache hit PKs, and have one `touch` block below this loop?

Or (maybe question for @mdellweg), does that `touch` really need to happen before the swap for a timing-related reason?

dralley:
Looks good!  A few things to look at, maybe we can make this even more efficient

pulpcore/plugin/stages/content_stages.py
jobselko: Agree. The name I chose is misleading, changed to `extra_fields`

pulpcore/plugin/stages/content_stages.py
jobselko: No, I just copied the logic from `content_q_by_type`. Changed

pulpcore/plugin/stages/content_stages.py
jobselko: Updated. I kept `touch` before swap for now

pulpcore/plugin/stages/content_stages.py
jobselko: I am not sure I understand this. Why do we need to call `touch` when content already exists? Is it because of orphan clean up?

dralley: Yes, it basically resets the orphan cleanup protection timer.

jobselko:
Measured benefit using https://fixtures.pulpproject.org/file-perf/ (20k files) with **immediate** policy sync:

**Without patch:**
Sync time: 508 seconds
Re-sync time: 569 seconds

**With patch:**
Sync time: 513 seconds
Re-sync time: 68 seconds

The biggest impact comes from the `QueryExistingArtifacts` fix.

Syncing with **on-demand** policy is already fast and is not significantly affected by this change (Sync time: 92 seconds, Re-sync time: 42 seconds).

jobselko:
@dralley I will squash the commits before merging, but I am keeping them separate for easier review now.

dralley:
@jobselko Keep the `QueryExistingArtifacts` commit separate and split it off into an independent PR.  I think we might end up backporting that one, because it's very self-contained and has a high impact.  Satellite might want it.

The rest can be squashed and stay in this PR.

Great work!  This is a huge performance improvement 🚀 👍

pulpcore/plugin/stages/content_stages.py
```diff
@@ -88,29 +88,30 @@ async def run(self):
...
+                        await sync_to_async(model_type.objects.filter(pk__in=pks).touch)()
+                    except AttributeError:
+                        raise TypeError(
+                            "Plugins which declare custom ORM managers on their content classes "
+                            "should have those managers inherit from "
+                            "pulpcore.plugin.models.ContentManager."
+                        )
+
+            for model_type, results in db_results_by_type.items():
+                for result in results:
```
dralley, line 114: A little worried that the additional loops here would outweigh the reduced # of queries

pulpcore/plugin/stages/content_stages.py
```diff
@@ -32,6 +32,27 @@ class QueryExistingContents(Stage):
...
+    def __init__(self, repo_version=None, deferred_fields=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._repo_version = repo_version
+        self._deferred_fields = deferred_fields or {}
+        self._content_cache = {}
+
+    def _fields_for_type(self, model_type):
+        base = ("pk",) + model_type._sanitized_natural_key_fields()
+        extra = self._deferred_fields.get(model_type, ())
+        return base + tuple(f for f in extra if f not in base)
```
dralley: So,there's 2 ways to implement this behavior:

1) Creating a list of fields to pass to `.only()`
2) Creating a list of fields to pass to `.defer()`

Thinking about this a bit more, `deferred_fields` might have been the right idea, just the wrong implementation.  

It's really only a couple of fields that are known to be really expensive, and it's probably easier and safer for plugins to just opt out of those specific fields (if the sync pipeline doesn't need them) than to just use a heuristic like "plugins probably only need the natural key fields", which might require doing a bit of an audit to make sure that's actually true, otherwise it might actually cause performance regressions.

So instead of trying to calculate the list of extra fields, it might be simpler to just take a list of deferred fields and pass it directly to `.defer()` (not `.only()`)

The main concern would be: if a plugin turned on caching without opting into some of this, it might increase memory consumption during the sync due to the cache being immediately pre-populated and held for the entire sync.  I don't know that that would be a concern for any plugin that doesn't already define their own pipeline though (who would need to switch this on deliberately anyway)

What do you think?

pulpcore/plugin/stages/content_stages.py
```diff
@@ -58,8 +98,9 @@ async def run(self):
                         "should have those managers inherit from "
                         "pulpcore.plugin.models.ContentManager."
                     )
+                fields = self._fields_for_type(model_type)
```
dralley: Could do

```python
                deferred_fields = self._deferred_fields_for_model(model_type)  # which is a simpler function to implement
                async for result in sync_to_async_iterable(
                    model_type.objects.filter(content_q).defer(*deferred_fields).iterator()
```

dralley: Might want to double check what happens when passing `.defer()` an empty set of arguments though

pulpcore/plugin/stages/content_stages.py
dralley: could do something like

```python
    def _deferred_fields_for_type(self, model_type):
        self._deferred_fields[model_type] if self._deferred_fields else []
```

pulpcore/plugin/stages/content_stages.py
jobselko: I can rework it, but that would result in "swap before touch". Alternatively, I can revert the last commit and keep the two queries. Which option would you prefer?

pulpcore/plugin/stages/content_stages.py
jobselko: I checked that calling `.defer()` without arguments loads all fields

pulpcore/plugin/stages/content_stages.py
jobselko: Agree that `.defer()` seems to be better than `.only()`. Since caching is turned on by default in `QueryExistingContents`, I exposed `deferred_fields` on `DeclarativeVersion`, so plugins can pass it without overriding `pipeline_stages()`. Does it sound good?

pulpcore/plugin/stages/content_stages.py
dralley: As long as it doesn't fail

pulpcore/plugin/stages/content_stages.py
dralley: Yes, this looks good.  Simpler and less risky, +1

pulpcore/plugin/stages/content_stages.py
dralley: @mdellweg I feel like it was either you or Dennis that dealt with the `touch` issues last, do you know if `touch` need to happen before the existing model swap for correctness reasons?

dralley:
@jobselko Can you squash commits 1, 2, and 4, and address the merge conflict?

I think it's just the `wip - reduce redundant queries` that needs more attention before merge.  We could make that a separate PR also, to get everything else merged today for the release?

jobselko:
Discussed with @dralley that we are not going to merge this today.

pulpcore/plugin/stages/content_stages.py
```diff
@@ -42,25 +66,46 @@ async def run(self):
...
-                        "Plugins which declare custom ORM managers on their content classes "
-                        "should have those managers inherit from "
-                        "pulpcore.plugin.models.ContentManager."
-                    )
+                deferred = self._deferred_fields.get(model_type, ())
                 async for result in sync_to_async_iterable(
-                    model_type.objects.filter(content_q).iterator()
+                    model_type.objects.filter(content_q).defer(*deferred).iterator()
                 ):
+                    db_results_by_type[model_type].append(result)
```
gerrod3, line 97:
```suggestion
                    db_results_by_type[model_type].append(result)
                    for d_content in d_content_by_nat_key[result.natural_key()]:
                        d_content.content = result
```
Move this here and delete the added loops on line 107-108.

pulpcore/plugin/stages/content_stages.py
```diff
@@ -42,25 +66,46 @@ async def run(self):
...
+
+            all_types = set(cache_hits_by_type.keys()) | set(db_results_by_type.keys())
+            for model_type in all_types:
+                pks = cache_hits_by_type.get(model_type, set())
+                pks = pks | {r.pk for r in db_results_by_type.get(model_type, [])}
```
gerrod3:
```suggestion
                pks |= {r.pk for r in db_results_by_type.get(model_type, [])}
```

pulpcore/plugin/stages/content_stages.py
```diff
@@ -30,8 +30,32 @@ class QueryExistingContents(Stage):
...
     This stage drains all available items from `self._in_q` and batches everything into one large
     call to the db for efficiency.
+
+    When `repo_version` is provided, content from that version is cached in memory on first
+    access (per content type) so that repeat syncs can resolve most items without per-batch
+    database queries.
+
+    Plugins with expensive fields that are not needed during sync can pass
+    `deferred_fields` - a mapping of content model class to a list of field names
+    to exclude from queries via Django's `defer()`.
```
gerrod3, line 40: Is `deferred` the right way? Models can have a lot of fields, but only a few are ever used during a sync. Wouldn't `only` be the more practical option?

dralley: See this discussion (hidden because it was marked resolved) https://github.com/pulp/pulpcore/pull/7754#discussion_r3345783904

TL;DR yes `.only()` could be theoretically more efficient, but it could also easily cause a regression without plugin intervention.  You have to keep track of which fields are being used rigorously.

OTOH `.defer()` would not cause a regression, and since it's really only a small number of fields on a small number of models that are a problem, it is less invasive.

pulpcore/plugin/stages/content_stages.py
jobselko: But this would cause "swap before touch" which we wanted to avoid.

dralley: We need to verify that it's OK to swap before touching.  It MIGHT be ok, but IIRC the `touch` stuff was particularly subtle and the order it was done in might matter.

pulpcore/plugin/stages/content_stages.py
jobselko: Keeping it as is. We discussed this on the call.

pulpcore/plugin/stages/content_stages.py
jobselko: Keeping it as is. We discussed this on the call.

CHANGES/+optimize_resync.feature
```diff
@@ -0,0 +1 @@
+Added content caching in `QueryExistingContents` to speed up re-syncs. Plugins with large content fields can pass `deferred_fields` to `DeclarativeVersion` to reduce memory usage.
```
dralley: Move the "plugins" note to a separate changelog entry under `CHANGES/plugin_api`.  You can make that one a little more specific / detailed also.  e.g. the fields passed should be fields not used elsewhere in the sync pipeline, and the impact is not just less memory but also lighter queries.

pulpcore/plugin/stages/content_stages.py
```diff
@@ -30,8 +30,32 @@ class QueryExistingContents(Stage):
...
     """
 
+    def __init__(self, repo_version=None, deferred_fields=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._repo_version = repo_version
+        self._deferred_fields = deferred_fields or {}
+        self._content_cache = {}
+
+    def _ensure_type_cache(self, model_type):
```
dralley, line 49: A comment here might be good.

pulpcore/plugin/stages/declarative_version.py
```diff
@@ -142,7 +146,7 @@ def pipeline_stages(self, new_version):
             [
                 ArtifactDownloader(resource_budget=resource_budget),
                 ArtifactSaver(resource_budget=resource_budget),
-                QueryExistingContents(),
+                QueryExistingContents(new_version, self.deferred_fields),
```
dralley: This works fine but since they're keyword arguments it's best to just be explicit, e.g. the lines above

CHANGES/+optimize_resync.feature
dralley: Also you don't really need to mention `QueryExistingContents` in the user-facing changelog.  It's just a performance improvement due to caching during syncs.

dralley (approved):

PR #7770, opened 2026-06-04
Dependabot: Add one week cooldown

daviddavis:
More info about cooldowns and reasoning behind using them: https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns


mdellweg:
Or here: https://opensourcesecurity.io/2025/2025-12-renovate-jamie/

mdellweg (approved):

PR #7771, opened 2026-06-04
Use a CS10-based CI image

dralley:


pyproject.toml
```diff
@@ -55,7 +55,7 @@ dependencies = [
   "protobuf>=4.21.1,<7.0",
   "pulp-glue>=0.30.0,<0.40",
   "pygtrie>=2.5,<=2.5.0",
-  "psycopg[binary]>=3.1.8,<3.4",  # SemVer, not explicitely stated, but mentioned on multiple changes.
+  "psycopg[binary]>=3.3.4,<3.4",  # SemVer, not explicitely stated, but mentioned on multiple changes.
```
gerrod3, line 57: Does cs10 require a bump in this requirement?

dralley: The lowerbounds PR runner was failing because `psycopg[binary]` didn't have pre-built wheels for Python 3.12

With that said the lowerbounds probably does not need to be bumped *this* much and we can relax it if we need to,  I just wanted to move past that issue.

dralley: OTOH, pip installs are sticky so users will never get an upgrade unless we bump the minimum bound.  Might be worth just doing it.

mdellweg: If we cannot test it we should most definitely bump it.

dkliban (approved):

PR #7776, opened 2026-06-07
Use ThreadedResolver to honor /etc/hosts and nsswitch.conf

dralley:
Pulpcore depended on aiodns, which causes aiohttp to default to AsyncResolver (c-ares) for DNS resolution. c-ares is a DNS client that speaks the wire protocol directly -- it does not go through glibc's getaddrinfo() and therefore ignores nsswitch.conf and /etc/hosts. This causes sync to fail when the remote hostname is only resolvable via /etc/hosts or other non-DNS nsswitch sources.

Explicitly use aiohttp.resolver.ThreadedResolver, which wraps socket.getaddrinfo() in a thread pool. Since getaddrinfo() uses the system resolver, it respects nsswitch.conf ordering and reads /etc/hosts when configured.

The performance difference is negligible for Pulp's use case: DNS resolution happens once per host (aiohttp caches resolved addresses on the connector), and syncs typically talk to a small number of distinct hosts.

Since we now bypass c-ares entirely, drop the aiodns dependency and the related pycares version constraint from CI.

### 📜 Checklist

### DIFF
--- a/.ci/assets/ci_constraints.txt
+++ b/.ci/assets/ci_constraints.txt
@@ -15,5 +15,3 @@ azure-storage-blob!=12.28.*
 # Apparently does not work with current azurite.
 
 
-pycares<5
-# older aiodns versions don't pin pycares UB, and are broken by pycares>=5
--- a/pulpcore/download/factory.py
+++ b/pulpcore/download/factory.py
@@ -144,7 +144,13 @@ class DownloaderFactory:
         # I don't see why...
         # https://github.com/aio-libs/aiohttp/pull/3372
         return aiohttp.ClientSession(
-            connector=aiohttp.TCPConnector(loop=asyncio.get_event_loop(), **tcp_conn_opts),
+            # ThreadedResolver uses getaddrinfo() which honors /etc/hosts and nsswitch.conf;
+            # the default AsyncResolver (c-ares) bypasses the system resolver entirely.
+            connector=aiohttp.TCPConnector(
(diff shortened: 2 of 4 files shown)


dralley:
My understanding is that we've had a number of issues with `aiodns` and `pycares` - as best I can tell we only ever depended on `aiodns` for purely theoretical performance reasons.  But at least for syncs, you really only expect to talk to one host, or maybe a small handful, so `aiodns` provides no benefit.

Original issue: https://pulp.plan.io/issues/6190

I don't think we ever measured a practical benefit nor would one really be expected.  But in this case we're being asked explicitly to follow system configurations which `aiodns` / `c-ares` does not support.

https://redhat.atlassian.net/browse/PULP-1788
https://redhat.atlassian.net/browse/SAT-45512

So, we can ditch `aiodns` and also ditch the pinning of `pycares`.

.ci/assets/ci_constraints.txt
```diff
@@ -15,5 +15,3 @@ azure-storage-blob!=12.28.*
 # Apparently does not work with current azurite.
 
 
-pycares<5
-# older aiodns versions don't pin pycares UB, and are broken by pycares>=5
```
mdellweg, line 19: This will come in via the plugin template, but also it is not-a-problem, since constraints will not pull a package, but just restrict it's version range when selected elsewhere.

mdellweg (approved):
Transcending the problem, I love it!

dkliban (approved):

PR #7783, opened 2026-06-08
Fix delete timeouts by batching PublishedArtifact bulk deletes

git-hyagi:
Materialize publication PKs upfront to avoid nested IN subqueries that produce poor execution plans in PostgreSQL. Delete PublishedArtifacts in batches of 500 to limit per-statement WAL size. In RepositoryVersion deletion, pre-delete PublishedArtifacts before the select_for_update block to avoid holding row locks during the slow bulk delete.

### 📜 Checklist

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -544,20 +544,28 @@ class Repository(MasterModel):
         #
         # [0] https://docs.djangoproject.com/en/4.2/ref/models/querysets/#delete
         with transaction.atomic():
-            repo_versions = RepositoryVersion.objects.filter(repository=self)
             repo_contents = RepositoryContent.objects.filter(repository=self)
-            publications = Publication.objects.filter(
-                repository_version__in=repo_versions.values_list("pk", flat=True)
-            )
-            published_artifacts = PublishedArtifact.objects.filter(
-                publication__in=publications.values_list("pk", flat=True)
+
+            # Materialize publication PKs to avoid nested IN subqueries that
+            # produce poor execution plans in PostgreSQL.
+            publication_pks = list(
+                Publication.objects.filter(
+                    repository_version__repository=self,
+                ).values_list("pk", flat=True)
             )
 
             # PublishedArtifact and RepositoryContent are the two most numerous object types
             # PublishedMetadata would be trickier to delete because it's a Content subclass
             # that is ignored by orphan cleanup, so to delete those in this way would require
             # manual intervention
-            published_artifacts._raw_delete(published_artifacts.db)
+            # Delete published artifacts in batches to limit per-statement WAL size.
+            batch_size = 500
+            for start in range(0, len(publication_pks), batch_size):
+                batch = publication_pks[start : start + batch_size]
+                PublishedArtifact.objects.filter(publication_id__in=batch)._raw_delete(
+                    PublishedArtifact.objects.db
+                )
+
             repo_contents._raw_delete(repo_contents.db)
 
             # Anything not deleted manually above will be caught up in Django cascade. Deleting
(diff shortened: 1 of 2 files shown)


pulpcore/app/models/repository.py
```diff
@@ -544,20 +544,25 @@ def delete(self, **kwargs):
...
+            publication_pks = list(
+                Publication.objects.filter(
+                    repository_version__repository=self,
+                ).values_list("pk", flat=True)
             )
 
-            # PublishedArtifact and RepositoryContent are the two most numerous object types
-            # PublishedMetadata would be trickier to delete because it's a Content subclass
-            # that is ignored by orphan cleanup, so to delete those in this way would require
-            # manual intervention
```
dralley, line 559: Don't delete this note (I know Claude did, I just hate when Claude does this)

dralley:
@git-hyagi Do you have any information about the practical impact this has?

git-hyagi:
> @git-hyagi Do you have any information about the practical impact this has?

Hi @dralley,

We ran some tests with an RPM repository defined with autopublish=true and 100k pkgs. Without this change, whenever we tried to delete the repo, it was taking ~5 minutes. After this change, we saw different results. Sometimes we saw 3–4 minutes, but we also saw the same task taking less than 10 seconds.
The environment where we ran the tests is not isolated (it is used by several other users), so it is hard to measure the impact of the other workloads running at the same time we were doing the tests, but we saw some improvements with the timeouts we were seeing: https://redhat.atlassian.net/browse/PULP-1361.
We believe that, even though the delete itself can sometimes take a lot of time to finish, this change can still be helpful because instead of "locking" the entire database until the delete finishes, other tasks seem to be able to run in between the batch executions.

pulpcore/app/models/repository.py
```diff
@@ -544,20 +544,28 @@ def delete(self, **kwargs):
...
+                ).values_list("pk", flat=True)
             )
 
             # PublishedArtifact and RepositoryContent are the two most numerous object types
             # PublishedMetadata would be trickier to delete because it's a Content subclass
             # that is ignored by orphan cleanup, so to delete those in this way would require
             # manual intervention
-            published_artifacts._raw_delete(published_artifacts.db)
+            # Delete published artifacts in batches to limit per-statement WAL size.
+            batch_size = 500
```
dralley, line 562: 500 feels quite small.  Is there a reason to choose such a number over 2000, 2500, 5000?

In comparison to the largest repositories any of those batch sizes would still be small

pulpcore/app/models/repository.py
```diff
@@ -544,20 +544,28 @@ def delete(self, **kwargs):
...
-            )
-            published_artifacts = PublishedArtifact.objects.filter(
-                publication__in=publications.values_list("pk", flat=True)
+
+            # Materialize publication PKs to avoid nested IN subqueries that
+            # produce poor execution plans in PostgreSQL.
+            publication_pks = list(
+                Publication.objects.filter(
+                    repository_version__repository=self,
+                ).values_list("pk", flat=True)
```
dralley, line 554: I'm a bit surprised that passing the sub-query optimizes worse, I would have expected the opposite.

Unless the point of comparison is repeatedly executing the sub-query for each batch.

pulpcore/app/models/repository.py
dralley: Ah, so this is 500 *publications* at a time, not 500 *publishedartifacts* at a time

git-hyagi: > _500 feels quite small. Is there a reason to choose such a number over 2000, 2500, 5000?_

hmm.... good question. We didn't test with different batch sizes to measure the impact on the database or other task executions and timeouts. Let me rerun the tests.

dralley: 500 publications at a time feels much more reasonable to me

With that said, at that point it might be more efficient to go one publication at a time

dralley (dismissed):

dralley:
@git-hyagi Needs a changelog entry, then LGTM

dralley (approved):

PR #7802, opened 2026-06-16
Enable API-versioning and allow for both v3 and v4 versions.

### LINKED ISSUE #6462: API namespacing

We want to support a v4 API in parallel with the existing v3 APIs to enable easier transition to Pulp 4.  

**Additional context**

ggainey:
This is a tech-preview PR, and will only "turn on" if the ENABLE_V4_API is set to True in settings.

It will allow urls for /v3/ and /v4/, and reject other versions.

For urlpatterns registered with a `/<str:version>/` slug, views will be called with a `version=` kwarg.

NOTE: To play this game, plugins will need to insure that all views accept `**kwargs` first, and then update the patterns in their `urls.py` to include the version-slug.

See the `/status/` view and serializer(s) for an example of how to respond based on the incoming request.

NOTE to implementers: **existing tests must pass without changes** whether the ENABLE flag is True or False. If that isn't the case - you're doing something wrong.

closes #6462


pulpcore/app/models/task.py
```diff
@@ -143,6 +144,8 @@ class Task(BaseModel, AutoAddObjPermsMixin):
 
     result = models.JSONField(default=None, null=True, encoder=DjangoJSONEncoder)
 
+    version = models.TextField(default=REST_FRAMEWORK.get("DEFAULT_VERSION", "v3"))
```
mdellweg: Can we call this `api_version`?

pulpcore/app/serializers/task.py
```diff
@@ -118,6 +119,11 @@ class TaskSerializer(ModelSerializer):
         help_text=_("The result of this task."),
     )
 
+    version = serializers.CharField(
+        help_text=_("The API-version that was invoked when creating the task."),
+        default=REST_FRAMEWORK.get("DEFAULT_VERSION", "v3"),
```
mdellweg:
```suggestion
```

I don't think we need this, tasks will never be created via this serializer.

pulpcore/app/settings.py
```diff
@@ -194,6 +195,11 @@
     "DEFAULT_VERSIONING_CLASS": "rest_framework.versioning.URLPathVersioning",
     "DEFAULT_SCHEMA_CLASS": "pulpcore.openapi.PulpAutoSchema",
 }
+ENABLE_V4_API = True
+if ENABLE_V4_API:
+    REST_FRAMEWORK["ALLOWED_VERSIONS"] = ["v3", "v4"]
+    REST_FRAMEWORK["DEFAULT_VERSION"] = "v3"
+
```
mdellweg: This looks like experimentation leftover.

Also i believe if the `REST_FRAMEWORK` adjustments should be dependent on an external `ENABLE_V4_API` setting, it needs to move to a dynaconf hook.

pulpcore/app/models/task.py
ggainey: "version" as an attribute is "magic" at the DRF level - it's what gets filled in by the content of the `<str:version>` slug in the urlpattern. I debated on exactly this - but making it "version" means it's filled directly by the incoming **kwargs to the view, naming it anything else means calling it out specifically. I could be convinced of either.

pulpcore/app/serializers/task.py
ggainey: Possibly so, will dig a bit.

pulpcore/app/settings.py
ggainey: It's not - having these two keys set in REST_FRAMEWORK is what lets DRF "do the right thing" in terms of deciding what versions are allowed/supported.

Could be a hook I guess.

pulpcore/app/models/task.py
mdellweg: I was just abit afraid that we could be confusing here. And changing a database column name later is a pain.

(Pulp is in the domain for versioning on all levels. Pulp-version, pulp-api-version, repository-version, deb-version, rpm-version, ansible-collection-version, ...)

pulpcore/app/settings.py
mdellweg: I think it should be a hook, or the direct succession of assigning `ENABLE_V4_API=True` and the `if ENABLE_V4_API` won't work as expected.

There's another `ENABLE_V4_API=False` in this file.

pulpcore/app/contexts.py
```diff
@@ -4,10 +4,15 @@
 from asgiref.sync import sync_to_async
 from django_guid import clear_guid, get_guid, set_guid
 
+from pulpcore.app.settings import REST_FRAMEWORK
+
 _current_task = ContextVar("current_task", default=None)
 _current_user_func = ContextVar("current_user", default=lambda: None)
 _current_domain = ContextVar("current_domain", default=None)
 x_task_diagnostics_var = ContextVar("x_profile_task")
+_current_pulp_version = ContextVar(
```
ggainey: Comment: Is this really how we want to "know" what pulp-version we are operating under? Author, please review!

(also - even if we keep this, name needs to be _current_pulp_api_version)

ggainey:
RE ENABLE_V4_API : I am going to turn it on *in this PR*, in order to make sure CI runs clean with it enabled. **BEFORE WE MERGE**, we will turn it OFF, and let ppl who want to experiment "opt-in" to it, for several releases.

pulpcore/app/contexts.py
mdellweg: That I kind ow liked. It's non-invasive, yet omnipresent inside the task code.

pulpcore/app/serializers/task.py
dralley: >I don't think we need this, tasks will never be created via this serializer.

You would think, but I remember there being some surprising ways and places in which the serializers were getting initialized.

pulp_file/app/viewsets.py
```diff
@@ -258,7 +258,7 @@ class FileRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin, Role
         responses={202: AsyncOperationResponseSerializer},
     )
     @action(detail=True, methods=["post"], serializer_class=FileRepositorySyncURLSerializer)
-    def sync(self, request, pk):
+    def sync(self, request, pk, **kwargs):
```
dralley, line 261: Apart from `version`, what arguments can end up being passed here?  And what is the likelihood that any would overlap with task arguments?

dralley: Especially provided this applies across basically all tasks, so there are a wide variety of inputs that could potentially overlap.

pulpcore/app/contexts.py
```diff
@@ -64,6 +75,10 @@ def _fetch(task):
     domain, user = await _fetch(task)
     with with_domain(domain), with_guid(task.logging_cid), with_user(user):
         task_token = _current_task.set(task)
+        if not task.version:
+            vers_token = _current_pulp_version.set(REST_FRAMEWORK.get("DEFAULT_VERSION", "v3"))
+        else:
+            vers_token = _current_pulp_version.set(task.version)
```
dralley: So, is this a correct description

"The task needs to know about the API version it was spawned from because the task might use the serializers internally, and the serializers need to know what API version they are operating from"

and

"if the task didn't need to potentially provide that data to Serializers, the context var (within tasks) and the field on the task model would be needless"

pulpcore/app/urls.py
```diff
@@ -152,103 +223,67 @@ class PulpDefaultRouter(routers.DefaultRouter):
     vs_tree.add_decendent(ViewSetNode(viewset))
 
 special_views = [
-    path("login/", LoginViewSet.as_view()),
-    path("repair/", RepairView.as_view()),
-    path("datarepair/7272/", DataRepair7272View.as_view()),
-    path(
-        "orphans/cleanup/",
+    re_path(r"^login/$", LoginViewSet.as_view()),
```
dralley, line 225: What purpose does regex matching serve for these simple paths?

ggainey: "Consistency", mostly. I can do one last experiment where I'm more selective about that.

pulp_file/app/viewsets.py
ggainey: To address this, changed to be specific about version - and also responded to @mdellweg 's "can this be api_version" comment, because that will *also* reduce collision-chance.

pulpcore/app/models/task.py
ggainey: Made the change to api_version - it bent some of the "magic" connections, but I think clarity is more important.

pulpcore/app/models/task.py
```diff
@@ -23,6 +23,7 @@
 from pulpcore.app.models.fields import EncryptedJSONField
 from pulpcore.app.models.status import AppStatus
 from pulpcore.app.role_util import get_users_with_perms
+from pulpcore.app.settings import REST_FRAMEWORK
```
mdellweg: This needs to be `from django.conf import settings`.

mdellweg (dismissed):

ggainey:
> RE ENABLE_V4_API : I am going to turn it on _in this PR_, in order to make sure CI runs clean with it enabled. **BEFORE WE MERGE**, we will turn it OFF, and let ppl who want to experiment "opt-in" to it, for several releases.

ENABLE_V4_API is now `False` by default in this PR (see the warning in the commit-message for why). The PR is now ready to merge.

mdellweg (dismissed):

pulp_file/app/viewsets.py
```diff
@@ -276,16 +276,21 @@ def sync(self, request, pk):
...
             raise ValidationError("Cannot use mirror mode with autopublished repository.")
+
+        task_kwargs = {
+            "remote_pk": str(remote.pk),
+            "repository_pk": str(repository.pk),
+            "mirror": mirror,
+            "optimize": optimize,
+        }
+        if kwargs.get("version"):
+            task_kwargs["api_version"] = kwargs.get("version")
```
dralley: Would it be worth making this even more specific, `pulp_api_version` or similar?  To fully minimize the risk of overlap.

One could potentially imagine passing an `api_version` field into a sync task for example.

pulpcore/app/models/task.py
```diff
@@ -143,6 +144,8 @@ class Task(BaseModel, AutoAddObjPermsMixin):
 
     result = models.JSONField(default=None, null=True, encoder=DjangoJSONEncoder)
 
+    api_version = models.TextField(default=settings.REST_FRAMEWORK.get("DEFAULT_VERSION", "v3"))
```
dralley: Can we not rely on `DEFAULT_VERSION` existing?  I'd prefer that be reliable and we not need a local default fallback here.

pulpcore/app/serializers/task.py
```diff
@@ -118,6 +119,11 @@ class TaskSerializer(ModelSerializer):
         help_text=_("The result of this task."),
     )
 
+    api_version = serializers.CharField(
+        help_text=_("The API-version that was invoked when creating the task."),
+        default=settings.REST_FRAMEWORK.get("DEFAULT_VERSION", "v3"),
```
dralley, line 124: Likewise, having local fallbacks kinda sucks.

dralley: Also is there a good reason to serialize this?  It's useful for debugging I suppose but otherwise I'm not sure it's relevant to the user at all.  We don't serialize task arguments for example.

pulp_file/app/viewsets.py
ggainey: We could - but it feels like overkill? We own this code, and "api version" is WAY more likely to be "Pulp". The edge-case would be adding api-version for...someting else?

pulpcore/app/models/task.py
ggainey: My thinking here (and in the case below) is that I'd rather Absolutely Positively Guarantee we get something here, than have code that wants to *use* api-version have to deal with it maybe being None.

pulpcore/app/serializers/task.py
ggainey: Hm - is there a good reason to *not* do so?

pulpcore/app/views/status.py
```diff
@@ -100,7 +100,18 @@ def get(self, request):
...
+
+        # If V4 is enabled, we'll set up to use the "old" serializer for v3
+        # and the new one with api_version for anything else
+        if settings.ENABLE_V4_API:
+            if request.version == "v3":
+                serializer = StatusSerializer(data, context=context)
+            else:
+                data["api_version"] = request.version
+                data["supported_api_versions"] = settings.REST_FRAMEWORK["ALLOWED_VERSIONS"]
+                serializer = V4StatusSerializer(data, context=context)
```
dralley: Do we need two different versions of `StatusSerializer` or can we just alter the behavior dynamically?

ggainey: There are multiple ways to deal with versioning - this is intended as "an example". Altering "inside of" the view/serializer/model is another one, absolutely.

pulpcore/app/contexts.py
dralley: @ggainey Did you mean to rename this?  I agree that it should be renamed and it looks like you intended to do so

pulpcore/tests/functional/api/test_v4_status.py
```diff
@@ -0,0 +1,28 @@
...
+@pytest.mark.parallel
+@pytest.mark.skipif(
+    not settings.ENABLE_V4_API,
+    reason="Test is pointless if V4 isn't enabled",
+)
+@pytest.mark.parametrize(
+    "version,expect_pass,new_fields",
+    [("v3", True, False), ("v4", True, True), ("v5", False, False)],
+)
+def test_v4_status(version, expect_pass, new_fields, pulp_api_v3_url, pulp_settings):
```
dralley, line 15: Wouldn't it be better to have a `pulp_api_v4_url` fixture?  Or even one which takes an argument.

pulpcore/app/models/task.py
dralley: If it was somehow `None`, then that would be a misconfiguration, and I'd personally rather that failure be more obvious than have the behavior be dependent on a dozen local fallbacks which might accidentally drift apart over time.  It introduces ways for inconsistency to creep in.

pulpcore/app/serializers/task.py
dralley: It just feels like an implementation detail, and I would default to not exposing those.

ggainey: Mmm, I guess I see this differently - knowing how something was created is (or at some point will be, as we modify based-on-version more) important for knowing *what to expect back*. So to me, this is useful information for a user looking at the task.

pulpcore/tests/functional/api/test_v4_status.py
ggainey: Mmm, good point. Lemme experiment :)

pulpcore/app/serializers/task.py
dralley: The API user doesn't learn anything new from it though, because the API user submitted the request in the first place.  So it may get back a task HREF in response to the request that it submitted, but given it submitted the request, it already knows which API version to be using.

Also, both APIs are supposed to work in parallel (there's no situation where an object is tied down to just one version of the API), so there's no reason to choose "what to expect" dynamically.  Consuming code would just pick one API version or the other to use.

Even if you were to dynamically select API version based on the task for some reason, you have to pick one to initially query the tasks endpoint.  Just feels like an unnecessary thing to do.

pulp_file/app/viewsets.py
dralley: Perhaps but I'm in favor of overkill if it's just a matter of naming and has no other downsides.  It also just feels like maintaining consistency with e.g. pulp_domain

pulpcore/tests/functional/api/test_v4_status.py
ggainey: "takes an argument" is more exciting in fixture-land than one might think :)

pulpcore/app/serializers/task.py
ggainey: The API user who created the task may know. The user who is looking at "currently waiting tasks inb the queue", for example, doesn't. The task, at the time it's picked up by the worker, may run differently depending on how it was invoked. I think this is valid information for "a" user to be able to see, and I see no way that it is dangerous/nonperformant.

dralley: I'm not concerned that it's dangerous or nonperformant or anything like that, just that it is unnecessary.  If it were to influence how the task runs (unlikely in most cases), that's still a deep implementation detail, and not even as relevant as many other things that we also don't serialize like `mirror` (which is a sync task request parameter and not presented back to the user in any way AFAICT).

ggainey: I'm not sure how to resolve "you don't think it's useful and I do" :)

dralley: Well, what is a concrete potential use for that info?  I'm fine with exposing it if there's a solid case for it being useful, I just don't want to expose implementation details otherwise.  Especially if it could be added back later should one arise.

pulpcore/tests/functional/api/test_v4_status.py
ggainey: Experimenting with pytest's "indirect" fixture-parameters resulted in code that was, honestly, less clear and more "magic" than this. For now at least, I'm going to leave this as-is.

pulpcore/app/serializers/task.py
ggainey: I can't predict the future. I *can* imagine a task's method being invoked from a worker, looking at the task-context, and deciding "oh, I don't do this The Old Way, I do this the vN way", which results in a different code path/data/whatever. I don't see "what API version was I created with" to be "an implementation detail" - it's specifically public information.

pulpcore/tests/functional/api/test_v4_status.py
dralley: What about `pulp_api_v4_url`?

pulp_file/app/viewsets.py
ggainey: I found an even-better reason to change to pulp_api_version: https://github.com/pulp/pulpcore/blob/main/pulpcore/app/serializers/domain.py#L295

pulpcore/tests/functional/api/test_v4_status.py
ggainey: And pulp_api_v5_url? I don't see that having those as fixtures, makes the code "better". One could even consider the existing "replace v3 in the existing url" to be part of the test for this PR - the URL *should* be the same as for v3, but for the new version.

pulpcore/app/serializers/task.py
dralley: It's already tracked in the task arguments though (specifically `Task.enc_kwargs`).  Strictly speaking it doesn't even need to be a separate field on the `Task` model, much less the serializer.

IMO if we need to debug, we would just do so in the exact same way we would debug anything else.  I don't see a need to special case this.

All sorts of other options can trigger different codepaths within the task - `mirror` already being mentioned, but also the likes of `skip_types`, `optimize`, `x-task-diagnostics` (which is probably the most similar case), etc.  We don't expose those publicly.

pulp_file/app/viewsets.py
dralley: That is indeed a pretty good reason!

(well, it's not a direct conflict, but it's an example of potential overlap issues)

pulpcore/tests/functional/api/test_v4_status.py
dralley: Ok, I think this is fine for now then, at least until we start getting more V4  API tests in general.  This can be a "later" problem.

pulpcore/tasking/tasks.py
```diff
@@ -421,6 +422,7 @@ def get_task_payload(
         "deferred": deferred,
         "profile_options": x_task_diagnostics_var.get(None),
         "app_lock": app_lock,
+        "api_version": kwargs.get("version", default_vers) if kwargs else default_vers,
```
dralley: This code could check the context variable, thus making it automatically self-propagating should a task be spawned from another task.

dralley: I will submit a followup PR with this.

pulpcore/app/serializers/task.py
dralley: I will probably submit a followup PR with this, but won't block this PR as-is.

dralley:
@ggainey @mdellweg What do you think about these tweaks?  https://github.com/ggainey/pulpcore/pull/275

(they can be considered and taken separately if needed)

ggainey:
> @ggainey @mdellweg What do you think about these tweaks? [ggainey#275](https://github.com/ggainey/pulpcore/pull/275)
> 
> (they can be considered and taken separately if needed)

I included your commits #1 and #3 and added you as co-author :)

dralley (approved):

PR #7801, opened 2026-06-16
Add a test for the postgresql parameter limit workaround

dralley:
Assisted By: Claude Code

### 📜 Checklist

- [ ] Commits are cleanly separated with meaningful messages (simple features and bug fixes should be [squashed](https://pulpproject.org/pulpcore/docs/dev/guides/git/#rebasing-and-squashing) to one commit)
- [ ] A [changelog entry](https://pulpproject.org/pulpcore/docs/dev/guides/git/#changelog-update) or entries has been added for any significant changes
- [ ] Follows the [Pulp policy on AI Usage](https://pulpproject.org/help/more/governance/ai_policy/)
- [ ] (For new features) - User documentation and test coverage has been added

See: [Pull Request Walkthrough](https://pulpproject.org/pulpcore/docs/dev/guides/pull-request-walkthrough/)

PR #7803, opened 2026-06-16
Parameter limit

dralley:
### 📜 Checklist

### DIFF
--- a/pulpcore/app/util.py
+++ b/pulpcore/app/util.py
@@ -26,15 +26,13 @@ from pulpcore.app.contexts import _current_domain, _current_user_func, current_p
 from pulpcore.app.loggers import deprecation_logger
 from pulpcore.exceptions.validation import InvalidSignatureError
 
-POSTGRES_MAX_QUERY_PARAMS = 65535
-
 
 class AnyArray(Lookup):
-    """PostgreSQL `= ANY(%s)` lookup that passes a list as a single array parameter.
+    """PostgreSQL ``= ANY(%s)`` lookup that passes a list as a single array parameter.
 
     psycopg3 adapts the Python list into a PostgreSQL array, so the entire list
     counts as **one** bind parameter regardless of size.  This avoids the
-    protocol-level 65535-parameter limit that `IN ($1, $2, …)` hits.
+    protocol-level 65,535-parameter limit that ``IN ($1, $2, …)`` hits.
     """
 
     lookup_name = "any_array"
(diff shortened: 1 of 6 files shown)


dralley:
Follow-up to https://github.com/pulp/pulpcore/pull/6784

This is probably a candidate for using the "run plugin CI against pulpcore main branch" automation.

pulpcore/app/util.py
```diff
@@ -26,6 +26,47 @@
...
+    * If the collection has fewer than 65 535 items, `__in` is used as-is.
+    * Otherwise `__any_array` is used so the whole list travels as a single
+      PostgreSQL array parameter.
+    """
+    if not isinstance(values, (list, set, tuple, frozenset)):
+        return Q(**{f"{field_name}__in": values})
+    values = list(values)
+    if len(values) < POSTGRES_MAX_QUERY_PARAMS:
+        return Q(**{f"{field_name}__in": values})
+    return Q(**{f"{field_name}__any_array": values})
```
mdellweg: Would there be a downside when we _always_ used this array method?

dralley: This was something I was wanting to investigate a bit more before undrafting. This is draft because it's pretty much just "what Claude said" and I wanted to at least look at some query plans and compare before going with it.

I was trying to make a reliable unit test as well but, unfortunately, the unit test in https://github.com/pulp/pulpcore/pull/7801 is not reliable for reasons that are not entirely clear to me.  
> 
> 
> The two SQL strategies are:
> 
> `IN ($1, $2, ..., $N)` — N separate bind parameters, one per value
> `= ANY($1)` — one bind parameter containing a PostgreSQL array
> Parsing/planning: With IN, PostgreSQL must parse N parameter placeholders and the planner builds an OR-tree of comparison nodes. At 100K values, that's significant parse time and planner overhead. With = ANY(array), the query structure is always the same single ScalarArrayOpExpr node regardless of list size.
> 
> Prepared statement caching: IN produces a different query shape for each different N, so the plan can't be reused across different list sizes. = ANY(array) always has the same shape — one parameter — so the plan is reusable.
> 
> Index usage: Both use btree indexes equally well.
> 
> Small lists: For tiny lists (1-10 items), IN is marginally cheaper because there's no array construction. The difference is negligible in practice.
> 
> So why not always use = ANY?
> 
> There's no strong PostgreSQL-level reason not to. The threshold in safe_in() is mostly conservatism:
> 
> __in is Django's standard, battle-tested lookup — it handles querysets (subqueries), empty lists, None values, and all the edge cases Django has polished over years. A custom lookup is more code to maintain.
> __in works across all database backends. = ANY(array) is PostgreSQL-specific.
> For Pulp specifically (always PostgreSQL), you could always use = ANY for Python lists and it'd be fine.
> If you want to simplify, you could drop the threshold and always use any_array for concrete lists. The threshold just avoids the custom path when there's no benefit.

dralley:
@pedro-psb This is my experimentation with the import/export issue, some of it (`safe_in()`) might be useful for the content view issue

PR #7807, opened 2026-06-17
Claude suggestions

### LINKED ISSUE #7808: 2x Improvement requests for the storage migration endpoint
I have been using the storage migrate endpoint to switch from OCI storage to S3.
While it worked okay for small domains, when I hit a domain with more artifacts (160K), some flaws popped up. 
- The migration running artifact by artifact is slow for big domain. Some concurrency might be helpful. 
- The migration is currently "locking" the whole domain for the duration of it. Which for bigger domains becomes service affecting, not being able to perform anything for 24 hours for example. 

I created this draft PR. Feel free to close it if you find the approach bad, it's just something that Claude suggested. https://github.com/pulp/pulpcore/pull/7807

YasenT:
These are some ideas that Claude came with to speed up the migration of big domains
Raising it just as an idea, in relation to an issue which I am also opening.

### DIFF
--- a/pulpcore/app/tasks/migrate.py
+++ b/pulpcore/app/tasks/migrate.py
@@ -1,6 +1,8 @@
 import logging
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from gettext import gettext as _
 
+from django.conf import settings
 from django.utils.timezone import now
 from rest_framework.serializers import ValidationError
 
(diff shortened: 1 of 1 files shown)


pulpcore/app/tasks/migrate.py
```diff
@@ -10,6 +12,35 @@
 
 _logger = logging.getLogger(__name__)
 
+MIGRATION_WORKERS = getattr(settings, "MIGRATION_WORKERS", 8)
```
dralley, line 15: I would call it `BACKEND_MIGRATION_WORKERS` or something, this is quite general and potentially misleading

dralley:
The strategy looks reasonable, fix the lint warnings though

PR #7816, opened 2026-06-23
Revert "Add a test for the postgresql parameter limit workaround"

dralley:
Reverts pulp/pulpcore#7801


dralley:
Even still, it is not a reliable test and takes a long time to run relatively speaking.  It gives false negatives, passes even on a branch that ought to exhibit this issue.

PR #7817, opened 2026-06-25
[PULP-578] Optimize filtering of repo versions by content

### LINKED ISSUE #7799: Use new memoized field in repository content API filtering
The WithContent filter (https://github.com/pulp/pulpcore/blob/main/pulpcore/app/viewsets/custom_filters.py#L375) uses the `with_content` method which still uses the old RepositoryContent model. We need to test to see if this filter has performance issues, and check to see if it can be replaced with the new field.

Follow-up to #7466

jobselko:
closes #7799

### 📜 Checklist

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -893,26 +893,19 @@ class RepositoryVersionQuerySet(models.QuerySet):
         Filters repository versions that contain the provided content units.
 
         Args:
-            content (django.db.models.QuerySet): query of content
+            content (django.db.models.QuerySet or list): Content queryset or list of PKs
 
         Returns:
             django.db.models.QuerySet: Repository versions which contains content.
         """
-        # TODO: Evaluate if this can be optimized with content_ids field
-        query = models.Q(pk__in=[])
-        repo_content = RepositoryContent.objects.filter(content__pk__in=content)
+        if isinstance(content, models.QuerySet):
+            content_pks = content.values_list("pk", flat=True)
+        elif not content:
+            return self.none()
+        else:
+            content_pks = content
 
-        for rc in repo_content.iterator():
-            filter = models.Q(
-                repository__pk=rc.repository.pk,
-                number__gte=rc.version_added.number,
-            )
-            if rc.version_removed:
-                filter &= models.Q(number__lt=rc.version_removed.number)
-
-            query |= filter
-
-        return self.filter(query)
+        return self.filter(content_ids__overlap=content_pks)
 
 
 class RepositoryVersion(BaseModel):
--- a/CHANGES/7799.feature
+++ b/CHANGES/7799.feature
@@ -1,0 +1,1 @@
+Optimized response time when filtering repository versions by content.
--- a/pulpcore/app/viewsets/custom_filters.py
+++ b/pulpcore/app/viewsets/custom_filters.py
@@ -371,7 +371,7 @@ class WithContentFilter(Filter):
         for url in value:
             content_units[extract_pk(url)] = url
 
-        content_units_pks = set(content_units.keys())
+        content_units_pks = list(content_units.keys())
         existing_content_units = Content.objects.filter(pk__in=content_units_pks)
         raise_for_unknown_content_units(existing_content_units, content_units)
 


pulpcore/app/models/repository.py
```diff
@@ -898,21 +898,15 @@ def with_content(self, content):
...
-            filter = models.Q(
-                repository__pk=rc.repository.pk,
-                number__gte=rc.version_added.number,
-            )
-            if rc.version_removed:
-                filter &= models.Q(number__lt=rc.version_removed.number)
+        if isinstance(content, models.QuerySet):
+            content_pks = list(content.values_list("pk", flat=True))
+        else:
+            content_pks = list(content)
```
mdellweg: What is this branch for? (According to the docstring, `content` is always a queryset.)

pulpcore/app/models/repository.py
```diff
@@ -898,21 +898,15 @@ def with_content(self, content):
...
-
-        for rc in repo_content.iterator():
-            filter = models.Q(
-                repository__pk=rc.repository.pk,
-                number__gte=rc.version_added.number,
-            )
-            if rc.version_removed:
-                filter &= models.Q(number__lt=rc.version_removed.number)
+        if isinstance(content, models.QuerySet):
+            content_pks = list(content.values_list("pk", flat=True))
```
mdellweg: Are we really sure we want to evaluate the queryset here (the filtering through `list`).
If we don't, i believe we can let it compose into the filter on line 909 with one less db roundtrip.

mdellweg (commented):
Nice work.

pulpcore/app/models/repository.py
jobselko: Good point, updated

pulpcore/app/models/repository.py
jobselko: `DistributionWithContentFilter.filter` sends a list (`versions = RepositoryVersion.objects.with_content([content.pk]).values_list("pk", flat=True)`), so the docstring was already outdated. Updated now.

jobselko:
Measured benefit by repeatedly adding/removing the same content unit into a repo (creating 2 RepositoryVersion and 1 RepositoryContent per cycle):

**Without patch:**
Response time grows with the number of cycles (e.g. ~0.5s at 100, ~2-3s at 700)

**With patch:**
Response time stays constant at ~0.45s regardless of the number of cycles

pulpcore/app/models/repository.py
mdellweg: `values_list` should still return a (lazily evaluated) queryset:

https://docs.djangoproject.com/en/6.0/ref/models/querysets/#values-list

jobselko: `values_list` is not what is passed in. `[content.pk]` is the argument to `with_content()`, and this is a plain list, not a queryset

jobselko: So I changed `list(content)` to just `content` as `__overlap` accepts any iterable.

EDIT: `overlap` only works on lists, `list()` is still needed

mdellweg: I may be missing something. But the queryset from line 902 works with overlap perfectly, right?

jobselko: Yeah, `overlap` is happy with both querysets and lists (so not only with lists, I was wrong).

dralley: So does that mean the branch is unnecessary?

mdellweg: Well, my thinking is that composing querysets (being lazy) can reduce the back and forth between app and db. Calling "list" on a queryset evaluates it forcing the id's to be loaded into django.

(Whether leaving a much more complex query for postgres to digest is actually an optimization, i can only assume...)

pulpcore/app/models/repository.py
```diff
@@ -893,26 +893,20 @@ def with_content(self, content):
...
-            if rc.version_removed:
-                filter &= models.Q(number__lt=rc.version_removed.number)
+        if isinstance(content, models.QuerySet):
+            content_pks = content.values_list("pk", flat=True)
+        else:
+            content_pks = content
 
-            query |= filter
+        if not content_pks:
+            return self.none()
```
mdellweg: Now i wonder if these to lines give us anything.
In case of a queryset, the truthy test will probably call a premature evaluation defeating the benefit from composing SQL.
Maybe in the case of the parameter being an empty list of id's, this can spare us from makeing an SQL call altogether. But it should be part of the "else" branch then.

mdellweg: All this still depends on what kind of argument we actually allow. I see you changed the docstring...

jobselko: Updated. We allow lists of PKs (this behavior existed before any of my changes) or querysets.

mdellweg (approved):

dralley (approved):

PR #7821, opened 2026-06-29
Fix race condition inside SingleArtifactContentUploadSerializer

gerrod3:
The serializer approach runs into the issue of the UniqueValidator sometimes throwing an exception if we try to save a new artifact, but it was created concurrently inbetween our first retrieval and the new save. Not sure why we were using a serializer in the first place (maybe I'm missing something?), but just copy-paste the database pattern we use elsewhere should handle it.

### 📜 Checklist


mdellweg (approved):
I think we just never ran into that. The window really is rather small.

PR #7832, opened 2026-06-30
Expose opt-in content_ids on repository version detail

### LINKED ISSUE #7831: Need a performant solution to find content diffs between non-adjacent repo versions

acheng-01:
Add a content_ids SerializerMethodField to RepositoryVersionSerializer that returns the version's content unit UUIDs only when the request includes the content_ids=true query parameter (null otherwise). This lets clients diff content sets between versions directly instead of issuing expensive content filter queries.

Closes #7831

Assisted by: Claude Opus 4.8

### 📜 Checklist


acheng-01:
Closed as this is deemed not ideal. Replacing this PR with #7844 as a more viable alternative.

PR #7844, opened 2026-07-02
Add added_between/removed_between filters for content diffs across arbitrary repository versions

### LINKED ISSUE #7831: Need a performant solution to find content diffs between non-adjacent repo versions

acheng-01:
This PR adds two range filters to the content viewset, added_between and removed_between, that
compute the net content diff between two arbitrary (non-adjacent) repository versions, rather
than only the single-step diff against the immediate predecessor that repository_version_added
and repository_version_removed provide.

Each filter takes exactly two comma-separated Repository Version (or Repository) HREF/PRNs in
'base,target' order. added_between=base,target returns the content present in target but not in
base; removed_between=base,target returns the content present in base but not in target. The two
are symmetric: added_between=a,b is equivalent to removed_between=b,a.

Benchmarks on a synthetic 200k-unit system with 1000-unit diffs between repo versions show this
approach is ~5-15x faster than running an equivalent complex q query, with the query going from
1486 ms -> 99 ms for a 50k/50k repo version pair and 2403 ms -> 479 ms for a 199k/199k repo
version pair. Planning time also decreases significantly and stays ~0.2 ms.

repository_version_added and repository_version_removed are unchanged and continue to compute the
single-step diff against the immediate predecessor, preserving backward compatibility.

Closes https://github.com/pulp/pulpcore/issues/7831

Assisted by: Claude Opus 4.8

### 📜 Checklist

### DIFF
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -997,15 +997,24 @@ class RepositoryVersion(BaseModel):
         if content_qs is None:
             content_qs = Content.objects
 
-        content_ids = self.content_ids
-        if len(content_ids) >= 65535:
-            # Workaround for PostgreSQL's limit on the number of parameters in a query
-            content_ids = (
-                RepositoryVersion.objects.filter(pk=self.pk)
-                .annotate(cids=Func(F("content_ids"), function="unnest"))
-                .values_list("cids", flat=True)
-            )
-        return content_qs.filter(pk__in=content_ids)
+        return content_qs.filter(pk__in=self.content_ids_subquery())
+
+    def content_ids_subquery(self):
+        """
+        Return this version's ``content_ids`` as a database-side ``unnest`` subquery.
+
+        Using a subquery keeps the content unit UUIDs inside PostgreSQL instead of loading the
+        whole array into Python and passing each UUID as a bound query parameter. This avoids the
+        per-query parameter limit and the memory/serialization cost for large repository versions.
+
+        Returns:
+            django.db.models.QuerySet: A values queryset yielding the content unit UUIDs.
+        """
+        return (
+            RepositoryVersion.objects.filter(pk=self.pk)
+            .annotate(cids=Func(F("content_ids"), function="unnest"))
+            .values_list("cids", flat=True)
+        )
 
     @property
     def content(self):
@@ -1119,8 +1128,8 @@ class RepositoryVersion(BaseModel):
         if not base_version:
             return Content.objects.filter(version_memberships__version_added=self)
 
-        return Content.objects.filter(pk__in=self.content_ids).exclude(
-            pk__in=base_version.content_ids
+        return Content.objects.filter(pk__in=self.content_ids_subquery()).exclude(
+            pk__in=base_version.content_ids_subquery()
         )
 
     def removed(self, base_version=None):
@@ -1134,8 +1143,8 @@ class RepositoryVersion(BaseModel):
         if not base_version:
             return Content.objects.filter(version_memberships__version_removed=self)
 
-        return Content.objects.filter(pk__in=base_version.content_ids).exclude(
-            pk__in=self.content_ids
+        return Content.objects.filter(pk__in=base_version.content_ids_subquery()).exclude(
+            pk__in=self.content_ids_subquery()
         )
 
     def contains(self, content):
--- a/pulpcore/app/models/repository.py
+++ b/pulpcore/app/models/repository.py
@@ -1028,13 +1028,24 @@ class RepositoryVersion(BaseModel):
         if content_qs is None:
             content_qs = Content.objects
 
-        # Try to not even attempt to evaluate the content_ids on the python side.
-        content_ids_subquery = (
+        return content_qs.filter(pk__in=self.content_ids_subquery())
+
+    def content_ids_subquery(self):
+        """
+        Return this version's ``content_ids`` as a database-side ``unnest`` subquery.
+
+        Using a subquery keeps the content unit UUIDs inside PostgreSQL instead of loading the
+        whole array into Python and passing each UUID as a bound query parameter. This avoids the
+        per-query parameter limit and the memory/serialization cost for large repository versions.
+
+        Returns:
+            django.db.models.QuerySet: A values queryset yielding the content unit UUIDs.
+        """
+        return (
             RepositoryVersion.objects.filter(pk=self.pk)
             .annotate(cids=Func(F("content_ids"), function="unnest"))
             .values_list("cids", flat=True)
         )
-        return content_qs.filter(pk__in=content_ids_subquery)
 
     @property
     def content(self):
@@ -1148,7 +1159,9 @@ class RepositoryVersion(BaseModel):
         if not base_version:
             return Content.objects.filter(version_memberships__version_added=self)
 
-        return Content.objects.filter(pk__in=self.content).exclude(pk__in=base_version.content)
+        return Content.objects.filter(pk__in=self.content_ids_subquery()).exclude(
+            pk__in=base_version.content_ids_subquery()
+        )
 
     def removed(self, base_version=None):
         """
@@ -1161,7 +1174,9 @@ class RepositoryVersion(BaseModel):
         if not base_version:
             return Content.objects.filter(version_memberships__version_removed=self)
 
-        return Content.objects.filter(pk__in=base_version.content).exclude(pk__in=self.content)
+        return Content.objects.filter(pk__in=base_version.content_ids_subquery()).exclude(
+            pk__in=self.content_ids_subquery()
+        )
 
     def contains(self, content):
         """
--- a/pulpcore/app/viewsets/base.py
+++ b/pulpcore/app/viewsets/base.py
@@ -157,7 +157,7 @@ class NamedModelViewSet(viewsets.GenericViewSet):
         return match.func.cls.queryset.model
 
     @staticmethod
-    def get_resource(uri, model=None):
+    def get_resource(uri, model=None, deferred_fields=None):
         """
         Resolve a resource URI/PRN to an instance of the resource.
 
@@ -168,6 +168,9 @@ class NamedModelViewSet(viewsets.GenericViewSet):
             uri (str): A resource URI/PRN.
             model (django.models.Model): A model class. If not provided, the method automatically
                 determines the used model from the resource URI/PRN.
+            deferred_fields (iterable): Optional field names to defer when loading the resource, so
+                large columns are not fetched from the database. Field names that do not exist on
+                the resolved model are ignored.
 
         Returns:
             django.models.Model: The resource fetched from the DB.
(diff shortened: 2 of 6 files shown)


pulpcore/app/models/repository.py
```diff
@@ -1119,8 +1132,8 @@ def added(self, base_version=None):
         if not base_version:
             return Content.objects.filter(version_memberships__version_added=self)
 
-        return Content.objects.filter(pk__in=self.content_ids).exclude(
-            pk__in=base_version.content_ids
+        return Content.objects.filter(pk__in=self.content_ids_subquery()).exclude(
+            pk__in=base_version.content_ids_subquery()
```
mdellweg, line 1132: This sounds like the real improvement here.
Can we even mark `content_ids` as `deferred` so that the humongous list never makes it to python in the first place? (Maybe that's already the case...)

mdellweg: Also does this change already improve the q filter scenario?

dkliban:
We have fixed a couple of DB performance issues by switching to using subqueries. I foresee us doing this more broadly throughout the codebase.

pulpcore/app/models/repository.py
daviddavis: I'll try to sum what we talked about at our pulpcore meeting but generally we agree on investigating these two items here. For the latter, we think that the `q` filter could/should be using subqueries. If it's not, it's worth understanding why it's not. Fixing that should alleviate the performance problem.

acheng-01: I'm not sure if this will correctly answer the question of whether the subquery I used here improves `q` performance, but:
- The filter we pass through `q` originally does eventually reach the `ContentRepositoryVersionFilter` logic and calls `get_content()`. However, there's a condition there where a subquery strategy is only used if a version has more than 65535 units. Below that, it's still inlining the content id list.
- `q` also has a different query shape compared to the implementation here that directly targets the repository version table. The `NOT repository_version=y` part becomes `Content.objects.all() EXCEPT (content in Y)`, which means it's still running through the entire table
- The `base_version` implementation here works around the expression filter strategy that is `q` (which I presume has other uses) and goes straight to `filter(pk__in=X).exclude(pk__in=Y)`. The subquerying optimizes this filter path to be sure, but I'm hesitant in changing things related to `q` because I don't fully understand its other uses.

acheng-01:
Here are some added insights to some of the performance concerns raised. I'm mainly trying to answer these two questions:
1. How the query shape differs when using `q` with subqueries vs `base_version`
2. What is the performance like for `q` with subqueries vs `base_version`

Tl;dr - the `q=` filter _can_ be optimized to the same performance as a dedicated filter as two end up as the same query, but it has a much larger blast radius and could have regressive consequences if we're not careful.

To review, `q=` is slow today (in our use case but maybe others too) because for "content in version X but not in base Y", `repository_version=X AND NOT repository_version=Y` compiles through the expression engine's set operations and becomes:

(X's content) `INTERSECT` ( (ALL content) `EXCEPT` (Y's content) )

`_NotAction` → `qs.difference()` (`EXCEPT`) and `_AndAction` → `qs.intersection()` (`INTERSECT`). The `NOT` term's left operand is the whole table, so the plan is a `HashSetOp Intersect` over a `HashSetOp Except` that seq-scans the entire content table and sorts/hashes every leg, spilling to disk (`temp read/written`, external merge sorts). That full-table set-op work — not the `pk__in` sub-selects — is the cost.

If the existing `q` expression filter engine composed operands as `pk__in` subqueries instead of set-ops — i.e. `_NotAction` → `exclude(pk__in=…)`, `_AndAction` → `filter(pk__in=…)` — the plan collapses to a nested loop over an indexed anti-join (`unnest(content_ids)` → pkey `Index Scan` + `NOT (hashed SubPlan)`). No full-table seq scan, no `SetOp` , no disk spill. This is the same plan a dedicated `base_version` filter (`filter(pk__in=X).exclude(pk__in=Y)`) produces.

And here are some benchmarks comparing all three scenarios. These are measured on a 200k-unit dataset with 1000-unit diff between repo versions. Results are taken over a 3-run average.

| Approach | 199k-unit versions | 50k-unit versions |
| ----------- | -------------------- | ------------------- |
| `q=` today with set-ops | ~1400 ms | ~450 ms |
| `q=` with `pk__in` subqueries | ~680 ms | ~150 ms |
| `base_version` filter | ~600 ms | ~140 ms |

As the results show, optimizing `q=` gets ~2× faster and lands within noise of a dedicated filter. The remaining time is inherent to the diff (unnesting/deduping ~199k ids and probing the index) rather than anything the query shape can avoid.

The worrying part is where we would implement the fix for `q` vs adding in `base_version` for our specific use case. The three `evaluate` methods (`NotAction` / `_AndAction` / `_OrAction`) are the generic `q=` engine used by every plugin and content type. Some specific concerns:

• The `pk__in` rewrite only holds if every operand is a pk-semijoinable queryset on the same model. If any sub-filter adds annotations/joins or filters a different model, `.values("pk")` wrapping is needed and must be validated — `.difference()` / `.intersection()` were chosen likely because they compose arbitrary querysets generically.
• Nested `pk__in (… values("pk"))` subqueries usually plan as semi/anti-joins, but it's unclear what would happen with deeply nested `AND` / `OR` / `NOT` queries. Pulp's docs say the limit is 8, which would mean a lot of nesting.
• `union()/intersection()/difference()` return querysets with restricted downstream operations; the `filter` / `exclude` form actually lifts those restrictions, which is a behavior change (generally an improvement) to test.

As an aside, I've also explored removing the 65535 content unit conditional on `get_content()` before subquerying happens for the repository version table as a smaller change to all of `q`'s operands. It will improve planning performance but won't affect execution performance because we'll still have the `EXCEPT` and `INTERSECT` operations that take up a bulk of the resources.

acheng-01:
I've opened a [draft PR](https://github.com/pulp/pulpcore/pull/7848) on what subqueries for `q` would look like. The existing filter tests appear to be passing with the changes. Let me know which direction would be better.

pulpcore/app/models/repository.py
mdellweg: > * However, there's a condition there where a subquery strategy is only used if a version has more than 65535 units.

I always wondered why we should have this in the first place. I cannot think of the subquery on an small or empty table being an issue at all.

acheng-01: In the latest commit, I've implemented changes to defer `content_ids` and also dropped the conditional for using the subquery in `get_content()`. Tests are passing. Lmk what you think.

mdellweg:
copied from #7848; discussion should really happen here.

> > Given the speedups we see here, I'd say we most definitely want this. Looks like a great piece of work. (I'm almost inclined to keep this in limbo for another week to unlock the next level of significant speedup of the parsing. Not actually!) Too bad, it doesn't sweepingly address your initial issue. Do you believe we could add a dedicated filter receiving two versions returning the set-difference (maybe that would "just" boil down to merging the base_version into the original 'version_added' filter)?
> > Drafting from the top of my head (and using a terrible name): `added_between=/p/a/v3/r/v/3/,/p/a/v3/r/v/6/`
> > Maybe this is helping (probably only as a blueprint, not as an implementation): https://django-filter.readthedocs.io/en/stable/ref/filters.html#django_filters.filters.BaseRangeFilter
> 
> Thanks! I'm glad that this turned out well.
> 
> Regarding our initial issue, as you mentioned I _think_ what you're proposing is actually what my `base_filter` [PR](#7844) already does. If we were to go the `added_between` / `removed_between` route, it'd really be an alternative packaging of the same logic because both ultimately call `RepositoryVersion.added()/removed()` with a base version.
> 
> I did a bit of research on `BaseRangeFilter`, and I believe its contribution here is thin: • It just splits `"a,b"` into two values and validates the count; its default `range` lookup (SQL `BETWEEN`) doesn't apply, so we'd override `.filter()` anyway. • We'd still resolve each href/PRN via the existing `get_repository_version()` (called twice) and need two new filters.
> 
>     * Its only positive is that it binds the two versions together, so the current "base alone is a no-op" scenario becomes impossible.
> 
> 
> Since #7844 is more or less ready and tested, I would personally lean toward keeping `base_version` and, if the silent no-op is the concern, I could add a validation error when `base_version` is passed without `repository_version_added / removed`. If you prefer the single-param ergonomics then I can add `added_between / removed_between` to it. Lmk!

Exactly. That's what I meant with blueprint. The implementation of the RangeFilter was never going to help us.
The "single-param egonomics" (nicely phrased!) is my personal preference for the api design (after all, to go full circle, it would allow sophisticated use of the filter in the "q-Language"). But I really want others to weigh in.

acheng-01:
@gerrod3 thoughts on the comment @mdellweg copied over?

gerrod3:
My preference is always for designs that are more explicit in their behavior and explanation. So in this case I think `added_between` and `removed_between` are the more intuitive filters.

acheng-01:
> My preference is always for designs that are more explicit in their behavior and explanation. So in this case I think `added_between` and `removed_between` are the more intuitive filters.

Okay sure. Will work on using `added_between` and `removed_between` for the next commit.

pedro-psb:
@acheng-01 lint is failing

ggainey:
@acheng-01 can you fix the lint issues and repush please?

pulpcore/app/viewsets/base.py
```diff
@@ -157,7 +157,7 @@ def get_resource_model(uri):
         return match.func.cls.queryset.model
 
     @staticmethod
-    def get_resource(uri, model=None):
+    def get_resource(uri, model=None, deferred_fields=None):
```
mdellweg, line 160: I'd like to handle the deferring differently (on the model/manager level not the view):

https://github.com/pulp/pulpcore/pull/7867

acheng-01: @mdellweg I left a comment on your PR. I'm fine with your changes, but how would you like to proceed so that we don't conflict?

mdellweg: To make matters worse, there's a third attempt in the same arena.

https://github.com/pulp/pulpcore/pull/7803

acheng-01: That one has been in draft mode for a few months at this point though. I'm leaning toward either you or I merge first and have the other rebase and resolve conflicts. I think #7803 can rebase again off our changes if needed, but happy to discuss further if you disagree.

mdellweg: Finally we got moving. Can you maybe now rebase?

acheng-01: @mdellweg I've rebased and removed my implementation in the viewset. Can you review again?

pulpcore/app/viewsets/custom_filters.py
```diff
@@ -163,6 +163,90 @@ def filter(self, qs, value):
         raise NotImplementedError()
 
 
+class RepositoryVersionBetweenFilter(BaseRangeFilter, RepoVersionHrefPrnFilter):
+    """
+    Base class for range filters that diff content between two arbitrary repository versions.
+
+    The filter accepts exactly two comma-separated Repository Version (or Repository) HREF/PRN
+    values in ``base,target`` order: the first is the base version and the second is the target
+    version. Subclasses implement :meth:`diff` to return the appropriate net set of content.
```
mdellweg, line 172: Does this mean `added_between=1,2` will produce the same results as `removed_between=2,1`?

Not a blocker, just an observation.

mdellweg (approved):
I think this is ready to go.

Thank you for sticking with us for so long.

pulpcore/app/viewsets/custom_filters.py
acheng-01: Yes that is correct. They're technically the same depending on how you use them but the parameter naming hopefully makes things a little clearer.

PR #7848, opened 2026-07-07
Optimize q filter to use pk__in subqueries instead of set operations

### LINKED ISSUE #7831: Need a performant solution to find content diffs between non-adjacent repo versions

acheng-01:
The `q` complex filter composed its NOT/AND/OR operands with QuerySet .difference()/.intersection()/.union(), which compile to SQL EXCEPT/INTERSECT/UNION. For NOT in particular, the left operand is the full unfiltered queryset, so evaluating an expression like `repository_version=X AND NOT repository_version=Y` seq-scans the entire content table and sorts/hashes every leg, spilling to disk on large repositories.

Rewrite the three operand evaluators to compose with `pk__in` subqueries:

  - _NotAction:  qs.exclude(pk__in=expr.evaluate(qs).values("pk"))
  - _AndAction:  chained qs.filter(pk__in=...) semi-joins
  - _OrAction:   OR of Q(pk__in=...) subqueries

This collapses the plan to indexed anti-/semi-joins with no full-table scan, no SetOp, and no disk-spilling sorts. Results are unchanged (each operand still filters the same base queryset on a unique pk, so the implicit de-duplication of the set operations is preserved); the existing q filter functional tests pass without modification.

Benchmarked on a 200k-unit repository computing a 1000-unit diff between two repository versions (steady-state, work_mem=4MB):

| Version sizes | Before (set-ops) | After (pk__in) |
| -------------- |  ----------------- |  -------------- |
| 199k units | ~1400 ms | ~680 ms |
| 50k units | ~450 ms | ~150 ms |

Planning time also drops (e.g. ~1.4 ms -> ~0.5 ms) as the large inlined operand lists are replaced by parameterized subqueries.

Closes #7831

Assisted by: Claude Opus 4.8

### 📜 Checklist

### DIFF
--- a/CHANGES/7831.bugfix
+++ b/CHANGES/7831.bugfix
@@ -1,0 +1,3 @@
+Optimized the ``q`` complex filter to compose ``NOT``/``AND``/``OR`` operands using ``pk__in``
+subqueries instead of the ``EXCEPT``/``INTERSECT``/``UNION`` set operations, avoiding full-table
+scans and disk-spilling sorts on large content sets.
--- a/CHANGES/7831.bugfix
+++ b/CHANGES/7831.bugfix
@@ -1,3 +1,5 @@
-Optimized the ``q`` complex filter to compose ``NOT``/``AND``/``OR`` operands using ``pk__in``
-subqueries instead of the ``EXCEPT``/``INTERSECT``/``UNION`` set operations, avoiding full-table
-scans and disk-spilling sorts on large content sets.
+Optimized the `q` complex filter to compose `NOT`/`AND`/`OR` operands as flat `WHERE`
+conditions in a single SQL statement for standard field filters, falling back to `pk__in`
+subqueries only for opaque operands (label, `repository_version`, and href/prn/id resolvers).
+This replaces the previous `EXCEPT`/`INTERSECT`/`UNION` set operations, avoiding full-table
+scans and disk-spilling sorts, and keeps the query flat regardless of expression nesting depth.
--- a/pulpcore/filters.py
+++ b/pulpcore/filters.py
@@ -10,7 +10,7 @@ from django.core.exceptions import FieldDoesNotExist
 from django.db import models
 from django.forms.utils import ErrorList
 from django.urls import Resolver404, resolve
-from django_filters.constants import EMPTY_VALUES
+from django_filters.constants import EMPTY_VALUES as STANDARD_EMPTY_VALUES
 from django_filters.rest_framework import BaseInFilter, DjangoFilterBackend, filters, filterset
 from drf_spectacular.contrib.django_filters import DjangoFilterExtension
 from drf_spectacular.plumbing import build_basic_type
@@ -20,7 +20,7 @@ from rest_framework import serializers
 
 from pulpcore.app.util import extract_pk, get_domain_pk, resolve_prn
 
-EMPTY_VALUES = (*EMPTY_VALUES, "null")
+EMPTY_VALUES = (*STANDARD_EMPTY_VALUES, "null")
 UPMatch = namedtuple("UPMatch", ("model", "pk"))
 
 
@@ -185,6 +185,23 @@ class PulpTypeInFilter(BaseInFilter, PulpTypeFilter):
 
 
 class ExpressionFilterField(forms.CharField):
+    """Parse and compile a ``q=`` complex-filter expression into a queryset transform.
+
+    Each parsed node (leaf ``_FilterAction`` and the ``NOT``/``AND``/``OR`` combinators) implements
+    a two-method protocol:
+
+    * ``as_flat_q()`` -> ``Q | None``: a pure, side-effect-free attempt to represent the *whole*
+      subtree as a single flat ``Q`` object. It returns ``None`` if any operand is "opaque" (a
+      filter with a custom ``.filter()`` override, a ``method``, or ``.distinct()``) and therefore
+      cannot be expressed as a plain ``WHERE`` condition.
+    * ``evaluate(qs)`` -> ``QuerySet``: applies the subtree to ``qs``. It uses ``as_flat_q()`` when
+      the subtree is fully reducible (one flat ``WHERE``) and otherwise falls back to composing the
+      opaque operands as ``pk__in`` semi-/anti-join subqueries.
+
+    ``evaluate()`` is the sole external entry point: :meth:`ExpressionFilter.filter` calls
+    ``value.evaluate(qs)`` on the outermost node.
+    """
+
     class _FilterAction:
         def __init__(self, filterset, tokens):
             key = tokens[0].key
@@ -206,11 +223,14 @@ class ExpressionFilterField(forms.CharField):
             self.value = form.cleaned_data[key]
             self.complexity = 1
 
-        def as_q(self):
-            # A leaf reduces to a plain WHERE condition (Q) only when it is a standard
-            # django-filter Filter: no custom .filter() override, no method, no .distinct().
-            # Anything else (href/prn resolvers, repository_version, method filters, ...) is
-            # opaque and must be applied via its own queryset transform in evaluate().
+        def as_flat_q(self):
+            """Return this leaf as a flat ``Q``, or ``None`` if the filter is opaque.
+
+            A leaf reduces to a plain ``WHERE`` condition only when it is a standard django-filter
+            ``Filter``: no custom ``.filter()`` override, no ``method``, and no ``.distinct()``.
+            Anything else (href/prn resolvers, ``repository_version``, method filters, ...) is
+            opaque and must be applied via its own queryset transform in :meth:`evaluate`.
+            """
             f = self.filter
             if (
                 type(f).filter is not filters.Filter.filter
@@ -218,13 +238,18 @@ class ExpressionFilterField(forms.CharField):
                 or getattr(f, "distinct", False)
             ):
                 return None
-            if self.value in EMPTY_VALUES:
+            # Mirror filters.Filter.filter exactly: an empty/unspecified value means the filter is
+            # not applied (identity Q), not "IS NULL". Use django-filter's own EMPTY_VALUES rather
+            # than the module-augmented set so the sentinel "null" is not swallowed here; null-aware
+            # filters override .filter() and are handled on the opaque evaluate() path above.
+            if self.value in STANDARD_EMPTY_VALUES:
                 return models.Q()
             q = models.Q(**{f"{f.field_name}__{f.lookup_expr}": self.value})
             return ~q if f.exclude else q
 
         def evaluate(self, qs):
-            q = self.as_q()
+            """Apply the leaf: a flat ``Q`` when reducible, else the filter's own transform."""
+            q = self.as_flat_q()
             if q is not None:
                 return qs.filter(q)
             return self.filter.filter(qs, self.value)
@@ -234,12 +259,14 @@ class ExpressionFilterField(forms.CharField):
             self.expr = tokens[0][0]
             self.complexity = self.expr.complexity + 1
 
-        def as_q(self):
-            child = self.expr.as_q()
+        def as_flat_q(self):
+            """Negate the child's flat ``Q``; ``None`` if the child is opaque."""
+            child = self.expr.as_flat_q()
             return None if child is None else ~child
 
         def evaluate(self, qs):
-            q = self.as_q()
+            """Apply ``NOT``: negate a flat ``Q`` when reducible, else anti-join the operand."""
+            q = self.as_flat_q()
             if q is not None:
                 return qs.filter(q)
             # Opaque operand: anti-join via a NOT IN subquery.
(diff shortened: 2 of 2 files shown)


pulpcore/filters.py
```diff
@@ -215,27 +215,29 @@ def __init__(self, tokens):
...
-            )
+            result = qs
+            for expr in self.exprs:
+                result = result.filter(pk__in=expr.evaluate(qs).values("pk"))
+            return result
 
     class _OrAction:
         def __init__(self, tokens):
             self.exprs = tokens[0]
             self.complexity = sum((expr.complexity for expr in self.exprs)) + 1
```
mdellweg, line 311: FYI, the complexity is meant to prevent easy dos attacks by generating arbitrarily complex queries. I guess you can already get quite far with the current setting.

CHANGES/7831.bugfix
```diff
@@ -0,0 +1,3 @@
+Optimized the ``q`` complex filter to compose ``NOT``/``AND``/``OR`` operands using ``pk__in``
+subqueries instead of the ``EXCEPT``/``INTERSECT``/``UNION`` set operations, avoiding full-table
+scans and disk-spilling sorts on large content sets.
```
mdellweg: This is supposed to be markdown, not ReST.
Single backticks work better, as far as I recall.

mdellweg (commented):
I would be willing to go this way.
It sounds like it can be a performance boost for all uses of q.

Do I understand correctly that your concern is manly how the performance is impacted by deeply nesting constructs?

acheng-01:
> I would be willing to go this way. It sounds like it can be a performance boost for all uses of q.
> 
> Do I understand correctly that your concern is manly how the performance is impacted by deeply nesting constructs?

Kind of. The benchmark copilot ran yesterday were done from more theoretical flattened SQL queries that `q=` can not actually achieve and thus produced misleading improvement figures. Because each sub-filter now returns `qs.filter(...)` and the operators compose by wrapping, every operand expands into several stacked `core_content` self-joins. For example, `repository_version=X` turns into:
```SQL
content WHERE pk IN (SELECT pk FROM content WHERE pk IN (
    SELECT pk FROM content WHERE pk IN (SELECT unnest(content_ids) FROM version)))
```
Switching from set-ops to `pk__in` subqueries removes the `EXCEPT` / `INTERSECT` disk-spill sorts, which is an improvement for all of `q` usage, but it trades in one more wrapping layer, so the net gain through the actual filter is more modest than a flattened query (going the `base_version` filter route) would suggest (~15% rather than ~100%). I just want to be upfront that it optimizes the shape rather than eliminating the underlying per-operand nesting.

acheng-01:
Following up on my own comment — I pushed a commit that I think actually addresses the performance issue.

• **Before:** each operand of an `AND / OR / NOT` got its own `pk__in` subquery, so nesting just stacked up more layers of SQL.
• **Now:** each operand can report an `as_q()` — if it's a plain field filter (name, dates, numbers, `__in`, etc.) it returns a `Q` object, and the operators just combine those with `& / | / ~`. A whole expression of standard fields therefore folds into a single flat SQL `WHERE`, no matter how deep it's nested.
• Operands that genuinely can't be reduced — labels, `repository_version`, href/prn/id lookups — return no `Q` and fall back to a `pk__in` subquery like before.
• Since nearly every filter on every endpoint is a plain field, this helps the common case across the board.
• `repository_version` on its own is unchanged (~2.4s on 200k units) since it stays a subquery, it just doesn't get the flattening benefit.

The existing `q` filter tests all pass without changes.

acheng-01:
Some more benchmarks run on actual differences in code:

100k rows (200k for `repository_version`), best of 4.
The number next to the dot is the number of `SELECT`s made by the query.

**ContentGuard (name)**

| Expression | set-ops | subquery | hybrid |
|---|---|---|---|
| `A AND B` | 68 ms · 2 | 41 ms · 3 | 14 ms · 1 |
| `A OR B` (tiny result) | 11 ms · 2 | 23 ms · 3 | 13 ms · 1 |
| `(A AND B) OR C` | 128 ms · 3 | 101 ms · 5 | 17 ms · 1 |
| `NOT A` | 124 ms · 2 | 31 ms · 2 | 12 ms · 1 |

**Repository (name / number)**

| Expression | set-ops | subquery | hybrid |
|---|---|---|---|
| `name AND name` | 164 ms · 2 | 43 ms · 3 | 15 ms · 1 |
| `(A AND B) OR (C AND D)` | 170 ms · 4 | 98 ms · 7 | 20 ms · 1 |
| `(A AND (B OR C)) AND NOT D` (complexity 8) | 559 ms · 5 | 166 ms · 8 | 22 ms · 1 |
| `name AND retain>=x AND retain<y` | 251 ms · 3 | 52 ms · 4 | 15 ms · 1 |
| `(num range) OR num=0` | 263 ms · 3 | 117 ms · 5 | 13 ms · 1 |

**Task**

| Expression | set-ops | subquery | hybrid |
|---|---|---|---|
| `name AND NOT name` | 303 ms · 3 | 69 ms · 4 | 17 ms · 1 |
| `name AND (NOT state OR state)` | 478 ms · 4 | 105 ms · 6 | 97 ms · 5 |

**Content — `repository_version` (opaque, 200k units)**

| Expression | set-ops | subquery | hybrid |
|---|---|---|---|
| 199k `AND NOT` 199k | 2662 ms · 7 | 2307 ms · 8 | 2234 ms · 8 |
| 50k `AND NOT` 50k | 1900 ms · 5 | 1458 ms · 6 | 1558 ms · 6 |

As the results show, the changes made in the new commit will in fact improve `q=` performance overall for _all_ queries. However, for operands that can't be reduced/flattened, performance doesn't have as dramatic of an improvement. So in our case with repository_version = X AND NOT repository_version = Y, we would get marginal benefits from this vs the other `base_version` filter change. We could still look into that route and/or we should do something about the double wrapper on `ContentRepositoryVersionFilter` that might not need to be there. Happy to hear any thoughts.

daviddavis:
It would seem to me based on @acheng-01's investigation that the performance with the q filter probably won't match [adding a base_version filter](https://github.com/pulp/pulpcore/pull/7844). So maybe it's worth pursuing and merging this PR and the base_version filter PR? Any thoughts @gerrod3 @mdellweg?

pulpcore/filters.py
```diff
@@ -206,36 +206,101 @@ def __init__(self, filterset, tokens):
...
+            # opaque and must be applied via its own queryset transform in evaluate().
+            f = self.filter
+            if (
+                type(f).filter is not filters.Filter.filter
+                or getattr(f, "method", None) is not None
+                or getattr(f, "distinct", False)
+            ):
+                return None
+            if self.value in EMPTY_VALUES:
+                return models.Q()
```
mdellweg: What role does "EMPTY_VALUES" play here?
Is it "the filter should not apply"? Would you just "not mention" a filter in "q=" if you didn't want it?
Or does it mean the value is NULL?

Could this mean we needed `models.Q(**{f"{f.field_name}__isnull": not f.exclude})`?

pulpcore/filters.py
```diff
@@ -206,36 +206,101 @@ def __init__(self, filterset, tokens):
...
+            if has_q:
+                result = result.filter(combined)
+            return result
 
     class _OrAction:
         def __init__(self, tokens):
             self.exprs = tokens[0]
             self.complexity = sum((expr.complexity for expr in self.exprs)) + 1
 
+        def as_q(self):
```
mdellweg: So now every subexpression is tried with `as_q()` on succeed we are done, if not, we fall back to `evaluate`. I wish these two methods had names reflecting that.

Ideas:
evaluate_q -> evaluate_opaque
as_q -> as_opaque
evaluate_flat -> evalueate_complex
evaluate_optimized -> evaluate

The catch probably is that the outermost is always called by `evaluate`, right?

mdellweg:
Given the speedups we see here, I'd say we most definitely want this. Looks like a great piece of work. (I'm almost inclined to keep this in limbo for another week to unlock the next level of significant speedup of the parsing. Not actually!)
Too bad, it doesn't sweepingly address your initial issue.
Do you believe we could add a dedicated filter receiving two versions returning the set-difference (maybe that would "just" boil down to merging the base_version into the original 'version_added' filter)?

Drafting from the top of my head (and using a terrible name):
`added_between=/p/a/v3/r/v/3/,/p/a/v3/r/v/6/`

Maybe this is helping (probably only as a blueprint, not as an implementation):
https://django-filter.readthedocs.io/en/stable/ref/filters.html#django_filters.filters.BaseRangeFilter

acheng-01:
> Given the speedups we see here, I'd say we most definitely want this. Looks like a great piece of work. (I'm almost inclined to keep this in limbo for another week to unlock the next level of significant speedup of the parsing. Not actually!) Too bad, it doesn't sweepingly address your initial issue. Do you believe we could add a dedicated filter receiving two versions returning the set-difference (maybe that would "just" boil down to merging the base_version into the original 'version_added' filter)?
> 
> Drafting from the top of my head (and using a terrible name): `added_between=/p/a/v3/r/v/3/,/p/a/v3/r/v/6/`
> 
> Maybe this is helping (probably only as a blueprint, not as an implementation): https://django-filter.readthedocs.io/en/stable/ref/filters.html#django_filters.filters.BaseRangeFilter

Thanks! I'm glad that this turned out well.

Regarding our initial issue, as you mentioned I _think_ what you're proposing is actually what my `base_filter` [PR](#7844) already does. If we were to go the `added_between` / `removed_between` route, it'd really be an alternative packaging of the same logic because both ultimately call `RepositoryVersion.added()/removed()` with a base version.

I did a bit of research on `BaseRangeFilter`, and I believe its contribution here is thin:
• It just splits `"a,b"` into two values and validates the count; its default `range` lookup (SQL `BETWEEN`) doesn't apply, so we'd override `.filter()` anyway.
• We'd still resolve each href/PRN via the existing `get_repository_version()` (called twice) and need two new filters.
- Its only positive is that it binds the two versions together, so the current "base alone is a no-op" scenario becomes impossible.

Since #7844 is more or less ready and tested, I would personally lean toward keeping `base_version` and, if the silent no-op is the concern, I could add a validation error when `base_version` is passed without `repository_version_added / removed`. If you prefer the single-param ergonomics then I can add `added_between / removed_between` to it. Lmk!

pulpcore/filters.py
acheng-01: `EMPTY_VALUES` here means "the filter isn't applied". It's mirroring `django_filters.Filter.filter`, which returns the queryset unchanged for empty values. So the flat equivalent is an identity `models.Q()`, not `IS NULL`. We don't want to use `__isnull` for a few reasons:
• An empty/unspecified value shouldn't be reinterpreted as "field IS NULL" — that would change the meaning of `key=` vs. not mentioning `key` at all.
• The filters that do have null semantics override `.filter()`, so `as_q()` classifies them as opaque and returns `None` — they keep their own `isnull` handling on the `evaluate()` path. So the leaf reduction never needs to deal with NULL.

That said, you did catch something. `as_q()` was using the module-level `EMPTY_VALUES`, which we augment with the `"null"` sentinel. That meant `name=null` would collapse to a no-op (match-all) instead of matching the literal string `"null"` as it does on `main`. In the latest commit, I've fixed it to check django-filter's own `EMPTY_VALUES` so the leaf mirrors `Filter.filter` exactly. The `"null"` sentinel stays reserved for the null-aware filters that actually opt into it.

pulpcore/filters.py
acheng-01: Agreed - thanks for pointing this out. I've renamed `as_q` to `as_flat_q` and kept `evaluate` as the general apply method, plus added docstrings to make the protocol explicit.

And yes, `evaluate()` is the sole entry point; `ExpressionFilter.filter` calls `value.evaluate(qs)` on the outermost node, which then recurses. I also updated the docstring on `ExpressionFilterField` spelling out that two-method protocol so it's a bit clearer on what it does.

mdellweg:
> > Given the speedups we see here, I'd say we most definitely want this. Looks like a great piece of work. (I'm almost inclined to keep this in limbo for another week to unlock the next level of significant speedup of the parsing. Not actually!) Too bad, it doesn't sweepingly address your initial issue. Do you believe we could add a dedicated filter receiving two versions returning the set-difference (maybe that would "just" boil down to merging the base_version into the original 'version_added' filter)?
> > Drafting from the top of my head (and using a terrible name): `added_between=/p/a/v3/r/v/3/,/p/a/v3/r/v/6/`
> > Maybe this is helping (probably only as a blueprint, not as an implementation): https://django-filter.readthedocs.io/en/stable/ref/filters.html#django_filters.filters.BaseRangeFilter
> 
> Thanks! I'm glad that this turned out well.
> 
> Regarding our initial issue, as you mentioned I _think_ what you're proposing is actually what my `base_filter` [PR](#7844) already does. If we were to go the `added_between` / `removed_between` route, it'd really be an alternative packaging of the same logic because both ultimately call `RepositoryVersion.added()/removed()` with a base version.
> 
> I did a bit of research on `BaseRangeFilter`, and I believe its contribution here is thin: • It just splits `"a,b"` into two values and validates the count; its default `range` lookup (SQL `BETWEEN`) doesn't apply, so we'd override `.filter()` anyway. • We'd still resolve each href/PRN via the existing `get_repository_version()` (called twice) and need two new filters.
> 
>     * Its only positive is that it binds the two versions together, so the current "base alone is a no-op" scenario becomes impossible.
> 
> 
> Since #7844 is more or less ready and tested, I would personally lean toward keeping `base_version` and, if the silent no-op is the concern, I could add a validation error when `base_version` is passed without `repository_version_added / removed`. If you prefer the single-param ergonomics then I can add `added_between / removed_between` to it. Lmk!

Exactly. That's what I meant with blueprint. The implementation of the RangeFilter was never going to help us.
The "single-param egonomics" (nicely phrased!) is my personal preference for the api design (after all, to go full circle, it would allow sophisticated use of the filter in the "q-Language"). But I really want others to weigh in.

(I am going to copy this comment over to the issue it is actually talking about, let's continue there.)

